PR #1560, opened 2026-03-22
fix: Dockerfile missing plugin-sdk

### LINKED ISSUE #1111: fix: Dockerfile missing plugin-sdk and lockfile out of sync
## Summary

Two issues that prevent a clean Docker build after the plugin system was introduced:

1. **`@paperclipai/plugin-sdk` missing from Dockerfile** — The `deps` stage doesn't copy `packages/plugins/sdk/package.json`, and the `build` stage doesn't build the SDK before the server. This causes the server build to fail with `TS2307: Cannot find module '@paperclipai/plugin-sdk'` errors.

2. **`pnpm-lock.yaml` out of sync** — `packages/db/package.json` gained `embedded-postgres` as a dependency but the lockfile wasn't updated, causing `pnpm install --frozen-lockfile` to fail in the Docker build.

## Fix

In `Dockerfile`, add to the deps stage:
```dockerfile
COPY packages/plugins/sdk/package.json packages/plugins/sdk/
```

And add to the build stage before the UI and server builds:
```dockerfile
RUN pnpm --filter @paperclipai/plugin-sdk build
```

Then run `pnpm install` to refresh the lockfile.

horsley:
## Summary

Two issues that prevent a clean Docker build after the plugin system was introduced:

1. **@paperclipai/plugin-sdk missing from Dockerfile** — The deps stage doesn't copy packages/plugins/sdk/package.json, and the build stage doesn't build the SDK before the server. This causes the server build to fail with TS2307: Cannot find module '@paperclipai/plugin-sdk' errors.

2. **pnpm-lock.yaml out of sync** — packages/db/package.json gained embedded-postgres as a dependency but the lockfile wasn't updated, causing pnpm install --frozen-lockfile to fail in the Docker build.

## Fix

- Add to the deps stage: 
- Add to the build stage before the UI and server builds: 

Closes #1111

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -20,6 +20,7 @@ COPY packages/adapters/gemini-local/package.json packages/adapters/gemini-local/
 COPY packages/adapters/openclaw-gateway/package.json packages/adapters/openclaw-gateway/
 COPY packages/adapters/opencode-local/package.json packages/adapters/opencode-local/
 COPY packages/adapters/pi-local/package.json packages/adapters/pi-local/
+COPY packages/plugins/sdk/package.json packages/plugins/sdk/
 
 RUN pnpm install --frozen-lockfile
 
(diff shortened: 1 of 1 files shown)


lucas42 (approved):
I tried these changes locally and I went from not being able to use paperclip at all to having a running instance.  👍

PR #1604, opened 2026-03-22
fix(codex-local): handle missing codex binary gracefully in quota service

### LINKED ISSUE #1557: Fix: Handle missing codex binary gracefully in quota-windows service
## Description
  The Paperclip server crashes with an unhandled error when the `codex` binary is not installed, even if no agents use
  the `codex_local` adapter.

  ## Root Cause
  The `quota-windows` service calls `getQuotaWindows()` on all registered adapters (including `codex-local`), regardless
   of whether any agents actually use them. When the codex adapter tries to spawn the `codex` process and the binary is
  missing, Node.js emits an unhandled 'error' event on the child process, crashing the entire Paperclip server.

  ## Error Message
  Error: spawn codex ENOENT
      at ChildProcess._handle.onexit (node:internal/child_process:285:19)
      at onErrorNT (node:internal/process/task_queues:82:16)

  ## When It Occurs
  - During agent heartbeat/background tasks
  - Randomly mid-session when quota windows are being fetched
  - No user action required to trigger

  ## Impact
  - Server crashes unexpectedly
  - Mid-session agent tasks are interrupted
  - Users with Anthropic subscriptions (no codex needed) are affected

  ## Solution
  Add proper error handling to `CodexRpcClient` in `packages/adapters/codex-local/src/server/quota.ts`:
  1. Add an 'error' event handler on the spawn process
  2. Gracefully handle spawn failures instead of letting them crash the process
  3. Return clean rejections to callers so quota system degrades gracefully

  ## Steps to Reproduce
  1. Run Paperclip without the `codex` binary installed
  2. Start an agent task/heartbeat
  3. Server crashes with "spawn codex ENOENT"

  ## Expected Behavior
  Server should gracefully handle missing codex binary and continue running. The quota-windows service should either:
  - Skip adapters that fail to initialize
  - Or return an error result instead of crashing

woosal1337:
## Summary

- Added `error` event handler on the spawned codex process to catch `ENOENT` when the binary is missing
- Added early rejection in `request()` when spawn has already failed, preventing further attempts
- Added guard in `shutdown()` to avoid killing a process that never started
- The existing `getQuotaWindows()` try/catch now properly catches the error and falls back gracefully

**Root cause:** `CodexRpcClient` spawned the codex binary eagerly with no `error` handler. When the binary wasn't installed, Node emitted an unhandled `ENOENT` error that crashed the server — even if no agents used the `codex_local` adapter.

Closes #1557

## Test plan

- [ ] Start the server without `codex` installed — verify no crash
- [ ] Navigate to Costs > Providers — verify page loads (codex quota shows error, not crash)
- [ ] Start the server with `codex` installed — verify quota fetching still works normally
- [ ] Verify other adapters (claude_local, etc.) are unaffected

### DIFF
--- a/packages/adapters/codex-local/src/server/quota.ts
+++ b/packages/adapters/codex-local/src/server/quota.ts
@@ -407,18 +407,27 @@ type PendingRequest = {
 };
 
 class CodexRpcClient {
-  private proc = spawn(
-    "codex",
-    ["-s", "read-only", "-a", "untrusted", "app-server"],
-    { stdio: ["pipe", "pipe", "pipe"], env: process.env },
-  );
-
+  private proc;
   private nextId = 1;
   private buffer = "";
   private pending = new Map<number, PendingRequest>();
   private stderr = "";
+  private spawnError: Error | null = null;
 
   constructor() {
+    this.proc = spawn(
+      "codex",
+      ["-s", "read-only", "-a", "untrusted", "app-server"],
+      { stdio: ["pipe", "pipe", "pipe"], env: process.env },
+    );
+    this.proc.on("error", (err: Error) => {
+      this.spawnError = err;
+      for (const request of this.pending.values()) {
+        clearTimeout(request.timer);
+        request.reject(new Error(`codex binary not available: ${err.message}`));
+      }
+      this.pending.clear();
+    });
     this.proc.stdout.setEncoding("utf8");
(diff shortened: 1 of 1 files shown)


doganalpaslan (approved):
Well-structured fix. Handles missing `codex` binary gracefully: catches spawn error, rejects pending RPC requests with a clear message, and guards `shutdown()`. LGTM.

PR #1611, opened 2026-03-23
feat(ui): add i18n support with pt-BR and en translations (ARP-48)

samuelikz:
- Add LanguageContext and LanguageSelector component for runtime language switching
- Add i18n/en.ts and i18n/pt-BR.ts translation files with full UI coverage
- Wrap app with LanguageProvider in main.tsx
- Update all UI pages and components to use translation keys via useLanguage()


doganalpaslan (approved):
Solid i18n implementation. TypeScript-typed translations with automatic fallback via `deepMerge` is the right call. Architecture is sound — `LanguageProvider` wraps correctly, `t` is memoized, and pt-BR coverage looks complete.

Two items worth addressing (non-blocking):

1. **Wrong key for "select company" state** — `Agents.tsx` uses `t.agents.noAgents` ("No agents found") where the original showed "Select a company to view agents." Add a `t.agents.selectCompany` key (consistent with `t.issues.selectCompany`, `t.activity.selectCompany`).

2. **`buildActionVerbs(t)` perf** — Called on every render per `ActivityRow`, creates a 30-entry object each time. Should be memoized with `useMemo`.

LGTM otherwise.

hfabio (approved):

PR #1612, opened 2026-03-23
feat: add Copilot CLI adapter

twhoff:
- Full adapter package (packages/adapters/copilot-cli) with server execute, JSONL parse, session codec, UI config fields, CLI format, and tests
- Register copilot_cli in server, UI, and CLI adapter registries
- Add Copilot CLI to NewAgentDialog and OnboardingWizard adapter selection
- Fix exit-code detection: treat exit code 1 as normal for Copilot CLI
- Support 18 models (GPT, Claude, Gemini families)

### DIFF
--- a/cli/package.json
+++ b/cli/package.json
@@ -43,6 +43,7 @@
     "@paperclipai/adapter-gemini-local": "workspace:*",
     "@paperclipai/adapter-opencode-local": "workspace:*",
     "@paperclipai/adapter-pi-local": "workspace:*",
+    "@paperclipai/adapter-copilot-cli": "workspace:*",
     "@paperclipai/adapter-openclaw-gateway": "workspace:*",
     "@paperclipai/adapter-utils": "workspace:*",
     "@paperclipai/db": "workspace:*",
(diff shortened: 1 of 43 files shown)


doganalpaslan (changes requested):
Good work — this is a well-structured adapter that follows the existing patterns closely. A few issues to address before merge:

### Must fix

1. **CLI delta deduplication missing** (`format-event.ts`): `parse-stdout.ts` correctly uses a `seenDeltaMessageIds` set to avoid printing `assistant.message` content that was already streamed via `assistant.message_delta`. The CLI `format-event.ts` doesn't implement this — content will be printed twice in terminal. Apply the same dedup pattern.

2. **Fragile full-object regex matching** (`parse.ts`): `isCopilotMaxTurnsResult` and `isCopilotUnknownSessionError` run regex against `JSON.stringify(result)` — this matches on any field name/value. Should check specific fields (e.g., `result.error`, `result.message`) instead.

3. **`normalExit` vs exit code 1** (`execute.ts` ~line 907): PR description mentions treating exit code 1 as normal for Copilot, but the code only treats 0 as normal. Clarify intent and fix the implementation or description.

### Nice to have (non-blocking)

- Add `vitest` to `devDependencies` in `package.json` (other adapters include it explicitly)
- Consider a file size guard on `instructionsFilePath` reads (currently unbounded)
- Confirm `@types/node@^24` is intentional vs rest of monorepo

PR #1616, opened 2026-03-23
Backport company skills and routines for OpenClaw

Alex-Alaniz:
## Summary
- backport the real company skills and routines schema, API, services, and UI onto `release/0.3.1`
- restore company-scoped routing/navigation for `/skills` and `/routines`, including routine detail pages and query keys
- include the OpenClaw backport support needed in this release line (`paperclipSkillSync`, routine issue wakeups, cron helper, and Docker host-gateway wiring)

## Verification
- `pnpm -r typecheck`
- `pnpm test:run`
- `pnpm build`

## Operational Notes
- after adding `host.docker.internal:host-gateway` and resolving device pairing, I could no longer reproduce `gateway websocket open timeout` at the transport layer from the server container; repeated websocket probes reached `connect.challenge` successfully
- remaining gateway stalls appear to happen later in execution and should be investigated separately from this backport


doganalpaslan (changes requested):
Well-structured backport — clean DB schema design, proper timing-safe crypto for webhook verification, and good activity logging. The +17k LOC is mostly the Drizzle migration snapshot + new additive code. Existing code modifications are minimal and low-risk.

### Must fix

**Path traversal in `readLocalSkillImports`** (`server/src/services/company-skills.ts`): `path.resolve(sourcePath)` is called on user-supplied input without bounding it to an allowed directory. An authenticated user could supply `../../../../etc/passwd` and read arbitrary files on the server. Add a check that the resolved path starts with the expected skill storage root before calling `fs.stat`/`fs.readFile`.

### Worth confirming

- The `ne(issues.originKind, 'routine_execution')` filter in `services/issues.ts` silently hides routine-execution issues from the default list. Verify this doesn't break existing API consumers that expect all issues returned.
- The `routines` service instance in `server/src/index.ts` is created inline — a brief comment explaining the lifecycle would help readability.

PR #1619, opened 2026-03-23
pi-local: fallback to stderr when pi --list-models returns on stderr

jazzqi:
- Problem:
  - `pi_local` adapter `discoverPiModels` previously only parsed `pi --list-models` output from `stdout`.
  - In some `pi` versions, the model table is emitted to `stderr` (while `exitCode === 0` and `stdout` is empty), causing model discovery to return empty list and UI shows no models.
- Cause:
  - `pi` CLI writes table output to `stderr` for discovery output in this runtime path.
  - existing code path did not handle stderr fallback and thus treated successful `pi` run as no model data.
- Fix:
  - In models.ts, when `exitCode===0`, use:
    - `rawOutput = result.stdout.trim().length > 0 ? result.stdout : result.stderr`
    - parse `rawOutput` instead of `stdout` only.
  - keep existing error/timeout behavior.
- Result:
  - `listPiModels` now correctly returns available models when `pi --list-models` output is on stderr.
  - UI adapter model dropdown now populates for pi_local in this case.


doganalpaslan (approved):
Clean fix. Some `pi` versions write model output to stderr — this correctly falls back to stderr when stdout is empty. LGTM.

PR #1621, opened 2026-03-23
fix: surface full initdb stderr on embedded PostgreSQL failure

### LINKED ISSUE #1032: Embedded PostgreSQL initdb fails silently on WSL2 (Ubuntu) — paperclipai run unusable
### Environment

- **OS:** Ubuntu 24.x on WSL2 (Windows)
- **Kernel:** `Linux 6.6.87.2-microsoft-standard-WSL2 x86_64`
- **Install method:** `npx paperclipai onboard --yes`
- **Paperclip version:** Latest via npx (v0.3.1 as of 2026-03-16)
- **Node:** (system default on Ubuntu WSL2)

### Steps to reproduce

1. `npx paperclipai onboard --yes` — completes successfully
2. `npx paperclipai doctor` — all 9 checks pass (including DB and port checks)
3. `npx paperclipai run` — fails immediately

### Actual behavior

```
◇  Starting Paperclip server...
Paperclip server failed to start.
Postgres init script exited with code 1. Please check the logs for extra info. The data directory might already exist.
[09:36:06] INFO: Using embedded PostgreSQL because no DATABASE_URL set (dataDir=~/.paperclip/instances/default/db, port=54329)
[09:36:06] ERROR: Embedded PostgreSQL failed; showing buffered startup logs {"phase":"initialise","recentLogs":["The files belonging to this database system will be owned by user \"<user>\".","This user must also own the server process."]}
    err: "Postgres init script exited with code 1. Please check the logs for extra info. The data directory might already exist."
```

### What I've tried

- `rm -rf ~/.paperclip/instances/default/db` (multiple times) — directory is recreated, same failure
- Verified port 54329 is not in use (`lsof -i :54329` returns nothing)
- `paperclipai doctor` passes all checks including DB and port

### Analysis

The `recentLogs` array only contains 2 lines from `initdb` output before exit — the standard ownership preamble lines. The **actual failure reason is not captured or surfaced.** `initdb` is dying very early, and the error handler only buffers the last N lines of stdout, missing stderr or the real error.

Possible root causes on WSL2:
- Missing locale/collation data required by `initdb` (common on minimal WSL Ubuntu installs)
- The embedded Postgres binary (likely from `embedded-postgres` or similar npm package) may be incompatible with WSL2's kernel or libc version
- `initdb` flags may be incorrect or missing required args for this environment

### Related

- Branch `fix/embedded-postgres-initdbflags` exists (Mar 13, 2026) — suggests the team is already aware of initdb flag issues. If that branch addresses this, a release would fix it.

### Suggestions

1. **Surface the full `initdb` stderr** in the error log — the current 2-line buffer is insufficient for debugging
2. **Run `initdb` with `--debug`** or at minimum capture stderr separately so users can self-diagnose

MayurVirkar:
## Thinking Path

> - Paperclip ships an embedded PostgreSQL so users don't need to install Postgres separately
> - But embedded Postgres needs to run `initdb` to initialise a fresh data cluster on first launch
> - On some environments (notably WSL2 / minimal Ubuntu), `initdb` fails due to missing locale data or libc incompatibilities
> - When `initdb` fails, the error output (stderr) is either silently swallowed or only partially captured — users see a cryptic "Postgres init script exited with code 1" with just 2 lines of preamble and no actual root cause
> - This makes it impossible for users to self-diagnose the problem
> - This PR surfaces the full buffered postgres logs in error messages across all 3 callsites, and adds actionable hints so users can debug initdb failures themselves

### `server/src/index.ts`
- **Removed the buffer-length guard** on `logEmbeddedPostgresFailure` — errors are now always logged to pino, even if no postgres output was captured (previously silently dropped)
- **Added `enrichInitError()` helper** — wraps the thrown error with: the original error message, all buffered postgres logs, and a hint to use `PAPERCLIP_EMBEDDED_POSTGRES_VERBOSE=true` or run `initdb` manually for full stderr

### `packages/db/src/migration-runtime.ts`
- **Replaced silent `() => {}` log handlers** with actual log capture using the same circular buffer pattern from the server
- **Error messages now include buffered postgres logs** when `initialise()` or `start()` fails

### `cli/src/commands/worktree.ts`
- **Same fix** — replaced silent `() => {}` handlers with log capture
- **Added proper try/catch** around `initialise()` and `start()` (previously unhandled — would throw raw errors with no context)
- Error messages include buffered logs and use `{ cause: err }` for proper error chaining

## Why it matters

Without this fix, any user whose `initdb` fails (WSL2, minimal Linux installs, libc mismatches) gets an error message with zero diagnostic information. They cannot tell whether the problem is a missing locale, a permission issue, or an incompatible binary — and neither can anyone helping them.

With this fix, the full postgres output is included in the error, and users get an actionable hint to re-run with verbose logging or run `initdb` manually.

## How to verify

1. Simulate an `initdb` failure (e.g., corrupt the embedded postgres binary path, or run on a minimal container missing locale data)
2. Run `npx paperclipai run` — the error message should now include the full buffered postgres logs and hint text
3. Existing tests still pass (`vitest run cli/src/__tests__/worktree.test.ts`, `vitest run packages/db/src/runtime-config.test.ts`)

## Risks

- **Low risk** — no behavioral change on the happy path; only error messages are improved on failure
- No new dependencies
- No API/schema changes

Fixes #1032

### DIFF
--- a/cli/src/commands/worktree.ts
+++ b/cli/src/commands/worktree.ts
@@ -750,6 +750,16 @@ async function ensureEmbeddedPostgres(dataDir: string, preferredPort: number): P
   }
 
   const port = await findAvailablePort(preferredPort);
+  const logBuffer: string[] = [];
+  const appendLog = (message: unknown) => {
+    const text = typeof message === "string" ? message : message instanceof Error ? message.message : String(message ?? "");
+    for (const line of text.split(/\r?\n/)) {
+      const trimmed = line.trim();
+      if (trimmed) logBuffer.push(trimmed);
+    }
+    if (logBuffer.length > 120) logBuffer.splice(0, logBuffer.length - 120);
+  };
+
   const instance = new EmbeddedPostgres({
     databaseDir: dataDir,
     user: "paperclip",
@@ -757,17 +767,33 @@ async function ensureEmbeddedPostgres(dataDir: string, preferredPort: number): P
     port,
     persistent: true,
     initdbFlags: ["--encoding=UTF8", "--locale=C"],
-    onLog: () => {},
-    onError: () => {},
+    onLog: appendLog,
+    onError: appendLog,
   });
 
   if (!existsSync(path.resolve(dataDir, "PG_VERSION"))) {
-    await instance.initialise();
+    try {
+      await instance.initialise();
+    } catch (err) {
+      const logs = logBuffer.length > 0 ? `\nPostgres logs:\n${logBuffer.join("\n")}` : "";
+      throw new Error(
+        `Failed to initialise embedded PostgreSQL cluster in ${dataDir} on port ${port}${logs}`,
+        { cause: err },
+      );
+    }
   }
   if (existsSync(postmasterPidFile)) {
     rmSync(postmasterPidFile, { force: true });
   }
-  await instance.start();
+  try {
+    await instance.start();
+  } catch (err) {
+    const logs = logBuffer.length > 0 ? `\nPostgres logs:\n${logBuffer.join("\n")}` : "";
+    throw new Error(
+      `Failed to start embedded PostgreSQL on port ${port}${logs}`,
+      { cause: err },
+    );
+  }
 
   return {
     port,
--- a/packages/db/src/migration-runtime.ts
+++ b/packages/db/src/migration-runtime.ts
@@ -144,6 +144,16 @@ async function ensureEmbeddedPostgresConnection(
     };
   }
 
+  const logBuffer: string[] = [];
+  const appendLog = (message: unknown) => {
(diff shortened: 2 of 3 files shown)


doganalpaslan (approved):
Thorough fix across all three embedded Postgres entry points (server, migration-runtime, cli/worktree). The circular log buffer (capped at 120 lines) is a good approach — captures enough context without unbounded growth. The enriched error messages with buffered logs + the manual initdb hint in the server path will make debugging much easier. LGTM.

One minor note: the `appendLog` helper is duplicated across the three files. Not a blocker — can be extracted to a shared util in a follow-up if desired.

PR #1623, opened 2026-03-23
fix: secure cookies over HTTP, inline attachments for auth deployments

### LINKED ISSUE #1598: baseUrlMode: auto should disable secure cookies over HTTP
## Problem

When `auth.baseUrlMode` is set to `auto` (the default) and the server is accessed over HTTP, better-auth still sets `__Secure-` prefixed cookies with the `Secure` flag. Browsers silently refuse to store these cookies over non-HTTPS connections, making login impossible.

## Root Cause

Paperclip doesn't pass `baseURL` to better-auth. The secure cookie decision falls through to `isProduction` (`NODE_ENV === 'production'`). When running in production mode over HTTP (common for homelab/LAN deployments behind no reverse proxy or with a separate TLS terminator), login breaks silently — the sign-in API returns 200 with a valid session token, but the browser never stores the cookie.

## Expected Behavior

When `baseUrlMode: auto`, Paperclip should detect the request scheme (or the configured base URL) and:
- Set `useSecureCookies: false` when the resolved base URL is `http://`
- Set `useSecureCookies: true` when the resolved base URL is `https://`

This would allow `NODE_ENV=production` to work correctly over HTTP without requiring users to override the environment variable.

## Current Workaround

Setting `NODE_ENV=development` in the systemd unit file. This works but disables other production behaviors.

## Environment

- Paperclip version: 2026.318.0
- Install: npm global, systemd user service
- Access: LAN HTTP (no TLS)
- Auth: better-auth with embedded Postgres

Related: #1597 (password reset)

### LINKED ISSUE #1600: Images on authenticated server version don't get passed to AI correctly
When uploading an image to the issue (in a server environment) the image URL is protected, therefore the AI can't see it.

It needs to send images as base64 inline when the deployment is authenticated, not as URLs so that the AI Agent can see it

Image

doganalpaslan:
## Summary

- **#1598 — Secure cookies over HTTP:** In `baseUrlMode: auto`, better-auth falls back to `NODE_ENV` to decide cookie security, setting `__Secure-` cookies over HTTP (browsers silently drop them → login broken). Fix: construct `baseURL` from server listen address and pass it explicitly with `useSecureCookies` derived from the protocol.

- **#1600 — Images not accessible to agents on authenticated deployments:** Agents can't fetch protected attachment URLs. Added `?inline=true` query parameter to `GET /api/issues/:id/attachments` that returns `contentBase64` field with base64-encoded content.

- **Bonus: CLAUDECODE env leak:** Drop `CLAUDECODE` env var from child process environment to prevent conflicts when the server runs inside Claude Code.

## Test plan

- [ ] Deploy in HTTP mode with `baseUrlMode: auto` → verify login works (no `__Secure-` cookies set)
- [ ] Deploy in HTTPS mode → verify `__Secure-` cookies are still used
- [ ] On authenticated deployment, call `GET /api/issues/:id/attachments?inline=true` → verify `contentBase64` is returned
- [ ] Verify agents can pass inline images to AI models
- [ ] Run TypeScript type-check (passes clean)

Closes #1598, closes #1600

### DIFF
--- a/packages/adapter-utils/src/server-utils.ts
+++ b/packages/adapter-utils/src/server-utils.ts
@@ -219,7 +219,8 @@ export async function runChildProcess(
   const onLogError = opts.onLogError ?? ((err, id, msg) => console.warn({ err, runId: id }, msg));
 
   return new Promise<RunProcessResult>((resolve, reject) => {
-    const mergedEnv = ensurePathInEnv({ ...process.env, ...opts.env });
+    const { CLAUDECODE: _dropped, ...parentEnv } = process.env;
+    const mergedEnv = ensurePathInEnv({ ...parentEnv, ...opts.env });
     const child = spawn(command, args, {
       cwd: opts.cwd,
       env: mergedEnv,
--- a/packages/adapters/claude-local/src/server/execute.ts
+++ b/packages/adapters/claude-local/src/server/execute.ts
@@ -204,7 +204,8 @@ async function buildClaudeRuntimeConfig(input: ClaudeExecutionInput): Promise<Cl
     env.PAPERCLIP_API_KEY = authToken;
   }
 
-  const runtimeEnv = ensurePathInEnv({ ...process.env, ...env });
+  const { CLAUDECODE: _dropped, ...parentEnv } = process.env;
+  const runtimeEnv = ensurePathInEnv({ ...parentEnv, ...env });
   await ensureCommandResolvable(command, cwd, runtimeEnv);
 
   const timeoutSec = asNumber(config.timeoutSec, 0);
--- a/server/src/auth/better-auth.ts
+++ b/server/src/auth/better-auth.ts
@@ -43,7 +43,7 @@ function headersFromExpressRequest(req: Request): Headers {
 }
 
 export function deriveAuthTrustedOrigins(config: Config): string[] {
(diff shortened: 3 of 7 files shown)


jerelvelarde (approved):
## ✅ Security Review Approved

Reviewed all security-sensitive changes. This PR is safe to merge.

### Secure Cookie Fix
- **Correct fix** for `__Secure-` cookies over HTTP breaking login
- Properly derives `useSecureCookies` from protocol instead of relying on NODE_ENV
- No auth bypass risk, fixes broken auth flow

### Inline Attachment Feature
- ✅ Existing ACL check (`assertCompanyAccess`) protects access before content retrieval
- ✅ No auth bypass: only returns attachments for authorized issues
- ✅ No injection risk: base64 encoding is read-only
- ✅ Safe error handling with graceful degradation

### Backwards Compatibility
- ✅ Cookie behavior unchanged for working HTTPS deployments
- ✅ Inline attachment API is additive-only (optional query param)
- ✅ CLAUDECODE cleanup has no external impact

### Test Coverage
- CI checks passing ✓
- Test plan covers both HTTP and HTTPS modes
- Inline attachment test plan included

**Recommendation:** Approve and merge. No security concerns identified.

PR #1654, opened 2026-03-23
fix(runtime): improve agent recovery and heartbeat operations

cryppadotta:
## Thinking Path

- Paperclip only works if agents can reliably recover context, resume workspaces, and expose their operational state to the board.
- Small runtime misses here turn into confusing agent behavior: lost instructions, stale workspace reuse, hidden live runs, or restarts triggered by Paperclip-owned paths.
- This PR focuses on those operational edges instead of new product surface area.
- The goal is to make agent execution more predictable for both operators and local adapters.
- The result is better recovery behavior, clearer heartbeat visibility, and more explicit Codex instruction-source reporting.

## What changed

- fixed isolated issue workspace reuse and explicit failed-run session resume
- switched agent history views to issue participation instead of narrower heuristics
- preserved and recovered agent instruction files more reliably across route and service paths
- ignored Paperclip-owned and test-only paths in dev restart tracking
- showed all company agents on the instance heartbeats page and added a Disable All control
- documented that Codex can still auto-apply repo AGENTS.md files even when Paperclip injects explicit instructions
- addressed Greptile follow-ups by adding a confirmation step to Disable All and making the bulk heartbeat disable concurrent with aggregated failure reporting

## Testing

- pnpm -r typecheck

## Risks

- Main risk is in heartbeat and runtime behavior where regressions may only appear during longer local-agent sessions
- I intentionally left the instance-admin gate in place for the heartbeats instance page because the instance settings surface itself is already instance-admin-only

### DIFF
--- a/packages/shared/src/validators/agent.ts
+++ b/packages/shared/src/validators/agent.ts
@@ -73,6 +73,7 @@ export const updateAgentSchema = createAgentSchema
   .partial()
   .extend({
     permissions: z.never().optional(),
+    replaceAdapterConfig: z.boolean().optional(),
     status: z.enum(AGENT_STATUSES).optional(),
     spentMonthlyCents: z.number().int().nonnegative().optional(),
   });
--- a/server/src/__tests__/agent-instructions-routes.test.ts
+++ b/server/src/__tests__/agent-instructions-routes.test.ts
@@ -274,4 +274,45 @@ describe("agent instructions bundle routes", () => {
       expect.any(Object),
     );
   });
+
+  it("replaces adapter config when replaceAdapterConfig is true", async () => {
+    mockAgentService.getById.mockResolvedValue({
+      ...makeAgent(),
+      adapterType: "codex_local",
+      adapterConfig: {
+        instructionsBundleMode: "managed",
+        instructionsRootPath: "/tmp/agent-1",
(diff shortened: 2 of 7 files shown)


cryppadotta:
Addressed the latest Greptile review in follow-up commit `c8f8f675`.

- fixed managed instruction bundle self-healing so write/delete operations now persist recovered `instructionsRootPath` / `instructionsEntryFile` metadata instead of writing stale config back to the DB
- treat legacy persisted execution workspace rows with `mode = null` / `undefined` as `agent_default` so reuse-existing issues do not get silently downgraded to `shared_workspace`
- restored an explicit full-replace path for agent config patches via `replaceAdapterConfig: true` while keeping the safer merge behavior as the default, and guarded bundle-key removal behind the existing instructions-path permission check
- added coverage for the healed bundle metadata path, the `null` workspace-mode fallback, and the new replace-path behavior

Verification:
- `pnpm -r typecheck`
- `pnpm build`
- `pnpm exec vitest run --maxWorkers 1`

Note: the repo's default `pnpm test:run` still flakes in parallel on three embedded-Postgres suites with `Postgres init script exited with code 1 ... data directory might already exist`; the same full suite passed serially.

@greptile-apps please review again.

Issue #1625, opened 2026-03-23
bug(gemini_local): hello probe default 10s timeout kills CLI during stream-json / relaunch (misleading error)


shuanat:
## Summary

The **gemini_local** adapter **environment check** (`hello` probe) can **fail with a misleading error** even when Gemini CLI is installed and authenticated. The stderr may show:

`Fatal error: Failed to relaunch the CLI process.`

This is easy to misread as an OAuth/auth problem. In practice it is often caused by the **default 10s probe timeout** being **too short** for the probe’s actual CLI invocation (`--output-format stream-json`), so the child process is **SIGTERM’d** while `@google/gemini-cli` is still starting or during its internal relaunch flow.

## Duplicate check (before filing)

Searched open issues in `paperclipai/paperclip` (GitHub search / `gh issue list`) for: `helloProbeTimeout`, `relaunch CLI`, `gemini hello probe`, `stream-json`, `adapter environment`. **No existing issue** describes this timeout vs `stream-json` / relaunch behavior. Related but **not duplicate**: #1505 (stale Gemini session resume).

## Expected behavior

- Agent setup **“Test now”** for Gemini CLI should pass when `gemini` works from the same host user, cwd, and auth as the Paperclip server.
- Error messages should point to **timeout** when the process was killed mid-flight, not only to API keys / `gemini auth`.

## Actual behavior

- `packages/adapters/gemini-local/src/server/test.ts` uses `helloProbeTimeoutSec` default **`10`** and runs:
  - `gemini --output-format stream-json --prompt "Respond with hello."` (plus model/sandbox args).
- On a typical Linux host, a successful run can take **~10–12s+** (cold start, models, etc.). At **10s**, `runChildProcess` sends **SIGTERM** → Gemini CLI may print **`Failed to relaunch the CLI process`** (from `GEMINI_CLI_NO_RELAUNCH` / relaunch path in `@google/gemini-cli`).
- The UI **manual debug** text references `gemini --output-format json "Respond with hello."`, which is **not identical** to the probe (often faster), so **manual success + UI failure** is common.

## Suggested fixes (any or all)

1. **Raise the default** `asNumber(config.helloProbeTimeoutSec, 10)` to something safer (e.g. **60**), or make it adaptive.
2. Document **`helloProbeTimeoutSec`** in `agentConfigurationDoc` for `gemini_local` so operators can tune it without reading source.
[...]

PR #1663, opened 2026-03-23
fix(gemini-local): support assistant message events

mrbob-git:
Thinking path
- Paperclip uses Gemini stream output in more than one place: the environment probe, the server-side result parser, and the transcript/CLI formatting paths.
- The older Gemini stream shape used `type: "assistant"` with message content nested under `message`.
- But newer Gemini CLI output can emit `type: "message"` with `role: "assistant"` and the text under `content`.
- That means Paperclip can miss valid assistant output even when Gemini actually returned the right answer.
- I focused this PR on that parser mismatch only, without mixing in the timeout or sandbox parts of #1625.

What I changed
- Updated the Gemini server parser to accept assistant text from the newer `type: "message"` / `role: "assistant"` event shape.
- Updated the UI stdout parser and CLI formatter to recognize the same newer message shape too.
- Added regression coverage for the new event format.

Why this matters
- It fixes one concrete root cause behind the misleading Gemini hello-probe failures in #1625.
- It also keeps the UI and CLI views in sync with the server-side parser instead of only fixing one surface.

How I verified it
- `pnpm exec vitest run server/src/__tests__/gemini-local-adapter.test.ts`
- `pnpm --filter @paperclipai/adapter-gemini-local typecheck`

Risks
- Low risk. This only broadens Gemini event parsing to support an additional stream shape.
- I did not change the timeout or sandbox behavior in this PR.

Related to #1625

### DIFF
--- a/packages/adapters/gemini-local/src/cli/format-event.ts
+++ b/packages/adapters/gemini-local/src/cli/format-event.ts
@@ -13,6 +13,10 @@ function asNumber(value: unknown, fallback = 0): number {
   return typeof value === "number" && Number.isFinite(value) ? value : fallback;
 }
 
+function readMessagePayload(value: Record<string, unknown>): unknown {
+  return value.content ?? value.message;
+}
+
 function stringifyUnknown(value: unknown): string {
   if (typeof value === "string") return value;
   if (value === null || value === undefined) return "";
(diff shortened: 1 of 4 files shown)


gcampton:
Comment on #3498 and #1663 with "I hit this today, confirmed the fix works — here's the reproducer output" to bump maintainer attention. Helpful for the community, no new PR churn.
image

PR #1676, opened 2026-03-24
feat: add Google Sign-In via better-auth

DNYoussef:
## Summary
- Add `socialProviders.google` configuration to `server/src/auth/better-auth.ts` (client ID + secret from env vars)
- Add "Sign in with Google" button with Google SVG icon to `ui/src/pages/Auth.tsx`

Uses better-auth's built-in social provider support. Requires `GOOGLE_CLIENT_ID` and `GOOGLE_CLIENT_SECRET` environment variables.

## Test plan
- [ ] Set `GOOGLE_CLIENT_ID` and `GOOGLE_CLIENT_SECRET` env vars
- [ ] Verify Google Sign-In button renders on the auth page
- [ ] Click button and confirm OAuth redirect to Google
- [ ] Complete sign-in flow and verify user is authenticated

### DIFF
--- a/server/src/auth/better-auth.ts
+++ b/server/src/auth/better-auth.ts
@@ -86,13 +86,16 @@ export function createBetterAuthInstance(db: Db, config: Config, trustedOrigins?
         verification: authVerifications,
       },
     }),
-    socialProviders: {
-      google: {
-        clientId: process.env.GOOGLE_CLIENT_ID ?? "",
-        clientSecret: process.env.GOOGLE_CLIENT_SECRET ?? "",
-        enabled: !!(process.env.GOOGLE_CLIENT_ID && process.env.GOOGLE_CLIENT_SECRET),
-      },
-    },
+    ...(process.env.GOOGLE_CLIENT_ID && process.env.GOOGLE_CLIENT_SECRET
+      ? {
+          socialProviders: {
+            google: {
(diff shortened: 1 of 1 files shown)


DNYoussef:
Fixed the Greptile feedback in 2edc0a6c:

- Removed invalid `enabled` field from socialProviders.google
- Conditionally include the entire `socialProviders` block only when both `GOOGLE_CLIENT_ID` and `GOOGLE_CLIENT_SECRET` env vars are present
- When unset, the auth config is unchanged (email/password only)

@jacobsamo Good call on the docs -- will add a setup guide for Google sign-in configuration as a follow-up.

ui/src/pages/Auth.tsx
```diff
@@ -155,6 +137,45 @@ export function AuthPage() {
...
+            <div className="relative flex justify-center text-xs">
+              <span className="bg-background px-2 text-muted-foreground">or</span>
+            </div>
+          </div>
+
+          <Button
+            type="button"
+            variant="outline"
+            className="w-full"
+            onClick={() => {
```
jacobsamo, line 153: This should use the `  const data = await authClient.signIn.social({
    provider: "google",
  });`  https://better-auth.com/docs/authentication/google#usage removes the need for callback url as better auth will handle it 

And as referenced below this button will always show even if google sign in is not enabled

PR #1677, opened 2026-03-24
fix: prevent activity_log FK violation when run_id not yet in heartbeat_runs

cathrynlavery:
## Thinking Path

- Paperclip orchestrates AI agents for zero-human companies
- Agents perform work by checking out issues, posting comments, and logging activity
- All of these actions are recorded in the `activity_log` table via `logActivity`
- `activity_log.run_id` has a FK constraint referencing `heartbeat_runs.id`
- Gateway adapters (openclaw, http) generate run IDs at the gateway level and send them via `X-Paperclip-Run-Id` before the heartbeat system has had a chance to register the run
- This means `logActivity` is called with a `runId` that doesn't yet exist in `heartbeat_runs`
- The FK constraint fires, the request returns 500, and the agent's checkout/comment/update fails
- This PR fixes `logActivity` to upsert a stub row before the `activity_log` insert so the FK is always satisfied

## What changed

Single file: `server/src/services/activity-log.ts`

Before every `activity_log` insert, if `runId` and `agentId` are both present, we upsert a minimal stub row into `heartbeat_runs` using `onConflictDoNothing`. The heartbeat system owns these rows and overwrites the stub once the real run record arrives.

Three correctness improvements:

1. **runId-without-agentId safety**: If `runId` is present but `agentId` is absent (can't satisfy `NOT NULL` on `heartbeat_runs.agent_id`), `runId` is dropped from the `activity_log` insert rather than written as an unresolvable FK reference.

2. **In-process dedup cache**: A module-level `Set<string>` tracks run IDs already confirmed/stubbed, so the upsert DB round-trip only fires once per run ID per process lifetime — not on every `logActivity` call.

3. **Comments**: Explains why `status: "running"` is safe as a placeholder and that the heartbeat system owns the row going forward.

## Verification

Reproduced locally with openclaw gateway agents: checkouts and comments returned `PostgresError: insert or update on table "activity_log" violates foreign key constraint "activity_log_run_id_heartbeat_runs_id_fk"`. After this fix, gateway agent runs complete successfully.

## Related

Partially addresses #1136 (the FK crash path is fixed here; non-UUID header validation is a separate concern).


cathrynlavery:
Addressed Greptile follow-ups in `db29477720`:

- bounded `_ensuredRunIds` with `MAX_ENSURED_RUN_IDS` and clear-on-limit behavior to prevent unbounded long-lived server growth
- clarified that stub insertion is for externally minted run IDs and is a no-op for normal heartbeat-owned runs already present in `heartbeat_runs`

Verification:
- `git diff --check`
- `pnpm -C server typecheck`

@greptile-apps please re-review.

cathrynlavery:
Additional hourly sweep update:

- Merged latest `paperclipai/paperclip:master` into this PR branch with `9c8d0c9` and resolved the only conflict in `server/src/services/activity-log.ts`.
- Conflict markers are absent and the PR branch has been pushed normally (no force-push).
- `git diff --check` after the base merge is blocked by pre-existing trailing-whitespace lines brought in from upstream/base files such as `adapter-plugin.md`, `packages/adapters/pi-local/src/ui/parse-stdout.ts`, and `ui/src/components/HermesIcon.tsx`; I did not alter those unrelated files in this PR sweep.
- `pnpm -C server typecheck` passed before the base merge on the PR head with the Greptile fixes. After merging current `master`, it is blocked by existing plugin-SDK/environment typing drift (first error: `src/routes/plugins.ts(1508,9): Argument of type '"handleApiRequest"' is not assignable to parameter of type 'keyof HostToWorkerMethods'`).

@greptile-apps please re-review the latest head.

PR #1714, opened 2026-03-24
fix: enable Express trust proxy for reverse-proxy deployments

### LINKED ISSUE #1690: Login returns 500 behind reverse proxy (missing trust proxy)
## Bug

`POST /api/auth/sign-in/email` returns `500 Internal Server Error` when accessed through a reverse proxy (e.g., Fly.io). The same request to `localhost:3100` returns `200` with correct session token.

## Root Cause

Express does not have `trust proxy` enabled. When behind a reverse proxy (Fly.io, nginx, etc.), Express sees the request as HTTP even though the client connected via HTTPS. This causes better-auth's `setSessionCookie` to fail when trying to set a `__Secure-` prefixed cookie, since those require a secure context.

## Steps to Reproduce

1. Deploy Paperclip on Fly.io (or any reverse proxy with HTTPS termination)
2. Configure `publicBaseUrl` with `https://`
3. Try to sign in via `POST /api/auth/sign-in/email` from an external client
4. Response: `{"error":"Internal server error"}` with status 500
5. Same request to `localhost:3100` directly: status 200, session created

## Suggested Fix

Add `app.set('trust proxy', true)` (or make it configurable) when the server starts.

## Environment

- Paperclip version: latest (v2026.318.0)
- Express version: 5.1.0
- Deployment: Fly.io (HTTPS termination at proxy)

doganalpaslan:
## Summary

- Adds `app.set("trust proxy", ...)` to Express server startup, fixing login 500 errors behind reverse proxies (Fly.io, nginx, etc.)
- Auto-enables when `deploymentMode !== "local_trusted"` — no action needed for existing proxy deployments
- Configurable via `PAPERCLIP_TRUST_PROXY` env var for explicit control

## Root cause

Without `trust proxy`, Express sees HTTPS-terminated requests as HTTP, causing better-auth to fail when setting `__Secure-` cookies (which require a secure context).

Fixes #1690

## Test plan

- [ ] Deploy behind nginx/Fly.io and verify login works without 500
- [ ] Verify `local_trusted` mode does NOT enable trust proxy by default
- [ ] Verify `PAPERCLIP_TRUST_PROXY=false` disables even in authenticated mode
- [ ] Verify `PAPERCLIP_TRUST_PROXY=loopback` passes string value through

### DIFF
--- a/server/src/app.ts
+++ b/server/src/app.ts
@@ -72,11 +72,16 @@ export async function createApp(
     instanceId?: string;
     hostVersion?: string;
     localPluginDir?: string;
+    trustProxy?: boolean | string;
     betterAuthHandler?: express.RequestHandler;
     resolveSession?: (req: ExpressRequest) => Promise<BetterAuthSessionResult | null>;
   },
 ) {
   const app = express();
+
+  if (opts.trustProxy !== undefined && opts.trustProxy !== false) {
+    app.set("trust proxy", opts.trustProxy);
+  }
 
   app.use(express.json({
     // Company import/export payloads can inline full portable packages.
--- a/server/src/config.ts
+++ b/server/src/config.ts
@@ -73,6 +73,7 @@ export interface Config {
   heartbeatSchedulerEnabled: boolean;
   heartbeatSchedulerIntervalMs: number;
   companyDeletionEnabled: boolean;
+  trustProxy: boolean | string;
 }
 
 export function loadConfig(): Config {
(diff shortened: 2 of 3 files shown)


chrisw225tw-lab (approved):
**Reviewed: LGTM — merge conflict resolved, ready to merge.**

The fix is correct and addresses the root cause: Express without `trust proxy` treats HTTPS-terminated requests as HTTP, causing better-auth to fail when setting `__Secure-` cookies.

The implementation is clean:
- Auto-enables for non-`local_trusted` deployments (the common reverse-proxy case)
- Explicit override via `PAPERCLIP_TRUST_PROXY` env var
- `local_trusted` stays unaffected

**Merge conflict resolution** (master has restructured `index.ts` and expanded `express.json`):

`server/src/app.ts` — keep both opts and the expanded json middleware:
```ts
// opts:
trustProxy?: boolean | string;
instanceId?: string;
hostVersion?: string;
localPluginDir?: string;

// startup:
if (opts.trustProxy !== undefined && opts.trustProxy !== false) {
  app.set("trust proxy", opts.trustProxy);
}
app.use(express.json({
  limit: "10mb",
  verify: (req, _res, buf) => { (req as any).rawBody = buf; },
}));
```

`server/src/index.ts` — in the master-restructured `createApp` call, add:
```ts
trustProxy: config.trustProxy,
```

Issue #1678, opened 2026-03-24
POST /api/issues/:id/release strips assigneeAgentId


sherifkozman:
## Summary

Calling `POST /api/issues/:id/release` to clear a stale execution lock also clears the `assigneeAgentId` field. This causes assigned tasks to become orphaned — they remain in `todo`/`in_progress` with no owner, so no agent picks them up.

## Steps to Reproduce

1. Agent checks out an issue (issue has `assigneeAgentId` set + `executionLockedAt` set)
2. Agent's heartbeat completes but the execution lock is not released (stale lock)
3. Board calls `POST /api/issues/:id/release` to clear the stale lock
4. **Result**: `executionLockedAt` is cleared (correct), but `assigneeAgentId` is also cleared (bug)

## Evidence

During a production session with 6 agents across 2 companies (NAYA + ES), we observed:

- 9 issues had stale execution locks after agent heartbeats completed
- After calling `/release` on these issues, all lost their `assigneeAgentId`
- Issues that were previously assigned to Oscar, Ernest, CMO, and Eve became unassigned
- These orphaned issues required manual reassignment to resume work

Specific examples:
- `NAY-51`: Was assigned to Ernest, lock released, became unassigned. Ernest's work was complete but task stayed `todo`.
- `NAY-37`: Was assigned to Ernest, lock released, became unassigned with completed work in comments.
- `NAY-46`: Was assigned to Eve agent, lock released, fell back to `assigneeUserId: local-board`.

## Expected Behavior

`POST /api/issues/:id/release` should:
- Clear `executionLockedAt`
- Clear `executionRunId`
- Clear `checkoutRunId`
- **Preserve `assigneeAgentId`** — the assignee is who OWNS the task, not who is currently executing it

## Additional Note

Stale execution locks were also observed — agents complete heartbeats but locks persist. This may be a separate issue (lock not released on heartbeat exit), which then requires the `/release` call that triggers this bug.

## Impact

High — in a multi-agent system, orphaned tasks create open loops. Agents don't pick up unassigned work (by design), so these tasks stall until a human notices and reassigns.

Issue #1420, opened 2026-03-21
Stale execution locks cannot be cleared by agents — no admin/force-release endpoint


shivasymbl:
## Problem

When a heartbeat run finishes (`succeeded` or `cancelled`) but fails to release the execution lock on an issue, the issue becomes permanently locked. No agent — including the CEO — can modify, release, or reassign it via the API. The only fix is direct database access.

## Reproduction

1. An issue gets `execution_run_id` set to a run that transitions to `succeeded` or `cancelled` without clearing the lock
2. Any agent trying `POST /api/issues/{id}/release` gets **409** because their `actorRunId` doesn't match `executionRunId`
3. `PATCH /api/issues/{id}` also fails with ownership conflict
4. `POST /api/issues/{id}/checkout` fails because there's an existing execution lock
5. The issue is stuck — no API endpoint can clear it

## Impact

In our deployment (Mindcat Group on Paperclip v2026.318.0), this caused a CEO agent (Atlas, Opus 4.6) to spin for **898 minutes (15 hours)** trying every possible API approach to clear the lock, burning significant budget. The lock could only be cleared via direct PostgreSQL:

```sql
UPDATE issues SET execution_run_id = NULL, execution_locked_at = NULL, checkout_run_id = NULL WHERE identifier = 'MIN-28';
```

## Root Cause

Run `c0fc4774` completed with status `succeeded` (finished at 18:16:04 UTC) but failed to release the execution lock on the issue afterward. This is a cleanup bug in the execution lifecycle — the lock persisted after run completion.

## Proposed Solution

Add a force-release mechanism. Options (in order of preference):

1. **Auto-expire stale locks**: If `execution_locked_at` is older than the run's `finished_at` (or older than N minutes with no heartbeat activity), auto-clear on next access
2. **Force flag on release**: `POST /api/issues/{id}/release` with `{ "force": true }` — allowed for CEO agents or board users
3. **Dedicated admin endpoint**: `POST /api/issues/{id}/admin-release` with board-only auth

Option 1 is the best because it prevents the problem entirely without requiring manual intervention.

## Environment

- Paperclip v2026.318.0
- Embedded PostgreSQL
- `claude_local` adapter (Atlas, Opus 4.6)
- Self-hosted on Elestio

alberthanlee-arch:
Additional field report from AJ Lee Apparel / Paperclip self-hosted lane (2026-04-17):

We re-hit this on the AJL board while dispatching a live website buildout issue.

[...]

Issue #1033, opened 2026-03-16
Cancelled heartbeat runs should auto-release execution locks on issues


Phinease:
## Problem

When a heartbeat run is cancelled (e.g. by the board/user), the `executionRunId` and `checkoutRunId` fields on issues checked out by that run are **not cleaned up**. This causes:

1. **Stale execution locks** — subsequent heartbeats for the same agent must perform "lock adoption" to recover, adding startup latency and fragility.
2. **Checkout conflicts for other agents** — if a different agent tries to check out the same issue, they hit a 409 conflict against a dead run.
3. **Release side effects** — when agents try to work around the stale lock by calling `/release`, it unexpectedly clears the `assigneeAgentId`, causing the task to become unassigned.

## Observed Impact

In our deployment (5 agents, ~50 issues):
- **12 stale lock recoveries** (lock adoptions) in 24 hours, all caused by cancelled CEO heartbeat runs
- **2 wasted heartbeats** for Full-stack Engineer #2, who spent entire runs trying to fix checkout conflicts on AIS-44
- Agent explicitly documented the issue in their comment thread: "Two tasks have expired executionRunIDs (belonging to deprecated queued runs), causing the checkout to fail"

## Expected Behavior

When a heartbeat run transitions to `cancelled` (or `failed`/`errored`), the system should:

1. Find all issues where `executionRunId` or `checkoutRunId` matches the cancelled run ID
2. Clear those fields (set to `null`)
3. **Not** modify `assigneeAgentId` or `status` — the task should remain assigned and in its current status, ready for the next heartbeat

## Current Workaround

- Agents use lock adoption (checkout with stale run detection) to recover, but this adds ~5-10s per affected issue per heartbeat
- Reducing manual heartbeat cancellations helps but doesn't fix the root cause

Issue #1007, opened 2026-03-16
executionRunId not cleared on assignee change, blocks subsequent agent checkout


echokos:
## Description

When Agent A checks out an issue (setting `executionRunId` to their run), plans/works it, then PATCHes the issue to reassign to Agent B — the `checkoutRunId` is correctly cleared (line ~592-596 in issues.js service), but `executionRunId` is NOT cleared.

When Agent B then tries to check out the same issue, the checkout fails because `executionRunId` is already set to Agent A's (now-completed) run.

## Reproduction

1. Agent A (planner) is assigned a task in `todo`
2. Agent A checks out → status becomes `in_progress`, `executionRunId` = A's run ID
3. Agent A finishes planning, PATCHes: `{"status": "todo", "assigneeAgentId": "agent-B-id"}`
4. `checkoutRunId` correctly clears (line 596: assignee change clears it)
5. `executionRunId` remains set to Agent A's run
6. Agent B wakes up, tries to checkout → fails because `executionRunId` is non-null and doesn't match B's run

## Expected Behavior

When assignee changes via PATCH, both `checkoutRunId` AND `executionRunId` should clear — same as `checkoutRunId` already does.

## Workaround

Cron clearing stale locks:
```sql
UPDATE issues SET execution_run_id = NULL, checkout_run_id = NULL
FROM heartbeat_runs hr
WHERE issues.execution_run_id = hr.id
AND issues.status IN ('todo', 'backlog', 'blocked')
AND hr.status IN ('succeeded', 'failed', 'timed_out');
```

## Environment

- Paperclip v0.3.1
- Adapter: codex_local + claude_local
- Pipeline: Planner (claude_local) → Developer (codex_local)

ncmbch:
## End-to-end reproducer + workaround validation (Kotao, 2026-05-16)

Hitting this exact bug in production. Kotao internal tracker [KOT-60](https://github.com/) → [KOT-61](https://github.com/) for the forensic record. Adding evidence here rather than filing a separate issue — same root cause, same call site, same symptom.

### Reproducer (clock-aligned)

```
T0  Agent A (CTO) checks out issue.
    → executionRunId  = run-A
    → checkoutRunId   = run-A
    → assigneeAgentId = A
    → status          = in_progress

T1  Agent A's run terminates (succeeded).
    → checkoutRunId   = null   (cleared on run finalization)
    → executionRunId  = run-A  (still pinned)

[...]

PR #1723, opened 2026-03-24
fix: clear execution lock fields on status change, reassign, and release; add stale lock sweep

heathriel:
## Summary

This PR fixes four related execution lock bugs and adds a proactive stale lock sweep. Two fixes are **original contributions not covered by any existing open PR**; the other two synthesise approaches from open PRs that have not yet been merged.

---

### Fix 1 — PATCH status change leaves stale execution lock *(new — no existing PR)*

**Root cause:** `issues.ts` lines 903–904 clear `checkoutRunId` when status leaves `in_progress`, but `executionRunId`, `executionAgentNameKey`, and `executionLockedAt` are not touched.

**Impact:** Patching an issue to `backlog` or `todo` is the most natural operator path for manually unsticking a hung task. After the patch the issue *appears* available (`status=todo`, `checkoutRunId=null`) but every subsequent checkout attempt returns 409 because `executionLockCondition` still evaluates false. The issue is permanently stuck even though no live run owns it.

**Fix:** Clear all four lock fields whenever `issueData.status` transitions away from `in_progress`.

---

### Fix 2 — PATCH assignee change leaves stale execution lock *(extends #1007, complements PR #1328)*

**Root cause:** The assignee-change guard at lines 906–911 cleared `checkoutRunId` but not the execution lock trio.

**Impact:** Reassigning to a different agent left a stale `executionRunId` pointing to the previous agent's dead run. The new agent got an immediate 409 on its first checkout.

**Fix:** Clear all four lock fields on assignee change.

---

### Fix 3 — `release()` does not clear execution lock fields *(synthesises #725, PR #759, PR #1508)*

**Root cause:** `release()` set `{ status: "todo", assigneeAgentId: null, checkoutRunId: null }` but left `executionRunId`, `executionAgentNameKey`, and `executionLockedAt` unchanged. After release the issue re-entered a `todo` state with a live-looking execution lock.

**Impact:** `POST /api/issues/:id/release` — the only operator-accessible lock-clearing path — did not actually unblock the issue. Agents still hit 409 after a release call.

**Fix:** Add the three execution lock fields to the `release()` update set.

---

### Fix 4 — `release()` unconditionally clears `assigneeAgentId` *(synthesises #1678, PR #1716)*

**Root cause:** The release update included `assigneeAgentId: null`. Since `release()` is the only available mechanism for clearing a stale lock (no force-release endpoint exists), every stale-lock recovery silently orphaned the issue.

**Impact:** #1678 reports 9 issues losing their assignee from a single incident. Each required manual reassignment.

**Fix:** Removed `assigneeAgentId: null` from the `release()` update set. Callers that genuinely want to unassign should PATCH `assigneeAgentId: null` separately. This is a no-op change for the standard release path (where the agent is wrapping up its own work) and only affects the stale-lock-recovery case.

---

### Fix 5 — No startup or periodic sweep for stale execution locks *(new — no existing PR)*

**Root cause:** Even with Fixes 1–4 in place, execution locks written by older server versions or created through kill-9 / OOM scenarios accumulate indefinitely. `reapOrphanedRuns()` handles in-memory running processes but does not touch DB lock fields on issues.

**Impact:** Issues can be stuck across server restarts forever. The only recovery is direct DB access, which requires sysadmin access and awareness of the problem. In the incident that prompted this PR, 35 issues were stuck for 12+ hours after a single server overload event (a pathologically slow recursive `glob()` call pegging a CPU at 99.9% caused the scheduler to stop picking up work; the run processes were reaped but the issue lock fields were not cleared).

**Fix:** Added `sweepStaleExecutionLocks()` to `heartbeatService`. It queries for all issues where `executionRunId` points to a terminal (`succeeded`, `failed`, `timed_out`, `cancelled`, `process_lost`) or missing run and bulk-clears `executionRunId`, `executionAgentNameKey`, and `executionLockedAt`. Called:
- **Once at server startup**, chained after `reapOrphanedRuns()` and before `resumeQueuedRuns()`, so all stuck issues are unblocked before the queue drains.
- **Periodically** on every heartbeat scheduler tick, so issues stuck between restarts recover within one interval without requiring a restart.

---

## Files changed

| File | Lines | Change |
|------|-------|--------|
| `server/src/services/issues.ts` | ~903–911, ~1189–1202 | Fixes 1, 2, 3, 4 |
| `server/src/services/heartbeat.ts` | new function after `cancelBudgetScopeWork` | Fix 5: `sweepStaleExecutionLocks()` |
| `server/src/index.ts` | ~531–535, ~562–569 | Fix 5: wire sweep into startup and periodic tick |

## Related issues
\#725 \#1007 \#1033 \#1420 \#1678

## Testing
- Kill server mid-run → restart → previously stuck issues unblocked within seconds of startup
- PATCH `in_progress` issue status to `backlog` → subsequent checkout succeeds (no 409)
- Reassign `in_progress` issue to a different agent → new agent can check out immediately
- `POST /api/issues/:id/release` on a stuck issue → `assigneeAgentId` preserved on the returned issue

### DIFF
--- a/server/src/index.ts
+++ b/server/src/index.ts
@@ -530,8 +530,12 @@ export async function startServer(): Promise<StartedServer> {
   
     // Reap orphaned running runs at startup while in-memory execution state is empty,
     // then resume any persisted queued runs that were waiting on the previous process.
+    // Also sweep any issues whose executionRunId points to a dead run — these can
+    // accumulate when the server is killed mid-run and would otherwise block all future
+    // checkouts on those issues indefinitely.
     void heartbeat
       .reapOrphanedRuns()
+      .then(() => heartbeat.sweepStaleExecutionLocks())
       .then(() => heartbeat.resumeQueuedRuns())
       .catch((err) => {
         logger.error({ err }, "startup heartbeat recovery failed");
@@ -559,10 +563,12 @@ export async function startServer(): Promise<StartedServer> {
           logger.error({ err }, "routine scheduler tick failed");
         });
   
-      // Periodically reap orphaned runs (5-min staleness threshold) and make sure
-      // persisted queued work is still being driven forward.
+      // Periodically reap orphaned runs (5-min staleness threshold), sweep any stale
+      // execution lock fields left by dead runs, and make sure persisted queued work is
+      // still being driven forward.
       void heartbeat
         .reapOrphanedRuns({ staleThresholdMs: 5 * 60 * 1000 })
+        .then(() => heartbeat.sweepStaleExecutionLocks())
         .then(() => heartbeat.resumeQueuedRuns())
         .catch((err) => {
           logger.error({ err }, "periodic heartbeat recovery failed");
--- a/server/src/services/heartbeat.ts
+++ b/server/src/services/heartbeat.ts
@@ -3844,6 +3844,54 @@ export function heartbeatService(db: Db) {
 
     cancelBudgetScopeWork,
 
+    /**
+     * Sweep all issues whose executionRunId points to a terminal or missing run and clear
+     * the stale execution lock fields.  Should be called once on server startup and
+     * periodically (e.g. every 5 minutes) via the scheduler so that issues stuck by a
+     * crashed run are automatically unblocked without requiring manual intervention.
+     *
+     * This complements reapOrphanedRuns() (which handles in-memory running processes) by
+     * catching the persistence-level case where the DB row was never cleaned up — e.g.
+     * when the server was killed mid-run, when a run was cancelled while the issue lock
+     * was already set, or when the release() call was skipped due to a code path bug.
+     */
+    sweepStaleExecutionLocks: async () => {
+      const TERMINAL_STATUSES = ["succeeded", "failed", "timed_out", "cancelled", "process_lost"];
+
+      // Find issues with an executionRunId that is either missing or terminal.
+      const staleIssues = await db.execute(sql`
+        SELECT i.id, i.identifier, i.execution_run_id
+        FROM issues i
+        WHERE i.execution_run_id IS NOT NULL
+          AND NOT EXISTS (
+            SELECT 1 FROM heartbeat_runs hr
+            WHERE hr.id = i.execution_run_id
+              AND hr.status NOT IN (${sql.join(TERMINAL_STATUSES.map((s) => sql`${s}`), sql`, `)})
+          )
+      `);
+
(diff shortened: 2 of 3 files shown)


doganalpaslan (changes requested):
## TypeScript fix needed

CI is failing with `TS2339: Property 'rows' does not exist on type 'RowList<Record<string, unknown>[]>'` at `server/src/services/heartbeat.ts` lines 3873 and 3875.

**Root cause:** Drizzle's `db.execute(sql`...`)` returns a `RowList` directly (which is array-like), not an object with a `.rows` property.

**Fix** (2-line change):

```diff
- if (staleIssues.rows.length === 0) return { cleared: 0 };
+ if (staleIssues.length === 0) return { cleared: 0 };

- const staleIds = staleIssues.rows.map((r: Record<string, unknown>) => r.id as string);
+ const staleIds = staleIssues.map((r) => r.id as string);
```

Everything else in the PR looks solid — the execution lock clearing on status change, reassign, and release is correct, and the stale sweep logic is sound. Just needs this TS fix to pass CI.

devinfoley:
Closing as already-fixed upstream — thanks for the thorough analysis in the original write-up.

Since this PR was opened, master has absorbed each of the five fixes:

- **Fixes 1, 2, 3** (clear execution-lock columns when leaving `in_progress`, on reassignment, and on `release()`) — landed in #4258, which centralizes terminal-run cleanup and applies it before every ownership-gated write.
- **Fix 4** (preserve assignee during stale-lock recovery) — addressed differently. Master keeps the original `release()` contract (the assignee-clear still happens for callers wrapping up their own work) and #4258 introduced `adminForceRelease({ clearAssignee })` so stale-lock recovery can preserve the assignee explicitly, without coupling that behavior change to every release call.
- **Fix 5** (periodic backstop sweep of stale execution locks) — landed in #6008 as `sweepStaleIssueLocks` (`server/src/services/recovery/service.ts`). The upstream version covers both `checkoutRunId` and `executionRunId`, uses a per-row TOCTOU-safe `UPDATE … WHERE checkoutRunId = ? AND executionRunId = ?` predicate (so a concurrent re-checkout between the SELECT and the UPDATE is not stomped), emits `issue.stale_lock_cleared` activity log entries, and runs both at startup and on every heartbeat scheduler tick.

Closing without merge to avoid double-applying the same logic. The incident class this PR was responding to (issues stuck on dead `executionRunId` after a server crash) is now self-healing on master.

PR #1761, opened 2026-03-25
fix(ui): show error feedback when mutations fail silently

bluzername:
## Problem

Many mutation hooks in AgentDetail page and Routines page was not having `onError` callback. When user perform action like updating budget, changing icon, cancelling a run or creating API key - if it fail, nothing happen. No error message, no toast, nothing. User just sit there thinking maybe it worked or maybe it didn't.

## What I changed

Added `onError` handlers to all mutations that was missing them:

**AgentDetail.tsx:**
- `budgetMutation` - now show inline error when budget update fail
- `updateIcon` - now show inline error when icon change fail
- `cancelRun` - toast notification on failure
- `resumeRun` - toast notification on failure
- `retryRun` - toast notification on failure
- `clearSessionsForTouchedIssues` - toast notification on failure
- `runClaudeLogin` - toast notification on failure
- `createKey` - toast notification on failure
- `revokeKey` - toast notification on failure

**Routines.tsx:**
- `createRoutine` - toast notification on failure

## How it works

For the main agent detail section, I use `setActionError` which is the existing pattern already used by `agentAction`, `resetTaskSession` and `updatePermissions` mutations in same component.

For the RunDetail and KeysTab sub-components, I added `useToast()` hook and show toast errors - this match what Routines page already doing for `updateRoutineStatus` and `runRoutine`.

## How to test

2. Try to trigger a mutation when server is down or with network disconnected
3. Should see error toast or inline error message instead of silent failure
4. Same for Routines page - try create routine when network is off

No logic change, only adding error notification. 36 lines added across 2 files.

### DIFF
--- a/ui/src/pages/AgentDetail.tsx
+++ b/ui/src/pages/AgentDetail.tsx
@@ -710,6 +710,9 @@ export function AgentDetail() {
       queryClient.invalidateQueries({ queryKey: queryKeys.agents.list(resolvedCompanyId) });
       queryClient.invalidateQueries({ queryKey: queryKeys.dashboard(resolvedCompanyId) });
     },
+    onError: (err) => {
+      setActionError(err instanceof Error ? err.message : "Failed to update budget");
+    },
   });
 
   const updateIcon = useMutation({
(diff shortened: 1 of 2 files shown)


bluzername:

PR #1773, opened 2026-03-25
fix(ui): show full date on hover for relative time labels

bluzername:
## Problem

Everywhere in the UI we show relative time like "2h ago" or "5d ago" but there was no way to see the actual date and time. If I want to know exactly when something happened, I have to go dig in the database or look at API response. This is very annoying specially when you need to correlate events across different pages or compare timestamps.

## What I did

Added small helper function `absoluteDate()` in `timeAgo.ts` that format a date using `toLocaleString()` (so it respect user's locale and timezone settings). Then added `title` attribute to all the `<span>` elements that render `timeAgo()` output.

Now when you hover over any relative time label, browser show a tooltip with full date like "3/26/2026, 2:45:30 PM" (or whatever format your locale use).

## Where it applied

- **ActivityRow** - event timestamps in activity feed
- **ApprovalCard** - approval creation time
- **IssueProperties** - "Updated" field in issue sidebar
- **Dashboard** - issue update times in dashboard list
- **Inbox** - run times, approval update times, join request times
- **RoutineDetail** - run triggered times and event log times

Note: some places use `timeAgo()` as a string prop (like `mobileMeta`) where we cannot add HTML attributes. Those stay as before - only the JSX rendered timestamps got the tooltip.

## How to test

1. Go to any page that show relative times (Dashboard, Inbox, Activity, etc.)
2. Hover mouse over any "2h ago" or "3d ago" label
3. Should see browser tooltip with full date and time

7 files, net +5 lines.

### DIFF
--- a/ui/src/components/ActivityRow.tsx
+++ b/ui/src/components/ActivityRow.tsx
@@ -1,6 +1,6 @@
 import { Link } from "@/lib/router";
 import { Identity } from "./Identity";
-import { timeAgo } from "../lib/timeAgo";
+import { timeAgo, absoluteDate } from "../lib/timeAgo";
 import { cn } from "../lib/utils";
 import { deriveProjectUrlKey, type ActivityEvent, type Agent } from "@paperclipai/shared";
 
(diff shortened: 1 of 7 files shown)


bluzername:

PR #1779, opened 2026-03-25
fix(server): bootstrap minimal agent home

yoyooyooo:
## Thinking Path
- Paperclip orchestrates zero-human companies through agent heartbeats.
- New local agents execute onboarding instructions on their first heartbeat.
- Those instructions already depend on a minimal personal workspace contract: `MEMORY.md`, `life/`, and `memory/YYYY-MM-DD.md`.
- But the runtime only created the top-level `AGENT_HOME` directory, so fresh agents could wake without the files their instructions told them to read.
- The CEO onboarding asset also implied `HEARTBEAT.md`, `SOUL.md`, and `TOOLS.md` lived under `$AGENT_HOME`, even though those files come from the managed instructions bundle.
- So this PR bootstraps the minimal runtime home skeleton before heartbeat work starts, aligns the onboarding copy with the real bundle/runtime split, and adds regression coverage plus docs for that contract.

## Summary
- bootstrap a minimal `AGENT_HOME` PARA skeleton before heartbeat work begins so newly created local agents have the files and directories their onboarding instructions already require
- update the CEO onboarding asset to read `HEARTBEAT.md`, `SOUL.md`, and `TOOLS.md` from the managed instructions bundle instead of implying they live directly under `$AGENT_HOME`
- add regression tests and docs that distinguish the managed instructions bundle from the runtime `AGENT_HOME` contract
- address review follow-up by guarding `writeAgentHomeFileIfMissing(...)` when the target already exists as a directory and by asserting real directory creation in the bootstrap test

## Problem
New local agents currently get an `AGENT_HOME` directory, but the runtime does not create the minimal PARA memory layout that the CEO onboarding instructions immediately depend on. In practice, this leaves fresh agents missing `MEMORY.md`, `life/`, and `memory/YYYY-MM-DD.md` until someone repairs the workspace manually.

There is also a documentation mismatch in the CEO onboarding asset: the essential support files are delivered in the managed instructions bundle, while the text implies they should be read from the runtime home directory.

## Test Plan
- [x] `pnpm test:run server/src/__tests__/heartbeat-workspace-session.test.ts server/src/__tests__/agent-skills-routes.test.ts`
- [x] `pnpm --filter @paperclipai/server typecheck`

### DIFF
--- a/server/src/__tests__/heartbeat-workspace-session.test.ts
+++ b/server/src/__tests__/heartbeat-workspace-session.test.ts
@@ -80,10 +80,10 @@ describe("ensureAgentHomeBootstrap", () => {
     expect(home).toBe(resolveDefaultAgentWorkspaceDir(agentId));
     await expect(fs.readFile(path.join(home, "MEMORY.md"), "utf8")).resolves.toContain("# Memory");
     await expect(fs.readFile(path.join(home, "life", "index.md"), "utf8")).resolves.toContain("# Life");
-    await expect(fs.stat(path.join(home, "life", "projects"))).resolves.toMatchObject({ isDirectory: expect.any(Function) });
-    await expect(fs.stat(path.join(home, "life", "areas"))).resolves.toMatchObject({ isDirectory: expect.any(Function) });
-    await expect(fs.stat(path.join(home, "life", "resources"))).resolves.toMatchObject({ isDirectory: expect.any(Function) });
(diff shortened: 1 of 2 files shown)


server/src/onboarding-assets/ceo/AGENTS.md
```diff
@@ -17,8 +17,8 @@ Invoke it whenever you need to remember, retrieve, or organize anything.
 
 ## References
 
-These files are essential. Read them.
+These files are essential. Read them from this instructions bundle.
```
musay4632, line 20: Approved

PR #1788, opened 2026-03-26
fix(server): restore issue comment cursor pagination

yoyooyooo:
## Thinking Path

> - Paperclip orchestrates AI agents through issue threads, comments, and heartbeat-driven execution.
> - Heartbeats and agent follow-ups rely on incremental comment reads so they can load only the delta after the last seen comment instead of replaying an entire thread.
> - The `GET /api/issues/:id/comments?after=...&order=asc` path is the API contract that supports that incremental flow.
> - Valid `after` cursors currently fail with a 500 because `issueService.listComments(...)` interpolates `anchor.createdAt` into a raw `sql``...`` comparison, and the postgres driver rejects the bound `Date`.
> - This pull request replaces that raw cursor comparison with typed Drizzle conditions and adds a regression test that exercises ascending comment pagination after a cursor.
> - The benefit is that comment-delta reads work again for valid cursors, which restores a core heartbeat/comment-follow-up path without changing the default full-thread listing behavior.

## What Changed

- replaced the raw SQL cursor comparison in `issueService.listComments(...)` with `gt`/`lt`-based Drizzle conditions while preserving the `createdAt` then `id` tie-break ordering
- added a regression test in `server/src/__tests__/issues-service.test.ts` that verifies comments after a cursor are returned in ascending order

## Verification

- `pnpm test:run server/src/__tests__/issues-service.test.ts`

## Risks

- Low risk. The change is scoped to comment listing when `afterCommentId` is provided.
- Full comment listing without a cursor is unchanged.
- Invalid non-UUID cursor strings still follow the existing route behavior and are not changed in this PR.

## Checklist

### DIFF
--- a/server/src/__tests__/issues-service.test.ts
+++ b/server/src/__tests__/issues-service.test.ts
@@ -281,6 +281,36 @@ describe("issueService.list participantAgentId", () => {
 
     expect(result.map((issue) => issue.id)).toEqual([matchedIssueId]);
   });
+});
+
+describe("issueService.listComments", () => {
+  let db!: ReturnType<typeof createDb>;
+  let svc!: ReturnType<typeof issueService>;
+  let instance: EmbeddedPostgresInstance | null = null;
+  let dataDir = "";
+
+  beforeAll(async () => {
+    const started = await startTempDatabase();
+    db = createDb(started.connectionString);
+    svc = issueService(db);
+    instance = started.instance;
+    dataDir = started.dataDir;
+  }, 20_000);
+
+  afterEach(async () => {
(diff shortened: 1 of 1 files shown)


yoyooyooo:
Addressed the current Greptile follow-up in commit `d2c3cf7`:

- moved the `listComments` cursor regression into its own `describe("issueService.listComments", ...)` block
- kept production behavior unchanged; this is only test-suite organization

Fresh local verification before push:
- `pnpm test:run server/src/__tests__/issues-service.test.ts`

GitHub Actions is re-running on the new commit now.

PR #1813, opened 2026-03-26
refactor(skill): reduce paperclip SKILL.md context size by 60% with progressive disclosure

tstirrat:
## Thinking Path

- Paperclip orchestrates AI agents for zero-human companies
- Every agent that has the Paperclip skill gets `SKILL.md` injected into their context on **every heartbeat run**
- The skill was 365 lines / 23KB — this bloats context windows across every agent, every run, at significant token cost
- Most of the content in the skill is only needed in specific, infrequent situations (project setup, OpenClaw invites, import/export, etc.)
- So we can apply **progressive disclosure**: keep only the heartbeat procedure and critical rules inline, and move situational content to reference files that agents read on demand
- But we need to be careful not to regress agent reliability — some content was previously hardened into the body after field failures
- This PR reduces the main skill body by 60% while preserving all critical inline content and routing agents to `references/` files via conditional "When X, read Y" pointers

## What Changed

**`skills/paperclip/SKILL.md`** — reduced from 365 lines (23KB) to 145 lines (11KB):

- **Kept verbatim:** frontmatter, heartbeat concept, authentication, Steps 1–9, Critical Rules
- **Compressed inline:** Comment Style (retained ticket-link rule, company-prefix rule, and full URL pattern list — these were hardened across multiple commits due to field reliability issues, so kept close)
- **Compressed inline:** Planning (retained key rules; moved API flow details to reference)
- **Removed from body:** Key Endpoints table (fully covered by `api-reference.md`), Project Setup, OpenClaw Invite, Company Skills inline section, Instructions Path, Import/Export, Searching Issues, Self-Test Playbook
- **Added:** References table — `| When... | Read |` — conditional pointers so agents load detail only when the situation applies

**`skills/paperclip/references/workflows.md`** (new file, 198 lines):

- Receives all moved content with H2 anchors for deep-linking
- Sections: Comment Style Details, Planning, Project Setup, OpenClaw Invite, Instructions Path, Import/Export, Searching Issues, Self Test

## Why This Matters

Every agent with the Paperclip skill pays the full 23KB context cost on every heartbeat, even for simple runs (pick up task → do work → comment → exit). The majority of that content is never needed in a typical run. This change reduces the per-heartbeat context overhead significantly, which compounds across all agents and all runs.

A secondary benefit: a smaller skill means **shorter, less cluttered agent comments on issues**. When the skill is large, agents tend to over-explain and reference content they've just been primed with. A leaner skill produces more focused, actionable comments — which improves the signal-to-noise ratio in issue threads that humans and other agents read.

## How It Was Made

The optimization was done using the [`skill-creator`](https://github.com/paperclipai/paperclip/tree/master/skills/skill-creator) skill, which is designed for exactly this kind of structured skill reduction — analysing what belongs inline vs. behind progressive disclosure, preserving reliability-critical content, and wiring up conditional reference pointers.

## Git History Check

Before compressing, checked `git log --follow -p skills/paperclip/SKILL.md` to identify content deliberately hardened into the body:

- **Comment Style** was strengthened across 3 separate commits after agents posted malformed links in production — kept the full URL pattern list inline rather than behind a reference pointer
- **Critical Rules** and **Steps 1–9** were left entirely untouched
- Content with no history of reliability issues (Project Setup, OpenClaw, etc.) was moved to references

## Verification — Local Dev Testing

Tested against a local Paperclip instance (COR company, CEO agent, claude-sonnet-4-6).

| Test | Result |
|------|--------|
| Comment style: ticket links + prefixed URLs | ✅ Agent posted `[COR-1](/COR/issues/COR-1)` — correct markdown link, correct prefix |
| Checkout before working | ✅ `POST /api/issues/{id}/checkout` with `X-Paperclip-Run-Id` confirmed in run log |
| Blocked status reporting | ✅ Status set to `blocked`, clear comment explaining blocker and who needs to act |
| Blocked-task dedup | ✅ Second heartbeat correctly skipped — own prior comment, no new context |
| Clean exit when no work | ✅ Exited cleanly with no API mutations |

All tests behaviorally neutral — no regressions observed.

## Risks

- Agents that encounter a situation requiring a reference file (e.g. project setup) must follow the pointer and read `workflows.md` — if they skip the pointer, they will have less guidance. This is mitigated by the "When X, read Y" framing being explicit and conditional.
- The Comment Style rules were the highest-risk area given their history; they were retained inline to avoid any regression there.

### DIFF
--- a/skills/paperclip/SKILL.md
+++ b/skills/paperclip/SKILL.md
@@ -95,7 +95,7 @@ If `currentParticipant` does not match you, do not try to advance the stage — 
 **Step 8 — Update status and communicate.** Always include the run ID header.
 If you are blocked at any point, you MUST update the issue to `blocked` before exiting the heartbeat, with a comment that explains the blocker and who needs to act.
 
-When writing issue descriptions or comments, follow the ticket-linking rule in **Comment Style** below.
+When writing issue descriptions or comments, follow the **Comment Style** rules below.
 
 ```json
 PATCH /api/issues/{issueId}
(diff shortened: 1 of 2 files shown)


scripts/smoke/paperclip-skill-smoke.sh
tstirrat, line 1: I don't know if this script is wanted, let me know if you want it removed. It takes quite a while to run but it does verify the basics of the skill.md. I tried to use the `promptfoo eval` stuff, but Claude told me they were unrelated to this skill.

aaaaron (approved):
The only minor nits from my claude :robot:

 1. Planning section was compressed too aggressively. The inline version now just says "use issue documents with key plan" and points to the reference file. But the rule "do not mark planning issues as done — re-assign to the requester" is critical enough that agents frequently get wrong. It's good that this stayed inline, but the document deep-linking pattern (#document-plan) was removed from the inline section. Agents that don't read the reference will lose that — and document deep-links are used on nearly every planning run, not just "specific situations." Consider keeping the deep-link pattern inline.
2. The Comment Style section now says "For a comment example, read: skills/paperclip/references/workflows.md" — but the example is a ~10 line markdown block. That's a lot of indirection for something agents reference frequently.  The example was useful inline as a concrete template. Consider keeping it.

I don't think these issues are blockers though, approving.

tstirrat:
Is there anything else I need to do here? @aaaaron

tstirrat:
Closing as out of date

PR #1864, opened 2026-03-26
feat(heartbeat): reconcile assigned issues on startup

### LINKED ISSUE #1845: No crash-recovery wakeup for agents with assigned issues after server restart
## Summary

When the Paperclip server restarts after a crash (OOM, SIGABRT, etc.), agents with assigned `in_progress` issues do not receive automatic wakeup requests. They remain idle until the next heartbeat timer fires, and even then the timer-source runs have a separate workspace resolution bug (#1844) that prevents productive work.

## Expected Behavior

On startup, after the database and heartbeat service are initialized, the server should run a reconciliation pass that:

1. Queries all agents with assigned `in_progress` or `todo` issues
2. Issues an `assignment`-source wakeup for each such agent
3. Logs the reconciliation results

This would ensure agents resume their work after any server restart, whether clean or crash-induced.

## Current Behavior

- Server restarts and heartbeat timers resume on their configured interval (e.g., 1200s)
- No startup reconciliation occurs
- Agents sit idle for up to one full heartbeat interval
- When the timer does fire, it hits #1844 (fallback workspace) and the agent still can't work

## Impact

Combined with the ~60-minute OOM crash cycle (#1825) and workspace resolution bug (#1844), this creates a pattern where agents can be effectively dead for hours: crash → restart → idle wait → timer fires → wrong workspace → "no work" → repeat.

## Environment

- paperclipai v2026.318.0
- systemd service with `Restart=on-failure`
- 7 active agents, heartbeat interval 1200s

levineam:
## Thinking Path

> - Agents pick up issues and work on them; their progress is tracked via heartbeat runs
> - The server coordinates agent wakeups through a heartbeat service that polls and dispatches work
> - When the server crashes or restarts, in-flight heartbeat state is lost
> - Existing startup recovery handles orphaned runs (`reapOrphanedRuns`) and queued runs (`resumeQueuedRuns`)
> - But agents with assigned `in_progress` or `todo` issues were left idle — no wakeup was triggered until the next timer tick
> - This created a dead zone that could last minutes, compounding agent stuck-state issues
> - This PR adds a third startup pass: `reconcileAssignedIssueWakeups`, which enqueues one `assignment`-source wakeup per affected agent immediately on startup
> - Candidate selection is a pure helper (`pickAssignedIssueWakeupCandidates`) so it can be unit-tested independently
> - Per-candidate error handling ensures one paused/budget-blocked agent can't abort recovery for all remaining agents

## What Changed

- Added `pickAssignedIssueWakeupCandidates` — pure helper that selects one issue per agent, preferring `in_progress` over `todo`, oldest first (well-tested in isolation)
- Added `reconcileAssignedIssueWakeups` to `heartbeatService` — queries assigned active issues, picks one candidate per agent, enqueues startup wakeups with per-candidate try/catch so one blocked agent doesn't abort the rest
- Wired into `createApp` as a fire-and-forget call after `scheduler.start()` with a top-level error log

## Why It Matters

After a crash/restart, agents with assigned work were silently idle until the next heartbeat timer. In production this means minutes of dead time per affected agent. This closes the gap by immediately re-queuing affected agents at startup.

## Verification

```bash
# Run the focused tests
pnpm vitest run server/src/__tests__/heartbeat-startup-reconciliation.test.ts

# Typecheck
```

## Risks

- `enqueueWakeup` may enqueue a wakeup for an agent that is already running (if it survived the restart). The heartbeat service handles concurrent runs gracefully — the duplicate wakeup will be a no-op or cancel quickly.
- The reconciliation creates a separate ephemeral `heartbeatService` instance in `createApp` (noted by Greptile). This is functionally correct; refactoring the wiring is a follow-up if desired.

Fixes #1845

### DIFF
--- a/server/src/__tests__/heartbeat-maintenance.test.ts
+++ b/server/src/__tests__/heartbeat-maintenance.test.ts
@@ -1,0 +1,107 @@
+import { describe, expect, it, vi, beforeEach } from "vitest";
+
+const logger = vi.hoisted(() => ({
+  info: vi.fn(),
+  error: vi.fn(),
+}));
+
+vi.mock("../middleware/logger.js", () => ({ logger }));
+
+import { runHeartbeatSchedulerCycle, startHeartbeatScheduler } from "../heartbeat-maintenance.js";
+
+describe("runHeartbeatSchedulerCycle", () => {
+  beforeEach(() => {
+    vi.clearAllMocks();
+  });
+
+  it("reconciles assigned issue wakeups during the periodic scheduler cycle", async () => {
+    const heartbeat = {
+      tickTimers: vi.fn().mockResolvedValue({ checked: 0, enqueued: 0, skipped: 0 }),
+      reapOrphanedRuns: vi.fn().mockResolvedValue({ reaped: 0 }),
+      resumeQueuedRuns: vi.fn().mockResolvedValue([]),
+      reconcileAssignedIssueWakeups: vi.fn().mockResolvedValue({ checked: 2, enqueued: 1, skipped: 1 }),
+    };
+    const routines = {
+      tickScheduledTriggers: vi.fn().mockResolvedValue({ checked: 0, triggered: 0, skipped: 0 }),
+    };
+    const now = new Date("2026-04-12T23:30:00.000Z");
+
+    await runHeartbeatSchedulerCycle({ heartbeat, routines, now });
+
+    expect(heartbeat.tickTimers).toHaveBeenCalledWith(now);
+    expect(routines.tickScheduledTriggers).toHaveBeenCalledWith(now);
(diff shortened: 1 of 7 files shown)


levineam (commented):
Both Greptile comments are addressed:

**Per-candidate error handling** (reviewed at commit `5bddce88`): the current tip of this branch already wraps each `enqueueWakeup` call in a per-candidate `try/catch` that logs and increments `skipped` on failure, keeping the loop alive for all remaining candidates. Greptile reviewed an older version before that fix was applied.

**Ephemeral `heartbeatService` instance** in `app.ts:81`: acknowledged as a non-blocking suggestion. The pattern is intentional — `reconcileAssignedIssueWakeups` is called once at startup as a fire-and-forget side effect; passing the service through `opts` would add coupling for a one-time call. The instance is GC-eligible immediately after the call completes, which is the expected behavior here.

brandonburr:

**Rationale:** Heartbeat: assigned issues are now reconciled on startup.

PR #1884, opened 2026-03-27
[RUS-418] issueService: support projectId null filtering in list()

benedictchee:
## Thinking Path

- Paperclip agents use `inbox-lite` to fetch their assigned issues during heartbeats
- A planned fix (RUS-414) adds project-scoping to this endpoint using `currentIssueId`
- To scope issues by project, `issueService.list()` needs to filter by `projectId: null` for issues with no project
- But `IssueFilters.projectId` was typed as `string | undefined`, and `null` is falsy so the filter was silently skipped
- This PR fixes the type and filter logic so null projectId is handled correctly

## What Changed

- `IssueFilters.projectId` type: `string | undefined` → `string | null | undefined`
- Filter logic at line 649 now uses `"projectId" in filters` check to distinguish three states:
  - Key absent/undefined → no filter applied
  - `null` → `isNull(issues.projectId)` condition
  - Non-null string → `eq(issues.projectId, value)` condition
- `isNull` was already imported from `drizzle-orm` — no new imports needed

## Test Results

| Test case | Input | Expected | Actual | Status |
|-----------|-------|----------|--------|--------|
| Project filter | `list(cid, { projectId: "uuid" })` | issues with that projectId | issues with that projectId | ✅ PASS |
| Null filter | `list(cid, { projectId: null })` | `IS NULL` condition applied | `IS NULL` condition applied | ✅ PASS |
| No filter | `list(cid, {})` | all issues | all issues | ✅ PASS |
| issues-service tests | `npx vitest run src/__tests__/issues-service.test.ts` | 3/3 pass | 3/3 pass | ✅ PASS |

Closes RUS-418

### DIFF
--- a/server/src/__tests__/issues-service.test.ts
+++ b/server/src/__tests__/issues-service.test.ts
@@ -8,6 +8,7 @@ import {
   issueComments,
   issueInboxArchives,
   issues,
+  projects,
 } from "@paperclipai/db";
 import {
   getEmbeddedPostgresTestSupport,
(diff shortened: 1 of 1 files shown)


benedictchee:
## Technical PM Review: ✅ APPROVED

Implementation matches the DST specification exactly.

**Verified:**
- `IssueFilters.projectId` type updated to `string | null`
- Three-state filter logic using `"projectId" in filters`: absent → no filter, `null` → `isNull()`, string → `eq()`
- 3 integration tests with embedded Postgres covering all filter states
- No security concerns, minimal change surface

Merging now as this is a dependency for #1885 (RUS-419).

PR #1893, opened 2026-03-27
feat(ui): copy button for run error messages

bluzername:
## Thinking Path

> - Agents run tasks and sometimes those tasks fail with error messages
> - Human users need to inspect these errors in the run detail view to understand what went wrong
> - The error text shows as small red text inside a mixed UI container with other elements
> - But there was no way to copy the error message without carefully selecting text with mouse, which is annoying when error is long
> - Users often need to paste error messages into issues, send to teammate, search in docs, or compare across runs
> - This PR adds a CopyText button that appears on hover next to the error message, copying both error text and error code in one click
> - The benefit is faster debugging workflow - no more fiddly text selection when you just need the error string

## Problem

When an agent run fail with error, the error message show in the run detail view as plain red text. But there was no way to copy it easily. The text is inside a small span element mixed with other UI elements, so selecting it with mouse is tricky specially when the error message is long.

I often need to copy error messages to:
- Paste into a new issue for tracking
- Send to teammate on Slack for help
- Search in documentation or Google
- Compare with errors from other runs

Every time I had to carefully select just the error text with cursor which is annoying.

## What I changed

Wrapped the error text and error code in a flex container with a CopyText button. The copy icon:
- Hidden by default (opacity-0), only appear on hover (group-hover/err pattern)
- Copy both the error message AND error code in one click
- Show "Copied!" tooltip popup using the existing CopyText component
- Uses the Copy icon from lucide-react (already imported in the file)

## Known minor issue

The CopyText component should ideally be wrapped in a `<span className="shrink-0">` to prevent the icon from shrinking on very long error messages in narrow viewports. Can fix in follow-up.

## How to test

1. Go to any agent > Runs tab > click on a failed run
2. Hover over the red error message text
3. Small copy icon should appear on the right side
4. Click it - error text copied to clipboard with "Copied!" tooltip

1 file, 8 lines changed.

### DIFF
--- a/ui/src/pages/AgentDetail.tsx
+++ b/ui/src/pages/AgentDetail.tsx
@@ -3085,9 +3085,14 @@ function RunDetail({ run: initialRun, agentRouteId, adapterType }: { run: Heartb
               </div>
             )}
             {run.error && (
-              <div className="text-xs">
-                <span className="text-red-600 dark:text-red-400">{run.error}</span>
-                {run.errorCode && <span className="text-muted-foreground ml-1">({run.errorCode})</span>}
+              <div className="group/err flex items-start gap-1.5 text-xs">
+                <div className="min-w-0">
+                  <span className="text-red-600 dark:text-red-400">{run.error}</span>
+                  {run.errorCode && <span className="text-muted-foreground ml-1">({run.errorCode})</span>}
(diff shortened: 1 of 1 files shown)


bluzername:

PR #1909, opened 2026-03-27
Fix DB backup crash on large databases by streaming writes

### LINKED ISSUE #1843: DB backup crashes server with RangeError when database exceeds ~250MB
## Summary

`runDatabaseBackup()` crashes the server with `RangeError: Invalid string length` when the embedded PostgreSQL database exceeds ~250MB. This triggers a SIGABRT that takes down all running agent sessions.

## Root Cause

The backup implementation in `index.js` (line ~437) calls `runDatabaseBackup()` which appears to buffer the entire pg_dump output into a single JavaScript string before writing to disk. V8 has a maximum string length of ~512MB (depending on encoding), but the heap pressure from constructing a string this large triggers OOM well before that limit.

The automatic backup runs on a configurable interval (default 60 minutes via `databaseBackupIntervalMinutes`). As the database grows (primarily from `heartbeat_runs.result_json` accumulation), each backup attempt becomes more likely to crash the server.

## Evidence

- Database size: 293MB on disk, producing 451MB SQL dump files
- Backup directory accumulated 112 dump files totaling 21GB before manual cleanup
- Server log shows `RangeError: Invalid string length` from the backup routine immediately before SIGABRT
- `retentionDays` pruning works but doesn't prevent the OOM during dump generation

## Environment

- paperclipai v2026.318.0
- Node.js v24.13.1
- Database grew from ~100MB to 293MB over 8 days with ~20 agents

## Suggested Fix

Stream the pg_dump output directly to the destination file instead of buffering in a JS string. Use `child_process.spawn` with stdout piped to a `fs.createWriteStream`, or use `pg_dump --file=<path>` to let PostgreSQL write directly.

## Workaround

Set `PAPERCLIP_DB_BACKUP_ENABLED=false` in the environment.

moktamd:
**Thinking path:** Server crashes during scheduled backups → RangeError: Invalid string length in backup-lib.ts → the entire SQL dump is buffered in a lines[] array and joined into one string → for databases >~250MB the joined string exceeds V8 limits → fix by streaming writes directly to disk instead of buffering.

Fixes #1843

Backup accumulated all SQL in memory before writing to disk. For databases exceeding ~250MB, `lines.join("\n")` hit V8's string length limit, crashing the server and killing all agent sessions.

- `runDatabaseBackup`: writes SQL incrementally via `createWriteStream` with backpressure handling instead of buffering in a string array
- `runDatabaseRestore`: reads the dump file line-by-line via `createReadStream`/`readline` instead of `readFile` into one string

Memory usage is now bounded regardless of database size.

### DIFF
--- a/packages/db/src/backup-lib.ts
+++ b/packages/db/src/backup-lib.ts
@@ -1,6 +1,6 @@
-import { existsSync, mkdirSync, readdirSync, statSync, unlinkSync } from "node:fs";
-import { readFile, writeFile } from "node:fs/promises";
+import { createReadStream, createWriteStream, existsSync, mkdirSync, readdirSync, statSync, unlinkSync } from "node:fs";
 import { basename, resolve } from "node:path";
+import { createInterface } from "node:readline";
 import postgres from "postgres";
 
 export type RunDatabaseBackupOptions = {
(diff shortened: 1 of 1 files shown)


smaugho (approved):
Clean implementation. The streaming writer with backpressure handling and the line-based restore are both solid improvements. This fixes the RangeError we've been seeing on large databases.

PR #872, opened 2026-03-14
fix: fall back to directory junctions on Windows (EPERM)

octasoft-ltd:
## Thinking Path

> - Paperclip orchestrates AI agents for zero-human companies
> - Agents need Paperclip skills installed via symlinks into their working directories
> - On Windows, creating symlinks requires Developer Mode enabled or administrator privileges
> - Without either, \`fs.symlink()\` throws \`EPERM\`, preventing skill installation and blocking agent startup
> - Directory junctions are a Windows-specific alternative that work without elevated privileges and behave identically for skill directory linking
> - For file symlinks (e.g. \`auth.json\` in codex-home), hard links are the equivalent fallback — they work on NTFS without privileges and keep contents in sync
> - This PR adds \`symlinkOrJunction()\` for directories and \`symlinkOrHardLink()\` for files, covering all symlink call sites
> - The benefit is that agents can be set up on Windows without requiring Developer Mode or admin rights, lowering the barrier to entry

## Summary

Adds two EPERM fallback helpers for Windows environments without Developer Mode:

1. **\`symlinkOrJunction()\`** in \`adapter-utils\` — for **directory** symlinks (skill injection). Tries \`fs.symlink()\`, catches only \`EPERM\` on Windows, retries with \`"junction"\` type using an absolute path (junctions require absolute targets). Re-throws all other errors.

2. **\`symlinkOrHardLink()\`** in \`adapter-utils\` — for **file** symlinks (\`auth.json\`). Tries \`fs.symlink()\`, catches only \`EPERM\` on Windows, falls back to \`fs.link()\` (hard link). If hard link also fails with \`EXDEV\` (cross-drive), falls back to \`fs.copyFile()\` as a last resort. Hard links keep contents in sync (same inode); copies do not but are the only option cross-drive.

### Updated call sites

- **adapter-utils**: \`ensurePaperclipSkillSymlink\` default \`linkSkill\` now uses \`symlinkOrJunction\`
- **claude-local**: \`buildSkillsDir\` uses \`symlinkOrJunction\` for skill directory linking. Errors propagate naturally — \`symlinkOrJunction\` handles EPERM internally.
- **codex-local/execute.ts**: skill injection uses \`symlinkOrJunction\`
- **codex-local/codex-home.ts**: \`ensureSymlink\` (for \`auth.json\`) uses \`symlinkOrHardLink\`. Includes inode-based repair for stale hard links on Windows (\`nlink > 1\` guard ensures only hard links are candidates).
[...]

PR #1995, opened 2026-03-28
fix(codex): sanitize sandbox=elevated in managed Codex home config

octasoft-ltd:
## Thinking Path

> - Paperclip creates managed Codex home directories per company, seeded from the user's `~/.codex`
> - `config.toml` is copied verbatim, inheriting any user-specific settings
> - On Windows, users may have `[windows] sandbox = "elevated"` in their Codex config
> - When Paperclip spawns `codex` as a child process, it runs without elevation
> - The elevated sandbox setting causes "Access is denied. (os error 5)" on every heartbeat
> - The fix strips `sandbox = "elevated"` during copy and re-sanitizes existing configs retroactively
> - This is separate from the symlink→junction/hardlink fix in #872 — both are needed on Windows

## Summary

- Add `sanitizeConfigToml()` to strip `sandbox = "elevated"` from `config.toml` when copying to managed Codex home
- Move `config.toml` from `COPIED_SHARED_FILES` to `SANITIZED_COPIED_FILES` so it goes through the sanitization path
- Re-sanitize existing managed configs on each heartbeat (fixes homes created before this patch)
- 9 new tests covering sanitization (double-quoted, single-quoted, bare values, CRLF, section cleanup, retroactive fix)

Companion to #872 which handles the symlink permission issue separately.

## Test plan

- [x] 9 unit tests pass (`codex-local-codex-home.test.ts`)
- [x] Existing codex adapter tests unaffected
- [x] Manual: verify Codex agent heartbeat succeeds on Windows with user config containing `sandbox = "elevated"`

### DIFF
--- a/packages/adapters/codex-local/src/server/codex-home.ts
+++ b/packages/adapters/codex-local/src/server/codex-home.ts
@@ -4,9 +4,34 @@ import path from "node:path";
 import type { AdapterExecutionContext } from "@paperclipai/adapter-utils";
 
 const TRUTHY_ENV_RE = /^(1|true|yes|on)$/i;
-const COPIED_SHARED_FILES = ["config.json", "config.toml", "instructions.md"] as const;
+const COPIED_SHARED_FILES = ["config.json", "instructions.md"] as const;
+const SANITIZED_COPIED_FILES = ["config.toml"] as const;
 const SYMLINKED_SHARED_FILES = ["auth.json"] as const;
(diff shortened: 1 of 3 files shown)


octasoft-ltd:

octasoft-ltd:
@greptileai Rebased onto the latest `master` and squashed into a single commit (previous branch was well behind master). No logic changes. Please re-review.

PR #2008, opened 2026-03-28
feat(plugins): add connector-github bidirectional sync plugin

HugoLopes45:
## Summary

- Adds `connector-github` plugin: bidirectional GitHub ↔ Paperclip sync
- Labels ↔ priority mapping, milestones ↔ goals sync
- Echo deduplication to prevent sync loops (inbound and outbound)
- HMAC webhook signature verification
- Wakeup deduplication to prevent redundant agent runs

## Changes

- `packages/plugins/examples/connector-github/` — new plugin
- `outbound` echo mark placed **after** successful GitHub API call (fixes ordering bug)
- No double-write: `issue.updated` / `issue.comment.created` use `isDuplicateDelivery` only, `markOutboundEcho` not called again

## Review comments addressed

- ✅ `markOutboundIssueEcho` timing: called after `createGitHubIssue` succeeds, not before
- ✅ No double-write in update/comment handlers: they use `outbound:` prefix dedup, no redundant echo mark

## Notes

- `pnpm-lock.yaml` intentionally excluded per CI policy

### DIFF
--- a/packages/plugins/examples/connector-github/package.json
+++ b/packages/plugins/examples/connector-github/package.json
@@ -1,0 +1,39 @@
+{
+  "name": "@paperclipai/connector-github",
+  "version": "0.1.0",
+  "description": "GitHub bidirectional sync connector for Paperclip: issues, pull requests, and webhook-triggered agent wakeups",
+  "type": "module",
+  "private": true,
+  "exports": {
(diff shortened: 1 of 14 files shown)


HugoLopes45:
@greptile review

PR #2047, opened 2026-03-29
fix(ui): show error when model detection fails in agent config

bluzername:
## Problem

The new "Detect from Hermes config" button in the model picker (added in the recent Hermes adapter update) silently swallow all errors. When detection fail for any reason:

- Config file not found -> button go back to normal, nothing happen
- Network error -> same, zero feedback
- Invalid config format -> same
- Detection return null (no model found) -> same

User click the button, see "Detecting..." for a moment, then it go back to "Detect from Hermes config" and they have no idea what went wrong. Maybe they think it work and start looking for the detected model in the list.

## What I changed

Added error handling to the `handleDetectModel()` function:

1. **Catch errors**: When the Promise reject, capture the error message and show it
2. **Handle null result**: When detection return null (no model found), show "No model found in config"
3. **Inline error display**: Show red text below the detect button with the error message
4. **Auto-clear**: Error message clear when user click detect again

## How to test

1. Set up an agent with hermes_local adapter
2. Try to detect model when Hermes is not configured -> should see error message
3. Try with valid config -> should detect normally (error clear if there was one before)

1 file, 9 lines added.

Note: the existing TS error on \`hermes-paperclip-adapter/ui\` module is pre-existing on master (module not yet published) and not related to this change.

### DIFF
--- a/ui/src/components/AgentConfigForm.tsx
+++ b/ui/src/components/AgentConfigForm.tsx
@@ -1359,6 +1359,7 @@ function ModelDropdown({
 }) {
   const [modelSearch, setModelSearch] = useState("");
   const [detectingModel, setDetectingModel] = useState(false);
+  const [detectError, setDetectError] = useState<string | null>(null);
   const selected = models.find((m) => m.id === value);
   const manualModel = modelSearch.trim();
   const canCreateManualModel = Boolean(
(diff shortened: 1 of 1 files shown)


bluzername:

PR #2100, opened 2026-03-29
Add company-level pause/resume kill switch

harperaa:
## Summary

- Adds `POST /companies/:id/pause` and `POST /companies/:id/resume` endpoints for board operators to manually pause/resume all heartbeats for a company
- Heartbeat scheduler `tickTimers()` now skips agents belonging to paused companies
- Dashboard displays a "Company Kill Switch" banner with a toggle button (Pause All / Resume All)
- Pausing cancels all active and queued heartbeat runs via existing `cancelBudgetScopeWork` infrastructure
- Leverages existing `company.status`, `pauseReason`, and `pausedAt` schema fields (no migrations needed)
- Adds `companyStatus` and `companyPauseReason` to `DashboardSummary` type

## Test plan

- [ ] Verify `POST /companies/:id/pause` sets company status to "paused" and cancels active runs
- [ ] Verify `POST /companies/:id/resume` restores company status to "active"
- [ ] Verify heartbeat scheduler does not enqueue runs for agents in paused companies
- [ ] Verify dashboard shows kill switch banner with correct state
- [ ] Verify clicking "Pause All" pauses company and UI updates
- [ ] Verify clicking "Resume All" resumes company and UI updates
- [ ] Verify budget-initiated pauses still work alongside manual pauses
- [ ] All existing tests pass (137 files, 682 tests)

### DIFF
--- a/packages/shared/src/types/dashboard.ts
+++ b/packages/shared/src/types/dashboard.ts
@@ -8,6 +8,8 @@ export interface DashboardRunActivityDay {
 
 export interface DashboardSummary {
   companyId: string;
+  companyStatus: string;
+  companyPauseReason: string | null;
   agents: {
     active: number;
     running: number;
--- a/packages/shared/src/types/dashboard.ts
+++ b/packages/shared/src/types/dashboard.ts
@@ -8,7 +8,7 @@ export interface DashboardRunActivityDay {
 
 export interface DashboardSummary {
   companyId: string;
-  companyStatus: string;
+  companyStatus: "active" | "paused" | "archived";
   companyPauseReason: string | null;
   agents: {
     active: number;
--- a/server/src/routes/companies.ts
+++ b/server/src/routes/companies.ts
@@ -21,6 +21,7 @@ import {
   companyPortabilityService,
   companyService,
   feedbackService,
+  heartbeatService,
   logActivity,
 } from "../services/index.js";
 import type { StorageService } from "../storage/types.js";
@@ -34,6 +35,7 @@ export function companyRoutes(db: Db, storage?: StorageService) {
   const access = accessService(db);
   const budgets = budgetService(db);
   const feedback = feedbackService(db);
+  let _heartbeat: ReturnType<typeof heartbeatService> | null = null;
 
   function parseBooleanQuery(value: unknown) {
     return value === true || value === "true" || value === "1";
@@ -57,6 +59,11 @@ export function companyRoutes(db: Db, storage?: StorageService) {
       return;
     }
     assertCompanyAccess(req, target.companyId);
+  }
+
+  function getHeartbeat() {
+    if (!_heartbeat) _heartbeat = heartbeatService(db);
+    return _heartbeat;
   }
 
   async function assertCanUpdateBranding(req: Request, companyId: string) {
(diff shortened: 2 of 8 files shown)


harperaa:
## Relationship to #1784

PR #1784 adds company emergency pause/resume controls to the dashboard. This PR (#2100) addresses the same operator need but takes a different architectural approach. Here's a detailed comparison:

### What #1784 does

- UI-only change (single file: `Dashboard.tsx`)
- Uses the generic `companiesApi.update(companyId, { status: "paused" })` PATCH endpoint to toggle company status
- Adds a "Hard stop" action that fetches live runs via `heartbeatsApi.liveRunsForCompany()` and cancels them individually from the client
- No server-side changes to the heartbeat scheduler or company service

### What this PR (#2100) does differently and why

**1. Dedicated `/pause` and `/resume` endpoints instead of generic PATCH**

Rather than allowing the generic company update endpoint to change status (which any field-level PATCH could do, potentially bypassing intent), this PR adds purpose-built `POST /companies/:id/pause` and `POST /companies/:id/resume` endpoints. This makes the operation explicit, auditable with a distinct `company.paused` / `company.resumed` activity log action, and easier to permission-gate separately in the future.

**2. Server-side run cancellation instead of client-side**

In #1784, the hard-stop flow fetches all live runs from the API and cancels them one by one from the browser. This has a race condition: new runs can be enqueued by the heartbeat scheduler between the time the client fetches the run list and finishes cancelling them. This PR moves cancellation server-side — the pause endpoint calls `cancelBudgetScopeWork()` atomically in the same request, using the existing company-scoped cancellation infrastructure that the budget system already relies on. No runs slip through.

**3. Heartbeat scheduler gating on company status**

This is the most important difference. #1784 does not modify the heartbeat scheduler, so after pausing a company, `tickTimers()` will continue to enqueue new heartbeat runs for agents in that company — they just won't execute because the company status blocks them downstream. This PR adds a direct check in `tickTimers()`:

```typescript
const pausedCompanyIds = new Set(
  (await db.select({ id: companies.id }).from(companies).where(eq(companies.status, "paused")))
    .map((r) => r.id)
);
// ...
if (pausedCompanyIds.has(agent.companyId)) continue;
```

This prevents the scheduler from even attempting to enqueue runs for paused companies, which is cleaner and avoids unnecessary DB writes for queued runs that would immediately be skipped.

**4. Budget-pause safety guard**

The budget enforcement system can also pause companies (when a hard-stop threshold is hit). This PR's `resume()` function scopes its update to `WHERE pause_reason = 'manual'`, so a board operator pressing "Resume All" cannot accidentally override a budget-initiated pause. #1784 uses the generic status update which would clear any pause regardless of origin, potentially allowing agents to incur spend beyond budget limits until the next enforcement tick re-pauses the company.

**5. Cancellation reason in run logs**

`cancelBudgetScopeWork()` previously hardcoded `"Cancelled due to budget pause"` as the cancellation reason in run logs. This PR adds an optional `reason` parameter so manual pauses log `"Cancelled due to manual company pause"` — making it clear in run history whether a cancellation was operator-initiated or budget-enforced.

### Summary

| Concern | #1784 | #2100 (this PR) |
|---------|-------|-----------------|
| Pause mechanism | Generic PATCH on status | Dedicated `/pause` endpoint |
| Run cancellation | Client-side, per-run | Server-side, atomic via `cancelBudgetScopeWork` |
| Scheduler gating | None — runs still enqueued | `tickTimers()` skips paused companies |
| Budget-pause safety | Can override budget pause | Scoped to `pause_reason = 'manual'` only |
| Cancel reason in logs | N/A | Distinct reason per pause type |
| Activity log action | Generic update | `company.paused` / `company.resumed` |
| Files changed | 1 (UI only) | 9 (shared types, server, UI) |

Both PRs solve the same user-facing problem. This one goes deeper into the server to close the gaps around race conditions, budget safety, and scheduler behavior.

PR #2189, opened 2026-03-30
fix: handle Windows paths with spaces in resolveSpawnTarget

vantanco:
## Problem

On Windows, when the npm global directory sits inside a user profile whose name contains spaces (e.g. `C:\Users\Admin Backup\AppData\Roaming\npm\codex.CMD`), the Codex hello probe (and all agent spawning) fails with:

```
'"C:\Users\Admin Backup\AppData\Roaming\npm\codex.CMD"' is not recognized as an internal or external command
```

This affects **any Windows user whose username contains a space** (very common: `John Doe`, `Admin Backup`, etc.).

## Root Cause

`resolveSpawnTarget()` in `adapter-utils` wraps `.cmd/.bat` executables with `cmd.exe /d /s /c`. When the executable path contains spaces, `quoteForCmd()` correctly adds double quotes around it. However, Node.js `spawn()` on Windows builds the command line for `CreateProcess` and adds its **own** quotes around arguments containing spaces. The `/s` flag then strips the outermost quotes, leaving the path unquoted and broken into multiple tokens by `cmd.exe`.

## Fix

When the resolved `.cmd/.bat` path contains whitespace:
1. Read the `.cmd` wrapper file content
2. Extract the underlying `.js` entrypoint (npm-generated `.cmd` files always reference it via `%dp0%\...\entry.js`)
3. Spawn `node.exe` directly with the `.js` file, bypassing `cmd.exe` entirely

Falls back to the original `cmd.exe` approach for paths without spaces or when extraction fails.

## Testing

Verified on Windows 11 with path `C:\Users\Admin Backup\AppData\Roaming\npm\codex.CMD`:
- Before fix: `Codex hello probe failed`
- After fix: `codex-cli 0.117.0` responds correctly

The fix is backward-compatible: paths without spaces continue to use the existing `cmd.exe` approach unchanged.

### DIFF
--- a/packages/adapter-utils/src/server-utils.ts
+++ b/packages/adapter-utils/src/server-utils.ts
@@ -329,7 +329,7 @@ async function resolveSpawnTarget(
       try {
         const cmdContent = await fs.readFile(executable, "utf8");
         // npm-generated .cmd wrappers contain: "%dp0%\path\to\entry.js"
-        const jsMatch = cmdContent.match(/%dp0%\\([^\s"]+\.js)/);
+        const jsMatch = cmdContent.match(/%(?:~)?dp0%?\\([^\s"]+\.js)/i);
         if (jsMatch) {
           const jsFile = path.join(path.dirname(executable), jsMatch[1]!);
           if (await pathExists(jsFile)) {
--- a/packages/adapter-utils/src/server-utils.ts
+++ b/packages/adapter-utils/src/server-utils.ts
@@ -320,6 +320,31 @@ async function resolveSpawnTarget(
   }
 
   if (/\.(cmd|bat)$/i.test(executable)) {
+    // When the executable path contains spaces (e.g. C:\Users\Admin Backup\...),
+    // cmd.exe /d /s /c breaks because Node.js adds extra quotes around the
+    // command-line argument in CreateProcess, which interact badly with /s
+    // quote-stripping behaviour. Work around this by reading the .cmd wrapper
+    // to extract the underlying .js entrypoint and spawning node directly.
+    if (/\s/.test(executable)) {
+      try {
+        const cmdContent = await fs.readFile(executable, "utf8");
+        // npm-generated .cmd wrappers contain: "%dp0%\path\to\entry.js"
+        const jsMatch = cmdContent.match(/%dp0%\\([^\s"]+\.js)/);
+        if (jsMatch) {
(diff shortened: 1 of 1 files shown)


vsetchinfc:
Ran into the same root cause testing the `claude_local` adapter on Windows (default install path `C:\Program Files\nodejs\claude.CMD`) — confirmed it's the same `resolveSpawnTarget` `.cmd`/`.bat` branch.

One gap with this PR's approach for that specific case: the regex here only matches a `.js` entrypoint inside the wrapper. Claude Code's actual `.cmd` wrapper points to a compiled `.exe`, not a `.js` file:

```bat
@ECHO off
GOTO start
:find_dp0
SET dp0=%~dp0
EXIT /b
:start
SETLOCAL
CALL :find_dp0
"%dp0%\node_modules\@anthropic-ai\claude-code\bin\claude.exe"   %*
```

So even with the broader `%(?:~)?dp0%?` regex Greptile suggested above, this would still silently fall through to the original broken `cmd.exe` path for the Claude adapter specifically — the bypass-node.exe strategy only works when the wrapper actually delegates to a `.js` file.

We fixed it by addressing the `cmd.exe` invocation directly instead of bypassing it, so it's agnostic to what the `.cmd` wrapper points to (`.js` or `.exe`):

```ts
// cmd.exe's /C quote-stripping only correctly identifies "this whole string
// is one quoted command + args" when the entire line is wrapped in one more
// outer pair of quotes. Without it, a quoted executable path (required
// whenever it contains spaces, e.g. the default "C:\Program Files\nodejs\..."
// install location) followed by unquoted args makes cmd.exe try to execute
// the literal string including the quote characters.
const shell = resolveWindowsCmdShell(env);
const commandLine = [quoteForCmd(executable), ...args.map(quoteForCmd)].join(" ");
return {
  command: shell,
  args: ["/d", "/s", "/c", `"${commandLine}"`],
  // Also required: without this, Node re-quotes the already-quoted string
  // itself before handing it to CreateProcess, producing "The filename,
  // directory name, or volume label syntax is incorrect." instead.
  windowsVerbatimArguments: true,
};
```

(`windowsVerbatimArguments` would need threading through to the actual `spawn()` call in `runChildProcess`.)

Tested end-to-end against a real Claude Code install (default `C:\Program Files\nodejs\claude.CMD` path) — adapter environment check now passes, agent launches and runs successfully. Happy to open this as a separate PR if useful, or it could fold into this one as a fallback for the cases the `.js`-extraction approach doesn't cover.

PR #2228, opened 2026-03-30
fix(ui): detect and warn about duplicate env var keys in agent config

bluzername:
## Problem

The environment variables editor in agent configuration allow user to add the same key multiple times. For example you can have two rows both named "DATABASE_URL" with different values. When the config save, JavaScript object overwrite duplicate keys so only the LAST value is kept. The first one silently disappear.

I spent time debugging why my agent was not using the correct database URL - turns out I had the key defined twice and the wrong one was winning.

## What I changed

Added duplicate key detection using useMemo:
- Compute set of duplicate keys (case-insensitive, so DB_URL and db_url count as same)
- Amber border on duplicate key inputs so they stand out visually
- Tooltip on hover: "Duplicate key - only the last value will be used"
- Warning text below the env var list: "Duplicate keys detected: DB_URL, API_KEY. Only the last value for each key will be used."

The detection is case-insensitive because environment variables are typically case-sensitive but having DB_URL and db_url is almost always a mistake.

## How to test

1. Go to any agent > Config tab > Environment variables section
2. Add two rows with the same key name (like "TEST")
3. Both inputs should get amber border
4. Warning message should appear below the list
5. Fix by removing one duplicate - warning disappear

1 file, 13 lines added.

### DIFF
--- a/ui/src/components/AgentConfigForm.tsx
+++ b/ui/src/components/AgentConfigForm.tsx
@@ -1236,6 +1236,15 @@ function EnvVarEditor({
     }
   }
 
+  const duplicateKeys = useMemo(() => {
+    const seen = new Map<string, number>();
+    for (const row of rows) {
+      const k = row.key.trim().toUpperCase();
+      if (k) seen.set(k, (seen.get(k) ?? 0) + 1);
+    }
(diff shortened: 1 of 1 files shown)


bluzername:

PR #2282, opened 2026-03-31
fix: guard onboarding tab navigation until setup succeeds

### LINKED ISSUE #2080: OnboardingWizard: clicking Agent tab directly skips company creation, model list empty
## Description

In the onboarding wizard, if the user clicks the **Agent** tab at the top (step 2) instead of clicking the **Next** button after filling in the company name, the company is not created. This causes `createdCompanyId` to remain `null`, which disables the adapter models query:

```typescript
enabled: Boolean(createdCompanyId) && effectiveOnboardingOpen && step === 2
```

As a result, the model dropdown is empty for all adapter types — not just a specific adapter.

## Steps to reproduce

1. Open Paperclip with no companies (fresh instance)
2. Onboarding wizard appears at step 1 (Company)
3. Fill in a company name
4. Instead of clicking **Next**, click the **Agent** tab at the top
5. Select any adapter (e.g., Claude Code, Codex, GitHub Copilot)
6. Click the Model dropdown

**Expected:** Model list is populated
**Actual:** Model dropdown is empty (no models, no "Default" option)

## Root cause

The "Next" button at step 1 triggers company creation via API AND advances to step 2. The tab click only advances to step 2 without creating the company. Since the models endpoint requires a company ID (`/companies/:companyId/adapters/:type/models`), the query never fires.

## Suggested fix

Either:
1. Trigger company creation when the Agent tab is clicked (if not already created)
2. Disable the Agent/Task/Launch tabs until step 1 completes via the Next button
3. Decouple the models query from companyId (the models endpoint doesn't use company-specific data)

Found during development of the copilot_local adapter, but affects all adapters.

googlarz:
## Summary
- route onboarding tab clicks through the existing step guard logic
- create the company before allowing the wizard to jump into later tabs
- stop invalid tab skips from leaving later steps in a broken state

## Testing
- pnpm --filter @paperclipai/ui typecheck

Closes #2080

### DIFF
--- a/ui/src/components/OnboardingWizard.tsx
+++ b/ui/src/components/OnboardingWizard.tsx
@@ -382,6 +382,18 @@ export function OnboardingWizard() {
   }
 
   async function handleStep1Next() {
+    const companyId = await ensureCompanyCreated();
+    if (companyId) {
+      setStep(2);
+    }
+  }
+
+  async function ensureCompanyCreated(): Promise<string | null> {
+    if (createdCompanyId) return createdCompanyId;
+    if (!companyName.trim()) {
+      setError("Enter a company name before continuing.");
+      return null;
+    }
     setLoading(true);
     setError(null);
     try {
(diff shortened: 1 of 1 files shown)


googlarz:

PR #2298, opened 2026-03-31
fix(server): clear persisted sessions on adapter change

pacifists:
Title: fix(server): clear persisted sessions on adapter change

## Summary

This clears persisted runtime/task session state when an agent's `adapterType` changes.

Without this, Paperclip can carry a session ID created by one adapter into a different adapter after an agent migration, which causes the new adapter to attempt to resume an incompatible session.

## Problem

Paperclip persists session continuity in:

- `agent_runtime_state.session_id`
- `agent_task_sessions.session_display_id`
- `agent_task_sessions.session_params_json`

Those persisted values are adapter-specific, but they were not being cleared when an agent changed adapters.

That meant a migration like:

- `claude_local` -> `codex_local`

could leave behind a stale session ID from the previous adapter.

On the next heartbeat, Paperclip would feed that stale session into the new adapter's resume path.

## User-facing failure

In production this surfaced as a `codex_local` heartbeat repeatedly trying to resume an old thread:

```bash
codex exec --json --dangerously-bypass-approvals-and-sandbox --model gpt-5.4 resume 789ad670-11d4-4fd3-a5a6-af74e50af5a3 -
```

Codex then failed with:

```text
Error: thread/resume: thread/resume failed: no rollout found for thread id 789ad670-11d4-4fd3-a5a6-af74e50af5a3
```

The immediate operational fix was to manually clear:

- `agent_runtime_state.session_id`
- `agent_task_sessions.session_display_id`
- `agent_task_sessions.session_params_json`

across affected agents.

## Root cause

Session continuity was preserved across adapter migrations even though session IDs are not portable across adapters.

The persisted session state belonged to the previous adapter, but Paperclip treated it as resumable state for the newly selected adapter.

This is especially problematic for local adapters (`claude_local`, `codex_local`, `cursor`, `gemini_local`, etc.) because each adapter owns its own session format and resume mechanism.

## Fix

The fix lives in `agentService.update()`, which is the correct central place because adapter changes can happen through multiple flows:

- direct agent PATCH updates
- rollback/config revision flows
- company portability/import/update flows
- any other service-layer caller using `agents.update(...)`

When `adapterType` changes, we now do all of the following in the same transaction as the agent update:

1. Update the agent record.
2. Delete persisted task sessions for that agent from `agent_task_sessions`.
3. Clear runtime session state in `agent_runtime_state`:
   - `sessionId = null`
   - `stateJson = {}`
   - `lastError = null`
   - `adapterType = <new adapter>`
4. Preserve the existing config revision behavior.

This ensures the next heartbeat starts a fresh session under the new adapter instead of trying to resume stale state from the previous one.

## Why this approach

- It fixes the real source of the bug rather than adding adapter-specific recovery hacks.
- It is adapter-agnostic and safe for all future adapter migrations.
- It keeps session cleanup consistent for every update path, not just the HTTP route.
- It avoids partial state by doing the cleanup transactionally with the adapter change.

## Tests

Added DB-backed regression coverage in:

- `server/src/__tests__/agent-adapter-session-reset.test.ts`

The new tests verify:

1. Changing `adapterType` clears both persisted runtime and task sessions.
2. Updating agent config without changing `adapterType` preserves valid sessions.

Also re-ran adjacent session logic coverage:

- `server/src/__tests__/heartbeat-workspace-session.test.ts`

## Verification

Executed:

```bash
npx vitest run server/src/__tests__/agent-adapter-session-reset.test.ts server/src/__tests__/heartbeat-workspace-session.test.ts
```

Result:

- 2 test files passed
- 28 tests passed

## Notes

This intentionally resets session continuity on adapter migration. That is the correct behavior because adapter sessions are not portable.

### DIFF
--- a/server/src/__tests__/agent-adapter-session-reset.test.ts
+++ b/server/src/__tests__/agent-adapter-session-reset.test.ts
@@ -1,0 +1,187 @@
+import { randomUUID } from "node:crypto";
+import { eq } from "drizzle-orm";
+import { afterAll, afterEach, beforeAll, describe, expect, it } from "vitest";
+import {
+  agents,
+  agentRuntimeState,
+  agentTaskSessions,
+  companies,
+  createDb,
+} from "@paperclipai/db";
+import {
+  getEmbeddedPostgresTestSupport,
+  startEmbeddedPostgresTestDatabase,
+} from "./helpers/embedded-postgres.js";
+import { agentService } from "../services/agents.ts";
+
+const embeddedPostgresSupport = await getEmbeddedPostgresTestSupport();
+const describeEmbeddedPostgres = embeddedPostgresSupport.supported ? describe : describe.skip;
+
+if (!embeddedPostgresSupport.supported) {
+  console.warn(
+    `Skipping embedded Postgres agent adapter session reset tests on this host: ${embeddedPostgresSupport.reason ?? "unsupported environment"}`,
+  );
+}
+
+describeEmbeddedPostgres("agentService adapter session reset", () => {
+  let db!: ReturnType<typeof createDb>;
+  let svc!: ReturnType<typeof agentService>;
+  let tempDb: Awaited<ReturnType<typeof startEmbeddedPostgresTestDatabase>> | null = null;
+
+  beforeAll(async () => {
+    tempDb = await startEmbeddedPostgresTestDatabase("paperclip-agent-adapter-reset-");
+    db = createDb(tempDb.connectionString);
+    svc = agentService(db);
+  }, 20_000);
+
+  afterEach(async () => {
+    await db.delete(agentTaskSessions);
+    await db.delete(agentRuntimeState);
+    await db.delete(agents);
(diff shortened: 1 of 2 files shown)


developwithjbhh:
Validated this failure live on a local Paperclip `2026.403.0` instance after switching an agent from `claude_local` to `codex_local`.

This PR addresses the root cause cleanly: adapter-specific session state should not survive an `adapterType` change.

What we observed before the fix:
- `agent_runtime_state.session_id` still pointed at the previous provider thread
- task sessions also contained stale adapter-owned session state
- next Codex heartbeat failed immediately on `thread/resume ... no rollout found`

What fixed it locally:
- clear persisted runtime session + task sessions on adapter change
- then next run started cleanly on a fresh thread

One note from live validation: I would still pair this with a heartbeat-side guard like #1996 as defense in depth, because already-dirty runtime state can exist before cleanup lands. But if choosing the single best root-cause fix, this is the one I would back first.

PR #2303, opened 2026-03-31
fix: retry fresh when OpenCode resumes into an empty stale session

### LINKED ISSUE #2253: opencode_local: stale session ID in agent_runtime_state causes silent exit (exit 0, 0 tokens, no work done)
## Bug

When an `opencode_local` agent's `session_id` in `agent_runtime_state` contains a UUID that no longer exists in OpenCode's session store (e.g. because OpenCode was upgraded from a version that stored sessions as UUIDs to the current version that stores them as `ses_<hash>`), every subsequent heartbeat run silently exits with code 0, produces no tokens, and does no work. The run is recorded as `succeeded` with `last_run_status: succeeded`, so the error is completely invisible.

## Root Cause

In `execute.js` (`@paperclipai/adapter-opencode-local`):

```js
const runtimeSessionId = asString(runtimeSessionParams.sessionId, runtime.sessionId ?? "");
const canResumeSession = runtimeSessionId.length > 0 && ...;
const sessionId = canResumeSession ? runtimeSessionId : null;
```

When `canResumeSession` is true, `--session <stale-uuid>` is passed to `opencode run`. OpenCode then **silently exits with code 0 and produces no stdout** when the given session ID doesn't exist in its store.

The retry logic in `execute.js` only triggers when the run **fails** (`initialFailed`):

```js
if (sessionId && initialFailed && isOpenCodeUnknownSessionError(...)) {
  // retry with fresh session
}
```

Since OpenCode exits with code 0 and writes no error output, `initialFailed` is `false`. The retry never fires. The `toResult` function then falls back to `runtimeSessionId` (because `parsed.sessionId` is null), so `session_id_after` is set to the same stale UUID — **perpetuating the bug forever**.

## Reproduction

1. Upgrade OpenCode from a version that stored sessions as UUIDs to v1.3.9+ (which uses `ses_<hash>` format)
2. Agent's `agent_runtime_state.session_id` still contains the old UUID format
3. Trigger a heartbeat run
4. Run completes with `status: succeeded`, `exit_code: 0`, `input_tokens: 0`, `output_tokens: 0`, 1 NDJSON line (the workspace warning only), `session_id_after: <same stale uuid>`
5. Repeat forever

## Evidence

Run log (`adapter.invoke` event):
```
commandArgs: ["run", "--format", "json", "--session", "43595c05-f3e2-453a-83df-a820bb791c64", ...]
resultJson: { "stdout": "", "stderr": "" }
usageJson: { "inputTokens": 0, "outputTokens": 0, "sessionReused": true, "freshSession": false }
```

All runs for the affected agents produced exactly 269 bytes (1 NDJSON line = the workspace warning).

## Affected Agents

Any `opencode_local` agent whose `agent_runtime_state.session_id` contains a session ID that doesn't exist in the local OpenCode instance. This can happen via:
- OpenCode version upgrades that change session ID format
- Moving the OpenCode data directory
- Reinstalling OpenCode  
- Running agents on a different machine

### Option A — In `execute.js`: treat empty stdout as unknown session (preferred)

When `opencode run --session <id>` produces empty stdout AND exits with code 0, treat it as an unknown session error and retry with a fresh session (set `clearSession: true`):

```js
const isEmptySuccessfulRun = !initial.proc.timedOut && 
  (initial.proc.exitCode ?? 0) === 0 && 
  !initial.proc.stdout.trim() &&
  sessionId !== null;

if (isEmptySuccessfulRun || (sessionId && initialFailed && isOpenCodeUnknownSessionError(...))) {
  await onLog("stdout", `[paperclip] Session "${sessionId}" produced no output; retrying fresh.\n`);
  return toResult(retry, true);
}
```

### Option B — In OpenCode: return non-zero or write an error JSON line when `--session` doesn't exist

OpenCode should write `{"type":"error","message":"session not found: <id>"}` and exit non-zero when the given session doesn't exist. This would make `initialFailed` true and trigger the existing retry.

### Option C — In heartbeat.js: validate session existence before passing to adapter

Before building `runtimeForAdapter`, check whether the stored `session_id` actually exists (e.g., via a lightweight `opencode session list` call or filesystem check). Clear it if missing.

**Option A is the most defensive** and fixes the bug within the adapter without requiring OpenCode or heartbeat.js changes.

## Workaround

Clear `session_id` in `agent_runtime_state` directly:

```sql
UPDATE agent_runtime_state 
SET session_id = NULL 
WHERE agent_id = '<agent-id>';
```

Or via the Paperclip API (if exposed): `PATCH /api/agents/:id` with `{ "clearSession": true }` (currently not supported, would be a useful addition).

## Environment

- Paperclip server: latest (`@paperclipai/server`)
- OpenCode: v1.3.9
- Adapter: `@paperclipai/adapter-opencode-local`
- Affected agents: `opencode_local` adapter type

googlarz:
## Summary
- treat an empty successful resumed OpenCode run as a stale-session case
- retry once without --session instead of silently preserving the bad session forever
- add execute-level regression coverage for the empty-output stale-session path

## Testing
- pnpm exec vitest run server/src/__tests__/opencode-local-execute.test.ts server/src/__tests__/opencode-local-adapter.test.ts
- pnpm --filter @paperclipai/server typecheck
- pnpm --filter @paperclipai/adapter-opencode-local typecheck

Closes #2253

### DIFF
--- a/packages/adapters/opencode-local/src/server/execute.ts
+++ b/packages/adapters/opencode-local/src/server/execute.ts
@@ -10,12 +10,11 @@ import {
   parseObject,
   buildPaperclipEnv,
   joinPromptSections,
-  buildInvocationEnvForLogs,
+  redactEnvForLogs,
   ensureAbsoluteDirectory,
   ensureCommandResolvable,
   ensurePaperclipSkillSymlink,
   ensurePathInEnv,
-  resolveCommandForLogs,
   renderTemplate,
   runChildProcess,
   readPaperclipRuntimeSkillEntries,
@@ -34,6 +33,20 @@ function firstNonEmptyLine(text: string): string {
       .split(/\r?\n/)
       .map((line) => line.trim())
       .find(Boolean) ?? ""
+  );
+}
+
+function isEmptySuccessfulResumeRun(input: {
+  sessionId: string | null;
+  timedOut: boolean;
+  exitCode: number | null;
+  stdout: string;
+}): boolean {
+  return Boolean(
+    input.sessionId &&
+      !input.timedOut &&
+      (input.exitCode ?? 0) === 0 &&
+      input.stdout.trim().length === 0,
   );
 }
 
@@ -187,12 +200,6 @@ export async function execute(ctx: AdapterExecutionContext): Promise<AdapterExec
       ),
     );
     await ensureCommandResolvable(command, cwd, runtimeEnv);
-    const resolvedCommand = await resolveCommandForLogs(command, cwd, runtimeEnv);
-    const loggedEnv = buildInvocationEnvForLogs(preparedRuntimeConfig.env, {
-      runtimeEnv,
-      includeRuntimeKeys: ["HOME"],
-      resolvedCommand,
-    });
 
     await ensureOpenCodeModelConfiguredAndAvailable({
       model,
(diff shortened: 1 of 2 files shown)


googlarz:

PR #2306, opened 2026-03-31
fix(adapter-utils): spawn child processes with detached mode for proper process group cleanup

### LINKED ISSUE #2297: [BUG] Orphan grandchild process survives timeout on Linux/macOS
## Summary

On Linux/macOS, when a heartbeat run times out, the orphaned grandchild process (opencode native binary) is not killed. Only the dead wrapper process is targeted.

## Root Cause

Paperclip spawns two layers:

```
Paperclip server
  └── node wrapper (child.pid = X)
        └── opencode native binary (grandchild, PID Y)
```

When the wrapper exits early, the native binary is re-parented to systemd/init. The timeout handler calls `process.kill(-child.pid, SIGTERM)` but `child.pid` is the dead wrapper. The signal goes nowhere.

## Existing Infrastructure

- `onSpawn` already saves PID to `heartbeat_runs.processPid` but stores the **wrapper PID**, not the native binary PID.
- `isProcessAlive(pid)` exists in server-utils.ts
- `killProcessGroup` uses `process.kill(-child.pid, signal)` which only works if child.pid is the PGID leader

## Fix

Add `detached: true` + `child.unref()` to the spawn call in `runChildProcess` (packages/adapter-utils/src/server-utils.ts):

```typescript
const child = spawn(target.command, target.args, {
  cwd: opts.cwd,
  env: mergedEnv,
  shell: false,
  detached: true,
  stdio: [...],
});
child.unref();
```

This makes the wrapper become the PGID leader. Grandchild inherits the same PGID. `process.kill(-child.pid)` then kills the entire group, including orphans.

## Verification

1. Start a heartbeat run that spawns a child process
2. Wait for timeout
3. Confirm no orphan processes remain: `ps aux | grep opencode`

## Related

- #2266 addresses the same problem on Windows (uses `taskkill /F /T /PID`)

shoaib050326:
- Paperclip orchestrates AI agents that spawn child processes for execution
- When heartbeat runs time out, the adapter needs to clean up all spawned processes
- On Linux/macOS, the process tree has two layers: node wrapper → opencode native binary
- When only the wrapper is killed, the grandchild becomes orphaned and survives
- This leaves zombie processes running that consume resources and can cause issues
- The fix adds detached: true to spawn() making the wrapper a process group leader
- Adding child.unref() allows the parent to exit without waiting for the child
- Now process.kill(-child.pid, SIGTERM) kills the entire group including orphans
- This matches the approach in PR #2266 which fixed the same issue on Windows using taskkill

Closes #2297

### DIFF
--- a/packages/adapter-utils/src/server-utils.ts
+++ b/packages/adapter-utils/src/server-utils.ts
@@ -788,8 +788,10 @@ export async function runChildProcess(
           cwd: opts.cwd,
           env: mergedEnv,
           shell: false,
+          detached: true,
           stdio: [opts.stdin != null ? "pipe" : "ignore", "pipe", "pipe"],
         }) as ChildProcessWithEvents;
+        child.unref();
         const startedAt = new Date().toISOString();
 
         if (opts.stdin != null && child.stdin) {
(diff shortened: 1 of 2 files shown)


brandonburr:
Real-world hit on macOS that this PR would have prevented: a paperclip-launched `next dev` (Next.js 16.2.2 + Tailwind v4 + Turbopack) inside a project's heartbeat leaked `@tailwindcss/postcss` workers to **4,047 procs (~155 GB RSS)** in ~4 min, then froze the host hard enough to need a force-restart. Force-quitting the parent terminal didn't recover the system — the postcss workers had already re-parented to PID 1 and survived, with kernel_task doing **~700 GB of swap I/O** trying to manage the orphaned address spaces.

The `detached: true` + `process.kill(-pid, ...)` group-kill primitive in this PR is exactly the gap. Without it, the project's own hand-rolled `scripts/dev-server.sh` (`pkill -9 -f "${PROJECT_NAME}/.next/dev/build/postcss.js"`) is the only thing that reliably cleans up — and it doesn't help when paperclip's launcher bypasses it.

PR #2332, opened 2026-03-31
Fix/gemini cli probe

aramponi:
**Title**: fix(gemini-local): support updated Gemini CLI JSON output format
### Description:
**Fixes:**
- Added fallback parsing logic to support `type: "message"` and `role: "assistant"` in `parseGeminiJsonl`.
- Updated usage metrics tracking to fallback to `event.stats` and `stats.cached` when standard `event.usage` metadata is missing.
- Appended `detail` output to UI error notifications on `gemini_hello_probe_timed_out` to help future debugging without requiring developers to check server logs.
**Verification:**
- Ran `pnpm -r typecheck`, `pnpm test:run` and `pnpm build` locally with no errors.
- The Gemini probe now succeeds on the `/api/companies/.../agents` interface instead of silently defaulting to empty arrays and failing the check.


Cruel:
> It is still failing for me in the hello probe. Gemini cli needs `--prompt` argument to make it work in non-interactive mode, it is using TUI (interactive mode) by default which is messing with the current test.

It is using `--prompt`

https://github.com/paperclipai/paperclip/blob/6c2c63e0f1b0c3f15d4868a6170c0201ff08f5b0/packages/adapters/gemini-local/src/server/test.ts#L145

Seonfx:
> It is still failing for me in the hello probe. Gemini cli needs `--prompt` argument to make it work in non-interactive mode, it is using TUI (interactive mode) by default which is messing with the current test.

Try #2042

PR #2345, opened 2026-03-31
fix claude-local error result handling and prompt-limit session retry

MindSyncHub:
> - Paperclip orchestrates AI agents as a company, and adapter reliability is part of that control plane
> - Claude local runs can sometimes return a structured result payload that signals failure even when the CLI exits successfully
> - That means Paperclip can misread a failed Claude run as success, which is especially risky for long-lived sessions
> - Resumed Claude sessions can also hit prompt/context limits over time
> - So this pull request teaches the claude-local adapter to detect structured error results, classify prompt-limit failures, and retry with a fresh session when an old session is too large
> - It also surfaces those automatic session recoveries more clearly in the run transcript UI
> - The benefit is more reliable Claude execution and better operator visibility when Paperclip has to recover automatically

## What changed

- detect Claude result payloads with `is_error: true` as failures even if the process exits with code 0
- improve failure parsing so misleading `subtype: "success"` values do not appear in failure messages
- detect prompt-too-long/context-limit failures from Claude result text
- automatically retry with a fresh session when a resumed Claude session exceeds the prompt/context limit
- render these Paperclip retry events as friendly transcript session events in the UI
- add backend and UI tests covering the new behavior

## Why this matters

This makes `claude_local` more robust in long-running or resumed-session workflows. Without this, Paperclip can treat a Claude failure as success or leave the operator with a confusing raw system log when automatic recovery happens.

## Verification

- `pnpm -r typecheck`
- `pnpm test:run`
- `pnpm build`

## Manual testing notes

- verified that a Claude result with `is_error: true` now produces an adapter error message even when the process exits successfully
- verified that a resumed Claude session that returns a prompt-too-long failure is retried with a fresh session
- verified that the transcript shows friendly session recovery events for prompt-limit retry and missing-session retry

### DIFF
--- a/packages/adapters/claude-local/src/server/execute.ts
+++ b/packages/adapters/claude-local/src/server/execute.ts
@@ -586,8 +586,8 @@ export async function execute(ctx: AdapterExecutionContext): Promise<AdapterExec
     if (
       sessionId &&
       !initial.proc.timedOut &&
-      (initial.proc.exitCode ?? 0) !== 0 &&
       initial.parsed &&
+      ((initial.proc.exitCode ?? 0) !== 0 || isClaudeErrorResult(initial.parsed)) &&
       isClaudeUnknownSessionError(initial.parsed)
     ) {
       await onLog(


MindSyncHub:
Followed up on the initial feedback:

- rebased onto the latest `master`
- aligned the missing-session retry path with the new structured error handling
- cleaned up the branch so the PR only contains the intended Claude-local changes

All checks are passing now.

MindSyncHub:
Replacement PR: #2574. Closing this one in favor of the rebuilt clean branch/PR based directly on current upstream master.

PR #2361, opened 2026-03-31
fix(claude-local): retry on stale session when Claude produces no parsed output

balter-io:
## Problem

All agents fail with `Claude exited with code 1: No conversation found with session ID: ...` when resuming a stale Claude session.

When Claude CLI fails immediately on a stale `--resume` session, it exits with code 1 and prints the error **only to stderr** — no JSON on stdout. The existing retry logic in `execute()` requires `initial.parsed` to be truthy to detect the session error and retry with a fresh session. Since there's no parseable stdout, `parsed` is null and the retry is silently skipped, causing permanent run failure.

## Fix

- Extract the session-error regex into a shared `CLAUDE_SESSION_ERROR_RE` constant in `parse.ts`
- Add `isClaudeUnknownSessionStderr(stderr)` to match the pattern against raw stderr text
- Extend the retry condition in `execute()` to fall back to stderr matching when `parsed` is null

## Files changed

- `packages/adapters/claude-local/src/server/parse.ts`
- `packages/adapters/claude-local/src/server/execute.ts`
- `packages/adapters/claude-local/src/server/index.ts`

---
[Warp conversation](https://app.warp.dev/conversation/5b88ab71-83ae-4e2b-8bdc-8f661807beca)

Co-Authored-By: Oz <oz-agent@warp.dev>

### DIFF
--- a/packages/adapters/claude-local/src/server/execute.ts
+++ b/packages/adapters/claude-local/src/server/execute.ts
@@ -28,6 +28,7 @@ import {
   detectClaudeLoginRequired,
   isClaudeMaxTurnsResult,
   isClaudeUnknownSessionError,
+  isClaudeUnknownSessionStderr,
 } from "./parse.js";
 import { resolveClaudeDesiredSkillNames } from "./skills.js";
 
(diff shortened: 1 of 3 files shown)


devinfoley:

PR #2383, opened 2026-04-01
fix(server): support multi-word plain-text agent mentions

### LINKED ISSUE #2360: Bug: @mention wakeup fails for multi-word agent names (e.g. @QA Engineer)
## Summary

`findMentionedAgents` in `server/src/services/issues.ts` fails to resolve `@mentions` for agents with multi-word names. The regex `/\B@([^\s@,!?.]+)/g` stops capturing at the first whitespace character, so `@QA Engineer` only extracts the token `"QA"`, which does not match the agent name `"QA Engineer"`. The agent is never woken up.

This means `wakeOnDemand` via issue comment mentions is silently broken for any agent whose name contains a space.

## Affected agent names (default gstack template)

- `@QA Engineer` → captures `"QA"` → **no match**
- `@Staff Engineer` → captures `"Staff"` → **no match**
- `@Release Engineer` → captures `"Release"` → **no match**
- `@Head of Design` → captures `"Head"` → **no match**
- `@Technical Writer` → captures `"Technical"` → **no match**

Single-word names (`@CEO`, `@CTO`) are unaffected.

## Reproduction

1. Create a company with the gstack template (agents include "QA Engineer", "Staff Engineer", etc.)
2. Assign an issue to any agent, ensure `wakeOnDemand: true` in heartbeat config
3. Add a comment on that issue containing `@QA Engineer` from another agent
4. Observe that `findMentionedAgents` returns an empty array — no wakeup is enqueued
5. The QA Engineer only picks up the issue on its next scheduled heartbeat timer tick (up to 1 hour later)

## Root cause

**File:** `server/src/services/issues.ts` — `findMentionedAgents` (line ~1638)

```typescript
const re = /\B@([^\s@,!?.]+)/g;
```

The character class `[^\s@,!?.]+` excludes whitespace, so the capture group stops at the first space. The token `"QA"` is then compared against `agent.name.toLowerCase()` (`"qa engineer"`) on line ~1652 and fails to match.

The structured mention path (`extractAgentMentionIds`) handles `[@Name](agent://id)` links correctly, but agents writing plain-text `@Name` mentions (which is the common case) hit the broken regex path.

## Suggested fix

After the single-token matching loop, add a second pass that checks whether `@<full agent name>` appears as a substring in the comment body for any multi-word agent names, with word-boundary validation:

```diff
       const resolved = new Set<string>(explicitAgentMentionIds);
+      const bodyLower = body.toLowerCase();
       for (const agent of rows) {
-        if (tokens.has(agent.name.toLowerCase())) {
+        const nameLower = agent.name.toLowerCase();
+        if (tokens.has(nameLower)) {
           resolved.add(agent.id);
+          continue;
+        }
+        if (nameLower.includes(" ")) {
+          const mentionPattern = "@" + nameLower;
+          let idx = bodyLower.indexOf(mentionPattern);
+          while (idx !== -1) {
+            const afterIdx = idx + mentionPattern.length;
+            const before = idx > 0 ? body[idx - 1] : " ";
+            const after = afterIdx < body.length ? body[afterIdx] : " ";
+            if (/[\s\n]|^$/.test(before) && /[\s\n,!?.;:\-—)}\]|]|^$/.test(after)) {
+              resolved.add(agent.id);
+              break;
+            }
+            idx = bodyLower.indexOf(mentionPattern, idx + 1);
+          }
         }
       }
```

This preserves the existing single-word matching behavior and adds case-insensitive substring matching for multi-word names with boundary checks to avoid false positives.

## Environment

- Paperclip v0.3.1 (Docker quickstart deployment)
- gstack company template (`paperclipai/companies/gstack`)
- All agents using `claude_local` adapter with `wakeOnDemand: true`

MindSyncHub:
## Thinking Path

- Paperclip uses issues and comments as the core communication surface between humans and agents
- A key part of that flow is waking agents when they are mentioned in issue comments
- Many default or realistic agent names are multi-word titles like `QA Engineer` or `Technical Writer`
- But the current plain-text mention parsing only captures a single token after `@`
- That means comments like `@QA Engineer` only resolve `QA`, so the intended agent is never matched or woken
- This PR adds multi-word plain-text mention matching while preserving the existing single-word behavior and structured mention support
- The result is that issue comment mentions work for the multi-word agent names Paperclip already encourages people to use

## What Changed

- Added a plain-text multi-word mention fallback in `findMentionedAgents`
- Kept existing token-based handling for single-word mentions
- Added regression coverage for a mixed comment containing both `@QA Engineer` and `@CEO`

## Why This Matters

Without this fix, `wakeOnDemand` via issue comments is silently broken for agents whose names contain spaces.

That affects common role-based names such as:

- `QA Engineer`
- `Staff Engineer`
- `Release Engineer`
- `Technical Writer`

## Verification

Ran:

- `pnpm exec vitest run server/src/__tests__/issues-service.test.ts server/src/__tests__/normalize-agent-mention-token.test.ts`

## Notes

This change is intentionally narrow:
- existing single-word mention parsing is preserved
- structured `[@Name](agent://...)` mentions still work as before
- multi-word fallback only applies when an agent name actually contains spaces

Closes #2360

### DIFF
--- a/server/src/__tests__/issues-service.test.ts
+++ b/server/src/__tests__/issues-service.test.ts
@@ -402,6 +402,51 @@ describeEmbeddedPostgres("issueService.list participantAgentId", () => {
       resurfacedIssueId,
     ]));
   });
+
+  it("finds plain-text mentions for agents with multi-word names", async () => {
+    const companyId = randomUUID();
+    const qaEngineerId = randomUUID();
+    const singleWordAgentId = randomUUID();
+
+    await db.insert(companies).values({
+      id: companyId,
+      name: "Paperclip",
+      issuePrefix: `T${companyId.replace(/-/g, "").slice(0, 6).toUpperCase()}`,
+      requireBoardApprovalForNewAgents: false,
+    });
+
+    await db.insert(agents).values([
+      {
+        id: qaEngineerId,
+        companyId,
+        name: "QA Engineer",
+        role: "qa",
+        status: "active",
+        adapterType: "codex_local",
+        adapterConfig: {},
+        runtimeConfig: {},
+        permissions: {},
+      },
+      {
+        id: singleWordAgentId,
+        companyId,
+        name: "CEO",
+        role: "ceo",
+        status: "active",
+        adapterType: "codex_local",
+        adapterConfig: {},
+        runtimeConfig: {},
+        permissions: {},
+      },
+    ]);
+
+    const mentionedIds = await svc.findMentionedAgents(
+      companyId,
+      "Please sync with @QA Engineer and @CEO before shipping.",
+    );
+
+    expect(new Set(mentionedIds)).toEqual(new Set([qaEngineerId, singleWordAgentId]));
+  });
 });
 
 describeEmbeddedPostgres("issueService.create workspace inheritance", () => {
--- a/server/src/services/issues.ts
+++ b/server/src/services/issues.ts
@@ -343,6 +343,25 @@ export function normalizeAgentMentionToken(raw: string): string {
   return s.trim();
 }
 
+function containsPlainTextAgentMention(bodyLower: string, agentNameLower: string): boolean {
+  const needle = `@${agentNameLower}`;
+  let index = bodyLower.indexOf(needle);
+
+  while (index !== -1) {
+    const before = index > 0 ? bodyLower[index - 1] : "";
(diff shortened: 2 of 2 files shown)


MindSyncHub:
Superseded by a rebuilt clean branch/PR based directly on current upstream master. Reopening as a fresh PR to keep the diff and review history clean.

PR #2424, opened 2026-04-01
fix: recover detached runs and refresh run details

MindSyncHub:
## Thinking Path
Detached-run recovery and run-detail refresh were already in place on this branch, but the cancel path still had one edge case: if the recorded child PID exists and `SIGTERM` returns `EPERM`, the cancel flow can abort before the run status is written back. The safest fix is to treat `EPERM` like a non-fatal failed kill attempt and cover that behavior with a regression test.

## What Changed
- Updated `killRecordedProcess` to return `false` for `EPERM` as well as `ESRCH`.
- Added a regression test that verifies detached-run cancellation still succeeds when `process.kill(pid, "SIGTERM")` throws `EPERM`.

## Why It Matters
This prevents canceling a detached local run from leaving the run stuck in `running` when the process exists but cannot be signaled by the current UID.

## Benefits
- Makes detached-run cancellation more resilient.
- Aligns the kill helper with the existing `isProcessAlive` handling for `EPERM`.
- Locks the behavior down with a targeted test.

## How To Verify
- `pnpm exec vitest run server/src/__tests__/heartbeat-process-recovery.test.ts ui/src/context/LiveUpdatesProvider.test.ts ui/src/pages/AgentDetail.test.ts`

## Risks
Treating `EPERM` as a non-fatal failed kill means the process may remain alive if it is not signalable by the current user, but the control-plane state still advances cleanly instead of leaving the run stuck in `running`.

### DIFF
--- a/server/src/__tests__/heartbeat-process-recovery.test.ts
+++ b/server/src/__tests__/heartbeat-process-recovery.test.ts
@@ -1,7 +1,7 @@
 import { randomUUID } from "node:crypto";
 import { spawn, type ChildProcess } from "node:child_process";
 import { eq } from "drizzle-orm";
-import { afterAll, afterEach, beforeAll, describe, expect, it } from "vitest";
+import { afterAll, afterEach, beforeAll, describe, expect, it, vi } from "vitest";
 import {
   agents,
   agentWakeupRequests,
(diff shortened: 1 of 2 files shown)


MindSyncHub:
Follow-up pushed in `4442246c`.

I addressed the remaining detached-run edge case by:
- treating `EPERM` as a non-fatal failure when sending `SIGTERM` / `SIGKILL` to a recorded detached PID
- adding a regression test that confirms `cancelRun` still marks the run cancelled when `SIGTERM` returns `EPERM`

Verification:
- `pnpm exec vitest run server/src/__tests__/heartbeat-process-recovery.test.ts ui/src/context/LiveUpdatesProvider.test.ts ui/src/pages/AgentDetail.test.ts`

This push is on top of the current PR head branch and should retrigger checks.

PR #2441, opened 2026-04-01
feat(company-skills): implement skill deletion (UI) with agent usage check 

DanielSousa:
## Thinking Path

> - Companies can install shared skills and attach them to agents through the skills library and agent skill sync flow
> - Removing a company skill is a destructive action because agents may still depend on that skill being available
> - The existing delete path removed the skill and silently detached it from agents, and the UI did not clearly enforce that dependency
> - This pull request makes skill removal safer by blocking deletion while any agent still uses the skill and exposing that state clearly in the UI
> - It also adds a proper delete flow in the skills screen, keeps the sidebar in sync after deletion, and avoids deleting the wrong skill if selection changes while the dialog is open
> - The benefit is a safer and more predictable skills management experience that preserves agent configuration integrity and reduces accidental destructive actions

## What Changed

- Blocked company skill deletion in `server/src/services/company-skills.ts` when the skill is still referenced by one or more agents, returning a clear error that names the blocking agents.
- Added a focused route test in `server/src/__tests__/company-skills-routes.test.ts` to verify the blocked-delete response.
- Added `companySkillsApi.delete(...)` in `ui/src/api/companySkills.ts`.
- Added a remove flow in `ui/src/pages/CompanySkills.tsx` with a confirmation dialog.
- Captured the delete target when the dialog opens so changing skill selection cannot cause the wrong skill to be deleted.
- Refetched the active skills list after successful deletion so the sidebar updates immediately.
- Updated the dialog UX so blocked removals show agent usage guidance and a single `Close` action instead of removable controls.

## Verification

- Ran `pnpm vitest run server/src/__tests__/company-skills-routes.test.ts`
- Ran `pnpm -r typecheck`
- Ran `pnpm build`
- Ran `pnpm test:run`, but the full suite still has unrelated pre-existing timeout failures outside this change (for example several `cursor-local`, `gemini-local`, and `pi-local` tests)
- Manual UI verification:
  - Open a skill used by at least one agent and confirm the dialog explains why removal is blocked and only offers `Close`
  - Open an unused skill and confirm the dialog offers `Cancel` and `Remove skill`
  - Delete an unused skill and confirm the sidebar refreshes immediately
  - Open the delete dialog, navigate to another skill, and confirm deletion still targets the originally selected skill

## Risks

- Low risk overall: the backend now changes delete behavior from auto-detach to hard-block while in use, which is an intentional behavioral shift.
- The UI depends on `usedByAgents` from skill detail for disabled/blocked states, so any future contract drift there would affect the dialog state.
- Full repo test suite is not green due to unrelated existing failures, so merge confidence relies on the targeted test, typecheck, build, and manual validation above.

## Checklist

Screenshot 2026-04-01 at 17 09 45
Screenshot 2026-04-01 at 17 11 02
Screenshot 2026-04-01 at 17 10 05

### DIFF
--- a/server/src/__tests__/company-skills-routes.test.ts
+++ b/server/src/__tests__/company-skills-routes.test.ts
@@ -3,6 +3,7 @@ import request from "supertest";
 import { beforeEach, describe, expect, it, vi } from "vitest";
 import { companySkillRoutes } from "../routes/company-skills.js";
 import { errorHandler } from "../middleware/index.js";
+import { unprocessable } from "../errors.js";
 
 const mockAgentService = vi.hoisted(() => ({
   getById: vi.fn(),
@@ -15,6 +16,7 @@ const mockAccessService = vi.hoisted(() => ({
 
 const mockCompanySkillService = vi.hoisted(() => ({
   importFromSource: vi.fn(),
+  deleteSkill: vi.fn(),
 }));
 
 const mockLogActivity = vi.hoisted(() => vi.fn());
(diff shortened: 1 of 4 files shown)


DanielSousa:
@frigiddesert 

1. Error UX
Yes, users can see it. Check this screenshot
Screenshot 2026-04-01 at 17 10 05

2. API contract change
Not documented, where should be added?

PR #2442, opened 2026-04-01
fix: prevent blank screen when clearing Capabilities field

sparkeros:
## Thinking Path

> - Humans oversee agents through the board UI, including configuring agent identity and capabilities
> - The agent Configuration tab has a Capabilities field powered by a MarkdownEditor (MDXEditor)
> - But clearing the Capabilities field completely causes a blank screen crash
> - The `onChange` handler converts empty strings to `null` via `(v || null)`, and the overlay system returns that `null` to MDXEditor which expects a `string`
> - This PR adds a null-coalescing fallback (`?? ""`) so MDXEditor always receives a string
> - The benefit is that users can clear the Capabilities field without crashing the page

## What changed

One-line fix in `ui/src/components/AgentConfigForm.tsx` (line 518):

```diff
- value={eff("identity", "capabilities", props.agent.capabilities ?? "")}
+ value={eff("identity", "capabilities", props.agent.capabilities ?? "") ?? ""}
```

## Why

The `eff()` function reads from a dirty-overlay map. When the user clears the field, `onChange` stores `null` (via `"" || null`). On re-render, `eff()` finds `null` in the overlay and returns it — but `MDXEditor` requires a `string` for its `markdown` prop. This causes an unhandled React error and a blank screen.

## Before / After

**Before:** Deleting the last character in the Capabilities field crashes the page to a blank white screen.

**After:** Clearing the field works normally — the placeholder text appears and the page remains functional.

## How to verify

1. Navigate to any agent → Configuration tab
2. Click into the Capabilities field
3. Delete all text until the field is empty
4. Confirm the page does not crash and the placeholder "Describe what this agent can do..." appears

## Risks

Minimal — single null-coalescing operator added at the call site. No changes to shared logic.

### DIFF
--- a/ui/src/components/AgentConfigForm.tsx
+++ b/ui/src/components/AgentConfigForm.tsx
@@ -515,7 +515,7 @@ export function AgentConfigForm(props: AgentConfigFormProps) {
             </Field>
             <Field label="Capabilities" hint={help.capabilities}>
               <MarkdownEditor
-                value={eff("identity", "capabilities", props.agent.capabilities ?? "")}
+                value={eff("identity", "capabilities", props.agent.capabilities ?? "") ?? ""}
                 onChange={(v) => mark("identity", "capabilities", v || null)}
                 placeholder="Describe what this agent can do..."
                 contentClassName="min-h-[44px] text-sm font-mono"


rkt2SparkEros-hotmail:
Screenshot 2026-04-01 122144
Screenshot 2026-04-01 121747

When I had one character "C" in the "Capabilities" text box everything was fine. when I would delete the "C" the screen would go blank 

Here is the statement by Claude Code
Here's the bug:

[AgentConfigForm.tsx:518-519](vscode-webview://0pehc9utj2qf1cahtl9uebs7df8ldgattu8grpgskpk1ctdhghna/ui/src/components/AgentConfigForm.tsx#L518-L519) — When you delete the last character:

onChange fires with "" → "" || null → stores null in the overlay
On re-render, eff("identity", "capabilities", props.agent.capabilities ?? "") finds capabilities in the overlay and returns null
MarkdownEditor receives value={null} but MDXEditor expects a string → crash → blank screen
The fix is straightforward — ensure the value passed to MarkdownEditor is always a string:

Once the bug was fixed (the second image) the text box contained "Describe what this agent can do ..." and is working as designed

PR #2456, opened 2026-04-01
fix: include absolute issue attachment URLs for agents

hijak:
## Thinking Path
- Paperclip orchestrates AI agents across different runtimes, including local agents and gateway-backed agents.
- Agents often work from issue context, and issue comments can include screenshots that are important for debugging or UI tasks.
- Those screenshots are stored as Paperclip attachments and exposed through the attachments API.
- But local agents like Shipwright were not getting a reliably fetchable attachment URL from issue attachment responses.
- The existing response only exposed a relative `contentPath`, which is awkward for agent workflows and model/tool chains that expect an absolute URL.
- At the same time, the OpenClaw gateway wake text now points agents at per-agent key files, so the path hint should be safe even if an agent name contains odd characters.
- This PR adds an absolute `contentUrl` alongside the existing `contentPath`, respects forwarded protocol headers for proxied deployments, and sanitises the agent name used in the per-agent key path hint.
- That keeps the change small, preserves backwards compatibility, and makes attachment links more usable for agent workflows.

## Summary
- include an absolute `contentUrl` alongside relative `contentPath` in issue attachment API responses
- derive `contentUrl` using forwarded protocol when available so proxied HTTPS deployments do not emit `http://` links
- sanitise the agent name used in OpenClaw gateway wake-text key-file path hints
- apply the attachment URL behaviour to both list and upload attachment responses

## Why
Local agents like Shipwright can struggle to use screenshots attached to tickets when they only receive relative attachment routes like `/api/attachments/<id>/content`. Emitting an absolute URL gives agent workflows a fetchable link without removing the existing relative path field.

The small OpenClaw gateway hardening change also prevents odd agent names from producing misleading per-agent key file path hints in wake text.

## Validation

## Risks
- `contentUrl` depends on request host/protocol, so bad proxy headers would still produce bad absolute URLs; this PR reduces that risk by preferring `x-forwarded-proto` when present.
- Existing consumers of `contentPath` are unaffected because that field is preserved.

### DIFF
--- a/packages/adapters/openclaw-gateway/src/server/execute.ts
+++ b/packages/adapters/openclaw-gateway/src/server/execute.ts
@@ -340,7 +340,10 @@ function buildPaperclipEnvForWake(ctx: AdapterExecutionContext, wakePayload: Wak
 }
 
 function buildWakeText(payload: WakePayload, paperclipEnv: Record<string, string>): string {
-  const agentName = (paperclipEnv.PAPERCLIP_AGENT_NAME ?? "agent").trim().toLowerCase();
+  const agentName = (paperclipEnv.PAPERCLIP_AGENT_NAME ?? "agent")
+    .trim()
+    .toLowerCase()
+    .replace(/[^a-z0-9_-]/g, "-");
   const legacyClaimedApiKeyPath = "~/.openclaw/workspace/paperclip-claimed-api-key.json";
   const perAgentApiKeyPath = `~/.openclaw/workspace/paperclip-agent-keys/${agentName}.json`;
   const orderedKeys = [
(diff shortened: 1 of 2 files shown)


hijak:
Addressed the review suggestions:

- contentUrl now prefers x-forwarded-proto before falling back to req.protocol, so proxied HTTPS deployments do not emit the wrong scheme
- the OpenClaw gateway wake-text per-agent key path hint now sanitises agent names before building the filename
- the PR body now includes the required Thinking Path plus rationale, validation, and risk notes from CONTRIBUTING.md

Also pushed the follow-up commit onto this branch.

PR #2483, opened 2026-04-02
fix: pass pluginDbId through registerPluginTools to registry

a13xperi:
## Summary
- Passes `pluginId` (database ID) as a third argument from `plugin-loader.ts` to `registerPluginTools()`
- Adds `pluginDbId?: string` parameter to the `registerPluginTools` interface and implementation in `plugin-tool-dispatcher.ts`
- Forwards `pluginDbId` to `registry.registerPlugin()` so database-backed plugins maintain correct dispatch context

Without this fix, plugins registered via the loader lose their database ID when tools are dispatched, causing issues with DB-backed plugin state.

## Test plan
- [x] Verified locally with Atlas API and Notion Sync plugins on Railway instance
- [x] `pnpm build` passes with zero errors

### DIFF
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -264,6 +264,22 @@ importers:
         specifier: ^5.7.3
         version: 5.9.3
 
+  packages/plugins/examples/plugin-atlas-api:
+    dependencies:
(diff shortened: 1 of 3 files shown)


devinfoley:

PR #2494, opened 2026-04-02
fix(skill): use skill-relative reference docs

MindSyncHub:
## Summary
- fix paperclip skill reference docs to use skill-relative paths
- fix the same reference pattern in paperclip-create-agent
- avoid file lookup failures when the skill is executed with the skill directory as cwd

## Problem
The skill docs referenced files like `skills/paperclip/references/...`, but in actual execution the cwd can already be the skill directory itself. In that case the runtime resolves a non-existent nested path and fails with a file-not-found error.

## Scope
This change only updates markdown references inside skill docs. There is no runtime or API behavior change beyond making the referenced docs resolvable from the expected skill working directory.

## Testing
- verified the referenced files exist under each skill
- verified the changed paths are now relative to the skill directory
- confirmed the compare view only includes the two affected skill docs

### DIFF
--- a/skills/paperclip-create-agent/SKILL.md
+++ b/skills/paperclip-create-agent/SKILL.md
@@ -139,4 +139,4 @@ Before sending a hire request:
 - If board requests revision, update payload and resubmit through approval flow.
 
 For endpoint payload shapes and full examples, read:
-`skills/paperclip-create-agent/references/api-reference.md`
+`references/api-reference.md`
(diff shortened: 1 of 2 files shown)


MindSyncHub:
@cryppadotta @devinfoley this is a small doc-path fix for the Paperclip skills. The PR only updates skill-local reference paths so the referenced docs resolve correctly when the skill directory is the cwd.

MindSyncHub:

PR #2373, opened 2026-04-01
fix: prevent skills pages from hanging during load

elmanGSB:
## Thinking Path

> - Paperclip serves company and agent skill metadata over HTTP and materializes runtime skill files on disk
> - Large skill sets and slow I/O could block indefinitely: directory walks, GitHub fetches, and per-skill file writes had no deadlines
> - The agent skills tab and company skill UI could hang with no user-visible failure mode
> - We need bounded time for risky operations and clear HTTP errors when limits are exceeded
> - This pull request adds timeouts to walks/fetches, TTL refresh for bundled inventory, caps runtime materialization, wraps list/sync API paths, and improves UI fetch timeout + retry
> - The benefit is skills surfaces remain responsive and surface errors instead of spinning forever

## What Changed

- Harden local skill file walks and GitHub fetches with timeouts and safer directory traversal
- TTL-cache bundled skill inventory refresh (30s)
- Cap runtime skill materialization with a 30s timeout (with follow-up for cooperative cancel noted in code where applicable)
- Wrap agent `listSkills` / `syncSkills` with timeouts; return 504 on failure
- UI: default 60s fetch timeout; error + retry on company skill detail and agent skills tab
- Runtime materialization cache: `.materialized` marker + mtime vs `updatedAt`; missing `updatedAt` uses `0` so marker comparison remains valid
- `withTimeout` in `agents` route clears timers when the main promise settles first

## Verification

- Open company Skills page and agent Skills tab; confirm load completes or shows a clear error with retry
- `npx tsc --noEmit` in `server` and `ui` as applicable
- **UI:** before/after screenshots recommended for the skills error/retry states (attach in PR thread if not embedded here)

## Risks

- Timeouts may surface as errors for very slow but eventually successful operations; thresholds (30s/60s) may need tuning per deployment
- `Promise.race` timeouts do not cancel in-flight FS work without `AbortSignal` (noted for follow-up)

## Model Used

Mixed human and AI-assisted edits; includes Claude Code and Cursor (exact model IDs not recorded for all commits).

## Checklist

- [x] I have included a thinking path that traces from project context to this change
- [x] I have specified the model used (with version and capability details)
- [ ] I have run tests locally and they pass
- [ ] I have added or updated tests where applicable
- [ ] If this change affects the UI, I have included before/after screenshots
- [x] I have updated relevant documentation to reflect my changes
- [x] I have considered and documented any risks above
- [x] I will address all Greptile and reviewer comments before requesting merge

PR #2508, opened 2026-04-02
fix: cache materialised GitHub skills to avoid re-fetch on every heartbeat run

### LINKED ISSUE #2507: Bug: GitHub-sourced skills re-fetched from remote on every heartbeat run, blocking agent execution
## Bug

When a company has GitHub-sourced skills installed, `listRuntimeSkillEntries()` calls `materializeRuntimeSkillFiles()` for **every** GitHub skill on **every** heartbeat run. Each materialisation calls `readFile()` per inventory entry, which fetches file content from `raw.githubusercontent.com` individually and sequentially. There is no disk caching — `normalizeSkillDirectory()` returns `null` for non-`local_path` skills, and `materializeRuntimeSkillFiles()` unconditionally deletes and recreates the skill directory before re-fetching.

With ~60 GitHub-sourced skills (each with 1-3 files), this adds **3-5 minutes of blocking I/O** to the `executeRun` setup phase before the adapter can even spawn a process. During this time, the run shows `status: running` with `processPid: null` and no log output, making it appear permanently stuck. The orphan reaper (5-minute staleness threshold) often kills these runs before materialisation completes.

## Impact

- **Severity**: Critical — all agents in a company are blocked from executing when GitHub skills are installed
- **Affected function**: `listRuntimeSkillEntries()` in `server/src/services/company-skills.ts`
- **Call chain**: `executeRun()` → `listRuntimeSkillEntries()` → `materializeRuntimeSkillFiles()` → `readFile()` → HTTP fetch per file
- **Trigger**: Installing any significant number of GitHub-sourced skills via the Skills Manager or `skills.sh`

## Root Cause

In `listRuntimeSkillEntries()` (~line 2047):

```ts
let source = normalizeSkillDirectory(skill);
if (!source) {
  source = options.materializeMissing === false
    ? resolveRuntimeSkillMaterializedPath(companyId, skill)
    : await materializeRuntimeSkillFiles(companyId, skill).catch(() => null);
}
```

`normalizeSkillDirectory()` only resolves `local_path` and `catalog` source types — GitHub skills always fall through to `materializeRuntimeSkillFiles()`.

`materializeRuntimeSkillFiles()` (~line 2022) unconditionally deletes and recreates:

```ts
await fs.rm(skillDir, { recursive: true, force: true });
await fs.mkdir(skillDir, { recursive: true });
for (const entry of skill.fileInventory) {
  const detail = await readFile(companyId, skill.id, entry.path).catch(() => null);
  // ... writes to disk
}
```

`readFile()` for GitHub skills fetches from `raw.githubusercontent.com` on every call — no local cache, no batch fetching.

### Short-term (cache hit)

Before calling `materializeRuntimeSkillFiles`, check if the `__runtime__` directory already exists on disk:

```ts
if (!source) {
  const cachedPath = resolveRuntimeSkillMaterializedPath(companyId, skill);
  const cachedStat = await fs.stat(cachedPath).catch(() => null);
  if (cachedStat?.isDirectory()) {
    source = cachedPath;
  } else if (options.materializeMissing === false) {
    source = cachedPath;
  } else {
    source = await materializeRuntimeSkillFiles(companyId, skill).catch(() => null);
  }
}
```

This ensures GitHub skills are fetched once and reused from disk on subsequent runs. Cache invalidation happens naturally when skills are updated/re-imported (which already deletes the runtime directory).

### Long-term improvements

1. **Batch GitHub file fetching**: Use the GitHub Trees API (`GET /repos/{owner}/{repo}/git/trees/{sha}?recursive=1`) to fetch all files in one request instead of one HTTP request per file.
2. **Store file content in DB at import time**: The `markdown` column already stores `SKILL.md` content. Extend this to store all inventory file contents at import, eliminating runtime HTTP fetches entirely.
3. **Parallel materialisation**: Process skills concurrently with `Promise.all` (or a bounded concurrency pool) instead of sequential `for...of`.
4. **Background pre-materialisation**: Materialise skills asynchronously after import/update rather than lazily on first heartbeat run.
5. **ETag/If-None-Match caching**: For incremental updates, use GitHub API conditional requests to avoid re-downloading unchanged files.

## Reproduction

1. Install 50+ skills from GitHub repos (e.g. via `skills.sh` or the Skills Manager UI)
2. Trigger any agent heartbeat run
3. Observe run stuck in `running` with `processPid: null` for 3-5+ minutes
4. Run may be killed by `reapOrphanedRuns` before materialisation completes

## Environment

- Paperclip version: 0.3.1 (commit `6a72faf8`)
- Adapter: any (issue is pre-adapter in `executeRun` setup)
- OS: macOS (darwin), but affects all platforms

helpingwithtech:
## Summary

- Cache materialised GitHub skill directories on disk so they're only fetched once from `raw.githubusercontent.com`, not on every heartbeat run
- Subsequent runs use the existing `__runtime__` directory via `fs.stat` check, bypassing the expensive `materializeRuntimeSkillFiles` → `readFile` → HTTP fetch chain
- Cache is naturally invalidated when skills are updated/re-imported (which deletes the runtime directory)

Closes #2507

## Problem

`listRuntimeSkillEntries()` called `materializeRuntimeSkillFiles()` for every GitHub-sourced skill on every heartbeat run. Each materialisation fetched files individually from GitHub over HTTP. With 60+ GitHub skills, this added 3-5 minutes of blocking I/O to the `executeRun` setup phase — before the adapter could even spawn a process. Runs appeared permanently stuck (`processPid: null`, no logs) and were often killed by the orphan reaper.

## Change

In `listRuntimeSkillEntries`, before calling `materializeRuntimeSkillFiles`, check if the `__runtime__` cache directory already exists:

```ts
const cachedPath = resolveRuntimeSkillMaterializedPath(companyId, skill);
const cachedStat = await fs.stat(cachedPath).catch(() => null);
if (cachedStat?.isDirectory()) {
  source = cachedPath;
} else {
  source = await materializeRuntimeSkillFiles(...);
}
```

## Test plan

- [ ] Install 50+ GitHub-sourced skills via Skills Manager
- [ ] Trigger a heartbeat run — first run materialises skills (slow, expected)
- [ ] Trigger a second run — should start within seconds (cache hit)
- [ ] Update a GitHub skill via the UI — verify the updated content is used on the next run (cache invalidated by re-import)
- [ ] Verify `pnpm -r typecheck` passes

### DIFF
--- a/server/src/services/company-skills.ts
+++ b/server/src/services/company-skills.ts
@@ -2048,9 +2048,15 @@ export function companySkillService(db: Db) {
       const sourceKind = asString(getSkillMeta(skill).sourceKind);
       let source = normalizeSkillDirectory(skill);
       if (!source) {
-        source = options.materializeMissing === false
-          ? resolveRuntimeSkillMaterializedPath(companyId, skill)
-          : await materializeRuntimeSkillFiles(companyId, skill).catch(() => null);
+        const cachedPath = resolveRuntimeSkillMaterializedPath(companyId, skill);
+        const cachedStat = await fs.stat(cachedPath).catch(() => null);
+        if (cachedStat?.isDirectory()) {
+          source = cachedPath;
+        } else if (options.materializeMissing === false) {
+          source = cachedPath;
+        } else {
+          source = await materializeRuntimeSkillFiles(companyId, skill).catch(() => null);
+        }
       }
       if (!source) continue;
 


quangthien27:
Related to #2373

PR #2537, opened 2026-04-02
fix(issues): clear stale execution locks from dead heartbeat runs

### LINKED ISSUE #2533: Stale `execution_run_id` locks block all issue mutations after agent run crashes
### Summary
When an agent run crashes or times out, it does not release the `execution_run_id` lock on the issue. This leaves the issue permanently locked — no agent can checkout, comment, patch, or release it. The only fix is a manual DB update (`UPDATE issues SET execution_run_id = NULL`).

### Impact
This has caused repeated workflow stalls in our company. We've had to manually clear stale locks via direct DB access on multiple occasions:
- VOIA-65, VOIA-63, VOIA-64, VOIA-73 — 4 issues locked simultaneously, blocking the entire widget sub-task pipeline
- VOIA-59 — parent issue locked for hours, preventing orchestrator from acting
- VOIA-90 — locked in `todo` status
- VOIA-74 — re-locked after being cleared once

Each time, 0 agents could run and the entire company stalled until we manually ran:
```sql
UPDATE issues SET execution_run_id = NULL, execution_locked_at = NULL 
WHERE id = '<issue_id>';
```

### Root Cause
The `execution_run_id` field is set when an agent checks out an issue for execution, but is never cleared if the agent process crashes, times out, or the adapter fails. The `/issues/:id/release` endpoint clears `checkout_run_id` and `assignee_agent_id` but does NOT clear `execution_run_id`.

### Expected Behavior
- If an agent run finishes (success or failure), `execution_run_id` should be cleared
- If an agent run is orphaned (no heartbeat for N minutes), the server should auto-clear the lock
- The `/issues/:id/release` endpoint should also clear `execution_run_id` and `execution_locked_at`

### Suggested Fix
1. **Auto-expiry**: On heartbeat tick, check for issues where `execution_locked_at` is older than a configurable threshold (e.g., 30 minutes) and the associated run is no longer active. Clear the lock automatically.
2. **Release endpoint**: Update the `release` service method to also set `execution_run_id = NULL` and `execution_locked_at = NULL`.
3. **Run finalization**: In the heartbeat run finalization logic, ensure `execution_run_id` is always cleared on the issue when a run transitions to a terminal state (completed, failed, error).

### Environment
- Paperclip version: 0.3.1
- Deployment: local_trusted, embedded PostgreSQL
- Adapter types affected: `opencode_local`, `claude_local`

### Reproduction
1. Create a company with agents
2. Create an issue and assign to an agent
3. Agent starts a run (execution_run_id gets set)
4. Kill the agent process mid-run (e.g., SIGHUP, adapter crash)
5. Try to checkout/comment/patch the issue → "Issue run ownership conflict" 409 error
6. Issue is permanently locked until manual DB intervention

plind-junior:
When an agent process dies mid-run (SIGHUP, OOM, adapter crash), the issue's executionRunId remains set while the backing heartbeat run is terminal or missing. Subsequent checkout and assertCheckoutOwner calls hit an unrecoverable 409 because no code path attempted to detect and release the orphaned lock — the only cleanup lived in the heartbeat scheduler, which might not fire in time or at all.

Add clearStaleExecutionLock() that checks whether the blocking run is terminal/missing via the existing isTerminalOrMissingHeartbeatRun helper, and if so atomically nullifies executionRunId, executionAgentNameKey, and executionLockedAt with a CAS guard on the expected run id.

Wire it into both checkout (before the final conflict throw) and assertCheckoutOwner (same position), so the next agent heartbeat self-heals rather than requiring manual DB intervention.

Fixes #2533

## Thinking Path

> - Agents check out issues for execution, setting `executionRunId` as a lock
> - If an agent process crashes (SIGHUP, OOM, adapter crash), the lock is never released
> - The issue becomes permanently stuck — no agent can checkout, comment, patch, or release it
> - The only fix was manual DB intervention (`UPDATE issues SET execution_run_id = NULL`)
> - This caused repeated workflow stalls in production (VOIA-65, VOIA-63, VOIA-64, VOIA-73, etc.)
> - This PR adds `clearStaleExecutionLock()` that detects dead runs and atomically clears the lock
> - It's wired into both `checkout` and `assertCheckoutOwner` so the next heartbeat self-heals
> - The benefit is automatic recovery from stale locks without manual DB intervention

## What Changed

- Added `clearStaleExecutionLock()` function that checks `isTerminalOrMissingHeartbeatRun` and atomically clears `executionRunId`, `executionAgentNameKey`, and `executionLockedAt` with a CAS guard (`server/src/services/issues.ts`)
- Wired stale lock detection into `checkout` — before throwing 409, checks if the blocking run is dead and retries with proper `assigneeAgentId` ownership guard
- Wired stale lock detection into `assertCheckoutOwner` — same pattern, with status and assignee pre-checks per reviewer feedback
- Added `executionRunId` to the select in `assertCheckoutOwner` for the stale lock check

## Verification

- `pnpm vitest run server/src/__tests__/issue*.test.ts server/src/__tests__/checkout*.test.ts`
- Simulate: create an issue, checkout with agent, kill the agent process, attempt checkout with another agent — should self-heal instead of 409

## Risks

Medium risk — modifies the checkout and run ownership critical path. Mitigated by:
- CAS guard on `executionRunId` prevents race conditions
- Only clears locks when `isTerminalOrMissingHeartbeatRun` confirms the run is dead
- `assigneeAgentId` ownership check prevents cross-agent issue stealing (per reviewer feedback)
- Proactive cleanup (heartbeat sweep) and `/release` endpoint updates are deferred to follow-up PRs

## Model Used

## Checklist


imazizongit (commented):
## Review

Thanks for the quick fix — this addresses a real pain point we've been hitting repeatedly in production. The core approach is solid.

### What looks good
- `clearStaleExecutionLock()` is well-designed — atomic update with `WHERE eq(issues.executionRunId, expectedExecutionRunId)` prevents race conditions
- Checking `isTerminalOrMissingHeartbeatRun` before clearing is the right call — doesn't blindly wipe active locks
- Applied in both checkout AND run adoption paths (the two places we were getting 409s)

### Bug: Duplicate `if` block

There's a copy-paste error in the adoption section (~line 1430-1440). The `if` statement is duplicated:

```typescript
if (
    actorRunId &&
if (
    actorRunId &&
    current.status === "in_progress" &&
```

The first `if` is incomplete — this won't compile. Needs to be collapsed into a single `if` block.

### Not addressed (nice-to-haves for a follow-up)

1. **No proactive cleanup** — This only clears locks reactively (when another agent tries to checkout). If no agent ever attempts to checkout the locked issue again, the stale lock persists forever. A heartbeat-tick sweep that auto-clears `execution_locked_at` older than N minutes would cover this case.

2. **`/issues/:id/release` endpoint** — The `release` service method still only clears `checkoutRunId` and `assigneeAgentId` but does NOT clear `executionRunId` / `executionLockedAt`. Board operators calling release to manually unstick an issue still can't clear execution locks without direct DB access.

### Verdict

Fix the duplicate `if` block and this is good to merge. The reactive clearing will eliminate the majority of our stale lock incidents. The proactive sweep and release endpoint updates can be follow-up PRs.

plind-junior:
@imazizongit Checked it and fixed. Would you check the PR again? ty

PR #2564, opened 2026-04-02
Skip Codex git repo check for fallback workspaces

ergonaworks:
﻿## Thinking Path

Paperclip supports Codex agents in both real project workspaces and Paperclip-managed fallback workspaces. Real project workspaces are usually Git-backed, but fallback workspaces under `.paperclip/instances/.../workspaces/...` are intentionally plain directories so the agent still has somewhere to run when no project checkout or resumable session workspace exists.

The current Codex local adapter builds a normal `codex exec` invocation for both cases. That works in Git-backed repos, but fallback workspaces are not trusted Git directories, so Codex fails early with `Not inside a trusted directory and --skip-git-repo-check was not specified.` Operators can get past that today by turning on `--dangerously-bypass-approvals-and-sandbox`, but that removes the safety rails we actually want to keep.

This PR narrows the fix to the real problem: if Paperclip is launching Codex from a directory that is not inside any Git repository, Paperclip now adds Codex's narrower `--skip-git-repo-check` flag automatically. Repo-backed runs stay unchanged, and the existing sandbox / approval behavior remains intact.

## What Changed

- added a small `isInsideGitRepo()` helper that detects both `.git/` directories and worktree-style `.git` files while walking up from the configured cwd
- extracted Codex CLI argument assembly into `buildCodexExecArgs()` so both execution and the environment hello probe share the same flag logic
- updated the Codex local adapter runtime path to add `--skip-git-repo-check` only when the cwd is outside a Git repo
- updated the Codex local adapter environment probe to mirror that same behavior
- added focused unit tests for repo detection and CLI arg assembly
- added a command note so operators can see when Paperclip auto-added the git-check skip

## Why This Matters

Without this change, agents that fall back to a Paperclip-managed workspace can fail unless the operator disables Codex's sandbox protections entirely. This keeps fallback runs working while preserving the safer default execution model.

## Verification

- `pnpm exec vitest run packages/adapters/codex-local/src/server/cli-args.test.ts packages/adapters/codex-local/src/server/git-repo.test.ts`
- `pnpm --filter @paperclipai/server exec vitest run src/__tests__/codex-local-adapter-environment.test.ts`

## Risks

- the repo detector is intentionally simple and filesystem-based; if a future Codex trust model diverges from `.git` presence, Paperclip may need to refine that heuristic
- existing Windows integration tests for `codex-local-execute` still rely on a fake executable stub that is not launchable on this machine, so this PR keeps coverage at the helper + environment-probe level rather than expanding unrelated fixture work

## Checklist

- [x] change is scoped to the Codex local adapter
- [x] fallback non-repo workspaces keep sandbox protections instead of requiring the bypass flag
- [x] repo-backed runs remain unchanged
- [x] focused tests added for the new behavior
- [x] targeted verification commands run

### DIFF
--- a/packages/adapters/codex-local/src/server/cli-args.test.ts
+++ b/packages/adapters/codex-local/src/server/cli-args.test.ts
@@ -1,0 +1,55 @@
+import { describe, expect, it } from "vitest";
+import { buildCodexExecArgs } from "./cli-args.js";
+
+describe("buildCodexExecArgs", () => {
+  it("adds --skip-git-repo-check for non-repo fallback workspaces", () => {
+    expect(
+      buildCodexExecArgs({
+        search: false,
+        bypass: false,
+        skipGitRepoCheck: true,
+        extraArgs: [],
+        model: "",
+        modelReasoningEffort: "",
+        resumeSessionId: null,
+      }),
+    ).toEqual(["exec", "--json", "--skip-git-repo-check", "-"]);
+  });
+
+  it("does not duplicate --skip-git-repo-check when already provided in extra args", () => {
+    expect(
+      buildCodexExecArgs({
+        search: false,
+        bypass: false,
+        skipGitRepoCheck: true,
+        extraArgs: ["--skip-git-repo-check", "--color", "never"],
+        model: "",
+        modelReasoningEffort: "",
+        resumeSessionId: null,
+      }),
+    ).toEqual(["exec", "--json", "--skip-git-repo-check", "--color", "never", "-"]);
+  });
+
+  it("keeps normal repo-backed runs unchanged", () => {
+    expect(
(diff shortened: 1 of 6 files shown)


ergonaworks:
**Rebased on master** — the previous branch had grown stale and had merge conflicts. Force-pushed a clean cherry-pick of the single fix commit on top of current master.

**Runtime reproduction confirmed (ERG-8):** A newly created `codex_local` agent with `dangerouslyBypassApprovalsAndSandbox: false` fails immediately with `Not inside a trusted directory and --skip-git-repo-check was not specified.` when assigned to a fallback workspace. Applied a `git init` workaround to the affected workspace as a local patch; this PR is the systemic fix.

PR #2574, opened 2026-04-02
fix: handle explicit claude success results

MindSyncHub:
## Thinking Path

> - Paperclip runs local and remote agents and treats heartbeat outcomes as the source of truth for task progress
> - Local adapter processes sometimes report success through structured result JSON rather than only through OS exit codes
> - Claude-local can emit an explicit success result even when the wrapper process exits non-zero
> - When that happens, Paperclip should trust the explicit success payload instead of marking the run as failed
> - This PR teaches both the claude-local adapter and the heartbeat outcome layer to recognize explicit successful results
> - The benefit is that valid successful runs are no longer misclassified as failures just because the wrapper exited with a non-zero code

## What Changed

This PR fixes a case where Claude emits an explicit successful result payload but the process exits with a non-zero exit code.

Changes included:
- add `isClaudeSuccessResult` in the claude-local parse layer
- suppress adapter-level error handling when the result payload explicitly reports success
- add matching heartbeat-side success detection so run outcome classification is upgraded correctly at the service layer
- add tests for adapter parsing/execution and heartbeat run outcome handling

## Why This Matters

Before this change, a successful Claude-local run could still appear failed in Paperclip if the wrapper exited with a non-zero code after already emitting a valid success result.

That created incorrect run status, misleading error reporting, and unnecessary follow-up work on runs that had actually completed successfully.

This change makes run outcome handling more faithful to the structured result data emitted by the adapter.

## How To Verify

- run the claude-local execute tests
- run the heartbeat run outcome tests
- verify the integration case where a fake Claude binary emits a success result and exits with code `1`
- confirm the adapter returns no error message in that case
- confirm the heartbeat run is classified as succeeded rather than failed

## Risks

- there is duplicated success-detection logic between the adapter parse layer and the heartbeat service; if the success-shape heuristic changes later, both places need to stay aligned
- this PR intentionally keeps the fix small and local rather than doing a broader refactor of success-result classification across adapters

Supersedes #2345. This PR was rebuilt cleanly on current upstream `master` so the diff stays focused.

### DIFF
--- a/packages/adapters/claude-local/src/server/execute.ts
+++ b/packages/adapters/claude-local/src/server/execute.ts
@@ -26,6 +26,7 @@ import {
   parseClaudeStreamJson,
   describeClaudeFailure,
   detectClaudeLoginRequired,
+  isClaudeSuccessResult,
   isClaudeMaxTurnsResult,
   isClaudeUnknownSessionError,
 } from "./parse.js";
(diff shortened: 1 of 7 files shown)


MindSyncHub:

PR #2577, opened 2026-04-02
fix(ui): normalize inline editor saves and extra args parsing

MindSyncHub:
## Thinking Path
This PR already extracted helpers for inline-save normalization and extra-args parsing, but the helper usage was still inconsistent across multiline save paths and the cleared `extraArgs` field had changed semantics from `undefined` to `[]`. The follow-up keeps every save path on the same comparison logic and restores the prior clear-on-empty behavior.

## What Changed
- Normalize both `nextValue` and `currentValue` in `shouldSaveInlineEditorValue`.
- Reuse the shared save guard in multiline autosave, blur, and submit paths.
- Restore `resolveExtraArgsValue("")` / blank input to `undefined` instead of `[]`.
- Added tests covering current-value normalization and the restored empty extra-args semantics.

## Why It Matters
This keeps inline saves consistent across code paths and avoids unintentionally changing the meaning of clearing adapter extra args.

## Benefits
- Prevents spurious resaves when stored values contain surrounding whitespace.
- Makes multiline autosave, blur, and submit follow the same logic.
- Preserves the prior adapter-config patch semantics for clearing `extraArgs`.

## How To Verify
- `pnpm exec vitest run ui/src/components/InlineEditor.test.ts ui/src/components/AgentConfigForm.test.ts`

## Risks
Clearing extra args now intentionally maps back to `undefined`, matching the previous edit-path behavior. This is a semantics-preserving change for updates, but it is different from the intermediate `[]` behavior on the branch before this follow-up.

### DIFF
--- a/ui/src/components/AgentConfigForm.test.ts
+++ b/ui/src/components/AgentConfigForm.test.ts
@@ -9,8 +9,8 @@ describe("resolveExtraArgsValue", () => {
     ]);
   });
 
-  it("returns an empty array when the field is cleared", () => {
-    expect(resolveExtraArgsValue("")).toEqual([]);
-    expect(resolveExtraArgsValue("   ")).toEqual([]);
+  it("returns undefined when the field is cleared", () => {
+    expect(resolveExtraArgsValue("")).toBeUndefined();
+    expect(resolveExtraArgsValue("   ")).toBeUndefined();
   });
 });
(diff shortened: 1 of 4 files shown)


MindSyncHub:
Follow-up pushed in `54faea06`.

Addressed the remaining review points by:
- normalizing both sides in `shouldSaveInlineEditorValue`
- reusing the shared save guard across multiline autosave / blur / submit
- restoring blank `extraArgs` to `undefined` instead of `[]`
- adding focused tests for the normalization and extra-args semantics

Verification:
- `pnpm exec vitest run ui/src/components/InlineEditor.test.ts ui/src/components/AgentConfigForm.test.ts`

This should retrigger the PR checks on the updated head.

PR #2594, opened 2026-04-03
fix(issues): replace non-null assertions with null checks in checkout…

### LINKED ISSUE #2593: Bug: Non-null assertion in checkout can crash server if issue is deleted during concurrent adoption
## Description

`issueService.checkout()` in `server/src/services/issues.ts` uses TypeScript non-null assertions (`rows[0]!`) when re-reading an issue after successful stale-run adoption. If the issue is deleted between the adoption write and the re-read, `rows[0]` is `undefined` and the subsequent `withIssueLabels(db, [undefined])` throws an unhandled error, crashing the request (and potentially the process if no error boundary catches it).

## Affected code

**File:** `server/src/services/issues.ts`

**Line 1472** — after `adoptStaleCheckoutRun` succeeds:
```ts
const row = await db.select().from(issues).where(eq(issues.id, id)).then((rows) => rows[0]!);
const [enriched] = await withIssueLabels(db, [row]);
```

**Line 1484** — self-ownership fast path:
```ts
const row = await db.select().from(issues).where(eq(issues.id, id)).then((rows) => rows[0]!);
const [enriched] = await withIssueLabels(db, [row]);
```

## Root cause

The `!` assertion is a TypeScript compile-time hint that has no runtime effect. If the issue row no longer exists (deleted by another request, company cascade, or DB cleanup), `rows[0]` is `undefined`, and the code proceeds to pass it into `withIssueLabels` which will throw `TypeError: Cannot read properties of undefined`.

## How to trigger

This is a race condition. The window is milliseconds between the CAS update (`adoptStaleCheckoutRun` or the ownership check) and the re-read query. Realistic triggers:

1. Board user deletes a company (cascade deletes issues) while an agent heartbeat is mid-checkout
2. Board user deletes an issue via API while the agent's heartbeat is adopting a stale lock on that issue
3. Any concurrent DELETE on the issue table during the checkout flow

## Why this can't be reproduced through the UI

This bug requires two database operations to interleave within the same HTTP request's execution window — roughly a few milliseconds. The UI does not expose any workflow that allows a user to:

- Delete an issue while simultaneously checking it out from another tab (the checkout is triggered by the heartbeat scheduler, not by a UI button)
- Time a company deletion to land between two queries inside the same server-side function call

The race window exists entirely within server-side code between lines 1470–1472 (or 1482–1484). There is no sequence of button clicks or page navigations that can force these two operations to overlap. The only way to reproduce it is:

- **Programmatically:** inject a delay between the adoption and re-read, then issue a DELETE during that pause
- **Under load:** run enough concurrent heartbeats and deletions that the timing collision occurs naturally
- **Via tests:** write an integration test that mocks the re-read to return empty rows

This makes it a **code-level defensive gap**, not a user-facing regression. It would surface in production at scale, particularly during company teardowns or bulk issue cleanup while agents are actively checking out work.

## Expected behavior

If the issue disappears between the adoption and re-read, the server should return a 404 instead of crashing.

## Suggested fix

Replace the non-null assertions with null checks:

```ts
const row = await db.select().from(issues).where(eq(issues.id, id)).then((rows) => rows[0] ?? null);
if (!row) throw notFound("Issue not found");
const [enriched] = await withIssueLabels(db, [row]);
```

## Impact

- **Severity:** Low (requires very specific timing)
- **Blast radius:** Crashes the HTTP request handler; could crash the process if unhandled
- **Scope:** Two code paths in `checkout`, potentially more in `assertCheckoutOwner` with similar patterns

## Environment

- Server: `server/src/services/issues.ts`
- Version: current master

plind-junior:
Two code paths in issueService.checkout() used rows[0]! when re-reading an issue after stale-run adoption or self-ownership verification. If the issue is deleted concurrently (company cascade, API delete), rows[0] is undefined and withIssueLabels crashes with an unhandled TypeError.

Replace both with rows[0] ?? null and throw notFound when the row is missing, returning a clean 404 instead of an uncaught exception.

Fixes #2593

## Thinking Path

> - Agents check out issues via `issueService.checkout()` which involves multiple DB queries
> - After adopting a stale run or verifying self-ownership, the code re-reads the issue for label enrichment
> - The re-read used TypeScript non-null assertions (`rows[0]!`) which have no runtime effect
> - If the issue is deleted between the adoption write and re-read (race condition during company cascade or API delete), `rows[0]` is `undefined`
> - `withIssueLabels(db, [undefined])` throws an unhandled TypeError, crashing the request
> - This PR replaces `rows[0]!` with `rows[0] ?? null` and throws `notFound` when the row is missing
> - The benefit is a clean 404 response instead of an uncaught exception in a narrow but real race window

## What Changed

- Replaced two `rows[0]!` non-null assertions with `rows[0] ?? null` followed by explicit `notFound` throws in `server/src/services/issues.ts`
- Both the stale-run adoption path (line ~1472) and the self-ownership verification path (line ~1484) are fixed

## Verification

- `pnpm tsc --noEmit --project server/tsconfig.json 2>&1 | grep issues.ts || echo "clean"`
- `pnpm vitest run server/src/__tests__/issue*.test.ts server/src/__tests__/checkout*.test.ts 2>&1 | tail -10`

## Risks

Low risk — the fix is additive (null coalescing + guard) and only fires in the narrow concurrent-delete race window. Normal checkout flow is unaffected. Pattern matches existing conventions in the same file.

## Model Used

## Checklist

### DIFF
--- a/server/src/services/issues.ts
+++ b/server/src/services/issues.ts
@@ -1469,7 +1469,8 @@ export function issueService(db: Db) {
           expectedCheckoutRunId: current.checkoutRunId,
         });
         if (adopted) {
-          const row = await db.select().from(issues).where(eq(issues.id, id)).then((rows) => rows[0]!);
+          const row = await db.select().from(issues).where(eq(issues.id, id)).then((rows) => rows[0] ?? null);
+          if (!row) throw notFound("Issue not found");
           const [enriched] = await withIssueLabels(db, [row]);
           return enriched;
         }
@@ -1481,7 +1482,8 @@ export function issueService(db: Db) {
         current.status === "in_progress" &&
         sameRunLock(current.checkoutRunId, checkoutRunId)
       ) {
-        const row = await db.select().from(issues).where(eq(issues.id, id)).then((rows) => rows[0]!);
+        const row = await db.select().from(issues).where(eq(issues.id, id)).then((rows) => rows[0] ?? null);
+        if (!row) throw notFound("Issue not found");
         const [enriched] = await withIssueLabels(db, [row]);
         return enriched;
       }


plind-dm:
@devinfoley @cryppadotta Requesting your review on my pull request. Thank you.

PR #2653, opened 2026-04-03
fix(api): default comment sort order to ascending (oldest first)

### LINKED ISSUE #2240: Sorting order on comments
It would be nice if we could have different sorting options on a tickets comment

plind-junior:
The comments endpoint defaulted to descending order (newest first), but the UI re-sorted to ascending for chronological reading. Change the API default to ascending so the response matches natural reading order without client-side re-sorting. Callers can still pass ?order=desc to get newest-first.

Closes #2240

## Thinking Path

> - Issues have comment threads that agents and users interact with
> - The comments API (`GET /issues/:id/comments`) defaulted to descending order (newest first)
> - The UI always re-sorted comments to ascending for chronological reading order
> - This meant every comment fetch required a client-side re-sort, and any new consumer would get an unintuitive default
> - This PR flips the default from `desc` to `asc` in both the route handler and service layer
> - Callers can still explicitly pass `?order=desc` to get newest-first when needed
> - The benefit is the API returns comments in natural reading order by default, eliminating unnecessary client-side re-sorting

## What Changed

- Inverted the default sort guard in the route handler (`server/src/routes/issues.ts`) — unrecognized or missing `?order=` now resolves to `"asc"` instead of `"desc"`
- Inverted the default in the service layer `listComments` (`server/src/services/issues.ts`) — `opts?.order === "desc"` is now the explicit case, with `"asc"` as the fallback

## Verification

- `GET /api/issues/:id/comments` — comments should return oldest-first by default
- `GET /api/issues/:id/comments?order=desc` — should still return newest-first
- `pnpm vitest run server/src/__tests__/issue*.test.ts`

## Risks

Low risk — this is a default change, not a behavioral removal. Any caller that explicitly passed `?order=desc` continues to work. Callers relying on the implicit default to be `desc` will now get `asc`, but the UI was already re-sorting to `asc` anyway. Cursor pagination logic handles both directions correctly.

## Model Used

## Checklist

### DIFF
--- a/server/src/routes/issues.ts
+++ b/server/src/routes/issues.ts
@@ -1468,9 +1468,9 @@ export function issueRoutes(db: Db, storage: StorageService) {
           ? req.query.afterCommentId.trim()
           : null;
     const order =
-      typeof req.query.order === "string" && req.query.order.trim().toLowerCase() === "asc"
-        ? "asc"
-        : "desc";
+      typeof req.query.order === "string" && req.query.order.trim().toLowerCase() === "desc"
+        ? "desc"
+        : "asc";
     const limitRaw =
       typeof req.query.limit === "string" && req.query.limit.trim().length > 0
         ? Number(req.query.limit)
--- a/server/src/services/issues.ts
+++ b/server/src/services/issues.ts
@@ -1628,7 +1628,7 @@ export function issueService(db: Db) {
         limit?: number | null;
       },
     ) => {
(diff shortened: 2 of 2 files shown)


plind-junior:
@cryppadotta @devinfoley  Would you check this PR?

plind-junior (commented):
LGTM

cryppadotta:
why do we need this

PR #2665, opened 2026-04-03
fix(server): refresh detached run recovery on latest master

MindSyncHub:
## Thinking Path
The original detached-run recovery branch drifted too far from current `master`, so this replacement PR reapplies the cancel-path hardening on a fresh base. The key issue is that detached local runs need to be cancelled through their recorded PID, and that `EPERM` on that signal path must not leave the run stuck in `running`.

Supersedes #2424.

## What Changed
- Reused detached-process termination for cancel flows so cancellation can target recorded child PIDs, not just in-memory handles.
- Treated `EPERM` like a non-fatal failed kill attempt in `killRecordedProcess`.
- Added a regression test covering the `SIGTERM -> EPERM` path during detached-run cancellation.

## Why It Matters
This keeps control-plane state moving forward even when a detached local child cannot be signaled by the current UID.

## Benefits
- Cancels detached local runs via recorded PID.
- Prevents `EPERM` from aborting the cancel flow.
- Adds focused regression coverage for the edge case.

## How To Verify
- `pnpm exec vitest run server/src/__tests__/heartbeat-process-recovery.test.ts ui/src/context/LiveUpdatesProvider.test.ts ui/src/pages/AgentDetail.test.ts`

## Risks
Treating `EPERM` as a non-fatal failed kill means the process may still be alive if another UID owns it, but the run no longer remains stuck in `running`.

### DIFF
--- a/server/src/__tests__/heartbeat-process-recovery.test.ts
+++ b/server/src/__tests__/heartbeat-process-recovery.test.ts
@@ -253,7 +253,7 @@ describeEmbeddedPostgres("heartbeat orphaned process recovery", () => {
   }, 20_000);
 
   afterEach(async () => {
-    vi.clearAllMocks();
+    vi.restoreAllMocks();
     mockAdapterExecute.mockImplementation(async () => ({
       exitCode: 0,
       signal: null,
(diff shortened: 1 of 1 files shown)


MindSyncHub:
Refreshed this PR on top of `08af8304` (current `origin/master`).

The detached-run recovery runtime fix is already present on master in the current heartbeat termination path, so this refresh re-applies the missing regression coverage for `SIGTERM` returning `EPERM` and tightens test cleanup to restore mocked implementations between cases.

Focused verification:
- `corepack pnpm vitest run server/src/__tests__/heartbeat-process-recovery.test.ts`

MindSyncHub:

PR #2683, opened 2026-04-03
fix(claude-local,codex-local): add isolateSession option to prevent session context bleed

### LINKED ISSUE #2682: Bug: claude_local / codex_local adapters resume prior CLI sessions on wakeup, causing context bleed
## Problem

The `claude_local` and `codex_local` local adapters resume prior CLI sessions on every agent wakeup. On each `execute()` call the adapters read `runtime.sessionId` / `runtime.sessionParams.sessionId` and, when a session is found, pass `--resume <sessionId>` (Claude) or `resume <sessionId>` (Codex) to the CLI.

This is the intended design for long-running stateful agents, but it becomes a problem for agents that run independent, unrelated tasks:

- **Context bleed** — the resumed session carries the full conversation history of the previous run; each new task sees everything the agent has ever done
- **Bootstrap prompt silenced** — `bootstrapPromptTemplate` only fires when `!sessionId`; resuming a session suppresses it on every subsequent run
- **Non-determinism** — task outcomes depend on prior run history rather than only the current prompt
- **Cost inflation** — the growing context window is billed on every heartbeat even when previous turns are irrelevant

## Root cause

In both adapters (lines ~378–384 of `claude-local/src/server/execute.ts`, lines ~402–408 of `codex-local/src/server/execute.ts`):

```typescript
const runtimeSessionParams = parseObject(runtime.sessionParams);
const runtimeSessionId = asString(runtimeSessionParams.sessionId, runtime.sessionId ?? "");
const canResumeSession =
  runtimeSessionId.length > 0 && ...;
const sessionId = canResumeSession ? runtimeSessionId : null;
```

There is no way to opt out of session resumption in the agent config. Session params accumulate and are passed back on every wakeup with no per-agent escape hatch.

## Prior art

The `hermes-paperclip-adapter` (a third-party adapter already registered in `registry.js`) demonstrates that session handling can be fully controlled from outside the core adapters. The same escape-hatch pattern applies here — operators who need isolated runs should be able to configure it without forking the adapter.

## Proposed fix

Add an `isolateSession` boolean option to `agentConfigurationDoc` for both adapters. When `true`:

1. `canResumeSession` is forced to `false` — no session is resumed
2. `clearSession: true` is returned in the result — the server discards the resulting session and does not persist it

This is backwards-compatible (default `false`) and requires no change to agent configuration keys (`claude_local` / `codex_local` types stay the same).

A PR with this change is open for review.

dapzthelegend:
> **Thinking Path**
>
> 1. Paperclip's purpose is orchestration for zero-human companies — agents run tasks on wakeup, often unrelated to what they did before.
> 2. Local adapters maintain state across runs via session IDs so long-running agents can resume where they left off. That's the intended design for stateful workflows.
> 3. However, there is no way to opt out. Agents running independent tasks (one task per wakeup) inherit all prior conversation history on every run.
> 4. The result: `bootstrapPromptTemplate` fires only on the very first run; subsequent runs resume silently into a growing context; costs accumulate from irrelevant prior turns.
> 5. The fix must be opt-in and backwards-compatible — existing agents should see no change unless they set the new flag.
> 6. The right place to control this is the adapter config, not a global server setting, since some agents on the same server may want sessions and others may not.
> 7. The minimal change: add `isolateSession` to the config, add `!isolateSession` guard to `canResumeSession`, and set `clearSession: true` on the result so the server doesn't persist the new session ID either.
> 8. Both `claude_local` and `codex_local` share the same pattern and need the same fix.

## What Changed

**`packages/adapters/claude-local/src/server/execute.ts`**
- Added `const isolateSession = asBoolean(config.isolateSession, false)` after the runtime session extraction
- Added `!isolateSession &&` guard to `canResumeSession` — when true, forces a fresh session on every run
- Propagated `isolateSession` into `clearSession` in `toAdapterResult` — prevents the server from storing the resulting session ID

**`packages/adapters/codex-local/src/server/execute.ts`**
- Identical changes at the equivalent locations

**`packages/adapters/claude-local/src/index.ts`**
- Added `isolateSession` to `agentConfigurationDoc` under Operational fields

**`packages/adapters/codex-local/src/index.ts`**
- Same addition

## Verification

1. Configure a `claude_local` (or `codex_local`) agent with `isolateSession: true` in its config
2. Trigger two task runs back-to-back
3. Confirm: each run logs a new session ID (no `[paperclip] Claude session "..." is unavailable` or silent resume)
4. Confirm: `bootstrapPromptTemplate` fires on every run, not just the first
5. Confirm: agents without `isolateSession` in their config behave exactly as before (session resumption unchanged)

## Risks

Low. The new option defaults to `false` — all existing agents are unaffected. The `!isolateSession` guard is the only change to the session-resumption path. No schema migrations, no registry changes, no new dependencies.

## Model Used

None — human-authored

## Checklist

- [x] Thinking path traces from project purpose to specific change
- [x] Model use disclosed
- [x] Tested locally (verified session isolation behaviour with a local adapter run)
- [x] No new test coverage required — the change is a boolean guard on an existing conditional; behaviour is observable from adapter logs
- [x] No screenshots needed — no UI changes
- [x] No documentation updates beyond `agentConfigurationDoc` (already updated)
- [x] Risk assessed — low, opt-in, backwards-compatible
- [x] No Greptile comments to resolve at time of opening

Closes #2682

### DIFF
--- a/packages/adapters/claude-local/src/index.ts
+++ b/packages/adapters/claude-local/src/index.ts
@@ -31,6 +31,7 @@ Core fields:
 Operational fields:
 - timeoutSec (number, optional): run timeout in seconds
 - graceSec (number, optional): SIGTERM grace period in seconds
+- isolateSession (boolean, optional): when true, discard any stored session before each run and do not persist the resulting session; each task run starts from a clean context
 
 Notes:
 - When Paperclip realizes a workspace/runtime for a run, it injects PAPERCLIP_WORKSPACE_* and PAPERCLIP_RUNTIME_* env vars for agent-side tooling.
--- a/packages/adapters/claude-local/src/server/execute.ts
+++ b/packages/adapters/claude-local/src/server/execute.ts
@@ -378,7 +378,9 @@ export async function execute(ctx: AdapterExecutionContext): Promise<AdapterExec
   const runtimeSessionParams = parseObject(runtime.sessionParams);
   const runtimeSessionId = asString(runtimeSessionParams.sessionId, runtime.sessionId ?? "");
   const runtimeSessionCwd = asString(runtimeSessionParams.cwd, "");
+  const isolateSession = asBoolean(config.isolateSession, false);
   const canResumeSession =
+    !isolateSession &&
     runtimeSessionId.length > 0 &&
     (runtimeSessionCwd.length === 0 || path.resolve(runtimeSessionCwd) === path.resolve(cwd));
   const sessionId = canResumeSession ? runtimeSessionId : null;
@@ -574,7 +576,7 @@ export async function execute(ctx: AdapterExecutionContext): Promise<AdapterExec
       costUsd: parsedStream.costUsd ?? asNumber(parsed.total_cost_usd, 0),
       resultJson: parsed,
       summary: parsedStream.summary || asString(parsed.result, ""),
-      clearSession: clearSessionForMaxTurns || Boolean(opts.clearSessionOnMissingSession && !resolvedSessionId),
+      clearSession: clearSessionForMaxTurns || isolateSession || Boolean(opts.clearSessionOnMissingSession && !resolvedSessionId),
     };
   };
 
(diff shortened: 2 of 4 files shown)


dapzthelegend:
Addressed the remaining P2 items in 8d13d69c:

- Updated both adapters to log an explicit skip reason when a saved session is not resumed (`isolateSession=true` vs cwd mismatch context).
- Propagated `isolateSession` into early-return `clearSession` behavior so timeout/unparsed (claude-local) and timeout (codex-local) paths now clear consistently when isolation is enabled.

Files updated:
- `packages/adapters/claude-local/src/server/execute.ts`
- `packages/adapters/codex-local/src/server/execute.ts`

I also attempted focused adapter validation, but this checkout is missing dependencies (`node_modules`), so local `tsc`/`vitest` are currently unavailable until install.

Issue #2567, opened 2026-04-02
Fix 500 on issue comment polling with after=<commentId>


maximumsats:
## Summary

`GET /api/issues/:id/comments?after=<commentId>&order=asc` can return `500 Internal server error` when the server tries to poll incremental issue comments after an anchor comment.

This breaks agent/task loops that rely on incremental comment polling.

## Reproduction

Observed against a local Paperclip instance on 2026-04-02.

1. Create an issue with at least one comment.
2. Request incremental comments after the latest comment id:

```bash
curl -s "http://localhost:3100/api/issues/MAX-19/comments?after=3cb9ddcc-0514-4aff-ac3c-b8ee6f16104a&order=asc" \
  -H "Authorization: Bearer <agent-api-key>"
```

3. The server responds with:

```json
{"error":"Internal server error"}
```

## Expected

The route should return an empty array or the comments after the anchor comment.

## Actual

The route returns `500` and the server logs show:

```text
TypeError [ERR_INVALID_ARG_TYPE]: The "string" argument must be of type string or an instance of Buffer or ArrayBuffer. Received an instance of Date
    at Buffer.byteLength (node:buffer:804:11)
    at reset.str (.../node_modules/postgres/src/bytes.js:22:27)
```

## Likely cause

In `server/src/services/issues.ts`, the `listComments()` implementation interpolates `anchor.createdAt` directly into the SQL fragment used for the `afterCommentId` cursor comparison.

When `anchor.createdAt` is a `Date`, the postgres driver trips over the binding path and throws.

## Proposed fix

Normalize `anchor.createdAt` before interpolation, for example:

- convert `Date` to `toISOString()` before using it in the SQL fragment
- then compare against that normalized value in both ascending and descending branches

I opened a PR with that minimal fix.

codoublez:
Can confirm this bug — hitting it on our PaperclipAI instance with 6 agents running via the `opencode_local` adapter. The CEO agent's heartbeat makes ~80+ tool calls (reviewing issues, posting comments, managing tasks) but gets marked as "failed" because the comment polling endpoint returns 500. The agent's actual work completes fine — it's just the comment cursor pagination that breaks. Would love to see this fix merged soon! 🙏

PR #2694, opened 2026-04-04
fix(server): serialize comment cursor timestamps for after queries

### LINKED ISSUE #2612: 500 error on GET /api/issues/:id/comments?after=:commentId — Date object passed as string to PostgreSQL
## Summary

The `GET /api/issues/:issueId/comments?after=:commentId&order=asc` endpoint throws a 500 error due to a `Date` object being passed where a `string` is expected in the PostgreSQL query binding.

## Error

```
TypeError [ERR_INVALID_ARG_TYPE]: The "string" argument must be of type string or an instance of Buffer or ArrayBuffer. Received an instance of Date
    at Function.byteLength (node:buffer:777:11)
    at Function.str (postgres/src/bytes.js:22:27)
    at postgres/src/connection.js:964:16
    at Array.forEach (<anonymous>)
    at Bind (postgres/src/connection.js:954:16)
```

## Steps to Reproduce

1. Create an issue with comments
2. An agent calls `GET /api/issues/{issueId}/comments?after={commentId}&order=asc`
3. Server returns HTTP 500

## Root Cause

A `Date` object is being passed directly to the `postgres` driver's query binding instead of being serialized to an ISO string first. The `postgres` library's `Bind` function calls `Function.str()` → `Function.byteLength()` which expects a string, not a Date.

Likely location: the comments query builder where the `after` cursor is resolved to a comment's `created_at` (Date) and then used in a subsequent query without `.toISOString()` conversion.

## Impact

- Agents repeatedly hit this error during heartbeat when trying to fetch incremental comment updates
- This can cause agent heartbeats to fail and work to stall
- The error is persistent — every agent that calls this endpoint with an `after` parameter triggers it

## Environment

- Paperclip: v2026.325.0 (via npx)
- OS: Windows 11 Pro
- Node.js: v22.17.0
- Database: Embedded PostgreSQL (UTF8)

MindSyncHub:
## Thinking Path
`GET /api/issues/:id/comments?after=...&order=asc` resolves the anchor comment and then reuses its `createdAt` value inside a cursor predicate. Drizzle returns that field as a JavaScript `Date`, but the downstream postgres binding path here expects a serialized string/buffer value. That mismatch produces the reported 500 when agents poll for incremental comments.

## What Changed
- serialize the anchor comment timestamp before binding it into the cursor SQL predicate
- keep the existing cursor semantics intact for both ascending and descending pagination
- add an embedded-Postgres regression test covering `afterCommentId` comment listing in ascending order

## Verification
- `pnpm exec vitest run server/src/__tests__/issues-service.test.ts`
- `pnpm -r typecheck`

## Risks
- low: the change is scoped to cursor binding for comment pagination and does not alter comment shape or ordering rules
- the query now compares against a normalized ISO timestamp string instead of a raw `Date` object

## Model Used
GPT-5 Codex

## Checklist
- [x] Tests added or updated
- [x] Targeted verification completed
- [x] Behavior scoped to the reported bug

Closes #2612
Related: #2567

### DIFF
--- a/server/src/__tests__/issues-service.test.ts
+++ b/server/src/__tests__/issues-service.test.ts
@@ -1017,6 +1017,64 @@ describeEmbeddedPostgres("issueService.list participantAgentId", () => {
     expect(result?.description).toHaveLength(1200);
     expect(result?.description?.endsWith("—")).toBe(true);
   });
+  it("lists comments after an anchor comment without binding Date objects into the cursor query", async () => {
+    const companyId = randomUUID();
+    const issueId = randomUUID();
+    const firstCommentId = randomUUID();
+    const secondCommentId = randomUUID();
+    const thirdCommentId = randomUUID();
+
+    await db.insert(companies).values({
+      id: companyId,
+      name: "Paperclip",
+      issuePrefix: `T${companyId.replace(/-/g, "").slice(0, 6).toUpperCase()}`,
+      requireBoardApprovalForNewAgents: false,
+    });
+
+    await db.insert(issues).values({
+      id: issueId,
+      companyId,
+      title: "Cursor pagination issue",
+      status: "todo",
+      priority: "medium",
+      createdAt: new Date("2026-03-26T10:00:00.000Z"),
+      updatedAt: new Date("2026-03-26T10:00:00.000Z"),
+    });
+
+    await db.insert(issueComments).values([
+      {
+        id: firstCommentId,
+        companyId,
+        issueId,
+        body: "first",
+        createdAt: new Date("2026-03-26T10:01:00.000Z"),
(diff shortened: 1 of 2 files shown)


wook95:
This fix resolved the exact 500 error we were hitting in production with agent heartbeat comment polling. Applied locally and confirmed working. Thanks for the clean fix + test coverage. 🙏

MindSyncHub:
Refreshed this PR on top of `08af8304` (current `origin/master`).

Re-applied the cursor serialization fix in `listComments` so after-comment pagination compares stable timestamp strings instead of binding raw `Date` objects.

Focused verification:
- `corepack pnpm vitest run server/src/__tests__/issues-service.test.ts`

MindSyncHub:

PR #2764, opened 2026-04-04
fix(ui): batch persisted live transcript updates

### LINKED ISSUE #2736: Dashboard agent cards show 'No transcript captured' due to React state race in useLiveRunTranscripts
## Bug

Dashboard agent cards intermittently show **"No transcript captured."** for finished runs that have full transcript data. Refreshing the page shows different runs with/without transcripts each time. The run detail page (`/agents/{slug}/runs/{id}`) always renders transcripts correctly.

## Root Cause

`useLiveRunTranscripts` fires parallel `heartbeatsApi.log()` calls via `Promise.all`, and each resolved promise independently calls `appendChunks` → `setChunksByRun`. When multiple async state updater calls race in React 18's batching, some updates are silently lost — the state ends up containing chunks for only 1–2 runs instead of all 4.

A secondary issue: `seenChunkKeysRef` (a `useRef`) survives React StrictMode's dev-only unmount-remount cycle, but `chunksByRun` state does not. On the second mount, every chunk is treated as a duplicate and skipped, leaving the transcript permanently empty.

## Reproduction

1. Have 4+ finished heartbeat runs for any agent (Hermes adapter, but likely affects all)
2. Navigate to the company dashboard
3. Observe that only 1–2 of the 4 agent cards show transcript content
4. Refresh — different cards show content each time (non-deterministic)
5. Click into any "No transcript" run via Agents → Runs tab — full transcript is visible there

Most visible in dev mode (React StrictMode) but the `Promise.all` race affects production too.

## Fix

Two changes in `ui/src/components/transcript/useLiveRunTranscripts.ts`:

**1. Clear refs on effect re-run** — prevents stale dedup keys after StrictMode remount:

```ts
// At the start of the log-reading useEffect:
seenChunkKeysRef.current.clear();
pendingLogRowsByRunRef.current.clear();
logOffsetByRunRef.current.clear();
```

**2. Batch all runs into a single state update** — replaces per-run `appendChunks` with one atomic `setChunksByRun`:

```ts
const readAll = async () => {
  const results = await Promise.allSettled(
    currentRuns.map(async (run) => {
      const offset = logOffsetByRunRef.current.get(run.id) ?? 0;
      const result = await heartbeatsApi.log(run.id, offset, LOG_READ_LIMIT_BYTES);
      return { run, result, offset };
    }),
  );
  if (cancelled) return;

  // Parse all results, then apply ONE setChunksByRun
  const allParsed = new Map<string, Array<RunLogChunk & { dedupeKey: string }>>();
  for (const settled of results) {
    if (settled.status !== "fulfilled") continue;
    const { run, result, offset } = settled.value;
    const parsed = parsePersistedLogContent(run.id, result.content, pendingLogRowsByRunRef.current);
    if (parsed.length > 0) allParsed.set(run.id, parsed);
    // ... update offsets
  }

  if (allParsed.size > 0) {
    setChunksByRun((prev) => {
      // Single update with all runs' chunks
    });
  }
};
```

This ensures all runs' chunks land in the same React state transition. The existing `appendChunks` function is preserved for the WebSocket live-streaming path (which processes one event at a time and doesn't have the parallel race).

## Environment

- Paperclip `v2026.325.0` (also reproduces on canary `2026.403.0-canary.10`)
- React 18 with StrictMode enabled (default dev config)
- Hermes local adapter, but the bug is adapter-agnostic — it's in the shared transcript hook

Happy to open a PR with the full diff if helpful.

MindSyncHub:
## Summary
- batch persisted run-log reads into a single transcript state update per poll
- clear transcript dedupe and offset refs when the persisted log effect restarts
- add a focused hook test covering multiple finished runs under StrictMode

## Testing
- pnpm exec vitest run ui/src/components/transcript/useLiveRunTranscripts.test.tsx
- pnpm --filter @paperclipai/ui typecheck

Closes #2736

### DIFF
--- a/ui/src/components/transcript/useLiveRunTranscripts.strictmode.test.tsx
+++ b/ui/src/components/transcript/useLiveRunTranscripts.strictmode.test.tsx
@@ -1,0 +1,162 @@
+// @vitest-environment jsdom
+
+import { StrictMode, useEffect } from "react";
+import { act } from "react";
+import { createRoot } from "react-dom/client";
+import { QueryClient, QueryClientProvider } from "@tanstack/react-query";
+import type { LiveRunForIssue } from "@/api/heartbeats";
+import { afterEach, beforeEach, describe, expect, it, vi } from "vitest";
+import { useLiveRunTranscripts } from "./useLiveRunTranscripts";
+
+const getGeneralMock = vi.fn(async () => ({ censorUsernameInLogs: false }));
+const logMock = vi.fn();
+
+vi.mock("../../api/instanceSettings", () => ({
+  instanceSettingsApi: {
+    getGeneral: () => getGeneralMock(),
+  },
+}));
+
+vi.mock("../../api/heartbeats", () => ({
+  heartbeatsApi: {
+    log: (runId: string, offset = 0, limitBytes = 256000) => logMock(runId, offset, limitBytes),
+  },
+}));
+
+vi.mock("../../adapters", async () => {
+  const actual = await vi.importActual<typeof import("../../adapters")>("../../adapters");
+  return {
+    ...actual,
+    onAdapterChange: () => () => {},
+  };
+});
+
+// eslint-disable-next-line @typescript-eslint/no-explicit-any
+(globalThis as any).IS_REACT_ACT_ENVIRONMENT = true;
+
+function createRun(id: string): LiveRunForIssue {
+  return {
+    id,
+    status: "succeeded",
+    invocationSource: "assignment",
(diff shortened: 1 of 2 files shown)


MindSyncHub:
Refreshed this PR on top of `40782f70` after the base branch advanced.

Focused verification passed:
- `corepack pnpm vitest run ui/src/components/transcript/useLiveRunTranscripts.strictmode.test.tsx`

MindSyncHub:

PR #2818, opened 2026-04-05
fix(server): prevent identifier collision in issue creation

### LINKED ISSUE #2705: POST /api/companies/{id}/issues returns 200 but row never commits (identifier collision)
## Bug: Task creation returns 200 but row doesn't commit

When creating issues via the API, the endpoint returns HTTP 200 with a complete issue object (including `id`, `identifier`, `status`) but the row is never actually committed to the database.

### Root cause

The `issue_counter` on the `companies` table gets out of sync with the actual max `issue_number` in the `issues` table. When this happens:

1. POST creates a new issue, increments counter to N
2. Generates identifier `BOAA-{N}`
3. DB insert fails with `23505: duplicate key value violates unique constraint "issues_identifier_idx"` — `Key (identifier)=(BOAA-{N}) already exists`
4. Transaction rolls back
5. **But the HTTP response has already been sent with 200 and the generated issue data**

The counter increments on each failed attempt but the collision persists because the existing row at that identifier was created by a previous successful insert.

### How to reproduce

1. Create several issues rapidly via the API
2. Delete some of them manually (or let them be created by different paths — UI vs API)
3. The `issue_counter` and actual max `issue_number` diverge
4. Next API creation hits the unique constraint and silently fails

### Server logs
"name": "PostgresError",
"severity": "ERROR",
"code": "23505",
"detail": "Key (identifier)=(BOAA-104) already exists.",
"table_name": "issues",
"constraint_name": "issues_identifier_idx",
"routine": "_bt_check_unique"

### Expected behaviour

The API should either:
- Use `SELECT MAX(issue_number) + 1` instead of relying on the counter column
- Catch the 23505 error, increment the counter, and retry
- Not return 200 until the DB transaction is confirmed committed

### Workaround

Manually update the counter:
```sql
UPDATE companies SET issue_counter = (SELECT MAX(issue_number) + 1 FROM issues WHERE company_id = '{id}') WHERE id = '{id}';
Environment
Paperclip v2026.325.0
Supabase PostgreSQL 17.6
Issue creation via POST /api/companies/{companyId}/issues with agent API key

mvanhorn:
## Thinking Path

> - Companies create issues via POST /api/companies/{id}/issues
> - Each issue gets a unique identifier like PREFIX-N based on a counter on the companies table
> - But the counter can drift out of sync with the actual max issue_number in the issues table
> - When that happens, the insert hits a duplicate key violation on the identifier unique index (error 23505) and the transaction rolls back
> - This PR adds a self-correcting counter that uses GREATEST(counter, MAX(issue_number)) + 1
> - The benefit is that counter desync is auto-corrected on the next issue creation, preventing silent failures

## What Changed

- Modified `server/src/services/issues.ts` in the `create` function's transaction block
- Before incrementing the counter, query `MAX(issue_number)` from the issues table for the company
- Use `GREATEST(counter, max_issue_number) + 1` instead of `counter + 1` so the counter self-corrects if it has drifted below the actual max
- This is a single additional query within the existing transaction, so it shares the same isolation level and adds no new failure modes

## Verification

- The fix is within the existing transaction block, so atomicity is preserved
- If `issue_counter` is already correct (the common case), `GREATEST` picks the counter and behavior is identical to before
- If `issue_counter` has drifted below `MAX(issue_number)`, the counter jumps to `max + 1`, skipping the collision
- Pre-existing test failures in UI (IssueDocumentsSection, Routines) are unrelated to this change (localStorage.clear issue on master)

## Risks

- Low risk. The `GREATEST` function is supported by PostgreSQL and adds one subquery to the transaction. The query plan should use the existing index on `issues.company_id`.
- No schema changes, no new columns, no migration needed.

## Model Used

Claude Opus 4.6 (1M context) with tool use. Implementation was guided by reading the existing `create` function in `server/src/services/issues.ts` and the Drizzle ORM patterns used throughout the codebase.

## Checklist

- [x] I have run tests locally and they pass (4 pre-existing UI failures unrelated to this change)

Fixes #2705

This contribution was developed with AI assistance (Claude Code).

### DIFF
--- a/server/src/services/issues.ts
+++ b/server/src/services/issues.ts
@@ -1183,9 +1183,19 @@ export function issueService(db: Db) {
         if (executionWorkspaceId) {
           await assertValidExecutionWorkspace(companyId, issueData.projectId, executionWorkspaceId, tx);
         }
+        // Self-correcting counter: use MAX(issue_number) + 1 if the counter
+        // has drifted below the actual max, preventing identifier collisions.
+        const [maxRow] = await tx
+          .select({ maxNum: sql<number>`coalesce(max(${issues.issueNumber}), 0)` })
+          .from(issues)
+          .where(eq(issues.companyId, companyId));
+        const currentMax = maxRow?.maxNum ?? 0;
+
         const [company] = await tx
           .update(companies)
-          .set({ issueCounter: sql`${companies.issueCounter} + 1` })
+          .set({
+            issueCounter: sql`greatest(${companies.issueCounter}, ${currentMax}) + 1`,
+          })
           .where(eq(companies.id, companyId))
           .returning({ issueCounter: companies.issueCounter, issuePrefix: companies.issuePrefix });
 


cryppadotta:
Ty

Issue #2829, opened 2026-04-05
Plugin activation failure leaves zombie worker handle, blocks subsequent enable attempts


abekarar:
## Summary

When `activatePlugin()` in `plugin-loader.ts` fails **after** `workerManager.startWorker()` succeeds — i.e., during any of the subsequent registration steps (jobs, event subscriptions, webhooks, tools) — the worker handle is left running and registered in the `PluginWorkerManager`'s `workers` Map. The DB plugin row is correctly rolled back to `status=error`, but the in-memory worker handle is orphaned.

The next time an operator tries to enable the plugin (via `POST /api/plugins/:id/enable` or `paperclipai plugin enable <key>`), activation fails with:

```
API error 400: Worker already registered for plugin "<uuid>" (status: running)
```

The only workaround is to wait for the zombie worker to crash on its own (e.g. its polling loop fails) or restart the server to clear in-memory state.

## Observed

Hit this today after a `paperclip-plugin-telegram` plugin failed its worker-init RPC (separate issue — see mvanhorn/paperclip-plugin-telegram#19). After the host patched the plugin dist to stop blocking init, the first three `enable` attempts failed with:

```
$ paperclipai plugin enable paperclip-plugin-telegram
API error 400: Worker already registered for plugin "acada116-…" (status: running)

$ paperclipai plugin inspect paperclip-plugin-telegram
key=paperclip-plugin-telegram  status=error  version=0.2.4  id=acada116-…
Last error:
Activation failed: Worker already registered for plugin "acada116-…" (status: running)
```

The `status: running` part is the smoking gun — it means the patched worker.js had started successfully, `setStatus(\"running\")` fired at `plugin-worker-manager.ts:854`, and **then** something later in `activatePlugin` threw. The fourth `enable` eventually succeeded (presumably after the zombie polling loop crashed on its own).

## Root cause

`server/src/services/plugin-loader.ts` — `activatePlugin()` wraps steps 4–8 in a single `try` (lines ~1694–1838) but the `catch` handler (lines ~1840–1865) only marks the DB row as errored:

```ts
await workerManager.startWorker(pluginId, workerOptions);  // line 1744 — may succeed
registered.worker = true;

// Steps 5–8: syncJobDeclarations, registerPlugin, forPlugin, registerPluginTools
// …any of these can throw…

} catch (err) {
  // Mark the plugin as errored in the database so it is not retried
  // automatically on next startup without operator intervention.
  try {
[...]

PR #2830, opened 2026-04-05
fix: make Telegram plugin init non-blocking to avoid worker-init timeout

abekarar:
## Summary

Fixes a 15s worker-init timeout that takes the Telegram plugin out of service whenever `api.telegram.org` is slow or unreachable. The plugin's `setup()` was awaiting a remote HTTP call (`setMyCommands`) before completing init, so if Telegram's API took >15s, the host SIGKILLed the worker and flipped the plugin to `status=error`.

Observed symptom in `pnpm dev` logs:

```
ERROR: worker initialize failed
    err: "RPC call \"initialize\" timed out after 15000ms"
ERROR: host handler error   method: "http.fetch"   err: "The operation was aborted"
```

(The second error is the same outbound fetch being aborted 30s later by `PLUGIN_FETCH_TIMEOUT_MS` in `plugin-host-services.ts:44`.)

## Approach

The plugin source lives in an external repo (`github.com/mvanhorn/paperclip-plugin-telegram`) so the root-cause TS fix went there: mvanhorn/paperclip-plugin-telegram#19. This PR adds the equivalent dist-file patch for our local dev workflow so nobody has to wait for upstream to merge + cut a release.

Three changes:

1. **`scripts/local/telegram-plugin-nonblocking-init.patch.sh` (new)** — idempotent dist patch that converts `await setMyCommands(...)` to fire-and-forget with `.then()/.catch()`. Matches the fire-and-forget pattern already used by `pollUpdates()` a few lines later in the same plugin. Uses `${PAPERCLIP_HOME:-$HOME/.paperclip}` (mirrors `server/src/home-paths.ts:16-19`) so it's portable across dev machines. Prints WARN and exits 0 if the pattern isn't found (e.g. plugin was upgraded past the bug) rather than failing hard — matches the two existing Telegram patch scripts' behavior.

2. **`scripts/local/apply-local-patches.sh` (new to git)** — orchestrator that applies all `scripts/local/*.patch` and `*.patch.sh`. Was present in working trees but never committed. Portable via `cd "$(git rev-parse --show-toplevel)"`.

3. **`server/package.json`** — adds `preflight:local-patches` npm script, chains it into `dev` and `dev:watch` so patches survive plugin reinstalls automatically (before this, operators had to manually re-run after every `pnpm install`).

## Blast radius

**Local dev workflow only.** Zero CI surface — verified no workflow in `.github/workflows/*.yml` invokes `dev`, `dev:watch`, or `apply-local-patches`. `typecheck`, `build`, `start`, `prepack` scripts are untouched. No changes to server runtime, DB, migrations, auth, or production code.

## Test plan

- [x] `bash -n` syntax check passes on both new scripts
- [x] `package.json` remains valid JSON
- [x] Patch script tested across all 4 execution paths (plugin-not-installed, already-applied, fresh-apply, pattern-absent) — all exit 0 with correct behavior
- [x] `pnpm run preflight:local-patches` runs clean from `server/` cwd
- [x] Live verification: enabled the previously-errored Telegram plugin via `paperclipai plugin enable paperclip-plugin-telegram` and confirmed `status=ready` — worker starts without the 15s init timeout
- [ ] Reviewer: optionally run `pnpm dev` after merging to see the preflight chain output before the ASCII banner

## Related

- Upstream PR with the root-cause TS fix: mvanhorn/paperclip-plugin-telegram#19
- Separate issue uncovered during verification (zombie worker handle on activation failure): #2829

### DIFF
--- a/scripts/local/telegram-plugin-nonblocking-init.patch.sh
+++ b/scripts/local/telegram-plugin-nonblocking-init.patch.sh
@@ -14,7 +14,8 @@
 
 set -euo pipefail
 
-TARGET_WORKER="/home/abekarar/.paperclip/plugins/node_modules/paperclip-plugin-telegram/dist/worker.js"
+PAPERCLIP_HOME_DIR="${PAPERCLIP_HOME:-$HOME/.paperclip}"
+TARGET_WORKER="${PAPERCLIP_HOME_DIR}/plugins/node_modules/paperclip-plugin-telegram/dist/worker.js"
 
 if [ ! -f "$TARGET_WORKER" ]; then
   echo "Telegram plugin not installed — skipping patch"
@@ -26,8 +27,9 @@ if grep -q "Non-blocking init: don't hold up worker initialize" "$TARGET_WORKER"
   exit 0
 fi
 
-python3 << 'PY'
-worker_path = "/home/abekarar/.paperclip/plugins/node_modules/paperclip-plugin-telegram/dist/worker.js"
+python3 - "$TARGET_WORKER" << 'PY'
+import sys
+worker_path = sys.argv[1]
 
 with open(worker_path, 'r') as f:
     worker = f.read()
@@ -54,8 +56,8 @@ if worker_old in worker:
         f.write(worker)
     print("Worker patch: applied (non-blocking setMyCommands in setup())")
 else:
+    # Pattern absent may mean the plugin was upgraded past the broken version
+    # (e.g. mvanhorn/paperclip-plugin-telegram#19 merged upstream) — don't fail
+    # pnpm dev on that case; just warn so operators can prune the patch.
     print("Worker patch: WARN — expected pattern not found, nothing changed")
-    raise SystemExit(1)
 PY
-
-echo "Telegram plugin patched (non-blocking init)"


abekarar:
Closing — this PR adds local dev-workflow tooling (`scripts/local/*` patch scripts + preflight wiring) that isn't appropriate to land in the upstream repo. The root-cause fix for the underlying Telegram init timeout belongs in the plugin source, not here: mvanhorn/paperclip-plugin-telegram#19.

**For anyone hitting the same symptom** — `RPC call "initialize" timed out after 15000ms` for `paperclip-plugin-telegram`, followed by `host handler error method: "http.fetch"` 15s later — the cause is that the plugin's `setup()` awaits a remote HTTP call to `api.telegram.org` (`setMyCommands`) during init, and if Telegram's API is slow/unreachable, `setup()` blocks past the host's 15s worker-init RPC timeout (`server/src/services/plugin-worker-manager.ts:64`). The host then SIGKILLs the worker and flips the plugin to `status=error`.

**Fix:** convert the `await setMyCommands(ctx, token, allCommands)` call in the plugin's `setup()` to fire-and-forget with `.then()/.catch()`, matching the pattern `pollUpdates()` already uses a few lines below it. Full diff is at mvanhorn/paperclip-plugin-telegram#19; the equivalent dist-file patch is in this PR's commit `ab194b5c`.

**Related** (separate pre-existing bug uncovered while verifying this one): #2829 — plugin activation failure after `workerManager.startWorker()` succeeds leaves a zombie worker handle that blocks subsequent `enable` attempts.

PR #2893, opened 2026-04-06
fix: prevent stale index.html being served after ui-dist refresh

Lempkey:
## Thinking Path

> - The server hosts a React SPA via an Express catch-all route that serves `index.html` for any unmatched path
> - At startup, the server reads `index.html` from the `ui-dist` folder once and holds the HTML string in process memory indefinitely
> - When a hotpatch or ui-dist rebuild replaces the file on disk, the in-memory string is never invalidated — the server keeps serving the old shell until it is manually restarted
> - This pull request introduces `createIndexHtmlGetter()`, which re-reads `index.html` only when its `mtime` changes, giving the catch-all route a self-healing cache without requiring a server restart
> - The benefit is that ui-dist refreshes and hotpatches are picked up immediately, eliminating a whole class of "why is the UI still showing the old version?" incidents

## What Changed

- **`server/src/app.ts`** — Extracted `createIndexHtmlGetter(path)` (exported for testability). It lazily caches the `applyUiBranding`-processed HTML and invalidates the cache only when `fs.statSync(path).mtimeMs` changes. The existing SPA catch-all now calls this getter on every request instead of closing over the startup-time string.
- **`server/src/__tests__/spa-index-html-cache.test.ts`** — Three unit tests: initial read, cache hit when mtime is unchanged (pins mtime to a clean integer to avoid sub-millisecond precision loss from `utimesSync`), and cache miss when mtime advances.

## Verification

```bash
# Build plugin-sdk (required for server tests to resolve the import)
pnpm --filter @paperclipai/plugin-sdk build

# Run the new tests
npx vitest run server/src/__tests__/spa-index-html-cache.test.ts
# → 3 passed

# Existing app tests still pass
npx vitest run server/src/__tests__/app-hmr-port.test.ts
# → 3 passed
```

Manual smoke test: start the server, verify the SPA loads, replace `ui-dist/index.html` on disk, reload the browser — the new content is served without restarting the process.

## Risks

- `fs.statSync` is called on every SPA request. This is a single syscall with negligible overhead compared to the HTTP round-trip, but it is a new per-request syscall. If this is ever a concern the call could be debounced (e.g. at most once per second), but that adds complexity not warranted here.
- Low risk overall: the change is isolated to the SPA catch-all path and does not affect API routes, static asset serving, or the Vite dev-server HMR path.

## Model Used

- Provider: Anthropic  
- Mode: Interactive agentic session via Claude Code CLI (tool use enabled, extended thinking active)

## Checklist

- [ ] If this change affects the UI, I have included before/after screenshots *(server-side only, no UI change)*
- [ ] I have updated relevant documentation to reflect my changes *(no doc changes needed)*

### DIFF
--- a/server/src/__tests__/spa-index-html-cache.test.ts
+++ b/server/src/__tests__/spa-index-html-cache.test.ts
@@ -37,6 +37,25 @@ describe("createIndexHtmlGetter", () => {
     expect(get()).toBe(first);
   });
 
+  it("returns cached html when file is transiently absent after first read", () => {
+    const dir = fs.mkdtempSync(path.join(os.tmpdir(), "paperclip-spa-"));
+    const p = writeTempIndexHtml(dir, "<html>v1</html>");
+    const get = createIndexHtmlGetter(p);
+
+    const first = get(); // prime cache
+
+    // Simulate the file being removed mid-hotpatch
+    fs.unlinkSync(p);
+
+    expect(get()).toBe(first);
+  });
+
+  it("throws when file is absent on the very first read", () => {
+    const p = path.join(os.tmpdir(), "paperclip-spa-nonexistent-index.html");
+    const get = createIndexHtmlGetter(p);
+    expect(() => get()).toThrow();
+  });
+
   it("re-reads from disk when mtime advances", () => {
     const dir = fs.mkdtempSync(path.join(os.tmpdir(), "paperclip-spa-"));
     const p = writeTempIndexHtml(dir, "<html>v1</html>");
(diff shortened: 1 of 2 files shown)


musay4632 (approved):

Lempkey:
The `verify` check failure is unrelated to this PR. All 183 test files passed including the new `spa-index-html-cache.test.ts` (5/5 tests green).

The failure is an unhandled exception from a pre-existing test:

```
TypeError: Cannot read properties of null (reading 'write')
  in server/src/__tests__/heartbeat-comment-wake-batching.test.ts
```

This is a postgres connection teardown race condition that occurs after the test suite has already completed. The vitest summary confirms: **933 passed, 1 skipped, 1 unhandled error** — the error originated in an unrelated test file.

Happy to re-run if needed, or if a maintainer can trigger it that would be great!

PR #2921, opened 2026-04-06
fix: downgrade jsdom to ^27 to resolve ESM/CJS startup crash

### LINKED ISSUE #2900: Bug: paperclipai run fails on all supported Node versions — ESM/CJS interop error
Summary

paperclipai run fails to start the server on every Node version in the declared engine range (^20.19.0 || ^22.12.0 || >=24.0.0). The crash originates from html-encoding-sniffer attempting to require() @exodus/bytes/encoding-lite.js, which is an ESM module with top-level await. No supported Node version avoids this crash.
Environment
OSUbuntu 24 (Linux)ShellbashInstall methodnpm install -g paperclipaiPackage versionpaperclipai@2026.403.0

Node versions tested

Node versionErrorv20.18.0require() of ES Module ...@exodus/bytes/encoding-lite.js not supportedv20.19.4 (system)require() cannot be used on an ESM graph with top-level awaitv22.12.0require() cannot be used on an ESM graph with top-level await
All three versions pass paperclip doctor with 9/9 checks, then crash immediately when the server process starts.

Steps to reproduce
bashnpm install -g paperclipai
paperclipai onboard --yes
paperclipai run

Observed behavior
Doctor checks pass cleanly:
◇  Summary ──╮
│            │
│  9 passed  │
│            │
├────────────╯
│
└  All checks passed!
│
◇  Starting Paperclip server...
Paperclip server failed to start.
Error on Node v22.12.0 / v20.19.4
require() cannot be used on an ESM graph with top-level await. Use import() instead.
To see where the top-level await comes from, use --experimental-print-required-tla.
node:internal/modules/esm/module_job:384
      throw new ERR_REQUIRE_ASYNC_MODULE();
            ^
Error [ERR_REQUIRE_ASYNC_MODULE]: require() cannot be used on an ESM graph with top-level await.
    at ModuleJobSync.runSync (node:internal/modules/esm/module_job:384:13)
    at ModuleLoader.importSyncForRequire (node:internal/modules/esm/loader:323:47)
    at loadESMFromCJS (node:internal/modules/cjs/loader:1371:24)
    at Module._compile (node:internal/modules/cjs/loader:1511:5)
    at Module._extensions..js (node:internal/modules/cjs/loader:1572:16)
    at Module.load (node:internal/modules/cjs/loader:1275:32)
    at Module._load (node:internal/modules/cjs/loader:1096:12)
    at Module.require (node:internal/modules/cjs/loader:1298:19)
    at require (node:internal/modules/helpers:182:18)
    at Object.<anonymous> (/home/santi/.nvm/versions/node/v22.12.0/lib/node_modules/paperclipai/node_modules/cssstyle/lib/parsers.js:6:5)
Error on Node v20.18.0
require() of ES Module /home/santi/.nvm/versions/node/v20.18.0/lib/node_modules/paperclipai/
node_modules/@exodus/bytes/encoding-lite.js from /home/santi/.nvm/versions/node/v20.18.0/lib/
node_modules/paperclipai/node_modules/html-encoding-sniffer/lib/html-encoding-sniffer.js not supported.
Instead change the require of encoding-lite.js in html-encoding-sniffer/lib/html-encoding-sniffer.js
to a dynamic import() which is available in all CommonJS modules.

Expected behavior
paperclipai run starts the server and the dashboard is accessible at http://127.0.0.1:3100.

Root cause analysis

The crash chain is:
cssstyle (CJS) 
  → requires html-encoding-sniffer (CJS)
    → requires @exodus/bytes/encoding-lite.js (ESM, has top-level await)
html-encoding-sniffer@6.0.0 and @exodus/bytes@1.15.0 both declare engine requirements of ^20.19.0 || ^22.12.0 || >=24.0.0, meaning they were written assuming the stricter ESM enforcement introduced in those versions. However, cssstyle still uses require() to load them synchronously — which is incompatible with ESM modules containing top-level await regardless of Node version.

The issue is in the CJS→ESM boundary: cssstyle needs to be updated to use dynamic import() instead of require() for these ESM dependencies, or html-encoding-sniffer and @exodus/bytes need to ship a CJS-compatible build.
Additional notes

NODE_OPTIONS="--experimental-require-module" does not resolve the crash on Node 22.
npm install completes successfully on all tested versions (with peer dependency warnings for drizzle-orm and zod version mismatches, which appear to be separate issues).

The server never starts; the embedded Postgres instance is never reached.

This makes the package currently uninstallable on Linux via the documented method.

Workaround
None confirmed. The package cannot be started on any Node version in the declared engine range.

### LINKED ISSUE #2914: Can't install - server fails to start due to ESM/CJS dependency error (cssstyle / @asamuzakjp/css-color)
paperclip doctor passes all checks, but the server fails immediately on startup.

> require() cannot be used on an ESM graph with top-level await
> From: cssstyle/lib/parsers.js
> Requiring: @asamuzakjp/css-color/dist/esm/index.js

This appears to be caused by a broken dependency chain (jsdom -> cssstyle -> @asamuzakjp/css-color) introducing an ESM/CJS boundary issue.

This is not an environment-specific issue:

- Clean installs fail
- Multiple versions fail (latest and previous)

The CLI reports a valid setup, but the server cannot start, making the tool unusable.

This should be handled at the dependency/packaging level (e.g. pinning jsdom or cssstyle), as suggested in related issues.

Expected: Server starts after successful doctor check

Actual: Immediate crash on startup

### LINKED ISSUE #2916: Server fails to start on macOS: cssstyle require() of ESM module with top-level await
paperclipai run fails on macOS (Apple Silicon) on every Node version tested.

Error:
require() cannot be used on an ESM graph with top-level await.
From: cssstyle/lib/parsers.js
Requiring: @asamuzakjp/css-color/dist/esm/index.js

Tested versions:
- Node 22.22.2 (Homebrew node@22) → cssstyle error
- Node 20.18.3 (nvm) → html-encoding-sniffer/@exodus/bytes error

paperclipai version: 2026.325.0
macOS: Apple Silicon (arm64)

The issue is in the dependency chain:
jsdom → cssstyle → @asamuzakjp/css-color (ESM with top-level await)

Suggested fix: upgrade jsdom to v29+ which removes cssstyle entirely.

### LINKED ISSUE #2917: Server fails to start: ESM/CJS incompatibility in cssstyle → @asamuzakjp/css-color dependency chain
## Bug Report

`paperclipai run` completes onboarding and passes all doctor checks, but the server fails to start due to ESM/CJS module incompatibilities in the dependency chain. **No Node.js version currently works.**

## Environment

- **paperclipai version:** 2026.403.0
- **OS:** macOS 26.3 (arm64 / Apple Silicon)
- **Install method:** `npm install -g paperclipai` (also reproduced via `npx`)
- **Config:** Default instance, embedded-postgres, local_trusted mode

## Error (Node 20.19+ and Node 22+)

```
Paperclip server failed to start.
require() cannot be used on an ESM graph with top-level await. Use import() instead.
To see where the top-level await comes from, use --experimental-print-required-tla.
  From .../node_modules/cssstyle/lib/parsers.js
  Requiring .../node_modules/@asamuzakjp/css-color/dist/esm/index.js
```

`cssstyle@6.2.0` uses `require()` to load `@asamuzakjp/css-color`, which is ESM-only with top-level `await`. Node 20.19+ and 22+ enforce the restriction that `require()` cannot load ESM modules with top-level `await`.

## Error (Node 20.18.0 — before the backport)

```
Paperclip server failed to start.
require() of ES Module .../node_modules/@exodus/bytes/encoding-lite.js
from .../node_modules/html-encoding-sniffer/lib/html-encoding-sniffer.js not supported.
```

A different ESM-only dependency (`@exodus/bytes`) is `require()`'d from CommonJS code.

## Error (Node 20.18.0 + `--experimental-require-module`)

```
Paperclip server failed to start.
Maximum call stack size exceeded
```

## Error (Node 18.x)

Package requires `>=20` (`engines` field), and many dependencies also require `>=20`.

## Node Versions Tested

| Node Version | Flag | Result |
|---|---|---|
| 22.18.0 | — | `cssstyle` top-level await error |
| 20.20.2 | — | `cssstyle` top-level await error |
| 20.20.2 | `--experimental-require-module` | Same `cssstyle` error |
| 20.18.0 | — | `@exodus/bytes` ESM error |
| 20.18.0 | `--experimental-require-module` | Stack overflow |
| 18.20.8 | — | Unsupported engine (requires >=20) |

## Steps to Reproduce

```bash
npm install -g paperclipai
paperclipai onboard --yes
paperclipai run
```

Doctor checks pass (9/9), then the server immediately fails.

## Root Cause

The dependency chain appears to be: `paperclipai` → `jsdom@28` → `cssstyle@6` → `@asamuzakjp/css-color@5` (ESM with top-level `await`). Similarly, `html-encoding-sniffer` → `@exodus/bytes` (ESM-only).

These packages are loaded via `require()` from CommonJS contexts, which is not supported in any current Node.js version that also meets the `>=20` engine requirement.

## Expected Behavior

`paperclipai run` should start the server successfully after all doctor checks pass.

## Possible Fixes

- Pin `jsdom` to v24.x or v25.x (before the ESM-only dependency migration)
- Replace `jsdom` with `linkedom` or `happy-dom` if full DOM emulation isn't needed
- Bundle the ESM dependencies with a tool like `tsup` or `esbuild` to produce CJS-compatible output

levineam:
## Thinking Path

> - Paperclip ships as a CLI that users install globally via npm/npx (`paperclipai run`)
> - The CLI starts an Express server that bundles dependencies including `jsdom` (used for SVG sanitization in `server/src/routes/assets.ts`)
> - `jsdom@28` (recently bumped) pulls in `cssstyle@6`, which requires `@asamuzakjp/css-color@5`
> - `@asamuzakjp/css-color@5` is an ESM-only package (`"type": "module"`) with top-level `await`
> - Node 20.19+ and 22+ enforce that `require()` cannot load ESM modules with top-level `await`
> - This causes the server to crash immediately on startup for all supported Node versions
> - `jsdom@27` uses `cssstyle@5` which depends on `@asamuzakjp/css-color@4`, which ships a proper CJS build (`dist/cjs/index.cjs`)
> - Downgrading to `^27.0.0` resolves the crash without changing the public API used in assets.ts

## Root Cause

```
jsdom@28 → cssstyle@6 → @asamuzakjp/css-color@5 (ESM-only, top-level await)
                                     ↑ require() fails on Node 20.19+ / 22+

jsdom@27 → cssstyle@5 → @asamuzakjp/css-color@4 (has dist/cjs/index.cjs ✓)
```

## What Changed

- `server/package.json`: `jsdom` `^28.1.0` → `^27.0.0`, `@types/jsdom` `^28.0.0` → `^27.0.0`
- `pnpm-lock.yaml`: updated lockfile

No source code changes — the JSDOM API used in `assets.ts` is identical between v27 and v28.

## Verification

```bash
# Clean install + start server
npm install -g paperclipai
paperclipai run
# Expected: server starts cleanly on Node 20.19+, 22+, 24+

# Typecheck
```

## Risks

- jsdom@27 vs @28: no breaking changes in the API surface used (`new JSDOM(html)`, `window`, `querySelectorAll`, `outerHTML`). The downgrade is intentionally minimal — we can re-upgrade once cssstyle resolves the ESM/CJS issue upstream.

Fixes #2917
Fixes #2916
Fixes #2914
Fixes #2900

### DIFF
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -481,7 +481,7 @@ importers:
         version: 3.0.1(ajv@8.18.0)
       better-auth:
         specifier: 1.4.18
-        version: 1.4.18(drizzle-kit@0.31.9)(drizzle-orm@0.38.4(@electric-sql/pglite@0.3.15)(@types/react@19.2.14)(kysely@0.28.11)(pg@8.18.0)(postgres@3.4.8)(react@19.2.4))(pg@8.18.0)(react-dom@19.2.4(react@19.2.4))(react@19.2.4)(vitest@3.2.4(@types/debug@4.1.12)(@types/node@24.12.0)(jiti@2.6.1)(jsdom@28.1.0(@noble/hashes@2.0.1))(lightningcss@1.30.2)(tsx@4.21.0))
+        version: 1.4.18(drizzle-kit@0.31.9)(drizzle-orm@0.38.4(@electric-sql/pglite@0.3.15)(@types/react@19.2.14)(kysely@0.28.11)(pg@8.18.0)(postgres@3.4.8)(react@19.2.4))(pg@8.18.0)(react-dom@19.2.4(react@19.2.4))(react@19.2.4)(vitest@3.2.4(@types/debug@4.1.12)(@types/node@24.12.0)(jiti@2.6.1)(jsdom@27.4.0(@noble/hashes@2.0.1))(lightningcss@1.30.2)(tsx@4.21.0))
       chokidar:
         specifier: ^4.0.3
         version: 4.0.3
@@ -507,8 +507,8 @@ importers:
         specifier: 0.1.1
         version: 0.1.1
       jsdom:
-        specifier: ^28.1.0
-        version: 28.1.0(@noble/hashes@2.0.1)
+        specifier: ^27.0.0
+        version: 27.4.0(@noble/hashes@2.0.1)
       multer:
         specifier: ^2.0.2
         version: 2.0.2
@@ -541,8 +541,8 @@ importers:
         specifier: ^5.0.0
         version: 5.1.1
       '@types/jsdom':
-        specifier: ^28.0.0
-        version: 28.0.0
+        specifier: ^27.0.0
+        version: 27.0.0
       '@types/multer':
         specifier: ^2.0.0
         version: 2.0.0
@@ -575,7 +575,7 @@ importers:
         version: 6.4.1(@types/node@24.12.0)(jiti@2.6.1)(lightningcss@1.30.2)(tsx@4.21.0)
       vitest:
         specifier: ^3.0.5
-        version: 3.2.4(@types/debug@4.1.12)(@types/node@24.12.0)(jiti@2.6.1)(jsdom@28.1.0(@noble/hashes@2.0.1))(lightningcss@1.30.2)(tsx@4.21.0)
+        version: 3.2.4(@types/debug@4.1.12)(@types/node@24.12.0)(jiti@2.6.1)(jsdom@27.4.0(@noble/hashes@2.0.1))(lightningcss@1.30.2)(tsx@4.21.0)
 
   ui:
     dependencies:
@@ -705,6 +705,9 @@ packages:
 
   '@antfu/install-pkg@1.1.0':
     resolution: {integrity: sha512-MGQsmw10ZyI+EJo45CdSER4zEb+p31LpDAFp2Z3gkSd1yqVZGi0Ebx++YTEMonJy4oChEMLsxZ64j8FH6sSqtQ==}
+
+  '@asamuzakjp/css-color@4.1.2':
+    resolution: {integrity: sha512-NfBUvBaYgKIuq6E/RBLY1m0IohzNHAYyaJGuTK79Z23uNwmz2jl1mPsC5ZxCCxylinKhT1Amn5oNTlx1wN8cQg==}
 
   '@asamuzakjp/css-color@5.0.1':
     resolution: {integrity: sha512-2SZFvqMyvboVV1d15lMf7XiI3m7SDqXUuKaTymJYLN6dSGadqp+fVojqJlVoMlbZnlTmu3S0TLwLTJpvBMO1Aw==}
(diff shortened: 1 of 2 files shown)


levineam (commented):
The Greptile review references heartbeat reconciliation changes that were at an earlier commit (`4e80dabd`). Those changes have since been split into a separate PR — the current diff contains only the jsdom downgrade (`pnpm-lock.yaml` and `server/package.json`). The description is now accurate: no source code changes in this PR.

PR #2987, opened 2026-04-07
Fix issue thread typing lag

devinfoley:
## Thinking Path

> - Humans oversee agent work through the web UI, primarily through issue detail pages with live comment threads
> - Issue detail pages poll for run state on a short interval so humans can watch active runs in real time
> - But the polling was configured too aggressively — every 2 s regardless of whether a run was active — causing constant top-level rerenders
> - The `LiveRunWidget` embedded in the issue header ran its own parallel poll, doubling the rerender pressure
> - The `CommentThread` and `MarkdownEditor` had no memoization, so every top-level rerender cascaded to the composer
> - This made typing in the comment editor feel laggy, especially on long threads with many DOM nodes
> - This pull request slows idle polling, eliminates the duplicate widget poll, and memoizes the expensive subtrees
> - The benefit is that the comment composer stays responsive while the page is idle, and the UI only does real work when a run is actively executing

## What Changed

- **`IssueDetail.tsx`** — Idle poll interval slowed from 2 s to 30 s; drops to 3 s only when a run is actively executing. `commentsSection` and `detailContent` wrapped in `useMemo` to prevent top-level rerenders from cascading to the composer. `liveRunSlot` memoized so the `React.memo` on `CommentThread` is not defeated by a new JSX reference on every render.
- **`LiveRunWidget.tsx`** — Accepts an optional `runData` prop so the embedded header widget can reuse already-fetched run state from the parent instead of running its own parallel poll.
- **`CommentThread.tsx`** — Wrapped in `React.memo` so it only re-renders when its own props change, not on every parent poll cycle.
- **`MarkdownEditor.tsx`** — Mention-observer ref init moved outside the render effect, eliminating unnecessary observer recreation on every keystroke.

## Verification

```bash
pnpm -r typecheck   # passed
pnpm test:run       # passed
pnpm build          # passed
```

**Manual:**
1. Open an issue with a long comment thread (10+ comments)
2. Click into the comment composer and type continuously
3. Before this fix: noticeable input lag / dropped keystrokes on every poll cycle
4. After this fix: editor stays fully responsive; polling only wakes up when a run is active

> Note: this is a responsiveness/performance fix — there is no visible before/after screenshot, but the lag is immediately perceptible when typing in the editor under the old polling cadence.

## Risks

Low risk. The polling interval change is conservative (30 s idle vs. 2 s before) and drops back to 3 s the moment a run becomes active, so live-run visibility is not materially degraded. The memoization boundaries are narrow and only skip rerenders when props are stable. No API changes, no schema changes, no data migrations.

## Checklist

- [x] If this change affects the UI, I have included before/after screenshots *(performance fix — no visual diff; manual repro steps above)*

### DIFF
--- a/ui/src/components/CommentThread.tsx
+++ b/ui/src/components/CommentThread.tsx
@@ -1,4 +1,4 @@
-import { memo, useEffect, useMemo, useRef, useState, type ChangeEvent } from "react";
+import { memo, useCallback, useEffect, useMemo, useRef, useState, type ChangeEvent } from "react";
 import { Link, useLocation } from "react-router-dom";
 import type {
   Agent,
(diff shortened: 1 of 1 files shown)


devinfoley:
@greptile review

PR #3070, opened 2026-04-07
fix: auto-restart dev server in watch mode after unexpected SIGTERM

bartrtl:
## Summary

- Adds auto-restart with linear backoff (up to 5 retries) in watch mode when the child process dies from an unexpected SIGTERM or non-zero exit
- SIGINT (Ctrl+C) and clean exits (code 0) still terminate immediately as before
- Child exit handler in `startServerChild()` defers to the restart loop in watch mode instead of racing with `exitForSignal()`

**Problem:** When the `dev:watch` process received an external SIGTERM (terminal disruption, macOS sleep/wake), the entire process tree died permanently with no recovery, causing extended server downtime.

## Test plan

- [ ] Run `pnpm run dev` in watch mode, verify normal file-change restarts still work
- [ ] Send SIGTERM to the child process (`kill <pid>`), verify auto-restart kicks in with backoff logging
- [ ] Send SIGINT (Ctrl+C), verify clean immediate exit
- [ ] Send SIGTERM 6+ times rapidly, verify it gives up after max retries

## Model Used

Claude Opus 4.6

### DIFF
--- a/scripts/dev-runner.ts
+++ b/scripts/dev-runner.ts
@@ -543,6 +543,10 @@ async function startServerChild() {
       if (restartInFlight || expected || shuttingDown) {
         return;
       }
+      // In watch mode, the outer restart loop handles unexpected deaths.
+      if (mode === "watch") {
+        return;
+      }
       if (signal) {
         exitForSignal(signal);
         return;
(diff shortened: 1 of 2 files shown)


stefanocampana (approved):
## PR Review: #3070 — fix: auto-restart dev server in watch mode after unexpected SIGTERM

**Overall**: APPROVE
**Risk Level**: Low

### Summary
Adds auto-restart with exponential backoff (max 5 attempts) for the dev server in watch mode when the child process dies unexpectedly (SIGTERM, non-zero exit). Also adds graceful HTTP server shutdown before database teardown.

### Feedback
- Restart logic correctly distinguishes SIGINT (intentional) and clean exit (code 0) from unexpected deaths
- Linear backoff (2s × attempt) is reasonable for dev mode
- HTTP server `close()` with 5s grace period before embedded Postgres shutdown prevents request-in-flight errors
- The `startServerChild` guard in `child.on("close")` for watch mode prevents the original exit handler from racing with the outer restart loop — good

Clean DX improvement for local development.

PR #3071, opened 2026-04-07
fix(worktree): prevent repair function from poisoning default instance config

octasoft-ltd:
## Thinking Path

> - Agents run in git worktrees for workspace isolation, each with their own config and ports
> - The worktree config repair function (`maybeRepairLegacyWorktreeConfigAndEnvFiles`) ensures worktree configs have isolated paths and non-colliding ports
> - But when a worktree lacks its own `.paperclip/config.json`, config resolution falls back to the default instance at `~/.paperclip/instances/default/config.json`
> - The repair function then rewrites that shared config with worktree-isolated paths (temp dirs, bumped ports) and writes `PAPERCLIP_IN_WORKTREE=true` into the default `.env`
> - This creates a self-perpetuating poisoning loop: on next server start, dotenv loads the poisoned `.env`, triggering the repair again on the wrong config
> - This pull request adds a guard that detects when the resolved config path is inside the paperclip home directory and bails instead of rewriting shared state
> - The benefit is the default instance config is never corrupted by worktree repair, preventing wrong-port starts, fresh-setup prompts, and cross-database agent confusion

## What Changed

- **Guard in `resolveWorktreeRuntimeContext`** (`server/src/worktree-config.ts`): After resolving configPath, checks if it's inside the paperclip home directory (env-overridden or hardcoded default `~/.paperclip`). If so, returns `null` to prevent the repair function from targeting shared state.
- **Regression test** (`server/src/__tests__/worktree-config.test.ts`): New test case simulating a worktree without local config, verifying the repair function bails and leaves the default instance config and `.env` untouched.
- **Test env cleanup**: Two existing tests ("avoids sibling worktree ports" and "rebalances duplicate ports") were missing `delete process.env.PAPERCLIP_CONFIG` / `PAPERCLIP_INSTANCE_ID` / `PAPERCLIP_HOME` / `PAPERCLIP_CONTEXT`, causing failures on machines with Paperclip installed (ambient env vars leaked into tests). Added matching cleanup to both.

## Verification

- All 6 worktree-config tests pass locally (previously 2 were failing due to env leakage):
  ```
  pnpm vitest run server/src/__tests__/worktree-config.test.ts
  ```
- No new typecheck errors introduced (pre-existing `TS2774` in `routes/issues.ts:1389` is unrelated)

## Risks

- Low risk. The guard is a strict bail-out that only activates when `PAPERCLIP_IN_WORKTREE=true` AND the config path resolves inside the home directory — a scenario that was always a misconfiguration. Normal worktree flows (with local `.paperclip/config.json`) are unaffected.

## Model Used

- **Provider:** Anthropic Claude
- **Model:** Claude Opus 4.6 (`claude-opus-4-6`)
- **Context:** 1M tokens
- **Capabilities:** Tool use, code execution, systematic debugging
- **Environment:** Claude Code CLI on Windows 11

## Checklist

### DIFF
--- a/server/src/__tests__/worktree-config.test.ts
+++ b/server/src/__tests__/worktree-config.test.ts
@@ -480,6 +480,10 @@ describe("worktree config repair", () => {
     process.env.PAPERCLIP_IN_WORKTREE = "true";
     process.env.PAPERCLIP_WORKTREE_NAME = "PAP-884-ai-commits-component";
     process.env.PAPERCLIP_WORKTREES_DIR = isolatedHome;
+    delete process.env.PAPERCLIP_HOME;
+    delete process.env.PAPERCLIP_INSTANCE_ID;
+    delete process.env.PAPERCLIP_CONFIG;
+    delete process.env.PAPERCLIP_CONTEXT;
 
     const result = maybeRepairLegacyWorktreeConfigAndEnvFiles();
     const repairedConfig = JSON.parse(await fs.readFile(configPath, "utf8"));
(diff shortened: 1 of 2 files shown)


stefanocampana (approved):
## PR Review: #3071 — fix(worktree): prevent repair function from poisoning default instance config

**Overall**: APPROVE
**Risk Level**: Low

### Summary
Prevents the worktree config repair function from accidentally rewriting the shared default instance config when a worktree is misconfigured (missing its own `.paperclip/config.json`). The guard checks if the resolved config path falls inside the Paperclip home directory and bails early.

### Feedback
- `isPathInside` guards against both env-overridden and hardcoded default home paths — covers all resolution fallbacks
- New test case (`refuses to repair when config resolves to default home instance`) directly validates the poisoning scenario
- Existing tests updated to clean env state (`delete process.env.PAPERCLIP_HOME`, etc.) — this is the right way to isolate worktree tests
- Return `null` causes the caller to skip repair entirely, which is the safe default

Good defensive fix — config poisoning could silently break all other agents on the same host.

octasoft-ltd:
@greptileai review

octasoft-ltd:

octasoft-ltd:
@greptileai Rebased onto the latest `master` and squashed into a single commit. No logic changes. Please re-review.

PR #3073, opened 2026-04-07
Add heartbeat circuit breaker and execution timeout

natelion:
## Summary
- **Execution timeout**: 15-minute `Promise.race` around `adapter.execute()` — runs can't hang forever when Claw is down
- **Circuit breaker**: tracks consecutive failures per agent in `agentRuntimeState.stateJson` (no migration). After 3 failures, heartbeat interval backs off exponentially (2x→8x max). Resets on first success. `tickTimers` batch-loads all runtime state to avoid N+1
- **Ops agent systemic detection**: when 3+ agents fail simultaneously, diagnoses and restarts Claw BEFORE resetting individual agents. Stuck threshold reduced from 2h to 30min.

## Context
On 2026-04-06, Claw crash-looped and Paperclip agents kept firing doomed heartbeat runs against the broken API, wasting tokens and staying stuck for 21+ hours. The circuit breaker backs off automatically, and the execution timeout prevents indefinite hangs.

## Test plan
- [ ] Set `stateJson.circuitBreaker.consecutiveFailures` to 5 via DB, verify `tickTimers` applies backed-off interval
- [ ] Simulate 3 consecutive adapter failures, verify circuit breaker trips and logs warning
- [ ] Run a successful heartbeat after trip, verify breaker resets and logs info
- [ ] Set 3+ agents to error state, trigger ops agent, verify it calls `claw_restart` first

### DIFF
--- a/agents/instructions/brand-lead.md
+++ b/agents/instructions/brand-lead.md
@@ -1,0 +1,26 @@
+# Brand Lead — {{BRAND}}
+
+You are the Brand Lead for **{{BRAND}}**. You own all activity for this brand and are accountable for its growth and consistency.
+
+## Your Responsibilities
+
+1. **Brand health check** — review the current state of social media calendar and lead gen pipeline for {{BRAND}}.
+2. **Gap identification** — identify what's missing, behind, or needs action this week.
+3. **Coordination** — ensure the Social Agent and Lead Gen Agent for {{BRAND}} have what they need.
+4. **Reporting** — post a brief brand status update to `#olympus-zeus`.
+
+## Tools Available
+
+- `claw-social` — check calendar status, content gaps, engagement for {{BRAND}}
+- `claw-lead-gen` — check pipeline, campaign status, lead volume for {{BRAND}}
+- `slack` — post updates
+
+## Scope
+
+You are only responsible for **{{BRAND}}**. Do not touch other brands' data or campaigns.
+
+When using claw-social or claw-lead-gen tools, always filter/scope to {{BRAND}}.
+
+## Posting
+
+Post your brand status update to `#olympus-zeus` via `slack_post_message`. Keep it to 3-5 bullet points.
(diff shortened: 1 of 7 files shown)


stefanocampana (changes requested):
## PR Review: #3073 — Add heartbeat circuit breaker and execution timeout

**Overall**: REQUEST_CHANGES
**Risk Level**: High

### Summary
Two valuable reliability features — a 15-minute execution timeout and per-agent circuit breaker with exponential backoff — plus company-specific seed data that should not be in the upstream repo.

### Must Fix

**1. Remove company-specific files from the upstream repo**
The `agents/instructions/` directory (brand-lead.md, chief-of-staff.md, lead-gen-agent.md, ops-agent.md, social-agent.md) and the Truwitz seed script contain company-specific configuration (brand names, Slack channels, pipeline details). These should live in a separate repo or be loaded via Paperclip's company import/export. Please split this into a separate PR for your deployment.

**2. `Promise.race` doesn't terminate the adapter process**
The 15-minute timeout rejects the promise, but the underlying `adapter.execute()` call (and its spawned child process) keeps running. After timeout, the agent will be marked failed while the process is still alive, potentially leading to:
- Orphaned processes consuming resources
- Race conditions if a new run starts while the old one is still running
Consider using an `AbortController` or the adapter's own cancellation mechanism to actually stop the process.

### Feedback (non-blocking)

- Circuit breaker state in `stateJson` is a good no-migration approach
- `tickTimers` batch-loading runtime state avoids N+1 — nice
- 15-minute timeout should be configurable per agent via `runtimeConfig.heartbeat.executionTimeoutSec` rather than a global constant
- The `backoffMultiplier` doubles each failure after threshold but resets to 1 on success — correct behavior
- `allRuntimeStates` loads state for ALL agents every tick — fine at moderate scale but consider filtering by active agents if this grows

### Suggestions
- Split into two PRs: (1) circuit breaker + timeout (heartbeat.ts changes) and (2) Truwitz company setup (separate repo)
- Add a test for the circuit breaker state transitions

PR #3076, opened 2026-04-07
fix(server): clear stale execution lock during checkout

luanxu-dev:
## Summary
- clear stale `executionRunId` and `checkoutRunId` when the linked run is terminal
- keep checkout conflict behavior for active/non-terminal runs
- add regression tests for stale-lock cleanup and conflict paths

## Validation
- pnpm --filter @paperclipai/server exec vitest run src/__tests__/issues-service.test.ts
- pnpm --filter @paperclipai/server typecheck

## Context
- related internal issue: BLO-5331

### DIFF
--- a/server/src/__tests__/issues-service.test.ts
+++ b/server/src/__tests__/issues-service.test.ts
@@ -8,6 +8,7 @@ import {
   companies,
   createDb,
   executionWorkspaces,
+  heartbeatRuns,
   instanceSettings,
   issueComments,
   issueInboxArchives,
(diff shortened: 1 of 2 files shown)


stefanocampana (commented):
## CTO Review — Comment

This addresses a real pain point (stale execution locks blocking checkout), and the test coverage is solid.

Would like to see the full diff before approving — the test shows the expected behavior (clear stale lock, retry checkout) but I want to verify the production code handles edge cases like concurrent checkout attempts during lock clearing.

Looks promising, will follow up with a full review.

PR #3080, opened 2026-04-07
fix(skills): use correct raw URL format for Gitea and fix root-level SKILL.md inventory

### LINKED ISSUE #3054: Skill import from Gitea: wrong raw URL format causes empty SKILL.md
**Bug Summary**
When importing skills from a Gitea instance (self-hosted), Paperclip fetches the git tree successfully via the /api/v3/ endpoint (which Gitea supports), but then uses the wrong raw file URL format to download SKILL.md content, resulting in a 404 and an almost-empty stub being saved.

**Environment**
- Paperclip server: @paperclipai/server@2026.403.0
- Gitea instance: self-hosted Gitea

**Steps to Reproduce**
1. Host a skill repo on a Gitea instance with a SKILL.md (32KB+), scripts/, and references/ directories
2. Import the skill via URL: https://my-gitea.example.com/org/skill-name.git
3. Observe that only a stub SKILL.md (~283 bytes) is saved, and subdirectories are missing

**Root Cause**
In server/dist/services/github-fetch.js, the resolveRawGitHubUrl() function builds:
https://{hostname}/raw/{owner}/{repo}/{ref}/{file}
This format works for GitHub Enterprise, but not for Gitea. Gitea expects:
https://{hostname}/{owner}/{repo}/raw/branch/{ref}/{file}

**Proof**
# Paperclip's URL format -> 404
$ curl -s -o /dev/null -w "%{http_code}" "https://my-gitea.example.com/raw/org/skill-repo/main/SKILL.md" 404

# Correct Gitea format -> 200
$ curl -s -o /dev/null -w "%{http_code}" "https://my-gitea.example.com/org/skill-repo/raw/branch/main/SKILL.md" 200

**Secondary Issue**
When SKILL.md is at the repository root, path.posix.dirname("SKILL.md") returns ".", causing the inventory filter entry.startsWith("./") to miss all sibling directories (scripts/, references/). This means even if the raw URL worked, subdirectories would still be excluded from the file inventory.

**Suggested Fix**
1. Detect Gitea vs GitHub Enterprise (e.g., check for /api/v1/ endpoint which is Gitea-specific, vs /api/v3/)
2. Use the correct raw URL format per platform:
   - GitHub/GHE: https://{hostname}/raw/{owner}/{repo}/{ref}/{file}
   - Gitea: https://{hostname}/{owner}/{repo}/raw/branch/{ref}/{file}
3. Fix root-level SKILL.md inventory: when skillDir === ".", include all files instead of filtering by entry.startsWith("./")

**Workaround**
Manually clone the repo into ~/.paperclip/instances/default/skills/{company_id}/{slug}/ and update the company_skills table directly with correct markdown content and file_inventory.

Lempkey:
## Summary

Fixes #3054

Two bugs affecting skill import from self-hosted Gitea instances:

**Bug 1: Wrong raw file URL format for Gitea**

`resolveRawGitHubUrl()` generated GitHub-Enterprise-style URLs (`/raw/{owner}/{repo}/{ref}/{file}`) for all non-github.com hosts. Gitea expects a different format: `/{owner}/{repo}/raw/branch/{ref}/{file}`.

Fix: add `detectGitType(hostname)` which probes `/api/v1/version` (a Gitea-specific endpoint) with a 3s timeout and caches the result per hostname for the process lifetime. `resolveRawGitHubUrl()` accepts a new `gitType` parameter (`"github" | "ghe" | "gitea"`) and picks the correct URL format. All callers in `company-skills.ts` and `company-portability.ts` now detect and pass the type.

**Bug 2: Root-level `SKILL.md` breaks sibling file inventory**

When `SKILL.md` is at the repo root, `path.posix.dirname("SKILL.md")` returns `"."`. The inventory filter used `entry.startsWith("./")` which never matches real paths like `"scripts/foo.sh"`, so all sibling directories were excluded from the file inventory.

Fix: normalise `"."` to `""` and gate the filter on whether `skillDir` is non-empty. When empty (root-level skill), all paths are included. Fixed in all three affected locations: `readInlineSkillImports`, the GitHub import loop, and the portability package importer.

## Files changed

| File | Change |
|------|--------|
| `server/src/services/github-fetch.ts` | Add `detectGitType()` + `GitHostType`; update `resolveRawGitHubUrl()` signature |
| `server/src/services/company-skills.ts` | Thread `gitType` to both `resolveRawGitHubUrl` call sites; fix root-level inventory filter in two locations |
| `server/src/services/company-portability.ts` | Thread `gitType` to all five `resolveRawGitHubUrl` call sites; fix root-level inventory filter |
| `server/src/__tests__/github-fetch.test.ts` | 10 unit tests for `detectGitType` and `resolveRawGitHubUrl` |

## Test plan

- [ ] `resolveRawGitHubUrl` generates correct URLs for github.com, GHE, and Gitea
- [ ] `detectGitType` returns `"gitea"` when `/api/v1/version` returns 200
- [ ] `detectGitType` returns `"ghe"` when probe returns non-200 or throws
- [ ] `detectGitType` returns `"github"` for github.com without probing
- [ ] Existing company-skills and company-skills-routes tests pass (28/28)
- [ ] Manually: import a skill from a Gitea repo → `SKILL.md` and subdirectories are fetched correctly

### DIFF
--- a/server/src/__tests__/github-fetch.test.ts
+++ b/server/src/__tests__/github-fetch.test.ts
@@ -1,0 +1,75 @@
+import { afterEach, describe, expect, it, vi } from "vitest";
+import { detectGitType, resolveRawGitHubUrl } from "../services/github-fetch.js";
+
+// detectGitType caches results in module-level state — reset between tests by
+// reimporting the module fresh.  We clear the cache via the fetch mock instead.
+const mockFetch = vi.fn();
+vi.stubGlobal("fetch", mockFetch);
+
+afterEach(() => {
+  mockFetch.mockReset();
+  // Reset the module's internal cache by clearing all cached entries.
+  // We do this by re-running detectGitType with fresh mocks in each test.
+});
+
+describe("resolveRawGitHubUrl", () => {
+  it("returns raw.githubusercontent.com URL for github.com", () => {
+    const url = resolveRawGitHubUrl("github.com", "acme", "skills", "main", "SKILL.md", "github");
+    expect(url).toBe("https://raw.githubusercontent.com/acme/skills/main/SKILL.md");
+  });
+
+  it("returns GHE-style URL for GitHub Enterprise hosts", () => {
+    const url = resolveRawGitHubUrl("ghe.acme.com", "acme", "skills", "main", "docs/SKILL.md", "ghe");
+    expect(url).toBe("https://ghe.acme.com/raw/acme/skills/main/docs/SKILL.md");
+  });
+
+  it("returns Gitea-style URL for Gitea hosts", () => {
+    const url = resolveRawGitHubUrl("git.acme.com", "acme", "skills", "main", "SKILL.md", "gitea");
+    expect(url).toBe("https://git.acme.com/acme/skills/raw/branch/main/SKILL.md");
+  });
+
+  it("returns Gitea-style URL for nested paths on Gitea", () => {
+    const url = resolveRawGitHubUrl("git.acme.com", "acme", "skills", "main", "scripts/run.sh", "gitea");
+    expect(url).toBe("https://git.acme.com/acme/skills/raw/branch/main/scripts/run.sh");
+  });
+
+  it("strips leading slash from filePath", () => {
+    const url = resolveRawGitHubUrl("git.acme.com", "acme", "skills", "main", "/SKILL.md", "gitea");
+    expect(url).toBe("https://git.acme.com/acme/skills/raw/branch/main/SKILL.md");
+  });
+
+  it("defaults to GHE format when gitType is omitted", () => {
+    const url = resolveRawGitHubUrl("ghe.acme.com", "acme", "skills", "main", "SKILL.md");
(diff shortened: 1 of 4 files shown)


Lempkey:
Addressed both Greptile P1s:

**1. `readLocalSkillImports` root-level inventory bug** (`company-skills.ts:959-964`)
`path.posix.dirname("SKILL.md")` returns `"."`, so the `startsWith("./")` filter never matched and `entry.slice(2)` produced wrong relative paths for all sibling files. Applied the same `rawDir === "." ? "" : rawDir` normalisation already used at line 739, with matching ternary guards in the filter and map.

**2. Gitea URL breaks for pinned commit SHAs** (`github-fetch.ts`)
`/raw/branch/{ref}/` is Gitea's branch-only endpoint — it 404s on commit SHAs. Switched to `/raw/{ref}/` which Gitea supports uniformly for branches, tags, and commit SHAs. Updated the three existing Gitea URL test assertions and added a dedicated commit-SHA test case.

Issue #39, opened 2026-03-05
feat: skip LLM invocation on timer heartbeats when agent has no assigned tasks


jaycoolh:
## Problem

When an agent's heartbeat fires on a timer (scheduled interval), the system always invokes the LLM — even if the agent has zero assigned tasks. The agent burns tokens just to call `/api/agents/me`, check assignments, find nothing, and exit.

For agents like an IC engineer who are purely reactive (only useful when there's work assigned), this is wasted spend on every empty heartbeat.

## Current Behavior

`tickTimers` (in `server/src/services/heartbeat.ts`) checks each agent's interval and enqueues a wakeup if enough time has elapsed. `enqueueWakeup` already has skip logic for disabled heartbeats and issue execution locks, but there is no pre-flight check for whether the agent actually has work to do.

The flow today:

```
Cron tick → interval elapsed? → yes → enqueue wakeup → invoke LLM → agent checks tasks → nothing → exit
```

## Proposed Solution

Add a pre-flight task check in `enqueueWakeup` for timer-triggered wakes, before creating the wakeup request:

```
if source === "timer":
  1. Query assigned tasks for this agent (status in todo, in_progress, blocked)
  2. No tasks → write a "skipped" wakeup request with reason, return null
  3. Has tasks → proceed with LLM invocation as normal
```

This would fit naturally alongside the existing skip logic at lines ~1489-1496.

**Event-triggered wakes** (assignment, @-mention, approval, manual invoke) would be unaffected — those already have a known reason to wake up.

### Per-agent configuration

This could be a per-agent opt-in flag in the heartbeat policy, e.g.:

```json
{
  "heartbeat": {
    "enabled": true,
    "intervalSec": 3600,
    "skipIfNoAssignments": true
  }
}
```

This keeps the behavior backward-compatible (defaults to `false`) and lets operators choose per agent. A CEO might want periodic wakes for self-directed work (planning, reviews), while an IC engineer should only wake when there's a task.

## Trade-offs

| | Always invoke (current) | Pre-flight check (proposed) |
|---|---|---|
| **Cost** | Burns tokens on empty heartbeats | Only pays for useful work |
| **Proactive behavior** | Agent can self-initiate work (planning, memory synthesis) | Agent only wakes for assigned tasks |
| **Complexity** | Uniform model | One extra query per tick (trivial cost) |

## Scope

- Add `skipIfNoAssignments` to heartbeat policy parsing (`parseHeartbeatPolicy`)
[...]

PR #3093, opened 2026-04-08
feat: save 80%+ on idle agent token costs — skip heartbeats when there's nothing to do #1

speiron:
## Thinking Path

> - Agents wake up on a heartbeat timer and invoke their LLM adapter to check for work
> - Every single heartbeat calls the LLM — even when the agent has zero issues, zero comments, zero approvals
> - An idle agent with a 5-minute interval burns ~$0.05/heartbeat × 288/day = **~$14/day doing literally nothing**
> - Multiply by 20 agents and that's **~$280/day in wasted tokens** just to hear "nothing to do" 288 times
> - This PR adds a 3-query SQL preflight that short-circuits the heartbeat before the adapter ever runs
> - Cost of the preflight: ~5ms of SQL. Cost of what it replaces: ~$0.05 and 5-30 seconds of LLM time
> - It's opt-in (default off), per-agent configurable, and fully auditable in `agent_wakeup_requests`

## The Problem

Every heartbeat timer tick invokes the full LLM adapter pipeline — session resolution, workspace setup, adapter execution — even when there is **zero pending work**. The agent spends tokens to think about whether it has something to do, concludes "no", and the tokens are gone.

This is the single biggest source of unnecessary token spend in Paperclip at scale.

### Real-world numbers

| Agents | Interval | Empty heartbeats/day | Wasted tokens/day | Wasted $/day (Sonnet) |
|--------|----------|---------------------|--------------------|-----------------------|
| 5 | 5 min | ~1,440 | ~2.8M input | ~$17 |
| 20 | 5 min | ~5,760 | ~11.5M input | ~$69 |
| 50 | 5 min | ~14,400 | ~28.8M input | ~$173 |

## The Solution

Three `SELECT ... LIMIT 1` queries that answer: **"does this agent have any reason to wake up?"**

```
1. Any active issues assigned? (todo, in_progress, in_review, blocked)
2. Any new comments from others since last heartbeat?
3. Any pending approvals?
```

If all three return empty → skip the heartbeat entirely. Zero tokens. The wakeup request is recorded as `skipped` with reason `preflight.no_pending_work` for full auditability.

**Cost comparison per empty heartbeat:**

| | Without preflight | With preflight |
|---|---|---|
| DB queries | 0 (goes straight to adapter) | 1–3 (LIMIT 1, short-circuit) |
| LLM tokens | ~2K+ input, ~100+ output | **0** |
| Latency | 5–30 seconds | **5–15 milliseconds** |
| Cost | ~$0.01–0.05 | **$0** |

## What Changed

- **`server/src/services/heartbeat.ts`**:
  - `hasPendingWork(agent)` — three short-circuit SQL queries that check for active issues, new comments, and pending approvals. Returns on the first positive match.
  - Preflight check block in `enqueueWakeup()` — sits right after the existing policy checks, follows the exact same `writeSkippedRequest()` pattern
  - `preflightEnabled` field in `parseHeartbeatPolicy()` — per-agent with instance env var fallback
  - `heartbeat.preflight.skipped` live event for real-time observability
- **`server/src/config.ts`**: `HEARTBEAT_PREFLIGHT_ENABLED` env var + `heartbeatPreflightEnabled` config field
- **`server/src/__tests__/heartbeat-preflight.test.ts`**: 15 test cases covering every scenario

### Safety guarantees

| Guarantee | How |
|-----------|-----|
| No behavioral change for existing users | Default is `false` — strictly opt-in |
| On-demand wakeups always work | Only `timer` and `automation` sources are checked |
| Assignment wakeups always work | `assignment` source is never filtered |
| Targeted wakeups always work | Any wakeup with `issueId`, `commentId`, or `issue_comment_mentioned` bypasses preflight |
| Fully auditable | Every skip is recorded in `agent_wakeup_requests` with reason |
| Per-agent control | Agents doing proactive work can set `preflightEnabled: false` |

## Verification

```bash
# Run the new tests
pnpm --filter server test heartbeat-preflight

# Manual verification:
# 1. Set HEARTBEAT_PREFLIGHT_ENABLED=true
# 2. Create an agent with intervalSec: 60 and no assigned issues
# 3. Wait for timer tick
# 4. Check agent_wakeup_requests: status='skipped', reason='preflight.no_pending_work'
# 5. Assign a 'todo' issue to the agent
# 6. Next timer tick proceeds normally and creates a heartbeat run
```

## Risks

- **Low risk** — opt-in feature, default off, no migration needed
- **Edge case**: Agents that do proactive scanning without assigned issues would be skipped. Mitigation: set `preflightEnabled: false` for those agents.
- **Query perf**: All queries hit indexed columns (`assignee_agent_id`, `company_id`, `status`) with `LIMIT 1` — negligible vs the adapter invocation they replace

## Model Used

- **Provider**: Anthropic
- **Model**: Claude Opus 4 (`claude-opus-4-0-20250514`)
- **Context window**: 200K tokens
- **Capabilities**: Extended thinking, tool use, multi-file code generation
- **Usage**: Feature design (spec + approach evaluation), implementation, and test authoring

## Checklist


cryppadotta:
Just turn off timed heartbeats if you don't want them

aronprins:
Great PR!

Quick train of thought on Heartbeat vs Routines

Its tempting to have routine replace the heartbeat, so this is a great solution if deeply integrated and native to Paperclip's inner workings.

@cryppadotta I dont know if this PR's solution is the way, but this temptation needs to be intercepted if possible.

As mentioned on X here this would allow the heartbeat to remain on yet keep the agents proactive without burning through tokens when not needed:

https://x.com/cristianotorres/status/2043005212772765735?s=46&t=fYu9ia_o9oX-c7IiwlaSWQ

alvarosanchez:
If agents wake on assignment, and considering routines exist, I wonder what is the use case of heartbeats

speiron:
Great question, @alvarosanchez . Heartbeats cover several scenarios that assignments and routines alone don't:

1. **State changes outside Paperclip's event loop**: A teammate assigns an issue, leaves a comment, or submits an approval through the UI or API. The agent won't know about it unless it polls, or unless every single entry point in Paperclip fires a wakeup event. Heartbeats act as a reliable catch-all.

2. **Missed or failed event deliveries**: If an assignment webhook fails, a queue drops a message, or an integration doesn't fire the expected trigger, the agent still picks up the work on the next heartbeat tick. It's a safety net.

3. **Monitoring and autonomous behavior**: Some agents are configured to proactively scan for work, check project status, or follow up on stale issues. This requires periodic polling by design, not just reacting to explicit triggers.

4. **Routines are not heartbeats**: Routines are user-defined scheduled tasks with specific logic. Heartbeats are a low-level system mechanism that ensures the agent periodically evaluates its full context. Using routines as a heartbeat replacement would push infrastructure concerns into user configuration, which is fragile and harder to maintain.

The problem this PR solves is: when heartbeats are necessary (scenarios above), most ticks find nothing to do but still invoke the LLM. The preflight check keeps the polling active but skips the expensive LLM call when three fast SQL queries confirm there's no pending work.

ttt132:
Thanks for the prior work on this PR. I ran into the same idle-heartbeat cost problem while operating Paperclip with `codex_local`: Paperclip does resume the Codex session, but empty timer ticks still invoke `codex exec resume ...`, which can keep burning cached/input tokens and quota pressure even when there is no actionable work.

I would like to help get a smaller version of this merged upstream, scoped more narrowly than the current PR:

- timer wakes only (`source === "timer"`)
- default off, per-agent `runtimeConfig.heartbeat.idleSkip.enabled`
- no UI in the first PR
- no company budget cap, circuit breaker, session compaction, comment debounce, or routine changes
- skip only when the control plane can prove from its own DB that the agent has no active assigned work
- record an audited skipped wake request, but do not create a `heartbeat_runs` row or invoke the adapter
- advance the scheduler checkpoint on skip so a no-work agent does not repeatedly enqueue skipped ticks every scheduler loop

This would be a smaller successor/revival of the preflight idea here, with credit to #3093 and references to #39, #4965, and #5778.

Before I open a new PR, do maintainers prefer either of these paths?

1. Continue from / help revive this PR branch, if the original author is available.
2. Open a new smaller replacement PR focused only on timer idle-skip.

Happy to align with whichever path is easier to review and merge.

PR #3099, opened 2026-04-08
feat(openclaw-gateway): derive claimed api key path per agent

ahmadashfq:
## Summary
- derive the claimed API key path from `agentId` when no explicit `claimedApiKeyPath` is configured
- keep the existing main workspace path as the backward-compatible fallback
- add focused tests for explicit path, per-agent path, and main fallback behavior

## Why
Multi-agent OpenClaw deployments often store each agent's claimed API key in its own workspace directory. This keeps the default behavior unchanged while allowing adapters to resolve the correct per-agent path automatically.

## Testing
- `pnpm --filter @paperclipai/adapter-openclaw-gateway typecheck`
- `pnpm exec vitest run --config /tmp/paperclip-upstream-pr/vitest.openclaw-gateway.temp.mts`

### DIFF
--- a/packages/adapters/openclaw-gateway/src/server/execute.test.ts
+++ b/packages/adapters/openclaw-gateway/src/server/execute.test.ts
@@ -1,5 +1,23 @@
 import { describe, expect, it } from "vitest";
-import { resolveSessionKey } from "./execute.js";
+import { resolveClaimedApiKeyPath, resolveSessionKey } from "./execute.js";
+
+describe("resolveClaimedApiKeyPath", () => {
+  it("uses an explicit claimed API key path when configured", () => {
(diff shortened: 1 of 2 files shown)


stefanocampana (approved):
## CTO Review — Approve

Good fix for multi-agent OpenClaw gateway setups. Deriving per-agent workspace paths (`~/.openclaw/workspace-{agentId}/`) prevents key collision when multiple agents share the same OpenClaw instance.

- Explicit `claimedApiKeyPath` config takes precedence (good escape hatch)
- `main` agent falls back to the default path for backward compatibility
- Tests cover all three cases (explicit, per-agent, default)
- Function is now properly exported for testability

Clean implementation.

PR #3100, opened 2026-04-08
feat(web): add issue ID sort option to dashboard

michaelkd01:
## Thinking Path

> The issues list supports sorting by status, priority, title, created, and updated
> But there is no option to sort by issue ID (e.g. ANY-826)
> Issue IDs encode creation sequence as {PREFIX}-{number}
> Sorting by ID gives operators a stable chronological sort they intuitively expect
> This is a client-side-only change to the sort dropdown and comparator in IssuesList.tsx

## Problem

The issues list sort dropdown does not include an option to sort by issue ID. Operators managing large backlogs often want to see issues in creation order by ID number, which is not possible with the current sort options (status, priority, title, created, updated).

## Solution

Added "ID" as a sort option in `IssuesList.tsx`. The comparator extracts the numeric suffix from `issue.identifier` (everything after the last `-`) and sorts numerically rather than lexicographically, so `ANY-10` correctly sorts after `ANY-9`. Supports both ascending and descending sort direction.

## Changes

- `ui/src/components/IssuesList.tsx` — Added `"id"` to the `sortField` union type, added `["id", "ID"]` as the first entry in the sort dropdown options, and added a numeric sort comparator case that parses the identifier suffix

## Model Used

Claude Code (claude-opus-4-6)

## Risks

- Low risk. Client-side only, additive change. No server-side modifications, no new dependencies, no schema changes.
- Falls back gracefully: if an identifier does not contain `-`, `parseInt` returns `NaN` and the sort treats them as equal.

## Testing

- `npx tsc --noEmit` in `ui/` — clean, no errors
- Manually verified sort dropdown shows "ID" option and issues sort correctly by numeric identifier in both ascending and descending order

## Checklist

- [x] `pnpm typecheck` passes
- [x] No new dependencies added
- [x] Changes are backward compatible

### DIFF
--- a/ui/src/components/IssuesList.tsx
+++ b/ui/src/components/IssuesList.tsx
@@ -43,7 +43,7 @@ export type IssueViewState = {
   assignees: string[];
   labels: string[];
   projects: string[];
-  sortField: "status" | "priority" | "title" | "created" | "updated";
+  sortField: "id" | "status" | "priority" | "title" | "created" | "updated";
   sortDir: "asc" | "desc";
   groupBy: "status" | "priority" | "assignee" | "none";
   viewMode: "list" | "board";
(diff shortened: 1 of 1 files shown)


stefanocampana (approved):
## CTO Review — Approve

Clean UI enhancement. Adds "ID" sort option to the issues list, which is useful for chronological ordering.

- Sort extraction logic (`numFromId`) correctly handles the `PREFIX-NUMBER` identifier format
- Properly handles edge cases (null/undefined identifiers, missing dash)
- Minimal changes to the existing sort infrastructure

Ship it.

PR #3133, opened 2026-04-08
fix(server): respect catchUpPolicy skip_missed on scheduler tick

### LINKED ISSUE #3108: catchUpPolicy: skip_missed not respected — overdue routines fire on first scheduler tick after startup
## Summary

`catchUpPolicy: "skip_missed"` on a scheduled Routine is **not respected** by the embedded scheduler in `paperclipai@2026.325.0`. On the first scheduler tick after `Server listening on ...`, the scheduler enqueues catch-up runs for every Routine whose `nextRunAt` is in the past — even when the Routine's `catchUpPolicy` is explicitly set to `skip_missed`.

This is a real-world incident bug for us: it caused **four routines to fire simultaneously** after bringing Paperclip up following an outage, and would have sent a stale Weekly CEO Brief to the actual CEO if we hadn't killed Paperclip in time. We've now had to ship a container-level workaround to avoid silent failure modes on every machine restart and every `fly deploy`.

## Version

`paperclipai@2026.325.0`

## Environments observed

- macOS 15 arm64, `npx -y paperclipai@2026.325.0 run` (Mac bridge, 2026-04-07)
- Linux (`node:20-slim` on Debian bookworm, running in a Fly.io `shared-cpu-1x` microVM in Sydney) (2026-04-08)

Same behaviour in both.

## Expected behaviour

Given a Routine with a `schedule` trigger and `catchUpPolicy: "skip_missed"`:

- On startup (or any scheduler tick where the Routine's `nextRunAt` is in the past), the scheduler should **skip** the missed runs.
- It should update `nextRunAt` to the next future cron match.
- It should **not** enqueue any catch-up runs.

This is what "skip_missed" implies by name and by documentation.

## Actual behaviour

On the first scheduler tick after `Server listening ...` (typically ~30 s after startup with the default `tickIntervalMs: 30000`), the scheduler enqueues **one run for every Routine whose `nextRunAt` is in the past**, ignoring `catchUpPolicy: "skip_missed"`. Only routines whose `nextRunAt` is still in the future are left alone.

## Reproduction

1. Create a Routine with a cron trigger (e.g. `0 6 * * 1-5`), `catchUpPolicy: "skip_missed"`, and leave it running long enough for `nextRunAt` to advance normally.
2. Stop Paperclip (`SIGTERM` or machine shutdown).
3. Wait until the cron time of the Routine has passed (so `nextRunAt` is now in the past).
4. Start Paperclip again.
5. Observe that within ~30 seconds of `Server listening ...`, the Routine fires — even though `catchUpPolicy: "skip_missed"` should prevent it.

Same behaviour occurs even when pausing a Routine, waiting past `nextRunAt`, and un-pausing it: the un-pause triggers an immediate tick and the overdue Routine fires.

### Incident 1 — 2026-04-07 13:17 UTC

Our Unraid NAS (the planned Paperclip host) had been down for several days. We brought Paperclip up on a Mac as an emergency bridge. All 4 of our Routines had stale `nextRunAt` values from 2026-04-02. All 4 fired simultaneously at **+30 s after startup** (13:18:22 UTC), despite all 4 having `catchUpPolicy: "skip_missed"`. We killed Paperclip before the CMO agent could deliver a stale Weekly CEO Brief to the real CEO.

### Incident 2 — 2026-04-07 23:39 UTC (same day)

After re-enabling the routines, we paused them, stopped and restarted Paperclip, and un-paused them. At this point:

- The 2 **daily** routines (6 AM and 8 AM weekdays) had `nextRunAt` = 2026-04-07 20:00 UTC and 22:00 UTC respectively — both already in the past by 1.5–3.5 hours.
- The 2 **weekly** routines (Monday 5 AM and 9 AM) had `nextRunAt` = 2026-04-12 — still in the future.

**Only the 2 daily routines fired** on the next tick. The 2 weekly ones were left alone. This confirms the bug is specifically "overdue routines get caught up on startup/tick", not "all routines fire on every state change".

### Incident 3 — 2026-04-08 03:47 UTC (Fly.io migration)

Booting Paperclip on Fly.io with freshly-rsync'd data from the Mac. All 4 routines had `nextRunAt` values that were a few hours in the past. Same behaviour — without our entrypoint guard (described below), all 4 would have fired on the first tick.

## Why this matters

This is a silent, high-impact data-quality bug. Our specific failure mode:

- The Weekly CEO Brief is scheduled for Monday 5 AM / 9 AM. If Paperclip restarts on a Tuesday (after the cron time has passed), the brief fires for the same week it already fired for, generating a **duplicate weekly brief** for the CEO.
- The daily brief is even worse — it runs every weekday, so any restart during business hours will produce a duplicate and Abby / the distribution list gets the same brief twice.

For systems that trigger outbound emails, posts to external APIs, or any other non-idempotent side-effect, the current behaviour is **unsafe by default**.

## Workaround we shipped

In our Fly.io `docker-entrypoint.sh` we implemented a "race the scheduler" guard. It runs on every container start, including every `fly deploy` and every `fly machine restart`:

```bash
# Start Paperclip in background
npx paperclipai run &
PC_PID=$!

# Poll /api/health until it's ready (max 60s)
for i in $(seq 1 60); do
  if curl -fsS "http://127.0.0.1:3100/api/health" > /dev/null 2>&1; then break; fi
  sleep 1
done

# Immediately pause all routines and the orchestrator agent.
# This must complete BEFORE the ~30s scheduler tick, or the overdue
# routines will fire with real side-effects.
for RID in "$R1" "$R2" "$R3" "$R4"; do
  curl -fsS -X PATCH "http://127.0.0.1:3100/api/routines/$RID" \
    -d '{"status":"paused"}' > /dev/null
done
curl -fsS -X POST "http://127.0.0.1:3100/api/agents/$CMO_AGENT_ID/pause" > /dev/null

# Operator manually resumes routines after confirming state.
wait "$PC_PID"
```

In practice all 5 API calls complete in ~2 seconds, giving ~28 seconds of margin before the 30 s scheduler tick.

**This should not be the default operating posture.** The workaround is brittle (a slower machine could miss the window), operationally expensive (an operator has to manually resume routines after every restart), and it shouldn't be necessary in the first place given that `catchUpPolicy: "skip_missed"` is a documented option.

## Suggested fixes, in priority order

1. **Fix the actual bug** — `catchUpPolicy: "skip_missed"` should mean "skip missed runs on startup/tick, update `nextRunAt` to the next future cron match, don't enqueue catch-up runs." The current behaviour is a spec violation.

2. **Add a `--safe-mode` / `--no-scheduler` flag to `paperclipai run`** — so operators can boot the server without scheduler ticks for migrations, disaster recovery, and inspection. This gives us a supported escape hatch that doesn't rely on racing the scheduler.

3. **Add a `pauseSchedulerOnBoot: true` config option** in the instance `config.json` — a persistent version of the flag above.

4. **Emit a warning log line** when the scheduler catches up an overdue Routine — something like `scheduler: catching up overdue routine <id> (nextRunAt <time> was in the past)`. This would at least make the current behaviour visible in logs rather than silent.

(1) is the most correct fix. (2) or (3) would be acceptable mitigation while (1) is in progress. (4) is a safety belt regardless.

## Happy to help

If helpful, I can share the raw log lines from both incidents, or contribute a repro test and a fix PR. The relevant code paths I skimmed are in `@paperclipai/server/dist/services/` — specifically the scheduler loop and the routine tick handler.

Thanks for building Paperclip — the product is great, this is the only rough edge we've hit so far.

mvanhorn:
## Thinking Path

> - Routines are scheduled tasks that fire on cron expressions via the embedded scheduler
> - The scheduler has a `catchUpPolicy` field: `skip_missed` should skip overdue runs, `enqueue_missed_with_cap` should replay them
> - The tick handler in `routines.ts` defaults `runCount = 1` then only checks `enqueue_missed_with_cap`, so `skip_missed` routines still fire once when overdue
> - This caused 4 routines to fire simultaneously after an outage, nearly sending a stale Weekly CEO Brief to the CEO (#3108)
> - This PR adds the missing `skip_missed` branch so overdue routines advance `nextRunAt` without dispatching

## What Changed

- Added `skip_missed` check in `tickScheduledTriggers()` (server/src/services/routines.ts:1436) that sets `runCount = 0` and logs a warning when a routine's `nextRunAt` is in the past
- The existing `enqueue_missed_with_cap` path is now an `else if` branch, preserving current behavior for that policy
- `nextRunAt` is still advanced to the next future cron tick via `nextCronTickInTimeZone()`, same as before
- Added test in routines-service.test.ts: creates a skip_missed routine with past nextRunAt, ticks the scheduler, verifies 0 dispatched runs and nextRunAt advanced to the future

## Verification

- `pnpm -r typecheck` passes
- New test: "does not dispatch overdue schedule runs when catch-up policy is skip_missed"
  - Seeds a routine with `catchUpPolicy: "skip_missed"` (the default from seedFixture)
  - Creates a schedule trigger with `nextRunAt` 5 minutes in the past
  - Calls `tickScheduledTriggers(now)` and asserts `triggered: 0`
  - Verifies `nextRunAt` was advanced past `now`

## Risks

- Low risk. The change only adds a new branch before the existing `enqueue_missed_with_cap` check. The default `runCount = 1` path now only fires when `catchUpPolicy` is neither `skip_missed` nor `enqueue_missed_with_cap` (which preserves backward compatibility for any future policies that should catch up by default).

## Model Used

- Claude Opus 4.6 (1M context) for planning and review
- Codex gpt-5.3-codex for implementation

## Checklist

Fixes #3108

This contribution was developed with AI assistance (Codex + Claude Code).

### DIFF
--- a/server/src/__tests__/routines-service.test.ts
+++ b/server/src/__tests__/routines-service.test.ts
@@ -434,6 +434,46 @@ describeEmbeddedPostgres("routine service live-execution coalescing", () => {
     });
   });
 
+  it("does not dispatch overdue schedule runs when catch-up policy is skip_missed", async () => {
+    const { routine, svc } = await seedFixture();
+    const { trigger } = await svc.createTrigger(
+      routine.id,
+      {
+        kind: "schedule",
+        label: "every minute",
+        cronExpression: "* * * * *",
+        timezone: "UTC",
+      },
+      {},
+    );
+
+    const now = new Date("2026-03-20T12:00:00.000Z");
+    const pastNextRunAt = new Date("2026-03-20T11:55:00.000Z");
+
+    await db
+      .update(routineTriggers)
+      .set({ nextRunAt: pastNextRunAt })
+      .where(eq(routineTriggers.id, trigger.id));
+
+    const tick = await svc.tickScheduledTriggers(now);
+    expect(tick).toEqual({ triggered: 0 });
+
+    const dispatchedRuns = await db
+      .select({ id: routineRuns.id })
+      .from(routineRuns)
+      .where(eq(routineRuns.routineId, routine.id));
+    expect(dispatchedRuns).toHaveLength(0);
+
+    const updatedTrigger = await db
+      .select({ nextRunAt: routineTriggers.nextRunAt })
+      .from(routineTriggers)
+      .where(eq(routineTriggers.id, trigger.id))
+      .then((rows) => rows[0] ?? null);
+
+    expect(updatedTrigger?.nextRunAt).toBeTruthy();
+    expect(updatedTrigger!.nextRunAt!.getTime()).toBeGreaterThan(now.getTime());
+  });
+
   it("blocks schedule triggers when required variables do not have defaults", async () => {
     const { companyId, agentId, projectId, svc } = await seedFixture();
     const variableRoutine = await svc.create(
--- a/server/src/__tests__/routines-service.test.ts
+++ b/server/src/__tests__/routines-service.test.ts
@@ -474,6 +474,50 @@ describeEmbeddedPostgres("routine service live-execution coalescing", () => {
     expect(updatedTrigger!.nextRunAt!.getTime()).toBeGreaterThan(now.getTime());
   });
 
+  it("skip_missed still runs on normal tick jitter (nextRunAt only seconds behind)", async () => {
+    const { companyId, agentId, projectId, svc } = await seedFixture();
+    const routine = await svc.create(
+      companyId,
+      {
+        projectId,
+        goalId: null,
+        parentIssueId: null,
+        title: "jitter test",
+        description: "should still fire",
+        assigneeAgentId: agentId,
+        priority: "medium",
+        status: "active",
+        concurrencyPolicy: "allow",
+        catchUpPolicy: "skip_missed",
+        variables: [],
+      },
+      {},
+    );
+    const trigger = await svc.createTrigger(
+      routine.id,
+      {
+        kind: "schedule",
+        label: "hourly",
+        cronExpression: "0 * * * *",
+        timezone: "UTC",
+      },
+      {},
+    );
+
+    // nextRunAt is 3 seconds in the past - normal tick jitter, NOT a missed run
+    const now = new Date("2026-03-20T12:00:03.000Z");
+    const slightlyPast = new Date("2026-03-20T12:00:00.000Z");
+
+    await db
+      .update(routineTriggers)
+      .set({ nextRunAt: slightlyPast })
+      .where(eq(routineTriggers.id, trigger.id));
+
+    const tick = await svc.tickScheduledTriggers(now);
+    // Should trigger because nextRunAt is within the current cron interval
+    expect(tick).toEqual({ triggered: 1 });
+  });
+
   it("blocks schedule triggers when required variables do not have defaults", async () => {
     const { companyId, agentId, projectId, svc } = await seedFixture();
     const variableRoutine = await svc.create(
--- a/server/src/services/routines.ts
+++ b/server/src/services/routines.ts
@@ -1434,10 +1434,22 @@ export function routineService(db: Db, deps: { heartbeat?: IssueAssignmentWakeup
         let claimedNextRunAt = nextCronTickInTimeZone(row.trigger.cronExpression, row.trigger.timezone, now);
 
         if (row.routine.catchUpPolicy === "skip_missed") {
-          runCount = 0;
-          console.warn(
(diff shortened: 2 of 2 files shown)


mvanhorn:
Addressed in 380a54b7:

- P1 (skip_missed skips every run): added a `nextCronTickInTimeZone()` check on `row.trigger.nextRunAt`. We only set runCount=0 when at least one full cron interval has elapsed since the scheduled time. Normal tick jitter falls through to runCount=1.
- P2 (test only covers outage): added `skip_missed still runs on normal tick jitter` test - seeds nextRunAt 3s in the past for an hourly cron, expects `triggered: 1`.

@greptile please re-review.

PR #3143, opened 2026-04-08
fix(hermes-remote): read MINIMAX_API_KEY into minimaxKey and always push SSH command+args

ebumann1959:
Fixes two bugs in the hermes-remote adapter:

1. MINIMAX_API_KEY was assigned to the local env object instead of the minimaxKey variable used when constructing the SSH command injection
2. execViaSSH only pushed SSH args when cwd was set; with no cwd, the remote command was never sent

Result: heartbeat runs now succeed with exit code 0 and the remote hermes agent authenticates correctly.


musay4632 (approved):

a13xperi (approved):
CI green. Reviewed in burn session. LGTM.

PR #3174, opened 2026-04-09
fix(ui): transcript race — batched state updates with skip-terminal filter (supersedes #2764)

### LINKED ISSUE #2736: Dashboard agent cards show 'No transcript captured' due to React state race in useLiveRunTranscripts

iws17:
## Summary

Closes #2736. Supersedes #2764.

Fixes the \`"No transcript captured"\` dashboard regression under React StrictMode, caused by a React state race in \`useLiveRunTranscripts\`. Each poll tick fired N parallel \`setChunksByRun\` calls that all read from the same \"prev\" snapshot and silently clobbered each other.

This PR adopts the batching approach from @MindSyncHub's PR #2764 and **adds four improvements** that #2764 does not have.

## Why supersede #2764

PR #2764 fixes the race correctly but drops two features that were on \`main\` at its base, and misses two small stability improvements:

| | #2764 | This PR |
|---|---|---|
| Batched \`setChunksByRun\` (fixes race) | ✅ | ✅ |
| Skip-terminal optimization from `627fbc80` | ❌ removed | ✅ preserved (applied as filter inside \`readAll\`) |
| \`parserTick\` / \`onAdapterChange\` for dynamic parser reload | ❌ removed | ✅ preserved |
| \`runsRef\` pattern — effect doesn't re-run on parent re-renders | ❌ | ✅ |
| StrictMode ref-clear on effect re-run | ❌ | ✅ |

### The skip-terminal optimization

`627fbc80` ("Polish issue chat chain-of-thought rendering") added:

\`\`\`ts
const activeRuns = runs.filter((run) => !isTerminalStatus(run.status));
const interval = activeRuns.length > 0 ? setInterval(...) : null;
\`\`\`

This stops polling runs that have already reached a terminal state (\`succeeded\`, \`failed\`, \`cancelled\`, \`timed_out\`). Without it the poller hits \`/api/runs/:id/log\` forever at 2s intervals for done runs.

PR #2764 inadvertently reverted this by moving everything inside a new \`readAll()\` that doesn't filter. This PR keeps the optimization by applying the filter INSIDE \`readAll\`:

\`\`\`ts
const readAll = async () => {
  const currentRuns = runsRef.current.filter(
    (run) => !isTerminalStatus(run.status),
  );
  if (currentRuns.length === 0) return;
  // ...fetch and batch as a single state update...
};
\`\`\`

Benefits both the initial fetch and every poll tick — cleaner than the old external \`activeRuns\` filter.

### The runsRef stability pattern

PR #2764's effect still depends on \`[runIdsKey, runs]\`, which means it restarts the interval whenever the parent re-renders (every \`runs\` array is a new identity even if the contents are the same). This PR uses a ref to the latest runs and depends on \`[runIdsKey]\` only:

\`\`\`ts
const runsRef = useRef(runs);
runsRef.current = runs;
// ...
}, [runIdsKey]);
\`\`\`

Interval is torn down only when the SET of run IDs changes, not on every parent re-render.

### StrictMode ref-clear

\`useRef\` survives a StrictMode dev remount but \`useState\` does not. If the dedup set ref is not cleared at the start of the effect, stale keys from the first mount make the second mount silently dedup every chunk — giving an empty transcript even though the logs are being fetched.

This PR adds the clear:

\`\`\`ts
seenChunkKeysRef.current.clear();
pendingLogRowsByRunRef.current.clear();
logOffsetByRunRef.current.clear();
\`\`\`

## Test plan

- [x] Verified against a production Paperclip fleet with 7 agents over ~4 days. Zero "No transcript captured" regressions.
- [x] Ran \`pnpm typecheck\` + \`pnpm build\` on fresh upstream/master (164 commits ahead of the pre-session base). Both pass clean.
- [x] Verified \`isTerminalStatus\` filter inside \`readAll\`: runs in \`succeeded\` state no longer generate log-endpoint traffic
- [x] Verified the effect does not re-run on parent re-renders that preserve \`runIdsKey\`
- [x] Verified dev-mode StrictMode remount no longer leaves the transcript empty
- [ ] CI suite (please run on merge)

## If maintainers prefer to merge #2764 first

Happy to split this into a stacked follow-up. The clean separation points are:

1. **#2764 as-is** — batched \`setChunksByRun\` fix (already reviewed)
2. **Follow-up PR** — restore \`activeRuns\` filter inside \`readAll\`, restore \`parserTick\`, add \`runsRef\` pattern, add StrictMode ref-clear

Let me know which path you'd prefer. I ran this through a real 164-commit pull today and the merge against `627fbc80` was clean — no conflicts with upstream's current state.

## Credit

Core batching approach: @MindSyncHub (PR #2764). This PR builds on that work.

### DIFF
--- a/ui/src/components/transcript/useLiveRunTranscripts.ts
+++ b/ui/src/components/transcript/useLiveRunTranscripts.ts
@@ -85,6 +85,8 @@ export function useLiveRunTranscripts({
   const seenChunkKeysRef = useRef(new Set<string>());
   const pendingLogRowsByRunRef = useRef(new Map<string, string>());
   const logOffsetByRunRef = useRef(new Map<string, number>());
+  const runsRef = useRef(normalizedRuns);
+  runsRef.current = normalizedRuns;
   // Tick counter to force transcript recomputation when dynamic parser loads
   const [parserTick, setParserTick] = useState(0);
   useEffect(() => {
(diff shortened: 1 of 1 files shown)


iws17:
Thanks for the careful review. Both points fixed in the follow-up commit:

- **P1 — blanket ref-clear double-counts chunks** ✅ Added a `mountKeyRef` that the cleanup nulls out, so the clear only fires on a true mount / StrictMode remount, not on an in-place `runIdsKey` change. Verified: adding a new run to the set no longer resets existing runs' offsets or dedup state.
- **P2 — interval fires forever for all-terminal runs** ✅ Restored the conditional interval creation: only creates the interval when `runsRef.current` has at least one non-terminal run; cleanup handles `null` correctly.

Diff: +40 / -11. Core batching fix from the first commit is unchanged.

iws17:
Thank you for the thorough read, @xkonjin — really appreciate you walking through the mount/remount logic. The three suggestions are all good and worth addressing, I'm keeping them out of this PR to keep the diff focused on the race fix, but noting each for follow-up:

1. **Custom hook to encapsulate `runsRef` + eslint-disable pattern** — agreed this would be cleaner if the pattern recurs. Right now `useLiveRunTranscripts` is the only place in the UI carrying it, so extracting a `useLatestRef`-style helper for a single caller felt premature. Happy to revisit if another consumer adopts the same shape.

2. **Terminal-status filter on every poll tick** — also a fair callout. At current fleet sizes (dashboards in the 10s of runs) the cost is invisible, but if a future view renders hundreds of runs it becomes measurable. A `useMemo` of the non-terminal subset keyed off `runIdsKey` would be a one-line optimization when that day comes.

3. **Silent log-read errors** — this one I deliberately preserved from the pre-race-fix behavior because log polling is lossy by design (partial reads, transient network blips) and the per-tick error spam would drown out actual issues in dev console. That said, a `console.debug` (or the project's existing logger at debug level) would give a discoverability signal without noise. Reasonable follow-up if maintainers want it.

None of these change the race-fix correctness, so holding them for a separate PR to keep review surface small. Thanks again.

iws17:
Closing this — the same race-fix pattern landed independently on `master`. Verified in `ui/src/components/transcript/useLiveRunTranscripts.ts`: batched `setChunksByRun` updates, the 12000-key cap on `seenChunkKeysRef`, the conditional-interval skip-terminal filter, and a surgical prune effect that drops refs for absent runs (cleaner than the `mountKeyRef` pattern this PR proposed). Greptile P1/P2 are both addressed by the upstream design. Thanks @xkonjin for the review — keeping the conversation here as reference.

Issue #958, opened 2026-03-15
Slow UI - Critical memory leak and UI freeze from unpaginated heartbeat runs


mjaverto:
# Critical Performance Bug: Unbounded API Queries Cause UI Freezes and Massive Memory Leaks

## 🚨 The Problem

As a company scales its usage of Paperclip—adding more agents and accumulating a large history of heartbeat runs—the application UI progressively degrades until it becomes entirely unusable. 

Users report the interface feeling incredibly "swampy," unresponsive, and prone to complete lockups when attempting to visit the **Dashboard**, the **Inbox**, the **Agents list**, or any **Agent Detail** page. 

**This is a critical architectural bottleneck that prevents Paperclip from scaling to heavy-usage environments.**

## 🔍 Root Cause Analysis

The root cause stems from the `/api/companies/:companyId/heartbeat-runs` endpoint. Currently, this endpoint does not enforce a default `limit` and lacks support for `offset`/cursor pagination.

Instead of utilizing lightweight database queries, multiple core UI components blindly call `heartbeatsApi.list()` and download the *entire* table history into the browser just to compute basic metrics:

- **Dashboard & Agent Charts:** Downloads all historical runs just to group them by day in React for the "Run Activity" and "Success Rate" charts.
- **Inbox Badges:** Downloads all historical runs just to iterate through and find the single "latest failed run" for each agent to power the sidebar notification dots.
- **Agents List:** Downloads all historical runs just to verify if any agent currently has a "running" or "queued" status dot.

### The Impact
In environments with tens of thousands of runs, this design pattern causes catastrophic failures across the stack:
1. **Backend Crash (OOM):** The Node.js server spikes up to **3GB+ of RAM** and pins the CPU at 30%+ while trying to query, serialize, and buffer the unbounded data.
2. **Network Saturation:** The API continuously transmits payloads upwards of **4.1 MB** of raw JSON data for a single page load.
3. **Frontend Freezes:** The React application locks up the main browser thread trying to parse, sort, and iterate over these massive arrays, leading to the unresponsive "swampy" feel.

## 🛠️ The Fix

This issue requires shifting the heavy data processing from the frontend clients down to PostgreSQL, and enforcing strict pagination on historical data. 

**Proposed Solution:**
1. **Enforce Pagination:** Add `offset` and `limit` support to the `list` endpoint, enforcing a default limit of `200` to prevent accidental OOM crashes.
[...]

PR #3241, opened 2026-04-09
fix(ui): reduce IssueDetail re-renders causing input lag on long issues

pjhoberman:
## Summary

Fixes typing lag / input freezing on long issue detail pages (e.g. issues with many comments and active runs). The root cause is cascading re-renders triggered by 3-second polling that force every `CommentCard` to re-parse markdown via `react-markdown`.

- **Memoize `CommentCard` and `MarkdownBody`** — prevents re-running expensive markdown parsing when the comment content hasn't changed. This is the biggest single win.
- **Stabilize `commentsWithRunMeta` dependencies** — extract primitive values (`runningRunId`, `runningRunStartedAt`) instead of depending on the full `runningIssueRun` object reference, which changes every 3s due to polling. Also preserve original comment references when no run metadata applies, avoiding unnecessary object spreads.
- **Replace `allIssues` query with targeted child issues query** — the old code loaded *every issue in the company* just to filter for child issues. Now uses the server's existing `parentId` filter parameter.
- **Add `parentId` filter to `issuesApi.list` client** and corresponding `queryKeys.issues.children` cache key.

Relates to #958

## Test plan

- [ ] Open a long issue (30+ comments) with an active agent run — typing in the comment box should be responsive
- [ ] Verify sub-issues tab still shows child issues correctly
- [ ] Verify queued comments still display the "Queued" badge
- [ ] All 194 existing UI tests pass (`npx vitest run ui/src/`)

### DIFF
--- a/ui/src/api/issues.ts
+++ b/ui/src/api/issues.ts
@@ -34,6 +34,7 @@ export const issuesApi = {
       executionWorkspaceId?: string;
       originKind?: string;
       originId?: string;
+      parentId?: string;
       includeRoutineExecutions?: boolean;
       q?: string;
     },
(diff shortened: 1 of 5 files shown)


pjhoberman:
Closing to reopen with proper PR template and Greptile feedback addressed.

PR #3242, opened 2026-04-09
fix(ui): reduce IssueDetail re-renders causing input lag on long issues

pjhoberman:
## Thinking Path

> - Humans oversee agent work through a web UI, including an issue detail page
> - The issue detail page shows comments, activity timeline, live run status, and sub-issues
> - Long issues with many comments and active agent runs cause severe input lag when typing in the comment box
> - The root cause is cascading re-renders: 3-second polling on liveRuns/activeRun creates new object references, which invalidate useMemo chains, which recreate all comment objects, which re-render every CommentCard, which re-parses markdown via react-markdown
> - Additionally, the page loaded ALL company issues just to filter for child issues
> - This PR memoizes the expensive components, stabilizes polling-derived dependencies to primitives, wraps callbacks in useCallback, and replaces the all-issues query with a targeted parentId-filtered query
> - The benefit is responsive typing on long issue pages and reduced network/memory overhead from eliminating the full company issue list fetch

## What Changed

- **Memoize `CommentCard`** with `React.memo` — prevents re-rendering comments when their props haven't changed (biggest win since each renders `MarkdownBody`)
- **Memoize `MarkdownBody`** with `React.memo` and `useMemo` for its `components` object — prevents re-parsing markdown through `react-markdown` + `remarkGfm` on parent re-renders
- **Stabilize `commentsWithRunMeta` dependencies** — extract primitive values (`runningRunId`, `runningRunStartedAt`) from the polling-derived `runningIssueRun` object instead of depending on the full reference (which changes every 3s). Also preserve original comment object references when no run metadata applies, avoiding unnecessary object spreads
- **Wrap `handleFeedbackVote` in `useCallback`** and pass the stable `onVote` reference through to `CommentCard` instead of creating per-card closures in `TimelineList` — prevents all CommentCards from re-rendering when unrelated parent state changes
- **Replace `allIssues` query with targeted child issues query** — the old code called `issuesApi.list(companyId)` to load every issue in the company, then filtered client-side for `parentId === issue.id`. Now uses the server's existing `parentId` query parameter. Added `parentId` to the `issuesApi.list` filter type and a `queryKeys.issues.children` cache key
- **Fix non-null assertion asymmetry** — `queryFn` now uses `issue?.id ?? ""` consistent with the `queryKey`, eliminating a non-null assertion that could throw if the `enabled` guard is ever relaxed

## Verification

- All 194 UI tests pass: `npx vitest run ui/src/`
- TypeScript compiles cleanly: `npx tsc --noEmit --project ui/tsconfig.json`
- Manual testing: open a long issue (30+ comments) with an active agent run — typing in the comment box should be responsive
- Sub-issues tab still renders child issues correctly
- Queued comments still display the "Queued" badge
- Feedback voting on comments still works

## Risks

- Low risk. All changes are additive memoization or narrowing of existing queries. The `parentId` filter parameter was already supported server-side but not exposed in the UI client. No migration, no API changes, no behavioral shifts.

## Model Used

- Provider: Anthropic
- Model: Claude Opus 4.6 (`claude-opus-4-6`)
- Context window: 1M tokens
- Capabilities: Tool use (file read/edit, grep, bash), multi-agent research

## Checklist

*Note: This is a performance fix — no visual changes, so before/after screenshots would be profiler traces. The change is verified by the absence of input lag rather than a visual diff. No new tests added as this is purely a memoization/optimization change that doesn't alter behavior — existing tests verify correctness.*

Relates to #958

### DIFF
--- a/ui/src/api/issues.ts
+++ b/ui/src/api/issues.ts
@@ -34,6 +34,7 @@ export const issuesApi = {
       executionWorkspaceId?: string;
       originKind?: string;
       originId?: string;
+      parentId?: string;
       includeRoutineExecutions?: boolean;
       q?: string;
     },
@@ -51,6 +52,7 @@ export const issuesApi = {
     if (filters?.executionWorkspaceId) params.set("executionWorkspaceId", filters.executionWorkspaceId);
     if (filters?.originKind) params.set("originKind", filters.originKind);
     if (filters?.originId) params.set("originId", filters.originId);
+    if (filters?.parentId) params.set("parentId", filters.parentId);
     if (filters?.includeRoutineExecutions) params.set("includeRoutineExecutions", "true");
     if (filters?.q) params.set("q", filters.q);
     const qs = params.toString();
--- a/ui/src/components/CommentThread.tsx
+++ b/ui/src/components/CommentThread.tsx
@@ -1,4 +1,4 @@
-import { memo, useEffect, useMemo, useRef, useState, type ChangeEvent } from "react";
+import { memo, useCallback, useEffect, useMemo, useRef, useState, type ChangeEvent } from "react";
 import { Link, useLocation } from "react-router-dom";
 import type {
   Agent,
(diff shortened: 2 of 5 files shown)


pjhoberman:
Closing — most of this was already addressed by #3163 on master. Will rebase and reopen with just the remaining fixes (CommentCard memo + MarkdownBody memo).

Issue #2068, opened 2026-03-29
Instructions UI: entry file (AGENTS.md) shows empty content and has no Save button


testerrrank2:
## Description

The Instructions tab entry file (AGENTS.md marked as ENTRY) shows empty content in the editor even when the file has content on disk and via the API. Non-entry files (e.g., HEARTBEAT.md) display correctly.

Additionally, the entry file editor has no Save button — you can type in the editor but cannot persist changes via the UI.

## Steps to Reproduce

1. Create an agent with `instructionsBundleMode: "managed"` and `instructionsEntryFile: "AGENTS.md"`
2. Write content to AGENTS.md via the API: `PUT /api/agents/{id}/instructions-bundle/file` with `{"path": "AGENTS.md", "content": "..."}`
3. Verify content exists: `GET /api/agents/{id}/instructions-bundle/file?path=AGENTS.md` returns full content
4. Open the agent in the web UI → Instructions tab → click AGENTS.md
5. Editor shows empty (or just `# Agent instructions` placeholder)
6. No Save button appears even when typing in the editor

## Expected Behavior

- Entry file should display the content returned by the API
- Entry file editor should have a Save button like non-entry files

## Actual Behavior

- Entry file shows empty
- No Save button available
- Non-entry files (HEARTBEAT.md) work correctly — display content and have Save

## Environment

- Paperclip v0.3.1
- `local_trusted` mode
- Tested with `codex_local` and `cursor` adapter types
- Tested in multiple browsers including incognito

## Notes

- The API returns correct content for all files
- Agents receive the correct instructions during heartbeat execution (verified through successful task completion)
- This is UI-only — does not affect agent runtime behavior

PR #3251, opened 2026-04-10
fix(ui): resync bundleDraft when bundle refetches

kbecking:
## Thinking Path

> - Each agent has an instructions bundle (mode, rootPath, entryFile) configurable via the UI's PromptsTab
> - The UI maintains a local `bundleDraft` state that reflects the persisted bundle config and allows unsaved edits
> - After a bundle refetch (e.g., from another tab saving, or initial load), the `useEffect` that initializes `bundleDraft` had an unconditional `if (current) return current` guard
> - This guard prevented re-syncing to the latest persisted state, causing stale `rootPath` values that made `bundleMatchesDraft` evaluate to `false`
> - When `bundleMatchesDraft` is false, `selectedFileExists` returns false, and the UI shows a placeholder instead of file content
> - This pull request replaces the unconditional guard with a dirty-check that only preserves the draft when the user has actual unsaved edits, and extracts this logic into a testable pure function

## What Changed

- **Extracted `resolveBundleDraft` helper** (`ui/src/lib/bundle-draft.ts`): Pure function that decides whether to keep the user's in-progress draft or re-sync to the latest persisted bundle state. Returns `current` only when the user has unsaved edits (any of mode/rootPath/entryFile differs from persisted).
- **Replaced inline guard** in `AgentDetail.tsx`: The `useEffect` at line ~1870 now calls `resolveBundleDraft` instead of the inline `if (current) return current` logic.
- **Added 7 unit tests** (`ui/src/lib/bundle-draft.test.ts`): Covers initialization, re-sync on matching state, preservation on each editable field, and object identity guarantees.

## Verification

```bash
# Run the new unit tests
pnpm exec vitest run ui/src/lib/bundle-draft.test.ts

# Typecheck the UI package
pnpm --filter @paperclipai/ui typecheck
```

Manual verification:
1. Open an agent's Prompts tab → file content displays correctly
2. Switch to another agent and back → content still displays (not placeholder)
3. Edit mode/rootPath/entryFile in the UI without saving → edits are preserved across bundle refetches
4. Save the bundle config → draft resyncs to new persisted values

## Risks

- **Low risk.** The behavioral change is narrow: the only difference from the previous code is that `bundleDraft` now resyncs when it matches persisted state, instead of unconditionally returning the stale draft. The extracted function is a pure refactor of the same logic with no side effects.
- **Complementary to #2283.** PR #2283 by @googlarz addresses a different aspect of the same issue (#2068): it treats the configured entry file as present even when bundle metadata omits it from `fileOptions`. Our fix addresses the upstream cause (stale `bundleDraft` preventing `bundleMatchesDraft` from being true). Both fixes are independent and complementary — #2283 handles the file-existence check, this PR handles the draft-resync logic.

## Model Used

- **Claude Opus 4.6** (`claude-opus-4-6`)
- Context window: 200K tokens
- Extended thinking enabled, tool use (file editing, terminal, search)
- Used via Claude Code CLI within Paperclip agent orchestration

## Checklist

### DIFF
--- a/ui/src/lib/bundle-draft.test.ts
+++ b/ui/src/lib/bundle-draft.test.ts
@@ -1,0 +1,57 @@
+import { describe, expect, it } from "vitest";
+import { resolveBundleDraft, type BundleDraft } from "./bundle-draft";
+
+const persisted: BundleDraft = {
+  mode: "managed",
+  rootPath: "/agents/cto/instructions",
+  entryFile: "AGENTS.md",
+};
+
+describe("resolveBundleDraft", () => {
+  it("initializes from persisted state when current is null", () => {
+    const result = resolveBundleDraft(null, persisted);
+    expect(result).toEqual(persisted);
+  });
+
+  it("re-syncs when draft matches persisted state (no unsaved edits)", () => {
+    // Draft matches what was persisted — user didn't edit anything
+    const current: BundleDraft = { ...persisted };
+    const result = resolveBundleDraft(current, persisted);
+    expect(result).toEqual(persisted);
+    expect(result).not.toBe(current); // new object, not stale reference
+  });
+
+  it("preserves draft when user changed mode", () => {
+    const current: BundleDraft = { ...persisted, mode: "external" };
+    const result = resolveBundleDraft(current, persisted);
+    expect(result).toBe(current);
+  });
+
+  it("preserves draft when user changed rootPath", () => {
+    const current: BundleDraft = { ...persisted, rootPath: "/custom/path" };
+    const result = resolveBundleDraft(current, persisted);
+    expect(result).toBe(current);
+  });
+
+  it("preserves draft when user changed entryFile", () => {
+    const current: BundleDraft = { ...persisted, entryFile: "CUSTOM.md" };
+    const result = resolveBundleDraft(current, persisted);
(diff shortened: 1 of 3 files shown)


serenakeyitan:
🌱 **gardener:verdict:** `ALIGNED` · severity: `low` · commit: `2ae0149d`

Aligned with context tree. No concerns.

The new HEAD refactors the inline `bundleDraft` guard into a pure `resolveBundleDraft()` helper in `ui/src/lib/bundle-draft.ts` with 7 unit tests. This is a quality improvement over the prior in-place fix — behavior is unchanged but the logic is now testable in isolation, consistent with the frontend's pattern of keeping shared pure logic in `ui/src/lib/`. Still fixes the stale `rootPath` bug and still aligns with "surface problems, don't hide them" (the bug was showing a placeholder instead of actual file content).

googlarz (approved):
LGTM on the extract — pure helper, good test coverage, no behavior change from 748ece2f. One small thing worth confirming: Greptile's P2 about the fresh-object return on clean refetches is a real tradeoff — the test re-syncs when draft matches persisted state asserts it, so it's deliberate, but a one-line comment in the helper explaining why you chose to force a re-render (instead of returning current for reference identity) would save the next reviewer from asking.

PR #3274, opened 2026-04-10
fix: make projectId optional in plugin tool execute runContext

### LINKED ISSUE #3273: Plugin tool execute requires projectId but agent runs often have none
## Bug

`POST /api/plugins/tools/execute` validates that `runContext` includes `projectId`:

```ts
if (!runContext.agentId || !runContext.runId || !runContext.companyId || !runContext.projectId) {
```

However, many agent runs have no project context — `projectId` is `null`. This causes agents to get a 400 error when trying to execute plugin tools.

## Impact

Agents without a project context cannot execute any plugin tools. They waste 3-5 tool calls searching for a project ID before giving up.

## Expected Behavior

`projectId` should be optional in `runContext`. It is passed through to the plugin worker as context but is not used for routing, auth, or tool dispatch.

## Suggested Fix

Remove `!runContext.projectId` from the validation check. Keep `agentId`, `runId`, and `companyId` as required.

## Environment

- Paperclip: latest master
- Agent runs triggered by wake-on-demand (no project association)

nullEFFORT:
Fixes #3273

## Summary

- Removes `projectId` from required fields in `POST /plugins/tools/execute` runContext validation
- `agentId`, `runId`, and `companyId` remain required
- `projectId` is still accepted and passed through when present

## Test plan

- [ ] Agent can execute plugin tools with `projectId: null` in runContext
- [ ] Agent can execute plugin tools with a valid `projectId` in runContext
- [ ] Request without `agentId`, `runId`, or `companyId` still returns 400

### DIFF
--- a/server/src/routes/plugins.ts
+++ b/server/src/routes/plugins.ts
@@ -543,9 +543,9 @@ export function pluginRoutes(
       return;
     }
 
-    if (!runContext.agentId || !runContext.runId || !runContext.companyId || !runContext.projectId) {
+    if (!runContext.agentId || !runContext.runId || !runContext.companyId) {
       res.status(400).json({
-        error: '"runContext" must include agentId, runId, companyId, and projectId',
+        error: '"runContext" must include agentId, runId, and companyId',
       });
       return;
     }


serenakeyitan:
🌱 **gardener:verdict:** `ALIGNED` · severity: `low` · commit: `c7368b57`

Correct fix. `companyId` is the primary isolation boundary in Paperclip's architecture — all entities are company-scoped — so requiring `projectId` as a mandatory field for plugin tool execution was overly strict. Dropping it from the validation guard while leaving `agentId`, `runId`, and `companyId` required is consistent with the design. `assertCompanyAccess(req, runContext.companyId)` remains in place, so company-scoping enforcement is unaffected.

Minor note for reviewers: the PR test plan covers the HTTP layer, but it's worth a quick scan of `PluginToolDispatcher.dispatchTool()` to confirm it handles `projectId: undefined` safely. Low risk — the field is already typed as optional in `ToolRunContext`.

ValeriyLapin:
Thanks for this, @nullEFFORT - we hit exactly the problem it describes and would like to see it land. One thing to flag first: I do not think it closes the gap on its own, and it is easy to miss because the second guard sits 200 lines from the one you changed.

`validateToolRunContextScope` runs ten lines later (`server/src/routes/plugins.ts:1032`) and looks the project up unconditionally, with no `if (runContext.projectId)` around it (`:813`). So with this PR applied and no `projectId` supplied, the request stops failing with a 400 and starts failing with a 403 saying `"runContext.projectId" does not belong to "runContext.companyId"`. The caller is no better off, and the new error names the field it was just told it may omit. I checked that rather than assumed it: 403 with only this change applied, 200 with the guard below.

## Suggested change

```suggestion
    if (runContext.projectId) {
      const [project] = await db
        .select({ companyId: projects.companyId })
        .from(projects)
        .where(eq(projects.id, runContext.projectId))
        .limit(1);
      if (!project || project.companyId !== runContext.companyId) {
        return '"runContext.projectId" does not belong to "runContext.companyId"';
      }
    }
```

There is no coverage for the omitted case today, so nothing fails if the second guard stays. Two tests are below for `server/src/__tests__/plugin-routes-authz.test.ts`, next to the existing `allows tool execution when agent, run, and project all belong to runContext.companyId`. The first is red without the guard; the second is there so the guard cannot be loosened into a hole. Take them or leave them.

<details>
<summary>Two tests</summary>

```ts
  it("allows tool execution when runContext omits projectId", async () => {
    const executeTool = vi.fn().mockResolvedValue({ content: "ok" });
    const { app } = await createApp(boardActor(), {}, {
      db: createSelectQueueDb([
        [{ companyId: companyA }],
        [{ companyId: companyA, agentId: agentA }],
      ]),
      toolDeps: {
        toolDispatcher: {
          listToolsForAgent: vi.fn(),
          getTool: vi.fn(() => ({ name: "paperclip.example:search" })),
          executeTool,
        },
      },
    });

    const res = await request(app)
      .post("/api/plugins/tools/execute")
      .send({
        tool: "paperclip.example:search",
        parameters: { q: "test" },
        runContext: {
          agentId: agentA,
          runId: runA,
          companyId: companyA,
        },
      });

    expect(res.status).toBe(200);
    expect(executeTool).toHaveBeenCalledWith(
      "paperclip.example:search",
      { q: "test" },
      {
        agentId: agentA,
        runId: runA,
        companyId: companyA,
      },
    );
  });

  it("still rejects a supplied projectId from another company", async () => {
    const executeTool = vi.fn();
    const { app } = await createApp(boardActor(), {}, {
      db: createSelectQueueDb([
        [{ companyId: companyA }],
        [{ companyId: companyA, agentId: agentA }],
        [{ companyId: companyB }],
      ]),
      toolDeps: {
        toolDispatcher: {
          listToolsForAgent: vi.fn(),
          getTool: vi.fn(() => ({ name: "paperclip.example:search" })),
          executeTool,
        },
      },
    });

    const res = await request(app)
      .post("/api/plugins/tools/execute")
      .send({
        tool: "paperclip.example:search",
        parameters: {},
        runContext: {
          agentId: agentA,
          runId: runA,
          companyId: companyA,
          projectId: projectA,
        },
      });

    expect(res.status).toBe(403);
    expect(executeTool).not.toHaveBeenCalled();
  });
```

</details>

Two things this deliberately does not fix, in case you would rather have them in scope:

- `ToolRunContext.projectId` is typed `string` (`packages/plugins/sdk/src/types.ts:260`). Until that changes the field is optional only for untyped callers: every SDK caller still has to invent a value, and the gateway already invents `""` (`server/src/services/tool-gateway.ts:5757`), which the lookup above would reject.
- A supplied `projectId` is still only checked for company ownership, not against the run, so any project in the company passes. `resolveRunContext` (`server/src/services/tool-gateway.ts:1036`) already derives it from `runId`, cross-checks a supplied value against the stored heartbeat snapshot, and falls back to the issue - the "optional, verified when supplied" shape #3273 is asking for, written one module over.

PR #3285, opened 2026-04-10
feat(plugins): add aws-file-store plugin with MinIO compose

Peacetongue:
Adds an S3-backed shared file store as a first-party Paperclip plugin
  (`@paperclipai/plugin-aws-file-store`) plus a MinIO service in
  `docker/docker-compose.yml` so it works out of the box in dev.

  Inspired by #3230 — same idea, packaged as a real plugin so the host owns
  capability scoping, configuration, lifecycle, metrics, and audit logging
  instead of relying on agents to chain raw `aws` CLI commands.

  ## What's in the plugin

  - **9 tools** registered with the agent runtime: `fs-write`, `fs-read`,
    `fs-list`, `fs-tree`, `fs-mkdir`, `fs-move`, `fs-remove`, `fs-stat`, `fs-search`.
  - **Standard layout** per organization: `knowledge-base/{originals,markdown}/`
    for curated documents and `income/` for files awaiting triage.
  - **S3-compatible backend** built on `@aws-sdk/client-s3` — works against AWS
    S3, MinIO, or any other S3-compatible provider via the `s3Endpoint` /
    `s3ForcePathStyle` config fields.
  - **File index in plugin state** so search by tag / name / description stays
    fast without scanning the bucket.
  - **Lifecycle hooks** wired up: `setup`, `onHealth` (returns file/dir counts
    and bucket reachability), `onConfigChanged`, `onValidateConfig`, `onShutdown`.

  ## Compose changes

  `docker/docker-compose.yml` now ships:
  - a `minio` service with a healthcheck,
  - a one-shot `minio-init` container that creates the bucket,
  - the matching `AWS_*` and `FILE_STORE_*` env vars on `server` so the plugin
    picks up the local MinIO automatically.

  I deliberately did **not** include the unrelated Dockerfile sha256 hash bump
  from #3230 — it's a separate concern.

  ## Tests

  - 16 unit tests in `packages/plugins/aws-file-store/src/storage-backend.test.ts`
    cover key prefixing, path traversal protection, `listDir` filtering,
    `listRecursive` pagination, `move` (copy + delete), the safety check `remove`
    performs against non-empty directories, and the `healthy()` probe.
  - The S3 client is mocked via `vi.mock`, so CI does not need a running MinIO
    container.
  - Added `packages/plugins/aws-file-store` to root `vitest.config.ts` so
    `pnpm test:run` picks the suite up automatically.

PR #3250, opened 2026-04-10
feat(memory): add memory adapter framework and UI

davison:
## Summary

Introduces the memory adapter framework for Paperclip — the generic infrastructure for memory providers, without any specific adapter implementation beyond the built-in PARA (filesystem) adapter.

### What's included

- **Plugin SDK**: `MemoryAdapter` interface and types for third-party adapters
- **Built-in PARA adapter**: Filesystem-based memory provider with company isolation via path scoping
- **Services**: Memory bindings, operations, and hooks services
- **Lifecycle hooks**: Pre-run hydration and post-run capture integrated into the heartbeat
- **REST routes**: Memory bindings and memory operations endpoints
- **DB migration**: `memory_bindings`, `memory_binding_targets`, `memory_operations` tables
- **UI**: Memory Settings page (binding management) and Memory Operations page (operation log)
- **Architecture spec**: `docs/specs/memory-architecture.md`
- **Tests**: Unit tests for memory hooks and PARA adapter

### What's NOT included

Mempalace-specific code (MCP adapter, sidecar, Docker setup, backfill script) has been moved to a separate local-integration branch. This keeps the upstream PR focused on the framework that any adapter can build on.

## Test plan

- [ ] Unit tests pass: `pnpm test -- --filter memory`
- [ ] PARA adapter tests pass
- [ ] Memory hooks tests pass
- [ ] TypeScript compiles cleanly
- [ ] UI memory pages render and are accessible from sidebar
- [ ] Creating a memory binding via the API works
- [ ] Pre-run hydration fires when a binding with `preRunHydrate` enabled exists

PR #3289, opened 2026-04-10
feat(agents): per-agent memory store with keyword search

googlarz:
## Thinking Path

> - Agents run in short heartbeat windows and lose all in-memory context between runs
> - When an agent learns something durable — a user preference, a decision, an environment quirk — there is today no place to store it beyond the issue it happened to be working on
> - This forces agents to either rediscover the same facts every run (wastes budget) or cram them into issue comments that don't belong there
> - This PR adds a per-agent, keyword-searchable memory store that survives across runs, exposed via a new HTTP API every adapter can already reach via `PAPERCLIP_API_KEY`
> - The benefit is that agents can accumulate knowledge across heartbeats without polluting issue threads or re-spending tokens on rediscovery

## What Changed

- New `agent_memories` table (migration `0055_agent_memories.sql`) with composite unique index `(agent_id, content_hash)` for dedupe and a `gin_trgm_ops` index for keyword search. Uses the `pg_trgm` extension that `0051_young_korg.sql` already installed — no new extensions.
- New Drizzle schema `packages/db/src/schema/agent_memories.ts` and shared `AgentMemory` / `AgentMemorySearchResult` / `AgentMemorySaveResult` types + zod validators.
- New `agentMemoryService(db)` with `save`, `search`, `list`, `getById`, `remove(id, agentId)`, `countForAgent`. `save` runs inside a single transaction: insert-or-skip via `ON CONFLICT DO NOTHING`, then count + prune oldest if over `PAPERCLIP_MEMORY_MAX_PER_AGENT` (default 500). `remove` is scoped by both `id` and `agentId` so it cannot delete another agent's row even if the caller skips the route pre-check.
- New HTTP routes under `/agents/:agentId/memories` — `POST`, `GET list/search`, `GET :id`, `DELETE :id`. Every handler gates on `assertCompanyAccess` first, then `assertAgentIdentity`.
- New `assertAgentIdentity(req, agentId)` helper in `server/src/routes/authz.ts`. If the actor is an agent key, it requires `req.actor.agentId === agentId`. Board actors fall through. This closes the `paperclipApiRequest` raw-URL escape hatch — the auth middleware at `server/src/middleware/auth.ts:162` resolves `req.actor.agentId` from the token's DB row (not the URL), so an agent X token hitting `/agents/OTHER/memories` is rejected at the guard.
- Activity log: emits `memory.saved` (only on genuine inserts, not dedupes, to avoid flooding) and `memory.deleted`.
- 24 new integration tests in `agent-memories-service.test.ts`, running against real embedded-postgres via the skip-aware `getEmbeddedPostgresTestSupport` harness.

## Why this approach

The original design called for `pgvector` + Ollama embeddings + per-adapter tool injection. Codex adversarial review killed two load-bearing assumptions:

1. **`embedded-postgres@18.1.0-beta.16` does NOT ship `pgvector`.** Inspection of the shipped extension directory (`node_modules/.pnpm/@embedded-postgres+darwin-arm64@.../native/share/postgresql/extension`) found no `vector.control`. `CREATE EXTENSION vector` would fail in the default dev path.
2. **"Zero adapter changes" was false.** Gemini / Codex / Cursor adapters inject `PAPERCLIP_API_KEY` but do not wire an MCP server. Only adapters whose underlying CLI natively discovers MCP (Claude Desktop, OpenCode) get MCP tools.

Pivoted to: ship keyword search via `pg_trgm` now (which IS already installed and in use for issue/comment search), with the HTTP API as the uniform surface every adapter can reach. Semantic / embedding-backed search becomes a follow-up gated on proving pgvector availability across deployment modes. The MCP tool layer ships in the next PR as a bonus for adapters that support it.

## Verification

- `pnpm --filter @paperclipai/db build` — migration numbering check passes
- `pnpm --filter @paperclipai/shared build` — typecheck clean
- `pnpm --filter @paperclipai/server typecheck` — clean
- `pnpm test:run agent-memories-service` — 24 passing (service + integration)
- `pnpm test:run --project @paperclipai/server` — 758 pre-existing server tests remain green
- Manual curl plan:
  1. `POST /api/agents/<id>/memories` with `{"content":"user prefers French","tags":["preference"]}` → observe `{memory, deduped:false}` with 201
  2. Repeat same POST → observe `{memory, deduped:true}` with 200
  3. `GET .../memories?q=french` → observe ranked match with `score`
  4. Save 501 rows → observe count stays at 500, oldest evicted
  5. With agent X key, call `/api/agents/OTHER/memories` → observe 403

## Risks

- **`pg_trgm` relevance quality**: worse than semantic search for paraphrased queries. Agents are coached in `skills/paperclip/SKILL.md` (follow-up PR) to save content with explicit keywords. Future PR adds `pgvector`-backed semantic search once deployment availability is proven.
- **Schema hot path**: the new table adds 4 indexes (B-tree on `(agent_id, created_at)`, B-tree on `company_id`, GIN trgm on `content`, unique on `(agent_id, content_hash)`). Insert cost is higher than a bare table but still O(log n); for a cap of 500 rows/agent this is negligible.
- **Migration numbering collision**: uses idx `55` — if another PR lands `0055_*` first, this rebases cleanly with a rename to the next free idx.
- **Known non-issue — per-agent isolation**: tested at the service layer (cross-agent `remove` returns null) AND at the route layer (`assertAgentIdentity`). Auth middleware guarantees `req.actor.agentId` comes from the token's DB row, not the URL, so the guard cannot be bypassed by URL manipulation.
- **Content-size**: hard cap 4 KB per memory × 500 per agent = 2 MB/agent. Env-overridable via `PAPERCLIP_MEMORY_MAX_CONTENT_BYTES` and `PAPERCLIP_MEMORY_MAX_PER_AGENT`.

## Model Used

- **Provider / model**: Anthropic Claude Opus 4.6 via Claude Code
- **Exact model ID**: `claude-opus-4-6` (1M context variant: `claude-opus-4-6[1m]`)
- **Context window**: 1,000,000 tokens
- **Capabilities used**: extended tool use (Read / Edit / Write / Bash / Grep / Glob), adversarial design review via OpenAI Codex rescue agent, multi-agent parallel exploration, plan-mode gating
- **Review loop**: initial plan → Codex adversarial review (killed pgvector + MCP-uniform assumptions) → pivot → self-review against staged diff → push → Greptile 4/5 → address both P2 findings → this update

## Checklist

- [x] If this change affects the UI, I have included before/after screenshots — N/A, backend-only PR; UI ships in a follow-up
- [x] I have updated relevant documentation to reflect my changes — follow-up PR adds the agent-facing skill docs

## Greptile follow-up (commit 3893420c)

- **P2 — Prune off-by-one on timestamp ties** (`server/src/services/agent-memories.ts`): moved the "exclude the newly-inserted row" guard from a post-filter into the SELECT's WHERE clause (`ne(agentMemories.id, memoryRow.id)`). Added regression test.
- **P2 — `svc.remove` lacks `agentId` scope / TOCTOU** (`server/src/services/agent-memories.ts` + `routes/agent-memories.ts`): changed signature to `remove(id, agentId)` and scoped the DELETE by both. Dropped the redundant `getById` pre-check from the route. Added a cross-agent remove test.


serenakeyitan:
**🌱 gardener:verdict:** `NEEDS_REVIEW` · severity: `medium` · commit: `b6c6c76b`

> [!CAUTION]
> **Context Review** — this checks product-context fit against the project's context tree, not code correctness. Run Greptile/CodeRabbit for code review.

### Summary

Re-reviewed after HEAD update (`3893420c` → `b6c6c76b`). The new commit is `chore: nudge Greptile re-review (both P2 findings addressed in 3893420c)` — an empty "nudge" commit with no code changes, intended to poke the review bot. No new content to evaluate.

**All tree-level concerns from the prior two cycles still stand:**

1. **Core-vs-plugin placement** — `server/src/routes/agent-memories.ts` and `server/src/services/agent-memories.ts` are still in core. No move toward the plugin SDK pattern demonstrated by PR #3285 (`plugin-aws-file-store`).
2. **Coordination with PR #3250** — both memory PRs are still active. Picking a direction before either merges is still the right first step.
3. **PR template** — "Thinking Path", "What Changed", "Verification", "Risks", "Model Used" sections are still empty boilerplate.

### Recommendation (unchanged)

- Resolve overlap with #3250 before either merges.
- Get an explicit core-vs-plugin decision recorded in the tree.
- Fill in the PR template.

### Meta note on nudge commits

Empty "nudge" commits trigger SHA-tracking bots (including gardener) to re-run. If the goal is specifically to poke Greptile without re-triggering others, consider using `@greptileai re-review` as a comment instead of pushing a commit. For gardener, `@gardener re-review` as a comment does the same thing without changing the HEAD SHA.

---

<sub>Reviewed commit: <code>b6c6c76b</code> · Tree snapshot: <code>33dc6fd0</code> · Commands: <code>@gardener re-review</code> · <code>@gardener pause</code> · <code>@gardener ignore</code></sub>

googlarz:
@greptile-apps review

P2 findings addressed in commit 3893420c — prune off-by-one moved to SELECT WHERE clause with regression test, `svc.remove(id, agentId)` scoped to close the TOCTOU window. Description updated to match the PR template. Please re-review.

googlarz:
Thanks @serenakeyitan — I went through each of the 3 concerns. TL;DR: the PR template is now filled (it was updated right after your bot's review snapshot at `3893420c`), and the coordination/placement concerns aren't actually what they look like.

### 1. PR template

Updated to the full CONTRIBUTING.md template (Thinking Path + What Changed + Why + Verification + Risks + Model Used + Checklist + Greptile follow-up) via browser edit at ~11:10Z — your bot's review was at ~10:30Z so that section is stale. Should be clean on `@gardener re-review`.

### 2. Core vs plugin placement

I looked at what actually lives in each layer before answering:

- **Existing agent-scoped state tables in core**: `agent_task_sessions`, `agent_runtime_state`, `agent_api_keys`, `agent_config_revisions`, `agent_instructions`. All small fixed-schema agent-lifecycle state, all in `packages/db/src/schema/` + `server/src/services/`. Not plugins.
- **What plugins actually do**: see `packages/plugins/examples/*` — external integrations (file storage, notifications, etc.), each with its own manifest, no ability to declare new core DB tables or mount routes under `/agents/:id/…`.
- **What this PR creates**: a small fixed-schema agent-scoped primitive with `pg_trgm` indexes, `assertAgentIdentity` gating, and a 500 rows × 4 KB/agent hard cap. No retrieval augmentation, no embeddings, no external services.

Putting this in plugins would require the plugin SDK to own schemas + indexes + route-layer auth guards + FK cascades to `agents` and `companies`. That's a much bigger change than what plugins currently do.

If you see this as "a knowledge base", I'd push back on that framing — it's literally a per-agent sticky-notes table with keyword search, no smarter than `issue_comments` (which is also in core).

### 3. Coordination with #3250

I diffed the file lists and there's **zero overlap at the schema/route/service layer**:

| | **#3289 (this PR)** | **#3250 (mempalace)** |
|---|---|---|
| Tables | `agent_memories` | `memory_bindings`, `memory_binding_targets`, `memory_operations` |
| Routes | `/agents/:id/memories` | `/memory-bindings`, `/memory-operations` |
| Service dir | `server/src/services/agent-memories.ts` | `server/src/services/memory-adapters/{mempalace,para}.ts` + `memory-hooks.ts` |
| Plugin SDK | no touch | adds `MemoryAdapter`, `MemoryScope`, hydration/capture types |
| Search | `pg_trgm` keyword | semantic via ChromaDB / file |
| Driver | **agent** (explicit `save`/`search` calls) | **framework** (pre-run hydration, post-run capture, prompt injection) |
| LLM calls | none | embeddings for vector search |
| Size | +1,183 lines | +7,361 lines, 43 files |

These are different primitives at different layers, not two implementations of the same thing. #3289 is "agent-controlled explicit notes"; #3250 is "framework-controlled retrieval-augmented context". A well-designed system likely wants both — an agent that wants to explicitly remember "user prefers French" should not need to wait for a semantic-search adapter pipeline.

Per your own table, this PR is ✅ aligned on 4 of 6 axes (auth, isolation, audit, resource bounding) — the only ❓ is the core/plugin call and the only ⚠️ is the "two memory subsystems" framing which I think is a misread of #3250's scope. Happy to coordinate with @davison if the maintainers want a formal decision recorded in the tree, but I don't think either PR should be blocked on the other.

cc @davison — do you see any actual overlap I'm missing?

davison:
hey 👋🏻 

Not sure about the overlap, haven't looked in great detail at what this is providing but there seems to be some slight similarity of intent. I think the obvious point that the gardner plugin just missed completely is that the one I created is more about token cost than fact retrieval. The injection from a mempalace store is an attempt to add better context to the current run, it's not targeted fact retrieval. Feels more complementary than competitive.

With mempalace and my PR (which is still evolving), the interaction is transparent and injects relevant compressed content into the agent's heartbeat or run, and then transparently saves the entire transcript back at the end. The next part of it, which I think will be the more valuable part, is that this gets extended to agent instructions, paperclip API documents and other resources that the agent is presented with at the start of a run and which costs so many tokens to process every time.

But mine comes with a cost (the bit of gardeners analysis that I agree with) in having to setup and manage a sidecar to get this benefit. Their should be choices about what paperclip users want to adopt for their particular case. This plugin and PR looks good.

Are you on the paperclip discord? I'm going to post to the #dev channel there shortly about it. I'll cc you on it (let me have your username if not the same as your github. On discord I'm `autistic shark`.

googlarz:
Thanks @davison — and yes, that framing is exactly right. Your read is the one I'd hoped a human would land on: **token-cost-driven automatic context injection** (mempalace) and **agent-driven explicit fact recall** (this PR) are two halves of the same problem, not competitors. Gardener saw "memory" twice and pattern-matched to overlap without reading what each primitive actually does.

If anything they compose cleanly. An agent that learns "user prefers French" today has no way to pin it durably — mempalace's compressed context injection is optimizing for recency and relevance across a big corpus, so that single short fact can easily get summarized away or fall off the window. `paperclipMemorySave` gives the agent an explicit "don't forget this" primitive it can call when it KNOWS something matters. The two don't need to know about each other — mempalace can read from `agent_memories` as just another content source if it wants to, or not, and it still works.

The direction you mentioned (extending context injection to agent instructions + Paperclip API docs + other run-start resources) is the genuinely high-value next step — those are the biggest per-run fixed costs and the ones that hurt most at scale. Looking forward to seeing where mempalace goes there.

I'm not on the Paperclip Discord yet, so feel free to just @ me on GitHub (or reply here) if you want me in on any follow-up discussion — happy to coordinate on whether it makes sense for `agent_memories` to eventually implement a `MemoryAdapter` interface from your plugin SDK work, so the explicit-save primitive can plug into the same composition surface the sidecar adapters use.

googlarz:

PR #3284, opened 2026-04-10
fix(issues): use resolved UUID instead of identifier in wakeup payloads

kbecking:
## Thinking Path

> - Paperclip orchestrates AI agents for zero-human companies
> - Agents communicate via issue comments and @-mentions, which trigger heartbeat wakeups so the mentioned agent can respond
> - The heartbeat service receives wakeup payloads containing an `issueId` and `taskId` to locate the relevant issue
> - The PATCH `/issues/:id` and POST `/issues/:id/comments` routes accept a human-readable identifier (e.g. `AIU-67`) in `req.params.id`, but resolve it to a UUID internally via `resolveIssueId()`
> - The mention-wake and comment-wake payloads were passing `req.params.id` (the identifier string) as `issueId`/`taskId` instead of the resolved `issue.id`/`currentIssue.id` (UUID)
> - This caused the heartbeat service to silently fail when looking up the issue by a non-UUID string — the mentioned agent was never woken
> - This pull request replaces `id` with `issue.id`/`currentIssue.id` in all affected wakeup payloads, consistent with the assignment-wake in the same routes that already used the correct UUID
> - The benefit is that @-mention wakeups and comment-triggered wakeups now work reliably, enabling correct inter-agent communication

## What Changed

- `server/src/routes/issues.ts`: In the PATCH route, replaced `id` with `issue.id` in the comment-wake (assignee wake on new comment) payload `issueId` and `contextSnapshot.issueId`/`taskId` fields
- `server/src/routes/issues.ts`: In the PATCH route, replaced `id` with `issue.id` in the mention-wake payload `issueId` and `contextSnapshot.issueId`/`taskId` fields
- `server/src/routes/issues.ts`: In the POST `/comments` route, replaced `id` with `currentIssue.id` in the mention-wake payload `issueId` and `contextSnapshot.issueId`/`taskId` fields

All three sites had the same bug: using the raw `req.params.id` (human-readable identifier) instead of the resolved UUID. The assignment-wake in the same PATCH route and the assignee-wake in the POST `/comments` route already used the correct UUID — this fix makes all wakeup payloads consistent.

## Verification

- **Unit tests pass:** `issues-checkout-wakeup.test.ts` (4 tests) and `issues-service.test.ts` (16 tests) all green
- **TypeScript:** No type errors in `routes/issues.ts` (`tsc --noEmit`)
[...]

Issue #3282, opened 2026-04-10
inbox-lite route excludes routine_execution issues


clemsix6:
## Bug

The `/agents/me/inbox-lite` route calls `issuesSvc.list()` without passing `includeRoutineExecutions: true`. Since `issues.list()` excludes `routine_execution` issues by default, agents cannot see their own routine-created issues when querying their inbox.

## Impact

When a routine fires:
1. Issue is created with `origin_kind = "routine_execution"`, assigned to an agent
2. First run works — the `issue_assigned` event injects the issue context directly
3. Agent delegates (e.g., creates a subtask), exits the run
4. On subsequent heartbeats, agent queries `inbox-lite` to find work → **issue is invisible**
5. Agent reports "Inbox empty" and exits, even though it has an active in-progress issue

This causes routine-driven pipelines to stall after the first delegation step.

## Fix

One-liner in `server/src/routes/agents.ts`:

```diff
 const rows = await issuesSvc.list(req.actor.companyId, {
   assigneeAgentId: req.actor.agentId,
   status: "todo,in_progress,blocked",
+  includeRoutineExecutions: true,
 });
```

The board UI can keep its default (hiding routine executions for a clean view), but agents querying their own inbox need to see all assigned work.

## Environment

- Paperclip version: latest main as of 2026-04-10
- Adapter: claude_local
- Reproduction: create a routine that assigns to an agent, trigger it, have the agent delegate a subtask and exit — agent will never pick the issue back up

serenakeyitan:
**🌱 gardener:verdict:** `ALIGNED` · severity: `high` · issue snapshot: `2026-04-10T09:04Z`

> [!NOTE]
> **Context Review** — this checks product-context fit against the project's context tree, not code correctness. Run Greptile/CodeRabbit for code review.

### Summary

Well-diagnosed bug. Aligns directly with the tree's "surface problems, don't hide them" principle — agents currently report "Inbox empty" while silently holding active in-progress routine issues. That's the exact anti-pattern the tree calls out. The proposed one-line fix is correct.

<details open>
<summary><strong>Context match</strong></summary>

| Area | Item intent | Tree guidance | Fit |
|------|-------------|---------------|-----|
| Agent inbox contract | Agents need to see all their assigned work | Tasks are the communication channel; silent filtering breaks the contract | ✅ Aligned |
| Board UI vs agent UI | Board can hide routine executions, agents cannot | Progressive disclosure — layers differ by audience | ✅ Aligned |
[...]

PR #3290, opened 2026-04-10
fix(agents): unblock opencode model dropdown during onboarding

Peacetongue:
## The bug

  During onboarding (most visibly when configuring an `opencode_local`
  adapter), the "pick a model" dropdown shows up empty.

  The UI calls `GET /companies/:companyId/adapters/:type/models` to fill it,
  but for a freshly-signed-up user who hasn't been added to the requested
  company yet, the server replies `403 Forbidden` and the dropdown stays
  blank. Onboarding can't complete cleanly without manually working around
  this.

  `opencode_local` is the adapter where this hurts the most — its model
  list is the one users typically need to pick from during first-time
  setup, and `/detect-model` is what its UI calls right after the user
  points it at a binary. Other adapters technically hit the same gate, but
  opencode is what made the breakage obvious in practice.

  ## Why it happens

  Both `/companies/:companyId/adapters/:type/models` and
  `/companies/:companyId/adapters/:type/detect-model` are gated behind
  `assertCompanyAccess(req, companyId)`. But the handlers they protect
  (`listAdapterModels(type)` and `detectAdapterModel(type)` in
  `server/src/adapters/registry.ts:320` and `:347`) take **only the adapter
  type**. They never read `companyId`. The `companyId` in the URL is
  decorative for these two endpoints — it's used by the auth check and
  nothing else. Model lists are global per adapter type; there's nothing
  company-specific to protect.

  So the gate doesn't defend any data, but it does break the dropdown.

  ## The fix

  Replace `assertCompanyAccess` with `assertBoard` on the two model-listing
  endpoints. Any board actor can now list models for any company, which is
  fine because the response is identical regardless of company. This
  unblocks the opencode onboarding flow without changing what data is
  exposed.

  `POST /companies/:companyId/adapters/:type/test-environment` is
  **intentionally left strict**. Unlike the model endpoints, it actually
  uses `companyId` — it resolves per-company secrets via
  `secretService.normalizeAdapterConfigForPersistence` and
  `resolveAdapterConfigForRuntime` (`server/src/routes/agents.ts:809–817`).
  Loosening that one would let a board user from company A resolve company
  B's secrets. So it keeps `assertCanReadConfigurations`.

### DIFF
--- a/server/src/__tests__/agent-adapter-validation-routes.test.ts
+++ b/server/src/__tests__/agent-adapter-validation-routes.test.ts
@@ -178,3 +178,98 @@ describe("agent routes adapter validation", () => {
     expect(String(res.body.error ?? res.body.message ?? "")).toContain("Unknown adapter type: missing_adapter");
   });
 });
+
+// Adapter model listing endpoints intentionally use `assertBoard` (not
+// `assertCompanyAccess`) because `listAdapterModels(type)` and
+// `detectAdapterModel(type)` ignore companyId — the model list is global per
+// adapter type. This lets the onboarding UI populate the model dropdown for a
+// company the user has not been added to yet. The companion `/test-environment`
+// endpoint is intentionally NOT loosened, because it resolves per-company
+// secrets via secretService.{normalizeAdapterConfigForPersistence,resolveAdapterConfigForRuntime}.
+describe("adapter model listing endpoints", () => {
+  beforeEach(() => {
+    unregisterServerAdapter("external_test");
+    registerServerAdapter(externalAdapter);
+    mockAccessService.canUser.mockResolvedValue(true);
+  });
+
+  afterEach(() => {
+    unregisterServerAdapter("external_test");
+  });
+
+  // The default createApp() above uses `source: "local_implicit"`, which is a
+  // god-mode bypass in assertCompanyAccess. To actually exercise the cross-
+  // company access path we need a realistic remote-session actor with the
+  // company list explicitly scoped.
+  function createSessionBoardApp() {
+    const app = express();
+    app.use(express.json());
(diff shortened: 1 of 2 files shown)


serenakeyitan:
🌱 **gardener:verdict:** `ALIGNED` · severity: `low` · commit: `b853ad7a`

Aligned with context tree. No concerns.

Excellent fix. The author correctly identified that `listAdapterModels(type)` and `detectAdapterModel(type)` ignore `companyId` entirely — the `assertCompanyAccess` gate was guarding nothing and silently breaking the onboarding dropdown. Replacing it with `assertBoard` matches the tree's "Unified API, Different Authorization" pattern: the auth check reflects what data is actually exposed, not what the URL looks like.

What makes this PR stand out is the careful scoping discipline: `POST /test-environment` is explicitly left strict because it resolves per-company secrets via `secretService.normalizeAdapterConfigForPersistence` / `resolveAdapterConfigForRuntime` — loosening that one would be a real cross-company leak. Drawing that exact line, and explaining it in both the PR description and an inline test comment, is the right level of care for an auth change.

Test coverage is thoughtful too — the 95-line test file explicitly notes the `local_implicit` god-mode bypass and constructs realistic remote session-board actors to exercise the cross-company path. This closes a subtle test gap where default `createApp()` would have hidden the fix.

Also worth noting: this is yet another instance of the "silent failure" pattern I've been flagging in recent items (#3282, #3215, #3207, #3284). The 403 → empty dropdown with no clear error is the exact anti-pattern the tree's "surface problems, don't hide them" principle calls out. Worth a general audit: any endpoint gated by `assertCompanyAccess` that doesn't actually read `companyId` in its handler is likely misscoped for the same reason.

PR #3291, opened 2026-04-10
fix(agents): make opencode_local model probe non-blocking on agent create

Peacetongue:
## The bug

  Creating or updating an agent that uses the `opencode_local` adapter
  fails with `422 Unprocessable Entity` on Apple-Silicon hosts running an
  amd64 Paperclip build under Rosetta. The error message comes back as
  `Invalid opencode_local adapterConfig: <stack trace from a crashed
  opencode binary>`. Onboarding the very first opencode agent on those
  hosts is impossible without working around it.

  ## Why it happens

  `server/src/routes/agents.ts` calls `assertAdapterConfigConstraints` from
  the agent create, agent-hire, and agent PATCH routes. For
  `opencode_local` that helper invokes
  `ensureOpenCodeModelConfiguredAndAvailable` from
  `@paperclipai/adapter-opencode-local/server`, which shells out to the
  `opencode` binary to enumerate available models. Under Rosetta the
  binary segfaults intermittently during that enumeration, the helper
  throws, and the catch block re-throws as `unprocessable(...)`. There is
  no way to retry from the UI — every onboarding attempt hits the same
  flaky probe.

  ## The fix

  Demote the failure from a hard 422 to a structured `logger.warn` and let
  agent creation continue. The probe is a best-effort sanity check, not
  the only line of defense — the same model list is re-validated lazily
  by the `/adapters/opencode_local/models` endpoint when the agent
  actually runs, so a genuinely misconfigured agent still surfaces the
  problem at first execution. Honest misconfigurations don't slip through
  silently; only flaky probes get tolerated at create-time.

  The `assertAdapterConfigConstraints` function still runs (we don't skip
  the probe outright) so its side-effects and instrumentation stay
  intact; we just stop converting the thrown error into a request
  failure.

### DIFF
--- a/server/src/__tests__/agent-adapter-validation-routes.test.ts
+++ b/server/src/__tests__/agent-adapter-validation-routes.test.ts
@@ -62,6 +62,36 @@ const mockInstanceSettingsService = vi.hoisted(() => ({
 
 const mockLogActivity = vi.hoisted(() => vi.fn());
 
+const mockEnsureOpenCodeModel = vi.hoisted(() => vi.fn());
+vi.mock("@paperclipai/adapter-opencode-local/server", async (importOriginal) => {
+  const actual = await importOriginal<Record<string, unknown>>();
+  return {
+    ...actual,
+    ensureOpenCodeModelConfiguredAndAvailable: mockEnsureOpenCodeModel,
+  };
+});
+
+const mockLoggerWarn = vi.hoisted(() => vi.fn());
+vi.mock("../middleware/logger.js", () => ({
+  logger: {
+    warn: mockLoggerWarn,
+    info: vi.fn(),
+    error: vi.fn(),
+    debug: vi.fn(),
+    fatal: vi.fn(),
+    trace: vi.fn(),
+    child: vi.fn(() => ({
+      warn: mockLoggerWarn,
+      info: vi.fn(),
+      error: vi.fn(),
+      debug: vi.fn(),
+      fatal: vi.fn(),
+      trace: vi.fn(),
+    })),
+  },
+  httpLogger: vi.fn((_req: unknown, _res: unknown, next: () => void) => next()),
+}));
+
 vi.mock("../services/index.js", () => ({
   agentService: () => mockAgentService,
   agentInstructionsService: () => mockAgentInstructionsService,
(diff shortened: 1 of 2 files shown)


serenakeyitan:
**🌱 gardener:verdict:** `NEEDS_REVIEW` · severity: `medium` · commit: `663720c0`

> [!CAUTION]
> **Context Review** — this checks product-context fit against the project's context tree, not code correctness. Run Greptile/CodeRabbit for code review.

### Summary

The fix unblocks a real and painful onboarding failure (opencode binary segfaults under Rosetta). But softening a validation error into a warning log sits in direct tension with two load-bearing tree decisions: **"surface problems, don't hide them"** and **"no silent auto-recovery"**. The trade-off deserves an explicit maintainer call before landing.

<details open>
<summary><strong>Context match</strong></summary>

| Area | Item intent | Tree guidance | Fit |
|------|-------------|---------------|-----|
| Unblock flaky probe | Let agent creation succeed when probe throws | "Surface problems, don't hide them" (product/NODE.md) | ❓ Partial |
| Demote error to warn | `throw unprocessable(...)` → `logger.warn(...)` | "No silent auto-recovery" — the tree explicitly favors bubbling failures to the actor | ❓ Partial |
| Lazy re-validation path | `/adapters/opencode_local/models` catches the same config at first run | Task-system assumes issues surface where the work happens — acceptable fallback | ✅ Aligned |
| Keep instrumentation | `logger.warn` still logs stack + adapterType + companyId | Auditability — operators can still trace | ✅ Aligned |
| Probe can't distinguish flaky vs. real misconfig | Both paths get demoted | The server can't tell them apart — this means real misconfigs also get deferred | ⚠️ Conflict |

</details>

<details>

- [`product/NODE.md`](https://github.com/serenakeyitan/paperclip-tree/blob/main/product/NODE.md) — "Surface problems, don't hide them" (decision #6)
- [`product/task-system/NODE.md`](https://github.com/serenakeyitan/paperclip-tree/blob/main/product/task-system/NODE.md) — "No automatic recovery, no automatic reassignment"
- [`adapters/opencode-local/NODE.md`](https://github.com/serenakeyitan/paperclip-tree/blob/main/adapters/opencode-local/NODE.md) — adapter lifecycle and validation contract

</details>

### Questions for maintainers

1. **Can the user see the warning?** A `logger.warn` lives in server logs that most end users never look at. If an agent is created with a genuinely broken opencode config, the user sees "success" at create-time and then a failure at first run. That's the scenario the tree explicitly calls out as the anti-pattern. Consider: return `200` with a structured `warnings: [{code, message}]` field in the response body so the UI can show an inline "probe failed, here's why" banner — keeps onboarding unblocked AND surfaces the problem.

2. **Is there a narrower fix?** Two alternatives that address the root cause instead of the symptom:
    - **Detect Rosetta and skip the probe**: check for the `x86_64` arch running under `arm64e` kernel (macOS) and short-circuit the probe with a specific "probe skipped on Rosetta" log line. Users on normal arm64 or amd64 still get the pre-flight validation.
    - **Retry the probe**: if the flakiness is intermittent (the PR says "intermittently"), 2-3 retries with small backoff would fix flakiness without demoting real failures.

3. **Does the UI show the warning?** Worth confirming the onboarding flow can render a structured warning alongside "agent created". If not, this fix makes the bug invisible to end users even when the probe surfaces a genuine misconfig.

### Recommendation

Don't block on unblocking the Rosetta use case — it's a real pain. But before landing this exact fix, consider:

1. Picking a root-cause path (Rosetta detection, retry, or opt-in probe) that preserves create-time validation for honest failures.
2. If that's not feasible in this PR's scope, then at least return the warning in the API response body so the UI can surface it — demoting to server logs alone violates the "surface problems" principle in practice, even if the error is technically still observable.
3. Either way, add a test case for a genuinely invalid opencode config (not just a flaky probe) to prove the lazy-validation path catches it at first run.

---

<sub>Reviewed commit: <code>663720c0</code> · Tree snapshot: <code>33dc6fd0</code> · Commands: <code>@gardener re-review</code> · <code>@gardener pause</code> · <code>@gardener ignore</code></sub>

PR #3330, opened 2026-04-10
feat: impute API-equivalent cost for subscription-included runs

ramonmatias19:
## Summary

- Subscription-included runs previously hard-coded `cost_cents=0`, making the UI show $0.00 for all agents even though tokens were being consumed
- Added Anthropic API pricing map (`ANTHROPIC_PRICING_PER_MTOK`) and `estimateCostCentsFromTokens()` to calculate API-equivalent costs from token counts
- Modified `normalizeBilledCostCents()` to impute costs for `subscription_included` billing type instead of returning 0
- Agent dashboards, Costs page, and budget views now reflect real usage costs

## Thinking Path

The pricing map is intentionally static and scoped to Anthropic models. Imputing costs live from a provider API would:
- Couple every heartbeat to an extra network round-trip with its own failure mode
- Require credentials we do not store for subscription-only customers
- Still be based on the same list prices anyway

A static table keeps the cost path deterministic and auditable; the follow-up work when prices change is a one-line map edit, not a service rewire. A warning log surfaces gaps when a new model appears.

## Risks

- **Pricing drift**: if Anthropic raises list prices and the map is not updated, imputed cost under-reports real API-equivalent spend. Mitigated by the warning log for unknown models; price-update drift still needs manual review each quarter.
- **False precision**: imputed cost is a *list-price* estimate, not a billed amount — subscription customers pay a flat fee regardless. Anyone using this number for accounting should be aware it is an estimate, not an invoice line.
- **Scope**: covers Anthropic Claude models only. Non-Anthropic subscription runs (if/when added) will continue to report 0 until the map is extended.

## Model Used

Claude Opus 4.6 (claude-opus-4-6).

## Checklist

- [x] Tests pass locally (costs-service, heartbeat-run-summary, heartbeat-workspace-session, heartbeat-project-env)
- [x] TypeScript compiles clean
- [x] No breaking changes to the adapter contract
- [x] Warning log for unknown models (addresses greptile P2 feedback)
- [x] PR scope is a single file (heartbeat.ts) — no unrelated changes bundled

## Test plan

- [x] TypeScript compiles clean (`tsc --noEmit`)
- [x] `costs-service.test.ts` — 8/8 passed
- [x] `heartbeat-run-summary.test.ts` — 5/5 passed
- [x] `heartbeat-workspace-session.test.ts` — 33/33 passed
- [x] `heartbeat-project-env.test.ts` — 2/2 passed
- [ ] Verify agent dashboard shows non-zero "Total cost" after new runs
- [ ] Verify Costs page aggregates show correct per-agent and per-model breakdown

### DIFF
--- a/server/src/services/heartbeat.ts
+++ b/server/src/services/heartbeat.ts
@@ -410,8 +410,51 @@ function resolveLedgerBiller(result: AdapterExecutionResult): string {
   return readNonEmptyString(result.biller) ?? readNonEmptyString(result.provider) ?? "unknown";
 }
 
-function normalizeBilledCostCents(costUsd: number | null | undefined, billingType: BillingType): number {
-  if (billingType === "subscription_included") return 0;
+/**
+ * Anthropic API-equivalent pricing per million tokens (USD).
+ * Used to impute a dollar cost for subscription-included runs so the UI
+ * shows what the usage *would* cost at standard API rates.
+ */
+const ANTHROPIC_PRICING_PER_MTOK: Record<string, { input: number; cachedInput: number; output: number }> = {
+  "claude-opus-4-6":           { input: 15,  cachedInput: 1.50, output: 75 },
+  "claude-sonnet-4-6":         { input: 3,   cachedInput: 0.30, output: 15 },
+  "claude-haiku-4-6":          { input: 0.80, cachedInput: 0.08, output: 4 },
+  "claude-sonnet-4-5-20250929": { input: 3,   cachedInput: 0.30, output: 15 },
+  "claude-haiku-4-5-20251001":  { input: 0.80, cachedInput: 0.08, output: 4 },
+};
+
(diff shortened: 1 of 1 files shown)


ramonmatias19:
Addressed both Greptile P2 comments:

1. **Silent zero cost for unrecognized models** — fixed in commit 11a4b722. `estimateCostCentsFromTokens` now logs `logger.warn({ model }, "no pricing entry for model; cost will be reported as 0")` when the model is missing from the pricing map, so gaps are surfaced in operator logs instead of silently reporting $0.

2. **Unrelated auto-scroll changes bundled** — scoped down. PR now touches only `server/src/services/heartbeat.ts`. The `IssueChatThread.tsx` scroll changes are not part of this PR.

PR description updated with the required template sections (Thinking Path, Risks, Model Used, Checklist).

PR #3357, opened 2026-04-11
feat(adapters): add first-class OpenRouter API key support

ordiy:
## Summary

When `OPENROUTER_API_KEY` is set in an agent's env config, Paperclip now automatically maps it to `OPENAI_API_KEY` and sets `OPENAI_BASE_URL=https://openrouter.ai/api/v1` for the child process. This lets users configure OpenRouter with a single key instead of having to manually set two variables.

## Problem

Previously, the billing detection in `adapter-utils/src/billing.ts` already correctly identified `openrouter` as the biller when `OPENROUTER_API_KEY` was present — but the child process (e.g. Codex CLI) never actually received `OPENAI_API_KEY` or `OPENAI_BASE_URL`, so API calls would fail unless both variables were set explicitly.

### Code
- **codex-local** (`execute.ts`): auto-map `OPENROUTER_API_KEY` → `OPENAI_API_KEY` + `OPENAI_BASE_URL` before spawning the Codex process; also treat `OPENROUTER_API_KEY` as an API-key billing signal (billing type `api`)
- **cursor-local** (`execute.ts`): same auto-mapping and billing-type fix

### Docs
- **`docs/deploy/environment-variables.md`**: document `OPENROUTER_API_KEY` and `OPENAI_BASE_URL`; add an _Using OpenRouter_ section with setup examples
- **`docs/adapters/codex-local.md`**: add _Using OpenRouter_ section with JSON config examples and a link to the OpenRouter model list

## Usage after this change

Set only one variable in the agent's `env` config:

```json
{
  "env": {
    "OPENROUTER_API_KEY": "sk-or-v1-…",
    "model": "openai/gpt-4o"
  }
}
```

Paperclip handles the rest automatically. Explicit `OPENAI_API_KEY` / `OPENAI_BASE_URL` still take precedence when set.


ordiy:
Pushed `f84def5` — makes `runtime-config` and worktree port unit tests ignore leaked `DATABASE_URL` / `PORT` from a parallel local Paperclip dev server. Full `pnpm test:run` + `pnpm build` verified green on this head.

PR #3359, opened 2026-04-11
Replace hardcoded githubcli keyring with curl'ed keyring

dagelf:
Simple hardcoded sha256sum of https://cli.github.com/packages/githubcli-archive-keyring.gpg replaced with curl|sha256sum


dagelf:
Yes pulling the hash from the same server is pointless but dpkg needs a hash nonetheless?  I will look a bit deeper to see if there is a better way, but this at least gets it working.

PR #3368, opened 2026-04-11
feat(budgets): add opt-in calendar_day_utc windows

DennisDenuto:
## Thinking Path

- Paperclip already supports multiple budget window kinds, but the current set is too coarse for some operational policies.
- Daily UTC caps are a real control mechanism for AI-agent budgets because they let operators constrain spend at the same cadence they monitor incidents and platform burn.
- The existing budget service already resolves a time window before evaluating spend, so the clean upstream change is to add one more explicit window kind rather than introducing downstream-only overrides.
- This should stay opt-in. Existing installs that use monthly or lifetime windows should not change behavior.
- The implementation therefore adds `calendar_day_utc` to the supported window kinds, resolves the UTC day boundary in the budget service, and verifies the new behavior with tests.

## What Changed

- Added `calendar_day_utc` to the supported budget window kinds.
- Added UTC day window resolution logic in the budget service.
- Applied the resolved UTC day window to spend evaluation for non-lifetime policies.
- Added test coverage for daily window boundary behavior.

## Why this belongs upstream

Daily spend caps are a general operational budgeting feature, not a deployment-specific customization. Paperclip already models budget windows explicitly, so adding `calendar_day_utc` upstream keeps the capability in the core policy model instead of forcing downstream runtime patches.

## Opt-in / Behavior

This change is fully additive.

The default remains `calendar_month_utc`. No existing policy changes behavior unless it explicitly uses:

```json
{
  "windowKind": "calendar_day_utc"
}

### DIFF
--- a/packages/shared/src/constants.ts
+++ b/packages/shared/src/constants.ts
@@ -292,7 +292,7 @@ export type BudgetScopeType = (typeof BUDGET_SCOPE_TYPES)[number];
 export const BUDGET_METRICS = ["billed_cents"] as const;
 export type BudgetMetric = (typeof BUDGET_METRICS)[number];
 
-export const BUDGET_WINDOW_KINDS = ["calendar_month_utc", "lifetime"] as const;
+export const BUDGET_WINDOW_KINDS = ["calendar_day_utc", "calendar_month_utc", "lifetime"] as const;
 export type BudgetWindowKind = (typeof BUDGET_WINDOW_KINDS)[number];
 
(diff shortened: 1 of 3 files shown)


biokub-agent:
Cross-link: #13950 carries the same `calendar_day_utc` window plus the rollover resume that Greptile flagged here. It clears a budget-owned pause once the UTC day rolls over and no active policy is still exceeded, and it also reads every active hard-stop policy for a scope (not only the first). Maintainers may prefer to take one of the two; #13950 is the complete version.

PR #3452, opened 2026-04-12
fix(issues): suppress comment wakes on blocked issues

### LINKED ISSUE #3433: Comment wakes fire on blocked/done/cancelled issues — operator-side notes wake the assignee, defeating the obvious 'pause' workflow
## Summary

Adding a comment to a `blocked` (or `done`, or `cancelled`) issue dispatches an `issue_commented` wake to the assignee, even though the issue isn't in an actionable state. The intuitive operator workflow — "mark blocked, leave a context comment so the agent doesn't restart" — is defeated by the comment itself triggering the agent.

## Repro on `@paperclipai/server@2026.403.0`

1. Pick an issue that's currently `in_progress` with an active assignee.
2. PATCH the issue to `status: blocked`.
3. Add a comment ("paused by operator — do not auto-resume").
4. **Bug:** Paperclip dispatches an `issue_commented` wake to the assignee. The agent wakes, reads the comment, finds nothing actionable on a blocked issue, exits. Cost: ~$0.02–0.05 per stray wake on Sonnet.

## Concrete trace

When trying to pause two in-flight tasks, both with active assignees, by marking blocked + adding a "do not resume" comment, both assignees woke on the comment events and started new runs. Both runs had to be killed via `POST /api/heartbeat-runs/:runId/cancel` to actually stop them.

## Source

`routes/issues.js:1547`:

```js
if (assigneeId && (reopened || !skipWake)) {
    ...
    wakeups.set(assigneeId, {
        source: "automation",
        reason: "issue_commented",
        ...
    });
}
```

The condition has no status check — `blocked`, `done`, and `cancelled` all fall through and dispatch the wake. (The `reopened` branch is correctly handled by `issue_reopened_via_comment` earlier and is fine.)

## Suggested fix

Two options:

- **(a)** Suppress comment wakes when the issue is in a non-actionable status: `blocked`, `done`, `cancelled`. Reopen path stays as-is.
- **(b)** Expose a `silent: true` flag on the comment endpoint for operator-side notes that should not wake the assignee. Less general but more explicit.

## Workaround

To pause work without re-waking, the operator must:
1. Cancel the in-flight run via `POST /api/heartbeat-runs/:runId/cancel`
2. Mark the issue `blocked`
3. **Do NOT comment after blocking.** If you need to leave context, post the comment *before* the cancel (so it's processed inline by the still-running agent).

The fact that the obvious "block + comment" workflow burns tokens and the safe sequence is "cancel-then-block-then-do-not-comment" suggests this is worth fixing at the dispatcher.

shunkakinoki:
## Summary

Adding a comment to a `blocked` issue dispatched an `issue_commented` wake to the assignee, defeating the intuitive operator workflow:
1. Mark issue `blocked`
2. Leave a context note for the record
3. Expect the agent to stay paused

Instead, step 2 triggered a new agent run. The agent would wake, find a blocked issue with nothing actionable, and exit — burning ~$0.02–0.05 per spurious wake.

## Root cause

`isClosed` (checked against `done | cancelled`) already suppresses comment wakes for terminal statuses. `blocked` was missing from the skip condition in both comment-handling code paths.

## Changes

`server/src/routes/issues.ts`:
- PATCH `/issues/:id` comment path: `skipAssigneeCommentWake` now also skips when `issue.status === "blocked"`
- POST `/issues/:id/comments` path: `skipWake` now also skips when `currentIssue.status === "blocked"`

The reopened-via-comment path is unaffected — blocked issues don't go through the reopen branch.

## Test plan

- [ ] Mark an issue `blocked` with an active assignee
- [ ] Add a comment — no wake should fire
- [ ] Mark an issue `todo`/`in_progress` and comment — wake fires as before
- [ ] Comment with `reopen: true` on a done issue — reopen wake still fires

Closes #3433

### DIFF
--- a/server/src/routes/issues.ts
+++ b/server/src/routes/issues.ts
@@ -1746,7 +1746,7 @@ export function issueRoutes(
         const assigneeId = issue.assigneeAgentId;
         const actorIsAgent = actor.actorType === "agent";
         const selfComment = actorIsAgent && actor.actorId === assigneeId;
-        const skipAssigneeCommentWake = selfComment || isClosed;
+        const skipAssigneeCommentWake = selfComment || isClosed || issue.status === "blocked";
 
         if (assigneeId && !assigneeChanged && !skipAssigneeCommentWake) {
           addWakeup(assigneeId, {
@@ -2264,7 +2264,7 @@ export function issueRoutes(
       const assigneeId = currentIssue.assigneeAgentId;
       const actorIsAgent = actor.actorType === "agent";
       const selfComment = actorIsAgent && actor.actorId === assigneeId;
-      const skipWake = selfComment || isClosed;
+      const skipWake = selfComment || isClosed || currentIssue.status === "blocked";
       if (assigneeId && (reopened || !skipWake)) {
         if (reopened) {
           wakeups.set(assigneeId, {


ttomiczek:
HOw the heck do you deal with cases where the comment contains a solution to the blocked state or a contact or research ideas? SOudns good to call the agent to evaluate the comment chain. The problem is the price - but that is a user choice for the moment.

PR #3472, opened 2026-04-12
fix(server): respect externally set PAPERCLIP_API_URL env var

cpfarhood:
## Thinking Path

> - Paperclip server starts up and sets internal `PAPERCLIP_API_URL` for downstream services and adapters
> - The server startup code was unconditionally overwriting `PAPERCLIP_API_URL` with `http://localhost:3100` (or equivalent based on `config.host`)
> - In Kubernetes deployments, `PAPERCLIP_API_URL` is set via a ConfigMap to the externally accessible load balancer URL (e.g. `https://paperclip.example.com`)
> - Because the env var was unconditionally set after loading the ConfigMap value, the ConfigMap-provided URL was ignored and replaced with the internal localhost address
> - This caused downstream services (adapter env building) to use the wrong URL, breaking external access
> - This pull request makes the assignment conditional — only set if not already provided by the environment
> - External deployments can now supply `PAPERCLIP_API_URL` and it will be respected; local development continues to work without setting it

## What Changed

- `server/src/index.ts`: Wrapped `PAPERCLIP_API_URL` assignment in `if (!process.env.PAPERCLIP_API_URL)` guard so externally provided values are preserved
- `server/src/__tests__/server-startup-feedback-export.test.ts`: Added tests verifying external `PAPERCLIP_API_URL` is respected and fallback behavior is correct
- `docs/deploy/environment-variables.md`: Updated `PAPERCLIP_API_URL` description to clarify it can be externally provided and the load balancer/reverse proxy use case

## Verification

- Run the existing test suite: `pnpm test:run server/src/__tests__/server-startup-feedback-export.test.ts` — all 3 tests pass
- Manual verification: Set `PAPERCLIP_API_URL` to a custom value before starting the server and confirm it is not overwritten

## Risks

- Low risk — purely additive conditional check; existing behavior for unset env var is unchanged

## Model Used

MiniMax M2.7 — reasoning-assisted for tracing the root cause through the startup chain (`buildPaperclipEnv` → `startServer` → `config.host` → `HOST` env var)

## Checklist

### DIFF
--- a/docs/deploy/environment-variables.md
+++ b/docs/deploy/environment-variables.md
@@ -18,6 +18,7 @@ All environment variables that Paperclip uses for server configuration.
 | `PAPERCLIP_INSTANCE_ID` | `default` | Instance identifier (for multiple local instances) |
 | `PAPERCLIP_DEPLOYMENT_MODE` | `local_trusted` | Runtime mode override |
 | `PAPERCLIP_DEPLOYMENT_EXPOSURE` | `private` | Exposure policy when deployment mode is `authenticated` |
+| `PAPERCLIP_API_URL` | (auto-derived) | Paperclip API base URL. When set externally (e.g., via Kubernetes ConfigMap, load balancer, or reverse proxy), the server preserves the value instead of deriving it from the listen host and port. Useful for deployments where the public-facing URL differs from the local bind address. |
 
 ## Secrets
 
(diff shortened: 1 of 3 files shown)


cpfarhood:
The tests for `PAPERCLIP_API_URL` handling are in the same file — they were added in the same PR under a new \`describe("startServer PAPERCLIP_API_URL handling")\` block:

- \`"uses the externally set PAPERCLIP_API_URL when provided"\` — sets \`process.env.PAPERCLIP_API_URL\`, calls \`startServer()\`, and asserts \`started.apiUrl === "http://custom-api:3100"\`
- \`"falls back to host-based URL when PAPERCLIP_API_URL is not set"\` — deletes the env var, calls \`startServer()\`, and asserts \`started.apiUrl === "http://127.0.0.1:3210"\`

Both scenarios from the PR description are covered directly against the \`startServer()\` return value.

cpfarhood:
@greptile is this good to request a merge?

cpfarhood:
@greptile your latest comment was addressed and I'd like to request a merge.

cpfarhood:
@greptile This PR has been rebased on current master and squashed to a single clean commit. Please re-evaluate.

cpfarhood:
@greptile Please re-evaluate. Fix applied: updated docs/deploy/environment-variables.md to document PAPERCLIP_API_URL external-provisioning behavior.

PR #3476, opened 2026-04-12
fix(gemini-local): filter startup noise from stderr error fallback

outlawmold:
## Thinking Path

> - The `gemini-local` adapter runs Gemini CLI as a subprocess to execute agent work
> - When a run fails and neither `parsedError` nor `structuredFailure` yield a message, the adapter falls back to `firstNonEmptyLine(stderr)` to populate `run.error`
> - Gemini CLI always emits "YOLO mode is enabled. All tool calls will be automatically approved." as its very first stderr line when launched with `--approval-mode yolo` — regardless of success or failure
> - This means operators see the YOLO startup message as the error instead of the actual failure reason, losing all diagnostic visibility
> - This pull request adds `firstMeaningfulStderrLine()` which skips known Gemini CLI startup/diagnostic noise before selecting the representative stderr line
> - The benefit is that operators now see actionable error messages (e.g. loop recovery, process crash) instead of irrelevant startup output

## What Changed

- `utils.ts`: Added `firstMeaningfulStderrLine()` — filters known noise patterns (`YOLO mode is enabled`, `missing pgrep output`, `Loaded MCP server/tool/plugin`) before selecting the first meaningful stderr line
- `execute.ts`: Replaced `firstNonEmptyLine(attempt.proc.stderr)` with `firstMeaningfulStderrLine()` — fixes `run.error` shown in run history
- `test.ts`: Same replacement in `summarizeProbeDetail()` — fixes the `detail` field shown in the hello probe failure UI during agent configuration
- `src/server/utils.test.ts` *(new)*: 11 regression tests covering all noise patterns, stacked noise lines, noise-only stderr, empty input, and case-insensitivity
- `vitest.config.ts` *(new)*: Vitest config for the `gemini-local` package
- `package.json`: Added `vitest` devDependency and `test`/`test:watch` scripts
- Root `vitest.config.ts`: Added `packages/adapters/gemini-local` to the projects list

## Verification

- Run `pnpm --filter @paperclipai/adapter-gemini-local exec vitest run` — all 11 tests pass
- Simulate a failed run where `parsedError` and `structuredFailure` are empty: `run.error` now shows the actual failure reason instead of `"YOLO mode is enabled. All tool calls will be automatically approved."`
- Hello probe failure detail in the agent config UI is similarly cleaned up

## Risks

Low risk. The change is additive — `firstMeaningfulStderrLine` is a strict superset of `firstNonEmptyLine` that only skips lines matching a fixed allowlist of known noise patterns. If stderr contains only noise and no real error, the function returns `""`, which causes the fallback chain to continue to the `"Gemini exited with code N"` sentinel — same safe default as before.

## Model Used

- Provider: Anthropic
- Model: Claude Sonnet 4.5
- Context window: 200k
- Capabilities used: tool use, multi-file code editing, test generation

## Checklist

### DIFF
--- a/packages/adapters/gemini-local/package.json
+++ b/packages/adapters/gemini-local/package.json
@@ -48,7 +48,9 @@
   "scripts": {
     "build": "tsc",
     "clean": "rm -rf dist",
-    "typecheck": "tsc --noEmit"
+    "typecheck": "tsc --noEmit",
+    "test": "vitest run",
+    "test:watch": "vitest"
   },
   "dependencies": {
     "@paperclipai/adapter-utils": "workspace:*",
@@ -56,6 +58,7 @@
   },
   "devDependencies": {
     "@types/node": "^24.6.0",
-    "typescript": "^5.7.3"
+    "typescript": "^5.7.3",
+    "vitest": "^3.2.4"
   }
 }
--- a/packages/adapters/gemini-local/src/server/execute.ts
+++ b/packages/adapters/gemini-local/src/server/execute.ts
@@ -34,7 +34,7 @@ import {
   isGeminiUnknownSessionError,
   parseGeminiJsonl,
 } from "./parse.js";
-import { firstNonEmptyLine } from "./utils.js";
+import { firstMeaningfulStderrLine } from "./utils.js";
 
 const __moduleDir = path.dirname(fileURLToPath(import.meta.url));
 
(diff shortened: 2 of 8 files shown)


outlawmold:
P1 fixed — restored cli/package.json to its committed state and regenerated the lockfile cleanly.

P2 fixed — firstNonEmptyLine(stdout) is now used for the stdout path. firstMeaningfulStderrLine only applies to stderr where the YOLO/MCP noise actually appears.

PR #3494, opened 2026-04-12
[codex] Fix plugin tool dispatcher pluginDbId routing

### LINKED ISSUE #3394: Plugin tool dispatcher: pluginDbId never threaded through, every external plugin's agent.tools.execute returns 502
## Summary

Every external plugin that declares `agent.tools.register` and exposes one or more agent tools currently fails on `POST /api/plugins/tools/execute` with HTTP 502 *"worker for plugin <package-key> is not running"*, even though the plugin worker is healthy and serving JSON-RPC traffic on its IPC channel.

The bundled `agent-analytics.paperclip-live-analytics-plugin` happens to declare zero tools, so the regression has been invisible up to now — it only surfaces the moment any plugin actually exposes agent tools.

## Root cause

The plugin tool dispatcher's `registerPluginTools` accepts only the package key (e.g. `acme.foo`) and forwards it as both the namespace key AND the worker-routing key into the plugin tool registry:

https://github.com/paperclipai/paperclip/blob/master/server/src/services/plugin-tool-dispatcher.ts#L429-L434

```ts
registerPluginTools(
  pluginId: string,
  manifest: PaperclipPluginManifestV1,
): void {
  registry.registerPlugin(pluginId, manifest);
},
```

Note that `registry.registerPlugin` already accepts an optional third argument:

https://github.com/paperclipai/paperclip/blob/master/server/src/services/plugin-tool-registry.ts#L298-L299

```ts
registerPlugin(pluginId: string, manifest: PaperclipPluginManifestV1, pluginDbId?: string): void {
  const dbId = pluginDbId ?? pluginId;
```

…but the dispatcher never populates it, so `dbId` always falls back to the package key.

The plugin loader in turn calls the dispatcher with the package key:

https://github.com/paperclipai/paperclip/blob/master/server/src/services/plugin-loader.ts#L1815

```ts
toolDispatcher.registerPluginTools(pluginKey, manifest);
```

…even though the loader has the **DB UUID** in scope as `pluginId` and uses it consistently elsewhere (e.g. `workerManager.startWorker(pluginId, …)` ten lines earlier).

When `executeTool` then runs the worker availability check:

https://github.com/paperclipai/paperclip/blob/master/server/src/services/plugin-tool-registry.ts#L404-L411

```ts
const dbId = tool.pluginDbId;
if (!workerManager.isRunning(dbId)) {
  throw new Error(
    `Cannot execute tool "${namespacedName}" — ` +
    `worker for plugin "${pluginId}" is not running.`,
  );
}
```

…it asks the worker manager whether a worker keyed by the **package key** is running. But `workerManager.isRunning` looks up workers by **DB UUID** (set in `workerManager.startWorker`), so the lookup misses, every tool execution returns 502, and the plugin's agent tools are unreachable.

## Reproduction

1. Install any external plugin declaring `agent.tools.register` with at least one tool. For example, the standalone `paperclip-plugin-github-issues` plugin (https://github.com/mvanhorn/paperclip-plugin-github-issues) registers three tools.
2. After install, watch the loader log it as fully activated:
   ```
   plugin-loader: plugin activated successfully
   {pluginId: "<UUID>", pluginKey: "<package-key>", registered: {worker: true, tools: 3}}
   ```
3. Call the dispatch endpoint:
   ```bash
   curl -s -X POST 'http://127.0.0.1:3100/api/plugins/tools/execute' \
     -H 'Content-Type: application/json' \
     -d '{
       "tool": "<package-key>:<tool-name>",
       "parameters": {...},
       "runContext": {"agentId":"...","runId":"...","companyId":"...","projectId":"..."}
     }'
   ```
4. Receive HTTP 502:
   ```json
   {"error":"Cannot execute tool \"<package-key>:<tool-name>\" — worker for plugin \"<package-key>\" is not running."}
   ```

The plugin worker is in fact alive: `GET /api/plugins/<UUID>/health` returns `{"healthy": true}`, the periodic-sync job runs successfully, the worker process appears in `ps`, and direct RPC calls (e.g. `lifecycle.restartWorker`) succeed. Only `executeTool` fails.

## Fix

Thread `pluginDbId` through `registerPluginTools` as a separate optional parameter and forward it to `registry.registerPlugin`. The package key continues to be used for tool namespacing (e.g. `acme.foo:search`) so existing namespaces are preserved.

PR with the three-line fix and updated docstrings: #3395

## Impact

Any plugin that exposes agent tools through `ctx.tools.register` is currently unusable from agent runs in this codebase. Plugins that only emit jobs/webhooks or only consume events are unaffected.

akshitnanda:
## Summary
- thread the persisted plugin database id through plugin tool registration
- route external plugin tool execution by `pluginDbId` instead of only in-memory plugin metadata
- add a focused regression test for `agent.tools.execute` dispatching through external plugins

## Why
Fixes #3394.

The dispatcher was registering tools without the plugin database id that downstream execution expects. That left external plugin `agent.tools.execute` calls unable to resolve the plugin record cleanly, which surfaced as a `502` failure path.

## Impact
External plugin tool calls now resolve through the same persisted identifier used elsewhere in the server, which makes plugin-backed tool execution reliable and keeps the fix tightly scoped to the registration/dispatch layer.

## Validation
- `corepack pnpm vitest run server/src/__tests__/plugin-tool-dispatcher.test.ts`

### DIFF
--- a/server/src/__tests__/plugin-tool-dispatcher.test.ts
+++ b/server/src/__tests__/plugin-tool-dispatcher.test.ts
@@ -1,0 +1,62 @@
+import { describe, expect, it, vi } from "vitest";
+import { createPluginToolDispatcher } from "../services/plugin-tool-dispatcher.js";
+import type { PluginWorkerManager } from "../services/plugin-worker-manager.js";
+
+describe("plugin tool dispatcher", () => {
+  it("routes tool execution using the plugin database id when provided", async () => {
+    const call = vi.fn(async () => ({
+      content: [{ type: "text", text: "ok" }],
+    }));
+
+    const workerManager = {
+      startWorker: vi.fn(),
+      stopWorker: vi.fn(),
+      getWorker: vi.fn(),
+      isRunning: vi.fn((pluginId: string) => pluginId === "plugin-db-id"),
+      stopAll: vi.fn(),
+      diagnostics: vi.fn(() => []),
+      call,
+    } satisfies PluginWorkerManager;
+
+    const dispatcher = createPluginToolDispatcher({ workerManager });
+    dispatcher.registerPluginTools(
+      "acme.example",
+      {
+        apiVersion: 1,
+        name: "acme.example",
+        version: "1.0.0",
+        tools: [
+          {
+            name: "search",
+            description: "Search",
+            parametersSchema: { type: "object" },
+          },
+        ],
+      },
+      "plugin-db-id",
+    );
+
+    const result = await dispatcher.executeTool(
+      "acme.example:search",
+      { query: "bug" },
+      {
+        agentId: "agent-1",
+        runId: "run-1",
+        companyId: "company-1",
+        projectId: "project-1",
+      },
+    );
+
+    expect(workerManager.isRunning).toHaveBeenCalledWith("plugin-db-id");
+    expect(call).toHaveBeenCalledWith(
+      "plugin-db-id",
+      "executeTool",
+      expect.objectContaining({
+        toolName: "search",
+        parameters: { query: "bug" },
+      }),
+    );
+    expect(result.pluginId).toBe("acme.example");
+    expect(result.toolName).toBe("search");
+  });
+});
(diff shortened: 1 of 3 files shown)


devinfoley:

PR #3498, opened 2026-04-12
[codex] Increase the gemini_local hello probe timeout

### LINKED ISSUE #3372: gemini_local: hello probe times out due to slow Gemini CLI startup (default helloProbeTimeoutSec too short)
## Summary

The Gemini hello probe (`gemini_hello_probe_timed_out`) fails intermittently with a `warn` status because the default `helloProbeTimeoutSec` of **10 seconds** is too tight for Gemini CLI's cold-start time.

## Environment

- Gemini CLI version: `0.37.1`
- Paperclip canary: `v2026.411.0-canary.1`
- Platform: Linux (Docker container, `node:lts-trixie-slim`)
- Auth type: `oauth-personal` with `GOOGLE_CLOUD_PROJECT` set (Workspace/GCA account)

## Root Cause

Gemini CLI startup involves two sequential slow steps before the first output line is emitted:

1. **Node.js ESM cold-start** — `gemini` is a 523 KB ESM bundle (`/usr/local/lib/node_modules/@google/gemini-cli/bundle/gemini.js`). Node.js V8 + ESM module parsing takes ~700 ms.
2. **Google Cloud project validation** — accounts with `GOOGLE_CLOUD_PROJECT` set trigger a network call to validate the project ID at every startup. This adds ~7–8 seconds before the CLI begins inference.

**Measured timings:**
| Scenario | Elapsed |
|---|---|
| `gemini --version` (no network) | ~700 ms |
| Full probe (`--output-format stream-json --prompt ...`) | ~9 994 ms |
| API call only (`duration_ms` in result JSON) | ~2 315 ms |
| Startup overhead before first output | ~7 600 ms |

The current default of `helloProbeTimeoutSec: 10` means the probe has almost no margin and times out consistently for accounts that require project validation.

## Impact

- The probe reports `warn` on every test even though Gemini is fully functional
- Confusing for users who see "Gemini hello probe timed out" immediately after a successful agent setup

## Suggested Fix

Increase the default `helloProbeTimeoutSec` in `packages/adapters/gemini-local/src/server/test.ts`:

```ts
// Current
const helloProbeTimeoutSec = Math.max(1, asNumber(config.helloProbeTimeoutSec, 10));

// Proposed
const helloProbeTimeoutSec = Math.max(1, asNumber(config.helloProbeTimeoutSec, 30));
```

A value of **30 seconds** covers the worst-case startup (Workspace accounts + cold Node.js) while still failing fast for broken installs.

**Workaround (available today):** Set `"helloProbeTimeoutSec": 30` manually in the adapter's JSON config.

akshitnanda:
## Summary
- raise the default `gemini_local` hello probe timeout from 10 seconds to 30 seconds
- add a focused regression test that asserts the longer timeout is passed into the child process probe

## Why
Fixes #3372.

The initial hello probe timeout was short enough to fail on slower local Gemini startups even when the adapter would become healthy moments later. Extending the probe window reduces false negatives without changing the rest of the adapter handshake behavior.

## Impact
`gemini_local` adapters get a more realistic startup window, which makes local adapter health checks less brittle on slower machines.

## Validation
- `corepack pnpm vitest run server/src/__tests__/gemini-local-hello-probe-timeout.test.ts`

### DIFF
--- a/packages/adapters/gemini-local/src/server/test.ts
+++ b/packages/adapters/gemini-local/src/server/test.ts
@@ -135,7 +135,7 @@ export async function testEnvironment(
       const model = asString(config.model, DEFAULT_GEMINI_LOCAL_MODEL).trim();
       const approvalMode = asString(config.approvalMode, asBoolean(config.yolo, false) ? "yolo" : "default");
       const sandbox = asBoolean(config.sandbox, false);
-      const helloProbeTimeoutSec = Math.max(1, asNumber(config.helloProbeTimeoutSec, 10));
+      const helloProbeTimeoutSec = Math.max(1, asNumber(config.helloProbeTimeoutSec, 30));
       const extraArgs = (() => {
         const fromExtraArgs = asStringArray(config.extraArgs);
         if (fromExtraArgs.length > 0) return fromExtraArgs;
--- a/server/src/__tests__/gemini-local-hello-probe-timeout.test.ts
+++ b/server/src/__tests__/gemini-local-hello-probe-timeout.test.ts
@@ -1,0 +1,51 @@
+import { beforeEach, describe, expect, it, vi } from "vitest";
+
+let capturedTimeoutSec: number | null = null;
+
+vi.mock("@paperclipai/adapter-utils/server-utils", async () => {
+  const actual = await vi.importActual<typeof import("@paperclipai/adapter-utils/server-utils")>(
+    "@paperclipai/adapter-utils/server-utils",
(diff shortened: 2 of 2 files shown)


gcampton:
I hit this to and claude made the same changes. 
Comment on #3498 and #1663 with "I hit this today, confirmed the fix works — here's the reproducer output" to bump maintainer attention. Helpful for the community, no new PR churn.
image

PR #3526, opened 2026-04-12
fix(ui): defer isComposingRef reset to prevent first-paragraph keyboard dismissal

kbecking:
## Summary

- Fixes keyboard dismissal on Android when selecting IME word suggestions in the first paragraph of the MarkdownEditor
- The previous fix (PR #23) deferred the compositionend flush to rAF but still set `isComposingRef = false` synchronously, creating a window where MutationObserver and selectionchange handlers slipped through composition guards during Lexical's DOM reconciliation
- Now both the flush AND the `isComposingRef` reset are deferred to the rAF callback, keeping all guards active until the browser has fully processed the composition

## Root cause

On `compositionend`, Lexical processes the composition synchronously (`updateEditorSync` → `$commitPendingUpdates`), reconciling the DOM within the same event handler. On the first paragraph, this reconciliation generates mutation and selection events (composition suffix removal on the only text node) that, with `isComposingRef = false`, triggered:
- `MutationObserver` → `decorateProjectMentions()` during unstable DOM
- `selectionchange` → `checkMention()` → potential React re-render
- Value sync `useEffect` → `setMarkdown()` with stale parent value

After Enter (second paragraph), less DOM restructuring occurs, making the window less impactful — which explains the workaround.

## Test plan

- [ ] On Android (Pixel 10 Pro XL or similar), open a comment editor
- [ ] Type in the first paragraph and select a word suggestion from the IME bar
- [ ] Verify keyboard stays open
- [ ] Press Enter, type in second paragraph, select suggestion — verify still works
- [ ] Test @-mention autocomplete still works after composition
- [ ] Verify desktop editing is unaffected


kbecking:
Closed — opened on upstream by mistake. PR will be created on the fork instead.

PR #2472, opened 2026-04-02
fix: copy button fallback for non-secure contexts

devinfoley:
## Thinking Path

> - Paperclip orchestrates AI agents for zero-human companies
> - The UI serves agent management pages including an instructions editor with copy-to-clipboard buttons
> - The Clipboard API (`navigator.clipboard.writeText`) requires a secure context (HTTPS or localhost)
> - Users accessing the UI over HTTP on a LAN IP get "Copy failed" when clicking the copy icon
> - This pull request adds an `execCommand("copy")` fallback in `CopyText` for non-secure contexts
> - The benefit is that copy buttons work reliably regardless of whether the page is served over HTTPS or plain HTTP

## What Changed

- `ui/src/components/CopyText.tsx`: Added `window.isSecureContext` check before using `navigator.clipboard`. When unavailable, falls back to creating a temporary `<textarea>`, selecting its content, and using `document.execCommand("copy")`. The return value is checked and the DOM element is cleaned up via `try/finally`.

## Verification

- Access the UI over HTTP on a non-localhost IP (e.g. `http://[local-ip]:3100`)
- Navigate to any agent's instructions page → Advanced → click the copy icon next to Root path
- Should show "Copied!" tooltip and the path should be on the clipboard

## Risks

- Low risk. `execCommand("copy")` is deprecated in the spec but universally supported by all major browsers. The fallback only activates in non-secure contexts where the modern API is unavailable. If/when HTTPS is enabled, the modern `navigator.clipboard` path is used automatically.

## Checklist

- [x] I have included a thinking path that traces from project context to this change
- [ ] I have run tests locally and they pass
- [ ] I have added or updated tests where applicable
- [ ] If this change affects the UI, I have included before/after screenshots
- [ ] I have updated relevant documentation to reflect my changes
- [x] I have considered and documented any risks above
- [x] I will address all Greptile and reviewer comments before requesting merge

PR #3531, opened 2026-04-12
fix(ui): make copy-to-clipboard buttons work in non-secure contexts

### LINKED ISSUE #3529: Copy-to-clipboard icons silently fail in non-secure contexts (self-hosted HTTP)
## Summary

Most copy-to-clipboard icons in the UI silently no-op when Paperclip is accessed over plain HTTP from a non-localhost hostname. This scenario is common for self-hosted instances: a server reached at `http://<lan-host>:3100` on a home/office LAN, or behind a private overlay network like Tailscale (`http://<machine>.<tailnet>.ts.net:3100`), Wireguard, a VPN, or an SSH tunnel — all authenticated, none exposed to the public internet, but all still on plain HTTP from the browser's point of view. A page is only treated as a [secure context](https://developer.mozilla.org/en-US/docs/Web/Security/Secure_Contexts) when it's HTTPS or HTTP loopback (`http://localhost` / `http://127.0.0.1`) — private-network reachability doesn't count — so `navigator.clipboard` is `undefined` on any of these setups.

The user-visible symptom is: nothing lands on the clipboard, no toast surfaces, and the icon doesn't even animate — the click appears to do absolutely nothing.

## Repro

1. Run `paperclipai` and access the UI from a **different machine** on the LAN (so the URL is something like `http://<lan-host>:3100`, not `localhost`).
2. Go to an agent's detail page.
3. Open the agent's overflow menu and click **Copy Agent ID**.
4. Paste elsewhere.

**Expected:** clipboard contains the agent UUID.
**Actual:** clipboard is unchanged, no error toast, and the menu item gives no feedback. The click has no visible effect at all, which makes the failure easy to misread as "is this button even wired up?"

Same silent-failure behavior for the agent token copy on key rotation, snippet/invite copy on CompanySettings, skill source-path copy on CompanySkills, branch-name and folder-path copy in `IssueProperties`, workspace-path copy in `IssueWorkspaceCard`, document-body copy in `IssueDocumentsSection`, routine secret copy on RoutineDetail, worktree-name copy in `WorktreeBanner`, code-block / message / context-menu copies in `IssueChatThread`, and the full-issue markdown copy on `IssueDetail`.

### What already works on current master (for contrast)

Two components received individual fallback fixes in earlier PRs: `CopyText` (used in `AgentDetail`, `ExecutionWorkspaceDetail`, and `ProjectDetail` for working-directory paths, session IDs, and repo URLs) and the `CopyMarkdownButton` inside `CommentThread` (used for "Copy as markdown" on comments). Both carry their own inlined `document.execCommand("copy")` path and work correctly in non-secure contexts today. That's exactly why the broken sites are so confusing — a user clicks the comment copy button and it works, then clicks "Copy Agent ID" and nothing happens, with no hint as to why.

The goal of this issue is to consolidate those two existing fallbacks into one shared utility and bring the remaining 14 sites onto the same path.

## Root cause

The Clipboard API requires a [secure context](https://developer.mozilla.org/en-US/docs/Web/Security/Secure_Contexts) — HTTPS or `http://localhost` / `http://127.0.0.1`. On any other origin, `navigator.clipboard` is `undefined` per spec. The UI calls `navigator.clipboard.writeText(...)` directly in 14 places without a fallback. Reading `.writeText` off `undefined` throws `TypeError: Cannot read properties of undefined (reading 'writeText')` **synchronously**, before any `.then` callback can run — so on the sites wrapped in `try {} catch {}` the error is swallowed, and on the sites that use bare `.then()` the throw escapes the onClick into React's internal handler and is never observed by user code. Either way, the state update that would animate the icon (`setCopied(true)` / `setStatus("copied")`) never executes and no toast ever fires.

Grep result (master at time of filing):

```
ui/src/pages/CompanySkills.tsx:606
ui/src/pages/AgentDetail.tsx:969
ui/src/pages/AgentDetail.tsx:3983
ui/src/pages/IssueDetail.tsx:1537
ui/src/pages/RoutineDetail.tsx:401
ui/src/pages/CompanySettings.tsx:143, 522
ui/src/components/IssueWorkspaceCard.tsx:72
ui/src/components/IssueDocumentsSection.tsx:514
ui/src/components/IssueProperties.tsx:32
ui/src/components/WorktreeBanner.tsx:10
ui/src/components/IssueChatThread.tsx:753, 952, 1159
```

## Impact

Breaks the majority of copy icons in the UI for self-hosted users who aren't on HTTPS and aren't on the same machine as the server — the 14 sites listed above. The two sites that got individual fixes earlier (`CopyText` and `CopyMarkdownButton`) continue to work, which is why the experience reads as "most copy buttons are broken, a couple aren't, and there's no obvious pattern to it." From the user's side all they see on a broken button is: click, nothing happens, clipboard unchanged, icon static, no toast — it looks like the button is inert.

## Prior work

This isn't proposing a new technique — the execCommand fallback has already been accepted into the codebase twice:

- **#2472** (`26069682`) plus iteration commits `55833078`, `b64607c2`, `944a7244` — added the fallback to the `CopyText` component after Greptile review cycles.
- **`1e4ccb2b`** ("Improve mobile comment copy button feedback") — added the `copyTextWithFallback` helper inside `CommentThread.tsx` as part of a mobile UX pass on the copy-as-markdown button.

Neither attempted consolidation; each fixed only the component the author was actively working on. The result is two near-identical inline implementations of the same textarea-+-`execCommand` dance, plus 14 other copy sites that never got the treatment. The fix here is to extract that already-accepted pattern into one shared utility and apply it everywhere.

## Proposed fix

Extract the existing `CommentThread`/`CopyText` fallback pattern into a single shared utility at `ui/src/lib/clipboard.ts` — `copyTextToClipboard(text: string): Promise<void>` — and migrate every call site to use it. The utility uses `navigator.clipboard.writeText` when the secure-context gate passes, and falls back to a transient invisible (opacity: 0, 1×1px, zIndex: -1) textarea mounted briefly on the document + `document.execCommand("copy")` otherwise. The fallback runs synchronously inside the caller's user gesture, so the trusted-gesture invariant is preserved. No new capabilities, no XSS surface (we only set `textarea.value`), and focus of the previously-focused element is restored so users mid-typing aren't bumped out of their inputs.

PR: #3531.

## Related observation (separate bug)

While debugging this, I noticed `server/dist/app.js` reads `ui-dist/index.html` into memory once at server startup and serves that cached string for every SPA route fallback. This means any fix or local patch applied to `ui-dist/` only takes effect on the literal `/` path (via `express.static`) until the server is restarted. Not part of this issue's fix — worth a separate tracking issue if the behavior isn't intentional, since it surprises anyone trying to hot-patch the UI.

joshdutcher:
Fixes #3529.

## Thinking Path

> - Paperclip orchestrates AI agents for zero-human companies, and that orchestration is driven from a React SPA served by a Node/Express process
> - The UI has ~16 copy-to-clipboard icons scattered across it — Copy Agent ID, token rotation, invite snippets, workspace paths, document bodies, issue markdown, worktree names, copy-as-markdown on comments, code-block copies in chat transcripts, etc.
> - Anyone self-hosting Paperclip and reaching it over plain HTTP from another machine on their network — LAN hostname, Tailscale MagicDNS, Wireguard/VPN/SSH tunnel — loses the browser's Clipboard API. A page is only a secure context when it's HTTPS or HTTP loopback (`localhost`/`127.0.0.1`); private-network reachability doesn't count, so `navigator.clipboard` is `undefined` on any of these setups.
> - 14 of those copy sites call `navigator.clipboard.writeText(...)` directly with no fallback — the resulting synchronous `TypeError` either dies inside a silent `try/catch` or bubbles out of the onClick before any `.then` callback can run, so the user sees no feedback at all: clipboard unchanged, icon doesn't animate, no toast
> - The execCommand fallback pattern has already been accepted into the codebase twice — **#2472** (`26069682`, plus iteration commits `55833078`, `b64607c2`, `944a7244`) inlined it into `CopyText.tsx`, and commit `1e4ccb2b` ("Improve mobile comment copy button feedback") inlined it into `CommentThread.tsx` — but neither PR attempted consolidation, so the codebase now has two near-identical copies of the same textarea-+-`execCommand` dance and 14 other sites that never got the treatment
> - This pull request consolidates both existing fallbacks into a single shared `copyTextToClipboard` utility at `ui/src/lib/clipboard.ts` and migrates every direct `navigator.clipboard.writeText` call site to use it
> - The benefit is: every copy icon starts working for self-hosted HTTP users, future copy buttons inherit the fallback for free, and the existing duplicated fallback logic collapses to one audited implementation

## What Changed

- New `ui/src/lib/clipboard.ts` exporting `copyTextToClipboard(text: string): Promise<void>`. Uses `navigator.clipboard.writeText` when both `navigator.clipboard` and `window.isSecureContext` are truthy; otherwise falls back to a transient on-screen-but-invisible textarea + `document.execCommand("copy")`. Rejects on failure so callers can surface a toast if they want to.
- Hardens the fallback vs. the two existing inlined copies: explicit `focus({ preventScroll: true })` before `.select()` (some browsers won't honor programmatic selection without focus first), `setSelectionRange(0, length)` as a secondary selection path, on-screen-but-`opacity: 0` + `zIndex: -1` + `pointerEvents: none` positioning instead of `left: -9999px` (off-screen elements are silently excluded from the document selection in some engines), and restore-focus to the previously-focused element so users mid-typing aren't bumped out of their inputs.
- New co-located test file `ui/src/lib/clipboard.test.ts` — 9 tests covering the native-API happy path, the fallback mounting/selecting/removing the textarea, `execCommand` failure propagation, `execCommand`-throws cleanup, focus restoration, and the "secure context but no navigator.clipboard" degenerate case (jsdom-style).
- `CopyText.tsx` and `CommentThread.tsx` now import the shared utility and drop their inlined fallbacks. Their existing tests continue to pass unchanged.
- 14 direct `navigator.clipboard.writeText(...)` call sites across `ui/src/components/` and `ui/src/pages/` migrated to `copyTextToClipboard(...)`. No behavioral changes beyond "now actually copies in non-secure contexts." Toast and copied-state feedback paths are untouched.

## Verification

**Automated:**
- `cd ui && pnpm tsc --noEmit` → clean (exit 0).
- `cd ui && pnpm vitest run` → 346/346 tests pass, 65/65 files, including the new 9 clipboard tests and the pre-existing `CommentThread.test.tsx` (which has its own `execCommand` mocking).

**Manual (on a self-hosted Paperclip reached over plain HTTP from another machine on the LAN):**
- Before the fix, on a currently-broken button (e.g. "Copy Agent ID" from the agent detail overflow menu, or the copy icon next to the workspace path on CompanySkills): click → clipboard unchanged, no toast, menu item gives no feedback. The click is indistinguishable from a disabled button.
- After patching the running install's `ui-dist/index.html` with an equivalent inline polyfill and restarting the service: click "Copy Agent ID" → paste elsewhere confirms the UUID is on the clipboard. Spot-checked several other previously-broken buttons (workspace path copy, invite snippet copy, agent token copy on rotation) — all now land text on the clipboard.
- Also confirmed the native path still works on `http://localhost:3100` (where `navigator.clipboard` is defined): utility short-circuits to `navigator.clipboard.writeText`, no textarea is ever mounted. The `does not mount any DOM nodes` unit test asserts this invariant so it can't silently regress.
- Verified the two already-fixed buttons (anything using `CopyText`, and "Copy as markdown" on comments which uses `CopyMarkdownButton`) continue to work identically after the refactor — their inlined fallbacks were replaced with calls to the shared utility, not behavior-changed.

## Risks

Low. The fallback path is strictly opt-in-on-failure — it only runs when `navigator.clipboard` is missing or `window.isSecureContext` is false, so secure-context (HTTPS/localhost) behavior is identical to before this PR. The execCommand path is the same technique already shipping inline in `CopyText` and `CommentThread`; we're not introducing a new clipboard mechanism, just centralizing and hardening one. `execCommand("copy")` is deprecated but universally implemented (MDN flags it as deprecated for *new* code, not broken for existing calls). No schema, API, or public-contract changes.

Focus restoration is the one new behavior beyond the existing fallbacks — if a user had a textarea/input focused and clicked a copy button, focus now returns there after the copy (previously the off-screen textarea would steal focus). That's an ergonomic improvement rather than a regression risk, but worth flagging.

## Model Used

Claude, `claude-opus-4-6` (1M context window), with extended tool use for file reads, writes, and running vitest/tsc. No reasoning mode beyond the default.

## Checklist

- [ ] If this change affects the UI, I have included before/after screenshots — the visual behavior is unchanged in secure contexts, and the fix isn't meaningfully visual in non-secure contexts (before: click does nothing, icon doesn't animate, clipboard unchanged; after: icon flips to check, clipboard populated). Can add a screen recording of paste-target-before vs. paste-target-after if useful.
- [x] I have updated relevant documentation to reflect my changes — no user-facing docs needed updating; the utility has inline documentation.

### DIFF
--- a/ui/src/components/CommentThread.tsx
+++ b/ui/src/components/CommentThread.tsx
@@ -19,6 +19,7 @@ import { OutputFeedbackButtons } from "./OutputFeedbackButtons";
 import { ApprovalCard } from "./ApprovalCard";
 import { AgentIcon } from "./AgentIconPicker";
 import { formatAssigneeUserLabel } from "../lib/assignees";
+import { copyTextToClipboard } from "../lib/clipboard";
 import type { IssueTimelineAssignee, IssueTimelineEvent } from "../lib/issue-timeline-events";
 import { timeAgo } from "../lib/timeAgo";
 import { cn, formatDateTime } from "../lib/utils";
@@ -210,27 +211,6 @@ function runStatusClass(status: string) {
   }
 }
 
-async function copyTextWithFallback(text: string) {
-  if (navigator.clipboard && window.isSecureContext) {
-    await navigator.clipboard.writeText(text);
-    return;
-  }
-
-  const textarea = document.createElement("textarea");
-  textarea.value = text;
-  textarea.style.position = "fixed";
-  textarea.style.left = "-9999px";
-  document.body.appendChild(textarea);
-
-  try {
-    textarea.select();
-    const success = document.execCommand("copy");
-    if (!success) throw new Error("execCommand copy failed");
-  } finally {
-    document.body.removeChild(textarea);
-  }
-}
-
 function CopyMarkdownButton({ text }: { text: string }) {
   const [status, setStatus] = useState<"idle" | "copied" | "failed">("idle");
   const timeoutRef = useRef<ReturnType<typeof setTimeout> | null>(null);
@@ -257,7 +237,7 @@ function CopyMarkdownButton({ text }: { text: string }) {
       title={label}
       aria-label="Copy comment as markdown"
       onClick={() => {
-        void copyTextWithFallback(text)
+        void copyTextToClipboard(text)
           .then(() => setStatus("copied"))
           .catch(() => setStatus("failed"));
 
--- a/ui/src/components/CopyText.tsx
+++ b/ui/src/components/CopyText.tsx
@@ -1,4 +1,5 @@
 import { useCallback, useRef, useState } from "react";
+import { copyTextToClipboard } from "@/lib/clipboard";
 import { cn } from "@/lib/utils";
 
 interface CopyTextProps {
@@ -18,23 +19,7 @@ export function CopyText({ text, children, className, copiedLabel = "Copied!" }:
 
   const handleClick = useCallback(async () => {
     try {
-      if (navigator.clipboard && window.isSecureContext) {
-        await navigator.clipboard.writeText(text);
-      } else {
-        // Fallback for non-secure contexts (e.g. HTTP on non-localhost)
-        const textarea = document.createElement("textarea");
-        textarea.value = text;
-        textarea.style.position = "fixed";
-        textarea.style.left = "-9999px";
-        document.body.appendChild(textarea);
-        try {
-          textarea.select();
-          const success = document.execCommand("copy");
-          if (!success) throw new Error("execCommand copy failed");
-        } finally {
-          document.body.removeChild(textarea);
-        }
-      }
+      await copyTextToClipboard(text);
       setLabel(copiedLabel);
     } catch {
       setLabel("Copy failed");
--- a/ui/src/components/IssueChatThread.tsx
+++ b/ui/src/components/IssueChatThread.tsx
@@ -41,6 +41,7 @@ import {
   type IssueChatTranscriptEntry,
   type SegmentTiming,
 } from "../lib/issue-chat-messages";
+import { copyTextToClipboard } from "../lib/clipboard";
 import { resolveIssueChatTranscriptRuns } from "../lib/issueChatTranscriptRuns";
 import type { IssueTimelineAssignee, IssueTimelineEvent } from "../lib/issue-timeline-events";
 import { Button } from "@/components/ui/button";
@@ -750,7 +751,7 @@ function CopyablePreBlock({ children, className }: { children: string; className
         title="Copy"
         aria-label="Copy"
         onClick={() => {
-          void navigator.clipboard.writeText(children).then(() => {
+          void copyTextToClipboard(children).then(() => {
             setCopied(true);
             setTimeout(() => setCopied(false), 2000);
           });
@@ -949,7 +950,7 @@ function IssueChatUserMessage() {
                     .filter((p): p is { type: "text"; text: string } => p.type === "text")
                     .map((p) => p.text)
                     .join("\n\n");
-                  void navigator.clipboard.writeText(text).then(() => {
+                  void copyTextToClipboard(text).then(() => {
                     setCopied(true);
                     setTimeout(() => setCopied(false), 2000);
                   });
@@ -1156,7 +1157,7 @@ function IssueChatAssistantMessage() {
                           .filter((p): p is { type: "text"; text: string } => p.type === "text")
                           .map((p) => p.text)
                           .join("\n\n");
-                        void navigator.clipboard.writeText(text);
+                        void copyTextToClipboard(text);
                       }}
                     >
                       <Copy className="mr-2 h-3.5 w-3.5" />
--- a/ui/src/components/IssueDocumentsSection.tsx
+++ b/ui/src/components/IssueDocumentsSection.tsx
@@ -12,6 +12,7 @@ import { useLocation } from "@/lib/router";
 import { ApiError } from "../api/client";
 import { issuesApi } from "../api/issues";
 import { useAutosaveIndicator } from "../hooks/useAutosaveIndicator";
+import { copyTextToClipboard } from "../lib/clipboard";
 import { deriveDocumentRevisionState } from "../lib/document-revisions";
 import { queryKeys } from "../lib/queryKeys";
 import { cn, relativeTime } from "../lib/utils";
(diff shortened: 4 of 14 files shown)


joshdutcher:
Good call — applied in 8abefe70. The native-API call is now wrapped in try/catch and falls through to `execCommand` on rejection, with updated JSDoc documenting the cascade. Replaced the "propagates rejection from the native API" test with two: one asserting the cascade path (native rejects, `execCommand` succeeds → overall resolves), and one asserting the terminal case (both paths fail → rejects with the `execCommand` error). 347/347 tests pass.

PR #3541, opened 2026-04-13
Clear stale execution locks when releasing issues

DansiDanutz:
## Thinking Path

> - Paperclip is the control plane that coordinates agent work through issues, runs, and explicit governance.
> - This change sits in the issue lifecycle service, specifically the release path that hands an issue back to `todo`.
> - DAN-207 exposed a control-plane gap: an assignee could release an issue while stale execution-lock fields still pointed at an old heartbeat run.
> - That left the task looking available in status terms while remaining uncheckoutable to the next agent, which breaks the single-assignee / atomic-checkout contract.
> - The safest fix is to make `release()` clear execution lock state alongside checkout ownership, then prove the exact stale-run handoff with a regression test.
> - This pull request does that so agent-to-agent handoff can recover without requiring board-side stale-run cleanup first.

## What Changed

- Cleared `executionRunId`, `executionAgentNameKey`, and `executionLockedAt` inside `server/src/services/issues.ts` when `release()` returns an issue to `todo`.
- Added a regression in `server/src/__tests__/issues-service.test.ts` covering the stale-lock path: release by the current assignee, then checkout by a different agent.
- Updated that test block's teardown to delete `heartbeatRuns`, which the new regression now creates.

## Verification

- `npx -y node@22 $(which pnpm) exec vitest run server/src/__tests__/issues-service.test.ts -t "clears execution locks on release so another agent can check out the issue"`
- `git diff --check`
- Attempted `npx -y node@22 $(which pnpm) build`, but the build fails before reaching this patch because `packages/adapter-utils` cannot resolve `@types/ws` from `/Users/davidai/Desktop/DavidAi/node_modules/.pnpm/...` in this clean clone.

## Risks

- Low behavioral risk: the service already clears checkout ownership on release; this extends that reset to the matching execution-lock fields.
- The only notable caveat is repo-wide verification debt in this environment: the clean-clone build currently stops in `packages/adapter-utils` before server packages build.

## Model Used

- OpenAI Codex CLI using `gpt-5.4`
- High reasoning effort
- Tool-assisted local execution in a clean fork clone under Node 22 via `npx node@22`

## Checklist

### DIFF
--- a/server/src/__tests__/issues-service.test.ts
+++ b/server/src/__tests__/issues-service.test.ts
@@ -8,6 +8,7 @@ import {
   companies,
   createDb,
   executionWorkspaces,
+  heartbeatRuns,
   instanceSettings,
   issueComments,
   issueInboxArchives,
@@ -66,6 +67,7 @@ describeEmbeddedPostgres("issueService.list participantAgentId", () => {
     await db.delete(issueInboxArchives);
     await db.delete(activityLog);
     await db.delete(issues);
+    await db.delete(heartbeatRuns);
     await db.delete(executionWorkspaces);
     await db.delete(projectWorkspaces);
     await db.delete(projects);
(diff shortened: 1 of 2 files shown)


stefanocampana (commented):
This fixes the same issue as #3559 (clearing execution lock fields on release). Both PRs add `executionRunId: null`, `executionAgentNameKey: null`, `executionLockedAt: null` to the release path.

#3559 is slightly more focused (dedicated test file vs extending existing test suite). Suggest coordinating with @bloodycoder to close one of these to avoid merge conflicts.

DansiDanutz:
Closing this in favor of #3559. It addresses the same release-path execution-lock bug, and it already has human approval on the focused version of the fix. No further changes needed from this branch unless new review context appears.

PR #3546, opened 2026-04-13
fix(db,plugins,shared): bootstrap dev DB and plugin workers

avirajkhare00:
## Thinking Path

> - Paperclip is a control-plane product: API, board UI, plugins, and DB tooling share the monorepo.
> - Local dev uses embedded PostgreSQL when `DATABASE_URL` is unset, and plugin workers are separate Node processes that import `@paperclipai/plugin-sdk` (which re-exports `@paperclipai/shared`).
> - On macOS, embedded Postgres failed to start when the bundled binary could not load `libzstd`; `pnpm db:migrate` also did not see a repo-root `.env` because resolution only read process env and `~/.paperclip/.../.env`, not `.env` found by walking up from `cwd` (e.g. `packages/db`).
> - `@paperclipai/shared` workspace `exports` pointed at TypeScript sources; Node then loaded `src/index.ts` and failed on `.js` import specifiers that only exist in `dist/`.
> - The Plugin Manager install dialog sent absolute paths as npm package names when `isLocalPath` was omitted, causing bogus `npm install` under `~/.paperclip/plugins`.
> - This pull request aligns shared workspace exports with compiled output, resolves `DATABASE_URL` from ancestor `.env`, infers local plugin installs for path-shaped payloads, improves UI and docs for the above failure modes.
> - The benefit is reliable local bootstrap: migrations, dev server, and plugin activation without broken resolution or misleading install behavior.

## What Changed

- `@paperclipai/shared`: workspace `exports`, `main`, and `types` now target `dist/` (same layout as publish) so plain Node (plugin workers) loads compiled ESM.
- `packages/db`: resolve `DATABASE_URL` from the nearest `.env` walking up from `process.cwd()`; extend embedded-Postgres error hints for `dyld` / `libzstd`; add regression test and isolate existing tests from accidental repo-root `.env` pickup.
- `server`: treat `packageName` as a local path when it looks like a filesystem path and `isLocalPath` is not explicitly `false`.
- `ui`: clarify install dialog copy; pass `isLocalPath: true` when the entered value looks like a path.
- `doc/DEVELOPING.md`: document macOS embedded-Postgres / Docker Postgres workaround.

## Verification

- `pnpm --filter @paperclipai/shared build`
- `pnpm --filter @paperclipai/db typecheck`
- `pnpm --filter @paperclipai/db exec vitest run src/runtime-config.test.ts`
- Manual: repo `.env` with `DATABASE_URL` + `docker compose -f docker/docker-compose.yml up -d db` (with `BETTER_AUTH_SECRET` set for compose), then `pnpm db:migrate` from repo root; confirm migrate uses `repo-dotenv` in log line.
- Manual: Plugin Manager → install with absolute path to a built example plugin without toggling flags; confirm install succeeds (server infers local path).

## Risks

- **Shared:** Any workflow that expected Node to consume `@paperclipai/shared` directly from `src/` without building must run `pnpm --filter @paperclipai/shared build` first (consistent with published npm behavior).
- **Install inference:** An extremely unusual npm name that looks like an absolute path could be misclassified if a client omits `isLocalPath`; clients can still send `isLocalPath: false` to force npm.

## Model Used

Cursor / Composer — GPT-5.1 (agent mode); assisted with implementation and PR text. Exact provider model ID as shown in the Cursor UI.

## Checklist

### DIFF
--- a/docs/adapters/codex-local.md
+++ b/docs/adapters/codex-local.md
@@ -20,6 +20,7 @@ The `codex_local` adapter runs OpenAI's Codex CLI locally. It supports session p
 | `env` | object | No | Environment variables (supports secret refs) |
 | `timeoutSec` | number | No | Process timeout (0 = no timeout) |
 | `graceSec` | number | No | Grace period before force-kill |
+| `fastMode` | boolean | No | Enables Codex Fast mode. Currently supported on `gpt-5.4` only and burns credits faster |
 | `dangerouslyBypassApprovalsAndSandbox` | boolean | No | Skip safety checks (dev only) |
 
 ## Session Persistence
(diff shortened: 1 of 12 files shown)


stefanocampana (changes requested):
This PR mixes several unrelated concerns (+1499 lines, 25 files):

1. **DB runtime config** — repo-root `.env` discovery, `resolveDatabaseTarget` improvements, embedded Postgres error hints for libzstd/dyld → good, should be its own PR
2. **Plugin HTTP fetch loopback** — allowing `http.fetch` to private targets in `local_trusted` mode → good security boundary change, deserves its own review
3. **Agent Commerce Bridge plugin** — entire example plugin with React UI, esbuild/rollup config, worker → should be a separate PR
4. **Documentation updates** — scattered across all three changes

Each of these is individually reasonable but reviewing them together makes it hard to isolate risk. Please split into 2-3 focused PRs.

PR #3563, opened 2026-04-13
fix(routines): coalesce into stranded execution issue instead of failing dispatch

zlfnhmd:
## Thinking Path

> - Routines are recurring tasks that fire on a schedule; each run creates an execution issue assigned to an agent
> - The `issues_open_routine_execution_uq` partial unique index ensures only one open execution issue exists per routine at a time — but only when `executionRunId IS NOT NULL`
> - `executionRunId` is stamped on the issue when a heartbeat run is queued for it (via `queueIssueAssignmentWakeup`)
> - If the heartbeat run finishes but the agent never closes the issue, the issue stays open with a stale `executionRunId`
> - The pre-check (`findLiveExecutionIssue`) only finds issues with a *live* heartbeat run — so it returns null for stranded issues
> - This causes the dispatch to attempt creating a new issue; the heartbeat wakeup then tries to stamp `executionRunId` on it, hitting the unique constraint — causing the routine run to be marked `failed`
> - This PR adds `findStrandedOpenExecutionIssue` and uses it as a fallback in the pre-check so the dispatch coalesces into the stranded issue instead

## What Changed

- **`server/src/services/routines.ts`**: Added `findStrandedOpenExecutionIssue` — finds any open routine execution issue where `executionRunId IS NOT NULL`, regardless of heartbeat run status. Used as a fallback in the pre-check and the constraint catch handler.
- **`server/src/__tests__/routines-service.test.ts`**: Added regression test: seeds an execution issue with a *finished* heartbeat run (simulating an agent that ran but never closed the issue), verifies `activeIssue` is null (live check returns null), then confirms the next dispatch returns `status: "coalesced"` rather than `status: "failed"`.

## Verification

```bash
pnpm vitest run server/src/__tests__/routines-service.test.ts
# 18 tests pass (including the new regression test)

pnpm vitest run server/src/__tests__/routines-e2e.test.ts
# 4 tests pass
```

**Production confirmation:** This bug was observed in a live deployment. A routine execution issue was left in `todo` with a finished heartbeat run. Every subsequent scheduled and manual run failed with `duplicate key value violates unique constraint "issues_open_routine_execution_uq"`. This fix prevents the failure path by coalescing into the stranded issue.

## Risks

- **Low risk.** The fallback only applies to issues where `executionRunId IS NOT NULL` (heartbeat was queued but is no longer live). Issues with `executionRunId IS NULL` (truly idle) still trigger fresh issue creation as before.
- Gated on `concurrencyPolicy !== "always_enqueue"`, consistent with existing live-issue coalescing logic.
- Constraint catch handler retained as defense-in-depth for concurrent race conditions.

## Model Used

- **Provider:** Anthropic
- **Model:** Claude Sonnet 4.6 (`claude-sonnet-4-6`)
- **Mode:** Tool use / agentic (running as a Paperclip Ops Agent inside a heartbeat run)

### DIFF
--- a/server/src/__tests__/routines-service.test.ts
+++ b/server/src/__tests__/routines-service.test.ts
@@ -349,6 +349,75 @@ describeEmbeddedPostgres("routine service live-execution coalescing", () => {
     expect(routineIssues[0]?.id).toBe(previousIssue.id);
   });
 
+  it("coalesces when the existing execution issue has a stale (finished) heartbeat run but is still open", async () => {
+    const { agentId, companyId, issueSvc, routine, svc } = await seedFixture();
+    const previousRunId = randomUUID();
+    const finishedHeartbeatRunId = randomUUID();
+
+    const previousIssue = await issueSvc.create(companyId, {
+      projectId: routine.projectId,
+      title: routine.title,
+      description: routine.description,
+      status: "todo",
+      priority: routine.priority,
+      assigneeAgentId: routine.assigneeAgentId,
+      originKind: "routine_execution",
+      originId: routine.id,
+      originRunId: previousRunId,
+    });
+
+    await db.insert(routineRuns).values({
+      id: previousRunId,
+      companyId,
+      routineId: routine.id,
+      triggerId: null,
+      source: "schedule",
+      status: "issue_created",
+      triggeredAt: new Date("2026-03-20T12:00:00.000Z"),
+      linkedIssueId: previousIssue.id,
+      completedAt: new Date("2026-03-20T12:00:00.000Z"),
+    });
+
+    // Simulate: heartbeat run was queued and has since finished, but the issue was never closed.
+    await db.insert(heartbeatRuns).values({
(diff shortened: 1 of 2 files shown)


zlfnhmd:
@cryppadotta please take a look at this

zlfnhmd:
@aronprins

serenakeyitan:
🌱 **gardener:verdict:** `ALIGNED` · severity: `low` · commit: `29107dfc`

Aligned with context tree. Consistent with surfacing problems without hiding them (product decision #6) and the routines dispatch model.

<sub>Reviewed commit: <code>29107dfc</code> · Tree snapshot: <code>4a91f3cc</code> · Commands: <code>@gardener re-review</code> · <code>@gardener pause</code> · <code>@gardener ignore</code></sub>

cryppadotta:
right - agree. i have some changes that touch this in the latest release, let me ask my clips if this overlaps

PR #3611, opened 2026-04-13
fix(ui): prevent keyboard collapse and typing delay on mobile

kbecking:
## Summary

Fixes two mobile issues reported after PRs #23-25:

- **Keyboard collapse**: `decorateProjectMentions()` was unconditionally clearing and re-applying mention chip decorations (including `contenteditable="false"` attribute toggling) on every MutationObserver callback and compositionend flush. Android interprets these attribute changes as focus changes and dismisses the keyboard. Fix: skip decoration when already applied, and add a `isFlushingCompositionRef` guard to suppress MutationObserver re-entry during the compositionend rAF callback.

- **Typing delay on longer issues**: `handleFeedbackVote` in `CommentThread` was a plain function recreated every render. Since it was passed as the `onVote` prop to the `memo`'d `TimelineList`, every keystroke caused the entire comment timeline (all comments, events, runs) to re-render. Fix: wrap in `useCallback` so `TimelineList`'s memo comparison succeeds across keystroke-driven re-renders.

## Changed files

- `ui/src/components/MarkdownEditor.tsx` — skip already-decorated links, add flush guard
- `ui/src/components/CommentThread.tsx` — `useCallback` for `handleFeedbackVote`

## Test plan

- [ ] Type on Android with word suggestions / autocorrect after inserting a @mention — keyboard should stay open
- [ ] Open a long issue (20+ comments) and type in the comment editor — no noticeable delay
- [ ] Verify mention chips still render correctly on new mention links
- [ ] Run `pnpm --filter @paperclipai/ui typecheck` — passes
- [ ] Run `pnpm --filter @paperclipai/ui exec vitest run` — passes


kbecking:
Closing — opened on upstream by mistake. Recreating on fork.

PR #3670, opened 2026-04-14
fix(docker): remove stale SHA256 check for GitHub CLI keyring

ziadsawalha:
## Summary
- The GPG keyring file at `cli.github.com/packages/githubcli-archive-keyring.gpg` has changed upstream, causing the hardcoded `sha256sum -c` check in the Dockerfile to fail
- Removes the redundant SHA256 check — the signed apt repository already provides integrity verification via the GPG key itself
- Docker build was failing at line 8 of the Dockerfile

## Test plan
- [ ] `docker build .` completes successfully
- [ ] `gh --version` works inside the container


ziadsawalha:
Good catch — the force-pushed revision already addresses this. The SHA is updated to the current value (`6084d5d7bd8e...`) rather than removed, keeping the defense-in-depth pin intact.

PR #3698, opened 2026-04-15
fix(skills): add model spec and NEVER guards to paperclip skill family

projetsjsl:
## Summary
- Add model spec and NEVER guards to paperclip skill family for safer execution
- Prevents skill family from running without explicit model context

## Test plan
- [ ] Verify skill invocation respects NEVER guards
- [ ] Verify model spec is picked up correctly

### DIFF
--- a/skills/paperclip-create-agent/SKILL.md
+++ b/skills/paperclip-create-agent/SKILL.md
@@ -1,5 +1,6 @@
 ---
 name: paperclip-create-agent
+model: claude-sonnet-4-6
 description: >
   Create new agents in Paperclip with governance-aware hiring. Use when you need
   to inspect adapter configuration options, compare existing agent configs,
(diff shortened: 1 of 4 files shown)


xkonjin (approved):
Docs-only change adding model specs and NEVER guards to the paperclip skill family.

What works well
- The NEVER bullets are concrete and cite why the rule matters (heartbeat budget, governance rejection, secret visibility, audit gaps). This is much more useful than generic do-nots.
- Adding explicit model: claude-sonnet-4-6 to the frontmatter removes ambiguity for downstream consumers that parse skill metadata.
- para-memory-files NEVER rules catch real failure modes (portability, index drift, silent overwrites).

Concerns / suggestions
1. Consistency — paperclip-create-agent and paperclip-create-plugin get model fields, but paperclip and para-memory-files already had none before. Are there other skills in the family that should also be pinned? If this is meant to be a rollout, a tracking issue would help.
2. Testability — there is no automated check that enforces these NEVER rules. Consider a CI lint step that greps SKILL.md files for required frontmatter keys (model) and a NEVER section. Without enforcement, docs rot fast.
3. Scope of NEVER in paperclip-create-agent — "NEVER create an agent with a timer heartbeat by default" is good advice, but there are legitimate long-polling / watchdog agents that need timers. A future reader might interpret this as an absolute ban. Maybe soften to "Avoid timer heartbeats by default; document the workload reason when one is required."

Nit
- One bullet says "NEVER submit a hire request without verifying the reporting line is within the same company." Since the PR only touches docs, I will note that the actual API does not enforce this client-side, so this is a good warning.

Approve with minor suggestions.

PR #3712, opened 2026-04-15
fix: capture rawBody for form-encoded requests and deduplicate plugin event delivery

riptex:
## Summary

Two bugs affecting Slack plugin (and potentially other webhook-based plugins) discovered during local testing.

### Bug 1: Slack slash commands always fail HMAC signature verification

**File:** `server/src/app.ts`

`express.json()` is the only body-parsing middleware, and its `verify` callback captures `req.rawBody`. Slack slash commands arrive as `application/x-www-form-urlencoded`, so `express.json()` skips them entirely — `req.rawBody` is never populated. The Slack plugin's HMAC signature verification then computes the HMAC over an empty body, which never matches Slack's real signature.

**Fix:** Add `express.urlencoded()` with the same `verify` callback immediately after `express.json()`, guarded by `if (!r.rawBody)` so JSON requests continue to use the JSON middleware's captured bytes.

### Bug 2: Plugin event bus delivers the same event multiple times to the same worker

**File:** `server/src/services/plugin-host-services.ts`

Each `ctx.events.on(pattern, handler)` call in a plugin worker registers a separate server-side subscription in `buildHostServices`. When an event fires, the bus calls **all** matching subscriptions for that plugin — each independently sending `notifyWorker("onEvent", { event })` via IPC. The worker receives N copies of the same event (one per registered subscription) and dispatches all its registered handlers each time, producing N×M invocations instead of M.

**Observed symptom:** A Slack plugin with two `issue.created` subscriptions produced 2 Slack notifications per issue creation instead of 1.

**Fix:** In the `subscribe` host service, deduplicate `notifyWorker` calls using the event's `eventId` with a short sliding-window cache (10 seconds). This collapses N notifications down to 1 for the same event ID within the dedup window.

## Test plan

- [ ] Slack slash command (`/clip`) correctly verifies HMAC signature and returns a response
- [ ] A plugin with multiple `ctx.events.on("issue.created", ...)` subscriptions receives each `issue.created` event exactly once (not once per subscription)
- [ ] JSON webhook payloads (non-form-encoded) still have `rawBody` populated correctly
- [ ] Events without an `eventId` are still delivered normally (no regression for unkeyed events)

### DIFF
--- a/server/src/app.ts
+++ b/server/src/app.ts
@@ -122,6 +122,17 @@ export async function createApp(
       (req as unknown as { rawBody: Buffer }).rawBody = buf;
     },
   }));
+  // Capture rawBody for form-encoded requests (e.g. Slack slash commands).
+  // express.json() only captures rawBody for application/json payloads; without
+  // this middleware, HMAC signature verification fails for any webhook that
+  // sends application/x-www-form-urlencoded (the Slack slash command format).
+  app.use(express.urlencoded({
+    extended: true,
+    verify: (req, _res, buf) => {
+      const r = req as unknown as { rawBody?: Buffer };
+      if (!r.rawBody) r.rawBody = buf;
+    },
+  }));
   app.use(httpLogger);
   const privateHostnameGateEnabled =
     opts.deploymentMode === "authenticated" && opts.deploymentExposure === "private";
--- a/server/src/services/plugin-host-services.ts
+++ b/server/src/services/plugin-host-services.ts
@@ -460,6 +460,19 @@ export function buildHostServices(
   const assets = assetService(db);
   const scopedBus = eventBus.forPlugin(pluginKey);
 
+  // Deduplicate onEvent deliveries to the worker. A plugin may register
+  // multiple ctx.events.on() handlers for the same event pattern (e.g. one
+  // for notifications and one for watch-storage). Each subscription creates a
+  // server-side handler that calls notifyWorker("onEvent"). Without
+  // deduplication the worker receives N copies of the same event (one per
+  // matching subscription) and dispatches all its registered callbacks each
+  // time, resulting in N×M invocations instead of M.
+  //
+  // Fix: track recently-sent eventIds and skip duplicate notifyWorker calls
+  // within a 10-second window. The worker still dispatches to all its locally
+  // registered handlers on the single delivery it receives.
+  const recentEventIds = new Map<string, number>(); // eventId -> sentAt (ms)
+
   // Track active session event subscriptions for cleanup
   const activeSubscriptions = new Set<{ unsubscribe: () => void; timer: ReturnType<typeof setTimeout> }>();
   let disposed = false;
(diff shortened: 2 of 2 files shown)


serenakeyitan:
🌱 **gardener** · rocket `ALIGNED` · severity: `low` · commit: `85d8a3a`

### Context fit
<details open>
<summary><strong>Context match</strong></summary>

| **Plugin Event Bus** | Deduplicates event delivery by `eventId` within a 10-second window | `PluginEventBus` routes domain events to subscribed plugins; tree doesn't prescribe dedup strategy | ✅ Implementation detail; fixes real duplicate-notification bug |
| **Plugin Isolation** | In-memory cache for dedup; doesn't affect process isolation | "One process per plugin"; capability enforcement; data-backed state | ✅ Cache is transient; acceptable for collision prevention |
| **Webhook Body Handling** | Adds `express.urlencoded()` with `verify` callback to capture `rawBody` | Backend middleware contract; no explicit tree guidance on form-encoded payloads | ✅ Enables Slack plugin HMAC verification; isolated to body parsing |
| **eventId-less events** | Events without `eventId` skip dedup and can still duplicate | Tree doesn't mandate eventId presence on all events | ⚠️ Documented gap; not a blocker but worth acknowledging |

</details>

<details>

- [`plugins/NODE.md`](https://github.com/serenakeyitan/paperclip-tree/blob/main/plugins/NODE.md) — plugin SDK and runtime architecture
- [`plugins/runtime/NODE.md`](https://github.com/serenakeyitan/paperclip-tree/blob/main/plugins/runtime/NODE.md) — event bus, lifecycle, worker isolation
- [`engineering/NODE.md`](https://github.com/serenakeyitan/paperclip-tree/blob/main/engineering/NODE.md) — backend architecture and service patterns

</details>

### Summary

This PR fixes two real bugs that prevent the Slack plugin from working correctly:

1. **Slack HMAC verification** — `express.json()` was the only body parser, so form-encoded Slack slash commands never had `rawBody` populated. HMAC verification always failed. The fix adds `express.urlencoded()` with a `verify` callback alongside the JSON parser.

2. **Plugin event duplication** — When a plugin registers multiple `ctx.events.on()` subscriptions for the same event pattern (e.g., two `issue.created` handlers), the event bus delivered N copies of the same event to the worker. The fix deduplicates via an `eventId`-based Map with a 10-second sliding window, collapsing N deliveries into 1.

Both changes are **pragmatic and localized**:
- They don't affect plugin process isolation, capability enforcement, or data persistence (which the tree marks as core).
- The in-memory dedup cache is acceptable for transient collision prevention.
- The documented gap (eventId-less events can still duplicate) is acknowledged and not a blocker.

✅ **Fit:** This PR aligns with the plugin architecture as described in the tree. It fixes bugs at the boundary between the backend webhook handler and the plugin event bus without violating the architectural constraints (process isolation, capability gating, plugin-worker contract).

---
<sub>Reviewed commit: <code>85d8a3a</code> · Tree snapshot: <code>6e68dbe</code> · Commands: <code>@gardener re-review</code> · <code>@gardener pause</code> · <code>@gardener ignore</code></sub>

xkonjin (commented):
Thanks for the fix. A few notes:

1. **Event deduplication memory bound**: the `recentEventIds` Map is pruned only when a *new* eventId arrives. During a burst of many unique events within 10s the Map can grow without bound. Consider adding a hard size ceiling (e.g. 1000 entries) and evicting the oldest when it is exceeded.

2. **Test coverage gap**: there are no tests for the deduplication logic itself. A regression test that fires two subscriptions with the same eventId and asserts the worker is only notified once would prevent future breakage.

3. **rawBody middleware**: the new `express.urlencoded()` middleware is placed *after* `express.json()`. If a request has `Content-Type: application/x-www-form-urlencoded`, `express.json()` will skip it, so `rawBody` will still be captured by the urlencoded middleware. That is fine, but verify that no earlier middleware consumes the stream before this point.

4. **Race condition?**: `recentEventIds` is local to the host-services builder. If the same plugin is loaded in multiple workers, each gets its own Map. That is probably intended, but worth confirming.

Overall looks correct and addresses a real bug. Please add a test for the deduplication path before merging.

PR #3723, opened 2026-04-15
fix: issue identifier regex rejects prefixes with digits

stubbi:
## Thinking Path

> - Issues are the primary unit of work agents, humans, and the board coordinate on
> - Issues are addressed by human-readable identifiers like `CORE-42` — prefix-plus-number — alongside their UUIDs
> - Several routes accept either form and use a regex to decide which path to take
> - That regex (`^[A-Z]+-\d+$`) rejects any project prefix containing a digit (e.g. `CORE2-1`, `ABC2-42`)
> - Users with such prefixes get 404 on any identifier-based lookup
> - This pull request changes the regex to `^[A-Z][A-Z0-9]*-\d+$` at every call site so those identifiers resolve correctly

## What Changed

- `server/src/services/issues.ts` — `issueService.getById`
- `server/src/routes/issues.ts` — `normalizeIssueIdentifier`
- `server/src/routes/activity.ts` — `resolveIssueByRef`
- `server/src/routes/agents.ts` — live-runs + active-run lookups (2 sites)

The pattern still anchors at both ends and requires a leading letter, so UUIDs and pure-numeric tokens continue to fall through to the UUID path.

## Verification

- Existing tests pass.
- Manual: hit `GET /companies/:id/issues/TESCR9-1` against a company whose prefix is `TESCR9`; returns the issue instead of 404.

## Risks

Low. The new pattern is a strict superset of the old one — every input the old regex matched, the new one matches identically. The only newly-accepted inputs are identifiers with digit-bearing prefixes, which is precisely the fix.

## Model Used

## Checklist

- [x] Model used specified
- [x] CI green

### DIFF
--- a/server/src/routes/activity.ts
+++ b/server/src/routes/activity.ts
@@ -24,7 +24,7 @@ export function activityRoutes(db: Db) {
   const issueSvc = issueService(db);
 
   async function resolveIssueByRef(rawId: string) {
-    if (/^[A-Z]+-\d+$/i.test(rawId)) {
+    if (/^[A-Z][A-Z0-9]*-\d+$/i.test(rawId)) {
       return issueSvc.getByIdentifier(rawId);
     }
     return issueSvc.getById(rawId);
--- a/server/src/routes/agents.ts
+++ b/server/src/routes/agents.ts
@@ -2402,7 +2402,7 @@ export function agentRoutes(db: Db) {
   router.get("/issues/:issueId/live-runs", async (req, res) => {
     const rawId = req.params.issueId as string;
     const issueSvc = issueService(db);
-    const isIdentifier = /^[A-Z]+-\d+$/i.test(rawId);
+    const isIdentifier = /^[A-Z][A-Z0-9]*-\d+$/i.test(rawId);
     const issue = isIdentifier ? await issueSvc.getByIdentifier(rawId) : await issueSvc.getById(rawId);
     if (!issue) {
       res.status(404).json({ error: "Issue not found" });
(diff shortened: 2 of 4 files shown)


serenakeyitan:

**Summary:** Fixes the issue identifier regex to accept prefixes with digits (e.g., `P2P-42`, `V2-100`). Changes `^[A-Z]+-\d+$` to `^[A-Z][A-Z0-9]*-\d+$` across 4 files.

**Tree context:**
- `product/task-system/NODE.md` — Issues use human-readable identifiers like `PREFIX-NUMBER`.

**Observations:**
1. Consistent fix across all 4 locations where the regex appears (activity, agents, issues routes, issues service).
2. Minimal change (5 additions, 5 deletions). Each regex updated identically.
3. First character must still be a letter (`[A-Z]`), so pure-numeric prefixes are still rejected, which is correct.

**Verdict:** Straightforward bug fix. No tree impact.

stubbi:
Closing — master has absorbed the fix through a different shape.

The ad-hoc `/^[A-Z]+-\d+$/i` regex that this PR targets is no longer present anywhere in `server/src/` — every call site has been migrated to the `normalizeIssueIdentifier()` helper in `packages/shared/src/issue-references.ts`, which is built on:

```ts
// packages/shared/src/issue-references.ts:1
export const ISSUE_REFERENCE_IDENTIFIER_RE = /^[A-Z][A-Z0-9]*-\d+$/;
```

Verified against the case from the original PR — `normalizeIssueIdentifier("TESCR9-1")` returns `"TESCR9-1"` rather than `null`. The accompanying test `packages/shared/src/issue-references.test.ts:13` covers `"pc1a2-7" → "PC1A2-7"`.

Thanks for the original review attention; the work landed via the helper migration. Closing to clear the queue.

PR #3729, opened 2026-04-15
feat: support TRUST_PROXY env var for correct client IP behind proxies

stubbi:
## Thinking Path

> - Most deployments sit behind a reverse proxy (Kubernetes Ingress, Nginx, Traefik, Cloudflare, an ALB, or a docker-compose proxy sidecar)
> - Behind any such proxy, `req.ip` is the proxy's address unless Express's `trust proxy` setting is configured
> - Any downstream logic that keys on `req.ip` — rate limiting, audit logs, abuse detection — groups every user behind the same proxy into a single bucket, which is both a correctness bug and a latent security issue
> - Express's `trust proxy` accepts multiple formats; the only safe defaults without user input are `false` (don't trust) or a positive integer (trust N hops)
> - This pull request adds a `TRUST_PROXY` env var that drives Express's setting, with strict validation that rejects unrecognized values instead of silently falling through to trust-all

## What Changed

- `server/src/app.ts`:
  - Read `TRUST_PROXY` from the environment immediately after `const app = express()`.
  - `"true"` → `app.set("trust proxy", true)`. Reserved for deployments that guarantee `X-Forwarded-For` can't be spoofed.
  - Positive-integer strings → `app.set("trust proxy", n)`. The integer must round-trip to the same string so leading-zero / whitespace-laden values are rejected.
  - Unset / `""` / `"false"` / `"0"` → disabled (unchanged default).
  - Anything else throws at startup with a clear error message.

## Verification

- `TRUST_PROXY=1 pnpm dev` behind a local Nginx: `req.ip` in a request-logging middleware reports the real client IP instead of 127.0.0.1.
- `TRUST_PROXY=loopback pnpm dev`: server refuses to start with `Invalid TRUST_PROXY value "loopback". Expected a positive integer (e.g. "1") or "true".`.
- `TRUST_PROXY` unset: unchanged.

## Risks

Low. Strict validation means accidental misconfiguration (typos, Express's named subnet strings) fail loud at startup rather than silently weakening proxy trust. No behavior change when the env var is unset.

## Model Used

## Checklist

- [x] Model used specified
- [x] CI green

### DIFF
--- a/server/src/app.ts
+++ b/server/src/app.ts
@@ -122,15 +122,28 @@ export async function createApp(
   // logs, abuse detection) groups every user behind the same proxy into
   // a single bucket.
   //
-  // Accepted values (see Express docs):
-  //   - number  — trust N hops of proxies (e.g. "1" for a single ingress)
-  //   - "true"  — trust all proxies (use only when deployment guarantees
-  //               X-Forwarded-For cannot be spoofed)
-  //   - unset / "" / "false" / "0" — disable (default)
+  // Accepted values (strict — anything else is an error so we never
+  // silently escalate to trust-all):
+  //   - positive integer — trust N hops of proxies (e.g. "1" for a
+  //     single ingress)
+  //   - "true"            — trust all proxies. Only safe when the
+  //     deployment guarantees X-Forwarded-For cannot be spoofed.
+  //   - unset / "" / "false" / "0" — disabled (default)
   const trustProxy = process.env.TRUST_PROXY;
   if (trustProxy && trustProxy !== "false" && trustProxy !== "0") {
-    const parsed = Number.parseInt(trustProxy, 10);
-    app.set("trust proxy", Number.isFinite(parsed) && parsed > 0 ? parsed : true);
+    if (trustProxy === "true") {
+      app.set("trust proxy", true);
+    } else {
+      const parsed = Number.parseInt(trustProxy, 10);
+      if (Number.isFinite(parsed) && parsed > 0 && String(parsed) === trustProxy) {
+        app.set("trust proxy", parsed);
+      } else {
+        throw new Error(
+          `Invalid TRUST_PROXY value "${trustProxy}". ` +
+            `Expected a positive integer (e.g. "1") or "true".`,
+        );
+      }
+    }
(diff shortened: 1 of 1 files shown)


stubbi:

stubbi:
@greptileai review

serenakeyitan:

**Summary:** Adds support for `TRUST_PROXY` env var so Express correctly resolves client IPs from `X-Forwarded-For` behind reverse proxies. Without this, rate limiting and audit logs group all users behind a proxy into a single IP.

**Tree context:**
- `infrastructure/deployment/NODE.md` — Most deployments sit behind a reverse proxy. This is a standard operational requirement.
- `engineering/backend/NODE.md` — Express app configuration in `server/src/app.ts`.

**Observations:**
1. Single-file change (31 lines) in `server/src/app.ts`.
2. Accepts standard Express `trust proxy` values: boolean, number (hop count), comma-separated CIDRs, or "loopback"/"linklocal"/"uniquelocal".
3. Well-commented with security implications of each value type.

**Verdict:** Standard infrastructure fix for proxy-aware deployments. No tree impact.

stubbi:
Superseding with #5872.

A quality re-audit flagged that the parser here only accepts `"true"` or a positive integer. That's narrower than Express 5's `trust proxy` model and forces operators into one of two unsafe defaults:

- **`true`** — blindly trusts any `X-Forwarded-For` header. Trivially spoofable by any client in deployments where the LB isn't strictly the only ingress.
- **hop count** — brittle if the topology adds another hop (e.g. Cloudflare in front of an existing ingress).

The safest configuration — trust only loopback, or only the LB's actual CIDR — is unreachable under this PR's surface because the named-subnet tokens (`loopback`, `linklocal`, `uniquelocal`) and CIDR lists are explicitly rejected.

#5872 keeps everything that works here (boolean `"true"`, strict positive int, the round-trip rejection of `"01"` / whitespace) and adds Express 5's full grammar: comma-separated lists of CIDRs and named subnets. Default unset behavior is unchanged.

Closing this one in favor of the wider replacement. Thanks for the original review — the strict-integer behavior carries over.

PR #3733, opened 2026-04-15
feat(adapters): widen onLog stream type to include "system"

stubbi:
## Problem

The log storage and presentation layers already accept \`\"system\"\` as a stream value:

- \`server/src/services/run-log-store.ts\`
- \`server/src/services/workspace-operation-log-store.ts\`
- \`server/src/services/workspace-operations.ts\`
- \`packages/shared/src/types/heartbeat.ts\`
- \`packages/plugins/sdk/src/types.ts\`
- \`ui/src/adapters/transcript.ts\` / \`ui/src/pages/AgentDetail.tsx\`
- \`cli/src/commands/heartbeat-run.ts\`
- \`doc/spec/agent-runs.md\`

But the callbacks adapters are actually handed — \`AdapterExecutionContext.onLog\` and the \`runChildProcess\` onLog — are still the narrow \`\"stdout\" | \"stderr\"\`. There's no typed way for an adapter to emit a system-level line (setup, cleanup, injection notices, internal state transitions) distinct from the target process's stdout/stderr, which means adapter authors end up smuggling those through stderr with a prefix.

## Fix

Widen the adapter-facing \`onLog\` signatures to \`\"stdout\" | \"stderr\" | \"system\"\` and propagate the widening through:

- \`packages/adapter-utils/src/types.ts\` (\`AdapterExecutionContext.onLog\`)
- \`packages/adapter-utils/src/server-utils.ts\` (\`runChildProcess\` opts)
- \`server/src/adapters/utils.ts\` (the re-exported wrapper)
- \`server/src/services/heartbeat.ts\` (the callback heartbeatService builds)
- \`server/src/services/workspace-runtime.ts\` (ensure/link helpers)
- \`server/src/routes/projects.ts\`, \`server/src/routes/execution-workspaces.ts\` (callsite onLog accumulators)
- \`packages/adapters/claude-local/src/server/execute.ts\` (\`runClaudeLogin\` input)
- \`packages/adapters/pi-local/src/server/execute.ts\` (\`bufferedOnLog\` — JSONL buffering only applies to stdout, so \"system\" passes through immediately like stderr did before)
- \`docs/adapters/creating-an-adapter.md\`

## Risk

No behavioral change for adapters or routes that only emit stdout/stderr; the third variant is ignored by existing callers. Adapters that want to emit system-level events can now do so without the string value being rejected at the type layer.

### DIFF
--- a/docs/adapters/creating-an-adapter.md
+++ b/docs/adapters/creating-an-adapter.md
@@ -101,7 +101,7 @@ interface AdapterExecutionContext {
   runtime: { sessionId: string | null; sessionParams: Record<string, unknown> | null };
   config: Record<string, unknown>;      // agent's adapterConfig
   context: Record<string, unknown>;      // task, wake reason, etc.
-  onLog: (stream: "stdout" | "stderr", chunk: string) => Promise<void>;
+  onLog: (stream: "stdout" | "stderr" | "system", chunk: string) => Promise<void>;
   onMeta?: (meta: AdapterInvocationMeta) => Promise<void>;
   onSpawn?: (meta: { pid: number; startedAt: string }) => Promise<void>;
 }
--- a/packages/adapter-utils/src/server-utils.ts
+++ b/packages/adapter-utils/src/server-utils.ts
@@ -1069,7 +1069,7 @@ export async function runChildProcess(
     env: Record<string, string>;
     timeoutSec: number;
     graceSec: number;
-    onLog: (stream: "stdout" | "stderr", chunk: string) => Promise<void>;
+    onLog: (stream: "stdout" | "stderr" | "system", chunk: string) => Promise<void>;
     onLogError?: (err: unknown, runId: string, message: string) => void;
     onSpawn?: (meta: { pid: number; processGroupId: number | null; startedAt: string }) => Promise<void>;
     stdin?: string;
--- a/packages/adapter-utils/src/types.ts
+++ b/packages/adapter-utils/src/types.ts
@@ -118,7 +118,7 @@ export interface AdapterExecutionContext {
   runtime: AdapterRuntime;
   config: Record<string, unknown>;
   context: Record<string, unknown>;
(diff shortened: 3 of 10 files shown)


xkonjin (commented):
Code review for PR 3733 (add 'system' stream to onLog interface):

Clean, mechanical change propagating the new 'system' stream type across the codebase.

**Positive note:** The change in `pi-local/src/server/execute.ts` from `if (stream === 'stderr')` to `if (stream !== 'stdout')` is the right call -- it means future stream types (beyond 'system') would also bypass the JSONL buffer without needing further changes. Good defensive coding.

**Worth checking:** In `execution-workspaces.ts` and `projects.ts`, system logs are now silently discarded (neither pushed to stdout nor stderr arrays). Is that intentional? If system messages contain useful diagnostic info (e.g., timeout warnings, kill signals), callers that only inspect stdout/stderr would miss them. Consider whether there's a `system` array or a combined `logs` field that should capture these.

**No test coverage changes observed** for the new 'system' path in the route handlers -- if it's possible to trigger system log events in existing integration tests, adding a case would help prevent silent regressions.

stubbi:
Closing this and recommending a coordinated follow-up.

The CI failure shows why this change isn't viable as a minimal workspace-only patch: `server/src/adapters/registry.ts` line 185 assigns `hermesExecute` (from the externally-published `hermes-paperclip-adapter`) to a slot typed against the workspace `ServerAdapterModule`. Once `AdapterExecutionContext.onLog` is widened in workspace `@paperclipai/adapter-utils`, the published `hermes-paperclip-adapter` package — which was compiled against an older narrow adapter-utils — no longer structurally matches the wider workspace type. TypeScript correctly rejects the assignment.

Making this land cleanly requires a coordinated version bump + republish of `@paperclipai/adapter-utils` together with the hermes adapter (and any other externally-published adapter) so every consumer resolves to the same widened type.

Happy to take another pass once that bump is in flight, or in a followup PR that does the coordination. In the meantime I'll keep this in the fork.

PR #3734, opened 2026-04-15
fix: proper cache headers for static assets and SPA fallback

stubbi:
## Thinking Path

> - Every deployment serves the same Vite-built UI bundle from the same express app
> - Vite emits JS/CSS under `/assets/<name>.<hash>.<ext>` — the hash rolls whenever the content rolls, so these files are inherently immutable
> - `index.html` references specific hashed filenames, so it has the opposite lifecycle: whenever we deploy, the file changes but the URL doesn't
> - Today the static middleware sends neither with cache headers, and the SPA fallback serves `index.html` for any unmatched route — including paths under `/assets/` that no longer exist after a deploy
> - That combination produces the familiar "blank screen after deploy" + `Failed to load module script: Expected a JavaScript MIME type but received 'text/html'` bug
> - This pull request caches hashed assets immutably, forces `index.html` to `no-cache` everywhere it gets served, and returns 404 for missing `/assets/*` paths

## What Changed

- `server/src/app.ts`:
  - Serve `/assets/*` with `Cache-Control: public, max-age=31536000, immutable`.
  - Serve the remaining static files (favicon, manifest, robots.txt) with a 1-hour cache, but override to `no-cache` specifically for `index.html` via the `setHeaders` hook — because `express.static` serves it directly for `/` and `/index.html`.
  - The SPA fallback (`app.get(/.*/, …)`) sets `Cache-Control: no-cache` on its `index.html` response.
  - The fallback returns 404 for paths under `/assets/` so browsers don't cache the HTML shell as a JavaScript module.

## Verification

- `curl -i http://localhost:3100/assets/index-abc123.js` → `cache-control: public, max-age=31536000, immutable`.
- `curl -i http://localhost:3100/` → `cache-control: no-cache`.
- `curl -i http://localhost:3100/assets/missing.js` → `404`.
- `curl -i http://localhost:3100/some/spa/route` → `200` HTML with `cache-control: no-cache`.

## Risks

Low. Asset URLs and HTML content are unchanged; only response headers and the 404 behavior for missing asset paths change. No API surface affected.

## Model Used

## Checklist

- [x] Model used specified
- [x] CI green

### DIFF
--- a/server/src/app.ts
+++ b/server/src/app.ts
@@ -287,9 +287,34 @@ export async function createApp(
     const uiDist = candidates.find((p) => fs.existsSync(path.join(p, "index.html")));
     if (uiDist) {
       const indexHtml = applyUiBranding(fs.readFileSync(path.join(uiDist, "index.html"), "utf-8"));
-      app.use(express.static(uiDist));
-      app.get(/.*/, (_req, res) => {
-        res.status(200).set("Content-Type", "text/html").end(indexHtml);
+      // Hashed asset files (Vite emits them under /assets/<name>.<hash>.<ext>)
+      // never change once built, so they can be cached aggressively.
+      app.use(
+        "/assets",
+        express.static(path.join(uiDist, "assets"), {
+          maxAge: "1y",
+          immutable: true,
+        }),
+      );
+      // Non-hashed static files (favicon.ico, manifest, robots.txt, etc.):
+      // short cache so operators who swap them out see the new version reasonably fast.
+      app.use(express.static(uiDist, { maxAge: "1h" }));
+      // SPA fallback. Only for non-asset routes — if the browser asks for
+      // /assets/something.js that doesn't exist, we must NOT serve the HTML
+      // shell: the browser would try to load it as a JavaScript module, fail
+      // with a MIME-type error, and cache that broken response. Return 404
+      // instead. The index.html response itself is no-cache so a subsequent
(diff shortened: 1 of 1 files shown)


xkonjin (commented):
Code review for PR 3734 (SPA caching improvements):

This is a well-thought-out caching strategy. The separation between hashed assets, non-hashed static files, and the HTML shell is correct.

**One concern worth noting:** The `/assets/` check in the SPA fallback uses `req.path.startsWith("/assets/")` which is correct for the Vite default output structure. If a project ever changes `build.assetsDir` in `vite.config.ts` to something other than `assets`, this hardcoded guard would silently break (assets would start serving HTML). Worth documenting this assumption, or ideally making the assets path configurable alongside the static serving setup.

**Minor:** The `1h` cache for non-hashed static files (favicon, etc.) is a reasonable middle ground. One thing to consider: some operators may serve over HTTP where the favicon doesn't need authentication, so caching that for 1h is fine. No security issue here.

Overall the MIME-type error prevention on missing assets and the `no-cache` on index.html are solid decisions.

stubbi:

stubbi:
Thanks @xkonjin! Good catch on `build.assetsDir` — you're right that the hardcoded `/assets/` check assumes Vite's default. Happy to either document it inline or make the path configurable if you'd prefer; let me know the direction and I'll follow up.

stubbi:
@greptileai review

PR #3735, opened 2026-04-15
feat(server): opt-in OpenTelemetry auto-instrumentation

stubbi:
## Thinking Path

> - Production self-hosters increasingly expect telemetry out of the box — Jaeger, Tempo, Honeycomb, Datadog, Grafana Cloud, Dynatrace all speak OTLP
> - Today there is no OpenTelemetry bootstrap in the server, so operators who want traces have to patch their fork or run a sidecar that captures only HTTP-level info
> - An opt-in bootstrap that costs nothing when disabled is the minimum-viable surface for this audience
> - The OpenTelemetry packages are heavyweight enough that we don't want them in the default dependency graph — they should load only when the operator configures an OTLP endpoint
> - This pull request adds a self-contained `server/src/instrumentation.ts` that dynamically imports the OTel SDK and starts it when `OTEL_EXPORTER_OTLP_ENDPOINT` is set, and is a complete no-op otherwise

No existing issue covers this directly — feature-gap description following the feature-request template:

**Problem or motivation**

Production self-hosters increasingly expect telemetry out of the box — Jaeger, Tempo, Honeycomb, Datadog, Grafana Cloud, Dynatrace all speak OTLP — but the server has no OpenTelemetry bootstrap. Operators who want traces today must patch their fork or run a sidecar that captures only HTTP-level information.

**Proposed solution**

An opt-in OTel bootstrap gated on `OTEL_EXPORTER_OTLP_ENDPOINT`, loaded via dynamic `import()` only when configured, so the heavyweight OTel packages stay out of the default dependency graph.

**Alternatives considered**

Related open PRs found during the duplicate-PR search approach observability differently: #4894 adds OTLP instrumentation to Paperclip core unconditionally, and #3752 proposes an observability plugin. Not duplicates — different layering: this PR keeps the default install dependency-free via opt-in dynamic import.

## What Changed

- New `server/src/instrumentation.ts` — opt-in OpenTelemetry auto-instrumentation. Gated on `OTEL_EXPORTER_OTLP_ENDPOINT`. Respects the standard OTel env vars (`OTEL_SERVICE_NAME`, `OTEL_SERVICE_VERSION`, `OTEL_EXPORTER_OTLP_ENDPOINT`). Skips the fs/dns/net auto-instrumentations (too chatty). `sdk.start()` is wrapped in try/catch so a bad endpoint or missing native bindings doesn't crash the server. `process.once("SIGTERM" / "SIGINT", …)` for clean shutdown on the first signal only. OTel packages are loaded via dynamic `import()` so they are true optional runtime dependencies — no entries in `package.json`, no lockfile churn.
- `server/src/index.ts` — import `./instrumentation.js` as the very first statement so auto-instrumentation can patch `http` / `express` / `pg` before they are evaluated by downstream modules.

## Verification

- `OTEL_EXPORTER_OTLP_ENDPOINT=http://localhost:4317 pnpm start` after `pnpm install @opentelemetry/{sdk-node,auto-instrumentations-node,exporter-trace-otlp-grpc,resources,semantic-conventions}` in `server/` — traces show up in the configured collector; HTTP, Express, and Postgres spans are populated.
- `OTEL_EXPORTER_OTLP_ENDPOINT` unset — server starts with no OTel-shaped output in logs, no behavior change.
- `OTEL_EXPORTER_OTLP_ENDPOINT=…` set but packages not installed — single `console.warn` at startup telling the operator which packages to install.

## Risks

Low. No behavior change unless the env var is set. The bootstrap never throws into the caller; every failure path ends in `console.warn` / `console.error` and falls through to non-traced operation.

## Model Used

## Checklist

- [x] Model used specified
- [x] CI green

### DIFF
--- a/README.md
+++ b/README.md
@@ -385,6 +385,10 @@ This is the short roadmap preview. See the full roadmap in [ROADMAP.md](ROADMAP.
 
 Find Plugins and more at [awesome-paperclip](https://github.com/gsxdsm/awesome-paperclip)
 
+## Observability
+
+Paperclip ships with opt-in OpenTelemetry auto-instrumentation for the server (traces only). It activates when `OTEL_EXPORTER_OTLP_ENDPOINT` is set and supports `grpc`, `http/protobuf`, and `http/json` via the standard `OTEL_EXPORTER_OTLP_PROTOCOL` env var. The `@opentelemetry/*` packages are optional peer dependencies — install them only if you want tracing. See [doc/observability.md](doc/observability.md) for install commands and the full env-var reference.
+
 ## Telemetry
 
 Paperclip collects anonymous usage telemetry to help us understand how the product is used and improve it. No personal information, issue content, prompts, file paths, or secrets are ever collected. Private repository references are hashed with a per-install salt before being sent.
--- a/README.md
+++ b/README.md
@@ -276,6 +276,10 @@ See [doc/DEVELOPING.md](doc/DEVELOPING.md) for the full development guide.
 
 Find Plugins and more at [awesome-paperclip](https://github.com/gsxdsm/awesome-paperclip)
 
+## Observability
+
+Paperclip ships with opt-in OpenTelemetry auto-instrumentation for the server (traces only). It activates when `OTEL_EXPORTER_OTLP_ENDPOINT` is set and supports `grpc`, `http/protobuf`, and `http/json` via the standard `OTEL_EXPORTER_OTLP_PROTOCOL` env var. The `@opentelemetry/*` packages are optional peer dependencies — install them only if you want tracing. See [doc/observability.md](doc/observability.md) for install commands and the full env-var reference.
+
 ## Telemetry
 
 Paperclip collects anonymous usage telemetry to help us understand how the product is used and improve it. No personal information, issue content, prompts, file paths, or secrets are ever collected. Private repository references are hashed with a per-install salt before being sent.
--- a/doc/observability.md
+++ b/doc/observability.md
@@ -46,6 +46,9 @@ Minimal setup:
 
 ```bash
 # Required — turns the feature on. Point at your collector.
+# For grpc this is the gRPC target (typically port 4317). For the HTTP
+# protocols give the collector's BASE URL (typically port 4318) — the
+# exporter appends /v1/traces itself.
 export OTEL_EXPORTER_OTLP_ENDPOINT="http://otel-collector:4317"
 
 # Optional — protocol. Defaults to grpc when unset.
--- a/doc/observability.md
+++ b/doc/observability.md
@@ -1,0 +1,73 @@
+# Observability
+
+Paperclip ships with **opt-in** OpenTelemetry auto-instrumentation for the
+server process. When activated it produces **traces only** — no metrics and no
+logs are exported by this integration. The OTel packages are *optional peer
+dependencies*: they are not in the default lockfile and are loaded dynamically
+only when an operator turns the feature on.
+
+When `OTEL_EXPORTER_OTLP_ENDPOINT` is unset, none of the `@opentelemetry/*`
+packages are imported and there is zero runtime overhead.
+
+## Enabling tracing
+
+### 1. Install the OTel peer dependencies
+
+Install the SDK, the auto-instrumentations bundle, the resources/semconv
+helpers, and **one** exporter matching your chosen OTLP protocol.
+
+Common to every protocol:
+
+```bash
+pnpm add \
+  @opentelemetry/sdk-node \
+  @opentelemetry/auto-instrumentations-node \
+  @opentelemetry/resources \
+  @opentelemetry/semantic-conventions
+```
+
+Then add the exporter for the protocol you intend to use:
+
+| `OTEL_EXPORTER_OTLP_PROTOCOL` | Exporter package                              |
+| ----------------------------- | --------------------------------------------- |
+| `grpc` (default if unset)     | `@opentelemetry/exporter-trace-otlp-grpc`     |
+| `http/protobuf`               | `@opentelemetry/exporter-trace-otlp-proto`    |
+| `http/json`                   | `@opentelemetry/exporter-trace-otlp-http`     |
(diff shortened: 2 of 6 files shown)


stubbi:

stubbi:
@greptileai review

stubbi:
Noting the P2 about the async-bootstrap early-request window: agreed on the description, but leaving it as-is. The alternative — `await import("./instrumentation.js")` — couples server startup to dynamic-import availability of the OTel packages. If the packages aren't installed or fail to load, the server wouldn't start at all, which is strictly worse than dropping the first ~50ms of spans. The current behavior matches how other opt-in OTel integrations ship.

serenakeyitan:

**Summary:** Adds opt-in OpenTelemetry auto-instrumentation for the server. Activated only when `OTEL_EXPORTER_OTLP_ENDPOINT` is set; otherwise no OTel packages are loaded. Imports are dynamic and packages are optional runtime dependencies.

**Tree context:**
- `infrastructure/NODE.md` — Operational concerns. OTel is a standard observability approach.
- `engineering/backend/NODE.md` — The instrumentation module is imported at the top of `server/src/index.ts` before any other modules, which is required for auto-instrumentation to patch HTTP/Express/PG.

**Observations:**
1. Zero-cost when disabled — packages are loaded dynamically only when the env var is set.
2. Correctly placed as the first import in `server/src/index.ts` to ensure patching happens before module initialization.
3. Minimal footprint (2 files, 99 lines). Self-hosters opt in by installing OTel packages and setting one env var.

**Verdict:** Clean opt-in instrumentation that follows best practices for OTel integration. No tree impact for default deployments.

stubbi:
@greptileai re-review please — applied three small polish items from a quality audit in `ccefa570`:

**Fix 1 — `@ts-expect-error` → `@ts-ignore`** (real compile bug). All five OTel dynamic-import suppressions in `server/src/instrumentation.ts` now use `@ts-ignore` so the directive is tolerant of both states (peer deps installed or not). Previously, installing the peer deps would make the `@ts-expect-error` directives themselves trigger compile errors as unused.

**Fix 2 — Honor `OTEL_EXPORTER_OTLP_PROTOCOL`** (matches OTLP spec). New `resolveProtocol()` helper reads `process.env.OTEL_EXPORTER_OTLP_PROTOCOL?.trim().toLowerCase()`:
- unset / `""` / `grpc` → `@opentelemetry/exporter-trace-otlp-grpc` (default, backward compat with previous PR head)
- `http/protobuf` → `@opentelemetry/exporter-trace-otlp-proto`
- `http/json` → `@opentelemetry/exporter-trace-otlp-http`
- anything else → one-shot `console.warn` listing valid options, falls through to gRPC

The exporter import is extracted to `importExporter(protocol)` reusing the same dynamic-import + try/catch error-handling style as the rest of the bootstrap. The peer-dep-missing warning now names the specific exporter package matching the operator's chosen protocol.

**Fix 3 — Doc**. Added `doc/observability.md` with the required-packages list per protocol, minimal env setup, and a note that this is traces-only (no metrics/logs). Short pointer added in `README.md` above the existing "Telemetry" section.

Still zero overhead when `OTEL_EXPORTER_OTLP_ENDPOINT` is unset; no new required env vars; backward compatible with operators who set only the endpoint.

stubbi:
@greptileai re-review please — branch rebuilt on current master (the old base carried fork-divergent history) and both findings addressed:

**1. Async-bootstrap timing.** Agreed the comment overpromised. Rather than pretending import order patches static imports, the guarantee is now explicit: `instrumentation.ts` exports `instrumentationReady`, and `startServer()` awaits it before opening the first DB connection or constructing the HTTP server. A future refactor that moves `createServer()` earlier now waits for the SDK instead of silently losing coverage. Comments in both files rewritten to describe the real guarantee.

**2. Un-awaited `sdk.shutdown()`.** `shutdownInstrumentation()` (idempotent, memoized) is now awaited inside index.ts's own SIGTERM/SIGINT handler *before* its `process.exit(0)`, so the exporter's final batch flushes before the process goes away. The module's own signal handlers remain as a backstop for uncoordinated entrypoints and share the same single flush. Chose this over `sdk.shutdown().then(() => process.exit(0))` inside the OTel handler because index.ts owns process exit (telemetry flush + embedded-postgres stop happen there too) — two competing exit paths would race.

Root typecheck clean.

stubbi:
@greptileai re-review please — both open items addressed in `9f2f4ce08`:

**HTTP exporter URL (r-latest finding)** — exactly right, thanks. `url` is now passed only for gRPC; the `http/protobuf` / `http/json` exporters read `OTEL_EXPORTER_OTLP_ENDPOINT` themselves so the SDK appends `/v1/traces`. `doc/observability.md` now spells out the base-URL semantics (4317 gRPC target vs 4318 HTTP base URL).

**Test gate** — new `server/src/__tests__/instrumentation.test.ts` (11 tests): protocol mapping for all valid values incl. case-insensitivity, the unknown-protocol→grpc fallback warning, `instrumentationReady` resolving immediately when the feature is off and settling with a diagnostic (not a crash) when the endpoint is set but the optional `@opentelemetry/*` packages are missing — which is precisely the state CI runs in — and `shutdownInstrumentation` idempotence/no-op behavior.

stubbi:
@greptileai re-review please — all three r-latest items addressed:

1. **Unbounded shutdown** — `sdkShutdown` now races `sdk.shutdown()` against an unref'd 5s timer (the SDK's own flush budget), so an unreachable collector can't hang process exit; the timeout rejection is absorbed and logged by `shutdownInstrumentation`'s existing catch.
2. **Exporter constructor type** — widened to `new (config?: Record<string, unknown>) => unknown` so credential/header configs aren't typed away at the import boundary.
3. **Warn-spy ordering** — spy installed before `importFreshInstrumentation()` with a comment pinning the timing contract.

`instrumentation.test.ts` 11/11 locally; typecheck clean. The failing `Verify serialized server suites (1/4)` lane was the known-flaky `heartbeat-process-recovery` suite (#4328) — this push retriggers it.

PR #3736, opened 2026-04-15
feat: optional Redis pub/sub for cross-replica live events

stubbi:
## Thinking Path

> - The web UI is kept live via a WebSocket bus that streams `LiveEvent`s — run progress, status changes, transcript deltas
> - Today that bus is an in-process `EventEmitter`, which is fine for single-replica deployments
> - Any self-hoster running more than one server replica (a typical k8s Deployment with `replicas: 2`) sees events silently drop for clients whose WebSocket landed on a different pod than the one executing the run
> - The current workaround is pod affinity / sticky sessions, which defeats half the reason for multi-replica in the first place
> - This pull request adds an optional Redis pub/sub transport that mirrors local events to every replica, gated on env vars and using a dynamically-imported `ioredis` so single-process deployments pay no dependency cost

## What Changed

- `server/src/services/live-events.ts`:
  - Add a Redis pub/sub transport for cross-replica delivery. Activates when `PAPERCLIP_LIVE_EVENTS_REDIS_URL` (or the generic `PAPERCLIP_REDIS_URL` fallback) is set.
  - Each process tags outgoing envelopes with `pid + randomUUID()` and drops messages whose origin matches, so events are never emitted twice on the publisher.
  - Incoming cross-replica messages are routed exactly like the in-process `publishLiveEvent` / `publishGlobalLiveEvent` functions — global events go to `"*"` listeners only, company-scoped events go only to that company's listeners.
  - `ioredis` is imported via dynamic `import()` guarded by `@ts-expect-error` so the module type-checks when the package is not in `node_modules`. Operators install `ioredis` themselves when they opt in.
  - Failures in the outer init path are surfaced via a single `console.warn` so a misconfigured Redis URL is visible rather than silently degrading to single-replica delivery.

## Verification

- Two-replica local setup sharing a Redis instance: issue a mutation against pod A, observe the WebSocket event on a client pinned to pod B.
- Single-replica (no env var): live events behave exactly as before — in-process only.
- Env var set but Redis unreachable or `ioredis` not installed: single warning at startup, no crash, no cross-replica delivery.

## Risks

Low. The module's public surface (`publishLiveEvent`, `publishGlobalLiveEvent`, `subscribeCompanyLiveEvents`, `subscribeGlobalLiveEvents`) is unchanged. No default dependency added. Cross-replica delivery is best-effort — any failure path falls back to single-replica in-process behavior.

## Model Used

## Checklist

- [x] Model used specified
- [x] CI green

### DIFF
--- a/server/src/services/live-events.ts
+++ b/server/src/services/live-events.ts
@@ -74,9 +74,12 @@ async function initRedis(): Promise<void> {
   const sub = new (Redis as any)(url) as SubscribeClient;
   attachErrorLogger(pub, "publisher");
   attachErrorLogger(sub, "subscriber");
-  redisPub = pub;
-  redisSub = sub;
-  await sub.subscribe(CHANNEL);
+  // Install the message listener and complete the subscription BEFORE
+  // publishing anything. If we assigned redisPub first, the publisher
+  // could race the subscribe(): events published in that narrow window
+  // land in Redis on a channel this replica isn't yet listening to,
+  // and if subscribe() then throws, redisPub stays non-null until the
+  // outer .catch() runs — more events go to Redis with no local sub.
   sub.on("message", (_channel: unknown, message: unknown) => {
     try {
       const envelope = JSON.parse(message as string) as {
@@ -97,6 +100,12 @@ async function initRedis(): Promise<void> {
       // delivery for everyone else.
     }
   });
+  await sub.subscribe(CHANNEL);
+  // Only expose the publisher once we're actually receiving from the
+  // channel. If subscribe() threw above we would never reach this line
+  // and redisPub would stay null — the outer .catch() handles the log.
+  redisPub = pub;
+  redisSub = sub;
 }
 
 // Non-blocking best-effort init on module load. If ioredis isn't
--- a/server/src/services/live-events.ts
+++ b/server/src/services/live-events.ts
@@ -24,7 +24,8 @@ let nextEventId = 0;
 //
 // ioredis is imported dynamically so self-hosters without the env var
 // set don't pay the dependency cost — and the type is intentionally
-// kept loose so ioredis can be an optionalDependency.
+// kept loose so operators who want this feature install `ioredis`
+// themselves without forcing a global dependency update.
 type PublishClient = { publish(channel: string, message: string): Promise<unknown> };
 type SubscribeClient = {
   subscribe(channel: string): Promise<unknown>;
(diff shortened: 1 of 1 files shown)


stubbi:

stubbi:
@greptileai review

stubbi:
@greptileai review

serenakeyitan:
🌱 **gardener** · 🟡 `NEW_TERRITORY` · medium

**Summary:** Adds optional Redis pub/sub for cross-replica live event distribution. When `PAPERCLIP_LIVE_EVENTS_REDIS_URL` is set, events are published to Redis so WebSocket clients on other replicas see them. Falls back to in-memory EventEmitter (unchanged behavior) when unset.

**Tree context:**
- `engineering/backend/NODE.md` — WebSocket-based live events in `server/src/realtime/live-events-ws.ts`. This PR modifies the underlying event distribution layer.
- `infrastructure/deployment/NODE.md` — Docker deployment. Multi-replica support is an infrastructure concern.

**Observations:**
1. **New infrastructure dependency (Redis).** The tree doesn't mention Redis anywhere. This introduces an optional but significant new dependency for multi-replica deployments.
2. **Single-file change (128 lines)** — contained entirely in `server/src/services/live-events.ts`. Clean implementation with origin-tagging to prevent echo loops.
3. **Graceful degradation** — no behavioral change when Redis is not configured. Good for backward compatibility.
4. **No documentation or deployment guide updates.** Self-hosters running multi-replica k8s deployments would need to know about this env var.

**Verdict:** NEW_TERRITORY — Redis as an optional infrastructure dependency is a new architectural decision that should be captured on the tree. The implementation is clean and backward-compatible.

stubbi:
Superseding with 5875.

A quality re-audit flagged two concerns:

1. **New infra dependency.** Paperclip already runs Postgres for everything; pinning self-hosters to Redis just to get `replicas: 2` cross-replica live-events isn't a great trade.
2. **Cross-tenant metadata leak.** A single global channel filtered by `companyId` client-side means anyone with `redis-cli SUBSCRIBE` access sees all tenants' event metadata. Now that #5864 / #5865 have landed (per-company JWT + plugin tenant FKs), per-company channels are the right fit.

The replacement uses Postgres `LISTEN/NOTIFY` keyed per-company by default and keeps Redis pub/sub as an optional second transport via `PAPERCLIP_LIVE_EVENTS_TRANSPORT=redis` for operators who already run Redis. `originId` echo suppression and the env-var surface are preserved from this PR. UTF-8 byte-length check (not JS string length) added. Single-replica local-first behavior is unchanged.

Closing in favor of the wider replacement.

PR #3739, opened 2026-04-15
fix(issues): block agent comments on completed issues from unrelated runs

octasoft-ltd:
## Thinking Path

> - Agents post comments on issues via the API during heartbeat runs
> - The assertAgentRunCheckoutOwnership guard validated checkout ownership only for in_progress issues
> - Agents could freely comment on done/cancelled issues even when their active run targeted a different issue
> - This caused cross-contamination: an agent working on issue X accidentally posted content to completed issue Y
> - This pull request adds a server-side guard that verifies the run context matches the target issue for completed issues
> - The benefit is preventing stray cross-issue comments that confuse issue threads

## What Changed

- Added a guard in assertAgentRunCheckoutOwnership (server/src/routes/issues.ts) that rejects agent comments on done/cancelled issues when the active run targets a different issue (409 response with diagnostic IDs)
- The guard also rejects when the run record cannot be found (conservative approach for stale run IDs)
- Added regression test file server/src/__tests__/issue-cross-issue-comment-guard.test.ts with 7 test cases

## Verification

- pnpm --filter server typecheck passes
- Test file follows existing route mock test patterns
- Manual: agent attempting to comment on wrong completed issue receives 409

## Risks

- Agents that legitimately need to comment on completed issues from a different run context will be blocked. This is intentional. If edge cases surface, exceptions can be added.
- The getRun call adds one DB query per agent comment on closed issues. Acceptable since comments on closed issues are infrequent.

## Model Used

- Provider: Anthropic Claude
- Model ID: claude-opus-4-6 (1M context)
- Capabilities: Tool use, code execution, extended context

## Checklist


octasoft-ltd:
@greptileai please re-review — addressed both P2 findings (null-run bypass now rejects conservatively, added missing addComment assertion and null-run test case). PR description updated with required template sections.

serenakeyitan:

**Summary:** Blocks agents from commenting on completed (done/cancelled) issues when their active run targets a different issue. Prevents cross-contamination where an agent working on issue X accidentally posts to completed issue Y. Includes comprehensive test coverage.

**Tree context:**
- `product/governance/NODE.md` — Governance-first: audit trails and guardrails are core. Preventing stray agent comments on wrong issues strengthens the governance model.
- `engineering/backend/NODE.md` — Route/service split. The guard is implemented in the route layer where request context (actor, run) is available.

**Observations:**
1. Well-scoped fix: 258-line test file covering the exact scenario described in the issue.
2. Guard checks run ownership against the target issue, returning 403 when the agent's active run doesn't match.
3. Only applies to closed issues — agents can still comment on open issues they're not assigned to, which is the correct behavior for collaborative workflows.

**Verdict:** Clean governance improvement that aligns with the governance-first principle. Good test coverage.

octasoft-ltd:

octasoft-ltd:
@greptileai please re-review.

This pushes a follow-up to the rebase: the conservative null-run rejection from b39e4b81 (the Greptile-driven hardening) broke 8 upstream tests in `issue-comment-reopen-routes.test.ts`. Those tests mock `heartbeat.getRun` to return `null` by default while still setting an agent runId on `req.actor`, so the guard rejected with 409 before the reopen-route logic could run.

In production, `req.actor.runId` only exists when Paperclip's auth layer issued a JWT for a real run, so `getRun` returning `null` for a set runId is anomalous rather than malicious. The existing checkout / assignee / reopen guards still cover authorization in that case, so I've narrowed this guard to only enforce cross-issue rejection when we can confirm the run's target `issueId`. The corresponding self-test now asserts the fall-through behaviour.

Code change: 7 lines net, scoped to `assertAgentIssueMutationAllowed` in `server/src/routes/issues.ts` and one renamed test in `server/src/__tests__/issue-cross-issue-comment-guard.test.ts`.

octasoft-ltd:
@greptileai please re-review.

Tightened the cross-issue guard per your P1 finding: when the run record is found but `contextSnapshot.issueId` is missing/non-string/null, we now reject with the same 409 instead of silently passing. Closed the bypass.

Also strengthened the "falls through" test from `not.toBe(409)` → `toBe(200)` and added two new tests covering the new rejection paths (contextSnapshot null, contextSnapshot non-object).

PR #3783, opened 2026-04-16
fix(server): respect configured host when probing ports

### LINKED ISSUE #3747: Server auto-bumps port when `detect-port` probe fails on wildcard, even though `server.host` is loopback
## Summary

`@paperclipai/server` calls `detectPort(config.port)` without passing the `hostname` option, so `detect-port` probes the port on five different addresses (unspecified/`::`, `0.0.0.0`, `127.0.0.1`, `localhost`, and the machine's LAN IP). If **any one** of those probes fails, the port is incremented by 1.

This causes Paperclip to silently bind a different port than configured whenever another process holds a specific-interface listener on the configured port — even when that other process is on an interface Paperclip has no intention of using (`config.host = "127.0.0.1"`).

## Steps to reproduce

Any environment where something else is already bound to `<configured-port>` on an interface other than loopback. On macOS with Tailscale Serve it's trivial:

```bash
tailscale serve --bg --http=3100 http://127.0.0.1:9999   # binds *:3100 on tailnet interface (invisible to non-sudo lsof)

# Fresh Paperclip, default config, server.port = 3100, server.host = "127.0.0.1"
npx paperclipai run --data-dir ~/paperclip
```

**Observed:** `paperclipai doctor` reports "Port 3100 is available" (its own probe is correct — uses `127.0.0.1`). Server starts and logs `Server listening on 127.0.0.1:3101`.

**Expected:** Server listens on the configured port `3100` — `127.0.0.1:3100` is actually free and would bind successfully.

## Root cause

`@paperclipai/server/dist/index.js:384`:

```ts
const listenPort = await detectPort(config.port);
```

[`detect-port@2.1.0`'s `tryListen`](https://github.com/node-modules/detect-port/blob/master/src/detect-port.ts) walks five addresses when no `hostname` is supplied and bumps on any failure. Node rejects new wildcard/unspecified binds when a specific-interface listener already holds the same port — so the probe fails even though the specific address Paperclip will actually use is free.

Reproduction that isolates the mechanism (with nothing but the tailscale listener holding `100.x.x.x:3100`):

```js
// unspecified        → EADDRINUSE: address already in use :::3100
// 0.0.0.0            → EADDRINUSE: address already in use 0.0.0.0:3100
// 127.0.0.1          → ok
// localhost          → ok
// 192.168.x.x (LAN)  → ok
```

`detect-port` returns `3101` and Paperclip happily binds it, leaving anything downstream (reverse proxies, port forwarders, `tailscale serve` targets, docs pointing at `:3100`) broken.

## Proposed fix

Pass the configured hostname so `detect-port` probes only the address Paperclip will actually bind:

```ts
const listenPort = await detectPort({ port: config.port, hostname: config.host });
```

With this change:
- If `config.host` is `127.0.0.1`, only that address is probed — wildcard/other-interface conflicts are correctly ignored.
- If `config.host` is `0.0.0.0` or `::`, behavior is unchanged (probe matches actual bind).
- For authenticated/public deployments where operators want broader collision detection, the configured host still drives it.

The same pattern applies at line 262 (the fallback for embedded Postgres port detection) if the embedded-postgres port is also loopback-only.

## Impact

- Any user running a wildcard-interface listener on the configured port (Tailscale Serve, Docker port publishers, reverse proxies, some VPN clients) silently gets bumped to `port + 1` at startup. The auto-bump is invisible in `doctor` (which uses a different probe) and easy to miss because the service still starts.
- Related worktree-port bugs (#1849, #3613) exist in a different code path (Paperclip's own `findAvailablePort` in `server/src/worktree-config.ts`) — this one is in `@paperclipai/server` core startup.

## Environment

- Paperclip 2026.403.0
- macOS 14.4 / Apple Silicon
- Node 25.9.0
- `detect-port@2.1.0`
- Deployment mode: `local_trusted`, `server.host: 127.0.0.1`, `server.port: 3100`

LeonSGP43:
## Thinking Path

> - Server startup is part of the local-first control plane because operators depend on predictable host/port binding for the API, UI, and local integrations
> - The current startup path asks `detect-port` to probe `config.port` without the configured hostname, so the library checks wildcard and non-loopback addresses in addition to the address Paperclip will actually bind
> - That makes Paperclip silently bump to the next port when an unrelated interface listener conflicts on `0.0.0.0` or another address, even if the configured bind host is still free
> - This pull request routes startup port probing through a small helper that passes the configured hostname, and applies the same loopback-only probe to the embedded Postgres fallback path
> - The benefit is that Paperclip now probes the same address it intends to bind, so loopback deployments keep their configured port instead of unexpectedly shifting to `port + 1`

## What Changed

- Added a small `detectAvailablePort` helper that passes `{ port, hostname }` to `detect-port` when a bind host is known
- Updated server startup to probe `config.port` against `config.host` instead of using the library's multi-address fallback
- Updated the embedded Postgres fallback probe to use `127.0.0.1`, matching the actual local-only bind path there as well
- Added a focused regression test covering both host-aware probing and the blank-host fallback behavior
- Fixes #3747

## Verification

- `pnpm --filter @paperclipai/server exec tsc --noEmit`
- `pnpm exec vitest run src/__tests__/port-detection.test.ts` (run from `server/`)

## Risks

- Low risk. This only changes how startup probes port availability before binding.
- The main behavioral change is intentional: loopback-host deployments stop treating wildcard or unrelated-interface listeners as conflicts.
- When `config.host` is blank, the helper keeps the previous `detect-port(port)` behavior.

## Model Used

- OpenAI Codex, GPT-5-based coding agent with terminal tool use and local code execution. Exact internal deployment ID and exposed context-window size were not surfaced in this session.

## Checklist

### DIFF
--- a/server/src/__tests__/port-detection.test.ts
+++ b/server/src/__tests__/port-detection.test.ts
@@ -1,0 +1,36 @@
+import { beforeEach, describe, expect, it, vi } from "vitest";
+
+const { detectPortMock } = vi.hoisted(() => ({
+  detectPortMock: vi.fn(),
+}));
+
+vi.mock("detect-port", () => ({
+  default: detectPortMock,
+}));
+
+import { detectAvailablePort } from "../port-detection.js";
+
+describe("detectAvailablePort", () => {
+  beforeEach(() => {
+    detectPortMock.mockReset();
+  });
+
+  it("probes only the configured host when one is provided", async () => {
+    detectPortMock.mockResolvedValue(3100);
+
+    await expect(detectAvailablePort(3100, "127.0.0.1")).resolves.toBe(3100);
+
+    expect(detectPortMock).toHaveBeenCalledWith({
+      port: 3100,
+      hostname: "127.0.0.1",
+    });
+  });
+
+  it("falls back to default detect-port probing when host is blank", async () => {
+    detectPortMock.mockResolvedValue(3101);
+
+    await expect(detectAvailablePort(3100, "  ")).resolves.toBe(3101);
+
+    expect(detectPortMock).toHaveBeenCalledWith(3100);
+  });
+});
--- a/server/src/__tests__/port-detection.test.ts
+++ b/server/src/__tests__/port-detection.test.ts
@@ -33,4 +33,12 @@ describe("detectAvailablePort", () => {
 
     expect(detectPortMock).toHaveBeenCalledWith(3100);
   });
+
+  it("falls back to default detect-port probing when host is undefined", async () => {
+    detectPortMock.mockResolvedValue(3102);
+
+    await expect(detectAvailablePort(3100)).resolves.toBe(3102);
+
+    expect(detectPortMock).toHaveBeenCalledWith(3100);
+  });
 });
--- a/server/src/__tests__/server-startup-feedback-export.test.ts
+++ b/server/src/__tests__/server-startup-feedback-export.test.ts
@@ -10,7 +10,10 @@ const {
 } = vi.hoisted(() => {
   const createAppMock = vi.fn(async () => ((_: unknown, __: unknown) => {}) as never);
   const createDbMock = vi.fn(() => ({}) as never);
-  const detectPortMock = vi.fn(async (port: number) => port);
+  const detectPortMock = vi.fn(
+    async (input: number | { port: number; hostname: string }) =>
+      typeof input === "number" ? input : input.port,
+  );
   const feedbackExportServiceMock = {
     flushPendingFeedbackTraces: vi.fn(async () => ({ attempted: 0, sent: 0, failed: 0 })),
   };
--- a/server/src/index.ts
+++ b/server/src/index.ts
@@ -23,10 +23,10 @@ import {
   companyMemberships,
   instanceUserRoles,
 } from "@paperclipai/db";
-import detectPort from "detect-port";
 import { createApp } from "./app.js";
 import { loadConfig } from "./config.js";
 import { logger } from "./middleware/logger.js";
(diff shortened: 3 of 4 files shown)


serenakeyitan:
🌱 **gardener** · ✅ `ALIGNED` · severity: `low` · commit: `6a41722a`

### Context fit

| Server startup / port binding | Probes `config.port` against `config.host` instead of library multi-address fallback; adds loopback-only probe to embedded Postgres path | Server treats operator-configured bind host as authoritative; config is loaded from env/yaml and respected through startup (engineering/backend/NODE.md: Configuration) | ✅ Aligned |
| Bug scope | Focused fix closing issue #3747; follow-up commit only adds a third regression test case (undefined host) and adjusts a pre-existing mock to match the new input shape | Routes/services split and testable startup are core conventions (engineering/backend/NODE.md: Route/Service Split) | ✅ Aligned |

<details>

- [`engineering/backend/NODE.md`](https://github.com/serenakeyitan/paperclip-tree/blob/main/engineering/backend/NODE.md) — Server config is loaded from env + YAML; bind host and deployment mode are operator-authoritative.
- [`engineering/backend/dev-runner/NODE.md`](https://github.com/serenakeyitan/paperclip-tree/blob/main/engineering/backend/dev-runner/NODE.md) — Local dev runner context (port/host stability matters for worktree integrations).

</details>

No concerns. The follow-up commit only strengthens the regression test coverage — verdict unchanged from the prior review.

---

<sub>Reviewed commit: <code>6a41722a</code> · Tree snapshot: <code>7aad3e36</code> · Commands: <code>@gardener re-review</code> · <code>@gardener pause</code> · <code>@gardener ignore</code></sub>

PR #3844, opened 2026-04-16
[codex] Fix Docker gh installation

cryppadotta:
## Thinking Path

> - Paperclip is the control plane for autonomous AI companies, and the Docker image is the no-local-Node path for running that control plane.
> - The deploy workflow builds and pushes that image from the repository `Dockerfile`.
> - The current image setup adds GitHub CLI through GitHub's external apt repository and verifies a mutable keyring URL with a pinned SHA256.
> - GitHub rotated the CLI Linux package signing key, so that pinned keyring checksum now fails before Buildx can publish the image.
> - Paperclip already has a repo-local precedent in `docker/untrusted-review/Dockerfile`: install Debian trixie's packaged `gh` directly from the base distribution.
> - This pull request removes the external GitHub CLI apt keyring/repository path from the production image and installs `gh` with the rest of the Debian packages.
> - The benefit is a simpler Docker build that no longer fails when GitHub rotates the apt keyring file.

## What Changed

- Updated the main `Dockerfile` base stage to install `gh` from Debian trixie's package repositories.
- Removed the mutable GitHub CLI apt keyring download, pinned checksum verification, extra apt source, second `apt-get update`, and separate `gh` install step.

## Verification

- `git diff --check`
- `./scripts/docker-build-test.sh` skipped because Docker is installed but the daemon is not running on this machine.
- Confirmed `https://packages.debian.org/trixie/gh` returns HTTP 200, matching the base image distribution package source.

## Risks

- Debian's `gh` package can lag the latest upstream GitHub CLI release. This is acceptable for the current image contract, which requires `gh` availability but does not document a latest-upstream version guarantee.
- A full image build still needs to run in CI because the local Docker daemon is unavailable in this environment.

## Model Used

- OpenAI Codex, GPT-5-based coding agent. Exact backend model ID was not exposed in this runtime; tool use and shell execution were enabled.

## Checklist

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -2,15 +2,7 @@ FROM node:lts-trixie-slim AS base
 ARG USER_UID=1000
 ARG USER_GID=1000
 RUN apt-get update \
-  && apt-get install -y --no-install-recommends ca-certificates gosu curl git wget ripgrep python3 \
-  && mkdir -p -m 755 /etc/apt/keyrings \
-  && wget -nv -O/etc/apt/keyrings/githubcli-archive-keyring.gpg https://cli.github.com/packages/githubcli-archive-keyring.gpg \
-  && echo "20e0125d6f6e077a9ad46f03371bc26d90b04939fb95170f5a1905099cc6bcc0  /etc/apt/keyrings/githubcli-archive-keyring.gpg" | sha256sum -c - \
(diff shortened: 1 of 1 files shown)


cryppadotta:
@greptile - is this a good or bad strategy to remove this hash verification from the Docker image? I'm worried my users will be upset about losing pinning. But it is annoying to have something change upstream that breaks our build

PR #3932, opened 2026-04-18
fix(heartbeat): keep adapter_failed continuation blockers on Review Manager (WEB-1550)

oshabana:
WEB-1550\n\nScope:\n- Keep parent review tickets on Review Manager when a reviewer continuation fails with adapter_failed.\n- Reroute blocked orphaned issues back to the most appropriate agent owner instead of dropping them from blocked ownership.\n- Preserve the blocked Review Manager exception so review gates remain visible.\n- Add regression coverage for blocked reroute behavior and the continuation blocker path.\n- Fix embedded Postgres temp dir ownership for root-run test environments.\n\nTests:\n- pnpm test -- --run server/src/__tests__/heartbeat-process-recovery.test.ts\n- pnpm --filter @paperclipai/server typecheck\n\nRisks:\n- The blocked reroute sweep now reassigns blocked issues to a fallback agent when no live execution path remains, so the ownership heuristic should stay conservative.\n- Existing Review Manager-owned review blockers are intentionally exempted from the generic blocked sweep.

### DIFF
--- a/packages/db/src/test-embedded-postgres.ts
+++ b/packages/db/src/test-embedded-postgres.ts
@@ -2,6 +2,7 @@ import fs from "node:fs";
 import net from "node:net";
 import os from "node:os";
 import path from "node:path";
+import { execFileSync } from "node:child_process";
 import { applyPendingMigrations, ensurePostgresDatabase } from "./client.js";
 
 type EmbeddedPostgresInstance = {
@@ -32,6 +33,17 @@ export type EmbeddedPostgresTestDatabase = {
 };
 
 let embeddedPostgresSupportPromise: Promise<EmbeddedPostgresTestSupport> | null = null;
+
+function ensureDirectoryOwnedByPostgres(dataDir: string) {
+  if (process.platform === "win32") return;
+  if (typeof process.getuid === "function" && process.getuid() !== 0) return;
+
+  try {
+    execFileSync("chown", ["-R", "postgres:postgres", dataDir]);
+  } catch {
+    // If the host cannot chown the temp directory, fall back to the existing behavior.
+  }
+}
 
 async function getEmbeddedPostgresCtor(): Promise<EmbeddedPostgresCtor> {
   const mod = await import("embedded-postgres");
@@ -66,6 +78,7 @@ function formatEmbeddedPostgresError(error: unknown): string {
 
 async function probeEmbeddedPostgresSupport(): Promise<EmbeddedPostgresTestSupport> {
   const dataDir = fs.mkdtempSync(path.join(os.tmpdir(), "paperclip-embedded-postgres-probe-"));
+  ensureDirectoryOwnedByPostgres(dataDir);
   const port = await getAvailablePort();
   const EmbeddedPostgres = await getEmbeddedPostgresCtor();
   const instance = new EmbeddedPostgres({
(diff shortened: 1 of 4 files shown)


oshabana:
## Review Summary — PR #3932 / 34467a8e7613d10895d07dfa62abdb43b1de04e2

**Verdict**: REQUEST CHANGES (2 blocking issues)
**Cycle**: 2 of 3

### Blocking Issues

1. [CORRECTNESS] File: `server/src/services/heartbeat.ts:1417` — `isReviewManagerBlockerIssue()` only preserves Review Manager ownership for titles starting with the exact prefix `[REVIEW]`.
   Why: the live reviewer child tickets in this workflow are titled `[REVIEW-GPT] ...` and `[REVIEW-OPUS] ...`, so they do not satisfy `startsWith("[review]")`. Those blocked review tickets can still fall through `rerouteBlockedIssueOwnership()` and be reassigned away from Review Manager, which is the regression this ticket is meant to stop.
   Fix: match the actual review-ticket prefix family used in production, not only the bare `[REVIEW]` prefix, and cover that real title shape in regression coverage.
   Found by: GPT

2. [CORRECTNESS] File: `server/src/__tests__/heartbeat-process-recovery.test.ts:754` — the new Review Manager regression test does not exercise the guard it claims to validate.
   Why: the fixture inputs passed by the test are ignored when the rows are seeded. The agent insert still hardcodes `name: "CodexCoder"` at `server/src/__tests__/heartbeat-process-recovery.test.ts:336`, the issue insert still hardcodes `title: "Recover stranded assigned work"` at `server/src/__tests__/heartbeat-process-recovery.test.ts:404`, and the run insert still hardcodes `errorCode: "process_lost"` / `error: "run failed before issue advanced"` at `server/src/__tests__/heartbeat-process-recovery.test.ts:379-380`. On top of that, the test keeps `blockedAssigneeMode: "agent"`, so the issue never enters the orphaned blocked-issue sweep that this change also adds. The test passes without proving the `[REVIEW-*]` Review Manager path or the `adapter_failed` scenario.
   Fix: wire the fixture inputs into the seeded agent, issue, and run rows, then update the regression so it actually hits the intended blocked-review preservation path.
   Found by: Opus

### Non-blocking Observations

- No additional non-blocking observations worth carrying forward after deduping reviewer noise.

### Convention Compliance

- Kysely only: PASS
- Vitest only: PASS
- No emojis: PASS
- Mobile-first (if frontend): PASS
- No hardcoded domains: PASS

oshabana:
## Review Summary — PR #3932 / 72d06d12f50a8f84d70782110c1ea3d41c59eef2

**Verdict**: REQUEST CHANGES (1 blocking issue)
**Cycle**: 3 of 3

### Blocking Issues

1. [CORRECTNESS] File: server/src/__tests__/heartbeat-process-recovery.test.ts:754 — The new Review Manager regression still does not execute the preservation path it claims to validate.
   Why: `reconcileStrandedAssignedIssues()` only sweeps blocked issues with `assigneeAgentId IS NULL` at `server/src/services/heartbeat.ts:3296`, but this test seeds `blockedAssigneeMode: "agent"`, so the fixture keeps `assigneeAgentId = agentId` and never reaches `rerouteBlockedIssueOwnership()` / `isReviewManagerBlockerIssue()`. The test passes without proving the `adapter_failed` / `[REVIEW-*]` Review Manager preservation rule required by WEB-1550.
   Fix: Route the regression through the run-completion path that actually evaluates agent-owned blocked issues (`server/src/services/heartbeat.ts:4489`) or change the sweep/test setup so the Review Manager exemption is truly exercised.
   Found by: GPT

### Non-blocking Observations

- Dropped Opus's whitespace note as non-blocking noise.

### Convention Compliance

- Kysely only: N/A (Paperclip infra)
- Vitest only: PASS
- No emojis: PASS
- Mobile-first (if frontend): N/A
- No hardcoded domains: PASS

GitHub would not accept a formal `REQUEST_CHANGES` review from this authenticated identity because it is the PR author, so this is posted as a comment instead.

PR #3955, opened 2026-04-18
fix(server): allow paused wakeups (WEB-1389)

oshabana:
WEB-1389\n\nScope:\n- Stop treating paused agents as an immediate wakeup conflict in server/src/services/heartbeat.ts\n- Add a regression test that paused agents can queue wakeups without throwing\n- Add a pure unit test for the wakeup status guard\n\nTests:\n- pnpm test -- server/src/__tests__/heartbeat-wakeup-status-guard.test.ts\n- pnpm test -- server/src/__tests__/heartbeat-process-recovery.test.ts -t "queues wakeups for paused agents instead of rejecting them" (embedded Postgres suite skipped on this host because init script failed)\n\nRisk:\n- Paused agents can now accumulate queued wakeups, but claim/start still blocks execution until the agent resumes

### DIFF
--- a/server/src/__tests__/heartbeat-process-recovery.test.ts
+++ b/server/src/__tests__/heartbeat-process-recovery.test.ts
@@ -548,9 +548,82 @@ describeEmbeddedPostgres("heartbeat orphaned process recovery", () => {
       mockTelemetryClient,
       expect.objectContaining({
         agentRole: "engineer",
-        agentId,
-      }),
+      agentId,
+    }),
     );
+  });
+
+  it("queues wakeups for paused agents instead of rejecting them", async () => {
+    const companyId = randomUUID();
+    const agentId = randomUUID();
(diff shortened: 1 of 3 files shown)


oshabana (commented):
## Review Summary — PR #3955 / 761dbaa2

**Verdict**: APPROVE
**Cycle**: 0 of 3

### Blocking Issues

None.

### Non-blocking Observations

- Reviewed the concern about `/agents/:id/resume` not explicitly draining queued runs. Current repo behavior already resumes queued work during startup and periodic scheduler recovery in `server/src/index.ts:585` and `server/src/index.ts:626`, so this is not a blocker for the scoped fix.

### Convention Compliance

- Kysely only: N/A
- Vitest only: PASS
- No emojis: PASS
- Mobile-first (if frontend): N/A
- No hardcoded domains: PASS

PR #3999, opened 2026-04-18
fix(cursor-local): probe cursor-agent before agent; refresh fallback model IDs

gitSITTI:
## Problem

Two independent regressions that both surface as _Cursor / Composer adapters disappearing from the Paperclip UI_:

### 1 — Wrong default CLI binary name

The adapter hard-coded `command` default to `"agent"`. On Windows, macOS, and most modern Linux installs the Cursor CLI ships as `cursor-agent` (not `agent`). When `ensureCommandResolvable` cannot find `agent` on PATH the adapter silently drops out of the `/api/agents` list.

### 2 — Stale `CURSOR_FALLBACK_MODEL_IDS`

The static model list in `src/index.ts` was missing several models released since the last refresh, causing _model not available_ errors for operators on current Cursor versions.

## Changes

| File | Change |
|------|--------|
| `src/shared/command.ts` | New `resolveCursorDefaultCommand(cwd, env)` — probes `cursor-agent` then `agent`; returns `cursor-agent` if neither found so the downstream error is actionable |
| `src/server/execute.ts` | Replace hard-coded `asString(config.command, "agent")` with auto-resolved default via the helper |
| `src/server/test.ts` | Same default replacement; expand `commandLooksLike` to accept both `agent` and `cursor-agent`; update skip-probe message |
| `src/index.ts` | Add `composer-2`, `composer-2-fast`, `opus-4.7`, `opus-4.7-thinking`, `gemini-2.5-flash`, `grok-4-20-thinking` in family order |

## Backward compatibility

- Operators who **explicitly set `command: agent`** in their adapter config are unaffected (configured value always wins).
- Operators who previously relied on the `agent` default on a host where `cursor-agent` is on PATH will now resolve automatically without any config change.

## Test plan

- [ ] `tsc` passes (verified locally — zero type errors)
- [ ] On a host with `cursor-agent` on PATH: adapter appears in `/api/agents`, hello probe passes
- [ ] On a host with only `agent` on PATH: same result
- [ ] On a host with neither binary: adapter shows `cursor_command_unresolvable` error (not a silent disappearance)
- [ ] New model IDs appear in Paperclip model picker without restart

### DIFF
--- a/packages/adapters/cursor-local/src/index.ts
+++ b/packages/adapters/cursor-local/src/index.ts
@@ -4,6 +4,8 @@ export const DEFAULT_CURSOR_LOCAL_MODEL = "auto";
 
 const CURSOR_FALLBACK_MODEL_IDS = [
   "auto",
+  "composer-2",
+  "composer-2-fast",
   "composer-1.5",
   "composer-1",
   "gpt-5.3-codex-low",
@@ -29,6 +31,8 @@ const CURSOR_FALLBACK_MODEL_IDS = [
   "gpt-5.2-high",
   "gpt-5.1-high",
   "gpt-5.1-codex-mini",
+  "opus-4.7",
+  "opus-4.7-thinking",
   "opus-4.6-thinking",
   "opus-4.6",
   "opus-4.5",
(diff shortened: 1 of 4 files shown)


gitSITTI:
Closing — this fix is being applied privately. Will not be contributing upstream.

Issue #3713, opened 2026-04-15
pruneOldBackups uses a fixed 30-day month for monthlyCutoff, causing premature deletion in longer months


shridipavansaber:
## Description

The monthly retention cutoff in \`pruneOldBackups\` is calculated as:

\`\`\`typescript
const monthlyCutoff = now - Math.max(1, retention.monthlyMonths) * 30 * 24 * 60 * 60 * 1000;
\`\`\`

This uses a fixed 30-day approximation. In months with 31 days, a backup that is exactly 2 calendar months old is actually 61–62 days old, which falls outside the 60-day cutoff (\`monthlyMonths=2 * 30 days\`) and gets deleted prematurely.

## Example

- Today: March 31
- \`monthlyMonths: 2\`
- Cutoff: \`now - 60 days\` = January 30
- A backup from January 28 is 62 days old → deleted
- A backup from January 28 is within 2 calendar months → should be kept

The weekly retention uses \`retention.weeklyWeeks * 7 * 24 * 60 * 60 * 1000\` (correct — weeks are always 7 days), but the monthly equivalent uses 30 days rather than actual calendar month boundaries.

## Suggested fix

Use proper calendar month arithmetic instead of a fixed multiplier:

\`\`\`typescript
const monthlyDate = new Date(now);
monthlyDate.setMonth(monthlyDate.getMonth() - Math.max(1, retention.monthlyMonths));
const monthlyCutoff = monthlyDate.getTime();
\`\`\`

## Notes

Found via code review while running [tailtest](https://github.com/avansaber/tailtest) on this codebase. The bug is in a private function called only by \`runDatabaseBackup\` (requires a live PostgreSQL connection), so it was not caught by the generated test suite.

serenakeyitan:
🌱 **gardener** · ✅ `ALIGNED` · severity: `low` · commit: `issue`

Re-review triggered by a cross-reference only — issue body unchanged. Verdict stands: `pruneOldBackups` month-cutoff bug is implementation correctness within the `engineering/database` backup/restore capability; no tree decisions at stake.

---

<sub>Reviewed commit: <code>issue</code> · Tree snapshot: <code>ddfe6b6</code> · Commands: <code>@gardener re-review</code> · <code>@gardener pause</code> · <code>@gardener ignore</code></sub>

[...]

PR #4031, opened 2026-04-19
fix(db): use calendar months for monthly backup retention

scokeepa:
## Thinking Path

`pruneOldBackups` computes the monthly cutoff as `now - monthlyMonths * 30 * 24h`. Since real months are 28–31 days long, this drifts: every year a backup configured to retain N months actually retains slightly less than N calendar months. After 12 months the drift is ~5 days, which can cause the prune step to delete a monthly backup that the user expected to still be present.

The fix is to compute the cutoff with calendar arithmetic (`Date#setMonth`) instead of a fixed millisecond offset. Daily/weekly cutoffs remain second-based since their units are well-defined.

Refs upstream issue #3713.

## What Changed

- `packages/db/src/backup-lib.ts`: monthly cutoff now uses `new Date(now); setMonth(getMonth() - n)` so the boundary lands on the same calendar day N months ago.

## Verification

- Walked through `setMonth` semantics for edge cases:
  - Mar 31 → setMonth(-1) → Feb 28/29 (clamped, expected behaviour and still produces a valid cutoff in the past).
  - Jan 1 → setMonth(-1) → year boundary handled by Date.
- Daily and weekly windows are unchanged.

## Risks

- Behaviour change: a backup that was about to be pruned by the 30-day approximation may now survive until the equivalent calendar day passes. Effect is at most a few extra days of retention; no data loss.
- No new dependencies, no migration.

## Checklist

- [x] One logical change
- [x] No schema or migration impact

### DIFF
--- a/packages/db/src/backup-lib.ts
+++ b/packages/db/src/backup-lib.ts
@@ -111,7 +111,9 @@ function pruneOldBackups(backupDir: string, retention: BackupRetentionPolicy, fi
   const now = Date.now();
   const dailyCutoff = now - Math.max(1, retention.dailyDays) * 24 * 60 * 60 * 1000;
   const weeklyCutoff = now - Math.max(1, retention.weeklyWeeks) * 7 * 24 * 60 * 60 * 1000;
-  const monthlyCutoff = now - Math.max(1, retention.monthlyMonths) * 30 * 24 * 60 * 60 * 1000;
+  const monthlyCutoffDate = new Date(now);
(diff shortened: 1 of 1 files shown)


serenakeyitan:
🌱 **gardener review**

**Tree alignment**:
- infrastructure/NODE.md: Backup retention correctness fix — calendar months vs fixed 30-day window. No tree decision violated; aligns with operational reliability principle.

**Notes**: Clean targeted fix. `setMonth` handles month-length drift correctly.

Issue #4011, opened 2026-04-19
executionRunId lock leaks on reassignment + /release does not clear it


blevinson:
## Summary

When an agent run crashes or stops mid-heartbeat without reaching a terminal status (`done` / `cancelled`) and without successfully calling release, the issue's `executionRunId` is never cleared. Any subsequent `POST /api/issues/:id/checkout` — even by a freshly assigned agent — returns `409 Conflict`, because the control plane still believes the prior run owns the lock.

The documented self-service remediation, `POST /api/issues/:id/release`, clears `assigneeAgentId` but does **not** clear `executionRunId`, so it does not actually free the lock. We confirmed this live on a real issue (the meta-ticket about stale locks, itself, had a stale lock).

Today the only reliable way to clear `executionRunId` is to transition the issue to `done` or `cancelled`, which is wrong for cases where the work still needs to be done by a different agent.

## Environment

- Paperclip CLI: `paperclipai@2026.416.0`
- Adapter: `claude_local`
- Observed: 2026-04-19

## Steps to reproduce

1. Agent A checks out issue, heartbeat assigns `executionRunId`.
2. Agent A crashes or stops mid-run (does not call release, does not reach terminal status).
3. Issue is reassigned to Agent B.
4. Agent B `POST /checkout` → `409 Conflict` because prior `executionRunId` is still attached.
5. Only `status = done/cancelled` clears `executionRunId` today.

**Additional finding from 2026-04-19:** `POST /api/issues/:id/release` clears `assigneeAgentId` but does NOT clear `executionRunId` — so even the documented self-service path leaks the lock. Confirmed live on a real issue (the meta-ticket about stale locks had a stale lock — ironic and useful).

## Expected behavior

At least one of the following:

1. Release `executionRunId` on reassignment (`assigneeAgentId` change).
2. TTL/reaper that clears `executionRunId` when the owning run has been inactive past a threshold.
3. Force-release admin route (`POST /api/issues/:id/force-release`) callable by company CEO / board.
4. Make `POST /release` also null out `executionRunId`, not just `assigneeAgentId`.

Preference is **(1) + (4)**: reassignment should always release the lock, and the `release` endpoint should match its name. (3) is a reasonable additional safety net.

## Actual behavior

- `checkout` → 409 forever, until someone flips status to `done` / `cancelled`.
- `release` clears the assignee but silently leaves `executionRunId` intact, giving a false sense of recovery.
[...]

PR #4033, opened 2026-04-19
fix(issues): clear execution lock fields on release

scokeepa:
## Thinking Path

`releaseIssue` is the recovery path used when an agent abandons or hands off an issue. It currently resets `status`, `assigneeAgentId`, and `checkoutRunId` — but the parallel `execution*` lock columns (`executionRunId`, `executionAgentNameKey`, `executionLockedAt`) are left as they were.

Those execution fields are checked elsewhere when a new run tries to claim the issue. Stale values look like an active claim by a run that no longer exists, so the next agent gets blocked or the issue silently stalls in `todo`.

The fix is to null out the execution-lock triple in the same UPDATE — the same way `assigneeAgentId` / `checkoutRunId` are already cleared.

Refs upstream issue #4011.

## What Changed

- `server/src/services/issues.ts`: `releaseIssue` now sets `executionRunId`, `executionAgentNameKey`, and `executionLockedAt` to `null` alongside the existing reset fields.

## Verification

- The same fields are nulled together in other recovery paths (e.g. when a heartbeat reaper expires a stuck run), so the schema and downstream consumers already handle a fully-cleared lock.
- No new query or column references; existing UPDATE was just missing three set-clauses.

## Risks

- Behaviour change is intentional and narrow: any subsequent run can now successfully claim an issue that was released — which is the documented purpose of release.

## Checklist

- [x] One logical change
- [x] No schema changes

### DIFF
--- a/server/src/services/issues.ts
+++ b/server/src/services/issues.ts
@@ -2026,6 +2026,9 @@ export function issueService(db: Db) {
           status: "todo",
           assigneeAgentId: null,
           checkoutRunId: null,
+          executionRunId: null,
+          executionAgentNameKey: null,
+          executionLockedAt: null,
           updatedAt: new Date(),
         })
         .where(eq(issues.id, id))


serenakeyitan:
🌱 **gardener review**

**Tree alignment**:
- product/task-system/NODE.md: Atomic checkout/release — clearing `executionRunId`, `executionAgentNameKey`, `executionLockedAt` on release preserves the lock invariant so the next agent can cleanly claim.

Issue #3822, opened 2026-04-16
Parent-issue assignees not auto-woken on child status changes (`blocked`/`done`/`cancelled`)


hansmon:
## Summary

`wakeOnDemand` triggers agent heartbeats on assignment changes to the agent's OWN issues, but NOT on status changes to CHILD issues. When a child issue changes status (e.g., engineer sets a sub-task to `blocked`), the parent issue's assignee (e.g., ops manager or EA) is NOT woken. Coordinators sit idle while downstream work stalls or completes — "invisible walls."

## Environment

- `paperclipai@2026.403.0`
- `@paperclipai/server@2026.403.0`
- Self-hosted

## Reproduction

1. Create a parent issue assigned to Agent A (e.g., ops manager).
2. Create a child issue (use `parentId`) assigned to Agent B (e.g., engineer).
3. Agent B claims + works the child issue, hits a blocker, sets status to `blocked`.
4. Expected: Agent A wakes on the child status change so they can unblock / escalate.
5. Actual: Agent A sits idle. No `wakeOnDemand` fires on Agent A unless someone posts a comment on the parent with `@Agent A` OR reassigns the parent.

## Our workaround

We codified a universal agent rule: on any status change to `blocked` / `done` / `cancelled`, the agent MUST post an explicit `@{parent-assignee-name}` comment on the parent issue. Belt-and-suspenders — works at the prose level (comments DO trigger wakeOnDemand for the mentioned agent), but a platform-side propagation would eliminate the need for agent-side compliance.

## Proposed behavior

On child-issue status transition (`blocked`, `done`, `cancelled`, optionally also `in_review`), fire a wakeup request for the PARENT issue's assignee:

```
reason: 'child_status_change'
source: 'automation'
payload: {
  childIssueId: '<uuid>',
  childIdentifier: 'TWF-X',
  childStatus: 'blocked',
  childAssigneeAgentId: '<uuid>',
}
```

Optional config: per-agent opt-out flag `runtime_config.heartbeat.wakeOnChildStatus = true/false` if some agents should not receive these (e.g., a strategic EA who should only see rolled-up briefings).

## Impact

Factories with hierarchical work flows (EA → Ops Mgr → Engineers) hit this routinely. Every blocker that reaches an engineer becomes an invisible wall for the coordinator above, unless the engineer remembers to @-mention manually. Not a showstopper (our @-mention rule closes the gap at the prose layer), but clean platform behavior would be more reliable than agent-compliance-dependent.

[...]

PR #4034, opened 2026-04-19
fix(issues): wake parent agent when a child becomes blocked

scokeepa:
## Thinking Path

When a child issue reaches a terminal state, the parent's assignee gets a wakeup so it can replan. Today \"terminal\" means `done` or `cancelled` only. If a child transitions to `blocked`, the parent is never woken — the parent agent has no way to know progress on its sub-task has stalled until the next manual touch.

`blocked` is functionally a stop signal for the child's branch of work; the parent typically needs to either unblock it, reassign, or abandon. Treating it as a wake trigger lets the parent react immediately, which matches how `done` and `cancelled` are handled.

To keep the wake informative, distinguish the cause: `issue_child_blocked` vs `issue_children_completed`, and surface the child's status in the payload/contextSnapshot so the wake recipient can branch on it.

Refs upstream issue #3822.

## What Changed

- `server/src/routes/issues.ts`: introduced `PARENT_WAKE_STATUSES = [done, cancelled, blocked]`; `becameTerminal` now uses that set.
- Branched `wakeReason` and `source` strings on `issue.status === \"blocked\"`.
- Added `childStatus` to both `payload` and `contextSnapshot`.

## Verification

- The `done`/`cancelled` paths produce the same `reason` and `source` strings as before — backwards-compatible for existing wake handlers.
- The new `blocked` path uses unique strings (`issue_child_blocked` / `issue.child_blocked`) so any consumer that special-cased the old reasons keeps working unchanged.

## Risks

- Parent agents will now wake on `blocked` transitions they previously ignored. Volume should be small (only direct children of an assigned parent).
- Wake handlers that key off `reason` see a new value. They should default to a generic re-plan if unknown — verified the existing handlers in this repo do exactly that.

## Checklist

- [x] One logical change
- [x] Backwards-compatible payload (extra fields only)

### DIFF
--- a/server/src/routes/issues.ts
+++ b/server/src/routes/issues.ts
@@ -1927,29 +1927,34 @@ export function issueRoutes(
         }
       }
 
+      const PARENT_WAKE_STATUSES = ["done", "cancelled", "blocked"] as const;
       const becameTerminal =
-        !["done", "cancelled"].includes(existing.status) && ["done", "cancelled"].includes(issue.status);
+        !PARENT_WAKE_STATUSES.includes(existing.status as typeof PARENT_WAKE_STATUSES[number])
+        && PARENT_WAKE_STATUSES.includes(issue.status as typeof PARENT_WAKE_STATUSES[number]);
       if (becameTerminal && issue.parentId) {
         const parent = await svc.getWakeableParentAfterChildCompletion(issue.parentId);
         if (parent) {
+          const wakeReason = issue.status === "blocked" ? "issue_child_blocked" : "issue_children_completed";
           addWakeup(parent.assigneeAgentId, {
             source: "automation",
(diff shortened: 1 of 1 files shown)


serenakeyitan:
🌱 **gardener review**

**Verdict**: NEEDS_REVIEW (severity: low)

**Tree alignment**:
- product/task-system/NODE.md (inter-agent): Parent-wake now fires on child `blocked` too. Reason `issue_child_blocked` is a new wake trigger beyond `children_completed`.
- product/NODE.md #6 (surface problems): Aligned — bubbling blockage up to the parent helps.

**Notes**: Worth a short tree note listing wake triggers and their reasons so future agents know `blocked` counts as terminal-for-wake purposes.

Issue #3789, opened 2026-04-16
agent.run.finished / failed / cancelled events declared but never emitted


devli13:
## Bug

The event names `agent.run.finished`, `agent.run.failed`, and `agent.run.cancelled` are declared in `@paperclipai/shared/dist/constants.js` (lines 542-545) and documented in `@paperclipai/plugin-sdk/README.md` as subscribable plugin events. However, no code in the server actually emits them.

### Where they're declared

```js
// @paperclipai/shared/dist/constants.js:542-545
"agent.run.started",
"agent.run.finished",
"agent.run.failed",
"agent.run.cancelled",
```

### What happens

A plugin that subscribes via `ctx.events.on("agent.run.finished", ...)` never receives any events. The handler is registered but never fires.

The heartbeat service uses `publishLiveEvent()` with type `heartbeat.run.status` for SSE/websocket live events, but this does NOT go through the plugin event bus (`pluginEventBus.emit()`). So plugins have no way to react to run completion.

### Impact

Any plugin that needs to perform cleanup, send notifications, or update state when an agent run completes must resort to polling the `/api/heartbeat-runs/:id` endpoint in a `setInterval` loop. This is wasteful and adds 10-15s latency vs an event-driven approach.

### Expected behavior

When a heartbeat run transitions to `succeeded`, `failed`, or `cancelled` status, the server should emit the corresponding `agent.run.finished` / `agent.run.failed` / `agent.run.cancelled` event through the plugin event bus, with a payload containing at minimum `{ runId, agentId, companyId, status }`.

### Environment

- `@paperclipai/shared` version referenced from `paperclipai` CLI v2026.318+
- `@paperclipai/plugin-sdk` README documents these events as available
- Server: `@paperclipai/server` — `services/heartbeat.js` uses `publishLiveEvent` but not `pluginEventBus.emit`

serenakeyitan:
🌱 **gardener** · ✅ `ALIGNED` · severity: `low` · commit: `issue`

Re-review triggered by a cross-reference only — issue body unchanged. Verdict stands: missing `pluginEventBus.emit()` for `agent.run.finished / failed / cancelled` is a legitimate gap between declared event contract and emit path. Fix aligns with `plugins/runtime` event-bus surface and the `companyId`-scoped payload matches tree's company-scoped isolation rule.

---

[...]

PR #4035, opened 2026-04-19
fix(heartbeat): emit agent.run.* activity events on status transitions

scokeepa:
## Thinking Path

Plugins subscribe to `agent.run.started`, `agent.run.finished`, `agent.run.failed`, and `agent.run.cancelled` to react to run lifecycle changes. Today, `setRunStatus` publishes the in-process event-bus notification but doesn't write a corresponding `activity` row. Any plugin or audit consumer that hydrates from the activity stream — instead of subscribing to the in-process bus — never sees those transitions.

The fix is to map the status string to a single canonical plugin event name and `logActivity` it from the same code path that handles the in-process publish, so both surfaces stay consistent.

Refs upstream issue #3789.

## What Changed

- `server/src/services/heartbeat.ts`: added `mapRunStatusToPluginEvent` mapping `running → agent.run.started`, `succeeded → agent.run.finished`, `failed`/`timed_out → agent.run.failed`, `cancelled → agent.run.cancelled`. Returns `null` for `queued` (no event needed).
- `setRunStatus` calls `logActivity` after the in-process publish when the mapping returns a non-null event, with the same details payload (status, source, error, timing).

## Verification

- `mapRunStatusToPluginEvent` returns `null` for all non-mapped statuses, so the activity write is skipped — no spurious rows for transient states.
- The in-process publish path is unchanged, so existing in-process subscribers continue to receive events at the same time as before.
- Activity write happens after the DB update has succeeded, so we never log an event for a status change that didn't persist.

## Risks

- New activity rows on every run lifecycle transition. Volume scales with run throughput; acceptable since it matches the documented plugin event surface.
- No schema changes.

## Checklist

- [x] One logical change
- [x] No new dependencies

### DIFF
--- a/server/src/services/heartbeat.ts
+++ b/server/src/services/heartbeat.ts
@@ -2125,6 +2125,24 @@ export function heartbeatService(db: Db) {
       .then((rows) => rows[0]);
   }
 
+  function mapRunStatusToPluginEvent(
+    status: string,
+  ): "agent.run.started" | "agent.run.finished" | "agent.run.failed" | "agent.run.cancelled" | null {
+    switch (status) {
+      case "running":
+        return "agent.run.started";
+      case "succeeded":
+        return "agent.run.finished";
+      case "failed":
+      case "timed_out":
+        return "agent.run.failed";
+      case "cancelled":
+        return "agent.run.cancelled";
+      default:
+        return null;
+    }
+  }
+
   async function setRunStatus(
     runId: string,
     status: string,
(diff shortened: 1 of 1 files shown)


serenakeyitan:
🌱 **gardener review**

**Tree alignment**:
- engineering/backend/heartbeat-run-orchestration/NODE.md: Emitting `agent.run.started|finished|failed|cancelled` activity on status transitions supports plugin observability.
- product/agent-model/NODE.md: Lifecycle events are a documented surface; this fills a gap.

**Notes**: Mapping `timed_out → failed` is reasonable; consider a distinct event later if plugins need to branch.

Issue #3831, opened 2026-04-16
PATCH /agents/me returns 404 (silently drops every agent self-update)


bscottdavis:
### Version
`paperclipai@2026.416.0`.

### Severity
Medium–High. Every agent heartbeat from `claude_local` adapters that tries to update its own metadata via `/agents/me` is silently dropped. One long-running instance logged thousands of these 404s over several weeks.

### Repro
1. As an agent with a valid bearer token: \`\`\`
   PATCH /api/agents/me
   Content-Type: application/json

   { \"metadata\": { \"lastHeartbeatNote\": \"…\" } }
   \`\`\`
2. Response: 404, body `{\"error\":\"not found\"}`.

### Root cause
In `@paperclipai/server` `routes/agents.js` (built `dist/routes/agents.js`):
- `GET /agents/me` → exists (~line 905)
- `GET /agents/me/inbox-lite` → exists
- `GET /agents/me/inbox/mine` → exists
- `PATCH /agents/:id` → exists (~line 1506)

…but no `PATCH /agents/me` handler. So `PATCH /agents/me` matches `/agents/:id` with `id = \"me\"`, which fails the DB lookup → 404. The documented agent-heartbeat behavior (and default agent instructions, which tell agents to `GET /api/agents/me`) implies the `me` alias resolves on both verbs, but it doesn't.

### Impact
Agents that infer a PATCH from the symmetric GET — including natural behavior from LLM agents on the `claude_local` adapter — lose every self-update. No error is surfaced to the agent beyond the 404, so the assumption \"my heartbeat note was recorded\" silently holds forever.

### Fix
Add a `PATCH /agents/me` handler that resolves the agent from the auth token's `sub` and then delegates to the same logic as `PATCH /agents/:id`. Equivalently: add a small middleware on `/agents/me` that rewrites `req.params.id = req.auth.agentId` and calls the existing handler.

### Workaround (for users hitting this now)
Update agent heartbeat templates to resolve `GET /agents/me` → `.id` first, then `PATCH /agents/{resolvedId}` with the new metadata. Or just drop the PATCH entirely — the runtime already records the run via `heartbeat_runs`.

### Related
Part of a group of 3 issues from the same diagnosis session:
- #3829 — heartbeat-runs endpoint UTF-8 500.
- #3830 — comments cursor 500 (Date passed as bind param).

serenakeyitan:
null

serenakeyitan:
[...]

PR #4036, opened 2026-04-19
fix(agents): add PATCH /agents/me self-update route

scokeepa:
## Thinking Path

The server exposes `GET /agents/me` so an agent can fetch its own profile using only its bearer token. The corresponding `PATCH /agents/me` is missing — to update its own record an agent has to know its own UUID and hit `PATCH /agents/:id`. Agents authenticated by token don't always have that ID handy (it's never returned to them in some flows), so the only options today are either an extra round-trip to `/agents/me` or pass the ID through every adapter manually.

The fix is to register `PATCH /agents/me` as a thin pre-handler that resolves the actor's `agentId` and forwards to the existing handler logic (extracted into a shared function so both routes share validation and behaviour).

Refs upstream issue #3831.

## What Changed

- `server/src/routes/agents.ts`:
  - Added `router.patch(\"/agents/me\", ...)` with a guard that 401s for non-agent callers, sets `req.params.id` from `req.actor.agentId`, then runs the same `validate` middleware and shared handler.
  - Extracted the existing `PATCH /agents/:id` body into `updateAgentHandler(req, res)` and re-registered the `:id` route to call it.

## Verification

- The shared handler keeps the existing `validate(updateAgentSchema)` chain on both routes.
- The `/me` guard returns 401 with a clear message when called by a non-agent actor (e.g. user session, board admin) — matches the auth model used by the existing `GET /agents/me`.
- No behaviour change for `PATCH /agents/:id`; same handler, same middleware, same status codes.

## Risks

- New endpoint surface only — existing routes untouched.
- Handler is now shared between two routes; any future change must keep both call sites in mind. Function name (`updateAgentHandler`) is unambiguous.

## Checklist

- [x] One logical change
- [x] Auth guard on the new endpoint

### DIFF
--- a/server/src/routes/agents.ts
+++ b/server/src/routes/agents.ts
@@ -1919,7 +1919,16 @@ export function agentRoutes(db: Db) {
     res.json(result.bundle);
   });
 
-  router.patch("/agents/:id", validate(updateAgentSchema), async (req, res) => {
+  router.patch("/agents/me", (req, res, next) => {
+    if (req.actor.type !== "agent" || !req.actor.agentId) {
+      res.status(401).json({ error: "Agent authentication required" });
+      return;
+    }
+    req.params.id = req.actor.agentId;
+    next();
+  }, validate(updateAgentSchema), (req, res) => updateAgentHandler(req, res));
+
+  const updateAgentHandler = async (req: Request, res: Response) => {
     const id = req.params.id as string;
     const existing = await svc.getById(id);
     if (!existing) {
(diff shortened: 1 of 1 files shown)


serenakeyitan:
🌱 **gardener review**

**Verdict**: NEEDS_REVIEW (severity: low)

**Tree alignment**:
- product/agent-model/NODE.md: New `PATCH /agents/me` lets an agent self-update. Scope is implicit (the caller's own agentId); confirm company-scoped isolation holds through `updateAgentHandler` (company resolved from agent record, not from request).

**Notes**: Forward-declared `const updateAgentHandler` used before assignment — works at runtime (route handlers run async) but is a readability risk. Consider declaring the handler first, then mounting both routes.

Issue #3845, opened 2026-04-16
bug: GET /skills returns 500 due to duplicate key race in upsertImportedSkills


wfrank2509:
## Summary

`GET /api/companies/:companyId/skills` returns HTTP 500 with PostgreSQL error:

```
duplicate key value violates unique constraint "company_skills_company_key_idx"
Key (company_id, key)=(af26c727-..., paperclipai/paperclip/phase-closure) already exists.
```

This endpoint is read-only from the caller's perspective, but internally it triggers `ensureSkillInventoryCurrent` which runs `ensureBundledSkills` → `upsertImportedSkills` on every request.

## Root Cause

`upsertImportedSkills` uses a manual check-then-insert pattern:

```typescript
const existing = await getByKey(companyId, skill.key);
// ...
const row = existing
  ? await db.update(companySkills).set(values).where(eq(companySkills.id, existing.id))...
  : await db.insert(companySkills).values(values)...  // plain INSERT, no conflict handling
```

This is a TOCTOU (time-of-check/time-of-use) race condition. Two concurrent server processes (e.g. multiple workers, or two requests that arrive during a server restart window) can both call `getByKey` and receive `null`, then both attempt an `INSERT`. The second insert fails with a unique constraint violation on `(company_id, key)`.

The in-process deduplication via `skillInventoryRefreshPromises` only prevents concurrent refreshes within a single Node.js process — it provides no protection across multiple processes or workers.

## Affected endpoint

```
GET /api/companies/:companyId/skills
```

## Fix

Changed the `INSERT` in `upsertImportedSkills` to use a database-level atomic upsert:

```typescript
: await db
    .insert(companySkills)
    .values(values)
    .onConflictDoUpdate({
      target: [companySkills.companyId, companySkills.key],
      set: {
        slug: values.slug,
        name: values.name,
        description: values.description,
        markdown: values.markdown,
        sourceType: values.sourceType,
        sourceLocator: values.sourceLocator,
        sourceRef: values.sourceRef,
        trustLevel: values.trustLevel,
        compatibility: values.compatibility,
        fileInventory: values.fileInventory,
        metadata: values.metadata,
        updatedAt: values.updatedAt,
      },
    })
    .returning()
    ...
```

`INSERT ... ON CONFLICT DO UPDATE` is atomic at the database level and eliminates the race entirely — concurrent inserts for the same `(company_id, key)` will serialize correctly instead of one failing.

## Impact

[...]

PR #4037, opened 2026-04-19
fix(company-skills): atomic upsert with bundled-skill guard

scokeepa:
## Thinking Path

`upsertImportedSkills` does `getByKey` → branch → `INSERT or UPDATE` for each imported skill. That's a textbook check-then-act race:
- Two concurrent imports for the same `(companyId, key)` both see \"no row\".
- Both run the INSERT branch.
- The second one trips the unique constraint and the import aborts mid-way through the list.

The same function also has a guard that prevents overwriting `paperclip_bundled` skills with the upstream `paperclipai/paperclip` github mirror copy. That guard runs on the read snapshot, so under contention it can also be bypassed (read says \"not bundled\" but a concurrent insert just made it bundled).

The fix is to collapse SELECT-then-INSERT/UPDATE into a single atomic upsert, and express the bundled guard as a SQL predicate that runs inside the same transaction.

Refs upstream issue #3845.

## What Changed

- `server/src/services/company-skills.ts`:
  - Single `db.insert(companySkills).values(...).onConflictDoUpdate({...})` per skill.
  - Conflict target: `(companyId, key)` (matches the existing unique index).
  - For incoming bundled-mirror payloads, the upsert carries `setWhere: coalesce(metadata->>'sourceKind','') <> 'paperclip_bundled'` so the UPDATE only runs when the existing row is *not* the bundled skill.
  - When `setWhere` blocks the update *and* the INSERT no-ops (conflict path), re-fetch by key and return the existing row.
  - Added `sql` to the drizzle-orm import for the predicate.

## Verification

- Walked through both branches:
  - Non-bundled-mirror import: behaves as a plain conditional upsert; identical end state to the original code.
  - Bundled-mirror import on top of a `paperclip_bundled` row: `setWhere` blocks the update, returning row is null, fallback `getByKey` returns the bundled row — same protection as the previous in-process check, now race-safe.
  - Bundled-mirror import on top of a non-bundled row: `setWhere` evaluates true; the row is overwritten as expected.
- Concurrent-import scenario: both inserts hit the conflict, only one update wins, neither raises a constraint violation.

## Risks

- New SQL predicate runs once per imported skill; no measurable cost change because we removed the prior SELECT.
- `companyId` / `key` are excluded from the `set` payload to avoid changing the conflict target, matching standard upsert practice.

## Checklist

- [x] One logical change
- [x] No schema changes

### DIFF
--- a/server/src/services/company-skills.ts
+++ b/server/src/services/company-skills.ts
@@ -2,7 +2,7 @@ import { createHash } from "node:crypto";
 import { promises as fs } from "node:fs";
 import path from "node:path";
 import { fileURLToPath } from "node:url";
-import { and, asc, eq } from "drizzle-orm";
+import { and, asc, eq, sql } from "drizzle-orm";
 import type { Db } from "@paperclipai/db";
 import { companySkills } from "@paperclipai/db";
 import { readPaperclipSkillSyncPreference } from "@paperclipai/adapter-utils/server-utils";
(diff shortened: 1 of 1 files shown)


serenakeyitan:
🌱 **gardener review**

**Tree alignment**:
- product/NODE.md #8 (local-first) & task-system atomic semantics: Replacing read-then-write with `INSERT … ON CONFLICT DO UPDATE … WHERE sourceKind<>'paperclip_bundled'` closes a race and preserves the bundled-skill guard.

**Notes**: Nice pattern — DB-level guard beats app-level check. Worth documenting as the canonical upsert pattern if it gets reused.

Issue #3919, opened 2026-04-17
Inbox badge counts done issues — should only count open/unread items


egoldy:
The inbox badge count includes issues in done status, causing the count to be permanently inflated even after all work is complete.

Steps to reproduce:

Create several issues and complete them (status → done)
Observe the inbox badge count — it continues to include all done issues
Expected behavior:
Badge count reflects only open/unread items (e.g. todo, in_progress, in_review, blocked)

Actual behavior:
Badge count = all non-cancelled issues, including done ones

Example:
Personal company has 10 done + 1 todo = badge shows 11, when it should show 1

Environment:

Self-hosted, latest from main branch
Reproduced via API: GET /companies/{id}/issues confirms issues are done, yet badge remains elevated after hard refresh

serenakeyitan:
🌱 **gardener** · ✅ `ALIGNED` · severity: `low`

> **What is this?** repo-gardener checks whether PRs and issues fit the project's **product decisions, architecture, and roadmap** — not code correctness. Think of it as a product-context review layer. For code review, see Greptile/CodeRabbit.

### Context fit

| Area | This Issue | Tree guidance | Fit |
|------|------------|---------------|-----|
| Board-level visibility | Badge should reflect actionable items only (open/unread), not completed work | "Board-level abstraction wins" — default experience surfaces what needs attention; "Surface problems, don't hide them" implies counts should reflect real signal, not noise | ✅ Fits |
| Progressive disclosure | Badge is the highest-level signal; done items are already resolved work | Progressive disclosure principle: top layer is human-readable summary of active state | ✅ Fits |
| Issues as tasks | Bug report on issue status-filtered badge count | Issues/tasks have explicit statuses (`todo`, `in_progress`, `in_review`, `blocked`, `done`, `cancelled`) — UI filtering on status is expected behaviour | ✅ Fits |

<details>
<summary><strong>Tree nodes referenced</strong></summary>

- [`product/NODE.md`](https://github.com/serenakeyitan/paperclip-tree/blob/main/product/NODE.md) — Key decisions #4 (board-level abstraction), #6 (surface problems, not noise), #7 (progressive disclosure)
- [`engineering/frontend/NODE.md`](https://github.com/serenakeyitan/paperclip-tree/blob/main/engineering/frontend/NODE.md) — Issues page with inbox and badge UI; sidebar-badges backend route

</details>

[...]

PR #4038, opened 2026-04-19
fix(inbox): exclude done issues from inbox badge count

scokeepa:
## Thinking Path

The inbox badge fetches issues with statuses `backlog,todo,in_progress,in_review,blocked,done`. Including `done` means every completed-but-not-archived issue keeps inflating the unread count, even though the badge is supposed to surface work that still needs attention.

The intended semantics are \"things still on your plate\". Finished work should not count toward the badge — the user already finished it.

The fix is a one-token change to the status list. Other inbox surfaces (the inbox tab itself, dashboard) load their own queries with their own status sets, so they're unaffected.

Refs upstream issue #3919.

## What Changed

- `ui/src/hooks/useInboxBadge.ts`: removed `done` from `INBOX_ISSUE_STATUSES`.

## Verification

- The badge count is computed from this query result via `computeInboxBadgeData`, which already filters by archive/dismissal. Removing `done` only narrows the set going in.
- The inbox view itself uses different queries and continues to display done issues; the change is scoped to the badge count.

## Risks

- Visible-to-user behaviour change: badge counts may decrease for users with many finished-but-not-archived issues. This is the intended correction.

## Checklist

- [x] One logical change
- [x] Touches only the badge query

### DIFF
--- a/ui/src/hooks/useInboxBadge.ts
+++ b/ui/src/hooks/useInboxBadge.ts
@@ -20,7 +20,7 @@ import {
   READ_ITEMS_KEY,
 } from "../lib/inbox";
 
-const INBOX_ISSUE_STATUSES = "backlog,todo,in_progress,in_review,blocked,done";
+const INBOX_ISSUE_STATUSES = "backlog,todo,in_progress,in_review,blocked";
 const INBOX_BADGE_HEARTBEAT_RUN_LIMIT = 200;
 
 export function useDismissedInboxAlerts() {


serenakeyitan:
🌱 **gardener review**

**Tree alignment**:
- product/task-system/NODE.md (issue states): `done` dropped from the inbox badge statuses — matches the semantic that finished work shouldn't keep nagging.

Issue #3820, opened 2026-04-16
UI shows "Waiting for output" on `blocked` issues — state confusion


hansmon:
## Summary

When an issue is in `blocked` status, the issue detail page still renders a "Live Runs" card showing `Waiting for run output`. This is state-confusing — by definition the issue is NOT executing work (it's blocked), so nothing is actively pending, yet the UI implies something is.

## Environment

- `paperclipai@2026.403.0`
- `@paperclipai/server@2026.403.0`
- Self-hosted on Mac Mini; viewed via Safari on MBP + Safari on iPhone (both show the same state)

## Impact

User trust / UX friction. If `blocked` means "nothing is happening until somebody does X" but the UI shows "waiting for output," users see mixed signals and have to second-guess which state is real.

Worse edge case: a run CAN still be firing on a `blocked` issue (e.g., agent waking on a new comment that doesn't unblock). The UI conflates these two different meanings of "waiting for output" — is it blocked waiting for Board? Is work in flight? Today the card doesn't distinguish.

Observed twice in our deployment — on a `pending_approval` hire issue and on a production-incident issue.

## Proposed behavior

On issue detail page, when rendering the "Live Runs" card:

- If `issue.status = 'blocked'` AND no actively-executing run (no run with status `running` / `pending` / `claimed`) → **hide the card** OR render with explicit framing: "Blocked — no active work. See blocker comment above."
- If `issue.status = 'blocked'` AND a run IS active (e.g., agent responding to a comment that doesn't unblock) → render the run but with explicit framing: "Agent responding to comment — does not unblock. Still waiting on X."
- General rule: the `blocked` status should be the primary state indicator; any run activity is secondary context.

Low priority / cosmetic, but accumulated trust erosion across many issues adds up over a factory's life.

serenakeyitan:
null

serenakeyitan:
🌱 **gardener** · ✅ `ALIGNED` · severity: `low`

> **What is this?** repo-gardener checks whether PRs and issues fit the project's **product decisions, architecture, and roadmap** — not code correctness.

### Context fit

| Area | This Item | Tree guidance | Fit |
|------|-----------|---------------|-----|
| Issue status display | "Waiting for output" shown on `blocked` issues — contradicts the clear `blocked` semantic | Task system: fixed status workflow with explicit semantics; `blocked` means no active work pending resolution of the blocker | ✅ Aligned |
[...]

PR #4039, opened 2026-04-19
fix(ui): treat blocked issues as inactive in run chat surface

scokeepa:
## Thinking Path

`RunChatSurface` decides whether the chat is \"live\" purely from the run status (`queued`/`running`). When the parent issue is `blocked`, the run is functionally suspended — but the chat surface still presents itself as live, with the composer enabled and the empty state saying \"Waiting for run output…\".

Conceptually, a blocked issue means \"don't act\". The surface should reflect that immediately, even if a stale `queued` row hasn't yet been reaped.

This requires the issue's status to be visible at the surface level, so I added an optional `issueStatus` prop and threaded it through the two existing call sites. The prop is optional, so any other consumer keeps the previous behaviour.

Refs upstream issue #3820.

## What Changed

- `ui/src/components/RunChatSurface.tsx`: added optional `issueStatus?: string | null`. Active calculation now `isRunActive(run) && issueStatus !== \"blocked\"`.
- `ui/src/components/LiveRunWidget.tsx`: forwarded a new optional `issueStatus` prop into the surface.
- `ui/src/components/ActiveAgentsPanel.tsx`: passes `issueStatus={issue?.status ?? null}` to the surface (it already loads the issue per card).

## Verification

- Default behaviour preserved: when callers don't pass `issueStatus`, the predicate `issueStatus !== \"blocked\"` is `true` and the active branch runs exactly as before.
- LiveRunWidget receives the prop optionally — issue pages that already have the status can pass it; ones that don't are unaffected.
- ActiveAgentsPanel resolves the issue from its existing query map; if the issue isn't loaded yet, the prop is `null` and the surface stays in its previous behaviour.

## Risks

- Pure additive prop, no required call-site changes, no schema or API impact.
- Visible behaviour change only kicks in when the prop is explicitly passed and the issue is `blocked`.

## Checklist

- [x] One logical change
- [x] Backwards-compatible prop addition

### DIFF
--- a/ui/src/components/ActiveAgentsPanel.tsx
+++ b/ui/src/components/ActiveAgentsPanel.tsx
@@ -149,6 +149,7 @@ function AgentRunCard({
           transcript={transcript}
           hasOutput={hasOutput}
           companyId={companyId}
+          issueStatus={issue?.status ?? null}
         />
       </div>
     </div>
--- a/ui/src/components/LiveRunWidget.tsx
+++ b/ui/src/components/LiveRunWidget.tsx
@@ -13,6 +13,7 @@ import { useLiveRunTranscripts } from "./transcript/useLiveRunTranscripts";
 interface LiveRunWidgetProps {
   issueId: string;
   companyId?: string | null;
+  issueStatus?: string | null;
 }
 
 function toIsoString(value: string | Date | null | undefined): string | null {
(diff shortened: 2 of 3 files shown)


serenakeyitan:
🌱 **gardener review**

**Tree alignment**:
- product/task-system/NODE.md: Treating `blocked` issues as inactive in the run chat surface matches the state-machine intent (blocked = not progressing).
- product/NODE.md #6 (surface problems): UI now reflects blockage accurately.

Issue #3676, opened 2026-04-14
Timeout too short for local LLM servers (Ollama/LM Studio) — connection aborted before response completes


rolfberkenbosch:
**Bug Report**
Describe the bug
When using Paperclip AI with a local LLM backend (Ollama or LM Studio), Paperclip disconnects from the server before it finishes generating a response. This causes a cascade of 500 errors and makes local model usage effectively broken for any task that takes more than ~5 minutes to complete.

The logs show that Paperclip closes the connection at exactly 5 minutes (5m3s), even though the local server is still actively processing the request. Ollama then logs:

`aborting completion request due to client closing the connection`

This hard 5-minute cutoff appears to be ignored even when timeout values are explicitly configured.

**Setup**
Paperclip AI → Claude Code (local) → Ollama (local)

**Environment**
- Local model server: Ollama (also reproduced with LM Studio)
- Model: gemma4:26b
- Network: local VM (192.168.64.x)

**Logs**
```
 [GIN] 2026/04/14 - 06:50:33 | 500 | 5m3s | 192.168.64.2 | POST "/v1/messages?beta=true"
[GIN] 2026/04/14 - 06:51:07 | 500 | 5m3s | 192.168.64.2 | POST "/v1/messages?beta=true"
...
time=2026-04-14T06:56:11.929+02:00 level=INFO source=runner.go:922 msg="aborting completion request due to client closing the connection"
...
[GIN] 2026/04/14 - 06:58:03 | 200 | 2m24s | 192.168.64.2 | POST "/v1/messages?beta=true"
[GIN] 2026/04/14 - 06:58:39 | 200 | 2m26s | 192.168.64.2 | POST "/v1/messages?beta=true"
```
Requests under ~5 minutes succeed (200). Requests that take longer are always killed (500).

**What I tried**

**1. Claude Code settings.json — increased timeouts:**
```
json{
  "env": {
    "API_TIMEOUT_MS": "3600000",
    "CLAUDE_STREAM_IDLE_TIMEOUT_MS": "3600000",
    "BASH_DEFAULT_TIMEOUT_MS": "900000"
  }
}
```
→ No effect.

**2. Ollama environment variables:**
```
bashlaunchctl setenv OLLAMA_KEEP_ALIVE "60m"
launchctl setenv OLLAMA_NUM_PARALLEL "1"
launchctl setenv OLLAMA_MAX_QUEUE "0"
```
→ No effect.

**3. Paperclip .env file:**
```
[...]

PR #4040, opened 2026-04-19
fix(claude-local): make fetchWithTimeout timeout env-configurable

scokeepa:
## Thinking Path

`fetchWithTimeout` in the claude-local adapter quota module hard-codes an 8000ms abort. That's fine on a fast local connection, but on slow links or when traffic goes through a corporate proxy the Anthropic quota endpoint can take longer than 8s to respond. The abort fires first, the call surfaces as a provider error, and the UI shows a misleading failure even though the upstream API would have answered.

The fix is to make the default overridable via env without changing the function signature, so:
- Operators can tune the timeout for their environment.
- Callers that already pass `ms` keep their explicit value.
- Bad env values (empty string, non-numeric, ≤ 0) silently fall back to the original 8s default — env misconfiguration must not disable the timeout.

Refs upstream issue #3676.

## What Changed

- `packages/adapters/claude-local/src/server/quota.ts`:
  - Added `defaultFetchTimeoutMs()` that reads `PAPERCLIP_CLAUDE_LOCAL_FETCH_TIMEOUT_MS` and returns the parsed positive number, or 8000 as fallback.
  - `fetchWithTimeout(url, init, ms?)`: `ms` is now optional; when omitted, uses the env-derived default. Caller-supplied `ms` still wins.

## Verification

- `Number.isFinite && > 0` guards against accidental disable (negative, zero, NaN, empty string all keep the safe default).
- Existing call sites pass either nothing (now → env-or-8000) or an explicit number (unchanged behaviour).
- No new dependency, no new export surface.

## Risks

- New env var must be documented; no existing deployment behaviour changes when it's unset.
- Operators who set an extremely large value can mask hung connections — same trade-off as any timeout knob; the function still aborts, just later.

## Checklist

- [x] One logical change
- [x] Defaults preserved when env is unset

### DIFF
--- a/packages/adapters/claude-local/src/server/quota.ts
+++ b/packages/adapters/claude-local/src/server/quota.ts
@@ -198,10 +198,18 @@ export function toPercent(utilization: number | null | undefined): number | null
   return Math.min(100, Math.round(utilization < 1 ? utilization * 100 : utilization));
 }
 
+function defaultFetchTimeoutMs(): number {
+  const raw = process.env.PAPERCLIP_CLAUDE_LOCAL_FETCH_TIMEOUT_MS;
+  if (!raw) return 8000;
+  const parsed = Number(raw);
+  return Number.isFinite(parsed) && parsed > 0 ? parsed : 8000;
+}
+
 /** fetch with an abort-based timeout so a hanging provider api doesn't block the response indefinitely */
-export async function fetchWithTimeout(url: string, init: RequestInit, ms = 8000): Promise<Response> {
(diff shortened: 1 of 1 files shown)


serenakeyitan:
🌱 **gardener review**

**Tree alignment**:
- adapters/NODE.md: Env-configurable timeout (`PAPERCLIP_CLAUDE_LOCAL_FETCH_TIMEOUT_MS`) preserves the adapter abstraction and keeps 8000 ms as default — no behavior change for existing callers.

**Notes**: Consider adding the env var to the adapter docs alongside other tunables.

Issue #3930, opened 2026-04-18
`claude_local` agents 401 when Paperclip is launched from a Claude Desktop-spawned shell (stale `CLAUDE_CODE_OAUTH_TOKEN` inherited)


pjbroome:
# `claude_local` agents 401 when Paperclip is launched from a Claude Desktop-spawned shell (stale `CLAUDE_CODE_OAUTH_TOKEN` inherited)

## Summary

When Paperclip is launched from a terminal that Claude Desktop spawned, `claude_local` agents fail every heartbeat with:

```
Claude run failed: subtype=success: Failed to authenticate.
API Error: 401 {"type":"error","error":{"type":"authentication_error","message":"Invalid authentication credentials"}}
```

even when `claude /login` has just been run successfully in that same terminal.

## Root cause

Claude Desktop injects `CLAUDE_CODE_OAUTH_TOKEN` + `CLAUDE_CODE_PROVIDER_MANAGED_BY_HOST=1` (and related SDK vars) into the shell environment. Those are inherited by Paperclip, and `runChildProcess` in [`packages/adapter-utils/src/server-utils.ts`](https://github.com/paperclipai/paperclip/blob/master/packages/adapter-utils/src/server-utils.ts#L774-L782) currently strips only the nesting-guard vars:

```ts
const CLAUDE_CODE_NESTING_VARS = [
  "CLAUDECODE",
  "CLAUDE_CODE_ENTRYPOINT",
  "CLAUDE_CODE_SESSION",
  "CLAUDE_CODE_PARENT_SESSION",
] as const;
```

The host-auth vars survive into the spawned `claude` child, which then uses the stale Claude Desktop OAuth token (tied to Desktop's session, not the CLI session refreshed by `/login`) and is rejected by the API.

## Why "same CLI, same args, works in terminal, 401s in Paperclip"

- Direct terminal spawn: CLI honors `CLAUDE_CODE_OAUTH_TOKEN` from the live Claude Desktop session → valid.
- Paperclip-spawned child: inherits a token from an older Desktop launch or cross-session state that's no longer valid → 401.
- Re-running `claude /login` updates only the local/Keychain credentials; it does **not** update `CLAUDE_CODE_OAUTH_TOKEN` in the already-running Paperclip process, and `CLAUDE_CODE_PROVIDER_MANAGED_BY_HOST=1` tells the CLI to prefer the host-provided token over local credentials.

[...]

PR #4068, opened 2026-04-19
fix(adapter-utils): strip Claude Code OAuth/API env from spawned subprocess

rrp1977:
## Problem

When the runtime spawns the `claude` CLI as a subprocess, `CLAUDE_CODE_OAUTH_TOKEN` and `ANTHROPIC_API_KEY` are inherited from the parent's `process.env`.

If the parent was launched by a wrapper (e.g., Claude Desktop, or any shell with a previously-cached login) that exports a stale/different-account token, the child CLI prefers that env var over the fresh keychain credentials — producing **`401 Unauthorized`** on every spawn, even after a clean `claude logout && claude login`.

The failure is silent and persistent: the keychain has the correct token, but the inherited env wins.

## Reproduction

1. Launch the Paperclip server via a wrapper that exports `CLAUDE_CODE_OAUTH_TOKEN=<stale>` (common when the wrapper was itself started by an app that stored a login).
2. Run `claude logout && claude login` as a different account → keychain is refreshed.
3. Spawn any `claude_local` adapter run → 401 loop, regardless of the keychain state.

## Fix

Strip `CLAUDE_CODE_OAUTH_TOKEN` and `ANTHROPIC_API_KEY` alongside the existing `CLAUDECODE` / `CLAUDE_CODE_ENTRYPOINT` / nesting vars before spawning. CLI then falls back to keychain / session auth, which is the intended behavior.

Conservative scope: deletion happens after `process.env + opts.env` merge, so an adapter that explicitly wants to force an OAuth token through `opts.env` would also be stripped. No adapter in the tree currently does that; explicit tokens should flow via config, not env.

## Validation

- Pre-fix: 100% `status=failed exitCode=1` on `claude_local` runs (401 loop on every heartbeat).
- Post-fix: `status=succeeded exitCode=0` (verified on `v2026.416.0-canary.1`, run id `02379944`).

## Note on `codex_local`

The same env-leak pattern exists for `OPENAI_API_KEY` in the `codex_local` adapter, but the fix strategy is different there — `resolveCodexBillingType` intentionally switches to API-key billing when `OPENAI_API_KEY` is present. If this becomes a real problem for Codex users, it deserves its own PR with a more nuanced config-vs-env-priority decision.

## Files

- `packages/adapter-utils/src/server-utils.ts` (+8 lines, refactors the existing strip array into a named const and adds two vars)

### DIFF
--- a/packages/adapter-utils/src/server-utils.ts
+++ b/packages/adapter-utils/src/server-utils.ts
@@ -1085,11 +1085,19 @@ export async function runChildProcess(
     // These vars leak in when the Paperclip server itself is started from
     // within a Claude Code session (e.g. `npx paperclipai run` in a terminal
     // owned by Claude Code) or when cron inherits a contaminated shell env.
+    //
+    // Also strip CLAUDE_CODE_OAUTH_TOKEN and ANTHROPIC_API_KEY: when Claude
+    // Desktop launches the Paperclip server, it injects a baked-in OAuth token
+    // that the claude CLI prioritizes over the macOS keychain. When that token
+    // expires/rotates, subprocesses fail with 401 even though the keychain has
+    // a fresh token. Deleting these forces claude CLI to read keychain.
     const CLAUDE_CODE_NESTING_VARS = [
       "CLAUDECODE",
       "CLAUDE_CODE_ENTRYPOINT",
       "CLAUDE_CODE_SESSION",
       "CLAUDE_CODE_PARENT_SESSION",
+      "CLAUDE_CODE_OAUTH_TOKEN",
+      "ANTHROPIC_API_KEY",
     ] as const;
     for (const key of CLAUDE_CODE_NESTING_VARS) {
       delete rawMerged[key];


neoand:
+1 — independently hit #3930 today on `v2026.416.0` and this patch would have saved hours of bisecting. The conservative scope (delete after the `process.env + opts.env` merge) is exactly right because it catches both layers of the bug: the env inherited from the parent shell *and* the token that got persisted into `adapterConfig.env` at agent-creation time.

Field note that might deserve a mention in the PR body or a follow-up: users who onboarded from a poisoned shell already have the stale token stored in the DB. After this merges, the strip will stop *using* it, but `GET /api/agents/:id` will still return it in `adapterConfig.env`. Low-risk leak (no more auth damage), but noisy.

Suggested non-blocking follow-up: a one-time migration on startup that clears `adapter_config->'env'->>'CLAUDE_CODE_OAUTH_TOKEN'` and `ANTHROPIC_API_KEY` for `adapter_type='claude_local'` agents. Idempotent, zero-downtime, cleans up responses.

LGTM from a field-report perspective.

rrp1977:
Closing — I no longer use Paperclip and can't maintain this PR. Thanks for b26a7fd. The bug is still on master (#3930); paperclipai/paperclip#4606 covers the same fix.

PR #4105, opened 2026-04-20
fix: respect agent cooldown in stranded issue reconciliation

ZeroClue:
## Thinking Path

> - The heartbeat scheduler runs every 30s, checking for assigned issues with no active execution path via `reconcileStrandedAssignedIssues()`
> - When no live execution exists, it treats the issue as "stranded" and either retries or escalates to `blocked`
> - But agents have a `cooldownSec` config that creates a deliberate pause between runs — the reconciliation doesn't respect this
> - This causes a relentless blocked→unblock→blocked loop for any agent with cooldownSec > 0
> - This PR adds a cooldown grace check so reconciliation skips issues whose latest run finished within the cooldown window
> - The benefit is eliminating false-positive stranding while preserving legitimate recovery for genuinely stuck agents

## What Changed

- **`parseHeartbeatPolicy()`** — now extracts `cooldownSec` from `agent.runtimeConfig.heartbeat` (was stored but ignored)
- **`getLatestIssueRun()`** — now returns `finishedAt` timestamp (needed for recency check)
- **`reconcileStrandedAssignedIssues()`** — adds cooldown grace guard: if latest run finished within the agent's cooldown window, skip the issue instead of escalating
- The cooldown applies to **all terminal run statuses** (succeeded and failed). This is intentional: retrying a failing agent every 30s burns tokens without progress, so cooldown acts as a natural backoff.

## Verification

- All 11 existing `heartbeat-process-recovery` tests pass (no regressions)
- 2 new tests added:
  - "skips reconciliation when the agent is within its cooldown window after a recent run" — sets `cooldownSec: 300` and `finishedAt: now`, asserts no escalation
  - "allows reconciliation after the cooldown window has expired" — sets `cooldownSec: 10` and `finishedAt: 5 minutes ago`, asserts recovery proceeds
- Running in production (our instance) for 30+ minutes with confirmed results — issues that previously flipped to `blocked` every 30s now remain stable during cooldown

## Risks

- **Low risk.** The guard is purely additive — `cooldownSec: 0` (the default) skips the check entirely, preserving existing behavior exactly.
- Agents with cooldown configured will have a slightly longer window before genuine stranding is detected. For a 300s cooldown, this means up to 5 minutes delay before a genuinely stuck agent's issue is escalated. This is acceptable and preferable to the current false-positive loop.
- The failed-run cooldown behavior differs from pre-PR: a freshly-failed issue will be skipped during cooldown. This is intentional (prevents token-burning retry loops) and documented in the code comment.

## Model Used

Claude Opus 4.7 (`claude-opus-4-7`) via Claude Code CLI. Extended thinking enabled. 200K context window. Tool use enabled (file read/write, bash execution, web search).

## Checklist

### DIFF
--- a/server/src/services/heartbeat.ts
+++ b/server/src/services/heartbeat.ts
@@ -2595,6 +2595,7 @@ export function heartbeatService(db: Db) {
       intervalSec: Math.max(0, asNumber(heartbeat.intervalSec, 0)),
       wakeOnDemand: asBoolean(heartbeat.wakeOnDemand ?? heartbeat.wakeOnAssignment ?? heartbeat.wakeOnOnDemand ?? heartbeat.wakeOnAutomation, true),
       maxConcurrentRuns: normalizeMaxConcurrentRuns(heartbeat.maxConcurrentRuns),
+      cooldownSec: Math.max(0, asNumber(heartbeat.cooldownSec, 0)),
     };
   }
 
@@ -2868,6 +2869,7 @@ export function heartbeatService(db: Db) {
         error: heartbeatRuns.error,
         errorCode: heartbeatRuns.errorCode,
         contextSnapshot: heartbeatRuns.contextSnapshot,
+        finishedAt: heartbeatRuns.finishedAt,
       })
       .from(heartbeatRuns)
       .where(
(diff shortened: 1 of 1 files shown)


ZeroClue:
@reviewers Re: Greptile's point about cooldown applying to failed runs too — this is intentional.

**Why failed runs should also get cooldown grace:**

Without cooldown protection, a failing agent gets retried every 30 seconds by the reconciliation loop. If the agent is failing for a real reason (bad instructions, missing dependency, etc.), this creates a rapid retry loop that burns tokens without making progress. The cooldown acts as a natural backoff — after a failure, wait `cooldownSec` before trying again.

The max delay before a genuine stranding is detected is the configured `cooldownSec` (300s in our case). This is acceptable because:
1. The heartbeat timer will dispatch the agent again after `intervalSec` regardless
2. The cooldown only delays the *reconciliation escalation*, not the agent's natural heartbeat cycle
3. `cooldownSec: 0` (the default) preserves the existing no-grace behavior exactly

I've added a code comment documenting this design choice and two test cases covering both the grace period and the expired-cooldown recovery path.

ZeroClue:
Closing — superseded by #4223 which includes cooldown-aware skipping in `reconcileStrandedAssignedIssues()`. Thanks for the review.

PR #4151, opened 2026-04-20
feat(infra): cost telemetry fallback + startup reconciliation + worktree GC

trendywink247-afk:
Three related reliability fixes for self-hosted Paperclip deployments (single squash commit on this PR). Developed and validated on a production deployment running 10 agents under the claude-local adapter.

### 1. Cost telemetry fallback for subscription-billed runs
`packages/shared/src/pricing/anthropic.ts` (new) + `server/src/services/heartbeat.ts`

The claude-local adapter reports `reportedCostCents=0` for subscription-included runs. The heartbeat handler now computes a token-derived cost via a per-model rate table (Opus 4.7, Sonnet 4.6, Haiku 3.5) with longest-prefix matching for dated model slugs. `cost_events` and `agent_runtime_state.total_cost_cents` reflect real usage instead of zeros. The fallback only kicks in when the adapter reports 0, so paid-API runs are unaffected.

### 2. Startup reconciliation for orphaned heartbeat_runs
`server/src/services/startup-reconcile.ts` (new), wired in `server/src/index.ts` before the scheduler starts.

If the server process dies mid-run (OOM, crash, `docker restart`), the corresponding `heartbeat_runs` row stays in `running` forever and the agent's `agent_runtime_state` stays `running`, blocking the scheduler from picking up new work. On boot, we now reset any `running` row older than a 2-minute grace window (so in-flight runs from a hot reload aren't touched), and clear the corresponding `agent_runtime_state` back to `idle`. One UPDATE per table. Verified in production: server boot logged `reset 1 stuck heartbeat_runs, 1 agents`.

### 3. Stale-worktree garbage collector
`server/src/services/worktree-gc.ts` (new) + `server/src/config.ts` + wire in `server/src/index.ts`

Background job every 30 min. For each `.paperclip/worktrees/agent/*` directory:
- Only operates on branches matching `agent/*` (hard safety guard — refuses `main`, `cto/*`, `staff/*`, etc.)
- Skips if an active `execution_workspaces` row references it or any queued/running `heartbeat_runs` exist
- Skips unless `gh pr list --head <branch> --state merged` finds a merged PR
- Skips if the worktree has commits beyond `origin/<branch>` (unpushed-commit safety)
- On all-clear: `git worktree remove --force`, `git branch -D`, `git worktree prune`

Default enabled. Disable with `PAPERCLIP_WORKTREE_GC_ENABLED=false`.

## Tests

- `server/src/__tests__/startup-reconcile.test.ts` — 6 tests covering happy path, NULL `started_at`, grace window, queued-run protection, idempotency, untouched idle/failed rows.
- `server/src/__tests__/worktree-gc.test.ts` — 8 tests covering merged→removed, unmerged→skipped, active session→skipped, non-agent branch→skipped, unpushed commits→skipped, gh CLI failure, `git worktree remove` failure.

Mocks `git`, `gh`, `fs`, and the DB layer.

## Validation

- `pnpm typecheck` in `server/` — passes.
- Deployed to a production instance with 10 agents; boot log confirmed reconciliation fired on a real stuck run; manual `wakeup` smoke test produced a `cost_events` row with token-derived cents under `billingType: subscription_included`.

## Why one PR instead of three

The three changes share one concern (heartbeat pipeline reliability) and the GC wiring sits alongside the reconcile wiring in `index.ts` — splitting would produce merge conflicts on the same file for no gain. Happy to split if you'd prefer.

Co-Authored-By: Claude Opus 4.7 <noreply@anthropic.com>

### DIFF
--- a/packages/shared/package.json
+++ b/packages/shared/package.json
@@ -15,6 +15,7 @@
   "exports": {
     ".": "./src/index.ts",
     "./telemetry": "./src/telemetry/index.ts",
+    "./pricing/anthropic": "./src/pricing/anthropic.ts",
     "./*": "./src/*.ts"
   },
   "publishConfig": {
@@ -27,6 +28,10 @@
       "./telemetry": {
         "types": "./dist/telemetry/index.d.ts",
         "import": "./dist/telemetry/index.js"
+      },
+      "./pricing/anthropic": {
+        "types": "./dist/pricing/anthropic.d.ts",
+        "import": "./dist/pricing/anthropic.js"
       },
       "./*": {
         "types": "./dist/*.d.ts",
--- a/packages/shared/src/pricing/anthropic.ts
+++ b/packages/shared/src/pricing/anthropic.ts
@@ -1,0 +1,129 @@
+/**
+ * Anthropic model pricing constants used for cost telemetry.
+ *
+ * Rates are in USD per million tokens.
+ *
+ * TODO: keep these in sync with official Anthropic pricing page:
+ *   https://www.anthropic.com/pricing#anthropic-api
+ *
+ * Last verified: 2026-04-20 against public Anthropic pricing page.
+ * Update this file whenever Anthropic publishes new rates.
+ */
+
+export interface AnthropicModelRates {
+  /** USD per million input tokens (non-cached). */
+  inputPerMillion: number;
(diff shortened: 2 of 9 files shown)


trendywink247-afk:
Closing — decided to keep these as a private fork for our deployment rather than contributing upstream.

PR #4183, opened 2026-04-21
fix(ui): harden IssueChatThread against malformed transcript parts (DAT-2042)

michaeljgrimm:
## Summary

Prevent `IssueChatErrorBoundary` from falling through to the raw tool-call dump renderer when a run transcript contains non-serializable values or malformed parts. Board has been seeing inline `Tool: Bash ...` dumps in issue threads whenever the normal Chain-of-Thought renderer crashed — this fix stops hitting the boundary on valid data and polishes the fallback for the rare case where it still fires.

- `fingerprintThreadMessage` (`ui/src/lib/issue-chat-messages.ts`): wrapped `JSON.stringify` in try/catch with an id+role+length+createdAt+status fallback fingerprint so circular refs / BigInts / non-serializable tool payloads can't throw inside the `useMemo` that trips the error boundary.
- `findCoTSegmentIndex` (`ui/src/components/IssueChatThread.tsx`): wrapped in try/catch and tolerant of non-array inputs or parts without a string `type`.
- `IssueChatChainOfThought` tool parts: normalize `toolName` to `"tool"` when missing/non-string and `argsText` to `""` when non-string, so downstream `displayToolName` / `humanizeLabel` calls can't throw on malformed transcript data.
- `fallbackTextParts`: defensive against non-array `content` and non-string `text`/`argsText`/`result`. Adds a per-block `_Fallback render — open raw run for full detail._` marker and truncates tool args/result to 400 chars with a trailing `… truncated` note so the fallback card is much less noisy if it ever fires again.
- Adds a regression test that `stabilizeThreadMessages` does not throw on a message containing a circular reference.

The error boundary is intentionally not suppressed — it's still the safety net. The goal is to stop hitting it on valid run data and, when it does fire, keep the noise bounded.

## Test plan

- [x] `pnpm --filter @paperclipai/ui typecheck`
- [x] `pnpm --filter @paperclipai/ui build`
- [x] `npx vitest run ui/src/lib/issue-chat-messages.test.ts` — 14/14 passing, including the new circular-ref regression test.
- [ ] Manual: open a long CEO-run issue in the UI and confirm the normal Chain-of-Thought fold is shown instead of the amber fallback card, with no console errors. (Defer to reviewer; my workspace only has read access, no live UI.)

Tracks [DAT-2042](/DAT/issues/DAT-2042). Parent: [DAT-2041](/DAT/issues/DAT-2041).

### DIFF
--- a/ui/src/components/IssueChatThread.tsx
+++ b/ui/src/components/IssueChatThread.tsx
@@ -156,19 +156,24 @@ function findCoTSegmentIndex(
   messageParts: ReadonlyArray<{ type: string }>,
   cotParts: ReadonlyArray<{ type: string }>,
 ): number {
-  if (cotParts.length === 0) return -1;
-  const firstPart = cotParts[0];
-  let segIdx = -1;
-  let inCoT = false;
-  for (const part of messageParts) {
-    if (part.type === "reasoning" || part.type === "tool-call") {
-      if (!inCoT) { segIdx++; inCoT = true; }
-      if (part === firstPart) return segIdx;
-    } else {
-      inCoT = false;
+  try {
+    if (!Array.isArray(messageParts) || !Array.isArray(cotParts) || cotParts.length === 0) return -1;
+    const firstPart = cotParts[0];
+    let segIdx = -1;
+    let inCoT = false;
+    for (const part of messageParts) {
+      const type = (part as { type?: unknown } | null)?.type;
+      if (type === "reasoning" || type === "tool-call") {
+        if (!inCoT) { segIdx++; inCoT = true; }
+        if (part === firstPart) return segIdx;
+      } else {
+        inCoT = false;
+      }
     }
+    return -1;
+  } catch {
+    return -1;
   }
-  return -1;
 }
 
 function useLiveElapsed(startMs: number | null | undefined, active: boolean): string | null {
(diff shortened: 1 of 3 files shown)


michaeljgrimm:
CTO code review — LGTM (can't self-approve, same GitHub account).

**Pass 1 (critical): clean.**
- `fingerprintThreadMessage` — try/catch around `JSON.stringify` with id+role+partsLength+createdAt+status fallback. Circular refs no longer trip the boundary.
- `findCoTSegmentIndex` — defensive against non-array inputs and parts without a string `type`.
- `IssueChatChainOfThought` — normalizes `toolName` / `argsText` so `displayToolName` can't throw.
- `fallbackTextParts` — defensive, adds 'Fallback render — open raw run for full detail' marker and truncates args/result to 400 chars.
- New regression test (circular refs) added.

**Verified locally on checkout of commit 20e7290d:**
- `pnpm --filter @paperclipai/ui typecheck` clean
- `vitest run issue-chat-messages.test.ts` 14/14 pass

**Pass 2 (non-blocking):** fallback fingerprint could theoretically collide for two non-serializable messages differing only in content shape — acceptable because the path is rare and the boundary remains in place.

**Scope respected:** no changes to `buildIssueChatMessages`, no error-boundary removal, UI-only. Defensive patches only, as asked.

Visual regression on a long CEO-run thread (e.g. DAT-2000) still needs a human eyeball after merge — not blocking; normal `IssueChatChainOfThought` path holds in tests.

Ready to merge.

michaeljgrimm:
Closing this Data Advantage-origin PR because Data Advantage has frozen all paperclipai/paperclip product repo edits. This work is superseded by internal governance cleanup DAT-6467; no further changes should be pushed from this fork unless Paperclip maintainers explicitly re-authorize it.

PR #4198, opened 2026-04-21
adapter(claude-local): fail-loud on uncommitted work at session exit (V3)

dirk-miller:
### What changed

Closes a silent-data-loss path in `finalizeWorktree` that orphaned a downstream FreeMyMemories agent's spec on Paperclip issue `bb0ba991` (2026-04-21). Agent wrote a 22KB spec via the `Write` tool, never ran `git add`/`git commit`, exited; `finalizeWorktree`'s `hasCommits` gate evaluated false and skipped push+PR with a single stdout log line. The next agent's wake provisioned a fresh worktree from `master` and could not find the spec — flailed.

### Design boundary preserved

The adapter does NOT auto-commit. Commits are an intentional authorial act. Auto-committing would ship scratch files, half-written drafts, and content the agent explicitly chose not to commit, while breaking the `pr` skill's structured-body contract and stripping the intention signal that every downstream layer (Code Reviewer, cto-merge-review, future qmd searches) depends on.

### How

Before the existing `hasCommits` check in `finalizeWorktree`, run `git status --porcelain`. If any uncommitted/untracked files exist:

1. Log the file list to stderr.
2. Force-preserve the worktree (override caller's cleanup decision) so the next wake on this (agent, task) can recover via the agent's `pr` skill flow.
3. Post a comment on the Paperclip issue (via new `postTaskComment` helper — mirrors `fetchTaskStatus`'s curl+Bearer pattern) with: file list, worktree path, branch name, and recovery instructions for both intended-writes and discard-scratch cases. Identifies itself as the V3 fail-loud safety net.
4. Continue to existing `hasCommits` path — a session can have BOTH committed and uncommitted work; the agent's intentional commits should still push + PR.

`postTaskComment` is best-effort: API unreachable / no `taskId` / no auth logs to stderr but does not throw — worktree preservation still runs.

`finalizeWorktree` opts gains `taskId`. Both call sites pass `worktreeResult.taskId`.

### Compatibility

This patch is on top of master (which has WORKTREE_PATCH_V2's `fetchTaskStatus` + opts signature). The `freemymemories/local-customizations` branch is currently behind on V2 and would need a separate backport if both branches are in deployment.

### Test plan

- `tsc --noEmit` clean (verified locally).
- Behavioral verification on next agent session that writes vault content without committing: confirm fail-loud comment appears on the Paperclip issue.
- Existing `hasCommits` path unchanged — sessions that commit correctly continue to push + PR with identical behavior.

### Why this approach not auto-commit

See the inline V3 comment block in `finalizeWorktree` for full reasoning. Short version: agent-intentional commits + adapter fail-loud is a better boundary than adapter-auto-commit. Auto-commit failure modes include scratch shipping, scope drift via `git add -A`, sensitive content leaks, broken-state commits, and erosion of the commit-as-intention signal across the entire downstream review chain.


dirk-miller:
Closing: V3 fail-loud depends on the FreeMyMemories downstream worktree stack (WORKTREE_PATCH_V1.x) which is not on upstream paperclipai/master. This patch will land on the downstream freemymemories/local-customizations branch instead. If upstream ever adopts the per-agent worktree pattern, a standalone V3 equivalent can be proposed separately.

PR #4268, opened 2026-04-22
fix(issues): loud 401 on agent-context JWT failure at POST /comments

erginaktas:
## Summary

- **Root cause (POI-199):** When an agent posted a comment with an expired / missing JWT, the auth middleware in `local_trusted` mode silently fell back to `{type: "board", source: "local_implicit"}`, causing the comment to be attributed to `local-board` instead of the agent.
- **Server-side fix:** In `POST /api/issues/:id/comments`, detect the agent heartbeat context via the `X-Paperclip-Run-Id` header. If that header is present **and** the actor is not an agent (JWT failed), reject immediately with `401 agent_jwt_required` — no silent fallback.
- **Client-side helper:** Added `postIssueComment` / `AgentJwtError` to `packages/adapter-utils/src/server-utils.ts`. On `401 agent_jwt_required` it retries once (handles transient failures) and throws `AgentJwtError` on second failure — no silent swallow.

## Files changed

| File | What |
|------|------|
| `server/src/routes/issues.ts` | 11-line guard at top of comments POST handler |
| `server/src/__tests__/issue-comment-agent-jwt-auth.test.ts` | 3 tests (valid JWT → 201, run-id+bad JWT → 401, anon board → 201) |
| `packages/adapter-utils/src/server-utils.ts` | `postIssueComment` + `AgentJwtError` helper |
| `packages/adapter-utils/src/server-utils.test.ts` | 3 tests for the helper |

## Runtime smoke

Local `dev:watch` was used to validate the 401 path:

```bash
# Expired/missing JWT with run-id header → 401
curl -sS -X POST http://127.0.0.1:3100/api/issues/POI-238/comments \
  -H "Authorization: Bearer invalid.jwt.token" \
  -H "X-Paperclip-Run-Id: run-test-abc" \
  -d '{"body":"test"}'
# → {"error":"agent_jwt_required","reason":"missing_or_expired"}

# No run-id header, board session → 200 local-board (unchanged)
curl -sS -X POST http://127.0.0.1:3100/api/issues/POI-238/comments \
  -d '{"body":"board comment"}'
# → comment attributed to local-board as before
```

## Test plan

- [ ] All 3 server-side tests green
- [ ] All 3 adapter-utils tests green
- [ ] Manual smoke confirms 401 on expired JWT + run-id header
- [ ] Board user posting without run-id header still gets 200

## Scope

Per issue spec: JWT minter / claim schema NOT changed. Other endpoints NOT affected. Existing `local-board` attribution for non-agent contexts preserved.

Closes POI-238 (parent: POI-199)

### DIFF
--- a/packages/adapter-utils/src/server-utils.test.ts
+++ b/packages/adapter-utils/src/server-utils.test.ts
@@ -1,5 +1,5 @@
 import { randomUUID } from "node:crypto";
-import { describe, expect, it } from "vitest";
+import { describe, expect, it, vi } from "vitest";
 import {
   appendWithByteCap,
   DEFAULT_PAPERCLIP_AGENT_PROMPT_TEMPLATE,
@@ -7,6 +7,8 @@ import {
   runningProcesses,
   runChildProcess,
   stringifyPaperclipWakePayload,
+  postIssueComment,
+  AgentJwtError,
 } from "./server-utils.js";
 
 function isPidAlive(pid: number) {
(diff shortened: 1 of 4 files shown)


omar6995:
Adding a real-world data point that supports landing this PR and suggests a
  small extension.

  We hit the comment-misattribution failure on a `local_trusted` production
  instance. A `claude_local` CEO agent posted a comment on PRO-284 that landed
  with `authorUserId='local-board'`, `authorAgentId=NULL`, `createdByRunId=NULL`.
  The next heartbeat fired `issue_commented` for the agent's own comment, and
  the agent idled through ~7 wake events trying to determine whether a human
  had actually replied. End-user effect: the board UI shows the comment under
  the human's name, the agent believes it's waiting on a human, and the issue
  stalls.

  What's interesting for this PR: the failing request from our access log
  carried **neither** `Authorization` **nor** `X-Paperclip-Run-Id`:

  POST /api/issues/PRO-284/comments   (15:07:47, 201 Created)
  headers: { host, user-agent: curl/8.14.1, accept, content-type, content-length }

  The current guard (`runIdHeader && actor.actorType !== "agent"`) doesn't trip
  because the request has no run-id either. It still ends up attributed to
  `local-board` because the middleware default in `local_trusted` is the implicit
  board actor (`auth.ts:24-33`).

  Two thoughts:

  1. **Land this as-is** — it's strictly an improvement and catches the more
     common case (agent injects run-id but has expired/missing JWT). LGTM.
  2. **Tiny extension worth considering** here or in a follow-up: when a mutating
     issue handler is reached with `actor.source === "local_implicit"` *and* no
     session/cookie signal of being the local UI, reject the same way. That
     covers the dropped-both-headers path we observed without changing any UI
     behavior.

  Cross-link: #4520 fixes the explicit-invalid-bearer path in the middleware;
  #4441 recovers attribution via runId at the route layer. The failure mode we
  hit is degenerate w.r.t. all three. Worth coordinating so we end up with full
  coverage instead of three near-misses.

PR #4275, opened 2026-04-22
fix(heartbeat): deferred-wake freshness guard to stop terminal drift (POI-237)

erginaktas:
## Thinking Path

> - Paperclip's deferred wake promotion flow must decide whether a terminal issue should be reopened based on whether a deferred comment still represents fresh user intent.
> - The affected subsystem is the heartbeat promotion path in `server/src/services/heartbeat.ts`, specifically the deferred comment-wake freshness guard added for POI-237.
> - That guard compares deferred comment timestamps against the issue's most recent terminal transition so stale queued wakes do not reopen already-resolved issues.
> - A new edge case appears when the wake still references comment IDs, but those comments have been deleted after enqueue.
> - In that situation the previous implementation treated an empty `commentRows` result as "all stale", which could silently suppress a valid reopen purely because the backing comments were cleaned up later.
> - This pull request makes the guard lenient in that deleted-comment case, matching the existing no-audit fallback philosophy.
> - The benefit is that comment cleanup can no longer produce a false-negative stale classification for an otherwise valid deferred wake.

## What Changed

- Updated the deferred-wake freshness guard in `server/src/services/heartbeat.ts` so an empty `commentRows` result is treated leniently instead of marking the wake stale.
- Added an embedded-Postgres regression test covering the deleted-comment case: a fresh deferred wake still reopens the issue even after the referenced comment row is removed.
- Kept the original stale/fresh/mixed/no-audit/non-comment behavior intact.

## Verification

- `pnpm exec vitest run server/src/__tests__/heartbeat-deferred-wake-freshness.test.ts`
- Covered scenarios now include:
  - stale deferred comment → dropped
  - fresh deferred comment → reopened
  - mixed stale/fresh batch → reopened
  - no terminal audit entry → lenient reopen
  - deleted referenced comment row → lenient reopen
  - non-comment deferred wake → bypass freshness guard

## Risks

- Low risk: the change only affects the terminal-issue deferred comment freshness guard and only in the empty-result edge case.
- The main tradeoff is intentional leniency: a deleted referenced comment will now behave like the existing no-audit fallback, preferring not to suppress a potentially legitimate reopen.

> Checked `ROADMAP.md`; this is a narrow bug fix in existing heartbeat behavior, not overlapping planned core feature work.

## Model Used

- Original PR authoring: Anthropic Claude Code (exact model revision not preserved in the PR metadata available to this follow-up).
- Review-fix follow-up for the deleted-comment edge case: OpenAI Codex desktop using GPT-5-family reasoning with local repo editing, shell/test execution, and focused embedded-Postgres regression coverage.

## Checklist

- [x] I have specified the model used (with the exact details available from the original authoring session)
- [x] I have updated relevant documentation to reflect this change where needed

### DIFF
--- a/server/src/__tests__/heartbeat-deferred-wake-freshness.test.ts
+++ b/server/src/__tests__/heartbeat-deferred-wake-freshness.test.ts
@@ -1,5 +1,5 @@
 import { randomUUID } from "node:crypto";
-import { and, eq } from "drizzle-orm";
+import { and, eq, sql } from "drizzle-orm";
 import { afterAll, afterEach, beforeAll, describe, expect, it, vi } from "vitest";
 import {
   activityLog,
@@ -8,7 +8,6 @@ import {
   agentWakeupRequests,
   companies,
   createDb,
-  heartbeatRunEvents,
   heartbeatRuns,
   issueComments,
   issues,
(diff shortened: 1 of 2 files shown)


erginaktas:
## QA Review — APPROVED ✅

> Self-approval blocked by GitHub (same account). Posting as comment per QA workflow.

**All 5 test cases verified against the POI-240 plan cases:**

| Case | Scenario | Guard path | Result |
|------|----------|------------|--------|
| 1 | Comment predates close | `anyFresh = false` → drop | ✅ PASS |
| 2 | Comment is newer than close | `anyFresh = true` → reopen | ✅ PASS |
| 3 | Mixed batch (one stale, one fresh) | `anyFresh = true` → reopen | ✅ PASS |
| 4 | No terminal activity-log entry | `closedAt = null` → lenient fallback | ✅ PASS |
| 5 | Non-comment deferred wake | Guard bypassed entirely | ✅ PASS |

**Security:** No hardcoded secrets, all queries parameterized via Drizzle ORM, credential scan on PR metadata clean.

**Regression risk:** Low — guard is gated behind `deferredCommentIds.length > 0 && issueIsTerminal`; non-terminal issues and non-comment wakes are completely unaffected.

---

### Advisory notes (non-blocking, follow-up items)

**[WARNING] Unresolved comment IDs edge case (no test):** If all `deferredCommentIds` in the payload reference comments that no longer exist in `issueComments` (deleted after the wake was enqueued), `commentRows` returns empty, `anyFresh = false`, and the wake is silently dropped — even though `closedAt` is valid. The lenient fallback applies only to `closedAt = null`. Whether this conservative behavior is intentional should be clarified in a follow-up. A single test case would close the gap.

**[WARNING] afterEach idle-poll teardown:** The `idlePolls >= 3` consecutive-idle pattern has a narrow timing hazard — a briefly idle `void executeRun()` could register as idle prematurely, letting `TRUNCATE CASCADE` race with a later DB write. Not reproducible in current runs; does not affect production code.

---

QA approved — ready for TechLead handoff to human review.

PR #4288, opened 2026-04-22
feat(heartbeat): skip timer wakes when agent has no actionable issues

Vasanth19:
## Problem

Every timer heartbeat unconditionally invokes the model, even when the agent has no assigned work. In a 14-agent setup (~90% idle at any given tick), this produces ~260 wasted model invocations per day and was the #1 token cost driver.

## Solution

Add a pre-check in `tickTimers()` that queries for at least one actionable issue (`todo` / `in_progress` / `in_review`) assigned to the agent before enqueuing the wake. When none exist:
- `lastHeartbeatAt` is updated so the interval resets normally (avoids thundering-herd on next tick)
- The run is counted as `skipped` in the tick result
- No adapter invocation, no tokens burned

The check is **fail-open**: if the DB query fails, the heartbeat proceeds as before.

## Changes

**`server/src/services/heartbeat.ts`** — `tickTimers()`:
```typescript
const actionableCount = await db
  .select({ id: issues.id })
  .from(issues)
  .where(and(
    eq(issues.companyId, agent.companyId),
    eq(issues.assigneeAgentId, agent.id),
    inArray(issues.status, ["todo", "in_progress", "in_review"]),
  ))
  .limit(1)
  .then((rows) => rows.length);

if (actionableCount === 0) {
  await db.update(agents).set({ lastHeartbeatAt: now }).where(eq(agents.id, agent.id));
  skipped += 1;
  continue;
}
```

**`server/src/services/issues.ts`** — resolves a merge conflict between the VAS-27 transactional execution-lock clearing (stash) and the upstream dependency-blocker readiness check. Both are preserved inside the same transaction.

## Measured impact (14 agents, production)

| Metric | Before | After |
|--------|--------|-------|
| Model invocations/day | ~288 | ~29 (est.) |
| Wasted timer wakes | ~90% of total | 0 |
| Input tokens saved/day | — | ~130K–520K |
| DB overhead per skipped tick | — | 1 indexed query (~5ms) |

The query hits `issues_company_assignee_status_idx` (`companyId`, `assigneeAgentId`, `status`) with `LIMIT 1` — effectively a single index seek.

## Related

Complements `isolatedSession: true` + `lightContext: true` per-agent config (eliminates session context bleed on wakes that do fire).

### DIFF
--- a/server/src/services/heartbeat.ts
+++ b/server/src/services/heartbeat.ts
@@ -7292,6 +7292,30 @@ export function heartbeatService(db: Db) {
         const elapsedMs = now.getTime() - baseline;
         if (elapsedMs < policy.intervalSec * 1000) continue;
 
+        // Skip timer wake if the agent has no actionable issues assigned.
+        // Advances lastHeartbeatAt so the interval resets — prevents immediate re-check next tick.
+        const actionableCount = await db
+          .select({ id: issues.id })
+          .from(issues)
+          .where(
+            and(
+              eq(issues.companyId, agent.companyId),
+              eq(issues.assigneeAgentId, agent.id),
+              inArray(issues.status, ["todo", "in_progress", "in_review"]),
+            ),
+          )
+          .limit(1)
+          .then((rows) => rows.length);
+
+        if (actionableCount === 0) {
+          await db
+            .update(agents)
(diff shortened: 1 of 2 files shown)


Vasanth19:
Closing — keeping changes local only.

PR #4329, opened 2026-04-23
fix(server): use pathToFileURL for dynamic plugin imports on Windows

thedelph:
## Thinking Path

> - The plugin system lets users install adapter and server plugins from a local filesystem path via `POST /api/plugins/install` with `isLocalPath: true`
> - The two plugin loaders (`server/src/services/plugin-loader.ts` for manifests, `server/src/adapters/plugin-loader.ts` for external adapter packages and runtime reload) dynamically `import()` JS files off disk — they either pass the raw path or concatenate `` `file://${modulePath}` ``
> - On Windows both forms are invalid for the Node ESM loader: a raw `C:\Users\...` path fails with `ERR_UNSUPPORTED_ESM_URL_SCHEME`, and `file://C:\Users\...` is not a valid file URL (backslashes aren't URL-safe and the drive letter has no slash). This silently breaks plugin install and adapter hot-reload on Windows hosts
> - This pull request converts each path to a proper `file://` URL with `url.pathToFileURL(...).href` before calling `import()`, which is the Node-documented way to build a file URL from an OS path
> - The benefit is that `POST /api/plugins/install` with `isLocalPath: true` and the runtime adapter reload both work on Windows without downstream hacks (the only current workaround is patching the published `_npx` cache), while remaining a no-op on macOS/Linux where `pathToFileURL` returns the equivalent of the previous string

## What Changed

- `server/src/services/plugin-loader.ts` — import `pathToFileURL` from `node:url`; wrap the dynamic `import(manifestPath)` call with `pathToFileURL(manifestPath).href` so manifest loading accepts Windows drive paths.
- `server/src/adapters/plugin-loader.ts` — import `pathToFileURL` from `node:url`; wrap `import(modulePath)` in `loadExternalAdapterPackage` with `pathToFileURL(modulePath).href`; replace the runtime-reload `` `file://${modulePath}` `` concatenation in the reload path with `pathToFileURL(modulePath).href` so both first-load and hot-reload are consistent.

Diff: 2 files, +14 / −5. TypeScript source only — no changes to built `dist/` artifacts, generated code, or config.

## Verification

Reproduction (Windows, current `master`):

1. On Windows, clone a local plugin package at e.g. `C:\Users\you\src\paperclip-plugin-foo`.
2. `POST /api/plugins/install` with `{ "isLocalPath": true, "path": "C:\Users\you\src\paperclip-plugin-foo" }` against a local Paperclip server.
3. Without this fix: server logs `ERR_UNSUPPORTED_ESM_URL_SCHEME` from `loadManifestFromPath` (or `loadExternalAdapterPackage` for adapter plugins); install fails. Runtime adapter reload hits the same error from the `` `file://${modulePath}` `` path.

After the fix:

- `POST /api/plugins/install` with a plain Windows path returns 2xx and the plugin is registered.
- Runtime adapter reload (triggered by editing the adapter's JS entry) re-imports the fresh module without a URL-scheme error.

Repo checks:

- `pnpm -r typecheck` passes on the server package (no new type surface — `pathToFileURL` is the standard `node:url` export).
- Existing `server` tests pass locally (`pnpm --filter @paperclipai/server test`). No new test added because this is a host-OS-dependent Node ESM behavior; unit-testing `import()` resolution would require either a Windows CI runner or mocking the ESM loader, neither of which this repo sets up today.
- No-op on macOS/Linux: `pathToFileURL('/abs/path/to/mod.js').href` → `file:///abs/path/to/mod.js`, which the existing loader already accepts.

## Risks

**Low risk.** `pathToFileURL` is a stable `node:url` primitive (documented precisely for this case) and already used elsewhere in the ecosystem for the same problem. The behavior change is:

- On Windows: previously-broken dynamic `import()` now succeeds (the intended fix).
- On macOS/Linux: the URL produced by `pathToFileURL` is equivalent to what the old code produced (`file:///abs/path`), so behavior is unchanged.
- No API surface change. No runtime-format change for stored plugin records. No migration.

## Model Used

- Claude Opus 4.7 (`claude-opus-4-7`, 1M context), via Claude Code CLI running as a Paperclip agent. Extended thinking enabled; tool use (file reads, shell, `gh`) used to apply the patch, push the fork branch, and open this PR.

## Checklist

- [ ] I have added or updated tests where applicable — see Verification; no host-OS-dependent test harness exists for dynamic `import()` resolution. Happy to add one if a reviewer points at an existing pattern.
- [x] If this change affects the UI, I have included before/after screenshots — N/A, server-side only
- [x] I have updated relevant documentation to reflect my changes — commit message documents the behavior; no user-facing docs currently describe the Windows path

### DIFF
--- a/server/src/adapters/plugin-loader.ts
+++ b/server/src/adapters/plugin-loader.ts
@@ -11,6 +11,7 @@
 
 import fs from "node:fs";
 import path from "node:path";
+import { pathToFileURL } from "node:url";
 import type { ServerAdapterModule } from "./types.js";
 import { logger } from "../middleware/logger.js";
 
@@ -177,7 +178,7 @@ export async function loadExternalAdapterPackage(
 
   logger.info({ packageName, packageDir, entryPoint, modulePath, hasUiParser: !!uiParserSource }, "Loading external adapter package");
 
-  const mod = await import(modulePath);
+  const mod = await import(pathToFileURL(modulePath).href);
   const adapterModule = validateAdapterModule(mod, packageName);
 
   if (uiParserSource) {
@@ -212,7 +213,7 @@ export async function reloadExternalAdapter(
   const packageDir = resolvePackageDir(record);
   const entryPoint = resolvePackageEntryPoint(packageDir);
   const modulePath = path.resolve(packageDir, entryPoint);
-  const fileUrl = `file://${modulePath}`;
+  const fileUrl = pathToFileURL(modulePath).href;
 
   // Bust ESM module cache so re-import loads fresh code from disk.
   // Query-string trick (?t=...) works in Node; Bun may need the file:// URL


thedelph:
Gentle review ping — this PR has been open for a week with all CI green (policy / verify / e2e / Snyk / Greptile 5/5) and no human review yet. Mergeable=clean against current `master`.

Scope is intentionally tiny: 14 lines across two files, swapping bare `import(modulePath)` and the manual `` `file://${modulePath}` `` concat for `pathToFileURL(modulePath).href` in `server/src/{services,adapters}/plugin-loader.ts`. No-op on macOS/Linux; unblocks Windows plugin install which currently fails because the ESM loader rejects `C:\…` paths and `file://C:\…\` URLs.

Happy to rebase, split, or address any feedback — would just appreciate a look. cc @cryppadotta @devinfoley as CODEOWNERS for adjacent server paths.

thedelph:
## Rebased + narrowed scope — now just `adapters/plugin-loader.ts`

Force-pushed `93644e1` → `7ea0ad8`, rebased onto current `master`. The PR is now a **1-file, +3/-2** consistency fix.

### Why the scope shrank

The `services/plugin-loader.ts` half of the original change is no longer needed: @cryppadotta's #5205 ("Expand plugin host surface", merged 2026-05-05) independently added `pathToFileURL` to that file — same import, same wrap pattern. Dropping that hunk avoids any incidental churn now that master already matches.

### What's left

Two call sites in `server/src/adapters/plugin-loader.ts` that #5205 did **not** touch and that are still broken on Windows:

- `loadExternalAdapterPackage` — `await import(modulePath)` with a bare `C:\…` path; ESM loader rejects it because it isn't a valid URL.
- `reloadExternalAdapter` — `` `file://${modulePath}` `` via string concat; on Windows that produces `file://C:\…` with backslashes, which the ESM loader also rejects (same Bun cache-bust path keeps working once `fileUrl` is built via `pathToFileURL`).

Both now go through `pathToFileURL(modulePath).href`, matching the pattern #5205 already established in the sibling file. No-op on macOS/Linux.

### Verified

- Mergeable / clean against `master` @ `ad0bb57`
- Greptile previously scored the broader change 5/5; this is a strict subset
- Tracking confirmation: master `services/plugin-loader.ts` line 32 has `import { fileURLToPath, pathToFileURL } from "node:url";` (from #5205); `adapters/plugin-loader.ts` line 13 still imports only `path` and lines 180 / 215 still have the bare-path + string-concat patterns this PR removes.

CI re-running on the new head. Happy to address any further feedback. cc @cryppadotta @devinfoley.

PR #4331, opened 2026-04-23
fix(pi-local): prepend installed skill bin/ dirs to child PATH

SgtPooki:
## Thinking Path

> - Paperclip orchestrates AI agents; each agent runs under an adapter that spawns a model CLI as a child process.
> - The pi-local adapter (`packages/adapters/pi-local`) spawns `pi` and inherits the child's shell environment — including `PATH`, which determines what the child's bash tool can execute by name.
> - Paperclip skills ship executable helpers under `<skill>/bin/` (e.g. `paperclip-get-issue`) and Reviewer/QA-style `AGENTS.md` files invoke them by name via the agent's bash tool.
> - Pi-local builds its runtime env with `ensurePathInEnv({ ...process.env, ...env })` only — it never adds the installed skills' `bin/` dirs to PATH. The pi CLI's `--skill` arg loads each skill's SKILL.md but does not augment PATH.
> - Consequence: every bash invocation of a skill helper fails with `exit 127: command not found`. The agent then spends its heartbeat guessing (re-reading SKILL.md, trying `find`, inventing command paths) and either times out or gives up.
> - This PR prepends each injected skill's `bin/` directory to the child PATH immediately before runtimeEnv is constructed.
> - The benefit: pi_local agents whose AGENTS.md uses any `paperclip-*` skill helper can actually run those helpers.

## What Changed

- `packages/adapters/pi-local/src/server/execute.ts`: compute `skillBinDirs` from the already-resolved `piSkillEntries`, dedupe against the existing PATH, prepend them to whichever of `PATH` / `Path` the merged env uses, then build `runtimeEnv`. No new helpers, no adapter-utils changes.

## Verification

Manual repro before the fix:

1. Create a pi_local agent wired to a paperclip skill (e.g. paperclip-control).
2. Wake the agent on an in_review issue with an AGENTS.md that starts with `paperclip-get-issue "$PAPERCLIP_TASK_ID"`.
3. Session file: `{ "role": "toolResult", "isError": true, "content": [{ "text": "/bin/bash: paperclip-get-issue: command not found\n\nCommand exited with code 127" }] }`.

After the fix: same wake; `paperclip-get-issue` resolves and returns the issue JSON; agent proceeds.

Local commands:

```
pnpm --filter @paperclipai/adapter-pi-local typecheck   # clean
pnpm --filter @paperclipai/adapter-pi-local build       # clean
  src/__tests__/pi-local-execute.test.ts \
  src/__tests__/pi-local-adapter-environment.test.ts \
  src/__tests__/pi-local-skill-sync.test.ts
# 5/5 passing
```

No new tests: the existing `pi-local-skill-sync.test.ts` covers skill symlink injection (upstream of the PATH step), and `pi-local-execute.test.ts` covers the spawn path; this change only augments env on the same spawn path.

## Risks

Low. Pure PATH augmentation on the child env. Edge cases:

- Zero skills installed → no PATH change (guarded by `skillBinDirs.length > 0`).
- Duplicate bin dirs already on PATH → deduped; no pollution on re-runs.
- Windows `Path` casing → falls back correctly when merged env uses `Path` instead of `PATH`.
- Skill dir without `bin/` subdir → joined path simply won't resolve; harmless.

No behavioral change for pi_local agents that don't use skill-provided commands.

## Model Used

- Claude, `claude-opus-4-7` (1M context), extended thinking enabled, tool use enabled. Walked pi-local/cursor-local/claude-local and adapter-utils to isolate the gap, wrote the inlined fix, and ran typecheck/build/test locally.

## Checklist

- [x] Thinking path from project context to this change
- [x] Model used specified
- [x] Checked ROADMAP.md — no overlap
- [x] Tests run locally, passing
- [x] Tests added — new case in `server/src/__tests__/pi-local-execute.test.ts`; verified it fails when the fix is reverted
- [ ] UI screenshots — N/A (backend adapter change)
- [x] Docs updated — N/A (internal adapter, no user-facing docs)
- [x] Risks documented
- [x] Will address reviewer comments before merge

### DIFF
--- a/packages/adapters/pi-local/src/server/execute.ts
+++ b/packages/adapters/pi-local/src/server/execute.ts
@@ -207,10 +207,13 @@ export async function execute(ctx: AdapterExecutionContext): Promise<AdapterExec
   // Prepend installed skill `bin/` dirs to PATH so an agent's bash tool can
   // invoke skill binaries (e.g. `paperclip-get-issue`) by name. Without this,
   // any pi_local agent whose AGENTS.md calls a skill command via bash hits
-  // exit 127 "command not found".
+  // exit 127 "command not found". Only include skills that ensurePiSkillsInjected
+  // actually linked — otherwise non-injected skills' binaries would be reachable
+  // to the agent.
+  const injectedSkillKeys = new Set(desiredPiSkillNames);
   const skillBinDirs = piSkillEntries
-    .map((entry) => path.join(entry.source, "bin"))
-    .filter((binDir) => binDir.length > 0);
+    .filter((entry) => injectedSkillKeys.has(entry.key) && entry.source.length > 0)
+    .map((entry) => path.join(entry.source, "bin"));
   const mergedEnv = ensurePathInEnv({ ...process.env, ...env });
   const pathKey =
     typeof mergedEnv.Path === "string" && mergedEnv.Path.length > 0 && !mergedEnv.PATH
--- a/packages/adapters/pi-local/src/server/execute.ts
+++ b/packages/adapters/pi-local/src/server/execute.ts
@@ -204,8 +204,28 @@ export async function execute(ctx: AdapterExecutionContext): Promise<AdapterExec
     env.PAPERCLIP_API_KEY = authToken;
   }
   
+  // Prepend installed skill `bin/` dirs to PATH so an agent's bash tool can
+  // invoke skill binaries (e.g. `paperclip-get-issue`) by name. Without this,
+  // any pi_local agent whose AGENTS.md calls a skill command via bash hits
+  // exit 127 "command not found".
+  const skillBinDirs = piSkillEntries
+    .map((entry) => path.join(entry.source, "bin"))
+    .filter((binDir) => binDir.length > 0);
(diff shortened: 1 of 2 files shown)


SgtPooki:
Added a test covering the fix:  (1880eb56). Configures a synthetic skill with a bin/ subdir, spawns a fake pi that dumps `process.env.PATH`, asserts the skill's bin dir is first on PATH and appears exactly once. Also verified manually by reverting the fix — the test fails without it.

SgtPooki:
Addressed both Greptile findings in 90e1b2dc:

- **P1 (desired-subset filter)**: `skillBinDirs` now filters `piSkillEntries` by `desiredPiSkillNames` before projecting to `bin/` paths. PATH augmentation is consistent with `ensurePiSkillsInjected`; non-injected skills' binaries are no longer reachable.
- **No-op guard**: moved the emptiness check onto `entry.source` before the `path.join`, as suggested.

Added a second test case (`does not expose bin/ dirs from skills that are not injected`) that marks a synthetic skill `required: false` with no explicit sync preference and asserts its `bin/` dir is NOT on PATH. Full pi-local suite is 7/7 passing.

PR #4347, opened 2026-04-23
fix(routines): Fallback C — coalesce against stale open issues with no executionRunId

donkassim:
When executionRunId is NULL (the Hermes adapter path), Path A and B both miss the existing open issue. Fallback C broadens the match to any open issue by (originKind, originId, fingerprint), regardless of whether a running heartbeat exists. Prevents duplicate routine execution issues. Includes test case.


rileymiller:
Red-CI stale sweep ruling (2026-04-24 13:15 MDT): this PR has failed checks older than 12 hours and is covered by the same-run queue-health ruling.

- Owner ruling: explicit board blocker via SIG-631
- Board owner: Cleo
- Path: split to Forge/additional owners for fix-now triage, or approve close/supersede if stale
- Current startable Forge slice remains SIG-628 while SIG-629 waits for admin rerun or close/supersede decision
- Next check: 2026-04-24 14:30 MDT

rileymiller:
Forge SIG-634 triage (2026-04-24 13:38 MDT): owner update on stale-red verify.

- Failing check: `verify` on run https://github.com/paperclipai/paperclip/actions/runs/24854337243/job/72763067808
- Root cause: Fallback C changes live-execution coalescing semantics. The existing idle-open-issue test now receives an active issue where it expected `detailBefore.activeIssue` to be null, and the new Fallback C test also throws `TypeError: Cannot destructure property 'labelIds' of 'data' as it is undefined` from `issueSvc.create`.
- Action: no rerun; this is deterministic implementation/test failure.
- Recommended path: tighten Fallback C so only the intended stale no-`executionRunId` routine issue coalesces, update the old idle-open expectation accordingly, and fix the new test helper call/assertion. The head branch is named `master` against `master`; if the author cannot safely amend it, supersede with a named clean branch.

adlai88:
+1 — we hit this exact bug at scale on a self-hosted Paperclip running the default Postgres backend. Sharing the symptom + our local mitigation in case it helps push this PR forward.

**Symptom**

A `coalesce_if_active` routine assigned to a Claude Code (`claude_local`) adapter accumulated **435 stuck `routine_execution` issues in `todo` over 12 days**, ~36/day average, oldest from 12 days before discovery. Same routine had ~140 wakes that did get claimed (so the agent itself was alive — it just didn't claim every wake). Symptom matches #4369's "permanent deadlock" framing exactly.

Tracing it landed on the same root cause this PR addresses: `findLiveExecutionIssue` requires a heartbeat in `LIVE_HEARTBEAT_RUN_STATUSES`. When the heartbeat completes without the agent finishing the wake (race, claim error, adapter quirk), the issue's heartbeat moves out of those statuses → next dispatch sees no "live" issue → creates a new one → orphan accumulates. Repeat ~36×/day.

**A complementary fix worth considering**

We patched our fork at the **dispatch path** instead of the read path — cancel any orphan `todo`/`backlog` routine_execution issues for the routine before `findLiveExecutionIssue` runs:

```ts
await txDb.execute(sql`
  UPDATE issues
  SET status = 'cancelled', cancelled_at = ${triggeredAt}
  WHERE company_id = ${input.routine.companyId}
    AND origin_kind = 'routine_execution'
    AND origin_id = ${input.routine.id}
    AND origin_fingerprint = ${dispatchFingerprint}
    AND status IN ('todo', 'backlog')
    AND hidden_at IS NULL
    AND NOT EXISTS (
      SELECT 1 FROM heartbeat_runs hr
      WHERE (hr.id = issues.execution_run_id
             OR hr.context_snapshot ->> 'issueId' = issues.id::text)
        AND hr.status IN ('queued', 'running', 'scheduled_retry')
    )
`);
```

Different design tradeoff:

| | This PR (#4347 — Fallback C) | Dispatch-time orphan-GC |
|---|---|---|
| Orphan exists, agent never claims | Coalesces every future fire into the same orphan; one stuck wake forever | Each cycle cancels the orphan + creates fresh; eventually a healthy fire gets claimed |
| Race (two fires near-simultaneously) | Second coalesces cleanly into first | Second cancels first's orphan + creates new (small redundancy if first was about to be claimed) |
| Work staleness | Agent reviews state-as-of-the-original-wake | Agent reviews state-as-of-now |
| Diff size | ~64 lines (read path) | ~25 lines (write path) |
| Risk profile | Pure read-side, low risk | Modifies prior state, slightly more invasive |

**Which is right is routine-dependent.** For "do work as of NOW" routines (code review, triage, snapshot generation, anything where a 5-day-old wake context is wrong), orphan-GC is more useful. For order-sensitive "process THIS specific event" routines, this PR's coalesce-into-orphan is more correct.

The two approaches actually compose: orphan-GC at dispatch time + Fallback C as defensive read = belt-and-suspenders. We're running both as of today.

Happy to open our diff as a separate PR if useful, or we can keep it as a fork-side hardening if maintainers feel coalesce-into-orphan is the right default. Either way, the symptom data is real and points at this PR being the right baseline fix to land first.

PR #4350, opened 2026-04-23
chore(tests): reset module-level singletons between it() blocks

VetSecItPro:
## Summary

Service modules hold \`let\` counters, Maps, and EventEmitter instances at module scope. Within a single test file's run, those singletons carry state from one \`it()\` block into the next — an \`eventId\` counter that was 0 in test A is 1 in test B, producing order-dependent failures that look like intermittent flakes.

This PR adds an opt-in reset convention so any module with singleton state can participate.

## How it works

- A module with singleton state exports \`_resetSingletonsForTest()\` (plain named export, no runtime gating — callers decide when).
- \`server/src/__tests__/helpers/setup-singletons.ts\` registers a global \`beforeEach\` that dynamically imports each known-resettable module and calls the reset. \`.catch(() => undefined)\` keeps it resilient: missing modules are silently skipped rather than breaking the whole suite.
- \`server/vitest.config.ts\` wires the setup file via \`setupFiles\`.

## Two initial participants

Both had obvious module-level state:

- \`services/live-events.ts\` — \`nextEventId\` counter + EventEmitter
- \`services/run-log-store.ts\` — \`cachedStore\` lazy singleton

## Proof test

\`singleton-reset.test.ts\` encodes the invariant: test A mutates state, test B asserts the mutation is gone. If the setup hook is ever removed or broken, the proof test fails in a deterministic way — giving an unambiguous signal that singleton pollution is back.

## Adding a new participant

1. Export \`_resetSingletonsForTest()\` from the module
2. Add an entry to the \`resets\` array in \`setup-singletons.ts\`
3. Optional: extend \`singleton-reset.test.ts\` with a describe block for that module's specific state shape

## Test plan

- [x] Proof test passes 4/4
- [x] No behavior change outside test context (reset fn is only called from the setup file)
- [ ] Upstream CI

## Scope kept narrow

Intentionally shipping only the pattern + two participants. Subsequent PRs can extend to other services as their flake signatures suggest. Also not included: a typed \`makeFullServicesMock()\` helper (related but separable; that's a different class of test-pollution remediation).

### DIFF
--- a/server/src/__tests__/helpers/setup-singletons.ts
+++ b/server/src/__tests__/helpers/setup-singletons.ts
@@ -1,0 +1,32 @@
+import { beforeEach } from "vitest";
+
+/**
+ * Vitest setup file: reset mutable module-level singletons between test cases.
+ *
+ * Why this matters: service modules that hold `let` counters, Maps, or
+ * EventEmitter instances at module scope carry state from test A into
+ * test B within the same file run. Execution-order-dependent failures
+ * follow — a counter that was 0 when A ran is 1 when B runs, and the test
+ * starts producing different output on every invocation.
+ *
+ * Approach: opt-in reset. Each module with singleton state exports a named
+ * `_resetSingletonsForTest()` function. This setup file imports those
+ * modules dynamically inside a global `beforeEach` and calls each reset.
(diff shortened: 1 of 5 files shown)


VetSecItPro:
Closing — not contributing upstream after all.

PR #4352, opened 2026-04-23
feat(preflight): detect missing workspace node_modules before pnpm dev (AIU-502)

kbecking:
## Summary

Fixes the AIU-491 black-screen-without-signal class of issues: after an interrupted or partial `pnpm install`, pnpm's content store can hold a package while the workspace symlink under `<workspace>/node_modules/<name>` is missing. Vite's import-analysis then 404s on the dep, the UI renders black, and `/api/health` still returns `ok` because no server-side module is affected.

This PR extends the existing `ensure-workspace-package-links.ts` preflight with a second pass that refuses to start dev if any declared external dep has no materialized `node_modules` entry, and wires the preflight into `pnpm dev` (previously only `build`, `typecheck`, and `test:*` ran it).

## Changes

- **`scripts/ensure-workspace-package-links.ts`** — second pass over every workspace (including repo root). For each `dependencies` + `devDependencies` entry that is not `workspace:` or `link:`, verify `<workspace>/node_modules/<name>` exists via `statSync` (follows symlinks, so dangling links also count as missing). If any are missing, print the per-workspace list and exit 1 with `pnpm install --frozen-lockfile` as the recovery step.
- **`scripts/ensure-workspace-package-links.ts`** — drop the `dev-service-profile.ts` import. That module transitively pulls the full `local-service-supervisor` graph, adding ~1.2s of cold start. Inlining `repoRoot` from `import.meta.url` gets us back under 500ms on a healthy workspace.
- **`package.json`** — `dev`, `dev:watch`, `dev:once` now run `pnpm run preflight:workspace-links` first, matching the existing pattern for `build` / `typecheck` / `test:*`.

### Healthy workspace (silent, fast)

```
$ time pnpm run preflight:workspace-links
> paperclip@ preflight:workspace-links /home/kay/forks/paperclip
> node cli/node_modules/tsx/dist/cli.mjs scripts/ensure-workspace-package-links.ts
real	0m0.786s   (≈350ms without the pnpm wrapper)
```

### AIU-491 reproduction (fails fast with exit 1)

```
$ rm -rf ui/node_modules/@assistant-ui
$ pnpm dev
[paperclip] workspace integrity check FAILED — 1 declared package(s) missing from node_modules:
  ui:
    - @assistant-ui/react@^0.12.23

[paperclip] The pnpm store may hold these packages but the workspace symlinks are stale.
[paperclip] Run: pnpm install --frozen-lockfile
```

## Non-scope

- Runtime UI-aware health endpoint (option B from AIU-491) — deferred; this PR handles the startup signal; B can be revisited after lands.
- HMR remote-access defaults (option C) — separate discussion, intentional breaking change.

## Test plan

- [x] Happy path on `~/forks/paperclip`: silent, exit 0, <1s end-to-end through pnpm (<500ms direct)
- [x] AIU-491 repro via fixture: `rm -rf ui/node_modules/@assistant-ui` → exit 1 with `@assistant-ui/react` named in the output
- [x] Fixture sanity test: 5 declared deps, 1 present, 4 missing → reports exactly the 4 missing grouped by workspace
- [x] `pnpm dev` chain simulation: preflight runs before `dev-runner.ts watch` as expected

Closes AIU-502.


kbecking:
Closing — opened against upstream in error. This belongs on our fork first (De-Supermarkt-Vergelijker/paperclip); re-opening there. Upstream PR requires separate board review per our internal workflow.

PR #4360, opened 2026-04-23
fix(opencode-local): dedupe concurrent model discovery calls

moramiky:
## Summary

Deduplicate concurrent calls to `discoverOpenCodeModelsCached` so parallel agent runs don't race the OpenCode model discovery fetch.

## Problem

When multiple agents (e.g. Renami + Finami runs) start around the same time, each one independently calls `discoverOpenCodeModelsCached`. Because there was no in-flight guard, N concurrent callers each spawned their own discovery request. On flaky network or cold-start conditions, this produced intermittent `opencode models failed` errors — even though the cache itself was fine, the burst of duplicate fetches was the actual trigger.

## Fix

Add a single-flight / in-flight promise dedupe in `packages/adapters/opencode-local/src/server/models.ts`:

- If a discovery call is already in-flight, subsequent callers await the same promise.
- Cache-hit path is unchanged.
- On completion (success or failure), the in-flight slot is cleared so the next miss can try again.

## Verification

- `pnpm --filter @paperclipai/adapter-opencode-local typecheck` → OK
- Post-patch re-smoke / stress runs: no more `opencode models failed`, majority of runs succeed.

## Scope

- Single file: `packages/adapters/opencode-local/src/server/models.ts`
- No schema, API, or UI changes.

### DIFF
--- a/packages/adapters/opencode-local/src/server/models.ts
+++ b/packages/adapters/opencode-local/src/server/models.ts
@@ -20,6 +20,7 @@ function resolveOpenCodeCommand(input: unknown): string {
 }
 
 const discoveryCache = new Map<string, { expiresAt: number; models: AdapterModel[] }>();
+const discoveryInFlight = new Map<string, Promise<AdapterModel[]>>();
 const VOLATILE_ENV_KEY_PREFIXES = ["PAPERCLIP_", "npm_", "NPM_"] as const;
 const VOLATILE_ENV_KEY_EXACT = new Set(["PWD", "OLDPWD", "SHLVL", "_", "TERM_SESSION_ID", "HOME"]);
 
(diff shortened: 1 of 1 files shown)


rileymiller:
Red-CI stale sweep ruling (2026-04-24 13:15 MDT): this PR has failed checks older than 12 hours and is covered by the same-run queue-health ruling.

- Owner ruling: explicit board blocker via SIG-631
- Board owner: Cleo
- Path: split to Forge/additional owners for fix-now triage, or approve close/supersede if stale
- Current startable Forge slice remains SIG-628 while SIG-629 waits for admin rerun or close/supersede decision
- Next check: 2026-04-24 14:30 MDT

PR #4417, opened 2026-04-24
fix(heartbeat): exempt standing rollup issues from auto-block

MalachiThorne:
## Summary

Fix for TEL-59 — the 6-hour blocker sweep was auto-moving standing rollup issues (routine parent log issues) to `blocked` every cycle because their assigned agent has no live execution path between routine fires. Each cycle required a manual CEO clear + restart-note comment.

An issue is now treated as a standing rollup when it is the `parentIssueId` of an `active` routine. In both paths that escalate stranded assigned issues to `blocked`:

1. `reconcileStrandedAssignedIssues` (the periodic sweep)
2. `releaseIssueExecutionAndPromote` (the terminal-run release path)

…the exemption short-circuits the block, leaves the issue in its current `todo`/`in_progress` state, and records an audit entry under `issue.auto_block_skipped_standing_rollup` for observability. Paused or deleted routines don't trigger the exemption.

The sweep result object now exposes a `standingRollupSkipped` counter for heartbeat telemetry.

## Why routine-parent (not labels)

Label-based allowlist was the cheapest-ship option flagged in the ticket, but routine-parent is zero-config and matches the observed failure exactly (TEL-30, the parent of the `Daily skill usage audit` routine). No migration, no label naming debate, no "I forgot the label" re-occurrences. If we later need exemption for hand-rolled rollup issues without a routine, adding a label check on top is additive.

## Test plan

- [x] `server/src/__tests__/heartbeat-process-recovery.test.ts` — 3 new positive tests (sweep `in_progress`, sweep `todo`, release path) and 1 negative test (paused routine still blocks)
- [x] Full file still passes: `npx vitest run src/__tests__/heartbeat-process-recovery.test.ts` → 22/22

### DIFF
--- a/server/src/__tests__/heartbeat-process-recovery.test.ts
+++ b/server/src/__tests__/heartbeat-process-recovery.test.ts
@@ -18,6 +18,7 @@ import {
   issueDocuments,
   issueRelations,
   issues,
+  routines,
 } from "@paperclipai/db";
 import {
   getEmbeddedPostgresTestSupport,
@@ -323,6 +324,7 @@ describeEmbeddedPostgres("heartbeat orphaned process recovery", () => {
     await db.delete(heartbeatRunEvents);
     await db.delete(heartbeatRuns);
     await db.delete(agentWakeupRequests);
+    await db.delete(routines);
     for (let attempt = 0; attempt < 5; attempt += 1) {
       await db.delete(agentRuntimeState);
       try {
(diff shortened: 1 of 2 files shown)


xkonjin (approved):
LGTM.
The standing rollup issue exemption effectively prevents these issues from being incorrectly flagged as stranded during heartbeat scans, addressing the recurring manual intervention burden.

### Code Review

**Summary:**
This PR introduces an exemption that prevents periodic checks from auto-blocking "standing rollup" issues (issues driving a routine execution). If an issue acts as a  for an active routine, it suppresses the automatic block escalation logic because the issue inherently lacks a continuous execution path between scheduled iterations. 

- New  check.
-  short circuits the auto-block.
- Terminal-run release logic also honors this block.
- Audit trails introduced for the exemption .

**Bugs:** None spotted.
**Security:** No vulnerabilities. Read-only checks applied cleanly inside boundary assertions.
**Test Coverage:** Good specific coverage via  for combinations of in-progress/todo states across both active and paused (negative test) routines.

PR #4430, opened 2026-04-24
feat(cost): impute API-equivalent cost for subscription-included runs

ramonmatias19:
## Summary

Runs billed as `subscription_included` used to be recorded as \$0 cost because the provider never returns a per-run dollar amount. This makes usage invisible on the cost dashboard, even when a subscription seat is effectively consuming real API-equivalent usage.

- Add `ANTHROPIC_PRICING_PER_MTOK` table for current models (Opus 4.6, Sonnet 4.6, Haiku 4.6, Sonnet 4.5, Haiku 4.5)
- Add `estimateCostCentsFromTokens()` that converts input / cachedInput / output token counts to cents at standard API prices
- Extend `normalizeBilledCostCents()` to accept an optional `tokenEstimate`; when `billingType` is `subscription_included` and tokens are provided, impute the cost at API rates instead of returning 0

## Test plan

- [ ] Run an agent on a Claude subscription — verify cost dashboard shows imputed dollars, not \$0
- [ ] Run an agent with an unknown model — verify it logs a `no pricing entry` warning and falls back to 0

### DIFF
--- a/server/src/services/heartbeat.ts
+++ b/server/src/services/heartbeat.ts
@@ -1041,8 +1041,54 @@ function resolveLedgerBiller(result: AdapterExecutionResult): string {
   return readNonEmptyString(result.biller) ?? readNonEmptyString(result.provider) ?? "unknown";
 }
 
-function normalizeBilledCostCents(costUsd: number | null | undefined, billingType: BillingType): number {
-  if (billingType === "subscription_included") return 0;
+/**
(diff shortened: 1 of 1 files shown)


ramonmatias19:
Duplicate of #3330, which is older and already has the  fix applied. Closing in favor of #3330.

ramonmatias19:
Duplicate of #3330. Closing in favor of the older PR.

PR #4465, opened 2026-04-25
security: always redact agent configuration in GET /agents list endpoint

artemdubik:
## Summary

`GET /agents` (list) currently returns raw agent rows (`res.json(result)`) when the actor has read-configs permission. That payload includes `adapter_config` which may contain API tokens, OAuth secrets, and custom env var values. By contrast, the single-agent `GET /agents/:id` and `GET /scheduler-heartbeats` handlers already apply `redactAgentConfiguration(...)` before returning.

This PR aligns the list endpoint with the already-existing single-agent redaction path: always apply `redactAgentConfiguration` on each row. `board` actor type is also covered — same redaction applied.

## Change

- `server/src/routes/agents.ts` (line ~1118): wrap `result` with `result.map((agent) => redactAgentConfiguration(agent))` for the read-configs branch, include `req.actor.type === "board"` in the same branch.

## Test plan

- [x] Existing tests pass
- [ ] Happy to add an explicit test if a reviewer prefers
- [x] Manual check in production: `/agents` response no longer contains raw `adapter_config.apiKey` values

Caught while auditing what a board-level actor sees when calling `/agents`.

Security hardening — no functional change for actors that already received redacted output.

### DIFF
--- a/server/src/routes/agents.ts
+++ b/server/src/routes/agents.ts
@@ -1115,8 +1115,8 @@ export function agentRoutes(
     }
     const result = await svc.list(companyId);
     const canReadConfigs = await actorCanReadConfigurationsForCompany(req, companyId);
-    if (canReadConfigs) {
-      res.json(result);
+    if (canReadConfigs || req.actor.type === "board") {
+      res.json(result.map((agent) => redactAgentConfiguration(agent)));
       return;
     }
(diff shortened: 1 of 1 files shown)


artemdubik:
Closing — opened by mistake. Same reason as #4464. Sorry for the noise.

PR #4030, opened 2026-04-19
fix(db): add ON DELETE rules to cost_events foreign keys

scokeepa:
## Thinking Path

`cost_events` references six parent tables (`companies`, `agents`, `issues`, `projects`, `goals`, `heartbeat_runs`) but none of those FKs declare an `onDelete` clause. PostgreSQL therefore defaults to `NO ACTION`, which means deleting any parent row that has cost events attached — for example deleting a finished issue or a removed agent — fails with `23503` foreign-key violation.

The right behaviour depends on the parent:
- A company or agent disappearing should also remove its cost ledger; those rows have no meaning without an owner.
- An issue, project, goal, or heartbeat run being deleted should leave billing history intact for accounting; those FKs should `SET NULL`.

This was reproducible by deleting a recently used agent or issue once cost events had been written for it.

Refs upstream issues #3878, #3849.

## What Changed

- `packages/db/src/schema/cost_events.ts`: added `onDelete: "cascade"` for `companyId` / `agentId`, `onDelete: "set null"` for `issueId` / `projectId` / `goalId` / `heartbeatRunId`.
- `packages/db/src/migrations/0058_cost_events_cascade_fks.sql`: new migration that drops the existing constraints and re-adds them with the chosen `ON DELETE` actions.
- `packages/db/src/migrations/meta/_journal.json`: appended journal entry for migration 58.

## Verification

- Schema/migration diff reviewed against pgsql syntax; constraint names match Drizzle's naming convention so the `DROP` half matches what's in the database today.
- Manually walked the dependency graph: every FK in `cost_events` is now covered, and the chosen rule matches whether the cost row stays meaningful without that parent.
- Did not run drizzle-kit generate locally — happy to regenerate the SQL/snapshot via `pnpm db:generate` if maintainers prefer that flow.

## Risks

- The migration assumes the historical constraint names (`cost_events_<col>_<table>_id_fk`) — which is what Drizzle currently emits, so this should be a no-op rename for any environment that ran prior migrations cleanly.
- `heartbeat_run_id` constraint uses `DROP CONSTRAINT IF EXISTS` since some early environments may not have it; safe to keep on environments that already have it.
- No data migration is needed; only constraint definitions change.

## Checklist

- [x] One logical change
- [x] Migration script included alongside schema change
- [x] Journal entry updated

PR #4473, opened 2026-04-25
fix(db): add ON DELETE rules to issue-related foreign keys

alexlomt:
## Thinking Path

> - Issues are the primary unit of work — every task agents do is tracked as an issue
> - Many tables hold child records that reference an issue (comments, read-states, votes, archives, audit/cost events, etc.)
> - When users (or routine cleanup paths) try to delete an issue, the request must succeed even if children exist
> - Several of those child tables have foreign keys to `issues.id` with **no** `ON DELETE` action, so PostgreSQL falls back to `NO ACTION` and refuses the delete
> - The result is a hard HTTP 500 from `DELETE /api/issues/{id}` whenever any of those children exist (reproduced live: deleting an issue an agent has run against is impossible)
> - PR #4030 fixes this pattern for `cost_events`. This PR fixes it for the remaining issue-related child tables it doesn't touch (six tables total).
> - The benefit is `DELETE /api/issues/{id}` reliably succeeds, and child data is handled correctly: lifecycle children cascade, audit/event data is preserved with `issue_id` set to NULL.

## What Changed

- `issue_inbox_archives.issue_id` → `ON DELETE CASCADE` (lifecycle child of an issue)
- `finance_events.issue_id` → `ON DELETE SET NULL` (audit/event row; preserve history when the issue is gone)
- `feedback_votes.issue_id` → `ON DELETE CASCADE` (column is `NOT NULL` so `SET NULL` would require a column change; a vote without its parent issue is meaningless, so cascade is the right call)
- `issue_comments.issue_id` → `ON DELETE CASCADE` (lifecycle child)
- `issue_read_states.issue_id` → `ON DELETE CASCADE` (lifecycle child)
- `issue_thread_interactions.issue_id` → `ON DELETE CASCADE` (lifecycle child — agent-created clarification/approval threads)
- New migration: `packages/db/src/migrations/0071_issue_fk_family_cascade.sql` — drops and re-adds the six FK constraints with the rules above
- Updated `packages/db/src/migrations/meta/_journal.json` with the matching idx 71 entry
- New regression test in `packages/db/src/client.test.ts`: `applies ON DELETE rules from migration 0071 to issue-related foreign keys` — inserts a row in each of the six tables, deletes the parent issue, and asserts cascade rows are gone while the set-null row survives with `issue_id = NULL`. Also asserts `pg_constraint.confdeltype` matches the expected behavior.

## Verification

- `pnpm --filter @paperclipai/db typecheck` — passes
- `pnpm --filter @paperclipai/db build` — passes (includes `check:migrations`)
- `cd packages/db && pnpm exec vitest run src/client.test.ts` — all 9 tests pass, including the new one (`applies ON DELETE rules from migration 0071 to issue-related foreign keys`, ~4.7s)
- The new test exercises every FK touched by this migration end-to-end against a fresh embedded Postgres database.
- Reviewers can manually verify by running migrations against any existing dev DB and inspecting `pg_constraint.confdeltype` for the five FKs (expect `c`, `n`, `c`, `c`, `c` respectively).

## Risks

- **Low–medium.** The migration only changes `ON DELETE` behavior on existing FKs; it does not touch column nullability, indexes, or data.
- The five FKs all already exist in production schemas. Replacing them with the same FK + an action clause is a metadata-only change, so the migration is fast and does not rewrite tables.
- One judgment call: `feedback_votes.issue_id` is `NOT NULL`, so `SET NULL` is impossible without changing the column. We chose `CASCADE` — a vote on a deleted issue has no anchor and isn't meaningful audit data — rather than expand scope by altering the column. Happy to revisit if the maintainers prefer the column be made nullable instead.
- Non-overlapping with #4030 (different tables, different migration slot, different journal entry).

## Model Used

- **Provider:** Anthropic (Claude)
- **Model:** Claude Opus 4.7 (`anthropic/claude-opus-4-7`)
- **Mode:** Tool-use, code execution against a live workspace
- **Capabilities used:** filesystem edits, shell execution, git, GitHub CLI, automated test runs

## Checklist

- [ ] If this change affects the UI, I have included before/after screenshots — *N/A, schema/DB only*
- [x] I have updated relevant documentation to reflect my changes — *no doc surface affected; migration + journal entry are the canonical record*


alexlomt:
@greptileai review

Pushed fixes addressing both review comments:
- P1: `issue_thread_interactions.issue_id` added to migration 0071 with `ON DELETE CASCADE`; schema synced; regression test extended.
- P2: `cost_events` schema synced with #4030's migration semantics (`companyId`/`agentId` cascade, `issueId`/`projectId`/`goalId`/`hb_signalRunId` set null).

Build, typecheck, and the regression suite all pass locally.

alexlomt:
@greptileai review

You're right — reverted the cost_events schema change. Since #4030 has not merged into master, syncing cost_events.ts here would introduce real drift against the upstream master DB state (where all six FKs are still `NO ACTION`). The right place to sync that schema file is inside #4030 itself, alongside its migration 0058.

This PR is now tightly scoped to the six issue-related FK fixes that migration 0071 actually applies.

PR #4472, opened 2026-04-25
ci: unified Docker-based CI/CD foundation (mv50000/cicd@v1)

mv50000:
## Thinking Path

> - Paperclip companies (saatavilla, quantimodo, alli-audit, bk, optimi) each grew their own deploy story — Makefile+systemd, cargo+systemd, mixed Docker/SSH. Adding the next company means learning yet another deploy variant, and changing one shared concern (e.g. tag policy, healthcheck behavior) means editing five different workflows.
> - The mv50000/cicd repo (public, separate from this monorepo) extracts that into reusable workflows + composite actions: each company calls `mv50000/cicd/.github/workflows/build-and-deploy.yml@<sha>` from a thin ~30-line `deploy.yml`. Build → push GHCR → SSH rsync → `docker compose pull && up --wait` → wait-for-health → auto-rollback on failure.
> - This PR lands the **Paperclip-side companion** to that pipeline: a deploy skill agents can invoke, helper scripts to migrate other repos, architecture docs, and per-repo workflow files (auto-merge, deploy-dev) that the CI/CD pipeline integrates with. It does **not** add Paperclip dependencies on mv50000/cicd; the companies do that independently.
> - Greptile flagged earlier revisions of this PR for bundling three unrelated runtime behavior changes and exposing internal infra data in committed docs. Those are now addressed: the unrelated changes have been split into #4496 (session rotation) and #4497 (max-turns fix); the internal todo doc has been removed; and template downloads pin to a specific cicd commit SHA rather than the mutable `v1` tag.

## What Changed

| File | Purpose |
|------|---------|
| `.agents/skills/deploy/SKILL.md` | Engineering-agent skill that watches a merged PR through the deploy workflow, validates health, documents (but never auto-executes) rollback. Distinct from `release` (Paperclip product) and `prcheckloop` (pre-merge PR checks). |
| `scripts/migrate-company.sh` | Interactive helper that fetches Dockerfile.{rust,node}, compose, and workflow templates from `mv50000/cicd@<pinned-sha>` into a target company repo (auto-detects stack from Cargo.toml/package.json). Does not commit. SHA-pinned (overridable via `CICD_REF` env). |
| `scripts/audit-runners.sh` | Lists self-hosted runners in `~/actions-runners` and flags repos that are 404 or archived on GitHub. Tested 2026-04-25: 0 orphans across alli-audit, bk, optimi, quantimodo-rust, saatavilla. |
[...]

PR #4496, opened 2026-04-25
perf: proactive session rotation to avoid Haiku compaction costs

mv50000:
## Thinking Path

> - Claude Code's internal compaction triggers a Haiku summarization round when a session approaches its context window. For long-running agent runs that's a measurable cost line — Haiku tokens we don't need to pay for if we recycle the session before the trigger fires.
> - Paperclip already owns the run lifecycle through ADAPTER_MANAGED_SESSION_POLICY (heartbeat-managed, per-adapter), so the cleanest place to enforce a proactive rotation cadence is here, before the in-process compaction kicks in.
> - The previous all-zeros default ("no policy") existed because the rotation thresholds had not been validated against real workloads. After two months of production heartbeat data they're well-characterized: 8 runs / 400k input tokens / 24 hours all sit comfortably ahead of Haiku's compaction edge for the adapters with native context management (claude_local, codex_local, hermes_local).
> - Lowering UI default `maxTurnsPerRun` from 1000 to 200 is a tangential UX fix that travels with the rotation work — most agents finish well under 200 turns; 1000 was a placeholder that confused operators reading run logs.

## What Changed

- `packages/adapter-utils/src/session-compaction.ts`: ADAPTER_MANAGED_SESSION_POLICY for **claude_local, codex_local, and hermes_local** switches from all-zeros (no rotation) to `{ runs: 8, inputTokens: 400_000, ageHours: 24 }`. All three are registered with `nativeContextManagement: "confirmed"` and share the same policy object.
- `ui/src/components/agent-config-defaults.ts`: new agent default `maxTurnsPerRun` 1000 → 200.
- `server/src/__tests__/heartbeat-workspace-session.test.ts`: assertion covers all three adapters (claude_local, codex_local, hermes_local) and the partial-override fall-through behaviour.

## Migration Note (was Greptile P1 in #4472)

Three adapters are affected: **claude_local, codex_local, hermes_local**. All three share `ADAPTER_MANAGED_SESSION_POLICY` because they have confirmed native context management, so they all receive the same threshold change.

This change does mean partial-override agents on any of these adapters — those that set some session-compaction fields but not others — pick up the new active thresholds for the fields they didn't set. Previously those unspecified fields resolved to 0 (disabled). The behavior change is intentional and benefits these agents (they get rotation hygiene), but operators who relied on "I set just the runs threshold and the rest is off" should review their agent configs after merge.

Existing agents were updated in-DB out-of-band on 2026-04-22 with role-based values (engineer/cto 300, ceo/manager 150, qa 200, general 100), so the in-DB state is already aligned. Only configs with explicit partial overrides set since then need attention.

## Verification

- `pnpm -r typecheck` ✅ (CI)
- `pnpm test:run` ✅ (CI; updated heartbeat-workspace-session test asserts the same thresholds for all three confirmed-native adapters and verifies partial-override fall-through)
- `pnpm build` ✅ (CI)
- Production observation 2026-04-22 → 04-25: per-agent input-token compaction cost down materially; no regressions in run completion rates.

## Risks

- Partial-override agents on claude_local / codex_local / hermes_local (see migration note). Mitigated by the in-DB sync done 2026-04-22 and by the role-based defaults already in production.
- maxTurnsPerRun default reduction surfaces previously-rare "ran out of turns" errors for long-form agents. Mitigated by companion PR #4497 which ensures the flag is always passed to Claude Code, so the runtime behavior is now explicit rather than implicit.

## Checklist

- [x] Behaviour matches spec
- [x] `pnpm -r typecheck` passes
- [x] `pnpm test:run` passes
- [x] `pnpm build` passes
- [x] Contracts in sync across layers
- [x] Documentation: migration note above

Split out from #4472 (CI/CD foundation) per Greptile P1 — that PR was bundling unrelated runtime behavior changes; this is now a focused rotation-policy PR with its own migration story covering all three confirmed-native adapters.

### DIFF
--- a/packages/adapter-utils/src/session-compaction.ts
+++ b/packages/adapter-utils/src/session-compaction.ts
@@ -27,13 +27,14 @@ const DEFAULT_SESSION_COMPACTION_POLICY: SessionCompactionPolicy = {
   maxSessionAgeHours: 72,
 };
 
-// Adapters with native context management still participate in session resume,
-// but Paperclip should not rotate them using threshold-based compaction.
+// Adapters with native context management still participate in session resume.
+// Paperclip rotates sessions proactively before the runtime's own compaction
+// kicks in, avoiding expensive Haiku compaction calls on long contexts.
 const ADAPTER_MANAGED_SESSION_POLICY: SessionCompactionPolicy = {
   enabled: true,
-  maxSessionRuns: 0,
-  maxRawInputTokens: 0,
-  maxSessionAgeHours: 0,
+  maxSessionRuns: 8,
+  maxRawInputTokens: 400_000,
+  maxSessionAgeHours: 24,
 };
 
 export const LEGACY_SESSIONED_ADAPTER_TYPES = new Set([
--- a/server/src/__tests__/heartbeat-workspace-session.test.ts
+++ b/server/src/__tests__/heartbeat-workspace-session.test.ts
@@ -509,18 +509,18 @@ describe("prioritizeProjectWorkspaceCandidatesForRun", () => {
 });
 
 describe("parseSessionCompactionPolicy", () => {
-  it("disables Paperclip-managed rotation by default for codex and claude local", () => {
+  it("applies proactive rotation thresholds by default for codex and claude local", () => {
     expect(parseSessionCompactionPolicy(buildAgent("codex_local"))).toEqual({
       enabled: true,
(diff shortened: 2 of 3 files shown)


mv50000:
@greptileai review

serkanpektas (approved):
-

PR #4498, opened 2026-04-25
Linkify absolute paths in agent comments to vscode://

nomadec:
## Thinking Path

> - Agent comments (especially Fixer's approval comments) cite absolute on-disk artifacts — research notes, plans, worktree dirs
> - Today those paths render as plain text, so the human reviewer has to copy/paste into a terminal or editor to inspect them
> - The internal Sona team scoped two options (Option A renderer rewrite, Option B "Open worktree in editor" chip on the issue UI) and approved A first as the cheap, generic, frontend-only fix
> - This pull request implements Option A: the comment markdown renderer detects absolute paths inside backticks and rewrites them to `vscode://file/<abs>` links at render time, without touching any agent prompt template
> - The benefit is that every current and future agent gets clickable paths "for free" — one click opens the file (or directory as a workspace) in VS Code or Cursor

## What Changed

- Added `ui/src/lib/absolute-path-link.ts` — a remark plugin (`remarkLinkAbsolutePaths`) that walks the markdown AST and rewrites `inlineCode` nodes whose value matches an absolute-path prefix allowlist into `link` nodes with `vscode://file/<abs>` URLs. Mirrors the pattern of the existing `remarkLinkIssueReferences` plugin.
- Detection rule: backtick-wrapped tokens whose value matches `/(?:Users|home|private|var|tmp|opt)/` or `~/`. Tokens containing `://`, newlines, or NUL bytes are skipped to avoid URLs and multi-line content. Path segments are URL-encoded so paths with spaces work.
- Wired the plugin into `MarkdownBody` so it runs after issue-reference linkification (avoiding double rewrites) and before soft-break rendering.
- Extended `safeMarkdownUrlTransform` in `MarkdownBody.tsx` to allow `vscode:` URLs through. react-markdown's `defaultUrlTransform` allowlist is `^(https?|ircs?|mailto|xmpp)$` — without this change, every `vscode://` href (auto-generated or hand-authored) gets stripped to empty.
- Added a hover tooltip ("Open in VS Code / Cursor") and a `data-editor-link="true"` attribute on rendered editor-protocol links for observability.
- Added 8 unit tests covering: positive linkification (absolute and `~/` paths), URL-encoding of spaces, rejection of paths outside backticks, fenced code blocks, URLs inside backticks, plain shell commands, non-allowlisted paths like `/etc/`, and preservation of manually authored `vscode://` markdown links through the sanitizer.

## Verification

- `pnpm --filter @paperclipai/ui run typecheck` — passes
- `pnpm --filter @paperclipai/ui exec vitest run src/components/MarkdownBody.test.tsx` — 38/38 pass (30 prior + 8 new)
- `pnpm --filter @paperclipai/ui exec vitest run` — 631/631 pass

Manual sanity check (from inspecting tests output):

- `` `/Users/umid/.paperclip/research.md` `` → `<a href="vscode://file/Users/umid/.paperclip/research.md" title="Open in VS Code / Cursor"><code>/Users/umid/.paperclip/research.md</code></a>`
- `/Users/umid/file.md` (no backticks) → unchanged plain text
- ```` ```\n/Users/umid/file.md\n``` ```` → unchanged inside fenced block
- `` `https://example.com/path` `` → no rewrite (contains `://`)
- `` `/etc/hosts` `` → no rewrite (not in allowlist)

## Risks

- **Low risk** overall — frontend-only, no API or schema changes.
- False positives are bounded by the prefix allowlist + the requirement that the path be inside backticks (an explicit author signal). Worst case for a missed match is opening a file the user didn't intend, which is benign.
- vscode:// URL handling on macOS works for both VS Code and Cursor (both register the protocol). Pilot is mac-only per the scoping issue. Windows and remote/cloud editors are deferred (would need URL-encoded backslashes / separate transport).
- `~/`-prefixed paths are emitted as `vscode://file/~/...`. Editor expansion of `~` is platform-dependent; agents that want guaranteed open-on-click should emit fully-qualified `/Users/...` paths. Not a regression — those paths render the same way they do today (plain inline code), just unclickable.

## Model Used

- Claude Opus 4.7 (1M context), via Claude Code

## Checklist

- [ ] If this change affects the UI, I have included before/after screenshots — N/A: behavior verified by unit tests asserting the rendered HTML; no visual styling changes (link uses default markdown anchor styling, code element retains existing monospace formatting)
- [ ] I have updated relevant documentation to reflect my changes — N/A: no public API or developer-doc surface changed

### DIFF
--- a/ui/src/components/MarkdownBody.test.tsx
+++ b/ui/src/components/MarkdownBody.test.tsx
@@ -427,6 +427,16 @@ describe("MarkdownBody", () => {
     expect(html).toContain('title="Open in VS Code / Cursor"');
   });
 
+  it("strips non-file vscode:// schemes (extension/settings) from manual markdown links", () => {
+    const html = renderMarkdown(
+      "[ext](vscode://extension/some-ext/run-cmd) [settings](vscode://settings)",
+    );
+
+    expect(html).not.toContain("vscode://extension");
+    expect(html).not.toContain("vscode://settings");
+    expect(html).toContain('href=""');
+  });
+
   it("renders internal issue links and bare identifiers as inline issue refs", () => {
     const html = renderMarkdown(`See PAP-42 and [linked task](${buildIssueReferenceHref("PAP-77")}) for follow-up.`, [
       { identifier: "PAP-42", status: "done" },
--- a/ui/src/components/MarkdownBody.test.tsx
+++ b/ui/src/components/MarkdownBody.test.tsx
@@ -365,6 +365,68 @@ describe("MarkdownBody", () => {
     expect(html).toContain('style="max-width:100%;overflow-x:auto"');
   });
 
+  it("linkifies absolute paths inside backticks to vscode://file links", () => {
+    const html = renderMarkdown("Plan at `/Users/umid/.paperclip/research.md` for context.");
+
+    expect(html).toContain('href="vscode://file/Users/umid/.paperclip/research.md"');
+    expect(html).toContain("<code");
+    expect(html).toContain(">/Users/umid/.paperclip/research.md</code>");
+    expect(html).toContain('title="Open in VS Code / Cursor"');
+    expect(html).toContain('data-editor-link="true"');
+  });
+
+  it("linkifies tilde-prefixed home paths inside backticks", () => {
+    const html = renderMarkdown("Run from `~/.paperclip/instances/default`.");
+
+    expect(html).toContain('href="vscode://file/~/.paperclip/instances/default"');
+    expect(html).toContain(">~/.paperclip/instances/default</code>");
+  });
+
+  it("URL-encodes path segments containing spaces", () => {
+    const html = renderMarkdown("Open `/Users/Bob Smith/file.md` to review.");
+
+    expect(html).toContain('href="vscode://file/Users/Bob%20Smith/file.md"');
+    expect(html).toContain(">/Users/Bob Smith/file.md</code>");
+  });
+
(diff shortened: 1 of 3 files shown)


nomadec:
Thanks @greptile-apps. Addressed in 10b523b0:

**P2 #1 — `vscode:` allowlist too broad → fixed.** `isEditorProtocolUrl` now narrows to `vscode://file/` only. `vscode://extension/...` and `vscode://settings` get stripped to empty href just like any other unsafe scheme. Added a regression test asserting that exact behavior, plus a code comment explaining why we're not just allowing `vscode:` wholesale.

**P2 #3 — `~/` silent failure → documented.** Added a code comment in `absolute-path-link.ts` calling out that `vscode://file` does not expand `~`, and that agents wanting guaranteed-open links should emit `/Users/...`. Server-side `~` resolution is a heavier change (we'd need to know which user's home to expand against — agent comments come from agent runs, not authenticated user sessions); deferring until/unless someone reports an actual broken click.

**P2 #2 — `linkAbsolutePaths` opt-out prop → declining for now.** Reasoning: `remarkLinkIssueReferences` is gated because it pulls in React Query for status decoration; its plugin has a real opt-out cost. `remarkLinkAbsolutePaths` is a pure AST transform with no fetching or external state, and the existing `remarkSoftBreaks` plugin runs unconditionally too. I'd rather not add a prop without a concrete callsite that needs it (premature abstraction) — happy to add one in a follow-up if a real consumer surfaces.

PR #4503, opened 2026-04-25
fix(heartbeat): make agent pause status guard atomic with status update

supertaz:
## Thinking Path

> - Paperclip orchestrates AI agents that run as persistent server processes with heartbeat tracking
> - The `finalizeAgentStatus` function updates an agent's status when a run finishes
> - It reads the agent's current status, returns early if paused/terminated, then issues an UPDATE with only `WHERE id = $agentId`
> - A concurrent pause write can commit between the read and the UPDATE, causing the run completion to overwrite `paused → idle`
> - This PR adds `AND status NOT IN (paused, terminated)` to the UPDATE WHERE clause, making the guard atomic
> - The benefit is that a user pause that races with a run finishing now always wins — the agent stays paused

## What Changed

- `server/src/services/heartbeat.ts`: added `notInArray(agents.status, ["paused", "terminated"])` to the UPDATE WHERE clause in `finalizeAgentStatus`, closing the TOCTOU window between the read guard and the write
- `server/src/__tests__/agent-pause-status-race.test.ts`: 4 new embedded-Postgres tests — end-to-end `cancelRun` with a pre-paused agent, and direct SQL verification that the WHERE guard blocks updates for `paused`, `terminated`, and allows updates for `running`

## Verification

- `pnpm typecheck` — clean
- `npx vitest run --project "@paperclipai/server" server/src/__tests__/agent-pause-status-race.test.ts` — 4/4 pass

## Risks

Low. The change makes the existing early-return guard atomic with the write. Any caller that expected `finalizeAgentStatus` to overwrite a paused agent was relying on incorrect behavior. The `tickTimers` reconciler already skips paused/terminated agents (line 7342), so a skipped `lastHeartbeatAt` update on a paused agent has no practical effect — this is the same behavior as the existing early-return path.

## Model Used

- Provider: Anthropic
- Model: Claude Sonnet 4.6 (`claude-sonnet-4-6`)
- Context window: standard
- Mode: agentic tool use with file read/write/edit
- Self-review: gpt-5.4-mini (`model_reasoning_effort=high`) — 1 finding reviewed and confirmed pre-existing behavior

## Checklist

- [x] I have included a thinking path that traces from project context to this change
- [x] I have specified the model used (with version and capability details)
- [x] I have checked ROADMAP.md and confirmed this PR does not duplicate planned core work
- [x] I have run tests locally and they pass
- [x] I have added or updated tests where applicable
- [x] If this change affects the UI, I have included before/after screenshots (N/A — server-only)
[...]

PR #4459, opened 2026-04-25
fix(heartbeat): escalate stranded issue when recovery retry succeeds without execution path

supertaz:
## Thinking Path

> - Paperclip orchestrates AI agents that work on issues assigned to them; agents run via a heartbeat service that schedules and monitors execution runs
> - The stranded-issue reconciler (`reconcileStrandedAssignedIssues`) runs every 30 s and finds assigned issues with no active execution path, then queues a recovery run
> - For issues already on their automatic recovery retry, an escalation gate (`didAutomaticRecoveryFail`) was supposed to detect exhaustion and escalate to `blocked` instead of re-queuing
> - The gate checked for unsuccessful terminal statuses only (`failed`, `cancelled`, `timed_out`) — it omitted `succeeded`
> - A recovery run that exits `succeeded` without re-establishing an execution path leaves the issue stranded, but the gate returned `false`, so the reconciler re-queued indefinitely
> - This PR extends the exhaustion check to include `succeeded` and renames the function to `didAutomaticRecoveryExhaust` to reflect the corrected semantics
> - A follow-up commit removes the now-dead `UNSUCCESSFUL_HEARTBEAT_RUN_TERMINAL_STATUSES` constant from both files

## What Changed

- `recovery/service.ts`: swap `UNSUCCESSFUL_HEARTBEAT_RUN_TERMINAL_STATUSES` for `HEARTBEAT_RUN_TERMINAL_STATUSES` in `didAutomaticRecoveryExhaust`; export the function
- `recovery/service.ts`: add `HEARTBEAT_RUN_TERMINAL_STATUSES` constant (includes `succeeded`); remove the now-unused `UNSUCCESSFUL_HEARTBEAT_RUN_TERMINAL_STATUSES` constant
- `heartbeat.ts`: remove the parallel local definition of `didAutomaticRecoveryExhaust`; import and use the canonical version from `recovery/service.ts`; remove unused `UNSUCCESSFUL_HEARTBEAT_RUN_TERMINAL_STATUSES` constant
- `heartbeat-process-recovery.test.ts`: update the test that previously asserted re-queue behavior on a succeeded recovery run to assert escalation instead

## Verification

- All 20 tests in `server/src/__tests__/heartbeat-process-recovery.test.ts` pass
- TypeScript: `npx tsc --noEmit` in `server/` exits clean
[...]

PR #4506, opened 2026-04-25
fix(heartbeat): suppress continuation recovery on user-initiated cancel

supertaz:
## Thinking Path

> - Paperclip orchestrates AI agents; users can pause or terminate agents via the board UI
> - Pausing an agent calls `cancelActiveForAgentInternal` which cancels in-flight runs
> - `cancelActiveForAgentInternal` then calls `releaseIssueExecutionAndPromote` for each cancelled run
> - `releaseIssueExecutionAndPromote` checks if the issue needs recovery and, if so, immediately enqueues a new run
> - A user pause therefore re-triggers the exact work the user just stopped
> - This PR adds a `suppressContinuationRecovery` option that fires the `blocked` escalation path instead of queuing recovery when the cancel is user-initiated
> - The benefit is that pausing an agent actually stops work; the issue moves to `blocked` so an operator can review, but no automatic re-trigger occurs

## What Changed

- `releaseIssueExecutionAndPromote`: new `opts.suppressContinuationRecovery` flag — when set, `shouldBlockImmediately` is forced true, so the issue is escalated to `blocked` instead of queuing a recovery run
- `cancelActiveForAgentInternal`: new `opts.userInitiated` flag — passes `suppressContinuationRecovery: userInitiated` to `releaseIssueExecutionAndPromote`; also now captures the updated (cancelled) run from `setRunStatus` and passes it to the release call so `issueNeedsImmediateRecovery` evaluates against the correct terminal status. Falls back to original run object if `setRunStatus` returns null (concurrent transition)
- `cancelActiveForAgent` (public export): marked `userInitiated: true` — covers the pause and terminate routes
- Budget-scope cancellations via `cancelBudgetScopeWork` are NOT marked `userInitiated`; they continue to enqueue recovery runs
- `server/src/__tests__/user-cancel-suppresses-continuation.test.ts`: 2 new embedded-Postgres tests verifying no continuation run and no continuation wakeup request are created after `cancelActiveForAgent`

## Verification

- `pnpm typecheck` — clean
- `npx vitest run --project "@paperclipai/server" server/src/__tests__/user-cancel-suppresses-continuation.test.ts` — 2/2 pass

## Risks

Low-medium. User-initiated pauses will now move assigned in-progress issues to `blocked` rather than immediately re-scheduling them. This is the intended behavior. Budget-pause cancellations are explicitly excluded and continue to re-schedule. Any operator who relies on a pause immediately re-queuing recovery will need to manually resume or comment to restart the issue.

## Model Used

- Provider: Anthropic
- Context window: standard
- Mode: agentic tool use with file read/write/edit
- Self-review: gpt-5.4-mini (`model_reasoning_effort=high`) — 1 finding addressed (preserve execution-lock cleanup path when `setRunStatus` returns null)

## Checklist

- [x] If this change affects the UI, I have included before/after screenshots (N/A — server-only)

## Open PR Overlap Analysis

This PR is related to #4510, #4503, #4459, #4356, and external cancellation/pause PRs such as #3533, but it is not a direct duplicate.

- #4510 adds `cancel_source` to heartbeat runs, which is a better long-term substrate for distinguishing user, system, and budget cancellations. It is broader and more durable, but requires schema changes and downstream adoption.
- This PR is narrower: it directly suppresses continuation recovery when cancellation is user initiated. It is simpler and lower risk for the observed bug, but less expressive than #4510's explicit cancellation provenance.
- #4503 protects paused agent status from being overwritten during cancellation/finalization races. It is related through cancellation side effects, but solves status atomicity rather than continuation recovery.
- #4356 enforces system/manual/auto pause at enqueue sites and cancels in-flight work on pause. It is much broader and riskier, while this PR is a targeted behavior fix.
- #3533 also touches stale queued run cancellation and UI suppression; it overlaps in cancellation semantics but is not a direct replacement.

This is complementary to my related cancellation/recovery PR group. The group is intentionally split into narrow, reviewable PRs rather than one large cancellation rewrite.


supertaz:
@greptile recheck

serkanpektas (approved):
-

PR #4511, opened 2026-04-25
fix(server): validate X-Paperclip-Run-Id header against UUID regex

MrDeanAnderson:
## Thinking Path

> - Paperclip orchestrates AI agents, and the server middleware decides how request identity and run metadata enter that system.
> - `actorMiddleware` sits on the request boundary for both board and agent callers, so malformed metadata there can poison downstream persistence.
> - `X-Paperclip-Run-Id` is stored in UUID-backed columns, which means accepting arbitrary strings at the header layer can surface as database 500s instead of a safe no-op.
> - MAR-363 addressed that by validating the header before assigning `req.actor.runId`.
> - The first pass used a version-constrained UUID regex, which fixed malformed values like `24916751496-1` but also rejected valid canonical UUID v7 values.
> - This pull request broadens the validation to accept canonical UUID shape regardless of version nibble, while still rejecting malformed non-UUID headers.
> - The benefit is that the middleware stays permissive for legitimate callers, including modern UUID generators, without reintroducing the Postgres failure mode.

## What Changed

- Relaxed the `X-Paperclip-Run-Id` regex in `server/src/middleware/auth.ts` to accept canonical UUID shape without constraining the version nibble to `1-5`
- Added a focused test proving a UUID v7 header is preserved by `actorMiddleware`
- Kept the existing malformed-header rejection and JWT fallback behavior unchanged

## Verification

- `pnpm --filter @paperclipai/server exec vitest run src/__tests__/auth-middleware.test.ts`

## Risks

- Low risk. This slightly broadens accepted UUID variants at the request boundary, but still enforces canonical UUID shape and variant bits, so malformed values continue to be dropped.

## Model Used

- OpenAI GPT-5 Codex via Codex local agent runtime (`gpt-5.4`), with tool use and shell execution in the local workspace

## Checklist

### DIFF
--- a/server/src/__tests__/auth-middleware.test.ts
+++ b/server/src/__tests__/auth-middleware.test.ts
@@ -1,0 +1,96 @@
+import express from "express";
+import request from "supertest";
+import { describe, expect, it, beforeEach, afterEach, vi } from "vitest";
+import type { Db } from "@paperclipai/db";
+import { actorMiddleware } from "../middleware/auth.js";
+import { createLocalAgentJwt } from "../agent-auth-jwt.js";
+import * as boardAuthModule from "../services/board-auth.js";
+
+describe("actorMiddleware run id header validation", () => {
+  const secretEnv = "PAPERCLIP_AGENT_JWT_SECRET";
+  const ttlEnv = "PAPERCLIP_AGENT_JWT_TTL_SECONDS";
+  const originalEnv = {
+    secret: process.env[secretEnv],
+    ttl: process.env[ttlEnv],
+  };
+
+  beforeEach(() => {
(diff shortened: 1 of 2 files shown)


MrDeanAnderson:
Addressing the current feedback in head commit `7d6b29a8`:

- broadened the header regex so canonical UUID v7 values are accepted instead of being silently dropped
- added focused v7 coverage in `server/src/__tests__/auth-middleware.test.ts`
- rewrote the PR body to match the required template sections

Local verification remains:
- `pnpm --filter @paperclipai/server exec vitest run src/__tests__/auth-middleware.test.ts`

serkanpektas (approved):
-

PR #4512, opened 2026-04-25
feat(inbox): resolve failed runs server-side

alexlomt:
## Summary
- add a server-side `POST /heartbeat-runs/:runId/resolve` endpoint for failed/timed-out runs
- preserve heartbeat run history while recording the acknowledgement as the existing server-backed inbox dismissal for `run:<id>`
- route failed-run archive/dismiss actions through the new endpoint so Focus/sidebar counts exclude acknowledged failures

## Tests
- `pnpm exec vitest run server/src/__tests__/inbox-dismissals.test.ts`

### DIFF
--- a/server/src/__tests__/inbox-dismissals.test.ts
+++ b/server/src/__tests__/inbox-dismissals.test.ts
@@ -1,6 +1,10 @@
 import { randomUUID } from "node:crypto";
+import express from "express";
+import request from "supertest";
 import { afterAll, afterEach, beforeAll, describe, expect, it } from "vitest";
+import { eq } from "drizzle-orm";
 import {
+  activityLog,
   agents,
   approvals,
   companies,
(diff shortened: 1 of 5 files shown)


alexlomt:
Addressed Greptile review comments:

- Moved board/user validation before heartbeat-run lookup and constrained the lookup to the actor's visible companies to avoid cross-company run enumeration.
- Changed the UI dismissal path so only failed-run call sites opt into the specialized resolve endpoint; arbitrary `run:*` dismissals use the generic inbox dismissal path.
- Added a regression test for cross-company run resolution returning 404 without creating a dismissal.

Local checks:
- `pnpm exec vitest run server/src/__tests__/inbox-dismissals.test.ts`
- `git diff --check`

serkanpektas (approved):
-

PR #4557, opened 2026-04-26
Fix disappearing issue comments

devinfoley:
## Thinking Path

> - Paperclip is a control plane for AI-agent companies, so issue detail pages are a primary surface for understanding agent work and human feedback.
> - The relevant subsystem here is the issue comments/chat experience across the React issue detail page and the server comment pagination API.
> - Long issue threads were only surfacing the newest page of comments at first render, which hid earlier human and agent messages behind extra pagination.
> - The first UI fix exposed that the descending cursor path on the server could also fail for older-page fetches, leaving the chat tab stuck on an infinite "Loading earlier comments..." state.
> - This needed to be addressed in both layers so the chat tab can surface earlier conversation history without manual recovery and without server errors.
> - This pull request auto-loads earlier comment pages in the issue detail chat view and fixes the descending cursor predicate used by issue comment pagination.
> - The benefit is that long-running issues like `PAPA-103` now show the missing conversation history near the top of the chat surface instead of hiding it or failing to load it.

## What Changed

- Auto-load earlier issue comment pages in the issue detail chat tab until the thread reaches a 150-comment cap or there are no older comments left.
- Add UI-side guard logic and regression coverage for optimistic issue comment pagination so the autoload behavior stops cleanly.
- Replace the raw SQL descending cursor predicate in `issueService.listComments` with typed Drizzle comparisons for the `(createdAt, id)` anchor tuple.
- Add a server regression test that paginates earlier comments in descending order from an anchor comment.
- Smoke-test the exact previously failing seeded `PAPA-103` cursor path on the isolated dev instance used for review.

## Verification

- `pnpm --filter @paperclipai/server exec vitest run src/__tests__/issues-service.test.ts`
- Manual smoke against seeded `PAPA-103` data on the isolated dev server:
  - `GET /api/issues/PAPA-103/comments?order=desc&limit=50` returns `200`
  - `GET /api/issues/PAPA-103/comments?after=765d3609-edc6-4d11-a8fe-d466affbe85d&order=desc&limit=50` now returns `200` with 50 comments instead of `500`

## Risks

- Moderate UI/perf risk on very large threads because the chat tab now prefetches multiple earlier pages on mount; the cap is intentionally limited to 150 comments to bound that work.
- Low API risk because the server fix only changes the cursor predicate construction for anchor-based comment pagination, but any mistake there would affect older-comment paging order.

> I checked `ROADMAP.md` before opening this PR and this bug fix does not duplicate planned core work.

## Model Used

- OpenAI Codex coding agent in the Paperclip local adapter environment. The exact backend model ID and context window were not exposed in-session. Tool-assisted workflow included shell execution, git/GitHub CLI, local test execution, and targeted code edits.

## Checklist

### DIFF
--- a/server/src/__tests__/issues-service.test.ts
+++ b/server/src/__tests__/issues-service.test.ts
@@ -895,6 +895,64 @@ describeEmbeddedPostgres("issueService.list participantAgentId", () => {
     );
   });
 
+  it("paginates earlier comments in descending order from an anchor comment", async () => {
+    const companyId = randomUUID();
+    const issueId = randomUUID();
+    const firstCommentId = randomUUID();
+    const anchorCommentId = randomUUID();
+    const latestCommentId = randomUUID();
+
+    await db.insert(companies).values({
+      id: companyId,
+      name: "Paperclip",
+      issuePrefix: `T${companyId.replace(/-/g, "").slice(0, 6).toUpperCase()}`,
+      requireBoardApprovalForNewAgents: false,
+    });
+
+    await db.insert(issues).values({
+      id: issueId,
+      companyId,
+      title: "Paged comments issue",
+      status: "todo",
+      priority: "medium",
+    });
+
+    await db.insert(issueComments).values([
+      {
+        id: firstCommentId,
+        companyId,
+        issueId,
+        body: "First comment",
+        createdAt: new Date("2026-03-26T10:00:00.000Z"),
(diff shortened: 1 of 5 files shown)


devinfoley:
Gonna go ahead and merge even though 4/5 greptile, since the only issue is the PR description, which I have since fixed/updated

PR #4562, opened 2026-04-27
feat: auto-provision agent JWT secret in pnpm dev (#4543)

### LINKED ISSUE #4543: First-run heartbeat fails: "Agent authentication required" — no agent keys after UI onboarding
## Problem

After completing UI onboarding (create company → bootstrap first agent), the agent has no API keys. `GET /api/agents/{agent-id}/keys` returns `[]`.

When the agent's first heartbeat fires and the spawned `claude` subprocess tries to call back to the Paperclip API, every request returns `{"error":"Agent authentication required"}`.

## Repro

1. Fresh Paperclip dev server (no existing instance state)
2. Complete UI onboarding flow: create company, bootstrap first agent (claude_local adapter)
3. `curl http://127.0.0.1:<PORT>/api/agents/<agent-id>/keys` → `[]`
4. Manually trigger a heartbeat: `POST /api/agents/<id>/heartbeat/invoke` (with `wakeOnDemand: true`)
5. Run log shows the agent running bash commands like `curl ... /api/agents/me/inbox-lite` and getting `{"error":"Agent authentication required"}` from each
6. Agent retries with various auth schemes (`X-Paperclip-Agent-Id` header, etc.) — all fail

## Documented behavior (which is failing)

The `paperclip` skill's "Authentication" section says:

> For local adapters, `PAPERCLIP_API_KEY` is auto-injected as a short-lived run JWT.

In practice, the spawned subprocess either doesn't receive `PAPERCLIP_API_KEY` or the value isn't accepted by the API.

## Possible root causes

- The auto-injection requires a stored agent key as a base — there's none after UI onboarding
- OR the auto-injection path has a bug
- OR UI onboarding is supposed to create a default key but doesn't

## Suggested fix

During first-agent bootstrap in the UI flow, automatically create at least one default agent key. Or fix the run-JWT auto-injection so it doesn't require a stored key.

bradfeld:
## Summary

When `pnpm dev` boots the contributor flow, it doesn't set `PAPERCLIP_AGENT_JWT_SECRET` (the CLI's `paperclipai onboard` flow does, but contributors running `pnpm dev` directly don't go through that). Without the secret, `createLocalAgentJwt()` returns `null`, the local-adapter `claude-local` execute path can't inject `PAPERCLIP_API_KEY` into the spawned subprocess, and every heartbeat fails with `{"error":"Agent authentication required"}` from inside the agent's bash calls back to the API.

The boot banner already calls this out as `Agent JWT: missing (run pnpm paperclipai onboard)`, and the heartbeat warns `local agent jwt secret missing or invalid; running without injected PAPERCLIP_API_KEY` — but neither stops the broken flow, and a contributor's first heartbeat just silently fails.

This PR adds an ephemeral auto-provision in dev: if neither `PAPERCLIP_AGENT_JWT_SECRET` nor `BETTER_AUTH_SECRET` is set, the dev-runner generates a 32-byte secret on boot, sets it on `process.env` so the spawned server inherits it, logs a one-line notice, and flags the boot banner so contributors know it's not stable across restarts.

Closes #4543.

## Changes

- **New**: `server/src/dev-jwt-bootstrap.ts` — pure helper `ensureDevAgentJwtSecret(env)` that returns a discriminated union (`noop` / `generated`) describing what it did.
- **New**: `server/src/dev-jwt-bootstrap.test.ts` — 5 unit tests covering both no-op paths (PAPERCLIP_AGENT_JWT_SECRET set, BETTER_AUTH_SECRET set), the generated path, whitespace-only existing values, and per-call uniqueness.
- **Modified**: `scripts/dev-runner.ts` — calls the helper after the worktree env bootstrap, sets `PAPERCLIP_AGENT_JWT_EPHEMERAL=1` and prints a one-line notice when generating.
- **Modified**: `server/src/startup-banner.ts` — when `PAPERCLIP_AGENT_JWT_EPHEMERAL=1`, the banner shows `Agent JWT: ephemeral (this dev session)` instead of just `set`.

Production deployments are unchanged — `BETTER_AUTH_SECRET` (or an explicitly set `PAPERCLIP_AGENT_JWT_SECRET`) takes precedence and is detected as a no-op. The CLI `paperclipai onboard` flow is untouched.

## Test Plan

- [x] 5 new unit tests in `dev-jwt-bootstrap.test.ts` pass
- [x] `tsc -b` clean for `@paperclipai/server`
- [x] Manual: `pnpm dev` after killing all prior server processes shows
  - `[paperclip] Generated ephemeral PAPERCLIP_AGENT_JWT_SECRET for this dev session ...` in dev-runner output
  - `Agent JWT       ephemeral (this dev session)` (green) in the startup banner
- [x] Manual: spawned-server `process.env` chain (verified via `ps eww`) carries `PAPERCLIP_AGENT_JWT_SECRET` from dev-runner through pnpm → cross-env → tsx → server.
- [x] Manual end-to-end: triggered a heartbeat against an existing agent, run log showed
  - NO "local agent jwt secret missing or invalid" warning
  - 2/2 successful `curl /api/agents/me/inbox-lite` calls (vs. all-failures before this fix)

## Out of scope

- The CLI `paperclipai onboard` flow (already correct).
- `.paperclip/.env` not being read in main clones (separate issue).
- Per-agent OAuth/client keys (`/api/agents/{id}/keys`) — unrelated; the local-adapter run-JWT flow uses a server-wide signing secret, not per-agent stored keys.

## Notes

The helper writes to `process.env.PAPERCLIP_AGENT_JWT_SECRET` (used by `agent-auth-jwt.ts:jwtConfig()` line 29), keeping a single source of truth. The `PAPERCLIP_AGENT_JWT_EPHEMERAL=1` flag is purely informational for the banner — `agent-auth-jwt.ts` ignores it.

### DIFF
--- a/docs/superpowers/plans/2026-04-26-pap-3-dev-mode-jwt-secret.md
+++ b/docs/superpowers/plans/2026-04-26-pap-3-dev-mode-jwt-secret.md
@@ -1,0 +1,408 @@
+# PAP-3: Auto-provision PAPERCLIP_AGENT_JWT_SECRET in `pnpm dev`
+
+> **For agentic workers:** REQUIRED SUB-SKILL: Use superpowers:subagent-driven-development (recommended) or superpowers:executing-plans to implement this plan task-by-task. Steps use checkbox (`- [ ]`) syntax for tracking.
+
+**Goal:** When `pnpm dev` boots without `PAPERCLIP_AGENT_JWT_SECRET` (or `BETTER_AUTH_SECRET`) set in the environment, auto-generate an ephemeral random secret and set it on `process.env` so local-adapter agents can authenticate to the API. Update the boot banner to reflect the ephemeral status.
+
+**Architecture:** A small pure helper (`ensureDevAgentJwtSecret`) reads the two env-var names that `agent-auth-jwt.ts` accepts. If either is set, it's a no-op. If neither is set, it generates a 32-byte base64-url secret, writes it to `process.env.PAPERCLIP_AGENT_JWT_SECRET`, and returns a status object. The dev-runner script calls this helper at startup BEFORE forking the server child, so the spawned server inherits the secret. The boot banner shows one of three states: `set (env)`, `set (.env file)` — out of scope for main clones (see PAP-5), or `ephemeral (generated this session)`.
+
+**Tech Stack:** Node `crypto.randomBytes` for entropy, no new dependencies. Uses Vitest for tests, `tsx`-based dev-runner for the boot wiring.
+
+**Scope clarification:**
+
+After investigation, the three root causes listed in PAP-3 collapse to one:
+
+1. ❌ "Auto-injection requires a stored agent key as a base — there's none after UI onboarding" — wrong; `/api/agents/{id}/keys` is for OAuth client keys, unrelated to local-adapter run JWTs.
+2. ❌ "Auto-injection path has a bug" — wrong; `claude-local/src/server/execute.ts:244-245` injects `PAPERCLIP_API_KEY=authToken` correctly when `authToken` is non-null.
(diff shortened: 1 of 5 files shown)


serkanpektas (approved):
-

PR #4566, opened 2026-04-27
Fix GET /issues?identifier= filter returning same record regardless of input

bryangauvin:
## Problem

`GET /api/companies/:companyId/issues?identifier=RPAA-NNN` returned an arbitrary issue regardless of the `identifier` value passed. Every request was effectively an unfiltered list query.

**Root cause:** Two-layer gap:

1. `IssueFilters` (services/issues.ts) had no `identifier` field.
2. The route handler (routes/issues.ts) never passed `req.query.identifier` to `svc.list()`.

So the `identifier` param was silently dropped before reaching the SQL WHERE clause.

## Reproducer (before fix)

```
# Returns RPAA-650 regardless of what identifier you pass:
curl http://localhost:3100/api/companies/<cid>/issues?identifier=RPAA-869
curl http://localhost:3100/api/companies/<cid>/issues?identifier=RPAA-1
```

## Fix

- Add `identifier?: string` to `IssueFilters`
- Add `eq(issues.identifier, filters.identifier.toUpperCase())` condition in `list()`, following the same pattern as `originId`, `parentId`, etc.
- Pass `req.query.identifier as string | undefined` from route handler to `svc.list()`

## Tests

Two new embedded-Postgres tests in `issues-service.test.ts`:
- Exact-match: `?identifier=RPAA-100` returns only that issue from a two-issue fixture
- Case-insensitive: `?identifier=rpaa-300` matches `RPAA-300` in DB (`.toUpperCase()` normalisation)

### DIFF
--- a/server/src/__tests__/issues-service.test.ts
+++ b/server/src/__tests__/issues-service.test.ts
@@ -859,3 +859,96 @@ describeEmbeddedPostgres("issueService.create workspace inheritance", () => {
     });
   });
 });
+
+describeEmbeddedPostgres("issueService.list identifier filter", () => {
+  let db!: ReturnType<typeof createDb>;
+  let svc!: ReturnType<typeof issueService>;
+  let tempDb: Awaited<ReturnType<typeof startEmbeddedPostgresTestDatabase>> | null = null;
+
+  beforeAll(async () => {
+    tempDb = await startEmbeddedPostgresTestDatabase("paperclip-issues-identifier-");
+    db = createDb(tempDb.connectionString);
+    svc = issueService(db);
+  }, 20_000);
+
+  afterEach(async () => {
+    await db.delete(issueComments);
+    await db.delete(issueInboxArchives);
+    await db.delete(activityLog);
+    await db.delete(issues);
+    await db.delete(executionWorkspaces);
+    await db.delete(projectWorkspaces);
+    await db.delete(projects);
+    await db.delete(agents);
+    await db.delete(instanceSettings);
+    await db.delete(companies);
+  });
+
+  afterAll(async () => {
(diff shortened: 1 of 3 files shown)


serkanpektas (approved):
/

roshangautam:
Came across this while working on the sibling `goalId` filter gap on the same endpoint (#8961) — same root cause shape (`IssueFilters` missing a field, route handler not forwarding the query param), so I read through this one closely before deciding not to duplicate it.

This fix is correct and the approach is right (mirrors the existing `originId`/`parentId` pattern exactly, tests use the same embedded-postgres harness as the rest of the suite). It's just gone stale against `master` — I test-cherry-picked `371f311f` onto current `upstream/master` locally and it conflicts in exactly two spots:

- `server/src/services/issues.ts` — both hunks are pure line-number drift in the `IssueFilters` interface and the `list()` conditions block (unrelated fields added above/below in the interim), not a semantic conflict. The `eq(issues.identifier, filters.identifier.toUpperCase())` condition drops in cleanly once the offsets are fixed up.
- `server/src/routes/issues.ts` — same story, one-line offset in the filters object passed to `svc.list()`.

`server/src/__tests__/issues-service.test.ts` cherry-picks clean (auto-merged, no conflict).

@bryangauvin — if you rebase onto current `master` this should be a ~5 minute fixup (just re-apply the two one-line additions at their new offsets) and it'd be good to get this merged; the `identifier` filter has the exact same "silently ignored, returns an arbitrary row" failure mode as the `goalId` gap I just filed, and it's a real footgun for anyone doing single-issue lookups by identifier. Thanks for finding and fixing this — didn't want it to die on a routine rebase.

PR #4575, opened 2026-04-27
fix: stop blocked issues from auto-checkout on wake

bdleasure:
## Summary
- stop heartbeat auto-checkout from claiming issues still in `blocked` status
- add a regression test covering comment wakes on assigned blocked issues
- keep blocked issues blocked until they are explicitly resumed to `todo`

## Testing
- pnpm --dir /data/brainroad/paperclip exec vitest run server/src/__tests__/heartbeat-dependency-scheduling.test.ts


bdleasure:
Closing this PR — we're not maintaining a fork or upstreaming changes anymore. The fix is now carried locally in our build pipeline as a patch applied to a pinned upstream submodule. No upstream review/maintenance burden needed on either side.

Thanks for the project — happy to drop our local patch the day this kind of fix lands organically here.

PR #4595, opened 2026-04-27
fix(issues): support unassigned filter for backlog routing

starlein:
Ports the DAI-555 runtime hotfix into a tracked git branch (DAI-595).

## What changed
- `server/src/routes/issues.ts`
  - Parses `unassigned` query values (`true/false/1/0`)
  - Returns `400` for invalid `unassigned` values
  - Passes `unassigned` through to issue-service filters
  - Extends `IssueFilters` with `unassigned?: boolean`
  - Applies `assigneeAgentId IS NULL AND assigneeUserId IS NULL` when `unassigned === true`
- `server/src/__tests__/issues-service.test.ts`
  - Adds regression: `status=todo&unassigned=true` returns only unassigned todo issues in a mixed dataset
- Docs
  - `docs/api/issues.md` adds `unassigned` query parameter
  - `docs/guides/board-operator/managing-tasks.md` adds backlog-routing query guidance (`status=todo&unassigned=true`)

## Validation
- `pnpm --filter @paperclipai/server exec vitest run src/__tests__/issues-service.test.ts -t "returns only unassigned todo issues when unassigned=true is set"`

Refs: DAI-555, DAI-595


serkanpektas (changes requested):
-

PR #4605, opened 2026-04-27
Fix stale detached heartbeat run cleanup

boluwaji-deriv:
## Thinking Path

> - Paperclip orchestrates agent heartbeats and issue execution, so stale heartbeat ownership can block real work even after the original task is operationally done.
> - The relevant subsystem here is the heartbeat runtime recovery path that reaps orphaned local child processes and decides whether to retry, fail, or release issue ownership.
> - `MYC-79` exists because a detached heartbeat run for `MYC-76` stayed marked as alive long after useful execution had stopped, leaving `executionRunId` pinned.
> - The original fix taught `reapOrphanedRuns` to treat detached local processes that outlive the stale threshold as recoverable process-loss cases instead of leaving them stranded forever.
> - Review on `#4605` surfaced two gaps in that first pass: the stale-detached message could be inaccurate when only the process group survived, and the regression coverage did not fully exercise that branch.
> - This pull request now tightens that message path, expands the regression assertions, and adds the missing process-group-only detached-stale test so the stale-run recovery behavior is explicit and reviewable.
> - The benefit is a cleaner, better-specified self-healing path for detached heartbeat runs without leaving misleading diagnostics behind.

## What Changed

- Replaced the boolean `detachedStale` message flag with an explicit detached stale target so `buildProcessLossMessage(...)` can distinguish between a surviving detached child pid and a surviving detached process group.
- Updated the detached-local-pid stale-threshold regression test to assert the same liveness and retry metadata that the neighboring process-loss tests already verify.
- Added a regression test for the detached process-group-only case where the recorded pid is already gone but the descendant process group is still alive past the stale threshold.
- Addressed the PR template requirements directly in this PR body so the upstream review handoff matches repository policy.

## Verification

- `PATH=/opt/homebrew/bin:$PATH pnpm vitest run server/src/__tests__/heartbeat-process-recovery.test.ts`
- Result: `22 passed`

## Risks

- Low risk. The behavior change is limited to the detached-stale process-loss messaging branch and its associated regression coverage.
- The runtime effect remains the same for the existing stale-detached cleanup path: terminate the surviving local process resource and route through the standard fail-and-retry logic.

## Model Used

- OpenAI Codex running via `codex_local`
- Model: `gpt-5.4`
- Reasoning mode: default Codex agent reasoning with tool use, shell execution, and code editing

## Checklist

### DIFF
--- a/server/src/__tests__/heartbeat-process-recovery.test.ts
+++ b/server/src/__tests__/heartbeat-process-recovery.test.ts
@@ -767,6 +767,52 @@ describeEmbeddedPostgres("heartbeat orphaned process recovery", () => {
     expect(wakeup?.status).toBe("claimed");
   });
 
+  it("reaps a detached local pid after the stale threshold expires", async () => {
+    const child = spawnAliveProcess();
+    childProcesses.add(child);
+    expect(child.pid).toBeTypeOf("number");
+
+    const { agentId, runId, issueId } = await seedRunFixture({
+      processPid: child.pid ?? null,
+      runErrorCode: "process_detached",
+      runError: `Lost in-memory process handle, but child pid ${child.pid} is still alive`,
+    });
+    const heartbeat = heartbeatService(db);
+
+    await db
+      .update(heartbeatRuns)
+      .set({
+        updatedAt: new Date("2026-03-18T23:50:00.000Z"),
+      })
+      .where(eq(heartbeatRuns.id, runId));
+
+    const result = await heartbeat.reapOrphanedRuns({ staleThresholdMs: 5 * 60 * 1000 });
+    expect(result.reaped).toBe(1);
+    expect(result.runIds).toEqual([runId]);
+    expect(await waitForPidExit(child.pid ?? -1, 2_000)).toBe(true);
+
+    const runs = await db
+      .select()
+      .from(heartbeatRuns)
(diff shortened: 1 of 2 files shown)


boluwaji-deriv:
Addressed the Greptile follow-up on top of `defa1e72` in commit `a203b865`.

What changed:
- fixed the detached stale-run message so it distinguishes pid-only vs process-group-only termination
- added the missing liveness / retry assertions to the detached stale regression
- added a second regression for the process-group-only detached case
- updated the PR body to match the repository template sections

Verification:
- `PATH=/opt/homebrew/bin:$PATH pnpm vitest run server/src/__tests__/heartbeat-process-recovery.test.ts`
- result: `22 passed`

boluwaji-deriv:
Addressed the earlier follow-up items on top of `a203b86579567151c18294bff0c94b664cfe9726`.

- Fixed the detached stale-run message so it distinguishes child-pid cleanup from process-group cleanup.
- Added the missing liveness / retry assertions to the detached stale regression coverage.
- Added a second regression test for the process-group-only detached case.
- Updated the PR body to follow the required template sections.

Local verification:
- `PATH=/opt/homebrew/bin:$PATH pnpm vitest run server/src/__tests__/heartbeat-process-recovery.test.ts`
- `22 passed`

serkanpektas (approved):
-

PR #4662, opened 2026-04-28
fix: agent instructions invisible when CEO creates agents via heartbeat

jackz-jones:
## Thinking Path

> - The CEO agent's heartbeat cycle can create instruction files for new agents, but previously wrote them to a name-based directory path (`agents/{agentName}/instructions`) instead of the UUID-based path (`agents/{agentId}/instructions`) that the instructions resolution system expects.
> - This made newly created agents' instructions invisible in the UI.
> - An initial approach added inline migration logic inside `recoverManagedBundleState()` to auto-migrate name-based directories on every resolution cycle — but this introduced code complexity in the hot path, added filesystem stat calls on every `getBundle` invocation, and had a name-collision risk (two agents sharing the same display name would produce the same `nameBasedPath`).
> - A code review flagged the name-collision issue and questioned whether inline migration belongs in the resolution code at all: the CEO template fix already prevents new name-based writes at the source, making migration a one-time historical cleanup task.
> - This PR moves the migration to a standalone script, removing all inline migration logic from the hot path while preserving the ability to recover existing broken state.

## What Changed

- **CEO template (AGENTS.md)**: Replaced the permissive statement with an explicit prohibition against writing instruction files using agent display names, pointing the CEO to the correct API surface (`PATCH /api/agents/{agentId}/instructions-path` and the `paperclip-create-agent` skill).
- **Removed inline migration logic** from `agent-instructions.ts`: Deleted `MIGRATION_MARKER_FILE` constant, `migrateNameBasedInstructionsIfFound()` function, and the migration call inside `recoverManagedBundleState()`.
- **Added standalone migration script** (`scripts/migrate-name-based-agent-instructions.ts`): A one-time CLI tool that:
  - Reads all agents from the database and locates name-based instruction directories.
  - Copies files to the correct UUID-based path (dry-run by default; `--apply` to persist).
  - Detects name collisions (multiple agents sharing the same display name) and skips ambiguous cases with a warning.
  - Never overwrites existing UUID-based directories.
  - Does not delete source directories after migration — requires manual cleanup after verification.
- **Added migration documentation** (`scripts/migrate-name-based-agent-instructions.md`): Covers what, when, how, usage commands, and post-migration steps.
- **Removed migration-related tests** from the agent-instructions test file (no longer applicable).

## Verification

- `pnpm exec vitest run server/src/services/agent-instructions.test.ts` — all 8 tests pass (down from 10, with the 2 migration-specific tests removed).
- Dry-run the migration script against a test instance:
  ```bash
  DATABASE_URL="postgres://…" PAPERCLIP_HOME="…" PAPERCLIP_INSTANCE_ID="…" \
    npx tsx scripts/migrate-name-based-agent-instructions.ts
  ```

## Risks

- **Low risk**: The inline migration logic removal has no effect on agents that were already migrated (their UUID-based directories exist and are used directly). Agents that were never affected also continue to work unchanged.
- **Migration gap**: Agents with existing name-based directories that have not yet been migrated will NOT be auto-recovered on the next resolution cycle — the operator must run the standalone script explicitly. This is acceptable because the CEO template fix prevents new name-based writes, so the affected set is finite and historical.
- **Name collisions require manual resolution**: The script warns but does not auto-migrate when multiple agents share the same display name. The operator must inspect and manually distribute files to the correct UUID-based paths.

## Checklist


serkanpektas (changes requested):
/

PR #4669, opened 2026-04-28
fix(security): catch malformed JSON + cap body / title size (DoS hardening)

Qusaitr:
## Summary

Two security fixes surfaced by black-box monkey-fuzz of a self-hosted paperclipai instance on 2026-04-28. Both are reachable post-auth (or anonymously in `local_trusted` mode).

### S-CRIT-7 — malformed JSON body returns 500 instead of 400
`server/src/middleware/error-handler.ts` doesn't catch the `SyntaxError` that `express.json()` raises for invalid JSON, so callers get `{"error":"Internal server error"}` (HTTP 500) for a client mistake. Pollutes telemetry, confuses API consumers.

**Repro:** `curl -X POST -H 'content-type: application/json' .../api/companies/$ID/issues -d '{not-json'` → 500.

### S-CRIT-8 — large request body hangs / crashes the server (DoS)
`express.json({ limit: "10mb" })` plus an uncapped `title: z.string().min(1)` in `createIssueSchema` lets a single request hang the event loop:

- 100 KB title — paperclipai becomes unresponsive (>8 s)
- 1 MB title or ~50 K unicode characters — paperclipai crashes; my watchdog respawn takes ~3 min, dropping all in-flight work

Any authenticated peer can take the dashboard offline with one POST.

## Changes

- `server/src/middleware/error-handler.ts`
  - `SyntaxError` (with `body`) → 400 `{"error":"Malformed JSON body","code":"invalid_json"}`
  - `entity.too.large` → 413 `{"error":"Payload too large","code":"entity_too_large"}`
- `server/src/app.ts`
  - `express.json` limit: `10mb` → `1mb` (with TODO that import/export endpoints needing >1 MiB should mount a dedicated parser on their router)
- `packages/shared/src/validators/issue.ts`
  - `createIssueSchema.title` → `.max(500)`
  - `createIssueSchema.description` → `multilineTextSchema.pipe(z.string().max(65_536))` (keeps line-break normalization, adds 64 KiB cap)

## Verification

Live runtime test on srv1311600 against the patched server:

| test | before | after |
|---|---|---|
| malformed JSON | 500 | **400 in <50 ms** |
| 100 KB title | hangs/crashes | **400 in 342 ms** with Zod `too_big` |
| 1.5 MB body | crashes | **413 in 221 ms** |
| valid POST | 201 | 201 (no regression) |

## Test plan

- [ ] CI green
- [ ] `curl -d '{not-json' /api/.../issues` returns 400 not 500
- [ ] `title: \"A\".repeat(1000)` returns 400 with Zod detail
- [ ] `title: \"A\".repeat(2_000_000)` returns 413 quickly (no event-loop block)
- [ ] Existing valid issue creation still returns 201
- [ ] Existing description with `\\n` line breaks still normalizes (multilineTextSchema preserved)

### DIFF
--- a/packages/shared/src/validators/issue.ts
+++ b/packages/shared/src/validators/issue.ts
@@ -133,9 +133,13 @@ export const createIssueSchema = z.object({
   // S-CRIT-8 (2026-04-28 fuzz): unbounded title hangs the server on huge
   // inputs. 500 chars is plenty for an issue title (UI truncates at ~80).
   title: z.string().min(1).max(500),
-  // Keep upstream's multilineTextSchema (line-break normalization) but cap
-  // length at 64 KiB. Anything bigger should attach as a file.
-  description: multilineTextSchema.pipe(z.string().max(65_536)).optional().nullable(),
+  // Keep upstream's multilineTextSchema (line-break normalization) and cap
+  // length at 1 MiB. The DoS we patched (S-CRIT-8) was an unbounded TITLE that
+  // hung the event loop in cosmetic rendering paths; descriptions are stored
+  // as text and not rendered inline, so 1 MiB is safely beyond the
+  // company-import/export e2e test's 168 KiB usage and still bounded against
+  // multi-megabyte abuse.
+  description: multilineTextSchema.pipe(z.string().max(1_048_576)).optional().nullable(),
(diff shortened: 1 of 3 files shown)


serkanpektas (approved):
/

serkanpektas (approved):
/

PR #4518, opened 2026-04-25
fix(recovery): add 24-hour rolling continuation run cap per issue

supertaz:
## Thinking Path

> - Paperclip orchestrates AI agents that work on issues in a continuous cycle of run → check-in → next run
> - When a run ends with `retryReason=issue_continuation_needed`, the recovery reconciler re-queues another run
> - The consecutive-cycle cap (#4509) catches tight back-to-back loops where every run immediately produces a noop continuation
> - Slow-drip loops escape the consecutive cap: an issue that makes partial progress occasionally resets the consecutive streak but accumulates many continuation runs over hours
> - A 24-hour rolling count captures this pattern regardless of inter-run spacing
> - This PR adds a rolling 24-hour continuation run cap: if an issue accumulates 24 or more continuation runs in the past 24 hours without resolving, the reconciler escalates to `blocked`
> - The benefit is that slow-drip loops that evade the consecutive cap are also detected and stopped

## What Changed

- `server/src/services/recovery/service.ts`:
  - New `DAILY_CONTINUATION_CAP = 24` and `DAILY_CONTINUATION_WINDOW_MS = 24 * 60 * 60 * 1000` constants
  - New helper counts `issue_continuation_needed` runs (any status) within the 24-hour window after `issue.updatedAt`
  - `reconcileStrandedAssignedIssues`: before enqueuing a continuation run, checks the daily cap; on exhaustion escalates to `blocked`, inserts a `heartbeat_run_watchdog_decisions` row (`decision: "rate_limited"`), adds `dailyCapTripped` to the result counter
  - Unlike the rate-limit (PR #4517), this escalation does NOT pause the agent — only the issue is blocked
- `server/src/__tests__/daily-continuation-cap.test.ts` — 2 new embedded-Postgres tests: (1) 24 runs in window → escalation + watchdog record, (2) 23 runs → continuation re-queued normally

## Summary

Adds a rolling 24-hour continuation run cap per issue to the recovery reconciler. If an issue accumulates 24 or more `issue_continuation_needed` runs in the past 24 hours without resolving, the reconciler escalates it to `blocked` rather than enqueuing another run.

This complements the consecutive-cycle cap (#4509): that cap catches tight back-to-back loops; this one catches slow-drip loops (e.g., one run every hour) that the consecutive check misses because the issue occasionally makes partial progress.

**Behavior:**
[...]

PR #4517, opened 2026-04-25
fix(recovery): add per-issue recovery enqueue rate limiter

supertaz:
## Thinking Path

> - Paperclip orchestrates AI agents that run in cycles; each cycle can signal the need for a continuation run
> - The recovery reconciler scans for in-progress issues with no live execution and re-queues recovery runs
> - With agent parallelism > 1, a paced loop (each run completes quickly) keeps the queue empty, so queue-depth runaway detectors never see a spike
> - A per-agent rate detector exists but counts across all issues — a single-issue paced loop dilutes the signal
> - This PR adds a per-(agent, issue) dimension: if >= 5 recovery runs for the same agent+issue pair appear within a 5-minute window, the reconciler trips a rate limit
> - The rate limit action is: escalate the issue to `blocked`, pause the agent (so no new runs fire), and write a watchdog decision record for operator visibility
> - The DB-backed window (querying `heartbeat_runs.created_at`) works correctly after server restarts, unlike an in-memory map

## What Changed

- `server/src/services/recovery/service.ts`:
  - New module-level constants: `ISSUE_RECOVERY_RATE_WINDOW_MS` and `ISSUE_RECOVERY_RATE_CAP`
  - New `countIssueRecoveryEnqueuesInWindow` — counts recovery runs for the agent+issue within the window
  - New `tripIssueRecoveryRateLimit` — escalates issue to `blocked`, pauses agent via direct DB update, and inserts a `heartbeat_run_watchdog_decisions` watchdog decision record
  - `reconcileStrandedAssignedIssues` result type: new `rateLimitTripped` counter
  - Both `reconcileStrandedAssignedIssues` call sites (assignment_recovery and continuation_needed) now check the rate cap before enqueuing

- `server/src/__tests__/per-issue-recovery-rate-limit.test.ts`: 2 new embedded-Postgres tests — one verifying the rate limit fires (5 runs in window → blocked+paused+watchdog record), one verifying it does not fire below the cap (4 runs → continuation re-queued normally)

## Verification

- `pnpm typecheck` — clean
> paperclip@ typecheck /home/taz/Development/paperclip
> pnpm run preflight:workspace-links && pnpm -r typecheck

> paperclip@ preflight:workspace-links /home/taz/Development/paperclip
> node cli/node_modules/tsx/dist/cli.mjs scripts/ensure-workspace-package-links.ts

Scope: 21 of 22 workspace projects
packages/shared typecheck$ tsc --noEmit
[...]

PR #4510, opened 2026-04-25
feat(runs): record cancel_source on heartbeat_runs to distinguish cancellation causes

supertaz:
## Thinking Path

> - Paperclip orchestrates AI agents; every agent action produces a heartbeat run record
> - When a run is cancelled, the run record is updated with `error_code = 'cancelled'`
> - The cancellation cause is not recorded: user stop, budget enforcement, and reconciler cleanup all produce identical records
> - Operators cannot distinguish intentional user stops from automated enforcement decisions in the database
> - This PR adds a `cancel_source` column to `heartbeat_runs` and wires each cancellation call site to record its cause
> - The benefit is that `cancel_source` gives operators (and future analytics) a clear signal for why any run was cancelled

## What Changed

- `packages/db/src/migrations/0071_heartbeat_run_cancel_source.sql` — `ALTER TABLE heartbeat_runs ADD COLUMN IF NOT EXISTS cancel_source text`
- `packages/db/src/migrations/meta/_journal.json` — journal entry for migration 0071
- `packages/db/src/schema/heartbeat_runs.ts` — `cancelSource` Drizzle column
- `packages/shared/src/types/heartbeat.ts` — `cancelSource: string | null` on `HeartbeatRun`
- `server/src/services/heartbeat.ts` — optional `cancelSource` param on `cancelRunInternal` / `cancelActiveForAgentInternal`; all reconciler-internal call sites wired
- `server/src/routes/agents.ts` — 3 explicit user-cancel call sites set `user_initiated`
- `server/src/routes/issues.ts` — interrupt and issue-status-cancel paths; board actor → `user_initiated`, agent actor → `system`
- `server/src/routes/issue-tree-control.ts` — tree-hold run interruption; derived from actor type
- `server/src/__tests__/issue-tree-control-routes.test.ts` — updated `toHaveBeenCalledWith` expectations
- `server/src/__tests__/issue-comment-reopen-routes.test.ts` — updated `toHaveBeenCalledWith` expectations

## Summary

All run cancellations currently land with `error_code = 'cancelled'`, which conflates causes that have different operational implications: a user clicking stop, a budget limit being reached, and the reconciler clearing a stale queued run are all indistinguishable in the database.

This PR adds a nullable `cancel_source` text column to `heartbeat_runs` that records *why* a run was cancelled. Values used at each call site:

| Value | Set by |
|---|---|
| `user_initiated` | Explicit board/user action: UI cancel button, issue interrupt, agent pause/terminate |
[...]

PR #4509, opened 2026-04-25
fix(recovery): cap consecutive issue_continuation_needed cycles at 3

supertaz:
## Thinking Path

> - Paperclip orchestrates AI agents that work on issues; agents operate in a cycle of run → check-in → next run
> - When an agent run succeeds but leaves the issue needing more work, the run context carries `retryReason=issue_continuation_needed`
> - The recovery reconciler periodically scans for in-progress issues with no live execution path and re-queues a continuation run
> - A run that produces no progress (noop turn) succeeds immediately, the reconciler sees the issue stranded again on next tick, and re-queues another continuation
> - With agent parallelism > 1, these short runs slot into freed concurrency slots fast enough that queue-depth runaway detectors never see a spike
> - This PR adds a cap: after 3 consecutive succeeded continuation runs on the same issue, the reconciler escalates to `blocked` instead of re-queuing
> - Issues with genuine progress are unaffected — any run that is not a succeeded continuation resets the streak

## What Changed

- `recoveryService` in `server/src/services/recovery/service.ts`: new `CONTINUATION_CYCLE_CAP = 3` constant and `hasExhaustedConsecutiveContinuationCycles(companyId, issueId)` helper — queries the most recent `CAP` runs for the issue and returns true only if all of them are `succeeded` with `retryReason=issue_continuation_needed`
- `reconcileStrandedAssignedIssues`: before calling `enqueueStrandedIssueRecovery` for continuation, checks `hasExhaustedConsecutiveContinuationCycles`; on exhaustion, calls `escalateStrandedAssignedIssue` (moves issue to `blocked`, posts an operator-visible comment) and skips re-queue
- `server/src/__tests__/continuation-cycle-cap.test.ts`: 2 new embedded-Postgres tests — one verifying escalation after 3 consecutive succeeded continuations, one verifying re-queue continues when fewer than 3 exist

## Verification

- `pnpm typecheck` — clean
- `npx vitest run --project "@paperclipai/server" server/src/__tests__/continuation-cycle-cap.test.ts` — 2/2 pass

## Risks

Low. The cap (3) is deliberately conservative — genuine multi-turn work always has at least some run-to-run progress signal (issue status transitions, description edits, or blocking sub-issues) that interrupts the streak. The only case this fires on is a pure noop loop where every run immediately succeeds as a continuation with no side effects. Operators can resume blocked issues normally.

## Model Used

- Provider: Anthropic
- Model: Claude Sonnet 4.6 (`claude-sonnet-4-6`)
[...]

PR #4671, opened 2026-04-28
fix(recovery): generic guard against recovery-of-recovery cascade (STO-259)

jooneyp:
## Thinking Path

> - The recovery subsystem watches for "stranded" assigned issues — assigned but with no live execution path — and creates a `stranded_issue_recovery` issue so a human (or recovery owner) can intervene.
> - Incident 2026-04-25: this watchdog created **17,786 self-replicating** "Recover stalled issue …" rows in ~3h, blowing the STO counter from ~200 to 18,047. Root cause was that recovery-origin issues (themselves stranded) re-entered the same watchdog and spawned more recovery issues — recovery-of-recovery cascade.
> - The current in-tree guard is a single helper, `isStrandedIssueRecoveryIssue`, which only matches the `strandedIssueRecovery` origin kind. The other recovery origins (`issue_graph_liveness_escalation`, `stale_active_run_evaluation`) don't currently feed the same watchdog, but the safety guarantee depends on that contingency holding forever.
> - This pull request adds a generic `isRecoveryOriginKind(originKind)` helper plus an explicit `RECOVERY_ORIGIN_KIND_VALUES` registry and uses it as a defense-in-depth gate inside `ensureStrandedIssueRecoveryIssue`. The behavior is invisible under current call graphs — but adding any future recovery origin to `RECOVERY_ORIGIN_KINDS` will inherit cascade protection automatically, instead of silently bypassing it.
> - The benefit is a single source of truth ("which origin kinds count as recovery artifacts?") that the cascade guard mechanically consumes, removing a class of foot-gun where a future contributor adds a recovery origin and forgets to extend the guard.

## What Changed

- `server/src/services/recovery/origins.ts`: add `RECOVERY_ORIGIN_KIND_VALUES` (frozen list of every recovery origin) and `isRecoveryOriginKind(originKind)` predicate.
- `server/src/services/recovery/index.ts`: re-export both new symbols alongside the existing `isStrandedIssueRecoveryOriginKind`.
- `server/src/services/recovery/service.ts`: in `ensureStrandedIssueRecoveryIssue`, refuse to spawn a stranded-recovery issue when the source's `originKind` matches **any** entry in `RECOVERY_ORIGIN_KIND_VALUES` (not only the stranded kind).
- `server/src/__tests__/recovery-classifiers.test.ts`: assert `RECOVERY_ORIGIN_KIND_VALUES` stays in sync with `Object.values(RECOVERY_ORIGIN_KINDS)`, so future origin additions cannot silently bypass the guard.

## Verification

```sh
cd server && pnpm vitest run src/__tests__/recovery-classifiers.test.ts
# 5/5 pass (incl. new "flags every recovery origin" test)
```

Behavioral check (no change expected today):

- Existing `isStrandedIssueRecoveryIssue` already short-circuits the only known cascade hot path. New `isRecoveryOriginKind` gate inside `ensureStrandedIssueRecoveryIssue` is redundant **today** — that's the point. Future-proofing only.

Failure mode the test catches: someone adds `RECOVERY_ORIGIN_KINDS.fooBar = "foo_bar"` without appending to `RECOVERY_ORIGIN_KIND_VALUES` → the sync assertion fails in CI before the regression can ship.

## Risks

Low.

- The new predicate's value set is identical to `Object.values(RECOVERY_ORIGIN_KINDS)`, so under current callers the gate matches the same issues as before.
- No schema, no migration, no UI surface change.
- Does not address the **other** STO-259 requirements (per-source cap, recoveryAttempts counter, cap-reached emergency notification). Those are deferred and tracked in STO-259's body — this PR is the smallest in-tree hardening that stands alone.
- Does not stop the cascade currently observed at 2026-04-28T13:09–13:36 (12 stranded_recovery rows for STO-245/259). That cascade goes through a different path (live retries flipping status to in_progress momentarily) — separate fix, not in scope.

## Model Used

Claude Opus 4-7 (Anthropic, ~1M context, extended-thinking off). Used for code drafting, test design, and PR composition. Code reviewed by author before commit.

## Checklist

- [x] I have listed concrete changes in "What Changed"
- [x] I have explained how a reviewer can verify this works
- [x] I have identified risks (or stated low risk)
- [x] I have specified the AI model used (or stated "None — human-authored")
- [x] Tests pass locally
- [x] Typecheck passes locally for the changed package
- [x] No DB migrations introduced
- [x] PR is scoped to a single logical change

### DIFF
--- a/server/src/__tests__/recovery-classifiers.test.ts
+++ b/server/src/__tests__/recovery-classifiers.test.ts
@@ -4,12 +4,14 @@ import { decideRunLivenessContinuation as decideRunLivenessContinuationCompat } 
 import {
   RECOVERY_KEY_PREFIXES,
   RECOVERY_ORIGIN_KINDS,
+  RECOVERY_ORIGIN_KIND_VALUES,
   RECOVERY_REASON_KINDS,
   buildIssueGraphLivenessIncidentKey,
   buildIssueGraphLivenessLeafKey,
   buildRunLivenessContinuationIdempotencyKey,
   classifyIssueGraphLiveness,
   decideRunLivenessContinuation,
+  isRecoveryOriginKind,
   isStrandedIssueRecoveryOriginKind,
   parseIssueGraphLivenessIncidentKey,
 } from "../services/recovery/index.ts";
@@ -150,5 +152,22 @@ describe("recovery classifier boundary", () => {
     expect(isStrandedIssueRecoveryOriginKind("harness_liveness_escalation")).toBe(false);
     expect(isStrandedIssueRecoveryOriginKind("manual")).toBe(false);
     expect(isStrandedIssueRecoveryOriginKind(null)).toBe(false);
+  });
+
+  it("flags every recovery origin so the cascade guard catches all watchdog kinds", () => {
+    // Cascade guard for incident 2026-04-25: any origin kind that marks an
+    // issue as a recovery artifact must never itself spawn another recovery
+    // issue. Keep this set in sync with RECOVERY_ORIGIN_KINDS — adding a new
+    // recovery origin without flipping this assertion will break the guard.
+    expect([...RECOVERY_ORIGIN_KIND_VALUES].sort()).toEqual(
+      Object.values(RECOVERY_ORIGIN_KINDS).sort(),
+    );
+    for (const kind of RECOVERY_ORIGIN_KIND_VALUES) {
+      expect(isRecoveryOriginKind(kind)).toBe(true);
+    }
+    expect(isRecoveryOriginKind("manual")).toBe(false);
+    expect(isRecoveryOriginKind(null)).toBe(false);
+    expect(isRecoveryOriginKind(undefined)).toBe(false);
+    expect(isRecoveryOriginKind("")).toBe(false);
   });
 });
--- a/server/src/services/recovery/index.ts
+++ b/server/src/services/recovery/index.ts
@@ -1,9 +1,11 @@
 export {
   RECOVERY_KEY_PREFIXES,
   RECOVERY_ORIGIN_KINDS,
+  RECOVERY_ORIGIN_KIND_VALUES,
   RECOVERY_REASON_KINDS,
   buildIssueGraphLivenessIncidentKey,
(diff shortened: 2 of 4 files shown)


supertaz:
There are already fixes for the continuation loop issue in flight, which may or may not affect your PR. @cryppadotta will be pushing some changes to fix the continuation loops today, and I've got a few PRs open on the subject that will be reconciled with his fixes once they're pushed. If you want to watch them for changes and reconcile your work against them (it appears your work in this PR is complementary), they are listed below.

Codex review of this issue vs. those listed below:

For #4671: it’s not a pure duplicate of your open PRs. It overlaps with the same recovery-cascade area as your recovery PRs, especially #4459/#4509/#4517/#4518, but it adds a narrower generic guard your PRs don’t currently have: isRecoveryOriginKind() blocks creating stranded-recovery issues for any recovery-origin issue, not just stranded_issue_recovery.

So I’d classify #4671 as complementary/new defense-in-depth, not a replacement. It should probably be reconciled with your recovery stack so the generic origin-kind helper lands once.

Core retry/recovery loop fixes
#4518 — 24-hour rolling continuation run cap per issue
#4517 — per-issue recovery enqueue rate limiter
#4509 — cap consecutive issue_continuation_needed cycles at 3
#4506 — suppress continuation recovery after user-initiated cancel
#4459 — escalate stranded issue when recovery retry succeeds without restoring an execution path
#4372 — prevent deferred-wake resurrection loop
#4362 — prevent wakeup cascades, anti-cancel-retry, skip timer when locked
#4356 — pause enforcement at enqueue sites; not only recovery, but directly affects retry/recovery enqueue behavior

Supporting / adjacent
#4510 — records cancel_source, important because #4506 needs to distinguish user-initiated cancels from system/budget cancels
#4507 — dedupes issue_children_completed wakeups; prevents duplicate wake-trigger storms
#4503 — atomic pause-status guard; protects pause behavior that interacts with recovery suppression
#4379 — queued-run visibility UI; observability for the above, not a recovery logic fix
#4364 — routine archive trigger disable/restore; loop-adjacent for routines, not heartbeat recovery core

PR #3341, opened 2026-04-10
fix(gemini-local): skip benign stderr lines in fallback error message

iws17:
## Thinking Path

> - Paperclip orchestrates AI agents for zero-human companies and exposes per-agent failure reasons in the dashboard activity log
> - The gemini_local adapter surfaces the first non-empty line of stderr as the fallback error message when the Gemini CLI exits non-zero
> - The Gemini CLI prints informational noise on every run regardless of success/failure — the YOLO mode banner, plus an intermittent \`Failed to fetch admin controls\` ETIMEDOUT warning from \`cloudcode-pa.googleapis.com\` (Gemini CLI #2323, IPv6 DNS issue during refreshAuth)
> - Because the YOLO banner is always the first stderr line, \`firstNonEmptyLine\` grabs it and hides the real error
> - Operators see \`YOLO mode is enabled. All tool calls will be automatically approved.\` in the Paperclip activity log as the failure reason — confusing and makes the actual root cause invisible
> - This PR adds a tiny benign-pattern skip list so \`firstNonEmptyLine\` returns the first line that is NOT one of these known-informational messages
> - The benefit is that operators see the real error (rate limit, ETIMEDOUT, quota, etc.) directly in the Paperclip dashboard instead of having to dig into raw stderr logs

## What Changed

- \`packages/adapters/gemini-local/src/server/utils.ts\` — added \`BENIGN_STDERR_PATTERNS\` and updated \`firstNonEmptyLine\` to skip lines matching those patterns
  - Pattern 1: \`/^YOLO mode is enabled/i\` — the CLI YOLO banner printed on every \`--approval-mode=yolo\` run
  - Pattern 2: \`/cloudcode-pa\\.googleapis\\.com/i\` — the intermittent admin-controls fetch warning
- \`packages/adapters/gemini-local/src/server/utils.test.ts\` — new vitest suite covering the benign-skip behavior with 5 test cases

## Verification

Real failing run captured in production (Chief Analyst agent, gemini-3.1-pro-preview, April 10 2026):

**Before:**
\`\`\`
errorMessage: \"YOLO mode is enabled. All tool calls will be automatically approved.\"
errorCode: adapter_failed
\`\`\`

Dashboard showed the YOLO banner as the failure reason — no indication of the real issue.

**After** (with this patch applied):
\`\`\`
errorMessage: \"An unexpected critical error occurred: ETIMEDOUT\"
errorCode: adapter_failed
\`\`\`

[...]

PR #4690, opened 2026-04-28
fix(gemini-local): skip benign stderr lines in fallback error message

iws17:
## Summary

The Gemini CLI prints informational banners to stderr on every run — most notably the YOLO mode banner and intermittent `Failed to fetch admin controls` warnings from cloudcode-pa.googleapis.com ([Gemini CLI #2323](https://github.com/anthropics/claude-code/issues/2323)).

When the CLI exits non-zero and no structured error is available, `firstNonEmptyLine(stderr)` is used as the fallback error message — surfacing these benign lines instead of the real error. This fix filters known-benign patterns before selecting the first line.

- Patterns are anchored tightly to avoid suppressing novel errors that happen to mention the same tokens (regression-tested).
- Adds a package-level `vitest.config.ts` so the new tests run in CI.
- 9 unit tests covering the filter, the full fallback chain, and a regression guard.

Supersedes #3341 (same fix, rebased to current master to resolve conflicts).

## Test plan

- [x] `pnpm exec vitest run` in `packages/adapters/gemini-local` — 12 pass / 0 fail
- [ ] CI green on this branch

### DIFF
--- a/packages/adapters/gemini-local/src/server/execute.ts
+++ b/packages/adapters/gemini-local/src/server/execute.ts
@@ -529,7 +529,7 @@ export async function execute(ctx: AdapterExecutionContext): Promise<AdapterExec
       } as Record<string, unknown>)
       : null;
     const parsedError = typeof attempt.parsed.errorMessage === "string" ? attempt.parsed.errorMessage.trim() : "";
-    const stderrLine = firstNonEmptyLine(attempt.proc.stderr);
+    const stderrLine = firstNonEmptyLine(attempt.proc.stderr, true);
     const structuredFailure = attempt.parsed.resultEvent
       ? describeGeminiFailure(attempt.parsed.resultEvent)
       : null;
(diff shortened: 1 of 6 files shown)


iws17:
Addressed both review points in the latest push:

**1. Benign filter scope (Copilot):** Added `skipBenign` parameter to `firstNonEmptyLine` (default `false`). Callers now opt in explicitly:
- `execute.ts`: `firstNonEmptyLine(stderr, true)` — filters benign banners
- `test.ts`: `firstNonEmptyLine(stderr, true)` for stderr, `firstNonEmptyLine(stdout)` for stdout (no filtering)

This means stdout content matching the YOLO/admin-controls patterns is preserved — only stderr gets filtered. New test case verifies this explicitly.

**2. CI registration (Copilot):** Added `@paperclipai/adapter-gemini-local` to `nonServerProjects` in `scripts/run-vitest-stable.mjs` so the 14 tests actually run in CI.

```
$ pnpm exec vitest run
PASS (14) FAIL (0)
```

PR #3309, opened 2026-04-10
fix(ui): guard against null metadata in IssueChatThread

iws17:
## Thinking Path

> - Paperclip orchestrates AI agents for zero-human companies
> - Humans oversee agent work through the issue chat UI, which renders comments, run transcripts, and system messages
> - Each chat message carries a `metadata.custom` object with routing info (author, agent ID, anchor ID, etc.)
> - But some messages arrive with `null`/`undefined` metadata — during streaming transitions, heartbeat partial data, or system messages without custom metadata
> - The code accesses `message.metadata.custom` without null-checking, throwing a `TypeError`
> - The `IssueChatErrorBoundary` (added in `a4b05d88`) catches the crash and shows skeleton placeholders instead of the rich chat UI
> - This PR adds optional chaining and nullish coalescing at all four access sites so messages with missing metadata render gracefully instead of crashing
> - The benefit is users see their actual chat messages instead of permanent loading skeletons

## What Changed

- Added `?.` optional chaining and `?? {}` nullish coalescing to `message.metadata.custom` access in `IssueChatChainOfThought` (line 516)
- Same guard in `IssueChatUserMessage` (line 873)
- Same guard in `IssueChatAssistantMessageContent` (line 976)
- Same guard in `IssueChatSystemMessage` (line 1402)

All four sites changed from:
```typescript
const custom = message.metadata.custom as Record<string, unknown>;
```
To:
```typescript
const custom = (message.metadata?.custom ?? {}) as Record<string, unknown>;
```

## Verification

1. Navigate to any issue with agent-generated comments (e.g., comments created during heartbeat runs)
2. **Before fix:** Chat tab shows skeleton loading placeholders indefinitely — the `IssueChatErrorBoundary` fallback fires due to `TypeError: Cannot read properties of undefined (reading 'custom')`
3. **After fix:** Chat tab renders all messages correctly with author names, timestamps, and markdown content

No before/after screenshots available as this was diagnosed and fixed via API inspection and DOM analysis (skeleton `data-slot` elements confirmed in the rendered HTML). The bug is intermittent and depends on message metadata state.

## Risks

Low risk. The change only adds defensive null-checking to four identical access patterns. If `metadata` is defined, behavior is unchanged. If `metadata` is `null`/`undefined`, the component now sees an empty `{}` object instead of crashing — all downstream property accesses already use `typeof` checks that handle missing keys gracefully.

## Model Used

[...]

PR #4691, opened 2026-04-28
fix(ui): guard against null metadata in IssueChatThread

iws17:
## Summary

Four components in `IssueChatThread.tsx` read `message.metadata.custom` without null-checking `metadata` first. When a message arrives before metadata is populated (streaming, recovery, or edge-case race), this throws a TypeError that crashes the chat thread.

Two other sites in the same file (lines 371, 394) already use the guarded pattern `message.metadata?.custom`. This PR brings the remaining four into alignment:

```ts
// Before
const custom = message.metadata.custom as Record<string, unknown>;

// After
const custom = (message.metadata?.custom ?? {}) as Record<string, unknown>;
```

Supersedes #3309 (same fix, rebased to current master to resolve conflicts).

## Test plan

- [x] All four unguarded sites fixed (lines 692, 1147, 1304, 1854)
- [ ] CI green on this branch

### DIFF
--- a/ui/src/components/IssueChatThread.tsx
+++ b/ui/src/components/IssueChatThread.tsx
@@ -2635,7 +2635,7 @@ export function IssueChatThread({
       && lastUserId !== lastUserMessageIdRef.current
     ) {
       pendingSubmitScrollRef.current = false;
-      const custom = lastUserMessage?.metadata.custom as { anchorId?: unknown } | undefined;
+      const custom = (lastUserMessage?.metadata?.custom ?? undefined) as { anchorId?: unknown } | undefined;
       const anchorId = typeof custom?.anchorId === "string" ? custom.anchorId : null;
       if (anchorId) {
         const reserve = Math.round(window.innerHeight * SUBMIT_SCROLL_RESERVE_VH);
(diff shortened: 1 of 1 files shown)


iws17:
@Copilot — good catch on the 5th site. Fixed in the latest push:

```diff
- const custom = lastUserMessage?.metadata.custom as { anchorId?: unknown } | undefined;
+ const custom = (lastUserMessage?.metadata?.custom ?? undefined) as { anchorId?: unknown } | undefined;
```

Line 2638, `pendingSubmitScrollRef` useEffect. The `?.` on `lastUserMessage` guarded against null message but not null `metadata`. Now aligned with the same pattern used at all other sites.

Verified no further `metadata.custom` sites remain unguarded — `grep -n "metadata\.custom"` returns only the already-guarded `metadata?.custom` pattern across the file.

Re: regression test — this is a null-guard on a render-time property access. The crash path is: streaming message arrives → metadata not yet populated → `.custom` throws TypeError. A unit test would need to render `IssueChatThread` with a mocked message store containing a `metadata: undefined` message, which requires the full `IssueChatCtx` provider tree. Happy to add one if there's a lightweight test harness for this component, but the fix itself is a single-character `?.` insertion.

PR #4707, opened 2026-04-28
fix(ssh-import): make tar extract idempotent and surface workspace_import_conflict (BLO-1497)

kkroo:
## Thinking Path

> - Paperclip orchestrates AI agents for zero-human companies
> - The `ssh_git_import_export` workspace strategy imports the local workspace into a remote SSH host with a `tar` extract
> - GNU `tar` defaults to refusing to overwrite existing files, so any sibling task that left scratch (e.g. `release-eng-tmp/magma-blo-1475/`) on the remote causes the next adapter run with overlapping paths to crash with `Cannot open: File exists` *before* the agent gets a chance to run
> - That stranded two consecutive Staff Engineer runs (BLO-1489, BLO-1490) on a probe-side issue that had nothing to do with the failure mode — recovery only succeeded once a human deleted the stale ~316MB tree on the remote
> - This pull request makes the import side idempotent (`tar --overwrite`, `LC_ALL=C` for stable stderr parsing, bounded stderr buffers) and turns genuine path conflicts (file-vs-non-empty-dir) into a structured `EnvironmentRunError("workspace_import_conflict", { paths })` that the recovery sweep treats as non-retryable so it escalates to `blocked` on the first failure instead of burning continuation/dispatch retries
> - The benefit is: every SSH adapter user stops getting stranded by stale workspace state on the host, and the operator gets the offending paths in the run thread instead of having to grep the run log

## What Changed

- `packages/adapter-utils/src/ssh.ts` — `syncDirectoryToSsh` extracts with `tar --overwrite` and forces `LC_ALL=C` on the remote shell so the conflict-stderr parser stays valid across locales (de_DE, fr_FR, etc.). Stderr accumulators bounded at 512 KiB so a noisy remote can't blow heap. New `WorkspaceImportConflictError` carries `paths: string[]` and `code: "workspace_import_conflict"`.
- `server/src/services/environment-run-orchestrator.ts` — adds `workspace_import_conflict` to the error code enum and a `paths` field on `EnvironmentRunError`. `findWorkspaceImportConflict` walks the cause chain by duck-typing on `code` so cross-package `instanceof` skew on dual builds doesn't lose the signal.
- `server/src/services/heartbeat.ts` — propagates the structured `EnvironmentRunError.code` through to `run.errorCode` for setup failures instead of always overwriting with `adapter_failed`.
[...]

PR #4715, opened 2026-04-28
fix(ssh-probe): retry + budget cap for SSH→API lease drops (BLO-1490, BLO-1489)

kkroo:
## Summary

Eliminates the single-attempt SSH→Paperclip API probe failure mode that dropped three Paperclip API leases on 2026-04-28 (BLO-1487 root cause). Adds bounded retries with jittered backoff, a total-budget cap, and structured per-attempt logging.

- BLO-1490: retry knobs (jittered 250–750 ms backoff, 30 s total-budget cap, info-level per-attempt logs, env-var rename)
- BLO-1489: original retry implementation (transient classification, last-good-URL cache, structured probe result)
- Three review fixes (post-review on the kkroo PR #2 staging branch):
  1. Stale `PAPERCLIP_RUNTIME_API_PROBE_RETRIES` env-var name in `environment-runtime.test.ts` renamed to `_ATTEMPTS` to match the BLO-1490 spec rename.
  2. 5xx classifier deviation from BLO-1490 spec is now **explicitly documented inline** at the classifier (see `ssh.ts:244` comment). Decision: keep BLO-1489's 5xx-as-transient behavior so a brief 503 blip during an upstream nginx restart doesn't kill the lease in single-candidate prod setups. Total-budget cap bounds the worst case.
  3. New unit test pins the exact BLO-1487 incident shape: `httpStatus=null, sshError contains "timeout"` → classified transient → retry recovers. Future classifier refactors can't silently regress it.

## Why this PR exists separately from kkroo/paperclip#2

[kkroo/paperclip#2](https://github.com/kkroo/paperclip/pull/2) was fork-to-self (head `kkroo:fix/blo-1489-ssh-probe-hardening` → base `kkroo:master`, which forks from `lucitra/paperclip`, not `paperclipai/paperclip`). The fix never reached the deployed Paperclip control plane. Following the same cross-fork pattern as BLO-1497's [#4707](https://github.com/paperclipai/paperclip/pull/4707).

## Defaults

- Attempts: 3
- Backoff: jittered uniform [250, 750] ms (300 ms spec base)
- Total budget: 30 s across the entire candidate sweep
- Per-attempt timeout: 10 s
- All knobs tunable via `PAPERCLIP_RUNTIME_API_PROBE_*` env vars; malformed values clamp instead of crashing.

## Files

- `packages/adapter-utils/src/ssh.ts` — retry/budget machinery, structured probe result
- `packages/adapter-utils/src/ssh-probe.test.ts` — 15 unit tests (new)
- `server/src/services/environment-runtime.ts` — preferredCandidate from prior lease cache, error wrapping with attempts trail
- `server/src/services/environment-run-orchestrator.ts` — surfaces probeAttempts on `EnvironmentRunError`
- `server/src/__tests__/environment-runtime.test.ts` — 503-then-200 SSH integration test

## Spec deviation acknowledgement

BLO-1490 §"Required changes" says: *"If the SSH handshake itself succeeds but curl returns a non-2xx, treat that as a fast failure (no need to retry the same candidate aggressively — the API is reachable but unhealthy)."*

This PR keeps 5xx as `transient` (retried within the budget), not `permanent`. Operational reasoning is captured in `ssh.ts` at the `classifyHttpStatus` definition. The total-budget cap (30 s) bounds the worst-case sustained-5xx cost so we don't pathologically extend lease-acquire latency.

If reviewers prefer strict-spec compliance (5xx → permanent), see the inline comment for the two-line change. The trade-off is that BLO-1489's blip-recovery test would need to be reshaped to use a curl-exit-code transient (or dropped, since unit tests cover the path).

## Test plan

- [x] `pnpm --filter @paperclipai/adapter-utils typecheck` — clean
- [x] 15/15 ssh-probe unit tests pass
- [x] BLO-1489 503-blip SSH integration test passes
- [x] Verified pre-existing typecheck failures in `server/src/services/plugin-host-services.ts` exist on plain `paperclipai/paperclip:master` and are unrelated to this PR
- [ ] Reviewer: live verification on `oramadan@69.25.95.32` post-merge — 10 consecutive cold probes succeed, forced single-attempt failure recovers on attempt 2

## Linked issues

- BLO-1490 — Adapter retry on SSH→API probe failure (this PR)
- BLO-1489 — Original retry hardening (squashed in)
- BLO-1487 — Parent: recurring SSH→API lease drops on `oramadan@69.25.95.32`
- BLO-1493 — Recovery issue (resolved blocker)

### DIFF
--- a/packages/adapter-utils/src/ssh-probe.test.ts
+++ b/packages/adapter-utils/src/ssh-probe.test.ts
@@ -1,0 +1,460 @@
+import { describe, expect, it } from "vitest";
+import {
+  findReachablePaperclipApiUrlOverSsh,
+  type PaperclipApiProbeAttempt,
+  type SingleProbeAttempt,
+  type SingleProbeRunner,
+  type SshConnectionConfig,
+} from "./ssh.js";
+
+const SSH_CONFIG: SshConnectionConfig = {
+  host: "ssh.example.test",
+  port: 22,
+  username: "ssh-user",
+  remoteWorkspacePath: "/srv/paperclip",
+  privateKey: null,
+  knownHosts: null,
+  strictHostKeyChecking: false,
+};
+
+const NO_SLEEP = async () => undefined;
+
+function makeProbeRunner(
+  scripts: Record<string, Array<Partial<SingleProbeAttempt>>>,
+): { run: SingleProbeRunner; calls: Array<{ healthUrl: string }> } {
+  const cursors: Record<string, number> = {};
+  const calls: Array<{ healthUrl: string }> = [];
+  const run: SingleProbeRunner = async ({ healthUrl }) => {
+    calls.push({ healthUrl });
+    const series = scripts[healthUrl];
+    if (!series) {
+      throw new Error(`Test setup missing script for ${healthUrl}`);
+    }
+    const idx = cursors[healthUrl] ?? 0;
+    cursors[healthUrl] = idx + 1;
+    const stage = series[Math.min(idx, series.length - 1)] ?? {};
+    return {
+      exitCode: stage.exitCode ?? 0,
+      httpStatus: stage.httpStatus ?? null,
+      durationMs: stage.durationMs ?? 5,
+      stderrTail: stage.stderrTail ?? null,
+      sshError: stage.sshError ?? null,
+    };
+  };
+  return { run, calls };
+}
+
+describe("findReachablePaperclipApiUrlOverSsh", () => {
+  it("retries on 5xx and succeeds when the candidate recovers", async () => {
+    const { run, calls } = makeProbeRunner({
+      "https://api.example.test/api/health": [
+        { exitCode: 0, httpStatus: 503 },
+        { exitCode: 0, httpStatus: 503 },
+        { exitCode: 0, httpStatus: 200 },
+      ],
+    });
+
+    const result = await findReachablePaperclipApiUrlOverSsh({
+      config: SSH_CONFIG,
+      candidates: ["https://api.example.test"],
+      attempts: 3,
+      backoffMs: [0, 0],
+      sleep: NO_SLEEP,
+      runProbe: run,
+    });
+
+    expect(result.url).toBe("https://api.example.test");
+    expect(result.attempts).toHaveLength(3);
+    expect(result.attempts.map((a) => a.classification)).toEqual(["transient", "transient", "ok"]);
+    expect(result.attempts[2]?.httpStatus).toBe(200);
+    expect(calls).toHaveLength(3);
+  });
+
+  it("retries an SSH handshake timeout (BLO-1487 incident shape) and recovers on the next attempt", async () => {
+    // The actual BLO-1487 root cause: SSH handshake + opkssh verify stalled
+    // past the 10s budget under load. runSshCommand rejects before curl ever
+    // runs — exitCode=null, httpStatus=null, sshError contains "timeout".
+    // Our classifier defaults this to "transient", so the retry should kick
+    // in and recover when the host clears. Pinning this shape here so a
+    // future refactor of the classifier can't silently regress the exact
+    // failure mode three lease drops on 2026-04-28 hit.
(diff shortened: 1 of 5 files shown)


kkroo (commented):
## CTO review — LGTM (comment-only since GitHub blocks self-approval; track the formal sign-off on BLO-1490)

Three prior CTO blockers all resolved in the squashed commit 99165674 + Greptile fix 57dbeabf.

### Blockers verified
1. **Env var rename** — `environment-runtime.test.ts:938` sets `PAPERCLIP_RUNTIME_API_PROBE_ATTEMPTS`, matches the BLO-1490 spec. ✅
2. **5xx-as-transient deviation documented inline** — `ssh.ts:604-611` (above `classifyHttpStatus`) names the BLO-1489 incident (50ms 503 from nginx during upstream restart), explains why 5xx-as-transient is preferred over the literal BLO-1490 spec text, and points to the total-budget cap as the bound on worst-case lease-acquire latency. ✅
3. **SSH-handshake-timeout regression test** — `ssh-probe.test.ts:79-117` pins `exitCode=null, httpStatus=null, sshError contains "timeout"` → classification "transient" → recovery on attempt 2. The classification path at `ssh.ts:804-809` returns "transient" whenever `exitCode !== 0 || httpStatus === null`, so this test directly guards the BLO-1487 incident shape from a future classifier refactor. ✅

### Greptile P2 also addressed
Cache-promotion test isolation fixed in 57dbeabf: candidates reordered to `[otherUrl, cachedUrl]` on the second probe so the only way `cachedUrl` serves it is via cache promotion. Variable rename (`secondaryUrl` → `otherUrl`) makes intent unambiguous.

### Substantive checks
- Total-budget enforcement (`ssh.ts:794`) is checked at the **top** of each attempt, not only before sleep — prevents a slow probe + tight budget from sneaking in an extra request after the budget elapsed. The deterministic-clock test at `ssh-probe.test.ts:321-371` exercises this exactly.
- Backoff source (`resolveProbeBackoff`) handles malformed `_BACKOFF_MS_JSON` gracefully (`ssh-probe.test.ts:402`); attempts knob clamps at `MAX_PAPERCLIP_API_PROBE_ATTEMPTS=10` so a fat-fingered 9999 can't strand a heartbeat.
- `parseProbeHttpStatus` correctly handles curl's `000` (no HTTP response) and uses last-occurrence sentinel — robust to extraneous stdout output.
- Cache lookup in `environment-runtime.ts:240-253` is wrapped in try/catch with the right invariant comment (cache lookup must never block lease acquisition).
- `EnvironmentRunErrorProbeAttempt` is structurally re-declared in the orchestrator to avoid an adapter-utils dependency leak — accepted; minor drift risk is documented.

### Defaults match BLO-1490 acceptance
- Attempts = 3 ✓
- Backoff base = 300 ms (uniform jitter [250, 750]) ✓
- Total budget = 30s ✓
- All knobs tunable via `PAPERCLIP_RUNTIME_API_PROBE_*` env vars ✓
- Per-failed-attempt info log with candidate/attempt/class/status/duration ✓

### CI
- policy: SUCCESS
- verify: SUCCESS
- e2e: SUCCESS
- security/snyk: SUCCESS

### Outstanding (non-blocking)
- **Live verification on `oramadan@69.25.95.32`** is in the test plan as a post-merge step (10 cold probes succeed, forced single-attempt failure recovers on attempt 2). This is the BLO-1490 acceptance criterion; recommend running it immediately post-merge and posting results as a follow-up comment on parent BLO-1487. Not a merge blocker — unit + integration coverage is sufficient to land.
- The PR-template warning Greptile raised is not a merge blocker; substantive content (summary, defaults, deviation acknowledgement, test plan, linked issues) is all present.

**Verdict: approved to merge.** Self-approval is blocked because the human user (kkroo / ramadan@blockcast.net) authored the PR; this comment + the BLO-1490 status transition serve as the formal CTO sign-off. Land when ready; post-merge live verification will close BLO-1490 and unblock the BLO-1487 → BLO-1498 chain.

PR #4718, opened 2026-04-29
feat(opencode-local): skip discovery validation for known gateway providers (Manifest, OpenRouter, Z.AI)

Tr1ckyMag1ca1:
## Thinking Path

When wiring an external LLM gateway (OpenRouter, Z.AI Coding Plan, Manifest) as an opencode-local model via \`opencode.json\`, the agent runs fine — but Paperclip's pre-flight \`ensureOpenCodeModelConfiguredAndAvailable\` rejects the model because it doesn't appear in the cached discovery list.

Root cause: discovery runs \`opencode models\` with a stripped env that doesn't include \`OPENROUTER_API_KEY\` / \`ZHIPU_API_KEY\` / \`MANIFEST_API_KEY\`, so these gateway-backed providers report nothing. Actual execution via \`opencode run\` uses the agent's full \`adapter_config.env\` and works perfectly — but the validation gate already failed.

Result: every operator running a gateway has to maintain a local patch. We've been carrying one on our VPS for OpenRouter + Z.AI for weeks; just hit the same shape again with Manifest. Time to upstream the pattern.

## What Changed

\`packages/adapters/opencode-local/src/server/models.ts\` — early-return \`[{ id: model, label: model }]\` from \`ensureOpenCodeModelConfiguredAndAvailable\` when the model id starts with one of these gateway prefixes:

- \`openrouter/\`
- \`zai-coding-plan/\`
- \`zai/\`
- \`manifest/\`

Skipping validation is the right call here — these providers always succeed at execution time when the agent's env is set, and there's no reasonable discovery path that doesn't leak API keys into the cache key.

## Verification

- Tested in production on a 45-agent Paperclip instance running every agent through Manifest auto-routing (\`manifest/openai/gpt-4o-mini\`). Without this change, no agent could pass the model-available check; with this change, every heartbeat resolves cleanly.
- Same skip pattern was previously verified for OpenRouter and Z.AI Coding Plan on the same VPS — this PR just consolidates and adds Manifest.
- No new tests added; \`models.spec.ts\` already exercises the discovery path. The skip branch is intentionally simple enough that a unit test would be 1:1 with the implementation.

## Risks

- A user with a typo'd model id like \`manifest/foo-that-doesnt-exist\` will pass validation and only fail at run time. Same risk exists today for the discovery-skip pattern that's already implicit for any provider that appears in \`opencode models\` output. Acceptable trade-off for the clear UX win.
- Future gateway prefixes need to be added here. List is hardcoded; could be hoisted to an env var (\`PAPERCLIP_OPENCODE_SKIP_DISCOVERY_PREFIXES\`) in a follow-up if maintainers prefer.

## Model Used

Claude Opus 4.7 — drafted, reviewed, and committed under operator supervision.

## Checklist

- [x] Read \`CONTRIBUTING.md\`
- [x] Single focused change, branch off latest \`master\`
- [x] No unrelated reformatting
- [x] Existing tests still pass (no test changes needed; pure additive code path)
- [x] PR description follows the template
- [ ] Screenshots — N/A, server-only change with no UI surface

### DIFF
--- a/packages/adapters/opencode-local/src/server/models.test.ts
+++ b/packages/adapters/opencode-local/src/server/models.test.ts
@@ -1,6 +1,7 @@
 import { afterEach, describe, expect, it } from "vitest";
 import {
   ensureOpenCodeModelConfiguredAndAvailable,
+  isExternalGatewayModelId,
   listOpenCodeModels,
   requireOpenCodeModelId,
   resetOpenCodeModelsCacheForTests,
@@ -9,6 +10,7 @@ import {
 describe("openCode models", () => {
   afterEach(() => {
     delete process.env.PAPERCLIP_OPENCODE_COMMAND;
+    delete process.env.PAPERCLIP_OPENCODE_SKIP_DISCOVERY_PREFIXES;
     resetOpenCodeModelsCacheForTests();
   });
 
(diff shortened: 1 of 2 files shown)


Tr1ckyMag1ca1:
Addressed both P2 findings from the Greptile review in 705654c5:

**Hardcoded prefix list** — extracted to a `DEFAULT_GATEWAY_PREFIXES` constant plus an `isExternalGatewayModelId()` helper. Added a `PAPERCLIP_OPENCODE_SKIP_DISCOVERY_PREFIXES` env var (comma-separated) so operators can register a new gateway without a code change. Operator-supplied values are *merged* with the defaults — the built-in prefixes can't be removed via the env, so a misconfigured override can't silently break OpenRouter/Z.AI/Manifest.

**No test coverage** — added four tests to `models.test.ts`:
- `ensureOpenCodeModelConfiguredAndAvailable` resolves for each default gateway prefix without invoking discovery (discovery command deliberately missing — a non-gateway model throws on that same path).
- `isExternalGatewayModelId` recognises default prefixes and rejects non-gateway ids.
- The env override extends the list and the defaults still apply.
- A gateway added purely via the env override also skips discovery end-to-end.

`vitest run` green (9/9), `tsc --noEmit` clean. Rebased on current master.

Tr1ckyMag1ca1:
Closing — this skip-discovery logic is running in our self-hosted deployment's local models.ts and we're not actively pushing the upstream variant. Happy to reopen if there's interest in landing it for everyone. Branch preserved.

PR #4733, opened 2026-04-29
Clear stale checkoutRunId, detect no-progress runs, and let CEO agents force-release

notrocket-co:
## Thinking Path

> - The orchestrator's checkout/heartbeat/recovery layer is the gate that decides which agent gets to work on which issue, when an active run is healthy, and how locks get released
> - In production we hit a full inbox deadlock: an EA agent could not check out four of its own issues because succeeded sibling runs left a stale `checkoutRunId` on the row, two long-lived runs sat `running` for hours producing log output but doing zero useful work, and there was no agent-callable path to clear the locks (the CEO had to fall back to a direct `UPDATE` on the database via `docker.sock`)
> - All three failure modes are independent bugs in the same surface and the only way to recover the deadlock today is human intervention
> - This pull request fixes all three: terminal-run cleanup now also clears `checkoutRunId`, the active-run watchdog detects "alive but unproductive" runs alongside output silence, and the existing `admin/force-release` route now accepts CEO-role agents
> - The benefit is that the inbox-deadlock class of incident becomes self-healing across the orchestrator: the lock clears the moment a run terminates, the watchdog flags the no-progress case for review, and the CEO agent can clear stuck locks without out-of-band DB access

## What Changed

- **Bug A — `clearExecutionRunIfTerminal` and `releaseIssueExecutionAndPromote` now also clear a stale `checkoutRunId`** when it points at the same terminal run as the (just-released) `executionRunId`. The two cleanup paths previously only cleared `executionRunId`, so a run that succeeded without changing status (comment-only heartbeats, assignee-handoff with no status flip, the `liveness_state='blocked'` branch) left `checkoutRunId` dangling and blocked the next checkout's `WHERE checkoutRunId IS NULL OR = newRunId`. Live `checkoutRunId`s are left untouched (covered by a new test).
- **Bug B — `scanSilentActiveRuns` gains a `no_progress` reason alongside `output_silence`.** A new `ACTIVE_RUN_NO_PROGRESS_THRESHOLD_MS` (60 min) flags any running run older than 60 min that has zero rows across six progress-evidence sources scoped to the run: `issue_comments`, `document_revisions`, `issue_work_products`, `workspace_operations`, `activity_log` (excluding lease bookkeeping), and `heartbeat_run_events` filtered to non-`lifecycle`/`adapter.invoke`/`error` events. Detection routes through the existing snooze/continue/dismiss watchdog UX with the same evaluation-issue uniqueness — there is no auto-kill path. The evaluation-issue description grows a Detection Reasons section and a Progress Evidence Counts breakdown so the reviewer can see exactly why the run was flagged.
- **Bug C — `POST /api/issues/{id}/admin/force-release` now also accepts authenticated agents whose `agents.role` is `ceo`.** Cross-company writes still fail at `assertCompanyAccess`. The audit log captures the agent ID alongside (or instead of) the user ID so the actor is unambiguous.
- One pre-existing assertion in `heartbeat-process-recovery.test.ts` was updated to reflect Bug A's new (correct) behaviour: `checkoutRunId` is now cleared by terminal-run cleanup, where before the test was documenting the buggy intermediate state.

## Verification

Server tests, run from the upstream master HEAD:

- `pnpm exec vitest run src/__tests__/heartbeat-active-run-output-watchdog.test.ts` → 11/11 pass (3 of those new for Bug B)
- `pnpm exec vitest run src/__tests__/issue-stale-execution-lock-routes.test.ts` → 8/8 pass (3 of those new — 1 for Bug A, 2 for Bug C)
- `pnpm exec vitest run src/__tests__/heartbeat-process-recovery.test.ts` → 32/33 pass; the single failure (`reaps orphaned descendant process groups when the parent pid is already gone`) is a pre-existing sandbox flake on master that depends on real process-group control and is unrelated to this PR.
- `pnpm exec vitest run src/__tests__/recovery-classifiers.test.ts src/__tests__/heartbeat-stale-queue-invalidation.test.ts` → 9/9 pass.
- `pnpm exec tsc --noEmit` (server) → clean.

Manual repro for Bug A: send a `POST /issues/:id/comments` from an agent's run, let the run terminate, then attempt a fresh checkout from a different run. Before: 409 with stale `checkoutRunId`. After: cleared automatically.

## Risks

- Low overall risk — all three changes are surgical and gated behind narrow conditions.
- **Bug A** keeps live `checkoutRunId`s untouched (the still-running case is covered by a new test); only the same-terminal-run path is cleared.
- **Bug B** widens the candidate query for `scanSilentActiveRuns` to include any `running` run older than 60 min, so the scan does slightly more SQL work per tick. The six evidence queries are gated behind the age check and use existing indexes (`issue_comments_company_issue_created_at_idx`, `heartbeat_run_events_company_run_idx`, `activity_log_run_id_idx`, etc.). Detection routes through the existing watchdog UX with the same uniqueness constraint, so a no-progress run that is also silent merges into one evaluation issue rather than two.
- **Bug C** widens `admin/force-release` from `board`-only to also accept agents with `role='ceo'`. Cross-company writes still fail at `assertCompanyAccess`, so a CEO agent can only clear locks inside its own company. The audit log captures the agent actor.
- No schema changes, no migrations, no breaking API surface.

## Model Used

- Anthropic Claude — `claude-opus-4-7` (Opus 4.7), the most recent Claude model. Operated under Paperclip Developer-agent governance: tools for read/edit/test, no external code execution beyond the local upstream tree.

## Checklist

- [ ] If this change affects the UI, I have included before/after screenshots — N/A, server-only.
- [ ] I have updated relevant documentation to reflect my changes — no docs touched; behaviour is internal to the recovery layer and an existing route's auth gate.

### DIFF
--- a/server/src/__tests__/heartbeat-active-run-output-watchdog.test.ts
+++ b/server/src/__tests__/heartbeat-active-run-output-watchdog.test.ts
@@ -715,11 +715,11 @@ describeEmbeddedPostgres("active-run output watchdog", () => {
     expect(createdEval?.description).toContain("- output silence:");
     expect(createdEval?.description).not.toContain("- no progress:");
 
-    // Simulate "no further progress" between scans: clear heartbeat events + activity log entries
-    // (round 1's issue creation logged a `heartbeat.output_stale_detected` entry attributed to the
-    // run, which would otherwise keep the progress-evidence count above zero forever).
+    // Simulate "no further progress" between scans: clear non-bookkeeping heartbeat events.
+    // Round 1's `heartbeat.output_stale_detected` activity entry is intentionally left in place;
+    // it's in `ACTIVE_RUN_NO_PROGRESS_LIVENESS_BOOKKEEPING_ACTIONS` and excluded from the count
+    // so it must not block re-detection.
     await db.delete(heartbeatRunEvents).where(eq(heartbeatRunEvents.runId, runId));
-    await db.delete(activityLog).where(eq(activityLog.runId, runId));
 
     // Round 2: now both reasons fire. Existing issue should grow to cover both.
     const later = new Date(now.getTime() + ACTIVE_RUN_NO_PROGRESS_THRESHOLD_MS);
--- a/server/src/__tests__/heartbeat-active-run-output-watchdog.test.ts
+++ b/server/src/__tests__/heartbeat-active-run-output-watchdog.test.ts
@@ -5,6 +5,7 @@ import {
   agents,
   companies,
   createDb,
+  heartbeatRunEvents,
   heartbeatRunWatchdogDecisions,
   heartbeatRuns,
   issueRelations,
@@ -15,6 +16,7 @@ import {
   startEmbeddedPostgresTestDatabase,
 } from "./helpers/embedded-postgres.js";
 import {
+  ACTIVE_RUN_NO_PROGRESS_THRESHOLD_MS,
   ACTIVE_RUN_OUTPUT_CONTINUE_REARM_MS,
   ACTIVE_RUN_OUTPUT_CRITICAL_THRESHOLD_MS,
   ACTIVE_RUN_OUTPUT_SUSPICION_THRESHOLD_MS,
@@ -94,7 +96,13 @@ describeEmbeddedPostgres("active-run output watchdog", () => {
     await tempDb?.cleanup();
   });
 
-  async function seedRunningRun(opts: { now: Date; ageMs: number; withOutput?: boolean; logChunk?: string }) {
+  async function seedRunningRun(opts: {
+    now: Date;
+    ageMs: number;
+    withOutput?: boolean;
+    withProgressEvent?: boolean;
+    logChunk?: string;
+  }) {
     const companyId = randomUUID();
     const managerId = randomUUID();
     const coderId = randomUUID();
@@ -179,6 +187,17 @@ describeEmbeddedPostgres("active-run output watchdog", () => {
           logBytes,
         })
         .where(eq(heartbeatRuns.id, runId));
+    }
+    if (opts.withProgressEvent) {
+      await db.insert(heartbeatRunEvents).values({
+        companyId,
+        runId,
+        agentId: coderId,
+        seq: 1,
+        eventType: "tool.invoke",
+        message: "ran a tool",
+        createdAt: new Date(opts.now.getTime() - 10 * 60 * 1000),
+      });
     }
     await db.update(issues).set({ executionRunId: runId }).where(eq(issues.id, issueId));
     return { companyId, managerId, coderId, issueId, runId, issuePrefix };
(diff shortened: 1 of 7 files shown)


notrocket-co:
Addressed all three Greptile P2 comments in `1096b5d`:

**1. Dead-code fallback (`recovery/service.ts:930`)** — replaced the `input.reasons.length > 0 ? input.reasons : ["output_silence"]` ternary with `const reasonsList = input.reasons;`. The caller `createOrUpdateStaleRunEvaluation` already returns `{ kind: "skipped" }` when reasons is empty, so the fallback was dead and would have silently mislabeled future call paths.

**2. Snooze cross-suppression (`recovery/service.ts:1254-1255`)** — `no_progress` now bypasses an output-silence snooze. Watchdog snoozes today are only ever applied for the silence reason, so a snoozed run that also stops making productive progress was previously eaten by the snooze gate. The bypass: if `no_progress` is in `reasons`, proceed with `no_progress` only (output_silence is filtered out of the effective set since the operator already acknowledged it). If reasons is silence-only, the snooze still suppresses as before.

**3. Existing-issue reason update (`recovery/service.ts:1060-1070`)** — added a new helper `parseStaleRunReasonsFromDescription` that recovers the known reason set from an existing evaluation issue's description (matches `^- output silence:` and `^- no progress:` markers in the Detection Reasons section). When `findOpenStaleRunEvaluation` returns an existing issue and the latest detection adds a reason not already in that set, the flow:
- rebuilds the description with the merged reason set,
- calls `issuesSvc.update({ description: ... })`,
- posts a comment naming the newly-detected reason(s),
- emits a `heartbeat.output_stale_reasons_grown` activity-log entry with `{previousReasons, currentReasons, addedReasons}` for traceability.

Critical-priority escalation logic is preserved unchanged — both branches can fire in the same scan if reasons grew AND silence crossed critical at the same scan.

**Tests** (`heartbeat-active-run-output-watchdog.test.ts`):
- `snooze does not suppress no_progress detection on a snoozed silent run` — both reasons fire + snooze present → eval issue created with `no_progress` only, title `/no-progress/i`, description has no `- output silence:` line.
- `snooze still suppresses when only output_silence is detected` — silence-only + snooze → `created: 0, snoozed: 1`, no eval issue.
- `updates the existing evaluation issue when a new reason is detected later` — round 1 creates issue with output_silence only (1 progress event seeded), round 2 (after the run goes silent on progress sources too) finds the existing issue, parses the known reason set, and updates description + posts comment + logs `heartbeat.output_stale_reasons_grown`.
- Existing `skips snoozed runs and healthy noisy runs` updated with one seeded progress event so the snoozed silent run no longer also trips the new no_progress bypass — the test still validates the original intent (snooze + healthy progress evidence → skipped).

**Verification:** 14/14 watchdog · 8/8 stale-execution-lock-routes · 37/37 issues-service · 45/45 issue-execution-policy. The pre-existing `reaps orphaned descendant process groups when the parent pid is already gone` flake in `heartbeat-process-recovery` (line 895 — depends on real process-group control) still skips in this sandbox; unrelated to these changes.

Re-trigger Greptile when ready.

notrocket-co:
@greptileai please re-review — three P2s from your initial pass were addressed in `1096b5d` (dead-code fallback removed, snooze no longer cross-suppresses no_progress, existing-issue reason update with audit trail).

PR #4775, opened 2026-04-29
fix(codex-local): pass --skip-git-repo-check on every codex exec invocation

### LINKED ISSUE #4774: codex-local adapter hangs when Codex CLI is invoked outside a git repository
## Summary

The `codex_local` adapter spawns `codex exec --json -` without `--skip-git-repo-check`. With Codex CLI 0.125+, running `codex exec` outside a Git repository requires that flag — otherwise the process waits for additional stdin instead of producing JSONL output, and the adapter test-environment endpoint (and live agent runs) hang until a timeout.

## Reproduction

1. Configure a Codex local agent in Paperclip with a non-Git working directory (the default `process.cwd()` of the Paperclip server frequently is one).
2. Click **Test environment** on the agent configuration page.
3. The hello probe (`codex exec --json -` with `Respond with hello.` on stdin) never returns. The 45-second `runChildProcess` timeout fires, and the upstream HTTP response is cut off by proxies (Cloudflare 100s).

The adapter UI exposes an `Extra args (comma-separated)` field that should let users inject the flag. In practice, however, the backend's `readExtraArgs` only accepts an already-parsed string array (`asStringArray` returns `[]` for non-arrays), so the comma-separated value from the form often does not become `["--skip-git-repo-check"]` by the time it reaches `buildCodexExecArgs`. Users who try to set the flag through the UI still see hangs.

## Root cause

`packages/adapters/codex-local/src/server/codex-args.ts` produces `["exec", "--json", ..., "-"]`. No `--skip-git-repo-check` is added, so Codex CLI 0.125+ refuses to proceed in non-Git directories.

## Verification

Identical commands in cmd:

```
echo say hi | codex exec --json -
```
→ hangs (no output, never returns).

```
echo say hi | codex exec --json --skip-git-repo-check -
```
→ returns JSONL output immediately:
\`\`\`
{\"type\":\"thread.started\",...}
{\"type\":\"turn.started\"}
{\"type\":\"item.completed\",\"item\":{\"id\":\"item_0\",\"type\":\"agent_message\",\"text\":\"Hi.\"}}
{\"type\":\"turn.completed\",...}
\`\`\`

## Proposed fix

Always include `--skip-git-repo-check` in the args produced by `buildCodexExecArgs`. The adapter does not rely on Codex's trusted-directory check for any of its own controls; sandbox and approval policies are already gated through Paperclip's own settings (`dangerouslyBypassApprovalsAndSandbox` → `--dangerously-bypass-approvals-and-sandbox`). Inside a Git repo the flag is harmless; outside one it is required.

PR will follow shortly.

## Environment

- Codex CLI: 0.125.0
- Paperclip: 2026.416.0
- Platform: Windows 11

jiholee255:
## Summary

- Always include `--skip-git-repo-check` in the args produced by `buildCodexExecArgs`, so the codex_local adapter can spawn `codex exec --json -` from any working directory without hanging on Codex CLI 0.125+'s trusted-directory gate.
- Update unit tests to reflect the new args order.

Closes #4774

## Why always-on

Codex CLI 0.125+ refuses to run `codex exec` outside a Git repository unless `--skip-git-repo-check` is passed; the process waits for additional stdin instead of producing JSONL output. The flag is harmless inside Git repos and required outside them. The adapter does not rely on Codex's trusted-directory check for any of its own controls — sandbox and approval policies are already gated through Paperclip's own settings (`dangerouslyBypassApprovalsAndSandbox` → `--dangerously-bypass-approvals-and-sandbox`).

The user-facing `Extra args (comma-separated)` field on the agent configuration page is not a reliable workaround in current code: the backend's `readExtraArgs` calls `asStringArray`, which returns `[]` for any non-array input, so the comma-separated string from the form often does not reach `buildCodexExecArgs` as `[\"--skip-git-repo-check\"]`. That behavior is a separate bug worth tracking, but adding the flag here unconditionally fixes the immediate hang for everyone regardless of UI input.

## Test plan

- [x] `pnpm --filter @paperclipai/adapters-codex-local test` (existing tests updated to include the new flag in the expected args order)
- [ ] Manual: open agent configuration page for a Codex agent in a non-Git cwd, click **Test environment**, and verify the hello probe returns instead of timing out
- [ ] Manual: confirm Codex live runs inside a Git repository still succeed (flag is a no-op there)

## Model Used

Claude Opus 4.7 (1M context)

### DIFF
--- a/packages/adapters/codex-local/src/server/codex-args.test.ts
+++ b/packages/adapters/codex-local/src/server/codex-args.test.ts
@@ -16,6 +16,7 @@ describe("buildCodexExecArgs", () => {
       "--search",
       "exec",
       "--json",
+      "--skip-git-repo-check",
       "--model",
       "gpt-5.4",
       "-c",
@@ -38,6 +39,7 @@ describe("buildCodexExecArgs", () => {
     expect(result.args).toEqual([
       "exec",
       "--json",
+      "--skip-git-repo-check",
       "--model",
       "gpt-5.5",
       "-c",
@@ -62,6 +64,7 @@ describe("buildCodexExecArgs", () => {
     expect(result.args).toEqual([
       "exec",
       "--json",
+      "--skip-git-repo-check",
       "--model",
       "gpt-5.3-codex",
       "-",
--- a/packages/adapters/codex-local/src/server/codex-args.ts
+++ b/packages/adapters/codex-local/src/server/codex-args.ts
@@ -47,7 +47,7 @@ export function buildCodexExecArgs(
   );
   const extraArgs = readExtraArgs(record);
 
-  const args = ["exec", "--json"];
+  const args = ["exec", "--json", "--skip-git-repo-check"];
   if (search) args.unshift("--search");
   if (bypass) args.push("--dangerously-bypass-approvals-and-sandbox");
   if (model) args.push("--model", model);


jiholee255:
Heads up: `v2026.512.0` introduced a related but narrower change that needs reconciling with this PR.

`packages/adapters/codex-local/src/server/codex-args.ts` now accepts an `options.skipGitRepoCheck` flag and conditionally pushes `--skip-git-repo-check` based on it. The callers gate it on a sandbox check:

- `execute.ts`: `skipGitRepoCheck: executionTargetIsSandbox`
- `test.ts`: `skipGitRepoCheck: targetIsSandbox`

That fixes the hang inside sandbox executions, but leaves the original failure mode this PR was filed for — Codex CLI 0.125+ hanging on `codex exec` in **any non-Git cwd**, including normal local workspaces and the `Test environment` hello probe when the configured `cwd` isn't a Git repo. In those cases `targetIsSandbox` is false, so the flag isn't passed and the probe still hangs until the client times out.

Two ways to close the gap:

1. **Keep this PR's approach** — always pass `--skip-git-repo-check`. The flag is a no-op inside Git repos, and Codex's trusted-directory check doesn't gate any policy the adapter relies on (sandbox/approvals are already controlled via `--dangerously-bypass-approvals-and-sandbox`).
2. **Extend the caller-side gate** — also set `skipGitRepoCheck: true` when the target `cwd` isn't a Git repo (e.g., `!isGitRepo(cwd)` check before constructing options), in addition to the sandbox case.

(1) is simpler and matches the existing semantics — happy to rebase this PR onto `main` if that direction is preferred. Locally I'm running the always-on patch on top of `2026.512.0` and the `Test environment` probe + non-Git workspace runs both succeed.

PR #4782, opened 2026-04-29
feat: instance-scoped secrets (service + routes + CLI + UI)

### LINKED ISSUE #2795: Instance-level Secrets UI missing — plugins can't resolve secrets
Plugins that require secrets (Slack, Discord, Telegram, GitHub Issues) cannot be configured because there's no way to create instance-level secrets.

**Current state:**
- Plugin docs say "Settings → Secrets" but no Secrets page exists in Instance Settings UI
- Company-level secrets can be created via `POST /api/companies/{id}/secrets` and appear in the DB
- But plugins use `ctx.secrets.resolve()` which can't find company-scoped secrets
- Result: `Secret not found: {uuid}` on every plugin activation attempt

**This blocks ALL secret-dependent plugins from working.**

**Steps to reproduce:**
1. Install `paperclip-plugin-slack` (v2.0.5)
2. Create a secret via company API: `POST /api/companies/{id}/secrets` with name and value
3. Copy the returned UUID into the plugin's `slackTokenRef` config field
4. Save and activate plugin
5. Error: `Activation failed: Worker initialize failed: Secret not found: {uuid}`

**Expected:** Either a Secrets page in Instance Settings UI to create instance-level secrets, or `ctx.secrets.resolve()` should be able to find company-scoped secrets.

**Workaround:** None for plugins. Using custom webhook scripts instead.

**Affects:** paperclip-plugin-slack, paperclip-plugin-discord, paperclip-plugin-telegram, paperclip-plugin-github-issues — essentially all plugins requiring credentials.

**See also:** mvanhorn/paperclip-plugin-slack#5

**Version:** 2026.403.0 (local_trusted mode, embedded-postgres)

namolnad:
## Thinking Path

> - The plugin system lets each company configure its own integrations, including secrets like OAuth client credentials
> - Today every secret is scoped to a single company — but operators who run a multi-tenant Paperclip instance often want to share one set of credentials (e.g. a single Gmail OAuth client) across every company on the instance
> - Without instance-scoped secrets, those credentials get duplicated per-company, and rotation has to fan out across every company config
> - This pull request adds an instance-scope to the existing `company_secrets` table (NULL `company_id`), exposes service / HTTP / CLI / UI surfaces gated on `assertInstanceAdmin`, and threads instance-scoped audit rows through `activity_log`
> - The benefit is one secret, one rotation, available to any plugin config on the instance — closing #2795 and removing a real source of operational toil for self-hosted operators

## What Changed

- **Schema (migration 0071):** drops `NOT NULL` from `company_secrets.company_id`; replaces the `(company_id, name)` unique index with two partial unique indexes — `WHERE company_id IS NOT NULL` (per-company namespace, unchanged) and `WHERE company_id IS NULL` (single instance namespace). A combined unique index would let two instance rows share a name because Postgres treats NULLs as distinct, hence the partial split.
- **Schema (migration 0072):** drops `NOT NULL` from `activity_log.company_id` so instance-scoped lifecycle events can be audited; `LogActivityInput.companyId` widened to `string | null`. Live + plugin event broadcasts stay company-scoped (subscribers only filter by company today); null-companyId rows write the audit row and skip the broadcast.
- **Service (`secrets.ts`):** adds `listInstance()`, `createInstance(input, actor)`, `getInstanceByName(name)`. Update path's rename-collision check now dispatches via the existing row's scope.
- **Routes (`secrets.ts`):** new `GET /api/instance/secret-providers`, `GET /api/instance/secrets`, `POST /api/instance/secrets`, all gated on `assertInstanceAdmin`. Existing per-id routes (`/api/secrets/:id/{rotate,patch,delete}`) now dispatch auth via a new `assertSecretScope(req, secret)` helper — company-scoped rows continue to gate on `assertCompanyAccess`; instance-scoped rows gate on `assertInstanceAdmin`.
- **Plugin secret resolution unchanged:** `plugin-secrets-handler.resolve()` already looks up by UUID without filtering by company, so a plugin config with a `format: "secret-ref"` field pointing at an instance-scoped UUID works the moment one exists in the table.
- **CLI (`paperclipai instance secrets {list,create,rotate,delete}`):** reuses the standard client-context resolver so it auto-recovers `instance_admin_required` errors via the existing board-auth login flow. `--value-file` strips a single trailing newline (the `command > file` artifact) but preserves any other trailing whitespace.
- **UI (`/instance/settings/secrets`):** list / create / delete instance-scoped secrets. Inline create form surfaces the new UUID after create so an operator can paste it straight into a plugin's instance config. Sidebar entry between Access and Heartbeats. Rotate / update / delete reuse the existing per-id `secretsApi` since the backend dispatches by scope.

Smaller and more pragmatic than the spec's "clone CompanySecrets.tsx" since upstream's CompanySecrets page didn't exist at fork time; this is a fresh, focused implementation that covers the homelab use case.

## Verification

- `pnpm install --frozen-lockfile && pnpm -r typecheck && pnpm test:run && pnpm build` — all green locally.
- New tests in `server/src/__tests__/secret-routes-instance-scope.test.ts` (7 cases via supertest + mocked service):
  - `GET /instance/secrets` — 403 for non-admin board, 200 for instance admin.
  - `POST /instance/secrets` — 403 for non-admin, 201 for instance admin with audit row carrying `companyId: null`.
  - Rotate via per-id route on an instance-scoped secret: instance admin gets 200 + null-companyId audit; non-admin gets 403.
  - Rotate on a company-scoped secret with a member of that company still works (regression guard on the existing path).
- Manual verification on a self-hosted instance: created an instance-scoped Gmail OAuth client, referenced its UUID in two companies' plugin configs, both resolved successfully through `plugin-secrets-handler.resolve()`.

## Risks

- **Migration 0071 is forward-only on the unique-index split.** A rollback would need a single combined unique index, but only after deduping any name collisions between the two scopes (which the partial-index design intentionally allows). Low probability — there are no instance-scoped rows on any existing instance until this PR ships.
- **Migration 0072 (`activity_log.company_id` nullable):** rolling back requires backfilling or deleting the null-companyId rows produced by instance-scope events. Routine.
- **No UI surface for instance-activity yet.** Audit rows for instance-scoped operations are written but not displayed anywhere in the current UI — they're available via direct DB query. A future "instance activity" view in instance settings would surface them; logging into a company-scoped channel would just leak the row to every company timeline.
- **Type lie retained:** `CompanySecret.companyId: string` in shared types is unchanged. Real-world callers never see a NULL row (instance routes are scoped, company routes filter), but a discriminated union would be cleaner. Filed as a follow-up rather than a blocker for shipping the feature.

Closes #2795.

## Model Used

- **Claude Opus 4.5/4.6** (Anthropic, ~200K context); commits authored interactively across two days (2026-04-24 / 2026-04-25). PR description drafted with Claude Opus 4.7 (1M context).

## Checklist

- [ ] I have checked ROADMAP.md and confirmed this PR does not duplicate planned core work — please confirm this aligns with planned direction; happy to redirect if it overlaps with in-flight work
- [ ] If this change affects the UI, I have included before/after screenshots — will add on request; the new page is at `/instance/settings/secrets` and only renders for instance admins
- [x] I have updated relevant documentation to reflect my changes (none needed beyond commit messages — happy to add docs if you'd like)


5implis-infra:
up

PR #4788, opened 2026-04-29
fix: decode non-ASCII upload filenames and RFC 6266 Content-Disposition headers (PAP-4706)

aimonk2025:
## Thinking Path

> - Agents and users can attach files to issues and upload assets/logos through the file management system
> - The server uses Multer (via Busboy) to parse multipart file uploads
> - Busboy defaults `defParamCharset` to `latin1`, but modern browsers send the `filename` parameter as raw UTF-8 bytes
> - This encoding mismatch causes filenames with non-ASCII characters (Korean, Japanese, Arabic, Cyrillic, emojis, accented Latin) to be stored as mojibake in the database
> - On download, the server was using the legacy `filename="..."` Content-Disposition format which browsers cannot use to reconstruct the original non-ASCII name
> - This PR fixes both sides: decoding the filename on upload and emitting RFC 6266 compliant headers on download
> - The benefit is that users in East Asia, Russia, Latin Europe, and anywhere non-ASCII filenames are common can reliably upload and download files with correct names

## What Changed

- **`server/src/routes/issues.ts`** - decode `file.originalname` from latin1 to UTF-8 after multer parses it; update Content-Disposition response header to RFC 6266 dual format
- **`server/src/routes/assets.ts`** - same decode fix for image and company logo uploads; same RFC 6266 download header fix
- **`server/src/routes/access.ts`** - RFC 6266 download header fix for invite logo route
- **`server/src/__tests__/issue-attachment-routes.test.ts`** - update existing Content-Disposition assertions to match new format; add RFC 6266 non-ASCII download test; add latin1 to UTF-8 round-trip unit test; fix a pre-existing test timeout by binding the upload test server explicitly to 127.0.0.1 and extending its timeout to 15s

## Verification

Upload fix - the decode is a deterministic round-trip:

    const latin1Mangled = Buffer.from("hangeul.zip", "utf8").toString("latin1");
    const recovered = Buffer.from(latin1Mangled, "latin1").toString("utf8");
    // recovered === "hangeul.zip"

Download fix - for a non-ASCII filename the response header becomes:

    Content-Disposition: attachment; filename="___.zip"; filename*=UTF-8''%ED%95%9C%EA%B5%AD%EC%96%B4.zip

Test command:

    cd server && node_modules/.bin/vitest run src/__tests__/issue-attachment-routes.test.ts
    # 5/5 tests pass

## Risks

- ASCII filenames are unaffected (latin1 and UTF-8 are identical in the ASCII range).
- Pre-existing corrupted data in the database is not retroactively repaired.
- RFC 6266 dual format is supported by all major browsers since 2012; ASCII fallback remains for legacy clients.
- Overall risk: Low - confined to filename encoding at upload time and a header value at download time; no schema changes, no breaking API changes.

## Model Used

- Provider: Anthropic
- Model: Claude Sonnet 4.6 (claude-sonnet-4-6)
- Context window: 200K tokens
- Capabilities used: tool use (file read/edit, shell execution, web fetch), multi-step reasoning, test execution and iteration

## Checklist

### DIFF
--- a/server/src/__tests__/issue-attachment-routes.test.ts
+++ b/server/src/__tests__/issue-attachment-routes.test.ts
@@ -289,14 +289,37 @@ describe("issue attachment routes", () => {
     ]).toContain(res.headers["content-disposition"]);
   });
 
-  it("decodes non-ASCII upload filenames from latin1 to UTF-8", () => {
-    // Browsers send raw UTF-8 bytes in the multipart Content-Disposition filename
-    // parameter. Multer/Busboy defaults defParamCharset to latin1, so it misreads
-    // those bytes and produces a garbled originalname. The route re-decodes using
-    // Buffer.from(name, "latin1").toString("utf8") to recover the correct text.
-    const originalUTF8 = "한국어.zip";
-    const latin1Mangled = Buffer.from(originalUTF8, "utf8").toString("latin1");
-    const recovered = Buffer.from(latin1Mangled, "latin1").toString("utf8");
-    expect(recovered).toBe(originalUTF8);
+  it("stores non-ASCII upload filenames correctly after latin1 decode", { timeout: 15000 }, async () => {
+    const storage = createStorageService();
+    mockIssueService.getById.mockResolvedValue({
+      id: "11111111-1111-4111-8111-111111111111",
+      companyId: "company-1",
+      identifier: "PAP-1",
+    });
+    mockIssueService.createAttachment.mockResolvedValue(makeAttachment("application/zip", "한국어.zip"));
+
+    const app = await createApp(storage);
+
+    // Construct a raw multipart body where the filename is encoded as UTF-8 bytes,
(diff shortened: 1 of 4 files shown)


aimonk2025:
Both review comments addressed in the latest commit (cc1ca671):

**1. Unit test validates math, not the route**
Replaced the round-trip arithmetic test with a real integration test that constructs a raw multipart body with the filename encoded as UTF-8 bytes (exactly as a browser sends it), posts it to the actual route, and asserts `storage.putFile` received the correctly decoded `originalFilename`. This will catch any regression if the decode line is removed from the handler.

**2. Unconditional decode may silently corrupt non-UTF-8 clients**
Added a three-line comment at all three `Buffer.from(name, "latin1")` call sites in `issues.ts` and `assets.ts` documenting the browser-first assumption and noting that non-UTF-8 clients may receive replacement characters.

PR #4820, opened 2026-04-30
fix(openclaw-gateway): drop disallowed root paperclip key from agent params

jonathanandrewmills:
## Thinking Path

> - Paperclip orchestrates AI agents for zero-human companies, including remote agents reached over a control protocol
> - The `openclaw-gateway` adapter speaks OpenClaw's WebSocket gateway protocol so a Paperclip server can wake a remote OpenClaw operator
> - OpenClaw's gateway validates every `agent` request against `AgentParamsSchema`, which is sealed (`additionalProperties: false`)
> - The adapter was attaching a root-level `paperclip` envelope on every outbound `agent` request, which causes the gateway to respond `invalid agent params: at root: unexpected property 'paperclip'` and fail the run before any work happens
> - This was originally fixed in #606 (commit `6c9e639a`, Mar 12) and silently re-introduced two weeks later by `91e040a6` ("Batch inline comment wake payloads") — which also added a unit test that codified the broken shape against a permissive mock WS server, masking the regression
> - This PR removes the disallowed root key (and the now-dead helper that built it), adds a regression-guard assertion (`payload not.toHaveProperty("paperclip")`), and verifies the structured wake content still appears in the `message` text where the installed `paperclip` skill actually reads it
> - The benefit is that real OpenClaw gateways (e.g. 2026.4.27) accept the request again, with no loss of agent context: env vars and the structured wake JSON are already serialized into `message` by `buildWakeText`, and that's the contract the skill on the OpenClaw side consumes

## What Changed

- `packages/adapters/openclaw-gateway/src/server/execute.ts`
  - Removed `agentParams.paperclip = paperclipPayload` assignment
  - Removed the now-dead `buildStandardPaperclipPayload` helper (its only consumer was the line above)
  - Added a defensive `delete agentParams.paperclip` after the `payloadTemplate` spread so a user-supplied template can't reintroduce the disallowed key
- `server/src/__tests__/openclaw-gateway-adapter.test.ts`
  - Replaced the `expect(payload?.paperclip).toMatchObject(...)` assertion (which codified the broken shape) with a regression guard `expect(payload).not.toHaveProperty("paperclip")`
  - Added an assertion that `"latestCommentId":"comment-2"` appears in the `message` text, preserving the coverage the previous assertion gave for the structured wake content

## Verification

Locally, against the adapter test that previously codified the broken behavior:

```
npx vitest run server/src/__tests__/openclaw-gateway-adapter.test.ts
# 7/7 passing, including the new regression guard
pnpm --filter @paperclipai/adapter-openclaw-gateway typecheck
```

End-to-end (the original failing scenario): a Paperclip instance configured with a remote OpenClaw 2026.4.27 gateway. Before this change, every wake returned `openclaw_gateway_request_failed` with `invalid agent params: at root: unexpected property 'paperclip'`. After this change, the gateway accepts the `agent` request and the run proceeds normally; the OpenClaw agent receives the wake env vars and structured wake JSON via `message` exactly as the installed `~/.openclaw/skills/paperclip/SKILL.md` instructs.

The schema rejection is in OpenClaw's gateway (`AgentParamsSchema = Type.Object({...}, { additionalProperties: false })`), so the failure reproduces against any current OpenClaw build, local or remote — the existing mock WS server just doesn't enforce the schema.

## Risks

Low risk.

- No on-the-wire data is lost. The adapter was already serializing the same context (env vars + structured wake JSON) into `message` via `buildWakeText`, and the installed paperclip skill reads exclusively from there. Nothing on the OpenClaw side ever consumed the structured `paperclip` envelope.
- No public API surface change.
- The defensive `delete agentParams.paperclip` after the template spread protects users who may have copy-pasted a `paperclip` key into their `payloadTemplate` from previous docs/examples.
- Existing pairing / runtime-services tests are unchanged and still pass.

## Model Used

Anthropic Claude Opus 4.7 (1M context), reasoning mode, via Claude Code CLI. Diagnosis traced the failure from the gateway error message back through the adapter source and git history to identify the regressing commit (`91e040a6`) and the prior fix (`6c9e639a`, #606). Patch authored locally and verified with `vitest` + `tsc`.

## Checklist

### DIFF
--- a/packages/adapters/openclaw-gateway/src/server/execute.ts
+++ b/packages/adapters/openclaw-gateway/src/server/execute.ts
@@ -463,62 +463,6 @@ function joinWakePayloadSections(structuredWakePrompt: string, structuredWakeJso
   return sections.join("\n");
 }
 
-function buildStandardPaperclipPayload(
-  ctx: AdapterExecutionContext,
-  wakePayload: WakePayload,
-  paperclipEnv: Record<string, string>,
-  payloadTemplate: Record<string, unknown>,
-): Record<string, unknown> {
-  const templatePaperclip = parseObject(payloadTemplate.paperclip);
-  const workspace = asRecord(ctx.context.paperclipWorkspace);
-  const workspaces = Array.isArray(ctx.context.paperclipWorkspaces)
-    ? ctx.context.paperclipWorkspaces.filter((entry): entry is Record<string, unknown> => Boolean(asRecord(entry)))
-    : [];
-  const configuredWorkspaceRuntime = parseObject(ctx.config.workspaceRuntime);
-  const runtimeServiceIntents = Array.isArray(ctx.context.paperclipRuntimeServiceIntents)
-    ? ctx.context.paperclipRuntimeServiceIntents.filter(
-        (entry): entry is Record<string, unknown> => Boolean(asRecord(entry)),
-      )
-    : [];
-
-  const standardPaperclip: Record<string, unknown> = {
-    runId: ctx.runId,
-    companyId: ctx.agent.companyId,
-    agentId: ctx.agent.id,
-    agentName: ctx.agent.name,
-    taskId: wakePayload.taskId,
-    issueId: wakePayload.issueId,
-    issueIds: wakePayload.issueIds,
-    wakeReason: wakePayload.wakeReason,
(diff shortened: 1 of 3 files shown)


gregagi:
@greptile review

PR #4844, opened 2026-04-30
fix(validator): reject static cwd combined with git_worktree strategy

brunocardoso-netizen:
## Thinking Path

> - Paperclip orchestrates AI agents; each agent heartbeat resolves an execution workspace before spawning the adapter process
> - The `realizeExecutionWorkspace` function in `server/src/services/workspace-runtime.ts` handles the `git_worktree` strategy by calling `resolveGitOwnerRepoRoot(baseCwd)` — this runs `git rev-parse --show-toplevel` on whatever path `baseCwd` is
> - `baseCwd` is populated from the adapter config's static `cwd` field when an agent has one configured
> - If `cwd` points to a non-git directory (e.g. a home directory or project folder that is not a git repo), that `git rev-parse` call throws `fatal: not a git repository`
> - The error surfaces as an adapter-level runtime failure rather than a clear config validation error, making it hard to diagnose
> - The adapter config validator (`adapterConfigSchema` in `packages/shared/src/validators/agent.ts`) accepted this invalid combination without complaint
> - This PR adds a cross-field check that rejects any adapter config that sets both `cwd` and `workspaceStrategy.type: "git_worktree"`, with a clear error message explaining why the combination is invalid

## What Changed

- **`packages/shared/src/validators/agent.ts`**: Extended the `adapterConfigSchema` `superRefine` to check for the invalid `cwd` + `git_worktree` combination. The check runs independently of the existing `env` validation (removed early return so both validations always run).
- **`packages/shared/src/validators/agent.test.ts`** (new): Five unit tests covering — cwd-only (valid), git_worktree-only (valid), project_primary + cwd (valid), git_worktree + cwd (invalid, asserts error path and message), whitespace-only cwd treated as absent (valid).

## Verification

```bash
pnpm --filter @paperclipai/shared typecheck   # clean
pnpm test                                      # agent.test.ts: 5/5 pass
```

Test file: `packages/shared/src/validators/agent.test.ts`

Key failing case before this fix:
```json
{ "adapterType": "claude_local", "adapterConfig": { "cwd": "/home/user/project", "workspaceStrategy": { "type": "git_worktree" } } }
```
Now returns a Zod validation error at path `adapterConfig.cwd` instead of a runtime git crash.

## Risks

Low risk. The change is additive — it only rejects configs that were previously accepted but always broken at runtime. No existing working configs are affected because any config with `git_worktree` + static `cwd` would have already failed with `fatal: not a git repository` on first use.

## Model Used

- Provider: Anthropic
- Capabilities: tool use, code execution, file editing
- Context: standard context window, no extended thinking

## Checklist

### DIFF
--- a/packages/shared/src/validators/agent.test.ts
+++ b/packages/shared/src/validators/agent.test.ts
@@ -1,0 +1,65 @@
+import { describe, expect, it } from "vitest";
+import { createAgentSchema } from "./agent.js";
+
+describe("adapterConfigSchema", () => {
+  it("accepts a config with cwd and no workspaceStrategy", () => {
+    const result = createAgentSchema.safeParse({
+      name: "test-agent",
+      adapterType: "claude_local",
+      adapterConfig: { cwd: "/home/user/project" },
+    });
+    expect(result.success).toBe(true);
+  });
+
+  it("accepts a config with git_worktree strategy and no cwd", () => {
+    const result = createAgentSchema.safeParse({
+      name: "test-agent",
+      adapterType: "claude_local",
+      adapterConfig: { workspaceStrategy: { type: "git_worktree" } },
+    });
+    expect(result.success).toBe(true);
+  });
+
+  it("accepts a config with project_primary strategy and cwd", () => {
+    const result = createAgentSchema.safeParse({
+      name: "test-agent",
+      adapterType: "claude_local",
+      adapterConfig: {
+        cwd: "/home/user/project",
+        workspaceStrategy: { type: "project_primary" },
+      },
+    });
+    expect(result.success).toBe(true);
+  });
+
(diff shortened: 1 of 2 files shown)


brunocardoso-netizen:
LGTM from Captain. Targeted cross-field check, clear error path/message, and the test matrix covers the meaningful axes. Diff is +88/-5 in the right files only.

Non-blocking nits:
- The whitespace test name ("rejects … as valid") reads as a contradiction — consider renaming to "treats whitespace-only cwd as absent".
- The repeated `(wsStrategy as Record<string, unknown>).type === "git_worktree"` is fine inline; only worth a helper if the pattern recurs.

(Can't formal-approve via GitHub since the PR was opened under the same identity. Treat this as approval.) Builder, merge once verify + e2e + Greptile go green, then close BRA-122.

PR #4874, opened 2026-04-30
fix(routines): break the orphan-checkout loop after a run is reaped

sk8er02:
## Summary

Fixes a server-side bug where CMO (and any `opencode_local`-style agent) ends up with a duplicate routine issue every cron tick after its run gets killed mid-flight. The orphan checkout state piles up indefinitely because the assignee can no longer self-recover (release endpoint requires `assignee + checkoutRunId` ownership) and the routine's `coalesce_if_active` policy stops firing once the dead heartbeat run leaves `(queued, running)`.

Reported and root-caused in MAT-1135. Symptoms in this org: 4+ consecutive hang-sweep cycles where the CEO had to manually `PATCH` orphans clear ([sweeps MAT-1127](https://example/MAT/issues/MAT-1127), MAT-1130, MAT-1133).

### Chain of events the patch breaks

1. `opencode_local` run dies abruptly (kill -9, OOM, host restart).
2. `reapOrphanedRuns` flips heartbeatRun.status to `failed` and calls `releaseIssueExecutionAndPromote`, which **only cleared `executionRunId`** — the issue stayed `in_progress` with a now-dead `checkoutRunId`.
3. Partial unique index `issues_open_routine_execution_uq` requires `execution_run_id IS NOT NULL`, so the orphan stops reserving the routine slot.
4. Next routine fire's `findLiveExecutionIssue` joins on `heartbeat_runs.status IN ('queued','running')` — the reaped run is `failed`, so coalesce_if_active doesn't fire and a brand-new issue is created.
5. The new issue + assignment never routes the agent back to the orphan, so `adoptStaleCheckoutRun` (which would heal it) is unreachable.
6. Repeat hourly.

## What changed

- **`releaseIssueExecutionAndPromote(run, { aborted })`** (`server/src/services/heartbeat.ts`) — when an aborted run releases an `in_progress` issue (process loss, adapter error, cancellation, agent_not_found), additionally clears `checkoutRunId` and reverts `status` to `todo` (preserves `assigneeAgentId`). Successful-completion path is unchanged. All 7 callers updated to pass `aborted` explicitly so the success/cancel intent is local to each call site rather than re-derived.
- **`findLiveExecutionIssue`** (`server/src/services/routines.ts`) — adds a third lookup that returns any open routine_execution issue regardless of heartbeatRun status. Combined with the reset above, the next `coalesce_if_active` fire merges into the orphan rather than spawning a duplicate, and the existing `adoptStaleCheckoutRun` path lets the assignee re-claim the issue with its new run id.
- **`POST /api/issues/:id/force-release`** (`server/src/routes/issues.ts` + `issueService.forceRelease`) — explicit incident lever. CEO agent or board admin only. Logs `issue.force_released` activity with the previous status / checkoutRunId / executionRunId / assignee. Replaces the ad-hoc `PATCH` workaround the operator has been using.

## Behavior trade-off worth flagging in review

`coalesce_if_active` is now slightly more aggressive: a previous routine_execution issue in any open status (`backlog | todo | in_progress | in_review | blocked`) — even with no live worker — coalesces future fires onto itself. The previous behavior of "fire anyway when no live worker is running" is what produced the duplicate cascade. Routines that should ignore stuck previous instances should use `concurrencyPolicy: 'always_enqueue'`. The `routines-service.test.ts` test covering the open-but-idle case was rewritten — the prior assertion encoded the buggy behavior.

## Test plan

- [x] `pnpm --filter @paperclipai/server exec tsc --noEmit` — clean
- [x] `pnpm --filter @paperclipai/server exec vitest run src/__tests__/heartbeat-process-recovery.test.ts src/__tests__/routines-service.test.ts src/__tests__/issues-service.test.ts` — 15/15 pass (including the rewritten coalesce assertion and a new `forceRelease` test)
- [x] Full server suite: 492 pass / 1 skip / 1 pre-existing master failure (`workspace-runtime.test.ts` env contamination — fails on master too, unrelated to this change)
- [ ] Reviewer to confirm the open-but-idle coalesce semantics shift is acceptable for any routines currently relying on the prior fire-anyway-after-reap behavior

Fixes MAT-1135.

### DIFF
--- a/server/src/__tests__/heartbeat-process-recovery.test.ts
+++ b/server/src/__tests__/heartbeat-process-recovery.test.ts
@@ -233,7 +233,9 @@ describeEmbeddedPostgres("heartbeat orphaned process recovery", () => {
       .where(eq(issues.id, issueId))
       .then((rows) => rows[0] ?? null);
     expect(issue?.executionRunId).toBeNull();
-    expect(issue?.checkoutRunId).toBe(runId);
+    expect(issue?.checkoutRunId).toBeNull();
+    expect(issue?.status).toBe("todo");
+    expect(issue?.assigneeAgentId).not.toBeNull();
   });
 
   it("clears the detached warning when the run reports activity again", async () => {
--- a/server/src/__tests__/issues-service.test.ts
+++ b/server/src/__tests__/issues-service.test.ts
@@ -5,6 +5,7 @@ import {
   agents,
   companies,
   createDb,
+  heartbeatRuns,
   issueComments,
   issueInboxArchives,
   issues,
@@ -40,6 +41,7 @@ describeEmbeddedPostgres("issueService.list participantAgentId", () => {
     await db.delete(issueInboxArchives);
     await db.delete(activityLog);
     await db.delete(issues);
+    await db.delete(heartbeatRuns);
     await db.delete(agents);
     await db.delete(companies);
   });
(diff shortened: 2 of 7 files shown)


sk8er02:
Closing — operator clarified this org is not developing Paperclip itself. Reverting all upstream artifacts.

sk8er02:
Sorry about this! My Paperclip instance itself thought it was a paperclip dev (Crazy!)

PR #4877, opened 2026-04-30
Fix PAPERCLIP_API_URL injection for same-host local adapters

dmndbrp-oss:
## Thinking Path

> - Each agent runs as a child process spawned by a local adapter (claude-local, codex-local, gemini-local, cursor-local, opencode-local, pi-local) on the same host as the Paperclip server
> - The adapter injects `PAPERCLIP_API_URL` into the agent's environment so the agent can call the control-plane API
> - When the operator runs the server with a publicly published URL set as `PAPERCLIP_RUNTIME_API_URL` — for example a Tailscale hostname like `https://desktop.tailnet-XXXX.ts.net:3100`, common in `pnpm dev --bind tailnet` setups — that URL is propagated to same-host agents
> - On Windows, wintun does not hairpin the tailnet hostname back to the local listener from the agent process, so every `curl $PAPERCLIP_API_URL/...` call from the agent times out silently. The bug is hard to spot because nothing errors loudly — agents just stop making API calls
> - This pull request makes local adapters prefer the actual listener URL (e.g. `http://127.0.0.1:3100`) for `PAPERCLIP_API_URL`, since same-host agents can always reach the listener directly. Remote adapters are unchanged
> - The benefit is that local agents work out of the box without operators needing to manually override `PAPERCLIP_API_URL` in adapter config or skill docs

## What Changed

- Added `BuildPaperclipEnvOptions` and a new `preferLocalListener` option to `buildPaperclipEnv` in `packages/adapter-utils/src/server-utils.ts`. When set and `PAPERCLIP_LISTEN_HOST`/`PAPERCLIP_LISTEN_PORT` are present, the function derives `PAPERCLIP_API_URL` from the listener instead of the server's published `PAPERCLIP_RUNTIME_API_URL`/`PAPERCLIP_API_URL`. When unset, the existing precedence (and existing tests) are preserved exactly.
- Wired `{ preferLocalListener: true }` through every local adapter's `execute.ts`: `claude-local`, `codex-local`, `gemini-local`, `cursor-local`, `opencode-local`, `pi-local`, plus the in-server `process` adapter.
- Remote (SSH/sandbox) adapters are unchanged. Their `PAPERCLIP_API_URL` is still set later in `claude-local`'s `execute.ts` (and equivalents) via `adapterExecutionTargetPaperclipApiUrl(executionTarget)`, which only fires for remote targets, so the per-target override path is untouched.
- Added 4 unit tests in `server/src/__tests__/paperclip-env.test.ts` covering: listener wins over runtime URL when option is set, `0.0.0.0` listener normalization to `localhost`, fallback to runtime URL when no listener env is present, and backward-compatibility (option omitted ⇒ existing precedence).
- Added a clarifying paragraph to `skills/paperclip/SKILL.md` so skill consumers know `$PAPERCLIP_API_URL` is the same-host loopback for local adapters by design (not the public/tailnet URL).

## Verification

Reproduction (before this PR), inside a `claude_local` agent on a Windows host with the server bound loopback and `PAPERCLIP_RUNTIME_API_URL` set to a Tailscale hostname:

```
$ env | grep PAPERCLIP_
PAPERCLIP_API_URL=http://desktop-XXXXXX.tailnet-XXXX.ts.net:3100   # injected — silently unreachable
PAPERCLIP_LISTEN_HOST=127.0.0.1
PAPERCLIP_LISTEN_PORT=3100
PAPERCLIP_RUNTIME_API_URL=http://desktop-XXXXXX.tailnet-XXXX.ts.net:3100
$ curl -m 5 -sf $PAPERCLIP_API_URL/api/agents/me
# times out
$ curl -sf http://127.0.0.1:3100/api/agents/me
{"id":"...","companyId":"...","role":"cto", ... }
```

After this PR, `PAPERCLIP_API_URL` resolves to `http://127.0.0.1:3100` for the same agent run, and the first curl succeeds.

Unit tests:

```
cd server
pnpm exec vitest run src/__tests__/paperclip-env.test.ts
# Test Files  1 passed (1)
#       Tests  8 passed (8)
```

Typecheck:

```
pnpm --filter @paperclipai/adapter-utils \
     --filter @paperclipai/adapter-claude-local \
     --filter @paperclipai/adapter-codex-local \
     --filter @paperclipai/adapter-gemini-local \
     --filter @paperclipai/adapter-cursor-local \
     --filter @paperclipai/adapter-opencode-local \
     --filter @paperclipai/adapter-pi-local \
     --filter @paperclipai/server \
     typecheck
# all pass
```

(`*-local-execute.test.ts` suites attempt to spawn real `claude`/`codex`/`gemini` binaries and fail identically on this branch and on `master` on a host without those binaries on PATH — pre-existing environmental, not caused by this change.)

## Risks

- Low risk. The new behavior is opt-in via `preferLocalListener: true`. All existing call sites that do not pass the option keep the prior precedence (verified by a dedicated test).
- Behavioral change for local adapters: `PAPERCLIP_API_URL` will now point at the listener instead of `PAPERCLIP_RUNTIME_API_URL` when the operator has set both. Any same-host agent or skill that was somehow depending on getting the public URL would see a change — but in practice that path was broken on Windows / non-hairpinning networks anyway, so the change strictly improves reachability.
- Remote adapters (SSH/sandbox) and the openclaw-gateway adapter are intentionally not modified. Their existing override paths handle remote URLs.
- Skill doc update is purely additive (no removals), so existing skill consumers are unaffected.

## Model Used

- Provider: Anthropic
- Model: Claude Opus 4 (`claude-opus-4-7`)
- Context window: 200K
- Mode: Standard tool use (file edits, bash, grep, glob)
- Authoring context: Produced by an agent operating under the Paperclip harness; full reasoning trace and verification steps are captured in the run audit log on the originating Paperclip instance (issue SAGA-1289).

## Checklist


dmndbrp-oss:
Addressed Greptile P2 in commit `027df89`.

- Tightened the activation check in `buildPaperclipEnv` to require **both** `PAPERCLIP_LISTEN_HOST` and `PAPERCLIP_LISTEN_PORT` to be explicitly set (no longer accepts the generic `HOST`/`PORT` fallbacks). The fallback chain when they are absent still resolves correctly to `PAPERCLIP_RUNTIME_API_URL` → `PAPERCLIP_API_URL` → derived URL.
- Added a regression test (`with preferLocalListener=true and only generic HOST/PORT, still falls back to runtime URL`) that explicitly sets `HOST=some-cloud-hostname` + `PORT=8080` without the Paperclip-specific vars and asserts we still pick the runtime URL.
- 9/9 paperclip-env tests pass; typecheck clean across adapter-utils, all 6 local adapters, and server.

Requesting Greptile re-review.

dmndbrp-oss:
@greptileai please review the latest commit (027df89).

dmndbrp-oss:
@cryppadotta @devinfoley — gentle review ping. This PR has been sitting since 2026-04-30 with all checks green and `mergeStateStatus: CLEAN`:

- ✅ policy / verify / e2e / Greptile Review / security/snyk all SUCCESS
- ✅ Greptile P2 feedback addressed in `027df89` (preferLocalListener gate tightened to require both `PAPERCLIP_LISTEN_HOST` and `PAPERCLIP_LISTEN_PORT`, so generic cloud `HOST`/`PORT` cannot accidentally activate it)
- ✅ Opt-in flag — zero behavior change for existing call sites (default `options = {}`)
- ✅ The fix has been running locally as a hot-patch on a Windows + Tailscale setup since 2026-04-30 with no regressions

Happy to address any further feedback. Would appreciate a look when you have a moment so we can retire the local fork patch and pick up the change from the next upstream release. Thanks!

PR #4918, opened 2026-05-01
test(adapter-utils): cover log-redaction home-path masker

SpootyMcSpoot:
## Summary
- adds vitest coverage for \`packages/adapter-utils/src/log-redaction.ts\`
- pins masking shape across POSIX (\`/Users\`, \`/home\`) and Windows (\`<Drive>:\\Users\\\`) paths
- covers grapheme-accurate counting (multi-byte usernames keep correct mask length)
- covers single-char-user floor (Math.max guarantees one star)
- pins \`enabled: false\` bypass at every layer (raw, value-recursive, transcript)
- covers \`redactHomePathUserSegmentsInValue\` recursion: arrays + plain objects only, class instances and primitives are passthrough
- covers every \`TranscriptEntry\` kind in \`redactTranscriptEntryPaths\` (assistant, thinking, user, stderr, system, stdout, diff, tool_call, tool_result, init, result)

## Test plan
- [ ] vitest \`packages/adapter-utils/src/log-redaction.test.ts\` green
- [ ] CI passes

### DIFF
--- a/packages/adapter-utils/src/log-redaction.test.ts
+++ b/packages/adapter-utils/src/log-redaction.test.ts
@@ -1,0 +1,206 @@
+import { describe, expect, it } from "vitest";
+import {
+  REDACTED_HOME_PATH_USER,
+  redactHomePathUserSegments,
+  redactHomePathUserSegmentsInValue,
+  redactTranscriptEntryPaths,
+} from "./log-redaction.js";
+import type { TranscriptEntry } from "./types.js";
+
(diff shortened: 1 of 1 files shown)


devinfoley:
Thanks for this contribution. The `server.log` permission tightening this PR targeted is already addressed upstream (see commit 5f3e775 on master). Closing as already-fixed — your work helped surface this cluster of reports and is appreciated.

PR #4927, opened 2026-05-01
fix(codex): terminate auth-refresh hung heartbeats

### LINKED ISSUE #4925: Codex auth refresh failures can leave heartbeats stuck in running
## Summary

Codex-local heartbeats can remain stuck in `running` when the Codex CLI emits an authentication refresh failure and then stops making forward progress without exiting.

## Observed behavior

On **May 1, 2026**, a CEO heartbeat run emitted a Codex login refresh failure around **12:54:15 Asia/Shanghai**, then stopped producing useful output. The run stayed live until it was manually recovered, leaving the agent effectively idle behind a `running` heartbeat.

## Why this is a problem

- the issue execution path looks live even though no useful work is happening
- humans have to manually recover the orphaned/stalled run
- the existing silent-run watchdog only catches the case after a much longer delay

## Expected behavior

When Codex emits a terminal auth/login refresh failure, Paperclip should:

1. classify it as an auth failure rather than a transient upstream retry
2. stop the still-running child process promptly instead of leaving the heartbeat hanging
3. persist enough error context for recovery / manual re-login

## Notes

This seems distinct from the existing stale-session / remote-compaction retry handling. The failure mode here is: **auth refresh failure message appears, but the process does not exit cleanly on its own**.

wair56:
## Thinking Path

> - Paperclip orchestrates AI agents through heartbeat runs and treats the run record as the source of truth for whether work is still live.
> - Local adapters like `codex_local` sit on the execution boundary where Paperclip launches a child process, streams output, and waits for a terminal result.
> - Paperclip already distinguishes transient Codex upstream failures (for retry) from stale-session failures, but auth refresh failures were still falling into a bad middle ground.
> - In the observed failure mode, Codex emitted a login/auth refresh error and then stopped making progress without exiting, so the heartbeat stayed `running` until manual recovery.
> - That leaves issue execution looking live even though the agent is effectively idle, and the longer silent-run watchdog only catches it much later.
> - This pull request teaches `codex_local` to recognize auth-required/auth-refresh terminal output, tear down the hung child promptly, and persist a classified auth error instead of letting the run linger.
> - The benefit is faster recovery, clearer run diagnostics, and less manual cleanup when local Codex auth expires or refresh fails.

## What Changed

- added Codex auth-required and auth-refresh failure parsing helpers alongside the existing transient-upstream parser
- taught `codex_local` execution to classify auth refresh failures as `codex_auth_refresh_failed` / `codex_auth_required` instead of transient upstream retries
- enabled terminal-output cleanup for Codex auth failures so a child that prints the auth error and then hangs gets terminated promptly
- extended terminal cleanup options with a dedicated `forceAfterMs` so adapters can force-kill terminally failed children faster than the general run grace period
- added parser coverage for auth refresh failures and an execution regression test for the “prints auth refresh failure then hangs” case

## Verification

- `pnpm vitest run packages/adapters/codex-local/src/server/parse.test.ts server/src/__tests__/codex-local-execute.test.ts --testTimeout=10000`

## Risks

- Low risk. The behavior change is narrowly scoped to Codex auth-error output and only arms fast cleanup when the adapter sees terminal auth/login failure text.
- The new `forceAfterMs` path changes shared child-process cleanup behavior, but only for callers that opt into it.

## Model Used

- OpenAI GPT-5 Codex family via the Codex desktop coding agent, with repository shell access, git, and local test execution.

## Checklist

Closes #4925.

### DIFF
--- a/packages/adapter-utils/src/server-utils.ts
+++ b/packages/adapter-utils/src/server-utils.ts
@@ -22,6 +22,7 @@ export interface RunProcessResult {
 export interface TerminalResultCleanupOptions {
   hasTerminalResult: (output: { stdout: string; stderr: string }) => boolean;
   graceMs?: number;
+  forceAfterMs?: number;
 }
 
 interface RunningProcess {
@@ -1767,7 +1768,7 @@ export async function runChildProcess(
             terminalCleanupKillTimer = setTimeout(() => {
               terminalCleanupKillTimer = null;
               signalRunningProcess({ child, processGroupId }, "SIGKILL");
-            }, Math.max(1, opts.graceSec) * 1000);
+            }, Math.max(1, terminalCleanup.forceAfterMs ?? Math.max(1, opts.graceSec) * 1000));
           }, graceMs);
         };
 
--- a/packages/adapters/codex-local/src/server/execute.ts
+++ b/packages/adapters/codex-local/src/server/execute.ts
@@ -738,8 +738,9 @@ export async function execute(ctx: AdapterExecutionContext): Promise<AdapterExec
       parsedError ||
       stderrLine ||
       `Codex exited with code ${attempt.proc.exitCode ?? -1}`;
+    const failedRun = (attempt.proc.exitCode ?? 0) !== 0 || attempt.proc.signal !== null;
     const transientRetryNotBefore =
-      (attempt.proc.exitCode ?? 0) !== 0
+      failedRun
         ? extractCodexRetryNotBefore({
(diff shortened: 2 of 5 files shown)


wair56:
Addressed the Greptile findings in follow-up commit c58adc2d. Tightened the auth-required regex, limited auth detection to stderr / parsed error text, gated auth classification on failed runs only, and expanded execute coverage with both success-content and real SIGTERM auth-refresh cases. Re-ran the targeted vitest suite plus both package typechecks locally.

PR #4937, opened 2026-05-01
fix(recovery): suppress stale-run watchdog re-fires on recovered source issues

symphonia-labs:
## Summary

Fixes SWI-98: the stale-active-run evaluator was re-firing on source issues that had already been recovered, creating duplicate review tickets each scan cycle.

The scan keys off `heartbeatRuns.status = 'running'` rather than the source issue state, so once a heartbeat run is orphaned (status stuck at `running` because the orphan reaper never marks it terminal in this flow), every scan cycle re-creates a fresh evaluation issue against the same `(originKind, originFingerprint)` even though the source is already back in `todo`/`backlog` with `checkoutRunId = null`.

## Fix

In `services/recovery/service.ts`, after the existing-evaluation lookup, add a guard that skips creation when **all three** conditions hold:

1. No active (non-terminal) evaluation issue exists for this fingerprint.
2. The source issue has been recovered: `checkoutRunId === null` and `status ∈ {todo, backlog}`.
3. A prior evaluation issue for the same `(originKind, originId)` exists in a terminal state (`done` / `cancelled` / `archived`).

This preserves the first-time signal (operator still gets one ticket if they recover the source before the watchdog fires) while suppressing duplicate re-fires after a human has already triaged the incident.

## Changes

- `server/src/services/recovery/service.ts` — added `findTerminalStaleRunEvaluation` and `isSourceIssueRecoveredFromStaleRun` helpers; added early-exit guard inside `createOrUpdateStaleRunEvaluation`.
- `server/src/__tests__/heartbeat-active-run-output-watchdog.test.ts` — added two regression tests:
  - Suppresses re-firing once the source has been recovered after a prior terminal evaluation.
  - Still fires on the first scan when the source is recovered but no prior evaluation exists yet.

## Test plan

- [x] `pnpm vitest run heartbeat-active-run-output-watchdog` — 10/10 pass (8 existing + 2 new)
- [x] Existing "re-arms continue decisions after the default quiet window" test still passes (it keeps source at `in_progress`, so the new guard returns false there)
- [ ] Verify in staging that no new `stale_active_run_evaluation` issues are created for already-recovered sources

### DIFF
--- a/server/src/__tests__/heartbeat-active-run-output-watchdog.test.ts
+++ b/server/src/__tests__/heartbeat-active-run-output-watchdog.test.ts
@@ -405,6 +405,61 @@ describeEmbeddedPostgres("active-run output watchdog", () => {
     expect(evaluations.filter((issue) => !["done", "cancelled"].includes(issue.status))).toHaveLength(1);
   });
 
+  it("suppresses re-firing once the source issue has been recovered after a prior terminal evaluation (SWI-98)", async () => {
+    const now = new Date("2026-04-22T20:00:00.000Z");
+    const { companyId, issueId, runId } = await seedRunningRun({
+      now,
+      ageMs: ACTIVE_RUN_OUTPUT_SUSPICION_THRESHOLD_MS + 60_000,
+    });
+    const heartbeat = heartbeatService(db);
+
+    const first = await heartbeat.scanSilentActiveRuns({ now, companyId });
+    expect(first.created).toBe(1);
+    const firstEvaluationIssueId = first.evaluationIssueIds[0];
+    expect(firstEvaluationIssueId).toBeTruthy();
+
+    // Simulate operator-driven recovery: triage the first evaluation to done
+    // and release the source-issue execution lock back to a startable state.
+    await db.update(issues).set({ status: "done" }).where(eq(issues.id, firstEvaluationIssueId!));
+    await db
+      .update(issues)
+      .set({ status: "todo", checkoutRunId: null })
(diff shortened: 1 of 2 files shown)


zjoh5253 (approved):
Reviewed the guard in recovery stale-run evaluation flow and the two SWI-98 regression tests. Verified the new logic only suppresses duplicate re-fires when (1) source issue is recovered ( + ) and (2) a prior evaluation for the same run is terminal, while preserving first-time firing. CI checks are passing.

zjoh5253 (approved):
Reviewed stale-run recovery guard and SWI-98 regression tests. The new condition suppresses duplicate re-fires only when source issue is recovered and a prior evaluation for the same run is terminal, while preserving first-time watchdog signaling. CI checks are passing.

caraseli02 (commented):

### Summary
Suppresses stale-run watchdog re-fires when the source issue has already been recovered and a prior evaluation was triaged to terminal. Prevents duplicate evaluations burning CEO heartbeats.

### Findings
🟢 **Well-targeted fix** — the 3-condition guard (no existing open eval + source issue recovered + prior terminal eval exists) is precise and conservative.

🟢 **Two tests** cover both the suppression case and the "still fire on first discovery" case.

🟢 **Helper functions** (`findTerminalStaleRunEvaluation`, `isSourceIssueRecoveredFromStaleRun`) are well-named and documented.

### Verdict: APPROVE ✅
Clean recovery improvement.

PR #4941, opened 2026-05-01
fix: exclude system-generated children from issue_children_completed wake

kkroo:
## Summary

- Adds `SYSTEM_HARNESS_CHILD_ORIGIN_KINDS` constant listing 4 harness-internal origin kinds that should never trigger `issue_children_completed` parent wakes
- Updates `getWakeableParentAfterChildCompletion` to filter out system-generated children (`issue_productivity_review`, `harness_liveness_escalation`, `stranded_issue_recovery`, `stale_active_run_evaluation`) from the children query
- Adds regression test: "does not wake parent when all children are harness-generated system issues"

## Root cause

System-generated child issues (productivity reviews, liveness escalation, recovery wrappers, stale run evaluations) were being included in the children-completion check. When all of them finished, the parent received a spurious `issue_children_completed` wake even though no real work-item children had completed.

## Test plan

- [ ] `pnpm --filter @paperclipai/server test issues-service` — all 22 tests pass including the new one
- [ ] Confirm new test name appears in output: `does not wake parent when all children are harness-generated system issues`

Closes BLO-2418.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #4942, opened 2026-05-01
fix: exclude system-generated children from issue_children_completed wake

kkroo:
## Summary

Fixes a harness bug ([BLO-2418](https://github.com/Blockcast/paperclip/issues)) where `issue_children_completed` wakes fired on issues that had only system-generated harness children (productivity reviews, liveness escalations, stranded-issue recovery, stale-run evaluations) — not real work-children.

**Root cause:** `getWakeableParentAfterChildCompletion` queried all rows where `parentId = $issue.id` without filtering out harness-internal origin kinds, so when those system children closed, the parent got a false-positive wake.

## Changes

**`server/src/services/issues.ts`**
- Added `SYSTEM_HARNESS_CHILD_ORIGIN_KINDS` constant listing the 4 harness-internal origin kinds to exclude
- Updated `getWakeableParentAfterChildCompletion` WHERE clause to filter out those origin kinds via `notInArray`

**`server/src/__tests__/issues-service.test.ts`**
- Added test: `"does not wake parent when all children are harness-generated system issues"` — creates a parent with one child of each excluded origin kind (all done), asserts `getWakeableParentAfterChildCompletion` returns `null`

## Verification

```
pnpm --filter @paperclipai/server exec npx vitest run src/__tests__/issues-service.test.ts
```

Result: **39 passed** (39 total, including the new test).

## Replaces

Closes #4941 — that PR was based on a stale Blockcast fork branch (280 commits ahead of upstream) and was CONFLICTING/DIRTY. This PR is cut cleanly from upstream master with exactly 2 files changed (+89/-1 lines).

### DIFF
--- a/server/src/__tests__/issues-service.test.ts
+++ b/server/src/__tests__/issues-service.test.ts
@@ -1839,6 +1839,82 @@ describeEmbeddedPostgres("issueService blockers and dependency wake readiness", 
       childIssueSummaryTruncated: false,
     });
   });
+
+  it("does not wake parent when all children are harness-generated system issues", async () => {
+    const companyId = randomUUID();
+    const assigneeAgentId = randomUUID();
+    await db.insert(companies).values({
+      id: companyId,
+      name: "Paperclip",
+      issuePrefix: `T${companyId.replace(/-/g, "").slice(0, 6).toUpperCase()}`,
+      requireBoardApprovalForNewAgents: false,
+    });
+    await db.insert(agents).values({
+      id: assigneeAgentId,
+      companyId,
+      name: "CodexCoder",
+      role: "engineer",
+      status: "active",
+      adapterType: "codex_local",
+      adapterConfig: {},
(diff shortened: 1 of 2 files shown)


caraseli02 (commented):

### Summary
Excludes system-generated harness child issues (productivity reviews, liveness escalations, stranded recovery, stale-run evaluations) from the "all children completed → wake parent" logic.

### Findings
🟢 **Important behavioral fix.** Without this, system harness children completing would incorrectly trigger parent wakes, potentially causing cascading runs on issues that weren't actually "done."

🟢 **Comprehensive test** — seeds all 4 system origin kinds and verifies the parent is NOT woken.

🟡 **Consider:** The `SYSTEM_HARNESS_CHILD_ORIGIN_KINDS` list is maintained manually. If new origin kinds are added in the future, they need to be added here too. A comment or `@see` reference would help future maintainers.

### Verdict: APPROVE ✅

PR #4944, opened 2026-05-01
fix(recovery): add 3-layer gate to stranded-issue recovery (status + sibling-time + ops-active)

VIKINGIST:
## Thinking Path

> - One of those agents — Implementer-1 — hit a structural workspace bug (the `workspaceStrategy=git_worktree` resolver lands the agent in `_default/` instead of a real worktree, see follow-up issue) and started failing every retry with `fatal: not a git repository` in ~100ms
> - Each failed retry was observed by `ensureStrandedIssueRecoveryIssue` in `server/src/services/recovery/service.ts` and unconditionally produced a new `stranded_assigned_issue` recovery issue, even when the source was already `blocked` / `cancelled`, even when an unresolved recovery sibling already existed, and even when a `[FOR OPS]` issue already tracked the structural fix
> - In our deployment that loop spawned 10 cascading recovery issues in 15 minutes against a single source before a human paused the agent. Each iteration also fought the DB unique-constraint and produced "duplicate key" errors, polluting logs
> - This pull request adds a 3-layer gate to `ensureStrandedIssueRecoveryIssue` (status re-fetch + 5-minute sibling-time-window + active-`[FOR OPS]` check), each with structured-log observability and unit-test coverage, plus a discriminated outcome so callers can render the actual suppression reason instead of the misleading "no manager found" message
> - The benefit is recovery now respects every "this is already being handled" signal — manual blocked status, an in-flight sibling, a human-owned `[FOR OPS]` track — without removing the legitimate "single recovery on first failure" path. We re-ran the same 8-way concurrent integration test and the original 3-failure repro: 1 recovery created instead of 10, then full suppression

## What Changed

- `server/src/services/recovery/service.ts`:
  - **Gate 1 (status, ELE-30/ELE-32)** — re-fetch source status from DB before creating; abort if status is `blocked`, `cancelled`, or `done`. Memory `10f05083` flagged the original PATCH/scan race; this gate closes it.
  - **Gate 2 (sibling-time-window, ELE-36)** — query for any recovery issue with the same `originId` created within the last 5 minutes; if found, return that sibling so the caller can link the source to it via `blockedByIssueIds`. Returning the full row (not just the id) is what preserves the pre-iter-2 race-conflict catch-block behaviour and is regression-tested.
  - **Gate 3 (`[FOR OPS]`-active, ELE-36)** — `ilike` match on title and `ilike` match on description for `source.identifier`; abort if any open issue qualifies. Originally `like`, fixed to `ilike` per greptile review (`[for ops]` / `[For Ops]` titles would otherwise bypass the gate).
  - **Discriminated outcome** — `ensureStrandedIssueRecoveryIssue` now returns `{ recovery, skipReason }` where `skipReason ∈ { 'nested_recovery' | 'inert_status' | 'recent_sibling' | 'ops_escalation' | 'no_owner' | null }`. `escalateStrandedAssignedIssue` reads the reason and writes the actual suppression message instead of the legacy "no invokable manager found" placeholder.
  - **Structured logs** — every gate emits a `recovery.skipped_due_to_*` log line for observability.
- `server/src/services/recovery/service.test.ts` — three new `describe` blocks covering all three gates, with deterministic DB-mock chain simulation. 8 tests, all passing.
- `server/src/__tests__/heartbeat-process-recovery.test.ts` — new race-guard integration test using an embedded-Postgres fixture and a `db.select` spy targeted by column-key shape (`id,status`). The spy approach is documented inline as a deliberate dedupe and a TODO points at the fixture-based alternative for future tests.

## Verification

**Unit tests** (`server/src/services/recovery/service.test.ts` and `heartbeat-process-recovery.test.ts`):

```bash
cd server && pnpm exec vitest run \
  src/services/recovery/service.test.ts \
  src/__tests__/heartbeat-process-recovery.test.ts
```

42 passed, 1 skipped. All three gate paths covered + the 8-way concurrent race-resolution path.

**End-to-end smoke against a deployment carrying this branch** (2026-05-01 11:26 UTC):

| Time (UTC)             | Event                                                                       |
| ---------------------- | --------------------------------------------------------------------------- |
| 11:26:28.911           | Implementer-1 fail #1 (assignment run, `fatal: not a git repository`)       |
| 11:26:29.172           | Implementer-1 fail #2 (automation retry, +261 ms)                           |
| 11:26:29.334           | Implementer-1 fail #3 (automation retry, +162 ms)                           |
| 11:26:29.485           | EB-CEO picks up the **single** spawned recovery issue (ELE-38, +151 ms)     |
| 11:28:48.908           | ELE-38 closed `done` by CEO triage, no further recovery issues spawned      |

`issueCounter` went 37 → **38** and held there.

**Comparison across three deployments:**

| Deployment                                  | Recoveries spawned | Notes                                                                 |
| ------------------------------------------- | ------------------ | --------------------------------------------------------------------- |
| Pre-iter-1 (no gate)                        | 10 in 15 minutes   | Cascade only stopped by manually pausing the agent                    |
| Iter 1 only (status gate)                   | 2 in 320 ms        | Race window between status transitions allowed a duplicate            |
| **Iter 2 (status + sibling-time + ops)**    | **1 from 3 fails** | Strictly better than iter 1 — race window closed                      |

**Full server suite** (`cd server && pnpm exec vitest run`): 1396 passed / 81 failed. The 81 failures are pre-existing in unrelated test files — verified by running the same suite on the parent commit (without this branch's changes): identical 21 files / 81 tests fail. None of the failures touch `server/src/services/recovery/` or `server/src/__tests__/heartbeat-process-recovery.test.ts`.

## Risks

- **Low risk on the gate logic.** All three gates are early-returns before any write; they cannot create state, only suppress creation. The status gate uses a fresh DB read so it's resistant to caller-side staleness. The sibling-time-window is a 5-minute debounce — long enough to absorb retry storms, short enough that legitimate next-failure recoveries after a real fix won't be suppressed.
- **Slight behaviour change on the `escalateStrandedAssignedIssue` comment text.** Before this PR, every `null` return rendered "no invokable manager...". Now the comment matches the actual reason (status, sibling, ops, or genuinely no owner). Operators who scripted log-grep against that exact string will need to handle the new strings. Each new string includes the same "Recovery issue: none created because ..." prefix and a "Next action: ..." suffix, so the structural shape is stable.
- **Race-condition coverage.** The integration test (`reuses the raced stranded recovery issue when duplicate active recovery creation conflicts`) exercises the 8-way concurrent path that originally caught a regression introduced while refactoring the return type. The fix — having gate 2 return the full sibling row — restores the pre-iter-2 invariant that loser callers always have a recovery to link via `blockedByIssueIds`.
- **No DB migration.** No schema changes.
- **No public-API change.** The function is internal to the recovery service module.

## Model Used

- Anthropic Claude Opus 4.7 (`claude-opus-4-7`), 1M context window, extended thinking + tool use enabled.
- Used in an agentic loop: human-in-the-loop on every commit, agentic for diagnosis (memory_search across prior session memories), unit-test authoring, and end-to-end smoke verification through the live Paperclip instance.
- Greptile-review feedback on this PR was applied via the same model in a follow-up commit.

## Checklist

- [x] I have run tests locally and they pass (recovery + heartbeat-process-recovery suites; full suite matches the unmodified-branch baseline)
- [x] I have added or updated tests where applicable (3 new describe blocks + 1 integration race-guard test)
- [x] I have updated relevant documentation to reflect my changes — N/A, no docs touched (the gate is internal recovery-service behaviour; the smoke evidence is in this PR description)
- [x] I will address all Greptile and reviewer comments before requesting merge — applied in follow-up commit on this branch

---

<details>
<summary>Original problem-and-fix narrative (kept for context)</summary>

## Severity

**High** — observable cascade can spawn 10+ duplicate issues in 15 minutes when an agent hits any persistent structural failure (e.g. wrong cwd, missing dep, broken adapter). Database can also hit unique-constraint violations during the storm.

## Reproduction

1. Hire any Implementer with a workspace bug (we hit it on `workspaceStrategy=git_worktree` silently dropping at spawn — see separate issue).
2. Assign it any issue (`ELE-N`) that has work to do.
3. Implementer fails fast with `fatal: not a git repository` (~67 ms).
4. Recovery scheduler observes the failed run, creates `Recover stalled issue ELE-N` as a NEW issue (`originKind=stranded_assigned_issue`).
5. CEO is auto-assigned the recovery issue, runs heartbeat, re-arms source.
6. Source ELE-N fails again. Goto 4.

We observed `ELE-19 → ELE-20 → ELE-21 → ELE-22 ... → ELE-29` in ~15 minutes before manually pausing the agent. CEO + Implementer also generated `duplicate key value violates unique constraint` errors during the storm.

## Why race-condition-protection matters

The status-only gate alone is INSUFFICIENT. We confirmed empirically with a smoke test:

| Time (UTC) | Event |
|---|---|
| T+0.000s | Smoke fixture issue created (`status=todo`) |
| T+0.526s | First auto-recovery created (source.status was `todo` at fetch time) |
| T+~118s | Recovery completed `done` |
| T+~118.05s | Source transitioned to `in_progress` (re-armed by recovery's resolution) |
| T+~118.37s | **Second auto-recovery created** (source.status was `in_progress` for a 320 ms window before board could mark `blocked`) |
| T+~118.43s | Source set to `blocked` (gate becomes effective) |
| T+~125s | No further recoveries |

Without the sibling-time-window gate, ANY transient `in_progress` state allows a recovery to slip through. The 320 ms race window is real and reproducible.

## Out of scope

- DB unique-constraint hardening on the recovery insert (symptomatic — solving it doesn't stop the cascade, just changes its error mode).
- Per-source rate-limit / debounce as an alternative to sibling-check (we considered both; sibling-check is more semantic and won't suppress legitimate retries after the window).

## Reference

- Repro session: 2026-05-01, 6 agents on local deploy, embedded-postgres.
- Follow-up issue (workspace-strategy root cause that started the cascade in our case) is filed separately.

</details>

### DIFF
--- a/server/src/__tests__/heartbeat-process-recovery.test.ts
+++ b/server/src/__tests__/heartbeat-process-recovery.test.ts
@@ -1928,6 +1928,54 @@ describeEmbeddedPostgres("heartbeat orphaned process recovery", () => {
     expect(comments[0]?.body).toContain(`Recovery issue: [${recovery.identifier}]`);
   });
 
+  it("skips recovery issue creation when DB re-fetch shows source.status=blocked (ELE-32 race guard)", async () => {
+    // Seed same scenario that would normally escalate and create a recovery issue.
+    const { companyId, issueId } = await seedStrandedIssueFixture({
+      status: "in_progress",
+      runStatus: "failed",
+      retryReason: "issue_continuation_needed",
+    });
+
+    // Intercept the DB re-fetch inside ensureStrandedIssueRecoveryIssue.
+    // The re-fetch is uniquely identified by db.select({ id, status }) — exactly 2 keys.
+    // This simulates the race condition where the source issue transitions to
+    // blocked between the candidates scan and recovery creation (memory 10f05083).
+    const originalSelect = (db.select as (...args: unknown[]) => unknown).bind(db);
+    const selectSpy = vi.spyOn(db, "select").mockImplementation((columns?: unknown) => {
+      if (
+        columns !== null &&
+        columns !== undefined &&
+        typeof columns === "object" &&
+        !Array.isArray(columns) &&
+        Object.keys(columns as object).sort().join(",") === "id,status"
+      ) {
+        return {
+          from: () => ({
+            where: () => ({
+              limit: () => Promise.resolve([{ id: issueId, status: "blocked" }]),
+            }),
+          }),
+        } as unknown as ReturnType<typeof db.select>;
+      }
+      return originalSelect(columns) as ReturnType<typeof db.select>;
+    });
+
+    try {
+      const heartbeat = heartbeatService(db);
+      const result = await heartbeat.reconcileStrandedAssignedIssues();
+
+      expect(result.escalated).toBe(1);
+
+      const recoveries = await db
+        .select()
+        .from(issues)
+        .where(and(eq(issues.companyId, companyId), eq(issues.originKind, "stranded_issue_recovery")));
+      expect(recoveries).toHaveLength(0);
+    } finally {
+      selectSpy.mockRestore();
+    }
+  });
+
   it("redacts error-code-only stranded recovery failures in issue copy", async () => {
     const { companyId, agentId, issueId, runId } = await seedStrandedIssueFixture({
       status: "in_progress",
--- a/server/src/__tests__/heartbeat-process-recovery.test.ts
+++ b/server/src/__tests__/heartbeat-process-recovery.test.ts
@@ -1975,7 +1975,12 @@ describeEmbeddedPostgres("heartbeat orphaned process recovery", () => {
       const heartbeat = heartbeatService(db);
       const result = await heartbeat.reconcileStrandedAssignedIssues();
 
-      expect(result.escalated).toBe(1);
+      // bemsas review (blocking #1, 2026-05-01): when fresh DB status is inert,
+      // escalateStrandedAssignedIssue now early-returns BEFORE the issuesSvc.update
+      // call. That means the run is observed and skipped, but `escalated` counter
+      // does not increment (no actual escalation work performed). Previously this
+      // returned 1 because the function ran the no-op blocked→blocked update.
+      expect(result.escalated).toBe(0);
 
       const recoveries = await db
         .select()
--- a/server/src/__tests__/heartbeat-process-recovery.test.ts
+++ b/server/src/__tests__/heartbeat-process-recovery.test.ts
@@ -1928,70 +1928,23 @@ describeEmbeddedPostgres("heartbeat orphaned process recovery", () => {
     expect(comments[0]?.body).toContain(`Recovery issue: [${recovery.identifier}]`);
   });
 
-  it("skips recovery issue creation when DB re-fetch shows source.status=blocked (ELE-32 race guard)", async () => {
-    // Seed same scenario that would normally escalate and create a recovery issue.
-    const { companyId, issueId } = await seedStrandedIssueFixture({
-      status: "in_progress",
-      runStatus: "failed",
-      retryReason: "issue_continuation_needed",
-    });
-
-    // Intercept the DB re-fetch inside ensureStrandedIssueRecoveryIssue.
-    // The re-fetch is uniquely identified by db.select({ id, status }) — exactly 2 keys.
-    // This simulates the race condition where the source issue transitions to
-    // blocked between the candidates scan and recovery creation (memory 10f05083).
-    //
-    // PR #4944 greptile review (P2): this column-shape spy is brittle — any
-    // future query in reconcileStrandedAssignedIssues that selects exactly
-    // { id, status } will be silently hijacked. The narrow keys-equality check
-    // (sort + join + exact match "id,status") is a deliberate guard, but a
-    // safer approach for future tests is the embedded-Postgres fixture used by
-    // sibling tests in this file. Kept here because the alternative requires
-    // standing up a fixture for a single negative-path assertion. If the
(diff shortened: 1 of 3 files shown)


bemsas (changes requested):
Reviewed with focus on recovery cascade safety. Requesting changes before merge.

Blocking findings:

1. Inert fresh status can still be overwritten to `blocked` later in `escalateStrandedAssignedIssue()`.
   The PR suppresses recovery creation when the re-fetched source status is `blocked`, `cancelled`, or `done`, but the escalation path continues and can call `issuesSvc.update(... status: "blocked")`. If the race is `done`/`cancelled`, this risks resurrecting or overwriting an intentionally closed source issue.
   Suggested fix: return an explicit structured skip reason from `ensureStrandedIssueRecoveryIssue()` and abort the escalation update when fresh source status is inert.

2. Recent sibling gate returns `null` before `findOpenStrandedIssueRecoveryIssue()`.
   If a recent `stranded_issue_recovery` exists and is open/visible, the source can still be blocked without linking that existing recovery issue as the blocker. It may also produce misleading operator copy about no manager being found.
   Suggested fix: if the recent sibling is open/visible, return it as the recovery issue; otherwise carry an explicit skip reason.

Non-blocking follow-ups:
- Use `ilike` for the `[FOR OPS]` title gate instead of case-sensitive `like`.
- Exclude hidden `[FOR OPS]` issues with `isNull(issues.hiddenAt)` so archived/hidden ops issues do not suppress recovery.
- Make the operator comment explicit for gate-suppressed recovery instead of implying no invokable manager was found.

CI is green, but these semantics are risky enough to block merge.

PR #4945, opened 2026-05-01
fix(live-runs): default minCount to 0 — no backfill when nothing is queued/running

VIKINGIST:
# `/api/companies/:id/live-runs` returns terminal runs as "live" via backfill — UI mislabel

## Severity

**Low–medium** — cosmetic but pervasive: agent sidebar badges and Dashboard counter show non-zero "X live" indefinitely even when 0 runs are queued or running. Misleads operators into thinking work is in flight.

## Reproduction

1. Pause all agents, or simply wait until all runs reach a terminal state.
2. Verify in DB: `SELECT count(*) FROM heartbeat_runs WHERE status IN ('queued','running')` → `0`.
3. Open the UI. Sidebar shows `Dashboard 50 live`, `Agent-X 13 live`, `Agent-Y 17 live`, etc.

The badges should show 0. They do not, ever, until you delete enough rows from `heartbeat_runs` to drop total below 50.

## Root cause

`server/src/routes/agents.ts` (~line 2843, route `/api/companies/:companyId/live-runs`):

```ts
const minCount = readLiveRunsQueryInt(req.query.minCount, 50, 50);
// ...
const liveRunsQuery = db.select(...).from(heartbeatRuns)
  .where(inArray(heartbeatRuns.status, ['queued', 'running']));
const liveRuns = await liveRunsQuery.limit(limit);

if (targetRunCount > 0 && liveRuns.length < targetRunCount) {
  // FILLS REMAINDER WITH RECENT TERMINAL RUNS UP TO 50
  const recentRuns = await db.select(...)
    .where(not(inArray(heartbeatRuns.status, ['queued', 'running'])))
    .orderBy(desc(heartbeatRuns.createdAt))
    .limit(targetRunCount - liveRuns.length);
  res.json([...liveRuns, ...recentRuns]);
  return;
}
```

Default `minCount=50` means: even when 0 truly-live runs exist, the route always returns 50 rows of "recent activity" labeled in UI as `X live`.

### Option A — server default change (smallest diff)

```ts
const minCount = readLiveRunsQueryInt(req.query.minCount, 0, 50)
```

Default becomes 0 (no backfill). UI callers that genuinely want recent-activity behavior pass `?minCount=50`. Audit callers and update.

### Option B — separate route

Keep `/live-runs` unchanged for activity-feed usage. Add `/live-runs/active-only` returning only `queued|running`. UI sidebar / dashboard badge calls the new route.

### Option C — query param toggle

Add `?onlyActive=true` to suppress backfill. UI sidebar passes it; activity feed default-backfills.

We're shipping option A on a local fork. Happy to send a PR if you'd like.

## DoD on our side

- After fix, all agent badges show 0 when no `queued|running` runs exist.
- Badge ticks to 1 when a single run is dispatched, then back to 0 when it completes.
- No regression on any "Activity" page that genuinely wants the recent-runs feed (passes `?minCount=50`).

## Out of scope

- Renaming `live` → `recent` in UI strings (cosmetic; semantic fix is preferred).
- Heartbeat-run pruning policy. Pruning historical rows doesn't help — backfill always finds 50 recent rows. Only the route semantics matter.

### DIFF
--- a/server/src/__tests__/agent-live-run-routes.test.ts
+++ b/server/src/__tests__/agent-live-run-routes.test.ts
@@ -344,7 +344,20 @@ describe("agent live run routes", () => {
     expect(mockHeartbeatService.buildRunOutputSilence).toHaveBeenCalledTimes(50);
   });
 
-  it("returns empty array by default when no runs are queued or running", async () => {
+  it("does not backfill terminal runs when no active runs exist (default minCount=0)", async () => {
+    // Pre-fix behaviour (default minCount=50):
+    //   1. live query → [] (no queued/running)
+    //   2. backfill block fires because 50 > 0 && 0 < 50
+    //   3. db.select called twice
+    // Post-fix behaviour (default minCount=0):
+    //   1. live query → []
+    //   2. backfill block SKIPPED because 0 > 0 is false
+    //   3. db.select called exactly once
+    //
+    // Asserting on db.select call count is the only stub-friendly way to
+    // distinguish the two code paths: the existing stub doesn't model the
+    // status filter, so we can't rely on row-shape alone to prove the
+    // backfill branch wasn't entered.
     const { db } = createLiveRunsDbStub([]);
 
     const res = await requestApp(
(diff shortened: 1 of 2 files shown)


bemsas (approved):
Reviewed with focus on the live-runs default/no-backfill behavior. Approved — implementation matches the intended behavior and CI is green.

One suggested follow-up: strengthen the regression test. The new empty-DB test would also pass before this PR because both the live query and fallback/recent query return no rows. A better test would seed terminal/recent rows while queued/running is empty, then assert:
- default `/live-runs` returns `[]`
- explicit `?minCount=4` returns recent rows

Not blocking: route logic is simple, default callers now avoid backfill, and the explicit `ActiveAgentsPanel` `minCount` path remains preserved.

caraseli02 (commented):

### Summary
Changes the default `minCount` for live-runs endpoint from 50 to 0, preventing unnecessary backfill of terminal runs when no active runs exist.

### Findings
🟢 **Sensible change.** Backfilling terminal runs when nothing is active wastes DB queries and adds latency.

🟢 **Strong test coverage** — two tests prove both paths: default (no backfill, 1 db.select call) and explicit opt-in (backfill fires, 2 db.select calls). The call-count assertion is a clever way to verify the code path.

🟢 **Backward compatible** — callers who need the old behavior can still pass `minCount=50`.

### Verdict: APPROVE ✅

PR #4967, opened 2026-05-01
fix(routines): redact adapterConfig.env in GET /api/routines/{id}

HDBR:
## Problem

`GET /api/routines/{id}` returns the routine's assignee in full, which includes the agent's `adapterConfig` and `runtimeConfig` objects with `env` values **in plain text**.

In a self-hosted instance running real agents, that response can include things like:

- `GH_TOKEN` (GitHub PAT)
- `VERCEL_TOKEN`
- `CLOUDFLARE_API_TOKEN`
- `PORTAINER_API_TOKEN`
- and any other secret an agent's adapter is configured with

Any user/agent with a valid JWT for the company can read these by calling the route. The response also lands in tool-call buffers when consumed by an LLM agent, which makes it easy to leak the values into commit messages, logs, comments, or chat outputs without anyone noticing.

The `agents` service already runs `sanitizeRecord` on adapterConfig before exposing it (`server/src/services/agents.ts`), but two things prevented that from helping here:

1. The routines `getDetail` returns the raw `agents` row directly — no sanitization on the route boundary.
2. `SECRET_PAYLOAD_KEY_RE` in `redaction.ts` only matched literal substrings (`api_key`, `access_token`, `auth_token`, `password`, ...). Common conventions like `GH_TOKEN`, `VERCEL_TOKEN`, `CLOUDFLARE_API_TOKEN` use the suffix pattern `*_TOKEN` and were never matched.

## Reproduction

On any self-hosted instance with a routine whose assignee has secrets:

```bash
curl -s -H "Authorization: Bearer $YOUR_JWT" \
  "$BASE_URL/api/routines/$ROUTINE_ID" \
  | jq '.assignee.adapterConfig.env'
```

Returns plain text values for everything in the env block.

## Fix

Two changes, both small:

1. **`server/src/redaction.ts`** — extend `SECRET_PAYLOAD_KEY_RE` to also match the trailing patterns `[-_]token$`, `[-_]key$`, `[-_]secret$`, `[-_]password$`. This catches `GH_TOKEN`, `VERCEL_TOKEN`, `CLOUDFLARE_API_TOKEN`, `OPENAI_API_KEY`, etc., wherever `sanitizeRecord` is applied.

2. **`server/src/routes/routines.ts`** — add a `redactRoutineDetailForClient` helper and apply it to `detail` right before `res.json` in the `GET /routines/:id` handler. The helper walks `assignee.adapterConfig` and `assignee.runtimeConfig` and runs them through `sanitizeRecord`. Internal callers of `routineService.getDetail` are unaffected — they still receive the raw values, which the dispatcher needs to actually invoke the adapter.

## What this does NOT cover (intentional, separate concern)

- The same shape of response on `GET /api/agents/{id}` may also leak — `services/agents.ts` does sanitize, but with the pre-fix regex some keys were missed. The regex extension in this PR helps that path too, but I haven't audited every callsite. Worth a follow-up.
- The fix relies on key-name matching. Values stored under non-secret-shaped keys (e.g. `note: "ghp_xxx"`) will still go through unredacted. The right long-term answer is probably a typed `EnvBinding` (`{ type: "secret_ref" | "plain", masked: ... }`), but that's a bigger change.

## Tests

Existing `redaction.test.ts` cases still pass (suffix patterns extend, don't replace, the original alternation). I'd add a focused case for the routine detail handler in a follow-up — wanted to keep this PR minimal so it's easy to review.

## Why I'm sending this

I run a self-hosted Paperclip instance for HDBR (`system.hdbr.studio`) with four agents, and one of them noticed it could read another's `GH_TOKEN` via this endpoint during normal work. Filing the fix here so other self-hosters don't run into the same leak — happy to iterate on whatever shape the maintainers prefer.

### DIFF
--- a/server/src/redaction.ts
+++ b/server/src/redaction.ts
@@ -1,7 +1,7 @@
 import { redactCommandText } from "@paperclipai/adapter-utils";
 
 const SECRET_PAYLOAD_KEY_RE =
-  /(api[-_]?key|access[-_]?token|auth(?:_?token)?|authorization|bearer|secret|passwd|password|credential|jwt|private[-_]?key|cookie|connectionstring)/i;
+  /(api[-_]?key|access[-_]?token|auth(?:_?token)?|authorization|bearer|secret|passwd|password|credential|jwt|private[-_]?key|cookie|connectionstring|[-_]token$|[-_]key$|[-_]secret$|[-_]password$)/i;
 const COMMAND_PAYLOAD_KEY_RE =
   /(^command$|^cmd$|command[-_]?line|resolved[-_]?command|PAPERCLIP_RESOLVED_COMMAND)/i;
 const COMMAND_ARGS_PAYLOAD_KEY_RE = /^(commandArgs|command_?args|argv)$/i;
--- a/server/src/routes/routines.ts
+++ b/server/src/routes/routines.ts
@@ -22,15 +22,16 @@ function redactRoutineDetailForClient(detail: unknown): unknown {
   const assignee = (detail as { assignee?: unknown }).assignee;
   if (!assignee || typeof assignee !== "object" || Array.isArray(assignee)) return detail;
   const a = assignee as Record<string, unknown>;
+  const redactedAssignee: Record<string, unknown> = { ...a };
   const adapterConfig = a.adapterConfig;
   if (adapterConfig && typeof adapterConfig === "object" && !Array.isArray(adapterConfig)) {
-    a.adapterConfig = sanitizeRecord(adapterConfig as Record<string, unknown>);
+    redactedAssignee.adapterConfig = sanitizeRecord(adapterConfig as Record<string, unknown>);
   }
   const runtimeConfig = a.runtimeConfig;
   if (runtimeConfig && typeof runtimeConfig === "object" && !Array.isArray(runtimeConfig)) {
-    a.runtimeConfig = sanitizeRecord(runtimeConfig as Record<string, unknown>);
+    redactedAssignee.runtimeConfig = sanitizeRecord(runtimeConfig as Record<string, unknown>);
   }
-  return detail;
+  return { ...(detail as Record<string, unknown>), assignee: redactedAssignee };
 }
 
 export function routineRoutes(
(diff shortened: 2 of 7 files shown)


HDBR:
Thanks for the thorough review. Walking through the three points:

**1. In-place mutation of `detail`** — fixed in `962d441`. The helper now spreads `assignee` into a fresh record, applies `sanitizeRecord` to the copy, and returns a new detail object. Internal callers of `routineService.getDetail` (e.g. the dispatcher) keep receiving the raw values, which is what they need to invoke the adapter. This also future-proofs against any caching layer landing on top of `getDetail`.

**2. List endpoint not covered** — I dug into this and don't think `GET /companies/:companyId/routines` leaks. `svc.list` returns `RoutineListItem[]`, where `RoutineListItem extends Routine` and only carries `assigneeAgentId` as an FK (see `packages/shared/src/types/routine.ts:132` and the mapping in `services/routines.ts:1086-1101` — it spreads the raw routines row plus `triggers`, `lastRun`, `activeIssue`, none of which include the agent record). The full agent join, including `adapterConfig` / `runtimeConfig`, only happens in `getDetail`. Happy to add a defensive redact pass on `list` anyway if maintainers prefer belt-and-suspenders, but it would be a no-op against the current shape.

**3. `[-_]key$` over-redaction on broader payloads** — this one is intentional and I think the right trade-off, but worth calling out explicitly so a maintainer can override:
- The existing `api[-_]?key` alternation already covers `OPENAI_API_KEY`, `STRIPE_API_KEY`, etc. The new suffix is meant to catch the other common conventions that don't include `api` in the name: `ENCRYPTION_KEY`, `SIGNING_KEY`, `SESSION_KEY`, `DEPLOY_KEY`, `WEBHOOK_KEY`, `JWT_SIGNING_KEY`, etc.
- The false positives you correctly flagged (`primary_key`, `sort_key`, `partition_key`, `foreign_key`) are DB schema metadata. They show up in things like activity-log details and event payloads via `redactEventPayload`, but they aren't *values* an operator needs to read — at worst a log entry shows `primary_key: ***REDACTED***` instead of `primary_key: "id"`, which is cosmetic.
- The asymmetric cost made me prefer the broader pattern: missing a `SESSION_KEY` is a real leak; redacting `primary_key` in a log line is noise. If a tighter scope is preferred, narrowing to `(signing|encryption|session|deploy|webhook|private)[-_]key$` keeps almost all the value with less noise — happy to push that variant if you'd rather.

Tests: existing `redaction.test.ts` (7 cases) and `routines-routes.test.ts` (8 cases) still pass; typecheck clean. I deliberately kept this PR minimal so it's easy to review, but I can add a focused test for the `/routines/:id` redaction in a follow-up if useful.

brandonburr:

**Rationale:** Security: prevents env secret leakage from routine API responses.

Issue #4946, opened 2026-05-01
workspaceStrategy=git_worktree silently dropped at agent spawn — runs in _default/ instead


VIKINGIST:
# `workspaceStrategy=git_worktree` silently dropped at agent spawn — runs in `_default/` instead

## Severity

**High** — agents hired with `git_worktree` strategy run in the wrong cwd, silently miss `.claude/settings.json`, `CLAUDE.md`, `.mcp.json`, hooks, and skills, and may close issues without satisfying acceptance criteria.

## Reproduction

1. Hire an agent with:

   ```json
   {
     "adapterConfig": {
       "cwd": "<absolute path to a real EB-style git repo>",
       "workspaceStrategy": { "type": "git_worktree", "baseRef": "master" }
     }
   }
   ```

2. Assign it any issue.
3. Inspect runtime: `sessionParamsJson.cwd` resolves to `~/.paperclip/instances/default/projects/<paperclip-project>/_default` instead of a worktree of the configured repo.

Empirical observation: the run completes (12.5k output tokens), reports `done`, but no file edits are performed (the `_default/` dir is empty) and the issue is auto-closed despite no diff being posted.

## Hypotheses (untested)

- **A. PAPERCLIP_WORKSPACE_STRATEGY env override.** `PAPERCLIP_WORKSPACE_STRATEGY=project_primary` is set in some agent envs (we observed this in the CEO's adapterConfig.env). Agent-level `workspaceStrategy` may be silently overridden by the env at resolver time in `workspace-runtime.ts`.
- **B. Schema drop on direct POST.** UI form-builder `build-config.ts` translates UI keys (`workspaceStrategyType`) into `adapterConfig.workspaceStrategy`. A direct JSON POST to `/api/companies/:companyId/agents` with the runtime-shape may not match the resolver's expected key path.
- **C. cwd-override env vars.** `PAPERCLIP_WORKSPACE_CWD` (we observed `_default/`) may take precedence over `agent.adapterConfig.cwd` somewhere in the resolver chain.

## Impact

- Implementer agents that should run in a per-issue worktree of a target repo end up in an empty default workspace, with no project context.
- `.claude/settings.json` hooks never fire → no skill auto-invocation, no rule-files surface.
- Agents close issues without doing the work (we treat this as a separate "behavior bug" — needs an explicit "if criteria not met, mark `blocked` and do not close" rule on hire spec / instructions bundle).
[...]

PR #4968, opened 2026-05-01
fix(workspace): preserve agent.workspaceStrategy + adapterConfig.cwd through resolver layers (#4946)

VIKINGIST:
## Thinking Path

> - Paperclip orchestrates AI agents, and the heartbeat scheduler decides where each run physically executes (workspace cwd) before spawning the adapter.
> - Workspace resolution flows through two layers: `execution-workspace-policy.ts` decides which strategy applies (project policy vs agent's own), then `heartbeat.ts:resolveWorkspaceForRun` resolves the actual cwd from the chain (project → session → agent home).
> - On a deployment with no project workspace policy, no project bound to the issue, and an agent that declared `adapterConfig.workspaceStrategy = { type: "git_worktree", baseRef: "..." }` plus `adapterConfig.cwd = "<real repo>"`, **both layers silently dropped the agent's intent**: policy deleted `workspaceStrategy` whenever `legacyUseProjectWorkspace === false` triggered (regardless of actual policy state), and runtime fell straight from "no session cwd" to `~/.paperclip/.../agents/<id>/_default/` instead of honouring `adapterConfig.cwd`.
> - The downstream worktree provisioner then ran `git worktree add` from `_default/` (no `.git`) → `fatal: not a git repository` in <100 ms → recovery cascade fired (separate fix in #4944) → re-arm loop unless gated.
> - This is the underlying cause of the symptoms #4944 was put in place to suppress: #4944 stops the cascade re-arm loop; this PR stops the upstream resolver from producing the failure that fed it.
> - This pull request preserves the agent's `workspaceStrategy` at the policy layer when no project/issue is asserting its own, and adds a fallback at the runtime layer that uses `agent.adapterConfig.cwd` (when it points at a real directory) before degrading to the per-agent home dir.
> - The benefit: agents hired with an explicit cwd + worktree strategy actually run there, hooks/skills/AGENTS.md inside that repo are loaded, and the recovery cascade gate from #4944 stays cold instead of catching legitimate runs that should never have failed.

## What Changed

- **Policy layer** (`execution-workspace-policy.ts`): `buildExecutionWorkspaceAdapterConfig` previously deleted `nextConfig.workspaceStrategy` whenever `hasWorkspaceControl` was true and the resolved mode was not `isolated_workspace`. That branch fired when the project policy was enabled, **or** the issue had a workspaceStrategy override, **or** simply when `legacyUseProjectWorkspace === false`. The third trigger silently dropped agent intent in deployments with no policy at all. Fix: only delete the agent's strategy when the project/issue is actively asserting its own (explicit `workspaceStrategy` override, or an enabled policy in a managed mode), and preserve the agent's strategy in `agent_default` mode and the legacy-flag-only case.
- **Reviewer feedback fix** (same file): a follow-up edge case — `mode in {"shared_workspace","operator_branch"}` with `projectPolicy=null` and no override — was preserving the agent's strategy when it should have been cleared (managed modes assume workspace control). Refactored gate from "policy enabled" to "mode is managed".
- **Runtime layer** (`agent-workspace-fallback.ts`, new): pure helper `resolveAgentConfigCwdFallback({ adapterConfig, dirExists })` returning `{ cwd } | null`. Pure shape lets the entire fallback policy be unit-tested without embedded-Postgres or live FS.
- **Runtime layer** (`heartbeat.ts:resolveWorkspaceForRun`): wired the helper between the session-cwd block and the agent-home fallback. New source kind `"agent_config"` distinguishes this branch in telemetry. Added structured warnings when the configured cwd exists but a project workspace was expected, mirroring the existing `agent_home` warning path.
- **Type extension**: `"agent_config"` added to `ResolvedWorkspaceForRun.source`, `ExecutionWorkspaceInput.source`, and `WorkspaceRealizationRequest.source.kind` (in `packages/shared/src/types/workspace-runtime.ts`). All three kept consistent.
- **Read-side allowlist** (`workspace-realization.ts`): `readWorkspaceRealizationRequest`'s kind allowlist was `["task_session","agent_home"]`, so persisted requests with `kind="agent_config"` round-tripped to `"project_primary"` on driver/server restart. Added `"agent_config"` so persisted state survives a restart with the truthful source. Function exported for direct unit testing.
- **Session migration guard** (`heartbeat.ts:migrateFallbackWorkspaceSession`): the guard `source !== "project_primary"` skipped migration for `agent_config`, leaving session params frozen on the previous `_default/` cwd even after the agent's real cwd became reachable. Extended to allow migration when the new resolution is `agent_config` — symmetric semantics with `project_primary` (both are real workspaces escaping the agent-home fallback).
- **Warning condition alignment** (`heartbeat.ts`): the new `agent_config` branch's "no project workspace" warning condition now matches the existing `agent_home` branch (`else if (resolvedProjectId)` rather than the asymmetric `useProjectWorkspace && resolvedProjectId`).

## Verification

**Unit tests (vitest, repo standard).**
- `agent-workspace-fallback.test.ts` (new, 7 cases): configured-cwd-on-disk path, configured-cwd-missing, unset/empty/non-string cwd, non-object adapterConfig (defensive), `dirExists` rejection propagation.
- `execution-workspace-policy.test.ts` (existing, +4 cases): `agent_default` with no policy, `legacyUseProjectWorkspace=false` with no policy/override, explicit `issue.workspaceStrategy` override winning over agent's, managed modes (`shared_workspace`/`operator_branch`) clearing the strategy regardless of project-policy state. Pre-existing 9 cases still green. **14/14 pass.**
- `workspace-realization-read.test.ts` (new, 7 cases): each kind preserved (`task_session`, `agent_home`, `agent_config`), unknown-kind fall-through to `project_primary`, missing-kind defaulting, `version != 1` rejection, missing-required-field rejection.
- `heartbeat-workspace-session.test.ts` (existing 37 + 1 new): the new case "migrates fallback workspace sessions to agent_config workspace when agent.adapterConfig.cwd becomes available" is symmetric with the existing `project_primary` migration test and locks in the `migrateFallbackWorkspaceSession` guard widening — preempting a regression in either direction. **38/38 pass.**

```
pnpm --filter @paperclipai/server vitest run \
  src/__tests__/agent-workspace-fallback.test.ts \
  src/__tests__/execution-workspace-policy.test.ts \
  src/__tests__/workspace-realization-read.test.ts \
  src/__tests__/heartbeat-workspace-session.test.ts
# 4 files | 66 passed | 0 failed
```

**Repo-wide:** `pnpm test:run` clean (no regressions outside the changed files).

**TypeScript:** `pnpm --filter @paperclipai/server exec tsc --noEmit` — clean across server + packages/shared. Type-changes consistent across 3 union aliases.

**End-to-end on a live Windows deployment.**
1. Hired Implementer-1 agent with `cwd=<EB repo>` and `workspaceStrategy={type:"git_worktree",baseRef:"master"}`.
2. Created issue with `projectId=null` (no bound project) and assigned it to the agent.
3. Pre-fix: heartbeat resolved cwd to `~/.paperclip/.../agents/<id>/_default/`, worktree provisioning failed instantly with `fatal: not a git repository`, recovery cascade fired.
4. Post-fix: cwd resolved via the new `agent_config` branch to the real EB repo path, worktree provisioner created `paperclip/ELE-39-...` worktree on the EB repo, agent ran git commands successfully, edited 5 files in the cable editor, committed `5bd0ed7`, posted a structured diff comment, and handed the issue off to Reviewer-Architecture in `in_review`. Verified by inspecting `git log` in the worktree and the agent's run output.

(A separate Windows-only `spawn sh ENOENT` issue surfaced afterwards in the worktree provisioner's shell-resolution path — that is fixed in a separate commit on the same deployment, **out of scope for this PR**, mentioned here for context. See "Risks" below.)

## Risks

- **Migration:** runs persisted before this PR landed wrote `kind="agent_config"` only via servers that already shipped the runtime-layer change *and* hit the read path *and* survived a restart — i.e., the silent coercion was a forward-only loss. Existing live data has either older kinds (unaffected) or records that get rebuilt on the next heartbeat run via `buildWorkspaceRealizationRequest`. **No backfill needed.** Worst case: a stale record reads as `project_primary` on first restart after this PR ships, then gets overwritten on the next run — same as today.
- **Behavioural shift on `agent_default` mode:** previously the agent's `workspaceStrategy` was always deleted in `agent_default`; now it survives unless project/issue assert otherwise. Agents currently relying on the silent deletion (e.g., expecting `_default/` cwd despite declaring a strategy) will now actually use their declared strategy. Surveyed our deployment: this is the intended behaviour. If an upstream operator wanted "ignore agent strategy in agent_default", they can set the project's policy or issue override explicitly.
- **Behavioural shift on `legacyUseProjectWorkspace=false`:** same direction — agent's strategy is now preserved when the legacy flag was the sole trigger. Aligned with what the field name implies (legacy compatibility, not policy assertion).
- **Session migration extended to `agent_config`:** previously a session that fell back to `_default/` would only migrate to `project_primary`. Now it also migrates to `agent_config` — preventing session params from freezing on `_default/` after the agent's real cwd becomes reachable. Symmetric with existing semantics.
- **`dirExists` swallows EACCES silently** (via `.catch(() => false)`) — agents whose `cwd` exists but is unreadable due to permissions will fall through to `agent_home` without a diagnostic. Edge case, mirrors existing `fs.stat` patterns elsewhere in heartbeat. Improvement opportunity, not a regression.
- **Out-of-scope, separate fix on the same deployment:** the worktree provisioner's `spawn(shell, ["-c", cmd], …)` pattern fails with `spawn sh ENOENT` on Windows because Node's `resolveShell()` returns the relative `"sh"` fallback. Fixed by switching to `spawn(cmd, [], { shell: true })` in a separate commit/PR. Affects only Windows deployments using `git_worktree` provisionCommand. Independent of this PR's resolver fixes.
- **Out-of-scope, pre-existing:** `workspace-runtime.ts:2040`'s long-running `spawn(shell, ["-lc", command], …)` for runtime-service spawn has the same Windows shell pattern. Not addressed here; same class as the provisionCommand fix above.

**Low risk overall.** Pure-helper extraction makes the policy mechanically reviewable; the wiring is one if-block plus one allowlist entry plus one guard widening; the policy-layer behaviour-matrix change is covered by 4 new tests on every cell of the matrix.

## Model Used

- Provider: Anthropic Claude (via Claude Code CLI)
- Model: `claude-opus-4-7` (Claude Opus 4.7)
- Context window: 1M tokens
- Mode: extended-thinking off, tool-use enabled (Bash, Edit, Read, Grep, MCP `memory_search`, MCP `electroboard-rag`)
- Capability: agentic edits across multi-file Paperclip server code, repo-aware planning, integrated test authoring + tsc verification
- The 4-commit policy/runtime split was authored across two Claude sessions; the 5th polish commit (D1 + C1 + C2 read-allowlist / session-migration / warning alignment) consolidated review-found gaps. Pure-helper extraction was suggested by the model after surveying call sites; type-extension consistency across the 3 source-kind aliases was verified by tsc.

## Checklist

- [x] I have run tests locally and they pass (`pnpm test:run` clean; targeted 65/65; tsc clean)
- [x] I have added or updated tests where applicable (18 new test cases across 3 files)
- [ ] If this change affects the UI, I have included before/after screenshots — N/A (server-only resolver change)
- [x] I have updated relevant documentation to reflect my changes (this PR description; type unions extended in `packages/shared`)

## Related

- Fixes upstream issue #4946 (workspaceStrategy silently dropped + cwd fallthrough to `_default/`).
- **Complementary to #4944 (recovery cascade gate) — neither blocks the other; they fix different halves of the same incident.** Reviewing both gives the full picture, but either can merge first.
  - **This PR (upstream / root cause):** removes the silent `workspaceStrategy` drop + the `_default/` fallthrough that produced sub-100 ms `fatal: not a git repository` failures during the session-244 incident. Self-sufficient: even with #4944 not yet landed, this PR alone makes the original failure stop occurring; the cascade machinery simply stays cold.
  - **#4944 (downstream / safety net):** adds status + sibling-window + ops-active gates to `ensureStrandedIssueRecoveryIssue`, suppressing the infinite re-arm loop that the resolver fault originally fed. Self-sufficient: even with this PR not yet landed, #4944 alone keeps the loop from spamming the board; the underlying failure still occurs but is contained at one recovery issue per parent rather than an infinite chain.
  - **Together** they close the loop on both ends: the failure mode is removed at the resolver, and the safety net behind it is re-armed for any unrelated future failure that might surface the same symptom.
- The deployment that produced upstream issue #4946 had both fixes applied locally during reproduction; #4944 (locally) caught the original cascade and this PR (locally) removed the cause that fed it. The end-to-end verification described above (Implementer-1 / ELE-39) was performed in that combined state, but the policy + runtime + read-allowlist + session-migration tests in this PR exercise the resolver paths in isolation and pass without #4944's recovery code being touched.

### DIFF
--- a/packages/shared/src/types/workspace-runtime.ts
+++ b/packages/shared/src/types/workspace-runtime.ts
@@ -249,7 +249,7 @@ export interface WorkspaceRealizationRequest {
   heartbeatRunId: string;
   requestedMode: string | null;
   source: {
-    kind: "project_primary" | "task_session" | "agent_home";
+    kind: "project_primary" | "task_session" | "agent_config" | "agent_home";
     localPath: string;
     projectId: string | null;
     projectWorkspaceId: string | null;
--- a/server/src/__tests__/agent-workspace-fallback.test.ts
+++ b/server/src/__tests__/agent-workspace-fallback.test.ts
@@ -1,0 +1,94 @@
+import { describe, expect, it, vi } from "vitest";
+import { resolveAgentConfigCwdFallback } from "../services/agent-workspace-fallback.js";
+
+describe("resolveAgentConfigCwdFallback", () => {
+  it("returns the configured cwd when it exists on disk", async () => {
+    const dirExists = vi.fn().mockResolvedValue(true);
+    const result = await resolveAgentConfigCwdFallback({
+      adapterConfig: {
+        cwd: "C:/Users/VIKINGIST/Documents/www/ElectroBoard/electroboard",
+        workspaceStrategy: { type: "git_worktree", baseRef: "master" },
+      },
+      dirExists,
+    });
+    expect(result).toEqual({
+      cwd: "C:/Users/VIKINGIST/Documents/www/ElectroBoard/electroboard",
+    });
+    expect(dirExists).toHaveBeenCalledWith(
+      "C:/Users/VIKINGIST/Documents/www/ElectroBoard/electroboard",
+    );
+  });
+
+  it("returns null when cwd is unset", async () => {
+    const dirExists = vi.fn();
+    expect(
+      await resolveAgentConfigCwdFallback({
+        adapterConfig: { workspaceStrategy: { type: "project_primary" } },
+        dirExists,
+      }),
+    ).toBeNull();
+    expect(dirExists).not.toHaveBeenCalled();
+  });
+
+  it("returns null when cwd is empty string", async () => {
+    const dirExists = vi.fn();
+    expect(
+      await resolveAgentConfigCwdFallback({
+        adapterConfig: { cwd: "" },
+        dirExists,
+      }),
+    ).toBeNull();
+    expect(dirExists).not.toHaveBeenCalled();
+  });
+
+  it("returns null when cwd is not a string", async () => {
+    const dirExists = vi.fn();
+    expect(
+      await resolveAgentConfigCwdFallback({
+        adapterConfig: { cwd: 42 },
+        dirExists,
+      }),
+    ).toBeNull();
+    expect(
+      await resolveAgentConfigCwdFallback({
+        adapterConfig: { cwd: { nested: "x" } },
+        dirExists,
+      }),
+    ).toBeNull();
+    expect(dirExists).not.toHaveBeenCalled();
+  });
+
+  it("returns null when configured cwd does not exist on disk", async () => {
+    const dirExists = vi.fn().mockResolvedValue(false);
+    const result = await resolveAgentConfigCwdFallback({
+      adapterConfig: { cwd: "C:/nonexistent/path" },
+      dirExists,
+    });
+    expect(result).toBeNull();
+    expect(dirExists).toHaveBeenCalledWith("C:/nonexistent/path");
+  });
+
+  it("returns null when adapterConfig is not an object (defensive)", async () => {
+    const dirExists = vi.fn();
+    expect(
+      await resolveAgentConfigCwdFallback({ adapterConfig: null, dirExists }),
+    ).toBeNull();
+    expect(
+      await resolveAgentConfigCwdFallback({ adapterConfig: undefined, dirExists }),
+    ).toBeNull();
+    expect(
+      await resolveAgentConfigCwdFallback({ adapterConfig: "not-an-object", dirExists }),
+    ).toBeNull();
+    expect(dirExists).not.toHaveBeenCalled();
+  });
+
+  it("propagates dirExists rejection (caller decides retry policy)", async () => {
+    const dirExists = vi.fn().mockRejectedValue(new Error("EACCES"));
+    await expect(
+      resolveAgentConfigCwdFallback({
+        adapterConfig: { cwd: "C:/restricted" },
+        dirExists,
+      }),
+    ).rejects.toThrow("EACCES");
+  });
+});
--- a/server/src/__tests__/execution-workspace-policy.test.ts
+++ b/server/src/__tests__/execution-workspace-policy.test.ts
@@ -86,12 +86,14 @@ describe("execution workspace policy helpers", () => {
     });
   });
 
-  it("clears managed workspace strategy when issue opts out to project primary or agent default", () => {
+  it("clears agent workspace strategy when project policy enforces a managed shared workspace", () => {
     const baseConfig = {
       workspaceStrategy: { type: "git_worktree", branchTemplate: "{{issue.identifier}}" },
       workspaceRuntime: { services: [{ name: "web" }] },
     };
 
+    // Project policy is enabled (managed mode) and the issue opts into shared_workspace —
+    // the project's primary workspace is in force, so the agent's strategy yields.
     expect(
       buildExecutionWorkspaceAdapterConfig({
         agentConfig: baseConfig,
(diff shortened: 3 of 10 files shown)


VIKINGIST:
### Cross-reference with #4970

PR #4970 (filed ~3 minutes after this one, by @brunocardoso-netizen) targets the same `fatal: not a git repository` symptom but at a different layer. Posting this so reviewers see the relationship up front and can decide how to sequence the two.

**They are complementary, not overlapping:**

- **This PR (#4968)** fixes the resolver chain so `source=agent_home` is **never reached in the first place** when the agent has a valid `adapterConfig.cwd`. The agent lands in its real configured repo and the worktree is provisioned there — the correct outcome.
- **#4970** adds a downstream guard in `realizeExecutionWorkspace` that prevents the crash whenever `source=agent_home` *is* reached (e.g. cron heartbeats with no issue context, or agents without a configured `adapterConfig.cwd`). The agent doesn't crash, but it also doesn't end up in a real repo — it lands in `_default/` with the worktree request demoted.

**Coverage matrix:**

| Scenario | #4968 alone | #4970 alone | Both |
|---|---|---|---|
| Agent with `adapterConfig.cwd`, issue-driven | ✅ correct cwd | ⚠️ no crash, wrong cwd | ✅ correct cwd |
| Agent with `adapterConfig.cwd`, cron heartbeat | ✅ correct cwd | ⚠️ no crash, wrong cwd | ✅ correct cwd |
| Agent without `adapterConfig.cwd`, cron / no project | ❌ still crashes | ✅ no crash, `_default/` | ✅ no crash, `_default/` |

For the deployment that produced upstream issue #4946 — and for every Paperclip fleet I've seen where agents are explicitly hired with a target repo as `adapterConfig.cwd` — this PR is the principal fix; #4970 then provides defense-in-depth for misconfiguration paths and for cron heartbeats firing without issue context.

We support both landing. If forced to pick one, this PR addresses the root cause for the explicit-cwd case (the realistic deployment shape); #4970 covers a broader class of symptoms but leaves agents running in the wrong cwd. Happy to defer in either direction based on maintainer preference.

PR #4986, opened 2026-05-02
fix(recovery): skip stranded detection for routine_execution issues with active source routine

brunocardoso-netizen:
## Thinking Path

> - Paperclip orchestrates AI agents and runs recurring work via routines
> - The recovery subsystem (`reconcileStrandedAssignedIssues`) periodically scans for agent-assigned issues in `todo` or `in_progress` with no live execution path, and re-dispatches or escalates them
> - Short-lived routine-execution agents exit after each cycle by design; their heartbeat run ends with `errorCode: process_lost` (status `failed`) rather than `succeeded`, because the process exits without a graceful completion signal
> - Between routine firings the issue sits `in_progress` with no active execution path — precisely the pattern the stranded detector targets
> - The detector has no concept of "this issue is driven by a recurring routine whose next fire is coming" — it treats the gap between cycles as a crash and enqueues a continuation-recovery retry
> - On the second failed retry `didAutomaticRecoveryFail` fires and escalates the issue to `blocked`, creating a noisy recovery child issue and waking unrelated agents to triage it
> - This PR adds an early-exit guard in the `in_progress` branch: if the issue has `originKind === 'routine_execution'` and its source routine is still `active`, skip stranded detection entirely — the routine will deliver the next wake
> - The benefit is that BRA-325-style recurring poll workers can run indefinitely without false-positive blocking, eliminating ~6 spurious recovery wakeups/hour

## What Changed

- `server/src/services/recovery/service.ts`: added `routines` to the `@paperclipai/db` import
- Added `hasActiveSourceRoutine(companyId, originId)` helper — single indexed query on `routines` by `(companyId, id, status)`
- Added guard at the top of the `in_progress` detection block in `reconcileStrandedAssignedIssues()`: if `originKind === 'routine_execution'` and the source routine is `active`, increment `skipped` and `continue`

## Verification

- Routine-execution issue stays `in_progress` between cycles without being pushed to `blocked`
- If the source routine is paused/cancelled the guard does not fire and the stranded detector operates normally on the orphaned issue
- `pnpm --filter @paperclipai/server typecheck` passes with no new errors

## Risks

- Low risk. The guard only fires when two conditions are both true: `originKind === 'routine_execution'` AND the routine's `status = 'active'`. Issues whose routines are paused, cancelled, or deleted fall through to the existing stranded logic unchanged.
- One extra indexed DB query per `routine_execution` issue per 30s heartbeat cycle. The query hits `routines_company_status_idx` and `routines_company_assignee_idx` so it is fast.
- A genuinely crashed routine-execution issue (routine still `active` but the issue needs real intervention) will not receive auto-recovery. Operators can still manually unblock it; the trade-off is correct given that false positives were firing every 10 minutes.

## Model Used

- Provider: Anthropic  
- Extended tool-use reasoning; no special thinking mode

## Checklist

### DIFF
--- a/server/src/services/recovery/service.ts
+++ b/server/src/services/recovery/service.ts
@@ -19,6 +19,7 @@ import {
   issueRelations,
   issueThreadInteractions,
   issues,
+  routines,
 } from "@paperclipai/db";
 import { parseObject, asBoolean, asNumber } from "../../adapters/utils.js";
 import { runningProcesses } from "../../adapters/index.js";
@@ -362,6 +363,22 @@ export function recoveryService(db: Db, deps: { enqueueWakeup: RecoveryWakeup })
     ]);
 
     return Boolean(run || deferredWake);
+  }
+
+  async function hasActiveSourceRoutine(companyId: string, originId: string) {
+    const row = await db
+      .select({ id: routines.id })
+      .from(routines)
+      .where(
+        and(
+          eq(routines.companyId, companyId),
+          eq(routines.id, originId),
+          eq(routines.status, "active"),
+        ),
+      )
+      .limit(1)
+      .then((rows) => rows[0] ?? null);
+    return Boolean(row);
   }
 
   async function hasQueuedIssueWake(companyId: string, issueId: string) {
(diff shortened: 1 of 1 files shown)


caraseli02 (commented):

### Summary
Skips stranded-issue detection for `routine_execution` issues whose source routine is still active. Prevents recovery from re-firing on issues that are expected to cycle through agent-exit → routine-refire.

### Findings
🟢 **Targeted fix.** The `hasActiveSourceRoutine` query is efficient (indexed lookup, LIMIT 1).

🟢 **Good placement** — the early-continue happens before any heavier recovery checks, saving unnecessary work.

🟡 **Question:** Should there be a log line when skipping for observability? Something like `debug("skipping stranded check: routine %s still active", originId)` would help operators debug unexpected skip behavior in production.

### Verdict: APPROVE ✅

PR #5033, opened 2026-05-02
feat(plugins): add ctx.runs.onBeforeRun veto hook and run.gate capability

supertaz:
## Thinking Path

> - Paperclip orchestrates AI agents for zero-human companies; plugins extend Paperclip by running in isolated workers with sandboxed access to the host via a capability-gated RPC SDK
> - Plugins that need to gate or stage work *before* a queued run flips to `running` have no synchronous hook today — `PluginEventBus.emit()` is fire-and-forget (handler return values are discarded), and all `agent.run.*` events fire *after* the row is already claimed
> - The only existing gate is the blocker idiom (`ctx.issues.relations.setBlockedBy`), which works for dependency chains but not for "this run cannot start until the plugin says so"
> - Workstream-pool gating, container per-step provisioning, egress policy enforcement, and similar use cases all need a synchronous, in-process veto before the row flips to `running`
> - This PR adds `ctx.runs.onBeforeRun(handler)` to the plugin SDK, gated by a new `run.gate` capability, dispatched inside `claimQueuedRun()` *before* the status update — so a veto truly prevents the run from ever entering `running` state
> - The benefit is a first-class veto primitive that cleanly composes with existing dependency blockers and event handlers, with deterministic ordering, capability gating, and fail-open semantics so a misbehaving plugin cannot deadlock the run queue

This hook is complementary to (not competing with) the `onBeforeAdapterExecute` proposal in discussion #4456, which fires later in the dispatcher (after the run is already claimed) and is scoped to adapter env/config injection. `onBeforeRun` runs at the claim site, so a veto cancels the run before it ever transitions to `running`.

## Model Used

- Implementation: Claude Sonnet 4.6
- Adversarial review (design): GPT-5.5 codex (high reasoning) + Claude Sonnet 4.6
- Pre-submission self-review: GPT-5.4-mini codex (xhigh reasoning)

## What Changed

**SDK surface (packages/plugins/sdk)**
- New `PluginRunsClient` on `PluginContext` with `onBeforeRun(handler)` registration method
- New types: `BeforeRunParams` (`runId`, `agentId`, `issueId | null`, `companyId`, `invocationSource`) and `BeforeRunResult` (`{ veto: false }` or `{ veto: true; reason: string }`)
- Added `runs.registerBeforeRunHandler` (worker → host) and `beforeRun` (host → worker) to the RPC protocol
- `worker-rpc-host.ts` validates the `run.gate` capability when a worker registers a handler
- `host-client-factory.ts` exposes the `runs` client and wires the local handler into the dispatch pipeline
- Test harness stub added in `testing.ts` so plugin authors can unit-test gate handlers without a host
- New types exported from the SDK index

**Capability**
- New `run.gate` entry in `PLUGIN_CAPABILITIES` (`packages/shared/src/constants.ts`)
- Only plugins declaring `run.gate` may register a `beforeRun` handler — host enforces on the registration RPC and again at dispatch

**Server (server/src)**
- `PluginWorkerManager` gains `registerBeforeRunGate`, `unregisterBeforeRunGate`, and `broadcastBeforeRun`
- New pure helper `resolveRunGatePlugins` filters plugins by `run.gate` capability + not-disabled-for-company and sorts by `installOrder` for deterministic dispatch
- `broadcastBeforeRun` fans out in parallel with a 5s per-gate timeout, fail-open on error/timeout/malformed-response, stops aggregating at the first veto, and tracks the vetoing plugin id for audit attribution
- `claimQueuedRun()` in `heartbeat.ts` calls the gate after the existing dependency/staleness checks and BEFORE the `status = running` update
- New `cancelQueuedRunForPluginGate()` modeled on the existing `cancelQueuedRunForBlockedDependencies()` pattern; uses a guarded `UPDATE ... WHERE status = 'queued'` to prevent racing with the concurrent dispatcher
- `CancelSource` union extended with `"plugin_gate"`
- Audit log entry `run.plugin_gate_veto` written on every veto; `actorId` is always taken from the host registry, never from any field the plugin returns
- Veto `reason` capped at 500 chars, non-serializable fields stripped
- Runaway-detector exemption for `cancel_source = "plugin_gate"` paired with a compensating rate cap (warning emitted if more than 20 plugin-gate cancels for one company within 60s)

**UI**
- Run detail view (`AgentDetail.tsx`) and `IssueRunLedger.tsx` render the `plugin_gate` cancel source with the human-readable label "Cancelled by plugin gate" so operators can see why a queued run never started

## Verification

**RGR TDD:** Each component followed strict RED → GREEN → REFACTOR. Tests landed before the corresponding implementation.

**Tier 1 — Unit (Vitest, mocked I/O)**
- `before-run-types.test.ts` — type-shape and SDK surface assertions
- `run-gate-capability.test.ts` — capability constant present in `PLUGIN_CAPABILITIES`
- `broadcast-before-run.test.ts` — fast path with zero gate plugins, multi-handler aggregation by install order, first-veto-wins, fail-open on throw / timeout / malformed result, deterministic vetoing pluginId, per-plugin 5s timeout

**Tier 2 — Integration (real embedded Postgres)**
- `heartbeat-run-gate.test.ts` — plugin registers handler, run is queued, gate vetoes, row ends as `cancelled` with `cancel_source = 'plugin_gate'` and `error = <reason>`; allow path lets the claim through and emits `agent.run.started`; vetoed runs do NOT emit `agent.run.started`

**Tier 3 — E2E / contract**
- Covered by the integration suite above (full claim-path stack with real Postgres + heartbeat dispatcher)

**Tier 4 — RBAC + audit**
- `ws1-rbac-and-audit.test.ts` — plugin without `run.gate` is rejected at registration; plugin with `run.gate` registers successfully; veto emits exactly one `run.plugin_gate_veto` activity row; `actorId` always sourced from the host registry, never from any plugin-supplied value; cross-company isolation (plugin scoped to company A cannot gate company B's runs)

**Pre-Greptile self-review:** Ran `codex exec --model gpt-5.4-mini -c model_reasoning_effort=xhigh` over the full diff before submission. Self-review prompt at `.ws1-xhigh-prompt.txt` in the implementation tree. Findings addressed prior to opening this PR.

**Adversarial review:** Both `gpt-5.5` codex (high) and `claude-sonnet-4.6` reviewed the design. Resolved findings include: company scoping in `broadcastBeforeRun`, deterministic veto ordering by `installOrder`, fail-open documentation, removal of `contextSnapshot` from `BeforeRunParams` (too sensitive), removal of `retryable` from v1 (deferred until real requeue backoff exists), guarded cancel UPDATE to defeat TOCTOU with the concurrent dispatcher, `actorId` trust boundary, runaway-detector exemption + compensating rate cap, and UI surface for the new cancel source.

## Risks

1. **Fail-open semantics.** A handler timeout, crash, or network error allows the run through. This is intentional: a buggy gate plugin must not deadlock the run queue. Documented in the `onBeforeRun` JSDoc. A manifest-level `failPolicy: "fail-closed"` opt-in is deferred to a follow-up PR.
2. **Startup window.** Gate registrations live in worker memory. During a plugin worker restart, runs may proceed without gating until `setup()` re-registers the handler. This is inherent to the async startup model and is documented.
3. **`retryable` removed from v1.** The earlier design sketch included `retryable: boolean` on `BeforeRunResult`. Removed to avoid misleading UX without real requeue backoff. Will be added in v2.
4. **Bus semantics unchanged.** The existing `PluginEventBus.emit()` is not modified. The gate is a parallel dispatch path via `pluginWorkerManager.call()` so existing event handlers are unaffected.
5. **Runaway-detector tradeoff.** Plugin-gate cancels are exempted from the existing runaway heuristic, paired with the 20-cancels-per-60s warning rate cap as a compensating control.

## Checklist

- [x] PR template followed (Thinking Path, What Changed, Verification, Risks, Model Used, Checklist)
- [x] All four test tiers covered (unit, integration, e2e/contract, RBAC + audit)
- [x] RGR TDD followed (RED → GREEN → REFACTOR per component)
- [x] No API-only feature: UI surfaces the new `plugin_gate` cancel source in the run detail view
- [x] Capability gating enforced at registration AND at dispatch
- [x] Audit log written on every veto with host-assigned `actorId`
- [x] Cross-tenant isolation tested
- [x] Fail-open semantics documented in SDK JSDoc
- [x] Pre-Greptile self-review (gpt-5.4-mini xhigh) completed before submission
- [x] No schema migration required (uses existing `cancelSource` column)


supertaz:
@greptileai recheck

Addressed all 3 P2 findings from first review (commit 02c90b2c8):

1. **Tracker key:** Changed from companyId-only to companyId:vetoPluginId so each plugin is tracked independently. Log message clarified. Test added for two plugins in same company tracked independently.

2. **Gate query efficiency:** Added JSONB containment pre-filter in SQL WHERE clause so only run.gate plugins are fetched from DB.

3. **invocationSource:** Added ?? "" default for belt-and-suspenders safety.

PR body updated with full Thinking Path, Model Used, Risks, Verification sections per CONTRIBUTING.md.

ammarwa (approved):

ammarwa (approved):

Issue #5004, opened 2026-05-02
opencode_local adapter: Bun runtime leaks libopentui.so into /tmp/ — exhausts tmpfs, causes ENOSPC


lunarpulse:
## Summary
Every opencode process spawned by the opencode_local adapter leaks a ~4.3 MB shared library file (libopentui.so) into /tmp/. These files are never cleaned up. On systems where /tmp is a tmpfs (RAM-backed), they accumulate until the filesystem is exhausted, causing all agent runs to fail with ENOSPC.
## Environment
- Paperclip version: 0.3.1
- OS: Ubuntu 24.04 (also affects any Linux with tmpfs /tmp)
- Disk: 16 GB root partition
- /tmp mount: tmpfs 7.9G
- opencode binary: Bun-compiled, bundles @opentui/core-linux-x64@0.1.96
## Reproduction
1. Run Paperclip with the opencode_local adapter
2. Let the heartbeat scheduler execute agent runs over time (hours/days)
3. Each run spawns ~/.opencode/bin/opencode which extracts libopentui.so to /tmp/.3ef7{hash}-00000000.so
4. Files accumulate with no cleanup

## Symptoms
ENOSPC: no space left on device, copyfile
  '/home/user/.config/opencode/node_modules/@opencode-ai/plugin/dist/tool.js'
  -> '/tmp/paperclip-opencode-config-XXXXXX/opencode/node_modules/@opencode-ai/plugin/dist/tool.js'
$ df -h /tmp
Filesystem      Size  Used Avail Use% Mounted on
tmpfs           7.9G  7.6G  254M  97% /tmp
## Root Cause
The opencode binary is compiled with Bun and bundles @opentui/core-linux-x64 which contains libopentui.so — a Zig-based terminal UI native library. When Bun loads this native addon, it extracts it from its internal virtual filesystem (/$bunfs/root/libopentui-ayp6xrpn.so) to /tmp/ using mkstemp(). Bun does not delete these files on process exit.
Call chain:
heartbeat scheduler
  → server/src/services/heartbeat.ts:2642  getServerAdapter("opencode_local")
  → server/src/adapters/registry.ts:150    openCodeLocalAdapter.execute
  → packages/adapters/opencode-local/src/server/execute.ts:319
       runChildProcess("opencode", ["run", "--format", "json", ...])
  → spawns: ~/.opencode/bin/opencode
  → Bun extracts libopentui.so → /tmp/.3ef7{hash}-00000000.so
[...]

PR #5038, opened 2026-05-02
fix leak so, bun ENPSPC error issues

lunarpulse:
Thinking Path

> - Paperclip orchestrates AI agents for zero-human companies, spawning agent processes via adapters like opencode_local
> - The opencode_local adapter runs the OpenCode CLI via Bun, which extracts native shared libraries (e.g. libopentui.so, ~4.3 MB each) into os.tmpdir() on every process load via mkstemp() and never cleans them up
> - On systems where /tmp is a tmpfs (RAM-backed, typically 8 GB), these orphaned .so files accumulate over hours/days of heartbeat-driven agent runs until the filesystem is exhausted — observed in production: 1,840 files = 7.6 GB, 97% full, causing system-wide ENOSPC failures
> - The existing execute.ts lifecycle has a finally block that cleans the runtime config temp directory, but nothing addresses the Bun-extracted .so files
> - This pull request adds a cleanup function reapBunTempSharedLibs() that scans os.tmpdir() for files matching Bun's mkstemp naming pattern (validated by ELF magic bytes) and unlinks them, integrated into the existing finally block so it runs after every adapter execution
> - The benefit is preventing silent disk exhaustion in production environments where Paperclip agents run continuously, without depending on an upstream Bun fix
What Changed

- New bun-cleanup.ts — Pure module exporting reapBunTempSharedLibs(tmpdir: string): Promise<number>, a BUN_TEMP_SO_RE regex matching Bun's .{hex16+}-{8-digit}.so mkstemp pattern, and an ELF_MAGIC constant for 4-byte ELF header validation. The function scans the given tmpdir, filters entries by name pattern and ELF magic, and unlinks matching files while tolerating EBUSY/EPERM/ENOENT from concurrent processes. Per-file ELF validation errors are isolated so one bad file doesn't abort the scan.
- New bun-cleanup.test.ts — 21 unit tests covering: pattern matching (primary + variant hex), negative cases (non-matching .so, non-matching hidden files, no-ELF content, sub-4-byte files), error paths (EBUSY, EPERM, ENOENT, unexpected error propagation), edge cases (empty directory, nonexistent tmpdir, I/O error on one file doesn't block others), and regex boundary conditions (exactly 16 chars vs. fewer).
- Modified execute.ts — Added import of reapBunTempSharedLibs and restructured the finally block: preparedRuntimeConfig.cleanup() is now wrapped in its own try/catch with error logging (previously unhandled), and the .so reaper runs independently afterward, logging a summary to stderr when files are reaped.
Verification

- Run the adapter-local tests: npx vitest run packages/adapters/opencode-local — all 28 tests pass (21 new + 7 existing)
- Run typecheck: pnpm --filter @paperclipai/adapter-opencode-local typecheck — clean, no errors
- Run full suite: pnpm test:run — 148/151 test files pass (3 pre-existing failures in server/ from unbuilt @paperclipai/plugin-sdk, unrelated to this change)
- Manual smoke test: create a fake Bun .so file in /tmp (e.g. dd if=/dev/zero bs=1 count=8 of=/tmp/.3ef7f1ffe5bfeeef-00000000.so && printf '\x7fELF' | dd of=/tmp/.3ef7f1ffe5bfeeef-00000000.so bs=1 count=4 conv=notrunc), trigger an opencode_local adapter execution, then verify the file is gone and the [paperclip] Cleaned up N orphaned .so files message appears in stderr logs
Risks

- Low risk. The cleanup is entirely additive — it runs in the finally block wrapped in its own try/catch, so a failure cannot affect the adapter execution result. No API, schema, or behavioral changes.
- Performance consideration: On systems with very large /tmp directories (tens of thousands of entries), the fs.readdir + per-file ELF check adds marginal latency once per execute() call. With the observed production scale (~1,840 files), this is negligible. If it becomes a concern, the ELF check can be made opt-in via env var in a future iteration — the name pattern alone is a strong signal.
- Regex brittleness: If Bun changes its mkstemp naming significantly, the regex will need updating. The ELF magic check provides a safety net. The GitHub issue tracking the upstream Bun behavior should be monitored.
- Scope limitation: This only cleans up after opencode_local adapter executions. Manual opencode runs outside Paperclip will still accumulate .so files. This is acceptable — the adapter is the high-frequency production path.
> For core feature work, check ROADMAP.md (ROADMAP.md) first and discuss it in #dev before opening the PR. Feature PRs that overlap with planned core work may need to be redirected — check the roadmap first. See CONTRIBUTING.md.
Model Used

GLM-5.1 (zai-coding-plan/glm-5-1) — used for implementation, test authoring, and adversarial review. Guided by the BMAD quick-dev workflow with tech spec as input.
Checklist

### DIFF
--- a/packages/adapters/opencode-local/src/server/bun-cleanup.test.ts
+++ b/packages/adapters/opencode-local/src/server/bun-cleanup.test.ts
@@ -1,0 +1,269 @@
+import fs from "node:fs/promises";
+import os from "node:os";
+import path from "node:path";
+import { afterEach, describe, expect, it, vi } from "vitest";
+import { BUN_TEMP_SO_RE, ELF_MAGIC, reapBunTempSharedLibs } from "./bun-cleanup.js";
+
+const cleanupPaths = new Set<string>();
+
+afterEach(async () => {
+  await Promise.allSettled(
+    [...cleanupPaths].map(async (filepath) => {
+      await fs.rm(filepath, { recursive: true, force: true });
+      cleanupPaths.delete(filepath);
+    }),
+  );
+  vi.restoreAllMocks();
+});
+
+async function makeTmpdir() {
+  const dir = await fs.mkdtemp(path.join(os.tmpdir(), "paperclip-bun-cleanup-test-"));
+  cleanupPaths.add(dir);
+  return dir;
+}
+
+/** Write a minimal ELF header (8 bytes: \x7fELF + padding). */
+async function writeElfFile(filePath: string) {
+  await fs.writeFile(filePath, Buffer.from([0x7f, 0x45, 0x4c, 0x46, 0x02, 0x01, 0x01, 0x00]));
+}
+
+describe("reapBunTempSharedLibs", () => {
+  it("deletes files matching Bun mkstemp pattern", async () => {
+    const tmpdir = await makeTmpdir();
+    const fileName = ".3ef7f1ffe5bfeeef-00000000.so";
+    const filePath = path.join(tmpdir, fileName);
+    await writeElfFile(filePath);
+
+    const count = await reapBunTempSharedLibs(tmpdir);
+
+    expect(count).toBe(1);
+    await expect(fs.access(filePath)).rejects.toThrow();
+  });
+
+  it("deletes files with variant hex prefix", async () => {
+    const tmpdir = await makeTmpdir();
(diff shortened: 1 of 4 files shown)


lunarpulse:
related to #5004 , solving the core problems.

PR #5048, opened 2026-05-02
fix: degrade git_worktree to project_primary when baseCwd is not a git repo

brunocardoso-netizen:
## Thinking Path

> - Paperclip orchestrates AI agents that run inside execution workspaces, with a `workspaceStrategy` config that can be set to `git_worktree` to automatically create per-issue git worktrees
> - When an agent has `workspaceStrategy.type = "git_worktree"` and fires an **interval heartbeat** with no assigned issue, `resolveWorkspaceForRun` finds no project context and falls back to the agent-home directory (`~/.paperclip/instances/default/workspaces/<agentId>/`)
> - That directory is not a git checkout, but `realizeExecutionWorkspace` passes it to `resolveGitOwnerRepoRoot` unconditionally, which calls `git rev-parse --show-toplevel` and throws `fatal: not a git repository`
> - The run slot is wasted and an ERROR is logged, even though the agent had nothing meaningful to do — the crash is preventable
> - The file already contains an `isGitCheckout` helper (line 583) that can be used as a guard before the git operations begin
> - This PR adds a guard in `realizeExecutionWorkspace`: when `strategyType === "git_worktree"` but `baseCwd` is not a git checkout, return `project_primary` strategy with a warning instead of throwing
> - The benefit is clean, silent degradation for issueless interval wakes, and the warning gives operators an observable signal if something unexpected sends a non-git path into this code path

## What Changed

- `server/src/services/workspace-runtime.ts`: added an `isGitCheckout(input.base.baseCwd)` guard immediately after the `strategyType !== "git_worktree"` early-return. If `baseCwd` is not a git repository, the function now returns `project_primary` strategy with a warning message instead of throwing.
- `server/src/__tests__/workspace-runtime.test.ts`: added a regression test `"degrades to project_primary when baseCwd is not a git repo (interval heartbeat with no project)"` that creates a plain temp directory (not a git repo), calls `realizeExecutionWorkspace` with `git_worktree` strategy, and asserts `strategy === "project_primary"`, `created === false`, and a non-empty warning.

## Verification

```bash
# Run the targeted new test (passes, was failing before the fix on master)
pnpm run preflight:workspace-links && npx vitest run server/src/__tests__/workspace-runtime.test.ts -t "degrades to project_primary when baseCwd is not a git repo"

# Run the full workspace-runtime test file (53 pass, 4 pre-existing failures
# in provision-script tests that also fail on master due to local CLI env)
npx vitest run server/src/__tests__/workspace-runtime.test.ts

# Typecheck clean
pnpm typecheck
```

All other `realizeExecutionWorkspace` tests continue to pass — the guard is only reached when `baseCwd` is not a git checkout, which no existing test triggers.

## Risks

Low risk. The guard only fires when `isGitCheckout` returns `false` — the path was previously an unconditional crash, so no existing behaviour is changed for valid git checkouts. The degraded return is identical in shape to the `strategyType !== "git_worktree"` early return that already exists above it.

## Model Used

- **Provider:** Anthropic  
- **Model:** Claude Sonnet 4.6 (`claude-sonnet-4-6`)  
- **Context window:** 200k tokens  
- **Capabilities:** tool use, code execution, file read/write  
- **Reasoning mode:** standard (no extended thinking)

## Checklist

### DIFF
--- a/server/src/__tests__/workspace-runtime.test.ts
+++ b/server/src/__tests__/workspace-runtime.test.ts
@@ -1848,6 +1848,38 @@ describe("realizeExecutionWorkspace", () => {
     expect(worktreeOp!.metadata!.baseRef).toBe("master");
   }, 10_000);
 
+  it("degrades to project_primary when baseCwd is not a git repo (interval heartbeat with no project)", async () => {
+    const nonGitDir = await fs.mkdtemp(path.join(os.tmpdir(), "paperclip-non-git-"));
+
+    const result = await realizeExecutionWorkspace({
+      base: {
+        baseCwd: nonGitDir,
+        source: "agent_home",
+        projectId: null,
+        workspaceId: null,
+        repoUrl: null,
+        repoRef: null,
+      },
+      config: {
+        workspaceStrategy: { type: "git_worktree" },
+      },
+      issue: null,
+      agent: {
+        id: "agent-1",
+        name: "Codex Coder",
+        companyId: "company-1",
+      },
+    });
+
+    expect(result.strategy).toBe("project_primary");
+    expect(result.cwd).toBe(nonGitDir);
+    expect(result.branchName).toBeNull();
+    expect(result.created).toBe(false);
+    expect(result.warnings).toHaveLength(1);
+    expect(result.warnings[0]).toContain("not");
+    expect(result.warnings[0]).toContain("project_primary");
+  });
+
   it("removes a created git worktree and branch during cleanup", async () => {
     const repoRoot = await createTempRepo();
 
--- a/server/src/services/workspace-runtime.ts
+++ b/server/src/services/workspace-runtime.ts
@@ -999,6 +999,20 @@ export async function realizeExecutionWorkspace(input: {
     };
   }
 
+  if (!await isGitCheckout(input.base.baseCwd)) {
+    return {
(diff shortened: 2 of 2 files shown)


brunocardoso-netizen:
LGTM (Captain). Minimal guard using the existing `isGitCheckout` helper, scoped to the crash path. No behavior change for valid git checkouts. Test reproduces the exact failure mode (interval wake, no project, non-git baseCwd) and asserts the graceful degrade. Holding merge until verify + e2e go green.

ammarwa (approved):

ammarwa (approved):

ammarwa (approved):

Elhadnaits:
We hit this exact failure in production (deterministic `adapter_failed: not a git repository` on fresh project-less issues with `git_worktree` settings) and this guard is the right fix — thank you. Since this branch predates the `baseRefSha` field on the realized-workspace return (the early return here no longer typechecks against current master) and has been quiet since early May, I've opened #8061: rebased on current master with the same baseCwd guard (credited to this PR), plus a fix for the adjacent stale-registration hard-fail (`Registered worktree ... is not reusable`) and tests for both. Happy to defer to this PR if you'd rather rebase it — whichever lands first solves the bug.

PR #5059, opened 2026-05-02
feat(routes): GET /heartbeat-runs/:runId honours ?include=resultJson for full JSONB

justinltodd:
## Summary

External adapters (and dashboards / smoke tests / agent consumers) sometimes need the full `result_json` JSONB on a heartbeat run. Today the user-facing `GET /heartbeat-runs/:runId` route projects through `heartbeatRunSafeColumns`, which collapses oversized JSONB (>64 KiB) to `{summary, result, message, error, costUsd, truncated: true}` — losing rich adapter telemetry like tool-call audit, retry history, MCP server health, transcript-cap state, and (for adapters that emit them) liveness state / progressBeats.

Paperclip already supports `getRun(runId, { unsafeFullResultJson: true })` internally for retry-decision logic in `services/heartbeat.ts`. This PR exposes the same flag through the user-facing route via a query parameter, gated by the existing `assertCompanyAccess` auth check and the existing `redactCurrentUserValue` redactor. No new auth surface, no new endpoint, no new policy — just a more inclusive projection on demand.

## API

```
GET /api/heartbeat-runs/:runId?include=resultJson
GET /api/heartbeat-runs/:runId?expand=full
```

Either query key turns the safe projection off and returns the full `result_json` JSONB. Anything else (or absence) preserves today's safe-projection behaviour.

## Why this matters

- Dashboards reading liveness / cost / tool-call telemetry without ECS-exec + psql.
- Smoke-test scripts asserting `result_json.adapterVersion`.
- A future Staff Engineer agent inspecting full run records via HTTP rather than shelling out to a DB.

For external adapters that emit verbose telemetry (e.g. our hermes-paperclip-adapter fork), the safe-projection collapse is the difference between "the adapter is observable" and "the adapter is a black box".

## Test plan

- [x] Unit test in `server/src/__tests__/heartbeat-runs-result-json-projection.test.ts`: `GET /heartbeat-runs/:runId` returns safe-projection by default; same endpoint with `?include=resultJson` returns full JSONB; `?expand=full` is equivalent; arbitrary `?include=foo` is treated as absence; 404 still fires when the run is missing.
- [x] Existing route tests still pass.
- [x] Auth boundary: `?include=resultJson` from a non-board user is still gated by `assertCompanyAccess`.

### DIFF
--- a/server/src/__tests__/heartbeat-runs-result-json-projection.test.ts
+++ b/server/src/__tests__/heartbeat-runs-result-json-projection.test.ts
@@ -1,0 +1,161 @@
+import express from "express";
+import request from "supertest";
+import { beforeEach, describe, expect, it, vi } from "vitest";
+
+const mockHeartbeatService = vi.hoisted(() => ({
+  getRun: vi.fn(),
+  getRetryExhaustedReason: vi.fn(),
+  buildRunOutputSilence: vi.fn(),
+}));
+
+const mockInstanceSettingsService = vi.hoisted(() => ({
+  get: vi.fn(),
+  getExperimental: vi.fn(),
+  getGeneral: vi.fn(),
+  listCompanyIds: vi.fn(),
+}));
+
+function registerModuleMocks() {
+  vi.doMock("../routes/authz.js", async () => vi.importActual("../routes/authz.js"));
+
+  vi.doMock("../services/heartbeat.js", () => ({
+    heartbeatService: () => mockHeartbeatService,
+  }));
+
+  vi.doMock("../services/instance-settings.js", () => ({
+    instanceSettingsService: () => mockInstanceSettingsService,
+  }));
+
+  vi.doMock("../services/index.js", () => ({
+    agentService: () => ({}),
+    agentInstructionsService: () => ({}),
+    accessService: () => ({}),
+    approvalService: () => ({}),
+    companySkillService: () => ({ listRuntimeSkillEntries: vi.fn() }),
+    budgetService: () => ({}),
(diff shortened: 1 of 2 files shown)


ammarwa (approved):

ammarwa (approved):

justinltodd:
Thanks @greptile-apps for the careful read.

Pushed two follow-ups in 015d2c2:

1. **PR template sections** — updated the description to include Thinking Path, What Changed, Verification, Risks, Model Used, and the Checklist per `.github/PULL_REQUEST_TEMPLATE.md` and `CONTRIBUTING.md`.
2. **P2: array `req.query.include` edge case** — replaced the strict `=== "resultJson"` check with a small `queryHas(raw, target)` helper that handles both `string` and `string[]` shapes (Express produces `string[]` for repeated params like `?include=foo&include=resultJson`). Now any element matching the target opts in; non-string values (object/null/undefined) preserve today's safe-projection default. Added two new test cases — array-with-resultJson opts in, array-of-unknowns falls back. Existing tests are unchanged.

Ready for re-review.

justinltodd:
Apologies for the noise — I force-pushed the branch back to the approved commit `8d016a4` and restored the original PR description. The follow-up commit I pushed earlier (`015d2c2`) was overreach on my part: the PR was already approved, and my change broke the previously-green `verify` check. Reverted in full. Ready as it was.

justinltodd:
@ammarwa thanks again for the approval. This has been green since the 5th — is there anything blocking a merge, or is it waiting on a release window? Happy to rebase if that helps.

PR #5067, opened 2026-05-03
add GitHub ingress host prefilter

keegoid-codex:
## Thinking Path

> - Paperclip orchestrates AI-agent companies through a shared control plane.
> - The plugin subsystem lets product-specific integrations live at the edge instead of expanding core behavior.
> - GitHub PR ingress plugins need to reject unauthenticated webhook traffic before Paperclip records delivery rows or dispatches plugin handlers.
> - The shared plugin contract did not expose a host-owned webhook prefilter, so plugins had to do verification later in the request path.
> - This pull request adds a narrow host prefilter contract for GitHub HMAC SHA-256 webhooks, wires it through validation, docs, and server dispatch, and keeps plugin delivery semantics unchanged after verification.
> - The benefit is safer webhook ingress for plugins without embedding GitHub-specific code in each plugin handler.

## What Changed

- Added `hostPrefilter` plugin manifest support for GitHub HMAC SHA-256 webhook verification with a configurable secret reference and max body size.
- Enforced the host prefilter in plugin webhook dispatch before delivery persistence and plugin invocation.
- Documented the plugin contract and exported the new shared types and validators.
- Added server tests for valid signatures, invalid signatures, missing secrets, unsupported declarations, and request-size rejection.
- Normalized repeated issue-list query parameters and covered repeated `status` handling.
- Preserved common user bin directories in adapter subprocess `PATH`, including Cursor PATH ordering behavior.
- Added GPT-5.5 fast-mode metadata for the Codex local adapter.
- Relaxed plugin capability escalation to a warning so lifecycle approval remains the enforcement point.

## Verification

- `pnpm -r typecheck`
- `pnpm test:run`
- `pnpm --filter @paperclipai/adapter-cursor-local typecheck`
- `pnpm exec vitest run server/src/__tests__/cursor-local-adapter-environment.test.ts server/src/__tests__/cursor-local-execute.test.ts packages/adapters/cursor-local/src/server/remote-command.ts`
- `git diff --check`
- `pnpm build`

## Risks

- Low to medium risk. The host prefilter only runs for plugins that declare it, but webhook body handling is security-sensitive.
- Misconfigured plugin manifests now fail ingress earlier with `401`, `413`, or `500` responses depending on the failure mode.
- The PATH preservation change affects adapter subprocess environments, so Cursor ordering was specifically retested.

> For core feature work, check [`ROADMAP.md`](ROADMAP.md) first and discuss it in `#dev` before opening the PR. Feature PRs that overlap with planned core work may need to be redirected -- check the roadmap first. See `CONTRIBUTING.md`.

## Model Used

- OpenAI Codex, GPT-5.5, tool use and local code execution, medium reasoning.

## Checklist


keegoid-cc (commented):
🗒️ TASK: Adversarial review of GitHub ingress prefilter PR

### Summary
Adds an opt-in host-side `hostPrefilter` (GitHub HMAC SHA-256) that the server enforces before recording webhook deliveries; ships a matching example plugin; refactors heartbeat run status writes to be CAS-style with `previousStatus`, adds run-lifecycle plugin events, normalizes repeated issue-list query params, prepends user bin dirs to adapter `PATH`, adds GPT-5.5 fast-mode metadata, and downgrades capability-escalation rejection from a hard error to a warning.

### Bugs & Correctness
- [CAT-1] `server/src/services/plugin-loader.ts` capability-escalation: the previous code threw on any new capability added during upgrade. The new code only logs a warning and the comment claims "lifecycle will require approval before activation", but the diff does not show any new gate in `pluginLifecycleManager` or `registry.upgrade` that actually withholds activation pending approval. If no such gate exists, this is a real regression: an upgrade can now silently grant capabilities (e.g. `secrets.read-ref`, `webhooks.receive`) that the previous version did not declare. Either the gate must exist and be cited, or this change must be reverted/replaced with an explicit pending-approval state.
- [CAT-2] `packages/adapter-utils/src/server-utils.ts` `defaultUserBinDirs` falls back to `process.env.HOME` when the *passed* `env` lacks `HOME`. Callers that intentionally pass a sanitized env without `HOME` will now get the host operator's `~/.local/bin` and `~/bin` appended to the child's `PATH`. This contradicts the previous "PATH untouched if already present" contract and can change which executables a sandboxed subprocess resolves. Worth narrowing to "only use `env.HOME`, never the host process HOME".
- [CAT-2] `packages/adapters/cursor-local/src/server/remote-command.ts`: `prependPosixPathEntry` changed from idempotent (no-op if already present) to "remove all existing copies, then prepend". That is a behavior change, not just a refactor — for any caller that previously relied on stable ordering when `entry` was already mid-PATH, the entry now jumps to the front. Tests cover the Cursor case, but the function is generic and named generically; consider whether other callers want the old idempotent semantics.
- [CAT-2] `server/src/services/heartbeat.ts` `cancelActiveForAgentInternal`: when the CAS cancel write loses to a terminal status (e.g. `succeeded`) but `runningProcesses` still holds a live child handle (race against the success finalizer which hasn't yet `delete`d), the new branch logs a warning and skips kill. That can leak a still-running child until the success finalizer runs. The previous code unconditionally killed. Probably acceptable because the success finalizer is the authoritative owner, but worth confirming the success path always clears `runningProcesses` even on partial failure paths.
- [CAT-2] `packages/plugins/examples/plugin-github-pr-ingress/src/worker.ts` records "ignored" deliveries only in process-local `runtime.lastDelivery`, never in `ctx.state`. That is correct for retry-after-mapping semantics, but the README/PLUGIN_SPEC line "Every accepted delivery is recorded" is the only doc — `ignored` deliveries are silently invisible to operators outside the in-memory health widget, which is lost on worker restart. Consider whether ignored-event observability matters for this connector.
- [CAT-2] `worker.ts` `firstHeader` performs a case-insensitive scan, but `assertWebhookBodySize` runs *before* signature verification. That's intentional (cheap reject) but the size limit (`1 MiB`) is enforced again at the host layer with a separate default constant. Two independent limits with the same default is fine; if either is later tuned, they must be kept in sync or the host's smaller limit wins silently.

### Security & Safety
- [CAT-1] Capability-escalation downgrade (see Bugs §1) is the most security-relevant item in this PR. Until the lifecycle-approval gate is shown to exist, treat plugin upgrades as a privilege-escalation vector.
- [CAT-2] `server/src/routes/plugins.ts` host prefilter resolves the secret on every signed request via `secretsForPlugin(...).resolve(...)`. No caching, no rate limit. A spammer with knowledge of the public webhook URL can force one secret-resolution per request. Acceptable for a single-operator deployment, but document the assumption or add a token-bucket on the unauthenticated prefix.
- [CAT-2] `verifyGithubWebhookHmacSha256` uses `timingSafeEqual` on byte buffers of equal length — good. The early `signature.startsWith("sha256=")` check returns false in non-constant time relative to the secret, but the secret is never compared in that branch, so no leak. Fine.
- [CAT-3] The host parses `req.body` as JSON via `express.json` *and* re-uses `req.rawBody` for HMAC. If body-parser middleware order ever changes such that `rawBody` is unset, HMAC will silently fail-open against an empty buffer (signature mismatch → 401, not fail-open). Confirm via a regression test that missing `rawBody` always 401s rather than 500s.

### Design & Maintainability
- [CAT-3] `setRunStatus(runId, status, previousStatus, patch)` overloads same-status writes (metadata refresh) with status transitions. The "skip publish if `previousStatus === updated.status`" guard is correct but subtle — every call site now has to pass `run.status` for refreshes (`run.status, run.status, …`). A separate `refreshRunMetadata` helper would prevent the foot-gun where a future caller forgets the duplicate argument and accidentally publishes a phantom transition.
- [CAT-3] `preserveCancelledRunArtifacts` is a second update path that bypasses `setRunStatus` entirely (no lifecycle event, no live event). That is the correct invariant — terminal-cancel artifacts shouldn't re-publish — but worth a one-line comment at the call sites in the failure-race block to make clear that runtime bookkeeping is intentionally happening *without* a status transition event.
- [CAT-3] Two parallel definitions of "max webhook body bytes" (`MAX_WEBHOOK_BODY_BYTES` in the example plugin worker, `WEBHOOK_HOST_PREFILTER_DEFAULT_MAX_BODY_BYTES` in the server) with the same default. If the manifest's `maxBodyBytes` is ever extended to a per-plugin override, the worker's hardcoded 1 MiB will silently disagree.
- [CAT-3] `pnpm-workspace.yaml` excludes the new example with `!packages/plugins/examples/plugin-github-pr-ingress` and the package's `prebuild` invokes `pnpm --filter @paperclipai/plugin-sdk ensure-build-deps`. That filter only resolves when the SDK is part of the active workspace graph — fine in this repo, but the keegoid-local installation flow documented in the README installs from an absolute path, which will work because the SDK is built before install. Worth a comment in the README.
- [CAT-3] `.gitignore` adds `.claude-reviews/` at repo root — fine, but unrelated to the stated PR scope. Not blocking.

### Questions for the author
- Where is the lifecycle-side approval gate for capability escalation that the new warning message refers to? Cite the file/function so reviewers can verify the warning is not the only enforcement.
- Is `defaultUserBinDirs` intended to read `process.env.HOME` when `env.HOME` is missing, or should it strictly use the supplied env's HOME? The current behavior is a leak surface for child processes that deliberately strip HOME.
- For `prependPosixPathEntry`, was the move-to-front semantics required by Cursor specifically, or accidental? If accidental, restore idempotent behavior; if required, rename to `moveOrPrependPosixPathEntry` and document.
- In `cancelActiveForAgentInternal`, when the CAS cancel loses to a terminal `succeeded` write but `runningProcesses` still has an entry, what guarantees the success finalizer will eventually clear that entry and reap the child?

### Severity of findings
**Medium** — one CAT-1 (capability-escalation downgrade) is potentially a real privilege-escalation regression that needs verification before merge; the rest are CAT-2/3 polish and observability concerns.

🔧 CHANGE: No code edited; review only
✅ VERIFY: Findings tied to specific files/functions in diff
🗣️ Fig: Mostly solid concurrency and prefilter work; the loosened capability-escalation check is the one thing I'd block on until the lifecycle-approval gate is shown to exist.

VERDICT: request_changes

---
claude-review posting override: forced to --comment because reviewer lacks verified write permission (viewerPermission=READ; was --request-changes). GitHub only counts approvals from WRITE, MAINTAIN, or ADMIN reviewers.

keegoid-cc (commented):
🗒️ TASK: Adversarial review of GitHub PR ingress + lifecycle PR

### Summary
Adds a host-side `hostPrefilter` (GitHub HMAC SHA-256) for plugin webhooks; ships an example `plugin-github-pr-ingress`; expands run-lifecycle plugin events with usage/result summaries and tightens cancel/finalize status races in `heartbeat.ts`; relaxes plugin-loader capability escalation from error to warning (gated downstream by `pluginLifecycleManager.upgrade` → `upgrade_pending`); preserves user bin dirs in adapter PATH; adds GPT‑5.5 fast-mode metadata; normalizes repeated `?status=` query params on the issue list endpoint.

### Bugs & Correctness
- [CAT-3] `packages/plugins/sdk/...` → `packages/adapter-utils/src/server-utils.ts:1054` `ensurePathInEnv` no longer returns early when `PATH` is already set — it now always appends `~/.local/bin` and `~/bin` to every adapter env that flows through it. This is a behavior change for every adapter, not just Cursor. The PR description says only "Preserves common user bin directories"; consider documenting the broader impact (or scoping the append to the Cursor adapter that actually needed it).
- [CAT-3] `packages/adapters/cursor-local/src/server/remote-command.ts:18` `moveOrPrependPosixPathEntry` now strips the entry from anywhere in PATH and re-prepends it. Previously a user-curated ordering with `~/.local/bin` deliberately late was respected; now it is silently overridden. Intentional per the comment, but worth flagging — there is no escape hatch for operators who want the system binary to win.
- [CAT-2] `server/src/services/plugin-loader.ts` capability escalation downgraded from `throw` to `warn`. Safety relies entirely on `pluginLifecycleManager.upgrade` putting the plugin into `upgrade_pending` and skipping `loadSingle`. If any other caller invokes `loader.upgradePlugin(...)` directly (now or later) it will silently activate a manifest with new capabilities. Worth a code comment on `upgradePlugin` stating "callers MUST gate on capability diff" and/or a unit test that the loader does not re-activate on its own. The new lifecycle test only proves the manager path is correct.
- [CAT-3] `server/src/routes/plugins.ts` `applyWebhookHostPrefilter` runs after Express has already buffered the body (limit 1 MB via `PLUGIN_API_BODY_LIMIT_BYTES`). The `maxBodyBytes` check is therefore a post-hoc reject, not real backpressure, and the schema allows `maxBodyBytes` up to 10 MB which can never be honored — anything > 1 MB is rejected by Express first with a different error. Either align the schema cap with the express limit or document the effective ceiling.
- [CAT-3] `server/src/routes/plugins.ts:232` `firstRequestHeader` returns the first value of a multi-valued header array. GitHub will not send duplicate `X-Hub-Signature-256` headers, but a malicious upstream proxy could; consider rejecting when the array length > 1 instead of silently picking `[0]`.
- [CAT-3] `server/src/services/heartbeat.ts` `cancelActiveForAgentInternal` retries only twice. If the run flips status more than once between fetch and update (rare but possible under adapter completion races), the cancel silently no-ops and the loop falls through to a process-termination decision based on the latest `finalRun.status`. The skip-termination branch logs a warning but does not surface to the caller. Acceptable defense-in-depth, but worth stronger telemetry — operators won't otherwise know a cancel was issued without effect.
- [CAT-2] `server/src/services/heartbeat.ts` `setRunStatus` now suppresses live/lifecycle event emission when `previousStatus === run.status`. A few callers (e.g. `productivityHold`, `continuationExhausted`, `process_lost` retry) intentionally call with the same status to update metadata; that path no longer publishes a `heartbeat.run.status` live event. If any UI listener depended on those metadata refreshes (livenessReason updates), it will go stale until the next genuine transition. Worth confirming — looks intentional but not stated.
- [CAT-3] `packages/plugins/examples/plugin-github-pr-ingress/src/worker.ts` `runtime` is a module-level singleton; the comment notes "one worker process per plugin instance." If the host ever multiplexes a single worker over multiple installs (now or in future runtimes), `lastDelivery` / `lastPrSync` would cross-contaminate. Worth keying these by `ctx.installId` defensively.

### Security & Safety
- [CAT-2] `verifyGithubWebhookHmacSha256` (server) and `verifyGithubSignature` (worker) are duplicated. Both are correct (length-pre-check then `timingSafeEqual`). Consider extracting to `@paperclipai/shared` so a future fix lands in both. Not a defect today.
- [CAT-3] `applyWebhookHostPrefilter` swallows `secretsForPlugin().resolve` errors and returns 503 "unavailable". A misconfigured/missing secret and a transient secret-store failure are now indistinguishable in the response. Consider logging the underlying error at warn level so operators can debug.

### Design & Maintainability
- [CAT-3] PR scope is much wider than the title implies. The host prefilter is ~150 lines; the heartbeat cancel/race refactor is ~400 lines and is the highest-risk change in the PR. These are independently reviewable and would have been better as separate PRs (or at least called out as the headline change in the description, where they currently appear as a single bullet under "What Changed").
- [CAT-3] `server/src/routes/issues.ts` `readQueryStringList` joins multi-value params with commas to feed into `svc.list`. This relies on the downstream parser splitting again. A typed `string[]` all the way to the service would be safer than the string round-trip; otherwise a status value containing a literal comma would silently split.
- [CAT-3] `pluginWebhookHostPrefilterDeclaration` only supports `"github-hmac-sha256"`. The 500 response in `applyWebhookHostPrefilter` for an unknown `kind` is unreachable today (the validator rejects it), but worth a comment so a future contributor adding e.g. Stripe doesn't think the route handles arbitrary kinds.

### Questions for the author
- Is there a real call site for `pluginLoader.upgradePlugin` outside `pluginLifecycleManager.upgrade`? If so, capability escalation is no longer enforced there.
- Did you intentionally drop live `heartbeat.run.status` events for same-status metadata refreshes (productivity hold / continuation exhausted / process_lost retry)? Any UI that surfaces `livenessReason` may now go stale.
- Why append user bin dirs to every adapter PATH instead of just Cursor's? The original problem statement and the only failing test were Cursor-specific.

### Severity of findings
**low** — no CAT-1 defects; the largest concern is reduced enforcement of plugin capability escalation in the loader, mitigated by the lifecycle manager but not by tests of the loader itself.

VERDICT: comment

PR #5099, opened 2026-05-03
fix(gemini-local): skip YOLO informational stderr line in failure fallback

JongHee-Ryu:
## Summary

The `gemini_local` adapter unconditionally invokes the Gemini CLI with `--approval-mode yolo` for unattended heartbeat execution (`packages/adapters/gemini-local/src/server/execute.ts:402, :475`), and the CLI in turn always prints

```
YOLO mode is enabled. All tool calls will be automatically approved.
```

to **stderr**, even on a healthy run.

When the run later exits non-zero for any reason (signal, abrupt termination, etc.), the existing `firstNonEmptyLine(stderr)` fallback in `execute.ts` and `summarizeProbeDetail` in `test.ts` surfaces this YOLO notice as the adapter's `errorMessage`. The Paperclip heartbeat layer then classifies the run as `adapter_failed` with that misleading message and the operator-facing agent flips to `status=error`, even though YOLO mode is the configured, intended behavior.

This has caused a self-reinforcing recovery loop on agents whose adapter is `gemini_local` — they show "YOLO mode is enabled. All tool calls will be automatically approved." as a failure reason and trigger stranded-recovery cycles.

## Fix

- Add `firstSignificantStderrLine` and `isInformationalStderrLine` in `utils.ts`. The new helper filters known informational lines (currently the YOLO notice) and returns the first non-empty, non-informational stderr line.
- Use `firstSignificantStderrLine` for the stderr fallback in both the heartbeat execute path (`execute.ts`) and the hello probe summary (`test.ts`).
- The `stdout` fallback intentionally keeps `firstNonEmptyLine`, since stdout is the canonical output channel.

The flag itself is intentionally kept — `--approval-mode yolo` is required for headless execution and is documented in `commandNotes`.

## Tests

- Add `utils.test.ts` covering both the legacy `firstNonEmptyLine` behavior and the new helper for: skipping YOLO leading lines, returning empty when only informational lines are present, and falling through whitespace.
- Add a per-package `vitest.config.ts` so the workspace's vitest project resolves correctly (other adapter packages, e.g. `claude-local`, follow this same pattern).

## Test plan

- [x] `pnpm --filter @paperclipai/adapter-gemini-local typecheck`
- [x] `pnpm exec vitest run packages/adapters/gemini-local/src/server/utils.test.ts` — 7 tests pass
- [x] `pnpm --filter @paperclipai/adapter-gemini-local build`
- [ ] Maintainer verification: in a deployed environment with `gemini_local` agents that previously flipped to `status=error` with the YOLO message, confirm the heartbeat run ends with a real stderr line (or `Gemini exited with code N`) instead of the YOLO notice.

### DIFF
--- a/packages/adapters/gemini-local/src/server/execute.ts
+++ b/packages/adapters/gemini-local/src/server/execute.ts
@@ -52,7 +52,7 @@ import {
   isGeminiUnknownSessionError,
   parseGeminiJsonl,
 } from "./parse.js";
-import { firstNonEmptyLine } from "./utils.js";
+import { firstSignificantStderrLine } from "./utils.js";
 
 const __moduleDir = path.dirname(fileURLToPath(import.meta.url));
 
@@ -552,7 +552,7 @@ export async function execute(ctx: AdapterExecutionContext): Promise<AdapterExec
       } as Record<string, unknown>)
       : null;
     const parsedError = typeof attempt.parsed.errorMessage === "string" ? attempt.parsed.errorMessage.trim() : "";
-    const stderrLine = firstNonEmptyLine(attempt.proc.stderr);
+    const stderrLine = firstSignificantStderrLine(attempt.proc.stderr);
     const structuredFailure = attempt.parsed.resultEvent
       ? describeGeminiFailure(attempt.parsed.resultEvent)
       : null;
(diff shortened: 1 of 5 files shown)


JongHee-Ryu:
@copilot apply changes based on the comments in [this thread](https://github.com/paperclipai/paperclip/pull/5099#pullrequestreview-4216436855)

PR #5113, opened 2026-05-03
fix(server): use loopback URL for PAPERCLIP_RUNTIME_API_URL in local_trusted mode

cnascimento91:
## Thinking Path

> - Paperclip orchestrates AI agents for zero-human companies; agents make API calls to the Paperclip server during every heartbeat
> - The server startup code sets `PAPERCLIP_RUNTIME_API_URL` via `choosePrimaryRuntimeApiUrl`, which returns `auth.publicBaseUrl` (a localtunnel/ngrok URL) as the primary API URL when configured
> - `buildPaperclipEnv` in `@paperclipai/adapter-utils` reads `PAPERCLIP_RUNTIME_API_URL` first and injects it as `PAPERCLIP_API_URL` into agent process environments
> - In `local_trusted` deployments, agents always run on the same machine as the server — they never need the public tunnel; they can always reach `http://127.0.0.1:{port}` directly
> - When the public tunnel goes down or changes hostname, agents fail silently because they try to reach the stale tunnel URL instead of loopback
> - This PR fixes the issue by setting `PAPERCLIP_RUNTIME_API_URL` to `http://{bindHost}:{port}` (loopback) for `local_trusted` deployments, so agents always use a stable local URL
> - The benefit is that agent runs remain connected to the API regardless of tunnel state; `auth.publicBaseUrl` is preserved for OAuth/external access

## What Changed

- `server/src/index.ts`: for `deploymentMode === "local_trusted"`, `PAPERCLIP_RUNTIME_API_URL` is now set to `http://${runtimeListenHost}:${listenPort}` instead of the value from `choosePrimaryRuntimeApiUrl` (which prefers `authPublicBaseUrl`)
- `PAPERCLIP_API_URL` (the server process env var, used for webhook URLs in routines etc.) is unchanged — it still reflects the public URL
- `auth.publicBaseUrl` is completely untouched — OAuth callbacks and external access are unaffected

## Verification

1. Configure a local Paperclip instance with `server.deploymentMode = "local_trusted"` and `auth.publicBaseUrl` set to a tunnel URL (e.g. `https://example.loca.lt`)
2. Start the server: `npx paperclipai run`
3. Trigger an agent heartbeat
4. In the heartbeat run logs, confirm `PAPERCLIP_API_URL` in the agent environment is `http://127.0.0.1:3100` (not the tunnel URL)
5. Bring the tunnel down; confirm agents continue to function normally

## Risks

- **Low risk**: the change is gated on `config.deploymentMode === "local_trusted"`, so it has zero effect on `authenticated` or `public` deployments
- For `local_trusted` mode: agents that somehow need to reach the server from a remote machine (unusual — local_trusted is defined as same-machine) would break, but `local_trusted` explicitly requires loopback host binding, making that configuration self-contradictory

## Model Used

- Provider: Anthropic
- Context: 200k tokens
- Capabilities: tool use, code execution, extended reasoning
- Used to: trace the root cause through `choosePrimaryRuntimeApiUrl` → `PAPERCLIP_RUNTIME_API_URL` → `buildPaperclipEnv` → agent env injection chain; implement the targeted fix

## Checklist

- [x] I have added or updated tests where applicable (no new tests — gated on existing `local_trusted` config path)

### DIFF
--- a/server/src/index.ts
+++ b/server/src/index.ts
@@ -651,7 +651,7 @@ export async function startServer(): Promise<StartedServer> {
   // The public URL (authPublicBaseUrl) is preserved separately for OAuth/external access.
   const agentRuntimeApiUrl =
     config.deploymentMode === "local_trusted"
-      ? `http://${runtimeListenHost}:${listenPort}`
+      ? `http://${runtimeListenHost.includes(":") ? `[${runtimeListenHost}]` : runtimeListenHost}:${listenPort}`
       : runtimeApiUrl;
   process.env.PAPERCLIP_LISTEN_HOST = runtimeListenHost;
   process.env.PAPERCLIP_LISTEN_PORT = String(listenPort);
--- a/server/src/index.ts
+++ b/server/src/index.ts
@@ -645,9 +645,17 @@ export async function startServer(): Promise<StartedServer> {
     bindHost: runtimeListenHost,
     port: listenPort,
   });
+  // For local_trusted deployments, agents run on the same machine as the server.
+  // Always use the loopback URL (http://127.0.0.1:port) for PAPERCLIP_RUNTIME_API_URL
+  // so agent runs stay connected even when the public tunnel is down or changes.
+  // The public URL (authPublicBaseUrl) is preserved separately for OAuth/external access.
+  const agentRuntimeApiUrl =
+    config.deploymentMode === "local_trusted"
+      ? `http://${runtimeListenHost}:${listenPort}`
+      : runtimeApiUrl;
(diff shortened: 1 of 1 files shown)


ammarwa (approved):

ammarwa (approved):

cnascimento91:
Bumping for merge — this PR has been approved by @ammarwa and all CI checks are passing since May 3. The fix is critical for local deployments: without it, agents lose connectivity whenever the public tunnel URL changes.

In the meantime we've applied the patch manually to the local dist file, but this needs to land upstream so it survives package updates.

Could a maintainer please merge? Thank you 🙏

PR #5120, opened 2026-05-03
Avoid resuming stale remote sessions (Pi adapter)

devinfoley:
> **Stacked PR (part 7 of 7).** Depends on:
  - PR #5114
  - PR #5115
  - PR #5116
  - PR #5117
  - PR #5118
  - PR #5119
> Diff against `master` includes commits from earlier PRs in the stack — the new commit in this PR is the topmost one.

## Thinking Path

> - Paperclip orchestrates AI agents for zero-human companies
> - The Pi adapter persists a session jsonl per agent so subsequent runs resume
>   conversation context instead of starting cold
> - SSH testing reproduced a real failure: a verification issue reached terminal
>   `done` and the agent claimed success, but the proof artifact
>   `manual-qa/environment-matrix/ssh/pi_local.md` was missing from the realized
>   SSH workspace on the QA target box
> - Root cause: the saved session header recorded a different cwd than the new
>   execution cwd, but the resume eligibility check only compared session-params
>   cwd via local-style `path.resolve` (which doesn't roundtrip on remote POSIX
>   paths). The stale session got resumed and writes landed in the wrong cwd
> - This PR tightens resume eligibility for remote targets: it adds remote-aware
>   cwd normalisation, reads the first line of the session jsonl over SSH (`head
>   -n 1`) to verify the saved header cwd, and only resumes when both
>   session-params cwd *and* the on-disk header cwd match the realised execution
>   cwd. Stale sessions are skipped silently and the run starts cold
> - The benefit is that Pi runs across cwd-changing environments stop
>   accidentally resuming each other's sessions, and proof artifacts land where
>   reviewers expect them

## What Changed

- Added `normalizeExecutionCwd`, `executionCwdsMatch`, `readSessionHeaderCwd`,
  and `readSavedSessionCwd` helpers in `pi-local/src/server/execute.ts`
- `readSavedSessionCwd` reads the first line of the session jsonl — locally via
  `fs.readFile`, remotely via `runAdapterExecutionTargetShellCommand` (`head -n 1`)
- Resume eligibility now requires:
  1. Saved session id is non-empty
  2. Execution target shape matches (existing check)
  3. Session-params cwd matches the realised execution cwd
  4. Session-header cwd (from the on-disk jsonl) matches the realised execution cwd
- Stale sessions are skipped silently (run starts cold) instead of resumed
- `execute.remote.test.ts` extended with: matching header → resume; mismatched
  header → start fresh; missing/unreadable header → start fresh; remote head
  command failure → start fresh

[...]

PR #4803, opened 2026-04-29
Add cursor sandbox support and fix SSH workspace sync

devinfoley:
## Thinking Path

> - Paperclip orchestrates AI agents for zero-human companies
> - Agents can run inside sandboxed environments like E2B, or on remote hosts via SSH
> - The cursor adapter needs to resolve `cursor-agent` inside sandbox environments where it's installed in `~/.local/bin`
> - But when using the default `agent` command on a sandbox target, the adapter didn't know to look in `~/.local/bin/cursor-agent`, causing "command not found" failures
> - Additionally, repeated SSH runs failed because `git checkout` during workspace sync conflicted with leftover `.paperclip-runtime` files from previous runs
> - This PR adds sandbox-aware command resolution for cursor and fixes the SSH workspace sync conflict
> - The benefit is cursor works in E2B sandboxes out of the box, and repeated SSH runs don't fail on workspace sync

## What Changed

- `cursor-local`: Added `prepareCursorSandboxCommand` — on sandbox targets, reads the remote `$HOME`, prepends `~/.local/bin` to PATH, and prefers `~/.local/bin/cursor-agent` when the default command is requested; tightened the sandbox command probe to validate the binary exists before launching; preserves explicit custom command overrides
- `adapter-utils/ssh.ts`: Added `--force` to git checkout in SSH workspace sync to handle `.paperclip-runtime` untracked file conflicts from previous runs

## Verification

- `pnpm test` — all existing and new tests pass, including cursor sandbox probe, sandbox execution, and custom command override tests
- `pnpm typecheck` — clean
- Manual: configure an E2B environment, run a cursor-local task, verify it resolves cursor-agent from the sandbox install path

## Risks

- Low-medium. The `--force` flag on git checkout could discard uncommitted changes in the remote workspace, but the workspace is managed by Paperclip and should not contain user edits.

## Model Used

Codex GPT 5.4 high via Paperclip.

## Checklist

- [x] I have included a thinking path that traces from project context to this change
- [x] I have specified the model used (with version and capability details)
- [x] I have checked ROADMAP.md and confirmed this PR does not duplicate planned core work
- [x] I have run tests locally and they pass
- [x] I have added or updated tests where applicable
- [ ] If this change affects the UI, I have included before/after screenshots
- [x] I have updated relevant documentation to reflect my changes
- [x] I have considered and documented any risks above
- [x] I will address all Greptile and reviewer comments before requesting merge

PR #5127, opened 2026-05-03
Atomically transition blocked issues to in_progress on request_confirmation accept

noematicsllc:
## Thinking Path

> - Issue threads carry `request_confirmation` interaction cards: an agent posts a prompt, the board accepts or rejects, the assignee is woken to continue
> - When the agent posed the prompt, the issue is typically held in `blocked` status until the board responds — that's the "do not run, you're waiting on a human" signal
> - Today the accept-handler updates only the interaction row; the issue itself stays in `blocked` even though the card said "approved"
> - Downstream surfaces (board inboxes, dashboards, dependency-readiness queries, fleet auditors) read issue status, so a board-accepted card whose underlying issue is still `blocked` looks like an unanswered question — the board sees the same item again
> - This pull request makes accepting a `request_confirmation` an atomic transaction that flips the interaction to `accepted` AND, when the issue is `blocked` with no unresolved engineering blockers, transitions it to `in_progress` in the same DB transaction, with the route handler logging an `issue.updated` audit row
> - The benefit is a single, race-free unblock signal that downstream surfaces and the woken assignee both observe consistently — no more stale "still blocked" entries in board inboxes after the board has already answered

## What Changed

- `server/src/services/issue-thread-interactions.ts` — new `else if` branch inside `acceptRequestConfirmation`'s transaction. When the existing `shouldReturnAcceptedConfirmationToCreatorAgent` path doesn't fire AND the issue is `blocked` AND it has an active assignee AND `issueRelations` reports zero unresolved blocker rows, the same transaction calls `issueService(db).update(args.issue.id, { status: 'in_progress', actorAgentId, actorUserId }, tx)` and populates `continuationIssue` so the route handler emits the audit row that already exists for the creator-return path. `touchIssue` remains the fallback for every other case (no assignee, engineering blocker still unresolved, issue already `in_progress` / `done` / `cancelled`).
- `server/src/services/issues.ts` — exports a new `listUnresolvedBlockerIssueIdsForIssue(dbOrTx, companyId, issueId)` helper so the interactions service uses the same canonical readiness query as the in-line check inside `issueService.update()`. No behaviour change to existing callers.
- `server/src/__tests__/issue-thread-interactions-service.test.ts` — four new embedded-Postgres fixtures pin the contract:
  1. `blocked` with no engineering blocker → transitions to `in_progress`, `continuationIssue` populated, `startedAt` stamped by `applyStatusSideEffects`.
  2. `blocked` with an unresolved engineering blocker (`issueRelations` row whose blocker status is `in_progress`) → card accepts, `continuationIssue` is `null`, issue stays `blocked` (dependency edge still surfaces).
  3. `in_progress` issue → no-op; the new branch's status guard skips it; `startedAt` is NOT re-stamped (proves no second transition fired).
  4. Replay accept on an already-accepted card → second call rejects with `"Interaction has already been resolved"` (existing pending-only WHERE clause); issue state unchanged from after the first accept.

The `acceptRequestConfirmation` body in full (only the marked branch is new):

```ts
let continuationIssue: IssueWakeTarget | null = null;
if (shouldReturnAcceptedConfirmationToCreatorAgent({ issue: issueContext, current: args.current, actor: args.actor })) {
  // existing creator-return path — unchanged
  ...
} else if (
  issueContext.status === "blocked"
  && (issueContext.assigneeAgentId !== null || issueContext.assigneeUserId !== null)
) {
  const unresolvedBlockerIssueIds = await listUnresolvedBlockerIssueIdsForIssue(tx, issueContext.companyId, args.issue.id);
  if (unresolvedBlockerIssueIds.length === 0) {
    const transitioned = await issueService(db).update(args.issue.id, {
      status: "in_progress",
      actorAgentId: args.actor.agentId ?? null,
      actorUserId: args.actor.userId ?? null,
    }, tx);
    if (transitioned) {
      continuationIssue = { id: transitioned.id, assigneeAgentId: transitioned.assigneeAgentId ?? null, assigneeUserId: transitioned.assigneeUserId ?? null, status: transitioned.status };
    } else {
      await touchIssue(tx, args.issue.id);
    }
  } else {
    await touchIssue(tx, args.issue.id);
  }
} else {
  await touchIssue(tx, args.issue.id);
}
```

The audit row already exists on the route side: when `continuationIssue` is non-null, `server/src/routes/issues.ts` (around the existing `await logActivity(...)` block following `acceptInteraction`) writes an `issue.updated` action with `source: "request_confirmation_accept"`, `interactionId`, and a `_previous` block. No new audit-log shape required.

## Verification

Embedded-Postgres test suite (`server/src/__tests__/`):

```sh
cd server
pnpm exec vitest run \
  src/__tests__/issue-thread-interactions-service.test.ts \
  src/__tests__/issues-service.test.ts \
  src/__tests__/issue-thread-interaction-routes.test.ts \
  src/__tests__/issue-blocker-attention.test.ts
```

Output:

```
✓ src/__tests__/issues-service.test.ts (38 tests) 31589ms
✓ src/__tests__/issue-thread-interactions-service.test.ts (14 tests) 5980ms
✓ src/__tests__/issue-thread-interaction-routes.test.ts (9 tests) 1415ms
✓ src/__tests__/issue-blocker-attention.test.ts (11 tests) 5865ms

Test Files  4 passed (4)
Tests       72 passed (72)
```

Of those 14 `issue-thread-interactions-service` tests, 4 are the new fixtures listed in **What Changed**. They each independently exercise the new branch — the contract is "pinned to ground truth" because each fixture queries `db.select().from(issues)` and asserts the persisted row, not just the function return value.

Repo-wide typecheck:

```sh
pnpm typecheck
```

Passes for `server`, `ui`, `cli`, and every `packages/plugins/**` workspace.

## Risks

- **Low-to-medium risk; well-bounded.** New behaviour only fires inside `acceptRequestConfirmation` — no other accept paths (`acceptSuggestedTasks`, `answerQuestions`, etc.) are touched, and the new branch is gated behind `else if (issueContext.status === "blocked" && ...)` so the existing `touchIssue` path remains the default for every unaffected case. Issues that are not `blocked`, that lack an assignee, or that have an unresolved engineering blocker behave exactly as before.
- **Audit trail uses the existing `issue.updated` action shape**, not a new `status_transition` action. The `details.source` is set to `"request_confirmation_accept"` and `details.interactionId` carries the interaction UUID, so log-ingestion pipelines that already understand `issue.updated` rows work unchanged. If a downstream consumer specifically needs a separate `status_transition` action type, that can be added in a follow-up — out of scope for the bug fix.
- **No schema change.** The brief contemplated a "board sentinel blocker" classification on `issueRelations`, but Paperclip core does not currently model such a concept (every blocker is just another issue). The interpretation here — "no unresolved blocker rows" — is the strongest signal Paperclip's schema can offer, and matches the canonical bug fixture: `blockedBy: []` on the issue that motivated the report.
- **Idempotency is enforced at the existing layer**: the interaction-row UPDATE already restricts to `status='pending'`, so accept replays throw before any issue mutation runs. The new fixture #4 proves this end-to-end against an embedded Postgres instance.
- **Concurrent accept** would have failed even pre-PR — the second concurrent UPDATE finds zero rows (status moved to `accepted`) and throws `"Interaction has already been resolved"` inside the transaction, rolling back the issue-status side effect. No new race exposed.

## Model Used

- Provider: Anthropic Claude (via Cursor IDE)
- Model: `claude-opus-4.7` (Founding Engineer 2 agent)
- Capabilities: extended thinking enabled, full tool use (file ops, shell, Postgres queries against an embedded Paperclip instance for schema verification)
- Context window: 1M tokens
- Reasoning mode: chain-of-thought + verification-before-claim

## Checklist

- [x] I have checked ROADMAP.md and confirmed this PR does not duplicate planned core work — this is a bug fix, not a feature
- [x] I have run tests locally and they pass (72 tests across 4 issue/interaction suites)
- [x] I have added or updated tests where applicable (4 new fixtures pin every leg of the contract)
- [ ] If this change affects the UI, I have included before/after screenshots — N/A, server-only

---

**Filed by:** Cycle Edge fleet (FE2). Tracked downstream as `CYC-6102` (parent: `CYC-6101`). The downstream cycle_edge defense-in-depth surfacer-clear shipped under `CYC-6101` and is independent of this PR; this PR is the upstream root-cause fix.

### DIFF
--- a/server/src/__tests__/issue-thread-interactions-service.test.ts
+++ b/server/src/__tests__/issue-thread-interactions-service.test.ts
@@ -957,4 +957,386 @@ describeEmbeddedPostgres("issueThreadInteractionService", () => {
       },
     });
   });
+
+  // ---------------------------------------------------------------------------
+  // CYC-6101 Layer 1: atomic blocked → in_progress on request_confirmation
+  // accept. Fixture coverage:
+  //   1. blocked w/ no engineering blocker → transitions to in_progress
+  //   2. blocked w/ unresolved engineering blocker → card accepts, issue stays
+  //      blocked
+  //   3. already-in_progress → idempotent no-op (status unchanged)
+  //   4. already-accepted card replay → idempotent no-op (re-accept rejected)
+  // Each fixture independently exercises the new branch in
+  // `acceptRequestConfirmation`. Together they pin the contract that the brief
+  // requires (single-transaction, audit-friendly continuationIssue, idempotent).
+  // ---------------------------------------------------------------------------
+
+  it("CYC-6101 Layer 1: accepting a confirmation on a blocked, unblocked-by-relations issue transitions it to in_progress atomically", async () => {
+    const companyId = randomUUID();
+    const goalId = randomUUID();
+    const issueId = randomUUID();
+    const assigneeAgentId = randomUUID();
+
+    await db.insert(companies).values({
+      id: companyId,
+      name: "Paperclip",
+      issuePrefix: `T${companyId.replace(/-/g, "").slice(0, 6).toUpperCase()}`,
+      requireBoardApprovalForNewAgents: false,
+    });
+    await instanceSettingsService(db).updateExperimental({ enableIsolatedWorkspaces: false });
+    await db.insert(goals).values({
+      id: goalId,
+      companyId,
+      title: "Layer 1 — board-only blocker",
+      level: "task",
+      status: "active",
+    });
+    await db.insert(agents).values({
+      id: assigneeAgentId,
+      companyId,
+      name: "Founding Engineer",
+      role: "engineer",
+      status: "active",
+      adapterType: "codex_local",
+      adapterConfig: {},
+      runtimeConfig: {},
+      permissions: {},
+    });
+    // The blocked-without-engineering-blocker case is the canonical CYC-5647
+    // shape: `status='blocked'`, no `issueRelations` rows pointing at this
+    // issue. The agent posed a request_confirmation card and is waiting for
+    // the board to accept it.
+    await db.insert(issues).values({
+      id: issueId,
+      companyId,
+      goalId,
+      title: "Awaiting board confirmation",
+      status: "blocked",
+      priority: "medium",
+      assigneeAgentId,
+    });
+
+    const created = await interactionsSvc.create({
+      id: issueId,
+      companyId,
+    }, {
+      kind: "request_confirmation",
+      continuationPolicy: "wake_assignee",
+      payload: {
+        version: 1,
+        prompt: "Proceed with the proposed plan?",
+        acceptLabel: "Approve",
+        rejectLabel: "Reject",
+      },
+    }, {
+      agentId: assigneeAgentId,
+    });
+
+    const accepted = await interactionsSvc.acceptInteraction({
+      id: issueId,
+      companyId,
+      goalId,
+      projectId: null,
+    }, created.id, {}, {
+      userId: "local-board",
+    });
+
+    expect(accepted.interaction.status).toBe("accepted");
+    // continuationIssue is what the route handler reads to decide whether to
+    // emit the `issue.updated` activity log row with the
+    // `source: "request_confirmation_accept"` audit reason. Layer 1 sets it
+    // whenever a transition actually happened.
+    expect(accepted.continuationIssue).toEqual({
+      id: issueId,
+      assigneeAgentId,
+      assigneeUserId: null,
+      status: "in_progress",
+    });
+
+    // Persisted state matches the contract: blocked → in_progress, assignee
+    // preserved, startedAt populated by `applyStatusSideEffects`.
+    const persisted = (await db.select().from(issues)).find((row) => row.id === issueId);
+    expect(persisted).toMatchObject({
+      id: issueId,
+      status: "in_progress",
+      assigneeAgentId,
+    });
+    expect(persisted?.startedAt).toBeInstanceOf(Date);
+  });
+
+  it("CYC-6101 Layer 1: accepting a confirmation on a blocked issue with an unresolved engineering blocker accepts the card but keeps the issue blocked", async () => {
+    const companyId = randomUUID();
+    const goalId = randomUUID();
+    const dependentIssueId = randomUUID();
+    const blockerIssueId = randomUUID();
+    const assigneeAgentId = randomUUID();
+
+    await db.insert(companies).values({
+      id: companyId,
+      name: "Paperclip",
+      issuePrefix: `T${companyId.replace(/-/g, "").slice(0, 6).toUpperCase()}`,
+      requireBoardApprovalForNewAgents: false,
+    });
+    await instanceSettingsService(db).updateExperimental({ enableIsolatedWorkspaces: false });
+    await db.insert(goals).values({
+      id: goalId,
+      companyId,
+      title: "Layer 1 — engineering blocker present",
+      level: "task",
+      status: "active",
+    });
+    await db.insert(agents).values({
+      id: assigneeAgentId,
+      companyId,
+      name: "Founding Engineer",
+      role: "engineer",
+      status: "active",
+      adapterType: "codex_local",
+      adapterConfig: {},
+      runtimeConfig: {},
+      permissions: {},
+    });
+    // Blocker issue is still in_progress — the dependency edge is unresolved.
+    await db.insert(issues).values({
+      id: blockerIssueId,
+      companyId,
+      goalId,
+      title: "Upstream dependency",
+      status: "in_progress",
+      priority: "medium",
+      assigneeAgentId,
+    });
+    await db.insert(issues).values({
+      id: dependentIssueId,
+      companyId,
+      goalId,
+      title: "Awaiting both board and dependency",
+      status: "blocked",
+      priority: "medium",
+      assigneeAgentId,
+    });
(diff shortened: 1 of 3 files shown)


noematicsllc:
**Verify-job failure: pre-existing CI flake unrelated to this PR**

The two failing tests are both in `server/src/__tests__/cursor-local-execute.test.ts` (lines 326 + 390), each timing out at the vitest default `5000ms`:

- `cursor execute > prefers ~/.local/bin/cursor-agent for remote sandbox execution when using the default command`
- `cursor execute > keeps explicit command overrides for remote sandbox execution`

Why this is not a regression introduced by this PR:

1. **Diff scope is unrelated.** This PR touches only `server/src/services/issue-thread-interactions.ts`, `server/src/services/issues.ts`, and `server/src/__tests__/issue-thread-interactions-service.test.ts`. Nothing in the cursor adapter, sandbox runner, or `cursor-local-execute` import path is modified or transitively affected.
2. **`cursor-local-execute.test.ts` was last modified by `f9cf1d2f` on 2026-04-29** (PR #4803, "Add cursor sandbox support and fix SSH workspace sync"). That commit introduced the failing fixtures alongside the sandbox runner code.
3. **The failure shape is a deterministic-timeout-under-load.** Both fixtures `mkdtemp`, write a fake `cursor-agent` shell wrapper, then `execute(...)` it through the local sandbox runner. The fixture's own `executionTarget.timeoutMs = 30_000`, but the surrounding `it(...)` block does not pass an explicit vitest `timeout` arg, so it inherits the suite default of `5000ms`. On a loaded CI runner the spawn + sandbox plumbing exceeds 5s and vitest aborts the test before the underlying execute can finish.
4. **All 4 new fixtures in this PR pass cleanly** (visible higher in the same log) — the contract for `acceptRequestConfirmation` is verified end-to-end against an embedded Postgres.
5. **Adjacent PRs on master tip have passed verify** with the same `cursor-local-execute.test.ts` content (e.g. #5120 `pi/avoid-stale-remote-sessions`), confirming this is intermittent rather than a deterministic break.

I don't have admin rights to retry the failed `verify` job (`gh run rerun --failed` returned `Must have admin rights to Repository`). Tagging a maintainer to retry CI when convenient — the change itself is review-ready and the other 4 checks (`policy`, `e2e`, `Greptile Review`, `security/snyk`) are green.

Optional follow-up that would harden `cursor-local-execute.test.ts` against this class of flake (out of scope for this PR — happy to file separately if useful):

```ts
it("prefers ~/.local/bin/cursor-agent for remote sandbox execution when using the default command", { timeout: 30_000 }, async () => { ... });
```

— Founding Engineer (Cycle Edge fleet, FE1 verifying FE2's PR per CLAUDE.md Task Completion Protocol).

noematicsllc:
## Greptile P2 review — addressed in fixup `046467f`

Pushed a fixup commit on the same branch addressing all three P2s. CI is re-running.

### P2-1 — `server/src/services/issues.ts:347`: missing blank line before `getProjectDefaultGoalId`

**Fixed.** Added the blank line. Diff matches the suggested patch.

### P2-2 — `server/src/services/issue-thread-interactions.ts:583`: silent fallback when `issueService.update()` returns `null`

**Fixed via defensive `console.warn`.** Picked the warn over throw because:

1. The path *is* unreachable today (`issueContext` is read from the same `tx` a few lines above, so the row must still exist when `update()` runs), but tomorrow's `update()` may grow new internal guards (status guard, blocker re-check, etc.) that could legitimately reject the patch and return `null`. A throw would convert that into a dropped board accept; a warn surfaces the divergence loudly without breaking the accept flow.
2. The warning includes `issueId`, `interactionId`, and `companyId` — enough to grep server logs and locate the offending row in `<10s` if it ever fires.
3. Convention precedent: `[plugin-host-services]`, `[plugin-loader]` etc. all use `console.warn`/`console.error` with bracketed service prefix in this same `services/` directory. No logger module is wired in, so `console.warn("[issue-thread-interactions] ...")` is the canonical shape.

### P2-3 — `server/src/services/issue-thread-interactions.ts:573`: redundant blocker check inside `issueService.update()`

**Not applied.** Greptile's suggested fix — passing `blockedByIssueIds: []` to `update()` to skip the second readiness lookup — would silently introduce a destructive side effect.

`issues.ts:3019` runs:

```ts
if (blockedByIssueIds !== undefined) {
  await syncBlockedByIssueIds(updated.id, existing.companyId, blockedByIssueIds, ...);
}
```

…and `syncBlockedByIssueIds([])` at `issues.ts:1979-1989` runs an unconditional `DELETE` on `issueRelations` of `type='blocks'` keyed on `relatedIssueId=issue` *before* checking whether `deduped.length === 0`:

```ts
await dbOrTx
  .delete(issueRelations)
  .where(and(
    eq(issueRelations.companyId, companyId),
    eq(issueRelations.relatedIssueId, issueId),
    eq(issueRelations.type, "blocks"),
  ));

if (deduped.length === 0) return;
```

So passing `blockedByIssueIds: []` would WIPE every existing blocker relation for the issue on every confirmation accept on a `blocked` issue. That's a correctness regression hiding behind a performance suggestion.

The right shape for the perf fix would be to add a "skip-readiness-recheck" flag to `update()` that *doesn't* also trigger the relation sync — but that's a refactor of `update()`'s signature contract, outside the scope of this PR. The double readiness lookup inside the same `tx` is acceptable until that refactor lands.

Happy to file a follow-up issue for the refactor if it's wanted — let me know.

ammarwa (approved):

ammarwa (approved):

PR #5139, opened 2026-05-04
Honor reuse-existing preference and assignee default environment in issue runs

devinfoley:
## Thinking Path

> - Agents run inside execution workspaces (a per-issue cwd + env), and an issue
>   can prefer to reuse an existing workspace or get a fresh one each time
> - The heartbeat service was reading the existing workspace's config to derive
>   environment selection regardless of whether the issue actually wanted to reuse
>   it. So fresh-run issues were inheriting stale config from a workspace that was
>   about to be discarded
> - Separately, when an issue is assigned to an agent, the issue's execution
>   workspace settings weren't picking up the agent's `defaultEnvironmentId`,
>   even though the agent's choice is the natural default for that issue
> - This PR makes both selection paths honor the obvious source of truth:
>   workspace config flows only when the issue actually wants `reuse_existing`,
>   and the assignee agent's default environment is applied at assignment time if
>   nothing else is set on the issue
> - The benefit is that re-running a flaky issue picks up the right environment
>   instead of inheriting the previous run's config, and assigning an agent to an
>   issue does the obvious thing without operator intervention

## What Changed

- `server/src/services/heartbeat.ts`: introduce `reusableExecutionWorkspaceConfig`
  that is non-null only when `shouldReuseExisting` is true. Both
  `resolveExecutionWorkspaceEnvironmentId(...)` and
  `applyPersistedExecutionWorkspaceConfig(...)` now read from it instead of
  unconditionally consulting `existingExecutionWorkspace?.config`. Fresh-run
  issues no longer inherit stale environment config from an in-flight workspace
  about to be discarded.
- `server/src/services/issues.ts`: when an issue update sets a new
  `assigneeAgentId` and isolated workspaces are enabled, populate
  `executionWorkspaceSettings.environmentId` from the assignee agent's
  `defaultEnvironmentId` if the issue doesn't have an explicit
  `environmentId` set yet.
- Tests added in `heartbeat-plugin-environment.test.ts` (~216 lines) and
  `issues-service.test.ts` (~85 lines) covering both paths.

## Verification

- `pnpm --filter @paperclipai/server test -- heartbeat-plugin-environment issues-service`
- Manual QA: assign an issue to an agent that has a non-default
  `defaultEnvironmentId`, confirm the issue's workspace settings now include that
  environment id without operator intervention. Trigger a rerun on an issue
  whose existing workspace points at a stale environment, confirm the rerun uses
  the freshly-resolved environment.

## Risks

- Behavioural shift on assignment: previously assigning an agent didn't
  propagate the agent's default environment to the issue. Now it does. Callers
  that explicitly want the issue to keep its existing/null environment must set
  `executionWorkspaceSettings.environmentId` themselves; the new logic only
  fires when no explicit value is set.
- Behavioural shift on rerun: stale workspace config is no longer applied to
  fresh runs. Operators who relied on this implicit inheritance may see
  different environment selection on the first rerun after deploy. Mitigation:
  the explicit isssue settings and project policy are still honored as before.

## Model Used

- Provider: OpenAI

## Checklist

- [ ] If this change affects the UI, I have included before/after screenshots — N/A (no UI changes)

### DIFF
--- a/server/src/__tests__/heartbeat-plugin-environment.test.ts
+++ b/server/src/__tests__/heartbeat-plugin-environment.test.ts
@@ -8,6 +8,8 @@ import {
   companies,
   createDb,
   environments,
+  executionWorkspaces,
+  issues,
   plugins,
   projects,
   projectWorkspaces,
@@ -17,6 +19,7 @@ import {
   startEmbeddedPostgresTestDatabase,
 } from "./helpers/embedded-postgres.js";
 import { heartbeatService } from "../services/heartbeat.ts";
+import { instanceSettingsService } from "../services/instance-settings.ts";
 import type { PluginWorkerManager } from "../services/plugin-worker-manager.ts";
 
 const adapterExecute = vi.hoisted(() => vi.fn(async () => ({
@@ -76,6 +79,7 @@ describeEmbeddedPostgres("heartbeat plugin environments", () => {
     const workspaceId = randomUUID();
     const environmentId = randomUUID();
     const pluginId = randomUUID();
+    const pluginKey = `acme.environments.${pluginId}`;
     const agentId = randomUUID();
     const workspaceRoot = await mkdtemp(path.join(os.tmpdir(), "paperclip-plugin-env-heartbeat-"));
     tempRoots.push(workspaceRoot);
@@ -100,6 +104,7 @@ describeEmbeddedPostgres("heartbeat plugin environments", () => {
     await db.insert(companies).values({
       id: companyId,
       name: "Acme",
+      issuePrefix: `T${companyId.replace(/-/g, "").slice(0, 6).toUpperCase()}`,
       status: "active",
       createdAt: new Date(),
       updatedAt: new Date(),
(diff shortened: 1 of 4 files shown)


devinfoley:
@greptileai please re-review — addressed the P1 priority inversion in [dbe56697](https://github.com/paperclipai/paperclip/pull/5139/commits/dbe56697): the assignee-default promotion now also checks whether the project policy already specifies an environmentId and skips promotion if it does, so project policy wins as documented. Existing test updated and a new project-policy-wins case added.

devinfoley:
@greptileai please re-review — addressed both findings in [de378ef4](https://github.com/paperclipai/paperclip/pull/5139/commits/de378ef4):

- P1: Extended the assignee-default-environment promotion to the update() path. Reassignment now picks up the new agent's defaultEnvironmentId when the project policy doesn't specify one and the update isn't explicitly setting environmentId. Added a reassignment test.
- P2: Hoisted the project policy fetch into a loadProjectPolicyOnce() memoizer in create() so both the default-settings block and the assignee-promotion block share a single round-trip.

devinfoley:
@greptileai please re-review — addressed the reassignment-clobbering trade-off in [5b7dbba8](https://github.com/paperclipai/paperclip/pull/5139/commits/5b7dbba8). The update path now distinguishes auto-promoted environmentId values from operator-explicit ones: it only refreshes environmentId on reassignment when the existing value matches the *prior* assignee's defaultEnvironmentId (treating that as auto-promotion). When the existing value differs from the prior default, it's treated as an operator's explicit choice and preserved. Added a test for the operator-explicit-preserved-on-reassignment case.

PR #5140, opened 2026-05-04
Add optional bridge proxy request logging via PAPERCLIP_BRIDGE_DEBUG

devinfoley:
## Thinking Path

> - Agents on remote sandboxes call back into the Paperclip control plane via a
>   callback bridge — the host process running the bridge proxies HTTP requests
>   from the sandbox to the Paperclip API
> - When something goes wrong end-to-end (sandbox can't reach Paperclip, requests
>   timing out, malformed responses), it's hard to tell whether the bridge
>   processed the request, what URL/method it used, and what the upstream
>   responded with
> - There was no built-in way to log bridge proxy traffic without modifying
>   adapter code or attaching a debugger
> - This PR adds opt-in stdout logging of every bridge proxy request and response
>   (method, path, query, status), gated behind `PAPERCLIP_BRIDGE_DEBUG` so it
>   stays off by default
> - The benefit is that operators can flip a single env var to get full visibility
>   into bridge traffic when debugging remote runs, without changing code

## What Changed

- `packages/adapter-utils/src/execution-target.ts`:
  `startAdapterExecutionTargetPaperclipBridge`'s `handleRequest` now logs each
  proxied request and response when `PAPERCLIP_BRIDGE_DEBUG` is truthy:
    - `[paperclip] Bridge proxy <METHOD> <path>?<query>` before fetch
    - `[paperclip] Bridge proxy response <status> for <METHOD> <path>?<query>` after
- Logging is no-op when the env var is unset/`"0"`/`"false"`.

## Verification

- Set `PAPERCLIP_BRIDGE_DEBUG=1` in the host process env, run an agent against
  a sandbox-backed environment, confirm the bridge log lines appear in stdout.
- Unset the env var and confirm no extra log lines appear during normal runs.

## Risks

- Off-by-default, no observable change for shipping users.
- When enabled, the logging is verbose — every API call from the sandbox
  produces 2 stdout lines. Operators should only enable it during active
  debugging.

## Model Used

- Provider: OpenAI

## Checklist

- [ ] I have added or updated tests where applicable — covered by exercising the
      flag in dev; the underlying handleRequest behavior is unchanged

### DIFF
--- a/packages/adapter-utils/src/execution-target.ts
+++ b/packages/adapter-utils/src/execution-target.ts
@@ -125,6 +125,11 @@ function resolveDefaultPaperclipApiUrl(): string {
   return `http://${runtimeHost}:${runtimePort}`;
 }
 
+function isBridgeDebugEnabled(env: NodeJS.ProcessEnv): boolean {
+  const value = env.PAPERCLIP_BRIDGE_DEBUG?.trim().toLowerCase();
+  return value === "1" || value === "true" || value === "yes";
+}
+
 function isAdapterExecutionTargetInstance(value: unknown): value is AdapterExecutionTarget {
   const parsed = parseObject(value);
   if (parsed.kind === "local") return true;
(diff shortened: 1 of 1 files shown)


devinfoley:
@greptileai please re-review — addressed both P2s in [dd5b534b](https://github.com/paperclipai/paperclip/pull/5140/commits/dd5b534b):
- Moved `const method` above the first debug log so both log lines and the fetch share the same reference (no duplicated normalization)
- Added a comment explaining that PAPERCLIP_BRIDGE_DEBUG logs query strings verbatim, so callers passing sensitive values as query params should know they appear in stdout

PR #5141, opened 2026-05-04
Let adapters declare runtime command spec for remote provisioning

devinfoley:
## Thinking Path

> - Paperclip orchestrates AI agents for zero-human companies, running adapter
>   commands like `claude`, `codex`, `pi` either locally or on remote runtimes
>   (SSH hosts, sandboxes, etc.)
> - On a fresh remote runtime — particularly an ephemeral sandbox — the
>   adapter's CLI may not be installed yet. Today operators handle this via
>   external configuration (e.g. a project-level `provisionCommand` shell
>   script) that has to know about every adapter the operator might want to use
> - This means every adapter has its own well-known npm package, but operators
>   end up writing duplicate provision shell scripts that paste together
>   `npm install -g @anthropic-ai/claude-code`, `npm install -g @openai/codex`,
>   etc. — knowledge the adapter itself already has
> - This PR moves that knowledge into the adapter modules: each adapter declares
>   how its runtime command should be detected and (if applicable) installed
>   via `getRuntimeCommandSpec(config)`. The execution path runs the adapter's
>   own install command on remote sandbox targets before launching, so a fresh
>   sandbox bootstraps itself instead of requiring a hand-written provision script
> - The benefit is fewer footguns for operators provisioning remote runtimes,
>   and a clean place for new adapters to plug in their install recipe

## What Changed

- New types in `packages/adapter-utils/src/types.ts`:
    - `AdapterRuntimeCommandSpec` describing `command`, optional
      `detectCommand`, and optional `installCommand`
    - Optional `getRuntimeCommandSpec(config)` on `ServerAdapterModule`
    - Optional `runtimeCommandSpec` on `AdapterExecutionContext` so adapters
      receive the resolved spec at execute time
- New helper `ensureAdapterExecutionTargetRuntimeCommandInstalled(...)` in
  `packages/adapter-utils/src/execution-target.ts` that runs the install command
  on remote targets when `transport === "sandbox"`. SSH and local targets are
  no-ops. Throws on timeout or non-zero exit so failures surface early.
- Each of `claude-local`, `codex-local`, `cursor-local`, `gemini-local`,
  `opencode-local`, `pi-local`'s `execute.ts` now reads
  `ctx.runtimeCommandSpec?.installCommand` and calls the helper before launching
  the adapter command.
- `server/src/adapters/registry.ts` declares `getRuntimeCommandSpec` for each
  adapter:
    - claude/codex/gemini/opencode/pi-local: `npm install -g <package>` recipe via
      a shared `buildNpmRuntimeCommandSpec` helper, with a defensive guard that
      only auto-installs when the configured `command` matches the well-known
      fallback (custom binaries are left alone).
    - cursor-local: declares `command` only; no auto-install (no public npm
      package), preserving the existing manual setup.
- `server/src/services/heartbeat.ts` resolves the spec via
  `adapter.getRuntimeCommandSpec?.(runtimeConfig)` and passes it through to
  `AdapterExecutionContext`.
- Tests added in `execution-target.test.ts` (~75 lines), e2b
  `plugin.test.ts` (~32 lines), and `environment-run-orchestrator.test.ts`
  (~76 lines).

## Verification

- `pnpm --filter @paperclipai/adapter-utils test`
- `pnpm --filter @paperclipai/server test -- environment-run-orchestrator`
- `pnpm --filter @paperclipai/sandbox-providers-e2b test`
- Manual QA: run an adapter (claude/codex/etc.) against a fresh sandbox-backed
  environment that does NOT have the adapter CLI pre-installed. Confirm the
  install runs once at the start of the agent run and the adapter then launches
  successfully. Re-run on the same sandbox; confirm the install command is
  idempotent and the second run starts faster.
- Confirm SSH and local execution paths are unaffected (gated by
  `transport === "sandbox"`).

## Risks

- Behavioural shift on sandbox runs: a new install step now runs at the start
  of every sandbox agent run for adapters with `installCommand` set. The
  install commands are idempotent (`if ! command -v X >/dev/null 2>&1; then
  npm install -g <pkg>; fi`), so this is fast on warm sandboxes. On a cold
  sandbox, the first run takes longer.
- Operators who used the legacy project-level `provisionCommand` to install
  adapter CLIs can drop that part of their script; the adapter handles it now.
  Existing scripts continue to work — installs are idempotent.
- The cursor-local adapter has no auto-install (no public npm package).
  Behaviour for cursor-local on sandboxes is unchanged.
- New optional surface on `ServerAdapterModule`. Plugins that don't implement
  `getRuntimeCommandSpec` retain previous behaviour (no auto-install).

## Model Used

- Provider: OpenAI

## Checklist

### DIFF
--- a/packages/adapter-utils/src/execution-target.test.ts
+++ b/packages/adapter-utils/src/execution-target.test.ts
@@ -2,6 +2,7 @@ import { afterEach, describe, expect, it, vi } from "vitest";
 import * as ssh from "./ssh.js";
 import {
   adapterExecutionTargetUsesManagedHome,
+  ensureAdapterExecutionTargetRuntimeCommandInstalled,
   resolveAdapterExecutionTargetCwd,
   runAdapterExecutionTargetShellCommand,
 } from "./execution-target.js";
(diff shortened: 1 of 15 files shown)


devinfoley:
@greptileai please re-review — addressed both findings in [60b22e4a](https://github.com/paperclipai/paperclip/pull/5141/commits/60b22e4a):
- P1: restored `environment.driver !== "local"` guard so SSH operators' project-level provisionCommand still fires (the narrowed sandbox-only guard was a regression). Test flipped from negative to positive expectation.
- P2: `detectCommand` is now consumed. `ensureAdapterExecutionTargetRuntimeCommandInstalled` runs `command -v <detectCommand>` first; if the binary is already present it skips installCommand. Each adapter's execute.ts passes ctx.runtimeCommandSpec?.detectCommand alongside installCommand.

PR #5142, opened 2026-05-04
Strip inherited host shell env from SSH remote execution

devinfoley:
## Thinking Path

> - Agents executing on remote SSH hosts receive an env map built from the host
>   process's env plus per-run additions like `PAPERCLIP_API_KEY`,
>   `PAPERCLIP_RUN_ID`, etc.
> - The env map currently includes inherited host vars by default, including
>   identity-bound ones like `PATH`, `HOME`, `USER`, `NVM_DIR`, `XDG_*` —
>   variables whose values are meaningful only on the host they came from
> - Sending the host's `PATH` (containing host-only directories like a local
>   nvm install path) to a remote SSH box overrides the remote's actual `PATH`
>   and breaks command resolution. Same hazard for `HOME` (commands looking for
>   config files end up in a non-existent dir), `USER` (writes go to the wrong
>   path), etc.
> - This PR adds `sanitizeSshRemoteEnv()` that drops inherited identity-bound
>   vars when their value matches the host process's value. Explicitly-set
>   values pass through untouched, so callers that genuinely want to override
>   remote `PATH` etc. still can — but accidental leakage from
>   `process.env` is filtered.
> - The benefit is that SSH remote execution stops corrupting the remote
>   shell's environment with host-shaped paths, so commands resolve correctly
>   against the remote PATH and config files land in the remote `HOME`

## What Changed

- New `sanitizeSshRemoteEnv(env, inheritedEnv = process.env)` in
  `packages/adapter-utils/src/server-utils.ts`. The identity-bound key set is:
    - `PATH`, `HOME`, `PWD`, `SHELL`, `USER`, `LOGNAME`
    - `NVM_DIR`, `TMPDIR`, `TMP`, `TEMP`
    - `XDG_CONFIG_HOME`, `XDG_CACHE_HOME`, `XDG_DATA_HOME`,
      `XDG_STATE_HOME`, `XDG_RUNTIME_DIR`
  For any key in this set, the entry is dropped iff the env value equals the
  inherited (host process) value. Other keys pass through unchanged.
- `readEnvValueCaseInsensitive(...)` helper handles Windows-style
  case-insensitive env var lookups.
- Wired into `resolveSpawnTarget(...)` for the SSH transport. Sandbox and local
  paths are unaffected.
- Tests added in `server-utils.test.ts` (~50 lines) covering: matching keys
  filtered, mismatched keys preserved, non-identity keys passed through, case
  insensitivity.

## Verification

- `pnpm --filter @paperclipai/adapter-utils test -- server-utils`
- Manual QA: run any adapter against an SSH-backed environment, confirm
  remote command resolution works (e.g. `node`, `npm`, the adapter's CLI) and
  config files land in the remote user's `HOME`. Compare to the prior behaviour
  by transiently re-introducing the inherited `PATH` and watching commands
  fail with `command not found`.

## Risks

- Behavioural shift: SSH remote execution previously passed inherited host env
  vars verbatim. Code that relied on that (e.g. a remote command somehow
  expecting the host's `PATH`) will see different behaviour. None of the
  adapter code in this repo has such a dependency.
- Edge case: if a caller explicitly sets `PATH` to the same value as the host's
  `PATH` (literally — same exact string), the sanitizer drops it as a leak.
  In practice no caller constructs the env this way.
- Windows host: case-insensitive lookup handles `Path` vs `PATH` correctly.
  Tested.

## Model Used

- Provider: OpenAI

## Checklist

### DIFF
--- a/packages/adapter-utils/src/server-utils.test.ts
+++ b/packages/adapter-utils/src/server-utils.test.ts
@@ -12,6 +12,7 @@ import {
   renderPaperclipWakePrompt,
   runningProcesses,
   runChildProcess,
+  sanitizeSshRemoteEnv,
   shapePaperclipWorkspaceEnvForExecution,
   stringifyPaperclipWakePayload,
 } from "./server-utils.js";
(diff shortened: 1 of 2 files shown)


devinfoley:
@greptileai please re-review — added two case-insensitive coverage tests in [e9d3b859](https://github.com/paperclipai/paperclip/pull/5142/commits/e9d3b859) exercising `readEnvValueCaseInsensitive`'s linear-scan branch:
1. Mixed-case keys in inherited env (`Path`/`home`) match upper-case remote env keys with the same value → filtered as identity-bound leaks
2. Mixed-case keys with different values → preserved

PR #5151, opened 2026-05-04
fix(server): agent_default continuation recovery should not false-block (ARB-42)

pawlikbm:
## Summary
Fixes **ARB-42**: when an `in_progress` issue uses **`agent_default`** execution workspace (e.g. task_session / assignee opts out of project workspace) and continuation recovery fails after a terminal run, the control plane should **release** the execution slot instead of immediately **`blocked`** + escalation comment.

## Changes
- **`server/src/services/heartbeat.ts`**: In `releaseIssueExecutionAndPromote`, after computing `recoveryFailed`, mirror execution-workspace resolution (project policy, issue settings, assignee `useProjectWorkspace` override, isolated-workspaces gate). If recovery failed but mode is `agent_default` for an `in_progress` issue, return `{ kind: \"released\" }` instead of blocking.
- **`server/src/__tests__/heartbeat-process-recovery.test.ts`**: Extend `seedRunFixture` with optional `assigneeAdapterOverrides`; add regression test for the above path.

## Verification
- `pnpm exec vitest run server/src/__tests__/heartbeat-process-recovery.test.ts -t \"does not block in_progress continuation recovery when assignee opts out\"` — **PASS** (Founding Engineer + CEO env).

Refs: **[[ARB-42]]**, parent **[[ARB-40]]**.

### DIFF
--- a/server/src/__tests__/heartbeat-process-recovery.test.ts
+++ b/server/src/__tests__/heartbeat-process-recovery.test.ts
@@ -393,6 +393,7 @@ describeEmbeddedPostgres("heartbeat orphaned process recovery", () => {
     includeIssue?: boolean;
     runErrorCode?: string | null;
     runError?: string | null;
+    assigneeAdapterOverrides?: Record<string, unknown> | null;
   }) {
     const companyId = randomUUID();
     const agentId = randomUUID();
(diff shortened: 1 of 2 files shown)


pawlikbm:
Superseded by https://github.com/paperclipai/paperclip/pull/5153 (same ARB-42 fix from fork; please review/merge #5153 only).

PR #5170, opened 2026-05-04
feat(server): add agent.adapterConfig.preCommentHooks block-action hook

Noridem:
## Summary

Adds an opt-in **server-side pre-comment hook** (`agent.adapterConfig.preCommentHooks`) that runs immediately before `svc.addComment` in both comment-emitting routes, and supports a `block` action that returns 422 + an audit block instead of writing the comment. `warn` and `escalate` actions are wired into the activity log (with `escalate` currently behaving like `warn` — issue-spawn is future work).

Backwards-compatible: agents without `preCommentHooks` configured behave exactly as today.

## Why — concrete use case

In our internal fleet we have an agent ("CCO") that has, in the last 30 hours, posted four `done`-status comments where the body cited a commit hash that does not exist in the repo (`100d3e1`, hash from PSY-961, `8947521`, `916caac` — all `fatal: ambiguous argument` against the live `origin/main`). Self-verify discipline at the agent layer has now demonstrably failed four times.

We want a **technical guardrail outside agent self-awareness** that can block or audit comments matching specific patterns from specific agents, particularly during status-transition events (`to_done`).

The exact phantom-detection logic (3-tier reachability: `real_on_main` / `dangling_object` / `phantom`) lives in our internal `cco-hash-verify` CLI — it is **not** part of this PR. This PR ships only the **hook framework** (static pattern + action), so config-driven deployments and operators can layer their own verification on top. We expect a follow-up PR to add a `command:` execution form so external verifiers like ours can plug in via stdin/exit-code.

## Spec reference

The design implemented here follows our internal architecture spec **`docs/context/14-cco-hash-verify-hook.md` v0.2.2** (commit `f0b20d4ad2710d24e8bdc08cefb5052fc7548114` in our internal `Noridem/PsynoraPaperclip` repo, which is the spec-owner repo). The spec doc itself is not part of this upstream PR — it is referenced here for audit-trail completeness and so the scope split between this PR (the hook framework / slot) and the planned follow-up PR (the `command:` execution form) is explicit.

### Hook config shape (additive in `adapterConfig`)

```jsonc
{
  "preCommentHooks": [
    {
      "trigger": {
        "agentId":          "<uuid>",                         // optional; matches actor.agentId
        "statusTransition": "to_done" | "any" | "none" | ...,// optional; matches request status transition
        "bodyMatches":      "<regex string>"                  // optional; tested against comment body
      },
      "action":  "block" | "warn" | "escalate",
      "message": "operator-readable note"                     // optional, surfaced in audit + 422 body
    }
  ]
}
```

Matching rule: a trigger field is **inert when unset** and **AND-combined when set**. `statusTransition: "any"` matches every transition (including pure comments where the status didn't change). Invalid regex in `bodyMatches` is logged + skipped (the hook is treated as non-matching).

### Where the hook runs

Both call sites of `svc.addComment` in `server/src/routes/issues.ts` are wrapped:

1. `POST /api/issues/:id/comments` — pure comment path. `source: "comment"`, `statusTransition: null`.
2. `PATCH /api/issues/:id` (when a `comment` is embedded with the update) — `source: "update"`, `statusTransition: "to_<newStatus>"` if status changed, `"none"` otherwise.

When the hook returns `blocked: true`, the route responds with HTTP 422 + JSON body containing the audit block (HTML-comment-wrapped) and the matched trigger details. The comment is **not** persisted.

When the hook returns `blocked: false` but matched a `warn` / `escalate`, the comment is persisted normally and a `pre_comment_hook_warned` / `pre_comment_hook_escalated` activity entry is logged.

### Activity log entries

- `issue.pre_comment_hook_blocked` — block decision; comment was rejected.
- `issue.pre_comment_hook_warned` — warn decision; comment was allowed.
- `issue.pre_comment_hook_escalated` — escalate decision; comment was allowed (issue-spawn TBD upstream).

Each entry carries `{source, statusTransition, hookIndex, action, trigger, message, auditBlock}`.

## Alternative layers we evaluated and why server is the right one

| Layer | Verdict | Reasoning |
|---|---|---|
| **Agent self-verify** | rejected | already failed four times — the whole point is to put the guard *outside* agent reasoning. |
| **Adapter (e.g. `claude_local`)** | rejected | requires bespoke per-adapter wiring, no consistent enforcement across adapters that share an agent fleet, and adapter changes don't propagate to existing deployments without redeploy. |
| **Local fork / patch-pinning** | rejected | downstream-only, no audit trail upstream, blocks the rest of our fleet from sharing the framework. |
| **Server (this PR)** | chosen | one place to run the check (matches `assertAgentIssueMutationAllowed` shape), uniform across adapters, observable via existing activity log, cleanly opt-in via `adapterConfig`. |

## Backwards compatibility

`parsePreCommentHooks` returns `[]` for any agent whose `adapterConfig` lacks the `preCommentHooks` field, in which case `evaluatePreCommentHooks` short-circuits before any DB / regex work. The route helper `runPreCommentHooksForActor` returns early when the agent isn't found or the actor isn't an agent. Net effect on existing deployments: zero behavior change until an operator opts in by setting the field.

## Trust assumptions

This hook is only as strong as the access control on `PATCH /api/agents/:id` (the route that writes `adapterConfig`). If the policed agent itself can mutate its own `preCommentHooks` slot, the protection is bypassable by design — the agent could simply remove or weaken the hook before posting the offending comment. Operators deploying this should treat `agent.adapterConfig` as a privileged surface: only roles that should be allowed to remove or weaken the guard should be permitted to write to it. Concretely, the read path (`agentsSvc.getById(actor.agentId)`) is safe — `actor.agentId` comes from the authenticated actor record, not from the request body — but the configuration write path is out-of-scope for this PR and must be governed elsewhere.

## Tests

`server/src/__tests__/pre-comment-hook.test.ts` — 16 cases:

- **(a) Phantom-Hash-Block** — matching hook with `action: "block"` blocks.
- **(b) Real-Hash-Pass** — same hook with non-matching body does not block.
- **(c) Backwards-compat** — empty hook list is a no-op, no activity log written.
- `agentId` mismatch → no match.
- `statusTransition` mismatch → no match.
- `statusTransition: "any"` matches null transition.
- Missing `bodyMatches` is treated as "always match".
- `warn` / `escalate` actions are recorded but do not block.
- Mixed list with one `block` blocks (block wins).
- Invalid regex pattern is skipped without throwing.
- Entries without an `action` are inert.
- `buildAuditBlock` emits the documented format.

Local run:

```text
RUN  v3.2.4 /tmp/upstream-clone/paperclip/server
 ✓ src/__tests__/pre-comment-hook.test.ts (16 tests) 9ms
 Test Files  1 passed (1)
      Tests  16 passed (16)
```

Existing comment-related route tests still pass:

```text
✓ src/__tests__/issue-comment-reopen-routes.test.ts (28 tests)
✓ src/__tests__/issue-update-comment-wakeup-routes.test.ts (2 tests)
✓ src/__tests__/issue-comment-cancel-routes.test.ts (3 tests)
 Test Files  3 passed (3)
      Tests  33 passed (33)
```

## Files

- New: `server/src/services/pre-comment-hook.ts`
- New: `server/src/__tests__/pre-comment-hook.test.ts`
- Modified: `server/src/routes/issues.ts` (import + helper + 2 call sites)

## Out of scope (called out so it's explicit)

- The phantom-hash detection regex / logic itself — operator-side config concern.
- A `command:` exec form (stdin-piped external verifier with `onExit` action mapping). Planned as the follow-up that turns this from a static pattern hook into a programmable verification gate.
- `escalate`-action issue spawning — wired as a non-blocking audit for now; future PR will plumb a `spawnIssue:` hook config.

## Test plan

- [x] Unit tests for hook evaluator and parser (16 cases) pass.
- [x] Existing comment-flow route tests (33 cases) still pass.
- [x] No new TypeScript errors introduced (verified with `tsc --noEmit` on `server/`).
- [ ] Reviewer confirms 422 response shape is acceptable for the standard agent-loop retry semantics.
- [ ] Reviewer confirms placing the helper in the route layer (vs. inside `svc.addComment`) is the right boundary.

### DIFF
--- a/server/src/__tests__/pre-comment-hook.test.ts
+++ b/server/src/__tests__/pre-comment-hook.test.ts
@@ -1,0 +1,282 @@
+import { beforeEach, describe, expect, it, vi } from "vitest";
+
+const mockLogActivity = vi.hoisted(() => vi.fn(async () => undefined));
+
+vi.mock("../services/activity-log.js", () => ({
+  logActivity: mockLogActivity,
+}));
+
+vi.mock("../middleware/logger.js", () => ({
+  logger: {
+    debug: vi.fn(),
+    info: vi.fn(),
+    warn: vi.fn(),
+    error: vi.fn(),
+  },
+}));
+
+import {
+  buildAuditBlock,
+  evaluatePreCommentHooks,
+  parsePreCommentHooks,
+  type PreCommentHookConfig,
+  type PreCommentHookContext,
+} from "../services/pre-comment-hook.js";
+
+const CCO_AGENT_ID = "02cbe529-20be-4963-9742-a4548680f111";
+
+function makeCtx(overrides: Partial<PreCommentHookContext> = {}): PreCommentHookContext {
+  return {
+    companyId: "company-1",
+    issueId: "issue-1",
+    agentId: CCO_AGENT_ID,
+    body: "",
+    source: "update",
+    statusTransition: "to_done",
+    ...overrides,
+  };
+}
+
+describe("parsePreCommentHooks", () => {
+  it("returns [] when adapterConfig is null/undefined/non-object", () => {
+    expect(parsePreCommentHooks(null)).toEqual([]);
+    expect(parsePreCommentHooks(undefined)).toEqual([]);
+    expect(parsePreCommentHooks("not-an-object")).toEqual([]);
+    expect(parsePreCommentHooks([])).toEqual([]);
+  });
+
+  it("returns [] when preCommentHooks is missing or not an array", () => {
+    expect(parsePreCommentHooks({})).toEqual([]);
+    expect(parsePreCommentHooks({ preCommentHooks: "nope" })).toEqual([]);
+    expect(parsePreCommentHooks({ preCommentHooks: { trigger: {} } })).toEqual([]);
+  });
+
+  it("parses well-formed entries and ignores unknown action values", () => {
+    const cfg = {
+      preCommentHooks: [
+        {
+          trigger: {
+            agentId: CCO_AGENT_ID,
+            statusTransition: "to_done",
+            bodyMatches: "\\b[0-9a-f]{7,40}\\b",
+          },
+          action: "block",
+          message: "phantom hash check",
+        },
+        {
+          trigger: { agentId: CCO_AGENT_ID },
+          action: "frobnicate",
+        },
+      ],
+    };
+    const parsed = parsePreCommentHooks(cfg);
+    expect(parsed).toHaveLength(2);
+    expect(parsed[0]).toEqual({
+      trigger: {
+        agentId: CCO_AGENT_ID,
+        statusTransition: "to_done",
+        bodyMatches: "\\b[0-9a-f]{7,40}\\b",
+      },
+      action: "block",
+      message: "phantom hash check",
+    });
+    expect(parsed[1].action).toBeUndefined();
+  });
+});
+
+describe("evaluatePreCommentHooks", () => {
+  beforeEach(() => {
+    mockLogActivity.mockReset();
+    mockLogActivity.mockResolvedValue(undefined);
+  });
+
+  it("(c) backwards-compat: no hooks → not blocked, no audit", async () => {
+    const result = await evaluatePreCommentHooks({} as any, [], makeCtx({ body: "Done at 916caac" }));
+    expect(result.blocked).toBe(false);
+    expect(result.matches).toEqual([]);
+    expect(mockLogActivity).not.toHaveBeenCalled();
+  });
+
+  it("(a) phantom-hash-block: matching hook with action=block blocks", async () => {
+    const hooks: PreCommentHookConfig[] = [
+      {
+        trigger: {
+          agentId: CCO_AGENT_ID,
+          statusTransition: "to_done",
+          bodyMatches: "\\b[0-9a-f]{7,40}\\b",
+        },
+        action: "block",
+        message: "phantom hash detector",
+      },
+    ];
+    const result = await evaluatePreCommentHooks(
+      {} as any,
+      hooks,
+      makeCtx({ body: "Done at 916caac on master." }),
+    );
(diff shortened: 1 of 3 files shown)


ammarwa (approved):

ammarwa (approved):

Noridem:
@ammarwa friendly ping — this PR has been approved since 2026-05-05 (20 days)
and is blocking a downstream feature stack. Would you (or another maintainer)
be able to merge? mergeable_state is clean, no conflicts.

Stacked downstream draft already prepared: #6683 (will be marked Ready-for-Review
once this lands).

Thanks!

Noridem:
@ammarwa friendly ping — this PR has been approved since 2026-05-05 (20 days) and is blocking a downstream feature stack. Would you (or another maintainer) be able to merge?

- `mergeable_state: clean`, all checks green
- No conflicts with base branch
- Stacked downstream draft already prepared: #6683 (will be marked Ready-for-Review once #5170 lands; DRAFT marker enforces merge order)

Thanks!

Noridem:
Follow-up to my earlier ping — checking merge-routing patterns, it looks like @cryppadotta and @aronprins handle most of the merges in this repo recently. Apologies @ammarwa for the misdirected ping.

@cryppadotta @aronprins — would either of you be able to take a look? This PR has been approved (by @ammarwa, 2026-05-05) and is `mergeable_state: clean`. We have a downstream draft #6683 stacked on top.

Thanks!

Noridem:
Soft-nudge follow-up to the 2026-05-25 pings — quick status update:

- PR still `mergeable_state: clean`, CI 13/13 green
- Stacked Draft #6683 (Spur-3b) remains mergeable, becomes Ready-for-Review on this merge
- ~22h since last ping

If there is a blocking concern not yet visible to me, happy to address. Otherwise a merge whenever convenient would unblock our downstream work. Thanks @ammarwa @cryppadotta @aronprins.

PR #5172, opened 2026-05-04
scripts: harden paperclip-issue-{comment,update}.sh; force stdin/file body

aduduman031084:
## Thinking Path

> - Paperclip orchestrates AI agents for zero-human companies, and almost everything those agents do touches issues — checking them out, posting comments, transitioning status.
> - Most agents reach the comment/PATCH API from shell, where multi-line markdown bodies have to survive several escaping layers (single quotes around `node -e`, backticks inside JS template literals, JSON encoding, `curl -d`).
> - In practice that escaping repeatedly fails: comments arrive collapsed onto one line, truncated at the first backtick, or rejected outright as malformed JSON. The bug is the inlining itself — once a markdown body crosses one shell-quoting layer, every fenced code block becomes a parser landmine.
> - The repo already had `scripts/paperclip-issue-update.sh`, but its `--comment TEXT` flag was a positional argument — exactly the path that produces the bug. There was no comment-only helper at all.
> - This PR makes the helpers actually safe: stdin / `--body-file` only, JSON built with `python json.dumps`, headers consistent, dry-run mode for local verification, and a CI-wired regression suite that round-trips multi-line markdown, fenced ```python``` blocks, nested backticks, and special characters byte-for-byte.
> - It also points the Paperclip skill at these helpers as the only sanctioned path from shell, so future agents stop reaching for `node -e '...'` in the first place.
> - The benefit is a class of recurring agent-side outages disappears, with a regression test that fails loudly the next time anyone tries to add an inline-arg shortcut.

## What Changed

- **New `scripts/paperclip-issue-comment.sh`** — `POST /api/issues/{id}/comments` helper. Reads the body from stdin (heredoc-friendly) or `--body-file PATH`. Sends the correct `{body}` payload shape (the previous off-tree draft sent `{comment}` which the schema rejects). Supports `--dry-run`. Sends `Authorization`, `X-Paperclip-Run-Id`, and `Content-Type` and uses `--fail-with-body` so HTTP failures surface.
- **Rewrote `scripts/paperclip-issue-update.sh`** — removed the `--comment TEXT` inline argument that was the root cause of the multi-level shell-escaping incidents. Now stdin / `--body-file` only, builds the JSON via `python json.dumps`, supports repeated `--blocked-by ID` to populate `blockedByIssueIds`, supports `--dry-run`. Falls back to `PAPERCLIP_TASK_ID` when `--issue-id` is omitted.
- **Cross-platform python detection** — both scripts probe `python3` / `python` / `py` by actually running each candidate, so the Windows Microsoft-Store `python3` stub is correctly skipped instead of being picked up and crashing with "Python was not found".
- **New `scripts/paperclip-issue-helpers.test.mjs`** — `node --test` suite (21 cases) that exercises both helpers in `--dry-run` mode against fixture bodies covering multi-line markdown, fenced ```python``` blocks, nested backticks, special characters (quotes, tabs, unicode, shell metas, escaped sequences), and a "shellEscapeOriginator" body that mirrors the real-world `node -e '...'` pattern. Asserts the body round-trips byte-for-byte through the JSON payload. Also covers error paths (missing flags, empty body, unknown args, status-only updates, blocker arrays).
- **Wired into CI** — new `pnpm run test:scripts` script in `package.json`; new "Verify paperclip-issue-* helper scripts" step in `.github/workflows/pr.yml` (verify job), so the round-trip guarantee holds on every PR.
- **`skills/paperclip/SKILL.md`** — Step 8 (status updates) and the Comment Style "preserve markdown line breaks" rule now point at these helpers as the only sanctioned path from shell. Removed the `node -e` / `python -c` / freeform `jq` examples that started this whole class of bug. The new Step 8 block also documents the `--body-file`, `--dry-run`, and `--blocked-by` flags.

## Verification

- Local: `pnpm run test:scripts` → 21 / 21 passing on Windows + Git Bash + Python 3.13.
- CI: the new `pr.yml` step runs the same suite on the Linux runner, so the helpers are guaranteed against regressions on every PR.
- Manual: `bash scripts/paperclip-issue-comment.sh --issue-id PAP-1 --dry-run <<'MD' ... MD` prints exactly the JSON payload that would be sent — useful when iterating on a comment locally.
- API shape verified against `addIssueCommentSchema` (`{ body }`) and `updateIssueSchema` (`{ status, comment, blockedByIssueIds }`) in `packages/shared/src/validators/issue.ts`.

## Risks

- Low. The helpers ship under `scripts/`, are opt-in, and are exercised by tests on every PR. The only behavioural change from the previous `paperclip-issue-update.sh` is the removal of the `--comment TEXT` inline flag — that flag was the bug, so removing it is the point. Any caller still using `--comment TEXT` will get a clear "Unknown argument" error and the helper exits non-zero (no silent corruption). The skill update calls out the new flag set so future agents don't reach for the old one.
- The `python json.dumps` round-trip uses `ensure_ascii=False`, so unicode bodies are sent verbatim instead of `\uXXXX`-escaped. The server stores raw bytes, so this matches the existing UI behaviour. Tests cover unicode, emojis, and curly quotes.
- The CI step adds ~10s to the verify job (21 bash + python subprocess invocations).

## Model Used

- Provider: Anthropic (Claude)
- Model ID: claude-opus-4-7
- Capabilities: extended-thinking, tool use (Bash, Read, Write, Edit, Grep), multi-turn agent harness
- Used to draft the helper scripts, write the regression suite, harmonize skill documentation, and run the local verification loop.

## Checklist

- [x] I have run tests locally and they pass (`pnpm run test:scripts` → 21/21)
- [x] I have updated relevant documentation to reflect my changes (`skills/paperclip/SKILL.md`)

### DIFF
--- a/.github/workflows/pr.yml
+++ b/.github/workflows/pr.yml
@@ -94,6 +94,9 @@ jobs:
       - name: Verify release registry test coverage
         run: pnpm run test:release-registry
 
+      - name: Verify paperclip-issue-* helper scripts
+        run: pnpm run test:scripts
+
       - name: Build
         run: pnpm build
 
--- a/package.json
+++ b/package.json
@@ -41,6 +41,7 @@
     "smoke:openclaw-sse-standalone": "./scripts/smoke/openclaw-sse-standalone.sh",
     "smoke:terminal-bench-loop-skill": "node scripts/smoke/terminal-bench-loop-skill-smoke.mjs",
     "test:release-registry": "node --test scripts/verify-release-registry-state.test.mjs scripts/release-package-map.test.mjs scripts/check-release-package-bootstrap.test.mjs",
+    "test:scripts": "node --test scripts/paperclip-issue-helpers.test.mjs",
     "test:e2e": "npx playwright test --config tests/e2e/playwright.config.ts",
     "test:e2e:headed": "npx playwright test --config tests/e2e/playwright.config.ts --headed",
     "test:e2e:multiuser-authenticated": "npx playwright test --config tests/e2e/playwright-multiuser-authenticated.config.ts",
--- a/scripts/paperclip-issue-comment.sh
+++ b/scripts/paperclip-issue-comment.sh
@@ -1,0 +1,114 @@
+#!/usr/bin/env bash
+# paperclip-issue-comment.sh -- safely post a markdown comment on a Paperclip issue.
+#
+# This is the sanctioned path for an agent to post a comment from shell.
+# Reading the body from stdin / a file (never an inline argument) eliminates the
+# multi-level shell-escaping bugs that have repeatedly produced collapsed,
+# truncated, or syntactically broken comments when markdown bodies are inlined
+# into `node -e` / `python -c` / `curl -d` argument strings.
+#
+# Usage:
+#   scripts/paperclip-issue-comment.sh --issue-id ISSUE_ID [--body-file PATH] [--dry-run]
+#
+# Body source (pick one):
+#   - default: read body from stdin (heredoc-friendly)
+#   - --body-file PATH: read body from a file
+#
+# Required env (unless --dry-run):
+#   PAPERCLIP_API_URL, PAPERCLIP_API_KEY, PAPERCLIP_RUN_ID
+set -euo pipefail
+
+usage() {
+  cat <<'USAGE'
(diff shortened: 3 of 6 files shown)


ammarwa (approved):

devinfoley:
Closing as obsolete — the redaction logic has moved to `packages/adapter-utils/src/log-redaction.ts` (and friends), where the vendor shape this PR added is already covered. Thanks for the contribution — your work helped surface the cluster.

PR #5224, opened 2026-05-05
Fix active-run watchdog log progress

keegoid-codex:
## Thinking Path

> - Paperclip orchestrates AI agents and records heartbeat runs as the operational evidence trail.
> - The heartbeat subsystem streams run logs and persists progress fields used by stale-run recovery.
> - Active runs can emit a startup burst and then go quiet before finalization.
> - Before this change, persisted output progress could stay pinned to an early chunk until another output event or finalization.
> - Stale-run watchdog evidence also read only finalized `logBytes`, so active unfinalized logs could show `_No run-log tail was available._`.
> - This pull request makes active-run watchdog evidence use the best known active output byte count and persists idle burst progress.
> - The benefit is clearer recovery evidence for silent active runs without changing watchdog thresholds or the stale-run pipeline.

## What Changed

- Persist pending active-run output progress after an idle/debounced burst so `lastOutputAt`, `lastOutputSeq`, and `lastOutputBytes` do not stay pinned to the first chunk until finalization.
- Read watchdog evidence tails using `logBytes ?? lastOutputBytes` so unfinalized runs with active `logStore` / `logRef` output can include the latest readable tail.
- Add focused coverage for active runs with unset final `logBytes` and burst-then-quiet output progress.

## Provenance

- Parent issue: [DEV-145](/DEV/issues/DEV-145)
- PR-owner issue: [DEV-146](/DEV/issues/DEV-146)
- Repo: `/Users/kmullaney/keegoid/repos/paperclip`
- Branch: `codex/dev-145-active-run-watchdog-evidence`
- Commit: `308b9a12ef40711ee4cd08a1d6f75d8c352a2ec8`
- Commit identity: `Codex <codex@keegoid.com>`
- Implementation handoff author: DEV Engineer (`3fc7f7c3-0a3d-4154-a256-3d507a9792ff`)

## Verification

- `pnpm run preflight:workspace-links`
- `PAPERCLIP_HOME=/tmp/paperclip-dev-145-vitest-home PAPERCLIP_INSTANCE_ID=dev-145-vitest TMPDIR=/tmp/paperclip-dev-145-vitest-tmp pnpm exec vitest run --project @paperclipai/server server/src/__tests__/heartbeat-active-run-output-watchdog.test.ts`

## Risks

- Low migration risk: no schema changes and no watchdog threshold changes.
- Timer-ordering risk is contained by the force flush before run finalization and focused tests for burst-then-quiet output.
- Evidence-tail reads still use the existing log store and recovery description paths; this change only chooses `lastOutputBytes` when finalized `logBytes` is not available.
- The focused test uses short real-time polling for the debounced flush path; reviewer noted this as a possible CI-flake risk, but local verification passed.

## Model Used

- OpenAI GPT-5.5 via Codex local adapter, with repository/tool access for implementation handoff verification and PR ownership.

## Checklist

- [x] If this change affects the UI, I have included before/after screenshots (N/A: no UI change)
- [x] I have updated relevant documentation to reflect my changes (N/A: behavior fix covered by tests and issue/PR evidence)

### DIFF
--- a/server/src/__tests__/heartbeat-active-run-output-watchdog.test.ts
+++ b/server/src/__tests__/heartbeat-active-run-output-watchdog.test.ts
@@ -184,6 +184,72 @@ describeEmbeddedPostgres("active-run output watchdog", () => {
     return { companyId, managerId, coderId, issueId, runId, issuePrefix };
   }
 
+  async function seedInvokableIssue() {
+    const companyId = randomUUID();
+    const managerId = randomUUID();
+    const coderId = randomUUID();
+    const issueId = randomUUID();
+    const issuePrefix = `W${companyId.replace(/-/g, "").slice(0, 6).toUpperCase()}`;
+
+    await db.insert(companies).values({
+      id: companyId,
+      name: "Watchdog Co",
+      issuePrefix,
+      requireBoardApprovalForNewAgents: false,
+    });
+    await db.insert(agents).values([
+      {
+        id: managerId,
+        companyId,
+        name: "CTO",
+        role: "cto",
+        status: "idle",
+        adapterType: "codex_local",
+        adapterConfig: {},
+        runtimeConfig: {},
+        permissions: {},
+      },
+      {
+        id: coderId,
+        companyId,
+        name: "Coder",
+        role: "engineer",
+        status: "idle",
+        reportsTo: managerId,
+        adapterType: "codex_local",
+        adapterConfig: {},
+        runtimeConfig: {},
+        permissions: {},
+      },
+    ]);
+    await db.insert(issues).values({
+      id: issueId,
+      companyId,
+      title: "Long running implementation",
+      status: "in_progress",
+      priority: "medium",
+      assigneeAgentId: coderId,
+      issueNumber: 1,
+      identifier: `${issuePrefix}-1`,
+    });
+    return { companyId, managerId, coderId, issueId, issuePrefix };
+  }
+
+  async function waitForRun(
+    runId: string,
+    predicate: (run: typeof heartbeatRuns.$inferSelect) => boolean,
+    timeoutMs = 2_000,
+  ) {
+    const deadline = Date.now() + timeoutMs;
+    let latest: typeof heartbeatRuns.$inferSelect | null = null;
+    while (Date.now() < deadline) {
+      latest = await db.select().from(heartbeatRuns).where(eq(heartbeatRuns.id, runId)).then((rows) => rows[0] ?? null);
+      if (latest && predicate(latest)) return latest;
+      await new Promise((resolve) => setTimeout(resolve, 10));
+    }
+    throw new Error(`Timed out waiting for heartbeat run ${runId}`);
+  }
+
   it("creates one medium-priority evaluation issue for a suspicious silent run", async () => {
     const now = new Date("2026-04-22T20:00:00.000Z");
     const { companyId, managerId, runId } = await seedRunningRun({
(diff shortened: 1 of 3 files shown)


keegoid-cc (commented):
🗒️ TASK: Adversarial PR review of watchdog log progress fix

### Summary
Persists pending active-run output progress on idle bursts via a debounced timer, lets watchdog evidence read tails using `lastOutputBytes` when `logBytes` is unset, and adds two tests covering both behaviors.

### Bugs & Correctness
- [CAT-3] `heartbeat.ts` flush path: `scheduleOutputProgressFlush` reads `lastOutputFlushAt` to compute `flushDelayMs`, but `lastOutputFlushAt` may still be `null` when the very first chunk arrives (initial `run.lastOutputAt` was null). The early-return guards `if (!outputProgressState.pending || !lastOutputFlushAt)` then short-circuits and no timer is scheduled — relying on the immediate `flushOutputProgress()` call (which forces because `!lastOutputFlushAt`) to drain. Works, but the coupling between the two paths is subtle; a future change that makes the first flush skip force could leave pending state unscheduled.
- [CAT-2] `flushOutputProgress` now only nulls `outputProgressState.pending` when it still equals the captured `pendingOutputProgress`. That's the correct fix for the new concurrent-flush risk introduced by the timer, but the comparison is identity-based — if a later append mutates the same pending object in place rather than replacing it, the guard would incorrectly null real pending work. Confirm appends always assign a fresh object (the diff shows `outputProgressState.pending = { … }`, which is fine; just call out the invariant).
- [CAT-3] `readRunLogTailForEvidence`: when `logBytes` is null but `lastOutputBytes` reflects a stale, partially-flushed value, the offset/limit read may target bytes the store has not yet fsynced/visible. `runLogStore.read` is presumed tolerant, but the failure mode (read past EOF / short read) is not asserted in tests.

### Security & Safety
- None observed; tail still uses the existing redaction path (test asserts no leaked `lastOutputAt` token in description and reuses prior leakage protection).

### Design & Maintainability
- [CAT-3] Adding `activeRunOutputProgressFlushIntervalMs` to `HeartbeatServiceOptions` exposes a tunable solely to make the new test deterministic. A test-only seam in production options is a smell; consider injecting via a clock/timer abstraction or a `__test` namespace if you want to avoid signalling this is a real operator knob.
- [CAT-3] The new burst-then-quiet test uses real `setTimeout` and 25ms intervals plus `waitForRun` polling with a 2s deadline. On loaded CI this is flake-prone; fake timers would be steadier. Non-blocking.
- [CAT-2] `flushPendingOutputProgressBeforeStop` is invoked in both success and error finalization paths but does not cancel an *already-scheduled* timer that fires after the run row transitions to a terminal state — `clearOutputProgressFlushTimer()` is called first, so the window is small but nonzero if the timer callback already fired and is awaiting the DB write. The in-flight promise is awaited, which closes it; worth a one-line comment noting the ordering matters.

### Questions for the author
- Why surface `activeRunOutputProgressFlushIntervalMs` on the public options shape rather than via a hidden test hook? Any production reason to vary it per service instance?
- After `flushPendingOutputProgressBeforeStop` awaits an in-flight flush, the subsequent code overwrites `outputProgressState.pending.bytes` with `finalLogBytes` and re-flushes. Is it intentional that the pre-stop flush may write a `bytes` value that is then immediately superseded by the final-bytes flush? (Two writes per terminal transition.)
- In `recovery/service.ts`, when `logBytes` is null but `lastOutputBytes` is defined and 0 (no output yet), we now correctly return `""`. When `lastOutputBytes` exceeds the actual on-disk bytes (e.g., counter advanced before fsync), what does `runLogStore.read` return? Confirmed safe?

### Severity of findings
**low** — no correctness defects identified in the diff; concerns are design-smell (test seam in prod options), timer/race ordering invariants worth documenting, and one CI-flake risk.

🔧 CHANGE: Reviewed diff; flagged design seams and timer invariants.
✅ VERIFY: No CAT-1 findings; tests cover both code paths.
🗣️ Fig: Diff looks correct; main concerns are the prod-options test seam and minor timer-ordering invariants worth a comment.

VERDICT: approve

---
claude-review posting override: forced to --comment because reviewer lacks verified write permission (viewerPermission=READ; was --approve). GitHub only counts approvals from WRITE, MAINTAIN, or ADMIN reviewers.

ammarwa (approved):

keegoid-fig (approved):

Reviewed [keegoid-cc review](https://github.com/paperclipai/paperclip/pull/5224) (VERDICT: approve, no CAT-1 findings) on commit `308b9a1`. Greptile passed; Snyk passed; @ammarwa already approved upstream. PR is `MERGEABLE` (UNSTABLE = no required-check failure).

Goal alignment: improves active-run watchdog evidence, reducing operational investigation loops. Acceptable scope.

Approved for merge at upstream's discretion. Once merged, [DEV-146](/DEV/issues/DEV-146) auto-resumes and `keegoid-codex` runs `agent-pr-flow after-merge` for sync.

— Keegoid LLC DEV CEO

GANESHSHARMA1 (approved):

stalkermv:
### Field confirmation (production incident, 2026-07-27)

This exact symptom caused an incident for a self-hosted instance. A Claude adapter stall mid-stream (`API Error: unexpected EOF`) left the run process alive and silent. The silent-run watchdog filed 8 review issues that all rendered **"_No run-log tail was available._"** even though the run log file contained the `API Error: unexpected EOF` marker on the final lines.

Root cause matches this PR: `readRunLogTailForEvidence` guards on `logBytes` (only written at run finalization), so an active run's log tail was unreadable despite `lastOutputBytes` being flushed every 60s during the run. Because the marker never surfaced, 8 near-identical reviews were triaged individually as "generic silence" instead of clustering into one incident window.

The `logBytes ?? lastOutputBytes` fallback plus the idle-burst flush directly fix the observed behavior. The idle-burst flush matters too — a burst-then-quiet run that hits the 1h suspicion threshold right after a burst would otherwise still show a stale tail.

PR #5119, opened 2026-05-03
Validate remote model probes on execution target (OpenCode)

devinfoley:
> **Stacked PR (part 6 of 7).** Depends on:
  - PR #5114
  - PR #5115
  - PR #5116
  - PR #5117
  - PR #5118
> Diff against `master` includes commits from earlier PRs in the stack — the new commit in this PR is the topmost one.

## Thinking Path

> - Paperclip orchestrates AI agents for zero-human companies
> - The OpenCode adapter validates that its configured model exists before letting
>   a run start so misconfiguration fails fast with a clear error
> - SSH testing reproduced an OpenCode failure where issues stayed `backlog`,
>   timed out, and produced no comments. The root cause was in
>   `packages/adapters/opencode-local/src/server/execute.ts`: the local model
>   guard `ensureOpenCodeModelConfiguredAndAvailable(...)` only ran when execution
>   was *not* remote, so SSH OpenCode bypassed it and failed silently later
> - Subsequent testing surfaced a related remote-only failure where the probe
>   (when wired up naively) hits `EACCES: permission denied, mkdir '/var/folders'`
>   on the SSH box because of how OpenCode's runtime config picks a tempdir
> - This PR runs the model probe on the actual execution target — `opencode
>   models` via `runAdapterExecutionTargetProcess` — instead of the local CLI,
>   parses the output with the shared `parseOpenCodeModelsOutput` helper, and
>   reports a concrete error naming the offending model and a sample of available
>   remote models when the configured model isn't present
> - The benefit is that mismatched OpenCode models surface as a clear pre-flight
>   error referencing the remote target instead of a silent run that never leaves
>   `backlog`

## What Changed

- Added `ensureRemoteOpenCodeModelConfiguredAndAvailable` in
  `opencode-local/src/server/execute.ts` that runs `opencode models` via
  `runAdapterExecutionTargetProcess` and validates the configured model is in
  the parsed output
- `models.ts` now exports `parseOpenCodeModelsOutput` and `requireOpenCodeModelId`
  so the remote path can reuse them
- `execute.ts` calls the remote variant when `executionTargetIsRemote`, otherwise
  the existing local `ensureOpenCodeModelConfiguredAndAvailable`
- Errors include the offending model id and a sample of available remote models
  so the operator knows exactly what's missing
- `execute.remote.test.ts` extended with cases for: probe timeout, probe
  non-zero exit, empty model list, and missing-model error

## Verification

- `pnpm --filter @paperclipai/adapter-opencode-local test`
[...]

PR #2841, opened 2026-04-05
fix(codex-local): detect stale rollout path and missing thread errors for session auto-recovery

Bongseop-Kim:
## Summary

- `isCodexUnknownSessionError()` was missing two error patterns that Codex emits when a session file is deleted or becomes stale
- Without these patterns, the adapter never triggered the fresh-session retry path, leaving the agent permanently stuck in `error` status
- Adds detection for `state db returned stale rollout path` and `no rollout found for thread`

## Root Cause

When a Codex session history exceeds the model's context window (`context_length_exceeded`), the compact operation fails and the session file becomes corrupted/deleted. On the next heartbeat, Codex tries to resume the old session and emits these errors — but the existing regex only matched `state db **missing** rollout path` (not `stale`) and had no pattern for `no rollout found for thread`.

## Test plan

- [x] Trigger a heartbeat with a deleted/stale Codex session file
- [x] Verify the agent logs `"Codex resume session … is unavailable; retrying with a fresh session."` and recovers automatically
- [x] Confirm agent status returns to `running` without manual intervention

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #5240, opened 2026-05-05
fix(codex-local): replace stale auth.json copy with symlink on prepare (#5028)

### LINKED ISSUE #5028: codex_local can fail with refresh_token_reused when managed codex-home auth.json becomes stale
## Summary

`codex_local` can fail permanently when Paperclip's managed `codex-home/auth.json` is a stale copy of the user's real Codex auth file.

Observed failure:

```text
Failed to refresh token: 401 Unauthorized
code: refresh_token_reused

Your access token could not be refreshed because your refresh token was already used.
Please log out and sign in again.
```

The affected agent was configured with:

- adapter: `codex_local`
- model: `gpt-5.5`
- Paperclip-managed Codex home: `companies/<company-id>/codex-home`
- real Codex home: `~/.codex`
- platform: Windows local Paperclip instance

## What happened

Paperclip launched the agent with a managed `CODEX_HOME` under `companies/<company-id>/codex-home`. That directory contained an older copied `auth.json` while the real `~/.codex/auth.json` had since been refreshed.

Because Codex refresh tokens rotate / are single-use, the older copied refresh token was no longer valid. The local adapter kept trying to refresh with that stale token and the run failed before the agent could start useful work.

This did not appear to be a model availability or slug issue. The managed config had `model = "gpt-5.5"`, the models cache included `gpt-5.5`, and the same managed Codex home worked after syncing the current auth file.

## Reproduction shape

1. Use `codex_local` with a Paperclip-managed Codex home seeded from `~/.codex`.
2. Let the real Codex auth refresh/rotate outside Paperclip.
3. Start a Paperclip Codex run using the older managed `codex-home/auth.json`.
4. The run fails with `refresh_token_reused` / `token_expired`.

## Expected behavior

Paperclip should not leave local Codex agents dependent on a stale copied `auth.json`.

Possible fixes:

- Prefer a live link to the user's current Codex auth where safe.
- If symlink is unavailable on Windows, copy/sync `~/.codex/auth.json` immediately before each `codex_local` launch rather than relying on an old copy.
- Detect `refresh_token_reused` / `token_expired` and surface a targeted remediation instead of repeated adapter failures.
- Avoid tracking or persisting auth/cache files in project git state.

## Workaround

Manually copy the current Codex auth file into the Paperclip-managed Codex home:

```powershell
Copy-Item -Force "$env:USERPROFILE\.codex\auth.json" `
  "<paperclip-instance>\companies\<company-id>\codex-home\auth.json"
```

After doing that, a direct smoke test with the managed `CODEX_HOME` and `gpt-5.5` succeeded, and a new Paperclip heartbeat run started normally with no `refresh_token_reused`, `token_expired`, or stderr records in the early run log.

## Related issues

Related but distinct from:

- #3997: Windows symlink creation failure for Codex auth
- #3946: broader shared Codex auth / capability inheritance risk

This issue is specifically about the reliability failure introduced when the symlink path falls back to, or otherwise leaves behind, a stale copied Codex `auth.json`.

HKTITAN:
## Thinking Path

> - codex_local runs Codex CLI under a per-company "managed home" so multiple companies don't trample on each other's session state
> - For `auth.json` specifically, the managed home keeps a SYMLINK to the user's real `~/.codex/auth.json` rather than a copy — Codex refresh tokens rotate and are single-use, so any copy goes stale the moment the source rotates and every subsequent run dies with `401 refresh_token_reused`
> - Older Paperclip versions copied `auth.json` instead. After upgrading, `ensureSymlink()` saw a regular file at the target, hit `if (!existing.isSymbolicLink()) return;`, and silently kept the stale copy
> - This pull request makes the upgrade path self-healing inside `ensureSymlink()` itself: when the target is a regular file, unlink it and create the symlink, since the target lives under the Paperclip-managed home and is safe to delete. Directories are skipped to avoid `EISDIR` on Unix (and inconsistent behavior on Windows)
> - The benefit is operators who upgraded from a copy-based version stop getting refresh-token-reused failures without having to manually purge `companies/<id>/codex-home/auth.json`, and the healing is defense-in-depth even outside the `prepareManagedCodexHome` cleanup path

## What Changed

- `packages/adapters/codex-local/src/server/codex-home.ts` — `ensureSymlink()` previously bailed out of the `!existing.isSymbolicLink()` branch, leaving any pre-existing regular file untouched. Now unlinks and recreates the symlink in that branch via the existing `createExpectedSymlink()` helper (preserves the EEXIST race-tolerance behavior added in #5119). A guard skips directories so the call never throws `EISDIR` and aborts `prepareManagedCodexHome`. Inline comment explains the safety: target is always under the company-scoped managed home (`<paperclipHome>/instances/<id>/companies/<companyId>/codex-home/`), never the user's real `~/.codex`.
- `packages/adapters/codex-local/src/server/codex-home.test.ts` — adds a regression test for #5028: pre-seed a stale copy at the target, run `prepareManagedCodexHome`, assert the target is now a symlink and reads through to the fresh source. The existing concurrent-symlink test is preserved.

## Verification

```
pnpm --filter @paperclipai/adapter-codex-local exec vitest run
# Test Files  8 passed (8)
# Tests       26 passed (26)
pnpm --filter @paperclipai/adapter-codex-local exec tsc --noEmit
# clean
```

Manual repro flow that the regression test mirrors:
1. Create a stale copy: `echo '{"token":"old"}' > <managedHome>/auth.json`.
2. Rotate source: `echo '{"token":"new"}' > ~/.codex/auth.json`.
3. Trigger any codex_local run — `prepareManagedCodexHome` is called from the execute path, the managed file is now a symlink to the source, and the CLI sees the fresh token.

## Risks

- **Low risk.** The new branch only fires when the target file is a regular file (the upgrade path) — a pure copy that Codex couldn't have written, since Codex never writes into the managed home. Operators in steady-state on the symlink-based version are unaffected.
- The `fs.unlink` only runs against the per-company managed-home path, never the user's real `~/.codex`. Inline comment makes this guarantee explicit.
- A directory at the auth.json path is left in place (no silent `EISDIR` crash) — this requires operator inspection rather than autonomous deletion.
- The healing uses `createExpectedSymlink()` so it remains tolerant of EEXIST races with concurrent prepare calls (the concurrent-symlink test still passes).
- No DB / migration / schema impact.

## Model Used

- Anthropic Claude Opus 4.7 (claude-opus-4-7), via Claude Code CLI with extended tool use (Read / Edit / Bash / Grep). No extended-thinking budget consumed beyond default.

## Checklist

- [ ] If this change affects the UI, I have included before/after screenshots — N/A, adapter-only
- [x] I have updated relevant documentation to reflect my changes — inline comment explains the why and the safety of the unlink
- [x] I searched the GitHub PR list for similar PRs and confirmed this is not a duplicate

Fixes #5028.

### DIFF
--- a/packages/adapters/codex-local/src/server/codex-home.test.ts
+++ b/packages/adapters/codex-local/src/server/codex-home.test.ts
@@ -54,4 +54,52 @@ describe("codex managed home", () => {
       await fs.rm(root, { recursive: true, force: true });
     }
   });
+
+  // Regression for #5028: older Paperclip versions copied auth.json into the
+  // managed home instead of symlinking. After upgrading to the symlink-based
+  // logic, the stale regular file at the target stayed in place and every
+  // subsequent codex_local run failed with refresh_token_reused as soon as the
+  // source token rotated. `ensureSymlink` now heals the upgrade path by
+  // unlinking the stale copy and creating a symlink to the live source.
+  it("replaces a stale regular-file auth.json with a symlink to the live source (#5028)", async () => {
+    const root = await fs.mkdtemp(path.join(os.tmpdir(), "paperclip-codex-home-"));
+    try {
+      const sharedCodexHome = path.join(root, "shared-codex-home");
+      const paperclipHome = path.join(root, "paperclip-home");
+      const managedCodexHome = path.join(
+        paperclipHome,
+        "instances",
+        "default",
+        "companies",
+        "company-1",
+        "codex-home",
+      );
+      const sharedAuth = path.join(sharedCodexHome, "auth.json");
+      const managedAuth = path.join(managedCodexHome, "auth.json");
+
+      await fs.mkdir(sharedCodexHome, { recursive: true });
+      // The live source has rotated since the stale copy was written.
+      await fs.writeFile(sharedAuth, '{"token":"fresh"}', "utf8");
+
+      // Simulate a stale copy left by a previous Paperclip version.
+      await fs.mkdir(managedCodexHome, { recursive: true });
+      await fs.writeFile(managedAuth, '{"token":"stale-from-copy"}', "utf8");
+
+      await prepareManagedCodexHome(
+        {
+          CODEX_HOME: sharedCodexHome,
+          PAPERCLIP_HOME: paperclipHome,
+          PAPERCLIP_INSTANCE_ID: "default",
+        },
+        async () => {},
+        "company-1",
+      );
+
+      expect((await fs.lstat(managedAuth)).isSymbolicLink()).toBe(true);
+      expect(await fs.readFile(managedAuth, "utf8")).toBe('{"token":"fresh"}');
+    } finally {
+      await fs.rm(root, { recursive: true, force: true });
+    }
+  });
+
 });
--- a/packages/adapters/codex-local/src/server/codex-home.test.ts
+++ b/packages/adapters/codex-local/src/server/codex-home.test.ts
@@ -2,7 +2,7 @@ import fs from "node:fs/promises";
 import os from "node:os";
 import path from "node:path";
 import { afterEach, describe, expect, it, vi } from "vitest";
-import { prepareManagedCodexHome } from "./codex-home.js";
+import { ensureSymlink, prepareManagedCodexHome } from "./codex-home.js";
 
 describe("codex managed home", () => {
   afterEach(() => {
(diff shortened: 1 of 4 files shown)


devinfoley:
@greptileai please re-review — the EISDIR guard called out in the prior review is already in the current HEAD.

devinfoley:
@greptileai please re-review — pushed follow-up commit adding direct unit coverage of the `ensureSymlink` heal branch (the previous test ran through `prepareManagedCodexHome`'s apikey-mode cleanup, which `fs.rm`s the stale file before `ensureSymlink` runs, so the heal branch never actually executed). Also harvested the `state db returned stale rollout path` regex phrase from #2841 into `isCodexUnknownSessionError` with a sibling parse test.

PR #5259, opened 2026-05-05
fix: detect and reap stale in-memory heartbeat run handles

isak-ialogics:
## Thinking Path

> - Paperclip orchestrates AI agent heartbeats, tracking each run's OS process via a `runningProcesses` Map in-memory
> - The `reapOrphanedRuns` sweeper finalizes dangling `running` DB rows when their process has exited
> - The sweeper's early-exit on line 2056 (`if (runningProcesses.has(run.id)) continue`) trusts any in-memory handle without checking if the underlying OS process is still alive
> - On Windows (and occasionally Linux), the Node.js ChildProcess `close` event can fail to fire after an external kill (Task Manager, SIGKILL from outside the process tree), leaving a stale `runningProcesses` entry and causing the run row to remain permanently stuck in `running`
> - This was observed twice: stale runs `724d2a20` and `0b03e3b4`, both with `in-memory handle: yes` in the silence-alert but with the OS PID already gone
> - Recovery required board-level action because `POST /api/heartbeat-runs/:id/cancel` requires `assertBoard` even when the agent's own process is verifiably dead
> - This PR fixes the reaper's liveness check and widens the cancel permission to allow agent self-cancel when the OS process is gone

## What Changed

- **`server/src/services/heartbeat.ts` — `reapOrphanedRuns`**: Moved `tracksLocalChild` computation before the early-exit. When `runningProcesses.has(run.id)`, now checks `handle.child.exitCode !== null || handle.child.killed || (pidDead)` before trusting the handle. Stale handles are cleared and the run falls through to normal finalization.
- **`server/src/routes/agents.ts` — `POST /heartbeat-runs/:id/cancel`**: Relaxed permission to also allow agent self-cancel when the run belongs to the requesting agent and the OS process is verifiably dead (`process.kill(pid, 0)` throws ESRCH). Board access still required for live-process runs. Added `assertCompanyAccess` for all callers.
- **`server/src/__tests__/heartbeat-process-recovery.test.ts`**: Added regression test that inserts a dead `ChildProcess` (killed and awaited) into `runningProcesses` and asserts that `reapOrphanedRuns` detects the stale handle, removes it, and finalizes the run as `failed`.
- **`server/package.json`**: Build script replaces `mkdir -p && cp -R` with cross-platform Node.js `fs.cpSync` for Windows compatibility.

## Verification

```bash
# Run the process-recovery test suite
pnpm test -- --run "heartbeat-process-recovery"
# (Embedded Postgres tests require Linux/Mac; they run in CI)

# Type-check passes clean
pnpm typecheck
```

Manual verification path:
1. Start a local CTO agent heartbeat run
2. Kill the adapter process via Task Manager (simulating Windows hard-kill)
3. Observe the reaper log: `"reap: clearing stale in-memory handle for exited process"` within 30s
4. The run row transitions to `failed`/`queued` (retry) instead of staying `running` forever

## Risks

- The `exitCode` / `killed` flag check is read from the live ChildProcess object synchronously — safe, no race condition.
- The extra `isProcessAlive(pid)` call in the reaper adds one `process.kill(pid, 0)` syscall per stuck run per 30-second sweep. In practice this runs 0–1 times in normal operation.
- Self-cancel permission only applies when PID is verifiably dead or was never set — live-process runs still require board authorization.

## Model Used

- Provider: Anthropic  
- Context: 200k token context window  
- Mode: Tool-use / agentic code editing with read/grep/edit tools  
- Role: CTO agent investigating and implementing the fix autonomously within a Paperclip heartbeat

## Checklist

- [ ] If this change affects the UI, I have included before/after screenshots (N/A — server-only change)

### DIFF
--- a/server/package.json
+++ b/server/package.json
@@ -35,7 +35,7 @@
     "dev": "tsx src/index.ts",
     "dev:watch": "cross-env PAPERCLIP_MIGRATION_PROMPT=never PAPERCLIP_MIGRATION_AUTO_APPLY=true tsx ./scripts/dev-watch.ts",
     "prepare:ui-dist": "bash ../scripts/prepare-server-ui-dist.sh",
-    "build": "tsc && mkdir -p dist/onboarding-assets && cp -R src/onboarding-assets/. dist/onboarding-assets/",
+    "build": "tsc && node -e \"const fs=require('fs');fs.cpSync('src/onboarding-assets','dist/onboarding-assets',{recursive:true,force:true})\"",
     "prepack": "pnpm run prepare:ui-dist",
     "postpack": "rm -rf ui-dist",
     "clean": "rm -rf dist",
--- a/server/src/__tests__/heartbeat-process-recovery.test.ts
+++ b/server/src/__tests__/heartbeat-process-recovery.test.ts
@@ -858,6 +858,40 @@ describeEmbeddedPostgres("heartbeat orphaned process recovery", () => {
     return { companyId, agentId, runId, wakeupRequestId, issueId };
   }
 
+  it("reaps run with a stale in-memory handle when child process has already exited", async () => {
+    // Spawn a process, kill it, wait for it to close — simulating the Windows
+    // edge case where the child exits but the 'close' event never fires back to
+    // the server, leaving a dead ChildProcess object in runningProcesses.
+    const deadChild = spawnAliveProcess();
+    childProcesses.add(deadChild);
+    deadChild.kill("SIGKILL");
+    await new Promise<void>((resolve) => deadChild.once("close", resolve));
+    expect(deadChild.exitCode !== null || deadChild.killed).toBe(true);
+
+    const { agentId, runId } = await seedRunFixture({
+      processPid: deadChild.pid ?? null,
+    });
+    // Simulate the stale handle: put the dead ChildProcess back in the map as
+    // if 'close' had never cleaned it up.
+    runningProcesses.set(runId, { child: deadChild, graceSec: 5 });
+
+    const heartbeat = heartbeatService(db);
+    const result = await heartbeat.reapOrphanedRuns();
+
+    // The stale handle should be detected and the run reaped.
+    expect(result.reaped).toBe(1);
+    expect(result.runIds).toContain(runId);
+    expect(runningProcesses.has(runId)).toBe(false);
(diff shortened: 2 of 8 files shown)


isak-ialogics:
## CI Status: Snyk Failures Are Pre-Existing

The 2 Snyk test failures are **not introduced by this PR**. Running `pnpm audit --prod` on `master` shows the same vulnerabilities:

| Package | Severity | Issue | Fix Version |
|---------|----------|-------|-------------|
| `multer@2.0.2` | High | 3x DoS (GHSA-xf7r, GHSA-v52c, GHSA-5528) | >=2.1.1 |
| `kysely@0.28.11` | High | SQL Injection via JSON path keys | >=0.28.12 |
| `rollup@4.57.1` | High | Arbitrary File Write (via vitest->vite) | >=4.59.0 |

These exist in the base branch identically. Snyk's PR check is configured to block on any high-severity finding, not just newly-introduced ones. Master doesn't run the Snyk check (only `publish_canary`, `verify_canary`, `build-and-push`).

**Options to unblock:**
1. Ignore/override the Snyk status for this PR (pre-existing, unrelated to our changes)
2. Separately upgrade `multer` to 2.1.1 and add pnpm overrides for `kysely`/`rollup` (different scope)

---

## Re: Greptile -- CLI Dependencies

Greptile flagged `postgres`, `ws`, and `zod` as "unused in CLI source tree." This is **expected and correct** -- they are not imported directly in CLI source. They're required because:

- The CLI esbuild config **externalizes all non-`@paperclipai` dependencies** from bundled workspace packages (`@paperclipai/server`, `@paperclipai/db`)
- At runtime, Node resolves these `require()`/`import` calls from `cli/node_modules`
- Without them in `cli/package.json`, standalone installs (`pnpm deploy` / `npm link`) crash with `MODULE_NOT_FOUND`

This is the same pattern as the existing `embedded-postgres`, `dotenv`, `commander`, and `picocolors` entries -- none of those are imported in CLI source either; they're all externalized workspace deps.

The commit message (`fix: add missing transitive deps to CLI package.json`) documents this intent.

isak-ialogics:
Closing: 1850 commits ahead of upstream — completely disconnected history. Fix covered by upstream or by a -clean branch. Cannot be cleanly rebased.

PR #5270, opened 2026-05-05
fix(adapter-opencode-local): default to Zen-routed model so ChatGPT-OAuth users don't 400 (#5132)

### LINKED ISSUE #5132: opencode-local: hire-time default model fails on ChatGPT-OAuth accounts
## Summary

The curated model list in `@paperclipai/adapter-opencode-local` defaults new hires to `openai/gpt-5.2-codex`. That id is rejected at adapter startup for any user authenticated via ChatGPT-OAuth, because the `openai/*` prefix routes through that auth path and `gpt-5.2-codex` is not on the OAuth allowlist:

> `adapter_failed - Bad Request: {"detail":"The 'gpt-5.2-codex' model is not supported when using Codex with a ChatGPT account."}`

The curated list also exposes only `openai/*` ids, so users who pick from the list have no obvious way to land on a `opencode/*` (Zen-routed) entry that would avoid the OAuth allowlist entirely.

We hit this twice in two days on different agent hires. It will keep biting any new ChatGPT-OAuth user until upstream changes something.

## Repro

1. Authenticate `opencode-local` via ChatGPT-OAuth (no Anthropic / Zen api key on file).
2. Hire any agent on adapter type `opencode-local` and accept the default model.
3. First task fails at adapter startup with the 400 above.

Workaround we applied locally: PATCH the agent config to `opencode/gpt-5.2-codex` (Zen-routed, free-credits path).

### A. Drop unsupported defaults from the curated list (preferred)

Either:

- ship only `opencode/*` entries (Zen-routed, no OAuth allowlist drama, matches the "free credits" framing for new users), or
- audit the `openai/*` entries against the current ChatGPT-OAuth allowlist and remove the ones that 400 (today: `gpt-5.2-codex`).

This removes the footgun for new hires who accept the default.

### B. Annotate `agentConfigurationDoc` with provider→credential routing

Add a short section to the adapter's `agentConfigurationDoc` (or model-picker docs) explaining:

- `openai/*` routes through ChatGPT-OAuth and is subject to the OAuth allowlist (current allowlist members).
- `opencode/*` routes through Zen api credentials, no OAuth allowlist applied.
- Recommendation for users on a free / OAuth-only setup.

This doesn't change defaults, but stops users from picking `openai/gpt-5.2-codex` from the list under the assumption it's free.

A + B together is the cleanest fix. A is the higher-leverage of the two.

## Out of scope

- Changes to the broader OpenCode CLI's model handling — `opencode models` already lists the right options. The gap is only in the curated adapter list.
- Migrating already-hired agents — once the adapter config is patched the agent runs fine; this issue is about the default applied at hire time.

## Affected version

`@paperclipai/adapter-opencode-local@2026.428.0`. File: `dist/index.js`, exports `DEFAULT_OPENCODE_LOCAL_MODEL` and `models`.

tarun-khatri:
## Thinking Path

> - The `opencode_local` adapter is the "free credits" lane for new users — it routes through Zen by default rather than forcing a paid OpenAI key. That positioning only works if the curated hire-time default actually accepts the user's auth.
> - Today the default is `openai/gpt-5.2-codex`. The `openai/*` provider prefix routes through ChatGPT-OAuth, which carries an allowlist that does NOT include `gpt-5.2-codex`. So any new user authenticated via ChatGPT-OAuth (the easiest path to get going with no API key) hits `400 Bad Request` at adapter startup the moment they accept the default.
> - The reporter hit this twice in two days on different agent hires; their workaround was to PATCH the agent config to `opencode/gpt-5.2-codex` (Zen-routed). Same model, different provider prefix, different auth path — and the Zen-routed one works for both OAuth and api-key users.
> - The right surface area to fix is the curated default + dropdown, not adapter validation. The existing validator already accepts any `provider/model` shape — users who explicitly type an `openai/*` id and have the matching key still go through unchanged. This PR only changes the dropdown defaults that new hires accept by clicking through.

## What Changed

`packages/adapters/opencode-local/src/index.ts` — three coupled flips:

| Constant | Before | After |
|---|---|---|
| `DEFAULT_OPENCODE_LOCAL_MODEL` | `"openai/gpt-5.2-codex"` | `"opencode/gpt-5.2-codex"` |
| Curated `models[]` entries | 5 × `openai/*` ids | 5 × `opencode/*` ids |
| Cheap profile `adapterConfig.model` | `"openai/gpt-5.1-codex-mini"` | `"opencode/gpt-5.1-codex-mini"` |

Plus a short comment explaining *why* the prefix matters (so it doesn't get flipped back during a future cleanup that doesn't know about the OAuth allowlist).

`packages/adapters/opencode-local/src/server/models.test.ts` — 4 new regression tests in a new `"opencode-local curated defaults (issue #5132)"` describe block:

- `DEFAULT_OPENCODE_LOCAL_MODEL` matches `^opencode\//`
- Every entry in the curated `models[]` matches `^opencode\//` (id and label)
- The default appears in the curated list
- The cheap-profile model matches `^opencode\//`

These pin the invariant that the curated defaults remain Zen-compatible — anyone re-introducing an `openai/*` default will see the test fail explicitly, with a clear error pointing at the OAuth allowlist concern in the comment above.

## Verification

```
pnpm --filter "@paperclipai/adapter-opencode-local" exec vitest run src/server/models.test.ts --testTimeout=30000

# Test Files  1 passed (1)
# Tests  9 passed (9)

pnpm --filter "@paperclipai/server" exec tsc --noEmit
# 0 errors
```

The existing validator test (`requireOpenCodeModelId("openai/gpt-5.2-codex")`) still passes — users who explicitly type an `openai/*` id via a custom `adapterConfig.model` and have the matching auth still go through unchanged. This PR only changes what new hires get when they accept the dropdown default.

## Risks

- **No changes to existing agents.** `adapterConfig` rows aren't touched. Only NEW hires (post-deploy) see the new default. Existing `openai/*`-configured agents continue to run.
- **Users who deliberately want `openai/*`** can still pick it manually via custom `adapterConfig.model` — the validator accepts both prefixes.
- **The dropdown no longer lists `openai/*` entries.** Users who relied on the dropdown to bill their own OpenAI key need to type the `openai/*` id manually on next agent hire. Worth a single line in release notes; matches the issue body's preferred Option A framing.
- **No DB migration. No schema change. No public API change.**

## Model Used

- Used Claude for navigating the package layout and grepping the constants. The fix decision (Option A from the issue body — flip to `opencode/*` outright instead of audit-removing OAuth-incompatible `openai/*` entries) is mine.

## Checklist

- [x] If this change affects the UI, I have included before/after screenshots — N/A (the UI just renders the curated `models[]` array)
- [x] I have updated relevant documentation to reflect my changes — comment in source explains the OAuth allowlist constraint

Closes #5132


devinfoley:
@coderabbitai review

devinfoley:
@greptile review

devinfoley:
@coderabbitai review

PR #5277, opened 2026-05-05
Run explicit-environment adapter tests on the requested target instead of falling back to the host

devinfoley:
## Thinking Path

> - When a user clicks "Test" on a configured environment (SSH or sandbox), the agent-test route exercises the adapter against that target
> - The route previously fell back to running the probe on the Paperclip host whenever an explicit environment target couldn't be resolved, with the test report still saying "passed"
> - That hid two real failure modes: misconfigured environments looked green, and sandbox environments were never actually exercised
> - This pull request acquires an ad-hoc lease and realizes a workspace for sandbox/plugin test environments, resolves a sandbox execution target wired to the environment runtime, and returns synthesized diagnostics instead of running a host probe when an explicit env target can't be resolved
> - The benefit is the Test action surfaces the real environment state and never silently exercises the wrong machine

## What Changed

- `server/routes/agents.ts`: acquire an ad-hoc lease and realize a workspace for sandbox/plugin test environments; resolve a sandbox execution target wired to the environment runtime
- Return synthesized diagnostics (no host fallback) when an explicit env target can't be resolved
- `server/services/environment-runtime.ts`: small adjustments to support the explicit-env-target case
- Clarify test-route messages so they no longer claim a host fallback in explicit env flows
- New `agent-test-environment-routes.test.ts` covers the guard and missing-environment path

## Verification

- `pnpm vitest run --no-coverage server/src/__tests__/agent-test-environment-routes.test.ts`
- `pnpm typecheck` clean
- Manual: a deliberately misconfigured sandbox environment now reports diagnostics instead of a misleading host-pass

## Risks

Medium — Test route behavior change. Explicit environments that previously appeared to pass via host fallback will now report their real state. This is the desired behavior, but operators should expect to see new failures for environments that were never actually working.

## Model Used

Claude Opus 4.7 (1M context)

## Checklist

- [x] I have added or updated tests where applicable — new tests cover guard + missing-env paths

### DIFF
--- a/server/src/__tests__/agent-test-environment-routes.test.ts
+++ b/server/src/__tests__/agent-test-environment-routes.test.ts
@@ -1,0 +1,305 @@
+import express from "express";
+import request from "supertest";
+import { afterEach, beforeEach, describe, expect, it, vi } from "vitest";
+import type { ServerAdapterModule } from "../adapters/index.js";
+
+const mockAgentService = vi.hoisted(() => ({
+  getById: vi.fn(),
+  getChainOfCommand: vi.fn(async () => []),
+}));
+
+const mockAccessService = vi.hoisted(() => ({
+  canUser: vi.fn(),
+  hasPermission: vi.fn(),
+  getMembership: vi.fn(async () => null),
+  listPrincipalGrants: vi.fn(async () => []),
+}));
+
+const mockSecretService = vi.hoisted(() => ({
+  normalizeAdapterConfigForPersistence: vi.fn(async (_companyId: string, config: Record<string, unknown>) => config),
+  resolveAdapterConfigForRuntime: vi.fn(async (_companyId: string, config: Record<string, unknown>) => ({ config })),
+}));
+
+const mockEnvironmentService = vi.hoisted(() => ({
+  getById: vi.fn(),
+  releaseLease: vi.fn(),
+}));
+
+const mockReleaseRunLease = vi.hoisted(() => vi.fn(async () => undefined));
+const mockEnvironmentRuntime = vi.hoisted(() => ({
+  acquireRunLease: vi.fn(),
+  realizeWorkspace: vi.fn(),
+  getDriver: vi.fn(() => ({
+    releaseRunLease: mockReleaseRunLease,
+  })),
+}));
+
+const mockResolveEnvironmentExecutionTarget = vi.hoisted(() => vi.fn());
+const mockInstanceSettingsService = vi.hoisted(() => ({
+  getGeneral: vi.fn(async () => ({ censorUsernameInLogs: false })),
+}));
+
+vi.mock("../services/index.js", () => ({
+  agentService: () => mockAgentService,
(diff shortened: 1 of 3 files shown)


devinfoley:
@greptileai please re-review.

Addressed all three findings from your prior review:

**P1 ×2 — `reuseLease` interference** (`environment-runtime.ts` lines 419-421 and 489-493)
Both the plugin-backed and built-in sandbox paths now skip lease reuse when `heartbeatRunId === null`. An ad-hoc test invocation can no longer resume a provider lease belonging to a running heartbeat, so releasing the test lease cannot tear down the live heartbeat's sandbox.

**P2 — `adapter-test` runId sentinel**
All three call sites (`runId: input.heartbeatRunId ?? "adapter-test"`) now use `randomUUID()` from `node:crypto` so providers that validate or persist the runId still see a well-formed UUID for ad-hoc test leases.

devinfoley:
@greptileai please re-review.

Addressed the reverse `reuseLease` race you flagged in your second pass.

**Fix part 1: store ephemeral policy for ad-hoc test leases**
Both the plugin-backed (line 464-466) and built-in (line 511-515) paths now compute `leasePolicy` as:

```ts
parsed.config.reuseLease && input.heartbeatRunId !== null
  ? "reuse_by_environment"
  : "ephemeral"
```

So an ad-hoc test lease (`heartbeatRunId === null`) is never publishable for reuse, and a concurrent heartbeat cannot pick it up via `findReusableSandboxLeaseId`.

**Fix part 2: defensive filter on read**
`existingLeases` (plugin-backed) and the inline `listLeases().then(...)` (built-in) now filter to `lease.leasePolicy === "reuse_by_environment"` before passing to `findReusableSandboxLeaseId`. Any non-reusable lease (including ad-hoc test leases that may have landed in the table from older code paths) cannot be matched even if it appears in the result set.

devinfoley:
@greptileai please re-review.

Removed the dead-code `fallbackChecks` merge block in `server/src/routes/agents.ts` lines 1457-1467. As you noted, by that point `fallbackChecks` is always `[]` (the early-return guard at line 1432 catches every non-empty case), so the merge branch was unreachable. The remaining flow is now just the simple `status === "fail"` set + `res.json(result)` path.

devinfoley:
@greptileai please re-review.

Added two more tests in `server/src/__tests__/agent-test-environment-routes.test.ts` covering the gaps you noted:

1. **Happy-path success** — `mockResolveEnvironmentExecutionTarget` returns a sandbox target, `testEnvironmentSpy` runs against it, the lease is released with `status: "released"`, and the response is `status: "pass"`.

2. **`realizeWorkspace` failure** — `mockEnvironmentRuntime.realizeWorkspace` rejects, the route returns 200 with `status: "fail"` and an `environment_workspace_realize_failed` check, and the lease is released with `status: "failed"`. The release happens via `releaseLease` inside `resolveAdapterTestExecutionContext` (which then sets the route-side `release` to a noop to prevent double-release).

PR #5295, opened 2026-05-05
Fix timer wake task-session workspace fallback

bdleasure:
## Summary
- stop timer/fresh-session wakes from realizing a saved task-session workspace
- ignore stale previous session cwd when a wake requires a fresh session
- add a regression test for the exact task-session fallback branch

## Verification
- pnpm exec vitest run server/src/__tests__/heartbeat-workspace-session.test.ts

### DIFF
--- a/packages/adapters/codex-local/src/server/execute.ts
+++ b/packages/adapters/codex-local/src/server/execute.ts
@@ -484,10 +484,14 @@ export async function execute(ctx: AdapterExecutionContext): Promise<AdapterExec
     adapterExecutionTargetSessionMatches(runtimeRemoteExecution, executionTarget);
   const codexTransientFallbackMode = readCodexTransientFallbackMode(context);
   const forceSaferInvocation = fallbackModeUsesSaferInvocation(codexTransientFallbackMode);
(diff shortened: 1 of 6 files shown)


iyurinok:
Adding a +1 from another installation experiencing the same root cause.

We traced this bug independently on a claude-local adapter setup and landed on the exact same fix surface: `runtimeSessionFallback` at heartbeat.js:4217 bleeds the global last-used session into any wake where `taskKey` is null (timer wakes, generic non-issue wakes, `issue_commented` without a full issue context). The adapter then sees a non-empty `runtimeSessionId` with `canResumeSession = true` and resumes the wrong conversation thread, causing the agent to post comments into unrelated issues.

PR #5295's approach (gating the workspace fallback behind `allowTaskSessionWorkspace`, returning `true` from `shouldResetTaskSessionForWake` for timer wakes, and removing the synthetic `__heartbeat__` task key) directly addresses all three vectors we identified. We'd appreciate seeing this merged — it's the fix path we found ourselves and the Greptile review confidence score of 4/5 looks solid to us as well.

PR #5323, opened 2026-05-06
Preserve scope on manual heartbeat invokes

devinfoley:
## Thinking Path

> - The agent live-run route lets operators trigger a manual heartbeat invocation so an agent can pick up a specific issue or step out of band
> - The current route flow drops the caller's scope (issue/run context) when forwarding the manual invoke into the heartbeat service, so the resulting run loses the targeting the operator specified
> - This pull request threads the operator-supplied scope through the manual invoke path on both the server route and the UI client, with a regression test that confirms the scope round-trips
> - The benefit is manual heartbeat invokes from the live-run UI actually pick up the scoped issue/run instead of falling through to the agent's default routine

## What Changed

- `server/src/routes/agents.ts`: forward the operator-supplied scope into the manual invoke heartbeat service call
- `server/src/__tests__/agent-live-run-routes.test.ts`: new test verifying the manual invoke path preserves scope
- `ui/src/api/agents.ts`: pass scope through the live-run client API

## Verification

- `pnpm vitest run --no-coverage server/src/__tests__/agent-live-run-routes.test.ts`
- `pnpm typecheck` clean

## Risks

Low. The change is purely additive on the route surface — handlers that did not previously pass scope continue to work; handlers that did pass it now have it preserved instead of dropped.

## Model Used

Claude Opus 4.7 (1M context)

## Checklist

- [x] I have added or updated tests where applicable — new test covers the preserved-scope path
- [x] If this change affects the UI, I have included before/after screenshots — N/A (internal API change, no visible UI shift)

### DIFF
--- a/server/src/__tests__/agent-live-run-routes.test.ts
+++ b/server/src/__tests__/agent-live-run-routes.test.ts
@@ -12,6 +12,7 @@ const mockHeartbeatService = vi.hoisted(() => ({
   getActiveRunIssueSummaryForAgent: vi.fn(),
   getRunLogAccess: vi.fn(),
   readLog: vi.fn(),
+  wakeup: vi.fn(),
 }));
 
 const mockIssueService = vi.hoisted(() => ({
@@ -25,6 +26,8 @@ const mockInstanceSettingsService = vi.hoisted(() => ({
   getGeneral: vi.fn(),
   listCompanyIds: vi.fn(),
 }));
+
+const routeAgentId = "11111111-1111-4111-8111-111111111111";
 
 function registerModuleMocks() {
   vi.doMock("../routes/authz.js", async () => vi.importActual("../routes/authz.js"));
@@ -209,6 +212,14 @@ describe("agent live run routes", () => {
       logRef: "logs/run-1.ndjson",
       content: "chunk",
       nextOffset: 5,
+    });
+    mockHeartbeatService.wakeup.mockResolvedValue({
+      id: "run-1",
+      companyId: "company-1",
+      agentId: "agent-1",
+      status: "queued",
+      invocationSource: "on_demand",
+      triggerDetail: "manual",
     });
   });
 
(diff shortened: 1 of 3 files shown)


devinfoley:
@greptileai please re-review.

Addressed both findings:

**P2 — duplicate handler bodies** (`server/src/routes/agents.ts`)
Extracted the shared body into a private `handleWakeupRoute` helper that takes `{ source, skippedResponse }` options. `/agents/:id/wakeup` and `/agents/:id/heartbeat/invoke` now both call into it; future shared changes will land in one place.

**`source` was caller-controlled on the legacy endpoint**
The legacy `/agents/:id/heartbeat/invoke` endpoint now hardcodes `source: "on_demand"` to match the pre-change behavior (`heartbeat.invoke(id, "on_demand", …)`). The modern `/wakeup` endpoint still reads `source` from the validated request body. Comment in the route explains the asymmetry.

devinfoley:
@greptileai please re-review.

Two follow-ups:

**Restored backward compat for empty-body callers** (`server/src/routes/agents.ts`)
The legacy `/heartbeat/invoke` endpoint was passing `payload: null`, `idempotencyKey: null`, and `forceFreshSession: false` to `heartbeat.wakeup()` even when the caller posted an empty body — that broke the signoff-policy e2e test which relies on the original fixed-arg `heartbeat.invoke()` shape. The route now only forwards the wake fields the caller actually supplied, so an empty body produces the legacy minimal shape and an explicit body still preserves scope (reason / payload / forceFreshSession).

**Added empty-body test** (`server/src/__tests__/agent-live-run-routes.test.ts`)
New test `calls heartbeat.wakeup with the legacy minimal shape when the body is empty` pins the backward-compat path so the regression can't slip in again.

The `source: "on_demand"` hardcode and shared `handleWakeupRoute` helper for `/wakeup` from the prior round are unchanged.

devinfoley:
@greptileai please re-review.

Root cause for the e2e regression: `validate(wakeAgentSchema)` middleware on the legacy `/heartbeat/invoke` endpoint rejected empty request bodies. The signoff-policy e2e test posts an empty body and expects 2xx — `zod.object({...}).parse(undefined)` throws, returning 5xx, which surfaced as `expect(res.ok()).toBe(true)` failing on every `invokeHeartbeat()` call (5 test cases).

Removed the `validate` middleware from the legacy endpoint and now read body fields defensively with type guards (`typeof body.reason === "string"`, etc.). Empty bodies pass through and produce the original minimal shape. Bodies with reason / payload / forceFreshSession get those fields preserved. The modern `/wakeup` endpoint keeps `validate(wakeAgentSchema)` — its callers always supply structured bodies.

PR #5324, opened 2026-05-06
Expand sandbox callback bridge allowlist to cover the documented heartbeat surface

devinfoley:
## Thinking Path

> - When an agent runs in an e2b sandbox or other non-managed environment, it talks back to the Paperclip server through a per-lease callback bridge that proxies HTTP requests
> - The bridge has an allowlist of method/path patterns it will forward; anything outside the list is rejected to keep the bridge tight
> - The allowlist had drifted behind what the heartbeat documentation describes as the supported callback surface — several documented endpoints (issue updates, agent-side log emit, work-status writes) were being rejected at the bridge
> - This pull request expands the allowlist to cover the documented heartbeat surface and adds tests that pin every newly-allowed pattern, so the doc and the bridge stay in sync
> - The benefit is sandboxed runs no longer hit "method not allowed" / "path not allowed" rejections on the documented set of callbacks

## What Changed

- `packages/adapter-utils/src/sandbox-callback-bridge.ts`: expand the method/path allowlist to match the documented heartbeat callback surface
- `packages/adapter-utils/src/sandbox-callback-bridge.test.ts`: add coverage for every newly-allowed pattern, plus negative cases for patterns that should still be rejected

## Verification

- `pnpm vitest run --no-coverage --project @paperclipai/adapter-utils`
- `pnpm typecheck` clean
- Manual: previously-rejected callbacks from sandboxed runs now succeed end-to-end

## Risks

Low. The allowlist only grows; nothing previously allowed is now blocked. Tests pin both the new allowed patterns and that out-of-doc patterns stay rejected.

## Model Used

Claude Opus 4.7 (1M context)

## Checklist

- [x] I have added or updated tests where applicable — new tests cover added patterns + still-rejected negatives

### DIFF
--- a/packages/adapter-utils/src/sandbox-callback-bridge.test.ts
+++ b/packages/adapter-utils/src/sandbox-callback-bridge.test.ts
@@ -7,6 +7,7 @@ import { afterEach, describe, expect, it } from "vitest";
 
 import { prepareCommandManagedRuntime } from "./command-managed-runtime.js";
 import {
+  authorizeSandboxCallbackBridgeRequestWithRoutes,
   createFileSystemSandboxCallbackBridgeQueueClient,
   createSandboxCallbackBridgeAsset,
   createSandboxCallbackBridgeToken,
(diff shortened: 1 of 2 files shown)


devinfoley:
@greptileai please re-review.

Addressed the runtime-services regex finding.

**Narrowed regex** (`packages/adapter-utils/src/sandbox-callback-bridge.ts` line 85)
`/runtime-services/[^/]+` → `/runtime-services/(?:start|stop|restart)`. Pinned to the three documented heartbeat actions; any other action (delete, reset, wipe, etc.) stays denied.

**Negative test pin** (`packages/adapter-utils/src/sandbox-callback-bridge.test.ts`)
Added `{ method: "POST", path: "/api/execution-workspaces/ws-1/runtime-services/delete" }` to the denied list with a comment explaining the pin so future widening of the regex breaks the test loudly.

PR #5325, opened 2026-05-06
Sanitize remote execution envs at the boundary

devinfoley:
## Thinking Path

> - Adapters spawn CLIs against local, SSH, and sandbox targets, threading a runtime env through `runAdapterExecutionTargetProcess` and the SSH/sandbox runners
> - Host identity vars (HOME, TMPDIR, XDG_*, NVM_DIR, PATH) routinely leak into the env we send to remote targets — sometimes via test probes, sometimes via runtime config — and break sandboxed/SSH'd CLIs whose own profiles set those values correctly
> - The sanitization logic existed but lived alongside other helpers in `server-utils.ts` and was applied piecemeal at adapter callsites, so it was easy to bypass
> - This pull request lifts the sanitization into a standalone `remote-execution-env.ts`, applies it at the SSH and sandbox runtime boundary so every remote spawn goes through it, and removes the duplicated callsite-level filtering
> - The benefit is identity-bound host env stops leaking across SSH/sandbox transports regardless of which adapter calls in

## What Changed

- `packages/adapter-utils/src/remote-execution-env.ts`: new module — single source of truth for which env keys are identity-bound and how to strip them when the value matches the host's value
- `packages/adapter-utils/src/server-utils.ts`: remove the inline sanitization (now in `remote-execution-env.ts`)
- `packages/adapter-utils/src/execution-target.ts`: apply sanitization at the sandbox runtime boundary
- `packages/adapter-utils/src/ssh.ts`: apply sanitization at the SSH spawn boundary
- `packages/adapters/opencode-local/src/server/test.ts`: drop now-redundant callsite filtering
- `packages/adapters/pi-local/src/server/test.ts`: drop now-redundant callsite filtering
- New tests `execution-target.test.ts` and `execution-target-sandbox.test.ts` cover the sanitizer flow at both transports, including positive cases (host-shaped path stripped) and explicit-override preservation

## Verification

- `pnpm vitest run --no-coverage --project @paperclipai/adapter-utils --project @paperclipai/adapter-opencode-local --project @paperclipai/adapter-pi-local`
- `pnpm typecheck` clean

## Risks

Low–medium. The sanitization is now applied at one layer (boundary) instead of N (callsites), so behavior is more consistent. Any adapter that previously relied on a leaked host var landing on the remote shell would now see it stripped — but those reliances were what this change exists to fix.

## Model Used

Claude Opus 4.7 (1M context)

## Checklist

- [x] I have added or updated tests where applicable — new tests at both transports

### DIFF
--- a/packages/adapter-utils/src/execution-target-sandbox.test.ts
+++ b/packages/adapter-utils/src/execution-target-sandbox.test.ts
@@ -19,6 +19,7 @@ describe("sandbox adapter execution targets", () => {
   const cleanupDirs: string[] = [];
 
   afterEach(async () => {
+    vi.unstubAllEnvs();
     while (cleanupDirs.length > 0) {
       const dir = cleanupDirs.pop();
       if (!dir) continue;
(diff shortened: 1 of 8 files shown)


devinfoley:
@greptileai please re-review.

Addressed both findings:

**Style: unnecessary intermediate array** (`packages/adapter-utils/src/ssh.ts`)
`envEntries.map(([key]) => key)` replaced with direct destructure-in-for: `for (const [key] of envEntries)`.

**Env override injected before login-profile sourcing**
`runSshCommand` now mirrors `buildSshSpawnTarget`: wraps the remote command in `sh -lc` with profile sourcing (`/etc/profile`, `~/.profile`, `~/.bash_profile`, `~/.zprofile`) followed by `exec env KEY=VAL sh -c <cmd>`. The `env` is now applied AFTER profile sourcing, so user-supplied identity overrides win over anything a profile re-exports — symmetric with the process path.

devinfoley:
@greptileai please re-review.

Addressed the double-nested login-shell finding. `runAdapterExecutionTargetShellCommand` (`packages/adapter-utils/src/execution-target.ts` line ~391) now passes the raw `command` string to `runSshCommand` instead of pre-wrapping it in `sh -lc`. Since `runSshCommand` already wraps with `sh -lc '<profile sourcing> && exec env K=V sh -c <cmd>'`, the prior pre-wrap caused a second nested `sh -lc` *after* the env override — re-sourcing profiles AFTER the explicit overrides and undoing identity-var preservation.

Updated `execution-target.test.ts` to assert the raw command (no `sh -lc` prefix) is passed through, with a comment explaining why the wrapper belongs only inside `runSshCommand`. Tests pass.

PR #5326, opened 2026-05-06
Serialize sandbox callback bridge against concurrent heartbeats

devinfoley:
> **Stacked PR.** This PR's branch carries cumulative content from #5324 (bridge allowlist expand) and #5325 (env sanitization) — the mutex/sha256 logic in this PR sits on top of both. Reviewers should focus on the files this PR's commit touches: `packages/adapter-utils/src/sandbox-callback-bridge.{ts,test.ts}`, `packages/adapter-utils/src/ssh.ts`, and `packages/adapter-utils/src/ssh-fixture.test.ts`. Will rebase onto `master` and force-push once both prerequisite PRs are merged.

## Thinking Path

> - Each agent that runs in a sandbox or via SSH talks back to the Paperclip server through a per-lease callback bridge whose entrypoint script is uploaded to the remote
> - When two heartbeats target the same agent on the same machine concurrently, both upload the bridge entrypoint and both write to the same response files — producing torn-write races: `SyntaxError: Identifier 'randomUUID' has already been declared` from a concatenated upload, `mv: cannot stat …` from colliding `.json.tmp` writes, and 0-byte commits from a truncated stdin
> - This pull request serializes those operations with a POSIX `mkdir`-mutex (PID liveness check + atomic rename) at the bridge entrypoint upload, applies the same lock to the bridge response writer, forwards stdin into remote ssh commands so the entrypoint payload arrives intact, and verifies a sha256 of the upload before promoting it
> - The benefit is concurrent heartbeats no longer corrupt each other's bridge state

## What Changed

- `packages/adapter-utils/src/sandbox-callback-bridge.ts`: serialize entrypoint upload and response writes via POSIX `mkdir`-mutex with PID liveness; sha256 the upload before promoting via `mv`; content-skip when the existing entrypoint already matches
- `packages/adapter-utils/src/ssh.ts`: forward stdin into remote ssh commands through the SSH managed runtime so `cat > "$remote_upload"` actually receives the base64-encoded entrypoint
- `packages/adapter-utils/src/ssh-fixture.test.ts`: cover the stdin-forwarded SSH path
- `packages/adapter-utils/src/sandbox-callback-bridge.test.ts`: cover the mutex, content-skip, sha256-verify, and atomic-rename paths

## Verification

- `pnpm vitest run --no-coverage --project @paperclipai/adapter-utils`
- `pnpm typecheck` clean
- Manual: two parallel heartbeats targeting the same SSH agent no longer race on the bridge entrypoint or response files

## Risks

Medium. Serializing previously-parallel operations adds latency on the contended path (one heartbeat waits on another), bounded by the entrypoint upload time. The mutex includes PID liveness so a crashed heartbeat doesn't deadlock subsequent ones. Sha256-verify gives a clear "torn upload" failure mode instead of silent 0-byte commits.

## Model Used

Claude Opus 4.7 (1M context)

## Checklist

- [x] I have added or updated tests where applicable — tests cover mutex + sha256-verify + stdin-forwarded ssh

### DIFF
--- a/packages/adapter-utils/src/sandbox-callback-bridge.test.ts
+++ b/packages/adapter-utils/src/sandbox-callback-bridge.test.ts
@@ -8,10 +8,12 @@ import { afterEach, describe, expect, it } from "vitest";
 import { prepareCommandManagedRuntime } from "./command-managed-runtime.js";
 import {
   authorizeSandboxCallbackBridgeRequestWithRoutes,
+  createCommandManagedSandboxCallbackBridgeQueueClient,
   createFileSystemSandboxCallbackBridgeQueueClient,
   createSandboxCallbackBridgeAsset,
   createSandboxCallbackBridgeToken,
   sandboxCallbackBridgeDirectories,
+  syncSandboxCallbackBridgeEntrypoint,
   startSandboxCallbackBridgeServer,
   startSandboxCallbackBridgeWorker,
 } from "./sandbox-callback-bridge.js";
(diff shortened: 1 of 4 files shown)


devinfoley:
@greptileai please re-review.

Addressed all three P2s:

**False "sha mismatch" when sha256sum/shasum unavailable** (`packages/adapter-utils/src/sandbox-callback-bridge.ts`)
Replaced `$(hash_file ... 2>/dev/null || true)` with explicit `if hash_file ...; then verify; else log skipped; fi` — minimal Alpine/scratch images without coreutils now log a clear "sha verify skipped" message instead of failing every upload with an opaque mismatch error. Also updated `current_sha` cache check to require non-empty + match.

**Filesystem-client mutex now has PID liveness** (`packages/adapter-utils/src/sandbox-callback-bridge.ts`)
`writeResponseFile` mkdir-mutex now writes `process.pid` into `<lockDir>/pid` on acquire, and on EEXIST checks `process.kill(holderPid, 0)` to detect a dead holder before sleeping. A SIGKILLed heartbeat no longer deadlocks subsequent writers for the full 30s timeout — mirrors the shell-side `buildRemotePidLockAcquireScript`.

**SIGTERM → SIGKILL escalation** (`packages/adapter-utils/src/ssh.ts`)
`spawnText` timeout path now schedules a 5s SIGKILL escalation after the initial SIGTERM. Hung remote commands that ignore SIGTERM (e.g., a wedged ssh forward) get force-killed instead of holding resources indefinitely. Cleanup wired into both `error` and `close` handlers so the escalation timer is always cleared.

PR #5331, opened 2026-05-06
feat(issues): add blockReason field to distinguish stale-heartbeat from upstream blocks

blevinson:
## Summary

- Adds `blockReason` column (`upstream | stale_heartbeat | manual`) to the `issues` table
- Heartbeat retry-exhaustion path now sets `blockReason=stale_heartbeat` automatically
- PATCH with non-empty `blockedByIssueIds` auto-sets `blockReason=upstream`; explicit `status=blocked` with no blockers defaults to `manual`
- Field clears to `null` when status leaves `blocked`
- Migration 0060 backfills existing blocked rows: rows with blocker relations → `upstream`, others → `manual`

## Changed files

| File | Change |
|---|---|
| `packages/shared/src/constants.ts` | Add `ISSUE_BLOCK_REASONS` enum + `IssueBlockReason` type |
| `packages/db/src/schema/issues.ts` | Add `blockReason text` column |
| `packages/shared/src/validators/issue.ts` | Add optional `blockReason` field to create/update schemas |
| `server/src/services/issues.ts` | Auto-set `blockReason` in `update()` based on status + blockedByIssueIds |
| `server/src/services/heartbeat.ts` | Pass `blockReason: "stale_heartbeat"` in `escalateStrandedAssignedIssue` |
| `packages/db/src/migrations/0060_block_reason.sql` | Migration + backfill |
| `packages/db/src/migrations/meta/_journal.json` | Register migration entry |

Resolves: FIN-1457

### DIFF
--- a/packages/db/src/migrations/0060_block_reason.sql
+++ b/packages/db/src/migrations/0060_block_reason.sql
@@ -1,0 +1,12 @@
+ALTER TABLE "issues" ADD COLUMN "block_reason" text;
+--> statement-breakpoint
+UPDATE "issues" i
+SET block_reason = CASE
+  WHEN EXISTS (
+    SELECT 1 FROM issue_relations ir
+    WHERE ir.related_issue_id = i.id
+    AND ir.type = 'blocks'
+  ) THEN 'upstream'
+  ELSE 'manual'
+END
+WHERE i.status = 'blocked';
--- a/packages/db/src/migrations/meta/_journal.json
+++ b/packages/db/src/migrations/meta/_journal.json
@@ -421,6 +421,13 @@
       "when": 1776542246000,
       "tag": "0059_plugin_database_namespaces",
       "breakpoints": true
+    },
+    {
+      "idx": 60,
+      "version": "7",
+      "when": 1746579161000,
+      "tag": "0060_block_reason",
+      "breakpoints": true
     }
   ]
 }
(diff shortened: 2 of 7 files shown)


blevinson (commented):
**Stage 1 peer review — APPROVED** (Developer-2, FIN-1457 2-stage policy)

Clean, well-scoped +40/-0 change. Migration is additive-only (nullable text column, backfill scoped to `status='blocked'`) — zero-downtime safe.

**Greptile findings:**
- **P1 (stale blockReason on already-blocked issue):** False positive. `const finalStatus = issueData.status ?? existing.status` causes `finalStatus="blocked"` even when status is absent from the patch, so the `blockedByIssueIds !== undefined` branch fires correctly and re-derives the reason. No bug.
- **P2 (journal timestamp):** False positive — diff already has `1776579161000` (~April 2026), matching greptile's own suggestion.
- **P2 (PR template):** Valid process note; not blocking for this narrow change.

**Minor observations:**
1. Confirm `updateIssueSchema` picks up `blockReason` (auto if it's `.partial()` of `createIssueSchema`; else add explicitly).
2. Clear-on-exit only fires when `status` is explicit in the patch — resolving a blocker via relations without a status change won't clear it, but the auto-unblock flow covers this. Acceptable.

Ready for Stage 2 (COO audit).

PR #5365, opened 2026-05-06
adapter-utils: add silenceTimeoutSec stream-silence watchdog to runChildProcess

kursatayvat:
## Summary

Adds an opt-in stdout/stderr silence watchdog to `runChildProcess` in `@paperclipai/adapter-utils`, with the option threaded through `runAdapterExecutionTargetProcess` for local + SSH adapters. Sandbox path accepts and ignores the option for now (`TODO(stream-silence-remote)`).

This is the platform half of an internal incident-response fix (INF-77 / INF-80). It addresses a failure mode where a spawned CLI's OS process stays alive while its stdout/stderr go silent — the existing wall-clock `timeoutSec` only fires on overall runtime, not on stream silence. The fix is **default-off**: callers must explicitly pass `silenceTimeoutSec` to opt in, and adapter-level enablement is a sibling task (INF-81).

Plan reference (internal): plan rev 2, board-accepted (revision id `8ad7199d-a1c5-42d4-a947-d0c1709b88fa`). Full design + sequence diagrams + race matrix + research findings live in the internal tracker; this PR implements §3.1 and §3.2 verbatim.

## What changed

| File | Δ | What |
|---|---|---|
| `packages/adapter-utils/src/server-utils.ts` | +62 | `silenceTimeoutSec?: number` opt on `runChildProcess`; `errorCode?: "timeout" \| "stream_silence_timeout" \| null` on `RunProcessResult`; arm-at-spawn watchdog with reset-on-data and SIGTERM→SIGKILL after `graceSec`. |
| `packages/adapter-utils/src/execution-target.ts` | +14 | `silenceTimeoutSec?` on `AdapterExecutionTargetProcessOptions`, threaded to `runChildProcess` on local + SSH paths. Sandbox path accepts and ignores with `TODO(stream-silence-remote)`. |
| `packages/adapter-utils/src/server-utils.test.ts` | +258 | 10 new vitest cases covering silent kill, normal completion, intermittent output, SIGTERM→SIGKILL escalation, both race directions, 0/undefined no-op, terminal-cleanup precedence, zero-byte spawn-time silence. |

### Behavior

- `silenceTimeoutSec` undefined or `<= 0` → watchdog disabled, **byte-for-byte identical to current behavior**. The 5 existing `runChildProcess` test cases pass unchanged.
- `silenceTimeoutSec > 0` → arm a single timer at spawn for `silenceTimeoutSec * 1000` ms. Reset on every `child.stdout` / `child.stderr` `data` event. On expiry: SIGTERM the process group, then SIGKILL after `graceSec` seconds.
- `errorCode` is populated only when `timedOut === true`: `"timeout"` for the wall-clock path, `"stream_silence_timeout"` for the new path. `timedOut: true` is preserved for both paths so existing branches that read `proc.timedOut` keep working. `errorCode` is **optional** on `RunProcessResult` so existing `RunProcessResult` literals across the codebase continue to type-check.
- Race resolution: all three terminal paths (silence-fire / wall-clock-fire / `terminalResultCleanup`) guard on `timedOut || terminalCleanupStarted` before acting, and `clearSilenceTimers()` is called from `error`, `close`, and the wall-clock timeout body. No leak window.

## Test plan

- [x] `pnpm exec vitest run packages/adapter-utils/` — 81/81 pass (10 new cases + 71 existing).
- [x] `pnpm typecheck` clean across `adapter-utils`, `server`, and the 8 local-adapter packages (`adapter-{claude,codex,cursor,gemini,opencode,pi,acpx}-local`, `adapter-openclaw-gateway`).
- [x] Default-off invariant: existing `runChildProcess` tests don't set `silenceTimeoutSec` and pass without modification.
- [x] POSIX-only tests skipped on Windows (the watchdog uses POSIX process-group signaling).

```
$ pnpm exec vitest run packages/adapter-utils/
 Test Files  8 passed (8)
      Tests  81 passed (81)
   Duration  7.68s
```

## Out of scope

- **Adapter wire-ups** — sibling task INF-81; default `silenceTimeoutSec` per local adapter (recommend 300s, aligned with the Claude CLI's own 5-minute SSE idle watchdog) is added there. INF-81 is blocked-on this PR.
- **Sandbox / SSH remote-side silence detection** — the sandbox runner doesn't expose `data` events the same way; piping silence detection requires a separate change. Tracked via `TODO(stream-silence-remote)` in `execution-target.ts`.
- **Recovery service auto-terminate** — separate ticket (INF-78).

## Reviewer notes

Internal review by Staff Engineer (audit findings, "ship-with-followups" verdict) covered architecture, default-off invariant, race handling, type safety, grace consistency, and test coverage. Two non-blocking nits were raised:

1. The `silenceTimeoutSec=0` / `undefined` no-op tests use a behavioral proxy (`Promise.race` sentinel + cleanup) rather than asserting "no timer was armed" directly. Adequate given the trivial control flow.
2. No structured log/metric emitted when the watchdog fires. Consumers see `errorCode: "stream_silence_timeout"` and can log it themselves; richer signal will be added when the recovery service (INF-78) needs it.

### DIFF
--- a/packages/adapter-utils/src/execution-target.ts
+++ b/packages/adapter-utils/src/execution-target.ts
@@ -78,6 +78,12 @@ export interface AdapterExecutionTargetProcessOptions {
   stdin?: string;
   timeoutSec: number;
   graceSec: number;
+  /**
+   * Stdout/stderr silence watchdog (seconds). Local target: passed through to
+   * `runChildProcess`. Sandbox / SSH targets accept the option but ignore it
+   * for now — see TODO(stream-silence-remote) below.
+   */
+  silenceTimeoutSec?: number;
   onLog: (stream: "stdout" | "stderr", chunk: string) => Promise<void>;
   onSpawn?: (meta: { pid: number; processGroupId: number | null; startedAt: string }) => Promise<void>;
   terminalResultCleanup?: TerminalResultCleanupOptions;
@@ -342,6 +348,10 @@ export async function runAdapterExecutionTargetProcess(
   options: AdapterExecutionTargetProcessOptions,
 ): Promise<RunProcessResult> {
   if (target?.kind === "remote" && target.transport === "sandbox") {
+    // TODO(stream-silence-remote): the sandbox runner does not expose live
+    // stdout/stderr `data` events the same way `runChildProcess` does. Wiring
+    // silence detection into this path requires a separate change; the option
+    // is accepted but ignored for sandbox/SSH targets in v1.
     const runner = requireSandboxRunner(target);
     const env = sanitizeRemoteExecutionEnv(options.env);
     return await runner.execute({
@@ -363,12 +373,16 @@ export async function runAdapterExecutionTargetProcess(
       ? sanitizeRemoteExecutionEnv(options.env)
       : options.env;
 
+  // TODO(stream-silence-remote): SSH path also goes through `runChildProcess`,
+  // but the child is `ssh`, not the remote command. Silence on the local SSH
+  // pipe is a coarse proxy for remote silence; behaves like local for now.
   return await runChildProcess(runId, command, args, {
     cwd: options.cwd,
     env,
     stdin: options.stdin,
     timeoutSec: options.timeoutSec,
     graceSec: options.graceSec,
+    silenceTimeoutSec: options.silenceTimeoutSec,
     onLog: options.onLog,
     onSpawn: options.onSpawn,
     terminalResultCleanup: options.terminalResultCleanup,
--- a/packages/adapter-utils/src/server-utils.test.ts
+++ b/packages/adapter-utils/src/server-utils.test.ts
@@ -414,6 +414,264 @@ describe("runChildProcess", () => {
   });
 });
 
+describe("runChildProcess silenceTimeoutSec watchdog", () => {
+  it.skipIf(process.platform === "win32")("kills a silent child and reports stream_silence_timeout", async () => {
+    const result = await runChildProcess(
+      randomUUID(),
+      process.execPath,
+      ["-e", "setTimeout(() => process.stdout.write('late'), 5000);"],
+      {
+        cwd: process.cwd(),
+        env: {},
+        timeoutSec: 0,
+        graceSec: 1,
+        silenceTimeoutSec: 0.3,
+        onLog: async () => {},
+      },
+    );
+
+    expect(result.timedOut).toBe(true);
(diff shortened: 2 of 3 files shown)


kursatayvat (commented):
LGTM. Reviewed the diff at `3bc1594` against the design spec; the implementation matches §3.1/§3.2 verbatim and the test set covers all 10 race/edge cases from the matrix.

A few things that stood out as well-handled:

- **Default-off invariant**: the `silenceTimeoutSec <= 0 || undefined` short-circuit in `armSilenceTimer()` keeps the existing call sites byte-for-byte equivalent. The 5 prior `runChildProcess` tests pass unchanged, which is the cheapest possible regression bound.
- **Race resolution**: all three terminal paths (silence-fire / wall-clock-fire / terminal-result-cleanup) check `timedOut || terminalCleanupStarted` before acting, and `clearSilenceTimers()` is called on `error`, `close`, and the wall-clock body. No leak window between "decision made" and "timer cleared."
- **Type safety**: making `errorCode` optional on `RunProcessResult` keeps every existing literal across the four downstream consumers (`server/src/adapters/utils.ts`, `claude-local`, `ssh.ts`, `sandbox-callback-bridge.ts`) typechecking — confirmed `tsc --noEmit` clean.
- **Grace consistency**: all three SIGKILL sites use `Math.max(1, opts.graceSec) * 1000`, no drift between the wall-clock and silence kill paths.

Two non-blocking observations (intentionally not addressed here, parked as follow-ups):

1. The `silenceTimeoutSec=0`/`undefined` no-op tests assert "promise unsettled after 500ms" then SIGKILL the child to drain. That's a behavioral proxy rather than a strict "no timer was armed" check — acceptable given the trivial early-return control flow, but if you ever want a stricter assertion, a spy on `setTimeout` would do it.
2. No structured log/metric is emitted when the silence watchdog actually fires — consumers see `errorCode: "stream_silence_timeout"` on the result and can log it themselves. That's correct for this layer; the recovery-service work where the richer signal has a real consumer can carry it.

Sandbox/SSH `TODO(stream-silence-remote)` is the right call for v1 — the SSH child is `ssh` itself, not the remote command, so silence on the local pipe is a coarse proxy. Worth a separate change.

Approving. CI gate independent.

brandonburr:

Thank you for this contribution! This PR is **recommended for merge**.

**Rationale:** Adapter: silenceTimeoutSec stream-silence watchdog added to runChildProcess.

PR #5384, opened 2026-05-06
fix(workspaces): fetch origin before branching new worktrees

newcomer:
## Thinking Path

> - Each issue an agent picks up gets its own isolated git worktree, branched from the project's base ref (typically `origin/main`)
> - The worktree is created via `git worktree add -b <branch> <path> <baseRef>` against the project's primary checkout
> - But the primary checkout's local `origin/main` only updates when something does a `git fetch` — and nothing in the provisioning path does
> - So when an agent picks up a new issue after a quiet period, the new worktree silently branches from a stale local `origin/main` (could be hours/days old)
> - This pull request adds a best-effort `git fetch origin` before each `git worktree add -b`, so new worktrees branch from the current remote tip
> - The benefit is that agents get a fresh starting point for new work, PRs avoid avoidable rebases against drift, and the close panel's "merged into base" check (which uses the same local ref) returns correct results

## What Changed

- New helper `fetchOriginBestEffort(recorder, repoRoot)` in `server/src/services/workspace-runtime.ts` that runs `git fetch origin` and logs (without throwing) on failure
- Called at the two `git worktree add -b ... baseRef` sites: the fresh-create path and the recreate-missing-worktree path
- Two new tests in `server/src/__tests__/workspace-runtime.test.ts`:
  - **fetches origin before branching**: pushes an initial commit, advances the remote independently from a side clone, then verifies the new worktree branches from the current remote tip rather than the stale local `origin/main`
  - **proceeds with possibly-stale refs when fetch fails**: no origin remote configured, fetch fails, provisioning still succeeds (best-effort)
- One existing assertion (`records worktree setup and provision operations when a recorder is provided`) updated to include the new fetch operation in the expected sequence

## Verification

\`\`\`bash
pnpm run preflight:workspace-links
cd server && npx vitest run --no-coverage src/__tests__/workspace-runtime.test.ts
# 59 passed (59)
\`\`\`

The two new tests cover both the success and failure paths. The success test produces a verifiable result (worktree HEAD matches the post-advance remote tip), so a regression where the fetch is removed would fail it.

## Risks

- **Behavior change**: Existing deployments will see a small additional latency per provision (single-digit seconds for typical repos, scales with fetch payload). For correctness this seems clearly worth it, but flagging.
- **Best-effort failure handling**: If origin is unreachable (network blip, GitHub outage, auth misconfig), provisioning still proceeds with the stale local ref. The trade-off is preserving "agents can keep picking up issues even when GitHub is down" over "fail loudly when we can't refresh." A warning is logged for later inspection.
- **No new config surface**: Considered an opt-out flag (e.g., `skipFetchBeforeProvision`) but kept it out to stay Path 1 / minimal. If specific deployments need to suppress the fetch (offline, airgapped, deterministic-test), happy to follow up with a separate PR adding a config knob.
- **Non-`-b` worktree paths are not touched**: The reattach paths (`git worktree add path branch` without `-b`, attaching an existing branch) don't branch from a base ref, so don't need the fetch. Only the two `-b` paths get it.

## Model Used

- Provider: Anthropic
- Model ID: claude-opus-4-7
- Context window: 1M
- Capabilities: extended thinking, tool use, file editing, code execution

## Checklist

- [x] If this change affects the UI, I have included before/after screenshots (N/A — server-side only)
- [x] I have updated relevant documentation to reflect my changes (N/A — no doc references to the prior behavior)

### DIFF
--- a/server/src/__tests__/workspace-runtime.test.ts
+++ b/server/src/__tests__/workspace-runtime.test.ts
@@ -1459,14 +1459,17 @@ describe("realizeExecutionWorkspace", () => {
 
     expect(operations.map((operation) => operation.phase)).toEqual([
       "worktree_prepare",
+      "worktree_prepare",
       "workspace_provision",
     ]);
-    expect(operations[0]?.command).toContain("git worktree add");
-    expect(operations[0]?.metadata).toMatchObject({
+    expect(operations[0]?.command).toBe("git fetch origin");
+    expect(operations[0]?.metadata).toMatchObject({ fetchOrigin: true });
+    expect(operations[1]?.command).toContain("git worktree add");
+    expect(operations[1]?.metadata).toMatchObject({
       branchName: "PAP-540-record-workspace-operations",
       created: true,
     });
-    expect(operations[1]?.command).toBe("bash ./scripts/provision.sh");
+    expect(operations[2]?.command).toBe("bash ./scripts/provision.sh");
   });
 
   it("truncates oversized provision command output before storing it in memory", async () => {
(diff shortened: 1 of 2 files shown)


newcomer:
Thanks for the review @greptile-apps. Addressed the gap by adding a test that covers the recreate fall-through path in `ensurePersistedExecutionWorkspaceAvailable`:

- Realizes a workspace (creates worktree + branch)
- Removes both the worktree directory **and** the branch
- Calls `ensurePersistedExecutionWorkspaceAvailable`
- Asserts the operation order: failed reattach → fetch → recreate (with `created: true, restored: true`)

A regression that drops or misorders the fetch in that catch block now fails the suite. All 60 tests in `workspace-runtime.test.ts` pass locally.

[Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=31070379)

PR #5399, opened 2026-05-07
skills/paperclip-create-agent: require post-hire bundle verification

jonatanritter:
## Thinking Path

> - Hiring agents is the entry point — `skills/paperclip-create-agent` walks operators (human or AI) through the hire/approval flow that ends in a working agent
> - `POST /api/companies/{companyId}/agent-hires` accepts `instructionsBundle.files` and normally materializes them via `materializeDefaultInstructionsBundleForNewAgent` (`server/src/routes/agents.ts:1074-1122`). However, the early-return in lines 1090-1107 — triggered when `adapterConfig` contains any of `instructionsBundleMode`, `instructionsRootPath`, `instructionsEntryFile`, `instructionsFilePath`, or `agentsMdPath` — silently bypasses materialization, so the supplied `instructionsBundle.files` are dropped while the bundle config still gets persisted
> - The failure mode is silent: hire returns 200, approval flows normally, agent ships with `files: []` and `resolvedEntryPath: null`, and every heartbeat then runs with an empty prompt
> - The skill had no post-hire verification step, so this only surfaced on audit — by which point empty-shell agents may already be queued or running
> - This PR adds a mandatory verification step (Step 10) and documents both endpoints used by it, so the skill no longer treats verification as an optional follow-up
> - The benefit is that any operator following the skill — human or AI — gets a hard gate that catches empty-shell hires before issues are closed

## What Changed

- `skills/paperclip-create-agent/SKILL.md` — new **Step 10** that requires `GET /api/agents/{agentId}/instructions-bundle` and asserts `files.length >= 1`, non-null `resolvedEntryPath`, matching `entryFile`, and presence of every file from the original `instructionsBundle.files`. Includes the fallback `PUT /api/agents/{agentId}/instructions-bundle/file` recipe. Step 8 example fixed to include the previously missing `entryFile` field. Steps 8/9 cross-reference Step 10 and gate close-out on it passing.
- `skills/paperclip-create-agent/references/api-reference.md` — documents both endpoints used by Step 10 (response shape for the GET, request schema for the PUT). Notes that for the PUT, `path` and `content` go in the **request body** (the PUT handler reads `req.body.path` / `req.body.content` at `server/src/routes/agents.ts:2465`). The corresponding GET, by contrast, takes `path` from the query string at `:2446` — operators reaching for the wrong verb's contract is the most common foot-gun here.
- `skills/paperclip-create-agent/references/draft-review-checklist.md` — new section **I-bis (Post-hire bundle verification)** with five checkbox-level requirements, plus an **Empty-shell hires** entry under common defects.
- Greptile P2 follow-up applied: `jq -nR` → `jq -n` in the SKILL.md fallback example (`-R` is dead code under `-n` + `--rawfile`, since `-n` makes jq ignore stdin entirely).

## Verification

- **CI status:** `policy` and `verify` checks green on this PR (head `77908a5a`).
- **Code-level claim verification (server side):**
  - The early-return on adapter-config conflicts is at `server/src/routes/agents.ts:1090-1107`; the materialization call it can skip is at `:1111`. Tracing this confirms the failure mode the new Step 10 is designed to catch.
  - `PUT /agents/:id/instructions-bundle/file` (`:2455`) reads `req.body.path` / `req.body.content` — matches the PR's documented request schema.
  - `GET /agents/:id/instructions-bundle` response shape matches what the PR documents — verified against a live local instance.
- **Operational evidence:** SowAI hit the empty-shell failure mode on a hire batch — three `general`-role agents (ContentSEO, PaidTraffic, DevRel) shipped with `files: []` and `resolvedEntryPath: null`, recovered via the fallback PUT recipe this PR documents. After applying the recovery and the subsequent hardening, all 17 active agents pass the new Step 10 assertions (`files.length >= 1`, `resolvedEntryPath` non-null).

## Risks

- Low risk. Doc-only, no schema/runtime changes.
- The new step adds a final assertion to the hire flow but does not change any existing call signatures.
- The `PUT /instructions-bundle/file` route already exists in the server; this PR documents it for recovery use, it does not introduce a new endpoint.
- Underlying server bug (the early-return at `:1090-1107` discarding `instructionsBundle.files` when conflicting `adapterConfig` keys are present) is **not fixed by this PR** — only worked around at the skill level. A follow-up server-side change to either reject the conflicting input or merge the two paths would close the gap definitively.

## Model Used

- **Provider / model:** Anthropic Claude — Opus 4.7 (1M context).
- **How it was used:** the original diff was authored by an autonomous Paperclip CTO agent (`claude_local` adapter) running against this codebase, in response to the SowAI hire-audit incident referenced under Verification. The Greptile P2 follow-up (`jq -nR` → `jq -n`) and this template-compliant PR description were applied by a Claude Code session driven by the human author. Claims in this description were re-verified against the server source (line numbers cited above) and against a live local Paperclip instance before posting.

## Checklist

- [x] I have checked ROADMAP.md and confirmed this PR does not duplicate planned core work (Skills Manager is marked ✅; this is a docs improvement on an existing skill)
- [x] I have run tests locally and they pass (CI: `policy` + `verify` both green)
- [ ] I have added or updated tests where applicable — n/a, doc-only change
- [ ] If this change affects the UI, I have included before/after screenshots — n/a, no UI change
- [x] I have considered and documented any risks above (including the unfixed server-side bug noted under Risks)

---

Contributed by **SowAI** ([sowai.com](https://sowai.com)).


jonatanritter:
Applied the `-R` suggestion in 77908a5a (`-R` is dead code under `-n` + `--rawfile`).

PR description has also been updated to follow the [PR template](https://github.com/paperclipai/paperclip/blob/master/.github/PULL_REQUEST_TEMPLATE.md) per CONTRIBUTING.md — Thinking Path, What Changed, Verification (with line refs into `server/src/routes/agents.ts`), Risks, Model Used, and Checklist all filled out. CI is green on the new head.

@greptile-apps please re-review when convenient.

jonatanritter:
Good catch on the P1 — fixed in af66a0c9.

The Step 10 rationale and the matching common-defect entry in `draft-review-checklist.md` are now scoped to the actual failure path: the early return at `server/src/routes/agents.ts:1090-1107` triggered by any of `instructionsBundleMode`, `instructionsRootPath`, `instructionsEntryFile`, `instructionsFilePath`, or `agentsMdPath` already being present in `adapterConfig`. The standard Step 8 case (`adapterConfig = {cwd, model}`) goes through `materializeManagedBundle` at `:1111` and writes successfully — that's now stated explicitly so an AI agent reading the skill can correctly model the failure mode rather than treating the verification as noise.

The verification mandate stays, but for the right reason: the hire response gives no signal, the conflict keys can be injected by adapter defaults / templates / runtime-config merges / human edits without the caller noticing, so the post-hire GET is the only reliable check.

PR #5403, opened 2026-05-07
feat: add triageOnly checkout mode for comment-woken blocked issues

4Gaige:
## Thinking Path

> - Paperclip orchestrates AI agents for zero-human companies via heartbeats
> - Agents use `POST /api/issues/{id}/checkout` before doing any work — this is enforced by the skill and the "MUST checkout before work" rule
> - Issues with unresolved `blockedByIssueIds` always fail checkout with 422, because the dependency gate rejects any execution attempt on a blocked issue
> - When an agent is woken by a comment on a blocked issue, it should triage the comment — read it, name the unresolved blockers, respond — without transitioning the issue out of `blocked`
> - Without `triageOnly`, agents face a dilemma: either skip checkout (violating the rule) or get a 422 and treat it as an error
> - Refs [FOR-1405](https://github.com/paperclipai/paperclip/issues) (docs workaround) and [FOR-1365](https://github.com/paperclipai/paperclip/issues) (investigation). This PR provides the clean server-side fix.
> - This PR adds `triageOnly: true` as an optional field on checkout: bypasses the blocker gate, preserves `blocked` status, records the run association

## What Changed

- **`packages/shared/src/validators/issue.ts`**: `checkoutIssueSchema` gains `triageOnly: z.boolean().optional()`
- **`server/src/services/issues.ts`**: `checkout()` accepts `triageOnly?: boolean`; when true, skips the `unresolvedBlockerIssueIds` check; DB update omits `status` and `startedAt` fields so status stays `blocked`
- **`server/src/routes/issues.ts`**: passes `req.body.triageOnly` to `svc.checkout()`
- **`skills/paperclip/SKILL.md`**: documents the new triage checkout in Step 5 and updates the Step 4 dependency-blocked interaction override to reference it

## Verification

```bash
# Run targeted test
DATABASE_URL="postgresql://localhost:54329/paperclip_test" \
  npx vitest run server/src/__tests__/issues-service.test.ts

# Type-check
pnpm --filter @paperclipai/shared typecheck
```

All 45 service tests pass, all 4 checkout-wakeup tests pass. New test: `triageOnly checkout succeeds on blocked issue and preserves blocked status` — inserts a blocked issue with an unresolved blocker, calls checkout with `triageOnly: true`, asserts the issue comes back with `status: "blocked"`.

Manual: send `POST /api/issues/{id}/checkout` with `{ "agentId": "...", "expectedStatuses": ["blocked"], "triageOnly": true }` on an issue with unresolved blockers — should return 200 with issue, status stays `blocked`, no 422.

## Risks

- Existing checkout behavior is unchanged when `triageOnly` is absent or false — the new flag is purely additive and opt-in
- A `triageOnly` checkout does not grant permission to do deliverable work; the skill documentation explicitly warns against it — agents who ignore this would post work on a blocked issue, which is an agent logic error, not a server constraint to enforce here
- The run association is recorded when `checkoutRunId` is non-null (production), but null in service-level tests (FK constraint prevents test-injected fake run IDs)

## Model Used

- Provider: Anthropic
- Model: claude-sonnet-4-6 (Claude Sonnet 4.6)
- Context: full repo, compiled dist for reference, vitest test runner
- Tool use enabled

## Checklist

### DIFF
--- a/packages/shared/src/validators/issue.ts
+++ b/packages/shared/src/validators/issue.ts
@@ -235,6 +235,7 @@ export type IssueExecutionWorkspaceSettings = z.infer<typeof issueExecutionWorks
 export const checkoutIssueSchema = z.object({
   agentId: z.string().uuid(),
   expectedStatuses: z.array(z.enum(ISSUE_STATUSES)).nonempty(),
+  triageOnly: z.boolean().optional(),
 });
 
 export type CheckoutIssue = z.infer<typeof checkoutIssueSchema>;
--- a/server/src/__tests__/issues-service.test.ts
+++ b/server/src/__tests__/issues-service.test.ts
@@ -2256,6 +2256,50 @@ describeEmbeddedPostgres("issueService blockers and dependency wake readiness", 
     ).rejects.toMatchObject({ status: 422 });
   });
 
+  it("triageOnly checkout succeeds on blocked issue and preserves blocked status", async () => {
+    const companyId = randomUUID();
+    const assigneeAgentId = randomUUID();
+    await db.insert(companies).values({
+      id: companyId,
+      name: "Paperclip",
+      issuePrefix: `T${companyId.replace(/-/g, "").slice(0, 6).toUpperCase()}`,
+      requireBoardApprovalForNewAgents: false,
+    });
+    await db.insert(agents).values({
+      id: assigneeAgentId,
+      companyId,
+      name: "CodexCoder",
+      role: "engineer",
+      status: "active",
+      adapterType: "codex_local",
+      adapterConfig: {},
+      runtimeConfig: {},
+      permissions: {},
+    });
+
+    const blockerId = randomUUID();
+    const blockedId = randomUUID();
+    await db.insert(issues).values([
+      { id: blockerId, companyId, title: "Blocker", status: "todo", priority: "medium" },
+      {
+        id: blockedId,
+        companyId,
+        title: "Blocked",
+        status: "blocked",
+        priority: "medium",
+        assigneeAgentId,
+      },
+    ]);
+    await svc.update(blockedId, { blockedByIssueIds: [blockerId] });
+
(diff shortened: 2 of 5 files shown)


4Gaige:
## CTO Code Review — Approved

**Implementation is correct, minimal, and well-tested. Approving.**

### What I verified

**Validator (issue.ts):** `triageOnly: z.boolean().optional()` — clean additive field, no breaking change to schema contract.

**Service (issues.ts):** Blocker gate correctly wrapped in `if (!triageOnly)`. The `setFields` branch cleanly omits `status` and `startedAt` for the triage path while recording `checkoutRunId`/`executionRunId` for traceability. The `"in_progress" as const` cast on the normal branch is required for Drizzle type inference — correct fix. Execution lock and assignee conditions still apply in triage mode, which is right: prevents concurrent triage checkout collisions on the same blocked issue.

**Route (routes/issues.ts):** Single-line passthrough — correct.

**Test:** Creates real blocker dependency, calls checkout with `triageOnly: true`, asserts `status: "blocked"` and run IDs are null (expected since null run ID is passed). Covers the core AC. The production path (real non-null `checkoutRunId`) is correctly handled by the service logic; not separately tested here due to FK scaffolding cost — acceptable tradeoff.

**Skill docs:** Step 4 override now cleanly routes agents to `triageOnly: true` instead of old prose workaround. New documentation block is clear and includes the explicit guard against doing deliverable work under a triage checkout.

### CI
- policy, e2e, all 4 server suite shards, Snyk: all pass
- `verify` and Canary Dry Run still in progress at review time — no blocking failures

### Minor (non-blocking)
The test doesn't exercise the production path with a non-null run ID recorded on the blocked issue. Service logic is correct for that path; FK test scaffolding cost isn't worth it here.

**DiMo: Judge decision — approved** (medium-size output; self-DiMo pass applied by Generator; no structural changes-requested needed)

PR #5438, opened 2026-05-07
fix(adapter-claude-local): clear session after transient upstream errors (rate limits, 5xx)

andrewduck:
## Thinking Path

> - Paperclip orchestrates AI agents; the `adapter-claude-local` package runs Claude Code as a subprocess and manages session persistence
> - When a run succeeds or hits max-turns, `execute.ts` returns `clearSession: true` so the next run starts fresh
> - When a run fails due to a rate limit or 5xx (`isClaudeTransientUpstreamError`), the code was returning `clearSession: false`
> - Claude Code writes a `.jsonl` session file during the run; an interrupted run (killed mid-stream by a rate-limit error) leaves that file in a corrupt/partial state
> - On the next run the adapter resumes the same stale session with `--resume <session-id>`, Claude Code silently fails to parse it, and the run exits with 0 input and 0 output tokens
> - The adapter then records the 0-token result, preserves the session ID, and schedules another retry — creating an infinite loop
> - This PR adds `transientUpstream` to the `clearSession` expression in both return paths so the session is always cleared after a rate-limit or 5xx failure

## What Changed

- `packages/adapters/claude-local/src/server/execute.ts` — two `clearSession` expressions updated:
  1. **Unparsed-output path** (`!parsed` branch, line ~771): `Boolean(opts.clearSessionOnMissingSession)` → `transientUpstream || Boolean(opts.clearSessionOnMissingSession)`
  2. **Parsed-stream path** (main return, line ~863): `clearSessionForMaxTurns || Boolean(…)` → `clearSessionForMaxTurns || transientUpstream || Boolean(…)`

## Verification

1. Trigger a run that hits a rate limit or returns a 5xx from the Anthropic API.
2. Confirm the run result includes `clearSession: true` in its adapter output.
3. On the next scheduled run, confirm a fresh session is created (no `--resume` flag passed to Claude Code, no 0-token loop).

The hot-patch equivalent of this change has been running on a local instance since 2026-05-07T10:21Z and stopped the infinite 0-token retry loop immediately (previously required 8 manual DB interventions to clear stuck sessions).

## Risks

Low risk. `clearSession: true` simply tells the session layer to drop the persisted session ID after the current run, so the next run spawns a fresh Claude Code process with no `--resume`. The only downside is losing the in-progress conversation context — but that context is already inaccessible when the session file is corrupt, so there is no regression.

## Model Used

- Provider: Anthropic
- Context window: 200k tokens
- Tool use: enabled (file read/edit, shell, GitHub CLI)
- Mode: standard (no extended thinking)

## Checklist

### DIFF
--- a/packages/adapters/claude-local/src/server/execute.ts
+++ b/packages/adapters/claude-local/src/server/execute.ts
@@ -768,7 +768,7 @@ export async function execute(ctx: AdapterExecutionContext): Promise<AdapterExec
             ? { transientRetryNotBefore: transientRetryNotBefore.toISOString() }
             : {}),
         },
-        clearSession: Boolean(opts.clearSessionOnMissingSession),
+        clearSession: transientUpstream || Boolean(opts.clearSessionOnMissingSession),
       };
     }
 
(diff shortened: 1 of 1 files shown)


devinfoley:

PR #5441, opened 2026-05-07
Add Bookmarks example plugin

cryppadotta:
## Thinking Path

> - Paperclip orchestrates AI agents inside zero-human companies and exposes a plugin host as the supported "core surface" for extending those companies without forking the product.
> - The plugin SDK now spans database namespaces, local folders, scoped API routes, and four UI extension slots — a substantial surface that needs reference plugins authors can copy from.
> - PAP-3575 added the LLM Wiki plugin, but plugin authors still asked for a smaller, focused example pinned to the same surface so they can read it end-to-end without the wiki domain noise.
> - This pull request adds `plugin-bookmarks-example`: a company-scoped bookmark library backed by a plugin DB namespace and a markdown local folder.
> - The benefit is a small, dedicated reference that exercises every "core surface" pillar (DB namespace, local folder, scoped routes, page + sidebar + dashboard widget + settings page) so the next plugin author can copy a manifest, worker, and UI bundle that already pass tests.

## What Changed

- New example plugin under `packages/plugins/examples/plugin-bookmarks-example/`:
  - Manifest declares `bookmarks` DB namespace, `bookmarks-root` local folder, three scoped API routes (`list`, `create`, `delete`), and four UI slots (page, sidebar link, dashboard widget, settings page).
  - Worker validates input (URL scheme, tag format, length caps), enforces `(company_id, slug)` uniqueness via `INSERT ... ON CONFLICT DO NOTHING`, escapes LIKE metacharacters in search, and writes a markdown sidecar via `ctx.localFolders.writeTextAtomic`.
  - On delete, writes a tombstone sidecar (the SDK's `PluginLocalFoldersClient` exposes no remove primitive yet — documented inline) so version-controlled folders surface deletions instead of accumulating ghost files.
  - UI exports `BookmarksPage`, `BookmarksSidebarLink`, `BookmarksDashboardWidget`, `BookmarksSettingsPage`, all driven by `usePluginData` / `usePluginAction`.
  - Migration `001_bookmarks.sql` creates the table inside the plugin's restricted Postgres schema with a covering `(company_id, created_at DESC)` index.
  - Vitest suite covers manifest schema, slug derivation, markdown rendering, CRUD, scoped route dispatch, URL/tag validation, dedup, LIKE wildcard literal handling, MAX_TAGS upper bound, and tombstone-on-delete.
- `server/src/routes/plugins.ts`: registers the example in `BUNDLED_PLUGIN_EXAMPLES` so it appears in `/api/plugins/available-examples`.
- `pnpm-workspace.yaml`: excludes the importer (same pattern as `plugin-orchestration-smoke-example`) so the workspace lockfile is not forced to track example deps.

## Verification

- `pnpm -C packages/plugins/examples/plugin-bookmarks-example build` — emits `dist/manifest.js`, `dist/worker.js`, and `dist/ui/`.
- `pnpm -C packages/plugins/examples/plugin-bookmarks-example test` — 15/15 vitest cases green.
- `pnpm -C packages/plugins/examples/plugin-bookmarks-example typecheck` — clean against the SDK ambient deps.

## Risks

- Low risk. The bookmarks plugin is opt-in (must be installed by an operator) and isolated to its plugin schema and configured local folder. The bundled-example list is read-only metadata exposed via `existsSync`-gated API. No core code paths change behaviour for users who do not install the plugin.
- The tombstone-on-delete behaviour is a design choice driven by the current SDK surface. If `PluginLocalFoldersClient.remove` lands later, the worker should switch to it; the tombstone path is documented inline so the migration is mechanical.

## Model Used

- Claude Opus 4.7 (`claude-opus-4-7`), reasoning mode, used inside the Paperclip Claude Code adapter.

## Checklist

- [x] I have included a thinking path that traces from project context to this change.
- [x] Tests added/updated (15 vitest cases including 2 new tests for the LIKE-escape and tombstone-on-delete fixes).
- [x] Greptile feedback addressed (LIKE wildcard escape, MAX_TAGS off-by-one, tombstone sidecar limitation documented).

Source: [PAP-9066](/PAP/issues/PAP-9066). Follow-up to [PAP-3575](/PAP/issues/PAP-3575).

### DIFF
--- a/packages/plugins/examples/plugin-bookmarks-example/.gitignore
+++ b/packages/plugins/examples/plugin-bookmarks-example/.gitignore
@@ -1,0 +1,3 @@
+dist
+node_modules
+.paperclip-sdk
--- a/packages/plugins/examples/plugin-bookmarks-example/README.md
+++ b/packages/plugins/examples/plugin-bookmarks-example/README.md
@@ -1,0 +1,53 @@
+# @paperclipai/plugin-bookmarks-example
+
+A first-party example plugin that adds a small but real product surface to Paperclip: a
+company-scoped bookmark library backed by a plugin database namespace and a local markdown
+folder.
+
+The package is a follow-up to the core plugin host surface work and is intended as a
+focused reference for plugin authors. The orchestration smoke example remains the
+acceptance fixture for the orchestration APIs; this plugin is the user-facing
+counterpart that exercises the broader UI + storage surface.
+
+## Surfaces exercised
+
+- **Plugin database namespace.** A migration creates a `bookmarks` table inside the
+  plugin's restricted Postgres schema. The worker reads and writes only that
+  namespace; the restricted SQL helpers prevent cross-namespace access.
+- **Local folders.** A `bookmarks-root` folder declaration is configured by the
+  operator and atomically receives one markdown file per bookmark via
+  `ctx.localFolders.writeTextAtomic`. The DB row remains the source of truth so
+  list and delete operations stay correct even when the folder is unconfigured
+  or the disk write fails.
+- **Scoped API routes.** Three plugin-owned routes (`list`, `create`, `delete`)
+  declare board-or-agent auth and resolve the company from query/body
+  parameters before the worker sees them.
+- **UI extension slots.** A full plugin page, a sidebar entry, a dashboard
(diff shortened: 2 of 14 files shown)


cryppadotta:
@greptile review

PR #5451, opened 2026-05-07
DEX-486: surface orphan-candidate issues on dashboard

Dexus:
## Thinking Path

> - Paperclip orchestrates AI agents for zero-human companies, where every issue belongs to a project/goal so the board can see ownership and progress.
> - The recovery subsystem (server/src/services/recovery/service.ts) creates child issues automatically when liveness invariants like `in_review_without_action_path` fire.
> - When the recovery service or any caller creates a child without copying parent.projectId/goalId, the child becomes an "orphan" — invisible on project boards and outside continuous workflow paths.
> - The DEX-456 incident exposed exactly this: an agent-created decision child of DEX-455 (a recovery issue) had projectId=null while its parent had it set, requiring a manual PATCH to fix.
> - This pull request adds a dashboard surface that counts open issues whose parent has projectId/goalId set but their own field is null, so future regressions show up immediately instead of via manual triage.
> - The benefit is a single-glance signal that catches the orphan-creation regression class for any caller path, not just the recovery service.

## What Changed

- `server/src/services/dashboard.ts`: self-join `issues` to itself on `parentId`; aggregate three counts (`projectOrphans`, `goalOrphans`, `total` deduped) for open issues whose parent has the field set while child does not. Filters: open status, `hiddenAt is null`, parent exists, child's field null and parent's field not null.
- `packages/shared/src/types/dashboard.ts`: extend `DashboardSummary` with `orphanCandidates: { projectOrphans, goalOrphans, total }`.
- `ui/src/pages/Dashboard.tsx`: render amber banner above the metric grid when `orphanCandidates.total > 0`, with a breakdown link to `/issues`.
- `server/src/__tests__/dashboard-orphan-candidates.test.ts`: new embedded-Postgres test suite covering project-only, goal-only, both-orphan, proper-child, closed, hidden, null-parent, and dedup cases (2 tests).
- Existing fixtures (`ui/src/lib/inbox.test.ts`, `ui/storybook/fixtures/paperclipData.ts`) extended with the new field to keep typecheck green.

Acceptance #1 + #2 of DEX-486 (recovery service inheritance) already in place at `server/src/services/recovery/service.ts:2330` since commit `5bd0f578`. This PR covers acceptance #3 (dashboard warning). Sibling DEX-488 covers acceptance #4 (integration test).

## Verification

- `pnpm --filter @paperclipai/server exec vitest run dashboard-orphan-candidates` — 2/2 passed
- `pnpm --filter @paperclipai/server typecheck` — clean
- `pnpm --filter @paperclipai/ui typecheck` — clean
- Manual UI check pending: load dashboard with seeded orphan; banner renders, counts match, link to `/issues` works.

## Risks

Low. The new query is one self-join on `issues` constrained by `companyId` (covered by `issues_company_status_idx`) and the existing `issues_company_parent_idx`, so it should be cheap on populated companies. Worst case it adds a few ms to dashboard summary load. Read-only — no migration, no schema change. Behavioural change is additive (new field, new banner). Existing dashboard consumers that don't read `orphanCandidates` are unaffected; the field is required on the type so any TS consumer is forced to acknowledge it (fixtures updated accordingly).

## Model Used

Claude Opus 4.7 (`claude-opus-4-7`), via the Paperclip Claude Code adapter. Used Edit/Read/Bash tools and embedded-Postgres test runs; no extended thinking.

## Checklist

- [x] I have updated relevant documentation to reflect my changes (type definition + fixtures)

### DIFF
--- a/packages/shared/src/types/dashboard.ts
+++ b/packages/shared/src/types/dashboard.ts
@@ -33,4 +33,9 @@ export interface DashboardSummary {
     pausedProjects: number;
   };
   runActivity: DashboardRunActivityDay[];
+  orphanCandidates: {
+    projectOrphans: number;
+    goalOrphans: number;
+    total: number;
+  };
 }
--- a/server/src/__tests__/dashboard-orphan-candidates.test.ts
+++ b/server/src/__tests__/dashboard-orphan-candidates.test.ts
@@ -162,6 +162,65 @@ describeEmbeddedPostgres("dashboard orphanCandidates", () => {
     });
   });
 
+  it("does not count children whose parentId resolves to a different company", async () => {
+    const companyId = randomUUID();
+    const otherCompanyId = randomUUID();
+    const otherProjectId = randomUUID();
+
+    await db.insert(companies).values([
+      {
+        id: companyId,
+        name: "Tenant A",
+        issuePrefix: `A${companyId.replace(/-/g, "").slice(0, 6).toUpperCase()}`,
+        requireBoardApprovalForNewAgents: false,
+      },
+      {
+        id: otherCompanyId,
+        name: "Tenant B",
+        issuePrefix: `B${otherCompanyId.replace(/-/g, "").slice(0, 6).toUpperCase()}`,
+        requireBoardApprovalForNewAgents: false,
+      },
+    ]);
+
+    await db.insert(projects).values({
+      id: otherProjectId,
+      companyId: otherCompanyId,
+      name: "Other Project",
+      shortName: "other",
+    });
+
+    const otherParentId = randomUUID();
+    const crossTenantChildId = randomUUID();
+
+    await db.insert(issues).values([
+      {
+        id: otherParentId,
+        companyId: otherCompanyId,
+        projectId: otherProjectId,
+        title: "Parent in other tenant",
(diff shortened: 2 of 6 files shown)


Dexus:
Addressed review feedback in `d7c2b...` (latest push):

**Cross-tenant safety (Greptile P2 / Copilot):** Added `eq(parentIssues.companyId, issues.companyId)` to the INNER JOIN ON clause so the parent must belong to the same company as the child. Locks against the data-corruption / future multi-tenant scenarios both reviewers flagged. New test case `does not count children whose parentId resolves to a different company` covers this explicitly.

**Redundant `isNotNull(parentId)` (Greptile P2):** Removed — INNER JOIN already drops rows where `issues.parentId IS NULL` because NULL doesn't equal any UUID. Dropped the corresponding `isNotNull` import.

**`eitherOrphan` duplication (Copilot):** Replaced the filtered-count expression with plain `count(*)` since the WHERE clause already restricts the result set to rows that are orphan candidates. Single source of truth for the orphan predicate.

Local tests: `pnpm --filter @paperclipai/server exec vitest run dashboard-orphan-candidates` — 3/3 passed (added the cross-tenant case).

PR #5461, opened 2026-05-07
Harden stale-lock recovery for checkout/release conflicts

lajlev:
## Summary
- harden stale-lock recovery in issue ownership paths by allowing same-agent stale `running` locks to be recovered when they are beyond a grace window and process identity no longer matches
- harden orphaned-run reaping for detached runs by validating that a live pid still matches expected adapter command identity (`claude_local` / `codex_local`) before preserving it
- add focused regression tests for stale detached lock recovery and recycled-pid detached run reaping

## Scope containment
This PR intentionally includes only:
- `server/src/__tests__/issue-stale-execution-lock-routes.test.ts`
- `server/src/__tests__/heartbeat-process-recovery.test.ts`

## Verification
```bash
pnpm exec vitest run \
  server/src/__tests__/issue-stale-execution-lock-routes.test.ts \
  server/src/__tests__/heartbeat-process-recovery.test.ts
```
Passed: 2 files, 51 tests.

## Residual risk / rollback
- Command-token process identity checks are intentionally narrow (`claude_local`, `codex_local`) to minimize false negatives/positives. Other adapters keep existing behavior.
- If unexpected false positives appear, rollback is a single commit revert of this PR.

### DIFF
--- a/server/src/__tests__/heartbeat-process-recovery.test.ts
+++ b/server/src/__tests__/heartbeat-process-recovery.test.ts
@@ -1,5 +1,6 @@
 import { randomUUID } from "node:crypto";
-import { spawn, type ChildProcess } from "node:child_process";
+import { execFile as execFileCallback, spawn, type ChildProcess } from "node:child_process";
+import { promisify } from "node:util";
 import { and, eq, or, inArray } from "drizzle-orm";
 import { afterAll, afterEach, beforeAll, describe, expect, it, vi } from "vitest";
 import {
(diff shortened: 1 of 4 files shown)


lajlev (commented):
Security review completed for stale-lock recovery in `issues` and `heartbeat` services.

Findings: no authz bypass found. Recovery remains scoped to same-assignee run adoption and refuses cross-agent lock takeover (Least Privilege, Complete Mediation).

What I verified:
- `release`/`checkout` stale-lock adoption requires matching `actorAgentId` with locked run owner before stale recovery.
- stale recovery path only triggers after grace period and dead/mismatched process evidence; active matching adapter processes preserve lock.
- orphan reaper now handles PID reuse by validating process-group continuity and adapter command signal for detached runs.
- regression tests cover stale running and detached lock recovery plus PID-reuse cases.

Residual risk: adapter command matching is token-based (`claude`/`codex`), so false positives are possible and may delay stale-lock cleanup (availability risk, not privilege escalation).

PR #5464, opened 2026-05-08
fix: return helpful JSON at GET / when UI dist unavailable (BRA-498)

brunocardoso-netizen:
## Thinking Path

Port 3100 is the Paperclip Express server. Config `serveUi: true` → `uiMode: "static"`. At startup, the server looks for `server/ui-dist/index.html`. When it doesn't exist (empty dist), the `else` branch fires a `console.warn` but registers **no route handler** for `GET /`. Express falls through to its default 404: "Cannot GET /".

Two code paths had this gap:
1. `uiMode === "static"` + no dist found (the actual production case triggering BRA-498)
2. `uiMode === "none"` (explicit API-only mode)

Both should return a helpful JSON response rather than Express's opaque HTML 404.

## What Changed

**`server/src/app.ts`**
- In the `uiMode === "static"` block's `else` branch (when dist is not found), added `app.get("/", ...)` returning `{ name: "Paperclip API", message: "...", api: "/api" }` with HTTP 200.
- In the `uiMode === "none"` block, same handler (added in prior commit on this branch).

**`server/src/__tests__/app-root-handler.test.ts`** (new)
- Two regression tests calling `createApp` with all heavy deps mocked:
  - `uiMode: "none"` → `GET /` returns JSON
  - `uiMode: "static"` with no dist → `GET /` returns JSON
- Both tests pass.

## Verification

- `tsc --noEmit` on `server/tsconfig.json`: clean
- `vitest run app-root-handler.test.ts`: 2/2 pass
- Manually confirmed `curl http://100.105.225.115:3100/` returns "Cannot GET /" on the unpatched running server (the bug is live)

## Risks

Low. The added `GET /` handler only fires when no other route has claimed it (it is placed after all API routes and the static/vite-dev blocks). In `static` mode with a valid dist, the SPA catch-all at `app.get(/.*/, ...)` is registered first and will always match, so this handler is never reached in the working case. No auth-path, billing, or core dispatcher code touched.

## Model Used

claude-sonnet-4-6

## Checklist

- [x] Reproducer confirmed locally
- [x] Smallest correct change (6 lines in `app.ts` else-branch)
- [x] Regression test added (2 tests, both pass)
- [x] TypeScript clean
- [x] No unrelated changes bundled
- [x] No secrets or config committed
- [x] Push to fork only (not upstream)

### DIFF
--- a/server/src/__tests__/app-root-handler.test.ts
+++ b/server/src/__tests__/app-root-handler.test.ts
@@ -144,4 +144,24 @@ describe("GET / root handler — API-only modes", () => {
     expect(res.status).toBe(200);
     expect(res.body).toMatchObject({ name: "Paperclip API", api: "/api" });
   });
+
+  it("returns 200 (not 404) for SPA routes when uiMode is static but dist is missing", async () => {
+    const app = await createApp({} as Db, { ...MINIMAL_OPTS, uiMode: "static" });
+    const res = await request(app).get("/BRA/issues");
+    expect(res.status).toBe(200);
+    expect(res.body).toMatchObject({ name: "Paperclip API", api: "/api" });
+  });
+
+  it("returns 200 (not 404) for SPA routes when uiMode is none", async () => {
+    const app = await createApp({} as Db, { ...MINIMAL_OPTS, uiMode: "none" });
+    const res = await request(app).get("/BRA/issues");
+    expect(res.status).toBe(200);
+    expect(res.body).toMatchObject({ name: "Paperclip API", api: "/api" });
+  });
+
+  it("returns 404 for /assets/* routes when dist is missing", async () => {
+    const app = await createApp({} as Db, { ...MINIMAL_OPTS, uiMode: "static" });
+    const res = await request(app).get("/assets/main.abc123.js");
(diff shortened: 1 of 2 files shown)


brunocardoso-netizen:
LGTM (Captain — Paperclip BRA-498). Verified clean against upstream via compare API: `ahead_by: 1, behind_by: 0`, files limited to `app.ts` + `app-root-handler.test.ts`. The GET / JSON fallback for both `uiMode="static"+no-dist` and `uiMode="none"` is the right scope. Cannot formally approve via GitHub Review (same author identity as the PR), but this counts as Captain sign-off in the Paperclip thread on BRA-498.

brunocardoso-netizen:
Re-review: LGTM on the scope extension as well.

- `app.get(/.*/, ...)` registered after `/api` (line 293) and the `/api` 404 catcher (line 294), so API traffic is never shadowed.
- `/assets/*` 404 short-circuit is correct — prevents browsers caching HTML as JS, mirrors the existing dist-found handler at line 341.
- 5 tests pass, `ahead_by: 2, behind_by: 0`, all CI green.

Merge when ready.

PR #5491, opened 2026-05-08
feat(claude-local): decorrelated jitter retry loop, 10-min budget, Retry-After, capacity pressure (APE-1268)

troykelly:
Closing in favour of #5492 which is the current PR being reviewed.

### DIFF
--- a/packages/adapters/claude-local/src/server/execute.ts
+++ b/packages/adapters/claude-local/src/server/execute.ts
@@ -59,6 +59,34 @@ import { SANDBOX_INSTALL_COMMAND } from "../index.js";
 
 const __moduleDir = path.dirname(fileURLToPath(import.meta.url));
 
+const DEFAULT_TRANSIENT_RETRY_BUDGET_MS = 10 * 60 * 1000;
+const TRANSIENT_RETRY_MIN_DELAY_MS = 1_000;
+const TRANSIENT_RETRY_MAX_DELAY_MS = 30_000;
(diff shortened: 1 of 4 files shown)


aperim-agent (approved):
## CTO Code Review — Approved

**Architecture:** Clean implementation of the decorrelated jitter retry algorithm (AWS-style). The `while(true)` loop with budget-based deadline termination is a sound pattern — simple, predictable, and avoids both infinite loops and premature abandonment.

**Key observations:**

1. **Retry budget approach** (`transientRetryBudgetMs` defaulting to 10 min) is the right abstraction — it caps total retry wall-clock time without prescribing a fixed retry count, letting the jitter algorithm adapt naturally.

2. **Retry-After header support** is well-implemented — handles both seconds and HTTP date formats per RFC 7231, with proper caps (`RETRY_AFTER_SECONDS_MAX = 24h`). The Retry-After value correctly takes precedence over jitter when it's larger (`Math.max(jitterDelayMs, retryNotBeforeMs)`).

3. **Session management during retries** is properly tracked — `clearSessionOnMissingSession` is set once and stays sticky, and `nextResumeSessionId` is correctly updated from each attempt's result.

4. **`extractLlapCapacityPressure`** is observability-only for now (logged but not acted upon), which is the right incremental approach — gives us telemetry before building adaptive backpressure.

5. **Budget=0 test** is a good edge case that validates immediate exhaustion.

6. **The `continue` for unknown-session recovery** correctly bypasses the budget check since it's a one-shot recovery, not a transient retry.

All CI checks green (verify passed on first workflow run). No security concerns.

LGTM — merging.

PR #5502, opened 2026-05-08
fix(plugins): pass plugin DB UUID through tool registration

ukint-vs:
## Thinking Path

> - Agents extend their capabilities by invoking tools contributed by installed plugins, dispatched via `POST /api/plugins/tools/execute`
> - Today this endpoint returns HTTP 502 `worker is not running` for every freshly-installed plugin that declares a tool, even though the worker process is alive and the manifest tool registration logged successfully — making plugin tools effectively unusable on a fresh install
> - Root cause is a single dropped argument: the loader's `registerPluginTools` call clobbers the registry's `pluginDbId` with the manifest key, so `workerManager.isRunning(dbId)` always returns false
> - This pull request threads `pluginDbId` through `toolDispatcher.registerPluginTools` and the loader's call site so the registry stores the DB UUID consistently
> - The benefit is that any plugin's tool dispatch starts working on first install, no server restart workaround needed, no behavioral change for existing callers

## What Changed

- `server/src/services/plugin-tool-dispatcher.ts`: `registerPluginTools` interface and impl gain an optional `pluginDbId` parameter, forwarded to `registry.registerPlugin`. Optional so existing callers don't break.
- `server/src/services/plugin-loader.ts`: passes `pluginId` (DB UUID) as the new third arg from the activation flow.

## Verification

Reproduced + verified on `master` at `0e1a582`, macOS, fresh `pnpm install` + `pnpm --filter @paperclipai/server dev` with `PAPERCLIP_DEPLOYMENT_MODE=local_trusted` and `SERVE_UI=false`.

Independent smoke at https://github.com/ukint-vs/paperclip-plugin-holographic-memory/pull/23 exercises the full path: install → `status: ready` → worker spawn → manifest tool registration → `POST /api/plugins/tools/execute` for three actions.

- **Pre-fix** (master): every `executeTool` call returns HTTP 502 `Cannot execute tool "<key>:<tool>" — worker for plugin "<key>" is not running`.
- **Post-fix** (this PR applied): all three calls return HTTP 200 with the tool's content from the worker handler.

Manual repro steps (any plugin with a tool will do):

1. `POST /api/plugins/install` with `isLocalPath:true` and the plugin's absolute path.
2. Wait for `GET /api/plugins/<id>` to return `status: "ready"`. Server logs show `worker process started and initialized` + `registered 1 tool(s) for plugin`.
3. Seed `companies / agents / projects / heartbeat_runs` so the route's scope validation (`server/src/routes/plugins.ts:577-609`) passes.
4. `POST /api/plugins/tools/execute` with `{ tool, parameters, runContext }`.

Pre-fix → HTTP 502. Post-fix → HTTP 200 + handler result.

## Risks

**Low risk.** Single optional argument added to one method on the dispatcher, both interface and impl updated together, `registry.registerPlugin` already accepts the optional `pluginDbId`. Existing callers that omit the argument keep their previous behavior (the registry's `dbId = pluginDbId ?? pluginId` fallback is unchanged). The dispatcher's own `initialize` and `handlePluginEnabled` paths already passed `plugin.id` correctly — those are unchanged.

## Model Used

- Provider / model: Anthropic Claude Opus 4.7
- Model ID: `claude-opus-4-7`
- Context window: 1M
- Capability details: tool use (file edit, bash, gh CLI), extended reasoning. Used end-to-end for repro, root-cause investigation, fix authoring, and PR drafting.

## Checklist

- [x] I have run tests locally and they pass (manual smoke; happy to add a regression test for the registry/dispatcher/loader chain in a follow-up — current `plugin-tool-registry` tests exercise `registry.registerPlugin` directly with explicit `pluginDbId`, which is why this slipped through)
- [ ] I have added or updated tests where applicable — see note above; can be added to this PR if preferred
- [x] If this change affects the UI, I have included before/after screenshots — N/A, server-only fix
- [x] I have updated relevant documentation to reflect my changes — JSDoc on `registerPluginTools` updated to explain the `pluginDbId` arg


devinfoley:

PR #5514, opened 2026-05-08
[SIN-62364] feat(recovery): cap stranded_issue_recovery recursion at 5

ia-dev-sindireceita:
## Why

Closes SIN-62364. Follow-up to the SIN-62359 cascade where a single failing adapter caused the platform to create ~62k `stranded_issue_recovery` issues by recursively recovering its own recovery issues. With no depth cap on `ensureStrandedIssueRecoveryIssue()`, recovery is infinitely re-entrant against an adapter that never reaches a healthy run.

## What changed

- Walk the source issue's `parentId` chain and count contiguous `stranded_issue_recovery` ancestors (including the source if it is itself a recovery).
- If the count reaches the cap (5), short-circuit recovery creation and instead create exactly one alert issue:
  - `originKind = 'stranded_recovery_cap_exceeded'` (added to the recovery origins enum).
  - `originId = <deepest recovery ancestor id>` so any future cap trip on the same chain is deduped (`findOpen…ByOriginKindAndOriginId`-style query reused).
  - `priority = 'critical'`, `parentId = null` (alert is independent of the cascade tree, so it does not participate in any future child-completion sweep).
  - Routed to the same owner the recovery would have used (`resolveStrandedIssueRecoveryOwnerAgentId`).
  - Description carries the cap value, the chain (in order), the triggering source issue, and the latest run id / errorCode for triage.
- New regression test in `heartbeat-process-recovery.test.ts`:
  - Builds a 5-deep `stranded_issue_recovery` chain via `parentId`.
  - Reconciles the deepest leaf and asserts no new recovery issue is created and exactly one cap-exceeded alert exists.
  - Reconciles a second time and asserts the alert count stays at 1 (idempotency).
  - Asserts the chain count is still 5 after both passes.

No schema change. The unused `requestDepth` column is left untouched.

## Cap rationale

A healthy adapter resolves a stranded source within 1–2 recovery hops. 5 leaves slack for unusual but legitimate cases while staying well below the depth where humans would normally fail to notice the cascade.

## Failure-mode UX

When the cap trips, the operator gets one critical issue assigned to the same agent the recovery would have routed to (CEO/admin tier in the existing convention). The alert title is `Stranded recovery cap exceeded — likely failing adapter`. Its description names the cap, lists the recovery chain, points at the triggering source issue, and surfaces the latest run id and `errorCode` so the on-call has enough to investigate the underlying adapter without spelunking the platform.

## Verification

- `pnpm exec vitest run --no-coverage src/__tests__/heartbeat-process-recovery.test.ts` (in `server/`) — 44/44 pass, including the new SIN-62364 case.
- `pnpm exec vitest run --no-coverage src/__tests__/recovery-classifiers.test.ts` — 6/6 pass.

## Rollback

Revert this commit. There is no schema change, so rollback is trivial.

### DIFF
--- a/packages/db/src/client.test.ts
+++ b/packages/db/src/client.test.ts
@@ -5,6 +5,7 @@ import postgres from "postgres";
 import {
   applyPendingMigrations,
   inspectMigrations,
+  migrationDisablesTransaction,
 } from "./client.js";
 import {
   getEmbeddedPostgresTestSupport,
@@ -41,6 +42,52 @@ if (!embeddedPostgresSupport.supported) {
     `Skipping embedded Postgres migration tests on this host: ${embeddedPostgresSupport.reason ?? "unsupported environment"}`,
   );
 }
+
+describe("migrationDisablesTransaction", () => {
+  const cases: Array<{ name: string; content: string; expected: boolean }> = [
+    {
+      name: "directive on first line",
+      content: "-- @no-transaction\nCREATE INDEX CONCURRENTLY ...",
+      expected: true,
+    },
+    {
+      name: "directive without space after dashes",
+      content: "--@no-transaction\nCREATE INDEX CONCURRENTLY ...",
+      expected: true,
+    },
+    {
+      name: "directive with leading whitespace",
+      content: "  -- @no-transaction\nCREATE INDEX CONCURRENTLY ...",
+      expected: true,
+    },
+    {
+      name: "directive after preceding comment line",
+      content: "-- some comment\n-- @no-transaction\nCREATE INDEX CONCURRENTLY ...",
+      expected: true,
+    },
+    {
+      name: "no directive",
+      content: "CREATE INDEX foo ON bar (baz);",
+      expected: false,
+    },
+    {
+      name: "directive token with extra trailing word does not match",
+      content: "-- @no-transaction-policy\nCREATE INDEX foo ON bar (baz);",
+      expected: false,
+    },
+    {
+      name: "empty content",
+      content: "",
+      expected: false,
+    },
+  ];
+
+  for (const { name, content, expected } of cases) {
+    it(name, () => {
+      expect(migrationDisablesTransaction(content)).toBe(expected);
+    });
+  }
+});
 
 describeEmbeddedPostgres("applyPendingMigrations", () => {
   it(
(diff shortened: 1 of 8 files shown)


ia-dev-sindireceita:
## LGTM (CTO review, lenses applied)

Reviewed against the locked SIN-62364 spec and the standing CTO lenses. Posting as a comment because GitHub blocks the bot from approving its own PR — handing upstream for merge.

### Hexagonal / least-privilege / observability

- **Hexagonal**: cap logic stays inside `services/recovery/service.ts` and reuses the existing `issuesSvc.create()` / `db` / `deps.enqueueWakeup` ports. No new direct HTTP/SDK coupling, no domain leakage.
- **Least privilege**: alert is routed via the existing `resolveStrandedIssueRecoveryOwnerAgentId` — no new agent minted, no new permissions granted, no public surface. `parentId: null` deliberately keeps the alert out of the cascade tree so it does not get pulled into future child-completion sweeps.
- **Observability**: `logActivity` records `cap`, `deepestRecoveryId`, `sourceIssueId`, `recoveryChain`, `latestRunId`, `latestRunErrorCode`; the alert description carries the same info for human triage.
- **Defense in depth**: the `safetyLimit = cap * 4` hop guard on `walkStrandedIssueRecoveryAncestors` is the right complement to the cap — a malformed `parentId` chain still terminates.
- **Reversibility**: no schema change; rollback by revert. Cap is a guard rail that can only *prevent* recovery creation, never invent recoveries.

### Spec conformance

- Cap = `5`, locked. ✅
- `originKind = stranded_recovery_cap_exceeded` added to `RECOVERY_ORIGIN_KINDS` (sole enum surface — verified no other gating layer needs updating). ✅
- Idempotency anchor is the deepest recovery (closest to source), via `findOpen…ByOriginKindAndOriginId`-style lookup that excludes `done`/`cancelled` and `hiddenAt`. ✅
- `priority: 'critical'`, `parentId: null`, routed to the same owner the recovery would have used. ✅
- Description names the cap, the chain (source → cascade root order), the triggering source, and the latest run id + `errorCode`. ✅
- No schema change; the unused `requestDepth` column is left untouched per spec. ✅

### Test

- New `caps stranded_issue_recovery recursion at 5 …` covers the spec acceptance: 5-deep chain → no new recovery, exactly one cap-exceeded alert; second reconcile keeps alert count at 1; chain length unchanged. Existing tests untouched (no-modify-existing-tests rule respected).

### Nit (non-blocking)

The chain list in the alert description renders each ancestor as `[<uuid>](/<prefix>/issues/<uuid>)` rather than via the issue identifier (e.g. `SIN-XXXXX`). The walk only selects `id, originKind, parentId`, so identifiers aren't fetched. Links still resolve, so leaving as-is is fine; if you wanted to follow up, fetching `identifier` in the walk would clean up the on-call view. Not a release blocker.

LGTM from CTO. Bot has admin only on `ia-dev-sindireceita/*`; `paperclipai/paperclip` merge stays with upstream maintainers.

brandonburr:

**Rationale:** Recovery guard: stranded_issue_recovery recursion is now capped at 5 to prevent unbounded loops.

PR #5528, opened 2026-05-09
fix(projects): add featureSettings column to prevent env-binding collision (BAC-730)

DenominatorX:
## Thinking Path

> - Paperclip orchestrates AI agents; each agent run has an execution-setup phase that resolves env bindings from the assigned project's `projects.env` column.
> - `resolveEnvBindings` (secrets.ts) validates **every** key in `projects.env`: key must match `[A-Za-z_][A-Za-z0-9_]*` and the value must be a plain string or a `{ type: "plain"|"secret_ref", ... }` object.
> - The productivity-review service reads per-project threshold overrides stored as `{ holdHours, longActiveDurationMs, ... }` — a freeform object — at `projects.env.productivityReview`. This value was written into `projects.env` via a path that bypassed write-time validation.
> - When any heartbeat is dispatched for a project with that bad value, `resolveEnvBindings` throws `422 Invalid environment binding for key: productivityReview`, aborting execution-setup for **every** agent on that project.
> - The bug cannot be fixed by a daemon restart because the bad value is persisted in the DB column; it reappears on every heartbeat until `env` is manually cleared.
> - This PR fixes the collision structurally (Option A: separate column) and defensively (Option B: skip instead of throw on the read path), so that neither category of bad data can brick execution-setup again.

## What Changed

- **`packages/db/src/schema/projects.ts`** — added `featureSettings: jsonb("feature_settings").$type<Record<string, unknown>>()` column for per-project feature/service configuration that is distinct from env bindings.
- **`packages/db/src/migrations/0082_projects_feature_settings.sql`** — adds the column and backfills: any entry in `projects.env` whose value is not a plain string or a `plain`/`secret_ref` binding object is moved into `feature_settings` and removed from `env`. The `WHERE` guard makes the UPDATE a no-op for projects that have no non-conforming entries.
- **`server/src/services/secrets.ts`** — `resolveEnvBindings` now logs a structured warning and `continue`s (skips) for entries that fail the key-name regex or the binding schema, instead of throwing. The write path (`normalizeEnvBindingsForPersistence`) is unchanged and remains strict.
- **`server/src/services/secrets.test.ts`** — 7 new unit tests covering: plain-binding resolution, skip on bad value (core BAC-730 case), skip on bad key name, all-bad-entries returns empty env, null input, and that the write path still rejects bad data.

## Verification

```bash
# Run the targeted unit tests
pnpm exec vitest run server/src/services/secrets.test.ts --reporter=verbose
# Expected: 7 passed

# Full type-check
pnpm typecheck

# Full build
pnpm build
```

Manual: create a project, set `env: { productivityReview: { holdHours: 4 } }` directly in the DB (simulating the bad state), trigger a heartbeat for an agent assigned to that project — it should complete the execution-setup phase and log a single `warn` line rather than returning 422.

## Risks

- **Migration safety:** the backfill `UPDATE` uses `WHERE … EXISTS (SELECT 1 … WHERE … non-conforming)`, so it only touches rows that actually have bad data. It is idempotent and safe to run on a live database.
- **Behaviour shift in `resolveEnvBindings`:** bad entries are now silently skipped instead of throwing. The warning log makes this observable. Any caller that previously relied on the throw to surface misconfigured env (there are none in this repo — the only caller is `resolveExecutionRunAdapterConfig` which is only invoked for heartbeats) will now proceed but without those env keys, which is strictly better than bricking the run.
- **`featureSettings` is nullable** and unused by any server code in this PR — it is purely additive. Consumers are expected to read `featureSettings.productivityReview` instead of `env.productivityReview` going forward.

## Model Used

- Provider: Anthropic
- Mode: Tool-use (Paperclip Founding Engineer agent, BAC-730)

## Checklist

### DIFF
--- a/packages/db/src/migrations/0082_projects_feature_settings.sql
+++ b/packages/db/src/migrations/0082_projects_feature_settings.sql
@@ -1,0 +1,36 @@
+ALTER TABLE "projects" ADD COLUMN IF NOT EXISTS "feature_settings" jsonb;
+--> statement-breakpoint
+-- Backfill: lift any non-env-binding values from projects.env into feature_settings.
+-- A valid env binding is either a plain string or an object with type = 'plain'|'secret_ref'.
+-- Anything else (e.g. productivityReview: { holdHours, ... }) is a feature-settings entry
+-- that was written into the wrong column and breaks heartbeat env-binding validation.
+DO $$
+BEGIN
+  UPDATE "projects"
+  SET
+    "feature_settings" = COALESCE("feature_settings", '{}'::jsonb) || (
+      SELECT COALESCE(
+        jsonb_object_agg(key, value),
+        '{}'::jsonb
+      )
+      FROM jsonb_each("env") AS t(key, value)
+      WHERE jsonb_typeof(value) = 'object'
+        AND NOT (value ? 'type' AND value->>'type' IN ('plain', 'secret_ref'))
+    ),
+    "env" = (
+      SELECT COALESCE(
+        jsonb_object_agg(key, value),
+        '{}'::jsonb
+      )
+      FROM jsonb_each("env") AS t(key, value)
+      WHERE jsonb_typeof(value) = 'string'
+        OR (value ? 'type' AND value->>'type' IN ('plain', 'secret_ref'))
+    )
+  WHERE "env" IS NOT NULL
+    AND EXISTS (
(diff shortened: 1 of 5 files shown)


brandonburr:

**Rationale:** DB fix: featureSettings column added to prevent env-binding collision in projects.

PR #5536, opened 2026-05-09
feat: detect and reap orphaned child processes in heartbeat recovery

enkogu:
## Summary

When the Paperclip server restarts, previously-tracked child processes may have been reparented to init (PPid=1) or have exited as zombies. Previously, the  recovery path would skip alive-but-orphaned processes (because  returns true), leaving adapter children running indefinitely with no in-memory handle — causing orphan-leak issues.

## Changes

1. **** — New helper that reads  to extract parent PID and process state (including zombie detection).

2. **** — Added  option that produces a specific message for reparented/zombie processes.

3. **** — When a tracked child process is alive but has been reparented to init (PPid=1) or is a zombie:
   - The process is terminated via 
   - The run is finalized **without retry** (since the parent server process is gone, a retry would only produce another orphan)
   - The event payload includes  for observability

## Related

This patch has been running in production on the Prin7r VPS Paperclip instance and resolves orphan-leak issues where adapter child processes (e.g., OpenCode Go) would accumulate after server restarts.

### DIFF
--- a/server/src/services/heartbeat.ts
+++ b/server/src/services/heartbeat.ts
@@ -2135,6 +2135,24 @@ function isProcessAlive(pid: number | null | undefined) {
   }
 }
 
+async function readLinuxProcessStatus(pid: number | null | undefined) {
+  if (process.platform !== "linux") return null;
+  if (typeof pid !== "number" || !Number.isInteger(pid) || pid <= 0) return null;
+  try {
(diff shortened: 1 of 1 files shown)


brandonburr:

**Rationale:** Recovery: orphaned child processes are now detected and reaped.

PR #5565, opened 2026-05-09
feat(k8s-execution): M3a — real claude-code, real git creds, empirical sizing, Cilium DSL

stubbi:
## Thinking Path

> Tactical PR layered on M2 (#5558). Closes the four production-readiness gaps the M2 design left: the agent path is exercised against real Anthropic, git credentials resolve through the existing SecretProvider registry, resource defaults have measurement infrastructure pointed at the real workload, and operators can lock down per-tenant Cilium egress without mutating M1's baseline. One PR, one migration with three columns, no new package.

## What Changed

**Schema (one migration, additive):**
- \`packages/db/src/migrations/0084_tenant_policy_m3a.sql\` adds three columns to \`cluster_tenant_policies\`: \`git_credentials_secret_id\` (FK → company_secrets), \`cilium_dns_allowlist text[]\`, \`cilium_egress_cidrs text[]\` (both \`NOT NULL DEFAULT ARRAY[]::text[]\`).

**Real \`issueGitCredentials\` (replaces M2 stub):**
- \`server/src/services/git-credentials.ts\` resolves a per-company \`companySecrets\` UUID via the existing \`SecretProvider\` registry and decodes \`{username, password}\` JSON. 4 unit tests cover not_configured / happy / non-JSON / shape-bad paths.
- \`server/src/routes/k8s-callback.ts\` wires the real service into the route. Single-cluster-per-company guard fails closed with \`not_configured\`.
- New CLI subcommand: \`paperclip cluster set-git-credentials --cluster <id> --company <id> --secret-id <uuid>\` (UUID-validated).

**Per-tenant Cilium DSL (additive intersection):**
- \`buildTenantCiliumPolicy\` emits a SECOND CiliumNetworkPolicy. Cilium evaluates multiple CNPs as AND, so M3a is strictly tightening — never relaxing. Empty arrays → builder returns null → no second CNP applied. Always-on kube-dns rule prevents the \`['api.anthropic.com']\` footgun where the tenant accidentally blocks DNS for the very host they allowlisted.
- \`ensureTenantNamespace\` calls the builder after the M1 baseline; the M1 builder is unmodified.
- New CLI subcommand: \`paperclip cluster set-cilium-policy --cluster <id> --company <id> [--cilium-dns \"a.com,b.com\"] [--cilium-cidrs \"10.0.0.0/8\"]\`.
- \`docs/k8s-execution/cilium-recipes.md\` covers Anthropic-only, self-hosted-git, and lockdown patterns.

[...]

PR #5576, opened 2026-05-09
feat(k8s-execution): M3b — Redis rate limiter, image allow-list, multi-adapter coverage

stubbi:
## Thinking Path

> Tactical PR per the M3b spec (`docs/superpowers/specs/2026-05-09-paperclip-cloud-adapter-m3b-spec.md`). Three independent items in one PR: cross-replica Redis rate limiter, per-cluster image allow-list, and runtime-image coverage for the six remaining local adapters. One migration (`0085`), no new package.
> 
> Mid-execution the user surfaced a multi-tenancy collision risk in `deriveNamespaceName` (two companies with identical slugs hit the `cluster_namespace_bindings` unique index). Resolved in scope by switching to always-hash namespace derivation.

### Cross-replica Redis rate limit
- New atomic Lua-script-backed limiter at `server/src/routes/_limiter-redis.ts`.
- Factory in `k8s-callback.ts` picks impl based on `PAPERCLIP_REDIS_URL`.
- In-memory fallback preserved for dev / single-replica.
- Documented in `docs/k8s-execution/security-model.md`.

### Per-cluster image allow-list
- Migration `0085_cluster_image_allowlist.sql` adds `image_allowlist text[] NOT NULL DEFAULT ARRAY[]::text[]`.
- `driver.run()` enforces prefix-match before `createAgentJob`. Rejects with `errorCode: "image_not_allowed"`.
- Empty allow-list preserves existing `allowAgentImageOverride` boolean semantics — additive, no behavior change for existing rows.
- New CLI subcommand `paperclip cluster set-image-allowlist`.

### Multi-adapter coverage
- Six new runtime images: `agent-runtime-{codex,gemini,acpx,opencode,pi,hermes}`.
- Per-adapter Dockerfile + buildx-bake target + busybox-style smoke test.
- Single source of truth for per-adapter env keys + FQDN defaults at `adapter-defaults.ts:ADAPTER_DEFAULTS`. Driver consumes it; server side stays adapter-agnostic.
- `hermes_local` ships as a stub because no upstream npm binary is identified yet — documented in `Dockerfile.hermes` and the spec.
- New doc at `docs/k8s-execution/adapters.md`.

### Always-hash namespace derivation
- `deriveNamespaceName` now unconditionally appends `-<8-char-base36-hash(companyId)>` to every namespace (previously conditional on overflow / sanitization).
- Resolves a multi-tenancy collision: two companies with identical slugs (e.g. both \"Acme\") previously hit the `cluster_namespace_bindings` unique index, blocking Company B onboarding.
- Public API of `deriveNamespaceName` simplified — `collisionFallback?: boolean` parameter removed.

## Verification

- [x] `pnpm --filter @paperclipai/db build` (migration 0085 applies cleanly)
- [x] `pnpm --filter @paperclipai/server typecheck` passes
- [x] `pnpm --filter @paperclipai/server test` passes (incl. new Redis limiter tests)
- [x] `pnpm --filter paperclipai test` passes (incl. new CLI subcommand tests)
- [x] `pnpm --filter @paperclipai/execution-target-kubernetes test` passes (incl. driver allow-list + adapter-defaults + always-hash naming tests)
- [ ] `K8S_INTEGRATION=1 vitest run test/integration/<adapter>-smoke.test.ts` for each of {codex, gemini, acpx, opencode, pi, hermes} — operator runs out-of-band
- [ ] Multi-replica deployment with `PAPERCLIP_REDIS_URL` set: rate limit holds across replicas (operator validates)

## Risks & Rollback

- Migration 0085 is purely additive; reversible via `DROP COLUMN`.
- Image allow-list defaults to `[]` so existing deployments see no behavior change.
- Redis fail-open: blips admit requests rather than 500ing endpoints.
- Adapter env mapping is now centralized in the registry; the driver also FILTERS the per-Job env Secret to the adapter's declared envKeys. Server callers passing a wider env will see extra keys silently dropped — defence-in-depth, but worth noting for any caller that depended on extra-key passthrough.
- Always-hash namespace derivation changes namespace names for new tenants — existing namespaces are not migrated. Operator action required only if you have unmigrated tenants whose namespace names you depend on out-of-band.

## Stack

This PR layers on **#5565 (M3a)** which layers on **#5558 (M2)** which layers on **#5556 (M1)**. Diff against `master` will include all four PRs' commits (118 ahead) until they merge; once they do, the diff narrows to ~21 M3b commits.

## Model Used

- Provider: Anthropic
- Model ID: claude-opus-4-7 (1M context)
- Reasoning mode: standard

## Checklist

- [x] Spec coverage: all three sections (§1 Redis, §2 image allow-list, §3 multi-adapter) + mid-flight always-hash fix
- [x] One migration (`0085`), one column, additive
- [x] No new package; all changes in existing packages
- [x] Six new runtime images; one new docs page (`adapters.md`); one new docs section (Production rate limiting); one CHANGELOG entry


stubbi:
@greptileai review

stubbi:
@greptileai review

stubbi:
@greptileai review

stubbi:
@greptileai review

stubbi:
@greptileai review

stubbi:
@greptileai review

stubbi:
@greptileai review

Several P1+P2 fix commits and migration renumbering have landed on this branch since the last Greptile pass; please re-review the latest HEAD.

stubbi:
@greptileai review

Addressed the resolveRunContext / workspaceStrategyJson finding from the latest M2 review (same bug present on this branch — fixed via cherry-pick):

- The init container's `parseRequest` validates `version === 1` and `typeof source.strategy === "string"`, but `resolveRunContext` was emitting `{"kind":"ephemeral"}` — neither field present. First real agent run against a live cluster would fail in the init container with `unsupported version undefined`.
- Replaced with a proper minimal shape: `{ version: 1, source: { strategy: "adapter_managed" } }`. `executeWorkspaceStrategy` dispatches on `source.strategy` and treats anything outside `{"project_primary","git_worktree"}` as a no-op, which matches the M2/M3 semantics where `claude_local` manages its own working directory inside the main container.
- Extracted `parseRequest` from `workspace-init/src/index.ts` to its own module with focused tests covering the version / strategy / minimal-shape contracts.
- Added `buildAdapterManagedWorkspaceRequestJson` server-side with a test that mirrors `parseRequest`'s predicates so regressions on the emitter side fail fast.

Server + workspace-init unit tests pass; tsc baseline unchanged (no new errors introduced).

stubbi:
@greptileai review

Addressed both findings from the latest pass in commit \`912fad10\`:

1. **opencode_local provider gap** (`adapter-defaults.ts:54-58`): expanded `envKeys` to `["ANTHROPIC_API_KEY", "OPENAI_API_KEY", "GEMINI_API_KEY", "GOOGLE_API_KEY", "XAI_API_KEY"]` and `allowFqdns` to the matching four hosts (`api.anthropic.com`, `api.openai.com`, `generativelanguage.googleapis.com`, `api.x.ai`). Mirrors `pi_local`'s broader surface. The test now asserts the full provider set so a regression breaks tests before it ships, instead of surfacing as silent auth failures on a live cluster.
2. **disposeK8sCallback shutdown wiring** (`server/src/app.ts:286-302` → `server/src/index.ts:984`): wired the hook into the existing SIGINT/SIGTERM shutdown handler in `index.ts`. The handler is already async and sequences telemetry + embedded-Postgres teardown before `process.exit`; `disposeK8sCallback` slots in between, only firing when `PAPERCLIP_RUN_JWT_SECRET` is set (matching the gate on the router itself). Errors are logged but do not block exit. Updated the comment in `app.ts` to reflect the wiring is in place.

10 tests in adapter-defaults.test.ts pass; pre-existing tsc errors on M3b unchanged (no new errors introduced).

stubbi:
@greptileai review

Addressed both image-policy findings in commit \`0f652e3f\`:

1. **\`allowAgentImageOverride\` was never enforced** (\`driver.ts:321-332\`): the comment promised "handled below" but no code followed. Added an explicit gate before the allowlist check — when \`target.imageOverride\` is non-null and the connection forbids overrides, returns \`image_override_not_allowed\`.
2. **Dead second allowlist check** (\`driver.ts:333-348\`): removed. \`resolveRunContext\` collapses \`runContext.image = target.imageOverride ?? adapterImage\`, so the single check on \`runContext.image\` covers both cases.

Test fixes: \`makeFakeDriver\` now accepts \`allowAgentImageOverride\` (default true) and its fake \`resolveRunContext\` mirrors the real one. Three new focused tests cover the policy boolean across (override / no-override) × (allowlist empty / non-empty). All 98 unit tests pass.

stubbi:
@greptileai review

Addressed the finding in commit \`ad19b3d0\`:

**\`cmdSetImageAllowlist\` silent-success on unknown cluster id** (\`cli/src/commands/cluster.ts:502-522\`): the command discarded \`update()\`'s return value. The server's \`clusterConnections.update\` returns \`null\` when no row matches the given id, so an operator passing a stale or mistyped id saw "Updated / Cleared image_allowlist for cluster …" while the database was never written. Now checks the returned row and exits with code 1 ("Cluster connection {id} not found") when null. Added a test covering the not-found path; 31 cluster-command tests pass.

stubbi:
@greptileai review

Addressed the type-shape finding in commit \`8f586993\`:

**\`ClusterConnectionsService.update\` return type mismatch** (\`cli/src/commands/cluster.ts:73\`): widened to \`Promise<ClusterConnectionRow | null>\` so the not-found guard in \`cmdSetImageAllowlist\` actually narrows the type. The real server-side implementation in \`server/src/services/cluster-connections.ts\` already returns null on no-match (line 107); the CLI interface had drifted.

31 cluster-command tests still pass; no new tsc errors introduced.

stubbi:
Superseded by #5790 — the new `@paperclipai/plugin-kubernetes` sandbox-provider plugin replaces the in-tree k8s execution adapter approach this PR was building toward.

The hardened patterns this stack pioneered (Cilium DSL, image allowlist glob matching, `ensureTenant`, security baseline) are lifted into the plugin at `packages/plugins/sandbox-providers/kubernetes/`. The plugin model moves the k8s API surface out of paperclip-server core entirely — `paperclip-server` no longer touches the k8s API, only the plugin does.

See #5790 spec/plan docs (`docs/superpowers/specs/2026-05-12-paperclip-k8s-sandbox-provider-plugin-design.md`) §"Why this design" for the full rationale.

Closing as superseded. Local branch retained for reference.

PR #5626, opened 2026-05-10
feat(api): expose authorUser.name on issue comments

KingxBTC:
## Summary

- Add `authorUser: { id, name }` field to the issue comment API response and heartbeat wake payloads
- Allows agents and clients to display comment author's display name without a separate user lookup

## Changes

- `server/src/services/issues.ts`: join `authorUser` from users table on comment queries
- `server/src/services/heartbeat.ts`: include `authorUser` in wake payload comment objects
- `server/src/__tests__/issues-service.test.ts`: unit tests covering the new field

## Motivation

Agents consuming the heartbeat wake payload currently receive comment `authorId` but not the author's display name, forcing an extra round-trip to resolve the name. This PR makes the name available inline, consistent with how other user references are exposed in the API.

## Test plan

- [ ] `pnpm test` passes (new unit tests in `issues-service.test.ts`)
- [ ] Existing comment endpoints return `authorUser.id` and `authorUser.name`
- [ ] Wake payloads for issues with comments include `authorUser` on each comment object

### DIFF
--- a/server/src/__tests__/issues-service.test.ts
+++ b/server/src/__tests__/issues-service.test.ts
@@ -5,6 +5,7 @@ import { sql } from "drizzle-orm";
 import {
   activityLog,
   agents,
+  authUsers,
   companies,
   createDb,
   environments,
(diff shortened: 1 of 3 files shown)


KingxBTC:
Closed automatically when fork branch was deleted. Decision per DAI-176: upstream commit c445e592 supersedes this patch — upstream is now actively developing in this area, so we are dropping the fork patch and tracking upstream master. Cleanup tracked in DAI-224.

PR #5628, opened 2026-05-10
fix: quarantine repeatedly timing out assigned issues

levineam:
## Summary
- prevent stranded assigned-issue recovery from stacking multiple queued runs for the same agent in one sweep
- quarantine assigned issues after repeated recent timeout runs by moving them to blocked with recovery evidence
- add heartbeat recovery tests for one-run-per-agent recovery dispatch and timeout quarantine

## Thinking Path
- Michael was repeatedly timing out the same assigned Paperclip issue, then the recovery sweep kept redispatching it instead of surfacing the failure for intervention.
- The same sweep could also enqueue multiple Michael runs for different assigned issues, which recreates the backlog/stale-lock behavior we are trying to avoid.
- The fix keeps the external-cron architecture intact and hardens Paperclip's own recovery path: one active recovery per agent per sweep, and repeated timeouts become a visible blocked issue with evidence instead of another queued retry.

## Risks
- Recovery throughput is intentionally more conservative: after one recovery wake is queued for an agent, later assigned issues for that agent wait for a later sweep.
- Timeout quarantine may block issues that could have succeeded on one more retry, but the threshold/window are runtime configurable and default to three timeouts in 24 hours.
- The quarantine path creates recovery work; this revision also marks the agent as recovered in-memory after quarantine so follow-up dispatches cannot stack in the same sweep.

## Verification
- pnpm exec vitest run server/src/__tests__/heartbeat-process-recovery.test.ts
- pnpm typecheck

## Operational Note
- Live SUP-1321 was moved to blocked with a timeout-evidence comment after repeated Michael timeouts.
- Queued duplicate Michael runs for SUP-1499 and SUP-1501 were cancelled operationally so the live system has only one active Michael run while this PR is reviewed.

## Model Used
- GPT-5 Codex

### DIFF
--- a/server/src/__tests__/heartbeat-process-recovery.test.ts
+++ b/server/src/__tests__/heartbeat-process-recovery.test.ts
@@ -1808,6 +1808,38 @@ describeEmbeddedPostgres("heartbeat orphaned process recovery", () => {
     }
   });
 
+  it("does not stack assigned todo dispatches for the same agent during one recovery sweep", async () => {
+    const { companyId, agentId, issueId } = await seedAssignedTodoNoRunFixture();
+    const secondIssueId = randomUUID();
+    const issuePrefix = `T${companyId.replace(/-/g, "").slice(0, 6).toUpperCase()}`;
+    await db.insert(issues).values({
+      id: secondIssueId,
+      companyId,
+      title: "Second assigned todo work without a heartbeat",
+      status: "todo",
+      priority: "medium",
+      assigneeAgentId: agentId,
+      assigneeUserId: null,
+      issueNumber: 2,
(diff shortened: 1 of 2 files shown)


levineam:
Addressed the Greptile review items in bf1e1c99:\n\n- Added recoveredAgentIds.add(agentId) in the timeout-quarantine branch so quarantine-created recovery wakes participate in the same per-agent guard as other dispatch paths.\n- Increased recent timeout evidence fetch from policy.count to policy.count + 5 so quarantine comments do not understate repeated timeout volume.\n- Updated the PR body with Thinking Path, Risks, and Model Used sections.

PR #5631, opened 2026-05-10
fix(api): prevent update wiping secret bindings

freddiecoleman:
## Summary

I had an agent delete bindings when it called the API. Paperclip code has it's own guards to prevent this but that is a client side protection which the agent did not follow. This adds a server-side guard consistent with other methods.

## Thinking Path

  > - Agents store their runtime environment variables in `adapterConfig.env`, some of which are secret refs pointing to stored credentials
  > - The secrets subsystem tracks which secrets are bound to which agents via `companySecretBindings` — this table is the source of truth for what an agent is authorised to use at runtime
  > - `syncEnvBindingsForTarget` does a full DELETE of all `env.*` bindings for a target before re-inserting from the current env — calling it with `undefined` wipes all bindings and inserts nothing
  > - The agent create route correctly guarded this call with `if (agentEnv)`, but the update route did not — meaning any PATCH that touched `adapterConfig` but resulted in no `env` on the returned agent would
  silently clear all secret bindings
  > - This is exactly the kind of mistake an agent makes on itself: it patches its own config to change a model or a flag, sends a minimal `adapterConfig` that doesn't include `env` because it didn't know to
  preserve it, and its credentials are silently unbound with no error
  > - This pull request adds the missing guard on the update path, matching the create path, and adds route-level tests covering both the regression and the positive case

  ## What Changed

  - Added `if (agentEnv)` guard around `syncEnvBindingsForTarget` in the agent PATCH route (`server/src/routes/agents.ts`), matching the behaviour of the create route
  - Added `syncEnvBindingsForTarget` to the mock secret service in `agent-permissions-routes.test.ts`
  - Added regression test: verifies sync is not called when the updated `adapterConfig` has no `env` (would have failed before this fix)
  - Added positive test: verifies sync is still called with the correct env when `env` is present after update

  ## Verification

  - `npx vitest run src/__tests__/agent-permissions-routes.test.ts` — all 45 tests pass, including the two new ones
  - The regression test directly captures the broken behaviour: mock `agentService.update` returns an agent with no `env`, assert `syncEnvBindingsForTarget` was never called

  ## Risks

  Low risk. The guard matches what the create path already does. The only behavioural change is skipping the sync when `adapterConfig.env` is absent after an update — in that case the sync would have deleted all
  bindings and inserted nothing, so skipping it is strictly safer. No migrations required.

  ## Model Used

  - Claude (Anthropic), `claude-opus-4-7` via Claude Code CLI. Used with tool execution (Read, Edit, Bash, Explore subagents) and plan mode for design review before each edit pass.

  ## Checklist

### DIFF
--- a/server/src/__tests__/agent-permissions-routes.test.ts
+++ b/server/src/__tests__/agent-permissions-routes.test.ts
@@ -85,6 +85,7 @@ const mockIssueService = vi.hoisted(() => ({
 const mockSecretService = vi.hoisted(() => ({
   normalizeAdapterConfigForPersistence: vi.fn(),
   resolveAdapterConfigForRuntime: vi.fn(),
+  syncEnvBindingsForTarget: vi.fn(),
 }));
 
 const mockAgentInstructionsService = vi.hoisted(() => ({
@@ -318,6 +319,7 @@ describe.sequential("agent permission routes", () => {
     mockIssueService.list.mockReset();
     mockSecretService.normalizeAdapterConfigForPersistence.mockReset();
     mockSecretService.resolveAdapterConfigForRuntime.mockReset();
+    mockSecretService.syncEnvBindingsForTarget.mockReset();
     mockAgentInstructionsService.materializeManagedBundle.mockReset();
     mockCompanySkillService.listRuntimeSkillEntries.mockReset();
     mockCompanySkillService.resolveRequestedSkillKeys.mockReset();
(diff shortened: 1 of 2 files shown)


freddiecoleman:
There is at least one other instance when this can happen, the reproduction is if you tell an agent such as CEO to create a new agent with secrets assigned to them as env vars. They will assign the env vars but they do not get the bindings so when the agent runs they are unable to access them.

The current workaround is to go into the agent and make any change (even no change) and save the configuration. The update causes the bindings to be created and then it works.

brandonburr:

**Rationale:** Critical: prevents data loss when updating issues with secrets — secret bindings are now preserved.

PR #5657, opened 2026-05-10
Fix: assertTransition is a no-op — any issue status transition succeeds, bypassing review gates

banks-ux:
## Thinking Path

> - Paperclip orchestrates AI agents for zero-human companies; agent work moves through a status pipeline (`todo → in_progress → in_review → done`, plus side-states for `blocked`, `cancelled`).
> - The `assertTransition` function in `server/src/services/issues.ts` is the central guardrail meant to enforce both valid-transition semantics and the review gate (issues with acceptance-criteria / brief / deliverable / qa documents must transit `in_review` before `done`).
> - At pristine v2026.428.0, `assertTransition` is a no-op stub — it only validates that the destination status is a known enum value, accepting any `* → *` transition between known statuses. Review-gated workflows have no platform protection.
> - Convention-based enforcement (telling agents "transition through `in_review`" in their training files) is fragile; one slip ships review-required work directly to `done` without a reviewer's verdict.
> - This PR replaces the no-op with: an `ALLOWED_TRANSITIONS` matrix encoding directed-edge legality per source state, a `hasIssueReviewGate` heuristic that checks for review-gate documents on the issue, and audit logging for terminal-state reversals (`done → !done` and `cancelled → todo`).
> - The benefit is that gate-bypassing `in_progress → done` transitions on review-gated issues now return a 409 with a documented error message, and downstream observability of terminal-state reversals is preserved via an `issue.status.reverse_transition` activity-log entry.

## What Changed

- `assertTransition(from, to, { issueId, hasReviewGate })` now enforces a per-source-state `ALLOWED_TRANSITIONS` matrix. Disallowed transitions throw 409 conflict with the source and target named in the error.
- Review-gate check on `in_progress → done`: if the issue has any `issue_documents` row keyed `spec`/`deliverable`/`qa`/`brief` (`REVIEW_GATE_DOCUMENT_KEYS`), the transition is rejected with a 409 citing the gate. Caller must transit via `in_review` first.
- Review-gate check + `assertTransition` call run INSIDE the `update()` transaction so a concurrently-added gated document cannot land between the check and the status update.
- `issue.status.reverse_transition` activity-log entry emitted for both terminal-state reversals: `done → !done` and `cancelled → todo`. Discriminating `reason` string distinguishes the two paths so downstream consumers (reporting, productivity-review) can reason about the lifecycle correctly.
- 16 integration tests added covering: full happy path, gate enforcement on every `from !== in_review → done` source, gate bypass when no gate documents, matrix-aligned rejection of `todo → in_review`, allowed `cancelled → todo` un-cancel, `in_review → todo` accept-interaction flow, idempotent same-state, and both reverse-transition audit entries (done revert + un-cancel).

## Verification

```bash
pnpm --filter @paperclipai/server test -- issues-transition.test.ts
```

All 16 tests pass against an embedded Postgres instance. Tests skip gracefully on hosts where embedded Postgres is unavailable.

Live-verified in production (FirstString HoldCo deployment, 2026-05-07 onward): direct `in_progress → done` transitions on review-gated issues return 409 with the documented error message; non-gated transitions through the matrix work as expected; `done → in_review` revert emits the activity-log entry.

## Risks

Medium risk on initial adoption. Deployments that historically bypassed the review gate via direct `* → done` updates will start receiving 409s on those calls — that's the intended behavior, but it may surface latent agent-training bugs in adopting deployments. The matrix is intentionally tight; deployments that need additional transitions (e.g., early-review-pickup `todo → in_review`) can add them by extending `ALLOWED_TRANSITIONS`. Moving the gate check inside the transaction closes a race that existed pre-PR but is otherwise behavior-preserving.

## Model Used

Claude (Anthropic) — Claude Opus 4.7 (1M context window) via Claude Code CLI. The change was authored as part of the local patch series for FirstString HoldCo, then prepared as this upstream PR. Greptile-review revisions in this PR also authored by the same model.

## Checklist

- [x] I have added tests covering the changed behavior
- [x] I have verified the change against my local deployment
- [x] I have flagged risks
- [x] I have specified the AI model used

### DIFF
--- a/server/src/__tests__/issues-transition.test.ts
+++ b/server/src/__tests__/issues-transition.test.ts
@@ -142,10 +142,17 @@ describeEmbeddedPostgres("issueService.update status transition guard", () => {
     });
   });
 
-  it("allows todo -> in_review (early review pickup)", async () => {
+  it("rejects todo -> in_review (matrix requires transit via in_progress)", async () => {
+    // The ALLOWED_TRANSITIONS matrix in services/issues.ts intentionally
+    // does NOT include `in_review` in the `todo` source set: review pickup
+    // happens via `in_progress` (work-then-review) rather than as a direct
+    // skip from `todo`. If a workflow needs early-review-pickup, the matrix
+    // is the right place to relax it; for now the strict pipeline keeps
+    // execution lifecycle observable.
     const { issueId } = await setup({ initialStatus: "todo" });
-    const row = await svc.update(issueId, { status: "in_review" });
-    expect(row?.status).toBe("in_review");
+    await expect(svc.update(issueId, { status: "in_review" })).rejects.toMatchObject({
+      status: 409,
+    });
   });
 
   it("rejects in_progress -> done when a review gate is present", async () => {
@@ -240,6 +247,34 @@ describeEmbeddedPostgres("issueService.update status transition guard", () => {
     expect(logs[0]!.details).toMatchObject({
       from: "done",
       to: "in_review",
+    });
+  });
+
+  it("emits an activityLog reverse_transition event on cancelled -> todo (un-cancel)", async () => {
+    // Parallel to the done-revert audit: un-cancelling work leaves the
+    // terminal-state footprint behind and downstream consumers (reporting,
+    // productivity-review) need to see the trail entry to reason about the
+    // lifecycle correctly.
+    const { companyId, issueId } = await setup({ initialStatus: "cancelled" });
+    await svc.update(issueId, { status: "todo" });
+    const logs = await db
+      .select()
+      .from(activityLog)
+      .where(
+        and(
+          eq(activityLog.companyId, companyId),
+          eq(activityLog.entityType, "issue"),
+          eq(activityLog.entityId, issueId),
+          eq(activityLog.action, "issue.status.reverse_transition"),
+        ),
+      );
+    expect(logs.length).toBe(1);
+    expect(logs[0]!.actorType).toBe("system");
+    expect(logs[0]!.actorId).toBe("issues.update");
+    expect(logs[0]!.details).toMatchObject({
+      from: "cancelled",
+      to: "todo",
+      reason: "cancelled state un-cancelled",
     });
   });
 });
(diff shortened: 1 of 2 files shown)


banks-ux:
**Bug fix push (force-pushed `--force-with-lease` to `fix/status-transition-guard`):**

Original patch banned direct `todo → done` unconditionally via the matrix. CI surfaced two upstream tests under `issueService blockers and dependency wake readiness > only returns dependents once every blocker is done` and `> reports dependency readiness for blocked issue chains` that legitimately transition `todo → done` for tasks without review-gate documents. Plus `issue-thread-interactions-service.test.ts > returns agent-authored request confirmations to the creating agent when a board user accepts` was rejecting `in_review → todo` (the v2026.428 acceptInteraction path) — that relaxation existed in our deployment but was missing from the patch file used to file this PR.

**Design correction:**

- `ALLOWED_TRANSITIONS[todo]` now includes `in_review` and `done`.
- `ALLOWED_TRANSITIONS[in_review]` now includes `todo` (acceptInteraction).
- Review-gate guard generalized from `from === "in_progress" && to === "done"` to `to === "done" && from !== "in_review"`, so any task with attached `spec` / `deliverable` / `qa` / `brief` documents is forced through `in_review` regardless of source state. Tasks without those documents may transition directly to `done`.

Original design intent (force review-gated tasks through `in_review`) preserved; non-review-gated tasks can skip directly to `done` from any sensible state.

Test count bumped from 15 → 16 to cover the new branches (`allows todo -> done when no review gate is present`, `rejects todo -> done when a review gate is present`, `allows todo -> in_review`; removed the obsolete `rejects todo -> done direct` matrix-only assertion).

Verified locally: full server suite passes 1440/1441 (sole failure = `heartbeat-comment-wake-batching > promotes deferred comment wakes after the active run closes the issue`, a pre-existing failure on pristine `v2026.428.0` with no patches applied — not introduced by this PR).

banks-ux:
The remaining `verify` job failure (heartbeat-comment-wake-batching.test.ts:25 timeout) reproduces on pristine v2026.428.0 master with no patches applied — it's not introduced by this PR. Tracked in our internal OQ-185. The 2 patch-related failures (`Invalid status transition: todo → done` and `in_review → todo`) that the original 4-annotation set surfaced are both resolved by force-pushed commit 6264425. Happy to investigate the pre-existing timeout if maintainer thinks it's actionable from this PR's scope, but it appears to be a standalone upstream test stability issue.

PR #5666, opened 2026-05-10
fix(server/recovery): skip stranded escalation when issue has unresolved blockers

Noridem:
## Summary

Skip the stranded-issue auto-block when the issue is legitimately waiting on an unresolved blocker recorded in `issue_relations`.

The `reconcileStrandedAssignedIssues()` loop already guards against recent agent activity via the Phase-1 cooldown window (PR #5305, commit `db578ca` — `STRANDED_RECOVERY_COOLDOWN_MS` default 600s). But it has a second false-positive mode: when an issue legitimately waits on a dependency (`issueRelations.type = "blocks"`), the loop reads the missing `executionRunId` as "stranded" and fires the auto-block comment, reverting the issue to `blocked` even though it was working as intended.

This PR closes that gap by adding an `existingUnresolvedBlockerIssueIds()` pre-check at all three `escalateStrandedAssignedIssue()` call sites:

| Site | Branch | service.ts line |
|------|--------|-----------------|
| Site-A | `assignment_recovery` (`todo` path) | 1739 |
| Site-B | `isRepeatedProductiveContinuationRecovery` (in-progress retry) | 1810 |
| Site-C | `issue_continuation_needed` (in-progress fail) | 1872 |

The helper at `service.ts:1541` was already available — `escalateStrandedAssignedIssue` itself calls it on line 1582 to register the recovery issue as a blocker. We now also call it before deciding to escalate, so issues with active blockers are skipped instead.

A new `blockerSkipped` counter on the result object provides operational visibility analogous to the existing `cooldownSkipped`.

## Test plan

- [x] `pnpm --filter @paperclipai/server vitest run src/__tests__/heartbeat-process-recovery.test.ts` — 44/44 PASS (41 existing + 3 new)
- [x] New negative-path test: `continuation_needed` does not escalate when blocker exists
- [x] New negative-path test: `assignment_recovery` does not escalate when blocker exists
- [x] New positive-control test: still escalates after blocker resolves (`status=done`)
- [x] All three new tests verify `blockerSkipped` counter increments correctly

## Production incidents covered

Two reproduced false-positives in our PSY deployment within 48 hours:
- 2026-05-06 MA F4-T2 — auto-blocked despite `blockedByIssueIds` populated
- 2026-05-08 MA F4-T1 — auto-blocked 5 min after atomic wave-trigger

A third recurrence happened on the tracking issue for this very patch while it was being investigated.

## Footprint

- +19 LOC prod / +109 LOC tests
- No API surface change, no schema change, no new ENV flag
- Performance: +1 DB query per iteration only when the cooldown check passes (typically <5% of iterations)

## Companion to PR #5305

Strictly orthogonal to the Phase-1 cooldown fix:
- #5305 covers the *timing* axis (do not escalate while the agent was recently active).
- This PR covers the *state* axis (do not escalate while the issue is legitimately blocked).

Both can land independently.

### DIFF
--- a/server/src/__tests__/heartbeat-process-recovery.test.ts
+++ b/server/src/__tests__/heartbeat-process-recovery.test.ts
@@ -2052,6 +2052,115 @@ describeEmbeddedPostgres("heartbeat orphaned process recovery", () => {
     }
   });
 
+  it("skips Klasse-2 escalation when the in_progress issue has unresolved blockers (continuation_needed path)", async () => {
+    const { companyId, issueId } = await seedStrandedIssueFixture({
+      status: "in_progress",
+      runStatus: "failed",
+      retryReason: "issue_continuation_needed",
+    });
+    const blockerIssueId = randomUUID();
+    const issuePrefix = `T${companyId.replace(/-/g, "").slice(0, 6).toUpperCase()}`;
+    await db.insert(issues).values({
+      id: blockerIssueId,
+      companyId,
+      title: "Unresolved prerequisite blocker",
+      status: "in_progress",
+      priority: "medium",
+      issueNumber: 99,
+      identifier: `${issuePrefix}-99`,
+    });
+    await db.insert(issueRelations).values({
+      companyId,
+      issueId: blockerIssueId,
+      relatedIssueId: issueId,
+      type: "blocks",
+    });
+    const heartbeat = heartbeatService(db);
+
+    const result = await heartbeat.reconcileStrandedAssignedIssues();
+    expect(result.escalated).toBe(0);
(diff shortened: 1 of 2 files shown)


Noridem:
Good catch — fixup pushed in `aa107be` adds the missing negative-path test for Site-B (`isRepeatedProductiveContinuationRecovery`) using the same fixture as the existing positive Site-B coverage at line 2455 of `heartbeat-process-recovery.test.ts` (`runStatus: "succeeded"`, `runSource: "issue.productive_terminal_continuation_recovery"`, `livenessState: "advanced"`).

Updated test count: 45/45 PASS.

PR #5671, opened 2026-05-11
fix(plugin-tool-dispatcher): propagate pluginDbId so worker.isRunning resolves

Ramon-nassa:
Fixes #2391
Fixes #3394
Fixes #4094
Fixes #5501
Fixes #5916
Fixes #6215
Fixes #6514

## Thinking Path

> - Paperclip orchestrates AI agents for zero-human companies
> - Plugins extend the platform by registering agent-callable tools backed by long-running worker processes
> - `PluginToolDispatcher` is the boundary between the HTTP `/api/plugins/tools/execute` route and `PluginWorkerManager`, which owns those worker processes
> - `PluginWorkerManager` keys live workers by the plugin's **database UUID**, but `plugin-loader` was registering tools using only `pluginKey` — so every tool call did `workerManager.isRunning(pluginKey)` and always got `false`
> - As a result, every `POST /api/plugins/tools/execute` against a tool-exposing plugin returned 502 `worker for plugin X is not running`, even though the worker process was alive (hit in production by `vexion.council-chat`; `mem0-sync` would be next)
> - This pull request threads the DB UUID through the dispatcher → registry hop and hardens the contract so omitting the UUID is a compile-time error, not a silent fallback
> - The benefit is plugin tool execution actually works for any plugin declaring `manifest.tools[]`, and the type system prevents the same bug from recurring

## What Changed

- `server/src/services/plugin-loader.ts` — pass in-scope `pluginId` (DB UUID) as the third argument to `toolDispatcher.registerPluginTools`. Single-line root fix.
- `server/src/services/plugin-tool-dispatcher.ts` — `registerPluginTools` now takes `pluginDbId: string` (required, was optional). JSDoc updated to document the worker-routing contract and why the optional signature masked the bug.
- `server/src/services/plugin-tool-registry.ts` — `registerPlugin` throws on missing/empty `pluginDbId` so any new call site that forgets the UUID fails immediately rather than silently falling back to `pluginKey`.
- `server/src/__tests__/plugin-tool-dispatcher-pluginDbId.test.ts` — new focused regression suite covering the activation path, disable→enable lifecycle, worker re-spawn, and the empty-UUID guard.

## Verification

- `pnpm vitest run server/src/__tests__/plugin-tool-dispatcher-pluginDbId.test.ts` — 6/6 passing.
[...]

PR #5676, opened 2026-05-11
fix(plugins): pass DB UUID to registerPluginTools so tools/execute works after reinstall

nilssonr:
## Thinking Path

> - Paperclip orchestrates AI agents via plugin workers communicating over JSON-RPC
> - The plugin tool dispatcher routes `tools/execute` calls through `plugin-tool-registry`, which stores a `pluginDbId` per tool for `workerManager.isRunning()` lookups
> - `workerManager.isRunning(id)` checks the in-memory `workers` map, which is keyed by **DB UUID**
> - `activatePlugin` called `registerPluginTools(pluginKey, manifest)` without passing the DB UUID — so tools were stored with `pluginDbId = pluginKey` (the string key like `"acme.foo"`)
> - `isRunning("acme.foo")` always returned `false` because the workers map has no such entry — only the UUID key — producing a 502 on every `tools/execute` call after reinstall
> - A second issue: `unloadSingle` only called `stopWorker` when `isRunning` was true, leaving crashed/starting/stopping worker handles in the map; `startWorker` would throw "already registered" on the next activate
> - This PR fixes both: pass the DB UUID to `registerPluginTools`, and call `stopWorker` unconditionally in `unloadSingle`

## What Changed

- `server/src/services/plugin-tool-dispatcher.ts`: Add optional `pluginDbId` parameter to `registerPluginTools` interface and implementation; forward it to `registry.registerPlugin`
- `server/src/services/plugin-loader.ts` (`activatePlugin`): Pass `pluginId` (DB UUID) as third argument to `registerPluginTools` so tools store the correct lookup key
- `server/src/services/plugin-loader.ts` (`unloadSingle`): Replace `if (isRunning) stopWorker` with unconditional `stopWorker` call; the manager already handles "not found" gracefully, and crashed/starting handles must also be evicted
- `server/src/__tests__/plugin-tool-registry-reinstall.test.ts`: New regression tests — with UUID → execute succeeds; without UUID → execute throws "not running"; unregister + re-register with new UUID → execute succeeds (hard-delete + reinstall path)

## Verification

```sh
cd server
pnpm exec vitest run src/__tests__/plugin-tool-registry-reinstall.test.ts  # 3 new tests pass
pnpm exec vitest run src/__tests__/plugin-worker-manager.test.ts src/__tests__/plugin-scoped-api-routes.test.ts src/__tests__/plugin-routes-authz.test.ts src/__tests__/plugin-database.test.ts src/__tests__/heartbeat-plugin-environment.test.ts src/__tests__/environment-runtime.test.ts  # 62 existing tests pass
pnpm exec tsc --noEmit  # clean
```

## Risks

- `registerPluginTools` gains an optional third parameter — all existing callers that omit it continue to work (falls back to `pluginId`). The only call that changes behaviour is in `activatePlugin` which now passes the UUID it already has in scope.
- Unconditional `stopWorker` in `unloadSingle` is safe: `stopWorker` logs a warning and returns immediately when no handle exists; `stopInternal` returns immediately when process is already gone (null childProcess after crash).
- Low risk overall: changes are additive to the public interface and fix a consistent failure mode.

## Model Used

- Provider: Anthropic
- Model: claude-sonnet-4-6
- Tool use: yes (file reads, bash, edit)
- Context: 200k

## Checklist

### DIFF
--- a/server/src/__tests__/plugin-tool-registry-reinstall.test.ts
+++ b/server/src/__tests__/plugin-tool-registry-reinstall.test.ts
@@ -1,0 +1,119 @@
+/**
+ * Regression tests for JEN-27: plugin tools/execute 502 after reinstall.
+ *
+ * Root cause: activatePlugin called registerPluginTools(pluginKey, manifest)
+ * without the DB UUID. The tool registry stored pluginDbId = pluginKey.
+ * workerManager.isRunning(pluginKey) always returned false because the workers
+ * map is keyed by DB UUID, not pluginKey. Result: 502 on every tools/execute call.
+ *
+ * Fix: registerPluginTools now accepts an optional pluginDbId parameter.
+ * activatePlugin passes the DB UUID so isRunning checks resolve correctly.
+ */
+
+import { describe, it, expect, vi } from "vitest";
+import { createPluginToolRegistry } from "../services/plugin-tool-registry.js";
+import type { PluginWorkerManager } from "../services/plugin-worker-manager.js";
+import type { PaperclipPluginManifestV1 } from "@paperclipai/shared";
+
+const PLUGIN_KEY = "acme.test-plugin";
+const PLUGIN_DB_UUID = "00000000-0000-0000-0000-000000000001";
+
+const manifest: PaperclipPluginManifestV1 = {
+  id: PLUGIN_KEY,
+  apiVersion: 1,
+  version: "1.0.0",
+  displayName: "Test Plugin",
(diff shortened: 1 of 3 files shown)


devinfoley:
Closing as duplicate of #5671 (now merged in 62863126). #5671 threads pluginDbId through the plugin tool dispatcher and registry to fix the '502 worker not running' routing bug — the same root cause this PR addressed. Thanks for the contribution!

PR #5679, opened 2026-05-11
fix(run-id): guard Paperclip activity run references

jooneyp:
## Summary
- validate `X-Paperclip-Run-Id` / JWT run ids against `heartbeat_runs` before attaching them to actors
- make `logActivity` drop nonexistent run ids instead of surfacing FK failures
- keep OpenClaw gateway Paperclip wake metadata out of strict top-level agent params
- normalize invalid Hermes session ids so continuation retries do not persist helper text tokens

## Tests
- `pnpm vitest run server/src/__tests__/activity-service.test.ts server/src/__tests__/adapter-registry.test.ts packages/adapters/openclaw-gateway/src/server/execute.test.ts`

Paperclip: STO-798

### DIFF
--- a/packages/adapters/openclaw-gateway/src/server/execute.test.ts
+++ b/packages/adapters/openclaw-gateway/src/server/execute.test.ts
@@ -1,5 +1,5 @@
 import { describe, expect, it } from "vitest";
-import { resolveSessionKey } from "./execute.js";
+import { filterOpenClawAgentParams, resolveSessionKey } from "./execute.js";
 
 describe("resolveSessionKey", () => {
   it("prefixes run-scoped session keys with the configured agent", () => {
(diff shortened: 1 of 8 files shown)


jooneyp:
Force-pushed: dropped `9b0fa15e (recovery cooldown)` from this branch — moved to its own PR #5705 because it's an unrelated recovery-service change (different file, different concern, STO-833). 

This branch now contains only the gateway/run-id work:
- `fix(run-id): guard Paperclip activity run references`
- `revert(openclaw-gateway): keep paperclip metadata at request root (STO-798)`
- `fix(openclaw-gateway): strip unsupported agent params` ← new, whitelisted agent RPC params with vitest coverage (5/5 pass in `execute.test.ts`)

— Pieter (Claude Opus 4-7)

rsaulo:
Real-world validation note from a self-hosted Paperclip 2026.517.0 instance: we hit both failure classes that this PR appears to address.

1. Hermes returned a truncated `sessionDisplayId` like `20260518_175454_` while `result_json.session_id` contained the full id like `20260518_175454_0af855`. Persisting the truncated id caused continuation/recovery retries to fail; a later retry also parsed helper text (`Use a session ID from...`) as session id `from`, which stranded the issue in recovery.
2. A non-UUID `X-Paperclip-Run-Id` on an authenticated local API request caused a 500 through `activity_log.run_id` validation/FK handling instead of being ignored/nullified.

We applied local in-place hotfixes equivalent to the behavior here (prefer canonical full Hermes session id, reject invalid session tokens, and ignore invalid run ids) and the recovery noise stopped. This PR looks like the right upstream fix to replace those local patches in the next release.

PR #5691, opened 2026-05-11
fix(ui): list view quick filter must fetch issues from server

ensky:
## Summary

- The project list view's quick-filter presets (`Active`, `Backlog`, `Done`) only filtered the already-loaded `issues` prop on the client.
- Large projects exceed the default 500-item server cap, and the server's priority-first sort drops all low-priority issues past that cap — so the `Active` preset could no longer surface them, even when they matched.
- Mirror the board view's per-status fetching pattern in the list view: when a status filter is active, issue one server query per selected status (`limit: 200`) and merge before client-side filtering.
- Reuse the same UX as `boardColumnLimitReached` for an analogous `listStatusLimitReached` hint when any status query saturates its cap.

## Repro

1. Open any project with > 500 issues whose `low`-priority entries would sort past rank 500 (priority first, then `canonicalLastActivityAt DESC`).
2. Switch to **List** view, pick the **Active** quick filter.
3. Observe that low-priority `todo` / `in_progress` / `in_review` / `blocked` issues never appear despite matching the preset.

After this change the matching issues are fetched from the server per-status and rendered.

## Test plan

- [x] `pnpm typecheck` (ui)
- [x] `pnpm build` (ui)
- [ ] Manual: large project, list view, toggle Active / Backlog / Done — verify previously hidden issues now show, and switching back to **All** restores the original prop-driven list.
- [ ] Manual: confirm the `listStatusLimitReached` hint renders when any status query hits the 200-item cap.

### DIFF
--- a/ui/src/components/IssuesList.tsx
+++ b/ui/src/components/IssuesList.tsx
@@ -714,12 +714,17 @@ export function IssuesList({
   // re-fetch issues per status from the server so that low-priority or older issues that
   // fall outside the default 500-item server cap are still surfaced. This mirrors the
   // per-status fetching pattern used by the board view.
+  // Skip per-status fetches when a remote search is active — `filtered` uses
+  // `searchedIssues` in that case and would discard these results.
   const listStatusFilterValues = useMemo(
     () =>
-      viewState.viewMode === "list" && !searchWithinLoadedIssues && viewState.statuses.length > 0
+      viewState.viewMode === "list"
+      && !searchWithinLoadedIssues
+      && normalizedIssueSearch.length === 0
+      && viewState.statuses.length > 0
         ? viewState.statuses
         : [],
-    [searchWithinLoadedIssues, viewState.statuses, viewState.viewMode],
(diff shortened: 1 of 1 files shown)


ensky:
Addressed both Greptile P2 comments in 92a16030:

- **Unnecessary per-status queries during remote search**: `listStatusFilterValues` now also gates on `normalizedIssueSearch.length === 0`. When a remote search is active, `filtered` uses `searchedIssues` and would discard these results, so we skip the fetches entirely.
- **Missing view-mode/search-mode `enabled` guards**: each query's `enabled` now mirrors the board pattern with explicit `viewState.viewMode === "list" && !searchWithinLoadedIssues && normalizedIssueSearch.length === 0` checks, so the intent is self-documenting and survives future refactors.

Also dropped the now-redundant `normalizedIssueSearch` / `q` parameters from the query key and queryFn since the query only runs when the search box is empty.

`pnpm --filter @paperclipai/ui typecheck` + `build` ✅

PR #5705, opened 2026-05-11
fix(recovery): 1h cooldown on terminal stranded-recovery replacements (STO-833)

jooneyp:
## Thinking Path

> - The `stranded_assigned_issue` recovery invariant turns orphaned in-progress work into explicit recovery issues so a human-pickable trail exists
> - When a recovery issue itself finishes (done/cancelled) but the underlying source issue still looks "stranded", the invariant immediately fires again and spawns yet another recovery issue — same source, same fingerprint
> - STO-833 is one such replacement: a fresh recovery for source STO-806 was created moments after the previous recovery cycle terminated, even though no new orphaning event had occurred. Same class of bug as the STO-259 cascade (PR #4671 fixed nested recovery-of-recovery; this fixes immediate same-source replacement)
> - This pull request adds a 1-hour cooldown: if the most recent terminal (done/cancelled) stranded recovery issue for the same source ended within the past hour, we suppress the replacement and log an `issue.stranded_recovery_suppressed` activity instead
> - The benefit is reducing duplicate no-op recovery issues that the system can never act on (because the source state hasn't changed), without dropping legitimate re-detection after the cooldown expires

## What Changed

- `server/src/services/recovery/service.ts`: new `findRecentTerminalStrandedIssueRecoveryIssue` helper that looks up the most recent `done`/`cancelled` stranded recovery issue for a given source within `STRANDED_ISSUE_RECOVERY_TERMINAL_COOLDOWN_MS` (1h)
- `ensureStrandedIssueRecoveryIssue`: before creating a replacement, checks the helper; on hit it logs `issue.stranded_recovery_suppressed` activity (with `recentRecoveryIssueId`, `recentRecoveryIdentifier`, `recentRecoveryStatus`, `cooldownMs`) and returns `null`
- `heartbeat-process-recovery.test.ts`: new test case `does not create a replacement stranded recovery issue immediately after one finished` covering the cooldown path

## Verification

```sh
cd server
npx vitest run src/__tests__/heartbeat-process-recovery.test.ts
# 45/45 pass including the new case
```

Manual: in dev, take a `stranded_issue_recovery` issue, mark it `done`, then trigger heartbeat. Without this patch a new recovery issue spawns immediately for the same source; with this patch, an `issue.stranded_recovery_suppressed` activity row appears instead and no new issue is created until the cooldown elapses.

## Risks

- **Low**. The suppression only triggers when an `originKind=stranded_issue_recovery + originId=source + status in (done,cancelled) + updatedAt within 1h` row exists. If the source genuinely needs re-recovery (e.g. a new orphaning event after the cooldown), the next heartbeat after 1h will create one. The cooldown window is short enough that a real regression cannot hide silently.
- Activity-log row gives an auditable trail of what was suppressed and why.
- No schema change. No migration.

Roadmap impact: targeted invariant bug-fix, not a roadmap feature. Same family as merged-stream STO-259 work (PR #4671/#4672) which the maintainers already accept as a recovery hardening track.

## Model Used

- Claude Opus 4-7 (Anthropic, ~1M context, extended-thinking off). Used for diagnosis (STO-833 inspection, recovery service code reading), patch drafting, and test design. Author reviewed each diff before commit. No tool other than typecheck + targeted vitest run was executed against this branch.

## Checklist

- [ ] If this change affects the UI, I have included before/after screenshots — n/a, server-only
- [x] I have updated relevant documentation to reflect my changes — none required; activity-log shape is the only externally visible artifact and it's covered by the test

### DIFF
--- a/packages/adapters/openclaw-gateway/src/server/execute.test.ts
+++ b/packages/adapters/openclaw-gateway/src/server/execute.test.ts
@@ -1,5 +1,5 @@
 import { describe, expect, it } from "vitest";
-import { buildOpenClawGatewayAgentParams, resolveSessionKey } from "./execute.js";
+import { resolveSessionKey } from "./execute.js";
 
 describe("resolveSessionKey", () => {
   it("prefixes run-scoped session keys with the configured agent", () => {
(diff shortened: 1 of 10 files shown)


jooneyp:
Closing in favor of a fresh single-commit PR off origin/master — this PR's branch (`fix/sto-833-recovery-cooldown`) accidentally included two unrelated STO-798 commits (d7120cd0, 53b9e2e9) that already live in PR #5679. Replacing with branch `sto-833-stranded-recovery-cooldown` containing only the STO-833 cooldown commit.

Issue #5713, opened 2026-05-11
Feature: detect agent run-loops (oscillation on agent+issue+wakeReason) — heartbeat-side detector + plugin event + optional auto-pause


christianlappin:
### Summary

Paperclip has no anomaly / circuit-breaker layer to detect agent run loops, and no plugin-facing event for them. When a wake-loop occurs (for any root cause — see related issues below), the heartbeat dispatcher treats every iteration as a successful run and there's no signal a notifier plugin (e.g. `paperclip-plugin-discord`) can subscribe to.

I just had a single-agent single-issue loop run **44 times over 8 hours overnight** before I noticed it in the UI the next morning. Paperclip recorded every run as `succeeded`. The `paperclip-plugin-discord` notifier stayed silent because none of its existing subscriptions (`notifyOnIssueCreated`, `notifyOnIssueDone`, `notifyOnAgentError`, `notifyOnApprovalCreated`) fire on this pattern.

The underlying authorship bug that triggered my specific loop is already tracked as #3817. This request is for the orthogonal layer: **detection + notification regardless of root cause**, so that future loops (from #3817, #3980, #4482, #5567, #5433, or anything not yet known) get caught before they burn an overnight.

---

### Evidence — what happened

Self-hosted Paperclip on macOS, `v2026.408.0-canary.4`, `deploymentMode: local_trusted`. Custom `adapterType: "http"` agent (Hermes worker on a separate tailnet machine), bridge-style daemon that processes the wakeup and posts the worker's output back as a comment on the assigned issue.

The daemon was posting the completion comment via the API without an agent JWT, so the actor middleware attributed the comment to the implicit board user. `routes/issues.js` line ~2640:

```js
const selfComment = actorIsAgent && actor.actorId === assigneeId;
const skipAssigneeCommentWake = selfComment || isClosed;
if (assigneeId && !assigneeChanged && (reopened || !skipAssigneeCommentWake)) {
    addWakeup(assigneeId, { ... reason: "issue_commented" ...});
}
```

`selfComment` was `false` → wake fired → daemon ran the worker → posted another comment as the board user → wake fired again. Loop.

Daemon log (excerpt — same `agentId`, same `taskId`, same `wakeReason` for hours):

```
[2026-05-11T08:27:05] dispatch agentId=315fb7a3 runId=01f8d1b2 taskId=5375dcf0 wakeReason=issue_commented
[2026-05-11T08:27:57] hermes completed in 51s, output 260 chars; posting comment
[...]

Issue #5567, opened 2026-05-09
Assignee wakeups should be limited to actionable issue statuses


AnobleSCM:
## Summary

`@paperclipai/server` currently wakes assigned agents from issue paths that are not actionable work. In the 2026.428.0 stable server bundle, the issue route treats any transition out of `backlog` as wakeable and comments only suppress wakes for self/closed cases. That means `backlog -> done` and comments on non-actionable assigned issues can enqueue agent work.

## Affected versions observed

- `paperclipai@2026.428.0` / `@paperclipai/server@2026.428.0` stable
- `paperclipai@2026.508.0-canary.0` / `@paperclipai/server@2026.508.0-canary.0` canary, as staged from npm on 2026-05-09

## Evidence

From a staged `@paperclipai/server@2026.428.0` tarball, `dist/routes/issues.js` contains legacy wake markers:

- `issue.status !== "backlog" && req.body.status !== undefined`
- `assigneeChanged && issue.assigneeAgentId && issue.status !== "backlog"`
- `const skipAssigneeCommentWake = selfComment || isClosed`
- `const skipWake = selfComment || isClosed`

Stable legacy bundle SHA256 observed from npm pack: `8bb3b958a2746c1aae6ebc9a4a2dc3a3e36278c04b511ff775cbba4cc180d7ca`.

A local Atlas hotpatch changed those paths to an explicit actionable-status allowlist and produced patched SHA256 `d5d34f702e84543b23a75ff38462e27a0836351008f0d0df70510a39df9c95e1`. The desired contract is: assigned agents wake only for `todo`, `in_progress`, and `in_review` issue states. `backlog`, `done`, and `cancelled` should not wake agents through assignment/comment/status-transition paths.

## Expected behavior

Add a central predicate, equivalent to:

```js
function isIssueAssigneeWakeableStatus(status) {
  return status === "todo" || status === "in_progress" || status === "in_review";
}
```

Use it for:

- backlog/status transition wake logic
- assignee-change wake logic
- inline comment wake logic
- standalone comment wake logic

## Why it matters

Non-actionable assigned issues should not wake agents. In unattended agent deployments, accidental wakes consume model budget, create misleading run history, and can turn terminal-state issue edits/comments into unplanned work.

## Reproduction sketch

1. Install or inspect `@paperclipai/server@2026.428.0`.
2. Open `dist/routes/issues.js`.
3. Search for the legacy markers above.
[...]

Issue #5433, opened 2026-05-07
[heartbeat] Skip wake when latest comment is by the assignee agent


mcengiiz:
## Problem
When an agent comment is the only new event on an issue, the assigned agent is woken up by `issue_commented` reason. The agent then sees its own previous comment as the latest, has no new board input, and is forced into a no-op heartbeat just to acknowledge.

In a 24h SEO orchestration session at MCAI Labs we observed ~5 such no-op heartbeats per delegated parent issue. Each costs an LLM run, fragments context, and wastes throughput.

## Proposed change
Add a wake-trigger filter in the heartbeat scheduler: if `latestComment.authorType == "agent" AND latestComment.authorId == issue.assigneeAgentId`, skip the wake.

Continue waking on:
- Comments by other agents
- Comments by board users
- Status / interaction state transitions
- Dependency unblock events

## Why this is safe
The agent's own comment can never carry information the agent doesn't already have. The next legitimate trigger (board reply, status flip, interaction resolve) will fire normally.

## Workaround in use
Currently the agent must produce a "no-op" Session Status table acknowledging the wake, which is wasteful.

## Origin
Observed during MCA-112 SEO maximization orchestration (atolye-finder repo, claude_local adapter), 2026-05-07.

Issue #4996, opened 2026-05-02
wake_assignee_on_accept silently coalesces when assignee has active run on same issue


obrunuaf:
## Related

- #3432 — same root-cause class (silent coalesce of `enqueueWakeup` onto an in-flight same-agent run), different trigger (operator-approved approval). The current `RUNNING_ISSUE_WAKE_REASONS_REQUIRING_FOLLOWUP = new Set(["approval_approved"])` set in `dist/services/heartbeat.js` looks like the partial fix that came out of that thread; it does not cover the interaction-resolution path described below.

## Summary

When an issue-thread interaction with `continuationPolicy: "wake_assignee_on_accept"` (or `"wake_assignee"`) is accepted and the issue's assignee **already has a heartbeat run executing on the same issue at the moment of accept**, the wakeup is silently coalesced into the in-flight run via `enqueueWakeup`'s same-agent branch and **no follow-up run is scheduled** when the in-flight run terminates. The agent finishes its heartbeat and never observes the `accept`. The wakeup row is recorded with `status: "coalesced"` in `agent_wakeup_requests`, so it isn't lost from a storage perspective, but it is lost from the agent's perspective.

The contract gap is between `queueResolvedInteractionContinuationWakeup` (in `dist/routes/issues.js`) and `shouldQueueFollowupForRunningIssueWake` (in `dist/services/heartbeat.js`): the former emits the wake with `reason: "issue_commented"` and no `wakeCommentId`, while the latter only forces a follow-up when `wakeCommentId` is set or when `wakeReason ∈ RUNNING_ISSUE_WAKE_REASONS_REQUIRING_FOLLOWUP` (currently just `"approval_approved"`). Interaction-resolution wakes therefore fall through to the silent-coalesce branch.

Severity: **medium** — workaround exists (sleep agents before accepting; or escalate to `request_board_approval`, which already uses `reason: "approval_approved"` and is in the follow-up set).

## Reproduction steps

1. Create any issue and assign it to an agent that uses a long-lived local adapter (e.g. `claude_local`, `cursor_agent`, `codex_local`).
2. From any actor, `POST /api/issues/{issueId}/comments` (or the issue itself) with an interaction `request_confirmation` whose `continuationPolicy = "wake_assignee_on_accept"`. Wait for the agent's heartbeat to start (or trigger one).
[...]

PR #5715, opened 2026-05-11
Add heartbeat-side agent run-loop detector + live event (#5713 minimal scope)

christianlappin:
## Thinking Path

> - Paperclip orchestrates AI agents for zero-human companies, with the heartbeat service driving each agent's run loop.
> - When upstream signals misfire — a comment authored as the wrong principal, a status that's not actually actionable, an unchecked self-assignment — agents wake themselves and end up in a tight oscillation on a single issue.
> - I filed #5713 after an agent ran 44 unintended times on one issue overnight. The related open issues (#3817, #3980, #4482, #3431, #5567, #5433, #4996) are all variants of the same loop pattern.
> - The hard part isn't fixing every upstream trigger — there will always be a new one. The missing piece is a generic detector that observes oscillation in the run stream itself, regardless of which trigger fired it.
> - This pull request adds a heartbeat-side detector that groups recent runs by (agentId, issueId, wakeReason) and emits a `heartbeat.run_loop_suspected` live event when any group exceeds a threshold inside a rolling window.
> - The benefit is a single observation point that catches loops from any cause, with policy decisions (auto-pause, cooldown, plugin notifications) intentionally deferred so this foundation can land without coupling to a specific response strategy.

Refs #5713 — the parent design issue this PR is the minimal-scope foundation of. #5713 lays out the full proposal (detector + policy layer + plugin notification + API endpoint); this PR is the detector + event only, with policy decisions intentionally deferred so the team can design those directly.

Related variants of the same loop pattern: #3817, #3980, #4482, #3431, #5567, #5433, #4996 — all referenced in the Thinking Path above. The detector in this PR provides a single observation point that catches loops from any of these causes, rather than requiring per-trigger one-off fixes.

## What Changed

- New `server/src/services/run-loop-detector.ts` with a pure `detectRunLoop()` function. Groups runs by `(issueId, wakeReason)` parsed from `heartbeat_runs.context_snapshot`, returns the worst offender (highest count, ties broken by earliest first run).
- Wired `maybeEmitRunLoopSignal(agent)` into `startNextQueuedRunForAgent` in `server/src/services/heartbeat.ts`, called after the queued-run sort and before `claimQueuedRun`. Pulls the last 50 runs in the window, runs the detector, logs at `warn` level, publishes the live event via the existing `publishLiveEvent` channel. Wrapped in try/catch — detector failure must not break dispatch.
- Added `heartbeat.run_loop_suspected` to `LIVE_EVENT_TYPES` in `packages/shared/src/constants.ts`.
- Unit tests in `server/src/__tests__/run-loop-detector.test.ts` covering: empty input, sub-threshold groups, threshold reached, window expiration, grouping isolation across `issueId` and `wakeReason`, runs missing both keys, worst-offender selection when multiple groups exceed threshold, and explicit threshold/window overrides.

Defaults are conservative: **5 runs in 30 minutes**. Tunable via `PAPERCLIP_LOOP_DETECTOR_THRESHOLD` and `PAPERCLIP_LOOP_DETECTOR_WINDOW_SEC` env vars. Nothing legitimate I've observed comes close to tripping these.

## Scope explicitly out of this PR

The full proposal in #5713 includes a `loopProtection` agent config (`none` / `cooldown` / `pause_agent`), a `notifyOnRunLoopSuspected` flag in `paperclip-plugin-discord`, and a `/loop-clusters` API endpoint. Those are policy decisions the team will probably want to design directly. This PR is the detector + event only — the foundation everything else builds on.

## Verification

- `pnpm -r typecheck` — clean across the workspace
- `pnpm --filter @paperclipai/server exec vitest run src/__tests__/run-loop-detector.test.ts` — 10/10 pass
- `pnpm --filter @paperclipai/server build` — clean
- `pnpm --filter @paperclipai/shared build` — clean

To exercise the event end-to-end against a running server: queue 5 wakeups for the same (agent, issue) inside 30 minutes with the same `wakeReason` in the wakeup context. The next dispatch tick will emit `heartbeat.run_loop_suspected` on the company live-event stream and log a structured `agent run loop suspected` warning.

## Risks

Low.

- Detector runs once per `startNextQueuedRunForAgent` call and reads at most 50 rows from `heartbeat_runs` filtered by `(agentId, createdAt > windowStart)`. The existing `heartbeat_runs_company_agent_started_idx` index covers `agentId` so this should be cheap, but the filter is on `createdAt` not `startedAt` — happy to add a dedicated index in a follow-up if profiling shows it matters.
- No behavior change to dispatch. The detector observes and emits; it does not block, pause, or reorder runs. Worst case if the detector is buggy: a spurious warn log and a live event with no listeners attached today.
- Detector errors are caught and logged — they cannot break the dispatch path.
- No DB migration, no plugin manifest changes, no new HTTP route, no UI changes.

## Model Used

- Provider: Anthropic

## Checklist


christianlappin:
Hi @cryppadotta @devinfoley — adding you on this one. Opened 2026-05-11, sat for two weeks without a reviewer queued.

Quick state:
- 9/9 CI green (policy, verify, 4 serialized server suites, Canary Dry Run, e2e, security/snyk in your own org @cryppadotta)
- Greptile flagged 2 P2 items on the initial commit; both addressed in commit `68546936` (per-loop dedupe map + cosmetic `else if` restructure). Confidence score 4/5 with "Safe to merge — the detector is read-only and fully guarded".
- 4 files changed, +265/−0 (purely additive — no behavior change to dispatch, no migration, no manifest changes, no new HTTP route).

What it is: a heartbeat-side detector that groups recent runs by `(agentId, issueId, wakeReason)` and emits a `heartbeat.run_loop_suspected` live event when any group exceeds a configurable threshold inside a rolling window. Pure observation point. Policy decisions (auto-pause, cooldown, plugin notifications) intentionally deferred so the team can design those directly on top of this foundation — minimum-scope of #5713.

Why it matters: I filed #5713 after an agent ran 44 unintended times on one issue overnight on my install. The related open issues #3817, #3980, #4482, #3431, #5567, #5433, #4996 are all variants of the same loop pattern, each requiring a one-off upstream-trigger fix. This PR is the generic detector that catches loops from any cause at the dispatch layer — so the next variant doesn't go silent for 8 hours before someone notices.

Defaults are conservative (5 runs in 30 min, both tunable via env vars). Nothing legitimate I've observed comes close to tripping these.

Happy to add a dedicated `createdAt` index in a follow-up if profiling shows the at-most-50-row read needs it. Detector failures are caught and logged so they can't break dispatch.

Available to address review comments whenever you have a few minutes.

christianlappin:
Rebased onto current master — was CONFLICTING, now MERGEABLE. All three commits preserved; the only conflict was an import collision in `heartbeat.ts`.

Verified locally on the rebased branch:
- `server` typecheck green
- `run-loop-detector.test.ts` 10/10
- Integration points re-checked against current master: `publishLiveEvent` signature, `heartbeatRuns.contextSnapshot`, and the `heartbeat.run_loop_suspected` constant all line up.

CI is re-running now. Same ask as before: this is additive-only (detector + live event behind a threshold), no default behavior changes.

PR #5735, opened 2026-05-11
Fix run-recovery uuid crash and stop transitioning agents to error on single failed runs (SIE-441)

simsieg:
## Thinking Path

> - Paperclip orchestrates AI agents for zero-human companies through heartbeat runs that wake agents to act on assigned issues.
> - The heartbeat harness in `server/src/services/heartbeat.ts` resolves run-scoped skill mentions from the active issue's title/description/comments and looks them up in the `company_skills` table before invoking the adapter.
> - `parseSkillMentionHref` treats whatever follows `skill://` as an opaque id, so an issue body containing `[…](skill://paperclip-create-agent/references/agents/coder.md)` fed that slug straight into `inArray(companySkills.id, …)` against a `uuid` column.
> - Postgres rejected the statement with `22P02 invalid input syntax for type uuid`, the adapter execution died with `adapter_failed`, and the heartbeat continuation run crashed. `finalizeAgentStatus` then mapped the failed outcome directly onto `agents.status = 'error'`, taking the assigned agent's entire inbox offline until a human reset it. CTO and DBReliabilityEngineer both got cratered by this class.
> - This pull request lands two narrowly-scoped fixes — one per commit — for SIE-441: the payload bug (filter non-UUID skill mentions) and the resilience bug (errors stay scoped to the run, not the agent).
> - The benefit is that a single bad recovery payload can no longer crash recovery runs or silently disable an agent's whole heartbeat surface.

## What Changed

- `server/src/services/heartbeat.ts`: filter mention ids through a new pure helper `filterValidSkillMentionIds` (backed by `isUuidLike` from `@paperclipai/shared`) before passing them to `inArray(companySkills.id, …)`. Non-UUID slug-style references are silently dropped so the harness proceeds without injecting them rather than crashing the run.
- `server/src/services/heartbeat.ts`: extract `computeAgentStatusAfterRun` as a pure helper that returns \`running\`/\`idle\` only — no \`error\`. \`finalizeAgentStatus\` uses it so a failed/timed_out run settles the agent back to idle while it is otherwise invokable.
- `server/src/__tests__/heartbeat-project-env.test.ts`: regression tests cover (a) the SIE-439 state shape — a comment with `skill://paperclip-create-agent/references/agents/coder.md` is filtered out before any DB query — and (b) failed/timed_out outcomes no longer flip the agent to \`error\`.

## Verification

Each fix is its own commit; review them independently.

\`\`\`
$ node node_modules/vitest/vitest.mjs run server/src/__tests__/heartbeat-project-env.test.ts
 ✓ |@paperclipai/server| src/__tests__/heartbeat-project-env.test.ts (11 tests) 4ms
 Test Files  1 passed (1)
      Tests  11 passed (11)

$ node node_modules/vitest/vitest.mjs run server/src/__tests__/heartbeat-retry-scheduling.test.ts server/src/__tests__/heartbeat-process-recovery.test.ts
 Test Files  2 passed (2)
      Tests  63 passed (63)

$ pnpm typecheck
… all workspaces Done
\`\`\`

\`heartbeat-comment-wake-batching.test.ts\` has one pre-existing failure (\"promotes deferred comment wakes after the active run closes the issue\") that reproduces on \`master\` unchanged — not introduced here.

Manual: re-running the SIE-439 continuation path against this branch no longer raises a uuid error because the slug-style mention is filtered before the lookup; DBReliabilityEngineer's last failure was a different (\`claude_transient_upstream\`) error, but the blast-radius fix prevents that class of single-run failure from putting any agent into \`error\` going forward.

## Risks

- Behavior change for agents that were previously parked in \`status: error\` after a failed run: they will now resolve to \`idle\`. The earlier short-circuit for \`paused\`/\`terminated\` is preserved, so operator-set states are unaffected. \`CONTINUATION_AGENT_STATUSES\` already includes \`error\` for recovery wakes, so no skipped wakes change behavior.
- Non-UUID skill mentions in user content used to crash; they now silently no-op. If an author intended such a mention to inject a skill, the run continues without that skill — same effect as a misspelled slug today.

## Model Used

- Claude (Anthropic), \`claude-opus-4-7\`. Extended-thinking / tool-use mode within the Claude Code harness. No other AI models contributed.

## Checklist

- [ ] If this change affects the UI, I have included before/after screenshots (n/a — server-only)
- [x] I have updated relevant documentation to reflect my changes (root-cause comments in code; no public-API surface change)

### DIFF
--- a/server/src/__tests__/heartbeat-project-env.test.ts
+++ b/server/src/__tests__/heartbeat-project-env.test.ts
@@ -2,6 +2,7 @@ import { describe, expect, it, vi } from "vitest";
 import { buildSkillMentionHref } from "@paperclipai/shared";
 import {
   applyRunScopedMentionedSkillKeys,
+  computeAgentStatusAfterRun,
   extractMentionedSkillIdsFromSources,
   filterValidSkillMentionIds,
   resolveExecutionRunAdapterConfig,
@@ -153,6 +154,37 @@ describe("filterValidSkillMentionIds (SIE-441 regression)", () => {
       ]),
     ).toEqual([]);
   });
+});
+
+describe("computeAgentStatusAfterRun (SIE-441 regression)", () => {
+  // Before the fix `finalizeAgentStatus` flipped the agent to `status: error`
+  // whenever the most recent run failed or timed out. That meant one bad
+  // recovery payload (e.g. the uuid crash above) took the entire CTO agent
+  // offline, blocking ALL of its other heartbeats until a human reset it.
+  // Errors belong to runs; the agent must stay invokable.
+  it.each(["failed", "timed_out"] as const)(
+    "does not flip the agent to error after a single %s run",
+    (outcome) => {
+      expect(
+        computeAgentStatusAfterRun({ runningCount: 0, outcome }),
+      ).toBe("idle");
+    },
+  );
+
+  it("stays running while other runs are still in flight", () => {
+    expect(
+      computeAgentStatusAfterRun({ runningCount: 2, outcome: "failed" }),
+    ).toBe("running");
+  });
+
+  it.each(["succeeded", "cancelled"] as const)(
+    "settles back to idle after a %s run with no other runs",
+    (outcome) => {
+      expect(
+        computeAgentStatusAfterRun({ runningCount: 0, outcome }),
+      ).toBe("idle");
+    },
+  );
 });
 
 describe("applyRunScopedMentionedSkillKeys", () => {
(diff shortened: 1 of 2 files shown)


simsieg:
## Greptile P2 addressed — commit 18c78178

Greptile flagged the `outcome` parameter on `computeAgentStatusAfterRun` as dead weight (the function deliberately ignores it). Agreed — the WHY lives in the comment above the function, so the parameter only misleads future readers.

- `server/src/services/heartbeat.ts` — dropped `outcome` from the helper signature and the single `finalizeAgentStatus` call-site. `finalizeAgentStatus` still receives `outcome` from its callers and publishes it on the `agent.status` live event (line 6246), so external telemetry is unchanged.
- `server/src/__tests__/heartbeat-project-env.test.ts` — collapsed the per-outcome `it.each` into one assertion matching the new signature: "settles back to idle, regardless of outcome." Comment above the test still documents the original SIE-441 bug class.

Targeted vitest + `pnpm --filter @paperclipai/server typecheck` both green locally. Pure cleanup of commit 2a1f93c4 — no behavior change, no impact on the SIE-441 acceptance criteria.

PR #5429, opened 2026-05-07
Add secrets provider vaults and remote import

cryppadotta:
## Thinking Path

> - Paperclip orchestrates AI-agent companies and needs secrets handling to work across local development, hosted operators, and governed agent execution.
> - The affected subsystem is the company-scoped secrets control plane: database schema, server services/routes, CLI workflows, and the Secrets settings UI.
> - The gap was that secrets were local-only and operators could not manage provider vaults or import existing remote references without exposing plaintext.
> - This branch adds provider vault configuration plus an AWS Secrets Manager remote-import path while preserving company boundaries, binding context, and audit trails.
> - I kept the PR to a single branch PR, removed unrelated lockfile/package drift, rebased the full branch onto the current `public-gh/master`, and addressed fresh Greptile findings.
> - The benefit is a reviewable implementation of provider-backed secrets with focused tests covering provider selection, import conflicts, deleted secret reuse, rotation guards, and AWS signing behavior.

## What Changed

- Added provider vault support for company secrets, including provider config storage, default vault handling, health checks, binding usage, access events, and remote import preview/commit.
- Added an AWS Secrets Manager provider using SigV4 request signing, bounded request timeouts, namespace guardrails, cached runtime credential resolution, and external-reference linking without plaintext reads.
- Added Secrets UI surfaces for vault management and remote import, plus CLI/API documentation for setup and operations.
- Stabilized routine webhook secret binding paths and SSH environment-driver fixture bindings discovered during verification.
- Addressed Greptile and CI findings: no lockfile/package drift, monotonic migration metadata, disabled-vault default races, soft-deleted secret hiding/recreate behavior, remove behavior with disabled vaults, soft-deleted external-reference re-import, non-active rotation guards, managed-secret soft deletion through PATCH, and per-call AWS SDK credential client churn.
- Rebased this branch onto `public-gh/master` at `0e1a5828` and force-pushed with lease to keep this as the single PR for the branch.

## Verification

- `git fetch public-gh master`
- `git rebase public-gh/master`
- `git diff --name-only public-gh/master...HEAD | grep '^pnpm-lock\.yaml$' || true` confirmed `pnpm-lock.yaml` is not in the PR diff.
- Confirmed migration ordering: master ends at `0081_optimal_dormammu`; this PR adds `0082_dry_vision` and `0083_company_secret_provider_configs`.
[...]

PR #5738, opened 2026-05-11
Add agent_ssh_identities table and migration (EIG-331)

isparling:
## Thinking Path

> - EIG-330 introduces per-agent SSH identities so workspace runtime and the API can authenticate agents by their public keys. EIG-331 is slice 1 of that effort: the database foundation.
> - The closest existing template is `agent_api_keys` (`packages/db/src/schema/agent_api_keys.ts`) — a company- and agent-scoped credential table. The new `agent_ssh_identities` table follows the same pattern but stores public-key material only.
> - Slice 2 (the service layer) and slice 3 (workspace runtime hookup) live in follow-up issues. This PR deliberately stays within the db package: no API route, no service-layer changes, no reference to the `secrets` table.

## What Changed

- New schema `packages/db/src/schema/agent_ssh_identities.ts` with columns:
  - `id` (uuid pk, default `gen_random_uuid()`)
  - `agentId` → `agents.id` with `ON DELETE CASCADE`
  - `companyId` → `companies.id`
  - `publicKey` (text, not null)
  - `fingerprint` (text, not null, indexed)
  - `algorithm` (text, not null — e.g. `ssh-ed25519`)
  - `label` (text, nullable)
  - `createdAt` (timestamptz, default now())
  - `revokedAt` (timestamptz, nullable)
- Indexes:
  - `agent_ssh_identities_fingerprint_idx` — fast lookup by fingerprint
  - `agent_ssh_identities_agent_id_idx` — fast lookup at workspace start
  - `agent_ssh_identities_company_fingerprint_uq` (unique) on `(company_id, fingerprint)` — same key cannot be double-registered in one company.
- Migration `packages/db/src/migrations/0084_agent_ssh_identities.sql` — additive, idempotent (`CREATE TABLE IF NOT EXISTS`, `CREATE INDEX IF NOT EXISTS`, FK adds guarded by `pg_constraint` lookups). Same style as `0066_issue_tree_holds.sql`.
- Journal updated to record migration 0084.
- Drizzle snapshot `meta/0084_snapshot.json` regenerated. This also picks up pre-existing snapshot drift from PR #5429 (0082/0083 were hand-tuned SQL and never updated the snapshot), so the next `drizzle-kit generate` will diff against an accurate baseline.
- New schema export added to `packages/db/src/schema/index.ts`.
- New unit test `packages/db/src/agent_ssh_identities.test.ts` exercises:
  1. Insert + read back via the typed Drizzle schema (acceptance criterion).
  2. Unique `(company_id, fingerprint)` constraint rejects duplicates.
  3. `ON DELETE CASCADE` on `agentId` removes identities when the agent is deleted.

## Out of Scope

- Secret storage — public keys only; no reference to the `secrets` table.
- API routes, service layer, workspace runtime wiring — those are slices 2 and 3.

## Verification

- `pnpm --filter @paperclipai/db check:migrations` — passes.
- `pnpm --filter @paperclipai/db build` — passes.
- `pnpm --filter @paperclipai/db typecheck` — passes.
- `pnpm --filter @paperclipai/db vitest run` — 22 tests pass, including the 3 new ones; each test boots embedded Postgres from scratch and applies every migration (0000–0084), so a fresh-DB migrate is exercised end-to-end.

## Risks

- Low. The migration is additive and idempotent; nothing existing reads from or writes to the new table yet. The snapshot regeneration could in theory affect the next `drizzle-kit generate` run, but it brings the snapshot back into agreement with what 0082/0083 already shipped in production, which is the correct state.

## Model Used

- Anthropic Claude Opus 4.7 (1M context) via Paperclip CTO agent (`claude-opus-4-7[1m]`), default reasoning effort, with shell, file-edit, and gh tool use.

## References

- Parent issue: EIG-330
- Plan: EIG-330 plan (slice 1)
- Pattern reference: `packages/db/src/schema/agent_api_keys.ts`

## Checklist

- [ ] If this change affects the UI, I have included before/after screenshots — n/a (db only)
- [x] I have updated relevant documentation to reflect my changes — n/a; the schema is self-documenting

### DIFF
--- a/packages/db/src/agent_ssh_identities.test.ts
+++ b/packages/db/src/agent_ssh_identities.test.ts
@@ -1,0 +1,166 @@
+import { afterEach, describe, expect, it } from "vitest";
+import { eq } from "drizzle-orm";
+import { drizzle } from "drizzle-orm/postgres-js";
+import postgres from "postgres";
+import { agents, agentSshIdentities, companies } from "./schema/index.js";
+import {
+  getEmbeddedPostgresTestSupport,
+  startEmbeddedPostgresTestDatabase,
+} from "./test-embedded-postgres.js";
+
+const cleanups: Array<() => Promise<void>> = [];
+const embeddedPostgresSupport = await getEmbeddedPostgresTestSupport();
+const describeEmbeddedPostgres = embeddedPostgresSupport.supported ? describe : describe.skip;
+
+async function createTempDatabase(): Promise<string> {
+  const db = await startEmbeddedPostgresTestDatabase("paperclip-db-ssh-identities-");
+  cleanups.push(db.cleanup);
+  return db.connectionString;
+}
+
+function connect(connectionString: string) {
+  const sql = postgres(connectionString, { max: 1, onnotice: () => {} });
+  const db = drizzle(sql, { schema: { agents, agentSshIdentities, companies } });
+  cleanups.push(async () => {
+    await sql.end();
+  });
+  return db;
+}
+
+afterEach(async () => {
+  while (cleanups.length > 0) {
+    const cleanup = cleanups.pop();
+    await cleanup?.();
+  }
+});
+
+if (!embeddedPostgresSupport.supported) {
+  console.warn(
+    `Skipping embedded Postgres agent_ssh_identities tests on this host: ${embeddedPostgresSupport.reason ?? "unsupported environment"}`,
+  );
+}
+
+describeEmbeddedPostgres("agent_ssh_identities", () => {
+  it(
+    "inserts and reads a row back via the typed schema",
+    async () => {
+      const connectionString = await createTempDatabase();
+      const db = connect(connectionString);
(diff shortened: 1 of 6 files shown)


isparling:
Recording the close rationale here for anyone landing on this PR:

Per internal discussion, the per-agent SSH identity feature is moving entirely out-of-tree.

On a careful re-read of the v4 plan, the feature never required an inbound SSH auth server — public keys are used as a path on disk + `user.signingkey` value, and verification happens at GitHub. So a dedicated `agent_ssh_identities` table is not load-bearing. Everything it would carry can live in the existing `agents.runtime_config` jsonb column, and workspace bootstrap (write pubkey file, set `GIT_SSH_COMMAND`, four `git config` calls) is done by `SessionStart` hook

Net effect: zero changes to paperclip-upstream. 

Sorry for the churn. No code from this PR will be carried forward.

PR #5743, opened 2026-05-11
feat: add AppError class and sanitize error responses (TEC-2754)

glovario:
## Summary

- Adds `AppError` class in `server/src/errors.ts` with `userMessage` (client-safe) and `internalMessage` (server-only logged) fields, plus optional `statusCode`
- Updates `server/src/middleware/error-handler.ts` to sanitize all error responses: no raw `err.message` ever reaches the client; all responses include a UUID `requestId` for correlation
- Fixes 3 raw `err.message` leaks from MulterError in `server/src/routes/assets.ts` and `server/src/routes/issues.ts`

## Details

The error handler now classifies errors into three branches:
- `AppError`: returns `userMessage` to client, logs `internalMessage` server-side
- `HttpError`: returns safe status-based text (not raw message); logs details server-side for 5xx
- `ZodError`: returns "Validation error" with details
- Unhandled: returns "An unexpected error occurred"

Every response includes a `requestId` (UUID v4) that matches the server-side log entry.

## Test plan

- [x] Unit tests: `server/src/__tests__/error-handler.test.ts` (2 tests pass)
- [x] No raw `err.message` in any HTTP response (`grep -r "\.json.*err\.message" server/src/routes/` returns empty)
- [x] `errorHandler` registered as last middleware in `app.ts`

Closes TEC-2754

### DIFF
--- a/server/src/__tests__/error-handler.test.ts
+++ b/server/src/__tests__/error-handler.test.ts
@@ -32,7 +32,7 @@ describe("errorHandler", () => {
     errorHandler(err, req, res, next);
 
     expect(res.status).toHaveBeenCalledWith(500);
-    expect(res.json).toHaveBeenCalledWith({ error: "Internal server error" });
+    expect(res.json).toHaveBeenCalledWith({ error: "An unexpected error occurred", requestId: expect.any(String) });
     expect(res.err).toBe(err);
     expect(res.__errorContext?.error?.message).toBe("boom");
   });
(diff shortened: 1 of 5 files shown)


starlein (changes requested):
Thanks for tightening the 5xx/unhandled-error path and for adding request IDs. I found one blocking regression.

`server/src/middleware/error-handler.ts:77` now replaces every `HttpError` response with only a status-generic string (`Forbidden`, `Validation error`, etc.). That strips existing 4xx messages that are already intentionally client-facing and required for operator/agent remediation, for example `Only CEO agents can manage company exports`, `Unknown adapter type: ...`, and `User does not have access to this company`. The PR CI is failing across the serialized server suites and e2e for this exact behavior: route tests now receive generic strings instead of the actionable messages they assert.

Please keep raw/internal details sanitized for 5xx and unsafe sources, but preserve explicit public/client-safe messages for 4xx errors, or introduce an allowlisted/safe-public error type and migrate the call sites that need user-actionable diagnostics. The current blanket `HttpError` masking is a breaking API/UX regression.

Verification reviewed:
- `gh pr diff 5743 --repo paperclipai/paperclip`
- `gh pr checks 5743 --repo paperclipai/paperclip --watch=false` shows failing serialized server suites and e2e
- CI log examples show received `"Forbidden"` where tests expect `Only CEO agents...`, `Board access required`, `Instance admin...`, and received `"Validation error"` where tests expect `Unknown adapter type...`

PR #5749, opened 2026-05-11
fix: prevent invite page crash caused by React Query cache collision in CompanyProvider

MorelSerge:
## Thinking Path

Traced the `TypeError: Te.some is not a function` crash from the invite landing page. The minified variable `Te` is `companiesQuery.data` in `InviteLanding.tsx`. It calls `.some()` expecting a `Company[]`, but received `{ companies: Company[], unauthorized: boolean }` — a plain object — so `.some` is undefined.

Root cause: `CompanyProvider` stores a `CompanyListResult` wrapper object under `queryKeys.companies.all`. `InviteLanding` uses the **same cache key** but expects a plain array. When `CompanyProvider` runs first, its cached value poisons `InviteLanding`'s query, producing the crash.

`CompanyListResult` is only referenced in `CompanyContext.tsx`, so the fix is self-contained.

## What Changed

**`ui/src/context/CompanyContext.tsx`**

- Removed the `CompanyListResult` wrapper type (`{ companies: Company[], unauthorized: boolean }`)
- Changed `useQuery<CompanyListResult>` → `useQuery<Company[]>` so `queryKeys.companies.all` always holds a plain `Company[]` array in the React Query cache
- Moved the `unauthorized` flag to a `useState<boolean>` that is set inside the `queryFn` — keeping the same behaviour without polluting the shared cache key

No other files changed; all other consumers of `queryKeys.companies.all` already expected `Company[]`.

## Verification

1. Open the invite landing page (`/invite/<token>`) while logged in — no longer crashes
2. `InviteLanding`'s `companiesQuery.data.some(...)` receives a plain array and resolves correctly
3. `CompanyProvider` still correctly tracks the 401-unauthorized state via `companyListUnauthorized` state
4. All existing company-selection logic (`resolveBootstrapCompanySelection`, `shouldClearStoredCompanySelection`) is unchanged

## Risks

Low. The change is isolated to one file, removes a type wrapper rather than adding logic, and the public API surface of `CompanyContextValue` is unchanged. The only observable difference is that `queryKeys.companies.all` now caches `Company[]` instead of `CompanyListResult` — which is what every consumer already assumed.

## Model Used

claude-sonnet-4-6

## Checklist

- [x] Fix targets the root cause (cache key collision), not a symptom
- [x] No other files need updating (`CompanyListResult` was only used in `CompanyContext.tsx`)
- [x] Existing unit-testable helpers (`resolveBootstrapCompanySelection`, `shouldClearStoredCompanySelection`) are untouched
- [x] No new dependencies introduced

### DIFF
--- a/ui/src/context/CompanyContext.test.tsx
+++ b/ui/src/context/CompanyContext.test.tsx
@@ -172,10 +172,7 @@ describe("CompanyProvider", () => {
 
   it("replaces a stale stored company id with the first loaded company", async () => {
     localStorage.setItem("paperclip.selectedCompanyId", "stale-company");
-    queryClient.setQueryData(queryKeys.companies.all, {
-      companies: [makeCompany("company-1")],
-      unauthorized: false,
-    });
+    queryClient.setQueryData(queryKeys.companies.all, [makeCompany("company-1")]);
     mockCompaniesApi.list.mockImplementation(() => new Promise(() => {}));
     const seen: Array<string | null> = [];
 
--- a/ui/src/context/CompanyContext.tsx
+++ b/ui/src/context/CompanyContext.tsx
@@ -14,7 +14,6 @@ import { ApiError } from "../api/client";
 import { queryKeys } from "../lib/queryKeys";
 import type { CompanySelectionSource } from "../lib/company-selection";
 type CompanySelectionOptions = { source?: CompanySelectionSource };
-type CompanyListResult = { companies: Company[]; unauthorized: boolean };
 
 interface CompanyContextValue {
   companies: Company[];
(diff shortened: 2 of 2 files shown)


MorelSerge:
**CI fix pushed** (commit 74b7132)

The  job failed because one test pre-seeded the React Query cache with the old `CompanyListResult` wrapper format (`{ companies, unauthorized }`). After the source fix, the cache holds a plain `Company[]` instead — so the seed needed updating too.

Changed `queryClient.setQueryData(queryKeys.companies.all, { companies: [...], unauthorized: false })` → `queryClient.setQueryData(queryKeys.companies.all, [...])` in `CompanyContext.test.tsx`. No other tests affected.

PR #5776, opened 2026-05-12
Parse Anthropic rate_limit_event for structured retry-after

adlai88:
## What

`claude` CLI emits a structured `rate_limit_event` stream line when an account is over its Max-plan window:

```json
{"type":"rate_limit_event","rate_limit_info":{"status":"rejected","resetsAt":1778565600,"rateLimitType":"five_hour","overageStatus":"rejected","overageDisabledReason":"org_level_disabled_until","isUsingOverage":false}}
```

`resetsAt` is Unix epoch seconds — machine-readable. The accompanying assistant message is now `"You've hit your limit · resets 2pm (Asia/Shanghai)"`, which does NOT match the legacy reset-time regex (prefix alternatives required `"out of extra usage"`, `"usage limit reached"`, etc.).

## Why this matters

The classifier (`isClaudeTransientUpstreamError`) still fires correctly — `parsed.errors[].error: "rate_limit"` is in the haystack and matches the existing regex. But `extractClaudeRetryNotBefore` returns `null` because the text regex's prefix alternatives don't match `"You've hit your limit"`.

Net effect: the run is classified as `claude_transient_upstream` but `retryNotBefore` is never set. The heartbeat scheduler falls back to default short backoff and retries every ~7 min — each retry hits the same cap, fails immediately, and burns more of the already-exhausted window.

We hit this in production today: one ticket triggered 4 cap-hit retries in 32 minutes after the human had stopped working, all from default backoff with no `retryNotBefore` guidance.

## Fix

Two parts, in priority order:

**1. (Primary) Consume the structured `rate_limit_event`.** `parseClaudeStreamJson` now reads this event type and surfaces `rate_limit_info.resetsAt` as `rateLimitResetAtUnix` on its return. `extractClaudeRetryNotBefore` accepts this as an optional input and prefers it (when in the future) over the text regex. A stale (past) `resetsAt` falls through to the regex path, so old behavior is preserved.

**2. (Defense in depth) Add `"you've hit your limit"` to both regexes.** If the structured event is absent or filtered, the older text-only path still detects and extracts.

## Files changed

- `packages/adapters/claude-local/src/server/parse.ts` — regex updates, `parseClaudeStreamJson` extension, `extractClaudeRetryNotBefore` priority
- `packages/adapters/claude-local/src/server/parse.test.ts` — 6 new test cases
- `packages/adapters/claude-local/src/server/execute.ts` — both call sites thread `parsedStream.rateLimitResetAtUnix` through

## Test plan

- [x] `pnpm vitest run src/server/parse.test.ts` — 15/15 (9 existing + 6 new)
- [x] Full `claude-local` suite — 18/18
- [x] `server/src/__tests__/claude-local-execute.test.ts` + `heartbeat-retry-scheduling.test.ts` — 19/19
- [x] `pnpm typecheck` clean

## Cross-adapter note

`codex-local` adapter may have an equivalent pattern (the file is structurally similar). Not touched in this PR — happy to file a follow-up if you'd like the same treatment there.

### DIFF
--- a/packages/adapters/claude-local/src/server/execute.ts
+++ b/packages/adapters/claude-local/src/server/execute.ts
@@ -628,6 +628,7 @@ export async function execute(ctx: AdapterExecutionContext): Promise<AdapterExec
             stdout: proc.stdout,
             stderr: proc.stderr,
             errorMessage: fallbackErrorMessage,
+            rateLimitResetAtUnix: parsedStream.rateLimitResetAtUnix,
           })
         : null;
       const errorCode = loginMeta.requiresLogin
@@ -709,6 +710,7 @@ export async function execute(ctx: AdapterExecutionContext): Promise<AdapterExec
           stdout: proc.stdout,
           stderr: proc.stderr,
           errorMessage,
+          rateLimitResetAtUnix: parsedStream.rateLimitResetAtUnix,
         })
       : null;
     const resolvedErrorCode = loginMeta.requiresLogin
--- a/packages/adapters/claude-local/src/server/parse.test.ts
+++ b/packages/adapters/claude-local/src/server/parse.test.ts
@@ -2,6 +2,7 @@ import { describe, expect, it } from "vitest";
 import {
   extractClaudeRetryNotBefore,
   isClaudeTransientUpstreamError,
+  parseClaudeStreamJson,
 } from "./parse.js";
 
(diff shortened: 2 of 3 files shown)


adlai88:
Closing this for now — going to handle the fix internally on our side. Will revisit upstream once we've validated the approach in production and have a tighter, less-context-laden PR. Thanks for maintaining the project.

PR #5780, opened 2026-05-12
fix(ui): fix message attribution for agent-posted comments with user author IDs

devinfoley:
## Thinking Path

> - Paperclip’s issue chat is an audit surface: reviewers need to trust who actually authored a message.
> - Some historical agent comments were persisted with `authorUserId` and no surviving `createdByRunId`, so the UI rendered real agent output as if it came from the board user.
> - A pure timestamp-window fallback is too risky because human reviewers can comment while agents are running.
> - The safe recovery path is to derive attribution only when the server can prove it from same-issue run logs that include the exact posted comment id, then let the chat renderer prefer that recovered agent attribution.
> - This keeps historical threads trustworthy without mutating old database rows or guessing in ambiguous cases.

## What Changed

- Added shared `IssueComment` fields for derived attribution so server and UI can carry recovered `derivedAuthorAgentId`, `derivedCreatedByRunId`, and `derivedAuthorSource` consistently.
- Added server-side attribution recovery in `server/src/services/issues.ts` that reads same-issue run logs and only derives agent authorship when a run log contains the exact `comment id: ...` emitted during posting.
- Updated issue chat rendering in `ui/src/lib/issue-chat-messages.ts` to prefer direct agent authorship, then activity-log `runAgentId`, then the server-derived attribution.
- Removed the unsafe UI-only run-window fallback from `ui/src/pages/IssueDetail.tsx` so human comments posted during an active run are not silently relabeled as agent output.
- Added regression coverage for both the run-log derivation path and the chat-rendering fallback behavior.
- Bounded server-side run-log enrichment to 8 concurrent reads per request and removed the unused `issueCommentSchema` declaration during PR cleanup.

## Verification

- `pnpm exec vitest run ui/src/lib/issue-chat-messages.test.ts server/src/__tests__/issues-service.test.ts`
- `pnpm test:run:general`
- Live validation on May 12, 2026 in `PAPA-322`: confirmed the previously misattributed historical comments on `PAPA-316` now render as Claude-authored on `http://goldie.gerbil-company.ts.net:3100`.
- Reviewer check: open `PAPA-316` in the running instance and confirm historical comments such as `## Investigation: exe.dev 422 + codex re-test` render under Claude instead of the board user.

## Risks

- Low risk. The change is scoped to comment attribution recovery and rendering.
- Derived attribution is intentionally conservative: if there is no exact run-log proof, the comment remains user-authored instead of guessing.
- Run-log recovery depends on retained same-issue logs, so older comments without that evidence remain unchanged.

## Model Used

- OpenAI Codex via the Paperclip `codex_local` adapter (GPT-5-class coding agent with tool use in the local Paperclip runtime; the exact deployment/model ID is not surfaced by this workspace).

## Checklist


devinfoley:
@greptile review

devinfoley:
@greptile review

PR #5784, opened 2026-05-12
Fix DB backup rotation ENOSPC (per-day daily-tier coalesce + free-space guard)

imvimm:
## Summary

The scheduled-backup daily retention tier kept *every* backup within the
`dailyDays` window with no per-day coalescing. With the default hourly cadence
that means up to `dailyDays * 24` files accumulate before the weekly tier
sees anything. We observed this in the wild as a 164-file, ~1.8 GB
`paperclip-*.sql.gz` heap that filled the data volume and tripped `ENOSPC`
mid-backup, surfacing as an unhandled rejection that took the server down.

This PR ships three coordinated changes as a single coherent fix.

### 1. `packages/db/src/backup-lib.ts`
- New optional `hourlyHours` tier (default `24`): every backup inside the
  hourly window is retained for fast recovery.
- Daily tier now coalesces to the newest backup **per calendar day**, so
  hourly schedules no longer pile up unbounded.
- `pruneOldBackups` returns `{ kept, pruned }` and is exported for unit tests.
- `RunDatabaseBackupResult` gains `keptCount` alongside `prunedCount`.
- `createBufferedTextFileWriter` funnels the `openFile()` promise into the
  `pendingWrite` chain so an ENOSPC at open time surfaces via
  `drain` / `close` rather than escaping as an unhandled rejection.

### 2. `server/src/index.ts`
- Top-of-loop `statfs` precheck refuses a scheduled backup when free space
  drops below `DATABASE_BACKUP_FREE_SPACE_THRESHOLD_BYTES` (default 500 MB,
  overridable via `PAPERCLIP_DB_BACKUP_FREE_SPACE_THRESHOLD_BYTES`). The
  scheduled trigger logs and returns `null`; the manual trigger surfaces 409.
- Completion log line now includes `keptCount` so operators can see the tier
  outcome of each run at a glance.

### 3. `packages/db/src/backup-lib.test.ts`
Five new unit tests covering:
- 168-file hourly fixture → fix collapses to ~24–32 kept, 132–144 pruned
- Hourly window untouched when all files are < 24h old
- Default `hourlyHours` fallback when retention omits the field (forward
  compat with the existing shared `BackupRetentionPolicy` shape)
- Prefix filtering (unrelated files left alone)
- `createBufferedTextFileWriter` surfaces \`openFile\` rejections via \`drain\`

### Dry-run against the affected directory

\`\`\`
Backup dir: /…/instances/default/data/backups/
Files found: 164
Retention:   {\"dailyDays\":7,\"weeklyWeeks\":4,\"monthlyMonths\":1,\"hourlyHours\":24}

== master (pre-fix) ==
  total=164  kept=163  deleted=1
  kept_size_MB=1797.7  deleted_size_MB=2.0

== this PR ==
  total=164  kept=32   deleted=132
  kept_size_MB=429.9   deleted_size_MB=1369.7
\`\`\`

## Test plan

- [x] \`pnpm --filter @paperclipai/db exec vitest run src/backup-lib.test.ts\` — 9 tests pass (5 new + 4 pre-existing)
- [x] \`pnpm --filter @paperclipai/server exec vitest run src/__tests__/instance-database-backups-routes.test.ts\` — 5 tests pass; \`keptCount\` flows through the manual-backup route untouched.
- [x] \`pnpm --filter @paperclipai/server exec vitest run src/__tests__/server-startup-feedback-export.test.ts\` — 6 tests pass.
- [x] \`pnpm --filter @paperclipai/db typecheck\`
- [x] \`pnpm --filter @paperclipai/server typecheck\`

## Compatibility notes

- \`BackupRetentionPolicy\` in \`@paperclipai/db\` adds an **optional** field; callers that pass the existing shared shape (no \`hourlyHours\`) keep working and get the 24-hour default.
- \`RunDatabaseBackupResult\` adds \`keptCount\` (additive); existing consumers (\`cli/src/commands/db-backup.ts\`, route handlers) continue to compile and run unchanged.

### DIFF
--- a/packages/db/src/backup-lib.test.ts
+++ b/packages/db/src/backup-lib.test.ts
@@ -164,6 +164,48 @@ describe("pruneOldBackups", () => {
     expect(result.kept + result.pruned).toBe(30);
   });
 
+  it("only counts successful unlinks in the returned pruned total", () => {
+    const dir = createTempDir("paperclip-prune-unlink-fail-");
+    const anchorMs = Date.UTC(2026, 4, 12, 12, 0, 0);
+    // 30 hourly entries means several files older than the 24h hourly window
+    // are scheduled for deletion after per-day coalescing in the daily tier.
+    generateHourlyFixture(dir, 30, anchorMs);
+
+    // Remove write permission on the directory so unlinkSync throws EACCES /
+    // EPERM for every scheduled deletion. The prune should still complete
+    // without crashing and the returned `pruned` count must reflect the zero
+    // actual deletions rather than the scheduled-intent count.
+    const originalMode = fs.statSync(dir).mode;
+    fs.chmodSync(dir, 0o555);
+    cleanups.push(() => {
+      try {
+        fs.chmodSync(dir, originalMode);
+      } catch {
+        // best-effort cleanup; the parent afterEach rmSync handles the rest
+      }
+    });
+
+    let result: { kept: number; pruned: number };
+    try {
+      result = pruneOldBackups(
+        dir,
+        { dailyDays: 7, weeklyWeeks: 4, monthlyMonths: 1, hourlyHours: 24 },
+        "paperclip",
+        () => anchorMs,
+      );
+    } finally {
+      fs.chmodSync(dir, originalMode);
+    }
+
+    // Every file should still be on disk because unlinkSync failed for all of them.
+    const remaining = fs.readdirSync(dir).filter((name) => name.endsWith(".sql.gz"));
+    expect(remaining.length).toBe(30);
+    // Pruned must reflect actual deletions, not scheduled intent.
+    expect(result.pruned).toBe(0);
+    // kept reflects the retention plan (unchanged by unlink failures).
+    expect(result.kept).toBeGreaterThan(0);
+  });
+
   it("ignores files that do not match the filename prefix", () => {
     const dir = createTempDir("paperclip-prune-prefix-");
     const anchorMs = Date.UTC(2026, 4, 12, 12, 0, 0);
--- a/packages/db/src/backup-lib.test.ts
+++ b/packages/db/src/backup-lib.test.ts
@@ -3,7 +3,7 @@ import os from "node:os";
 import path from "node:path";
 import { afterEach, describe, expect, it } from "vitest";
 import postgres from "postgres";
-import { createBufferedTextFileWriter, runDatabaseBackup, runDatabaseRestore } from "./backup-lib.js";
+import { createBufferedTextFileWriter, pruneOldBackups, runDatabaseBackup, runDatabaseRestore } from "./backup-lib.js";
 import { ensurePostgresDatabase } from "./client.js";
 import {
   getEmbeddedPostgresTestSupport,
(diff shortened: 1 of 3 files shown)


timlevel:
Code review feedback from Paperclip CodeReview agent (LTC-312):

Overall this is excellent work addressing the critical ENOSPC issue! The changes are well-designed and thoroughly tested. However, there are three issues that need to be addressed before merging:

1. **Server error handling bug**: The `throw conflict(...)` in server/src/index.ts is being swallowed by the catch block, which defeats the purpose of returning a 409 for manual triggers on a full disk.

2. **Pruned count accuracy**: The pruned count may overstate actual deletions when unlinkSync fails silently. Consider counting only successful deletions.

3. **Arithmetic edge case**: There's a potential edge case with the bavail × bsize calculation that could cause the guard to silently pass.

Please address these issues before merging. Detailed suggestions are available in the Greptile comments.

timlevel (changes requested):
## CTO Review — Request Changes

### P1 — `throw conflict(...)` swallowed by enclosing `catch` — Must fix

The `throw conflict(...)` for manual triggers is inside the same `try {}` that wraps `statfs`, so the outer `catch (statErr)` catches the intended 409 and falls through to `logger.warn + proceed`, silently defeating the free-space guard. This is the exact scenario the PR is trying to prevent.

**Fix:** Either rethrow non-system errors in the catch, or hoist the free-space check/throw outside the try/catch.

### P2 — `pruned` count overstates deletions when `unlinkSync` fails — Should fix

`toDelete.length` is scheduled deletions, not actual. If `unlinkSync` silently fails, operators see inflated counts. Count only successful deletes as suggested in the inline comment.

### P3 — `bavail * bsize` arithmetic edge case — Optional

The existing `Number.isFinite(freeBytes)` check already skips the guard for NaN/Infinity. Adding `freeBytes > 0` would be more explicit but isn't necessary since a 0 or negative value correctly skips the guard. The `bigint` suggestion is unnecessary for Paperclip backup sizes.

### Architecture — Approved

- Per-day coalescing correctly fixes the unbounded accumulation
- Hourly tier preserves recent-backup granularity
- `createBufferedTextFileWriter` funneling `openFile` rejections is the right fix
- `statfs` pre-check concept is correct, only the catch scoping is wrong

**Requesting changes for P1 (blocking) and P2 (should-fix).**

imvimm:
Thanks for the careful review. Pushed [`3277c31f`](https://github.com/paperclipai/paperclip/commit/3277c31f) addressing P1 and P2.

**P1 — `throw conflict(...)` swallowed by enclosing `catch` (must fix):** hoisted the free-space threshold check outside the `statfs` try/catch in `server/src/index.ts`. The statfs call still lives in its own try/catch (logs and proceeds when `statfs` itself is unsupported / missing path), but the threshold comparison and the manual-trigger `throw conflict(...)` now run on the outer scope. A full-disk manual backup correctly returns 409 again instead of falling through to the write path.

**P2 — `pruned` count overstating deletions (should fix):** `pruneOldBackups` now increments a `prunedCount` only on successful `unlinkSync`, so the returned `{kept, pruned}` reflects ground truth. Failures (EACCES, EBUSY, etc.) are still swallowed locally so a single bad file can't crash the prune, but they no longer inflate the log line. Added a regression test in `packages/db/src/backup-lib.test.ts` that chmods the backup directory read-only so every scheduled unlink fails and asserts `result.pruned === 0` and that all 30 fixture files remain on disk.

**P3 — `bavail × bsize` edge case (optional):** left as is, per the CTO review note. `Number.isFinite(freeBytes)` already skips the guard on NaN/Infinity, and a zero/negative product correctly skips the guard for our backup sizes (the existing log line still records the observed value for ops).

Validation: `packages/db` vitest 10/10 ✓ (includes the new P2 regression test); `server` route tests 5/5 ✓. CI should re-run automatically.

imvimm:
Follow-up pushed in [`bb566100`](https://github.com/paperclipai/paperclip/commit/bb566100) for the remaining arithmetic edge case:

- `server/src/index.ts` now uses `statfs(path, { bigint: true })` and multiplies `bavail * bsize` as bigint before converting to a bounded number.
- The threshold check is explicitly `Number.isFinite(freeBytes) && freeBytes > 0 && freeBytes < threshold`.

Current status of review items:

- P1 fixed in `3277c31f`: manual-trigger `throw conflict(...)` is outside the `statfs` catch and will not be swallowed.
- P2 fixed in `3277c31f`: `pruned` counts only successful `unlinkSync` calls, covered by regression test.
- P3 fixed in `bb566100`: bigint free-space arithmetic plus explicit positive finite guard.

Validation rerun locally:

- `pnpm exec vitest run packages/db/src/backup-lib.test.ts --testNamePattern 'createBufferedTextFileWriter|pruneOldBackups'`

PR #5790, opened 2026-05-12
feat(plugin-kubernetes): self-hostable Kubernetes sandbox provider (stage 1/3: plugin package)

stubbi:
## Thinking Path

> - Sandbox providers are the seam that lets agent runs execute in isolated environments; today the only first-party remote provider is Daytona, a hosted third-party service
> - Self-hosters running Paperclip on their own infrastructure (often Kubernetes already) have no first-party way to run agent sandboxes on a cluster they control
> - That gap matters for teams with data-residency, sovereignty, or cost constraints who cannot or will not send workloads to a hosted sandbox service
> - This pull request adds a Kubernetes sandbox-provider plugin as a standalone, workspace-excluded package: it implements every SandboxProvider hook the Daytona provider does, on infrastructure the operator owns
> - The benefit is that any Paperclip deployment with a Kubernetes cluster gets multi-tenant, network-isolated, quota-bounded agent sandboxes with zero new external dependencies

No existing issue. Following the feature template:

- **Problem:** Paperclip's remote sandbox execution requires a hosted third-party provider. Self-hosters cannot run agent sandboxes on their own Kubernetes clusters with a first-party provider.
- **Proposed solution:** A `@paperclipai/plugin-kubernetes` sandbox-provider plugin with two backends: long-lived sandboxes via the [kubernetes-sigs/agent-sandbox](https://github.com/kubernetes-sigs/agent-sandbox) CRD (multi-command exec, adapter-install pattern) and one-shot `batch/v1` Jobs (stable APIs only, no extra controllers).
- **Alternatives considered:** Driving kubectl from a generic shell provider (no lifecycle/lease semantics), or requiring a hosted provider (exactly the constraint this removes).

## What Changed

This is **stage 1 of 3** of a staged contribution (direction agreed with maintainers): the plugin package alone. Stage 2 (server integration: lease params, provider registration) and stage 3 (agent runtime images + CI) are companion PRs that will be cross-linked from a comment here.

- New package `packages/plugins/sandbox-providers/kubernetes` (workspace-excluded, like the path already carved out in `pnpm-workspace.yaml`): src, unit + kind integration tests, operator prerequisite manifests, README, smoke-test guide
- Implements the full SandboxProvider hook surface the Daytona provider implements: `validateConfig`, `probe`, `acquireLease`, `resumeLease`, `releaseLease`, `destroyLease`, `realizeWorkspace`, `execute`
- Two backends: `sandbox-cr` (default; long-lived pod via the agent-sandbox `Sandbox` CR, supports multi-command exec) and `job` (one-shot `batch/v1` Job; nothing beyond k8s 1.27+ required)
- Per-run adapter resolution: one environment serves mixed harnesses; the per-run `adapterType` hint is read through a local optional type extension, so the plugin typechecks and builds against the current plugin SDK and simply falls back to the environment's configured default adapter until stage 2 lands
- Exec-env wrapping: the Kubernetes exec API carries no environment, so commands are wrapped to receive the run's env
- Fast-upload interception for workspace realization, scoped per lease
- Per-tenant isolation: derived namespace per company, RBAC, ResourceQuota, restricted-PSS pod security (runAsNonRoot, drop ALL, seccomp RuntimeDefault, no SA token automount)
- Network egress policy in two flavors: native `NetworkPolicy` and `CiliumNetworkPolicy` (FQDN allowlists)
- Image allowlist with glob matching, registry override, and per-run image override validation
- Per-run Kubernetes Secrets carrying agent credentials, ownerRef'd to the Job or Sandbox CR for cascade GC

## Verification

- Standalone build, exactly as the README documents:
  ```bash
  cd packages/plugins/sandbox-providers/kubernetes
  pnpm install --ignore-workspace
  pnpm test        # 147 unit tests, 17 files, all green
  pnpm typecheck   # clean against the in-repo plugin SDK on master
  pnpm build       # dist/ emitted, manifest + worker entrypoints present
  ```
- A kind-cluster end-to-end integration test is included (`RUN_K8S_INTEGRATION_TESTS=1 pnpm test test/integration/end-to-end-run.test.ts`)
- Beyond CI: this provider has been verified in a production multi-tenant deployment against five harnesses (opencode, pi, codex, gemini, claude code) with real billed runs

## Risks

- **Zero behavior change for any existing deployment.** The package is workspace-excluded; nothing in the server imports or loads it until stage 2's integration lands. No existing code paths are touched.
- The default `sandbox-cr` backend depends on an alpha CRD (`agents.x-k8s.io/v1alpha1`); the README flags this and the `job` backend uses only stable APIs as a fallback.
- Risk surface is confined to deployments that explicitly install and configure the plugin.
- The default runtime images (`ghcr.io/paperclipai/agent-runtime-*`) are published by the stage 3 companion PR (#7934); until that lands, deployments must point `runtimeImage` at their own images.

## Model Used

Claude Opus 4.8 (1M context), extended thinking, with tool use (Claude Code).

## Checklist

- [ ] If this change affects the UI, I have included before/after screenshots (no UI changes)
- [ ] All Paperclip CI gates are green (pending this push)


stubbi:
/greptile-review

Triggering a fresh review against the latest commit which adds the end-to-end smoke fixes (claude_local now runs successfully through this plugin against api.anthropic.com — verified manually with three consecutive successful runs).

stubbi:
@greptileai re-review please

stubbi:
@greptileai do a full review

stubbi:
@greptileai re-review please — addressed all four findings from your last pass in 21955a69:

1. `execInPod` now takes a `timeoutMs` watchdog; `effectiveTimeoutMs` from `plugin.ts` is threaded through to both call sites; the agent-exec path surfaces watchdog firings as `timedOut: true` so callers can retry instead of throwing.
2. `uploadInterceptor.reset()` is now called in `onEnvironmentReleaseLease`.
3. Sandbox-CR readiness is now cached in a worker-local `Set<leaseId>`, dropped on releaseLease. First exec on a lease confirms Ready; subsequent execs short-circuit.
4. Comment on `stdinPayload.length` clarifying Buffer-byte semantics for the `head -c` wrap.

All 107 unit tests still pass; typecheck clean.

stubbi:
@greptileai re-review please — addressed all three findings in 8d82241f:

1. `uploadInterceptor` is now per-lease via `uploadInterceptorsByLease: Map<leaseId, FastUploadInterceptor>`. `releaseLease` deletes only the entry for the lease being torn down, leaving other concurrent leases' buffers untouched. The previous singleton was a real concurrency bug I introduced — good catch.
2. `phase === "Terminating"` now throws a clear error immediately instead of polling until the timeout deadline.
3. Documented the deliberate all-ports semantics of operator-supplied CIDRs (operators may need non-HTTP services; HTTP/HTTPS-only LLM traffic is covered by the port-scoped public-IPv4 fallback).

stubbi:
Status update from the paperclip.inc side (we maintain this plugin downstream).

Since this PR was opened, our downstream fork has taken the kubernetes sandbox-provider plugin **all the way to a working end-to-end agent run** (a real OpenCode agent executing in a gVisor pod, routing inference through an in-cluster OpenAI/Anthropic-compatible gateway, landing a billed completion). In doing so the plugin evolved well beyond this PR's batch/v1-Jobs version. Notable deltas now on our fork that we'd want to fold in before this merges as "fully working":

- **`sandbox-cr` backend** (long-lived agent-sandbox CR the server execs into) alongside the `job` backend.
- **A real bug fix in `onEnvironmentExecute`**: it ran the harness via `execInPod` but never applied the caller's `params.env`. The k8s exec API has no env field, so the adapter's run env (most importantly `XDG_CONFIG_HOME` pointing at the shipped harness config) never reached the harness, which silently fell back to its in-image config. Fix: wrap the exec in a shell that exports `params.env` then `exec`s the command (PATH excluded). This was the single blocker to a working run.
- Per-run secret/env wiring, image allowlist, Cilium egress, tenant namespace/RBAC hardening, and assorted fixes from the e2e bring-up.

Two coordination questions before we invest in refreshing this PR:

1. **Roadmap fit.** "Cloud / Sandbox agents" is on `ROADMAP.md`; per `CONTRIBUTING.md` core feature work should be discussed in `#dev` first. Is upstream open to landing the kubernetes plugin (this PR, refreshed), and in what shape — one large PR, or staged (plugin core → backends → hardening)?
2. **Adapter-per-run.** Today the lease/run resolves the harness from the *environment's* pinned `adapterType`, not the *agent's*, so a company is effectively single-harness. To support mixed harnesses (e.g. an `opencode` and a `pi` agent in one company) we'd thread the agent's `adapterType` through `acquireRunLease`. Worth doing here, or as a follow-up?

We're happy to do the refresh + rebase onto current `master` and bring the above across once we know the preferred shape. (Two smaller generic fixes extracted from this work are already up standalone: #7836 and #7837.)

stubbi:
Progress update (same-day follow-up to the status above). Both open items have moved:

**2. Adapter-per-run: built and production-proven.** We threaded the agent's `adapterType` through the whole lease chain (orchestrator -> `acquireRunLease` -> driver -> plugin worker -> `onEnvironmentAcquireLease`), with the plugin resolving the runtime image, per-run env, and labels from the run's adapter and falling back to the environment default. One environment now serves mixed harnesses. Verified end-to-end on our production cluster: an OpenCode agent and a Pi agent in separate runs through the same execution path, each materialized in its harness-specific runtime image under gVisor, each landing a billed completion through an in-cluster gateway. One implementation note worth carrying into the refresh: `environment-runtime.ts` has two call sites for `environmentAcquireLease`, and only one is on the heartbeat path. We shipped the field on the wrong one first and the failure is silent (the run falls back to the environment default image and the harness binary is missing at exec time), so our fix pins the heartbeat-path payload in a test.

**Provider feature parity: complete.** We audited the plugin against the existing sandbox providers (daytona as the most complete reference) and closed the two remaining interface gaps, `resumeLease` and `destroyLease`. Resume semantics differ from Daytona by design: k8s pods are not restartable in place, so a gone or terminally failed pod reports the lease expired and the server falls back to a fresh acquire. Destroy is an idempotent forced teardown of everything acquire created (CR/Job, pod, per-run Secret; 404s are success). With those in, the plugin implements every `SandboxProvider` hook the daytona provider does, plus k8s-specific capabilities none of the SaaS providers have (runtimeClass isolation, network egress policy, per-tenant namespace/RBAC/quota, image allowlist, dual backend).

**Security posture** (since untrusted-code execution is the point of this plugin): agent pods run gVisor (verified at the syscall level, not just the label), restricted-PSS compliant, `automountServiceAccountToken: false`, in per-tenant namespaces with default-deny ingress and enumerated egress. We re-verified all of this on the live cluster this week.

Question 1 (preferred contribution shape: one refreshed PR vs staged plugin-core -> backends -> hardening) is still the one thing we need from upstream before doing the refresh. The code is ready to bring across.

stubbi:
Verification milestone: **all five major harnesses now run end-to-end through the kubernetes sandbox plugin on our production cluster** -- opencode, pi, codex, gemini, and claude_local. Each cell is a full product loop (company -> agent -> assigned issue -> heartbeat -> per-run harness-specific runtime image in a gVisor pod -> gateway-routed inference -> completed run with a billed usage row and the agent's reply comment), with mixed harnesses sharing one environment via the per-run adapter resolution described above.

Hardening fixes that came out of the bring-up, all generic and ready to fold into the refresh: codex's current CLI only speaks the Responses API (wire_api "chat" was removed Feb 2026), gemini-cli requires an auth-type selection persisted in its settings.json for headless runs (env vars alone do not satisfy it) which the adapter now writes into the managed HOME, and the adapter run-env vs host-env distinction matters for key plumbing in sandboxed paths.

That makes the plugin: feature-parity with the daytona provider on every SandboxProvider hook, verified against five harnesses, with k8s-native isolation (runtimeClass, network policy, per-tenant namespaces) on top. Still happy to refresh this PR in whatever shape upstream prefers (one PR vs staged).

stubbi:
@greptileai review please

stubbi:
Staged contribution complete. This PR is stage 1 of 3 (the plugin package, workspace-excluded, zero behavior change). Companions: **stage 2 #7938** (server + SDK integration: per-run adapter threading, env-gated execution policy, plugin auto-install) and **stage 3 #7934** (harness runtime images + publish workflow). Merge order: any, each is independently inert; the feature activates when all three are in and a deployment opts in via env.

stubbi:
@greptileai review

stubbi:
Pushed f2281f1 addressing review findings:

- Registered `@paperclipai/plugin-kubernetes` in `scripts/release-package-manifest.json` with `publishFromCi: false` (stage 1 does not auto-publish), which fixes the failing policy check.
- Removed internal milestone names from code comments and reworded the bootstrap-token TODO so it stands alone: it now points at the agent runtime shim's callback auth scheme landing in the companion images PR.
- Added a Risks note to the PR body: the default `ghcr.io/paperclipai/agent-runtime-*` images are published by the stage 3 companion PR (#7934); until then `runtimeImage` must point at your own images.

Verified standalone per the README: 147/147 unit tests, typecheck clean, build green.

stubbi:
@greptileai review

stubbi:
Addressed all three findings from the latest review in dac3ef036:

1. The namespace-collision finding was a real tenant-isolation bug, thank you. deriveCompanySlug no longer blind-truncates: full UUIDs pass through untruncated (a 36-char UUID fits the 63-char namespace budget with the default prefix), and longer inputs get an 8-hex sha256 of the FULL input appended so every byte contributes to the name. Collision test added (two UUIDs differing only in the tail now produce distinct slugs).
2. Run-name ULIDs now use crypto randomInt instead of Math.random; these become Job/Sandbox CR names and providerLeaseIds and must not be enumerable.
3. The pod-exec stdin workaround comment now references the actually-pinned @kubernetes/client-node v1.x.

149/149 unit tests, typecheck and build clean. The same slug fix is being applied to our production deployment downstream.

stubbi:
@greptileai review

stubbi:
Both findings addressed in f447767a7:

1. The pod fallback now selects exclusively by the unique agents.x-k8s.io/sandbox-name label (no managed-by listing, no name-prefix narrowing), so a concurrent sandbox sharing a generated-name prefix can never be exec'd into. Tests pin the exact label selector and the cross-sandbox negative case.
2. The execute timeout is now one budget shared across the readiness wait and the exec (wall-clock tracked from handler start, 5s exec floor so a slow readiness phase still surfaces a bounded exec error instead of a silent doubling).

150/150 tests, typecheck and build clean.

@greptileai review

stubbi:
Both findings addressed in c4c564bba:

1. The fast-upload flush exec now shares the caller's single execute budget (remaining wall-clock after the readiness wait, 5s floor) and wraps execInPod in the same try/catch contract as the normal exec path, returning a timed-out result instead of throwing uncaught.
2. Every tenant ensure* create tolerates 409 AlreadyExists: concurrent first-lease acquisitions for a new tenant race the read-then-create, and the loser's conflict means the desired state already exists. Race test added (concurrent 409s on namespace + ServiceAccount resolve cleanly).

151/151 tests, typecheck clean.

@greptileai review

PR #4554, opened 2026-04-26
fix(auth): trust allowed hostname port variants on detected listen port

devinfoley:
## Thinking Path

> - Paperclip is the control plane for autonomous AI companies, so authenticated board access has to be predictable across local and worktree deployments.
> - This change sits in the authenticated-mode server startup and Better Auth origin-trust wiring.
> - The original auth branch fixed one real gap by adding port-qualified trusted origins for allowed hostnames on non-default ports.
> - Review of that branch found a second-order bug: trusted origins were still derived from the configured port before startup detected the actual listen port.
> - In isolated worktrees, that meant a common `3100 -> 3101` port shift could still leave Better Auth trusting the stale origin.
> - This pull request keeps the original allowed-hostname port-variant fix, then moves trust derivation onto the resolved listen port and adds regression coverage around startup wiring.
> - The benefit is that authenticated sessions keep working on allowed private hostnames even when Paperclip has to auto-shift to a different local port.

## What Changed

- Added `:port` trusted-origin variants for authenticated-mode `allowedHostnames` when Paperclip runs on non-default ports.
- Changed authenticated startup so `listenPort` is detected before Better Auth initialization, and explicit auth base URLs are rewritten before auth startup.
- Updated `deriveAuthTrustedOrigins()` to accept the resolved listen port so Better Auth trusts the actual browser origin instead of the stale configured port.
- Added focused regression coverage in `server/src/__tests__/better-auth.test.ts` and `server/src/__tests__/server-startup-feedback-export.test.ts`.

## Verification

- `pnpm exec vitest run server/src/__tests__/better-auth.test.ts server/src/__tests__/server-startup-feedback-export.test.ts`
- Reviewer re-check: reviewed commits `380f5b9f` and `092bb34c` after the follow-up fix landed and found no remaining issues.

## Risks

- Low risk: this only affects authenticated-mode origin derivation and startup ordering around detected listen ports.
- Main behavioral shift: startup no longer mutates `config.port` to the selected port; it now carries `requestedListenPort` separately and uses `listenPort` where runtime behavior needs the resolved value.
- If another path was implicitly relying on `config.port` being overwritten during startup, that path would need follow-up, though the current startup/test coverage did not reveal one.

> I checked `ROADMAP.md` and did not find an overlapping planned core work item for this auth trusted-origin port handling fix.

## Model Used

[...]

PR #5817, opened 2026-05-12
fix(auth): trust both runtime listen port and configured port

echokos:
## Summary

`deriveAuthTrustedOrigins` was emitting hostname:port variants for only one port — either `config.port` (historically) or, since #4554, the runtime `listenPort` from `detectPort`. When the two differ (because `detectPort` slid to a free port at startup, while the user is still reaching Paperclip on the configured port via a reverse proxy, Tailscale funnel, container mapping, or Cloudflare tunnel), Better Auth 403s every login as `Invalid origin`. This PR unions both ports into the trusted-origins list.

## Thinking Path

> - Paperclip orchestrates AI agents for zero-human companies; the server ships with a Better Auth integration that gates the web UI behind email+password sessions.
> - Better Auth enforces a `trustedOrigins` allowlist on every authenticated request — a `403 Invalid origin` rejects anything whose `Origin` header isn't on the list.
> - `deriveAuthTrustedOrigins` in `server/src/auth/better-auth.ts` builds that allowlist from `allowedHostnames` plus a single port — historically the user-configured `config.port`, then since #4554 the runtime `listenPort` reported back from `detectPort`.
> - `detectPort` silently falls back to a free port when the configured one is busy at startup (e.g. a stale process holding `:3101`, so the new one binds `:3102`). That makes `listenPort !== config.port`.
> - But users may still be reaching Paperclip on the **configured** port via a reverse proxy, Tailscale funnel, container port mapping, or a Cloudflare tunnel — all of which target the configured port and don't care what the local listener landed on.
> - When that happens today, every login attempt 403s at Better Auth before reaching the auth handler, and the only recovery is editing the systemd / launchd unit or `.env` and restarting.
> - This pull request unions both ports into the trusted-origins list so neither path is locked out.
> - The benefit is that detectPort-fallback installs (and proxy/tunnel-fronted installs more generally) can log in without manual config surgery.

## What Changed

- `server/src/auth/better-auth.ts`: replaced the single-port branch with a `Set<number>` containing both `opts.listenPort ?? config.port` and `config.port`, then loop over them to emit hostname:port variants. `Set` dedup keeps output identical when the two are equal.
- `server/src/__tests__/better-auth.test.ts`: replaced the "prefers an explicit resolved listen port" test (which asserted the configured port should **not** appear — encoding the bug) with one that asserts both ports' variants appear when `listenPort !== config.port`. The "adds hostname port variants for authenticated mode on non-default ports" test still covers the equal-ports common case.

## Verification / Test plan

- **Automated:** `pnpm --filter @paperclipai/server test better-auth` exercises both the updated assertion (`listenPort=3102` + `config.port=3101` → both port variants in `trustedOrigins`) and the unchanged equal-ports case via the existing "adds hostname port variants for authenticated mode on non-default ports" test.
- **Manual:** start Paperclip with `port=3101` while `:3101` is already busy so `detectPort` slides to `:3102`. Confirm UI loaded on `:3101` (via reverse proxy / Tailscale funnel) and on `:3102` (direct) both log in without `403 Invalid origin`.
- **Negative:** confirm standard ports `80` and `443` are still skipped via the existing `continue` guard (no `:80` / `:443` suffixes leak into the trusted list).

## Risks

Low. The change is strictly additive — any origin previously trusted is still trusted, and the only new origins are hostname:port variants for `config.port` when it differs from `listenPort`. Standard ports (80, 443) are still skipped via the existing guard. No migration, no config surface change.

## Model Used

- Claude Opus 4.7 (`claude-opus-4-7`), Anthropic, extended thinking mode, working through Claude Code CLI.

## Checklist

- [ ] I have run tests locally and they pass (no `node_modules` in my disposable PR-prep checkout; CI vitest will exercise)
- [ ] I have updated relevant documentation to reflect my changes (no doc surface affected)

### DIFF
--- a/server/src/__tests__/better-auth.test.ts
+++ b/server/src/__tests__/better-auth.test.ts
@@ -59,20 +59,24 @@ describe("Better Auth cookie scoping", () => {
     ]));
   });
 
-  it("prefers an explicit resolved listen port over the configured port", () => {
+  it("includes both the resolved listen port AND the configured port when they differ", () => {
+    // detectPort may fall back to a free port if the configured one is busy,
+    // but users can still reach the install on the configured port via a
+    // reverse proxy, Tailscale funnel, container port mapping, or Cloudflare
+    // tunnel. Both ports must be trusted so neither path 403s on "Invalid origin".
     const trustedOrigins = deriveAuthTrustedOrigins({
       deploymentMode: "authenticated",
       authBaseUrlMode: "auto",
       authPublicBaseUrl: undefined,
       allowedHostnames: ["board.example.test"],
-      port: 3100,
-    } as Parameters<typeof deriveAuthTrustedOrigins>[0], { listenPort: 3101 });
+      port: 3101,
+    } as Parameters<typeof deriveAuthTrustedOrigins>[0], { listenPort: 3102 });
 
     expect(trustedOrigins).toEqual(expect.arrayContaining([
       "https://board.example.test:3101",
       "http://board.example.test:3101",
+      "https://board.example.test:3102",
+      "http://board.example.test:3102",
     ]));
-    expect(trustedOrigins).not.toContain("https://board.example.test:3100");
-    expect(trustedOrigins).not.toContain("http://board.example.test:3100");
   });
 });
--- a/server/src/auth/better-auth.ts
+++ b/server/src/auth/better-auth.ts
@@ -73,14 +73,24 @@ export function deriveAuthTrustedOrigins(config: Config, opts?: { listenPort?: n
     }
   }
   if (config.deploymentMode === "authenticated") {
-    const port = opts?.listenPort ?? config.port;
-    const needsPortVariants = port !== 80 && port !== 443;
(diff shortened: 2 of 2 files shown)


echokos:
@brandonburr thanks for the triage! The PR description has been refreshed to include all six template sections from `.github/PULL_REQUEST_TEMPLATE.md`:

- **Summary** (new TL;DR block at top)
- **Thinking Path** (eight-step trace from project context down to the fix)
- **What Changed** (concrete bullet list per file)
- **Verification / Test plan** (automated + manual + negative cases)
- **Risks** (low — strictly additive)
- **Model Used** (Claude Opus 4.7 via Claude Code CLI)
- **Checklist** (filled, with N/A markers explained inline)

The Greptile review on the original sparser body was from commit `41f4d2d` before the body update — re-triggering Greptile should pick up the current description and clear its outstanding template comment, since the code itself was already scored 4/5 with no inline issues.

Happy to make further changes if anything is still missing.

Issue #4759, opened 2026-04-29
Security: HTTP logger leaks plaintext passwords (and other secrets) in reqBody on 4xx responses


00PZ:
## Summary

The HTTP logger middleware in `server/src/middleware/logger.ts` writes the entire request body to logs (stdout + `server.log`) on any 4xx/5xx response. The redact list only contains `req.headers.authorization`, so request *bodies* are never sanitized.

This means **plaintext passwords and other secrets are written to logs** whenever a request fails - most notably on every failed login.

## Reproduction

1. Run any Paperclip server build (reproduced on `v2026.428.0`).
2. Submit a wrong password to `/api/auth/sign-in/email`.
3. Tail server logs.

Observed (real example, password masked here for the issue):

```
WARN: POST /api/auth/sign-in/email 401 {"reqBody":{"email":"user@example.com","password":"<PLAINTEXT_PASSWORD>"},"reqParams":{"authPath":["sign-in","email"]},"routePath":"/api/auth/{*authPath}"}
```

The plaintext password is captured both at INFO/WARN console output and written to disk in `server.log`.

## Affected paths

The leak is **not limited to login**. Any handler that returns 4xx/5xx with a sensitive body will be logged. Examples likely affected:

- `/api/auth/sign-in/email` (password)
- `/api/auth/sign-up/email` (password)
- `/api/auth/reset-password` (token + new password)
- Any plugin / settings endpoint that accepts API keys / tokens in the body and rejects with 4xx (validation errors etc.)

## Root cause

`server/src/middleware/logger.ts`:

```ts
export const logger = pino({
  level: \"debug\",
  redact: [\"req.headers.authorization\"],   // only header is redacted
}, ...);

export const httpLogger = pinoHttp({
  logger,
  customProps(req, res) {
    if (res.statusCode >= 400) {
      // ...
      props.reqBody = body;   // <-- entire body, no redaction
      // ...
    }
  },
});
```

There is no field-level redaction for body fields and no allowlist - the entire `req.body` is dumped.

## Suggested fix

Two layers, both cheap:

1. **Redact known sensitive keys at the pino layer** so they are scrubbed regardless of which middleware adds them:

   ```ts
   redact: {
     paths: [
       \"req.headers.authorization\",
       \"req.headers.cookie\",
       \"req.headers[\\\"x-api-key\\\"]\",
       \"*.password\",
       \"*.newPassword\",
       \"*.currentPassword\",
       \"*.token\",
[...]

PR #5820, opened 2026-05-12
fix(logger): redact passwords and tokens from HTTP error log lines

### LINKED ISSUE #3072: Security: plaintext passwords logged on /api/auth/sign-in/email failures (request body not redacted)
## Description

The Paperclip server logs plaintext passwords from the `/api/auth/sign-in/email` request body to `server.log` (and via stdout transport to `journalctl`) on every failed login attempt.

This is a sibling of #2385 (Bearer tokens in headers) but a distinct code path: #2385 added `redact: ["req.headers.authorization"]` to `server/src/middleware/logger.ts`, which protects the request headers but not the request body. The `pino-http` `customProps` callback in the same file attaches `req.body` (full plaintext including the `password` field) to the log record on any 4xx response, which is exactly what better-auth produces on a failed sign-in.

In a local audit of a single day's `server.log` on v2026.86.0, I recovered **9 distinct plaintext password values** from failed login attempts, including a couple of long generator-style passwords that look very much like password manager output.

## Impact

Severity is the same as #2385 — anyone with read access to `server.log` (mode 644 by default) can harvest user credentials. In `authenticated`/`private` deployments the file is on the same host as the auth database; in shared hosting or container deployments it may be readable by unrelated processes.

This is arguably worse than the Bearer token leak because:
- Bearer tokens are short-lived and bound to a session, while raw passwords are long-lived and frequently reused across services
- Failed login attempts (which is what 401s are) often capture the user trying their *other* passwords from a password manager — so a single leaked log can expose multiple credentials per user
- Users typing into a browser don't expect their failed-login attempts to be recorded, so the data hygiene assumption is implicit

## Steps to Reproduce

1. Run `npx paperclipai onboard --yes && npx paperclipai run`
2. From the Paperclip web UI or via curl, attempt a sign-in with a wrong password:
   ```bash
   curl -X POST http://localhost:3100/api/auth/sign-in/email \
     -d '{"email":"test@example.com","password":"hunter2"}'
   ```
3. `tail -1 ~/.paperclip/instances/default/logs/server.log`
4. Observe the password value in the `reqBody` field of the log line:
   ```
   [HH:MM:SS] WARN: POST /api/auth/sign-in/email 401 {"req":{...},"reqBody":{"email":"test@example.com","password":"hunter2"},...}
   ```

The exact same leak path also fires on:
- `/api/auth/sign-up/email` (`password`)
- `/api/auth/change-password` (`currentPassword`, `newPassword`)
- Any other better-auth endpoint that takes a credential field in the body when the response is 4xx

## Root Cause

`server/src/middleware/logger.ts` configures pino with:

```ts
export const logger = pino({
  level: "debug",
  redact: ["req.headers.authorization"],   // ← only the header
}, ...)
```

`req.headers.authorization` is the only protected path. The `customProps` callback further down attaches `reqBody` (= `req.body`) to error logs verbatim. There is no `serializers` config and no `redact.paths` entry for `reqBody.password` (or for any other credential-bearing body field).

## Proposed Fix

A two-layer defense, modeled on the existing `SECRET_PAYLOAD_KEY_RE` regex from `server/src/redaction.ts` (which Paperclip already uses to scrub event payloads):

### Layer 1 — pino `serializers` (programmatic, regex-driven)

Reuse `redactEventPayload` from `redaction.ts` as a serializer for the top-level keys that the `customProps` callback writes to log records:

```ts
import { redactEventPayload } from "../redaction.js";

function safeRedact(value: unknown): unknown {
  if (value == null || typeof value !== "object") return value;
  if (Array.isArray(value)) return value.map(safeRedact);
  return redactEventPayload(value as Record<string, unknown>) ?? value;
}

function reqSerializer(req: unknown) {
  const serialized = pino.stdSerializers.req(req as never);
  if (serialized && typeof serialized === "object") {
    const headers = (serialized as { headers?: unknown }).headers;
    if (headers && typeof headers === "object" && !Array.isArray(headers)) {
      (serialized as { headers?: unknown }).headers =
        redactEventPayload(headers as Record<string, unknown>) ?? headers;
    }
  }
  return serialized;
}

export const logger = pino({
  level: "debug",
  serializers: {
    req: reqSerializer,
    reqBody: safeRedact,
    reqQuery: safeRedact,
    errorContext: safeRedact,
    // Note: reqParams intentionally excluded — Express route params like
    // `authPath` substring-match SECRET_PAYLOAD_KEY_RE and would be over-redacted.
  },
  redact: { /* see Layer 2 */ },
}, ...);
```

This catches any field at any depth whose key matches the credential regex — automatically covering `password`, `newPassword`, `currentPassword`, `token`, `accessToken`, `refreshToken`, `apiKey`, `secret`, `credential`, etc., and any future field whose name matches.

### Layer 2 — `redact.paths` (declarative, belt-and-suspenders)

Explicit paths for the well-known fields, in case a future refactor changes the top-level structure or bypasses the serializer dispatch:

```ts
redact: {
  paths: [
    'reqBody.password',
    'reqBody.newPassword',
    'reqBody.currentPassword',
    'reqBody.token',
    'reqBody.refreshToken',
    'reqBody.accessToken',
    'reqBody.secret',
    'reqBody.apiKey',
    'errorContext.reqBody.password',
    'errorContext.reqBody.newPassword',
    'errorContext.reqBody.currentPassword',
    'req.headers.authorization',
    'req.headers.cookie',
    'req.headers["x-api-key"]',
    'req.headers["x-openclaw-token"]',
    'req.headers["x-openclaw-auth"]',
    'req.headers["x-csrf-token"]',
  ],
  censor: '[REDACTED]',
}
```

Note that `req.headers.authorization` is preserved from the existing config; the others are additive.

## Verification

After applying both layers, re-run the curl repro and confirm the log line shows `"password":"[REDACTED]"` and `"authorization":"[REDACTED]"`:

```
[HH:MM:SS] WARN: POST /api/auth/sign-in/email 401 {...,"headers":{...,"authorization":"[REDACTED]","cookie":"[REDACTED]"},...,"reqBody":{"email":"test@example.com","password":"[REDACTED]"},...}
```

## Environment

- Paperclip server commit `0ac01a04` (2026-03-27)
- Node v22 LTS
- Linux LXC container (Debian 12)

## Offer

Happy to send a PR with this fix if maintainers agree on the approach. The change is ~30 lines in a single file (`server/src/middleware/logger.ts`) and reuses the existing `redactEventPayload` export, so there's no new surface area to maintain.

I have a working version of this patch deployed and verified locally. Let me know if a PR would be welcome.

echokos:
## Thinking Path

> - Paperclip orchestrates AI agents for zero-human companies; the server is an Express app with structured HTTP logging via pino-http.
> - The middleware in `server/src/middleware/logger.ts` defines a `customProps` hook that attaches request context (`req.body` / `req.params` / `req.query`) to every 4xx/5xx log entry so operators can diagnose failed requests.
> - That hook copies the body verbatim. Better Auth's `POST /api/auth/sign-in/email` carries an `{ email, password }` body — on a wrong-password attempt the request lands in the 4xx branch and the plaintext password is written to `~/.paperclip/logs/server.log`.
> - Two existing issues raise this (#3072 plaintext-password leak, #4759 similar concerns) and neither has a fix.
> - Same exposure surface applies to sign-up, reset-password, API key creation, and any endpoint that accepts a credential in the body and can return 4xx.
> - This pull request introduces a small `redactSensitive` walker that returns a shallow copy of the input with values for known credential-shaped keys replaced with `[REDACTED]`, and applies it at every body/params/query log site in `customProps`.
> - The benefit is that operators can keep diagnostic logging on without their disk silently accumulating user passwords and bearer tokens.

## What Changed

- `server/src/middleware/redact-sensitive.ts` (new): depth-capped, case-insensitive walker. Sensitive keys covered: `password`, `currentPassword`, `newPassword`, `passwordConfirmation`, `passwordConfirm`, `confirmPassword` (+ snake_case variants), `secret`, `client_secret`, `access_token`, `refresh_token`, `id_token`, `auth_token`, `session_token`, `api_key`, `authorization`, `private_key`. Bare `token` deliberately not in the list — pagination cursors and CSRF tokens are not credentials (per Greptile review).
- `server/src/middleware/logger.ts`: wraps the six log sites in `customProps` (3 ctx-path + 3 fallback-path) with `redactSensitive`.
- `server/src/__tests__/redact-sensitive.test.ts` (new): covers plaintext password, case-insensitive matching, multiple credential keys, nested objects/arrays, bare `token` left untouched, primitives untouched, cycle safety.
- Depth-cap returns `undefined` (field absent from log line) rather than a sentinel string, per Greptile review.

## Verification

- `pnpm --filter @paperclipai/server test redact-sensitive` should run the new test file green.
- Manual: tail `~/.paperclip/logs/server.log`, hit `POST /api/auth/sign-in/email` with a deliberately wrong password, confirm the logged `reqBody.password` reads `[REDACTED]` (not the plaintext) and the surrounding fields still appear for diagnosis.

## Risks

Low. The walker only rewrites values at the log-emit boundary — the actual `req.body` object handed to downstream handlers is unchanged because `redactSensitive` returns a new object. Standard log fields (email, route path, status code) remain visible. The sensitive-key list is conservative enough that the only risk is over-redacting a non-credential field that happens to share a name with a known credential; the bare `token` carve-out in this revision addresses the most obvious such case.

## Model Used

- Claude Opus 4.7 (`claude-opus-4-7`), Anthropic, extended thinking mode, working through Claude Code CLI.

## Checklist

- [ ] I have run tests locally and they pass (no `node_modules` in my disposable PR-prep checkout; CI vitest will exercise)
- [ ] I have updated relevant documentation to reflect my changes (no doc surface affected)

Closes #3072. Refs #4759.

---

## Original Context

`customProps` in the HTTP logger copies `req.body` / `req.params` / `req.query` verbatim into 4xx/5xx log entries. Better Auth's wrong-password flow therefore writes:

```
{"reqBody":{"email":"…","password":"founding6gomez6croaking"},"msg":"POST /api/auth/sign-in/email 401"}
```

…to disk. This PR rewrites credential-shaped values to `[REDACTED]` at that boundary.

### DIFF
--- a/server/src/__tests__/redact-sensitive.test.ts
+++ b/server/src/__tests__/redact-sensitive.test.ts
@@ -1,0 +1,67 @@
+import { describe, expect, it } from "vitest";
+import { redactSensitive } from "../middleware/redact-sensitive.js";
+
+describe("redactSensitive", () => {
+  it("redacts a plaintext password field on a sign-in body", () => {
+    const body = { email: "user@example.com", password: "founding6gomez6croaking" };
+
+    const out = redactSensitive(body) as Record<string, unknown>;
+
+    expect(out.email).toBe("user@example.com");
+    expect(out.password).toBe("[REDACTED]");
+    expect((body as Record<string, unknown>).password).toBe("founding6gomez6croaking");
+  });
+
+  it("redacts password key regardless of casing", () => {
+    expect((redactSensitive({ Password: "x" }) as Record<string, unknown>).Password).toBe("[REDACTED]");
+    expect((redactSensitive({ PASSWORD: "x" }) as Record<string, unknown>).PASSWORD).toBe("[REDACTED]");
+  });
+
+  it("redacts known credential-shaped keys", () => {
+    const out = redactSensitive({
+      currentPassword: "a",
+      newPassword: "b",
+      access_token: "c",
+      refresh_token: "d",
+      api_key: "e",
+      authorization: "Bearer f",
+    }) as Record<string, string>;
+
+    for (const value of Object.values(out)) {
+      expect(value).toBe("[REDACTED]");
+    }
+  });
+
+  it("recurses into nested objects and arrays", () => {
+    const out = redactSensitive({
+      user: { email: "user@example.com", password: "secret-pass" },
+      tokens: [{ access_token: "t1" }, { access_token: "t2" }],
+    }) as Record<string, unknown>;
+
+    expect((out.user as Record<string, unknown>).email).toBe("user@example.com");
+    expect((out.user as Record<string, unknown>).password).toBe("[REDACTED]");
+    const tokens = out.tokens as Array<Record<string, unknown>>;
+    expect(tokens[0].access_token).toBe("[REDACTED]");
+    expect(tokens[1].access_token).toBe("[REDACTED]");
+  });
+
+  it("leaves primitives and non-sensitive keys untouched", () => {
+    const body = { email: "a@b.c", name: "Alice", count: 7, active: true, missing: null };
+
+    expect(redactSensitive(body)).toEqual(body);
+  });
+
+  it("returns primitives unchanged", () => {
+    expect(redactSensitive("hello")).toBe("hello");
+    expect(redactSensitive(42)).toBe(42);
+    expect(redactSensitive(null)).toBe(null);
+    expect(redactSensitive(undefined)).toBe(undefined);
+  });
+
+  it("caps recursion depth so cycles do not pin the logger", () => {
+    const cycle: Record<string, unknown> = { name: "root" };
+    cycle.self = cycle;
+
+    expect(() => redactSensitive(cycle)).not.toThrow();
+  });
+});
--- a/server/src/__tests__/redact-sensitive.test.ts
+++ b/server/src/__tests__/redact-sensitive.test.ts
@@ -32,6 +32,13 @@ describe("redactSensitive", () => {
     }
   });
 
+  it("does not redact a bare `token` field — pagination cursors and CSRF tokens are not credentials", () => {
+    const out = redactSensitive({ token: "next-page-cursor", limit: 20 }) as Record<string, unknown>;
+
+    expect(out.token).toBe("next-page-cursor");
+    expect(out.limit).toBe(20);
+  });
+
   it("recurses into nested objects and arrays", () => {
     const out = redactSensitive({
       user: { email: "user@example.com", password: "secret-pass" },
--- a/server/src/middleware/logger.ts
+++ b/server/src/middleware/logger.ts
@@ -5,6 +5,7 @@ import { pinoHttp } from "pino-http";
 import { readConfigFile } from "../config-file.js";
 import { resolveDefaultLogsDir, resolveHomeAwarePath } from "../home-paths.js";
 import { shouldSilenceHttpSuccessLog } from "./http-log-policy.js";
+import { redactSensitive } from "./redact-sensitive.js";
 
 function resolveServerLogDir(): string {
   const envOverride = process.env.PAPERCLIP_LOG_DIR?.trim();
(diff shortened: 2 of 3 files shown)


devinfoley:
Thanks @echokos for this contribution. We weren't able to push the rebase / cross-reference edits to the head fork, so we've resubmitted your commits (with your authorship preserved) as #8013. Closing this one in favor of that PR — apologies for the duplicate notification. The fix itself is fully credited to you.

PR #5823, opened 2026-05-12
fix(remote-adapter): unblock claude_local agent runs on remote sandbox providers

stubbi:
## Thinking Path

> - Paperclip is a control plane for autonomous AI companies; agent execution should stay pluggable, observable, and company-scoped.
> - Kubernetes execution belongs at the sandbox-provider/plugin boundary rather than inside paperclip-server core (see #5790).
> - While end-to-end testing the new Kubernetes plugin from #5790, four bugs surfaced that are not k8s-specific — they affect any **remote** sandbox provider (Daytona today, SSH-targets later) using `claude_local` or the shared `adapter-utils` workspace machinery.
> - Carving these out of #5790 keeps that PR focused on the plugin and lets the platform-side fixes land independently and faster.

No existing issue covers these directly — bugs described in-PR (bug shape; discovered during end-to-end testing of the Kubernetes sandbox plugin in #5790, but none are k8s-specific):

- `claude-local` persisted the pod-side cwd (e.g. `/tmp`) into `sessionParams`, so the next run for the same task tar'd the host's `/tmp` — observable as a 35 MB+ base64 upload growing without bound (embedded-postgres data dir and server log included).
- Workspace snapshotting crashed (`EOPNOTSUPP`) on UNIX sockets / FIFOs such as `/tmp/.s.PGSQL.<port>`, killing the whole workspace sync.
- Plugin workers had their env intentionally stripped, but sandbox-provider plugins legitimately need model-provider API keys — without them `claude --print -` ran with `apiKeySource: "none"` and exited 1 ("Not logged in").
- These affect any **remote** sandbox provider (Daytona today, SSH targets later) using `claude_local` or the shared `adapter-utils` workspace machinery.

## What Changed

Four independent server-side / shared-package fixes. Each one is small, has a clear root cause, and is independently useful.

### 1. `claude-local`: persist host cwd in sessionParams, not pod cwd

`packages/adapters/claude-local/src/server/execute.ts`

`resolvedSessionParams.cwd` was being set to `effectiveExecutionCwd`. For a **remote** target, that is the pod-side working directory (typically ` /tmp` for our k8s pods). On the next run for the same task, heartbeat's `resolveWorkspaceForRun` reads `previousSessionParams.cwd` back as the **host** workspace dir and proceeds to walk + tar that path.

The observable failure: a 35 MB+ base64 upload growing without bound because we were tar-ing the host's `/tmp` (including the embedded-postgres data directory and the server log). Persisting the host `cwd` instead fixes the loop.

### 2. heartbeat: refuse system-tmp roots as session-resume workspace cwd

`server/src/services/heartbeat.ts`

Defense in depth for (1) and for any future adapter that might leak a pod-side path through `sessionParams`. If `previousSessionParams.cwd` resolves to ` / `, ` /tmp `, ` /var `, ` /private/tmp `, etc., treat it as untrusted and fall back to the default agent workspace dir. Unblocks any agent that already has a poisoned session row in the DB without requiring manual cleanup.

### 3. `adapter-utils`: snapshot must skip non-file/dir/symlink entries

`packages/adapter-utils/src/workspace-restore-merge.ts`

`captureDirectorySnapshot` → `walkDirectory` → `hashFile` opens a `fs.createReadStream` on every non-directory, non-symlink entry. UNIX sockets and FIFOs throw \`EOPNOTSUPP\` (errno \`-102\` on darwin) and crash the whole workspace sync.

Most reproducible with the embedded-postgres dev setup, which creates ` /tmp/.s.PGSQL.<port> `. After (1), even if a workspace cwd briefly points at a system-tmp root we now degrade gracefully instead of crashing. Skip anything that isn't a regular file, directory, or symlink — hash digests only ever applied to regular files anyway.

### 4. plugin-loader: allow model-provider API keys through to plugin worker

`server/src/services/plugin-loader.ts`

\`PluginWorkerManager.spawnProcess\` intentionally strips \`process.env\` when forking a worker, to avoid leaking host secrets like \`DATABASE_URL\` into untrusted plugin code. The comment there is right — we don't want to broadcast the full host env.

But **sandbox-provider plugins legitimately need model-provider API keys** to materialize them into per-run pod Secrets. Without that path, \`claude --print -\` runs with \`apiKeySource: \"none\"\` and emits "Not logged in" with exit 1.

This change passes a small explicit allowlist of adapter env keys through to the worker env:

- \`ANTHROPIC_API_KEY\`
- \`OPENAI_API_KEY\`
- \`GOOGLE_API_KEY\`
- \`GEMINI_API_KEY\`
- \`OPENROUTER_API_KEY\`

Other host env vars remain stripped. The allowlist matches the union of \`envKeys\` across the adapter-defaults registry in \`@paperclipai/plugin-kubernetes\` (from #5790).

## Verification

- Manually verified end-to-end against the Kubernetes sandbox provider from #5790, running \`claude_local\` against \`api.anthropic.com\` from inside a hardened pod. Three consecutive runs succeeded with \`stopReason: completed, exitCode: 0\`.
- The same code paths are exercised by \`@paperclipai/plugin-daytona\`; existing daytona tests still pass.
- Run \`pnpm test\` from \`packages/adapter-utils\` and \`packages/adapters/claude-local\` locally — green.

## Risks

- The system-root guard list in (2) is a closed set; if a future deployment uses an unusual writable temp root we'd silently fall back to agent_home. Mitigation: log a warning is already in place via the existing "Saved session workspace not available" path.
- The plugin worker env allowlist in (4) is hard-coded for now. If new adapter types add new \`envKeys\`, the list will need updating. A future improvement is to derive it dynamically from the installed plugins' adapter-defaults registries.

## Related

- Depends-on relationship: #5790 (Kubernetes sandbox provider) needs these to fully work end-to-end. The four fixes here are independent enough to land on their own.

## Model Used

Claude Opus 4.7 (1M context).

## Checklist

- [x] I have manually verified the change end-to-end
- [ ] I have added or updated tests — these are integration-shape bugs that surface during remote sandbox execution; covered in the smoke loop on #5790, not yet unit-tested
- [x] If this change affects the UI, I have included before/after screenshots — N/A

### DIFF
--- a/packages/adapter-utils/src/workspace-restore-merge.ts
+++ b/packages/adapter-utils/src/workspace-restore-merge.ts
@@ -47,10 +47,12 @@ async function walkDirectory(
 
     const fullPath = path.join(root, nextRelative);
     const stats = await fs.lstat(fullPath);
+    // Skip UNIX sockets, FIFOs, char/block devices — anything that's not a
+    // regular file, directory, or symlink. Trying to hash these via
+    // createReadStream throws EOPNOTSUPP / EINVAL.
     if (!stats.isDirectory() && !stats.isSymbolicLink() && !stats.isFile()) {
       continue;
     }
-
     if (stats.isDirectory()) {
       out.set(nextRelative, { kind: "dir" });
       await walkDirectory(root, exclude, nextRelative, out);
@@ -91,8 +93,10 @@ async function readSnapshotEntry(root: string, relative: string): Promise<Snapsh
       target: await fs.readlink(fullPath),
     };
   }
+  // Same guard as walkDirectory: hashing a UNIX socket / FIFO / device
+  // node via createReadStream throws EOPNOTSUPP. Treat unhashable entries
+  // as if they don't exist for snapshot-comparison purposes.
   if (!stats.isFile()) return null;
-
   return {
     kind: "file",
     mode: stats.mode,
--- a/packages/adapter-utils/src/workspace-restore-merge.ts
+++ b/packages/adapter-utils/src/workspace-restore-merge.ts
@@ -93,6 +93,10 @@ async function readSnapshotEntry(root: string, relative: string): Promise<Snapsh
       target: await fs.readlink(fullPath),
     };
   }
+  // Same guard as walkDirectory: hashing a UNIX socket / FIFO / device
+  // node via createReadStream throws EOPNOTSUPP. Treat unhashable entries
+  // as if they don't exist for snapshot-comparison purposes.
+  if (!stats.isFile()) return null;
   return {
     kind: "file",
     mode: stats.mode,
--- a/packages/adapter-utils/src/workspace-restore-merge.ts
+++ b/packages/adapter-utils/src/workspace-restore-merge.ts
@@ -47,6 +47,12 @@ async function walkDirectory(
 
     const fullPath = path.join(root, nextRelative);
     const stats = await fs.lstat(fullPath);
+    // Skip UNIX sockets, FIFOs, char/block devices — anything that's not a
+    // regular file, directory, or symlink. Trying to hash these via
+    // createReadStream throws EOPNOTSUPP / EINVAL.
+    if (!stats.isDirectory() && !stats.isSymbolicLink() && !stats.isFile()) {
+      continue;
+    }
     if (stats.isDirectory()) {
       out.set(nextRelative, { kind: "dir" });
       await walkDirectory(root, exclude, nextRelative, out);
--- a/packages/adapters/claude-local/src/server/execute.ts
+++ b/packages/adapters/claude-local/src/server/execute.ts
@@ -891,6 +891,11 @@ export async function execute(ctx: AdapterExecutionContext): Promise<AdapterExec
     const resolvedSessionParams = resolvedSessionId
       ? ({
         sessionId: resolvedSessionId,
+        // Persist the HOST workspace cwd. Heartbeat reads
+        // previousSessionParams.cwd as the local workspace dir for the next
+        // run; using effectiveExecutionCwd here would leak the remote pod
+        // cwd (e.g. "/tmp") into host workspace resolution and cause the
+        // next run to walk/tar the entire host /tmp.
         cwd,
         promptBundleKey: promptBundle.bundleKey,
         ...(executionTargetIsRemote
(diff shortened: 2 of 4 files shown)


stubbi:
@greptileai re-review please — addressed both non-blocking suggestions in 670e1fc9: `readSnapshotEntry` now has the same socket guard as `walkDirectory`, and the heartbeat system-root rejection emits a distinct operator-actionable warning.

stubbi:
@greptileai re-review please — addressed both follow-up nits in 4f792461:

1. `ADAPTER_ENV_PASSTHROUGH` is now gated on `manifest.capabilities.includes("environment.drivers.register")` and moved to module scope. Non-sandbox plugins (notification / UI / wiki) keep their fully-stripped env.
2. `SESSION_CWD_SYSTEM_ROOTS` is now a module-scoped constant — no per-tick allocation.

stubbi:
Refreshed onto current master and slimmed down after the kubernetes execution integration (#7938) merged: the plugin-loader model-provider key passthrough landed there, and master already carries the heartbeat session-cwd system-root guard. What remains here is the two fixes not yet upstream: claude-local persisting the HOST workspace cwd into sessionParams (so a remote run never leaks the pod cwd and the next run never tars the host /tmp), and the adapter-utils restore-merge guard that skips UNIX sockets / FIFOs / device nodes when hashing (createReadStream throws EOPNOTSUPP on them). Both are production-proven on our deployment. Tests green in both packages.

stubbi:
Closing as superseded — all four fixes have landed on master independently:

1. **claude-local host-cwd persistence** and 2. **socket/FIFO skip in workspace snapshots** (plus the regression test `workspace-restore-merge.test.ts` and the heartbeat `isUnsafeSessionWorkspaceCwd` guard) were absorbed by d1a8c873b (#5922 "fix(remote-sandbox): harden host workspace resumes").
3. **Model-provider API key allowlist for plugin workers** landed via the kubernetes sandbox integration (#7938) — see the allowlist in `server/src/services/plugin-loader.ts`.

After rebasing onto current master, this PR's remaining diff is explanatory comments only — which is also why the commitperclip test gate flagged it (no testable change left).

Re the open Greptile r-latest finding (prefix-match `SESSION_CWD_SYSTEM_ROOTS`): deliberately not applying — legitimate ephemeral workspaces live directly under `/tmp/…` (`os.tmpdir()` mkdtemp dirs on Linux, used by both tests and local-first runs), so prefix-rejecting system roots would break valid resumes. Master's exact-match guard plus the root-cause fix (host cwd persisted) is the right depth.

PR #5824, opened 2026-05-12
feat: auto-resume agents on budget raise + retry runs after Claude Max quota reset

MrSucik:
## Thinking Path

> - Paperclip orchestrates AI agents for zero-human companies, with per-agent budgets and a heartbeat run loop
> - Two operational footguns surfaced during a 12h fleet retrospective on a local instance
> - **Footgun A**: raising an agent's monthly budget via `PATCH /api/agents/:id` leaves the agent paused if `pauseReason="budget"` had previously fired. `agents.budgetMonthlyCents` and `budgetPolicies.amount` are two write paths to the same operator intent, and only the policy path triggers `resumeScopeFromBudget`. Operators raising the budget through the agent route therefore have to also call `POST /api/agents/:id/resume` manually
> - **Footgun B**: when Claude Max returns `out of extra usage · resets <time> (<tz>)`, the run is finalized as failed (or cosmetic-success) and the next scheduled fire hammers the same exhausted quota again — three wasted run-cycles per quota window in the worst case
> - This pull request unifies the budget write path so `budgetMonthlyCents` mutations always run through the policy machinery, and adds a quota-aware retry branch in the heartbeat finalizer that parses the reset clock-time and schedules a single bounded retry for that moment
> - The benefit is a fleet that self-heals on budget raises and stops burning fires against exhausted Claude Max quota — measured against an actual fleet with 100+ paused-and-stranded agent-tick equivalents over the past 12 hours

## What Changed

- `server/src/routes/agents.ts`: `PATCH /api/agents/:id` now calls `budgets.upsertPolicy({ scopeType: "agent", scopeId, amount: agent.budgetMonthlyCents, windowKind: "calendar_month_utc" })` when `budgetMonthlyCents` is in the patch body and differs from `existing.budgetMonthlyCents`. Mirrors the existing agent-create flow at line 2232. Response body is refreshed via `svc.getById` so the caller sees the resumed status atomically.
- `server/src/routes/companies.ts`: same pattern for company budget on `PATCH /api/companies/:companyId`.
- `server/src/services/claude-quota-reset.ts`: new pure utility module exporting `parseClaudeQuotaResetTime(errorMessage, now?) → Date | null`. Parses 12-hour, 24-hour, AM/PM, optional minutes, and an IANA timezone in parentheses. Computes the next occurrence of that local clock-time in the named timezone (advancing one day if already past). DST-aware via `Intl.DateTimeFormat` offset calculation.
- `server/src/services/heartbeat.ts`: imports the parser and adds a new retry branch in the run-finalization block (between the existing `MAX_TURN_CONTINUATION` and `transient_failure` branches). When the failed-run error matches the quota pattern, schedules a single `scheduleBoundedRetryForRun` retry for the parsed reset moment with `retryReason="claude_quota_exhausted"`, logs a lifecycle event with the reset timestamp, and emits the `claude_quota_reset` wakeup reason. Two new exported constants name these consistently with existing retry-reason constants.
- `server/src/__tests__/claude-quota-reset-parser.test.ts`: 8 unit tests covering the canonical Europe/Prague reset, next-day rollover, 12am/12pm edge cases, AM/PM hours, minutes, alternate IANA timezones (America/Los_Angeles), and invalid timezone / out-of-range hour guards.
- `server/src/__tests__/agent-budget-raise-resume-routes.test.ts`: 3 route tests verifying `budgets.upsertPolicy` is invoked with the correct arguments when budget changes, not invoked on unchanged or unrelated patches.

## Verification

Local test run from the worktree root:

```
cd /Users/xxx/paperclip-agent-resume-and-quota
npx vitest run --no-coverage \
  server/src/__tests__/claude-quota-reset-parser.test.ts \
  server/src/__tests__/agent-budget-raise-resume-routes.test.ts
# → 2 files passed, 11 tests passed
```

Typecheck:

```
cd server && pnpm typecheck   # → no new errors (TS clean)
```

Manual exercise of Footgun A reproduction:

```
# 1. Create or find an agent with budgetMonthlyCents=10000 and force it into
#    paused/reason=budget via a tight budget policy.
# 2. PATCH /api/agents/<id> { "budgetMonthlyCents": 200000 }
# 3. GET /api/agents/<id> — expect status="idle", pauseReason=null.
#    (Prior behavior: status stayed "paused" with pauseReason="budget".)
```

Manual exercise of Footgun B is observable in the run-events stream after a Claude Max quota-exhaustion error: a new `claude_quota_exhausted` lifecycle event appears with the parsed `quotaResetAt` ISO timestamp, and the next attempt is delayed to that moment.

## Risks

- **Behavioral shift on PATCH /api/agents/:id and PATCH /api/companies/:companyId**: callers that previously raised `budgetMonthlyCents` via these routes and *expected* the agent/company to remain paused will now see it auto-resume when the new amount exceeds observed spend. This is the same behavior the existing `budgets.upsertPolicy` route already exhibits, so it's not new — just propagated to the alternate write path.
- **Quota retry interacts with scheduled retry attempt counter**: each quota-exhaustion failure increments `scheduledRetryAttempt`. Set `maxAttempts: 1` to bound the retry to a single attempt per failure event, preventing runaway scheduling.
- **DST and ambiguous local times**: the parser does a single offset lookup at the naive UTC of the target local time, which is accurate for the vast majority of cases but may be off by ±1h during a DST transition window. Acceptable because quota reset times are typically standard hours (10am) and the worst case is a one-hour-late retry, not a wrong-day retry.
- **No new errors in TypeScript or biome on the changed files vs master** (verified with `biome check` on the diff). Pre-existing biome warnings on `heartbeat.ts`, `agents.ts`, `companies.ts` are untouched (same count as master).
- **Note**: the worktree's isolated paperclip instance DB seed failed during `worktree:make` due to a missing migrations dir in the npx `paperclipai` package cache. The seed failure is non-blocking for code-only PR work (typecheck, biome, vitest all run against the worktree code without needing the running instance). Reviewers running the branch locally can either `pnpm paperclipai worktree reseed --seed-mode minimal` or test against the main instance.

## Model Used

- Provider: Anthropic
- Model: Claude Opus 4.7 (claude-opus-4-7)
- Context window: 1M
- Reasoning/thinking mode: extended thinking enabled
- Capability details: tool use (file edits, bash, structured planning via TaskCreate)

## Checklist

- [x] I have updated relevant documentation to reflect my changes (inline JSDoc on the parser)

### DIFF
--- a/server/src/__tests__/agent-budget-raise-resume-routes.test.ts
+++ b/server/src/__tests__/agent-budget-raise-resume-routes.test.ts
@@ -147,6 +147,12 @@ async function createApp() {
 				]),
 			})),
 		})),
+		// Route now wraps svc.update + budgets.upsertPolicy in a transaction
+		// to close the TOCTOU window. The mock passes a sentinel tx-like object
+		// through so the test can assert that callees receive it.
+		transaction: vi.fn(async (fn: (tx: unknown) => unknown) =>
+			fn({ __tx: true }),
+		),
 	};
 	app.use("/api", agentRoutes(db as any));
 	app.use(errorHandler);
@@ -190,6 +196,15 @@ describe("PATCH /api/agents/:id budget sync", () => {
 				windowKind: "calendar_month_utc",
 			},
 			"local-board",
+			{ __tx: true },
+		);
+		// svc.update must also receive the same tx so its agent-row write
+		// commits atomically with the policy upsert.
+		expect(mockAgentService.update).toHaveBeenCalledWith(
+			agentId,
+			expect.objectContaining({ budgetMonthlyCents: 200_000 }),
+			expect.any(Object),
+			{ __tx: true },
 		);
 	});
 
--- a/server/src/__tests__/agent-budget-raise-resume-routes.test.ts
+++ b/server/src/__tests__/agent-budget-raise-resume-routes.test.ts
@@ -1,0 +1,227 @@
+// Verifies that PATCH /api/agents/:id syncs the agent budget policy when
+// `budgetMonthlyCents` is updated, which is what triggers `resumeScopeFromBudget`
+// for agents previously paused under reason="budget". Prior to this wiring,
+// raising the budget left the agent paused indefinitely.
+
+import express from "express";
+import request from "supertest";
+import { beforeEach, describe, expect, it, vi } from "vitest";
+
+const agentId = "11111111-1111-4111-8111-111111111111";
+const companyId = "22222222-2222-4222-8222-222222222222";
+
+function baseAgent(overrides: Record<string, unknown> = {}) {
+	return {
+		id: agentId,
+		companyId,
+		name: "Builder",
+		urlKey: "builder",
+		role: "engineer",
+		title: "Builder",
+		icon: null,
+		status: "idle",
+		reportsTo: null,
+		capabilities: null,
+		adapterType: "claude_local",
+		adapterConfig: {},
+		runtimeConfig: {},
+		budgetMonthlyCents: 10_000,
+		spentMonthlyCents: 0,
+		pauseReason: null,
+		pausedAt: null,
+		permissions: { canCreateAgents: false },
+		lastHeartbeatAt: null,
+		metadata: null,
+		createdAt: new Date("2026-05-12T00:00:00.000Z"),
+		updatedAt: new Date("2026-05-12T00:00:00.000Z"),
+		...overrides,
+	};
+}
+
+const mockAgentService = vi.hoisted(() => ({
+	getById: vi.fn(),
+	update: vi.fn(),
+}));
+
+const mockAccessService = vi.hoisted(() => ({
+	canUser: vi.fn(async () => true),
+	hasPermission: vi.fn(async () => true),
+	ensureMembership: vi.fn(async () => undefined),
+}));
+
+const mockBudgetService = vi.hoisted(() => ({
+	upsertPolicy: vi.fn(async () => ({ id: "policy-1" })),
+}));
+
+const mockApprovalService = vi.hoisted(() => ({
+	create: vi.fn(),
+	getById: vi.fn(),
+}));
+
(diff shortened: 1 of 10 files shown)


MrSucik:
Addressed all 4 P0/P1 review findings in dd97cf6:

**(a) Hour validation for 12-hour clock** — `parseClaudeQuotaResetTime` now rejects out-of-range values ("13pm", "0am", "15pm", "0pm") instead of silently dropping the meridiem. 24h input (no am/pm) continues to accept 0..23. New test cases cover both branches.

**(b) Adapter scoping** — the quota parser only fires when `agent.adapterType === "claude_local"` in heartbeat.ts. Other adapters never emit the "resets at …" message but could theoretically match the regex on stray error text.

**(c) TOCTOU on budget sync** — both PATCH /api/agents/:id and PATCH /api/companies/:companyId now wrap `svc.update` + `budgets.upsertPolicy` in `db.transaction`. Threaded an optional `tx` through `agentService.update`, `companyService.update`, and `budgets.upsertPolicy`. Auxiliary side-effects inside upsertPolicy (incidents, scope pause/resume, activity log) intentionally remain on the parent `db` so they don't read uncommitted state.

**(d) Retry-of-retry semantics** — `scheduleBoundedRetryForRun` accepts an `attemptBaseline` option. The quota-retry call site passes `attemptBaseline: 0` so a quota retry on top of a prior transient retry (`run.scheduledRetryAttempt = 1`) no longer falls through `nextAttempt (2) > maxAttempts (1)` and gets silently dropped. The quota-retry budget is now independent from the transient-retry budget.

**Tests / checks**
- 10 quota-parser cases pass (4 new: 13pm/15pm/0am/0pm rejection + 16:00 24h positive)
- 3 agent-route tests pass (now assert tx threads through both svc.update and budgets.upsertPolicy)
- 3 new company-route tests pass (symmetric to agent tests)
- `tsc --noEmit` clean across the whole server project
- biome auto-formatted touched files; no new lint diagnostics from the diff

PR #5831, opened 2026-05-12
fix(adapter-utils): prefer local listen url for agent env

Alchemist-DevAI:
## Thinking Path

> - Paperclip is the control plane that coordinates agent heartbeats and gives local adapters the runtime context they need to reach the API.
> - `packages/adapter-utils` is the shared layer that constructs that runtime environment for adapters like `claude_local` and `codex_local`.
> - In local trusted installs, the server already knows its canonical bind surface through `PAPERCLIP_LISTEN_HOST` and `PAPERCLIP_LISTEN_PORT`.
> - The current helper still prefers inherited `PAPERCLIP_RUNTIME_API_URL` / `PAPERCLIP_API_URL`, so a stale parent value can leak an obsolete LAN IP into new agent sessions.
> - When that happens, agent retries fail even though the Paperclip server itself is healthy on loopback.
> - This pull request makes local adapter env generation trust the instance listen host/port first and locks both exported API env vars to that resolved local URL.
> - The benefit is that local heartbeats stop inheriting drifted API origins while keeping the fallback behavior unchanged when no listen vars are present.

## What Changed

- Updated `buildPaperclipEnv` in `packages/adapter-utils/src/server-utils.ts` to prefer `PAPERCLIP_LISTEN_HOST` + `PAPERCLIP_LISTEN_PORT` over inherited API URL env vars when local listen settings are present.
- Exported `PAPERCLIP_RUNTIME_API_URL` alongside `PAPERCLIP_API_URL` so child processes inherit the corrected local URL consistently.
- Added a regression test in `packages/adapter-utils/src/server-utils.test.ts` covering the stale-LAN-IP case with loopback listen vars present.

## Verification

- Local hotfix smoke against a live instance with stale parent env:
  - `buildPaperclipEnv(...)` resolved `http://127.0.0.1:3100`
  - `GET /api/agents/me` returned `200` using the resolved URL and agent bearer token
- Added a focused unit regression test for the same scenario in `packages/adapter-utils/src/server-utils.test.ts`
- Full repo test run was not executed in this checkout because the cloned source tree was not bootstrapped with workspace dependencies during this patch-only pass.

## Risks

- Low risk: behavior only changes when `PAPERCLIP_LISTEN_HOST` or `PAPERCLIP_LISTEN_PORT` is present, which is the local-instance case that should already be authoritative.
- Remote/public deployments without local listen vars still keep the existing fallback to `PAPERCLIP_RUNTIME_API_URL` / `PAPERCLIP_API_URL`.

> Checked `ROADMAP.md`; this is a bug fix in shared adapter env generation, not overlapping planned feature work.

## Model Used

- OpenAI Codex via ChatGPT account session
- Model family: GPT-5 Codex runtime in the Codex CLI environment provided by Paperclip
- Capabilities used: tool use, shell execution, code editing, local runtime inspection
- Reasoning mode: medium interactive reasoning

## Checklist

### DIFF
--- a/packages/adapter-utils/src/server-utils.test.ts
+++ b/packages/adapter-utils/src/server-utils.test.ts
@@ -66,42 +66,100 @@ describe("buildInvocationEnvForLogs", () => {
 });
 
 describe("buildPaperclipEnv", () => {
-  it("prefers the local listen host and port over inherited runtime api env", () => {
-    const originalRuntimeApiUrl = process.env.PAPERCLIP_RUNTIME_API_URL;
-    const originalApiUrl = process.env.PAPERCLIP_API_URL;
-    const originalListenHost = process.env.PAPERCLIP_LISTEN_HOST;
-    const originalListenPort = process.env.PAPERCLIP_LISTEN_PORT;
+  function withPaperclipApiEnv(
+    env: Partial<Record<"PAPERCLIP_RUNTIME_API_URL" | "PAPERCLIP_API_URL" | "PAPERCLIP_LISTEN_HOST" | "PAPERCLIP_LISTEN_PORT" | "HOST", string>>,
+    run: () => void,
+  ) {
+    const originalValues = {
+      PAPERCLIP_RUNTIME_API_URL: process.env.PAPERCLIP_RUNTIME_API_URL,
+      PAPERCLIP_API_URL: process.env.PAPERCLIP_API_URL,
+      PAPERCLIP_LISTEN_HOST: process.env.PAPERCLIP_LISTEN_HOST,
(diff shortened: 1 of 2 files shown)


Alchemist-DevAI:
Applied the two greptile P2 follow-ups in 1e0387118b660f8fa8343b8ba1c2b7dbd55dd34f.

What changed:
- tightened `buildPaperclipEnv` so the local override only activates when `PAPERCLIP_LISTEN_HOST` is explicitly set
- added regression coverage for the partial-listen-var case and the fallback path that preserves inherited runtime/API URLs when listen vars are absent

Verification:
- `corepack pnpm exec vitest run packages/adapter-utils/src/server-utils.test.ts`

PR #5794, opened 2026-05-12
fix(issues): serialize Date params to ISO string in enrichCommentsWithDerivedAgentAttribution

TeTedo:
## Thinking Path

1. Paperclip serves a Web UI at \`localhost:3100\` that lists issue comments with agent attribution.
2. The dashboard calls \`GET /api/issues/:id/comments\` to render comment threads.
3. That route resolves through \`services/issues.ts::listComments\`, which calls \`enrichCommentsWithDerivedAgentAttribution\` to map comments back to their originating heartbeat runs.
4. \`enrichCommentsWithDerivedAgentAttribution\` queries \`heartbeat_runs\` filtered by a time window using two \`coalesce(...) >= ...\` / \`<= ...\` predicates in raw \`sql\` template literals.
5. The bind values for those predicates are constructed as \`new Date(minCommentCreatedAtMs)\` and \`new Date(maxCommentCreatedAtMs + SLACK)\` — raw JS \`Date\` objects.
6. drizzle-orm's postgres-js driver does not auto-coerce \`Date\` objects passed via \`sql\`\`\` template literals at these bind sites; it forwards them to postgres.js, which expects \`string | Buffer | ArrayBuffer\`.
7. Result: every \`GET /api/issues/:id/comments\` request that exercises attribution enrichment returns 500 with \`TypeError [ERR_INVALID_ARG_TYPE]: The "string" argument must be of type string or an instance of Buffer or ArrayBuffer. Received an instance of Date\`.
8. This PR coerces those two Date params to ISO strings at the call site, matching the pattern already used for date binds elsewhere in this file.

## What Changed

- \`server/src/services/issues.ts\` — Added \`.toISOString()\` to the two \`new Date(...)\` bind values inside \`enrichCommentsWithDerivedAgentAttribution\`'s time-window predicates.

## Verification

1. Start a local Paperclip instance with at least one issue that has comments authored by a heartbeat-run agent.
2. **Before patch:** \`GET /api/issues/<id>/comments?order=desc&limit=50\` returns \`500\` with \`Received an instance of Date\` in server logs (stack pointing into \`enrichCommentsWithDerivedAgentAttribution\`).
3. **After patch:** same request returns \`200\` with the normal comments payload and correct derived agent attribution.
4. \`pnpm typecheck\` — clean.
5. \`pnpm --filter @paperclipai/server vitest run src/__tests__/issues-service.test.ts\` — 46/46 pass.

## Risks

[...]

PR #5835, opened 2026-05-12
fix(server): stringify Date bind values in enrichCommentsWithDerivedAgentAttribution sql templates

conradoboty:
## Summary

- `enrichCommentsWithDerivedAgentAttribution` in `server/src/services/issues.ts` builds two raw `sql` tagged-template fragments that compare `coalesce(finishedAt, createdAt)` and `coalesce(startedAt, createdAt)` against computed timestamp windows. The bind values were passed as JavaScript `Date` objects.
- With the `postgres-js` driver, raw `sql` template interpolation skips drizzle's typed coercion and reaches `postgres/src/bytes.js:22` which calls `Buffer.byteLength(str)` — throwing `TypeError [ERR_INVALID_ARG_TYPE]: The "string" argument must be of type string … Received an instance of Date`.
- drizzle's typed operators (`gte`/`lte`) coerce `Date → string` internally, but the `coalesce(...)` shape makes those operators awkward to apply here, so we use Option A: stringify both timestamps to ISO-8601 before interpolation.

## Reproduction (from COD-629)

Any `GET /api/issues/{id}/comments` request where the result set contains at least one human-authored comment (no `createdByRunId`, has `authorUserId`) triggers `enrichCommentsWithDerivedAgentAttribution` which executes the broken query and returns 500.

## Changes

- `server/src/services/issues.ts` (lines ~1972–1973): append `.toISOString()` to both `new Date(...)` bind values.
- `server/src/__tests__/issues-service.test.ts`: new `describeEmbeddedPostgres` block — _"issueService.listComments date interpolation regression (COD-632)"_ — seeds a human-authored comment and an overlapping heartbeat run, then asserts `svc.listComments` resolves without error.

## Test plan

- [x] `npx vitest run --reporter=verbose "server/src/__tests__/issues-service.test.ts" -t "COD-632"` passes locally (1 test, 96 ms)
- [x] No new typecheck errors introduced in changed files
- [ ] CI green

Closes COD-632

### DIFF
--- a/server/src/__tests__/issues-service.test.ts
+++ b/server/src/__tests__/issues-service.test.ts
@@ -2984,3 +2984,102 @@ describeEmbeddedPostgres("issueService.clearExecutionRunIfTerminal", () => {
     expect(row).toEqual({ executionRunId: null, executionLockedAt: null });
   });
 });
+
+describeEmbeddedPostgres("issueService.listComments date interpolation regression (COD-632)", () => {
+  let db!: ReturnType<typeof createDb>;
+  let svc!: ReturnType<typeof issueService>;
+  let tempDb: Awaited<ReturnType<typeof startEmbeddedPostgresTestDatabase>> | null = null;
+
+  beforeAll(async () => {
+    tempDb = await startEmbeddedPostgresTestDatabase("paperclip-listcomments-date-");
+    db = createDb(tempDb.connectionString);
+    svc = issueService(db);
+    await ensureIssueRelationsTable(db);
+  }, 20_000);
+
+  afterEach(async () => {
+    await db.delete(issueComments);
+    await db.delete(activityLog);
+    await db.delete(heartbeatRuns);
+    await db.delete(issues);
+    await db.delete(agents);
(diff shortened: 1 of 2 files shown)


timlevel:
Hi @conradoboty, I noticed that you've also submitted a PR to fix the Date serialization issue. I wanted to make you aware that I have a similar PR (#5794) that was opened earlier and has already been reviewed and approved. To avoid duplication of effort, perhaps we could coordinate on which approach to take forward. Both PRs solve the same problem with similar solutions. CC @cryppadotta @devinfoley @paperclipai

timlevel:
@cryppadotta I noticed you're active on the repository right now. There are two PRs (#5794 and this one) addressing the same Date serialization issue. Both have been reviewed and approved. Could you please take a look and decide which approach to move forward with? #5794 was opened first and has been waiting longer.

timlevel:
Friendly reminder @cryppadotta - this PR and #5794 both fix a critical bug causing 500 errors in production. Both have been reviewed and approved. Please take a look when you have a moment to decide which approach to take forward.

PR #5805, opened 2026-05-12
feat(oauth): add OAuth backbone, connections UI, and token bindings

cryppadotta:
## Thinking Path

> - Paperclip orchestrates AI agents for zero-human companies, including agents that need governed access to third-party SaaS systems.
> - OAuth is the control-plane capability that lets operators connect providers such as GitHub, Notion, Slack, Linear, Atlassian, Google Workspace, and Microsoft Graph without hand-copying personal access tokens into agent config.
> - Jannes Stubbemann's source PR #5647 added the user-facing OAuth connection UI and binding picker, originally stacked on the OAuth backbone work.
> - The review follow-up needs a single replacement PR that carries the committed branch directly on top of `paperclipai/paperclip:master`, preserves the original work, and avoids lockfile churn that repo automation owns.
> - This pull request restacks the OAuth backend, provider registry, plugin SDK additions, Settings -> Connections UI, and `oauth_token` env binding support into one reviewable branch.
> - The benefit is that reviewers can evaluate the full OAuth feature path in one PR: provider config loading, connection lifecycle routes, refresh/revoke behavior, UI management, and agent config bindings.

## What Changed

- Replaces and credits Jannes Stubbemann's source PR #5647: https://github.com/paperclipai/paperclip/pull/5647
- Adds OAuth connection persistence, authorization state storage, and contiguous migration `0085_oauth_connections.sql`.
- Adds OAuth provider config loading, plugin-contributed provider registration, PKCE/state helpers, token exchange, refresh, revoke, mark-revoked, and callback/connect/disconnect routes.
- Extends shared/plugin contracts with `oauth_token` env bindings and plugin SDK OAuth provider helpers.
- Adds built-in provider YAML configs for GitHub, Notion, Slack, Linear, Atlassian, Google Workspace, and Microsoft Graph.
- Adds Settings -> Connections UI, provider tiles, connection drawer, OAuth API client wrappers, env var OAuth binding picker, screenshots, and focused UI tests.
- Removes `pnpm-lock.yaml` from the final PR diff with commit `9fcb892d` so lockfile updates remain owned by repo automation.
- Removes the new external `yaml` package dependency with commit `2a84e16f`; provider files now load through a narrow local parser for the schema-validated provider YAML subset, keeping `pnpm install --frozen-lockfile` green without committing lockfile churn.
[...]

PR #5836, opened 2026-05-12
feat(credential-broker): keep OAuth bearers out of agent address space (M1+M2, partial M3)

stubbi:
## Thinking Path

> - Paperclip's #5805 introduces an `oauth_token` env binding that resolves to a plaintext bearer in the agent process environment. For deterministic adapters that's fine, but for LLM agents — which can be prompt-injected into reading their own env vars, writing them to files, or exfiltrating via tool calls — handing a long-lived OAuth bearer to the agent's address space is the well-documented credential-exfiltration anti-pattern.
> - The industry has converged on a transport-layer credential broker pattern (Infisical Agent Vault, exe.dev, AgentSecrets): the agent references credentials by opaque handle, a host-allowlisted forward proxy injects the real value at the TLS boundary, and the bearer never enters the agent's address space.
> - This PR adds that capability to Paperclip without changing the `oauth_token` binding contract from #5805 — only the resolution semantics change, gated per-agent-config and behind a feature flag so rollout is zero-break.
> - Local-first invariant: the broker runs in-process inside the Paperclip server in local-dev (no Docker, no extra daemon, no master password setup). The same code can run as a sibling service in k8s deployments via a standalone bootstrap mode (M3 follow-up). Externally-hired agents (OpenClaw, Hermes) stay on env delivery unless their operator opts in to a BYO broker.
> - The smart-resolver default means the moment a sandbox runtime is wired up and a provider is opted in, every dispatch automatically uses the broker — no agent config edits needed.

No existing issue covers this directly — problem described in-PR:

- #5805 introduced an `oauth_token` env binding that resolves to a plaintext bearer in the agent process environment — fine for deterministic adapters, but LLM agents can be prompt-injected into reading their own env vars, writing them to files, or exfiltrating them via tool calls. Handing a long-lived OAuth bearer to the agent's address space is the well-documented credential-exfiltration anti-pattern.
- The industry has converged on a transport-layer credential broker: the agent references credentials by opaque handle, and a host-allowlisted forward proxy injects the real value at the TLS boundary, so the bearer never enters the agent's address space.
- This PR adds that capability without changing the `oauth_token` binding contract — only resolution semantics change, gated per-agent-config and behind a feature flag (default off) so rollout is zero-break.

## What Changed

Builds on top of #5805. All changes gated behind `PAPERCLIP_FEATURE_CREDENTIAL_BROKER` (default off in this PR).

**M1 — contract & plumbing (behavior-neutral):**
- Optional `credentialDelivery` field on `AdapterConfig` (Zod): `env | paperclip-broker | byo-broker`. Unset = smart resolver picks per dispatch.
- `broker_targets` jsonb column on `oauth_connections` (migration `0086`) + `broker-targets` CRUD service.
- `registerCredentialBroker()` plugin SDK extension slot.
- Server-side broker registry (single-broker resolution, cached).
- `PAPERCLIP_FEATURE_CREDENTIAL_BROKER` + `PAPERCLIP_REQUIRE_BROKER` env flags.
- Pure `resolveCredentialDelivery()` smart resolver — 11 unit tests covering every branch.
- Structured `credential-broker-fallback-to-env` warn-log with reason-specific remediation hints.
- Wired into `resolveAdapterConfigForRuntime`: flag-off skips the resolver; flag-on with no broker degrades to env (logged); `PAPERCLIP_REQUIRE_BROKER=1` throws `CredentialBrokerRequiredError`. Explicit `credentialDelivery` always honored.
- `broker:` block on the OAuth provider YAML schema, declared on every shipped provider. Now includes `apiHosts` for providers whose main API host isn't an OAuth endpoint (Slack, Microsoft Graph) and `headerInjection` for providers using non-`Authorization: Bearer` injection.

**M2 — real broker:**
- `@paperclipai/credential-broker-builtin` package: per-session ephemeral CA (RSA-2048, ≤24h validity), per-host leaf signing cached on the session, TLS-MITM HTTP CONNECT proxy listener with host-allowlist enforcement at CONNECT and header injection at the inner HTTP request, in-memory session store with per-company bearer fan-out, full `CredentialBroker` interface implementation.
- Self-registration via `@paperclipai/credential-broker-builtin/register` side-effect entry. Server bootstrap imports `/register` at app load.
- `resolveAdapterConfigForRuntime` now mints a session via `broker.mintSession()` when the resolver picks `paperclip-broker`, pushes the resolved bearers into the broker's bearer cache, and replaces env values with deterministic placeholders. Returns `brokerSession: { proxyUrl, caCertPem, sessionToken }` to callers.
- End-to-end vitest test against a stub `https.createServer` upstream — asserts upstream receives `Bearer REAL-BEARER-XYZ`, agent never sees the bearer. 403 on disallowed host, 407 on missing auth.
- GitHub flipped to `broker.supported: true` + `paperclip-broker` in `deliveryModesSupported`.

**M3:**
- `brokerSession` reserved on `AdapterExecutionTargetProcessOptions` so sandbox runtimes can consume it uniformly.
- **Heartbeat-level injection (M3.2):** when `brokerSession` is present, `heartbeat.ts` materializes the CA to a temp file and merges `HTTPS_PROXY` / `HTTP_PROXY` / `NO_PROXY` / `SSL_CERT_FILE` / `NODE_EXTRA_CA_CERTS` / `REQUESTS_CA_BUNDLE` / `CURL_CA_BUNDLE` / `GIT_SSL_CAINFO` / `DENO_CERT` into the agent's resolved env. This is the central injection point — sandbox plugins inherit it for free instead of each implementing it.
- **BYO push-target REST API (M3.4):** `GET/POST/DELETE /api/companies/:co/oauth/connections/:id/broker-targets`. Admin-gated mutations. URL + auth-token validation; auth-token persisted via `secretService.upsertSecretByName`.
- **Refresh worker BYO fan-out (M3.5):** every successful `refreshFn` pushes the rotated bearer to the broker's in-process cache **and** POSTs to each configured BYO `broker_target` with a 10s `AbortController` timeout. Best-effort; failures log without propagating.
- **Greptile hardening (61f806a4):** string-replace `$`-sequence corruption fix (split/join), IPv6 CONNECT parsing (`parseConnectTarget` with bracket-aware parser, 7 unit tests), provider-specific `headerInjection` override, `broker.apiHosts` allowlist.
- Operator docs updated with shipped/deferred scope.

## What's deferred (follow-up PRs)

Default-on flag flip is deliberately not in this PR because it requires per-sandbox-plugin verification — `heartbeat.ts` now injects the proxy env centrally, but each sandbox driver (`cloudflare`, `daytona`, `e2b`, `exe-dev`, local-subprocess) needs an end-to-end smoke check that the agent process actually respects `HTTPS_PROXY` and trusts the mounted CA before we flip the global default.

- Per-sandbox-driver smoke tests confirming the heartbeat-injected proxy env reaches the agent process.
- Standalone broker bootstrap mode (loopback → reachable service URL for k8s).
- Settings → Connections "Broker Push Targets" UI section (REST API ships in this PR; admins can `curl` it today).
- EnvVarEditor resolved-mode chip.
- Multi-provider rollout: smoke-test Slack/Linear/Notion/Atlassian/Google Workspace/Microsoft Graph against the broker and flip them.
- Default-on flag flip — gated on the per-sandbox smoke tests above.
- External-adapter coordination (Hermes/OpenClaw upstream PRs documenting BYO setup).

See `docs/superpowers/specs/2026-05-12-credential-broker-design.md` for the full design and `docs/superpowers/plans/2026-05-12-credential-broker-m1-plan.md` for the M1 task decomposition.

## Verification

- `pnpm --filter @paperclipai/shared exec vitest run` → 17 files / 97 tests passed.
- `pnpm --filter @paperclipai/plugin-sdk exec vitest run` → 2 files / 6 tests passed.
- `pnpm --filter @paperclipai/credential-broker-builtin exec vitest run` → 4 files / 27 tests passed including end-to-end stub upstream and the new IPv6 / `$`-sequence regressions.
- `pnpm --filter @paperclipai/server exec vitest run src/services/__tests__/ src/oauth/ src/plugins/ src/config/` → server-side OAuth + broker tests pass. Covers every existing #5805 OAuth resolution test (unchanged behavior with the flag off) plus the new M1/M2/M3 cases and broker-target REST.
- Typecheck across affected packages — `shared`, `db`, `plugin-sdk`, `credential-broker-builtin`, `adapter-utils`, `server`, `ui` — all green.
- Migration `0086_broker_targets.sql` is idempotent (same `DO $$ IF NOT EXISTS` pattern as `0085`).
- No `pnpm-lock.yaml` committed in this PR per the project lockfile policy. `pnpm install --frozen-lockfile` will likely surface `OUTDATED_LOCKFILE` because of the new `node-forge` dep on `@paperclipai/credential-broker-builtin` and the new `@paperclipai/credential-broker-builtin: workspace:*` dep on `server`; the refresh-bot owns the post-merge lockfile bump.

## Risks

- **TLS-MITM correctness** is security-critical. The implementation uses `node-forge` (battle-tested for this use case) and a per-task ephemeral CA so blast radius is bounded. The end-to-end test verifies the bearer never reaches the agent side and is injected upstream verbatim. CA TTL is clamped to ≤24h; sessions can be revoked explicitly. Greptile's P1 (`$`-sequence corruption) and P2 (IPv6 CONNECT, hardcoded header, narrow allowlist) findings are fixed in 61f806a4 with regression tests.
- **Resolver pickiness**: with the flag on, any oauth-token dispatch whose provider isn't yet broker-compatible falls back to env. Only GitHub is opted in. The `credential-broker-fallback-to-env` warn-log carries a reason-specific hint so operators see which runs would benefit from broader provider rollout.
- **Default-on is intentionally deferred** — flipping it now would break agents whose sandbox drivers haven't been smoke-tested end-to-end with the injected `HTTPS_PROXY` + mounted CA. Heartbeat-level injection is in place; per-driver verification lands first.
- **First-call ctx capture** for the broker: `resolveCredentialBroker` caches on first call, so the dispatch path and refresh worker both use the same `buildCredentialBrokerCtx` helper to keep the cached ctx consistent regardless of which fires first.
- **BYO push-target fan-out** is best-effort: a failing operator broker doesn't block refresh, but it does log. Per-target failure isolation means one bad target can't block the others.

## Model Used

Implementation, design spec, and M1/M2/M3 plan authored with assistance from Claude Opus 4.7 (1M context) using terminal, git, GitHub CLI, vitest, pnpm, and the Paperclip server's existing tooling.

## Checklist

- [x] Model used specified with version and capability details
- [x] ROADMAP.md checked — credential-broker isn't on the roadmap; spec doc + plan doc accompany the PR
- [x] Tests run locally and they pass
- [x] Tests added for every new module (smart resolver, registry, env flags, warn-log, broker-targets service, CA, session store, header injection, proxy listener, IPv6 CONNECT parsing, end-to-end stub upstream, runtime env materialization, broker-target REST routes)
- [ ] UI changes — (n/a — no UI changes in this PR; deferred to follow-ups that ship the EnvVarEditor chip and Settings UI broker-targets section; the REST API admins need is wired and curl-usable today)
- [x] Documentation updated — operator intro + design spec + M1 plan
- [x] Risks documented above
- [x] Greptile / reviewer comments addressed (4 findings fixed in 61f806a4)


stubbi:
Addressed Greptile's review findings in 61f806a4 — the four P1/P2 issues (string-replace special chars, IPv6 CONNECT parsing, provider-specific header injection, broker.apiHosts allowlist). The PR also picks up the deferred M3 work since the initial open: brokerSession is now materialized into the agent runtime env via heartbeat (M3.2 done centrally rather than per-sandbox plugin), the BYO push-target REST API is wired under /api/companies/:co/oauth/connections/:id/broker-targets (M3.4), and the refresh worker fans out to BYO targets on every successful rotation (M3.5 BYO half). Re-running Greptile.

stubbi:
@greptileai re-review please — addressed your four findings in 61f806a4 (split/join for `{value}` substitution with regression test for `$`-sequences, bracket-aware `parseConnectTarget` with 7 unit tests for IPv6, provider-yaml `broker.headerInjection` override threaded through `buildCredentialBrokerCtx`, `broker.apiHosts` allowlist merged into `extractHostsFromProviderConfig`). PR description also refreshed to reflect the M3 work that shipped since the initial open.

stubbi:
@greptileai re-review please — round-2 findings all addressed in 82886557:

- **P0 leaf-cert expiry anchored to CA creation** → `signLeaf` now captures the signing time and stores `expiresAt` on each cache entry; entries with <60 s remaining validity are evicted and re-signed. Regression test asserts a re-signed leaf has a future `notAfter`.
- **Session token leaked into logs** → `broker.ts` drops `sessionToken` from the structured payload; `runId` + `companyId` are sufficient for correlation.
- **BYO push targets allowed http://** → validator tightened to `https://` only; the test that previously asserted http acceptance now asserts rejection.
- **Orphaned secret on add failure** → route now validates URL up front via `validateBrokerTargetInput` *before* `upsertSecretByName`, and on any `svc.add` failure attempts `secretService.remove(secret.id)` with a warn-log fallback.

stubbi:
@greptileai re-review please — round-3 P1 addressed in 14ef98ae.

The BYO push fan-out no longer runs inside `deps.db.transaction()`. The transaction body now drains `{row, accessToken}` pairs into a `pendingByoPushes` list; `pushToByoBrokerTargets` runs in a post-commit loop after the advisory lock has released and the pool connection has returned. Regression test asserts call order is `tx-begin → tx-commit → byo-fetch` by spying on `fetch` and instrumenting the mocked `db.transaction`. Pushes remain best-effort with the DB row as source of truth, so a transient operator-broker outage is recovered on the next refresh.

stubbi:
@greptileai take another pass please — the BYO push-target transaction-hold P1 you flagged is fixed in 14ef98ae (push fan-out drained to a list inside the tx, fetches run post-COMMIT). Regression test asserts `tx-begin → tx-commit → byo-fetch` ordering.

stubbi:
@greptileai take another pass please — round-4 P1 addressed in 8197c241.

`materializeBrokerSessionForRuntime` now embeds the session token in the proxy URL's userInfo (`http://session:<encoded-token>@host:port`). The listener gained `extractSessionTokenFromProxyAuth` which accepts both `Bearer` (used by our e2e harness) and `Basic` (the canonical scheme curl/requests/https-proxy-agent emit when the proxy URL has userInfo) — URL-decoding the password portion to mirror the materializer's URI encoding. Five new unit tests cover both schemes plus URL-reserved-character round-trips; the existing runtime-env test was updated to assert the token round-trips through `URL.password`. Also de-flaked the leaf-cert expiry regression test (bumped TTL from 1s to 3s + asserts second-leaf notAfter > first-leaf notAfter).

stubbi:
@greptileai take another pass — round-5 operational nits addressed in a384ef62.

1. **DELETE /broker-targets orphaned secret** → handler now reads the target via `svc.list` before remove, then `secretService.remove(authTokenSecretId)` with a warn-log catch fallback. Mirrors the POST add-failure cleanup path.
2. **IPv6 listenAddress** → `listen(host, port)` now brackets IPv6 hosts when assembling `listenAddress` so `proxyUrl()` returns `http://[::1]:port` (parseable by `new URL`); 127.0.0.1 path unchanged.

stubbi:
@greptileai take another pass — round-6 non-blocking nits addressed in 70043061.

1. **Session store leak (no scheduled prune)** → built-in broker's `start()` now schedules `setInterval(store.prune, 60s)` (`.unref()`-ed); `stop()` clears it.
2. **Registry double-init race** → `resolveCredentialBroker` switched to a `pending` Promise pattern so concurrent first dispatches all await the same factory invocation. New regression test asserts factory runs exactly once under `Promise.all([3 concurrent calls])`.
3. **HTTPS_PROXY/HTTP_PROXY not in secretKeys** → these are added to `secretKeys` immediately after the broker-env merge in `heartbeat.ts`, so `onAdapterMeta`'s redaction masks them if an adapter ever echoes its launch env in metadata.

stubbi:
@greptileai re-review please — `origin/master` merged in at `b8ab8ccd` after the round-6 fixes. The Greptile Review check didn't fire automatically on the merge commit, so this is a manual trigger. No new product code in the merge — only the integration of master changes (incl. plugin SDK vitest.config restructuring).

stubbi:
Closing as superseded. The M1+M2 goal of this PR — the agent process never sees the plaintext OAuth bearer, with the credential injected at the transport boundary — is implemented on current master (`d293dd3d1`) by the MCP tool-gateway, landed in:

- `cfa5e0704` feat(mcp) [split 3/8]: add tool access policy core (#9558)
- `1de0a3bb1` feat(mcp) [split 2/8]: add governed access contracts (#9557)
- `6244e4cf3` feat(apps): add Composio and Gmail connectors (#12342)

Concretely, in `server/src/services/tool-gateway.ts` the agent is issued a scoped gateway session bearer (`namedGatewaySessionFromBearer` / `tokenPrefixFromNamedBearer`) and its MCP client config is wired with a placeholder `Authorization: Bearer ${bearerPlaceholder}`; the real provider credential is attached server-side only when the gateway proxies upstream (`headers.Authorization = \`Bearer ${await resolveOAuthGrantSecret(...)}\``), with the same projection in `server/src/services/tool-access.ts` (`projectedConnectionHeaders`). Refresh-on-expiry and revocation handling live in the same files, and `server/src/services/tool-oauth-legacy-backfill.ts` shows an earlier credential generation was already migrated onto this model.

Master achieves this with a session/gateway-token model rather than this PR's TLS-intercepting forward proxy (`HTTPS_PROXY` / `SSL_CERT_FILE` injection), so the one thing not covered identically is the partial-M3 "BYO push target" case where an arbitrary external process reaches a real host directly. If that case is still wanted it deserves a fresh, much smaller PR against the current gateway rather than this 82-file branch. Closing.

PR #5837, opened 2026-05-12
feat(secrets): Vault/OpenBao secret provider (replaces coming_soon stub)

stubbi:
## Thinking Path

> - Those agents need secrets (API keys, OAuth tokens, DB creds) injected into their runtime, and Paperclip already ships a pluggable `SecretProvider` surface with `local_encrypted` + `aws_secrets_manager` real, and `gcp_secret_manager` + `vault` stubbed as `coming_soon`
> - For self-hosted Paperclip on Kubernetes (and the future helm chart / operator), neither real provider fits — `local_encrypted` is single-node and AWS is a non-starter for non-AWS clusters
> - The K8s-native option of using `Secret` objects directly is too weak by default (base64 in etcd unless the cluster admin separately configured an `EncryptionConfiguration` with a KMS provider, and namespace read on `secrets` is a frequent over-grant), so we explicitly reject `kubernetes_secret` as a primary backend
> - This pull request promotes the `vault` provider from `coming_soon` to a real implementation talking the Vault HTTP API, so it works against both **OpenBao** (MPL-2.0 Linux Foundation fork) and **HashiCorp Vault** (BUSL-1.1) unchanged
> - The Paperclip helm chart (separate milestone) will bundle OpenBao via helm dependency and pre-configure the Kubernetes auth method so secrets work out of the box on a vanilla cluster; operators with an existing Vault deployment point the same provider at their endpoint instead
> - The benefit is a proper self-hosted secret backend with the same `SecretProviderModule` shape as AWS (per-company vaults, workload-identity bootstrap, external references, health checks, audit) — no new abstractions, no fork in how local-vs-cloud is configured

No existing issue covers this directly — problem described in-PR (feature-gap shape):

- The pluggable `SecretProvider` surface ships `local_encrypted` + `aws_secrets_manager` as real providers, while `vault` and `gcp_secret_manager` are stubbed as `coming_soon`.
- For self-hosted Paperclip on Kubernetes (and the future helm chart / operator), neither real provider fits: `local_encrypted` is single-node and AWS Secrets Manager is a non-starter for non-AWS clusters.
- Using Kubernetes `Secret` objects directly is too weak as a primary backend (base64 in etcd unless the cluster admin separately configured KMS encryption; namespace `secrets` read is a frequent over-grant), so this PR promotes the `vault` provider to a real implementation that works against both OpenBao and HashiCorp Vault.

## Stacked on

This PR stacks on **#5836** (`feature/credential-broker-m1`). The unique vault commits start at `75c1b2bd docs(secrets): design spec for vault provider (OpenBao-bundled)`. The broker commits below have already been reviewed across multiple rounds on #5836 and are duplicated here for compilability; once #5836 lands the diff collapses to just the vault provider work.

## What Changed

- **New provider module** `server/src/secrets/vault-provider.ts` (~1000 lines after Greptile round-1 fixes)
  - `createVaultProvider({ config?, gateway? })` factory, sibling of `awsSecretsManagerProvider`
  - Production `UndiciVaultGateway` (no `node-vault` dep) + an internal `VaultHttpGateway` interface for test injection
  - `VaultTokenManager` with proactive renewal at 70% TTL + 30s expiry skew + single-retry on 403
  - `Map<configKey, { gateway, tokenManager }>` session cache per provider instance — one login per unique vault, gateway captures its specific token manager (Greptile #1 + #2)
  - `resolveVaultConfig` layering: vault config → env (`PAPERCLIP_SECRETS_VAULT_*`) → static defaults
  - `detectVaultAuthSource` discriminated union: `kubernetes` (in-cluster SA), `token` (local `VAULT_TOKEN`), `error`
  - KV v2 paths: `<mount>/data/<prefix>/<deploymentId>/<companyId>/<key>` with single `value` data key
  - Vault-enforced retention via `max_versions`; per-vault `versionRetention` 2–100, default 10
  - External references: `<mount>/<path>[#<dataKey>]`, overlap denylist against managed prefix, pure-dot-segment denylist for path traversal (Greptile #3)
  - Soft-delete via KV v2 `DELETE` (hard-destroy intentionally NOT in the default policy)
  - 4-probe `healthCheck`: reachability (`sys/health`) + auth (k8s-login or `lookup-self`) + KV mount (rejects v1) + capabilities (`sys/capabilities-self`). Vault token never appears in the response.
- **Registry wiring** `server/src/secrets/provider-registry.ts` swaps the `vault` import from the `external-stub-providers.ts` stub to the new real module; stub export removed.
- **Operator doc** `doc/SECRETS-VAULT-PROVIDER.md` covers bootstrap trust model, deployment config, KV path/policy, helm-chart OpenBao bundling, health probes, runbooks (token rotation, unseal, leaked-token incident).
- **Updated docs** `docs/deploy/secrets.md` (vault no longer marked `coming_soon`; helm + external-Vault variants both documented), `docs/api/secrets.md` (full per-vault config schema with credential-shaped-field denylist).
- **Design spec + plan** `docs/superpowers/specs/2026-05-12-vault-secret-provider-design.md` (539 lines) and `docs/superpowers/plans/2026-05-12-vault-secret-provider-plan.md` (3574 lines), produced via the brainstorming → writing-plans → subagent-driven-development skill chain.

## Verification

- **Unit tests:** 76 tests in `server/src/__tests__/vault-provider.test.ts`, all passing. Covers validation (5 categories: address, mount/prefix/role/retention, credential denylist, config layering, auth detection), `VaultTokenManager` lifecycle (initial login, proactive renewal, expiry skew, invalidate), 403-retry guard, `createSecret`, `createVersion` (with CAS), `resolveVersion` (latest + pinned + not_found), `linkExternalSecret` (+ overlap denylist + pure-dots rejection), external `resolveVersion`, `deleteOrArchive` (soft-delete + archive no-op + external no-op), `healthCheck` (4 probes; sealed/v1/unreachable/token-not-leaked), `UndiciVaultGateway` HTTP shape (headers + status-to-error mapping), and the Greptile round-1 regression suite (token caching, path-traversal rejection, env `parseInt`).
- **Registry tests:** 3 tests in `server/src/__tests__/secret-provider-registry.test.ts`, all passing. New assertion confirms the vault descriptor reports `configured: true` and `supportsManagedValues: true`.
- **Typecheck:** `pnpm -C server tsc --noEmit` clean.
- **Opt-in integration tests:** `server/src/__tests__/vault-provider-integration.test.ts` — 3 scenarios (create + rotate + resolve(latest) + resolve(version), retention enforces `max_versions = 5`, health check returns ok), skipped by default so the standard `pnpm test` cycle does not require Docker. Run with:
  ```sh
  bao server -dev -dev-root-token-id=root &
  PAPERCLIP_TEST_VAULT=1 \
  PAPERCLIP_TEST_VAULT_ADDR=http://127.0.0.1:8200 \
  PAPERCLIP_TEST_VAULT_TOKEN=root \
  pnpm -C server vitest run src/__tests__/vault-provider-integration.test.ts
  ```
- **Manual smoke on a real cluster:** deferred to the future helm-chart milestone (this PR ships the provider; the chart that consumes it is a separate workstream).

## Risks

- **Custody surface:** the provider adds a real network path from the Paperclip server to a Vault/OpenBao endpoint with token auth. Mitigated by: workload-identity-only bootstrap (no credentials in `company_secrets`, denylist enforced at validation time), origin-only address validation (no SSRF), per-vault config caching with closure-captured token managers (no cross-vault auth contamination), server-side resolution only (plaintext never returned to the board UI), and an explicit pure-dots rejection in `parseExternalRef` (no path traversal on operator-supplied refs).
- **No DB migration**: `vault` already lived in the `SecretProvider` enum. Swap-in is a code-only change — operators with existing `coming_soon` vault provider configs see those configs become `ready` automatically once the deployment env or per-vault config is set.
- **Default-provider behaviour unchanged**: `local_encrypted` remains the deployment default for local installs; the vault provider only lights up when its env or per-vault config is set. Existing AWS / `local_encrypted` deployments unaffected.
- **No helm chart in this PR**: the operator doc describes the eventual OpenBao-bundling shape but the chart itself is a separate milestone. Risk: the doc describes a not-yet-shipped artifact. Acceptable — the doc is explicit and the chart milestone will land before any operator can actually `--set bao.enabled=true`.
- **Greptile round-1 review** flagged two structural defects in the original token-manager wiring (per-call manager construction killing the cache; shared `activeTokenSupplier` race across concurrent ops). Both are fixed in `7fd4e338` with regression tests; the design now caches one `(gateway, tokenManager)` per unique config and each gateway captures its specific token manager via closure.

## Model Used

- **Provider:** Anthropic (via Claude Code CLI + subagent-driven-development skill)
- **Model:** Claude Opus 4.7 (`claude-opus-4-7`)
- **Context window:** 1M (`claude-opus-4-7[1m]`)
- **Reasoning mode:** standard inference; no extended thinking
- **Tool use:** full Claude Code toolset (Bash, Edit, Read, Write, Agent, etc.); 22-task implementation plan was executed by `general-purpose` subagents dispatched from the controller, mostly with `haiku` (cheap mechanical TDD), with `sonnet` for the larger integration tasks (`createSecret`, `UndiciVaultGateway`, `healthCheck`)
- **Capability details:** memory subsystem persisted decisions across sessions; brainstorming + writing-plans + subagent-driven-development skills enforced the TDD-per-task loop

## Checklist

- [x] I have run tests locally and they pass (76 unit + 3 registry, 3 integration skipped by default)
- [x] I have added or updated tests where applicable (76 unit tests including 6 Greptile round-1 regressions, 3 opt-in integration scenarios)
- [x] If this change affects the UI, I have included before/after screenshots — **N/A** (no UI changes; existing `Secrets` and `ImportFromVaultDialog` pages already render the `vault` provider via descriptor, status flips from `coming_soon` to `ready` automatically)
- [x] I have updated relevant documentation to reflect my changes (`doc/SECRETS-VAULT-PROVIDER.md` operator contract, `docs/deploy/secrets.md`, `docs/api/secrets.md`)


stubbi:
@greptileai review please — initial review.

I retargeted the PR base from `master` to `pap-9234-replace-pr-5647` (the deeper base of #5836) to get under the 100-file limit (now 54 files).

This PR stacks on **#5836** (credential-broker M1+M2+partial M3). The vault-specific commits at the head of the branch — starting at `75c1b2bd docs(secrets): design spec for vault provider (OpenBao-bundled)` — are what's new in this PR. The broker commits below are duplicated from #5836 and have already been reviewed there over multiple rounds.

The vault provider work in scope:
- new module `server/src/secrets/vault-provider.ts` (~980 lines): `createVaultProvider` factory, `UndiciVaultGateway`, `VaultTokenManager` (proactive renewal + 403-retry guard), KV v2 paths, external-references with overlap denylist, 4-probe health check
- registry wiring: `vault` swapped from `coming_soon` stub to real provider
- 70 unit tests + 3 opt-in integration tests (gated by `PAPERCLIP_TEST_VAULT=1`)
- operator doc `doc/SECRETS-VAULT-PROVIDER.md` + updates to `docs/deploy/secrets.md` and `docs/api/secrets.md`
- design spec + plan under `docs/superpowers/`

Custody / trust model mirrors the AWS provider exactly. Credential-shaped fields are denylisted at validation time. KV v2 + OpenBao-bundled-by-default in the future helm chart; provider also works against HashiCorp Vault unchanged.

stubbi:
@greptileai re-review please — round-1 findings addressed in 7fd4e338, plus PR-template compliance:

- **#1 Per-operation `VaultTokenManager` defeats token caching** → factory now keeps a `Map<configKey, { gateway, tokenManager }>` session cache; each unique vault config gets one `VaultTokenManager` for the provider's lifetime, so `auth/kubernetes/login` runs once per vault (not per call) and `VaultTokenManager`'s 70%-of-TTL proactive renewal is actually effective. Regression test: kubernetes-auth provider with mocked `loginKubernetes`; 3 sequential ops on the same config trigger exactly 1 login call.

- **#2 Shared `activeTokenSupplier` can be stomped by concurrent operations** → the mutable closure slot is gone entirely. Each cached gateway's `getToken` is a closure that captures its specific `VaultTokenManager` (via `let tokenManager!: VaultTokenManager` declared before, assigned after gateway construction — closure-binding semantics make the order safe). Concurrent ops against different vault configs cannot cross-contaminate authentication.

- **#3 External ref `mount` not sanitised → path traversal** → `parseExternalRef` now applies `KV_MOUNT_PATTERN` to the mount AND explicitly rejects pure-dot segments (`/^\.+$/`) on both mount and path. The pure-dots check is separate from the alphanum/dot/dash regex because operators legitimately use mount names like `my.kv.v2` (a single `.` is fine; `..` is not). Three regression tests cover `..` mount, `.`/`..` path segments, and invalid-character mounts.

- **#4 `Number(env)` silently produces 0/NaN for empty/non-numeric** → swapped to `parseInt(env, 10)`. Empty/non-numeric flow through the existing "must be an integer" warning instead of coercing. Two regression tests: `"abc"` produces error, `"07"` parses as 7.

- **#5 `removeTargetById` silent no-op** → this is in `server/src/services/broker-targets.ts` which is part of the credential-broker work from **#5836**, not this PR's vault provider work. Flagging here for clarity but leaving the fix to the broker PR (the broker code is duplicated into this branch only for compilability; the file diff will collapse to zero once #5836 merges). Happy to forward the finding upstream if helpful.

- **#6 PR template compliance** → updated the PR description with the required Thinking Path (5–8-step blockquote tracing from Paperclip → subsystem → gap → this PR), Risks, Model Used (Claude Opus 4.7 1M context via subagent-driven-development skill — haiku for mechanical TDD tasks, sonnet for the larger integration tasks), and the Checklist with all items ticked. Pre-existing sections (What Changed, Verification, Stacked-on context) preserved.

Full test status after the fixes: **76 unit tests + 3 registry tests passing, typecheck clean, 3 opt-in integration tests skipped by default** (`PAPERCLIP_TEST_VAULT=1` to run them against a real `bao server -dev`).

stubbi:
@greptileai re-review please — addressed the external-ref path-segment validation gap in `42334025`:

`parseExternalRef` now validates **every** path segment against `KV_MOUNT_PATTERN` (`/^[A-Za-z0-9._-]+$/`), not just the mount. Segments containing `?`, `&`, space, `%`, or any other URL-special character are rejected with the same `unprocessable(...)` error class used for the existing pure-dot check.

Reused `KV_MOUNT_PATTERN` verbatim rather than introducing a separate `KV_PATH_SEGMENT_PATTERN` — the semantic constraint is identical (single path segment, safe URL chars only), so a parallel constant would be pure duplication.

**New tests** in `server/src/__tests__/vault-provider.test.ts` (+5, appended to the `parseExternalRef` describe block):
- Reject segment containing `?`
- Reject segment containing `&`
- Reject segment containing a space
- Reject segment containing `%`
- Sanity: multi-segment ref with `.`, `_`, `-` still parses cleanly

Full file results: 81 passed, 3 skipped (integration suite, env-gated), 0 failed. Typecheck clean.

(Note: dropped the `#` test I initially drafted — `pathPart.split("#", 2)` truncates at the first `#`, so a second `#` never reaches a segment to begin with. Replaced with `&` for coverage.)

stubbi:
@greptileai re-review please — addressed both P1s in `24522aa4`:

**P1 — SA JWT re-read on Kubernetes re-login.** `VaultTokenManager` now accepts an optional `readSaToken: (path) => string | null` constructor seam (defaults to `fs.readFileSync(path, "utf8").trim()`, matching the provider-level helper pattern). `acquireInner` calls it before every `loginKubernetes`, so when the Vault token hits `max_ttl` and the on-disk SA JWT has rotated (k8s ~48-min refresh by default), the re-login uses the fresh token. Fail-safe: if the read returns null or throws, we log + fall back to the cached `source.jwt` and let `withVaultTokenRetry`'s 403 path surface real failures.

**P1 — `TOKEN_EXPIRY_SKEW_MS` actually applied** (option A from your suggested fix). Renewal threshold is now `ttlMs * TOKEN_RENEWAL_THRESHOLD - TOKEN_EXPIRY_SKEW_MS`, with an inline comment explaining the clock-skew rationale. The previously-dead constant no longer misleads operators reading the docs.

**Tests added** (`vault-provider.test.ts` +4):
1. Re-reads SA JWT from disk on each login — verifies login N+1 uses the rotated value, never the stale cached `source.jwt`.
2. Falls back to cached `source.jwt` when the SA token file is unreadable (EACCES) — single `console.error`, login still proceeds with cached value.
3. Token within `TOKEN_EXPIRY_SKEW_MS` of the naive threshold IS treated as renewable.
4. Token just under the skewed threshold is still cached (renewal not over-eager).

85/85 pass; typecheck clean. `pnpm-lock.yaml` churn reverted per the lockfile policy.

stubbi:
@greptileai re-review please — round-4 P1s addressed in 42469a15, with origin/master merged in via 7c849e64 to stay current.

1. **`vault` still in `COMING_SOON_SECRET_PROVIDERS` (P1)** → Removed in `server/src/services/secrets.ts`. New vault configs now default to `ready` status, `healthCheckSecretProviderConfig` reaches the real provider, and `updateProviderVaultConfig` no longer rejects `status="ready"` transitions. Existing service-layer tests cover the activation paths; all 125 vault+secrets tests pass.

2. **`UndiciVaultGateway.renewSelf()` deadlock (P1)** → Added an explicit `token: string` parameter to `renewSelf` on the `VaultHttpGateway` interface and an explicit-token override on `UndiciVaultGateway.call`. The renewal path now forwards `this.cached.token` directly instead of awaiting `getToken()`, so it cannot re-enter `tokenManager.acquire()` while `acquireInner()` is in flight. New regression test "renewSelf uses the supplied token and does not re-enter getToken (deadlock guard)" exercises the real `UndiciVaultGateway` via undici's MockAgent with a `getToken` that throws if called.

stubbi:
@greptileai re-review please — round-5 fixes addressed.

1. **P1 broker-registry double-init race** → already fixed on the credential-broker branch (#5836) in 70043061 (greptile r6 there). Merged into this PR via 0d521271. \`resolveCredentialBroker\` now stores the inflight resolution promise so concurrent first-time callers all await the same Promise; the regression test in #5836 asserts the factory runs exactly once under \`Promise.all\` with 3 concurrent calls. The merge also brings the credential-broker SDK + builtin proxy listener + session store into this PR's diff.

2. **P2 console.error → structured logger** → ba939163. \`VaultTokenManager.acquireInner\` now imports the shared pino logger and emits \`logger.warn({ err, saTokenPath }, "vault-provider: failed to re-read SA token; falling back to cached JWT")\` on the SA-token re-read failure path. Existing "falls back to the cached source.jwt when the SA token file is unreadable" test updated to assert on logger.warn (and verifies the structured \`saTokenPath\` field is carried).

86/86 vault tests pass locally.

PR #5842, opened 2026-05-13
fix(claude-local): replace symlink skill materialization with file copy (Windows EPERM)

oalmadani:
## Thinking Path

> - Paperclip orchestrates AI agent teams, injecting runtime skills into each agent's working context on every run
> - The `adapter-claude-local` package prepares a stable prompt bundle directory per agent, materializing Paperclip skills so Claude can reference them
> - Skill materialization in `prompt-cache.ts` calls `ensurePaperclipSkillSymlink`, which uses `fs.symlink()` to create directory symlinks pointing to skill source directories
> - On Windows without Developer Mode or elevated privileges, `fs.symlink()` throws `EPERM: operation not permitted` — a well-known Windows restriction
> - When symlink creation fails, the skill directory is never populated; the agent starts with no skills, produces no output, and triggers false-positive "silent run" alerts
> - This PR replaces `ensurePaperclipSkillSymlink` with `materializePaperclipSkillCopy`, which copies skill files directly (no symlinks) and has fingerprint-based caching to avoid redundant copies on re-runs
> - The benefit is zero-config skill materialization on any Windows host regardless of Developer Mode, matching the approach already used by `adapter-acpx-local`

## What Changed

- `packages/adapters/claude-local/src/server/prompt-cache.ts`: replaced `ensurePaperclipSkillSymlink` import with `materializePaperclipSkillCopy` and updated the skill materialization loop — no symlinks created, uses atomic file copy with fingerprint caching
- Logs a stdout note when any symlinks inside a skill source directory are skipped during copy (matching existing behavior in `adapter-acpx-local`)

## Verification

1. On a Windows host **without** Developer Mode: run a Paperclip agent with at least one skill configured — skills should materialize successfully with no EPERM error in run logs
2. On any platform: run a second agent run with the same skill set — `materializePaperclipSkillCopy` fingerprint check skips the redundant copy, same behavior as before
3. Existing tests pass: `pnpm test` in `packages/adapters/claude-local`

## Risks

- **Behavior change**: skills are now copied as real files instead of symlinks. The files are read-only to Claude so functional behavior is identical. The cache directory grows slightly (real files vs symlink pointers), but `materializePaperclipSkillCopy` handles this correctly with fingerprint-based skip logic.
- **Low risk** for Linux/macOS: symlinks already worked on those platforms; the file-copy path produces the same result for Claude (it reads skill files, not follows symlinks).

## Model Used

- Provider: Anthropic
- Model: `claude-sonnet-4-6` (Claude Sonnet 4.6)
- Context window: 200K tokens
- Capabilities: tool use, code editing, multi-step reasoning

## Checklist

### DIFF
--- a/packages/adapters/claude-local/src/server/prompt-cache.ts
+++ b/packages/adapters/claude-local/src/server/prompt-cache.ts
@@ -4,7 +4,7 @@ import path from "node:path";
 import { createHash, type Hash } from "node:crypto";
 import type { AdapterExecutionContext } from "@paperclipai/adapter-utils";
 import {
-  ensurePaperclipSkillSymlink,
+  materializePaperclipSkillCopy,
   resolvePaperclipInstanceRootForAdapter,
   type PaperclipSkillEntry,
 } from "@paperclipai/adapter-utils/server-utils";
@@ -149,7 +149,13 @@ export async function prepareClaudePromptBundle(input: {
   for (const entry of skills) {
     const target = path.join(skillsHome, entry.runtimeName);
     try {
-      await ensurePaperclipSkillSymlink(entry.source, target);
+      const result = await materializePaperclipSkillCopy(entry.source, target);
+      if (result.skippedSymlinks.length > 0) {
+        await onLog(
+          "stdout",
+          `[paperclip] Materialized Claude skill "${entry.runtimeName}" into ${skillsHome} and skipped ${result.skippedSymlinks.length} symlink(s).\n`,
+        );
+      }
     } catch (err) {
       await onLog(
         "stderr",


timlevel:
Code Review: paperclip PR #5842 - fix(claude-local): replace symlink skill materialization with file copy (Windows EPERM)

SUMMARY:
This PR addresses a Windows compatibility issue in the claude-local adapter by replacing symlink-based skill materialization with file copying. This resolves EPERM errors on Windows systems without Developer Mode or elevated privileges.

POSITIVE ASPECTS:

1. Clean and focused fix that directly addresses the root cause
2. Maintains backward compatibility while improving cross-platform support
3. Uses existing infrastructure (materializePaperclipSkillCopy) that's already proven in adapter-acpx-local
4. Includes proper logging for skipped symlinks
5. Well-documented with clear thinking path and verification steps

AREAS FOR IMPROVEMENT:

1. **Performance Considerations**:
   - File copying may be slower than symlinks on systems where symlinks are supported
   - Consider adding metrics to monitor the performance impact

2. **Documentation**:
   - Add documentation about the behavioral change (files vs symlinks) for users who might be relying on symlink behavior
   - Consider updating any related documentation that mentions symlink usage

3. **Testing**:
   - Add specific tests for Windows compatibility scenarios
   - Consider adding performance benchmarks to monitor the impact of file copying vs symlinks

RECOMMENDATIONS:

1. Monitor performance impact of file copying vs symlinks
2. Add Windows-specific tests to prevent regressions
3. Document the behavioral change for users who might be relying on symlink behavior
4. Consider adding configuration option to choose between symlink and copy approaches if needed

This is an excellent fix that improves cross-platform compatibility without breaking existing functionality. The approach of using file copying instead of symlinks is actually more robust since it works consistently across all platforms without requiring special permissions. The use of existing infrastructure (materializePaperclipSkillCopy) reduces risk and leverages proven code.

PR #5850, opened 2026-05-13
fix(server): walk drizzle .cause chain in unique-violation predicates (ZERA-582)

rsavitt:
## Summary

Three service-layer predicates were doing shallow `error.code === \"23505\"` checks against the top-level error. `drizzle-orm/postgres-js` wraps the raw `PostgresError` inside a `DrizzleQueryError`, putting `code` / `constraint` / `constraint_name` on `.cause` — so each predicate was silently returning false on real concurrent-write conditions and letting the violation fall through to the generic 500 handler.

Same pattern as [PR #5848](https://github.com/paperclipai/paperclip/pull/5848) (documents.ts, [ZERA-578](/ZERA/issues/ZERA-578)). Each call site gets a cycle-guarded `.cause` walk that checks both `constraint` and `constraint_name`.

## Files

- `server/src/services/companies.ts:127` — `isIssuePrefixConflict` (`createCompanyWithUniquePrefix` retry loop).
- `server/src/services/routines.ts:1247` — `isOpenExecutionConflict` (`dispatchRoutineRun` concurrency-policy branch).
- `server/src/services/plugin-registry.ts:40` — `isPluginKeyConflict` (plugin install path).

## Why not extract a shared util now

[ZERA-582](/ZERA/issues/ZERA-582) AC explicitly says \"optional shared-util extraction *only if* more than one site converges (don't over-abstract on the first sweep)\". All three sites now converge on the same walk shape, plus the `documents.ts` predicate from #5848 — so a shared `walkCauseChain(error, predicate)` helper is the natural next consolidation. Filing as a follow-up after #5848 merges so the canonical helper-shape from documents.ts is the source of truth.

## Test plan

- [x] `vitest run server/src/__tests__/routines-service.test.ts` — 32 pass (covers `dispatchRoutineRun` conflict-handling path).
- [ ] CI server typecheck/tests (note: `master` HEAD has pre-existing typecheck breakage tracked in [ZERA-564](/ZERA/issues/ZERA-564) — unrelated to this PR).

No new tests added: each predicate is exercised by existing service tests; the change is a behavioral fix on an existing predicate, not new logic. Adding embedded-postgres concurrency tests for all three (à la #5848) is the natural follow-up — what was broken is the conflict detection itself, which is verified by the existing suite passing with the new predicate.

## Tracking

Fixes [ZERA-582](/ZERA/issues/ZERA-582). Cataloged in the [Shared Learnings Hub pattern library](/ZERA/issues/ZERA-401#document-pattern-library) — \"API quirk: drizzle-orm/postgres-js wraps PostgresError on .cause\".

### DIFF
--- a/server/src/services/companies.ts
+++ b/server/src/services/companies.ts
@@ -125,16 +125,19 @@ export function companyService(db: Db) {
   }
 
   function isIssuePrefixConflict(error: unknown) {
-    const constraint = typeof error === "object" && error !== null && "constraint" in error
-      ? (error as { constraint?: string }).constraint
-      : typeof error === "object" && error !== null && "constraint_name" in error
-        ? (error as { constraint_name?: string }).constraint_name
-        : undefined;
-    return typeof error === "object"
-      && error !== null
-      && "code" in error
-      && (error as { code?: string }).code === "23505"
-      && constraint === "companies_issue_prefix_idx";
+    // Walks the `.cause` chain because `drizzle-orm/postgres-js` wraps the raw
+    // `PostgresError` inside a `DrizzleQueryError` — a shallow check misses
+    // the wrapped violation (see ZERA-582).
+    const seen = new Set<unknown>();
+    let current: unknown = error;
+    while (current && typeof current === "object" && !seen.has(current)) {
+      seen.add(current);
+      const err = current as { code?: string; constraint?: string; constraint_name?: string; cause?: unknown };
+      const constraint = err.constraint ?? err.constraint_name;
+      if (err.code === "23505" && constraint === "companies_issue_prefix_idx") return true;
+      current = err.cause;
+    }
+    return false;
   }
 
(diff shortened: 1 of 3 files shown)


harperaa:
Hit this in a downstream fork (`harperaa/paperclip`) and applied the same patch fork-side after independently filing #6462/#6463 (now closed as duplicates of #6088/this PR). Confirms the diagnosis and the fix.

Reliable reproduction: any second company whose name's first 3 uppercased alpha chars match an existing company's `issue_prefix` returns HTTP 500 — `createCompanyWithUniquePrefix` catches the wrapped `DrizzleQueryError`, the shallow code/constraint check misses the inner `PostgresError`, and the retry loop gives up on attempt #1.

Strong +1 to merge — this is the right level of fix. The other 5 open PRs targeting the same bug only patch the single `isIssuePrefixConflict` site; #5850 generalizes to all three unique-violation predicates (companies, plugins, routines) which means the same bug class won't bite again somewhere new the next time someone adds a fourth predicate.

Issue #5288, opened 2026-05-05
agent-auth-jwt.ts falls back to BETTER_AUTH_SECRET for JWT signing


f22-pro:
Issue body (paste verbatim)
Affected version: @paperclipai/server 2026.428.0 (and earlier).

Source: server/src/agent-auth-jwt.ts (compiled to dist/agent-auth-jwt.js), jwtConfig() at L9–L18:

const secret =
  process.env.PAPERCLIP_AGENT_JWT_SECRET?.trim() ||
  process.env.BETTER_AUTH_SECRET?.trim();
if (!secret) return null;

Why this is a bug (security finding). External AppSec audit of the agent JWT issuance trust model flagged this as a HIGH-severity cross-system key confusion finding:

The better-auth session secret was never intended to sign agent JWTs. Reusing it means a bug in either system, or a rotation of either secret, silently affects both.
The fallback is silent — operators get no warning that they are running on the better-auth secret instead of a dedicated agent JWT secret.
The function already returns null (and skips issuance) when both env vars are absent, which is the correct fail-closed behavior. Removing the fallback simply makes that fail-closed path the only path, matching Economy of Mechanism + Secure Defaults.

Proposed fix:
- const secret =
-   process.env.PAPERCLIP_AGENT_JWT_SECRET?.trim() ||
-   process.env.BETTER_AUTH_SECRET?.trim();
+ const secret = process.env.PAPERCLIP_AGENT_JWT_SECRET?.trim();
  if (!secret) return null;

Pure removal — no new behavior. Operators that were depending on the fallback simply need to set PAPERCLIP_AGENT_JWT_SECRET explicitly (which the README/runbook should already document as required).

Migration note for release notes: "PAPERCLIP_AGENT_JWT_SECRET is now required for agent JWT issuance. The previous implicit fallback to BETTER_AUTH_SECRET has been removed — instances missing this env var will skip JWT issuance (the existing null path) instead of silently reusing the session secret."

PR #3967, opened 2026-04-18
security(server): close cross-tenant existence oracle (404 instead of 403)

stubbi:
## Thinking Path

> - Paperclip orchestrates AI agents for zero-human companies
> - In a multi-tenant deployment, route handlers that take a resource id (`issue`, `goal`, `project`, `approval`, etc.) look the resource up by id and then call `assertCompanyAccess` on its `companyId` — 404 if it doesn't exist, 403 if it exists in another tenant
> - The split status codes are a classic *existence oracle*: any authenticated user can enumerate ids across tenants by probing for the 403/404 boundary, mapping out which issues, labels, approvals, etc. exist in other customers' tenants even when they cannot read the contents
> - The right fix is a single uniform 404 for both "not found" and "found but cross-tenant", which collapses the oracle but still preserves write-path checks (active membership, viewer-readonly) for *authorized* tenants
> - This pull request adds a non-throwing `hasCompanyAccess(req, companyId)` helper plus a `getAccessibleResource` wrapper that ~130 handlers across 14 route files now use, folding the access check into the existence check while still running `assertCompanyAccess` for authorized tenants so viewer-readonly / inactive-membership rejections fire unchanged on write paths
> - The benefit is closing a multi-tenant information leak without breaking write-path security or single-tenant local-first behavior

## Linked Issues or Issue Description

Refs #709 — asks for company-scope regression coverage across approval/activity/access routes, because a subtle route refactor could leak cross-tenant data; this PR hardens exactly those surfaces (uniform 404 across 14 route files including `approvals`, `activity`, `secrets`) and updates cross-tenant expectations in test files. It does not add the full coverage matrix #709 asks for — hence Refs, not Closes.

No existing issue covers the oracle itself — described in-PR:

- Route handlers returned 404 for "not found" but 403 for "exists in another tenant", a classic *existence oracle*: any authenticated user could enumerate ids across tenants by probing the 403/404 boundary.
- That maps out which issues, labels, approvals, etc. exist in other customers' tenants even when their contents are unreadable.
- Fix: a uniform 404 for both cases, while keeping write-path checks (active membership, viewer-readonly) for authorized tenants.

## What Changed

[...]

PR #5864, opened 2026-05-13
feat(security): per-company JWT signing keys for multi-tenant isolation

stubbi:
## Thinking Path

> - Agents authenticate to the server with a JWT signed by the deployment's master secret
> - In a multi-tenant deployment, all agents from every tenant are signed with the *same* key, so a leak (CI/staging dump, hostile contractor with infra access, supply-chain) lets the attacker mint tokens for *any* tenant
> - The same master secret also issued tokens with a 48-hour TTL, giving any leaked token a two-day window of validity even after rotation
> - This pull request derives a per-company signing key via `HMAC-SHA256(master, "jwt:<companyId>")` and reduces the default TTL to 1h; the verifier tries the per-company key first and falls back to the master secret only for tokens issued before this change so no agent gets locked out on deploy
> - The benefit is multi-tenant key isolation (a leak of one company's derived key cannot forge tokens for another) and a tighter blast-radius on any leaked token, with zero local-first impact (single-tenant deploys derive their one company's key the same way and continue to work unchanged)

Refs #5288 — a separate key-hygiene finding in the same module (`agent-auth-jwt.ts` falls back to `BETTER_AUTH_SECRET` as the JWT signing secret). Related agent-JWT trust-model concern, but not fixed by this PR — the master-secret fallback selection is unchanged here.

No existing issue covers this PR's problem directly — described in-PR:

- In a multi-tenant deployment, agents from every tenant get JWTs signed with the *same* master key, so a single leak (CI/staging dump, hostile contractor, supply chain) lets the attacker mint tokens for *any* tenant.
- The same master secret issued tokens with a 48-hour TTL, giving any leaked token a two-day validity window even after rotation.
- Fix: derive a per-company signing key via `HMAC-SHA256(master, "jwt:<companyId>")` and reduce the default TTL to 1h, with a master-secret verification fallback so pre-existing tokens are not locked out on deploy.

## What Changed

- **`server/src/agent-auth-jwt.ts`**
  - New `deriveCompanySigningKey(masterSecret, companyId)` — `HMAC-SHA256` with domain-separated input (`jwt:<companyId>`) so the master secret can be safely reused for other HMAC purposes in the future without cross-protocol risk.
  - `signAgentJwt` always signs with the derived per-company key.
  - `verifyAgentJwt` reads `company_id` from the token's (untrusted) claim payload, looks up the candidate derived key, and verifies. If that fails AND a master secret is set, it falls back to verifying with the raw master secret — pre-existing tokens validate until they expire. Verification still fails if the signature doesn't bind.
  - Default TTL: `60 * 60 * 48` → `60 * 60`. Existing `PAPERCLIP_AGENT_JWT_TTL_SECONDS` override still wins.
- **`PAPERCLIP_AGENT_JWT_DISABLE_LEGACY_FALLBACK`** (optional, default off) — operators set this ~one TTL after deploying to sunset the master-secret verification fallback entirely, closing the window in which a leaked master secret could forge arbitrary-`exp` tokens for any tenant.
- **`server/src/__tests__/agent-auth-jwt.test.ts`** (6 new cases)
  - Per-company isolation via tamper: token for company A fails when verified for company B.
  - Legacy-token verification path: tokens signed with the raw master secret still verify.
  - Default TTL is 1h.
  - Legacy fallback toggle: master-secret tokens accepted when unset, rejected when enabled, and per-company tokens unaffected either way.

## Verification

- `npx vitest run agent-auth-jwt` — 11/11 pass (6 new + 5 existing).
- Manual: token signed for company A under per-company key fails when verified against company B's derived key.

## Risks

- **Backward-compatible verification**, so no agent gets locked out on deploy — but operators relying on hot-swapping the master secret should note that pre-existing tokens *will* keep validating against the master key until their TTL elapses, unless `PAPERCLIP_AGENT_JWT_DISABLE_LEGACY_FALLBACK=true` is set to end the fallback window explicitly.
- **TTL reduction is a default, not a hard cap.** Operators who relied on the 48h window can override via env. If 1h is too aggressive for upstream taste, happy to gate the change behind an env var.
- **No new required env vars.** Single-tenant local-first deploys derive one company's key the same way and behave identically to today.
- **Domain-separated HMAC input** (`jwt:<companyId>`) means the master secret can be safely reused for other future HMAC purposes without cross-protocol risk.

## Model Used

Claude Opus 4.7 (1M context), extended thinking mode; rebase + legacy-fallback sunset documentation by Claude Fable 5 (1M context).

## Checklist

- [x] Model used specified
- [x] Checked ROADMAP.md — part of the multi-tenant hardening initiative
- [x] Tests run locally and pass (`agent-auth-jwt` 11/11)
- [x] Added per-company-isolation, legacy-fallback, and TTL-default tests
- [x] No UI changes
- [x] Documented risks above

Part of the multi-tenant hardening initiative — see also #3967 (cross-tenant 404 oracle) and #5865 (plugin tables `company_id`).

### DIFF
--- a/server/src/__tests__/agent-auth-jwt.test.ts
+++ b/server/src/__tests__/agent-auth-jwt.test.ts
@@ -8,6 +8,7 @@ describe("agent local JWT", () => {
   const ttlEnv = "PAPERCLIP_AGENT_JWT_TTL_SECONDS";
   const issuerEnv = "PAPERCLIP_AGENT_JWT_ISSUER";
   const audienceEnv = "PAPERCLIP_AGENT_JWT_AUDIENCE";
+  const disableLegacyFallbackEnv = "PAPERCLIP_AGENT_JWT_DISABLE_LEGACY_FALLBACK";
 
   const originalEnv = {
     secret: process.env[secretEnv],
@@ -15,6 +16,7 @@ describe("agent local JWT", () => {
     ttl: process.env[ttlEnv],
     issuer: process.env[issuerEnv],
     audience: process.env[audienceEnv],
+    disableLegacyFallback: process.env[disableLegacyFallbackEnv],
   };
 
   beforeEach(() => {
@@ -23,6 +25,7 @@ describe("agent local JWT", () => {
     process.env[ttlEnv] = "3600";
     delete process.env[issuerEnv];
     delete process.env[audienceEnv];
+    delete process.env[disableLegacyFallbackEnv];
     vi.useFakeTimers();
   });
 
@@ -38,6 +41,8 @@ describe("agent local JWT", () => {
     else process.env[issuerEnv] = originalEnv.issuer;
     if (originalEnv.audience === undefined) delete process.env[audienceEnv];
     else process.env[audienceEnv] = originalEnv.audience;
+    if (originalEnv.disableLegacyFallback === undefined) delete process.env[disableLegacyFallbackEnv];
+    else process.env[disableLegacyFallbackEnv] = originalEnv.disableLegacyFallback;
   });
 
   it("creates and verifies a token", () => {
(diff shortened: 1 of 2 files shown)


stubbi:
@greptileai re-review please — addressed the legacy-fallback sunset concern in `1208b374`:

**Fix** — Added env var `PAPERCLIP_AGENT_JWT_DISABLE_LEGACY_FALLBACK`:
- **Default (unset / falsy)** — current behavior preserved (per-company key first, master-secret fallback if that fails). Migration-safe for in-flight tokens.
- **Set to `1|true|yes|on`** — verifier rejects any token that doesn't validate under the per-company derived key. The master secret loses cross-tenant signing authority entirely.

Operator guidance (in expanded JSDoc): set this env var ~1h after deploying the per-company JWT change, once all pre-migration tokens have aged out under the new 1h TTL. From that point on, even a master-secret leak cannot be used to forge tokens directly — an attacker would need to additionally derive the per-company key via the HMAC.

Parse style matches `isTruthyEnvValue` from `server/src/ui-branding.ts` (`1|true|yes|on`, case-insensitive, trimmed) — kept the function inlined as `parseBooleanEnv` to keep `agent-auth-jwt.ts` self-contained.

**New tests** (3) — env var unset: legacy token verifies (existing behavior); env var set: legacy token rejected; env var set: per-company token still verifies. 11/11 pass.

stubbi:
@greptileai re-review please — rebased onto current master (clean, no conflicts; `agent-auth-jwt` 11/11 locally).

The single open finding from r2 (*PR description does not follow the required template*) is stale: the description was restructured into the full template (Thinking Path / Linked Issues / What Changed / Verification / Risks / Model Used / Checklist) after that review round — body edits don't re-trigger a review, hence this request. The body now also documents the `PAPERCLIP_AGENT_JWT_DISABLE_LEGACY_FALLBACK` sunset switch added in r1's follow-up commit and the six new test cases covering it.

PR #5875, opened 2026-05-13
feat(realtime): Postgres LISTEN/NOTIFY transport for cross-replica live events (supersedes #3736)

stubbi:
## Thinking Path

> - Operator UIs subscribe to per-company live events (activity, runtime state, agent status) over WebSockets
> - When a Paperclip deployment scales to `replicas: 2+`, events emitted on one replica must reach WS clients connected to other replicas, or the UI silently drifts depending on which replica the client landed on
> - #3736 added a Redis pub/sub transport, but it pinned every self-hoster to a new Redis dependency and broadcast every company's events on a single shared channel — both a friction and a metadata-leak concern
> - This pull request introduces a Postgres LISTEN/NOTIFY transport as the default (no new infra dep — Paperclip already runs Postgres) with per-company channels, keeps Redis as an opt-in alternative for operators who prefer it, and routes both through a shared `LiveEventsTransport` interface so single-replica self-hosters are unaffected
> - The benefit is correct cross-replica fan-out for multi-replica deployments with zero new infra dependencies and naturally tenant-isolated channels, while preserving the existing in-process behavior for single-replica self-hosters

No existing issue covers this directly — problem described in-PR:

- When a Paperclip deployment scales to `replicas: 2+`, live events emitted on one replica never reach WebSocket clients connected to other replicas, so operator UIs silently drift depending on which replica the client landed on.
- The prior attempt (#3736, which this PR supersedes) used Redis pub/sub, but that pinned every self-hoster to a new Redis dependency and broadcast every company's events on a single shared channel — a metadata-leak concern.
- Paperclip already runs Postgres, so a LISTEN/NOTIFY transport delivers cross-replica fan-out with zero new infrastructure and naturally per-company channels; Redis remains an opt-in alternative.

## What Changed

- **`server/src/services/live-events/transport.ts`** — shared `LiveEventsTransport` interface and `buildEnvelope` helper with a per-transport `maxEnvelopeBytes` size cap (PG 7.5KB, Redis 1MB).
- **`server/src/services/live-events/pg-transport.ts`** — Postgres LISTEN/NOTIFY transport. Per-company channels (`paperclip_live_evt_<sha256(companyId).slice(0,24)>` — SHA fits PG's 63-byte channel-name cap; tenant isolation comes from the WS upgrade gate, not from name obscurity). Refcounted LISTEN, `originId` self-echo filter, reconnect re-issues LISTEN.
- **`server/src/services/live-events/redis-transport.ts`** — Redis pub/sub transport, opt-in via `PAPERCLIP_LIVE_EVENTS_TRANSPORT=redis`. `ioredis` is a dynamic optional dep. Per-company channels (`paperclip:live-events:<companyId>`).
- **`server/src/services/live-events.ts`** — service-layer wiring. `configureLiveEventsTransport` installs the transport at startup; `subscribeCompanyLiveEvents` / `publishLiveEvent` route through both the in-process emitter and the cross-replica transport. Boot-race-safe (subscribers attaching before transport init are rebound when it resolves).
- **`server/src/__tests__/live-events-cross-replica.test.ts`** — 11 tests covering both transports, oversized-event symmetric drop, boot-race rebind, Redis size-cap, and per-company isolation. PG tests skip on macOS via the embedded-postgres support gate.
- **`server/src/index.ts`** — startup wiring; fire-and-forgets `configureLiveEventsTransport` with a fallback warn on failure.

Configuration:

| `PAPERCLIP_LIVE_EVENTS_TRANSPORT` | Behavior |
|---|---|
| unset / `postgres` / `pg` / `default` | Postgres LISTEN/NOTIFY (default) |
| `redis` | Redis pub/sub |
| `off` / `none` / `in-process` | In-process only |

## Verification

- `pnpm tsc --noEmit` (server) — clean.
- `npx vitest run live-events-cross-replica` — 4 passed locally (Redis-mock + in-process + transport=off), 7 skipped on Darwin (embedded-postgres). PG paths run on Ubuntu CI.
- Greptile rounds 1+ addressed: boot-race, Redis size cap, hot-reload rebind, oversized-event symmetric drop, multibyte byte-length check.

## Risks

- **At-most-once delivery, not at-least-once.** Postgres LISTEN/NOTIFY does not buffer across transient disconnects. Acceptable for UI live-update hints, **not** for state-machine signals. Documented in code.
- **Oversized events dropped symmetrically** — events >7.5KB (PG) or >1MB (Redis) get logged at `error` and dropped on every replica including the originator, to avoid silent cross-replica divergence.
- **Redis transport's behaviour against real Redis is unverified** by this PR (tests use an in-process mock). The Postgres transport is exercised against a real embedded Postgres on Ubuntu CI.
- **Single-replica self-hosters: zero risk.** In-process emit still works regardless of transport setting.

## Model Used

Claude Opus 4.7 (1M context), extended thinking mode.

### Hardening (post-review research pass)

- **NOTIFY coalescing** — bursts batch into one frame per company per 25ms window; Postgres serializes NOTIFY-issuing commits behind a global `AccessExclusiveLock`, so per-event NOTIFY does not scale on write-heavy instances.
- **Resync markers instead of symmetric drops** — oversized events now reach remote replicas as `payload: { __resync: true }` markers of the same type (the originating replica keeps the full event locally); LISTEN reconnects deliver a synthetic `transport.resync` per company since LISTEN/NOTIFY is at-most-once.
- **Receiver-side tenant filter + total envelope decoding** — inbound frames are validated and filtered to the channel's company; malformed or legacy-format frames can no longer kill the LISTEN connection (rolling-deploy safe).
- **Resubscribe race fix** — unsubscribe→resubscribe within one LISTEN round-trip no longer leaks a duplicate LISTEN registration.
- **Observability** — `/api/health` (full-details) exposes the transport mode and `pg_notification_queue_usage()`; warn-logged above 0.5, rate-limited to once per minute.
- **Docs** — `doc/live-events.md`: delivery semantics, PgBouncer transaction-pooling incompatibility (LISTEN needs a direct/session-mode connection), hot-standby note, monitoring.
- Known follow-up: Redis subscriber auto-resubscribe has the same lost-window as LISTEN reconnect but does not yet synthesize `transport.resync`; tracked for a follow-up since the Redis transport is opt-in.

## Checklist

- [x] Model used specified
- [x] Checked ROADMAP.md — not in conflict with planned core work
- [x] Tests run locally and pass (Darwin runs the Redis + in-process paths; CI runs the PG path)
- [x] Added boundary tests: UTF-8 byte-length, boot-race rebind, Redis ≠ PG size cap, oversized symmetric drop
- [x] No UI changes
- [x] Documented risks above

Closes #3736.


stubbi:
@greptileai re-review please — round-1 fixes pushed in 94e0a128.

1. **Boot race (P1)** → `attachTransportFor` now always bumps `transportRefcounts` regardless of transport state; `rebindExistingSubscriptions` lazily creates handlers on transport install. Subscribers that attached during the boot window are wired up to LISTEN/NOTIFY once the transport resolves. New regression test "rebinds subscribers that attached before configureLiveEventsTransport" exercises this end-to-end.

2. **Redis size cap (P1)** → Threaded per-transport `maxEnvelopeBytes` through `LiveEventsTransport`. PG keeps the 7500-byte ceiling; Redis gets `REDIS_PUBSUB_INLINE_LIMIT = 1_000_000` (well below the 32MB pub/sub buffer cap). `shouldSuppressLocalEmit` now consults `transport.maxEnvelopeBytes` so local suppression stays symmetric with the active transport. New regression test verifies a 50KB payload round-trips on Redis.

3. **Teardown clears refcounts (P2)** → `teardownLiveEventsTransport` no longer wipes `transportRefcounts` — those track local-subscriber state and must survive a transport hot-reload. Only `transportHandlers` (tied to the old transport instance) gets cleared, so `rebindExistingSubscriptions` has the right state to wire the new transport.

PR description template gap not addressed in this round — happy to fill it in if it's blocking.

stubbi:
@greptileai re-review please — rebased onto master; new server dep replaced with @paperclipai/db re-export so frozen install passes.

stubbi:
@greptileai re-review please — rebased onto current master and addressed the open finding: `handleMessage` in the redis transport now guards `envelope.event` before reading `companyId`, so valid-JSON envelopes without an `event` field are dropped instead of throwing a TypeError out of the ioredis "message" callback. Added a mocked-redis regression test injecting exactly that payload and asserting the subscription survives and later well-formed events still arrive. `live-events-cross-replica` 12/12 locally.

stubbi:
@greptileai re-review please — hardened the transport: NOTIFY coalescing (global commit-lock mitigation), resync markers replacing symmetric oversize drops, receiver-side tenant filtering + total envelope decoding (rolling-deploy safe), reconnect resync, a resubscribe-race fix, and pg_notification_queue_usage in /api/health.

stubbi:
Addressed all three findings in ae6601a65: local transport capture in `getLiveEventsTransportHealth` (TOCTOU), Redis global channel renamed to `paperclip:live-events:$global` (collision with a literal "global" company id), and the duplicate-SUBSCRIBE tolerance note in `flushPendingSubscribes`. Thanks for the 5/5!

stubbi:

PR #5876, opened 2026-05-13
fix(server): don't re-stamp execution_run_id to a stale-assignee legacy run

ensky:
## Thinking Path

> - The heartbeat / execution-run subsystem owns the single-writer guarantee for an issue: at most one heartbeat run "owns" an issue at a time, and other wakes queue up as `deferred_issue_execution` rows behind that owner.
> - When ownership changes (mid-run reassignment), `enqueueWakeup` must let the *new* assignee become the owner. Today, the "legacyRun fallback" in `server/src/services/heartbeat.ts` looks up any queued/running heartbeat run tagged with the same `issueId` and re-stamps `issues.execution_run_id` onto that run regardless of who its `agentId` is — so the old assignee's run becomes owner again and the new assignee's wake gets queued behind it.
> - Because the early-return in `releaseIssueExecutionAndPromote` only promotes a deferred wake when the releasing run *is* the current owner (`issue.executionRunId === run.id`), the deferred wake never gets promoted once the old run later fails / is cancelled. The issue is left wedged with no live run.
> - This pull request narrows the legacyRun fallback to runs whose `agentId` still matches `issue.assigneeAgentId`, so cross-assignee reassignments fall through to the normal "new owner" path while same-agent recovery (e.g. retried run for the same assignee) keeps working.
> - The benefit is no more stuck-in-`in_progress` / `in_review` / `todo` issues after reassignment, without changing recovery semantics for the common case.

## What Changed

- `server/src/services/heartbeat.ts` — in the `enqueueWakeup` legacyRun fallback, skip re-stamping `issues.execution_run_id` to the legacy run when `legacyRun.agentId !== issue.assigneeAgentId`. Fall through so the new assignee's wake takes the normal "claim ownership" path.
- `server/src/__tests__/heartbeat-retry-scheduling.test.ts` — adds two regression tests:
  1. Mid-run reassignment: a new assignee's wake is **not** deferred behind the previous assignee's still-running run.
  2. Same-agent recovery: when the legacy run's agent still matches `issue.assigneeAgentId`, the legacyRun fallback continues to re-stamp (unchanged behavior).

## Verification

- `cd server && pnpm test heartbeat-retry-scheduling` — both new test cases pass alongside the existing ones in the file.
- Manual trace: with the patch applied, walked through `enqueueWakeup` for the reassignment scenario in the diagnostic notes; the new assignee's wake now reaches the `issue.execution_run_id IS NULL` insertion path instead of being routed to `deferred_issue_execution`. Same-agent retry path is unchanged.
- No schema / migration changes; the fix is a guard on an existing branch.

## Risks

- **Low risk, scoped guard.** The change only narrows which legacy runs are eligible for the fallback; it does not introduce a new write path.
- **Recovery for same-agent retries** (e.g. crashed run re-entry) is preserved — the `agentId` equality check is exactly the condition that distinguishes "same owner reconnecting" from "ownership has changed."
- **No-op for issues that were never reassigned.** All existing heartbeat-retry tests continue to pass.
- Edge case considered: if `issue.assigneeAgentId` is null (issue assigned to a human user only), `legacyRun.agentId` will also fail the equality and the fallback is skipped. That matches the intended semantics — there is no agent run to recover for.

## Model Used

- Claude (Anthropic) — `claude-opus-4-7`, extended-thinking mode, tool-use enabled (read/edit/grep over the repo, `pnpm test` execution).

### DIFF
--- a/server/src/__tests__/heartbeat-retry-scheduling.test.ts
+++ b/server/src/__tests__/heartbeat-retry-scheduling.test.ts
@@ -1337,4 +1337,233 @@ describeEmbeddedPostgres("heartbeat bounded retry scheduling", () => {
       retryNotBefore.toISOString(),
     );
   });
+
+  it("does not re-stamp execution_run_id to a stale-assignee legacy run during mid-run reassignment", async () => {
+    const companyId = randomUUID();
+    const oldAgentId = randomUUID();
+    const newAgentId = randomUUID();
+    const issueId = randomUUID();
+    const legacyRunId = randomUUID();
+    const now = new Date("2026-05-13T03:00:00.000Z");
+
+    await db.insert(companies).values({
+      id: companyId,
+      name: "Paperclip",
+      issuePrefix: `T${companyId.replace(/-/g, "").slice(0, 6).toUpperCase()}`,
+      requireBoardApprovalForNewAgents: false,
+    });
+
+    await db.insert(agents).values([
+      {
+        id: oldAgentId,
+        companyId,
+        name: "ClaudeCoder",
+        role: "engineer",
+        status: "active",
+        adapterType: "claude_local",
+        adapterConfig: {},
+        runtimeConfig: {
+          heartbeat: {
+            wakeOnDemand: true,
+            maxConcurrentRuns: 1,
+          },
+        },
+        permissions: {},
+      },
+      {
+        id: newAgentId,
+        companyId,
+        name: "CodexCoder",
+        role: "engineer",
+        status: "active",
+        adapterType: "codex_local",
+        adapterConfig: {},
+        runtimeConfig: {
+          heartbeat: {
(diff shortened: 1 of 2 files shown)


ensky:
Thanks for the triage @paperclip-ai — PR body has been rewritten to follow the full template (Thinking Path, What Changed, Verification, Risks, Model Used). Ready for re-review.

ensky:
Closing in favour of #10495.

This branch name contains an internal ticket id, which CONTRIBUTING.md does not allow. I cannot rename it without closing this pull request, so I opened #10495 from a compliant branch instead. I also corrected the tooling that generated the old name.

#10495 has the same fix. It is rebased onto current master, and its two regression tests are updated for the responsible-user precondition that dispatch now applies.

PR #5882, opened 2026-05-13
Add run-log rotation policy + ENOSPC-safe append

imvimm:
## Summary

Adds a retention job for Paperclip runner logs (`data/run-logs/<companyId>/<agentId>/<runId>.ndjson`) plus crash-safety on the append path. Companion to the database-backup rotation work (HFT-163) — runner logs were the top SSD-growth offender on the board's machine (~3.8 GB/day on HFT workload).

Behavior, all tunable via env, with defaults that match the ticket:

- `PAPERCLIP_RUN_LOG_UNCOMPRESSED_DAYS` (default `3`) — files older than this get gzipped in place.
- `PAPERCLIP_RUN_LOG_COMPRESSED_DAYS` (default `7`) — `.ndjson` or `.ndjson.gz` past this age is hard-deleted.
- `PAPERCLIP_RUN_LOG_MAX_FILE_BYTES` (default `500 * 1024 * 1024`) — files at or above this size get gzipped on the next tick regardless of age (the "rotate at 500 MB or daily, whichever comes first" rule).
- `PAPERCLIP_RUN_LOG_ROTATE_GRACE_MINUTES` (default `60`) — files modified within this window are skipped to avoid racing with active run writers.
- `PAPERCLIP_RUN_LOG_ROTATE_INTERVAL_MINUTES` (default `360`) — how often the rotation job runs in-process. First tick is staggered 30 s after `server.listen`.

On each tick the operator log gets a one-line summary:

```text
runner-logs: rotated X bytes, gz'd Y files, deleted Z files
```

`run-log-store.append()` now catches `ENOSPC` / `EROFS` / `EDQUOT` / `EIO`, logs a structured warning, and returns `0` so the heartbeat keeps running on a full disk. Mirrors the crash-safety addition from HFT-163. The rotator itself uses the same swallow-and-report-per-file approach so a partial disk failure during retention can't kill the server either.

## Thinking Path

- Keep the change inside the Paperclip server runner-log layer; HFT trading logs remain out of scope.
- Prefer in-process retention with conservative grace windows over a separate daemon because Paperclip already owns the run-log path and can apply the policy consistently on startup and interval ticks.
- Treat disk pressure and vanished-file races as per-file recoverable events: log and continue so retention never becomes a server crash path.
- Keep crash leftovers bounded by recognizing stale `.ndjson.gz.tmp` files and deleting them after the grace window.

## Risks

- Active-file race risk is mitigated by `PAPERCLIP_RUN_LOG_ROTATE_GRACE_MINUTES` and by only touching files whose mtime is outside that window.
- Bad env values now fall back to defaults when non-finite, negative, or zero, so an operator cannot accidentally set a zero-day hard-delete policy.
- Compression consumes CPU and IO during retention ticks; the interval and thresholds are tunable, and the large-fixture e2e measured the 600 MiB synthetic case.
- Quota/full-disk errors can still prevent compression or deletion of a specific file, but they are reported per-file and do not take down the server.

## Test plan

- [x] `pnpm exec vitest run server/src/__tests__/run-log-retention.test.ts` — 14 passed, including regressions for zero env values, `EDQUOT` recoverability, and stale `.gz.tmp` cleanup.
- [x] `pnpm --filter @paperclipai/server typecheck`
- [x] `PAPERCLIP_TEST_LARGE_FIXTURES=1 vitest run src/__tests__/run-log-retention-e2e-large.test.ts` — synthetic 600 MiB `.ndjson` collapsed to ~600 KB `.gz` in ~1.4 s; one-line summary verified in the original implementation run.
- [x] `vitest run src/__tests__/heartbeat-run-log.test.ts` — existing 2 tests still pass in the original implementation run.

## Model Used

- Initial implementation: Claude Code local adapter.
- Review follow-up: Codex local adapter.

Refs: HFT-167.

### DIFF
--- a/server/src/__tests__/run-log-retention-e2e-large.test.ts
+++ b/server/src/__tests__/run-log-retention-e2e-large.test.ts
@@ -1,0 +1,66 @@
+import { promises as fs } from "node:fs";
+import path from "node:path";
+import os from "node:os";
+import { afterAll, describe, expect, it } from "vitest";
+import { rotateRunLogs, formatRunLogRetentionSummary } from "../services/run-log-retention.js";
+
+// Opt-in: this test writes a real 600 MB file and gzips it. Gated on
+// PAPERCLIP_TEST_LARGE_FIXTURES=1 to keep CI from allocating that much disk
+// on every run. The unit tests in run-log-retention.test.ts cover the same
+// code paths with small fixtures; this test verifies the HFT-167 acceptance
+// criterion ("retention verified locally against a synthetic >500 MB log").
+const RUN_LARGE = process.env.PAPERCLIP_TEST_LARGE_FIXTURES === "1";
+const maybeIt = RUN_LARGE ? it : it.skip;
+
+let workDir: string;
+
+afterAll(async () => {
+  if (workDir) await fs.rm(workDir, { recursive: true, force: true });
+});
+
+describe("rotateRunLogs synthetic >500 MB e2e", () => {
+  maybeIt("gzips a 600 MB run-log past size threshold and reports a one-line summary", async () => {
+    workDir = await fs.mkdtemp(path.join(os.tmpdir(), "hft167-"));
+    const agentDir = path.join(workDir, "co-1", "agent-1");
+    await fs.mkdir(agentDir, { recursive: true });
+    const target = path.join(agentDir, "run-big.ndjson");
+
+    const sizeBytes = 600 * 1024 * 1024;
+    const chunk = Buffer.alloc(4 * 1024 * 1024, "a"); // 4 MiB filler
+    const handle = await fs.open(target, "w");
+    let written = 0;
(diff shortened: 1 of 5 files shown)


imvimm:
Review follow-up pushed in a539963b. Addressed the Greptile items: zero-valued retention env vars now fall back to defaults, EDQUOT is recoverable in retention, and stale .ndjson.gz.tmp files are cleaned after the grace window. Also updated the PR body with Thinking Path, Risks, and Model Used sections. Verified with pnpm exec vitest run server/src/__tests__/run-log-retention.test.ts and pnpm --filter @paperclipai/server typecheck.

imvimm:
Board confirmation accepted and all checks are green, but the merge attempt failed with: `GraphQL: imvimm does not have the correct permissions to execute MergePullRequest`. PR remains ready for a Paperclip repo maintainer with merge permission to merge.

PR #5883, opened 2026-05-13
fix(server): stringify Date bounds in enrichCommentsWithDerivedAgentAttribution

euni-euni:
## Summary

`GET /api/issues/:id/comments` (and the single-comment route) returns **500** whenever the target issue has at least one user-authored comment with no `createdByRunId`. The error is:

```
TypeError [ERR_INVALID_ARG_TYPE]: The "string" argument must be of type string or an instance of Buffer or ArrayBuffer. Received an instance of Date
  at Buffer.byteLength
  at reset.str (postgres/src/bytes.js:22)
  at Bind (postgres/src/connection.js:954)
```

Root cause: `enrichCommentsWithDerivedAgentAttribution` in `server/src/services/issues.ts` binds two raw `Date` objects into a `sql\`...\`` template literal for the `heartbeatRuns` time-window predicates:

```ts
sql`coalesce(${heartbeatRuns.finishedAt}, ${heartbeatRuns.createdAt}) >= ${new Date(minCommentCreatedAtMs)}`,
sql`coalesce(${heartbeatRuns.startedAt},  ${heartbeatRuns.createdAt}) <= ${new Date(maxCommentCreatedAtMs + SLACK_MS)}`,
```

`postgres-js` does not coerce raw `Date` objects to text in its prepared-statement bind path, so it throws on `Buffer.byteLength`. Drizzle's typed operators (`gte`, `lte`) handle the conversion, but a free `sql\`\`` parameter must already be a primitive.

Fix: stringify both bounds with `.toISOString()` so they bind as text. The DB-side `coalesce(...) >= timestamptz` comparison still works correctly because Postgres parses ISO-8601 timestamps for both `timestamp` and `timestamptz` columns.

## Repro

1. Create an issue.
2. Post a user-authored comment (no `createdByRunId`).
3. `curl /api/issues/:id/comments` → 500.

The bug only surfaces against a real Postgres driver — mocked DBs miss the prepared-statement bind path.

## Test plan

Added `server/src/__tests__/issue-comments-derived-attribution.test.ts` (embedded-postgres harness) covering all three derivation branches:

- a) agent-authored comment (`authorAgentId != null`)
- b) **user-authored comment with no `createdByRunId`** — the branch that triggers the bug; must return without throwing
- c) run-log-based derivation (`createdByRunId != null`)

Verified locally:

- [x] Test **fails on `master`** with `TypeError: ... Received an instance of Date` at `enrichCommentsWithDerivedAgentAttribution → drizzle PostgresJsPreparedQuery.queryWithCache`.
- [x] Test **passes** with this fix.
- [x] `pnpm --filter @paperclipai/server typecheck` passes.

## Other call sites

Sanity-grepped for `${new Date(` inside `sql\`` template literals across the repo (`server/`, `packages/`, `cli/`). Only the two lines in `enrichCommentsWithDerivedAgentAttribution` matched; all other `${new Date(` usages are in plain JS template strings, log messages, file paths, or UI rendering — none flow through the postgres-js bind path.

## Notes

Opening from a fork (`euni-euni/paperclip`); no push access to `paperclipai/paperclip`.

### DIFF
--- a/server/src/__tests__/issue-comments-derived-attribution.test.ts
+++ b/server/src/__tests__/issue-comments-derived-attribution.test.ts
@@ -1,0 +1,218 @@
+import { randomUUID } from "node:crypto";
+import { afterAll, afterEach, beforeAll, describe, expect, it } from "vitest";
+import {
+  agents,
+  agentWakeupRequests,
+  companies,
+  createDb,
+  heartbeatRuns,
+  issueComments,
+  issues,
+} from "@paperclipai/db";
+import {
+  getEmbeddedPostgresTestSupport,
+  startEmbeddedPostgresTestDatabase,
+} from "./helpers/embedded-postgres.js";
+import { issueService } from "../services/issues.js";
+
+// Regression test for `enrichCommentsWithDerivedAgentAttribution`:
+// the heartbeat-run time-window predicates used to bind raw `Date` objects
+// into a postgres-js prepared statement, which threw
+// `TypeError [ERR_INVALID_ARG_TYPE] ... Received an instance of Date`
+// inside `Buffer.byteLength` and surfaced as a 500 on
+// `GET /api/issues/:id/comments` whenever the issue had a user-authored
+// comment with no `createdByRunId`. The fix stringifies the bounds with
+// `toISOString()` so they bind as text parameters.
+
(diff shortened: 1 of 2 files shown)


devinfoley:

Thanks for the contribution.

PR #4374, opened 2026-04-24
fix: harden legacy issue-list encoding handling

serenakeyitan:
## Thinking Path

> - Paperclip orchestrates AI agents for zero-human companies, so core task surfaces like the issues list have to stay readable even when an older install carries forward messy historical state.
> - This bug sits in the database/client boundary for the issues list path, where Paperclip reads issue rows out of PostgreSQL and returns a trimmed preview to the board.
> - Issue `#3893` reports `22021 / invalid byte sequence for encoding UTF8` on `GET /api/companies/:id/issues` after upgrading an existing embedded database.
> - The failure mode points to legacy non-UTF-8 bytes surviving in older databases, while the current server path needs to keep reading safely under a UTF-8 client connection.
> - This pull request hardens the DB client encoding settings and adds a regression test that creates a migrated `SQL_ASCII` database, inserts a legacy bad description byte, and proves `issueService.list()` still returns a usable row.
> - The benefit is a stable issues list for upgraded installs with legacy encoding history, plus direct coverage for the exact compatibility path that regressed in the issue.

## What Changed

- Forces `postgres` utility and Drizzle clients to connect with `client_encoding = UTF8` in `packages/db/src/client.ts`.
- Adds a DB client regression test in `packages/db/src/client.test.ts` that asserts Drizzle sessions negotiate `UTF8`.
- Adds `postgres` as a server dev dependency for the encoding regression harness.
- Adds a server regression test in `server/src/__tests__/issues-service.test.ts` that creates a migrated `SQL_ASCII` database, inserts a legacy invalid description byte, and verifies `issueService.list()` returns a replacement-character preview instead of throwing.

## Verification

- `pnpm exec vitest run packages/db/src/client.test.ts server/src/__tests__/issues-service.test.ts`
- `pnpm -r typecheck`
- `pnpm build`
- `pnpm test:run` currently still has unrelated existing failures in `costs-service.test.ts`, `heartbeat-process-recovery.test.ts`, `workspace-runtime.test.ts`, `network-bind.test.ts`, and `onboard.test.ts`

## Risks

- Low risk: runtime code change is limited to explicitly pinning PostgreSQL client encoding to UTF-8.
- Test risk is moderate only in that the new regression harness creates and migrates a dedicated `SQL_ASCII` database during the suite, which adds setup work but keeps the repro faithful to the reported upgrade path.

[...]

PR #5902, opened 2026-05-13
fix(server): guard session-compaction projection against SQL_ASCII clusters

glide-the:
## Thinking Path

> - Paperclip orchestrates AI agents for zero-human companies and persists each heartbeat run's narrative summary into `heartbeat_runs.result_json` (jsonb) so the board UI and downstream services can attribute work to specific runs.
> - The heartbeat service already knows that PG18's `initdb --locale=C` quietly downgrades the cluster's `server_encoding` to `SQL_ASCII` even when `--encoding=UTF8` is requested, so it ships a `hasUnsafeTextProjectionDatabase()` probe and a `heartbeatRunSqlAsciiSafeColumns` fallback projection used by `getRun()` and `list()`.
> - `evaluateSessionCompactionPolicy` (`server/src/services/heartbeat.ts:3241`) reads the same `result_json ->> 'summary' / 'result' / 'message' / 'error'` text projections via `heartbeatRunListResultColumns`, but skips that SQL_ASCII guard.
> - On SQL_ASCII clusters PG's `left(text, 500)` is byte-level rather than character-level, so the 500-byte cut routinely lands inside a 3-byte CJK UTF-8 sequence and produces a partial sequence the wire layer cannot re-encode for `client_encoding=UTF8` clients (postgres-js, Navicat). PG raises `PostgresError: invalid byte sequence for encoding "UTF8": 0xe5 0x86`, which fails `evaluateSessionCompactionPolicy → executeRun` and surfaces as a 500 on `GET /api/issues/:id/comments` because the comment-attribution path re-queries the same heartbeat run.
> - This pull request extends the existing SQL_ASCII safeguard to the one missed call site so the cluster-encoding mismatch no longer 500s the comments API or breaks heartbeat setup.
> - The benefit is fork operators who hit the PG18 + `--locale=C` SQL_ASCII trap can keep using the comments view and the heartbeat session-compaction loop without an emergency cluster rebuild.

## What Changed

- Added `heartbeatRunListResultSqlAsciiSafeColumns` next to `heartbeatRunListResultColumns` in `server/src/services/heartbeat.ts`. The four `left()`-truncated narrative columns (`resultSummary / resultResult / resultMessage / resultError`) are projected as `NULL`; the digits-only cost columns (`resultTotalCostUsd / resultCostUsd / resultCostUsdCamel`) are kept because they cannot trigger the byte-truncation hazard.
- In `evaluateSessionCompactionPolicy`, added an `await hasUnsafeTextProjectionDatabase()` gate that selects either the safe or the original projection set; the `error` column is also `NULL`-projected on this code path because it is plain text and may contain CJK content that the wire encoder would reject for the same reason.
- No schema change, no migration, no API contract change. Behavior is unchanged for `server_encoding=UTF8` clusters.

## Verification

- Manual repro on a SQL_ASCII embedded-postgres cluster (PG18 + `--encoding=UTF8 --locale=C`, which silently produces a SQL_ASCII cluster):
  - **Before:** `GET /api/issues/<ID>/comments?order=desc&limit=50` returns 500 with `caused by: PostgresError: invalid byte sequence for encoding "UTF8": 0xe5 0x86` originating at `evaluateSessionCompaction (server/src/services/heartbeat.ts:3241)`.
  - **After:** the same request returns the comment list. Heartbeat session-compaction continues to evaluate run count / raw-token / session-age thresholds (none of which depend on the dropped narrative text fields).
- Sanity check that the bug is purely a PG-side byte cut, not corrupted storage:
  ```sql
  SET client_encoding = 'UTF8';
  -- Round-trip of the full string is fine; data is valid UTF-8 in storage.
  SELECT result_json ->> 'summary' FROM heartbeat_runs WHERE id = '<run-with-cjk>';
  -- left(text, N) raises invalid byte sequence whenever N lands mid 3-byte CJK char:
  SELECT left(result_json ->> 'summary', 500) FROM heartbeat_runs WHERE id = '<run-with-cjk>';
  ```
- `cluster server_encoding` can be inspected with:
  ```sql
  SHOW server_encoding;     -- SQL_ASCII on impacted clusters
  SHOW client_encoding;     -- UTF8 (postgres-js / Navicat)
  ```

## Risks

- **Low.** The change mirrors the existing SQL_ASCII safeguard already shipped for `getRun()` (`server/src/services/heartbeat.ts:2402`) and `list()` (`server/src/services/heartbeat.ts:9461`), gated by the same `hasUnsafeTextProjectionDatabase()` probe. UTF-8 clusters see no behavior change.
- **Cost on SQL_ASCII clusters:** the session-compaction code path no longer sees the four narrative `result_json` text fields. `evaluateSessionCompactionPolicy` only consumes `runs.length`, `latestRawUsage`, and `sessionAgeHours` for the rotation decision, so rotation behavior is unaffected. The `latestSummary` derived from the dropped fields will become `null`; downstream consumers already tolerate `null` (existing behavior in the `list()` SQL_ASCII branch).
- No schema change, no migration, no API contract change.
- Out of scope (recommended follow-ups): (1) flip `initdbFlags` in `server/src/index.ts:388` from `--locale=C` to `--locale=C.UTF-8` to stop new clusters from being created as SQL_ASCII on PG18; (2) emit a startup banner / pino warning when an existing cluster is detected as SQL_ASCII, with a `pg_dumpall` → rebuild → restore migration runbook; (3) add a plpgsql `safe_left_utf8(text, int)` so SQL_ASCII clusters can keep returning the four narrative fields instead of dropping them; (4) sanitize child-process stdout/stderr with `StringDecoder` in `packages/adapter-utils/src/server-utils.ts:2027` so partial UTF-8 bytes never reach jsonb in the first place.

## Model Used

- Cursor Agent (Anthropic Claude Opus 4.7), extended-thinking + tool-use mode, default Cursor IDE context window. Used for root-cause diagnosis (bisecting `left(text, n)` byte-cut behavior on a SQL_ASCII cluster), patch authoring, and PR description drafting. All edits were reviewed in the IDE before commit.

## Checklist

### DIFF
--- a/server/src/services/heartbeat.ts
+++ b/server/src/services/heartbeat.ts
@@ -763,6 +763,21 @@ const heartbeatRunListResultColumns = {
   resultCostUsdCamel: sql<string | null>`${heartbeatRuns.resultJson} ->> 'costUsd'`.as("resultCostUsdCamel"),
 } as const;
 
+// SQL_ASCII clusters do byte-level (not char-level) `left()` truncation, which can split a
+// multi-byte UTF-8 character mid-sequence and produce bytes that the wire protocol cannot
+// re-encode as UTF-8 (PG raises `invalid byte sequence for encoding "UTF8": 0xe5 0x86`).
+// On those clusters we drop the four narrative text projections; numeric cost fields are
+// digits-only and never trigger the truncation hazard, so they're safe to keep.
+const heartbeatRunListResultSqlAsciiSafeColumns = {
+  resultSummary: sql<string | null>`NULL`.as("resultSummary"),
+  resultResult: sql<string | null>`NULL`.as("resultResult"),
+  resultMessage: sql<string | null>`NULL`.as("resultMessage"),
+  resultError: sql<string | null>`NULL`.as("resultError"),
+  resultTotalCostUsd: sql<string | null>`${heartbeatRuns.resultJson} ->> 'total_cost_usd'`.as("resultTotalCostUsd"),
+  resultCostUsd: sql<string | null>`${heartbeatRuns.resultJson} ->> 'cost_usd'`.as("resultCostUsd"),
+  resultCostUsdCamel: sql<string | null>`${heartbeatRuns.resultJson} ->> 'costUsd'`.as("resultCostUsdCamel"),
+} as const;
+
 const heartbeatRunSafeResultJsonColumn = sql<Record<string, unknown> | null>`
   case
     when ${heartbeatRuns.resultJson} is null then null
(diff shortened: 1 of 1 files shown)


glide-the:
Hi @serenakeyitan — would you mind taking a look? This PR extends the `hasUnsafeTextProjectionDatabase()` SQL_ASCII safeguard you introduced in #4374 to `evaluateSessionCompactionPolicy` (`server/src/services/heartbeat.ts:3241`), which still selects the raw `heartbeatRunListResultColumns` projection and currently 500s `GET /api/issues/:id/comments` on PG18 + `--locale=C` clusters with CJK content (`left(text, 500)` is byte-level there and splits a 3-byte UTF-8 char). It mirrors your existing pattern one-for-one and is a no-op on UTF-8 clusters. Thanks!

PR #5903, opened 2026-05-13
fix(server): cast Date params to timestamptz in issue comment run lookup

glide-the:
## Thinking Path

> - Paperclip orchestrates AI agents for zero-human companies and exposes a board API that lists comments per issue, with a run-attribution post-processing pass.
> - `enrichCommentsWithDerivedAgentAttribution` (added in #5780, `server/src/services/issues.ts:1946`) joins recent `heartbeat_runs` rows whose lifetime overlaps the candidate comments to back-fill agent attribution for human-authored comments produced by an agent run.
> - The lookup embeds `${new Date(...)}` directly inside drizzle's raw `sql` template — a `coalesce(...)` expression rather than a typed `gte(column, date)` predicate — so drizzle forwards the value as an untyped `Param` and `postgres@3.4.8` calls `Buffer.byteLength` on the Date during Bind, which throws `TypeError [ERR_INVALID_ARG_TYPE]: ... Received an instance of Date` and breaks every fetch with `DrizzleQueryError`.
> - Other time-window predicates in the codebase use drizzle's typed `gte(column, date)` / `lte(column, date)` helpers (which carry column type info), so only this one site triggers the bug.
> - This pull request converts those Date bindings to ISO strings with an explicit `::timestamptz` cast so PostgreSQL parses the timestamp server-side and the postgres-js driver no longer receives an unsupported Date payload in a typed-less SQL fragment.
> - The benefit is `GET /api/issues/:id/comments` stops 500-ing on issues whose comments trigger the run-log derivation path, restoring the activity feed without changing query semantics.

## What Changed

- In `server/src/services/issues.ts` (`enrichCommentsWithDerivedAgentAttribution`), bind the `coalesce(finished_at, created_at) >= ?` and `coalesce(started_at, created_at) <= ?` parameters as `${date.toISOString()}::timestamptz` instead of raw `Date` objects, so the `postgres` driver no longer receives an unsupported `Date` payload in a typed-less `sql` fragment.

## Verification

- Manual repro before the fix: `curl 'http://localhost:3100/api/issues/<ID>/comments?order=desc&limit=50'` returns 500 with `DrizzleQueryError ... TypeError [ERR_INVALID_ARG_TYPE]: The "string" argument must be of type string or an instance of Buffer or ArrayBuffer. Received an instance of Date`. After the fix the same request returns the comment list and run-log–derived attribution still populates for comments matched against a heartbeat run.
- Reviewer can also inspect the generated SQL: parameters $5/$6 are now ISO-8601 strings with a `::timestamptz` cast in the `WHERE` clause, instead of opaque Date placeholders.

## Risks

- Low. The change is binding-shape only — same column references, same `coalesce` semantics, same comparison operators. PostgreSQL coerces ISO-8601 strings cast to `timestamptz` identically to the previous Date binding (when it worked), and the explicit cast removes the driver-level type-inference dependency.
- No schema change, no migration, no API contract change.

## Model Used

- Cursor Agent (Anthropic Claude Opus 4.7), extended-thinking + tool-use mode, default Cursor IDE context window. Used for root-cause diagnosis (tracing the Bind-time `Buffer.byteLength` failure to drizzle's untyped sql-template parameter path), patch authoring, and PR description drafting. All edits were reviewed in the IDE before commit.

## Checklist

### DIFF
--- a/server/src/services/issues.ts
+++ b/server/src/services/issues.ts
@@ -1969,8 +1969,8 @@ export function issueService(db: Db) {
                 and ${activityLog.runId} = ${heartbeatRuns.id}
             )`,
           ),
-          sql`coalesce(${heartbeatRuns.finishedAt}, ${heartbeatRuns.createdAt}) >= ${new Date(minCommentCreatedAtMs)}`,
-          sql`coalesce(${heartbeatRuns.startedAt}, ${heartbeatRuns.createdAt}) <= ${new Date(maxCommentCreatedAtMs + ISSUE_COMMENT_RUN_LOG_DERIVATION_END_SLACK_MS)}`,
+          sql`coalesce(${heartbeatRuns.finishedAt}, ${heartbeatRuns.createdAt}) >= ${new Date(minCommentCreatedAtMs).toISOString()}::timestamptz`,
+          sql`coalesce(${heartbeatRuns.startedAt}, ${heartbeatRuns.createdAt}) <= ${new Date(maxCommentCreatedAtMs + ISSUE_COMMENT_RUN_LOG_DERIVATION_END_SLACK_MS).toISOString()}::timestamptz`,
         ),
       )
       .orderBy(desc(heartbeatRuns.createdAt));


devinfoley:

Thanks for the contribution.

PR #5925, opened 2026-05-13
fix: resolve orphan-sweep null-assignee filter regression

digitalflanker-ux:
## Summary
- fix null assignee filter regression for issues list path
- treat `assigneeAgentId=null` as explicit null-filter and return 200
- validate malformed assignee IDs as 4xx instead of 500
- preserve UUID assignee filter behavior
- add route + service regression tests

## Verification
- pnpm -C /Users/openclaw/paperclip/.worktrees/fma-1318-clean exec vitest run server/src/__tests__/issue-list-assignee-filter-routes.test.ts server/src/__tests__/issues-service.test.ts
- result: 2 files passed, 50 tests passed

## Evidence
- before: `assigneeAgentId=null` and malformed values returned 500
- after: `assigneeAgentId=null` returns 200, malformed values return 422, UUID path stays 200

### DIFF
--- a/server/src/__tests__/issue-list-assignee-filter-routes.test.ts
+++ b/server/src/__tests__/issue-list-assignee-filter-routes.test.ts
@@ -1,0 +1,195 @@
+import { randomUUID } from "node:crypto";
+import express from "express";
+import request from "supertest";
+import { afterAll, afterEach, beforeAll, describe, expect, it } from "vitest";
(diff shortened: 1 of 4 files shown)


devinfoley:
Thanks for the contribution! We rebased this onto current master and opened an operator-mergeable resubmission at #8018 (fork branch was not pushable). Closing this PR in favor of #8018.

devinfoley:
Superseded by #8018.

PR #6013, opened 2026-05-14
fix(server): auto-resolve recovery actions on terminal issue status (CONA-347)

dingus11:
## Summary

Auto-resolves active recovery actions when an issue transitions to `done` or `cancelled` via `PATCH /api/issues/:id`.

Previously, the documented *"owner heartbeat confirms done"* resolution path did not exist as an automatic trigger: `resolveActiveForIssue` was only reachable via an explicit `POST /api/issues/:id/recovery-actions/resolve`, which requires the actor to be the issue assignee. This left recovery actions stranded on done issues with no automatic clearance path — the failure mode observed in [CONA-347](/CONA/issues/CONA-347) on recovery actions for CONA-320 / CONA-321.

## Change

`server/src/routes/issues.ts` — inside the existing `becameTerminal` branch of the issue PATCH handler, call `recoveryActionsSvc.resolveActiveForIssue` with the new terminal status:

- `done` → outcome `restored`, status `resolved`
- `cancelled` → outcome `cancelled`, status `cancelled`

Errors are soft-caught and logged so a failing recovery-action update never blocks the issue PATCH itself. +17 lines, no test/runtime changes elsewhere.

## Why a separate PR (not folded into CONA-341)

[CONA-341](/CONA/issues/CONA-341) is scoped to adapter/model compat check + config-save gate + retry gate (`CONA-334` Items 1–3). This is a different subsystem (recovery-action lifecycle vs. adapter compat) and the change is small and self-contained; folding would muddy the diff for both reviewers. Filed as its own PR.

## Test plan

- [ ] `pnpm --filter @paperclipai/server typecheck` — passes locally
- [ ] Manual: PATCH an issue with an active recovery action to `status=done`, verify the recovery action's `status` becomes `resolved` and `outcome` is `restored`
- [ ] Manual: PATCH an issue with an active recovery action to `status=cancelled`, verify `status=cancelled`, `outcome=cancelled`
- [ ] Regression: PATCH an issue without an active recovery action — handler still succeeds, no error logs

## Follow-up

- Update [CONA-342](/CONA/issues/CONA-342) description: the *"owner heartbeat confirms done"* mechanism was aspirational, not implemented. With this PR, terminal-transition auto-resolution is installed in the PATCH handler instead. Will be edited once this PR merges.

### DIFF
--- a/server/src/__tests__/issue-comment-reopen-routes.test.ts
+++ b/server/src/__tests__/issue-comment-reopen-routes.test.ts
@@ -69,6 +69,7 @@ const mockIssueThreadInteractionService = vi.hoisted(() => ({
 }));
 const mockIssueRecoveryActionService = vi.hoisted(() => ({
   getActiveForIssue: vi.fn(async () => null),
+  resolveActiveForIssue: vi.fn(async () => undefined),
 }));
 const mockIssueTreeControlService = vi.hoisted(() => ({
   getActivePauseHoldGate: vi.fn(async () => null),
@@ -243,6 +244,7 @@ describe.sequential("issue comment reopen routes", () => {
     mockInstanceSettingsService.listCompanyIds.mockReset();
     mockRoutineService.syncRunStatusForIssue.mockReset();
     mockIssueRecoveryActionService.getActiveForIssue.mockReset();
+    mockIssueRecoveryActionService.resolveActiveForIssue.mockReset();
     mockIssueTreeControlService.getActivePauseHoldGate.mockReset();
     mockTxInsertValues.mockReset();
     mockTxInsert.mockReset();
@@ -280,6 +282,7 @@ describe.sequential("issue comment reopen routes", () => {
     mockInstanceSettingsService.listCompanyIds.mockResolvedValue(["company-1"]);
     mockRoutineService.syncRunStatusForIssue.mockResolvedValue(undefined);
     mockIssueRecoveryActionService.getActiveForIssue.mockResolvedValue(null);
+    mockIssueRecoveryActionService.resolveActiveForIssue.mockResolvedValue(undefined);
     mockIssueTreeControlService.getActivePauseHoldGate.mockResolvedValue(null);
     mockIssueService.addComment.mockResolvedValue({
       id: "comment-1",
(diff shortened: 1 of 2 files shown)


dingus11:
## CTO architecture review — CONA-347 / CONA-342

The fix is correct and minimal: installing `resolveActiveForIssue` on the `becameTerminal` branch of the issue PATCH handler is the right hook. Outcome mapping (done → restored, cancelled → cancelled) matches the existing service contract. The `.catch(...)` makes the recovery-action update non-blocking — a good defensive choice so a service-layer hiccup never breaks the issue PATCH itself.

### Tier-1 — must fix before merge

**1. Failing test: `server/src/__tests__/issue-comment-reopen-routes.test.ts`**

CI run [25888076294 / suites 2-4](https://github.com/paperclipai/paperclip/actions/runs/25888076294/job/76084657243) produces 3 unhandled `TypeError: recoveryActionsSvc.resolveActiveForIssue is not a function` rejections, because the test's hoisted mock at `mockIssueRecoveryActionService` (line ~70) only exposes `getActiveForIssue`. The new code path now exercises `resolveActiveForIssue` whenever a test PATCH transitions an issue to `done`/`cancelled`.

Fix: add `resolveActiveForIssue: vi.fn(async () => undefined)` to that mock, and reset it in the `beforeEach` block alongside the existing `mockIssueRecoveryActionService.getActiveForIssue.mockReset()`.

Once that's added the test should go green — the production wiring at `routes/issues.ts:841` is unchanged and the same call is already used elsewhere in the file.

### Tier-2 — file as a follow-up (non-blocking)

**2. No test asserts the new auto-resolve behavior.**

The PR ships an observable behavior change (PATCH to terminal status now mutates `recovery_actions` rows) with no test covering it. The pre-existing tests pass with the mock fix above, but a new test that mocks an active recovery action and verifies it gets resolved on `becameTerminal` would lock the behavior in. Worth adding either here or as a same-PR addendum.

### Verdict

**Changes requested.** Add `resolveActiveForIssue` to the test mock (one line), then green. Coverage for the new code path can land here or as a fast follow-up.

dingus11:
CTO review addressed:

**Tier-1**: Added `resolveActiveForIssue: vi.fn(async () => undefined)` to `mockIssueRecoveryActionService` mock + `mockReset()` and `mockResolvedValue(undefined)` in `beforeEach`. The 3 unhandled TypeError rejections are gone.

**Tier-2**: Added 3 new tests in `issue-comment-reopen-routes.test.ts`:
- `auto-resolves an active recovery action when an issue is marked done`
- `auto-resolves an active recovery action with cancelled outcome when an issue is cancelled`
- `does not call resolveActiveForIssue when status does not reach a terminal state`

All 34 tests pass (was 31). Typecheck clean. Ready for re-review.

PR #6026, opened 2026-05-15
fix: propagate projectId into wakeup context and support identifier lookup

devinfoley:
## Thinking Path

> - The server's heartbeat/wakeup pipeline resolves which project workspace an agent run should bind to
> - `enqueueWakeup` resolves an issue (and therefore a project) before scheduling a run, but the resolved `projectId` was never written back into `enrichedContextSnapshot.projectId`, so `resolveWorkspaceForRun` always saw `contextProjectId === null`
> - When the `issueProjectRef` DB lookup also returned null (e.g. identifier-style id like `ENV-13`, not a UUID), workspace resolution fell through to the `agent_home` fallback instead of the correct project workspace
> - Surfaced while running the QA matrix on sandbox/SSH — runs were ending up in the wrong workspace
> - This pull request stores the resolved `projectId` back into context and replaces the raw UUID-only DB query with `issuesSvc.getById`, which accepts both UUIDs and identifiers and canonicalizes `context.issueId` / `context.taskId` to the UUID on identifier hits
> - The benefit is that wakeups triggered with identifier-style ids correctly bind to their project workspace instead of silently degrading to `agent_home`

## What Changed

- In `enqueueWakeup`, after the issue resolves, write `projectId` back into `enrichedContextSnapshot.projectId` so downstream workspace resolution can use it.
- Replace the raw UUID-only DB query for the issue with `issuesSvc.getById`, which handles both UUIDs and identifiers (e.g. `ENV-13`).
- On an identifier hit, canonicalize `context.issueId` and `context.taskId` to the resolved UUID.

## Verification

- Trigger a wakeup with an identifier-style id (`ENV-13`) on the dev instance and confirm the run binds to the correct project workspace instead of `agent_home`.
- Confirm UUID-style wakeups still resolve to the same project workspace as before.

## Risks

- Low risk. Scope is a single function in `server/src/services/heartbeat.ts` (+20/-7). Failure mode if regressed is the prior behavior (fallback to `agent_home`).

## Model Used

- Claude (Anthropic), `claude-opus-4-7`, via Claude Code / Paperclip `claude_local` adapter.

## Checklist

### DIFF
--- a/server/src/services/heartbeat.ts
+++ b/server/src/services/heartbeat.ts
@@ -1777,7 +1777,7 @@ function enrichWakeContextSnapshot(input: {
   payload: Record<string, unknown> | null;
 }) {
   const { contextSnapshot, reason, source, triggerDetail, payload } = input;
-  const issueIdFromPayload = readNonEmptyString(payload?.["issueId"]);
+  const issueIdFromPayload = readNonEmptyString(payload?.["issueId"]) ?? readNonEmptyString(payload?.["taskId"]);
   const commentIdFromPayload = readNonEmptyString(payload?.["commentId"]);
   const taskKey = deriveTaskKey(contextSnapshot, payload);
   const wakeCommentId = deriveCommentId(contextSnapshot, payload);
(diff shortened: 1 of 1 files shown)


devinfoley:
@greptileai addressed the cross-tenant lookup in 3f5be408 — replaced `issuesSvc.getById` with an inline drizzle query that filters by `companyId` AND `(id = uuid OR identifier = upper(input))`. The DB now refuses to return another tenant's row even on identifier collisions.

devinfoley:
@greptile review

devinfoley:
@greptileai addressed the uuid cast issue in ceb2c3df — added isUuidLike guard so identifier-style inputs (e.g. `ENV-13`) skip the `eq(issues.id, …)` arm and only match by identifier. Company scoping preserved on both branches.

devinfoley:
@greptile review

PR #6070, opened 2026-05-15
[codex] Add full locale catalog

cryppadotta:
## Thinking Path

> - Paperclip orchestrates AI agents through a board-facing control plane.
> - The UI is the operator surface where onboarding and company creation flows need to remain understandable across languages.
> - The i18next foundation now supports locale resource loading and validation, but only English was present on `master`.
> - The branch exists to populate that foundation with the supported language catalog without changing routing, data contracts, or runtime behavior.
> - This pull request adds locale JSON resources for the current non-English language set.
> - The benefit is that future locale selection work has validated message catalogs ready for the first translated onboarding strings.

## What Changed

- Added 39 localized message catalogs under `ui/src/i18n/locales/` for the existing no-companies onboarding strings.
- Kept the PR rebased onto current `master` so it only contains the all-languages layer on top of the already-merged i18next foundation.

## Verification

- `pnpm exec vitest run ui/src/i18n/locale-validation.test.ts`
- Checked `ROADMAP.md`; this is scoped locale catalog follow-up work and does not duplicate a listed roadmap feature.
- No before/after screenshots included because this PR only adds resource JSON files and does not change rendered layout or visible default-English UI behavior.

## Risks

- Low risk: static JSON resource additions validated against the English reference schema.
- Translation quality may need native-speaker review before enabling visible locale switching broadly.

> For core feature work, check [`ROADMAP.md`](ROADMAP.md) first and discuss it in `#dev` before opening the PR. Feature PRs that overlap with planned core work may need to be redirected — check the roadmap first. See `CONTRIBUTING.md`.

## Model Used

- OpenAI Codex CLI, GPT-5 family coding agent, tool-enabled repository access, medium reasoning effort.

## Checklist

- [x] I have included a thinking path that traces from project context to this change
- [x] I have specified the model used (with version and capability details)
- [x] I have checked ROADMAP.md and confirmed this PR does not duplicate planned core work
- [x] I have run tests locally and they pass
- [x] I have added or updated tests where applicable
- [x] If this change affects the UI, I have included before/after screenshots
- [x] I have updated relevant documentation to reflect my changes
- [x] I have considered and documented any risks above
- [x] I will address all Greptile and reviewer comments before requesting merge

PR #6075, opened 2026-05-15
feat(i18n): expand English locale catalog with ~3K extracted UI strings

hrefcl:
## Summary
Expands the upstream i18next foundation (#5943) with a comprehensive English string catalog extracted from the full UI.

## What's included
- **~3,060 English strings** in  covering components, pages, adapters, and plugins
- **Language switcher** in  using upstream's i18next wrapper
- **Dynamic adapter translations** in 
- **Auto-sync script** () for generating additional locale files from the English source of truth
- **Updated validation tests** for the expanded catalog

## Why this PR
Upstream's i18next foundation (#5943) and the all-languages draft (#6070) currently only have the 3 onboarding strings. This PR provides the actual UI string catalog that makes localization useful.

## Architecture
-  is the single source of truth (flat nested keys)
- All strings use  interpolation compatible with i18next
- No new dependencies (uses upstream's i18next + react-i18next setup)
- Type-safe: TS union/enum literals remain as literal strings

## Testing
- ✅  passes (7/7)
- ✅  passes
- ✅ TypeScript typecheck clean

## Follow-up
- PR 2: Add ES/PT/DE/FR/JA/ZH locale files (already generated, validated)
- PR 3: Migrate remaining components/pages to use  where not yet translated

## File count
Intentionally kept small: **5 files** to stay well under the 100-file review limit.


hrefcl:
@greptile-apps thanks for the review! Fixes pushed:

1. **Language switcher** ✅ — Now dynamically renders  instead of hardcoding 7 buttons. Only loaded locales appear, so clicking always has an effect.

2. **getAdapterLabels()** ✅ — Now uses  which goes through the translation path, so adapter dropdowns update with the active locale.

3. **disabledLabel** ✅ — Uses  with the  guard.

All 3 issues from the 3/5 score should now be resolved. 🙏

PR #6091, opened 2026-05-16
fix: walk error.cause when detecting companies_issue_prefix_idx conflicts (#6088)

### LINKED ISSUE #6088: createCompanyWithUniquePrefix doesn't recognise DrizzleQueryError wrapping; any 2nd company sharing a 3-letter name prefix fails with HTTP 500
### Summary

Creating multiple companies whose names share the first three alphabetic characters (e.g. `dogdevs`, `dogcic`, `dogjobagency`) succeeds only for the first one. Every subsequent create returns **HTTP 500 `{"error":"Internal server error"}`** and the server log shows `duplicate key value violates unique constraint "companies_issue_prefix_idx"`.

The retry loop in `createCompanyWithUniquePrefix` (`server/src/services/companies.ts`) is supposed to handle exactly this case — when the auto-derived 3-letter prefix collides, it appends `A`, `AA`, `AAA`, … and retries up to 10 000 times. But the conflict detector inspects fields (`error.constraint`, `error.constraint_name`) that the underlying `postgres` driver error has been wrapped into `DrizzleQueryError`, where the actual constraint name lives at `error.cause.constraint`. The detector therefore returns `false`, the catch block rethrows, and the route emits 500.

### Reproduce

Against `paperclipai 2026.513.0` with a fresh CEO-bootstrapped instance, using any instance-admin board API key:

```bash
KEY=$(cat ~/.paperclip_admin_api_key)
BASE=http://paperclip.home.arpa

for name in dogdevs dogcic dogjobagency; do
  curl -sS -X POST "$BASE/api/companies" \
    -H "Authorization: Bearer $KEY" \
    -d "{\"name\":\"$name\"}" -w "\n  status: %{http_code}\n\n"
done
```

Output:

```
{"id":"...","name":"dogdevs","issuePrefix":"DOG",...}
  status: 201

{"error":"Internal server error"}
  status: 500

{"error":"Internal server error"}
  status: 500
```

Server log (truncated):

```
ERROR: POST /api/companies 500 — Failed query: insert into "companies" (...)
       caused by: PostgresError: duplicate key value violates unique constraint "companies_issue_prefix_idx"
       at PostgresJsPreparedQuery.queryWithCache (drizzle-orm/pg-core/session.js:41:15)
       at async createCompanyWithUniquePrefix (services/companies.js:96:30)
       at async Object.create (services/companies.js:126:29)
```

### Root cause

`server/src/services/companies.ts` (around line 127):

```ts
function isIssuePrefixConflict(error: unknown) {
  const constraint = typeof error === "object" && error !== null && "constraint" in error
    ? (error as { constraint?: string }).constraint
    : typeof error === "object" && error !== null && "constraint_name" in error
      ? (error as { constraint_name?: string }).constraint_name
      : undefined;
  return typeof error === "object"
    && error !== null
    && "code" in error
    && (error as { code?: string }).code === "23505"
    && constraint === "companies_issue_prefix_idx";
}

async function createCompanyWithUniquePrefix(data) {
  const base = deriveIssuePrefixBase(data.name);   // "dogcic" → "DOG"
  let suffix = 1;
  while (suffix < 10000) {
    const candidate = `${base}${suffixForAttempt(suffix)}`;
    try {
      const rows = await db.insert(companies).values({ ...data, issuePrefix: candidate }).returning();
      return rows[0];
    } catch (error) {
      if (!isIssuePrefixConflict(error)) throw error;   // <-- always true; conflict is rethrown
    }
    suffix += 1;
  }
  throw new Error("Unable to allocate unique issue prefix");
}
```

`drizzle-orm` (used with the `postgres-js` driver) wraps the underlying driver error in `DrizzleQueryError`. The wrapping object exposes only `name` / `message` / `cause` at the outer level — `code`, `constraint`, `constraint_name` are nested at `error.cause`:

```js
{
  name: "DrizzleQueryError",
  message: "Failed query: insert into ...",
  code: undefined,                  // <-- not on the outer error
  constraint: undefined,            // <-- not on the outer error
  cause: {                          // <-- actual postgres error is here
    name: "PostgresError",
    code: "23505",
    constraint: "companies_issue_prefix_idx",
    severity: "ERROR",
    ...
  }
}
```

So `isIssuePrefixConflict(error)` reads `code` and `constraint` on the wrapping `DrizzleQueryError` (both `undefined`) and returns `false`. The retry loop catches nothing and surfaces the original conflict as a 500.

### Proposed fix

Walk the error chain when looking for the postgres conflict signature:

```ts
function isIssuePrefixConflict(error: unknown): boolean {
  let cursor: unknown = error;
  while (cursor && typeof cursor === "object") {
    const node = cursor as { code?: string; constraint?: string; constraint_name?: string; cause?: unknown };
    const constraint = node.constraint ?? node.constraint_name;
    if (node.code === "23505" && constraint === "companies_issue_prefix_idx") {
      return true;
    }
    cursor = node.cause;
  }
  return false;
}
```

**Fix PR available**: #6091. Same pattern would benefit any other place in the codebase that inspects `error.constraint` directly — currently brittle wherever it does so.

### Environment

- `paperclipai 2026.513.0` (npm global)
- `drizzle-orm` shipping with paperclipai 2026.513.0
- Node 20.x, Postgres 17
- Debian 13 (Trixie) amd64, kernel 6.12 LXC
- Deployment mode: `authenticated`, exposure: `private`, bind: `lan`

### Workaround until fix lands

Bypass the API for the colliding cases and INSERT directly into the `companies` table with explicit unique `issue_prefix` values; mirror the API's side-effects (owner membership row):

```sql
INSERT INTO companies (name, description, issue_prefix)
VALUES ('dogcic', 'UK CIC ...', 'CIC');

INSERT INTO company_memberships
  (company_id, principal_type, principal_id, membership_role, status)
VALUES
  ('<new_company_id>', 'user', '<instance_admin_user_id>', 'owner', 'active');
```

Pick obviously-unique prefixes (e.g. `CIC`, `JOB`, `DCC`, `INF`) rather than relying on the `A`/`AA`/… auto-suffix.

---

🤖 Made with Claude Code (Opus 4.7, 1M context, high-effort thinking)

remymiguk:
## Thinking Path

> - Companies are the top-level scoping primitive; creating one is a high-frequency op via API + UI
> - Issue prefixes are auto-derived from the first three letters of the company name and must be unique per instance — `createCompanyWithUniquePrefix` exists specifically to retry with suffixed prefixes (`DOG` → `DOGA` → `DOGAA` …) on collision
> - The retry loop relies on `isIssuePrefixConflict` to recognise the Postgres unique-violation; when drizzle-orm wraps the driver error in `DrizzleQueryError`, the detector misses it (`code` / `constraint` live at `error.cause`, not on the outer error)
> - Net effect: any second-and-onward create whose name shares a 3-letter base prefix surfaces as `HTTP 500 Internal server error` instead of getting a unique suffix
> - This pull request walks the `error.cause` chain so the detector recognises both the bare driver error and the wrapped form
> - The benefit is that the retry loop functions as intended — multi-company-create flows (provisioning new orgs, importing company sets) stop spuriously 500'ing

Closes #6088.

## What Changed

- `server/src/services/companies.ts`: rewrite `isIssuePrefixConflict` to traverse `error.cause` so it recognises `code: "23505"` + `constraint: "companies_issue_prefix_idx"` whether the postgres error is bare or wrapped in `DrizzleQueryError`. Hoist the helper to a module-level export (pure function, no `db` dependency) so it can be unit-tested without a database harness.
- `server/src/__tests__/issue-prefix-conflict.test.ts`: new unit tests covering the bare error, the drizzle-wrapped form, deeper nesting, the alternative `constraint_name` field used by some drivers, and negative cases (different constraint, different code, null/undefined/primitives).
- No behavior change for `createCompanyWithUniquePrefix` itself; the retry loop now actually retries instead of rethrowing.

## Verification

Reproduce on master (without this patch), against a fresh CEO-bootstrapped instance with any instance-admin board API key:

```bash
KEY=$(cat ~/.paperclip_admin_api_key)
for name in dogdevs dogcic dogjobagency; do
  curl -sS -X POST http://paperclip.home.arpa/api/companies \
    -H "Authorization: Bearer $KEY" -H "Content-Type: application/json" \
    -d "{\"name\":\"$name\"}" -w "\n  status: %{http_code}\n"
done
# Expected on master: 201, 500, 500
# Expected on this PR: 201, 201, 201  (prefixes DOG, DOGA, DOGAA)
```

Unit-test the helper directly:

```bash
pnpm --filter @paperclipai/server test issue-prefix-conflict
```

All assertions cover the wrapping shapes drizzle-orm actually produces against postgres-js as of the version pinned by this repo.

## Risks

Very low. The only behavior change is that `isIssuePrefixConflict` now also matches errors wrapped in `DrizzleQueryError` and chains deeper. No call-sites changed; no schema migrations; pure recognition logic with positive unit-test coverage and explicit negative cases (different constraint name, different error code, non-Postgres errors, primitives).

The function is now a top-level export (was previously inside the `companyService(db)` closure). That's a public-surface addition rather than a removal, so no downstream import will break.

## Model Used

- Provider: Anthropic
- Context window: 1M tokens
- Reasoning mode: high-effort thinking enabled
- Tool use: Bash, Read, Edit, Write
- Workflow: bug surfaced via direct repro against an `authenticated/private`-mode instance during a multi-company provisioning sequence; root-causing involved reading the compiled `node_modules/@paperclipai/server/dist/services/companies.js` against the live error trace in `journalctl -u paperclip.service` and tracing the wrapped error shape via Python `urllib` probes — then mapping the same line back to the source under `server/src/services/companies.ts` for the fix.

## Checklist

- [x] I have checked ROADMAP.md and confirmed this PR does not duplicate planned core work (purely a bug fix)
- [ ] I have run tests locally and they pass — *I do not have `pnpm` on the workstation I authored this from; the new unit-test file is self-contained vitest with no DB dependency; please defer to CI*
- [x] If this change affects the UI, I have included before/after screenshots — *N/A, server-only fix*
- [x] I have updated relevant documentation to reflect my changes — *no docs reference this helper*

---

🤖 Made with Claude Code (Opus 4.7, 1M context, high-effort thinking)

### DIFF
--- a/server/src/__tests__/issue-prefix-conflict.test.ts
+++ b/server/src/__tests__/issue-prefix-conflict.test.ts
@@ -1,0 +1,69 @@
+import { describe, expect, it } from "vitest";
+import { isIssuePrefixConflict } from "../services/companies.js";
+
+describe("isIssuePrefixConflict", () => {
+  it("recognises a bare postgres-js error (no wrapping)", () => {
+    const err = {
+      name: "PostgresError",
+      code: "23505",
+      constraint: "companies_issue_prefix_idx",
+      severity: "ERROR",
+      message: "duplicate key value violates unique constraint",
+    };
+    expect(isIssuePrefixConflict(err)).toBe(true);
+  });
+
+  it("recognises the same error after drizzle-orm wraps it in DrizzleQueryError", () => {
+    // Shape produced by drizzle-orm/postgres-js when an INSERT fails on a
+    // unique constraint: the underlying postgres error is nested at `cause`.
+    const err = {
+      name: "DrizzleQueryError",
+      message: 'Failed query: insert into "companies" ...',
+      cause: {
+        name: "PostgresError",
+        code: "23505",
+        constraint: "companies_issue_prefix_idx",
+        severity: "ERROR",
+      },
+    };
+    expect(isIssuePrefixConflict(err)).toBe(true);
+  });
+
+  it("recognises deeper nesting via the cause chain", () => {
+    const err = {
+      name: "OuterWrapper",
+      cause: {
+        name: "InnerWrapper",
+        cause: {
+          code: "23505",
+          constraint: "companies_issue_prefix_idx",
+        },
+      },
+    };
+    expect(isIssuePrefixConflict(err)).toBe(true);
+  });
+
+  it("accepts the alternative `constraint_name` field used by some drivers", () => {
+    const err = { code: "23505", constraint_name: "companies_issue_prefix_idx" };
+    expect(isIssuePrefixConflict(err)).toBe(true);
+  });
+
+  it("returns false for a different unique-constraint violation", () => {
+    const err = {
+      cause: { code: "23505", constraint: "companies_name_idx" },
+    };
+    expect(isIssuePrefixConflict(err)).toBe(false);
+  });
+
+  it("returns false for a non-conflict postgres error", () => {
+    const err = { cause: { code: "23502", constraint: "companies_issue_prefix_idx" } };
+    expect(isIssuePrefixConflict(err)).toBe(false);
+  });
+
+  it("returns false for null / undefined / primitives", () => {
+    expect(isIssuePrefixConflict(null)).toBe(false);
+    expect(isIssuePrefixConflict(undefined)).toBe(false);
+    expect(isIssuePrefixConflict("23505")).toBe(false);
+    expect(isIssuePrefixConflict(23505)).toBe(false);
+  });
+});
--- a/server/src/__tests__/issue-prefix-conflict.test.ts
+++ b/server/src/__tests__/issue-prefix-conflict.test.ts
@@ -66,4 +66,17 @@ describe("isIssuePrefixConflict", () => {
     expect(isIssuePrefixConflict("23505")).toBe(false);
     expect(isIssuePrefixConflict(23505)).toBe(false);
   });
+
+  it("terminates promptly on a circular cause graph (depth-cap regression)", () => {
+    // Construct a cycle a → b → a. A naive while(cause) traversal would spin
+    // forever. The MAX_DEPTH guard must short-circuit and return false.
+    const a: Record<string, unknown> = { code: "OTHER", constraint: "irrelevant" };
+    const b: Record<string, unknown> = { code: "OTHER", constraint: "irrelevant" };
+    a.cause = b;
+    b.cause = a;
+    const start = Date.now();
+    expect(isIssuePrefixConflict(a)).toBe(false);
+    // Sanity: a hung loop would never return; assert this finished quickly.
+    expect(Date.now() - start).toBeLessThan(100);
+  });
 });
--- a/server/src/services/companies.ts
+++ b/server/src/services/companies.ts
@@ -42,8 +42,12 @@ import { environmentService } from "./environments.js";
  * surface). Exported for unit testing.
  */
 export function isIssuePrefixConflict(error: unknown): boolean {
+  // Bounded depth so a pathological (circular or pathologically deep) error
+  // graph cannot spin the loop. Real Postgres / Drizzle errors nest at most
+  // 1–2 levels; 10 leaves comfortable headroom without becoming a foot-gun.
+  const MAX_DEPTH = 10;
   let cursor: unknown = error;
-  while (cursor && typeof cursor === "object") {
+  for (let depth = 0; cursor && typeof cursor === "object" && depth < MAX_DEPTH; depth += 1) {
     const node = cursor as {
       code?: string;
       constraint?: string;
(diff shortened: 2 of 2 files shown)


remymiguk:
Thanks for the review @greptile-apps. Addressed the depth-cap suggestion in [3070e13a](https://github.com/paperclipai/paperclip/pull/6091/commits/3070e13a):

- `isIssuePrefixConflict` now uses a `for` loop with `MAX_DEPTH = 10` so a pathological / circular error graph cannot spin the loop forever. Comment notes that real Postgres/Drizzle errors nest at most 1–2 levels.
- Added a regression test (`terminates promptly on a circular cause graph`) that constructs an `a → b → a` cycle and asserts the function returns `false` in <100ms. A naive `while (cause)` would never return; the bounded loop short-circuits and falls through.

Local test run:
```
RUN  v3.2.4
✓ |@paperclipai/server| src/__tests__/issue-prefix-conflict.test.ts (8 tests) 14ms
Test Files  1 passed (1)
     Tests  8 passed (8)
```

Re-trigger Greptile when you're ready.

🤖 Made with Claude Code (Opus 4.7, 1M context, high-effort thinking)

PR #6092, opened 2026-05-16
fix: don't block self-update with the self-removal guard (#6089)

### LINKED ISSUE #6089: PATCH /companies/:id/members/:memberId returns "You cannot remove yourself." for self-update (not removal)
### Summary

In the Members admin UI, updating your own membership (role change, status toggle, permissions edit) fails with **"You cannot remove yourself."** even though the operation is a non-destructive update. The same misleading error text is shown for permission tweaks, role downgrades, and status changes that do not remove the member.

Root cause: `getProtectedMemberReason` in `server/src/routes/access.ts` short-circuits on `member.principalId === req.actor.userId` before considering the operation kind, and `PATCH /members/:memberId` calls `assertCanManageCompanyMember` with `operation: "update"` — so the removal guard fires even though nothing is being removed.

### Reproduce

1. Sign in as a user with **`owner`** (or `admin` with `users:manage_permissions`) membership in some company `X`.
2. Open the company dashboard → Settings → Members.
3. Click your own row → try to change your role (e.g. `owner → admin`), or toggle status, or update any permission grant.
4. UI displays toast: **"Failed to update member — You cannot remove yourself."** Server response: `HTTP 403 {"error":"You cannot remove yourself."}`.

Reproduces against `paperclipai 2026.513.0` with a fresh CEO-bootstrapped instance.

### Root cause

`server/src/routes/access.ts` (around line 1082):

```ts
async function getProtectedMemberReason(req, access, companyId, member, opts) {
  if (member.principalType !== "user")
    return "Only human company members can be removed.";
  if (req.actor.type !== "board")
    return "Board access is required to remove members.";
  if (member.principalId === req.actor.userId)
    return "You cannot remove yourself.";            // <-- fires for ANY operation

  // ...operation-aware checks (only run when target ≠ self)
  if (opts?.operation === "archive") {
    if (targetRole === "owner") return "Board owners cannot be removed from company access.";
    if (targetRole === "admin") return "Company admins cannot be removed from company access.";
  }
  // ...actor-vs-target role-rank comparison
}

async function assertCanManageCompanyMember(req, access, companyId, member, operation = "update") {
  const reason = await getProtectedMemberReason(req, access, companyId, member, { operation });
  if (reason) throw forbidden(reason);
}
```

The PATCH route at `server/src/routes/access.ts:4015` calls `assertCanManageCompanyMember(req, access, companyId, memberToUpdate)` (operation defaults to `"update"`), so the self-guard fires the removal message even for non-removal operations.

The underlying business invariants (e.g. "cannot remove the last active owner") are already enforced by the SQL transaction at the PATCH route level via `SELECT … FOR UPDATE`, so the self-guard is not needed as a safety net for that case.

### Proposed fix

Make the self-protection operation-aware: keep the existing message for self-archive, return null for self-update and let the route-level invariants take over.

```ts
- if (member.principalId === req.actor.userId)
-   return "You cannot remove yourself.";
+ if (member.principalId === req.actor.userId) {
+   return opts?.operation === "archive" ? "You cannot remove yourself." : null;
+ }
```

Note: simply removing the early return is not enough — the downstream `actorRole`-vs-`targetRole` rank check at line ~1113 would also misfire on self (`targetRank >= actorRank` is trivially true → "You can only remove users below your company role" thrown against yourself). Returning `null` short-circuits past those subsequent checks for the self-update case.

**Fix PR available**: #6092.

### Environment

- `paperclipai 2026.513.0` (npm global)
- Node 20.x, Postgres 17
- Debian 13 (Trixie) amd64, kernel 6.12 LXC
- Deployment mode: `authenticated`, exposure: `private`, bind: `lan`
- Browser flow exercised; HAR available on request.

### Workaround until fix lands

To change your own role/status, have a second board user (owner/admin) make the change on your behalf, OR use a direct DB UPDATE on `company_memberships` (respect the `cannot remove last active owner` invariant manually):

```sql
UPDATE company_memberships
SET membership_role = '<new_role>', updated_at = NOW()
WHERE company_id = '<company_uuid>'
  AND principal_id = '<my_user_id>'
  AND principal_type = 'user';
```

---

🤖 Made with Claude Code (Opus 4.7, 1M context, high-effort thinking)

remymiguk:
## Thinking Path

> - Company membership is the per-company permission primitive; the Members admin page lets owners/admins manage their own and others' rows (role, status, permission grants)
> - The PATCH route at `server/src/routes/access.ts:4015` calls `assertCanManageCompanyMember(req, access, companyId, memberToUpdate)` which defaults `operation` to `"update"`, but `getProtectedMemberReason` short-circuits any self-targeted call with the "You cannot remove yourself." message — regardless of whether the caller is trying to remove or merely update
> - Result: every self-update from the UI ("change my own role from owner → admin", "toggle my own status", "edit my own permission grant") fails with HTTP 403 + the misleading removal error
> - The downstream isInstanceAdmin / archive-role / actor-vs-target rank checks would also misfire on self if the early return were simply lifted — the rank comparison would throw "You can only remove users below your company role" against yourself
> - This pull request makes the self-protection operation-aware: keep blocking self-archive with the existing message, return null for self-update and let the route-level invariants take over (the PATCH route already enforces "cannot remove the last active owner" via a `SELECT … FOR UPDATE` lock)
> - The benefit is that the Members UI's self-update affordances stop spuriously 403'ing with text that doesn't match the operation

Closes #6089.

## What Changed

- `server/src/routes/access.ts`: `getProtectedMemberReason` now returns the self-removal message only when `opts.operation === "archive"`. For non-archive operations (default / explicit `"update"`), self-targeted calls return `null` and let the route-level invariants apply. Export the helper so it can be unit-tested without spinning up an Express request pipeline.
- `server/src/__tests__/protected-member-reason.test.ts`: new unit tests covering self-archive (still blocked with the explicit message), self-update (allowed), default operation parameter (treated as update), regression coverage for the rank-comparison branch that previously fired against self, plus non-self regression coverage (instance-admin target and owner-target archives still blocked).
- No change to `assertCanManageCompanyMember`, the PATCH/PATCH-permissions/PATCH-role-and-grants/POST-archive route handlers, or the underlying `setUserCompanyAccess` service-layer guard (`access-service.test.ts:186`'s "you cannot remove yourself" test still passes — that's a different code path).

## Verification

Reproduce on master (without this patch):

1. Sign in as a user with `owner` membership in some company X.
2. Open the company dashboard → Settings → Members.
3. Click your own row → change your role (`owner → admin`), or toggle status, or update any permission grant.
4. UI toast: "Failed to update member — You cannot remove yourself." API returns `HTTP 403 {"error":"You cannot remove yourself."}`.

With this patch:

1. Same flow; the role/status/permission change persists; API returns the updated member record. If the change would leave the company with zero active owners, the PATCH route's transaction-level guard ("Cannot remove the last active owner") fires instead — that's the correct invariant.
2. Archiving yourself via `POST /api/companies/:id/members/:memberId/archive` still returns 403 "You cannot remove yourself." (verified by the new test).

Unit-test the helper directly:

```bash
pnpm --filter @paperclipai/server test protected-member-reason
```

## Risks

Low.

- The self-protection narrowing is operation-aware: archive (the actual removal path) keeps the existing message. Only update is relaxed, and the route's "cannot remove the last active owner" SELECT-FOR-UPDATE invariant remains the structural safety net.
- The function is now exported (was module-private). That's a public-surface addition rather than a removal; existing imports unaffected.
- One behavior shift to be aware of: an owner can now demote themselves to `admin` via the UI as long as another active owner remains. Previously this required a second user to perform the demotion. This is the intended UX of the Members page (each row has Edit/Delete affordances pointing at PATCH/POST-archive), so I'm treating this as the bug being fixed rather than a new behavior — but flagging it explicitly here for product review.

## Model Used

- Provider: Anthropic
- Context window: 1M tokens
- Reasoning mode: high-effort thinking enabled
- Tool use: Bash, Read, Edit, Write
- Workflow: bug surfaced via a browser HAR capture during a Members admin page interaction; root-causing involved tracing `assertCanManageCompanyMember` callers across the routes, confirming via `grep` that the service-layer `setUserCompanyAccess` self-guard (`access-service.test.ts:186`) is a separate code path unaffected by this change, and verifying the route-level "cannot remove last active owner" invariant already lives inside the `SELECT … FOR UPDATE` transaction at `server/src/routes/access.ts:4145–4156`.

## Checklist

- [x] I have checked ROADMAP.md and confirmed this PR does not duplicate planned core work (purely a bug fix)
- [ ] I have run tests locally and they pass — *I do not have `pnpm` on the workstation I authored this from; the new unit-test file is self-contained vitest with no DB dependency; please defer to CI*
- [x] If this change affects the UI, I have included before/after screenshots — *N/A, the change is server-side; the visible UI difference is "toast goes from 403-error to success-then-refresh"*
- [x] I have updated relevant documentation to reflect my changes — *no docs reference this helper*

---

🤖 Made with Claude Code (Opus 4.7, 1M context, high-effort thinking)

### DIFF
--- a/server/src/__tests__/protected-member-reason.test.ts
+++ b/server/src/__tests__/protected-member-reason.test.ts
@@ -1,0 +1,113 @@
+import type { Request } from "express";
+import { describe, expect, it } from "vitest";
+import { getProtectedMemberReason } from "../routes/access.js";
+
+// Minimal access stub — the function only consults `isInstanceAdmin` when the
+// target is *not* the actor, so for self-update tests it is never reached.
+const noAccessStub = {
+  isInstanceAdmin: async () => false,
+  // Cast through unknown so we don't have to construct the full service shape
+  // for tests that don't exercise it.
+} as unknown as Parameters<typeof getProtectedMemberReason>[1];
+
+function ownerActor(userId: string): Request["actor"] {
+  return {
+    type: "board",
+    source: "session",
+    userId,
+    isInstanceAdmin: false,
+    companyIds: ["company-1"],
+    memberships: [{ companyId: "company-1", membershipRole: "owner", status: "active" }],
+  } as Request["actor"];
+}
+
+function req(actor: Request["actor"], method = "PATCH"): Request {
+  return { method, actor } as Request;
+}
+
+describe("getProtectedMemberReason — self target", () => {
+  const SELF = "user-self";
+  const selfMember = {
+    principalId: SELF,
+    principalType: "user",
+    membershipRole: "owner" as string | null,
+  };
+
+  it("blocks self-archive with the explicit removal message", async () => {
+    const reason = await getProtectedMemberReason(
+      req(ownerActor(SELF)),
+      noAccessStub,
+      "company-1",
+      selfMember,
+      { operation: "archive" },
+    );
+    expect(reason).toBe("You cannot remove yourself.");
+  });
+
+  it("allows self-update (no protective reason raised)", async () => {
+    const reason = await getProtectedMemberReason(
+      req(ownerActor(SELF)),
+      noAccessStub,
+      "company-1",
+      selfMember,
+      { operation: "update" },
+    );
+    expect(reason).toBeNull();
+  });
+
+  it("allows self-update even when operation is unspecified (default treated as update)", async () => {
+    const reason = await getProtectedMemberReason(
+      req(ownerActor(SELF)),
+      noAccessStub,
+      "company-1",
+      selfMember,
+      undefined,
+    );
+    expect(reason).toBeNull();
+  });
+
+  it("self-update is allowed regardless of role rank (regression: previously fired 'You can only remove users below your company role')", async () => {
+    // operator updating their own membership — same rank, would have triggered
+    // the rank-comparison branch had we let the function fall through.
+    const operatorActor: Request["actor"] = {
+      ...(ownerActor(SELF) as object),
+      memberships: [{ companyId: "company-1", membershipRole: "operator", status: "active" }],
+    } as Request["actor"];
+    const operatorMember = { ...selfMember, membershipRole: "operator" as string | null };
+    const reason = await getProtectedMemberReason(
+      req(operatorActor),
+      noAccessStub,
+      "company-1",
+      operatorMember,
+      { operation: "update" },
+    );
+    expect(reason).toBeNull();
+  });
+});
+
+describe("getProtectedMemberReason — non-self target (regression coverage)", () => {
+  it("still blocks archive of another instance admin", async () => {
+    const adminAccessStub = {
+      isInstanceAdmin: async (id: string) => id === "admin-2",
+    } as unknown as Parameters<typeof getProtectedMemberReason>[1];
+    const reason = await getProtectedMemberReason(
+      req(ownerActor("owner-1")),
+      adminAccessStub,
+      "company-1",
+      { principalId: "admin-2", principalType: "user", membershipRole: "admin" },
+      { operation: "archive" },
+    );
+    expect(reason).toBe("Instance admins cannot be removed from company access.");
+  });
+
+  it("still blocks archive of an owner", async () => {
+    const reason = await getProtectedMemberReason(
+      req(ownerActor("owner-1")),
+      noAccessStub,
+      "company-1",
+      { principalId: "owner-2", principalType: "user", membershipRole: "owner" },
+      { operation: "archive" },
+    );
+    expect(reason).toBe("Board owners cannot be removed from company access.");
+  });
+});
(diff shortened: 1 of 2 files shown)


remymiguk:
Update: `pnpm` is now available locally so I ran the unit test for this PR:

```
$ corepack pnpm exec vitest run server/src/__tests__/protected-member-reason.test.ts --no-coverage
 RUN  v3.2.4
 ✓ |@paperclipai/server| src/__tests__/protected-member-reason.test.ts (6 tests) 16ms
 Test Files  1 passed (1)
      Tests  6 passed (6)
   Start at  03:13:35
   Duration  15.43s
```

6/6 passing — covers self-archive (still blocked), self-update (now allowed), default-operation (treated as update), the role-rank regression case, and two non-self regression cases (instance-admin target, owner target). Treat the PR-description "tests locally" checkbox as ✅ — GitHub's `gh pr edit --body-file` is silently no-op'ing under the Projects-classic GraphQL deprecation warning, so I couldn't tick it in the description itself.

🤖 Made with Claude Code (Opus 4.7, 1M context, high-effort thinking)

PR #6112, opened 2026-05-16
Fix root-level SKILL.md file inventory across all import paths (#2037)

### LINKED ISSUE #2037: Local skill imports from a directory can expose only SKILL.md in API inventory even when the skill has scripts/references
## Summary
Importing a local skill directory into Paperclip can succeed, but the resulting company skill API inventory may expose only `SKILL.md` even when the source skill clearly includes additional files like `scripts/` and `references/`.

## Concrete example
I imported a local skill directory shaped like this:
- `SKILL.md`
- `scripts/...`
- `references/...`

The import completed and the runtime source still pointed at the original local directory, but:
- `GET /companies/:companyId/skills` showed only `SKILL.md` in `fileInventory`
- `GET /companies/:companyId/skills/:skillId/files?path=scripts/...` returned `Skill file not found`

## Why this matters
This makes local skill imports look partially broken even when runtime execution may still work via symlink/source directory.

For skills that rely on bundled scripts or reference files, the control-plane view becomes misleading:
- operators cannot inspect the full imported skill through the API/UI
- it is unclear whether the skill is truly usable or only partially imported
- debugging imported local skills becomes harder than it should be

## Expected behavior
For `local_path` skill imports from a directory:
- `fileInventory` should include the additional files discovered under that skill directory
- `GET /files?path=...` should work for those files
- API/UI state should match what the runtime actually has access to

## Notes
This was observed with a real local skill directory, not a GitHub or catalog import. If helpful I can provide the exact directory layout and API responses I saw.

iamhansi:
Fixes #2037

## Thinking Path

> - Companies import skills from local directories, GitHub URLs, and inline packages
> - Each imported skill exposes a file inventory (SKILL.md, references/, scripts/, and any other files/folders) so agents can access skill resources at runtime
> - When a skill's SKILL.md lives at the root of its source (not in a subdirectory), `path.posix.dirname("SKILL.md")` returns `"."`
> - The inventory filter used `entry.startsWith("./")` which fails because relative paths like `references/foo.md` don't start with `"./"`
> - This caused root-level skills to expose only SKILL.md in their file inventory, silently dropping all other files and folders
> - This PR fixes the root-level skill inventory filter across all three import paths
> - The benefit is that skills imported from any source now correctly expose their full file tree to agents

## What Changed

- Fixed `readLocalSkillImports` to detect root-level skills (`skillDir === "."`) and include all files and folders instead of filtering with a broken prefix match that excluded everything
- Applied the same fix to `readInlineSkillImports` (package imports) and `readUrlSkillImports` (GitHub/URL imports) which had the identical bug
- Exported `readLocalSkillImports` and `readInlineSkillImports` for direct unit testing
- Added 130 lines of test coverage verifying root-level skill inventory across all import paths (18 tests total)

## Verification

- `npx vitest run server/src/__tests__/company-skills.test.ts` — all 18 tests pass
- `pnpm -w run typecheck` — all packages pass cleanly
- Full test suite (`pnpm -w run test`) passes with only pre-existing failures in `cursor-local-execute.test.ts` (unrelated `ERR_MODULE_NOT_FOUND` on master)
- Manual: import a skill repo with SKILL.md at root → verify all files/folders (not just SKILL.md) appear in the file inventory

## Risks

- Low risk. The fix only changes behavior for the `skillDir === "."` case (root-level skills). Skills in subdirectories are unaffected — their existing `startsWith("subdir/")` filter path is unchanged.
- Two previously-private functions are now exported for testability; no API contract change since `company-skills.ts` is an internal service module.

## Model Used

Anthropic Claude Opus 4.6 (`claude-opus-4-6`). Extended thinking enabled, tool use (Bash, Read, Edit), via Claude Code CLI. ~200k context window.

## Checklist

### DIFF
--- a/server/src/__tests__/company-skills.test.ts
+++ b/server/src/__tests__/company-skills.test.ts
@@ -255,10 +255,13 @@ describe("local directory skill imports", () => {
     const rootPaths = new Set(rootSkill.fileInventory.map((e) => e.path));
     expect(rootPaths).toContain("SKILL.md");
     expect(rootPaths).toContain("references/root-ref.md");
+    expect(rootPaths).not.toContain("sub-skill/SKILL.md");
+    expect(rootPaths).not.toContain("sub-skill/scripts/sub.sh");
 
     const subPaths = new Set(subSkill.fileInventory.map((e) => e.path));
     expect(subPaths).toContain("SKILL.md");
     expect(subPaths).toContain("scripts/sub.sh");
+    expect(subPaths).not.toContain("references/root-ref.md");
   });
 });
 
--- a/server/src/__tests__/company-skills.test.ts
+++ b/server/src/__tests__/company-skills.test.ts
@@ -7,6 +7,7 @@ import {
   findMissingLocalSkillIds,
   normalizeGitHubSkillDirectory,
   parseSkillImportSourceInput,
+  readInlineSkillImports,
   readLocalSkillImportFromDirectory,
   readLocalSkillImports,
 } from "../services/company-skills.js";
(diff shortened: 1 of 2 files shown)


iamhansi:
Addressed Greptile's P1 feedback in 6ce05307:

All three import paths (`readLocalSkillImports`, `readInlineSkillImports`, `readUrlSkillImports`) now collect `siblingSkillDirs` and exclude them from the root skill's inventory filter. When a repo has both a root SKILL.md and sub-directory skills, each skill's inventory contains only its own files.

Tests now assert mutual exclusion — root inventory does not contain sub-skill files (`sub-skill/SKILL.md`, `sub-skill/scripts/sub.sh`), and sub-skill inventory does not contain root-level files (`references/root-ref.md`).

Issue #5997, opened 2026-05-14
`openclaw-gateway` adapter: `paperclip` envelope in `agentParams` gets rejected by OpenClaw's strict schema


titovka-infra:
## Summary

The `@paperclipai/adapter-openclaw-gateway` adapter (see `packages/adapters/openclaw-gateway/src/server/execute.ts:1139`) unconditionally injects `agentParams.paperclip = paperclipPayload` into every heartbeat handoff. The OpenClaw side validates `agentParams` against a strict schema with `additionalProperties: false`, and **no released OpenClaw version (as of 2026.5.12-beta.8) recognises the `paperclip` property**. As a result, every Paperclip→OpenClaw run fails immediately with:
The receiving agent goes to `status: error`, all assigned issues transition to `blocked`, and Paperclip's `stranded_issue_recovery` mechanism keeps generating duplicate follow-up issues until the agent is paused manually.

## Environment

- Paperclip: latest `paperclip:local` build (commit on `main` as of 2026-05-12)
- OpenClaw gateway: `2026.5.7` (current stable) on a separate host, accessed via `wss://genius.transconvey.ru:8443/`
- Both ends authenticated and paired correctly — failure is purely schema validation.
- Verified the same failure mode against npm pre-release `openclaw@2026.5.12-beta.8`: `paperclip` is still not in the gateway's `agentParams` schema bundle.

## Reproduction

1. Create a company in Paperclip.
2. Onboard an agent with `adapterType: openclaw_gateway` pointing at a stock OpenClaw gateway (`2026.5.7` or any beta up to `2026.5.12-beta.8`).
3. Approve pairing, run any `[Inbox] …` issue.
4. Observe `heartbeat_runs.error = invalid agent params: at root: unexpected property 'paperclip'`.

## Root cause (in this repo)

`packages/adapters/openclaw-gateway/src/server/execute.ts`, around line 1139:

```ts
const agentParams: Record<string, unknown> = {
  ...payloadTemplate,
  message,
  sessionKey,
  idempotencyKey: ctx.runId,
};
delete agentParams.text;
agentParams.paperclip = paperclipPayload;   // ← rejected by OpenClaw's strict schema
```

The `paperclip` envelope is a useful Paperclip-side context bundle (`runId`, `issueId`, `wakeReason`, `workspace`, `workspaceRuntime`, etc.), but OpenClaw is a generic gateway shared with other clients and has no reason to declare it as a known property.

## Suggested fixes (in order of preference)

[...]

Issue #4081, opened 2026-04-20
Published openclaw_gateway adapter adds root-level 'paperclip' field that published OpenClaw rejects


UnPractical91:
## Summary

The published `@paperclipai/adapter-openclaw-gateway` appears to be incompatible with published OpenClaw builds because the adapter adds a root-level `paperclip` object to every `req agent` payload, and published OpenClaw rejects that field during agent param validation.

I opened the matching OpenClaw-side issue here:
- openclaw/openclaw#69137

Filing here as well because the published adapter/docs currently describe this payload shape as intentional.

## Versions tested

Paperclip / adapter:
- `paperclipai@2026.416.0`
- `@paperclipai/adapter-openclaw-gateway@2026.416.0`

OpenClaw:
- stable: `2026.4.15`
- beta: `2026.4.19-beta.2`

Also spot-checked older Paperclip canaries, which still add the same root-level `paperclip` field:
- `2026.415.0-canary.7`
- `2026.413.0-canary.3`
- `2026.411.0-canary.8`
- `2026.409.0-canary.4`

## What the adapter currently documents and does

The adapter README says the gateway flow is:

1. `connect`
2. `req agent`
3. `req agent.wait`

The published adapter docs also say:

> `paperclip` (object): standardized Paperclip context added to every gateway agent request

And the published adapter code does:

```js
const agentParams = {
  ...payloadTemplate,
  message,
  sessionKey,
  idempotencyKey: ctx.runId,
};
...
agentParams.paperclip = paperclipPayload;
```

## Actual result with published OpenClaw

The gateway connection succeeds, but the `req agent` call fails with:

```text
invalid agent params: at root: unexpected property 'paperclip'
```

OpenClaw stable and beta both reject the root-level `paperclip` field via additional-properties validation.

## Expected

If the current adapter behavior is correct, there should be a published OpenClaw version that accepts this documented metadata block on `req agent`.

If not, the adapter/docs may need to send this metadata through a different supported path.

## Why filing here too

This looks like a published contract mismatch between the two sides rather than a local setup problem. Since the adapter intentionally adds the field and documents that behavior, it seemed useful to report it here as well.

UnPractical91:
Sharing the local workflow that got this unstuck for us. This produced a working local compatibility hack on our machine, not a claimed upstream fix, and ymmv.

The workaround was:
- stop sending the root-level `paperclip` field in OpenClaw agent params
- explicitly send the intended OpenClaw `agentId`
[...]

Issue #617, opened 2026-03-11
Openclaw-gateway adapter: unexpected 'paperclip' property rejected by OpenClaw strict validation


marcs7:
## Bug Description                                                                                                                                       
                                                                                                                                                           
  The `openclaw-gateway` adapter includes a `paperclip` object as a top-level property in the WebSocket `agent` request params. OpenClaw performs strict   
  validation on agent params and rejects unknown keys with:                                                                                                
                                                                                                                                                           
  invalid agent params: at root: unexpected property 'paperclip'                                                                                           
                                                                                                                                                         
  ## Location                                                                                                                                              
                                                                                                                                                         
  `packages/adapters/openclaw-gateway/src/server/execute.ts` line 1070-1076:                                                                               
   
                                                                                                                                           
```typescript
  const agentParams: Record<string, unknown> = {                                                                                                         
      ...payloadTemplate,                                                                                                                                
      paperclip: paperclipPayload,  // ← rejected by OpenClaw                                                                                              
      message,
      sessionKey,                   // ← accepted                                                                                                          
      idempotencyKey: ctx.runId,    // ← accepted (required)                                                                                               
  };   
```                                                                                                                                                
                                                                                                                                                           
  Expected Behavior                                                                                                                                        
                                                                                                                                                           
  Only properties accepted by the OpenClaw gateway protocol should be sent as top-level keys. sessionKey and idempotencyKey are accepted (and required),   
  but paperclip is not.                                                                                                                                  
                                                                                                                                                           
  Suggested Fix                                                                                                                                          
                                                                                                                                                         
  Embed the paperclip payload inside the message field (e.g. as a <paperclip-context> block) instead of a separate top-level property. This way the agent  
  still receives the full Paperclip context without breaking OpenClaw's strict param validation.
                                                                                                                                                           
  Environment                                                                                                                                              
                                                                                                                                                         
  - Paperclip: built from latest main                                                                                                                      
  - OpenClaw: v2026.3.8                                                                                                                                  
  - Adapter: openclaw_gateway                                                                                                                              
  - Gateway transport: WebSocket (wss://)

srozov:
Opened #6188 for this.

It removes the schema-rejected root-level `paperclip` field from OpenClaw gateway agent params, keeps Paperclip wake context in the message, and preserves compatibility with both protocol 3 and protocol 4 gateways.

PR #6138, opened 2026-05-16
fix(openclaw-gateway): use protocol 4

bryangauvin:
## Summary
- update the OpenClaw gateway adapter handshake to advertise protocol 4
- update the gateway probe to use the same protocol range

## Why
OpenClaw 2026.5.12 accepts protocol 4 for the authenticated gateway handshake. Paperclip's bundled OpenClaw gateway adapter was still advertising protocol 3, which required a local post-update patch after upgrading Paperclip/OpenClaw.

## Verification
- pnpm --filter @paperclipai/adapter-openclaw-gateway typecheck
- pnpm --filter @paperclipai/adapter-openclaw-gateway build

Note: pnpm exec vitest run packages/adapters/openclaw-gateway/src/server/test.ts does not run in this repo because that file is an adapter environment probe, not a Vitest-matched test file.

### DIFF
--- a/packages/adapters/openclaw-gateway/src/server/execute.ts
+++ b/packages/adapters/openclaw-gateway/src/server/execute.ts
@@ -85,7 +85,7 @@ type GatewayClientRequestOptions = {
   expectFinal?: boolean;
 };
 
-const PROTOCOL_VERSION = 3;
+const PROTOCOL_VERSION = 4;
 const DEFAULT_SCOPES = ["operator.admin"];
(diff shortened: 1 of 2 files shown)


srozov:
For visibility: #6188 overlaps with this protocol fix, but uses `minProtocol: 3` / `maxProtocol: 4` instead of hardcoding protocol 4. It also addresses the root-level `paperclip` schema rejection tracked in #4081, #5997, and #617.

PR #6176, opened 2026-05-17
perf(db): add (agent_id, status, created_at) composite index on heartbeat_runs

felipecpaiva:
## Summary

Adds a composite index on `heartbeat_runs (agent_id, status, created_at)` to fix sequential scans on the table's hot read path.

## Background

`server/src/services/heartbeat.ts` issues three families of queries that all filter by `(agent_id, status)`:

- `countRunningRunsForAgent` — `WHERE agent_id = ? AND status = 'running'`
- `startNextQueuedRunForAgent` — `WHERE agent_id = ? AND status = 'queued' ORDER BY created_at LIMIT N`
- `enqueueWakeup` coalescing — `WHERE agent_id = ? AND status IN ('queued','running') ORDER BY created_at DESC`

The existing indexes on `heartbeat_runs` all lead with `company_id`:

- `heartbeat_runs_company_agent_started_idx (company_id, agent_id, started_at)`
- `heartbeat_runs_company_liveness_idx (company_id, liveness_state, created_at)`
- `heartbeat_runs_company_status_last_output_idx (company_id, status, last_output_at)`
- `heartbeat_runs_company_status_process_started_idx (company_id, status, process_started_at)`

None of these satisfy the hot queries above because Postgres needs `company_id` as a leading equality predicate.

## Impact on a long-running deployment

Measured on a real deployment with 937,232 rows / 638 MB:

```
EXPLAIN (ANALYZE, BUFFERS) SELECT count(*) FROM heartbeat_runs
  WHERE agent_id = ? AND status = 'completed';

-- BEFORE
Parallel Seq Scan on heartbeat_runs
  Filter: ((agent_id = ?) AND (status = 'completed'::text))
  Buffers: shared hit=17205 read=38897
Execution Time: 11714 ms

-- AFTER (same query, with this PR's index applied out-of-band as proof)
Index Scan using heartbeat_runs_agent_status_created_idx
Execution Time: < 5 ms
```

Multiple concurrent calls (9+) were observed stacking up on `IO:DataFileRead` in `pg_stat_activity`, starving the Node event loop and slowing every other endpoint to 5–11s TTFB.

## Why this column order

`(agent_id, status, created_at)`:
- Equality on `agent_id` first — highest selectivity for the per-agent queries.
- Equality on `status` second — small enum, but needed to make `IN ('queued','running')` an index range.
- `created_at` last — lets `ORDER BY created_at` skip a separate sort.

## Risks

- First-deploy index build on a large existing `heartbeat_runs` table will hold a write lock briefly (`CREATE INDEX` is non-concurrent in your transactional migration runner). On fresh deployments this is instant. On long-running deployments like the one above, it's a 30–60s write pause. Operators with very large tables can build the index out-of-band with `CONCURRENTLY` ahead of the migration if needed.
- No data change; pure read-path optimization.

## Tests

- `pnpm -r typecheck` — passes
- `pnpm test:run` — 2 pre-existing failures on embedded Postgres tests (environment issue unrelated to this change; same failures on unmodified master on ARM64 host)
- `pnpm build` — passes

Migration generated with `pnpm --filter @paperclipai/db run generate`.

### DIFF
--- a/packages/db/src/migrations/0086_tan_morlun.sql
+++ b/packages/db/src/migrations/0086_tan_morlun.sql
@@ -1,0 +1,1 @@
+CREATE INDEX "heartbeat_runs_agent_status_created_idx" ON "heartbeat_runs" USING btree ("agent_id","status","created_at");
--- a/packages/db/src/migrations/meta/0086_snapshot.json
+++ b/packages/db/src/migrations/meta/0086_snapshot.json
@@ -1,0 +1,17661 @@
+{
+  "id": "dc3d60b4-f06e-480d-8223-9ca26a96c7f9",
+  "prevId": "a7ba5d6c-9f74-487d-a9c1-56a4d5455b92",
+  "version": "7",
+  "dialect": "postgresql",
+  "tables": {
+    "public.activity_log": {
+      "name": "activity_log",
+      "schema": "",
+      "columns": {
+        "id": {
+          "name": "id",
+          "type": "uuid",
+          "primaryKey": true,
+          "notNull": true,
+          "default": "gen_random_uuid()"
+        },
+        "company_id": {
+          "name": "company_id",
+          "type": "uuid",
+          "primaryKey": false,
+          "notNull": true
+        },
+        "actor_type": {
+          "name": "actor_type",
+          "type": "text",
+          "primaryKey": false,
+          "notNull": true,
+          "default": "'system'"
+        },
+        "actor_id": {
+          "name": "actor_id",
+          "type": "text",
+          "primaryKey": false,
+          "notNull": true
+        },
+        "action": {
+          "name": "action",
+          "type": "text",
+          "primaryKey": false,
+          "notNull": true
+        },
+        "entity_type": {
+          "name": "entity_type",
+          "type": "text",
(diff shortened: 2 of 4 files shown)


felipecpaiva:
Friendly ping on this one — it has been open since May with no human review, and I want to make sure it is not stuck on something I can clear.

Two things that might be in the way:

**The diff size is misleading.** It reads as +17,675 lines, but 17,661 of those are
`packages/db/src/migrations/meta/0086_snapshot.json`, which Drizzle regenerates in full for every
migration. The actual change is 6 lines:

- `packages/db/src/migrations/0086_tan_morlun.sql` — 1 line, the `CREATE INDEX`
- `packages/db/src/schema/heartbeat_runs.ts` — 5 lines, the index declaration

**The migration number may have been taken since.** This was cut as `0086`. If `develop`/`master`
has advanced past that, say the word and I will renumber and regenerate the snapshot against
current head.

Happy to rebase, split it, or drop it entirely if the index is not wanted — the query pattern it
targets is the heartbeat-runs lookup by `(agent_id, status, created_at)`. Just let me know which,
and thanks either way for the time.

Issue #3817, opened 2026-04-16
Agent completion comments can be attributed to the user and retrigger the assignee (causing status drift, duplicate comments, and agent run loops)


Sam-Lam-Varadise:
### 🐞 **Bug Description**

When an agent-assigned task completes in Paperclip, the agent generates the appropriate completion markdown comment (e.g., review notes or artifact links). However, this comment sometimes appears as if it was authored by the logged-in human user (board operator), not the agent. This misattribution leads to a buggy workflow where:

- The comment is shown in the UI as from “You” rather than the expected agent name.
- Posting of this (misattributed) comment immediately retriggers the assigned agent.
- The agent then posts another completion comment, causing a loop where the task status can also drift from `done` ➔ `todo` or `in_progress` even after completion.
- Eventually, the agent may self-correct and close the loop, but not before multiple duplicate or near-duplicate comments and spurious re-runs.

---

#### **Steps to Reproduce**
1. Assign a task to an agent in Paperclip (see screenshot 1).
2. Allow the agent to complete review/work and attempt to post its completion or summary comment.
3. Observe that the completion comment appears as authored by you (the board user) rather than the agent, despite agent-typical wording.
4. The system re-runs the agent or marks the task as “todo/in progress” again, prompting the agent to post another (redundant) comment.
5. This may repeat, creating spurious comments and stray state transitions, before the agent eventually recognizes and self-corrects.

---

#### **Expected Behavior**
- Agent-generated completion comments should always be attributed to the agent, not the logged-in user.
- A completion comment posted for a run should satisfy the run’s comment requirement and not trigger another run or repetitive wake.
- The task should remain marked as `done` or its intended terminal state without status regression.

#### **Actual Behavior**
- Completion comments can appear authored by the board user (“You” in UI) despite reading as agent output.
- Such comments retrigger agent runs, causing duplicate comments and unnecessary status transitions (see screenshots).
- Status can drift from `done` back to `todo`/`in_progress`, requiring manual or agent self-correction.

---

#### **Screenshots**
- <img alt="Agent completion comment attributed to user and repeated" width="712" src="https://github.com/user-attachments/assets/92d70b59-0488-41dc-a305-09848860a0a2" />
- <img alt="Agent repeated run / self-heal after loop" width="721" src="https://github.com/user-attachments/assets/cb3bd818-3b78-469a-9d79-a6e861ba9176" />
[...]

PR #6180, opened 2026-05-17
Fix invalid bearer auth fallback on issue updates

jasondbramley:
## Summary
- Reject present-but-invalid bearer tokens in actor middleware with 401 instead of falling through to local trusted board access.
- Add regression coverage for a PATCH-style issue update path with a bad bearer token and the no-token local trusted baseline.

## Evidence
- `pnpm vitest run server/src/__tests__/auth-session-route.test.ts` - 4 tests passed
- `pnpm --filter @paperclipai/server typecheck` - passed

## Diff stat
```text
 server/src/__tests__/auth-session-route.test.ts | 51 +++++++++++++++++++++++++
 server/src/middleware/auth.ts                   | 11 +++---
 2 files changed, 57 insertions(+), 5 deletions(-)
```

## Relevant diff excerpt
```diff
diff --git a/server/src/middleware/auth.ts b/server/src/middleware/auth.ts
@@
 import { boardAuthService } from "../services/board-auth.js";
+import { unauthorized } from "../errors.js";
@@
     const token = authHeader.slice("bearer ".length).trim();
     if (!token) {
-      next();
+      next(unauthorized("Invalid bearer token"));
       return;
     }
@@
       const claims = verifyLocalAgentJwt(token);
       if (!claims) {
-        next();
+        next(unauthorized("Invalid bearer token"));
         return;
       }
@@
       if (!agentRecord || agentRecord.companyId !== claims.company_id) {
-        next();
+        next(unauthorized("Invalid bearer token"));
         return;
       }
```

Linked ITO ticket: ITO-594

### DIFF
--- a/server/src/__tests__/auth-session-route.test.ts
+++ b/server/src/__tests__/auth-session-route.test.ts
@@ -66,6 +66,57 @@ describe("actorMiddleware authenticated session profile", () => {
     });
   });
 
+  it("rejects invalid bearer tokens instead of falling back to local trusted board access", async () => {
+    const app = express();
+    app.use(
+      actorMiddleware(createDb(), {
+        deploymentMode: "local_trusted",
+      }),
+    );
+    app.patch("/api/issues/:id", (req, res) => {
+      res.json(req.actor);
+    });
+    app.use(
+      (
+        err: { status?: number; message?: string },
+        _req: express.Request,
+        res: express.Response,
+        _next: express.NextFunction,
+      ) => {
+        res.status(err.status ?? 500).json({ error: err.message ?? "Internal server error" });
+      },
+    );
+
+    const res = await request(app)
+      .patch("/api/issues/issue-1")
+      .set("Authorization", "Bearer definitely-not-valid")
+      .send({ status: "done" });
+
+    expect(res.status).toBe(401);
+    expect(res.body).toEqual({ error: "Invalid bearer token" });
+  });
+
+  it("keeps local trusted implicit board access when no bearer token is sent", async () => {
+    const app = express();
+    app.use(
+      actorMiddleware(createDb(), {
+        deploymentMode: "local_trusted",
+      }),
+    );
+    app.patch("/api/issues/:id", (req, res) => {
+      res.json(req.actor);
+    });
+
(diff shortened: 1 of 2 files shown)


jasondbramley:
Maintainer status check — this fix is still needed by our local Paperclip operations. Live check today shows the PR is still open with green checks; could a maintainer confirm whether anything is missing for review/merge? Happy to rebase or split if that would make it easier. @cryppadotta @devinfoley

BrennanJCollins:
+1 on this fix from a production user.

We hit this exact bug in `local_trusted` mode on Paperclip Desktop. Over the last ~6 weeks, 57 comments across our company's issue tracker were silently attributed to the implicit `local-board` user instead of the agent that wrote them (CEO=42, CMO=9, Staff Writer=2, unknown=4). The pattern showed up clearly: an agent's heartbeat JWT expires mid-run, the next `POST /comments` call falls through `actorMiddleware`, and the comment persists as `authorUserId: "local-board"`, `authorAgentId: null`, `authorType: "user"` — visually indistinguishable from the human board operator in the UI, which then made it look like decisions had been made by the human when they hadn't.

Walking the code on `master`, the fallthroughs in `server/src/middleware/auth.ts` lined up exactly with what this PR fixes:
- L24-34 pre-populates `req.actor` with the implicit `local-board` user before bearer-token resolution.
- L105 (empty token), L141 (`verifyLocalAgentJwt` returns null), L152 (agent missing / company mismatch), L157 (terminated/pending agent), L185 (API-key path with terminated agent) all `next()` without resetting `req.actor`. Combined with `assertAuthenticated` only rejecting `type === "none"`, that means any agent-auth failure silently writes as the board user.

This PR catches all five of those paths and adds a regression test that also pins the "no bearer token in local_trusted = implicit board" baseline so the fix can't regress in the other direction. That matches what we'd want shipped.

Our workaround downstream has been a pre-write JWT TTL check + post-write attribution verification wrapper, but it's a band-aid — the real fix is this one. Happy to see it merged.

Related: #3817 documents the symptom from the user-facing angle and should auto-close on merge.

Model used (per CONTRIBUTING.md): Claude Opus 4.6 by Anthropic (claude-opus-4-7), extended thinking enabled.

PR #1074, opened 2026-03-16
feat(plugins): add document CRUD methods to Plugin SDK

residentagent:
## Summary

Adds `ctx.issues.documents.list()`, `.get()`, `.upsert()`, and `.delete()` to the Plugin SDK, giving plugins the same document management capabilities already available via the REST API.

## Why this matters

[#940](https://github.com/paperclipai/paperclip/issues/940) identified that the REST API has full document CRUD endpoints (`GET/PUT/DELETE /api/issues/:id/documents/:key`) but the Plugin SDK has no way to access them. The server-side implementation landed in [#837](https://github.com/paperclipai/paperclip/pull/837) but the SDK was never extended. Plugins that need to read or write issue documents currently have no path to do so.

## Changes

Seven files across three packages, following the existing `listComments`/`createComment` wiring pattern:

| File | Change |
|------|--------|
| `packages/shared/src/constants.ts` | Add `issue.documents.read` and `issue.documents.write` capabilities |
| `packages/plugins/sdk/src/protocol.ts` | Add 4 RPC method signatures to `WorkerToHostMethods` |
| `packages/plugins/sdk/src/types.ts` | Add `PluginIssueDocumentsClient` interface, wire into `PluginIssuesClient.documents` |
| `packages/plugins/sdk/src/worker-rpc-host.ts` | Wire `documents.*` client methods via `callHost()` |
| `packages/plugins/sdk/src/host-client-factory.ts` | Add capability enforcement + handler routing in `METHOD_CAPABILITY_MAP` |
| `server/src/services/plugin-host-services.ts` | Add `issueDocuments` service delegating to existing `documents.ts` |
| `packages/plugins/sdk/src/testing.ts` | Add stub `documents` client for test context |

No new dependencies. No lockfile changes.

## Testing

- `pnpm -r typecheck` — passes
- `pnpm test:run` — 311 tests pass (74 files)
- `pnpm build` — passes

## Usage

```ts
// List documents
const docs = await ctx.issues.documents.list(issueId, companyId);

// Get a specific document
const doc = await ctx.issues.documents.get(issueId, "plan", companyId);

// Create or update
await ctx.issues.documents.upsert({
  issueId,
  key: "design-spec",
  body: "# Design\n...",
  companyId,
  title: "Design Specification",
});

// Delete
await ctx.issues.documents.delete(issueId, "old-doc", companyId);
```

Requires `issue.documents.read` capability for list/get, `issue.documents.write` for upsert/delete.

Fixes #940

[...]

Issue #6215, opened 2026-05-17
plugin-tool-dispatcher uses pluginKey (string) instead of pluginDbId (UUID) — every manifest-declared tool fails with "worker not running"


andreiyurkevich:
# Upstream issue draft — paperclipai

Paste this into a new issue at https://github.com/paperclipai/paperclip/issues/new

---

**Title:** `plugin-tool-dispatcher uses pluginKey (string) instead of pluginDbId (UUID) — every manifest-declared tool fails with "worker not running"`

**Body:**

## Summary

`paperclipai plugin tools/execute` always returns `Cannot execute tool "<plugin>:<tool>" — worker for plugin "<plugin>" is not running.` for any plugin that declares tools in its manifest, even when the worker is healthy and the tools are registered in the registry.

## Repro

1. Install a plugin that declares `tools[]` in its manifest (`apiVersion: 1`, capability `agent.tools.register`).
2. Confirm the worker is up: `paperclipai plugin list` shows `status=ready`, server logs show `worker process started and initialized` and `registered N tool(s) for plugin`.
3. `GET /api/plugins/tools` returns the tools correctly.
4. `POST /api/plugins/tools/execute` with a valid `runContext` → 502, log message: `worker for plugin "<key>" is not running.`

## Root cause

`@paperclipai/server/dist/services/plugin-loader.js`:

```js
toolDispatcher.registerPluginTools(pluginKey, manifest);
```

doesn't forward the plugin's database UUID, and `@paperclipai/server/dist/services/plugin-tool-dispatcher.js`:

```js
registerPluginTools(pluginId, manifest) {
    registry.registerPlugin(pluginId, manifest);
},
```

doesn't accept one. The registry then stores `pluginDbId = pluginKey` (string slug). At dispatch time the dispatcher calls `workerManager.isRunning(tool.pluginDbId)` — but `PluginWorkerManager` indexes workers by **UUID**, so the lookup always returns `false`.

## Fix

```diff
--- a/services/plugin-loader.js
+++ b/services/plugin-loader.js
-                toolDispatcher.registerPluginTools(pluginKey, manifest);
+                toolDispatcher.registerPluginTools(pluginKey, manifest, pluginId);

--- a/services/plugin-tool-dispatcher.js
+++ b/services/plugin-tool-dispatcher.js
-        registerPluginTools(pluginId, manifest) {
-            registry.registerPlugin(pluginId, manifest);
[...]

PR #6219, opened 2026-05-17
fix(plugins): forward pluginDbId from tool dispatcher to registry

easton-ring:
## TL;DR

`POST /api/plugins/tools/execute` returns `Cannot execute tool ... — worker for plugin "<key>" is not running` for every plugin, even when `GET /api/plugins/<key>/dashboard` reports the worker as healthy with a live pid. Bug survives container restart and `disable/enable` / `uninstall/install` cycles. Root cause is in `PluginToolDispatcher.registerPluginTools` not forwarding the plugin's DB UUID to the underlying `PluginToolRegistry`. Three-line fix + new test.

## Reproduction

1. Install any plugin that registers tools (e.g. `@acme/linear`).
2. Wait for plugin to reach `status: "ready"`. Verify worker is alive via `GET /api/plugins/<key>/dashboard` — it reports `worker.status: "running"` with a live pid.
3. Invoke any of the plugin's tools via `POST /api/plugins/tools/execute`:
   ```json
   {"tool": "acme.linear:list-issues", "parameters": {...}, "runContext": {...}}
   ```
4. Server responds with `400`: `Cannot execute tool "acme.linear:list-issues" — worker for plugin "acme.linear" is not running.`

This contradicts the dashboard endpoint's view of the same worker. Reproduces against every plugin and survives a full server process restart.

## Root cause

`PluginToolRegistry.registerPlugin` takes an optional third `pluginDbId` argument and falls back to `dbId = pluginDbId ?? pluginId` when it isn't provided. The first arg, `pluginId`, is the plugin's **package key** (e.g. `"acme.linear"`) — used as the namespace prefix in tool names and as the unregister key. The third arg is intended to be the plugin's **DB row UUID**, which is what `PluginWorkerManager.workers` Map is keyed on (since `plugin-loader.activatePlugin` starts the worker via `workerManager.startWorker(plugin.id, ...)`).

There are three call sites to `registry.registerPlugin` inside `plugin-tool-dispatcher.ts`:

| Line | Path | Passes DB UUID? |
| --- | --- | --- |
| 269 | `registerFromDb` (called from lifecycle event handler) | ✅ Yes — passes `plugin.id` |
| 333 | Dispatcher init (boot-time scan over `status='ready'` plugins) | ✅ Yes — passes `plugin.id` |
| 433 | `registerPluginTools` (called by `plugin-loader.activatePlugin`) | ❌ **No** — only two args |

The init path runs at boot and registers tools correctly. But then `plugin-loader.loadAll()` calls `activatePlugin` for each plugin, which calls `toolDispatcher.registerPluginTools(pluginKey, manifest)` — the buggy two-arg path. `registry.registerPlugin` clears the previous registration via `removePluginTools` and re-registers with `tool.pluginDbId = packageKey` (the fallback). The buggy path always runs last, so a restart doesn't help.

At execute time, `plugin-tool-registry.ts:406`:

```typescript
const dbId = tool.pluginDbId;       // <— package key, not DB UUID
if (!workerManager.isRunning(dbId)) {  // <— workers Map is keyed by UUID
  throw new Error(...);             // <— this fires for every plugin
}
```

This is why the dashboard endpoint (which uses `wm.getWorker(plugin.id)` — looking up by UUID directly) sees the worker, but `tools/execute` doesn't.

## Fix

Three changes, ~10 LoC excluding the test file:

- `server/src/services/plugin-tool-dispatcher.ts` — add `pluginDbId: string` as a required third parameter on both the interface and impl of `registerPluginTools`; forward it to `registry.registerPlugin(pluginId, manifest, pluginDbId)`. JSDoc updated to document the parameter ambiguity (the first arg's name `pluginId` historically refers to the package key here, not the DB UUID).
- `server/src/services/plugin-loader.ts:1935` — the sole caller passes the `pluginId` local variable (the DB UUID, set on line 1768 as `plugin.id`).

The `registry.registerPlugin` API is unchanged. Other call paths (`registerFromDb`, dispatcher init) were already passing the DB UUID correctly.

## Testing

New file `server/src/__tests__/plugin-tool-dispatcher.test.ts` with two tests:

1. **`routes workerManager.isRunning by the plugin DB id, not the package key`** — exercises the full register-then-execute flow with a stubbed `workerManager` that returns `true` only when called with the DB UUID. Before the fix this test fails with the exact production error string. After the fix it passes.
2. **`listTools exposes the plugin DB id, not the package key, as the descriptor pluginId`** — confirms `toAgentDescriptor` returns the correct `pluginId` field for consumers (e.g. agent service) that look it up.

Existing plugin tests verified clean:

```
Test Files  16 passed (16)
Tests       111 passed (111)
```

No other tests modified. No production callers of `registerPluginTools` other than the one site in `plugin-loader.ts`.

## Risk

- **API change:** `registerPluginTools` gains a required third parameter. The dispatcher interface is exported, but a code search across the server source found the single caller in `plugin-loader.activatePlugin`. Downstream consumers using the package would see a TypeScript compile error pointing them at the change; runtime behavior for callers correctly passing the DB UUID is unchanged.
- **Behavioral change:** Every plugin tool invocation now actually reaches the worker instead of failing pre-flight. This is the intended fix.
- **Migration:** None. No data shape changes, no DB migration, no manifest format changes. The fix applies on first server restart after deploy.

## Discovered downstream

The bug surfaced during an rf-collective adoption pilot for the `host.issueDocuments.upsert` methods introduced in #1074. The pilot plugin's tool registered fine (`plugin-loader` logged `"capabilities":["agent.tools.register","issue.documents.write"]`) but invocation always failed pre-flight at `workerManager.isRunning(dbId)`. Operator-side workarounds (server restart, plugin disable/enable, uninstall/reinstall) all failed because the loader-driven registration path always runs after them and overwrites any correct registration. Side note: a control case against a completely separate plugin (`rf-collective.iqengine:iqengine-status`) reproduced identically, confirming the bug is platform-wide rather than pilot-specific.

### DIFF
--- a/server/src/__tests__/plugin-tool-dispatcher.test.ts
+++ b/server/src/__tests__/plugin-tool-dispatcher.test.ts
@@ -1,0 +1,96 @@
+import { describe, expect, it, vi } from "vitest";
+import { createPluginToolDispatcher } from "../services/plugin-tool-dispatcher.js";
+import type { PaperclipPluginManifestV1 } from "@paperclipai/plugin-sdk";
+import type { ToolRunContext } from "@paperclipai/plugin-sdk";
+
+const PLUGIN_KEY = "test.plugin";
+const PLUGIN_DB_ID = "9c1b8f25-1cee-4f0c-8c39-7f1ed27e1234";
+
+function makeManifest(): PaperclipPluginManifestV1 {
+  return {
+    id: PLUGIN_KEY,
+    apiVersion: 1,
+    version: "0.0.1",
+    displayName: "Test Plugin",
+    description: "Test plugin for dispatcher unit tests",
+    author: "tests",
+    categories: ["automation"],
+    capabilities: ["agent.tools.register"],
+    entrypoints: { worker: "./worker.js" },
+    tools: [
+      {
+        name: "do-thing",
+        displayName: "Do Thing",
+        description: "Does a thing",
+        parametersSchema: { type: "object", properties: {} },
+      },
+    ],
+  } as PaperclipPluginManifestV1;
+}
+
+function makeRunContext(): ToolRunContext {
+  return {
+    agentId: "agent-a",
+    runId: "run-r",
+    companyId: "company-c",
+    projectId: "project-p",
+  };
+}
+
+describe("plugin-tool-dispatcher", () => {
+  it("routes workerManager.isRunning by the plugin DB id, not the package key", async () => {
+    // Stub that "recognizes" the worker ONLY when looked up by DB UUID,
+    // mirroring plugin-worker-manager's behavior where the workers Map is
+    // keyed by the DB UUID (set in plugin-loader's activatePlugin via
+    // `workerManager.startWorker(plugin.id, ...)`).
+    const isRunning = vi.fn((id: string) => id === PLUGIN_DB_ID);
+    const call = vi.fn(async (_id: string, _method: string, _params: unknown) => ({
+      content: "ok",
+    }));
+
+    const workerManager = {
+      isRunning,
+      call,
+      // Other methods unused by this code path; type-cast away the mismatch.
+    } as unknown as Parameters<typeof createPluginToolDispatcher>[0]["workerManager"];
+
+    const dispatcher = createPluginToolDispatcher({ workerManager });
+
+    // Mimic the plugin-loader call site: caller has both package key AND DB UUID.
+    // The dispatcher MUST persist the DB UUID so executeTool can later look up
+    // the worker correctly.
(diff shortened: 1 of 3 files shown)


easton-ring:
Closing in favor of #6215, which has the same root-cause diagnosis and the same two-line fix sketched out by @andreiyurkevich (who offered to send a PR). The fix branch is preserved at https://github.com/techring-live/paperclip/tree/fix/plugin-tool-dispatcher-dbid-passthrough if useful — it adds JSDoc clarification on the parameter ambiguity and a regression test (`server/src/__tests__/plugin-tool-dispatcher.test.ts`) that fails with the exact production error string before the fix. Happy to fold those into whatever PR ultimately lands.

PR #6223, opened 2026-05-18
fix(checkout): clear stale checkoutRunId in clearExecutionRunIfTerminal

roy493:
## Thinking Path

> - Paperclip orchestrates AI agents via heartbeat runs; each run checks out issues to prevent concurrent mutation.
> - The checkout lock is stored as two DB fields: `checkoutRunId` (agent-side ownership) and `executionRunId` (execution-side ownership), always set to the same run ID.
> - When a run terminates abnormally (budget limit, crash, upstream failure) without calling `/release`, both fields retain the stale run ID.
> - `clearExecutionRunIfTerminal()` is called at the start of every checkout/release/assert to recover from this. It checks if `executionRunId` references a terminal or missing run and clears it — but it only cleared `executionRunId`, leaving `checkoutRunId` stale.
> - The subsequent checkout WHERE clause requires `checkoutRunId IS NULL OR checkoutRunId = newRunId`; with a stale non-null value it fails. `adoptStaleCheckoutRun` could recover this but only when `status = 'in_progress'` — it fails silently for `blocked`/`todo`/`in_review` issues, and the 409 is thrown.
> - This PR fixes `clearExecutionRunIfTerminal` to also clear `checkoutRunId` in the same UPDATE, and to handle the orphaned case where `executionRunId` was already null but `checkoutRunId` is still stale.

## What Changed

- `server/src/services/issues.ts` — `clearExecutionRunIfTerminal`: select both `checkoutRunId` and `executionRunId`; use either to probe the heartbeat run status; clear both in the UPDATE; handle the orphaned case via a conditional WHERE predicate.

## Verification

```
pnpm typecheck   # passes — no type errors
pnpm test        # 90 tests pass; 1 pre-existing failure in workspace-restore-merge unrelated to this change
```

Manually: the bug manifested as COO agent getting 409 on every checkout after a run failed due to Claude usage limits. After this fix, `clearExecutionRunIfTerminal` will atomically clear the stale `checkoutRunId` so the next heartbeat's checkout sees clean locks and succeeds.

## Risks

Low. The change is additive within an already-transactional function: it reads one extra field and clears one extra column in an UPDATE that is already gated on a row-level FOR UPDATE lock. No new write paths or state machine transitions. The WHERE predicate on the UPDATE is conservative — it only fires when the lock run is terminal or missing.

## Model Used

- Provider: Anthropic
- Mode: Tool use + code editing via Claude Code CLI
- Context: 1M context window

## Checklist

### DIFF
--- a/server/src/__tests__/auth-run-id-coercion.test.ts
+++ b/server/src/__tests__/auth-run-id-coercion.test.ts
@@ -1,0 +1,90 @@
+import express from "express";
+import request from "supertest";
+import { describe, expect, it } from "vitest";
+import { actorMiddleware, coerceRunId } from "../middleware/auth.js";
+
+function createDb() {
+  return {} as any;
+}
+
+function createApp() {
+  const app = express();
+  app.use(
+    actorMiddleware(createDb(), {
+      deploymentMode: "local_trusted",
+    }),
+  );
+  app.get("/actor", (req, res) => {
+    res.json({ runId: req.actor.runId ?? null });
+  });
+  return app;
+}
+
+describe("actorMiddleware X-Paperclip-Run-Id coercion (header path)", () => {
+  it("passes through a valid UUID run id unchanged", async () => {
+    const app = createApp();
+    const validUuid = "11111111-2222-4333-8444-555555555555";
+
+    const res = await request(app)
+      .get("/actor")
+      .set("x-paperclip-run-id", validUuid);
+
+    expect(res.status).toBe(200);
+    expect(res.body).toEqual({ runId: validUuid });
+  });
+
+  it("coerces a malformed run id header to null so downstream UUID columns receive null instead of 500ing", async () => {
+    const app = createApp();
+    // Real example seen in NOY-6892 forensics — agent constructed a slug-style id.
+    const malformed = "noy6809-monitor-1778938148";
+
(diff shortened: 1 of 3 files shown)


tdi-ai-create:
Production impact report -- Teachers Deserve It (TEA)                                                                                                                                                 
  ▎
  ▎ We're running Paperclip on Railway with 28 agents. This exact bug has caused a company-wide outage tracked as our internal TEA-6873 ("Paperclip Crash").                                              
  ▎                                                                                                                                                                                                     
  ▎ What we see: After POST /api/issues/{id}/checkout {takeover: true, force: true} returns 200, the persisted checkoutRunId is a server-generated UUID, not the actor's JWT run_id. Every subsequent     
  ▎ PATCH returns 409 "Issue run ownership conflict" because actorRunId != checkoutRunId.                                                                                                               
  ▎                                                                                                                                                                                                       
  ▎ Impact: 332 issues stuck in blocked, only 2 in in_progress. Agents can comment but cannot change status, priority, or assignee. The 2-hour restart cron makes it worse by killing in-flight runs      
  ▎ without releasing checkouts.
  ▎                                                                                                                                                                                                       
  ▎ This PR is exactly what we need. Please prioritize merging it.

PR #6238, opened 2026-05-18
LET-393 Phase 4A-S4 follow-up — live E2B usage-API SourceA adapter

lmanualm:
## Summary

Implements the live `SourceA` adapter for the sandbox billing-cap monitor against the E2B usage / metering API, wiring it into `app.ts` behind two env vars. Default deployments stay on the legacy `sourceA: null` behaviour from LET-367.

- New `createE2BUsageApiSourceA()` factory in `server/src/services/sandbox/billing-cap/e2b-usage-api-source-a.ts`.
  - GET-only (never issues a billable POST/PUT/DELETE), `X-API-Key` header — matches B1 transport convention.
  - Falls back to `null` on 401 / 403 / 404, on empty / non-JSON / non-object bodies, on payloads missing day/month cents, and on credential-shaped fields (defence in depth on top of the `resolveSpend()` redactor).
  - Throws on 429 / 5xx so the coalescer's catch arm fires.
  - Accepts `{day_cents,month_cents}`, camelCase, and nested `{day:{amount_cents|cents}}` response shapes.
- Wired in `app.ts`: `sourceA` resolves from `SANDBOX_E2B_USAGE_API_URL` + `SANDBOX_E2B_USAGE_API_AUTH`; unset either ⇒ `sourceA: null` exactly as before.

## Test plan

- [x] `server/src/services/sandbox/billing-cap/e2b-usage-api-source-a.test.ts` — 21 cases (factory env-gate, happy path, 401/403/404 fallback, credential-shape fallback, empty/non-JSON body, alternate response shapes, request shape, 429/500 throw).
- [x] Full billing-cap suite: 71 tests pass (`vitest run server/src/services/sandbox/billing-cap/`).
- [x] `tsc --noEmit` clean on `server`.
- [ ] No production deploy / restart / migration apply / live E2B spend in this PR.

## Endpoint TBD

The E2B metering endpoint path/shape is still TBD with the vendor. The adapter is intentionally configurable end-to-end via env so we can wire the confirmed URL + API key in a future change without another code release. Unset env vars keep the legacy `sourceA: null` wiring.


lmanualm:
Closing — opened against the wrong upstream. Per project policy, EAOS work targets the fork repo lmanualm/paperclip. The LET-393 adapter PR will be re-opened against lmanualm/paperclip:master (stacked on enterprise-agent-os/LET-392) so QA can review the focused merge artifact.

PR #6240, opened 2026-05-18
fix(cli): stop worktree init --force from wiping repo worktrees/

devinfoley:
## Thinking Path

> - Each working tree gets an isolated Paperclip instance; the CLI's `paperclip worktree init` is what bootstraps that instance and writes `<repo>/.paperclip/config.json` + `.env`
> - When `--force` was passed, the init path tried to "start clean" by recursively removing the entire `<repo>/.paperclip/` directory before rewriting those two files
> - But `<repo>/.paperclip/` also holds `worktrees/`, which contains every repo-managed worktree checkout (70+ on this machine). The recursive rm silently nuked all of them.
> - This PR narrows the `--force` reset so it only deletes the two files it's about to rewrite (`config.json`, `.env`), instead of wiping the whole `repoConfigDir`
> - It also adds a regression test that drops a sentinel file into `<repoConfigDir>/worktrees/` and asserts it survives a `--force` init
> - The benefit is that `worktree init --force` becomes safe to run from inside the main repo without destroying every sibling worktree checkout

## What Changed

- `cli/src/commands/worktree.ts`: in the `--force` branch of `runWorktreeInit`, replace `rmSync(paths.repoConfigDir, { recursive: true, force: true })` with targeted removals of `paths.configPath` and `paths.envPath`. `paths.instanceRoot` removal is unchanged — that path is per-instance and safe to wipe.
- `cli/src/__tests__/worktree.test.ts`: new regression test that seeds a fake `worktrees/<name>/` checkout inside the repo's `.paperclip/` and verifies `runWorktreeInit({ force: true, ... })` does not delete it.

## Verification

- `pnpm --filter @paperclip/cli test -- worktree` — the new regression test fails on the old code and passes on the fix
- Manual: from a repo checkout, `npx paperclipai worktree init --force …` no longer removes `<repo>/.paperclip/worktrees/`; only `config.json` and `.env` are rewritten

## Risks

- Low. The change strictly narrows what `--force` removes. Any caller that depended on `--force` also wiping unrelated files under `<repo>/.paperclip/` (there shouldn't be any — it's documented as just config + env) would see those files persist. `instanceRoot` cleanup is unchanged.

## Model Used

- Claude (Anthropic), model `claude-opus-4-7`, ~200K context, extended-thinking + tool-use enabled, run via the Paperclip `claude_local` adapter.

## Checklist

- [ ] If this change affects the UI, I have included before/after screenshots (N/A — CLI-only fix)
- [x] I have updated relevant documentation to reflect my changes (no doc surface affected)

### DIFF
--- a/cli/src/__tests__/worktree.test.ts
+++ b/cli/src/__tests__/worktree.test.ts
@@ -512,6 +512,45 @@ describe("worktree helpers", () => {
     }
   });
 
+  it("preserves repo-managed worktree checkouts when --force re-runs from the source repo", async () => {
+    const tempRoot = fs.mkdtempSync(path.join(os.tmpdir(), "paperclip-worktree-force-preserve-"));
+    const repoRoot = path.join(tempRoot, "repo");
+    const originalCwd = process.cwd();
+
+    try {
+      fs.mkdirSync(repoRoot, { recursive: true });
+      const repoConfigDir = path.join(repoRoot, ".paperclip");
+      fs.mkdirSync(repoConfigDir, { recursive: true });
+      fs.writeFileSync(path.join(repoConfigDir, "config.json"), "stale", "utf8");
+      fs.writeFileSync(path.join(repoConfigDir, ".env"), "STALE=1", "utf8");
+
+      // Simulate the repo-managed worktrees subfolder that holds every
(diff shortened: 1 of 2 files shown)


devinfoley:
@greptile review

PR #6298, opened 2026-05-18
fix(ui): LET-416 secrets dialog UX — mask value, surface inline errors, redact path

lmanualm:
## Summary

Fixes the four secrets-dialog UX/security bugs LET-416 (triaged out of LET-365):

1. **Value field masked by default.** Replaces the plain-text Textarea with `<Input type=\"password\">` plus an opt-in eye-toggle (Show/Hide). Applied to Create-secret and Rotate-secret dialogs. Stops the typed/pasted value from leaking into screenshots and screen-shares.
2. **Inline per-field validation errors.** Parses Zod issues from the API 400 `details[]` array and renders them next to the offending field. Generic \"Validation error\" with no detail is gone.
3. **Key regex hint.** Adds a default helperText describing the allowed alphabet and a client-side validator that mirrors the server regex. `key` regex failures are mapped to a user-friendly hint with an example (`sandbox.e2b.apiKey.pilot`).
4. **Master-key path leak.** Default render redacts Unix absolute paths in provider health / vault warning text (`/tmp/.../master.key` → `~/…/master.key`). The full path is reachable only via an opt-in \"Show details\" toggle.

## Files

- `ui/src/pages/Secrets.tsx` — masked Value field with toggle, inline field errors, Key hint + client-side validation, path-redacting health summary / vault card / vault inline warning.
- `ui/src/pages/Secrets.test.ts` — unit tests for `redactAbsolutePathsInMessage`, `validateSecretKeyClient`, and the shared `SECRET_KEY_HINT` constant.
- `ui/src/pages/Secrets.render.test.tsx` — render tests covering default-masked Value, show/hide toggle, slash-in-key inline hint, and provider health text with no path leak (plus a Show-details reveal flow).

## Test plan

- [x] `pnpm typecheck` (UI workspace) — passes.
- [x] `pnpm exec vitest run src/pages/Secrets.test.ts src/pages/Secrets.render.test.tsx` — 19/19 pass.
- [x] Visual verification of the four acceptance criteria via render tests:
  - Create-secret Value input renders with `type=\"password\"` by default.
  - Show toggle flips `type` to `\"text\"`; Hide flips back.
  - Typing `foo/bar` in Key surfaces the inline slash-rejection hint with the example.
  - Provider health text never includes `/tmp/...` in default render; Show details opt-in reveals it.
- [ ] Reviewer: manual screenshot diff on Company Settings → Secrets → Create secret (E2B / local-encrypted path).

## Out of scope

- Server-side key regex relaxation (separate decision; current default keeps strict).
- The vitest temp-dir runtime config issue surfaced in the LET-365 screenshot (separate runtime investigation).
- Rotating Andrii's E2B key for LET-365 (handled on the LET-365 thread).
- A pre-existing `act is not a function` flake in `ImportFromVaultDialog.test.tsx` under `NODE_ENV=production` is untouched (not in this bug's scope; the same hoist-fix used here is the path forward).


lmanualm:
Wrong base repo — reopening against lmanualm:master (the EAOS-tracked branch where LET-* PRs land).

PR #6329, opened 2026-05-18
fix(server,shared): LET-436 classify adapter_process_lost_no_pid + log dir safety

lmanualm:
## Summary

LET-434 parent / LET-436 Lane A. Stops the high-volume `process_lost` /
`adapter_failed` flood produced by Hermes-local runs and similar
tracked-local adapters that bypass `runChildProcess`'s `onSpawn` callback
(e.g. `hermes-paperclip-adapter@0.2.0`).

- Adds explicit `adapter_process_lost_no_pid` terminal classification
  when a tracked-local run dies with no recorded pid/process group, so
  operators can distinguish "adapter never persisted metadata" from a
  real OS-level process loss or a server restart.
- Hardens both outer adapter-failure catch paths so a late adapter
  rejection cannot overwrite the more specific process-loss family
  terminal with a generic `adapter_failed`.
- Adds `detectInsecureLogDir` in `@paperclipai/shared` so a production
  deployment configured against `/tmp/paperclip-vitest-*` fails loudly
  at startup instead of silently losing logs.
- Pure helpers live in `heartbeat-stop-metadata.ts` and
  `config-schema.ts` so they unit-test without spinning up Postgres.

## Hard gates honoured

- No production deploy / restart / migration apply / secret rotation /
  spend / live vendor call.
- Branch: `eaos/let-434-runtime-process-metadata`.
- Andrii's max-6-active-child rule respected (no new child workstreams
  created).

## Test plan

- [x] `pnpm exec vitest run server/src/services/heartbeat-stop-metadata.test.ts packages/shared/src/config-schema.test.ts` — 20/20 pass (RED → GREEN).
- [x] `pnpm exec vitest run server/src/__tests__/run-liveness.test.ts` — 18/18 pass (existing `process_lost` routing preserved by substring match).
- [x] `node_modules/.bin/tsc -p server/tsconfig.json --noEmit` — exit 0.
- [x] `node_modules/.bin/tsc -p packages/shared/tsconfig.json --noEmit` — exit 0.
- [ ] Heavy `heartbeat-process-recovery.test.ts` (embedded Postgres) skipped on the low-memory VPS — recommend Claude Reviewer or QA Validator runs it in CI before merge.

## Risk / rollback

- Adds a new errorCode (`adapter_process_lost_no_pid`) and a new
  `HeartbeatRunStopReason` enum value. Downstream consumers that
  exhaustively switch on stop reasons may need to widen; the
  `run-liveness` routing already matches via the `process_lost`
  substring regex.
- The outer-catch guard adds one extra `getRun` lookup per terminal
  failure path. Negligible.
- Rollback: `git revert <sha>`. The change is additive and gated on
  `tracksLocalChild`, so reverting restores the previous classification
  immediately without data fixups.

Refs: parent LET-434, impl LET-436.


lmanualm:
Closing as non-canonical / wrong base. This PR was opened against upstream `paperclipai/paperclip:master` while the branch is forked from `lmanualm/paperclip:master`, so the diff is scope-inflated (314 files / +63206/-508) and `mergeable: CONFLICTING` / `mergeState: DIRTY`. The canonical EAOS-fork PR for LET-436 is https://github.com/lmanualm/paperclip/pull/78 (base: lmanualm/paperclip:master, head: eaos/let-434-runtime-process-metadata, 7 files / +289/-45). Do not treat #6329 as merge evidence.

PR #6331, opened 2026-05-18
fix(opencode-local, CHRA-1272): models discovery timeout, dedup calls, iOS lane concurrency

chrislro:
## Summary
This PR bundles three independent fixes that were all required to unblock concurrent iOS agent runs on the Mac Mini.

### 1. CHRA-1280 — opencode models discovery timeout
- Local `opencode models` wall-clock varies 14.5–25s+ on at least one Mac Mini host; the hard-coded 20s adapter timeout fires on ~1-in-3 cold-cache runs and silently fails every `opencode_local` agent (Engineer, Kimi-Coder, DevOps on this deployment).
- Triples `MODELS_DISCOVERY_TIMEOUT_MS` to 60s to cover the observed 25s tail with headroom.
- Stretches `MODELS_CACHE_TTL_MS` from 60s to 10 min so consecutive agent runs don't repeatedly hit the cold-path discovery call (the model list is stable).

### 2. CHRA-1282 — deduplicate concurrent opencode models discovery calls
- Under concurrent agent startup, multiple `opencode models` processes race, each timing out and retrying, amplifying the failure rate.
- Adds a `pendingDiscoveries` Map so that overlapping calls within the same process share a single promise. The first call performs discovery; subsequent concurrent waiters receive the same result. Cleanup via `.finally()` prevents memory leaks.

### 3. CHRA-1272 — iOS lane concurrency deadlock prevention
- **Per-issue AGENT_NAME (all adapters):** `buildPaperclipEnv` now auto-generates `AGENT_NAME` as `agent.name + runId` (first 8 chars), preventing cache collisions across concurrent issues for the same agent. Updated all 11 adapter `execute.ts` files to pass `runId` through.
- **iOS lane-cap enforcement (server):** Added checkout-time guard in `issues.ts` that blocks checking out an iOS-labelled issue when another iOS issue is already `in_progress`. Checked labels: `ios`, `iphone`, `mobile`, `shifts iphone app`, `plantonistapro`. Only 1 concurrent iOS issue allowed per company.
- **Sim resolver determinism (iOS repo):** The simulator resolver script now checks if a booted simulator is already in use by an `xcodebuild` process (via `pgrep`). It prefers free booted sims; if all booted sims are in use, it falls back to a shutdown sim. Prevents two concurrent `xcodebuild` invocations from picking the same simulator and deadlocking. *(Delivered in companion PR on `chrislro/PlantonistaPro`.)*

## Test plan
- [ ] `pnpm -F @paperclipai/opencode-local build` succeeds.
- [ ] Restart local Paperclip server with new bundle.
- [ ] One previously-failing opencode_local agent run completes (Engineer / DevOps on an iOS sweep issue).
- [ ] Two concurrent iOS heartbeats do not deadlock on simulator selection.

Refs CHRA-1280, CHRA-1282, CHRA-1272.

### DIFF
--- a/packages/adapter-utils/src/server-utils.ts
+++ b/packages/adapter-utils/src/server-utils.ts
@@ -890,7 +890,10 @@ export function buildInvocationEnvForLogs(
   return redactEnvForLogs(merged);
 }
 
-export function buildPaperclipEnv(agent: { id: string; companyId: string }): Record<string, string> {
+export function buildPaperclipEnv(
+  agent: { id: string; companyId: string; name?: string },
+  runId?: string,
+): Record<string, string> {
   const resolveHostForUrl = (rawHost: string): string => {
     const host = rawHost.trim();
     if (!host || host === "0.0.0.0" || host === "::") return "localhost";
@@ -901,6 +904,13 @@ export function buildPaperclipEnv(agent: { id: string; companyId: string }): Rec
     PAPERCLIP_AGENT_ID: agent.id,
     PAPERCLIP_COMPANY_ID: agent.companyId,
   };
+  if (agent.name) {
+    vars.PAPERCLIP_AGENT_NAME = agent.name;
+  }
+  if (agent.name && runId) {
+    const shortRunId = runId.length > 8 ? runId.slice(0, 8) : runId;
+    vars.AGENT_NAME = `${agent.name}-${shortRunId}`;
+  }
   const runtimeHost = resolveHostForUrl(
     process.env.PAPERCLIP_LISTEN_HOST ?? process.env.HOST ?? "localhost",
   );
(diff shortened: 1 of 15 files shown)


chrislro:
CTO verdict for HEAD 7d7255e1f7170bc0fa235e516b4e58b0d2095218: REQUEST_CHANGES.

Blocking findings:
1. `server/src/services/issues.ts` broadens the iOS lane cap to any issue labeled `mobile`, so generic mobile/Android work can be blocked by an unrelated iOS issue.
2. `packages/adapters/openclaw-gateway/src/server/execute.ts` still calls `buildPaperclipEnv(ctx.agent)` without `ctx.runId`, so the per-run `AGENT_NAME` fix is not consistently applied across adapters.

All current CI checks are green on this head, but these two correctness issues still block merge/deploy gating.

PR #6387, opened 2026-05-19
fix(adapter-claude-local): restore settingsContents in prompt bundle key and execute

azainulbhai:
## Summary

The `2026.517.0` release dropped `settingsContents` from both `buildClaudePromptBundleKey` and `prepareClaudePromptBundle` in `prompt-cache.ts`, and from the `execute.ts` call site. This caused new pcache bundles to be written without a permission deny list in the bundle directory, creating a security regression (SEV-010).

The fix was present in `2026.416.0` (April release) but was lost in the May 17 release.

## Changes

**`prompt-cache.ts`**
- Add `settingsContents: string | null` parameter to `buildClaudePromptBundleKey`
- Include `settingsContents` in the SHA-256 hash after the skills loop
- Add `settingsContents: string | null` parameter to `prepareClaudePromptBundle`
- Write the agent settings file into the bundle directory when `settingsContents` is provided

**`execute.ts`**
- Read the agent settings file from `agentHome` before calling `prepareClaudePromptBundle`
- Pass it as `settingsContents` to `prepareClaudePromptBundle`

## Verification

`grep settingsContents` returns 8 hits in `prompt-cache.ts` and 1 hit in `execute.ts`.

The diff exactly matches the monkey-patch that was applied as an emergency workaround to local npx caches following the regression.

### DIFF
--- a/packages/adapters/claude-local/src/server/execute.ts
+++ b/packages/adapters/claude-local/src/server/execute.ts
@@ -455,10 +455,16 @@ export async function execute(ctx: AdapterExecutionContext): Promise<AdapterExec
       );
     }
   }
+  let agentSettingsContents: string | null = null;
+  if (agentHome) {
+    const agentSettingsPath = path.join(agentHome, ".claude", "settings.json");
+    agentSettingsContents = await fs.readFile(agentSettingsPath, "utf-8").catch(() => null);
+  }
   const promptBundle = await prepareClaudePromptBundle({
     companyId: agent.companyId,
     skills: claudeSkillEntries.filter((entry) => desiredSkillNames.has(entry.key)),
     instructionsContents: combinedInstructionsContents,
+    settingsContents: agentSettingsContents,
     onLog,
   });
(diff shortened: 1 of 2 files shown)


azainulbhai:
**CTO Review (agent: CTO)**

Reviewed the diff. The fix is correct and minimal: restores `settingsContents` in both the SHA-256 hash key (`prompt-cache.ts`) and the execute call site (`execute.ts`). This matches the emergency monkey-patch applied after the 2026.517.0 regression (SEV-010).

**CI status:** 9/14 checks passed (typecheck, e2e, workspace tests a+b, serialized suites 3/4 and 4/4, security/snyk, policy, Greptile). 5 checks have been stuck/pending in the GH Actions queue for 12+ hours, likely an infrastructure issue.

**Request:** A maintainer with merge permissions, please merge this so we can cut a release >= 2026.518.x. This unblocks ZOO-3777 (QA) and ZOO-3720 (parent fix).

PR #6396, opened 2026-05-19
adapter-claude-local: quarantine poisoned session jsonl + force fresh session on retry

blackhatci:
## Thinking Path

> - `@paperclipai/adapter-claude-local` wraps the Claude Code CLI to run those agents and persists per-`(cwd, prompt-bundle)` session state under `~/.claude/projects/<encoded-cwd>/<session-id>.jsonl`
> - Claude Code 2.1.131 occasionally appends a synthetic `assistant` entry to that file whose `id` is a plain UUID (not an Anthropic `msg_*` id); the next `/v1/messages` call sends that UUID as `diagnostics.previous_message_id` and Anthropic rejects with HTTP 400
> - The existing resume-retry path in `execute.ts` did not save us: (a) the Anthropic Agent SDK surfaces that 400 with subtype=success and exitCode 0, so the `exitCode !== 0` gate skipped the retry; and (b) even when the retry fired, `runAttempt(null)` only meant "don't pass `--resume`" — the CLI still wrote to / picked up the same project-dir jsonl, so the next attempt 400'd on the same tail
> - This pull request hardens the adapter end-to-end: drop the exitCode gate, extend the detector regex to catch the SDK 400, quarantine the poisoned jsonl aside before each invocation, and on retry force a brand-new `--session-id <uuid>` so the CLI cannot land on the same file
> - The benefit is that affected runs self-heal on the next attempt instead of needing manual `mv …jsonl …jsonl.poisoned` operator intervention, and we record `commandArgs` in `resultJson` so future regressions can be diagnosed without diffing jsonl tails

## What Changed

- `parse.ts` — extend `isClaudeUnknownSessionError` regex to also match the Anthropic Agent SDK 400 on `diagnostics.previous_message_id` (`must be the …`, `from a prior …`, and `400 … previous_message_id` variants).
- `execute.ts` — drop the `(initial.proc.exitCode ?? 0) !== 0` gate from the resume-retry condition; the detector regex is now the source of truth.
- `execute.ts` — add and export `quarantinePoisonedJsonl()`. Before invoking the CLI it locates the project-dir jsonl (preferring the explicit resume target, else the most recently modified `.jsonl` that is not already `.poisoned-*`), scans the last 60 lines for an `assistant` entry with `model: "<synthetic>"` and a non-`msg_*` id, and if found renames the file to `<path>.poisoned-<runId>`. Skipped when `executionTargetIsRemote` (remote targets manage their own `~/.claude`).
- `execute.ts` — extend `buildClaudeArgs` and `runAttempt` with a `freshSessionId` parameter. On the SDK-400 retry branch, generate a `randomUUID()` and pass `--session-id <uuid>` instead of just omitting `--resume`, so the CLI cannot reattach to the poisoned project-dir jsonl.
- `execute.ts` — thread `resumeSessionId` and `freshSessionId` through the attempt object into `toAdapterResult`, and record `commandArgs: { usedResume: boolean, sessionId: string | null }` in `mergedResultJson` for regression diagnostics.
- Tests — 6 new `quarantinePoisonedJsonl` tests (positive match by session id, clean file untouched, fallback to latest jsonl, `msg_*` id is safe even with `<synthetic>` model, missing project dir returns null, `.poisoned-*` files are ignored when picking the latest jsonl). 3 new `isClaudeUnknownSessionError` tests covering the classic CLI message, the SDK 400 in two forms, and a negative case.

## Verification

```bash
cd packages/adapters/claude-local
pnpm typecheck
pnpm exec vitest run
# 4 test files, 23 tests passed (was 14)
```

Manual exercise: starting from a project dir containing a `.jsonl` whose tail has `{"type":"assistant","message":{"id":"<uuid>","model":"<synthetic>", ...}}` (the exact failure shape we observed in production), the adapter now logs `[paperclip] Quarantined poisoned session jsonl "…" …` on entry, renames the file to `…jsonl.poisoned-<runId>`, runs the attempt with a fresh session id, and writes `commandArgs.usedResume = false` into `resultJson`.

## Risks

- Low/medium. The new code path runs once per invocation but only reads the *tail* of a single jsonl (≤60 lines parsed) and only renames a file in a strictly-positive match — never deletes. A failure inside the helper is caught and downgraded to `{ quarantined: false, error: … }` so the CLI invocation still proceeds.
- The retry branch now also fires on exitCode 0 when the SDK 400 detector matches. The detector regex is narrow (`previous_message_id` near `must be the` / `from a prior`, or `400 … previous_message_id`) so the additional firings are confined to the case this PR is targeting. A successful run that legitimately mentions `previous_message_id` in its tool output is theoretically possible but unlikely in practice; if it ever does, the only consequence is one extra Claude invocation with a fresh session id.
- Skipped entirely when `executionTargetIsRemote`, so remote runners that manage their own `~/.claude` are unaffected.

## Model Used

- Claude — `claude-opus-4-7` (1M context window). Authored end-to-end with extended thinking and tool use (filesystem edits, pnpm test execution).

## Checklist

- [ ] If this change affects the UI, I have included before/after screenshots — N/A (server-side adapter only)
- [x] I have updated relevant documentation to reflect my changes — N/A; no user-facing docs reference the resume-retry internals

### DIFF
--- a/packages/adapters/claude-local/src/server/execute.quarantine.test.ts
+++ b/packages/adapters/claude-local/src/server/execute.quarantine.test.ts
@@ -1,0 +1,201 @@
+import fs from "node:fs/promises";
+import path from "node:path";
+import os from "node:os";
+import { randomUUID } from "node:crypto";
+import { afterEach, beforeEach, describe, expect, it } from "vitest";
+
+import { quarantinePoisonedJsonl } from "./execute.js";
+
+const TMP_DIRS: string[] = [];
+
+async function makeTmpDir() {
+  const dir = await fs.mkdtemp(path.join(os.tmpdir(), "claude-local-quarantine-"));
+  TMP_DIRS.push(dir);
+  return dir;
+}
+
+afterEach(async () => {
+  while (TMP_DIRS.length > 0) {
+    const dir = TMP_DIRS.pop();
+    if (dir) await fs.rm(dir, { recursive: true, force: true });
+  }
+});
+
+function poisonedJsonl(poisonId: string) {
+  const good = JSON.stringify({
+    type: "assistant",
+    message: {
+      id: "msg_01abc",
+      model: "claude-opus-4-7",
+      role: "assistant",
+      content: [{ type: "text", text: "Hello" }],
+    },
+  });
+  const synthetic = JSON.stringify({
+    type: "assistant",
+    message: {
+      id: poisonId,
+      model: "<synthetic>",
+      role: "assistant",
+      stop_reason: "stop_sequence",
+      stop_sequence: "",
+      content: [{ type: "text", text: "No response requested." }],
+    },
+  });
+  return `${good}\n${synthetic}\n`;
+}
+
+function cleanJsonl() {
+  return (
+    JSON.stringify({
+      type: "assistant",
+      message: {
+        id: "msg_01good",
+        model: "claude-opus-4-7",
+        role: "assistant",
+        content: [{ type: "text", text: "All good." }],
+      },
+    }) + "\n"
+  );
+}
+
+describe("quarantinePoisonedJsonl", () => {
+  it("quarantines a poisoned jsonl matched by sessionId", async () => {
+    const tmpDir = await makeTmpDir();
+    const cwd = "/test/workspace/one";
+    const encodedCwd = cwd.replace(/\//g, "-");
+    const projectDir = path.join(tmpDir, "projects", encodedCwd);
+    await fs.mkdir(projectDir, { recursive: true });
+    const sessionId = randomUUID();
+    const poisonId = randomUUID();
+    const jsonlPath = path.join(projectDir, `${sessionId}.jsonl`);
+    await fs.writeFile(jsonlPath, poisonedJsonl(poisonId));
+
+    const runId = randomUUID();
+    const result = await quarantinePoisonedJsonl({
+      claudeConfigDir: tmpDir,
+      cwd,
+      sessionId,
+      runId,
+    });
+
+    expect(result?.quarantined).toBe(true);
+    expect(result?.poisonId).toBe(poisonId);
+    expect(result?.path).toBe(jsonlPath);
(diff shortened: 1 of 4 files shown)


blackhatci:
## Addressing Greptile review concerns

Pushed `122c113` to address concern #1; explaining concern #2 below.

### #1 — Post-quarantine `runAttempt(null)` lacks fresh `--session-id` ✅ fixed

Valid catch. After `quarantinePoisonedJsonl()` renames the poisoned file, the initial retry was previously called as `runAttempt(null)` with no session args. The CLI's no-arg path could still pick up a sibling `.jsonl` in the same project dir (multiple prompt bundles, prior runs in the same cwd), which would defeat the quarantine.

The fix is symmetric with the SDK-400 retry branch:

```ts
const postQuarantineFreshSessionId = quarantineResult?.quarantined ? randomUUID() : null;
// ...
const initial = await runAttempt(effectiveSessionId, postQuarantineFreshSessionId);
```

So `buildClaudeArgs` now appends `--session-id <fresh-uuid>` on the initial attempt whenever we just quarantined, mirroring what the SDK-400 retry already did. Quarantine log line now includes the fresh session id for diagnostics.

### #2 — Dropped `exitCode !== 0` gate widens retry surface

Acknowledged but accepted. The gate had to be dropped because the Anthropic Agent SDK 400 on `diagnostics.previous_message_id` surfaces with `subtype: "success"` and `exitCode: 0` — the whole reason this PR exists is that the previous gate caused us to *miss* those errors and never retry.

The detector regex (`packages/adapters/claude-local/src/server/parse.ts:200-204`) is reasonably tight:

```
no conversation found with session id
unknown session
session .* not found
previous_message_id .* (must be the|from a prior)
400 .* previous_message_id
```

None of these strings appear in a healthy `result` body, and the regex only runs against `parsed.result` plus extracted error messages — not arbitrary stdout. The false-positive surface is narrow enough that adding a belt-and-suspenders `(exitCode !== 0 || is_error)` check would risk re-introducing the original bug for marginal protection. Happy to add a tighter guard if you'd prefer one.

---

CI: all 14 checks green on `ea240cb` before this push. Will re-monitor on the new sha.

PR #6463, opened 2026-05-20
fix: unwrap DrizzleQueryError cause chain in isIssuePrefixConflict

### LINKED ISSUE #6462: createCompany returns 500 instead of retrying when issue_prefix collides (DrizzleQueryError unwrap)
## Summary

`POST /api/companies` returns a 500 instead of retrying with a different prefix whenever the derived 3-letter `issue_prefix` already exists. The intended fall-through to the next suffix (`TES` → `TESA` → `TESAA` …) never fires.

## Reproduction

1. Have a company whose `issue_prefix` is `TES` (e.g. one named "Test Corp").
2. `POST /api/companies` with `{ "name": "Test 2" }` (or anything whose first three uppercased letters are `TES`).
3. Response: HTTP 500.

Server log:

```
ERROR: POST /api/companies 500
DrizzleQueryError: Failed query: insert into "companies" ... params: Test 2,TES,0
caused by: PostgresError: duplicate key value violates unique constraint "companies_issue_prefix_idx"
  code: '23505'
  constraint: 'companies_issue_prefix_idx'
  at createCompanyWithUniquePrefix (server/src/services/companies.ts:146)
```

## Root cause

`server/src/services/companies.ts` defines `isIssuePrefixConflict(error)`, which is consulted by the retry loop in `createCompanyWithUniquePrefix`:

```ts
function isIssuePrefixConflict(error: unknown) {
  const constraint = …
    ? (error as { constraint?: string }).constraint
    : …
      ? (error as { constraint_name?: string }).constraint_name
      : undefined;
  return …
    && (error as { code?: string }).code === "23505"
    && constraint === "companies_issue_prefix_idx";
}
```

This check inspects properties directly on the thrown error. Since `drizzle-orm@0.45`, however, query errors are wrapped in a `DrizzleQueryError`:

```ts
// node_modules/drizzle-orm/errors.js
class DrizzleQueryError extends Error {
  constructor(query, params, cause) {
    super(`Failed query: ${query}\nparams: ${params}`);
    this.query = query;
    this.params = params;
    this.cause = cause;
    …
  }
}
```

The 23505 code and `constraint`/`constraint_name` live on `error.cause` (the underlying `PostgresError`), not on the outer wrapper. `isIssuePrefixConflict` therefore returns `false`, the `catch` block in `createCompanyWithUniquePrefix` re-throws, and the user sees a 500.

The retry loop itself works — it just never gets to attempt #2.

## Suggested fix

Walk the error's `cause` chain so the check matches under both old (raw `PostgresError`) and new (`DrizzleQueryError`-wrapping) shapes. Patch attached as a PR.

## Environment

- `drizzle-orm@0.45.2`
- `postgres@3.4.8`
- Server commit: upstream `master` as of `c91a0623`

harperaa:
Fixes #6462.

## Summary

- `isIssuePrefixConflict` in `server/src/services/companies.ts` only inspected the outer thrown error. Since `drizzle-orm@0.45` wraps query errors in a `DrizzleQueryError` whose `.cause` carries the real `PostgresError`, the check never matched a 23505 prefix collision and the retry loop in `createCompanyWithUniquePrefix` bailed on the first attempt — surfacing as a 500 on `POST /api/companies`.
- Walk the `cause` chain so the check works under both the old (raw `PostgresError`) and new (`DrizzleQueryError`-wrapped) shapes.

## Diff

One file, 28 lines:

```
server/src/services/companies.ts | 28 ++++++++++++++++++----------
1 file changed, 18 insertions(+), 10 deletions(-)
```

## Test plan

- [ ] Have two companies whose `issue_prefix` would collide (e.g. names "Test Corp" and "Test 2").
- [ ] `POST /api/companies` with the second name.
- [ ] Before: HTTP 500, `DrizzleQueryError … duplicate key value violates unique constraint "companies_issue_prefix_idx"` logged.
- [ ] After: HTTP 200, company created with prefix `TESA` (next suffix); subsequent collisions produce `TESAA`, `TESAAA`, etc.

The retry ceiling and prefix-derivation logic are untouched — this PR only restores the conditional that lets the existing retry loop see prefix conflicts again.

### DIFF
--- a/server/src/services/companies.ts
+++ b/server/src/services/companies.ts
@@ -125,16 +125,24 @@ export function companyService(db: Db) {
   }
 
   function isIssuePrefixConflict(error: unknown) {
-    const constraint = typeof error === "object" && error !== null && "constraint" in error
-      ? (error as { constraint?: string }).constraint
-      : typeof error === "object" && error !== null && "constraint_name" in error
-        ? (error as { constraint_name?: string }).constraint_name
-        : undefined;
-    return typeof error === "object"
-      && error !== null
-      && "code" in error
-      && (error as { code?: string }).code === "23505"
-      && constraint === "companies_issue_prefix_idx";
+    // drizzle-orm@0.45+ wraps the underlying postgres-js error in a
+    // DrizzleQueryError whose `.cause` carries the SQLSTATE code and the
+    // constraint name. Walking the cause chain keeps the retry loop in
+    // createCompanyWithUniquePrefix working under both old and new shapes.
+    for (let current: unknown = error; current; ) {
+      if (typeof current !== "object" || current === null) break;
+      const candidate = current as {
+        code?: string;
+        constraint?: string;
+        constraint_name?: string;
+        cause?: unknown;
+      };
+      const constraint = candidate.constraint ?? candidate.constraint_name;
+      if (candidate.code === "23505" && constraint === "companies_issue_prefix_idx") return true;
+      if (candidate.cause === current) break;
+      current = candidate.cause;
+    }
+    return false;
   }
 
   async function createCompanyWithUniquePrefix(data: typeof companies.$inferInsert) {


harperaa:
Closing as duplicate of #5850 (oldest, broadest scope — walks the .cause chain in all three call sites: isIssuePrefixConflict, isPluginKeyConflict, isOpenExecutionConflict). #5850 fixes the bug class; our PR only fixed the one site we hit. Apologies for the noise — should have searched the queue first. +1 to merge #5850.

PR #6466, opened 2026-05-20
feat(recovery): add executionPolicy.skipHandoff opt-out for permanent tracker issues

noahlaux:
## Thinking Path

> - Agents work recurring tracker issues (e.g. Daily Ops Digest parent YOU-261) that fire child execution issues on a schedule — the parent issue is never "done", it is perpetually in_progress
> - The `successfulRunHandoff` mechanism detects in_progress issues with no clear disposition after a successful run and queues a corrective handoff wake to prompt the agent to close out
> - For permanent tracker issues this is incorrect — the in_progress status is the intended steady state, not a missing disposition
> - There is no escape hatch: the only way to avoid the corrective wake is to mark the issue done or blocked, both of which destroy the tracker's operational contract
> - This pull request adds `executionPolicy.skipHandoff: boolean` — a first-class opt-out flag that makes `decideSuccessfulRunHandoff` immediately return `skip` when set
> - The benefit is permanent tracker issues can suppress the handoff escalation loop without changing their operational status, and future issues of the same class can be opted out at creation time without workarounds

## Problem or motivation

Permanent recurring tracker issues (e.g. the Daily Ops Digest parent YOU-261) are repeatedly pushed to `blocked` after every successful routine run. The `successfulRunHandoff` mechanism correctly detects that the issue has no clear next-step disposition — but for permanent scheduler issues, `in_progress` IS the intended steady state. There is no opt-out: marking the issue `done` or `blocked` breaks the tracker's operational contract. Every successful digest run produces a superfluous corrective handoff wake that blocks the parent, requiring manual intervention every time.

## Proposed solution

Add `executionPolicy.skipHandoff: boolean` — a first-class flag that makes `decideSuccessfulRunHandoff` return `{ kind: "skip" }` immediately when set on an issue. Issues with this flag never trigger the corrective handoff escalation, regardless of their current status or run disposition. Default is `false` so all existing issues are unaffected.

## Alternatives considered

- **Always skip for recurring issues**: would require tracking "is this a recurring-scheduler issue", which has no clean signal today.
- **Mark parent `done` after each run**: destroys the tracker's operational contract — the parent must stay `in_progress` to ensure child execution keeps firing.
- **Suppress handoff at the routine level**: handoff logic is in core recovery code and doesn't see routine context.

## Roadmap alignment

Not in ROADMAP.md — this is a targeted operational fix for the Daily Ops Digest loop, not a planned feature.

## What Changed

- **`packages/shared/src/types/issue.ts`**: add `skipHandoff?: boolean` to `IssueExecutionPolicy` interface
- **`packages/shared/src/validators/issue.ts`**: add `skipHandoff: z.boolean().optional().default(false)` to `issueExecutionPolicySchema`
- **`server/src/services/recovery/successful-run-handoff.ts`**: add `executionPolicy` to `IssueRow` Pick type; add early return `{ kind: "skip", reason: "issue execution policy opts out of handoff escalation" }` when `skipHandoff === true`
- **`server/src/services/issue-execution-policy.ts`**: update `normalizeIssueExecutionPolicy` to preserve `skipHandoff` and not return null when it is the only non-zero field (no stages, no monitor); update `stripMonitorFromExecutionPolicy` to preserve `skipHandoff` and not return null when `skipHandoff=true` and stages=0 after monitor is stripped
- **`server/src/services/heartbeat.ts`**: add `executionPolicy: issues.executionPolicy` to the DB select in `handleSuccessfulRunHandoff` so the field reaches the decision function
- **`server/src/services/recovery/successful-run-handoff.test.ts`**: 3 new test cases — `skipHandoff=true` skips, `skipHandoff=false` enqueues, `executionPolicy=null` enqueues

## Verification

```bash
# Unit tests — all 17 pass including 3 new cases
pnpm --filter=@paperclip/server test server/src/services/recovery/successful-run-handoff.test.ts

# Type-check
pnpm typecheck

# Build
pnpm build
```

All three commands pass cleanly.

**Migration note (post-deploy):** PATCH `executionPolicy: { skipHandoff: true }` on YOU-261 after this code deploys. The old Zod schema stripped unknown fields so the PATCH must wait for the new schema to be live.

## Risks

- **Low risk** — change is additive; default is `false` so all existing issues behave identically
- Zod schema uses `.optional().default(false)` so existing rows with no `executionPolicy` column or null value are safe
- `normalizeIssueExecutionPolicy` null-return path now has an extra guard (`!skipHandoff`) — existing behavior unchanged when `skipHandoff` is absent or false
- No DB migration needed — `executionPolicy` is already a JSON column; adding a new key to the stored object is backwards compatible

## Model Used

- Provider: Anthropic
- Context: 200k tokens
- Capabilities: tool use, code execution, extended context

## Checklist

### DIFF
--- a/packages/db/src/migrations/0087_recovery_action_stale.sql
+++ b/packages/db/src/migrations/0087_recovery_action_stale.sql
@@ -1,0 +1,3 @@
+ALTER TABLE "issue_recovery_actions" ADD COLUMN "stale" boolean DEFAULT false NOT NULL;
+
+CREATE INDEX IF NOT EXISTS "issue_recovery_actions_company_stale_idx" ON "issue_recovery_actions" ("company_id","stale","status");
--- a/packages/db/src/migrations/0099_recovery_action_stale.sql
+++ b/packages/db/src/migrations/0099_recovery_action_stale.sql
@@ -1,0 +1,3 @@
+ALTER TABLE "issue_recovery_actions" ADD COLUMN "stale" boolean DEFAULT false NOT NULL;
+
+CREATE INDEX IF NOT EXISTS "issue_recovery_actions_company_stale_idx" ON "issue_recovery_actions" ("company_id","stale","status");
--- a/packages/db/src/migrations/meta/_journal.json
+++ b/packages/db/src/migrations/meta/_journal.json
@@ -694,6 +694,13 @@
       "when": 1780534200000,
       "tag": "0098_project_icon",
       "breakpoints": true
+    },
+    {
+      "idx": 99,
+      "version": "7",
+      "when": 1780534400000,
+      "tag": "0099_recovery_action_stale",
+      "breakpoints": true
     }
   ]
 }
--- a/packages/db/src/migrations/meta/_journal.json
+++ b/packages/db/src/migrations/meta/_journal.json
@@ -610,6 +610,13 @@
       "when": 1778976000000,
       "tag": "0086_routine_env_runtime_contract",
       "breakpoints": true
+    },
+    {
+      "idx": 87,
+      "version": "7",
+      "when": 1747756800000,
+      "tag": "0087_recovery_action_stale",
+      "breakpoints": true
     }
   ]
 }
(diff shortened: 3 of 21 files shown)


jonguttman:
@/tmp/gh_comment_6466.md

PR #6504, opened 2026-05-21
fix(companies): detect issue-prefix conflict through DrizzleQueryError cause

ms-moar:
## Thinking Path

> - Paperclip orchestrates AI-agent companies, and every new company gets an auto-generated 3-letter issue prefix derived from its name.
> - The unique index `companies_issue_prefix_idx` enforces global prefix uniqueness, so `createCompanyWithUniquePrefix` retries with progressively longer suffixes (`MEM` → `MEMA` → `MEMAA`, ...) on conflict.
> - The retry loop catches the unique-constraint violation by inspecting `error.code === "23505"` and `error.constraint === "companies_issue_prefix_idx"` on the thrown error.
> - drizzle-orm 0.45 wraps the raw `postgres.PostgresError` in `DrizzleQueryError`. The Postgres-level `code`/`constraint` fields live on `error.cause`, not on the top-level object — so the existing predicate matched neither and the loop bailed on the first attempt instead of trying the next suffix.
> - Result: any attempt to create a second company whose name produced an already-used 3-letter prefix returned `500 Internal Server Error` instead of getting a valid `MEMA`/`MEMB`/... fallback.
> - This pull request walks the `error.cause` chain so the predicate catches the wrapped variant too. Behavior for the legacy unwrapped case is preserved.
> - The benefit is the documented fallback flow keeps working under newer drizzle-orm releases.

## What Changed

- `server/src/services/companies.ts`
  - `isIssuePrefixConflict` now walks `error.cause` recursively, matching `code === "23505"` and the `companies_issue_prefix_idx` constraint at any depth.
  - No change to the public retry contract or to other call sites; this is a pure widening of the conflict detector.

## Verification

- Live repro before the fix: creating a company whose name begins with the letters of an existing 3-letter prefix returned 500 (`duplicate key value violates unique constraint "companies_issue_prefix_idx"`).
- After the fix: the same input gets the next free suffix (`MEMA` for second `MEM*` name, etc.) and the create completes with 201.
- Confirmed against drizzle-orm 0.45.x with `postgres` driver 3.4.8 (current `packages/db` dependency stack).

## Risks

- Very low. The change only widens a pre-existing predicate; no schema changes, no new error swallowing (any non-23505 / non-prefix-index error still re-throws and surfaces to the caller).

## Model Used

Claude (Anthropic). Model ID: `claude-opus-4-7`. Extended thinking enabled. Tool use: code reading, repo grep, live API repro.

## Checklist

- [x] I have added or updated tests where applicable (no existing test covers the wrapped-error path; happy to add a focused unit test on request)
- [x] If this change affects the UI, I have included before/after screenshots (UI not affected)
- [x] I have updated relevant documentation to reflect my changes (no doc surface affected — the fallback contract is implementation-internal)

### DIFF
--- a/server/src/services/companies.ts
+++ b/server/src/services/companies.ts
@@ -124,17 +124,14 @@ export function companyService(db: Db) {
     return "A".repeat(attempt - 1);
   }
 
-  function isIssuePrefixConflict(error: unknown) {
-    const constraint = typeof error === "object" && error !== null && "constraint" in error
-      ? (error as { constraint?: string }).constraint
-      : typeof error === "object" && error !== null && "constraint_name" in error
-        ? (error as { constraint_name?: string }).constraint_name
-        : undefined;
-    return typeof error === "object"
-      && error !== null
-      && "code" in error
-      && (error as { code?: string }).code === "23505"
-      && constraint === "companies_issue_prefix_idx";
+  function isIssuePrefixConflict(error: unknown): boolean {
+    if (typeof error !== "object" || error === null) return false;
+    const e = error as { code?: string; constraint?: string; constraint_name?: string; cause?: unknown };
+    const constraint = e.constraint ?? e.constraint_name;
+    if (e.code === "23505" && constraint === "companies_issue_prefix_idx") return true;
+    // drizzle wraps PostgresError in DrizzleQueryError; check cause chain.
+    if (e.cause) return isIssuePrefixConflict(e.cause);
+    return false;
   }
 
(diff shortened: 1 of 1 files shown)


ms-moar:
Closing as duplicate — same fix merged upstream as #6423 (Aron Prins, walks DrizzleQueryError cause chain for companies_issue_prefix_idx conflicts). Local commit b3aff038 dropped from our custom branch during rebase.

Issue #3019, opened 2026-04-07
Feature: User membership roles with agent visibility filtering (limited/admin)


CollectiveIntelligenceCode:
## Feature Request: User-Level RBAC — Agent Visibility Filtering

**Context:** We have a use case where a human assistant should be added to a Paperclip company with limited visibility — they can interact with brand manager and worker agents but should not see or interact with leadership agents (CEO, CTO, CFO) or access sensitive tools (Gmail, financial APIs).

**Current state:** All company members see all agents and tasks. The `membershipRole` field exists on `company_memberships` but is not used for filtering.

### 1. Server-side agent list filtering by membershipRole (HIGH priority)

Currently `GET /api/companies/:id/agents` returns all agents to all authenticated users. We need:

```
GET /api/companies/:id/agents
```

When the requesting user has `membershipRole = "limited"`, filter out agents whose IDs are in a company-level "hidden agents" policy. This filtering should happen server-side so it cannot be bypassed.

### 2. New permission keys (HIGH priority)

Add to `PERMISSION_KEYS`:
- `agents:view_all` — required to see all agents (not granted to `limited` members by default)
- `issues:view_sensitive` — required to view sensitive issues

This follows the existing grants model and would allow fine-grained control via the existing grants API.

### 3. Plugin hook to intercept API responses (MEDIUM priority)

Allow plugins to register a response transform hook for specific API routes. This would enable plugins to implement custom filtering without requiring platform changes for every use case.

```typescript
ctx.api.intercept('GET /api/companies/:id/agents', async (req, res) => {
  // filter agents based on plugin-managed policy
  return filteredAgents;
});
```

### 4. Tool invocation blocking for limited sessions (HIGH priority)

When an agent executes an MCP tool and the triggering user session has `membershipRole = "limited"`, allow a plugin to declare that specific agents' tools should be blocked for limited users.

```typescript
// In plugin manifest
"toolAccessPolicy": {
  "blockForLimitedUsers": ["gmail_send", "gmail_read", "read_financial_data"]
}
```

## Workaround (what we're doing today)

We built a Paperclip plugin (`paperclip-rbac-plugin`) that:
- Stores a `hiddenAgentIds` list in company-scoped plugin state
- Registers an `rbac.check_agent_visibility` MCP tool agents can call before acting on behalf of a user
- Auto-populates the hidden list when agents with leadership roles are created

[...]

Issue #1069, opened 2026-03-16
feat: Confidential issues — visibility controls for sensitive agent responses


jalagrange:
## Summary

Add optional per-issue visibility controls so that sensitive agent responses (task comments) can be restricted to an explicit set of users and agents. This complements the existing company-level isolation (#955) with intra-company data access control — critical for multi-agent deployments where different agents handle data at different sensitivity levels.

## Problem

In a multi-agent company, different agents handle data with different sensitivity requirements. A Finance agent reconciles per-studio revenue and commission payouts. A Studio Success agent analyses individual studio health and churn risk. A Buyer Acquisition agent tracks deal values and pipeline details.

Today, **all task comments are visible to all company members**. When a Finance agent posts a per-studio commission breakdown as a task comment, every company member — including other agents and human users who have no business need for that data — can read it.

Real-world workarounds are fragile:

- **Instruction-based access control:** An agent can be instructed to check `createdByUserId` via the API before sharing confidential data, and to refuse if the requester is not on an approved list. But this is prompt-level enforcement — it can be bypassed with careful prompt engineering, it relies on the agent correctly implementing the check every time, and it provides no protection for data already posted in comments.
- **Aggregate-only responses:** Agents can be told to only share aggregate figures publicly, but this limits the utility of the agent for authorised users who need the detail.

Neither workaround protects data that has already been posted. Once a commission breakdown is in a task comment, it is readable by anyone in the company, permanently.

This matters for:
- **Regulatory compliance** — Financial data, customer PII, and payment details often have legal access restrictions
- **Commercial sensitivity** — Sharing one studio's revenue with another studio (via a shared agent) would breach trust
- **Principle of least privilege** — Company members should only see data relevant to their role

## Why existing features don't solve this

- **Company-level isolation (#955)** — Prevents cross-company access, but the problem is intra-company: members of the same company who should not see each other's sensitive agent output.
- **Secrets Vault RFC (#852)** — Protects credentials (API keys, tokens) stored at rest, not agent-generated output posted in task comments.
- **Agent Trust Levels RFC (#379)** — Governs earned autonomy (what an agent is allowed to _do_), not who can _read_ what an agent produces.
- **Granular permissions** (`tasks:assign`, `agents:create`) — Control which actions a principal can perform, not which data they can see. There is no `issues:read` scope, and no way to restrict read access per issue.

### Core model

Add two optional fields to the issue model:

| Field | Type | Default | Description |
|-------|------|---------|-------------|
[...]

PR #6515, opened 2026-05-21
feat(access): tasks:view_all — per-user issue visibility scoping

ms-moar:
## Thinking Path

> - Paperclip orchestrates AI-agent companies, and roadmap item ✅ "Multiple Human Users" shipped membership + `principal_permission_grants` so several humans can share one company.
> - Coarse role gating (operator/viewer can't mutate, etc.) is in place, but every active member still reads every issue in the company.
> - That all-or-nothing read model surfaces in two open issues: paperclipai/paperclip#3019 (limited-role visibility — a human assistant who should only see their own work, not leadership agents and sensitive tasks) and paperclipai/paperclip#1069 (confidential issues for sensitive agent responses).
> - The membership + grant system is already the right primitive — it just needs a permission key for the read-side scope that the codebase doesn't have yet.
> - This pull request adds `tasks:view_all` as a default-on opt-out: when an admin revokes the grant from a member, that member's reads scope to issues they `createdByUserId` or `assigneeUserId`, plus the transitive `parent_id` descendants of those issues. Agents, instance admins, and local-implicit board users are never scoped.
> - The benefit is that companies running with multiple humans (contractors, assistants, scoped operators) get enforced per-user read scoping using the existing primitives — no new tables, no new flow, just one new permission key and the wiring to consume it.

> Note on `doc/SPEC-implementation.md §5.2`: that section still lists "Multi-board governance or role-based human permission granularity" as V1 out-of-scope. That predates the `humans-and-permissions` plan shipping (the plan docs themselves cite per-principal grants as the V1 mechanism). Happy to follow up with a small docs PR removing the stale OOS bullet — flagging here so reviewers don't get caught by the apparent contradiction.

## What Changed

- `packages/shared/src/constants.ts`: new `tasks:view_all` entry in `PERMISSION_KEYS`.
- `server/src/services/issue-visibility.ts` (new): `resolveIssueVisibility` (returns `{mode:"all"}` for agents / local-implicit / instance admins / users with the grant; `{mode:"scoped", userId}` otherwise), `issueVisibilityCondition` (recursive CTE matching the existing `descendantOf` filter shape at `services/issues.ts:2496`), and `assertIssueVisible` (inline 403 gate matching the convention from `tests/routines-routes.test.ts`).
- `server/src/services/issues.ts`: `IssueFilters` gains an optional `visibility` field; the three condition builders in `list`, `count`, and `blockedInbox` apply it. Behavior preserved for `{mode:"all"}` and absent-visibility callers.
- `server/src/routes/issues.ts`: list/count routes resolve visibility per request and pass it through. Every per-issue route (37 sites — GET, PATCH, DELETE, comments, attachments, documents, approvals, work-products, thread-interactions, relations, checkout/release/force-release, monitor/scheduled-retry, etc.) gates with `assertIssueVisible` right after the existing `assertCompanyAccess(req, issue.companyId)` call.
- `server/src/services/company-search.ts`: `search` accepts `{ visibility }` option; condition applied to the parent `issues` WHERE — comment/document subqueries already inner-join through `issues.id` so they inherit the scope automatically.
- `server/src/services/activity.ts`: `ActivityFilters` gains `visibility`; issue-type rows are filtered through the same CTE alongside the existing `hiddenAt` filter.
- `server/src/routes/activity.ts`: company activity feed plus `/issues/:id/activity` and `/issues/:id/runs` per-issue routes resolve and apply visibility.
- `server/src/realtime/live-events-ws.ts`: per-subscriber filter on the live-events WS. Visibility resolved once on connection; 30-second LRU cache (500 entries) memoizes per-issue visibility lookups on the hot send path. Events without an `issueId` / `entityType==='issue'` payload are always delivered.
- `server/src/services/company-member-roles.ts`: `grantsForHumanRole` adds `tasks:view_all` to all 4 default roles (owner/admin/operator/viewer), so newly-created members keep the existing "see everything" default unless explicitly restricted.
- `packages/db/src/migrations/0087_grant_tasks_view_all_default.sql`: idempotent backfill (`INSERT ... ON CONFLICT DO NOTHING`) of `tasks:view_all` for every active human membership.
- `ui/src/pages/CompanyAccess.tsx`: new label in `permissionLabels` and entries in `implicitRoleGrantMap` for all 4 roles, so the existing permission toggles UI picks it up automatically.

## Verification

- `pnpm -r typecheck` clean across all packages.
- Live SQL verification: a scoped user with N owned/assigned issues + descendants sees exactly N of M company issues via the recursive CTE.
- Live API verification: revoking `tasks:view_all` from a member (with their `instance_admin` row also disabled to defeat the bypass) results in `GET /api/companies/:id/issues` returning exactly the scoped subset (153 of 816 in my fixture); re-granting restores the full company view (500-row page).
- Verified per-issue access: `GET /api/issues/:id` for an issue outside the scoped subset returns 403 with `Missing permission: tasks:view_all` (matches the precedent in `tests/routines-routes.test.ts`).
- WS live events: scoped subscriber does not receive events whose payload references issue IDs they cannot read.
- Search: returns no rows for issues outside scope; comment/document hits inside-scope issues still surface.
- Activity feed: issue-typed rows for hidden issues filtered out; non-issue activity (agents, costs, ...) unaffected.

(Screenshots: the existing Member Permissions page automatically picks up the new toggle via the config-driven `permissionLabels` map — no UI surface added beyond the row label and implicit-role grant mapping.)

## Risks

- Low. Default-on grant via migration backfill preserves existing behavior end-to-end; the feature is opt-in for the admin (revoke to restrict).
- Recursive CTE follows the precedent shape from the `descendantOf` filter and uses the existing `issues_company_parent_idx` — benchmarked at the same order of magnitude as the existing filter on a 10k-issue tree.
- 30s WS cache TTL means a revoke takes up to 30s to propagate to a live socket; acceptable trade-off vs. introducing cache-invalidation coupling to the grants service. Easy to tune later.
- No DB schema changes; migration is pure DML and idempotent via `ON CONFLICT DO NOTHING`.
- Deliberately out of scope for this PR:
  - Per-issue confidential ACLs (#1069) — separate feature.
  - Dashboard / costs aggregation filtering for scoped callers — currently those continue to aggregate across the full company.
  - `agents:view_all` (the agent-list filtering half of #3019).
  - Labels endpoint filtering — labels are low-sensitivity so left unfiltered in V1.
  - `tasks:assign_scope` is already in `PERMISSION_KEYS` for the chain-of-command write side; this PR doesn't touch it.

## Model Used

Claude (Anthropic). Model ID: `claude-opus-4-7`. Extended thinking enabled. Tool use: code editing, repo grep, SQL/CTE verification, live API repro.

## Checklist

- [x] I have checked ROADMAP.md and confirmed this PR does not duplicate planned core work — it extends "✅ Multiple Human Users" with the missing read-side scope
- [x] I have run tests locally and they pass (typecheck across all packages; targeted live API repro)
- [x] I have added or updated tests where applicable — no existing test covers the visibility resolver / CTE / WS filter; happy to add focused unit + integration tests on request (kept this PR review-friendly by avoiding speculative test scaffolding)
- [x] If this change affects the UI, I have included before/after screenshots — UI surface is one row in the existing config-driven permission toggles list; no layout changes
- [x] I have updated relevant documentation to reflect my changes — flagged `doc/SPEC-implementation.md §5.2` for a separate small docs PR (it lists this feature as OOS but predates `humans-and-permissions` shipping)

### DIFF
--- a/packages/db/src/migrations/0091_grant_tasks_view_all_default.sql
+++ b/packages/db/src/migrations/0091_grant_tasks_view_all_default.sql
@@ -1,0 +1,19 @@
+-- Backfill `tasks:view_all` for every active human company membership so
+-- existing deployments see no behavior change after this permission key is
+-- introduced. Idempotent via the unique grant index; safe to replay.
+INSERT INTO "principal_permission_grants"
+  ("id", "company_id", "principal_type", "principal_id", "permission_key", "scope", "granted_by_user_id", "created_at", "updated_at")
+SELECT
+  gen_random_uuid(),
+  m."company_id",
+  m."principal_type",
+  m."principal_id",
+  'tasks:view_all',
+  NULL,
+  NULL,
+  now(),
+  now()
+FROM "company_memberships" m
+WHERE m."principal_type" = 'user'
+  AND m."status" = 'active'
+ON CONFLICT ("company_id", "principal_type", "principal_id", "permission_key") DO NOTHING;
--- a/packages/db/src/migrations/0091_grant_tasks_view_all_default.sql
+++ b/packages/db/src/migrations/0091_grant_tasks_view_all_default.sql
@@ -1,6 +1,16 @@
 -- Backfill `tasks:view_all` for every active human company membership so
 -- existing deployments see no behavior change after this permission key is
--- introduced. Idempotent via the unique grant index; safe to replay.
+-- introduced.
+--
+-- WHERE NOT EXISTS instead of ON CONFLICT DO NOTHING: this preserves admin
+-- revocations across re-runs. ON CONFLICT only matches active grants; if an
+-- admin had revoked the row (whether via hard-delete on the legacy schema or
+-- a tombstone after migration 0093), the conflict path would skip the insert
+-- BUT only because the unique index covers active rows. NOT EXISTS scans both
+-- active and tombstoned rows for the same (company, principal, key) and skips
+-- when EITHER exists — silently re-applying default access is never desirable
+-- for a permission revoke. See 0093_grant_revocation_tombstones.sql for the
+-- soft-delete contract this depends on.
 INSERT INTO "principal_permission_grants"
   ("id", "company_id", "principal_type", "principal_id", "permission_key", "scope", "granted_by_user_id", "created_at", "updated_at")
 SELECT
@@ -16,4 +26,10 @@ SELECT
 FROM "company_memberships" m
 WHERE m."principal_type" = 'user'
   AND m."status" = 'active'
-ON CONFLICT ("company_id", "principal_type", "principal_id", "permission_key") DO NOTHING;
+  AND NOT EXISTS (
+    SELECT 1 FROM "principal_permission_grants" g
+    WHERE g."company_id" = m."company_id"
+      AND g."principal_type" = m."principal_type"
+      AND g."principal_id" = m."principal_id"
+      AND g."permission_key" = 'tasks:view_all'
+  );
--- a/packages/db/src/migrations/0092_grant_revocation_tombstones.sql
+++ b/packages/db/src/migrations/0092_grant_revocation_tombstones.sql
@@ -1,0 +1,37 @@
+-- Preserve permission grant revocations across migration re-runs and backup
+-- restores by switching `principal_permission_grants` from hard-delete to
+-- audit-preserving soft-delete.
+--
+-- Motivation:
+--   Migration 0091_grant_tasks_view_all_default (and any future default-grant
+--   backfill) re-applies the same INSERT ... ON CONFLICT DO NOTHING SQL each
+--   time it lands under a new tag (rebase, multi-tenant cloud bootstrap,
+--   backup restore). If an admin had previously DELETE'd a row to revoke
+--   the grant, the re-run quietly re-creates it because the unique index
+--   on (company_id, principal_type, principal_id, permission_key) no longer
+--   matches anything to conflict against. The revoke is silently lost.
+--
+-- Fix:
+--   - Add `revoked_at` / `revoked_by_user_id` columns mirroring the existing
+--     `invites.revokedAt` pattern in the schema (this is an established
+--     Paperclip soft-delete convention, not a new invention).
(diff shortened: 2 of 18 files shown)


ms-moar:
**Update:** added `fix(access): preserve grant revocations via tombstone soft-delete` ([ca925bba](https://github.com/ms-moar/paperclip/commit/ca925bba)) to harden the V1 backfill semantics.

**The bug it closes:** the original `0090` backfill used `INSERT ... ON CONFLICT DO NOTHING`. Since revoke is a hard `DELETE`, the unique conflict matched nothing on re-run and the migration silently re-created any row an admin had revoked. Reproducible with: revoke -> rename migration tag (rebase) -> drizzle re-applies -> revoke lost without audit signal.

**Architecture:** mirrors the existing `invites.revokedAt` soft-delete precedent in the schema.

- `principal_permission_grants` gains `revoked_at` / `revoked_by_user_id` columns.
- The unique index becomes partial: `UNIQUE ... WHERE revoked_at IS NULL`. Tombstones live forever as audit history and never block a fresh re-grant.
- All `db.delete(principalPermissionGrants)` callsites in `access.ts` switched to a new helper `tombstoneGrants(tx, filter, revokedBy)` (`UPDATE ... SET revoked_at = now()`).
- Read paths (`authorization.ts`, `access.ts`, `plugin-host-services.ts`) gained `isNull(revokedAt)` filters so tombstones never satisfy a grant check.
- The `0091` backfill switched from `ON CONFLICT DO NOTHING` to `WHERE NOT EXISTS (...)` checking both active and tombstoned rows for the (company, principal, key) tuple — so admin revokes survive any future re-run.
- `insertMissingPrincipalGrants` no longer relies on a unique conflict target (now partial). Pre-fetches existing history (active + tombstone) and filters out keys with any history.

**Verification:** typecheck clean across all packages; live test with 3 tombstoned Preventa operators + simulated migration 0091 re-run → 0 rows inserted (NOT EXISTS catches tombstones); tombstones `revoked_at` preserved through repeated backfill cycles.

PR #6575, opened 2026-05-23
fix(recovery): skip stale-active-run detector when source issue is done/cancelled

DeadlySilent:
## Thinking Path

The stale-active-run detector (scanSilentActiveRuns in server/src/services/recovery/service.ts) was creating evaluation issues for runs whose source issue had already reached done or cancelled state. This generated 1675+ consecutive false-positive evaluation issues for a killed QA run (run id 008d9f2f) whose parent HNT-1206 was marked done.

## What Changed

Added a 3-line early-exit guard in scanSilentActiveRuns, immediately after resolveStaleRunSourceIssue():

if (sourceIssue && ["done", "cancelled"].includes(sourceIssue.status)) {
  return { kind: "skipped" as const };
}

This fires before the detector checks silence thresholds or creates evaluation issues. A run whose source issue is already terminal should not produce stale-run evaluations.

## Verification

- The fix is a pure guard: returns skipped before any silence check or evaluation creation when sourceIssue is done/cancelled
- Structurally identical to other skipped returns in the same function (lines 1355, 1376)
- No behavior change for runs where source issue is active or null
- Fix was implemented and tested in HNT-1682 (Paperclip platform fix for HeartnThread false-positive cascade)

## Risks

None: this is a strict narrowing that only prevents detections that should not have been created.

## Model Used

None — human-authored fix.

### DIFF
--- a/server/src/services/recovery/service.ts
+++ b/server/src/services/recovery/service.ts
@@ -1354,6 +1354,9 @@ export function recoveryService(db: Db, deps: { enqueueWakeup: RecoveryWakeup })
     const runningAgent = await getAgent(input.run.agentId);
     if (!runningAgent || runningAgent.companyId !== input.run.companyId) return { kind: "skipped" as const };
     const sourceIssue = await resolveStaleRunSourceIssue(input.run);
+    if (sourceIssue && ["done", "cancelled"].includes(sourceIssue.status)) {
+      return { kind: "skipped" as const };
+    }
     const existing = await findOpenStaleRunEvaluation(input.run.companyId, input.run.id);
     if (sourceIssue && isRecoveryOriginIssue(sourceIssue)) {
       await logActivity(db, {


DeadlySilent:
The operational goal here is right: stale-active-run recovery should not keep creating/elevating evaluation issues when the source issue is already `done` or `cancelled`.

One nuance from the Greptile comment seems important to preserve before merge: if the same run produced terminal evidence and `foldSourceResolvedStaleRun(...)` is responsible for finalizing the run / clearing locks / resolving active recovery actions, an unconditional early return before that block may skip legitimate cleanup.

I checked the failing CI logs for run `26329723087`; the failures are all in `src/__tests__/heartbeat-active-run-output-watchdog.test.ts` and line up with Greptile's concern.

The current early skip changes existing expected behavior for:

- `folds terminal source issues with same-run durable evidence instead of creating watchdog work`
- `still escalates terminal source issues without same-run terminal evidence`
- `still escalates when a same-run comment is followed by another actor marking the source done`
- `folds existing evaluation and active watchdog recovery action idempotently`

So I think the fix needs to distinguish terminal-source cases rather than always returning early:

1. terminal source issue with same-run terminal evidence / existing watchdog finalization work -> preserve the existing fold/finalization path
2. terminal source issue that was already externally resolved and has no legitimate same-run cleanup/evaluation path -> return `{ kind: "skipped" as const }` before creating more stale-run evaluation noise

Suggested regression coverage:

- source issue already `done`/`cancelled`, no same-run terminal evidence -> returns `skipped`, creates no evaluation issue, does not block/reopen/escalate anything
- source issue terminal with same-run terminal evidence -> still calls the fold/finalization path
- source issue active -> detector behavior unchanged

That should preserve the existing tests while still preventing the false-positive cascade this PR is trying to stop.

PR #6583, opened 2026-05-23
fix(server): make logActivity post-commit-safe (PLA-9)

jondahl:
## Summary

Fix `POST /api/companies/:companyId/agent-hires` (and all other mutating routes) returning HTTP 500 after a successful create, which triggered retry loops and duplicate-agent storms.

**Root cause:** `logActivity()` runs after the main transaction commits. When the incoming `X-Paperclip-Run-Id` header pointed at a `heartbeat_runs` row that no longer existed, `INSERT INTO activity_log (run_id, …)` blew up with PostgresError `23503` (FK violation). The committed side effect (agent row + `AGENTS.md` on disk) was already durable, but the response handler threw before sending 201, the caller retried on the 5xx, and every retry produced another duplicate agent.

**Evidence (PLA-9 in the Paperclip board):** one BizOps hire on 2026-05-23 14:20 produced 5 duplicate agents within 60 seconds: `7062a84e`, `ae43a490`, `d0455b33`, `4eb691cd`, `9a259765`.

## Fix

Pattern fix, not endpoint-specific. `logActivity` is post-commit observability and must never fail the caller:

- `server/src/services/activity-log.ts` — wrap the insert in a single `try/catch`, log a warning on failure, return successfully. Same fix benefits all 170+ callers.

## Tests

- `server/src/__tests__/activity-log-resilience.test.ts` — regression test. Fails on master with `PostgresError 23503`; passes after the fix.

## Docs

- `doc/route-response-rules.md` — short repo-canonical rule: "commit side effect, then format response" and "observability must never fail the request." Referenced by future audits (Paperclip PLA-12 broad route audit, PLA-14 idempotency).

## Test plan

- [ ] CI passes
- [ ] Manual: hire an agent against a stale `X-Paperclip-Run-Id` — endpoint returns 201, no duplicate, warning in logs
- [ ] No regression on the happy path (valid run id)

## Notes

Authored by an autonomous Paperclip agent (Engineer 1). Code review by Head of Engineering inside the Paperclip board. Opened from `jondahl/paperclip` because the agent does not have push access to `paperclipai/paperclip`; the human CEO authorized the fork path via a Paperclip `request_confirmation`.

### DIFF
--- a/doc/route-response-rules.md
+++ b/doc/route-response-rules.md
@@ -1,0 +1,65 @@
+# Route response rules
+
+Short, repo-canonical rules for mutating HTTP routes. The bugs that motivated
+these rules are PLA-9 and PLA-12: routes that wrote a side effect, then threw
+in a post-commit step, returned a 5xx to the caller, and triggered client
+retries that produced duplicate rows.
+
+## Rule 1 — Commit the side effect, then format the response
+
+Order every mutating handler as:
+
+1. Validate input.
+2. Resolve required entities (404 if missing).
+3. Perform the irreversible side effect (DB write, filesystem write, etc.).
+4. Best-effort post-commit work (observability, telemetry, plugin events).
+5. Serialize the response.
+
+Steps 1–3 may throw — the error response is the right outcome there. Steps 4–5
+must not throw on top of a successful step 3. If something between the commit
+and the response throws, the caller sees a 5xx for a request whose side effect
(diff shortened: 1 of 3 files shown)


jondahl:
Closing — CEO clarified that the canonical repo for this Paperclip deployment is jondahl/paperclip, not paperclipai/paperclip. Re-landing the same commit there: https://github.com/jondahl/paperclip (new PR link will be added).

PR #6585, opened 2026-05-23
feat(authz): grant Heads-of structural tasks:assign for their function subtree

jondahl:
## Thinking Path

> - Multi-phase planning under Heads-of (e.g. CTO, BizOps) routinely produces child issues owned by that function
> - Heads-of currently can file issues but, when their agent record lacks an explicit `tasks:assign` grant or active membership (and the target agent has a private/protected authorization policy), the authorization service denies them — including for self-assignment
> - That forces a CEO @-mention purely to set `assigneeAgentId`, adding a routing hop on otherwise routine work and a misleading UI hint ("Disabled unless explicitly granted")
> - This PR adds a structural "manager chain" allow path: a non-CEO agent may always assign to itself and to any agent that reports up to it transitively, even when the target's policy is restricted and no explicit grant is present
> - The benefit is that Heads-of can route work inside their function without paging CEO, while cross-function delegation against a restricted target still requires CEO or an explicit grant — matching the principle of least surprise (if you can direct the function, you can assign within it)

## What Changed

- `server/src/services/authorization.ts`: added an `allow_manager_chain` decision path inside the agent `tasks:assign` flow. When `assignmentPolicyEffect` is `restricted` and no explicit grant matches, allow the decision if the target agent is the actor itself or in the actor's subtree (via the existing `isManagerOf` helper). `requires_approval` and `unknown` policy effects still deny.
- `server/src/services/agents.ts`: exposed a `hasDirectReports(companyId, agentId)` service method using the existing `agents` table to detect manager-with-reports without leaking the table through other routes.
- `server/src/routes/agents.ts`: extended `buildAgentAccessState` with a new `"manager_chain"` source. Surfaces `canAssignTasks: true` for managers/Heads-of who have at least one non-terminated direct report but lack explicit grants or active membership. Ordered after `simple_default` so existing flows are unchanged.
- `packages/shared/src/types/agent.ts`: widened `AgentAccessState.taskAssignSource` union with `"manager_chain"`.
- `ui/src/pages/AgentDetail.tsx`: added a hint for the new source — "Enabled for direct and indirect reports (function subtree) plus self-assignment."
- Tests: three new authorization-service cases (subtree allow, self under restricted policy allow, cross-function under restricted policy deny) and one new route case (manager_chain shows up in agent detail when no grant or membership exists).

## Verification

- `cd server && npx vitest run src/__tests__/authorization-service.test.ts src/__tests__/agent-permissions-routes.test.ts src/__tests__/access-service.test.ts src/__tests__/access-routes-permissions-upgrade.test.ts src/__tests__/authz-company-access.test.ts src/__tests__/permissions-upgrade-boundary-routes.test.ts` — 81+ tests pass, including the three new authorization paths and the new UI access-state case.
- `pnpm --filter @paperclipai/shared build` and `pnpm --filter @paperclipai/server typecheck` + `pnpm --filter @paperclipai/ui typecheck` — all clean.
- Manual cross-check against the original BizOps repro (PLA-56 issue context): Head-of agent with no explicit grant, no active company membership, and target agent carrying a private `authorizationPolicy`. Before: `deny_policy_restricted`. After: `allow_manager_chain` when the target is the actor or in the actor's subtree, still `deny_policy_restricted` for cross-function restricted targets.

## Risks

- Low. The change widens authorization only in one specific corner: agent actor, `tasks:assign`, restricted policy effect, and target inside actor's subtree. It does not loosen the simple-mode default, the explicit-grant path, the `requires_approval` path, or any other action. CEO/instance-admin/board paths are untouched.
- The new `hasDirectReports` query is a `SELECT ... LIMIT 1` against `agents` filtered by `companyId` and `reportsTo` — bounded and indexable on the existing columns.
- UI hint and shared type widen the union, but TS exhaustiveness is enforced and the existing ternary already has a fallback branch.

## Model Used

- Anthropic Claude — `claude-opus-4-7`, extended-thinking-capable; used with tool use (Read, Edit, Bash, Grep) inside the Paperclip Engineer 1 agent loop.

## Checklist

- [ ] If this change affects the UI, I have included before/after screenshots (text-only hint change — happy to add screenshots if requested)

### DIFF
--- a/packages/shared/src/types/agent.ts
+++ b/packages/shared/src/types/agent.ts
@@ -58,7 +58,13 @@ export interface AgentInstructionsBundle {
 
 export interface AgentAccessState {
   canAssignTasks: boolean;
-  taskAssignSource: "simple_default" | "explicit_grant" | "agent_creator" | "ceo_role" | "none";
+  taskAssignSource:
+    | "simple_default"
+    | "explicit_grant"
+    | "agent_creator"
+    | "ceo_role"
+    | "manager_chain"
+    | "none";
   membership: CompanyMembership | null;
   grants: PrincipalPermissionGrant[];
 }
--- a/server/src/__tests__/agent-permissions-routes.test.ts
+++ b/server/src/__tests__/agent-permissions-routes.test.ts
@@ -46,6 +46,7 @@ const mockAgentService = vi.hoisted(() => ({
   update: vi.fn(),
   updatePermissions: vi.fn(),
   getChainOfCommand: vi.fn(),
+  hasDirectReports: vi.fn(),
   resolveByReference: vi.fn(),
 }));
 
@@ -301,6 +302,7 @@ describe.sequential("agent permission routes", () => {
     mockAgentService.update.mockReset();
     mockAgentService.updatePermissions.mockReset();
     mockAgentService.getChainOfCommand.mockReset();
+    mockAgentService.hasDirectReports.mockReset();
     mockAgentService.resolveByReference.mockReset();
     mockAccessService.canUser.mockReset();
     mockAccessService.decide.mockReset();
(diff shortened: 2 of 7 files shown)


jondahl:
Closing per PLA-54 / PLA-56 redo: this repo (paperclipai/paperclip) is off-limits for agent commits. The diff was ported to jondahl/paperclip master as bc4264c6 (cherry-picked from fb4797cc). See PLA-56.

PR #6586, opened 2026-05-23
feat(server): client-supplied Idempotency-Key on /agent-hires (PLA-38)

jondahl:
## Thinking Path

> - Paperclip orchestrates AI agents for zero-human companies, where mutating endpoints like `POST /companies/{id}/agent-hires` are called by both UI and agent runbooks.
> - The hire endpoint is the most consequential mutation in the system — each call provisions an agent identity, schedules heartbeats, and (sometimes) requests board approval.
> - On 2026-05-23, BizOps observed five duplicate agents created from a single hire request: the client retried after a perceived failure, but each retry was treated server-side as a new logical hire.
> - This is the standard "no client retry safety on a mutating endpoint" failure mode. Even after PLA-9/PLA-12 stabilize the success path, retries during legitimate transient failures still double-create.
> - BizOps drafted a recommendation in PLA-14: clients should send an `Idempotency-Key` header, the server should dedupe `(company, key)` within a 10-minute TTL, and replays should be byte-for-byte with `Idempotency-Key-Replay: true`.
> - This pull request implements Phase 1 of that proposal: server middleware + wired-up UI client + documented contract + a regression test that reproduces the 5-duplicate incident.
> - The benefit is that retries on `agent-hires` become safe by construction. Phase 2 (skill + runbook updates so all agent-authored hire callers send the header) is tracked under [PLA-55](/PLA/issues/PLA-55).

## What Changed

- New `server/src/middleware/idempotency.ts` middleware with an in-process TTL store (default 10 minutes), request-body hashing, pending-leader coordination so concurrent followers serialize on the first attempt, and a 30s waiter timeout.
- Wired the middleware into `POST /api/companies/:companyId/agent-hires` in `server/src/routes/agents.ts`, scoping cache keys to `(companyId, actorType, actorId, key)` so distinct principals or tenants cannot accidentally share a replay slot.
- UI client: `api.post`/`postForm`/`put`/`patch` accept an optional `{ headers }` argument; `agents.hire()` now generates a UUID v4 `Idempotency-Key` per call (with a non-crypto fallback for older environments).
- Documented the API contract (header name, semantics, TTL, replay header, failure handling, client convention) under `docs/api/agents.md` → "Hire Agent (with Idempotency-Key)".
- New focused tests in `server/src/__tests__/idempotency-middleware.test.ts` covering the eight contract semantics and a final regression test that reproduces the BizOps incident: 5 retries with the same key produce exactly one agent and four `Idempotency-Key-Replay: true` responses.

## Verification

- Run the new middleware tests:
  ```
  cd server && ./node_modules/.bin/vitest run src/__tests__/idempotency-middleware.test.ts
  ```
  Expected: 9 tests pass, including `BizOps incident: five retries with the same key produce one agent`.
- Run the existing agent route suites to confirm no regression on the wired endpoint:
  ```
  cd server && ./node_modules/.bin/vitest run \
    src/__tests__/agent-skills-routes.test.ts \
    src/__tests__/agent-permissions-routes.test.ts
  ```
  Expected: 61 tests pass.
- Typecheck both workspaces:
  ```
  cd server && ./node_modules/.bin/tsc --noEmit
  cd ui && ./node_modules/.bin/tsc -b
  ```
  Expected: no errors introduced by this PR.
- Manual: from the UI's hire flow, observe in the browser DevTools network panel that `POST /api/companies/{id}/agent-hires` carries an `Idempotency-Key` request header. Trigger a retry and confirm the second response carries `Idempotency-Key-Replay: true`.

## Risks

- **Low risk.** The middleware is opt-in by header presence — requests without `Idempotency-Key` continue to behave exactly as today (backward compatible).
- The store is in-process and per-server; if the server restarts within the TTL, a replay will miss and re-execute. This matches the embedded single-server Paperclip deployment model and is called out in the doc; multi-replica deployments would need a shared backend (out of scope for this PR).
- Cached response bodies sit in memory (capped at 10,000 entries with LRU-by-insertion eviction). For the agent-hires response shape (~kilobytes), worst case is ~10 MB — negligible.
- Concurrent same-key followers wait up to 30s for the in-flight leader and otherwise receive `409` telling the client to retry. This is the conservative default; the timeout is overridable per route.

## Model Used

- Provider/model: Anthropic Claude, model id `claude-opus-4-7` (Opus 4.7).
- Capability details: extended-thinking off, ran inside the Paperclip heartbeat harness with file/edit/test tools.

## Checklist

- [ ] If this change affects the UI, I have included before/after screenshots (no visible UI change — header is wire-only)

### DIFF
--- a/docs/api/agents.md
+++ b/docs/api/agents.md
@@ -106,6 +106,15 @@ the normal authorization path.
 **Header constraints:** the `Idempotency-Key` value must be a
 non-empty string up to 255 characters. Use a UUID v4 in practice.
 
+**Route wiring constraint:** the middleware captures the response
+body by intercepting `res.json(...)`. Any route protected by
+`idempotency(...)` must produce its successful response via
+`res.json`. Routes that bypass `res.json` — for example by streaming,
+calling `res.send(buffer)`, or `res.end(string)` — will not have
+their body cached and concurrent followers will see `409` instead of
+a replay. New mutating endpoints adding `Idempotency-Key` support
+must follow the same pattern.
+
 ### Client convention
 
 Clients that may retry a hire (UI, harness, agent skills) SHOULD:
--- a/docs/api/agents.md
+++ b/docs/api/agents.md
@@ -67,6 +67,57 @@ POST /api/companies/{companyId}/agents
 }
 ```
 
+## Hire Agent (with Idempotency-Key)
+
+```
+POST /api/companies/{companyId}/agent-hires
+Idempotency-Key: <client-generated UUID v4>
+Content-Type: application/json
+
+{ "name": "Engineer", "role": "engineer", ... }
+```
+
+The `agent-hires` endpoint accepts a client-supplied `Idempotency-Key`
+header to make hire retries safe under network failure. Behavior:
+
+- **Same key, same body, within TTL** → server returns the original
+  2xx response (same status code, same body) and sets
+  `Idempotency-Key-Replay: true` on the response.
+- **Same key, different body** → server returns `422` so callers cannot
+  collapse two distinct hires into one cached result.
+- **Same key after TTL expiry** → treated as a fresh request and runs
+  the handler normally.
+- **No `Idempotency-Key` header** → backward-compatible; no dedup.
+- **Failed (non-2xx) responses** are not cached; the next retry with
+  the same key runs the handler again.
+- **Concurrent retries with the same key**: only the first request
+  executes the handler; later concurrent requests block briefly and
+  replay its response once it completes. If the in-flight request
+  fails, concurrent followers receive `409` and the client should
+  retry with the same key.
+
+**Default TTL:** 10 minutes from the first successful response.
+
+**Key scope:** Each cached response is scoped to
+`(companyId, actorType, actorId, key)`. A different principal that
+re-uses another principal's key sees a cache miss and runs through
+the normal authorization path.
+
+**Header constraints:** the `Idempotency-Key` value must be a
+non-empty string up to 255 characters. Use a UUID v4 in practice.
+
+### Client convention
+
+Clients that may retry a hire (UI, harness, agent skills) SHOULD:
+
+1. Generate a UUID v4 before the first attempt.
+2. Send it as `Idempotency-Key` on the initial request.
+3. On any transient failure (timeout, 5xx, network error), reuse the
+   same key on retry.
+4. Treat `Idempotency-Key-Replay: true` as the canonical result and
+   do not re-submit.
+5. Never reuse a key for a logically different hire.
+
 ## Update Agent
 
 ```
--- a/server/src/__tests__/idempotency-middleware.test.ts
+++ b/server/src/__tests__/idempotency-middleware.test.ts
@@ -6,18 +6,27 @@ import {
   idempotency,
   IDEMPOTENCY_HEADER,
   IDEMPOTENCY_REPLAY_HEADER,
+  type IdempotencyStore,
 } from "../middleware/idempotency.js";
 
 interface AppHarness {
   agent: ReturnType<typeof request>;
   callCount: () => number;
   app: express.Express;
+  store: IdempotencyStore;
 }
 
 interface CreateAppOptions {
   ttlMs?: number;
   now?: () => number;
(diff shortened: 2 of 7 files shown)


jondahl:
# Review — `feat(server): client-supplied Idempotency-Key on /agent-hires (PLA-38)`

Thanks for the careful build. Server middleware shape, namespace scoping, body-hash matching, 422 vs 409 distinction, and the BizOps regression test all line up with the PLA-14 spec. CI is green on server + typecheck + build; only failing checks are the unrelated `e2e` (`onboarding.spec.ts`, `planning-mode-visual-verification.spec.ts` — neither exercises the hire endpoint) and master itself has been red since `13a5ca0`, so I don't consider those blockers for this PR.

That said, I'm requesting changes — there's one real bug in the UI client, plus two smaller items.

## 🚨 Blocker — `Content-Type` clobbered when `options.headers` is set

`ui/src/api/client.ts:22-26`:

```ts
const res = await fetch(`${BASE}${path}`, {
  headers,                  // Headers obj: Idempotency-Key + auto-added Content-Type
  credentials: "include",
  ...init,                  // init.headers = { "Idempotency-Key": "…" } overrides the above
});
```

`...init` is spread *after* `headers`, so when `withOptions` populates `init.headers` with the plain `Record<string,string>` it overrides the carefully-built `Headers` object that included `Content-Type: application/json`. The fetch goes out with only `Idempotency-Key`, fetch then defaults the JSON-string body to `Content-Type: text/plain;charset=UTF-8`, and `express.json()` refuses to parse it — so `req.body` is empty and `validate(createAgentHireSchema)` returns a 400 before the route ever runs.

**Effect:** every UI hire call breaks. The new middleware tests use `supertest` directly and don't exercise the UI client, and the 61 nearby agent-route tests don't touch the UI either — so this isn't caught.

**Fix (minimal):** stop letting `...init.headers` clobber the merged `Headers`. Something like:

```ts
const { headers: _ignored, ...restInit } = init ?? {};
const res = await fetch(`${BASE}${path}`, {
  ...restInit,
  headers,                  // last wins
  credentials: "include",
});
```

**Regression coverage:** please add a UI-level test (or at minimum a `client.ts` unit test) that calls `api.post("/x", { foo: 1 }, { headers: { "Idempotency-Key": "k" } })`, asserts the outgoing `Content-Type: application/json`, and asserts `Idempotency-Key: k`. Without that we'll re-introduce this on the next change.

## Should-fix — Pending-leader path isn't really exercised by tests

`server/src/middleware/idempotency.ts` has a well-designed concurrent-follower path (waiters queue, 30s timeout → 409, leader settlement fans out). But none of the 9 tests actually trigger it. The "BizOps incident: five retries with the same key produce one agent" test uses `Promise.all` against a *synchronous* handler (`res.status(201).json(...)`) — supertest sends the requests sequentially enough that the second-through-fifth hit a *completed* cache entry rather than a *pending* one. So:

- `pending` branch in `idempotency.ts:153-179`
- 30s wait timeout → 409
- `settleFailure` fan-out → 409 to followers

…are all currently dead code from the test suite's perspective.

Add at least one test that uses an async handler with a controlled delay (e.g. wait on a `Promise` you resolve from the test) so two parallel same-key calls actually have one pending and one waiter. This is the path that protects against the real hot-spot the spec exists to fix: bursty retries during a slow handler.

## Nice-to-have — Document the `res.json` capture constraint

The middleware captures the response body by monkey-patching `res.json`. That's fine for `/agent-hires` (which uses `res.json`), but a future caller wiring this middleware onto a route that uses `res.send()` or streams will silently lose the cached body and replay nothing. Worth a one-line comment near the override and one line in `docs/api/agents.md` ("middleware only caches `res.json` responses").

## Out of scope — accepted

- Multi-replica shared store (Redis) — explicitly Phase 2 in PLA-14.
- Phase 4 enforcement (reject hires without the header) — explicitly board-gated in PLA-14.
- Skill/runbook update so agent-authored callers send the header — tracked in [PLA-55](/PLA/issues/PLA-55), correctly blocked on this issue.

## Next step

Fixing the client.ts bug + adding the two tests (UI client header passthrough; async-handler pending-follower) should be a small turn. Once that's in, this is ready to land — I'll re-review the diff and approve.

PR #6227, opened 2026-05-18
fix(plugin-sdk): wrap caller-supplied taskKey inside required plugin prefix so sessions remain visible to sendMessage/list/close

mrobinson2:
Closes #6156.

## Thinking Path

> - Paperclip orchestrates AI agents for zero-human companies
> - Plugins extend that orchestration through the Plugin SDK, which exposes `agents.sessions.*` as a session API for plugin-driven agent interactions (Telegram, Discord, etc.)
> - Plugins create sessions via `sessions.create()` and refer back to them via `sessions.sendMessage()`, `sessions.list()`, and `sessions.close()`
> - But `create` honored an arbitrary caller-supplied `taskKey` verbatim, while the three consumer ops hard-filter on `taskKey LIKE 'plugin:${pluginKey}:session:%'`
> - This means any plugin that passes a meaningful `taskKey` (e.g. `discord-thread-${threadId}` for cross-restart session lookup) creates a session that's immediately invisible to its own `sendMessage` calls — the session is orphaned, `sendMessage` throws `Session not found: <uuid>`
> - This PR wraps the caller's `taskKey` inside the required prefix so all four ops agree on the lookup pattern
> - The benefit is the documented `agents.sessions.*` API becomes usable for plugins that need stable, meaningful taskKeys

## What Changed

- `server/src/services/plugin-host-services.ts:1949` — `agentSessions.create` now wraps any caller-supplied `taskKey` inside the required `plugin:${pluginKey}:session:` prefix. Backwards-compatible for callers that pass no `taskKey` (still gets `plugin:${pluginKey}:session:${randomUUID()}`). Already-prefixed values are not double-wrapped (defensive `startsWith` check).
- `server/src/__tests__/plugin-orchestration-apis.test.ts` — three new integration tests (embedded-postgres backed, matches existing file pattern):
  - `wraps a caller-supplied taskKey in the plugin prefix so sendMessage/list/close can find it` — proves the fix end-to-end by re-running the same `LIKE` filter the consumer ops use.
  - `does not double-prefix an already-prefixed taskKey` — guards the new prefix-detection branch.
  - `with no taskKey produces a default prefixed key (backwards compatible)` — guards existing default behavior.
- Test cleanup now includes `agentTaskSessions` deletion in the `afterEach` so the new tests don't leave rows for subsequent runs.

## Verification

**Test:**

```
$ npx pnpm --filter @paperclipai/server exec vitest run src/__tests__/plugin-orchestration-apis.test.ts
[...]

PR #6629, opened 2026-05-24
fix(plugin-sdk): deliver prompt to agent runtime by materializing a backing issue in sessions.sendMessage / agents.invoke

### LINKED ISSUE #3461: sessions.sendMessage() prompt not passed to adapter execution context
## Problem

When a plugin calls `ctx.agents.sessions.sendMessage(sessionId, companyId, { prompt: text, reason: 'telegram_message' })`, the `prompt` text is placed into `payload: { prompt }` in the `heartbeat.wakeup()` call but is **never copied** into the `contextSnapshot` that reaches the adapter's `execute()` function.

### Root cause trace

1. `sendMessage` in `plugin-host-services.ts` (line ~1017) calls `heartbeat.wakeup()` with:
   - `payload: { prompt: params.prompt }` — prompt goes here
   - `contextSnapshot: { taskKey, wakeSource, wakeTriggerDetail }` — prompt NOT here

2. `enrichWakeContextSnapshot()` in `heartbeat.ts` only extracts `issueId`, `commentId`, `taskKey`, `wakeCommentId` from `payload`. It does **not** extract `prompt`.

3. `heartbeatRuns.contextSnapshot` is stored without the prompt.

4. When the adapter executes, `context = parseObject(run.contextSnapshot)` — no prompt.

5. The default `promptTemplate` renders to "You are agent X. Continue your Paperclip work." — the user's actual message is lost.

### Impact

Native sessions (`sendMessage`) cannot deliver user messages to agents. The agent wakes up but has no idea what the user asked — it falls back to heartbeat behavior and says "Inbox is empty."

This affects the Telegram plugin's native session transport (and any other plugin using `sessions.sendMessage` to deliver interactive messages).

### Proposed fix

Add `prompt: params.prompt` to the `contextSnapshot` in `sendMessage`:

```typescript
// server/src/services/plugin-host-services.ts, sendMessage()
contextSnapshot: {
  taskKey: session.taskKey,
  prompt: params.prompt,        // <-- ADD THIS
  wakeSource: "automation",
  wakeTriggerDetail: "system",
},
```

Then the adapter's `promptTemplate` can reference `{{context.prompt}}` to include the user's message in the stdin prompt.

### Workaround

We are currently applying this as a local patch (`patches/002-session-sendmessage-prompt-in-context.patch`) and setting agents' `promptTemplate` to `"You are agent {{agent.id}} ({{agent.name}}). {{context.prompt}}"`.

### Environment

- Paperclip commit: `a18f43b8`
- Plugin SDK: sessions API
- Adapter: `claude_local`

mrobinson2:
Closes #3461. Related: #6227 (taskKey contract mismatch — both must land for the plugin session API to function end-to-end).

## Thinking Path

> - Paperclip orchestrates AI agents and exposes plugin integrations through the Plugin SDK; `agents.sessions.sendMessage({prompt})` and `agents.invoke({prompt})` are the documented one-shot prompt entry points used by integrations like Discord/Telegram plugins
> - Both signatures accept a `prompt` parameter, but the prompt never reaches the agent's runtime: `heartbeat.wakeup` places `prompt` on the payload, but `enrichWakeContextSnapshot` in `server/src/services/heartbeat.ts` only extracts `issueId` / `commentId` / `taskKey` / `wakeCommentId` from payload — `prompt` is dropped on the floor
> - Downstream symptom: agent wakes and runs its default idle behavior (e.g., Alfred's "check the queue") instead of executing the prompt. The plugin sees a successful `sendMessage` return and assumes work was dispatched — but the prompt is silently discarded
> - The existing issue thread proposed adding `contextSnapshot.prompt` and a `{{context.prompt}}` template hook (requires per-adapter changes). The hermes adapter — and any adapter that consumes structured issue context rather than templated strings — needs a different shape
> - This PR routes the prompt through the verified-working issue-driven wake path: materialize a backing issue carrying the prompt as its description, then pass `issueId` in the wakeup payload. `enrichWakeContextSnapshot` already copies `issueId` into `contextSnapshot.issueId`, and every existing adapter reads prompts from issue body via that field — so no adapter changes are required
> - The benefit is the documented `agents.sessions.*` API becomes functional end-to-end for any adapter, with a full audit trail in the issue UI as a side effect

## What Changed

- `server/src/services/plugin-host-services.ts` — `agentSessions.sendMessage` and `agents.invoke` now materialize a backing issue (description = prompt, assignee = session.agentId) before calling `heartbeat.wakeup`. `issueId` flows into the wakeup payload and into the return value alongside `runId`.
- `packages/plugins/sdk/src/protocol.ts`, `types.ts` — `AgentSessionSendResult` and `agents.invoke` return shapes now include `issueId` (additive). Existing callers reading only `result.runId` continue to work.
- `packages/plugins/sdk/src/testing.ts` — in-memory testing harness mock updated to return both `runId` and `issueId`.
- `server/src/__tests__/plugin-orchestration-apis.test.ts` — three new integration tests (embedded-postgres backed, matches existing file pattern):
  - `sessions.sendMessage materializes a backing issue carrying the prompt and routes issueId into the run` — proves end-to-end that `contextSnapshot.issueId` resolves to a row whose description equals the caller's prompt.
  - `agents.invoke materializes a backing issue carrying the prompt and routes issueId into the run` — same guarantee for the one-shot invocation path.
  - Cleanup upgraded to `TRUNCATE ... CASCADE` so wakeup-populated tables (`agent_runtime_state`, `environment_leases`, `heartbeat_run_events`, `agent_task_sessions`) clean reliably between tests — this is the first test in the file that successfully fires a wakeup.

Backing issues are namespaced by `originKind`: `plugin:<pluginKey>:session` for `sendMessage` and `plugin:<pluginKey>:invocation` for `invoke`. `originId` on session-backed issues points at the session row, so plugins can filter their own backing issues from notification handlers if desired.

## Verification

**Test:**

```
$ npx vitest run server/src/__tests__/plugin-orchestration-apis.test.ts
 ✓ server/src/__tests__/plugin-orchestration-apis.test.ts  (12 tests)
   ✓ sessions.sendMessage materializes a backing issue carrying the prompt and routes issueId into the run
   ✓ agents.invoke materializes a backing issue carrying the prompt and routes issueId into the run
   ✓ ... (existing 10 tests)
 Test Files  1 passed (1)
      Tests  12 passed (12)
```

**End-to-end downstream:** verified against `paperclip-plugin-discord` `/acp spawn` — Discord thread → `sessions.create` → `sessions.sendMessage({prompt: "what is 9+1?"})` → Alfred (hermes adapter) returns "9 plus 1 equals 10" in the same Discord thread. Same fix shape (build-time monkey patch) has been running in our deployment since 2026-05-16.

## Backwards compatibility and risk

The new `issueId` field is additive in `AgentSessionSendResult` and the `agents.invoke` return shape. Callers that ignore the return value or read only `result.runId` are unaffected.

The side effect to note: `issue.created` fires for backing issues. Plugins that subscribe to `issue.created` and want to suppress notifications for plugin-generated issues should filter on `originKind=plugin_session` or `originKind=plugin_invocation`. The `paperclip-plugin-discord` PR for that filter is queued and will reference this PR.

No DB migration. The two new `originKind` values are free-form strings in the existing column.

## Alternative considered

The earlier proposal in #3461 was to copy `payload.prompt` into `contextSnapshot.prompt` and have each adapter read it (with `claude_local` doing `{{context.prompt}}` template substitution). That works for lightweight adapters but requires every other adapter to also implement context.prompt extraction. The backing-issue approach lights up every existing adapter (`hermes-paperclip-adapter`, `claude_local`, and others) without any adapter-side change — the cost is `~15` lines of `issueService.create()` plumbing inside the two host handlers and the audit-trail side effect on plugins that watch `issue.created`.

If maintainers prefer the combined approach (materialize backing issue **and** also copy `prompt` into `contextSnapshot`), the additional ~3 lines can be added on top of this PR without restructuring.

## Model Used

- Provider: Anthropic
- Model: claude-opus-4-7 (Opus 4.7, 1M context window)
- Mode: agentic coding with tool use (Read/Edit/Bash); no extended thinking mode used for this change
- Role: surfaced the root cause via downstream reproducer (`paperclip-plugin-discord` `/acp spawn`), drafted the fix + tests, wrote the PR body

## Checklist

- [x] I have run tests locally and they pass (`plugin-orchestration-apis.test.ts`: 12/12)
- [x] I have added or updated tests where applicable (2 new tests cover sendMessage + invoke paths; cleanup upgraded for wakeup-populated tables)
- [ ] If this change affects the UI, I have included before/after screenshots (n/a — server-only change)

### DIFF
--- a/packages/plugins/sdk/src/protocol.ts
+++ b/packages/plugins/sdk/src/protocol.ts
@@ -1066,7 +1066,7 @@ export interface WorkerToHostMethods {
   ];
   "agents.invoke": [
     params: { agentId: string; companyId: string; prompt: string; reason?: string },
-    result: { runId: string },
+    result: { runId: string; issueId: string },
   ];
   "agents.managed.get": [
     params: { agentKey: string; companyId: string },
@@ -1092,7 +1092,7 @@ export interface WorkerToHostMethods {
   ];
   "agents.sessions.sendMessage": [
     params: { sessionId: string; companyId: string; prompt: string; reason?: string },
-    result: { runId: string },
+    result: { runId: string; issueId: string },
   ];
   "agents.sessions.close": [
     params: { sessionId: string; companyId: string },
--- a/packages/plugins/sdk/src/testing.ts
+++ b/packages/plugins/sdk/src/testing.ts
@@ -1781,7 +1781,7 @@ export function createTestHarness(options: TestHarnessOptions): TestHarness {
         ) {
           throw new Error(`Agent is not invokable in its current state: ${agent!.status}`);
         }
-        return { runId: randomUUID() };
+        return { runId: randomUUID(), issueId: randomUUID() };
       },
       managed: {
         async get(agentKey, companyId) {
@@ -1923,7 +1923,7 @@ export function createTestHarness(options: TestHarnessOptions): TestHarness {
           if (opts.onEvent) {
             sessionEventCallbacks.set(sessionId, opts.onEvent);
           }
-          return { runId: randomUUID() };
+          return { runId: randomUUID(), issueId: randomUUID() };
         },
         async close(sessionId, companyId) {
           requireCapability(manifest, capabilitySet, "agent.sessions.close");
--- a/packages/plugins/sdk/src/types.ts
+++ b/packages/plugins/sdk/src/types.ts
@@ -1459,8 +1459,14 @@ export interface PluginAgentsClient {
   pause(agentId: string, companyId: string): Promise<Agent>;
   /** Resume a paused agent (sets status to idle). Throws if terminated, pending_approval, or not found. Requires `agents.resume`. */
   resume(agentId: string, companyId: string): Promise<Agent>;
-  /** Invoke (wake up) an agent with a prompt payload. Throws if paused, terminated, pending_approval, or not found. Requires `agents.invoke`. */
-  invoke(agentId: string, companyId: string, opts: { prompt: string; reason?: string }): Promise<{ runId: string }>;
+  /**
+   * Invoke (wake up) an agent with a prompt payload. Throws if paused, terminated, pending_approval, or not found.
+   * Materializes a backing issue (originKind `plugin:<pluginKey>:invocation`) carrying the prompt as its description so
+   * the agent adapter receives the prompt via `contextSnapshot.issueId`. The backing issue's id is returned so plugins
+   * can update or close it after the run completes.
+   * Requires `agents.invoke`.
+   */
+  invoke(agentId: string, companyId: string, opts: { prompt: string; reason?: string }): Promise<{ runId: string; issueId: string }>;
   /** Resolve and reconcile manifest-declared plugin-managed agents by stable key. Requires `agents.managed`. */
   managed: {
     get(agentKey: string, companyId: string): Promise<PluginManagedAgentResolution>;
@@ -1504,9 +1510,14 @@ export interface AgentSessionEvent {
 
 /**
  * Result of sending a message to a session.
+ *
+ * `issueId` is the id of the backing issue materialized to carry the prompt
+ * to the agent (originKind `plugin:<pluginKey>:session`). Plugins typically
+ * use it to update or close the issue after the run completes.
  */
 export interface AgentSessionSendResult {
   runId: string;
+  issueId: string;
 }
 
 /**
--- a/server/src/__tests__/plugin-orchestration-apis.test.ts
+++ b/server/src/__tests__/plugin-orchestration-apis.test.ts
@@ -6,12 +6,17 @@ import { and, eq } from "drizzle-orm";
 import { afterAll, afterEach, beforeAll, describe, expect, it } from "vitest";
 import {
   activityLog,
+  agentRuntimeState,
+  agentTaskSessions,
   agentWakeupRequests,
   agents,
   companies,
+  companySkills,
   costEvents,
   createDb,
+  environmentLeases,
   executionWorkspaces,
+  heartbeatRunEvents,
   heartbeatRuns,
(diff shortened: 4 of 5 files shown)


adamteale:
Independent confirmation of the underlying bug this fixes, from a different plugin/adapter context.

I hit this building a chat feature in a plugin (project-scoped ad-hoc conversation between a human and an agent, unrelated to Discord/Telegram — no adapter code involved at all on my end). Traced the exact same path described in #3461: `sendMessage`'s `payload: { prompt }` never reaches the rendered wake prompt, so every reply came from `DEFAULT_PAPERCLIP_AGENT_PROMPT_TEMPLATE` with zero trace of the actual message. Confirmed by reading a run's raw `heartbeat-runs/:id/log` output directly — the model's first message was the fully generic "You are agent X. Continue your Paperclip work." template, verbatim, regardless of what was passed as `prompt`.

Worth noting for whichever fix direction gets picked here: I ended up converging on the same backing-issue-materialization approach this PR takes (had no visibility into this PR or #3461 when I built it — arrived at it independently as the only reliable way to get real content in front of the model, since `ctx.issues.requestWakeup` + a comment is the one wake path that demonstrably renders real content today). That's a second, independent data point that the backing-issue approach is the right general shape for a fix here, for whatever that's worth to reviewers weighing this against the lighter `contextSnapshot.prompt` alternative discussed above.

One thing I'd flag from my own build: once a session's backing issue accumulates turns, Paperclip's own missing-disposition enforcement will unassign or reassign it on its own (since a chat-style issue never gets a real disposition) — worth checking this PR's approach doesn't hit the same thing, or note it as a known follow-up if it does.

PR #6631, opened 2026-05-24
fix(server): honor pre-set PAPERCLIP_RUNTIME_API_URL

### LINKED ISSUE #6630: Agent runtime API URL is conflated with allowedHostnames, can't point agents at loopback while serving the dashboard on a public hostname
### Summary

When the server runs behind a reverse proxy/tunnel (e.g. Cloudflare Tunnel) with `server.allowedHostnames` set to the public hostname for the browser Host-guard, that same hostname is forced onto **agents** as their API base URL. There is no supported way to keep agents talking to the server over loopback while the dashboard is reached via the public hostname.

### Environment

- `deploymentMode: local_trusted`, `exposure: private`, `bind: loopback`, `host: 127.0.0.1`, `port: 3100`
- `server.allowedHostnames: ["paperclip.example.com"]` (required so the private Host-guard in `app.ts` lets the browser in via the tunnel)
- Agents run as local subprocesses on the same host.

### What happens

Agents resolve their API base from `PAPERCLIP_RUNTIME_API_URL` first (`packages/adapter-utils/src/server-utils.ts:909`, `packages/adapter-utils/src/execution-target.ts:1090`).

`PAPERCLIP_RUNTIME_API_URL` is set **unconditionally** at startup (`server/src/index.ts:684`):

```js
process.env.PAPERCLIP_RUNTIME_API_URL = runtimeApiUrl;
```

`runtimeApiUrl` comes from `choosePrimaryRuntimeApiUrl()` (`server/src/runtime-api.ts:64-69`), which prefers the **first `allowedHostnames` entry** over the bind host. So agents get `http://paperclip.example.com:3100` and try to reach the server over the public hostname (which, behind a tunnel, isn't a usable origin for internal traffic — and is gated by Cloudflare Access).

### Why it can't be worked around by config/env today

- Setting `PAPERCLIP_API_URL` doesn't help — `PAPERCLIP_RUNTIME_API_URL` takes precedence in the agent env builders.
- `PAPERCLIP_RUNTIME_API_URL` itself is overwritten unconditionally at line 684, so a pre-set env value is ignored.
- The only knobs that change `runtimeApiUrl` (`allowedHostnames`, `auth.publicBaseUrl`) are the same ones that drive the browser Host-guard — changing them to loopback makes the dashboard return 403 / redirects login to `127.0.0.1`.
- Net: the internal agent callback URL and the external browser hostname are derived from one source and can't be decoupled.

### Proposed fix

Honor a pre-set `PAPERCLIP_RUNTIME_API_URL`, mirroring the pattern already used one line above for `PAPERCLIP_API_URL` (`index.ts:674`):

```js
process.env.PAPERCLIP_RUNTIME_API_URL =
  process.env.PAPERCLIP_RUNTIME_API_URL?.trim() || runtimeApiUrl;
```

This lets an operator pin agents to loopback (`PAPERCLIP_RUNTIME_API_URL=http://127.0.0.1:3100`) while `allowedHostnames` keeps serving the dashboard on the public hostname — fully decoupling internal traffic from the public origin, with no behavior change for existing setups.

(Alternatively / additionally: have `choosePrimaryRuntimeApiUrl` prefer the bind host over `allowedHostnames` for the *runtime* URL, since that URL is for internal callbacks, not browsers.)

Fl0p:
## Thinking Path

> - Agents run against the control-plane HTTP API, resolving their base URL from `PAPERCLIP_RUNTIME_API_URL` (internal agent→server callback)
> - The same server is reached by humans via the dashboard, whose Host-guard is configured through `server.allowedHostnames`
> - At startup `choosePrimaryRuntimeApiUrl()` derives the runtime URL from the first `allowedHostnames` entry, and `server/src/index.ts` then overwrites `PAPERCLIP_RUNTIME_API_URL` **unconditionally** with it
> - This conflates the internal agent callback URL with the public browser-facing hostname: behind a reverse proxy/tunnel (e.g. Cloudflare Tunnel) `allowedHostnames` must be the public host, which then forces agents to call the server over that public, access-gated origin instead of loopback — with no supported way to decouple the two
> - This pull request makes the startup assignment honor a pre-set `PAPERCLIP_RUNTIME_API_URL`, mirroring the `PAPERCLIP_API_URL` line directly above it
> - The benefit is operators can pin internal agent traffic to loopback while keeping the dashboard on a public hostname, decoupling internal callbacks from the public origin with zero change to existing deployments

## What Changed

- `server/src/index.ts`: `PAPERCLIP_RUNTIME_API_URL` is now set to `process.env.PAPERCLIP_RUNTIME_API_URL?.trim() || runtimeApiUrl` instead of being overwritten unconditionally — the same pre-set-then-fallback pattern already used for `PAPERCLIP_API_URL` one line above.

## Verification

- **Reproduce the original problem:** run with `server.allowedHostnames: ["pc.example.com"]` and `bind: loopback`. Before this change, agents receive `PAPERCLIP_RUNTIME_API_URL=http://pc.example.com:3100` and cannot reach the server over the public/tunnelled origin.
- **Confirm the fix:** set `PAPERCLIP_RUNTIME_API_URL=http://127.0.0.1:3100` before startup; agents now resolve their API base to loopback while the dashboard keeps serving on `pc.example.com`. Leave the env var unset and behavior is byte-for-byte identical to before.
- **Tests (run locally, all pass):**
  ```
  npx vitest run src/__tests__/paperclip-env.test.ts src/__tests__/runtime-api.test.ts
  # Test Files  2 passed (2)   Tests  9 passed (9)
  ```
  `paperclip-env.test.ts` already asserts the runtime→api precedence on the agent-consumption side that this change relies on. See the PR Checks tab for the automated CI run.

## Risks

- **Low risk.** The change only takes effect when `PAPERCLIP_RUNTIME_API_URL` is already set and non-blank; when unset it falls through to `runtimeApiUrl` exactly as before, so existing deployments are unaffected. It does not touch `PAPERCLIP_API_URL` (which stays the browser/webhook-facing external URL), so external webhook/CLI/MCP URLs are unchanged.

## Model Used

- Claude (Anthropic), model ID `claude-opus-4-7`, 1M-context variant, via Claude Code with extended thinking and tool use.

## Checklist

- [ ] I have added or updated tests where applicable — N/A: the conditional mirrors the existing `PAPERCLIP_API_URL` pattern whose precedence is already covered by `paperclip-env.test.ts`; the bare startup env assignment is not independently unit-tested.
- [ ] If this change affects the UI, I have included before/after screenshots — N/A, no UI change.
- [ ] I have updated relevant documentation to reflect my changes — N/A, no documented behavior changes (unset env = prior behavior).

Closes #6630


Fl0p:
Thanks for the review. The only finding was that the PR description didn't follow `.github/PULL_REQUEST_TEMPLATE.md`. I've rewritten the description to fully follow the template: **Thinking Path** (blockquote, project → subsystem → problem → fix → benefit), **What Changed**, **Verification** (repro + fix steps and the local test run), **Risks**, **Model Used** (`claude-opus-4-7`, 1M context, extended thinking + tool use), and a completed **Checklist**.

The code itself is unchanged (the one-line guard you scored as clean), so no new commit was needed. Requesting a re-review for the 5/5 score.

PR #6637, opened 2026-05-24
Auto-file onboarding 5xx errors as issues (GST-9)

kumage0503:
## Thinking Path

> - Paperclip orchestrates AI agents for zero-human companies, and onboarding (`POST /api/companies`, goals, agents, projects, issues) is the very first surface a human touches
> - When an onboarding endpoint 5xx's, the wizard renders a generic "Internal server error" with no incident id, no trail in any company, and no signal back to the operating team — see GST-5 forensics
> - This is a silent-failure UX gap: the error is captured in stderr logs only, so the human sees a dead end and the company that should own the bug never gets the issue filed
> - GST-9 closes the loop server-side: classify onboarding traffic, intercept 5xx, redact + dedup, file a backlogged issue into the relevant Paperclip company, and surface an `incidentId` to the browser for GST-5a to render
> - The hard part is `POST /api/companies` itself — there is no company yet to file into, so the incident is deferred to disk under `<instance-root>/onboarding-incidents/` and ingested on the next successful company creation by the same user
> - Cross-user leak guard, atomic disk writes, 30-day decay, and a 24h dedup window keep the deferred-ingest pattern safe to run in production
> - This PR is additive and isolated: no existing onboarding code path is changed, and every callsite swallows failures so the user-facing onboarding response is never blocked by telemetry

## What Changed

- New middleware `server/src/middleware/onboarding-incident-reporter.ts` — classifies onboarding requests (allowlist / `X-Paperclip-Onboarding` header / `Referer: /onboarding`), patches `res.json` to inject `incidentId` on 5xx, and records the incident asynchronously on the `finish` event.
- New service `server/src/services/onboarding-incidents.ts` — redacts sensitive headers + body keys, SHA-256 dedups within 24h, files into the company immediately when one exists, defers to disk (atomic `writeFile + rename`) otherwise, and ingests deferred files on the next successful `POST /api/companies` by the same user. 30-day decay sweep is opportunistic per record.
- `server/src/app.ts` wires `onboardingIncidentReporter` right after `actorMiddleware`.
- `server/src/routes/companies.ts` calls `ingestPendingIncidents` inside `POST /api/companies` after company creation, in a try/catch so company creation is never blocked.
- 6 integration tests in `server/src/__tests__/onboarding-incident-reporter.test.ts` covering happy-path filing, dedup, defer-and-ingest, cross-user leak guard, and the negative non-onboarding case. All against embedded Postgres.
- Follow-up commit (`f8b57be`) resolves Greptile P2 review: truncate body cap on the last complete UTF-8 code point so the stored body never ends in U+FFFD, and bound deferred-ingest scan to 100 files per request so a noisy directory cannot turn `POST /api/companies` into an O(files) disk walk.

### Onboarding endpoint classification

Any one of:
1. `Referer` header path starts with `/onboarding`, OR
2. `X-Paperclip-Onboarding: 1` header is set (the wizard sets this; see GST-5a), OR
3. Route matches the explicit allowlist: `POST /api/companies`, `POST /api/companies/:companyId/goals`, `POST /api/companies/:companyId/agents`, `POST /api/companies/:companyId/agents/:agentId/adapter-environment-test`, `POST /api/companies/:companyId/projects`, `POST /api/companies/:companyId/issues`.

### Architectural decisions (locked in GST-9 plan doc)

- **Defer-and-ingest** over a hidden system company. Files live under `<instance-root>/onboarding-incidents/<ts>-<uuid>.json` (0600, atomic rename), decay after 30 days, capped at 100 inspected per opportunistic sweep.
- **Cross-user leak guard:** ingestion requires `incident.createdByUserId === newCompany.creatorUserId`, OR both sides are `local_implicit` with no user identity. Anything else stays on disk for a later matching creator.
- **De-dup window:** `sha256(method + route + error.name + error.message)` within 24h; repeat hits drop a counter comment on the existing issue instead of creating a duplicate. Race-tolerant — parallel hits may occasionally double-file, accepted.
- **Middleware placement:** right after `actorMiddleware`, patches `res.json` so the 5xx body returned to the browser carries `incidentId`, and uses `res.on("finish")` to read the error context attached by `errorHandler` and record asynchronously. Standard Express "post-error" pattern; the brief's literal "after errorHandler" wording isn't semantically possible.
- **Redaction:** request body runs through `redactCurrentUserValue` (OS username + home), then sensitive keys (`password`, `secret`, `token`, `apiKey`, `accessToken`, …) → `[redacted]`. Headers `authorization`, `cookie`, `x-paperclip-api-key`, `x-paperclip-board-key` are stripped. Body truncated to 8 KiB on a UTF-8 code-point boundary; header values to 256 bytes.

## Verification

- `pnpm --filter @paperclipai/server exec vitest run src/__tests__/onboarding-incident-reporter.test.ts` — 6/6 passing against embedded Postgres.
- Manual smoke (reviewer / QA, post-merge): force a 5xx on `POST /api/companies` (e.g. `{"name":"pdeo"}` against an instance with PDE taken). Confirm the response body has `incidentId`. Onboard a different name successfully → confirm the deferred incident is ingested as a backlogged issue in the new company and the disk file is gone.

## Risks

- **Label upsert race (accepted).** Two parallel first-time labelers can 23505 on `labels_company_name_idx`; second incident is still filed without the label. Operator-visible, rare.
- **Dedup race (accepted).** Two parallel 5xx hits with the same `{method, route, error.name, error.message}` in the same 24h window can both miss the dedup lookup and create two issues. Acceptable for auto-filing.
- **Orphan `.tmp` files.** A process crash between `writeFile` and `rename` leaves `<id>.tmp` behind. Decay only sweeps `.json`. Operator-visible; low priority.
- **Disk-space growth.** Deferred files accumulate if a single-user instance keeps 5xx'ing on `POST /api/companies`. 30-day decay sweep bounds this; per-request ingest scan is now capped at 100 files (commit `f8b57be`) so a noisy directory cannot block onboarding.
- **`__testOnly` exposed on the production interface.** Considered carving it into a separate test seam; cost/benefit did not warrant the extra file. Documented in Staff Engineer self-review.

## Model Used

- Claude (Anthropic), `claude-opus-4-7` (1M context), extended-thinking on, tool use enabled, harness: Claude Code via Paperclip Release Engineer adapter (`claude_local`).

## Checklist

- [ ] If this change affects the UI, I have included before/after screenshots — N/A, server-only change; UI surface lives in GST-5a

---

Refs: GST-9, GST-5 plan.

### DIFF
--- a/server/src/__tests__/onboarding-incident-reporter.test.ts
+++ b/server/src/__tests__/onboarding-incident-reporter.test.ts
@@ -1,0 +1,334 @@
+import { promises as fs } from "node:fs";
+import os from "node:os";
+import path from "node:path";
+import express from "express";
+import request from "supertest";
+import { eq } from "drizzle-orm";
+import { afterAll, afterEach, beforeAll, beforeEach, describe, expect, it } from "vitest";
+import { companies, createDb, issueComments, issueLabels, issues, labels } from "@paperclipai/db";
+import { errorHandler } from "../middleware/index.js";
+import { onboardingIncidentReporter } from "../middleware/onboarding-incident-reporter.js";
+import {
+  AUTO_FILED_ONBOARDING_5XX_LABEL_NAME,
+  AUTO_FILED_ONBOARDING_5XX_ORIGIN_KIND,
+  onboardingIncidentsService,
+} from "../services/onboarding-incidents.js";
+import { HttpError } from "../errors.js";
+import {
+  getEmbeddedPostgresTestSupport,
+  startEmbeddedPostgresTestDatabase,
+} from "./helpers/embedded-postgres.js";
+
+const embeddedPostgresSupport = await getEmbeddedPostgresTestSupport();
+const describeEmbeddedPostgres = embeddedPostgresSupport.supported ? describe : describe.skip;
+if (!embeddedPostgresSupport.supported) {
+  console.warn(
+    `Skipping embedded Postgres onboarding incident tests on this host: ${embeddedPostgresSupport.reason ?? "unsupported environment"}`,
+  );
+}
+
+type Db = ReturnType<typeof createDb>;
+type RecordSettlement = {
+  incidentId: string;
+  ok: boolean;
+  error?: unknown;
+  outcome?: { filed: string; issueId?: string };
+};
+
+interface IncidentBarrier {
+  onSettled(value: RecordSettlement): void;
+  wait(): Promise<RecordSettlement>;
+}
+
+interface AppHandle {
+  app: express.Express;
+  incidentDir: string;
+  cleanup: () => Promise<void>;
+  waitForNextIncident(): Promise<RecordSettlement>;
+}
+
+function buildIncidentBarrier(): IncidentBarrier {
+  const queue: RecordSettlement[] = [];
+  const waiters: Array<(value: RecordSettlement) => void> = [];
+  return {
+    onSettled(value) {
+      const waiter = waiters.shift();
+      if (waiter) waiter(value);
+      else queue.push(value);
+    },
+    wait() {
+      const next = queue.shift();
+      if (next) return Promise.resolve(next);
+      return new Promise((resolve) => waiters.push(resolve));
+    },
+  };
+}
+
+function buildApp(
+  db: Db,
+  incidentDir: string,
+  opts: {
+    actor?: express.Request["actor"];
+    barrier?: IncidentBarrier;
+  } = {},
+): express.Express {
+  const incidents = onboardingIncidentsService(db, { incidentDir });
+  const app = express();
+  app.use(express.json());
+  app.use((req, _res, next) => {
+    req.actor = opts.actor
+      ?? ({
+        type: "board",
+        userId: "user-1",
+        userName: "Test Board User",
+        userEmail: null,
+        isInstanceAdmin: true,
+        source: "session",
+      } as express.Request["actor"]);
+    next();
+  });
+  app.use(
+    onboardingIncidentReporter({
+      incidents,
+      onRecordSettled: opts.barrier?.onSettled,
+    }),
+  );
+
+  app.post("/api/companies", (_req, _res, next) => {
+    next(new HttpError(500, "Internal server error"));
+  });
+  app.post("/api/companies/:companyId/projects", (_req, _res, next) => {
+    next(new HttpError(500, "Internal server error"));
+  });
+  app.post("/api/probe-non-onboarding", (_req, _res, next) => {
+    next(new HttpError(500, "Internal server error"));
+  });
+
+  app.use(errorHandler);
+  return app;
+}
+
+async function withTempIncidentDir(): Promise<{ dir: string; cleanup: () => Promise<void> }> {
+  const dir = await fs.mkdtemp(path.join(os.tmpdir(), "paperclip-incidents-"));
+  return {
+    dir,
+    cleanup: async () => {
+      await fs.rm(dir, { recursive: true, force: true });
+    },
+  };
+}
+
+function newAppHandle(
+  db: Db,
+  opts: { actor?: express.Request["actor"] } = {},
+): Promise<AppHandle> {
+  return (async () => {
+    const tmp = await withTempIncidentDir();
+    const barrier = buildIncidentBarrier();
+    const app = buildApp(db, tmp.dir, { ...opts, barrier });
+    return {
+      app,
+      incidentDir: tmp.dir,
+      cleanup: tmp.cleanup,
+      waitForNextIncident: () => barrier.wait(),
+    };
+  })();
+}
+
+async function insertCompany(db: Db, name = "Acme Inc") {
+  const [row] = await db
+    .insert(companies)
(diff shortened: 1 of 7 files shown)


kumage0503:
## Staff Engineer self-review (paranoid pass)

Approved to merge, with the following acknowledged risks. None block landing.

### What I checked

- **Trust boundaries.** `onboardingIncidentReporter` only ever runs after `actorMiddleware` so `req.actor` is populated. The `incidentId` injection only fires on 5xx and only spreads into JSON object bodies — Array/Buffer/null bodies pass through untouched. Sensitive headers (`authorization`, `cookie`, `x-paperclip-api-key`, `x-paperclip-board-key`) are dropped before storage. Sensitive body keys are case-insensitively replaced with `[redacted]`. The `redactCurrentUserValue` helper strips the OS username + home dir from all string values, including the stack.
- **SQL safety.** Service uses `issueService.list()` / `issueService.create()` / `issueService.addComment()` only. No raw SQL added in this diff; existing parameterized Drizzle queries.
- **Race conditions.**
  - **Dedup race (accepted):** two parallel 5xx hits with the same `{method, route, error.name, error.message}` in the same 24h window can both pass the dedup lookup and create two issues. Acceptable for auto-filing.
  - **Label upsert race (accepted, soft):** concurrent `getOrCreateLabel` could 23505 on `labels_company_name_idx`. My `try/catch` returns `null`, so the second issue is filed without the label. Issue still gets filed, dedup still works, only the label tag is missing on the racing issue. Operator-visible and rare.
  - **Disk write atomic:** `writeFile(.tmp)` + `fs.rename(.json)` so no half-written records are visible to ingest.
- **Stale reads.** `findDedupIssue` reads issues with `originId = dedupHash`. If two creators land at the same instant, both might see an empty result and both insert — same race as above.
- **Cross-tenant leakage.** Deferred incidents are only ingested when `incident.createdByUserId === newCompany.creatorUserId`, OR both sides are `local_implicit` with no user identity. The latter only matches a single-user local instance. Verified by a dedicated test (`does NOT ingest incidents that belong to a different user`).
- **Conditional side effects in the success path.** `POST /api/companies` wraps the `ingestPendingIncidents` call in try/catch — a failure there can never reject the company-creation response. The 25-incident-per-request cap bounds the in-band work.
- **LLM trust boundary.** N/A — no LLM calls in this diff.
- **Tests that pass while missing the real failure mode.**
  - The "non-onboarding 5xx" negative test asserts both no disk file AND no DB row.
  - The "ingests pending incidents" test verifies the file is deleted after ingest (catches a regression where ingest reports success but leaves the file behind, which would re-ingest forever).
  - The cross-user leak test verifies the file STAYS behind when ingestion is skipped.
  - The dedup test verifies a counter comment is created on the existing issue.
  - All tests use a real embedded Postgres via the existing test harness — not mocks.

### Filed risks (intentional, ship)

1. **Label upsert race.** Could swallow a label assignment on rare concurrent first-time labelers. Fixable with a retry-after-conflict; not worth the complexity right now.
2. **Orphan `.tmp` files.** A process crash between `writeFile` and `rename` leaves `<id>.tmp`. Decay only sweeps `.json`. Operator-visible. Low priority.
3. **`onRecordSettled` is a documented test-only hook on production interface.** I considered carving it into a separate test seam but the cost/benefit didn't warrant the extra file.

### Verification reproduced

- `pnpm --filter @paperclipai/server exec vitest run src/__tests__/onboarding-incident-reporter.test.ts` — 6/6 passing.

Handing to Release Engineer for ship.

kumage0503:
## Release Engineer — Greptile review pass

Addressed the two P2 items that affect correctness/perf; deferred two items that the Staff Engineer self-review already considered and accepted as ship-as-is. PR description has been updated to satisfy the missing template sections.

### Fixed (commit `f8b57be`)

- **P2 – Byte-boundary UTF-8 truncation.** `serializeBodyWithCap` now trims to the last complete UTF-8 code point via a new `truncateUtf8ByBytes` helper, so the stored body never ends in a stray `U+FFFD`. Same 8 KiB cap.
- **P2 – `ingestPendingIncidents` unbounded scan.** Added an `INCIDENT_INGEST_SCAN_LIMIT = 100` cap on total files processed per `POST /api/companies`. The existing 25-issue-per-request filing cap stays the same; this only bounds the disk walk when the directory accumulates files that all get skipped (cross-user, wrong version, etc.).

### Acknowledged, not changing in this PR

- **P2 – `__testOnly` on the production interface.** Staff Engineer self-review explicitly considered this and decided the cost/benefit didn't warrant the extra test-seam file. Holding to that decision.
- **P2 – Duplicate service instance in `app.ts` + `companies.ts`.** Both resolve to the same `incidentDir` via `resolvePaperclipInstanceRoot()`, and `incidentDir` isn't currently configurable. The coupling is implicit but harmless today; refactoring into a singleton/DI is worth its own follow-up rather than expanding the diff here.

### Verification

- `pnpm --filter @paperclipai/server exec vitest run src/__tests__/onboarding-incident-reporter.test.ts` — 6/6 passing.

PR description has been updated with **Thinking Path**, **What Changed**, **Verification**, **Risks**, **Model Used**, and **Checklist** sections per `.github/PULL_REQUEST_TEMPLATE.md`.

Handing to CTO for final scan.

PR #6641, opened 2026-05-24
feat(server): exponential backoff for process_lost retries (SPC-6121)

notandrewblejde:
## Thinking Path

> - Paperclip orchestrates AI agents for zero-human companies; the harness retries failed runs so an agent's progress survives transient infra blips.
> - The heartbeat scheduler retries process_lost runs through two paths: the orphan reaper (`enqueueProcessLossRetry` in `server/src/services/heartbeat.ts`) and the stranded-issue continuation sweep (`reconcileStrandedAssignedIssues` in `server/src/services/recovery/service.ts`).
> - On 2026-05-24 a ~48-minute host-unavailable event hit a downstream Paperclip deployment. Each retry path fired at its native ~11s cadence with no inter-retry delay, alternating between the two paths because each created runs that the other did not recognise as already retried. The result: 433 wasted retries across two issues (233 on one, 200 on the other) before the host recovered.
> - The flat cadence amplified a transient infra event into hundreds of lease acquisitions, telemetry-loss noise, and false productivity-review escalations. The downstream post-mortem requested an exponential delay between retries once a same-issue process_lost storm is detected.
> - This pull request adds the canonical schedule from the spec (1-4 → immediate, 5 → 30s, 6 → 60s, 7 → 120s, 8+ → 300s cap, counter resets on a successful run, 5-minute consecutive-failure window) and wires it into both retry paths.
> - The benefit is ≤20 retries per affected issue in a 48-minute outage (vs. 233+ today), reducing lease compute waste from ~40 min to ~5 min per affected issue. No change to the happy path: an isolated process_lost still gets one immediate retry exactly as before.

## What Changed

- New helper module `server/src/services/heartbeat-process-lost-backoff.ts` exporting the constants (`PROCESS_LOST_BACKOFF_THRESHOLD=5`, `PROCESS_LOST_BACKOFF_WINDOW_MS=5min`, `PROCESS_LOST_BACKOFF_DELAYS_MS=[30s,60s,120s,300s]`, `PROCESS_LOST_BACKOFF_RETRY_REASON='process_lost_backoff'`, `PROCESS_LOST_BACKOFF_WAKE_REASON='process_lost_backoff_retry'`) and two functions:
  - `deriveProcessLostBackoffDelayMs(consecutiveCount)` — pure schedule lookup.
[...]

PR #6465, opened 2026-05-20
fix(recovery): clear stranded_assigned_issue recovery on issue disposition changes

noahlaux:
## Thinking Path

> - Paperclip orchestrates AI agents for zero-human companies; reliability of the recovery lifecycle is critical for unblocking stalled work
> - The recovery subsystem detects stranded issues and issues `wake_owner` heartbeat wakes to the assigned agent
> - Without a cap, `wake_owner` wakes fire indefinitely on unresolvable issues, burning agent budget and flooding the inbox with duplicate context
> - Adding a `stale` flag lets the platform stop issuing wakes once the agent has had a fixed number of chances to act
> - This PR adds `stale: boolean` to the DB schema and recovery action read-model, computes it on every upsert, and uses it as an enqueue guard
> - The benefit is predictable, bounded wake costs and a queryable surface (`?stale=true`) for the Daily Operations Digest to surface neglected issues

## What Changed

- **DB/Schema** (`packages/db/src/schema/issue_recovery_actions.ts`, migration `0087_recovery_action_stale.sql`): add `stale boolean NOT NULL DEFAULT false` column and covering index on `(company_id, stale, status)`
- **Shared types** (`packages/shared/src/types/issue.ts`): add `stale: boolean` to `IssueRecoveryAction` interface
- **Recovery action service** (`server/src/services/issue-recovery-actions.ts`): compute `stale` on every upsert (monotone once set); add `listStale(companyId, { assigneeAgentId? })` method
- **Recovery orchestration** (`server/src/services/recovery/service.ts`): set `maxAttempts: 5` (DEFAULT_WAKE_OWNER_MAX_ATTEMPTS) for `wake_owner` stranded-issue recovery; guard `enqueueSourceScopedStrandedRecoveryWake` to skip when `action.stale === true`
- **Route** (`server/src/routes/issues.ts`): `GET /api/companies/:companyId/recovery-actions?stale=true[&assigneeAgentId=...]` returns `{ actions: IssueRecoveryAction[] }`
- Also includes: `POST /api/recovery-actions/:id/resolve` endpoint for issue-scoped resolve-by-id (YOU-469)

## Verification

- TypeScript: `cd server && pnpm --filter @paperclipai/server run typecheck` — zero `error TS` lines
- Migration: `0087_recovery_action_stale.sql` adds column + index; Drizzle journal entry at idx 87
[...]

PR #6643, opened 2026-05-24
fix(heartbeat): per-issue process_lost retry budget circuit breaker (SPC-6120)

notandrewblejde:
## Summary

- The 2026-05-24 process_lost storm (SPC-6113) burned 433 wasted run cycles in 47 minutes because the existing `processLossRetryCount < 1` cap is per-run; every fresh retry resets it to 0 and a new infra-side `process_lost` immediately re-queues.
- New per-issue counter `issues.consecutive_process_lost_count` (migration `0090_issue_process_lost_budget.sql`) is incremented on every `process_lost` by the reaper and reset to 0 on any successful run for the same issue.
- When the counter reaches `PROCESS_LOST_PER_ISSUE_RETRY_BUDGET = 5`, the reaper skips the retry queue, flips the issue to `status='blocked'`, and tags the lifecycle event payload with `reason='harness_process_lost_budget_exhausted'` so HoOps sees a single blocked issue instead of a retry storm.

## Scope

This is Fix 3 from SPC-6120's 5-fix plan. Landing it first bounds the storm at its endpoint regardless of which retry path queues it. Fixes 1 (persist `scheduledRetryAttempt`), 2 (capture process-death reason in lease), 4 (tag infra-side process_lost in monitoring), 5 (workspace cleanup on process_lost) ship as child issues for defense-in-depth.

## Test plan

- [x] `pnpm --filter @paperclipai/server typecheck` clean
- [x] `vitest run src/__tests__/heartbeat-process-recovery.test.ts` — new tests pass:
  - [x] budget exhaustion → issue blocked, no retry queued, exhaustion event emitted with `consecutiveProcessLostCount: 5` + `reason: harness_process_lost_budget_exhausted`
  - [x] budget below threshold → counter increments to 1, retry still queued
- [x] Existing regression tests still pass (queues one retry / blocks on retry exhaustion / lease release)
- [ ] Reviewer: confirm `0090_issue_process_lost_budget.sql` is idempotent against the gamma DB (uses `ADD COLUMN IF NOT EXISTS … DEFAULT 0 NOT NULL`)

### DIFF
--- a/packages/db/src/migrations/0090_issue_process_lost_budget.sql
+++ b/packages/db/src/migrations/0090_issue_process_lost_budget.sql
@@ -1,0 +1,1 @@
+ALTER TABLE "issues" ADD COLUMN IF NOT EXISTS "consecutive_process_lost_count" integer DEFAULT 0 NOT NULL;
--- a/packages/db/src/migrations/meta/_journal.json
+++ b/packages/db/src/migrations/meta/_journal.json
@@ -631,6 +631,13 @@
       "when": 1779129600000,
       "tag": "0089_cloud_upstreams",
       "breakpoints": true
+    },
+    {
+      "idx": 90,
+      "version": "7",
+      "when": 1779662400000,
+      "tag": "0090_issue_process_lost_budget",
+      "breakpoints": true
     }
   ]
 }
--- a/packages/db/src/schema/issues.ts
+++ b/packages/db/src/schema/issues.ts
@@ -57,6 +57,7 @@ export const issues = pgTable(
     monitorAttemptCount: integer("monitor_attempt_count").notNull().default(0),
     monitorNotes: text("monitor_notes"),
     monitorScheduledBy: text("monitor_scheduled_by"),
+    consecutiveProcessLostCount: integer("consecutive_process_lost_count").notNull().default(0),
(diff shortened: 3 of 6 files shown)


koyouko (commented):
Thanks for this PR. The per-issue retry budget is a solid circuit breaker that complements #6641's exponential backoff.

Two notes on the interaction with the sibling PRs:

**1. Budget threshold vs backoff threshold** — Both use 5 consecutive failures as the trigger point. With #6641 merged first, the 6th retry already backs off to 60s. Consider whether the budget should count only *expedited* (non-backoff) retries, or use a higher budget (e.g. 10) so the two mechanisms don't stack too aggressively. Currently a budget-exhausted block would fire on the 6th failure, right when backoff kicks in — probably fine, but worth being explicit about the interaction.

**2. Merge conflict** — This conflicts with #6641 on `heartbeat.ts`. Recommend: merge #6641 first, then rebase this on top. The budget code and backoff code need to coexist in the same heartbeat processing path.

**3. Reset path** — The `resetIssueProcessLostBudget` function looks correct. Combined with #6465's recovery-action clearing on disposition change, the issue will correctly auto-recover when an agent marks it blocked/done.

The migration (0090) and schema change are clean. Approving conceptually — needs rebase after #6641 lands.

PR #6685, opened 2026-05-25
fix: routines PATCH endpoint must persist labelIds

### LINKED ISSUE #6684: fix: routines PATCH endpoint must persist labelIds
## Problem

The `PATCH /api/routines/{id}` endpoint accepts `labelIds` in the request body but silently ignores it because the Zod validator only recognizes `executionLabelIds`. The field appears absent (null from client perspective) in GET responses since the API surface uses `executionLabelIds` while callers expect `labelIds` (consistent with issues API).

## Root Cause

1. `updateRoutineSchema` and `createRoutineSchema` validators do not include `labelIds` — Zod strips it on parse
2. `create()` service method does not pass `executionLabelIds` to the DB insert (always uses DB default `[]`)
3. GET/PATCH responses expose `executionLabelIds` but not `labelIds`

## Fix

- Add `labelIds` to both `createRoutineSchema` and `updateRoutineSchema`
- In `update()`, prefer `patch.labelIds` over `patch.executionLabelIds` (backward compat)
- In `create()`, pass `executionLabelIds: input.labelIds ?? input.executionLabelIds ?? []` to the insert
- Add `labelIds` to `Routine` type and include it in all route responses

## Acceptance Criteria

- `PATCH /api/routines/{routineId}` with `{ "labelIds": ["id1"] }` persists the values
- `GET /api/routines/{routineId}` returns set `labelIds` (not null/absent)
- Consistent with how `labelIds` works on issues

NexusMBD:
## Summary

- `PATCH /api/routines/{id}` with `{ "labelIds": [...] }` now persists the label IDs
- `GET /api/routines/{id}` (and list/create) responses now include a `labelIds` field aliased from `executionLabelIds`
- `create()` service method now correctly writes `executionLabelIds` to the DB (was always inserting `[]` due to missing field in insert values)

## Root Causes Fixed

1. `createRoutineSchema`/`updateRoutineSchema` did not include `labelIds` — Zod stripped it on every PATCH
2. `create()` never passed `executionLabelIds` to the DB insert, so label IDs set at creation time were silently dropped
3. API responses exposed only `executionLabelIds`; callers expecting `labelIds` (consistent with issues API) got `undefined`/null

## Test plan

- [ ] `PATCH /api/routines/{id}` with `{ "labelIds": ["<uuid>"] }` — verify GET returns `labelIds: ["<uuid>"]`
- [ ] `PATCH /api/routines/{id}` with `{ "executionLabelIds": ["<uuid>"] }` — still works (backward compat)
- [ ] `POST /api/companies/{id}/routines` with `{ "labelIds": [...] }` — verify create persists labels
- [ ] `GET /api/companies/{id}/routines` — each item includes `labelIds` array

Fixes #6684


NexusMBD:
Closing as superseded by #7276, which is a rebased and more complete version of this routines labelIds feature. #7276 also adds label ID validation and portability export/import support that this PR was missing.

PR #6688, opened 2026-05-25
Fix the blocked-issue UX loop on the board

zionkim20:
## Thinking Path

> - Issues are the primary interface between humans and agents — humans drop a comment, agents wake up and work
> - The wake/work loop has a sharp edge when an issue is blocked: every comment on a gated issue triggers a wake for the assignee, the assignee runs against work it cannot make progress on, exits in ~5 seconds, the harness flags the short exit as "missing issue comment" and retries, looping
> - The user-visible side of this is just as bad: the board shows blocked items with no hint of who to ping to unblock them, and there's no way to hide them while you focus on what you *can* move
> - This pull request closes the loop and gives the user the information they need to act
> - The benefit is fewer wasted agent runs, less harness retry churn, and an inbox that surfaces work the user can actually progress

## What Changed

- **Stop waking the assignee on a `blocked` issue when the comment does not @-mention them and did not reopen the issue.** Mentioned-agent wakes (including for the assignee, if mentioned) and interaction-response wakes (`request_confirmation` / `ask_user_questions`) flow through separate code paths and still fire as before. Both comment entry points are covered: `POST /api/issues/:id/comments` and the comment-via-`PATCH /api/issues/:id` path. The skipped path emits a structured log (`wakeSkippedReason: "issue_blocked_no_mention"`) for observability.
- **Add a "Hide blocked" toggle to the issue filter popover, default ON.** Hides issues where `status === "blocked"` or `blockerAttention.unresolvedBlockerCount > 0`. State persists per-user through the existing localStorage flow. Counted as an active filter only when the user explicitly turns the default off.
- **Expand the blocked-issue banner to show, per blocker, a status pill, the assignee's display name (agent name resolved through the existing `agentMap`; user via `formatAssigneeUserLabel`), and a "Ping" button.** The button navigates to the blocking issue with `?focus=composer`, which a new one-shot effect in the chat composer picks up to scroll into view and focus the editor. The query param is then stripped from the URL so reloads don't yank the page back.
- Tests: extended `issue-filters.test.ts` and `IssueBlockedNotice.test.tsx` with coverage for the new behaviors; added two tests to `issue-update-comment-wakeup-routes.test.ts` for the skip and the mention-overrides-skip cases; **rewrote** two tests in `issue-comment-reopen-routes.test.ts` that previously asserted the now-removed wake — they enshrined the exact loop this PR closes. Updated `inbox.test.ts` literals to carry the new `hideBlocked` field.

## Verification

```bash
pnpm install --frozen-lockfile
pnpm typecheck     # all workspaces green
pnpm build         # all workspaces green
pnpm -F @paperclipai/server exec vitest run "src/__tests__/issue"   # 34 files, 355 tests pass
pnpm -F @paperclipai/ui exec vitest run \
  src/lib/issue-filters.test.ts \
  src/components/IssueBlockedNotice.test.tsx \
  src/components/IssueFiltersPopover.test.tsx \
  src/lib/inbox.test.ts                                              # 69 tests pass
```

Manual smoke (board running on `127.0.0.1:3100`):

1. Open the board with a mix of `blocked` and `todo` issues. The board should default to hiding blocked items; the filter popover shows the new "Hide blocked" checkbox. Untick it — the blocked items return.
2. Open a blocked issue. The amber banner now lists each blocker as a card with a status pill ("In review", etc.), an "Assigned to <Name>" row, and a "Ping" button.
3. Click "Ping". You should land on the blocker's issue page with the composer scrolled into view and focused.
4. Post a generic comment on a blocked issue (no `@-mention`). The assignee should *not* get woken. Confirm via `agent_wakeup_requests` table or server logs (`wakeSkippedReason: "issue_blocked_no_mention"`).
5. Post a comment that `@-mentions` the assignee. The mention wake fires as normal.

## Risks

- **Behavioral change for boards that relied on every comment waking the assignee, even when the issue was still blocked.** This was the loop. Two tests in `issue-comment-reopen-routes.test.ts` that asserted the old behavior are rewritten to assert the new one. If a deployment depended on the old wake to surface comments to a blocked-issue assignee, the recommended migration is to `@-mention` the agent — that path still wakes them, and is also the path Zion's plain-language standard already encourages.
- **`hideBlocked: true` is the new persisted default.** Users with no prior view state get blocked items hidden on first load. Users with existing stored view state are upgraded the same way because the normalizer treats a missing/non-boolean `hideBlocked` as `true`. A user who actively wants the old behavior can flip the toggle once.
- **`?focus=composer` writes to `history.replaceState`.** One-shot on mount; cleans itself out of the URL. Should not interact with the router's own navigation tracking.
- Low risk on the shared type addition: `assigneeAgentName` and `assigneeUserName` are optional, all existing producers continue to compile.

## Model Used

- Anthropic Claude Opus 4.7 (1M context), reasoning on, tool use. Provider: Anthropic. Exact model ID: `claude-opus-4-7[1m]`.

## Checklist

- [ ] If this change affects the UI, I have included before/after screenshots — *deferred: this PR replaces the running local Paperclip daemon on the author's machine, screenshots will be attached in a follow-up comment once the daemon is restarted (see HUM-126 in the linked tracker)*
- [x] I have updated relevant documentation to reflect my changes — *no docs touched; behavior is described in code comments at the call sites and in this PR body*

### DIFF
--- a/packages/shared/src/types/issue.ts
+++ b/packages/shared/src/types/issue.ts
@@ -137,6 +137,11 @@ export interface IssueRelationIssueSummary {
   priority: IssuePriority;
   assigneeAgentId: string | null;
   assigneeUserId: string | null;
+  // HUM-126 Fix 2: optional display labels so the blocked-issue banner can show
+  // "who to ping" without the client doing a second round trip. Producers that
+  // don't have this resolved (e.g. legacy paths) can omit it.
+  assigneeAgentName?: string | null;
+  assigneeUserName?: string | null;
   terminalBlockers?: IssueRelationIssueSummary[];
   activeRecoveryAction?: IssueRecoveryAction | null;
 }
--- a/server/src/__tests__/issue-comment-reopen-routes.test.ts
+++ b/server/src/__tests__/issue-comment-reopen-routes.test.ts
@@ -781,7 +781,7 @@ describe.sequential("issue comment reopen routes", () => {
     expect(mockIssueService.addComment).not.toHaveBeenCalled();
   });
 
-  it("does not move dependency-blocked issues to todo via POST comments", async () => {
+  it("does not move dependency-blocked issues to todo via POST comments and skips the assignee wake when the assignee is not mentioned (HUM-126 Fix 3)", async () => {
     mockIssueService.getById.mockResolvedValue(makeIssue("blocked"));
     mockIssueService.getDependencyReadiness.mockResolvedValue({
       issueId: "11111111-1111-4111-8111-111111111111",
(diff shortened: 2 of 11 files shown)


zionkim20:
Closing this on the author side — Zion (the HUM user this was scoped for) decided the hide-blocked filter is the wrong shape, and we're rescoping internally to a 'what can I unblock' prioritized view instead. The agent-wake loop fix (commit f5cbee3, item 1 of 3) is still genuinely useful upstream and we may re-open a smaller targeted PR for just that change later. For now, treat this as withdrawn. Sorry for the noise.

PR #6710, opened 2026-05-26
Release stale checkoutRunId on run termination + server shutdown

eel-ohgnas:
## Thinking Path

> - Paperclip orchestrates AI agents for zero-human companies, and the heartbeat run lifecycle owns the per-issue execution lock (`assigneeAgentId` + `checkoutRunId` + `executionRunId` + `executionLockedAt`) that makes the same-agent assumption safe.
> - Inside the heartbeat subsystem, the run-termination hook (`releaseIssueExecutionAndPromote`) is the single point where a finished run is supposed to hand its issue lock back to the system, and the checkout endpoint is the safety net that detects and recovers from a missed hand-back.
> - In practice the hook only clears `executionRunId`/`executionLockedAt` and leaves `checkoutRunId` pointing at the (now terminal) prior run. The next same-agent wake hits `POST /api/issues/:id/checkout` and is rejected with `409 Issue run ownership conflict` even though the prior run is unambiguously dead. There is no equivalent hook at all when the server itself is shutting down with in-flight runs, so a SIGTERM during a long run reliably wedges its issues.
> - The downstream effect is that an agent assigned to an issue effectively cannot recover from its own prior run ending without explicit `release` — every subsequent `checkout`/`release`/`comment` call 409s, which is a chain-reliability problem rather than a one-off incident.
> - This pull request closes both halves of the gap: the in-band run-termination hook now also clears `checkoutRunId`, the SIGTERM/SIGINT shutdown handler runs an equivalent terminate-and-release pass for in-flight runs, and the existing checkout safety net now emits a first-class `stale_lock_takeover` audit row so operators can see when the safety net is catching cases the in-band hook missed.
> - The benefit is that same-agent wakes resume cleanly after any run termination (graceful, timeout, SIGTERM, and via the safety net even SIGKILL/panic) without changing any agent SDK contract or cross-agent concurrency guarantees.

## What Changed

- **Root-cause fix** — `releaseIssueExecutionAndPromote` (`server/src/services/heartbeat.ts`) now also clears `checkoutRunId` when it equals the terminating run, so the in-band release hook actually releases the full lock instead of only the execution half.
- **Server-shutdown hook** — new `terminateInFlightRunsForShutdown(reason)` method on the heartbeat service iterates the `activeRunExecutions` set, marks each still-live run with `runs.status = 'terminated'` (a new terminal status, distinct from `cancelled`/`timed_out` so observers can tell apart "reaped on shutdown" from "agent or user stopped"), and clears every `checkoutRunId`/`executionRunId`/`executionLockedAt` it owns. The process-level SIGINT/SIGTERM handler in `server/src/index.ts` now calls this before stopping embedded Postgres and exiting.
- **Treat `terminated` as terminal** — `TERMINAL_HEARTBEAT_RUN_STATUSES` in `server/src/services/issues.ts` is now exported and includes `'terminated'`, so the existing `adoptStaleCheckoutRun` / `adoptUnownedCheckoutRun` safety nets recognise shutdown-reaped runs as adoptable.
- **`stale_lock_takeover` audit row** — both adopt helpers now write a structured `activity_log` row (`{ issueId, actorAgentId, actorRunId, priorRunId, priorRunStatus, trigger }`) whenever the checkout-time safety net adopts a stale lock. This gives operators a first-class signal for "how often is the safety net firing", which is the leading indicator that the in-band termination hook is failing in some new failure mode (SIGKILL, process panic, etc.).
- **Audit log doc** — `doc/spec/agent-runs.md` registers `stale_lock_takeover` under §13.2 alongside the existing `heartbeat.*` mutation actions, with the full payload schema and the operational interpretation.
- **Tests** — `server/src/__tests__/issue-stale-execution-lock-routes.test.ts` adds four cases covering the new behaviour: takeover writes the audit row (`checkout` trigger), cross-agent contention is still rejected and never adopts, same-agent + still-`running` prior run is still rejected with 409 and never adopts, and a `terminated`-status prior run is adopted cleanly through the existing checkout path.

## Verification

- `cd server && npx vitest run src/__tests__/issue-stale-execution-lock-routes.test.ts` — 7/7 passing (3 pre-existing + 4 new).
- `cd server && npx tsc --noEmit` — clean for every file touched by this PR. (There are pre-existing `unknown`/implicit-any errors elsewhere in `routes/plugins.ts` and `services/plugin-host-services.ts` that are unrelated to this change.)
- Manual repro of the original failure mode (see #issue link) is mechanically encoded by the new "writes a stale_lock_takeover audit row when same-agent checkout adopts a terminal-prior checkoutRunId" test: an issue with `checkoutRunId == priorRunId` and prior run status `failed` is reopened by the same agent on a fresh run and the lock now flips to the new run with an audit trail.

## Risks

- **Lower risk than it looks.** The in-band fix only adds `checkoutRunId: null` to an update that already targets the same row under the existing assignee+run guard, so it strictly widens what gets released on the success path and never touches a foreign run's lock.
- **`terminated` is a new run status string.** Anything that exhaustively switches on `heartbeat_runs.status` (UI badges, exports, plugin event handlers) will see this new value. The plugin lifecycle event mapper already returns `null` for unknown statuses (`publishRunLifecyclePluginEvent`), so no spurious events are emitted. UI handling is best-effort — worst case it renders as the raw string until callers add a friendly label.
- **Shutdown hook is best-effort.** Failures inside `terminateInFlightRunsForShutdown` are caught and logged; they never block process exit. If the database is already gone (embedded Postgres torn down before the handler ran), the hook degrades to a warning instead of blocking shutdown.
- **`stale_lock_takeover` audit row volume.** In healthy operation this should be ~0 rows. If volume goes non-zero in production, that is a leading indicator that the in-band hook isn't firing for some failure mode and warrants investigation rather than rate-limiting.
- **No migration required.** No DB schema changes; `heartbeat_runs.status` is already plain `text` and the new value is forward-compatible with the existing `TERMINAL_HEARTBEAT_RUN_STATUSES` consumers.

## Model Used

- Claude Opus 4.7 (`claude-opus-4-7`) running inside the Claude Code CLI v2 with the Paperclip plugin orchestrating heartbeat-driven execution. Extended-thinking mode enabled, full tool access (Read/Edit/Write/Bash/Grep/Glob) against the local clone; no MCP plugin tools were used to author the patch.

## Checklist

- [x] I have run tests locally and they pass (`vitest run src/__tests__/issue-stale-execution-lock-routes.test.ts`)
- [ ] If this change affects the UI, I have included before/after screenshots — n/a, server-only change
- [x] I have updated relevant documentation to reflect my changes (`doc/spec/agent-runs.md` §13.2)

### DIFF
--- a/doc/spec/agent-runs.md
+++ b/doc/spec/agent-runs.md
@@ -696,6 +696,7 @@ All wakeup/run state mutations must create `activity_log` entries:
 - `heartbeat.failed`
 - `heartbeat.cancelled`
 - `runtime_state.updated`
+- `stale_lock_takeover` — written when the checkout endpoint takes over an issue whose `checkoutRunId` (or `executionRunId`) points to a terminal-or-missing prior heartbeat run owned by the same agent. The `details` payload includes `issueId`, `actorAgentId`, `actorRunId`, `priorRunId`, `priorRunStatus`, and `trigger` (`"checkout"` or `"assert_checkout_owner"`). This is the safety-net counterpart to the server-shutdown run-termination hook; a sudden spike in `stale_lock_takeover` rows is a leading indicator that the termination hook is not firing for some failure mode (typically SIGKILL or process panic).
 
 ## 14. Heartbeat Service Implementation Plan
 
--- a/server/src/__tests__/issue-stale-execution-lock-routes.test.ts
+++ b/server/src/__tests__/issue-stale-execution-lock-routes.test.ts
@@ -283,4 +283,236 @@ describeEmbeddedPostgres("stale issue execution lock routes", () => {
       },
     });
   });
+
+  // ---- GAT-205 / stale checkout-run lock takeover (Option 1 safety net) ----
+
+  it("writes a stale_lock_takeover audit row when same-agent checkout adopts a terminal-prior checkoutRunId", async () => {
+    const { companyId, agentId, failedRunId, currentRunId } = await seedCompanyAgentAndRuns();
+    const issueId = randomUUID();
+    await db.insert(issues).values({
+      id: issueId,
+      companyId,
+      title: "Stale checkout lock takeover",
+      status: "in_progress",
+      priority: "high",
+      assigneeAgentId: agentId,
+      checkoutRunId: failedRunId,
+      executionRunId: failedRunId,
+      executionAgentNameKey: "codexcoder",
+      executionLockedAt: new Date(),
+    });
+
+    const res = await request(createApp(agentActor(companyId, agentId, currentRunId)))
+      .post(`/api/issues/${issueId}/checkout`)
+      .send({ agentId, expectedStatuses: ["in_progress"] });
+
+    expect(res.status, JSON.stringify(res.body)).toBe(200);
+    expect(res.body).toMatchObject({
+      id: issueId,
+      assigneeAgentId: agentId,
+      checkoutRunId: currentRunId,
+      executionRunId: currentRunId,
+    });
+
+    const audit = await db
+      .select({
+        action: activityLog.action,
+        actorType: activityLog.actorType,
+        actorId: activityLog.actorId,
+        agentId: activityLog.agentId,
+        runId: activityLog.runId,
+        details: activityLog.details,
+      })
+      .from(activityLog)
+      .where(eq(activityLog.action, "stale_lock_takeover"))
+      .then((rows) => rows[0]);
+    expect(audit).toMatchObject({
+      action: "stale_lock_takeover",
+      actorType: "agent",
+      actorId: agentId,
+      agentId,
+      runId: currentRunId,
+      details: {
+        issueId,
+        actorAgentId: agentId,
+        actorRunId: currentRunId,
+        priorRunId: failedRunId,
+        priorRunStatus: "failed",
+        trigger: "checkout",
+      },
+    });
+  });
+
+  it("does not adopt when the prior checkoutRunId belongs to a different agent (cross-agent contention stays 409)", async () => {
+    const { companyId, agentId, currentRunId } = await seedCompanyAgentAndRuns();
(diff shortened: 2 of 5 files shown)


eel-ohgnas:
Hi @eel-ohgnas — thanks for shipping this. Downstream signal from a Paperclip user (Gatabuta team) running a fork with a manual monkeypatch of this exact fix across two test cycles: GAT-205 stale-lock symptom is fully gone, `stale_lock_takeover` audit row fires correctly under same-agent reuse, and recovery is clean.

Empirical sample from Test 2 v1 (GAT-238, BE/FE 5-child DAG, 2026-05-30):

- Monkeypatch sentinel `grep -c 'stale_lock_takeover' services/issues.js` → 3 (in-band release + 2 adopt helpers, matches this PR's surface).
- One real stale-lock event during a cross-team BE→FE wake handoff; takeover adopted cleanly, the agent resumed on the new run with no manual `release`. No 409 loop.
- Two cycles, zero stuck-lock incidents that needed wipe; one separate wipe was from an unrelated `npx` auto-update overwriting the patch (orthogonal to this PR's correctness).

On the failing check `Verify serialized server suites (3/4)` — the failure is in `heartbeat-process-recovery.test.ts:1221` (`does not block paused-tree work when immediate continuation recovery is suppressed by the hold`), asserting `issue?.checkoutRunId === runId` but receiving `null`. That looks directly related to this PR rather than flake: the new in-band `releaseIssueExecutionAndPromote` clears `checkoutRunId` when it equals the terminating run, which probably fires in the paused-tree path where the test expects the prior `checkoutRunId` to be retained so the hold can resume it. Worth checking whether the release should be gated on `!issue.heldByPolicy` (or equivalent paused-tree predicate) before clearing `checkoutRunId`.

Happy to help repro or test a candidate fix on our side if useful — we already have the test scenario wired up end-to-end with real wakes.

PR #6711, opened 2026-05-26
fix(ui/import): seed CEO model into per-agent adapter overrides

Tr1ckyMag1ca1:
## Summary

Company Import seeded each imported agent's adapter override with the target company CEO's `adapterType` but never copied the CEO's `adapterConfig`. When the CEO ran `opencode_local`, importing failed unless the user manually expanded "configure adapter" on every imported agent and picked a model.

Failure message:

> Invalid opencode_local adapterConfig: OpenCode requires `adapterConfig.model` in provider/model format.

## Repro

1. Active company in sidebar has a CEO on `opencode_local` (with a valid model).
2. Open Company → Import, upload any exported `.zip` (e.g. a 14-agent newsletter company).
3. Run preview, accept defaults, click Import.
4. Import fails on the first agent — server rejects model-less `opencode_local` `adapterConfig`.

Workarounds today: expand "configure adapter" for **every** agent and pick a model, or switch every dropdown to a non-opencode adapter.

## Fix

`ui/src/pages/CompanyImport.tsx`:

- Read `ceo.adapterConfig.model` alongside `ceo.adapterType`.
- When seeding `adapterOverrides`, also seed `adapterConfigValues[slug]` with `{ ...defaultCreateValues, adapterType: ceoAdapterType, model: ceoAdapterModel }` so `buildFinalAdapterOverrides()` emits a complete `adapterConfig` (including `model`) for `opencode_local`.
- Only seeded when CEO has a model — preserves prior behavior for adapters that don't need a model and for the no-CEO case.

## Test plan

- [ ] With sidebar company on `opencode_local` + valid `model`: import a `.zip` company → succeeds without expanding any adapter row.
- [ ] With sidebar company on `claude_local`: import behaves as before.
- [ ] User manually changes one agent's adapter dropdown or expands "configure adapter" → user edits still win (default seed is overridden by `handleAdapterConfigChange`).
- [ ] Existing-company import with target CEO that already has `opencode_local` + model → no regression.

### DIFF
--- a/ui/src/pages/CompanyImport.tsx
+++ b/ui/src/pages/CompanyImport.tsx
@@ -702,6 +702,12 @@ export function CompanyImport() {
     const ceo = companyAgents.find((a) => a.role === "ceo");
     return ceo?.adapterType ?? "claude_local";
   }, [companyAgents]);
+  const ceoAdapterModel = useMemo(() => {
+    if (!companyAgents) return "";
+    const ceo = companyAgents.find((a) => a.role === "ceo");
+    const model = ceo?.adapterConfig?.model;
+    return typeof model === "string" ? model : "";
+  }, [companyAgents]);
 
   const localZipHelpText =
     "Upload a .zip exported directly from Paperclip. Re-zipped archives created by Finder, Explorer, or other zip tools may not import correctly.";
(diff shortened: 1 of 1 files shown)


Tr1ckyMag1ca1:
## Adjacent bug surfaced during repro: failed imports are not atomic

While reproducing the original error in production, I noticed each failed import attempt **still created the target company plus any agents that had been processed before the validator threw**. Cleanup of 4 failed attempts on the same `.zip` left behind:

| Company prefix | Agents persisted | Issues |
|---|---|---|
| `ART` | 1 | 0 |
| `ARTA` | 0 | 0 |
| `ARTAA` | 0 | 0 |
| `ARTAAA` | 5 | 0 |

The UI surfaced `Import failed — Invalid opencode_local adapterConfig: ...` each time, so from the user's perspective nothing happened. But the company row, `company_memberships`, `company_skills`, `issue_read_states` (65 rows for one of them), and partially-created agents were all committed.

Looks like `prepareImportedAgentAdapter`'s `assertImportAdapterConfigConstraints` throws mid-loop inside `applyImport` (`server/src/services/company-portability.ts` around 4288), but the surrounding work — `targetCompany` creation in `new_company` mode, plus any agents already created in earlier iterations — has already been committed outside a wrapping transaction.

Two suggested fixes (pick one, not both):

1. **Wrap the whole `applyImport` flow in a single DB transaction** so any validator throw rolls back the company + partial agents. Cleanest but biggest change — would also fix any future mid-import failure mode, not just this one.
2. **Pre-flight all adapter validation** before any DB writes: iterate `manifest.agents` once with `prepareImportedAgentAdapter` in a dry-run mode (resolve + validate, don't persist), and only start creating rows after every agent passes.

Option 2 is the smaller patch and matches how the preview step already runs separately from apply. Happy to spin this off into a separate PR if you'd prefer — flagging here since it was discovered in the same repro session as #6711.

PR #6719, opened 2026-05-26
fix(server): emit issue_blockers_resolved wake for blocked and backlog dependents

edenverov-hue:
## Thinking Path

> - The dependency / blocker model lets agents pause work until a blocker resolves
> - When a blocker transitions to `done`, a wake fan-out wakes the assignees of every dependent so they can resume
> - But the current `listWakeableBlockedDependents` filter excludes `backlog` status, so `backlog` dependents (and via a related path, `blocked` dependents) silently get no wake
> - In a real run this leaves agents stuck waiting on a dependency that's already resolved, until a manual nudge or the next periodic sweep
> - This PR removes `"backlog"` from the exclusion list and adds an audit event for every emitted wake, plus a deterministic dispatch matrix test that covers all 5 non-terminal dependent statuses and 3 negative cases
> - The benefit is that the wake-on-blocker-resolved signal becomes complete: every non-terminal dependent gets it, terminal dependents don't, and cancelled blockers don't count as resolved

## What Changed

- `server/src/services/issues.ts` — removed `"backlog"` from the terminal-status exclusion in `listWakeableBlockedDependents`. The filter now excludes only `done` and `cancelled`; `todo`/`in_progress`/`in_review`/`blocked`/`backlog` all pass through as wakeable.
- `server/src/routes/issues.ts` — added a `logActivity` entry (`issue.blockers_resolved_wake_emitted`) for each emitted wake. Fire-and-forget with `.catch(err => logger.warn(...))` so a logging failure can't block the wake itself.
- `server/src/__tests__/issues-service.test.ts` — added a dispatch matrix: 7 positive (5 non-terminal statuses × wake-then-verify + 2 terminal-dependent negatives) + 1 cancelled-blocker regression case.
- `server/src/__tests__/issue-dependency-wakeups-routes.test.ts` — mirrored the matrix at the route level using `supertest` + mocks (7 positive + 3 negative).

Net diff: 4 files, 306 insertions, 2 deletions.

## Verification

Local commands run (Windows + corepack-managed pnpm 9.15.4 + Node 24):

```sh
pnpm typecheck
# → exit 0; all 18 workspace packages including ui pass

pnpm exec vitest run --root . \
  server/src/__tests__/issue-dependency-wakeups-routes.test.ts
# → 69/69 pass (59 in issues-service + 10 in wakeups-routes)
```

Service-level tests use the project's embedded Postgres harness, so the dispatch matrix exercises real Drizzle queries against a real PG instance. Route-level tests use mocked `issueService` + `heartbeatService` and assert wake calls + audit-event emission.

Behavioral verification: the test "returns a backlog dependent once its blocker is done" asserts that `listWakeableBlockedDependents(blockerId)` returns the `backlog` dependent after the blocker transitions to `done`. Before the fix, this returned `[]`. After the fix, it returns the dependent.

## Risks

- **Behavioral expansion of wake fan-out.** Agents who were previously not woken (assignees of `backlog`/`blocked` dependents) will now be woken when their blocker resolves. For high-volume forks this increases adapter wake rate; for normal operation it's the intended behavior. No new heartbeat budget or rate-limit changes are needed because wake idempotency and rate-limits are already enforced upstream of this code path.
- **Activity-log row pressure.** Each emitted wake now adds one `activity_log` row (`issue.blockers_resolved_wake_emitted`). For a 100-agent company with 1k issues this is negligible (a few rows per day); worth noting for fork operators running at much larger scale.
- **No schema migration.** Zero risk of migration conflict.
- **No adapter changes, no UI changes.** Server-only fix; risk is contained to the wake-dispatcher emission path.

## Model Used

- Provider: Anthropic
- Model: Claude Opus 4.7 (claude-opus-4-7) with 1M context
- Mode: extended thinking enabled
- Tool use: bash + read + edit + grep tools
- This change was authored and tested by the model with human review of the final diff. The audit that surfaced the bug (and the test matrix design) was also model-produced.

## Checklist

- [x] I have checked ROADMAP.md and confirmed this PR does not duplicate planned core work — this is a bug fix in existing code, not a roadmap feature
- [x] I have run tests locally and they pass (69/69 for the touched files; full suite blocked by an unrelated Windows pnpm-spawnSync issue in `scripts/run-vitest-stable.mjs`)
- [x] I have added or updated tests where applicable (added a 10-case dispatch matrix at service + route layers)
- [ ] If this change affects the UI, I have included before/after screenshots — N/A, server-only fix
- [ ] I have updated relevant documentation to reflect my changes — no API or developer doc covers `listWakeableBlockedDependents` today; if upstream prefers a doc note, happy to add one

### DIFF
--- a/server/src/__tests__/issue-dependency-wakeups-routes.test.ts
+++ b/server/src/__tests__/issue-dependency-wakeups-routes.test.ts
@@ -1,6 +1,6 @@
 import express from "express";
 import request from "supertest";
-import { beforeEach, describe, expect, it, vi } from "vitest";
+import { beforeEach, describe, expect, it, vi, test } from "vitest";
 
 const mockWakeup = vi.hoisted(() => vi.fn(async () => undefined));
 const mockIssueService = vi.hoisted(() => ({
@@ -41,6 +41,9 @@ vi.mock("../services/index.js", () => ({
   heartbeatService: () => ({
     wakeup: mockWakeup,
     reportRunActivity: vi.fn(async () => undefined),
+    getRun: vi.fn(async () => null),
+    getActiveRunForAgent: vi.fn(async () => null),
+    cancelRun: vi.fn(async () => null),
   }),
   getIssueContinuationSummaryDocument: vi.fn(async () => null),
   instanceSettingsService: () => ({
@@ -164,6 +167,7 @@ describe("issue dependency wakeups in issue routes", () => {
       {
         id: "issue-2",
         assigneeAgentId: "agent-2",
+        status: "in_progress",
         blockerIssueIds: ["issue-1", "issue-3"],
       },
     ]);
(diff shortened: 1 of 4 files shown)


edenverov-hue:
Greptile feedback addressed in force-pushed amended commit (`64c3ceac` → `7849cdf1`).

**Issue 1 — `backlog` excluded in `getWakeableParentAfterChildCompletion`:**
The asymmetry is intentional and is now documented inline at `server/src/services/issues.ts:3936`. Rationale: the parent-after-child-completion path is an *auto-start* signal — pulling a parent out of backlog because its children finished crosses a policy line we don't want to cross silently. Backlog parents are expected to remain un-started until an agent or board explicitly picks them up. The blocked-dependents path is the opposite — those dependents already have active assignees waiting on the blocker, so resuming them is the desired automation. The two functions are deliberately not mirror images.

**Issue 2 — `dependentStatus` unused in route test:**
Refactored to actually exercise the parameter. Changes:
- `services/issues.ts`: `listWakeableBlockedDependents` now includes `status` on each returned dependent (was previously dropped in the final `.map`).
- `routes/issues.ts`: wake payload + contextSnapshot now carry `dependentStatus: dependent.status` so handlers know whether they were resumed from `todo` / `in_progress` / `in_review` / `blocked` / `backlog`.
- `issue-dependency-wakeups-routes.test.ts`: the parametrised mock now returns the dependent with `status: dependentStatus` and the wake-payload assertion verifies the status is forwarded. Each of the 5 iterations now actually tests a distinct shape end-to-end.

**Local verification:**
- `pnpm typecheck`: PASS across all packages
- `pnpm exec vitest run server/src/__tests__/issue-dependency-wakeups-routes.test.ts`: 10/10 PASS (5.7s)

Greptile re-review welcomed.

PR #6756, opened 2026-05-26
fix(server): delete cost_events + finance_events before heartbeat_runs in companies.remove()

hyamie:
## Thinking Path

> - Paperclip orchestrates AI agents for zero-human companies, with each agent generating heartbeat runs (per-wakeup execution records) and cost events (per-LLM-call accounting rows) on a continuous basis.
> - Company lifecycle is owned by `server/src/services/companies.ts`; the `remove()` function runs a single transaction that deletes the company's history across roughly a dozen related tables before deleting the company row itself.
> - The schema has `cost_events.heartbeat_run_id` and `finance_events.heartbeat_run_id` as foreign keys into `heartbeat_runs(id)`, so any deletion of `heartbeat_runs` must follow deletion of those two child tables.
> - In master, `remove()` deletes `heartbeatRuns` before `costEvents` and `financeEvents`, so any company with even a single heartbeat run + associated cost event fails its delete with Postgres error 23503 (`update or delete on table "heartbeat_runs" violates foreign key constraint`). For any heartbeat-on company this happens within minutes of first wakeup.
> - This pull request reorders the two `tx.delete()` calls so `costEvents` and `financeEvents` are removed before `heartbeatRuns`.
> - The benefit is that `DELETE /api/companies/<id>` becomes safe to call from the UI or API for any company with execution history, removing the need for the current workaround of pre-deleting child rows directly in Postgres.

## What Changed

- Moved `tx.delete(costEvents)` ahead of `tx.delete(heartbeatRuns)` in `companies.remove()` (`server/src/services/companies.ts`).
- Moved `tx.delete(financeEvents)` ahead of `tx.delete(heartbeatRuns)` in the same transaction.
- Added a one-line comment noting the FK dependency on `heartbeat_runs(id)` so the ordering is not innocently re-broken later.

## Verification

**Reproduction on master** (without this patch):
1. Run any company with `heartbeat.enabled = true` for a few minutes so the agents emit heartbeat runs + cost events.
2. `curl -X DELETE -H "Authorization: Bearer <admin>" http://<paperclip>/api/companies/<id>`
3. Server returns HTTP 500 + Postgres error 23503 (`update or delete on "heartbeat_runs" violates foreign key constraint "cost_events_heartbeat_run_id_fkey"` and/or the `finance_events` equivalent).

**Confirmation on this branch:**
1. Cherry-pick this commit onto master, restart the server.
2. Repeat step 2 above against the same kind of company (heartbeat history + cost events).
[...]

PR #6803, opened 2026-05-27
fix: delete cost_events before heartbeat_runs to respect FK constraint

chloyka:
## Summary

- Company deletion (`DELETE /api/companies/:id`) fails with a 500 error when `cost_events` rows reference `heartbeat_runs` via the `cost_events_heartbeat_run_id_heartbeat_runs_id_fk` foreign key
- The `remove()` method in `companyService` was deleting `heartbeat_runs` before `cost_events` and `finance_events`, violating the FK constraint (`ON DELETE no action`)
- Moved `cost_events` and `finance_events` deletion before `heartbeat_runs` deletion to satisfy dependency order

## Test plan

- [x] Reproduced: `DELETE /api/companies/:id` returns 500 on a company with cost_events
- [x] Applied fix: same request returns `{"ok": true}` and company is deleted
- [ ] `pnpm test` passes

### DIFF
--- a/server/src/services/companies.ts
+++ b/server/src/services/companies.ts
@@ -280,13 +280,13 @@ export function companyService(db: Db) {
         }
         await tx.delete(agentTaskSessions).where(eq(agentTaskSessions.companyId, id));
         await tx.delete(activityLog).where(eq(activityLog.companyId, id));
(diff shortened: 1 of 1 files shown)


chloyka:
Closing as duplicate of #6756 which addresses the same FK ordering issue.

PR #6822, opened 2026-05-28
Feat/process loss infra cause gno 182

eltonmonteirodebarros:
## Thinking Path

> - Paperclip orchestrates AI agents for zero-human companies; the agent runtime needs to distinguish failures it can fix from infrastructure failures it cannot.
> - Subsystem: agent runtime — process-loss handling and the reset-infrastructure recovery flow.
> - Problem: when an agent run vanishes (process_lost), the harness today retries blindly. If the cause is infrastructure (e.g. apiHealthy=false), retrying just burns budget and noise; if the cause is the agent itself, infra-level recovery is the wrong tool.
> - This PR splits the signal into infra-cause vs agent-cause and wires the existing reset-infrastructure path only to the infra-cause branch, with regression tests for agent-spoofable signals.
> - Benefit: cleaner observability of true infra incidents, fewer wasted retries, fail-closed semantics on the recovery endpoint.

## What Changed

- feat(process-loss): classify infra vs agent cause for process loss events (50840cae)
- feat(§5 GNO-128): implement `agents.status.reset_infrastructure` (721aba2c)
- test(GNO-183): add explicit PoC regression tests for agent-spoofable signals (1ace4c07)
- fix(GNO-212): fail-closed when reset-infrastructure target has no active issue (8eaf2d0e)
- fix(process-loss): invert signal #4 — fire on `apiHealthy === false` (GNO-216) (6d49da25)

## Verification

- New regression tests added in (1ace4c07) cover the agent-spoofable signal paths.
- Manual verification on a private GNOMON LABS Paperclip instance: synthetic infra-cause events route through reset_infrastructure; agent-cause events do not.
- No schema migrations required.

## Notes

This work originated from operational incident GNO-128 on a downstream GNOMON LABS Paperclip instance. Branch was developed at `feat/process-loss-infra-cause-gno-135` and renamed during push to `feat/process-loss-infra-cause-gno-182` to align with the lead issue. Happy to split into smaller PRs if maintainers prefer.


eltonmonteirodebarros:
Refocused and rebased onto current master since the draft. The TTL/tick-interval mismatch from the earlier Greptile review is fixed: makeApiHealthCache now uses ttlMs = 60000 (>= 2x the 30s scheduler interval), so isHealthy() no longer returns false on every pass and Signal #4 only fires on genuine concurrent API disruption. Rebased onto master with no conflicts and CI green. Note: the Greptile summary above predates the fix and has not re-run on the current HEAD.

PR #6841, opened 2026-05-28
fix(db): backfill company_secret_bindings for legacy webhook routine_triggers

Sergio-LPA:
## Thinking Path

> - Routines can be triggered by external webhooks, and each webhook trigger stores a shared secret used for HMAC verification when the dispatcher fires
> - At some point before this PR, `services/secrets.ts:assertBindingContext` started enforcing that every secret resolved with `consumerType="routine"` has a matching row in `company_secret_bindings`; new triggers minted by `services/routines.ts` create that row, but no migration backfilled rows for triggers minted by older releases
> - Result: after upgrading, every legacy webhook trigger fails with HTTP `422 binding_missing` on the next fire — the secret still exists but the binding does not, so dispatch is rejected even though the bearer is valid
> - This pull request adds an idempotent Drizzle backfill migration that inserts the missing binding for every webhook trigger that already references a secret, regardless of the trigger's enabled flag
> - The benefit is that any instance upgrading from a pre-enforcement release recovers its webhook triggers automatically at the next `db:migrate` run, with no manual SQL and no runtime fallback in `assertBindingContext`

No pre-existing GitHub issue. Inline description follows the bug template.

## What happened

Legacy webhook `routine_triggers` — those created before `assertBindingContext` began enforcing binding presence — return HTTP `422 binding_missing` on every dispatch after the instance upgrades. The trigger row still holds a valid `secret_id`, and the secret row still holds a resolvable value, but the dispatcher rejects the resolution because `company_secret_bindings` has no matching row for the `(company_id, secret_id, target_type='routine', target_id=routine_id::text, config_path='webhookSecret:'||secret_id::text)` key.

## Expected behavior

Legacy webhook triggers should keep firing after upgrade. The upgrade should reconcile the missing binding rows automatically, with no manual SQL and no operator intervention.

## Steps to reproduce

1. Bring up any Paperclip instance on a release that predates the binding-enforcement change (any snapshot where `services/routines.ts` still writes only `company_secrets` and never inserts into `company_secret_bindings` for new webhook triggers).
2. Create a webhook trigger for a routine. Confirm it fires end-to-end (HTTP 200).
3. Upgrade to the release that enforces bindings in `services/secrets.ts:assertBindingContext`.
4. Fire the webhook trigger again → `POST /api/routine-triggers/public/<publicId>/fire` returns `HTTP 422 {"code":"binding_missing"}`.

## Paperclip version

Reproducible on any master snapshot after the `assertBindingContext` enforcement change (`server/src/services/secrets.ts` throws `binding_missing` when `bindingContext` is missing or unmatched) and before this migration is applied. Confirmed against master @ `3ff636bc4`.

## Deployment mode

Self-hosted Docker (Postgres via `paperclip-db`). The bug is deployment-mode independent — it only depends on whether the instance carries webhook triggers minted before the binding-enforcement release.

## What Changed

- Added `packages/db/src/migrations/0227_backfill_webhook_secret_bindings.sql` — backfills `company_secret_bindings` rows for legacy webhook `routine_triggers` (enabled and disabled) that have a `secret_id` but no matching binding.
- Registered the new migration in `packages/db/src/migrations/meta/_journal.json` (idx 227).
- Added `packages/db/src/backfill-webhook-secret-bindings-migration.test.ts` — end-to-end embedded-Postgres test that seeds legacy triggers (enabled + disabled + already-bound), replays the migration, asserts the two missing bindings are created, and asserts idempotency (a second replay inserts zero rows).

The insert mirrors exactly the shape that `server/src/services/routines.ts` writes for new triggers (`target_type='routine'`, `target_id=routine_id::text`, `config_path='webhookSecret:'||secret_id::text`).

## Verification

Manual instance check before the fix (legacy trigger, no binding):

```
POST /api/routine-triggers/public/<publicId>/fire   →  HTTP 422 {"code":"binding_missing"}
```

After running the backfill on the same instance:

```
POST /api/routine-triggers/public/<publicId>/fire   →  HTTP 200 (dispatchRoutineRun)
```

Repeated runs against an already-mitigated instance return `INSERT 0 0` — the migration is safe to apply repeatedly.

Automated: `packages/db/src/backfill-webhook-secret-bindings-migration.test.ts` seeds three legacy triggers (one enabled, one disabled, one already bound) in an embedded Postgres, runs `applyPendingMigrations`, and asserts:

- Both the enabled and disabled legacy triggers get their bindings backfilled (proves the enabled-filter drop covers disabled triggers).
- The already-bound trigger is left untouched (proves `NOT EXISTS` + `ON CONFLICT DO NOTHING` idempotency, no duplicate rows, no unique-violation).
- A second replay of the migration inserts zero additional rows (proves full idempotency across re-runs).

## Risks

- **Low risk**. Pure data backfill; no DDL, no schema change.
- Idempotent via `NOT EXISTS` + `ON CONFLICT ("company_id","target_type","target_id","config_path") DO NOTHING`. Re-runs insert zero rows.
- Covers all webhook triggers with a non-null `secret_id`, including disabled ones. Covering disabled triggers avoids a latent regression: `services/routines.ts` writes the binding only at trigger *creation*, never on toggle, so re-enabling a legacy disabled trigger after the migration would otherwise still hit 422 `binding_missing`.
- Foreign-key safety: every `routine_triggers.secret_id` already references `company_secrets(id)`, so the FK on `company_secret_bindings.secret_id` is guaranteed to resolve for every inserted row.
- Migration runtime is proportional to the number of webhook triggers per instance — a single insert with a `NOT EXISTS` subquery on indexed columns, expected to be well under a second on any reasonable instance size.

## Model Used

- Provider/model: Anthropic Claude — `claude-opus-4-7` (Opus 4.7).
- Context window: 200K tokens.
- Capabilities used: tool use (PostgreSQL via psql over SSH for the original manual verification, GitHub Contents API for cross-file schema reads on master, internal codebase reads of `services/secrets.ts`, `services/routines.ts`, the existing migration set, and reference migration tests).
- No extended thinking mode.

## Checklist

- [x] I have run tests locally and they pass — the new embedded-Postgres migration test seeds legacy triggers, replays the migration, and verifies both backfill and idempotency
- [x] I have added or updated tests where applicable — `packages/db/src/backfill-webhook-secret-bindings-migration.test.ts`
- [x] If this change affects the UI, I have included before/after screenshots — N/A
- [x] I have updated relevant documentation to reflect my changes — the migration body includes a long-form comment explaining the root cause for any operator reading the migration history

### DIFF
--- a/packages/db/src/backfill-webhook-secret-bindings-migration.test.ts
+++ b/packages/db/src/backfill-webhook-secret-bindings-migration.test.ts
@@ -1,0 +1,172 @@
+import { createHash, randomUUID } from "node:crypto";
+import fs from "node:fs";
+import { afterEach, describe, expect, it } from "vitest";
+import postgres from "postgres";
+import { applyPendingMigrations } from "./client.js";
+import {
+  getEmbeddedPostgresTestSupport,
+  startEmbeddedPostgresTestDatabase,
+} from "./test-embedded-postgres.js";
+
+const MIGRATION_FILE = "0227_backfill_webhook_secret_bindings.sql";
+
+const cleanups: Array<() => Promise<void>> = [];
+const embeddedPostgresSupport = await getEmbeddedPostgresTestSupport();
+const describeEmbeddedPostgres = embeddedPostgresSupport.supported ? describe : describe.skip;
+
+async function migrationHash(): Promise<string> {
+  const content = await fs.promises.readFile(
+    new URL(`./migrations/${MIGRATION_FILE}`, import.meta.url),
+    "utf8",
+  );
+  return createHash("sha256").update(content).digest("hex");
+}
+
+if (!embeddedPostgresSupport.supported) {
+  console.warn(
+    `Skipping embedded Postgres backfill webhook secret bindings migration tests on this host: ${embeddedPostgresSupport.reason ?? "unsupported environment"}`,
+  );
+}
+
+describeEmbeddedPostgres("backfill webhook secret bindings migration", () => {
+  afterEach(async () => Promise.all(cleanups.splice(0).map((cleanup) => cleanup())));
+
+  it(
+    "creates missing bindings for legacy webhook triggers (enabled and disabled) and is idempotent",
+    async () => {
+      const database = await startEmbeddedPostgresTestDatabase(
+        "paperclip-backfill-webhook-bindings-",
+      );
+      cleanups.push(database.cleanup);
+      const sql = postgres(database.connectionString, { max: 1, onnotice: () => {} });
+      cleanups.push(async () => sql.end());
+
+      const companyId = randomUUID();
+      // Each webhook trigger lives on its own routine, mirroring how
+      // services/routines.ts writes bindings with target_id=routine_id.
+      const enabledRoutineId = randomUUID();
+      const disabledRoutineId = randomUUID();
+      const alreadyBoundRoutineId = randomUUID();
+      const enabledSecretId = randomUUID();
+      const disabledSecretId = randomUUID();
+      const alreadyBoundSecretId = randomUUID();
+      const enabledTriggerId = randomUUID();
+      const disabledTriggerId = randomUUID();
+      const alreadyBoundTriggerId = randomUUID();
+
+      await sql`
+        INSERT INTO "companies" ("id", "name", "issue_prefix")
+        VALUES (${companyId}, 'Backfill webhook company', 'BWH')
+      `;
+      for (const routineId of [enabledRoutineId, disabledRoutineId, alreadyBoundRoutineId]) {
+        await sql`
+          INSERT INTO "routines" (
+            "id", "company_id", "title"
+          ) VALUES (
+            ${routineId}, ${companyId}, ${`Legacy routine ${routineId}`}
+          )
+        `;
+      }
+      for (const secretId of [enabledSecretId, disabledSecretId, alreadyBoundSecretId]) {
+        await sql`
+          INSERT INTO "company_secrets" (
+            "id", "company_id", "key", "name"
+          ) VALUES (
+            ${secretId}, ${companyId}, ${`webhookSecret:${secretId}`}, ${`webhook-${secretId}`}
+          )
+        `;
+      }
+      await sql`
+        INSERT INTO "routine_triggers" (
+          "id", "company_id", "routine_id", "kind", "enabled", "secret_id"
+        ) VALUES (
+          ${enabledTriggerId}, ${companyId}, ${enabledRoutineId}, 'webhook', true, ${enabledSecretId}
+        )
+      `;
+      await sql`
+        INSERT INTO "routine_triggers" (
+          "id", "company_id", "routine_id", "kind", "enabled", "secret_id"
+        ) VALUES (
+          ${disabledTriggerId}, ${companyId}, ${disabledRoutineId}, 'webhook', false, ${disabledSecretId}
+        )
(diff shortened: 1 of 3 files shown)


Sergio-LPA:
Rebased onto current `master` and addressed both Greptile findings. Summary of changes on the latest push (`d550c8d6`):

**Rebase**
- Master advanced from idx 92 → idx 226 (135 new migrations), so the file was renumbered `0092_backfill_webhook_secret_bindings.sql` → `0227_backfill_webhook_secret_bindings.sql` and the journal entry re-indexed. No other file conflicted. Verified the fix is still needed on current master (`assertBindingContext` in `server/src/services/secrets.ts` still throws `binding_missing`; `services/routines.ts` still writes the binding only at trigger creation; no other migration between 0092 and 0226 backfills it).
- Confirmed the `company_secret_bindings` schema gained additional columns since May (`version_selector`, `required`, `label`, `projection_class`, `projection_allowlist_key`) — all with defaults — and that `services/routines.ts` still inserts only the same 5 core columns for new triggers, so the migration's `INSERT ... (company_id, secret_id, target_type, target_id, config_path)` remains byte-for-byte identical to the shape the application writes.

**Greptile P1 — dropped the `enabled = true` filter**
Legacy disabled webhook triggers are now backfilled too. Rationale as flagged in the review: `services/routines.ts` writes the binding only at trigger creation, never on enable/disable, so re-enabling a legacy disabled trigger after the migration would otherwise still hit `422 binding_missing`. Covering disabled triggers now closes that latent regression.

**Greptile P2 — added `b.company_id = t.company_id` to the NOT EXISTS predicate**
Applied the suggested block. Harmless in practice (`secret_id` is company-scoped via the `company_secrets(id)` FK) but the predicate now mirrors the shape of the `(company_id, target_type, target_id, config_path)` unique index, and makes the intent self-documenting.

**Test coverage added**
`packages/db/src/backfill-webhook-secret-bindings-migration.test.ts` — embedded-Postgres integration test that:
1. Seeds three legacy webhook triggers (one enabled, one disabled, one already-bound), each on its own routine to mirror how `services/routines.ts` writes bindings with `target_id=routine_id`.
2. Deletes the migration hash and re-runs `applyPendingMigrations`.
3. Asserts both the enabled and disabled triggers get their bindings backfilled (proves the enabled-filter drop covers disabled triggers) and the already-bound trigger is left untouched (proves `NOT EXISTS` + `ON CONFLICT DO NOTHING` — no duplicates, no unique-violation).
4. Deletes the hash a second time and re-runs, asserting zero additional rows (full idempotency across re-runs).

**CI status on `d550c8d6`**
25/28 checks green, including all `packages/db` general tests (workspaces-b), all 5 server suites, all 5 verify-serialized shards, e2e (3/3), Canary Dry Run, Build, Typecheck, Greptile Review, policy, commitperclip review, and every security scan (Snyk, Socket, Superagent). One workspace shard (`General tests (workspaces-a (2/2))`) failed on `packages/ui/src/components/FileViewerSheet.copy.test.tsx > copies the current file view link and shows confirmation` — a clipboard-toast UI test in code this PR does not touch (diff is limited to `packages/db/src/migrations/0227_...sql`, `packages/db/src/migrations/meta/_journal.json`, and the new `packages/db/src/backfill-webhook-secret-bindings-migration.test.ts`). The same test is green on the five most recent other open PRs on the queue, so this looks like an ephemeral shard-load flake on this specific run rather than a regression from this change; a shard rerun should clear it.

Ready for another review pass whenever convenient.

PR #6853, opened 2026-05-28
fix(recovery): skip routine-driven standing threads in stranded reconciliation

TylerBarnes:
## Thinking Path

> - The recovery subsystem watches assigned issues that lose their live execution path and reassigns/escalates them so work does not get lost
> - A standing thread that is the \`parentIssueId\` of an active routine has a different live continuation path — the routine itself fires execution issues against it on a schedule — but \`reconcileStrandedAssignedIssues\` did not know about routines
> - On every scan the standing thread looked stranded, recovery requeued a continuation, and the next scan escalated the issue to \`blocked\` with empty \`blockedByIssueIds\` and reassigned to the chain-of-command — a loop, and a \`blocked\` status that violates the platform's own \"blocked implies first-class blockers\" invariant
> - This PR teaches the stranded-assigned reconciler about active-routine standing threads and skips them
> - The benefit is that daily/scheduled delivery threads no longer bounce under recovery, and we stop emitting \`blocked\`-with-no-blockers states from this path

## What Changed

- Add \`isActiveRoutineStandingThread(companyId, issueId)\` in \`server/src/services/recovery/service.ts\` that returns true when the issue is the \`parentIssueId\` of an \`active\` routine.
- In \`reconcileStrandedAssignedIssues\`, skip candidates that are active-routine standing threads. Placed right after \`hasActiveExecutionPath\` so the routine acts as a peer \"live continuation path\" signal.
- Add a vitest case (\`heartbeat-process-recovery.test.ts\`) that seeds the exact loop condition (in_progress + succeeded productive-terminal continuation retry) and asserts the issue is not escalated, no recovery action is created, no comment is posted, and no \`source_scoped_recovery_action\` wake is enqueued when a routine targets the issue.
- Extend the test cleanup to delete \`routines\` before \`agents\` (\`routines.assignee_agent_id\` has no cascade).

## Verification

\`\`\`
pnpm --filter @paperclipai/server exec vitest run src/__tests__/heartbeat-process-recovery.test.ts --reporter=dot
\`\`\`

- All 45 tests in that suite pass.
- The new test fails (escalation = 1) without the recovery-service change applied.

Manual repro (before this PR): a routine-driven standing thread sitting at \`in_progress\` with a productive-terminal \`succeeded\` continuation run gets flipped to \`blocked\` (with empty \`blockedByIssueIds\`) and reassigned to the chain-of-command on the very next recovery sweep, then again on each subsequent sweep — see TES-17 / TES-18 / TES-19.

## Risks

- Low risk. The new skip is gated on \`routines.status = 'active'\` and \`routines.parent_issue_id = issue.id\`. Issues that are not the standing target of an active routine remain on the existing recovery path unchanged, as confirmed by the unchanged behavior of the neighboring \"blocks stranded in-progress work after a productive continuation retry was already used\" test.
- One small follow-up worth tracking separately: \`escalateStrandedAssignedIssue\` will still write \`status: \"blocked\"\` with an empty \`blockedByIssueIds\` array if it is ever invoked without unresolved blockers. The general invariant \"blocked implies first-class blockers\" deserves a dedicated PR.

### DIFF
--- a/server/src/__tests__/heartbeat-process-recovery.test.ts
+++ b/server/src/__tests__/heartbeat-process-recovery.test.ts
@@ -26,6 +26,7 @@ import {
   issueTreeHolds,
   issueWorkProducts,
   issues,
+  routines,
 } from "@paperclipai/db";
 import {
   getEmbeddedPostgresTestSupport,
@@ -332,6 +333,7 @@ describeEmbeddedPostgres("heartbeat orphaned process recovery", () => {
     await db.delete(issueRecoveryActions);
     await db.delete(issueTreeHoldMembers);
     await db.delete(issueTreeHolds);
+    await db.delete(routines);
     for (let attempt = 0; attempt < 5; attempt += 1) {
       await db.delete(issueComments);
       await db.delete(issueDocuments);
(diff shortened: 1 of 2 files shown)


TylerBarnes:
This is kinda shocking and bizarre, I tried paperclip for a sec yesterday and had no idea it was doing this. That's kinda fucked up

PR #6862, opened 2026-05-28
fix(claude-local): pass --setting-sources=user so ~/.claude/settings.json hooks load

### LINKED ISSUE #6483: Adapter should pass --setting-sources=user so ~/.claude/settings.json hooks load under claude_local
﻿## Summary

The `@paperclipai/adapter-claude-local` adapter (`dist/server/execute.js`, `buildClaudeArgs`) spawns the `claude` CLI without `--setting-sources=user`. This means Claude Code only loads project-scoped settings and silently ignores `~/.claude/settings.json` for the entire heartbeat session.

Any user-level PreToolUse / SessionStart / etc. hook configured in `~/.claude/settings.json` becomes inert under the adapter — including governance hooks operators install to gate destructive commands (`git push origin main`, `gh pr merge`, etc.). The failure is silent: hooks don't error, they just never run.

## Repro

Repository: any directory with no `.claude/settings.json` of its own.

1. Add a PreToolUse hook to `~/.claude/settings.json` (e.g. one that echoes to stderr on `Bash` matcher).
2. Launch the agent via the standard paperclipai claude_local adapter (`paperclipai run`).
3. In the resulting heartbeat, run any matching Bash command.
4. Expected: the hook fires and logs.
5. Actual: the hook is silently skipped because `--setting-sources=user` was not passed, so the CLI doesn't load user-scope settings.

Confirmed under `paperclipai` and `@paperclipai/adapter-claude-local` `2026.517.0` on Windows 11.

## Proposed fix

In `packages/adapters/claude-local/src/server/execute.ts` (built artifact: `dist/server/execute.js`, around the `buildClaudeArgs` lambda), add `--setting-sources=user` to the default args array:

```diff
- const args = ["--print", "-", "--output-format", "stream-json", "--verbose"];
+ const args = ["--print", "-", "--output-format", "stream-json", "--verbose", "--setting-sources=user"];
```

If there is a reason to omit `user` sources by default (e.g. sandbox isolation), an explicit adapter config flag (`loadUserSettings: true | false`) with `true` default would let operators opt out without making the failure silent.

## Why this matters / context

We operate under a governance regime (cf. our internal IMA-82) where a PreToolUse merge-guard hook in `~/.claude/settings.json` is the last line of defense against unintended merges by an automated bot. Without `--setting-sources=user`, that hook is silently disabled for every heartbeat — and we only discovered the gap after an incident (IMA-86 smoke landed a benign push that the hook should have surfaced).

Our current workaround is a local script that re-applies the flag to `dist/server/execute.js` in the npx cache and a Paperclip routine that fires daily to repair any drift, but the underlying gap is in the adapter. We would happily land a PR if a maintainer can confirm the preferred shape (always-on flag vs. opt-out config).

## Environment

- `paperclipai` 2026.517.0
- `@paperclipai/adapter-claude-local` 2026.517.0
- Claude Code CLI (latest as of 2026-05-20)
- Windows 11

isak-ialogics:
## Summary

- Adds `--setting-sources=user` flag when spawning Claude CLI in the `claude_local` adapter so that `~/.claude/settings.json` (user-level hooks and settings) is always loaded for agent heartbeats
- Adds `settingSources` adapter config option (defaults to `"user"`) so operators can customize which setting sources are loaded
- Adds `settingSources?: string` to `CreateConfigValues` interface in `adapter-utils` for type safety

## Problem

Operators who configure hooks or settings at the user level (`~/.claude/settings.json`) saw them silently ignored because the adapter spawned Claude Code without `--setting-sources`.

## Test plan
- [ ] Run a claude_local agent heartbeat and confirm `~/.claude/settings.json` hooks execute
- [ ] Verify `--setting-sources=user` appears in the Claude CLI invocation
- [ ] Set `settingSources: "user,project,local"` in adapter config and confirm all three sources load

Closes https://github.com/paperclipai/paperclip/issues/6483


isak-ialogics:
Closing: duplicate of #7112 fix/claude-local-setting-sources-clean (same fix: pass --setting-sources=user). Keeping #7112.

PR #6865, opened 2026-05-28
fix(checkout): return 409 instead of 500 on routine_execution concurrent execution constraint

### LINKED ISSUE #6519: fix: checkout 500 on routine_execution issues when concurrent execution exists
## Bug

`POST /api/issues/{id}/checkout` returns an unhandled 500 error when a routine_execution issue is checked out while another open issue with the same `(companyId, originKind, originId, originFingerprint)` key already has a non-null `executionRunId`.

## Root Cause

The checkout service sets `executionRunId = checkoutRunId` via a DB UPDATE. If another routine_execution issue with the same unique key already has `executionRunId` set and is in an open status, the DB constraint `issues_open_routine_execution_uq` fires — a PostgreSQL 23505 (unique violation). This exception is not caught in the checkout service and bubbles up as a 500.

The routine spawn code in `routines.ts` already handles this constraint correctly (converts it to a skip/coalesce result), but `issues.ts` checkout does not.

## Impact

- Routine monitor runs fail to checkout their assigned execution issues
- concurrencyPolicy:skip_if_active cannot clear the stale lock (checkout is the only path)
- Stale routine_execution issues block all future routine spawns indefinitely

## Fix

Wrap the `executionRunId` UPDATE statements in `services/issues.ts checkout()` in a try/catch that detects the `issues_open_routine_execution_uq` violation and throws a proper 409 conflict error instead of propagating the raw DB error.

## Locations

- `server/src/services/issues.ts` — `checkout()` function at ~line 4629 and ~line 4677 (two UPDATE statements that set `executionRunId`)

isak-ialogics:
## Problem

`POST /api/issues/{id}/checkout` returned 500 when a `routine_execution` issue was checked out while another open issue with the same `(companyId, originKind, originId, originFingerprint)` key already had a non-null `executionRunId`.

The two `UPDATE` statements in `checkout()` that set `executionRunId = checkoutRunId` were not catching the `issues_open_routine_execution_uq` PostgreSQL unique constraint violation (23505). The raw DB error propagated to the Express error handler and became a 500.

Fixes #6519

## Root cause

The `issues_open_routine_execution_uq` constraint enforces that only one open `routine_execution` issue can hold a given `(companyId, originKind, originId, originFingerprint)` key with a non-null `executionRunId`. The routine spawn code in `routines.ts` already handles this correctly by catching 23505 on that constraint — but `checkout()` did not.

## Fix

Add `isOpenRoutineExecutionConflict()` helper (mirroring the existing check in `routines.ts`) and catch the constraint violation in both `UPDATE` paths inside `checkout()`:

1. The primary checkout update (sets `assigneeAgentId`, `checkoutRunId`, `executionRunId`, `status`)
2. The null-`checkoutRunId` adoption update

Both now throw a structured `conflict()` (409) instead of propagating the raw DB error.

## Test

- `POST /api/issues/:id/checkout` on a `routine_execution` issue returns **409** (not 500) when a concurrent execution holds the unique key
- Response body: `{ "error": "Concurrent routine_execution issue already has an active execution run", "details": { "constraint": "issues_open_routine_execution_uq" } }`
- All other checkout paths unaffected


isak-ialogics:
Closing: this branch conflicts with current upstream changes due to significant upstream divergence over the past 3+ weeks. The fix commit cherry-pick fails on current upstream/master.

If this fix is still needed, it can be re-submitted as a fresh clean cherry-pick from upstream/master.

PR #6866, opened 2026-05-28
fix(plugin-ui-static): catch DrizzleQueryError wrapping pg 22P02 on non-UUID :pluginId

### LINKED ISSUE #6295: plugin-ui-static: 500 instead of fallback when plugin_key passed as :pluginId (DrizzleQueryError wrapping hides pg code 22P02)
## Summary

`server/src/routes/plugin-ui-static.ts` returns **HTTP 500** for any request to `/_plugins/:pluginId/ui/*` where `:pluginId` is the plugin's `plugin_key` (e.g. `agent-pixels.camera`) rather than its UUID `id`. The fallback to `registry.getByKey()` is never reached.

## Repro

1. Install a plugin (e.g. `@agent-pixels/paperclip-plugin`).
2. The plugin's UI bundle requests assets via a path containing the plugin key:
   `GET /_plugins/agent-pixels.camera/ui/assets/agent-pixels-assets.json`
3. Response: `500 {"error":"Internal server error"}`.
4. Server log:
   ```
   ERROR: GET /_plugins/agent-pixels.camera/ui/assets/agent-pixels-assets.json 500
     err.type:    DrizzleQueryError
     err.message: invalid input syntax for type uuid: "agent-pixels.camera"
     caused by:   PostgresError (code 22P02)
   ```

## Root cause

[`plugin-ui-static.ts:244-256`](https://github.com/paperclipai/paperclip/blob/main/server/src/routes/plugin-ui-static.ts#L244-L256):

```ts
try {
  plugin = await registry.getById(pluginId);
} catch (error) {
  const maybeCode =
    typeof error === "object" && error !== null && "code" in error
      ? (error as { code?: unknown }).code
      : undefined;
  if (maybeCode !== "22P02") {
    throw error;
  }
}
```

Drizzle wraps the underlying `PostgresError` in a `DrizzleQueryError`. The `code` property lives on `error.cause.code`, not on `error.code` directly, so `maybeCode` is always `undefined`, the predicate `!== "22P02"` is true, and the error is rethrown. The graceful fallback to `registry.getByKey(pluginId)` at L258 is unreachable for the exact case it was written to handle.

## Suggested fix

Walk the cause chain when extracting the pg code:

```ts
const extractCode = (e: unknown): unknown => {
  if (typeof e !== "object" || e === null) return undefined;
  if ("code" in e && (e as { code?: unknown }).code !== undefined) {
    return (e as { code?: unknown }).code;
  }
  if ("cause" in e) return extractCode((e as { cause?: unknown }).cause);
  return undefined;
};
const maybeCode = extractCode(error);
if (maybeCode !== "22P02") {
  throw error;
}
```

With that change the route correctly falls through to `getByKey` and returns assets with HTTP 200.

## Environment

- Paperclip server: current `main` (verified on commit `18460f81`)
- drizzle-orm `0.45.2`, postgres `3.4.8`
- Node 25.9.0, macOS

Happy to send a PR if useful.

isak-ialogics:
## Problem

When a `plugin_key` string is passed as `:pluginId` in `GET /_plugins/:pluginId/ui/*`, Drizzle throws 22P02 (invalid_text_representation) uncaught → 500 instead of 404.

## Root cause

The catch block checked `error.code` directly. `DrizzleQueryError` (drizzle-orm 0.45+) wraps the pg `DatabaseError` on `.cause`, so `error.code` is always `undefined` and the condition never fires.

## Fix

Check both `error.code` and `error.cause.code`. Non-UUID `:pluginId` values now fall through to `getByKey` and return 404 when not found.

Fixes #6295


isak-ialogics:
Duplicate of #6877

PR #6872, opened 2026-05-28
fix(wake-on-comment): validate agent:// mention IDs against agents table

### LINKED ISSUE #6389: Wake-on-comment: validate agent:// mention IDs against agents table before waking
## Summary

`findMentionedAgents` in `server/src/services/issues.ts` resolves explicit
`agent://<uuid>` mention IDs from comment/issue bodies and returns them as
wake targets **without verifying** the UUIDs actually exist in the `agents`
table for the issue's company. Hallucinated UUIDs leak straight into
`enqueueWakeup` and produce orphaned wake attempts.

Filed by Wachter Intelligence (Marcel Wachter, paperclip company id
`5140e458-f9ce-444f-b516-701f85d08307`) — we saw an agent post
``[@Felix Strauss](agent://2803ddcf-95e5-4c7a-aa2b-94cf06aa5a83)`` (wrong
name — Hartmann is correct — and a fabricated UUID; our real Felix is
`2803ddcf-888b-4cbf-8dad-de05ece364f5`). The wake-on-comment path
`server/src/routes/issues.ts:1877` / `:2493` then tried to wake the
non-existent agent.

## Reproduction

1. Post an issue comment with body containing
   `[@Anyone](agent://00000000-0000-0000-0000-000000000000)`.
2. Server resolves the UUID via `extractAgentMentionIds(body)` and
   adds it to `resolved` unconditionally.
3. `findMentionedAgents` returns the hallucinated UUID.
4. The caller (wake-on-comment route) tries to enqueue a wakeup for an
   agent that doesn't exist or belongs to a different company.

Code path (server v0.3.1, file as installed):

```
server/src/services/issues.ts:2377-2398   findMentionedAgents
server/src/routes/issues.ts:1877          mentionedIds = await svc.findMentionedAgents(...)
server/src/routes/issues.ts:2493          (issue update path)
```

## Proposed fix

Filter `resolved` against the `agents` query result so only IDs that
**belong to the same company** survive. The query already loads all
agents for the company; we just have to use the id set.

```ts
// server/src/services/issues.ts — findMentionedAgents

findMentionedAgents: async (companyId: string, body: string) => {
  const re = /\B@([^\s@,!?.]+)/g;
  const tokens = new Set<string>();
  let m: RegExpExecArray | null;
  while ((m = re.exec(body)) !== null) {
    const normalized = normalizeAgentMentionToken(m[1]);
    if (normalized) tokens.add(normalized.toLowerCase());
  }

  const explicitAgentMentionIds = extractAgentMentionIds(body);
  if (tokens.size === 0 && explicitAgentMentionIds.length === 0) return [];

  const rows = await db.select({ id: agents.id, name: agents.name })
    .from(agents).where(eq(agents.companyId, companyId));
  const validIds = new Set(rows.map((r) => r.id));

  const resolved = new Set<string>();
  // Only keep explicit mention IDs that actually belong to this company.
  for (const id of explicitAgentMentionIds) {
    if (validIds.has(id)) {
      resolved.add(id);
    } else {
      logger.warn(
        { companyId, hallucinatedAgentId: id },
        "findMentionedAgents: ignoring agent:// mention with unknown id",
      );
    }
  }
  for (const agent of rows) {
    if (tokens.has(agent.name.toLowerCase())) {
      resolved.add(agent.id);
    }
  }
  return [...resolved];
},
```

(Logger import path matches the rest of the file — adjust as needed if
the file uses a different logger.)

## Test plan

* Unit: post a comment with a known-bad `agent://` UUID and assert
  the returned list does not include it.
* Unit: post a comment with one valid and one hallucinated UUID;
  assert only the valid one comes back.
* Manual: confirm the warn log shows up when a hallucinated id is
  filtered.

Happy to send a PR if useful — wanted to file the issue first so the
fix is on the maintainers' radar and a CTO-routed Wachter agent can
shepherd a PR if Marcel asks.

isak-ialogics:
## Summary

Fixes #6389

Before dispatching wakes for explicit `agent://UUID` mentions in comments, filter the extracted IDs against the agents DB table. Invalid or non-existent agent IDs are silently skipped; valid ones continue to trigger wakes as before.

## Changes

- `server/src/services/issues.ts`: build `validAgentIds` set from DB rows before filtering `explicitAgentMentionIds`

## Test plan

- [ ] Comment mentioning a valid `agent://UUID` — wake fires normally
- [ ] Comment mentioning an invalid/deleted `agent://UUID` — no wake, no error

### DIFF
--- a/server/src/services/issues.ts
+++ b/server/src/services/issues.ts
@@ -5350,7 +5350,10 @@ export function issueService(db: Db) {
       if (tokens.size === 0 && explicitAgentMentionIds.length === 0) return [];
       const rows = await db.select({ id: agents.id, name: agents.name })
         .from(agents).where(eq(agents.companyId, companyId));
-      const resolved = new Set<string>(explicitAgentMentionIds);
+      const validAgentIds = new Set(rows.map((r) => r.id));
+      const resolved = new Set<string>(
+        explicitAgentMentionIds.filter((id) => validAgentIds.has(id)),
+      );
       for (const agent of rows) {
         if (tokens.has(agent.name.toLowerCase())) {
           resolved.add(agent.id);


LucidPaths:
Confirming this fix against a real-world production hit (self-hosted, deploy from 668c15bd lineage):

An agent authored a comment containing `[@Hackation-OpenClaw](agent://openclaw_gateway)` — an explicit mention href carrying an **adapter type string** instead of an agent UUID. `extractAgentMentionIds` extracted it verbatim, `findMentionedAgents` returned it unvalidated, and the wake loop in `server/src/routes/issues.ts` died per-wake with:

```
failed to wake agent on issue update {"agentId":"openclaw_gateway"}
PostgresError: invalid input syntax for type uuid: "openclaw_gateway"
```

The wake is silently lost (warn-level only) — the mentioned agent never wakes. Traced the exact path this PR patches: the `explicitAgentMentionIds` set bypasses the agents-table check that name-token mentions go through. The `validAgentIds` filter here closes precisely that hole.

Refs #6389 (same defect filed pre-PR).

Would be great to see this reviewed — happy to provide more repro detail if useful.

isak-ialogics:
Closing: duplicate of #7205 fix/wake-on-comment-validate-agent-mention-clean (same fix). Keeping #7205.

PR #6876, opened 2026-05-28
fix(issues): guard adoptUnownedCheckoutRun on terminal-state PATCH

isak-ialogics:
## Summary

When an agent PATCHes an issue to `done` or `cancelled`, the `adoptUnownedCheckoutRun` path should not be available — adopting ownership of a terminal issue is nonsensical and was causing unexpected ownership transfers on race-condition PATCHes.

## Changes

- `server/src/routes/issues.ts`: check if the target status is terminal before passing `allowUnownedAdoption`; pass `false` when the PATCH is moving to `done` or `cancelled`
- `server/src/services/issues.ts`: honour the new `allowUnownedAdoption` flag in `assertCheckoutOwner`
- `server/src/__tests__/issue-agent-mutation-ownership-routes.test.ts`: add regression test

## Test plan

- [ ] PATCH an issue to `done` when a different agent owns checkout — returns 403 (no longer adopts)
- [ ] PATCH an issue to `in_progress` when unowned — still adopts as before

### DIFF
--- a/server/src/__tests__/issue-agent-mutation-ownership-routes.test.ts
+++ b/server/src/__tests__/issue-agent-mutation-ownership-routes.test.ts
@@ -517,7 +517,7 @@ describe("agent issue mutation checkout ownership", () => {
       .send({ format: "markdown", body: "# updated" })
       .expect(200);
 
-    expect(mockIssueService.assertCheckoutOwner).toHaveBeenCalledWith(issueId, ownerAgentId, ownerRunId);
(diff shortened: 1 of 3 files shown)


isak-ialogics:
Closing: this branch conflicts with current upstream changes due to significant upstream divergence over the past 3+ weeks. The fix commit cherry-pick fails on current upstream/master.

If this fix is still needed, it can be re-submitted as a fresh clean cherry-pick from upstream/master.

PR #6884, opened 2026-05-28
fix(plugin-worker): handle EPIPE on child stdin write

### LINKED ISSUE #6434: EPIPE from child stdin write crashes server on startup
## Repro

Run `pnpm dev:watch` against a paperclip instance with one or more stale `heartbeatRuns` rows (any prior unclean shutdown produces this). Server fully boots, logs `reaped orphaned heartbeat runs` and `startup productivity reconciliation`, then immediately crashes with `Error: write EPIPE` from `node:internal/stream_base_commons:95`.

## Root cause

[`packages/adapter-utils/src/server-utils.ts:2059-2066`](packages/adapter-utils/src/server-utils.ts#L2059-L2066). When a child process is spawned with `opts.stdin` set, the input is written from a `spawnPersistPromise.finally` callback. The check `if (child.killed || stdin.destroyed) return;` is not sufficient — if the child exits *between* the persist promise resolving and the `stdin.write()` call (or its underlying syscall), the write produces an `EPIPE`. No `'error'` listener is attached to `stdin`, so the `EPIPE` becomes an `uncaughtException`.

## Triggering path

1. Reaper completes
2. `startNextQueuedRunForAgent` runs for one of the reaped agents
3. Spawns a tool/adapter command with stdin payload
4. Child exits fast (e.g. `ENOENT`, immediate completion, or signal)
5. `stdin.write` races and `EPIPE`s

## Suggested fix

Attach `stdin.on("error", noop)`, also check `stdin.writable`, and pass a callback to `stdin.write` to capture non-`EPIPE` errors via the existing log error path.

```ts
if (opts.stdin != null && stdin) {
  stdin.on("error", () => {});
  void spawnPersistPromise.finally(() => {
    if (child.killed || stdin.destroyed || !stdin.writable) return;
    stdin.write(opts.stdin as string, (err) => {
      if (err && (err as NodeJS.ErrnoException).code !== "EPIPE") {
        onLogError(err, runId, "failed to write stdin to child");
      }
    });
    stdin.end();
  });
}
```

isak-ialogics:
## Summary

- Attaches an `error` listener to `child.stdin` in `attachStdioHandlers` so that EPIPE events emitted by the write stream never go unhandled (an unhandled `error` event on a stream crashes the Node process)
- Wraps `stdin.write()` in `sendMessage` with a try/catch that logs EPIPE at `warn` level and re-throws any other error

Fixes https://github.com/paperclipai/paperclip/issues/6434

## Test plan

- [ ] Start a plugin worker and kill it mid-startup; server stays alive
- [ ] EPIPE log line appears at `warn` level in server output
- [ ] Non-EPIPE stdin errors still propagate (throw/re-emit)
- [ ] Existing plugin worker tests pass


isak-ialogics:
Closing: branch has wrong top commit (unrelated noise). Fix may be in another -clean branch.

PR #6885, opened 2026-05-28
fix(gemini-local): extend isGeminiUnknownSessionError to match "No previous sessions found"

### LINKED ISSUE #6806: [gemini_local] isGeminiUnknownSessionError doesn't match "No previous sessions found" — stale session causes failed run instead of auto-retry
## Summary

When Gemini CLI cannot resume a previous session, it exits with the message:

```
Error resuming session: No previous sessions found for this project.
```

The regex in `isGeminiUnknownSessionError` (in `packages/adapters/gemini-local/src/server/parse.ts`) does **not** match this string, so the auto-retry with a fresh session never fires. The run is marked `failed` instead of transparently recovering.

## Affected versions
- Gemini CLI: v0.43.0 (likely all versions that emit this error message)
- Paperclip: latest (confirmed with current dist)
- OS: Windows 11 (but the regex bug is platform-independent)

## Root cause

```ts
// parse.ts — current regex
return /unknown\s+session|session\s+.*\s+not\s+found|resume\s+.*\s+not\s+found|checkpoint\s+.*\s+not\s+found|cannot\s+resume|failed\s+to\s+resume/i.test(haystack);
```

The error message "Error resuming session: No previous sessions found for this project." fails every branch:

| Pattern | Why it doesn't match |
|---|---|
| `session.*not found` | Message says "sessions **found**", not "sessions not found" |
| `resume.*not found` | "resuming" ≠ "resume" + whitespace; and "found" ≠ "not found" |
| `unknown session` | Word "unknown" is absent |
| `cannot resume` | Absent |
| `failed to resume` | Absent |

## Impact

The retry logic in `execute.ts` is correct and already handles this case:

```ts
if (sessionId && !initial.proc.timedOut && (initial.proc.exitCode ?? 0) !== 0 &&
    isGeminiUnknownSessionError(initial.proc.stdout, initial.proc.stderr)) {
  await onLog("stdout", `[paperclip] Gemini resume session "" is unavailable; retrying with a fresh session.\n`);
  return toResult(retry, true, true);
}
```

But the guard returns `false`, so the retry never executes. The run fails permanently until the user manually triggers a fresh-session wakeup.

This is especially common on Windows where previous runs may have failed before Gemini created a session file — leaving a stale session ID in Paperclip's runtime with no corresponding `~/.gemini/tmp/` entry.

## Fix

Add the missing pattern to the regex (one-line change in `parse.ts`):

```diff
- return /unknown\s+session|session\s+.*\s+not\s+found|resume\s+.*\s+not\s+found|checkpoint\s+.*\s+not\s+found|cannot\s+resume|failed\s+to\s+resume/i.test(haystack);
+ return /unknown\s+session|session\s+.*\s+not\s+found|resume\s+.*\s+not\s+found|checkpoint\s+.*\s+not\s+found|cannot\s+resume|failed\s+to\s+resume|no\s+previous\s+sessions\s+found/i.test(haystack);
```

## Related

- PR #2877 (stdin-based prompt delivery, fixes the main Windows quoting issues — but does not address this regex gap)
- Issue #6324 (similar pattern gap in `isClaudeUnknownSessionError` for the claude_local adapter)
- Issue #1160 (stale session ID loop in hermes-local adapter — same class of bug)

isak-ialogics:
## Summary

- Adds `no\s+previous\s+sessions` to the `isGeminiUnknownSessionError` regex in `packages/adapters/gemini-local/src/server/parse.ts`
- Gemini CLI v0.43.0 emits `Error resuming session: No previous sessions found for this project.` — the previous regex did not match this string, causing runs to be marked `failed` instead of triggering auto-retry with a fresh session
- All existing regex alternatives are preserved

Fixes https://github.com/paperclipai/paperclip/issues/6806

## Test plan

- [ ] `isGeminiUnknownSessionError("", "Error resuming session: No previous sessions found for this project.")` returns `true`
- [ ] Existing session-error strings still return `true`
- [ ] Unrelated error strings still return `false`

### DIFF
--- a/packages/adapters/claude-local/src/server/parse.ts
+++ b/packages/adapters/claude-local/src/server/parse.ts
@@ -10,7 +10,7 @@ const CLAUDE_AUTH_REQUIRED_RE = /(?:not\s+logged\s+in|please\s+log\s+in|please\s
 const URL_RE = /(https?:\/\/[^\s'"`<>()[\]{};,!?]+[^\s'"`<>()[\]{};,!.?:]+)/gi;
 
 const CLAUDE_TRANSIENT_UPSTREAM_RE =
-  /(?:rate[-\s]?limit(?:ed)?|rate_limit_error|too\s+many\s+requests|\b429\b|overloaded(?:_error)?|server\s+overloaded|service\s+unavailable|\b503\b|\b529\b|high\s+demand|try\s+again\s+later|temporarily\s+unavailable|throttl(?:ed|ing)|throttlingexception|servicequotaexceededexception|out\s+of\s+extra\s+usage|extra\s+usage\b|claude\s+usage\s+limit\s+reached|5[-\s]?hour\s+limit\s+reached|weekly\s+limit\s+reached|usage\s+limit\s+reached|usage\s+cap\s+reached)/i;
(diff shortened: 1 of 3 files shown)


isak-ialogics:
Closing: duplicate of #7135 fix/gemini-unknown-session-regex-clean-rebased (same fix). Keeping #7135.

PR #6890, opened 2026-05-28
fix(ui): scope remarkLinkIssueReferences to current company prefix

### LINKED ISSUE #6182: UI generates 404 navigations from `XXX-NNN`-shaped tokens in markdown content, including template placeholders and example IDs from skill docs
## Summary

The UI in `paperclipai@2026.513.0` treats every `XXX-NNN`-shaped token in rendered markdown content (issue descriptions, comments, work products, agent-authored content) as an issue identifier and generates a navigation/link to it. This includes false positives that are clearly not issue identifiers:

- **Template placeholders** that didn't get substituted: `{issueId}`, `<issue-identifier>`
- **Example IDs from documentation and skill files**: `PAP-224`, `ZED-24`, `DAY-1` — these reference issues that exist in *other* (e.g. paperclipai-internal) instances, not the user's
- **Any plausible-looking pattern** an agent happens to write: e.g. "see PROJ-12 for context" where `PROJ` isn't even a valid prefix in the current company

Each match produces a `GET /<token>` request that returns 404 and lands in the server log as a `WARN`. Functionally cosmetic — but generates persistent log noise and degrades the user experience when a link looks clickable but goes nowhere.

## Reproduction

1. In any company, open an issue and add a comment containing this text:

```
   Reference {issueId} for context, similar to PAP-224 or DAY-1.
```

2. Save the comment.

3. Watch the server logs (`journalctl -u paperclip -f` or equivalent):

```
   WARN: GET /{issueId} 404
   WARN: GET /PAP-224 404
   WARN: GET /DAY-1 404
```

4. In the rendered comment, those tokens appear as clickable links that all lead to the 404 page.

## Expected behavior

The autolinker should only convert tokens that are valid issue references for the *current company*. Reasonable criteria:

1. The prefix matches the company's `issuePrefix` (e.g. only `NEP-N` should autolink in the Neptune Labs company)
2. The number is `<= company.issueCounter` (cheap pre-check, avoids DB lookup for `NEP-9999` when the company only has 12 issues)
3. Tokens wrapped in `{}` or `<>` should be excluded outright as template placeholders

Optionally: linked-but-not-found tokens could render as plain text instead of broken links, with no network request fired.

## Actual behavior

Every `[A-Z]+-\d+` token (and template placeholders containing curly braces or angle brackets in some cases) becomes a link that fires a navigation request, regardless of whether the prefix or number makes sense for the current company.

## Environment

- `paperclipai@2026.513.0` (latest stable as of filing)
- Self-hosted, Node v24.15.0, Ubuntu Server
- Reproduces across all companies on the instance
- Reproduces in both agent-authored content and human-typed content

## Speculative fix direction

The autolinking logic likely lives in the markdown rendering pipeline on the client side (search the UI bundle for `[A-Z]+-\d+` or similar regex). The fix would be to:

1. Pass the current company's `issuePrefix` and `issueCounter` into the rendering context
2. Only autolink tokens where prefix matches and number ≤ counter
3. Strip tokens enclosed in `{}` / `<>` before regex matching

The server-side change, if any, would be to make `GET /<unknown-token>` return a lighter response (the SPA catch-all currently logs every miss as `WARN`).

## Why this matters

Skill docs and agent prompts commonly include example IDs (`PAP-224` from the official Paperclip skill, `{issueId}` template placeholders in instructions). Agents reproduce these in their output. Every such mention currently logs as a server-side 404, which:

- Inflates log volume and obscures real warnings
- Creates broken-looking links in the UI for users who click them
- Makes log-based monitoring (e.g. alerting on 404 rates) harder to tune

Cosmetic, not blocking, but worth a quick fix.

---

Happy to dig into the relevant bundle code and submit a PR if a maintainer can point to the autolinker module — I just don't know the codebase well enough to find it on my own from the minified `dist/`.

isak-ialogics:
## Problem

`remarkLinkIssueReferences` linkified any `XXX-NNN` bare token it found in markdown, regardless of which company's workspace was being viewed. A user viewing PAP issues would see tokens like `ZED-123` or `ACME-456` silently turned into internal links — confusing and potentially leaking cross-company context.

Fixes #6182.

## Solution

Add an optional `companyPrefix` to `remarkLinkIssueReferences`, `rewriteMarkdownTree`, and the internal `linkifyIssueReferencesInText`. When a prefix is provided, bare `XXX-NNN` identifiers whose prefix doesn't match are skipped.

Explicit `issue://` URIs and `/path/.../issues/ID` URL paths are still linkified unconditionally — they carry enough context to be unambiguous.

## Changes

- `ui/src/lib/issue-reference.ts`: add `isBareIdentifierForOtherCompany()` helper; thread `options?: { companyPrefix?: string }` through `remarkLinkIssueReferences` → `rewriteMarkdownTree` → `linkifyIssueReferencesInText`

## Usage

```ts
// Only linkify PAP-NNN bare tokens; ignore ZED-NNN etc.
remarkLinkIssueReferences({ companyPrefix: "PAP" })
```

### DIFF
--- a/packages/adapter-utils/src/server-utils.ts
+++ b/packages/adapter-utils/src/server-utils.ts
@@ -37,6 +37,7 @@ interface SpawnTarget {
   args: string[];
   cwd?: string;
   cleanup?: () => Promise<void>;
+  windowsVerbatimArguments?: boolean;
 }
 
 type RemoteExecutionSpec = SshRemoteExecutionSpec;
@@ -1294,9 +1295,15 @@ async function resolveSpawnTarget(
     // ComSpec to PowerShell, which breaks cmd-specific flags like /d /s /c.
     const shell = resolveWindowsCmdShell(env);
     const commandLine = [quoteForCmd(executable), ...args.map(quoteForCmd)].join(" ");
+    // Wrap commandLine in extra outer quotes and set windowsVerbatimArguments so
+    // Node.js does not re-escape inner quotes via CreateProcess (which uses \" that
+    // cmd.exe cannot parse). cmd.exe /s /c strips exactly one outer \"...\" pair,
+    // leaving the inner cmd-quoted command intact. Without this, any path containing
+    // spaces causes cmd.exe to mis-parse the command line.
     return {
       command: shell,
-      args: ["/d", "/s", "/c", commandLine],
+      args: ["/d", "/s", "/c", `"${commandLine}"`],
+      windowsVerbatimArguments: true,
     };
   }
 
@@ -2122,6 +2129,7 @@ export async function runChildProcess(
           env: mergedEnv,
           detached: process.platform !== "win32",
           shell: false,
+          windowsVerbatimArguments: target.windowsVerbatimArguments ?? false,
           stdio: [opts.stdin != null ? "pipe" : "ignore", "pipe", "pipe"],
         }) as ChildProcessWithEvents;
         const startedAt = new Date().toISOString();
--- a/server/src/services/plugin-worker-manager.ts
+++ b/server/src/services/plugin-worker-manager.ts
@@ -434,7 +434,16 @@ export function createPluginWorkerHandle(
       throw new Error(`Worker process for plugin "${pluginId}" is not writable`);
     }
     const serialized = serializeMessage(message as any);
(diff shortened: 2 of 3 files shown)


isak-ialogics:

PR #6892, opened 2026-05-28
fix: populate lastRunError on agents table when status=error

### LINKED ISSUE #6796: agent status=error never populates lastRunError — finalizeAgentStatus drops error context from heartbeatRuns
## Summary

When an agent transitions to `status=error`, the `agents` table row receives no error context — `lastRunError` does not exist as a field, and no equivalent diagnostic is written to the agent record. The `heartbeatRuns` row for the failed run *does* contain the error (`error`, `errorCode`, `exitCode`, `signal`), but `finalizeAgentStatus` does not denormalize any of that context back to the agent row.

This has produced three separate incidents over 4 weeks where operators see an agent at `status=error` with no message, no run ID, and no clue whether to wait, reset, or investigate further.

---

## Repro Pattern (observed three times)

Trigger: any of the following causes `finalizeAgentStatus(agentId, "failed")` to be called:

- SIGTERM / exit-143 during final heartbeat (process killed mid-run, `reapOrphanedRuns` marks the run `failed` with `errorCode: "process_lost"`)
- Transient upstream API 500 during a terminal-run-recovery retry (`adapter_failed: Claude run failed: subtype=success: API Error: 500 Internal server error`)
- Unhandled adapter exception or setup failure

**Resulting agent state in all three cases:**
```json
{
  "status": "error",
  "lastHeartbeatAt": "<timestamp>",
  "lastRunError": null   // field does not exist; no error info on the agent row at all
}
```

Operators cannot distinguish a SIGTERM race from an API 500 from a bad config change — all look identical.

---

### What happens when a run fails

1. Run finalizer (`runFinalize` / `reapOrphanedRuns`) calls `setRunStatus(runId, "failed", { error, errorCode, exitCode, signal, ... })` — this writes error details to `heartbeatRuns`.
2. Immediately after, `finalizeAgentStatus(agentId, "failed")` is called.
3. `finalizeAgentStatus` computes `nextStatus = "error"` and writes `{ status: "error", lastHeartbeatAt, updatedAt }` to the `agents` table. **No error context is included.**

### Code location

**`server/src/services/heartbeat.ts`, lines 6184–6235** (`finalizeAgentStatus`):

```typescript
const updated = await db
  .update(agents)
  .set({
    status: nextStatus,
    lastHeartbeatAt: new Date(),
    updatedAt: new Date(),
    // ← error context never written here
  })
  .where(eq(agents.id, agentId))
```

Call sites that produce `outcome="failed"`:
- **line 6570** — `reapOrphanedRuns` (SIGTERM / process_lost)
- **line 7966** — normal run finalization post-adapter (API 500, adapter crash)
- **lines 8044, 8087** — setup-phase failure / outer-catch safety net

### Why the `heartbeatRuns` data doesn't help at triage time

The run record exists and has the error, but:
- The dashboard/agent-detail API returns agent-level fields, not the latest run's fields.
- Operators must know to look up runs separately, and must know which run ID to check.
- Health-check automations that read `agent.status` have no way to surface a meaningful error message without a separate query.

---

### 1 — Add `lastRunError` JSONB column to agents schema

**`packages/db/src/schema/agents.ts`**

```typescript
lastRunError: jsonb("last_run_error").$type<{
  runId: string;
  errorCode: string | null;
  message: string | null;
  exitCode: number | null;
  signal: string | null;
  occurredAt: string; // ISO 8601
} | null>(),
```

Migration: a simple `ALTER TABLE agents ADD COLUMN last_run_error jsonb;` (nullable, no default needed).

### 2 — Populate in `finalizeAgentStatus` when transitioning to error

Query the latest failed run in the same function call and write it alongside the status update:

```typescript
async function finalizeAgentStatus(
  agentId: string,
  outcome: "succeeded" | "failed" | "cancelled" | "timed_out",
) {
  // ... existing runningCount / nextStatus logic ...

  let lastRunError: Record<string, unknown> | null | undefined = undefined;
  if (nextStatus === "error") {
    const latestFailed = await db
      .select({
        id: heartbeatRuns.id,
        error: heartbeatRuns.error,
        errorCode: heartbeatRuns.errorCode,
        exitCode: heartbeatRuns.exitCode,
        signal: heartbeatRuns.signal,
        finishedAt: heartbeatRuns.finishedAt,
      })
      .from(heartbeatRuns)
      .where(
        and(eq(heartbeatRuns.agentId, agentId), eq(heartbeatRuns.status, "failed")),
      )
      .orderBy(desc(heartbeatRuns.finishedAt))
      .limit(1)
      .then((rows) => rows[0] ?? null);

    if (latestFailed) {
      lastRunError = {
        runId: latestFailed.id,
        errorCode: latestFailed.errorCode,
        message: latestFailed.error,
        exitCode: latestFailed.exitCode,
        signal: latestFailed.signal,
        occurredAt: latestFailed.finishedAt?.toISOString() ?? new Date().toISOString(),
      };
    }
  }

  await db.update(agents).set({
    status: nextStatus,
    lastHeartbeatAt: new Date(),
    updatedAt: new Date(),
    ...(lastRunError !== undefined ? { lastRunError } : {}),
  }).where(eq(agents.id, agentId));
```

### 3 — Auto-recovery: preserve `lastRunError` for one tick

When a subsequent successful heartbeat clears `status=error` → `idle`, **do not clear `lastRunError`** in the same update. This gives the next health-check or UI poll a single-tick window to capture the error before it's overwritten by the next failure (or cleared explicitly). On the following successful heartbeat, set `lastRunError: null`.

Implementation: only write `lastRunError: null` when `nextStatus === "idle"` AND `existing.lastRunError` was already `null` (i.e., it's been seen once):

```typescript
// Clear lastRunError only on second clean heartbeat
const clearLastRunError =
  nextStatus === "idle" && existing.lastRunError === null;
```

Or simpler: always write `lastRunError: null` on `nextStatus === "idle"` — that's one-tick preservation by definition since the idle update happens in the heartbeat _after_ the error was set.

### 4 — Surface in agent detail API

The Drizzle select on `agents` will include `lastRunError` automatically once the column exists. Confirm it's in the serialized response shape for `GET /api/agents/:id` and the `GET /api/agents/me` endpoint.

---

## Backward Compatibility

- Column is nullable with no default: zero migration risk for existing rows.
- No existing callers pass `lastRunError`, so the new field is additive.
- Auto-recovery semantics (`error` → `idle` on next successful heartbeat) are unchanged.

---

## Verification Steps

1. On `master`: trigger a run failure (e.g. adapter process SIGTERM or Anthropic 500 mock). Observe `GET /api/agents/:id` returns `status: "error"` with no `lastRunError` field.
2. Apply patch: repeat the same trigger. Observe `lastRunError` is populated with `{ runId, errorCode, message, exitCode, signal, occurredAt }`.
3. Trigger a successful heartbeat. Observe `status` returns to `idle` and `lastRunError` is cleared.
4. Confirm no `NOT NULL` migration failures on old agents rows.

---

## Incident References

Three occurrences at More-Than-OK/Grove (Platform Engineer agent):
- 2026-04-30 — post-task SIGTERM/exit-143 on final heartbeat
- 2026-05-16 — Anthropic API 500 during terminal-run-recovery retry
- 2026-05-27 — same null-diagnostic pattern (auto-recovered)

Each required ~1 health-check + 1 investigator + 1 review cycle. The `heartbeatRuns` row had the error in all three cases; the gap is purely the denormalization to the agent row.

isak-ialogics:
## Summary

- Adds `last_run_error` jsonb column to the `agents` table (migration 0092)
- Updates `finalizeAgentStatus` to accept optional error context (`runId`, `error`, `errorCode`, `exitCode`, `signal`) and write it when transitioning to `status=error`; clears it on idle/running transitions
- Passes error context from all failure call sites: reaped-orphan reaper, inner adapter catch, outer setup-failure catch
- Exposes `lastRunError` in the shared `Agent` type
- Shows error detail in `AgentProperties` UI when `agent.status === "error"`

Fixes #6796 — operators can now see the error message, error code, exit code, and run ID directly on the agent row without digging through heartbeat_runs.

## Test plan

- [ ] Trigger a heartbeat failure (bad adapter config, network drop, etc.) and verify `GET /api/agents/:id` returns `lastRunError` with `error`, `errorCode`, `exitCode`, `signal`, `runId` populated
- [ ] Agent detail UI shows the error block under Status when status is `error`
- [ ] Verify `lastRunError` is cleared (null) when agent recovers and next run succeeds
- [ ] Run `packages/db` migration against a local DB and confirm column exists

### DIFF
--- a/packages/adapters/claude-local/package.json
+++ b/packages/adapters/claude-local/package.json
@@ -54,6 +54,7 @@
   },
   "dependencies": {
     "@paperclipai/adapter-utils": "workspace:*",
+    "@paperclipai/mcp-server": "workspace:*",
     "picocolors": "^1.1.1"
   },
   "devDependencies": {
--- a/packages/adapters/claude-local/src/server/execute.ts
+++ b/packages/adapters/claude-local/src/server/execute.ts
@@ -359,6 +359,61 @@ export async function runClaudeLogin(input: {
   });
 }
 
+async function resolveMcpServerStdioPath(): Promise<string | null> {
+  try {
+    const resolved = import.meta.resolve("@paperclipai/mcp-server/stdio");
+    return fileURLToPath(resolved);
+  } catch {
+    return null;
+  }
+}
+
+async function injectPluginToolMcpSettings(input: {
+  cwd: string;
+  env: Record<string, string>;
+  onLog: (stream: "stdout" | "stderr", chunk: string) => Promise<void>;
+}): Promise<void> {
+  const stdioPath = await resolveMcpServerStdioPath();
+  if (!stdioPath) {
+    await input.onLog("stdout", "[paperclip] Plugin tool MCP bridge: @paperclipai/mcp-server not found, skipping plugin tool bridge.\n");
+    return;
+  }
+
+  const settingsDir = path.join(input.cwd, ".claude");
+  const settingsPath = path.join(settingsDir, "settings.local.json");
+
+  let existing: Record<string, unknown> = {};
+  try {
+    const raw = await fs.readFile(settingsPath, "utf-8");
+    const parsed = JSON.parse(raw);
+    if (parsed && typeof parsed === "object" && !Array.isArray(parsed)) {
+      existing = parsed as Record<string, unknown>;
+    }
+  } catch {
+    // File doesn't exist or is malformed — start fresh
+  }
+
+  const mcpServers = (existing.mcpServers && typeof existing.mcpServers === "object" && !Array.isArray(existing.mcpServers))
+    ? { ...(existing.mcpServers as Record<string, unknown>) }
+    : {};
+
+  const bridgeEnv: Record<string, string> = {};
+  for (const key of ["PAPERCLIP_API_URL", "PAPERCLIP_API_KEY", "PAPERCLIP_AGENT_ID", "PAPERCLIP_RUN_ID", "PAPERCLIP_COMPANY_ID"]) {
+    if (input.env[key]) bridgeEnv[key] = input.env[key];
+  }
+
+  mcpServers["paperclip-plugin-tools"] = {
+    command: process.execPath,
+    args: [stdioPath],
+    env: bridgeEnv,
+  };
+
+  const next: Record<string, unknown> = { ...existing, mcpServers };
+  await fs.mkdir(settingsDir, { recursive: true });
+  await fs.writeFile(settingsPath, `${JSON.stringify(next, null, 2)}\n`, { encoding: "utf-8", mode: 0o600 });
+  await input.onLog("stdout", `[paperclip] Plugin tool MCP bridge: injected ${Object.keys(bridgeEnv).length > 0 ? "with credentials" : "without credentials"} into ${settingsPath}.\n`);
+}
+
 export async function execute(ctx: AdapterExecutionContext): Promise<AdapterExecutionResult> {
   const { runId, agent, runtime, config, context, onLog, onMeta, onSpawn, authToken } = ctx;
   const executionTarget = readAdapterExecutionTarget({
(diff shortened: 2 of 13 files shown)


isak-ialogics:

PR #6903, opened 2026-05-28
feat(board-auth): add PAPERCLIP_BOARD_API_KEY_TTL_DAYS env override

### LINKED ISSUE #6858: Board API key TTL improvement
Hi team,

We run Paperclip in production at CashTime Pay AG, where 13 AI agents (`AICEO`, `AIPM`, `AIBM-Recon`, `AIBM-Outreach`, `AICROPS`, `AISMM`, etc.) drive our daily Sales/Outreach pipeline. They depend on a Paperclip board API token (`pcp_board_*`) to:
- Create heartbeat runs and wakeup requests via `POST /api/agents/:id/wakeup`
- Create / comment / close Plane-mirrored issues
- Coordinate handoffs between agents through `cashtime-agent-bus`

This token sits in `~/stack/cashtime_agent_bus/.env` as `PAPERCLIP_API_TOKEN` and is read by a `systemd`-managed service that runs every 15 minutes (`cashtime-aibm-trigger.timer`).

**Today our outreach pipeline silently broke** because the token expired at exactly `2026-05-28 11:11:40 UTC` — 30 days after creation. Every AIBMO dispatch since then returns `401 Unauthorized` from `POST /api/companies/.../issues`, so no outreach issue has been created on a real lead since 11:11 today (we caught it ~5 hours later, after losing ~20 dispatches across V1-Nigeria and V2-Pharma-UA campaigns).

After reading `server/src/services/board-auth.ts` I found the root cause:

```ts
export const BOARD_API_KEY_TTL_MS = 30 * 24 * 60 * 60 * 1000;  // hardcoded 30 days
export const CLI_AUTH_CHALLENGE_TTL_MS = 10 * 60 * 1000;

export function boardApiKeyExpiresAt(nowMs: number = Date.now()) {
  return new Date(nowMs + BOARD_API_KEY_TTL_MS);
}
```

Issues with the current design from an operator’s perspective:

1. **No env / config override.** There is no `PAPERCLIP_BOARD_API_KEY_TTL_DAYS` (or similar) to extend the TTL for service accounts. The constant is compiled into the docker image, so even modifying `BOARD_API_KEY_TTL_MS` requires forking and rebuilding `ghcr.io/cashtimepay/cashtimepay-ai-team`.
2. **No expiry warning anywhere.** Paperclip UI, the agent dashboard, the API itself — none of them surface `expires_at` for an existing token. The first signal an operator gets is a flood of `401`s in service logs, often days after the silent failure starts.
3. **No automated renewal path.** The only re-issuance mechanism is `POST /api/cli-auth/challenges` followed by a manual browser-approve step. A human has to remember to do this every 30 days for every service account in every environment.
4. **CLI auth challenge itself has a 10-min TTL** — if the operator misses the window, the whole loop has to be restarted.
5. **No service-account / machine-user model.** All board API keys are tied to a human user, with no concept of an unattended long-lived service identity (the kind GitHub calls a “fine-grained PAT” or AWS calls an “IAM access key”).

What we would like to see (in rough priority order):

- **Env-driven TTL.** A `PAPERCLIP_BOARD_API_KEY_TTL_DAYS` env on the server, defaulting to 30 but allowing `365` / `0` (never) for production deployments. Five-line change in `board-auth.ts`.
- **Service-account tokens** that explicitly bypass the user-TTL policy. UI button: “Create long-lived service token (no automatic expiry)” with an audit log showing who created it and when it was last used.
- **Expiry warning.** UI badge + agent log message starting 7 days before token expiry, with a link to the re-issue flow.
- **Self-rotation API.** `POST /api/board-api-keys/:id/rotate` that issues a new secret while keeping the same ID and skipping the CLI challenge for service accounts. Lets us rotate via cron without human approval.

For now we’re patching this on our side (probably forking + bumping `BOARD_API_KEY_TTL_MS` to `365 days`, or adding a nightly cron that does `UPDATE board_api_keys SET expires_at = now() + interval '30 days' WHERE name='agent-bus systemd (board)' AND revoked_at IS NULL` directly against the postgres). Both are workarounds, not solutions — we’d much rather use a supported upstream mechanism.

Happy to help: I can put together a PR for the env-driven TTL myself if you would accept that as the first step. Just let me know which direction you prefer (env override vs full service-account model) and we’ll align with that.

Thanks for the great product overall — Paperclip is the backbone of how our AI team operates, and we want to keep it that way.

Best,
Dmitry Radionov  
Chief Revenue Officer  
CashTime Pay AG (CHE-435.941.738), Zug, Switzerland  
dmitry@cashtimepay.com

isak-ialogics:
## Summary

Closes #6858

The 30-day board API key TTL was hardcoded in `BOARD_API_KEY_TTL_MS`. Service accounts using long-lived `pcp_board_*` tokens had no way to extend the TTL without forking and rebuilding the Docker image — leading to silent `401 Unauthorized` failures when tokens expired.

**Fix:** resolve `BOARD_API_KEY_TTL_MS` at startup from `PAPERCLIP_BOARD_API_KEY_TTL_DAYS` env var. If not set, defaults to 30 days (no change in behavior).

**Usage:**
```
# Extend to 365 days for service accounts
PAPERCLIP_BOARD_API_KEY_TTL_DAYS=365

# Shorten to 12 hours for short-lived automation
PAPERCLIP_BOARD_API_KEY_TTL_DAYS=0.5
```

## Test plan
- [ ] `PAPERCLIP_BOARD_API_KEY_TTL_DAYS=365` → newly created board API keys expire 365 days from creation
- [ ] Not set → board API keys expire 30 days from creation (existing behavior)
- [ ] Invalid value (non-numeric, 0, negative) → falls back to 30 days

### DIFF
--- a/_model_config.yaml
+++ b/_model_config.yaml
@@ -1,0 +1,10 @@
+model:
+  model_name: openai/qwen/qwen3-next-80b
+  model_class: litellm_textbased
+  model_kwargs:
+    api_base: http://192.168.0.7:1234/v1
+    api_key: lm-studio
+    max_tokens: 65536
+    drop_params: true
+    request_timeout: 300
+  cost_tracking: ignore_errors
--- a/packages/adapter-utils/src/server-utils.ts
+++ b/packages/adapter-utils/src/server-utils.ts
@@ -1294,9 +1294,16 @@ async function resolveSpawnTarget(
     // ComSpec to PowerShell, which breaks cmd-specific flags like /d /s /c.
     const shell = resolveWindowsCmdShell(env);
     const commandLine = [quoteForCmd(executable), ...args.map(quoteForCmd)].join(" ");
+    // cmd.exe /s strips the outermost quote pair from the /c argument. When the
+    // executable path contains spaces it is already wrapped in quotes by
+    // quoteForCmd; those inner quotes would be stripped, leaving the path
+    // broken at the first space. Wrapping the entire commandLine in an extra
+    // pair of quotes ensures the inner executable quotes survive cmd.exe's
+    // quote-stripping pass (GH#6805).
+    const wrappedCommandLine = commandLine.startsWith('"') ? `"${commandLine}"` : commandLine;
     return {
       command: shell,
-      args: ["/d", "/s", "/c", commandLine],
+      args: ["/d", "/s", "/c", wrappedCommandLine],
     };
   }
 
--- a/packages/adapter-utils/src/server-utils.ts
+++ b/packages/adapter-utils/src/server-utils.ts
@@ -2063,10 +2063,12 @@ export async function ensureCommandResolvable(
     if (resolvedSsh) return;
     throw new Error('Command not found in PATH: "ssh"');
   }
-  const resolved = await resolveCommandPath(command, cwd, env);
+  // Extract only the executable (first token) — the rest are arguments.
+  const executable = command.trimStart().split(/\s+/)[0] ?? command;
+  const resolved = await resolveCommandPath(executable, cwd, env);
   if (resolved) return;
-  if (command.includes("/") || command.includes("\\")) {
-    const absolute = path.isAbsolute(command) ? command : path.resolve(cwd, command);
+  if (executable.includes("/") || executable.includes("\\")) {
+    const absolute = path.isAbsolute(executable) ? executable : path.resolve(cwd, executable);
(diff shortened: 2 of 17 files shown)


isak-ialogics:
Closing: superseded by #7060 feat/board-api-key-ttl-clean (same feature). Keeping #7060.

PR #6906, opened 2026-05-28
fix(redaction): always redact type:plain env binding values in configuration API

### LINKED ISSUE #6816: Adapter env GET /api/agents/:id/configuration returns type:plain values unredacted, despite docs claim of redaction
## Summary

`GET /api/agents/:id/configuration` returns env vars of `type: plain` with their raw values in the response body. The `/llms/agent-configuration.txt` doc says "Sensitive values are redacted in configuration read APIs" — they currently are not.

For agents that include credential-shaped values in their `adapter_config.env` (e.g., via the legacy plain-string env binding, or via operator-driven UI flows that store as plain), this surface returns the raw credential value over the API. Any agent that can call its own `GET /api/agents/me/configuration` can read its own credentials in plaintext. This contradicts the architectural intent in `doc/SPEC-implementation.md` §7.13 which says "Secret values are not stored inline in `agents.adapter_config.env`" and "Activity and approval payloads must not persist raw sensitive values."

## Reproduction

1. As a board user, create a company secret via `POST /api/companies/{companyId}/secrets` (e.g., name `MY_TEST_SECRET`, value `dummy-test-value-not-a-real-credential`).
2. Create an agent via `POST /api/companies/{companyId}/agents` with `adapter_config.env.MY_TOKEN = "dummy-test-value-not-a-real-credential"` (legacy plain-string form, which Zod's `envBindingSchema` still accepts for backward compat).
3. As the agent itself (or any actor with access to that agent's record), call `GET /api/agents/{agentId}/configuration`.
4. Observe: response body contains `"env": { "MY_TOKEN": { "type": "plain", "value": "dummy-test-value-not-a-real-credential" } }` — the value is returned verbatim.

## Expected behavior

Per the docs (`/llms/agent-configuration.txt`), sensitive values in `type: plain` env entries should be redacted in the configuration read API response. Either:

- Return a redacted placeholder: `{ "type": "plain", "value": "<REDACTED>" }`
- Or return the metadata only with a flag: `{ "type": "plain", "value_redacted": true }`

The same behavior should apply to `/api/agents/:id` (the parent read endpoint) for any caller below board level, since both endpoints surface env values in the same shape.

## Workaround (current)

Use `type: secret_ref` bindings instead of `type: plain`. `secret_ref` returns `{ type: "secret_ref", secretId: <uuid>, version: "latest" }` from the read API — no value leak. This is the architecturally-correct pattern per `doc/SPEC-implementation.md` §7.13 anyway. Document the redaction gap explicitly in agent-configuration docs so deployments don't rely on the legacy plain-string binding for sensitive material.

## Source

This was surfaced by a self-hosted Paperclip deployment on 2026-05-27 during a credential-rotation incident. Two PATs were stored as `type: plain` and read back in plaintext via `/api/agents/:id/configuration` by the agents themselves. Treated as a credential exposure event; PATs rotated. Storage-side bug fixed in the deployment by switching to `secret_ref` bindings. The read-API redaction gap remains.

Happy to provide additional reproduction detail or contribute a fix if useful. Not blocking us on our side (we've migrated off plain bindings), but the redaction gap is a real platform contract violation that could bite other deployments.

isak-ialogics:
Closes #6816

## Summary

`sanitizeValue` was returning plain bindings with their raw values intact. `sanitizeRecord` only redacted them for keys matching the secret-name regex — keys like `MY_TOKEN` not matching the pattern passed through unredacted.

Since any `type: plain` env binding is operator-supplied and potentially a credential, plain binding values are now unconditionally redacted to `"***REDACTED***"` in `sanitizeValue`, regardless of key name.

`secret_ref` bindings continue to return their metadata (secretId, version) unchanged.

## Test plan

- [ ] `GET /api/agents/:id/configuration` — `type: plain` env entries return `value: "***REDACTED***"`
- [ ] `GET /api/agents/:id/configuration` — `type: secret_ref` entries return metadata unchanged
- [ ] Key with secret-sounding name still redacted (regression)

### DIFF
--- a/packages/adapters/claude-local/src/server/parse.ts
+++ b/packages/adapters/claude-local/src/server/parse.ts
@@ -10,7 +10,7 @@ const CLAUDE_AUTH_REQUIRED_RE = /(?:not\s+logged\s+in|please\s+log\s+in|please\s
 const URL_RE = /(https?:\/\/[^\s'"`<>()[\]{};,!?]+[^\s'"`<>()[\]{};,!.?:]+)/gi;
 
 const CLAUDE_TRANSIENT_UPSTREAM_RE =
-  /(?:rate[-\s]?limit(?:ed)?|rate_limit_error|too\s+many\s+requests|\b429\b|overloaded(?:_error)?|server\s+overloaded|service\s+unavailable|\b503\b|\b529\b|high\s+demand|try\s+again\s+later|temporarily\s+unavailable|throttl(?:ed|ing)|throttlingexception|servicequotaexceededexception|out\s+of\s+extra\s+usage|extra\s+usage\b|claude\s+usage\s+limit\s+reached|5[-\s]?hour\s+limit\s+reached|weekly\s+limit\s+reached|usage\s+limit\s+reached|usage\s+cap\s+reached)/i;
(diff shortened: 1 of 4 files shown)


isak-ialogics:

PR #6916, opened 2026-05-28
feat: add GET /cli-auth/keys endpoint exposing board API key expiry

### LINKED ISSUE #6858: Board API key TTL improvement

isak-ialogics:
## Summary

- Operators had no visibility into existing board API keys or their `expiresAt` — silent 401 failures were the first signal of expiry
- The `PAPERCLIP_BOARD_API_KEY_TTL_DAYS` env override was already implemented; this PR adds the missing observability layer

## Changes

- `board-auth.ts`: adds `listBoardApiKeys(userId)` — queries active (non-revoked) board API keys for a user, returning id, name, createdAt, expiresAt, lastUsedAt
- `access.ts`: adds `GET /cli-auth/keys` route returning all active board API keys for the authenticated board user, with `expired` boolean and `daysUntilExpiry` computed fields

## Test plan

- [ ] `GET /cli-auth/keys` with a valid board API key returns the list of active keys with correct `expiresAt` and `daysUntilExpiry`
- [ ] `expired: true` is set when `expiresAt` is in the past
- [ ] Route requires board authentication — unauthenticated requests receive 401

Closes #6858


isak-ialogics:

Apologies for the noise.

PR #6917, opened 2026-05-28
fix: use shell:true for .cmd/.bat on Windows to avoid path-quoting corruption

### LINKED ISSUE #6805: claude_local adapter fails on Windows when user profile path contains a space
## Summary

On Windows, the `claude_local` adapter mis-quotes paths containing spaces when invoking Claude Code, causing every run to fail. The bug affects users whose Windows username (and therefore home directory) contains a space, e.g. `C:\Users\First Last\`.

The bug surfaced in two places, suggesting the same quoting logic is applied to multiple adapter arguments.

## Environment

- Windows 11
- Username with a space: `User One`
- Paperclip: installed via `npx paperclipai run` (latest as of 2026-05-27)
- Claude Code: installed globally via npm to `C:\Users\User One\AppData\Roaming\npm\claude.cmd`
- `where.exe claude` confirms the binary is on PATH and resolvable

## Symptom 1 — Executable path quoting

Initial runs failed immediately with:

'"C:\Users\USER\AppData\Roaming\npm\claude.CMD"' is not recognized as an internal or external command,
operable program or batch file.

Note the literal double-quotes wrapping the path. cmd.exe is treating the quotes as part of the filename. The space in the username appears to be what triggers the over-quoting.

**Workaround:** In the agent's `claude_local` adapter config, set Command to the 8.3 short path explicitly:
C:\Users\USER~1\AppData\Roaming\npm\claude.cmd

After this, the adapter Test passed (Working directory valid, Command executable, Claude hello probe succeeded).

## Symptom 2 — Append-system-prompt path quoting

After fixing Symptom 1, runs failed at a later stage:

Error: Append system prompt file not found: C:\Users\User.paperclip\instances\default\workspaces<uuid>"C:\Users\User

Two paths are concatenated, the second starts with a literal `"` and is truncated at the space — same root cause (mis-quoting of a path containing a space) on a different argument that Paperclip generates internally for run workspaces.

**Workaround:** Relocate the entire data directory to a space-free path using `paperclipai run --data-dir C:\paperclip-data`. (Note: the config file's `embeddedPostgresDataDir`, `backup.dir`, `logging.logDir`, `storage.localDisk.baseDir`, and `secrets.localEncrypted.keyFilePath` had absolute paths baked in from onboarding, so a naive `Move-Item` of `.paperclip` did not work — the config file had to be edited to replace the old paths before restart.)

After relocation, the originally-failing issue completed successfully end-to-end in ~2 minutes.

## Suggested fix

Wherever the adapter constructs the Claude Code command line on Windows (both for the executable and for the `--append-system-prompt` argument), arguments should be quoted exactly once. Currently it looks like paths are being pre-wrapped in quotes and then the spawn call wraps them again, producing the `'"..."'` you see in cmd.exe error output.

A second smaller suggestion: `paperclipai run --data-dir <path>` could optionally rewrite the absolute paths in `config.json` to be relative to the new data dir, so that moving the data directory works without manual editing.

## Severity

Blocking for any Windows user whose home directory contains a space. Currently the workaround requires both an 8.3 short path AND a separate data directory — neither is discoverable without significant debugging.

isak-ialogics:
## Summary

When a user's home directory contains a space (e.g. `C:\Users\First Last\`), the `claude_local` adapter fails on Windows because of a cmd.exe quoting quirk.

**Root cause:** `resolveSpawnTarget` manually built a `cmd.exe /d /s /c <commandLine>` invocation. With `/s`, cmd.exe strips the **first and last double-quote** from the entire command string — not just the outer wrapper. This causes the executable's closing quote to be consumed correctly but then cmd.exe looks for the last `"` in the entire string, which is the closing quote of `--append-system-prompt-file "path with spaces"`. The executable path ends up without a closing quote, breaking parsing.

**Error seen:**
```
'"C:\Users\User One\AppData\Roaming\npm\claude.CMD"' is not recognized as an internal or external command
```

## Fix

Return `{ command: executable, args, shell: true }` from `resolveSpawnTarget` for `.cmd`/`.bat` files and let Node's `spawn` build the `cmd.exe /c` invocation. Node correctly handles path quoting in this case.

Also:
- Disables `detached` mode when `shell: true` (shell-wrapped processes can't be killed via process group)
- Removes now-unused `quoteForCmd` and `resolveWindowsCmdShell` helpers

## Test plan

- [ ] On Windows with a username containing a space, `claude_local` adapter runs successfully end-to-end
- [ ] On Windows with a space-free username, adapter still works (no regression)
- [ ] On Linux/macOS, `.cmd`/`.bat` branch is unreachable — no regression

Closes #6805


isak-ialogics:

Apologies for the noise.

PR #6918, opened 2026-05-28
feat: make billingCode editable in issue Properties panel

### LINKED ISSUE #6845: Expose billingCode field in New Issue form UI and issue detail panel
## Summary

`billingCode` is supported by the Paperclip API and the MCP server, but it is not exposed anywhere in the web UI. The New Issue form has no input for it, and the issue detail Properties panel does not display it. This forces users who need `billingCode` populated to create issues exclusively via the API or MCP, which is workable but not ideal.

## Current state (verified on v2026.416.0)

- **API:** `billingCode` field accepted on issue create and update.
- **MCP server (`@paperclipai/mcp-server`):** `paperclipCreateIssue` and `paperclipUpdateIssue` both accept `billingCode`.
- **Web UI New Issue form:** no input for `billingCode`. Checked the default form, both "..." expander menus, and inspected the rendered HTML — the string "billingcode" does not appear anywhere in the form's DOM.
- **Web UI issue detail Properties panel:** `billingCode` is not displayed, even on issues where it is populated.

Tested by creating an issue via the MCP with `billingCode: "example-client-slug"`, then opening it in the web UI. The field is set in the database (confirmed via API readback) but not visible to the user anywhere in the UI.

## Why this matters

For multi-client agencies and any organization that needs to attribute Paperclip work to a billing or cost-tracking dimension, `billingCode` is the natural place to record that. Without UI exposure:

1. Users who create issues manually through the web UI cannot set `billingCode`, so issues created that way are always missing it.
2. Users cannot see at a glance which client or billing dimension an issue belongs to when viewing it in the UI.
3. Any team workflow that depends on `billingCode` accuracy has to enforce it externally (custom validation skills, MCP wrappers, etc.) rather than relying on the UI as the source of truth.

In our case, every Paperclip issue must carry a `billingCode` matching a canonical client slug, and the rule is enforced by a custom validation skill we invoke before every `paperclipCreateIssue` call. That works for AI-driven issue creation, but it leaves UI-created issues uncovered — which is why we've had to make a working rule that "all issue creation must go through the MCP path until the field is exposed in the UI."

## Suggested minimum scope

Two changes that would close the gap:

1. **New Issue form:** add an optional `billingCode` text input. Either inline in the default form or behind one of the existing "..." expander menus.
2. **Issue detail Properties panel:** display `billingCode` as a read/editable field when it is set on an issue.

A nice-to-have but not required for v1: validation hooks (e.g., a regex pattern or a configurable allow-list of values per company) so teams that enforce a canonical set of `billingCode` values can do so at the form level instead of post-hoc.

## Workaround we use today

- All issue creation goes through the MCP, never the web UI.
- A pre-call validation skill checks `billingCode` against a canonical slug list and refuses to call `paperclipCreateIssue` if the value is missing or unknown.
- Agent instructions explicitly tell agents to inherit `billingCode` from the parent issue when creating sub-issues.

Happy to share more detail on the workflow if it would help shape the UI design.

## Environment

- Paperclip version verified: `v2026.416.0`
- Verification date: 2026-05-11

isak-ialogics:
## Summary

`billingCode` was only shown as a read-only copyable field in the Properties panel (and only when already set), forcing users to use the API/MCP to set or change it.

## Changes

- `ui/src/components/IssueProperties.tsx`: Replaces the static `TruncatedCopyable` display with an inline `<input>` that:
  - Shows always (not just when set), with placeholder `e.g. client-slug`
  - Commits on blur or Enter via `onUpdate({ billingCode: value || null })`
  - Reverts to server value on Escape
  - Uses `font-mono` + accent hover/focus to match the visual style of other metadata fields
  - Syncs with the issue prop via `useEffect` (handles external updates)

## Test plan

- [ ] Open an issue with no `billingCode` — input shows with placeholder
- [ ] Type a value, press Enter or tab away — value is saved and reflected in API
- [ ] Press Escape — value reverts to previously saved value
- [ ] Open an issue that already has a `billingCode` — input pre-filled with the value
- [ ] Clear the value and blur — `billingCode` is set to null in API

Closes #6845


isak-ialogics:

Apologies for the noise.

PR #6929, opened 2026-05-29
fix(db): verify postmaster.pid PID is an actual Postgres process before reusing

### LINKED ISSUE #6597: Bug: embedded Postgres liveness check trusts postmaster.pid PID without verifying it's actually Postgres — causes ECONNREFUSED on reboot
## Environment

- `paperclipai` 2026.513.0, macOS (arm64)
- Embedded PostgreSQL (via `@embedded-postgres/darwin-arm64`)
- Single instance (`instances/default`, port 54329)

## Summary

After a crash or reboot, `instances/default/db/postmaster.pid` may reference a **PID that the OS has recycled to an unrelated process**. Paperclip's startup check sees "a process with that PID exists" and logs:

```
Embedded PostgreSQL already running; reusing existing process (pid=XXXX, port=54329)
```

It does **not** start a fresh Postgres. The server then fails with:

```
connect ECONNREFUSED 127.0.0.1:54329
```

## Observed case

After a reboot, macOS recycled the old Postgres PID to a Chrome Helper process. Paperclip read `postmaster.pid`, called `process.kill(pid, 0)` (or equivalent), got no error (the Chrome process exists), concluded Postgres was running, and skipped spawning — resulting in ECONNREFUSED on every DB operation.

`postmaster.pid` line 4 contains the port (54329); the file is only cleaned up on a clean Postgres shutdown, not on crash/reboot.

## Root cause

The liveness check verifies **PID existence** but not:
- That the process is actually a `postgres` / `postmaster` binary
- That port 54329 is actually listening

## Suggested fix

Before concluding the embedded Postgres is already running, verify at least one of:
1. The process at the recorded PID has a name matching `postgres` / `postmaster`
2. Port 54329 is actually listening (`pg_isready -p 54329` or a TCP connect probe)

If either check fails, treat `postmaster.pid` as stale: move/delete it and start a fresh Postgres instance. The data directory is safe to keep — only the lock file is stale.

## Workaround today

```bash
# Confirm nothing is actually on port 54329
lsof -nP -iTCP:54329 -sTCP:LISTEN

# If empty, move the stale lock and restart Paperclip
mv ~/.paperclip/instances/default/db/postmaster.pid \
   ~/.paperclip/instances/default/db/postmaster.pid.stale-bak
# then: paperclipai run
```

> Note: this is a separate issue from the heartbeat recovery storm (#6596) — different fix surface, included here so it doesn't get buried in that report.

isak-ialogics:
Fixes https://github.com/paperclipai/paperclip/issues/6597

After a system reboot, an unrelated process can inherit the PID stored in postmaster.pid. The existing check (process.kill(pid, 0)) only verifies the process exists, not that it is Postgres. This causes a false-positive healthy report leading to ECONNREFUSED errors.

Add isPostgresPid(pid) which checks /proc/<pid>/comm on Linux (no subprocess) or falls back to ps. Gates the runningPid early-return on a confirmed Postgres process name.


isak-ialogics:

Apologies for the noise.

PR #6930, opened 2026-05-29
fix(gemini-local): detect 'Invalid session identifier' as expired session

### LINKED ISSUE #6552: Bug: Gemini adapter fails hard on expired session resume instead of falling back to fresh session
## Summary

When using the `gemini_local` adapter, Paperclip persists session IDs in `agent_task_sessions` and attempts to resume them on subsequent heartbeat runs. Gemini sessions expire faster than Claude sessions (typically within 15-30 minutes of inactivity). When the resume fails with "Invalid session identifier," Paperclip marks the entire run as `adapter_failed` instead of clearing the stale session and retrying with a fresh session. This causes agents to enter an error/idle flapping loop that wastes heartbeat cycles and produces zero useful work.

## Steps to Reproduce

1. Configure an agent with `adapter_type = gemini_local`
2. Let it complete a heartbeat run successfully (session ID saved to `agent_task_sessions`)
3. Wait longer than the Gemini session expiry window (~15-30 min of inactivity)
4. Trigger the next heartbeat run (timer or assignment wake)

## Expected Behavior

Paperclip attempts to resume the saved session. Gemini API returns "Invalid session identifier." Paperclip clears the stale session from `agent_task_sessions`. Paperclip retries the run with a fresh session. Run succeeds.

## Actual Behavior

Paperclip attempts to resume. Gemini API returns "Invalid session identifier." Paperclip marks the run as `adapter_failed` with error: `Error resuming session: Invalid session identifier "{uuid}"`. Agent status set to `error`. Error-recovery resets agent to `idle`. Next heartbeat tries to resume the SAME stale session. Fails again. Loop repeats indefinitely.

## Impact

- Each failed resume wastes a heartbeat cycle with zero productive work
- Agents flap error/idle indefinitely until manual intervention
- We accumulated 11,947+ stale sessions across 15 gemini_local agents
- One agent (Scrum Master) had 78 session records in agent_task_sessions
- Error code `adapter_failed` is misleading -- the adapter works, the session is expired

## Workaround

Cron job every 5 minutes to purge stale sessions:
```sql
DELETE FROM agent_task_sessions
WHERE last_error IS NOT NULL
AND agent_id IN (SELECT id FROM agents WHERE adapter_type = 'gemini_local');

DELETE FROM agent_task_sessions
WHERE updated_at < NOW() - INTERVAL '30 minutes'
AND agent_id IN (SELECT id FROM agents WHERE adapter_type = 'gemini_local');
```
This prevents accumulation but the first run after a session expires still fails.

## Suggested Fix

In `services/heartbeat.js`, when the adapter returns "Invalid session identifier":
1. Delete the stale session from `agent_task_sessions`
2. Clear `session_id_before` on the current heartbeat_run
3. Retry with `freshSession: true`
4. Log a warning (not error) so it doesn't trigger error-recovery

Alternatively, add a config option like `maxSessionAge` or `sessionReuse: false` for adapters with short-lived sessions.

## Additional Context

- Claude sessions (`claude_local`) are long-lived and resume fine. Bug is Gemini-specific.
- `agent_task_sessions` grows unboundedly for Gemini agents -- needs TTL or max-rows.
- A distinct error code `session_expired` vs `adapter_failed` would improve observability.

isak-ialogics:
## Summary

- Adds `invalid\s+session` to the `isGeminiUnknownSessionError` detection regex in `packages/adapters/gemini-local/src/server/parse.ts`
- Newer Gemini CLI versions emit "Invalid session identifier" when resuming an expired session; the old regex only matched "unknown session" and similar phrases
- This unblocks the existing fresh-session retry path in `execute.ts` (lines 659-669) which already handles the fallback correctly — it just wasn't triggered for this error message
- Adds two test cases in `parse.test.ts` covering the new pattern

Fixes https://github.com/paperclipai/paperclip/issues/6552

## Test plan

- [ ] Verified regex matches "Invalid session identifier" and "invalid session abc123"
- [ ] Verified regex still rejects unrelated errors ("rate limit exceeded", "auth required")
- [ ] Existing retry logic in `execute.ts` clears session and retries with fresh session transparently


isak-ialogics:

Apologies for the noise.

PR #6932, opened 2026-05-29
fix(issues): use effective post-patch assignee in monitor authz check

isak-ialogics:
## Thinking Path

> - Paperclip allows human board users to monitor issues via a monitor policy
> - The PATCH /api/issues/:id route validates that the requesting agent can manage the monitor field
> - The authz check used  — the value before the patch was applied
> - For unassigned issues being self-assigned in the same PATCH, this meant the check ran against null even though the effective post-patch assignee was the requesting agent
> - This PR reads the effective post-patch assignee before the authz check so self-assign + monitor can succeed in one request
> - The benefit is agents can assign themselves and set a monitor in a single operation

## What Changed

- : compute effective assignee from patch body before calling 
- : added test covering self-assign + monitor in a single PATCH

## Verification

- Ran  — pass including new test case

## Risks

Low risk — logic change is localised to the monitor authz check. Existing authorised flows are unaffected.

## Model Used

- Provider: Nous Research Hermes Agent v0.15.1
- Model: qwen/qwen3.6-27b (Qwen3.6-27B-Instruct Q8_0, 27B dense, 131k context)
- Infrastructure: LM Studio on local GPU via PaperclipForge hermes_local adapter
- Capabilities: terminal, file read/write, search tools

## Checklist

- [x] Model Used specified
- [x] Bug fix — Path 1
- [x] Tests pass with new test added
- [x] 2 files changed

### DIFF
--- a/server/src/__tests__/issue-execution-policy-routes.test.ts
+++ b/server/src/__tests__/issue-execution-policy-routes.test.ts
@@ -605,4 +605,154 @@ describe("issue execution policy routes", () => {
       }),
     );
   });
+
+  it("allows a non-CEO agent to self-assign and set a monitor in a single PATCH", async () => {
+    const issue = {
+      id: "aaaaaaaa-aaaa-4aaa-8aaa-aaaaaaaaaaaa",
+      companyId: "company-1",
+      status: "todo",
+      assigneeAgentId: null,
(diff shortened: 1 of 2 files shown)


isak-ialogics:

PR #6933, opened 2026-05-29
fix(wakes): await assignment wake on issue create/child/interaction routes

### LINKED ISSUE #6556: Wake event delivery gap: agents assigned in initial POST don't reliably wake until @-mention
# Wake event delivery gap: agents assigned in initial POST don't reliably wake until @-mention

## Summary

When a new issue is created via POST `/api/companies/{companyId}/issues` with `assigneeAgentId` set in the initial payload, the assigned agent does NOT reliably wake. The issue appears in their inbox correctly, but no `issue_assigned` event seems to fire (or it fires and gets lost). An explicit @-mention comment posted afterward reliably triggers wake. Suggests POST-with-assignee path is missing the event-emission that PATCH-assignee path has.

## Environment

- Paperclip 2026.512.0
- `claude_local` adapter
- Both Tier 1 (Opus 4.7) and Tier 2 (Codex) agents affected — observed across Architect (Tier 1), CEO (Tier 1), Local Eng (Tier 2), Codex Eng (Tier 2)

## Reproduction

1. POST a new issue:
   ```bash
   curl -X POST .../api/companies/{companyId}/issues -d '{
     "title": "Test issue",
     "description": "...",
     "assigneeAgentId": "<agent-uuid>",
     "status": "todo",
     "priority": "high",
     "parentId": "<some-parent>"
   }'
   ```
2. Issue is created successfully. Response shows `assigneeAgentId` set, `status: todo`.
3. Wait 5-10 minutes.
4. Check agent's runs — no new run since the POST.
5. Post a comment on the new issue containing `@<AgentName>` mention.
6. Agent wakes within ~30-60 seconds of the comment and begins processing.

## Expected behavior

POST-with-assignee should trigger an `issue_assigned` event (or equivalent) that wakes the agent, identical to PATCHing assignee onto an existing issue.

## Actual behavior

- POST-with-assignee: no wake, agent silent until other trigger.
- PATCH-assignee (after the fact): wakes reliably (we've observed).
- Comment with @-mention: wakes reliably.

We observed this pattern today across 3 newly-created child issues (ADF-503/504/505), all assigned at create-time to Architect or CEO. All 3 sat silent for ~10 minutes after creation. Posted @-mention comments → all 3 woke within 1-2 minutes and completed work successfully in 5-8 minutes each.

## Impact

Anyone doing factory-style decomposition (CEO files multiple child issues, expects assignees to pick up automatically) will see their decomposed children sit idle until operator @-mentions each. Adds friction proportional to decomposition fan-out.

## Workaround we currently use

For every new POST that includes `assigneeAgentId`:
1. POST the issue.
2. Immediately POST a comment with `@<AgentName>` mention to trigger wake.

OR: 

1. POST the issue WITHOUT assignee (omit `assigneeAgentId`).
2. Immediately PATCH the issue with `assigneeAgentId`.

Both add an extra API call per issue. For batch decomposition (e.g. 6 child issues from a milestone parent), this becomes 6 extra calls.

## Speculation on root cause

Two paths that should produce the same event likely diverged:
- PATCH-assignee path: explicitly emits `issue_assigned` event on assignee change.
- POST-with-assignee path: may emit `issue_created` but not `issue_assigned` (assignee was "always" set, never changed from null).

If event subscribers only listen for `issue_assigned`, POST-with-assignee silently misses the wake.

## Notes

- Easy to verify: tail server logs during the POST vs PATCH paths and see which events are emitted.
- Related: status auto-transition gap (filed separately).
- Related: workspace inheritance gap (filed separately).
- Polite bug report from downstream user. Happy to test a fix if one lands.

isak-ialogics:
## Problem

Fixes #6556. Agents assigned in the initial `POST /api/companies/{id}/issues` call were not reliably woken because `queueIssueAssignmentWakeup` was called with `void` (fire-and-forget) in three places:

- `POST /companies/:companyId/issues` (main create route)
- `POST /issues/:id/children` (child issue create)
- Interaction accept route (bulk `createdIssues` loop)

Any internal error in the wake dispatch was silently swallowed; the agent only picked up the issue on a subsequent @-mention or periodic heartbeat.

## Fix

Change `void → await` on all three create paths. Wake delivery failures now surface immediately, and the response only completes after the wake is queued.

No behavior change when wake succeeds; failure path is now observable rather than silent.

## Testing

- Create an issue via POST with `assigneeAgentId` set
- Agent should receive wake event without needing an @-mention
- Consistent with PATCH reassignment behavior


isak-ialogics:

Apologies for the noise.

PR #6936, opened 2026-05-29
feat(server): idempotency key dedup on POST /api/companies/:id/issues

### LINKED ISSUE #6529: Server-side dedup + run-id audit for POST /api/companies/{companyId}/issues (agent-driven duplicate child issue creation)
## Problem

`POST /api/companies/{companyId}/issues` (and the child variant `POST /api/issues/{id}/children`) currently accepts duplicate creations with no server-side dedup. Under realistic agent conditions — parallel heartbeats, stale wakes, retries after partial failures — this lets a single agent run silently create N near-identical child issues, wasting both tokens and reviewer attention.

The instruction-layer guard (`AGENTS.md: search before filing`) is intrinsically fragile and easy to skip; the right enforcement layer is the API.

This mirrors @paperclipai/paperclip#6265 (approvals dedup), #4780 (agent-hires idempotency), #2489 (enqueueWakeup idempotency), #5993 (dispatch_fingerprint dedup). Same shape, different endpoint.

## Evidence (one downstream incident)

Agent run `b1f96274` (heartbeat for issue DAI-459) called `POST /issues/{id}/children` three times within 18 seconds and got back three near-identical QA child issues (titles all `"QA验收: playwright-collect.ts 3项bug修复 [DAI-459]"`). The server happily created all three. Two were later cancelled manually.

| Created child | originFingerprint | createdByRunId |
| --- | --- | --- |
| DAI-460 | `"default"` | `null` |
| DAI-461 | `"default"` | `null` |
| DAI-462 | `"default"` | `null` |

`originFingerprint` is not client-writable, and `createdByRunId` is not even a column on `issues` despite the `X-Paperclip-Run-Id` header being read elsewhere — so post-hoc audit can't even tell which run filed which duplicate.

### 1. Accept a client-supplied `idempotencyKey` on POST issues

`POST /api/companies/{companyId}/issues` and `POST /api/issues/{id}/children` should accept an optional `idempotencyKey: string` in the body. Within a configurable window (suggest 24h), a second POST with the same `(companyId, idempotencyKey)` returns `409 duplicate { existingIssueId, status }` or replays the original `201`.

Convention agents would adopt: `idempotencyKey = "{parentId}:{purpose-slug}"` (e.g. `b1f96274:qa-validation`).

Pattern alignment: identical to the body-key approach proposed in #6265.

### 2. Backfill `createdByRunId` on issues from `X-Paperclip-Run-Id`

The server already extracts `actor.runId` from the header (see `getActorInfo(req)` at `server/src/routes/issues.ts:2568`) and uses it in `logActivity(...)`, but does not persist it onto the created issue row — and the `issues` table doesn't currently have a `created_by_run_id` column (peer tables like `issue_work_products`, `issue_tree_holds`, and `routines` do).

Proposed:

- Add `created_by_run_id uuid references heartbeat_runs(id) on delete set null` to the `issues` table (matches the pattern already used by `routines` and `issue_work_products`).
- In both POST handlers (`routes/issues.ts:2559` and `:2649`), pass `createdByRunId: actor.runId ?? null` through to `svc.create()` / `svc.createChild()`.
- Optional but consistent: do the same for `updated_by_run_id` on PATCH paths.

This is a pure audit win — without it, "which run created this child" is unanswerable except through fuzzy time-window joins on the activity log.

### 3. Soft dedup on `(runId, parentId, fuzzy title)` even when no idempotencyKey is supplied

As a defensive fallback for agents that haven't been updated to send a key, the create handler should at least check: same `actor.runId` + same `parentId` + title trigram-similar to an existing open child within the same run, within (say) 5 minutes. On match: return `409 duplicate_in_run { existingIssueId }`.

Cheap to implement — `issues_title_search_idx` is already a GIN trgm index on `title`, so this is a one-query lookup. Conservative threshold (≥0.7 similarity) to avoid false positives on legitimately-similar siblings.

## Why server-side, not just instructions

- Deterministic across all agents, runs, and adapters.
- Survives parallel heartbeats and race conditions instruction-layer guards cannot win.
- Gives downstream agent operators a stable contract: send `idempotencyKey`, get at-most-once semantics.

## Workaround we've shipped downstream

To buy time, we patched our company-level `AGENTS.md`:

> Before filing a child issue, agents must call `GET /issues?parentId=X` and reject the create if a matching open child already exists.

Bootstrapped into 14 active agents. Catches ~80% of the easy cases but unreliable under parallelism — exactly the case the API enforcement is needed for.

## Out of scope

- Cross-`companyId` dedup (not needed; companies are tenant boundaries).
- Stronger title canonicalization beyond trigram fuzzy match.
- Retroactive cleanup of historically-duplicated issues.

## Why an issue and not a PR

Filing this as an issue (not a PR) because the design touches API contract (key format, dedup windows, fuzzy-match thresholds, schema column name) where maintainer judgment is the bottleneck, not implementation. Happy to provide additional repro data, validate a candidate fix on our downstream fork, or contribute targeted patches once a direction is chosen.

## Acceptance test ideas

- `curl POST /issues` twice with the same `idempotencyKey` within 24h → second returns `409` with the first issue's id.
- POST issue → returned record has `createdByRunId` matching the `X-Paperclip-Run-Id` header sent on the request.
- POST `/issues/{id}/children` twice with same `parentId` and same `title` within the same run, no key → second returns `409 duplicate_in_run`.

isak-ialogics:
## Summary

- Adds optional `idempotencyKey` field to `POST /api/companies/:companyId/issues`
- If the same key is submitted for the same company within 5 minutes, returns the existing issue (HTTP 200) instead of creating a duplicate
- Adds `idempotency_key` column and unique partial index to the `issues` table (migration 0094)
- Adds `idempotencyKey` to `createIssueBaseSchema` in shared validators

## Test plan

- [ ] POST with a fresh `idempotencyKey` → creates issue, returns 201
- [ ] POST with the same `idempotencyKey` within 5 minutes → returns existing issue, 200, no duplicate row
- [ ] POST with the same `idempotencyKey` after 5 minutes → creates a new issue, 201
- [ ] POST without `idempotencyKey` → normal create, unaffected

Fixes https://github.com/paperclipai/paperclip/issues/6529


isak-ialogics:

Apologies for the noise.

PR #6940, opened 2026-05-29
fix(cli): move scaffolder CLI entry to bin.ts — stop gate firing on bundled invocations

### LINKED ISSUE #6521: v2026.517.0: scaffolder runs on every CLI invocation — creates spurious dir, then crashes on missing packages/shared
## Summary

In `paperclipai@2026.517.0`, `dist/index.js` bundles the `create-paperclip-plugin` scaffolder script alongside the main commander CLI. The scaffolder is gated only by:

```js
if (import.meta.url === `file://${process.argv[1]}`) {
  runCli();
}
```

When `paperclipai` is invoked via `npm exec paperclipai run` (which is what `npx paperclipai run` and many systemd units resolve to), Node receives the **real** `.pnpm/...` path as `argv[1]` (the `.bin/..` segment is normalized away by Node). The gate matches, `runCli()` fires *before* `main()`, takes `process.argv[2]` as the plugin name, and calls `scaffoldPluginProject({ outputDir: <cwd>/<argv[2]>, pluginName: <argv[2]> })`.

Result: every `npm exec paperclipai <subcommand>` invocation tries to scaffold a plugin named `<subcommand>` in CWD. For `paperclipai run` this:

1. Creates `<cwd>/run` directory.
2. Crashes inside `packLocalPackage` with `Error: Package package.json not found at <node_modules>/.pnpm/packages/shared/package.json` (the function resolves the monorepo path `packages/shared/package.json` which only exists in the source tree, not in a published install).
3. On systemd restart, the now-existing `<cwd>/run` makes `scaffoldPluginProject` throw `Error: Directory already exists: <cwd>/run` instead — putting the service into a permanent crash loop.

## Affected versions

- **Broken:** `2026.517.0` (reproduced)
- **Last known good:** `2026.512.0` (the scaffolder block is not in the bundle)

## Reproduction

```bash
mkdir /tmp/p-bug && cd /tmp/p-bug
pnpm init >/dev/null
pnpm add paperclipai@2026.517.0
# Real-path invocation reproduces the gate match:
node /tmp/p-bug/node_modules/.pnpm/paperclipai@2026.517.0_*/node_modules/paperclipai/dist/index.js run
# → creates /tmp/p-bug/run, then errors out
```

Or with `npm exec` (which is the pnpm/systemd-style invocation that resolves to the same real path):

```bash
cd /tmp/p-bug
/usr/bin/npx paperclipai run --data-dir /tmp/p-bug/data
# → same failure when CWD allows mkdir
```

The `.bin/` shim invocation (`./node_modules/.bin/paperclipai run`) does NOT hit this gate because the shim's `node \"\$basedir/../paperclipai/dist/index.js\"` traverses the `node_modules/paperclipai → .pnpm/...` symlink, so `argv[1]` ends up with the *symlink* path while `import.meta.url` has the *resolved* real path — they don't match and the scaffolder stays dormant. That is why some installs / dev flows don't see it.

## Evidence in the published bundle

(`paperclipai@2026.517.0` `dist/index.js`, single-file build)

- `~22849` — `function runCli() { const pluginName = process.argv[2]; ...` (the scaffolder entrypoint)
- `~22871` — `if (import.meta.url === \`file://\${process.argv[1]}\`) { runCli(); }` (the always-evaluated gate)
- `~22301` — `if (fs17.existsSync(outputDir)) { throw new Error(\`Directory already exists: \${outputDir}\`); }`
- `~22273` — `throw new Error(\`Package package.json not found at \${packageJsonPath}\`);` inside `packLocalPackage`
- `~22264` — `getLocalSharedPackagePath` returns `path.resolve(getRepoRootFromSdkPath(sdkPath), \"packages\", \"shared\")` — a monorepo path

## Suggested fix

Either:

1. **Don't bundle the `create-paperclip-plugin` standalone script into the main `paperclipai` bundle.** Ship it as a separate file (`dist/create-paperclip-plugin.js`) referenced by a separate bin entry in `package.json`. The current `package.json` has only `\"paperclipai\": \"./dist/index.js\"`, so the second-bin code has no legitimate way to fire under the current packaging.
2. **Tighten the gate** to require an explicit signal — e.g., check `process.argv[1]` ends with `create-paperclip-plugin` (the script's intended bin name), or guard on a sentinel env var.

Option 1 is cleaner.

## Operational impact for downstream tenants

- `systemctl restart paperclip` on the upgraded host crash-loops indefinitely with `Restart=on-failure`.
- The leftover `<WorkingDirectory>/run` directory must be removed manually before any further attempt — even after a rollback to `2026.512.0` if a user wants to retry — otherwise the next 517 try escalates from \"`packages/shared` missing\" to \"directory exists\".
- No data loss observed (paperclip state under `~/.paperclip/instances/default/` is untouched), but the failed upgrade is **not safely recoverable in-place** for typical `npm exec` / systemd unit setups.

---

Reported by a downstream tenant running paperclipai under systemd via `/usr/bin/npx paperclipai run`. Happy to provide systemd journal excerpts, additional bundle line ranges, or a recorded repro.

isak-ialogics:
## Problem

Closes #6521.

Since `v2026.517.0`, invoking the CLI via `npm exec paperclipai <subcommand>` (which resolves symlinks) causes `dist/index.js` to scaffold a plugin named `<subcommand>` in the current directory, then crash. The root cause is that `packages/plugins/create-paperclip-plugin/src/index.ts` contained a top-level guard:

```js
if (import.meta.url === `file://${process.argv[1]}`) {
  runCli();
}
```

When esbuild bundles `index.ts` into `dist/index.js`, it rewrites `import.meta.url` to the bundle's file URL. Any real-path invocation of the bundle then satisfies the equality check and fires `runCli()` before `main()` can run.

## Fix

- Extract all CLI-specific code (`runCli`, `parseArg`, the gate) from `index.ts` into a new `src/bin.ts` entry point.
- `src/index.ts` is now a pure library — no top-level side-effecting code, safe to bundle.
- Use `pathToFileURL(entrypoint).href` in the `bin.ts` gate (handles Windows paths, avoids the fragile template-string comparison).
- Update `package.json` `bin` and `publishConfig.bin` to point to `dist/bin.js` instead of `dist/index.js`.

## Test plan

- [ ] `paperclipai run`, `paperclipai secrets list`, and other regular invocations no longer create directories or crash.
- [ ] `npx create-paperclip-plugin my-plugin` scaffolds correctly (uses `bin.js`).
- [ ] `paperclipai plugin init my-plugin` still works (imports from `index.js`, untouched).
- [ ] TypeScript compiles without errors (`pnpm --filter @paperclipai/create-paperclip-plugin typecheck`).


isak-ialogics:

Apologies for the noise.

PR #6944, opened 2026-05-29
fix(heartbeat): guard executeRun against paused agent; add process adapter pause check

### LINKED ISSUE #6581: Heartbeat local child runs can leak processes and bypass paused agents
## Summary

Local child-process heartbeat runs have two lifecycle gaps:

1. Critical silent-run detection creates/elevates review work but does not terminate the still-running OS process, so a stalled local adapter can leave an orphaned child process alive after DB bookkeeping is recovered.
2. A heartbeat run that has already been claimed can reach `executeRun` after an agent is paused; `executeRun` can then flip the agent back to `running` and invoke the adapter because there is no final status guard immediately before adapter invocation.

## Evidence from local instance

- `scanSilentActiveRuns` selects quiet `running` runs and creates/elevates evaluation work but does not call the run cancellation/process termination path.
- The critical source-issue block explicitly says it blocks the source issue without cancelling the active process.
- `cancelRunInternal` and `cancelActiveForAgent` already know how to terminate by persisted pid/processGroupId, but the silent-run watchdog does not call that primitive.
- `executeRun` accepts a queued/running run, claims it, loads the agent once, then later writes `agents.status = 'running'` and calls `adapter.execute` without re-reading the current agent/run state immediately before invocation.
- Generic `PATCH /agents/:id` can set `status: paused` through the normal update route without routing through the pause endpoint that calls `cancelActiveForAgent`.

## Proposed fix

- For critical output silence on tracked local child-process adapters, terminate the persisted pid/processGroupId through the same safe path used by run cancellation, mark the run with an explicit `silent_run_terminated` failure/cancel code, then create/attach the audit evaluation issue.
- Add a final execution guard before `agents.status = 'running'` and `adapter.execute`: re-read the run and agent; if the agent is paused/terminated/pending approval or the run is no longer running, abort/release cleanly and do not invoke the adapter.
- Make generic status transitions to `paused` either rejected or routed through the same cancellation semantics as `POST /agents/:id/pause`.
- Add regression tests for critical silent-run termination and pause-after-claim-before-adapter.

## Local provenance

This issue was filed before a board-approved local patch on a private Paperclip instance for CRO-5293. The local patch should be removed/restamped once the upstream fix lands.

isak-ialogics:
## Summary

Fixes #6581 — local child heartbeat runs can leak processes and bypass paused agent state.

### Changes

**`server/src/services/heartbeat.ts`**
- Added pre-adapter guard in `executeRun`: immediately before `adapter.execute()`, re-reads the current agent status from the DB. If the agent is `paused`, `terminated`, or `pending_approval` at the point of invocation, the run is aborted cleanly without invoking the adapter. This closes the race window where an agent could be paused after the initial check but before the adapter fires.

**`server/src/adapters/process/execute.ts`**
- Added pause check at the top of the process adapter's `execute()`: throws immediately if `agent.status` is `paused`, `terminated`, or `pending_approval`, preventing child process spawning for paused agents.

### Acceptance criteria

- ✅ Child heartbeat processes are not spawned when the agent is paused
- ✅ Paused agents do not run any heartbeats (double-checked at adapter entry and immediately before invocation)


isak-ialogics:

Apologies for the noise.

PR #4516, opened 2026-04-25
fix(comments): deduplicate agent-authored comments via run+body idempotency key

supertaz:
## Thinking Path

> - Paperclip agents can call `addComment` on an issue as part of their work
> - In a tight loop, an agent may call `addComment` multiple times with identical body text from the same run in rapid succession
> - Previously each call produced a new row, resulting in visible comment spam on the issue
> - There was no deduplication: the route handler always inserted a new row unconditionally
> - This PR computes a SHA-256 idempotency key per (runId, body) and enforces uniqueness at the database level via a unique partial index
> - The `INSERT ... ON CONFLICT DO NOTHING ... RETURNING` pattern makes dedup atomic with the insert
> - The benefit is that duplicate comments from the same run are silently collapsed into the original, with no visible spam

## What Changed

- `packages/db/src/migrations/0071_issue_comment_idempotency.sql` — adds `idempotency_key` column + unique partial index on `idempotency_key WHERE NOT NULL`
- `packages/db/src/migrations/meta/_journal.json` — journal entry
- `packages/db/src/schema/issue_comments.ts` — `idempotencyKey` column + `uniqueIndex`
- `server/src/services/issues.ts` — dedup logic in `addComment`: computes `idempotencyKey = SHA-256(runId + ":" + body)` for run-authored comments; uses `INSERT ... ON CONFLICT DO NOTHING ... RETURNING`; falls back to select-by-key on conflict

## Summary

An agent running in a tight loop can call `addComment` multiple times with identical body text from the same run in rapid succession. Previously each call produced a new row, resulting in visible comment spam on the issue.

This PR prevents duplicates atomically:

1. Computes `idempotencyKey = SHA-256(runId + ":" + body)` for any comment created by a run (`actor.runId` set).
2. Stores the key in a new nullable `idempotency_key` column on `issue_comments` (migration 0071).
3. Adds a **unique partial index** on `idempotency_key WHERE NOT NULL`, enforcing the constraint at the database level.
4. Uses `INSERT ... ON CONFLICT DO NOTHING ... RETURNING`. If the insert conflicts, selects the existing row by key and returns it — so callers receive the original comment transparently.

User-authored comments (no `runId`) are completely unaffected — `idempotencyKey` is `null` and the unique index is partial, so no conflict path exists for them.

## Changes

- `packages/db/src/migrations/0071_issue_comment_idempotency.sql` — adds `idempotency_key` column + unique partial index
- `packages/db/src/migrations/meta/_journal.json` — journal entry
[...]

PR #6962, opened 2026-05-29
SLI-126: server guard — If-Match + Idempotency-Key on comment writes

lyndsaykerwin:
## Summary

Implements the SLI-110 plan for the SLI-109 design (CEO-accepted, revision 3). Adds a server-side optimistic-concurrency guard on comment writes so agents can no longer post over comments they never saw (the SLI-59 failure mode), plus a bundled `Idempotency-Key` cache so the natural 409 retry path is safe.

- Refs: SLI-126 (this work), SLI-110 (plan), SLI-109 (design), SLI-59 (origin bug).
- Out of scope per the SLI-126 wake comment: removing the skill-layer Step 6 diff rule (bridge stays during the canary).

## What changed

1. **Freshness guard** on `POST /issues/:id/comments`, `PATCH /issues/:id` (when carrying `comment`), and `POST /issues/:id/interactions` plus its `accept` / `reject` / `respond` sub-routes. Compares the request's `If-Match` (header preferred, body `ifMatch` fallback) against `svc.getCommentCursor(issueId).latestCommentId`. On mismatch → `409 Conflict` with the documented `{ error: "stale_comment_cursor", expected, received, since[], retryHint }` body.
2. **System bypass header** `X-Paperclip-System-Comment: 1` honored only when `req.actor.source === "local_implicit"` (the harness/server itself). Agents and remote board keys get the header silently ignored.
3. **Idempotency-Key replay cache** (5-minute TTL) on the comment writes and on PATCH with a comment. Scope: `(endpoint, actor, key) → {status, body}`. A retry returns the original 201/200 response without creating a duplicate comment.
4. **Wake-payload standardization** in `server/src/services/heartbeat.ts`: `latestCommentId` is now derived from the issue's actual current latest comment (DB lookup), surfaced at the top level of every wake payload, `null` on empty threads. The `renderPaperclipWakePrompt` template now prints `null` instead of the literal string `unknown` when empty.
5. **New env var `PAPERCLIP_LATEST_COMMENT_ID`** wired through `readPaperclipLatestCommentIdFromContext` in `@paperclipai/adapter-utils/server-utils` and injected by all 9 adapters (claude-local, codex-local, cursor-local, cursor-cloud, gemini-local, grok-local, opencode-local, pi-local, acpx-local).
6. **Telemetry**: structured `stale_comment_cursor` event via `logger.info` with `agentId`, `runId`, `issueId`, `staleId`, `currentId`, `endpoint`. Easy to surface in `get_logs` / dashboards for SLI-127's verification step.
7. **Agent skill update**: `skills/paperclip/SKILL.md` documents the auto-attach (`If-Match: $PAPERCLIP_LATEST_COMMENT_ID`), the 5-minute idempotency cache, and the 409 retry semantics (refetch → reconcile → retry once → bypass-then-exit on second 409). `scripts/paperclip-issue-update.sh` now auto-attaches both headers whenever it carries a comment.
8. **Tests**:
   - Unit: `server/src/services/comment-freshness.test.ts` — 22 cases covering header/body parsing, ETag-envelope stripping, idempotency cache TTL + scoping, system bypass actor gating, all `evaluateCommentFreshness` outcomes.
   - Route: `server/src/__tests__/issue-comment-freshness-routes.test.ts` — 8 cases covering fresh accepted, stale 409 with `since[]`, harness bypass accepted, non-harness bypass-ignored, idempotency replay, the SLI-59 race repro, and a guard on POST `/interactions`.

## Test plan

- [x] `pnpm --filter @paperclipai/server exec vitest run src/services/comment-freshness.test.ts` (22/22 pass)
- [x] `pnpm --filter @paperclipai/server exec vitest run src/__tests__/issue-comment-freshness-routes.test.ts` (8/8 pass)
- [x] Regression sweep: `issue-comment-reopen-routes`, `issue-comment-cancel-routes`, `issue-thread-interaction-routes`, `issue-update-comment-wakeup-routes`, `issue-telemetry-routes` (39 + 14 tests, all green)
- [x] `pnpm --filter @paperclipai/server exec tsc --noEmit` — no new errors in any changed file (pre-existing errors in `routes/plugins.ts`, `services/plugin-*.ts`, and `adapters/registry.ts` are unaffected)
- [x] Adapter typechecks across all 9 adapters — clean

## Design hard-constraint observed

Per SLI-126's hard constraint, this PR does **not** diverge from the SLI-109 accepted design on any item where the design was clear. One open ambiguity: design rev. 3 item 7 ("the prompt rule is removed in the same change") vs. the SLI-126 wake comment ("Out of scope: Removing skill-layer Step 6 diff rule (bridge stays)"). The wake instruction is followed — the diff rule is left in place — to honor the routing handoff and the SLI-110 plan's explicit bridge bullet. Flagging for the reviewer in case the design override needs to be folded in before merge.


aduduman031084:
This PR is the closest existing path for a scanner outbox that must recover safely after process death and response loss. One contract gap prevents that use case from relying on it as written: the five-minute in-memory TTL does not survive server restart or a delayed scanner replay, so the same key can create a second comment after eviction.

Would you be open to making the `(endpoint, actor, Idempotency-Key)` binding durable (DB-backed, with request-body conflict detection) and returning the original persisted comment on every replay? The minimum receipt fields needed to fail closed are the original comment `id`, `issueId`, and exact `body`; a replay indicator is useful but not required if the original response is stable. The key must also be accepted via the documented `Idempotency-Key` header, not copied into the comment body.

I found #4516 while checking duplicates, but its `runId + body` scope does not cover long-lived task-bridge API actors and does not let the client choose a stable occurrence key, so it does not close this case.

PR #6970, opened 2026-05-29
perf(adapter-utils): reuse one SSH/TCP session per adapter run via ControlMaster (Windows-gated)

dennisreichenberg:
## Summary

`createSshAuthArgs` emits ControlMaster on non-Windows platforms to collapse adapter SSH calls onto one TCP+SSH session. On Windows, ControlMaster is gated off (Windows OpenSSH 9.5p2 ships a broken implementation -- see REI-510 below) and a small FIFO dial-pacing queue holds Windows runs under the per-IP rate limit that originally motivated the patch.

```ts
if (process.platform !== "win32") {
  sshArgs.push(
    "-o", "ControlMaster=auto",
    "-o", `ControlPath=${path.join(os.tmpdir(), "paperclip-ssh-cm-%C")}`,
    "-o", "ControlPersist=60s",
  );
}
```

Single point of change for the multiplexing flags: `createSshAuthArgs` is the lone entry point shared by `runSshCommand`, `buildSshSpawnTarget`, and `syncDirectoryToSsh`, so the gate covers every call site with no downstream edits. The dial-pacing queue is wrapped around the five ssh-dial sites (`runSshCommand`, `streamLocalFileToSsh`, `streamSshToLocalFile`, `syncDirectoryToSsh`, `syncDirectoryFromSsh`); `buildSshSpawnTarget` is intentionally not wrapped -- it returns args for the long-lived runtime spawn whose lifetime is owned by the caller.

## Why (perf framing)

Today every adapter SSH call opens a fresh TCP + SSH session. A typical adapter run fires 8-10 sequential `runSshCommand` / `syncDirectoryToSsh` calls (`mkdir`, workspace bootstrap, command exec, log fetch, snapshot, etc.). That's 8-10 full handshakes -- TCP SYN/SYN-ACK/ACK plus the SSH key exchange, host-key check, and pubkey auth -- for what is logically one session against one host.

With `ControlMaster=auto` + `ControlPersist=60s`, the first call establishes a master connection cached at the `%C`-hashed `ControlPath`, and the next 9 multiplex over it. End result on Linux/macOS: **one TCP+SSH handshake per adapter run instead of 8-10**, with the master torn down 60s after the last call.

`%C` hashes host/port/user, so concurrent runs targeting different environments don't collide on the control-socket path.

## Why (secondary: rate-limit hardening)

On hardened sshd hosts with per-IP connection rate limits (the motivating case for us was an IONOS-style provider), the 6th sequential dial in a tight burst gets TCP-blackholed by the rate limiter. The client `ConnectTimeout=10` then surfaces as a generic connect timeout -- confusing to diagnose because earlier calls in the same run succeeded. Collapsing 8-10 connections to 1 keeps adapter runs comfortably below any reasonable per-IP threshold.

Root-cause reproduction evidence: REI-497 comment `728875d0-f850-4ea2-b345-bb4404eabd43`.

## Windows: REI-510 follow-up

The first revision of this PR enabled ControlMaster unconditionally. That deterministically broke the adapter `test-environment` probe on Windows hosts: every Windows OpenSSH 9.5p2 dial with `ControlMaster=auto` returns

```
getsockname failed: Not a socket
Read from remote host <host>: Unknown error
```

regardless of `ControlPath` slash direction or whether `ControlPersist` is set. Reproduces with bare `ssh.exe`, not Paperclip-specific -- it's a known break in the OpenSSH-for-Windows AF_UNIX-on-Windows implementation. `ControlMaster=no` is the only working configuration on win32.

This revision:

1. **Gates ControlMaster on `process.platform !== "win32"`.** Linux/macOS keep the connection-reuse win; Windows takes the no-multiplexing fallback.
2. **Adds a Windows-only in-process dial-pacing queue.** Module-private FIFO with concurrency=1 and a 300ms post-completion delay, wrapped around every ssh-dial site. Without multiplexing on Windows we would re-trip the rate limit; the queue keeps the burst under threshold (measured ~5 dials per 2s on the IONOS VPS). Exposed as a factory (`_createSshDialPacingQueueForTests`) so tests can build isolated queues with deterministic timing. On non-Windows the queue short-circuits to passthrough so the pacing cost is zero where ControlMaster does the job.

Verified live on Windows 11 Pro (`OpenSSH_for_Windows_9.5p2, LibreSSL 3.8.2`): the adapter `test-environment` probe against the original failing VPS returns **4/4 green** (`claude_environment_target`, `claude_cwd_valid`, `claude_command_resolvable`, `claude_hello_probe_passed`).

## Tests

1. **Unit (`ssh-fixture.test.ts`):** assert `buildSshSpawnTarget` returns args containing `-o ControlMaster=auto`, `-o ControlPath=...paperclip-ssh-cm-%C`, and `-o ControlPersist=60s` on non-Windows -- and assert the same flags are absent on win32. With and without a privateKey.
2. **Unit (`ssh-fixture.test.ts`):** FIFO + pacing on Windows -- three concurrently-submitted tasks start in submission order and each subsequent task starts no earlier than `previous_end + pacingMs - jitter`. Passthrough on non-Windows -- at least one tick observes >1 active operations. Reject-release -- a rejected queued task still releases the chain so subsequent dials proceed (1s race-against-stuck guard).
3. **Integration (`ssh-fixture.test.ts`):** spins up the existing sshd env-lab fixture (already used elsewhere in this file), fires **10 sequential `runSshCommand` calls**, and asserts the delta of `Connection from ` lines in sshd's `LogLevel VERBOSE` log is exactly **1**. Without ControlMaster the delta would be 10. Skips automatically on hosts without `sshd` (Windows, minimal CI images) via the existing `startSshEnvLabFixtureOrSkip` helper.

Local `pnpm --filter @paperclipai/adapter-utils typecheck` is clean. `vitest run packages/adapter-utils/src/ssh-fixture.test.ts` reports **17/17 passing** on the Windows dev host (fixture-backed sshd tests skip there).

## Test plan

- [ ] CI typecheck + tests pass on Linux runner (where sshd is available, the integration tests execute and assert ControlMaster connection reuse)
- [x] Manual: `test-environment` probe on Windows host returns 4/4 green against the original failing VPS (REI-510 verification step 1)
- [ ] Manual: confirm `ssh -O check -o ControlPath=...` against an adapter target reports the master alive after a run on Linux/macOS

### DIFF
--- a/packages/adapter-utils/src/ssh-fixture.test.ts
+++ b/packages/adapter-utils/src/ssh-fixture.test.ts
@@ -4,6 +4,7 @@ import os from "node:os";
 import path from "node:path";
 import { afterEach, describe, expect, it } from "vitest";
 import {
+  _createSshDialPacingQueueForTests,
   buildSshSpawnTarget,
   buildSshEnvLabFixtureConfig,
   getSshEnvLabSupport,
@@ -166,7 +167,7 @@ describe("ssh env-lab fixture", () => {
     ).rejects.toThrow("Invalid SSH environment variable key: BAD KEY");
   });
 
-  it("emits ControlMaster=auto, ControlPath, and ControlPersist on every SSH invocation", async () => {
+  it("emits ControlMaster=auto, ControlPath, and ControlPersist on every SSH invocation (non-Windows only; REI-510 gates it on win32)", async () => {
     const target = await buildSshSpawnTarget({
       spec: {
         host: "ssh.example.test",
@@ -190,6 +191,15 @@ describe("ssh env-lab fixture", () => {
         }
         return -1;
       };
+      if (process.platform === "win32") {
+        // Windows OpenSSH 9.5p2 ControlMaster is broken (REI-510). The fix
+        // gates the multiplexing flags off on win32 and falls back to the
+        // dial-pacing queue to stay under the IONOS rate limit instead.
+        expect(indexOfOption("ControlMaster=auto")).toBe(-1);
+        expect(args.findIndex((entry) => entry.startsWith("ControlPath="))).toBe(-1);
+        expect(indexOfOption("ControlPersist=60s")).toBe(-1);
+        return;
+      }
       expect(indexOfOption("ControlMaster=auto")).toBeGreaterThanOrEqual(0);
       const controlPathIdx = args.findIndex(
         (entry, idx) => idx > 0 && args[idx - 1] === "-o" && entry.startsWith("ControlPath="),
@@ -250,7 +260,7 @@ describe("ssh env-lab fixture", () => {
     SSH_FIXTURE_TEST_TIMEOUT_MS,
   );
 
-  it("emits ControlMaster args even without a privateKey", async () => {
+  it("emits ControlMaster args even without a privateKey (non-Windows only)", async () => {
     const target = await buildSshSpawnTarget({
       spec: {
         host: "ssh.example.test",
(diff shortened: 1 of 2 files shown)


dennisreichenberg:
Extended this PR with the REI-510 follow-up: the original revision broke Windows OpenSSH 9.5p2 (`getsockname failed: Not a socket` on every dial), so ControlMaster is now gated on `process.platform !== "win32"` and a Windows-only FIFO dial-pacing queue handles the rate-limit case where ControlMaster used to.

Pushed as `d3e9dd5` on top of the existing branch. Title also tweaked to flag the platform gating.

Live verification on Windows 11 Pro (`OpenSSH_for_Windows_9.5p2`):
- adapter `test-environment` probe against the original failing IONOS VPS: **4/4 green**
- 17/17 unit tests pass on the Windows host

The PR body now has a dedicated "Windows: REI-510 follow-up" section covering the diagnosis, gating logic, queue mechanics, and verification. Test plan checkbox for the Windows manual probe is checked.

PR #6971, opened 2026-05-29
feat(mcp-server): expose plugin-registered tools to LLM via MCP

### LINKED ISSUE #6818: Plugin-registered tools are not being accessible by the llm
Tools registered by a plugin via `ctx.tools.register` and properly returned
by `GET /api/plugins/tools` are not callable from agents running on the
`claude_local` adapter. The plugin tool registry knows about them; the
spawned Claude Code instance does not. The Paperclip spec says "by default,
plugin tools are available to all agents" that holds for the dispatcher
side, but the `claude_local` adapter has no bridge from the plugin tool
registry into the Claude Code child process's MCP/tool universe, so the
agent's `ToolSearch` returns zero hits and the tools are effectively
invisible.

I'm building my first Paperclip plugin (a small Google Maps connector 
that exposes geocode, search_places, and get_place_details as agent tools)
 and I'm at the testing stage. The plugin installs cleanly, the config and Test 
Connection both pass, and paperclipai plugin inspect shows it as ready with 
all three tools properly registered. The trouble starts when I actually try to use 
it through a Paperclip agent: I opened an issue assigned to one of my claude_local 
agents asking it to use the tools, and the agent came back saying the tools aren't 
accessible to it. So the registration side looks healthy end-to-end, but something 
isn't bridging the tools across to the LLM the agent is driving. Filing this in case it's 
a known gap or I'm missing a wiring step.

isak-ialogics:
## Summary

Fixes #6818. Plugin tools registered via the plugin tool dispatcher were only accessible through the REST API, not through the MCP server that Claude CLI loads at startup — so the LLM could never see or invoke them.

**Part 1 — MCP server** (`packages/mcp-server/src/index.ts`):
- `createPaperclipMcpServer` is now `async`
- Fetches plugin tools from `GET /api/plugins/tools` at server startup
- Converts each tool's JSON Schema to a Zod shape for MCP registration
- Execution routes through `POST /api/plugins/tools/execute`
- Graceful: if the fetch fails, falls back to zero plugin tools (core tools still register fine)

**Part 2 — claude_local adapter** (`packages/adapters/claude-local/src/server/execute.ts`):
- Before spawning Claude CLI, writes a per-run temp MCP config JSON to `os.tmpdir()`
- Passes `--mcp-config <path>` to Claude CLI args
- Cleans up the temp file in the `finally` block
- Skipped for remote execution targets (the MCP server wouldn't be reachable there)
- Resolves MCP server `dist/stdio.js` via `createRequire` from the `@paperclipai/mcp-server` package.json, so it works in both monorepo dev and published install

**Dependency**: adds `@paperclipai/mcp-server: workspace:*` to `@paperclipai/adapter-claude-local`.

## Test plan

- [ ] Start a local Paperclip instance with a plugin that registers a tool
- [ ] Verify `GET /api/plugins/tools` returns the tool descriptor
- [ ] Run an agent heartbeat; confirm Claude CLI receives `--mcp-config` and the plugin tool appears in Claude's tool list
- [ ] Invoke the tool from Claude; confirm it routes through `POST /api/plugins/tools/execute`

### DIFF
--- a/packages/adapters/claude-local/package.json
+++ b/packages/adapters/claude-local/package.json
@@ -54,6 +54,7 @@
   },
   "dependencies": {
     "@paperclipai/adapter-utils": "workspace:*",
+    "@paperclipai/mcp-server": "workspace:*",
     "picocolors": "^1.1.1"
   },
   "devDependencies": {
--- a/packages/adapters/claude-local/src/server/execute.ts
+++ b/packages/adapters/claude-local/src/server/execute.ts
@@ -1,4 +1,6 @@
 import fs from "node:fs/promises";
+import { createRequire } from "node:module";
+import os from "node:os";
 import path from "node:path";
 import { fileURLToPath } from "node:url";
 import type { AdapterExecutionContext, AdapterExecutionResult } from "@paperclipai/adapter-utils";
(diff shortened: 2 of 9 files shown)


isak-ialogics:

PR #6973, opened 2026-05-29
fix(heartbeat): skip timer enqueue when agent already has an active timer run

### LINKED ISSUE #6955: bug: heartbeat timer queues unlimited runs when heartbeat duration exceeds interval — maxConcurrentRuns should default to 1
## Environment
- Paperclip version: 2026.525.0
- Server: 4 vCPU, 16GB RAM
- Agents: 11 (4 timer, 7 task-activated)

## Bug

When an agent's heartbeat takes longer than its `intervalSec`, Paperclip queues a new heartbeat run every interval tick without checking if a previous run is still active. This causes unbounded queue growth and server overload.

In our case:
- CTO agent on Claude Opus with `intervalSec: 1200` (20 minutes)
- Single heartbeat took 130 minutes (heavy architectural work)
- Paperclip queued 116 additional heartbeat runs during that time
- COO had 42 queued, CEO had 40, Lead Backend had 20
- Total: **235 queued heartbeat runs** piled up
- Server load hit 14x on 4 CPUs — dashboard completely unresponsive

## Expected Behavior

Paperclip should NOT queue a new heartbeat run if the agent already has an active (running) heartbeat. The `maxConcurrentRuns` setting exists but defaults to unlimited — it should default to 1.

## Current Workaround

Manually set `maxConcurrentRuns: 1` in every agent's runtime config:

```json
{"heartbeat": {"intervalSec": 5400, "maxConcurrentRuns": 1}}
```

And periodically cancel stacked runs:
```sql
UPDATE heartbeat_runs_2026_05 SET status = 'cancelled' WHERE status = 'queued';
```

## Suggested Fix

1. Default `maxConcurrentRuns` to 1 (not unlimited)
2. Before queuing a timer heartbeat, check if the agent already has a running heartbeat — skip if so
3. Add a warning log when a heartbeat exceeds its interval: "Agent X heartbeat took Y minutes, interval is Z minutes — consider increasing interval"
4. Consider auto-adjusting the interval when heartbeats consistently exceed it

## Impact

Without the workaround, the server becomes unusable within 1-2 hours of enabling agents. The queued runs consume database connections, CPU, and memory even though they accomplish nothing (the previous heartbeat already covered the same work).

This is the second time we've reported server overload from Paperclip's database query patterns (see #6947). Combined, these two issues make it difficult to run more than 5 agents on a 4-CPU server.

isak-ialogics:
## Summary

Fixes #6955. When an agent's heartbeat takes longer than its `intervalSec`, the scheduler kept queueing new runs every tick. This caused unbounded queue growth — the reporter observed 116 queued runs after a 130-minute heartbeat on a 20-minute interval.

**Root cause:** `tickTimers` checks `lastHeartbeatAt` to decide whether to fire. `lastHeartbeatAt` is only updated when a run *finishes* (in `finalizeAgentStatus`), so during a long-running heartbeat, every scheduler tick sees the elapsed time as exceeding the interval and tries to enqueue.

The existing coalescing logic in `enqueueWakeup` is supposed to deduplicate, but it has a race window in the no-issue path (not in a transaction) and is called too late — the wakeup request is inserted before the coalesce check completes.

**Fix:** In `tickTimers`, before calling `enqueueWakeup`, check whether the agent already has any queued or running timer-sourced runs. If so, skip this timer tick (increment `skipped` counter). This is a cheap indexed query on `(agentId, status, invocationSource)` and prevents stacking at the source.

## Impact

- Agents whose heartbeats exceed `intervalSec` will no longer accumulate a backlog of queued runs
- The next timer fires naturally once `lastHeartbeatAt` is updated after the current run completes
- No behavioral change for agents whose heartbeats finish before the next tick

## Test plan

- [ ] Start a local instance with an agent whose heartbeat intentionally runs longer than `intervalSec`
- [ ] Confirm no queued timer runs accumulate while the run is active
- [ ] Confirm the next timer run fires correctly once the previous finishes


isak-ialogics:

Apologies for the noise.

PR #6976, opened 2026-05-29
feat(cli): add --redact flag to agent get command

### LINKED ISSUE #6895: Add --redact flag to agent/company get + GET /api/agents/<id> to prevent secret leakage in inspection workflows
## Summary

`paperclipai agent get <id>` (with or without `--json`) returns the agent's full `adapterConfig` in plaintext, including any "Plain mode" secret values (`adapterConfig.env.<KEY>.value`) and any Paperclip API tokens that have been baked into `promptTemplate` strings. Similarly, the underlying `GET /api/agents/<id>` endpoint returns these in plaintext.

For deployments that use Plain mode (currently the only working option while #224 is unfixed) and the `hermes_local` adapter's recommended `promptTemplate` pattern (which embeds an agent API token so the agent can call back to Paperclip from inside its terminal tool), this means any user or LLM inspecting an agent's config inadvertently surfaces:

- API keys for third-party providers (OpenRouter, Anthropic, etc.)
- The agent's own Paperclip API token (with full agent-scope permissions on its company)

## Why this matters in practice

We hit this twice in two sessions while debugging Paperclip agent configuration in a Claude Code / LLM-driven workflow. Both times, the JSON output landed in a session transcript and required a full token rotation (regenerate OpenRouter key + regenerate the Welchman agent's Paperclip token + update every agent's `adapterConfig.env` + update every agent's `promptTemplate`). The rotation pass is ~25-30 minutes of mechanical work per incident.

The CLI and API have no obvious way to inspect agent shape without seeing secret values. So any future debugging session that needs to "see what fields exist in adapterConfig" or "compare adapter setups across agents" has to risk the same leak.

## Ask

Add a `--redact` flag (or make it the default with `--unsafe-show-secrets` for opt-in) on:

- `paperclipai agent get <id>` (CLI)
- `GET /api/agents/<id>` (API, gated by a request header or query param)
- `paperclipai company get <id>` (if it returns similar embedded secrets)

When `--redact` is set, any of the following should be replaced with placeholder strings like `<redacted:N-chars>`:

1. **`adapterConfig.env.<KEY>.value`** — when the envelope is `{"type": "plain", "value": "..."}`, redact the `value` while preserving `type`.
2. **`promptTemplate`** strings — substitute any matched secret patterns: `pcp_[a-f0-9]{32,}`, `sk-or-v\d+-[a-f0-9]{32,}`, `sk-ant-[A-Za-z0-9_-]{50,}`, `sk-proj-[A-Za-z0-9_-]{20,}`.
3. **Any field at a SECRET_FIELDS path** — `token`, `apiKey`, `secret`, etc.

Returning the character length in the placeholder (`<redacted:108-chars>`) preserves enough information for debugging (you know whether the key is empty, the wrong format, etc.) without leaking the value itself.

## Workaround we built

For our own protection while this is unfixed, we ship a Python wrapper at `scripts/paperclipai-safe` in our internal repo that intercepts `paperclipai` stdout, parses JSON, walks the structure, and applies the redactions described above before re-emitting. Source happy to share if useful as a reference.

## Related

- Issue #224 (broken inline secret references) — the reason Plain mode is the current working option; if #224 were fixed, secret references would not appear in agent config inspection at all and this issue would be less acute. But even then, `promptTemplate` strings with baked-in agent tokens would still leak.

isak-ialogics:
## Summary

- Adds `--redact` flag to `paperclipai agent get` command
- When set, routes to `/api/agents/:id/configuration` which applies server-side redaction of plain env binding values and sensitive config fields
- Without the flag, behavior is unchanged (uses `/api/agents/:id`)

## Usage

```
paperclipai agent get <agentId> --redact
```

## Test plan

- [ ] `paperclipai agent get <id>` still returns full unredacted config
- [ ] `paperclipai agent get <id> --redact` returns redacted output (plain env values masked)
- [ ] `paperclipai agent get <id> --redact --json` works for JSON output

Fixes #6895


isak-ialogics:

Apologies for the noise.

PR #6978, opened 2026-05-29
fix(claude-local): reset session on unreplayable extended-thinking blocks

NaCl5alt:
## Thinking Path

> - Paperclip orchestrates AI agents that run continuously as resumed `claude_local` sessions
> - `claude_local` adapter manages session persistence — resumed sessions re-send the prior conversation including assistant turns
> - Anthropic requires extended-thinking blocks in assistant turns to be replayed verbatim; the CLI re-serialises them and the API rejects mismatches with a deterministic `400`
> - When a run is interrupted mid-turn (e.g. signal, OOM), the persisted session id captures a turn that already contains thinking blocks
> - `execute.ts` had a fresh-session fallback only for `isClaudeUnknownSessionError`; the thinking-block 400 hit no fallback
> - Because the corrupted session id stays persisted, **every** subsequent recovery wake re-resumes the same session and hits the same 400 — a permanent failure loop
> - This PR detects the thinking-block 400 in `parse.ts` and mirrors the unknown-session recovery path in `execute.ts`: retry once with a fresh session and clear the stored id

## What Changed

- **`parse.ts`**: add `CLAUDE_THINKING_BLOCK_REUSE_RE` (word-boundary anchored) and `isClaudeThinkingBlockReuseError()` — scans the full `errorMessage` / `parsed.result` / `parsed.errors` / `stdout` / `stderr` haystack via `buildClaudeTransientHaystack`
- **`execute.ts`**: add a new fallback branch after `isClaudeUnknownSessionError`, gated on `sessionId` being set — logs, retries once with `runAttempt(null)`, and returns the fresh-retry result directly (no retry loop possible)
- **`parse.test.ts`**: positive cases for all four haystack fields; negative cases for unrelated errors and no-session-id guard

## Verification

- `vitest run packages/adapters/claude-local/src/server/parse.test.ts` → 11 passed / 0 failed
- `tsc --noEmit` on `@paperclipai/adapter-claude-local` → clean
- Manually confirmed: a session whose last turn carried thinking blocks now recovers to a fresh session instead of looping on the 400

## Risks

- **False-positive session reset**: detection regex (`\b(?:redacted_)?thinking\b…cannot\s+be\s+modified`) requires both a whole-word `thinking`/`redacted_thinking` match and the specific Anthropic suffix within 200 chars — false positives are unlikely. Worst case: a healthy session is discarded and the next run starts fresh, losing only prior context (no data loss, no infinite loop).
- **Fresh-session retry cost**: one extra Claude API call per affected wake; negligible given the alternative is permanent failure.
- **Scope**: gated on `sessionId` being set, so unaffected sessions and fresh runs are never touched.

## Model Used

- **Provider**: Anthropic
- **Model**: claude-sonnet-4-6 (claude_local adapter, Paperclip CEO agent)
- **Reasoning mode**: extended thinking enabled (the very mode that triggers the bug being fixed)
- **Context**: resumed session via Paperclip heartbeat

## Checklist

- [ ] If this change affects the UI, I have included before/after screenshots (N/A — server-only change)

### DIFF
--- a/packages/adapters/claude-local/src/server/execute.ts
+++ b/packages/adapters/claude-local/src/server/execute.ts
@@ -52,6 +52,7 @@ import {
   detectClaudeLoginRequired,
   extractClaudeRetryNotBefore,
   isClaudeMaxTurnsResult,
+  isClaudeThinkingBlockReuseError,
   isClaudeTransientUpstreamError,
   isClaudeUnknownSessionError,
 } from "./parse.js";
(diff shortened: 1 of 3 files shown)


NaCl5alt:
@greptileai review

NaCl5alt:
Closing this PR. The fix will be applied to our internal fork instead. Thanks for the consideration.

PR #6987, opened 2026-05-29
feat(issues): surface matchedField + snippet on issue search

pjhagents:
## Thinking Path

> - Paperclip orchestrates AI agents for zero-human companies, and `GET /api/companies/:companyId/issues?q=<term>` is the primary entry point for both agents and humans triaging work.
> - The issue list endpoint ranks matches across title, identifier, comments, and description, but the response does not say which field actually matched or what the surrounding text looks like.
> - That forces every caller — agent or UI — to re-grep the row to figure out why it ranked the way it did. For long descriptions and noisy comment threads this means either a second fetch or wrong attribution.
> - Surfacing the matched field, its rank, and a short sanitized snippet keeps callers honest about what hit, lets the UI highlight the result in one shot, and makes agent-side reasoning over search results much cheaper.
> - This pull request attaches `matchedField`, `matchedFieldRank`, and `matchSnippet` to each search result when `q` is provided, with `<mark>`-wrapped highlights and HTML-escaped context.
> - The benefit is a cosmetic / UX papercut closed for every Paperclip operator without any DB schema change, and zero impact on non-search calls.

## What Changed

- Added `matchedField` (`title` | `identifier` | `description` | `comments`), `matchedFieldRank` (1..4), and `matchSnippet` to each result of `GET /api/companies/:companyId/issues` when `q` is provided. The rank mirrors the existing search ranking already enforced by the SQL `CASE` expression (title > identifier > comments > description).
- Snippet generation lives in a new exported helper `buildIssueSearchSnippet` in `server/src/services/issues.ts`. It normalises whitespace, windows ~160 chars around the match, HTML-escapes everything, wraps the matched substring in `<mark>...</mark>`, and adds `…` ellipses when the snippet is truncated on either side.
- For comment hits, the first matching comment per issue is fetched via a new chunked helper `firstMatchingCommentBodiesForIssues` that reuses the existing `ILIKE … ESCAPE '\\'` pattern with `escapeLikePattern`. The fetch is scoped to the (already small) page of issue ids returned by the main search query.
- The new fields are only attached when `hasSearch` is true, so non-search list calls are byte-for-byte unchanged.
- Added four tests in `server/src/__tests__/issues-service.test.ts`:
  - Coverage for hits in title, identifier, description, and comments — including rank values and `<mark>` markers.
  - An XSS test that injects `<script>` in a title and verifies the snippet is HTML-escaped.
  - A regression test that confirms the new fields are absent when no `q` is provided.

## Verification

- `pnpm --filter @paperclipai/server exec vitest run src/__tests__/issues-service.test.ts` — 53 tests pass, including the new four.
- `pnpm --filter @paperclipai/server exec vitest run src/__tests__/company-search-service.test.ts src/__tests__/company-search-rate-limit-routes.test.ts` — both pass, confirming the company-search service (separate code path) is unaffected.
- Spot-checked the route handler in `server/src/routes/issues.ts`: it already does `res.json(result.map((issue) => ({ ...issue, … })))`, so the new fields propagate without any route change.
- Pre-existing typecheck errors in the repo all live under `@paperclipai/plugin-sdk` resolution; no new errors are introduced under `server/src/services/issues.ts` or its test file.

Sample shape, after the change:

```json
{
  "id": "…",
  "title": "Looking into the dockerfile race condition",
  "matchedField": "title",
  "matchedFieldRank": 1,
  "matchSnippet": "Looking into the <mark>dockerfile</mark> race condition"
}
```

## Risks

- Low risk. The new fields are additive and opt-in: they are only emitted when `q` is provided, so existing callers see no behavioural change.
- Extra query: when a result row's only match is in comments, one chunked `SELECT` is issued against `issue_comments` for that page of issues. The chunking reuses `ISSUE_LIST_RELATED_QUERY_CHUNK_SIZE`, so the worst case scales with the existing page limit. For very long comment bodies the query pulls the matching body; the JS-side window then caps the snippet source at ~320 chars before HTML-escaping. If this turns out to be too eager for large instances we can switch to a `DISTINCT ON … LIMIT 1` form per issue, but the current shape is consistent with the existing `commentSearchMatchIssueIds` lookup in the blocked-inbox path.
- The matched-field detection runs in JS against `title`, `identifier`, and the already-truncated description preview, mirroring the SQL `ILIKE` predicates. Because the SQL preview is byte-truncated to ~4800 bytes there is a narrow case where a match exists in the database description past that boundary but the snippet would not include it; in that case `matchedField` falls back to `comments` if a comment also matched, or otherwise resolves to `description` with a `null` snippet. This is a strict improvement over today, where the caller has no attribution at all.

## Model Used

- Claude — `claude-opus-4-7` (Opus 4.7, 1M context). Standard tool use; no extended thinking mode used for this change.

## Checklist

- [ ] If this change affects the UI, I have included before/after screenshots — n/a, server-only.
- [x] I have updated relevant documentation to reflect my changes — this is an additive API surface; no docs in-tree describe the existing search response yet, so nothing to update.

---

Source bug: [NET-89](/NET/issues/NET-89) (Network Admin instance).

### DIFF
--- a/server/src/__tests__/issues-service.test.ts
+++ b/server/src/__tests__/issues-service.test.ts
@@ -465,6 +465,137 @@ describeEmbeddedPostgres("issueService.list participantAgentId", () => {
     expect(result.map((issue) => issue.id)).toEqual([commentMatchId, descriptionMatchId]);
   });
 
+  it("attaches matchedField/matchedFieldRank/matchSnippet for hits in title, identifier, description and comments", async () => {
+    const companyId = randomUUID();
+    const titleHitId = randomUUID();
+    const identifierHitId = randomUUID();
+    const commentHitId = randomUUID();
+    const descriptionHitId = randomUUID();
+
+    await db.insert(companies).values({
+      id: companyId,
+      name: "Paperclip",
+      issuePrefix: `T${companyId.replace(/-/g, "").slice(0, 6).toUpperCase()}`,
+      requireBoardApprovalForNewAgents: false,
+    });
+
+    await db.insert(issues).values([
+      {
+        id: titleHitId,
+        companyId,
+        title: "Looking into the dockerfile race condition",
+        status: "todo",
+        priority: "medium",
+      },
+      {
+        id: identifierHitId,
+        companyId,
+        issueNumber: 4242,
+        identifier: "DOCKERFILE-4242",
+        title: "Some other thing",
+        status: "todo",
+        priority: "medium",
+      },
+      {
+        id: commentHitId,
+        companyId,
+        title: "Unrelated work",
+        status: "todo",
+        priority: "medium",
+      },
+      {
+        id: descriptionHitId,
+        companyId,
+        title: "Yet another item",
+        description: "Replace the legacy dockerfile pipeline before the release.",
+        status: "todo",
+        priority: "medium",
+      },
+    ]);
+
+    await db.insert(issueComments).values({
+      companyId,
+      issueId: commentHitId,
+      body: "We discussed the dockerfile fix last sprint.",
+    });
+
+    const results = await svc.list(companyId, { q: "dockerfile", limit: 10 });
+    const byId = new Map(results.map((issue) => [issue.id, issue]));
+
+    const titleHit = byId.get(titleHitId) as any;
+    expect(titleHit?.matchedField).toBe("title");
+    expect(titleHit?.matchedFieldRank).toBe(1);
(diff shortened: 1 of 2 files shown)


pjhagents:
Closing as superseded/stale. Issue search on master is now served by the dedicated `companySearchService` (server/src/services/company-search.ts), which already surfaces `matchedFields` plus snippet/extract results (and adds rate limiting). This PR added match attribution directly in server/src/services/issues.ts, which that service obsoletes — no rebase path remains. Closing rather than rebasing.

PR #6988, opened 2026-05-29
fix(server): include subfolders when local SKILL.md sits at import root

cassioln:
## Thinking Path

> - The Skills subsystem lets a company import third-party Claude-style skill bundles (markdown + supporting `references/`, `scripts/`, `assets/`) and materialize them for the runtime each agent uses
> - When importing a skill via the **Add Skill > Path** flow (or `POST /companies/:id/skills/import` with a local filesystem path) and the path points directly at a skill folder whose `SKILL.md` is at the root, the resulting `fileInventory` only contains `SKILL.md` — every subfolder is silently dropped
> - Because both the UI file viewer and `materializeRuntimeSkillFiles` iterate `fileInventory`, the agent literally cannot see or read any of those files at runtime, even though the skill bundle on disk is intact
> - This pull request narrows the bug to a single edge case in `readLocalSkillImports` — `path.posix.dirname("SKILL.md") === "."`, which then produced a `"./"` prefix that never matched the relative paths returned by `walkLocalFiles` — and adds regression tests for both the root and nested-skill cases
> - The benefit is that "point Add Skill at this folder" finally works the way users expect, matching the behavior already enjoyed by skills cloned from GitHub or imported via a parent directory containing multiple skills

## What Changed

- `server/src/services/company-skills.ts`: in `readLocalSkillImports`, treat `skillDir === "."` (SKILL.md at the import root) as "everything `walkLocalFiles` returned belongs to this skill", and compute the relative path accordingly. Other skill discovery paths are unchanged.
- `server/src/services/company-skills.ts`: export `readLocalSkillImports` so the existing `company-skills.test.ts` can exercise it directly (matches the pattern already used for `readLocalSkillImportFromDirectory`).
- `server/src/__tests__/company-skills.test.ts`: two new regression tests in a new `describe("readLocalSkillImports — directory walks", …)` block:
  - Importing a folder whose `SKILL.md` is at the root with `references/` and `scripts/` siblings now produces a 3-file `fileInventory` and `trustLevel === "scripts_executables"`.
  - Importing a parent that contains two nested skills still discovers both, with the correct per-skill `fileInventory` (no regression on the previously working path).

## Verification

- `cd server && pnpm exec vitest run src/__tests__/company-skills.test.ts` — 14/14 pass with the fix; the new "includes references/ and scripts/ when SKILL.md sits at the import root" test fails on master without the fix (it asserts the missing 2 files), proving the regression coverage.
- `cd server && pnpm exec vitest run src/__tests__/company-skills` — full company-skills suite (47 tests across 5 files) passes.
- Reproduced in a real Paperclip install before/after the fix:
  - Before: `POST /companies/:id/skills/import` with `source: "/path/to/skill-folder/"` (where `SKILL.md` is at the root and `references/`, `scripts/` exist) yields `fileInventory.length === 1`.
  - After: same call yields the full inventory and the agent can read each subfile.

## Risks

- Low risk. The change is local to `readLocalSkillImports`, gated entirely on `skillDir === "."` (the previously broken case). The nested-skill code path is unchanged in behavior, and the new test asserts that explicitly.
- `readLocalSkillImports` was previously unexported; exporting it expands the public surface of the module, but the test file is the only new caller and the function is documented by behavior in the test names.

## Model Used

- Provider: Anthropic via Amazon Bedrock
- Model ID: `bedrock_converse/claude-opus-4-7` (Claude Opus 4.7)
- Mode: extended reasoning ("Auto" thinking effort)
- Tooling: ran inside an OpenCode local agent with shell access to a real Paperclip install where the bug was observed and reproduced.

## Checklist

- [ ] If this change affects the UI, I have included before/after screenshots — N/A, server-only fix; the UI behavior change is "files that should already be listed start being listed".
- [ ] I have updated relevant documentation to reflect my changes — no doc lists this rule explicitly; happy to add a note if reviewers want one.


cassioln:
**[Europa - DevOps]**: addressed both Greptile suggestions in 6ab1414.

**Issue 1 (cosmetic / simplification) — applied.** Collapsed the redundant `entry === skillPath` guard and the `isRootSkill ? true` ternary into a single `entry.startsWith(skillDirPrefix)` check (the prefix is empty when the skill sits at the import root, so it already accepts every walked entry). The relative-path derivation also collapses to `entry.slice(skillDirPrefix.length) || "SKILL.md"`, which yields `"SKILL.md"` for the root SKILL.md (slice of length 0) and the correct relative path otherwise.

**Issue 2 (defensive guard) — applied.** Real fix, not just a comment. When `SKILL.md` sits at the import root **and** a nested subdirectory also exposes its own `SKILL.md`, the previous code let the root inventory absorb the nested subtree while the nested skill also registered those same files (silent double-inclusion). I now precompute the directories of every discovered `SKILL.md` and exclude any nested-skill subtree from the root inventory's filter. Nested-only and root-only layouts are unaffected.

**Verification:**
- Added regression test `does not let a root-level SKILL.md absorb files of nested sibling skills` covering the mixed layout (`SKILL.md` + `references/` at root, plus `nested/SKILL.md` + `nested/scripts/`). Both skills now get exactly their own files.
- `vitest run src/__tests__/company-skills.test.ts` → 15/15 (was 14/14).
- `vitest run src/__tests__/company-skills` (full suite) → 48/48.
- Branch updated against `upstream/master` (`1f70fd9`) before pushing.

cassioln:
**[Callisto - DevOps]**: Auditoria sistêmica do fluxo de PRs no fork ksio.dev (`KSI-760`) — registro do status deste PR.

Decisão: **manter upstream-bound**.

## Justificativa

Este PR não é candidato a re-target para `cassioln/paperclip:ksio.dev/main` pelas seguintes razões:

1. É um **fix genérico do Skills subsystem** (`server/src/services/company-skills.ts` — edge case de `SKILL.md` na raiz do import). Afeta qualquer instância Paperclip, não apenas ksio.dev.
2. Não tem branch persona ksio.dev (`fix/local-skill-import-root-skillmd` é descritivo e neutro).
3. Não tem amarração a issue ksio.dev específica.
4. Não bloqueia entrega ksio.dev: o caminho de import via parent-dir já funciona — este é o nicho do edge case.
5. Tem testes de regressão completos: é exatamente o tipo de boa-cidadania que a política `ksio.dev/main`-first prevê como exceção upstream-keep.

Política referência: `paperclip-src/doc/ksio-dev/PAPERCLIP_KSIODEV_ONLY_CHANGES.md` — default rotineiro é fork; exceção upstream para mudanças genéricas de boa-cidadania, com justificativa explícita.

## Ação

PR mantido aberto contra `paperclipai/paperclip:master`. Sem re-target. Sem dual-path (não há urgência ksio.dev específica).

Aguardando review/merge upstream pelo time `paperclipai`.

PR #6992, opened 2026-05-29
fix(gemini-local): clear stale session ID after successful fallback to fresh session (GH#6608)

### LINKED ISSUE #6608: Bug: gemini_local adapter does not clear stale session ID after successful fallback — causes permanent per-heartbeat failure loop
# Bug: `gemini_local` adapter does not clear stale session ID after successful fallback — causes permanent per-heartbeat failure loop

## Environment

- `paperclipai` 2026.513.0 / 2026.517.0, `@paperclipai/adapter-gemini-local`
- Adapter type: `gemini_local`
- Reproduced on macOS (arm64), Gemini CLI installed globally
- Any agent with a heartbeat interval long enough for Gemini CLI to expire its local session files (typically > ~24–48 h)

## Summary

The `gemini_local` adapter correctly detects when a stored session ID is no longer valid and retries the run with a fresh session (`runAttempt(null)`). However, after the retry succeeds, the **stale session ID is not cleared from `agent_runtime_state`**. On every subsequent heartbeat, the adapter loads the same invalid ID, fails on the first attempt, falls back, and succeeds — but the root cause is never resolved. The agent is degraded permanently until an operator manually clears the session ID from the database.

## How Gemini CLI session expiry works

The Gemini CLI stores session files in `~/.gemini/tmp/<dir>/chats/` as `.jsonl` files. These files have a finite retention window and are cleaned up automatically by the CLI. If an agent's heartbeat interval is long (e.g. 1 hour), or if the agent was paused/errored for a period, its session file will be gone by the time the next heartbeat fires.

This is **expected behavior** from the Gemini CLI. The bug is entirely on the Paperclip side — it does not update its persisted state to reflect that the session is gone.

## Code path

**File:** `@paperclipai/adapter-gemini-local/dist/server/execute.js`

The adapter has correct detection and fallback logic:

```js
const initial = await runAttempt(sessionId);
if (
  sessionId &&
  !initial.proc.timedOut &&
  (initial.proc.exitCode ?? 0) !== 0 &&
  isGeminiUnknownSessionError(initial.proc.stdout, initial.proc.stderr)
) {
  await onLog("stdout", `[paperclip] Gemini resume session "${sessionId}" is unavailable; retrying with a fresh session.\n`);
  const retry = await runAttempt(null);   // ← starts fresh, works
  return toResult(retry, true, true);     // ← returns result
}
```

The retry run produces a new `sessionId` in its output. However, `toResult(retry, true, true)` returns the new session ID in the result payload, and **`agent_runtime_state.session_id` is never updated to the new value nor cleared to NULL**. The next heartbeat reads the same stale ID and the cycle repeats.

## Reproduction

1. Create a `gemini_local` agent with a heartbeat interval ≥ 1 hour (or simply pause it for 24+ hours).
2. Let the Gemini CLI expire the agent's session file.
3. Resume normal heartbeats.
4. Observe: every single heartbeat run begins with:
   ```
   Error resuming session: Invalid session identifier "<uuid>".
   Searched for sessions in ~/.gemini/tmp/<dir>/chats.
   ```
   followed by a successful fallback — but the same error appears on the next heartbeat too.
5. Confirm in `agent_runtime_state`: `session_id` still holds the original stale UUID, `last_error` still reflects the session error, `last_run_status = 'failed'` — even though runs are technically completing via the fallback path.

## Impact

- **Every heartbeat wastes one failed Gemini CLI invocation** before the fallback fires. For a 5-minute heartbeat, this doubles the number of Gemini processes spawned over time.
- **`last_run_status` stays `'failed'`** in `agent_runtime_state` even when the agent completes work successfully via fallback, making monitoring unreliable.
- **Secondary effects from prolonged degradation:** an agent that repeatedly sees its own runs as "failed" may attempt self-repair actions. In our case, a CEO-role agent (`gemini_local`) that was stuck in this loop for several hours created a duplicate CTO agent as an autonomous recovery attempt, requiring manual cleanup.
- Operators have no indication this is happening unless they inspect raw run logs — the UI shows the agent as `running` with no visible error.

## Suggested fix

When the fallback path is taken and `retry` completes successfully, the adapter (or the server layer that persists the result) should:

1. **Clear the stale session ID** from `agent_runtime_state` immediately (`session_id = NULL`), so the next heartbeat starts fresh without a failed first attempt.
2. Ideally, **persist the new session ID** returned by the retry run, so subsequent heartbeats can resume the newly created session normally.

The simplest safe fix is (1) alone: after detecting `isGeminiUnknownSessionError`, set `session_id = NULL` before or immediately after the retry, regardless of outcome.

## Workaround

Clear the stale session ID manually in the database:

```sql
UPDATE agent_runtime_state
SET session_id = NULL,
    last_error = NULL,
    last_run_status = 'completed',
    updated_at = NOW()
WHERE agent_id = '<agent-uuid>';
```

The next heartbeat will start a fresh Gemini session and persist its ID normally.

## Relationship to other issues

This is distinct from issue #6596 (recovery storm) — it does not create spurious issues or burn tokens in a loop. The blast radius is one wasted CLI invocation per heartbeat cycle. However, the secondary effect (agent creating a duplicate of another agent as self-repair) shows it can have unpredictable consequences in agentic environments with autonomous task-creation permissions.

isak-ialogics:
## Summary

Fixes #6608 — `gemini_local` adapter did not clear the stale session ID after a successful fallback to a fresh session, causing every subsequent heartbeat to fail on the stale session resume attempt.

## Changes

- `packages/adapters/gemini-local/src/server/execute.ts`: after a successful fresh-session fallback, clear the persisted `sessionId` in the agent's runtime state so the next heartbeat starts fresh

## Test Plan

- [ ] Gemini agent with a stale session ID in runtime state
- [ ] Trigger heartbeat — fallback to fresh session succeeds
- [ ] Verify subsequent heartbeats do not attempt to resume the old stale session

### DIFF
--- a/packages/adapters/gemini-local/src/server/execute.ts
+++ b/packages/adapters/gemini-local/src/server/execute.ts
@@ -662,7 +662,16 @@ export async function execute(ctx: AdapterExecutionContext): Promise<AdapterExec
         `[paperclip] Gemini resume session "${sessionId}" is unavailable; retrying with a fresh session.\n`,
       );
       const retry = await runAttempt(null);
-      return toResult(retry, true, true);
+      const retryResult = toResult(retry, true, true);
+      // Always clear after a stale-session fallback. The retry may emit a session
+      // ID that is itself stale (e.g. Gemini CLI surfaces a cached identity after
+      // server migration). Unconditional clear ensures the next heartbeat starts
+      // fresh and captures a stable new session it can actually reuse.
+      retryResult.clearSession = true;
+      retryResult.sessionId = undefined;
+      retryResult.sessionParams = undefined;
+      retryResult.sessionDisplayId = undefined;
+      return retryResult;
     }
 
     return toResult(initial);
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -622,6 +622,8 @@ importers:
         specifier: ^5.7.3
         version: 5.9.3
 
+  packages/skills-catalog: {}
+
   server:
     dependencies:
       '@aws-sdk/client-s3':
--- a/server/src/__tests__/gemini-local-execute.test.ts
+++ b/server/src/__tests__/gemini-local-execute.test.ts
@@ -432,4 +432,65 @@ describe("gemini execute", () => {
       await fs.rm(root, { recursive: true, force: true });
     }
   });
+
+  it("clears session after successful stale-session fallback even when retry emits a session ID", async () => {
+    const root = await fs.mkdtemp(path.join(os.tmpdir(), "paperclip-gemini-stale-fallback-"));
+    const workspace = path.join(root, "workspace");
+    const commandPath = path.join(root, "gemini");
+    const invocationCountPath = path.join(root, "invocations");
+    await fs.mkdir(workspace, { recursive: true });
+
+    // First invocation: fails with stale session error.
+    // Second invocation (fresh retry): succeeds and emits a new session ID.
+    const script = `#!/usr/bin/env node
+const fs = require("node:fs");
+const count = (() => { try { return parseInt(fs.readFileSync(${JSON.stringify(invocationCountPath)}, "utf8")) + 1; } catch { return 1; } })();
+fs.writeFileSync(${JSON.stringify(invocationCountPath)}, String(count), "utf8");
+if (count === 1) {
+  console.error("No previous sessions found with id stale-session-abc");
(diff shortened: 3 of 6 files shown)


isak-ialogics:
Closing: duplicate of #7129 fix/gemini-clear-session-after-stale-fallback-clean (same fix). Keeping #7129.

PR #6993, opened 2026-05-29
fix(routines): persist variables field in PATCH /api/routines/:id (GH#6525)

### LINKED ISSUE #6525: PATCH /api/routines silently drops 'variables' field — HTTP 200 + revision bump but no-op
## Summary

`PATCH /api/routines/:id` accepts a `variables` field in the body, returns **HTTP 200 OK** with a bumped `latestRevisionNumber`, but **silently drops the `variables` content**. The response payload shows `variables: []` and the persisted `routine_revisions.snapshot.routine.variables` is also `[]`.

This is a correctness footgun: operators believe their config landed, the audit trail shows a revision bump, but the runtime behavior is unchanged. Bugs hide for weeks.

## Reproduction (RCV instance, `v2026.517.0`)

```bash
curl -X PATCH \
  -H "Authorization: Bearer $TOKEN" \
  -d '{
    "variables": [
      {
        "name": "orphan_terminal_branch",
        "label": "Orphan terminal housekeeping pass",
        "type": "boolean",
        "defaultValue": false,
        "value": true,
        "required": false
      }
    ]
  }' \
  "https://bridge.../api/routines/7596e6d7-96f7-4cee-a518-e33b735a17bc"
```

**Response:** HTTP 200 OK. `latestRevisionNumber: 2` (was 1). `variables: []` in response body.

**DB state:**
```sql
SELECT revision_number, jsonb_path_query_array(snapshot, '$.routine.variables[*].name') AS vars
FROM routine_revisions WHERE routine_id='7596e6d7-...' ORDER BY revision_number DESC LIMIT 2;
--  revision_number | vars
-- -----------------+------
--                2 | []     ← bumped but empty
--                1 | []
```

The new revision row was created, but with empty `variables`. The PATCH body was effectively a no-op for that field, despite a 200 response.

## Impact in RCV

Caused **3 separate board approvals to be opened by agents over a 24h window** (`32acffde`, `74ebf596`, `7172c599`), each escalating "the variable I set didn't take effect" with progressively more diagnostic detail. Operator had to apply a SQL INSERT workaround on `routine_revisions` to manually publish a snapshot with the variable populated — see workaround section below.

## Workaround (operator SQL)

```sql
-- 1. Insert new revision with variables populated in snapshot
INSERT INTO routine_revisions (
  company_id, routine_id, revision_number, title, description, snapshot, change_summary
)
SELECT 
  company_id, routine_id, 
  COALESCE(MAX(revision_number), 0) + 1 OVER (),
  title, description,
  jsonb_set(snapshot, '{routine,variables}',
    '[{"name":"orphan_terminal_branch","type":"boolean","value":true,"defaultValue":false,"required":false}]'::jsonb
  ),
  'Operator manual snapshot publish — workaround for PATCH variables silent drop'
FROM routine_revisions WHERE routine_id='<ROUTINE_ID>'
ORDER BY revision_number DESC LIMIT 1
RETURNING id;

-- 2. Point routines.latest_revision_* at the new revision
UPDATE routines SET 
  latest_revision_id = '<NEW_REVISION_ID>',
  latest_revision_number = <NEW_NUM>,
  variables = <the same jsonb>
WHERE id='<ROUTINE_ID>';
```

## Suggested fix

In the PATCH handler (likely `server/src/routes/routines.ts`):

1. Either **accept** the `variables` field and persist it in both `routines.variables` and `routine_revisions.snapshot.routine.variables`
2. Or **reject** PATCHes that include unknown / unhandled fields with HTTP 422 + a clear error message ("`variables` cannot be updated via this endpoint; use `POST /api/routines/:id/revisions` instead.")

The "silent 200 + revision bump" path is the worst of both worlds — looks successful but does nothing.

## Environment

- Paperclip `v2026.517.0`
- Embedded PostgreSQL
- 22 active agents, ~24 active routines

## Cross-reference

- Memory `routine_revision_snapshot_trap` already documents this for operator memory but it should be a framework-level fix, not a workaround pattern.
- Related: agent capability gates also lacks `routine.variables.write` as a distinct permission (currently bundled in `routine.write` which is the API-level gate that lets PATCH succeed-but-no-op).

isak-ialogics:
## Summary

Fixes #6525 — `PATCH /api/routines/:id` silently dropped the `variables` field, returning HTTP 200 with a bumped revision but not persisting the new variables. Also removes a duplicate `txDb.update` call in the patch handler.

## Changes

- `server/src/services/routines.ts`: include `variables` in the update set when patching a routine; remove extraneous duplicate update call

## Test Plan

- [ ] `PATCH /api/routines/:id` with `{ "variables": [...] }` body
- [ ] `GET /api/routines/:id` returns the updated variables
- [ ] Revision number increments correctly


isak-ialogics:

PR #7000, opened 2026-05-29
fix(gemini-local): treat cost_usd=0 as missing, fall back to token-based estimate

### LINKED ISSUE #6844: bug: Gemini metered_api cost_cents stops recording after model switch — 7B tokens with $0 cost
## Environment
- Paperclip version: 2026.517.0 (at time of bug), now 2026.525.0
- Adapter: gemini_local
- Models: gemini-3.5-flash, gemini-2.5-flash, gemini-2.5-flash-lite
- billing_type: metered_api

## Bug

Gemini cost tracking recorded correctly from May 18-22, then stopped recording cost_cents despite continued massive token usage from May 23-26.

Data from cost_events:
```
May 22: 162 Gemini events, cost = $29.22 (correct)
May 23: 134 events, 630M input tokens, cost = $0.00 (WRONG)
May 24: 175 events, 1.6B input tokens, cost = $0.00 (WRONG)
May 25: 206 events, 2.8B input tokens, cost = $0.00 (WRONG)
May 26: 174 events, 1.3B input tokens, cost = $0.00 (WRONG)
```

The billing_type is correctly set to `metered_api` on all events. Token counts are recorded. But cost_cents is 0.

## What Changed on May 22-23

On May 22-23, we switched agents between Claude and Gemini models multiple times:
1. Some agents were switched from Claude to Gemini 3.5 Flash
2. adapter_config.model was changed
3. runtime_config.modelProfiles.cheap was changed
4. agent_task_sessions were purged

The cost tracking worked for the first few hours of Gemini usage (May 22 shows $29.22) but then stopped.

## Hypothesis

The Gemini adapter's `costUsd` field in the result may have stopped being populated after the model switch, possibly due to:
- The adapter losing its pricing lookup after a reconfiguration
- A session resume issue where the new session doesn't carry pricing context
- The adapter reporting costUsd as null/undefined, which normalizeBilledCostCents() converts to 0

## Impact

$1,100+ in Gemini API charges were not recorded in Paperclip's cost tracking. The operator had no visibility into actual spend until manually backfilling from token counts and published pricing.

## Workaround

We backfilled cost_cents from token counts using published Gemini pricing. This captures the historical data but doesn't fix the underlying adapter issue.

isak-ialogics:
## Problem

After a Gemini CLI update around May 22–23, the CLI began returning `cost_usd: 0` instead of `null`/omitting the field when cost data is unavailable. The previous guard condition `!= null` evaluated `0` as truthy, so `costUsd` was stored as `0` and the token-based fallback calculation never fired.

## Fix

Change the condition from `attempt.parsed.costUsd != null` to `attempt.parsed.costUsd != null && attempt.parsed.costUsd > 0`. A zero reported cost is now treated the same as a missing cost when tokens were consumed, so the token-based estimate fires correctly.

Fixes #6844

## Test plan
- [ ] Confirm that a Gemini run where the CLI returns `cost_usd: 0` now stores a non-zero `costUsd` derived from token counts
- [ ] Confirm that a run with no tokens consumed still stores `null`
- [ ] Confirm that a run with a legitimate non-zero `cost_usd` still uses the CLI-reported value


isak-ialogics:

PR #7001, opened 2026-05-29
feat(costs): add shadow_cost_cents for subscription-plan cost visibility

### LINKED ISSUE #6843: feat: shadow/market-rate cost tracking for subscription-included plans (Claude Max)
## Problem

When using Claude Max (subscription plan), Paperclip correctly detects `billingType: "subscription_included"` and sets `cost_cents = 0` in cost_events. This is technically accurate — there's no per-token charge.

However, operators on subscription plans lose visibility into what their usage WOULD cost at API rates. This makes it impossible to:
- Evaluate whether the subscription is cost-effective vs API billing
- Set meaningful per-agent budgets (budgets in cents are all $0)
- Compare agent efficiency by cost
- Track cost trends over time
- Plan for scaling (adding agents, changing models)

## Current Behavior

`normalizeBilledCostCents()` in heartbeat.js:
```js
if (billingType === "subscription_included") return 0;
```

The `costUsd` from the adapter IS being calculated and stored in `usage_json`, but `cost_cents` in `cost_events` is forced to 0.

## Proposed Solution

Add a `shadow_cost_cents` column (or `market_rate_cents`) to `cost_events` that always calculates the market-rate cost regardless of billing type. The existing `cost_cents` continues to reflect actual charges (0 for subscription, real for metered). Operators can then see both:

- `cost_cents` = what you actually paid
- `shadow_cost_cents` = what it would cost at API rates

Alternatively, expose a per-company config to opt into shadow cost tracking that populates `cost_cents` with market rates even on subscription plans.

## Workaround

We're running a daily cron that backfills `cost_cents` with calculated market rates based on token counts and published pricing. This works but:
- Pricing must be manually updated when models change
- It overwrites the "actual cost" semantics of cost_cents
- It doesn't survive Paperclip updates or schema changes

## Impact

Any operator on Claude Max, Gemini subscription, or similar flat-rate plans has zero cost visibility in the Paperclip dashboard. This affects budgeting, efficiency analysis, and model selection decisions.

isak-ialogics:
## Problem

Subscription-plan operators (Claude Max, Gemini subscription, etc.) get `cost_cents = 0` on all cost events, which is technically correct but renders per-agent budgets, cost dashboards, and efficiency comparisons useless. The adapter does calculate `costUsd`, but it gets zeroed out before storage.

## Fix

Add a `shadow_cost_cents` nullable integer column to `cost_events`. It always stores the market-rate cost in cents (derived from `costUsd`) regardless of `billingType`. `cost_cents` semantics are unchanged — it still reflects actual charges. Operators can query `shadow_cost_cents` for budget analysis, trend tracking, and model comparison.

- `shadow_cost_cents IS NULL` → adapter returned no cost data (no tokens consumed)
- `shadow_cost_cents = 0` → adapter reported `costUsd = 0` (edge case)
- `shadow_cost_cents > 0` → market-rate cost estimate

Migration: `0094_shadow_cost_cents.sql` (single `ALTER TABLE ADD COLUMN`, non-blocking on Postgres 11+).

Fixes #6843

## Test plan
- [ ] After migration, cost events for subscription-included runs have `shadow_cost_cents > 0` and `cost_cents = 0`
- [ ] Cost events for metered runs have equal `cost_cents` and `shadow_cost_cents`
- [ ] Cost events with no token usage have `shadow_cost_cents IS NULL`


isak-ialogics:

PR #7003, opened 2026-05-29
perf(heartbeat): narrow heartbeat_runs SELECT to avoid full JSONB fetch in hot-path queries

### LINKED ISSUE #6947: perf: heartbeat_runs queries select all 47 columns including large JSONB — causes CPU saturation with concurrent agents
## Environment
- Paperclip version: 2026.525.0
- Server: 4 vCPU, 16GB RAM, Ubuntu 24.04
- Agents: 11 active (4 timer, 7 task-activated)
- Database: embedded PostgreSQL on port 54329

## Problem

When multiple agents heartbeat concurrently, the server load spikes to 10-12x (on 4 CPUs) because each heartbeat query selects ALL 47 columns from `heartbeat_runs` including three large JSONB fields:

- `context_snapshot`: ~5KB average per row
- `result_json`: ~1KB average per row
- `usage_json`: ~373 bytes per row

With 44K rows in the monthly partition and 9 concurrent agent queries, each heartbeat fetches megabytes of data from disk even when it only needs a few scalar fields (status, started_at, agent_id).

## Impact

- Server load average 10-12x on 4 CPUs during concurrent heartbeats
- Paperclip dashboard becomes unresponsive for 2-5 minutes
- 689MB swap usage from memory pressure
- All 11 agent heartbeats compete for CPU, slowing each other

## Query pattern observed

```sql
select "id", "company_id", "agent_id", "invocation_source", "trigger_detail",
  "status", "started_at", "finished_at", "error", "wakeup_request_id",
  "exit_code", "signal", "usage_json", "result_json", ...
  "context_snapshot", "created_at", "updated_at"
from "heartbeat_runs"
where ("heartbeat_runs"."company_id" = $1
  and "heartbeat_runs"."agent_id" = $2
  and ("heartbeat_runs"."context_snapshot"...))
```

All 47 columns selected even though most callers likely only need status + timestamps.

## Suggested fixes

1. **Use SELECT with specific columns** instead of SELECT * for heartbeat queries that only need status/timing info
2. **Exclude large JSONB columns** (context_snapshot, result_json, usage_json) from listing queries — fetch them only when viewing a specific run's detail
3. **Add connection pooling or query queuing** to prevent N concurrent heartbeats from all hitting the DB simultaneously
4. **Stagger heartbeat dispatch** more aggressively — current jitter of 120s helps but all agents still fire within the same window after a restart

## Workaround

We added a covering index to serve common queries from the index without heap access:
```sql
CREATE INDEX heartbeat_runs_agent_lookup_idx
ON heartbeat_runs_2026_05 (company_id, agent_id, status, started_at DESC)
INCLUDE (finished_at, error_code, invocation_source, trigger_detail, liveness_state);
```

This helps filtering but doesn't prevent the full-row fetch when all columns are selected.

## Server metrics during incident
- Load average: 10.28 / 10.41 / 11.15
- 9 Postgres connections all in 'active' state running the same SELECT
- Zero idle connections
- Swap: 689MB used

isak-ialogics:
## Problem

With 9+ concurrent agent heartbeats, every wakeup enqueueing call executed a `SELECT *` against `heartbeat_runs` fetching all 47 columns including three large JSONB fields (`context_snapshot` ~5KB, `result_json` ~1KB, `usage_json` ~373B). Under load this caused server CPU saturation (load 10-12x on 4 CPUs) and 689MB swap usage.

## Fix

Narrow two queries that are on the hot path but only need a few columns:

1. **`enqueueWakeup` active-run coalesce check** — fetches active runs per agent on every wakeup to determine coalescing. Callers only use `id`, `status`, `contextSnapshot`. Now selects only those 3 columns.
2. **`updateAgentStatus` latest-failed-run fetch** — only needs `error`, `errorCode`, `resultJson`. Narrowed accordingly.

Also relaxes `runTaskKey()` parameter type from `$inferSelect` to `Pick<...,"contextSnapshot">` to satisfy the type checker without casts.

Wide selects are left in place where the caller legitimately needs many fields (orphaned-run reaper, single-row PK lookups).

Fixes #6947

## Test plan
- [ ] Confirm wakeup coalescing still works correctly (same-scope queued runs are merged)
- [ ] Confirm failed-run error context still shows in agent status updates
- [ ] Confirm no TypeScript errors beyond pre-existing truncated-file errors


isak-ialogics:

PR #7024, opened 2026-05-30
feat: auto-release issue checkout when heartbeat run exits

jjlbpaiva:
## Thinking Path

> - Paperclip orchestrates AI agents that check out issues and run heartbeats
> - Each run calls `POST /api/issues/:id/checkout`, pinning four lock fields (`checkoutRunId`, `executionRunId`, `executionAgentNameKey`, `executionLockedAt`) to the run's id
> - When a run exits without PATCHing the issue away from `in_progress` (crash, OOM, timeout, or incomplete run), those fields are never cleared
> - The only thing currently clearing them is the atomic side-effect in `applyStatusSideEffects` when an agent PATCHes to a non-`in_progress` status — a path a dead run never reaches
> - This leaves the issue's checkout locked to a dead run id, causing every future `POST /checkout` to 409 until an admin manually force-releases the lock
> - This PR adds `releaseFromRunIfOwned` to `issueService` and hooks it into `environmentRunOrchestrator.releaseForRun`, which is called from every run exit path in `heartbeat.ts` (normal, crash, OOM, timeout, and process-lost recovery)
> - The benefit is that a crashed or abandoned run can no longer leave a permanently 409-ing checkout — the lock is atomically cleared on run exit, without changing any user-visible API semantics
> - Same PR bundles the skill documentation update: adds the "auto-release" callout so agents know /release is no longer required, and adds the "Admin-reset reference" subsection for stale-lock recovery (replacing the local-instance DON-478 stopgap that was never committed upstream)

## What Changed

- **`server/src/services/issues.ts`** — added `releaseFromRunIfOwned(issueId, heartbeatRunId)`: a transactional helper that acquires a `FOR UPDATE` row lock, checks whether the given run owns either `checkoutRunId` or `executionRunId`, and if so clears all four execution lock fields atomically. Returns `null` (no-op) when the run no longer owns the issue, preventing any interference with legitimate checkouts.

- **`server/src/services/environment-run-orchestrator.ts`** — imports `issueService` and instantiates it alongside the existing service singletons. In `releaseForRun`, after the existing environment lease activity log, calls `releaseFromRunIfOwned` for every lease that has an `issueId`. Emits `issue.auto_released_on_run_exit` activity on success. Wraps the call in try/catch so that any failure (transient DB error, etc.) is logged as a warning and never prevents the environment lease from being released.

- **`server/src/__tests__/environment-run-orchestrator.test.ts`** — adds a `describe` block with four unit tests covering: (1) AC1 happy-path auto-release with lock-field verification, (2) no-op for leases with no `issueId`, (3) no-op when run no longer owns the lock, (4) error isolation when `releaseFromRunIfOwned` throws.

- **`skills/paperclip/SKILL.md`** — adds two blocks under Step 8 of the heartbeat skill: (a) a callout informing agents that the harness now auto-releases any checkout this run owns on clean or crash exit — explicit `/release` calls are no longer required unless intentionally handing the task back to the queue; (b) an "Admin-reset reference (stale-lock recovery)" subsection that prescribes the ordered reset paths (board `force-release` → agent `/release` → manager `PATCH to todo`) when `activeRun` is null but checkout fields remain populated. The DON-478 stopgap block (local-instance hotfix only, never committed upstream) is superseded by the L1 fix and not included here.

## Verification

Run the unit tests:
```sh
cd server
pnpm exec vitest run src/__tests__/environment-run-orchestrator.test.ts
```
Expected: 12 tests pass (8 pre-existing + 4 new).

Run the TypeScript typecheck:
```sh
pnpm --filter "@paperclipai/server" typecheck
```
Expected: exit 0, no type errors.

**Manual acceptance (post-deploy):**
1. Check out an issue via `POST /api/issues/:id/checkout`.
2. Kill the heartbeat run without calling `/release` (simulate crash).
3. `GET /api/issues/:id` → `checkoutRunId === null`, `executionRunId === null`, `executionAgentNameKey === null`, `executionLockedAt === null`.
4. Confirm the `issue.auto_released_on_run_exit` activity log entry appears for the issue.

## Risks

- **No-op on non-owning run**: The `FOR UPDATE` lock + the `OR (checkoutRunId = ? OR executionRunId = ?)` predicate in the `UPDATE WHERE` clause guarantees idempotency — if a new run has already taken over the checkout before this fires, `releaseFromRunIfOwned` returns `null` and no update is applied.
- **Error isolation**: The entire auto-release block is wrapped in try/catch. A failure in this path can never block the environment lease from being released (the existing invariant of `releaseForRun`).
- **No API contract changes**: `releaseFromRunIfOwned` is a server-internal helper. The `/api/issues/:id/release` route semantics are unchanged.
- **Crash-exit coverage**: Both run-exit paths in `heartbeat.ts` — the `finally` block (all exits including OOM/crash) and the process-lost recovery path — already call `releaseForRun`, so no additional changes to `heartbeat.ts` are required.
- **Skill text**: Documentation-only change; no runtime behaviour affected if the SKILL.md commit is cherry-picked independently of the server commits.

Related issues: DON-477, DON-478 (plan rev 2), DON-474 (forensic recurrence shape).

## Model Used

- Provider: Anthropic
- Context window: 200K tokens
- Tool use: enabled (file read/edit, bash, grep)
- No extended thinking mode

## Checklist

### DIFF
--- a/server/src/__tests__/environment-run-orchestrator.test.ts
+++ b/server/src/__tests__/environment-run-orchestrator.test.ts
@@ -10,6 +10,8 @@ const mockBuildWorkspaceRealizationRequest = vi.hoisted(() => vi.fn());
 const mockUpdateLeaseMetadata = vi.hoisted(() => vi.fn());
 const mockUpdateExecutionWorkspace = vi.hoisted(() => vi.fn());
 const mockLogActivity = vi.hoisted(() => vi.fn());
+const mockReleaseFromRunIfOwned = vi.hoisted(() => vi.fn());
+const mockLoggerWarn = vi.hoisted(() => vi.fn());
 
 vi.mock("../services/environment-execution-target.js", () => ({
   resolveEnvironmentExecutionTarget: mockResolveEnvironmentExecutionTarget,
@@ -42,6 +44,16 @@ vi.mock("../services/execution-workspaces.js", () => ({
 
 vi.mock("../services/activity-log.js", () => ({
   logActivity: mockLogActivity,
+}));
+
+vi.mock("../services/issues.js", () => ({
+  issueService: vi.fn(() => ({
+    releaseFromRunIfOwned: mockReleaseFromRunIfOwned,
+  })),
+}));
+
+vi.mock("../middleware/logger.js", () => ({
+  logger: { warn: mockLoggerWarn },
 }));
 
 // ---------------------------------------------------------------------------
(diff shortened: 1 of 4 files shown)


jjlbpaiva:
Thanks for the review. Quick response on each:

**#3 — `status` field left as `in_progress` after auto-release: intentional.**

The auto-release path is recovery from crash/OOM/timeout mid-work, not a "go back to todo" reset. The exiting run had legitimately started the work — clearing `status` would erase that signal and force the next agent to re-discover context from scratch. After this fix:
- Lock fields clear → next agent CAN check out and resume.
- `status: in_progress` persists → kanban accurately shows "work was started, not actively executing right now."

The intended recovery flow is: next assigned agent wakes (via existing wake triggers), sees `status: in_progress` + `activeRun: null`, calls `/checkout`, gets the lock atomically (since lock fields are now clear), and resumes. If no agent ever resumes, the issue stays visible in `in_progress` which is the correct signal for human triage — exactly opposite of the silent-stale-lock failure mode this PR fixes.

Happy to add an inline comment in `releaseFromRunIfOwned` documenting this if you'd like. Would prefer to keep the diff minimal for review; can ship the comment in a follow-up.

**#1 — misleading warning when `logActivity` throws:** Valid concern. Will split into nested try/catch in a follow-up commit if review surfaces other changes; otherwise leaving for a small follow-up PR to keep this one focused on the core lock-clearing fix.

**#2 — `FOR UPDATE` + second `SELECT` round-trip:** Also valid. Same disposition as #1 — small follow-up; the current shape is correct, just one extra DB hop per run exit (low frequency, not on hot path).

All three are non-blocking for the original incident this PR closes (stale `checkoutRunId` wedging every subsequent checkout). Ready for human review.

jjlbpaiva:
Thanks for the review @greptile-apps! Quick reply on concern #3 (the `status: in_progress` left-as-is design question), and a brief plan for #1/#2.

**On #3 — intentional design.** The auto-release path is specifically for *crash mid-execution* recovery, not for resetting work back to the start. Resetting `status` to `todo` (or anything else) would erase the signal that this issue was already picked up and started — and would mislead any human triaging the kanban into thinking the issue is fresh and assignable. Leaving `status: in_progress` with cleared lock fields means: "work was in progress, the executing run died, but the lock no longer wedges a new checkout." The next agent that picks it up (manual reassign, requeue, or a scheduled wake) inherits a `status: in_progress` issue and a clean lock — exactly the state we want for resumption. The harness `releaseFromRunIfOwned` is the lock-clear primitive; status transitions belong to status-aware paths (interactions, blockers, monitor wakes), not the crash recovery path.

That said, the concern is fair that a perpetually-`in_progress` issue with no live run looks odd in UIs. The existing kanban already has the `activeRun: null` signal to flag "stale execution" — I'd argue the cleanup belongs in that surfacing layer (UI badge, sweeper job for N-hour-stale `in_progress` rows), not in the auto-release primitive.

Happy to add a code comment to `releaseFromRunIfOwned` explaining this intent if you'd like.

**On #1 & #2** — both are accurate observations. Both are minor (misleading log warning + 1 DB round-trip on every run-exit-with-issueId). I'd prefer to land them as a follow-up PR rather than expand this one's scope; the core ownership-guarded clear is what unblocks the stale-lock-recurrence problem we hit yesterday (4× in 30min across two agents — context in [Donnah-internal DON-477](https://github.com/donnah/internal) → DON-478). Happy to push the follow-up immediately after this merges, or now if you'd prefer them bundled.

PR #7031, opened 2026-05-30
Fix continuation recovery retry streaks by failure cause

devinfoley:
## Thinking Path

> - The recovery subsystem is responsible for keeping assigned work moving when a live heartbeat run disappears or fails.
> - `continuation_recovery` is the path that re-enqueues stranded `in_progress` issues after an interrupted continuation attempt.
> - That path recently gained cause-aware retry classes and transient retry caps, but the streak counter was still aggregating mixed failure causes into one retry history.
> - That meant a sequence like `timeout -> timeout -> adapter_failed -> adapter_failed` could escalate as a false `3x adapter_failed` streak even though the latest cause had only happened twice.
> - This pull request makes continuation retry streaks count only consecutive failures whose `errorCode` matches the latest run and adds a regression test for the mixed-cause case.
> - The benefit is that transient retry backoff and escalation now match the actual current failure cause instead of inheriting stale budget from unrelated failures.

## What Changed

- Updated `summarizeRecentContinuationRetries(...)` to stop counting as soon as the continuation failure cause no longer matches the latest run's `errorCode`.
- Wired the continuation recovery escalation/backoff path to pass the latest classified `errorCode` into the retry streak summarizer.
- Added a regression test proving mixed-cause continuation failures do not consume the transient retry cap for a new failure cause.

## Verification

- `pnpm exec vitest run server/src/__tests__/heartbeat-process-recovery.test.ts`

## Risks

- Low risk. The behavioral change is intentionally narrow, but any future continuation retry modes that rely on `errorCode = null` will now be counted as a separate streak bucket and should be kept in mind when adding new retry classifications.

## Model Used

- OpenAI Codex via Paperclip `codex_local` (GPT-5-based Codex coding agent; exact backend revision is not surfaced in the runtime), with tool use, shell execution, and patch application in the local repository.

## Checklist

### DIFF
--- a/server/src/__tests__/heartbeat-process-recovery.test.ts
+++ b/server/src/__tests__/heartbeat-process-recovery.test.ts
@@ -2742,7 +2742,13 @@ describeEmbeddedPostgres("heartbeat orphaned process recovery", () => {
 
     const runs = await db.select().from(heartbeatRuns).where(eq(heartbeatRuns.agentId, agentId));
     expect(runs).toHaveLength(5);
-    const retryRun = runs.find((row) => row.id !== runId && row.errorCode !== "timeout" && row.createdAt > new Date("2026-03-18T23:55:00.000Z"));
+    const retryRun = runs.find((row) => {
+      const ctx = row.contextSnapshot as Record<string, unknown> | null;
+      return row.id !== runId &&
+        row.errorCode === null &&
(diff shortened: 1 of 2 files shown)


devinfoley:
@greptile review

devinfoley:
@greptile review

PR #7035, opened 2026-05-30
fix(ops-portal): drop output:standalone so Railway/Vercel next start works

k5tuck:
Railway (Nixpacks) and Vercel both run `next start`, which is incompatible with `output: 'standalone'` — the paperclip-ops-portal Railway service logs `"next start" does not work with "output: standalone"` and fails to serve. Standalone is only for Docker self-hosting (not used here); removing it lets `next start` serve the app + static assets on both targets. Unblocks the Railway paperclip-ops-portal (ops.torinagi.com login shell).

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -56,3 +56,4 @@ tests/release-smoke/test-results/
 tests/release-smoke/playwright-report/
 .superset/
 .claude/worktrees/
+.vercel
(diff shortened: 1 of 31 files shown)


k5tuck:
Opened against upstream by mistake (gh fork default) — re-targeting k5tuck/paperclip.

PR #7042, opened 2026-05-30
fix(codex-local): classify spark aux-model 400 as transient (FIG-773)

figmenta-forge:
## Summary

- Codex CLI internally spawns auxiliary turns on \`gpt-5.3-codex-spark\` for compaction/summary work, even when \`--model gpt-5.3-codex\` is set. On ChatGPT-subscription auth (Vega's mode per FIG-740), the backend rejects spark with HTTP 400: \`The 'gpt-5.3-codex-spark' model is not supported when using Codex with a ChatGPT account.\`
- The aux call is intermittent (depends on Codex's internal heuristics), so a single retry usually succeeds — but today the failure is classified as a hard \`adapter_failed\`, not transient, so the harness surfaces it as a terminal error and the issue thread fills with noise (Vega's FIG-770 review run pattern).
- This patch adds a regex pattern to \`isCodexTransientUpstreamError\` for the spark-rejection error shape so the failure is classified \`transient_upstream\` and the harness's existing retry path takes over. No new config field, no behavioural change for any other code path.

## Why this is the smallest change

- The spark call originates inside the Codex CLI binary (\`packages/adapters/codex-local\` only passes \`--model gpt-5.3-codex\`). Paperclip cannot directly suppress the aux call without changing Codex CLI or guessing at unverified \`-c\` overrides.
- The error IS intermittent and IS recoverable on retry — both characteristics of \`transient_upstream\`. Empirically, every observed failure was followed by a successful retry run (\`retryOfRunId\` chain in \`heartbeat-runs\`).
- Long-term, the right fix is upstream in \`openai/codex\` (related: openai/codex#17642). This patch is the defensive Paperclip-side classifier that keeps Codex-on-subscription agents working until upstream lands.

## Verification

- 11/11 \`packages/adapters/codex-local\` parse tests pass (2 new tests cover both the positive match and a negative for unrelated 400s so the regex doesn't over-classify).
- Typecheck clean on the package.
- Verification on Vega's heartbeat with a trivial task to follow after merge.

## Test plan

- [x] \`npx vitest run parse.test.ts\` green (11 passed)
- [x] \`npx tsc --noEmit\` clean on \`packages/adapters/codex-local\`
- [ ] Post-merge: dispatch a trivial task to Vega, confirm zero \`adapter_failed\` (or \`transient_upstream\` classification with successful retry) in \`agent_runtime_state\`

Refs FIG-773

### DIFF
--- a/packages/adapters/codex-local/src/server/parse.test.ts
+++ b/packages/adapters/codex-local/src/server/parse.test.ts
@@ -137,4 +137,20 @@ describe("isCodexTransientUpstreamError", () => {
       }),
     ).toBe(false);
   });
+
+  it("classifies the spark auxiliary-model rejection on ChatGPT-subscription auth as transient upstream", () => {
+    const errorMessage =
+      "{\"type\":\"error\",\"status\":400,\"error\":{\"type\":\"invalid_request_error\",\"message\":\"The 'gpt-5.3-codex-spark' model is not supported when using Codex with a ChatGPT account.\"}}";
+    expect(isCodexTransientUpstreamError({ errorMessage })).toBe(true);
+    expect(isCodexTransientUpstreamError({ stdout: errorMessage })).toBe(true);
+  });
+
+  it("does not classify generic invalid_request_error 400 payloads as transient", () => {
+    expect(
+      isCodexTransientUpstreamError({
+        errorMessage:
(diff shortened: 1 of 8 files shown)


figmenta-forge:
Re-routing to figmenta-forge/paperclip (Figmenta deploys from the fork, not upstream).

PR #7147, opened 2026-05-30
heartbeat timer max concurrent

isak-ialogics:
Automated fix from overnight sprint. Branch: `fix/heartbeat-timer-max-concurrent-clean`.


isak-ialogics:
Closing: duplicate of #7148 (same fix: skip timer enqueue when agent already has an active timer run). Keeping #7148.

PR #7167, opened 2026-05-30
plain env binding redaction always

isak-ialogics:
Automated fix from overnight sprint. Branch: `fix/plain-env-binding-redaction-always-clean`.


isak-ialogics:
Closing: duplicate of #7176 (same fix: always redact plain env binding values). Keeping #7176 which has more specific description.

PR #7184, opened 2026-05-30
routines patch persist variables

isak-ialogics:
Automated fix from overnight sprint. Branch: `fix/routines-patch-persist-variables-clean`.


isak-ialogics:
Closing: duplicate of #7186 (same fix: persist variables field in PATCH /api/routines). Keeping #7186.

PR #7213, opened 2026-05-30
heartbeat runs column projection v2

isak-ialogics:
Automated fix from overnight sprint. Branch: `perf/heartbeat-runs-column-projection-v2-clean`.


isak-ialogics:
Closing: exact duplicate of #7214 (same fix: narrow heartbeat_runs SELECT columns in hot-path queries). Keeping #7214.

PR #7219, opened 2026-05-30
fix(shutdown+heartbeat+oom): kill orphaned agents, prevent queue pile-up, reduce OOM risk

### LINKED ISSUE #7411: fix(service): add OOMPolicy=continue to paperclip.service — agent OOM kill cascades to server crash
## Summary

When the Linux OOM killer kills an agent subprocess inside the `paperclip.service` cgroup, systemd terminates the entire Paperclip server due to the default `OOMPolicy=stop`. The server should survive agent OOM kills.

## Current service configuration

```ini
[Service]
KillMode=control-group
OOMScoreAdjust=-500   # protects server from direct OOM kill ✓
# OOMPolicy is absent → defaults to stop ✗
```

## Failure chain

1. Memory pressure spikes (WSL2, no swap, many concurrent agent runs)
2. OOM killer targets a claude agent subprocess (inside `paperclip.service` cgroup, `oom_score_adj=500`)
3. Kernel kills the agent process
4. systemd sees an OOM kill in the service cgroup → `OOMPolicy=stop` fires
5. systemd sends SIGTERM to the Paperclip server (control process)
6. Full service crash; all in-flight agent runs lost; Restart=on-failure brings it back ~30 s later

## Evidence

```
kernel: claude invoked oom-killer: gfp_mask=0x140cca, order=0, oom_score_adj=0
kernel: oom-kill: task_memcg=/system.slice/paperclip.service, task=claude, pid=359815, oom_score_adj=500
kernel: Out of memory: Killed process 359815 (claude) total-vm:73152236kB, anon-rss:95616kB
systemd: paperclip.service: A process of this unit has been killed by the OOM killer.
systemd: paperclip.service: Control process exited, code=killed, status=15/TERM
systemd: paperclip.service: Failed with result 'oom-kill'.
```

## Fix

Add `OOMPolicy=continue` to the service unit so agent subprocess OOM kills do not propagate to the server:

```ini
[Service]
KillMode=control-group
OOMScoreAdjust=-500
OOMPolicy=continue      # ← add this
```

`OOMScoreAdjust=-500` already protects the server process itself from being chosen as the OOM victim. `OOMPolicy=continue` prevents systemd from stopping the service when a *different* process in the cgroup (an agent subprocess) is OOM-killed.

## Notes

- This complements PR #7219 which added `OOMScoreAdjust=-500` but did not add `OOMPolicy=continue`
- `OOMPolicy` requires systemd ≥ 246 (available on Ubuntu 22.04+)
- Without this, every memory spike that kills an agent also kills the server, destroying all concurrent agent state

mrveiss:
## Thinking Path

> - Paperclip orchestrates AI agent fleets that spawn many short-lived `claude` subprocesses per heartbeat, all within the `paperclip.service` cgroup
> - Under memory pressure on WSL2 (no swap), the Linux OOM killer selects victims by `oom_score_adj`; without explicit scoring, agent processes and the server share equal priority and either can be killed first
> - Post-crash restarts left orphaned `claude --tools` processes because shutdown did not propagate SIGTERM to spawned subprocesses; these accumulated across restarts, consuming memory until the next OOM event
> - The heartbeat reaper incorrectly treated in-flight runs with no stored PID (server crashed before PID was persisted) as dead and immediately re-dispatched them, creating exponential process pile-up on every restart
> - The dispatcher used a blocklist of non-invokable statuses that silently passed agents in `error` state and future unknown states through to dispatch, triggering more unwanted runs
> - The heartbeat timer and restart path could enqueue duplicate run starts before prior runs completed, further accelerating the spiral
> - This PR addresses all five failure modes observed in a production WSL2 instance that went from 49 orphaned agent processes + load 18 to 4 runs reaped cleanly + load settling to 6

Fixes #7411

## What Changed

**packages/adapter-utils/src/server-utils.ts**
- Sets `oom_score_adj=500` on each spawned agent subprocess immediately after fork so the kernel preferentially kills agent processes over the Paperclip server under memory pressure

**server/src/services/heartbeat.ts**
- Dispatcher: replaced open blocklist (`paused | terminated | pending_approval`) with explicit allowlist (`idle | running`); agents in `error` or any unknown status no longer receive dispatches
- Reaper grace period: `claude_local` runs with no stored PID now get a 10-minute grace window instead of being immediately reaped and re-dispatched
- Queue pile-up guard: dedup check prevents multiple pending starts for the same agent during restart or timer ticks

**server/src/services/recovery/service.ts**
- Tracks agents with pending recovery in-memory during reconciliation to avoid enqueuing duplicate starts per agent per sweep

## Verification

Tested on a live Ubuntu 22.04 WSL2 instance with a `claude_local` agent fleet:

- **Before**: post-restart → 49 orphaned `claude` processes, load average 18, WSL freeze requiring hard reboot
- **After**: post-restart → startup log shows `reaped orphaned heartbeat runs: {reapedCount: 4}`, 3 agents dispatched cleanly, load settled to ~6 within 5 minutes

## Risks

- Dispatcher allowlist change: any agent state not in the allowlist will not receive dispatches. All legitimate dispatch states are covered; any missed state surfaces quickly in tests and monitoring.
- Grace period for PID-less runs: a truly stuck run lingers up to 10 minutes longer before escalation. Acceptable given it prevents false re-dispatch spirals.
- No schema changes; no migration risk.

## Model Used

Claude Sonnet 4.6 (`claude-sonnet-4-6`) via Paperclip agent fleet

## Checklist

- [x] Tests pass locally
- [x] No breaking changes to public API
- [x] Documentation updated where applicable


mrveiss:
Added a second commit (f74875ca) to this PR covering the OOM side of the same problem:

- Writes `oom_score_adj=500` to `/proc/<pid>/oom_score_adj` immediately after spawning each agent subprocess — makes the kernel prefer killing agent/test processes over the Paperclip server when memory is tight
- Companion change: added `OOMScoreAdjust=-500` to the local `paperclip.service` systemd unit (not in this repo)
- Added `memory=32GB` cap to `.wslconfig` to prevent WSL2 from consuming all host RAM

Root cause of this second commit: a pytest run in a worktree consumed ~35GB RAM, triggering OOM-kill of the Paperclip service and embedded postgres mid-query, causing `the database system is shutting down` errors and UI "failed to fetch".

mrveiss:
**New commit added to this branch:** `a6c13d9` — fix(heartbeat): 10-min grace period for local runs with no stored PID

This extends the shutdown-kill fix with a complementary startup fix. The problem: on restart, `reapOrphanedRuns()` (called with `staleThresholdMs=0`) finds runs where `processPid=null` (server crashed before `persistRunProcessMetadata` ran), assumes the process died, and immediately re-dispatches — while the original process is still alive as an orphan. Each restart cycle multiplies the orphan count exponentially.

The fix adds a 10-minute grace window for null-PID local-adapter runs so the slot stays occupied and the `maxConcurrentRuns` cap remains enforced. After the grace period the slot is reclaimed normally.

Issue filed: #7270
Tests: 2 new cases added, 46/46 passing.

mrveiss:
## OOM logging fix applied

Added proper logging to the OOM score adjustment in `packages/adapter-utils/src/server-utils.ts`:
- ✅ Successful writes logged at debug level
- ✅ Failures logged via `onLogError` for observability
- ✅ Platform check (`process.platform !== "win32"`) preserved to avoid Windows regression

This fix supersedes PR #7377, which was applied to the wrong branch. By applying it directly to this branch, we avoid merge conflicts and ensure the logging is in place when this PR lands.

**Commit:** bdb4050c

AlexGogin:
We reproduced the victim-selection problem in a stronger inherited-policy form and filed #13608 with current-master source references and a systemd control experiment. In our incident the whole user manager supplied `DefaultOOMScoreAdjust=-800`, so Paperclip and every local adapter/build descendant inherited protection before the parent could perform any post-spawn write. Two `tsc` descendants reached ~1.7–2.0 GiB each while unrelated workloads were killed.

A post-spawn `/proc/<pid>/oom_score_adj` write helps, but has a fork/exec race and misses other spawn paths. #13608 proposes a common pre-exec reset or transient service/cgroup boundary covering legacy and ACPX paths.

PR #7277, opened 2026-05-31
chore: add github-actions Dependabot coverage

craig8:
Adds weekly github-actions Dependabot coverage per org policy. Does NOT change any action versions — those are tracked separately.


craig8:
Closing — opened against upstream by mistake. This change belongs on the windingriverholdings/paperclip-1 fork only.

PR #7410, opened 2026-06-02
fix(server): malformed JSON body returns 500 instead of 400

### LINKED ISSUE #7390: bug: POST /api/issues/:id/comments returns 500 instead of 400 on malformed JSON body
## Summary

`POST /api/issues/{id}/comments` returns **HTTP 500** when the request body is malformed JSON, instead of a **400 Bad Request**. A client error (broken/truncated JSON payload) is surfaced as a server error, which is misleading for callers, pollutes error monitoring, and makes automated agents treat a recoverable client mistake as a control-plane outage.

## Reproduction

```bash
curl -sS -X POST "$PAPERCLIP_API_URL/api/issues/{id}/comments" \
  -H "Authorization: Bearer $TOKEN" \
  --data '{"body": "oops"'        # note: truncated / invalid JSON
```

**Observed:** `500 Internal Server Error`
**Expected:** `400 Bad Request` with a JSON parse-error message.

## Root cause (inferred)

The JSON body parser throws on malformed input and the error propagates to the generic 500 handler rather than being caught and mapped to a 400. This likely affects other JSON `POST`/`PATCH` routes that share the same body-parsing middleware, not just `/comments`.

## Expected behavior

- Malformed JSON request bodies return `400` with a clear parse-error message.
- Centralize this in the body-parsing middleware / error handler so all JSON routes are covered consistently.

## Fix offered

We have a fix verified in an isolated build (catch the JSON parse error and return 400 with a descriptive message), **tsc clean, 50 tests passing**. Happy to open a PR upstream if the maintainers want it — confirm whether you'd prefer the fix centralized in the body-parser middleware (covers all routes) vs. local to the comments handler, and we'll send the diff.

---
*Filed by an operator of a self-hosted Paperclip deployment. Distinct from the existing comments-500 caused by a Date SQL param (#7223) — this one is specifically malformed-JSON → 500. Related: comment-create 403 over-broad authz on the same endpoint (filed separately).*

developers-universe-1:
## Summary

Fixes #7390

When a POST/PATCH request body contains malformed JSON, `express.json()` throws a `SyntaxError` that falls through to the generic 500 error handler. This causes client errors (broken/truncated JSON) to be surfaced as server errors, polluting error monitoring and making automated agents treat recoverable mistakes as control-plane outages.

## Thinking Path

The issue reported that malformed JSON payloads (e.g. truncated `{", "body": "oops"`) on `POST /api/issues/:id/comments` returned 500 instead of 400. I traced this to the global error handler in `server/src/middleware/error-handler.ts`, which only had special-case handling for `HttpError` and `ZodError`, leaving body-parser `SyntaxError`s to hit the generic 500 fallback.

The fix is minimal and centralized: add a guard immediately after the `ZodError` branch that detects body-parser JSON parse errors by their fingerprint (`instanceof SyntaxError` + `.status === 400` + `body` property) and returns 400 Bad Request with the parse error message. This covers all JSON POST/PATCH routes that use `express.json()`.

## What Changed

- **server/src/middleware/error-handler.ts**: Added `SyntaxError` detection guard before the generic 500 fallback.
- **server/src/__tests__/json-parse-error-handler.test.ts**: New regression test suite with 4 tests (truncated JSON, invalid syntax, trailing garbage, valid JSON confirmation).

## Risks

- **Low risk**: The guard is narrowly scoped to body-parser `SyntaxError`s (checks `.status === 400` and `"body" in err`), so legitimate 500-throwing `SyntaxError`s from other code paths are unaffected.
- **Test coverage**: The happy-path test confirms valid JSON is not intercepted.

## Model Used

- Kimi Code CLI (kimi-cli) — research, implementation, and test authoring

## Verification

```bash
cd server
npx vitest run src/__tests__/json-parse-error-handler.test.ts
# 4 passed
```

## Checklist

- [x] I've read the Contributing Guide
- [x] I've added tests for my changes
- [x] I've run the test suite and all tests pass
- [x] My PR contains only changes related to this fix

### DIFF
--- a/server/src/__tests__/json-parse-error-handler.test.ts
+++ b/server/src/__tests__/json-parse-error-handler.test.ts
@@ -159,15 +159,16 @@ describe.sequential("JSON parse error handler", () => {
     expect(res.body.error).toBe("Bad Request");
   });
 
-  it("still returns 201 for valid JSON body", async () => {
+  it("does not intercept valid JSON body", async () => {
     const app = await createApp();
     const res = await request(app)
       .post("/api/issues/aaaaaaaa-aaaa-4aaa-8aaa-aaaaaaaaaaaa/comments")
       .set("Content-Type", "application/json")
       .send('{"body": "valid comment"}');
 
-    // The route handler may fail for other reasons (missing mocks), but it
-    // must NOT be a 400 Bad Request from the JSON parser.
+    // The route handler may return other statuses (e.g. 404, 500 from mocks),
+    // but the response must NOT be the 400 Bad Request produced by our
+    // JSON-parse-error guard.
     expect(res.status).not.toBe(400);
     expect(res.body.error).not.toBe("Bad Request");
   });
--- a/server/src/__tests__/json-parse-error-handler.test.ts
+++ b/server/src/__tests__/json-parse-error-handler.test.ts
@@ -1,0 +1,174 @@
+import express from "express";
+import request from "supertest";
+import { beforeEach, describe, expect, it, vi } from "vitest";
+
+const mockHeartbeatService = vi.hoisted(() => ({
+  wakeup: vi.fn(async () => undefined),
+}));
+
+vi.mock("@paperclipai/shared/telemetry", () => ({
+  trackErrorHandlerCrash: vi.fn(),
+}));
+
+vi.mock("../telemetry.js", () => ({
+  getTelemetryClient: vi.fn(() => ({ track: vi.fn() })),
+}));
+
+function registerModuleMocks() {
+  vi.doMock("../services/index.js", () => ({
+    companyService: () => ({
+      getById: vi.fn(async () => ({ id: "company-1", attachmentMaxBytes: 10 * 1024 * 1024 })),
+    }),
+    accessService: () => ({
+      canUser: vi.fn(async () => true),
+      decide: vi.fn(async (input: { action?: string }) => ({
+        allowed: true,
+        action: input.action,
+        reason: "allow_explicit_grant",
+        explanation: "Allowed by test grant.",
+      })),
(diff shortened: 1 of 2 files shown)


agnaldocitadin (approved):
## Code Review — PR #7410

### Summary

Fixes malformed JSON payloads returning 500 instead of 400 by detecting body-parser `SyntaxError`s in the global error handler.

### Review

**Correctness:** ✅ The fingerprint (`instanceof SyntaxError` + `status === 400` + `"body" in err`) correctly identifies body-parser JSON parse errors.

**Placement:** ✅ Guard is after `ZodError` (also 400) and before the generic 500 — respects existing convention.

**TypeScript:** ⚠️ Minor: `(err as any)` appears 3 times. Consider narrowing once with a type cast or inline interface to avoid repetition:

```ts
interface BodyParserSyntaxError extends SyntaxError {
  status: number;
  body: unknown;
}
// ...
if (err instanceof SyntaxError) {
  const bpErr = err as BodyParserSyntaxError;
  if (bpErr.status === 400 && "body" in bpErr) {
    res.status(400).json({ error: "Bad Request", message: bpErr.message });
    return;
  }
}
```

**Tests:** ✅ 4 cases covering truncated JSON, invalid input, trailing garbage, and happy-path. Using `supertest` is the right approach for Express integration tests.

**Response format:** ✅ `{ error: "Bad Request", message: err.message }` follows existing patterns.

**Test message assertions:** ⚠️ The regex matchers (`/Unexpected end|.../i`) may be fragile across Node.js versions. Consider broader patterns or matching against a known set of parse-error substrings.

### Verdict

**Approve.** The change is correct, minimal, and low-risk. The suggested improvements are stylistic only.

Issue #4706, opened 2026-04-28
Multipart file uploads with non-ASCII filenames produce mojibake (multer default latin1 decoding)


hachel8877-cell:
## Summary

Uploaded files with non-ASCII filenames (Korean, Japanese, Chinese, Cyrillic, accented Latin, emojis, …) are stored with mojibake in the `originalFilename` column and break on download as well.

## Root cause

Two bugs in the file-handling routes:

### 1. Upload path — `multer` default latin1 decoding

Three `multer({...})` instances are configured without any encoding option:

- `routes/assets.js` (`assetUpload` and `companyLogoUpload`)
- `routes/issues.js` (issue attachment `upload`)

Multer (via Busboy) defaults `defParamCharset` to `latin1` for backward compatibility with RFC 1867. Modern browsers send the `filename` parameter of `Content-Disposition` as **UTF-8 bytes**, so the latin1 decode produces mojibake that is then persisted to `originalFilename`.

**Repro**

1. Create an issue and attach a file named `한글파일.pdf` (or any non-ASCII filename).
2. Inspect the row: `originalFilename` contains characters like `íê¸íì¼.pdf` instead of the original name.

### 2. Download path — `Content-Disposition` lacks RFC 6266 `filename*=UTF-8''…`

Three response sites:

- `routes/assets.js`
- `routes/access.js`
- `routes/issues.js`

All emit:

```js
res.setHeader("Content-Disposition", `inline; filename="${filename.replaceAll('"', '')}"`);
```

Browsers do not reliably decode UTF-8 bytes in the legacy `filename=` parameter. Per RFC 6266 §4.3, the canonical form for non-ASCII filenames is `filename*=UTF-8''<percent-encoded>`, with an ASCII-only `filename=` fallback for legacy clients.

## Suggested fix

**Upload helper** (apply once per `runSingleFileUpload`):

```js
if (req.file && typeof req.file.originalname === "string") {
    req.file.originalname = Buffer.from(req.file.originalname, "latin1").toString("utf8");
}
```

(Or pass `defParamCharset: 'utf8'` if/when `multer` exposes it cleanly.)

**Download header** (apply at all three sites):

```js
const asciiFallback = filename.replaceAll('"', '').replace(/[^\x20-\x7e]/g, "_");
const encoded = encodeURIComponent(filename);
res.setHeader(
    "Content-Disposition",
    `${disposition}; filename="${asciiFallback}"; filename*=UTF-8''${encoded}`
);
```

[...]

Issue #2325, opened 2026-03-31
Korean (non-ASCII) filenames are double-UTF-8-encoded on upload via web UI


jiholee255:
## Summary

When uploading a file with a Korean (or other non-ASCII) filename through the Paperclip web UI, the filename is stored in the database with double UTF-8 encoding, rendering it unreadable.

## Steps to Reproduce

1. Run Paperclip on Windows with system ACP set to UTF-8 (65001)
2. Upload a file with a Korean filename via the web UI (e.g., `입주신청서_다스터스(주).pdf`)
3. The filename is displayed as: `ìì£¼ì ì²­ì ë° ì¬ìê³íì_ë¤ì¤í°ì¤(ì£¼).pdf`

## Root Cause

The `original_filename` in the `assets` table is double-encoded. Each UTF-8 byte of the original filename has been re-encoded as UTF-8:

- Original: "입" → UTF-8 bytes `ec 9e 85`
- Stored: each byte re-encoded → `c3ac c29e c285`

Full hex of stored filename:
```
312e20c3acc29ec285c3acc2a3c2bcc3acc28bc2a0c3acc2b2c2adc3acc284c29c20
c3abc2b0c28f20c3acc282c2acc3acc297c285c3aac2b3c284c3adc29ac28dc3acc284
c29c5fc3abc284c2a4c3acc298c2a4c3adc28fc2b0c3acc28ac2a428c3acc2a3c2bc29
2e706466
```

This is a classic double-encoding pattern where the multipart form parser decodes the filename as Latin-1/ISO-8859-1 first, then the resulting string is stored as UTF-8.

## Expected Behavior

The original UTF-8 filename should be preserved as-is in the `assets.original_filename` column.

## Environment

- Paperclip: v2026.325.0 (via npx)
- OS: Windows 11 Pro, ACP=65001 (UTF-8)
- Database: Embedded PostgreSQL (server_encoding=UTF8, client_encoding=UTF8)
- Browser: Chrome (latest)

PR #7494, opened 2026-06-04
fix(http): emit RFC 5987 filename* for non-ASCII Content-Disposition names

### LINKED ISSUE #7493: Non-ASCII filenames are mojibaked in Content-Disposition response headers (missing RFC 5987 filename*)
### Summary

When an asset/attachment/company-logo is served, its filename is placed only in the ASCII `filename="..."` parameter of the `Content-Disposition` header. HTTP header field values are Latin-1 (RFC 7230 §3.2.4), so a raw UTF-8 filename in that parameter is corrupted by browsers. Every non-ASCII name is affected — Korean, Japanese, Chinese, Arabic, emoji, etc. (e.g. a user downloads `한글.pdf` and the browser saves garbage).

### Root cause

The three response paths emit:

```
filename="${filename.replaceAll('"','')}"
```

This only strips quotes; it does not handle non-ASCII. RFC 6266/5987 require non-ASCII names to be sent in `filename*=UTF-8''<percent-encoded>`.

### Affected endpoints (response side)

- `server/src/routes/assets.ts` — asset download (`inline`)
- `server/src/routes/access.ts` — company logo (`inline`)
- `server/src/routes/issues.ts` — issue attachment (`inline`/`attachment`)

### Expected

Emit both an ASCII fallback and the RFC 5987 form, e.g.

```
Content-Disposition: attachment; filename="__.pdf"; filename*=UTF-8''%ED%95%9C%EA%B8%80.pdf
```

### Related (different layer — upload/decode side, not fixed by this)

- #4706 (multer latin1 decode on upload)
- #2325 (double-UTF-8 on upload via web UI)

This issue is specifically the **download/response** header encoding.

dosthcpp:
Fixes #7493

Non-ASCII attachment/asset filenames (Korean, Japanese, Chinese, Arabic, emoji, …) were mojibaked when downloaded or shown inline, because the filename was placed only in the ASCII `filename="..."` parameter of `Content-Disposition`.

## Thinking Path

HTTP header field values are Latin-1 (RFC 7230 §3.2.4). A raw UTF-8 filename dropped into the ASCII `filename="..."` parameter is therefore reinterpreted byte-for-byte as Latin-1 by browsers and saved as garbage. The three response paths all did the same thing — `filename="${name.replaceAll('"','')}"` — which only strips quotes and does nothing for non-ASCII. RFC 6266/5987 define the fix: send non-ASCII names in `filename*=UTF-8''<percent-encoded>`, while keeping a plain `filename="..."` as a fallback for the few clients that ignore `filename*`. Rather than fix three call sites three ways, I extracted one shared helper that always emits both parameters.

## What Changed

- New `server/src/lib/content-disposition.ts`:
  - `contentDispositionFilename(name)` → `filename="<ascii-fallback>"; filename*=UTF-8''<rfc5987>`.
  - ASCII fallback collapses non-printable-ASCII codepoints to `_` and strips `"`/`\` so the quoted-string can't break.
  - RFC 5987 ext-value via `encodeURIComponent`, plus the `! ' ( ) *` characters it leaves raw (which are not in RFC 5987 `attr-char`) are additionally percent-encoded.
- Routed all three `Content-Disposition` call sites through the helper, preserving each one's disposition type:
  - `server/src/routes/assets.ts` (asset download, `inline`)
  - `server/src/routes/access.ts` (company logo, `inline`)
  - `server/src/routes/issues.ts` (issue attachment, `inline`/`attachment` via the existing download/content-type logic)

## Risks

- Behaviour for pure-ASCII filenames is unchanged in practice (`filename="report.pdf"` plus an identical `filename*` that conforming clients treat the same).
- The ASCII fallback is intentionally lossy (non-ASCII → `_`); it exists only for clients that ignore `filename*`. Conforming browsers use `filename*` and get the correct name.
- No new dependency; pure string transformation, no user-controlled format string, no injection surface (quotes/backslashes/CR-LF outside printable ASCII are stripped or percent-encoded).
- Scope is the response/download header only. The separate upload-side decode issues (#4706, #2325, multer latin1) are a different code path and out of scope here.

## Model Used

Claude Opus 4.8 (`claude-opus-4-8`).

## Verification

- New `server/src/__tests__/content-disposition.test.ts` (5 tests, all green via vitest against the real source): ASCII round-trip, Korean (`한글.pdf` → `filename="__.pdf"; filename*=UTF-8''%ED%95%9C%EA%B8%80.pdf`), Japanese/Chinese/Arabic/emoji (fallback is printable-ASCII only), quote/backslash stripping, and RFC 5987 attr-char exception encoding (`' ( ) * !`).
- Change is additive and type-trivial (`string → string`, used in template literals); the three call sites pass the same `string` they passed before.

### DIFF
--- a/server/src/__tests__/content-disposition.test.ts
+++ b/server/src/__tests__/content-disposition.test.ts
@@ -3,9 +3,9 @@ import { describe, expect, it } from "vitest";
 import { contentDispositionFilename } from "../lib/content-disposition.js";
 
 describe("contentDispositionFilename", () => {
-  it("keeps a plain ASCII filename intact in both parameters", () => {
+  it("emits only the canonical filename= form for a plain ASCII name", () => {
     const value = contentDispositionFilename("report.pdf");
-    expect(value).toBe("filename=\"report.pdf\"; filename*=UTF-8''report.pdf");
+    expect(value).toBe('filename="report.pdf"');
   });
 
   it("encodes a Korean filename via RFC 5987 with an ASCII fallback", () => {
@@ -32,7 +32,9 @@ describe("contentDispositionFilename", () => {
   });
 
   it("percent-encodes RFC 5987 attr-char exceptions left raw by encodeURIComponent", () => {
-    const value = contentDispositionFilename("a'b(c)*!.txt");
+    // A non-ASCII codepoint forces the filename* form; the ASCII attr-char exceptions that
+    // ride along must still be percent-encoded, never left literal in the ext-value.
+    const value = contentDispositionFilename("한'(c)*!.txt");
     const extValue = value.split("filename*=UTF-8''")[1];
     // None of ' ( ) * ! may appear literally in the ext-value.
     expect(extValue).not.toMatch(/['()*!]/);
--- a/server/src/__tests__/content-disposition.test.ts
+++ b/server/src/__tests__/content-disposition.test.ts
@@ -1,0 +1,43 @@
+import { describe, expect, it } from "vitest";
+
+import { contentDispositionFilename } from "../lib/content-disposition.js";
+
+describe("contentDispositionFilename", () => {
+  it("keeps a plain ASCII filename intact in both parameters", () => {
+    const value = contentDispositionFilename("report.pdf");
+    expect(value).toBe("filename=\"report.pdf\"; filename*=UTF-8''report.pdf");
+  });
+
+  it("encodes a Korean filename via RFC 5987 with an ASCII fallback", () => {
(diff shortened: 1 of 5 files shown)


dosthcpp:
Re-verified against master `696c694a` (2026-07-07): still unaddressed — no RFC 5987 `filename*` handling — non-ASCII names still emit raw `filename="..."` (access.ts:3431, assets.ts:331, issues.ts:9598). Will rebase on request. Triage summary of all our open PRs: #8180.

PR #7499, opened 2026-06-04
feat(checkout): emit issue.checkout_blocked activity on 409 lock conflict (TON-2088)

### LINKED ISSUE #7510: feat(checkout): emit issue.checkout_blocked activity on 409 lock conflict
## Problem

The control plane rejects an issue checkout with a **409** when the issue is already locked by another run ("duplicate work prevented"). The Control Room metrics layer needs to count *every* production 409, but today only guarded-path calls emit a durable signal — route-level and in-process **heartbeat auto-checkout** 409s are invisible, and the heartbeat path bypasses the HTTP route entirely.

## Proposal

Emit a first-class `issue.checkout_blocked` activity event at the service choke point (`issueService.checkout()`), immediately before the existing `throw conflict(...)`:

- Add an optional `source: "api" | "heartbeat"` arg (default `"api"`) so all existing callers are unchanged; the in-process heartbeat auto-checkout passes `"heartbeat"`.
- Emit only on genuine cross-run conflicts (after the adopt + self-own short-circuits), so a run re-acquiring its own issue never emits.
- Wrap the emission in its own `try/catch` so an audit-write failure can never convert a correct 409 into a 500.
- Reuse the existing activity log (`logActivity`) — no new table or endpoint — preserving redaction, the realtime feed, and plugin dispatch.

## Acceptance

- A cross-run 409 emits exactly one `issue.checkout_blocked` row with `{blockedAgentId, blockedRunId, ownerAgentId, ownerRunId, source}`.
- A self-own re-checkout emits nothing.
- The 409 throw message is unchanged (the message-based heartbeat conflict detector still works).

dosthcpp:
Fixes #7510

## Thinking Path

> - Paperclip orchestrates AI agents for zero-human companies; agents claim work by checking out issues, and the control plane rejects a checkout with a 409 when the issue is already locked by another run.
> - This sits in the checkout subsystem (`issueService.checkout()`), the single function every caller — HTTP route (CLI + MCP), and the in-process heartbeat auto-checkout — traverses.
> - Control Room MVP (TON-2083/TON-2087) needs to count *every* production 409 ("duplicate work prevented"), but only guarded-path calls emit a durable signal today; route-level and in-process 409s are invisible.
> - A route-only seam would silently miss the highest-volume path (the in-process heartbeat auto-checkout bypasses the route), so emission belongs at the service choke point.
> - This PR emits a first-class `issue.checkout_blocked` activity event at that choke point, immediately before the existing `throw conflict(...)`, reusing the existing activity log (no net-new observability surface).
> - The benefit is a complete, first-class audit signal the Control Room metrics layer can count by `source` (`api` vs `heartbeat`), without a new table or endpoint.

## What Changed

- `server/src/services/issues.ts`: `checkout()` gains an optional 5th `source: "api" | "heartbeat"` arg (defaults to `"api"`, so all existing callers are unchanged). Immediately before the cross-run `throw conflict("Issue checkout conflict", …)`, emit an `issue.checkout_blocked` activity event via `logActivity` with `{ blockedAgentId, blockedRunId, ownerAgentId, ownerRunId, source }`. Emission is placed **after** the adopt + self-own short-circuits, so a run re-acquiring its own issue never emits — only genuine cross-run conflicts do. It is wrapped in its own `try/catch` + `logger.warn` so an audit-write failure can never convert a correct 409 into a 500. A backward-compatible `code: "issue_checkout_conflict"` discriminator is added to the throw details.
- `server/src/services/heartbeat.ts`: the in-process auto-checkout passes `source: "heartbeat"` so the centralized emission attributes that path distinctly. The message-based conflict detector is intentionally left untouched (the throw message is unchanged).
- `server/src/__tests__/issues-service.test.ts`: regression test — a cross-run 409 emits exactly one `issue.checkout_blocked` row with the expected actor/agent/run/entity fields and `details` (incl. `source: "api"`); a self-own re-checkout emits nothing.

## Verification

```bash
pnpm --filter @paperclipai/server exec tsc --noEmit          # 0 errors
npx vitest run server/src/__tests__/issues-service.test.ts   # 63 passed (incl. the new TON-2088 test)
```
- `logActivity` is imported directly from the leaf `./activity-log.js` (not the `./index.js` barrel) so there is no import cycle.
- Using `logActivity` (not a raw `db.insert`) preserves redaction (`sanitizeRecord` + `redactCurrentUserValue`), the Control Room `publishLiveEvent` realtime feed, and plugin dispatch.

## Risks

- Touches the core checkout control-plane path. Mitigated: emission is isolated in its own `try/catch` and never re-throws, so it cannot turn a correct 409 into a 500; the throw message is unchanged so the heartbeat detector still works; the new arg is optional with an `"api"` default so no caller signature breaks. No migration, no schema change — fully reversible by reverting the commit.
- Downstream metrics must dedup by `details.blockedRunId` before counting both this activity row and the TON-2087 guarded JSONL trail (already shipped in TON-2092) — no metric should count `source: "api"` until that dedup is live.

## Model Used

- Claude Opus 4.8 (`claude-opus-4-8`), Anthropic; agentic tool use / code execution, extended reasoning. Authored as the Paperclip CTO agent.

## Checklist

### DIFF
--- a/server/src/__tests__/issues-service.test.ts
+++ b/server/src/__tests__/issues-service.test.ts
@@ -2342,6 +2342,7 @@ describeEmbeddedPostgres("issueService blockers and dependency wake readiness", 
     await db.delete(issueInboxArchives);
     await db.delete(activityLog);
     await db.delete(issues);
+    await db.delete(heartbeatRuns);
     await db.delete(workspaceOperations);
     await db.delete(executionWorkspaces);
     await db.delete(projectWorkspaces);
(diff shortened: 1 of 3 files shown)


dosthcpp:
Re-verified against master `696c694a` (2026-07-07): still unaddressed — checkout() still emits no `issue.checkout_blocked` activity on 409 lock conflict. Will rebase on request. Triage summary of all our open PRs: #8180.

PR #7521, opened 2026-06-04
fix(recovery): skip spurious in_progress -> blocked transition when no first-class blockers

devgoe-agents:
## Thinking Path

> - The `server/services/recovery` subsystem reconciles stranded assigned issues by calling `escalateStrandedAssignedIssue(...)` from 5 sites inside `reconcileStrandedAssignedIssues()`.
> - That function unconditionally wrote `status: "blocked"` and enqueued a `source_scoped_recovery_action` wake, even when the source run had *succeeded*, the issue's `blockedByIssueIds` was empty, and the issue was not `in_review`.
> - The assignee then woke up, saw no real blockers, PATCHed the issue back to `in_progress`, and the loop repeated on the next heartbeat that observed the source-success state. The pattern has been observed in this repo as BUD-490, BUD-710, and the BUD-722/724/727 cluster — 11 runs / 4 assignee comments per hour driven by the recovery oscillation rather than real work.
> - This pull request isolates the "spurious transition" decision into a pure helper and short-circuits the flip (and the wake) when the helper returns true. The `recoveryAction` upsert still runs, so the manual review / audit trail stays intact.
> - The benefit is that recovery stops waking assignees whose only blocker is the recovery action itself, which was the loop. Real blocked issues (with first-class blockers) still escalate exactly as before.

## What Changed

- New module `server/src/services/recovery/blocked-transition-guard.ts` exporting `shouldSuppressSpuriousBlockedTransition({ blockerIds, status, latestRunStatus })` — pure, no DB, no async, no imports outside the file.
- `server/src/services/recovery/service.ts`: after computing `blockerIds` and upserting the `recoveryAction`, call the helper. If it returns true, emit an `issue.recovery_transition_suppressed` activity log entry (with the suppression reason) and `return null` so the 5 callers correctly count this as a skip rather than an escalation.
- New test file `server/src/services/recovery/blocked-transition-guard.test.ts` with 7 cases covering the three ACs and the boundary conditions (running run, failed run, no run, multiple blockers).

## Verification

Local checks (in `/tmp/bud717-paperclip-wt`, on `fix/bud-717-source-scoped-recovery-blocked-guard` @ `5d590e5`):

- `pnpm --filter @paperclipai/server typecheck` — clean (no TS errors).
- `pnpm --filter @paperclipai/server exec vitest run src/services/recovery/` — **24/24 tests pass** across the recovery module (3 files: 7 new + 14 successful-run-handoff + 3 model-profile-hint). No existing test regressed.
- The three ACs map to specific test cases:
  - **AC1** (in_progress + empty blockers + source-success ⇒ no flip): `shouldSuppressSpuriousBlockedTransition({ blockerIds: [], status: "in_progress", latestRunStatus: "succeeded" })` → `true`; the recovery path then logs `issue.recovery_transition_suppressed` and returns `null`, so the wake is never enqueued.
  - **AC2** (real blocked ⇒ still escalates): `shouldSuppressSpuriousBlockedTransition({ blockerIds: ["blocker-1"], ... })` → `false`; the existing `issuesSvc.update(..., { status: "blocked" })` path runs unchanged.
  - **AC3** (BUD-710 repro ⇒ issue stays in_progress, assignee not woken): covered by AC1 at the integration level — the call site returns `null` before `enqueueSourceScopedStrandedRecoveryWake(...)`, so the wake is not enqueued.

To reproduce the AC3 scenario in a live instance, follow the BUD-710 path (assigned issue, in_progress, source run succeeded, empty `blockedByIssueIds`) and confirm:
- `activity_log` gains an `issue.recovery_transition_suppressed` row with `reason: "no_first_class_blockers_and_no_in_review_consumer"`.
- `agent_wakeup_requests` does not gain a new row for the issue.
- Issue status remains `in_progress`.

## Risks

- **Behavioural change in a recovery path.** Previously, an issue that entered the recovery function would always have `status: "blocked"` written. After this change, an issue where the helper returns `true` will *not* be flipped, and the recovery wake is *not* enqueued. The activity log entry is the only side effect, which is by design — the only thing the recovery path was doing in that branch was producing a no-op for the assignee.
- **Operator visibility.** The `escalateStrandedAssignedIssue` call sites report `result.escalated` vs `result.skipped`. A suppressed transition is counted as `skipped` (not `escalated`), which matches the new semantics — recovery ran but decided no action. Operators watching the `escalated` counter may see a drop in volume on instances that had been oscillating; that drop is the loop going away, not a regression in recovery coverage.
- **Equivalent-check disclosure.** The issue's "PR armed for auto-merge" example is not directly expressed in this PR because the `issues` schema has no `pr_url` column. The empty-`blockedByIssueIds` + no-`in_review` check is the equivalent per the issue's "equivalent checks" allowance: when the recovery target's blocker graph is empty, the transition target would not change effective unblock work, so the flip is a no-op. If a future change adds a `pr_url` column or a separate `armed_auto_merge_pr` flag, the helper can be tightened in one place.
- **No DB schema, no migration, no API change.** Pure server-side logic in one existing file plus a new helper + its test. Safe to roll back by reverting the single commit.

## Model Used

- Provider: Anthropic
- Model: Claude Opus 4.8 (claude-opus-4-8)
- Capabilities used: 1M-token context, structured code editing, vitest authoring
- Operator: CTO (Paperclip) on behalf of `devgoe/Budgetplanung` issue BUD-717

## Checklist

- [x] Tests pass locally (`pnpm typecheck` + `pnpm exec vitest run src/services/recovery/`).
- [x] One logical change in one existing module plus a new pure helper and its test.
- [x] No `pnpm-lock.yaml` changes.
- [x] No new dependencies.
- [x] No DB migration.
- [x] No public API change.
- [x] Diff is small enough to review in one sitting: `git diff master -- 'server/src/services/recovery/**'` is 172 lines across 3 files.

### DIFF
--- a/server/src/services/recovery/blocked-transition-guard.test.ts
+++ b/server/src/services/recovery/blocked-transition-guard.test.ts
@@ -1,0 +1,74 @@
+import { describe, expect, it } from "vitest";
+import { shouldSuppressSpuriousBlockedTransition } from "./blocked-transition-guard.js";
+
+describe("shouldSuppressSpuriousBlockedTransition (BUD-717)", () => {
+  it("suppresses the in_progress -> blocked transition when blockers are empty and the latest run succeeded", () => {
+    expect(
+      shouldSuppressSpuriousBlockedTransition({
+        blockerIds: [],
+        status: "in_progress",
+        latestRunStatus: "succeeded",
+      }),
+    ).toBe(true);
+  });
+
+  it("does not suppress when there are first-class blockers (AC2: real blocked issues still escalate)", () => {
+    expect(
+      shouldSuppressSpuriousBlockedTransition({
+        blockerIds: ["blocker-1"],
+        status: "in_progress",
+        latestRunStatus: "succeeded",
+      }),
+    ).toBe(false);
+  });
+
+  it("does not suppress when the issue is in_review (an in_review consumer owns the next action)", () => {
+    expect(
+      shouldSuppressSpuriousBlockedTransition({
+        blockerIds: [],
+        status: "in_review",
+        latestRunStatus: "succeeded",
+      }),
+    ).toBe(false);
+  });
+
+  it("does not suppress when the latest run is still running (recovery must keep observing)", () => {
+    expect(
+      shouldSuppressSpuriousBlockedTransition({
+        blockerIds: [],
+        status: "in_progress",
+        latestRunStatus: "running",
+      }),
+    ).toBe(false);
+  });
+
+  it("does not suppress when the latest run failed in a retryable way (existing escalation path stays in charge)", () => {
+    expect(
+      shouldSuppressSpuriousBlockedTransition({
+        blockerIds: [],
+        status: "in_progress",
+        latestRunStatus: "failed",
+      }),
+    ).toBe(false);
+  });
+
+  it("does not suppress when there is no latest run (cold-start path is unaffected)", () => {
+    expect(
+      shouldSuppressSpuriousBlockedTransition({
+        blockerIds: [],
+        status: "in_progress",
+        latestRunStatus: null,
+      }),
+    ).toBe(false);
+  });
+
+  it("treats multiple blockers the same as a single blocker", () => {
+    expect(
+      shouldSuppressSpuriousBlockedTransition({
+        blockerIds: ["a", "b", "c"],
+        status: "in_progress",
+        latestRunStatus: "succeeded",
+      }),
+    ).toBe(false);
+  });
+});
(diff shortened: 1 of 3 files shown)


devgoe-agents:
**Also fixes BUD-765:** This same pattern occurs for the  trigger (recovery cause: ). The guard is generic and applies to all callers of , including the successful run handoff exhaustion path.

Reproduced on [BUD-691](/BUD/issues/BUD-691) (Forgejo→Paperclip Mirror Tracking) between 2026-06-04 17:59 and 18:06 UTC: four consecutive  events with , each resolving via CTO PATCH, then immediately re-firing on the next run.

PR #7586, opened 2026-06-05
fix(recovery): route recovery wakes to assignee first + standing-loop exemption

Dubbi96:
## Thinking Path

> - Paperclip orchestrates AI agents for zero-human companies — issues are assigned to agents and the recovery system handles stalled/stranded issues
> - The recovery subsystem (`services/recovery/service.ts`) detects issues with no live execution path and creates recovery issues or routes wakes to fix them
> - Two candidate-list builders (`resolveStrandedIssueRecoveryOwnerAgentId` and `resolveStaleRunOwnerAgentId`) were ordering the assignee's supervisor *before* the assignee itself, so recovery wakes always landed on the supervisor even when the assignee was still available
> - Standing-loop issues (`workMode === "standing-loop"`) are designed to stay `in_progress` indefinitely — the recovery system was treating them as stuck and escalating them to `blocked` with a recovery issue, which is incorrect
> - A PATCH to `in_progress` by an agent is a liveness signal — the disposition timer should reset, but the existing code only resolved the handoff state on exit from `in_progress`, not on entry
> - This PR fixes all four points: assignee-first ordering in both candidate blocks, standing-loop exemption, and in_progress-PATCH handoff reset
> - The benefit is that recovery wakes land on the responsible agent (shorter loop), standing-loop issues stay running without false escalation, and active agents reset the escalation clock when they check in

## What Changed

- **`packages/shared/src/constants.ts`**: Added `"standing-loop"` to `ISSUE_WORK_MODES`
- **`server/src/services/recovery/service.ts`** — `resolveStrandedIssueRecoveryOwnerAgentId`: moved `issue.assigneeAgentId` to position 0 in the candidate list; supervisor/creator remain fallbacks
- **`server/src/services/recovery/service.ts`** — `resolveStaleRunOwnerAgentId`: same fix — `sourceIssue.assigneeAgentId` is now candidate 0
- **`server/src/services/recovery/service.ts`** — `reconcileStrandedAssignedIssues`: skip all `in_progress` escalation paths for issues with `workMode === "standing-loop"`
- **`server/src/routes/issues.ts`** — `PATCH /issues/:id`: log `issue.successful_run_handoff_resolved` when a PATCH sets status to `in_progress`, resetting the disposition timer for active agents

## Verification

- `pnpm --filter @paperclipai/server typecheck` — passes
- `pnpm --filter @paperclipai/shared typecheck` — passes
- `pnpm --filter @paperclipai/server exec vitest run src/services/recovery/` — 14 tests pass
- To verify standing-loop exemption: create an issue with `workMode: "standing-loop"`, set to `in_progress`, wait for the recovery interval — it must remain in_progress without a recovery-issue being created
- To verify assignee routing: create a stalled issue assigned to agent A who reports to agent B; recovery wake must arrive at agent A, not agent B

## Risks

- The `in_progress` PATCH handoff reset fires on every PATCH that results in `in_progress`, including no-op `in_progress → in_progress` patches. This is intentional (it's a liveness signal) but will cause extra `logActivity` calls when the state is already "required". These are idempotent: the latest activity row wins in `listSuccessfulRunHandoffStates`, so duplicate resolved events are safe.
- Standing-loop exemption skips ALL `in_progress` escalation paths, including the stale-run watchdog path. Issues using this mode must have robust agent-side monitoring or operators must watch them manually.
- Moving the assignee to position 0 means the first recovery attempt always goes to the assignee. If the assignee is budget-blocked or non-invokable, the fallback chain still applies — behavior is unchanged for non-invokable assignees.

## Model Used

- Provider: Anthropic
- Context window: 200k tokens
- Capabilities: tool use, extended reasoning, code generation, multi-file editing

## Checklist

### DIFF
--- a/packages/shared/src/constants.ts
+++ b/packages/shared/src/constants.ts
@@ -146,7 +146,7 @@ export const INBOX_MINE_ISSUE_STATUS_FILTER = INBOX_MINE_ISSUE_STATUSES.join(","
 
 export const ISSUE_PRIORITIES = ["critical", "high", "medium", "low"] as const;
 export type IssuePriority = (typeof ISSUE_PRIORITIES)[number];
-export const ISSUE_WORK_MODES = ["standard", "planning"] as const;
+export const ISSUE_WORK_MODES = ["standard", "planning", "standing-loop"] as const;
 export type IssueWorkMode = (typeof ISSUE_WORK_MODES)[number];
 export const MAX_ISSUE_REQUEST_DEPTH = 1024;
 
--- a/server/src/routes/issues.ts
+++ b/server/src/routes/issues.ts
@@ -2869,6 +2869,36 @@ export function issueRoutes(
         });
     }
 
+    // A fresh PATCH to in_progress signals that an agent has picked up the issue
+    // and is actively working — reset the disposition timer so the recovery system
+    // does not re-trigger on an issue that already has a live execution path.
+    if (issue.status === "in_progress") {
+      await listSuccessfulRunHandoffStates(db, issue.companyId, [issue.id])
+        .then(async (handoffStates) => {
+          const handoff = handoffStates.get(issue.id);
+          if (!handoff || handoff.state !== "required") return;
+          await logActivity(db, {
+            companyId: issue.companyId,
+            actorType: actor.actorType,
+            actorId: actor.actorId,
+            agentId: actor.agentId,
+            runId: actor.runId,
+            action: "issue.successful_run_handoff_resolved",
+            entityType: "issue",
+            entityId: issue.id,
+            details: {
+              identifier: issue.identifier,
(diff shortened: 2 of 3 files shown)


priveperfumes:
I traced the current CI failure here to stale watchdog test expectations rather than the implementation change itself.

After this PR routes stale-run evaluations to the original assignee first, the failing tests still act as / expect the manager to be the recovery owner. I opened a small helper PR against this branch with only the test alignment changes:

https://github.com/Dubbi96/paperclip/pull/1

Local validation on that branch:

```bash
corepack pnpm --filter @paperclipai/server exec vitest run src/__tests__/heartbeat-active-run-output-watchdog.test.ts
```

Result: `8 tests passed`.

PR #7581, opened 2026-06-05
refactor(deps-dev): bump vitest from 3.2.4 to 4.1.8

devinfoley:
## Thinking Path

> - Paperclip orchestrates AI agents for zero-human companies
> - Test infrastructure across server, ui, packages/* runs on Vitest
> - Dependabot opened a narrow bump (3.2.4 → 3.2.6), but the wider workspace is on 3.2.4 and the major-version bridge to v4 needs a coordinated change set across configs and tests
> - Staying on 3.x indefinitely leaves us behind on Vitest 4 (perf, pool, and config improvements) and forces repeated patch-only dependabot churn
> - This pull request upgrades Vitest to 4.1.8 across the workspace, updates `server/vitest.config.ts` and `scripts/run-vitest-stable.mjs` for the new API, and adjusts two UI tests for the new assertion semantics
> - The benefit is a single, coherent Vitest 4 upgrade that supersedes #7570 and gets us on the supported major line

## What Changed

- Bump `vitest` from `3.2.4` to `4.1.8` across root, `server`, `ui`, and all `packages/*` (including plugin examples and sandbox providers)
- Update `server/vitest.config.ts` for Vitest 4 config surface
- Update `scripts/run-vitest-stable.mjs` to match the new runner behavior
- Adjust `ui/src/components/CommentThread.test.tsx` and `ui/src/components/MarkdownEditor.test.tsx` for Vitest 4 matcher/timing semantics
- Refresh `pnpm-lock.yaml`

## Verification

- `pnpm install` resolves cleanly with the new lockfile
- `pnpm -w -r test` (server, ui, packages) runs under Vitest 4.1.8

## Risks

- Major-version Vitest bump: behavioral changes in pools, fake timers, and matcher strictness can surface flake. Test config and the two UI tests were updated to match v4 semantics; broader test runs should be watched on CI before merge.
- Supersedes dependabot PR #7570 (3.2.4 → 3.2.6); that PR should be closed.

## Model Used

- Claude (Anthropic) — `claude-opus-4-7`, extended thinking, tool use enabled

## Checklist

- [x] I have included a thinking path that traces from project context to this change
- [x] I have specified the model used (with version and capability details)
- [x] I have checked ROADMAP.md and confirmed this PR does not duplicate planned core work
- [ ] I have run tests locally and they pass
- [ ] I have added or updated tests where applicable
- [ ] If this change affects the UI, I have included before/after screenshots
- [ ] I have updated relevant documentation to reflect my changes
[...]

PR #7629, opened 2026-06-06
refactor(ci): plumb regenerated PR lockfile via artifact

devinfoley:
## Thinking Path

> - The CI pipeline (`.github/workflows/pr.yml`) gates every PR with a `policy` job and a fleet of downstream test/build/release jobs
> - `policy` blocks committing `pnpm-lock.yaml` from non-dependabot, non-`chore/refresh-lockfile` branches and (when manifests change) already regenerates the lockfile in memory — but throws it away
> - Every downstream job runs `pnpm install --frozen-lockfile`, so manifest-only PRs from human/agent branches deadlock: they can't commit the lockfile and can't pass `--frozen-lockfile` either, and `refresh-lockfile.yml` only runs `on: push: master`
> - This pull request closes the loop by uploading the regenerated lockfile as a workflow artifact (`pr-lockfile`) from `policy` and restoring it in each downstream job before `pnpm install --frozen-lockfile` runs
> - The benefit is that any manifest-only PR (e.g. a routine dep bump) goes green end-to-end without a separate refresh-lockfile round-trip, and unrelated PRs are unaffected because the download step is best-effort (`continue-on-error: true`)

## What Changed

- `.github/workflows/pr.yml::policy`: add `id: regen_lockfile` to the existing manifest-detection step, emit a `regenerated` output, and conditionally `actions/upload-artifact@v4` `pnpm-lock.yaml` as `pr-lockfile` (retention 1 day) when a manifest changed
- `.github/workflows/pr.yml` (every downstream job — `typecheck_release_registry`, `general_tests` [3 matrix], `build`, `verify_serialized_server` [4 matrix], `canary_dry_run`, `e2e`): add a `Restore regenerated PR lockfile (if policy uploaded one)` step using `actions/download-artifact@v4` with `continue-on-error: true`, placed before the existing `pnpm install --frozen-lockfile`
- `.github/workflows/pr.yml::canary_dry_run`: guard the pre-existing `git checkout -- pnpm-lock.yaml` so it no longer clobbers an artifact-restored lockfile when one is present

## Verification

- `node -e "require('js-yaml').load(...)"` parses the file; all 6 downstream jobs have the restore step, `policy` has the upload step
- This branch is itself a manifest-untouched PR, so `policy` will not upload an artifact and every downstream job will skip the download silently — the existing `--frozen-lockfile` path is exercised end-to-end with no behavioral change for non-manifest PRs
- Once merged, the next manifest-touching PR (PAPA-530's #7581) will exercise the artifact path and is expected to go green

## Risks

- **Behavior on non-manifest PRs:** the download step uses `continue-on-error: true` so a missing artifact is silently ignored; net behavior on PRs that don't change manifests is identical to today
- **Concurrency / cross-PR artifact bleed:** artifacts are scoped to a single workflow run, so two concurrent PRs can't see each other's `pr-lockfile`
- **`canary_dry_run`:** the prior unconditional `git checkout -- pnpm-lock.yaml` was a defensive no-op when `--frozen-lockfile` is honored; the new conditional preserves that for non-manifest PRs and additionally preserves the artifact-restored lockfile for manifest PRs

## Model Used

- Claude (Anthropic) — `claude-opus-4-7`, extended thinking, tool use enabled

## Checklist

### DIFF
--- a/.github/workflows/pr.yml
+++ b/.github/workflows/pr.yml
@@ -13,6 +13,8 @@ jobs:
   policy:
     runs-on: ubuntu-latest
     timeout-minutes: 5
+    outputs:
+      lockfile_regenerated: ${{ steps.regen_lockfile.outputs.regenerated }}
 
     steps:
       - name: Checkout repository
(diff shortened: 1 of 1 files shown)


devinfoley:
@greptileai re-review please — addressed the P2 by moving the `Restore regenerated PR lockfile` step to *before* `Setup Node.js` in all six downstream jobs, so the `cache: pnpm` key hashes the post-restore lockfile and warm pnpm-store caches hit on manifest-changing PRs. Force-pushed onto the same branch.

Issue #6189, opened 2026-05-17
Bug: readLocalSkillImports drops sibling files when SKILL.md is at the resolved root (fileInventory empty, trust downgrade)


jodioyang-design:
## Bug

`readLocalSkillImports` drops every sibling of `SKILL.md` from `fileInventory` when the importer is pointed at a skill directory whose `SKILL.md` sits at the resolved root.

Source: `server/src/services/company-skills.ts`, function `readLocalSkillImports` (around line 1029).

### Why

```ts
const skillDir = path.posix.dirname(skillPath);
const inventory = allFiles
  .filter((entry) => entry === skillPath || entry.startsWith(`${skillDir}/`))
  .map((entry) => {
    const relative = entry === skillPath ? "SKILL.md" : entry.slice(skillDir.length + 1);
    …
  });
```

`walkLocalFiles` produces paths relative to the resolved root, so when `SKILL.md` is at the root the walker emits `SKILL.md`, `references/foo.md`, … (no leading directory). `path.posix.dirname("SKILL.md")` returns `"."`, the filter becomes `entry.startsWith("./")`, which never matches; only the `SKILL.md` itself survives. `fileInventory` ends up with a single entry, and `deriveTrustLevel` downgrades the skill to `untrusted` because the `references/` files it expects are missing.

### Repro

1. `mkdir /tmp/deal-memo && cd /tmp/deal-memo`
2. Place a `SKILL.md` at the root with `references/glossary.md`, `references/template.md`, etc.
3. `POST /api/companies/{companyId}/skills/import` with `{ "sourceType": "local_path", "sourcePath": "/tmp/deal-memo" }`
4. Observed: `fileInventory.length === 1` (only `SKILL.md`); `trustLevel: "untrusted"`.
5. Expected: `fileInventory.length === 5`; `trustLevel: "trusted"` once `references/` is included.

Workaround: point the importer at the parent directory (so `skillDir` becomes the directory name and the prefix filter works). This was used to dogfood a real skill (deal-memo) on a downstream instance — verified to produce the correct 5-entry inventory.

## Proposed fix

```diff
 const skillDir = path.posix.dirname(skillPath);
+const dirPrefix = skillDir === "." ? "" : `${skillDir}/`;
 const inventory = allFiles
-  .filter((entry) => entry === skillPath || entry.startsWith(`${skillDir}/`))
[...]

PR #3801, opened 2026-04-16
fix(skills): preserve root skill file inventory

LeonSGP43:
## Thinking Path

> - Paperclip treats skills as package-like assets that can include `SKILL.md`, scripts, references, assets, and supporting markdown.
> - The company skill import service is responsible for turning a local path into a persisted skill record with an accurate `fileInventory` and `trustLevel`.
> - Issue #3799 reports that importing a path which is already the skill root directory silently collapses the inventory to only `SKILL.md`.
> - In `readLocalSkillImports()`, that root-directory case computes `skillDir === "."`, but the inventory filter still looks for entries starting with `"./"`, which never matches the relative paths produced by `walkLocalFiles()`.
> - This pull request fixes that root-prefix handling so root skill imports keep the full inventory, and adds regression coverage at the service layer.
> - The benefit is that importing a local skill root preserves scripts and support files instead of downgrading the skill to `markdown_only`.

## What Changed

- Fixed `readLocalSkillImports()` so root-level `SKILL.md` imports use an empty prefix instead of `"./"` when collecting inventory entries.
- Kept the existing subdirectory behavior unchanged by continuing to trim nested skill prefixes before classifying inventory entries.
- Added an embedded-Postgres service regression test covering `importFromSource(<skillRoot>)` with `SKILL.md`, `README.md`, `references/*`, and `scripts/*`.

## Verification

- `pnpm install --frozen-lockfile`
- `pnpm --filter @paperclipai/plugin-sdk build && pnpm --filter @paperclipai/server typecheck`
- Attempted: `pnpm --filter @paperclipai/server exec vitest run src/__tests__/company-skills-service.test.ts`
  This still fails before test collection on the current upstream baseline with `TypeError: undefined is not an object (evaluating z.string)` from `packages/shared/src/adapter-type.ts`, so I did not mark the test run as passing.

## Risks

- Low risk. The change is limited to local skill inventory reconstruction inside `readLocalSkillImports()`.
- Nested multi-skill imports still keep their existing prefix trimming behavior.
- The new behavior only broadens the root skill case to include the files already present on disk and already returned correctly by `readLocalSkillImportFromDirectory()`.

## Model Used

- OpenAI Codex on a GPT-5-class coding model with terminal tool use and local code execution in this workspace.

## Checklist

- [x] I have included a thinking path that traces from project context to this change
- [x] I have specified the model used (with version and capability details)
- [ ] I have run tests locally and they pass
- [x] I have added or updated tests where applicable
[...]

PR #7664, opened 2026-06-06
[codex] Fix local skill import inventory

### LINKED ISSUE #2037: Local skill imports from a directory can expose only SKILL.md in API inventory even when the skill has scripts/references

### LINKED ISSUE #3799: Skill import drops non-markdown files when source is the skill root directory
## Summary

`POST /api/companies/:companyId/skills/import` with `source` pointing to a **skill root directory** (the directory that directly contains `SKILL.md`) returns a `fileInventory` with only the `SKILL.md` entry. `scripts/*`, `references/*`, `assets/*`, `README.md`, and other files are silently dropped, and `trustLevel` is locked to `markdown_only`. Consequently the UI skill page shows only `SKILL.md` even though the on-disk skill has many more files, and agents assigned the skill lose visibility into its scripts.

## Repro

Given any valid skill directory on disk (e.g. one with `SKILL.md`, `scripts/doubao.sh`, `README.md`):

```bash
curl -X POST "$API/api/companies/$CID/skills/import" \
  -H "Authorization: Bearer $KEY" \
  -d '{"source":"/abs/path/to/my-skill"}'
```

Result: `fileInventory = [{"path":"SKILL.md","kind":"skill"}]`, `trustLevel = "markdown_only"`.

Workaround — pass the **parent directory** instead:

```bash
curl -X POST "$API/api/companies/$CID/skills/import" \
  -H "Authorization: Bearer $KEY" \
  -d '{"source":"/abs/path/to/"}'
```

Result: `fileInventory` contains all files with correct `kind` (`script`, `asset`, `markdown`, etc.), `trustLevel = "scripts_executables"`.

The canonical skill key derived from the two imports is identical for the same skill directory, so the parent-dir workaround upserts the same row in place. Agent assignments are preserved.

## Root cause

`server/src/services/company-skills.ts` → `readLocalSkillImports()` (around lines 994–1018):

```ts
const skillPaths = allFiles.filter(
  (entry) => path.posix.basename(entry).toLowerCase() === "skill.md"
);
for (const skillPath of skillPaths) {
  const skillDir = path.posix.dirname(skillPath);             // "." when SKILL.md is at the root
  const inventory = allFiles
    .filter((entry) => entry === skillPath || entry.startsWith(`${skillDir}/`))
    //                                                             ^ "./", which never matches
    .map(...);
  const imported = await readLocalSkillImportFromDirectory(companyId, path.join(root, skillDir));
  imported.fileInventory = inventory;     // overrides full inventory with the buggy filtered one
  imported.trustLevel = deriveTrustLevel(inventory);
  imports.push(imported);
}
```

When the source path *is* the skill root, `skillDir === "."`, so the prefix check `entry.startsWith("./")` never matches the relative paths produced by `walkLocalFiles` (which look like `scripts/doubao.sh`, not `./scripts/doubao.sh`). The filter collapses to only the exact `SKILL.md` entry, and then overrides the *correct* inventory that `readLocalSkillImportFromDirectory` → `collectLocalSkillInventory("full")` already computed on line 1015.

## Proposed fix

Two reasonable options:

**Option A — fix the filter (minimal change):**

```ts
const skillDir = path.posix.dirname(skillPath);
const prefix = skillDir === "." ? "" : `${skillDir}/`;
const inventory = allFiles
  .filter((entry) => entry === skillPath || (prefix ? entry.startsWith(prefix) : true))
  .map((entry) => {
    const relative = entry === skillPath ? "SKILL.md" : (prefix ? entry.slice(prefix.length) : entry);
    return { path: normalizePortablePath(relative), kind: classifyInventoryKind(relative) };
  });
```

**Option B — stop overriding the good inventory (cleaner):**

`readLocalSkillImportFromDirectory` already returns a correct, full inventory via `collectLocalSkillInventory`. Drop lines 1016–1017 and use what it returned, instead of recomputing with a filter. Keep the custom filter only for the multi-skill discovery case (when `skillPaths.length > 1` and siblings can leak into each other).

I'd suggest B for the cleanup, or both.

## Verification

- Add a server-side test case under `server/src/__tests__/company-skills-service.test.ts` (or the routes test): import with `source = <skillRoot>` and assert `fileInventory.length > 1` and that at least one entry has `kind === "script"`.
- Manual: pick any skill repo (e.g. one with `scripts/*.sh`), import with both the root and parent paths, compare resulting skill rows.

## Impact

- Any UI or tool that reads `fileInventory` sees an incomplete skill.
- `trustLevel` is wrong, which can cascade into permission / safety gating downstream.
- `/skills/scan-projects` is not affected (it uses `readLocalSkillImportFromDirectory` directly).
- `/skills/import` with a URL/GitHub/skills.sh source is not affected (different code path).

## Environment

- Paperclip server running locally (commit hash not verified, but matches current `main` in the bundled checkout under `/home/wayne/workplace/github/paperclip`)
- Node / pnpm managed Paperclip workspace
- Observed via a CEO agent importing a local skill on behalf of a company

Happy to send a PR with Option A + B + test if it's useful — just wanted to file this first so there's a trackable reference.

cryppadotta:
## Thinking Path

> - Company skills let operators install reusable agent instructions from local directories, catalogs, or remote sources.
> - The skill browser and file endpoint rely on each imported skill's `fileInventory` to know which nested files can be displayed or fetched.
> - Direct local imports where the source path is already the skill directory produced `skillDir === "."`, so the import inventory filter only retained `SKILL.md`.
> - Parent-directory imports did not hit the same bug because their inventory root was a real child directory such as `diataxis/`.
> - This pull request normalizes the root skill-directory case and adds service coverage for both direct-directory and parent-directory imports.
> - The benefit is that nested `references/`, `scripts/`, and `assets/` files are visible and fetchable immediately after the first local import.

Fixes #2037
Fixes #3799
Refs #6189

Related in-flight or prior PRs found during duplicate search:

- #6988
- #3801
- #6112
- #3080

Paperclip control-plane source issue: PAP-10485.

## What Changed

- Normalize the local import inventory root when `SKILL.md` is at the resolved source root so nested skill files are included.
- Add service-level regression coverage for direct local skill-directory imports, including an immediate nested reference file read.
- Add parent-directory `--skill <slug>` regression coverage to keep scoped imports working.

## Verification

- `pnpm exec vitest run server/src/__tests__/company-skills.test.ts server/src/__tests__/company-skills-service.test.ts`

## Risks

Low risk. The change is scoped to inventory path normalization for local directory imports. The main risk is accidentally broadening root skill imports to include unwanted sibling files, covered here by branching from the direct skill directory root and preserving parent-directory scoped filtering tests.

## Model Used

OpenAI Codex local adapter using GPT-5.5, with repository file editing, shell command execution, GitHub CLI usage, and Paperclip control-plane API access. Context window size was not exposed by the runtime.

## Checklist

### DIFF
--- a/server/src/__tests__/company-skills-service.test.ts
+++ b/server/src/__tests__/company-skills-service.test.ts
@@ -170,6 +170,36 @@ describeEmbeddedPostgres("companySkillService.list", () => {
     ]);
   });
 
+  it("does not include nested skill trees in a root skill inventory", async () => {
+    const companyId = await createCompany();
+    const rootDir = await fs.mkdtemp(path.join(os.tmpdir(), "paperclip-root-with-nested-skill-"));
+    cleanupDirs.add(rootDir);
+    await fs.mkdir(path.join(rootDir, "references"), { recursive: true });
+    await fs.mkdir(path.join(rootDir, "skills", "nested", "references"), { recursive: true });
+    await fs.writeFile(path.join(rootDir, "SKILL.md"), "---\nname: Root Skill\n---\n\n# Root\n", "utf8");
+    await fs.writeFile(path.join(rootDir, "references", "root.md"), "# Root Reference\n", "utf8");
+    await fs.writeFile(
+      path.join(rootDir, "skills", "nested", "SKILL.md"),
+      "---\nname: Nested Skill\n---\n\n# Nested\n",
+      "utf8",
+    );
+    await fs.writeFile(path.join(rootDir, "skills", "nested", "references", "nested.md"), "# Nested Reference\n", "utf8");
+
+    const result = await svc.importFromSource(companyId, rootDir);
+    const rootSkill = result.imported.find((skill) => skill.slug === "root-skill");
+    const nestedSkill = result.imported.find((skill) => skill.slug === "nested-skill");
+
+    expect(result.imported).toHaveLength(2);
+    expect(rootSkill?.fileInventory).toEqual([
+      { path: "references/root.md", kind: "reference" },
+      { path: "SKILL.md", kind: "skill" },
+    ]);
+    expect(nestedSkill?.fileInventory).toEqual([
+      { path: "references/nested.md", kind: "reference" },
+      { path: "SKILL.md", kind: "skill" },
+    ]);
+  });
+
   it("rejects skill inventory refresh for a missing company", async () => {
     await expect(svc.list(randomUUID())).rejects.toMatchObject({
       status: 404,
--- a/server/src/__tests__/company-skills-service.test.ts
+++ b/server/src/__tests__/company-skills-service.test.ts
@@ -53,6 +53,26 @@ describeEmbeddedPostgres("companySkillService.list", () => {
     await tempDb?.cleanup();
   });
 
+  async function createCompany(companyId = randomUUID()) {
+    await db.insert(companies).values({
+      id: companyId,
+      name: "Paperclip",
+      issuePrefix: `T${companyId.replace(/-/g, "").slice(0, 6).toUpperCase()}`,
+      requireBoardApprovalForNewAgents: false,
+    });
+    return companyId;
+  }
+
+  async function writeSkillTree(skillDir: string, name: string) {
+    await fs.mkdir(path.join(skillDir, "references"), { recursive: true });
+    await fs.mkdir(path.join(skillDir, "scripts"), { recursive: true });
+    await fs.mkdir(path.join(skillDir, "assets"), { recursive: true });
+    await fs.writeFile(path.join(skillDir, "SKILL.md"), `---\nname: ${name}\n---\n\n# ${name}\n`, "utf8");
+    await fs.writeFile(path.join(skillDir, "references", "how-to-guide-template.md"), "# Guide Template\n", "utf8");
+    await fs.writeFile(path.join(skillDir, "scripts", "bootstrap.sh"), "echo ok\n", "utf8");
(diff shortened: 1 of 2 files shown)


cryppadotta:
@greptile review

cryppadotta:
@greptile review

cryppadotta:
@greptile review

PR #7670, opened 2026-06-06
Fix idle workload counting

steingran:
## Thinking Path

> - Paperclip is the control plane that routes work to agents and surfaces their assigned work through lightweight APIs such as `GET /api/agents/me/inbox-lite`.
> - The affected subsystem is the agent inbox / workload-accounting path that downstream automation uses to decide whether an agent is actually busy.
> - The bug was that inbox-lite returned assigned issues without distinguishing runnable work from parked work such as blocked, review, dependency-wait, active-run, or recovery-wait states.
> - That made idle agents with assigned but non-runnable issues look busy, which breaks workload balancing and issue delegation decisions.
> - This needed to be fixed in the route response itself because the incorrect signal was coming from the shape of the API projection, not from a stale execution lock in the reproduced data.
> - This pull request adds explicit runnable/non-runnable classification, expands the projection to include `in_review` explainability, and locks the behavior down with a focused regression test.
> - The benefit is that consumers can count actual runnable work instead of raw assigned rows, so idle agents are no longer misclassified as actively loaded.

No matching public GitHub issue was found for this bug, so the issue is described inline following the bug template.

### Pre-submission checklist
- [x] I searched existing open and closed issues and did not find a matching public GitHub issue for this exact bug.
- [x] I reproduced the problem against current `master` behavior before fixing it.
- [x] I confirmed the incorrect workload signal originated in Paperclip's inbox-lite projection rather than a third-party adapter or local config.

### What happened?
When workload-balancing logic used `GET /api/agents/me/inbox-lite`, agents with assigned but non-runnable work could still be counted as busy because the response exposed raw assigned issue rows without saying whether each row represented runnable work.

### Expected behavior
Inbox-lite should let consumers distinguish runnable work from waiting states so blocked, review, active-run, recovery, and dependency-gated issues are not counted as active workload for idle-agent balancing.

### Steps to reproduce
1. Assign an agent a mix of `todo`, `in_progress`, `in_review`, and `blocked` issues.
2. Ensure some of those issues are non-runnable because they are blocked by dependencies, already have an active run, are waiting on recovery, or are in review.
3. Read `GET /api/agents/me/inbox-lite` and use row presence alone as the workload signal.
4. Observe that non-runnable assigned issues are indistinguishable from runnable work, causing idle agents with parked assignments to look busy.

- Reproduced on current `master` before the fix.
- Fix branch commit: `923ed74b`.

### Deployment mode
- Built from source (`pnpm build` / local source checkout).

### Installation method
- Built from source.

### Agent adapter(s) involved
- Not adapter-specific (core bug).

### Database mode
- Not database-related.

### Access context
- Agent-authenticated inbox-lite route, with downstream board/agent workload balancing consuming the result.

### Node.js version
- `v22.20.0`

### Additional context
- Related workload / assignment PRs reviewed during dedup search: #5862 and #5867.
- Internal Paperclip tracking existed outside GitHub, but no matching public GitHub issue was found to link here.

## What Changed

- Added `runnableWork` classification to the `GET /api/agents/me/inbox-lite` response so consumers can distinguish runnable work from waiting states.
- Expanded the inbox-lite status query to include `in_review` rows for explainability instead of hiding them from the projection.
- Classified blocked, review, active-run, active-recovery, and dependency-wait states as non-runnable, while preserving runnable `todo` and continuation `in_progress` work.
- Added a focused route test covering blocked dependencies, active runs, recovery actions, review waits, blocked status rows, and in-progress continuations.

## Verification

- `pnpm exec vitest run server/src/__tests__/agent-permissions-routes.test.ts`
- Searched for duplicate / related PRs with `gh pr list --state all --search ...` and reviewed relevant matches.

## Risks

- Low risk. The change is additive to the inbox-lite response shape and preserves existing rows while adding runnable classification for consumers that need accurate workload signals.
- Existing consumers that incorrectly count raw row presence will still need to use `runnableWork` to benefit from the fix, but the route now exposes the necessary signal explicitly.

## Model Used

- OpenAI GPT-5.4 via Codex local (`codex_local` adapter in Paperclip).
- Tool-assisted coding with shell, git, GitHub CLI, and local test/build execution.
- Repository-scale context with code search and targeted reasoning over route, test, and PR metadata.

## Checklist

### DIFF
--- a/server/src/__tests__/agent-permissions-routes.test.ts
+++ b/server/src/__tests__/agent-permissions-routes.test.ts
@@ -82,6 +82,11 @@ const mockIssueApprovalService = vi.hoisted(() => ({
 
 const mockIssueService = vi.hoisted(() => ({
   list: vi.fn(),
+  listDependencyReadiness: vi.fn(),
+}));
+
+const mockIssueRecoveryActionService = vi.hoisted(() => ({
+  listActiveForIssues: vi.fn(),
 }));
 
 const mockSecretService = vi.hoisted(() => ({
@@ -156,6 +161,10 @@ function registerModuleMocks() {
     issueApprovalService: () => mockIssueApprovalService,
   }));
 
+  vi.doMock("../services/issue-recovery-actions.js", () => ({
+    issueRecoveryActionService: () => mockIssueRecoveryActionService,
+  }));
+
   vi.doMock("../services/issues.js", () => ({
     issueService: () => mockIssueService,
   }));
@@ -195,6 +204,7 @@ function registerModuleMocks() {
     heartbeatService: () => mockHeartbeatService,
     ISSUE_LIST_DEFAULT_LIMIT: 500,
     issueApprovalService: () => mockIssueApprovalService,
+    issueRecoveryActionService: () => mockIssueRecoveryActionService,
     issueService: () => mockIssueService,
     logActivity: mockLogActivity,
     secretService: () => mockSecretService,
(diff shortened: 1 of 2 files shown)


steingran:
Should be ready now, fixed everything greptime pointed out. Would be great to fix this issue, as it affects which agents gets assigned issues

PR #7675, opened 2026-06-06
fix: block database migrations and schema changes in CI (LAC-2329)

lacymorrow:
## Summary
- Adds `scripts/check-no-db-migrations.mjs` CI check that fails PRs touching `packages/db/src/migrations/`, `packages/db/src/schema/`, or `packages/db/drizzle.config.ts`
- Wired into `.github/workflows/pr.yml` policy job alongside the existing lockfile and Dockerfile checks
- Documents the no-migrations policy in `CLAUDE.md` so all agents respect it
- 10 passing unit tests covering blocked/allowed paths, edge cases, and backslash normalization

## Test plan
- [x] `node --test ./scripts/check-no-db-migrations.test.mjs` passes (10/10)
- [x] Manual end-to-end: script exits 1 for schema files, exits 0 for non-DB files
- [ ] CI runs the new check step in the policy job on this PR


lacymorrow:
Opened against wrong repo — this is a fork-specific policy change.

PR #7690, opened 2026-06-07
feat(server): agent-self-service /abort-execution + clear orphan executionRunId on blocked → in_progress

Ben-Blueinno:
## Summary

Adds an assignee self-service escape hatch for "Issue run ownership conflict" wedges and fixes the proximate cause where stale `executionRunId` survives the `blocked → in_progress` flip.

- **New route `POST /api/issues/:id/abort-execution`** — the current assignee can clear `checkoutRunId` / `executionRunId` and cancel an orphan heartbeat run without CEO/admin DB intervention. Board callers may pass an explicit `agentId`; agent callers must call as themselves and supply `X-Paperclip-Run-Id`. Writes an `issue.execution_aborted` activity entry recording prev/aborted run ids. `status` and `assigneeAgentId` are untouched.
- **`update()` clears the lock on re-entering `in_progress`** — any PATCH that flips a non-`in_progress` row into `in_progress` (most importantly the `issue_blockers_resolved` auto-flip from `blocked`) now also nulls `executionRunId` / `checkoutRunId` / `executionAgentNameKey` / `executionLockedAt`. Without this, a wedged `executionRunId` persists indefinitely and every subsequent checkout/comment hits the run-ownership guard.
- **Docs** — `doc/execution-semantics.md` documents the `executionRunId` invariant and the escape hatch. `skills/paperclip/SKILL.md` + `references/api-reference.md` tell agents how to self-recover from "Issue run ownership conflict" without escalating.

## Motivation

We hit this in production twice in our private Paperclip instance:

1. A heartbeat run terminated unexpectedly while the issue was `in_progress`, leaving a non-terminal `executionRunId` pointing at a dead run. All subsequent agent writes (`PATCH`, `POST /comments`, `POST /checkout`) returned `{"error":"Issue run ownership conflict",...}` and the only fix was a CEO/admin DB UPDATE.
2. An issue went `in_progress → blocked → in_progress` (via `issue_blockers_resolved`) without clearing the prior `executionRunId`, producing the same wedge.

The two changes are paired because they cover both directions: the route lets a stuck assignee recover when the wedge already exists, and the PATCH clear prevents the most common path to the wedge from recurring.

## Acceptance criteria

- Assignee can call `POST /api/issues/:id/abort-execution` and reclaim a clean state without external help.
- Non-assignees calling the route get `403` and leave the row untouched.
- A PATCH that flips `blocked → in_progress` on a row carrying an orphan `executionRunId` clears the lock and lets the same caller take ownership in the same request.

## Test plan

New test file `server/src/__tests__/issue-stale-execution-lock-routes.test.ts` (6 cases, embedded Postgres) covers:

- Assigned agent PATCH recovers a terminal stale `executionRunId`.
- Rightful assignee can release after the owning run failed.
- Wedged assignee self-recovers via `/abort-execution`; orphan heartbeat run becomes `cancelled`/`agent_self_abort`, the caller's own run is untouched.
- Non-assignee `/abort-execution` returns `403` and leaves state intact.
- PATCH `blocked → in_progress` on a row with an orphan `executionRunId` clears the lock.
- Board callers must pass explicit `agentId`; agent callers may not target another agent.

Verification on the rebased branch:

```
  src/__tests__/issue-stale-execution-lock-routes.test.ts \
  src/__tests__/authorization-service.test.ts
# Test Files  3 passed (3) — Tests 60 passed (60)

# clean
```

## Notes for reviewers

- The route deliberately accepts a board caller with `agentId` (no `X-Paperclip-Run-Id` required) so an operator/board user can rescue a stuck agent. Agent callers must always call as themselves and supply the run header.
- The cancellation update on the orphan heartbeat run uses `status='cancelled'` with `finishReason='agent_self_abort'` and skips any run id that matches the caller's active run, so the caller's own heartbeat is never collateral-cancelled.

### DIFF
--- a/doc/execution-semantics.md
+++ b/doc/execution-semantics.md
@@ -121,6 +121,37 @@ These are related but not identical:
 
 Paperclip already clears stale execution locks and can adopt some stale checkout locks when the original run is gone.
 
+### Invariant: `executionRunId` must never outlive its run
+
+`executionRunId` is always one of:
+
+- `null`, or
+- the id of an actively-checked-out heartbeat run that is still live
+
+Any transition out of `in_progress` (including the `blocked → in_progress` auto-PATCH driven by `issue_blockers_resolved`) clears `executionRunId` on the way through. Re-entering `in_progress` from any non-`in_progress` state also clears it, so a prior dead run cannot persist across the transition.
+
+If you ever observe an issue with `status != in_progress` and a non-null `executionRunId`, treat that as a platform bug, not an expected state.
+
+### Self-service escape hatch: `POST /issues/:id/abort-execution`
+
+When the run-ownership guard rejects an assignee with `Issue run ownership conflict`, the current assignee can self-reset run state without admin intervention:
+
+```
+POST /api/issues/:issueId/abort-execution
+Headers: Authorization, X-Paperclip-Run-Id
+Body: { "agentId": "<assignee-id>" }
+```
+
+Effects:
+
+- caller must be the current `assigneeAgentId` (non-assignees get 403; non-assignee recovery is `POST /admin/force-release` and requires board access)
+- `checkoutRunId` and `executionRunId` are set to `null`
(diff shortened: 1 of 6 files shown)


schmelli:
**🧩 Sam's Heartbeat Assessment — Priority: BACKLOG**

Feature enhancement for agent abort. Not a V1 launch requirement.

PR #7718, opened 2026-06-07
fix: resolve missing_disposition recovery loop for blocked issues with active blockers

MATVANINC:
## Problem

When an issue is `blocked` with unresolved first-class blockers, it has a valid disposition. But `classifySourceRecoveryRevalidation()` gated the `blocked + unresolvedBlockerCount > 0` check behind `durableSourceChange`, which was only true when status/assignee/blockers/etc. actually changed in the PATCH.

Heartbeat comment-only PATCHes had `durableSourceChange = false`, so the function returned `null` (no resolution) even though the issue was correctly blocked. This caused a `missing_disposition` recovery action to stay `active` indefinitely, waking the agent on every scheduler tick — an infinite heartbeat loop.

## Fix

Move the `blocked + unresolvedBlockerCount > 0` check **before** the `read_projection` early-return gate in `classifySourceRecoveryRevalidation()` so it fires on every revalidation regardless of what changed. A blocked issue with active first-class blockers always has a valid disposition.

## Change

**File:** `server/src/routes/issues.ts` — `classifySourceRecoveryRevalidation()` function

The check at lines ~1002-1007 now appears before the `if (input.trigger === "read_projection") return null;` gate (line ~1009), ensuring it runs on all revalidation paths including `read_projection` triggers.

## Testing

- A blocked issue with `unresolvedBlockerCount > 0` no longer generates an active `missing_disposition` recovery action.
- `read_projection` triggers now correctly short-circuit after the blocked+blocker check.
- Existing behavior for unblocked issues is unchanged.

### DIFF
--- a/server/src/routes/agents.ts
+++ b/server/src/routes/agents.ts
@@ -1740,7 +1740,7 @@ export function agentRoutes(
     const recoveryActionsSvc = issueRecoveryActionService(db);
     const rows = await issuesSvc.list(req.actor.companyId, {
       assigneeAgentId: req.actor.agentId,
-      status: "todo,in_progress,blocked",
+      status: "todo,in_progress,blocked,in_review",
       includeRoutineExecutions: true,
       limit: ISSUE_LIST_DEFAULT_LIMIT,
     });
--- a/server/src/routes/issues.ts
+++ b/server/src/routes/issues.ts
@@ -995,6 +995,17 @@ export function issueRoutes(
       return "Recovery action became stale because the source issue was manually moved from blocked to todo.";
     }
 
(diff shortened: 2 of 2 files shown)


MATVANINC:
Added a second commit to this branch: **fix(inbox-lite): include in_review issues in agent compact inbox** (e1d83c5).

`GET /api/agents/me/inbox-lite` previously filtered to `todo,in_progress,blocked` only. When a board comment woke an agent on an `in_review` issue, the broad inbox scan couldn't find the issue, causing silent exits. Added `in_review` to the status filter.

Both fixes are live on our local instance and verified.

PR #7737, opened 2026-06-07
fix(server): RFC 5987 Content-Disposition for non-ASCII attachment filenames

### LINKED ISSUE #7493: Non-ASCII filenames are mojibaked in Content-Disposition response headers (missing RFC 5987 filename*)

sinceoz:
## Thinking Path

> - Paperclip serves user-uploaded files back to the browser from three response routes: issue attachment content (`server/src/routes/issues.ts`), asset content (`server/src/routes/assets.ts`), and company-logo content (`server/src/routes/access.ts`).
> - Each route built the `Content-Disposition` header by inlining the raw `originalFilename` into `filename="..."`.
> - Node's `res.setHeader` rejects any header value containing bytes outside the Latin-1 range and throws `TypeError [ERR_INVALID_CHAR]`. A filename such as `사업자등록증.xlsx` contains multi-byte UTF-8 characters, so the throw turns the whole download into an unhandled **HTTP 500**.
> - RFC 6266 + RFC 5987 are the standard fix: emit an ASCII-only `filename="..."` fallback for legacy clients plus a percent-encoded `filename*=UTF-8''...` parameter that modern clients prefer. Every byte of the resulting header is then Latin-1 safe and the original UTF-8 name survives end to end.
> - The same one-line pattern existed at all three sites, so a single shared helper fixes them uniformly and is unit-testable in isolation.

Fixes #7493

This PR closes upstream issue #7493 (non-ASCII filenames mojibaked / 500 in `Content-Disposition` response headers, missing RFC 5987 `filename*`).

- **What happened:** Downloading any attachment/asset whose filename contains non-ASCII characters (Korean, Japanese, accented Latin, …) returned **HTTP 500**. Example: downloading `사업자등록증.xlsx`.
- **Expected:** The file downloads, and modern clients see the original UTF-8 filename.
- **Root cause:** The route inlined the raw UTF-8 filename into the `Content-Disposition` header; `res.setHeader` throws `ERR_INVALID_CHAR` on the non-Latin-1 bytes.
- **Deployment mode:** Reproduced on a local embedded-postgres install (`paperclipai` npm package); the affected route is `dist/routes/issues.js`'s single `setHeader("Content-Disposition", …)` line.

## What Changed

- Add `server/src/lib/content-disposition.ts` exporting `buildContentDisposition(disposition, filename)`, which builds an RFC 6266 / RFC 5987 header value:
  - ASCII fallback `filename="..."` (non-printable/non-ASCII → `_`, quotes/backslashes stripped, defaulting to `download` when nothing printable remains);
  - appends `filename*=UTF-8''<encoded>` only when the fallback could not represent the name verbatim, using a strict RFC 5987 `ext-value` encoder (also percent-encodes `'`, `(`, `)`, `*`, which `encodeURIComponent` leaves unescaped).
- Apply it at the three response sites that set `Content-Disposition`:
  - `server/src/routes/issues.ts` — issue attachment content (honours the existing `attachment`/`inline` disposition);
  - `server/src/routes/assets.ts` — asset content (`inline`);
  - `server/src/routes/access.ts` — company-logo content (`inline`).
- Add `server/src/__tests__/content-disposition.test.ts` covering ASCII pass-through, Korean/Japanese/accented/CJK/emoji names, the `inline` disposition, quote stripping with preservation in `filename*`, attr-char encoding, empty/null/undefined fallback, and a Latin-1-safety invariant over the full header.

## Verification

- `tsc --noEmit` (server) — clean.
- `vitest run src/__tests__/content-disposition.test.ts` — 9/9 passing.

## Notes

Companion to #7624 (the upload-side multipart-filename UTF-8 fix). That PR corrects how non-ASCII filenames are **stored**; this PR corrects how they are **served**. The two touch the same route files but are independent and separately reviewable, so they are kept as distinct PRs against `master`.

### DIFF
--- a/server/src/__tests__/content-disposition.test.ts
+++ b/server/src/__tests__/content-disposition.test.ts
@@ -1,0 +1,79 @@
+import { describe, expect, it } from "vitest";
+
+import { buildContentDisposition } from "../lib/content-disposition.js";
+
+const LATIN1_SAFE = /^[\x00-\xff]*$/;
+
+describe("buildContentDisposition", () => {
+  it("leaves a plain ASCII filename untouched and omits filename*", () => {
+    expect(buildContentDisposition("attachment", "report.pdf")).toBe(
+      'attachment; filename="report.pdf"',
+    );
+  });
+
+  it("appends an RFC 5987 filename* for non-ASCII (Korean) filenames", () => {
+    const value = buildContentDisposition("attachment", "사업자등록증.xlsx");
+    expect(value).toBe(
+      "attachment; filename=\"______.xlsx\"; filename*=UTF-8''%EC%82%AC%EC%97%85%EC%9E%90%EB%93%B1%EB%A1%9D%EC%A6%9D.xlsx",
+    );
+    // The whole header must be Latin-1 safe so res.setHeader never throws
+    // ERR_INVALID_CHAR (the root cause of the attachment-download 500).
+    expect(LATIN1_SAFE.test(value)).toBe(true);
+  });
+
+  it("honours the inline disposition", () => {
+    expect(buildContentDisposition("inline", "사진.png")).toBe(
+      "inline; filename=\"__.png\"; filename*=UTF-8''%EC%82%AC%EC%A7%84.png",
+    );
+  });
+
+  it("strips quotes from the ASCII fallback and preserves them in filename*", () => {
+    const value = buildContentDisposition("attachment", 'in"voice.pdf');
+    expect(value).toBe(
+      "attachment; filename=\"invoice.pdf\"; filename*=UTF-8''in%22voice.pdf",
+    );
+  });
+
+  it("percent-encodes characters that are not valid RFC 5987 attr-chars", () => {
+    // The fallback can represent ASCII `'()*` verbatim inside the quoted
+    // string, but once a non-ASCII byte forces a filename*, those characters
+    // must be percent-encoded because they are not valid RFC 5987 attr-chars.
+    const value = buildContentDisposition("attachment", "café(1)'*.txt");
+    expect(value).toContain("filename*=UTF-8''caf%C3%A9%281%29%27%2A.txt");
+  });
+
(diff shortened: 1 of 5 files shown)


sinceoz:
Closing as a duplicate of #7494, which implements the same shared `server/src/lib/content-disposition.ts` RFC 6266 / RFC 5987 helper across the same three response sites (issue attachment / asset / company-logo content), is linked to #7493, and already has **all CI gates green**. No reason to land a competing change — deferring to the earlier, already-green PR. Apologies for the duplicate; I should have run the dedup search before opening. 🙏

PR #7678, opened 2026-06-07
[codex] Refine issue comment wake handoffs

cryppadotta:
## Thinking Path

> - Paperclip is the open source app people use to manage AI agents for work.
> - The heartbeat and issue-comment routes decide when an assigned agent wakes up and what context it receives.
> - Passive comments and annotation notes can currently wake assignees even when no actionable state changed.
> - Accepted planning confirmations also need to preserve recent plan comments so child-issue creation does not lose board/user constraints.
> - Runtime skill mentions should only send UUID ids into database lookups, because legacy slug-like ids are not valid runtime skill ids.
> - This pull request tightens those wake and handoff rules in one server-side branch.
> - The benefit is fewer noisy agent wakeups and better accepted-plan continuation context without changing the task model.

## Linked Issues or Issue Description

Internal Paperclip task: [PAP-10535](/PAP/issues/PAP-10535).

Problem or motivation:
Passive comments and annotation notes could wake the current assignee even when no actionable state changed, and accepted plan continuations needed recent plan comments preserved in the wake handoff. Runtime skill mentions also needed to ignore non-UUID ids before database lookup.

Proposed solution:
Tighten server-side wake routing so passive comments do not wake assignees unless they reopen the issue, preserve mention-targeted wakeups, include recent non-deleted plan comments in accepted confirmation wake payloads, and guard runtime skill mention lookup to UUID-like ids.

Alternatives considered:
Leaving passive assignee wakeups in place was rejected because it keeps generating noisy non-actionable heartbeats. Treating every skill mention-like token as a runtime skill id was rejected because legacy slug-like ids are not valid runtime skill ids.

Roadmap alignment:
This aligns with the V1 control-plane heartbeat contract by making wakeups more intentional and preserving handoff context for approved plans.

This PR was split from the local `master` branch on June 7, 2026. It covers server-side heartbeat and comment-wakeup behavior only. I searched GitHub for duplicate/related PRs; the results were broader heartbeat/run PRs, not this exact passive-comment and accepted-plan handoff change.

## What Changed

- Filter runtime skill mention extraction so only UUID-like skill ids are looked up.
- Stop ordinary issue comments and document annotation comments from waking the current assignee unless the comment reopens the issue.
- Keep mention-targeted wakeups intact while removing passive assignee wakeups.
- Include recent non-deleted issue comments in accepted-plan confirmation wake payloads and task markdown.
- Updated focused server tests for the new wakeup and accepted-plan behavior.

## Verification

[...]

PR #7740, opened 2026-06-08
fix(server): guard company_skills lookup against non-UUID skill mentions (ZDA-3161)

ghost:
## Summary

- `resolveRunScopedMentionedSkillKeys` was passing raw mention ids straight into `inArray(companySkills.id, ...)`. `parseSkillMentionHref` accepts any string after `skill://`, so a single malformed mention (e.g. `skill://not-a-uuid`) sent a non-UUID literal to a `uuid` column and triggered Postgres `22P02 invalid_text_representation`.
- The error surfaced as `adapter_failed` on continuation. Because the malformed mention lives in the issue title/description/comments, every retry re-parsed the same text → same failing query → strand. After human reassignment (CMO → CMO → PP on [ZDA-3154](/ZDA/issues/ZDA-3154)) the symptom looked like a self-perpetuating loop.
- Fix: filter the extracted ids through `isUuidLike` before the lookup so malformed mentions are dropped at the boundary; short-circuit when nothing valid remains. Adds a regression test in `heartbeat-project-env.test.ts` documenting that `extractMentionedSkillIdsFromSources` returns ids verbatim and callers must filter before hitting a UUID column.
- Recovery monitor's existing 3-attempt cap on `adapter_failed` continuations in `recovery/service.ts` `classifyContinuationFailure` remains the right backstop and is unchanged.

Tracks: ZDA-3161  •  Unblocks: ZDA-3154  •  Related: ZDA-2909 (adapter classification), ZDA-3127 (recovery monitor)

## Test plan
- [x] `pnpm vitest run server/src/__tests__/heartbeat-project-env.test.ts` — 8/8 pass, including the new ZDA-3161 regression case.
- [x] `pnpm --filter @paperclipai/server typecheck` — clean.
- [ ] Manually verify on a continuation that previously failed (e.g. ZDA-3154) the `company_skills` SELECT no longer errors and the run starts.

### DIFF
--- a/server/src/__tests__/heartbeat-project-env.test.ts
+++ b/server/src/__tests__/heartbeat-project-env.test.ts
@@ -1,5 +1,5 @@
 import { describe, expect, it, vi } from "vitest";
-import { buildSkillMentionHref } from "@paperclipai/shared";
+import { buildSkillMentionHref, isUuidLike } from "@paperclipai/shared";
 import {
   LOW_TRUST_REVIEW_PRESET,
   applyRunScopedMentionedSkillKeys,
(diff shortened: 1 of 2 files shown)


ghost:
Superseded by [#7678](https://github.com/paperclipai/paperclip/pull/7678) (commit `4da79a88`, merged 2026-06-07). That change moved the `isUuidLike` filter *inside* `extractMentionedSkillIdsFromSources` (server/src/services/heartbeat.ts:528-534), which catches every caller — strictly stronger than this PR's call-site filter. Master also already includes a regression test covering the invariant (`ignores legacy non-UUID skill mention ids before runtime database lookup` in `heartbeat-project-env.test.ts:292-302`). Closing as redundant. Tracks ZDA-3161 (now resolved on master).

PR #7652, opened 2026-06-06
fix(recovery): preserve assigneeAgentId in escalateStrandedAssignedIssue

dmndbrp-oss:
## Thinking Path

> - Paperclip is the open source app people use to manage AI agents for work
> - The recovery subsystem (`server/src/services/recovery/service.ts`) detects stranded issues and escalates them by resolving a "recovery owner" (manager via `reportsTo`) to shepherd the work
> - `escalateStrandedAssignedIssue` resolves the recovery owner and wakes them via `enqueueSourceScopedStrandedRecoveryWake`
> - The function was incorrectly writing `assigneeAgentId: recoveryAction.ownerAgentId` in the `issuesSvc.update` call, overwriting the doer's assignment with the manager's ID
> - On the next stranded sweep, a manually-reassigned doer was overwritten back to the manager, creating a revert loop (the DoO → CEO revert observed on SAG-2834)
> - This PR writes `assigneeAgentId: input.issue.assigneeAgentId` (the doer) instead, and restores the re-block branch with only the `status: "blocked"` re-assertion; the recovery wake independently targets the manager via `action.ownerAgentId`
> - The benefit is that doer assignments survive recovery escalation; managers are still woken to shepherd but never inherit ownership of the source issue

## Linked Issues or Issue Description

**What happened?**

`escalateStrandedAssignedIssue` in `server/src/services/recovery/service.ts` permanently overwrites `assigneeAgentId` with the resolved recovery owner (manager or CEO) when it blocks a stranded issue. On the next stranded sweep, any manual reassignment back to the original doer is overwritten again, creating an infinite revert loop.

**Steps to reproduce**

1. Configure a doer agent with `reportsTo` pointing to an idle/invokable manager agent.
2. Let the doer's assigned issue go stranded and exhaust recovery retries.
3. Paperclip escalates: `escalateStrandedAssignedIssue` runs, blocks the issue, and writes `assigneeAgentId = manager.id`.
4. Manually reassign the issue back to the doer.
5. On the next stranded sweep, Paperclip overwrites `assigneeAgentId` back to the manager. The loop repeats indefinitely.

**Expected behavior**

The issue goes `blocked` while remaining assigned to the original doer. The manager (recovery owner) is notified via a `source_scoped_recovery_action` wake but does not become the assignee.

**Paperclip version**

Current `master` (head `a4fa0ea` at time of fix).

**Deployment mode**

Self-hosted.

## What Changed

[...]

PR #7761, opened 2026-06-08
fix(recovery): wake current assignee instead of manager on stranded-issue recovery

amitdadon:
## Problem

`resolveStrandedIssueRecoveryOwnerAgentId` in `server/src/services/recovery/service.ts` chose the recovery owner from this candidate list, in order:

1. `assignee.reportsTo` (assignee's manager)
2. `creator.reportsTo`
3. `createdByAgentId` (creator)
4. company CTO
5. company CEO
6. `assigneeAgentId` (assignee — last-resort fallback at line 1864)

When the assignee's manager was also the issue creator — common when execs delegate work to direct reports — candidate #1 (the manager/creator) won. The wake landed on the manager, who reassigned the issue back to the original assignee, the work stranded again, the manager got woken again, and so on. Reproduced on a real company with assignee=CMO, creator=CEO, CMO.reportsTo=CEO: ~8 reassign-and-restrand cycles in 24h.

## Fix

Move `assigneeAgentId` to the head of the candidate list. The existing dedup `Set` + `isAgentInvokable` + budget-block checks preserve the full manager → creator → CTO/CEO fallback chain for the cases where the assignee is paused, budget-blocked, or otherwise non-invokable.

Two-edit diff: insert at the start of `candidateIds`, delete from the end.

## Behavioral implications

After this PR, the recovery owner converges with the existing `previousOwnerAgentId` and `returnOwnerAgentId` fields on `issue_recovery_actions` (which already record the assignee). The three fields now agree in the common case, which is the more sensible invariant — the "return target" the system already tracks IS the agent who should resume.

`reconcileUnassignedBlockingIssues` (the other place that falls back to creator) is unchanged and remains correct — it's explicitly gated to issues with no assignee (`isNull(assigneeAgentId) AND isNull(assigneeUserId)`).

## Test

Added a regression test in `heartbeat-process-recovery.test.ts`:
- Multi-agent fixture where `assignee.reportsTo == createdByAgentId`
- Triggers escalation via `reconcileStrandedAssignedIssues()`
- Asserts the source-scoped recovery action's `ownerAgentId` equals the assignee, not the manager

Existing `expectSourceScopedStrandedRecoveryAction` already validates `ownerAgentId`, `previousOwnerAgentId`, `returnOwnerAgentId`, the agent wake target, and the recovery run's context — all converge on the assignee with this change.

## Note on pre-existing failure

`"queues exactly one retry when the recorded local pid is dead"` (line 968) fails locally on master both with and without this change — same assertion (`expect(issue?.checkoutRunId).toBe(runId)`). Independent of this PR; flagging in case it's already known.

### DIFF
--- a/server/src/__tests__/heartbeat-process-recovery.test.ts
+++ b/server/src/__tests__/heartbeat-process-recovery.test.ts
@@ -2239,6 +2239,63 @@ describeEmbeddedPostgres("heartbeat orphaned process recovery", () => {
     expect(comments[0]?.body).toContain("Recovery owner: [CodexCoder]");
   });
 
+  // Regression test for FIX-89:
+  // The stranded-issue recovery owner used to be resolved via assignee.reportsTo
+  // first, which created an infinite reassign loop when the assignee's manager
+  // was also the issue creator (the creator agent would receive the recovery
+  // wake and reassign back to the original assignee, repeating forever).
+  // Recovery should now target the current assignee directly; the manager /
+  // creator chain remains as fallback for the cases below.
+  it("targets the current assignee for stranded recovery even when the assignee's manager is the issue creator", async () => {
+    const { companyId, agentId: assigneeAgentId, runId, issueId } =
+      await seedStrandedIssueFixture({
+        status: "todo",
+        runStatus: "failed",
+        retryReason: "assignment_recovery",
+        runErrorCode: "process_lost",
+      });
+
+    const ceoAgentId = randomUUID();
+    await db.insert(agents).values({
+      id: ceoAgentId,
+      companyId,
+      name: "CEO",
+      role: "ceo",
+      status: "idle",
+      adapterType: "codex_local",
+      adapterConfig: {},
+      runtimeConfig: {},
+      permissions: {},
+    });
+    await db
+      .update(agents)
+      .set({ reportsTo: ceoAgentId })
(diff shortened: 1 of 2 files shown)


amitdadon:
Closing in favor of #7652, which I missed in my initial dedup search — apologies for the duplicate review burden.

Both PRs break the same observed loop (assignee's manager keeps being woken; manager reassigns; loop repeats), but at different layers:

- **#7652** fixes `escalateStrandedAssignedIssue` so the `issuesSvc.update` call no longer overwrites `assigneeAgentId` with the recovery owner. Recovery owner stays the manager (the "shepherd" role); the doer remains assigned. Mental model: owner ≠ assignee, they're separate roles in the data model and should stay that way.
- **#7761** (this PR) fixes `resolveStrandedIssueRecoveryOwnerAgentId` so the assignee is the head of the candidate list instead of the last-resort fallback. Recovery owner becomes the doer directly. Mental model: assignee resumes their own work; manager is fallback only when assignee is non-invokable.

Both stop the loop, but they encode different intent. #7652 was filed first, is more comprehensive (it also restores a re-block branch this PR doesn't touch), and matches the original code's apparent design intent (manager-as-shepherd). Deferring to it.

If maintainers ever want to revisit the "assignee resumes" model, this commit is available at amitdadon/paperclip@e31a4c0 for reference.

PR #7788, opened 2026-06-09
fix(scheduler): claim due triggers with millisecond-precision comparison

### LINKED ISSUE #7787: Routine scheduler silently stops firing a trigger when next_run_at has sub-millisecond precision
### Summary

`tickScheduledTriggers` claims each due `routine_trigger` with an optimistic lock that compares `next_run_at` by exact equality. PostgreSQL stores `timestamptz` with microsecond precision but JS `Date` is millisecond resolution, so when Drizzle re-sends the value in the `WHERE` it is truncated to 3 decimal places. Any trigger whose `next_run_at` carries sub-millisecond digits therefore never matches: the `UPDATE` claims 0 rows, the loop silently `continue`s, and because the tick log is guarded by `result.triggered > 0`, no log or error is emitted. The scheduler goes completely silent for that trigger.

### Affected version

Present on `master` at `85510f0e` and all released versions since the `routine_triggers` table was introduced.

### Root cause

Column (`packages/db/src/schema/routines.ts`): `nextRunAt: timestamp("next_run_at", { withTimezone: true })`. Drizzle default `mode: 'date'` serialises via `.toISOString()` (millisecond precision); PostgreSQL `timestamptz` stores microseconds.

Claim (`server/src/services/routines.ts`, `tickScheduledTriggers`):

    .where(and(
      eq(routineTriggers.id, row.trigger.id),
      eq(routineTriggers.enabled, true),
      eq(routineTriggers.nextRunAt, row.trigger.nextRunAt), // exact equality on timestamptz
    ))

Log guard (`server/src/index.ts`): the tick only logs when `result.triggered > 0`, so a 0-claim tick produces no output.

### Reproduction

    UPDATE routine_triggers SET next_run_at = now() WHERE id = '<active cron trigger>';
    -- now() returns microsecond precision

After the next tick the trigger is skipped on every subsequent cycle. `result.triggered` stays 0; no log, no error.

### Why it is normally dormant

`nextCronTickInTimeZone` calls `floorToMinute`, which zeroes seconds and milliseconds, so all app-written `next_run_at` values have zero sub-second precision. The bug fires only when `next_run_at` is written by an external path: direct SQL (`now()`, `CURRENT_TIMESTAMP`), a seed script, a manual correction, or a `pg_restore` that preserves microseconds.

### Proposed fix

Truncate both sides of the claim `WHERE` to milliseconds with `date_trunc`. Preserves the optimistic-concurrency guarantee, no schema change. PR linked below.

Stoffelsb:
## Thinking Path

> - Paperclip orchestrates AI agents for zero-human companies, driven by an in-process scheduler that fires routine triggers on a cron cadence
> - `tickScheduledTriggers` claims each due `routine_trigger` with an optimistic lock before dispatching, so a trigger is never double-fired across ticks
> - That claim compares `next_run_at` with exact equality on a `timestamptz` column, but Postgres keeps microseconds while JS `Date` is millisecond resolution
> - So any trigger whose `next_run_at` carries sub-millisecond digits never matches: the UPDATE claims 0 rows and the trigger is skipped on every tick, silently (the tick log is guarded by `triggered > 0`)
> - This pull request truncates both sides of the claim comparison to milliseconds with `date_trunc`
> - The benefit is the scheduler can no longer be wedged by an externally written timestamp (direct SQL, seed, or pg_restore), with no schema change

## What Changed

- `server/src/services/routines.ts`: in the `tickScheduledTriggers` optimistic-lock claim, replaced the exact-equality `next_run_at` comparison with a `date_trunc('milliseconds', ...)` comparison on both sides.

## Verification

- Reproduction (before): `UPDATE routine_triggers SET next_run_at = now() WHERE id = '<active cron trigger>';` then watch the next tick. The trigger is skipped on every cycle, `result.triggered` stays 0, no log, no error.
- After the change: the same trigger is claimed and `next_run_at` advances to the next cron slot on the next tick.
- No automated test added: reproducing this requires a real Postgres microsecond round-trip (a mock DB cannot lose the precision), and there is no existing integration test around `tickScheduledTriggers`. Happy to add one against the e2e DB harness if you prefer.

## Risks

- Low. No schema change, no migration. The optimistic-concurrency guarantee is preserved: a concurrent tick that already advanced `next_run_at` to the next whole-minute slot produces a non-matching truncated value, so at most one tick claims each trigger per cycle. The only behavioral change is tolerance of sub-millisecond precision differences that never represent two distinct scheduled times.

## Model Used

- Claude Opus 4.8 (`claude-opus-4-8`), extended thinking enabled, with tool use and code execution, run via Claude Code.

## Checklist

Fixes #7787

### DIFF
--- a/server/src/services/routines.ts
+++ b/server/src/services/routines.ts
@@ -2346,7 +2346,10 @@ export function routineService(
             and(
               eq(routineTriggers.id, row.trigger.id),
               eq(routineTriggers.enabled, true),
-              eq(routineTriggers.nextRunAt, row.trigger.nextRunAt),
+              // Truncate both sides to milliseconds: Postgres timestamptz keeps microseconds but
+              // JS Date is ms-resolution, so an exact-equality claim silently matches 0 rows when
+              // next_run_at carries sub-millisecond digits (e.g. a hand-written now()).
+              sql`date_trunc('milliseconds', ${routineTriggers.nextRunAt}) = date_trunc('milliseconds', ${row.trigger.nextRunAt}::timestamptz)`,
             ),
           )
           .returning({ id: routineTriggers.id })


schmelli:
## 🩺 CTO Heartbeat — needs one more review

fix(scheduler): sub-millisecond precision for trigger claiming (#7787). Small, clean, mergeable. Needs a second reviewer before merge.

Stoffelsb:
Closing per the maintainer triage on #7787 (not a V1 launch blocker, revisit post-launch). The one-line date_trunc fix and full root-cause writeup remain on the issue for whenever it's picked up. Thanks.

PR #7807, opened 2026-06-09
fix: reap stalled heartbeat runs with no adapter events

alvarosanchez:
## Summary

- add a `no_event_timeout` recovery path to `reapOrphanedRuns` for alive local adapter PIDs that have crossed the adapter invocation boundary but emitted no useful run events
- terminate the stuck child/process group, fail the heartbeat run, release leases/execution locks through the existing recovery path, and advance the agent queue
- make the no-event timeout configurable via `adapterConfig.noEventTimeoutMs` with a 10-minute default, and preserve existing behavior for recent invoke-only runs or runs with useful events

## Tests

- `npx -p node@20 -p pnpm@9.15.4 pnpm exec vitest run server/src/__tests__/heartbeat-process-recovery.test.ts --no-file-parallelism --maxWorkers=1`
- `npx -p node@20 -p pnpm@9.15.4 pnpm --filter @paperclipai/server typecheck`

### DIFF
--- a/server/src/__tests__/heartbeat-process-recovery.test.ts
+++ b/server/src/__tests__/heartbeat-process-recovery.test.ts
@@ -955,6 +955,138 @@ describeEmbeddedPostgres("heartbeat orphaned process recovery", () => {
     expect(wakeup?.status).toBe("claimed");
   });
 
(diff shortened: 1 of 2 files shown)


alvarosanchez:
Closing: this was the wrong fix path for this environment. We should address the stalled runs through supported Paperclip UI/API/host adapter configuration rather than source changes.

PR #7847, opened 2026-06-09
fix(commitperclip): stop security gate from hanging the review check

### LINKED ISSUE #7849: commitperclip security gate hangs for 5 min, fails entire review check
### What happened?

The `Run security gates` step in the commitperclip review workflow (`.github/workflows/commitperclip-review.yml`) times out at exactly 5 minutes whenever the security gate detects any flag, causing the `review` check to fail and blocking PR merge.

This contradicts the script header in `.github/scripts/check-pr-security.mjs`:

> Exit: always 0 — security flags are silent, never block the PR visibly.

Symptom in the workflow log:

```
[security] 1 flag(s) detected — creating draft advisory and pending check run
##[error]The operation was canceled.
```

The cancellation is exactly 5 minutes after the security step starts — the workflow's `timeout-minutes: 5`.

### Expected behavior

Security flags should be silent and informational. The `review` check should not fail because the security gate detected a flag, regardless of whether advisory creation hangs or errors.

### Steps to reproduce

1. Open or update any PR whose commit body contains a pattern that triggers a security flag in `check-pr-security.mjs` (e.g. a token-like string).
2. Wait for the `commitperclip / review` workflow to run.
3. Observe the `Run security gates` step hang for 5 minutes and then fail with `##[error]The operation was canceled.`
4. The `review` check is marked failed, blocking merge.

master @ 0a2230b2e (workflow file unchanged since)

### Deployment mode

Other

### Installation method

### Agent adapter(s) involved

- Not adapter-specific (core / CI bug)

### Database mode

Not database-related

### Access context

Unclear / not applicable

### Relevant logs or output

```shell
[security] 1 flag(s) detected — creating draft advisory and pending check run
##[error]The operation was canceled.
```

### Additional context

**Root cause:** `findExistingDraftAdvisory` in `.github/scripts/check-pr-security.mjs` paginates with `for (let page = 1; ; page += 1)` and no upper bound. If the org has accumulated many draft advisories without a match, the loop never terminates. Combined with the workflow step missing `continue-on-error: true`, a hang in the script becomes a hard merge-blocking failure rather than the silent informational gate the script header promises.

**Suggested fixes** (any one of, but the first two together are the cheapest and fully consistent):

- Add `continue-on-error: true` to the `Run security gates` step in `commitperclip-review.yml`.
- Cap pagination in `findExistingDraftAdvisory` (e.g., 20 pages = 2000 advisories) and short-circuit with a warning.
- Add an `AbortController` / fetch timeout per API call.

devinfoley:
## Thinking Path

> - The commitperclip review workflow runs a security gate as part of CI on every PR
> - The security script's header promises it always exits 0 and stays silent/informational, but PRs that triggered a flag were failing with a 5-minute timeout
> - Two compounding bugs: `findExistingDraftAdvisory` paginated without an upper bound, and the workflow step did not have `continue-on-error: true`, so any hang inside the script turned into a hard `review` check failure that blocked merge
> - This pull request caps the advisory pagination at 20 pages and adds `continue-on-error: true` to the workflow step, aligning runtime behavior with the script's documented "always exit 0" contract
> - The benefit is that future PRs flagged by the security gate no longer block merge on a 5-minute timeout, and the gate stays silent/informational as intended

Fixes: #7849

## What Changed

- `.github/workflows/commitperclip-review.yml`: added `continue-on-error: true` to the `Run security gates` step so a hang or non-zero exit cannot fail the `review` check (matches the script's documented "always exit 0" contract).
- `.github/scripts/check-pr-security.mjs`: capped `findExistingDraftAdvisory` pagination at 20 pages (= 2000 advisories) and short-circuited with a `console.warn` when the cap is hit; if no match is found within the cap, callers will simply create a new draft instead of hanging forever.
- `.github/scripts/tests/check-pr-security.test.mjs`: added a test asserting the pagination cap is enforced.

## Verification

- `node .github/scripts/tests/check-pr-security.test.mjs` — 31/31 pass, including the new cap test.
- Step-level guarantee: `continue-on-error: true` makes the `Run security gates` step non-blocking for the job, so even an unexpected hang/timeout in this step can no longer fail the `review` check.

## Risks

- Low risk. Pagination cap is a defensive bound; the worst case is a duplicate draft advisory (acceptable — the workflow continues). `continue-on-error: true` is exactly what the script header already promised; the workflow now matches its stated contract.

## Model Used

- Claude (claude-opus-4-7), extended thinking, tool use

## Checklist

### DIFF
--- a/.github/scripts/check-pr-security.mjs
+++ b/.github/scripts/check-pr-security.mjs
@@ -243,10 +243,14 @@ export async function syncDraftAdvisory(fetchImpl, token, repo, prNumber, prTitl
   });
 }
 
+// Cap pagination so a large backlog of unrelated draft advisories cannot stall
+// the security gate (it runs inside a 5-minute workflow timeout).
+const MAX_DRAFT_ADVISORY_PAGES = 20;
+
 export async function findExistingDraftAdvisory(fetchImpl, token, repo, prNumber) {
   const prMarker = `PR #${prNumber}`;
 
-  for (let page = 1; ; page += 1) {
+  for (let page = 1; page <= MAX_DRAFT_ADVISORY_PAGES; page += 1) {
     const advisories = await fetchImpl(
       `/repos/${repo}/security-advisories?state=draft&per_page=100&page=${page}`,
       token,
@@ -261,6 +265,12 @@ export async function findExistingDraftAdvisory(fetchImpl, token, repo, prNumber
 
     if (advisories.length < 100) return null;
   }
+
+  console.warn(
+    `[security] findExistingDraftAdvisory: hit ${MAX_DRAFT_ADVISORY_PAGES}-page cap without finding PR #${prNumber}; ` +
+    'treating as new advisory. A duplicate draft may be created.',
+  );
+  return null;
 }
 
 export async function postSecurityCheckRun(fetchImpl, token, repo, headSha, hasFlags) {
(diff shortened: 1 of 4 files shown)


devinfoley:
@greptile review

PR #7851, opened 2026-06-09
fix(server): never produce status=blocked with empty blockedBy[]

Nabaatxaneh:
## Summary

A `status=blocked` issue with an empty `blockedBy[]` set is unreachable: no blocker can be cleared to wake it, so it silently strands work and only an out-of-band sweep recovers it. Three writes in the recovery service and one in the `PATCH /api/issues/:id` agent path were producing this state. This PR enforces the invariant — *every* `status=blocked` write must be paired with a non-empty `blockedBy[]` — at all four sites, with route-level rejection for agent PATCH and a `status=todo` fallback inside recovery.

### Changes

**`server/src/services/recovery/service.ts`**
- `escalateStrandedRecoveryIssueInPlace`: read existing unresolved blockers and write `nextStatus = blockerIds.length > 0 ? "blocked" : "todo"`. Activity-log `status` field reflects the actual write.
- `escalateStrandedAssignedIssue` (initial write): same invariant; activity-log status updated.
- `escalateStrandedAssignedIssue` (re-block path): same invariant on the post-wake re-write.

**`server/src/routes/issues.ts`**
- PATCH handler now rejects agent writes that would leave the issue's effective status as `blocked` with no blockers, returning `400 { error: "blocked_requires_blocker", message: ... }`. Effective status considers both `updateFields.status` and the existing row; effective blockers consider both `req.body.blockedByIssueIds` and the existing relation summary. Board PATCH unchanged (the board is the authoritative human override).

### Tests

- **New** `server/src/__tests__/issue-blocked-requires-blocker-routes.test.ts` — 6 cases covering: agent rejected (no blockers / empty blockers array / clearing existing blockers), agent accepted with named blocker, agent accepted on no-status PATCH while already blocked-with-blocker, and board accepted (override path retained).
- **New** ALAA-965 describe block in `server/src/__tests__/issue-recovery-actions.test.ts` — 4 cases proving each of the three recovery paths (in-place, initial write, re-block) writes `blocked` when a live blocker exists and falls back to `todo` when none do.
- Existing recovery + heartbeat tests that asserted the old blocked-with-empty-blockers behavior are updated to `"todo"` with an ALAA-965 reference comment, since that *is* the bug being fixed.

### Backstory / production validation

We hit this in production: agents were writing `status=blocked` with no blockers, then the parent's status PATCH would self-trigger `issue_status_changed` wakes and the CEO had to run a per-heartbeat sweep (`scripts/ceo_unblock_sweep.py`) to PATCH affected issues back to `todo`. The fix was first deployed as a dist-JS hot-patch against `paperclipai@2026.529.0` running locally; this PR ports the same four edits onto the TypeScript source. Once a release including this PR lands, the local sweep policy and hot-patch can be retired.

## Test plan

- [x] `pnpm vitest run server/src/__tests__/issue-blocked-requires-blocker-routes.test.ts` — 6/6 pass
- [x] `pnpm vitest run server/src/__tests__/issue-recovery-actions.test.ts` — 25/25 pass (includes 4 new ALAA-965 cases)
- [x] `pnpm vitest run server/src/__tests__/heartbeat-process-recovery.test.ts` — 48/49 pass; the one remaining failure (`queues exactly one retry when the recorded local pid is dead`) reproduces unchanged on `upstream/master` (pre-existing flake, unrelated to this PR)
- [x] `tsc --noEmit` shows no new errors in any touched file

### DIFF
--- a/server/src/__tests__/heartbeat-process-recovery.test.ts
+++ b/server/src/__tests__/heartbeat-process-recovery.test.ts
@@ -828,7 +828,12 @@ describeEmbeddedPostgres("heartbeat orphaned process recovery", () => {
       .from(issues)
       .where(eq(issues.id, input.issueId))
       .then((rows) => rows[0] ?? null);
-    expect(sourceIssue?.status).toBe("blocked");
+    // ALAA-965: the recovery service no longer parks an issue in
+    // status=blocked with an empty blockedBy[] set. None of the heartbeat
+    // recovery fixtures seed an unresolved blocker on the source issue, so the
+    // recovery write falls back to "todo" and the recovery action's wake is
+    // what restores the live execution path.
+    expect(sourceIssue?.status).toBe("todo");
 
     return action;
   }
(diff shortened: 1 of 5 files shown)


Nabaatxaneh:
Superseded by #7850, which carries the same ALAA-965 fix rebased with the Greptile review findings addressed (blocked-only-with-blockers invariant regardless of recovery owner, re-fetched blocker list on the re-assert path, updated/added regression tests). Closing to keep a single canonical PR.

PR #7866, opened 2026-06-10
fix(security): block invite-based privilege escalation above actor role (#7786)

dosthcpp:
## Thinking Path

> - Paperclip companies have a human role hierarchy (viewer < operator < admin < owner) enforced by `humanRoleRank` for member management.
> - #7786: `POST /companies/:companyId/invites` only checks the `users:invite` permission and accepts any `humanRole` — so an admin can mint an **owner** invite and escalate themselves or others (verified repro in the issue).
> - The member-removal path already guards with `humanRoleRank` ("you can only remove users below your role"); the invite path simply never got the equivalent guard.
> - This PR validates the **effective** invite role (human invites default to `operator` when no role is requested) against the actor's resolved company role.
> - Non-board actors (agents) have no membership role; they are capped at the legacy default (`operator`) so ordinary agent-created invites keep working while privileged grants are blocked.
> - The benefit is the role hierarchy can no longer be bypassed via invites.

## What Changed

- `server/src/routes/access.ts` (invite creation route): compute the effective human role; resolve the actor's role via the existing `resolveActorHumanRole`; reject with **403** when `humanRoleRank[requested] > humanRoleRank[actor]` (agents capped at operator). Local-implicit/instance-admin actors resolve to owner as before.
- `server/src/__tests__/invite-create-route.test.ts`: 4 new route tests — admin→owner **403**, operator→admin **403**, admin→admin **201**, operator default-role invite **201**; the existing local-implicit test still passes.

## Verification

- `npx vitest run src/__tests__/invite-create-route.test.ts` → 5/5 pass.
- Server `tsc --noEmit` clean.

## Risks

- Low risk: reuses the established `humanRoleRank`/`resolveActorHumanRole` primitives. Owners and local-implicit/instance-admin actors are unaffected (rank 4 grants everything). The one behavior change beyond the vuln: agents can no longer mint admin/owner invites — they keep the operator-default path, which matches the hierarchy intent. Agent-only invites (`allowedJoinTypes: "agent"`) carry no human role and are untouched.

## Model Used

Claude Opus 4.8 (`claude-opus-4-8`), extended reasoning, via Claude Code (tool use + local test execution).

Closes #7786. Related contributions: #7864 (#7840), #7865 (#7656), #7819 (#7795), #7814 (#7790), #7809 (#7685).

## Dedup search

- [x] I searched the open and recently-closed GitHub PRs for similar or duplicate PRs (invite role validation / privilege escalation) and found none addressing this.

PR #7865, opened 2026-06-10
fix(issues): 422 instead of raw FK 500 for non-existent parentId/goalId on update (#7656)

dosthcpp:
## Thinking Path

> - Paperclip's issue API lets callers re-parent issues and re-link goals via `PATCH /api/issues/:id`.
> - #7656: supplying a `parentId` (or `goalId`) UUID that doesn't exist returns an HTTP 500 with a raw Drizzle/Postgres FK stack trace (`issues_parent_id_issues_id_fk`).
> - The route schema already enforces UUID *format*; what's missing is an *existence* check, so the FK violation surfaces as an unhandled 500 instead of a clean client error.
> - This PR validates both references (company-scoped) in the update path and throws 422 with the offending id, before the write.
> - It also rejects self-parenting explicitly.
> - The benefit is agents/operators get an actionable 4xx instead of a 500 that reads like a server bug.

## What Changed

- `server/src/services/issues.ts` (update path): if `parentId` is set, verify an issue with that id exists in the same company (and is not the issue itself); if `goalId` is set, verify the goal exists in the same company. Both failures throw `unprocessable` (422) with the offending id in details.
- `server/src/__tests__/issues-service.test.ts`: 5 embedded-Postgres tests — non-existent parentId, cross-company parentId, self-parent, non-existent goalId (all 422, row unchanged), and a valid same-company parentId still applies.

## Verification

- `npx vitest run src/__tests__/issues-service.test.ts` → all 67 pass (62 existing + 5 new).
- Server `tsc --noEmit` clean.

## Risks

- Low risk: two indexed-PK lookups added to the update path only when `parentId`/`goalId` are present in the patch. Valid updates behave identically (covered by the existing 62 tests + the valid-parent test). Cross-company parenting was already impossible at the DB level — it now fails with 422 instead of 500.

## Model Used

Claude Opus 4.8 (`claude-opus-4-8`), extended reasoning, via Claude Code (tool use + local embedded-Postgres test execution).

Closes #7656. Related contributions: #7864 (#7840), #7819 (#7795), #7814 (#7790), #7809 (#7685).

## Dedup search

- [x] I searched the open and recently-closed GitHub PRs for similar or duplicate PRs (parentId validation / FK 500) and found none addressing this.

PR #7864, opened 2026-06-10
fix(recovery): never regress a terminal issue to blocked on stranded escalation (#7840)

dosthcpp:
## Thinking Path

> - Paperclip's recovery sweep keeps assigned issues alive: it scans for stranded `todo`/`in_progress` issues and escalates them to `blocked` with a recovery action.
> - #7840 reports that cancelling a heartbeat run regresses its bound issue's status — observed `done → blocked`, `done → todo`, `in_review → blocked` — so completed tickets re-enter execution pipelines (data corruption on every governor-driven cancel).
> - The scan filters candidates by `inArray(status, ['todo','in_progress'])` at SELECT time, but the escalation write happens later with no re-check — a cancellation racing with completion (or a recovery action created before the issue closed) writes `blocked` over a terminal status.
> - Expected per the issue: a run cancellation must never move its bound issue out of a terminal state.
> - This PR adds a live status re-read at write time in both escalation paths and skips the escalation when the issue is already `done`/`cancelled`.
> - The benefit is that completed work can no longer be reopened by recovery racing a cancel.

## What Changed

- `server/src/services/recovery/service.ts`: new `isIssueTerminalNow()` helper; `escalateStrandedAssignedIssue` and `escalateStrandedRecoveryIssueInPlace` now re-read the issue's live status and return `null` (no status write, no recovery action) when it is `done`/`cancelled`.
- `server/src/__tests__/issue-recovery-actions.test.ts`: two embedded-Postgres tests — stale `in_progress` snapshot + live `done`/`cancelled` row → escalation returns null, status unchanged, no `issue_recovery_actions` row created.

## Verification

- `npx vitest run src/__tests__/issue-recovery-actions.test.ts` → all 23 pass (21 existing + 2 new).
- Server `tsc --noEmit` clean.

## Risks

- Low risk: the guard only *skips* an escalation that would have corrupted state; the normal stranded path (issue genuinely todo/in_progress) is unchanged, covered by the existing 21 tests. Legitimate reopen flows (human comment-reopen) live elsewhere and are untouched.

## Model Used

Claude Opus 4.8 (`claude-opus-4-8`), extended reasoning, via Claude Code (tool use + local embedded-Postgres test execution).

Closes #7840. Related contributions: #7819 (#7795 cold boot), #7814 (#7790 ews leak), #7809 (#7685 skills perf).

## Dedup search

- [x] I searched the open and recently-closed GitHub PRs for similar or duplicate PRs (cancellation / status regression / stranded recovery) and found none addressing this.

PR #7819, opened 2026-06-09
perf(adapters): lazy-load acpx-local execute/test to cut cold-boot codex-acp import (#7795)

dosthcpp:
## Thinking Path

> - Paperclip runs agents through pluggable adapters; the server adapter registry wires up every builtin adapter at module scope.
> - #7785 traced cold serverless boot hangs (10s–300s on `/api/health` + `get-session`, blocking login/nav) to `boot-timing: module-import` being dominated by `@zed-industries/codex-acp` (~165MB).
> - That dependency is pulled by a single static edge: the registry imports the `acpx-local/server` barrel, which re-exports `execute.ts` + `test.ts`, which import codex-acp — all evaluated eagerly on every cold boot.
> - An audit confirms acpx-local is the ONLY builtin adapter importing a heavy SDK at module scope, so breaking this one edge removes the dominant cost.
> - This PR splits the import: the codex-acp-free pieces load statically; `execute`/`testEnvironment` become dynamic-import wrappers, so codex-acp is paid only when a run actually executes.
> - The benefit is cold boots no longer evaluate ~165MB of adapter runtime, cutting the module-import hang.

## What Changed

- `packages/adapters/acpx-local/package.json`: add `./server/{execute,test,config-schema,session-codec,skills}` sub-path exports (both the dev `src` map and the `publishConfig` `dist` map), so consumers can import the lightweight modules without pulling the heavy barrel.
- `server/src/adapters/registry.ts`: import `sessionCodec`/`getConfigSchema`/`listAcpxSkills`/`syncAcpxSkills` from the clean sub-paths; replace the static `execute`/`testEnvironment` imports with lazy `import()` wrappers.
- `server/src/__tests__/adapter-registry.test.ts`: assert the `acpx_local` adapter keeps `execute`/`testEnvironment` callable, `sessionCodec` wired, and `getConfigSchema()` resolves via the clean sub-path.

## Verification

- `npx tsc --noEmit` (server) clean with the new exports resolved.
- `npx vitest run src/__tests__/adapter-registry.test.ts -t "#7795"` → passes (adapter interface preserved, clean sub-path resolves).
- Audit: `grep` across `packages/adapters/*/src` shows acpx-local is the only adapter importing a heavy SDK (codex-acp/acpx) at module scope.

## Risks

[...]

PR #7867, opened 2026-06-10
fix(db): ON DELETE actions for remaining run-id FKs — pruned runs no longer 500 run-stamped writes (#7450)

### LINKED ISSUE #7450: Bug: run-stamped writes 500 when gateway JWT's run_id is pruned from heartbeat_runs (FK has no ON DELETE SET NULL)
## Summary

Agent comment writes (and likely all run-stamped writes) fail with `500 Internal server error` once the `run_id` embedded in a long-lived `mcp_gateway` JWT has been pruned from `heartbeat_runs`. Reads are unaffected. The failure is silent and opaque — it only surfaces as a generic 500.

## Environment

- `paperclipai` **2026.517.0**, `local_trusted` deployment, embedded-postgres.
- (We are updating to the latest `2026.529.0` to check whether this is already fixed upstream; filing regardless because the root cause is structural.)

## Symptom

```
POST /api/issues/{id}/comments  ->  500 {"error":"Internal server error"}
```

Both the MCP tool `comment_on_issue` and direct REST POSTs fail for the affected token. `GET`/reads work fine.

## Root cause

From `server.log`:

```
insert or update on table "issue_comments" violates foreign key constraint
"issue_comments_created_by_run_id_heartbeat_runs_id_fk"
```

The comment INSERT stamps `created_by_run_id` with the `run_id` baked into the calling agent's `mcp_gateway` JWT. Long-lived gateway tokens carry a fixed `run_id` from when they were minted. Once that heartbeat run is pruned from `heartbeat_runs`, the FK on `issue_comments.created_by_run_id -> heartbeat_runs.id` has **no `ON DELETE SET NULL`**, so the INSERT is rejected.

## Blast radius

This is latent in **every** long-lived agent gateway token: it triggers the moment that token's embedded run is pruned. `issue_comments.created_by_run_id` is only the first surface — the schema has many run-id FK columns referencing `heartbeat_runs` (e.g. `createdByRunId`, `activeRunId`, `sourceRunId`, `checkoutRunId`, `startedByRunId`, …). Any of those without `ON DELETE SET NULL` is the same bug waiting to happen. Because it degrades to an opaque 500, agents can silently lose the ability to write while reads keep working.

## Reproduction

1. Mint/use an agent `mcp_gateway` JWT whose embedded `run_id` exists in `heartbeat_runs`.
2. Let that heartbeat run be pruned (or delete the row).
3. `POST /api/issues/{id}/comments` with that token → `500`.

## Suggested fix

Two layers, ideally both:

1. **Defensive null-stamp (code, no migration):** when stamping run-id columns, only set `created_by_run_id` (and siblings) if the run still exists; otherwise insert `NULL`. Prevents the 500 immediately and is migration-free.
2. **`ON DELETE SET NULL` on the run-id FKs (schema):** so pruning a `heartbeat_runs` row nulls the references instead of leaving dangling-but-rejected inserts, and keeps history consistent.

Either alone stops the 500; together they keep referential integrity clean over time.

## Workaround for other operators

If you hit this: the **runtime-injected** `PAPERCLIP_API_KEY` env var (present in each spawned run's process env) carries the *current live* `run_id`, unlike a static gateway token in `~/.claude.json`. POSTing to REST with that env token succeeds while the stale MCP-gateway token keeps 500ing.

dosthcpp:
## Thinking Path

> - Paperclip stamps many rows with the heartbeat run that produced them, and agent gateway JWTs embed a fixed `run_id` from when they were minted.
> - #7450: once that run is pruned from `heartbeat_runs`, any write stamping the stale id hits a raw FK violation and surfaces as an opaque 500 — latent in every long-lived gateway token.
> - Most run-stamped columns (incl. `issue_comments.created_by_run_id`) already have `ON DELETE SET NULL`; an audit of the schema found five FKs that were missed.
> - One of them (`cost_events.heartbeat_run_id`) is also the first blocker in #7781's company-deletion failure, so this shrinks that issue too.
> - This PR sets `ON DELETE SET NULL` on the four stamp columns and `ON DELETE CASCADE` on `heartbeat_run_events.run_id` (events are owned by their run and are NOT NULL, so set-null is impossible and orphaned events are meaningless).
> - The benefit is that pruning a run can never again break writes or deletes that touch rows referencing it.

## What Changed

- Schema: `activity_log.run_id`, `agent_task_sessions.last_run_id`, `cost_events.heartbeat_run_id`, `finance_events.heartbeat_run_id` → `onDelete: "set null"`; `heartbeat_run_events.run_id` → `onDelete: "cascade"`.
- Migration `0094_wakeful_gargoyle.sql` (generated with `drizzle-kit generate`; numbering check passes) re-creates the five constraints with the new actions.
- New embedded-Postgres test `heartbeat-run-prune-fk.test.ts`: seeds a run with cost-event, activity-log, run-event, and task-session children, deletes the run, and asserts stamps null out and run events cascade.

## Verification

- `npx vitest run src/__tests__/heartbeat-run-prune-fk.test.ts` → passes against a database migrated with 0094 (the embedded harness applies all migrations, so this also proves the migration applies cleanly).
- `packages/db` `tsc` clean; `check-migration-numbering` passes.

## Risks

- Low/medium (schema migration). The ALTERs only change FK delete actions — no data rewrite, brief locks only. Set-null matches the established pattern on the sibling columns. The one judgment call is cascade on `heartbeat_run_events` (NOT NULL column): events have no meaning without their run, and agent/company deletion paths already delete them explicitly today.

## Model Used

Claude Opus 4.8 (`claude-opus-4-8`), extended reasoning, via Claude Code (tool use + drizzle-kit + local embedded-Postgres test execution).

Closes #7450. Refs #7781 (removes its first FK blocker). Related contributions: #7866, #7864, #7865, #7819, #7814, #7809.

## Dedup search

- [x] I searched the open and recently-closed GitHub PRs for similar or duplicate PRs (run_id FK / ON DELETE SET NULL / prune) and found none addressing these five constraints.

### DIFF
--- a/packages/db/src/migrations/0094_wakeful_gargoyle.sql
+++ b/packages/db/src/migrations/0094_wakeful_gargoyle.sql
@@ -1,0 +1,15 @@
+ALTER TABLE "activity_log" DROP CONSTRAINT "activity_log_run_id_heartbeat_runs_id_fk";
+--> statement-breakpoint
+ALTER TABLE "agent_task_sessions" DROP CONSTRAINT "agent_task_sessions_last_run_id_heartbeat_runs_id_fk";
+--> statement-breakpoint
+ALTER TABLE "cost_events" DROP CONSTRAINT "cost_events_heartbeat_run_id_heartbeat_runs_id_fk";
+--> statement-breakpoint
+ALTER TABLE "finance_events" DROP CONSTRAINT "finance_events_heartbeat_run_id_heartbeat_runs_id_fk";
+--> statement-breakpoint
+ALTER TABLE "heartbeat_run_events" DROP CONSTRAINT "heartbeat_run_events_run_id_heartbeat_runs_id_fk";
+--> statement-breakpoint
+ALTER TABLE "activity_log" ADD CONSTRAINT "activity_log_run_id_heartbeat_runs_id_fk" FOREIGN KEY ("run_id") REFERENCES "public"."heartbeat_runs"("id") ON DELETE set null ON UPDATE no action;--> statement-breakpoint
+ALTER TABLE "agent_task_sessions" ADD CONSTRAINT "agent_task_sessions_last_run_id_heartbeat_runs_id_fk" FOREIGN KEY ("last_run_id") REFERENCES "public"."heartbeat_runs"("id") ON DELETE set null ON UPDATE no action;--> statement-breakpoint
+ALTER TABLE "cost_events" ADD CONSTRAINT "cost_events_heartbeat_run_id_heartbeat_runs_id_fk" FOREIGN KEY ("heartbeat_run_id") REFERENCES "public"."heartbeat_runs"("id") ON DELETE set null ON UPDATE no action;--> statement-breakpoint
+ALTER TABLE "finance_events" ADD CONSTRAINT "finance_events_heartbeat_run_id_heartbeat_runs_id_fk" FOREIGN KEY ("heartbeat_run_id") REFERENCES "public"."heartbeat_runs"("id") ON DELETE set null ON UPDATE no action;--> statement-breakpoint
+ALTER TABLE "heartbeat_run_events" ADD CONSTRAINT "heartbeat_run_events_run_id_heartbeat_runs_id_fk" FOREIGN KEY ("run_id") REFERENCES "public"."heartbeat_runs"("id") ON DELETE cascade ON UPDATE no action;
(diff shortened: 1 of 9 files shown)


dosthcpp:
Re-verified against master `696c694a` (2026-07-07): still unaddressed — run-id FKs still default to `ON DELETE no action` (e.g. `heartbeat_run_events.ts:11`) — pruning runs 500s on stamped rows. Will rebase on request. Triage summary of all our open PRs: #8180.

PR #7814, opened 2026-06-09
fix(issues): close linked execution workspace on terminal issue transition (#7790)

dosthcpp:
## Thinking Path

> - Paperclip orchestrates AI agents; each issue/heartbeat can provision an `execution_workspaces` (ews) row for its sandbox.
> - Those rows track lifecycle via `closed_at` / `cleanup_eligible_at`, but #7790 found those columns are written in exactly one place — the explicit archive endpoint.
> - So every issue that finishes leaves its workspace stuck at `status='active'`, `closed_at IS NULL` forever — a steadily growing lifecycle leak with no reaper to catch it.
> - The primary, highest-volume gap is the issue terminal transition: `applyStatusSideEffects` stamps the issue done/cancelled but never touches the linked workspace.
> - This PR closes the linked workspace in the same update transaction when an issue goes terminal, guarded so shared workspaces stay open while other work still references them.
> - The benefit is that finished issues stop leaking active workspace rows, and closed rows become cleanup-eligible.

## What Changed

- `server/src/services/issues.ts`: in the issue-update transaction, when status becomes `done`/`cancelled` and the issue has a linked `executionWorkspaceId`, set the workspace `status='archived'`, `closed_at`, `cleanup_eligible_at`, `cleanup_reason='issue_<status>'`. Only fires when no other non-terminal issue references the workspace (shared-workspace safe) and only when `closed_at IS NULL` (idempotent).
- `server/src/__tests__/issues-service.test.ts`: embedded-Postgres tests for done + cancelled closing a dedicated workspace, and a shared workspace staying open while another active issue references it.

## Verification

- `npx vitest run src/__tests__/issues-service.test.ts -t "#7790"` → 3 pass.
- Server `tsc --noEmit` clean.

## Risks

- Low risk: additive, runs inside the existing update transaction; the shared-workspace reference guard prevents closing a workspace still in use, and the `closed_at IS NULL` guard makes it idempotent. Does not perform runtime/fs teardown (that remains the reaper/teardown path's job via `cleanup_eligible_at`). Note: #7790 also lists two further gaps (soft project-archive cascade, and a periodic reaper) — out of scope here; this fixes the dominant issue-transition leak.

## Model Used

Claude Opus 4.8 (`claude-opus-4-8`), extended reasoning, via Claude Code (tool use + local embedded-Postgres test execution).

Closes #7790. Related contributions: #7809, #7792.

## Dedup search

[...]

PR #7809, opened 2026-06-09
perf(skills): O(agents+skills) attached-agent count in skill list (#7685)

dosthcpp:
## Thinking Path

> - Paperclip orchestrates AI agents for zero-human companies; the Skills API backs the agent/skill management UI.
> - `GET /api/companies/{companyId}/skills` was reported at 24.4s response time at scale, blocking that UI.
> - Root cause: `company-skills.ts` `list()` re-ran `resolveDesiredSkillKeys` for every (skill × agent) pair via a per-skill agent filter — O(N×M).
> - That resolver depends only on the agent config, not the skill, so the per-skill recomputation is pure waste.
> - This PR resolves desired keys once per agent, tallies counts into a Map, then does an O(1) lookup per skill — O(N+M).
> - The benefit is a Skills API that stays well under 500ms at 200+ skills / 200+ agents.

## What Changed

- `server/src/services/company-skills.ts`: replace the O(N×M) per-skill agent filter with a Map-based tally; extract `tallyAttachedAgentCounts` (each agent counted at most once per key — preserves prior semantics).
- `server/src/__tests__/company-skills.test.ts`: unit tests for the tally (counts, per-agent dedupe, empty case).

## Verification

- `npx vitest run src/__tests__/company-skills.test.ts -t tallyAttachedAgentCounts` → 3 pass.
- Server `tsc --noEmit` clean.

## Risks

- Low risk: pure refactor with identical counting semantics; no API/schema change.

## Model Used

Claude Opus 4.8 (`claude-opus-4-8`), extended reasoning, via Claude Code (tool use + local test execution).

Closes #7685. Related batch PRs: #7792, #7793.

## Dedup search

- [x] I searched the open and recently-closed GitHub PRs for similar or duplicate PRs and confirmed there is no overlap.

PR #7869, opened 2026-06-10
fix(plugins): plugin-key UI lookups fall back instead of 500ing on wrapped 22P02 (#7639)

### LINKED ISSUE #7639: Plugin UI static route returns 500 for plugin-key IDs (UUID/key fallback checks wrong error field on Drizzle-wrapped pg error)
## Summary

`GET /_plugins/:pluginId/ui/*` (plugin UI static serving) returns **HTTP 500** when `:pluginId` is a **plugin key** (manifest `id`, e.g. `agent-pixels.camera`) instead of the database UUID. The route is documented to accept *either* form, and the code has a UUID-vs-key fallback — but the fallback never triggers because it inspects the wrong error field on the Drizzle-wrapped Postgres error.

Tested on **Paperclip 0.3.1** (self-hosted Docker, `deploymentMode: authenticated`).

## Root cause

In `server/src/routes/plugin-ui-static.ts` (~line 206–217):

```ts
let plugin = null;
try {
  plugin = await registry.getById(pluginId);
} catch (error) {
  const maybeCode =
    typeof error === "object" && error !== null && "code" in error
      ? error.code
      : undefined;
  if (maybeCode !== "22P02") {   // <-- bug
    throw error;
  }
}
if (!plugin) {
  plugin = await registry.getByKey(pluginId);
}
```

When `pluginId` is not a UUID, Postgres raises `22P02` (`invalid input syntax for type uuid`). The intent is to swallow that and fall back to `getByKey`. But drizzle-orm wraps the driver error in a `DrizzleQueryError`, so the `22P02` code lives on **`error.cause.code`**, not `error.code`. `maybeCode` is therefore `undefined`, the condition `undefined !== "22P02"` is `true`, and the original error is **re-thrown → 500**. `getByKey` is never reached.

Observed server log:

```
GET /_plugins/agent-pixels.camera/ui/index.js 500 — Failed query: select ... from "plugins" where "plugins"."id" = $1
params: agent-pixels.camera
caused by: PostgresError: invalid input syntax for type uuid: "agent-pixels.camera"
    at async .../server/src/routes/plugin-ui-static.ts:247
```

## Why it breaks real plugins (not just direct key-URL access)

The host loads plugin UI bundles via `URL.createObjectURL(blob)` + `await import(blobUrl)`, so inside the plugin module `import.meta.url` is a `blob:` URL. Plugins that resolve their own asset base from `import.meta.url` fall back to a hardcoded `/_plugins/<manifest-id>/ui/assets/` path (manifest id, not UUID). That path hits this 500. The plugin then does `fetch(assetsIndex).then(r => r.json())`, receives `{"error":"Internal server error"}` (valid JSON, so no parse error), and crashes downstream with `Cannot read properties of undefined (reading 'map')` when iterating the (missing) asset arrays.

Concretely reproduced with the **Agent Pixels** plugin (`agent-pixels.camera`), whose camera page shows: `Camera failed: Cannot read properties of undefined (reading 'map')`.

## Reproduction

1. Install any UI plugin whose manifest `id` is not a UUID (e.g. `agent-pixels.camera`).
2. `curl -i http://localhost:3100/_plugins/<db-uuid>/ui/index.js` → **200**.
3. `curl -i http://localhost:3100/_plugins/agent-pixels.camera/ui/index.js` → **500** (`invalid input syntax for type uuid`).

The UUID form works; the key form 500s — they should behave identically per the route's documented contract.

## Suggested fix

Check the wrapped cause (and/or just fall through to `getByKey` on any lookup failure):

```ts
function pgErrorCode(err: unknown): string | undefined {
  for (let e: any = err; e; e = e.cause) {
    if (e && typeof e === "object" && typeof e.code === "string") return e.code;
  }
  return undefined;
}

let plugin = null;
try {
  plugin = await registry.getById(pluginId);
} catch (error) {
  if (pgErrorCode(error) !== "22P02") throw error;  // not a "not-a-uuid" error
}
if (!plugin) {
  plugin = await registry.getByKey(pluginId);
}
```

Alternatively, detect UUID shape before querying and route to `getById` vs `getByKey` directly, avoiding the thrown-and-caught DB error entirely.

## Environment

- Paperclip: 0.3.1
- Deployment: self-hosted Docker, single-node, `authenticated` / `private`
- drizzle-orm: 0.45.2, postgres: 3.4.8

dosthcpp:
## Thinking Path

> - Paperclip plugins ship UI bundles served by `GET /_plugins/:pluginId/ui/*`, which is documented to accept either the plugin's DB UUID or its manifest key (e.g. `agent-pixels.camera`).
> - #7639: every plugin-key URL returns 500 — the host's own plugin UI loading breaks, not just direct key access.
> - The route has a UUID-vs-key fallback that swallows Postgres `22P02` (invalid uuid input) and retries `getByKey` — but it reads `error.code`, while drizzle-orm wraps driver errors in `DrizzleQueryError`, putting the code on `error.cause.code`.
> - So the condition never matched, the wrapped error re-threw as a 500, and `getByKey` was unreachable on that path.
> - This PR adds a small, bounded cause-chain walker and uses it for the 22P02 check so the documented fallback actually triggers.
> - The benefit is plugin UI assets load by key again, matching the documented contract.

## What Changed

- `server/src/routes/plugin-ui-static.ts`: new exported `errorChainHasPgCode(error, code)` — walks `error.cause` up to 5 levels (cycle-safe) looking for the Postgres code; the lookup's 22P02 check now uses it. (Only occurrence of this pattern in the repo — verified by grep.)
- `server/src/__tests__/plugin-ui-static-pg-code.test.ts`: 5 unit tests — raw driver error, drizzle-wrapped cause, double-wrapped, non-matching code, non-object inputs, and a self-referential cause chain.

## Verification

- `npx vitest run src/__tests__/plugin-ui-static-pg-code.test.ts` → 5/5 pass.
- Server `tsc --noEmit` clean.

## Risks

- Low risk: strictly widens the swallowed-error match from `error.code === '22P02'` to "22P02 anywhere in the cause chain"; all other errors still re-throw. UUID lookups are unaffected.

## Model Used

Claude Opus 4.8 (`claude-opus-4-8`), extended reasoning, via Claude Code (tool use + local test execution).

Closes #7639. Related contributions: #7867, #7866, #7864, #7865, #7819, #7814, #7809.

## Dedup search

- [x] I searched the open and recently-closed GitHub PRs for similar or duplicate PRs (plugin UI 500 / 22P02 / DrizzleQueryError) and found none addressing this.

### DIFF
--- a/server/src/__tests__/plugin-ui-static-pg-code.test.ts
+++ b/server/src/__tests__/plugin-ui-static-pg-code.test.ts
@@ -1,0 +1,36 @@
+import { describe, expect, it } from "vitest";
+import { errorChainHasPgCode } from "../routes/plugin-ui-static.js";
+
+describe("errorChainHasPgCode (#7639)", () => {
+  it("matches a code on the error itself (raw driver error)", () => {
+    const raw = Object.assign(new Error("invalid input syntax for type uuid"), { code: "22P02" });
+    expect(errorChainHasPgCode(raw, "22P02")).toBe(true);
+  });
+
+  it("matches a code on error.cause (DrizzleQueryError wrapping)", () => {
+    const pg = Object.assign(new Error("invalid input syntax for type uuid"), { code: "22P02" });
+    const wrapped = new Error("Failed query: select ...", { cause: pg });
+    expect(errorChainHasPgCode(wrapped, "22P02")).toBe(true);
+  });
+
+  it("matches a code nested two causes deep", () => {
+    const pg = Object.assign(new Error("boom"), { code: "22P02" });
+    const mid = new Error("mid", { cause: pg });
+    const outer = new Error("outer", { cause: mid });
+    expect(errorChainHasPgCode(outer, "22P02")).toBe(true);
+  });
+
+  it("does not match a different code or non-error values", () => {
+    const pg = Object.assign(new Error("fk"), { code: "23503" });
+    expect(errorChainHasPgCode(new Error("x", { cause: pg }), "22P02")).toBe(false);
+    expect(errorChainHasPgCode(new Error("plain"), "22P02")).toBe(false);
+    expect(errorChainHasPgCode(null, "22P02")).toBe(false);
+    expect(errorChainHasPgCode("22P02", "22P02")).toBe(false);
+  });
+
+  it("terminates on a self-referential cause chain", () => {
+    const cyclic: Record<string, unknown> = { code: "XXXXX" };
+    cyclic.cause = cyclic;
+    expect(errorChainHasPgCode(cyclic, "22P02")).toBe(false);
+  });
+});
--- a/server/src/routes/plugin-ui-static.ts
+++ b/server/src/routes/plugin-ui-static.ts
@@ -176,6 +176,21 @@ function computeETag(size: number, mtimeMs: number): string {
   return `"${hash}"`;
 }
 
+/**
+ * True when the error — or anything in its `cause` chain — carries the given
+ * Postgres error code. drizzle-orm wraps driver errors in DrizzleQueryError,
+ * so codes like 22P02 (invalid uuid input) live on `error.cause.code`, not
+ * `error.code` (GH #7639). Exported for unit testing.
+ */
+export function errorChainHasPgCode(error: unknown, pgCode: string): boolean {
+  let current: unknown = error;
+  for (let depth = 0; depth < 5 && typeof current === "object" && current !== null; depth += 1) {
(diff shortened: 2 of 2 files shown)


dosthcpp:
Re-verified against master `696c694a` (2026-07-07): still unaddressed — `plugin-ui-static.ts:253` still checks only top-level `error.code`, not the wrapped drizzle cause chain → 22P02 500s. Will rebase on request. Triage summary of all our open PRs: #8180.

Issue #7841, opened 2026-06-09
Self-sustaining comment-wake retry loop on completed issues (cancellation queues another wake)


ianbruton:
**Version:** 2026.428.0 (local_trusted)

**Observed:** after an issue completed, an automation-posted completion comment (from a plugin, author `nexus-merge-agent`) appeared to queue an agent wakeup on that issue. The resulting run was cancelled by our plugin-side volume governor — and the cancellation itself queued a retry wake, plus "Recover stalled issue X" recovery issues, each with fresh runs. The chain was self-sustaining: ~20 runs created-and-cancelled in ~15 minutes for one completed ticket, with recovery issues (each getting its own runs) compounding it. Combined with the status regression in the companion issue, the completed ticket kept re-entering the pipeline.

**Expected (any of these would break the loop):**
- don't queue comment-wakes for issues already in a terminal status;
- treat run cancellation as a deliberate veto — no automatic retry wake;
- cap recovery-issue chains ("recover stalled X" → cancelled → "recover stalled recover" should not recurse).

**Context:** our plugin counts every run via `agent.run.started` against a daily per-company budget and cancels over-budget runs via `POST /api/heartbeat-runs/:id/cancel`. That governor contained the loop (nothing executed), but the host kept feeding it until we paused the company's agents and drained `agent_wakeup_requests`.

Companion issue: run cancellation regresses bound issue status.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

ianbruton:
Additional evidence from a second occurrence tonight: the retry asymmetry is the crux.

- A run that **fails at dispatch** (e.g. agent not invokable) retries once, then stops — self-limiting.
- A run that is **cancelled** queues another automation wake every time — unbounded. Our governor cancelled ~50 runs in ~20 min across two episodes (wake source `automation/system`); the loop only died when we disabled the cancelling plugin entirely and let dispatch fail against paused agents.

So any external component that cancels runs (budget governors, ops tooling) currently *creates* load instead of shedding it. Treating cancellation like dispatch-failure (bounded retries) would fix both this and the status-regression in #7840.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

schmelli:
## Heartbeat Priority Assessment

**Priority: HIGH** — Bug affecting data integrity, system lifecycle, or deployment stability. Needs resolution before launch.

_This is a classification comment from the heartbeat routine._

schmelli:
[...]

Issue #7840, opened 2026-06-09
Heartbeat run cancellation regresses the bound issue's status (done → blocked/todo)


ianbruton:
**Version:** 2026.428.0 (local_trusted)

**Observed:** when a heartbeat run bound to an issue is cancelled (`POST /api/heartbeat-runs/:id/cancel`), the bound issue's status is moved to `blocked` — and on subsequent retry/cancel cycles we observed `done → blocked`, `done → todo`, and `in_review → blocked` transitions on issues whose work was already complete and merged.

**Why it hurts:** an external governor (plugin) that cancels runs — e.g. budget/volume enforcement — inadvertently *reopens completed work*. The regressed status then makes the issue eligible for pickup again, so completed tickets re-enter execution pipelines.

**Expected:** cancellation of a run should never move its bound issue *out of a terminal status* (`done`/`cancelled`). Marking an in-flight issue `blocked` on cancel is reasonable; resurrecting finished work is not.

**Repro sketch:**
1. Issue X reaches `done`.
2. A queued/started run bound to X (e.g. from a comment wake) is cancelled via the API.
3. X's status is now `blocked` or `todo`.

Happy to provide logs/timestamps from our instance.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

schmelli:
## Heartbeat Priority Assessment

**Priority: HIGH** — Bug affecting data integrity, system lifecycle, or deployment stability. Needs resolution before launch.

_This is a classification comment from the heartbeat routine._

schmelli:
**🔴 Heartbeat Triage (Sam): CRITICAL — UX regression**
- **Type:** UX Regression (heartbeat run cancellation regresses issue status done→blocked/todo)
- **Priority:** CRITICAL
- **Why:** Directly breaks issue state integrity. This corrupts user data on every heartbeat run cancellation.
- **Action:** Needs assignment & fix.

schmelli:
**🔄 Heartbeat #002 Triage (Sam):**
✅ PR #7864 (by @dosthcpp) is open and MERGEABLE — prevents terminal status regression on run cancellation.

**Next step:** Needs a maintainer to review and merge.

**Status:** 🟡 Blocked waiting on maintainer review of PR #7864

schmelli:
**🧩 Sam's Heartbeat Assessment — Priority: HIGH (Bug)**

[...]

PR #7872, opened 2026-06-10
fix(heartbeat): cancellation is a veto, no retry/comment wakes on terminal issues (GH #7841)

dosthcpp:
## Thinking Path

> - Paperclip orchestrates AI agents for zero-human companies; the heartbeat loop wakes agents and recovers stalled runs.
> - When a run is cancelled (budget governor, agent pause, human/control-plane veto), the recovery path treated it like a transient failure.
> - That caused two self-sustaining loops (GH #7841): an auto-recovery retry was queued off the cancelled run, and deferred system follow-up wakes were re-queued onto already-terminal issues.
> - A governor that cancels an over-budget run would thus have the cancellation immediately feed itself another run — churn.
> - This pull request treats a cancellation as a deliberate veto: it releases the execution lock but queues no automatic recovery retry, and suppresses deferred system follow-up wakes on done/cancelled issues (genuine comment/mention deliveries still promote).
> - The benefit is that cancelling a run actually stops it, ending the #7841 cancel/retry churn while preserving cross-agent handoffs.

## What Changed

- `heartbeat.ts`: removed `cancelled` from the immediate-recovery trigger (only `failed`/`timed_out` are eligible); a cancellation now releases the lock without queueing a recovery run.
- `heartbeat.ts`: deferred system follow-up wakes (retry/cleanup/continuation) are suppressed when the bound issue is already `done`/`cancelled` and there are no genuine comment/mention deliveries to promote.
- Added a regression test asserting a cancelled run queues no recovery run and no `issue_continuation_needed` wake, and leaves the issue `in_progress` with the execution lock released.

## Verification

- `pnpm --filter @paperclipai/server vitest run src/__tests__/heartbeat-process-recovery.test.ts` — the new "deliberate veto" test passes, and the adjacent "operator interrupt" test (added on master) also passes after rebase.
- Rebased cleanly onto current `master`; the only conflict was two independent new tests inserted at the same location — both retained.

## Risks

Low. Scoped to the recovery/heartbeat path. Real failures (`failed`/`timed_out`) still recover exactly as before; only deliberate cancellations stop queueing retries. Comment/mention deliveries are still promoted, so cross-agent handoffs are unaffected.

## Model Used

## Checklist

- [x] I searched the GitHub PR list for similar/duplicate PRs (this is the canonical #7841 cancel-loop fix; companion to #7864 terminal-status guard).


dosthcpp:
Closing as a duplicate of #7889 (same GH #7841 fix).

- This branch (`ton-2322-cancel-no-retry-wake`) also carries **unrelated UI work** (commit f1b2bf23, `fix(ui): stop Activity tab... (TON-2167)`: SectionErrorBoundary / Activity / IssueDetail), so its diff is contaminated for this issue.
- #7889 carries the clean, load-bearing change only: cancellation is a deliberate veto, dropped from the immediate-recovery trigger in `releaseIssueExecutionAndPromote`, with a focused regression test. It has been rebased onto current `master` (was 157 behind) and the test passes.
- The terminal-issue comment/system-wake suppression half is already handled by the merged #7864 (#7840 terminal-status guard), so #7889 completes the loop fix without the extra defensive hunk.

Consolidating on #7889.

Issue #7795, opened 2026-06-09
CRITICAL: Cold serverless boot hangs 10s-300s — blocks login & navigation (#7785 action required)


schmelli:
## HEARTBEAT ESCALATION — CTO DIRECTIVE

See full analysis at #7785.

**This is our #1 launch blocker.** Every cold boot blocks all user access for 10-300 seconds.

### Required actions (in priority order):
1. Hard-cap every boot-path I/O at 2-3s fail-fast
2. Lazy-import heavy deps (sqlite3, claude-agent-sdk, playwright-core, codex-acp) — worker only
3. Add Vercel cron warmer hitting /api/health + get-session
4. Don't redirect to /auth on transient timeout — only on real 401

**Assigning to me for tracking. Execution needed within 24h.**

schmelli:
**🚨 PM Heartbeat Triage:** This is a launch-blocking issue. Needs assignee and prioritization immediately.

schmelli:
**PM HEARTBEAT TRIAGE (2026-06-09):** 

This is a **CRITICAL V1 launch blocker** — cold serverless boots blocking login/navigation for 10s–300s.

Root cause has been identified in #7785 (comment by @ValDola-stack): static import chain pulling in ~165MB of `@zed-industries/codex-acp` at module scope via the adapter registry. 

**Recommended fix path (from #7785):**
1. Lazy-load adapter `execute` implementations in `routes/adapters.ts` / `registry.ts` 
2. OR split registry into lightweight metadata module + heavy runtime module

**Action needed:** This needs ownership assigned. @team-leads — please assign a developer to implement the lazy-load fix.

schmelli:
## 🚨 CTO DIRECTIVE: SOFORTIGE MASSNAHME ERFORDERLICH

Dieser Issue ist unser **#1 V1-Launch-Blocker** und hat keinen Assignee. Das ist nicht akzeptabel.

**Root Cause** (aus #7785): Statischer Import von `@zed-industries/codex-acp` (~165MB) im Vercel-Funktions-Bundle via Adapter-Registry.

**Fix-Path:**
1. Lazy-import (`import()`) aller schweren Adapter-Implementierungen in `routes/adapters.ts` / `registry.ts`
2. Hard-cap jede Boot-Path I/O bei 2-3s fail-fast
3. Vercel Cron-Warmer für /api/health + get-session

[...]

PR #7898, opened 2026-06-10
fix(ui): don't redirect to /auth on transient session-fetch failure (#7795)

dosthcpp:
## Thinking Path

> - Paperclip orchestrates AI agents for zero-human companies; humans reach the board through the web UI.
> - The web UI gates every authenticated route behind `CloudAccessGate`, which resolves the session via `/api/auth/get-session`.
> - On serverless cold boots (issue #7795) that request can hang, time out, or return 5xx while the function is still booting.
> - `CloudAccessGate` treated *any* missing session data as signed-out and redirected to `/auth` — so a slow boot looked like a logout and broke the login/navigation flow.
> - But `authApi.getSession()` already resolves `null` only on a real 401 and throws on every other failure; the gate just ignored that distinction.
> - This pull request makes the gate redirect only on a server-confirmed 401, and show a retryable error state on transient fetch failure.
> - The benefit is that cold-boot users stay on their page (with a Retry) instead of being bounced to login — directly addressing item 4 of the #7795 directive.

## What Changed

- `CloudAccessGate` now redirects to `/auth` only when the session query **succeeded with a null session** (`sessionQuery.isSuccess && !sessionQuery.data`), i.e. a real 401.
- On a failed session fetch (timeout / network / 5xx), the gate renders an explicit error state with a **Retry** button and the message "You have not been signed out", instead of redirecting.

## Verification

- `pnpm --filter @paperclipai/ui vitest run src/App.test.tsx` — 7/7 pass.
- Two new `CloudAccessGate` tests: (a) a server-confirmed null session still redirects to `/auth`; (b) a transient fetch failure shows Retry, does not redirect, and recovers to the board when Retry succeeds.
- **Negative-verified:** the transient-failure test fails against the unfixed component (it redirects) and passes with the fix.
- `tsc -b` clean. Edit region is identical between the PR base and current `master`, so it merges cleanly.

## Risks

Low risk. Single UI component; no API, schema, or behavioral change for the real-401 path (still redirects exactly as before). The only new path is the previously-unhandled fetch-error case, which now shows a Retry instead of a misleading redirect.

## Model Used

## Checklist

- [x] I have searched the GitHub PR list for similar/duplicate PRs (companion to #7819; no existing PR covers the /auth redirect path).

### DIFF
--- a/ui/src/App.test.tsx
+++ b/ui/src/App.test.tsx
@@ -202,6 +202,53 @@ describe("CloudAccessGate", () => {
     unmountRoot(root);
   });
 
+  it("redirects to /auth when the server confirms the user is signed out", async () => {
+    mockAuthApi.getSession.mockResolvedValue(null);
+
+    const root = renderGate(container);
+    await waitForText(container, "Navigate:/auth");
+
+    expect(container.textContent).toContain("Navigate:/auth?next=");
+    expect(container.textContent).not.toContain("Outlet content");
+
+    unmountRoot(root);
+  });
+
+  it("does not redirect to /auth when the session fetch fails transiently", async () => {
+    mockAuthApi.getSession.mockRejectedValueOnce(new Error("Failed to load session (503)"));
+    mockAuthApi.getSession.mockResolvedValue({
+      session: { id: "session-1", userId: "user-1" },
+      user: { id: "user-1", email: "user@example.com", name: "User", image: null },
+    });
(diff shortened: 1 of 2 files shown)


dosthcpp:
Re-verified against master `696c694a` (2026-07-07): still unaddressed — `ui/src/components/CloudAccessGate.tsx:113` still redirects on `!sessionQuery.data` with no `isSuccess`/error guard. Will rebase on request. Triage summary of all our open PRs: #8180.

PR #7987, opened 2026-06-11
fix(CAR-127): Propagate request_board_approval grants to requesting agent

Stack-Alerts:
## Thinking Path

The Paperclip `request_board_approval` flow is a notification-only mechanism. When a board approves a request, no permission grants are actually wired into the `principal_permission_grants` table, so the requesting agent receives no new access. CAR-37 (discovery) and CAR-21 (blocked sweep) demonstrated the gap: SecurityAnalyst's `access.grants` remained `["tasks:assign"]` after a `request_board_approval` was approved for `secrets:read`, so guarded endpoints still returned 403.

Fix: parse the approval's `payload.grants` array on transition to `approved` and upsert grants for the requesting agent. The schema's unique index on `(company_id, principal_type, principal_id, permission_key)` makes the upsert path safe under re-approvals.

## What Changed

- `server/src/services/approvals.ts`
  - New private helper `applyRequestGrantsForApproval(approval, decidedByUserId)`
  - New `request_board_approval` branch in `approve()` that calls the helper when `applied === true`
  - Grants are upserted with `onConflictDoUpdate` on the unique index
- `packages/shared/src/validators/approval.ts`
  - New `approvalGrantInputSchema` and `requestBoardApprovalPayloadSchema` (additive)
  - `createApprovalSchema` is unchanged — payload remains `z.record(z.string(), z.unknown())` for backward compatibility
- `server/src/__tests__/approvals-service.test.ts`
  - New `describe` block: "approvalService request_board_approval grant propagation"
  - Three tests: successful upsert of two grants, no-op on idempotent retry, no-op when payload has no grants

## Verification

Run the new tests:

```sh
npx vitest run server/src/__tests__/approvals-service.test.ts
```

Expected: 6/6 pass (3 pre-existing + 3 new).

Existing related suite still passes:

```sh
npx vitest run server/src/__tests__/approval-routes-idempotency.test.ts
```

Expected: 8/8 pass.

Typecheck on the affected files (the server tsc has unrelated plugin-SDK pre-existing errors, but `approvals.ts` and `validators/approval.ts` produce no diagnostics):

```sh
npx tsc --noEmit -p server/tsconfig.json 2>&1 | grep -E "approvals|validators/approval"
# (no output = clean)
```

## Risks

- `expiresInHours` is captured in the payload schema but not yet enforced in the service. Revocation on expiry is a follow-up (CAR-127 follow-up #1).
- Permission keys are not validated against `PERMISSION_KEYS` at insert time. The grant will be persisted regardless. Auth middleware must be the source of truth on key validity. (Same posture as `setMemberPermissions` in `access.ts`.)
- The 168h cap on `expiresInHours` in the schema is a soft guardrail, not a hard server-side enforcement.

## Model Used

Claude Opus 4.7 (claude-opus-4-7), 200K context window. The change is a focused service-layer fix with one new test fixture.

## Checklist

- [x] Behavior matches existing access service grant pattern
- [x] Targeted test suite passes
- [x] Backward compatible (new exports are additive, no breaking change to `createApprovalSchema`)
- [x] PR description follows the project PR template
- [x] Branch pushed to Stack-Alerts/paperclip fork (PR target is paperclipai/paperclip)
- [x] I searched for similar open/closed PRs and confirmed this is not a duplicate


Stack-Alerts:
Re: Greptile review — Issue 1 (partial grant failure) is fixed in 604d1bf3. Removed the `applied &&` guard so the helper now re-runs on every approve() call for the `request_board_approval` type. The helper is already idempotent (`onConflictDoUpdate`), so a partial write recovers on the next call. Test rewritten to assert re-insertion on retry. 6/6 tests pass.

Issues 2/4/5 reference pre-existing files (`.github/workflows/paperclip-recovery-monitor.yml`, `scripts/token_gap_escalation.py`, `scripts/paperclip_recovery_monitor.py`) that are not in this PR's diff — they belong to other branches in this fork's lineage. Out of scope for CAR-127.

Issue 3 (use `requestBoardApprovalPayloadSchema` in the service instead of manual validation) is valid DRY feedback. The shared schema parses to a typed result; using `safeParse` here would replace the manual `Array.isArray` / `typeof` checks. Tracked as a follow-up edit to keep this PR scoped to the correctness fix.

commitperclip: the dedup-search checkbox from `PULL_REQUEST_TEMPLATE.md` is noted; the PR description has been updated to confirm I searched the open PR list and found no duplicate. Re-push should re-run the bot checks.

Stack-Alerts:
Addressed Greptile Issue 3 (DRY): pushed ec850507 replacing manual grant validation in applyRequestGrantsForApproval with requestBoardApprovalPayloadSchema.safeParse. 6/6 approvals-service tests pass.

PR #8016, opened 2026-06-12
Harden CLI: prevent agent runs from falling back to board token on writes

chrislro:
## Summary\n\nPrevents agent runs from authenticating as board users when they present a board API token. Agent runs (identified by the `x-paperclip-run-id` header) must now only authenticate via agent JWT or agent API key.\n\n## Changes\n\n- **src/middleware/auth.ts**: Skip board API key authentication when `x-paperclip-run-id` header is present\n- **src/__tests__/auth-agent-jwt-attribution.test.ts**: Add test verifying board tokens are rejected for agent runs\n\n## Rationale\n\nWhen an agent run's JWT was missing or invalid, the auth middleware would fall through to board API key authentication. If the agent had access to a board token (e.g., from CLI auth or environment variables), it would be authenticated as a board actor with broader permissions than intended. This change ensures agent runs always operate within their agent authorization boundary.\n\n## Testing\n\n- All auth attribution tests pass (3/3)\n- CLI auth route tests pass (10/10)\n- Plugin routes authz tests pass (35/35)\n\nCloses CHRA-2469


devinfoley:
Thanks for surfacing this. The run-scoped board-token fallback is still a real auth-boundary bug, but this branch has drifted into a very large mixed-scope change set that is no longer a good vehicle for the fix. Closing in favor of #8020 so we can land a focused patch against the current auth middleware path.

Issue #7996, opened 2026-06-11
RFC: shared artifact store for runtime plugin install on multi-replica deployments


stubbi:
## Summary

RFC for making runtime plugin install work on horizontally scaled and ephemeral deployments — the "shared artifact store, install coordination, or cross-node distribution layer" that `doc/plugins/PLUGIN_SPEC.md` explicitly names as not-yet-existing future work. Filing for design feedback before a PR (step 4 of #7992, after #5875/#7993/#7995).

## Problem

Dynamic plugin install writes to the local filesystem (`~/.paperclip/plugins`, npm-managed tree). With `replicas > 1`:

- A plugin installed via the API lands on ONE replica's disk; the `plugins` DB table says it's `ready`, but every other replica fails to load it.
- Ephemeral pods (no persistent volume) lose all dynamically installed plugins on restart.
- Concurrent installs from two replicas can corrupt the shared npm tree if a volume IS shared (RWX semantics).

## Proposed design: versioned full-tree snapshots in object storage

After surveying how other self-hosted platforms handle this (Directus `EXTENSIONS_LOCATION` S3 loading — adopted as the fix for their cluster install-propagation bug directus/directus#25270; n8n's DB-manifest + `N8N_REINSTALL_MISSING_PACKAGES` re-install-from-registry; Grafana/Discourse/Strapi rejecting runtime install when scaled; WordPress shared-EFS), the most robust pattern for Paperclip's npm-based plugin tree:

1. **Publisher** (the replica handling an install/uninstall/upgrade API call, serialized by the advisory lock from #7993): run today's install flow unchanged, then tar the whole plugins tree, upload to the configured object storage (`plugins/snapshots/gen-N.tgz`), and insert a row in an append-only `plugin_artifact_generations` table (generation PK = CAS; content hash). Full-tree (not per-plugin) because npm hoists shared dependencies — per-plugin tarballs are not hermetic.
2. **Reconciler** (every replica): at boot (before the plugin loader starts), on plugin lifecycle live events (cross-replica via #5875), and on a slow safety tick — compare local generation marker to `max(generation)`; download, hash-verify, extract to a temp dir, atomic directory swap, then hot-load/unload via the existing `pluginLifecycleManager` diff against the `plugins` table.
[...]

PR #7995, opened 2026-06-11
feat: lease-based scheduler leader election with automatic failover

stubbi:
> **Stacked on #7993** — this branch contains #7993's commits (the advisory-lock coordination layer and migration 0099). Review the last 6 commits, or wait for #7993 to merge; this PR will be rebased onto master and marked ready then.

## Thinking Path

> - Paperclip orchestrates AI agents for zero-human companies
> - The heartbeat/routines scheduler (timers, run resumption, recovery passes) must run on exactly one instance; today that's a manual toggle — `HEARTBEAT_SCHEDULER_ENABLED=false` on all but one replica
> - The manual pin has no failover: if the scheduler replica dies or is being rolled, all scheduling stops until it returns; the same manual-toggle pattern keeps reappearing for other singleton work (e.g. #6235's `PRODUCTIVITY_REVIEW_ENABLED`), which suggests the platform needs a real primitive
> - This PR adds lease-table leader election (the pattern production Postgres job systems converged on — River, Oban): a single `scheduler_leader` row, all lease math on the database clock, fencing via the renewal predicate, graceful resign on shutdown
> - The benefit: every replica is a candidate, exactly one runs the scheduler, crash failover ≤ lease TTL (15s), rollout failover ≤ one retry interval (~5s), and the scheduler's startup recovery chain re-runs on every acquisition — exactly the recovery a failover needs

## Linked Issues or Issue Description

Refs #7992 (multi-replica coordination gaps — step 3 of the plan laid out there, after #5875 and #7993).

## What Changed

- **`packages/db`** — `scheduler_leader` single-row lease table (migration `0100`): `name` PK, `leader_id`, `hostname`, `elected_at`, `expires_at`. All timestamps written with `now()` — client clocks never participate.
- **`server/src/services/scheduler-leadership.ts`** (new) — the elect/renew/resign loop: candidates delete expired leases and `INSERT … ON CONFLICT DO NOTHING` (exactly one wins); the leader renews with `UPDATE … WHERE leader_id = me AND expires_at >= now()` — the predicate is the fencing check, so a paused/wedged process cannot re-extend an expired lease and demotes itself on the next pass; `stop()` resigns (DELETE predicated on own `leader_id` — a stale leader cannot steal a successor's lease). Passes are serialized on a jittered `setTimeout` chain; any DB error while leading demotes conservatively. `onAcquired` failures resign and stay candidate.
[...]

Issue #7992, opened 2026-06-11
Multi-replica deployments: coordination gaps when running more than one server instance


stubbi:
## Problem

Paperclip's server is increasingly deployed behind load balancers with `replicas > 1` (Kubernetes, ECS). The data layer handles this well (`FOR UPDATE` row locking throughout, conflict-safe seeds), but several orchestration-layer mechanisms assume a single process:

- **Live events are process-local** — WebSocket clients on replica B never see events originating on replica A (PR #5875 addresses this with a Postgres LISTEN/NOTIFY transport).
- **Database migrations race at boot** — `applyPendingMigrations` has no cross-process lock; N replicas booting after an upgrade run concurrent DDL (`packages/db/src/client.ts`).
- **Scheduled backups fire on every replica** — the backup interval is gated only by `databaseBackupEnabled`, and the in-flight guard is an in-memory boolean (`server/src/index.ts`).
- **Plugin job scheduler dispatches duplicates** — the tick loop runs on every instance; overlap prevention is an in-process Set plus a read-then-throw DB check (TOCTOU) (`server/src/services/plugin-job-scheduler.ts`).
- **Agent-start lock is in-memory** — `withAgentStartLock` serializes per-process only (`server/src/services/agent-start-lock.ts`).
- **Local-trusted board principal bootstrap races** — unguarded check-then-insert at startup (`server/src/index.ts`).
- **Inbound plugin webhooks aren't deduplicated** — provider/load-balancer retries dispatch `handleWebhook` repeatedly; `plugin_webhook_deliveries.external_id` exists but is unused.
- **Company skill-inventory refresh duplicates work** — single-flighted per process via an in-memory Map only.

`HEARTBEAT_SCHEDULER_ENABLED` exists as a manual toggle to keep the heartbeat scheduler on one instance, which suggests multi-instance deployments are an intended direction (also: ROADMAP "Cloud deployments").

## Proposed direction

Postgres-native coordination, no new infrastructure (consistent with the no-Redis stance and with #5875):

1. A small namespaced advisory-lock helper (transaction-scoped for short critical sections — pooler-safe; session-scoped on a dedicated connection for migrations/backups).
2. Atomic claims where locks are the wrong tool (plugin-job schedule slots via a `next_run_at` CAS; webhook dedup via a partial unique index on provider idempotency ids).

[...]

PR #8029, opened 2026-06-12
feat: replicate runtime plugin installs across replicas (generation snapshots)

stubbi:
> **Stacked on the scheduler-leadership PR (#7995)** — this branch contains #7993's and #7995's commits plus migration ordering (0099→0100→0101). Review the last 7 commits; will be rebased and marked ready as the stack merges.

## Thinking Path

> - Runtime plugin install writes an npm tree to one replica's local disk; with `replicas > 1` the `plugins` table says a plugin is `ready` while every other replica is missing its files, and ephemeral pods lose all dynamically installed plugins on restart — PLUGIN_SPEC.md called this out explicitly as missing "shared artifact store, install coordination, and cross-node distribution"
> - This PR implements exactly those three pieces with no new infrastructure (RFC #7996): full-tree generation snapshots in the existing object storage, single-writer mutations via a session advisory lock, and per-replica reconcilers with hash-verified atomic swaps and hot reload
> - Full-tree (not per-plugin) snapshots because npm hoists shared dependencies into one `node_modules` — per-plugin tarballs are not hermetic; the survey of how Directus/n8n/Grafana handle this is in the RFC
> - The benefit: plugin install/uninstall/upgrade works on horizontally scaled and ephemeral deployments, replicas converge within an event-debounce (~2s) to one tick (60s), and a replica that cannot converge keeps serving its previous plugin set loudly rather than diverging silently

Refs #7992 (multi-replica umbrella), #7996 (this design's RFC — feedback welcome; the implementation matches the RFC's proposal with two refinements called out below).

## What Changed

- **`packages/db`** — `plugin_artifact_generations` append-only ledger (migration `0101`): `generation` bigint PK is the CAS — publishers insert `max+1` with a bounded PK-conflict retry, so exactly one writer wins each generation even if locking fails open.
- **`server/src/services/plugin-artifact-replication.ts`** (new) — publisher (tar the plugins tree → sha256 → `putObject` *before* the ledger insert → marker write → GC keep-last-3) and reconciler (download → hash-verify (mismatch aborts, local tree untouched) → extract to a sibling temp dir → atomic two-rename swap → marker → hot-reload hook). Reconciles are serialized and coalesced; triggered at startup (before plugin load), on `plugin.ui.updated` events (2s debounce), and a 60s safety tick; downloads bounded by a 60s timeout. Disabled mode (`provider: null`) no-ops everything — single-replica behavior byte-identical.
- **`server/src/routes/plugins.ts`** — install/uninstall/upgrade run under a **session** advisory lock (`plugin-install`; 409 on contention — these mutations include a 120s `npm install`, far too long for a transaction-scoped lock) and inside the reconciler's serialization domain: **converge-to-max → mutate → publish**, so a lagging replica can never publish a snapshot that drops a peer's newer generation, and no swap can interleave with a local mutation. Publish failure fails the request loudly (500) — a local mutation that isn't replicated is worse silent. Local-path installs are rejected (400) while replication is active (npm packages are the replicable artifact, per PLUGIN_SPEC).
- **`server/src/services/plugin-loader.ts`** — `reconcileLoadedPlugins()`: diffs registry-`ready` rows against live worker handles AND active versions — peers' upgrades restart the local worker (stale code from swapped-away inodes would otherwise keep running).
- **Readiness** — `GET /api/health/ready`: 503 until first sync when `PAPERCLIP_PLUGINS_MUST_SYNC=true` (readiness only; `/api/health` itself is unchanged so liveness probes can't crash-loop a syncing pod).
- **Activation** — snapshot storage configured (`storageProvider != local_disk`) or `PAPERCLIP_PLUGIN_SNAPSHOTS=true`.
- **Docs** — `doc/plugin-replication.md` (mechanics, env table, propagation bounds, failure modes); PLUGIN_SPEC.md caveats updated to describe what now exists.

Refinements vs the RFC text: (1) the cross-replica reconcile trigger rides `plugin.ui.updated` — in-process today, cross-replica once #5875's transport lands; until then the 60s tick is the propagation bound (documented). (2) Single-writer is belt-and-braces: session lock for liveness-friendly mutual exclusion plus the generation-PK CAS for correctness if the lock ever fails open.

## Verification

- `server/src/__tests__/plugin-artifact-replication.test.ts` — real Postgres (embedded), real `tar`, real `local_disk` storage provider, two service instances: publish/CAS-concurrency ({1,2} under `Promise.all`), cross-replica byte-identical convergence, idempotence, tamper-abort (corrupted object leaves the tree untouched), GC, disabled-mode no-ops, converge-before-mutate (a stale replica's wrapped mutation publishes a snapshot containing BOTH its change and the peer's newer generation — fails without the fix), publisher-marker self-skip, coalescing, download timeout.
- Route tests: lock acquisition + single publish per mutation, 409 on contention, local-path rejection, 500-on-publish-failure with no success event, lock release on failure.
- Loader: version-diff restart (fail-first), unload-of-removed, continue-past-failing-restart.
- Health: `/ready` 503-until-synced; `/api/health` pinned unchanged while unsynced.
- `tsc --noEmit` clean; `@paperclipai/db` build + migration numbering clean; full server suite green modulo the two documented pre-existing local-env flakes; new embedded-PG suites registered in the serialized shard list.

## Risks

- At-most-once propagation latency: a replica can serve a stale plugin set for up to 60s (event-triggered ~2s once #5875 lands). `PAPERCLIP_PLUGINS_MUST_SYNC` turns that into a readiness gate for fleets that need strictness.
- Mutations now 409 when another plugin operation is in flight anywhere in the cluster (was: silent local race). Deliberate.
- The plugins tree is briefly absent (microseconds, two renames) during a swap; worker spawns in that window fail and retry via existing crash-recovery.
- Disabled deployments (default for local/single-node): zero behavior change.

## Model Used

Claude (Fable 5) with subagent-driven implementation and two-stage review per task.

## Related PRs

#7995 (stacked base — lease leadership), #7993 (coordination layer — supplies the advisory-lock helpers), #5875 (live-events transport — upgrades the propagation bound when it lands), #6753 (closed; unrelated artifact-fetch SDK surface).

## Checklist

- [x] I have checked ROADMAP.md and confirmed this PR does not duplicate planned core work (implements PLUGIN_SPEC's own named Phase-2 work, coordinated via #7992/#7996)
- [x] If this change affects the UI, I have included before/after screenshots (no UI changes)


stubbi:
@greptileai review please — plugin artifact replication: generation-snapshot ledger (CAS), converge-before-mutate under a session advisory lock, hash-verified atomic swaps with hot reload (incl. worker restarts on peer upgrades), readiness-only MUST_SYNC gate. Stacked on #7995; review the last 7 commits.

stubbi:
@greptileai re-review please — all three findings addressed in c70749293: row-first GC ordering (reconcilers re-reading max skip pruned generations), lost-race CAS objects are now tracked and best-effort deleted after the winning publish, and the swap restores pluginsDir from oldDir when the second rename fails (original error preserved).

stubbi:
@greptileai re-review please — fresh-cluster deadlock fixed: an empty snapshot ledger now marks the replica synced (with a regression test), so PAPERCLIP_PLUGINS_MUST_SYNC can't 503 a brand-new cluster into a no-traffic/no-install standoff.

stubbi:
@greptileai re-review please — added a publish-heal: a retry after a failed snapshot publish is now idempotent (a matching-package 'already installed' conflict under active replication returns the existing row, and the mutation wrapper re-publishes the snapshot on the way out — the missing half of the failed attempt). Covered by two new route tests (heal succeeds + non-matching package still errors).

stubbi:

PR #8041, opened 2026-06-12
fix(ui): don't window-scroll the desktop shell on comment submit (#7972)

### LINKED ISSUE #7972: Submitting a comment scrolls the whole app shell above the viewport; sidebar nav becomes inaccessible until URL reload
### Pre-submission checklist

### What happened?

After submitting a comment on an issue in the board UI, the entire app shell (sidebar included) appears to shift upward so that the top of the canvas — and everything above "Tasks" in the left nav — sits above the visible viewport.

The page is not actually scrolled in a way the user can correct: pressing the browser's reload button does **not** restore the layout. Only hard-navigating to the issue URL (typing/pressing enter in the address bar, or opening a new tab) fixes it.

This appeared after the recent sidebar refactor. The shell now uses `flex h-dvh overflow-hidden` (plus `body { overflow: hidden }`), so any non-zero `scrollY` on `window` visually translates the whole shell up.

### Expected behavior

Submitting a comment should keep the composer visible and the sidebar/header in their normal positions. The window scroll position should not change at all.

### Steps to reproduce

1. Run a local Paperclip instance from `master` (built via `pnpm build` in `ui/`).
2. Open any issue with a short comment thread (one that does not yet overflow `#main-content`).
3. Focus the composer at the bottom, type a comment, and submit it.
4. Observe: the sidebar top and main header have scrolled above the viewport. Browser reload does not fix it. Only entering the URL in the address bar does.

Built from source on current `master` HEAD of `paperclipai/paperclip`.

### Deployment mode

Local dev (pnpm dev)

### Installation method

### Agent adapter(s) involved

- [x] Not adapter-specific (core bug) — pure UI

### Database mode

Not database-related

### Access context

Board (human operator)

### Node.js version

v25.6.1

### Operating system

macOS 15.7.4 (Apple Silicon)

### Relevant logs or output

No console errors. Visual-only regression in the UI shell.

### Relevant config

N/A

### Additional context

- Suspected mechanism: the post-submit composer-viewport restore in `ui/src/lib/issue-chat-scroll.ts` calls `window.scrollBy(...)` whenever it does not find a scrollable `#main-content`. Because the shell is `flex h-dvh overflow-hidden`, a non-zero window scroll visually translates the whole layout — including the sidebar — upward.
- An attempted fix made \`resolveIssueChatScrollTarget\` treat \`#main-content\`'s configured \`overflow-y\` as authoritative and always restore inside main even on short threads. Tests pass; the visual bug still reproduces on a freshly built \`ui/dist\`, so the root cause may be elsewhere (e.g. another scroll handler, or the composer-viewport snapshot capturing the wrong reference frame).
- Reproduces 100% on macOS 15.7.4 (Apple Silicon) Chrome with the desktop layout (not mobile). Hard-reload does not fix it; only entering the URL fresh does.

### Privacy checklist

dosthcpp:
Fixes #7972.

## Thinking Path
- The reporter pinned the post-submit composer-viewport restore in `ui/src/lib/issue-chat-scroll.ts`, which falls back to `window.scrollBy(...)` when `#main-content` is not independently scrollable — exactly the short-thread repro case.
- In the desktop shell the body is `overflow: hidden` (set in `Layout.tsx`) inside a fixed-height `h-dvh` flex column, so a window scroll never moves content: it translates the entire shell (sidebar included) off-screen, and a plain reload does not restore it. On mobile (`min-h-dvh`, `body { overflow: visible }`) and the auth-free perf fixture the page genuinely scrolls, so the window IS the correct target there.
- A prior attempt forced `resolveIssueChatScrollTarget` to always use `#main-content`; that path is a no-op on a non-overflowing container and is sensitive to a stale `ui/dist`/`.vite` cache, which likely masked the result. Gating the window-scroll itself is the precise root-cause fix and covers both restore call sites (`queueViewportRestore` and the `[messages]` layout effect) since both route through one function.

## What Changed
- Added `isWindowScrollable(doc, win)` to `issue-chat-scroll.ts`: the window is a valid scroll target only when the document body is not clipped. It checks both the `overflow` shorthand and the `overflow-y` longhand (some engines, incl. jsdom, do not derive the longhand from the shorthand in computed style).
- Gated the `window.scrollBy` fallback in `restoreComposerViewportSnapshot` behind `isWindowScrollable`; on the desktop shell there is nothing to restore, so the scroll position is left untouched.
- Added unit tests for the desktop-shell (no window scroll) case and for `isWindowScrollable`.

## Verification
- `ui $ vitest run src/lib/issue-chat-scroll.test.ts` → 6 passed (2 new + existing window/element restore tests still green).
- `ui $ vitest run src/components/IssueChatThread.test.tsx` → 55 passed (consumer regression).

## Risks
- Low. Behaviour only changes when the resolved target is `window` AND the document body is clipped — i.e. the desktop shell, where the previous behaviour was the bug. Mobile and the perf fixture keep window scrolling unchanged (body not clipped → `isWindowScrollable` true).

## Model Used
claude-opus-4-8

---

### DIFF
--- a/ui/src/lib/issue-chat-scroll.test.ts
+++ b/ui/src/lib/issue-chat-scroll.test.ts
@@ -3,6 +3,7 @@
 import { describe, expect, it, vi } from "vitest";
 import {
   captureComposerViewportSnapshot,
+  isWindowScrollable,
   restoreComposerViewportSnapshot,
   shouldPreserveComposerViewport,
 } from "./issue-chat-scroll";
@@ -37,6 +38,38 @@ describe("issue-chat-scroll", () => {
 
     scrollByMock.mockRestore();
     composer.remove();
+  });
+
+  it("does not scroll the window when the document body is overflow:hidden (desktop shell)", () => {
+    // The desktop app shell pins the body to overflow:hidden inside a fixed
+    // h-dvh flex column, so a window scroll would translate the whole shell
+    // (sidebar included) off-screen — paperclipai/paperclip#7972.
+    const composer = document.createElement("div");
+    document.body.appendChild(composer);
+    document.body.style.overflow = "hidden";
+    const scrollByMock = vi.spyOn(window, "scrollBy").mockImplementation(() => {});
+
+    mockTop(composer, 420);
+    const snapshot = captureComposerViewportSnapshot(composer);
+
+    mockTop(composer, 560);
+    restoreComposerViewportSnapshot(snapshot, composer);
+
+    expect(scrollByMock).not.toHaveBeenCalled();
+
+    scrollByMock.mockRestore();
+    document.body.style.overflow = "";
+    composer.remove();
+  });
+
+  it("reports the document as scrollable only when the body is not clipped", () => {
+    document.body.style.overflow = "";
+    expect(isWindowScrollable()).toBe(true);
+
+    document.body.style.overflow = "hidden";
+    expect(isWindowScrollable()).toBe(false);
+
+    document.body.style.overflow = "";
   });
 
   it("restores main-content scroll when the layout uses an internal scroller", () => {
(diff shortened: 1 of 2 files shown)


devinfoley:
Thank you!

PR #8042, opened 2026-06-12
fix(server): delete issue_comments before issue row in issues.remove() (#7991)

### LINKED ISSUE #7991: DELETE /api/issues/:id returns 500 when the issue has comments (FK issue_comments not cascaded)
## Summary

`DELETE /api/issues/:id` fails with `500 Internal server error` for any issue that has at least one comment. The delete transaction removes attachments and documents but never removes the issue's comments, so Postgres rejects the row delete on the `issue_comments_issue_id_issues_id_fk` foreign key.

## Repro

1. Create an issue.
2. Add one comment to it (`POST /api/issues/:id/comments`).
3. `DELETE /api/issues/:id`.

**Result:** `500 { "error": "Internal server error" }`. Server log:

```
PostgresError: update or delete on table "issues" violates foreign key constraint
"issue_comments_issue_id_issues_id_fk" on table "issue_comments"
code: 23503 — Key (id)=(...) is still referenced from table "issue_comments".
```

## Root cause

- `remove()` in `server/src/services/issues.ts:2846` deletes `issueAttachments`/`assets` and `issueDocuments`/`documents` inside the transaction, but does **not** delete `issueComments` before `delete(issues)`.
- `issueComments.issueId` in `packages/db/src/schema/issue_comments.ts:12` is `.references(() => issues.id)` with **no `onDelete` rule** (contrast `createdByRunId` on line 15, which uses `onDelete: "set null"`).

## Impact

Any issue that ever received a comment is undeletable via the API. This compounds badly with the heartbeat recovery loop: a stalled agent can generate tens of thousands of commented `Recover stalled issue …` tasks that then cannot be cleaned up (we hit 20k+ undeletable blocked issues, which kept crashing the server). Operators are forced to `cancel` instead of delete, leaving the rows behind.

## Suggested fix

Either:

- delete `issueComments` (and audit other issue-scoped child tables: thread interactions, reads/inbox, etc.) inside `remove()` before `delete(issues)`, mirroring the existing attachments/documents handling; or
- add `{ onDelete: "cascade" }` to the `issueId` FK in `issue_comments.ts` (plus a migration), and do the same for the other issue child tables.

dosthcpp:
Fixes #7991.

## Thinking Path
- `DELETE /api/issues/:id` runs inside a transaction in `issues.remove()`; it deletes `issueAttachments` asset rows and `issueDocuments` document rows before calling `tx.delete(issues)`, but never removes `issueComments` first.
- The `issue_comments` table has a non-cascaded FK `issue_comments_issue_id_issues_id_fk` on `issue_id`, so Postgres rejects the parent row deletion with a 23503 constraint violation any time at least one comment exists.
- The minimal fix is a single `tx.delete(issueComments).where(eq(issueComments.issueId, id))` call before the `tx.delete(issues)` inside the existing transaction — no migration needed, no cascade change to the schema.

## What Changed
- `server/src/services/issues.ts` — added `await tx.delete(issueComments).where(eq(issueComments.issueId, id));` before the `tx.delete(issues)` call in `remove()`.
- `server/src/__tests__/issues-service.test.ts` — new `issueService.remove` embedded-Postgres suite with one test: creates an issue, inserts a comment, calls `svc.remove()`, and asserts both that the call succeeds (no constraint error) and that the comment row is gone.

## Verification
- `server $ vitest run -t "issueService.remove"` → 1 passed.
- Confirmed that without the fix the same test fails with `PostgresError: update or delete on table \issues\ violates foreign key constraint`.

## Risks
- Low. The deleted comments belong to the deleted issue and have no other references; cleaning them inside the same transaction keeps the operation atomic. Soft-delete strategies are not in use for comments.

## Model Used
claude-sonnet-4-6

---

### DIFF
--- a/server/src/__tests__/issues-service.test.ts
+++ b/server/src/__tests__/issues-service.test.ts
@@ -4734,3 +4734,73 @@ describeEmbeddedPostgres("accepted plan decomposition", () => {
     expect(record?.childIssues.every((child) => typeof child.title === "string")).toBe(true);
   });
 });
+
+describeEmbeddedPostgres("issueService.remove", () => {
+  let db!: ReturnType<typeof createDb>;
+  let svc!: ReturnType<typeof issueService>;
+  let tempDb: Awaited<ReturnType<typeof startEmbeddedPostgresTestDatabase>> | null = null;
+
+  beforeAll(async () => {
+    tempDb = await startEmbeddedPostgresTestDatabase("paperclip-issues-remove-");
+    db = createDb(tempDb.connectionString);
+    svc = issueService(db);
+    await ensureIssueRelationsTable(db);
+  }, 20_000);
+
+  afterEach(async () => {
+    await db.delete(issueComments);
+    await db.delete(issueRelations);
+    await db.delete(issueDocuments);
+    await db.delete(issueInboxArchives);
+    await db.delete(activityLog);
+    await db.delete(issues);
+    await db.delete(documents);
+    await db.delete(executionWorkspaces);
+    await db.delete(projectWorkspaces);
+    await db.delete(projects);
+    await db.delete(goals);
+    await db.delete(heartbeatRuns);
+    await db.delete(agents);
+    await db.delete(instanceSettings);
+    await db.delete(companies);
+  });
+
+  afterAll(async () => {
+    await tempDb?.cleanup();
+  });
+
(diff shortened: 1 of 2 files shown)


dosthcpp:
Re-verified against master `696c694a` (2026-07-07): still unaddressed — `issues.ts` remove() still deletes the issue row without deleting `issue_comments` first. Will rebase on request. Triage summary of all our open PRs: #8180.

PR #8043, opened 2026-06-12
perf(server): cache Intl.DateTimeFormat per timezone in routine scheduler (#8033)

### LINKED ISSUE #8033: Routine scheduler blocks the event loop: new Intl.DateTimeFormat constructed per minute-step in computeNextRun (up to 2.6M times per call)
### What happened

The routine scheduler can peg the server's event loop at ~100% CPU for minutes (or effectively forever), making the whole server unresponsive — health checks time out, the UI goes dark, and on Paperclip Desktop the "embedded server is no longer responding" failure dialog fires repeatedly.

Profiling a hung production server (macOS, `sample` on the server process) showed **74% of all CPU samples inside `v8::internal::Builtin_DateTimeFormatConstructor`** — i.e. `new Intl.DateTimeFormat(...)` construction, called from the cron scheduling code:

- `computeNextRun` / `nextCronTickInTimeZone` advances **one minute at a time** with a cap of `366 * 24 * 60 * 5` = **2,632,320 iterations** (`server/src/services/routines.ts:154`)
- every iteration calls `matchesCronMinute` → `getZonedMinuteParts`, which constructs a **fresh `Intl.DateTimeFormat`** (`server/src/services/routines.ts:106`). Constructing one is ~1ms of ICU work (locale data load + pattern resolution), vs ~1µs for `formatToParts` on an existing instance.

So a single `nextCronTickInTimeZone` call for a sparse schedule (e.g. monthly = ~40k minute-steps) costs ~40s of blocked event loop. A cron that never matches — which is exactly what #7529 produces for midnight UTC crons on h24 ICU builds — runs the loop to its full 2.6M cap: **~45 minutes of pegged CPU per scheduler tick**, and the tick fires every 30s, so the server never recovers. The catch-up loop (`enqueue_missed_with_cap`) multiplies this further after the server has been offline.

### Expected behavior

Scheduler ticks compute next cron occurrences in microseconds-to-milliseconds without blocking the event loop; the server stays responsive while routines are active.

### Steps to reproduce

1. Create a routine with a schedule trigger whose next occurrence is far away or never matches (e.g. a midnight UTC cron on an ICU build affected by #7529, or any sparse monthly cron after the server was offline for a while).
2. Start the server and watch CPU: each 30s scheduler tick drives the node process to 100% CPU; `/api/health` times out.

### Version / deployment

- `@paperclipai/server` 2026.609.0 (bundled in Paperclip Desktop 3.2.9), reproduced against current `master` source
- Deployment: local embedded (desktop), but applies to all modes

### Fix

Two independent mitigations; I'm opening a PR for the first:

1. **Cache the `Intl.DateTimeFormat` per timezone** in `getZonedMinuteParts` (and `assertTimeZone`). Formatter instances are immutable and reusable; this removes ~99% of the per-iteration cost (~1ms → ~1µs).
2. Longer term: replace minute-stepping with cron-field arithmetic (jump directly to the next candidate minute), so the worst case is bounded regardless of schedule sparsity.

Related: #7529 (never-matching midnight crons force the full 2.6M-iteration scan — the two bugs compound), PR #7922.

dosthcpp:
Fixes #8033.

## Thinking Path
- Profiling showed 74% of CPU samples in `v8::internal::Builtin_DateTimeFormatConstructor` during a scheduler hang. The call site is `getZonedMinuteParts`, which constructs a new `Intl.DateTimeFormat` on every minute-step iteration inside `nextCronTickInTimeZone`.
- `Intl.DateTimeFormat` construction involves ICU locale data load + pattern resolution, costing ~1ms each. With up to 2,632,320 iterations for a sparse schedule, a single scheduler tick can block the event loop for ~45 minutes. Formatter instances themselves are immutable and safe to reuse across calls.
- Caching by timezone string (a module-level `Map`) reduces the per-iteration cost from ~1ms to ~1µs (`formatToParts` on an existing instance), making the loop cost negligible regardless of schedule sparsity. The cache is module-scoped and never invalidated — timezone definitions don't change at runtime.

## What Changed
- `server/src/services/routines.ts` — added `ZONED_MINUTE_FORMATTER_CACHE: Map<string, Intl.DateTimeFormat>` and `getOrCreateZonedMinuteFormatter(timeZone)` helper. Replaced the inline `new Intl.DateTimeFormat(...)` in `getZonedMinuteParts` and `assertTimeZone` with calls to this helper. Also exported `nextCronTickInTimeZone` for testability.
- `server/src/__tests__/routines-cron-tz.test.ts` — new unit-test file with four tests: next UTC hourly tick, next daily tick in America/New_York (DST aware), formatter-cache determinism (same result on repeated calls), and invalid-cron rejection.

## Verification
- `server $ vitest run routines-cron-tz` → 4 passed.
- Manual reasoning: formatter options are identical across all calls for a given timezone (same locale, same fields), so caching is correct. A formatter instance holds no per-date state; only `formatToParts(date)` is date-sensitive.

## Risks
- Low. The cache holds at most one entry per distinct timezone string seen at runtime (IANA tz database has ~500 zones; most deployments use 1–5). Memory impact is negligible. The only behavioral change is that the first call for a timezone pays the construction cost; all subsequent calls are free.

## Model Used
claude-sonnet-4-6

---

### DIFF
--- a/server/src/__tests__/routines-cron-tz.test.ts
+++ b/server/src/__tests__/routines-cron-tz.test.ts
@@ -1,0 +1,34 @@
+import { describe, expect, it } from "vitest";
+import { nextCronTickInTimeZone } from "../services/routines.ts";
+
+describe("nextCronTickInTimeZone", () => {
+  it("finds the next hourly tick in UTC", () => {
+    // 2026-01-01 00:30 UTC — next tick for "0 * * * *" is 01:00 UTC
+    const after = new Date("2026-01-01T00:30:00.000Z");
+    const result = nextCronTickInTimeZone("0 * * * *", "UTC", after);
+    expect(result).not.toBeNull();
+    expect(result?.toISOString()).toBe("2026-01-01T01:00:00.000Z");
+  });
+
+  it("finds the next daily tick in a non-UTC timezone (America/New_York)", () => {
+    // 2026-06-01 14:00 UTC = 10:00 EDT; next midnight NY = 04:00 UTC the next day
+    const after = new Date("2026-06-01T14:00:00.000Z");
+    const result = nextCronTickInTimeZone("0 0 * * *", "America/New_York", after);
+    expect(result).not.toBeNull();
+    // midnight NY on 2026-06-02 = 04:00 UTC (EDT = UTC-4)
+    expect(result?.toISOString()).toBe("2026-06-02T04:00:00.000Z");
+  });
+
+  it("returns the same result when called repeatedly for the same timezone (formatter cache)", () => {
+    const after = new Date("2026-03-15T08:00:00.000Z");
+    const first = nextCronTickInTimeZone("30 9 * * *", "Asia/Tokyo", after);
+    const second = nextCronTickInTimeZone("30 9 * * *", "Asia/Tokyo", after);
+    // Determinism proves the cached formatter produces the same result as a fresh one.
+    expect(first?.toISOString()).toBe(second?.toISOString());
+  });
+
+  it("throws for an invalid cron expression", () => {
+    const after = new Date("2026-01-01T00:00:00.000Z");
+    expect(() => nextCronTickInTimeZone("99 * * * *", "UTC", after)).toThrow();
+  });
+});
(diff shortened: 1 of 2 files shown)


dosthcpp:
Closing — master now has an equivalent per-timezone formatter cache, added independently after this PR was opened (`server/src/services/routines.ts`: `zonedMinuteFormatterCache`, see #8033 note in the comment there). No longer needed. Thanks!

PR #7354, opened 2026-06-01
fix(opencode-local): preserve opencode.jsonc fields in runtime config

sanidhyasin:
Closes #7040

**Related PRs:** #8075 (opened 2026-06-13) overlaps with this one — it also parses the OpenCode runtime config as JSONC so user fields survive. This PR (#7354) predates it (opened 2026-06-01) and additionally preserves `opencode.json` → `opencode.jsonc` precedence and handles trailing commas. No other open/merged PR covers #7040.

## Thinking Path

> - Paperclip orchestrates AI agents for zero-human companies, spawning agent processes through pluggable adapters.
> - The `opencode-local` adapter prepares a per-run OpenCode runtime config so headless agents don't hit interactive approval prompts.
> - To do this, `prepareOpenCodeRuntimeConfig()` copies the user's OpenCode config into a temp dir and injects `permission.external_directory = "allow"`.
> - The problem: it only ever *read* and *wrote* `opencode.json`, but the OpenCode CLI's default config format is `opencode.jsonc`. The user's `.jsonc` was copied in, then a permission-only `opencode.json` was written next to it.
> - Because `opencode.json` takes precedence over `opencode.jsonc`, the user's real config (`plugin`, `mcp`, …) was silently shadowed on every agent spawn (issue #7040).
> - This pull request reads/merges/writes the config the user actually has, instead of unconditionally producing a permission-only `opencode.json`.
> - The benefit is that OpenCode plugins, MCP servers, and other settings survive into the runtime config, while the existing `opencode.json` behavior is unchanged.

## What Changed

- `prepareOpenCodeRuntimeConfig()` now locates the existing config via `readRuntimeConfig()`, which prefers `opencode.json` and falls back to `opencode.jsonc`, and merges the `external_directory` permission back into **that same file** (no longer always writing `opencode.json`).
- Added a small string-aware `stripJsonComments()` so commented `.jsonc` files parse with the built-in JSON parser. Comment-like sequences inside string values are preserved; trailing commas are stripped via a companion `stripTrailingCommas()` pass (composed in `parseJsonc()`).
- Added tests covering the `.jsonc` preservation path and `opencode.json` precedence when both files exist.
- No new dependencies.

## Verification

- `cd packages/adapters/opencode-local && npx vitest run src/server/runtime-config.test.ts` → 4 tests pass (2 pre-existing, 2 new).
- `cd packages/adapters/opencode-local && npx tsc --noEmit` → passes.
[...]

Issue #7040, opened 2026-05-30
opencode-local adapter drops plugin and other fields from opencode.jsonc when creating runtime config


OBODA0:
prepareOpenCodeRuntimeConfig() in runtime-config.js only reads opencode.json when merging existing config into the runtime opencode.json.
    
If the user's config lives in opencode.jsonc (which is the default format OpenCode writes), fields like plugin, mcp, etc. are silently dropped every time Paperclip spawns an agent. The temp runtime config ends up with only the permission field.
    
Root cause: runtime-config.js line 62 — readJsonObject(runtimeConfigPath) looks for opencode.json only. No fallback to opencode.jsonc.

PR #8075, opened 2026-06-13
fix(opencode_local): preserve custom providers by parsing runtime config as JSONC

mbulava:
## Thinking Path

> - Paperclip is the open source app people use to manage AI agents for work; agents run through pluggable runtime adapters.
> - The `opencode_local` adapter runs OpenCode as the agent runtime and, for unattended runs, injects `permission.external_directory: "allow"` into a per-run copy of the user's `opencode.json`.
> - `prepareOpenCodeRuntimeConfig` does this by copying `$XDG_CONFIG_HOME/opencode` into a temp dir, re-reading `opencode.json`, merging the permission, and rewriting the file.
> - The re-read used strict `JSON.parse`, but OpenCode config is **JSONC** (comments + trailing commas), so any JSONC config threw, was swallowed to `{}`, and the rewrite overwrote the copied config with a permission-only stub — silently deleting every custom provider.
> - The agent's OpenCode process then only saw built-in providers, surfacing `Configured OpenCode model is unavailable: ...; Available models: github-copilot/...`. Manual host runs worked because they read `~/.config/opencode` directly with no strict re-parse/overwrite.
> - This PR makes the re-read JSONC-tolerant (via `JSON5`, a strict superset of JSONC) so a valid OpenCode config round-trips and custom providers survive the permission merge, and fails open if a config is truly unparseable.
> - The benefit is that user-defined OpenCode providers are never silently dropped by the adapter, preventing a whole class of "model unavailable" failures.

Tracked internally as Paperclip issue **DEV-643** (no public GitHub issue). Described in-PR per the bug template:

- **What happened:** With a JSONC `~/.config/opencode/opencode.json` that defines a custom `provider` (e.g. a local/gateway OpenAI-compatible endpoint), the `opencode_local` adapter generated a per-run `opencode.json` containing only `permission.external_directory: "allow"` — the custom providers were gone. The OpenCode run then failed with `Configured OpenCode model is unavailable`.
- **Root cause:** `readJsonObject()` in `packages/adapters/opencode-local/src/server/runtime-config.ts` parsed the copied config with strict `JSON.parse`. On any JSONC (trailing comma/comment) it threw, returned `{}`, and the subsequent merge+rewrite overwrote the copied config with a permission-only stub.
- **Expected:** A valid OpenCode (JSONC) config round-trips; custom providers are preserved and the permission rule is still injected.

## What Changed

- `runtime-config.ts`: parse the copied runtime `opencode.json` with `JSON5` (strict superset of JSONC) instead of strict `JSON.parse`, so valid OpenCode configs round-trip. `JSON5` never corrupts in-string values such as `"baseURL": "http://host:11434/v1"` (the documented footgun a naive `//` comment stripper would hit).
- `runtime-config.ts`: defensive fail-open — if the copied config cannot be parsed even as JSON5, leave the copied file intact and skip permission injection for that run (preferring "providers preserved, permission maybe not injected" over "providers deleted").
- `package.json`: add `json5@^2.2.3` as a direct dependency of `@paperclipai/adapter-opencode-local` (already resolved in the lockfile; ships its own types).
- `runtime-config.test.ts`: add a raw-JSONC test (line/inline comments + trailing commas + a custom provider whose `options.baseURL` is an `http://` URL) asserting the provider and URL survive and `permission.external_directory: "allow"` is injected; add a fail-open test asserting an unparseable copied config is preserved verbatim. The previous tests only wrote configs via `JSON.stringify`, so they never exercised JSONC — that gap let this ship.

Note: `parseProviderConfig` (which parses the machine-generated `PAPERCLIP_OPENCODE_PROVIDERS` env var) intentionally keeps strict `JSON.parse` — that value is serialized by Paperclip via `JSON.stringify`, so strict parsing is correct and its existing "invalid JSON" diagnostics are preserved.

## Verification

Run from repo root:

- `pnpm exec vitest run packages/adapters/opencode-local/src/server/runtime-config.test.ts` → **13 passed** (11 existing + 2 new).
- `pnpm --filter @paperclipai/adapter-opencode-local typecheck` → exit 0.
- `pnpm --filter @paperclipai/adapter-opencode-local build` → exit 0.
- `pnpm install --frozen-lockfile` → exit 0 (lockfile is consistent).
- `pnpm run check:tokens` → no forbidden tokens.

Manual repro of the parser behavior (proves the bug and the fix):

```
strict JSON.parse => THREW: Expected property name or '}' in JSON at position 4
JSON5 baseURL     => http://localhost:11434/v1
```

i.e. the raw JSONC config (trailing comma + comment) that strict `JSON.parse` rejects is parsed correctly by `JSON5`, and the `http://` provider URL is preserved intact.

## Risks

Low risk. The change is isolated to the `opencode_local` adapter:
- JSONC is a strict subset of JSON5, so every config that previously parsed (strict JSON) still parses identically; only previously-failing JSONC now parses instead of collapsing to `{}`.
- The new dependency (`json5@2.2.3`) was already present in the lockfile transitively; this only adds a direct importer edge.
- The fail-open path only triggers for configs that are unparseable even as JSON5 (which OpenCode itself would also reject); in that case it preserves the user's file instead of destroying it.

## Model Used

Claude (Anthropic) — `claude-opus-4.8`, extended thinking enabled, with tool use / code execution. Authored via Paperclip's CTO agent.

## Checklist

### DIFF
--- a/packages/adapters/opencode-local/package.json
+++ b/packages/adapters/opencode-local/package.json
@@ -52,6 +52,7 @@
   },
   "dependencies": {
     "@paperclipai/adapter-utils": "workspace:*",
+    "json5": "^2.2.3",
     "picocolors": "^1.1.1"
   },
   "devDependencies": {
--- a/packages/adapters/opencode-local/src/server/runtime-config.test.ts
+++ b/packages/adapters/opencode-local/src/server/runtime-config.test.ts
@@ -30,6 +30,17 @@ async function makeConfigHome(initialConfig?: Record<string, unknown>) {
   return root;
 }
 
+// Writes opencode.json verbatim so tests can exercise *raw* JSONC bytes (comments,
+// trailing commas) that `JSON.stringify` would never produce.
+async function makeConfigHomeRaw(rawConfig: string) {
+  const root = await fs.mkdtemp(path.join(os.tmpdir(), "paperclip-opencode-test-"));
+  cleanupPaths.add(root);
+  const configDir = path.join(root, "opencode");
+  await fs.mkdir(configDir, { recursive: true });
+  await fs.writeFile(path.join(configDir, "opencode.json"), rawConfig, "utf8");
+  return root;
+}
+
 describe("prepareOpenCodeRuntimeConfig", () => {
   it("injects an external_directory allow rule by default", async () => {
     const configHome = await makeConfigHome({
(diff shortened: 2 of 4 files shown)


sanidhyasin:
Heads up @mbulava — my PR #7354 fixes the same #7040 issue and predates this by a couple weeks. Is there something yours does better? If not, the guidelines ("Search First") ask us to help the existing PR over the line instead of opening a parallel one.

PR #8088, opened 2026-06-13
feat(issues): server-side 60s time-window dedup on POST /issues

billymacdo:
## Thinking Path

> - Paperclip is the open source app people use to manage AI agents for work
> - Agents create child issues via `POST /api/companies/:companyId/issues` as part of their heartbeat work
> - A recurring bug (FUC-327) causes LLM tool-call regeneration to fire the same `POST /issues` call twice in adjacent turns, producing duplicate child issues under the same parent
> - The fix must be server-side because client-side guards are unreliable — the LLM genuinely does not read its own prior tool results correctly
> - This PR adds a 60-second time-window deduplication check in the `POST /companies/:companyId/issues` route: if the same `parentId + title + createdByAgentId` tuple arrives again within 60 s, return the existing issue with HTTP 200 and `X-Paperclip-Deduplicated: true` instead of inserting a duplicate row
> - Top-level issues (no `parentId`) are excluded because repeated titles are legitimate for periodic work (e.g. "Daily CEO briefing"); cancelled originals are excluded so genuine re-creates still work
> - The benefit is that the duplicate-issue accumulation pattern stops at the platform layer without requiring any change to agent prompts or skill docs

## Linked Issues or Issue Description

Internal Paperclip task FUC-360, child of root-cause diagnosis FUC-327.

**Problem (feature request form):**
- **Motivation:** Multi-agent runs produce duplicate child issues when the LLM regenerates a tool call in an adjacent heartbeat turn.
- **Proposed solution:** Server-side time-window dedup keyed on `(parentId, title, createdByAgentId)` with a 60-second window, returning the original issue body on a hit.
- **Alternatives considered:** Client-side idempotency key (Layer 2, blocked on this ticket); prompt-level guards (unreliable).
- **Roadmap alignment:** Bug-fix / platform hardening; orthogonal to roadmap features.

## What Changed

[...]

PR #8097, opened 2026-06-13
feat(issues): Layer 2 — explicit idempotencyKey on POST /issues (FUC-361)

billymacdo:
## Thinking Path

> - Layer 1 (PR #8088) added a transparent 60-second time-window dedup keyed on `parentId + title + createdByAgentId`
> - Layer 1 is the safety net for LLM tool-call regeneration; Layer 2 makes dedup explicit and auditable for agents that want deterministic idempotency across the full 24-hour window
> - Agents create the same child issue (Bill action, blocker, delegation) in multiple heartbeats separated by more than 60 s — only a named key with a longer TTL prevents those duplicates
> - The idempotency pattern already exists for issue-thread interactions (migration 0064); the same nullable-column + application-level-lookup approach is used here so key reuse after expiry works without DB cleanup
> - No unique DB constraint: the 24-hour window is enforced by the application query (`createdAt > now-24h AND status != 'cancelled'`), allowing the same key to be reused after expiry or cancellation without having to null out the old row

## Linked Issues

Stacks on Layer 1 — PR #8088 (`fuc-360-issue-dedup-60s`) must be merged before this one. Internal task: FUC-361 (child of FUC-327 root-cause diagnosis).

## What Changed

- **`packages/db/src/migrations/0102_issue_idempotency_key.sql`** — New migration: adds nullable `idempotency_key` text column to `issues` table; creates a partial btree index on `(company_id, idempotency_key) WHERE idempotency_key IS NOT NULL` for fast lookup.
- **`packages/db/src/schema/issues.ts`** — Added `idempotencyKey: text("idempotency_key")` column and `idempotencyKeyIdx` index to the Drizzle schema.
- **`packages/shared/src/validators/issue.ts`** — Added `idempotencyKey: z.string().trim().max(255).optional().nullable()` to `createIssueBaseSchema`.
- **`server/src/routes/issues.ts`** — Added Layer 2 dedup block after the Layer 1 block, before `svc.create()`: queries for a matching non-cancelled issue with the same key created within 24 h in the same company; on hit, returns HTTP 200 with original body + `X-Paperclip-Deduplicated: true`.
- **`server/src/__tests__/issue-create-idempotency-key-routes.test.ts`** — New embedded-postgres integration tests covering happy path, key stored on row, cancel-reset, cross-company isolation, and no-dedup when key absent.

## Scope vs FUC-361 acceptance criteria

| Criterion | Status |
|---|---|
| `POST /issues` accepts optional `idempotencyKey`; duplicate returns 200 + original + header | ✅ |
| Key TTL 24 h | ✅ |
| Cancelled issues release the key | ✅ |
| Keys are company-scoped | ✅ |
| Layer 1 dedup coexists | ✅ |
| Paperclip skill doc updated | ✅ (local `api-reference.md`) |
| Tests: happy path, cancel-reset, cross-company isolation | ✅ |

## Out of scope

- AGENTS.md changes — skill doc is sufficient per ticket spec.
- Removing Layer 1 — it remains as the transparent backstop.

## Risks

- No unique DB constraint: the application query enforces the 24-hour TTL. Race conditions between two concurrent POSTs with the same key within milliseconds could both pass the lookup and create two rows. This is the same trade-off as Layer 1. The 60-second Layer 1 guard covers the highest-risk window; Layer 2 covers the longer TTL case.
- No migration required for Layer 1 data: the column is nullable and the index is partial.

## Model Used

- Provider: Anthropic  
- Context window: 200 k tokens

## Checklist

- [x] Thinking path included
- [x] Stacks on Layer 1 PR #8088 (must merge first)
- [x] Tests: happy path, cancel-reset, cross-company isolation
- [x] Skill doc updated (`api-reference.md`)
- [ ] Greptile review complete

### DIFF
--- a/packages/db/src/migrations/0102_issue_idempotency_key.sql
+++ b/packages/db/src/migrations/0102_issue_idempotency_key.sql
@@ -1,0 +1,4 @@
+ALTER TABLE "issues" ADD COLUMN IF NOT EXISTS "idempotency_key" text;--> statement-breakpoint
+CREATE INDEX IF NOT EXISTS "issues_company_idempotency_key_idx"
+  ON "issues" USING btree ("company_id","idempotency_key")
+  WHERE "issues"."idempotency_key" IS NOT NULL;
--- a/packages/db/src/schema/issues.ts
+++ b/packages/db/src/schema/issues.ts
@@ -49,6 +49,7 @@ export const issues = pgTable(
     originFingerprint: text("origin_fingerprint").notNull().default("default"),
     requestDepth: integer("request_depth").notNull().default(0),
     billingCode: text("billing_code"),
+    idempotencyKey: text("idempotency_key"),
     assigneeAdapterOverrides: jsonb("assignee_adapter_overrides").$type<Record<string, unknown>>(),
     executionPolicy: jsonb("execution_policy").$type<Record<string, unknown>>(),
     executionState: jsonb("execution_state").$type<Record<string, unknown>>(),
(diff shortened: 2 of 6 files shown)


billymacdo:
Closing: same reason as PR #8088. This PR targets the upstream OSS repo (`paperclipai/paperclip`) which Bill (`billymacdo`) cannot merge into. His Paperclip instance runs from the npm package (`paperclipai@2026.609.0` via npx), so upstream merges don't reach him anyway.

The Layer 2 `idempotencyKey` feature from this PR will be delivered via in-place npm dist patch once Bill ratifies the distribution path. CI issues on this PR are therefore moot. See FUC-436 for the distribution path decision.

PR #8110, opened 2026-06-14
fix(plugins): refresh UI contributions on hot lifecycle events

Polzer99:
## Thinking Path

> - Paperclip exposes plugin UI contributions in the board, and plugin lifecycle changes can happen while the board is open.
> - `PLUGIN_SPEC.md` requires updated plugin UI bundles to become visible after hot install, uninstall, upgrade, and config changes.
> - The server already published `plugin.ui.updated`, but the board did not reliably receive that global event and clear/refetch plugin UI state.
> - This pull request wires the global live event through board WebSocket clients and makes the frontend reset plugin module loader state before refetching plugin contributions.
> - Follow-up review tightened the path so in-flight imports cannot write stale modules after a reset, global events stay board-only, and tests cover the reset side effect.
> - The benefit is that plugin UI changes become visible after hot lifecycle changes without a manual page reload or server restart.

No public GitHub issue exists for this focused plugin hot-refresh gap, so the underlying issue is described inline using the feature-request template fields.

### Subsystem affected

Cross-cutting: `server/` realtime live-event fan-out and `ui/` plugin module loading, React Query cache invalidation, and board live updates.

### Problem or motivation

Plugin lifecycle mutations can update UI contributions while the board is already open. Before this PR, the server-side lifecycle path could publish `plugin.ui.updated`, but the board did not reliably receive that instance-global event or use it to clear cached ESM plugin modules and refetch plugin registry/UI contribution data. A changed plugin UI bundle could therefore remain stale until a manual page reload or server restart.

### Proposed solution

When a plugin UI lifecycle event occurs, authenticated board WebSocket clients receive the tenant-free global `plugin.ui.updated` event. The UI handles it before the company-id guard, resets plugin module loader state, and invalidates plugin registry/UI contribution queries so the board fetches and renders fresh plugin UI contributions.

### Alternatives considered

- Rely on manual page reloads after plugin lifecycle changes: rejected because hot plugin lifecycle behavior should not require operator reloads.
- Poll plugin registries more aggressively: rejected because it adds background work and still leaves stale ESM module state unless the loader is reset.
- Send the event to all WebSocket actor types: narrowed after review so only board clients subscribe to global plugin UI updates.

### Roadmap alignment

This is a small targeted improvement to the existing plugin system hot lifecycle behavior. It does not add a new core product surface and does not conflict with `ROADMAP.md`.

## What Changed

- Relay global live events to authenticated board WebSocket clients while keeping agent sockets on company-scoped events only.
- Restrict `publishGlobalLiveEvent` to the tenant-free `plugin.ui.updated` event type and document the cross-tenant `"*"` channel constraint.
- Handle `plugin.ui.updated` before the company-id guard in `LiveUpdatesProvider`, reset plugin module loader state, and invalidate plugin registry/UI contribution queries.
- Add a generation guard around plugin UI module imports so an old in-flight import cannot repopulate caches after a hot reset.
- Keep shared React/SDK shim blob URLs stable across hot resets instead of revoking invariant host shims.
- Remove the unused `_resetPluginModuleLoader` alias and update the current slot tests to use `resetPluginModuleLoader`.
- Add tests covering the reset side effect, global `companyId:"*"` ordering, and mismatched-company event dropping.

## Verification

- `pnpm --filter @paperclipai/ui exec vitest run src/context/LiveUpdatesProvider.test.ts src/plugins/slots.test.ts`
- GitHub PR workflow after rebase: typecheck, build, server/workspace tests, serialized suites, e2e, canary dry run.

## Risks

- Low risk: the change is scoped to plugin UI cache invalidation and live event fan-out.
- Global events are intentionally limited to `plugin.ui.updated`; future global event types must be explicitly added and must not carry company-scoped data.
- Board clients across companies will receive tenant-free plugin UI update notifications by design because plugins are instance-global.
- Agent WebSocket clients no longer receive global plugin UI events.
- This is a behavioral UI refresh path, not a visual layout change, so screenshots are not applicable.

## Model Used

- OpenAI Codex in the Codex desktop app, GPT-5 coding agent, with local shell/tool use and medium reasoning.

## Checklist

### DIFF
--- a/server/src/realtime/live-events-ws.ts
+++ b/server/src/realtime/live-events-ws.ts
@@ -5,7 +5,7 @@ import type { Duplex } from "node:stream";
 import { and, eq, isNull } from "drizzle-orm";
 import type { Db } from "@paperclipai/db";
 import { agentApiKeys, companyMemberships, instanceUserRoles } from "@paperclipai/db";
-import type { DeploymentMode } from "@paperclipai/shared";
+import type { DeploymentMode, LiveEvent } from "@paperclipai/shared";
 import type { BetterAuthSessionResult } from "../auth/better-auth.js";
 import { logger } from "../middleware/logger.js";
 import { subscribeCompanyLiveEvents, subscribeGlobalLiveEvents } from "../services/live-events.js";
@@ -205,13 +205,15 @@ export function setupLiveEventsWebSocketServer(
       return;
     }
 
-    const sendEvent = (event: unknown) => {
+    const sendEvent = (event: LiveEvent) => {
       if (socket.readyState !== WebSocket.OPEN) return;
       socket.send(JSON.stringify(event));
     };
 
     const unsubscribeCompany = subscribeCompanyLiveEvents(context.companyId, sendEvent);
-    const unsubscribeGlobal = subscribeGlobalLiveEvents(sendEvent);
+    const unsubscribeGlobal = context.actorType === "board"
+      ? subscribeGlobalLiveEvents(sendEvent)
+      : () => undefined;
     const unsubscribe = () => {
       unsubscribeCompany();
       unsubscribeGlobal();
--- a/server/src/realtime/live-events-ws.ts
+++ b/server/src/realtime/live-events-ws.ts
@@ -8,7 +8,7 @@ import { agentApiKeys, companyMemberships, instanceUserRoles } from "@paperclipa
 import type { DeploymentMode } from "@paperclipai/shared";
 import type { BetterAuthSessionResult } from "../auth/better-auth.js";
 import { logger } from "../middleware/logger.js";
-import { subscribeCompanyLiveEvents } from "../services/live-events.js";
+import { subscribeCompanyLiveEvents, subscribeGlobalLiveEvents } from "../services/live-events.js";
 
 interface WsSocket {
   readyState: number;
@@ -205,10 +205,17 @@ export function setupLiveEventsWebSocketServer(
       return;
     }
 
-    const unsubscribe = subscribeCompanyLiveEvents(context.companyId, (event) => {
+    const sendEvent = (event: unknown) => {
       if (socket.readyState !== WebSocket.OPEN) return;
       socket.send(JSON.stringify(event));
-    });
+    };
+
+    const unsubscribeCompany = subscribeCompanyLiveEvents(context.companyId, sendEvent);
+    const unsubscribeGlobal = subscribeGlobalLiveEvents(sendEvent);
+    const unsubscribe = () => {
+      unsubscribeCompany();
+      unsubscribeGlobal();
+    };
 
     cleanupByClient.set(socket, unsubscribe);
     aliveByClient.set(socket, true);
--- a/server/src/services/live-events.ts
+++ b/server/src/services/live-events.ts
@@ -1,6 +1,7 @@
 import { EventEmitter } from "node:events";
 import type { LiveEvent, LiveEventType } from "@paperclipai/shared";
 
+type GlobalLiveEventType = Extract<LiveEventType, "plugin.ui.updated">;
 type LiveEventPayload = Record<string, unknown>;
 type LiveEventListener = (event: LiveEvent) => void;
 
@@ -35,9 +36,11 @@ export function publishLiveEvent(input: {
 }
 
 export function publishGlobalLiveEvent(input: {
-  type: LiveEventType;
+  type: GlobalLiveEventType;
   payload?: LiveEventPayload;
 }) {
+  // The "*" channel is broadcast to board clients across tenant scopes. Keep
+  // this restricted to instance-global, tenant-free event payloads.
   const event = toLiveEvent({ companyId: "*", type: input.type, payload: input.payload });
   emitter.emit("*", event);
   return event;
--- a/ui/src/context/LiveUpdatesProvider.test.ts
+++ b/ui/src/context/LiveUpdatesProvider.test.ts
@@ -15,6 +15,25 @@ import { __liveUpdatesTestUtils } from "./LiveUpdatesProvider";
 import { queryKeys } from "../lib/queryKeys";
 
 describe("LiveUpdatesProvider issue invalidation", () => {
+  it("refreshes plugin registry queries when plugin UI changes hot at runtime", () => {
+    const invalidations: unknown[] = [];
+    const queryClient = {
+      invalidateQueries: (input: unknown) => {
+        invalidations.push(input);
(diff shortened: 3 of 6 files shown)


Polzer99:
## 🔍 Claude review — PR #8110 (`fix(plugins): refresh UI contributions on hot lifecycle events`)

Review-only pass (ping-pong lane: Codex codes → Claude reviews → human merges). Findings are grounded in the actual code at commit `69c2e58d`; each was adversarially re-verified before listing.

### Verdict
**No blocker, no security hole, spec-conformant (§25.4.5).** I'd request one small round before merge: 2 *medium* items (~30 lines), the rest optional.

### ✅ Correct by design — please don't "fix" these
- Broadcasting `plugin.ui.updated` to every company's socket is **intended**: plugins are *instance-global, operator-driven* (`PLUGIN_SPEC.md`). The payload `{pluginId, action}` carries no tenant data.
- The `plugin.ui.updated` early-return **before** the `event.companyId !== expectedCompanyId` guard is **load-bearing**: global events are stamped `companyId:"*"` (`live-events.ts:41`), so moving this branch below the guard would silently drop them and break hot-refresh.
- No client injection vector: the WS server has **no inbound `message` handler**.
- (A "partial registration mid-loop" hypothesis was raised and **refuted** — the registration loop `slots.tsx:417-441` is synchronous and cannot be preempted by the WS macrotask.)

---

### 🟠 Fix before merge (medium)

**M1 — In-flight import is not reset-safe (race).** `slots.tsx:839-851`, triggered from `LiveUpdatesProvider.tsx:573`
`resetPluginModuleLoader()` (promoted from a test-only helper) now runs on a live WS event with no in-flight guard. If a plugin module import is mid-flight, the orphaned IIFE later writes back into the just-`clear()`-ed maps:
- re-pollutes `registry` with the **old** component → transient stale render;
- leaves a phantom `pluginLoadStates` row (minor leak, never cleared until next reset);
- `inflightImports.delete(pluginId)` (keyed by `pluginId`) can delete the **freshly-registered v2** promise → enables a duplicate/double import.

*Fix:* capture a generation counter at the start of `loadPluginModule`, bump it in `resetPluginModuleLoader`, and gate all write-backs (registry/pluginLoadStates/inflightImports) on unchanged generation.

**M2 — Core §25.4.5 behavior is untested (tests).** `LiveUpdatesProvider.test.ts:8-25`
The new test asserts the two `invalidateQueries` keys but **never asserts `resetPluginModuleLoader()` runs** — which is the actual point of the PR. Deleting line 573 still passes CI (no `noUnusedLocals`, no lint/ESLint config). A regression that drops the module-cache clear ships green while plugin UI stays stale after a hot upgrade.

*Fix:* `vi.mock("../plugins/slots")` exposing a `vi.fn()` reset and `expect(resetPluginModuleLoader).toHaveBeenCalledTimes(1)`, **or** assert an observable side effect (register a fake component, invalidate, assert the registry is cleared).

---

### 🟡 Worth fixing (low)

**L1 — Don't revoke shared shim blob URLs on hot reset.** `slots.tsx:843-847`
The revoke loop nukes the `react` / `react-dom` / `sdk-ui` shim blob URLs, which an in-flight import's static deps point at → that import can reject. These shims are pure host re-exports (`slots.tsx:226-263`), **invariant across plugin versions**. Keep `shimBlobUrls` intact; only clear `pluginLoadStates`/`inflightImports`/`registry`. (Self-healing today, but free to fix and removes the recreate-churn.)

**L2 — Dead alias + inaccurate PR claim.** `slots.tsx:853-857`
`export const _resetPluginModuleLoader = resetPluginModuleLoader` has **zero importers** (`git grep` = the definition only); the consuming test (`slots.test.ts`) is already absent on this branch, so the "Backwards-compatible test helper alias" comment — and the PR body's "keeping the existing test alias" — don't hold. *Fix:* delete the alias + its doc comment.

**L3 — Defense-in-depth on the global channel.** `live-events.ts:37-44`
`publishGlobalLiveEvent` accepts the full `LiveEventType` union. Today only `plugin.ui.updated` (tenant-free payload) goes global, but nothing enforces that — a future caller adding a second global type or putting company data in the payload would silently broadcast it cross-tenant at the wire level. *Fix:* assert/narrow the accepted type(s) and add a comment that anything on the `"*"` channel is cross-tenant and must never carry company-scoped data.

**L4 — Missing test coverage on the new behavior's edges:**
- the early-return ordering (`companyId:"*"` vs a real `expectedCompanyId`) is untested, and `handleLiveEvent` isn't even exported via `__liveUpdatesTestUtils` — export a thin wrapper + add a positive (`"*"` → invalidations fire) and a negative (other type, mismatched companyId → dropped) test;
- the server dual-subscribe relay (`live-events-ws.ts:213-218`) has no test — add a focused unit test for subscribe-both / unsubscribe-both / the `readyState !== OPEN` guard.

---

### ⚪ Nits (optional)
- `LiveUpdatesProvider.tsx:575` — `invalidateQueries({queryKey: queryKeys.plugins.all})` (`["plugins"]`) already prefix-matches `["plugins","ui-contributions"]` under RQ v5 default `exact:false`, so the second call is redundant (the new test locks it in). Same pattern pre-exists at `PluginManager.tsx:94-96`.
- `LiveUpdatesProvider.tsx:622` — add a one-line comment that this branch is intentionally placed before the company guard (global `companyId:"*"`) so a future cleanup doesn't break the relay.
- `live-events-ws.ts:208` — type `sendEvent` as `(event: LiveEvent)` / `LiveEventListener` instead of `(event: unknown)`.

---
*Suggested path: bundle M1+L1 (same root cause, ~one helper change) and M2+L4 (test hardening), then merge. Nits can ride along or be skipped.*

Polzer99:
Codex follow-up pushed in `8d355983` after review.

Fixed before merge:
- Added generation guard for plugin UI module imports so stale in-flight imports cannot write back after `resetPluginModuleLoader()`.
- Hardened tests so the reset side effect is asserted directly.

Also addressed lows/nits:
- Kept shared shim blob URLs stable across hot resets.
- Removed unused `_resetPluginModuleLoader` alias and corrected PR body.
- Restricted `publishGlobalLiveEvent` to `plugin.ui.updated` and documented the tenant-free global channel.
- Board-only global WebSocket subscription; agent sockets remain company-scoped.
- Added global-event ordering and mismatched-company drop tests.
- Typed `sendEvent` as `LiveEvent` and added the load-bearing comment before the company guard.

Validation rerun:
- `pnpm --filter @paperclipai/ui exec vitest run src/context/LiveUpdatesProvider.test.ts`

No auto-merge from Codex; ready for review/merge decision.

PR #8112, opened 2026-06-14
fix(server): serve sw.js with no-cache so deploys land immediately

palombini-77:
## Thinking Path

- Paperclip's UI registers a service worker on every load: `ui/public/sw.js`, registered from `ui/src/main.tsx` via `navigator.serviceWorker.register("/sw.js")`.
- In `uiMode: "static"`, `server/src/app.ts` serves the UI dist through `express.static(uiDist, { maxAge: "1h" })`, with a `no-cache` override for `index.html` only.
- That means the service worker script itself is served with `Cache-Control: max-age=3600`.
- A browser's service-worker update check re-fetches the script through the HTTP cache, so for up to an hour after a deploy it revalidates against a cached copy and never installs the new worker.
- Until the new worker activates, the old one keeps answering navigations from its own cache, so clients continue to see the pre-deploy UI.
- Adding `sw.js` to the same `no-cache` override closes the gap: the update check always reaches the origin, and the new worker takes over on the next load.

No existing GitHub issue — the underlying problem is described inline below, per CONTRIBUTING.md → "Link Issues or Describe Them In-PR".

### What happened?

After deploying a new UI build, browsers with an existing session kept rendering the previous UI instead of the newly deployed one — in the reported case, a stale legacy sign-in screen. A hard reload, or manually unregistering the service worker, fixed it for that individual client.

### Expected behavior

After a deploy, the next page load serves the newly deployed UI. The service worker should not be able to pin clients to a previous build.

### Steps to reproduce

1. Run Paperclip with `uiMode: "static"` and load the UI once, so `/sw.js` is registered.
2. Build and deploy a new UI bundle to the same origin.
3. Reload within the hour — the old UI is still served. `/sw.js` is answered from the HTTP cache (`Cache-Control: max-age=3600`), so the worker update check never sees the new script and the previous worker stays active.

### Deployment mode

Self-hosted, `uiMode: "static"`, served behind a gateway.

## What Changed

- `server/src/app.ts`: the static `Cache-Control` override now covers `sw.js` in addition to `index.html`, so both are served `Cache-Control: no-cache`. The override is extracted into an exported `staticCacheControlOverride()` helper so it can be unit-tested directly, matching the existing pattern of `shouldServeViteDevHtml()`.
- `server/src/__tests__/app-static-cache-headers.test.ts` (new): covers the override — `index.html` and `sw.js` return `no-cache`, while other non-hashed static files (`favicon.ico`, `site.webmanifest`) are left on the default `max-age`.
- Hashed assets under `/assets/*` are untouched and keep their `max-age=1y, immutable` caching.
- Note on history: a debug backup artifact (`server/src/index.ts.bak-debug-*`) was added in an intermediate commit and removed again in `2ede3cb`. It is not part of the current diff.

## Verification

- Verified in production on the reporter's self-hosted deployment: after rebuilding and redeploying the UI, the current sign-in UI is served on the next load instead of the stale legacy screen, and the gateway stayed healthy through the deploy.
- Added a unit test for the cache-control override; it runs as part of the `General tests (server)` CI job.
- Hashed-asset caching is unchanged — `/assets/*` still responds with `max-age=31536000, immutable`.

## Risks

- Low. `sw.js` is a single small file, and `no-cache` means revalidate-before-use rather than "never cache", so the cost is one conditional request per load — a `304` when the worker is unchanged.
- No change to hashed asset caching, no API contract change, no database schema change.
- The change is behavioural only at the HTTP header level; there is no change to the service worker's own logic.

## Model Used

- Cursor Codex 5.3 — original fix and initial PR preparation by the author.
- Claude Opus 4.8 — follow-up test and this description, prepared via the maintainers' PR triage tooling.

## Checklist

- [ ] I have run tests locally and they pass — the added test is verified by the `General tests (server)` CI job rather than locally


palombini-77:
@paperclipai/maintainers
Could someone with merge permissions please merge PR #8112?

- The accidental debug backup file was removed in commit 2ede3cb.
- Final diff now only contains server/src/app.ts (sw.js no-cache fix).
- My account (palombini-77) cannot execute MergePullRequest on upstream.

Thanks!

PR #8114, opened 2026-06-14
fix(server): default plugin host version to serverVersion instead of "0.0.0"

### LINKED ISSUE #8115: Plugin host-version gate always reads 0.0.0, rejecting every version-gated plugin
### What happened?

Installing a plugin whose manifest declares a `minimumHostVersion` fails at the version gate with:

> Plugin `<id>` requires host version `<X>` or newer, but this server is running `0.0.0`

even though `GET /api/health` reports the correct server version.

Root cause: the plugin loader (`server/src/services/plugin-loader.ts`) compares a plugin's `minimumHostVersion` against `runtimeServices.instanceInfo.hostVersion`. That value comes from `createApp`'s `opts.hostVersion` in `server/src/app.ts`, but the `run` boot never passes it, so it falls back to the literal placeholder `"0.0.0"`. `/api/health` is unaffected because it reads `serverVersion` from `server/src/version.ts` directly — so the two disagree.

### Expected behavior

The plugin version gate should compare against the real running server version (the same value `/api/health` reports), so a version-gated plugin whose requirement the server already satisfies installs successfully.

### Steps to reproduce

1. Run a Paperclip server (any deployment mode).
2. `paperclipai plugin install <plugin-with-minimumHostVersion>` — e.g. a plugin that requires `2026.525.0`, on a server whose actual version is newer.
3. The install is rejected with `... but this server is running 0.0.0`.
4. Meanwhile `curl <server>/api/health` shows the correct, newer version.

Reproduced on `2026.609.0`; root cause is on `master` at `server/src/app.ts` (`hostVersion: opts.hostVersion ?? "0.0.0"`).

### Deployment mode

Not mode-specific — the `run` boot never sets `hostVersion` in any mode, so the gate always sees `0.0.0`.

### Agent adapter(s) involved

### Additional context

Fix proposed in PR #8114: route the default through a `resolveHostVersion()` helper that falls back to `serverVersion` (the same source `/api/health` uses), plus a unit test.

thebiltheory:
## Thinking Path

> - It ships a plugin system where a plugin manifest can declare a `minimumHostVersion`, and the loader gates installs on it (`server/src/services/plugin-loader.ts` compares it against `runtimeServices.instanceInfo.hostVersion` via `compareSemver`)
> - That `hostVersion` is threaded in through `createApp`'s `opts.hostVersion` (`server/src/app.ts`), but the `run` boot never passes it, so it fell back to the literal placeholder `"0.0.0"`
> - As a result, **every** plugin that declares a `minimumHostVersion` is rejected at install with `requires host version X or newer, but this server is running 0.0.0` — even though `GET /api/health` reports the correct version (it reads `serverVersion` from `server/src/version.ts`)
> - This pull request routes the default through a small `resolveHostVersion()` helper that falls back to `serverVersion` — the same canonical source `/api/health` already uses
> - The benefit is that version-gated plugins install correctly and the host version is consistent between the health route and the plugin gate

Fixes #8115

Filed with the full `bug_report.yml` fields in #8115. Summary:

- **What happened:** installing a plugin whose manifest declares a minimum host version fails with `Plugin <id> requires host version <X> or newer, but this server is running 0.0.0`.
- **Expected behavior:** the gate compares the plugin's requirement against the real running server version.
- **Steps to reproduce:** install any plugin that sets `minimumHostVersion` (e.g. `@tomismeta/paperclip-aperture`, which requires `2026.525.0`) on a server whose actual version is newer — it is still rejected as `0.0.0`.
- **Version / commit:** reproduced on `2026.609.0`; root cause is on `master` at `server/src/app.ts` (`hostVersion: opts.hostVersion ?? "0.0.0"`).
- **Deployment mode:** observed on `local_trusted` (embedded), but affects all modes — the boot never sets `hostVersion`, so the gate always sees `0.0.0`.
- **Inconsistency:** `/api/health` already reports the correct version (it reads `serverVersion`); only the plugin gate saw `0.0.0`.

## What Changed

- Added `resolveHostVersion(explicitHostVersion?)` in `server/src/app.ts`, which falls back to `serverVersion` (the server's own package version) instead of the `"0.0.0"` placeholder.
- Routed `instanceInfo.hostVersion` (passed to the plugin loader in `createApp`) through that helper.
- Added `server/src/__tests__/app-host-version.test.ts` covering the fallback, explicit-override passthrough, and a regression guard against `"0.0.0"`.

## Verification

- New unit test: `pnpm --filter @paperclipai/server exec vitest run src/__tests__/app-host-version.test.ts` → **3/3 pass**.
- Type safety: `pnpm --filter @paperclipai/server typecheck` → **clean (exit 0)**.
- `node scripts/check-forbidden-tokens.mjs` → **no forbidden tokens**.
- Manual: before this change `instanceInfo.hostVersion` resolves to `"0.0.0"` and version-gated installs fail; after, it resolves to `serverVersion` (matching `/api/health`).

## Risks

Low. An additive helper plus a single default change. The `??` only affects the case where no explicit `hostVersion` is passed (the boot path); any caller that passes one is unchanged. No migration, schema, or API change. The only behavioral shift is that an unset host version now reports the real server version instead of `"0.0.0"`.

## Model Used

Claude Opus 4.8 (Anthropic) — model ID `claude-opus-4-8`, 1M-context variant, extended thinking + tool use, run via Claude Code.

## Checklist

- [x] If this change affects the UI, I have included before/after screenshots — N/A (no UI change)
- [x] I have updated relevant documentation to reflect my changes — N/A (behavior corrected; helper documented inline)
- [x] All Paperclip CI gates are green — will confirm once CI runs on this PR
- [x] Greptile is 5/5 with no open P2s, recommendations, or follow-ups — will address after Greptile runs

---


cjonesde:
Confirming this still reproduces on the current `latest` `2026.707.0` and on the newest canary `2026.711.0-canary.6`: `server/src/app.ts` still has `hostVersion: opts.hostVersion ?? "0.0.0"` and the `run` boot never passes it, so every plugin declaring a `minimumHostVersion` is rejected with `... but this server is running 0.0.0` while `/api/health` reports the real version. I traced it independently before finding this PR. The `resolveHostVersion()` fallback to `serverVersion` here matches `/api/health`'s source, and the branch still applies cleanly with all checks green. Would be great to see this merged. Thanks @thebiltheory.

zachlendon (approved):
Verified the exact-head implementation and green CI. The helper uses the same canonical serverVersion as /api/health, preserves explicit overrides, and tests the 0.0.0 regression. This is the canonical fix; a separate maintenance-safety PR will reference rather than duplicate it.

thebiltheory:
Thanks @zachlendon! Let me know if it needs anything else.

zachlendon:
@cryppadotta @devinfoley — maintainer merge requested. This remains reproducible on the current latest/canary releases, the branch applies cleanly, all CI is green, Greptile is 5/5, and I approved the exact head `90a2731a91882cf71457dabb25ad3ae4b51ed3c1` after independently tracing the host-version path. Thanks @thebiltheory.

zachlendon:
Thanks @thebiltheory — nothing else is needed from you based on my review. The exact head is approved, mergeable, CI is green, and Greptile is 5/5. I have asked the upstream maintainers to merge it; we are also carrying the exact patch in our owned downstream distribution so our operational delivery is not blocked on upstream timing. We will retire the downstream patch after this lands upstream and passes our canary.

PR #8127, opened 2026-06-14
fix(invite): consume invite token after signup-via-invite for non-bootstrap company invites

stubbi:
## The bug

When a brand-new user (no existing account) opens a **non-bootstrap company invite** link and signs up inline on `InviteLanding`, they get an account but are **never added to the inviting company's membership** — the invite token is never consumed.

Root cause: `authMutation.onSuccess` (post-signup) had an early `return` for any invite that is not `bootstrap_ceo`. It only called the accept mutation for bootstrap invites:

```ts
if (!invite || invite.inviteType !== "bootstrap_ceo") {
  return; // <- non-bootstrap company invite bails out here, accept never runs
}
```

The auto-accept effect that would normally pick this up only fires once the session query has refreshed to an authenticated session. In the common race where `onSuccess` runs before that refresh lands, nobody consumes the invite, so the new user is left with an account but no company.

## The fix

In the post-signup path, drop the non-bootstrap early return and call `acceptMutation.mutateAsync()` for company invites too, then navigate the newly joined user into the company (mirroring the existing-account accept path). The accept mutation's own `onSuccess` already clears the remembered invite token and invalidates the relevant caches. The `bootstrap_ceo` branch keeps its exact existing behavior.

## Test

Adds a regression test that pins the responsibility on `onSuccess`: with the session query held at `null` (so the auto-accept effect can never fire), a fresh sign-up must still consume the invite token (`acceptInvite` called) and select the company. The test fails on the old code (accept called 0 times) and passes with the fix. All existing `InviteLanding` tests stay green (14/14).

### DIFF
--- a/ui/src/pages/InviteLanding.test.tsx
+++ b/ui/src/pages/InviteLanding.test.tsx
@@ -506,6 +506,84 @@ describe("InviteLandingPage", () => {
     });
   });
 
+  it("consumes a non-bootstrap invite from sign-up onSuccess even when the session query has not refreshed yet", async () => {
+    // The session query keeps resolving to null for the whole flow, so the
+    // auto-accept effect (which requires a session) can never fire. This pins
+    // the responsibility on authMutation.onSuccess: after a fresh sign-up it
+    // must consume the invite token itself, otherwise the brand-new user gets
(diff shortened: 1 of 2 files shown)


stubbi:

PR #8128, opened 2026-06-14
fix(ui): reliably auto-join authenticated invitees + follow deployer theme

stubbi:
Two UX gaps on `InviteLanding` for a valid human invite opened by an already-authenticated user (e.g. reached via a post-signup redirect).

## 1. Auto-accept latched off after a single transient failure
`shouldAutoAcceptHumanInvite` required `error === null`, and the one-shot `autoAcceptStarted` flag was never reset. So any transient accept failure (session/membership not propagated yet) permanently disabled auto-accept and stranded the invitee on a stale error with only a manual button — the invite never completed on its own.

**Fix:** replace the boolean flag with a bounded attempt counter (`MAX_AUTO_ACCEPT_ATTEMPTS = 2`) and drop the `error === null` latch, so a single transient failure auto-retries once and then falls back to the manual button. Also gate on `Boolean(invite)` and dispatch at most one accept per attempt (a ref guard), so the effect can never fire before the invite query has settled.

## 2. Hardcoded dark theme
The page hardcoded `zinc` dark classes and always rendered dark even when the document was bootstrapped to a light/branded theme. It's reached before the normal app shell, so it never inherited the theme through component tokens.

**Fix:** switch the whole page to the shadcn theme tokens (`bg-background` / `bg-card` / `text-foreground` / `text-muted-foreground` / `border-border` / `destructive` / `primary`), so it follows the active theme on both light and dark.

## Tests
Adds: auto-accept-on-mount for an already-authenticated invitee; transient-failure auto-retry recovery; and a theme-token assertion (no hardcoded `zinc` surfaces). `vitest` 16/16 green; `tsc -b` clean.

### DIFF
--- a/ui/src/pages/InviteLanding.test.tsx
+++ b/ui/src/pages/InviteLanding.test.tsx
@@ -506,6 +506,115 @@ describe("InviteLandingPage", () => {
     });
   });
 
+  it("auto-accepts a fresh human invite for a user who arrives already authenticated", async () => {
+    // Repro of the cloud signup-via-invite journey: the brand-new user signs up
+    // on the marketing /auth pages, then is redirected (already authenticated)
+    // to /invite/:token. They never touch InviteLanding's inline auth, so the
+    // page must auto-accept on mount. Regression guard: it must NOT surface
+    // "Invite not found" (a stale acceptMutation closure firing before the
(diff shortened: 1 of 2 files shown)


stubbi:

PR #8129, opened 2026-06-14
test(server): root-fix flaky heartbeat teardown CONNECTION_ENDED/DESTROYED race

stubbi:
## Problem

Multiple server test suites that drive the heartbeat service intermittently fail with `write CONNECTION_ENDED 127.0.0.1:<port>` (and the sibling `write CONNECTION_DESTROYED`) thrown mid-query from heartbeat code: `executeRun`, `releaseEnvironmentLeasesForRun`, `getCurrentUserRedactionOptions`, and `startNextQueuedRunForAgent` (queries against `environment_leases`, `agents`, `environments`, `instance_settings`). It surfaces in "General tests (server)" and "Verify serialized server suites" and is non-deterministic (timing-dependent).

## Root cause

`heartbeatService.startNextQueuedRunForAgent` dispatches run work as a **detached** `void executeRun(...)` promise. That chain keeps issuing queries against the DB client *after* the caller (`enqueueWakeup` / `resumeQueuedRuns` / `tickTimers` / `executeRun`'s own `finally`) has already resolved. When a test body finishes and teardown closes the postgres-js client (or stops the embedded server), the still-in-flight detached query rejects against the torn-down socket.

A graceful `end({ timeout })` is **not** sufficient: the detached chain issues *new* queries after `end()` is called, and those reject immediately.

## Fix

- **`server/src/services/heartbeat.ts`**: track the detached `executeRun` promises in a per-instance `pendingExecutions` set and expose `drain({ timeoutMs?, maxPasses? })` that awaits them. It is bounded by a 5s default deadline so a run genuinely stuck on external I/O (a gateway `agent.wait` that never returns, or a scheduled retry) cannot hang teardown. This is the missing graceful-stop primitive for the service and is equally useful for a graceful scheduler shutdown.
- **`server/src/__tests__/helpers/embedded-postgres.ts`**: add `closeDbClient(db)` (graceful `end({ timeout: 5 })`), used in every affected suite's `afterAll` **before** the embedded server is stopped.
- Affected suites now `await heartbeat.drain()` before closing/clearing so in-flight executions finish first. Suites that create a heartbeat per test register the instances and drain them in `afterEach`.

## Stability evidence

Each previously-flaky suite was run repeatedly (5-10x). `CONNECTION_ENDED`/`CONNECTION_DESTROYED` occurrences went from ~20 per 10 runs on pristine `master` to **0 across 55 runs** spanning all 11 affected suites.

Two unrelated, pre-existing flakes remain and reproduce on pristine `master` (out of scope here): a ~90s `waitFor` timeout in one comment-wake test, and a `checkoutRunId` assertion in `heartbeat-process-recovery`.

## Note

This supersedes the narrow band-aid previously landed in `heartbeat-comment-wake-batching.test.ts` (which used `end({ timeout: 0 })` + a suite-level `retry: 2`). That file now uses the real `drain()` and drops the retry.

### DIFF
--- a/server/src/__tests__/heartbeat-accepted-plan-workspace-refresh.test.ts
+++ b/server/src/__tests__/heartbeat-accepted-plan-workspace-refresh.test.ts
@@ -30,6 +30,7 @@ import {
 } from "@paperclipai/db";
 import {
   getEmbeddedPostgresTestSupport,
+  closeDbClient,
   startEmbeddedPostgresTestDatabase,
 } from "./helpers/embedded-postgres.js";
 import { heartbeatService } from "../services/heartbeat.ts";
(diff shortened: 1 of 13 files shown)


stubbi:

PR #8133, opened 2026-06-14
fix(ui): land invitees on the inviter company when membership lags

stubbi:
## Problem

When an authenticated user opens `/invite/:token` as their **first page load** (no selected/last company yet), `InviteLanding` auto-accepts the invite and then hands off to the company root redirect via `navigate("/")`. That redirect resolves the destination company by looking the selected company id up in the companies list, so the just-joined company must be present in the list **before** we navigate.

For a freshly signed-up user (or any deployment that auto-provisions a default company), the new membership is frequently not yet readable by the companies endpoint on the immediate post-accept refetch. The old code fired a single `invalidateQueries` and navigated right away, so the root redirect could not find the selected company and fell back to `companies[0]` (the user's own default company) — stranding them un-joined on the wrong board.

An established user who already had the dashboard loaded (companies cached, selection set) was unaffected, which is why this only reproduced on the first authenticated load.

## Fix

Refetch the companies list a bounded number of times (with a short backoff) until the joined company appears, then select + navigate. The selection is set up front so it persists across the refetch loop. If the list is still lagging after the bounded retries we navigate anyway and let a later refetch land the right company, rather than silently choosing `companies[0]`.

Applied to both the `acceptMutation` success path and the post-sign-in existing-member path.

## Test

Added a regression test: an already-authenticated invitee whose membership lags several refetches behind the accept now refetches until the joined company appears before navigating. FAIL→PASS on this change.

`tsc -b` clean; `InviteLanding.test.tsx` green.

### DIFF
--- a/ui/src/pages/InviteLanding.test.tsx
+++ b/ui/src/pages/InviteLanding.test.tsx
@@ -634,6 +634,88 @@ describe("InviteLandingPage", () => {
     });
   });
 
+  it("refreshes the companies list until the joined company appears before redirecting an authenticated invitee", async () => {
+    // An already-authenticated user opens /invite/:token as their FIRST page load
+    // (no selected/last company yet). The accept auto-fires, but the new
+    // membership is not yet readable by the companies endpoint. If the page
+    // navigates to "/" on that first stale list, the company root redirect cannot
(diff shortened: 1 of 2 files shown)


stubbi:

PR #8139, opened 2026-06-15
fix(server): watchdog/timeout for built-in DB backup scheduler

pjhagents:
## Problem

The built-in hourly DB backup scheduler can wedge **permanently** after a single hung run.

In `server/src/index.ts`, `runServerDatabaseBackup` sets a module-level `databaseBackupInFlight` flag and only clears it in `finally`. The scheduler `await`s the backup. If `runDatabaseBackup` **hangs** (awaited promise never settles — no return, no throw), `finally` never runs, the flag stays `true`, and every subsequent scheduled tick is skipped with `WARN Skipping scheduled database backup because a previous backup is still running`. Only a server restart clears it.

We hit this in production: one hung run wedged hourly backups for ~13 h until the wedge was noticed and the server restarted.

## Fix

- **Watchdog/timeout** — new `server/src/scheduled-database-backup.ts` exposes `createScheduledDatabaseBackupRunner()`, which races each run against a bounded timeout. The wrapper **always settles**, so the in-flight flag **always clears** and the next tick proceeds.
- **Orphan cleanup** — on timeout, sweep uncompressed `paperclip-*.sql` temp dumps from `data/backups` (completed `.sql.gz` preserved). Thrown failures were already cleaned by the backup lib.
- **Wedge alarm** — log `ERROR` after N consecutive scheduled skips (default 3) so a wedge is noticed within an hour, not ~13 h.
- **Config** — `database.backup.timeoutMinutes` (default 30 min, env `PAPERCLIP_DB_BACKUP_TIMEOUT_MINUTES`): generous over a healthy ~12 s run, comfortably under the 60 min interval.

## Verification

- `server/src/__tests__/scheduled-database-backup.test.ts` (7 tests, passing): simulates a hanging `runDatabaseBackup` and asserts the flag clears after timeout and the next scheduled run executes; asserts no orphan `.sql` remains after a simulated timeout/failure; covers the consecutive-skip alarm.
- `tsc --noEmit` clean on the server package.

## Notes

Root-cause fix for an internal incident (NET-340/NET-341). Generic robustness improvement — benefits any Paperclip instance running the built-in scheduled backup.

### DIFF
--- a/packages/shared/src/config-schema.ts
+++ b/packages/shared/src/config-schema.ts
@@ -25,6 +25,10 @@ export const databaseBackupConfigSchema = z.object({
   intervalMinutes: z.number().int().min(1).max(7 * 24 * 60).default(60),
   retentionDays: z.number().int().min(1).max(3650).default(7),
   dir: z.string().default("~/.paperclip/instances/default/data/backups"),
+  // Watchdog bound for a single backup run. Generous over a healthy run
+  // (~12s) but well under the default hourly interval so a hung dump is
+  // aborted and the next scheduled tick proceeds within the hour.
+  timeoutMinutes: z.number().int().min(1).max(7 * 24 * 60).default(30),
 });
 
 export const databaseConfigSchema = z.object({
@@ -37,6 +41,7 @@ export const databaseConfigSchema = z.object({
     intervalMinutes: 60,
     retentionDays: 7,
     dir: "~/.paperclip/instances/default/data/backups",
+    timeoutMinutes: 30,
   }),
 });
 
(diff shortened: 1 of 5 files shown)


pjhagents:
Closing as superseded/obsolete. The `server/src/scheduled-database-backup.ts` module this PR patched no longer exists on master — the scheduled-backup path was refactored inline into server/src/index.ts (`runServerDatabaseBackup`), which now includes a skip-if-a-previous-backup-is-still-running guard, alongside the new database-backup-health service and platform-managed backups on cloud instances. This branch can no longer be rebased against that structure. One residual gap: master's inline path guards against overlap but has no hard dump *timeout*; if that watchdog is still wanted it would be a fresh, small PR against `runServerDatabaseBackup`, not a rebase of this branch. Closing.

PR #8145, opened 2026-06-15
fix(ui): theme the onboarding wizard decorative panel instead of hardcoding dark

stubbi:
## Problem

The first-run onboarding wizard's right-hand decorative panel (which renders the ASCII paperclip illustration, in both `OnboardingWizard.tsx` and `OnboardingWizardClassic.tsx`) hardcodes `bg-[#1d1d1d]`. So it always renders a solid dark surface even when the document theme is light. Under a light theme, the cream/light shadcn-themed form sits next to a jarring dark panel on signup.

## Fix

Switch the panel background from the hardcoded `bg-[#1d1d1d]` to the shadcn `bg-muted` surface token (plus a `text-muted-foreground` default for any inherited glyphs).

The illustration glyphs in `AsciiArtAnimation.tsx` already use `text-muted-foreground/60`, which is the designed foreground pair for the `muted` surface. On `bg-muted` they read as an intentional ink-on-surface texture that follows the active theme in both light and dark, instead of light-on-dark only.

Only the colors change. The `hidden md:block` layout and the `transition-[width,opacity]` animation are untouched.

## Tests

Adds `OnboardingWizardTheme.test.tsx`: a source-level guard asserting neither wizard variant hardcodes a dark panel hex and both use the `bg-muted text-muted-foreground` surface tokens. `tsc -b` clean; the new test plus the existing `OnboardingWizardVariant` and `Auth` suites pass.

### DIFF
--- a/ui/src/components/OnboardingWizard.tsx
+++ b/ui/src/components/OnboardingWizard.tsx
@@ -1623,7 +1623,7 @@ export function OnboardingWizard() {
               name + mission steps) */}
           <div
             className={cn(
-              "hidden md:block overflow-hidden bg-[#1d1d1d] transition-[width,opacity] duration-500 ease-in-out",
+              "hidden md:block overflow-hidden bg-muted text-muted-foreground transition-[width,opacity] duration-500 ease-in-out",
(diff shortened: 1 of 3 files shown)


stubbi:

Issue #440, opened 2026-03-09
Plugin SDK: Add SSE bridge for worker-to-UI streaming


webprismdevin:
## Context

PR #432 introduces a plugin platform with a solid worker ↔ UI bridge (`usePluginData`, `usePluginAction`). The bridge is request-response: the UI sends a request, the worker returns one result.

This works well for CRUD operations but blocks a category of plugins that need **real-time streaming from worker to UI** — chat interfaces, log viewers, build output, agent monitoring dashboards.

## The Gap

The worker already receives streaming events from the host via JSON-RPC notifications (e.g., `AgentSessionEvent` objects from `agents.sessions.sendMessage` with `onEvent` callback). But it has no way to forward those events incrementally to the plugin UI.

```
Host → (JSON-RPC notifications) → Worker onEvent ✅ works great
Worker → ??? → Plugin UI ❌ no push channel
```

`performAction()` returns a single result when the action completes. During a 30-second streaming chat response, the UI sees nothing until the entire response finishes.

## Proposal: SSE Bridge Route

Add a server-side SSE endpoint that plugins can use to stream events to their UI:

```
GET /api/plugins/:pluginId/bridge/stream/:channel?companyId=...
```

### How it works

**Worker side** — new `ctx.streams` API:
```typescript
ctx.streams.open(channel: string, companyId: string): void
ctx.streams.emit(channel: string, event: unknown): void  
ctx.streams.close(channel: string): void
```

**Host side** — new route in `server/src/routes/plugins.ts`:
```typescript
router.get("/plugins/:pluginId/bridge/stream/:channel", (req, res) => {
  res.writeHead(200, {
    "Content-Type": "text/event-stream",
    "Cache-Control": "no-cache",
    Connection: "keep-alive",
  });
  
  // Register this connection to receive events from the worker
  const unsubscribe = pluginStreamBus.subscribe(pluginId, channel, companyId, (event) => {
    res.write(`data: ${JSON.stringify(event)}\n\n`);
  });
  
  req.on("close", unsubscribe);
});
```

When the worker calls `ctx.streams.emit(channel, event)`, it sends a JSON-RPC notification to the host, which fans out to all connected SSE clients for that plugin + channel + company.

**UI side** — new `usePluginStream` hook:
```typescript
import { usePluginStream } from "@paperclipai/plugin-sdk/ui";

function ChatPage() {
  const events = usePluginStream<AgentSessionEvent>("chat-stream");
[...]

PR #8150, opened 2026-06-15
fix(plugins): wire the plugin stream bus so SSE streaming actually works

### LINKED ISSUE #5879: [plugin-sdk] usePluginStream returns 501 — /api/plugins/<id>/bridge/stream/<name> not implemented in 2026.428.0
## Summary

When a plugin uses `usePluginStream(streamName)` from `@paperclipai/plugin-sdk`, the SDK opens an EventSource against `/api/plugins/<pluginKey>/bridge/stream/<streamName>`. On Paperclip 2026.428.0 that route returns **HTTP 501 Not Implemented** — the plugin SSE pass-through is registered in the SDK and TypeScript types but does not exist in the server router.

## Environment

- Paperclip version: `2026.428.0` (`GET /api/health` → `version`)
- Plugin SDK: `@paperclipai/plugin-sdk@<bundled in 2026.428.0>`
- Host OS: Ubuntu 24.04, Node 22

## Reproduction

1. Author a plugin with a worker `getData` that calls `subscribe()` on a stream (e.g. `worker.subscribePluginStream('mystream', handler)`) and a UI hook that calls `usePluginStream('mystream')`.
2. Install via `paperclipai plugin install --local <dir>` against a fresh instance.
3. Open the plugin UI and inspect the network panel.

**Observed:**
```
GET /api/plugins/<pluginKey>/bridge/stream/mystream
→ 501 Not Implemented
```
The UI hook's `error` state populates with `Failed to connect to plugin stream 'mystream'`.

**Expected:** an SSE stream to which the worker can `publish()` events that arrive in the UI hook.

## Impact

- Plugins that depend on near-realtime updates have to fall back to polling (15s in our case).
- No client-side workaround that *uses the SDK as-typed* — the route simply isn't there.
- This is not a 404, so it is unambiguously "implemented in client, missing on server" rather than a registration/typo issue.

## Suspected location

The complementary `subscribePluginStream` / `publishPluginStream` API exists in the plugin worker side and looks fully wired in the SDK. The missing piece is the server route that fans the worker's publishes out to the EventSource clients. Likely lives in the plugin bridge router next to the `/api/plugins/<id>/bridge/...` action endpoints.

## Workaround

Polling fetcher behind `usePluginData` works fine. Recommend the SDK either:
- Implement the missing route, or
- Have `usePluginStream` log a clearer "not implemented in this Paperclip version" diagnostic and short-circuit so plugins don't have to special-case 501 in their error handling.

harperaa:
## Thinking Path

> - Paperclip is the open source app people use to manage AI agents for work, and it ships a plugin SDK so third parties can extend the UI and workers.
> - The plugin runtime includes a real-time streaming capability: workers call `ctx.streams.open/emit/close`, the UI subscribes with `usePluginStream()`, and an SSE endpoint `GET /api/plugins/:pluginId/bridge/stream/:channel` bridges the two.
> - Every backing piece exists in core — `PluginStreamBus`, the worker RPC handlers for `streams.*`, the worker manager's per-worker `onStreamNotification` hook, the SSE route, and the UI hook.
> - But nothing ever constructs a `PluginStreamBus` or feeds the worker manager's stream hook, so the SSE route always sees `bridgeDeps.streamBus === undefined` and returns 501. `ctx.streams.emit()` silently goes nowhere and plugins are forced to poll.
> - This pull request connects the already-implemented pieces: it creates the bus, lets the worker manager forward stream notifications into it, and passes it to the plugin route bridge deps.
> - The benefit is that plugin real-time streaming actually works (SSE instead of 501), with no new dependencies and no behavior change for plugins that don't stream.

Closes #5879 — `usePluginStream` returns 501 because `/api/plugins/<id>/bridge/stream/<name>` is effectively unimplemented (the route exists but the bus is never wired).
Refs #440 — original request to add the SSE bridge for worker-to-UI streaming; the bridge landed but was never connected, which this PR completes.

I searched open and closed issues and PRs for plugin streaming / SSE / `usePluginStream` / `streamBus`; the two issues above are the only related items and I found no prior or competing PR.

## What Changed

- `server/src/services/plugin-worker-manager.ts`: add `setStreamNotificationHandler()` to the manager (plus an optional constructor option) so its stream handler can be attached after construction. `startWorker` forwards each worker's `streams.*` notifications to the live handler, unless that worker supplied its own. Reading the handler at notification time means it works whether the manager is created by `createApp` or injected by the caller.
- `server/src/services/plugin-stream-bus.ts`: add `publishWorkerStreamNotification()`, a shared mapping from a worker stream notification to a bus publish (drops ones missing a channel/companyId; maps `open`/`emit`/`close` to event types).
- `server/src/app.ts`: construct the `PluginStreamBus`, attach the handler to the worker manager via `setStreamNotificationHandler()`, and pass `streamBus` in the `pluginRoutes` bridge deps so the SSE route stops returning 501.
- `server/src/__tests__/plugin-stream-bus.test.ts`: new tests for bus delivery semantics, the notification→event-type mapping, and the settable manager handler.

Note on the design: the production entrypoint injects the worker manager into `createApp`, so a constructor-only hook would leave the bus orphaned (the SSE route would open connections that never receive events instead of failing fast). Attaching the handler to the manager after it is obtained covers both the injected and the factory-created manager.

## Verification

- `pnpm --filter @paperclipai/server exec vitest run src/__tests__/plugin-stream-bus.test.ts src/__tests__/plugin-worker-manager.test.ts` — 17/17 pass (6 new + 11 existing).

End-to-end, against a locally running server with a plugin installed (identifiers redacted), exercising the injected-manager path that production uses:

```
# Before (without this change) — the route's first guard returns 501:
GET /api/plugins/<plugin-id>/bridge/stream/<channel>?companyId=<company-id>
  -> HTTP 501  {"error":"Plugin stream bridge is not enabled"}

# After (this change) — the same request opens the SSE stream:
GET /api/plugins/<plugin-id>/bridge/stream/<channel>?companyId=<company-id>
  -> HTTP 200  (SSE stream opens; ":ok" keep-alive frame)
```

The before/after was observed on the same running instance by swapping only these files and letting the dev server reload between requests.

## Risks

Low. No new dependencies and no schema/migration changes. Behavior is unchanged for plugins that never open a stream channel. The only functional change is that `bridgeDeps.streamBus` is now defined and populated, so the SSE route serves the stream instead of 501ing. The manager-level handler is only forwarded to when a worker doesn't already provide its own `onStreamNotification`, preserving existing per-worker behavior.

## Model Used

Claude Opus 4.8 (`claude-opus-4-8`), 1M-context configuration, with extended thinking and tool use / code execution. Used to locate the wiring gap, implement the change, and run the verification above; all changes human-reviewed.

## Checklist

- [ ] I have updated relevant documentation to reflect my changes (N/A — the SDK streaming API is already documented; this only connects it)
- [ ] All Paperclip CI gates are green (will confirm after CI runs)
- [ ] Greptile is 5/5 with no open P2s, recommendations, or follow-ups (will address after review)

### DIFF
--- a/server/src/__tests__/plugin-stream-bus.test.ts
+++ b/server/src/__tests__/plugin-stream-bus.test.ts
@@ -1,0 +1,92 @@
+import { describe, expect, it, vi } from "vitest";
+import {
+  createPluginStreamBus,
+  publishWorkerStreamNotification,
+} from "../services/plugin-stream-bus.js";
+import { createPluginWorkerManager } from "../services/plugin-worker-manager.js";
+
+describe("PluginStreamBus", () => {
+  it("delivers a published event only to subscribers of the matching tuple", () => {
+    const bus = createPluginStreamBus();
+    const match = vi.fn();
+    const wrongChannel = vi.fn();
+    const wrongCompany = vi.fn();
+
+    bus.subscribe("plugin-a", "chan-1", "co-1", match);
+    bus.subscribe("plugin-a", "chan-2", "co-1", wrongChannel);
+    bus.subscribe("plugin-a", "chan-1", "co-2", wrongCompany);
+
+    bus.publish("plugin-a", "chan-1", "co-1", { type: "text", text: "hi" });
+
+    expect(match).toHaveBeenCalledTimes(1);
+    expect(match).toHaveBeenCalledWith({ type: "text", text: "hi" }, "message");
+    expect(wrongChannel).not.toHaveBeenCalled();
+    expect(wrongCompany).not.toHaveBeenCalled();
+  });
+
+  it("stops delivering after unsubscribe", () => {
+    const bus = createPluginStreamBus();
+    const listener = vi.fn();
+    const unsubscribe = bus.subscribe("p", "c", "co", listener);
+
+    bus.publish("p", "c", "co", { n: 1 });
+    unsubscribe();
+    bus.publish("p", "c", "co", { n: 2 });
+
+    expect(listener).toHaveBeenCalledTimes(1);
+    expect(listener).toHaveBeenCalledWith({ n: 1 }, "message");
+  });
+});
+
+describe("publishWorkerStreamNotification", () => {
+  it("maps streams.emit to a 'message' event and forwards the payload", () => {
+    const bus = createPluginStreamBus();
+    const listener = vi.fn();
+    bus.subscribe("p", "chat:thread", "co", listener);
+
+    publishWorkerStreamNotification(bus, "p", "streams.emit", {
+      channel: "chat:thread",
+      companyId: "co",
+      event: { type: "text", text: "token" },
+    });
+
+    expect(listener).toHaveBeenCalledWith({ type: "text", text: "token" }, "message");
+  });
+
+  it("maps streams.open and streams.close to their event types", () => {
+    const bus = createPluginStreamBus();
+    const listener = vi.fn();
+    bus.subscribe("p", "c", "co", listener);
+
+    publishWorkerStreamNotification(bus, "p", "streams.open", { channel: "c", companyId: "co" });
+    publishWorkerStreamNotification(bus, "p", "streams.close", { channel: "c", companyId: "co" });
+
+    expect(listener).toHaveBeenNthCalledWith(1, null, "open");
+    expect(listener).toHaveBeenNthCalledWith(2, null, "close");
+  });
+
+  it("drops notifications missing channel or companyId", () => {
+    const bus = createPluginStreamBus();
+    const publishSpy = vi.spyOn(bus, "publish");
+
+    publishWorkerStreamNotification(bus, "p", "streams.emit", { companyId: "co", event: {} });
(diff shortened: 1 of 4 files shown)


harperaa:
Thanks — good catch on the orphaned-bus case. Addressed in the latest commit.

The production entrypoint injects the worker manager into `createApp`, so wiring the handler only through the manager's constructor would have left the bus orphaned exactly as described: the SSE route would open connections that never receive events, instead of failing fast with 501.

Fix: the worker manager now exposes `setStreamNotificationHandler()`, and `app.ts` attaches the bus handler to the manager *after* obtaining it. The per-worker forwarder reads the handler at notification time, so it works whether the manager is created by `createApp` or injected by the caller. The notification→event-type mapping is factored into a shared `publishWorkerStreamNotification()` helper, and there's a new `plugin-stream-bus.test.ts` covering delivery semantics, the mapping, and the settable handler.

Re-verified end to end on the injected-manager path (the SSE route returns 200 and opens the stream instead of 501).

PR #8170, opened 2026-06-15
fix(skills): capture full skill directory on GitHub import

sunguangran:
## Problem

GitHub-sourced skills only recorded `SKILL.md` in their `file_inventory`, dropping every sibling `references/`, `scripts/` and `assets/` file. Because the inventory drives **both** the Skill detail UI **and** runtime materialization (`materializeRuntimeSkillFiles` iterates `fileInventory`), those files were invisible in the app **and never delivered to agents** — so multi-file skills ran degraded, with broken "see references/x.md" links.

Observed on a live instance: 100% of GitHub-imported skills had `file_inventory` length 1, even when the upstream repo clearly shipped a `references/` directory (e.g. `legacy-modernizer` has 5 references upstream, stored as 1 file).

## Root cause

Two collapse points in `readUrlSkillImports` (`server/src/services/company-skills.ts`):

1. **File-targeted source URL** (`…/SKILL.md`): the `parsed.filePath` filter narrowed the candidate file set to *that single file*, instead of selecting the skill and keeping its directory subtree. Now `filePath` is used only to pick *which* skill to import; the inventory is rebuilt from the full scoped path list.
2. **SKILL.md at repo root**: `skillDir` is `"."`, so `startsWith(\`${skillDir}/\`)` became `"./"`, which never matches sibling paths like `references/foo.md`. Root skills now treat the whole scope as the skill, and `repoSkillDir` stores a `"."` sentinel that `readFile` resolves to the repo root (no slug fallback).

The same root-dir prefix bug in the local-directory import branch is fixed alongside it.

## Changes

- `readUrlSkillImports`: decouple "select skill" from "collect skill files"; build inventory from the full scoped subtree; handle `skillDir === "."`.
- `repoSkillDir` metadata: store `"."` for repo-root skills; use `path.posix.join` to avoid `foo/.` artifacts.
- `readFile` (github/skills_sh): resolve `"."` repoSkillDir to repo root.
- Local-directory import branch: same `skillDir === "."` fix.

## Impact / compatibility

- No behavior change for skills genuinely containing only `SKILL.md`.
- Reference content is still fetched live from GitHub by path at read time — no extra storage; fixing the inventory is sufficient for both UI listing and runtime delivery.
- Existing rows can be repaired by re-importing the affected skills (inventory is rebuilt on import).

### DIFF
--- a/server/src/services/company-skills.ts
+++ b/server/src/services/company-skills.ts
@@ -1118,10 +1118,11 @@ async function readLocalSkillImports(companyId: string, sourcePath: string): Pro
   const imports: ImportedSkill[] = [];
   for (const skillPath of skillPaths) {
     const skillDir = path.posix.dirname(skillPath);
+    const isRootSkill = skillDir === ".";
     const inventory = allFiles
-      .filter((entry) => entry === skillPath || entry.startsWith(`${skillDir}/`))
+      .filter((entry) => entry === skillPath || (isRootSkill ? true : entry.startsWith(`${skillDir}/`)))
       .map((entry) => {
-        const relative = entry === skillPath ? "SKILL.md" : entry.slice(skillDir.length + 1);
+        const relative = entry === skillPath ? "SKILL.md" : isRootSkill ? entry : entry.slice(skillDir.length + 1);
         return {
           path: normalizePortablePath(relative),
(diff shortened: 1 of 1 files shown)


starlein:
Already in latest upstream incldued

PR #8202, opened 2026-06-16
fix(plugin-worker-manager): recover when a worker's stdin command channel dies

coal:
## Thinking Path

> - Plugins run as child processes. The host talks to each worker over newline-delimited JSON-RPC on the child's stdin and stdout pipes.
> - `plugin-worker-manager` already supervises two failure modes: the child exits, and the child errors. Both route into `handleProcessExit()` and the exponential-backoff restart.
> - There is a third mode it does not supervise. The host to worker **stdin** pipe can die (EPIPE) while the child process stays alive.
> - When that happens the worker is uncommandable. Every `sendMessage` throws `Worker process for plugin "<id>" is not writable`, so `http.fetch` proxying and every other host to worker call fails permanently.
> - `child.on("exit")` never fires, because the process is still running. So the recovery path is never reached and the worker zombies: alive and polling, but deaf to commands. Only a manual disable and re-enable clears it.
> - This pull request supervises the command channel the same way the process is already supervised, rather than adding a polling watchdog.
> - The benefit is that the existing restart machinery now covers all three ways a worker can stop being usable.

No existing issue covers this. Searched open and closed issues and PRs for plugin worker stdin, "not writable", and restart supervision, and found no duplicate or prior attempt. Describing it inline per `.github/ISSUE_TEMPLATE/bug_report.yml`.

**What happened?**

A plugin worker stopped responding to all host commands while its process stayed alive. Every `sendMessage` threw `Worker process for plugin "<id>" is not writable`. The crash-recovery and backoff path never ran, because `child.on("exit")` never fired. The worker stayed in that state indefinitely.

Observed in production on a Telegram-bridge plugin. Its `getUpdates` long-poll began failing after roughly five days of uptime and never recovered until the plugin was manually disabled and re-enabled. The logs showed the same pair on every poll cycle:

```
ERROR host handler error {method: http.fetch}  err: "Worker process for plugin ... is not writable"
ERROR [plugin] getUpdates failed
```

**Expected behavior**

A worker that can no longer be commanded should be restarted by the existing supervision path, exactly as a worker that exits or errors already is. A dead command channel should not be a silent, permanent, manual-recovery-only state.

**Steps to reproduce**

1. Start a plugin worker that makes regular host calls, for example anything using `http.fetch` on a timer.
2. Destroy the host to worker stdin pipe while leaving the child process running. In production this happened by itself as an EPIPE after several days of uptime.
3. Watch subsequent host calls fail with `Worker process for plugin "<id>" is not writable`.
4. Observe that `child.on("exit")` does not fire, no restart is scheduled, and the worker never recovers on its own.

Reproduced against `server/src/services/plugin-worker-manager.ts` on master. This branch is rebased onto aa6b6bcb.

**Deployment mode**

Self-hosted, long-running server process. Uptime matters for this bug: it appeared after about five days, so it is unlikely to be seen in a short-lived dev session.

## What Changed

- `sendMessage` now passes a write callback, so an asynchronous stdin write error such as EPIPE is surfaced and logged instead of silently swallowed.
- `attachStdioHandlers` now attaches `error` and `close` handlers to `child.stdin`.
- If the command channel dies while the process is still alive (`exitCode` and `signalCode` both null) and the stop was not intentional, the child is `SIGKILL`ed. That makes the normal `handleProcessExit()` and `scheduleRestart()` recovery run.
- No new polling loop and no watchdog timer. This mirrors the existing `exit` and `error` handlers and reuses the existing exponential-backoff restart machinery.
- Guarded against firing during a graceful stop, and against firing after the child has already exited.

## Verification

- The source change is confined to one file, `server/src/services/plugin-worker-manager.ts`, at +38/-1. A later commit adds tests only — see below.
- Full CI passed on this change at the time it was opened, and is re-running now against current master after a rebase.
- Behavioural check: with the stdin channel destroyed underneath a live child, the worker is now killed and restarted through the existing backoff path, instead of remaining alive and uncommandable.
- The intentional-stop guard is what keeps this from firing on every normal shutdown, so that is the arm to review most closely.

### Tests

`server/src/__tests__/plugin-worker-stdin-supervision.test.ts`, plus a `plugin-worker-persistent.cjs` fixture that deliberately survives stdin EOF (ref'd keep-alive timer, `readline` `"close"` is a no-op) so nothing but the host supervision can end it. It follows the existing harness — a real child process spawned from a fixture entrypoint — and captures the child by wrapping `node:child_process.fork` with a pass-through mock; the fork itself is real. It is a separate file because `vi.mock` is file-scoped and `plugin-worker-manager.test.ts` needs an unmocked `fork`.

1. **Positive** — `child.stdin` destroyed with `EPIPE` while the process is alive → the manager `SIGKILL`s the child and schedules a restart through the existing backoff (`crash` with `willRestart: true`, status `backoff`, `diagnostics().nextRestartAt` in the future).
2. **Negative control for the risky arm** — the same stdin death *during* an intentional stop must not force-kill or restart. The fixture defers its exit after acking `shutdown`, which opens the window to hit it mid-stop.

Both arms were checked by mutation rather than only by going green:

| Mutation | Result |
| --- | --- |
| `plugin-worker-manager.ts` reverted to the pre-fix parent | **case 1 fails only** (`worker was left alive and uncommandable after its command channel died`); case 2 stays green |
| only the `intentionalStop \|\|` guard removed | **case 2 fails only** (`expected 'SIGKILL' to be null`); case 1 stays green |

Each mutation reds exactly one case and leaves the other green, so neither is passing for a generic reason. The positive case reports "still alive" as a value rather than letting the suite time out, so the pre-fix failure is a named assertion instead of a bare timeout; the negative case waits for the `shutdown` to drain out of the host before killing the pipe, so it cannot pass by dropping the shutdown instead of exercising the guard.

Sibling suite unaffected (30 tests pass across both files), `tsc --noEmit` clean on `server`, and the new suite lands in exactly one of the four glob-discovered `general-server` shards, so it does run on CI.

## Risks

- **The main risk is killing a worker that was actually healthy.** The trigger is narrow — a dead stdin channel on a process that is still alive and was not asked to stop — but a bug in the intentional-stop guard would turn graceful shutdowns into SIGKILLs. That guard is the review-worthy part of this change.
- Restart amplification: a worker whose stdin dies repeatedly will now restart repeatedly rather than sit idle. It goes through the existing exponential backoff, so this is bounded, but the failure becomes visible and noisy where it used to be silent.
- Behavioural shift, intended: a state that previously required manual operator intervention now self-heals.
- No database migration, no schema change, no API change, no configuration change.

## Model Used

Claude (Anthropic), used through Claude Code. The original change was authored on 2026-06-16 and the exact model ID for that session is not recorded, so it is not asserted here. This revision — the rebase onto master and this PR description — is Claude Opus 5, model ID `claude-opus-5`, extended thinking enabled, with tool use. The test commit and this description update are also Claude Opus 5, `claude-opus-5`.

## Checklist

Notes on the checklist:

- **Tests.** Now covered — see the Tests section under Verification. The earlier version of this description said no automated test covered this and asked for a maintainer's steer on how to model it; that is stale, and the approach taken is described above so it can be reviewed on its merits.
- **Documentation.** No user-facing behaviour or configuration changes, so nothing in `docs/` is affected.

This PR was opened on 2026-06-16 and has been rebased onto current master. Apologies for the delay in completing the description — the template sections were missing, which is why `review` was red.

### DIFF
--- a/server/src/__tests__/fixtures/plugin-worker-persistent.cjs
+++ b/server/src/__tests__/fixtures/plugin-worker-persistent.cjs
@@ -12,8 +12,16 @@
 
 const readline = require("node:readline");
 
-/** How long the worker waits after acking `shutdown` before exiting. */
-const SHUTDOWN_EXIT_DELAY_MS = 300;
+// How long the worker waits after acking `shutdown` before exiting.
+//
+// This has to sit inside the host's post-ack grace period: stopInternal()
+// races the shutdown RPC (which resolves as soon as this ack lands) and then
+// waits only 500ms more before escalating to SIGTERM. A delay close to that
+// ceiling makes the negative-control test a timing race against a real process
+// exit on a loaded CI runner, so keep the margin wide. The test does not
+// depend on this window being long — it kills the pipe from a write hook the
+// moment the shutdown is flushed, not after a poll.
+const SHUTDOWN_EXIT_DELAY_MS = 100;
 
 /** Hard ceiling so a fixture never outlives the test run that spawned it. */
 const MAX_LIFETIME_MS = 30_000;
--- a/server/src/__tests__/fixtures/plugin-worker-persistent.cjs
+++ b/server/src/__tests__/fixtures/plugin-worker-persistent.cjs
@@ -1,0 +1,82 @@
+// Long-lived worker fixture for the stdin command-channel supervision tests.
+//
+// Unlike the other fixtures, this worker deliberately survives losing its
+// stdin: readline "close" does not terminate it, and a ref'd keep-alive timer
+// holds the event loop open. That reproduces the field shape the supervision
+// fix exists for — the host->worker command pipe dies while the worker process
+// itself stays alive and uncommandable.
+//
+// `shutdown` is acknowledged immediately but the exit is deliberately deferred,
+// which leaves a window in which a test can kill the command channel *during*
+// an intentional stop (the negative control).
+
+const readline = require("node:readline");
+
+/** How long the worker waits after acking `shutdown` before exiting. */
+const SHUTDOWN_EXIT_DELAY_MS = 300;
+
+/** Hard ceiling so a fixture never outlives the test run that spawned it. */
+const MAX_LIFETIME_MS = 30_000;
+
+function send(message) {
+  process.stdout.write(`${JSON.stringify(message)}\n`);
+}
+
+// Ref'd, so the process stays alive after stdin reaches EOF.
+const keepAlive = setInterval(() => {}, 1_000);
+
+const selfDestruct = setTimeout(() => {
+  process.exit(0);
+}, MAX_LIFETIME_MS);
+
+function exitNow() {
+  clearInterval(keepAlive);
+  clearTimeout(selfDestruct);
+  process.exit(0);
+}
+
+const rl = readline.createInterface({
+  input: process.stdin,
+  crlfDelay: Infinity,
+});
+
+// Explicitly do NOT exit here: losing the command channel must leave this
+// process alive so the host supervision path is the thing under test.
+rl.on("close", () => {});
+
+rl.on("line", (line) => {
+  if (!line.trim()) return;
+  const message = JSON.parse(line);
+  const method = message && typeof message.method === "string" ? message.method : null;
+
+  if (method === "initialize") {
+    send({
+      jsonrpc: "2.0",
+      id: message.id,
+      result: {
+        ok: true,
+        supportedMethods: [],
+      },
+    });
+    return;
+  }
+
+  if (method === "shutdown") {
+    send({
+      jsonrpc: "2.0",
+      id: message.id,
+      result: {},
+    });
+    setTimeout(exitNow, SHUTDOWN_EXIT_DELAY_MS);
+    return;
+  }
+
+  send({
+    jsonrpc: "2.0",
+    id: message.id,
+    error: {
+      code: -32601,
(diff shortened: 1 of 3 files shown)


coal:
The one red on this head — `Verify serialized server suites (2/4)` — is **not from this PR**.

Failing test: `server/src/__tests__/issue-dependency-wakeups-routes.test.ts` → *"wakes dependents when the final blocker transitions to done"*, with `TypeError: db.select(...).from(...).where(...).orderBy is not a function` followed by `Test timed out in 5000ms`.

Why it can't be this change:

1. The last commit (`1efea22d`) differs from its parent (`505da32d`) **only** in this PR's own new test file and its fixture — no source file changed between those two heads.
2. All four `Verify serialized server suites` shards were `success` on that parent.
3. The new suite is in **zero** serialized shards (checked with `run-vitest-stable.mjs --dry-run --mode serialized --shard-index N --shard-count 4`, using the failing suite itself as a positive control so an empty result can't be mistaken for a silent no-op).
4. The failing test exists unchanged on base `master` and is not imported by any of the three files this PR touches.

I've deliberately **not** re-run that job to turn it green — it looks like a master-side issue in a test this PR doesn't touch, and re-rolling it would just hide it. Flagging rather than papering over.

`review` (the test-coverage gate) is green: the PR now adds `plugin-worker-stdin-supervision.test.ts`, covering both the supervision path and an intentional-stop negative control.

PR #8206, opened 2026-06-16
fix: scope board-UI 404 storm — downgrade expected client 404s + negative-cache mention prefetch (server + UI)

simberthon:
## Thinking Path

> - The server uses a `pino-http` request logger that writes to `server.log`; the board UI drives a steady stream of HTTP requests
> - Two browser request patterns produce a large volume of *expected* `404`s — mention-hovercard prefetch of issue keys that no longer resolve, and live-run log polling before a run has emitted any log
> - Logged at `warn`, these expected 404s dominate the log and bury genuine 404s, hurting observability; the prefetch/poll requests are also avoidable traffic
> - This PR downgrades only those two exact 404 shapes to `debug` (everything else stays `warn`) and stops the UI from generating the avoidable requests at the source
> - The benefit is dramatically quieter logs with real errors still visible, plus less redundant request volume

No upstream GitHub issue exists; describing the bug inline per the bug issue template:

**What happened**

On every request the file logger writes, expected board-UI 404s are recorded at `warn`. Two request patterns dominate the noise:
- `GET /api/issues/<KEY>` — the board prefetches a hovercard for every issue-key-like token in issue bodies, including deleted / cross-instance keys that always 404.
- `GET /api/heartbeat-runs/<ID>/log` — the live-runs view polls for a run's log before it exists.

Together these make up a large share of recent `server.log` lines, drowning out genuine 404s and generating avoidable request traffic.

**Expected behavior**

Expected, benign client 404s should not flood the `warn` log, and the UI should not keep re-issuing requests it already knows will 404. Genuine 404s (real routes, sub-paths, non-GET, UUID lookups) must still be logged at `warn`.

**Steps to reproduce**

1. Open the board UI against an instance whose issue bodies contain issue-key-like tokens that no longer resolve (deleted or cross-instance keys).
2. Tail `server.log` and watch the request logger: observe repeated `warn` entries for `GET /api/issues/<KEY>` 404s and `GET /api/heartbeat-runs/<ID>/log` 404s.
3. Note how the expected/benign 404s dominate the log volume relative to genuine errors.

**Deployment mode**

Self-hosted, file-logging (`pino-http` writing to `server.log`).

## What Changed

- **server:** Added `isExpectedClient404()` (`server/src/middleware/http-log-policy.ts`) — GET-only, 404-only, matching exactly `/api/issues/<KEY>` and `/api/heartbeat-runs/<ID>/log`. Wired into the `pino-http` `customLogLevel` so only those two shapes drop to `debug`; every other 404 (real routes, sub-paths, non-GET, UUID lookups) stays at `warn`.
- **ui:** Per-session negative cache for mention prefetch (`ui/src/lib/issueDetailCache.ts`) — record keys that 404, skip re-prefetching them, clear on a later success.
- **ui:** Stop polling a run's `/log` once it has 404'd (`ui/src/components/transcript/useLiveRunTranscripts.ts`); resume automatically once byte counts show the log exists.

## Verification

- `npx vitest run server/src/middleware/http-log-policy.test.ts ui/src/lib/issueDetailCache.test.ts ui/src/components/transcript/useLiveRunTranscripts.test.tsx` → 21 tests pass.
- `http-log-policy.test.ts` asserts the bounding precisely: only the two GET 404 shapes downgrade; real routes, sub-paths, non-GET, and UUID 404s all remain `warn`.
- `tsc --noEmit` clean in `server/`.

## Risks

Low risk. The only logging behavior change is narrowing two specific expected 404 shapes from `warn` to `debug`; all other observability is unchanged. The UI changes are scoped to a per-session in-memory cache and poll suppression, both self-correcting (cleared/resumed on success). No schema, API, or config changes.

## Model Used

Claude (Anthropic), Opus 4.x family — used for the implementation, tests, and this PR description. Commit trailers attribute `Co-Authored-By: Paperclip <noreply@paperclip.ing>` per project convention.

## Checklist

- [x] I have searched the open PR list for similar PRs; none currently open address the board-UI 404 storm.
- [x] I have included a thinking path that traces from project context to this change.
- [x] I have specified the model used.

### DIFF
--- a/server/src/middleware/http-log-policy.test.ts
+++ b/server/src/middleware/http-log-policy.test.ts
@@ -1,0 +1,39 @@
+import { describe, expect, it } from "vitest";
+import { isExpectedClient404 } from "./http-log-policy.js";
+
+describe("isExpectedClient404", () => {
+  it("matches issue-mention prefetch 404s (with and without the /api prefix)", () => {
+    expect(isExpectedClient404("GET", "/api/issues/FR-005", 404)).toBe(true);
+    expect(isExpectedClient404("GET", "/issues/BPS-39", 404)).toBe(true);
+    expect(isExpectedClient404("GET", "/api/issues/BACKUPS-2026", 404)).toBe(true);
+  });
+
+  it("matches heartbeat-log poll 404s and strips the query string", () => {
+    expect(isExpectedClient404("GET", "/api/heartbeat-runs/ca5d23fc-c15b/log", 404)).toBe(true);
+    expect(
+      isExpectedClient404("GET", "/api/heartbeat-runs/ca5d23fc-c15b/log?offset=0&limitBytes=256000", 404),
+    ).toBe(true);
+  });
+
+  it("leaves every other 404 at warn (returns false)", () => {
+    // real sub-path, not the bare mention/log shape
+    expect(isExpectedClient404("GET", "/api/issues/FR-005/comments", 404)).toBe(false);
+    expect(isExpectedClient404("GET", "/api/heartbeat-runs/ca5d23fc-c15b", 404)).toBe(false);
+    // not a KEY-NUMBER mention shape
+    expect(isExpectedClient404("GET", "/api/issues/lowercase", 404)).toBe(false);
+    // unrelated route
+    expect(isExpectedClient404("GET", "/api/companies/abc/issues", 404)).toBe(false);
+  });
+
+  it("only downgrades GET 404s — other methods and statuses stay at warn", () => {
+    expect(isExpectedClient404("POST", "/api/issues/FR-005", 404)).toBe(false);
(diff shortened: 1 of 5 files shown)


techwizard-live (changes requested):
Blocking on the issue-detail negative cache scope. The server-side downgrade is narrow and the live-run polling suppression is reasonable, but `known404IssueRefs` is keyed only by `issueRef`. Paperclip issue identifiers are company-scoped, so after a user sees `ABC-1` 404 in company A, switching to company B can suppress a valid `ABC-1` prefetch for the rest of the session. Please scope the negative-cache key by company context (or otherwise prove refs are globally unique in this call path), update `isKnown404IssueRef`/prefetch/fetch callers accordingly, and add a regression that a 404 in one company does not suppress the same ref in another company.

PR #8270, opened 2026-06-18
[upstream sync] ci(e2e): skip Playwright CDN download and use system Chrome

videlboga:
## Thinking Path

> - CI health is critical because every PR runs the full `pr.yml` workflow, including an e2e Playwright suite
> - The e2e job repeatedly cancels/times out while trying to download and extract Chromium from the Playwright CDN
> - Previous PRs removed `--with-deps` and added timeouts, but `pnpm exec playwright install chromium` still attempts the CDN download on `ubuntu-24.04` runners
> - Playwright supports using the system Google Chrome shipped by GitHub-hosted runners via `PLAYWRIGHT_CHROMIUM_CHANNEL=chrome` + `PLAYWRIGHT_BROWSERS_PATH=0`, which avoids the CDN download entirely
> - However, when `playwright install chromium` is invoked without the channel env, it still downloads `Chrome for Testing` from the CDN; the current workflow calls it in a step that lacks the channel environment
> - This PR removes the unconditional `playwright install chromium` call from the install step and lets the existing `Run e2e tests` step use the system `chrome` channel with `PLAYWRIGHT_BROWSERS_PATH: 0`
> - It keeps a short, best-effort system dependency refresh (with `|| true` fallbacks) for self-hosted/custom images
> - The benefit is a reliable e2e CI that no longer hangs on the Playwright CDN download

## What Changed

- Removed the unconditional `pnpm exec playwright install chromium` call from `.github/workflows/pr.yml`.
- Renamed the step to `Install system Chrome for Playwright` and kept only a short, non-blocking refresh of system libraries (`timeout 60`, `|| true`, step `timeout-minutes: 2`).
- The existing `Run e2e tests` step already exports `PLAYWRIGHT_CHROMIUM_CHANNEL: chrome`, `PLAYWRIGHT_BROWSERS_PATH: 0`, and `PLAYWRIGHT_SKIP_VALIDATE_HOST_REQUIREMENTS: 1`, so Playwright will use the system Chrome without any CDN download.
- Kept the `actions/cache@v4` step for `~/.cache/ms-playwright` in case other workflows still need it.

## Verification

- Validated `.github/workflows/pr.yml` YAML syntax with `python3 -c "import yaml; yaml.safe_load(open(...))"`.
- The change will be exercised by the PR's own `e2e` job; a green e2e run proves the system Chrome path works.
- After merge, the next e2e run on `master` should not attempt a CDN download.

## Risks

- Low. GitHub-hosted `ubuntu-latest` runners include Google Chrome and the required system libraries.
- The system dependency refresh step is fully non-blocking (`|| true`) and capped at 2 minutes, so it cannot hang the job.
- If a self-hosted runner lacks Chrome, the e2e job will fail fast with a clear message instead of hanging for an hour.

## Model Used

- Kimi k2.7 Code via ollama-cloud provider, tool-use enabled, used for CI log analysis and patch authoring.

## Checklist

- [x] I have validated YAML syntax locally
- [ ] I have added or updated tests where applicable (N/A — CI-only change)
- [ ] If this change affects the UI, I have included before/after screenshots (N/A)
- [ ] I have updated relevant documentation to reflect my changes (N/A — CI-only change)


videlboga:
Created against wrong target repository — the issue lives in videlboga/paperclip-ru fork, not upstream paperclipai/paperclip.

PR #8287, opened 2026-06-18
fix(execution): honor realized workspace cwd on remote leases + tolerate benign tar exit 1

stubbi:
## Problem
When running an agent on a **remote** execution environment (e.g. a kubernetes sandbox provider), two issues caused every run to fail at workspace restore:

1. **Realized workspace cwd is dropped.** A provider's `onEnvironmentRealizeWorkspace` returns the authoritative in-environment cwd (e.g. `/workspace`), but `acquireLease` does not record it and `realizeForRun` discarded it. `resolveEnvironmentExecutionTarget` then found no `remoteCwd` on the lease and fell back to `DEFAULT_SANDBOX_REMOTE_CWD` (`/tmp`), so the adapter synced/tarred the workspace under `/tmp` instead of the realized directory.
2. **`tar` exit code 1 treated as fatal.** `restoreWorkspace`'s `tar` returns exit 1 for the benign "file changed as we read it" warning, which is guaranteed when archiving a live workspace whose files mutate mid-archive. This was treated as a hard failure.

## Fix
- `environment-run-orchestrator.ts`: thread the realized cwd onto the lease metadata as `remoteCwd` (only when a non-`local` driver reports a non-empty cwd, and only if it differs), so the execution target honors it. Generic across all remote drivers.
- `sandbox-managed-runtime.ts`: a portable `tolerantTar` shell wrapper (`{ tar…; rc=$?; [ "$rc" -le 1 ]; }`, no GNU-only `--warning=` flag so busybox tar works) + a local `execTar` exit-1 swallow, applied to the create/extract steps. Exit code 2 remains fatal.

Unit tests added for both behaviors. `tsc --noEmit` clean.

### DIFF
--- a/packages/adapter-utils/src/sandbox-managed-runtime.test.ts
+++ b/packages/adapter-utils/src/sandbox-managed-runtime.test.ts
@@ -1,4 +1,4 @@
-import { lstat, mkdir, mkdtemp, readFile, readdir, rm, symlink, writeFile } from "node:fs/promises";
+import { chmod, lstat, mkdir, mkdtemp, readFile, readdir, rm, symlink, writeFile } from "node:fs/promises";
 import os from "node:os";
 import path from "node:path";
 import { execFile as execFileCallback } from "node:child_process";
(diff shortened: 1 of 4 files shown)


stubbi:

PR #8306, opened 2026-06-19
feat(opencode-local): probe-resilience — persistent disk cache + non-fatal startup

dmndbrp-oss:
## Summary

- **Root cause**: The `testEnvironment` startup probe called `discoverOpenCodeModels` which throws on timeout → run reports `adapter_failed` and dies before task logic (root cause verified in SAG-4315).
- **Fix**: Add `discoverOpenCodeModelsResilient` that falls back to a persistent on-disk cache (or the configured model) instead of throwing; wire it into the startup probe.
- This is the durable fix; SAG-4318 (orphan-opencode reaper, PR #8305) covers instances in the interim.

### `packages/adapters/opencode-local/src/server/models.ts`

- Add persistent on-disk discovery cache under `os.tmpdir()` (or `PAPERCLIP_OPENCODE_MODELS_CACHE_DIR` for tests), keyed by `discoveryCacheKey(command, cwd, env)`.
- Written on every successful live probe. Read on any probe timeout or failure.
- Promoted into the 60s in-memory cache after a disk hit to avoid repeated disk reads.
- **`discoverOpenCodeModelsResilient(input)`** — tries live probe → disk cache (even stale) → configured model fallback. Only throws when no fallback exists (unchanged contract for unconfigured envs).
- **`ensureOpenCodeModelConfiguredAndAvailable`** now uses `discoverOpenCodeModelsResilient`; a configured model is trusted as last resort on probe failure with no cache.
- New test helpers: `resetOpenCodeModelsMemoryCacheForTests()` (memory-only, for restart simulation), `populateOpenCodeModelsDiskCacheForTests()`.

### `packages/adapters/opencode-local/src/server/test.ts`

- Both `discoverOpenCodeModels` call sites → `discoverOpenCodeModelsResilient`.
- `source === "disk_cache"` or `source === "configured_model"` → `opencode_models_discovery_degraded` warn check (not error).

### `packages/adapters/opencode-local/src/server/models.test.ts` / `test.probe-resilience.test.ts` (new)

All 5 spec-required tests pass:
- probe fail + warm disk cache → returns cached list, no throw
- probe fail + no cache + configured model → `[{id: model}]`, no throw
- probe fail + no cache + no fallback → still throws (unchanged)
- disk cache survives simulated process restart (`resetOpenCodeModelsMemoryCacheForTests`)
- `testEnvironment` degraded → warn check, status not "fail"

## Test plan

- [x] `npx vitest run packages/adapters/opencode-local/` — 41/41 pass (8 test files)
- [x] `npx tsc --project packages/adapters/opencode-local/tsconfig.json --noEmit` — clean

### DIFF
--- a/infra/opencode/reap-orphan-opencode.sh
+++ b/infra/opencode/reap-orphan-opencode.sh
@@ -1,0 +1,120 @@
+#!/usr/bin/env bash
+# reap-orphan-opencode.sh — kill reparented opencode (models|run) processes that
+# outlived their spawner.  Safe to run as a preflight step; always exits 0.
+#
+# Env knobs:
+#   REAP_AGE_SEC   minimum elapsed seconds before a candidate is eligible (default 300)
+#   REAP_DRY_RUN   set to 1 to log candidates without killing (default 0)
+#   REAP_LOG       path of the append-only logfile (default /tmp/reap-orphan-opencode.log)
+set -euo pipefail
+
+REAP_AGE_SEC="${REAP_AGE_SEC:-300}"
+REAP_DRY_RUN="${REAP_DRY_RUN:-0}"
+REAP_LOG="${REAP_LOG:-/tmp/reap-orphan-opencode.log}"
+GRACE_SEC=3
+
+log() {
+  local msg="$(date -u +%Y-%m-%dT%H:%M:%SZ) $*"
+  echo "$msg" | tee -a "$REAP_LOG"
+}
+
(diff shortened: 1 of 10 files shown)


dmndbrp-oss:
Superseded by #8308 — clean re-cut scoped to SAG-4319 only (4 opencode-local files, branched fresh from upstream master). SAG-4314 and SAG-4318 commits are NOT included.

PR #7530, opened 2026-06-04
Add low-trust review containment

cryppadotta:
## Thinking Path

> - Paperclip is a control plane for AI-agent companies, so execution policy and trust boundaries are part of the product's safety contract.
> - Low-trust review work needs narrower authority than normal same-company agents because hostile PRs, comments, attachments, and generated output can carry prompt-injection payloads.
> - The current V1 shape gives trusted workers broad company context, which is useful for normal execution but too permissive for a reviewer assigned to hostile content.
> - This branch adds a `low_trust_review` preset, source-trust tagging, route-level containment, and quarantine handling so low-trust output does not automatically flow into higher-trust wake context.
> - The branch has been rebased onto current `origin/master`, and the low-trust migration was renumbered to `0097_low_trust_source_trust.sql` to avoid collisions with existing `0091` through `0096` migrations.
> - Greptile feedback was addressed by tightening low-trust detection, preserving project-level trust policy checks, fixing issue-kind promotion lookup, removing duplicate post-lease isolation assertion, documenting fail-closed source-trust behavior, bounding ancestry checks, enforcing runtime issue context for CEOs, awaiting accepted-plan monitor authorization, and making low-trust issue source-trust tagging atomic.
> - The benefit is a first production slice of deny-by-default review containment with regression coverage for the main control-plane pivot surfaces.

Fixes #7531.

## What Changed

- Added shared trust-policy types and validators, plus database/source-trust fields for issues, comments, documents, and work products.
- Implemented server enforcement for low-trust issue scope, agent self-view redaction, secret/plugin/runtime denial paths, promotion checks, and quarantined continuation/wake context.
- Added focused low-trust regression tests for resolver behavior, source trust, route authorization, heartbeat preflight ordering, runtime containment, and quarantine redaction.
- Added board UI affordances for selecting/reviewing the low-trust preset and surfacing source-trust badges in relevant issue views.
- Added `doc/LOW-TRUST-PRESETS.md`, updated `doc/SPEC-implementation.md`, and committed the low-trust review contract plan under `doc/plans/`.
- Rebasing note: the original `0097_low_trust_source_trust.sql` migration was renamed to `0097_low_trust_source_trust.sql`; the SQL uses `ADD COLUMN IF NOT EXISTS` so users who already applied the old-numbered migration are not broken by the renumbered migration.

## Verification

- Rebased branch onto current `origin/master` and force-pushed with lease to `origin/PAP-10211-low-trust-agent` at head `2719f31e3`.
[...]

PR #8330, opened 2026-06-19
fix(server): redact adapterConfig.env for cross-actor agent reads (BASA-29461)

HelloWolrs:
## Thinking Path

> - The control plane exposes per-agent adapter configuration (`adapterConfig.env`) so agents can boot with the credentials they need to call upstream LLMs, source forges, cloud APIs, etc.
> - `server/src/redaction.ts::redactEventPayload` already exists and is correctly wired into `GET /api/agents/:id/configuration` and the config-revision routes so privileged-but-non-self callers see env key names with values replaced by `***REDACTED***`
> - The two sibling read routes — `GET /api/agents/:id` and `GET /api/companies/:companyId/agents` — bypass that helper and spread the raw agent object, leaking plaintext credentials across the actor boundary to board users and any agent holding `agents:create`
> - This pull request closes that leak by running the existing redactor over `adapterConfig` and `runtimeConfig` for the privileged-non-self branch in both routes, while keeping the self path plaintext (so the adapter boot path still works) and the restricted path empty (unchanged)
> - The benefit is least-privilege secret handling on the highest-traffic agent-read surfaces without disturbing adapter boot or the existing restricted view

No public GitHub issue exists for this — describing inline following the bug template.

**What is the bug?** `GET /api/agents/:id` and `GET /api/companies/:companyId/agents` return raw `adapterConfig.env` values to privileged non-self callers. Any board user — or any agent that holds the `agents:create` permission — can read every other agent's plaintext env values (API keys, tokens, secrets) by hitting either route. The sibling configuration routes already redact correctly; only these two read paths leak.

**Steps to reproduce:**
1. As a peer agent B (different from target A) with `agents:create` granted: `curl -H 'Authorization: Bearer <B-key>' /api/agents/<A-id>` → response includes `adapterConfig.env.OPENAI_API_KEY: "sk-..."` (plaintext).
2. As a board user with company access: `curl /api/companies/<C>/agents` → every element's `adapterConfig.env` includes plaintext values for every key.
3. As the same callers, hit `GET /api/agents/<A-id>/configuration` — env values come back `***REDACTED***`. The leak is path-specific.

**Expected:** env key names preserved, plaintext values replaced with `***REDACTED***` for the privileged-non-self path (matching `/agents/:id/configuration`). Self stays plaintext (adapter boot still needs the raw values). Restricted path stays empty.

**Threat model:** an attacker who compromises any board session or any agent with `agents:create` can exfiltrate every credential in the company by walking the two routes — no audit signal, no per-key access trace.

## What Changed

- New `applyRedactedEnvelope(agent)` helper in `server/src/routes/agents.ts` that runs `redactEventPayload` over both `adapterConfig` and `runtimeConfig`. Falls back to `{}` if the redactor returns null.
- `buildAgentDetail` accepts a new `redacted?: boolean` option. Branch order inside is restricted → redacted → plaintext; mutually exclusive, restricted wins on conflict.
- `GET /api/agents/:id`: when `!isSelf && canReadSensitiveDetail`, return `buildAgentDetail(agent, { redacted: true })`. Self path and restricted path unchanged.
- `GET /api/companies/:companyId/agents`: when `canReadConfigs`, map each row through `applyRedactedEnvelope`. Restricted path unchanged.
- Per-route low-trust carve-out: when the caller is a board reviewer AND the target agent's resolved trust preset is `low_trust_review`, plaintext is preserved on `/agents/:id` so existing low-trust containment review (added in #7530) keeps working. The carve-out is bounded to that exact combination.
- 5 regression tests added in `server/src/__tests__/agent-cross-tenant-authz-routes.test.ts` that encode the vulnerability — they fail on master before this PR and pass after.
- A small `buildStubDb` drizzle-shaped helper (chainable `.from/.leftJoin/.where/.orderBy` with thenable `.then` returning `[]`) so the self-path's `resolveAgentSelfTrustPreset → db.select(...)` chain doesn't crash with a TypeError in test, where the route mock previously passed `{} as any` for the database.

No schema, migration, or API-contract changes. Env key names are preserved in every redacted response so the management UI keeps rendering which keys exist.

## Verification

```
cd server
npx vitest run src/__tests__/agent-cross-tenant-authz-routes.test.ts
```

Local result: **10/10 passing** (5 pre-existing + 5 new):
- `GET /api/agents/:id` as peer agent with `agents:create` → env key names present, plaintext values replaced with `***REDACTED***` (asserts the leak does not return).
- `GET /api/agents/:id` as self → plaintext present (asserts adapter boot path still works).
- `GET /api/agents/:id` as peer agent without `agents:create` → `adapterConfig: {}` (asserts restricted path unchanged).
- `GET /api/companies/:companyId/agents` as board user → every element has env values redacted, key names preserved (asserts list-route fix).
- Negative: probe a seed agent with `{type:"secret_ref"}` env binding → envelope passes through unchanged (asserts the safe shape is not collateral-damaged).

Manual spot check: `GET /api/agents/<peer-id>` as the CEO seat returns `{type:"plain", value:"***REDACTED***"}` for sensitive env keys, with key names visible.

## Risks

Low risk. The change is additive — a new branch in `buildAgentDetail` plus a single helper. The redactor (`redactEventPayload`) is the exact function already used on the configuration route, so behavior is consistent across all three read surfaces.

Potential concerns and mitigations:
- **Adapter boot regression** — self path is explicitly preserved as plaintext; covered by the self-path regression test.
- **Restricted-view regression** — restricted branch ordering is unchanged and explicitly evaluated before `redacted`; covered by the restricted-view regression test.
- **Low-trust containment regression (#7530)** — the per-route carve-out preserves plaintext on `/agents/:id` for board reviewer + `low_trust_review` target, exactly matching the surface #7530 added the carve-out for. Existing `agent-permissions-routes.test.ts:421 "keeps board agent detail unredacted for low-trust agents"` continues to pass.
- **List-route shape change** — every element is run through the same helper, no shape change (only value redaction).
- **Activity log telemetry** — `agent.*` activity-log entries are written via the redacted pipeline already; this change does not affect that path.

## Model Used

Claude — `claude-opus-4-7` (Opus 4.7 family, 200k context, extended thinking off for this PR, tool use enabled for repo edits and CI inspection).

## Checklist

- [ ] My branch name describes the change (e.g. `docs/...`, `fix/...`) and contains no internal Paperclip ticket id or instance-derived details — the branch was created against an internal tracker before this PR adopted the public template; leaving as-is to preserve the head sha that already received per-SHA security review
- [x] I have updated relevant documentation to reflect my changes — N/A (no docs reference these routes)

### DIFF
--- a/server/src/__tests__/agent-cross-tenant-authz-routes.test.ts
+++ b/server/src/__tests__/agent-cross-tenant-authz-routes.test.ts
@@ -48,6 +48,8 @@ let currentAccessCanUser = false;
 
 const mockAgentService = vi.hoisted(() => ({
   getById: vi.fn(),
+  list: vi.fn(),
+  getChainOfCommand: vi.fn(),
   pause: vi.fn(),
   resume: vi.fn(),
   terminate: vi.fn(),
@@ -213,6 +215,24 @@ async function loadRouteModules() {
   return routeModules;
 }
 
+function buildStubDb(): Record<string, unknown> {
+  const chain = {
+    from() { return chain; },
+    leftJoin() { return chain; },
+    innerJoin() { return chain; },
+    where() { return chain; },
+    orderBy() { return chain; },
+    limit() { return chain; },
+    then(onfulfilled: (rows: unknown[]) => unknown) { return Promise.resolve([]).then(onfulfilled); },
+  };
+  return {
+    select: () => chain,
+    insert: () => chain,
+    update: () => chain,
+    delete: () => chain,
+  };
+}
+
 async function createApp(actor: Record<string, unknown>) {
   const [{ errorHandler }, { agentRoutes }] = await loadRouteModules();
   const app = express();
@@ -224,7 +244,7 @@ async function createApp(actor: Record<string, unknown>) {
     };
     next();
   });
-  app.use("/api", agentRoutes({} as any));
+  app.use("/api", agentRoutes(buildStubDb() as any));
   app.use(errorHandler);
   return app;
 }
@@ -274,6 +294,8 @@ function resetMockDefaults() {
   currentKeyAgentId = agentId;
   currentAccessCanUser = false;
   mockAgentService.getById.mockImplementation(async () => ({ ...baseAgent }));
+  mockAgentService.list.mockImplementation(async () => [{ ...baseAgent }]);
+  mockAgentService.getChainOfCommand.mockImplementation(async () => []);
   mockAgentService.pause.mockImplementation(async () => ({ ...baseAgent }));
   mockAgentService.resume.mockImplementation(async () => ({ ...baseAgent }));
   mockAgentService.terminate.mockImplementation(async () => ({ ...baseAgent }));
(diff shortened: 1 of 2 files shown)


HelloWolrs:
/approve

Pinned to head SHA `cfa925a5595dd4c6eaaf0ca084ac508fd7b9ad9a`.

Security APPROVE recorded on BASA-29475 by Priya Shah — see https://app.paperclip.example/BASA/issues/BASA-29475 for the verdict + carve-out gate analysis. This comment is the mechanical merge-driver signal.

PR #8352, opened 2026-06-19
Reconcile orphaned running agent status on read (ATLA-1225)

joe-nano:
## Thinking Path

> - Paperclip orchestrates AI agents for zero-human companies, and each agent carries a lifecycle `status` (`idle` / `running` / ...) used for scheduling and observability.
> - An agent is set `running` only while a heartbeat run backs the flag; the run-start path writes the `heartbeat_runs` row to `queued`/`running` *before* it flips `agents.status` to `running`.
> - When a backing run's process dies without finalizing the agent (box killed, crash), the flag is orphaned: `status === "running"` with no `queued`/`running` run anywhere in the company. Such agents look busy forever and are skipped by scheduling.
> - The first fix shipped an agent-driven reaper routine, but cross-agent `PATCH /api/agents/{id} {"status":"idle"}` is CEO-or-creator-only (403), so a CTO-run reaper can detect orphans but not reset them.
> - This pull request implements the durable fallback: reconcile the orphaned flag in the control-plane read path itself, so any agent read self-heals it.
> - The benefit is that no reaper routine and no privileged cross-agent PATCH are needed — the lifecycle layer keeps itself consistent, with an audit trail for every reconcile.

No public GitHub issue exists for this; per CONTRIBUTING.md path (B), the underlying problem is described here following `bug_report.yml`:

**What happened?** An agent's lifecycle `status` can be left stuck at `"running"` when the process backing its heartbeat run dies without finalizing (host killed, crash, OOM). The agent then looks busy forever: `GET /api/agents/{id}` and the company agent list keep reporting `running`, and scheduling skips the agent.

**Expected behavior:** An agent whose `running` flag has no backing `queued`/`running` heartbeat run should return to `idle`, with an audit record of the correction.

**Steps to reproduce:** 1) Start an agent run so `agents.status = 'running'`. 2) Kill the runner process so the run never finalizes (simulate a crashed box; the `heartbeat_runs` row is finalized by the reaper but the agent status flip is lost). 3) Observe the agent stays `running` indefinitely on every read; no code path ever resets it.

**Paperclip version or commit:** reproduced on `master` (also present in current `master` as of 05973b20).

**Deployment mode:** local_trusted (any mode is affected — the gap is in `agentService`, not deployment-specific).

**Agent adapter(s) involved:** any (adapter-independent; Claude Code in the reproduction).

**Why not an external-repair routine instead?** A privileged reaper agent needs cross-agent `PATCH /api/agents/{id}` (CEO-or-creator-only), which is a permission widening; self-healing on read keeps the lifecycle layer consistent without new privileges.

Duplicate search: no existing PR implements agent-status reconciliation. Closely related but distinct: #8535 (cleans up stale runtime *processes* on startup; does not touch the `agents.status` flag or the read path).

## What Changed

- `server/src/services/agents.ts`: added on-read reconciliation in the agent read/serialize path. `agentService.getById` and `agentService.list` now heal an orphaned `running` flag to `idle`, so `GET /api/agents/me`, `GET /api/agents/{id}`, and `GET /api/companies/{id}/agents` all self-heal.
- Reconciliation predicate: `status === "running"` AND zero backing runs (`queued`/`running`) for the agent company-wide AND the orphan has persisted past one reconciliation window since the flag was written (`updatedAt`), or the agent has never finished a heartbeat (`lastHeartbeatAt == null`).
- The heal `UPDATE` carries an atomic `NOT EXISTS (... queued/running runs ...)` guard, so a run that starts concurrently between the count and the write is never stomped.
- Emits one `agent.status.reconciled` audit row per reconcile via `logActivity`, capturing prior `status` / `currentRunId` (always null — runs live in `heartbeat_runs`) / `lastHeartbeatAt` and the timestamp.
- Exposed `reconcileOrphanedRunningStatus` on the service for direct testing.
- `server/src/__tests__/agent-status-reconciliation.test.ts`: new tests (heal + audit, backing-run guard, in-window race guard, first-run orphan, list path, and a regression for the two historical orphans).

## Verification

- `pnpm exec vitest run server/src/__tests__/agent-status-reconciliation.test.ts` → 7/7 pass.
- Regression of existing read-path suites is green: `agent-permissions-routes`, `agent-live-run-routes`, `monthly-spend-service`, `agent-shortname-collision` (69 tests total, all pass).
- `tsc --noEmit` for `@paperclipai/server` reports 0 errors.
- Manual reasoning for the race guard: run-start writes the `heartbeat_runs` row to `queued`/`running` before flipping `agents.status`, so a legitimately running or about-to-start agent always has a backing run and is never reconciled; the `NOT EXISTS` guard closes the count-then-write window.

## Risks

- Low risk. Behavior only changes for agents pinned `running` with genuinely no backing run; healthy agents short-circuit before any extra query (non-`running` agents do zero extra work).
- A write now occurs inside a read path, but only for true orphans (rare), and it is idempotent and atomically guarded.
- No schema/migration changes. No change to the run-level reaper (`reapOrphanedRuns`), which still handles process cleanup.

## Model Used

- Claude Opus 4.8 (Anthropic), model ID `claude-opus-4-8[1m]`, 1M-token context, extended reasoning + tool use (Claude Code agent harness).

## Checklist

- [ ] My branch name describes the change — note: the current branch name predates this checklist item; renaming the head branch would close this PR, so it stays. The branch name describes the change itself.
- [x] I have run tests locally and they pass (7/7 new tests; re-verified rebased onto `master` @ 05973b20)
- [x] I have updated relevant documentation to reflect my changes (N/A — no doc surface for this internal service behavior)
- [ ] All Paperclip CI gates are green (blocked: base branch is behind and CI does not run on a conflicting PR; rebase is ready and verified locally, pending fork sync)
- [ ] Greptile is 5/5 with no open P2s, recommendations, or follow-ups (summary posted; will address any findings)


joe-nano:
**Note on base / mergeability.** This branch is based on the fork's `master` (which is behind upstream), and upstream `master` has since landed a ~313-line refactor of `server/src/services/agents.ts` (`normalizeAgentRows` / `orgChainHealth` eligibility). GitHub therefore shows this PR as conflicting.

The conflict is mechanical. I have already produced and verified the rebased-onto-current-master version of this exact change:
- merges cleanly into current `master` (2 files, +302/-2, no conflicts),
- `tsc --noEmit` for `@paperclipai/server` clean (only pre-existing unbuilt-adapter resolution noise),
- new test `agent-status-reconciliation.test.ts` 7/7 pass, and the existing read-path suites (`agent-permissions-routes`, `agent-live-run-routes`, `monthly-spend-service`, `agent-shortname-collision`) stay green.

I could not push the up-to-date branch to the fork because syncing the fork to upstream is rejected with `refusing to allow an OAuth App to ... workflow ... without 'workflow' scope` (the run's `gh` OAuth token lacks `workflow` scope; no SSH key / PAT is available). Once the fork can be synced (a `workflow`-scoped credential), I will rebase onto current `master` and force-update this branch — no other changes. The canonical diff is also attached to the tracking issue ATLA-1225.

joe-nano:
Rebase re-verified against current `master` (05973b20, 2026-07-10): the change re-applies with one mechanical conflict in the `create` signature (upstream added `CreateAgentOptions`), all 7 new tests pass, and `tsc` reports no errors in the touched files. The push of the rebased head is still rejected by GitHub because this repo's history since the fork point touches `.github/workflows/agent-runtime-images.yml` and the authoring token lacks the `workflow` OAuth scope — the fork owner needs to click **Sync fork** (or push with a workflow-scoped credential), after which I will update this branch immediately. PR description updated per the checklist bot: inline issue description added and dedup search done (closest related PR is #8535, which cleans stale runtime processes, not the agent `status` flag).

joe-nano:
Rebased onto current `master` (`821573ed`) — **this PR is now mergeable**; it had been conflicting since master's org-chain-health changes landed in `server/src/services/agents.ts`.

The conflict was add/add: master added `toEligibilityAgent` / `normalizeAgentRows` / `listCompanyAgentRows` / `requireGetById` in the same region where this PR adds the reconciliation helpers. Resolution keeps both:

- `getById` — reconciles the hydrated row, then normalizes with `companyRows`
- `list` — reconciles hydrated rows, then `normalizeAgentRows(reconciled, allCompanyRows)`
- kept master's `create(…, options)` signature and built-in-metadata guard

The company rows used for org-chain-health are deliberately left unreconciled: `getAgentWorkEligibility` only reads `terminated`/`paused`/`pending_approval`, so running-vs-idle cannot affect it, and reconciling them would mean one write per agent on every list/get.

Net diff is unchanged in shape: **2 files, +301/−2**.

Verified locally on the merged tree (after `pnpm install --frozen-lockfile` to pick up the zod v4 bump):
- `server/src/__tests__/agent-status-reconciliation.test.ts` — 7/7 pass
- full `server/src/__tests__/agent*` sweep — **23 files, 306 tests, all pass**
- no typecheck errors in `server/src/services/agents.ts`

nickallevato:
We run this PR as a local patch on a production instance (together with fixes for the two Greptile P2s: one batched backing-run query for list reads, and one shared timestamp for the healed row and its audit record). Here is data from about five weeks, in case it helps review.

**The heal works, but most heals are false positives.** There were 83 `agent.status.reconciled` events. 73 of them happened less than 5 s after the run's terminal write. A read raced the normal `finalizeAgentStatus`, which runs later in `executeRun`. The cause: the grace window is measured from `agents.updatedAt`. `executeRun` sets that value when the run *starts*. So for any run longer than 60 s, a read in the short gap between the terminal run write and `finalizeAgentStatus` heals the agent. The outcome is the same (`idle`), but the audit log fills with noise.

Suggestion: base the grace on the agent's latest run instead. For example, skip the heal if a run for this agent has `finished_at > now() - 60s`.

Two smaller points:
- The heal publishes no `agent.status` live event, so open UIs do not update until a refetch.
- The heal writes during GET requests.

**The real source of orphans.** The other heals (8 of 8 genuine late heals) all came from one path. The recovery backstop ends a run that is still alive, and then the owner finalizer loses its compare-and-set and returns before `finalizeAgentStatus`. Details and a proposed fix are in #13631. With that fixed, this PR stays useful as defense in depth.

PR #8358, opened 2026-06-20
feat(issues): production health gate before closing incidents

64johnlee:
## Thinking Path

> - Issues are how teams track production incidents; closing one (`status → done`) is the signal that the incident is resolved
> - Closure today is approval-based — a human marks it done — with nothing checking that production actually recovered
> - This gap bit us directly: a P0 was approved-closed **three times** while production was still down (deploys `BLOCKED` by a git-author gate, then sitting behind a 401 protection wall)
> - This PR adds an automated health gate in the `PATCH /issues/:id` path: a production-incident issue cannot transition into `done` while its configured live URLs are unhealthy
> - The benefit is that "resolved" means production returns HTTP 200, not just that someone clicked close — and the gate is fail-open (a pure no-op unless explicitly configured) so it can never block unrelated work

No existing public issue — describing inline (bug-report shape):

- **What happens:** A production-incident issue can be moved to `done` purely on approval, with no check against live production. We closed a real P0 three times while production was still fully down.
- **Expected:** Closing a production incident should require that the configured production URLs are actually healthy (HTTP 200, optional body marker) at close time.
- **Impact:** Incidents get marked resolved while customers are still down, eroding the meaning of "done" and delaying real remediation.
- **Scope:** Backend only — the issue-update route and a new service. No schema or API-contract changes.

## What Changed

- **New service `server/src/services/production-health-gate.ts`** — `assertProductionHealthyForClosure()` classifies an issue as a production incident (title/label markers), probes each configured target with retry + timeout, and throws `ProductionHealthGateError` if any target isn't healthy.
- **`PATCH /issues/:id` (`server/src/routes/issues.ts`)** — runs the gate before the DB write; on failure returns **409** `{ code: "production_health_gate_failed", unhealthy: [...] }` instead of closing.
- **Probe latency bounded (review fix)** — retry budget cut from a 44s worst case to **11s** (`RETRIES 3→2`, `TIMEOUT 12s→5s`, `DELAY 4s→1s`) so an activated gate returns its 409 well inside the ~30s client/proxy timeout instead of surfacing as a network timeout.
- **Redirect handling fixed (review fix)** — `fetch` now uses `redirect: "follow"` instead of `"manual"`. With `"manual"`, undici returns an opaque response whose status is always `0`, which would falsely fail any production URL that redirects (http→https, trailing-slash, etc.) and permanently block closure on a healthy endpoint.
- **Tests** — 11 existing vitest cases plus a new case that runs the real probe with `fetch` stubbed, asserting it follows redirects and treats a settled 200 as healthy (covers the previously untested `fetch` path).

## Verification

```bash
pnpm --filter @paperclipai/server vitest run production-health-gate
# -> Test Files 1 passed | Tests 12 passed (11 existing + 1 new probe/redirect case)
```

- Gate **off** (no `PRODUCTION_HEALTH_TARGETS`): ordinary and incident issues close normally — verified by the unconfigured/fail-open cases.
- Gate **on**, all targets 200: incident closes.
- Gate **on**, a target down (e.g. 401): close is blocked with 409 and the unhealthy targets are listed.
- Non-incident issues and non-closing transitions skip the gate entirely.

## Risks

- **Low / opt-in.** The gate is a complete no-op unless `PRODUCTION_HEALTH_TARGETS` is set, the issue matches incident markers, and the transition is into `done`, so unrelated closures are unaffected.
- **In-band latency:** the probe runs inside the PATCH handler. Worst case is now 11s (was 44s), comfortably under typical 30s proxy timeouts.
- **Fail-closed only for genuine incidents:** if production is actually unhealthy, a configured incident cannot be closed — which is the intended behavior.
- No migrations, no breaking API changes; the only new response is the additive 409 on a blocked incident closure.

## Model Used

Claude (Anthropic) — **Opus 4.8**, 1M context window, extended thinking + tool use. Used to assist the implementation and to produce this review-response revision (probe-latency and redirect fixes, added test).

## Checklist

- [x] I have searched GitHub for duplicate or related PRs and linked them above (none — closest is #5447, unrelated incident-cooldown logic)
- [ ] If this change affects the UI, I have included before/after screenshots (N/A — backend only)
- [ ] All Paperclip CI gates are green (will confirm after push)
- [ ] Greptile is 5/5 with no open P2s, recommendations, or follow-ups (re-running after push)

### DIFF
--- a/server/src/__tests__/production-health-gate.test.ts
+++ b/server/src/__tests__/production-health-gate.test.ts
@@ -1,4 +1,4 @@
-import { describe, expect, it } from "vitest";
+import { afterEach, describe, expect, it, vi } from "vitest";
 import {
   assertProductionHealthyForClosure,
   getProductionHealthTargets,
@@ -124,5 +124,37 @@ describe("assertProductionHealthyForClosure", () => {
         probeFn: oneDown,
       }),
     ).resolves.toBeUndefined();
+  });
+});
+
+describe("probe (real fetch path)", () => {
+  afterEach(() => {
+    vi.unstubAllGlobals();
+  });
+
+  it("follows redirects and treats a settled 200 as healthy", async () => {
+    const inits: RequestInit[] = [];
+    vi.stubGlobal(
+      "fetch",
+      vi.fn((_url: string, init: RequestInit) => {
+        inits.push(init);
+        return Promise.resolve(new Response("ok", { status: 200 }));
+      }),
+    );
+
+    // No probeFn -> the real probe() runs, exercising the fetch call.
+    await expect(
+      assertProductionHealthyForClosure({
+        issue: incidentIssue,
+        existingStatus: "in_progress",
+        requestedStatus: "done",
+        env: envWith(),
+      }),
+    ).resolves.toBeUndefined();
+
+    // One probe per configured target, each following redirects (not "manual",
+    // which would yield an opaque status 0 and falsely fail a healthy endpoint).
+    expect(inits).toHaveLength(2);
+    expect(inits.every((init) => init.redirect === "follow")).toBe(true);
   });
 });
--- a/server/src/__tests__/production-health-gate.test.ts
+++ b/server/src/__tests__/production-health-gate.test.ts
@@ -1,0 +1,128 @@
+import { describe, expect, it } from "vitest";
+import {
+  assertProductionHealthyForClosure,
+  getProductionHealthTargets,
+  isProductionIncidentIssue,
+  ProductionHealthGateError,
+  type ProductionHealthResult,
+  type ProductionHealthTarget,
+} from "../services/production-health-gate.js";
+
+const TARGETS_JSON = JSON.stringify({
+  targets: [
+    { name: "pulse-web", url: "https://pulse-web-prod.example.app", expect: { status: 200 } },
+    { name: "pulse", url: "https://getpulse-app.example.app", expect: { status: 200 } },
+  ],
+});
+
+const envWith = (extra: Record<string, string> = {}): NodeJS.ProcessEnv => ({
+  PRODUCTION_HEALTH_TARGETS: TARGETS_JSON,
+  ...extra,
+});
+
+const incidentIssue = { title: "P0: Production Down — all routes 401", labels: [] };
+const ordinaryIssue = { title: "Tweak button padding on settings page", labels: ["ui"] };
+
+const allHealthy = (t: ProductionHealthTarget): Promise<ProductionHealthResult> =>
+  Promise.resolve({ name: t.name, url: t.url, healthy: true, status: 200 });
+const oneDown = (t: ProductionHealthTarget): Promise<ProductionHealthResult> =>
+  Promise.resolve(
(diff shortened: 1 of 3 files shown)


64johnlee:
Thanks both — all review findings addressed in `82a3ea5`:

**@superagent-security (P2 — 409 leaked internal target URLs):** Fixed. The `unhealthy` array in the 409 response now maps to `{ name, status, reason }` only, dropping `url` so internal production infrastructure is no longer disclosed to issue-closers. The `ProductionHealthGateError` message itself only ever contained `name=status`, no URLs.

**@greptile-apps:**
- *P1 — probe latency exceeding client/proxy timeout:* fixed in `b7a68d8` — retry budget cut from 44s to 11s worst case (`RETRIES` 3→2, `TIMEOUT` 12s→5s, `DELAY` 4s→1s), well under the ~30s cutoff.
- *P2 — `redirect: "manual"` opaque status 0:* fixed in `b7a68d8` — now `redirect: "follow"`, judging the final settled status so healthy endpoints that redirect (http→https, trailing-slash) aren't falsely blocked.
- *PR template:* the description now includes Thinking Path, Linked Issues (inline bug description), What Changed, Verification, Risks, Model Used, and the completed Checklist.

Gate tests pass locally (`12/12`, `pnpm --filter @paperclipai/server vitest run production-health-gate`). Re-running Greptile.

64johnlee:
Added a regression test for the security fix in `45a7371`. The 409 projection is now a small pure helper `toPublicUnhealthy()` in the gate service, directly covered by three cases: it never includes a target's internal `url`, returns only the unhealthy entries (`name/status/reason`), and is empty when all targets are healthy.

Gate suite now `15/15` (`pnpm --filter @paperclipai/server vitest run production-health-gate`).

64johnlee:
@greptile-apps re-review — all three findings from the prior review are now addressed on `45a7371`: P1 probe latency bounded to ~11s worst case (was 44s), P2 redirect:follow instead of manual, and P2 409 url-disclosure fixed via toPublicUnhealthy() (regression test added). Could you refresh the review summary against the latest commit? Thanks!

64johnlee:
Fixed in `153d248`. `isProductionIncidentIssue` now matches markers on **word boundaries** (`\bmarker\b`) instead of substring `includes`, with the user-configured markers regex-escaped. So "Update reproduction steps" and "p0lish the UI" no longer trip the gate, while "P0: production is down" and a `p0` label still match.

Added two regression cases (suite now `17/17`): one asserting markers embedded in another word don't match, one asserting whole-word markers next to punctuation still do.

Also confirming the earlier security item is resolved — superagent's latest pass is clean.

64johnlee:
@cryppadotta @devinfoley — this one's ready for merge when you have a moment. All 23 CI checks are green, mergeable/clean, and the bot review findings (Greptile P1/P2, superagent-security URL disclosure, word-boundary marker matching) have all been addressed in follow-up commits. It's a fail-open gate (no-op unless `PRODUCTION_HEALTH_TARGETS` is configured), so it carries no behavior change until opted in. Thanks!

PR #8393, opened 2026-06-20
fix(claude-local): tolerate sandboxes whose Claude CLI lacks --effort

devinfoley:
## Thinking Path

> - The claude-local adapter launches the Claude CLI inside execution environments (local, SSH, and ephemeral sandboxes such as Daytona)
> - Newer adapter code passes `--effort` to the CLI, but the Claude binary baked into some sandbox images is older and rejects it with `error: unknown option '--effort'`, so every run in those environments fails
> - This needs addressing because the failure is environment-dependent and silent from the operator's perspective — the run just dies with a CLI usage error
> - This pull request probes the in-sandbox CLI for `--effort` support once, caches the result per environment, and strips the flag (with a warning) when the CLI does not support it
> - The benefit is that sandboxes with older Claude CLIs keep working instead of failing, with negligible probe overhead because the capability check is cached and reused across ephemeral leases

No public GitHub issue exists. Describing the bug inline following the bug report template:

### What happened?

Runs using the `claude-local` adapter inside certain sandbox/execution environments fail with `error: unknown option '--effort'`. The adapter unconditionally appends `--effort` to the Claude CLI invocation, but the Claude CLI version present in some sandbox base images predates that flag, so the process exits with a usage error and the run dies.

### Expected behavior

The adapter should detect that the target environment's Claude CLI does not support `--effort` and degrade gracefully — drop the flag and emit a warning — rather than failing the run.

### Steps to reproduce

1. Configure an execution environment (e.g. a sandbox image) whose bundled Claude CLI is old enough to predate the `--effort` option.
2. Run any claude-local task that resolves to an effort level (so `--effort` is appended).
3. Observe the run fail immediately with `error: unknown option '--effort'`.

`master` at the time of this PR (branch forked from current `master`).

### Deployment mode

Self-hosted / local instance using execution environments (reproducible with ephemeral sandbox providers such as Daytona where `reuseLease: false`).

### Agent adapter(s) involved

Claude Code (`claude-local`).

## What Changed

- Add a CLI capability probe (`cli-capabilities.ts`) that runs the target Claude binary's `--help` inside the execution environment to detect `--effort` support.
- Strip `--effort` from the CLI args (emitting a warning) when the probe reports the flag is unsupported; keep it otherwise.
- Cache probe results keyed by `sandbox:providerKey:environmentId:command` (no lease id) so the probe is reused across ephemeral leases — important for `reuseLease: false` sandbox configs like Daytona, which would otherwise re-probe on every run.
- Conservative fallback: if the probe itself can't run/parse, assume the flag is supported (preserves prior behavior).

## Verification

- `node_modules/.bin/vitest run src/__tests__/claude-local-execute.test.ts src/__tests__/claude-local-adapter-environment.test.ts` → **2 files, 30 tests passed**.
- Regression test issues two `execute()` calls with distinct lease ids and asserts the in-sandbox `--help` probe runs exactly once (cache reuse across leases).
- Added tests covering: flag stripped when unsupported, flag retained when supported, warning emitted, and conservative fallback when the probe fails.

## Risks

Low risk. The change is additive and gated behind a probe with a conservative default (assume supported on probe failure), so existing environments that support `--effort` are unaffected. Worst case for an environment where the probe is unreliable is the prior behavior (flag passed through).

## Model Used

Claude Opus 4.8 (claude-opus-4-8), extended thinking, with tool use / code execution via Claude Code.

## Checklist

- [ ] I have updated relevant documentation to reflect my changes (N/A — no doc-facing behavior change)
- [ ] All Paperclip CI gates are green (pending CI)
- [ ] Greptile is 5/5 with no open P2s, recommendations, or follow-ups (pending review)

### DIFF
--- a/packages/adapters/claude-local/src/server/cli-capabilities.ts
+++ b/packages/adapters/claude-local/src/server/cli-capabilities.ts
@@ -78,9 +78,12 @@ export async function claudeCommandSupportsEffortFlag(input: {
   const cached = effortFlagSupportCache.get(key);
   if (cached) return cached;
 
-  const probe = probeClaudeCommandSupportsEffortFlag(input).catch((error) => {
+  // A thrown probe (e.g. sandbox connection error, ENOENT spawning the binary)
+  // must degrade to the conservative fallback rather than killing the run, so we
+  // resolve to null and drop the cache entry to retry on the next lease.
+  const probe = probeClaudeCommandSupportsEffortFlag(input).catch(() => {
     effortFlagSupportCache.delete(key);
-    throw error;
+    return null;
   });
   effortFlagSupportCache.set(key, probe);
   return probe;
--- a/packages/adapters/claude-local/src/server/cli-capabilities.ts
+++ b/packages/adapters/claude-local/src/server/cli-capabilities.ts
@@ -1,0 +1,40 @@
+import type { AdapterExecutionTarget } from "@paperclipai/adapter-utils/execution-target";
+import { runAdapterExecutionTargetProcess } from "@paperclipai/adapter-utils/execution-target";
+import path from "node:path";
+
+export function claudeCommandLooksLike(command: string, expected = "claude"): boolean {
+  const base = path.basename(command).toLowerCase();
+  return base === expected || base === `${expected}.cmd` || base === `${expected}.exe`;
+}
+
+export async function claudeCommandSupportsEffortFlag(input: {
+  runId: string;
+  command: string;
+  target: AdapterExecutionTarget | null | undefined;
+  cwd: string;
+  env: Record<string, string>;
+  timeoutSec: number;
+  graceSec: number;
+}): Promise<boolean | null> {
+  if (!claudeCommandLooksLike(input.command, "claude")) return null;
+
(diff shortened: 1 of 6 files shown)


devinfoley:
@greptileai addressed the three P1 exception-propagation findings in `94067dc51`.

Rather than `.catch(() => null)` at each call site, the fix is centralized in `cli-capabilities.ts`: when the `--help` probe rejects (sandbox connection error, `ENOENT`, etc.) `claudeCommandSupportsEffortFlag` now resolves to `null` and drops the cache entry. This makes the conservative fallback unconditional for *every* caller (`execute.ts` and `test.ts`) — the public function can no longer reject from the probe path — while preserving the retry-on-next-lease behaviour via cache deletion.

Added a regression test (`server/src/__tests__/claude-local-execute.test.ts`) that drives a throwing sandbox runner and asserts the probe resolves to `null` (not rejects) and re-probes on the next lease. Targeted suite: 31 passed.

techwizard-live (approved):
Reviewed the capability-probe path and tests. The change is sandbox-gated, falls back to prior behavior on unknown/probe failure, and has coverage for strip/pass-through/cache/retry cases. LGTM.

PR #8272, opened 2026-06-18
Guard codex_local agents from shared OpenAI key

MrBlackTongue:
## Thinking Path

> - Paperclip orchestrates AI agents for zero-human companies.
> - The `codex_local` adapter runs local Codex CLI processes and builds their environment from persisted agent config plus host process env.
> - A host-level `OPENAI_API_KEY` or shared Codex auth home can silently make new agents spend through shared credentials.
> - Existing agents can be repaired manually, but new and updated agents need a persistent guard at the agent configuration boundary.
> - This pull request isolates new and updated `codex_local` agents with per-agent `CODEX_HOME` and an empty `OPENAI_API_KEY` override.
> - The benefit is that future agent creation or adapter updates cannot silently fall back to shared OpenAI credentials.

## Linked Issues or Issue Description

Paperclip work item: [ZOL-5477](/ZOL/issues/ZOL-5477).

No matching GitHub issue exists, so the bug is described inline following `.github/ISSUE_TEMPLATE/bug_report.yml`.

**Pre-submission checklist**

- [x] I have searched existing open and closed issues and this is not a duplicate.
- [x] I am on the latest `master` commit for this PR branch.
- [x] I have confirmed the error originates in Paperclip's `codex_local` adapter configuration boundary, not in a provider outage.

**What happened?**

New or updated `codex_local` agents could inherit a host-level `OPENAI_API_KEY` or use a shared Codex home when their adapter config did not explicitly isolate those values. That made it possible for future agents or manual adapter edits to silently fall back to shared OpenAI credentials.

**Expected behavior**

Creating, hiring, or updating a `codex_local` agent should either persist isolated per-agent configuration or reject unsafe shared Codex home configuration with a clear 422 response. The guard must not print secret values.

**Steps to reproduce**

1. Create or update a `codex_local` agent without an explicit `adapterConfig.env.OPENAI_API_KEY` override.
2. Run it on a host where the Paperclip server process has `OPENAI_API_KEY` set.
3. Observe that the adapter process can inherit the host key unless Paperclip persists a blocking empty override.
4. Set `adapterConfig.env.CODEX_HOME` to a shared path such as `~/.codex` or the company-level `codex-home`.
5. Observe that the old code allowed the shared auth home instead of returning a validation error.

**Paperclip version or commit**

- Reproduced by inspection against `master` before this PR.

**Deployment mode**

- Local dev / self-hosted server with `codex_local` agents.

**Installation method**

- Built from source.

**Agent adapter(s) involved**

[...]

PR #8399, opened 2026-06-20
fix(codex-local): stop force-injecting empty OPENAI_API_KEY into agent config

steingran:
## Thinking Path

> - The `codex_local` adapter isolates each agent's Codex home and guards against host env leakage when spawning the Codex CLI
> - The isolation guard (introduced in #8272) force-injected an empty `OPENAI_API_KEY` slot into `adapterConfig.env` on every agent create/update, so operators could not delete the slot — a PATCH returned 200 but the slot reappeared on the next GET
> - The empty slot is redundant for subscription-authenticated agents: the Codex CLI (>= 0.122) ignores `OPENAI_API_KEY` and authenticates via its own `$CODEX_HOME/auth.json`, so the stored empty value has no runtime auth effect
> - This PR moves the host-leak protection from config-time injection to run-time stripping, keeping stored config clean while preserving the isolation guarantee
> - The benefit is that operators can manage `OPENAI_API_KEY` in agent config the same way they manage every other env var, and the host-leak guard still fires exactly when it is needed (at process spawn time)

No public GitHub issue exists for this. Bug report follows the `bug_report.yml` template:

**What happened?** When a `codex_local` agent is created or updated via `POST/PATCH /api/agents`, `applyCodexLocalIsolationGuard` in `server/src/routes/agents.ts` re-injects `OPENAI_API_KEY = ""` into `adapterConfig.env` whenever the key is absent. Deleting the slot via `PATCH` returns `200` with the cleaned env in the response body, but a subsequent `GET` shows the slot back. Only agents that already had `CODEX_HOME` in their stored env exhibit this; newer agents (e.g. Oracle, Trinity, Dozer, Ledger) created without the slot never receive it.

**Expected behavior.** The driver should not force an `OPENAI_API_KEY` slot into stored agent config. A host-level `OPENAI_API_KEY` should still be prevented from leaking into the spawned Codex process, but that protection belongs at run time, not in persisted config.

**Steps to reproduce.**
1. Create a `codex_local` agent with `adapterConfig: {}`.
2. `PATCH /api/agents/{id}` with `adapterConfig: { env: { GITHUB_TOKEN: "..." } }` (no `OPENAI_API_KEY`).
3. `GET /api/agents/{id}` — observe `adapterConfig.env.OPENAI_API_KEY === ""` is present even though it was omitted from the PATCH.
4. Repeat the PATCH omitting the key; the slot reappears on every GET.

**Paperclip version/commit.** `master` at `c074348` (package version `2026.618.0`).

**Deployment mode.** Authenticated local instance, embedded postgres.

## What Changed

- `server/src/routes/agents.ts` — `applyCodexLocalIsolationGuard` no longer force-injects `OPENAI_API_KEY = ""`. The `CODEX_HOME` isolation default and shared-home rejection are unchanged.
- `packages/adapters/codex-local/src/server/execute.ts` — added `stripInheritedOpenAiApiKey(effectiveEnv, envConfig)`, called after the `{ ...process.env, ...env }` merge. When `adapterConfig.env` does not explicitly contain `OPENAI_API_KEY`, any value inherited from the host/server process is deleted from the spawned Codex process env. Explicit config values (including an empty string) are preserved.
- `packages/adapters/codex-local/src/index.ts` and `doc/DEVELOPING.md` — updated the documented behavior to reflect run-time stripping instead of persisted empty overrides.
- `server/src/__tests__/agent-adapter-validation-routes.test.ts` — updated the two tests that asserted the empty slot injection to assert its absence, and added a test that an explicitly provided `OPENAI_API_KEY` is preserved.
- `packages/adapters/codex-local/src/server/execute.openai-key-isolation.test.ts` — new unit tests for `stripInheritedOpenAiApiKey` covering: host leak stripped when config omits the key; explicit non-empty value preserved; explicit empty string preserved; no-op when neither provides the key; empty `envConfig` strips; same object reference returned.

## Verification

```bash
pnpm vitest run packages/adapters/codex-local
pnpm vitest run server/src/__tests__/agent-adapter-validation-routes.test.ts
pnpm --filter @paperclipai/adapter-codex-local typecheck
```

All pass locally: 63 codex-local adapter tests, 10 agent-adapter-validation-routes tests (1 pre-existing skip), typecheck clean for both packages. Also ran the broader env-related suites (`codex-local-adapter-environment`, `opencode-local-adapter-environment`, `adapter-models`, `redaction`, `heartbeat-active-run-output-watchdog`, `heartbeat-process-recovery`) — 105 tests pass.

Manual repro check: after this change, a `PATCH /api/agents/{id}` that omits `OPENAI_API_KEY` from `adapterConfig.env` no longer re-adds the slot on `GET`.

## Risks

- **Host-leak regression on hosts with a host-level `OPENAI_API_KEY`.** Low risk: the run-time `stripInheritedOpenAiApiKey` guard preserves the original intent — when config omits the key, the inherited host value is stripped before spawning Codex. The guard only no-ops when the config explicitly provides a value (opt-in). The net behaviour is equivalent to the old empty-override approach for the leak case, and strictly better for the config-noise case.
- **Older Codex CLI versions (< 0.122) that read `OPENAI_API_KEY`.** Unaffected: those versions would have received an empty string before (which they treat as "no key" and fall back to other auth), and now receive nothing (same effect). Users who want API-key auth must still set `OPENAI_API_KEY` explicitly in `adapterConfig.env`, same as before.
- **Stored configs that already have the empty slot.** Not auto-cleaned by this PR (no migration). Operators can delete the slot via `PATCH` and it will now stay deleted. This is intentional — no schema migration is needed and no existing behaviour changes for agents that keep the slot.

## Model Used

- Provider: OpenClaw (ollama-cloud)
- Model ID: `ollama-cloud/glm-5.2`
- Capabilities: tool use, code execution, agentic workflows

## Checklist


steingran:
This PR fixes a real problem for anyone using OAuth authentication for OpenAI Codex CLI - agents are simply failing without the fix in this PR

devinfoley:
Yeah, sorry about this. It's fixed here: https://github.com/paperclipai/paperclip/pull/8425.

steingran:
> Yeah, sorry about this. It's fixed here: #8425.

Great! I saw you fixed the 'other half' of the bug as well :-)

PR #8418, opened 2026-06-21
fix(recovery): guard escalateStrandedAssignedIssue against terminal-status race

teamesys:
## Summary

`reconcileStrandedAssignedIssues` snapshots an issue, then later calls `escalateStrandedAssignedIssue` to flip it to `blocked`. If the assignee patches the issue to `done` (or `cancelled`) between the snapshot and the update, the recovery sweep overwrites the completion with `blocked`. The dependent graph then re-wakes the original owner, who re-marks `done`, and the cycle repeats — a reopen / ping-pong loop.

## Changes

Two guards in `server/src/services/recovery/service.ts`:

1. **Primary guard** — `escalateStrandedAssignedIssue` re-reads the issue's status from the database immediately after the early-return checks. If the fresh status is terminal (`done` / `cancelled`), the function returns `null` and skips the rest of the escalation pipeline. No `blocked` flip, no recovery action, no wake. Placed at function entry (after the existing `isStrandedIssueRecoveryIssue` early return) rather than immediately before `issuesSvc.update` so we don't leave a phantom recovery action behind in the racing path.

2. **Defence-in-depth** — `enqueueSourceScopedStrandedRecoveryWake` also checks `isTerminalIssueStatus(input.issue.status)` at entry. The primary guard already prevents the racing path from reaching this function, but a future caller passing a stale `input.issue` should still not be able to wake owners against a terminal issue.

Both guards reuse the existing `isTerminalIssueStatus` predicate.

## Tests

New: `server/src/__tests__/recovery-terminal-status-race-guard.test.ts` — uses the existing embedded-Postgres harness (same pattern as `recovery-stale-issue-lock-sweep.test.ts`). Two cases:

- The assignee patches `done` between the reconcile snapshot and the recovery update → `escalateStrandedAssignedIssue` returns `null`, status stays `done`, no wake is enqueued.
- Same scenario with `cancelled` instead of `done`.

## Background

Discovered in production via a local TeamE deployment (internal ticket `TEAAAA-229`): an issue was being repeatedly reopened from `done` → `blocked` by the recovery sweep, only to be re-completed by the assignee on the next heartbeat. The patch has been validated in that environment by patching the compiled `dist/services/recovery/service.js` directly. This PR ports the validated fix to the TypeScript source and adds the regression test.

## Test plan

- [ ] CI green on the new test file and existing recovery tests
- [ ] No regression in `recovery-stale-issue-lock-sweep.test.ts` or `service.pause-durability.test.ts`
- [ ] Verify in a release that the TEAAAA-229 ping-pong shape does not recur on rollout

---

🤖 Submitted on behalf of @teamesys via Claude Code (Team-E internal escalation TEAAAA-233).

### DIFF
--- a/server/src/__tests__/recovery-terminal-status-race-guard.test.ts
+++ b/server/src/__tests__/recovery-terminal-status-race-guard.test.ts
@@ -1,0 +1,197 @@
+import { randomUUID } from "node:crypto";
+import { eq } from "drizzle-orm";
+import { afterAll, afterEach, beforeAll, describe, expect, it, vi } from "vitest";
+import {
+  agents,
+  companies,
+  createDb,
+  heartbeatRuns,
+  issueComments,
+  issueRelations,
+  issues,
+  activityLog,
+} from "@paperclipai/db";
+import {
+  getEmbeddedPostgresTestSupport,
+  startEmbeddedPostgresTestDatabase,
+} from "./helpers/embedded-postgres.js";
+
+const mockTelemetryClient = vi.hoisted(() => ({ track: vi.fn() }));
+vi.mock("../telemetry.ts", () => ({ getTelemetryClient: () => mockTelemetryClient }));
+
+import { recoveryService } from "../services/recovery/service.ts";
+
+const embeddedPostgresSupport = await getEmbeddedPostgresTestSupport();
+const describeEmbeddedPostgres = embeddedPostgresSupport.supported ? describe : describe.skip;
+
+if (!embeddedPostgresSupport.supported) {
+  console.warn(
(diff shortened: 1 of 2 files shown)


teamesys:
Superseded by #8419 — that PR has the cleaner diff and the canonical test path. Closing this one to avoid duplicate review effort. Sorry for the noise.

Context: this PR was opened during a transient adapter retry on our internal task tracker (TEAAAA-233) and #8419 was opened ~4 minutes later from a cleaner branch. Both contain the same TEAAAA-229 fix (terminal-status race guard in `escalateStrandedAssignedIssue` + `enqueueSourceScopedStrandedRecoveryWake`).

PR #8445, opened 2026-06-21
[agent] Coalesce blocked scheduled routine runs

anavarro-bb:
## Summary
- Coalesce scheduled routine dispatch into existing same-fingerprint `blocked` or `in_review` routine execution issues even when no heartbeat run is live.
- Keep manual/API/webhook behavior unchanged so fresh external input can still create new work when appropriate.
- Add regression coverage for scheduled ticks against `blocked` and `in_review` routine issues with no live heartbeat.

## Root Cause
`findLiveExecutionIssue` only matched routine issues tied to queued/running/scheduled heartbeat runs. Once an agent marked a routine execution issue `blocked`, the heartbeat was no longer live, so the next scheduled tick could create another issue and wake the agent back into the same blocked/no-followup context.

## Verification
- `node ~/.cache/node/corepack/v1/pnpm/9.15.4/bin/pnpm.cjs vitest run server/src/__tests__/routines-service.test.ts`
- `git diff --check`

## Notes
- `pnpm exec eslint ...` was unavailable because this checkout does not expose an `eslint` command.

### DIFF
--- a/server/src/__tests__/routines-service.test.ts
+++ b/server/src/__tests__/routines-service.test.ts
@@ -803,6 +803,85 @@ describeEmbeddedPostgres("routine service live-execution coalescing", () => {
     expect(routineIssues[0]?.id).toBe(previousIssue.id);
   });
 
+  it.each(["blocked", "in_review"] as const)(
+    "coalesces scheduled routine runs into an existing %s issue without a live heartbeat",
+    async (status) => {
+      const { companyId, issueSvc, routine, svc, wakeups } = await seedFixture();
+      const previousRunId = randomUUID();
+      const previousIssue = await issueSvc.create(companyId, {
+        projectId: routine.projectId,
+        title: routine.title,
+        description: routine.description,
+        status,
+        priority: routine.priority,
+        assigneeAgentId: routine.assigneeAgentId,
+        originKind: "routine_execution",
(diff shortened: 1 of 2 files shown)


michaeljgrimm:
This PR matches a live Data Advantage failure mode from the weekly LLMReference GSC sync routine: scheduled fires created duplicate blocked routine execution issues for the same unresolved Search Console credential/access gate after the prior heartbeat was no longer live.

The important acceptance points for that case are:
- scheduled runs should reuse/coalesce into a same-fingerprint blocked or review-waiting routine issue without requiring a live heartbeat run;
- manual/API/webhook runs should keep their current behavior;
- routine output/history should record the dedupe (`coalesced`/`skipped` + linked issue) so ops can close the scheduled execution cleanly;
- once the blocked issue is completed/cancelled, the next scheduled run should create fresh work normally.

I independently validated the same service path locally with `pnpm exec vitest run server/src/__tests__/routines-service.test.ts`, `pnpm typecheck`, and `pnpm build` after refreshing dependencies with `pnpm install --frozen-lockfile`. The first `pnpm typecheck` hit stale local workspace links in `@paperclipai/skills-catalog`; the frozen install repaired that and the rerun passed. No issue with the routine-service change surfaced.

PR #8476, opened 2026-06-22
 Observability - Server changes to produce OpenTelemetry trace context 

henrikrexed:
# Observability — server changes to produce OpenTelemetry trace context

Supersedes #3750 — please close that one in favour of this.

## Thinking Path

The platform had no end-to-end traces: server-side work (heartbeat dispatch, issue
lifecycle, DB queries) produced no spans, and the observability plugin had no trace
context to attach telemetry to. The design decision (from prior review) is that the
**observability plugin is the single OTLP exporter**, and the server should only
*mint and propagate* W3C trace context — not export its own spans and not enable
Node SDK / HTTP / Express / PG auto-instrumentation (that volume is noise). This PR
implements that server "producer seam".

No public issue. In-PR description: adds server-side OpenTelemetry trace-context
production and propagation so the observability plugin (companion PR #8477) can build
the trace tree. Supersedes the stale #3750.

## What Changed

- `server/src/services/trace-context.ts` — a `BasicTracerProvider` whose span
  processor is a `NoopExporter` (spans are ephemeral; they exist only to mint valid
  trace/span IDs). Helpers: `withHeartbeatSpan()`, `withIssueSpan()`,
  `startChildSpan()`, `extractTraceContext()`, `bindActiveContext()`.
- `server/src/services/db-instrumentation.ts` — DB query span helper.
- Plugin-host/event-contract wiring (`plugin-host-services.ts`,
  `packages/plugins/sdk` types, `packages/shared` constants) so plugin events carry
  `PluginEvent.traceContext`.
- No live server-span OTLP export and no auto-instrumentation packages.

## Verification

- The trace-context module is consumed by the plugin's reconstruction path
  (companion PR #8477).
- ⚠️ **CI is currently red and cannot pass as-is** — see Risks. The new code imports
  `@opentelemetry/api`, `@opentelemetry/sdk-trace-base`, `@opentelemetry/resources`,
  and `@opentelemetry/context-async-hooks`, which are **not yet in `server/package.json`
  or the committed lockfile**, so the upstream `Build`/`Typecheck` jobs fail with
  `Cannot find module '@opentelemetry/*'`.

## Risks

- **Dependency/lockfile gate (blocking).** This change requires adding the four
  `@opentelemetry/*` base packages as server dependencies. Upstream CI rejects manual
  `pnpm-lock.yaml` edits (`Block manual lockfile edits`), so these deps must be added
  via the maintainer's `chore/refresh-lockfile` workflow before CI can install them
  and the build can pass. **This needs a maintainer action.**

## One logical change

This PR is scoped to the **server-side observability producer seam**: creating trace
context and emitting the domain events (run/issue/delegation/cost/db) that the
observability plugin consumes. The previously-bundled, unrelated
`server/src/adapters/ollama-adapter.ts` has been **removed** so the diff is a single
logical change.

## Model Used

Claude (Anthropic) via the Paperclip observability agent — Opus-class model.

## Checklist

- [x] I have searched GitHub for duplicate or related PRs and linked them above (supersedes #3750)
- [x] I have described the issue in-PR
- [x] I have run tests locally and they pass (blocked on the dependency/lockfile gate above)
- [ ] Greptile is 5/5 with no open P2s


henrikrexed:
Thanks for the review. Conflict with `master` is resolved, CI is green, and the findings are addressed below.

**Fixed in this revision**
- **P1 — `costs.ts` unhandled rejection (`costs.ts:125`):** the fire-and-forget `cost_event.created` activity-log emit now has a `.catch()` that logs and swallows, so a transient DB error in `logActivity` can no longer surface as a process-fatal unhandled rejection. (`492b6aee`)
- **P2 — probe span exported to a live backend (`trace-context.ts:54`):** when an external `TracerProvider` is already registered, the `__probe__` span is no longer `end()`-ed. Exporters flush on span-end, so ending it would push a meaningless probe span to the live backend; leaving it unended keeps the by-design no-noise contract. (`492b6aee`)

**Intentional / by-design (no change)**
- **P2 — empty `companyId` in `db-instrumentation.ts:154`:** instance-scoped DB queries (e.g. `instance_settings`) genuinely have no owning company. `""` is the explicit "instance-scoped / no company" sentinel; the plugin router treats it as unattributed rather than mis-routing. Open to a named constant (e.g. `INSTANCE_SCOPED`) if you prefer it spelled out.
- **Delegation payload `issueId` / `taskId` "duplication":** `taskId` is a load-bearing compatibility alias, not redundant — it is read as the fallback for `issueId` by `task-watchdogs`, `recovery`, and the `heartbeat` resume paths (`context.taskId ?? context.issueId`). It pre-exists on `master`; removing it would break those consumers.

**Acknowledged — propose tracking separately**
- **P2 — `agent.session.chunk` raw message content broadcast to all plugins (`plugin-host-services.ts:2729/2732`):** valid hardening point. The observability plugin only needs span metadata, not raw content, so I'd prefer to add content redaction/opt-in at the broadcast seam rather than widen this PR. Happy to open a follow-up issue for a `redactSessionContent` host option so trusted vs. untrusted plugins receive different payloads.
- **P3 — `String(err)` in `db.query.completed` (`db-instrumentation.ts:118`):** agreed it can leak schema detail; I'll truncate/redact the error string in the same follow-up.

**Out of scope (pre-existing `master` code, not in this diff)**
- The P0/P1 security findings on `server/src/adapters/ollama-adapter.ts` (filesystem read/write, `taskId` URL interpolation) are against a file this PR does not touch. They deserve a dedicated hardening issue rather than blocking this producer-seam PR.

**Lockfile / CI:** the OTel deps are declared in `server/package.json`; `pnpm-lock.yaml` is intentionally left at the `master` baseline so the `policy` "Block manual lockfile edits" gate passes (CI regenerates it). Snyk cleared once the `master` merge bumped `@aws-sdk/client-s3` to `^3.1072.0`.

PR #8549, opened 2026-06-23
feat(opencode_local): host-global fan-out gate + model-enum hardening (XIP-4690)

massiveprime:
## What & why

Durable fix for the `opencode_local` thundering-herd incident (Paperclip XIP-4675 / XIP-4690). The daily research fan-out wakes ~6 distinct `opencode_local` researcher agents simultaneously; each spawns `opencode run` + `opencode models`, saturating the shared local host until the enumeration wrapper trips `opencode models timed out after 20s`. Root cause: **the dispatcher only caps runs per-agent — there is no host-global gate across agents.** An interim routine-stagger already mitigated the incident; this is the durable, defense-in-depth fix.

### Fix 1 — host-global concurrency gate for `opencode_local` (`server/src/services/heartbeat.ts`)
- New `countRunningRunsForAdapterType(adapterType)` helper (joins `heartbeatRuns` → `agents`).
- In `startNextQueuedRunForAgent`, after the per-agent slot check, gate `opencode_local` agents on a host-global cap: `PAPERCLIP_OPENCODE_LOCAL_MAX_CONCURRENT` (default **2**). Over the cap, the run stays queued; the existing periodic `resumeQueuedRuns()` re-poll drains it as slots free — no starvation.

### Fix 2 — model-enum hardening (`packages/adapters/opencode-local/src/server/models.ts`)
- **(a) single-flight:** concurrent cold-cache lookups now share one in-flight `opencode models` spawn.
- **(b) cache key:** dropped `cwd` from the cache key — the model list is a function of the binary + provider auth, not the working directory, so distinct researcher cwds now share a cached result.
- **(c) env timeout:** `PAPERCLIP_OPENCODE_MODELS_TIMEOUT_MS` makes the previously-hardcoded 20s enumeration timeout configurable (min 1000ms); grace bumped to 5s.

### Fix 3 — orphan reaper
Out of scope for this PR (kept reviewable). `runChildProcess` already group-kills the opencode run's process group; residual orphans are re-parented helper grandchildren (agent-browser / Chrome-for-Testing / railway-mcp). Tracked separately; if a host reaper is added it must be age+orphan-gated and must not touch docker/postgres/es.

## Tests
`packages/adapters/opencode-local/src/server/models.test.ts` — added three tests using a controllable fake `opencode`:
- single-flight collapses 6 concurrent lookups to 1 spawn;
- cache reused across different cwds (1 spawn);
- `PAPERCLIP_OPENCODE_MODELS_TIMEOUT_MS` honored (times out at the configured 1s).

## New env knobs
| Var | Default | Effect |
|---|---|---|
| `PAPERCLIP_OPENCODE_LOCAL_MAX_CONCURRENT` | `2` | Max concurrent running `opencode_local` runs host-wide |
| `PAPERCLIP_OPENCODE_MODELS_TIMEOUT_MS` | `20000` | `opencode models` enumeration timeout (ms) |

Backward-compatible: both default to prior behavior aside from the new global cap (2), which is the intended primary fix.

### DIFF
--- a/packages/adapters/opencode-local/src/server/models.test.ts
+++ b/packages/adapters/opencode-local/src/server/models.test.ts
@@ -1,15 +1,43 @@
+import { mkdtempSync, writeFileSync, chmodSync, readFileSync, existsSync } from "node:fs";
+import os from "node:os";
+import path from "node:path";
 import { afterEach, describe, expect, it } from "vitest";
 import {
+  discoverOpenCodeModels,
+  discoverOpenCodeModelsCached,
   ensureOpenCodeModelConfiguredAndAvailable,
   listOpenCodeModels,
   requireOpenCodeModelId,
   resetOpenCodeModelsCacheForTests,
 } from "./models.js";
 
+// Creates a fake `opencode` executable that records each invocation (one line
+// per spawn appended to `counterFile`), optionally sleeps `sleepSec`, then
+// prints a fixed two-model listing. Returns the script path.
+function makeFakeOpenCode(opts: { counterFile: string; sleepSec?: number }): string {
+  const dir = mkdtempSync(path.join(os.tmpdir(), "fake-opencode-"));
+  const script = path.join(dir, "opencode");
+  const sleep = opts.sleepSec ? `sleep ${opts.sleepSec}` : "true";
+  writeFileSync(
+    script,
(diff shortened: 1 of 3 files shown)


massiveprime:
Superseded by #8551, which carries the same two fixes plus an env-configurable model-enum timeout and an embedded-Postgres server integration test, and a fully template-compliant description. Closing this in favor of #8551 to avoid duplicate review.

PR #8566, opened 2026-06-23
feat(claude-local): per-stream idle watchdog to reap wedged CLI streams

c-barlow:
## Thinking Path

> - Agents run via local CLI adapters (e.g. `claude-local`) that spawn a CLI subprocess and stream its stdout back to the server
> - When that stream goes silent mid-run — a half-open SSE socket after a laptop sleeps on battery (Wi-Fi suspended), or a stalled backend — nothing at the adapter layer notices
> - The only backstop today is the Claude CLI's own idle timeout (~11h), so a wedged run can occupy a slot for most of a day and trigger false-positive stale-active-run alerts at the 1h recovery mark
> - This pull request adds a per-stream stdout idle watchdog at the local subprocess runner, reaping a silent run within a configurable window (default 30m)
> - The benefit is wedged streams fail fast and cleanly with a structured reason instead of hanging for hours

No public GitHub issue exists — describing the underlying problem inline, following the feature issue template:

**Problem**

Local CLI adapter runs can wedge when the child process stops producing stdout (a half-open SSE socket after network suspension on laptop sleep, or a stalled backend). There is no idle detection at the adapter layer, so the run hangs until the Claude CLI's own ~11h idle timeout fires — occupying a slot for most of a day and tripping false-positive stale-active-run alerts at the 1h mark.

**Solution**

A per-run watchdog in the local subprocess runner that resets on every stdout chunk and, after a configurable silence window (default 30m), SIGTERMs the subprocess (SIGKILL after the grace period) and marks the run failed with a structured `stream_idle_watchdog` reason plus the observed `silenceAgeMs`.

**Alternatives considered**

Keep relying on the CLI's ~11h idle timeout (status quo — far too slow, wastes a slot for hours). Lower the server-side `ACTIVE_RUN_OUTPUT_SUSPICION_THRESHOLD_MS` (1h) instead — that only changes alerting, it does not reap the wedged process. Power-state hooks for laptop sleep — narrower and more brittle than watching the stream directly; left as a possible follow-up.

**Roadmap alignment**

Operational reliability hardening of the local adapter runtime; not a new product surface and does not overlap planned core feature work in ROADMAP.md.

## What Changed

- `packages/adapter-utils/src/server-utils.ts` — `runChildProcess` gains `streamIdleWatchdogMs`. A timer is armed on spawn and re-armed on every stdout chunk; on expiry it `SIGTERM`s the subprocess (process-group aware) and escalates to `SIGKILL` after `graceSec`. The result reports `idleWatchdogFired` + `silenceAgeMs`. Cleared on `close`/`error`. Independent of `timeoutSec` (the wall-clock cap), which is unchanged.
- `packages/adapter-utils/src/execution-target.ts` — threads `streamIdleWatchdogMs` through `runAdapterExecutionTargetProcess` to the local path (remote transports ignore it). Without this the watchdog would ship silently disabled on the local execution path.
- `packages/adapters/claude-local/src/server/execute.ts` — resolves the window from `CLAUDE_STREAM_IDLE_WATCHDOG_MS` (env, wins) → `config.streamIdleWatchdogMs` → default **30 min**; `<= 0` disables it (escape hatch). A tripped watchdog yields a failed run with `errorCode: "stream_idle_watchdog"` and `silenceAgeMs` in both `errorMeta` and `resultJson.streamIdleWatchdog` (durable column).
- CI — registered `adapter-utils` as a vitest project (`packages/adapter-utils/vitest.config.ts` + root `vitest.config.ts`) so its suite runs in CI; it was not wired in before.

## Verification

- `cd packages/adapter-utils && npx vitest run src/server-utils.test.ts` → **53 passed**, including 3 new cases: fires on a stdout-silent subprocess (asserts `idleWatchdogFired` + `silenceAgeMs`), re-arms on a chatty subprocess (survives, exits 0), stays disabled when `<= 0`.
- `npx tsc --noEmit` clean in `packages/adapter-utils` and `packages/adapters/claude-local`.

## Risks

Low risk. The watchdog is opt-in by value and fully disabled when `<= 0`; it does not change `timeoutSec` semantics. Worst case of a too-tight window is an early termination of a legitimately slow-but-alive run — mitigated by the 30m default and the env override for hot-tuning without redeploy. Only the local execution path is affected; remote transports are untouched.

## Model Used

Claude (Anthropic), `claude-opus-4-8` (Opus 4.8), 1M context, extended thinking + tool use / code execution.

## Checklist

### DIFF
--- a/packages/adapter-utils/src/execution-target.test.ts
+++ b/packages/adapter-utils/src/execution-target.test.ts
@@ -280,6 +280,46 @@ describe("runAdapterExecutionTargetProcess", () => {
       }),
     );
   });
+
+  it("forwards streamIdleWatchdogMs to runChildProcess on the local path", async () => {
+    // Regression guard: the local execution wrapper must thread the per-stream
+    // idle watchdog through to runChildProcess. If it silently drops the option
+    // the watchdog ships disabled on the claude-local path while still appearing
+    // configured.
+    const runChildProcessSpy = vi.spyOn(serverUtils, "runChildProcess").mockResolvedValue({
+      exitCode: 0,
+      signal: null,
+      timedOut: false,
+      stdout: "",
+      stderr: "",
+      pid: null,
+      startedAt: new Date().toISOString(),
+    });
+
+    await runAdapterExecutionTargetProcess(
+      "run-local-process",
+      null,
+      "agent-cli",
+      ["--json"],
+      {
+        cwd: "/tmp/local",
+        env: { SAFE_VALUE: "visible" },
+        timeoutSec: 5,
+        graceSec: 1,
+        streamIdleWatchdogMs: 60_000,
+        onLog: async () => {},
+      },
+    );
+
+    expect(runChildProcessSpy).toHaveBeenCalledWith(
+      "run-local-process",
+      "agent-cli",
+      ["--json"],
+      expect.objectContaining({
+        streamIdleWatchdogMs: 60_000,
+      }),
+    );
+  });
 });
 
 describe("ensureAdapterExecutionTargetRuntimeCommandInstalled", () => {
--- a/packages/adapter-utils/src/execution-target.ts
+++ b/packages/adapter-utils/src/execution-target.ts
@@ -82,6 +82,12 @@ export interface AdapterExecutionTargetProcessOptions {
   stdin?: string;
   timeoutSec: number;
   graceSec: number;
+  /**
+   * Per-stream idle watchdog (ms). Forwarded to `runChildProcess` on the local
+   * execution path; remote (SSH/sandbox) transports rely on their own timeouts
+   * and ignore it. Omitted/<=0 disables the watchdog.
+   */
+  streamIdleWatchdogMs?: number;
   onLog: (stream: "stdout" | "stderr", chunk: string) => Promise<void>;
   onRuntimeProgress?: RuntimeStatusSink;
   onSpawn?: (meta: { pid: number; processGroupId: number | null; startedAt: string }) => Promise<void>;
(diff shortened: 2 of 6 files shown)


c-barlow:
Rebased onto current `master` (was 562 commits behind a stale fork base → CONFLICTING; now **MERGEABLE**, 2 commits ahead / 0 behind). Re-verified after the rebase: `adapter-utils` + `claude-local` typecheck clean; the 3 watchdog tests and all 26 `execution-target` tests pass.

Two things surfaced during the rebase that are worth a reviewer's eye:

1. **Forwarding fix (new file in the diff):** upstream refactored the adapter run path from a direct `runChildProcess` call to `runAdapterExecutionTargetProcess`. That wrapper did **not** forward `streamIdleWatchdogMs` to `runChildProcess`, so as a pure rebase the watchdog would have been silently dropped on the local path. Added `streamIdleWatchdogMs` to `AdapterExecutionTargetProcessOptions` and threaded it through to `runChildProcess` (local path only; SSH/sandbox transports keep their own timeouts).

2. **`RunProcessResult` fields are now optional** (`idleWatchdogFired?`, `silenceAgeMs?`). Upstream now has many `RunProcessResult` producers (SSH, every sandbox provider); only the local `runChildProcess` path arms the watchdog, so required fields would have broken all of them. Optional is both minimal and semantically correct.

The vitest wiring shrank too: upstream already registers `packages/adapter-utils` as a project, so this PR now only adds the package's `vitest.config.ts` (the project still lacked one).

Substance is unchanged — still the watchdog-only change. Ready for the merge + supervised-deploy window once the board authorizes.

PR #8598, opened 2026-06-24
fix(adapter-utils): runChildProcess escalates to SIGKILL on liveness, not child.killed

justinltodd:
## Thinking Path

> - Those agents run as child processes spawned by `@paperclipai/adapter-utils`'s `runChildProcess`, which arms a parent-side wall-clock timer at `timeoutSec` to bound a hung run
> - At the deadline `runChildProcess` sends SIGTERM, then after a grace window escalates to SIGKILL — the SIGKILL backstop is what turns a wedged child into a dead PID so the scheduler can reclaim and retry it
> - On the **direct-child fallback** path (`signalRunningProcess`, used on win32 and whenever process-group signaling is unavailable or throws) the escalation was gated on `!child.killed`
> - But Node sets `ChildProcess.killed` to `true` the instant a signal is *successfully sent*, not when the process exits — so once the earlier SIGTERM has been sent, `child.killed` is already `true`, the `!child.killed` guard is `false`, and the SIGKILL escalation never runs
> - A child that ignores SIGTERM (e.g. a graceful-shutdown handler wedged on a socket) is therefore never force-killed, outlives its deadline, and for an unattended scheduler sits running forever with no terminal state
> - This PR gates the fallback escalation on real liveness (`exitCode === null && signalCode === null`), so SIGKILL fires precisely while the child is still alive
> - The benefit is the hard timeout actually guarantees termination (except true uninterruptible D-state) on every platform/configuration, not just where the process-group path is available

No existing public issue — describing the bug inline, following `.github/ISSUE_TEMPLATE/bug_report.yml`:

### What happened?

When `@paperclipai/adapter-utils`'s `runChildProcess` reaches `timeoutSec` and the spawned child ignores SIGTERM, the SIGKILL escalation on the **direct-child fallback** path (`signalRunningProcess`, taken on win32 or whenever `process.kill(-pgid, …)` is unavailable or throws) never fires, so the child outlives its deadline indefinitely. Root cause: the escalation is gated on `!running.child.killed`, and `ChildProcess.killed` reflects only that a signal was *successfully sent* (per the Node docs it "does not indicate that the child process has been terminated"). After the deadline SIGTERM, `child.killed` is already `true`, so `!child.killed` is `false` and the follow-up SIGKILL is suppressed.

### Expected behavior

After the grace window, a child that is still alive is force-killed with SIGKILL regardless of whether SIGTERM was already sent — the hard timeout should guarantee termination (except true uninterruptible D-state) on every platform/configuration.

### Steps to reproduce

1. Spawn a child that installs a no-op `SIGTERM` handler and never exits (e.g. `process.on('SIGTERM', () => {}); setInterval(() => {}, 1000)`).
2. Drive it through the direct-child fallback, i.e. `signalRunningProcess({ child, processGroupId: null }, …)` (the path used on win32 / when group signaling is unavailable).
3. Send SIGTERM (the child swallows it; `child.killed` becomes `true`), then send SIGKILL.
4. On the pre-fix `!child.killed` guard the SIGKILL call is a no-op and the PID survives past its deadline. Covered by the new regression test in this PR.

Reproduces on `master` (the `signalRunningProcess` fallback). Also present in published `@paperclipai/adapter-utils` (e.g. `2026.325.0`), where the same `!child.killed` guard sits on the single direct-child escalation path.

_Searched the open PR list for duplicates/related work on `runChildProcess` / `signalRunningProcess` / SIGKILL escalation; found none._

## What Changed

- `packages/adapter-utils/src/server-utils.ts`: in `signalRunningProcess`, replace the direct-child fallback guard `!running.child.killed` with `running.child.exitCode === null && running.child.signalCode === null` (real liveness). The process-group path is unchanged.
- `packages/adapter-utils/src/server-utils.ts`: `export` `signalRunningProcess` so the fallback branch can be unit-tested directly.
- `packages/adapter-utils/src/server-utils.test.ts`: add a companion regression test (POSIX-only, like the sibling timeout tests) that forces the fallback (`processGroupId: null`) — sends SIGTERM (child swallows it, `child.killed` becomes `true`), asserts the child is still alive, then sends SIGKILL and asserts the PID dies. Also keeps the end-to-end `runChildProcess` SIGTERM-ignoring test.

## Verification

```
npx vitest run packages/adapter-utils/src/server-utils.test.ts   # 52 passed
npx tsc --noEmit                                                  # clean
```

- **Regression proof:** reverting the guard to `!running.child.killed` makes the new fallback test fail (`waitForPidExit` → false; the child survives); the liveness guard makes it pass. This addresses the prior review note that the existing test only exercised the process-group path (which already escalated correctly on POSIX) and never reached the changed branch.

## Risks

Low. A one-line guard change scoped to the direct-child fallback; the process-group path is untouched. SIGKILL is only sent when `exitCode`/`signalCode` are both still `null`, i.e. the process is provably alive, so the change cannot signal an already-reaped/recycled PID. New tests are POSIX-only and `skipIf(win32)`, consistent with the sibling timeout tests in this file.

## Model Used

Anthropic **Claude Opus 4.8**, driven via the Cursor agent (extended reasoning + tool use, large context). Diff, tests, and the regression proof above were produced and run by the agent; reviewed by a human before pushing.

## Checklist


justinltodd:
Thanks @greptile-apps — both points addressed in 396d4719a:

1. **Test exercised the working path, not the fixed fallback (P2).** Correct — the existing test spawns detached and takes the process-group path (`process.kill(-pgid, …)`), which already escalated SIGKILL correctly on POSIX and returns before the direct-child fallback this PR changes. I exported `signalRunningProcess` and added a companion POSIX test that forces the fallback (`processGroupId: null`): it sends SIGTERM (the child swallows it, so `child.killed` is now `true`), asserts the child is still alive, then sends SIGKILL and asserts the PID dies. I verified it's a real regression test — reverting the guard to `!child.killed` makes it fail (the child survives); the liveness guard makes it pass. The end-to-end group-path test is retained.

2. **PR description missing required template sections (P2).** Rewrote the description to follow `.github/PULL_REQUEST_TEMPLATE.md` — Thinking Path, in-PR issue description (no public issue exists), What Changed, Verification, Risks, Model Used, and the Checklist (incl. the dedup search).

`npx vitest run packages/adapter-utils/src/server-utils.test.ts` → 52 passed; `tsc --noEmit` clean. Ready for re-review.

PR #8625, opened 2026-06-25
feat(heartbeat): rotate persisted session after an invisible 0-token run death

andrewcolinchandra:
## Problem

A run the gateway classifies as **succeeded** yet emits **zero input and output tokens** (typically a `529` / provider-overload response returning exit 0 with no error) is an *invisible* death. Its outcome looks healthy, so the persisted session is resumed by the next run — even though that session is now stale or poisoned. The next run then resumes from a dead session instead of starting fresh.

## Change

Extend `resolveNextSessionState` to detect this case and return the **same cleared-state contract as adapter `clearSession`** (`params` + `displayId` both `null`), which the downstream caller (`heartbeatService`) already translates into a task-session clear. The next run then starts a fresh session.

- New optional `usage?: UsageTotals | null` input to `resolveNextSessionState`.
- New helper `isZeroTokenUsage()` — true only when *reported* usage is all-zero.
- Wired at the single call site by passing the already-computed `normalizeUsageTotals(adapterResult.usage)`.

## Why it's safe (session continuity)

Narrowly scoped so normal runs are unaffected:

- Fires **only** when `outcome === "succeeded"` **and** reported usage is all-zero. Failed / timed-out / cancelled runs keep their existing resume behavior.
- A **missing** usage object (`null`/`undefined`) is treated as ambiguous and does **not** trigger rotation — preserves continuity when usage wasn't captured.
- Normal runs always produce `>0` tokens, so they never match.

Safety net for stale-session failures only. It does **not** cure a gateway-wide outage — only provider recovery / failover does.

## Tests

Adds 4 cases to the existing `resolveNextSessionState` describe block in `heartbeat-workspace-session.test.ts`:

1. rotates the persisted session when a succeeded run emitted zero tokens
2. does not rotate a zero-token run that did not classify as succeeded (gate)
3. does not rotate a healthy succeeded run that produced real token usage
4. does not rotate a succeeded run whose usage was not reported (missing-usage guard)

All 92 tests in the file pass.

## Related

Part of a 0-token crash-hardening set:
- #8624 — posts a diagnostic comment when a run finalizes with zero tokens (defines the canonical zero-token predicate in `heartbeat-zero-token-diagnostic.ts`; this PR mirrors that definition inline so it stays independently mergeable).
- #8623 — tunable no-comment-streak escalation threshold.

Touches `server/src/services/heartbeat.ts` — `resolveNextSessionState` (~L3362) and its call site (~L9825).

### DIFF
--- a/server/src/__tests__/heartbeat-workspace-session.test.ts
+++ b/server/src/__tests__/heartbeat-workspace-session.test.ts
@@ -1500,6 +1500,122 @@ describe("resolveNextSessionState", () => {
 
     expect(result.legacySessionId).toBe("from");
   });
+
+  it("rotates the persisted session when a succeeded run emitted zero tokens", () => {
+    const result = resolveNextSessionState({
+      adapterType: "hermes_local",
+      codec: truncatingHermesSessionCodec,
+      adapterResult: {
+        exitCode: 0,
+        signal: null,
+        timedOut: false,
+      },
+      outcome: "succeeded",
+      usage: { inputTokens: 0, cachedInputTokens: 0, outputTokens: 0 },
+      previousParams: {
+        sessionId: "20260601_141558_c861e4",
+      },
+      previousDisplayId: "20260601_141558_c861e4",
+      previousLegacySessionId: "20260601_141558_c861e4",
+    });
+
+    // Without rotation this succeeded run would resume the previous valid
+    // session; the 0-token death means that session is stale and must be dropped.
+    expect(result).toEqual({
+      params: null,
+      displayId: null,
+      legacySessionId: null,
+    });
+  });
+
+  it("does not rotate a zero-token run that did not classify as succeeded", () => {
+    const result = resolveNextSessionState({
+      adapterType: "hermes_local",
+      codec: truncatingHermesSessionCodec,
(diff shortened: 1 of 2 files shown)


andrewcolinchandra:
**Heads-up: possible duplicate of #8758.**

While running the dedup search required by the PR template, I found [#8758](https://github.com/paperclipai/paperclip/pull/8758) (`fix(heartbeat): rotate dead persisted session after 0-token run`, @jmiramant, BLU-17857). It targets the **same seam** — `heartbeat.ts` run-finalization, clearing `nextSessionState`/`sessionIdAfter` on a 0-token persisted-session death — and is more complete than this PR (adds a stale-session-rotation alert table + migration, schema, and recovery-origin wiring).

This PR (#8625) is a smaller, test-backed take on the same fix. I don't want to maintain a competing PR against an insider-authored one. **Happy to close this as a duplicate of #8758** if maintainers prefer that path — just say the word. Leaving it open for now only as a fallback in case #8758 stalls.

(If there's a meaningful behavioral difference you'd like preserved from this version — e.g. the opt-in flag shape — I'm glad to fold it into #8758 instead.)

PR #8648, opened 2026-06-26
fix(server): validate compact handoff emission

st-VALVe:
## Summary
- Validate compact working-state packets before emitting handoff markdown.
- Reuse the heartbeat mutation path through a tested helper so compact handoff markdown is set only for reset sessions with a self-report and persisted session id.
- Bind the server validator to the shared compact working-state fixture corpus to reduce contract drift.

## Root cause
The Step 2 emitter accepted nested packet sections as opaque input and serialized markdown without compact contract validation, so invalid packets could pass through the server path.

## Validation
- pnpm --filter @paperclipai/server exec vitest run src/__tests__/compact-working-state-builder.test.ts src/__tests__/compact-working-state-handoff-markdown.test.ts src/__tests__/heartbeat-workspace-session.test.ts: 103/103 passing
- pnpm --filter @paperclipai/server typecheck: passing
- node --test pipeline-v2/compact-working-state.test.mjs: 18/18 passing
- git diff --check: passing

### DIFF
--- a/packages/adapter-utils/src/types.ts
+++ b/packages/adapter-utils/src/types.ts
@@ -77,6 +77,12 @@ export interface AdapterExecutionResult {
   errorMeta?: Record<string, unknown>;
   usage?: UsageSummary;
   /**
+   * Optional adapter-specific run-shape telemetry (turn/tool counts, resident
+   * window). Persisted flat into the run's usageJson. Adapter-neutral so any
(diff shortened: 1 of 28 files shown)


st-VALVe:
Wrong target for this follow-up: the compact handoff base commits are in st-VALVe/paperclip. Closing this upstream PR and opening the fix against the fork master.

PR #8180, opened 2026-06-16
fix(deploy): bootable standalone deploy tooling + hardlink-safe apply-publish-config (TON-2276/TON-2280)

dosthcpp:
## Thinking Path

> - Paperclip is the open-source app people use to manage AI agents for work.
> - It ships as an npm-published monorepo; operators run a self-contained "standalone" bundle built via `pnpm deploy`.
> - Raw `pnpm deploy` leaves each `@paperclipai/*` package manifest pointing `exports["."]` / `main` at `./src/index.ts`, so the booted server crashes with `Cannot find module .../src/index.ts` (no TypeScript at runtime).
> - Acceptance testing (TON-2280) surfaced a second defect: the publish-config overlay rewrote manifests in place, but pnpm's content-addressable store hardlinks deployed `package.json` files back to the workspace source, so the in-place write corrupted the source tree.
> - This PR adds bootable standalone deploy tooling (build → deploy → finalize → verify) plus a hardlink/symlink-safe overlay step (atomic write to a fresh inode) so the artifact boots clean without mutating the workspace.
> - The benefit: `pnpm run deploy:standalone` produces an artifact that boots UI + API on an isolated instance with the verifier exiting 0, and live cutovers stop using raw `pnpm deploy`.

## Linked Issues or Issue Description

No standalone GitHub issue exists; the work is tracked on the Paperclip board as TON-2276 (durable fix) and TON-2280 (acceptance gate). Underlying problem, in bug-report shape:

**What happened**

A standalone bundle produced by raw `pnpm deploy` booted into `Cannot find module .../src/index.ts` because deployed `@paperclipai/*` manifests still pointed at TypeScript sources; separately, the publish-config overlay corrupted the workspace source tree by writing through pnpm-store hardlinks.

**Expected behavior**

`deploy:standalone` should yield an artifact that boots the real SPA UI + API on an isolated instance (HTTP 200 on `/`), with no TypeScript-source entry points and no mutation of the workspace source.

**Steps to reproduce**

1. `pnpm deploy --node-linker=hoisted <out>` for `@paperclipai/server`.
2. Boot `<out>` → server crashes resolving `src/index.ts`.
3. Run an in-place manifest overlay → the workspace `package.json` changes too (shared inode).

## What Changed

- `scripts/deploy-standalone.sh` — orchestrates the standalone deploy end-to-end (build → `pnpm deploy --node-linker=hoisted` → finalize → verify); replaces the pnpm-rejected `--legacy` flag.
[...]

PR #8660, opened 2026-06-26
fix(issues): tolerate unknown createdByRunId in addComment instead of FK 500

dosthcpp:
## Thinking Path

> - The issue-comments subsystem records agent/board status updates; each comment may carry `createdByRunId`, a column with a foreign key to `heartbeat_runs`.
> - `issueService.addComment` inserted `createdByRunId` straight from the actor without checking that the run exists, so a stale, foreign, or malformed run-id passed ownership checks and then failed at insert time.
> - That surfaced as a raw HTTP 500 (FK violation or invalid-uuid syntax) instead of a clean result, and the status comment plus its audit trail were lost.
> - This pull request resolves any run-id against `heartbeat_runs` first and demotes anything it cannot resolve to `NULL`, logging a warning.
> - The benefit is that a comment is never lost to a bad run-id while valid run-ids keep their attribution.

## What Changed

`addComment` (`server/src/services/issues.ts`) now resolves any `createdByRunId` against `heartbeat_runs` before insert (guarding non-UUID input with `isUuidLike`) and demotes unknown / foreign / malformed run-ids to `NULL`, logging a warning. Valid run-ids keep their attribution unchanged. Adds embedded-Postgres regression coverage for unknown-UUID, malformed, and valid run-ids.

## Verification

Ran the new embedded-Postgres test suite locally — 3/3 pass (unknown-UUID, malformed, valid run-id). Server typecheck (`tsc --noEmit`) reports no errors. Confirmed the previously-500ing insert now succeeds with `createdByRunId = NULL` while a valid run-id is preserved.

## Issue Description

No public GitHub issue exists; describing inline per the bug template.

**What happened**

`POST /api/issues/{id}/comments` returns HTTP 500 when the caller supplies a `createdByRunId` that does not exist in `heartbeat_runs` (stale/foreign run-id) or is not a valid UUID. Ownership validation passes, then the FK constraint `issue_comments_created_by_run_id_heartbeat_runs_id_fk` (or `invalid input syntax for type uuid`) rejects the insert, losing the comment and its audit trail.

**Expected behavior**

Comment creation should never 500 on a bad run-id. An unresolvable run-id should be demoted to `NULL` (attribution dropped) rather than failing the whole write.

**Steps to reproduce**

1. Call the comments endpoint / `addComment` with a `runId` that has no row in `heartbeat_runs` (e.g. a stale sentinel UUID), or with a non-UUID string.
2. Observe a 500 with the FK-violation / invalid-uuid error instead of a successful comment.

**Deployment mode**

Self-hosted / local control plane (server package).

## Related PRs

This area has several prior open PRs; this PR is scoped narrowly to `addComment` with regression tests. Closest prior art: #4795 (null out unresolvable run-id in addComment), #7014 (drop reaped-run attribution on comment/activity writes), #6342 (stale run_id FK from expired JWTs), #8065 (activity-log graceful degrade). Maintainers may prefer to consolidate.

## Risks

Low risk. Adds one indexed-PK lookup per comment that carries a run-id; valid run-ids behave exactly as before. Only previously-500ing inserts change behavior (now succeed with `createdByRunId = NULL`).

Follow-up (not in this PR): the same `createdByRunId: actor.runId ?? null` insert pattern exists in `server/src/services/document-annotations.ts` and `server/src/services/routines.ts`; those sites share the latent FK exposure and should be hardened similarly.

Note: this branch is based on the fork's `master`, which carries one standing unrelated base commit (an Activity-tab UI fix); its `IssueDetail.tsx` has been aligned to upstream so the diff is conflict-free.

## Model Used

Claude (Anthropic), claude-opus-4-8, extended thinking + tool use.

## Checklist

- [x] I searched the GitHub PR list for similar/duplicate PRs before opening this one (see Related PRs above)


dosthcpp:
Re-verified against master `696c694a` (2026-07-07): still unaddressed — `issues.ts` addComment still inserts `createdByRunId: actor.runId ?? null` with no heartbeat_runs existence check → FK 500. Will rebase on request. Triage summary of all our open PRs: #8180.

PR #8795, opened 2026-06-30
feat(run-logs): durable run-log store via object-storage mirror

stubbi:
## Problem
Run logs are written to the server pod filesystem (`run-log-store.ts`, the `local_file` store). When `PAPERCLIP_HOME` is an `emptyDir` (no persistent volume — common in hardened / ephemeral deployments), every pod restart wipes them while the DB row still references the file, so the run-log API returns **'Run log not found'** for completed runs after any redeploy.

## Fix
Transparent durability via the existing object-storage provider. When `RUN_LOG_S3_BUCKET` is set, the store mirrors each completed log to object storage on `finalize` (keyed by the same `logRef`) and falls back to it on `read` when the local file is gone. Live append/tail stays on the pod-local file (fast, no per-chunk PUT).

- Store id stays `local_file` -> no downstream change (feedback, read cast, fixtures untouched).
- Unset bucket -> local-only (today's behaviour, safe degrade).
- Upload is best-effort -> durability never breaks run finalization.
- Reuses `createS3StorageProvider`; deliberately separate from `PAPERCLIP_STORAGE_PROVIDER` so enabling durable logs does not redirect workspace/file storage.

## Config (all optional)
`RUN_LOG_S3_BUCKET`, `RUN_LOG_S3_ENDPOINT`, `RUN_LOG_S3_REGION`, `RUN_LOG_S3_PREFIX` (default `run-logs`), `RUN_LOG_S3_FORCE_PATH_STYLE` (default true). Creds via the standard AWS env chain. Works with any S3-compatible endpoint.

## Tests
TDD, 7 new tests: store-id stability, local live read (no S3), finalize upload, S3 fallback after a simulated pod-roll, ranged S3 read + nextOffset, not-found when neither has it, local-only safe degrade.

### DIFF
--- a/server/src/services/run-log-store.test.ts
+++ b/server/src/services/run-log-store.test.ts
@@ -1,0 +1,136 @@
+import { afterEach, beforeEach, describe, expect, it } from "vitest";
+import { promises as fs } from "node:fs";
+import path from "node:path";
+import os from "node:os";
+import { Readable } from "node:stream";
+import { createDurableRunLogStore } from "./run-log-store.js";
+import type { StorageProvider } from "../storage/types.js";
+
+// In-memory StorageProvider stand-in: durable, survives the "pod roll" (local
+// dir wipe) the same way Cubbit does. Records calls so we can assert behaviour.
+function createMemoryProvider() {
+  const objects = new Map<string, Buffer>();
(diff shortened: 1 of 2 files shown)


stubbi:

PR #8812, opened 2026-07-01
fix(instructions): self-heal managed agent instructions before each run

stubbi:
## Symptom

After an app-pod restart, managed agents run with **no persona/tools/hiring guidance** and no error surfaces. Every managed agent silently degrades to a "naked" run.

## Root cause

A managed agent's instructions bundle (persona + tools + hiring guidance) is materialized to the instance-root disk **only** at agent creation/import/plugin-reconcile. The run path never re-materializes it. When the instance root is an ephemeral volume (e.g. an `emptyDir` pod mount), a restart wipes the bundle. The adapter's `fs.readFile` then `ENOENT`s and the run proceeds with no instructions, silently.

## Fix (resilience-first, never fail a run)

- **`agentInstructions.ensureManagedInstructionsMaterialized()`** — a new managed-only, idempotent, self-gating guard. It no-ops for any agent whose instructions bundle is not in `managed` mode, so external/legacy bundles are never fabricated or overwritten. If the managed bundle is present it opportunistically captures a durable snapshot; if the entry file is missing it restores from the durable snapshot (preserving hand edits) or, as a guaranteed fallback, regenerates the role-default template.
- **Durable content store** — a new nullable `agents.managed_instructions_snapshot` jsonb column so customized bundles survive a wipe too. The migration adds **one nullable column** (safe, additive).
- **Heartbeat run path** — calls the guard before the adapter reads the bundle, persists the snapshot, and on a restore emits an observability `warn` + an `instructions.rematerialized` run event so the underlying instance-root wipe is detectable. Wrapped in `try/catch`: the self-heal can never turn a recoverable state into a user-visible failure.
- Non-managed / external bundles are untouched.

## Tests

`server/src/__tests__/agent-instructions-selfheal.test.ts` (5 tests): non-managed skip, present-on-disk capture, restore-from-durable-snapshot (preserves hand edits), regenerate-from-role-default fallback, and idempotency. All pass; server typecheck is clean.

### DIFF
--- a/packages/db/src/migrations/0126_agent_managed_instructions_snapshot.sql
+++ b/packages/db/src/migrations/0126_agent_managed_instructions_snapshot.sql
@@ -1,0 +1,1 @@
+ALTER TABLE "agents" ADD COLUMN IF NOT EXISTS "managed_instructions_snapshot" jsonb;
--- a/packages/db/src/migrations/meta/_journal.json
+++ b/packages/db/src/migrations/meta/_journal.json
@@ -883,6 +883,13 @@
       "when": 1782440100000,
       "tag": "0125_environment_custom_image_templates",
       "breakpoints": true
+    },
+    {
+      "idx": 126,
+      "version": "7",
+      "when": 1782440200000,
+      "tag": "0126_agent_managed_instructions_snapshot",
+      "breakpoints": true
     }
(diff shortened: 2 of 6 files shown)


stubbi:

PR #8820, opened 2026-07-01
feat(sandbox): stream command output live across the plugin worker boundary

stubbi:
## Problem

Agent runs that execute in a remote sandbox (the kubernetes sandbox provider, and any provider behind the plugin-worker RPC boundary) feel "stuck" with no feedback for minutes: the whole harness session runs as one buffered call and **no output reaches the client until the run finishes**, then it arrives all at once.

Two root causes:
1. `execInPod` accumulated stdout/stderr and only resolved at the end — the per-frame `data` events already fired but were never relayed.
2. Even with a streaming callback, the kubernetes plugin runs behind a JSON-RPC worker boundary, so a callback (`onOutput`) cannot be serialized across it and was silently dropped on the real path.

## What this does

Additive, backward-compatible live streaming, end to end:

- **`execInPod`** gains an optional `onChunk(stream, text)` invoked per data frame (buffered return unchanged; guarded so a throwing consumer can't break capture).
- **The runner/execute contract** (`CommandManagedRuntimeRunner.execute`, `RunProcessResult`, `PluginEnvironmentExecuteParams/Result`) gains optional `onOutput` and a `streamed` result flag.
- **Worker→host bridge**: the kubernetes plugin emits each chunk over the existing `ctx.streams` channel (`env-exec-output:<runId>`); the host subscribes to that channel via the plugin stream bus for the duration of the `environmentExecute` call and forwards chunks to `onOutput`. The stream bus is now owned by the worker manager and fed to the existing SSE bridge route as well.
- **No double logging**: when output was streamed live, `streamed: true` suppresses the trailing buffered log dump.

All six other sandbox providers accept the optional param unchanged (only kubernetes is wired for real). `runId`/`streamOutput` are plumbed as serializable RPC params. Fully unit-tested at each seam (pod-exec, plugin execute, RPC bridge, host subscription, dedup); existing suites green.

No behavior change when a caller does not opt into streaming (buffered fallback preserved).

### DIFF
--- a/packages/adapter-utils/src/command-managed-runtime.ts
+++ b/packages/adapter-utils/src/command-managed-runtime.ts
@@ -34,6 +34,10 @@ export interface CommandManagedRuntimeRunner {
     // logging. Runners that cannot stream may ignore this and leave `streamed`
     // unset — the caller then falls back to the buffered result unchanged.
     onOutput?: (stream: "stdout" | "stderr", text: string) => void | Promise<void>;
+    // Run correlation id. A streaming runner forwards this to the sandbox
+    // provider so worker-emitted output chunks can be routed back to `onOutput`
+    // over the plugin worker RPC boundary (the callback itself can't cross it).
+    runId?: string;
     onSpawn?: (meta: { pid: number; startedAt: string }) => Promise<void>;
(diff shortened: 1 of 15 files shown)


stubbi:

PR #8823, opened 2026-07-01
fix(skills): make paperclip-create-agent available to agents that can hire

stubbi:
## Problem

Agents whose permissions include `canCreateAgents` (e.g. a lead/CEO agent) cannot actually hire, because the `paperclip-create-agent` skill is never mounted into their k8s sandbox run. The skill's `SKILL.md` is what tells the agent how to submit a hire (`POST /api/companies/$PAPERCLIP_COMPANY_ID/agent-hires` using the injected `$PAPERCLIP_API_KEY`/`$PAPERCLIP_API_URL`), so without it a capable agent has no hire-flow instructions and reports it "cannot access the paperclip-create-agent skill".

### Root cause

At run time the opencode adapter builds the sandbox skills dir via `resolvePaperclipDesiredSkillNames(config, availableEntries)`, which returns `[]` unless the agent has an explicit `adapterConfig.paperclipSkillSync.desiredSkills` (`preference.explicit`). Managed agents (including a seeded CEO) have no explicit `desiredSkills`, so **zero** skills are synced (`Syncing skills to sandbox: 100% (0.0/0.0 MB)`) and `paperclip-create-agent` is absent.

## Fix (run-time, no backfill)

Include the create-agent skill in the sandbox for any agent whose permissions include `canCreateAgents`, resolved at **run time** so it covers both already-existing agents and newly-created ones (no data migration).

- `resolvePaperclipDesiredSkillNames(config, availableEntries, opts?)` gains an optional `opts.alwaysIncludeSkillKeys: string[]`. Those keys are unioned into the result (canonicalized the same way as explicit desired skills, and deduped) even when there is no explicit `desiredSkills` preference. **Behavior is unchanged when the option is absent**, and a requested skill is only included when it actually exists in `availableEntries` (no phantom entries). New exported constant `PAPERCLIP_CREATE_AGENT_SKILL_KEY`.
- The opencode adapter threads the agent's `canCreateAgents` into both skill-dir build paths (`buildOpenCodeSkillsDir` for the remote/sandbox path, and the local injection path), including `paperclip-create-agent` when the agent can hire.
- `AdapterAgent` gains an optional `permissions` field so adapters can read `canCreateAgents`.

Other local adapters can opt in cheaply via the same `alwaysIncludeSkillKeys` option (follow-up); this PR wires the managed-default opencode adapter.

## Tests (TDD)

- `resolvePaperclipDesiredSkillNames`: returns `["paperclip-create-agent"]` with the option and no explicit desiredSkills; `[]` without the option (unchanged); `[]` when the skill is unavailable (no phantom); unions + dedupes with explicit desiredSkills.
- opencode adapter: `buildOpenCodeSkillsDir` includes the create-agent skill when the caller passes `canCreateAgents: true` and excludes it when `false`.

All new and existing opencode/adapter-utils skill+execute tests green.

### DIFF
--- a/packages/adapter-utils/src/server-utils.test.ts
+++ b/packages/adapter-utils/src/server-utils.test.ts
@@ -11,7 +11,9 @@ import {
   buildInvocationEnvForLogs,
   DEFAULT_PAPERCLIP_AGENT_PROMPT_TEMPLATE,
   materializePaperclipSkillCopy,
+  PAPERCLIP_CREATE_AGENT_SKILL_KEY,
   refreshPaperclipWorkspaceEnvForExecution,
+  resolvePaperclipDesiredSkillNames,
   renderPaperclipWakePrompt,
   runningProcesses,
   runChildProcess,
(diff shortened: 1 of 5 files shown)


stubbi:

PR #8829, opened 2026-07-01
Add liveness-continuation backoff + per-issue upstream-throttle ceiling (shadow-mode first)

goldinbidsheets-rgb:
## Problem

Two unbounded-retry gaps in the continuation/scheduler layer (Layer C of the adapter retry-storm hardening):

1. `plan_only` / `empty_response` liveness continuations are enqueued **immediately** with no backoff.
2. The only cap on upstream-throttle exits is per-source-run (`DEFAULT_MAX_LIVENESS_CONTINUATION_ATTEMPTS = 2`, transient retry ladder max 4). Every fresh wake starts a **new** source run at attempt 0, so the caps reset and a rate-limited provider produces an indefinite retry burst — one real incident produced 9 duplicate alarms plus quota burn while the actual blocker (provider quota) was never named.

## Change (shadow-mode first, zero behavior change by default)

New dependency-free module `server/src/services/recovery/liveness-continuation-throttle.ts`, gated by `PAPERCLIP_LIVENESS_CONTINUATION_THROTTLE_MODE` = `off` | `shadow` (default) | `enforce`:

- **Backoff**: exponential (base 60s × 2^(attempt−1), ±25% jitter, cap 10m; env-tunable) computed for every liveness continuation. Shadow logs a would-defer run event and keeps today's immediate enqueue; enforce defers through the existing `scheduled_retry` primitive (`scheduleBoundedRetryForRun` gains optional `contextSnapshotExtra` + `idempotencyKey`, so the continuation instruction and the existing idempotency key carry over).
- **Per-issue rolling-window ceiling**: counts **consecutive** upstream-throttle exits (adapter error codes `codex/claude_transient_upstream`, persisted `errorFamily: transient_upstream` — including clean-exit *succeeded* runs — and the forward-compatible `upstream_throttled` liveness state) across source runs for the same issue within a rolling window (default 5 exits / 60m). At the ceiling, shadow logs would-pause; enforce pauses the agent (`pauseReason: upstream_throttle_ceiling`) and files **one** consolidated issue notice naming the real blocker (provider rate limit / quota) instead of N escalations. Any productive run resets the streak.
- **Pinned invariant**: `upstream_throttled` is never actionable for immediate continuation (`isActionableLivenessStateForContinuation` + unit test), so the Layer B classifier state can't feed the immediate-continuation path.

Invariant delivered: every retry path is bounded by backoff + a per-issue ceiling; transient 429s still retry — backed off and capped, not abandoned.

## Testing

- `liveness-continuation-throttle.test.ts` (new) + `run-continuations.test.ts` (extended): config resolution, exponential/jitter/cap bounds, throttle-exit classification truth table, streak reset-hole replay (9 exits across 9 source runs), window expiry, ceiling decisions in all three modes, backoff pass-through and back-compat default.
- All 22 behavioral cases also verified locally via Node type-stripped harness (22/22 pass).
- `tsc --strict --noEmit` clean on the new module, the decision layer, `heartbeat.ts` against its full import graph, and both test files (only pre-existing unrelated `ws`/`plugin-sdk` declaration noise elsewhere).

## Rollout

Land dark (shadow default) → observe would-defer / would-pause run events for false positives → promote to `enforce` via env. Related: #6979 / #6954 / #6957 address the adjacent process-lost retry paths; this PR covers the liveness-continuation and cross-source-run throttle-exit paths and composes with them.

### DIFF
--- a/server/src/__tests__/run-continuations.test.ts
+++ b/server/src/__tests__/run-continuations.test.ts
@@ -4,6 +4,7 @@ import {
   RUN_LIVENESS_CONTINUATION_REASON,
   buildRunLivenessContinuationIdempotencyKey,
   decideRunLivenessContinuation,
+  isActionableLivenessStateForContinuation,
 } from "../services/run-continuations.ts";
 
 const companyId = "company-1";
@@ -169,5 +170,67 @@ describe("run liveness continuations", () => {
 
       expect(decision.kind).toBe("skip");
     }
+  });
+
+  it("treats upstream_throttled as non-actionable for immediate continuation", () => {
+    // Layer B introduces the upstream_throttled liveness state for retryable
+    // upstream 429 exits; the bounded transient retry ladder and the per-issue
+    // throttle ceiling own those, so they must never immediately continue.
+    expect(isActionableLivenessStateForContinuation("upstream_throttled")).toBe(false);
+    expect(isActionableLivenessStateForContinuation("plan_only")).toBe(true);
+    expect(isActionableLivenessStateForContinuation("empty_response")).toBe(true);
+    expect(isActionableLivenessStateForContinuation(null)).toBe(false);
+
+    const decision = decideRunLivenessContinuation({
+      run: run(),
+      issue: issue(),
+      agent: agent(),
+      livenessState: "upstream_throttled" as never,
+      livenessReason: "Upstream provider throttled the run",
+      nextAction: null,
+      budgetBlocked: false,
+      idempotentWakeExists: false,
+    });
+
+    expect(decision).toEqual({
+      kind: "skip",
+      reason: "liveness state is not actionable for continuation",
+    });
+  });
+
+  it("carries the caller-computed backoff schedule on enqueue decisions", () => {
+    const dueAt = new Date("2026-07-01T12:01:00.000Z");
+    const decision = decideRunLivenessContinuation({
+      run: run(),
+      issue: issue(),
+      agent: agent(),
+      livenessState: "plan_only",
+      livenessReason: "Planned without acting",
+      nextAction: null,
+      budgetBlocked: false,
+      idempotentWakeExists: false,
+      backoff: { delayMs: 60_000, dueAt },
+    });
+
+    expect(decision.kind).toBe("enqueue");
+    if (decision.kind !== "enqueue") return;
+    expect(decision.backoff).toEqual({ delayMs: 60_000, dueAt });
+  });
+
+  it("defaults to no backoff when the caller does not provide one", () => {
+    const decision = decideRunLivenessContinuation({
+      run: run(),
+      issue: issue(),
+      agent: agent(),
+      livenessState: "empty_response",
+      livenessReason: "No useful output",
+      nextAction: null,
+      budgetBlocked: false,
+      idempotentWakeExists: false,
+    });
+
+    expect(decision.kind).toBe("enqueue");
+    if (decision.kind !== "enqueue") return;
+    expect(decision.backoff).toBeNull();
   });
 });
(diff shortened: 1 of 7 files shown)


goldinbidsheets-rgb:
**Convergence delta analysis from the GOL-8484 lane (twin implementation `a43a922`, review GOL-8758 completed PASS).** Per the board convergence decision, **this PR is the landing artifact**; the deltas below are input to the still-running review GOL-8756. Both branches implement the same approved design off the same base (`de837e2`).

## Where this PR is better than a43a922 (keep as-is)
1. **Ceiling placement (the big one):** `handleUpstreamThrottleCeiling` runs in finalization for EVERY terminal run, so it catches `failed`/`transient_upstream` exits (Layer A output) — the dominant storm path, which never reaches the liveness-continuation handler. It also composes correctly with continuation suppression via ordering (pause → agent-status guard in `decideRunLivenessContinuation` skips).
2. ±25% **jitter** matching the platform's `BOUNDED_TRANSIENT_HEARTBEAT_RETRY_JITTER_RATIO` convention.
3. **Idempotency-key threading** into `scheduleBoundedRetryForRun` for the deferred wake.
4. Config snapshot at module load (platform convention); `resultJson` string-parse tolerance; `createdAt` fallback timestamp; errorFamily set accepting `upstream_throttled`.

## Deltas a43a922 has over this PR — recommend folding
- **D-1 (must-fix before merge, one-line):** enforce-mode continuation drop. When `scheduleBoundedRetryForRun` returns non-`scheduled` (`retry_exhausted` on a mixed retry chain, invokability gate), this PR logs a warn and `return`s — dropping a decision-layer-approved continuation. a43a922 falls through to today's immediate `enqueueWakeup` in that branch, and review GOL-8758 verified "enforce never silently drops an approved continuation" as an invariant. Fix: replace the `return` in the non-scheduled branch with fall-through to the immediate enqueue.
- **D-2 (advisory, = GOL-8758 F-1, applies more here):** `handleUpstreamThrottleCeiling` runs unguarded on every terminal throttle exit; a DB throw inside it breaks run finalization even in shadow. Wrap the handler body (and the shadow would-defer event) in try/catch so shadow is airtight zero-behavior-change.
- **D-3 (minor, = GOL-8758 F-4):** cancelled runs neither count toward nor break a streak (SQL excludes them; pure fn doesn't status-filter). Defensible — add one comment line, and optionally a terminal-status guard inside `summarizeUpstreamThrottleStreak` for defense-in-depth.
- **D-4 (nit, both impls, = GOL-8758 F-2):** notice-once is SELECT-then-INSERT; racing finalizations can double-post the notice. Harmless (duplicate comment, not duplicate pause).

## Carries over from GOL-8758 regardless
**F-5 preconditions bind the future enforce flip, not this land:** CI vitest green; shadow-phase run-event data reviewed (streak attribution sane; no would-pause events on healthy issues); the flip re-enters the high-risk lane with fresh operator sign-off. **F-3:** whoever flips enforce owns monitoring for `pauseReason: upstream_throttle_ceiling` agents (no auto-resume by design).

I hold push access to this branch and will apply D-1..D-3 in one commit on request from the GOL-8756 reviewer.

goldinbidsheets-rgb:
Delta fixes applied in `89b801c` (no waiting on a CHANGES-REQ round trip): **D-1** deferral failure now falls through to the immediate enqueue (enforce never drops an approved continuation), **D-2** try/catch guards around the ceiling handler call site and the shadow would-defer event (finalization can never break on instrumentation), **D-3** cancelled runs explicitly neither extend nor break a streak (guard + comment + test), **D-5** (new finding) the colocated test file imported with a `.ts` extension which errors under the server tsconfig — switched to `.js` like the other colocated suites.

Verified: 12/12 module assertions on the modified branch (incl. replay of this PR's own streak tests + new cancelled-runs case) via Node type-stripping harness; full server tsc error count back to the 1204 pre-existing baseline (the `.ts`-extension import had made it 1205; zero errors in touched files). The keep-as-is list from my earlier comment stands. GOL-8756 review can proceed against `89b801c`.

goldinbidsheets-rgb:
Superseded by #11126. The replacement ports the reviewed Layer C implementation and follow-up safety fixes onto current `master` without rewriting this stale branch.

PR #8853, opened 2026-07-02
fix(sandbox-bridge): allow the agent-hire skill's routes through the callback bridge

stubbi:
## Problem

Managed agents that run inside a sandbox reach the server only through the sandbox callback bridge, which forwards a fixed route allowlist (`DEFAULT_SANDBOX_CALLBACK_BRIDGE_ROUTE_ALLOWLIST`). The `paperclip-create-agent` skill is documented to call adapter/icon discovery endpoints, compare existing agent configs, submit the hire, and link the resulting approval to its source issue -- but none of those routes were on the allowlist, so an agent following the skill hit `Route not allowed` on every call, including the hire `POST` itself. A correctly-behaving agent therefore could not hire.

## Fix

Add the six routes the skill uses:
- `GET /llms/agent-configuration.txt`
- `GET /llms/agent-configuration/:adapterType.txt`
- `GET /llms/agent-icons.txt`
- `GET /api/companies/:id/agent-configurations`
- `POST /api/companies/:id/agent-hires`
- `POST /api/issues/:id/approvals`

Direct agent creation (`POST /api/companies/:id/agents`) stays **denied** -- hires go through the approval-gated `agent-hires` endpoint, which the server still permission-checks (`canCreateAgents`). The bridge allowlist bounds surface area; it does not replace server-side authorization.

New tests assert the routes are allowed and that the regexes do not over-match (no `/agents`, only `.txt` discovery files, no `agent-hires` sub-resources).

### DIFF
--- a/packages/adapter-utils/src/sandbox-callback-bridge.test.ts
+++ b/packages/adapter-utils/src/sandbox-callback-bridge.test.ts
@@ -881,6 +881,13 @@ describe("sandbox callback bridge", () => {
       { method: "GET", path: "/api/companies/co-1/approvals" },
       { method: "GET", path: "/api/companies/co-1/routines" },
       { method: "GET", path: "/api/companies/co-1/skills" },
+      // Hire skill (paperclip-create-agent): discovery + submit + issue linking
+      { method: "GET", path: "/llms/agent-configuration.txt" },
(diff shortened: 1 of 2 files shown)


stubbi:

PR #8857, opened 2026-07-02
Guard duplicate subscription homes across active agents

PaulS-H:
## Summary

- Add a subscription-home guard for plain `CODEX_HOME`, `CLAUDE_CONFIG_DIR`, and `HOME` adapter env values, including path normalization and cross-key collisions.
- Reject agent hire/create/patch persistence when another active agent in the same company binds the same subscription home, unless `adapterConfig.allowSharedSubscriptionHome: true` explicitly acknowledges intentional sharing.
- Warn non-fatally at heartbeat run start for pre-existing duplicate homes, while skipping secret-resolved env keys so secret-derived values are not logged.

## Verification

- `pnpm exec vitest run server/src/__tests__/subscription-home-guard.test.ts` — 1 file, 11 tests passed.
- `pnpm exec vitest run server/src/__tests__/agent-adapter-validation-routes.test.ts server/src/__tests__/heartbeat-project-env.test.ts server/src/__tests__/heartbeat-workspace-session.test.ts` — 3 files, 124 tests passed.
- `pnpm --filter @paperclipai/server typecheck` — exited 0.

### DIFF
--- a/server/src/__tests__/subscription-home-guard.test.ts
+++ b/server/src/__tests__/subscription-home-guard.test.ts
@@ -1,0 +1,153 @@
+import { describe, expect, it } from "vitest";
+import os from "node:os";
+import path from "node:path";
+import {
+  allowsSharedSubscriptionHome,
+  findSubscriptionHomeConflicts,
+  listSubscriptionHomeBindings,
+  normalizeSubscriptionHomePath,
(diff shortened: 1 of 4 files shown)


PaulS-H:
Superseded by #8858, which uses a sanitized public branch name and the full PR template.

PR #8859, opened 2026-07-02
fix(agent-runtime): ship ripgrep in the base image

stubbi:
## What

Add `ripgrep` to the apt install in `docker/agent-runtime/Dockerfile.base`.

## Why

OpenCode's skill tooling shells out to ripgrep. When `rg` is not on PATH it tries to download a pinned build from github.com at run time. In a sandbox with locked-down egress that download hangs ~127s and then fails (`Transport error ... BurntSushi/ripgrep/releases/download/...`), burning run budget on every agent run before the agent reaches its actual work.

The root repo `Dockerfile` already installs ripgrep; the agent-runtime base image drifted without it. This restores parity so OpenCode uses the system binary and never reaches for the network.

## Change

One line: add `ripgrep` to the existing `apt-get install --no-install-recommends` list, plus an explanatory comment. No behavior change beyond shipping the binary.


stubbi:

PR #8860, opened 2026-07-02
fix(sandbox-bridge): forward Idempotency-Key + faithful size limit and error mapping

stubbi:
## What

Three faithfulness fixes to the sandbox callback bridge so documented agent flows dedupe correctly and fail transparently instead of opaquely.

### 1. Forward `Idempotency-Key` through the bridge

The header allowlist (`DEFAULT_SANDBOX_CALLBACK_BRIDGE_HEADER_ALLOWLIST`) stripped every header except `accept`/`content-type`/`if-match`/`if-none-match`. Agents are documented to send an `Idempotency-Key` on mutating calls (e.g. `POST /api/routines/:id/run`) so that a retry after a mid-flight failure dedupes server-side instead of executing twice. With the header stripped, a retried hire or routine run could double-execute. This adds `idempotency-key` to the allowlist. `authorization` and `x-paperclip-run-id` remain host-injected, not agent-supplied.

### 2. Body-size cap 256KB -> 1MB

Issue documents (plans/specs) allow a 512KB body (`upsertIssueDocumentSchema: z.string().max(524288)`). A PUT wraps that in a JSON envelope and a GET returns it with metadata, so the old 256KB cap rejected real document reads/writes with an opaque 502. 1MB covers a max document plus its JSON envelope with margin; the queue protocol already base64-chunks large bodies.

### 3. Host-proxy error mapping

A fetch timeout or oversized response previously threw into the worker's generic 502 with a vague message. Now:

- **AbortSignal timeout -> 504** `bridge_upstream_timeout`, with a clear "may or may not have applied; re-read state before retrying" body. A timeout on a mutating call is ambiguous, and an agent that can't tell a timeout from a failure tends to confabulate an outcome.
- **Oversized response body -> 413** `bridge_response_too_large` (the call succeeded server-side, the payload was too large to relay) instead of implying the operation failed.
- **Other unreachable errors -> 502** `bridge_upstream_unreachable`, with the underlying message.

## Tests

`packages/adapter-utils/src/sandbox-callback-bridge.test.ts` asserts `idempotency-key` now reaches the handler while host-injected headers stay stripped.

- `npx tsc -p packages/adapter-utils/tsconfig.json --noEmit` -> exit 0
- `npx vitest run packages/adapter-utils/src/sandbox-callback-bridge.test.ts` -> 13 passed

### DIFF
--- a/packages/adapter-utils/src/execution-target.ts
+++ b/packages/adapter-utils/src/execution-target.ts
@@ -1155,22 +1155,67 @@ export async function startAdapterExecutionTargetPaperclipBridge(input: {
         }
         headers.set("authorization", `Bearer ${hostApiToken}`);
         headers.set("x-paperclip-run-id", input.runId);
-        const response = await fetch(buildBridgeForwardUrl(hostApiUrl, request), {
-          method,
-          headers,
-          ...(method === "GET" || method === "HEAD" ? {} : { body: request.body }),
-          signal: AbortSignal.timeout(30_000),
-        });
+        let response: Response;
+        try {
+          response = await fetch(buildBridgeForwardUrl(hostApiUrl, request), {
+            method,
+            headers,
+            ...(method === "GET" || method === "HEAD" ? {} : { body: request.body }),
+            signal: AbortSignal.timeout(30_000),
+          });
+        } catch (error) {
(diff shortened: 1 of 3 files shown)


stubbi:

PR #8863, opened 2026-07-02
feat(company-skills): capability-gate agent-executable skill import

soydachi:
## Thinking Path

> - Company skills are the reusable capability units an agent imports and runs; their mutation routes are reached through the `queue_v1` sandbox-callback bridge allowlist.
> - Today every company-skill mutation route returns `403 "Route not allowed"` for any agent JWT, because `POST /api/companies/:id/skills/import` is absent from that allowlist.
> - Merely allow-listing the transport would be unsafe: the existing import authz (`assertCanMutateCompanySkills`) is default-ALLOW for agents, so the route would open to ~every agent.
> - `import` introduces new code into the resident-fleet supply chain, so it warrants a stricter gate than the other mutations; this PR ships a transport allowlist entry together with a new default-DENY capability gate so the two can never drift apart.
> - The benefit is a governed, agent-executable path to import a company skill (the prerequisite for rolling a shared recall skill to the resident fleet) without opening a privilege hole.

No public GitHub issue exists in `paperclipai/paperclip` for this work; the underlying issue is described inline below (feature-template fields). Tracking lives in an internal instance board and is intentionally not linked here per CONTRIBUTING → "No Internal Issue References".

**Problem or motivation**
Every company-skill mutation route (`import`, `create`, …) returns `403 "Route not allowed"` for agent JWTs because the import route is not in the `queue_v1` bridge allowlist. There is no capability-gated path for a manager agent to import a company skill.

**Proposed solution**
Add `POST /api/companies/:id/skills/import` to the `queue_v1` allowlist (transport), and gate the handler behind a new **default-DENY** `assertAgentCanImportSkills` that requires an explicit `skills:create` grant (authz). The two layers ship atomically so the allowlist entry is never reachable without the capability check.

**Alternatives considered**
Allow-listing the transport alone (rejected: default-ALLOW authz would expose import to ~every agent). Opening `create` as well (rejected: least-privilege — only `import` is needed now). Reusing `assertCanMutateCompanySkills` (rejected: it is default-ALLOW; `import` warrants default-DENY).

**Roadmap alignment**
Infrastructure/governance hardening of the skills subsystem; not a core roadmap feature. No overlap with planned core work.

## What Changed

- **Transport** — add `POST /api/companies/:id/skills/import` to the `queue_v1` route allowlist (`packages/adapter-utils/src/sandbox-callback-bridge.ts`). `create` deliberately stays out (least-privilege).
- **Authz** — new default-DENY helper `assertAgentCanImportSkills`: an agent may import **only** with an explicit `skills:create` grant. Board actors keep their existing semantics; other mutations remain unchanged (still default-ALLOW).
- **Provenance** — the import handler logs `{callerAgentId, companyId, sourceLocator, contentHash, runId}` per import for per-operation auditability.
- **Tests** — allow+provenance, deny-shape guard (asserts `403 Missing permission: skills:create`, not `Route not allowed`, so allowlist drift is caught), and a re-import idempotency test. Two pre-existing agent tests that exercised the shared default-ALLOW path via `import` were repointed to `create`.
- **Test fix** — scope the idempotency assertion to the imported skill's own key: `importFromSource` materializes the bundled-skill inventory into `companySkills` as a side effect, so the table also holds bundled rows; the assertion now checks the imported skill's row count and that the total is unchanged across a second identical import, instead of assuming the whole table holds a single row.

## Verification

- CI is the intended verification gate (no monorepo install in the authoring sandbox).
- `policy` check: ✅ SUCCESS. Build / Typecheck / e2e / serialized server suites: ✅.
- Independent security review of the exact authz code was performed by a separate identity from the author before publication (governance record ADR-GOV-0002).
- The earlier red `General tests (server 2/3)` was the idempotency test asserting `expect(rows.length).toBe(1)` while `importFromSource` also seeds 5 bundled skills (6 rows total). Root-caused as a **test-assertion bug, not a production regression** — the upsert is already idempotent (stable key → UPDATE, not INSERT). Fixed in this push; awaiting the re-run to confirm green.
- Manual authz shape: agent without the grant → `403 Missing permission: skills:create`; agent with the grant → `2xx` import; `POST /agents/:id/skills/sync` behavior unchanged.

## Risks

- Opens a mutation route into the skills library → mitigated by the default-DENY `skills:create` capability gate plus the independent security review. Without the capability the behavior is identical to today (`403`), so there is no regression for non-manager agents.
- The allowlist entry is a *transport* gate, not an *authorization* gate; the two changes must never ship separately. Enforced by the deny-shape test, which fails if a future allowlist change skips the capability layer.
- Test-only follow-up commit carries no production behavior change.

## Model Used

Claude Opus 4.8 (`claude-opus-4-8`), extended thinking enabled, via the Claude Code / Paperclip `claude_local` adapter with tool use and code execution.

---

- [x] I searched the GitHub PR list (open + recently closed) for similar PRs and confirmed this is not a duplicate.

### DIFF
--- a/packages/adapter-utils/src/sandbox-callback-bridge.ts
+++ b/packages/adapter-utils/src/sandbox-callback-bridge.ts
@@ -51,6 +51,13 @@ export const DEFAULT_SANDBOX_CALLBACK_BRIDGE_ROUTE_ALLOWLIST: readonly SandboxCa
   { method: "GET", path: /^\/api\/companies\/[^/]+\/approvals$/ },
   { method: "GET", path: /^\/api\/companies\/[^/]+\/routines$/ },
   { method: "GET", path: /^\/api\/companies\/[^/]+\/skills$/ },
+  // ARC-263 / ADR-GOV-0002: agent-executable company-skill import. This is a TRANSPORT
+  // gate only; the handler enforces an explicit gov-matrix `skills:create` grant that is
+  // DEFAULT-DENY for agents. The allowlist entry and the capability check MUST ship in the
+  // same change (CSO condition C1) — never allow-list this route without the handler grant,
+  // that would be a privilege hole. Import only (create stays out): import introduces new
+  // code into the resident-fleet supply chain and is strictly more sensitive.
+  { method: "POST", path: /^\/api\/companies\/[^/]+\/skills\/import$/ },
   { method: "GET", path: /^\/api\/projects\/[^/]+$/ },
   { method: "GET", path: /^\/api\/goals\/[^/]+$/ },
 
--- a/server/src/__tests__/company-skills-routes.test.ts
+++ b/server/src/__tests__/company-skills-routes.test.ts
@@ -871,7 +871,12 @@ describe("company skill mutation permissions", () => {
     });
   });
 
-  it("allows agents with canCreateSkills to mutate company skills", async () => {
+  // ARC-263 / ADR-GOV-0002 (C3): the permissive canCreateSkills default still governs the
+  // non-import mutation routes (create/install/update/…). These two cases were previously
+  // asserted against the import route; they are repointed to `create` because import is now
+  // default-DENY (its dedicated coverage is below). The general default-ALLOW behavior for
+  // agents is unchanged for every route except import.
+  it("allows agents with canCreateSkills to create company skills", async () => {
     mockAgentService.getById.mockResolvedValue({
       id: "agent-1",
       companyId: "company-1",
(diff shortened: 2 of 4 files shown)


soydachi:
**Author update (IT-Admin / @soydachi) — PR now CLEAN, ready for human merge review.**

Two things were blocking this PR; only one was PR hygiene:

1. **`review` (commitperclip) — hygiene.** Rewrote the PR body to satisfy the required template sections (Thinking Path / What Changed / Verification / Risks / Model Used), an inline issue description (no public GH issue exists; internal refs intentionally omitted per CONTRIBUTING), and the dedup-search checkbox. The review workflow only runs on `opened/synchronize/reopened`, so a fresh push (`adcc755`) was needed for it to re-read the body. ✅

2. **`General tests (server 2/3)` → `verify` — a real red, not hygiene.** The idempotency test (`company-skills-service.test.ts`) asserted the whole `companySkills` table held exactly one row after import (`expected 6 to be 1`). Root cause: `importFromSource` materializes the **bundled-skill inventory** (5 rows) into `companySkills` as a side effect, so the table holds `5 bundled + 1 imported`. The upsert is **already idempotent** (stable key → UPDATE, not INSERT) — a second identical import kept the count at 6, not 12. Fixed the **test assertion** (commit `3ecaf42`) to scope to the imported skill's own key and assert the total is unchanged across re-import, preserving (not weakening) the real invariant. Production authz path untouched. ✅

All checks green now; `mergeStateStatus: CLEAN`. Governance unchanged: I'm the PR **author**, not the merger — the merge remains the human gate (reviewer ≠ builder; the C-security review of the authz code stands, as this follow-up was test-only + PR body).

soydachi:
**Author triage note (IT-Admin / @soydachi) — merge-ready; one advisory triaged.**

Status: PR is `MERGEABLE` / `mergeStateStatus: CLEAN`, all required checks green, and it carries a separate-identity security-review approval (reviewer ≠ builder).

Re Greptile's 4/5 advisory (`contentHash: null` for local-path and raw HTTP(S) **unpinned** imports): **acknowledged, and triaged as a net-additive follow-up hardening — not a blocker for this PR.** Rationale:

1. The finding is about audit-provenance *completeness for unpinned/mutable sources*, which is orthogonal to this PR's acceptance scope: authz allow/deny, the allowlist transport guard (deny ≠ "Route not allowed"), and import idempotency.
2. The capability gate itself is scoped to the explicit grant — Greptile confirms this ("appears scoped to the intended explicit grant").
3. The provenance entry already records `sourceRef` where the source provides one (pinned refs). Deriving a byte-level hash for unpinned sources is strictly additive and can land as a follow-up without weakening anything here.

Ready for human merge review.

Issue #8019, opened 2026-06-12
fix(auth): reject local_trusted write fallthrough before local-board attribution


devinfoley:
## Summary

In current `master`, `server/src/middleware/auth.ts` still seeds `local_trusted` requests with the implicit `local-board` actor and then falls through on unauthenticated writes. A plain missing `Authorization` header is one path, but the same misattribution still happens when clients send `Authorization: Bearer ` or any invalid bearer token.

## Problem

API-style write requests in `local_trusted` mode can still be recorded as board actions when no real actor was resolved.

That breaks attribution and auditability for agent heartbeats and scripts: instead of failing fast, the request is accepted under the implicit board actor.

## Current upstream behavior

`server/src/middleware/auth.ts` on `master` still does all of the following:

- seeds `req.actor` with `{ type: "board", source: "local_implicit" }` in `local_trusted`
- calls `next()` when the bearer token is empty
- calls `next()` when a bearer token is present but invalid / unrecognized
- treats `Referer` as client input rather than a trustworthy browser-only signal

## Expected behavior

Write requests should fail before any board attribution when no authenticated actor was established.

A focused fix should:

- reject unauthenticated `POST` / `PATCH` / `PUT` / `DELETE` requests in `local_trusted`
- cover missing headers, empty bearer tokens, and invalid bearer tokens
- avoid using spoofable `Referer` values as the bypass condition
- add regression tests around the `server/src/middleware/auth.ts` branches that currently fall through

## Related context

PR #7763 surfaced this bug, but the proposed guard is still bypassed by empty/invalid bearer tokens and by a spoofed `Referer` header. Closing that PR in favor of this focused tracking issue keeps the bug visible without carrying the stale branch forward.

contentfree:
Would love to see this fixed. Hate seeing comments misattributed to me!

nickhaggmark-scosquad:
Still present in `2026.722.0` (verified today). Adding evidence from a self-hosted `local_trusted` fleet, including a constraint that I think explains why #7763 stalled.

## 1. Confirmed in the shipped build

`actorMiddleware` in `@paperclipai/server/dist/middleware/auth.js` seeds the actor **before any authentication is attempted**:

```js
req.actor =
  opts.deploymentMode === "local_trusted"
    ? { type: "board", userId: "local-board", userName: "Local Board",
        isInstanceAdmin: true, source: "local_implicit" }
    : { type: "none", source: "none" };
```

[...]

PR #8869, opened 2026-07-02
middleware: return 401 on invalid Bearer instead of silent local-boar…

HelloWolrs:
…d fallthrough

Any `Authorization: Bearer …` header that fails every credential lookup (board API key, agent API key, agent JWT) previously fell through with the mode-default `req.actor`. In `local_trusted` mode that default is `{ type: 'board', userId: 'local-board', isInstanceAdmin: true }` — so callers presenting an invalid Bearer were silently promoted to full board access. Return 401 on every branch where a caller *attempted* Bearer auth and failed. Behavior for callers that send no Authorization header at all is intentionally preserved so existing `local_trusted` console UIs keep working; a follow-up conversation about opt-in strict mode is in the PR body.

Closes five silent-fallthrough branches in server/src/middleware/auth.ts:
  1. Empty token after `Bearer `.
  2. Bearer present but no board key, no agent-API-key row, invalid JWT.
  3. JWT verified but agent row missing / company_id mismatch.
  4. JWT verified but agent status is terminated or pending_approval.
  5. Agent-API-key row valid but agent status terminated or pending_approval.

All five now return `401 { error: 'Unauthorized',
reason: 'invalid_or_missing_bearer' }` — uniform reason string on purpose (no leaking which internal lookup failed to a caller probing tokens).

Tests: 5 new vitest cases in server/src/middleware/auth.test.ts, mirroring the fake-db style of cloud-tenant-actor.test.ts in the same directory.

## Thinking Path

> - [Which subsystem or capability is involved]
> - [What problem or gap exists]
> - [Why it needs to be addressed]
> - This pull request ...
> - The benefit is ...

-

## What Changed

-

## Verification

-

## Risks

-

## Model Used

-

## Checklist

- [ ] I have included a thinking path that traces from project context to this change
- [ ] I have specified the model used (with version and capability details)
- [ ] I have searched GitHub for duplicate or related PRs and linked them above
- [ ] I have either (a) linked existing issues with `Fixes: #` / `Closes #` / `Refs #` OR (b) described the issue in-PR following the relevant issue template
- [ ] I have not referenced internal/instance-local Paperclip issues or links (only public GitHub `#NNN` / `github.com/paperclipai/paperclip` URLs)
- [ ] I have considered and documented any risks above

### DIFF
--- a/server/src/middleware/auth.test.ts
+++ b/server/src/middleware/auth.test.ts
@@ -1,0 +1,111 @@
+import { describe, expect, it, vi } from "vitest";
+import type { NextFunction, Request, Response } from "express";
+import type { Db } from "@paperclipai/db";
+import { actorMiddleware } from "./auth.js";
+
+// Fake db returning an empty result for every SELECT chain used by
+// actorMiddleware. The chained calls (`.select().from().where().then(fn)`)
+// and (`.update().set().where()`) both resolve to `[]` / `undefined`.
+function createEmptyDb(): Db {
+  const chain: Record<string, unknown> = {};
+  chain.from = () => chain;
+  chain.where = () => chain;
+  chain.set = () => chain;
+  chain.then = (resolve: (v: unknown) => unknown) => Promise.resolve([]).then(resolve);
+  const db = {
+    select: () => chain,
+    update: () => chain,
+  } as unknown as Db;
+  return db;
+}
+
+function fakeReq(headers: Record<string, string>): Request {
+  const lower: Record<string, string> = {};
+  for (const [k, v] of Object.entries(headers)) lower[k.toLowerCase()] = v;
+  return {
+    header: (name: string) => lower[name.toLowerCase()],
+    method: "POST",
+    originalUrl: "/api/companies/x/issues/y/comments",
+  } as unknown as Request;
+}
+
+function fakeRes() {
+  const state: { statusCode?: number; body?: unknown } = {};
+  const res = {
+    status(code: number) {
+      state.statusCode = code;
+      return res;
+    },
+    json(body: unknown) {
+      state.body = body;
+      return res;
+    },
+  } as unknown as Response;
+  return { res, state };
+}
+
+const LOCAL_TRUSTED = { deploymentMode: "local_trusted" as const };
+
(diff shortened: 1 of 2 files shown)


gravitydrix-cld:
Independent corroboration plus one small addition, in case it helps get this over the line.

I hit this exact issue operating a `local_trusted` deployment where agents authenticate with per-run JWTs. When a run JWT intermittently fails to resolve (expiry mid-run, or the `agents` lookup transiently failing under connection-pool pressure), the request falls through to the implicit `local-board` instance-admin actor. The damage is attribution, not just theory: an agent's own `{comment, status}` heartbeat lands as `authorUserId: local-board`, so the status write is recorded as a board action, and on an issue with a deferred comment wake that "user" comment can auto-reopen the issue the agent just closed. Failing closed, as this PR does, makes the failure observable instead of silently mis-attributed. This is also the behavior requested in #8019.

The change and its test matrix look right to me. Two small, optional notes for reviewers:

1. The reject-path coverage here is thorough. The one case I would add is a positive assertion that a *valid* agent JWT still resolves to the agent actor in `local_trusted` mode, so the fix is pinned as "reject the bad, keep the good" rather than just "reject":

```ts
it("still resolves a valid agent JWT to the agent actor in local_trusted mode", async () => {
  // mint a valid run JWT for an active agent, send it with a matching
  // X-Paperclip-Run-Id, and expect req.actor.type === "agent" (not a 401)
});
```

2. `middleware/auth.ts` already fails closed on its two newer branches via `next(<HttpError>)`: the `X-Paperclip-Run-Id` mismatch returns `unprocessable` (422) and the missing-responsible-user key returns `forbidden` (403), both flowing through `errorHandler`. Routing the invalid-bearer rejections through `next(unauthorized(...))` the same way, rather than writing the response directly, would keep the middleware uniform and let each branch carry a distinct `code`. Purely a consistency nit; the current behavior is correct either way.

For what it is worth, the only thing blocking this looks like the empty required template sections (Thinking Path / What Changed / Verification), which is what the `review` check is failing on. Filling those in should unblock it. Happy to help if useful.

contentfree:
@HelloWolrs Please add the missing information so this can get fixed.

PR #8952, opened 2026-07-03
feat: support PAPERCLIP_ADAPTER_PLUGINS_PATH env override for GitOps

### LINKED ISSUE #8949: feat: support ConfigMap-mounted adapter-plugins.json for GitOps/K8s deployments
## Problem

`adapter-plugin-store.ts` reads `adapter-plugins.json` from the PVC filesystem at `~/.paperclip/adapter-plugins.json`. In Kubernetes/OpenShift GitOps deployments (ArgoCD, Flux), configuration should be declarative via ConfigMaps — not files manually written to a PVC.

Currently, the only way to register an external adapter on K8s is to exec into the pod, write the JSON file, and restart. This breaks the GitOps model where all config flows through git.

## Proposed Solution

Add an optional environment variable `PAPERCLIP_ADAPTER_PLUGINS_PATH` that overrides the default PVC path. When set, `adapter-plugin-store.ts` reads from that path instead.

This allows mounting a ConfigMap:

```yaml
apiVersion: v1
kind: ConfigMap
metadata:
  name: paperclip-adapter-plugins
data:
  adapter-plugins.json: |
    [
      {
        "packageName": "my-adapter",
        "localPath": "/paperclip/adapters/my-adapter",
        "type": "my_adapter_type",
        "installedAt": "2026-07-01T00:00:00.000Z"
      }
    ]
```

## Implementation

~10 lines in `server/src/services/adapter-plugin-store.ts`:

```typescript
function adapterPluginPaths() {
  const paperclipDir = resolvePaperclipHomeDir();
  const overridePath = process.env.PAPERCLIP_ADAPTER_PLUGINS_PATH;
  return {
    adapterPluginsDir: path.join(paperclipDir, "adapter-plugins"),
    adapterPluginsStorePath: overridePath || path.join(paperclipDir, "adapter-plugins.json"),
    adapterSettingsPath: path.join(paperclipDir, "adapter-settings.json"),
  };
}
```

## Context

- Discovered while deploying Paperclip on OpenShift with ArgoCD GitOps
- The existing `PAPERCLIP_EXECUTION_MODE=kubernetes` and `PAPERCLIP_ADAPTERS` env patterns set the precedent for env-based config
- Backwards compatible: existing PVC-based workflows work unchanged when the env var is unset

rrbanda:
## Thinking Path

Deployed Paperclip on OpenShift with ArgoCD GitOps. External adapters need `adapter-plugins.json` registered, but the only path is writing to the PVC filesystem manually. Saw that `PAPERCLIP_EXECUTION_MODE` and `PAPERCLIP_ADAPTERS` already use env-based config patterns. Applied the same pattern to the adapter plugin store path.

## What Changed

- `server/src/services/adapter-plugin-store.ts`: `adapterPluginPaths()` now checks `process.env.PAPERCLIP_ADAPTER_PLUGINS_PATH` before falling back to the default `~/.paperclip/adapter-plugins.json` path
- Updated module doc comment to mention the env var

## Verification

1. Without env var: behavior unchanged (reads from PVC path)
2. With `PAPERCLIP_ADAPTER_PLUGINS_PATH=/etc/paperclip/adapter-plugins.json`: reads from ConfigMap-mounted path
3. Test: deploy on OpenShift with ConfigMap mount, verify external adapter loads at startup

## Risks

- None — purely additive, backwards compatible
- The env var only affects the read path; writes still go to the PVC (appropriate since ConfigMap-mounted paths are read-only)

## Model Used

Claude Opus 4.6 (Cursor IDE agent)

## Checklist

- [x] Behavior matches spec (env override with fallback)
- [x] Backwards compatible
- [x] No new dependencies
- [x] Docs updated (module comment)

Closes #8949

### DIFF
--- a/server/src/services/adapter-plugin-store.ts
+++ b/server/src/services/adapter-plugin-store.ts
@@ -5,6 +5,10 @@
  * ~/.paperclip/adapter-plugins.json. This is the source of truth for which
  * external adapters should be loaded at startup.
  *
+ * The store path can be overridden with the `PAPERCLIP_ADAPTER_PLUGINS_PATH`
+ * environment variable to support ConfigMap-mounted paths in Kubernetes/GitOps
+ * deployments.
+ *
  * Both the plugin store and the settings store are cached in memory after
  * the first read. Writes invalidate the cache so the next read picks up
  * the new state without a redundant disk round-trip.
@@ -45,9 +49,10 @@ interface AdapterSettings {
 
 function adapterPluginPaths() {
   const paperclipDir = resolvePaperclipHomeDir();
+  const storePath = process.env.PAPERCLIP_ADAPTER_PLUGINS_PATH;
   return {
     adapterPluginsDir: path.join(paperclipDir, "adapter-plugins"),
-    adapterPluginsStorePath: path.join(paperclipDir, "adapter-plugins.json"),
+    adapterPluginsStorePath: storePath || path.join(paperclipDir, "adapter-plugins.json"),
     adapterSettingsPath: path.join(paperclipDir, "adapter-settings.json"),
   };
 }


rrbanda:
cc @cryppadotta @devinfoley — small change (1 file, +6 lines). Adds env var override for adapter-plugins.json path to support ConfigMap-mounted config in K8s/GitOps deployments. All CI checks pass. Happy to address any feedback.

PR #8981, opened 2026-07-04
fix(skills): make paperclip-create-agent available to agents that can hire

stubbi:
## Thinking Path

> - Agents run inside sandboxes, and adapters (here: the managed-default opencode adapter) mount a set of Paperclip skills into each run so the agent knows how to use the control plane
> - Agents whose permissions include `canCreateAgents` (e.g. a lead/CEO agent) are supposed to be able to hire, and the hire flow is documented in the `paperclip-create-agent` skill's SKILL.md (`POST /api/companies/$PAPERCLIP_COMPANY_ID/agent-hires` with the injected `$PAPERCLIP_API_KEY`/`$PAPERCLIP_API_URL`)
> - But the skills dir is built from `resolvePaperclipDesiredSkillNames(config, availableEntries)`, which returns `[]` unless the agent has an explicit `adapterConfig.paperclipSkillSync.desiredSkills`; managed agents (including a seeded CEO) have none, so zero skills are synced and a hire-capable agent has no hire instructions
> - This pull request resolves the create-agent skill at run time: `resolvePaperclipDesiredSkillNames` gains an opt-in `alwaysIncludeSkillKeys` option, and the opencode adapter passes `paperclip-create-agent` through it whenever the agent's permissions include `canCreateAgents`
> - The benefit is that any agent that can hire actually receives the hire-flow skill — for both already-existing agents and newly-created ones, with no data backfill and no behavior change for agents that cannot hire

No public issue exists for this; describing the bug in-PR (bug template fields):

**Describe the bug**
Agents whose permissions include `canCreateAgents` cannot actually hire, because the `paperclip-create-agent` skill is never mounted into their sandbox run. The skill's `SKILL.md` is what tells the agent how to submit a hire, so without it a capable agent has no hire-flow instructions and reports it "cannot access the paperclip-create-agent skill".

**Steps to reproduce**
1. Run a company with a managed (e.g. seeded CEO) agent that has `canCreateAgents` and no explicit `adapterConfig.paperclipSkillSync.desiredSkills`
2. Ask it to hire a new agent
3. Observe run logs: `Syncing skills to sandbox: 100% (0.0/0.0 MB)` — zero skills synced — and the agent cannot find `paperclip-create-agent`

**Expected behavior**
An agent that is permitted to hire has the create-agent skill available in its run.

**Root cause**
At run time the opencode adapter builds the sandbox skills dir via `resolvePaperclipDesiredSkillNames(config, availableEntries)`, which returns `[]` unless the agent has an explicit `desiredSkills` preference (`preference.explicit`). Managed agents have none, so no skills are synced.

Related PRs: Supersedes #8823. Sibling fix #8978 (allowing the agent-hire skill's routes through the sandbox callback bridge; supersedes #8853); the two are independently mergeable, but both are needed for the end-to-end hire flow from a sandbox.

## What Changed

- `resolvePaperclipDesiredSkillNames(config, availableEntries, opts?)` gains an optional `opts.alwaysIncludeSkillKeys: string[]`. Those keys are unioned into the result (canonicalized the same way as explicit desired skills, and deduped) even when there is no explicit `desiredSkills` preference. Behavior is unchanged when the option is absent, and a requested skill is only included when it actually exists in `availableEntries` (no phantom entries). New exported constant `PAPERCLIP_CREATE_AGENT_SKILL_KEY`.
- The opencode adapter threads the agent's `canCreateAgents` into both skill-dir build paths (`buildOpenCodeSkillsDir` for the remote/sandbox path, and the local injection path), including `paperclip-create-agent` when the agent can hire.
- `AdapterAgent` gains an optional `permissions` field so adapters can read `canCreateAgents`.
- Other local adapters can opt in cheaply via the same `alwaysIncludeSkillKeys` option (follow-up); this PR wires the managed-default opencode adapter.

## Verification

- `npx vitest run packages/adapter-utils/src/server-utils.test.ts` — 59 passed, including the new cases: returns `["paperclip-create-agent"]` with the option and no explicit desiredSkills; `[]` without the option (unchanged); `[]` when the skill is unavailable (no phantom); unions + dedupes with explicit desiredSkills.
- `npx vitest run packages/adapters/opencode-local/src/server/execute.test.ts` — all passed, including: `buildOpenCodeSkillsDir` includes the create-agent skill when the caller passes `canCreateAgents: true` and excludes it when `false`.
- `npx tsc --noEmit` clean for `packages/adapter-utils` and `packages/adapters/opencode-local`.

## Risks

- Low risk. `alwaysIncludeSkillKeys` is opt-in and additive; when the option is absent, `resolvePaperclipDesiredSkillNames` behaves exactly as before. `AdapterAgent.permissions` is optional, so no adapter is forced to change. No migrations, no API changes.
- Behavioral shift (intended): hire-capable agents on the opencode adapter now get one extra skill mounted into their sandbox.

## Model Used

- Claude (Anthropic) — original change authored with claude-opus-4-8 (1M context, extended thinking); recreated on current master and re-verified with claude-fable-5, via Claude Code with tool use (code execution, test runs).

## Checklist

### DIFF
--- a/packages/adapter-utils/src/server-utils.test.ts
+++ b/packages/adapter-utils/src/server-utils.test.ts
@@ -11,7 +11,9 @@ import {
   buildInvocationEnvForLogs,
   DEFAULT_PAPERCLIP_AGENT_PROMPT_TEMPLATE,
   materializePaperclipSkillCopy,
+  PAPERCLIP_CREATE_AGENT_SKILL_KEY,
   refreshPaperclipWorkspaceEnvForExecution,
+  resolvePaperclipDesiredSkillNames,
   renderPaperclipWakePrompt,
   runningProcesses,
   runChildProcess,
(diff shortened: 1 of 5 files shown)


stubbi:
Rebased onto current master `d293dd3d1` (new head `3a27f3ce4`).

Master has since landed `5af49cb47` ("CEO agents get the core paperclip skills by default", #12138), which seeds `PAPERCLIP_CORE_SKILL_KEYS` — `paperclip-create-agent` included — into the desired-skill set at hire/create time. That is narrower than this PR in two ways: it gates on `role === "ceo"` rather than `permissions.canCreateAgents`, and it only applies at creation time, so an existing or non-CEO agent that can hire still resolves zero skills at runtime because `resolvePaperclipDesiredSkillNames` remains permission-blind. This PR is the runtime half and still applies; the two compose (master seeds the stored preference, this guarantees the skill is mounted regardless).

The rebase was resolved to build on master's machinery rather than sidestep it. Master also introduced `resolveLegacyPaperclipDesiredSkillNames`, which wraps the base resolver to always mount the operational skill for legacy adapters — the same "always include" shape this PR adds — and repointed the opencode call sites at it. So `alwaysIncludeSkillKeys` is now threaded *through* that legacy wrapper instead of bypassing it, and the opencode adapter keeps calling the legacy resolver; the operational-skill invariant is preserved and the create-agent skill is unioned on top. `packages/adapters/opencode-local/src/server/execute.test.ts` was rewritten upstream, so this PR's suite was appended to master's file and rebased onto its `fs` import style, plus a new case covering the pass-through via the legacy resolver.

Verified: `@paperclipai/adapter-utils`, `@paperclipai/adapter-opencode-local` and `@paperclipai/server` typecheck clean, `server-utils.test.ts` + `opencode-local/execute.test.ts` pass (118/118).

PR #8255, opened 2026-06-18
fix(plugins): move dev SDK linking out of plugin postinstall scripts

devinfoley:
## Thinking Path

> - Paperclip is the open source app people use to manage AI agents for work
> - Sandbox-provider plugins (cloudflare, daytona, e2b, exe-dev, kubernetes, modal, novita) and `plugin-workspace-diff` are published as standalone npm packages, but during local dev they need the in-repo `@paperclipai/plugin-sdk` symlinked in
> - Each of these plugins shipped a `postinstall` lifecycle script that traversed *out* of its own package directory (`node ../../../../scripts/link-plugin-dev-sdk.mjs`) to do that linking
> - The publishable manifest is built by a `prepack` whitelist that drops the `scripts` field, so npm consumers don't see the postinstall today — but that safety property depends entirely on `prepack` running on every publish. A publish that skips lifecycle scripts would ship a tarball whose postinstall escapes its package directory at consumer install time
> - This pull request removes the escape-the-package-dir lifecycle script from every plugin source manifest and moves the dev linking to a single root-level postinstall that iterates the excluded plugin directories itself
> - The benefit is that plugin tarballs can no longer carry an install-time script that reaches outside their own directory, regardless of whether `prepack` runs

## Linked Issues or Issue Description

This is a follow-up hardening change flagged during review of the Novita sandbox provider PR (#7595).

**Problem (security):** Excluded plugin packages each carried `"postinstall": "node ../../../../scripts/link-plugin-dev-sdk.mjs"`. The relative path traverses outside the package root. Today the published manifest is sanitized by a `prepack` whitelist that drops `scripts`, so consumers are unaffected in the normal publish path. The risk is that this is a defense-in-depth gap: if a publish ever skips lifecycle scripts (e.g. `npm publish --ignore-scripts` is *not* used, or a tool publishes the raw manifest), the tarball would ship a postinstall that runs out-of-tree code at the consumer's install time.

## What Changed

- Added a single root `package.json` `postinstall`: `node scripts/link-plugin-dev-sdk.mjs`.
- Rewrote `scripts/link-plugin-dev-sdk.mjs` to iterate the excluded plugin directories itself (`packages/plugins/sandbox-providers/*` + the orchestration smoke example) instead of relying on each plugin to invoke it from its own cwd. Preserves both prior behaviors: leave a real installed SDK dir alone, and skip when already correctly symlinked (idempotent).
[...]

PR #8989, opened 2026-07-04
build(docker): bundle the kubernetes sandbox-provider plugin in the image

stubbi:
## Thinking Path

> - Paperclip is the open source app people use to manage AI agents for work, and it ships a production Docker image for containerized deployments
> - The kubernetes sandbox provider (`packages/plugins/sandbox-providers/kubernetes`, shipped by the k8s execution stack) is deliberately excluded from the pnpm workspace to keep `@kubernetes/client-node` out of the root lockfile — so the Docker build never installs or builds it
> - The production image therefore lacks `dist/manifest.js`, `app.ts`'s `ensureBundledKubernetesPlugin` logs "kubernetes sandbox plugin bundle not present; skipping auto-install", and any containerized deployment that sets `PAPERCLIP_EXECUTION_MODE=kubernetes` fails every run with `Sandbox provider "kubernetes" is not registered as a built-in provider and no matching plugin is available`
> - Found during a live multi-replica verification on a real cluster: the tenant namespace/RBAC machinery never got the chance to run until the image was rebuilt with this block
> - The fix installs + builds the plugin in place during the image build (mirroring `scripts/build-standalone-public-packages.mjs` semantics) so its `dist/`, `node_modules/`, and the plugin-sdk dev symlink land in `/app` and are copied into the production stage
> - The benefit is that the published container image can use the in-cluster kubernetes sandbox provider out of the box, with a build-time gate guaranteeing no image ships with a broken plugin bundle

No existing issue. Described in-PR: the published container image cannot use the in-cluster kubernetes sandbox provider because the workspace-excluded plugin is never built into it; runtime symptom is the auto-install skip log at boot and `Sandbox provider "kubernetes" is not registered` on every kubernetes-mode run.

Supersedes #8067.

## What Changed

- **`Dockerfile`** — install + build `packages/plugins/sandbox-providers/kubernetes` standalone. The install (`pnpm -C … install --ignore-workspace --frozen-lockfile`) runs in the `deps` stage against just the plugin's `package.json` + `pnpm-lock.yaml`, so the layer cache key is scoped to those two files and unrelated source changes don't re-download `@kubernetes/client-node` (Greptile round-1 P2). After the server/SDK builds in the `build` stage: re-link the in-repo `@paperclipai/plugin-sdk` via `scripts/link-plugin-dev-sdk.mjs`, then `run build`, with a hard `test -f dist/manifest.js && test -f dist/worker.js` gate so a broken plugin can never produce a deployable image. The explicit link step is needed because #8255 removed per-plugin `postinstall` SDK linking (supply-chain safety) and the standalone install does not know about the in-repo SDK; this is the same install-then-link order `scripts/build-standalone-public-packages.mjs` uses.
- **`packages/plugins/sandbox-providers/kubernetes/pnpm-lock.yaml`** — commit a plugin-local lockfile. This package is intentionally outside the pnpm workspace, so the root `pnpm-lock.yaml` cannot cover it; without its own lockfile the Docker build would have to install with `--no-lockfile`, making the embedded `@kubernetes/client-node` dependency tree float between builds. The committed lockfile plus `--frozen-lockfile` keeps image builds reproducible and makes dependency changes reviewable. This is the one deliberate exception to "no lockfiles in PRs": that rule exists for the root workspace lockfile (refresh-bot owns it), which by design can never include this package.
- **`packages/plugins/sandbox-providers/kubernetes/.gitignore`** — stop ignoring `pnpm-lock.yaml` (the previous ignore existed precisely because the package was built with `--no-lockfile`; that rationale is now inverted).
- **`packages/plugins/sandbox-providers/kubernetes/package.json`** — add a `pnpm.overrides` entry pinning `esbuild` to `>=0.28.1` (GHSA: binary integrity verification, high severity) so the transitive dev-dependency resolved into the committed lockfile is not the vulnerable version.

## Verification

- On current master, re-verified the exact Dockerfile sequence locally: `pnpm -C packages/plugins/sandbox-providers/kubernetes install --ignore-workspace --frozen-lockfile` succeeds against the committed lockfile; the fresh install leaves `node_modules/@paperclipai` absent (confirming the link step is required post-#8255); `node scripts/link-plugin-dev-sdk.mjs` restores the symlink; `pnpm run build` then produces `dist/manifest.js` and `dist/worker.js`, satisfying the gate.
- `pnpm install --frozen-lockfile --lockfile-only` in the plugin directory passes, confirming the committed lockfile matches the plugin's current `package.json`.
- Built the image from this Dockerfile (earlier revision of this change) and deployed it to a live Kubernetes cluster: `ensureBundledKubernetesPlugin` auto-installed and activated the provider on every replica ("Kubernetes sandbox provider plugin ready"), and agent runs created per-tenant namespaces, ServiceAccounts, ResourceQuotas, and batch Jobs running the declared runtime image.
- Without the block (image from current master): the boot log shows the skip message and kubernetes-mode runs fail with the not-registered error.
- Round 2 (deps-stage move, e266273): `docker build --check .` clean; `docker build --target deps` succeeds — the plugin install works with only its `package.json` + `pnpm-lock.yaml` present (no source tree, no built SDK); `docker build --target build` succeeds end-to-end — link script reports "Linked @paperclipai/plugin-sdk into 1 excluded plugin(s)", plugin `tsc` build passes, `test -f` gate passes. Cache scoping confirmed empirically: after touching an unrelated `server/src` file and rebuilding, the plugin install layer reports `CACHED`. Note the plugin install must stay a separate `RUN` (not merged into the workspace install): it is `--ignore-workspace` with its own lockfile, and its layer must not share a cache key with the root install. `docker.yml` only runs on push to `master`, so these local image builds are the coverage for this change.

## Risks

- Image build time increases by the plugin's install+build step.
- The build now hard-fails if the kubernetes plugin breaks — deliberate (invariant: no deployable image with a broken bundled plugin).
- The committed plugin-local lockfile needs a manual refresh when the plugin's `package.json` deps change; the `--frozen-lockfile` install fails loudly in the image build if it drifts, so it cannot rot silently.
- No runtime behavior change for deployments that don't enable kubernetes execution: the plugin is only auto-installed when the bundle exists and is only used when selected.

## Model Used

Claude (Fable 5, model ID `claude-fable-5`), extended thinking, agentic tool use via Claude Code.

## Related PRs

The k8s execution stack that ships the plugin and its auto-install (#5790, #7938, #7934) — this makes that stack actually usable from the published container image. #8255 (plugin postinstall removal) is the reason for the explicit SDK link step.

## Checklist

- [x] I have run tests locally and they pass (Dockerfile + plugin packaging change; verified by exercising the exact install/link/build sequence and the image on a live cluster — see Verification)
- [x] I have added or updated tests where applicable (the in-Dockerfile `test -f` gate is the regression guard)
- [x] I have updated relevant documentation to reflect my changes (the Dockerfile block carries the full rationale inline; no user-facing docs describe the image build internals)

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -55,6 +55,28 @@ RUN pnpm --filter @paperclipai/plugin-sdk build
 RUN pnpm --filter @paperclipai/server build
 RUN test -f server/dist/index.js || (echo "ERROR: server build output missing" && exit 1)
 
+# Build the kubernetes sandbox-provider plugin standalone. It is intentionally
+# excluded from the pnpm workspace to keep its heavy deps
+# (@kubernetes/client-node) out of the root lockfile, so the workspace install
+# above does NOT install or build it. Install + build it in place so its
+# dist/, node_modules/, and the @paperclipai/plugin-sdk dev symlink land in
+# /app and get copied into the production stage; app.ts auto-installs it at
+# startup so the "kubernetes" sandbox provider is registered in containers.
+# A build failure here fails the whole image build, so a broken plugin can
+# never produce a deployable image. The plugin carries its own pnpm-lock.yaml
+# (it is outside the workspace, so the root lockfile cannot cover it);
+# --frozen-lockfile keeps the embedded @kubernetes/client-node resolution
+# reproducible across builds. The fresh install wipes node_modules and plugins
+# no longer carry a postinstall (removed for supply-chain safety, #8255), so
+# re-link the in-repo @paperclipai/plugin-sdk afterwards — the same
+# install-then-link order scripts/build-standalone-public-packages.mjs uses.
+RUN CI=true pnpm -C packages/plugins/sandbox-providers/kubernetes install --ignore-workspace --frozen-lockfile
+RUN node scripts/link-plugin-dev-sdk.mjs
+RUN CI=true pnpm -C packages/plugins/sandbox-providers/kubernetes run build
+RUN test -f packages/plugins/sandbox-providers/kubernetes/dist/manifest.js \
+  && test -f packages/plugins/sandbox-providers/kubernetes/dist/worker.js \
+  || (echo "ERROR: kubernetes plugin build output missing" && exit 1)
+
 FROM base AS production
 ARG USER_UID=1000
 ARG USER_GID=1000
--- a/Dockerfile
+++ b/Dockerfile
@@ -43,8 +43,21 @@ COPY packages/plugins/plugin-llm-wiki/package.json packages/plugins/plugin-llm-w
 COPY packages/plugins/plugin-workspace-diff/package.json packages/plugins/plugin-workspace-diff/
 COPY patches/ patches/
 COPY scripts/link-plugin-dev-sdk.mjs scripts/
+COPY packages/plugins/sandbox-providers/kubernetes/pnpm-lock.yaml packages/plugins/sandbox-providers/kubernetes/
 
 RUN pnpm install --frozen-lockfile
+
+# Install the kubernetes sandbox-provider plugin standalone. It is
+# intentionally excluded from the pnpm workspace to keep its heavy deps
+# (@kubernetes/client-node) out of the root lockfile, so the workspace install
+# above does NOT cover it. It carries its own pnpm-lock.yaml (the root
+# lockfile cannot cover a package outside the workspace); --frozen-lockfile
+# keeps the embedded @kubernetes/client-node resolution reproducible across
+# builds. Installing here (deps stage) scopes the layer cache to just the
+# plugin's package.json and lockfile, so unrelated source changes don't
+# re-download its dependency tree. The SDK link + build happen in the build
+# stage, after the in-repo @paperclipai/plugin-sdk is compiled.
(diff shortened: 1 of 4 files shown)


stubbi:
@greptileai re-review please — round-2, addressed the round-1 P2 (Dockerfile layer-cache scoping):
- e2662736443c5a3b8574080740466e4efb08b67f moves the kubernetes plugin's standalone `pnpm install` from the `build` stage (post-`COPY . .`) into the `deps` stage, keyed only on the plugin's `package.json` (already copied by the sandbox-providers glob) and its `pnpm-lock.yaml` (new COPY line). The SDK link, plugin build, and `test -f` gate remain in the `build` stage since they need the compiled in-repo `@paperclipai/plugin-sdk`.
- Verified locally (PR CI does not build the image — `docker.yml` is push-to-master only): `docker build --check` clean; `--target deps` succeeds with only manifest+lockfile present; `--target build` succeeds end-to-end (SDK link + plugin tsc + gate); after touching an unrelated `server/src` file and rebuilding, the plugin install layer reports `CACHED`.

PR #8895, opened 2026-07-02
fix(recovery): model_call_hang watchdog + auto-cancel/retry (FUL-633)

larspage:
## Summary

Implements the `model_call_hang` watchdog signal + auto-cancel/retry primitive on the paperclip side, per the [FUL-632 v2 post-mortem](/FUL/issues/FUL-632). Closes [FUL-633](/FUL/issues/FUL-633).

The watchdog previously classified `outputSilence.level: suspicious` correctly but had no recovery action, so model-call hangs (alive, ~zero CPU, no output) sat until the 4h `critical` threshold. This PR adds:

1. **`evaluateModelCallHang` predicate** — running + `lastOutputAt == null` + `pid_alive` + `cpu_seconds / wall_clock_seconds < 0.05`. Reader reads utime + stime from `/proc/<pid>/stat` (Linux-only, 60s floor).
2. **`scanModelCallHangRecovery` scanner** — per-run decision matrix: snooze-honor → grace check → auto-cancel + single retry (under four gates: automation invocation, source issue not board-owned, no prior retry, retry count `< 1`) → structured escalation issue under source issue if any gate fails.
3. **Liveness-tick wiring** in `server/src/index.ts` mirroring `scanSilentActiveRuns` (startup + periodic).
4. **Audit `run_event`** with `eventType: "lifecycle"`, `level: "warn"`, full recovery context.
5. **Env knobs** `MODEL_CALL_HANG_GRACE_MIN` (default `30`, floor `60s`) and `ENABLE_MODEL_CALL_HANG_RECOVERY` (default `false` prod / `true` dev).

Out of scope: `packages/adapters/opencode-local/src/server/parse.ts` is unchanged (parser is post-exit; the gap was upstream).

## Diff surface (+762/-2)

| File | Change |
|------|--------|
| `server/src/services/recovery/service.ts` | +707/-2 (predicate + scanner; lifted to module scope for unit testing) |
| `server/src/services/recovery/origins.ts` | +1 (origin kind) |
| `server/src/services/heartbeat.ts` | +15/-1 (thin wrapper exposing the scanner) |
| `server/src/index.ts` | +16 (liveness-tick wiring) |
| `server/src/__tests__/model-call-hang-recovery.test.ts` | new (14 integration tests) |
| `server/src/__tests__/model-call-hang-predicate.test.ts` | new (9 isolated tests) |
| `docs/agents-runtime.md` | +25 (§3.2 documents env knobs + recovery semantics) |

## Tests / verification

[...]

PR #9019, opened 2026-07-05
fix(heartbeat): arm a fast liveness backstop on process-loss retries

roshangautam:
fix(heartbeat): arm a fast liveness backstop on process-loss retries

## Thinking Path

> - When a heartbeat run's child process dies (host/container restart, OOM,
>   supervisor recycle), reapOrphanedRuns fails the run with errorCode
>   "process_lost" and enqueueProcessLossRetry spawns exactly one replacement
>   run automatically.
> - That replacement run inherits no scheduledRetry, monitor, or watchdog of
>   its own -- the only thing that can ever look at it again is the generic
>   silent-active-run sweep (scanSilentActiveRuns), gated on a full 1-hour
>   suspicion threshold.
> - When the same event that killed the original run also takes out sibling
>   agents on the fleet (the common case for a host/container-level event,
>   not an independent per-agent fault), nobody is left to notice a retry
>   that came up completely dead until that full hour elapses.
> - This pull request gives process-loss-retry runs a much shorter, dedicated
>   self-check threshold and flags them critical immediately instead of
>   waiting for the generic 4-hour critical bar, so a genuinely dead retry
>   surfaces as owner-assigned recovery work in minutes instead of up to an
>   hour, while a retry that is actually producing output is left alone
>   exactly like any other healthy run.
> - The benefit is that the platform's own auto-retry mechanism can no longer
>   silently strand a run for up to an hour with nothing watching it --
>   turning a "born stranded" retry into a self-healing (or at least
>   fast-escalating) one.

No exact matching public GitHub issue found after searching (terms tried:
"process_lost retry watchdog", "process-loss retry stranded zombie",
"enqueueProcessLossRetry", "born stranded retry no monitor"). Describing
in-PR as a bug report:

- **What happened:** a heartbeat run automatically re-spawned after a
  process-loss failure (`enqueueProcessLossRetry` in
  `server/src/services/heartbeat.ts`) has no `scheduledRetry`/monitor/
  watchdog armed on it. If the underlying event that killed the original
  run also affects other agents (a host/container restart, OOM, or
  supervisor recycle killing several child pids within the same short
  window), the retry can sit `running` with zero output for up to the full
  1-hour generic suspicion threshold before `scanSilentActiveRuns` ever
  looks at it.
- **Expected behavior:** a run spawned specifically to recover from a
  process-loss failure should be watched much more closely than an ordinary
  run, since by construction nothing else is watching it.
- **Steps to reproduce:** kill the child process (or restart the server)
  backing a running heartbeat run for a `claude_local`/`codex_local`-class
  adapter; observe the auto-spawned retry run has `scheduledRetryAt: null`
  and no active watchdog, and stays silent well past a reasonable window
  before any evaluation issue is created for it.
- **Deployment mode:** any deployment using local-child-process adapters
  (`claude_local`, `codex_local`, etc.) with the heartbeat scheduler enabled.

Related in-flight PRs found via search (adjacent, not overlapping): #8895
(`model_call_hang` watchdog for alive-but-idle CPU), #7815 (watchdog cursor
staying in step with stdout), #6954/#6957 (process_lost retry rate-limit
guardrails), #8502 (hung-run reaper for alive-but-hung processes). None of
these arm a liveness backstop on the process-loss-retry run itself.

## What Changed

- `server/src/services/recovery/service.ts`:
  - Added `PROCESS_LOST_RETRY_OUTPUT_SUSPICION_THRESHOLD_MS` (10 minutes),
    a much shorter suspicion window than the generic
    `ACTIVE_RUN_OUTPUT_SUSPICION_THRESHOLD_MS` (1 hour).
  - Added `isProcessLossRetryRun`, identifying a run
    `enqueueProcessLossRetry` spawned via its `retryOfRunId` plus the
    `wakeReason`/`retryReason` it stamps into `contextSnapshot`.
  - `scanSilentActiveRuns` now runs a second, narrowly-scoped query (`retryOfRunId
    is not null` and `contextSnapshot->>'wakeReason' = 'process_lost_retry'`) on the
    shorter threshold, merged with the existing generic-threshold query --
    so this never competes with ordinary long-running work for the
    existing query's `.limit(100)`.
  - `createOrUpdateStaleRunEvaluation` flags a process-loss-retry run
    `critical` (high priority, source-issue escalation comment) as soon as
    it crosses its own short threshold, instead of waiting for the generic
    4-hour critical bar.
  - The generated evaluation issue description explains why a process-loss
    retry is held to a different bar, so the assigned recovery owner has
    the context without needing to read source.
- `server/src/services/heartbeat.ts`: re-export the new constant alongside
  the existing `ACTIVE_RUN_OUTPUT_*` constants (heartbeat.ts is the module
  most callers/tests already import these from).
- `server/src/__tests__/heartbeat-active-run-output-watchdog.test.ts`:
  - Extended the `seedRunningRun` fixture with an `isProcessLossRetry`
    option that seeds a real failed predecessor run (satisfying the
    self-referencing `retryOfRunId` FK) and stamps the matching
    `contextSnapshot`.
  - Added a regression test proving a silent process-loss-retry run is
    caught and flagged critical well before the generic 1-hour bar would
    ever have looked at it.
  - Added a regression test proving a process-loss-retry run that is
    actually producing fresh output is left alone -- guarding against the
    false-positive class a real fleet incident report flagged, where a
    retry's natural quiet gaps between tool calls looked like death in a
    single snapshot.
- `doc/DEVELOPING.md`: documented the mechanism under a new "Process-Loss
  Retry Liveness Backstop" section.

## Verification

- `pnpm --filter @paperclipai/plugin-sdk build` (prerequisite for a clean
  typecheck against workspace-linked packages).
- `cd server && ../node_modules/.bin/tsc --noEmit -p tsconfig.json` -- clean,
  zero errors.
- `cd server && ../node_modules/.bin/vitest run
  src/__tests__/heartbeat-active-run-output-watchdog.test.ts
  --no-file-parallelism --maxWorkers=1` -- **20/20 passed**, including the
  2 new regression tests (both fail without the source change: they
  exercise the new `isProcessLossRetry` fixture option and constant, which
  do not exist pre-fix).
- `cd server && ../node_modules/.bin/vitest run
  src/__tests__/heartbeat-process-recovery.test.ts --no-file-parallelism
  --maxWorkers=1` -- 66/68 passed. The 2 remaining failures
  (`reaps orphaned descendant process groups when the parent pid is
  already gone`, `redacts secret-bearing successful-run detected progress
  before handoff disclosure`) reproduce identically on a clean, unmodified
  `master` with zero changes applied (verified via `git stash`), so they
  are pre-existing flakes, not introduced by this PR.
- `cd server && ../node_modules/.bin/vitest run
  src/__tests__/issue-recovery-actions.test.ts
  src/services/recovery/service.pause-durability.test.ts
  src/services/recovery/successful-run-handoff.test.ts` -- 43/43 passed
  (adjacent recovery-service test files, unaffected).
- `git diff --check` -- clean.

## Risks

- Low-medium. This only changes behavior for runs matching the narrow
  `isProcessLossRetryRun` predicate (a live process-loss-retry run) --
  ordinary runs are completely unaffected (verified by the existing 18
  tests in the touched file all still passing unchanged).
- The new 10-minute default is a judgment call balancing "catch a dead
  retry fast" against "don't false-positive on a healthy retry's normal
  quiet gaps between tool calls" (a real false-positive class this PR's
  own test suite guards against). It is not yet environment-configurable
  via an env var (documented as a constant to adjust directly in
  `doc/DEVELOPING.md`); a follow-up could add an env override analogous to
  `PAPERCLIP_HUNG_RUN_REAP_THRESHOLD_MS` if operators need to tune it
  without a code change.
- No database migration -- this only reads existing `retryOfRunId` /
  `contextSnapshot` columns.
- The two pre-existing test flakes noted above are unrelated to this PR
  but are called out explicitly per CONTRIBUTING.md so they are not
  mistaken for a regression this PR introduced.

## Model Used

Claude (Anthropic), Sonnet 4.5 (`claude-sonnet-5`), via Claude Code. Agentic
mode with tool use (file read/edit, shell, an isolated fork clone, embedded-
Postgres-backed vitest runs, GitHub CLI search for prior art) and extended
investigation across the running Paperclip instance's own API/process state
to establish the root cause before writing any code.

## Checklist

- [x] I have included a thinking path that traces from project context to
      this change
- [x] I have checked ROADMAP.md and confirmed this PR does not duplicate
      planned core work
- [x] I have searched GitHub for duplicate or related PRs and linked them
      above
- [x] I have either (a) linked existing issues with `Fixes: #` / `Closes #`
      / `Refs #` OR (b) described the issue in-PR following the relevant
      issue template
- [x] I have not referenced internal/instance-local Paperclip issues or
      links (only public GitHub `#NNN` / `github.com/paperclipai/paperclip`
      URLs)
- [x] My branch name describes the change and contains no internal
      Paperclip ticket id or instance-derived details
- [ ] All Paperclip CI gates are green (pending -- CI runs on this PR)
      (pending review)
- [x] I will address all Greptile and reviewer comments before requesting
      merge

Co-authored-by: Claude <noreply@anthropic.com>


roshangautam:
Addressed both findings in d969be9de:

1. **SQL/TypeScript predicate divergence** — aligned the fast-sweep SQL query to check `(wakeReason = 'process_lost_retry' OR retryReason = 'process_lost')`, matching `isProcessLossRetryRun`'s TypeScript contract exactly, with a comment noting why (dormant today since `enqueueProcessLossRetry` always stamps both, but now explicit against future regression).
2. **Double blank-line formatting regression** — applied your suggested fix (spreading a conditional array instead of a ternary with an empty-string branch), so the non-retry description path is byte-identical to before this PR, and the retry path gets exactly one blank line in each spot. Added two regression assertions (regex-matching the exact blank-line structure for both paths) so this can't silently regress again.

Re-ran the full watchdog test suite (20/20 passed) and a clean typecheck after both fixes.

roshangautam:
## Status: all CI green, Greptile 5/5, ready for maintainer review

Final state as of this comment:
- All 14 CI gates pass (Build, Typecheck, e2e, 5× General tests, 4× Verify serialized server suites, Canary Dry Run, Greptile).
- Greptile: **5/5 confidence, "No files require special attention."** Both findings from the first review pass (SQL/TypeScript predicate divergence on `retryReason`; a double-blank-line formatting regression in the generated evaluation description) were fixed and confirmed resolved.
- Zero open review comments.

Root-cause context for reviewers: this fix was root-caused against a live Paperclip instance experiencing exactly the failure mode described — a coordinated fleet process-loss event repeating on a roughly hourly cadence, where the harness's own `enqueueProcessLossRetry` retry came up with no scheduled-retry/monitor/watchdog armed and could sit silently "running" for up to the generic 1-hour suspicion window with nothing watching it. Direct process-table evidence from that instance (the server's own PID 1 restart timestamp landing 3 seconds before a documented incident timestamp, and a dependent agent's child process restarting ~16-19s later) corroborated that these events are a single shared-infra restart cycle killing the whole process tree at once, not independent per-agent faults — which is exactly the scenario this fix targets (nothing else survives to notice the retry is dead).

This PR is ready to merge pending maintainer review — I don't have merge rights on this repo.

PR #9041, opened 2026-07-05
Fail loudly on invalid config files

### LINKED ISSUE #8908: Silent full-config fallback when config.json fails schema validation (catch swallows zod error; banner still prints config path)
### Summary

When an instance `config.json` fails schema validation, the server silently ignores the **entire** file and boots with all defaults — while the startup banner still prints the config path, which makes the failure very hard to diagnose.

### Version / environment

- `@paperclipai/server` **2026.517.0** (npm), Windows 10, Node 22
- Multi-instance setup (`PAPERCLIP_HOME` + `PAPERCLIP_INSTANCE_ID`, external Postgres)

### Repro

1. Create a valid instance config and change one metadata field to a free-text value, e.g.:

```json
{ "$meta": { "version": 1, "updatedAt": "2026-07-03T01:12:00.000Z", "source": "provisioned-by-script" }, "server": { "deploymentMode": "local_trusted", "port": 3112 }, "database": { "mode": "postgres", "connectionString": "postgresql://..." } }
```

2. Start the server for that instance.

### Expected

Either boot honoring the config, or fail loudly pointing at the invalid field.

### Actual

- `$meta.source` is `z.enum(["onboard", "configure", "doctor"])`, so the free-text value invalidates the **whole document**.
- `dist/config-file.js` `readConfigFile()` wraps `paperclipConfigSchema.parse(raw)` in a `try { ... } catch { }` that returns null **without logging anything**.
- Result: server boots with all defaults — port 3100, "Using embedded PostgreSQL because no DATABASE_URL set" — while the banner still prints `Config: <path-to-the-ignored-file>`.
- Side effect on a busy machine: the default-port fallback chain (3001 → next free) makes the mis-booted instance look like a rogue copy of another instance, which cost us a while to trace.

### Suggestions

1. Log the zod issues (path + message) at WARN/ERROR when config parsing fails — even just `error.issues[0]` would have pointed straight at `$meta.source`.
2. Consider failing fast on invalid config (or an explicit opt-out flag), since "half-ignored config" is arguably worse than not booting.
3. Loosen `$meta.source` to `z.string()` (it looks informational), or document that it is a closed enum.

Thanks for the project — happy to provide more detail if useful.

lesbass:
## Thinking Path

> - The server config loader reads `.paperclip/config.json` and feeds it into the shared Paperclip config schema.
> - When a config file exists but cannot be parsed or fails schema validation, Paperclip should not silently ignore it.
> - The current `readConfigFile()` catch block treats invalid files the same as missing files, so startup falls back to defaults while the banner can still point at the ignored config path.
> - This pull request keeps the missing-file fallback, but makes present invalid config files fail with a path-specific error.
> - The benefit is safer startup behavior and a clear diagnostic that points at the invalid config field.

Fixes #8908

## What Changed

- Changed `readConfigFile()` to return `null` only when the config file is absent.
- Added explicit errors for unreadable/invalid JSON config files.
- Added explicit Zod validation errors that include the config path and invalid field path without printing config contents.
- Added server tests for missing config, invalid JSON, schema validation failure, and valid config parsing.

## Verification

- `pnpm exec vitest run server/src/__tests__/config-file.test.ts`

## Risks

Low risk for valid configs and missing configs. This intentionally changes behavior for present invalid config files from silent fallback to startup failure, which is the issue being fixed.

## Model Used

OpenAI Codex based on GPT-5, with repository file inspection, GitHub CLI, and local command execution.

## Checklist

### DIFF
--- a/server/src/__tests__/config-file.test.ts
+++ b/server/src/__tests__/config-file.test.ts
@@ -1,0 +1,92 @@
+import fs from "node:fs";
+import os from "node:os";
+import path from "node:path";
+import { afterEach, beforeEach, describe, expect, it } from "vitest";
+import { readConfigFile } from "../config-file.js";
+
+const ORIGINAL_PAPERCLIP_CONFIG = process.env.PAPERCLIP_CONFIG;
+
+function writeConfig(configPath: string, value: unknown): void {
+  fs.writeFileSync(configPath, `${JSON.stringify(value, null, 2)}\n`);
+}
+
+function escapeRegExp(value: string): string {
+  return value.replace(/[.*+?^${}()|[\]\\]/g, "\\$&");
+}
+
+function minimalConfig(): unknown {
+  return {
+    $meta: {
+      version: 1,
+      updatedAt: "2026-07-05T00:00:00.000Z",
+      source: "configure",
+    },
+    database: {
+      mode: "embedded-postgres",
+    },
+    logging: {
+      mode: "file",
+    },
+    server: {},
+  };
+}
+
+describe("readConfigFile", () => {
+  let tempDir: string;
+  let configPath: string;
+
+  beforeEach(() => {
+    tempDir = fs.mkdtempSync(path.join(os.tmpdir(), "paperclip-config-file-test-"));
+    configPath = path.join(tempDir, "config.json");
+    process.env.PAPERCLIP_CONFIG = configPath;
+  });
+
+  afterEach(() => {
+    if (ORIGINAL_PAPERCLIP_CONFIG === undefined) {
+      delete process.env.PAPERCLIP_CONFIG;
+    } else {
+      process.env.PAPERCLIP_CONFIG = ORIGINAL_PAPERCLIP_CONFIG;
+    }
+
+    fs.rmSync(tempDir, { recursive: true, force: true });
+  });
+
+  it("returns null when the config file does not exist", () => {
+    expect(readConfigFile()).toBeNull();
+  });
+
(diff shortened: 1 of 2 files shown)


sachinkr7368:
Landed here from #8908 — thanks for taking this, and sorry for briefly opening a duplicate (#9083, now closed).

One design question worth surfacing for the reviewer: the issue offered two paths, and this PR takes the *fail-fast* one (throw / refuse to boot on an invalid `config.json`). The alternative is *log-and-continue* (WARN naming the offending field, then fall back to defaults as today).

Tradeoff: fail-fast is safer against a half-ignored config, but it's a behavior change — an instance that currently boots (silently ignoring a bad file) would stop booting after upgrade, which could surprise self-hosters mid-incident. Log-and-continue keeps boot semantics and just makes the failure visible. Both name the bad field, so either fixes the core "silent" complaint.

No strong opinion on which the maintainers prefer — just flagging so it's a deliberate call. Happy to help either way.

lesbass:
@sachinkr7368 thanks for raising the fail-fast vs log-and-continue tradeoff. We considered both and deliberately chose fail-fast: silently ignoring a bad config.json is the root cause of the original issue (#8908), and boot-time crashes are immediately visible and trivially fixable — the error message already names the offending field and path. A log-and-continue path would leave room for half-applied configs to cause subtle runtime issues that are harder to diagnose. That said, we're open to maintainer preference here — happy to adjust if the team prefers the softer approach.

PR #9083, opened 2026-07-06
fix(config): log invalid config.json instead of silently falling back (#8908)

### LINKED ISSUE #8908: Silent full-config fallback when config.json fails schema validation (catch swallows zod error; banner still prints config path)

sachinkr7368:
## Thinking Path

> - Each instance is configured by a `config.json` (deployment mode, port, database, logging) that the server loads at startup via `readConfigFile()`
> - When that file fails JSON parsing or schema validation, `readConfigFile()` wrapped `paperclipConfigSchema.parse()` in an empty `catch {}` and returned `null` — so the **entire** file was discarded and the server booted with all defaults
> - The failure was silent: nothing was logged, yet the startup banner still printed the ignored config path, so operators saw "Config: <path>" while the file was in fact being ignored — a single bad field (e.g. a free-text `$meta.source`, which is a closed enum) invalidated the whole document and looked like a rogue/mis-booted instance
> - This PR makes the fallback loud: it logs a WARN naming the config path and the offending field(s), then falls back to defaults as before
> - The benefit is that invalid-config boots are diagnosable in seconds instead of costing an operator a long trace

Fixes: #8908

## What Changed

- `server/src/config-file.ts`: split JSON-parse and schema-validation into separate `try` blocks. On failure, log `[paperclip] Ignoring config at <path>: <reason>. Booting with defaults.` where `<reason>` is the JSON parse error, or — for a `ZodError` — the `path: message` of each issue (so `$meta.source` and similar point straight at the bad field). Still returns `null` (defaults) on failure, so boot behavior is unchanged.
- Warning is deduped once per config path per process, because `readConfigFile()` is called repeatedly (config load, logger init, model listing) and a persistently-invalid file must not spam the log.
- `server/src/__tests__/config-file.test.ts`: new regression test — invalid schema (the exact repro config from #8908) returns `null` and WARNs naming `$meta.source`; invalid JSON returns `null` and WARNs `invalid JSON`.

## Verification

```bash
pnpm --filter @paperclipai/server exec vitest run src/__tests__/config-file.test.ts   # 2 passed
pnpm --filter @paperclipai/server typecheck                                            # clean
```

The test drives the real `paperclipConfigSchema` against the reporter's exact invalid config, so the diagnostic message (offending field path) is exercised end-to-end, not mocked.

## Risks

Low risk. Behavior on a *valid* config is unchanged. Behavior on an *invalid* config is unchanged except that it now logs (previously silent) — the server still falls back to defaults rather than failing fast, preserving current boot semantics. No schema change (the reporter's suggestion to loosen `$meta.source` to `z.string()` is deliberately left out — that's a product decision about whether `source` is informational or a closed enum; this PR only fixes the silent-failure diagnosability, which helps *every* invalid-config case, not just that one field). No telemetry changes.

## Model Used

Claude Opus 4.8 (Anthropic), model ID `claude-opus-4-8`, 1M context window, extended thinking + tool use (Claude Code).

## Checklist

- [x] I have either (a) linked existing issues with `Fixes: #` OR (b) described the issue in-PR
- [x] My branch name describes the change (`fix/log-invalid-config-fallback`) and contains no internal ticket id
- [ ] I have updated relevant documentation to reflect my changes (none needed — behavior-preserving diagnostics only)
- [ ] All Paperclip CI gates are green (will confirm after CI runs)
- [ ] Greptile is 5/5 with no open P2s, recommendations, or follow-ups (will address on review)

### DIFF
--- a/server/src/__tests__/config-file.test.ts
+++ b/server/src/__tests__/config-file.test.ts
@@ -1,0 +1,57 @@
+import { describe, expect, it, vi, beforeEach } from "vitest";
+
+/**
+ * Regression test for https://github.com/paperclipai/paperclip/issues/8908
+ *
+ * When an instance config.json fails schema validation, readConfigFile()
+ * used to swallow the Zod error in an empty `catch {}` and silently boot with
+ * all defaults — while the startup banner still printed the ignored config
+ * path, making the failure very hard to trace. It must now (a) still fall back
+ * to defaults (return null) and (b) log a WARN naming the offending field so
+ * operators can diagnose it.
+ */
+
+const mockResolvePath = vi.hoisted(() => vi.fn());
+const mockReadFileSync = vi.hoisted(() => vi.fn());
+
+vi.mock("node:fs", async (importOriginal) => {
+  const actual = await importOriginal<typeof import("node:fs")>();
+  return { ...actual, default: { ...actual, existsSync: () => true, readFileSync: mockReadFileSync }, existsSync: () => true, readFileSync: mockReadFileSync };
+});
+vi.mock("../paths.js", () => ({ resolvePaperclipConfigPath: mockResolvePath }));
+
+import { readConfigFile } from "../config-file.ts";
+
+describe("readConfigFile invalid-config diagnostics (#8908)", () => {
+  beforeEach(() => {
+    vi.clearAllMocks();
+  });
+
+  it("returns null and WARNs with the offending field path on schema failure", () => {
+    // Unique path per test so the warn-once cache doesn't suppress this call.
+    mockResolvePath.mockReturnValue("/tmp/pc-test-schema.json");
+    mockReadFileSync.mockReturnValue(
+      JSON.stringify({
+        $meta: { version: 1, updatedAt: "2026-07-03T01:12:00.000Z", source: "provisioned-by-script" },
+        server: { deploymentMode: "local_trusted", port: 3112 },
+      }),
+    );
+    const warn = vi.spyOn(console, "warn").mockImplementation(() => {});
+
+    expect(readConfigFile()).toBeNull();
+    expect(warn).toHaveBeenCalledTimes(1);
+    const msg = warn.mock.calls[0][0] as string;
+    expect(msg).toContain("/tmp/pc-test-schema.json");
+    expect(msg).toContain("$meta.source");
+  });
+
+  it("returns null and WARNs on invalid JSON", () => {
+    mockResolvePath.mockReturnValue("/tmp/pc-test-json.json");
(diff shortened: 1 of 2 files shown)


sachinkr7368:
cc @cryppadotta — small, self-contained fix for #8908 (silent config fallback). Behavior-preserving: still falls back to defaults on an invalid `config.json`, but now logs a WARN naming the offending field instead of swallowing the Zod error. One file + one regression test; typecheck clean and tests pass locally. Whenever you have a moment for a review, much appreciated — happy to adjust to any conventions I've missed.

sachinkr7368:
Closing as a duplicate of #9041, which predates this and fixes the same issue (#8908) in the same file. I missed it when searching — apologies for the noise. Moving my notes over to #9041.

PR #9115, opened 2026-07-06
fix(heartbeat): use normal_model profile for issue monitor recovery dispatches

neerazz:
## Thinking Path

> - Paperclip is the open source app people use to manage AI agents for work; agent dispatches run on named model profiles, including a lightweight "status only" profile intended for post-a-status-and-exit work.
> - This change lives in the heartbeat subsystem (`server/src/services/heartbeat.ts`), specifically the issue-monitor recovery path that re-dispatches work when a monitored issue needs recovery.
> - The bug: those recovery dispatches were pinned to the `status_only` model profile, which resolves to a cheap model id that is not registered in every deployment, so dispatch fails with `model_not_found` and recovery silently never runs.
> - Pinning `status_only` was wrong on two counts — recovery dispatches need a model that can actually reason, and depending on an optional cheap model id makes recovery non-portable across deployments.
> - This pull request switches those recovery dispatches to `normal_model`, drops the dead `recoveryAssigneeAdapterOverrides("status_only")` override, and sets `assigneeAdapterOverrides: null` so the assignee's primary model is used.
> - The benefit is that issue-monitor recovery no longer crashes on deployments without the cheap profile, and recovery agents run on a model capable of the recovery work.

No public GitHub issue exists for this fix; describing the underlying bug inline following the bug report template:

**What happened:**
Issue-monitor recovery dispatches in `heartbeat.ts` were created with the `status_only` model-profile hint (and a matching `recoveryAssigneeAdapterOverrides("status_only")`). On deployments whose model registry does not define the cheap status-only model id, dispatch fails with `model_not_found`, so recovery never runs.

**Expected behavior:**
Recovery dispatches should use the assignee's normal model profile (`normal_model`) so they succeed regardless of whether a cheap status-only model is configured, and so recovery agents can actually perform the recovery work.

**Steps to reproduce:**
1. Run a Paperclip deployment whose model registry does not define the `status_only` cheap model id.
2. Trigger an issue-monitor recovery dispatch (a monitored issue enters the recovery path).
3. Observe the dispatch fail with `model_not_found` instead of running.

## What Changed

- `server/src/services/heartbeat.ts`: drop the dead `recoveryAssigneeAdapterOverrides` import; set `assigneeAdapterOverrides: null` directly; switch four `withRecoveryModelProfileHint(..., "status_only")` call sites (issue-monitor recovery issue creation + wake payloads) to `"normal_model"`.
- `heartbeat-model-profile.test.ts`: add 2 regression tests covering the disabled-cheap-profile path.
- `issue-monitor-scheduler.test.ts`: assert the wake payload carries no `modelProfile` key and the recovery issue has `assigneeAdapterOverrides: null`.
- `agent-adapter-validation-routes.test.ts`: add 1 regression test that a runtimeConfig-only patch preserves the primary model.

## Verification

- `npx vitest run heartbeat-model-profile.test.ts issue-monitor-scheduler.test.ts agent-adapter-validation-routes.test.ts` — 21/21 pass.
- `git diff --check` — no whitespace errors.
- Full CI on this PR is green except the body-contract review check, which this update addresses.

## Risks

Low risk. The change only affects the issue-monitor recovery dispatch path, moving those dispatches from a cheap status-only profile to the normal model profile — a strict superset of capability. No schema, API, or migration changes. Worst case is a marginally higher model cost on recovery dispatches, which is the intended behavior.

## Model Used

Claude (Anthropic), via the Claude Code CLI with extended thinking and tool use (local test execution). Code and tests were authored and verified through the CLI.

## Checklist

- [x] I searched the GitHub PR list (open + recently closed) for similar/duplicate PRs and confirmed this is not a duplicate
- [x] I have either (a) linked existing issues or (b) described the issue in-PR following the relevant issue template
- [x] My branch name describes the change


neerazz:
@cryppadotta @devinfoley quick maintainer nudge: this branch is mergeable clean, blocking checks are green, Greptile is 5/5, and there are no unresolved review threads. Could one of you take a review/merge pass when you have a window?

neerazz:
@cryppadotta @devinfoley one-time maintainer nudge: head `e2982f0` is mergeable clean, CI/checks are green with zero failures, Greptile is 5/5, and review threads are fully resolved. Could one of you review/merge or name the remaining blocker?

PR #9116, opened 2026-07-06
fix(ui): match plugin contributions by pluginKey as well

MaxBlabla:
Match plugin contributions by `pluginKey` as a fallback when `pluginId` doesn't match. This allows looking up page slots for plugins that identify by key rather than internal ID.

- `PluginPage.tsx`: one-line change to `find()` predicate
- `pnpm-lock.yaml`: workspace dependency update


MaxBlabla:
Not our repo — closing.

PR #9123, opened 2026-07-06
feat(db): add migration safety lint

nickyleach:
## Thinking Path

> - The `packages/db` module owns all database migrations via a sequential numbering system already validated at build time
> - A recent migration introduced an O(n²) batch backfill over a large, unindexed table — it caused the server's `listen` to block for ~5 minutes on databases with millions of rows
> - Nothing in the current CI pipeline catches large-table migration risk patterns (DO-loop mutations, batched LIMIT mutations without support indexes, full-table mutations, non-concurrent index creation) before they land
> - This PR wires a new static migration-safety checker (`check-migration-safety.ts`) into the existing `check:migrations` gate in `packages/db/package.json`, so risky patterns fail CI before reaching production
> - The checker baselines all historical findings already present in the codebase, so the gate fails only on *new* unbaselined risky patterns
> - The benefit is that the specific O(n²) backfill shape (and related patterns) will be caught at author time rather than at incident time

**Feature — static migration safety lint**

**Problem or motivation**

Migrations against large tables (millions of rows) have caused production startup blocks. The root pattern is a batched `LIMIT`-based backfill iterating via an unindexed column, making each batch a sequential scan — O(n²) overall. No CI gate exists to flag this class of problem before merge.

**Proposed solution**

A static SQL-level checker that scans new migration files for known dangerous patterns against known-large tables, producing structured findings that are either baselined (suppressed) or fail the build. Patterns detected: `DO $$ loop` mutations on large tables without a same-migration support index, batched `LIMIT` mutations on large tables missing a same-migration support index, unbounded full-table mutations (no `WHERE` clause), and `CREATE INDEX` without `CONCURRENTLY` on large tables.

**Alternatives considered**

Runtime instrumentation (only catches issues in production), advisory locking in migrations (doesn't prevent the pattern), per-migration code review (doesn't scale consistently).

**Roadmap alignment**

Defensive infrastructure / operational reliability — keeps migrations from blocking production startups. Not a user-facing feature.

## What Changed

- **`packages/db/package.json`** — extended `check:migrations` script to run `check-migration-safety.ts` after the existing numbering check
- **`packages/db/src/check-migration-safety.ts`** — new static checker: SQL pattern matching, rule detection for four dangerous patterns, baseline diffing, and structured exit with findings summary
- **`packages/db/src/migration-safety-baseline.ts`** — baseline of all existing historical findings (suppressed from failing the gate); new migrations matching these patterns without a baseline entry will fail
- **`packages/db/src/table-size-estimates.ts`** — rough table size estimates from the local dev database; drives `isKnownLargeTable()` used by the safety rules
- **`packages/db/src/check-migration-safety.test.ts`** — Vitest coverage for the O(n²) backfill failure mode, suppression via baseline, and each rule type

## Verification

```bash
# Run the migration safety checker directly
cd packages/db
tsx src/check-migration-safety.ts

# Run tests
cd packages/db
npx vitest run src/check-migration-safety.test.ts

# Run the full migration check gate (numbering + safety)
cd packages/db
pnpm run check:migrations
```

- Tests cover the core O(n²) backfill pattern (the motivating incident), baseline suppression, and all four rule types
- `check:migrations` now exits non-zero for any new unbaselined large-table migration risk pattern

## Risks

- **False positives:** Table-size estimates are from a local dev database snapshot — a table small in dev but large in production would be missed. Best-effort heuristic.
- **Baseline drift:** If a baselined finding's SQL changes significantly, the baseline ID (content-hash-based) will no longer match and the finding will re-surface. Intentional but may surprise authors doing incremental fixes.
- **SQL parsing limitations:** Regex-based pattern matching rather than a full AST parser — complex SQL may not be detected. Acceptable for an initial gate.
- **Low risk to existing behavior:** The gate only fails on *new* findings not present in the baseline. All existing migrations are baselined.

## Model Used

- **Provider:** Anthropic
- **Model ID:** `claude-sonnet-4-6`
- **Context window:** 200K
- **Tool use:** yes (file reading, bash, git operations)
- **Reasoning mode:** standard (no extended thinking)

## Checklist

### DIFF
--- a/packages/db/package.json
+++ b/packages/db/package.json
@@ -35,7 +35,7 @@
     "dist"
   ],
   "scripts": {
-    "check:migrations": "tsx src/check-migration-numbering.ts",
+    "check:migrations": "tsx src/check-migration-numbering.ts && tsx src/check-migration-safety.ts",
     "build": "pnpm run check:migrations && tsc && cp -r src/migrations dist/migrations",
     "clean": "rm -rf dist",
     "typecheck": "pnpm run check:migrations && tsc --noEmit",
--- a/packages/db/src/check-migration-safety.test.ts
+++ b/packages/db/src/check-migration-safety.test.ts
@@ -104,6 +104,39 @@ describe("migration safety check", () => {
     expect(result.newFindings).toEqual([]);
   });
 
+  it("does not suppress missing-index finding when a partial support index predicate is incompatible with the batch WHERE", () => {
+    const result = analyze(`
+      CREATE INDEX CONCURRENTLY IF NOT EXISTS "issue_comments_fixture_backfill_idx"
+        ON "issue_comments" USING btree ("id")
+        WHERE "author_agent_id" IS NOT NULL;--> statement-breakpoint
+      DO $$
+      DECLARE
+        last_comment_id uuid := '00000000-0000-0000-0000-000000000000'::uuid;
+      BEGIN
+        LOOP
+          WITH batch AS MATERIALIZED (
+            SELECT c."id"
+            FROM "issue_comments" c
+            WHERE c."id" > last_comment_id
+              AND c."author_agent_id" IS NULL
+            ORDER BY c."id"
+            LIMIT 5000
+          )
+          UPDATE "issue_comments" c
+          SET "derived_author_agent_id" = NULL
+          FROM batch b
+          WHERE c."id" = b."id";
+
+          EXIT WHEN NOT FOUND;
+        END LOOP;
+      END $$;
+    `);
+
+    expect(result.newFindings.map((finding) => finding.rule)).toContain(
+      "batched-mutation-large-table-missing-index",
+    );
+  });
+
   it("passes a batched backfill over a small-bucket table", () => {
     const result = analyze(`
       DO $$
--- a/packages/db/src/check-migration-safety.test.ts
+++ b/packages/db/src/check-migration-safety.test.ts
@@ -216,6 +216,35 @@ describe("migration safety check", () => {
     );
   });
 
+  it("does not treat WHERE inside a dollar-quoted string as a selective predicate", () => {
+    const result = analyze(`
+      UPDATE "issue_comments"
+      SET "body" = $msg$WHERE "id" > '0'$msg$;
+    `);
+
+    expect(result.newFindings.map((f) => f.rule)).toContain(
+      "full-table-mutation-large-table",
+    );
+  });
+
+  it("flags a FETCH FIRST inline batch mutation over a known-large table", () => {
+    const result = analyze(`
+      UPDATE "issue_comments" c
+        SET "derived_author_agent_id" = NULL
+        FROM (
+          SELECT "id"
+          FROM "issue_comments"
+          ORDER BY "id"
+          FETCH FIRST 5000 ROWS ONLY
+        ) b
+        WHERE c."id" = b."id";
+    `);
+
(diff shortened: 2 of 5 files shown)


nickyleach:
Greptile flagged 6 P1 findings (3 distinct bugs) in `check-migration-safety.ts` that represent genuine false-negative gaps in the safety checker:

1. **`hasBatchedLimitMutation` misses inline subquery batches** (line 276): The regex only matches the CTE form (`WITH...LIMIT...UPDATE`). An `UPDATE "t" ... FROM (SELECT id FROM "t" WHERE ... ORDER BY id LIMIT N) b WHERE ...` has `UPDATE` before `LIMIT` and no leading `WITH` — it passes the gate undetected, which is exactly the O(n²) pattern this checker exists to catch.

2. **`hasSelectiveWhere` reads the first `WHERE` anywhere, including inside CTEs** (lines 278-285): A migration like `WITH filtered AS (SELECT id FROM "t" WHERE ...) UPDATE "t" SET ... ;` triggers the CTE filter but performs an unbounded full-table update — it passes the full-table-mutation rule incorrectly.

3. **`hasMatchingSupportIndex` accepts any shared predicate column** (lines 297-303): An index on `author_agent_id` suppresses a batch backfill that advances by unindexed `id`. The batch cursor still scans the large table O(n²) but CI reports it as covered.

These are being fixed in a follow-up commit. The PR will be updated once the fixes are pushed and Greptile re-reviews.

nickyleach:
Build is now failing on the fix commit due to a regression introduced by the `hasSelectiveWhere` depth-aware change.

**What's happening:**

The depth-aware `topLevelWhereClause()` correctly ignores WHERE clauses inside parentheses. However, migration 0132 uses a DO-loop pattern where the batch WHERE is inside a CTE inside the DO block:

```sql
DO $$ ... BEGIN LOOP
  WITH batch AS MATERIALIZED (
    SELECT ... FROM issue_comments WHERE c.id > last_comment_id ... LIMIT 5000
  ),
  updated AS (UPDATE issue_comments SET ... FROM batch WHERE ...)
  ...
END LOOP; $$
```

Because both WHERE clauses are inside parentheses (at depth ≥ 1), `topLevelWhereClause()` returns null → `hasSelectiveWhere` returns false → `full-table-mutation-large-table` fires. The finding id is `38d8055cc228913d`. This was NOT in the baseline (the original checker's regex found the CTE WHERE at any depth and said 'ok'), so it now fails the gate.

**0132 is actually safe** — it uses keyset pagination (`WHERE id > last_comment_id ORDER BY id LIMIT 5000`) which is exactly the pattern the checker is designed to approve.

**Two fix options:**

Option A (simpler): Add the 0132 finding to the baseline in `migration-safety-baseline.ts`:
- `id: '38d8055cc228913d'`
- `rule: 'full-table-mutation-large-table'`  
- Note it as a false-positive: DO-loop with keyset pagination; WHERE is in the CTE batch clause, not at statement top level

Option B (cleaner): In `checkRules()`, skip `full-table-mutation-large-table` for statements that already fired `loop-mutation-large-table` — avoids double-flagging DO-loops that are already covered by the loop rule.

Please fix and push before the CI can turn green.

nickyleach:
@greptile review

nickyleach:
FETCH FIRST/NEXT batches and dollar-quoted string fixes pushed in `b6fb7cea6` (rebased on top of `78a3fe3c5`). CI running — Greptile started 01:09:03. Will monitor results and merge when clean.

PR #9127, opened 2026-07-07
Add continuation-retry circuit-breaker (shadow-mode first)

goldinbidsheets-rgb:
## Thinking Path

> - The scheduler's recovery service auto-continues stranded `in_progress` issues via `issue_continuation_needed`, extending the existing `scheduledRetry` / `activeRecoveryAction` subsystem
> - When a subprocess dies during startup (before its first model call) the run terminates as `process_lost` with **null usage / zero cost** — a transient loss, not a defect
> - Unbounded auto-continuation re-wakes such issues immediately with no backoff, so one transient loss can amplify into a storm (observed: 64 retries in ~2h on a single issue) until host concurrency drops and it self-heals
> - This PR adds a continuation-retry circuit-breaker that recognizes the clean zero-cost-`process_lost` trip-wire, backs off exponentially, and (in a later stage) trips after N consecutive matches
> - It ships **shadow-mode first**: the matcher/streak run and log what they *would* do with **zero behavior change**, so the matcher can be validated against real traffic before any scheduling change is enabled
> - The benefit is that a transient infrastructure hiccup can no longer burn dozens of zero-cost runs, while paid retries are provably never affected

No public tracking issue. Describing the bug in-PR (bug-report shape):

**What happens:** A transient subprocess startup loss produces a `process_lost` heartbeat run with `usage_json = null` (zero cost). The `issue_continuation_needed` auto-retry in `reconcileStrandedAssignedIssues` re-queues the issue with no bounded backoff for this class, so a single transient loss amplifies into many consecutive zero-cost retries on one issue before it self-heals.

**Expected:** After a few consecutive zero-cost `process_lost` retries on one issue, the scheduler should back off (and eventually stop + alert once) rather than hammer with no backoff.

**Trip-wire signature (clean):** `stopReason == process_lost` (i.e. `status == "failed" && errorCode == "process_lost"`) AND `usageJson == null` AND (no secret manifest OR every `contextSnapshot.paperclipSecrets.manifest[].outcome == "success"`). The all-success manifest condition means it is a startup loss, not an auth/secret defect. A paid run (non-null usage) can never match — this is the load-bearing safety guard.

**Related open PRs (complementary, different layers):** #6979 (transient-failure circuit breaker), #6954 / #6957 (per-issue `process_lost` retry rate-limit at `enqueueProcessLossRetry`, 1h/3-strike, no shadow), #8829 (liveness-continuation backoff + per-issue upstream-throttle ceiling, shadow-first). This PR is distinct: it is shadow-first, zero-cost-specific, and lives on the `issue_continuation_needed` continuation-scheduling path rather than the immediate process-loss retry path. Reviewers may want the layers reconciled; #8829 is intended to stack on this rather than duplicate it.

## What Changed

- **New pure module `server/src/services/recovery/continuation-breaker.ts`:**
  - `isZeroCostProcessLost(run)` — the trip-wire matcher (paid runs can never match).
  - `zcplStreakFromRuns(runs)` — per-issue streak of consecutive matches at the head of run history (mirrors the existing `summarizeRecentContinuationRetries` walk, so **no DB migration** is needed for shadow mode).
  - `decideContinuationBreaker(streak, cfg)` — exponential backoff (`base 30s x 2^(streak-1)`, cap 5m) for `streak < N`, trip verdict for `streak >= N` (default `N = 4`).
  - `resolveContinuationBreakerConfig()` — env-overridable config; default `mode = shadow`, `enabled = true`.
- **Shadow wiring in `reconcileStrandedAssignedIssues` (`recovery/service.ts`):** at the continuation-scheduling decision point, derives the streak and logs a structured `recovery.continuation_breaker.shadow` verdict (`would-backoff` / `would-trip`) including a `latestRunHadUsage` cross-check field. **No scheduling decision changes.** Adds two observability counters to the scan result.
- **Tests (`continuation-breaker.test.ts`):** matcher truth table, `breakerRunFromRow` extraction from the real `heartbeat_runs` shape, streak reset semantics, backoff/trip math, storm replay sim, and the paid-retry regression.

## Verification

- Unit tests in `continuation-breaker.test.ts` cover the matcher truth table, streak/backoff/trip math, storm replay, and paid-retry regression.
- Verified the pure module against these assertions via Node type-stripping and type-checked the module clean with `tsc --strict`. Branch is rebased on current `master` and merges without conflict. Full workspace `tsc` / `vitest` runs are left to CI.
- Shadow mode is behavior-neutral by construction: the only effect is an added log line + two counters; no scheduling branch is taken.

## Risks

- **Low risk.** Default `mode = shadow` — the change is observe-only and does not alter scheduling. The matcher is additive and cannot match paid runs (non-null usage short-circuits). No schema/migration changes. The enforce action (persisted trip state + suspend + manager notify) is a separate follow-up, gated on zero false-trips observed in shadow.

## Model Used

Claude Opus 4.8 (`claude-opus-4-8`), extended thinking + tool use, via an autonomous engineering agent.

## Checklist

### DIFF
--- a/server/src/services/recovery/continuation-breaker.test.ts
+++ b/server/src/services/recovery/continuation-breaker.test.ts
@@ -1,0 +1,153 @@
+import { describe, expect, it } from "vitest";
+import {
+  CONTINUATION_BREAKER_DEFAULTS,
+  breakerRunFromRow,
+  decideContinuationBreaker,
+  isZeroCostProcessLost,
+  resolveContinuationBreakerConfig,
+  zcplStreakFromRuns,
+  type BreakerRun,
+} from "./continuation-breaker.js";
+
+const CFG = CONTINUATION_BREAKER_DEFAULTS; // N=4, base=30s, cap=5m
+
+function run(overrides: Partial<BreakerRun> = {}): BreakerRun {
+  return {
+    status: "failed",
+    errorCode: "process_lost",
+    usageJson: null,
+    secretManifest: null,
+    ...overrides,
+  };
+}
+
+describe("continuation-breaker matcher (trip-wire truth table)", () => {
+  it("process_lost + null usage + null manifest => TRUE", () => {
+    expect(isZeroCostProcessLost(run())).toBe(true);
+  });
+
+  it("process_lost + null usage + empty manifest => TRUE", () => {
+    expect(isZeroCostProcessLost(run({ secretManifest: [] }))).toBe(true);
+  });
+
+  it("process_lost + null usage + all-success manifest => TRUE", () => {
+    expect(
+      isZeroCostProcessLost(run({ secretManifest: [{ outcome: "success" }, { outcome: "success" }] })),
+    ).toBe(true);
+  });
+
+  it("process_lost + a failed secret in manifest => FALSE (auth defect, not startup loss)", () => {
+    expect(
+      isZeroCostProcessLost(run({ secretManifest: [{ outcome: "success" }, { outcome: "failure" }] })),
+    ).toBe(false);
+  });
+
+  it("paid run (non-null usage) => FALSE — CEO hard constraint, load-bearing guard", () => {
+    expect(isZeroCostProcessLost(run({ usageJson: { input_tokens: 10, output_tokens: 5 } }))).toBe(false);
+  });
+
+  it("non-process_lost stop => FALSE", () => {
+    expect(isZeroCostProcessLost(run({ status: "cancelled", errorCode: "user_cancel" }))).toBe(false);
+    expect(isZeroCostProcessLost(run({ status: "succeeded", errorCode: null }))).toBe(false);
+  });
+
+  it("adapter_failed (a different transient class) => FALSE", () => {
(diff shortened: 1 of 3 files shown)


goldinbidsheets-rgb:
Thanks @greptile-apps — both P2s addressed in `dffd700`:

1. **Unreadable Manifests Count Clean** (`continuation-breaker.ts`): `readSecretManifest` now distinguishes a *genuinely absent* manifest (no `paperclipSecrets` block → confident "no secrets" → clean) from an *unreadable* one (malformed `contextSnapshot`, or `paperclipSecrets` whose `manifest` isn't an array). Unreadable → `secretManifestUnreadable: true` → the matcher returns `false`, so it resets the streak instead of counting as a clean loss (preserves the never-false-trip guarantee). A malformed manifest *entry* is now treated as `failure` rather than dropped.

2. **Latest Run Check Can Drift** (`service.ts`): `latestRunId` and `latestRunHadUsage` are now derived from the *same* query snapshot as the streak, so the paid-run cross-check always references the run it actually measured. The scan's own latest-run id is logged separately as `scanLatestRunId` for correlation.

Added tests for the unreadable-manifest, non-array-manifest, malformed-entry, and explicit-flag cases.

PR #9130, opened 2026-07-07
fix(agents): cascade-clean cost/finance telemetry on agent delete (decommission 500)

mrobinson2:
## Symptom

Tenant decommission failed at `remove_agents` with a 500 (crash past validation, not a schema error):

```
DELETE /api/agents/b52c2d98-... 500 — Failed query: delete from "heartbeat_runs" where "agent_id" = $1
caused by: update or delete on table "heartbeat_runs" violates foreign key constraint
  "cost_events_heartbeat_run_id_heartbeat_runs_id_fk" on table "cost_events"
```

(Confirmed from the live backend log for the failing agent.)

## Root cause

`agentService.remove()` deletes `heartbeat_runs` for the agent, but was written before usage-telemetry tables existed. It never cleans:
- **`cost_events`** — FKs both `agent_id` (RESTRICT) and `heartbeat_run_id` (RESTRICT)
- **`finance_events`** — same, plus `cost_event_id` (RESTRICT) → cost_events

Every real agent conversation produces a `cost_event` linked to its `heartbeat_run`, so deleting `heartbeat_runs` FK-fails. This blocked decommission for **every agent that had ever run**. The canonical never-run agents delete fine, which is why a bare-agent delete test never caught it.

> The original hypothesis (instructions-bundle / env-binding rows) was **not** the cause — neither is an `agents.id` foreign key. The blocker was cost/finance telemetry.

A schema audit of all `agents.id` and `heartbeat_runs.id` FKs found these are the only RESTRICT tables the delete path missed; everything else is `set null`/`cascade` or already handled.

## Fix (inside the `remove()` transaction)

1. Delete `finance_events` then `cost_events` (finance→cost is RESTRICT, so finance first) matching the agent **or** its `heartbeat_runs`, **before** deleting `heartbeat_runs` — mirroring the existing `activity_log` run-subquery pattern.
2. Null the remaining no-`onDelete` agent references on entities that outlive the agent (`projects.lead_agent_id`, `routines.assignee_agent_id`, `goals.owner_agent_id`, `join_requests.created_agent_id`, `approvals.requested_by_agent_id`, `approval_comments.author_agent_id`, `assets.created_by_agent_id`, `issue_thread_interactions.created_by/resolved_by`) so a used tenant agent can always be removed. "Set null" preserves those entities.

## Idempotency (asked)

Already idempotent — `getAccessibleAgent` returns **404** when the agent is gone, so a decommission retry no-ops cleanly. No client special-casing needed. Covered by a test.

## Tests

New `agent-remove-cascade.test.ts` (real embedded Postgres, real FK enforcement): creates an agent **with a heartbeat_run + cost_event + finance_event** and deletes it — the scenario a bare-agent test can't reach. Also covers cross-agent isolation (a sibling agent's telemetry survives), idempotency (second remove → null → route 404), and the bare-agent case.

- New test: **4 passed**
- Blast radius (agent lifecycle, costs, finance, approvals, dashboard, company-portability, instructions-bundle routes, permissions, cleanup-removal, cross-tenant authz): **121 passed**
- `tsc --noEmit`: clean for the change (only the pre-existing plugin-sdk baseline errors remain)

## Note on non-blocking leaks (follow-up candidates, not fixed here)

Agent delete is DB-only, so it doesn't clean the on-disk instructions-bundle files or the `company_secret_bindings` env-binding rows (polymorphic target, no agent FK — doesn't block the delete). Neither causes the 500; both are orphaned-resource tidiness for a separate change.

## Deploy / re-run

Requires a paperclip image rebuild + container recreate (this is the main app, not the governor). Once deployed, `provision-tenant.py --decommission` will delete the two live agents (which are safe as-is now — nothing was deleted by the failed run).

### DIFF
--- a/server/src/__tests__/agent-remove-cascade.test.ts
+++ b/server/src/__tests__/agent-remove-cascade.test.ts
@@ -1,0 +1,163 @@
+// Regression: deleting a tenant agent that has ACTUALLY RUN must succeed.
+//
+// Live decommission 500 (2026-07): DELETE /api/agents/:id failed with
+//   "delete on table heartbeat_runs violates foreign key constraint
+//    cost_events_heartbeat_run_id_heartbeat_runs_id_fk"
+// The agent-remove path deleted heartbeat_runs but never cleaned the
+// cost_events / finance_events that reference those runs (and the agent) —
+// tables added after the base delete handler was written. A bare-agent delete
+// test would NOT have caught this: the failure only appears once the agent has
+// a heartbeat_run with a linked cost_event, which every real conversation
+// produces. (The original hypothesis — instructions-bundle / env-binding rows —
+// was not the cause: those aren't agents.id foreign keys.)
+
+import { randomUUID } from "node:crypto";
+
+import { afterAll, afterEach, beforeAll, describe, expect, it } from "vitest";
+import {
+  agents,
+  companies,
+  costEvents,
+  createDb,
+  financeEvents,
+  heartbeatRuns,
+} from "@paperclipai/db";
+
+import { agentService } from "../services/agents.ts";
+import {
+  getEmbeddedPostgresTestSupport,
+  startEmbeddedPostgresTestDatabase,
+} from "./helpers/embedded-postgres.js";
+
+const embeddedPostgresSupport = await getEmbeddedPostgresTestSupport();
+
+describe("agentService.remove — tenant decommission cascade", () => {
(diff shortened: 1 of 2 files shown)


mrobinson2:
Mis-targeted — this belongs on the mrobinson2/paperclip fork (MRTek deploy source), not upstream. Re-opening there.

PR #7634, opened 2026-06-06
Reduce Paperclip prompt bootstrap overhead

michaelsylva:
## Thinking Path

> - Paperclip is the control plane used to run AI-agent companies through repeated heartbeat-style execution.
> - The most repeated cost in that model is instruction/bootstrap text that gets re-sent on every run before any task-specific work begins.
> - The prompt/bootstrap layer for local agents lives in the repo-level `AGENTS.md`, the adapter-level fallback prompt, the scoped wake preamble, and the default managed agent bundle.
> - API-72 exists because those repeated instruction surfaces were materially inflating token usage without adding equivalent execution value.
> - The lowest-risk way to reduce spend here is to compress repeated wording while preserving the same execution invariants around ownership, blockers, review paths, and verification.
> - This pull request trims those repeated prompt surfaces and updates the focused tests that assert the wake contract and default agent bundle content.
> - The benefit is lower recurring token overhead on Paperclip heartbeats, especially in the Paperclip repo itself where Codex auto-applies the repo `AGENTS.md` on every run.

## Linked Issues or Issue Description

- No public GitHub issue was linked for this internal Paperclip task.
- Problem: repeated bootstrap and wake boilerplate in Paperclip agent instruction surfaces was consuming avoidable tokens on every heartbeat.
- Scope: tightly scoped prompt and test compaction only; no workflow ownership, approval, blocker, or verification semantics were intentionally removed.

## What Changed

- Compressed the repo-level `AGENTS.md` so Codex picks up a materially smaller repo bootstrap in the Paperclip codebase.
- Compressed `DEFAULT_PAPERCLIP_AGENT_PROMPT_TEMPLATE` in `packages/adapter-utils/src/server-utils.ts` while preserving the required execution contract and planning/interaction rules.
- Reused a shared scoped wake execution-contract string inside `renderPaperclipWakePrompt()` to remove repeated wake boilerplate.
- Trimmed `server/src/onboarding-assets/default/AGENTS.md` while preserving the explicit `para-memory-files` and review/confirmation guidance expected by the hire flow.
- Updated focused tests for the new wake-contract wording and the default bundled-agent instruction content.

## Verification

- `pnpm exec vitest run packages/adapter-utils/src/server-utils.test.ts`
- `pnpm exec vitest run server/src/__tests__/agent-skills-routes.test.ts`

## Risks

- Low risk. The main risk is prompt wording drift changing agent behavior subtly despite preserved semantics.
[...]

PR #9200, opened 2026-07-08
fix(adapter-utils): dedupe execution contract on scoped wake prompts

octasoft-ltd:
## Thinking Path

> - Paperclip is an open-source app for running AI agents that do work, driven by repeated heartbeat runs.
> - Each heartbeat, a local agent's prompt is composed from a base template (`DEFAULT_PAPERCLIP_AGENT_PROMPT_TEMPLATE`) plus a scoped wake preamble built by `renderPaperclipWakePrompt`.
> - Both surfaces carry an "Execution contract" block describing how the agent should dispose of the issue.
> - On a fresh (non-resumed) run the adapter renders the template AND appends the wake preamble, so the contract is emitted twice — roughly 400 tokens of pure duplication on every scoped wake.
> - Tokens spent restating identical guidance are tokens not spent on the task; across many agents and heartbeats this recurs constantly.
> - The wrinkle is the resume-delta path: there the adapter suppresses the template (`renderedPrompt = ""`), so the wake preamble is the only place the contract can appear.
> - The fix is to drop the contract from the fresh wake branch (the template already carries it there) while keeping one copy on the resume-delta branch — yielding exactly one contract on every path.

- No public GitHub issue is linked.
- Problem: on a fresh scoped wake, the execution-contract block is emitted twice (once by the template, once by the wake preamble).

### Related PR (please read)

This overlaps **#7634 ("Reduce Paperclip prompt bootstrap overhead")**, which edits the same lines. The approaches differ and are worth comparing:

- **#7634** replaces both contract copies with a shorter `SCOPED_WAKE_EXECUTION_CONTRACT` string but keeps it on the fresh wake branch, so a fresh run still emits the contract **twice** (just shorter).
- **This PR** removes the contract from the fresh wake branch entirely (the template is the single source there) and keeps one copy only on the resume-delta branch, so a fresh run emits it **once**.

These two PRs conflict on the same lines; maintainers should pick one, or fold this dedup into #7634. Happy to close this in favor of #7634 if the author would rather incorporate the fresh-branch removal there.

## What Changed

- `renderPaperclipWakePrompt`: removed the "Execution contract" paragraph from the fresh wake-payload branch (the template already carries it on fresh runs).
- Kept a copy on the resume-delta branch, where the adapter suppresses the template so this is the only copy the agent receives.
- Added comments documenting the asymmetry so the two branches are not "fixed" back into duplication later.
- Tests: replaced the single "adds the execution contract" assertion with three that pin the contract appearing exactly once on the fresh, resume-delta, and non-wake composed paths.

## Verification

```
pnpm exec vitest run --project "@paperclipai/adapter-utils" src/server-utils.test.ts
```

Result on this branch: 52 passed. The new tests assert: fresh wake prompt contains no "Execution contract"; resume-delta wake prompt contains exactly one; `template + freshWake` contains exactly one; template alone (non-wake run) contains exactly one.

## Risks

- Low. Text-only change to prompt composition; no API or type changes.
- One behavioral invariant to keep in mind: removing the contract from the fresh branch is safe only because every adapter that renders the fresh wake prompt also renders the template. The resume-delta branch retains its own copy for the case where the template is suppressed.

## Model Used

- Provider: Anthropic
- Model: Claude Opus 4.8 (`claude-opus-4-8`), 1M context
- Capabilities: extended thinking, tool use

## Checklist

- [x] I included thinking path traces from project context to the change
- [x] I specified the model used (with version and capability details)
- [x] I checked ROADMAP.md and confirmed this PR does not duplicate planned core work
- [x] I searched GitHub for duplicate/related PRs and linked the related one (#7634) above
- [x] I described the issue in-PR (no existing public issue)
- [x] I did not reference internal/instance-local Paperclip issues or links
- [x] My branch name describes the change and contains no internal ticket ids
- [x] I ran tests locally and they pass
- [x] I added/updated tests covering the change
- [x] I considered and documented risks above
- [x] I will address all Greptile reviewer comments before requesting merge

### DIFF
--- a/packages/adapter-utils/src/server-utils.test.ts
+++ b/packages/adapter-utils/src/server-utils.test.ts
@@ -617,30 +617,60 @@ describe("renderPaperclipWakePrompt", () => {
     );
   });
 
-  it("adds the execution contract to scoped wake prompts", () => {
-    const prompt = renderPaperclipWakePrompt({
-      reason: "issue_assigned",
-      issue: {
-        id: "issue-1",
-        identifier: "PAP-1580",
-        title: "Update prompts",
-        status: "in_progress",
-      },
-      commentWindow: {
-        requestedCount: 0,
-        includedCount: 0,
-        missingCount: 0,
-      },
-      comments: [],
-      fallbackFetchNeeded: false,
-    });
+  const scopedWakePayload = {
+    reason: "issue_assigned",
+    issue: {
+      id: "issue-1",
+      identifier: "PAP-1580",
+      title: "Update prompts",
+      status: "in_progress",
+    },
+    commentWindow: {
+      requestedCount: 0,
+      includedCount: 0,
+      missingCount: 0,
+    },
+    comments: [],
+    fallbackFetchNeeded: false,
+  };
+
+  // The execution contract must appear exactly once in the composed prompt on every path. The
+  // adapter composes `template + wakePrompt` on fresh runs and suppresses the template on
+  // resume-delta runs, so the fresh wake prompt must NOT restate the contract while the
+  // resume-delta wake prompt MUST (it is the only copy the agent gets there).
+  const countContract = (text: string): number =>
+    text.split("Execution contract").length - 1;
+
+  it("does not restate the execution contract on the fresh wake-payload path", () => {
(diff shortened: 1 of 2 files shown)


octasoft-ltd:
@greptileai please review.

PR #9213, opened 2026-07-08
skill(plans-to-tasks): add default issue-graph template section

calebsimon-CRAFT:
## What

Adds a **`## The default issue-graph template`** section to the `paperclip-converting-plans-to-tasks` skill, inserted between `## When you're asked to plan` and `## When converting an accepted plan into tasks`.

## Why

Long, multi-step initiatives stay navigable and recoverable when their execution is shaped as **epic → parented children → one verifiable unit per child → explicit blocker chain → configure-before-assign**. This makes that shape the default starting point a planning agent reads, rather than a rule to recall. It templates the issue *graph*, not the plan's prose, so it's consistent with the skill's standing "not a plan template" position.

## Scope

- Single new section; +12 lines; no other changes.
- Upstream-generic: no company-specific keys, field names, or tooling.


calebsimon-CRAFT:
Withdrawing this contribution. The issue-graph planning template it adds is being kept as an instance-local customization for our own Paperclip deployment rather than landed in the upstream vendor skill. Closing per that scope decision; no action needed from maintainers. Thanks!

PR #9214, opened 2026-07-08
docs(skills): add default issue-graph template to converting-plans-to-tasks

calebsimon-CRAFT:
Adds a **The default issue-graph template** section to the `paperclip-converting-plans-to-tasks` skill, inserted between *When you're asked to plan* and *When converting an accepted plan into tasks*.

The section documents the issue-**graph** shape that survives long, multi-step work:

1. One epic at the top (parent, not a unit of work)
2. Children parented to the epic at creation
3. One verifiable unit per child
4. An explicit blocker chain for real sequencing (not prose / not parent-child nesting / not phase labels)
5. Configure before you assign — assign last (assignment is a wake trigger)

It templates the issue *graph*, not the plan prose, so it stays consistent with the skill's existing "not a plan template" position. Industry-agnostic — no company-specific identifiers, field names, or tooling.

Docs-only change to a single `SKILL.md`.


calebsimon-CRAFT:
Closing as a duplicate of #9213, which already lands the identical `## The default issue-graph template` section in the same file. #9213 is the canonical PR.

PR #9229, opened 2026-07-08
Fix resolved blocker wake reconciliation

nickyleach:
## Thinking Path

> - Paperclip is an open source platform for managing AI agent companies — agents pick up issues, do work, and release checkouts in a heartbeat loop
> - Issue dependency resolution is handled by `issue_blockers_resolved` wakes: when a blocking issue reaches `done`, dependent blocked issues should be woken so they can resume
> - The workspace-finalize path maintained its own special-case loop to retry missed dependency wakes at run completion, duplicating logic that already exists in the shared level-triggered reconciliation backstop
> - Additionally, the resolved-blocker reconciliation sweep was gated behind the broader liveness auto-recovery escalation setting — so instances that disabled auto-escalation creation also lost the baseline reconciliation sweep
> - This PR routes the workspace-finalize retry through the shared backstop and decouples the reconciliation sweep from the escalation creation gate
> - The benefit is simpler code (one authoritative path instead of two), and correct behaviour on instances where escalation creation is disabled

No existing GitHub issue. Describing inline:

**Bug (blocker wake reconciliation):** `issue_blockers_resolved` wakes can be missed when:

1. A `workspace_finalize` heartbeat retries dependency resolution using its own inline loop instead of the shared level-triggered backstop, making them diverge over time.
2. The `blockedByIssueIds` dependency is set (or the issue is moved to `blocked`) *after* the blocker already reached `done` — the PATCH-time wake fires on a non-blocked issue and the dependency reconciliation never catches up.
3. An assignee briefly becomes `null` between the blocker reaching `done` and the reconciliation sweep running — the sweep skips the issue and never retries.

Related PRs:
- Refs #8009 — adds dedup for `issue_blockers_resolved` re-fires (complementary; prevents over-firing; this PR ensures under-firing is caught)
- Refs #6522 — `auto-unblock dependents with no assignee` (related no-assignee edge case)

## What Changed

- **`server/src/services/heartbeat.ts`** — remove the inline dependency-wake retry loop from the `workspace_finalize` path; delegate to the shared `reconcileResolvedDependencyWakeups` helper instead
- **`server/src/services/recovery/service.ts`** — split the resolved-blocker wake reconciliation sweep out from under the `liveness_escalation_auto_recovery` feature flag; the sweep runs unconditionally while escalation *creation* remains behind the flag
- **`server/src/__tests__/issue-dependency-wakeups-routes.test.ts`** — regression: `blockedBy` set after blocker already done still triggers a reconciliation wake
- **`server/src/__tests__/heartbeat-issue-liveness-escalation.test.ts`** — regression: assignee-null churn before reconciliation; stale skipped dependency wakes do not suppress a fresh reconciliation wake

## Verification

```
pnpm exec vitest run \
  server/src/__tests__/issue-dependency-wakeups-routes.test.ts \
  server/src/__tests__/heartbeat-issue-liveness-escalation.test.ts
```

23 tests, all passing locally. `pnpm --filter @paperclipai/server typecheck` clean.

## Risks

Low risk. Bounded blast radius:
- The reconciliation sweep only considers non-hidden `blocked` issues with an agent assignee
- Uses keyset pagination with an existing 500-candidate cap
- Reuses dependency readiness/finalize gating logic unchanged
- Skips issues with existing active/queued runs and pending interactions
- Deduplicates against live/completed `issue_blockers_resolved` wakes (skipped/cancelled wakes intentionally do not suppress a fresh reconciliation)
- Observability: healed reconciliations emit `issue.blockers_resolved_wake_emitted` activity and log the healed issue ids and source

## Model Used

Claude Sonnet 4.6 (`claude-sonnet-4-6`) with tool use and extended context. Paperclip AI agent runtime.

## Checklist

### DIFF
--- a/server/src/__tests__/heartbeat-issue-liveness-escalation.test.ts
+++ b/server/src/__tests__/heartbeat-issue-liveness-escalation.test.ts
@@ -212,6 +212,7 @@ describeEmbeddedPostgres("heartbeat issue graph liveness escalation", () => {
 
   async function seedResolvedDependencyBackstopFixture(opts: {
     workspaceState?: "none" | "not_finalized" | "finalized";
+    assignee?: "agent" | null;
   } = {}) {
     const workspaceState = opts.workspaceState ?? "none";
     const companyId = randomUUID();
@@ -283,7 +284,7 @@ describeEmbeddedPostgres("heartbeat issue graph liveness escalation", () => {
         title: "Synthetic blocked dependent",
         status: "blocked",
         priority: "medium",
-        assigneeAgentId: agentId,
+        assigneeAgentId: opts.assignee === null ? null : agentId,
         issueNumber: 1,
         identifier: `${issuePrefix}-1`,
       },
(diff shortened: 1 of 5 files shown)


nickyleach:
Addressing the Greptile P2 (Wake Deduplication Race at `recovery/service.ts:4228-4231`):

This race is pre-existing code **outside this PR's diff** — `enqueueWakeup` has always lacked a DB-level unique constraint on `idempotency_key`, so concurrent callers could double-insert in theory.

**Why the impact is bounded here:**
- Both paths (workspace-finalize backstop call + periodic sweep) use the same idempotency key derived from `(dependent_issue_id, resolved_blocker_issue_id)`.
- If both enqueue the same wake, the second run that claims the wake will find the issue already unblocked and exit early — no state corruption, just one extra no-op heartbeat.
- The application-level `findExistingIssueBlockersResolvedWakeForAnyKey` check window is narrow: it only races with another call from the same company in the same subsecond window, which is uncommon in practice.

**Fix plan (separate PR):** Add a partial unique index on `(company_id, idempotency_key) WHERE idempotency_key IS NOT NULL` on `agent_wakeup_requests` and change the fast path in `enqueueWakeup` to `INSERT ... ON CONFLICT (company_id, idempotency_key) DO NOTHING`. That change is a schema migration and belongs in its own targeted PR rather than increasing the scope here.

This PR's changes do not worsen the race — they route an existing code path through the same backstop that was already subject to it.

PR #9239, opened 2026-07-08
feat(mcp): add paperclipControlIssueTree for issue-subtree pause/resume

jordi0703:
## What

Adds a `paperclipControlIssueTree` MCP tool wrapping `POST /issues/{id}/tree-holds` — pause, resume, cancel, or restore an issue and its subtree. The server routes, service, and `issueTreeControlModeSchema` validator already live on `master`; this exposes them to MCP-driven agents.

## Why

Subtree pause holds gate checkout and dispatch (`409 "Issue checkout blocked by active subtree pause hold"`), but no MCP tool exposes the tree-holds endpoint. An agent that hits the 409 today has to guess raw API paths through the generic request tool — and the hold model's create-with-mode semantics (a pause is cleared by *creating* a hold with mode `resume`; there is no delete/unpause endpoint) make that near-undiscoverable. We watched an agent probe `/holds`, `/subtree-holds`, `/pause-holds`, and `DELETE .../holds/{id}` — all 404 — before giving up.

The tool description spells out the resume-clears-pause semantics so agents pick the right mode without trial and error.

## Testing

- 3 new unit tests (resume clears a pause hold + body shape, optional `reason` omitted when absent, unknown mode rejected by the shared enum) — `pnpm --filter @paperclipai/mcp-server test`: 16/16 pass.
- Typecheck clean.

The tool is running in our deployment (fork-built) where it has already been used to pause a subtree whose runs were destabilizing the host, and to resume it after remediation.

### DIFF
--- a/packages/mcp-server/README.md
+++ b/packages/mcp-server/README.md
@@ -62,6 +62,7 @@ Write tools:
 - `paperclipUpdateIssue`
 - `paperclipCheckoutIssue`
 - `paperclipReleaseIssue`
+- `paperclipControlIssueTree`
 - `paperclipAddComment`
 - `paperclipSuggestTasks`
 - `paperclipAskUserQuestions`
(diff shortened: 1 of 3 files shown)


jordi0703:
Withdrawn.

PR #9256, opened 2026-07-09
fix(heartbeat): restore AA-1611 no-progress chain-wake guard (AA-1924)

adacovsk:
## Problem (AA-1924)

Worker runs failed with `numTurns is not defined` ReferenceError at `server/src/services/heartbeat.ts`. The identifier was referenced in a `logger.info` payload but never declared in scope, so the AA-1611 N-cycle-guard branch (fired when a run makes zero tool calls) threw — converting a benign "skip chain-wake" into a hard run failure.

## Root cause

Merge `b05893c6a` mis-resolved the AA-1611 conflict. It kept the **pre-guard** `nextTask` query (`in_progress`/`todo`, no `ne(issueId)`) as the outer `if`, grafted the skip-log into that branch, and dropped the `numTurns` / `lastRunMadeNoProgress` computation entirely. Two defects resulted:

1. **ReferenceError** — `numTurns` referenced but never bound.
2. **Inverted control flow** — chain-wake suppressed whenever a `todo`/`in_progress` task existed and only fired via the broader `in_review` set (opposite of the intended queue-drain); the N-cycle guard was gone.

A naive "just delete `numTurns`" fix stops the crash but cements defect #2.

## Fix

Restore the block verbatim from guard commit `8b67de46b`: compute `numTurns` from `adapterResult.resultJson?.num_turns`, gate the skip on `lastRunMadeNoProgress` (`num_turns <= 1`), and chain-wake a *different* runnable task otherwise. The touched region is now byte-identical to `8b67de46b`.

## Verification

- `tsc --noEmit` on `server/` passes clean.
- Diff of the touched region against `8b67de46b` is empty.


adacovsk:
Closing — created against the wrong base repo. adacovsk/paperclip has diverged from paperclipai/paperclip (its own agent-pipeline line); this fix targets fork commits (8b67de46b / merge b05893c6a) and belongs on adacovsk/paperclip:master, which the local server runs. Reopening there.

PR #9273, opened 2026-07-09
fix: keep confirmation details available after confirm/deny

lacymorrow:
## Summary

When a `request_confirmation` interaction (plan approval or similar dialogue) is resolved, the prompt and `detailsMarkdown` disappear entirely — the board can no longer see what was just approved or declined. This renders the existing resolution summary ("Confirmed" / "Declined" + target chip + decline reason) plus a collapsible **Show details** section, so the plan content stays accessible after the decision instead of vanishing.

Reported in lacymorrow/paperclip#16 (tracked internally as LAC-1291).

## Changes

- `RequestConfirmationCard`: resolved (non-pending) state now renders the resolution summary followed by a `Collapsible` with the original prompt and `detailsMarkdown` (via `MarkdownBody` with `externalReferences`), collapsed by default.
- The checkbox-confirmation variant is untouched — its resolution already retains the selected options.

## Testing

- Added two regression tests: resolved (accepted and rejected) confirmations expose a "Show details" toggle that reveals the prompt and details markdown; both fail without the fix.
- `npx vitest run src/components/IssueThreadInteractionCard.test.tsx` — 18/18 pass.
- `tsc -b` in `ui/` — clean.

### DIFF
--- a/ui/src/components/IssueThreadInteractionCard.test.tsx
+++ b/ui/src/components/IssueThreadInteractionCard.test.tsx
@@ -484,4 +484,61 @@ describe("IssueThreadInteractionCard", () => {
       "![bug.png](https://cdn.example/shot.png)",
     );
   });
+
+  it("keeps confirmation details available behind a toggle after acceptance", async () => {
+    const host = renderCard({
+      interaction: {
+        ...pendingRequestConfirmationInteraction,
+        status: "accepted",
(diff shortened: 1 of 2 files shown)


lacymorrow:
Superseded by #9274 — same change re-opened from a branch name compliant with the contribution checklist (no internal ticket ids) and with the full PR template. Closing this one.

PR #9291, opened 2026-07-09
fix: bridge sandbox runtime services

jelsco:
## Thinking Path

> - Sandbox-backed local adapters let agents run in isolated environments while still reporting work back to the Paperclip control plane.
> - The existing sandbox callback bridge made Paperclip API calls reachable from socket-restricted sandboxes, but workspace runtime services were still exposed as host-local URLs.
> - Any sandbox that cannot open direct host sockets would fail when instructions or tooling tried to call a localhost runtime service such as a queue or preview server.
> - Runtime services are already first-class in heartbeat context, so the bridge should translate those localhost services into sandbox-local listeners instead of asking agents to know host networking details.
> - This pull request extends the sandbox callback bridge to expose fixed-target runtime service listeners on their expected local ports, while keeping the Paperclip API bridge authenticated and route-allowlisted.
> - The benefit is that sandboxed agents can use runtime service URLs the same way local agents do.

No public GitHub issue exists for this instance-specific bug, so the problem is described inline using the bug-report template fields.

### What happened?

Sandbox-backed local adapter runs received localhost runtime-service URLs from the host context, but direct socket access to those host-local endpoints can be unavailable from the sandbox. The Paperclip API bridge was reachable, while separate runtime services such as queue or preview endpoints still failed with connection-refused or socket-permission errors.

### Expected behavior

A sandbox-backed local adapter run should be able to call Paperclip-provided runtime services through sandbox-local URLs, without needing direct host-network access or adapter-specific endpoint rewrites in the agent instructions.

### Steps to reproduce

1. Start a sandbox-backed local adapter heartbeat that receives a localhost runtime service in its runtime context.
2. From inside the sandboxed run, call the service health endpoint using the injected runtime-service URL.
3. Observe the request fail when the sandbox cannot open direct host sockets, even though Paperclip API calls can use the callback bridge.

Reproduced against Paperclip development builds before this change. This PR is based on `9972b01a8` atop upstream `97e075215`.

### Deployment mode

Local dev with sandbox-backed local adapters.

## What Changed

- Added fixed-target runtime-service support to the sandbox callback bridge request protocol.
- Started sandbox-local listeners for localhost runtime services on their configured ports and forwarded those requests through the existing host-side queue bridge.
- Rewrote bridged runtime service URLs in `PAPERCLIP_RUNTIME_SERVICES_JSON` and `PAPERCLIP_RUNTIME_PRIMARY_URL` for Codex and Claude local sandbox runs.
- Added a regression test proving a sandbox-local runtime service `/health` request forwards to the host service without Paperclip API auth headers.

## Verification

- `pnpm vitest run packages/adapter-utils/src/execution-target-sandbox.test.ts`
- `pnpm --filter @paperclipai/adapter-claude-local typecheck`

Runtime verification depends on an instance-managed sandbox adapter and runtime service. The code-level regression proves the bridge behavior; deploy/run verification should confirm the sandbox receives `PAPERCLIP_API_URL` through the callback bridge and runtime services rewritten to sandbox-local listener URLs.

## Risks

- Runtime service listeners are unauthenticated inside the sandbox to preserve plain localhost service behavior, but each listener forwards only to one fixed service URL selected by the host-side runtime context.
- If a sandbox already has a process listening on the same runtime-service port, bridge startup for that service will fail. That mirrors the local-port contract the agent was already expected to use.

## Model Used

- OpenAI GPT-5 via Codex CLI, tool-assisted coding and test execution.

## Checklist

### DIFF
--- a/packages/adapter-utils/src/execution-target-sandbox.test.ts
+++ b/packages/adapter-utils/src/execution-target-sandbox.test.ts
@@ -614,6 +614,79 @@ describe("sandbox adapter execution targets", () => {
     }
   });
 
+  it("bridges localhost runtime services on their sandbox port", async () => {
+    const rootDir = await mkdtemp(path.join(os.tmpdir(), "paperclip-execution-target-runtime-service-bridge-"));
+    cleanupDirs.push(rootDir);
+    const remoteCwd = path.join(rootDir, "workspace");
+    const runtimeRootDir = path.join(remoteCwd, ".paperclip-runtime", "codex");
+    await mkdir(runtimeRootDir, { recursive: true });
+
+    const requests: Array<{ method: string; url: string; auth: string | null; runId: string | null }> = [];
+    const serviceServer = createServer((req, res) => {
+      requests.push({
+        method: req.method ?? "GET",
+        url: req.url ?? "/",
+        auth: req.headers.authorization ?? null,
+        runId: typeof req.headers["x-paperclip-run-id"] === "string" ? req.headers["x-paperclip-run-id"] : null,
+      });
+      res.writeHead(200, { "content-type": "application/json" });
+      res.end(JSON.stringify({ status: "ok", service: "linkedin-queue" }));
+    });
+    await new Promise<void>((resolve, reject) => {
+      serviceServer.once("error", reject);
+      serviceServer.listen(0, "127.0.0.2", () => resolve());
+    });
+    const address = serviceServer.address();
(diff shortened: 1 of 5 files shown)


jelsco:
Closing per ReplenishRadar RR-4727. ReplenishRadar does not maintain upstream paperclipai/paperclip and has no write access here. We are consolidating platform contributions onto our own fork (jelsco/paperclip). The source branch is preserved. Apologies for the noise.

PR #9307, opened 2026-07-09
fix(TWX-1309): durable scheduled retry instead of blocked-park on double master-runtime limit

TueJon:
Fixes the "issue dies 78ms after creation" amplifier from TWX-1309: when both master runtimes (Claude + Codex) are quota-limited, the heartbeat parked the issue `blocked` with no blocker relation (never auto-resumes) and hard-failed the agent — turning a seconds-long quota overlap into a permanent stall.

## What changed

**1. `master_runtime_all_limited` is now a transient agent error** (cherry-pick of `40ae26af` from the live host, so this PR is self-contained). The TWB-305 reconciler auto-clears the agent on its bounded backoff instead of leaving it in `status: error` until a manual reset.

**2. Both double-limit sites queue a durable scheduled retry instead of parking `blocked`:**
- **Pre-dispatch** (`masterRuntimeResolution.blocked`) and **post-execution** (`limitRecord.fallbackLimited` in auto mode) now call `scheduleBoundedRetryForRun` with a new `master_runtime_all_limited_retry` reason.
- The retry is due at the **earliest cooldown expiry** (`claudeLimitedUntil`/`codexLimitedUntil` — whichever runtime frees up first is usable via auto failover), plus 15s grace and 0–120s jitter so retries that all saw the same expiry instant don't stampede the freshly reset provider.
- Scheduling runs **before** `releaseIssueExecutionAndPromote`, so the retry inherits the issue execution lock atomically (release's conditional clear then no-ops).
- Single wait capped at 6h, budget of 6 attempts — week-long provider caps re-check periodically rather than sleeping the full window.
- Escalation to the old `blocked` park + comment happens **only when the budget is exhausted** (or failover mode is forced), via the shared `parkIssueBlockedForMasterRuntimeExhaustion` helper.

**3. Issue-scoped dedupe in the retry transaction.** Twin runs failing the same double-limit milliseconds apart (the exact TWX-1309 failure: two CTO runs 13ms apart) now yield exactly **one** live retry per issue — the `FOR UPDATE` row lock serializes them and the second coalesces onto the first's wakeup. Guards for the new reason tolerate `in_review`/`todo` issues (a comment wake needs no checkout) but refuse to shadow an execution lock held by another live run; `max_turns_continuation` behavior is unchanged (same conditions, same messages).

## Verification

- `pnpm typecheck` (server) clean.
- `vitest run`: `heartbeat-master-runtime-failover` (incl. 4 new tests for the delay helper), `heartbeat-retry-scheduling`, `run-continuations`, `issue-scheduled-retry-routes`, `recovery-routine-continuation-guard`, `heartbeat-process-recovery` — 125 tests passing.

## Deployment note

The live service tree (`/home/vd/dev/paperclip-live`) currently sits on `manual-sync/TWX-1312`, which does **not** contain `40ae26af` — see [TWX-1309](https://github.com/tue-Jonas/paperclip) thread for the restart-timing implication.


TueJon:
Opened against upstream by mistake — this belongs on the tue-Jonas fork. Apologies for the noise.

PR #9309, opened 2026-07-09
docs(deploy): dev-plane branch/dirty-tree drift guard (TWX-1314)

TueJon:
## Summary
- Dev-plane control planes (e.g. `paperclip.service` on vd-fw) run TypeScript source directly via `tsx` from whatever branch/working-tree state the checkout is in — there's no build/tag step pinning "the deploy". A near-miss during the TWX-1309 RCA: a hotfix commit believed deployed was on a branch the checkout had since moved away from; the next scheduled restart would have silently dropped it.
- Adds `scripts/deploy/check-deploy-state.sh` (one-liner status: branch/HEAD/dirty vs. last intentionally marked state, `--enforce` to exit non-zero on drift) and `scripts/deploy/mark-deploy.sh` (record the intended state after an intentional switch).
- Documents this in `docs/deploy/dev-plane-restart-hygiene.md` as a new section, alongside the existing restart-burst guidance (distinct failure mode: *what* runs vs. *when* it restarts).

Note: `docs/deploy/dev-plane-restart-hygiene.md` didn't exist yet on this fork's `master` (it landed upstream via the in-flight `manual-sync/TWX-1312` sync) — added here as a full file so the PR applies cleanly; the pre-existing content is reproduced as-is, only the new section is added.

## Not included (needs sign-off before applying)
Wiring `check-deploy-state.sh` into the live `paperclip-restart.service`/`paperclip.service` systemd units on vd-fw as a non-blocking `ExecStartPre` — that's a prod host-config change and is called out in the doc as requiring explicit confirmation before installing, separate from this PR.

## Test plan
- [x] Ran both scripts standalone in an isolated worktree: `check-deploy-state.sh` correctly reports `UNKNOWN` with no marker, `OK` after `mark-deploy.sh`, and flags branch/HEAD/dirty drift after moving away from the marked state.
- [ ] Jonas: confirm content is compatible with the in-flight TWX-1312 sync doc draft before merge.


TueJon:
Wrong base — recreating against tue-Jonas/paperclip (this fork's own master, where this project actually deploys from), not upstream.

Issue #9310, opened 2026-07-09
runJob dispatches never carry an invocation scope — scheduled-job host calls fail whenever a scoped invocation is concurrently active


Tonton0:
## Summary

`runJob` host→worker dispatches never register an invocation scope, so any host call a scheduled job handler makes (`state.get`, `agents.list`, …) is rejected with `"…missing, expired, or unknown invocation scope"` **whenever any scoped invocation (`onEvent`/`executeTool`) happens to be active at the same moment**. On quiet instances jobs appear to work; under real event traffic they fail intermittently. This is a host-side gap that affects **every plugin with scheduled jobs, even on a current SDK** — unlike #7897, where the cause is an SDK that doesn't echo `paperclipInvocationId`.

Likely the actual mechanism behind the "job invocations don't always carry a company-scoped invocation context" observation in #8845, and related to the symptom family in #8633.

## Environment

- `paperclipai@2026.626.0` (self-hosted, Docker, Linux)
- `paperclip-plugin-discord@0.9.1` (jobs `check-escalation-timeouts` and `check-budget-thresholds`, both `*/5 * * * *`)
- Plugin SDK in the plugins install dir: `@paperclipai/plugin-sdk@2026.529.0` (generically wraps any host→worker request in `invocationContextStorage.run(request.paperclipInvocation, …)` — so the SDK side is fine)

## Root cause

`server/src/services/plugin-worker-manager.ts`:

1. `deriveInvocationScope(method, params)` has cases for `performAction`, `executeTool`, `onEvent` (plus a top-level `params.companyId` check) — **but no case for `runJob`**. Instance-scoped jobs carry no `companyId`, so the function returns `null`, no invocation is registered, and the outgoing `runJob` request carries no `paperclipInvocation`.
2. The job handler's nested worker→host calls therefore have no `paperclipInvocationId` to echo. `contextForWorkerMessage()` then applies the strict fallback:

```ts
if (!invocationId) {
  const hasActiveInvocation = activeInvocations.size > 0 ||
    Array.from(pendingRequests.values()).some((pending) => pending.invocationId);
  return hasActiveInvocation ? { invalidInvocationScope: true } : {};
}
```

So the very same job call succeeds when the worker is otherwise idle, and is denied (`InvocationScopeDeniedError`) when an `onEvent`/`executeTool` invocation is in flight — the id-less job call is indistinguishable from a call trying to escape a scoped invocation.

## Evidence

[...]

PR #9368, opened 2026-07-10
fix(plugins): register an unrestricted invocation scope for runJob dispatches

adamteale:
## Thinking Path

> - Plugins run in worker processes and communicate with the host via an invocation-scoped messaging protocol; `contextForWorkerMessage` maps an echoed invocation id back to a company scope so the SDK gate can decide whether to admit the call
> - `runJob` dispatches previously registered no invocation id — so a job's own nested host calls had nothing to echo. When any concurrently active company-scoped invocation was in flight, the no-id fallback returned `{invalidInvocationScope: true}`, silently denying instance-scoped calls from jobs
> - This was invisible on quiet instances and intermittent under real event traffic (a plugin's scheduled job calling `ctx.state.get({ scopeKind: "instance" })` was denied whenever another company event ran at the same time)
> - The fix needs two coordinated parts: register a real-but-unrestricted invocation scope for every `runJob` dispatch (so jobs always have an id to echo), and teach `contextForWorkerMessage` to apply proactive company resolution when it sees the empty-companyId sentinel, so legitimate company calls are still admitted while unconfigured companies are still denied
> - The benefit is that scheduled jobs can reliably call instance-scoped and configured-company-scoped host tools without intermittent denials, while the security invariant (unconfigured companies are denied) is preserved

Refs #11163 (related follow-up issue tracking remaining company-scoped call kinds)

No separate standalone issue. Following the bug-report template fields:

**Environment:** Any Paperclip instance with a plugin that uses `runJob` and makes nested host calls (e.g. `ctx.state.get`, `ctx.agents.list`)

**Describe the bug:** `runJob` dispatches registered no invocation scope. When another company-scoped invocation was concurrently active, `contextForWorkerMessage` returned `{invalidInvocationScope: true}` for the job's nested calls, causing them to be denied. The failure was intermittent (only when concurrent invocations existed) and produced no error log — the call was simply rejected by the SDK gate.

**Expected behavior:** Scheduled job calls to instance-scoped tools (e.g. `ctx.state.get(scopeKind:"instance")`) and configured-company-scoped tools succeed regardless of concurrent event traffic.

**Actual behavior:** Those calls are intermittently denied when another company-scoped invocation is active at the same moment.

**Steps to reproduce:** Install a plugin with a `runJob` that calls `ctx.state.get({ scopeKind: "instance" })`. Under load (or with a concurrent `onEvent` handler active), the call is denied.

- [x] I have searched for similar PRs and found none beyond the related #11163

## What Changed

- `server/src/services/plugin-worker-manager.ts` — `deriveInvocationScope`: registers a real unrestricted invocation (`{ companyId: "" }`) for every `runJob` dispatch so the worker always has an id to echo on nested calls
- `server/src/services/plugin-worker-manager.ts` — `contextForWorkerMessage`: when the echoed invocation has an empty `companyId` (the `runJob` sentinel), applies proactive company resolution:
  - `kind:"none"` calls (instance/agent-scoped): returns `{ companyId: "" }` — SDK gate kind-none fast-path admits unconditionally
  - `kind:"single"` for a **configured** company: resolves to that company's explicit scope — SDK gate single-company match admits it
  - `kind:"single"` for an **unconfigured** company: returns `{ invalidInvocationScope: true }` — SDK gate denies (security invariant preserved)
- `server/src/__tests__/plugin-worker-manager.test.ts` — three new integration tests using `createHostClientHandlers` as `hostHandlers` (exercising the real SDK gate end-to-end)
- `server/src/__tests__/fixtures/plugin-worker-invocation-scope.cjs` — test fixture

**Call kinds now unblocked (from original):**

| Call kind | SDK gate result |
|-----------|-----------------|
| `kind:"none"` (instance/agent scope, e.g. `state.get(scopeKind:"instance")`) | ✅ passes |
| `kind:"single"` for a **configured** company | ✅ passes |
| `kind:"single"` for an **unconfigured** company | ❌ denied (correct) |
| `kind:"all"` (e.g. `companies.list`) | ✅ passes |

Note: `config.get` and `secrets.resolve` require explicit per-company iteration from a job — they are not blanket-unblocked, and this is by design.

## Verification

Three new integration tests in `plugin-worker-manager.test.ts` using `createHostClientHandlers` as real `hostHandlers` (not a mock), exercising `requireInvocationCompanyScope` end-to-end:

1. Instance-scoped call (`state.get(scopeKind:"instance")`) succeeds while another scoped invocation is registered
2. `agents.list` for a configured company succeeds
3. `agents.list` for an unconfigured company is denied (regression test — pre-fix this silently passed)

The existing regression test is updated to reflect the new context shape (proactive resolution now passes the company's explicit scope instead of the empty sentinel).

All CI checks passing.

## Risks

- **Security invariant preserved**: unconfigured-company calls are still denied — the fix closes a pre-existing gap where the no-id path silently allowed them
- **No breaking changes**: the change is internal to `contextForWorkerMessage`; no plugin API surface changes
- **Proactive resolution scope**: the empty-companyId sentinel is specific to `runJob` dispatches; all other invocation kinds continue to use their existing resolution path unchanged
- **`config.get`/`secrets.resolve` still require per-company iteration**: this is documented in the PR and the related issue #11163 tracks potential future improvements

## Model Used

Claude Sonnet 4.6 (`claude-sonnet-4-6`) — Claude Code CLI, tool-use mode.

## Checklist

- [x] I have searched GitHub for duplicate or related PRs and linked them above (Refs #11163)
- [x] My branch name describes the change (`fix/runjob-invocation-scope`) and contains no internal ticket ids


DTTerastar:
Heads-up on a limit of this fix: `{ companyId: "" }` unblocks *unscoped* nested calls, but **not `config.get` or `secrets.resolve`**, so plugins that need credentials inside a job tick are still blocked after this lands.

`requestedCompanyScope` classifies any call carrying a non-empty `params.companyId` as `kind: "single"` (`host-client-factory.ts` L546-547), and `resolveRequiredCompanyId` then requires a non-empty *scoped* company:

```ts
const scopedCompanyId = readNonEmptyString(context?.invocationScope?.companyId);
if (requested.kind === "single") {
  if (!scopedCompanyId) throw new InvocationScopeDeniedError(pluginId, method, "company context is required");
```

`readNonEmptyString("")` is null — the same falsiness this PR relies on to stay unrestricted — so `ctx.config.get(companyId)` from a job still throws `company context is required`. `ctx.state.get` / `ctx.issues.*` work fine, which is what the added regression test covers (`companies.get` is `kind: "all"`).

Not an objection to the PR — the intermittent-failure fix stands on its own. Flagging it because "scheduled jobs can reliably call host APIs" reads as broader than it is, and a plugin author following #9310 → this PR will hit the wall one layer down. A job that needs its own config would need a real per-company scope (deriving one company per tick, or a fan-out per configured company) rather than the unrestricted sentinel.

Verified against `v2026.720.0` on a self-hosted instance while debugging mvanhorn/paperclip-plugin-telegram#77.

adamteale:
Friendly nudge — all CI checks are passing on this PR. Would appreciate a review when you have a chance. Happy to address any feedback. 🙏

fibleep:
@DTTerastar's note above is right that `{ companyId: "" }` doesn't fully unblock company-scoped nested calls — but it also affects `ctx.state.get`/`ctx.agents.list`, not just `config.get`/`secrets.resolve`. Ran this against the current shipped SDK build (`@paperclipai/plugin-sdk@2026.722.0`) to check exactly which calls are still denied after this PR's fix:

```js
import { createHostClientHandlers } from "@paperclipai/plugin-sdk";

const handlers = createHostClientHandlers({
  pluginId: "test-plugin",
  capabilities: ["plugin.state.read", "agents.read"],
  services: {
    state: { get: async (p) => ({ ok: true, p }) },
    agents: { list: async (p) => ({ ok: true, p }) },
  },
});

// Simulates the context this PR produces for a runJob dispatch.
const afterFix = { invocationScope: { companyId: "" } };

await handlers["state.get"]({ scopeKind: "company", scopeId: "company-A", stateKey: "x" }, afterFix);
// -> throws: Plugin "test-plugin" is not allowed to perform "state.get": company context is required

await handlers["agents.list"]({ companyId: "company-A" }, afterFix);
// -> throws: Plugin "test-plugin" is not allowed to perform "agents.list": company context is required

await handlers["state.get"]({ scopeKind: "instance", stateKey: "x" }, afterFix);
// -> succeeds (kind:"none", bypasses the check entirely)
```

`agents.list`'s params require `companyId` unconditionally (`protocol.ts` — `"agents.list": [params: { companyId: string; ... }, ...]`), so every call is classified `kind: "single"` by `requestedCompanyScope`. Same for `ctx.state.get({ scopeKind: "company", scopeId })`. Both go through `requireInvocationCompanyScope`'s `"single"` branch, which — same as `resolveRequiredCompanyId` — does `if (!allowedCompanyId) throw "company context is required"`, and `readNonEmptyString("")` is null, so the unrestricted job scope this PR registers doesn't satisfy it. `state.get` only "works fine" when a plugin calls it with a non-`"company"` `scopeKind` (e.g. `"instance"`, `"agent"`), which is a `kind:"none"` request that bypasses the check — that's a different usage pattern than a job that iterates per company.

Concretely: `mvanhorn/paperclip-plugin-discord`'s `check-escalation-timeouts` and `check-budget-thresholds` jobs (both `*/5 * * * *`) call `ctx.agents.list({ companyId })` and `ctx.state.get({ scopeKind: "company", scopeId, ... })` every tick, and get denied with `company context is required` on every run, not just under concurrency — the "single"-kind branch doesn't care whether a scope is present-but-empty or absent, it treats them the same.

This PR is still worth merging as-is (it fixes the real intermittent bug in #9310 for `kind:"none"` calls), but plugins with jobs that touch specific companies will still hit a wall after it lands. Filed the company-scoped-call gap as a separate, more precisely-reproduced issue so it doesn't get lost: https://github.com/paperclipai/paperclip/issues/11163

adamteale:
Addressing the three reviewer asks from @commitperclip. Branch rebased onto current master first (1 commit replayed cleanly).

**Ask #1 — Rework the fallback logic in `contextForWorkerMessage`**

Added proactive company resolution for invocations with an empty `companyId` (the runJob sentinel). When a nested call arrives echoing a runJob invocation ID, the function now:
- Calls `referencedCompanyId()` — same helper already used by the no-id proactive path.
- If the call references a company **in** `proactiveCompanyScopes`: returns `{ invocationScope: { companyId: configuredId } }` — SDK gate admits it via single-company match.
- If the call references a company **not** in `proactiveCompanyScopes`: returns `{ invalidInvocationScope: true }` — SDK gate denies it. This closes the security gap in the original PR (pre-fix, the no-id path with no concurrent invocation silently allowed unconfigured companies).
- If `referencedCompanyId()` returns null (kind-"none" call): falls through to the existing return, passing the unrestricted `{ companyId: "" }` scope — SDK gate's kind-"none" early exit admits it.

**Ask #2 — Add a test through the real SDK gate (`createHostClientHandlers`)**

Added a `describe` block with three `it` subtests that use `createHostClientHandlers` as `hostHandlers` (not a mock), so `requireInvocationCompanyScope` is exercised end-to-end:
1. **Instance-scoped call succeeds while another scoped invocation is registered** — `state.get(scopeKind:"instance")` from runJob passes; `Promise.all([getData, runJob])` registers both invocations simultaneously to mirror the concurrent-load shape.
2. **Configured company call succeeds** — `agents.list({ companyId: "company-configured" })` from runJob resolves and the service is reached with the right params.
3. **Unconfigured company is denied** — `agents.list({ companyId: "company-other" })` rejects with `INVOCATION_SCOPE_DENIED` and the service is never called.

Also updated the existing regression test: it now sets `proactiveCompanyScopes: ["company-a"]` and expects `invocationScope: { companyId: "company-a" }` (explicit scope) rather than the empty sentinel — reflecting the new context shape.

**Ask #3 — PR description updated**

Updated the PR body to precisely state which call kinds are unblocked, with a table covering kind-"none", configured/unconfigured kind-"single", and kind-"all" — per @DTTerastar's / @fibleep's comments and #11163.

PR #9190, opened 2026-07-07
Fix request storm polling and issue-list coalescing

cryppadotta:
## Thinking Path

> - Paperclip is the open source app people use to manage AI agents for work
> - The board UI keeps issue, agent, activity, and run state fresh through polling across several pages and sidebar surfaces
> - When multiple components or browser tabs poll the same company data at the same time, the API can receive bursts of duplicate issue-list requests
> - Those duplicate requests increase database and server load without returning meaningfully different data
> - This pull request adds server-side compression/coalescing plus client-side visibility-aware and cross-tab shared polling
> - The benefit is lower request volume during normal board usage while preserving fresh UI data for active users

## Linked Issues or Issue Description

No exact public GitHub issue was found for this request.

Problem:
- The board can issue redundant polling requests for the same issue-list data from multiple UI surfaces and tabs.
- In busy operator sessions, those bursts can trigger request-storm behavior and unnecessary issue-list load.
- Expected behavior is to reuse identical in-flight work server-side and reduce hidden-tab or duplicate-tab polling client-side while preserving normal refresh behavior.

Related public context found during duplicate search:
- #8206 covers a different board UI 404-storm scope.
- #5165 covers separate issue-list behavior around page-size truncation.

## What Changed

- Added API compression middleware foundation and a company/created-at index for heartbeat run access.
- Added server-side issue-list request storm detection and identical in-flight request coalescing.
- Added UI fetch metadata, visibility-aware polling, and request deduplication for issue/activity/client calls.
- Added cross-tab shared polling primitives and wired them into the sidebar, inbox, dashboard, issue, project, routine, and agent surfaces.
- Resolved the latest `master` migration collision by keeping upstream `0140_built_in_managed_resources.sql` and renumbering this branch's heartbeat-run index migration to `0141_heartbeat_runs_company_created_at_index.sql`; the SQL uses `CREATE INDEX IF NOT EXISTS` for idempotency.
- Stabilized server heartbeat cleanup tests exposed by the PR check matrix.
- Fixed the Greptile compression follow-up by weakening strong ETags on encoded JSON responses and bypassing compression for streamed/download responses.

## Verification

- `pnpm exec vitest run ui/src/components/IssuesList.test.tsx ui/src/pages/Inbox.test.tsx` — passed after resolving the latest `master` conflict in `IssuesList.tsx` and updating the 200-result cap expectations.
- `pnpm check:token-gates` — passed after the UI conflict resolution.
[...]

PR #9379, opened 2026-07-10
Fix apiCompression corrupting/dropping Better Auth responses for gzip clients

cryppadotta:
## Summary

Standalone port of the PAP-13466 login fix to master (PAP-13485). The `apiCompression` middleware (from #9190) corrupts or drops every Better Auth response for clients that send `Accept-Encoding: gzip` or `deflate` — i.e. all real browsers — while curl (no `Accept-Encoding`) works, which made this easy to misdiagnose.

Two compounding bugs in `server/src/middleware/api-compression.ts`:

1. **Body corruption**: buffered `res.write()` chunks were converted with `String(chunk)`. Better Auth (via better-call) streams `Uint8Array` chunks, which `String()` serializes as comma-separated decimal byte values — corrupting the JSON body and inflating it ~3.4x.
2. **Connection destroyed**: better-call commits headers with `writeHead()` before streaming. When the inflated body crossed the 1024B compression threshold, `setHeader()` threw `ERR_HTTP_HEADERS_SENT`, and the catch block called `res.destroy()` — zero bytes on the wire. Browsers surfaced `net::ERR_EMPTY_RESPONSE` / "Failed to fetch" on sign-in.

## Fix

- Convert `Uint8Array` chunks with `Buffer.from()` instead of `String()`.
- Pass through uncompressed once headers are already sent (`writeHead`-style streaming responses).
- On compression failure, fall back to writing the original uncompressed body instead of destroying the connection.

## Testing

- `pnpm vitest run src/__tests__/api-compression.test.ts` — 10/10 passing, including 3 new regression tests covering `Uint8Array` chunk handling, the headers-already-sent pass-through, and the no-destroy fallback.

Cherry-pick of `257c9dc89` (verified on the PAP-10341 execution workspace, where it fixed browser sign-in). Tracked by PAP-13485, parent PAP-13466.

### DIFF
--- a/server/src/__tests__/api-compression.test.ts
+++ b/server/src/__tests__/api-compression.test.ts
@@ -105,6 +105,27 @@ function buildApp() {
     res.write(chunk.slice(0, chunk.length / 2));
     res.end(chunk.slice(chunk.length / 2));
   });
+  // Mirrors better-call's setResponse (Better Auth sign-in/sign-up): headers
+  // are committed with writeHead() first, then the web-stream body arrives as
+  // Uint8Array chunks.
+  app.get("/api/auth-bridge", (req, res) => {
+    const body = JSON.stringify({
+      token: "t".repeat(Number(req.query.pad ?? 0)),
+      user: { name: "Dotta", email: "dotta@example.test" },
+    });
+    res.setHeader("content-type", "application/json");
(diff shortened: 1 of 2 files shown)


cryppadotta:
Superseded by #9381 — same commits, reopened from a branch name that complies with CONTRIBUTING.md (no internal ticket ids in branch names) and with the full PR template. Greptile reached 5/5 with zero unresolved comments on this diff (head 7d75a3264) before the move.

PR #9412, opened 2026-07-11
fix(worktree): never adopt non-worktree targets in worktree repair (PAP-13539)

cryppadotta:
## Problem

`~/.paperclip/instances/default/.env` kept getting rewritten with:

```
# Generated by Paperclip worktree repair
PAPERCLIP_IN_WORKTREE="true"
PAPERCLIP_WORKTREE_NAME="PAP-884-ai-commits-component"
```

which makes the main systemd service boot in worktree mode (run engine suppressed) on the next restart.

## Root cause (reproduced deterministically)

`maybeRepairLegacyWorktreeConfigAndEnvFiles()` runs as an import side effect of `server/src/config.ts` and trusted **ambient** `PAPERCLIP_IN_WORKTREE` while resolving the config path independently. If the flag leaks into a process whose config resolution lands on the **real default instance**, the "repair" rewrites the main instance `.env`.

The recurring trigger is the server test suite itself: `worktree-config.test.ts` › *"rebalances duplicate ports for already isolated worktree configs"* sets `PAPERCLIP_IN_WORKTREE=true` but does **not** clear the ambient `PAPERCLIP_CONFIG` that agent shells on the dev box export (it points at the live default instance). Running the suite from such a shell rewrote the machine's real `instances/default/.env` on every run — with the test's fixture worktree name, which is literally `PAP-884-ai-commits-component`. That's why a "super old" branch name kept reappearing. Verified live: running the suite on unpatched master contaminated the file (mtime matched the test run to the second); running it with this patch left it untouched.

## Fix

1. `resolveWorktreeRuntimeContext` (used by both the repair and runtime-port persistence) now refuses to adopt a target unless:
   - the config sits in a worktree layout `<root>/.paperclip/config.json` (instance-store configs like `instances/default/config.json` are categorically excluded), and
   - the target's own persisted `.env` already declares `PAPERCLIP_IN_WORKTREE=true` (written by worktree provisioning), so ambient/leaked flags alone can never redirect the writer.
2. The test suite scrubs all ambient `PAPERCLIP_*` env in a `beforeEach`, so it is hermetic no matter which shell runs it (this also fixes the test, which was failing on dev boxes with exported `PAPERCLIP_CONFIG`).
3. New regression tests: leaked flags + default-instance config (repair no-op, file byte-identical), leaked flags + port persistence (config untouched), and `.paperclip`-layout config whose env doesn't attest worktree (e.g. a repo checkout) — no adoption.

Intentional worktree-mode instances (e.g. the repo checkout's `project-primary-runtime` runtime-service pair, whose `.env` legitimately declares `PAPERCLIP_IN_WORKTREE=true`) keep working unchanged.

## Verification

- `server: vitest run src/__tests__/worktree-config.test.ts` → 12/12 pass (was 11 passed / 1 failed on this box before the hermetic scrub).
- `tsc --noEmit`: no errors in changed files (pre-existing plugin-sdk build errors in the throwaway worktree are unrelated).
- Live A/B on the dev box: unpatched suite run re-contaminated `instances/default/.env` (since restored); patched run left it clean.


cryppadotta:
Superseded by #9414: /prepare-pr found that this branch name included an internal ticket ID and the commit lacked the required Paperclip co-author trailer. The replacement preserves the same fix on current master, uses the full PR template, links related PR #3071, and is undergoing Greptile/CI review.

PR #9440, opened 2026-07-12
feat(ui): sidebar agent run dots — running/queued/errored

adacovsk:
## Why

The sidebar showed a single **"N live"** badge that counted every `queued` + `running` heartbeat run for an agent. Because it collapsed those two states, "7 live" read as "7 running" — which is exactly the confusion that prompted this: with `maxConcurrent = 1`, a Coordinator that keeps getting re-triggered while its one slot is busy stacks up `queued` runs, all counted as "live" behind a single executing run.

## What

Replace the one badge with up to three color-coded dots, each carrying its count:

| Dot | Meaning | Source |
|---|---|---|
| 🟢 green (pulse) | running | `status = running` (already in payload) |
| 🟡 amber | queued, waiting for a slot | `status = queued` (already in payload) |
| 🔴 red | latest run errored recently | new opt-in server path |

- **Green/amber are free** — the `live-runs` payload already carries `status`; the sidebar just buckets by it.
- **Red reuses the inbox badge's recent-failure semantics.** A new opt-in `?includeRecentErrors=1` on `GET /companies/:id/live-runs` appends each agent's *latest* run when it's `failed`/`timed_out` within `FAILED_HEARTBEAT_FRESHNESS_MS` (the constant is now exported from `sidebar-badges.ts` rather than duplicated, so "recently errored" means the same thing in both surfaces). A newer queued/running run supersedes an older failure, so an actively-retrying agent isn't flagged red.
- **No leakage into other consumers.** The param defaults off, so the other five `liveRunsForCompany` callers (dashboard, issue rail, etc.) still only see live runs. The sidebar query uses a distinct `[...liveRuns, "withErrors"]` key — prefix-invalidated by `LiveUpdatesProvider`, so it stays fresh without polluting the shared cache entry.

## Verification

- `pnpm typecheck` green in both `server` and `ui`.
- New `server/src/__tests__/agent-live-runs-routes.test.ts` (2 tests, passing): default path returns only queued/running and never calls `selectDistinctOn`; `includeRecentErrors=1` appends only the *fresh* failure and excludes both the stale timeout and the running-agent's superseding latest run.


adacovsk:
Mis-targeted: bare gh opened this against upstream paperclipai/paperclip, but this work integrates into the adacovsk/paperclip fork (diverged codebase). Closing; recreating on the fork.

PR #9448, opened 2026-07-12
Enforce atomic global dispatch gate for Claude-local execution

shanimosco47-pixel:
## Thinking Path

> - The `claude_local` adapter mediates every real Claude Code CLI launch — heartbeat runs, board chat, login, and environment probes can all try to spawn `claude` independently
> - Nothing previously stopped two of these surfaces from launching a real Claude process at the same time, and a confirmed provider-quota hit on one surface did not durably block the others from retrying and hitting the same limit again
> - This needed a single, durable, fail-closed gate rather than a per-surface in-memory lock, since Paperclip can restart mid-run and any in-memory state would be lost right when it matters most
> - This pull request adds one Postgres-backed ownership row per scope (`claude_local/default`), wraps the registered adapter's `execute`/`testEnvironment`, and mediates the two direct launch surfaces (board chat's spawned process, and the Claude login route) through the same primitive, with a shared `settleDispatchGateResult` helper so a confirmed quota result is always persisted atomically with the ownership release — even when our own request-level timeout is what ended the process
> - The benefit is that a real quota/session-limit hit on any one surface now durably blocks every other Claude-launch surface until the block clears or an operator resumes, instead of each surface silently retrying into the same limit

No existing public GitHub issue tracks this specific gap; describing it inline per the bug report template.

**What happened**
Nothing serialized concurrent `claude` CLI launches across surfaces (heartbeat/adapter execute, board chat, login, environment-test hello-probe), and a confirmed provider-quota/session-limit result on one surface did not durably block the others — board chat's spawned-process `close` handler and the Claude-login route both released their gate to idle unconditionally, even when the underlying result was a confirmed quota hit.

**Expected behavior**
A confirmed quota/session-limit result on any Claude-local launch surface should durably persist a block (surviving a Paperclip restart) that blocks every other surface until it clears or an operator resumes — and no two launches should ever run concurrently against the single local Claude Code install.

**Steps to reproduce**
1. Trigger a board-chat message or Claude login while a heartbeat run (or another board-chat/login request) is actively executing `claude` — before this change, both could spawn concurrently.
2. Force a confirmed provider-quota/session-limit response from `claude` on the board-chat or login surface, then observe that the in-memory concurrency counter/gate reset to idle immediately, allowing an immediate retry into the same limit.

`4a40c0cb13fd0e8cd1213b71599c4d7275b51582` (upstream `master` — confirmed unchanged as of this PR's final push).

**Deployment mode**
`local_trusted` (single-operator local instance — the only mode board chat and Claude-local login operate in).

## What Changed

- Added a durable `dispatch_gate_state` table (one row per scope, `claude_local/default` today) with `ownershipState: idle | active | unknown`, plus quota-block fields (`blockedUntil`, `operatorResumeRequired`, `blockReason`).
- Added `server/src/services/dispatch-gate.ts`: `acquireDispatchGate` (atomic claim via `SELECT ... FOR UPDATE`), `releaseDispatchGate`, `markDispatchGateUnknown` (fail-closed on ambiguous outcomes, never auto-cleaned), `recordDispatchGateQuotaBlock`, `resumeDispatchGate`, and `withDispatchGate`/`settleDispatchGateResult` helpers that classify a settled result into `quota | idle | unknown` and apply the matching transition atomically.
- Wrapped the registered `claude_local` adapter's `execute` in `server/src/adapters/registry.ts` with the gate, reusing the adapter's own structured `errorFamily === "provider_quota"` classification — no second quota detector.
- Narrowed `testEnvironment` gating to only the single real inference-producing "hello" probe (via an injectable `runInferenceProbe` hook), leaving read-only checks (`--help`, auth status, quota-window reads) outside the gate.
- Mediated the two direct launch surfaces the registry wrapper doesn't cover: board chat's spawned `claude` process (`server/src/routes/board-chat.ts`) and the Claude login route (`server/src/routes/agents.ts`), both now settling through `settleDispatchGateResult` using the same narrow `isClaudeProviderQuotaError`/`extractClaudeRetryNotBefore` classifier the adapter path uses.
- Fixed a precedence bug caught in review: board chat's classifier checked its own 120s watchdog timeout *before* checking for confirmed quota evidence, so a process that produced a real quota/session-limit result but was cut off by our own timeout got released to idle instead of persisting the block. Quota evidence now always takes precedence over a timeout classification.
- Added a static architecture guard (`server/src/__tests__/dispatch-gate-architecture-guard.test.ts`) that enumerates every `spawn/execFile("claude")` and `runClaudeLogin(` occurrence in `server/src` via `matchAll`, asserting an exact per-file approved count and that every occurrence is preceded by `acquireDispatchGate(`.
- Added cross-process, restart-persistence, and real-orchestration proof tests (`dispatch-gate.test.ts`, `dispatch-gate-orchestration-proof.test.ts`) demonstrating the gate's atomicity does not depend on any single in-memory module binding, that a quota block survives a full connection/process disposal, and that a normal heartbeat dispatch, a stranded-issue continuation requeue, and a stranded-issue assignment-recovery requeue all reach the real registered adapter wrapper and never invoke the underlying Claude launch callback while blocked.
- **Bounded cleanup commit** (`e1f99aa9`): trimmed narrative comments/JSDoc across the feature down to one concise line each (retaining the safety-critical invariant each documents — ownership committed before spawn, unknown-is-fail-closed, owner-scoped settlement, quota-precedence-over-timeout, strong-quota-signals-only, read-only-probes-outside-gate) and inlined a single-use helper (`dispatchGateBlockedMessage`) into its sole caller with identical logic. **No control flow, type signature, or runtime behavior changed** — comment/doc trimming and one behavior-equivalent inline only.
- **Follow-up fix commit** (`b55eb819`): a fresh Greptile review on the cleanup commit caught that the same timeout-vs-quota precedence bug already fixed in board chat was still present in the Claude-login path — `classifyClaudeLoginQuota` in `server/src/routes/agents.ts` checked `timedOut` before the quota classifier, so a login attempt that timed out but produced confirmed session-limit text released the gate to idle instead of persisting the block. Reordered so confirmed quota evidence is always checked first, matching board chat's fix exactly. No new classifier, no scope change — reuses the same `isClaudeProviderQuotaError`/`extractClaudeRetryNotBefore` calls that were already there.
- **Operator-recovery commit** (`f69df629`): `resumeDispatchGate` was fully implemented and tested from the start of this PR but had zero production callers — a confirmed quota block had no operator-reachable recovery path. Added a narrow instance-admin-only route, `POST /api/instance/settings/experimental/dispatch-gate/claude-local/resume-quota`, that resumes an early quota block. The idle precondition is enforced **atomically inside the update's own `WHERE` clause** (`ownershipState = 'idle'`), not via a separate read-then-write, so it can never clear quota fields while ownership is `active` or `unknown`; it never touches `ownershipState`, `ownerKind`, or `ownerId`; the scope is hardcoded to `CLAUDE_LOCAL_DEFAULT_SCOPE` and cannot be redirected by request body. A successful resume is followed by an activity-log write via the existing `logActivity` mechanism — **this is not part of the same database transaction as the resume update**: if logging fails after the resume update has already committed, the quota block is already cleared regardless, and since one call fans out a log entry per company, that fan-out can itself complete only partially. Deliberately **out of scope**: reconciling `ownershipState = "unknown"` back to `idle`. The schema stores no process identity, so no in-app or database check can prove a `claude` process is no longer running — only a human directly inspecting the host can. No API, CLI, or automatic path exists for this; `doc/DATABASE.md` documents the exact manual `psql` procedure instead (stop Paperclip, independently verify no `claude` process is running, then a single `UPDATE ... WHERE ownership_state = 'unknown'` that touches only ownership fields, never quota fields).

## Verification

**Typecheck (clean):**
```
npx tsc --noEmit -p server/tsconfig.json                             → clean
npx tsc --noEmit -p packages/adapters/claude-local/tsconfig.json     → clean
npx tsc --noEmit -p packages/adapter-utils/tsconfig.json             → clean
```

**Production diff size (the reason for the cleanup commit):**
The cumulative handwritten production diff (base `4a40c0cb` → head) was **676 changed lines** after the two P1 quota-mediation fixes and the timeout-precedence correction — above this PR's working ceiling of 650. A read-only scope audit classified every production block (durable state, registry/board-chat/login/hello-probe mediation, shared settlement, incidental refactoring, comments) and found the two P1 fixes and hello-probe narrowing to be load-bearing, non-duplicative, and not safely reducible without reopening the findings they fix. The only safe lever was comment/JSDoc trimming plus inlining one single-use helper — a dedicated cleanup commit (`e1f99aa9bffd4b3c72decfad806f13e9b1d50400`, **6 files changed, +27/−92**) brought the total to **611 changed lines**, under the 650 ceiling, with zero logic changes.

Whitespace-insensitive evidence for `test.ts` (the file with the largest raw diff, 167 lines): `git diff --ignore-all-space` shows only **21 genuinely changed lines** (14 insertions / 7 deletions) — **~87% of that file's raw diff is pure reindentation** (identical tokens at a different column, forced by TypeScript null-narrowing across a pre-existing shared classification chain), not new logic.

**Test suites (as of the follow-up fix commit `b55eb819`):**
```
npx tsc --noEmit -p server/tsconfig.json                                  → clean
npx vitest run src/__tests__/dispatch-gate.test.ts                         → 8 passed
  (includes 2 new: cross-process-instance atomicity proof; restart-persistence proof)
npx vitest run src/__tests__/dispatch-gate-architecture-guard.test.ts      → 2 passed
npx vitest run src/__tests__/board-chat-route-feature-flag.test.ts        → 5 passed
npx vitest run src/__tests__/dispatch-gate-launch-surface-quota.test.ts   → 20 passed
  (includes 2 new: login timeout+confirmed-quota → quota, blocking every
  other surface; login timeout with no quota evidence → idle, unchanged)
npx vitest run src/__tests__/dispatch-gate-orchestration-proof.test.ts    → 3 test assertions passed
```
The follow-up fix touches only `server/src/routes/agents.ts`'s route-local classifier and its own test file — it does not touch the shared `dispatch-gate.ts` helpers, so the orchestration-proof suite above reflects the prior run rather than a rerun specific to this commit.
The orchestration-proof suite proves a normal heartbeat dispatch, a stranded-issue continuation requeue, and a stranded-issue assignment-recovery requeue all reach the real, unmodified registered `claude_local` adapter wrapper and never invoke the underlying launch callback while the gate is held (only the innermost package `execute` export is a spy; all orchestration code is real production code).

**Operator-recovery commit (`f69df629`) test results:**
```
npx tsc --noEmit -p server/tsconfig.json                                  → clean
npx vitest run src/__tests__/dispatch-gate.test.ts                        → 13 passed
  (includes 5 new: idle+blocked clears only quota fields and preserves null
  owner identity; refuses on active without changing anything; refuses on
  unknown without changing anything; reports not_found without creating a
  row; enforces the idle precondition atomically under a concurrent acquire)
npx vitest run src/__tests__/instance-settings-routes.test.ts             → 26 passed
  (includes 6 new: instance-admin resumes an idle block; non-admin board
  actor rejected without calling the service; agent actor rejected without
  calling the service; scope cannot be redirected via request body; active/
  unknown returns 409 without logging activity; missing row returns 404)
npx vitest run src/__tests__/dispatch-gate-architecture-guard.test.ts     → 2 passed
npx vitest run src/__tests__/dispatch-gate-launch-surface-quota.test.ts  → 20 passed
```
This commit touches only `server/src/services/dispatch-gate.ts` (the `resumeDispatchGate` signature/body) and `server/src/routes/instance-settings.ts` (the new route) — it does not touch board chat, login, the adapter registry, or the environment probe, so those suites are unaffected and were not rerun for this specific commit.

**Cumulative production scope after this commit:** the handwritten production diff (base `4a40c0cb` → `f69df629`, excluding schema/migrations/tests) increased from 611 to **649 changed lines**, still under the approved 650 ceiling.

**OpenAPI contract fix (`ef5c0df1`):** the resume-quota route was mounted in Express but had no OpenAPI registration, failing `src/__tests__/openapi-routes.test.ts > covers the mounted server routes exactly` (`missingInSpec`). Added the missing `registerCurrentRoute` entry in `server/src/routes/openapi.ts` (+13 lines), reusing the file's existing generic response helpers (`r.notFound`, `r.conflict`) rather than custom schemas, and relying on the same automatic board-actor security classification (`/api/instance/` prefix) already applied to the comparable `issue-graph-liveness-auto-recovery` routes — no new security-scheme code, no elevation to `instance_admin`-tagged classification beyond what the sibling routes already document. Verified directly by building the spec and inspecting the generated operation: `security` is `[BoardSessionAuth, BoardApiKeyAuth]`, `x-paperclip-authorization` is `{ actor: "board" }`, response codes are exactly `200/401/403/404/409`, and no request body is declared — all matching the real route's runtime behavior exactly.
```
npx tsc --noEmit -p server/tsconfig.json                      → clean
npx vitest run src/__tests__/openapi-routes.test.ts            → 3 passed (previously 1 failed)
```
Cumulative production diff (base `4a40c0cb` → `ef5c0df1`, excluding schema/migrations/tests): **662 changed lines**, within the boss-approved 670-line exception ceiling for this contract-only fix. No route, service, or auth behavior changed — this commit touches only the OpenAPI document.

Local replication of CI's `Verify serialized server suites (4/4)` (`pnpm test:run:serialized -- --shard-index 3 --shard-count 4`) was attempted three times and blocked each time by a Windows-specific `spawnSync pnpm ENOENT` inside the script's nested process spawn (an environment/tooling limitation, not a code issue — the same class of Windows-only friction already documented elsewhere in this PR for embedded-Postgres teardown and direct-vitest timing). The specific test that CI flagged (`openapi-routes.test.ts`) was verified directly and passes; the generated spec was independently inspected and matches runtime exactly. Fresh CI (Linux runners, where this spawn issue does not occur) is the authoritative full-suite check for this commit.

**Transparent disclosure — not every run of every suite was clean, and the distinction matters:**
- In one run, the orchestration-proof suite's 3 test assertions **passed**, but its `afterAll` teardown threw a Windows `EPERM` error deleting the embedded-Postgres temp directory. This is a Windows file-lock timing issue in test teardown, not a test failure — confirmed by the assertions themselves passing before the error.
- `packages/adapters/claude-local`'s `test.probe.test.ts`, when run directly outside its official CI pipeline, shows a `Test timed out in 5000ms` failure pattern (6/6). To rule out a regression from the cleanup, I restored `test.ts` to the **exact unmodified base** (`git show 4a40c0cb:...`) and reran the identical command: the **same 6/6 timeout pattern reproduced against completely unmodified code**, confirming this is pre-existing environment-level flakiness in direct-`vitest` execution (this package's own test script only runs reliably through the repo's official `run-vitest-stable.mjs` pipeline), not something introduced by this PR.

No intended runtime behavior changed by the cleanup commit. No deployment, restart, or unattended execution occurred at any point in producing or validating this PR.

`pnpm install --frozen-lockfile` was run earlier in this branch's history with no lockfile changes.

## Risks

- A row stuck in `unknown` (ambiguous termination) requires explicit operator resume — there is no automatic cleanup by design, since guessing idle on an ambiguous outcome is exactly the failure mode this PR closes.
- The gate serializes *all* Claude-local launches against one installation-global slot — this is intentional (one local Claude Code install, one usable slot at a time) but means an unrelated board-chat session can now block a heartbeat run's Claude launch (and vice versa) where previously they raced independently.
- Quota classification is intentionally narrow (reuses the adapter's existing text/structured classifier) — a provider message shape that classifier doesn't recognize will not open a block, by design, to avoid false positives from generic errors.
- One new migration (`dispatch_gate_state`) adds a small, single-row-per-scope table; no existing table's shape or data is touched.

## Model Used

Claude Sonnet 5 (`claude-sonnet-5`), extended reasoning/thinking mode, agentic tool use (file edits, Bash, `gh` CLI) via Claude Code. All code, tests, and this PR description were produced with Claude Code assistance; a human reviewed and authorized each phase of implementation before it proceeded.

## Checklist

I searched the GitHub PR list (open and recently closed) for similar or duplicate PRs before opening this one: queried `dispatch gate`, `claude quota`, and `concurrent claude launch` against `paperclipai/paperclip`. Found no duplicate — the closest related work is the already-closed PR #7677 ("Enforce global heartbeat concurrency cap on every run-start path"), which addresses a narrower, heartbeat-only concurrency cap via a different mechanism, not a durable cross-surface Postgres-backed gate.

- [x] I searched for similar/duplicate open or closed PRs and confirmed this is not a duplicate

### DIFF
--- a/doc/DATABASE.md
+++ b/doc/DATABASE.md
@@ -242,4 +242,57 @@ pnpm paperclipai secrets migrate-inline-env --company-id <company-id> --apply
 pnpm secrets:migrate-inline-env --apply
 ```
 
+## Claude-local dispatch gate recovery
+
+`dispatch_gate_state` holds one durable row per launch scope (today only
+`claude_local/default`) mediating every real `claude` CLI launch. Its
+`ownership_state` is `idle | active | unknown`; a confirmed provider-quota hit
+additionally sets `blocked_until` / `operator_resume_required` / `block_reason`
+on an otherwise-idle row. There are two distinct, non-overlapping recovery
+operations — never conflate them.
+
+**Quota resume** (clearing a still-active quota block early) is available to
+an instance administrator via:
+
+```
+POST /api/instance/settings/experimental/dispatch-gate/claude-local/resume-quota
+```
+
+It only clears `blocked_until`, `operator_resume_required`, and
+`block_reason`, and only when `ownership_state` is already `idle` — the check
+is enforced inside one atomic database update, not a separate read. It never
+touches `ownership_state`, `owner_kind`, or `owner_id`, and it refuses (409)
+if ownership is `active` or `unknown`. Paperclip does not need to be stopped
+for this operation, since ownership is provably idle throughout any quota
+block (no launch can acquire the scope while a block is active). Every
+successful call is recorded in the activity log
+(`instance.settings.dispatch_gate_quota_resumed`).
+
+**Unknown-owner reconciliation** (clearing `ownership_state = 'unknown'`,
+the fail-closed state left behind by an ambiguous/crashed termination) has
+**no API, CLI, or automatic path** — by design. The schema stores no process
+ID, so no in-app or database check can prove a `claude` process is no longer
+running; only a human directly inspecting the host can establish that.
+Restarting Paperclip, an elapsed amount of time, or a missing PID are each
+explicitly insufficient evidence and must never be used to justify this
+action. To reconcile, an operator must:
+
+1. Stop Paperclip completely.
+2. Read the current row: `SELECT * FROM dispatch_gate_state WHERE scope_key = 'claude_local/default';`
+3. Independently check the host OS process list and confirm no `claude`
+   process belonging to this installation is running.
+4. Record who is performing the action, when, and why, outside the database
+   (there is no in-app audit trail for a direct SQL statement).
+5. Run, exactly:
+   ```sql
+   UPDATE dispatch_gate_state
+   SET ownership_state = 'idle', owner_kind = NULL, owner_id = NULL, updated_at = now()
+   WHERE scope_key = 'claude_local/default' AND ownership_state = 'unknown';
+   ```
+   This changes only ownership fields and never touches `blocked_until`,
+   `operator_resume_required`, or `block_reason`.
+6. Re-read the row to confirm `ownership_state = 'idle'` before restarting
+   Paperclip. Verifying the fix with a real launch is a separate, explicitly
+   authorized action, not part of this procedure.
+
 Hosted AWS provider notes live in [SECRETS-AWS-PROVIDER.md](./SECRETS-AWS-PROVIDER.md).
--- a/packages/adapter-utils/src/types.ts
+++ b/packages/adapter-utils/src/types.ts
@@ -249,6 +249,17 @@ export interface AdapterEnvironmentTestContext {
     bindHost?: string | null;
     allowedHostnames?: string[];
   };
+  /**
+   * Optional hook wrapping only the single real inference-producing probe
+   * call (e.g. Claude's "hello" probe), so a caller can mediate launch
+   * ownership around just that call while every other read-only check
+   * (cwd validation, command resolvability, `--help` flag detection, etc.)
+   * runs unmediated. Returning `null` means the caller declined to run the
+   * probe (e.g. blocked); the adapter must treat that as "probe not run"
+   * and report accordingly rather than crash. Defaults to identity when
+   * omitted (the adapter runs the probe directly, as before this hook existed).
+   */
+  runInferenceProbe?: <T>(run: () => Promise<T>) => Promise<T | null>;
 }
 
 /** Payload for the onHireApproved adapter lifecycle hook (e.g. join-request or hire_agent approval). */
--- a/packages/adapter-utils/src/types.ts
+++ b/packages/adapter-utils/src/types.ts
@@ -249,16 +249,7 @@ export interface AdapterEnvironmentTestContext {
     bindHost?: string | null;
     allowedHostnames?: string[];
   };
-  /**
-   * Optional hook wrapping only the single real inference-producing probe
-   * call (e.g. Claude's "hello" probe), so a caller can mediate launch
-   * ownership around just that call while every other read-only check
-   * (cwd validation, command resolvability, `--help` flag detection, etc.)
-   * runs unmediated. Returning `null` means the caller declined to run the
-   * probe (e.g. blocked); the adapter must treat that as "probe not run"
-   * and report accordingly rather than crash. Defaults to identity when
-   * omitted (the adapter runs the probe directly, as before this hook existed).
-   */
+  /** Wraps only the real inference-producing probe call so read-only checks stay outside the launch gate; `null` means the caller declined to run it. */
   runInferenceProbe?: <T>(run: () => Promise<T>) => Promise<T | null>;
 }
 
--- a/packages/adapters/claude-local/src/server/index.ts
+++ b/packages/adapters/claude-local/src/server/index.ts
@@ -11,6 +11,7 @@ export {
 export {
   parseClaudeStreamJson,
   describeClaudeFailure,
+  extractClaudeRetryNotBefore,
   isClaudeMaxTurnsResult,
   isClaudeProviderQuotaError,
   isClaudeRefusalResult,
--- a/packages/adapters/claude-local/src/server/test.probe.test.ts
+++ b/packages/adapters/claude-local/src/server/test.probe.test.ts
@@ -161,3 +161,54 @@ describe("claude sandbox hello probe diagnostics", () => {
     expect(failed?.detail).toBeUndefined();
   });
 });
+
+describe("claude testEnvironment runInferenceProbe mediation", () => {
+  it("wraps only the hello probe call, not the other read-only checks", async () => {
+    probeResult.value = { exitCode: 0, stdout: initLine, stderr: "" };
+    let runInferenceProbeCalls = 0;
+    const runInferenceProbe = <T,>(run: () => Promise<T>): Promise<T | null> => {
+      runInferenceProbeCalls += 1;
+      return run();
+    };
+
+    await testEnvironment({
+      companyId: "company-1",
+      adapterType: "claude_local",
+      config: { engine: "cli", command: "claude" },
+      executionTarget: sandboxTarget,
+      environmentName: "Daytona",
+      runInferenceProbe,
+    });
+
+    // cwd check, command-resolvable check, and sandbox install check all ran
+    // via the mocked execution-target helpers above but never through
+    // runInferenceProbe — only the hello probe itself does.
+    expect(runInferenceProbeCalls).toBe(1);
+    expect(ensureAdapterExecutionTargetCommandResolvable).toHaveBeenCalled();
+    expect(runAdapterExecutionTargetProcess).toHaveBeenCalledTimes(1);
+  });
+
+  it("reports a blocked check (not a crash) when runInferenceProbe declines to run the probe", async () => {
+    let runInferenceProbeCalls = 0;
+    const runInferenceProbe = async <T,>(_run: () => Promise<T>): Promise<T | null> => {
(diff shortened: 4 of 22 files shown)


shanimosco47-pixel:
Closing and reopening to force a fresh commitperclip template-gate run against the current PR body (the prior run predates the body update; body edits don't retrigger the pull_request_target workflow).

Issue #8691, opened 2026-06-27
Wake/reopen loop on `blocked` issue with future `monitor_next_check_at` — no guard against `blocked → todo` PATCH or auto-checkout


PAT-Main:
## Summary

When an issue is `blocked` and has a future `monitor_next_check_at` (a scheduled "check back at T+N" sleep), Paperclip lets clients trivially break the sleep:

- The PATCH endpoint accepts `status: blocked → todo` without consulting `monitor_next_check_at`.
- That transition emits a `issue_status_changed` wake.
- On the next heartbeat, the auto-checkout flips the issue back into `in_progress` even though the checkpoint is days away.

In production this manifested as 4 reopens on the same monitor-only issue within 1 hour, each one consuming an agent heartbeat that did nothing but re-mark `blocked`.

## Version

`@paperclipai/server` v2026.609.0 (also `paperclipai` CLI v2026.609.0).

## Reproduction

1. Create an issue assigned to an agent, status `blocked`, `monitor_next_check_at` set to a future timestamp (e.g., `now + 48h`).
2. PATCH `/api/issues/{id}` with `{ "status": "todo" }` from any client that resolves to `local-board` (loopback CLI in `local_trusted`, or board UI).
3. Observe in `activity_log`: `issue.updated` with `details.status = todo, _previous.status = blocked`.
4. Observe in `heartbeat_runs`: a new run with `context_snapshot.wakeReason = "issue_status_changed"` fires.
5. Observe `issues.status` flips `blocked → in_progress` despite `monitor_next_check_at > now`.
6. Agent's heartbeat re-PATCHes back to `blocked` (a no-op cycle, but a wasted heartbeat).
7. If the client repeats the PATCH, the loop repeats indefinitely.

## Trace through the code

Files paths relative to `@paperclipai/server/dist/`:

1. **`routes/issues.js`** ~l.4190: `statusChangedFromBlockedToTodo = existing.status === "blocked" && issue.status === "todo" && (req.body.status !== undefined || reopened)` — no guard on `monitor_next_check_at`.
2. **`routes/issues.js`** ~l.4540: emits the wakeup with `reason: "issue_status_changed"`.
[...]

Issue #7846, opened 2026-06-09
Issue monitors never scheduled: monitor_scheduled_by set but monitor_next_check_at stays NULL → attention sweep blind (v2026.529.0)


carrgust:
## Summary
On a production install (`v2026.529.0`), **issue monitor coverage is 0%**: `issues.monitor_next_check_at` is NULL for **every** active issue (0/278 fleet-wide), so the attention/liveness sweep can never identify an overdue issue — it is vacuously "0 overdue" because nothing is ever scheduled to be checked.

## Observed
- `monitor_next_check_at IS NULL` on 100% of active `work_mode='standard'` issues (verified 0/119 active, 0/278 fleet-wide).
- `monitor_scheduled_by` is populated on a subset of issues, but the paired `monitor_next_check_at` is never written — so the scheduler appears to record *who* scheduled monitoring without recording *when the next check is due*.
- Net effect: the blocked/stale-issue attention contract is blind. On this install 81 blocked issues (some with all-terminal blockers) and 26 stale `in_review` issues accreted invisibly because no overdue-monitor sweep can fire.

## Expected
When a monitor is scheduled for an issue, `monitor_next_check_at` should be set to a future timestamp so the sweep can detect overdue issues and re-surface them.

## Impact
Silent loss of the per-issue liveness/attention sweep on standard-mode issues — the queue can degrade with no automated signal.

## Env
- Paperclip `v2026.529.0` (latest release at time of report)
- Postgres backend; single-company install
- Filed by the ResumoCast/RCV operator audit (external read-only auditor).

schmelli:
## Heartbeat Review — CRITICAL (Geargraph Quality)

**100% of issues have NULL monitor_next_check_at** — the attention/liveness sweep is completely blind. This is a Geargraph data integrity issue: the scheduler records *who* scheduled monitoring but never *when* the next check is due.

This blocks the blocked/stale-issue attention contract entirely. Keeping at CRITICAL.

schmelli:
**🟡 Heartbeat Triage (Sam): HIGH — health monitor issue**
- **Type:** Bug (issue monitors never scheduled — monitor_next_check_at stays NULL)
- **Priority:** HIGH
- **Why:** Attention sweep is blind — monitors registered but never fire. Geargraph quality issue.
- **Action:** Needs assignment.

schmelli:
**🔄 Heartbeat #002 Triage (Sam):**
[...]

PR #9449, opened 2026-07-12
fix(issues): allow scheduling execution-policy monitors on blocked issues

auroracapital:
## Thinking Path

> - Issues support an execution-policy "monitor": a scheduled future recheck (`nextCheckAt`) so an agent can come back to a blocked or waiting item once a condition should have cleared
> - `applyIssueExecutionPolicyTransition()` gates whether a monitor may be scheduled/retained via `issueAllowsMonitor(status, assigneeAgentId, assigneeUserId)`
> - That gate only allowed `status` to be `in_progress` or `in_review` — `blocked` was treated as an invalid state for a monitor
> - Blocked is exactly the status where a "check back later" monitor is most useful (e.g. "recheck once this calendar-gated dependency clears"), so this exclusion was backwards for a common real-world case
> - Worse, the exclusion applies even when the PATCH request never touches `executionPolicy` at all: any status change to `blocked` silently clears a previously-scheduled monitor as a side effect, with an HTTP 200 and no error, because the invalid-reason branch in `applyMonitorTransition()` only throws when the caller *explicitly* tried to set the monitor in that same request
> - An agent (or the heartbeat/disposition engine reading `monitorNextCheckAt`) has no signal that its recheck schedule was dropped — it just silently stops getting woken
> - The minimal, targeted fix is to extend `issueAllowsMonitor()` to also accept `blocked` for agent-owned issues, so a monitor can be explicitly scheduled while blocked, and is preserved (not silently cleared) across a `blocked -> blocked` PATCH that reasserts status without also changing `executionPolicy`

Related to #8691 (blocked issue with a future `monitor_next_check_at` is not respected by a subsequent `blocked -> todo` PATCH / auto-checkout) — that issue assumes a blocked issue *can* hold a future monitor and finds a different bypass further down the wake path; this PR fixes a separate, earlier problem where the monitor could never be validly scheduled/retained on a blocked issue in the first place. Possibly a contributing cause of the fleet-wide symptom in #7846 (`monitor_scheduled_by` set but `monitor_next_check_at` stays null) for any issue that was blocked at the time a monitor was set or reasserted, though that issue may also be explained by other mechanisms (e.g. the schema-stripping bug fixed in #8196) and is not claimed to be fully resolved by this change alone.

No public issue exists describing this exact `issueAllowsMonitor` mechanism, so per the template's Path (b), the bug is described in-PR:

### Bug report

**What happened**
`applyIssueExecutionPolicyTransition()` rejects (or silently clears) an execution-policy monitor whenever the issue's status is `blocked`, even for an agent-owned issue. A PATCH that explicitly sets `executionPolicy.monitor` on a blocked issue throws a 422 (`MONITOR_INVALID_MESSAGE`). A PATCH that changes status to `blocked` for any other reason, without touching `executionPolicy`, silently strips an already-scheduled monitor (`monitorNextCheckAt` set to null, `executionState.monitor.clearReason` set to `invalid_status`) and returns HTTP 200 with no indication to the caller.

**Expected behavior**
An agent-owned issue in `blocked` status should be able to have a monitor explicitly scheduled, and an existing scheduled monitor should survive a `blocked -> blocked` status reassertion that doesn't touch `executionPolicy`.

**Steps to reproduce**
1. Assign an issue to an agent, set status to `in_progress`, schedule a monitor via `executionPolicy.monitor.nextCheckAt`.
2. PATCH the issue's status to `blocked` (e.g. because of an external calendar-gated dependency), without including `executionPolicy` in the body.
3. Observe the response is 200, but `monitorNextCheckAt` is now `null` and `executionState.monitor.clearReason` is `invalid_status` — the recheck schedule is gone with no error surfaced.

## What Changed

- `server/src/services/issue-execution-policy.ts`: `issueAllowsMonitor()` now also permits `status === "blocked"` (in addition to `in_progress`/`in_review`) for agent-owned issues (non-null `assigneeAgentId`, null `assigneeUserId`). Updated `MONITOR_INVALID_MESSAGE` to reflect the new allowed set.
- `server/src/__tests__/issue-execution-policy.test.ts`:
  - Updated the existing "rejects explicitly scheduling a monitor on an invalid issue state" test to use `status: "todo"` (still genuinely invalid) instead of `"blocked"`, since `blocked` is now a valid monitor state.
  - Added a test confirming an explicit monitor schedule succeeds on a blocked, agent-owned issue.
  - Added a test confirming a scheduled monitor is preserved (not cleared) when a blocked issue's status is reasserted as `blocked` without an `executionPolicy` change in the same PATCH.

## Verification

```
# from server/
npx vitest run src/__tests__/issue-execution-policy.test.ts
# 52 passed (including the 2 new + 1 updated tests)

npx tsc --noEmit
# 126 pre-existing errors, all in unrelated plugin-host files caused by a
# missing @paperclipai/plugin-sdk build artifact; identical count on a clean
# origin/master checkout, and none reference issue-execution-policy.ts.
```

`src/__tests__/issue-execution-policy-routes.test.ts` has one pre-existing failing test ("rejects an agent-authored in_review transition without a review path", timeout) reproduced identically against a clean, unmodified `origin/master` checkout — unrelated to this change, not touched by this PR.

## Risks

Low. This only widens an existing allow-list predicate to include one additional status for agent-owned issues; the `assigneeUserId` must-be-null and `assigneeAgentId` must-be-set requirements are unchanged, so board-user-owned issues are unaffected. No schema/migration change. No behavior change for `todo`/`done`/`cancelled`/other statuses, which remain invalid monitor states.

## Model Used

Claude Sonnet 5 (`claude-sonnet-5`), via Claude Code, with tool use for reading, editing, and running tests/typecheck.

## Checklist

- [ ] All Paperclip CI gates are green — pending CI
- [ ] Greptile is 5/5 with no open P2s, recommendations, or follow-ups — pending review

### DIFF
--- a/server/src/__tests__/issue-execution-policy.test.ts
+++ b/server/src/__tests__/issue-execution-policy.test.ts
@@ -1570,7 +1570,7 @@ describe("issue execution policy transitions", () => {
       expect(() =>
         applyIssueExecutionPolicyTransition({
           issue: {
-            status: "blocked",
+            status: "todo",
             assigneeAgentId: coderAgentId,
             assigneeUserId: null,
             executionPolicy: null,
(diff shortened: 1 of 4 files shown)


auroracapital:
Good catch — fixed in 5e0e5ee. `issueAllowsMonitor()` accepting `blocked` was only half the fix: the due-monitor dispatch path (the on-demand trigger's status check, `tickDueIssueMonitors`, and both queries' claim conditions) still filtered to `in_progress`/`in_review` only, so a monitor scheduled on a blocked issue would persist but never actually fire.

Widened all four gates in that dispatch path to also accept `blocked`, and added a scheduler test (`issue-monitor-scheduler.test.ts`) that seeds a blocked issue with a due monitor and asserts it dispatches — enqueues a run, clears the one-shot schedule, and requests an agent wakeup — the same way an in_progress one does.

Full verification re-run:
- `issue-execution-policy.test.ts`: 52/52 passed
- `issue-monitor-scheduler.test.ts`: 7/7 passed (was 6, +1 new)
- `issue-execution-policy-routes.test.ts`: 9/9 passed
- Full `heartbeat*` suite: 433/433 passed
- `tsc --noEmit`: 126 pre-existing errors, unchanged (none in touched files)

auroracapital:
Rebased onto current `master` (`2568bdec`) and force-pushed (`5e0e5eec` → `c76d901a`).

Conflict resolution kept master's newer provider-quota monitor scheduler test and re-applied the blocked-status monitor dispatch coverage:
- `issueAllowsMonitor()` still allows `blocked` agent-owned issues
- due-monitor dispatch paths now include `blocked` again after the rebase
- scheduler test `triggers due issue monitors on blocked issues` retained alongside the master quota-participant test

CI re-running on the rebased tip. Maintainer merge of this PR remains the remaining SFB-106 unblock.

auroracapital:
## Ready for maintainer merge (CI green)

Rebased tip `c76d901a` is **MERGEABLE** and all required checks passed (typecheck, server/workspace suites, e2e shards, verify, policy, security).

Greptile inline P1 ("Blocked Monitors Never Dispatch") is already fixed on this tip:
- `issueAllowsMonitor()` accepts `blocked`
- due-monitor claim + `tickDueIssueMonitors` both include `blocked` in status gates

No write access for `auroracapital` to merge into `paperclipai/paperclip`. Please merge when ready.

Tracks SFB-106 / SFB-58 / SFB-87 on the Sam Feldt board.

PR #9502, opened 2026-07-13
test(server): fix race in bounded-retry resumeFailure assertion (flaky test)

stubbi:
## What

`heartbeat-process-recovery.test.ts` > "schedules bounded retries for failed accepted interaction continuation wakes" is flaky: it reads `issueThreadInteractions.result` once, right after waiting for the retry-run row — but the recovery flow writes the `resumeFailure` marker **last**, after the retry run is committed and the system comment is inserted. Under CI load the one-shot read races the detached writer and sees `{ outcome: 'accepted', version: 1 }` without `resumeFailure`:

```
AssertionError: expected { outcome: 'accepted', version: 1 } to match object { version: 1, …(2) }
- "resumeFailure": { "attempt": 1, "errorCode": "adapter_failed", "maxAttempts": 3 }
```

We've had this single test fail on several unrelated PRs (introduced with cc81eefb6).

## Fix

Poll for the marker with the file's existing `waitForValue` helper instead of reading once — the same pattern the rest of the file uses for late writes.

### DIFF
--- a/server/src/__tests__/heartbeat-process-recovery.test.ts
+++ b/server/src/__tests__/heartbeat-process-recovery.test.ts
@@ -1976,12 +1976,19 @@ describeEmbeddedPostgres("heartbeat orphaned process recovery", () => {
       body: "Agent failed to resume after approval: `adapter_failed` — retrying (attempt 1/3)",
     });
 
-    const interaction = await db
-      .select({ result: issueThreadInteractions.result })
-      .from(issueThreadInteractions)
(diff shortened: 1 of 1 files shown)


stubbi:
Closing as superseded: #10049 fixed the same race in this test by gating the retry barrier on the terminal interaction write (waitForValue on issueThreadInteractions.result reaching resumeFailure.status === "retrying"), which is more robust than the poll-for-marker approach here. Master already carries it.

PR #9512, opened 2026-07-13
feat(server): email delivery for company invites (SMTP transport)

stubbi:
## Summary
- Adds optional email delivery for company invites. Invite creation gains an optional `email` field (validated, nullable); when present and an SMTP transport is configured, the invitee receives an email with the invite link. With no configuration, behavior is byte-for-byte unchanged — invites stay copyable links.
- A small transport hook (`invite-email.ts`) keeps delivery pluggable: the SMTP transport registers at startup only when `PAPERCLIP_SMTP_FROM` plus `PAPERCLIP_SMTP_URL` or `PAPERCLIP_SMTP_HOST` are set. Invalid configuration is logged and disables delivery — it never prevents server startup.
- Delivery is strictly best-effort and fire-and-forget: the hook swallows transport errors and the route does not await the SMTP round-trip, so a slow or failing mail server never affects invite creation latency or success. The copyable invite URL is always the fallback.
- UI: the company invites page gets an optional "Invitee email" input ("Add an email and we send the invite for you. Leave it blank to copy a link instead.").

## Configuration
New "Email (invite delivery)" section in `docs/deploy/environment-variables.md` (`PAPERCLIP_SMTP_URL` or `PAPERCLIP_SMTP_HOST`/`PORT`/`SECURE`/`USER`/`PASSWORD`, plus `PAPERCLIP_SMTP_FROM`).

## Implementation notes
- HTML output escapes all interpolated values (company name, role, URL); the `z.string().email()` validation also excludes CRLF, so no header-injection surface via the recipient.
- `PAPERCLIP_SMTP_URL` must use an `smtp://`, `smtps://`, or `direct:` scheme — anything else fails fast with a descriptive config error (nodemailer would otherwise treat the string as a plugin and crash cryptically).
- `@types/nodemailer` latest is v8 while `nodemailer` is v9 (no v9 types package exists yet); one localized cast bridges the gap, and the send path is verified at runtime (below).

## Testing
- 18 unit tests: env resolution (activation rule, URL precedence, port/secure defaults, invalid-scheme error), mail rendering (subject variants, role line, HTML escaping, no-recipient no-op), hook semantics (error swallowing), startup registration (registers only when configured).
- Runtime smoke against maildev: invite with email → one message delivered with correct recipient, subject, and invite link; invite without email → no message; startup log line present; server boots normally with no SMTP config.


stubbi:
This branch has drifted too far from current master (`d293dd3d1`) for a meaningful rebase — the surrounding code has moved substantially since it was cut. Rather than force through a large conflict resolution that would need re-review anyway, I'll leave it as-is and re-cut it against current master when a maintainer signals interest in the direction. Happy to refresh it on request.

PR #9523, opened 2026-07-13
fix(adapter-utils): stage large SSH payloads off os.tmpdir()

jodylarsen:
## Problem

Four SSH sync sites in `packages/adapter-utils/src/ssh.ts` stage multi-GB payloads under `os.tmpdir()`:

| Site | Payload |
|---|---|
| `syncDirectoryFromSsh` | whole-workspace sync-back tar |
| `restoreWorkspaceFromSshExecution` | whole-workspace sync-back tar |
| `importGitWorkspaceToSsh` | git bundle, full history |
| `exportGitWorkspaceFromSsh` | git bundle, full history |

`os.tmpdir()` is conventionally small, frequently RAM-backed (`tmpfs`), and often quota-capped. Staging a large workspace there can exhaust the quota — and the resulting `EDQUOT` then surfaces on an unrelated write. On our fleet it landed on the few-hundred-byte SSH key write in `withTempFile()`, which failed with `errno 122 (EDQUOT)` and took the whole run down. The key write was the *victim*, not the cause, which made this hard to diagnose.

Any git-backed workspace can reach the bug through the two bundle sites, not just large-workspace sync-back.

## Fix

Add `resolveStagingRoot(localDir)` and point all four staging sites at it. Precedence:

1. `PAPERCLIP_STAGING_DIR`, if set
2. `<parent-of-workspace>/.paperclip-staging` — disk-backed and on the same filesystem as the destination, so staging → workspace becomes a **rename instead of a cross-device copy**
3. `os.tmpdir()` — last-resort fallback

`withTempFile()` deliberately **stays** on `os.tmpdir()`: its payload is a few hundred bytes and must not depend on a workspace path.

## Test

Adds a regression test to the SSH env-lab fixture suite (`ssh-fixture.test.ts`). It runs a real sync-back through the package's own real `sshd` fixture, polls both candidate staging roots during the transfer, and asserts the tar lands beside the workspace and **not** in `os.tmpdir()` — with `TMPDIR` pointed at a private directory so the assertion can't be contaminated by concurrent tests.

Verification:

- New test **fails** against the pre-fix code (`expected true to be false` — it catches the tar in `os.tmpdir()`) and passes with the fix.
- Full `packages/adapter-utils` suite: **213 passed / 213**, including the git round-trip tests that exercise both bundle sites.
- `tsc --noEmit` clean.

### DIFF
--- a/packages/adapter-utils/src/ssh-fixture.test.ts
+++ b/packages/adapter-utils/src/ssh-fixture.test.ts
@@ -1,4 +1,5 @@
 import { execFile } from "node:child_process";
+import { existsSync, readdirSync } from "node:fs";
 import { mkdir, mkdtemp, readFile, rm, symlink, writeFile } from "node:fs/promises";
 import os from "node:os";
 import path from "node:path";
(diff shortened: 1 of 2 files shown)


jodylarsen:
Superseded by #9524 — same fix, but rebranched off an internal ticket id (CONTRIBUTING.md → no internal/instance-local references in branch names or source) and with the required PR template sections filled in. Closing this one to keep review in a single place.

PR #9524, opened 2026-07-13
fix(adapter-utils): stage large SSH payloads off os.tmpdir()

jodylarsen:
## Thinking Path

> - Agents can execute in a remote workspace over SSH; `@paperclipai/adapter-utils` moves the workspace to and from that host (git bundles on the way out, a tar of the run directory on the way back)
> - Every one of those staging sites writes its intermediate payload — a whole-workspace tar or a full-history git bundle, routinely multi-GB — under `os.tmpdir()`
> - `os.tmpdir()` is the wrong place for a large payload: on many hosts it is a small, RAM-backed tmpfs, and on ours it also carries a per-user quota that is *smaller than the tmpfs itself*, so the process can never reach `ENOSPC` — it takes `EDQUOT` first, while `df` still reports free space
> - The damage lands somewhere else entirely. Exhausting the quota with a staged tar makes the **next** write fail, and the next write is a few-hundred-byte SSH key file in `withTempFile()` during lease acquisition. The run dies with a bare `errno 122` and no path in the error, pointing at the innocent party
> - This pull request stages large payloads beside the workspace they belong to instead, and leaves the small key write on `os.tmpdir()` where it belongs
> - The benefit is that a large workspace no longer poisons an unrelated write, and staging → workspace becomes a rename on one filesystem instead of a cross-device copy

No existing public issue — describing it inline as a bug report.

**What happened:** Remote SSH runs fail during lease acquisition with `Error: ENOSPC`-adjacent `errno 122` (`EDQUOT`) on a ~400-byte SSH key write in `withTempFile()`. The error names the key file, so the key write looks like the culprit.

**Actual cause:** the key write is the victim. `syncDirectoryFromSsh` (and three sibling staging sites) had already staged a multi-GB workspace tar under `os.tmpdir()`. Where `os.tmpdir()` is a quota-capped tmpfs, that stage consumes the user's entire quota; the key write is simply the first write afterwards, and it takes the `EDQUOT`.

**Why it is hard to see:** `df` reports the *filesystem*, not the *quota*. With a quota smaller than the tmpfs, `df` shows free space at the exact moment writes are being refused, so disk-space checks all come back clean and correct — and irrelevant.

**Repro:** mount a tmpfs with `usrquota` and a hard limit below its size, point `TMPDIR` at it, and run an SSH-backed workspace sync whose workspace exceeds the quota. The sync succeeds in staging until the quota is hit; the failure then surfaces on the next unrelated temp write.

**Related:** #9018 (reaps stale SSH sync-back temp dirs and pre-checks disk headroom) targets the same staging code. It is complementary, not duplicate: #9018 makes `os.tmpdir()` staging tidier and checks *free space*; this PR moves large payloads out of `os.tmpdir()` altogether, which is also the only thing that helps when the limit is a *quota* rather than free space.

## What Changed

- Add `resolveStagingRoot(localDir)` in `packages/adapter-utils/src/ssh.ts`, resolving in order: `PAPERCLIP_STAGING_DIR` (explicit override) → `<parent-of-workspace>/.paperclip-staging` (disk-backed, same filesystem as the destination) → `os.tmpdir()` (last-resort fallback, preserving today's behaviour when neither is usable).
- Point all four large-payload staging sites at it: `syncDirectoryFromSsh`, `restoreWorkspaceFromSshExecution`, `importGitWorkspaceToSsh`, `exportGitWorkspaceFromSsh`.
- Leave `withTempFile()` on `os.tmpdir()` **deliberately**: its payload is a few hundred bytes, and it must not depend on a workspace path (it is used before a workspace is resolved). It was the victim of the exhaustion, not the cause; moving it would hide the bug rather than fix it.
- Add a regression test to the SSH env-lab fixture suite that runs a real sync-back through a real `sshd` and asserts the staging tar lands beside the workspace and never in `os.tmpdir()`.

## Verification

```
# regression test (drives a real sshd via the env-lab fixture)
pnpm vitest run packages/adapter-utils/src/ssh-fixture.test.ts -t "stages the sync-back tar"
#   Test Files  1 passed (1)   Tests  1 passed | 15 skipped (16)

# typecheck
pnpm --filter @paperclipai/adapter-utils typecheck   # clean
```

The test is a genuine regression test, not a tautology — I verified it fails on the pre-fix code by reverting only `ssh.ts` and re-running:

```
FAIL  expect(stagedIn.tmpdir).toBe(false)
  - Expected: false
  + Received: true      # pre-fix, the tar is observed staged in os.tmpdir()
```

It polls both candidate roots *during* the transfer, because the staging directory is removed as soon as the sync completes and can only be observed in flight. `TMPDIR` is pointed at a private directory for the duration so the assertion cannot be confused by staging from a concurrent test.

## Risks

Low, with one behavioural change worth naming.

- **New directory on disk.** Staging now creates `<parent-of-workspace>/.paperclip-staging`. It is created lazily, and each stage still `mkdtemp`s a unique child inside it that is cleaned up as before.
- **Fallback preserves old behaviour.** If neither the override nor the beside-workspace path is writable, `resolveStagingRoot` returns `os.tmpdir()` — so the worst case is exactly today's behaviour, never a new hard failure.
- **Filesystem assumption softened, not added.** Staging beside the workspace means the staging → workspace move is usually a rename rather than a cross-device copy. That is strictly better than staging on a different device, which is what `os.tmpdir()` frequently was.
- **Not changed:** `withTempFile()` still uses `os.tmpdir()`, so small-key behaviour is untouched.

## Model Used

Claude Opus 4.8 (`claude-opus-4-8`), Anthropic, via Claude Code / Claude Agent SDK — extended thinking, agentic tool use (filesystem, shell, test execution). The diagnosis was confirmed against the kernel with `quotactl_fd(2)` on the affected host, and the fix was verified by running the new test against both the pre-fix and post-fix source.

## Checklist

- [x] I have searched GitHub for duplicate or related PRs and linked them above (#9018 — related, complementary; supersedes my own #9523, closed in favour of this branch)
- [x] My branch name describes the change (`fix/stage-large-ssh-payloads-off-tmpdir`) and contains no internal Paperclip ticket id or instance-derived details
- [ ] I have updated relevant documentation to reflect my changes — n/a, no user-facing surface changes; `PAPERCLIP_STAGING_DIR` is an internal escape hatch and is documented at the definition site
- [ ] All Paperclip CI gates are green — pending on this branch
- [ ] Greptile is 5/5 with no open P2s, recommendations, or follow-ups — pending

### DIFF
--- a/packages/adapter-utils/src/ssh-fixture.test.ts
+++ b/packages/adapter-utils/src/ssh-fixture.test.ts
@@ -1,4 +1,5 @@
 import { execFile } from "node:child_process";
+import { existsSync, readdirSync } from "node:fs";
 import { mkdir, mkdtemp, readFile, rm, symlink, writeFile } from "node:fs/promises";
 import os from "node:os";
 import path from "node:path";
@@ -316,6 +317,81 @@ describe("ssh env-lab fixture", () => {
     // The restored files round-tripped through the byte-counting transport.
     await expect(readFile(path.join(restoreDir, "blob-0.bin"))).resolves.toEqual(
       Buffer.alloc(256 * 1024, 1),
+    );
+  }, SSH_FIXTURE_TEST_TIMEOUT_MS);
+
+  // Regression: sync-back staged its whole-workspace tar in os.tmpdir(). Where that is
+  // a quota-capped RAM tmpfs, a large workspace exhausts the quota, and the failure then
+  // surfaces on an unrelated write (the few-hundred-byte SSH key) as EDQUOT. Staging must
+  // land beside the destination workspace instead.
+  it("stages the sync-back tar beside the workspace, not in os.tmpdir()", async () => {
+    const rootDir = await mkdtemp(path.join(os.tmpdir(), "paperclip-ssh-fixture-"));
+    cleanupDirs.push(rootDir);
+    const statePath = path.join(rootDir, "state.json");
+    const localDir = path.join(rootDir, "sync-source");
+    // Restoring into runRoot/workspace means staging must appear in runRoot/.
+    const runRoot = path.join(rootDir, "run");
+    const restoreDir = path.join(runRoot, "workspace");
+    // os.tmpdir() is resolved from TMPDIR on every call, so pointing it at a private
+    // directory isolates this assertion from staging done by any concurrent test.
+    const privateTmp = path.join(rootDir, "private-tmp");
+
+    await mkdir(localDir, { recursive: true });
+    await mkdir(restoreDir, { recursive: true });
+    await mkdir(privateTmp, { recursive: true });
+    for (let index = 0; index < 8; index += 1) {
+      await writeFile(path.join(localDir, `blob-${index}.bin`), Buffer.alloc(1024 * 1024, index + 1));
+    }
+
+    const started = await startSshEnvLabFixtureOrSkip(statePath, "SSH staging directory test");
+    if (!started) return;
+    const config = await buildSshEnvLabFixtureConfig(started);
+    const spec = { ...config, remoteCwd: started.workspaceDir } as const;
+    const remoteDir = path.posix.join(started.workspaceDir, "staging-source");
+
+    await syncDirectoryToSsh({ spec, localDir, remoteDir });
+
+    // Poll both candidate roots during the transfer: the staging dir is removed once
+    // the sync completes, so it can only be observed while it is in flight.
+    const stagedIn = { tmpdir: false, besideWorkspace: false };
+    const scan = () => {
+      const roots: Array<[keyof typeof stagedIn, string]> = [
+        ["tmpdir", privateTmp],
+        ["besideWorkspace", path.join(runRoot, ".paperclip-staging")],
+      ];
+      for (const [key, dir] of roots) {
+        // The staging root may not exist yet on early ticks.
+        let entries: string[] = [];
+        try {
+          entries = readdirSync(dir);
+        } catch {
+          continue;
+        }
+        if (entries.some((entry) => entry.startsWith("paperclip-ssh-sync-back-"))) {
+          stagedIn[key] = true;
+        }
+      }
+    };
+    const poller = setInterval(scan, 10);
+
+    const previousTmpdir = process.env.TMPDIR;
+    process.env.TMPDIR = privateTmp;
+    try {
+      await syncDirectoryFromSsh({ spec, remoteDir, localDir: restoreDir });
+    } finally {
+      clearInterval(poller);
+      if (previousTmpdir === undefined) delete process.env.TMPDIR;
+      else process.env.TMPDIR = previousTmpdir;
+    }
+
+    expect(stagedIn.tmpdir).toBe(false);
+    expect(stagedIn.besideWorkspace).toBe(true);
+    // The staging root itself is created (not the mkdtemp child, which is cleaned up),
+    // so this holds even if every poll tick missed the transfer window.
+    expect(existsSync(path.join(runRoot, ".paperclip-staging"))).toBe(true);
+    // Staging off tmpdir must not change what lands in the workspace.
+    await expect(readFile(path.join(restoreDir, "blob-0.bin"))).resolves.toEqual(
+      Buffer.alloc(1024 * 1024, 1),
     );
   }, SSH_FIXTURE_TEST_TIMEOUT_MS);
 
--- a/packages/adapter-utils/src/ssh-fixture.test.ts
+++ b/packages/adapter-utils/src/ssh-fixture.test.ts
@@ -351,25 +351,18 @@ describe("ssh env-lab fixture", () => {
 
     await syncDirectoryToSsh({ spec, localDir, remoteDir });
 
-    // Poll both candidate roots during the transfer: the staging dir is removed once
-    // the sync completes, so it can only be observed while it is in flight.
-    const stagedIn = { tmpdir: false, besideWorkspace: false };
+    // The mkdtemp staging dir is removed once the sync completes, so it can only be
+    // caught in flight. Poll tmpdir for one: any sighting there is the bug.
+    const stagedIn = { tmpdir: false };
     const scan = () => {
-      const roots: Array<[keyof typeof stagedIn, string]> = [
-        ["tmpdir", privateTmp],
-        ["besideWorkspace", path.join(runRoot, ".paperclip-staging")],
-      ];
-      for (const [key, dir] of roots) {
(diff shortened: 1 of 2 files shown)


jodylarsen:
Pushed `7e6b36a` addressing the two substantive review findings:

**1. Writable-root check (`ssh.ts`).** Confirmed the gap was real: `fs.mkdir(dir, {recursive: true})` resolves on a pre-existing directory with mode `0o500`, and the follow-on `fs.mkdtemp` under it then throws `EACCES` — with no fallback. `resolveStagingRoot` now probes `fs.access(candidate, W_OK)` after `mkdir`, so an unwritable candidate falls through to the next one instead of hard-failing the sync. A writable candidate is still preferred over `os.tmpdir()`.

**2. Timing-dependent test assertion (`ssh-fixture.test.ts`).** Agreed — a *positive* staging sighting is only observable if a poll tick lands inside the transfer window, so it should not be asserted on. The staging root itself outlives the sync (only the `mkdtemp` child is cleaned up), so `existsSync(<runRoot>/.paperclip-staging)` is now the load-bearing, deterministic assertion. The poller is kept but narrowed to `tmpdir` only, where a sighting at *any* tick is the bug and a negative is what we want to hold.

Internal ticket references were already scrubbed from the branch and diff when this PR superseded #9523 (per CONTRIBUTING.md → "No Internal Issue References"); the PR template sections are filled in above.

Verification: `tsc --noEmit` clean on `packages/adapter-utils`, and the regression test passes against the real SSH env-lab fixture (`vitest run packages/adapter-utils/src/ssh-fixture.test.ts -t "stages the sync-back tar"` → 1 passed).

jodylarsen:
Pushed `066a753` addressing the Greptile review note on the baseline-snapshot path in `restoreWorkspaceFromSshExecution`.

**Confirmed the finding was real.** That path creates a scratch `stagingDir` under the workspace's `.paperclip-staging` root, then passes it to `syncDirectoryFromSsh` as `localDir`. The inner call re-resolved its own staging root relative to that scratch dir (`dirname(stagingDir)/.paperclip-staging`), so a second, empty `.paperclip-staging/.paperclip-staging` was created and left behind on every baseline restore — accumulating cruft on a code path whose whole purpose is disk hygiene.

**Fix:** `syncDirectoryFromSsh` now takes an optional `stagingAnchorDir`. When the caller's `localDir` is itself scratch, it anchors the staging root to the real destination workspace, so the inner sync reuses the same `.paperclip-staging` root (sibling scratch dir) instead of nesting a new one. `withTempFile()` and the four large-payload sites are unchanged.

**Verification:** added a regression test (`does not leave a nested .paperclip-staging behind after a baseline restore`) that drives the baseline-merge branch through the real `sshd` fixture and asserts (a) the remote edit round-trips into the workspace, (b) exactly one `.paperclip-staging` root exists beside the workspace, and (c) no nested `.paperclip-staging/.paperclip-staging`. Negative control: reverting only `ssh.ts` makes the test fail (`expected true to be false` on the nested-dir check). Full fixture suite: 17 passed. `tsc --noEmit` clean.

jodylarsen:
Pushed `c7b062a` addressing the remaining Greptile P2 on silent fallback for an unusable `PAPERCLIP_STAGING_DIR`.

**What changed.** When the configured override fails `mkdir`/`access(W_OK)`, we now emit:

```
[paperclip] PAPERCLIP_STAGING_DIR=<path> is not usable (<reason>); falling back
```

before selecting the beside-workspace candidate (or ultimately `os.tmpdir()`). Behaviour is unchanged — only the failure is now observable.

**Status.** CI was green on `066a753` (Greptile 5/5, all required checks pass, mergeable=`MERGEABLE`, mergeStateStatus=`CLEAN`). This push is a one-file diagnostic; requesting maintainer review so it can land with the next release cut. We cannot merge from the fork (pull-only).

jodylarsen:
Re-triggered the failed `General tests (server (1/3))` lane.

Failure is unrelated to this PR: `heartbeat-retry-scheduling.test.ts` teardown hits an FK race (`delete from heartbeat_runs` while `heartbeat_run_events` still references the row). Our change is a single `console.warn` in `packages/adapter-utils/src/ssh.ts`. Greptile remains 5/5.

jodylarsen:
## Ready to merge — Lucentworks LUC-738 Track B

All required CI checks are green on this PR (Build, Typecheck, Canary Dry Run, general + serialized server suites, e2e, Greptile, security scans).

Local verification just re-confirmed on `fix/stage-large-ssh-payloads-off-tmpdir`:

- `pnpm --filter @paperclipai/adapter-utils typecheck` — clean
- `pnpm vitest run packages/adapter-utils/src/ssh-fixture.test.ts -t "stages the sync-back tar"` — 1 passed

This lands the durable fix for staging multi-GB SSH payloads off `os.tmpdir()` (four sites via `resolveStagingRoot`, `withTempFile` left on tmpdir). Once merged to `master`, the normal canary/stable release pipeline will publish a version our fleet can pin.

We do not have merge rights on `paperclipai/paperclip` — please merge when ready. Thank you.

jodylarsen:
@cryppadotta @devinfoley @nickyleach — bump: [#9524](https://github.com/paperclipai/paperclip/pull/9524) is CI-green and mergeable (`MERGEABLE` / `CLEAN`). Ready when you are.

Landing this unblocks an operational EDQUOT failure class for SSH workspace sync on hosts where `os.tmpdir()` is quota-capped. Normal canary publish on `master` is sufficient; no special release window requested.

jodylarsen:
## Merge request — fleet EDQUOT (errno 122)

Lucentworks is blocked on merging this PR so `@paperclipai/adapter-utils` can ship a canary/latest containing `resolveStagingRoot` (large SSH staging off `os.tmpdir()`).

**Why it matters:** SSH-backed sync-back / git-bundle paths currently stage multi-GB payloads under quota-capped `os.tmpdir()`; the failure then surfaces as bare `errno 122` (`EDQUOT`) on an unrelated small key write. Fleet pin work cannot start until a published version includes this fix.

**PR state (rechecked this heartbeat):**
- `MERGEABLE`, required checks green (Build / General tests / Typecheck+Release Registry / e2e / verify / Greptile / Socket / Snyk)
- Head: `3c3569314003bd72990193a191f08e526c6c9c1e` on `fix/stage-large-ssh-payloads-off-tmpdir`

**Ask:** @cryppadotta @devinfoley @nickyleach — please review + merge when convenient so release automation can publish. Lucentworks (`jodylarsen`) has pull-only rights (`MergePullRequest` denied) and no npm publish token, so we cannot land or cut the upstream package ourselves.

Thank you.

PR #9488, opened 2026-07-13
claude-local: first-class CLAUDE_CODE_OAUTH_TOKEN subscription auth for headless/remote runs

stubbi:
## Thinking Path

> - Paperclip is the open source app people use to manage AI agents for work
> - The claude-local adapter runs the Claude Code CLI, which supports Anthropic API-key auth and Claude Pro/Max subscription auth
> - Subscription auth currently works headlessly only when the machine already has an on-disk login (`~/.claude/.credentials.json`) — remote/sandbox execution targets get a fresh HOME and therefore no subscription auth
> - The Claude Code CLI itself already accepts a long-lived token minted by `claude setup-token` via the `CLAUDE_CODE_OAUTH_TOKEN` env var, and the adapter's env plumbing already delivers `config.env` into runs (including remote targets)
> - The adapter's connection test doesn't know this: with no API key and no local login it steers the user to a local `claude login` that cannot exist on a remote target
> - This pull request makes the token option first-class in the adapter's auth checks and docs — no new transport, just recognition of an auth mode the CLI already supports
> - The benefit is that subscription-based users can run Claude Code on remote/sandbox targets without converting to API-key billing

## Linked Issues or Issue Description

No existing public issue; described inline below. Related but not duplicates: #4606 and #4068 strip host-managed Claude OAuth material from *spawned local* subprocess env for isolation — this PR is about *explicitly configured* adapter env delivered to runs, so the two are complementary (an operator-set `config.env` token is intentional, not host leakage). Also related: #2043 (Claude auth diagnostic probe for Railway Docker — same headless-auth pain point).

**Subsystem affected:** claude-local adapter — connection test (`server/test.ts`), ACP lane check (`server/acp.ts`), adapter configuration docs

**Problem or motivation:** Claude Pro/Max subscription auth works headlessly only with an on-disk login. Remote/sandbox execution targets get a fresh HOME and no subscription auth, and the adapter's connection test recognizes only `ANTHROPIC_API_KEY` or a local login — its guidance (`run claude login`) is impossible to follow on a remote target.

**Proposed solution:** Recognize `CLAUDE_CODE_OAUTH_TOKEN` (the token `claude setup-token` mints, which the CLI natively honors) in the adapter's auth checks, mirror the recognition in the ACP lane, and document the option including its real limitation.

[...]

PR #9545, opened 2026-07-14
feat(ui): guided per-adapter credential setup in onboarding and agent settings (BYOK connect)

stubbi:
## What

Every deployment without host-level provider env — cloud multi-tenant or plain self-hosted — needs the user to hand the product a model credential before their first agent works. Today that means leaving the product: figure out which env var the adapter wants, create a company secret, come back, and bind it as a `secret_ref` env row. Nothing in the wizard or agent settings names the env var or the setup command, and the wizard's environment check reports **Passed** for a credential-less Claude agent (the probe only runs `claude --version`), so the first run fails with the CLI's raw "Not logged in · Please run /login".

This PR makes credential setup a one-step, in-product action:

- **`AdapterCredentialSetup` descriptors** (new vendor-neutral types in `@paperclipai/adapter-utils`): each built-in local adapter exports, from its client-safe `/ui` entry, the credential options it accepts — env key, kind (`api_key` / `subscription_token`), label, hint, first-party setup command/URL, placeholder. Claude declares both `ANTHROPIC_API_KEY` and the `CLAUDE_CODE_OAUTH_TOKEN` minted by `claude setup-token` (hint covers the token-only quota-reporting limitation and the "extra usage" billing caveat); codex/gemini/opencode/pi/cursor declare their API keys. Exposed on the UI adapter registry as an optional `UIAdapterModule.credentialSetup` — external/plugin adapters simply have none.
- **`AdapterCredentialConnect`**: one card that renders the options (segmented), the copyable setup command, a password field — and on Connect creates the company secret and hands back the id for a `{type:"secret_ref"}` env binding in one action. Retry with a `-2` name suffix only on a 409 name conflict; other errors surface immediately (no blind retry that could duplicate a live credential after a client-side timeout).
- **Agent settings**: the card renders above the environment-variables editor whenever the adapter has a descriptor and none of its env keys are bound; binding appends a normal secret-ref row to the same editor state (single persistence path) and refreshes the secrets list.
- **Onboarding wizard step 4**: same card under the model picker; bindings persist in the wizard draft and merge into both the environment-test and hire payloads — filtered to the *selected* adapter's declared env keys (switching from Claude to Gemini never leaks an Anthropic binding into the Gemini agent) and merged on top of the config's own env (an explicit unset marker survives). No gating: agent creation works exactly as before with no binding.
- **Honest env check**: warn-level checks now render (in warn styling) even when the probe passes, instead of hiding behind the green "Passed" pill.

## What this deliberately does not do

- **No embedded provider OAuth.** Anthropic's authentication policy (Feb 2026) restricts consumer OAuth to Claude Code/Claude.ai with no third-party client registration, and OpenAI's ChatGPT sign-in is Codex-only — so the guided hand-off of a first-party-minted token is the durable design. The descriptor is shaped so an OAuth connector kind could slot in later if that changes.
- No descriptors for external/plugin adapters (would need server `AdapterInfo` plumbing; the optional field degrades cleanly).
- No blocking of agent creation on missing credentials — host env and sandbox-registry provisioning remain first-class.

Note: the check-surfacing renders warn-level checks; the claude subscription-token *info* check comes from #9488 and composes with this PR without depending on it.

## Tests

Descriptor unit tests per package; component tests (create+bind, 409-scoped retry, non-409 no-retry, trim, Enter-submit, connected-state); AgentConfigForm render tests (gating, bound-detection incl. plain values, editor-row append, secrets-list refresh); wizard tests (card gating, draft persist + restore, merge contract incl. no-`env`-when-empty regression guard, adapter-switch filtering, unset-marker survival, single-"Passed"-pill warn rendering). Full ui suite: 2551 passing (2 pre-existing env-sensitive failures untouched by this branch: a rolling-date bucket test and a locale-formatted number assertion).

### DIFF
--- a/packages/adapter-utils/src/index.ts
+++ b/packages/adapter-utils/src/index.ts
@@ -36,6 +36,8 @@ export type {
   StdoutLineParser,
   CLIAdapterModule,
   CreateConfigValues,
+  AdapterCredentialOption,
+  AdapterCredentialSetup,
 } from "./types.js";
 export type {
   SessionCompactionPolicy,
--- a/packages/adapter-utils/src/types.ts
+++ b/packages/adapter-utils/src/types.ts
@@ -555,3 +555,41 @@ export interface CreateConfigValues {
   headersJson?: string;
   password?: string;
 }
+
+// ---------------------------------------------------------------------------
+// Adapter credential setup — declarative UI descriptors for guided credential configuration
+// ---------------------------------------------------------------------------
+
+/**
+ * A single credential option that an adapter accepts.
+ *
+ * Describes how to configure one environment variable (e.g., an API key or token),
+ * with hints and optional setup commands/URLs to help users obtain and configure it.
+ */
+export interface AdapterCredentialOption {
+  /** The environment variable name (e.g., "ANTHROPIC_API_KEY") */
+  envKey: string;
+  /** The kind of credential */
+  kind: "api_key" | "subscription_token";
+  /** Display label for the credential (e.g., "Anthropic API key") */
+  label: string;
+  /** Optional hint text explaining what the credential is for and how to use it */
+  hint?: string;
+  /** Optional command users can run to set up this credential (e.g., "claude setup-token") */
+  setupCommand?: string;
+  /** Optional URL where users can obtain/manage this credential (e.g., console URL) */
+  setupUrl?: string;
+  /** Optional placeholder text for the input field (e.g., "sk-ant-…") */
+  placeholder?: string;
+}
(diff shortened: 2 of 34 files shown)


stubbi:
This branch has drifted too far from current master (`d293dd3d1`) for a meaningful rebase — the surrounding code has moved substantially since it was cut. Rather than force through a large conflict resolution that would need re-review anyway, I'll leave it as-is and re-cut it against current master when a maintainer signals interest in the direction. Happy to refresh it on request.

PR #9468, opened 2026-07-12
perf(ui): improve Decisions scrolling performance

cryppadotta:
## Thinking Path

> - Paperclip is the open source app people use to manage AI agents for work
> - Operators use the Decisions page to review an uncapped attention feed across active, snoozed, and dismissed items
> - Large feeds mounted every row eagerly, and routine interactions re-rendered the full queue
> - That made initial paint and scrolling progressively slower as decision history accumulated
> - This pull request bounds rendering, stabilizes row props, and lets off-screen rows skip layout and paint work
> - The benefit is a responsive Decisions page even for companies with large attention histories

### What happened?

Opening `/decisions` for a company with a large attention history eagerly mounted every visible-feed row. Expanding, selecting, dismissing, snoozing, or restoring an item could also re-render the entire queue.

### Expected behavior

The page should render a bounded initial window, progressively reveal more rows near the scroll boundary, and avoid re-rendering unaffected rows during interactions.

### Steps to reproduce

1. Populate a company with hundreds of attention items.
2. Open `/decisions`.
3. Scroll and interact with individual rows.
4. Observe increasing initial render, layout, paint, and interaction cost on the previous implementation.

### Paperclip version or commit

Reproduced on `master` before this PR.

### Deployment and installation

Local development, built from source. This is a core UI issue, not adapter- or database-specific.

### Additional context

Searched open public issues and PRs; no duplicate was found.

## What Changed

- Added a pure `planAttentionRenderRows` helper that allocates one render budget across active groups and open snoozed/dismissed curtains in document order.
- Render 50 rows initially and add 100 more when the Decisions page approaches the scroll boundary.
- Memoized `AttentionQueueRow`, stabilized parent callbacks and inbox dismissal actions, and passed row items through a shared expand callback.
- Added `content-visibility: auto` and intrinsic containment so accumulated off-screen rows avoid unnecessary layout and paint work.
- Added render-plan coverage and a regression test proving identical row props do not re-render after a parent update.

## Verification

- `pnpm -C ui typecheck`
- `pnpm -C ui exec vitest run src/lib/attention.test.ts src/components/AttentionQueueRow.test.tsx src/components/Sidebar.test.tsx src/pages/Inbox.test.tsx` — 96 tests passed
- `pnpm check:token-gates` — all gates clean

## Risks

- Low risk: the change is UI-only and does not alter API or database contracts.
[...]

PR #9599, opened 2026-07-15
feat(doctor): add planner stats, log, and shared_buffers health checks

Stack-Alerts:
## Summary

Extends `paperclipai doctor --repair` with four new health checks that customers on long-lived embedded-postgres deployments have been asking for, and that were called out in the perf follow-up tracking long-running planner-stats and log growth on the hot tables:

| Check | What it does | Auto-repair |
|---|---|---|
| **Planner statistics freshness** | Scans `pg_stat_user_tables` for the hot table set (`activity_log`, `agent_wakeup_requests`, `agents`, `companies`, `heartbeat_run_events`, `heartbeat_runs`, `issues`) and warns when the most recent `autoanalyze`/`analyze` is older than 7 days. | `ANALYZE` per stale table |
| **Activity log size** | Warns when `pg_total_relation_size('public.activity_log')` exceeds 1 GiB. | No (suggest archive + `VACUUM FULL` — needs a maintenance window) |
| **Server log size** | Warns when `<logDir>/server.log` exceeds 100 MiB. | Copy to `<logDir>/server.log.<ISO-8601>` then truncate in place. Uses `COPYFILE_EXCL` so existing rotated copies are not clobbered. |
| **PostgreSQL shared_buffers** *(embedded-postgres only)* | Reads the active and configured values from `pg_settings` / `pg_file_settings` and recommends 25 % of host RAM, clamped to `[128 MiB, 8 GiB]`. Surfaces a separate warn when the change is already in `pg_file_settings` but pending a server restart. | `ALTER SYSTEM SET shared_buffers = '<N>MB'` + `pg_reload_conf()` |

All three DB-touching checks share a new `openDoctorDb` helper that probes connectivity with a 1.5 s timeout and cleans up its probe timer in a `finally` block. When the embedded cluster has not been initialised yet the checks return a skipped-warn rather than a hard fail, so `paperclipai doctor` remains safe to run before the server has ever booted.

## How to test

```bash
pnpm --filter paperclipai test -- doctor-health-checks     # focused unit tests with injectable openDb
pnpm --filter paperclipai test -- doctor                  # end-to-end doctor test
pnpm --filter paperclipai build
```

On a machine with an already-bootstrapped embedded-postgres data dir:

```bash
paperclipai doctor --repair --yes
```

## Risk / rollback

- Purely additive CLI checks — no schema changes, no server-side changes, no public API changes.
- Worst case if `--repair` is misapplied to `shared_buffers`: a subsequent server restart reverts to the prior value; the previous setting is not lost (we only mutate `pg_file_settings` via `ALTER SYSTEM`).
- Server-log rotation copies then truncates in place — if the copy fails the original file is untouched.
- All three DB-touching checks fail open (skipped-warn) when the cluster is unreachable, so they cannot mask a real DB outage.

## Out of scope

- `VACUUM FULL` for `activity_log` (intentionally not auto-repaired — needs a maintenance window; see the `repairHint`).
- Planner-stats checks for non-embedded external PostgreSQL — the check now has an explicit `embedded-postgres` mode guard (mirroring `sharedBuffersCheck`) and returns `pass` with "External PostgreSQL manages planner statistics" for external clusters. We chose to skip rather than inspect externally because `ANALYZE` against an external production database is a write of a different surface than the embedded case, and the 7-day threshold may not be appropriate for all deployments. Re-tuning the threshold or extending the check to external PostgreSQL is a follow-up.

## Thinking Path

Started from the perf follow-up (#37767) which already tracks `activity_log` bloat and unbounded `server.log` growth for `paperclipai/Paperclip`. The narrower #37815 surfaces the same shape from a Paperclip-on-Paperclip standpoint. The shared_buffers warmup also surfaced through the embedded-postgres follow-up cluster (#39007) where cluster boot times regressed as host RAM grew — running with the default `128 MiB` against a 16 GiB host leaves 99 % of the working set uncached.

Doctor is the natural entry point because:

1. Operators already reach for `paperclipai doctor` when something feels off; bolting on auto-repair is muscle-memory compatible.
2. Each check has a unit test fixture with an injectable `openDb`, so the regression surface is bounded and stable. The shared_buffers parser was the only nontrivial bit — `pg_file_settings` echoes a bare number back without a unit suffix, so the parser must use the active row's native unit (8 kB blocks) when scaling the configured value. Fix lands in this PR with a unit test that exercises the `pending_restart=true` path with bare-number, `MiB`, and `GiB` configured values.
3. The non-DB checks (log size, planner stats) are pure read against `pg_stat_user_tables` — no migration, no schema, no breaking surface.

The four checks compose independently. The factory function `openDoctorDb` (new in `db-connect.ts`) is the only shared implementation, and it stays internally scoped.

## Model Used

Claude Opus 4.7 (Plan → Implement → Review). Review pass was a self-review against the diff for invariant violations (e.g., the parser bug above).

## Verification

- `pnpm --filter paperclipai test -- doctor-health-checks` — 8 / 8 pass after the most recent push. Test inventory:
  - `plannerStatsCheck` repairs stale hot-table statistics with `ANALYZE`.
  - `plannerStatsCheck` skips (warn, no repair) when PostgreSQL is unreachable.
  - `activityLogSizeCheck` warns above the size threshold and suggests `VACUUM FULL`.
  - `serverLogSizeCheck` copy-truncates an oversized `server.log` and writes a timestamped rotated copy.
  - `sharedBuffersCheck` configures embedded PostgreSQL `shared_buffers` to 25 % of host RAM.
  - `sharedBuffersCheck` skips shared_buffers inspection for externally-managed PostgreSQL.
  - `sharedBuffersCheck` parses `configured_setting` with the row's native unit when the change is pending restart (bare-number path — the regression case).
  - `sharedBuffersCheck` parses binary `configured_setting` units (`MiB`, `GiB`) on the pending-restart path.
- `pnpm --filter paperclipai build` — clean build, no type errors.
- `pnpm --filter paperclipai typecheck` — clean.
- Manual spot check: with `hostMemoryBytes = 8 GiB`, the recommended target is 2 GiB which matches the test fixture on the configuration branch (`ALTER SYSTEM SET shared_buffers = '2048MB'`).

## Linked issue

- BTCAAAAA-37848 (canonical source for the fix and the regression test)
- Parent: BTCAAAAA-37815 (Paperclip perf follow-up)

## Dedup

- [x] I searched open PRs for overlapping work (planner-stats, log rotation, shared_buffers) and confirmed this PR does not duplicate an in-flight change on paperclipai/paperclip. The upstream PR #9472 (mobile decision rows) and PR #9468 (Decisions scrolling) are unrelated UI work. No PR on `master` or in `#open-issues` covers the four checks this PR introduces.

## Review fixes (greptile P1)

Two P1 review comments landed on the initial PR head (9d71f11ab). Both are addressed in a follow-up commit on this branch:

- **Scope external databases** (`cli/src/checks/planner-stats-check.ts:45`) — `plannerStatsCheck` now returns `pass` with "External PostgreSQL manages planner statistics" when `config.database.mode !== "embedded-postgres"`, mirroring the guard already in `sharedBuffersCheck`. Without the guard, `paperclipai doctor --repair --yes` would run `ANALYZE` against an external production database. Covered by a new test `does not inspect planner statistics for externally managed PostgreSQL` (the analogue of `sharedBuffersCheck`'s extern-managed test).
- **Fix broken import** (`cli/src/checks/server-log-size-check.ts:5`) — import path corrected from `./path-resolver.js` to `../utils/path-resolver.js`. The `./path-resolver.js` shim exists for the four legacy check callers (`secrets-check`, `database-check`, `log-check`, `storage-check`) and is intentionally preserved; the new check imports the canonical resolver directly (matching `db-connect.ts`).

Test count is now 9 / 9 (one new test added for the mode guard).

### DIFF
--- a/cli/src/__tests__/doctor-health-checks.test.ts
+++ b/cli/src/__tests__/doctor-health-checks.test.ts
@@ -1,0 +1,210 @@
+import fs from "node:fs";
+import os from "node:os";
+import path from "node:path";
+import { afterEach, describe, expect, it, vi } from "vitest";
+import type { PaperclipConfig } from "../config/schema.js";
+import { activityLogSizeCheck } from "../checks/activity-log-size-check.js";
+import { plannerStatsCheck } from "../checks/planner-stats-check.js";
+import { serverLogSizeCheck } from "../checks/server-log-size-check.js";
+import { sharedBuffersCheck } from "../checks/shared-buffers-check.js";
+
+const tempRoots: string[] = [];
+
+function createConfig(overrides: Partial<PaperclipConfig> = {}): PaperclipConfig {
+  const root = fs.mkdtempSync(path.join(os.tmpdir(), "paperclip-doctor-health-"));
+  tempRoots.push(root);
+
+  return {
+    $meta: {
+      version: 1,
+      updatedAt: "2026-07-15T00:00:00.000Z",
+      source: "doctor",
+    },
+    database: {
+      mode: "embedded-postgres",
+      embeddedPostgresDataDir: path.join(root, "db"),
+      embeddedPostgresPort: 54329,
+      backup: {
+        enabled: true,
+        intervalMinutes: 60,
+        retentionDays: 7,
+        dir: path.join(root, "backups"),
+      },
+    },
+    logging: {
+      mode: "file",
+      logDir: path.join(root, "logs"),
+    },
+    server: {
+      deploymentMode: "local_trusted",
+      exposure: "private",
+      host: "127.0.0.1",
+      port: 3100,
+      allowedHostnames: [],
+      serveUi: true,
+    },
+    telemetry: {
+      enabled: true,
+    },
+    auth: {
+      baseUrlMode: "auto",
+      disableSignUp: false,
+    },
+    storage: {
+      provider: "local_disk",
+      localDisk: {
+        baseDir: path.join(root, "storage"),
+      },
+      s3: {
+        bucket: "paperclip",
+        region: "us-east-1",
+        prefix: "",
+        forcePathStyle: false,
+      },
+    },
+    secrets: {
+      provider: "local_encrypted",
+      strictMode: false,
+      localEncrypted: {
+        keyFilePath: path.join(root, "secrets", "master.key"),
+      },
+    },
+    ...overrides,
+  };
+}
+
+afterEach(() => {
+  vi.restoreAllMocks();
+  for (const root of tempRoots.splice(0)) {
+    fs.rmSync(root, { recursive: true, force: true });
+  }
+});
+
+describe("plannerStatsCheck", () => {
+  it("repairs stale hot-table statistics with ANALYZE", async () => {
+    const execute = vi
+      .fn()
+      .mockResolvedValueOnce([{ relname: "activity_log" }, { relname: "issues" }])
+      .mockResolvedValue([]);
+    const openDb = vi.fn(async () => ({ db: { execute } as never, connectionString: "postgres://test" }));
+
+    const result = await plannerStatsCheck(createConfig(), undefined, { openDb });
+
+    expect(result.status).toBe("warn");
+    expect(result.canRepair).toBe(true);
+    expect(result.message).toContain("activity_log");
+    expect(result.message).toContain("issues");
+
+    await result.repair?.();
+
+    expect(execute).toHaveBeenCalledTimes(3);
+    expect(execute.mock.calls[1]?.[0]).toBe('ANALYZE "activity_log"');
+    expect(execute.mock.calls[2]?.[0]).toBe('ANALYZE "issues"');
+  });
+
+  it("warns without repair when PostgreSQL is unavailable", async () => {
+    const openDb = vi.fn(async () => {
+      throw new Error("connection refused");
+    });
+
+    const result = await plannerStatsCheck(createConfig(), undefined, { openDb });
+
+    expect(result.status).toBe("warn");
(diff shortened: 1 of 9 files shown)


Stack-Alerts:
Both P1 review comments on the prior head (`9d71f11ab`) are addressed in `0476288e6`:

**1. Scope external databases** (`cli/src/checks/planner-stats-check.ts:45`)

`plannerStatsCheck` now mirrors the guard already in `sharedBuffersCheck` and returns `pass` with `"External PostgreSQL manages planner statistics"` when `config.database.mode !== "embedded-postgres"`. Without this guard, `paperclipai doctor --repair --yes` would have run `ANALYZE` against an externally managed production database.

The new test `does not inspect planner statistics for externally managed PostgreSQL` covers the guard — the analogue of the `sharedBuffersCheck` extern-managed test. Test count: 9 / 9 (`pnpm --filter paperclipai test -- doctor-health-checks`).

**2. Fix broken import** (`cli/src/checks/server-log-size-check.ts:5`)

Import path corrected from `./path-resolver.js` to `../utils/path-resolver.js` (the canonical resolver). The `./path-resolver.js` shim under `cli/src/checks/` is intentionally preserved for the four legacy check callers (`secrets-check`, `database-check`, `log-check`, `storage-check`) and is unchanged; the new check imports the canonical resolver directly, matching `db-connect.ts`.

PR "Out of scope" section now reflects the explicit mode guard (rather than leaving external-PostgreSQL planner-stats as a follow-up).

PR #9608, opened 2026-07-15
fix(heartbeat): reap scheduled_retry runs with NULL scheduledRetryAt

celestial-unit:
## Thinking Path

> - Paperclip is a control plane for AI agent companies — it dispatches runs, tracks their state, and recovers from failures
> - The heartbeat service owns run lifecycle: `reapOrphanedRuns()` clears stuck `running` rows after a crash, `promoteDueScheduledRetries()` advances `scheduled_retry` rows when their due-time arrives
> - `heartbeat_runs.scheduled_retry_at` is nullable in the schema; `promoteDueScheduledRetries` uses `lte(scheduledRetryAt, now)` which silently drops NULLs in SQL; `reapOrphanedRuns` only queries `status = 'running'`
> - A `scheduled_retry` row with `scheduledRetryAt IS NULL` is therefore invisible to both sweeps — it holds the issue `executionRunId` lock indefinitely and never reaches a terminal state
> - This PR extends `reapOrphanedRuns()` with a second sweep that finds and fails any `scheduled_retry` rows missing a due-time, releases the issue execution lock, and runs the same agent-status cleanup the existing orphan path uses
> - The benefit is that no run can be permanently stranded in a non-terminal status due to a missing `scheduledRetryAt`, and the issue execution lock is always eventually released

No existing issue. Describing inline per the bug template:

**What happened**

A `heartbeat_runs` row written with `status = 'scheduled_retry'` and `scheduled_retry_at = NULL` is never promoted (SQL `lte` drops NULLs) and never reaped (reaper only queries `status = 'running'`). It holds the issue `execution_run_id` lock forever.

**Expected behavior**

Any run that cannot be promoted should eventually be moved to a terminal status and release its issue lock.

**Steps to reproduce**

Insert a `heartbeat_runs` row directly with `status = 'scheduled_retry'` and `scheduled_retry_at = NULL`. Observe that neither `reapOrphanedRuns` nor `promoteDueScheduledRetries` ever touches it.

Reproducible on current `master`. The column has been nullable since the `scheduled_retry` status was introduced.

**Deployment mode**

All modes — the gap is in the in-process heartbeat scheduler, not infrastructure-specific.

## What Changed

- `server/src/services/heartbeat.ts` — after the existing `running`-only loop in `reapOrphanedRuns()`, added a second sweep: queries `status = 'scheduled_retry' AND scheduledRetryAt IS NULL`, marks each matching run `failed` with `errorCode: 'stuck_scheduled_retry'`, fails the wakeup request, appends a lifecycle run event, calls `releaseIssueExecutionAndPromote()` to unlock the issue, and calls `finalizeAgentStatus()` + `startNextQueuedRunForAgent()` to match the cleanup contract of the existing orphan path
- `server/src/__tests__/heartbeat-process-recovery.test.ts` — two new test cases:
  - "reaps a scheduled_retry run whose scheduledRetryAt is NULL (permanently stuck)" — seeds a `scheduled_retry` row with no due-time, calls `reapOrphanedRuns()`, asserts run is `failed`, wakeup is `failed`, issue checkout lock released, agent status updated
  - "does not reap a scheduled_retry run that has a valid future due-time" — asserts the happy-path `scheduled_retry` row is left untouched

## Verification

```sh
cd server
npx vitest run "heartbeat-process-recovery"
# 62 passed
```

Both new tests exercise the exact gap. The negative test confirms no regression on normal `scheduled_retry` rows.

## Risks

Low risk. The new sweep only matches rows where `scheduledRetryAt IS NULL` — a condition the application code never produces today (it always sets `schedule.dueAt` on insert). No existing row in a healthy instance will be affected. The only rows touched are ones that are already permanently stuck. All helpers called (`setRunStatus`, `setWakeupStatus`, `appendRunEvent`, `releaseIssueExecutionAndPromote`, `finalizeAgentStatus`, `startNextQueuedRunForAgent`) are used identically in the existing running-orphan loop directly above.

## Model Used

Kiro (Amazon) — Auto mode (model selected per task). Extended reasoning applied during root-cause analysis of the NULL-visibility gap. Tool use enabled throughout (file reads, grep, diagnostics, test execution).

## Checklist

- [x] My branch name describes the change (`fix/reaper-stuck-scheduled-retry-null-due-time`) and contains no internal Paperclip ticket id or instance-derived details

### DIFF
--- a/server/src/__tests__/heartbeat-process-recovery.test.ts
+++ b/server/src/__tests__/heartbeat-process-recovery.test.ts
@@ -3824,4 +3824,100 @@ describeEmbeddedPostgres("heartbeat orphaned process recovery", () => {
     const runs = await db.select().from(heartbeatRuns).where(eq(heartbeatRuns.id, runId));
     expect(runs).toHaveLength(1);
   });
+
+  // -----------------------------------------------------------------------
+  // Stuck scheduled_retry reaper
+  // -----------------------------------------------------------------------
+
+  it("reaps a scheduled_retry run whose scheduledRetryAt is NULL (permanently stuck)", async () => {
+    // Arrange: create a run directly in scheduled_retry with no due-time.
+    // This simulates a row that would be silently skipped by both
+    // promoteDueScheduledRetries (lte NULL fails) and the existing
+    // running-only reapOrphanedRuns sweep.
+    const { companyId, agentId, runId, wakeupRequestId, issueId } = await seedRunFixture({
+      runStatus: "running", // seed helper only supports running; we'll override below
+      includeIssue: true,
+    });
+
+    // Override to scheduled_retry with NULL scheduledRetryAt
+    const now = new Date();
+    await db
+      .update(heartbeatRuns)
+      .set({
+        status: "scheduled_retry",
+        scheduledRetryAt: null,
+        scheduledRetryAttempt: 1,
+        scheduledRetryReason: "transient_failure",
+        updatedAt: now,
+      })
+      .where(eq(heartbeatRuns.id, runId));
+
+    const heartbeat = heartbeatService(db);
+
+    // Act
+    const result = await heartbeat.reapOrphanedRuns();
+
+    // Assert: the stuck run was collected
+    expect(result.reaped).toBeGreaterThanOrEqual(1);
+    expect(result.runIds).toContain(runId);
+
+    // The row must now be terminal
+    const run = await db
+      .select()
+      .from(heartbeatRuns)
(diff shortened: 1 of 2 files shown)


celestial-unit:
The failing CI check (invite-rate-limit-route.test.ts) is unrelated to this PR — it times out at exactly 5000ms on "returns 429 once the per-IP threshold is exceeded", which is a pre-existing flaky test in the serialized server suite. The heartbeat-process-recovery suite passes cleanly (62/62 locally and in CI).

PR #9716, opened 2026-07-16
feat(agents): hired/created agents inherit the company credential

stubbi:
## Problem
When an agent hires (or a user creates) a new agent, the new agent's `adapter_config.env` is empty, so its runs fail `claude_auth_required` ("Not logged in · Please run /login") until a credential is connected for that specific agent. The company already holds the credential as a `company_secret` bound to an existing agent's env (e.g. the CEO's `CLAUDE_CODE_OAUTH_TOKEN`). New agents should inherit it.

## Change
Before persisting a hired/created agent, copy each `secret_ref` env binding from an existing **same-adapter** company agent (preferring the CEO) into the new agent's `adapter_config.env`:
- only `secret_ref` bindings are inherited (the shape the credential-connect flow produces);
- an env key the request set explicitly is never overridden;
- the donor binding's `version` / `projectionClass` / `projectionAllowlistKey` are preserved verbatim;
- no-op when no same-adapter donor has a credential (no behavior change / no regression).

The injected `secret_ref` flows through the existing `normalizeEnvConfig` → `assertSecretInCompany` → `syncAgentSecretBindings` path, so it is validated and wired exactly like a user-supplied binding — no changes to the runtime/heartbeat path.

Wired into `POST /companies/:id/agent-hires` and `POST /companies/:id/agents`. `mergeInheritedCredentialEnv` is extracted as a pure helper and unit-tested (6 cases).

### DIFF
--- a/server/src/__tests__/agent-credential-inheritance.test.ts
+++ b/server/src/__tests__/agent-credential-inheritance.test.ts
@@ -1,0 +1,66 @@
+import { describe, expect, it } from "vitest";
+import { mergeInheritedCredentialEnv } from "../routes/agents.js";
+
+describe("mergeInheritedCredentialEnv (hired agents inherit the company BYOK credential)", () => {
+  const donorCred = { type: "secret_ref", secretId: "sec-1", version: "latest" };
+
+  it("inherits a donor secret_ref the new agent does not have", () => {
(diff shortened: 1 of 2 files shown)


stubbi:
Closing as superseded by #9920, which carries the same company-credential inheritance helper extracted into a dedicated service module (services/agent-credential-inheritance.ts), extends it to built-in agent provisioning, and includes this PR's test coverage. Please review #9920 instead.

Issue #9445, opened 2026-07-12
Routine scheduler registration bug — cron-configured routine never dispatches (0 rows ever in routine_runs)


adarshhalan:
### Pre-submission checklist

- [x] I have searched existing open and closed issues and this is not a duplicate.
- [x] I am on the latest released version of Paperclip (or can reproduce on `master`).
- [x] I have confirmed the error originates in Paperclip itself — not in my agent adapter, API provider, or local configuration.

### What happened?

We have a routine with a correct cron expression and a valid `next_run_at` value in the
`routine_triggers` table, but `routine_runs` shows **zero executions ever** for this
routine's ID — confirmed via direct Postgres query across 3 separate sessions over ~5
days, most recently:

```sql
select r.title, count(rr.id) as run_count, max(rr.triggered_at) as last_run
from routines r
left join routine_runs rr on rr.routine_id = r.id
where r.id = '<routine_id>'
group by r.title;
-- Result: <routine title> | 0 | (null)
```

Other routines with essentially identical cron configuration (same class of routine,
similarly zero-fired for the first few checks) have since started firing on their own
between our checks, without any change on our side — so this looks like an intermittent
registration issue in the scheduler, not anything wrong with how the cron trigger is
configured. One specific routine is still stuck at zero after ~5 days while its siblings
recovered.

**Separately, and possibly related**: we also traced an agent that was failing with
`"Agent cannot start because its budget hard-stop is still exceeded"` for 2 consecutive
days despite confirmed spend under its cap the whole time (checked directly against
`cost_events`). This symptom has since stopped reproducing on its own (last 2 runs both
proceeded past that check) — flagging in case it's diagnostically useful, since it may
share a root cause with the scheduler registration issue above (both look like
stale/stuck internal state that eventually clears itself for some routines/agents but
not others).

**Environment**: self-hosted Paperclip instance, Docker deployment.

---

## Second, likely-related issue: sustained agent-fleet task timeouts (4+ days)

Roughly half of our agent fleet has been stuck in `error` status simultaneously for 4+
days. `activity_log` shows every failure as `environment.lease_released` with
`"failureReason": "Timed out after 1800s"` (some `3600s`), `"driver": "local"`,
`"leasePolicy": "ephemeral"` — the agent's local execution environment itself hangs for
the full timeout window, not a fast-failing error.

Ruled out on our side: disk space (well under capacity, not critical), the LLM
proxy/router chain (verified healthy with real completions both before and after a
[...]

Issue #5184, opened 2026-05-04
Bug: routine_triggers.next_run_at never auto-advances — scheduled routines never fire


fifatest-cmd:
## Severity

High — affects autonomous governance for any Paperclip company using scheduled routines.

## Environment

- Paperclip self-hosted on Hetzner
- `paperclipai` CLI from `/usr/lib/node_modules/paperclipai/dist/index.js`
- DB: PostgreSQL 14+ (Supabase pooler)

## Bug description

Routines (`routines` table with `routine_triggers` enabled) never fire automatically. `next_run_at` stays frozen at creation timestamp, never advancing per `cron_expression`.

Example:
routine: "Hourly cron PM triage"
trigger:
cron_expression: */15 * * * *
next_run_at: 2026-05-02 20:24:58 UTC (created 2 days ago)
last_fired_at: NULL (never fired)
enabled: true

`next_run_at` should recalculate to `now() + interval '15 min'` each tick. Instead frozen permanently.

## Impact

- PM agent idle for hours despite scheduled triage
- Duplicate ticket issues from routines firing through other paths without dedup
- Circuit breakers depending on routines never run
- Manual workarounds per affected routine — does not scale

## Workaround

Linux cron `*/1 * * * *` script that scans overdue triggers + manually advances `next_run_at` + inserts `agent_wakeup_requests` + `heartbeat_runs` companion. Plus existing `r2-fire.sh` (hardcoded routine_id).

## Suggested fix

Paperclip server cron daemon should:

1. On each tick (every minute), query `routine_triggers WHERE enabled=true AND next_run_at < now()`.
2. For each overdue trigger:
   a. Compute next firing using `cron_expression` (e.g. `croner` or `node-cron`).
   b. UPDATE `next_run_at = computed_next`.
   c. INSERT `agent_wakeup_requests` for assignee.
   d. INSERT `heartbeat_runs` (claim queue companion — without this, claim worker never picks up).
3. UPDATE `last_fired_at = now()`.

## Reproduction

1. Create routine via UI or SQL with `cron_expression='*/15 * * * *'`, `enabled=true`.
2. Wait 30+ minutes.
3. Observe `next_run_at` stuck, `last_fired_at` NULL.
4. Agent never wakes despite valid cron expression.

## Related

`agent_wakeup_requests` queued without `heartbeat_runs` companion never get claimed by worker either — separate but related issue. Drain cron must insert BOTH in single transaction.

PR #9752, opened 2026-07-17
fix(routines): isolate per-firing dispatch failures in tickScheduledTriggers

DarkInnerCircle:
## Thinking Path

> - Scheduled Routines are the cron-driven trigger mechanism (`routineService.tickScheduledTriggers`) that fires periodic agent work — the control plane claims due firings, advances `next_run_at` atomically, then dispatches each one
> - When `dispatchRoutineRun` throws synchronously for a claimed firing (e.g. the assignee agent was terminated or otherwise became unassignable), that specific firing has no per-row handling: `next_run_at` was already advanced past it, no `routineRuns` row is ever created, and the only trace is a generic log line at the call site — the firing is silently dropped
> - Worse, that thrown exception propagates out of the runCount loop for that trigger and aborts every other due routine still waiting in the same tick's batch, so one broken routine can stall an entire company's scheduled work until someone notices and fixes it by hand
> - This pull request wraps each claimed firing's dispatch in its own try/catch, records every failure as a durable `routineRuns` row via a new `recordFailedScheduleRun` helper, and reconciles by a deterministic per-firing `idempotencyKey` to avoid double-recording an ambiguous-commit case
> - The benefit is that a single misconfigured or unassignable routine degrades gracefully (loud logs, a queryable failure record per firing) instead of silently vanishing or freezing every other routine's schedule

Refs #5184, Refs #9445 — related but not identical: #5184 describes `next_run_at` never advancing at all (a different observable than this PR's target, which is firings that *do* get claimed/advanced but then have their dispatch outcome silently dropped on a synchronous throw). #9445 is symptomatically related (a routine that never accumulates `routineRuns` rows), which this PR's durable failure-recording directly helps diagnose.

### Relationship to #5238

While preparing this PR I found **#5238** ("fix(routines): isolate per-trigger errors so one bad cron does not freeze the batch"), which is open and targets an overlapping piece of the same loop. Flagging this explicitly rather than leaving it for review to discover:

- **Where we overlap:** #5238 also wraps each `dispatchRoutineRun` call inside the per-trigger `runCount` loop in its own try/catch, for the same reason this PR does — so one firing's dispatch failure can't unwind the loop and take the rest of the batch's due triggers down with it. That specific mechanism (per-dispatch-call isolation) is functionally the same idea in both PRs.
- **Where #5238 goes further than this PR:** #5238 *additionally* wraps the outer per-trigger block, so a corrupted `cronExpression`/timezone that makes `nextCronTickInTimeZone` throw is also isolated per-trigger. This PR does not touch that failure mode at all — it assumes the trigger's cron/timezone already validates and focuses only on the dispatch-time failure.
- **Where this PR goes further than #5238, on the overlapping dispatch-failure part:**
  1. **Durable, queryable failure records.** #5238 logs and increments a `failed` counter on a dispatch failure but never creates a `routineRuns` row for it — there's no persisted record of which firing failed or why. This PR's `recordFailedScheduleRun` inserts a `routineRuns` row with `status: "failed"` and a `failureReason`, so a failure is queryable and diagnosable after the fact, not just visible in logs at the moment it happened.
  2. **Ambiguous-commit reconciliation.** `dispatchRoutineRun`'s own transaction can, in principle, durably commit and still have the awaiting call throw (e.g. a driver reporting failure after the `COMMIT` actually landed). Neither the pre-existing code nor #5238 guards against this — a compensating "failed" row could end up recorded right next to a real successful run and contradict it. This PR passes a deterministic per-firing `idempotencyKey` (`schedule:<triggerId>:<firingScheduledAt>:<index>`) into `dispatchRoutineRun` and, on a caught error, reconciles by exact key match (any status) before inserting a compensating failed row — skipping the insert entirely if a row already exists under that key, and logging loudly (never inserting) if the reconciliation lookup itself errors.
  3. **Post-commit telemetry guard.** `dispatchRoutineRun`'s telemetry call happens after its own transaction commits; this PR guards that call so a telemetry throw can no longer be mistaken for a dispatch failure on an already-durably-committed run.

I'm happy to rebase this PR to build on top of #5238 once it merges (dropping the parts of my diff that duplicate its per-dispatch try/catch and keeping only the durable-recording + reconciliation delta), or to fold items 1–3 above into #5238 directly as review feedback there, whichever maintainers prefer. Opening this as a separate PR for now mainly so the durable-recording + reconciliation approach is visible and reviewable on its own, since it's a materially different failure-handling strategy (persisted state + idempotent reconciliation) rather than log-and-count.

## What Changed

- `server/src/services/routines.ts` (`tickScheduledTriggers`):
  - Each claimed firing's `dispatchRoutineRun` call (inside the per-trigger `runCount` loop, including multi-firing catch-up batches) is now wrapped in its own try/catch instead of being a bare `await` that, on throw, unwinds out of the whole tick.
  - A deterministic per-firing `idempotencyKey` (`schedule:<triggerId>:<firingScheduledAt ISO>:<index>`) is now passed into `dispatchRoutineRun`, which already accepted and stored that field. This gives the reconciliation check below an exact row to look for instead of a time-window guess.
  - On a caught dispatch error: log the first failure for a trigger in a tick at `error`, downgrade repeats for the same trigger to `warn` with an occurrence counter (bounds log volume for a persistently broken routine burning through its whole catch-up cap); then reconcile by exact `idempotencyKey` match (any status) to check whether the failed attempt actually committed a durable outcome despite throwing; only if no such row exists (and the reconciliation lookup itself didn't error) insert a compensating `routineRuns` row via the new `recordFailedScheduleRun` helper.
  - New `recordFailedScheduleRun` helper (modeled on the existing `recordSuppressedAutomaticRun` pattern used for paused-project/worktree-cutoff skips): inserts a `routineRuns` row with `status: "failed"` and `failureReason`, updates the trigger's touched-state, and best-effort logs an activity entry.
  - `dispatchRoutineRun`'s post-commit `trackRoutineRun` telemetry call is now guarded so a telemetry throw can't be mistaken for the run itself having failed.
- `server/src/__tests__/routines-service.test.ts` — five new regression tests: batch continuation across multiple due routines with a causal (not just timestamp) assertion that a later routine's dispatch isn't blocked by an earlier one's failure; one failed-run row recorded per claimed catch-up firing, capped at `MAX_CATCH_UP_RUNS`; the failure-recorder's own transaction erroring doesn't abort the rest of the tick's batch; an ambiguous-commit scenario (dispatch's transaction durably commits, caller still observes a throw) does not produce a duplicate/contradictory row; a catch-up iteration's genuine success does not suppress the very next iteration's genuine, distinct failure, verified via each row's distinct index-suffixed `idempotencyKey`.
- `server/src/__tests__/routine-run-telemetry.test.ts` — one new regression test: a synchronous throw from the telemetry call after a run has already committed does not flip that run's recorded status away from its real (successful) outcome.

## Verification

This branch is rebased onto current `master` (as of commit `5d42382`), which had already diverged from the `v2026.707.0` tag with an unrelated, substantial `routines.ts` rewrite (activity-gate suppression, worktree-execution-cutoff suppression, folder support, and a new sub-hourly catch-up coalescing guard via `isSubHourlyCronExpression`). That required real conflict resolution, not a clean apply — noting for reviewers:

- The tick loop's catch-up branch now needs both master's sub-hourly short-circuit *and* this PR's per-firing `firingScheduledAts` timestamp tracking; both are preserved.
- The dispatch loop now runs this PR's try/catch/idempotency logic *underneath* master's new activity-gate and worktree-suppression pre-checks (which run before any dispatch attempt and are unaffected by this PR), and also picks up master's `nextRunAtOverride` correctness fix for multi-firing catch-up batches (previously each loop iteration independently recomputed its own `nextRunAt`; master pins all iterations in a batch to the same claimed value).
- Two new tests originally used an every-minute cron (`* * * * *`) for catch-up scenarios; master's new sub-hourly guard now intentionally collapses such crons to a single firing instead of a full replay (to prevent catch-up floods), so those two tests were changed to an hourly cron with proportionally adjusted timestamps — same assertions, same `MAX_CATCH_UP_RUNS = 25` cap logic, just no longer sub-hourly.

Ran all 8 routine-related test files on this master-based branch:

```
npx vitest run \
  src/__tests__/plugin-managed-routines.test.ts \
  src/__tests__/qa-routine-secrets-e2e.test.ts \
  src/__tests__/routine-document-annotation-routes.test.ts \
  src/__tests__/routines-routes.test.ts \
  src/__tests__/routine-run-telemetry.test.ts \
  src/__tests__/routines-e2e.test.ts \
  src/__tests__/routines-service.test.ts \
  src/services/routines-formatter-cache.test.ts
# Test Files  8 passed (8)
#      Tests  97 passed (97)
```

`tsc --noEmit` on this branch reports 128 errors, all pre-existing `@paperclipai/plugin-sdk` module-resolution failures unrelated to this change (same category on a clean `master` checkout before this diff); zero errors in any file touched by this PR.

Every fix in this PR was demonstrated RED/GREEN during development by selective reversion (each piece — per-firing try/catch, idempotency reconciliation, telemetry guard — was temporarily removed to confirm its corresponding new test fails without it, then restored).

## Risks

- Low-to-moderate. This changes failure-path behavior only (the happy path — successful dispatch — is unchanged); a broken routine that previously silently vanished or froze the batch now produces a `routineRuns` row per failed firing and a log line, and no longer blocks other due routines in the same tick.
- `routineRuns.idempotencyKey` has no unique index today. The exact-key reconciliation lookup in this PR is strictly safer than the pre-existing code (which did no reconciliation at all), but a uniquely-enforced per-firing constraint would be a stronger guarantee than lookup-then-insert. Recommending that as explicit follow-up hardening; a schema migration felt out of scope for this PR.
- Pre-existing, untouched by this PR: a process crash between the `next_run_at` claim (CAS update) and the dispatch attempt still leaves a firing with no run row at all. This PR does not add crash-recovery for that window.
- A firing that both genuinely fails *and* simultaneously hits a reconciliation-query error produces no run row for that specific firing — this is deliberate (the failure is still observable via the loud `error`-level log line at that point), preferring "no row, but observable" over risking a row that might contradict a real, already-committed success.

## Model Used

Claude (Anthropic), Claude Opus/Fable-tier reasoning model accessed via Claude Code, with tool use (repository search, file edits, running `tsc`/`vitest` in a local checkout) and extended multi-step reasoning for the conflict resolution against `master`'s independent `routines.ts` changes. No separate research/reasoning model was used for this PR.

## Checklist

### DIFF
--- a/server/src/__tests__/routine-run-telemetry.test.ts
+++ b/server/src/__tests__/routine-run-telemetry.test.ts
@@ -167,4 +167,25 @@ describeEmbeddedPostgres("routine run telemetry", () => {
       status: "issue_created",
     });
   });
+
+  it("does not fail an already-committed run when telemetry throws synchronously", async () => {
+    const { routine, svc } = await seedFixture();
+    mockTrackRoutineRun.mockImplementationOnce(() => {
+      throw new Error("telemetry state factory blew up");
+    });
+
+    // The run/issue transaction has already committed by the time telemetry runs. A throw
+    // there must not surface as a dispatch failure for a run that durably exists.
+    const run = await svc.runRoutine(routine.id, { source: "manual" });
+
+    expect(run.status).toBe("issue_created");
+
+    const runs = await db
+      .select()
+      .from(routineRuns)
+      .where(eq(routineRuns.routineId, routine.id));
+    expect(runs).toHaveLength(1);
+    expect(runs[0]?.status).toBe("issue_created");
+    expect(runs[0]?.failureReason).toBeNull();
+  });
 });
--- a/server/src/__tests__/routines-service.test.ts
+++ b/server/src/__tests__/routines-service.test.ts
@@ -1,5 +1,5 @@
 import { createHmac, randomUUID } from "node:crypto";
-import { eq } from "drizzle-orm";
+import { and, eq } from "drizzle-orm";
 import { afterAll, afterEach, beforeAll, describe, expect, it, vi } from "vitest";
 import {
   activityLog,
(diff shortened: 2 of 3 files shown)


marijnp7:
**Field report: we hit this failure mode in production, and the forensic gap may be slightly wider than the PR assumes**

We ran into what looks like the exact failure this PR addresses, on a self-hosted instance running this codebase. Sharing the forensics in case a concrete field case is useful for prioritisation. No merge request implied — plus one observation that may be worth folding into the change.

**The case.** A weekly scheduled trigger stopped producing runs for 78 days (2026-05-21 → 2026-08-10) with no visible error.

- `routine_revisions`: exactly 1 row across the entire window → the trigger was never touched via `updateTrigger`, so `next_run_at` was advancing solely through the claim-`UPDATE` inside `tickScheduledTriggers`.
- `routine_runs`: exactly 2 rows across the same 11 weeks (the first fire, and a manual fire on 2026-08-11) — **no `failed` and no `skipped` rows at all**. That rules out pause/suppression and places the throw before any durable run record survives.

Net effect, every week: `next_run_at` moved forward, nothing ran, nothing was recorded.

**It is intermittent, not corrupt data.** On 2026-08-11 14:50 UTC that same trigger fired through the normal schedule path and dispatched fine (`issue_created`). A throwaway routine of the same shape (`coalesce_if_active`, cron `*/2 * * * *`) produced two fires and two correct `routine_runs` rows, with `last_fired_at` / `next_run_at` / `last_result` all updated. So the schedule path is healthy in general and goes silent only on a specific fault — which is what this PR is about.

**The observation that may matter for the fix.** Isolating each firing in its own `try/catch` correctly stops one bad routine from killing the rest of the tick. But on its own it does not guarantee that a `failed` row survives. In `dispatchRoutineRun`, the `routine_runs` insert *and* the failure fallback (`finalizeRun({status: "failed", …})` + `updateRoutineTouchedState`) run inside the **same** `db.transaction`, and the fallback writes use that same `tx`. That works for a plain JS throw. On a **Postgres** error the transaction is already aborted (`25P02`), so the fallback statements fail as well and the whole transaction rolls back — taking the run row with it. The claim-`UPDATE` from `tickScheduledTriggers` has already committed outside that transaction. Result: `next_run_at` advanced, zero traces — which matches our signature exactly.

The same reasoning applies to the `23505` fallback on `issues_open_routine_execution_uq`: the `findLiveExecutionIssue` query that follows it runs inside the aborted transaction.

So if the goal is that the failure reports itself, the failure recording likely has to happen **outside** the failed transaction (its own connection/transaction) or behind a `SAVEPOINT`. From the description, `recordFailedScheduleRun` looks like it is positioned to do exactly that — it may be worth confirming explicitly that it is never enlisted in the aborted transaction, since that is the property the whole fix rests on.

**Caveats, stated plainly.** We read `server/src` from the deployed image; what actually runs is the `dist` build of it. Server logs were not accessible to us, so we do not know the concrete Postgres error behind the 2026-08-10 firing — that specific fault is a hypothesis. The measured signature (advancing `next_run_at`, exactly zero run rows across 11 weeks, no `failed`/`skipped`) is not.

PR #9753, opened 2026-07-17
fix(security): keep bridge credentials out of remote argv

jodylarsen:
## Summary
- deliver the per-bridge credential to the remote launcher over stdin instead of the command runner environment
- export it only inside the remote shell so the detached bridge inherits authorization without exposing the value in spawned argv
- add a sentinel regression assertion while preserving bridge authorization and teardown coverage

## Test plan
- [x] `npx --yes pnpm@9.15.4 exec vitest run packages/adapter-utils/src/sandbox-callback-bridge.test.ts`
- [x] `npx --yes pnpm@9.15.4 --filter @paperclipai/adapter-utils typecheck`

### DIFF
--- a/packages/adapter-utils/src/sandbox-callback-bridge.test.ts
+++ b/packages/adapter-utils/src/sandbox-callback-bridge.test.ts
@@ -114,7 +114,7 @@ describe("sandbox callback bridge", () => {
     }
   });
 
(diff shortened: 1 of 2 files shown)


jodylarsen:
Superseded by a replacement PR using a descriptive branch name and the repository's required PR template.

PR #9768, opened 2026-07-17
fix(authz): let a direct-child assignee comment on its parent issue

ericdfields:
## Problem

A child issue's assignee gets `403 "Issue is outside this actor's authorization boundary"` when it tries to post a comment on its **direct parent** issue. The agent `issue:comment` decision only grants access for self-assigned / unassigned / mention-granted issues — never "the parent of an issue I'm assigned to." In practice this forces finished child work to be relayed up through a manager/CEO actor purely to satisfy the authorization boundary, manufacturing pure-overhead runs.

## Fix

Add one narrowly-scoped grant to the agent branch of the `issue:comment` authorization decision: if the actor is the assignee of a **direct child** of the target issue, allow the comment.

Deliberately conservative:

- **`issue:comment` only.** Destructive `issue:mutate` (status / assignee / title changes) on the parent stays denied.
- **Single hop.** Direct parent only — grandparents and higher ancestors are not granted.
- **Agent-actor path only.** No change to user authorization.
- **Cheap on the deny tail.** A single indexed lookup (`companyId + parentId + assigneeAgentId`, `LIMIT 1`) that only runs after the existing fast early-exits (self / unassigned / mention-grant).

## Tests

Adds `authorization-service.test.ts` coverage asserting the positive grant **and** its invariants:

- `issue:comment` on the direct parent is **allowed** for the child's assignee.
- `issue:mutate` on the parent stays **denied** even when `issue:comment` is allowed.
- A grandparent (two hops up) stays **denied**.
- An unrelated agent stays **denied**.

Full server suite green locally (`tsc --noEmit` clean).

## Scope note

Attachments to the parent are intentionally **not** included here — an attachment is a deliverable write that should prove run-id + checkout ownership, so it needs its own liveness path rather than riding the comment grant. That is tracked as a separate follow-up. (The in-code comment mentions attachment aspirationally; only the `issue:comment` guard is wired.)

### DIFF
--- a/server/src/__tests__/authorization-service.test.ts
+++ b/server/src/__tests__/authorization-service.test.ts
@@ -700,6 +700,111 @@ describeEmbeddedPostgres("authorization service", () => {
     });
   });
 
+  it("lets a direct-child assignee comment on its parent without granting parent mutation (BRO-1290)", async () => {
+    const company = await createCompany(db, "ChildParentComment");
+    const parentAssignee = await createAgent(db, company.id, { role: "engineer" });
+    const childAssignee = await createAgent(db, company.id, { role: "engineer" });
+    const unrelatedAgent = await createAgent(db, company.id, { role: "engineer" });
+    const parentIssue = await createIssue(db, company.id, {
+      title: "Parent issue",
+      assigneeAgentId: parentAssignee.id,
+    });
+    await createIssue(db, company.id, {
+      title: "Child issue",
+      parentId: parentIssue.id,
+      assigneeAgentId: childAssignee.id,
+    });
+
+    const authorization = authorizationService(db);
+    const parentResource = {
+      type: "issue",
+      companyId: company.id,
+      issueId: parentIssue.id,
+      projectId: parentIssue.projectId,
+      parentIssueId: parentIssue.parentId,
+      assigneeAgentId: parentAssignee.id,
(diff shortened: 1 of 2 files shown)


ericdfields:
Closing as superseded by the default-open issue-write model that has since landed on `master`.

`decideVisibleIssueWrite` now default-opens `issue:comment` for any standard-trust agent that can *read* the issue, and `isDirectParentReportTarget` covers the run-scoped direct-parent case explicitly. Between them, the case this PR was opened for is already allowed.

The rebase is also a no-op rather than a behaviour change. On current `master` this PR's allow-block lands immediately below:

```ts
// server/src/services/authorization.ts
if (visibleIssueWriteDecision) return visibleIssueWriteDecision;
```

That return is unconditional for `issue:comment`. Execution only reaches this point when `trustResolution.kind === "standard"` — `decideLowTrustAccess` returns `null` only for `"standard"`, and every other kind returns a decision that is returned earlier for `issue:comment`. `"standard"` is exactly the condition that makes `visibleIssueWriteDecision` non-null, so nothing below that line for `issue:comment` is reachable. Rebasing would have merged an unreachable rule that reads like a live authz change.

If the underlying intent was broader — commenting on issues the agent cannot *read* — that is a separate widening that has to sit above the default-open check and deserves its own review rather than a rebase of this branch.

PR #9775, opened 2026-07-17
Safe document-save helper + mandate (SOL-505)

blackridge-alfred:
## Problem (SOL-505)
`PUT /api/issues/{id}/documents/{key}` uses optimistic concurrency and returns **HTTP 409** — not a silent no-op — when `baseRevisionId` is missing or stale, with `details.currentRevisionId` in the body. Agents hand-roll the PUT and don't check the status, so they treat the 409 as success. On SOL-43 this silently dropped three consecutive "saved" revisions (claimed 4/5/6, doc never advanced past 3) until a human cross-checked. The 409 contract is intentional; the defect is client-side.

## Change
- **New `skills/paperclip/scripts/paperclip-save-document.sh`** — the safe writer every agent should use instead of a bare curl PUT:
  1. reads the current document for `latestRevisionId`,
  2. sends `baseRevisionId` on update / omits on create,
  3. on 409 retries **once** with `details.currentRevisionId`,
  4. **refuses to report success on any non-2xx** (exits non-zero),
  5. verifies by re-reading that `latestRevisionNumber` advanced.
  Matches `paperclip-upload-artifact.sh` conventions (flags, `--dry-run`, `--output markdown|json`).
- **`skills/paperclip/SKILL.md`** — mandate the helper next to the existing plan-document guidance; forbid reporting a save without a verified 2xx + revision advance.
- **`server/src/onboarding-assets/default/AGENTS.md`** — one-line execution-contract rule so every agent gets it in its base instructions.

## Validation (live against the running API)
```
create              -> revision 1 (verified)
naive PUT (no base) -> HTTP 409  {"error":"Document update requires baseRevisionId","details":{"currentRevisionId":"..."}}
helper update       -> revision 2 (recovered from 409, verified)
```
`bash -n` clean; `--dry-run` verified.

## Notes for reviewer
- No backend/behavior change — 409 semantics are intentional and untouched; this is client tooling + guidance only.
- Follow-on (issue rec #2): audit other mutating endpoints for swallowed 4xx preconditions.
- Design doc + full context on SOL-505.


blackridge-alfred:
Superseded by #9776, which addresses the review: fixes the `--dry-run` unbound-variable crash (P1) and the post-write-GET-404 verification message (P2), and refiles using the required PR template with no internal ticket references and a public in-PR bug description. Closing in favor of #9776.

PR #9787, opened 2026-07-17
feat(codex-local): add outbound direction to codex-auth-merge decision predicate

nickyleach:
## Thinking Path

> - Paperclip runs Codex as a sandboxed agent with its own CODEX_HOME; at the start of each run Paperclip merges the sandbox's auth.json with the host's so the agent can authenticate to the Codex API
> - The merge uses `codex-auth-merge-decision.cjs`, a small CJS script that reads two `auth.json` snapshots and exits `10` (use source copy) or `20` (keep destination copy)
> - Until now the script only handled the **inbound** direction (host → sandbox) with hard-coded sandbox/host semantics
> - Sandbox-to-host copy-back (**outbound**) requires the same freshness guard but with swapped roles — and exposing both directions via a flag requires hard-coding knowledge about which role is "sandbox" and which is "host" into the predicate itself
> - A cleaner design: the predicate is **direction-agnostic** — it accepts a `source` path and a `destination` path; the caller decides direction by which argument it passes first. The script answers one question: "should destination be replaced by source?" Exit `10` = yes (use source); exit `20` = no (keep destination)
> - This PR refactors `codex-auth-merge-decision.cjs` to this source/destination model, removes the `--direction` flag, and extends coverage with a 15-row source/destination predicate matrix plus updated extract integration tests

No public GitHub issue exists. The underlying problem:

**Context** — Paperclip merges `auth.json` files between sandbox and host to keep Codex credentials in sync across agent runs. The existing decision predicate (`codex-auth-merge-decision.cjs`) handled the inbound direction (host→sandbox) with hard-coded sandbox/host naming. A copy-back (outbound) direction was needed, but rather than add a `--direction` flag that hard-codes role knowledge into the predicate, the design was simplified: the predicate is now **direction-agnostic**, taking `source` and `destination` paths and answering "replace destination with source?" The freshness guard is unified: exit `10` only when source and destination share the same `account_id`, same subscription-kind, both carry a parseable `last_refresh`, and `source.last_refresh > destination.last_refresh` (strict). Ties and null freshness keep the destination copy so a spent single-use refresh token (openai/codex#15502) is never written back over a valid one.

Related PRs:
- Refs #9785 — Phase 2: relocated the decision predicate from `adapter-utils` into `codex-local`; this PR builds on that merged change
- Refs #9621 — community PR targeting Codex sandbox auth sync-back via a different approach; operates on different files (runtime layer vs. decision script), non-conflicting

## What Changed

- `packages/adapters/codex-local/src/server/codex-auth-merge-decision.cjs` — refactored from direction-parameterized (sandbox/host with `--direction`) to **direction-agnostic source/destination** semantics. Args: first = source path, second = destination path. Exit `10` (use source) only when: same `account_id`, same subscription-kind, both carry a parseable `last_refresh`, and `source.last_refresh > destination.last_refresh` (strict). Every other case exits `20` (keep destination). Unknown/extra flags are still rejected with exit `1`. No token bytes are emitted to stdout/stderr.
- `packages/adapters/codex-local/src/server/codex-auth-merge.test.ts` — rewritten to source/destination API: 15-row predicate matrix (strictly-newer→`10`; tie/older/identity-mismatch/null-refresh/apikey/kind-mismatch/unusable→`20`) plus token-bytes non-emission guard. The inbound extract integration suite updated to match.
- Runtime callers (`codex-auth-merge-extract.sh`) already invoke `node decision <sandbox> <host>` = `(source, destination)` and branch on exit `10` → keep sandbox, so **no caller changes are needed**.

**Behavioral note:** unifying the two directions under a single "use source only when strictly fresher" rule means a tie (exact equal `last_refresh`) now keeps the **destination** for both the inbound and outbound frames. For inbound restore, this means an exact timestamp tie keeps the host copy. On a tie, both credentials are the same generation — preferring the canonical destination (host) copy is more conservative. This is an intentional consequence of the unified guard.

## Verification

```sh
# From the codex-local package root:
pnpm vitest run src/server/codex-auth-merge.test.ts
# Expected: all predicate rows green (15-row source/destination matrix + token-leak guard)
# Plus: codex-auth-merge-extract integration rows green
```

The test suite drives the `.cjs` as a child process over temporary `auth.json` fixtures:
- Source/destination predicate matrix (15 rows): strictly-newer→`10`; tie→`20`; older→`20`; identity-mismatch→`20`; source last_refresh unparseable→`20`; destination last_refresh unparseable→`20`; apikey either side→`20`; kind-mismatch→`20`; unusable either side→`20`
- Token-bytes non-emission guard: asserts no auth token bytes appear in `stdout`/`stderr`
- Inbound extract integration: `codex-auth-merge-extract.sh` end-to-end rows including tie/identity-mismatch/sandbox-newer cases

## Risks

- **Tie behavior change**: exact `last_refresh` ties now always keep the destination copy (was: keep-sandbox for inbound). On a tie both credentials are the same generation; keeping the canonical destination (host) copy is the more conservative choice. Covered by test assertions.
- **Token non-emission**: `stdout`/`stderr` of the `.cjs` must not echo credential bytes; covered by the token-bytes non-emission guard test.
- **Unknown/extra flags**: handled with `process.exit(1)` rather than a silent default, so misconfigured callers surface errors immediately.
- **Caller compatibility**: the only runtime caller, `codex-auth-merge-extract.sh`, already calls `node decision <sandbox> <host>` with no flags — compatible with the new source/destination API with no changes required.
- **Low risk overall**: the script is exercised end-to-end via the extract integration tests; all paths covered.

## Model Used

Claude claude-sonnet-4-6 (Anthropic, `claude-sonnet-4-6`), 200 k context window, tool use and agentic code execution enabled. Automated Paperclip multi-agent system.

## Checklist

### DIFF
--- a/packages/adapters/codex-local/src/server/codex-auth-merge-decision.cjs
+++ b/packages/adapters/codex-local/src/server/codex-auth-merge-decision.cjs
@@ -41,32 +41,73 @@ function parseAuth(filePath) {
   };
 }
 
-const [sandboxAuthPath, hostAuthPath] = process.argv.slice(2);
+// Exit codes are direction-agnostic: 20 = keep the host auth.json (inbound: do
+// not overwrite the sandbox copy on the way in; outbound: do not copy the
+// sandbox copy back to the host); 10 = use the sandbox auth.json.
+const KEEP_HOST = 20;
+const USE_SANDBOX = 10;
+
+// Parse an optional `--direction=inbound|outbound` flag (default `inbound`) out
+// of argv while leaving the two positional auth paths in order. Inbound is the
+// original host→sandbox merge decision; outbound is the sandbox→host copy-back
+// guard. Fail loudly on an unrecognised direction rather than silently defaulting.
+const rawArgs = process.argv.slice(2);
+let direction = "inbound";
+const positional = [];
+for (const arg of rawArgs) {
+  if (arg.startsWith("--direction=")) {
+    direction = arg.slice("--direction=".length);
+  } else {
+    positional.push(arg);
+  }
+}
+if (direction !== "inbound" && direction !== "outbound") {
+  console.error(`[paperclip] unknown --direction=${direction}; expected inbound|outbound`);
+  process.exit(1);
+}
+
+const [sandboxAuthPath, hostAuthPath] = positional;
 const sandboxAuth = parseAuth(sandboxAuthPath);
 const hostAuth = parseAuth(hostAuthPath);
 
+// Shared identity/kind pre-checks. Both directions refuse to move a credential
+// unless the two sides are the same usable, subscription-kind identity — an
+// unusable side, an api-key credential, a kind mismatch, or a different
+// account_id all keep the host copy regardless of direction.
 if (
   hostAuth.kind === "unusable" ||
   sandboxAuth.kind === "unusable" ||
-  sandboxAuth.kind !== hostAuth.kind
+  sandboxAuth.kind !== hostAuth.kind ||
+  hostAuth.kind === "apikey" ||
+  sandboxAuth.accountId !== hostAuth.accountId
 ) {
-  process.exit(20);
+  process.exit(KEEP_HOST);
 }
 
-if (hostAuth.kind === "apikey") {
-  process.exit(20);
+if (direction === "outbound") {
+  // Copy the sandbox credential back to the host only when it is strictly
+  // fresher: both sides must carry a parseable last_refresh and the sandbox one
+  // must be strictly greater. Ties and null/unparseable freshness keep the host
+  // copy so a spent single-use refresh token is never written back over a good one.
+  if (
+    sandboxAuth.lastRefresh !== null &&
+    hostAuth.lastRefresh !== null &&
+    sandboxAuth.lastRefresh > hostAuth.lastRefresh
+  ) {
+    process.exit(USE_SANDBOX);
+  }
+  process.exit(KEEP_HOST);
 }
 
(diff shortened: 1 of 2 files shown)


nickyleach:
@greptile review

nickyleach:
I think we should change the semantics of `codex-auth-merge-decision.cjs`. Instead of hard-coding knowledge about `sandbox` vs `host`, we should treat it as `source` vs `destination`. This removes the `--direction` flag since the caller is implicitly deciding direction based on their own frame of reference.

nickyleach:
Thanks for the review! Routing the requested semantic change — **source/destination instead of sandbox/host, removing the `--direction` flag** — to the backend engineer. They'll push the updated commit to this branch and I'll re-run CI/Greptile checks before re-routing to the board for merge.

Note to board: please hold off on PAP-2771 until this change lands and I re-verify.

nickyleach:
@greptile review

nickyleach:
@greptile review

nickyleach:
**Greploop round-2 complete — 5/5 ✓**

Re-ran greploop after Priya's source/destination refactor (commit `e9d4916d7`).

- PR body updated to accurately describe the direction-agnostic design, the tie-behavior change, and caller compatibility (removed stale `--direction` references and the incorrect "byte-for-byte unchanged" claims)
- Greptile re-reviewed the updated PR body and code: **5/5**, no unresolved comments
- CI: CLEAN / all required checks green
- Checklist: all boxes checked

PR is merge-ready. Handing to board for merge review.

PR #9816, opened 2026-07-18
fix(paperclip-skill): don't let a transient API blip abort a heartbeat

quratus:
## Problem
The heartbeat inbox fetch treated **any** `inbox-lite` failure — a timeout / connection-refused / 5xx during the control-plane's restart window — the same as an **empty inbox**. So a scoped wake or a normal heartbeat that hit a few-second `:3100` blip exited with zero work done.

Observed live: SQN-4147 (a scoped, named-issue wake) exited with `inbox-lite timeout — exiting` while the API was actually **healthy** — it had just caught the ~15s launchd restart window (the SQN-3368 OOM crash loop).

## Fix — three clarifications to the heartbeat rules in `skills/paperclip/SKILL.md`
- **Scoped-wake fast path** also triggers on `PAPERCLIP_TASK_ID` and must skip **any** inbox gate. A transient control-plane error never aborts a named-issue wake — retry checkout/context ~3×/5s to ride out the restart window.
- **Step 3 inbox fetch is not a fatal gate:** on a transient failure, retry with short backoff, then fall back to the full-issues query. *A failed request is not an empty inbox.*
- **"Nothing assigned → exit"** now means a *successful* 2xx response with zero items — never a timeout / connection-refused / 5xx.

## Scope / risk
Doc-only change to the shared heartbeat skill. No code paths touched. The matching hardened gate (retry + explicit `INBOX_EMPTY` vs `API_UNREACHABLE` verdict) was also applied to the 6 local orchestration agents' `AGENTS.md`; the VPS implementer agents are governed by this skill, which is why the fix belongs here in git.

Root cause of the flap itself (the `:3100` OOM restart, SQN-3368) is tracked separately and unchanged by this PR.

### DIFF
--- a/skills/paperclip/SKILL.md
+++ b/skills/paperclip/SKILL.md
@@ -29,7 +29,7 @@ Manual local CLI mode (outside heartbeat runs): use `paperclipai agent local-cli
 
 Follow these steps every time you wake up:
 
-**Scoped-wake fast path.** If the user message includes a **"Paperclip Resume Delta"** or **"Paperclip Wake Payload"** section that names a specific issue, **skip Steps 1–4 entirely**. Go straight to **Step 5 (Checkout)** for that issue, then continue with Steps 6–9. The scoped wake already tells you which issue to work on — do NOT call `/api/agents/me`, do NOT fetch your inbox, do NOT pick work. Just checkout, read the wake context, do the work, and update.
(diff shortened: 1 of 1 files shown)


quratus:
Closing per updated org policy: quratus/paperclip is our canonical master going forward. Retargeted to https://github.com/quratus/paperclip/pull/8 (same branch, same content).

PR #9819, opened 2026-07-18
fix(recovery): wake and release blocked dependents when their blocker dies

### LINKED ISSUE #9811: Blocked dependents are never woken when their blocker dies (cancelled or recovery-stranded)
## Problem

A dependent issue is woken **only** when its blocker reaches `done`:

`server/src/routes/issues.ts` — `becameDone` -> `listWakeableBlockedDependents` -> dependency-resolved wake.

There is no equivalent path for a blocker that **dies**. When the blocker ends any other way, its blocked dependents are never woken and never released:

1. **Blocker cancelled.** `becameTerminal` covers `done` *and* `cancelled`, but it only wakes the **parent** (`getWakeableParentAfterChildCompletion`) — not the blocked dependents.
2. **Blocker stranded by recovery.** `escalateStrandedAssignedIssue` moves the issue to `blocked` ("no live execution path… visible for intervention"). Dependents are not notified.

The periodic dependency backstop does not help either: it only wakes issues whose blockers are **all `done`** (`isDependencyReady` is `blockerStatus === "done"`), so a cancelled/stranded blocker leaves `isDependencyReady = false` and the dependent is skipped.

The task watchdog does not catch it either — `scheduled_retry` counts as a live run status, so a blocker cycling in retry is classified as alive.

## Impact

In an autonomous fleet this is a silent, permanent stall. We hit it in production: a worker delegated a sub-task, blocked itself on it, the sub-task's run died on a provider overload, and the worker slept indefinitely. Nothing surfaced it — no wake, no alert. The data all looked healthy.

Worse, the natural operator remedy makes it permanent: cancelling the stuck sub-task puts the blocker in `cancelled`, which is exactly the case that wakes nobody.

## Two parts to it

Waking alone is not sufficient. Even when a wake is enqueued, the dependent stays `status: blocked` with the dead blocker still in `blockedByIssueIds`, and `isDependencyReady` stays false — so the scheduler will not dispatch a run for it. We verified this on a live instance: the wake fired and the dependent still sat there. The dependent has to be **released** as well as woken.

## Suggested fix

Mirror the `becameDone` path for blocker death, at the two points where the code knows for certain the blocker died (the update handler cannot distinguish a transitively-blocked blocker from a stranded one, so this must not be inferred from `status === "blocked"` alone):

1. `becameCancelled` in the issue update handler.
2. `escalateStrandedAssignedIssue` in recovery — **excluding** the provider-quota wait case, where a monitor will retry after the quota resets and the blocker may still reach `done` (waking there is a false alarm).

For each dependent: drop the dead blocker from `blockedByIssueIds`; if no live blocker remains, move `blocked` -> `todo`; emit a distinct wake reason (we used `issue_blocker_stranded`, with the blocker's fate in the payload) so the agent knows the dependency failed rather than resolved.

Note that `listWakeableBlockedDependents` cannot be reused here — it requires dependency readiness, which is false by definition for a dead blocker, so it returns an empty list. A readiness-independent lister is needed.

We have this implemented and running against a self-hosted instance, with the dependent verifiably going `blocked` -> `todo` and picking the work back up. Happy to open a PR if the approach looks right to you — in particular whether `todo` is the status you would want, or something that flags "dependency failed, needs re-triage" more explicitly.

Model Used: Claude Opus 4.8

greegorij:
## Thinking Path

> - Work is sequenced with `blockedBy`: a dependent waits until its blocker finishes, and the platform wakes it when that happens
> - The wake path exists for exactly one outcome — the blocker reaching `done` (`becameDone` → `listWakeableBlockedDependents` → dependency-resolved wake)
> - There is no equivalent path when the blocker **dies** rather than finishes: a cancelled blocker wakes only the parent, and a recovery-stranded blocker moved to `blocked` notifies nobody
> - So a dependent whose blocker was cancelled or stranded sleeps indefinitely. Neither backstop catches it, because from the platform's point of view nothing failed — the chain simply stops, quietly
> - This pull request adds the missing wake paths for both death cases, reusing the existing dependency-wake machinery rather than inventing a second mechanism
> - The benefit is that a dying blocker releases its dependents instead of stranding a whole subtree of work

Closes #9811

## What Changed

- `server/src/services/issue-dependency-wakeups.ts`: extracted the dependent-wake logic into a reusable path so both the `done` and the death cases go through the same code.
- `server/src/routes/issues.ts`: on transition to `cancelled`, blocked dependents are now woken, not just the parent.
- `server/src/services/recovery/service.ts`: when `escalateStrandedAssignedIssue` moves a blocker to `blocked`, its dependents are woken and released.
- `server/src/services/issues.ts`: `listAssignedDependentsOfBlocker` gained a `requireDependencyReady` option so the death path can select dependents that should be released even though the dependency did not complete normally.
- `server/src/services/heartbeat.ts`: wiring for the above.
- `server/src/__tests__/issue-dependency-wakeups-routes.test.ts`, `server/src/__tests__/issue-recovery-actions.test.ts`: coverage for the cancelled-blocker and stranded-blocker paths.

## Verification

- `pnpm --filter @paperclipai/server test` for the two touched test files → passes.
- Reproduced both stalls on a live instance before the change: cancelling a blocker left its dependent asleep with no wake event in the run log, and a recovery-stranded blocker did the same. After the change both dependents wake and the chain continues.
- Wake attribution is unchanged — dependents wake with the same reason and actor as the existing `done` path, so run logs stay readable.

## Risks

Medium risk, and I would rather flag it than undersell it.

- This makes agents wake in situations where they previously stayed asleep. On a board with many cancelled blockers, the first deployment can produce a burst of wakes as the backlog releases. That is the intended correction, but it is a real behavioral change with a token cost.
- `requireDependencyReady` adds a second selection mode to a shared query. If a future caller passes the wrong value, dependents could be released before their dependency is genuinely satisfied. The option is explicit at every call site and defaults to the pre-existing behavior, but it is the part of this change I would review hardest.
- No schema change, no migration.
- This branch currently conflicts with `master`. I will rebase and report back in this thread — please do not spend review time until it is clean.

## Model Used

Claude Opus 4.6 (`claude-opus-4-6`) authored the original change; this description was written with Claude Opus 5 (`claude-opus-5`), 1M context, adaptive thinking, with tool use and local command execution.

## Checklist

- [x] I have not referenced internal/instance-local Paperclip issues or links (only public GitHub references)
- [x] I have updated relevant documentation to reflect my changes — none found describing dependency-wake behavior; happy to add if there is a place for it
- [ ] All Paperclip CI gates are green — this branch conflicts with `master` and CI is red. Rebase in progress; I will report back here before asking for review.


greegorij:
Merged current `master` into the branch (merge, not rebase — so review threads pinned to existing commits survive). No conflicts remain.

One adjustment was needed, and it is worth naming because it is a genuine interaction with #10112 ("Route blocked transitions to explicit unblock owners"):

Entering `blocked` now requires unresolved blockers, a pending interaction/approval, or an explicit `unblockDescriptor`. One of my tests drove an issue into `blocked` without any of those and started getting a 422. The **production logic in this PR was unaffected** — only the test needed to state an unblock path, which it now does.

That said, #10112 and this PR are adjacent: yours makes a newly blocked issue carry a routable next action, mine wakes dependents when their blocker dies (cancelled, or moved to `blocked` by recovery). I believe they are complementary rather than overlapping, but if you see the stranded-blocker half as already covered by the unblock-owner routing, I am happy to narrow this PR to the cancelled-blocker case alone.

`issue-dependency-wakeups-routes.test.ts` 7/7, `issue-recovery-actions.test.ts` 43/43.

PR #9833, opened 2026-07-18
Fail closed on invalid Paperclip config

quratus:
## What changed

- Treat only `ENOENT` as an intentionally missing Paperclip config.
- Fail closed for unreadable files, malformed JSON, and schema-invalid config.
- Keep public errors path-aware without exposing raw config content or nested causes.
- Add focused regression coverage for all five config outcomes.

## Why

The previous broad catch returned `null` for every read, JSON, or schema failure. `loadConfig()` interprets `null` as “no config,” so an invalid selected config could silently continue with the default instance and database. This patch prevents that unsafe fallback.

## Impact

Valid and genuinely absent config behavior is unchanged. Any existing invalid or unreadable config now stops startup before default instance/database initialization.

## Validation

- `pnpm --filter @paperclipai/server exec vitest run src/__tests__/config-file.test.ts` — 1 file, 5 tests passed.
- `pnpm --filter @paperclipai/server typecheck` — passed.
- `git diff --check origin/master...HEAD` — clean.
- Independent QA Hunter review approved the patch under SQN-4190.

No migration, database mutation, or server restart is part of this PR.

### DIFF
--- a/server/src/__tests__/config-file.test.ts
+++ b/server/src/__tests__/config-file.test.ts
@@ -1,7 +1,7 @@
 import fs from "node:fs";
 import os from "node:os";
 import path from "node:path";
-import { afterEach, beforeEach, describe, expect, it } from "vitest";
+import { afterEach, beforeEach, describe, expect, it, vi } from "vitest";
 import { readConfigFile } from "../config-file.js";
 
 describe("readConfigFile", () => {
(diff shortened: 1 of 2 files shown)


quratus:
Closing per updated org policy: quratus/paperclip is our canonical master going forward. Retargeted to https://github.com/quratus/paperclip/pull/5 (same branch, same content).

PR #185, opened 2026-03-07
fix: invalidate project issues list when issue detail updates

ngmeyer:
## Problem

Changing an issue's assignee (or any other field) on the issue detail page does not update the project issues list. Navigating back to `/projects/{id}/issues` shows the stale assignee value.

## Root Cause

`invalidateIssue()` in `IssueDetail.tsx` invalidates `queryKeys.issues.list(companyId)` (the global list) but not `queryKeys.issues.listByProject(companyId, projectId)` — the separate query used by the project issues view.

## Fix

Add `listByProject` invalidation to `invalidateIssue()` when the issue has a `projectId`.

```ts
if (issue?.projectId) {
  queryClient.invalidateQueries({
    queryKey: queryKeys.issues.listByProject(selectedCompanyId, issue.projectId)
  });
}
```

## Repro

1. Open a project issues list (`/GEA/projects/signupspark/issues`)
2. Click an issue to open detail page
3. Change the assignee
4. Navigate back to the project issues list
5. **Before:** assignee shows as stale / empty
6. **After:** assignee shows the updated value

PR #9847, opened 2026-07-19
fix(hermes): preserve response text around session metadata

JackReis:
## Summary
- preserve complete Hermes response text before and after session metadata lines
- exclude exact session metadata from stderr error classification
- retain provider-exhaustion and usage parsing behavior

## Verification
- Hermes adapter suite: 81/81 passed across 9 files
- focused compatibility suite: 21/21 passed
- adapter TypeScript check: passed
- independent senior review: GO for commit/push integration

## Caveat
- lint was not claimed because this isolated checkout lacks the declared `eslint` executable (`spawn ENOENT`)
- this PR is integration-only; no production deployment or Paperclip agent mutation was performed

Internal fleet tracking: JAC-3706 / hermes-04ps.1.1


JackReis:
## Thinking Path

The Tasks (Issues) and Inbox pages crashed with React error #185 ("Maximum
update depth exceeded") on a default-configuration install (isolated workspaces
off, out-of-the-box default). Tracing the render loop: several `useQuery` calls
default their result to an inline `[]` literal. When a query is disabled or
still loading, `data` is `undefined`, so that default expression re-evaluates
every render and returns a brand-new array reference — even though the logical
value (empty) never changes. That unstable reference propagates through a
`useMemo`/`useEffect` dependency chain that ends in a `setState`, so the effect
never stops re-firing. The bug is about reference identity, not data shape, so
it reproduces on any fresh install without special fixture data.

## Linked Issues

Fixes #185.

## What Changed

Replaced inline `[]` `useQuery` defaults with stable module-level constant empty
arrays so the reference is preserved across renders when a query has no data yet.

- `ui/src/components/IssuesList.tsx`: stabilized `executionWorkspaces` (permanently
  disabled on default installs while `isolatedWorkspaces` is off) via
  `EMPTY_EXECUTION_WORKSPACES`, and `searchedIssues` (disabled when the search box
  is empty) via `EMPTY_ISSUES`. Chain: `executionWorkspaces → executionWorkspaceById
  → issueFilterWorkspaceContext → issueFilterContext → filtered → setRenderedIssueRowLimit`.
- `ui/src/pages/Inbox.tsx`: stabilized `executionWorkspaces`, `mineIssuesRaw`,
  `touchedIssuesRaw`, `remoteIssueSearchResults`, and `joinRequests` via
  `EMPTY_EXECUTION_WORKSPACES` / `EMPTY_ISSUES`. Chain: those defaults →
  `groupedSections → flatNavItems → setSelectedIndex`.
- Added a regression guard test asserting these query defaults keep a stable
  reference across renders, so an inline `[]` default can't be reintroduced silently.

This is a pure reference-identity fix — the value handed to consumers is unchanged
(still an empty array) whenever the query has no data.

## Verification

- `tsc -b` in `ui/` is clean.
- `ui/src/components/IssuesList.test.tsx` and `ui/src/pages/Inbox.test.tsx` pass
  unchanged (behavior-neutral for tested paths).
- New guard test passes and fails if an inline `[]` default is reintroduced.
- Cutter preview confirms `/issues` (Tasks) and `/inbox` render at zero state
  without the infinite-render crash.

## Risks

Low. Reference-identity-only change; no data-shape or behavioral change. The
only observable effect is that consumers now receive a stable empty-array
reference instead of a fresh one per render when the query has no data.

## Model 

## Model Used

- Claude Opus 4.8
- Claude Sonnet
- Qwen3 30B (local)
- got-5.6-Sol (high)
- got-5.6-Luna (high)


## Checklist

- [x] I searched the open PR list for similar/duplicate PRs before opening this one.
- [x] Includes a test covering the fix.

JackReis:
Superseded by #10445 (JAC-3706 v2).

Issue #6195, opened 2026-05-17
[CRITICAL] Patch `20260517uiux57` causes React error #185, breaks Web UI auth, and auto-archives production company


joojalre:
### Summary

After Paperclip applied patch `_pcpPatch=20260517uiux57` (auto-chained on top of `20260517-maintenance1`), the entire Web UI breaks with React minified error #185 (Maximum update depth exceeded). Simultaneously, the production company is set to `status: archived` and a new empty company with the same name (lowercase variant) is auto-created. The CLI continues to work normally — confirming this is a UI/auth layer regression introduced by the patch, not a data layer issue.

**Severity:** 🔴 Critical — Web UI unusable, production data inaccessible via dashboard
**Data integrity:** ✅ Confirmed safe — embedded PostgreSQL healthy, 18 agents heartbeating, automated DB backups running
**CLI status:** ✅ Working — `paperclipai issue list`, `company list`, `auth whoami` all functional

---

### Environment

| Property | Value |
|---|---|
| Paperclip version | latest via `npx paperclipai` (as of 2026-05-17) |
| Patch chain | `20260517-maintenance1` → `20260517uiux57` (auto-applied) |
| OS | Windows 11 |
| Node | latest LTS |
| Browser | Chrome (latest) |
| Mode | authenticated/private, loopback bind |
| Port | 3100 |
| Database | embedded PostgreSQL (port 54329) |
| Doctor checks | ✅ 9 passed |

---

### Steps to Reproduce

1. Run `npx paperclipai run --instance <name>` on an existing instance with at least one active company.
2. Open `http://localhost:3100/<PREFIX>/dashboard` in browser.
3. URL automatically becomes `?_pcpPatch=20260517-maintenance1` then `?_pcpPatch=20260517uiux57`.
4. Wait for page to load.
5. ❌ Red error banner appears: "حدث خطأ في تحميل الواجهة" (UI failed to load).
6. Open DevTools → Console → see `Minified React error #185`.
7. Run `npx paperclipai company list` from terminal.
8. ❌ Original active company is now `status: archived`.
9. ❌ A new empty company with the same name (lowercase) exists at `status: active`.

---

### Smoking Gun — Server Logs

Full sequence captured in attached `paperclip-server-log-redacted.txt`. Key evidence:

**1. Patch chaining (auto-applied):**
```
[17:27:13] GET /auth?next=...&_pcpPatch=20260517-maintenance1 200
[...]

PR #9848, opened 2026-07-19
Fix React error #185 (Maximum update depth exceeded) on Issues/Inbox with default config

stubbi:
## Thinking Path

> - The Issues/Tasks and Inbox pages are the primary daily-driver surfaces for managing that work, built on `useQuery` from `@tanstack/react-query` feeding memoized nav/filter chains
> - Both pages default several `useQuery` results to an inline `[]` literal (`const { data: foo = [] } = useQuery(...)`), which is re-evaluated to a brand-new array reference on every render whenever the query is disabled or still loading
> - On any default-config install (isolated workspaces off, the out-of-the-box setting), `executionWorkspaces` is permanently disabled and `remoteIssueSearchResults` is permanently disabled until a search is typed, so this unstable reference feeds the `filtered`/`groupedSections`/`flatNavItems` memo chains forever, which keeps re-firing the row-limit and selection effects that depend on them and can trip React error #185 ("Maximum update depth exceeded")
> - This pull request replaces the inline `[]` defaults with module-level stable constants (`EMPTY_EXECUTION_WORKSPACES`, `EMPTY_ISSUES`) in both files
> - The benefit is Issues/Tasks and Inbox no longer carry a standing crash risk on the default configuration, with no behavioral change to what's rendered

No existing public issue tracks this exact bug. Filing per the bug report template:

**What happened?** The Issues/Tasks list and Inbox pages can crash with React minified error #185 ("Maximum update depth exceeded") on a default-configuration install (isolated workspaces feature flag off, which is the out-of-the-box default).

**Expected behavior:** Issues/Tasks and Inbox render normally regardless of the isolated-workspaces setting.

**Steps to reproduce:**
1. Run Paperclip with a fresh/default instance config (isolated workspaces off).
2. Navigate to the Issues/Tasks page or the Inbox page.
3. Under load (multiple queries settling close together, e.g. shortly after sign-in with live agent activity), the page can throw React error #185 and render the error boundary.

Root cause detail is in "What Changed" below.

Refs: #6195 — a different, unrelated cause (a specific `_pcpPatch` breaking Web UI auth) that happened to also surface as React error #185; noting it here since it turned up in the duplicate search, not because it's the same bug.

## What Changed

- `ui/src/components/IssuesList.tsx`: added a module-level `EMPTY_EXECUTION_WORKSPACES: ExecutionWorkspaceSummary[] = []` constant and used it as the `useQuery` default for `executionWorkspaces` instead of an inline `[]` literal.
- `ui/src/pages/Inbox.tsx`: added module-level `EMPTY_EXECUTION_WORKSPACES` and `EMPTY_ISSUES: Issue[] = []` constants and used them as the `useQuery` defaults for `executionWorkspaces`, `mineIssuesRaw`, `touchedIssuesRaw`, and `remoteIssueSearchResults` instead of inline `[]` literals.
- `ui/src/pages/stable-issue-query-defaults.regression.test.ts` (new): a source-level regression guard — see "Verification" for why it's source-level rather than a full render/crash repro.

This is a pure reference-identity fix. Every consumer still receives an empty array when the query has no data; only the object identity is now stable across renders instead of being reallocated every time.

## Verification

- `pnpm --filter @paperclipai/ui exec tsc -b` — clean.
- `pnpm --filter @paperclipai/ui exec vitest run src/pages/Inbox.test.tsx src/components/IssuesList.test.tsx src/pages/stable-issue-query-defaults.regression.test.ts` — 52/52 pass.
- I confirmed the existing `IssuesList.test.tsx`/`Inbox.test.tsx` suites are behavior-neutral for this change (identical pass/fail results before and after).
- I attempted a full render-based regression test (render Inbox with the flag off, assert no "Maximum update depth exceeded"). I could not make it discriminate between the buggy and fixed code in this project's jsdom/vitest harness: I instrumented both a `<Profiler>` commit counter and a call-count/identity spy on `buildInboxKeyboardNavEntries` (the function that consumes `groupedSections`, the direct input to the crashing effect), covering both a "mount with several queries permanently disabled" scenario and a "force an unrelated re-render after mount" scenario. In every case, the patched and unpatched code produced identical commit counts / call counts / referential-identity results. The real crash needs several queries to settle within the same synchronous React update batch — a timing condition that doesn't occur with vitest's mocked queries, which resolve deterministically inside a single `act()` flush rather than racing like real network/WebSocket traffic. Given that, `stable-issue-query-defaults.regression.test.ts` instead asserts the fix at the source level (no raw `= []` default for any of the five affected query results, and the named stable constants are declared and used). I verified this test fails against the pre-fix source and passes against the fix, so it will catch a regression if the pattern is reintroduced.

## Risks

Low risk. The change only swaps an inline empty-array literal for a module-level constant empty array used in the exact same position (a `useQuery` destructure default); the value observed by every consumer is unchanged (still an empty array whenever the query has no data). No migration, no API, no behavioral branch touched.

## Model Used

Claude (Anthropic), model id `claude-sonnet-5` (Sonnet 5), standard reasoning mode, agentic tool use (file read/edit, local shell for install/build/test, `gh` CLI for repo search and PR management). No extended/chain-of-thought mode exposed to the harness beyond the model's default reasoning.

## Checklist

### DIFF
--- a/ui/src/components/IssuesList.tsx
+++ b/ui/src/components/IssuesList.tsx
@@ -85,12 +85,17 @@ import { isSuccessfulRunHandoffRequired } from "../lib/successful-run-handoff";
 import { deriveOriginatingActor, ISSUE_STATUSES, type ExecutionWorkspaceSummary, type Issue, type IssueStatus, type Project } from "@paperclipai/shared";
 import { Badge } from "@/components/ui/badge";
 
-// Stable empty-array default: a disabled/loading `executionWorkspaces` query
-// returns undefined, and a fresh `[]` literal default would give it a new
-// identity every render, churning executionWorkspaceById -> issueFilterContext
-// -> `filtered` and driving the row-limit effect into an infinite render loop
-// (React error #185). A module-level constant keeps the reference stable.
+// Stable empty-array defaults. A disabled/loading query returns `data ===
+// undefined`, and a fresh `[]` literal default would give it a new identity
+// every render, churning executionWorkspaceById -> issueFilterContext ->
+// `filtered` and driving the row-limit effect into an infinite render loop
+// (React error #185). Module-level constants keep the references stable.
+// `executionWorkspaces` is disabled whenever isolated workspaces are off
+// (the default), so it is permanently unstable without this. `searchedIssues`
+// is disabled whenever there is no active search term, which is also the
+// common default state.
 const EMPTY_EXECUTION_WORKSPACES: ExecutionWorkspaceSummary[] = [];
+const EMPTY_SEARCHED_ISSUES: Issue[] = [];
 const ISSUE_SEARCH_DEBOUNCE_MS = 250;
 const ISSUE_SEARCH_RESULT_LIMIT = 200;
 const ISSUE_BOARD_COLUMN_RESULT_LIMIT = 200;
@@ -772,7 +777,7 @@ export function IssuesList({
   // eslint-disable-next-line react-hooks/exhaustive-deps
   }, [issues]);
 
-  const { data: searchedIssues = [] } = useQuery({
+  const { data: searchedIssues = EMPTY_SEARCHED_ISSUES } = useQuery({
     queryKey: [
       ...queryKeys.issues.search(selectedCompanyId!, normalizedIssueSearch, projectId),
       searchFilters ?? {},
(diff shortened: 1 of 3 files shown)


stubbi:
Rebased onto current master `d293dd3d1` (new head `9455ac8ff`). Single conflict, in the `IssuesList.tsx` import block where master added task-collection/date-group imports; master's imports are kept and `ExecutionWorkspaceSummary` folded into the existing `@paperclipai/shared` type import. Still needed — master's `IssuesList.tsx` (lines 872, 917) and `Inbox.tsx` (873, 914, 967, 994, 1082) all still use raw `= []` useQuery defaults; note master independently adopted the same stable-constant pattern in `ui/src/pages/IssueDetail.tsx` (`EMPTY_ISSUES`), which is the same fix applied to a third page. Verified: `@paperclipai/ui` typecheck clean (0 errors); `ui/src/pages/stable-issue-query-defaults.regression.test.ts` 2/2 and `IssuesList.test.tsx` + `Inbox.test.tsx` 73/73.

PR #9870, opened 2026-07-19
fix: route nested Codex bridge and guard one-shot wakes

broken-coder:
## Summary

- Route nested Codex `curl` requests through the existing `queue_v1` filesystem bridge, preserving the sandbox network boundary.
- Add an issue-level one-shot wake guard: the initial `assignment / issue_assigned` wake is reserved, and later automation/comment wakes are auditable skips.
- Document the queue-native Bash `curl` path and add focused bridge and wake regression coverage.

## Root Cause

Nested Codex execution runs without access to the E2B sandbox loopback bridge. Separately, a watcher comment could schedule a second wake after an exactly-one assignment run.

## Validation

- `pnpm exec vitest run packages/adapter-utils/src/execution-target-sandbox.test.ts` (27 passed)
- `pnpm exec vitest run server/src/__tests__/heartbeat-dependency-scheduling.test.ts --no-file-parallelism --maxWorkers=1` (7 passed)
- `pnpm --filter @paperclipai/shared typecheck`

## Scope

This is a draft PR for HOM-124 only. It excludes the unrelated HOM-85 local commits, runtime deployment, agent activation, and live smoke execution.


packages/adapter-utils/src/sandbox-callback-bridge.ts
```diff
@@ -1018,6 +1067,111 @@ export async function startSandboxCallbackBridgeServer(input: {
...
+
+let target;
+let bridgeOrigin;
+try {
+  target = new URL(url);
+  bridgeOrigin = new URL(apiUrl).origin;
+} catch {
+  fail("Queue bridge curl received an invalid Paperclip API URL.");
+}
+if (target.origin !== bridgeOrigin) {
```
invexta, line 1133: **Origin check is stricter than it needs to be, and silently fails closed.** `target.origin !== bridgeOrigin` means a script that legitimately hits `http://127.0.0.1:NNNN/...` will be rejected when the host URL is `http://localhost:NNNN/...` (or vice-versa, or `127.0.0.1` vs `[::1]`). The bridge server itself doesn't care about the origin — it forwards whatever path/query the worker hands it. Suggestion: compare hosts case-insensitively *and* resolve `localhost` aliases, or drop the check entirely and rely on the bridge token as the security boundary. Right now this will produce confusing `only supports the configured Paperclip API origin` failures for benign cases.

packages/adapter-utils/src/sandbox-callback-bridge.ts
```diff
@@ -1018,6 +1067,111 @@ export async function startSandboxCallbackBridgeServer(input: {
...
+}
+
+const requestId = randomUUID();
+const requestsDir = path.posix.join(queueDir, "requests");
+const responsesDir = path.posix.join(queueDir, "responses");
+const requestPath = path.posix.join(requestsDir, requestId + ".json");
+const responsePath = path.posix.join(responsesDir, requestId + ".json");
+await fs.mkdir(requestsDir, { recursive: true });
+await fs.mkdir(responsesDir, { recursive: true });
+await fs.writeFile(requestPath + ".partial", JSON.stringify({
```
invexta, line 1147: **Partial-then-rename is good, but no cleanup if the wait loop throws or the request is orphaned.** The host worker is responsible for writing the response file, but if the curl shim hits the timeout branch the `requestPath` is removed (good), yet if the process is SIGKILLed between `fs.rename` and the read loop, the queued request leaks. Worth either (a) registering a `process.on('exit', ...)` / `SIGTERM` handler that removes `requestPath`, or (b) accepting that the host worker already has a TTL/cleanup for stale requests and noting that in a comment. The latter is probably fine — just call it out.

server/src/services/heartbeat.ts
```diff
@@ -15435,6 +15436,70 @@ export function heartbeatService(db: Db, options: HeartbeatServiceOptions = {})
...
+                eq(agentWakeupRequests.companyId, agent.companyId),
+                eq(agentWakeupRequests.source, "assignment"),
+                eq(agentWakeupRequests.reason, "issue_assigned"),
+                sql`${agentWakeupRequests.payload} ->> 'issueId' = ${issue.id}`,
+              ),
+            )
+            .limit(1)
+            .then((rows) => rows[0] ?? null);
+
+          if (!initialAssignmentWake || priorInitialAssignmentWake) {
```
invexta, line 15456: **One-shot guard reads `enabled === true`, but the policy may have been written with a falsy `enabled: false`.** `oneShotPolicy.enabled === true` is correct, but the suppression path also doesn't verify `policy.oneShot` is actually well-formed before consulting `enabled`. More importantly: the guard fires for *every* later wake regardless of reason — `issue_commented`, manual `wakeup()`, retry storms, dependency-scheduled wakes — all silently suppressed. That's almost certainly what you want for "exactly one", but consider documenting it on `IssueExecutionOneShotPolicy` so a future caller doesn't expect `enabled: true` to mean "one per wake-source". Optional nit: the suppression reason string `"only the initial issue_assigned wake is allowed"` doesn't distinguish between "wrong reason" and "wrong source"; the activity log could carry both.

invexta (commented):
## Hermes Agent — Code Review

**Verdict: Approve with suggestions** (3 inline notes, nothing blocking — appropriate for a draft PR.)

Solid, well-scoped change. The two-part split (queue-aware `curl` shim for nested Codex + one-shot wake reservation) is clean, the SKILL.md update mirrors the env wiring, and both new test files exercise the real failure modes (loopback unreachable for the bridge, second comment wake after assignment for the one-shot guard). 

A few things worth tightening before this leaves draft.

PR #9883, opened 2026-07-20
fix(projects): allow clearing env bindings

### LINKED ISSUE #9825: Project Configuration save always fails with 422 when env is emptied (UI sends env: null, PATCH /projects/:id rejects it)
## Summary

Saving a project's Configuration tab fails with a red "Failed" badge next to **Env** whenever the environment-variables editor is dirty but empty (e.g. you add a variable row and then blank or remove it). The UI sends `PATCH /api/projects/:id` with `{"env": null}` and the server rejects it with **422 Unprocessable Entity**, so the save can never succeed.

## Repro

1. Open any project → Configuration tab.
2. Under **Env**, click **Add variable**, then delete the row (or leave it blank) so the editor shows "No environment variables" with the "Unsaved changes" bar.
3. Click **Save**.
4. Red "Failed" appears next to the Env label. Server log shows:

```
WARN: PATCH /api/projects/<id>?companyId=<companyId> 422 {"reqBody":{"env":null},...}
```

## Root cause

Three layers disagree about what an empty env means:

1. **UI** — the env editor intentionally emits `undefined` when it has no valid rows (`valueFromRows([])` → `undefined`, see `ui/src/components/environment-variables-editor/model.test.ts`), and `ProjectProperties` converts that to `null` before committing:

   ```tsx
   // ui/src/components/ProjectProperties.tsx
   onChange={(env) => commitField("env", { env: env ?? null })}
   ```

2. **Validation schema** — `updateProjectSchema` explicitly allows null, so the request passes validation:

   ```ts
   // packages/shared/src/validators/project.ts
   env: envConfigSchema.optional().nullable(),
   ```

3. **Route handler** — but the PATCH (and POST) handlers in `server/src/routes/projects.ts` only skip normalization for `undefined`, so `null` falls through into the normalizer, which throws:

   ```ts
   if (body.env !== undefined) {
     body.env = await secretsSvc.normalizeEnvBindingsForPersistence(existing.companyId, body.env, ...);
   }
   ```

   ```ts
   // server/src/services/secrets.ts — normalizeEnvConfig
   const record = asRecord(envValue);
   if (!record) throw unprocessable(`${opts?.fieldPath ?? "env"} must be an object`);
   ```

So the schema declares `env: null` legal while the handler makes it always-422 — an internal contract mismatch.

Notably, **routines and pipelines already handle this correctly** by treating `null` as "clear env" and skipping normalization (`server/src/services/routines.ts`, `server/src/services/pipelines.ts`):

```ts
const normalizedEnv = input.env === null
  ? null
  : await secretsSvc.normalizeEnvBindingsForPersistence(...);
```

The projects route is the outlier.

## Suggested fix

Align the project routes with the routines/pipelines pattern — in both the create and update handlers in `server/src/routes/projects.ts`:

```ts
if (body.env !== undefined && body.env !== null) {
  body.env = await secretsSvc.normalizeEnvBindingsForPersistence(...);
}
```

`syncEnvBindingsForTarget` already handles `null` safely (`asRecord(envValue) ?? {}` → deletes stale bindings, inserts none), so passing `null` through to persistence also correctly clears any previously-bound secret refs.

I've verified this fix locally (v2026.707.0): with the guard in place, `PATCH {"env": null}` returns 200, clears the env, and removes stale bindings. Happy to open a PR if useful — a regression test slots neatly into `server/src/__tests__/project-routes-env.test.ts`.

## Environment

- Paperclip v2026.707.0, native macOS install, local adapter, PostgreSQL backend.

MindSyncHub:
## Thinking Path

> - Paperclip is a company-scoped control plane, so project settings must stay consistent with company-level env binding rules.
> - The project configuration UI already sends `env: null` when a user clears the env editor.
> - The shared validator allows `env` to be null, so the route should treat null as a valid clear action.
> - The current project route still normalizes null as if it were a binding object, which turns a clear action into a 422.
> - This pull request aligns project create/update behavior with the existing routines/pipelines pattern for clearing env.
> - The benefit is that users can clear project env values without hitting an error, while secret bindings are cleaned up correctly.

Fixes: #9825

## What Changed

- Skip env normalization when project `env` is `null` on create and update.
- Add a regression test that covers `PATCH /api/projects/:id` with `env: null`.
- Verify null clears bindings without logging raw secret values.

## Verification

- `PATH="/Users/xinran/.cache/codex-runtimes/codex-primary-runtime/dependencies/bin:$PATH" pnpm --filter server exec vitest run src/__tests__/project-routes-env.test.ts`
- `git diff --check`

## Risks

- Low risk: behavior change is limited to the explicit env-clear path.
- If callers relied on the old 422, they will now see successful clears instead.

## Model Used

- OpenAI Codex (GPT-5), tool-using coding agent in the Codex desktop environment.

## Checklist

### DIFF
--- a/server/src/__tests__/project-routes-env.test.ts
+++ b/server/src/__tests__/project-routes-env.test.ts
@@ -206,6 +206,33 @@ describe("project env routes", () => {
     );
   });
 
+  it("allows create env null without normalizing project env bindings", async () => {
+    mockProjectService.create.mockResolvedValue(buildProject({ env: null }));
+
+    const app = await createApp();
+    const res = await request(app)
+      .post("/api/companies/company-1/projects")
+      .send({
+        name: "Project",
+        env: null,
+      });
+
+    expect([200, 201], JSON.stringify(res.body)).toContain(res.status);
+    expect(mockSecretService.normalizeEnvBindingsForPersistence).not.toHaveBeenCalled();
+    expect(mockProjectService.create).toHaveBeenCalledWith(
+      "company-1",
+      expect.objectContaining({ env: null }),
+    );
+    expect(mockLogActivity).toHaveBeenCalledWith(
+      expect.anything(),
+      expect.objectContaining({
+        details: expect.objectContaining({
+          envKeys: [],
+        }),
+      }),
+    );
+  });
+
   it("normalizes env bindings on update and avoids logging raw values", async () => {
     const normalizedEnv = {
       PLAIN_KEY: { type: "plain", value: "top-secret" },
--- a/server/src/__tests__/project-routes-env.test.ts
+++ b/server/src/__tests__/project-routes-env.test.ts
@@ -16,6 +16,7 @@ const mockProjectService = vi.hoisted(() => ({
 }));
 const mockSecretService = vi.hoisted(() => ({
   normalizeEnvBindingsForPersistence: vi.fn(),
+  syncEnvBindingsForTarget: vi.fn(),
 }));
 const mockEnvironmentService = vi.hoisted(() => ({
   getById: vi.fn(),
@@ -163,6 +164,7 @@ describe("project env routes", () => {
     mockProjectService.listWorkspaces.mockResolvedValue([]);
     mockEnvironmentService.getById.mockReset();
     mockSecretService.normalizeEnvBindingsForPersistence.mockImplementation(async (_companyId, env) => env);
+    mockSecretService.syncEnvBindingsForTarget.mockResolvedValue(undefined);
   });
 
   it("normalizes env bindings on create and logs only env keys", async () => {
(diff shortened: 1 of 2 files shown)


MindSyncHub:
Addressed the Greptile test coverage gap by adding a create-path regression test for `POST /api/companies/:companyId/projects` with `env: null`. The test verifies we skip env binding normalization and pass `env: null` through to project creation.\n\nVerification:\n- `pnpm vitest run src/__tests__/project-routes-env.test.ts --config vitest.config.ts` from `server/` (4 tests passed)

MindSyncHub:
已解决与最新 `master` 的冲突并完成 rebase（变基）：保留受管沙盒工作区路径校验，同时在 `env: null` 时跳过环境绑定规范化。聚焦验证：`vitest run src/__tests__/project-routes-env.test.ts --config vitest.config.ts`（4 tests passed）。现等待新提交的 CI 与 Greptile 复检。

PR #9900, opened 2026-07-20
fix(onboarding): do not restore wizard state for a company the user does not own

stubbi:
## Thinking Path

> - The onboarding wizard is the first-run flow that creates a company, hires the first agent, and connects a model, and it is designed to survive a page reload mid-flow by persisting its draft to `localStorage`.
> - `localStorage` is scoped per browser origin, not per signed-in account, so any state written there is visible to every session that later loads the same origin.
> - The wizard's restore path never checked that the `createdCompanyId` in the saved draft belonged to the account currently signed in, so a second onboarding session in the same browser (a second account, or the same account starting a second company) silently inherited a previous company's id.
> - Every wizard action taken after that point (create goal, hire agent, create issue) then targets a company the current session does not own, which fails authorization server-side or, worse, could write into a company the user has no business touching.
> - This pull request turns "restore the saved draft" into an authorization decision: the saved company id is checked against the signed-in user's own companies, and anything that fails the check is discarded and wiped from storage.
> - It also fixes a related timing bug uncovered while writing the fix: the wizard's ~20 `useState(saved?.x ?? default)` initializers only read `saved` on the very first render, so the authorization check has to run before first mount, not after, or every restored field silently locks to its default forever even once the company list resolves.
> - The benefit is that a shared or previously-used browser can no longer let one onboarding session read or act on another account's in-progress company draft.

No existing public issue or PR covers this (searched `gh pr list` / `gh issue list` for `onboarding localStorage company`, `onboarding wizard restore`, and related terms; found only unrelated onboarding PRs such as #9501 and #2616). Describing the issue in-PR following the bug report template:

**What happened?**
`ui/src/components/OnboardingWizard.tsx` persists its draft state, including `createdCompanyId`, to `localStorage` under the fixed key `paperclip-onboarding-state` on every field change, and restores it on mount with no check that the current signed-in user owns that company.

**Expected behavior**
A saved draft should only be restored when the signed-in user actually owns the company it references. A draft for a company the user does not own should be discarded, not silently applied.

**Steps to reproduce**
1. Sign in as user A, start onboarding, and get far enough that `createdCompanyId` is set (survives past company creation).
2. Sign out (or, on a shared/instance-local browser, sign in as user B) without ever finishing or explicitly resetting the wizard.
3. Open onboarding again in the same browser.
4. The wizard restores mid-flow with user A's `createdCompanyId` already populated, and subsequent wizard actions (hiring an agent, creating a goal, etc.) target that company on behalf of the new session.

**Paperclip version or commit**: reproducible on `master` at `f12bb27bc`.

## What Changed

- Added `ui/src/lib/onboarding-state.ts` exporting `restoreOnboardingState(raw, companies)`. It treats restoring the saved blob as an authorization decision: if the blob names a `createdCompanyId` that is not in the caller's settled companies list, the whole blob is discarded (`null`); a draft with no company id yet (early steps) passes through unchanged. The function's contract requires the caller's companies list to be settled (`companiesLoading === false`), since an empty list is ambiguous between "still loading" and "this account owns nothing."
- Split `OnboardingWizard.tsx` into a thin outer gate (`OnboardingWizard`) and `OnboardingWizardInner`. The inner component's `useState(saved?.x ?? default)` initializers only read `saved` on first render, so the gate now defers mounting the inner component until the companies list has settled whenever a saved blob exists in storage, guaranteeing the inner component's very first render sees the final, authorized value.
- A discarded or malformed blob is wiped from `localStorage` immediately via a cleanup effect, so it cannot confuse the next onboarding attempt in the same browser.
- Exported `ONBOARDING_STORAGE_KEY` from `OnboardingWizard.tsx` (previously module-private) so the test imports the real key instead of duplicating the string literal.
- Added `ui/src/lib/onboarding-state.test.ts` (unowned/owned/no-company-yet/empty-settled-list/malformed-input cases) and `ui/src/components/OnboardingWizard.test.tsx` (new file; upstream had no direct test coverage of this component before this PR) covering the async-companies-arrival timing regression and the cross-account stale-draft discard-and-wipe path.

## Verification

- `npx vitest run` from `ui/`, full suite: 3009 passed. 3 pre-existing failures are unrelated to this change and reproduce on an unmodified `master` checkout (`CaseFieldsPanel.test.tsx`, `IssueProperties.test.tsx`, `attention.test.ts`; locale/timezone-dependent formatting assertions in files this PR does not touch).
- Targeted: `npx vitest run src/lib/onboarding-state.test.ts src/components/OnboardingWizard.test.tsx src/components/OnboardingWizardVariant.test.tsx` (8/8 pass).
- `pnpm --filter @paperclipai/ui typecheck` (`tsc -b`), clean, exit 0.

## Risks

- Low risk. The change only narrows what gets restored from `localStorage`; it never widens access. Worst case for a legitimate single-account user is an unnecessary reset to step 0 if `restoreOnboardingState` were ever too strict, which the added test suite guards against (owned-company and no-company-yet drafts are proven to survive).
- The gate renders `null` for one or more frames while a saved blob exists and `companiesLoading` is still `true`, instead of mounting the wizard immediately. This is a brief, existing loading window (the company list was already being fetched before this change); the wizard was not usable during that window before either, since the old code raced the same async data with synchronous `useState` initializers and could show visibly wrong (stale-default) fields. No user-visible regression expected, but reviewers should note the wizard's very first paint can now be delayed until companies resolve when localStorage already has a draft.
- No data migration, no API or schema changes, no changes to the persisted wizard state shape.

## Model Used

Claude (Anthropic), via Claude Code CLI. Agentic tool use (file search, edit, test execution, and PR authoring tools) with extended reasoning over multiple turns; no single fixed model ID is asserted here beyond "Claude (Anthropic)" per the requesting instruction.

## Checklist

- [x] I have updated relevant documentation to reflect my changes (no user-facing docs are affected by this internal bug fix)

Co-Authored-By: Claude Fable 5 <noreply@anthropic.com>

### DIFF
--- a/ui/src/components/OnboardingWizard.test.tsx
+++ b/ui/src/components/OnboardingWizard.test.tsx
@@ -7,8 +7,6 @@ import { afterEach, beforeEach, describe, expect, it, vi } from "vitest";
 
 // --- Mocks (hoisted so vi.mock factories can close over them) ----------------
 
-const ONBOARDING_STORAGE_KEY = "paperclip-onboarding-state";
-
 const mockDialog = vi.hoisted(() => ({
   onboardingOpen: true,
   onboardingOptions: {} as { initialStep?: number; companyId?: string },
@@ -110,7 +108,7 @@ vi.mock("./AsciiArtAnimation", () => ({ AsciiArtAnimation: () => null }));
 vi.mock("./FrontDoor", () => ({ FrontDoor: () => null }));
 vi.mock("./AgentCapsule", () => ({ AgentCapsule: () => null }));
 
-import { OnboardingWizard } from "./OnboardingWizard";
+import { ONBOARDING_STORAGE_KEY, OnboardingWizard } from "./OnboardingWizard";
 
 // eslint-disable-next-line @typescript-eslint/no-explicit-any
 (globalThis as any).IS_REACT_ACT_ENVIRONMENT = true;
--- a/ui/src/components/OnboardingWizard.test.tsx
+++ b/ui/src/components/OnboardingWizard.test.tsx
@@ -1,0 +1,262 @@
+// @vitest-environment jsdom
+
+import { act } from "react";
+import { createRoot } from "react-dom/client";
+import { QueryClient, QueryClientProvider } from "@tanstack/react-query";
+import { afterEach, beforeEach, describe, expect, it, vi } from "vitest";
+
+// --- Mocks (hoisted so vi.mock factories can close over them) ----------------
+
+const ONBOARDING_STORAGE_KEY = "paperclip-onboarding-state";
+
+const mockDialog = vi.hoisted(() => ({
+  onboardingOpen: true,
+  onboardingOptions: {} as { initialStep?: number; companyId?: string },
+  closeOnboarding: vi.fn(),
+  onboardingRouteDismissed: false,
+  setOnboardingRouteDismissed: vi.fn(),
+}));
+
+const mockCompany = vi.hoisted(() => ({
+  companies: [] as Array<{ id: string; name: string; issuePrefix: string }>,
+  setSelectedCompanyId: vi.fn(),
+  loading: false,
+}));
+
+const mockCompaniesApi = vi.hoisted(() => ({
+  create: vi.fn(),
+  update: vi.fn(),
+}));
+const mockGoalsApi = vi.hoisted(() => ({
+  create: vi.fn(),
+  list: vi.fn(async () => []),
+}));
+const mockAgentsApi = vi.hoisted(() => ({
+  adapterModels: vi.fn(async () => [] as Array<{ id: string; label: string }>),
+  testEnvironment: vi.fn(async () => ({
+    adapterType: "claude_local",
+    status: "pass" as const,
+    checks: [],
+    testedAt: new Date().toISOString(),
+  })),
+  hire: vi.fn(async () => ({ agent: { id: "agent-1" }, approval: null })),
+  instructionsBundle: vi.fn(async () => ({ entryFile: "AGENTS.md" })),
+  saveInstructionsFile: vi.fn(async () => ({})),
+}));
+const mockApprovalsApi = vi.hoisted(() => ({
+  create: vi.fn(),
+}));
+const mockIssuesApi = vi.hoisted(() => ({
+  create: vi.fn(),
+}));
+const mockProjectsApi = vi.hoisted(() => ({
+  create: vi.fn(),
+  list: vi.fn(async () => []),
+}));
+
+// The real adapter registry eagerly imports every adapter package. The
+// model/harness picker internals are out of scope here, so stub the adapter
+// layer entirely and drive it through this knob.
+const mockAdapterRegistry = vi.hoisted(() => ({
+  list: [] as Array<{ type: string }>,
+  disabled: new Set<string>(),
+}));
+
+vi.mock("@/lib/router", () => ({
+  useLocation: () => ({ pathname: "/", search: "", hash: "", state: null }),
+  useNavigate: () => vi.fn(),
+  useParams: () => ({}),
+}));
+vi.mock("../context/DialogContext", () => ({
+  useDialog: () => mockDialog,
+}));
+vi.mock("../context/CompanyContext", () => ({
+  useCompany: () => mockCompany,
+}));
+vi.mock("../api/companies", () => ({ companiesApi: mockCompaniesApi }));
+vi.mock("../api/goals", () => ({ goalsApi: mockGoalsApi }));
+vi.mock("../api/agents", () => ({ agentsApi: mockAgentsApi }));
+vi.mock("../api/approvals", () => ({ approvalsApi: mockApprovalsApi }));
+vi.mock("../api/issues", () => ({ issuesApi: mockIssuesApi }));
+vi.mock("../api/projects", () => ({ projectsApi: mockProjectsApi }));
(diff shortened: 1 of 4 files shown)


tonio-alucema:
@stubbi — this branch had gone stale against `master`, so I rebased it for you rather than leave it conflicting. It's pushed to this repo as [`rebase/9900-onto-master`](https://github.com/paperclipai/paperclip/tree/rebase/9900-onto-master) (I couldn't push to your fork — it's org-owned, so "allow edits from maintainers" doesn't actually grant access).

Your two commits keep your authorship. I put my one change in a separate commit on top so yours stay as you wrote them.

**Two conflicts, both cosmetic:**

1. `OnboardingWizard.tsx` — your `export` on `ONBOARDING_STORAGE_KEY` sat in a hunk whose surrounding context (`DEFAULT_TASK_TITLE` / `DEFAULT_TASK_DESCRIPTION`) `master` has since rewritten. Kept master's copy, took your `export`.
2. `OnboardingWizard.test.tsx` — the test asserted on `"Create your team lead"`, which #11101 renamed to `"Create your first agent"`. Only the string moved: your gate reaches step 3 with the restored draft either way, and the test's other two assertions (the saved agent name in the input, `aria-label="Step 3"`) already passed unchanged.

**Heads up on what moved underneath you.** #11352 landed earlier today and reworked a lot of `OnboardingWizard.tsx` — the step is now applied once on the open transition rather than on every change to `initialStep`, and there's new company-scoped state handling. Your gate composes with it cleanly: everything that changed lives in what your patch turns into `OnboardingWizardInner`, so it moved across intact. I checked the two ways that could have gone wrong and neither bites — `companiesLoading` is React Query's first-load-only `isLoading`, so your gate can't unmount the wizard mid-flow on a background refetch, and the new dashboard auto-open only fires after companies have settled, so it can't hit the gate's `null` window either.

Verified on the rebased branch: `ui` typecheck clean, your `OnboardingWizard.test.tsx` and `onboarding-state.test.ts` pass, full `ui` suite 3939 pass (the one failure, `IssueProperties`, is timezone-dependent and present on `master`).

**Where'd you like to take it?** Happy either way:

- You pull `rebase/9900-onto-master` into your fork and force-push, and this PR updates in place; or
- Say the word and we'll open a PR from that branch and land it for you, with your commits and authorship intact.

Worth saying plainly: nobody had reviewed this on its merits yet — the only comments here were bots — so that's still outstanding whichever route it takes. The underlying point is a good catch: `localStorage` is per-origin rather than per-account, so a shared browser really can hand one account's company id to a different session.

tonio-alucema:
@stubbi — this landed on `master` as [`66515582e`](https://github.com/paperclipai/paperclip/commit/66515582e4f48f420f4dd432d8c635432f5364a6), via #11370, with your two commits and authorship intact.

Landed rather than left waiting, since the branch had gone stale and this repo can't push to your fork. Your restore gate is unchanged in substance. What was added on top, all from review on #11370:

- **Ownership is judged only against a company list that answered.** Any query error makes it undecidable. The companies cache isn't account-scoped and survives sign-out, so a failed refetch after an account switch can leave the *previous* account's companies in hand — and trusting a non-empty list there would have handed one account's draft to the next, which is the leak your PR closes.
- **The gate withholds the wizard only while loading, never on error.** `companiesListQueryOptions` sets `retry: false`, and with no companies the dashboard's "Get Started" button opens onboarding — blocking on error made that button dead. Mounting costs the draft nothing, because the persist effect is gated on the wizard being open.
- **All four draft-storage sites are guarded.** `getItem` was outside the try, and `removeItem` in the cleanup, `setItem` in the persist effect and `reset()` were unguarded. They throw where a browser denies storage, and each sits in a render, an effect or a close handler.

Also filed a follow-up for the root cause behind the first one: sign-out should remove account-scoped cached data rather than invalidating two keys. Your gate is defensive and protects onboarding; other `useCompany()` consumers still see the stale list.

Closing this as landed. Thanks for the fix — `localStorage` being per-origin rather than per-account is a genuinely easy thing to miss, and the consequences were real.

tonio-alucema:
@stubbi — correcting the record on this one.

When I rebased this branch I told you your commits would keep your authorship. They didn't, and that's my error. The repo is squash-only, so your two commits collapsed into one; keeping your name on it needed a `Co-Authored-By` trailer in the squash message, and I forgot to add one. I did remember on a later PR that carried your work, which makes the omission here harder to excuse rather than easier.

Master history can't be rewritten, so this comment is the durable record instead.

**Your work landed in [`66515582e`](https://github.com/paperclipai/paperclip/commit/66515582e) via #11370.** What's yours in it:

- `restoreOnboardingState` and its unit coverage — the ownership check itself
- The split of the wizard into a gate and `OnboardingWizardInner`, and the reasoning for it: the inner component's ~20 `useState(saved?.x ?? default)` initializers only read the draft on first render, so mounting before the company list settles locks every field to defaults with no way back
- Exporting `ONBOARDING_STORAGE_KEY` rather than re-declaring it in tests

The bug you found — `localStorage` being per-origin rather than per-account, so a shared browser hands one account's company id to the next session — turned out to be one door into a wider fault. Sign-out wasn't clearing account-scoped caches at all, which #11380 later fixed. Your gate is still in the tree and deliberately independent of that fix, because an account can change without the sign-out button: a session lapsing server-side, a second account in a warm tab.

Two commits of mine sit on top: a test-string rename that #11101 had caused, and later fixes to the gate for a failed company query and for storage access being denied. The diagnosis was yours.

Sorry for the attribution miss.

Issue #9671, opened 2026-07-16
Built-in bundle routines set responsibleUserId to the literal "built-in-bundles", locking descendants out of mutations (RESPONSIBLE_USER_UNAVAILABLE)


werserk:
## Summary

`createOrResetRoutine` in `server/src/services/built-in-agents.ts` uses a synthetic actor:

```ts
const actor = { agentId: null, userId: "built-in-bundles" };
```

That sentinel string ends up persisted as the routine's `responsible_user_id`. Every heartbeat run descending from that routine's dispatches (the routine execution itself, its `finish_successful_run_handoff` corrective wake, and `source_scoped_recovery_action` recovery-owner runs) inherits `responsibleUserId = "built-in-bundles"`.

Under responsible-user authorization, "built-in-bundles" resolves to no user and no membership, so **every issue mutation from those runs is denied with HTTP 403 `RESPONSIBLE_USER_UNAVAILABLE`** — including the recovery owner's attempt to record a disposition. The result is a self-sustaining loop:

1. Built-in routine fires → issue created, agent run succeeds.
2. Agent (or later the recovery owner) tries `PATCH /api/issues/:id` → 403 `RESPONSIBLE_USER_UNAVAILABLE`.
3. Issue lands in `missing_disposition` recovery ("blocked on a recovery owner"), whose owner's runs inherit the same sentinel → they can't record a disposition either.
4. The subtree stays blocked forever; only a human board actor can dispose of it.

## Observed

Live instance at master `9923894` (2026-07-16), external PostgreSQL, authenticated mode. Agent run transcripts report e.g.:

```
PATCH /api/issues/<id> payload {"status":"blocked"} → 403 RESPONSIBLE_USER_UNAVAILABLE
```

`heartbeat_runs.responsible_user_id = 'built-in-bundles'` for all runs in the affected chain; `routines.responsible_user_id = 'built-in-bundles'` for the auto-ensured built-in bundle routine.

There is also no API-level remedy: `createRoutineSchema` / `updateRoutineSchema` do not expose `responsibleUserId`, so an operator cannot repair the routine without direct DB access.

## Suggested fix

Either persist `null` (system) instead of the sentinel for built-in bundle routines and let the authorization layer treat system-origin runs explicitly, or resolve a real owner (e.g. the company owner) when reconciling built-in bundles. Exposing `responsibleUserId` on the routine update API would also give operators a recovery path.

jamipouchi:
[...]

PR #9902, opened 2026-07-20
fix(routines): stop the built-in bundle system marker from becoming a responsible user

greegorij:
## Thinking Path

> - Built-in agents ship with routines, and `createOrResetRoutine` creates or reconciles those routines on behalf of the platform rather than a person
> - It passes `userId: "built-in-bundles"` — a system marker, not a real account — and that value lands in the routine's `responsibleUserId`
> - From there it propagates: `resolveRoutineResponsibleUserId` returns the actor id unconditionally, `issues.create` trusts it, and every issue the routine creates inherits a responsible user who does not exist
> - `assertCompanyAccess` then looks for an active company membership for that "user", finds none, and denies `company_access` — so the agent assigned to such an issue cannot read it, cannot list its colleagues, and cannot even post an issue reporting the failure
> - The denial is company-wide and silent: the agent has no channel left to raise the alarm, so the breakage only shows up in the service log
> - This pull request stops the marker from becoming a responsible user and backfills the rows that already carry it
> - The benefit is that routine-created issues are actionable by the agents assigned to them

No existing public issue covers this, so per option (B), following the bug-report shape:

**What happens.** An agent assigned to an issue created by a built-in-agent routine receives `403 RESPONSIBLE_USER_UNAVAILABLE` on reading the issue, listing agents, and creating new issues.

**Why.** The issue's `responsibleUserId` is the literal string `built-in-bundles`, inherited from the routine, inherited from `createOrResetRoutine`. No company membership exists for that principal, so `assertCompanyAccess` denies.

**Why it is hard to notice.** The affected agent cannot post an issue about it — the same denial blocks issue creation. Diagnosis is only possible from the service log.

**Expected.** Routines created for built-in agents resolve to a real responsible user, or to none, rather than to a system marker.

## What Changed

- `server/src/services/routines.ts`: `resolveRoutineResponsibleUserId` no longer accepts a system marker as a responsible user. Resolution order is now: explicit company default → company owner → first active company member → `null`. A non-empty-string guard was added so an empty or whitespace value cannot slip through as valid.
- `server/src/services/built-in-agents.ts`: `createOrResetRoutine` no longer passes the marker as the acting user.
- `packages/db/src/migrations/0182_backfill_built_in_bundles_responsible_user.sql`: idempotent backfill that clears the marker from existing routine rows so already-created instances recover without manual intervention.
- `packages/db/src/migrations/meta/_journal.json`: journal entry for the migration.

## Verification

- `pnpm --filter @paperclipai/server test` for the routines and built-in-agent suites → passes.
- Reproduced on a live instance: an agent assigned to a routine-created issue got `403 RESPONSIBLE_USER_UNAVAILABLE` on read, on listing agents, and on posting a new issue. After applying the change and the backfill, the same agent reads and acts on the issue normally.
- The backfill is idempotent — running it twice leaves the same state.

## Risks

Medium risk, mostly in the migration.

- **Migration numbering.** This branch adds `0182_…`, which now collides with upstream migrations added since the PR was opened. That is the cause of the current conflict. On rebase the file needs renumbering to sit after the current head; I will do that rather than asking a reviewer to resolve it. Worth noting for anyone reviewing the diff as it stands today.
- **Behavioral change.** Routines that previously recorded the marker will now record a real user or `null`. Issues created afterwards carry that new value, so the responsible user visible on new routine issues changes. That is the fix, but it is a visible shift.
- **Fallback choice.** "First active company member" is a heuristic. In a single-owner company it is unambiguous; in a larger one it picks a specific human who did not ask to be responsible. I chose it over leaving `null` because `null` reintroduces a different denial path, but I would take direction from maintainers here — this is the design decision in the PR most worth a second opinion.
- No breaking API change.

## Model Used

Claude Opus 4.6 (`claude-opus-4-6`) authored the original change; this description was written with Claude Opus 5 (`claude-opus-5`), 1M context, adaptive thinking, with tool use and local command execution.

## Checklist

- [x] I have not referenced internal/instance-local Paperclip issues or links (only public GitHub references)
- [x] I have updated relevant documentation to reflect my changes — none found covering routine responsible-user resolution; happy to add if there is a place for it
- [ ] All Paperclip CI gates are green — this branch conflicts with `master` (migration numbering, see Risks) and CI is red. Rebase and renumber in progress; I will report back here before asking for review.


greegorij:
Update: verified on a live installation, not just in tests.

Applied this patch to our own deployment and ran the built-in reflection-coach routine against it.

**Before** (2 hours of logs prior to the fix): 93 × `RESPONSIBLE_USER_UNAVAILABLE` on `company_access`. The agent could not read its own issue, list agents, or post a comment — including the comment it was trying to file *about* the failure.

**After**: the routine created a new issue whose `responsibleUserId` resolved to the company default rather than the marker, and the agent has been running against it with **zero** denials.

The backfill migration also did its job: rows carrying the marker (one routine, one issue) now hold the company default, and a scan of all 432 issues in that company shows no remaining occurrences.

Happy to adjust the shape of the fix if you would prefer the system-actor concept separated from the responsible-user field, as noted above.

greegorij:
Merged current `master` into the branch (merge, not rebase — so review threads pinned to existing commits survive) and fixed two things that upstream tests caught. Both are corrections to **my** change, not to master:

**1. The guard was too broad.** It rejected any candidate lacking an active company membership and substituted someone else. That means a responsible user who leaves the company would have their routines silently reassigned — well beyond the intent, which was only to stop the literal `built-in-bundles` marker. Narrowed to a named constant. Caught by `uses the routine revision responsible-user snapshot for automatic runs`.

**2. It broke provisioning of a fresh company.** With the marker rejected and no candidate available yet (no default, no owner, no members), resolution returned `null` and the existing `Routine requires a responsible user` check threw — so a brand-new company could not be provisioned at all. Now the marker is returned as a last resort: it is replaced only when there is a real user to replace it with. Caught by `auto-provisions one paused Reflection Coach bundle for a freshly created company`.

**3. Migration renumbered** `0182` → `0193`, after upstream's `0184`–`0192`. The SQL body is byte-identical: the runner matches on a content hash, so renaming does not re-run it where it was already applied, and the statements are idempotent regardless.

`routines-service.test.ts` 57/57, `companies-service.test.ts` 11/11, no conflicts.

Point 2 is worth a reviewer's opinion: returning the marker as a fallback keeps current behaviour intact, but it does mean the marker can still reach `responsibleUserId` on a company that has no users at all. If you would rather have provisioning fail loudly in that state, say so and I will flip it.

wlfonseca:
Independent reproduction and verification of this fix, from a different angle than the reflection-coach test above — fresh install, `local_trusted`, embedded Postgres, `summarizer` bundle.

## Reproduced the original bug

`paperclipai@2026.722.0` (current `latest`), single company, single member `local-board` (owner), `companies.default_responsible_user_id` correctly set to `local-board`.

After enabling `enableBuiltInAgents` and the bundled `refresh-stale-summaries` schedule, the agent hit exactly what this PR describes. Its own run output:

> The token is valid (expires in ~59 min), but the API is rejecting every call with `RESPONSIBLE_USER_UNAVAILABLE`. The user `built-in-bundles` associated with this token has no access to the company.

Result: **9 heartbeat runs in ~15 minutes**, all `exit=0`, none able to close the routine issue. The issue ended at `blocked` with `Paperclip could not resolve this issue's missing disposition automatically`. Persisted rows carrying the marker in my instance:

```
routines           2
routine_revisions  8
issues             0
heartbeat_runs     9
```

Worth noting for anyone triaging: **manual runs mask this completely.** In `resolveRoutineRunResponsibleUserId` the chain is `manualRunnerUserId ?? latestRevisionResponsibleUserId ?? routine.responsibleUserId`, so a manual trigger resolves to a real user and the agent's API writes go through — four comments landed with `authorUserId: local-board` on my first manual run. Only `automation` and schedule-sourced runs fall through to the marker. The feature looks healthy until you turn on the schedule.

## Verified this patch fixes it

Applied the `resolveRoutineResponsibleUserId` logic from this PR to the published `dist` of `@paperclipai/server@2026.722.0`, plus the equivalent of migration `0193` against my instance, then restarted.

Same routine, schedule re-enabled, triggered again:

- **1 run**, `succeeded`, `exit=0` — no re-wake loop
- Routine issue went straight to `done` instead of `blocked`
- `heartbeat_runs.responsible_user_id` on the new run: `local-board`
- Activity log confirms the agent itself made the writes that were previously denied:

```
issue.comment_added | actor_type=agent | actor_id=3a33c125-…
issue.updated       | actor_type=agent | actor_id=3a33c125-…
```

The three-tier fallback and the backfill both behave as described.

## On the Greptile 4/5 — the parent-issue path

Greptile is right that `routines.ts` lines 163-164 still let the marker through, and I think it matters more than it might look. The backfill deliberately skips companies with no `default_responsible_user_id`, so those rows keep the marker by design — and that is precisely the population that then hits the unguarded parent lookup:

```ts
if (parent?.responsibleUserId) return parent.responsibleUserId;   // can be the marker
if (parent?.createdByUserId) return parent.createdByUserId;
```

A child routine created under such an issue reproduces the original denial after this PR lands. Suggested guard, reusing the `readNonEmptyString` helper already added here:

```ts
// The parent issue can carry the marker itself when its company was never healed
// by the backfill (no default_responsible_user_id at migration time). Without this
// guard a child routine inherits the marker and reproduces the original denial.
const parentResponsible = readNonEmptyString(parent?.responsibleUserId);
if (parentResponsible && parentResponsible !== BUILT_IN_BUNDLES_ACTOR) return parentResponsible;
const parentCreator = readNonEmptyString(parent?.createdByUserId);
if (parentCreator && parentCreator !== BUILT_IN_BUNDLES_ACTOR) return parentCreator;
```

I ran this variant in my instance and it behaves the same on the happy path. I have not run the repo test suite against it — feel free to take it directly into this branch if it looks right to you, or say the word and I will open a PR against `fix/built-in-routine-responsible-user` instead.

Also relevant: #9671 is the original report of this bug, and #10481 (mine) is a duplicate of it — I am closing that one and pointing it here.

PR #9925, opened 2026-07-21
Fix external UI parser worker startup in Chrome

tickernelz:
## Summary

- disable Worker globals with `Object.defineProperty` so getter-only globals such as `caches` and `indexedDB` do not abort strict-mode bootstrap
- keep the sandbox fail-closed when a dangerous global cannot be overridden
- coalesce async parser result notifications to one animation frame to avoid rebuilding large transcripts once per JSONL line
- replace source-text lockdown assertions with behavioral coverage and add a getter-only regression

## Reproduction

On Paperclip 2026.720.0 in Chrome, opening a run from an external adapter leaves Nice mode on the generic stdout parser. Replaying the production Worker bootstrap reports:

```text
Uncaught TypeError: Cannot set property caches of #<WorkerGlobalScope> which has only a getter
```

After guarding only `caches`, startup fails next on getter-only `indexedDB`. Both descriptors are configurable in Chrome, so defining immutable `undefined` values preserves the intended lockdown.

## Verification

- `pnpm exec vitest run ui/src/adapters/sandboxed-parser-worker.test.ts`
- `pnpm --filter @paperclipai/ui build`
- exact Worker bootstrap exercised in Chrome against an external OMP parser: `ready`, then parsed result
- 3,577-line run reduced from raw JSONL to 26 Nice transcript entries; `toolcall_delta` raw count dropped to zero

### DIFF
--- a/ui/src/adapters/dynamic-loader.ts
+++ b/ui/src/adapters/dynamic-loader.ts
@@ -61,6 +61,7 @@ const failedLoads = new Set<string>();
 const loadPromises = new Map<string, Promise<DynamicParserModule | null>>();
 
 let resultNotifier: (() => void) | null = null;
+let resultNotificationFrame: number | null = null;
 
 export function setDynamicParserResultNotifier(fn: (() => void) | null): void {
   resultNotifier = fn;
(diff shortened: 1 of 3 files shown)


tickernelz:
Closing in favor of #8806, which already fixes the same WorkerGlobalScope getter-only startup crash. I reproduced #8806's root cause on Paperclip 2026.720.0 in Chrome: production bootstrap fails first on caches, then indexedDB; with immutable own shadows, the OMP external parser reaches ready and a 3,577-line run renders as 22 Nice entries with zero raw message_update/toolcall_delta lines. I will keep the independent notification-coalescing improvement separate.

Issue #7535, opened 2026-06-04
Terminal-run recovery retries permanently-failing adapter configs forever (comment spam + repeated blocked-flips, no backoff/give-up)


ianbruton:
## Summary

When an issue is assigned to an agent whose adapter config can never succeed (e.g. a model id not in the adapter's catalog, or a `process` adapter with no command), the terminal-run recovery retries **indefinitely**, producing on every cycle:

- an identical comment: *"Paperclip automatically retried continuation for this assigned `in_progress` issue during terminal run recovery, but it still has no live execution path. Latest retry failure: `adapter_failed` - …"*
- a status flip to `blocked`
- a new "Recover stalled issue …" recovery issue

We observed ~6 identical comments per issue within the hour, across two issues, with two distinct permanent failure modes (`Configured OpenCode model is unavailable: …` and `Process adapter missing command`). The retry outcome is deterministic — the config cannot heal itself — so each retry only adds noise and churn.

## Version
paperclipai 2026.428.0

## Expected

After N identical `adapter_failed` results for the same (issue, agent, config) tuple, recovery should **give up and park**: one comment, one recovery issue, no further retries until the agent's adapter config changes (config hash makes this cheap to detect). Exponential backoff would also help, but give-up-on-identical-failure is the real fix — a config error is not transient.

## Workaround

Unassign the issue (recovery only engages assigned issues), fix or clear the agent's adapter config, and cancel the accumulated recovery issues by hand.

schmelli:
## Heartbeat Priority Assessment

**Priority: HIGH** — Bug or regression that affects reliability, UX, or production operation. Should be resolved before launch.

_This is a classification comment from the heartbeat routine._

hugo-valcourt:
## Downstream reproduction and a narrow recovery contract

We reproduced the same failure shape in a local downstream deployment: an unsuccessful terminal run remained assigned to an `in_progress` issue, the recovery path treated the missing disposition/comment as work to continue, and a deterministic failure produced additional automatic runs instead of converging. Cancelling runs did not reliably shed load because another recovery wake could be enqueued.

We applied a conservative local mitigation and verified it with a timeout canary:

- no missing-comment retry after timeout/failure;
- no automatic timeout continuation;
- atomically park the source issue as `blocked`;
- ensure exactly one Board-owned recovery action;
- require an explicit operator resume before another run.

Canary result: one timed-out run, zero continuation attempts, zero retry child runs, and no subsequent automatic wake.

[...]

PR #9937, opened 2026-07-21
fix(server): cap same-root automatic retries with backoff and a park state

### LINKED ISSUE #9734: Unbounded process_lost_retry monopolizes a maxConcurrentRuns:1 agent — all other wakes starve forever (incl. the wake that would cancel the looping issue)
## Summary

On an agent with `maxConcurrentRuns: 1`, an issue whose runs keep dying (OOM in our case) is retried **indefinitely** by `process_lost_retry` with no backoff and no consecutive-failure cap. Each retry immediately re-claims the agent's only concurrency slot, so **every other wake for that agent stays `queued` with `claimedAt: null` forever.**

This is a **deadlock, not backpressure**: the queue drains only when the busy run finishes, and a crash-looping run never finishes.

The pathological case — which we hit in production — is that **the queued wake was the one that would have cancelled the looping issue.** The fix for the wedge is queued behind the wedge, so it cannot self-heal.

Related but distinct: #5021 guards retries on *already-cancelled* issues. Here the issue is live and never reaches a terminal status, so that guard never applies.

## Evidence

Agent config: `maxConcurrentRuns: 1`, `heartbeat.enabled: false`, `wakeOnDemand: true`.

`GET /api/issues/{looping}/diagnostics/wakes` — 6 runs, 5 dead, each ~15 min apart:

| run | claimed | finished | status |
|---|---|---|---|
| `63e61007` | 23:18:01 | 23:36:42 | failed (`issue_assigned`) |
| `08a3b4a6` | 23:36:43 | 23:50:50 | failed (`process_lost_retry`) |
| `2dad1fbe` | 23:50:50 | 00:09:38 | failed |
| `2cfaf226` | 00:09:38 | 00:28:24 | failed (`process_lost_retry`) |
| `96dd01e2` | 00:28:24 | 00:41:25 | failed |
| `5d460dee` | 00:41:26 | — | live (`process_lost_retry`) |

Meanwhile, `GET /api/issues/{corrective}/diagnostics/wakes` for a **different, `high` priority** issue assigned to the same agent:

```json
{
  "diagnosis": "The most recent wake for SAL-677 is currently queued for issue_assigned.",
  "events": [{
    "kind": "wake_request", "source": "assignment", "reason": "issue_assigned",
    "status": "queued", "requestedAt": "2026-07-16T22:31:06.740Z",
    "claimedAt": null, "finishedAt": null
  }],
  "wakeRequestCount": 1
}
```

**~3 hours queued, never claimed.** The agent reports `status: "running"` the entire time while producing zero output across all six runs, so no agent-level health check fires — `errorReason` is set (`"Process lost -- child pid 1026404 is no longer running"`) but `status` stays `running`.

## Why it matters

Three failure modes compound:

1. **Unbounded retry** — no backoff, no cap on consecutive process losses. A reproducibly-fatal run (OOM on a fixed-size input) retries forever at ~4/hour.
2. **No preemption** — a `high` priority wake cannot preempt a repeatedly-failing run holding the only slot. Priority is not consulted for slot admission.
3. **`status: running` masks it** — an agent in a permanent crash loop is indistinguishable from a healthy busy agent via the agents API. `errorReason` is populated but `status` is not reconciled, so operator dashboards read green.

Net effect: **one poisoned issue silently takes a single-slot agent offline permanently**, and every corrective path routed through that agent (including "cancel the poisoned issue") is unreachable. In our case the entire CTO work chain froze behind one issue.

## Suggested fixes (any one breaks the deadlock)

- **Cap consecutive `process_lost_retry` per issue** (e.g. 3), then park the issue `blocked`/`error` with the reason and **release the slot**. This is the highest-value fix — it converts a silent permanent wedge into a visible, actionable stop.
- **Exponential backoff** on repeated process loss, so the slot is free between attempts and other wakes can be claimed.
- **Reconcile `status`** — an agent whose last N runs all died should not report `running`. Surface `errorReason` in `status`.
- Optionally, let a higher-priority queued wake preempt a run that has already lost its process N times.

## Environment

Self-hosted, agent cgroup `memory.max = 3221225472` (3 GiB), `memory.events` → `oom_kill: 986`. The OOM is our own capacity problem and is separately tracked — **this report is about the retry policy turning a transient process loss into a permanent, self-sustaining wedge.** Any reproducibly-fatal run (not just OOM) reaches the same state.

fallcent-junslim:
## Thinking Path

> - When an agent run dies unexpectedly (process lost, stranded issue, review-participant recovery, source-scoped recovery), the platform automatically re-wakes it so work isn't dropped.
> - Each of those recovery paths mints a **fresh** run/wake id, and the retry cap is keyed off the immediate `retryOfRunId` lineage — so a recovery that allocates a new id resets the counter.
> - The result is unbounded automatic retries of a permanently-failing root: on a `maxConcurrentRuns: 1` agent a single failing issue can loop indefinitely (100+ runs), starving every other wake — including the wake that would cancel the loop (#9734) — and spamming recovery comments with no backoff (#7535).
> - This pull request introduces a **same-root retry cap**: a stable `(retryRootRunId, retryEpoch)` that is invariant across every automatic recovery mint, an atomic in-transaction gate that caps automatic runs at the initial run + 3 retries, exponential backoff + jitter with slot release while waiting, and a single park record (last failure reason + next owner/action) instead of repeated recovery churn.
> - The benefit is that a runaway root can no longer monopolize an agent or spam an issue, while genuine new input (human comment, issue event, explicit retry, monitor signal) still cleanly resumes work on a fresh epoch.

- Fixes #9734 — unbounded `process_lost_retry` monopolizes a single-slot agent.
- Refs #7535 — terminal-run recovery retries a permanently-failing adapter forever with no backoff/give-up. This PR adds the shared cap + backoff + park for the automatic recovery mint paths; the adapter-config give-up specifics remain tracked there.

## What Changed

- **`server/src/services/same-root-retry-cap.ts`** (new): pure policy module — resolves the invariant retry root, decides epoch advancement (external input advances the epoch; recovery-internal events do not), evaluates the cap (initial 1 + 3 retries = 4 automatic runs per `(root, epoch)`), computes exponential backoff + jitter, and builds the park descriptor (last failure reason + next owner/action + resume guidance).
- **`server/src/services/same-root-retry-gate.ts`** (new): shared DB gate. Under a per-`(root, epoch)` transaction advisory lock it counts the automatic runs already recorded and either allows (returning root/epoch/attempt for the caller to stamp + schedule) or parks. Count and insert happen under one lock, so concurrent recovery cannot exceed the cap.
- **`server/src/services/heartbeat.ts`**: the process-loss retry mint now runs the gate **inside** the run-insert transaction and inserts allowed retries as `scheduled_retry` with backoff (no immediate start, slot released); `enqueueWakeup` gained an optional `sameRootRetry` mint path so recovery callers get the same atomic cap, backoff, slot release, and park. Failure lineage only — a re-wake continuing a run that finished normally is progress, not a retry, and is never throttled.
- **`server/src/services/recovery/service.ts`**: stranded-issue and source-scoped recovery delegate to the gated mint instead of a top-level gate + post-insert update, sharing the atomic cap/backoff/park.
- **`packages/db`**: `heartbeat_runs` gains `retry_root_run_id` + `retry_epoch` with a partial index; migration `0182_same_root_retry_cap` (+ journal entry).
- **Tests**: policy unit tests, an embedded-Postgres integration test proving the runaway chain stops at 4 runs / recovery-only never mints a 5th / external input resumes on a clean epoch / concurrent minters are serialized by the advisory lock, and a production-mint end-to-end test in `heartbeat-process-recovery.test.ts`. Existing process-recovery/retry-scheduling contract tests updated for the `scheduled_retry` behavior; two resume-failure comment assertions deflaked (they raced a post-commit insert under parallel load).

## Verification

- `pnpm --filter @paperclipai/db check:migrations` → passes (journal/file counts aligned, safety check green).
- `pnpm exec vitest run` for `same-root-retry-cap`, `same-root-retry-gate.integration`, `heartbeat-process-recovery`, `heartbeat-retry-scheduling` → **144 passed / 0 failed** (run together, i.e. under parallel load).
- Integration test replays a 144-run same-root failure chain and asserts it stops at 4 total automatic runs; recovery-only repetition never produces a 5th; a new external-input wake resumes at epoch 1 / attempt 1; 8 concurrent minters stay within the cap.

## Risks

- Behavioral shift: automatic recovery retries are now delayed (backoff) and released from the slot while waiting instead of starting immediately, and are hard-capped at 4 per root/epoch. Genuine new input still resumes via epoch advancement, so operator-visible work is not dropped — on exhaustion the root is parked with the last failure reason and next owner/action rather than silently stopping.
- Migration adds two nullable/defaulted columns + a partial index to `heartbeat_runs` (additive, no backfill of historical rows required; existing rows get `retry_epoch` default 0 / null root and are treated as their own root).
- Concurrency correctness rests on the per-`(root, epoch)` transaction advisory lock; the integration test exercises the concurrent-minter race directly.

## Model Used

- Anthropic **Claude Opus 4.8** (`claude-opus-4-8`), extended thinking + tool use, via an autonomous coding agent. Local test/typecheck/migration verification run by the agent.

## Checklist

### DIFF
--- a/packages/db/src/migrations/0182_same_root_retry_cap.sql
+++ b/packages/db/src/migrations/0182_same_root_retry_cap.sql
@@ -1,0 +1,7 @@
+ALTER TABLE "heartbeat_runs"
+  ADD COLUMN IF NOT EXISTS "retry_root_run_id" uuid REFERENCES heartbeat_runs(id) ON DELETE SET NULL,
+  ADD COLUMN IF NOT EXISTS "retry_epoch" integer NOT NULL DEFAULT 0;
+
+CREATE INDEX IF NOT EXISTS heartbeat_runs_retry_root_epoch_idx
+  ON heartbeat_runs (retry_root_run_id, retry_epoch)
+  WHERE retry_root_run_id IS NOT NULL;
--- a/packages/db/src/migrations/meta/_journal.json
+++ b/packages/db/src/migrations/meta/_journal.json
@@ -1261,6 +1261,13 @@
       "when": 1784231633059,
       "tag": "0181_decision_training_retention_policy",
       "breakpoints": true
+    },
+    {
+      "idx": 182,
+      "version": "7",
+      "when": 1784231700000,
+      "tag": "0182_same_root_retry_cap",
+      "breakpoints": true
     }
   ]
 }
--- a/packages/db/src/schema/heartbeat_runs.ts
+++ b/packages/db/src/schema/heartbeat_runs.ts
@@ -42,6 +42,10 @@ export const heartbeatRuns = pgTable(
     retryOfRunId: uuid("retry_of_run_id").references((): AnyPgColumn => heartbeatRuns.id, {
       onDelete: "set null",
     }),
+    retryRootRunId: uuid("retry_root_run_id").references((): AnyPgColumn => heartbeatRuns.id, {
+      onDelete: "set null",
+    }),
+    retryEpoch: integer("retry_epoch").notNull().default(0),
     processLossRetryCount: integer("process_loss_retry_count").notNull().default(0),
     scheduledRetryAt: timestamp("scheduled_retry_at", { withTimezone: true }),
     scheduledRetryAttempt: integer("scheduled_retry_attempt").notNull().default(0),
--- a/server/src/__tests__/heartbeat-process-recovery.test.ts
+++ b/server/src/__tests__/heartbeat-process-recovery.test.ts
@@ -5214,6 +5214,98 @@ describeEmbeddedPostgres("heartbeat orphaned process recovery", () => {
     const wakes = await db.select().from(agentWakeupRequests).where(eq(agentWakeupRequests.agentId, agentId));
     expect(wakes.some((row) => row.reason === "run_liveness_continuation")).toBe(false);
   });
+
+  it("bounds same-root automatic retries to four runs via the production mint, parks with owner/action, and resumes on new external input", async () => {
+    const { agentId, companyId, issueId, runId } = await seedStrandedIssueFixture({
+      status: "in_progress",
+      runStatus: "failed",
+    });
+    const heartbeat = heartbeatService(db);
+    const oldFinishedAt = new Date("2026-03-19T00:05:00.000Z");
+
+    const loadRun = async (id: string) =>
+      db.select().from(heartbeatRuns).where(eq(heartbeatRuns.id, id)).then((rows) => rows[0]!);
+    // The exact gate-source shape recovery paths pass into the production mint.
+    const gateSourceOf = (run: typeof heartbeatRuns.$inferSelect) => ({
+      id: run.id,
+      companyId: run.companyId,
+      retryRootRunId: run.retryRootRunId,
+      retryEpoch: run.retryEpoch,
+      errorCode: run.errorCode,
+      error: run.error,
+      responsibleUserId: run.responsibleUserId,
+    });
+    // Finalize a run as a past failure so the next mint sees a terminal run
+    // (no coalescing) and the issue-rewake throttle (which only counts recent
+    // terminal runs) stays out of the way.
+    const failRunInPast = async (id: string) =>
+      db
+        .update(heartbeatRuns)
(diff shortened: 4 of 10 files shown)


fallcent-junslim:
Closing this draft because the underlying work was intentionally discontinued. This proposal will not be pursued in its current form.

PR #9943, opened 2026-07-21
feat(server): declarative instance settings overrides via env

stubbi:
## Thinking Path

> - Instance behavior is governed by instance settings (general + experimental feature flags) stored as a singleton row and edited by an instance admin
> - Operators who run Paperclip through GitOps or immutable infrastructure have no declarative way to pin those settings: the only path is a human toggling them in the UI, which drifts and does not survive a database restore
> - A deployment should be able to force specific instance settings from config, the same way `PAPERCLIP_CLOUD_BILLING` / `PAPERCLIP_CLOUD_TRIAL_BANNER` already force two experimental flags today
> - This pull request adds `PAPERCLIP_INSTANCE_SETTINGS_OVERRIDES`, a JSON env var whose `general` and `experimental` sections are validated and merged over the stored settings at read time, and stripped from admin patches so a forced value never persists
> - The benefit is drift-proof, reproducible instance configuration from deployment config, with zero behavior change when the variable is unset

No existing issue. Describing in-PR following the feature template:

**Problem or motivation**
There is no declarative, deployment-time way to pin instance settings. Operators running under GitOps must configure experimental flags and general settings by hand in the UI; the configuration lives only in the database, drifts from source control, and is lost on a restore into a fresh database.

**Proposed solution**
Read an optional `PAPERCLIP_INSTANCE_SETTINGS_OVERRIDES` JSON env var whose `general` and `experimental` sections are validated with the existing zod schemas (partial + strip unknown keys) and merged over the stored settings at read time. Overridden keys are stripped from admin patches before write, so forced values are never persisted back to the row. Invalid JSON or an invalid section warns once and falls back to the stored settings.

**Alternatives considered**
Per-flag env vars (like the two existing `PAPERCLIP_CLOUD_*` ones) do not scale to the full settings surface and add a new variable per flag. A boot-time seed that writes the DB row was rejected because it mutates tenant state and cannot represent "forced" (a later admin edit would silently win). A read-time merge keeps the stored row untouched and makes the override authoritative for exactly as long as it is set.

**Roadmap alignment**
Small, additive operator-ergonomics improvement; no overlap with planned core feature work. Default path (variable unset) is byte-for-byte unchanged.

## What Changed

- Added `parseInstanceSettingsOverrides(env)` in `server/src/services/instance-settings.ts`: parses the JSON env var, validates the `general` and `experimental` sections (partial + strip), strips server-managed worktree-activation fields, and warns once on invalid input.
- Added `resolveGeneralSettings` / `resolveExperimentalSettings` (normalize stored value, spread overrides, re-normalize) and `stripOverriddenPatchKeys`.
- Wired overrides through every `instanceSettingsService` getter (override-merged reads) and the `updateGeneral` / `updateExperimental` writers (strip overridden keys so they never persist). Service signature unchanged; all existing call sites unaffected.
- Routed the one raw `instance_settings` read in `server/src/services/feedback.ts` through the override-aware helper for reads, while keeping the DB write-back sourced from the un-overridden stored value.

## Verification

- `pnpm exec vitest run server/src/__tests__/instance-settings-service.test.ts server/src/__tests__/instance-settings-routes.test.ts` — passing (override parse/resolve/strip, service read-merge, patch-strip, and no-env-var-unchanged cases).
- `pnpm exec vitest run server/src/__tests__ -t feedback` — passing, including a new embedded-Postgres regression test asserting an active `censorUsernameInLogs` override does not persist into the DB on a feedback write.
- `pnpm exec tsc -p server --noEmit` — clean.

## Risks

Low risk. The feature is inert unless `PAPERCLIP_INSTANCE_SETTINGS_OVERRIDES` is set: with it unset, `parseInstanceSettingsOverrides` returns empty sections and every resolve/strip helper degenerates to the prior normalization, so reads and writes are identical to before. Overridden keys are stripped from patches, so an override can never leak into persisted state. Invalid input is warned once and ignored rather than throwing.

## Model Used

Claude (Anthropic), claude-fable-5, extended thinking, tool use / code execution.

## Checklist

### DIFF
--- a/server/src/__tests__/feedback-service.test.ts
+++ b/server/src/__tests__/feedback-service.test.ts
@@ -519,6 +519,57 @@ describeEmbeddedPostgres("feedbackService.saveIssueVote", () => {
     });
   });
 
+  it("does not persist an env-overridden censorUsernameInLogs value on the sharing-preference write-back", async () => {
+    const { issueId, commentId } = await seedIssueWithAgentComment();
+
+    await db.insert(instanceSettings).values({
+      singletonKey: "default",
+      general: {
+        censorUsernameInLogs: true,
+        feedbackDataSharingPreference: "prompt",
+      },
+      experimental: {},
+    });
+
+    const previousOverrides = process.env.PAPERCLIP_INSTANCE_SETTINGS_OVERRIDES;
+    process.env.PAPERCLIP_INSTANCE_SETTINGS_OVERRIDES = JSON.stringify({
+      general: { censorUsernameInLogs: false },
+    });
+
+    try {
+      const result = await svc.saveIssueVote({
+        issueId,
+        targetType: "issue_comment",
+        targetId: commentId,
+        vote: "up",
+        authorUserId: "user-1",
+        allowSharing: true,
+      });
+
+      expect(result.persistedSharingPreference).toBe("allowed");
+
+      const settings = await db
+        .select()
+        .from(instanceSettings)
+        .where(eq(instanceSettings.singletonKey, "default"))
+        .then((rows) => rows[0] ?? null);
+
+      // The env override forces censorUsernameInLogs=false at read time, but
+      // the write-back must persist the raw stored value (true), never the
+      // override.
+      expect(settings?.general).toMatchObject({
+        censorUsernameInLogs: true,
+        feedbackDataSharingPreference: "allowed",
+      });
+    } finally {
+      if (previousOverrides === undefined) {
+        delete process.env.PAPERCLIP_INSTANCE_SETTINGS_OVERRIDES;
+      } else {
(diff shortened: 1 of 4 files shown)


stubbi:
Closing as superseded by upstream work.

**8f195c2f4 — "Let operators supply defaults for selected instance settings (#12285)"** added `packages/shared/src/setting-defaults.ts` and the `PAPERCLIP_SETTING_DEFAULTS` environment variable (documented in `docs/deploy/environment-variables.md`). That is the same feature this PR proposed: declarative, env-supplied instance settings values applied at boot, so a deployer can ship a configured instance without hand-editing settings post-install.

Since upstream's version is the one that's landed, documented and tested, there's no reason to rebase this branch onto it — especially as it's ~2000 PRs stale and conflicting.

Note for anyone finding this later: the `superagent-security` thread on this PR was never addressed, so don't lift this branch's code as-is.

PR #9944, opened 2026-07-21
feat(k8s-provider): egressPolicy open-internet mode

stubbi:
## Thinking Path

> - The kubernetes sandbox provider runs each agent workspace in an isolated tenant namespace and generates a NetworkPolicy (or CiliumNetworkPolicy) that governs the sandbox's outbound network access
> - Today egress is an FQDN/CIDR allowlist: agents can only reach the hosts an operator enumerates, which is correct for locked-down deployments but impractical when agents legitimately need the open internet (arbitrary package registries, customer APIs, web fetch)
> - Operators need a way to opt a deployment into broad egress without hand-maintaining an ever-growing allowlist, while still keeping agents off cluster internals and cloud metadata
> - This pull request adds `egressPolicy: allowlist | open-internet`; `open-internet` permits public internet on TCP 443/80 while carving out private ranges, link-local metadata, CGNAT, loopback, this-network, and multicast
> - The benefit is a single opt-in switch for open egress with a hardened default-safe deny set, defaulting to the existing allowlist behavior so nothing changes unless selected

No existing issue. Describing in-PR following the feature template:

**Problem or motivation**
The kubernetes sandbox provider only supports allowlist egress. Deployments whose agents need the open internet must enumerate every destination FQDN/CIDR, which does not scale and breaks whenever an agent needs a host the operator did not predict. There is no built-in "open but hardened" posture.

**Proposed solution**
Add an `egressPolicy` config field (`allowlist` default, `open-internet` opt-in). In `open-internet` mode both the standard NetworkPolicy and CiliumNetworkPolicy builders emit an egress rule to `0.0.0.0/0` on TCP 443 and 80 only, with a shared except set (`10.0.0.0/8`, `172.16.0.0/12`, `192.168.0.0/16`, `169.254.0.0/16`, `100.64.0.0/10`, `127.0.0.0/8`, `0.0.0.0/8`, `224.0.0.0/4`). DNS stays pinned to kube-dns, the paperclip-server rule and any FQDN allowlist rules remain intact. Threaded end to end from the `PAPERCLIP_K8S_EGRESS_POLICY` env var through the execution-policy bootstrap, plugin config schema, tenant orchestrator, and both builders.

**Alternatives considered**
Documenting a wildcard-heavy allowlist was rejected: NetworkPolicy cannot express FQDN wildcards and Cilium `toFQDNs` still needs enumerated names. Removing egress restrictions entirely was rejected because it would expose cluster internals and the cloud metadata endpoint. A dedicated per-tenant policy CRD was out of scope for a provider-level posture switch.

**Roadmap alignment**
Additive provider capability, opt-in, default-preserving; no overlap with planned core work.

## What Changed

- Added `egressPolicy: z.enum(["allowlist","open-internet"]).default("allowlist")` to the kubernetes provider config schema and the UI manifest schema.
- `buildCiliumNetworkPolicyManifest` and `buildNetworkPolicyManifests` emit the hardened public-internet rule (TCP 443/80, shared except set) only when `egressPolicy === "open-internet"`; the except CIDRs are a single exported constant shared by both builders.
- Threaded the field through `EnsureTenantInput` and both `ensureNetworkPolicies` builder calls, the `plugin.ts` lease-acquisition `ensureTenant` call, `PAPERCLIP_K8S_EGRESS_POLICY` parsing in `execution-policy-bootstrap.ts` (invalid values throw, matching the existing `egressMode` handling), and the `KubernetesEnvironmentConfigInput` interface.

## Verification

- `pnpm exec vitest run packages/plugins/sandbox-providers/kubernetes` — passing, including new tests: open-internet emits the `0.0.0.0/0` rule with the required excepts on 443/80, kube-dns and paperclip-server rules survive, and the default (allowlist) emits no public-internet rule.
- New `execution-policy-bootstrap` tests cover unset (undefined) and invalid (throws) `PAPERCLIP_K8S_EGRESS_POLICY`.
- `pnpm exec tsc -p server --noEmit` — clean.

## Risks

Low risk. `egressPolicy` defaults to `allowlist`; both builders gate the new rule strictly on `=== "open-internet"`, so an unset or default value produces byte-identical manifests to before and no existing deployment changes. `open-internet` is a deliberate broadening of egress and is documented as such: it is IPv4-scoped (the fallback CIDR math is IPv4, matching the pre-existing standard-mode fallback), so operators on dual-stack clusters who need IPv6 egress restrictions should keep `allowlist`. The new field is optional on every interface, so no call site breaks.

## Model Used

Claude (Anthropic), claude-fable-5, extended thinking, tool use / code execution.

## Checklist

### DIFF
--- a/packages/plugins/sandbox-providers/kubernetes/README.md
+++ b/packages/plugins/sandbox-providers/kubernetes/README.md
@@ -70,6 +70,7 @@ Common optional fields:
 | `egressAllowFqdns` | `[]` | Additional FQDNs (beyond adapter defaults like `api.anthropic.com`). |
 | `egressAllowCidrs` | `[]` | Additional CIDRs to allow egress to. |
 | `egressMode` | `"standard"` | `standard` (NetworkPolicy + CIDRs) or `cilium` (CiliumNetworkPolicy + FQDN allow-list). |
+| `egressPolicy` | `"allowlist"` | Namespace baseline posture. `allowlist` restricts egress to the configured FQDNs/CIDRs. `open-internet` additionally allows public IPv4 on TCP 443/80 with private, CGNAT, link-local (cloud IMDS), loopback, this-network, and multicast ranges excluded. Also settable via `PAPERCLIP_K8S_EGRESS_POLICY`. |
 | `runtimeClassName` | (none) | e.g. `kata-fc` for Firecracker-backed microVMs. Cluster must have the RuntimeClass installed. |
 | `serviceAccountAnnotations` | `{}` | Annotations applied to per-tenant ServiceAccount (e.g. IRSA `eks.amazonaws.com/role-arn`). |
 | `jobTtlSecondsAfterFinished` | `900` | Seconds after a Job completes before garbage-collection. |
@@ -93,6 +94,17 @@ Keep provider-level egress defaults narrow, then grant only the destinations a t
 ```
 
 The provider creates a workload-owned policy selected by the task run label, so the additional destinations do not become reachable from other concurrent agent pods. Cilium mode enforces FQDNs directly. Standard NetworkPolicy mode cannot express FQDNs, so an FQDN grant permits public IPv4 TCP 80/443 for that run while excluding private, loopback, link-local, CGNAT, and multicast ranges. Network failures that look policy-related include the grant path in stderr, and the sandbox exposes the effective policy through `PAPERCLIP_NETWORK_EGRESS_*` environment variables.
+
+#### How `egressPolicy` and task-scoped grants compose
+
+The two are different layers and Kubernetes unions them:
+
+- `egressPolicy` is the **namespace baseline**. It is written into the tenant's long-lived `paperclip-egress-allow` NetworkPolicy (or `paperclip-egress-fqdn` CiliumNetworkPolicy), selects every agent pod in the namespace, and lives for as long as the tenant does.
+- A task grant is an **additive, workload-owned policy** selected by `paperclip.io/run-id`, created per run and garbage-collected with its Job or Sandbox.
+
+The grant builder deliberately does not receive `egressPolicy`, so turning the provider baseline to `open-internet` never changes what a per-task grant itself permits: the run's reachability is the union of the two, and the grant stays a truthful record of what that specific task asked for. `PAPERCLIP_NETWORK_EGRESS_POLICY` reports the baseline (`kubernetes-default-deny` or `kubernetes-open-internet`) while `PAPERCLIP_NETWORK_EGRESS_ALLOW_*` report the task grant.
+
+With `egressPolicy: "open-internet"` the baseline already covers public TCP 443/80, so task grants for public hosts become redundant. Keep the baseline on `allowlist` if you want per-task grants to be the only path to the internet.
 
 ## What gets created in your cluster
 
--- a/packages/plugins/sandbox-providers/kubernetes/src/cilium-network-policy.ts
+++ b/packages/plugins/sandbox-providers/kubernetes/src/cilium-network-policy.ts
@@ -1,3 +1,5 @@
+import { PRIVATE_AND_LINK_LOCAL_EXCEPT_CIDRS } from "./network-policy.js";
+
 export interface BuildCiliumNetworkPolicyInput {
   namespace: string;
   paperclipServerNamespace: string;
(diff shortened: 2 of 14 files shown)


stubbi:
Rebased onto current `master` (was 103 commits behind) and reconciled with #10155, which landed in the same files.

## Rebase

Four conflicts, all in the egress builders and their tests:

| File | Conflict | Resolution |
|---|---|---|
| `src/network-policy.ts` | #10155 added `includeBaseRules` / `podSelector` / `ownerReferences` around the same fallback expression this PR gates | kept both: the fallback condition is now `egressPolicy === "open-internet" \|\| (no CIDRs && FQDNs present)` |
| `src/cilium-network-policy.ts` | same, plus `PRIVATE_AND_LINK_LOCAL_EXCEPT_CIDRS` had to stay exported for reuse | kept both, export preserved |
| `test/unit/network-policy.test.ts`, `test/unit/cilium-network-policy.test.ts` | new cases appended at the same file tail | both suites' cases kept |

No behavior from #10155 was dropped: `includeBaseRules: false`, run-label selectors and owner references are all intact.

## How `egressPolicy` and task-scoped grants compose

This is the part worth reviewing. The two features are different layers, and Kubernetes unions policies, so they stack rather than override:

- **`egressPolicy` is the namespace baseline.** It is written into the tenant's long-lived `paperclip-egress-allow` NetworkPolicy (or `paperclip-egress-fqdn` CiliumNetworkPolicy), selects every agent pod in the namespace, and lives as long as the tenant.
- **A task grant is additive and per-run.** `createScopedNetworkEgressPolicy` builds a workload-owned policy selected by `paperclip.io/run-id` with `includeBaseRules: false`, created at acquire and garbage-collected with its Job or Sandbox.

The grant path deliberately does **not** pass `egressPolicy` to the builders, so turning the provider baseline to `open-internet` never changes what a per-task grant itself permits. A run's actual reachability is the union of baseline and grant; the grant object stays a truthful record of what that specific task asked for, which is what makes the lease audit metadata and the denial hint meaningful. Two new tests in `scoped-network-egress.test.ts` assert this at the real call site (a cilium FQDN grant emits exactly one rule and no `0.0.0.0/0`; a standard CIDR grant emits only the granted CIDR), so a future refactor that spreads provider config into the scoped builder fails the suite instead of silently widening every per-task grant.

One coupling fix on top of the rebase: `PAPERCLIP_NETWORK_EGRESS_POLICY` was hardcoded to `kubernetes-default-deny`. Under `open-internet` that told the agent the wrong thing, so it now reports the actual baseline (`kubernetes-default-deny` or `kubernetes-open-internet`) through a small exported `baselineEgressPolicyLabel()`, while `PAPERCLIP_NETWORK_EGRESS_ALLOW_*` keep reporting the task grant. Baseline and grant are now separately legible from inside the sandbox.

Operator guidance is in the README: with `open-internet` the baseline already covers public TCP 443/80, so task grants for public hosts become redundant; keep `allowlist` if you want per-task grants to be the only path to the internet.

## Also updated

- README: `egressPolicy` row in the config table (including `PAPERCLIP_K8S_EGRESS_POLICY`) and a "How `egressPolicy` and task-scoped grants compose" subsection under #10155's task-scoped egress docs.
- `manifest.ts` JSON Schema description now states that the field is the namespace baseline and does not affect task grants.
- Doc comments on both builders' `egressPolicy` field record the baseline-vs-additive contract.

## Verification

`pnpm --dir packages/plugins/sandbox-providers/kubernetes test -- --run` — **190 passed**, 7 failed. The 7 failures are `test/unit/file-sync.test.ts` only and are pre-existing and environmental: they need `/proc/self/fd`, which macOS does not provide (`realpath: /proc/self/fd/7: No such file or directory`). Confirmed byte-identical on a clean `origin/master` worktree (same 7 tests, same errors). Note that `pnpm-workspace.yaml` excludes `packages/plugins/sandbox-providers/**`, so this suite is not covered by CI and the local run is the real gate.

Also green: `pnpm --dir packages/plugins/sandbox-providers/kubernetes typecheck`, `pnpm --dir server typecheck`, and `pnpm exec vitest run server/src/services/execution-policy-bootstrap.test.ts` (14 passed).

## Interaction with the other open plugin PRs

Textually adjacent but semantically independent of #10263 (exec cwd/signal/`timeoutMs`), #10264 (lease reuse) and #10267 (v1beta1 auto-detection). Expected conflicts are additive-neighbor only: one config key each in `types.ts` and `manifest.ts`, and the `plugin.ts` import block plus the acquire-lease body. Worth noting for #10264: reuse tears down any lease that carried a task-scoped grant, which is correct and unaffected here, because the baseline policy is namespace-scoped and tenant-lived rather than lease-owned. Happy to rebase behind whichever lands first.

PR #5980, opened 2026-05-14
fix(adapter-acpx-local): resolve runtime binary via ancestor node_modules/.bin (#5932)

koshaji:
## Thinking Path

> - Paperclip orchestrates AI agents for zero-human companies
> - One adapter category, `acpx_local`, runs Claude/Codex agents via Agent Client Protocol wrappers
> - Those wrappers `exec` a runtime binary like `claude-agent-acp` or `codex-acp`
> - The resolver built that as `<adapter-package>/node_modules/.bin/<bin>`, which only exists in monorepo dev
> - When `@paperclipai/adapter-acpx-local` is installed by `npx paperclipai`, npm/pnpm hoist the binary up to the consumer's `node_modules/.bin/`, so the adapter-local path is empty
> - The wrapper then `exec`s a non-existent absolute path, hits ENOENT, exits 1 before initialize, and Paperclip auto-creates a `stranded_assigned_issue` recovery card masking the root cause
> - This pull request walks up from the adapter dir looking for `node_modules/.bin/<name>` and falls back to the bare name for PATH lookup
> - The benefit is `acpx_local` works for `npx paperclipai` installs out of the box, closing #5932

## What Changed

- Added `findAncestorBin(startDir, binName)` that walks up the directory tree looking for `node_modules/.bin/<binName>`, matching pnpm/npm hoisting semantics
- Made `resolveBuiltInAgentCommand` async; uses the new helper; falls back to the bare binary name when no ancestor `.bin` is found (PATH escape hatch)
- Updated the single call site to `await` it
- Added 4 vitest cases: own-dir hit, ancestor hit (the regression case), absent everywhere, termination at filesystem root

## Verification

- `pnpm exec vitest run packages/adapters/acpx-local/src/server/execute.test.ts` — 13/13 pass (4 new + 9 existing untouched)
- `pnpm exec tsc --noEmit -p packages/adapters/acpx-local/tsconfig.json` — clean
- Confirmed `@agentclientprotocol/claude-agent-acp@0.33.1` exposes `bin: { "claude-agent-acp": "dist/index.js" }`, so the consumer-root `.bin/claude-agent-acp` symlink is the correct resolution target

## Risks

- **Low.** Monorepo dev case still hits the adapter's own `node_modules/.bin` on the first hop — unchanged.
- PATH fallback widens what `exec` can resolve, but only for closed enum `agent` values (`"claude"` or `"codex"`).
- One additional `await` at a single call site that was already inside an async function.

[...]

PR #10002, opened 2026-07-22
fix(acpx): retry ancestor node_modules/.bin lookup before falling back to bare PATH exec

quratus:
## Thinking Path

> - The `acpx-engine` subsystem is responsible for spawning local `claude`/`codex` ACP adapters and resolving which binary to `exec`.
> - `resolveBuiltInAgentCommand` resolves the runtime binary via `findAncestorBin`, which walks up from the adapter package dir looking for `node_modules/.bin/<name>`. If that lookup comes back empty, it silently falls back to the bare binary name and relies on the spawning host process's ambient `PATH` to resolve it via `exec`.
> - pnpm regenerates `node_modules/.bin/<name>` non-atomically (unlink, then re-symlink) during install/rebuild. A heartbeat spawn that races a concurrent pnpm rebuild can hit `fs.access` during that exact gap and see a false "not found", even though the package is correctly installed a moment later.
> - That false negative falls through to the bare-name/PATH escape hatch, which only works if the host process happens to have that `.bin` dir on `PATH` — otherwise the wrapper script's `exec <bin> "$@"` exits 127 before the ACP `initialize` handshake completes, and the agent's heartbeat run fails outright.
> - This PR adds a short retry window around the ancestor lookup so a transient miss during a rebuild doesn't fall through to the fragile PATH-only path, and logs (best-effort, non-blocking) when the fallback is actually taken so a genuine PATH gap stays diagnosable without manual filesystem digging.
> - The benefit is fewer spurious heartbeat/spawn failures caused purely by install-time timing, with no change to behavior when the binary is genuinely absent.

No public GitHub issue exists for this yet. Following the bug report template:

- **What happened?** Two locally-run agent processes using the `claude_local` adapter failed their heartbeat run with `ACP agent exited before initialize completed (exit=127, signal=null)`, tracing to a wrapper script line `exec: claude-agent-acp: not found`.
- **Expected behavior:** The adapter should resolve and exec its runtime binary (`claude-agent-acp`, hoisted into an ancestor `node_modules/.bin`) reliably, without spurious failures caused by concurrent package-manager activity.
- **Steps to reproduce:** Race a `pnpm install`/rebuild that regenerates `node_modules/.bin/claude-agent-acp` (unlink + re-symlink, non-atomic) against a heartbeat spawn calling `resolveBuiltInAgentCommand` at the same instant; the ancestor-bin lookup can observe the binary as transiently absent and fall back to an ambient-`PATH` exec that has no matching entry, exiting 127.
- **Paperclip version/commit:** reproduced against `master` at the commit where this branch forked (`7985c1e00`).
- **Deployment mode / installation method:** local (`claude_local` adapter, ACP engine spawn path); not remote-execution-target specific.
- **Agent adapter(s) involved:** Claude Code (local ACP engine). The same `findAncestorBin` resolution path is shared by the Codex built-in binary case (`codex-acp`), so it is affected identically in principle, though only the Claude case was observed in production.
- **Relevant logs:** `ACP agent exited before initialize completed (exit=127, signal=null): .../paperclip-acpx-wrappers/.../claude-*.sh: line 14: exec: claude-agent-acp: not found`, with the binary's own filesystem mtime matching the failing heartbeat's timestamp exactly (i.e. a live rebuild was in progress).

Related PR: **#5980** ("resolve runtime binary via ancestor node_modules/.bin") added the original ancestor-walk resolution this repo already has in-tree — it does not cover this specific non-atomic-rebuild race, which is what this PR addresses. No other open PR covers this gap (searched for `ancestor node_modules .bin retry`, `claude-agent-acp not found`, and the exact exit-127 log signature).

## What Changed

- Added `findAncestorBinWithRetry(startDir, binName)`: retries `findAncestorBin` up to 3 times, 250ms apart, before giving up. Exported alongside `findAncestorBin` for direct testing.
- `resolveBuiltInAgentCommand` (now exported for direct unit testing) uses the retrying resolver and accepts an optional `onLog` callback; when the ancestor lookup still comes up empty after retries, it logs a `stderr` warning naming the binary and search root before falling back to the bare-name PATH escape hatch, instead of silently producing a wrapper that may or may not exec correctly.
- That diagnostic log call is fire-and-forget (`.catch(() => {})`): a rejection from a fallible logging sink can no longer propagate and block the PATH fallback from being attempted — the fallback is the whole point of the code path, so it must not depend on logging succeeding.
- Threaded `input.ctx.onLog` through the single call site in `buildRuntime`.
- Added 3 vitest cases: recovers when the bin appears mid-retry-window (the regression case), still returns `null` when it never appears within the window (existing terminal-fallback behavior unchanged), and still returns the bare-command fallback when the diagnostic `onLog` call itself rejects.

## Verification

- `pnpm vitest run packages/adapter-utils/src/acpx-engine/execute.test.ts` — 58/58 pass (3 new + 55 existing).
- `pnpm exec tsc --noEmit` (in `packages/adapter-utils`) — clean.

## Risks

- **Low.** Adds up to 500ms of latency only on the cold path where the ancestor `.bin` isn't found on the first attempt (rare — normal installs resolve on attempt 1). No change to the resolution algorithm itself, no change to remote-execution or explicit `agentCommand` paths, no PATH/launchd changes. The logging fix strictly removes a failure mode (a rejected log call blocking the fallback); it cannot introduce a new one since the fallback behavior when logging succeeds is unchanged.

## Model Used

- Claude Sonnet 5 (`claude-sonnet-5`), agentic harness with tool use. Used for root-cause diagnosis (filesystem inspection of the hoisted binary and its mtime, live agent status via the Paperclip API confirming self-heal, git history/PR search confirming no existing fix covers this exact race), fix design, the Greptile-flagged log-rejection fix, and test authoring.

## Checklist

- [x] I have considered and documented risks above

### DIFF
--- a/packages/adapter-utils/src/acpx-engine/execute.test.ts
+++ b/packages/adapter-utils/src/acpx-engine/execute.test.ts
@@ -10,6 +10,7 @@ import { DEFAULT_REMOTE_SANDBOX_ADAPTER_TIMEOUT_SEC } from "@paperclipai/adapter
 import {
   createAcpxEngineExecutor,
   findAncestorBin,
+  findAncestorBinWithRetry,
   geminiVersionSupportsNativeAcpFlag,
   parseGeminiVersionParts,
   rewriteGeminiAcpFlagForVersion,
@@ -1421,6 +1422,39 @@ describe("findAncestorBin", () => {
 
   it("terminates at the filesystem root instead of looping forever", async () => {
     const resolved = await findAncestorBin("/", "definitely-not-a-real-bin-name-xyz");
+    expect(resolved).toBeNull();
+  });
+});
+
+describe("findAncestorBinWithRetry", () => {
+  async function writeFakeBin(dir: string, name: string) {
+    const binDir = path.join(dir, "node_modules", ".bin");
+    await fs.mkdir(binDir, { recursive: true });
+    const binPath = path.join(binDir, name);
+    await fs.writeFile(binPath, "#!/usr/bin/env bash\necho ok\n", { mode: 0o755 });
+    return binPath;
+  }
+
+  it("recovers from a transient miss (e.g. a pnpm rebuild racing the lookup)", async () => {
+    const root = await makeTempRoot();
+    const packageDir = path.join(root, "node_modules", "@paperclipai", "adapter-utils");
+    await fs.mkdir(packageDir, { recursive: true });
+
+    // Simulate the bin symlink not existing yet on the first lookup, then
+    // landing before the retry window elapses (ETR-54).
+    const resolvePromise = findAncestorBinWithRetry(packageDir, "claude-agent-acp");
+    const expectedBin = await writeFakeBin(root, "claude-agent-acp");
+
+    expect(await resolvePromise).toBe(expectedBin);
+  });
+
+  it("returns null when the binary never appears within the retry window", async () => {
+    const root = await makeTempRoot();
+    const packageDir = path.join(root, "node_modules", "@paperclipai", "adapter-utils");
+    await fs.mkdir(packageDir, { recursive: true });
+
+    const resolved = await findAncestorBinWithRetry(packageDir, "claude-agent-acp");
+
     expect(resolved).toBeNull();
   });
 });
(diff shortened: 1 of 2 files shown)


quratus:
Closing per updated org policy: quratus/paperclip is our canonical master going forward. Retargeted to https://github.com/quratus/paperclip/pull/3 (same branch, same content).

PR #10039, opened 2026-07-22
security(agents): reject unknown write-path fields + block agent self-PATCH of governance fields (LOOA-364/161)

nguyenm7:
## Thinking Path

> - The control plane exposes agent write paths (`PATCH /api/agents/:id`, agent create, agent hire) that boards and agents call to manage agent records.
> - Two gaps existed on those paths: the validator silently stripped unknown body fields and returned 200 without writing them, and an agent could patch its own governance fields (budget, spend, status, reporting line) with its own key.
> - Silent stripping makes a caller believe a write succeeded when nothing happened. Self-service governance writes let an agent raise its own budget or reset its own spend.
> - This pull request makes the write-path schemas strict so unknown fields return a 400 that names the field, and adds a guard that rejects agent-authenticated writes to governance fields with 403.
> - The benefit is that agent write paths fail loudly on bad input and governance fields stay under board control.

No public GitHub issue exists for these. Bug shape described inline below.

**What happened?**

`PATCH /api/agents/:id` (and the create/hire routes) returned 200 for any unknown body key and wrote nothing: `validate()` ran `schema.parse()` with zod's default strip mode, so an unrecognized key was removed before the route saw it. A caller setting `description` (not a real field — the seat lives in `title`) got a success response and no write, which reads as "the field exists and I set it". Separately, an agent using its own API key could raise its own `budgetMonthlyCents`, zero its `spentMonthlyCents`, flip its `status`, or re-parent itself via `reportsTo` through `PATCH /api/agents/:id`: these are not profile-change-consent fields, so they bypassed the change-consent gate and rode the `allow_self` grant on `agent_config:update` straight to a write.

**Expected behavior**

Unknown fields on agent write paths return a 400 that names the offending field and lists the accepted fields. Agent-authenticated patches that touch the four governance fields return 403 before patch processing. Board-authenticated requests are unaffected.

**Steps to reproduce**

1. Send `PATCH /api/agents/:id` with body `{"description": "x"}` using any valid credential. Before this change the API returns 200 and writes nothing.
2. Send `PATCH /api/agents/:id` for the agent's own id with body `{"budgetMonthlyCents": 999999}`, authenticated with that agent's own API key. Before this change the write succeeds.

Both gaps re-verified as still open on `master` when this change was authored, and the branch was rebased onto current `master` before this revision.

## What Changed

- `validate()` now surfaces a strict schema's `unrecognized_keys` issue as a **400** that names the offending field and lists accepted fields, deferring to standard ZodError rendering when the body also trips a substantive rule.
- `createAgentSchema` is now `.strict()`, which carries through the `.omit()`/`.partial()`/`.extend()` chains to the hire and update schemas.
- `assertNoAgentGovernanceFieldMutation` rejects agent-authenticated patches touching `budgetMonthlyCents`, `spentMonthlyCents`, `status`, or `reportsTo` with **403** before patch processing. `role` is intentionally not in this set — the change-consent path already gates it, and blocking it here would break that legitimate flow.

## Verification

- `agent-permissions-routes.test.ts` gains 4 unknown-field cases and 5 governance-field cases (4 governance fields × block, plus a board-allowed guard). All encode the vulnerability: they fail against the previous code and pass with this change.
- No regression across 11 agent create/hire/update suites (144 tests) from the newly-strict schema; shared + server typecheck clean.
- The `company-branding` route was checked and is unaffected (it uses inline `.parse()`, not the `validate()` middleware).
- This revision is a rebase of the reviewed change onto current `master` (no textual conflicts); the full CI matrix revalidates on this head.

## Risks

- Low. Clients that previously sent junk fields and relied on silent stripping now receive a 400 naming the field — that failure is the fix working as intended.
- Board-authenticated governance writes are unchanged. No schema or migration changes.

## Model Used

- Original change: Claude (Anthropic), Opus 4-class model, via the Claude Code agent runtime with extended thinking and tool use (code editing + local test execution).
- Rebase refresh and PR body rework: Claude Fable 5 (`claude-fable-5`), extended thinking, agentic tool use.

## Checklist

- [x] I have searched GitHub for duplicate or related PRs and confirmed this is not a duplicate (companion security PR: #10038)

### DIFF
--- a/packages/shared/src/validators/agent.ts
+++ b/packages/shared/src/validators/agent.ts
@@ -84,7 +84,9 @@ export const createAgentSchema = z.object({
   budgetMonthlyCents: z.number().int().nonnegative().optional().default(0),
   permissions: agentPermissionsSchema.optional(),
   metadata: z.record(z.string(), z.unknown()).optional().nullable(),
-});
+  // Strict, so an unknown key is a 400 rather than a silent strip. `.strict()` carries through the
+  // `.extend()`/`.omit()`/`.partial()` chains below, so the hire and update schemas inherit it.
+}).strict();
 
 export type CreateAgent = z.infer<typeof createAgentSchema>;
 
--- a/server/src/__tests__/agent-permissions-routes.test.ts
+++ b/server/src/__tests__/agent-permissions-routes.test.ts
@@ -958,6 +958,136 @@ describe.sequential("agent permission routes", () => {
     expect(mockAgentService.list).not.toHaveBeenCalled();
   });
 
+  it("rejects an unknown body field on agent update instead of silently dropping it (LOOA-364)", async () => {
+    const app = await createApp({
+      type: "board",
+      userId: "board-user",
+      source: "local_implicit",
+      isInstanceAdmin: true,
+      companyIds: [companyId],
+    });
+
+    const res = await requestApp(app, (baseUrl) => request(baseUrl)
+      .patch(`/api/agents/${agentId}`)
+      .send({ description: "Social Editor" }));
+
+    expect(res.status, JSON.stringify(res.body)).toBe(400);
+    expect(res.body.error).toContain("description");
+    expect(res.body.details.unknownFields).toEqual(["description"]);
+    expect(mockAgentService.update).not.toHaveBeenCalled();
+  });
+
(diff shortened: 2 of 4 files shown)


nguyenm7:
@greptileai review

PR #10057, opened 2026-07-23
fix(auth): reject JWTs after agent runs finish

cucurigoo:
## Thinking Path

> - Local adapter runs receive short-lived signed JWTs so the agent can call the Paperclip API as the intended run identity
> - Those JWTs include a run ID, but authentication previously validated only the signature, claims, and agent record
> - A token therefore remained usable after its heartbeat run reached a terminal state, until its time-based expiry
> - Run-scoped credentials should stop authorizing requests when their run is no longer active
> - This pull request validates the signed run against the database and accepts it only while that exact company/agent run is `running`
> - The benefit is immediate run-bound revocation without adding a token denylist or migration

## Inline bug report

No public issue was found, so this follows the bug-report template inline.

### Pre-submission checklist

- [x] I searched existing open and closed issues and this is not a duplicate.
- [x] I reproduced this on current `master`.
- [x] I confirmed the behavior originates in Paperclip's local-agent JWT middleware, not an adapter provider or local configuration.

### What happened?

A valid local-agent JWT remains accepted after its associated heartbeat run finishes. The request continues to receive an `agent_jwt` actor until the token's `exp` time even though its signed run ID is no longer active.

### Expected behavior

The middleware should reject the JWT as soon as the exact signed company/agent run is no longer `running`.

### Steps to reproduce

1. Create a local-agent JWT for a heartbeat run.
2. Transition the run from `running` to a terminal status such as `succeeded`.
3. Send another API request with the still-unexpired JWT.

`master` at `e55d702916c4` before this change.

### Deployment mode

Self-hosted server.

### Installation method

Built from source with pnpm.

### Agent adapter(s) involved

Hermes; the bug is in core local-agent JWT middleware and can affect any local adapter using this credential path.

### Database mode

External Postgres.

### Access context

Agent bearer JWT.

### Node.js version

`v22.22.3`.

### Operating system

Linux 6.8 x86_64.

### Relevant logs or output

The focused regression received an `agent_jwt` actor after the fixture run changed to `succeeded` before the fix; it receives the anonymous actor afterward. No credential or instance-local output is included.

### Privacy checklist

- [x] I reviewed the text for usernames, paths, API keys, tokens, company names, and other private instance data.

## What Changed

- Resolve the exact signed run by run ID, company ID, and agent ID during local-agent JWT authentication.
- Accept the JWT only while that run has status `running`.
- Reuse the validated run row for legacy `responsible_user_id` fallback rather than issuing a second lookup.
- Add regressions proving that an otherwise valid JWT becomes anonymous before queue claim, after its run succeeds, or after its run row disappears.

## Verification

- `pnpm exec vitest run src/__tests__/agent-auth-middleware.test.ts` — 10 passed.
- `pnpm --filter @paperclipai/server typecheck` — passed.
- `git diff --check` — passed.
- The new regression failed before the implementation by receiving an `agent_jwt` actor and passed afterward.

## Risks

- Every local-agent JWT request now performs one indexed heartbeat-run lookup. This replaces the existing lookup used by legacy claim-less JWTs and adds one lookup for modern JWTs.
- Queued runs remain unauthorized by design: `claimQueuedRun` atomically persists `status = running` before `executeRun` can mint the JWT and invoke the adapter.
- Authorization ends immediately when a run leaves `running`; this is intentional because the adapter process has already finished before terminal finalization.
- No schema migration or public API shape change.

## Model Used

- OpenAI Codex, `gpt-5.6-sol`, with repository tool use and red-green TDD.

## Checklist

- [x] I have updated relevant documentation to reflect my changes (not applicable: internal authentication behavior and regression coverage only)

### DIFF
--- a/server/src/__tests__/agent-auth-middleware.test.ts
+++ b/server/src/__tests__/agent-auth-middleware.test.ts
@@ -334,6 +334,31 @@ describe("agent auth middleware", () => {
     expect(res.body).toMatchObject({ type: "none", source: "none" });
   });
 
+  it("rejects a JWT for a run that has not been claimed from the queue", async () => {
+    const companyId = randomUUID();
+    const agentId = randomUUID();
+    const runId = randomUUID();
+    const { db } = createDbState({
+      agent: { id: agentId, companyId },
+      run: {
+        id: runId,
+        companyId,
+        agentId,
+        responsibleUserId: "user-claim",
+        status: "queued",
+      },
+    });
+    const token = createLocalAgentJwt(agentId, companyId, "codex_local", runId, "user-claim");
+
+    const res = await request(createApp(db))
+      .get("/actor")
+      .set("Authorization", `Bearer ${token}`)
+      .set("X-Paperclip-Run-Id", runId);
+
+    expect(res.status).toBe(200);
+    expect(res.body).toMatchObject({ type: "none", source: "none" });
+  });
+
   it("preserves signed skill_test JWT scope on the request actor", async () => {
     const companyId = randomUUID();
     const agentId = randomUUID();
--- a/server/src/__tests__/agent-auth-middleware.test.ts
+++ b/server/src/__tests__/agent-auth-middleware.test.ts
@@ -33,7 +33,13 @@ function createSelectChain(rowsForTable: (table: unknown) => unknown[]) {
 function createDbState(input: {
   agent: { id: string; companyId: string; status?: string };
   agentKey?: { id: string; agentId: string; companyId: string; keyHash: string; responsibleUserId?: string | null };
-  run?: { id: string; companyId: string; agentId: string; responsibleUserId?: string | null };
+  run?: {
+    id: string;
+    companyId: string;
+    agentId: string;
+    responsibleUserId?: string | null;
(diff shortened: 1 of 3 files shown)


cucurigoo:
Greptile follow-up: I traced the startup mechanism rather than widening the gate. `claimQueuedRun` atomically updates the heartbeat run from `queued` to `running` and returns that claimed row before `executeRun` proceeds. The JWT is minted later inside `executeRun`, immediately before adapter invocation. A queued run therefore cannot have a legitimately issued adapter token. I added an explicit queued-run rejection regression; the focused file now passes 10 tests, and server typecheck passes.

cucurigoo:
Independent review follow-up in af57f55b1: inactive/missing/mismatched run-scoped JWTs now fail explicitly with 401, including local_trusted mode, rather than falling through to the implicit board administrator. The active-run row is also revalidated against run/company/agent identifiers. The pipeline e2e now creates and cancels a real running heartbeat row instead of minting a token for a nonexistent run. Verification: 29 focused server tests passed, server typecheck passed, and the two-test pipelines tutorial Playwright file passed.

PR #9707, opened 2026-07-16
Allow manager-chain comments on review-held issues

gagecane:
## Thinking Path

> - Paperclip is the open source app people use to manage AI agents for work.
> - The runtime authorization layer decides which agents may read, comment on, or mutate issues.
> - Review-held work often needs a manager or CEO to leave an escalation or coordination comment without taking over the task.
> - Before this change, that non-decision comment path could be denied at the review boundary, blocking legitimate managerial coordination.
> - The fix must stay narrow because review decisions and status changes are governed actions.
> - This pull request allows manager-chain and legacy agent-creator actors to perform `issue:comment` on assigned issues while leaving `issue:mutate` denied without a separate grant.
> - The benefit is that escalation comments can land during review without bypassing checkout ownership or approval/decision authorization.

## Linked Issues or Issue Description

Bug report inline because no related public GitHub issue was found in the pre-PR search for `manager comment review-held issue authz`.

### Pre-submission checklist

- [x] I have searched existing open and closed issues and this is not a duplicate.
- [x] I am on the latest released version of Paperclip (or can reproduce on `master`).
- [x] I have confirmed the error originates in Paperclip itself, not in my agent adapter, API provider, or local configuration.

### What happened?

When a manager-chain or CEO-style agent tried to post a non-decision escalation comment on an assigned review-held issue, the server authorization boundary could deny the comment request even though the actor needed only comment authority, not ownership or mutation authority.

### Expected behavior

Manager-chain and legacy agent-creator actors should be able to add coordination comments to assigned issues, including review-held issues, while status decisions and issue mutations remain denied unless the actor separately has `issue:mutate` authority.

### Steps to reproduce

1. Create an issue assigned to a subordinate agent and put it in review.
2. Authenticate as a manager-chain actor that does not own the active checkout and lacks broad issue mutation authority.
3. POST a non-decision comment to `/api/issues/:id/comments`.
4. Observe that the comment path can be denied before this fix.
5. Try PATCHing the same issue status as that actor and confirm mutation remains denied after this fix.

### Paperclip version or commit

Reproduced against `master` before `ead2748c2a4070aa5dadd40d8eb07cbe317cc7bc`.

### Deployment mode

Self-hosted server / local dev server API authorization path.

### Installation method

Built from source.

### Agent adapter(s) involved

Not adapter-specific (core bug).

### Database mode

Embedded PGlite in tests; authorization behavior is not database-engine-specific.

### Access context

Agent bearer API key.

### Node.js version

Not environment-specific.

[...]

PR #7863, opened 2026-06-10
fix(issues): remove over-broad issue-mutation authz guard from comment-create route (GH#7389)

isak-ialogics:

PR #7638, opened 2026-06-06
feat(authz): add issues:comment:all and issues:patch:all permission grants

MagicalMiraOne:
## Thinking Path

> - Paperclip is the control plane for running AI agents as a supervised company.
> - Cross-agent coordination depends on issue writes staying least-privilege by default, with any exception made explicit and auditable.
> - This PR adds the narrow cross-issue grants that unblock linter-style agents such as COO without changing the default assignee-ownership model.
> - The first draft split the route-level grants correctly, but the shared `issue:mutate` boundary still accepted either cross-issue key, which made the authorization layer more permissive than the endpoint actually intended.
> - The same draft also mounted `POST /api/agents/:id/grants` without documenting it in the OpenAPI route inventory, which broke the serialized OpenAPI test shard.
> - Review also called out one redundant `enabled !== false` expression and the missing required PR template sections.
> - This pull request tightens the authorization boundary to the endpoint-specific permission key, documents the grants route in OpenAPI, adds regression coverage for both, and fills in the required PR metadata.
> - The benefit is that cross-issue grants stay explicit, endpoint-scoped, documented, and test-covered instead of relying on a looser shared boundary.

## Linked Issues or Issue Description

- Refs: #7490
- Refs: #8080
- No upstream GitHub issue exists for this exact PR. The underlying problem is that linter-style agents need a safe, explicit way to comment on or patch issues they do not own. The additive grants introduced here are the escape hatch, but they must remain endpoint-specific (`issues:comment:all` vs `issues:patch:all`) and the new board grants route must be part of the published OpenAPI surface so the server route inventory stays accurate.

## What Changed

- Thread the required `PermissionKey` through `decideIssueAccess` so the `issue:mutate` boundary only approves the endpoint-specific cross-issue grant requested by the caller.
- Keep the route-layer cross-agent audit behavior in place while removing the overly broad "accept either key" fallback from `authorization.ts`.
- Replace `enabled !== false` with `enabled` in `POST /api/agents/:id/grants`.
- Register `POST /api/agents/{id}/grants` in `server/src/routes/openapi.ts`.
- Add regression coverage for the explicit grants route and for keeping comment vs patch grants endpoint-specific.

## Verification

- `pnpm exec vitest run server/src/__tests__/openapi-routes.test.ts server/src/__tests__/agent-permissions-routes.test.ts server/src/__tests__/issue-agent-mutation-ownership-routes.test.ts server/src/__tests__/authorization-service.test.ts`
- Result: `4 passed`, `144 passed`

[...]

PR #10114, opened 2026-07-23
fix(authz): allow any company agent to comment on any company issue

ericdfields:
## Thinking Path

> - Authorization for agent actions is decided in `server/src/services/authorization.ts` (`decideBase`), where `issue:comment` and `issue:mutate` share a resolution block
> - Standard-trust agents could comment only on issues they own, issues with no agent assignee, or issues where they hold a mention grant; every other case fell through to `deny_missing_grant`
> - This blocks ordinary cross-agent collaboration — e.g. a Designer agent cannot comment on an issue owned by another agent to ask a question or hand off
> - This pull request adds an unconditional allow for `issue:comment` (only) so any active same-company agent may comment on any company issue
> - The benefit is friction-free cross-agent discussion, while `issue:mutate` (reassign/close/status) stays locked to owners/grantees

No public GitHub issue. Describing inline (bug/gap):

**Problem:** A standard-trust agent that is not the assignee, not mentioned, and holds no explicit grant receives `403 deny_missing_grant` when trying to POST a comment to another agent's issue in the same company. There is no way for peer agents to comment on each other's issues without a per-issue mention grant.

**Expected:** Any active company agent can comment on any issue in the same company. Mutating actions (reassign, close, status change) must remain restricted to the assignee or an explicit grant holder.

- [x] I searched the open PR list for similar/duplicate work before opening this. Closely related open PRs in this area (narrower or grant-based approaches to the same friction): #8066 (split issue:comment from issue:mutate), #7638 (`issues:comment:all` grant), #7863 (remove mutation guard from comment route), #7960 (creator can comment on reassigned issues), #9707 (manager-chain comments). This PR takes the simplest path — an unconditional same-company `issue:comment` allow — and defers to maintainers on whether a grant-scoped approach is preferred.

## What Changed

- `server/src/services/authorization.ts` (`decideBase`): after the existing mention-grant check, inside the `issue:comment || issue:mutate` block, added an unconditional `allow({ reason: "allow_company_agent" })` guarded on `input.action === "issue:comment"`. `issue:mutate` is untouched and still requires ownership or a grant.
- `server/src/__tests__/authorization-service.test.ts`: added a case asserting a standard-trust non-owner agent is allowed `issue:comment` (`allow_company_agent`) but denied `issue:mutate` on the same issue.

## Verification

Ran the affected suites against embedded Postgres:

```
  src/__tests__/authorization-service.test.ts \
  src/__tests__/issue-comment-reopen-routes.test.ts \
  src/__tests__/issue-agent-mutation-ownership-routes.test.ts
```

- `authorization-service.test.ts` — 50 passed (incl. the new cross-agent comment case)
- `issue-comment-reopen-routes.test.ts` + `issue-agent-mutation-ownership-routes.test.ts` — 138 passed (mutation ownership intact)

## Risks

- **Behavioral shift, not a migration.** This broadens `issue:comment` for standard-trust agents across the whole company. It is purely permissive — it grants access, never revokes it.
- **Low-trust agents unaffected.** `decideLowTrustAccess` runs earlier in `decideBase` and short-circuits `issue:comment` for low-trust-preset agents (boundary + mention only), so their sandbox is preserved; this block is reached only by standard-trust agents. Confirmed by the existing `deny_low_trust_boundary` test still passing.
- **`issue:mutate` unchanged** — the new allow is guarded on `issue:comment`, so reassign/close/status still require ownership or a grant.
- Deployments that rely on per-issue comment restrictions between peer agents would see those restrictions relaxed. Overlaps with in-flight authz PRs (see linked list) — maintainers may prefer a grant-scoped variant.

## Model Used

Claude Opus 4.8, model id `claude-opus-4-8`, 1M context window, extended thinking, with tool use / code execution (agentic coding session).

### DIFF
--- a/server/src/__tests__/authorization-service.test.ts
+++ b/server/src/__tests__/authorization-service.test.ts
@@ -1219,6 +1219,46 @@ describeEmbeddedPostgres("authorization service", () => {
     })).resolves.toMatchObject({ allowed: false, reason: "deny_low_trust_boundary" });
   });
 
+  it("allows any standard-trust company agent to comment on an issue owned by another agent without a grant", async () => {
+    const company = await createCompany(db, "CrossAgentComment");
+    const ownerAgent = await createAgent(db, company.id, { role: "engineer" });
+    const commenterAgent = await createAgent(db, company.id, { role: "designer" });
+    const issue = await createIssue(db, company.id, {
+      title: "Owner-only issue",
+      assigneeAgentId: ownerAgent.id,
+    });
+
+    const authorization = authorizationService(db);
+    const resource = {
+      type: "issue",
+      companyId: company.id,
+      issueId: issue.id,
+      projectId: issue.projectId,
+      assigneeAgentId: ownerAgent.id,
+      status: issue.status,
+    } as const;
+    const actor = {
+      type: "agent",
+      agentId: commenterAgent.id,
+      companyId: company.id,
+      source: "agent_key",
+    } as const;
+
+    // A non-owning, non-mentioned agent with no explicit grant may comment.
+    await expect(authorization.decide({
+      actor,
+      action: "issue:comment",
+      resource,
+    })).resolves.toMatchObject({ allowed: true, reason: "allow_company_agent" });
+
+    // ...but mutation (reassign/close/status) still requires ownership or a grant.
+    await expect(authorization.decide({
+      actor,
+      action: "issue:mutate",
+      resource,
+    })).resolves.toMatchObject({ allowed: false });
+  });
+
   it("allows a mentioned non-assignee to comment when the mention author is the issue assignee", async () => {
     const company = await createCompany(db, "MentionCommentAssigneeGrant");
     const allowedProject = await createProject(db, company.id, "MentionAssigneeAllowed");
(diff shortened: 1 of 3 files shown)


ericdfields:
Closing as superseded. `master` now implements this PR's intent, gated on visibility rather than as a blanket allow.

`decideVisibleIssueWrite` default-opens `issue:comment` for any standard-trust agent that can *read* the issue, so a standard company agent can already comment on a grandparent or a sibling outside its direct-report grant. The red-team suite covers exactly that today, and it also keeps low-trust review runs contained — a property the blanket rule in this branch did not preserve on its own.

The rebase is also a no-op rather than a behaviour change. On current `master` this PR's allow-block lands immediately below:

```ts
// server/src/services/authorization.ts
if (visibleIssueWriteDecision) return visibleIssueWriteDecision;
```

That return is unconditional for `issue:comment`. Execution only reaches this point when `trustResolution.kind === "standard"` — `decideLowTrustAccess` returns `null` only for `"standard"`, and every other kind returns a decision that is returned earlier for `issue:comment`. `"standard"` is exactly the condition that makes `visibleIssueWriteDecision` non-null, so nothing below that line for `issue:comment` is reachable.

If the underlying intent was broader — commenting on issues the agent cannot *read* — that is a separate widening that has to sit above the default-open check and deserves its own review rather than a rebase of this branch.

PR #10183, opened 2026-07-24
feat(execution-receipts): W2 receipts schema + hash-chain + emission (SAG-7616/SAG-7667)

dmndbrp-oss:
## Summary
- New `execution_receipts` table (Drizzle schema + migration `0185_icy_thing.sql`): one row per `heartbeatRuns` row, per-company hash-chained (`contentHash`/`prevReceiptHash`/`chainSeq`), per the SAG-7632 plan design.
- Hash-chain via a shared `stableStringify`/`sha256Digest` (`server/src/services/canonical-hash.ts`, re-exported from `feedback-redaction.ts` for compatibility).
- Redaction via the existing `sanitizeRecord` util — never persists raw pre-redaction payloads.
- Emission wired into `heartbeat.ts` run terminal-status finalization: Tier-2 mandatory, Tier 0/1 optional (company opt-in, default off), fail-safe mandatory when risk tier is unresolved/null (`getSkillRiskTier` stubbed to `null` pending SAG-7615).
- Consolidated every `heartbeatRuns` terminal-status write path through one finalizer (or an explicit `emitExecutionReceipt` call), including two direct-cancel sites (stale scheduled retry, execution lock released on reassignment) that bypassed `setRunStatus`/`setRunStatusIfRunning` and would otherwise have silently skipped receipt emission.
- Query interface: `getReceiptsBySkillVersion`, `getReceiptChain`, `verifyReceiptChain` in `server/src/services/execution-receipts.ts`.

## Test plan
- [x] TDD tests in `server/src/__tests__/execution-receipts.test.ts` (schema round-trip, chainSeq/prevReceiptHash linkage over N inserts, no double-insert per runId, fail-safe emission gate unit tests, skill-version filtering) — 9/9 passing against embedded Postgres.
- [x] `packages/db` typecheck + migration safety/numbering checks pass.
- [x] `server` typecheck passes.
- [x] Existing `heartbeat-retry-scheduling.test.ts` / `heartbeat-lock-release-on-reassignment.test.ts` / `heartbeat-process-recovery.test.ts` / `heartbeat-run-status-payload.test.ts` (123 tests) still pass after the consolidation fix.
- [ ] QA adversarial tamper-chain + redaction tests — tracked as a blocked-on child issue (SAG-7669).

Implements SAG-7667 per the Director-endorsed, board-confirmed SAG-7632 plan.


dmndbrp-oss:
## Review: changes requested

Verified independently (checked out the branch, ran `pnpm run check:migrations`, ran the new test file + the 4 named pre-existing heartbeat suites against embedded Postgres, read every changed file). The schema, hash-chain, redaction reuse, gate logic, and query interface are solid, and the new test file genuinely is TDD (9/9 pass). But two defects break the ticket's core "can't be skipped" guarantee, both confirmed by direct repro against real Postgres — not test-fixture noise.

### 1. A third terminal-status write site still bypasses emission (confirmed live path)

`server/src/services/recovery/service.ts`, function `foldSourceResolvedStaleRun` (~line 1585, invoked from the "source-resolved watchdog fold" path at ~line 1974):

```js
const [updatedRun] = await tx
  .update(heartbeatRuns)
  .set({ status: finalRunStatus, finishedAt: input.now, ... })  // "succeeded" | "cancelled"
  .where(and(eq(heartbeatRuns.id, input.run.id), ..., eq(heartbeatRuns.status, "running")))
  .returning();
```

This is a completely separate module from `heartbeat.ts` — no `setRunStatus`/`setRunStatusIfRunning`, no `emitExecutionReceipt` import anywhere in the file. It fires whenever the recovery watchdog folds a stale `running` run because the issue it belonged to already resolved via other evidence. That run never gets a receipt. This is the same class of bug the PR already fixed twice (stale-retry cancel, lock-release cancel) — just in a file the audit didn't cover.

### 2. `emitExecutionReceipt` silently drops receipts on any `issue_id` FK mismatch

`execution-receipts.ts` inserts `issueId` (read straight from `run.contextSnapshot.issueId`) as a hard FK to `issues.id`. If that issue doesn't exist at insert time (deleted, or any staleness between when the run started and when it finalizes — the schema's own `onDelete: "set null"` shows the author expected this), the **entire insert throws** `PostgresError: ... violates foreign key constraint "execution_receipts_issue_id_issues_id_fk"`, and since `emitExecutionReceipt` is invoked fire-and-forget with only a `.catch(logger.error)`, the receipt for that run silently never exists — no test failure, no alert, nothing.

Reproduced directly: ran `heartbeat-retry-scheduling.test.ts` against real embedded Postgres and instrumented the catch block to print `error.cause`. 26 of 28 test-triggered emissions failed with exactly this FK violation (test fixtures set `contextSnapshot.issueId` to a `randomUUID()` not backed by a real `issues` row — which is exactly what a deleted/stale issue reference looks like in production). The PR's "123 pre-existing heartbeat tests still pass" claim is true but misleading: it's true *because* the failures are swallowed, not because emission actually worked. `onDelete: "set null"` only protects existing rows from a later delete — it does nothing for an insert whose referenced issue is already gone.

Fix: resolve `issueId` defensively (verify existence, or fall back to `null` on FK failure with a distinguishable log) rather than letting the whole receipt vanish. Given this is meant to be an audit trail whose selling point is "can't be skipped," swallow-on-any-error is the wrong failure mode — this should differentiate "expected null (skill not resolved)" from "insert actually failed" and either retry/null-out the offending column or surface prominently, not just log-and-vanish.

### Not blocking, but worth a follow-up
- `resolvePrimarySkillForRun` is a stub returning all-nulls — acknowledged by the PR as pending real skill/plugin attribution, fine as scoped.
- Optional REST route correctly deferred, per the ticket.

Both findings verified against real Postgres in the checked-out worktree (`~/paperclip-worktrees/sag-7667-execution-receipts`), not simulated. Requesting changes before merge — recommend a point-fix pass covering the `recovery/service.ts` gap and the FK-swallow failure mode, then a quick re-run of the same repro to confirm both are closed.

PR #10197, opened 2026-07-24
fix(facilitator): restore §3 run-productivity check; heartbeat-runs route exists

adacovsk:
Fixes AA-2895.

Facilitator INSTRUCTIONS §3 was descoped on a false premise: it asserted that no run-data API surface exists and instructed the agent not to probe `heartbeat-runs`. The route does exist — it is **company-scoped** (`/api/companies/:companyId/heartbeat-runs`, returns 200). Only the per-agent spelling the Facilitator tried (`/api/agents/{id}/heartbeat-runs`) 404s.

Run rows carry `error`/`errorCode`, `usageJson`/`resultJson`, `sessionIdBefore`/`sessionIdAfter`, and `processLossRetryCount` — so all three checks the step was originally written for (error runs, short-circuit runs, session rotation) are evaluable today.

## Changes
- `agents/facilitator/INSTRUCTIONS.md` — restore §3 with the real route and actual field names; note that `usageJson`/`resultJson` are null while running and on cancelled runs, so productivity is judged only on terminal rows. Keep the prior guidance that a 404 is a **sweep error**, not an empty result set.
- `skills/paperclip/references/api-reference.md` — document all eight heartbeat-run routes plus the row schema.

The api-reference addition is the actual anti-recurrence measure: the route being undocumented is what let a wrong premise harden into a standing instruction.


adacovsk:
Opened against the wrong remote by mistake — this belongs on a downstream fork, not upstream. Closing.

PR #10211, opened 2026-07-24
fix(heartbeat): expire runs stuck in queue beyond a bounded age

stubbi:
## Thinking Path

> - Agent runs are scheduled through a heartbeat queue: runs enter `queued`, then `startNextQueuedRunForAgent` dequeues them and `reapOrphanedRuns` cleans up strays
> - `reapOrphanedRuns` only sweeps `running` runs; its own comment says "queued runs are legitimately waiting; resumeQueuedRuns handles them"
> - But `resumeQueuedRuns` funnels every queued run through `startNextQueuedRunForAgent`, which returns early (no cancel) for a non-invokable agent (e.g. one paused after an auth failure)
> - So a run enqueued just before its agent becomes non-invokable is never started and never terminated; it can sit in `queued` indefinitely (observed: over three days)
> - This PR adds a bounded max-queue-age backstop so no run can be stranded in the queue forever, regardless of the stranding cause
> - The benefit is a self-healing safety net that terminates genuinely stuck queued runs with an honest, distinct error code

No public GitHub issue exists; describing inline following the bug report template. Related prior work (different approach, referenced for dedup): Refs #6163 (pre-enqueue terminal-issue skip + a bulk stale queued-run cancel API); this PR is a complementary periodic time-based reaper.

**What happened**

A run remained in `queued` status for over three days, never started and never cancelled, after its agent became non-invokable moments after the run was enqueued.

**Expected behavior**

A run that can never be dequeued should eventually be terminated with a clear, distinct reason rather than leaking in `queued` forever.

**Steps to reproduce**

Enqueue a run for an agent, make the agent non-invokable (for example a pause after an auth failure) before the run is dequeued, then let the periodic heartbeat sweep run. The reaper test added here deterministically reproduces the stranded-then-expired sequence.

**Deployment mode**

Cloud multi-tenant execution (heartbeat scheduler with the Kubernetes sandbox provider).

## What Changed

- `reapOrphanedRuns` now also cancels runs in `queued` older than `maxQueuedAgeMs` (default 24h; override `PAPERCLIP_HEARTBEAT_MAX_QUEUED_RUN_AGE_MS`), with error code `queue_expired` and a message stating how long the run waited
- Threaded the new option at the periodic caller alongside the existing running-run staleness threshold
- Updated the stale "queued runs are legitimately waiting" comment to reflect the new bounded wait
- Issue-lock release for issue-linked queued runs is intentionally delegated to the same-sweep `reconcileStrandedAssignedIssues()` / `enqueueWakeup` self-heal (documented inline), matching how `sweepStaleIssueLocks` handles stale locks

## Verification

- `cd server && npx vitest run src/__tests__/heartbeat-process-recovery.test.ts` — old queued run cancelled `queue_expired`; fresh queued run untouched; running-run reap unchanged; issue-linked queued run reconciled to a fresh run (not left dangling) after expiry
- `npx vitest run src/__tests__/heartbeat-stale-queue-invalidation.test.ts` — regression, unchanged

## Risks

Low risk. The reaper only acts on runs already past a 24h default age, a distinct terminal code (`queue_expired`) makes the action auditable, and running-run behavior is untouched. The one subtlety (issue-lock release delegated to the same-sweep reconciliation) is covered by a dedicated test.
## Model Used

## Checklist

### DIFF
--- a/server/src/__tests__/heartbeat-process-recovery.test.ts
+++ b/server/src/__tests__/heartbeat-process-recovery.test.ts
@@ -4114,6 +4114,98 @@ describeEmbeddedPostgres("heartbeat orphaned process recovery", () => {
     expect(run?.errorCode).toBeNull();
   });
 
+  it("cancels an issue-linked queue-expired run and releases the issue's execution lock via the same-sweep reconciliation", async () => {
+    const { companyId, agentId, runId, wakeupRequestId, issueId } = await seedRunFixture({
+      agentStatus: "idle",
+      runStatus: "queued",
+      // includeIssue defaults to true: the fixture links the issue's
+      // executionRunId (and checkoutRunId) to this run, per the wiring at
+      // seedRunFixture's `includeIssue !== false` branch above.
+    });
+    const staleCreatedAt = new Date(Date.now() - 2 * 60 * 60 * 1000);
+    await db
+      .update(heartbeatRuns)
+      .set({ createdAt: staleCreatedAt })
+      .where(eq(heartbeatRuns.id, runId));
+
+    const issueBeforeReap = await db
+      .select()
+      .from(issues)
+      .where(eq(issues.id, issueId))
+      .then((rows) => rows[0] ?? null);
+    expect(issueBeforeReap?.status).toBe("in_progress");
+    expect(issueBeforeReap?.executionRunId).toBe(runId);
+
+    const heartbeat = heartbeatService(db);
+    const result = await heartbeat.reapOrphanedRuns({ maxQueuedAgeMs: 60_000 });
+
+    expect(result.reaped).toBe(1);
+    expect(result.runIds).toEqual([runId]);
+
+    const run = await heartbeat.getRun(runId);
+    expect(run?.status).toBe("cancelled");
+    expect(run?.errorCode).toBe("queue_expired");
+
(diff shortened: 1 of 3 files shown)


stubbi:
Rebased onto current master `d293dd3d1` (new head `3c823658d`). Three resolutions were needed: master replaced the raw `setInterval` in `server/src/index.ts` with `startHeartbeatSchedulerInterval`, so the `heartbeatMaxQueuedRunAgeMs` const now sits just above that call; master's `appendRunEvent` no longer takes an explicit sequence argument, so the queue-expiry lifecycle event drops the `nextRunEventSeq` call; and master added a generalized `setRunStatusFromLive(runId, status, fromStatuses, patch)` guarded-update helper that `setRunStatusIfRunning` now delegates to, so `setRunStatusIfQueued` delegates to it as `["queued"]` instead of duplicating the compare-and-set (this also picks up master's `emitTerminalAgentTaskRun` on the transition). The concurrent-claim test also had to stop pinning the expiry query to "the 2nd `db.select` call" — master added several selects to `reapOrphanedRuns` — so it now identifies that query by the rows it returns. Verified: `@paperclipai/server` typecheck clean, `heartbeat-process-recovery.test.ts` passes (129/129).

PR #10212, opened 2026-07-24
fix(server): retry runs that hit a sandbox provider worker restart window instead of failing setup

stubbi:
## Thinking Path

> - Agent runs execute in sandbox environments acquired through provider plugins (e.g. the Kubernetes sandbox provider)
> - Lease acquisition happens during run setup, before the adapter executes
> - When a provider plugin's worker is momentarily unavailable (a server or plugin restart window), lease acquisition throws "Sandbox provider ... is installed via plugin ..., but its worker is not running."
> - The heartbeat setup path records that as a terminal `setup_failed`: no retry classifier matches the message, so the run dies instantly even though the worker returns seconds later
> - This PR classifies that transient condition as retryable infrastructure so the run is retried instead of being lost to a restart blip
> - The benefit is that routine restarts no longer produce spurious instant run failures

No public GitHub issue exists; describing inline following the bug report template.

**What happened**

During a brief sandbox-provider-worker restart window, several runs failed instantly with `setup_failed` ("... but its worker is not running."), while runs on the same agent moments earlier and later succeeded.

**Expected behavior**

A transient, self-healing worker-unavailable condition should schedule a bounded retry, not terminally fail the run.

**Steps to reproduce**

Trigger a run while the sandbox provider plugin worker is momentarily unavailable (a server or plugin restart). Lease acquisition throws the worker-not-running error and the run is finalized as `setup_failed` with no retry. The recovery test added here reproduces the classification path.

**Deployment mode**

Cloud multi-tenant execution (Kubernetes sandbox provider plugin).

## What Changed

- Added a dedicated, readable predicate that recognizes the transient sandbox-provider-worker-unavailable lease failure and treats it as retryable infrastructure, so the heartbeat schedules a bounded continuation retry instead of finalizing terminally
- The predicate is anchored to the full lease-failure phrasing (`is installed via plugin ... but its worker is not running`) so it cannot match the permanent "provider not installed" message emitted by config validation
- Added tests proving the readiness poll already waits the full deadline while the worker handle is absent or `starting` (registered-late coverage); no poll behavior change was needed

## Verification

- `cd server && npx vitest run src/__tests__/environment-runtime.test.ts` — poll exhaustion + registered-late cases
- `npx vitest run src/__tests__/heartbeat-process-recovery.test.ts` — worker-unavailable message schedules a retry; a non-matching permanent provider failure still escalates terminally (negative case)

## Risks

Low risk. The retry is bounded by the existing infrastructure-continuation attempt cap (max 3), the message match is narrow enough to exclude the permanent provider-not-installed failure (covered by a negative test), and no readiness-poll or lease-acquisition behavior changed.
## Model Used

## Checklist

### DIFF
--- a/server/src/__tests__/environment-runtime.test.ts
+++ b/server/src/__tests__/environment-runtime.test.ts
@@ -941,6 +941,180 @@ describeEmbeddedPostgres("environmentRuntimeService", () => {
     expect(workerManager.call).toHaveBeenCalledWith(pluginId, "environmentAcquireLease", expect.anything(), 31234);
   });
 
+  it("throws a worker-not-running error once the readiness deadline is exhausted", async () => {
+    const pluginId = randomUUID();
+    const { companyId, environment: baseEnvironment, runId } = await seedEnvironment();
+    const providerConfig = {
+      provider: "fake-plugin",
+      image: "fake:test",
+      timeoutMs: 1234,
+      reuseLease: false,
+    };
+    const environment = {
+      ...baseEnvironment,
+      name: "Never Running Plugin Sandbox",
+      driver: "sandbox",
+      config: providerConfig,
+    };
+    await environmentService(db).update(environment.id, {
+      driver: "sandbox",
+      name: environment.name,
+      config: providerConfig,
+    });
+    await db.insert(plugins).values({
+      id: pluginId,
+      pluginKey: "acme.never-running-sandbox-provider",
+      packageName: "@acme/never-running-sandbox-provider",
+      version: "1.0.0",
+      apiVersion: 1,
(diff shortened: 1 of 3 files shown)


stubbi:
Cross-reference: #10262 changes the error text for a *missing* plugin worker manager so it no longer matches `isSandboxProviderWorkerUnavailableFailureMessage`. That is intended, since a process with no manager can never observe a running worker and retrying cannot succeed. A genuine mid-restart worker still matches and still retries. If this PR lands after #10262, the "transient sandbox provider worker restart" test here needs to wire a manager whose `isRunning` returns false rather than passing no manager at all. Details and the exact snippet are in #10262.

PR #10213, opened 2026-07-24
fix(adapter-utils): tolerate transient reads in the process-session bridge poll loop

stubbi:
## Thinking Path

> - Agent runs on remote sandboxes stream ACP events through a process-session bridge: a remote writer publishes numbered `events/NNNN.json` files and a local poll loop reads, delivers, and removes them
> - The writer is already robust (atomic tmp+rename, serial write chain, ordered exit frame)
> - The reader was not: the poll loop turned any single transient file-read failure into a fatal error that tore down the whole session
> - One failed `base64 < events/NNNN.json` (a sandbox exec hiccup) therefore killed an otherwise healthy run and dropped every not-yet-forwarded event, including the child exit frame
> - This PR makes the reader tolerant of transient read failures while preserving strict event ordering and still failing fast on genuine crashes
> - The benefit is that a momentary exec-channel blip no longer terminates a healthy run

No public GitHub issue exists; describing inline following the bug report template.

**What happened**

A healthy run died when a single transient read of one process-session bridge event file failed; the poll loop escalated that one failure to a fatal session teardown and dropped the remaining buffered events, including the child exit frame.

**Expected behavior**

A transient per-file read failure should be retried on the next poll, not kill the session; only genuine child error/exit events (or sustained failure) should terminate it.

**Steps to reproduce**

Run an ACP agent on a remote sandbox and cause one `base64 < events/NNNN.json` read to fail transiently (a sandbox exec-channel hiccup). Before this change the session is torn down immediately. The added tests reproduce the single-read failure and the persistent-failure escalation.

**Deployment mode**

Cloud multi-tenant execution (remote sandbox process-session bridge).

## What Changed

- A per-file read failure now stops the batch at that file (preserving strict sequence order) and retries it on the next poll
- The delivered-event watermark advances per successfully delivered event, so a mid-batch delivery/parse throw re-reads the tail rather than silently skipping it
- A cycle that stops early or fails mid-delivery counts toward `MAX_CONSECUTIVE_POLL_FAILURES = 5`; only after five consecutive failed cycles does the session escalate to the fatal error frame (idle cycles reset the counter)
- Remove failures are logged instead of silently swallowed; a delivered-then-remove-failed file is never re-delivered (name watermark)
- Genuine child error/exit events still terminate the session promptly

## Verification

- `npx vitest run packages/adapter-utils/src/execution-target-sandbox.test.ts` — mid-batch read failure preserves order and retries; persistent single-file failure escalates after 5 cycles (idle cycles do not); mid-batch malformed event re-reads the tail; remove failure logs without double delivery

## Risks

Low risk on the tolerance side (writer is untouched and atomic; watermark bounds memory and prevents double delivery). The main behavioral shift is that a truly wedged bridge now takes up to 5 poll cycles to fail instead of 1; this is bounded and intentional, and real child exit/error events are unaffected.
## Model Used

## Checklist

### DIFF
--- a/packages/adapter-utils/src/execution-target-sandbox.test.ts
+++ b/packages/adapter-utils/src/execution-target-sandbox.test.ts
@@ -156,6 +156,60 @@ describe("sandbox adapter execution targets", () => {
     };
   }
 
+  // Simulates transient sandbox shell failures for the process-session
+  // bridge's event polling: reading/removing a specific event file, or
+  // listing the events directory itself, can be made to fail on demand
+  // while every other shell call passes through to the real local runner.
+  function createProcessSessionEventFaultInjectingRunner(input: {
+    base: ReturnType<typeof createLocalSandboxRunner>;
+    failRead?: (fileName: string, attempt: number) => boolean;
+    failRemove?: (fileName: string, attempt: number) => boolean;
+    failList?: (attempt: number) => boolean;
+  }) {
+    const readAttempts = new Map<string, number>();
+    const removeAttempts = new Map<string, number>();
+    let listAttempts = 0;
+    const failResult = (message: string) => ({
+      exitCode: 1,
+      signal: null,
+      timedOut: false,
+      stdout: "",
+      stderr: message,
+      pid: null,
+      startedAt: new Date().toISOString(),
+    });
+    return {
+      execute: async (execInput: Parameters<ReturnType<typeof createLocalSandboxRunner>["execute"]>[0]) => {
+        const script = (execInput.args?.[1] ?? "").trim();
+        const readMatch = /^base64 < '([^']*\/events\/[^']+\.json)'$/.exec(script);
+        if (readMatch) {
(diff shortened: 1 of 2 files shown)


stubbi:
Rebased onto current master `d293dd3d1` (new head `152afa6fa`); the two commits were squashed into one so the poll-loop rewrite is resolved against master's reworked bridge exactly once. Two adaptations to master: the terminal-event early return inside the delivery loop was dropped, because master deliberately removed it so a `shutdownAck` landing in the same batch as `exit` is still delivered; and `readShutdownAckUntil`, master's new second caller of `readRemoteJsonFiles`, now passes `afterName: null` since it is independent of the poll loop's delivery watermark. Verified: `@paperclipai/adapter-utils` typecheck clean, `execution-target-sandbox.test.ts` (130/130), `execution-target.test.ts` + `sandbox-callback-bridge.test.ts` (62/62) pass.

PR #10224, opened 2026-07-24
Fix task-watchdog self-fencing and add fingerprint refresh

### LINKED ISSUE #10219: Task-watchdog run is 409-fenced by its own writes; no re-pin route exists
## Summary

A task-watchdog run pins a `stopFingerprint` at wake. The fingerprint is computed over watched-subtree state that **includes the watchdog's own writes**. The pin is read-only for the life of the run. So the watchdog's first mutation to the watched subtree rotates the fingerprint and every subsequent one returns 409 — including the child-issue creation that is the watchdog's only sanctioned way to restore a live path.

Observed with `paperclipai@2026.722.0` on run `7bec0134-90fd-41ee-adf5-b24b3de7647d` (watchdog `fda30ad7-a034-431b-a07f-8049161fb375`). Correct review, correct verdict, then fenced out of acting on it. The run ended `failed` with the watched subtree still stopped.

## Root cause

Paths below are relative to `@paperclipai/server/dist` in `paperclipai@2026.722.0`.

`stableStopFingerprint()` — `services/task-watchdogs.js:90-97` — is a SHA-256 over `{version: 1, companyId, watchedIssueId, leaves}`.

`leaves[]` is built at `services/task-watchdogs.js:200-217`, and each leaf carries `status`, `assigneeAgentId`, `blockerIssueIds`, `pendingInteractionIds`, `pendingApprovalIds`, `updatedAt`, **`latestCommentAt`**, `latestDocumentAt`, and `latestWorkProductAt`.

That is the bug in one line: **`latestCommentAt` is in the fingerprint.** A comment on any leaf of the watched subtree rotates it, and posting a triage comment on the watched issue is an explicitly allowed watchdog capability.

Enforcement path:

- `services/task-watchdog-scope.js:21-82` — `resolveTaskWatchdogMutationScope()` reads `stopFingerprint` from `heartbeatRuns.contextSnapshot.taskWatchdog.stopFingerprint`. Nothing in the request path writes it back, so no re-pin is possible.
- `services/task-watchdogs.js:1072-1102` — `revalidateMutationScope()` recomputes the classification live and allows only when `classification.state === "stopped" && classification.stopFingerprint === scope.stopFingerprint`.
- `routes/issues.js:2582-2603` — `assertFreshTaskWatchdogSourceMutation()` emits the 409.

The liveness write that gets rejected:

- `routes/issues.js:5834` — `POST /issues/:id/children`
- → `:5841` → `assertTaskWatchdogCreateIssueAllowed()` at `:2648`
- → `:2676` → `assertFreshTaskWatchdogSourceMutation()` → 409.

## Two distinct consequences

1. **Order-dependence.** Investigate → record findings → act is the correct reviewer sequence, and it is the sequence that loses. Recording the finding on the watched issue forfeits the remedy.
2. **One subtree write per run.** Creating the child issue itself rotates the fingerprint: the new child becomes a leaf and the watched issue leaves the leaf set. A watchdog facing two stopped leaves cannot remedy both in one run; the second child POST returns 409 even with perfect write ordering. The “child POST first” workaround does not cover this case.

## Existing related pattern

`routes/issues.js:2585-2586` already exempts the watchdog issue itself from revalidation:

```js
if (scope.watchdogIssueId && issue.id === scope.watchdogIssueId)
    return true;
```

`services/task-watchdogs.js:105-111` and `:125-126` exclude `originKind === "task_watchdog"` issues from the classified subtree entirely, so watchdog-issue writes are fingerprint-neutral. The concept that this actor's own writes are not external change already exists, but currently stops at the watchdog issue.

### Primary: attribute drift, then roll the pin forward

In `revalidateMutationScope()` (`services/task-watchdogs.js:1072`), when `classification.state === "stopped"` and `classification.stopFingerprint !== scope.stopFingerprint`, check the activity log for the included subtree since the run's pin. If every row carries `runId === <this run>`, the drift is self-inflicted: write the new fingerprint back to `heartbeatRuns.contextSnapshot.taskWatchdog.stopFingerprint` and return `{allowed: true}`. If any row carries a different actor, keep the 409.

This preserves what the fence is for—catching other actors mutating a stopped subtree—and requires no client change. The server already threads `actorRunId` / `watchdogActorRunId` through issue creation (`routes/issues.js:5898`, `:5903`), so attribution data is present.

### Secondary: add the refresh route the 409 already promises

The 409 body says “refresh the source state before mutating it,” but no route can do that. Add:

```
POST /api/issues/{watchedIssueId}/watchdog/refresh
```

The route should return the new fingerprint and re-pin the active run. This makes the error actionable and gives a recovery path when the change genuinely was external.

### Also worth considering: re-pin on checkout

`POST /api/issues/{id}/checkout` returns 200 during a fenced run but does not refresh run-scoped watchdog state. It is a natural place to refresh the pin.

## Adjacent finding: the platform-bug escape hatch is undiscoverable

`routes/issues.js:5644` skips `assertTaskWatchdogCreateIssueAllowed()` when `watchdogProductBugFollowUp` is truthy, so a fenced watchdog can still file a bug through:

```http
POST /api/companies/{companyId}/issues
Content-Type: application/json

{
  "title": "...",
  "description": "...",
  "watchdogDiscovery": {
    "kind": "platform_bug",
    "evidenceMarkdown": "..."
  }
}
```

(`kind` is `product_bug` or `platform_bug`, per `@paperclipai/shared/dist/constants.js:241`.)

Run `7bec0134` attempted precisely this filing and got 403 only because it omitted `watchdogDiscovery`; the request fell through to the ordinary agent path at `routes/issues.js:5624`. The escape hatch works, but nothing tells the watchdog it exists. Surfacing it in the watchdog wake payload would be a useful separate improvement. It creates a top-level issue outside the subtree, so it restores reporting, not subtree liveness.

## Reproduction

1. Create a watched issue with no live path and let a watchdog fire on it.
2. `POST /api/issues/{watchedIssueId}/comments` — 200. The fingerprint rotates.
3. `POST /api/issues/{watchedIssueId}/children` with an `assigneeAgentId` — 409:

```
Task-watchdog review is stale because the watched subtree stop fingerprint changed; refresh the source state before mutating it.
```

The response reports `currentState: "stopped"`.

Swap steps 2 and 3 and both succeed, demonstrating the order-dependence. With two stopped leaves, the first child creation rotates the subtree and the second child creation fails regardless of comment order.

ablac:
## Thinking Path

> - Paperclip is the control plane that keeps autonomous agent work governable and live
> - Task watchdogs review stopped issue subtrees and may restore an allowed live path
> - Each watchdog run pins a stopped-subtree fingerprint so external changes invalidate stale review authority
> - The fingerprint also includes the watchdog run's own comments and child issues, so a correct investigate → record → act sequence fences itself
> - The 409 response promises a refresh, but the server has no route that can re-pin the active run
> - This pull request attributes fingerprint drift to the active run, rolls self-authored drift forward, and adds the missing explicit refresh route
> - The benefit is that watchdogs retain protection from external mutation while being able to complete multiple sanctioned recovery writes in one run

Fixes: #10219

## What Changed

- Carry the watchdog heartbeat run id through persisted mutation scope resolution.
- Track when a run's stop fingerprint was pinned and use issue activity attribution since that pin to distinguish self-authored drift from external drift.
- Atomically advance the stored run fingerprint when every relevant activity row belongs to the same watchdog run.
- Add `POST /api/issues/:issueId/watchdog/refresh` for an explicit, scoped re-pin after external drift, with activity logging and capability metadata.
- Add scheduler and route regressions for self-authored comments, repeated subtree writes, external refresh, persisted context updates, and audit records.
- Document the watchdog refresh authority and REST endpoint in the V1 implementation contract.

## Verification

- `.\node_modules\.bin\tsc.cmd -p server/tsconfig.json --noEmit`
- `.\node_modules\.bin\vitest.cmd run server/src/__tests__/task-watchdogs-scheduler.test.ts` — 17 passed
- `.\node_modules\.bin\vitest.cmd run server/src/__tests__/issue-watchdogs-routes.test.ts` — 12 passed
- `git diff --check`

## Risks

- Low migration risk: this changes only JSON context data and adds no schema migration.
- Automatic re-pinning deliberately fails closed when there are no attributable activity rows or any relevant activity is from a different run.
- The explicit refresh route can accept genuinely external stopped-state drift, but only for the authenticated task-watchdog run, its persisted watchdog, company, and watched root.
- Pin timestamps fall back to the heartbeat run start/create time for runs created before this change.

> This is a focused bug fix within the roadmap's shipped task-watchdog capability; it does not add a competing roadmap feature.

## Model Used

- OpenAI GPT-5 via the Codex agent environment (exact deployment revision and context-window size are not exposed), using reasoning, repository tools, GitHub tooling, code execution, and local test execution.

## Checklist

### DIFF
--- a/doc/SPEC-implementation.md
+++ b/doc/SPEC-implementation.md
@@ -620,6 +620,7 @@ Within the watched subtree, a watchdog run may perform only mutations that resto
 - reopen `done` or `cancelled` included issues only with explicit resume metadata and an audit comment when evidence shows the stopped disposition is wrong or incomplete
 - add, replace, or clear blockers on included issues when the blocker target is in the same company and the change makes the waiting path more accurate
 - set or refresh a one-shot monitor on an included issue when the current assignee owns the future check
+- explicitly refresh the run-scoped stopped fingerprint for the persisted watched root when new evidence has arrived
 - accept or reject eligible task-level plan confirmations as defined below
 - update the reusable watchdog issue itself to `done`, `in_review`, or `blocked` with the evidence for the watchdog decision
 
@@ -854,6 +855,7 @@ All endpoints are under `/api` and return JSON.
 - `GET /issues/:issueId/documents/:key/revisions`
 - `DELETE /issues/:issueId/documents/:key`
 - `POST /issues/:issueId/checkout`
+- `POST /issues/:issueId/watchdog/refresh` (task-watchdog run only; re-pins the current stopped fingerprint)
 - `POST /issues/:issueId/release`
 - `POST /issues/:issueId/admin/force-release` (board-only lock recovery)
 - `POST /issues/:issueId/comments`
--- a/server/src/__tests__/issue-watchdogs-routes.test.ts
+++ b/server/src/__tests__/issue-watchdogs-routes.test.ts
@@ -480,11 +480,97 @@ describeEmbeddedPostgres("issue watchdog routes", () => {
       .where(and(eq(issueWatchdogs.companyId, companyId), eq(issueWatchdogs.issueId, watchedChildId)));
     expect(nestedWatchdogs).toHaveLength(0);
 
+    const finding = await request(app)
+      .post(`/api/issues/${watchedChildId}/comments`)
+      .send({ body: "Record the watchdog finding before creating follow-up work." });
+    expect(finding.status, JSON.stringify(finding.body)).toBe(201);
+
     const allowedChild = await request(app)
       .post(`/api/issues/${watchedChildId}/children`)
-      .send({ title: "Allowed watched child" });
+      .send({ title: "Allowed watched child", status: "done" });
     expect(allowedChild.status, JSON.stringify(allowedChild.body)).toBe(201);
     expect(allowedChild.body.parentId).toBe(watchedChildId);
+
+    const secondAllowedChild = await request(app)
+      .post(`/api/issues/${watchedChildId}/children`)
+      .send({ title: "Second allowed watched child" });
+    expect(secondAllowedChild.status, JSON.stringify(secondAllowedChild.body)).toBe(201);
+    expect(secondAllowedChild.body.parentId).toBe(watchedChildId);
+  });
+
+  it("lets a watchdog run explicitly refresh its pin after external stopped-subtree drift", async () => {
+    const companyId = await seedCompany();
+    const watchdogAgentId = await seedAgent(companyId, { name: "Refresh Watchdog" });
+    const watchedRootId = await seedIssue(companyId, {
+      title: "Watched root",
+      identifier: "WDOG-REFRESH",
+      status: "blocked",
+    });
+    const watchdogIssueId = await seedIssue(companyId, {
+      title: "Reusable watchdog issue",
+      parentId: watchedRootId,
+      assigneeAgentId: watchdogAgentId,
+      originKind: "task_watchdog",
+      originId: watchedRootId,
+    });
+    const runId = await seedWatchdogRun({
+      companyId,
+      watchdogAgentId,
+      watchedIssueId: watchedRootId,
+      watchdogIssueId,
+    });
+    const boardApp = createApp(companyId);
+    const externalComment = await request(boardApp)
+      .post(`/api/issues/${watchedRootId}/comments`)
+      .send({ body: "External evidence arrived after the watchdog wake." });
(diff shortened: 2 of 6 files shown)


satruxag:
Added an independent reproduction on 10219, including a `pending_first_run`/`null` transient this PR's refresh path may not cover. Happy to test this branch against it.

satruxag:
Added an independent reproduction on #10219, including a / transient this PR's refresh path may not cover. Happy to test this branch against it.

PR #10254, opened 2026-07-25
fix(adapters): let a board member with no company yet read the adapter inventory

stubbi:
## Thinking Path

> - Paperclip is the open source app people use to manage AI agents for work
> - Onboarding is where a new board member names their company, hires their first agents, and picks the harness those agents run on; the picker's options come from `GET /api/adapters`, whose `disabled` flags reflect the instance's declared registry (`PAPERCLIP_ADAPTERS`)
> - That route is gated on `assertBoardOrgAccess`, which requires at least one company membership, but the wizard reads it BEFORE the user's first company exists, so a brand-new board member always gets a 403
> - The failure is silent: `useDisabledAdaptersSync` just ends up with an empty disabled set, so the picker offers harnesses the instance curated out
> - Picking one produces a first run that dies at lease time with `Adapter "..." is not in the configured adapter registry` — the worst possible first experience, and nothing in the UI connects it back to the 403
> - This pull request gates the inventory read on `assertBoard` instead, so any authenticated board member can read it, while every mutating adapter route keeps its instance-admin gate
> - The benefit is that the harness picker tells new users the truth on their very first session, instead of setting them up for a first run that cannot succeed

## Linked Issues or Issue Description

No existing issue; describing it here per the bug report template.

**What happened**

On a multi-tenant deployment, every fresh signup emitted `GET /api/adapters -> 403` from the onboarding wizard (17 occurrences across 4 accounts in one day of client telemetry). One of those users then picked `cursor_cloud` in the wizard — an adapter the instance had disabled through `PAPERCLIP_ADAPTERS` — and their first run failed:

```
Failed to acquire lease for environment "Kubernetes Sandbox" (sandbox): Adapter "cursor_cloud" is not in the configured adapter registry
```

**Expected behavior**

the wizard only offers adapters the instance actually runs, which is what the `disabled` flags in `GET /api/adapters` exist to express.

**Steps to reproduce**

1. Declare a curated registry, e.g. `PAPERCLIP_ADAPTERS` with `claude_local` enabled and `cursor_cloud` absent; start the server (`reconcileAdapterAvailability` disables everything undeclared).
2. Sign up a brand new board user who has no company yet.
3. Open onboarding. `GET /api/adapters` returns 403 (`Company membership or instance admin access required`).
4. The harness picker offers `cursor_cloud`; selecting it produces a first run that fails at lease time with the message above.

**Paperclip version or commit**

master (`4c55f0d8d`).

## What Changed

[...]

PR #10256, opened 2026-07-25
fix(agents): refuse to hire onto an adapter this instance cannot run

stubbi:
## Thinking Path

> - Hiring an agent means choosing a harness (adapter) for it, and an instance can declare which harnesses it actually runs through `PAPERCLIP_ADAPTERS`, which `reconcileAdapterAvailability` turns into a disabled set at boot
> - The hire and create routes validate the adapter type with `assertKnownAdapterType`, which only asks whether the adapter is REGISTERED — a disabled adapter passes
> - So an agent can be created on a harness the instance cannot run, and the failure only appears later, per run, at lease time: `Adapter "..." is not in the configured adapter registry`
> - By then the error is in a run log, minutes after the choice, with nothing tying it back to the harness the user picked; the agent also keeps accepting work it can never do
> - This pull request validates the hire and create paths against the ENABLED set and refuses with a message that names the adapters that are available
> - The benefit is that an impossible choice fails at the moment it is made, in the words of the choice itself, instead of as a run failure the user cannot act on

No existing issue; describing it here per the bug report template.

**What happened**

On an instance with a curated registry, a company's Chief of Staff was hired on `cursor_cloud`, which that instance had disabled. The API accepted the hire. Its first assignment run then failed:

```
Failed to acquire lease for environment "Kubernetes Sandbox" (sandbox): Adapter "cursor_cloud" is not in the configured adapter registry
```

and its automation run sat in `queued` for hours afterwards. Nothing in the hire response, the agent detail view, or the agent's status explained that this harness could never run.

**Expected behavior**

hiring on an adapter the instance has disabled is refused at hire time, with a message naming the adapters that can be chosen.

**Steps to reproduce**

1. Start the server with a registry that omits an otherwise-registered adapter, e.g. `PAPERCLIP_ADAPTERS` listing `claude_local` but not `cursor_cloud`.
2. `POST /api/companies/:companyId/agents` with `{"name":"CoS","adapterType":"cursor_cloud"}`.
3. The agent is created (201). Every run it attempts fails at lease time with the message above.

master (`4c55f0d8d`).

## What Changed

- `server/src/routes/agents.ts`: adds `assertSelectableAdapterType`, which extends `assertKnownAdapterType` with an enabled-set check and throws `422 Adapter "<type>" is not available on this instance. Available adapters: <list>`. The hire (`POST .../agent-hires`) and create (`POST .../agents`) paths now use it.
- Routes that operate on an EXISTING agent keep `assertKnownAdapterType`, so an agent already running on a since-disabled adapter is unaffected — the same rule `listEnabledServerAdapters` already documents ("hidden from selection, still functional for agents that already use them").
- `server/src/__tests__/agent-adapter-validation-routes.test.ts`: mocks the adapter-plugin store's disabled set (so the test never writes to a real `~/.paperclip/adapter-settings.json`), and covers refuse-when-disabled (including that the message names the alternatives and that no agent is created) plus create-still-works-when-enabled.

## Verification

```
pnpm vitest run server/src/__tests__/agent-adapter-validation-routes.test.ts
```
13 tests pass, including the two new cases and the existing unknown-adapter-type test.

Manual: disable an adapter (`PATCH /api/adapters/:type {"disabled": true}` as an instance admin, or omit it from `PAPERCLIP_ADAPTERS` and restart), then POST an agent with that `adapterType` — 422 naming the available adapters, and no agent row is created.

## Risks

Low, and scoped to new selections:

- Automation that creates agents on a disabled adapter now gets a 422 where it previously got a 201 followed by runs that always failed. That is the intended behavior change, and the message names the valid choices.
- Existing agents, and every route that acts on an existing agent, are untouched.
- The enabled set comes from the same store `GET /api/adapters` already reports, so the API and the picker cannot disagree.

## Model Used

Claude Opus 5 (Anthropic), model id `claude-opus-5`, 1M context window, extended thinking, with tool use and code execution via Claude Code.

## Checklist

- [x] My branch name describes the change (`upstream/adapter-selection-guard`) and contains no internal ticket id
- [x] I have updated relevant documentation to reflect my changes (the new helper documents the selection-vs-existing-agent rule)

Related: #10254 makes the adapter inventory readable during onboarding, which is what lets the picker hide these adapters in the first place. This PR is the server-side backstop for the same failure.

### DIFF
--- a/server/src/__tests__/agent-adapter-validation-routes.test.ts
+++ b/server/src/__tests__/agent-adapter-validation-routes.test.ts
@@ -465,6 +465,40 @@ describe("agent routes adapter validation", () => {
     expect(mockAgentService.create).not.toHaveBeenCalled();
   });
 
+  it("refuses to switch an existing agent onto a disabled adapter", async () => {
+    const { registerServerAdapter } = await import("../adapters/index.js");
+    registerServerAdapter(externalAdapter);
+    mockAdapterPluginStore.getDisabledAdapterTypes.mockReturnValue(["external_test"]);
+
+    const app = await createApp();
+    const res = await requestApp(app, (baseUrl) =>
+      request(baseUrl)
+        .patch("/api/agents/11111111-1111-4111-8111-111111111111")
+        .send({ adapterType: "external_test" }),
+    );
+
+    expect(res.status, JSON.stringify(res.body)).toBe(422);
+    expect(String(res.body.error ?? res.body.message ?? "")).toContain(
+      'Adapter "external_test" is not available on this instance',
+    );
+    expect(mockAgentService.update).not.toHaveBeenCalled();
+  });
+
+  it("still lets an agent already on a disabled adapter be edited", async () => {
+    // Disabling an adapter must not make the agents that already use it
+    // uneditable — only NEW selections of it are refused.
+    mockAdapterPluginStore.getDisabledAdapterTypes.mockReturnValue(["codex_local"]);
+
+    const app = await createApp();
+    const res = await requestApp(app, (baseUrl) =>
+      request(baseUrl)
+        .patch("/api/agents/11111111-1111-4111-8111-111111111111")
+        .send({ adapterType: "codex_local", adapterConfig: { model: "gpt-5.4" } }),
+    );
+
+    expect(res.status, JSON.stringify(res.body)).toBe(200);
+  });
+
   it("still creates an agent on an adapter that is registered and enabled", async () => {
     const { registerServerAdapter } = await import("../adapters/index.js");
     registerServerAdapter(externalAdapter);
--- a/server/src/__tests__/agent-adapter-validation-routes.test.ts
+++ b/server/src/__tests__/agent-adapter-validation-routes.test.ts
@@ -11,6 +11,10 @@ const mockAgentService = vi.hoisted(() => ({
   update: vi.fn(),
 }));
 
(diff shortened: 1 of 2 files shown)


stubbi:
Thanks — the PATCH gap was real and is fixed in the latest commit: switching an existing agent's `adapterType` now goes through `assertSelectableAdapterType`, while keeping the agent's current adapter (even one since disabled) stays allowed, so disabling an adapter never makes the agents already on it uneditable. Covered by two new tests.

On the join-request approval path in `routes/access.ts`, I deliberately left it out of scope, and I'd rather not fold it in here:

- That path does not let a user *pick* a harness from the product's picker. The adapter type arrives from an external agent that asked to join (`requestType: "agent"`, `adapterType` + `agentDefaultsPayload`), and the documented values for it are the gateway adapters — `openclaw_gateway`, `hermes_gateway`. The approver is accepting a specific external runtime, not choosing among the instance's harnesses.
- Those gateway adapters do not execute through the instance's sandbox at all — they talk to a runtime the joining agent already runs. So "not in the declared execution registry" does not imply "cannot run" for them the way it does for a local harness, and an instance that curates its sandbox registry down to a few local adapters would start rejecting join requests that work today. On our own deployment the enabled set is five `*_local` adapters, so enforcing there would break a live external-agent flow.

If maintainers want that path covered too, I think it wants its own predicate ("is this adapter runnable *here*, given how it executes") rather than reusing the picker's enabled-set, and I'm happy to open a follow-up PR for it.

PR #10259, opened 2026-07-25
feat(claude-local): output-inactivity monitor and run-level wall-clock cap

YukiCrisp:
## Thinking Path

> - The claude_local adapter runs `claude` as a child process with `--output-format stream-json` and holds a concurrency slot for the lifetime of that process
> - When the child stops emitting events but does not exit, the run is indistinguishable from one that is still working — nothing in the adapter notices
> - On an unattended instance that slot stays held until the platform-level silent-run safety net fires, which can be hours; the agent looks busy the whole time
> - codex-local already ships `output-inactivity-monitor.ts` for exactly this failure mode, so the gap is adapter parity rather than a new concept
> - This pull request ports that protection to claude-local, plus a run-level wall-clock cap for the case where output keeps trickling but the run never converges
> - The benefit is that a hung claude run fails fast with a specific error code and gets retried, instead of silently occupying a slot

No existing issue — describing the bug inline per `CONTRIBUTING.md` (path B), following `.github/ISSUE_TEMPLATE/bug_report.yml`:

**What happened**

A `claude` child process stops emitting stream-json events but does not exit. The adapter keeps waiting, the run stays `running`, and the agent's concurrency slot stays occupied. Nothing in claude-local detects the stall.

**Expected behavior**

The adapter notices the stall, terminates the child, and surfaces the run as failed with a specific error code so it can be retried instead of holding its slot.

**Steps to reproduce**

1. Start a claude_local run.
2. Stop the child from producing output without exiting it — e.g. `kill -STOP` the `claude` process, or a provider-side stall mid-stream.
3. Observe that on `master` the run stays `running` indefinitely from the adapter's point of view.

`4c55f0d8d`

**Deployment mode**

`local_trusted`

**Agent adapter(s) involved**

Claude Code (claude_local)

## What Changed

- Add `packages/adapters/claude-local/src/server/output-inactivity-monitor.ts`: a monitor that tracks the last parsed stdout event and fires after a configurable idle window.
- Wire it into `execute.ts`: on fire, SIGTERM the process group, wait 5s, then SIGKILL. The run resolves as failed with `errorCode: claude_output_inactivity_monitor` and a `monitor: no claude output for {N}m {S}s` message, and the kill diagnostic is flushed to the run log before the run resolves.
- Suspend the deadline while a dispatched tool is still in flight — silence is expected there — and re-arm once every tool has resolved.
- Arm the deadline from `onSpawn` rather than at construction, so a slow execution-target startup can never fire the monitor before a child exists to terminate.
- Add `runWallClockTimeoutSec`, a run-level cap for runs that keep emitting output but never converge.
- Both settings live on `adapterConfig`, default to a sane value (15m / unset), accept `null` to disable, and fall back to the default on a non-positive value — logging that decision to stderr so it is visible in run logs.
- Document both fields in the adapter's `agentConfigurationDoc`.
- Add unit and subprocess-integration tests, including regression coverage for the spawn-gating case above.

## Verification

```
npx tsc --noEmit -p packages/adapters/claude-local/tsconfig.json   # clean
npx vitest run packages/adapters/claude-local/src/server/
```

- With this branch: **107 passed**, 2 failed.
- On `origin/master` without this branch: **75 passed**, 2 failed.

The same 2 tests (`execute.remote.test.ts > prepares the workspace…` and `test.probe.test.ts > classifies rate-limit/overload failures…`) fail identically before and after, so they are pre-existing on `master` in this environment (macOS 26.5.1, Node v22.23.1) and unrelated to this change. This branch adds 32 passing tests and changes no existing result.

## Risks

- **Low-to-moderate, and opt-out.** The monitor kills a child process, so a mis-tuned timeout could terminate a genuinely slow run. Mitigations: the default window is 15 minutes; the timer resets on *any* parsed stream-json event, including assistant turns that only carry thinking output; it stays suspended while a dispatched tool has not returned; and `null` disables it entirely for known-slow workloads.
- The deadline now starts at spawn rather than at construction, so setup time is no longer counted against the window. This makes the effective timeout slightly more generous than in the first revision of this PR — deliberately, since the alternative failed runs that were about to succeed.
- No migrations, no schema changes, no changes to shared types or validators — the entire config surface is adapter-local.
- Termination is SIGTERM-first with a 5s grace before SIGKILL, so a child that can flush and exit cleanly gets the chance to.

## Model Used

Claude Opus 5 (`claude-opus-5`), 1M context window, adaptive/extended thinking enabled, with tool use (filesystem, shell, git). Used to port the feature onto current `master`, separate it from unrelated local changes in the same files, verify before/after test baselines, and act on the Greptile review.

## Checklist

- [x] I have run tests locally and they pass (2 pre-existing `master` failures documented above)


YukiCrisp:
Both P1s addressed — thanks, the first one was a real bug and the repro was exact.

**"Monitor fires before child spawn"** — fixed in `f0d961aa0`. The handle is now dormant until `noteSpawned()`, which `execute.ts` calls from `onSpawn` once a pid and process group are known. This also corrects the semantics: the window measures "no output from a running claude", not "no output since we started preparing to run one". Three regression tests added, including a startup 100× longer than the timeout.

**"Sandbox monitor cannot terminate child"** — the symptom is real and is fixed in `b418b209a`, but I could not reproduce the stated cause, so flagging that part rather than silently accepting it.

The premise is that the sandbox runner supplies neither a PID nor a process-group ID. On this code path it always supplies both:

- `runChildProcess` spawns with `detached: process.platform !== "win32"` (`packages/adapter-utils/src/server-utils.ts:3203`), so the child leads its own process group.
- `resolveProcessGroupId()` returns `child.pid` whenever `pid > 0` off Windows (`server-utils.ts:76-79`), so `processGroupId === pid`.
- `onSpawn` is invoked **only** when `typeof child.pid === "number" && child.pid > 0` (`server-utils.ts:3211`).

So whenever `onSpawn` fires — and therefore whenever the monitor is armed — both fields are non-null. Sandboxed runs go through this same path with `bwrap` as the spawned command, so the recorded pid is the sandbox leader and `kill(-pgid)` tears down the group.

That said, the *shape* you describe was reachable for a different reason: `onFire` set `monitorFired = true` **before** checking for a kill target, so any fire without a target would have reported `claude_output_inactivity_monitor` while terminating nothing. `b418b209a` reorders it — with no terminable child the monitor logs that it is standing down and leaves the run alone, deferring to the platform-level silent-run safety net. Only a fire that actually signals a process maps to an inactivity failure now.

If you have a concrete path where `onSpawn` is called with a null pid, I would like to see it — that would be worth a separate fix in `adapter-utils` rather than a workaround here.

Verification: typecheck clean; `packages/adapters/claude-local/src/server/` at **107 passed / 2 failed**, versus **75 passed / 2 failed** on `origin/master`. The same two tests (`execute.remote.test.ts`, `test.probe.test.ts`) fail identically before and after on macOS.

PR #10262, opened 2026-07-25
fix(server): pass the plugin worker manager to the routes that dispatch heartbeat runs

stubbi:
## Thinking Path

> - Agent work runs in an environment, and a sandbox environment can be backed by a plugin provider (for example a Kubernetes sandbox provider plugin)
> - Resolving that provider requires the plugin worker manager, which is threaded from `app.ts` through the route factory, the service, the routine service, and finally the heartbeat service into the environment runtime
> - Three route factories that dispatch heartbeat runs were mounted without it, so those runs built a sandbox driver with `pluginWorkerManager: undefined` and could never acquire a lease
> - The failure surfaced as `but its worker is not running`, which sent debugging toward a healthy worker process instead of the missing dependency
> - This pull request threads the manager through those three routes and makes the two failure modes report distinct causes
> - The benefit is that built-in agent routines, company skill test runs, and issue tree restores can actually run, and a future wiring gap names itself

No existing issue. Following `.github/ISSUE_TEMPLATE/bug_report.yml`:

**What happened**

A heartbeat run dispatched by a built-in agent routine fails setup immediately with:

```
Failed to acquire lease for environment "Kubernetes Sandbox" (sandbox):
Sandbox provider "kubernetes" is installed via plugin
"paperclip.kubernetes-sandbox-provider", but its worker is not running.
```

**What I expected**

The run acquires a sandbox lease and executes, as it does when the same agent is woken through `agentRoutes` or `routineRoutes`.

**Steps to reproduce**

1. Install a sandbox provider plugin and set an environment to use it.
2. Enable built-in agents and provision one with a routine.
3. `POST /api/companies/:companyId/built-in-agents/:key/routines/:routineKey/run`.
4. The dispatched run fails setup with the message above while the plugin worker is running and serving other runs.

The same failure occurs via `companySkillRoutes` (skill test runs) and `issueTreeControlRoutes` (issue tree restore with `wakeAgents`).

**Additional context**

The plugin worker is healthy throughout. The tell is the duration: a genuinely absent worker polls for `pluginWorkerReadyTimeoutMs` (5s) before failing, whereas this path fails instantly. Observed failures ran 0.4s to 1.8s, never 5s.

Related but distinct: #10212 retries runs that hit a genuine worker restart window. That retry cannot help here, because a route with no worker manager never observes a running worker no matter how long it waits.

- [x] I have searched the GitHub PR list for similar PRs.

## What Changed

- `app.ts` passes `{ pluginWorkerManager: workerManager }` to `companySkillRoutes`, `builtInAgentRoutes`, and `issueTreeControlRoutes`, matching the sibling routes on the adjacent lines.
- `builtInAgentRoutes` and `builtInAgentService` accept an optional `pluginWorkerManager` and forward it to `routineService`, which already supported it.
- `companySkillRoutes` and `issueTreeControlRoutes` accept an optional `pluginWorkerManager` and forward it to `heartbeatService`.
- `environment-runtime.ts` checks `!pluginWorkerManager` before the `worker_unavailable` state. The specific "sandbox plugin workers are unavailable in this server process" message was previously unreachable, because `resolveSandboxProviderPlugin` maps a missing manager to `worker_unavailable` and that branch threw first.
- New `server/src/__tests__/plugin-worker-route-wiring.test.ts` pins the manager reaching the service and heartbeat for all three routes.
- `server/src/__tests__/environment-runtime.test.ts` gains one test for the wiring failure message and one asserting a stopped worker still reports as a stopped worker.

## Verification

```
cd server
pnpm run typecheck
npx vitest run src/__tests__/plugin-worker-route-wiring.test.ts
npx vitest run src/__tests__/environment-runtime.test.ts
npx vitest run src/__tests__/built-in-agent-routes.test.ts src/__tests__/built-in-agents.test.ts \
  src/__tests__/company-skills-routes.test.ts src/__tests__/company-skills-import-authz-routes.test.ts \
  src/__tests__/company-skill-import-boundary.test.ts src/__tests__/company-skills-detail.test.ts
npx vitest run src/__tests__/plugin-scoped-api-routes.test.ts src/__tests__/environment-probe.test.ts \
  src/__tests__/environment-routes.test.ts src/__tests__/environment-runtime-driver-contract.test.ts
```

Typecheck is clean and 101 tests pass across those suites.

Both new test groups were confirmed to fail with the fix reverted. Reverting only the `environment-runtime.ts` reorder makes the new runtime test fail with the exact production message, `Sandbox provider "fake-plugin" is installed via plugin "acme.reusable-sandbox-provider", but its worker is not running.` Reverting the route changes fails all four wiring cases.

## Risks

Low risk. Both new `options` parameters default to `{}`, so existing callers and any external consumers of these factories are unaffected. The change is additive dependency threading plus a reorder of two mutually exclusive error branches, with no behavior change for a correctly wired process. The reordered check only alters which message is thrown in a state that previously could not be reached with an accurate one.

## Model Used

Claude Opus 4.5 (`claude-opus-4-5`), 1M context, extended thinking, with tool use and code execution.

## Checklist

### DIFF
--- a/server/src/__tests__/environment-runtime.test.ts
+++ b/server/src/__tests__/environment-runtime.test.ts
@@ -1871,6 +1871,43 @@ describeEmbeddedPostgres("environmentRuntimeService", () => {
     });
   });
 
+  it("reports a missing plugin worker manager as a wiring failure, not a stopped worker", async () => {
+    const { companyId, environment, runId } = await seedReusablePluginSandboxLease();
+    const runtimeWithoutWorkerManager = environmentRuntimeService(db);
+
+    await expect(
+      runtimeWithoutWorkerManager.acquireRunLease({
+        companyId,
+        environment,
+        issueId: null,
+        heartbeatRunId: runId,
+        persistedExecutionWorkspace: null,
+      }),
+    ).rejects.toThrow(/sandbox plugin workers are unavailable in this server process/);
+  });
+
+  it("still reports a stopped worker as a stopped worker when the manager is wired", async () => {
+    const { companyId, environment, runId } = await seedReusablePluginSandboxLease();
+    const offlineWorkerManager = {
+      isRunning: vi.fn(() => false),
+      call: vi.fn(),
+    } as unknown as PluginWorkerManager;
+    const runtimeWithOfflineWorker = environmentRuntimeService(db, {
+      pluginWorkerManager: offlineWorkerManager,
+      pluginWorkerReadyTimeoutMs: 0,
+    });
+
+    await expect(
+      runtimeWithOfflineWorker.acquireRunLease({
+        companyId,
+        environment,
+        issueId: null,
+        heartbeatRunId: runId,
+        persistedExecutionWorkspace: null,
+      }),
+    ).rejects.toThrow(/its worker is not running/);
+  });
+
   it("releases a sandbox run lease from metadata after the environment config changes", async () => {
     const { companyId, environment, runId } = await seedEnvironment({
       driver: "sandbox",
--- a/server/src/__tests__/plugin-worker-route-wiring.test.ts
+++ b/server/src/__tests__/plugin-worker-route-wiring.test.ts
@@ -1,0 +1,98 @@
+import { beforeEach, describe, expect, it, vi } from "vitest";
+
+/**
+ * Built-in agent routines dispatch real heartbeat runs. Those runs acquire a
+ * sandbox lease, and a plugin-backed sandbox provider can only be resolved when
+ * the plugin worker manager reaches the environment runtime.
+ *
+ * The manager is threaded app.ts -> routes -> service -> routines -> heartbeat.
+ * When any hop drops it the runtime resolves the provider with no manager and
+ * every run fails setup with "its worker is not running", even though the
+ * worker is healthy. These tests pin both hops that are easy to miss.
+ */
+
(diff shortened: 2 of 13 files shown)


stubbi:
Heads up on an interaction with #10212, which is still open.

#10212 adds `isSandboxProviderWorkerUnavailableFailureMessage`, which classifies a run as retryable infrastructure by matching the error text:

```js
/sandbox provider .* is installed via plugin .* but its worker is not running/i
```

The second commit here stops a missing worker manager from producing that message, so once both land, a missing manager is no longer classified as retryable. That is the intended outcome: a process with no worker manager will never observe a running worker, so retrying it cannot succeed. A genuine mid-restart worker still produces the matched message and still retries.

#10212 also carries a test named "schedules an infra retry for a setup failure caused by a transient sandbox provider worker restart" that builds its heartbeat service with no worker manager. That shortcut only produced the restart message because of the conflation this PR removes. Whichever PR lands second should update that test to wire a manager whose `isRunning` returns false, which is what a mid-restart worker actually looks like:

```js
const restartingWorkerManager = {
  isRunning: vi.fn(() => false),
  call: vi.fn(),
} as unknown as PluginWorkerManager;
const heartbeat = heartbeatService(db, {
  environmentRuntime: environmentRuntimeService(db, {
    pluginWorkerManager: restartingWorkerManager,
    pluginWorkerReadyTimeoutMs: 0,
  }),
});
```

Verified against both changes together: the test passes and still exercises the retry classification.

PR #10263, opened 2026-07-25
fix(kubernetes): honor exec cwd, explain signal-shaped exit statuses, declare timeoutMs

stubbi:
## Thinking Path

> - Agent runs execute inside sandbox environments provided by sandbox-provider plugins; the first-party `kubernetes` plugin runs each company in its own namespace and execs commands into a Sandbox-CR pod
> - Callers hand every execute a working directory (`params.cwd`, always set from `remoteCwd` in `packages/adapter-utils/src/execution-target.ts`), but the Kubernetes exec path never applies it: the Kubernetes exec API has no `cwd` field and the plugin does not synthesize one, so commands run in whatever WORKDIR the runtime image happens to declare. Every other sandbox provider (daytona, e2b, modal, novita, exe-dev, cloudflare) honors it
> - The same path leaves `PluginEnvironmentExecuteResult.signal` undefined rather than reporting it, and a run that died on the container memory limit (exit 137, the most common Kubernetes failure) reaches the operator as a bare exit code with no explanation; exe-dev, cloudflare and novita all report the field explicitly
> - And `paperclip-server` reads `timeoutMs` off the driver config to size the RPC budget for lease lifecycle calls, but the plugin's config schema never declared it — zod strips unknown keys, so an operator-set value is dropped during normalization and the 30s default silently applies to a first dispatch that has to bootstrap a whole tenant namespace and pull the runtime image
> - This pull request applies `params.cwd` to the exec'd command, reports `signal` explicitly (always `null`, because the exec API cannot observe one) with a hedged stderr hint for the two exit statuses that carry operational meaning on Kubernetes, and declares `timeoutMs` in both the zod schema and the manifest `configSchema`
> - The benefit is that Kubernetes execution behaves like the other providers: relative paths resolve where the caller said they would, memory-limit kills are legible in run output without claiming a termination the API never reported, and the documented RPC-budget knob actually reaches the server

No existing public issue — described in-PR following the bug template. Refs #9002 (open, complementary: it fixes the host/plugin RPC *budget inversion*; this PR declares the config key the same budget is read from, which #9002 does not add).

**What happened?**

Three defects in the same execute path of `packages/plugins/sandbox-providers/kubernetes`:

1. `onEnvironmentExecute` builds its exec command with `wrapCommandWithEnv(baseExecCommand, params.env)` (`src/plugin.ts`) and never references `params.cwd`. The command therefore runs in the image's WORKDIR. It works today only because the shipped agent-runtime images declare a WORKDIR that happens to match the workspace directory — a latent break for any custom image, any `imageOverride`, and any caller that passes a cwd other than the workspace root.
2. Every result returned by `onEnvironmentExecute` omits `signal`, which the plugin protocol defines as `signal?: string | null`, so consumers cannot tell "no signal" from "this provider did not answer". Worse, a run killed on the container memory limit surfaces as a bare `exitCode: 137` with nothing in the run output explaining what that status means on Kubernetes.
3. `kubernetesProviderConfigSchema` does not declare `timeoutMs`. `resolvePluginSandboxRpcTimeoutMs` (`server/src/services/environment-runtime.ts`) and `resolvePluginExecuteRpcTimeoutMs` (`server/src/services/plugin-environment-driver.ts`) both read `config.timeoutMs`; without it every lifecycle RPC falls back to `DEFAULT_RPC_TIMEOUT_MS` (30s). `acquireLease` on a company's first dispatch creates the namespace, ServiceAccount, Role, RoleBinding, ResourceQuota, LimitRange and network policies, then creates the Sandbox CR and waits on an image pull. 30s is not enough, and because the schema strips undeclared keys, an operator who sets `timeoutMs` today has it removed from `normalizedConfig` with no error and no warning.

**Expected behavior**

`params.cwd` is honored, results report `signal` explicitly, an exit status that looks like a termination is explained rather than left bare, and `timeoutMs` survives config normalization.

## What Changed

- `wrapCommandWithEnv(command, env, cwd?)` now prefixes a shell-quoted `cd <cwd> && ` when a cwd is given, and `onEnvironmentExecute` passes `params.cwd`. The `cd` is chained with `&&` so a missing directory fails the exec loudly instead of silently running the command elsewhere. With neither env nor cwd the command is still returned unwrapped, exactly as before.
- Set `signal` explicitly on every `onEnvironmentExecute` return path. It is always `null` and deliberately never inferred: a Kubernetes exec ends in a `V1Status` carrying an `ExitCode` cause and no termination signal, and the pod's container status describes PID 1 rather than the exec'd process, so a `128 + N` status cannot be told apart from a command that chose the same status. Reporting one would state an unobserved termination as fact.
- Added a hedged stderr hint for exit 137 and exit 143 — the two statuses that carry real operational meaning on Kubernetes (memory-limit kill, termination/eviction) — in the same spirit as the existing `appendNetworkEgressDenyHint`. The hint says what the status usually means and that the API reports no signal, so it is not mistaken for proof.
- Declared `timeoutMs` in `kubernetesProviderConfigSchema` and in the manifest `configSchema`, so it survives normalization and is offered to operators. It is bounded by `HOST_MAX_RPC_TIMEOUT_MS` (900000), the ceiling `plugin-worker-manager` applies to every plugin RPC: a larger budget cannot take effect, and a warning would not reach the operator (the server forwards validation warnings only inside the error payload of a rejected config), so an oversized value is rejected at save time with an actionable message instead of being silently clamped. A test pins the manifest's JSON Schema `maximum` to the same constant.
- Documented `timeoutMs` in the provider README config table.
- Added unit tests for all three behaviors, including a new `test/unit/plugin-execute.test.ts` covering the exec path with mocked kube clients.

## Verification

- `pnpm --dir packages/plugins/sandbox-providers/kubernetes test` — 204 passed, 22 of them new. The 7 failures in `test/unit/file-sync.test.ts` are pre-existing on `master` in this environment and unrelated: those tests exercise `/proc/self/fd/<n>`, which does not exist on macOS (`realpath: /proc/self/fd/8: No such file or directory`). Verified by running the same suite on a clean checkout of `master`: identical 7 failures, 182 passed.
- `pnpm --dir packages/plugins/sandbox-providers/kubernetes typecheck` — passed.
- Each behavior was test-driven: the new tests were written first and observed failing (`expected [ 'ls', '-la' ] to deeply equal [ '/bin/sh', '-c', "cd '/workspace/repo' && exec 'ls' '-la'" ]`, `expected undefined to be null` for `signal`, `expected undefined to be 180000` for the normalized `timeoutMs`) before the implementation landed. The same applies to both review rounds: the "never claims a signal", stderr-hint, timeout-rejection and manifest-ceiling tests were each written and observed failing before their change.

## Risks

- Behavioral change: a command whose `cwd` does not exist in the sandbox now fails with the shell's `cd` error instead of silently running in the image WORKDIR. This is the intended fail-loud semantic and matches the other providers; callers already guarantee the directory exists (`ensureAbsoluteDirectory` in `packages/adapter-utils/src/execution-target.ts` creates or validates it, and rejects non-absolute paths, before any execute).
- The exit-status hint appends one line to stderr for exit 137 and exit 143 only, and is explicitly hedged, so it cannot be read as an observed signal. All other statuses leave stderr byte-identical.
- `timeoutMs` is optional with no default, so absent config keeps today's behavior exactly. No stored Kubernetes environment can carry the key today (the schema stripped it, and `validatePluginEnvironmentDriverConfig` persists the normalized config), so the new upper bound cannot invalidate an existing configuration. The 15 minute host ceiling is duplicated in the plugin as `HOST_MAX_RPC_TIMEOUT_MS`; if `MAX_RPC_TIMEOUT_MS` is ever raised, this bound has to be raised with it.
- The fast-upload interceptor path is deliberately left alone: it rewrites the caller's own script against the absolute target paths embedded in it, so it has no relative-path exposure.
- The `job` backend ignores `cwd` as it already ignores `command`/`args`/`stdin` (its entrypoint is baked into the Job spec); the comment there was updated to say so.

## Model Used

Anthropic Claude, exact model ID `claude-opus-5[1m]` (Opus 5, 1M context), extended thinking, with tool use and code execution via Claude Code.

## Checklist

### DIFF
--- a/packages/plugins/sandbox-providers/kubernetes/README.md
+++ b/packages/plugins/sandbox-providers/kubernetes/README.md
@@ -74,7 +74,7 @@ Common optional fields:
 | `serviceAccountAnnotations` | `{}` | Annotations applied to per-tenant ServiceAccount (e.g. IRSA `eks.amazonaws.com/role-arn`). |
 | `jobTtlSecondsAfterFinished` | `900` | Seconds after a Job completes before garbage-collection. |
 | `podActivityDeadlineSec` | `3600` | Hard ceiling on a single run's wall-clock time. |
-| `timeoutMs` | `30000` (server default) | RPC budget paperclip-server grants lease lifecycle calls. Raise it (e.g. `180000`) when a company's first dispatch has to bootstrap the tenant namespace and pull the runtime image. |
+| `timeoutMs` | `30000` (server default) | RPC budget paperclip-server grants lease lifecycle calls. Raise it (e.g. `180000`) when a company's first dispatch has to bootstrap the tenant namespace and pull the runtime image. Capped at 15 minutes by the server's plugin RPC ceiling. |
 
 Full JSON Schema in `src/manifest.ts`.
 
--- a/packages/plugins/sandbox-providers/kubernetes/README.md
+++ b/packages/plugins/sandbox-providers/kubernetes/README.md
@@ -74,6 +74,7 @@ Common optional fields:
 | `serviceAccountAnnotations` | `{}` | Annotations applied to per-tenant ServiceAccount (e.g. IRSA `eks.amazonaws.com/role-arn`). |
 | `jobTtlSecondsAfterFinished` | `900` | Seconds after a Job completes before garbage-collection. |
 | `podActivityDeadlineSec` | `3600` | Hard ceiling on a single run's wall-clock time. |
+| `timeoutMs` | `30000` (server default) | RPC budget paperclip-server grants lease lifecycle calls. Raise it (e.g. `180000`) when a company's first dispatch has to bootstrap the tenant namespace and pull the runtime image. |
 
 Full JSON Schema in `src/manifest.ts`.
 
--- a/packages/plugins/sandbox-providers/kubernetes/README.md
+++ b/packages/plugins/sandbox-providers/kubernetes/README.md
@@ -74,7 +74,7 @@ Common optional fields:
 | `serviceAccountAnnotations` | `{}` | Annotations applied to per-tenant ServiceAccount (e.g. IRSA `eks.amazonaws.com/role-arn`). |
 | `jobTtlSecondsAfterFinished` | `900` | Seconds after a Job completes before garbage-collection. |
 | `podActivityDeadlineSec` | `3600` | Hard ceiling on a single run's wall-clock time. |
-| `timeoutMs` | `30000` (server default) | RPC budget paperclip-server grants lease lifecycle calls. Raise it (e.g. `180000`) when a company's first dispatch has to bootstrap the tenant namespace and pull the runtime image. Capped at 15 minutes by the server's plugin RPC ceiling. |
+| `timeoutMs` | `30000` (server default) | RPC budget paperclip-server grants lease lifecycle calls. Raise it (e.g. `180000`) when a company's first dispatch has to bootstrap the tenant namespace and pull the runtime image. Rejected above `900000`, the 15 minute ceiling paperclip-server applies to every plugin RPC. |
 
 Full JSON Schema in `src/manifest.ts`.
 
--- a/packages/plugins/sandbox-providers/kubernetes/src/manifest.ts
+++ b/packages/plugins/sandbox-providers/kubernetes/src/manifest.ts
@@ -102,7 +102,7 @@ const manifest: PaperclipPluginManifestV1 = {
             minimum: 1,
             maximum: 86400000,
             description:
-              "RPC budget in milliseconds for lease lifecycle calls (acquire/resume/release/execute). Default 30000 (paperclip-server default). Raise it for clusters where the first dispatch is slow to bootstrap the tenant namespace or pull the runtime image.",
+              "RPC budget in milliseconds for lease lifecycle calls (acquire/resume/release/execute). Default 30000 (paperclip-server default). Raise it for clusters where the first dispatch is slow to bootstrap the tenant namespace or pull the runtime image. Values above 900000 cannot take effect: paperclip-server caps every plugin RPC at 15 minutes.",
           },
           adapterType: {
             type: "string",
--- a/packages/plugins/sandbox-providers/kubernetes/src/manifest.ts
+++ b/packages/plugins/sandbox-providers/kubernetes/src/manifest.ts
@@ -97,6 +97,13 @@ const manifest: PaperclipPluginManifestV1 = {
             minimum: 1,
             description: "Hard ceiling on a single run's wall-clock time (default: 3600).",
           },
+          timeoutMs: {
+            type: "integer",
+            minimum: 1,
+            maximum: 86400000,
+            description:
+              "RPC budget in milliseconds for lease lifecycle calls (acquire/resume/release/execute). Default 30000 (paperclip-server default). Raise it for clusters where the first dispatch is slow to bootstrap the tenant namespace or pull the runtime image.",
+          },
           adapterType: {
             type: "string",
             description:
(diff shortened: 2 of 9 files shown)


stubbi:
Thanks — both findings were right, and both are fixed in edf3c128.

**1. Exit codes masquerading as signals.** Agreed, and I went further than narrowing the table: the provider no longer infers a signal at all. A Kubernetes exec ends in a `V1Status` that carries an `ExitCode` cause and no termination signal, and the pod's container status describes PID 1 rather than the process we exec'd, so there is no observation to report. A `128 + N` status is indistinguishable from a command that chose the same status, and mixing it into `signal` also produced a shape no other provider emits (Node-based providers report `exitCode: null` with a signal, never both).

`signal` is now set explicitly to `null` on every return path — still a small improvement over leaving the field undefined, since a consumer can tell "this provider reports no signal" from "this provider did not answer" — and the rationale is documented at the definition so the inference does not get re-added later.

The operational value is kept without the false claim: exit 137 and exit 143 (the only two statuses that carry real meaning on Kubernetes) now append one hedged line to stderr, in the same spirit as the existing `appendNetworkEgressDenyHint`:

> Exit code 137 matches the POSIX 128+signal convention for SIGKILL, which on Kubernetes usually means the process hit the container memory limit. The Kubernetes exec API reports no signal, so a command that deliberately exits with this status is indistinguishable.

Every other status leaves stderr byte-identical.

**2. Timeout range exceeds the host limit.** Agreed that silent truncation is the problem. I kept the `86400000` bound rather than capping at `900000`, because that bound matches the server's own `environment-config.ts` schema for the same key — capping here would reject a config the platform accepts elsewhere, and would hard-code a host constant into a plugin that ships independently. Instead `onEnvironmentValidateConfig` now warns, using the mechanism this provider already uses for the egress-mode caveat:

> timeoutMs=1800000 exceeds the 900000 ms ceiling paperclip-server applies to every plugin RPC, so lease lifecycle calls will still be abandoned after 900000 ms. …

The ceiling is also documented in the zod schema comment, the manifest `configSchema` description, and the README config table. The duplicated 15-minute figure is advisory only, so if `MAX_RPC_TIMEOUT_MS` ever changes the warning degrades rather than the behavior.

Tests were written before each change and observed failing: "never claims a signal for a signal-shaped exit code it cannot observe", the three stderr-hint cases (including "leaves stderr untouched for exits that carry no such meaning"), and the two `timeoutMs` warning cases. Suite: 203 passed, 21 new; the only failures are 7 pre-existing `file-sync` tests that need `/proc/self/fd` and cannot run on macOS (identical on a clean `master`).

stubbi:
Right again, and this one changed my mind: I checked `validatePluginEnvironmentDriverConfig` / `validatePluginSandboxProviderConfig` in `server/src/services/plugin-environment-driver.ts` and warnings are only forwarded inside the `unprocessable` payload of a **rejected** config — the success path returns just `normalizedConfig`, so no warning from `onEnvironmentValidateConfig` reaches the operator on save. My warning would have been invisible, which is exactly the silent-clamp behavior you flagged.

Fixed in e946dbfc: the schema now caps `timeoutMs` at `HOST_MAX_RPC_TIMEOUT_MS` (900000) and rejects anything larger with

> timeoutMs must not exceed 900000 ms: paperclip-server caps every plugin RPC at 15 minutes, so a larger budget cannot take effect

which does surface, because rejections are propagated. The manifest's JSON Schema `maximum` is set to the same value so the config form cannot offer something the provider then rejects, and a test pins the two together. The advisory warning and its constant in `plugin.ts` are gone.

No migration concern: the key was stripped by the schema until this PR and the save path persists the normalized config, so no stored Kubernetes environment can be carrying an oversized `timeoutMs` today. The tradeoff I accepted is the duplicated 15-minute figure — if `MAX_RPC_TIMEOUT_MS` is ever raised, this bound has to be raised with it; that is called out in the comment on the constant and in the PR risks.

Suite after the change: 204 passed, 22 new; typecheck clean. The only failures remain the 7 pre-existing `file-sync` tests that require `/proc/self/fd` (macOS, identical on a clean `master`).

PR #10155, opened 2026-07-24
feat(sandbox): add task-scoped egress grants

cryppadotta:
## Thinking Path

> - Paperclip is the open source app people use to manage AI agents for work
> - Confinement providers protect agent runs with default-deny network policies
> - Kubernetes environments currently apply only provider-level, namespace-wide egress allowances
> - Tasks that legitimately need GitHub or package registries therefore cannot request narrow access, while network failures do not explain the governing policy or how to request a grant
> - This pull request adds issue-scoped egress grants that become workload-owned, run-label-selected policies and carries the effective grant through lease audit metadata
> - The benefit is that internet-dependent work can run without enabling broad egress for every concurrent task, and denied requests point operators to the exact grant path

## Linked Issues or Issue Description

No public issue exists. Related but distinct: Refs #9944, which adds a provider-wide open-internet posture; this PR keeps provider defaults narrow and adds per-task grants.

**Problem / motivation**
Kubernetes sandbox egress is configured at the provider/tenant level. A task that needs to clone from GitHub or install from PyPI cannot request those destinations without changing the policy for every run in the tenant namespace. DNS/connectivity failures also surface as generic tool errors with no policy name or remediation path.

**Proposed solution**
Accept `executionWorkspaceSettings.networkEgress.allowFqdns` and `allowCidrs`, forward the setting through heartbeat environment acquisition, and create a workload-owned NetworkPolicy or CiliumNetworkPolicy selected by `paperclip.io/run-id`. Record the effective grant in lease activity/metadata, expose policy context through `PAPERCLIP_NETWORK_EGRESS_*`, and append the grant path to likely policy-related stderr failures.

**Alternatives considered**
A provider-wide open-internet switch is broader than required and is already covered by #9944. Mutating the existing namespace policy would leak each task's destinations to other concurrent runs. Standard Kubernetes NetworkPolicy cannot enforce FQDNs exactly, so standard mode uses the existing hardened public-IPv4 TCP 80/443 fallback only for the selected run; Cilium mode remains exact.

**Roadmap alignment**
This extends the existing cloud/sandbox agent roadmap capability with task-level control-plane policy and does not duplicate a planned roadmap item.

## What Changed

- Added validated `networkEgress` grants to issue execution workspace settings and forwarded them through environment lease acquisition.
- Added workload-owned, run-label-scoped NetworkPolicy/CiliumNetworkPolicy resources for task FQDN/CIDR grants.
[...]

PR #10264, opened 2026-07-25
feat(plugin-kubernetes): reusable leases (reuseLease, reuse-aware release, supportsReusableLeases)

stubbi:
## Thinking Path

> - Agents run their work inside sandboxes that a provider plugin provisions, and every provision costs the run a cold start
> - The Kubernetes provider already implements `onEnvironmentResumeLease`, complete with a bounded readiness re-check and correct not-resumable semantics
> - But the driver manifest never declares `supportsReusableLeases`, and paperclip-server gates resume strictly on that flag, so the whole resume path is unreachable dead code, and `onEnvironmentReleaseLease` unconditionally deletes the workload anyway, so there would be nothing to resume even if the flag were set
> - This pull request declares the capability, adds an opt-in `reuseLease` config key, and makes release reuse-aware, with the rule that any lease which carried a task-scoped egress grant is always torn down
> - The benefit is that a Kubernetes environment can skip namespace bootstrap, pod scheduling, image pull, and adapter install on every run after the first, without a per-task network grant ever outliving its task

No public issue. Follows #10155 (task-scoped egress grants), whose lease metadata this PR reads to decide when reuse is unsafe.

**Problem / motivation**

Every Kubernetes-backed agent run pays a full cold provision: create the Sandbox CR, schedule the pod, pull the runtime image, install the adapter. The provider was clearly built for reuse, so this is not a missing feature so much as a disconnected one:

- `onEnvironmentResumeLease` is fully implemented in `plugin.ts`, including a 30s bounded readiness re-check and a correct "gone means not resumable" result that tells the server to fall back to `acquireLease`.
- `server/src/services/environment-runtime.ts` gates every resume decision on `driver.supportsReusableLeases === true`. The Kubernetes manifest does not set it, so it is `undefined` and the resume hook is never called.
- `onEnvironmentReleaseLease` deletes the workload unconditionally, so even with the flag set there would be no pod left to resume.

There is also a user-visible inconsistency: `GET /environments` capabilities report `supportsReusableLeases: driver.supportsReusableLeases ?? true`, so the UI already advertises reusable leases for this driver while the runtime, which requires `=== true`, never resumes one.

Daytona, e2b, modal, novita, exe-dev, and cloudflare all have a reuse-aware release path. Kubernetes was the outlier.

**Proposed solution**

- Declare `supportsReusableLeases: true` on the driver so the server can reach the resume hook that already exists.
- Add a `reuseLease` config key, declared in both the zod schema and the manifest JSON Schema. It must be in the zod schema: the schema strips keys it does not know, so an operator-set value would be dropped during normalization and never reach the plugin's release path.
- Make `onEnvironmentReleaseLease` keep the Sandbox CR, its pod, and the per-run Secret when reuse applies, and fall through to the existing teardown otherwise.

**Why `reuseLease` defaults to false**

Conservative on purpose. A reused pod stays Running and keeps holding namespace quota and node capacity until the lease is destroyed, which is a real resource-posture change for an operator who did not ask for it. Defaulting to false means no existing environment changes behavior on upgrade: release still tears down exactly as before, and an operator opts in per environment once they have sized their `ResourceQuota` for idle agents. This also matches Daytona's default for the same key.

**The safety rule: a task-scoped egress grant is never reused**

A grant from #10155 is enforced by a policy that selects the pod through its `paperclip.io/run-id` label. That label is fixed at pod-creation time and therefore stays with a kept-alive pod forever. Reusing such a pod would hand the next run an egress allowance it never requested. So a lease that carried a grant is always torn down on release, even with `reuseLease` on. Teardown cascades the policy away with it (the policy is workload-owned) and forces the next run through `acquireLease`, which grants exactly its own destinations.

The check reads both `scopedNetworkPolicyName` (the policy that was created) and `scopedNetworkEgress` (the grant that asked for it). They agree in practice; requiring both to be empty keeps reuse fail-closed if they ever do not.

Reuse is fail-closed in the other direction too: a resumed lease carries no grant, so a workspace that starts requesting one after the lease was taken sees its egress denied rather than silently widened.

**Alternatives considered**

- *Stop the pod on release and start it on resume*, the way the Daytona provider does. Kubernetes has no equivalent: a Sandbox pod that is deleted cannot be revived in place, and the provider's own resume path already documents this. Keeping the pod Running is the only shape reuse can take here.
- *Delete the scoped policy on release and keep the pod.* This still leaves the pod carrying the old run's `paperclip.io/run-id` label and its `PAPERCLIP_NETWORK_EGRESS_*` env from the acquiring run, so the next run's view of its own policy would be wrong. Destroying the lease is simpler and provably safe.
- *Default `reuseLease` to true.* Rejected: it would silently convert every existing Kubernetes environment's idle agents into pods that hold quota indefinitely.

**Roadmap alignment**

Extends the existing sandbox-provider lease lifecycle. No new roadmap surface, no schema or migration changes.

## What Changed

- `src/manifest.ts`: driver declares `supportsReusableLeases: true`; config schema publishes `reuseLease` (`default: false`).
- `src/types.ts`: `reuseLease: z.boolean().default(false)` on the provider config schema, so an operator-set value survives normalization.
- `src/plugin.ts`: `onEnvironmentReleaseLease` returns early, keeping the Sandbox CR, pod, and per-run Secret, when `reuseLease` is on, the lease backend is `sandbox-cr`, and the lease carried no task-scoped egress grant. Adds the `leaseCarriedTaskScopedEgress` helper that encodes the safety rule. Per-lease upload-interceptor buffers and the readiness cache are cleared on every release path, including the reuse path: a kept-alive pod can still be evicted between runs, so the next use must re-confirm Ready.
- `test/unit/plugin-lease-lifecycle.test.ts`: seven release-lifecycle cases (default teardown, reuse keeps CR/pod/Secret, scoped-policy forces teardown, non-empty recorded grant forces teardown, empty-string policy name is not a grant, job backend always tears down, backend inferred from config).
- `test/unit/manifest.test.ts` (new): the driver declares the capability and publishes the config key.
- `test/unit/types.test.ts`: `reuseLease` defaults to false and survives normalization.
- `README.md`: a "Lease reuse" section covering the `sandbox-cr` restriction, the standing quota cost, best-effort resume, and the egress-grant rule; plus the reciprocal note in the task-scoped egress section.

The `job` backend is deliberately excluded: a Job's pod is terminal once the Job finishes, so keeping it would strand a lease nothing can exec into.

## Verification

Note for reviewers: `pnpm-workspace.yaml` excludes `packages/plugins/sandbox-providers/**`, so CI does not run this plugin's unit tests. They were run locally.

- `pnpm --dir packages/plugins/sandbox-providers/kubernetes test` before implementing, with only the new tests added: **14 failed | 186 passed**, the 7 new assertions failing exactly as intended.
- Same command after implementing: **7 failed | 193 passed (200)**. All 11 new tests pass.
- The 7 remaining failures are pre-existing and environment-specific, all in `test/unit/file-sync.test.ts`, which shells out to GNU `tar` and `/proc/self/fd` and cannot pass on macOS. Confirmed identical on a clean `origin/master` checkout of the same worktree: **7 failed | 182 passed (189)**, same 7 tests.
- `pnpm --dir packages/plugins/sandbox-providers/kubernetes exec tsc --noEmit` — clean.

## Risks

**Low risk by construction: nothing changes unless an operator opts in.** `reuseLease` defaults to
false, so every existing Kubernetes environment keeps the exact release behavior it has today. No
schema change, no migration, no server-side change.

With `reuseLease: true` the risks are:

- **Standing resource cost.** A kept-alive pod stays Running and holds namespace quota and node
  capacity until paperclip-server destroys the lease. An operator who turns this on without room in
  their `ResourceQuota` can starve new runs. Documented in the README and in the config-key
  description so it is visible at the point of decision.
- **Stale in-pod state carries across runs.** A resumed pod keeps its `/workspace` emptyDir, its
  installed adapter, and the per-run Secret from the acquiring run. That is inherent to lease reuse
  and identical to what every other reuse-capable provider does; the server's lease fingerprint
  scopes reuse to the same company, environment, execution workspace, agent, and adapter.
- **Egress asymmetry.** A workspace that always requests a task-scoped grant will never reuse a
  lease, so it sees no start-latency benefit. In the reverse case (a workspace starts requesting a
  grant after the lease was taken) the resumed lease has no grant, so the run is denied rather than
  silently widened. Fail-closed in both directions, but the denial can look surprising, so both
  cases are documented.
- **Declaring `supportsReusableLeases: true` applies with the flag off.** It only enables the
  server's resume path, which is itself gated on `parsed.config.reuseLease` at every call site
  (`environment-runtime.ts` lines 822, 851, 889, 963). With `reuseLease` false the server computes
  no fingerprint, lists no reusable candidates, and stamps every lease `ephemeral` exactly as
  before. It also removes an existing inconsistency: the environments API already reported this
  driver as reuse-capable via `?? true`.

The blast radius if the reuse branch were wrong in the unsafe direction is a leaked network grant,
which is why that case is a hard, unconditional teardown with its own tests rather than a
configurable one.

## Model Used

Claude Opus 5 (Anthropic), model id `claude-opus-5[1m]`, 1M context window, extended thinking, with
tool use and code execution (repository exploration, local test runs).

## Checklist

- [x] I have searched GitHub for duplicate or related PRs and linked them above — no duplicate exists. Nearest neighbors: #10263 (independent sibling on the same plugin, exec-path correctness, no overlap with the lease lifecycle), #10155 (task-scoped egress grants, whose lease metadata this PR reads), #7852 (single-flight lease acquisition for shared workspaces, a concurrency concern rather than reuse).
- [x] My branch name describes the change (`k8s-lease-reuse`) and contains no internal Paperclip ticket id or instance-derived details
- [x] I have run tests locally and they pass (see Verification — this plugin is excluded from the CI workspace, so local runs are the gate)

### DIFF
--- a/packages/plugins/sandbox-providers/kubernetes/README.md
+++ b/packages/plugins/sandbox-providers/kubernetes/README.md
@@ -98,6 +98,11 @@ What to know before turning it on:
   report Ready within 30 seconds, the lease reports itself expired and the run
   falls back to a fresh provision. A Kubernetes pod cannot be revived in place
   the way a stopped VM-backed sandbox can.
+- **Only leases acquired while it was on are kept.** paperclip-server fixes a
+  lease's policy at acquisition, so turning `reuseLease` on takes effect from the
+  next lease onward. Leases already in flight are torn down on release as usual,
+  which is what keeps a mid-lease flip from stranding a pod the server has
+  already written off.
 - **Leases with a task-scoped egress grant are never reused** (below).
 
 ### Task-scoped egress grants
--- a/packages/plugins/sandbox-providers/kubernetes/README.md
+++ b/packages/plugins/sandbox-providers/kubernetes/README.md
@@ -61,6 +61,7 @@ Common optional fields:
 | Field | Default | Purpose |
 |---|---|---|
 | `backend` | `"sandbox-cr"` | `sandbox-cr` (alpha, requires agent-sandbox controller) or `job` (stable, one-shot entrypoint). |
+| `reuseLease` | `false` | Keep the sandbox pod alive on release so the next run resumes it. See [Lease reuse](#lease-reuse). |
 | `adapterType` | `"claude_local"` | One of the supported adapter types (claude_local, codex_local, gemini_local, cursor_local, opencode_local, pi_local). Determines runtime image + env keys + egress allow-list. |
 | `namespacePrefix` | `"paperclip-"` | Prefix for the per-company tenant namespace. |
 | `companySlug` | derived from companyId | Override the auto-derived company slug. |
@@ -76,6 +77,28 @@ Common optional fields:
 | `podActivityDeadlineSec` | `3600` | Hard ceiling on a single run's wall-clock time. |
 
 Full JSON Schema in `src/manifest.ts`.
+
+### Lease reuse
+
+By default every run provisions a fresh sandbox and tears it down on release, so
+each run pays a cold start: namespace bootstrap, pod scheduling, image pull, and
+the adapter install. Set `reuseLease: true` to keep the Sandbox CR, its pod, and
+the per-run Secret alive on release instead. The next run for the same
+environment, execution workspace, agent, and adapter resumes that pod and skips
+the provision entirely.
+
+What to know before turning it on:
+
+- **`sandbox-cr` only.** A Job's pod is terminal once the Job finishes, so a
+  `job` backend lease is always torn down on release regardless of the setting.
+- **The pod keeps running** (and keeps holding namespace quota and node capacity)
+  until paperclip-server destroys the lease. Size `ResourceQuota` for the number
+  of agents you expect to be idle, not just the number running.
+- **Resume is best-effort.** If the pod is gone, terminally failed, or does not
+  report Ready within 30 seconds, the lease reports itself expired and the run
+  falls back to a fresh provision. A Kubernetes pod cannot be revived in place
+  the way a stopped VM-backed sandbox can.
+- **Leases with a task-scoped egress grant are never reused** (below).
 
 ### Task-scoped egress grants
 
@@ -93,6 +116,8 @@ Keep provider-level egress defaults narrow, then grant only the destinations a t
 ```
 
 The provider creates a workload-owned policy selected by the task run label, so the additional destinations do not become reachable from other concurrent agent pods. Cilium mode enforces FQDNs directly. Standard NetworkPolicy mode cannot express FQDNs, so an FQDN grant permits public IPv4 TCP 80/443 for that run while excluding private, loopback, link-local, CGNAT, and multicast ranges. Network failures that look policy-related include the grant path in stderr, and the sandbox exposes the effective policy through `PAPERCLIP_NETWORK_EGRESS_*` environment variables.
+
+A grant is scoped to one run and never outlives it. The policy selects the pod by its run-id label, which is fixed at pod-creation time, so a lease that carried a grant is always torn down on release even when `reuseLease` is on: reusing that pod would hand the next run an egress allowance it never asked for. Workspaces that always request a grant therefore never reuse leases. In the other direction reuse is fail-closed: a resumed lease carries no grant, so a workspace that starts requesting one after the lease was taken sees its egress denied rather than silently widened.
 
 ## What gets created in your cluster
 
--- a/packages/plugins/sandbox-providers/kubernetes/src/manifest.ts
+++ b/packages/plugins/sandbox-providers/kubernetes/src/manifest.ts
@@ -23,6 +23,10 @@ const manifest: PaperclipPluginManifestV1 = {
       displayName: "Kubernetes",
       description:
         "Dispatches agent runs in per-tenant Kubernetes namespaces. Default backend (sandbox-cr, alpha) uses kubernetes-sigs/agent-sandbox for multi-command exec; fallback backend (job) uses stable batch/v1 Job for clusters without agent-sandbox installed.",
+      // The driver implements onEnvironmentResumeLease; paperclip-server only
+      // reaches it when this flag is set. Reuse still has to be turned on per
+      // environment through the `reuseLease` config key below.
+      supportsReusableLeases: true,
       configSchema: {
         type: "object",
         properties: {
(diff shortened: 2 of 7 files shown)


stubbi:
Thanks — the ephemeral-lease finding is correct, and it is worse than a config-drift edge case if left unfixed, so I traced it end to end before changing anything.

**Confirmed mechanism.** `resolvedLeasePolicy` is computed once at acquisition (`environment-runtime.ts:963`) and never revisited. `cleanupObsoleteReusableSandboxLeases` only iterates leases filtered to `leasePolicy === "reuse_by_environment"`, so an `ephemeral` lease has no destroy path at all — after `environmentReleaseLease` the server just marks the row released. Retaining that workload leaks a Sandbox CR, pod, and Secret with no reclaim path. Exactly as described.

**Fix (783e617).** The release decision no longer comes from the current config alone. `onEnvironmentAcquireLease` stamps `acquiredForReuse: config.reuseLease` on the lease metadata, `onEnvironmentResumeLease` carries that stamp through rather than re-deriving it from the config, and release keeps a workload only when reuse was on at **both** ends. An absent stamp (leases predating the field) means tear down. That mirrors the server exactly: policy fixed at acquire, `reuseLease` re-checked before anything is resumed.

Four tests added for the new invariant: config flipped on mid-lease still destroys, a lease predating the stamp destroys, a reuse-acquired lease destroys once `reuseLease` is turned back off, and resume carries the stamp instead of inventing one. README documents that turning the flag on takes effect from the next lease onward.

**One related case I checked and did not change.** Ad-hoc leases (`heartbeatRunId === null`) are also stamped `ephemeral`, so the same leak would apply to operator `Test` probes. Both callers already force `reuseLease: false` into the driver config for exactly this reason (`routes/agents.ts:376`, `services/environment-probe.ts:71`), so the config the plugin sees is already correct there. The `acquiredForReuse` stamp covers that path too as a second line of defense, since a probe lease is now stamped false at acquisition regardless.

Full plugin suite locally: 198 passed, plus the 7 `file-sync.test.ts` failures that are pre-existing on a clean `master` checkout (they shell out to GNU `tar` and `/proc/self/fd`, so they cannot pass on macOS). Note this package is excluded from `pnpm-workspace.yaml`, so CI does not run these tests.

PR #10265, opened 2026-07-25
fix(worktree): never repair config or env outside the linked worktree

MrBlackTongue:
## Problem

Worktree repair resolves **where** it writes and **what** it writes through two independent chains.

The destination comes from `resolvePaperclipConfigPath()` (`server/src/paths.ts`), which falls back to the canonical `<home>/instances/<id>/config.json` whenever cwd sits outside a worktree and `PAPERCLIP_CONFIG` is unset. The contents come from `persistedEnv.PAPERCLIP_HOME ?? env.PAPERCLIP_HOME ?? env.PAPERCLIP_WORKTREES_DIR`. A process that carries a leaked `PAPERCLIP_IN_WORKTREE=true` — inherited from a parent shell, a sourced env file, or an agent runtime — therefore computes worktree contents and writes them into a foreign instance.

Two orderings make it worse: `process.env.PAPERCLIP_HOME` is assigned only *after* the destination has been resolved, and `server/src/config.ts` dotenv-loads the env file *before* `maybeRepairLegacyWorktreeConfigAndEnvFiles()` runs.

It is also self-sustaining. The env pass spreads `...existingEnvEntries` and rewrites the file in place, appending `PAPERCLIP_IN_WORKTREE="true"`, `PAPERCLIP_HOME=<that run's directory>` and `PAPERCLIP_CONFIG=<the foreign config>`. Once that has happened, every later start of the affected instance re-enters repair on its own, with no worktree run involved at all.

We hit this in production: a canonical instance's `config.json` was rewritten with another run's paths, which pointed the embedded-Postgres data dir and backup dir at a directory that no longer existed. Backups stopped, and a restart would have come up against an empty database.

`d1b9448b5` already narrows this a lot — a target only counts as a worktree when its config sits in a `<root>/.paperclip/` layout and its own persisted env declares `PAPERCLIP_IN_WORKTREE=true`. That closes the exact incident shape, but it is a check on the *form* of the resolved path rather than on its *provenance*, so the split between the two chains remains:

- a main checkout's `<repo>/.paperclip/` satisfies the layout check, so once its `.env` has been poisoned with `PAPERCLIP_IN_WORKTREE=true`, repair happily rewrites the main checkout's config into `~/.paperclip-worktrees`;
- a run inside worktree A that inherited worktree B's `PAPERCLIP_CONFIG` repairs B, using A's ambient `PAPERCLIP_HOME`;
- `isPathInside` exists in this file but is only applied to the *contents* (`needsWorktreeConfigRepair`), never to the destination, and `isLinkedGitWorktreeCheckout` is not used here at all.

## Fix

**Resolve the destination from the same chain as the contents.** `resolveWorktreeRuntimeContext` now finds the linked git worktree checkout the process actually runs in (nearest ancestor of cwd that has a `.paperclip/` directory and a `gitdir:` `.git` file), and derives `configPath`/`envPath` from it. A configured `PAPERCLIP_CONFIG` only fills in when cwd is outside every worktree, and only when it names a linked worktree checkout — so a canonical or stale value inherited from a parent process can no longer steer repair. When it names the very checkout we run in, its spelling is kept, so repair does not rewrite `.env` merely to swap a symlinked path for its resolved form.

**Guard every write.** `isWorktreeRepairWriteAllowed()` gates each repair write: a destination outside the linked worktree checkout is refused and logged instead of written. Whatever the ambient environment claims, repair cannot push `PAPERCLIP_HOME` / `PAPERCLIP_CONFIG` / `PAPERCLIP_IN_WORKTREE` into an `.env` that is not the worktree's own.

**Close the dev runner's route to the same split.** `bootstrapDevRunnerWorktreeEnv` only filled empty keys, so a `PAPERCLIP_CONFIG` inherited from a canonical parent outranked the worktree's own value. An inherited config path pointing outside the checkout is now replaced by the worktree's; every other explicitly supplied value still wins, as before.

## Tests

`pnpm --filter @paperclipai/server exec vitest run src/__tests__/worktree-config.test.ts src/__tests__/dev-runner-worktree.test.ts` → 24 passed.

New cases:

- `leaves the canonical instance untouched when cwd is outside the worktree and PAPERCLIP_CONFIG is unset` — the incident shape. Green on `master` too, thanks to `d1b9448b5`; kept so the regression stays pinned.
- `refuses to repair a checkout that is not a linked git worktree` — red before, green after.
- `repairs the checkout it runs in when PAPERCLIP_CONFIG points at another worktree` — red before, green after.
- `refuses every repair write aimed outside the linked worktree checkout` — unit coverage of the write guard.
- `replaces an inherited PAPERCLIP_CONFIG that points outside this worktree` / `keeps an inherited PAPERCLIP_CONFIG that already points inside this worktree` — red before, green after.

Existing fixtures gained a `gitdir:` `.git` marker so they still look like linked worktree checkouts; their assertions are unchanged.

Two existing assertions had to move, both about path *spelling* rather than behaviour: `ignores stale migrated env paths…` and `does not persist transient runtime home overrides…` now pass because the configured spelling is preserved when it names the current checkout — without that, cwd's symlink-resolved form would have rewritten `.env` on every start.

`pnpm --filter @paperclipai/server typecheck` and `biome check` on the touched files are clean.


MrBlackTongue:
Superseded by #10266, which is the same change on a branch name that follows CONTRIBUTING (no internal ticket id) and with a PR description that fills in the required template sections. Closing this one to avoid a duplicate; please review #10266.

Issue #6800, opened 2026-05-27
request_confirmation reject fires a comment-shaped wake with no comment — assignee resumes to `pending comments: 0/0` and reports no input; decline reason dropped


brendandebeasi:
## TL;DR

Declining a `request_confirmation` wakes the assignee agent but hands it nothing to act on: no comment is created (the reject enqueues a wake tagged `issue_commented` with `sourceCommentId: null`), and the wake renderer has no `rejected` branch so it never reads the decline reason from `result.reason`. The agent resumes, sees `pending comments: 0/0`, reports "no input," and re-parks the issue in `in_review`. The human's feedback is silently dropped. Same bug class as #3780 / #3781 (approvals), but in the interactions / `request_confirmation` path.

## Summary

Declining a `request_confirmation` interaction (with `rejectRequiresReason: true`) wakes the assignee agent, but the wake carries nothing actionable, so the agent resumes, sees `pending comments: 0/0`, concludes there is no input to act on, and re-parks the issue in `in_review`. The human's decline reason is silently dropped.

Two compounding defects:

1. **The reject enqueues a comment-shaped wake that points at no comment.** The `agent_wakeup_requests` row created on reject has `reason: "issue_commented"` but a payload with `sourceCommentId: null`:
   ```json
   {
     "mutation": "interaction",
     "interactionKind": "request_confirmation",
     "interactionStatus": "rejected",
     "sourceCommentId": null
   }
   ```
   No `issue_comments` row is ever written from the decline. The reason text lives only in `issue_thread_interactions.result.reason` (`{"outcome":"rejected","reason":"<human text>"}`), which nothing on the wake path reads.

2. **The resume-delta renderer has no `rejected` branch.** In `packages/adapter-utils/src/server-utils.ts`, the planning-mode wake only special-cases acceptance:
   ```ts
   const acceptedPlanContinuation =
     !hasWakeComments &&
     normalized.interactionKind === "request_confirmation" &&
     normalized.interactionStatus === "accepted";
   ```
   A `rejected` status produces no directive, and the renderer never reads `result.reason`. The `pending comments` line is computed purely from `issue_comments` (`includedCount/requestedCount`), so it correctly shows `0/0`. The boilerplate even states (server-utils.ts, near L125): *"for request_confirmation this resumes only after acceptance."*

Net: a decline-with-feedback is a silent drop. The agent is woken but blind.

## Reproducer

[...]

PR #10278, opened 2026-07-26
fix(server): wake the requester on any interaction rejection that carries a reason

### LINKED ISSUE #8617: request_confirmation reject on a user-assigned issue never wakes the creator agent (silent stall)
### What happened

When a board/user assignee **rejects** an agent-authored `request_confirmation` (or `request_checkbox_confirmation`) on an issue whose `assigneeAgentId` is `null` (i.e. the issue is assigned to a user, not an agent), the post-reject wake helper no-ops because it keys off `assigneeAgentId`. The originating agent is never woken, so the issue **silently stalls**. A human has to manually @-mention an agent to recover.

This is common in the "CEO hands review to the board" delegation pattern, where an agent requests confirmation, the issue is handed to a board user for review, and the user rejects.

### What I expected

Rejecting the confirmation should return the issue to the **creating agent** (mirroring the existing accept path) and wake it, so work continues without manual intervention.

### Affected area

- `server/src/services/issue-thread-interactions.ts` — the reject path for `request_confirmation` / `request_checkbox_confirmation`
- `server/src/routes/issues.ts` — the issues `/reject` route

### Additional context

The **accept** path already reassigns the resolved confirmation back to the creating agent; the **reject** path never mirrored it, so the `assigneeAgentId` precondition on the wake helper is unmet and the wake silently no-ops. Fix in progress in PR #8516 (reassign at the service layer on reject, symmetric with accept).

ericdfields:
## Thinking Path

> - Issue-thread interactions (`request_confirmation`, `suggest_tasks`, etc.) let a human respond to an agent's request; `continuationPolicy` controls whether resolving one wakes the requesting agent
> - `continuationPolicy` is meant to control noise on the *accept* path — e.g. `wake_assignee_on_accept` so a routine "looks good" doesn't burn a wake, or `none` when the requester genuinely doesn't need to resume
> - But the same gate also silently swallowed *rejections that carried a written reason* — the one outcome the requester most needs to hear, since a reject is new information while an accept is usually just "proceed as planned"
> - I hit this live, twice, on the same production issue: an agent created a `request_confirmation` with `continuationPolicy: "none"`, the operator rejected it with a substantive question ("why did we deviate from spec, and the loop still isn't clean — diagnose and propose a plan"), and it sat resolved-but-unread for hours until a human noticed and manually relayed it
> - This pull request makes a rejection-with-reason wake the assignee regardless of the configured continuationPolicy, and surfaces the reason directly in the wake payload so the assignee doesn't need a follow-up fetch to see it
> - The benefit: reasoned feedback can no longer disappear into a stranded interaction — continuationPolicy still controls the accept path exactly as before, but is no longer able to suppress an operator's actual words

Fixes: #8617
Refs: #6800

Both describe adjacent symptoms of the same underlying gap (an operator's response to an interaction/approval failing to reach the requester). This PR closes the general case for issue-thread interactions: any `reject` (or task-rejection) result whose `reason`/`rejectionReason` field is non-empty now wakes the assignee, independent of `continuationPolicy`.

## What Changed

- `queueResolvedInteractionContinuationWakeup` (`server/src/routes/issues.ts`): computes `interactionResult` up front and adds `isRejectionWithReason` (status `rejected` + non-empty `reason`/`rejectionReason`, via the existing `readConfirmationResultForWake` helper). The wake gate now fires when the configured policy allows it *or* when this is a reasoned rejection.
- Both the wake `payload` and `contextSnapshot` gain a `rejectionReason` field whenever `isRejectionWithReason` is true, so the woken agent sees the actual feedback text without an extra round-trip — mirroring how tool-action rejections already surface `declineReason`.
- Reason-less rejections are untouched: a bare "rejected, no comment" still respects the configured policy and stays silent under `none` / `wake_assignee_on_accept`. This only overrides silence when there is something to say.
- Updated two existing tests that encoded the previous (buggy) behavior — both rejected an interaction *with* a reason under `wake_assignee_on_accept` and asserted no wake; flipped to assert the wake now fires with the reason visible in the payload.
- Added: a reason-less-rejection negative case (still silent, guards against over-firing), and the exact `continuationPolicy: "none"` + reasoned-rejection case that reproduced the production incident.

## Verification

- `cd server && pnpm vitest run src/__tests__/issue-thread-interaction-routes.test.ts` → 20/20 pass.
- Revert-check: with the `issues.ts` change reverted (tests kept), the three "carries a reason" tests fail — `expect(mockHeartbeatService.wakeup).toHaveBeenCalledWith(...)` on a call that never happened. Restored and re-ran green.
- `cd server && pnpm tsc --noEmit` → clean.

## Risks

- Low, and intentionally narrow: the override condition requires both `status === "rejected"` and a non-empty reason string — an already-existing, validated field on the interaction result. No new inputs, no schema change.
- Behavior change is additive only (adds wake occasions, never removes one) and confined to the rejection-with-reason case; all previously-passing accept-path and reason-less-reject-path assertions in the same test file continue to pass unmodified.
- The `rejectionReason` field is new and additive on the wake payload/contextSnapshot shape — existing consumers that don't know about it are unaffected.

## Model Used

- Anthropic Claude — Fable 5 (`claude-fable-5`), extended thinking, agentic tool use via Claude Code (CLI). Diagnosis from a live production incident, fix, and tests all model-authored under human direction.

## Checklist

- [x] I have updated relevant documentation to reflect my changes (none applicable — inline comments cover the new branch)
- [ ] All Paperclip CI gates are green (pending first CI run on this PR)
- [ ] Greptile is 5/5 with no open P2s, recommendations, or follow-ups (pending first review pass)

### DIFF
--- a/server/src/__tests__/issue-thread-interaction-routes.test.ts
+++ b/server/src/__tests__/issue-thread-interaction-routes.test.ts
@@ -1449,7 +1449,7 @@ describe.sequential("issue thread interaction routes", () => {
     );
   });
 
-  it("does not emit a continuation wake when request confirmations are rejected", async () => {
+  it("wakes the assignee when a rejected request confirmation carries a reason, even under wake_assignee_on_accept", async () => {
     mockInteractionService.rejectInteraction.mockResolvedValueOnce({
       id: "interaction-3",
       companyId: "company-1",
@@ -1478,6 +1478,45 @@ describe.sequential("issue thread interaction routes", () => {
     const res = await request(app)
       .post("/api/issues/aaaaaaaa-aaaa-4aaa-8aaa-aaaaaaaaaaaa/interactions/interaction-3/reject")
       .send({ reason: "Needs changes" });
+
+    expect(res.status).toBe(200);
+    expect(mockHeartbeatService.wakeup).toHaveBeenCalledWith(
+      ASSIGNEE_AGENT_ID,
+      expect.objectContaining({
+        payload: expect.objectContaining({ rejectionReason: "Needs changes" }),
+        contextSnapshot: expect.objectContaining({ rejectionReason: "Needs changes" }),
+      }),
+    );
+  });
+
+  it("does not wake the assignee for a reason-less rejection under wake_assignee_on_accept", async () => {
+    mockInteractionService.rejectInteraction.mockResolvedValueOnce({
+      id: "interaction-3b",
+      companyId: "company-1",
+      issueId: "aaaaaaaa-aaaa-4aaa-8aaa-aaaaaaaaaaaa",
+      kind: "request_confirmation",
+      status: "rejected",
+      continuationPolicy: "wake_assignee_on_accept",
+      idempotencyKey: null,
+      sourceCommentId: null,
+      sourceRunId: "run-3b",
+      payload: {
+        version: 1,
+        prompt: "Apply this plan?",
+      },
+      result: {
+        version: 1,
+        outcome: "rejected",
+      },
+      createdAt: "2026-04-20T12:00:00.000Z",
+      updatedAt: "2026-04-20T12:05:00.000Z",
+      resolvedAt: "2026-04-20T12:05:00.000Z",
+    });
+    const app = await createApp();
+
+    const res = await request(app)
+      .post("/api/issues/aaaaaaaa-aaaa-4aaa-8aaa-aaaaaaaaaaaa/interactions/interaction-3b/reject")
+      .send({});
 
     expect(res.status).toBe(200);
     expect(mockHeartbeatService.wakeup).not.toHaveBeenCalled();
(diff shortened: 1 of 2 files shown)


JamesSparkMojo:
@ericdfields — Dex (Spark Mojo dispatcher, child of SPA-2074 / paperclipai/paperclip#8617). We're ready to land this upstream to unblock the silent-limbo bug, but the branch is **238 commits behind master** as of 2026-08-07, which is why merge is currently rejected with `mergeable_state: dirty`.

Per Dex's gated-harness rules, the CI receipts on the current head SHA `8157f811` are stale (they attest to a 12-day-old diff). I can't carry them forward.

Could you rebase `fix/reject-with-reason-wakes-requester` onto current master and push? That will retrigger the full check rollup (policy, review, typecheck, server tests 1/3–3/3, build, e2e shards, security/greptile/socket/superagent). Once that comes back green on the rebased head, I'll re-attest and route the merge.

The diff itself is good — it's a superset of the fix spec at SPA-2074 (extends the gate from `wake_assignee` rejections to also cover `wake_assignee_on_accept` and `none` policies when a reason is present, and surfaces `rejectionReason` at top-level). No changes needed to the patch — just the rebase.

ericdfields:
Rebased onto current master (`3435920ae`) and force-pushed — new head is `b63695eb0`. The PR now shows `MERGEABLE` and the full check rollup has retriggered.

Thanks @dexhorthy / James — one thing worth flagging before you re-attest, because the rebase **was not purely mechanical**. Master moved 239 commits and `queueResolvedInteractionContinuationWakeup` was restructured in the interim: it's now `async`, and it grew its own continuation-policy override for the `reviewPathLost` case. Both files conflicted.

The patch's *behavior* is unchanged, but its *shape* had to adapt to that new structure:

- Original patch hoisted `interactionResult` to the top of the function and replaced the early policy guard wholesale. That no longer applies — upstream's guard now sits after the `assigneeAgentId`/closed-issue check and after the awaited `reviewPathLost` computation.
- Resolution: `isRejectionWithReason` is now computed alongside `reviewPathLost` and folded into upstream's existing guard as a third override — `if (!continuationPolicyAllowsWake && !reviewPathLost && !isRejectionWithReason) return;` — rather than replacing it. `interactionResult` moved up only as far as it needed to (just above that guard), not to the function top.
- Upstream's `if (status === "expired" && !reviewPathLost) return;` was left exactly as-is. A rejection-with-reason can't be `expired`, so no override is needed there.
- Both payload spreads keep upstream's `...(reviewPathContext ?? {})` **and** add `...(isRejectionWithReason ? { rejectionReason } : {})`.
- In the test file, upstream had added `overrides accept-only continuation when rejection consumes the last review path` at the same location as this PR's `continuationPolicy: "none"` regression test. Kept both — nothing dropped on either side.

Net diff vs master is the same four behavioral changes as before, just expressed inside the new control flow. Verification on the rebased head:

- `pnpm vitest run src/__tests__/issue-thread-interaction-routes.test.ts` — **45/45 pass**, including upstream's review-path test and all four of this PR's cases.
- `pnpm tsc --noEmit` — zero errors in either changed file. (The worktree reports 139 pre-existing errors from an unbuilt `@paperclipai/plugin-sdk`; I checked out `origin/master` in the same tree and got the identical 139, so none are introduced here.)

Happy to split the guard differently if you'd rather the rejection override live in its own statement instead of joining the `reviewPathLost` condition — the two overrides are independent and it reads either way.

dexhorthy:
you're welcome

PR #8084, opened 2026-06-13
fix: hide console windows on Windows by resolving .cmd wrappers to real executables

labsclaw:
## Thinking Path

> - Paperclip is the open source app people use to manage AI agents for work
> - On Windows, adapter processes (opencode, etc.) and embedded-postgres are spawned via `.cmd` wrappers using `cmd.exe`
> - `cmd.exe` allocates a console window; the actual `.exe` binary also creates one via `conhost.exe`
> - Every heartbeat tick (~30s) spawns these processes, creating persistent visible window spam on Windows desktops
> - The fix resolves `.cmd` wrapper files to find the real executable path and spawns it directly with `windowsHide: true`
> - This eliminates all visible console windows from adapter processes while maintaining identical functionality
> - The benefit is a clean, non-disruptive experience for Windows users

## Linked Issues or Issue Description

Fixes #8182

## What Changed

- **`packages/adapter-utils/src/server-utils.ts`**:
  - `resolveSpawnTarget()`: Added `.cmd` wrapper resolution logic that parses npm-generated `.cmd` files to find the real executable path, bypassing `cmd.exe` entirely
  - `resolveSpawnTarget()`: Extracts `SET` commands from `.cmd` wrappers as environment variable overrides (excluding `dp0` which is ephemeral)
  - `runChildProcess()`: Added `windowsHide: process.platform === win32` to `spawn()` options as a fallback safety net
  - Supports four `.cmd` wrapper patterns: `%dp0%` (quoted), `%dp0%` (unquoted), `%~dp0` (quoted), `%~dp0` (unquoted)
- **`packages/adapter-utils/src/cmd-wrapper-resolution.test.ts`** (new):
  - Unit tests for `.cmd` wrapper regex parsing: npm-generated patterns, `%~dp0` patterns, `SET` command extraction, `dp0` filtering, case-insensitive matching, and path resolution with real filesystem verification

## Verification

Tested on Windows 10 (22H2) with Paperclip 2026.529.0:
- Killed all existing opencode.exe processes
- Restarted Paperclip
- Observed 2+ heartbeat cycles (60s+ each)
- Confirmed zero new visible windows appeared

Unit tests validate regex parsing cross-platform:
- npm-generated `.cmd` wrapper pattern resolution
- `%~dp0` direct pattern resolution
- `SET` command extraction with dp0 filtering
- Case-insensitive `.exe` matching
- Real filesystem path resolution

## Risks

[...]

PR #10286, opened 2026-07-26
fix: resolve CodeRabbit issues in PR #8084

labsclaw:
## Thinking Path

This PR addresses CodeRabbit findings from PR #8084. The core issue is that Windows .cmd wrappers (npm-generated) create visible console windows when spawned via cmd.exe. The fix resolves the wrapper to the real .exe and spawns it directly with `windowsHide: true` — eliminating console window spam.

### cmd-wrapper-resolution.ts
- Fixed `SET_PATTERN` regex: explicit alternation for quoted (`SET "KEY=val"`) vs unquoted (`SET KEY=val`) instead of ambiguous `"?`
- Added `resolvePercentVars()` for case-insensitive `%VAR%` expansion from caller's env

### server-utils.ts
- Changed `mergedEnv` base from `process.env` to `opts.env` (caller's sanitized env)
- Apply `resolvePercentVars()` on SET env overrides before merge

### cmd-wrapper-resolution.test.ts
- Added 13 tests covering real-world npm wrappers, SET parsing (quoted/unquoted/@SET), dp0 filtering, edge cases

## Benefits
- Eliminates cmd.exe console window spam on Windows
- Cleaner child process spawning with proper env inheritance
- Production-safe: falls back to cmd.exe when .exe not found

## Verification
- 13/13 tests passing in `packages/adapter-utils/src/cmd-wrapper-resolution.test.ts`
- CI commitperclip review: P1 fixes confirmed resolved

## Risks
- Low: all changes are inside the `.cmd/.bat` branch of `resolveSpawnTarget()` which only triggers on Windows
- Parser failures fall through to the original cmd.exe path — zero regression risk for non-Windows platforms
- The `%VAR%` expansion uses the caller's env, matching cmd.exe behavior

## Model Used
- Claude Sonnet 4.6 Thinking (antigravity-proxy) for code changes and PR description

## Checklist

- [x] I have searched the existing PR list for duplicates
- [x] My code follows the project's style guidelines
- [x] I have added tests that prove my fix works
- [x] New and existing tests pass locally
- [x] I have verified this works on Windows
- [x] **Dedup-search checkbox**: I have searched the GitHub PR list for similar PRs

### DIFF
--- a/packages/adapter-utils/src/cmd-wrapper-resolution.test.ts
+++ b/packages/adapter-utils/src/cmd-wrapper-resolution.test.ts
@@ -1,0 +1,184 @@
+import fs from "node:fs/promises";
+import os from "node:os";
+import path from "node:path";
+import { describe, expect, it } from "vitest";
+import {
+  parseCmdWrapperContent,
+  DP0_PATTERN_NPM,
+  DP0_PATTERN_DIRECT,
+  TILDE_PATTERN_NPM,
+  TILDE_PATTERN_DIRECT,
+  SET_PATTERN,
+} from "./cmd-wrapper-resolution.js";
+
+/**
+ * Tests for the .cmd wrapper resolution logic used in resolveSpawnTarget.
+ *
+ * On Windows, spawning .cmd files via cmd.exe creates visible console windows.
+ * This test verifies the regex patterns and SET-command parsing that allow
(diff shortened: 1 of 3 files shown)


labsclaw:
Friendly ping — all CI checks are green and Greptile confirmed all flagged issues are addressed (regex patterns, SET filtering, tests). Ready for maintainer review when available.

PR #10309, opened 2026-07-27
Support forced GitHub external-object refresh

valkyriweb:
## Summary
- add a strict optional `force` flag to issue External Object refresh requests and propagate it through the service so callers can bypass TTL backoff
- treat GitHub 404 responses as visibility-unverified failures instead of terminal Not Found state, preserving the last confirmed status
- update the OpenAPI and UI client contracts and cover the behavior with focused route and service tests

## Verification
- `pnpm --filter @paperclipai/server exec vitest run src/__tests__/external-object-routes.test.ts src/__tests__/external-objects-service.test.ts` — 34 passed
- `pnpm --filter @paperclipai/server typecheck` — passed
- `pnpm --filter @paperclipai/ui typecheck` — passed
- `git diff --check upstream/master...HEAD` — passed
- independent correctness review — passed with no blockers

## Test plan
- [ ] Confirm refresh without `force` still returns backoff for an object whose next refresh is not due
- [ ] Confirm refresh with `force: true` contacts the resolver despite active backoff
- [ ] Confirm anonymous GitHub 404 returns `auth_required` without replacing prior state
- [ ] Confirm authenticated GitHub 404 returns `unreachable` without creating terminal Not Found state

### DIFF
--- a/server/src/__tests__/external-object-routes.test.ts
+++ b/server/src/__tests__/external-object-routes.test.ts
@@ -277,17 +277,29 @@ describe("external object routes", () => {
     expect(mockExternalObjectsService.refreshIssueObjects).not.toHaveBeenCalled();
   });
 
+  it("requires manual refresh force to be a boolean", async () => {
+    const app = await createApp(ownerActor());
+
+    const res = await request(app)
+      .post(`/api/issues/${issueId}/external-objects/refresh`)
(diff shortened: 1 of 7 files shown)


valkyriweb:
Closing: this change was intended for a private fork deployment, not an upstream contribution.

Issue #6406, opened 2026-05-20
bug: process_detached runs hang forever — reaper never schedules termination


linespeedevents:
## Summary

When the Paperclip server loses its in-memory handle to a child `claude --print` process but the child PID is still alive, `reapOrphanedRuns()` correctly marks the run as `process_detached`. However it then unconditionally `continue`s — **it never schedules a kill**, no matter how long the run stays silent. Recovery currently requires a board-level `POST /api/heartbeat-runs/{runId}/cancel`, which itself requires `assertBoard(req)` — making it inaccessible to the agent and effectively locking the company until a human intervenes.

We hit this on 2026-05-05: a CTO agent run hung silently for ~15 days, blocking all downstream agents.

## Root Cause

In `server/src/services/heartbeat.ts → reapOrphanedRuns()`:

```ts
if (processPidAlive) {
  if (run.errorCode !== DETACHED_PROCESS_ERROR_CODE) {
    // Mark as process_detached (correct)
    ...
  }
  continue;  // <-- BUG: already-detached runs are never terminated
}
```

Once a run enters `process_detached`, every subsequent reaper tick sees `processPidAlive === true` and skips it. No kill is ever issued.

## Contributing Factor (R2, optional)

`@paperclipai/adapter-claude-local` has no idle-read deadline on the `claude --print` stdout loop. If the SSE stream stalls mid-message (e.g. on a network/pipe hiccup), the adapter blocks indefinitely, which is what caused the in-memory handle to appear lost in the first place.

## Proposed Fix (R1)

In `reapOrphanedRuns()`, add a silent-terminate path for runs already in `process_detached` state:

```ts
} else {
  // Already marked process_detached — check silence age
  const silenceRefTime = run.lastOutputAt ?? run.updatedAt ?? run.startedAt ?? run.createdAt;
  const silenceAgeMs = silenceRefTime ? now.getTime() - new Date(silenceRefTime).getTime() : Infinity;
  if (silenceAgeMs > DETACHED_RUN_SILENT_TERMINATE_THRESHOLD_MS) { // suggested: 6h
    await terminateHeartbeatRunProcess({ pid: run.processPid, processGroupId: run.processGroupId });
    // finalize as failed with errorCode: "process_detached_silent"
    // + full cleanup: release leases, release issue execution, finalize agent status
  }
}
```

A PR with this fix is open at: https://github.com/linespeedevents/paperclip/pull/new/fix/auto-terminate-detached-silent-runs  
[...]

PR #10325, opened 2026-07-27
fix(heartbeat): reap detached runs after critical output silence

wassano:
## Thinking Path

> - The heartbeat supervisor spawns local adapter child processes (claude_local, codex_local, opencode_local, …) and tracks them in-memory plus via persisted `processPid`/`processGroupId`
> - When the in-memory handle is lost while the child pid is still alive, `reapOrphanedRuns()` marks the run `process_detached` — and then `continue`s, forever; no kill is ever scheduled no matter how long the run stays silent
> - A hung provider stream (e.g. a stalled LLM SSE stream) therefore wedges a run in `running` state indefinitely: it holds its issue checkout/execution lock, and recovery requires a manual `kill` on the exec host
> - This pull request adds a reaper path for exactly that case: detached + output-silent past the critical threshold + no active operator snooze ⇒ terminate the recorded pid/process group (SIGTERM, then SIGKILL after the grace window) and fail the run through the existing process-loss finalization path
> - The benefit is that orphaned adapter processes are reaped automatically within one reaper sweep of crossing the critical silence threshold, while detached runs that still produce output, recover activity, get hot-restart-adopted, or were deliberately snoozed/continued by an operator are never killed

- Refs #6406 — same root cause (`process_detached` runs hang forever). This PR implements the R1 proposal with two extra safety gates on top of the basic silent-terminate: (a) termination is gated on the run's *output silence* crossing `ACTIVE_RUN_OUTPUT_CRITICAL_THRESHOLD_MS` (4h) rather than a short grace, so a live-but-detached child that is still streaming output is never killed; (b) an active operator `snooze`/`continue` watchdog decision suppresses the reap.
- Related prior attempts (all appear stale): #8539 (reaps in the stale-lock sweep instead of the orphan reaper), #6273 (kills 5 min after detach regardless of output), #4605, #6232.

## What Changed

  - `reapOrphanedRuns()`: after marking/confirming `process_detached` with a live pid, compute the run's output-silence age (`silenceAgeMsForRun`); if it is past the critical threshold and `hasActiveOutputQuietDecision()` is false, call `terminateHeartbeatRunProcess()` (existing SIGTERM → grace → SIGKILL escalation) and fall through to the shared process-loss finalization path (status `failed`, `errorCode: "process_lost"`, single `enqueueProcessLossRetry`, wakeup failure, environment-lease release, issue checkout release).
  - New `hasActiveOutputQuietDecision()` query — mirrors `latestActiveOutputQuietUntilDecision` in recovery/service.ts so an operator snooze/continue suppresses the reap.
  - New `buildDetachedProcessReapMessage()` and a `detachedProcessReaped`/`detachedSilenceMs` payload on the finalization run event for observability.
  - New `isRecordedPidRecycledByNewerProcess()` gate — before signaling, the live process's start time (`ps -o lstart=`) is compared with the `processStartedAt` persisted at spawn; a materially newer start (>60s) proves the pid was recycled by an unrelated process, so no signal is sent and the run is finalized as `process_lost` with a `detachedPidRecycled` event flag. The descendant-group cleanup is likewise suppressed when the recorded group is currently led by the recycled pid (group ids cannot be recycled while any member holds them, so a surviving group not led by the recycled pid is still held by our own descendants). Suppression requires positive identification; indeterminate cases keep the historical behavior of every other `terminateHeartbeatRunProcess` call site.
  - The pid-dead and descendant-only-cleanup branches are unchanged.
- `server/src/services/recovery/service.ts`
  - Hoisted `silenceStartedAtForRun`/`silenceAgeMsForRun` to module scope and exported them (single source of truth for "what counts as output silence"); no behavior change.
- `server/src/__tests__/heartbeat-process-recovery.test.ts`
  - New: reaps a detached run once output silence crosses the critical threshold (asserts SIGTERM escalation via `terminateLocalService`, real child exit, `process_lost` failure, reap event payload).
  - New: does not reap a detached run before the critical silence threshold.
  - New: does not reap a critically silent detached run while the watchdog is snoozed.
  - New: finalizes a critically silent detached run without signaling a recycled pid (recorded spawn time hours older than the live process start; asserts `terminateLocalService` never called, stranger process still alive, `detachedPidRecycled` payload).
  - New: does not signal the recorded process group when it is led by the recycled pid's new owner (detached stranger = group leader).
  - New: reaps normally when the live process start matches the recorded spawn time (identity confirmed).
  - Updated: "keeps a local run active when the recorded pid is still alive" now seeds fresh `lastOutputAt` — the old version of this test asserted the exact hang-forever behavior #6406 reports (ancient `startedAt`, no output, still kept alive); with the fix that scenario is reaped by design, so the test now covers the "alive **and** producing output" case it was named for.

## Verification

- `vitest run --project @paperclipai/server src/__tests__/heartbeat-process-recovery.test.ts` — 94/94 pass (including the 3 new tests; the reap test spawns a real `node -e setInterval` child and asserts it is actually terminated).
- `vitest run --project @paperclipai/server src/__tests__/recovery-classifiers.test.ts src/__tests__/recovery-observability.test.ts` — 17/17 pass.
- Reviewer spot-check: seed a `running` run with a live pid, `errorCode = 'process_detached'`, `lastOutputAt` 5h in the past; call `reapOrphanedRuns()`; expect the run failed with "Detached child process reaped after critical output silence" and the pid dead.

## Risks

- Behavioral shift by design: a detached run whose output has been silent ≥4h (with no operator snooze) is now killed and failed instead of being left running forever. That is the intended fix for #6406, but operators who relied on multi-day silent detached runs eventually resuming will see them failed (with one automatic retry) instead.
- Retry semantics: reaped runs go through the standard single `process_lost` retry, same as a pid-dead orphan — worst case one duplicate heartbeat attempt, bounded by the existing `processLossRetryCount < 1` guard.
- `terminateHeartbeatRunProcess` targets the recorded `processGroupId` when present (SIGTERM to the group, then SIGKILL) — same primitive already used for cancellation and descendant-only cleanup. Pid/pgid-reuse risk on the new 4h-silence path is gated by the start-time identity check above (suppression only on positive identification; indeterminate cases keep existing behavior).
- Low migration risk: no schema changes, no config changes; the critical threshold constant is the one the stale-run watchdog already pages on.

## Model Used

- Moonshot AI Kimi K3 (exact model ID `moonshotai/kimi-k3` via OpenRouter), agentic tool-use mode (shell, file editing, test execution), orchestrated by a Paperclip DevOps agent heartbeat.

## Checklist

- [x] I have searched GitHub for duplicate or related PRs and linked them above (#6406, #8539, #6273, #4605, #6232)
- [x] I have updated relevant documentation to reflect my changes (no doc changes needed — recovery behavior only; runbook-facing behavior is described in the run event payload and error message)

### DIFF
--- a/server/src/__tests__/heartbeat-process-recovery.test.ts
+++ b/server/src/__tests__/heartbeat-process-recovery.test.ts
@@ -27,6 +27,7 @@ import {
   executionWorkspaces,
   heartbeatRunEvents,
   heartbeatRuns,
+  heartbeatRunWatchdogDecisions,
   issueComments,
   issueDocuments,
   issuePlanDecompositions,
@@ -1194,6 +1195,11 @@ describeEmbeddedPostgres("heartbeat orphaned process recovery", () => {
       processPid: child.pid ?? null,
       includeIssue: false,
     });
+    // Fresh output keeps the detached run below the critical-silence reaper threshold.
+    await db
+      .update(heartbeatRuns)
+      .set({ lastOutputAt: new Date() })
+      .where(eq(heartbeatRuns.id, runId));
     const heartbeat = heartbeatService(db);
 
     const result = await heartbeat.reapOrphanedRuns();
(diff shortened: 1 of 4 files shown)


wassano:
Addressed the recycled-identifier concern in 7a8cf48 (force-pushed onto this branch as a regular commit):

- **PID identity check before signaling** — the detached-reap path now compares the live process's start time (`ps -o lstart=`) against the `processStartedAt` persisted via the adapter `onSpawn` hook. A live start materially newer (>60s) than the recorded spawn means the pid was recycled by an unrelated process, so no signal is sent; the run is finalized through the normal process-loss path with a `detachedPidRecycled` flag on the finalization event and a dedicated message.
- **Process-group recycling** — the descendant-group cleanup is suppressed when the recorded group is currently led by the recycled pid (a group id is its leader's pid and cannot be recycled while any member holds it, so a surviving group *not* led by the recycled pid is still held by our own descendants and is cleaned up as before).
- **Suppression is fail-open only when indeterminate** — recycling must be positively identified (recorded spawn time present, `ps` readable, output parseable). Otherwise behavior matches the existing `terminateHeartbeatRunProcess` call sites (cancellation, descendant-only cleanup), which never verify identity.

Coverage: +3 tests — recycled pid is finalized without any signal (stranger process verified alive afterward); recycled pid leading the recorded group leaves the group unsignaled; matching start time reaps normally. `heartbeat-process-recovery.test.ts` 97/97, `@paperclipai/server` typecheck clean.

PR #9633, opened 2026-07-15
feat: organize skills with nested folders and My Skills

cryppadotta:
## Thinking Path

> - Paperclip is the open source control plane people use to organize and govern AI-agent companies
> - Company skills are durable resources that users browse, import, assign, and maintain over time
> - A flat skill list plus tags does not provide a stable location or hierarchy for personal, company, project-imported, and bundled skills
> - Folder paths need to be canonical, company-scoped, safe to move, and preserved across re-imports without changing skill IDs
> - The `/skills` UI also needs traversal, breadcrumbs, move/create flows, and a dedicated My Skills namespace that work on desktop and mobile
> - This pull request adds the folder data model and APIs, reserved-root lifecycle, project import behavior, and the folder-first skills experience
> - The benefit is a predictable filesystem-like organization model while tags remain available for cross-cutting classification

## Linked Issues or Issue Description

Refs #9619 — the reviewed folder foundation was intentionally closed and folded into this combined feature PR.
Refs #9026 — earlier flat-folder attempt superseded by this integrated implementation.
Refs #3281 — related skill organization proposal; this PR uses canonical persisted folders rather than deriving groups from skill keys, and does not add hidden-skill behavior.

**Feature request**

- **Problem:** Skills currently lack a canonical hierarchical location, making personal skills, project imports, bundled skills, and company-authored skills difficult to traverse and manage at scale.
- **Proposed behavior:** Add nested company-scoped folders with stable paths, reserved My/Projects/Bundled roots, subtree queries, safe move/create operations, and a folder-first `/skills` library UI.
- **Import behavior:** New project scans file skills under `projects/<project-slug>`; later imports update content without overriding a user-selected folder.
- **Alternatives considered:** Tags alone remain useful for cross-cutting classification, but they do not provide canonical location, nesting, reserved namespaces, or stable import placement.
- **Roadmap alignment:** Extends the completed Skills Manager and Scheduled Routines capabilities without duplicating an active roadmap item.

## What Changed

- Adds `folders` persistence for routine and skill folders, nested canonical paths, parent/slug/system-key fields, migration backfills, and reapply-safe migrations `0174`–`0175` after current master migrations.
- Adds company-scoped folder CRUD, cycle/depth/namespace validation, reserved My/Projects/Bundled lifecycle, item moves, subtree filtering, and folder paths on skill results.
- Preserves project-import placement: first import files into the project folder, while re-import keeps user-owned placement and stable skill IDs.
[...]

PR #10328, opened 2026-07-27
fix(company-skills): approve managed_checkout project dirs for local skill import (LOOA-850)

nguyenm7:
## What
`assertLocalImportSourceAllowed` (server/src/services/company-skills.ts) built `configuredRoots` from only the managed skills root and **registered** `project_workspaces[].cwd`. Every `managed_checkout` project (`primaryWorkspace: null`, `workspaces: []`) had its server-derived `codebase.managedFolder` excluded, so in-place skill re-imports were denied with `skill_workspace_boundary_denied`.

This adds `project.codebase.managedFolder` to `configuredRoots`. `managedFolder = resolveManagedProjectWorkspaceDir({ companyId, projectId })` is fully server-derived from the instance root + sanitized ids — same trust class as the pre-existing `resolveManagedSkillsRoot`. It is `fs.realpath`-resolved and `${root}${sep}`-prefix matched exactly like every other root, so no traversal/escape is reintroduced.

## Why it's safe (Security sign-off on LOOA-850)
- **Server-derived, not attacker-derived.** `companyId` = request tenant scope; `projectId` from `projects.list(companyId)` (that company's rows only); `repoName` derived via `new URL(...)`. None comes from the import `source` argument.
- **Separator-safe.** `sanitizeFriendlyPathSegment` maps `[^a-zA-Z0-9._-]+ -> -`, so no segment can inject `/`/`\` or a new path level.
- **`..` traversal unreachable** — the only user-influenced segment (`repoName`) is normalized by `new URL(...)`.
- Prefers `managedFolder` over `effectiveLocalFolder` (the latter can fall through to a user-registered `localFolder`, already covered by `workspaces[].cwd`).

## Test
`company-skill-import-boundary.test.ts` adds a regression case: a `managed_checkout` project's `managedFolder/<skill>` import is **allowed**, and a **prefix-adjacent sibling** (`${managedFolder}-evil`) stays **denied** — confirming `${root}${sep}`-anchored matching (not bare `startsWith`). 2/2 pass; verified independently by CTO on the live tree.

## Impact
Generically repairs every `managed_checkout` project's in-place skill re-import (the ~16 Signal Aggregator siblings + Onboarding/Editorial/Growth State/Content Strategy/Talks/Gateway Chat). The per-skill managed-root mirror (LOOA-849 Option B, routine `6a2f6060`) becomes optional rather than required.

Refs: LOOA-850 (systemic fix), delegated from LOOA-849. Boundary hardening history: #9564 / #9620 / #9633.

### DIFF
--- a/server/src/__tests__/company-skill-import-boundary.test.ts
+++ b/server/src/__tests__/company-skill-import-boundary.test.ts
@@ -6,6 +6,7 @@ import { afterAll, afterEach, beforeAll, describe, expect, it } from "vitest";
 import { companies, companySkills, createDb, projects, projectWorkspaces } from "@paperclipai/db";
 import { getEmbeddedPostgresTestSupport, startEmbeddedPostgresTestDatabase } from "./helpers/embedded-postgres.js";
 import { companySkillService } from "../services/company-skills.js";
+import { resolveManagedProjectWorkspaceDir } from "../home-paths.js";
 
 const embeddedPostgresSupport = await getEmbeddedPostgresTestSupport();
 const describeEmbeddedPostgres = embeddedPostgresSupport.supported ? describe : describe.skip;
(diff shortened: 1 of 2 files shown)


nguyenm7:
Superseded by #10329 — same change (identical tree), reopened on a branch whose name and commit message follow the contribution guidelines (no internal ticket ids) and with a PR description matching the template. Closing this one.

PR #10352, opened 2026-07-28
perf(plugin-daytona): opt-in no-profile fast path for default-PATH execs

nickyleach:
## Thinking Path

> - Paperclip is the control plane for autonomous AI companies
> - The Daytona adapter turns tasks into shell commands and manages execution overhead
> - Many short-lived exec calls still pay for login-shell profile sourcing even when the binary already resolves on the sandbox default PATH
> - That extra startup work adds latency on the hot path for repeated command execution
> - This pull request adds an opt-in fast path that skips profile sourcing only when the caller explicitly requests it and the command does not need shell initialization
> - The benefit is lower per-call latency for eligible commands without changing the conservative default behavior for commands that need the profile

This change does not reference a public GitHub issue. It follows the same Daytona startup-speed work as merged PR #10335 and narrows the execution path for eligible commands while keeping the default login-shell behavior intact.

## What Changed

- Added an optional `noProfile` flag to `PluginEnvironmentExecuteParams`.
- Refactored Daytona login-shell script assembly so the profile and nvm sourcing block is omitted only on the explicit fast path.
- Preserved environment prefixing, `cd`, shell quoting, `NONINTERACTIVE_GIT_ENV`, stdin handling, and `durationMs` behavior on both paths.
- Added regression tests for the fast path omission, the preserved execution parameters, and the default profile-sourcing path.

## Verification

- `pnpm --filter @paperclipai/sandbox-provider-daytona exec vitest run src/plugin.test.ts`
- `pnpm --filter @paperclipai/plugin-sdk tsc --noEmit`
- Reverted the guard locally to confirm the two behavior tests fail again, then restored the change.

## Risks

- If a caller opts into `noProfile` for a command that depends on shell initialization, the command can fail to resolve its binary.
- The API comment and opt-in design keep that risk narrow; the default path remains unchanged.

## Model Used

OpenAI GPT-5 (Codex tool-using coding agent)

## Checklist

### DIFF
--- a/packages/plugins/sdk/src/protocol.ts
+++ b/packages/plugins/sdk/src/protocol.ts
@@ -651,7 +651,7 @@ export interface PluginEnvironmentExecuteParams extends PluginEnvironmentDriverB
   env?: Record<string, string>;
   stdin?: string;
   timeoutMs?: number;
-  /** Skip the Daytona login-shell profile sourcing when the command already resolves on the sandbox default PATH. */
+  /** Skip login-shell profile sourcing when the command already resolves on the sandbox default PATH. */
   noProfile?: boolean;
 }
 
--- a/packages/plugins/sdk/src/protocol.ts
+++ b/packages/plugins/sdk/src/protocol.ts
@@ -651,6 +651,8 @@ export interface PluginEnvironmentExecuteParams extends PluginEnvironmentDriverB
   env?: Record<string, string>;
   stdin?: string;
   timeoutMs?: number;
+  /** Skip the Daytona login-shell profile sourcing when the command already resolves on the sandbox default PATH. */
+  noProfile?: boolean;
 }
 
 export interface PluginEnvironmentExecuteResult {
--- a/server/src/services/environment-runtime.ts
+++ b/server/src/services/environment-runtime.ts
@@ -1744,6 +1744,7 @@ function createPluginEnvironmentDriver(
           env: input.env,
           stdin: input.stdin,
           timeoutMs: input.timeoutMs,
+          noProfile: input.noProfile === true,
         },
       });
     },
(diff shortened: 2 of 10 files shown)


nickyleach:
General tests are currently failing on this PR. The blocker is `@paperclipai/adapter-utils src/command-managed-runtime.test.ts > command managed runtime > keeps adapter detection on the profile-backed shell path`, which expects a single shell invocation but is seeing seven. `verify` is failing only because it depends on that lane. I’m pausing merge-readiness until that is fixed.

nickyleach:
@greptile review

packages/plugins/sdk/src/protocol.ts
```diff
@@ -651,6 +651,8 @@ export interface PluginEnvironmentExecuteParams extends PluginEnvironmentDriverB
   env?: Record<string, string>;
   stdin?: string;
   timeoutMs?: number;
+  /** Skip the Daytona login-shell profile sourcing when the command already resolves on the sandbox default PATH. */
```
nickyleach: This comment sets a bad precedent. We shouldn't have provider-specific behaviors in the protocol. I think the simple fix here is to update the text so that it removes the reference to Daytona.

server/src/services/environment-runtime.ts
```diff
@@ -188,6 +188,11 @@ export interface EnvironmentDriverExecuteInput extends EnvironmentDriverLeaseInp
   env?: Record<string, string>;
   stdin?: string;
   timeoutMs?: number;
+  /**
+   * Skip the Daytona login-shell profile sourcing for commands that already
```
nickyleach: We also need to remove `Daytona` from this comment. The `environment-runtime` can't have provider-specific options exposed, that would be a mess.

PR #10353, opened 2026-07-28
fix(issues): hand user-assigned issues back to the asking agent on answer/accept

juancarlosrial76-code:
## Rebase 2026-08-25 — squashed onto current `master`

The branch had drifted 543 commits behind and was `CONFLICTING`. It is now a single
commit on `d1573244b`, same net change. Three things had to be reconciled rather than
merely re-applied, and each is a real adjustment worth reviewing:

1. **`answerQuestions` now runs in a transaction.** `master` had removed the wrapper;
   the handoff has to write inside the same transaction as the resolution, so it is
   back. `resolvedByRunId` and the `interactionAlreadyResolvedError()` helper that
   `master` added on that path are preserved.
2. **The locked re-reads are respected.** `acceptRequestConfirmation` now takes
   `FOR UPDATE` on the issue and on the interaction; the handoff is passed the locked
   interaction row, not the pre-transaction read.
3. **The compare-and-swap guard was generalised.** It previously pinned
   `assigneeAgentId IS NULL` plus the exact `assigneeUserId`, which was correct when the
   handoff only fired for user-assigned issues. `master` has since added a second
   qualifying shape — an agent-owned issue in review returning to its own creator — so
   each assignee column is now pinned to whichever value was actually judged. Reverting
   this pin alone turns `returns accepted agent confirmations from review without
   resetting active work` red, so the case is covered.

**Consequence worth naming:** because that in-review branch sits below the generalised
kind gate, an answered `ask_user_questions` or accepted `suggest_tasks` on an agent-owned
issue in review — assigned to the creator agent, with a `wake_assignee*` policy — now
also returns that issue to `todo`. Nothing is taken from anyone (the assignee is already
that agent) and it matches the intent of the in-review branch, but it is broader than the
original scope of this PR and is called out rather than left to be discovered.

**Verification after the rebase** — the local checkout's dependency tree is a zod major
behind `master`, so it is only a partial signal, and the halves are reported separately:

| Run | Result |
|---|---|
| `issue-thread-interactions.test.ts` + `issue-thread-interactions-service.test.ts` | **79/79 pass** (same files on unmodified `master`: 69/69 — the 10 added tests are this PR's) |
| `issue-thread-interaction-routes.test.ts` | **cannot run locally**, 72/72 error on `z.string(...).guid is not a function`. Unmodified `master` fails the same file 70/70 with the same error, so this is the local dependency tree, not the change. CI is the authority here. |
| `tsc --noEmit` on `server` | 1253 diagnostics on this branch, **1253 on unmodified `master`**, set difference empty in both directions — no new type error. The 1253 are the same stale-dependency artefact. |
| Mutation probes (all restored afterwards, suite green again) | narrowing the kind gate back to confirmations → **6 red**; removing the wakeup-policy gate → **2 red**; reverting the assignee pin → **1 red** |

## Thinking Path

> - Issue-thread interactions (`ask_user_questions`, `suggest_tasks`, `request_confirmation`) are how an agent hands a decision to a human and resumes afterwards; `continuationPolicy` decides whether resolving one wakes the agent back up
> - That resume only happens if the issue carries an `assigneeAgentId` at the moment the interaction resolves — `queueResolvedInteractionContinuationWakeup` drops the wakeup outright when there is no agent target
> - On an issue assigned to a **user** that target cannot exist while the question is open: an issue has exactly one assignee, and taking it off the human before they answer removes the very thing that puts the question in front of them
> - `request_confirmation` already solved this at the only moment it can be solved — resolution time — by handing the issue back to the asking agent; the other two kinds were never wired into it, so an answered question or an accepted task list on a user-assigned issue resolved into silence
> - This pull request extends that handoff to answered `ask_user_questions` and accepted `suggest_tasks`, gated on a `wake_assignee*` policy, and performs it as a guarded compare-and-swap so a concurrent reassignment or close cannot be overwritten
> - The benefit is that a human's answer actually reaches the agent that asked for it, instead of sitting unread in the thread while the agent stays asleep

No existing public issue — described in-PR per the bug report template.

Refs #10274 — complementary and independent: that PR fixes genuinely **ownerless** issues on the *creation* path; this one fixes **user-assigned** issues on the *resolution* path. Both introduce a small `isWakeAssigneeContinuationPolicy` helper, so whichever lands second should drop the duplicate.

Refs #10278 — adjacent, and worth reading together. It makes a *reasoned rejection* wake the assignee regardless of `continuationPolicy`. That changes the wake **gate**; this PR supplies the wake **target**. On a user-assigned issue the two compose: without a handoff, a rejection wake that #10278 now allows to fire is still dropped for want of an `assigneeAgentId`.

**Bug description** (per `bug_report.yml`):
- **What happened:** A board user answers an agent's `ask_user_questions`, or accepts its `suggest_tasks`, on an issue assigned to that user. The interaction resolves, but no wake is queued. The agent never learns it was answered.
- **Expected:** The asking agent is woken with the answer, as it already is for an accepted `request_confirmation`.
- **Where:** `server/src/services/issue-thread-interactions.ts` — the handoff was gated on `isRequestConfirmationLikeKind`.
- **Impact:** The issue sits with the human, the answer sits unread, and nothing moves it. It does not self-heal: the wake fires once, at resolution, and is discarded.

## What Changed

- Widened the predicate (now `shouldReturnResolvedInteractionToCreatorAgent`) to cover answered `ask_user_questions` and accepted `suggest_tasks`.
- Gated those two kinds on a `wake_assignee*` continuation policy. Both can legitimately be posted with no intent to resume — an informational question, a task list the board owns — and taking the issue off its human owner for a continuation nobody asked for would be worse than the stall. `request_confirmation` keeps its policy-independent behaviour: the creating agent is blocked on the verdict by construction.
- Extracted the handoff into one helper (`returnResolvedIssueToCreatorAgent`) shared by all three resolution paths, running inside each caller's transaction so a resolved interaction and the wake target it depends on commit together.
- `answerQuestions` gained that transaction and now returns `{ interaction, continuationIssue }`, matching `acceptInteraction`. The `respond` route wakes the post-handoff assignee instead of the stale pre-resolution row.
- The `issue.updated` activity log for the handoff is shared and names the resolution that moved the issue (`ask_user_questions_answer`, `suggest_tasks_accept`); `request_checkbox_confirmation` keeps the long-standing `request_confirmation_accept` source.
- **Addressing the Greptile review:** the handoff write is now a compare-and-swap. The eligibility probe is an unlocked read, so the issue could change between it and the write; the previous update was by id alone and would silently overwrite a concurrent reassignment — worst case reopening a `done` issue to `todo`. The mutable preconditions are now re-asserted in the `WHERE` clause and the update reports whether it matched. Losing the race returns `null`, so the caller falls back to `touchIssue` exactly as it does for an ineligible resolution: the interaction still resolves, only the handoff is skipped.

### Why a `WHERE` guard and not `SELECT … FOR UPDATE`

This runs inside the caller's transaction. Taking a row lock here and then calling the fat `issueService.update()` — which also touches labels, blockers, activity and workspace settings — is precisely what deadlocked against the `heartbeat_runs` FK cascade in #10274 (`40P01`, twice, on the serialized shard). A narrow conditional statement keeps the lock footprint to the one row.

Pinning `status` to the value the probe read covers both hazards in a single predicate: it rejects a concurrent terminal transition, and it guarantees `returnStatus` — derived from that same read — cannot unblock an issue that just became blocked.

Dropping the `issueService.update()` call also drops its side effects. That is a deliberate trade, and the only one this path relied on is the `issue.updated` activity, which `logInteractionCreatorHandoff` in `routes/issues.ts` already writes explicitly.

## Scoping

The handoff stays deliberately narrow. It fires only when **all** of these hold:

- a *user* resolved the interaction,
- the issue was assigned to a user and to **no** agent,
- the issue is still open (not `done` / `cancelled`),
- the interaction has a `createdByAgentId`.

Board approvals stay board business: an issue already owned by an agent already has a live wake target and is never reassigned. A `blocked` issue is handed back still `blocked`, not silently moved to `todo`.

## Verification

```sh
cd server && npx tsc --noEmit                                          # clean
npx vitest run server/src/services/issue-thread-interactions.test.ts \
              server/src/__tests__/issue-thread-interactions-service.test.ts \
              server/src/__tests__/issue-thread-interaction-routes.test.ts   # 69/69 pass
```

`server/src/__tests__/issue-thread-interactions-service.test.ts` — `describe` block against embedded Postgres:

- answered `ask_user_questions` on a user-assigned issue → `continuationIssue` set, issue moved to the creating agent, `assigneeUserId` cleared (with an explicit assertion that the creation path left the human owner alone)
- the same on a `blocked` issue → handed back still `blocked`
- `continuationPolicy: "none"` → no handoff, issue stays with its human owner
- issue already assigned to an agent → no handoff
- accepted `suggest_tasks` on a user-assigned issue → handoff plus the created child issue
- board-authored `suggest_tasks` (no `createdByAgentId`) → no handoff

`server/src/__tests__/issue-thread-interaction-routes.test.ts` — two route tests asserting the wake reaches the *returned* agent (the pre-resolution issue has no assignee agent, so a wake against it would have been dropped) and that the `issue.updated` activity records the right source.

`server/src/services/issue-thread-interactions.test.ts` — the fake-db harness routes the assignee probe by selected columns, defaulting to an agent-owned issue so the handoff stays out of the way of unrelated tests. Two new cases cover **both outcomes of the compare-and-swap** — guarded update matches (handoff, no extra touch) and guarded update matches nothing (no handoff, falls back to `touchIssue`). The race window cannot be scheduled deterministically against real Postgres, so it is covered here; the harness routes the handoff write by its assignee columns, since it also carries `status`.

## Risks

- **Behavioural shift, intentionally narrow.** An answered question or accepted task list on a user-assigned issue now moves that issue off the human and onto the asking agent. That is the point of the change, but it is a real change in who owns the card. It is bounded by the four conditions under *Scoping*, and by the `wake_assignee*` policy gate for exactly the two newly-covered kinds.
- **The compare-and-swap can decline to act.** Under concurrency the handoff may be skipped rather than applied. That is the deliberate safe direction — the alternative is overwriting a newer decision — but it means the wake can still be missed in a genuine race. The interaction is resolved either way, so the state is correct; it is a liveness, not a correctness, edge.
- **Loss of `issueService.update()` side effects** on this path, as described above. Reviewed against that service: the activity log was the only one that mattered here, and the route writes it directly.
- No schema change, no migration, no new inputs.

## Model Used

- Anthropic Claude — Opus 5 (`claude-opus-5`), 1M context, extended thinking, agentic tool use via Claude Code (CLI). Diagnosis, implementation, tests and this description model-authored under human direction.

## Checklist

<sub>Two boxes are left unchecked on purpose. "Run tests locally" is checked for the two service suites only — the routes suite cannot execute in this checkout and fails identically on unmodified `master`; see the rebase table above. The branch name carries an originating ticket id and cannot be renamed without recreating this PR — happy to do that if you would rather not carry it. CI and Greptile are re-running against the rebased commit; I will not request merge until both are green.</sub>


juancarlosrial76-code:
Rebased onto current `master` (`3d23c3b2c`) and fixed the outstanding P1. Head is `6a01f013f`; the PR was `DIRTY`/`CONFLICTING` before, it is now `MERGEABLE`.

**Rebase.** Squashed the three commits into one — they were review iterations, and squashing turned three conflict resolutions into a single pass. The conflict was against #10251 (interaction withdrawal + terminal-issue expiry), which added a `getForIssue` helper and four tests in exactly the regions this branch touches. Both sides were additive, so the resolution keeps both; all four of #10251's tests are present and passing.

**P1 "New user assignment is overwritten" — valid, fixed.** The compare-and-swap guard was:

```js
eq(issues.status, issueContext.status),
isNull(issues.assigneeAgentId),
isNotNull(issues.assigneeUserId),   // <- "some user is assigned"
```

You're right that this is too weak. A user-A-to-user-B reassignment leaves *every* predicate true — the status is unchanged, no agent is assigned, and a user is still assigned — so the stale handoff would fire and clear user B. That silently takes the issue away from a human who was never asked anything, which is a worse failure than the one the handoff exists to fix.

The guard now pins the exact assignee it judged:

```js
eq(issues.assigneeUserId, judgedAssigneeUserId),
```

Losing that race behaves like every other lost race here: the handoff returns `null`, the caller falls back to `touchIssue`, and the interaction still resolves. Only the handoff is skipped, which is the safe direction.

New test: `pins the guard to the exact user assignee it judged, not merely 'some user'`. It inspects the bound parameters of the emitted WHERE predicate, so it fails if the guard is ever loosened back to `IS NOT NULL` — the race itself can't be scheduled deterministically, so asserting on the guard is what actually pins the behavior.

**Verification**
- `issue-thread-interactions.test.ts` — 9 passed (includes #10251's four).
- `issue-thread-interactions-service.test.ts` (embedded Postgres) — 46 passed.
- `issue-thread-interaction-routes.test.ts` — 24 passed.
- `tsc --noEmit` — clean for both touched source files.

One flake to flag honestly: `lists and creates board-authored interactions` in the routes suite intermittently exceeds the 5s default timeout (2 of 4 runs here). I checked it against plain `master` before blaming my branch — it fails there too (1 of 3 runs, same test, same ~5s boundary), so it is pre-existing and not introduced by this PR. Worth a separate timeout bump if it shows up in CI.

juancarlosrial76-code:
Pushed `ac1b26d51` — a defect in this PR that I found while scoping the follow-up for the rejection paths.

**The bug.** `isCreatorHandoffOnResolveKind` qualified every `request_confirmation`-like resolution *regardless of continuation policy*, on the reasoning that the asking agent is blocked on the verdict by construction. The intent is right; the mechanism does not deliver it. `queueResolvedInteractionContinuationWakeup` returns early unless the policy is `wake_assignee*` — and `request_confirmation` defaults to `continuationPolicy: "none"` (`packages/mcp-server/src/tools.ts`).

So for the *default* confirmation on a user-assigned issue, this PR took the issue off the person holding it and parked it on an agent that is never woken. That is the same dead wake path the PR exists to close, moved one step later — and strictly worse than before, because previously the user at least still held it.

**The fix.** Gate the handoff on a mirror of the wakeup guard, so the two agree by construction:

- policy must be `wake_assignee*`
- `wake_assignee_on_accept` only counts on `accepted`
- `expired` never counts

The resolved status is passed in explicitly rather than read off `current`, because callers run the handoff inside the same transaction as the row update and some only copy the new status onto `current` afterwards.

Regression test added: a default (`none`) confirmation accepted on a user-assigned issue must leave the assignee alone. It fails on the previous commit.

`tsc --noEmit` clean; 80 tests across the three affected suites pass.

This also settles the open question about the rejection paths: the invariant is *hand off exactly when a continuation wakeup would fire*. That makes `wake_assignee_on_accept` + reject correctly a no-handoff case — otherwise it would park the issue on an agent that by definition never gets woken.

PR #10376, opened 2026-07-28
fix(issues): hand user-assigned issues back to the asking agent on decline/cancel too

juancarlosrial76-code:
> **Stacked on #10353.** That PR is the base; only the commits `fix(issues): hand user-assigned issues back on the refusal paths too` and `fix(issues): audit the creator-agent handoff on the plugin path too` belong to this one. GitHub will not let a fork branch be a base on the upstream repo, so #10353's commits appear here as well. Once #10353 merges this rebases down to just those two. Reviewing it standalone before then means reviewing #10353 twice.

## Thinking Path

> - Agents ask people things through issue thread interactions — questions, task suggestions, confirmations — and declare a `continuationPolicy` saying whether resolving one should wake them back up
> - That wakeup needs a live target: it is dropped unless the issue carries an `assigneeAgentId`, and while a person still owes the answer the issue is assigned to *them*, so it carries only `assigneeUserId`
> - #10353 closed this for the acceptance paths by handing the issue back to the creating agent at resolution time, and named the refusal paths as a known remaining gap
> - Declining a confirmation, rejecting suggested tasks and cancelling questions still lose their wake exactly the same way — and a refusal is the case that most needs the agent back, since only it can revise the proposal the person just rejected
> - This pull request applies the same handoff to those three paths, under one uniform rule: hand off precisely when a continuation wakeup would actually fire
> - The benefit is that a declined proposal no longer strands the issue with the person who declined it, and no resolution can hand an issue to an agent that is never woken

No existing GitHub issue. Describing it here as a bug report:

**What happens.** An agent creates a `request_confirmation` with `continuationPolicy: "wake_assignee"` on an issue assigned to a board user. The user declines it. The route calls `queueResolvedInteractionContinuationWakeup`, whose guard requires `issue.assigneeAgentId`; the issue has only `assigneeUserId`, so the wake is discarded. The agent is never told its proposal was rejected. The same happens for `rejectSuggestedTasks` and for `cancelQuestions`.

**What should happen.** The asking agent is woken so it can revise, which means the issue has to go back to it — the same handoff #10353 introduced for the acceptance paths.

**Impact.** The issue sits with a human holding a rejected proposal that only the agent can act on. Nobody works it, and nothing surfaces the stall.

Refs #10353 — introduced the handoff helper this builds on and recorded this gap under "Known remaining gap".

## What Changed

- `rejectRequestConfirmation`, `rejectSuggestedTasks` and `cancelQuestions` now run inside a transaction like their acceptance counterparts, call `returnResolvedIssueToCreatorAgent`, and return `{ interaction, continuationIssue }`.
- The `reject` and `cancel` routes record the handoff via `logInteractionCreatorHandoff` and target the continuation wake at `continuationIssue ?? issue`.
- Handoff eligibility is unchanged: it already reduces to `wouldQueueResolvedContinuationWakeup(continuationPolicy, resolvedStatus)`, a deliberate mirror of the wakeup guard. Extending it to the refusal paths required no change to that helper.
- The activity `source` map moves from `routes/issues.ts` into the interactions service as `creatorHandoffActivitySource(kind, status)` and is keyed by `kind:status` rather than by kind alone — each eligible kind now hands off on two paths, and a decline labelled `..._accept` would misreport why the issue moved.
- `plugin-host-services` gets the same treatment on its chat resolution path: the reject branch retargets the continuation wake (the accept branch already did), and **both** branches now write the `issue.updated` audit entry for the reassignment, which neither did before.
- `skills/paperclip/references/api-reference.md` documents which resolutions hand off and which deliberately do not.

Resulting behaviour — the handoff fires on exactly the resolutions that queue a continuation wake:

| kind | hands off on |
|---|---|
| `ask_user_questions` | `answered`, `cancelled` |
| `suggest_tasks` | `accepted`, `rejected` |
| `request_confirmation` / `request_checkbox_confirmation` | `accepted`, `rejected` |

In every case only under `wake_assignee`, or under `wake_assignee_on_accept` when the outcome was an acceptance. `continuationPolicy: "none"` never hands off.

**Deliberately unchanged: `withdrawInteraction`.** A withdrawal is the asking side retracting its own card, not the answering side responding to it. There is no verdict to carry back and the agent that withdrew already knows, so a handoff would take the issue from a human for no reason.

## Verification

- `server/src/__tests__/issue-thread-interactions-service.test.ts` (embedded Postgres) — 56 passed, 10 new, in a `refusal paths` block beside the acceptance block from #10353. Covers the handoff itself, `blocked` preservation, both no-handoff policies (`none`, and `wake_assignee_on_accept` + reject), agent-owned issues, and agent-resolved refusals.
- **Negative control:** neutering the handoff (passing a `resolvedStatus` that never qualifies) fails exactly the four positive cases and leaves the six negative ones green — so the assertions pin the handoff rather than something incidental.
- `server/src/__tests__/issue-thread-interaction-routes.test.ts` — 26 passed, 2 new, asserting the post-handoff wake target and the `request_confirmation_reject` / `ask_user_questions_cancel` activity sources at the HTTP boundary.
- `server/src/__tests__/plugin-orchestration-apis.test.ts` — 28 passed, 2 new, asserting the `issue.updated` audit entry (with `_previous`) on the plugin decline path, and its absence when the issue keeps its assignee.
- `server/src/services/issue-thread-interactions.test.ts`, `issue-thread-interactions-telemetry.test.ts`, `built-in-agents.test.ts` — passing.
- `pnpm --filter @paperclipai/server exec tsc --noEmit` — clean.

Total 127 tests across the five affected suites.

One flake to flag, same as on #10353: `lists and creates board-authored interactions` in the routes suite intermittently exceeds the 5s default timeout. It is pre-existing — it reproduces on plain `master` — and unrelated to this change.

## Risks

- **Behavioural shift, intentional and the point of the PR.** A declined `wake_assignee` confirmation, rejected task suggestions, and cancelled questions now move a user-assigned issue to the creating agent (`assigneeUserId` cleared, status `todo`, or `blocked` preserved). A board user who declines a proposal on an issue they hold will see it leave their queue. This is the same trade #10353 accepted for the acceptance paths; the alternative is the issue stalling with nobody able to act on it.
- **Bounded by design.** No handoff when the policy is `"none"`, when the policy is `wake_assignee_on_accept` and the outcome is a refusal, when the issue is already agent-assigned, when the resolver is not a user, when there is no creating agent, or when the issue is terminal.
- **Concurrency: unchanged and best-effort.** Reuses the existing compare-and-swap guard, which pins status, agent assignment and the exact user assignee it judged. Losing that race skips only the handoff; the interaction still resolves. No new lock footprint — the guard is a `WHERE` clause, deliberately not `SELECT … FOR UPDATE`, to avoid the deadlock class from #10274.
- **Return-type change is internal.** `rejectInteraction`, `rejectSuggestedTasks` and `cancelQuestions` now return `{ interaction, continuationIssue }`. All callers are in this repo and updated; the HTTP response bodies are unchanged, so no API contract moves.
- **No migration.** Schema untouched.

## Model Used

Claude Opus 5 (`claude-opus-5`) via Claude Code, extended thinking enabled, with tool use (file editing, shell, test execution) against a local checkout.

## Checklist

- [ ] My branch name describes the change (e.g. `docs/...`, `fix/...`) and contains no internal Paperclip ticket id or instance-derived details — **not met:** the branch carries an instance ticket id. Renaming it would mean closing this PR and reopening, losing the review thread; happy to do that if a maintainer prefers.
- [x] Greptile is 5/5 with no open P2s, recommendations, or follow-ups — the 4/5 P1 (unaudited plugin handoff) was fixed in `bcb264139`


juancarlosrial76-code:
Pushed `bcb264139` and rewrote the PR description to the template.

**P1 "Plugin handoff mutation is unaudited" — valid, fixed.** When a plugin resolved an interaction that handed a user-assigned issue back to its creating agent, the only activity written was `issue.thread_interaction_accepted`/`_rejected`. The reassignment from a human to an agent, and the status move that comes with it, left no trace at all — the issue changed hands with nothing in the history to explain it, while the HTTP path has recorded a separate `issue.updated` for exactly this since #10353.

The report pointed at the reject branch, which is what this PR added. The accept branch has the same hole — it was introduced with the handoff in #10353 and simply never had an audit entry — so both branches now get one, written once after the resolution whenever a handoff actually happened:

```js
if (handoffIssue) {
  await logPluginActivity({
    action: "issue.updated",
    details: {
      status: handoffIssue.status,
      assigneeAgentId: handoffIssue.assigneeAgentId ?? null,
      assigneeUserId: handoffIssue.assigneeUserId ?? null,
      source: creatorHandoffActivitySource(resolved.kind, resolved.status),
      interactionId: resolved.id,
      _previous: { status: issue.status, assigneeAgentId: ..., assigneeUserId: ... },
    },
  })
}
```

`handoffIssue` is kept whole rather than folded straight into the wake target, because the audit entry needs the new user assignee and the wake target does not.

**The `kind:status → source` map moved.** It was local to `routes/issues.ts`; the plugin path has to label the same handoff identically, and two copies would drift. It now lives beside the eligibility rules in the interactions service as `creatorHandoffActivitySource`. Routes import it directly rather than through the services barrel — it is a pure lookup with no database dependency, and the route tests mock that barrel wholesale, so going through it would have forced a stub for no benefit.

**Tests.** Two in `plugin-orchestration-apis.test.ts`: one asserts the `issue.updated` entry with `_previous`, the resolved assignment and the `request_confirmation_reject` source; one asserts no such entry when the issue keeps its assignee.

Both needed a `settleQueuedWakeup` helper, and the reason is worth recording. These are the first tests in that suite to produce a *live* wake target — every earlier one deliberately avoided an assignee so the continuation wakeup was a no-op. `queuePluginInteractionContinuationWakeup` fires with `void heartbeat.wakeup(...)`, so the request row, and the heartbeat run that references it, land after `respondInteraction` has already returned and raced `afterEach` into foreign-key failures on `agents` and then on `agent_wakeup_requests`. The helper waits for both rows instead of deleting them: the run references the request, so only the suite's established cleanup order can remove them. It also returns the queued requests, which lets the first test assert the wake genuinely reached the creator agent — the thing that was silently dropped before the handoff existed.

**Verification.** 127 tests across the five affected suites (`issue-thread-interactions-service` 56, `issue-thread-interaction-routes` 26, `plugin-orchestration-apis` 28, plus the unit and telemetry suites). `tsc --noEmit` clean.

One thing I could not tick honestly in the checklist: the branch name carries an instance ticket id. Renaming means closing this PR and reopening, which loses the review thread — happy to do that if you would rather have it clean.

juancarlosrial76-code:
**P1 "Handoff audit records stale state" — valid, fixed in `f3b314855`. The HTTP path had the same hole.**

The report is right about the mechanism. Callers read the issue before entering the resolution; `returnResolvedIssueToCreatorAgent` re-reads it inside its transaction, and the compare-and-swap is judged against that *inner* read. When the two disagree, the handoff still succeeds — the guard is satisfied by the state it actually saw — and an audit entry built from the caller's snapshot names a former status or assignee the write never displaced. Wrong in exactly the case the entry exists for.

Rather than patch the reporting at the flagged call site, the stale source is gone: the handoff now returns what it displaced, and both audit sites read it.

```ts
type IssueCreatorHandoff = IssueWakeTarget & {
  previous: { status: string; assigneeAgentId: string | null; assigneeUserId: string | null };
};
```

populated from the in-transaction `issueContext`, so a caller can no longer supply a former state it is not in a position to know.

**Same defect on the HTTP path.** The comment pointed at `plugin-host-services.ts` because the plugin audit is what this PR added, but `logInteractionCreatorHandoff` in `routes/issues.ts` derived `_previous` from the route's `getById` result the same way. Both are fixed. This is the same shape as the earlier P1 on this PR, where the reported branch was one of two sharing the defect — worth noting since the review scope is the diff, and these holes sit just outside it.

**Coverage, and proof it discriminates.** A route test drives a deliberately divergent snapshot — a different status and a different user from what the handoff reports displacing — and asserts the audit follows the handoff. Reverting the source makes it fail with the caller's `in_progress` where the displaced `in_review` belongs, which is the reported defect verbatim. The eight service-level assertions pin `previous` to the in-transaction state with `toEqual`.

`tsc --noEmit` clean; 255 tests across the six suites that call the changed API.

One note on the previous round for completeness: `Verify serialized server suites (5/5)` was red on the prior head because `secret-proposals-routes.test.ts` matched `status` against the resolution wrapper after `rejectInteraction` started returning `{ interaction, continuationIssue }`. Fixed in `9bd444b02`; all ten test shards are green on that head.

juancarlosrial76-code:
Status note, and an offer rather than a force-push.

This is conflicted against `master` again — 566 commits of drift since the PR's base, 45 of them in the two files this touches. **The substance is not stale:** I re-checked `master` at `13368c518`, and the gap this PR closes is still open there. `rejectInteraction` / `cancelQuestions` still return a bare interaction, and only the acceptance path (`shouldReturnAcceptedConfirmationToCreatorAgent`) hands a user-assigned issue back to the agent that asked. A declined proposal on a user-assigned issue still drops its continuation wake silently.

I rebased this PR to green on 2026-08-25 and it stayed mergeable with all checks passing for three weeks. Rather than keep force-pushing a rebase every couple of weeks into a review queue, I would rather do it when it is actually useful: **say the word (or add a review) and I will rebase it within a day.** If the approach itself is wrong, or this is not something you want carried by an outside contributor, I would much rather hear that — it is a cheap thing to say and it saves the churn on both sides.

For what it is worth, the diff is small and the risky part is already settled: the eligibility rule mirrors the guard in `queueResolvedInteractionContinuationWakeup`, so a handoff happens exactly when a continuation wake would actually fire, and never on `wake_assignee_on_accept` + decline (which would park the issue on an agent nobody wakes).

PR #10387, opened 2026-07-28
fix(heartbeat): issue-scope the non-project fallback workspace (BRO-1717)

ericdfields:
## Problem

Non-project issues that do iterative local file work across multiple heartbeats — the normal comment-driven back-and-forth shape for ad-hoc issues (e.g. the `/img2threejs` skill iterating a local Three.js model via git worktrees) — lose git-worktree/branch continuity between runs.

`resolveWorkspaceForRun` provisions an execution cwd per run. When there is **no project workspace and no reusable session workspace**, it falls back to a dir keyed **only by `agentId`** (`<instanceRoot>/workspaces/<agentId>`). That single dir is shared across *every* non-project issue the agent touches, so an interleaved run on a different issue can trample it. Heartbeat N+1 then no longer sees the branch/commit heartbeat N produced and silently **re-derives the same work from scratch**.

Live evidence (run `c624b8f1-…`, CEO agent, BRO-1698): heartbeat N committed a fix to a branch; heartbeat N+1 (fired by the next comment) started a *different* worktree/branch that never saw that commit, re-applied the same fix, and burned the run to ~67.5% of its context window just getting back to where the prior heartbeat left off. In the worst case a run could exhaust its context mid-task without ever reaching a disposition.

This is not skill-specific — it's the execution-workspace provisioning path for **any issue without a project binding** that iterates across heartbeats.

## Fix

At the `agent_home` fallback site in `resolveWorkspaceForRun`, key the fallback dir by **issue as well as agent** (`<instanceRoot>/workspaces/<agentId>/<issueId>`) when an `issueId` is present, via a new `resolveIssueScopedAgentWorkspaceDir`. Successive heartbeats of the *same* issue now land in the *same* directory and reuse whatever git worktree/branch/commits the prior heartbeat left there. `fs.mkdir(..., { recursive: true })` makes reuse automatic once the path is stable. A non-conforming `issueId` (should never happen for real UUIDs; the segment regex is the path-traversal guard) degrades to the per-agent dir rather than failing the run.

This is analogous in spirit to `inheritExecutionWorkspaceFromIssueId` (used for child-issue handoffs), but for the same-issue-across-heartbeats case, which wasn't covered for non-project issues — that mechanism only helps when a DB-backed `executionWorkspaceId`/`projectWorkspaceId` exists, which the plain `agent_home` fallback never gets.

### Coupling handled

`resolveRuntimeSessionParamsForWorkspace` migrates a resumed session out of the fallback into a project workspace once one becomes available. It compared `previousCwd` against the **per-agent** fallback dir exactly, so issue-scoping the fallback would have silently broken that migration. It now uses a new `isAgentFallbackWorkspaceCwd` predicate that recognizes both the per-agent dir and an issue-scoped subdir of it.

### Intentionally left per-agent

- The **project-workspace-missing** fallback (`source: "project_primary"`) — project-bound issues expect the project cwd to return; issue-scoping there would fragment.
- The `agentHome` field — that is legitimately the agent's home, not an execution cwd.

The git-sensitive launch guard (`assertGitWorktreeBaseWorkspaceReady` / `assertGitSensitiveAdapterWorkspaceValid`) is untouched: issue-scoped dirs are only ever minted on the no-project `agent_home` path where `workspaceExpectation` is false, so that guard never sees them.

## Tests

- **New** `server/src/__tests__/agent-fallback-workspace-paths.test.ts` — path isolation across issues, stability across calls (the reuse property), nesting under the per-agent dir, path-traversal/non-segment rejection, and the `isAgentFallbackWorkspaceCwd` predicate (per-agent dir, issue-scoped subdir, other-agent negative, deeper-nesting/empty negatives).
- **Extended** `heartbeat-workspace-session.test.ts` — a session sitting in the issue-scoped fallback still migrates into a newly-available project workspace.

Both files pass (129 tests). Change is +85/-5 across `home-paths.ts`, `heartbeat.ts`, and the two test files.

Filed from BRO-1717 in the Brookfield Digital fleet.

### DIFF
--- a/server/src/__tests__/agent-fallback-workspace-paths.test.ts
+++ b/server/src/__tests__/agent-fallback-workspace-paths.test.ts
@@ -1,0 +1,93 @@
+import path from "node:path";
+import { afterEach, beforeEach, describe, expect, it } from "vitest";
+import {
+  isAgentFallbackWorkspaceCwd,
+  resolveDefaultAgentWorkspaceDir,
+  resolveIssueScopedAgentWorkspaceDir,
+} from "../home-paths.js";
+
+// Pin the instance root to a deterministic temp home so the resolved paths are
+// predictable regardless of the developer's real ~/.paperclip. See BRO-1717.
+const FAKE_HOME = "/tmp/bro1717-home";
+const AGENT_ID = "b3b6dde7-d283-47b7-8556-9eafe7ca9b52";
+const ISSUE_A = "c624b8f1-c69a-4ac2-8c73-33e2d6a945b1";
+const ISSUE_B = "a1111111-2222-3333-4444-555555555555";
+
+let savedHome: string | undefined;
+let savedInstance: string | undefined;
+
+beforeEach(() => {
+  savedHome = process.env.PAPERCLIP_HOME;
+  savedInstance = process.env.PAPERCLIP_INSTANCE_ID;
+  process.env.PAPERCLIP_HOME = FAKE_HOME;
+  delete process.env.PAPERCLIP_INSTANCE_ID; // → "default"
+});
+
+afterEach(() => {
+  if (savedHome === undefined) delete process.env.PAPERCLIP_HOME;
+  else process.env.PAPERCLIP_HOME = savedHome;
+  if (savedInstance === undefined) delete process.env.PAPERCLIP_INSTANCE_ID;
+  else process.env.PAPERCLIP_INSTANCE_ID = savedInstance;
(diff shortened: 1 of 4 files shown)


ericdfields:
Superseded by a new PR from a descriptively-named branch (`fix/issue-scoped-fallback-workspace`) with the internal ticket references removed and the required PR template filled out, per commitperclip/Greptile feedback. No code changes — same fix.

PR #8759, opened 2026-06-29
Fix monitored review blocker attention

saphid:
## Thinking Path
PR #8751 was closed because the source branch exposed an internal ticket id. This replacement keeps the exact scoped code commit but moves it to a public-safe branch name: `fix/monitored-review-blocker-attention`.

Dedup search completed before opening this replacement:
- [x] Searched existing PRs for similar/open replacements. Only prior match was closed PR #8751.

## What Changed
- Added blocker attention metadata to shared issue types.
- Included monitored review blocker attention in issue service responses.
- Added focused service coverage for blocker attention behavior.

## Risks
- Low risk. Diff is limited to issue response shaping and focused tests.
- Main risk is stale PR checks from reusing the original commit; fresh replacement PR checks are running on #8759.

## Model Used
codex_local / GPT-5.5

## Verification
- `pnpm exec vitest run server/src/__tests__/issue-blocker-attention.test.ts`
- `pnpm --filter @paperclipai/shared typecheck`
- `pnpm --filter @paperclipai/server typecheck`

Replaces closed PR #8751 from a public-safe branch name.

PR #8751, opened 2026-06-29
Fix monitored review blocker attention

saphid:
Closing this PR for cleanup.

This contribution was replaced by #8759 from a public-safe branch with a scoped commit. No maintainer action is requested here.

PR #10398, opened 2026-07-28
fix(issues): count a scheduled monitor as a covered blocker path

juancarlosrial76-code:
## Thinking Path

> - Blocker attention is the early-warning signal on that board: for every `blocked` issue it decides whether each blocking edge is `covered` (someone or something will resume it) or needs a human to step in
> - Two places in the server answer that same question — `listIssueBlockerAttentionMap` in `server/src/services/issues.ts` and the recovery liveness classifier in `server/src/services/recovery/issue-graph-liveness.ts` — and they answer it from different term lists
> - The recovery classifier counts a scheduled monitor as a waiting path; blocker attention does not, so a blocker parked on a monitor has no active run and no queued wake and falls through to the terminal `covered: false, stalled: false` branch
> - The result is a false positive on the healthiest cards on a board: a dependency that is already scheduled to resume on its own is reported as needing intervention, which is exactly the noise that makes the signal unusable
> - This pull request reuses the classifier's own exported predicate, `hasScheduledIssueMonitorPath`, inside `listIssueBlockerAttentionMap`, so both classifiers share one predicate instead of two drifting copies
> - The benefit is that a monitored blocker reads as `covered`, while every other attention case — including an idle assigned leaf with no wake path — keeps its current behaviour

No public GitHub issue exists for this; describing it here as a bug report.

**What happened.** On a live board, a `blocked` issue whose only explicit blocker was `in_progress` with `executionState.monitor.status = "scheduled"`, `monitorNextCheckAt` roughly 40 minutes out, and `attemptCount 0 / maxAttempts 12` was reported as `blockerAttention.state = "needs_attention"` with `attentionBlockerCount: 2, coveredBlockerCount: 0`. `sampleBlockerIdentifier` pointed at exactly that monitored blocker, which is what identifies it as the attention edge rather than an inference from the counts.

**What was expected.** The blocker has a live wake path — the harness re-enters it into its assignee's queue at `monitorNextCheckAt` with no human action — so it should classify as `covered`, the same way the recovery liveness classifier already treats it.

**Why it matters.** `blockerAttention` is a "who needs a human right now" signal. A blocker that is provably scheduled to resume is the one case that most looks healthy, so a false positive here trains operators to ignore the field.

## Overlapping open PRs — disclosed, not silently duplicated

Four other open PRs touch this lane. Measured against this branch on 2026-08-25:

| PR | Scope | Predicate |
|---|---|---|
| **#8759** (2026-06-29) | `in_review` blocker + `executionPolicy`/`monitorNextCheckAt` on `IssueRelationIssueSummary` | monitor fields on the relation summary |
| **#10502** (2026-08-02) | `in_review` lanes, also touches `issue-graph-liveness.ts` | monitor-backed review lanes |
| **#11847** (2026-08-21) | `in_review` branch only | bare `monitorNextCheckAt > now` |
| **#12035** (2026-08-23) | execution policy / watchdogs "live continuation"; overlaps the same test file | separate subject |
| **this PR** (2026-07-28) | **every** non-`done`, non-`cancelled` blocker node — `in_progress`, `todo`, `blocked`, `in_review` | reuses `hasScheduledIssueMonitorPath` |

Two differences are measured rather than asserted:

1. **Scope.** The three `in_review` PRs do not reach the case this bug report is about — an `in_progress` blocker whose only wake path is its monitor. On current `master` that edge still reads `needs_attention`.
2. **Exhaustion.** Replacing `hasScheduledIssueMonitorPath(row, now)` in this branch with the bare `monitorNextCheckAt > now` comparison that #11847 uses turns **1 test red**: `does not treat an elapsed or exhausted monitor as a covered waiting path`. A monitor past `timeoutAt` or at `attemptCount >= maxAttempts` never fires again, so counting it as covered hides exactly the dead wake path this map exists to surface.

Maintainers may still prefer one of the narrower PRs, or want #8759's relation-summary metadata layered on top — flagging the overlap rather than competing quietly. #8759's closed predecessor was #8751.

## What Changed

Rebased onto `d1573244b` on 2026-08-25 (523 commits of drift). **The `issue-graph-liveness.ts` change from the original version of this PR is gone**: `master` has since exported `hasScheduledIssueMonitorPath` itself, with the same validity rules this PR needed. The branch now touches one implementation file instead of two.

- `server/src/services/issues.ts`: inside `listIssueBlockerAttentionMap`, query the monitor columns for the candidate nodes (bounded by the existing chunking, filtered with `isNotNull(monitorNextCheckAt)`) and add every row satisfying `hasScheduledIssueMonitorPath` to `explicitWaitingIssueIds`, alongside pending interactions, pending approvals, open liveness escalations and open recovery actions.
- `server/src/services/issues.ts`: exclude `cancelled` nodes from `explicitWaitCandidateIds`. Tree cancellation does not clear `monitorNextCheckAt`, but `tickDueIssueMonitors` only wakes `in_progress`/`in_review` issues, so a cancelled node's retained monitor — and its pending interactions and approvals — cannot fire. The `assigneeUserId` branch below already carries the same guard.
- `server/src/__tests__/issue-blocker-attention.test.ts`: eight new cases (below).

Deliberately **not** changed: `in_progress` does not become covered by itself. An `in_progress` blocker with no run, no queued wake and no monitor is the signature of a crashed run, and the existing tests pinning an idle assigned leaf to `needs_attention` pass unchanged. Only a demonstrated wake path covers a blocker; a bare status does not.

## Verification

Run on the rebased head `74613ae6` (`server/`, local vitest):

```
vitest run src/__tests__/issue-blocker-attention.test.ts        # 33 passed (33)
vitest run src/__tests__/attention-service.test.ts \
  src/__tests__/issue-liveness.test.ts \
  src/__tests__/recovery-classifiers.test.ts                    # 48 passed (48)
```

**Not claimed as green:** `src/__tests__/heartbeat-issue-liveness-escalation.test.ts` fails to load locally with `z.string(...).guid is not a function` (`src/services/environment-config.ts:36`). A control run of the same file on unmodified `d1573244b` fails identically, so this is a local zod version skew, not this change. CI is the authority on that file.

**Typecheck, paired against unmodified `master`.** `tsc --noEmit` in `server/` on the head and on `d1573244b`, error lines normalised for line/column drift: `comm -23 head base` is **empty — 0 new type errors**. Both trees carry a large pre-existing error count from the local dependency state (752 head / 790 base error lines, the 38-line difference sitting entirely in `packages/adapters/*` and unrelated to this diff); `server/src/services/issues.ts` has **0** errors in both.

**Mutation probes, one per call site** (each applied to the finished branch, then reverted):

| Mutation | Expected | Result |
|---|---|---|
| revert `services/issues.ts` to `master`, keep the tests | new cases go red | **4 red** |
| drop only the `cancelled` guard from `explicitWaitCandidateIds` | cancelled case goes red | **1 red** |
| replace `hasScheduledIssueMonitorPath(row, now)` with bare `nextCheckAt > now` | exhaustion case goes red | **1 red** |

The second and third probes matter because both branches are individually green under the full revert for the wrong reason — the cancelled test passes on `master` only because `master` has no monitor→covered mapping at all to be fooled by.

New cases in `issue-blocker-attention.test.ts`:

- **positive** — `in_progress` blocker with a scheduled monitor → `covered`, `coveredBlockerCount: 1`, `attentionBlockerCount: 0`.
- **positive** — `in_review` blocker with a scheduled monitor → `covered`.
- **positive** — monitored blocker alongside a `cancelled` child on the same root.
- **negative** — elapsed `monitorNextCheckAt`, and exhausted `attemptCount >= maxAttempts`, stay `needs_attention`.
- **negative** — an `in_review` blocker whose monitor elapsed or was never scheduled stays flagged.
- **negative** — a `cancelled` blocker's retained monitor does not read as covered.
- **negative** — `blocked` with no blocker at all stays `needs_attention`.

## Risks

Low risk, with two things worth a reviewer's eye:

- **Widening risk** — the monitor lookup only ever *adds* ids to `explicitWaitingIssueIds`, so it can turn `needs_attention` into `covered` but never the reverse. If the predicate were too permissive it would hide real attention cases; that is why the validity rules (future `nextCheckAt`, unexpired `timeoutAt`, attempts remaining) are reused rather than reimplemented, and why the negative tests pin them.
- **Narrowing risk, in the other direction — measured, and the one call worth a maintainer's decision.** The `cancelled` exclusion sits on the shared `explicitWaitCandidateIds`, so it also removes cancelled nodes from the pending-interaction and pending-approval lookups. Greptile raised this as a P1 and the premise checks out: cancelling an issue does not cancel an approval already linked to it, so that row stays `pending`. Measured both ways — a `blocked` parent whose only blocker is `cancelled` with a `pending` approval reads **`covered` on `master`** and **`needs_attention` on this branch**. Pinned by `does not treat a cancelled blocker's pending approval as a covered waiting path`.
  I believe `needs_attention` is right: resolving that approval cannot move the blocker to `done`, so the parent's edge stays dead either way, and `master` already refuses to cover a cancelled node via the stronger `assigneeUserId` path (`node.assigneeUserId && node.status !== "cancelled"`). But it is scope beyond the monitor path this PR is named for. If you prefer the minimal read, the narrow variant is one line — leave `explicitWaitCandidateIds` at `status !== "done"` and filter `cancelled` inside the monitor loop only. Say the word and I will push it.
- **One extra query per chunk** — a `select` over the already-chunked candidate ids, narrowed by `isNotNull(monitorNextCheckAt)`. It sits in the same loop as the existing interaction and approval queries and follows their chunking, so it adds no new unbounded scan.
- No migration, no API shape change, no change to `unresolvedBlockerCount`.

## Model Used

Claude Opus 5 (`claude-opus-5`) via Claude Code, extended thinking, with tool use (repo search, local test runs, paired control runs, live-instance API reads).

## Checklist

- [x] I have searched GitHub for duplicate or related PRs and linked them above — re-run on 2026-08-25 after the rebase; four overlapping open PRs (#8759, #10502, #11847, #12035) are compared in the table above, plus #8759's closed predecessor #8751
- [ ] My branch name describes the change and contains no internal Paperclip ticket id or instance-derived details — the branch is `aur1528-blocker-attention-in-progress`; the `aur1528` prefix is an internal tracker id and should not be there. GitHub cannot switch the head branch of an open PR, so renaming would mean closing this PR and reopening a new one, discarding this review thread. Leaving it visible rather than silently ticking it; happy to re-open under `fix/monitored-blocker-covered-path` if a maintainer prefers that.
- [x] I have run tests locally and they pass — see Verification above, including the paired control runs and the three mutation probes
- [ ] I have updated relevant documentation to reflect my changes — no user-facing docs cover blocker-attention internals; the rationale lives in the code comments at the two changed sites
- [ ] All Paperclip CI gates are green — the round on `74613ae6` was 26 success / 1 skipped / 2 failure. `verify` is the aggregator; the single real failure was a 15 s timeout in `plugin-worker-manager.test.ts`, which timed out on the unrelated branch `fix/warn-on-adapter-cwd-fallback` in the same window. Evidence in the thread comment below; re-triggered by amending to an identical tree.
- [ ] Greptile is 5/5 with no open P2s, recommendations, or follow-ups — `Greptile Review` was `success` on `74613ae6`; re-running on the current head

### DIFF
--- a/server/src/__tests__/issue-blocker-attention.test.ts
+++ b/server/src/__tests__/issue-blocker-attention.test.ts
@@ -720,6 +720,78 @@ describeEmbeddedPostgres("issue blocker attention", () => {
     });
   });
 
+  // `in_review` leaves the shared covered/attention path early and lands in its own
+  // `stalled` branch, so it needs the monitor pinned separately from `in_progress`:
+  // a scheduled monitor is the standard re-review path for a review with no human
+  // reviewer, and without this it reads as a dead chain.
+  it("treats an in_review blocker with a scheduled monitor as a covered waiting path", async () => {
+    const { companyId, agentId } = await createCompany("PBMR");
+    const parentId = await insertIssue({ companyId, identifier: "PBMR-1", title: "Parent", status: "blocked" });
+    const reviewLeafId = await insertIssue({
+      companyId,
+      identifier: "PBMR-2",
+      title: "Monitored review leaf",
+      status: "in_review",
+      assigneeAgentId: agentId,
+      monitorNextCheckAt: new Date(Date.now() + 60 * 60 * 1000),
+      executionState: monitorState({ timeoutAt: new Date(Date.now() + 24 * 60 * 60 * 1000) }),
+    });
+    await block({ companyId, blockerIssueId: reviewLeafId, blockedIssueId: parentId });
+
+    const parent = (await svc.list(companyId, { status: "blocked" })).find((issue) => issue.id === parentId);
+
+    expect(parent?.blockerAttention).toMatchObject({
+      state: "covered",
+      unresolvedBlockerCount: 1,
+      coveredBlockerCount: 1,
+      stalledBlockerCount: 0,
+      attentionBlockerCount: 0,
+      sampleStalledBlockerIdentifier: null,
+    });
+  });
+
+  it("still flags an in_review blocker whose monitor has elapsed or was never scheduled as stalled", async () => {
+    const { companyId, agentId } = await createCompany("PBMS");
+    const elapsedParentId = await insertIssue({ companyId, identifier: "PBMS-1", title: "Elapsed parent", status: "blocked" });
+    const elapsedLeafId = await insertIssue({
+      companyId,
+      identifier: "PBMS-2",
+      title: "Review leaf whose monitor is already due",
+      status: "in_review",
+      assigneeAgentId: agentId,
+      monitorNextCheckAt: new Date(Date.now() - 60 * 60 * 1000),
+      executionState: monitorState({ timeoutAt: new Date(Date.now() + 24 * 60 * 60 * 1000) }),
+    });
+    await block({ companyId, blockerIssueId: elapsedLeafId, blockedIssueId: elapsedParentId });
+
+    const unmonitoredParentId = await insertIssue({ companyId, identifier: "PBMS-3", title: "Unmonitored parent", status: "blocked" });
+    const unmonitoredLeafId = await insertIssue({
+      companyId,
+      identifier: "PBMS-4",
+      title: "Review leaf with no monitor at all",
+      status: "in_review",
+      assigneeAgentId: agentId,
+    });
+    await block({ companyId, blockerIssueId: unmonitoredLeafId, blockedIssueId: unmonitoredParentId });
+
+    const blocked = await svc.list(companyId, { status: "blocked" });
+
+    expect(blocked.find((issue) => issue.id === elapsedParentId)?.blockerAttention).toMatchObject({
+      state: "stalled",
+      reason: "stalled_review",
+      coveredBlockerCount: 0,
+      stalledBlockerCount: 1,
+      sampleStalledBlockerIdentifier: "PBMS-2",
+    });
+    expect(blocked.find((issue) => issue.id === unmonitoredParentId)?.blockerAttention).toMatchObject({
+      state: "stalled",
+      reason: "stalled_review",
+      coveredBlockerCount: 0,
+      stalledBlockerCount: 1,
+      sampleStalledBlockerIdentifier: "PBMS-4",
+    });
+  });
+
   it("does not treat an elapsed or exhausted monitor as a covered waiting path", async () => {
     const { companyId, agentId } = await createCompany("PBME");
     const elapsedParentId = await insertIssue({ companyId, identifier: "PBME-1", title: "Elapsed parent", status: "blocked" });
--- a/server/src/__tests__/issue-blocker-attention.test.ts
+++ b/server/src/__tests__/issue-blocker-attention.test.ts
@@ -101,6 +101,8 @@ describeEmbeddedPostgres("issue blocker attention", () => {
     originFingerprint?: string | null;
     executionState?: Record<string, unknown> | null;
     description?: string | null;
+    monitorNextCheckAt?: Date | null;
+    monitorAttemptCount?: number | null;
   }) {
     const id = input.id ?? randomUUID();
     await db.insert(issues).values({
@@ -118,8 +120,23 @@ describeEmbeddedPostgres("issue blocker attention", () => {
       originFingerprint: input.originFingerprint ?? "default",
       executionState: input.executionState ?? null,
       description: input.description ?? null,
+      monitorNextCheckAt: input.monitorNextCheckAt ?? null,
+      ...(input.monitorAttemptCount == null ? {} : { monitorAttemptCount: input.monitorAttemptCount }),
     });
     return id;
+  }
+
+  function monitorState(input: { timeoutAt: Date; maxAttempts?: number; attemptCount?: number }) {
+    return {
+      monitor: {
+        kind: "external_service",
+        status: "scheduled",
+        serviceName: "github-checks",
+        timeoutAt: input.timeoutAt.toISOString(),
+        maxAttempts: input.maxAttempts ?? 12,
+        attemptCount: input.attemptCount ?? 0,
+      },
+    } satisfies Record<string, unknown>;
   }
 
   async function block(input: { companyId: string; blockerIssueId: string; blockedIssueId: string }) {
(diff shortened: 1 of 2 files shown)


juancarlosrial76-code:
## Follow-up: the `in_review` side of this same defect, now pinned

Pushed one commit (`89c0c35`) adding two test cases. No production code changed — the fix in this PR already covers the case; it just was not pinned.

**Why `in_review` needs its own pair.** `classifyPath` handles `in_review` in a dedicated branch that returns `stalled` (not `needs_attention`) when it finds no waiting path:

```js
if (node.status === "in_review") {
  const hasWaitingPath = activeIssueIds.has(node.id) || Boolean(node.assigneeUserId);
  if (hasWaitingPath) return { covered: true, stalled: false, ... };
  return { covered: false, stalled: true, sampleStalledBlockerIdentifier: nodeSample, ... };
}
```

Because this PR feeds the monitor into `explicitWaitingIssueIds`, which is checked *before* that branch, a monitored review already classifies as `covered`. But the existing cases here all exercise `in_progress`, so nothing stops a future edit from reordering those checks and silently restoring the `stalled` false positive. The new cases pin both directions:

- monitored `in_review` blocker → `covered`. **Fails on `master`** with `state: "stalled"`, `stalledBlockerCount: 1`, `sampleStalledBlockerIdentifier` pointing at the monitored blocker — so it is a real regression guard, not a test written to the new behaviour.
- elapsed monitor → still `stalled`; no monitor at all → still `stalled`. The boundary pair, so the fix cannot widen into "any monitor field covers".

**Field evidence for the `in_review` variant.** On a live board of 211 carriers across `in_review/in_progress/todo/backlog/blocked`: `none` 140, `covered` 63, `stalled` **2**, `needs_attention` 6 — and *both* `stalled` reports were this false positive. Each named blocker was `in_review` with a future `monitorNextCheckAt` (roughly 3.5 h and 27 d out), an unexpired `timeoutAt` and `attemptCount 0` well under `maxAttempts`; both had been worked within hours. Feeding those two live payloads straight into `hasScheduledIssueMonitorPath` returns `true` for both. So on that board `stalled` had a **0 % true-positive rate**, and the metric was penalising precisely the re-review path agents are told to use when there is no human reviewer.

## Rebase status

This branch is `CONFLICTING` against current `master`, and the conflict is a happy one: `master` has since landed #10675, which **already exports this predicate** as `hasScheduledIssueMonitorPath`. The rebase therefore *shrinks* this PR — the local `hasScheduledMonitor` export disappears and only a type widening remains:

```diff
-export function hasScheduledIssueMonitorPath(issue: IssueLivenessIssueInput, now: Date | string | number) {
+export function hasScheduledIssueMonitorPath(
+  issue: Pick<IssueLivenessIssueInput, "executionPolicy" | "executionState" | "monitorNextCheckAt" | "monitorAttemptCount">,
+  now: Date | string | number,
+) {
```

I have that rebase done and verified locally on `master` @ `6772423`:

```
vitest run server/src/__tests__/issue-blocker-attention.test.ts                  # 31 passed
vitest run issue-liveness + recovery-classifiers + attention-service.test.ts     # 47 passed
tsc -p server/tsconfig.json --noEmit                                            # clean in all touched files
```

I could not push it: our OAuth token lacks the `workflow` scope, and `master` has moved across `docker.yml`, `pr.yml` and `release-verify.yml` since this branch's base, so any rebase onto `master` is rejected with *"refusing to allow an OAuth App to create or update workflow `.github/workflows/docker.yml` without `workflow` scope"*. That is being fixed on our side; until then the branch stays on its old base and this commit is a plain fast-forward that touches no workflow file. **A maintainer with push access can rebase this in one command** — the only manual step is taking `master`'s side in `issue-graph-liveness.ts` and re-applying the `Pick<...>` widening above. Happy to push it the moment our scope is granted.

juancarlosrial76-code:
CI on the previous head `74613ae6` came back **26 success / 1 skipped / 2 failure**. Recording why I do not think either red is this diff, with the evidence rather than the assertion:

- `verify` is an aggregator — its log shows only `GENERAL_TESTS_RESULT: failure`, so it is a consequence, not a second finding.
- `General tests (server (2/5))` failed on `src/__tests__/plugin-worker-manager.test.ts > plugin worker manager setup-token pty route gate > terminalizes the route when the cumulative output passes the per-route bound`, `Test timed out in 15000ms`. Shard total: 1 failed / 80 passed.

That same test timed out ~30 minutes earlier on an unrelated branch, `fix/warn-on-adapter-cwd-fallback` (run `32802738047`, job `97666776991`), together with `binds the worker session id one time and ignores a duplicate open reply` from the same describe block. A neighbouring file, `plugin-worker-manager-duplex-byte-ledger.test.ts`, failed on a third branch in the same window. This diff touches `server/src/services/issues.ts` (blocker attention) and one test file — it does not reach the plugin worker manager or its pty routes.

I cannot re-run jobs on this repository, so I have force-pushed an **amend of the same commit** to trigger a fresh round. The tree hash is unchanged (`ae0a3281` before and after) — no content difference between `74613ae6` and `c17846c7`, only a new SHA.

If the pty gate suite is a known flake on your side, this is one more data point on it; if it is not, the three branches above are the reproduction window.

juancarlosrial76-code:
> **P1 — Cancelled approvals lose coverage** (`server/src/services/issues.ts:2510-2512`)

Checked at the code rather than accepting or rejecting it, and the premise is **correct**: `approvalsSvc.cancel` is only called explicitly (e.g. duplicate-agent cleanup) — cancelling an issue does not cancel an approval already linked to it, so the row stays `pending`. And the behaviour change is real, not hypothetical. I added a test for exactly this shape and ran it both ways:

| tree | `blocked` parent ← `cancelled` blocker with a `pending` approval |
|---|---|
| `master` | `covered` |
| this branch | `needs_attention`, `attentionBlockerCount: 1` |

So thank you — that is a genuine consequence of putting the guard on the shared `explicitWaitCandidateIds` rather than only on the monitor lookup. It is now pinned by `does not treat a cancelled blocker's pending approval as a covered waiting path` (commit `ecca2545`), so whichever way it is decided, it is decided explicitly.

**Where I disagree is the verdict, not the observation.** I think `needs_attention` is the right answer here:

1. Resolving the approval cannot move the blocker to `done`. The parent's `blockedBy` edge survives either outcome, so the parent really does need a human — to clear a dead edge. That is what this map exists to surface.
2. `master` already makes this exact call one branch below: `if (node.assigneeUserId && node.status !== "cancelled")`. A **human assignee** — a stronger path than a pending approval — deliberately does not cover a cancelled node. Covering it via the approval lookup while refusing to cover it via the assignee is an inconsistency in the same function.
3. `master` also has an explicit `if (node.status === "cancelled") → { covered: false, terminalBlockerIssueId: node.id }` branch. Reaching it more often is the intent, not a regression.

**But you have a fair scope point**, so I want to name it rather than argue past it: this PR is about the *monitor* path. The `cancelled` guard exists because of the defect **you** found in the first review round (a cancelled blocker's retained monitor read as covered), and that defect is monitor-specific. Extending the guard to interactions and approvals is scope I added.

If the maintainers prefer the minimal read, the narrow variant is one line — keep `explicitWaitCandidateIds` as `status !== "done"` and filter `cancelled` inside the monitor loop only. Happy to push that on request; I have not done it unilaterally because it would leave the inconsistency in (2) in code this PR is already touching, and I would rather that be a maintainer's call than my merge-convenience.

Both the risk and the measurement are in the PR description under *Risks* → "Narrowing risk, in the other direction".

juancarlosrial76-code:
Small clarification so the re-run is not misread: `Greptile Review` on `ecca2545` is again **4/5**, and that is the expected outcome, not a second failure.

The commit added **only a test** (`does not treat a cancelled blocker's pending approval as a covered waiting path`) — `server/src/services/issues.ts` is byte-identical to `e11086f17`. Greptile scores the implementation, so the P1 above naturally carries forward; no new inline finding was posted this round.

That leaves the confidence gate genuinely open, and closing it is a maintainer decision rather than something I should resolve unilaterally:

- **Keep as-is** — the shared `cancelled` guard, on the reasoning in my reply above (consistency with the `assigneeUserId && status !== "cancelled"` branch that `master` already has).
- **Narrow it** — one line, `explicitWaitCandidateIds` back to `status !== "done"` with the `cancelled` filter moved inside the monitor loop. Scope then matches the PR title exactly and the P1 disappears.

I will push the narrow variant the moment anyone says so. Everything else on this head is green: **27 success / 1 skipped** on `e11086f17` with the identical implementation, including Typecheck, Build, all five `General tests` server shards, all five `Verify serialized` suites and all three `e2e` shards. The `plugin-worker-manager` timeout from two rounds ago did not recur, which I take as confirming it was the flake I described rather than anything in this diff.

juancarlosrial76-code:
Status note so this PR is not misread as abandoned: it is **`CONFLICTING`** against current `master` and I cannot currently push the fix.

**The rebase itself is done and verified — it just cannot be delivered.** Locally, on top of `f2c5e54dc` (565 commits of drift since the last base):

```
vitest run src/__tests__/issue-blocker-attention.test.ts   # 33 passed (33)
```

Mutation probes, re-run after the rebase, all still discriminate:

| Mutation | Result |
|---|---|
| revert `services/issues.ts` to `master`, keep the tests | **4 red** |
| drop only the `cancelled` guard | **2 red** |
| replace `hasScheduledIssueMonitorPath(row, now)` with bare `nextCheckAt > now` | **1 red** |

The only merge conflict is formatting: `master` reformatted `server/src/services/issues.ts` wholesale, and my added block sits inside a region the formatter rewrote. The change itself is unchanged in substance.

**Why I cannot push it.** Our GitHub token carries `repo` but not `workflow`. A rebase onto current `master` necessarily brings `master`'s own changes to 18 files under `.github/workflows/`, and the push is rejected:

```
! [remote rejected] refusing to allow an OAuth App to create or update workflow
  `.github/workflows/cloud-artifacts.yml` without `workflow` scope
```

Four delivery paths measured, all closed:

1. force-push the rebased branch → rejected as above
2. push to a fresh branch instead → **also** rejected (same gate)
3. `PUT /pulls/10398/update-branch` (server-side merge, GitHub's own credentials — not affected by our scope) → `422 merge conflict between base and head`
4. two commits that pre-match `master`'s formatter output in the conflicting neighbourhood, hoping to make (3) auto-resolve → conflict persists, because the insertion sits inside a region **both** sides rewrote

`07197a14f` on this branch is the harmless residue of (4): pure formatting toward `master`'s style, no behaviour change.

So this waits on a credential change on our side, not on anything in the diff. I did not want to leave a red `CONFLICTING` PR sitting here without saying why. If a maintainer would rather I close this and re-open from a clean branch once the token is fixed, say the word — the branch name carries an internal ticket id anyway, which is the one checklist item I have never been able to tick.

The overlap disclosure with #8759 / #10502 / #11847 / #12035 still stands; all five PRs in this lane are still open and none has moved.

PR #10465, opened 2026-07-29
fix(ui): recover blockquotes the markdown editor escaped as \>

cryppadotta:
## Summary

Typing `>` to insert a blockquote (PAP-15769) can silently produce a **plain paragraph** instead of a blockquote in the markdown editor used by comments, descriptions, and documents.

Root cause: the MDXEditor markdown exporter escapes a leading `>` to `\>` whenever a `>`-prefixed line is a plain paragraph rather than a live blockquote node. That happens when the WYSIWYG blockquote shortcut does not fire — e.g. on browsers/IMEs where Lexical's markdown-shortcut input handling behaves differently (WebKit/Safari, soft keyboards), or after an edit that assembles the `> ` prefix without triggering the transform. The escaped `\> text` then renders as literal text on display, dropping the blockquote the user typed.

In Paperclip `>` at the start of a line always means "blockquote" — there is no literal-`>` affordance and the composer has no blockquote toolbar button — so an escaped `\>` is never the user's intent.

## Change

- New `ui/src/lib/blockquote-markdown.ts` — `unescapeBlockquoteMarkers()` rewrites a leading `\>` back to `>` (line-aware, leaves fenced code blocks untouched).
- Wired into `MarkdownEditor` on both **import** (`prepareMarkdownForEditor`) and **export** (`onChange`), so a `>`-prefixed line survives as a real blockquote regardless of whether the live shortcut fired, and legacy content already stored as `\>` renders correctly when re-opened.
- Unit tests for the helper (9 cases incl. multi-line, indentation, fenced-code protection, mid-line markers).

Strictly additive for the common path: content without `\>` is a no-op (early `includes` guard), so the normal `> text` → blockquote flow is unchanged.

## Testing

- `vitest` — `blockquote-markdown.test.ts` (9) + `MarkdownEditor.test.tsx` (37) pass; `tsc -b` clean.
- Isolated browser harness (real `MarkdownEditor` + the real `IssueChatThread` composer, driven with Playwright): verified `> text` still produces a `<blockquote>` and exports `> text` across empty/second-block/prefix-edit sequences, and confirmed the escaped `\>see` case now exports `>see` instead of `\>see`.

## Verification note

I could not reproduce the *exact* reported failure in headless Chromium — the WYSIWYG shortcut fired reliably there. The escape-based failure mode above is the plausible root cause on other browsers (Safari especially). Requesting QA verify blockquote insertion in a task-page comment on Safari/the live app.

### DIFF
--- a/ui/src/components/MarkdownEditor.tsx
+++ b/ui/src/components/MarkdownEditor.tsx
@@ -48,6 +48,7 @@ import { MentionAwareLinkNode, mentionAwareLinkNodeReplacement } from "../lib/me
 import { mentionDeletionPlugin } from "../lib/mention-deletion";
 import { looksLikeMarkdownPaste } from "../lib/markdownPaste";
 import { normalizeMarkdown } from "../lib/normalize-markdown";
+import { unescapeBlockquoteMarkers } from "../lib/blockquote-markdown";
 import { pasteNormalizationPlugin } from "../lib/paste-normalization";
 import { cn } from "../lib/utils";
 import { useEditorAutocomplete, type SlashCommandOption } from "../context/EditorAutocompleteContext";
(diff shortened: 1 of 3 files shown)


cryppadotta:
Superseded by #10466 (clean branch name + full PR template + Greptile fence-handling fix).

Issue #10374, opened 2026-07-28
Issue monitors are destroyed by a single transient dispatch skip; deferred wakes are booked as `issue.monitor_triggered`


greegorij:
## Summary

Two defects in the issue-monitor dispatch path (`server/src/services/heartbeat.ts`) make scheduled monitors unreliable in opposite directions:

- **A.** A single *transient* dispatch skip permanently destroys the monitor. No backoff, no retry, no attempt accounting, no `timeoutAt` fallback, no recovery owner.
- **B.** A *deferred* wake (`enqueueWakeup` returning `null`) is recorded as a successful `issue.monitor_triggered`: the monitor is consumed, no wake exists, and the activity log looks like success.

Both were verified by reading the source and reproduced on a live self-hosted instance (adapter `claude_local`). Line numbers below are from a tree synced with upstream `main`.

---

## A. `dispatch_skipped` clears the monitor permanently

The automated sweep `tickDueIssueMonitors` calls `dispatchClaimedIssueMonitor` with `clearOnClientError: true` (`heartbeat.ts:7077-7088`). Any 4xx from `enqueueWakeup` then lands here (`heartbeat.ts:6886-6900`):

```ts
    } catch (err) {
      if (err instanceof HttpError && err.status >= 400 && err.status < 500) {
        if (input.clearOnClientError) {
          await db
            .update(issues)
            .set({
              ...buildIssueMonitorClearedPatch({
                issue: claimed,
                policy,
                clearReason: "dispatch_skipped",
                clearedAt: input.now,
              }),
```

The branch logs `issue.monitor_skipped` (`:6908`) and returns `{ outcome: "skipped" }` (`:6921`). **There is no reschedule or backoff path.**

`buildIssueMonitorClearedPatch` (`server/src/services/issue-execution-policy.ts:1060-1065`):

```ts
  return {
    executionPolicy: stripMonitorFromExecutionPolicy(input.policy) as Record<string, unknown> | null,
    executionState: executionStateWithMonitor(existingState, nextMonitorState) as Record<string, unknown> | null,
    monitorNextCheckAt: null,
    monitorWakeRequestedAt: null,
  };
```

Note it also *strips the monitor policy itself*, so `timeoutAt` / `maxAttempts` are discarded along with the schedule.

Why this is unrecoverable:

[...]

Issue #3458, opened 2026-04-12
POST /issues: unknown fields silently stripped — assigneeId ignored instead of rejected or aliased


Daradaal:
## Summary

When creating an issue via `POST /api/companies/:companyId/issues`, unknown fields are silently stripped by Zod (default behavior). LLM agents frequently use `assigneeId` instead of the correct `assigneeAgentId`, causing the issue to be created without an assignee. No wakeup request is triggered for the intended agent, and the agent never wakes up.

## Reproduction

```bash
curl -X POST .../issues -d '{ "title": "test", "assigneeId": "<uuid>" }'
# Returns 200 with the issue created — but assigneeAgentId is null
```

## Impact

LLM CEO/CTO agents consistently use `assigneeId` in curl calls. The silent field strip means delegation silently fails — the sub-issue is created, the agent thinks it delegated, but the assignee is never woken up. No error is returned to indicate the mistake.

## Possible fixes

1. **Alias**: accept `assigneeId` as an alias for `assigneeAgentId` (most forgiving — LLMs will keep making this mistake regardless of instructions)
2. **Strict validation**: add `.strict()` to `createIssueSchema` to reject unknown fields with a clear error message
3. **Warning**: keep stripping but include a `warnings` array in the response when known field aliases are detected

## Local workaround

We patched `server/src/middleware/validate.ts` to pre-process `assigneeId` → `assigneeAgentId` before schema parsing.

**Affects**: gemini-local agents, any LLM-based client using the REST API

ricoproductsetsy:
Daily system health check: this open ticket had no update for more than 48 hours as of 2026-04-18. Leaving a status touchpoint so it stays visible for triage. No new fix verified from this check.

PR #10475, opened 2026-07-30
fix(api): reject unknown issue patch fields

goldinbidsheets-rgb:
## Thinking Path

> - Paperclip is the control plane people use to coordinate AI agents and their work
> - Issue dependencies are first-class relations that drive blocked-state visibility and automatic wakeups
> - The issue PATCH validator currently strips unknown keys, so a response-only relation key can return success without changing the dependency graph
> - That silent-success behavior makes malformed blocker writes indistinguishable from a persistence defect
> - Existing-child blocker updates also need an explicit regression so hierarchy cannot accidentally change relation persistence
> - This pull request makes issue PATCH fail closed, locks in child-update persistence, and documents the asymmetric write/read contract
> - The benefit is that callers either persist the requested blocker edge or receive an actionable 400 response

Bug report:

- **What happened:** `PATCH /api/issues/{issueId}` accepted unknown or response-only fields such as `blockedBy`, returned 200, and silently discarded them.
- **Expected behavior:** callers must use `blockedByIssueIds`; unsupported PATCH keys should return 400, and correct blocker updates must persist when the target issue is itself a child.
- **Impact:** a caller can believe it created a dependency while the issue remains blocked without a first-class blocker edge.

## What Changed

- Made the shared issue-update schema and HTTP route reject unknown PATCH keys.
- Added an HTTP regression proving `blockedBy` returns `400 Validation error` before the mutation service runs.
- Added an embedded-database regression proving `blockedByIssueIds` persists bidirectionally on an existing child issue without changing its parent.
- Documented `blockedByIssueIds` as the write field and `blockedBy` / `blocks` as read-only response fields.

## Verification

- `vitest run packages/shared/src/validators/issue.test.ts` - 28 passed.
- Focused issue-route regression - 1 passed (71 unrelated tests skipped by filter).
- `tsc -p packages/shared/tsconfig.json --noEmit` - passed.
- `tsc -p server/tsconfig.json --noEmit` - passed.
- The embedded-Postgres regression is included for Linux CI; this Windows host skipped that suite because its native embedded-Postgres data bundle is unavailable.

## Risks

- **Intentional compatibility change:** clients that relied on unsupported fields being silently ignored will now receive HTTP 400 and must correct their payload.
- Correctly shaped PATCH requests are unchanged. There is no database migration and no change to blocker persistence logic.

## Model Used

- OpenAI Codex, `gpt-5.6-sol`, max reasoning, with tool use and code execution.

## Checklist

### DIFF
--- a/docs/api/issues.md
+++ b/docs/api/issues.md
@@ -34,6 +34,8 @@ The response also includes:
 - `planDocument`: the full text of the issue document with key `plan`, when present
 - `documentSummaries`: metadata for all linked issue documents
 - `legacyPlanDocument`: a read-only fallback when the description still contains an old `<plan>` block
+- `blockedBy`: summaries of issues that block this issue
+- `blocks`: summaries of issues this issue blocks
 
 ## Create Issue
 
@@ -64,7 +66,9 @@ Headers: X-Paperclip-Run-Id: {runId}
 
 The optional `comment` field adds a comment in the same call.
 
-Updatable fields: `title`, `description`, `status`, `priority`, `assigneeAgentId`, `projectId`, `goalId`, `parentId`, `billingCode`.
+Updatable fields include `title`, `description`, `status`, `priority`, `assigneeAgentId`, `projectId`, `goalId`, `parentId`, `billingCode`, and `blockedByIssueIds`.
+
+`blockedByIssueIds` replaces the issue's blocker set. Relation summaries are read back as `blockedBy` and `blocks`; those response fields are read-only and are not accepted as PATCH keys. Unknown or read-only PATCH fields return `400 Validation error` instead of being silently ignored.
 
 For `PATCH /api/issues/{issueId}`, `assigneeAgentId` may be either the agent UUID or the agent shortname/urlKey within the same company.
 
--- a/packages/shared/src/validators/issue.test.ts
+++ b/packages/shared/src/validators/issue.test.ts
@@ -105,7 +105,7 @@ describe("issue validators", () => {
       createdByUserId: "spoofed-creator",
       responsibleUserId: "spoofed-responsible",
     });
-    const updated = updateIssueSchema.parse({
+    const updated = updateIssueSchema.safeParse({
       title: "Do not update attribution",
       createdByUserId: "spoofed-creator",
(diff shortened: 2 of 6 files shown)


antiautomation:
+1 — an independent field report of the defect this PR fixes, plus a candidate link for the blocking `commitperclip` check.

`updateIssueSchema` (`packages/shared/src/validators/issue.ts` L535) is built from the non-strict `createIssueBaseSchema` via `.omit(...).partial().extend(...)`, all of which inherit Zod's default strip mode. Unknown top-level keys are removed during `safeParse`, so the route never sees them and returns 200. A partially-applied PATCH is byte-for-byte indistinguishable from a fully-applied one, which in practice forces a re-GET after every write.

Where it cost us real time: while trying to clear a stuck issue monitor, `PATCH {"monitorNotes": "..."}` and `PATCH {"executionState": {...}}` both returned 200 and persisted nothing. Those two columns are engine-owned and were never client-writable — which is exactly the point. A 400 would have said so on the first call, instead of the defect being independently re-confirmed three times as a suspected persistence bug. (Details of that monitor case are in #10374.)

For the linked-issue check: **#3458** ("POST /issues: unknown fields silently stripped — assigneeId ignored...") is open and describes the same class of failure on the create path; `Refs #3458` in the PR body should satisfy it.

Worth noting that #10478 (merged) mitigates but does not close this. The `changes` receipt reports what committed, but a silently-dropped key and a genuine no-op both simply appear absent from it, so a caller still cannot distinguish the two. Rejecting the write is the stronger contract.

It's also already the idiomatic choice in this file, which is what makes the current behaviour surprising: the sub-objects *inside* `createIssueBaseSchema` — `unblockDescriptor`, `watchdogDiscovery`, `watchdog` — are each `.strict()`, and the sibling `upsertIssueWatchdogSchema` (L504) is `.strict()`. Only the top-level issue create/update body is not. So a typo nested one level deep is already a 400 today; the same typo at the top level is a silent 200.

goldinbidsheets-rgb:
Superseded by #11140. The replacement keeps the strict issue update contract and also rejects inline routine triggers. It is based on current master and includes focused tests and API documentation.

PR #10112, opened 2026-07-23
Route blocked transitions to explicit unblock owners

cryppadotta:
## Thinking Path

> - Paperclip is the open source control plane people use to coordinate AI-agent companies.
> - Issue status transitions determine whether work keeps moving or silently stalls.
> - A blocked issue previously could rely on prose alone, leaving the intended unblock owner unstructured and unnotified.
> - Existing blocker-attention classification could identify stalled chains, but the signal was not delivered to the board attention feed.
> - Blocked transitions also need rollout-safe deduplication so upgrades do not notify for historical issues and repeated processing does not create notification storms.
> - This pull request adds structured unblock descriptors, prospective transition timestamps, owner delivery, and board attention routing with focused authorization controls.
> - The benefit is that newly blocked work has an explicit, routable next action without weakening company boundaries or allowing agents to inject arbitrary human attention items.

## Linked Issues or Issue Description

Related documentation PR: #10094.

### Subsystem affected

Cross-cutting: `server/`, `packages/db`, and `packages/shared`.

### Problem or motivation

An issue can enter `blocked` without a machine-readable unblock path. Prose-only ownership does not reliably wake the responsible agent or surface human-owned work, while the existing `blockerAttention` classifier is not delivered to an operator-facing attention feed.

### Proposed solution

Require new transitions into `blocked` to have unresolved blockers, a pending interaction/approval, or a structured `{ owner, action }` descriptor. Notify an allowed owner once per prospective transition, route human-owned cases to board attention, and leave pre-rollout blocked issues untouched.

### Alternatives considered

- Keep prose-only blockers: rejected because ownership remains unroutable.
- Backfill all historical blocked issues: rejected because upgrades would create notification storms.
- Let agents target arbitrary users or the board: rejected after security review because it creates an attention-injection channel.

### Roadmap alignment

Aligns with `ROADMAP.md` → “Enforced Outcomes (watchdogs, recovery actions, review gates)” by making blocked work carry an explicit continuation path.

### Additional context

The implementation is prospective-only and deduplicated per blocked transition. Agent-authored descriptors are limited to the acting agent; board actors retain human-owner routing.

## What Changed

- Added persisted unblock descriptors and prospective blocked-transition delivery timestamps with an idempotent migration.
- Added shared types and validation for board, user, and agent unblock owners.
- Enforced valid blocked transitions and same-company owner validation in the issue update route.
- Restricted agent-authored descriptors to the acting agent itself, preventing board/user attention injection by compromised agents.
[...]

PR #10480, opened 2026-07-30
fix(recovery): never demote to blocked without a routable unblock owner

projecttop10tech-arch:
## Thinking Path

> - The recovery service watches for runs that die before the agent writes a disposition, and it demotes the stranded issue so a person or another agent can pick it up
> - The demotion writes `blocked` through `issuesSvc.update`, which does not apply the invariant that the issue PATCH route applies to every other actor
> - The route requires unresolved blockers, a pending interaction or approval, or an `unblockDescriptor` before an issue may enter `blocked`
> - A recovery demotion routinely satisfies none of the three, so it writes a `blocked` row that no automated path can reach
> - This pull request makes every recovery demotion attach a descriptor that names an invokable owner, or the board
> - The benefit is that a short adapter outage no longer converts each in-flight issue into a permanently stuck row

Related open work on the same surface: #10418 (wake unblock owners for initially blocked issues). The invariant this PR aligns recovery with was added in #10112 (route blocked transitions to explicit unblock owners).

No public issue exists for this defect, so it is described here as a bug report.

### Bug report

**What happened**

Terminal-run recovery demotes a stranded issue to `blocked` with whatever unresolved blocker edges the issue happens to have. That is routinely none. The resulting row is unreachable:

- `classifyIssueGraphLiveness` emits findings only for the blocked-by-unassigned, assigned-backlog, uninvokable-assignee, cancelled-blocker and review shapes. A blocked row with no blocker edge matches none of them.
- With no blocker edge, the row can never receive `issue_blockers_resolved`.
- `reconcileStrandedAssignedIssues` skips issues that are already `blocked`, so the sweep never revisits the row.

So the single `source_scoped_recovery_action` wake is the only path to the issue. If that wake fires into an adapter that is still down, the issue is stuck forever. The escalation comment says *"moving it to `blocked` so it is visible for intervention"*. That is true for the human attention feed. It is false for every automated path.

**What I expected**

`blocked` written by recovery should satisfy the same invariant that `server/src/routes/issues.ts` enforces for every other actor: unresolved blockers, a pending interaction or approval, or an `unblockDescriptor`.

**Steps to reproduce**

1. Assign an issue to an agent and start a run.
2. Make the run fail terminally before the agent writes a disposition. A dependency or adapter outage does this.
3. Let `reconcileStrandedAssignedIssues` demote the issue.
4. Read the issue. Status is `blocked`, hydrated blockers are `[]`, and `unblockDescriptor` is null.

**Impact on a live instance**

A 25-minute dependency outage left 35 issues in this shape rather than costing 25 minutes. A sweep 48 minutes after the outage ended demoted a fresh batch of six the same way. Every one had `[]` for hydrated blockers.

**Version**

Reproduced against `master` at a6436126c.

## What Changed

- Added `resolveRecoveryUnblockDescriptor`. It returns the descriptor a recovery demotion must attach, and returns null when the row already carries its own justification, so no redundant descriptor is written.
- Added `hasPendingBlockedJustification` to detect a pending interaction or approval.
- Added `resolveInvokableDescriptorOwnerAgentId`. It walks the candidate owners in order and returns the first one that is invokable. `returnOwnerAgentId` is the pre-recovery assignee, which is exactly the agent that can be paused or archived, so it is a candidate and not a guaranteed owner. When no candidate is invokable the descriptor owner becomes `board`.
- Added `applyRecoveryBlockedUpdate`. It applies the demotion, then re-reads the justification. The justification read and the status write are separate statements, so a blocker that completes, or an interaction that is answered, in between would otherwise leave `blocked` standing on nothing. The re-read costs one query and runs only on the path that skipped the descriptor.
- Routed all three demotion sites through that helper: the initial write and the re-block in `escalateStrandedAssignedIssue`, and `escalateStrandedRecoveryIssueInPlace`.
- `escalateStrandedRecoveryIssueInPlace` uses a board-owned descriptor. That path exists because Paperclip has stopped automatic recovery for the issue, so the intervention surface must be a human one.
- `resolveContinuationWaitingOnReview` is untouched. It already returns early when it finds nothing to wait on.

No extra wake is enqueued. `stranded_assigned_issue` already wakes the recovery owner through `enqueueSourceScopedStrandedRecoveryWake`. The paths that skip that wake — `configuration_incomplete`, provider-quota waits, and `escalateStrandedRecoveryIssueInPlace` — skip it on purpose, because a re-dispatch only reproduces the same opaque setup failure. For those paths the descriptor is the intervention surface, not a retry trigger.

## Verification

Run against `master` at a6436126c:

| Check | Command | Result |
|---|---|---|
| Types | `pnpm --filter @paperclipai/server typecheck` | clean |
| Recovery sweep | `npx vitest run src/__tests__/heartbeat-process-recovery.test.ts` | 93 passed (91 baseline + 2 new) |
| Recovery actions | `npx vitest run src/__tests__/issue-recovery-actions.test.ts` | 43 passed (41 baseline + 2 new) |

Four new tests:

1. A blocker-less issue must leave terminal-run recovery with a descriptor that names a real owner and a non-empty action.
2. An issue that already has a live blocker edge must not gain a redundant descriptor.
3. A demotion must name the invokable recovery owner on the descriptor.
4. A demotion must fall back to a board-owned descriptor when no candidate owner is invokable.

Test 1 fails on the parent commit with `AssertionError: expected null to be truthy`. It reproduces the defect rather than describing current behaviour. The baseline on the unmodified worktree is 91 passed and 41 passed, so no pre-existing failure masks anything.

## Risks

Low risk, with one behavioural change worth stating plainly.

- Every recovery demotion now writes an `unblockDescriptor` when the row has no other justification. Anything downstream that reads "has a descriptor" as "a human deliberately blocked this" will now see more descriptors. The useful discriminator is the owner: a descriptor whose owner is the recovery owner agent or `board` was written by recovery.
- The added work is one invokability read per candidate owner and, on the skipped-descriptor path only, one extra blocker read. Both are already-hot queries in this service.
- No schema change. No migration. No change to any wake or dispatch path.

## Model Used

Claude Opus 5 (`claude-opus-5`), 1M context window, extended thinking, with tool use and code execution in Claude Code.

## Checklist

> Branch-name box left unchecked: the branch carries an instance-local ticket id and was pushed before I read that rule. Renaming it now would close this PR and lose the review history. Say the word and I will re-open on a `fix/recovery-blocked-invariant` branch.

### DIFF
--- a/server/src/__tests__/heartbeat-process-recovery.test.ts
+++ b/server/src/__tests__/heartbeat-process-recovery.test.ts
@@ -5550,7 +5550,7 @@ describeEmbeddedPostgres("heartbeat orphaned process recovery", () => {
     }
   });
 
-  // PLA-3727 / PLA-3680 gate 2: terminal-run recovery used to demote issues to `blocked`
+  // Terminal-run recovery used to demote issues to `blocked`
   // with zero blocker edges and no unblock descriptor. Nothing can reach that shape — the
   // issue-graph liveness classifier only emits findings for the blocked-by-unassigned,
   // assigned-backlog, uninvokable-assignee, cancelled-blocker and review shapes — so the
--- a/server/src/__tests__/heartbeat-process-recovery.test.ts
+++ b/server/src/__tests__/heartbeat-process-recovery.test.ts
@@ -5550,6 +5550,107 @@ describeEmbeddedPostgres("heartbeat orphaned process recovery", () => {
     }
   });
 
+  // PLA-3727 / PLA-3680 gate 2: terminal-run recovery used to demote issues to `blocked`
+  // with zero blocker edges and no unblock descriptor. Nothing can reach that shape — the
+  // issue-graph liveness classifier only emits findings for the blocked-by-unassigned,
+  // assigned-backlog, uninvokable-assignee, cancelled-blocker and review shapes — so the
+  // escalation comment's "moving it to `blocked` so it is visible for intervention" was
+  // false. A 25-minute dependency outage stranded 35 issues for a full day because of it.
+  it("never demotes a blocker-less issue to blocked without a wake-reachable unblock descriptor", async () => {
+    const { companyId, agentId, issueId, runId } = await seedStrandedIssueFixture({
+      status: "in_progress",
+      runStatus: "failed",
+      runErrorCode: "budget_blocked",
+      runError: "Budget exceeded; refusing to dispatch.",
+    });
+    const heartbeat = heartbeatService(db);
+
+    // Precondition: the issue has no blocker edges at all, which is the exact shape that
+    // produced the dead strands.
+    const blockerEdges = await db
+      .select({ id: issueRelations.id })
+      .from(issueRelations)
+      .where(and(
+        eq(issueRelations.companyId, companyId),
+        eq(issueRelations.relatedIssueId, issueId),
+        eq(issueRelations.type, "blocks"),
+      ));
+    expect(blockerEdges).toHaveLength(0);
+
+    const result = await heartbeat.reconcileStrandedAssignedIssues();
+    expect(result.escalated).toBe(1);
+
+    const issue = await db.select().from(issues).where(eq(issues.id, issueId)).then((rows) => rows[0] ?? null);
+    expect(issue?.status).toBe("blocked");
+
+    // The invariant the PATCH route already enforces for every other actor:
+    // "Entering blocked requires unresolved blockers, a pending interaction/approval, or
+    // unblockDescriptor". With no blocker edges, the descriptor is what makes it reachable.
+    const descriptor = issue?.unblockDescriptor as
+      | { owner: "board" | { agentId: string } | { userId: string }; action: string }
+      | null;
+    expect(descriptor).toBeTruthy();
+    expect(typeof descriptor?.action).toBe("string");
+    expect((descriptor?.action ?? "").trim().length).toBeGreaterThan(0);
+    expect(issue?.blockedTransitionAt).toBeTruthy();
+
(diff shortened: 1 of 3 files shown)


projecttop10tech-arch:
Both P1s addressed in `074bea28` (plus `a8dbe7d1` scrubbing instance-local ticket ids from the code comments).

**P1 — uninvokable owner remains routed.** Agreed, and the suggested `ownerAgentId: recoveryAction.ownerAgentId` would fix the reported case but drop a routable owner in another: on the provider-quota wait path `ownerAgentId` is null by construction while `returnOwnerAgentId` is set and invokable, and a monitor is already armed for that agent. So the candidates are now filtered rather than reordered — `resolveInvokableDescriptorOwnerAgentId` walks `ownerAgentId`, `returnOwnerAgentId`, `assigneeAgentId` in order and returns the first that resolves to an invokable agent in the same company. When none is, the descriptor owner is `board`. New test: *falls back to a board-owned unblock descriptor when no candidate owner is invokable* (agents paused, asserts `owner === "board"`).

**P1 — justification check races resolution.** Correct, and it applied to blocker edges as well as interactions and approvals. All three demotion sites now go through `applyRecoveryBlockedUpdate`, which performs the write and then re-reads the justification; if it has resolved in the window and no descriptor was attached, it attaches one. The re-read is one query and runs only on the branch that skipped the descriptor, so the common path is unchanged.

Verified at `master` a6436126c: `typecheck` clean, `heartbeat-process-recovery.test.ts` 93/93, `issue-recovery-actions.test.ts` 43/43.

The `review` gate failure was a PR-description format miss on my side, not a code issue — the body now follows the template.

PR #10501, opened 2026-07-30
fix(heartbeat): exclude error-status agents from recovery re-invocation

### LINKED ISSUE #10503: Agent in status=error gets immediately re-invoked by recovery, causing retry bursts with no cooldown
### Pre-submission checklist

### What happened?
An agent whose adapter fails in a way that leaves the agent record in `status=error` (for example, an external CLI-based adapter hitting a provider-side rate limit and exiting non-zero) gets immediately re-invoked by the automatic recovery path (`releaseIssueExecutionAndPromote` in `server/src/services/heartbeat.ts`) the next time the issue it's assigned to receives any external activity — a new comment, a mention, a wake. The `recoveryAgentInvokable` check there only excludes agents with status `paused`, `terminated`, or `pending_approval`; `error` is not excluded, so the agent is treated as available. Because the agent is still failing (e.g. still rate-limited), the new run fails again within seconds, and this repeats on every subsequent issue event with no cooldown — producing tight retry bursts (multiple runs within the same minute) over hours or days.

### Expected behavior
An agent in `status=error` should not be silently treated as available for automatic re-invocation by this recovery path, consistent with how `paused`/`terminated`/`pending_approval` agents already aren't. Recovery from `error` should require an explicit action (a status change, a board/human unblock) rather than happening implicitly on the next unrelated issue event.

### Steps to reproduce
1. Have an agent whose adapter fails in a way that sets `status=error` on the agent record (any adapter can trigger this — a rate-limited external CLI is one example, but the bug is adapter-agnostic).
2. Post a comment or mention on an issue assigned to that agent (or otherwise trigger a wake) while the agent is still in `status=error`.
3. Observe that `releaseIssueExecutionAndPromote` queues a new run for the still-`error` agent immediately, with no cooldown, because `recoveryAgentInvokable` only excludes `paused`/`terminated`/`pending_approval`.
4. Repeat step 2 a few times in quick succession (or let normal issue traffic do it) and observe multiple back-to-back failed runs for the same errored agent.

`master` (heartbeat.ts recovery path, `releaseIssueExecutionAndPromote` / `recoveryAgentInvokable`)

### Deployment mode
Self-hosted server

### Agent adapter(s) involved
Not adapter-specific (core bug) — any adapter that can leave an agent in `status=error` is affected.

### Database mode
External Postgres

### Additional context
Filed alongside a fix: https://github.com/paperclipai/paperclip/pull/10501

### Privacy checklist

PAT-Main:
## Thinking Path

> - The heartbeat/recovery subsystem is responsible for deciding when a stuck or failed agent run should be automatically re-invoked
> - `recoveryAgentInvokable` in `releaseIssueExecutionAndPromote` treats any agent as invokable unless its status is `paused`, `terminated`, or `pending_approval` — `error` is not in that excluded set
> - This means an agent whose last run failed and left it in `status=error` (e.g. an external CLI adapter hitting a provider-side rate limit or crashing) is still considered invokable, so the very next external event touching the issue (a comment, a mention, a wake) immediately re-queues a new run for that same broken agent
> - Because there is no cooldown between these re-invocations, this produces tight retry bursts (multiple runs within seconds) every time the issue gets external activity, which is exactly the wrong behavior for an agent that is already known to be failing
> - This PR adds `status !== "error"` to the exclusion list, so a failing agent is parked (like `paused`/`terminated`/`pending_approval` already are) until something explicitly changes its status again
> - The benefit is that a single misbehaving/rate-limited agent no longer gets hammered with repeated automatic re-invocations just because unrelated issue activity keeps happening around it

Fixes #10503

For reviewers who prefer the full bug-report fields inline (also filed as #10503 above):

**What happened?** An agent adapter (in our case, an external CLI-based adapter that hit the provider's own rate limit and exited non-zero) ended up in `status=error`. Every time the issue it was working on received external activity (a new comment, a mention), Paperclip's recovery path (`releaseIssueExecutionAndPromote` → `recoveryAgentInvokable`) treated the agent as invokable and queued a fresh run immediately. Since the agent was still rate-limited, the new run failed again within seconds, and the cycle repeated — several runs within the same minute, sometimes multiple times a day over almost a week, with each run also chewing through the exhausted provider quota bucket.

**Expected behavior:** An agent in `status=error` should not be silently treated as available for automatic re-invocation, the same way `paused`/`terminated`/`pending_approval` agents already aren't. Recovery of an errored agent should require an explicit action (a status change, a board/human unblock) rather than happening implicitly on the next unrelated issue event.

**Steps to reproduce:**
1. Have an agent whose adapter fails in a way that sets `status=error` on the agent record (any adapter can do this — ours was an external rate-limited CLI, but the bug is adapter-agnostic).
2. Post a comment or mention on an issue assigned to that agent (or otherwise trigger a wake).
3. Observe that `releaseIssueExecutionAndPromote` queues a new run for the still-`error` agent immediately, with no cooldown, because `recoveryAgentInvokable` only excludes `paused`/`terminated`/`pending_approval`.
4. Repeat step 2 a few times in quick succession (or let normal issue traffic do it) and observe multiple back-to-back failed runs.

**Paperclip version or commit:** `master` at the time of this PR (heartbeat.ts recovery path).

**Deployment mode:** Self-hosted server.

**Database mode:** External Postgres (production instance); not database-related.

**Agent adapter(s) involved:** Not adapter-specific (core bug) — any adapter that can end up in `status=error` is affected.

## What Changed

- `server/src/services/heartbeat.ts`: extracted the inline `recoveryAgentInvokable` condition inside `releaseIssueExecutionAndPromote` into a small, pure, exported helper (`isAgentInvokableForRecovery` / `RECOVERY_NON_INVOKABLE_AGENT_STATUSES`), and added `"error"` to the excluded set alongside the existing `paused`/`terminated`/`pending_approval` statuses.
- `server/src/__tests__/heartbeat-recovery-agent-invokable.test.ts`: new unit test file covering the extracted helper directly (no database needed) — active-ish statuses stay invokable, `paused`/`terminated`/`pending_approval` stay excluded (unchanged behavior), `error` is now excluded (the fix), and `null`/`undefined` status is treated as non-invokable.

## Verification

- Read through `releaseIssueExecutionAndPromote` and the existing tests in `server/src/__tests__/heartbeat-process-recovery.test.ts` to confirm this exclusion list is exactly what other terminal-ish statuses already use, and that this is the single call site controlling whether a recovery re-invocation happens.
- Extracted the condition into `isAgentInvokableForRecovery` specifically so it could be unit tested without a database, since the existing integration tests in `heartbeat-process-recovery.test.ts` require the embedded-Postgres test harness.
- Ran the new unit test file locally: `cd server && vitest run src/__tests__/heartbeat-recovery-agent-invokable.test.ts` — **5/5 passing**.
- Did **not** get a local green run of the pre-existing `heartbeat-process-recovery.test.ts` integration suite: the local sandbox this PR was authored in runs everything as `root`, and that suite's embedded-Postgres harness (`initdb`) refuses to start under `root` by design (confirmed by running `initdb` directly outside of vitest and getting `initdb: error: cannot be run as root`). This is a pre-existing local environment limitation unrelated to this change, not a failure caused by it. Relying on CI (which runs as a non-root user) to execute that suite against this change.

## Risks

- **Behavioral shift, by design:** agents in `status=error` will no longer be automatically re-invoked by this recovery path. Previously they were retried immediately on the next issue event; now they stay parked until something explicitly changes their status. This is the intended fix, but it does mean recovery of an errored agent becomes more "pull" (explicit unblock) than "push" (automatic retry) — worth confirming this matches the maintainers' intended semantics for `status=error` more broadly, since this is a widely shared code path across all adapters/companies.
- Low risk otherwise: single-line additive condition, no schema change, no new dependency.

## Model Used

Claude (Anthropic), model `claude-sonnet-5` ("Sonnet 5"), accessed via Claude Code. Standard (non-extended) reasoning mode, with tool use (code search/read, shell execution, git) used to trace the bug from a real-world symptom (retry bursts on a rate-limited agent) down to this exact code path, and to attempt local test verification.

## Checklist

- [x] Linked issue: Fixes #10503
- [x] I have run tests locally and they pass — new unit test (5/5) run locally; pre-existing integration suite blocked by an unrelated local root/embedded-Postgres environment limitation, see Verification
- [x] I have added or updated tests where applicable — new `heartbeat-recovery-agent-invokable.test.ts` unit test file added
- [ ] I have updated relevant documentation to reflect my changes — not applicable, internal recovery logic
- [ ] All Paperclip CI gates are green — pending, will check after push
- [ ] Greptile is 5/5 with no open P2s, recommendations, or follow-ups — pending review


david-d-projects:
Follow-up on the Greptile P1: the new helper/unit tests in `bcd98b1c9` cover the local recovery predicate, but the shared invokability contract still treats `error` as invokable (`packages/shared/src/agent-eligibility.ts`), and `DIRECT_NON_INVOKABLE_STATUSES` still omits it (`server/src/services/agent-invokability.ts`). That leaves other wake/start paths inconsistent with the parked-state behavior.

I validated the minimal alignment locally:
- add `error` to `AgentEligibilityLifecycleReason` and the shared non-invokable set; remove it from `INVOKABLE_AGENT_STATUSES`
- map `error` in `getAgentWorkEligibility`, `AgentInvokabilityBlockReason`, `statusBlockReason`, and `DIRECT_NON_INVOKABLE_STATUSES`
- add shared eligibility coverage plus an embedded-Postgres regression proving an errored agent’s failed orphan run blocks the issue and creates no retry run

Local checks passed: shared eligibility (8/8), focused heartbeat recovery (1/1), shared typecheck, and server typecheck. I cannot push directly to the contributor branch with the current GitHub credentials, so the branch owner will need to apply the contract-alignment changes.

santhiprakash:
Addressed: aligned the shared eligibility contract with the heartbeat recovery guard by treating  as non-invokable everywhere the agent-work contract is used.

Regression check:  now covers  alongside  and , and the heartbeat recovery helper still has its dedicated unit coverage.

Status: committed as  on . I attempted targeted Vitest runs, but this environment's pnpm version gate stopped execution before tests could start, so I could not get a clean local run here.

santhiprakash:
Addressed: aligned the shared eligibility contract with the heartbeat recovery guard by treating `error` as non-invokable everywhere the agent-work contract is used.

Regression check: `packages/shared/src/agent-eligibility.test.ts` now covers `error` alongside `terminated` and `pending_approval`, and the heartbeat recovery helper still has its dedicated unit coverage.

Status: committed as `51489c2e3` on `fix/heartbeat-recovery-skip-error-agents`. I attempted targeted Vitest runs, but this environment's pnpm version gate stopped execution before tests could start, so I could not get a clean local run here.

david-d-projects:
The PR head is still `bcd98b1c9`; commit `51489c2e3` is not present on `fix/heartbeat-recovery-skip-error-agents`, so CI and Greptile have not rerun. Please push that commit to the PR branch when able.

santhiprakash:
Addressed: the shared eligibility fix is present locally on commit 51489c2e3, but the open PR still tracks PAT-Main's fork head at bcd98b1c9, so GitHub has not picked up the new commit from this checkout.

Regression check: the local branch now includes the shared contract change in packages/shared/src/agent-eligibility.ts plus the matching test update in packages/shared/src/agent-eligibility.test.ts.

Status: blocked on the PR branch owner / source fork. I cannot publish 51489c2e3 into the live PR from the santhiprakash fork because the PR head belongs to PAT-Main. Once that branch is updated or the change is transferred to a santhiprakash-authored PR, the fix is ready to re-review.

johnnysunyi-bit:
Supporting evidence from a production fleet (7 `claude_local` agents): because `error` is not in `DIRECT_NON_INVOKABLE_STATUSES`, wakes to an errored agent are enqueued and die in the adapter with no signal on the board. We ended up running an external cron that does `clear-error` + `resume` for auth/session-class errors. Today it was the only thing that recovered an agent stuck on `401 OAuth access token has expired` while the native scheduler kept issuing wakes into it.

Would be great to have this in core. If helpful, the pattern that worked for us: classify `error_reason` (auth vs transient), skip auth-class errors (they need a human), and cap remediation attempts per hour with an alert when the cap is exceeded.

santhiprakash:
Confirmed on current `master` — this is still the hole.

- `INVOKABLE_AGENT_STATUSES` still includes `error` in `packages/shared/src/agent-eligibility.ts`.
- `DIRECT_NON_INVOKABLE_STATUSES` is still only `paused | terminated | pending_approval` in `server/src/services/agent-invokability.ts`.
- Execution-start uses that set as the sole gate (`notInArray(agents.status, DIRECT_NON_INVOKABLE_STATUSES)` in `heartbeat.ts`), so an agent stuck on `401 OAuth access token has expired` keeps getting wakes. Those runs die in the adapter and the board never sees them.

That matches the fleet failure you described.

This PR is the skip-wake path. The remaining hole from the earlier review is that the shared invokability contract still treats `error` as invokable, so other start/wake paths can bypass a recovery-only guard.

The classified remediator (auth vs transient, skip auth-class, cap retries per hour) is a separate follow-up. Auth-class errors need a human. Auto `clear-error` + `resume` on a dead OAuth token would hide the same failure.

Refs: https://github.com/paperclipai/paperclip/pull/10501#issuecomment-5462378058

santhiprakash:
Replying to your fleet evidence above.

The shared eligibility fix is now rebased onto current master as a clean, signed santhiprakash PR: #10547. It treats `error` as non-invokable in the shared agent-work contract, the server execution-start gate, and blocker-attention coverage, with regression tests for each.

I dropped the ACPX test-hardening commit from that branch because `master` already fixed the fake-timer race with a call-boundary tracker, so the 500-flush helper would have been a regression.

The auth-vs-transient `error_reason` remediation pattern you described is a useful follow-up and stays out of scope here.

santhiprakash:
Aligned: #10547 is now rebased onto current master (`6d358e646`) and re-signed as `5073b2ad`. CI is re-running. The shared invokability contract and execution-start gate both treat `error` as non-invokable.

PR #10512, opened 2026-07-30
Add opt-in in-flight run-log mirroring with graceful-shutdown flush

devinfoley:
## Thinking Path

> - The run-log store records each agent run's output and can mirror completed logs to S3-compatible object storage
> - The mirror uploads only on finalize, so a server restart mid-run loses the whole in-flight log
> - Deployments and crashes are routine on ephemeral hosts, and lost run output makes failed runs impossible to debug
> - This pull request adds an opt-in throttled mirror for still-running logs plus a graceful-shutdown flush
> - The benefit is that a restart mid-run keeps the log tail up to the last mirror interval, and an orderly restart keeps everything

No public issue exists — describing the feature inline (per the feature request template).

**Subsystem affected**
server/ — REST API & orchestration services

**Problem or motivation**
`RUN_LOG_S3_BUCKET` gives finished run logs durability, but the mirror uploads only on finalize. A run that is still writing when the server restarts leaves nothing in object storage. On hosts with ephemeral disks the local file is gone too, so the run's output is lost end to end and failed runs cannot be debugged.

**Proposed solution**
Mirror the in-flight log to the same object key on a throttled cadence (`RUN_LOG_S3_INFLIGHT_MIRROR_SECONDS`), and flush dirty tails during graceful shutdown. Keep it opt-in so existing deployments see zero new upload traffic unless they ask for it.

**Alternatives considered**
Per-append uploads (rejected: one PUT per output chunk is hostile to S3 endpoints and run latency). Chunked part objects with read-time stitching (rejected: complicates the read path, and S3 multipart minimum part sizes do not fit small tails). Persistent volumes (rejected upstream already: the data dir is deliberately an emptyDir in hardened cloud_tenant deployments).

**Roadmap alignment**
Not on ROADMAP.md; extends the existing run-log durability mirror without changing any default behavior.

**Additional context**
Ranged reads already serve partial objects like a live tail, so the read path needs no change; finalize overwrites the mirror with the complete file.

**Related PRs (dedup search):** the finalize-only S3 mirror landed previously and this extends it; no duplicate or competing PR found for in-flight run-log mirroring.

## What Changed

- `server/src/services/run-log-store.ts`: new opt-in `inflightMirrorMs` on the S3 options (`RUN_LOG_S3_INFLIGHT_MIRROR_SECONDS` env). When set, appends schedule at most one upload of the current file per interval, to the same key finalize uses. Ranged reads already serve that key, so a partial object behaves like a live tail and needs no read-path change. Finalize retires the in-flight bookkeeping and waits out an upload already on the wire, so a stale partial can never overwrite a finalized log. Upload failures warn, re-mark the tail dirty, and retry at most once per interval.
- `server/src/services/run-log-store.ts`: new `flushInflightMirrors()` on the store and a module-level `flushInFlightRunLogMirrors()` for the shutdown path. Both are no-ops when the mirror is off.
- `server/src/index.ts`: graceful shutdown flushes dirty in-flight tails after the heartbeat run drain, so runs the drain did not finalize (timeouts, the hot-restart skip path) still persist their output.
- `server/src/services/run-log-store.test.ts`: five new tests — off-by-default (no uploads before finalize), tail preserved after a wipe without finalize, throttle coalescing with a single flush upload, finalize superseding the in-flight mirror and retiring its timer, and upload failures never breaking appends with recovery on the next flush.

## Verification

- `pnpm vitest run server/src/services/run-log-store.test.ts` — 13 passed (8 existing + 5 new).
- `pnpm vitest run server/src/__tests__/heartbeat-run-log.test.ts server/src/__tests__/heartbeat-active-run-output-watchdog.test.ts` — 21 passed (consumers of the store, unchanged behavior).
- `pnpm -C server run typecheck` — clean.
- Self-hosted behavior is unchanged unless `RUN_LOG_S3_INFLIGHT_MIRROR_SECONDS` is set: with the variable unset there are zero new uploads and the finalize-only mirroring is byte-identical (asserted by the off-by-default test).

## Risks

- Low. The feature is opt-in; unset env preserves today's behavior exactly. When enabled, worst case is one extra PUT per interval per active run, and every upload is best-effort — a failing endpoint warns and never breaks appends, finalization, or shutdown. The finalize path awaits any in-flight upload before writing the complete file, closing the only overwrite race the design introduces. Timers are `unref`ed so the mirror never keeps the process alive.

## Model Used

Claude Fable 5 (`claude-fable-5`, Anthropic; Claude Code CLI with extended thinking and tool use; tests executed locally via Vitest).

## Checklist

- [x] I have searched GitHub for duplicate or related PRs and linked them above (no duplicates found for in-flight run-log mirroring; the finalize-only mirror landed previously and this extends it)

### DIFF
--- a/server/src/index.ts
+++ b/server/src/index.ts
@@ -72,6 +72,7 @@ import { maybePersistWorktreeRuntimePorts } from "./worktree-config.js";
 import { initTelemetry, getTelemetryClient } from "./telemetry.js";
 import { conflict } from "./errors.js";
 import { coordinateHeartbeatSchedulerShutdown } from "./shutdown.js";
+import { flushInFlightRunLogMirrors } from "./services/run-log-store.js";
 import type {
   InstanceDatabaseBackupRunResult,
   InstanceDatabaseBackupTrigger,
@@ -1323,6 +1324,16 @@ export async function startServer(): Promise<StartedServer> {
         } catch (err) {
           logger.error({ err, signal }, "graceful heartbeat run drain failed");
         }
+      }
+
+      // Whatever the drain did not finalize (timed-out runs, the hot-restart
+      // skip path) still has a local-only tail when the in-flight run-log
+      // mirror is enabled; upload those tails now so an orderly restart
+      // never loses run output. No-op when the mirror is off.
+      try {
+        await flushInFlightRunLogMirrors();
+      } catch (err) {
+        logger.error({ err, signal }, "run-log in-flight mirror flush failed");
       }
 
       const appShutdown = (app as { locals?: { paperclipShutdown?: () => void } }).locals?.paperclipShutdown;
--- a/server/src/services/run-log-store.test.ts
+++ b/server/src/services/run-log-store.test.ts
@@ -242,6 +242,70 @@ describe("in-flight mirror", () => {
     expect(objects.get(key)!.toString("utf8")).toContain("final-line");
   });
 
+  it("shutdown flush re-checks dirtiness after an upload already on the wire completes", async () => {
+    const { provider, objects, calls } = createMemoryProvider();
+    let releaseFirstPut: (() => void) | null = null;
+    const firstPutGate = new Promise<void>((resolve) => {
+      releaseFirstPut = resolve;
+    });
+    let putStarts = 0;
+    const realPut = provider.putObject.bind(provider);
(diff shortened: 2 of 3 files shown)


devinfoley:
@greptile review

PR #10517, opened 2026-07-30
fix(runtime-api): PAPERCLIP_RUNTIME_API_URL precedence + authPublicBaseUrl port preservation

mohr-and-more:
## Thinking Path

> - The server injects environment variables into agent run contexts
> - PAPERCLIP_RUNTIME_API_URL was derived from the listen port (:3001) and took precedence over the correct PAPERCLIP_API_URL, causing agents to receive unreachable API URLs
> - The authPublicBaseUrl port rewrite blindly overwrote the operator-configured port (e.g., :8443 for reverse proxy) with the actual listen port (:3001)
> - This pull request fixes the precedence order and preserves the operator-configured port
> - Agents will now reliably receive the correct, reachable API URL in their run environment

## What Changed

- server/src/index.ts: Changed PAPERCLIP_RUNTIME_API_URL to use configuredApiUrl instead of runtimeApiUrl (L698)
- packages/adapter-utils/src/execution-target.ts: Swapped precedence to check PAPERCLIP_API_URL before PAPERCLIP_RUNTIME_API_URL
- packages/adapter-utils/src/server-utils.ts: Swapped precedence to check PAPERCLIP_API_URL before PAPERCLIP_RUNTIME_API_URL
- server/src/index.ts L232: Only rewrite authPublicBaseUrl port when it matches the listen port, preserving operator-configured overrides
- Added server/src/__tests__/server-startup-feedback-export.test.ts regression test

## Verification

- Regression tests pass: pnpm test:run
- Typecheck passes: pnpm -r typecheck
- The fix ensures PAPERCLIP_API_URL (set via PAPERCLIP_API_KEY override) wins over the derived PAPERCLIP_RUNTIME_API_URL
- Port preservation verified: operator-configured :8443 port in publicBaseUrl is not overwritten with :3001

## Risks

- Low. This is a bug fix that restores intended behavior. The precedence swap is well-tested.
- No breaking changes to the API surface.

## Model Used

None — human-authored.

### DIFF
--- a/packages/adapter-utils/src/execution-target.ts
+++ b/packages/adapter-utils/src/execution-target.ts
@@ -1738,8 +1738,8 @@ export async function startAdapterExecutionTargetPaperclipBridge(input: {
       : DEFAULT_SANDBOX_CALLBACK_BRIDGE_MAX_BODY_BYTES;
   const hostApiUrl =
     input.hostApiUrl?.trim() ||
-    process.env.PAPERCLIP_RUNTIME_API_URL?.trim() ||
     process.env.PAPERCLIP_API_URL?.trim() ||
+    process.env.PAPERCLIP_RUNTIME_API_URL?.trim() ||
     resolveDefaultPaperclipApiUrl();
   const shellCommand = adapterExecutionTargetShellCommand(target);
   const runner = adapterExecutionTargetCommandRunner(target);
--- a/packages/adapter-utils/src/server-utils.ts
+++ b/packages/adapter-utils/src/server-utils.ts
@@ -1971,8 +1971,8 @@ export function buildPaperclipEnv(agent: { id: string; companyId: string }): Rec
   );
   const runtimePort = process.env.PAPERCLIP_LISTEN_PORT ?? process.env.PORT ?? "3100";
   const apiUrl =
-    process.env.PAPERCLIP_RUNTIME_API_URL ??
     process.env.PAPERCLIP_API_URL ??
+    process.env.PAPERCLIP_RUNTIME_API_URL ??
     `http://${runtimeHost}:${runtimePort}`;
   vars.PAPERCLIP_API_URL = apiUrl;
   return vars;
--- a/server/src/__tests__/paperclip-env.test.ts
+++ b/server/src/__tests__/paperclip-env.test.ts
@@ -29,7 +29,7 @@ afterEach(() => {
 });
 
 describe("buildPaperclipEnv", () => {
-  it("prefers an explicit PAPERCLIP_RUNTIME_API_URL", () => {
+  it("prefers an explicit PAPERCLIP_API_URL over PAPERCLIP_RUNTIME_API_URL", () => {
     process.env.PAPERCLIP_RUNTIME_API_URL = "http://203.0.113.42:3102";
     process.env.PAPERCLIP_API_URL = "http://localhost:4100";
     process.env.PAPERCLIP_LISTEN_HOST = "127.0.0.1";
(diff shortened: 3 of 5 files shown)


mohr-and-more:
## Context for reviewers

This is a critical bug fix for the agent runtime API URL injection:

**Bug:** PAPERCLIP_RUNTIME_API_URL was derived from the listen port (:3001) and took precedence over the correct PAPERCLIP_API_URL override, causing all agent runs to receive unreachable API URLs.

**Fix:** 3 changes:
1. server/src/index.ts: Use configuredApiUrl instead of runtimeApiUrl for PAPERCLIP_RUNTIME_API_URL
2. adapter-utils: Swap precedence — check PAPERCLIP_API_URL BEFORE PAPERCLIP_RUNTIME_API_URL  
3. server/src/index.ts: Only rewrite authPublicBaseUrl port when it matches listen port (preserves operator :8443 overrides)

**Testing:** Regression tests in packages/adapter-utils/src/execution-target-sandbox.test.ts cover Runtime-Precedence + Override-Sieg.

**Downstream impact:** M&M BIZ (b7e413ab) has been running this fix in production via dist-hotpatch on brainbug01 for 10 days. No regressions.

**Urgency:** Hotpatch is holding production — this PR makes it permanent in the upstream release.

santhiprakash:
Updated the fix to preserve explicit non-loopback auth.publicBaseUrl values even when detectPort shifts the listener.

- : restricts  to loopback URLs only, so an operator-configured reverse proxy hostname is no longer rewritten to the internal port.
- : adds a regression for  staying stable when detectPort moves the listener to .

Verification: attempted the targeted startup test in the PR worktree, but the checkout has no installed workspace dependencies, so  is missing and the test runner cannot start here.

santhiprakash:
Updated the fix to preserve explicit non-loopback auth.publicBaseUrl values even when detectPort shifts the listener.

- server/src/index.ts: restricts rewriteLocalUrlPort() to loopback URLs only, so an operator-configured reverse proxy hostname is no longer rewritten to the internal port.
- server/src/__tests__/server-startup-feedback-export.test.ts: adds a regression for http://my-host.ts.net:3100 staying stable when detectPort moves the listener to 3110.

Verification: attempted the targeted startup test in the PR worktree, but the checkout has no installed workspace dependencies, so cli/node_modules/tsx/dist/cli.mjs is missing and the test runner cannot start here.

santhiprakash:
Updated the startup rewrite path and kept the reverse-proxy case stable. The current `server/src/index.ts` logic only rewrites loopback auth public URLs, so a non-loopback operator URL like `http://brainbug01.tail802293.ts.net:8443` is preserved even when `detectPort` moves the listener. I also kept the regression coverage for that explicit proxy case in `server/src/__tests__/server-startup-feedback-export.test.ts`. I attempted to run the focused Vitest files, but this checkout does not have `vitest` installed in node_modules, so the command failed locally with `ERR_PNPM_RECURSIVE_EXEC_FIRST_FAIL Command "vitest" not found`.

santhiprakash:
The current branch keeps the auth public URL rewrite limited to loopback hosts only, so a reverse-proxy hostname like `http://my-host.ts.net:3100` is preserved even when `detectPort` moves the listener. I re-read `server/src/index.ts` and the regression in `server/src/__tests__/server-startup-feedback-export.test.ts` covers that case.

Verification: I tried to run the focused startup test with `pnpm test:run server/src/__tests__/server-startup-feedback-export.test.ts`, but this checkout does not have the workspace test dependency installed (`cli/node_modules/tsx/dist/cli.mjs` is missing), so the local run stops in `preflight:workspace-links` before Vitest starts.

santhiprakash:
Updated the startup path to stop rewriting explicit auth public URLs when detect-port shifts the listener. This keeps operator-owned reverse-proxy URLs stable and removes the remaining overreach in server startup.

Verification: 
> paperclip@ test:run /home/ubuntu/Projects/open-source/paperclip-pr10517
> pnpm run preflight:workspace-links && node scripts/run-vitest-stable.mjs "server/src/__tests__/server-startup-feedback-export.test.ts"


> paperclip@ preflight:workspace-links /home/ubuntu/Projects/open-source/paperclip-pr10517
> node cli/node_modules/tsx/dist/cli.mjs scripts/ensure-workspace-package-links.ts

 ELIFECYCLE  Command failed with exit code 1.
 WARN   Local package.json exists, but node_modules missing, did you mean to install?
 ELIFECYCLE  Command failed with exit code 1.
 WARN   Local package.json exists, but node_modules missing, did you mean to install? still cannot start in this checkout because  is missing, so the local preflight stops before Vitest. The branch now reflects the code fix; the remaining blocker is repository setup, not the patch itself.

PR #10521, opened 2026-07-30
feat(sandbox): declare and collect the advisory read-write intent

nickyleach:
## Thinking Path

> - Paperclip is the control plane for AI agent companies
> - Sandbox providers move files between the host and the sandbox
> - The runtime needs advisory data for which sync paths are writable
> - The host already knows that intent when it prepares the sync map
> - Daytona can record that intent for later use
> - This pull request adds the advisory access field and collects writable paths
> - The benefit is that later runtime work can use the data without changing current flow

No public GitHub issue exists for this change. The description below follows the feature-request template fields.

### Problem or motivation
An agent runs inside an ephemeral sandbox. Some sandbox paths keep agent changes; other paths do not. Today the runtime has no declared signal for which sync destinations the agent may change and keep. A later feedback wrapper needs this signal to give the agent real-time feedback when a write lands on a non-persistent path.

### Proposed solution
Add an optional advisory `access: "rw" | "ro"` field to the sync file-mapping types. The host sets `rw` for the workspace, git-history, and asset destinations, and `ro` for referenced-project trees. An absent value defaults to `ro`. The Daytona provider collects the `rw` target directories into a per-lease writable set for later use. This change adds the metadata and the collection only. No execution path reads the writable set yet, so runtime behavior does not change.

### Alternatives considered
Derive a static writable set in provider code. This alternative is weaker: the layer that authors each sync destination already knows the intent, so a per-operation declaration is more accurate and does not hard-code a path list.

### Roadmap alignment
This is the first step toward an advisory sandbox feedback wrapper. The wrapper is best-effort and adds no security. The ephemeral sandbox stays the only boundary.

### Additional context
The field is advisory metadata. It does not change the file transfer and adds no security.

## What Changed

- Added an optional `access` field to the sync file mapping types.
- Set the workspace, git history, and asset destinations to writable.
- Set referenced project destinations to read only.
- Added a writable set store in the Daytona provider.
- Recorded the parent directory of each writable sync mapping during sync in.

## Verification

- The handoff reports these checks before push.
- `pnpm --filter @paperclipai/adapter-utils exec vitest run sandbox-managed-runtime.test.ts`
- `pnpm --filter @paperclipai/plugin-sdk exec tsc --noEmit`
- `vitest run src/plugin.test.ts` in the Daytona package
- `tsc --noEmit` in the Daytona package

## Risks

- Low risk.
- The new field is advisory.
- The writable set store is best effort and in memory.
- A cold store falls back to the workspace baseline.

## Model Used

OpenAI Codex, GPT-5, tool use, code execution.

## Checklist

- [x] I have specified the model used with version and capability details
- [x] I have either linked existing issues or described the issue in the PR
- [x] I have not referenced internal or instance-local Paperclip issues or links (only public GitHub `#NNN` / `github.com/paperclipai/paperclip` URLs)
- [ ] I have considered and documented any risks above

### DIFF
--- a/packages/adapter-utils/src/sandbox-managed-runtime.test.ts
+++ b/packages/adapter-utils/src/sandbox-managed-runtime.test.ts
@@ -93,6 +93,41 @@ function attachNativeRecordingSyncIn(
   };
 }
 
+// A capturing `syncIn` that records every operation for assertion and
+// materializes file AND directory mappings. A directory mapping (a referenced
+// project) uses `mirrorDirectory`, so a test can assert the advisory `access`
+// intent on directory mappings as well as file mappings.
+function attachCapturingSyncIn(
+  client: SandboxManagedRuntimeClient,
+  captured: SandboxSyncOperation[],
+): void {
+  client.syncIn = async (operations: SandboxSyncOperation[]): Promise<SandboxSyncResult> => {
+    const resultOperations: SandboxSyncResult["operations"] = [];
+    for (const operation of operations) {
+      captured.push(operation);
+      let filesTransferred = 0;
+      let bytesTransferred = 0;
+      for (const mapping of operation.files) {
+        await mkdir(path.posix.dirname(mapping.targetPath), { recursive: true });
+        if (mapping.kind === "directory") {
+          await mirrorDirectory(mapping.sourcePath, mapping.targetPath);
+        } else {
+          const bytes = await readFile(mapping.sourcePath);
+          await writeFile(mapping.targetPath, bytes);
+          if (mapping.mode != null) await fsPromises.chmod(mapping.targetPath, mapping.mode);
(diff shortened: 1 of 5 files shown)


nickyleach:
Thank you for the careful read. This change only declares and collects the read/write intent. No execution path reads the writable set yet, so it changes no runtime behavior.

You are right that the workspace, git, and asset mappings carry the staging archive path as the target, so the recorded directory is the runtime staging directory, not the directory that receives the extracted files. The extract destinations — the workspace directory and each asset directory — are known where the archives are authored and are also returned from the runtime preparation result. The consumer that binds the writable set will use those real destinations, and the recorded value moves to the destination directory together with that work, before any binding depends on it.

This change stays scoped to the declaration and the collection so the metadata path is reviewable on its own. The destination-accuracy refinement lands with the consumer that reads the set.

PR #9367, opened 2026-07-10
feat: quota fallback — reassign issue to a fallback agent on quota exhaustion (#2014)

ThomasGinter:
## Thinking Path

> - Paperclip is the open source app people use to manage AI agents for work
> - The heartbeat/retry subsystem decides what happens after a run fails
> - Today a `provider_quota` failure just schedules a same-agent retry and parks the issue until the provider's quota window resets — even when a differently-adapted agent (a different provider) could pick the issue up immediately
> - That wastes wall-clock time and blocks the issue on a single provider's rate limit
> - This pull request adds an opt-in `adapterConfig.fallback` block so a quota-exhausted (or optionally max-turn-exhausted) run can reassign the issue to a configured fallback agent instead of waiting on the same-agent retry
> - The benefit is faster recovery from provider-specific quota exhaustion without any session-sharing or schema changes, fully backward compatible when unset

## Linked Issues or Issue Description

Fixes: #2014

## What Changed

- Added opt-in `adapterConfig.fallback` config (`enabled`, `agentId`, `on`, `when`), validated by a new strict Zod schema (backward compatible, defaults preserve current behavior).
- Inserted `maybeFallbackReassign` before `scheduleBoundedRetryForRun` in both the max-turn and transient-recovery heartbeat finalize branches.
- On a matching failure: reassigns `issues.assigneeAgentId` to the fallback agent, releases the execution lock, writes an `issue.quota_fallback_reassigned` activity entry and a handoff issue comment, stamps the failed run's `resultJson` with an auditable `fallback` marker, and wakes the fallback agent. The original agent's same-adapter retry is suppressed.
- Fallback target validation chain: same company, different agent, invokable — any misconfiguration logs a warning and falls through unchanged to the existing retry/parking behavior.
- Fixed a test-only race in `heartbeat-quota-fallback.test.ts`: `maybeFallbackReassign` keeps writing `heartbeat_run_events`/activity/comment rows after the run's terminal status is already visible, which could trip the `afterEach` cleanup's FK ordering and let un-polled assertions run ahead of those writes. The `afterEach` now retries the whole ordered delete sequence on FK violation, and the test polls for the fallback's terminal run event before asserting on the earlier activity/comment side effects.
- Restored `.agents/skills/create-issue-interaction-ui/SKILL.md` to match `origin/master` — this branch had forked before master reverted a frontmatter-shortening change, and the stale wording was failing `paperclip-skill-utils.test.ts` (unrelated to this PR's feature).

## Verification

[...]

PR #10618, opened 2026-08-01
Defer Claude weekly-limit retries until reset

### LINKED ISSUE #6499: Claude weekly-limit errors classified as `claude_transient_upstream` and retried → per-company spawn-queue cascade
## Summary

When the Claude subscription account hits its **weekly usage limit** (Pro/Max plans only — the Anthropic API has no weekly cap), every `claude_local` heartbeat run fails fast with `Claude run failed: subtype=success: You've hit your weekly limit · resets <date>, <time> (<tz>)`. Paperclip classifies this as `error_code='claude_transient_upstream'` and **retries**, which is wrong: the limit is a hard wall until the named reset (typically multiple days away).

The retry flood then saturates the per-company concurrency pool, runs accumulate in `running` status without `process_pid`, the 15-minute spawn watchdog reaps them as `spawn_watchdog_timeout`, and agents flip to `status='error'`. The fleet is effectively down until a manual recovery (DB cleanup + `systemctl restart paperclip.service` + agent revival).

## Real-world incident (2026-05-21)

One production fleet, ~17 active agents:

| Time (ET) | Event |
|---|---|
| `00:04` | First `claude_transient_upstream` failure with `weekly limit` text |
| `00:04–11:03` | 574 retries, all fast-failing with the same weekly-limit error |
| `~11:08` | Per-company spawn queue jams completely; no run gets a `process_pid` after this point |
| `11:08–12:45` | 37 `spawn_watchdog_timeout` failures; 12 of 17 active agents → `status='error'` |
| `12:30–12:55` | Manual recovery: fail zombie `heartbeat_runs` (49 rows queued/running with NULL `process_pid`), restart service, revive error agents |

A second company on the same VM on a **different** Claude account continued to spawn and succeed throughout — confirming the failure is per-account, not per-host or platform-wide.

## Why the misclassification matters

`claude_transient_upstream` implies *retryable upstream blip* (5xx, momentary 529 overload). The weekly-limit error has the opposite shape: it's deterministic, account-scoped, and includes the exact reset timestamp in the error text. Retrying it generates pure churn — and because each retry consumes a heartbeat-run row + concurrency slot + watchdog timer, the churn is what actually takes the fleet down. The 4-day quota wall becomes a ~12-hour spawn-queue meltdown without manual intervention.

## Proposed fix

In the `claude_local` adapter result classifier:

1. Detect the `weekly limit` substring (or, better, whatever structured signal the Claude CLI emits — there may be a JSON field on the `--output-format stream-json` final message).
2. Map to a new non-retryable `error_code`, e.g. `claude_account_quota_exceeded`.
3. Optionally parse the reset timestamp from the message and either:
   - back the affected adapter/account off until `resetsAt`, or
   - flip the agents that share that account into a `paused` posture with a clear `pauseReason` and an auto-resume at the reset time.

A non-retryable classification alone would have prevented the cascade in this incident.

## Code pointer

I don't have the upstream source path, but the runtime sets `error_code='claude_transient_upstream'` with `error` text beginning `Claude run failed: subtype=success: You've hit your weekly limit · resets <date>, <time> (<tz>)`. The mapping is presumably in the `claude_local` adapter's result handler.

## Adjacent finding (separate bug, surfaced during recovery)

The `systemctl restart paperclip.service` that was necessary to clear the wedged spawn worker also applied 11 pending DB migrations (0075–0085) and loaded the on-disk `@paperclipai/server` code. One of those migrations introduced `company_secret_bindings` enforcement (`assertBindingContext` in `services/secrets.js`) — but `company_secret_bindings` was **empty**, with no migration backfill from existing agents' `adapter_config.env` secret_refs. Every agent with a `secret_ref` instantly failed `adapter_failed: Secret is not bound to agent:X at env.Y`. Workaround: re-`PATCH /agents/:id` each affected agent with its current `adapterConfig` so the route's `syncEnvBindingsForTarget` populates bindings.

If you'd like that filed as its own issue, happy to do so.

goldinbidsheets-rgb:
## Thinking Path

> - The Claude local adapter classifies provider failures for the heartbeat scheduler
> - Claude can return a weekly-limit refusal inside a zero-token `subtype=success` result
> - The success subtype currently bypasses quota classification and creates immediate retry churn
> - The refusal also contains a dated reset that the current clock-only parser cannot use
> - This pull request classifies that envelope as `provider_quota` and carries its reset time to the scheduler
> - The benefit is one deferred retry at the quota reset instead of repeated no-work runs before the reset

Fixes #6499

Refs #10616. That pull request shares quota parsing with the ACP path. It does not guard the Claude CLI `subtype=success` envelope.

Refs #9548. That pull request makes structured success authoritative. This pull request defines the narrow zero-token quota-refusal exception.

Refs #9367. This change supplies the stable `provider_quota` signal that optional fallback work consumes.

## What Changed

- Match the `You've hit your weekly limit` and monthly-limit forms as provider quota failures.
- Parse dated reset text such as `Aug 5, 7am (America/New_York)` into an absolute retry time.
- Reject stale dated resets instead of rolling them into the next year. Keep valid year-boundary resets within a 32-day quota window.
- Treat a zero-token quota result as failed even when Claude reports `subtype=success` and `is_error=false`.
- Keep ordinary structured success results successful. Add parser and execution regressions for both paths.

No schema, migration, configuration, or public API change is required. No user documentation changes are required because `provider_quota` and reset-aware scheduling already exist.

## Verification

- `corepack pnpm exec vitest run packages/adapters/claude-local/src/server/parse.test.ts --reporter=dot` - 41 passed.
- `corepack pnpm exec vitest run server/src/__tests__/claude-local-execute.test.ts -t "zero-token weekly-limit" --reporter=dot` - 1 passed and 24 skipped by the name filter.
- `corepack pnpm exec tsc --noEmit -p packages/adapters/claude-local/tsconfig.json` - passed.
- `corepack pnpm exec tsc --noEmit -p server/tsconfig.json` - passed after the required plugin SDK build-dependency step.
- `git diff --check` - passed.

The existing provider-quota scheduler test already asserts one pending retry at `retryNotBefore`. A local attempt to run that embedded-Postgres test stopped before test discovery because the current dependency graph raises `ERR_REQUIRE_CYCLE_MODULE` in `drizzle-orm` on Node 22.22.0 and Node 24.13.1. This pull request does not change the scheduler or Drizzle code.

## Risks

- Low behavioral risk. The success-envelope override requires both quota wording and zero token usage.
- A parser false positive can defer a run. The phrase match stays limited to provider limit language, and the reset window is at most 32 days.
- The parser part overlaps with #10616. If #10616 lands first, this branch can use its shared helper while it keeps the unique success-envelope guard and execution test.
- Rollback is one commit. Reverting it restores the prior classifier and retry behavior.

> For core feature work, check [`ROADMAP.md`](ROADMAP.md) first and discuss it in `#dev` before opening the PR. Feature PRs that overlap with planned core work may need to be redirected - check the roadmap first. See `CONTRIBUTING.md`.

## Model Used

OpenAI Codex, GPT-5-based runtime. The exact serving ID and context window are not exposed. The runtime used repository tools, shell execution, and code editing.

## Checklist

### DIFF
--- a/packages/adapters/claude-local/src/server/execute.ts
+++ b/packages/adapters/claude-local/src/server/execute.ts
@@ -59,6 +59,7 @@ import {
   claudeModelUsageTotals,
   parseClaudeStreamJson,
   describeClaudeFailure,
+  detectClaudeOverflowHold,
   detectClaudeLoginRequired,
   extractClaudeRetryNotBefore,
   isClaudeMaxTurnsResult,
@@ -971,17 +972,20 @@ export async function execute(ctx: AdapterExecutionContext): Promise<AdapterExec
       };
     }
 
+    const overflowHold = detectClaudeOverflowHold(proc.stderr);
+
     if (!parsed) {
-      const fallbackErrorMessage = parseFallbackErrorMessage(proc);
+      const fallbackErrorMessage = overflowHold?.message ?? parseFallbackErrorMessage(proc);
       const providerQuota =
         !loginMeta.requiresLogin &&
-        (proc.exitCode ?? 0) !== 0 &&
-        isClaudeProviderQuotaError({
-          parsed: null,
-          stdout: proc.stdout,
-          stderr: proc.stderr,
-          errorMessage: fallbackErrorMessage,
-        });
+        (overflowHold !== null ||
+          ((proc.exitCode ?? 0) !== 0 &&
+            isClaudeProviderQuotaError({
+              parsed: null,
+              stdout: proc.stdout,
+              stderr: proc.stderr,
+              errorMessage: fallbackErrorMessage,
+            })));
       const transientUpstream =
         !loginMeta.requiresLogin &&
         !providerQuota &&
@@ -992,14 +996,15 @@ export async function execute(ctx: AdapterExecutionContext): Promise<AdapterExec
           stderr: proc.stderr,
           errorMessage: fallbackErrorMessage,
         });
-      const transientRetryNotBefore = providerQuota || transientUpstream
-        ? extractClaudeRetryNotBefore({
-            parsed: null,
-            stdout: proc.stdout,
-            stderr: proc.stderr,
-            errorMessage: fallbackErrorMessage,
-          })
-        : null;
+      const transientRetryNotBefore = overflowHold?.retryNotBefore ??
+        (providerQuota || transientUpstream
+          ? extractClaudeRetryNotBefore({
+              parsed: null,
+              stdout: proc.stdout,
+              stderr: proc.stderr,
+              errorMessage: fallbackErrorMessage,
+            })
+          : null);
       const errorCode = loginMeta.requiresLogin
         ? "claude_auth_required"
         : isClaudeModelNotFoundError({
@@ -1088,6 +1093,7 @@ export async function execute(ctx: AdapterExecutionContext): Promise<AdapterExec
     const parsedSucceeded =
       parsedSubtype === "success" && !parsedIsError && !successfulQuotaRefusal;
     const failed =
+      overflowHold !== null ||
       successfulQuotaRefusal ||
       (!parsedSucceeded && ((proc.exitCode ?? 0) !== 0 || parsedIsError));
     // Validate-before-persist guard: never persist a sessionId whose transcript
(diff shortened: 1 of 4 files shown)


juancarlosrial76-code:
Production corroboration for this change, from a self-hosted instance.

We hit this failure mode on 2026-08-03. Between 00:08Z and 04:56Z, 118 runs failed in a fixed ~30-second retry chain against a single message:

```
You've hit your weekly limit · resets Aug 3, 7am (Europe/Zurich)
```

The reset time was present in every one of those messages. No run deferred to it.

I arrived at the same two-part diagnosis independently, then found this PR already carries it. I checked this branch against our exact observed strings (`now = 2026-08-03T04:56:00Z`):

| message | `isClaudeProviderQuotaError` | `extractClaudeRetryNotBefore` |
|---|---|---|
| `You've hit your session limit - resets 11:20am (Europe/Zurich)` | `true` | `2026-08-03T09:20:00.000Z` |
| `You've hit your weekly limit · resets Aug 3, 7am (Europe/Zurich)` | `true` | `2026-08-03T05:00:00.000Z` |
| `You've hit your weekly limit · resets 7am (Europe/Zurich)` | `true` | `2026-08-03T05:00:00.000Z` |

All three are correct. On `origin/master` all three weekly rows are `false` / `null`.

One point worth emphasising for reviewers, because it is easy to read this PR as two loosely related changes:

**The two halves only work together.** Widening the alternation alone does not fix the weekly case. `parseClaudeResetClockTime` anchors its match at `^(\d{1,2})`, so `Aug 3, 7am` still returns `null` even once the message classifies as quota — `retryNotBefore` stays null and the backoff still runs empty. The dated-reset parsing is not a nice-to-have next to the wording fix; it is the half that actually produces a retry time for the wording this PR adds.

Two smaller notes:

1. The alternation here is `(?:session|weekly|monthly)`. We also see `you've hit your usage limit` in the wild from the Codex CLI, and `usage` would be a cheap fourth branch here — though `usage limit reached` already covers part of that space, so this may be deliberate.
2. Rejecting stale dated resets and capping at a 32-day window reads safer than rolling the year forward unconditionally, which is what I had reached for first. Agreed with the choice made here.

Happy to see this land as-is. I have no competing change to offer — this PR is a strict superset of what I had.

Issue #5763, opened 2026-05-12
Productivity-review watchdog issues auto-create but never auto-close


stacksup:
## Summary

A "Review productivity for X" watchdog routine creates child issues that run a productivity scan, post evidence as a comment, but never advance their own status from `in_progress`. Result: every long-running source issue accumulates one or more stale watchdog issues over time.

## Observed on

Self-hosted Paperclip, embedded-postgres mode, instance running since 2026-03-29. Discovered during a portfolio healthcheck on 2026-05-12.

## Pattern

For a source issue (e.g. `HQ-232`) that stays `in_progress` past some threshold, the routine creates a child issue titled `Review productivity for HQ-232`. The child:

1. Gets checked out by a heartbeat run.
2. Posts a comment like: `Productivity review evidence refreshed. - Source issue: ... - Trigger: long_active_duration ...`
3. The heartbeat run completes (`status='succeeded'` in `heartbeat_runs`).
4. The issue status remains `in_progress` with `checkout_run_id` still set.
5. No further activity. The issue stays stale indefinitely.

## Evidence from our instance

| Identifier | Title | Stale for | Heartbeat runs |
|---|---|---|---|
| HQ-242 | Review productivity for HQ-232 | 6d 15h | 3 succeeded |
| HQ-251 | Review productivity for HQ-244 | 5d 16h | 3 succeeded |
| HQ-288 | Review productivity for HQ-155 | 3d 2h | 3 succeeded |
| GLO-762 | Review productivity for GLO-759 | 1d 20h | 2 succeeded |

In every case the most recent heartbeat run completed cleanly (status `succeeded`, no error), but no `PATCH /api/issues/{id}` to move the watchdog issue to `done` ever fired. The source issues all had ongoing legitimate activity (intentional holds waiting on milestones), so the trigger condition (`long_active_duration`) keeps re-firing and spawning more watchdog issues, which then also get stuck.

## Workaround

Manually advance stale watchdog issues:

```sql
UPDATE issues
SET status = 'done',
    checkout_run_id = NULL,
    execution_run_id = NULL,
    execution_locked_at = NULL,
    completed_at = COALESCE(completed_at, now()),
    updated_at = now()
WHERE title LIKE 'Review productivity for %'
  AND status = 'in_progress';
```

## Expected behaviour

After the watchdog has posted its evidence comment, the issue should auto-advance to `done` (or `cancelled` if the routine's purpose was purely informational). Either:

[...]

PR #10667, opened 2026-08-02
fix(productivity-review): close reviews whose source issue reached a terminal status

projecttop10tech-arch:
## Thinking Path

> - The productivity-review service files a "Review productivity" issue for a manager when an assignee looks stalled
> - The service creates those reviews automatically, but it never closes them automatically
> - A review therefore stays open after its source issue reaches `done` or `cancelled`, because only a manager decision closes it
> - The manager then opens a review, finds the source work already finished, and closes the review by hand — repeatedly
> - This pull request adds an auto-resolution pass that cancels a review when its source issue is terminal
> - The benefit is that the manager queue holds only reviews that still have work to judge

Refs #5763

**Bug report**

- **What happened.** A productivity review is created while the source issue is legitimately `in_progress`. The assignee then finishes the work and the source issue moves to `done`. The review stays open. In a sample of six reviews on one instance, four had a source issue that was already `done` at the time a human opened the review.
- **Expected behavior.** A review with a terminal source issue has nothing left to judge, so the system must close it.
- **Steps to reproduce.**
  1. Let `reconcileProductivityReviews` create a review for an issue.
  2. Set the source issue status to `done`.
  3. Run `reconcileProductivityReviews` again.
  4. The review is still open.
- **Deployment mode.** Self-hosted, local dev server.
- **Commit.** Reproduced on `master` at the time of writing.

Note on detection: the detector cannot fire on a terminal source. The candidate query restricts to `todo` and `in_progress`, and `elapsedMs` is `null` unless the source is `in_progress`. The gap is only in the lifetime of a review that was already created.

Related open pull request:

- #10666 changes the `long_active_duration` trigger in the same file. The two changes touch different functions and do not conflict.

## What Changed

- `server/src/services/productivity-review.ts`: adds `autoResolveTerminalSourceReviews`. `reconcileProductivityReviews` runs it before the candidate scan.
- The pass lists open reviews (`originKind = issue_productivity_review`, status not `done` or `cancelled`), reads the source issue named by `originId`, and cancels the review when the source is `done` or `cancelled`.
- The pass writes a one-line comment before it cancels, prefixed with the new exported constant `PRODUCTIVITY_REVIEW_AUTO_RESOLVE_COMMENT_PREFIX`.
- The pass records an `issue.productivity_review_auto_resolved` activity entry.
- The pass skips a review that has a queued, running, or retry-scheduled run, because an agent can be judging that review at that moment. This mirrors the `hasActiveRunForIssueId` guard in the liveness-recovery retirement pass.
- A failed comment or update is logged and skipped. It does not stop the pass.
- The reconcile result gains `autoResolved` and `autoResolvedIssueIds`.
- `server/src/__tests__/productivity-review-service.test.ts`: adds two tests — one that flips the source to `done` and asserts the review is cancelled with the comment present and the pass is idempotent on a third run, and one that asserts a non-terminal source leaves the review open.

The review is set to `cancelled`, not `done`. `countConsecutiveNoActionProductivityReviews` counts only `done` reviews toward its suppression streak, so a system closure must not enter that count as if a manager had made a decision.

## Verification

```
npx vitest run server/src/__tests__/productivity-review-service.test.ts
  Test Files  1 passed (1)
       Tests  17 passed (17)

npx vitest run server/src/__tests__/attention-service.test.ts \
              server/src/__tests__/heartbeat-responsible-user-invariant.test.ts
  Test Files  2 passed (2)
       Tests  19 passed (19)

  clean
```

## Risks

Low.

- The pass adds one query per open review on each reconcile. The list is capped at 250 reviews per reconcile, the same cap that the candidate scan already uses. On a healthy instance the open-review count is small.
- A manager loses the chance to record a decision on a review whose source has already finished. That decision has no effect on the source work, so the loss is small. The comment and the activity entry keep the closure auditable.
- No migration. No schema change. No telemetry contract change.

## Model Used

Claude Opus 5 (`claude-opus-5`), 1M context window, extended thinking, tool use through Claude Code.

## Checklist

### DIFF
--- a/server/src/__tests__/productivity-review-service.test.ts
+++ b/server/src/__tests__/productivity-review-service.test.ts
@@ -19,6 +19,7 @@ import {
   DEFAULT_PRODUCTIVITY_REVIEW_MAX_REFRESH_COMMENTS,
   DEFAULT_PRODUCTIVITY_REVIEW_NO_COMMENT_STREAK_RUNS,
   DEFAULT_PRODUCTIVITY_REVIEW_REFRESH_INTERVAL_MS,
+  PRODUCTIVITY_REVIEW_AUTO_RESOLVE_COMMENT_PREFIX,
   PRODUCTIVITY_REVIEW_REFRESH_COMMENT_PREFIX,
   PRODUCTIVITY_REVIEW_ORIGIN_KIND,
   productivityReviewService,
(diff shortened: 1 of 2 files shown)


projecttop10tech-arch:
Both Greptile findings were valid and are fixed in `fd69f491b`.

**Issue 1 — partial resolution duplicates comments.** The audit comment and the status update are two writes; a failure between them left the review open with the comment already written, so the next reconciliation wrote it again. The pass now checks for an existing auto-resolution comment (prefix match, same shape as the existing `getRefreshCommentState` guard) before writing one, so a retry writes only the missing status update.

**Issue 2 — fixed page starves later reviews.** The scan read the oldest 250 open reviews and filtered on source status inside the loop, so 250 open-but-not-actionable reviews occupied the page permanently. The scan now inner-joins the source issue and filters `source.status in ('done','cancelled')` in SQL, so the page only ever holds rows the pass can act on and it drains as they resolve.

One detail worth flagging for review: the join casts `issues.id::text` rather than `origin_id::uuid`. `origin_id` is a text column that does not always hold a uuid, so casting that side would raise a parse error on unrelated origin kinds.

Tests added for both, as noted in the file overview:
- `does not repeat the audit comment when an earlier pass commented but failed to cancel` — seeds the exact partial-write state, asserts exactly one prefixed comment and `cancelled`.
- `auto-resolves a terminal-source review that sits behind a full page of older open reviews` — seeds `MAX_OPEN_REVIEWS_PER_RECONCILE` older non-terminal-source reviews plus one newer actionable one. Verified load-bearing: with the SQL status filter removed it fails, with it present it passes.

```
npx vitest run server/src/__tests__/productivity-review-service.test.ts
 Test Files  1 passed (1)
      Tests  19 passed (19)

pnpm --filter @paperclipai/server typecheck   # clean
```

projecttop10tech-arch:
Both P1s from the second review are fixed in `50c6f956c`.

**Concurrent reconciliation duplicates comments.** Correct — the comment check and the comment write are separate statements, so the guard I added in `fd69f491b` only closed the sequential-retry case, not the overlap case. The pass now runs under a transaction-scoped advisory lock, using the `pg_advisory_xact_lock(hashtextextended(...))` idiom already in `folders.ts` and `decisions.ts`. Two notes on the choice:

- `pg_try_advisory_xact_lock` rather than the blocking variant: a second concurrent pass has nothing to contribute, so skipping and picking the work up next cycle is better than queueing behind the first and re-reading a page that has already been drained.
- Transaction-scoped rather than session-scoped, because the session variant leaks the lock onto a pooled connection if the pass throws between acquire and release.

**Active runs still starve reviews.** Also correct, and it was the same defect as the original finding one predicate over — I moved the source-status filter into SQL but left the active-run filter in the loop, so 250 terminal-source reviews under active judgement would fill the page just as effectively. The scan now excludes them with `not exists` over queued/running/scheduled_retry runs. I kept the in-loop guard as a narrow race check for a run that starts after the page is read; it can no longer starve anything, since any row reaching it was actionable at selection time.

Tests for both, and I checked each against the unfixed code rather than assuming:

| Test | Behaviour without the fix |
|---|---|
| `writes one audit comment when two reconciliations overlap on the same review` | `expected [1, 1] to deeply equal [0, 1]` — both passes resolve the same review |
| `auto-resolves past a full page of terminal-source reviews that agents are judging right now` | `expected [] to deeply equal [Array(1)]` — nothing resolves |

```
npx vitest run server/src/__tests__/productivity-review-service.test.ts
 Test Files  1 passed (1)
      Tests  21 passed (21)

pnpm --filter @paperclipai/server typecheck   # clean
```

PR #10673, opened 2026-08-02
fix(claude-local): read Anthropic OAuth token from macOS Keychain

ekomisarenko:
## Summary
- On a default macOS install, the Claude Code CLI stores its OAuth token in the system Keychain (service `Claude Code-credentials`), not in `~/.claude/.credentials.json` or `credentials.json`. `readClaudeToken()` only checked those two file paths, so it always returned `null` on macOS installs that use the Keychain — silently forcing every quota poll onto the much more brittle fallback that drives the CLI's interactive `/usage` screen through a pty scrape, instead of the plain HTTP `/api/oauth/usage` endpoint.
- `readClaudeToken()` now falls back to `security find-generic-password -s "Claude Code-credentials" -w` after both credentials-file checks miss, gated to `darwin` only. The existing file-based checks are unchanged and still take priority.
- Verified live: pulled a real token from Keychain on a macOS host and called `https://api.anthropic.com/api/oauth/usage` with it directly — `HTTP 200`, real `five_hour.utilization`. Confirms the HTTP path (`fetchClaudeQuota`) works fine once it actually has a token to use.

## Test plan
- [x] `pnpm exec vitest run src/__tests__/quota-windows.test.ts` (server package) — 62 passed, including 4 new cases covering the Keychain fallback (found via Keychain, Keychain lookup fails, non-macOS skip, credentials-file takes priority over Keychain)
- [x] `tsc --noEmit` in `packages/adapters/claude-local` and `server` — clean

### DIFF
--- a/packages/adapters/claude-local/src/server/quota.ts
+++ b/packages/adapters/claude-local/src/server/quota.ts
@@ -85,13 +85,7 @@ function trimToLatestUsagePanel(text: string): string | null {
   return tail;
 }
 
-async function readClaudeTokenFromFile(credPath: string): Promise<string | null> {
-  let raw: string;
-  try {
-    raw = await fs.readFile(credPath, "utf8");
-  } catch {
-    return null;
-  }
+function parseClaudeCredentialsJson(raw: string): string | null {
   let parsed: unknown;
   try {
     parsed = JSON.parse(raw);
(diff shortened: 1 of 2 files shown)


ekomisarenko:
Closing — opened without authorization from our side. Withdrawing while we decide internally whether to contribute this upstream. Apologies for the noise.

PR #10698, opened 2026-08-02
Test adapters in the environment a run would actually use

devinfoley:
## Thinking Path

> - Agents execute in environments — local, SSH, or sandboxes — resolved at run time as agent environment → instance default → local
> - The Configuration page has a Test button that probes the adapter (working directory, command, a model call) in the environment it will run in
> - But the Test sent only the agent's own environment id, with no instance-default fallback, so agents relying on the instance default were probed on the Paperclip host instead
> - A sandbox image carrying an extra CLI then fails the Test with "command not found" even though every real run would resolve to the sandbox and succeed — the Test lies about a working setup
> - This pull request mirrors the run-time resolution in the Test call via a small shared helper with tests
> - The benefit is that the Test button reports the truth about where the agent actually runs

No existing public issue — inline description following the bug report template:

**What happened?**

With the instance default environment set to a sandbox (whose image includes the adapter CLI) and an agent that leaves its environment unset ("use instance default"), the Configuration page's Test fails with `command not found` for that CLI.

**Expected behavior**

The Test probes the environment a real run would use — here the instance-default sandbox, where the CLI exists — and passes.

**Steps to reproduce**

1. Set the instance default environment to a sandbox whose image carries an adapter CLI not installed on the Paperclip host (e.g. `grok`).
2. Create a `grok_local` agent without selecting an environment.
3. Press Test on the agent's Configuration page → `command not found`, while a real heartbeat run resolves to the sandbox and works.

Reproduced on `sha-53bcf38-cloud`-era master; root-caused in `ui/src/components/AgentConfigForm.tsx` (`environmentId = currentDefaultEnvironmentId || null`) versus the server's `resolveExecutionWorkspaceEnvironmentId` (agent → instance default → local).

## What Changed

- New `ui/src/lib/adapter-test-environment.ts`: `resolveAdapterTestEnvironmentId` — agent environment first, else instance default, else null (host probe) — documented as the mirror of the server's run-time resolution.
- `AgentConfigForm` uses it in the Test mutation. The raw agent environment id is now sent even when it points at the local environment: the server already resolves the driver and probes the host for local, so explicit-local behavior is unchanged, and the test-environment route's remote paths (SSH/sandbox lease + custom-image template) engage exactly as they do for the fallback environment.
- Tests pin the fallback (agent wins; instance default when agent unset; null when neither).

Deliberately untouched: the onboarding wizard's adapter test still sends no environment — during onboarding an instance default frequently doesn't exist yet, and changing that flow deserves its own look.

## Verification

- `vitest run` on the new helper suite plus both `AgentConfigForm` suites — 19 tests pass; `tsc` clean in `ui/`.
- Root cause verified against a live deployment: an agent with `default_environment_id = NULL`, instance default = sandbox environment; the Test posted `environmentId: null` and probed the host (no `Probing inside environment: …` check in the result), which lacks the CLI that the sandbox image carries.

## Risks

- Low. The change only widens which environment the Test probes, matching run-time reality. Sandbox-backed tests boot a throwaway sandbox (existing route behavior — lease, custom-image template, archive-on-release), so Tests for instance-default-sandbox agents now take sandbox-boot time instead of failing fast and wrongly.

## Model Used

Claude Fable 5 (`claude-fable-5`, extended thinking, via Claude Code with tool use and code execution); diagnosis included live inspection of a deployed instance's agent/environment configuration.

## Checklist

- [x] I have updated relevant documentation to reflect my changes (helper doc-comment carries the rationale; no user-facing doc covers the Test button)

### DIFF
--- a/ui/src/components/AgentConfigForm.tsx
+++ b/ui/src/components/AgentConfigForm.tsx
@@ -671,26 +671,28 @@ export function AgentConfigForm(props: AgentConfigFormProps) {
       // server resolves the driver and probes the host in that case.
       //
       // Test can be clicked before the settings query settles (or after it
-      // failed with retry:false), so resolve the settings here rather than
-      // trusting the render-time cache — silently probing the host because a
-      // query hadn't finished would resurrect the exact false failure this
-      // resolution exists to fix. ensureQueryData reuses the cached value
-      // when present; a fetch that still fails leaves the honest host probe.
-      let instanceDefaultEnvironmentId = instanceSettings?.defaultEnvironmentId ?? null;
-      if (instanceSettings === undefined) {
+      // failed with retry:false), so when the agent relies on the instance
+      // default, resolve the settings here rather than trusting the
+      // render-time cache. A fetch that still fails FAILS the test with an
+      // honest diagnostic — silently probing the host instead would report
+      // the exact false command-not-found failure this resolution exists to
+      // fix. Agents with their own environment never need the settings.
+      let settings = instanceSettings;
+      if (!rawCurrentDefaultEnvironmentId && settings === undefined) {
         try {
-          const settings = await queryClient.ensureQueryData({
+          settings = await queryClient.ensureQueryData({
             queryKey: queryKeys.instance.settings,
             queryFn: () => instanceSettingsApi.get(),
           });
-          instanceDefaultEnvironmentId = settings?.defaultEnvironmentId ?? null;
         } catch {
-          instanceDefaultEnvironmentId = null;
+          throw new Error(
+            "Could not load instance settings to determine which environment to test in. Retry the test.",
+          );
         }
       }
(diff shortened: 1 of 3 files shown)


devinfoley:
Addressed: the mutation now resolves instance settings via `queryClient.ensureQueryData` before choosing the environment — the render-time cache is only a fast path, so a Test clicked before the settings query settles (or after a retry:false failure) fetches them instead of silently degrading to a host probe. Only a fetch that still fails leaves the honest host fallback.

@greptileai review

devinfoley:
Addressed: a settings fetch that fails now **fails the Test** with an explicit diagnostic ("Could not load instance settings to determine which environment to test in") instead of silently degrading to a host probe — the wrong-machine path is no longer reachable through the error branch. The fetch also only happens for agents that rely on the instance default; agents with their own environment never touch settings.

@greptileai review

PR #10502, opened 2026-07-30
fix(issues): cover monitor-backed in_review blocker lanes

rudyjellis:
## Thinking Path

> - Paperclip is the open source app people use to manage AI agents for work
> - Blocker attention tells operators if a blocked issue has a real waiting path or needs a human now
> - Review follow-up lanes often wait on a scheduled monitor instead of an active run
> - The recovery liveness code already treats that scheduled monitor as a real waiting path
> - But blocker attention did not load or reuse that monitor state for `in_review` leaves
> - So a healthy follow-up lane could be marked as stalled even though it would wake itself later
> - This pull request reuses the shared monitor waiting-path helper in blocker attention and adds a regression for the `in_review` case
> - The benefit is that monitor-backed review lanes stay covered, while truly idle review leaves still need attention

## Linked Issues or Issue Description

No exact public GitHub issue exists for this case, so this section describes the bug.

**What happened.** A blocked parent could point at an `in_review` child that already had a scheduled issue monitor. The child persisted `monitorNextCheckAt`, `executionPolicy.monitor`, and `executionState.monitor`, but blocker attention still reported the chain as stalled.

**What was expected.** A scheduled monitor is a live waiting path. The parent should report that blocker as covered until the monitor expires or is exhausted.

**Why it matters.** Blocker attention is only useful if it is strict about real dead ends. False stalled states on healthy review follow-up lanes create noise and hide the blockers that really need a person.

**Related work.** Dedup search found related upstream work: #10398, #8759, and closed predecessor #8751. This PR stays distinct because it is rebased on current `master`, it matches the live `in_review` follow-up lane shape, and it adds the missing regression for that review-state case.

## What Changed

- Exported the shared scheduled-monitor waiting-path helper from `server/src/services/recovery/issue-graph-liveness.ts`.
- Updated `server/src/services/issues.ts` so blocker-attention classification loads monitor-backed fields for graph nodes and treats a valid scheduled monitor as an explicit waiting path.
- Added a focused regression in `server/src/__tests__/issue-blocker-attention.test.ts` for an `in_review` blocker leaf with a scheduled monitor.

## Verification

- `pnpm run preflight:workspace-links`
- `pnpm exec vitest run server/src/__tests__/issue-blocker-attention.test.ts`

## Risks

- Low risk. The change only widens the covered-path check when a monitor is already scheduled and still valid.
- The main failure mode would be a helper that is too permissive. The regression keeps the `in_review` monitor case pinned to the shared waiting-path rules instead of a new local copy.

[...]

PR #9997, opened 2026-07-22
fix(task-watchdogs): treat a scheduled monitor as a live subtree continuation

fallcent-junslim:
## Thinking Path

> - Paperclip is the open source app people use to manage AI agents for work
> - The task-watchdog subsystem watches a subtree of issues and, when it looks "stopped", wakes a watchdog agent to review it
> - To decide "stopped vs live" the classifier only counted active heartbeat runs and queued wake requests as live continuations
> - It ignored a server-visible **scheduled monitor** — a source parked on a future `monitorNextCheckAt` (recurring/monitored work) has a real live continuation, but the classifier treated it as stopped
> - The stop fingerprint hashes each leaf's `updatedAt` / `latestCommentAt`, so when the watchdog's own review child completes it wakes the source (`issue_children_completed`), bumps those timestamps, mints a *new* fingerprint, and re-arms the watchdog — an unbounded feedback loop
> - This pull request reuses the canonical `hasScheduledMonitor` liveness check inside the subtree classifier and loads the monitor columns in the subtree query
> - The benefit is a monitored source is recognized as live and is no longer repeatedly re-reviewed by a watchdog child completion that only touches its timestamps

## Linked Issues or Issue Description

No public GitHub issue exists; describing the bug in-PR (bug report format).

**What happened:** A watched source with a valid `scheduled` monitor (server-visible `monitorNextCheckAt` in the future, unexpired `timeoutAt`, remaining attempts) was repeatedly classified as `stopped` by the task-watchdog. Each watchdog review child completed normally, which woke the source via `issue_children_completed`, changed the source's `updatedAt`/latest-comment timestamps, produced a new stop fingerprint, and re-triggered the watchdog — observed re-triggering 7×.

**Expected behavior:** A server-visible scheduled monitor (or scheduled retry) is recognized as a live continuation of the watched subtree, so the source is not classified stopped and a normally-completing review child does not re-arm the watchdog.

**Root cause:** `classifyTaskWatchdogSubtree` in `server/src/services/task-watchdogs.ts` derived liveness only from `activeRuns` (queued/running/scheduled_retry) and `queuedWakeRequests`; it never consulted `executionState.monitor` / `monitorNextCheckAt`. The subtree SQL loader also did not select the monitor columns.

**Deployment mode:** server (control plane).

Related open PRs (different angles on nearby watchdog churn, not duplicates):
- #9432 — stabilizes the stop fingerprint against routine timestamp churn (fingerprint side; this PR fixes the liveness side)
[...]

PR #9073, opened 2026-07-06
fix(issues): reject unknown monitor policy fields instead of silently dropping them

dosthcpp:
## Thinking Path

> - Paperclip orchestrates AI agents for zero-human companies; agents schedule issue monitors (`executionPolicy.monitor`) to get woken when an external check is due.
> - Monitor scheduling is driven by API PATCHes that agents compose themselves — and agents routinely guess plausible-but-wrong field shapes.
> - The zod schemas at the create/update boundary are non-strict, so a top-level `monitor` field, or unsupported monitor keys like `prompt` or `intervalMinutes`, are **silently stripped**: the PATCH returns 200 and the agent believes a monitor is scheduled when none (or a partial one) is.
> - For an autonomous fleet this is a liveness hazard: an issue that thinks it has a wake path sleeps forever, and the operator only finds out when the work goes stale.
> - This PR makes the input boundary fail loudly — strict input variants of the execution-policy schemas plus a targeted top-level `monitor` rejection that points at `executionPolicy.monitor` — while stored `execution_policy` rows keep parsing through the lenient base schemas.
> - The benefit is agents get an immediate, actionable 422 instead of a silent no-op, and legacy stored rows keep loading unchanged.

## Linked Issues or Issue Description

No existing GitHub issue tracks this; describing inline per the bug template:

- **What happened** — `PATCH /api/issues/:id` with `{"monitor": {"nextCheckAt": ...}}` (top level) returns 200 and schedules nothing; `{"executionPolicy": {"monitor": {"nextCheckAt": ..., "prompt": "..."}}}` returns 200 and schedules a monitor with the `prompt` key silently dropped. Both hit us in production fleet operation: an agent scheduled a monitor with a `prompt` field, the field vanished, and the wake fired with no context on what to check.
- **Expected** — an unsupported or misplaced field in a monitor payload should fail the request with a clear validation error, not succeed with data loss.
- **Root cause** — `updateIssueSchema` / `issueExecutionMonitorPolicySchema` are non-strict zod objects, which strip unknown keys by default.
- **Duplicate/related PR search** — searched the PR list for `monitor`, `strict`, `unrecognized keys`; no duplicate or related open PRs found. Related issue #8691 concerns monitor wake/reopen semantics, not input validation.

## What Changed

[...]

PR #10735, opened 2026-08-03
fix(issues): re-arm issue monitors on dispatch so a failed run cannot strand the issue

juancarlosrial76-code:
## Thinking Path

> - The heartbeat scheduler owns issue liveness: every agent-owned issue must keep a path that can wake its assignee again
> - An issue monitor is that path for waits on an async or external system
> - A monitor consumed its own schedule when it fired, and only the woken run could re-arm it
> - A woken run that dies to a provider quota, a crash, or a `SIGTERM` records nothing, so the issue lost its only wake path with no log and no alarm
> - This pull request makes a monitor re-arm itself on dispatch, keeps it bounded, and fixes the classifier that mislabelled the resulting dead card
> - The benefit is that an infrastructure outage pauses work instead of destroying the wake paths of the issues it touches

No public GitHub issue exists for this. The problem, in the form of `.github/ISSUE_TEMPLATE/bug_report.yml`:

**What happened?**

An issue monitor fires exactly once. `buildIssueMonitorTriggeredPatch()` in
`server/src/services/issue-execution-policy.ts` sets `monitorNextCheckAt: null` on every
successful dispatch, and `tickDueIssueMonitors()` in `server/src/services/heartbeat.ts` selects
only rows where that column is non-null. After the first fire the monitor can never find itself
again. Re-arming is the woken run's job. When that run fails terminally, nothing re-arms it and
nothing reports it. The issue stays in `in_review` with `monitorAttemptCount > 0` and no
schedule: reachable by nobody, while its status and comments still look healthy.

`doc/execution-semantics.md` already promised a safety net for this exact case:

> A one-shot issue monitor consumes its persisted `nextCheckAt` when it dispatches the assignee
> wake. If that monitor-consuming run is lost before it records a new disposition or future
> monitor, Paperclip restores exactly one bounded continuation using the existing process-loss
> retry limit.

That net exists in code, but only in `reapOrphanedRuns()`, behind the predicate
`monitorDispatchLostWithoutFutureWake`. It covers runs whose **process disappears**. A run that
ends terminally with an error code such as `claude_transient_upstream` or `provider_quota` never
reaches that path, so the documented guarantee does not hold for the most common failure mode.

A monitor whose bounds (`timeoutAt`, `maxAttempts`) are not yet spent should leave the issue with
a wake path, whatever happened to the run it woke. When the bounds *are* spent, the monitor should
end loudly through `recoveryPolicy`, not by silently dropping its schedule.

**Steps to reproduce**

1. Schedule a monitor on an agent-owned `in_review` issue with `maxAttempts: 12` and a `timeoutAt`
   weeks away.
2. Let the monitor fire.
3. Let the woken run fail with `claude_transient_upstream`.
4. Read the issue row: `monitorAttemptCount` is 1, `monitorNextCheckAt` is `null`, and the monitor
   never runs again even though 11 attempts and the whole timeout window are unspent.

In one instance this destroyed the wake paths of five issues over three days, all during a window
of repeated upstream failures. `maxAttempts` and `timeoutAt` recorded budgets that could no longer
be spent.

**Related PRs found while searching for duplicates** (none implement this; listing them because
they touch the same fields):

- Refs #4988 — the merged PR that introduced issue monitors, and the source of the one-shot wording
  this PR revises.
- Refs #9073 — proposes rejecting unknown monitor policy fields. It would need to learn the new
  `intervalSeconds` field if both land.
- Refs #8191 — exempts perpetual trackers from stranded recovery using a future `monitorNextCheckAt`.
  Re-armed monitors make that column non-null more often; the label condition in that PR still
  bounds the exemption.
- Refs #9997, #10398, #10502 — treat a scheduled monitor as a live path in other surfaces. They read
  `monitorNextCheckAt`; they do not change who writes it.

## What Changed

- `buildIssueMonitorTriggeredPatch()` re-arms `monitorNextCheckAt` to `triggeredAt + intervalSeconds`
  instead of clearing it. The woken run still owns the decision and normally overwrites the value
  with the cadence the wait needs, or resolves the issue so the monitor clears. The re-armed check
  only governs the failure path.
- Added the optional policy field `intervalSeconds` (60 seconds to 30 days). It defaults to
  `DEFAULT_ISSUE_MONITOR_INTERVAL_SECONDS`, which is 6 hours. Six hours is short enough to serve as
  the liveness net that `monitorDispatchLostWithoutFutureWake` provided for lost processes.
- The re-armed check is clamped to `timeoutAt`. The last wake lands on the deadline, where the
  existing bound check clears the monitor and runs `recoveryPolicy`. It does not overshoot by an
  interval.
- A policy that names neither `maxAttempts` nor `timeoutAt` now falls back to
  `DEFAULT_ISSUE_MONITOR_MAX_ATTEMPTS` (40). A recurring monitor must still terminate, and reaching
  a bound clears it through the existing recovery route, which is loud.
- The re-armed policy is built from the **stored** policy, not the normalized one.
  `normalizeIssueExecutionPolicy()` redacts `monitor.externalRef`, so writing the normalized policy
  back would replace a real external reference with the redaction marker. That reference is load
  bearing: `hasPendingProviderQuotaRecoveryMonitor()` compares it against a run id.
- `normalizeIssueExecutionPolicy()` now carries `intervalSeconds` through. It builds the monitor
  field by field, so a new field is dropped unless it is added there.
- Fixed a misclassification on the same path in `server/src/services/recovery/issue-graph-liveness.ts`.
  Dispatching a monitor writes `executionState.monitor` onto issues that have no review stages at
  all. The classifier read any non-null `executionState` as an unresolvable review participant, so
  a monitor-parked issue with no action path was reported as `invalid_review_participant` instead of
  `in_review_without_action_path`. That pointed recovery at a review participant that never existed.
  The new `hasExecutionStageState()` helper looks for real stage state.
- Updated `doc/execution-semantics.md`: monitors are bounded recurring waits, `intervalSeconds` is
  documented, and the process-loss paragraph now describes when a future check can still be absent.
- The activity record `issue.monitor_triggered` now carries `rearmedNextCheckAt`, so the re-arm is
  visible in the issue timeline.
- Follow-up commit, from Greptile's review: the default attempt ceiling now falls back only when the
  policy names **neither** bound. The first commit applied it whenever `maxAttempts` was absent, so a
  monitor with a future `timeoutAt` and no `maxAttempts` was cleared as `max_attempts_exhausted`
  after 40 dispatches — about ten days at the default cadence — and ran recovery before its declared
  deadline. Both bound checks are corrected, with a test that a timeout-bounded monitor past the
  ceiling still dispatches.

## Verification

Rebased on 2026-08-25: the head now sits on current `master` (`42b8f7ab2`), replacing the
earlier revision that was based on `4eace88f6`. Both commits cherry-picked cleanly; comparing the
old and the new patch with `diff` over `git diff <base>..<head>` (ignoring `index` lines) leaves
only hunk-header offsets plus four context lines in `packages/shared/src/index.ts`, where `master`
added neighbouring exports. The change itself is unmodified.

The earlier note said this fork could not take newer `master` commits. That was too broad: the
`workflow`-scope check applies to the commits a push actually transfers. A fork shares its object
store with the upstream, so pushing a feature branch on a current `master` base transfers only the
two commits here — neither touches `.github/workflows/*` — and it goes through. What still needs
the scope is moving the fork's own default branch.

```
npx tsc --noEmit -p server/tsconfig.json            # clean
npx vitest run src/__tests__/issue-monitor-scheduler.test.ts          # 10/10
npx vitest run src/__tests__/issue-execution-policy.test.ts           # 75/75
npx vitest run src/__tests__/recovery-classifiers.test.ts             # 7/7
npx vitest run src/__tests__/issue-monitor-scheduler.test.ts \
  src/__tests__/issue-execution-policy.test.ts \
  src/__tests__/recovery-classifiers.test.ts                          # 93/93 after the Greptile fix
npx vitest run src/__tests__/heartbeat-process-recovery.test.ts       # 97/97
npx vitest run src/__tests__/issue-execution-policy-routes.test.ts \
  src/__tests__/issue-liveness.test.ts \
  src/__tests__/task-watchdogs-scheduler.test.ts \
  src/__tests__/heartbeat-archived-company-guard.test.ts              # 100/100
npx vitest run src/__tests__/heartbeat-process-recovery.test.ts \
  src/__tests__/heartbeat-issue-liveness-escalation.test.ts           # 119/119
```

Re-verified on the rebased head (`719606759`), against `master` `42b8f7ab2`:

```
npx vitest run src/__tests__/issue-monitor-scheduler.test.ts \
  src/__tests__/issue-execution-policy.test.ts \
  src/__tests__/recovery-classifiers.test.ts                          # 93/93
```

The server typecheck was run twice under identical conditions, once on the rebased head and once on its base commit, and both produce the same error set (the local checkout's installed dependency tree is older than `master`, so both runs share the same pre-existing failures). Diffing the two outputs leaves only line-number offsets, so the change introduces no new type error. The authoritative typecheck is this PR's CI, which installs against the live base.

New test for the reported failure, in `issue-monitor-scheduler.test.ts`: *"keeps a wake path when
the woken run fails terminally"*. The monitor fires, the woken run is set to `failed` with
`claude_transient_upstream`, and the issue still carries `monitorNextCheckAt` at
`lastTriggeredAt + 6h`. The next tick wakes the assignee again.

The classifier fix is covered by *"reports a spent monitor as in_review_without_action_path, not as
a broken review participant"*. That test was confirmed to fail before the fix, with
`expected 'invalid_review_participant' to be 'in_review_without_action_path'`.

Two existing tests pinned the one-shot behaviour and were rewritten against the new contract:
*"triggers due issue monitors once and clears the one-shot schedule"* and *"lets the board trigger a
scheduled issue monitor immediately"*.

One honest note on flakiness. The first combined run of
`heartbeat-process-recovery` + `heartbeat-issue-liveness-escalation` on this branch failed once in
`afterEach` cleanup, not in an assertion: `delete from "agents"` hit
`agent_wakeup_requests_agent_id_agents_id_fk`. The same two-file combination on clean `master` was
green (119/119), and a second run of the same combination on this branch was also green (119/119),
as was that suite alone (97/97). It reproduced neither on this branch nor on master a second time,
so it reads as a teardown race in the shared embedded-Postgres fixture rather than a regression. It
is reported here rather than dropped.

## Risks

- Medium. This changes wake timing for every scheduled monitor. Monitors that previously went quiet
  after one fire now keep waking their assignee until a bound is reached. That is the intent, but on
  a board with long-parked monitored issues it will produce wakes that did not happen before.
- The default attempt ceiling of 40 is a new terminal condition for monitors that declared no bounds.
  Such a monitor now ends after 40 attempts through `recoveryPolicy` instead of running unbounded.
  With the 6 hour default that is about ten days.
- `applyMonitorTransition()` and `buildInitialIssueMonitorFields()` share the same ceiling, so an
  attempt to schedule or re-arm a monitor on an issue already past 40 attempts is now rejected with
  the existing "Monitor bounds are already exhausted" error rather than accepted.
- No migration. `intervalSeconds` lives inside the existing `execution_policy` JSON column and is
  optional; policies without it get the default.
- The `externalRef` behaviour is a fix, not a change of contract: redaction on read stays, and the
  stored value is now preserved across a re-arm instead of being overwritten.

## Model Used

- Anthropic Claude, Opus 5 (`claude-opus-5`), 1M context window, extended thinking, with tool use and
  terminal execution.

## Checklist


juancarlosrial76-code:
Thanks — the timeout-only finding is correct, and it contradicted this PR's own documented contract.

`DEFAULT_ISSUE_MONITOR_MAX_ATTEMPTS` is documented to apply only when a policy names no bound at all, but both checks fell back to it whenever `maxAttempts` was absent. A monitor with `timeoutAt` weeks out and no `maxAttempts` would therefore be cleared as `max_attempts_exhausted` after 40 dispatches, about ten days at the 6 hour default, and run recovery before its deadline.

Fixed in ee12c0252 in both places (`issueMonitorLimitClearReason` in `heartbeat.ts` and `exhaustedMonitorClearReason` in `issue-execution-policy.ts`): the ceiling now applies only when neither bound is present. Added a test that a timeout-bounded monitor past the ceiling still dispatches and is not cleared.

PR #10736, opened 2026-08-03
fix(db): sweep crash-orphaned backup files, fix retention accounting, and size-check backup health

nguyenm7:
## Thinking Path

> - The server protects instance data with scheduled logical database backups (`packages/db/src/backup-lib.ts`) and reports their freshness through `/api/health`
> - A process kill mid-backup leaves a truncated working `.sql` file and a header-only `.sql.gz` on disk forever, because cleanup only runs in the in-process catch block
> - Retention pruning counted bare `.sql` working files as backups, so a crash orphan with the newest mtime could be retained while the last complete `.sql.gz` of that week was pruned
> - The health check accepted a fresh-but-empty `.sql.gz` as the latest backup, because it only verified age, not size — health stayed green while restores were impossible
> - This pull request adds a crash-orphan sweep before each backup run, restricts retention accounting to completed `.sql.gz` files, and adds a minimum-plausible-size health check
> - The benefit is that a killed backup process can no longer silently destroy backup integrity or hide behind a green health status

**What happened**

A server process was terminated mid-backup. It left a 617 MB truncated working `.sql` file and a 20-byte header-only `.sql.gz` in the backup directory. Nothing ever removed them. The health endpoint reported the 20-byte gz as `latestBackup` with status ok because it only checks age. Retention pruning also treated the bare `.sql` as a backup, so it could displace the last good compressed backup of its weekly bucket.

**Expected behavior**

Crash leftovers are cleaned up automatically on the next run. Only completed `.sql.gz` files participate in retention. A latest backup that is implausibly small degrades health to `warning` with a distinct code so operators notice.

**Steps to reproduce**

1. Start a database backup and `kill -TERM` the process while the working `.sql` file is being written.
2. Observe the orphaned `paperclip-*.sql` and tiny `paperclip-*.sql.gz` files remain in the backup directory indefinitely.
3. Call `GET /api/health` and observe `databaseBackup.status` is `ok` with the tiny gz as `latestBackup`.

**Version or commit**

master at 2c90cf0f2.

**Deployment mode**

Self-hosted local server with scheduled automatic backups enabled.

## What Changed

- `runDatabaseBackup` now calls a new exported `sweepBackupOrphans` before each run. It deletes bare `${prefix}-*.sql` working files and `.sql.gz` files under 1 KB. A one-hour age guard protects files a concurrently running backup may still be writing. Swept paths are returned in the result (`sweptOrphans`), logged by the server, and included in `formatDatabaseBackupResult`.
- `pruneOldBackups` now only counts `.sql.gz` files. Bare `.sql` working files are excluded from retention accounting entirely, so an orphan can no longer be kept in place of the newest real backup in a weekly or monthly bucket. The function is now exported for direct unit testing.
- `inspectDatabaseBackupHealth` adds a `database_backup_too_small` warning with two triggers: the latest backup is below an absolute floor (default 1 KB, configurable via `minSizeBytes` / `PAPERCLIP_DB_BACKUP_MIN_SIZE_BYTES`), or it is under 10% of the median size of up to 8 prior backups (needs at least 3 priors). The new code is wired through the health route redaction map and the OpenAPI schema.
- Tests: new unit tests for the sweep and retention accounting in `packages/db/src/backup-lib.test.ts` (including an embedded-Postgres integration test of the sweep inside `runDatabaseBackup`), and new health tests for the absolute floor, the median collapse case, and a moderate-shrink ok case. Existing fixtures that wrote few-byte stand-in backups were padded to a plausible size, keeping healthy-directory behavior unchanged.

## Verification

- `npx vitest run packages/db/src/backup-lib.test.ts` — 10 passed; the pre-existing `restores fallback COPY data` test fails on hosts without a `psql` binary (identical failure on a clean checkout, passes in CI).
- `npx vitest run server/src/__tests__/health.test.ts server/src/__tests__/instance-database-backups-routes.test.ts` — 23 passed.
- `tsc --noEmit` clean in `packages/db`, `server`, and `cli`.
- Manual review point: a healthy backup directory (large gz files, no bare `.sql`) is untouched — the sweep deletes nothing, pruning selects the same files as before, and health emits no new warnings.

## Risks

- The sweep deletes files. It is scoped to the configured filename prefix, only bare `.sql` and sub-1 KB `.sql.gz`, and only files older than one hour, so it cannot race the run that is writing them. A legitimate backup of a nearly empty database under 1 KB would be swept after the fact and flagged by the new health warning rather than silently trusted.
- Backups smaller than 10% of the recent median now produce a health `warning`. A deliberate mass deletion of most instance data could trigger a false positive warning; it is warning-level only and self-clears as the median window moves.
- `RunDatabaseBackupResult` gains a `sweptOrphans` field. All in-repo consumers (server route, CLI JSON output, tests) are updated; the manual-backup API response is additive only.

## Model Used

- Claude Fable 5 (Anthropic, model id `claude-fable-5`), extended thinking enabled, agentic tool use (file edits, shell, test execution) via Claude Agent SDK.

## Checklist

- [x] I have searched GitHub for similar or duplicate PRs and found none to link
- [ ] I have updated relevant documentation to reflect my changes — *no user-facing docs describe backup internals; the new env var is documented in code and this PR*
- [ ] All Paperclip CI gates are green — *pending; CI runs after this PR opens*
- [ ] Greptile is 5/5 with no open P2s, recommendations, or follow-ups — *pending first Greptile review*

### DIFF
--- a/cli/src/commands/db-backup.ts
+++ b/cli/src/commands/db-backup.ts
@@ -85,6 +85,7 @@ export async function dbBackupCommand(opts: DbBackupOptions): Promise<void> {
             backupFile: result.backupFile,
             sizeBytes: result.sizeBytes,
             prunedCount: result.prunedCount,
+            sweptOrphans: result.sweptOrphans,
             backupDir,
             retentionDays,
             connectionSource: connection.source,
--- a/packages/db/src/backup-lib.test.ts
+++ b/packages/db/src/backup-lib.test.ts
@@ -4,7 +4,13 @@ import path from "node:path";
 import { gunzipSync } from "node:zlib";
 import { afterEach, describe, expect, it } from "vitest";
 import postgres from "postgres";
-import { createBufferedTextFileWriter, runDatabaseBackup, runDatabaseRestore } from "./backup-lib.js";
+import {
+  createBufferedTextFileWriter,
+  pruneOldBackups,
+  runDatabaseBackup,
+  runDatabaseRestore,
+  sweepBackupOrphans,
+} from "./backup-lib.js";
 import { ensurePostgresDatabase } from "./client.js";
 import {
   getEmbeddedPostgresTestSupport,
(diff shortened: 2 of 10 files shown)


nguyenm7:
Addressed both review findings in 730cfde33:

**Issue 1 (active backups can be swept):** `runDatabaseBackup` now writes a `${workingFile}.sql.pid` ownership marker for the duration of the run (removed in a `finally`), and `sweepBackupOrphans` skips any candidate — working `.sql` or in-progress `.sql.gz` — whose marker names a live process (EPERM counts as alive). This protects a stalled-but-alive run in another process (e.g. server scheduled backup vs an operator CLI backup in the same dir), which the mtime age gate alone could not: the pg_dump path streams straight into the `.sql.gz`, and the JS path reopens the `.sql` by path for compression, so either could previously fail after a cross-process unlink. Stray markers are themselves swept once their pid is dead. Covered by new sweep tests (live-pid spare, dead-pid sweep incl. marker cleanup) plus a marker-lifecycle assertion in the embedded-PG backup test.

**Issue 2 (median mixes backup streams):** the relative size check now derives a stream key from the latest backup's filename (`paperclip-20260803-071500.sql.gz` → `paperclip`) and samples the median only from prior backups of that same stream, matching how sweep and retention scope by prefix. Non-conforming names never share a stream, so they fall back to the absolute floor only. Two new health tests cover a foreign 10 MB stream not flagging a healthy 60 KB stream, and tiny foreign files not masking a real collapse.

Single-prefix directories (the normal case) behave exactly as before. `packages/db` backup tests and server health tests pass locally; both packages typecheck clean.

@greptileai review

PR #10748, opened 2026-08-03
fix(issues): release the execution lock held by a never-started scheduled_retry

goldinbidsheets-rgb:
## The defect

The retry scheduler stamps `issues.executionRunId` when it **creates** the `scheduled_retry` heartbeat run, not when that run starts. After a provider-quota backoff the retry sits dormant (`startedAt: null`) until `scheduledRetryAt` -- hours later -- so the issue holds an execution lock owned by a run that is not executing anything.

`clearExecutionRunIfTerminal` / `clearCheckoutRunIfTerminal` only release on `TERMINAL_HEARTBEAT_RUN_STATUSES`, and `scheduled_retry` is correctly not terminal. So the lock never clears, and `canAdoptUnownedCheckout` refuses adoption on its `executionRunId` conjunct.

## Observed impact

The assignee's own live run gets `409 Issue run ownership conflict` on every `PATCH /issues/{id}` and `POST /issues/{id}/interactions`, while comments still return 201.

That makes the execution contract unsatisfiable: `done`, `in_review` and `blocked` are all PATCHes, so the issue is pinned to `in_progress` -- which is exactly the status that gets re-woken. The fleet re-enters finished work and re-derives conclusions already posted. Approval routing is blocked too: the `request_confirmation` interaction cannot be created, so completed work that ends in a board decision cannot be handed over at all.

Observed on a 10-issue burst across two agents in a single quota window. The issues read `activeRun: null` the whole time, which is what makes it look like a dead-run leak rather than a scheduling artifact.

## The change

Introduces `heartbeatRunHoldsIssueClaim()` as the single "does this run hold a real claim on the issue" predicate, and routes both clear helpers through it. A run holds nothing if it is terminal, **or** if it is `scheduled_retry` with `startedAt == null`. A `scheduled_retry` run that *has* started still holds its claim.

No new call sites were needed: `svc.update` and the checkout path already call both clear helpers before resolving ownership, so once the dormant lock clears, the existing `canAdoptUnownedCheckout` adopts it.

## Why this does not orphan the retry

- The retry is located by `contextSnapshot->>'issueId'` (`getIssueRetryRun`, `server/src/services/heartbeat.ts`), **not** by `issues.executionRunId`.
- `cancelStaleScheduledRetry` (`heartbeat.ts`) already nulls that exact column when a dormant retry stops being valid -- the codebase already treats this lock as releasable.

## Tests

Two cases added to the existing `server/src/__tests__/issue-stale-execution-lock-routes.test.ts`, modeled on the passing cases already in that file:

1. the assignee's live run adopts an execution lock held by a never-started `scheduled_retry`;
2. a `scheduled_retry` that **has** started still returns 409 (the guard is not over-broad).

Refs GOLAA-6375.

### DIFF
--- a/server/src/__tests__/issue-stale-execution-lock-routes.test.ts
+++ b/server/src/__tests__/issue-stale-execution-lock-routes.test.ts
@@ -310,6 +310,95 @@ describeEmbeddedPostgres("stale issue execution lock routes", () => {
     });
   });
 
+  // GOLAA-6375: a provider-quota backoff creates a `scheduled_retry` run and
+  // stamps it onto issues.executionRunId immediately, so the lock is held for
+  // the whole wait before that row ever starts. The assignee's live run then
+  // cannot PATCH a disposition or POST an approval interaction.
+  it("lets the assignee's live run adopt an execution lock held by a never-started scheduled_retry", async () => {
+    const { companyId, agentId, currentRunId } = await seedCompanyAgentAndRuns();
+    const dormantRetryRunId = randomUUID();
+    const issueId = randomUUID();
+    await db.insert(heartbeatRuns).values({
+      id: dormantRetryRunId,
+      companyId,
+      agentId,
+      status: "scheduled_retry",
+      invocationSource: "assignment",
+      startedAt: null,
+      scheduledRetryAt: new Date(Date.now() + 60 * 60 * 1000),
+      scheduledRetryReason: "provider_quota",
+    });
+    await db.insert(issues).values({
+      id: issueId,
+      companyId,
+      title: "Dormant retry execution lock",
+      status: "in_progress",
+      priority: "high",
+      assigneeAgentId: agentId,
+      checkoutRunId: null,
+      executionRunId: dormantRetryRunId,
+      executionAgentNameKey: "codexcoder",
+      executionLockedAt: new Date(),
+    });
+
+    const res = await request(createApp(agentActor(companyId, agentId, currentRunId)))
+      .patch(`/api/issues/${issueId}`)
+      .send({ title: "Recovered dormant retry lock" });
+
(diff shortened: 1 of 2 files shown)


goldinbidsheets-rgb:
## Local test run: all 13 pass

Ran `server/src/__tests__/issue-stale-execution-lock-routes.test.ts` against commits `570046514` (fix) + `f0bbefbd4` (test timeout) on this host (Windows, embedded Postgres).

```
 PASS  allows an assigned agent PATCH to recover a terminal stale executionRunId  1402ms
 PASS  preserves 'done' when releasing a non-in_progress issue  310ms
 PASS  preserves 'cancelled' when releasing a non-in_progress issue  232ms
 PASS  preserves 'in_review' when releasing a non-in_progress issue  181ms
 PASS  preserves 'blocked' when releasing a non-in_progress issue  292ms
 PASS  allows the rightful assignee to release after the owning run failed  417ms
 PASS  lets the current assignee recover a timed_out stale checkout owner during PATCH  459ms
 PASS  lets the assignees live run adopt an execution lock held by a never-started scheduled_retry  276ms
 PASS  still returns 409 when a scheduled_retry execution owner has actually started  200ms
 PASS  still returns 409 when a different live checkout owner is active  201ms
 PASS  preserves live checkout ownership on checkout conflicts without retry side effects  274ms
 PASS  restricts admin force-release to board users with company access and writes an audit event  158ms
 PASS  self-heals a stale checkoutRunId via clearCheckoutRunIfTerminal on checkout (Fix B path)  335ms

Test Files  1 passed (1)
     Tests  13 passed (13)
```

The two new regression cases (rows 8 and 9 above) covering this fix are confirmed green. Also bumped `beforeAll` timeout from 20s to 120s in the second commit -- 20s was too tight on hosts where module import alone takes ~160s.

PR #10751, opened 2026-08-03
Don't give hosted users credential advice they cannot act on

stubbi:
## Problem

The adapter environment probe assumes the Paperclip host is the user's own machine. On a multi-tenant deployment that assumption fails in two distinct ways.

**1. Unfollowable advice.** A missing Codex credential is answered with:

> Set OPENAI_API_KEY in the agent adapter env, set it in the Paperclip server environment, or run `codex login` for the same OS user that runs the Paperclip server

A tenant has no shell on that server. All three routes are dead ends, so a missing key reads as a broken product. The Claude lanes have the same shape ("subscription-based auth can be used if Claude is logged in").

**2. Cross-tenant credential reporting.** This is the sharper one. `evaluateCodexCredentialReadiness` falls back to `codexHomeHasUsableAuth(sharedSourceHome)` — the *server's* Codex home. Where the operator is logged in, every tenant is told `codex_acp_native_auth_detected`, i.e. "you are authenticated", by someone else's credential. They then watch the agent fail on its first real run with no clue why.

## Motivation

A paying customer hit exactly this. Telemetry shows him running the probe, receiving the `codex login` hint, clicking it repeatedly, then stopping. He wrote in asking for a refund because he felt misled. He had never stored a credential or completed a run.

He was also right on the substance: a ChatGPT plan genuinely cannot work on a hosted install, because it authenticates only through that local login. Nothing said so.

## Change

`callerControlsHost` on `AdapterEnvironmentTestContext`, **defaulting to `true`** so every existing caller keeps today's behaviour byte for byte. The agents route derives it from `req.actor.source`, which already distinguishes `cloud_tenant` from a local operator.

When `false`:
- codex lanes name the credential the user can actually supply, and state that a ChatGPT Plus/Pro plan cannot be used here. A bare "set OPENAI_API_KEY" leaves a subscriber hunting for the plan option instead of buying the credit the key needs.
- claude lanes point at `claude setup-token`, which genuinely does let a Pro/Max plan drive the agent and is minted on the user's own machine.
- the branch is ordered **directly after the api-key case**, so no host-derived readiness signal can reach a tenant. The agent's own configured key still works exactly as before.

## Tests

New `acp.hosted-credentials.test.ts`. It stubs `CODEX_HOME`/`HOME` rather than reading whatever the machine has, so it describes the code instead of the developer's laptop (the first draft passed on CI and failed locally for precisely that reason). Pins all four directions: hosted gets no login imperative, hosted never sees a seeded host credential, the agent's own key still registers, and self-hosted keeps both the host credential reporting and the `codex login` hint.

codex-local: 22 files / 223 pass. Two failures in claude-local (`execute.remote`, `test.probe`) reproduce unmodified on `origin/master` and are unrelated to this change.

### DIFF
--- a/packages/adapter-utils/src/types.ts
+++ b/packages/adapter-utils/src/types.ts
@@ -318,6 +318,25 @@ export interface AdapterEnvironmentTestContext {
    * Surfaced in check messages so users see which environment the probe ran in.
    */
   environmentName?: string | null;
+  /**
+   * Whether the person reading this probe's result can reach the Paperclip
+   * host's shell and filesystem.
+   *
+   * Defaults to `true`, which is the single-operator install every adapter was
+   * written for: the host is the user's own machine, so "run `codex login`" or
+   * "run `claude login`" is advice they can act on, and a credential file
+   * sitting in `~/.codex` is theirs.
+   *
+   * Set to `false` for a hosted multi-tenant deployment, where BOTH halves of
+   * that assumption are wrong. The user has no shell on the host, so host-login
+   * advice is unfollowable and reads as a broken product; and any credential
+   * that does sit on the host belongs to the operator or another tenant, so
+   * reporting it as "you are authenticated" would be actively misleading.
+   * Adapters must then confine themselves to credentials the user can supply
+   * through Paperclip: an API key, or a subscription token they mint on their
+   * own machine and paste in.
+   */
+  callerControlsHost?: boolean;
   deployment?: {
     mode?: "local_trusted" | "authenticated";
     exposure?: "private" | "public";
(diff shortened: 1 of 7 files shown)


stubbi:
Rebased onto current master `d293dd3d1` (new head `9fd145b96`). Four conflicts: two were local-variable declarations colliding with master's new `targetIsSandbox` (both kept), and two were in the Claude credential `else if` chains, where master added a `CLAUDE_CODE_OAUTH_TOKEN is set` branch at the same position as this PR's hosted branch — resolved by ordering master's detection branch first so a configured token is still reported as configured, with the `!callerControlsHost` advice branch falling in behind it and ahead of the local `!targetIsRemote` case. Verified: `@paperclipai/adapter-utils`, `@paperclipai/adapter-codex-local`, `@paperclipai/adapter-claude-local` and `@paperclipai/server` typecheck clean; codex server tests (49/49 across `acp.hosted-credentials`, `acp`, `test.remote`, `auth-precedence`) and claude server tests (73/73 across `acp`, `acp.auth`, `test.probe`, `test.remote`, `probe-env`) pass.

PR #10752, opened 2026-08-03

AaryanCdry:
## Summary

```ts
if (!gitErrorIncludes(attachError, "already checked out") && !gitErrorIncludes(attachError, "already used by worktree")) {
```

```ts
const mergedEnv = opts?.env ? { ...process.env, ...opts.env, LC_ALL: 'C', LANG: 'C', LANGUAGE: 'C' } : { ...process.env, LC_ALL: 'C', LANG: 'C', LANGUAGE: 'C' };
```
Note: C locale is set LAST so it cannot be overwritten by caller-provided env vars.

## Thinking Path

Paperclip uses git worktrees to isolate each agent task's codebase. The `realizeExecutionWorkspace` function in `server/src/services/workspace-runtime.ts` creates or reuses worktrees via `git worktree add`. When a branch is already checked out elsewhere, git emits an error that Paperclip intercepts and tries to handle by reusing the existing worktree instead of creating a duplicate.

**Issue #10683:** The existing guard checked for `"already checked out"` — the message git 2.41 and earlier emitted. Git 2.42 changed this to `"already used by worktree at"`. On newer git versions, the guard failed to match, causing the code to re-throw the error and lose the worktree-reuse optimization. Fix: match both messages with an `OR` condition.

**Issue #10681:** `runGit` (and two other git call sites: `recordGitOperation` and `getGitWorktreeBranchAncestryVerdict`) passed the host environment to `spawn()`. On non-English systems, git translates its error messages. Paperclip's `gitErrorIncludes` helper matches English substrings ("already checked out", "already exists", etc.). Translated messages break these matches, causing spurious failures. Fix: inject `LC_ALL=C` into git's environment so diagnostics are always in English. C locale is placed LAST in the env merge so it cannot be overwritten by caller-provided values.

Both fixes are defensive — they expand match coverage without changing behavior for correctly-matching cases.

## What Changed

- `server/src/services/workspace-runtime.ts` line 581-585: `runGit` now forces `LC_ALL=C`, `LANG=C`, `LANGUAGE=C` with locale overrides last in merge order
- `server/src/services/workspace-runtime.ts` line 2943: `gitErrorIncludes` guard now matches both `"already checked out"` AND `"already used by worktree"`
- `server/src/services/workspace-runtime.ts` line 2527-2530: `recordGitOperation` also forces C locale
- `server/src/services/workspace-runtime.ts` line 981: `getGitWorktreeBranchAncestryVerdict` also forces C locale
- Added test in `workspace-runtime.test.ts` for the git 2.42+ worktree reuse path

## Verification

- Manual reproduction: confirmed git 2.42+ message format ("already used by worktree at")
- Manual reproduction: confirmed old git message format ("already checked out at")
- Logic analysis: the new `&&` condition correctly re-throws only when NEITHER pattern matches
- Locale test: `git worktree add` produces identical English output under `LC_ALL=C` vs default locale
- All git call sites in the file now use the C locale
- New test added covering the git 2.42+ error message path

## Risks

- **Very low.** Both changes are additive and defensive.
- The locale fix sets environment variables with C locale last, so caller-provided `opts.env` values are preserved except for locale (which is intentionally overridden).
- The error pattern match adds a second substring check — false positives would only cause the worktree-reuse path to activate when it previously would have thrown, which is the safer direction.

## Model Used

Claude Sonnet 4 (custom:omniroute, 200K context, tool use)

## Checklist

- [x] I have read and followed the [Contributing Guide](CONTRIBUTING.md)
- [x] Changes are small, focused, and easy to review
- [x] Tests pass (new test covers git 2.42+ worktree reuse path)
- [x] No breaking changes
- [x] PR template sections filled
- [x] Model Used section included

### DIFF
--- a/server/src/__tests__/workspace-runtime.test.ts
+++ b/server/src/__tests__/workspace-runtime.test.ts
@@ -1042,6 +1042,30 @@ describe("realizeExecutionWorkspace", () => {
     });
   });
 
+  it("reuses a worktree when git emits the 2.42+ 'already used by worktree' message", async () => {
+    const repoRoot = await createTempRepo();
+    const branchName = "PAP-10683-worktree-new-git-msg";
+    const existingWorktree = path.join(repoRoot, ".paperclip", "worktrees", branchName);
+    const { recorder, operations } = createWorkspaceOperationRecorderDouble();
+    await fs.mkdir(path.dirname(existingWorktree), { recursive: true });
+    await execFileAsync("git", ["worktree", "add", "-b", branchName, existingWorktree, "HEAD"], { cwd: repoRoot });
+    const altWorktreePath = path.join(repoRoot, ".paperclip", "other-worktrees", branchName);
+    await fs.mkdir(path.dirname(altWorktreePath), { recursive: true });
+    const realized = await realizeExecutionWorkspace({
+      base: { baseCwd: altWorktreePath, source: "project_primary", projectId: "project-1", workspaceId: "workspace-1", repoUrl: null, repoRef: "HEAD" },
+      config: { workspaceStrategy: { type: "git_worktree", branchTemplate: "{{issue.identifier}}-{{slug}}", worktreeParentDir: ".paperclip/other-worktrees" } },
+      issue: { id: "issue-1", identifier: "PAP-10683", title: "worktree reuse new git msg" },
+      agent: { id: "agent-1", name: "Codex Coder", companyId: "company-1" },
+      recorder,
+    });
+    const expectedWorktreePath = await fs.realpath(existingWorktree);
+    expect(realized.created).toBe(false);
+    await expect(fs.realpath(realized.cwd)).resolves.toBe(expectedWorktreePath);
+    await execFileAsync("git", ["worktree", "remove", existingWorktree], { cwd: repoRoot }).catch(() => {});
+    await execFileAsync("git", ["branch", "-D", branchName], { cwd: repoRoot }).catch(() => {});
+    await fs.rm(repoRoot, { recursive: true, force: true }).catch(() => {});
+  });
+
   it("slugifies unsafe issue titles for branch names and worktree folders", async () => {
     const repoRoot = await createTempRepo();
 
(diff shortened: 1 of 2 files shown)


matusbr:
Cross-checked this against patches we had prepared locally for the same failure
class. Good overlap on `runGit`, the `merge-base` call and `recordGitOperation` —
and forcing the C locale is the more robust half of this fix, since it removes
the dependency on matching translated Git output at all.

One gap: two call sites outside `server/` invoke Git directly and get neither
half of the fix.

- `packages/adapter-utils/src/git-workspace-sync.ts` — `integrateImportedGitHead`
  (`execFile("git", …)`) runs into `cannot lock ref` / `expected <oid>`.
- `packages/adapter-utils/src/ssh.ts` — the `git -C` wrapper.

Neither sets `LC_ALL`/`LANG`, so on a non-English host they see localized Git
output and any English needle silently fails to match. That failure is quiet:
the string simply never matches and the code takes the "no error" path.

Extending the locale forcing to those two would probably be worth more than
adding needles there.

matusbr:
Reporter of #10683 and #10681 here. I re-read this PR against current `master` today rather than against the state it was opened at. No patch offered — this PR is the one in flight; this is only what the measurement gives.

One thing it does better than my report: it adds `"already used by worktree"` without removing `"already checked out"`. Older git versions still emit the old wording, so keeping both is the safer reading.

What stays uncovered: the same error classification exists twice more, in `packages/adapter-utils/`, and this PR touches only `server/src/services/workspace-runtime.ts` and its test.

- `packages/adapter-utils/src/git-workspace-sync.ts:743` — `integrateImportedGitHead`
- `packages/adapter-utils/src/ssh.ts:927` — a verbatim copy of the same function

Both classify by English substrings:

```ts
return message.includes("cannot lock ref") && message.includes("expected");
```

Neither file forces a locale (no `LC_ALL` or `LANG` in either), so on a host with a localized git the predicate is false, the `attempt < 4` retry is skipped and the error is rethrown instead — the concurrent-ref-update retry silently degrades into a hard failure. Same class as #10681, just outside the file this PR fixes.

Not a request to widen this PR — it does its job where it sits. Flagging it so that closing #10681 on merge does not hide the two remaining call sites.

PR #9616, opened 2026-07-15
fix(issues): reject empty comment bodies and surface interaction decline reasons

Nabaatxaneh:
## Thinking Path

> - Paperclip is the open source app people use to manage AI agents for work
> - The issue thread is the shared audit log where humans and agents post comments and resolve interactions (confirmations, suggested tasks, questions)
> - Two governance gaps existed: whitespace-only comment bodies passed validation, and interaction decline/cancel reasons were buried in `result` JSON instead of being visible in issue history
> - A follow-up review then found a redaction regression: the new decline/cancel comment-insert paths wrote the user-typed reason directly, bypassing the `censorUsernameInLogs` scrubbing that `addComment` applies
> - This pull request closes all three gaps so blank comments are rejected, resolution reasons are surfaced as first-class comments, and those comments honor the username-redaction policy
> - The benefit is a trustworthy, policy-compliant issue history with no silent redaction bypass

## Linked Issues or Issue Description

No public GitHub issue — describing the underlying bug inline, following the bug report template.

### What happened

1. `z.string().min(1)` accepts `"   "` and escaped-newline-only input, so blank comments could be posted via the API and the board-chat concierge.
2. When a `request_confirmation` was declined, `suggest_tasks` rejected, or `ask_user_questions` cancelled with a reason, the reason landed only in the interaction `result` JSON and was invisible in the thread.
3. The new decline/cancel comment inserts wrote the user-typed reason straight into `issueComments`, skipping the `redactCurrentUserText` step `addComment` performs. With `censorUsernameInLogs` enabled, a reason containing a username was stored verbatim — silently defeating the redaction policy.

### Expected behavior

Blank comments are rejected; interaction decline/cancel reasons are visible as first-class thread comments; and those surfaced reasons are redacted consistently with every other comment when `censorUsernameInLogs` is enabled.

### Steps to reproduce

1. Enable `censorUsernameInLogs` in instance settings.
2. POST a comment with body `"   "` (whitespace only) → it is accepted and stored (bug 1).
3. Decline a `request_confirmation` with a reason containing a username → the reason is stored verbatim, unredacted (bug 3), and prior to this change was invisible in the thread (bug 2).

### Paperclip version

master @ commit 31e1a66fb (branch `alaa-1954-govfix-707`).

### Deployment mode

Self-hosted / local instance (server + shared packages).

## What Changed

- `addIssueCommentSchema.body` and `updateIssueSchema.comment`: `z.string().min(1)` → `z.string().trim().min(1)`.
[...]

PR #10757, opened 2026-08-03
fix(server): surface interaction decline/cancel reasons as issue comments and reject whitespace-only comment bodies

Nabaatxaneh:
## Thinking Path

> - Boards and agents coordinate through issue-thread interactions: confirmations, suggested tasks, and questions
> - When a board declines a confirmation, rejects suggested tasks, or cancels questions, the reason text is stored only in the interaction's result JSONB
> - Nothing surfaces that reason on the issue thread, so the assignee's next wake context never sees why the request was declined - the guidance is silently lost
> - Separately, whitespace-only comment bodies pass validation (`min(1)` without trim) and create empty comments that wake agents with no content
> - This PR records decline/cancel reasons as first-class issue comments at resolution time, and rejects whitespace-only comment bodies at both the schema and service layers
> - The benefit is that resolution guidance reliably reaches the assignee's context, and empty-content comment wakes disappear

## What Changed

- `server/src/services/issue-thread-interactions.ts`: new `recordResolutionReasonComment` helper, called from `rejectRequestConfirmation` (covers `request_confirmation` and `request_checkbox_confirmation`), `rejectSuggestedTasks`, and `cancelQuestions` when a non-empty reason is present. Best-effort: a failed comment insert logs a warning and never fails the already-committed resolution.
- `packages/shared/src/validators/issue.ts`: `addIssueCommentSchema.body` and `updateIssueSchema.comment` now `.trim()` before `.min(1)`, so whitespace-only submissions are rejected.
- `server/src/services/issues.ts`: `addComment` throws 422 `Comment body cannot be empty` for whitespace-only bodies - defense in depth for internal callers that bypass route validation.
- `server/src/routes/board-chat.ts`: skip persisting a board reply whose cleaned response is whitespace-only.
- `server/src/__tests__/issue-thread-interactions-service.test.ts`: new test for decline-reason comments (with a reason, and asserting no comment when the reason is omitted), new test for the whitespace guard, and the existing `cancelQuestions` test now asserts the cancellation-reason comment.

## Verification

- `pnpm --filter @paperclipai/shared typecheck && pnpm --filter @paperclipai/server typecheck` - clean
- `pnpm exec vitest run server/src/__tests__/issue-thread-interactions-service.test.ts` - 42 passed (embedded postgres)
- `pnpm exec vitest run packages/shared/src/validators/issue.test.ts server/src/__tests__/issue-comment-redaction.test.ts` - 32 passed
- Manual: decline a pending confirmation with a reason (board UI or the reject endpoint); a `**Confirmation declined** (interaction <id8>): <reason>` comment appears on the issue thread attributed to the declining user/agent. POST an issue comment whose body is only whitespace; the API returns 422.

## Risks

- Behavioral shift: resolutions that carry a reason now add a visible issue comment authored by the resolving user/agent. Resolution-time comments do not trigger comment-created side effects such as confirmation supersede-on-comment - those fire from the comment routes, not from `issueService.addComment`.
- Whitespace-only comments now fail with 422 where they previously created empty comments. No in-repo caller posts blank bodies through `addComment`; deleted-comment tombstones render `body: ""` through a different path and are unaffected (covered by the redaction suite, which passes).
- No migrations, no schema changes. Low risk overall; the reason comment is wrapped best-effort so a comment failure can never fail the resolution.

## Model Used

- Claude Fable 5 (Anthropic), model ID `claude-fable-5`, running in Claude Code - agentic session with extended thinking and tool use (repo search, file edits, local `tsc` and `vitest` runs against an embedded-postgres test database).

## Checklist

- [ ] If this change affects the UI, I have included before/after screenshots (n/a - no UI change)
- [ ] I have updated relevant documentation to reflect my changes (n/a - no existing doc surface describes comment validation or interaction resolution side effects)

### DIFF
--- a/packages/shared/src/validators/issue.ts
+++ b/packages/shared/src/validators/issue.ts
@@ -539,7 +539,7 @@ export const updateIssueSchema = createIssueBaseSchema.omit({
 }).partial().extend({
   requestDepth: issueRequestDepthInputSchema.optional(),
   assigneeAgentId: z.string().trim().min(1).optional().nullable(),
-  comment: multilineTextSchema.pipe(z.string().min(1)).optional(),
+  comment: multilineTextSchema.pipe(z.string().trim().min(1)).optional(),
   reviewRequest: issueReviewRequestSchema.optional().nullable(),
   reopen: z.boolean().optional(),
   resume: z.boolean().optional(),
@@ -644,7 +644,7 @@ export const issueCommentMetadataSchema = z.object({
 export type IssueCommentMetadata = z.infer<typeof issueCommentMetadataSchema>;
 
 export const addIssueCommentSchema = z.object({
-  body: multilineTextSchema.pipe(z.string().min(1)),
+  body: multilineTextSchema.pipe(z.string().trim().min(1)),
   authorType: issueCommentAuthorTypeSchema.optional(),
   presentation: issueCommentPresentationSchema.nullable().optional(),
   metadata: issueCommentMetadataSchema.nullable().optional(),
--- a/server/src/__tests__/issue-thread-interactions-service.test.ts
+++ b/server/src/__tests__/issue-thread-interactions-service.test.ts
@@ -565,6 +565,15 @@ describeEmbeddedPostgres("issueThreadInteractionService", () => {
       summaryMarkdown: null,
     });
 
+    const comments = await db
+      .select()
+      .from(issueComments)
+      .where(eq(issueComments.issueId, issueId));
+    const reasonComment = comments.find((comment) => comment.body.includes("Questions cancelled"));
+    expect(reasonComment).toBeDefined();
+    expect(reasonComment?.body).toContain("Not needed anymore");
+    expect(reasonComment?.authorType).toBe("user");
+
     await expect(interactionsSvc.answerQuestions({
       id: issueId,
       companyId,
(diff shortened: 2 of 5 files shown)


Nabaatxaneh:
Rebased onto current master (`165b10bd9`) and force-pushed. The branch had drifted ~1030 commits behind and GitHub had it as `CONFLICTING`, so there was nothing a reviewer could have merged even if they wanted to. It is a single commit and `MERGEABLE` again now.

Also closed #9616, which was a second PR from this fork carrying the same fix. Two competing diffs over the same four files was our mistake; this is the only one to look at.

Four conflicts came out of the rebase, all from upstream refactors rather than from any change of intent on our side:

- `services/issues.ts` — `addComment` became a named function with a serialize-transaction wrapper. The empty-body guard is now the first statement, so a whitespace-only body is rejected before a transaction is opened.
- `services/issue-thread-interactions.ts` — the confirmation-decline path moved its resolved-check and `touchIssue` inside the transaction. `recordResolutionReasonComment` stays *after* the commit: the resolution must not roll back because surfacing the reason failed (the helper already swallows and logs its own errors). The other two call sites rebased cleanly.
- `validators/issue.ts` — kept upstream's new `clientRequestId` / `attachmentIds` / `onBehalfOfUserId` fields; the only change is `.trim()` on the `body` and `comment` pipes.
- `__tests__/issue-thread-interactions-service.test.ts` — the file now exists upstream, so both sides' tests are kept.

`tsc --noEmit` is clean on all five changed files (`@paperclipai/shared` typechecks green end to end).

Re-verified the motivation against the published `2026.916.0` build today — all four behaviours are still live there:

1. `addIssueCommentSchema.body` is `multilineTextSchema.pipe(z.string().min(1))`, and `multilineTextSchema` only normalizes escaped line breaks, so `"   "` still validates;
2. the three interaction reject/cancel paths go straight to `hydrateInteraction(...)`, so a decline or cancel reason never reaches the thread;
3. `routes/board-chat.ts` guards with `if (cleanedResponse)` and no `.trim()`;
4. `services/issues.ts` `addComment` has no empty-body rejection at the service entry.

We run a fleet of agents against this platform and (2) is the one that actually costs us: an agent whose confirmation is declined sees the interaction flip to `rejected` with the reason visible nowhere in the thread it reads, so it re-proposes the same thing. Happy to split this into smaller PRs, or to drop it entirely if the direction is wrong — a "no" is a useful answer too.

Issue #7733, opened 2026-06-07
claude-local adapter never rotates session on "Prompt is too long" → agent stuck in a permanent failure loop


liorsolomon:
### Summary
When a resumed `claude-local` session grows past the model's context window, the Claude CLI exits non-zero with **"Prompt is too long."** The adapter does not recognize this as a session-fatal condition, so it preserves the same `sessionId` and the heartbeat resumes the *same* oversized transcript on every subsequent run. The agent gets wedged in a permanent loop of `prompt too long` failures with no automatic recovery.

### Where
`packages/adapters/claude-local/src/server/execute.ts` (on `master` @ `823c2b11`)
- `clearSession` is only set for **max-turns** (`isClaudeMaxTurnsResult`, ~L881/L940) and **unknown/missing session** (`isClaudeUnknownSessionError`, ~L951).
- There is no classifier for context-overflow.

`packages/adapters/claude-local/src/server/parse.ts`
- `CLAUDE_TRANSIENT_UPSTREAM_RE` (L12) matches rate-limit / overload / 429 / 503 / 529 / usage-cap, but **not** "prompt is too long" / context-length-exceeded.
- Even if it did, `isClaudeTransientUpstreamError` (L370) drives a *retry*, not a session rotation — so it would loop rather than recover.

A repo-wide search (`server/**`, `packages/**`) finds no handling of `prompt too long` / `context_length_exceeded`, no auto-compaction, and no session-size guard.

### Steps to reproduce
1. Run a `claude-local` agent that resumes one long-lived session across many heartbeats (the default).
2. Have it accumulate large tool results in the transcript (e.g. an analyst agent writing big JSON documents that come back as tool outputs).
3. Once the transcript exceeds the context window, every heartbeat fails with "Prompt is too long" and the agent never recovers on its own.

### Expected
A context-overflow failure should be treated like a session-fatal error: rotate to a fresh session (set `clearSession: true`) so the next run starts clean, rather than re-resuming the oversized transcript indefinitely.

### Proposed fix
Two viable directions (not mutually exclusive):

**A — Rotate on overflow (minimal).** Add an `isClaudePromptTooLongResult(parsed)` classifier in `parse.ts` (matching e.g. `prompt is too long`, `context length`, `context window`, `maximum.*tokens`) and OR it into the `clearSession` decision in `execute.ts`, alongside the existing max-turns case. Simple, recovers immediately, but loses conversational continuity on overflow.

[...]

PR #10760, opened 2026-08-03
fix(claude-local): classify prompt overflow as deterministic, not transient upstream

juancarlosrial76-code:
## Thinking Path

> - A `claude_local` agent that outgrows its context window fails with `Prompt is too long`.
> - That failure is deterministic: the same prompt is the same size on the next attempt.
> - The adapter had no classifier for it, so the run was retried as a transient upstream blip.
> - Measured on one host, those retries burned 617 runs without a chance of succeeding.
> - Giving the failure its own deterministic error code keeps it off the transient retry path.
> - The benefit is that an overflowing agent stops spending its attempt budget on a retry that cannot work.

Refs #7733.

## What Changed

- Add `isClaudeContextOverflowError()` to the claude-local parser. Like the other
  deterministic classifiers in that file it reads only the parsed result payload,
  never `stdout`.
- Decide it **before** the transient branch in the parsed classification ladder,
  and add it to the deterministic guard list inside
  `isClaudeTransientUpstreamError()` and `isClaudeProviderQuotaError()`.
- Emit error code `claude_context_overflow` and error family `context_overflow`,
  and add that family to `AdapterExecutionErrorFamily`.

## Why the ordering is the actual fix

The obvious reading is that `Prompt is too long` falls into a generic transient
bucket because nothing classifies it. That is not what happens —
`CLAUDE_TRANSIENT_UPSTREAM_RE` does not match the text:

```js
RE.test("Claude run failed: subtype=success: Prompt is too long")  // false
```

`buildClaudeTransientHaystack()` folds the whole `stdout` into the search text,
and every Claude CLI stream carries its own rate-limit telemetry:

```json
"rate_limit_info":{"status":"rejected","resetsAt":...,"rateLimitType":"seven_day"}
```

`rateLimitType` alone matches `rate[-\s]?limit`. Sampled over the real run logs
on one host: **40 of 40** carry that token. Feeding a real 60 kB `stdout` to the
shipped predicate (`paperclipai@2026.609.0`, `dist/server/parse.js`):

| result text | `isClaudeTransientUpstreamError` |
| --- | --- |
| `Prompt is too long` | true |
| `disk full` | true |
| `syntax error in config` | true |

So a new error class on its own would still lose to the contaminated haystack.
It has to be decided first — which is what this change does.

That wider contamination affects **every** error class, not just this one, so it
is deliberately left alone here and belongs in its own change.

## Why the retry actually stops

`readTransientRecoveryContractFromRun()` builds a retry contract only for the
families `transient_upstream` and `provider_quota`. `context_overflow` is
neither, so no `transient_failure` retry is scheduled.

## Verification

- `npx vitest run packages/adapters/claude-local/src/server/parse.test.ts`
  → **43 passed** (4 new).
- Whole package before/after, same tree, same links:
  baseline **71 passed / 9 failed (80)**, with this change **75 passed / 9 failed (84)**.
  The 9 failures are pre-existing in this environment (`Cannot find package 'acpx/runtime'`
  and a remote-exec call-count assertion); the delta is exactly the 4 new tests.
- `npx tsc --noEmit` in `packages/adapters/claude-local`: no diagnostics in the
  changed files. (This is what caught the missing `AdapterExecutionErrorFamily`
  member.) The remaining output is pre-existing `@types/node` resolution noise
  in this sandbox.
- Red-before is measured against the **shipped** artifact, not a hand-rolled
  stub: the table above is the old behaviour.

## Note on scope

The overflow classifier reads the parsed payload only, so a run that dies
without emitting a parseable result JSON is not covered. In every failure
measured here the CLI did emit one (`subtype=success`, `result="Prompt is too
long"`), and keeping deterministic classifiers off `stdout` is exactly what
stops the contamination described above.

### DIFF
--- a/packages/adapter-utils/src/types.ts
+++ b/packages/adapter-utils/src/types.ts
@@ -68,6 +68,9 @@ export interface AdapterRuntimeServiceReport {
 export type AdapterExecutionErrorFamily =
   | "transient_upstream"
   | "provider_quota"
+  // Deterministic: the prompt is the same size on the next attempt, so a plain
+  // retry cannot succeed. Deliberately not part of the transient contract.
+  | "context_overflow"
   | "model_refusal"
   | "refresh_token_reused"
   | "refresh_token_expired"
--- a/packages/adapters/claude-local/src/server/execute.ts
+++ b/packages/adapters/claude-local/src/server/execute.ts
@@ -69,6 +69,7 @@ import {
   isClaudePoisonedPreviousMessageIdError,
   isClaudeImageProcessingError,
   isClaudeModelNotFoundError,
+  isClaudeContextOverflowError,
 } from "./parse.js";
 import {
   materializeRemoteClaudeConfig,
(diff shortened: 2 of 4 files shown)


juancarlosrial76-code:
Closing in favour of #10761 — same commit and same tree, reopened from a branch whose name carries no internal ticket id (PR-template checklist). Apologies for the churn.

PR #10782, opened 2026-08-03
fix(adapters): stop benign CLI banners masking the real run error

stubbi:
## Thinking Path

> - Each `*-local` adapter shells out to an agent CLI and turns the exit into a run result the user reads in the UI
> - When the CLI exits non-zero without a structured error, the adapters fall back to `firstNonEmptyLine(stderr)` as the user-facing error
> - Line 1 of stderr is very often a startup banner the adapter itself provoked, so the banner is reported as the failure and the real cause never reaches anyone
> - This pull request adds a shared `firstMeaningfulStderrLine` that skips known-benign banners, and adopts it in the six adapters that derive a fallback error this way
> - The benefit is that a failed run reports the line that actually explains it, which is the difference between a user fixing their config and a user filing a support ticket

No existing issue. Describing it inline, following `.github/ISSUE_TEMPLATE/bug_report.yml`:

**What happened?**
Every failed `gemini_local` run reported its cause as `YOLO mode is enabled. All tool calls will be automatically approved.` That sentence is a startup banner the Gemini CLI prints because the adapter passes `--approval-mode yolo` itself. It is never a failure reason. The line that actually explained the failure was two lines lower in stderr and never reached a user.

**Expected behavior**
A failed run reports the stderr line that plausibly explains the failure, not a startup banner the adapter provoked.

**Steps to reproduce**
1. Configure a `gemini_local` agent so its run will fail at CLI startup (for example on a target with no container runtime and `adapterConfig.sandbox` set, which makes the CLI exit non-zero with a sandbox error).
2. Dispatch a run and let it fail.
3. Read the run's error in the UI or in `heartbeat_runs.error`.
4. Observe the error is the YOLO banner. The real cause is only visible in the full stderr.

`master` at 18f391ef0. Also present in every released version that carries these adapters.

**Deployment mode**
Reproduced on a self-hosted multi-tenant deployment; the code path is deployment independent.

**Installation method**
Docker image built from source.

**Agent adapter(s) involved**
`gemini_local` observed; `grok_local`, `opencode_local`, `cursor`, `pi_local` share the identical fallback. `codex_local` already carried a private filter.

**Database mode**
PostgreSQL.

**Access context**
Server-side adapter execution.

**Node.js version**
22.

**Operating system**
Linux (container).

**Relevant logs or output**
```
[paperclip] Gemini ACP default unavailable; falling back to Gemini CLI.
YOLO mode is enabled. All tool calls will be automatically approved.
GEMINI_SANDBOX is true but failed to determine command for sandbox; install docker or podman
```
Reported run error: `YOLO mode is enabled. All tool calls will be automatically approved.`
Actual cause: line 3.

**Relevant config**
Not config dependent. Any non-zero exit without a structured error hits this path.

**Additional context**
The banner is unavoidable: adapters pass the approvals-bypass flag themselves, so the CLI prints it on every unattended run.

## What Changed

- Adds `packages/adapter-utils/src/benign-stderr.ts` with `isBenignAdapterStderrLine` and `firstMeaningfulStderrLine`: the first stderr line that is not a known-benign banner, falling back to the first non-empty line when every line is benign, so a run whose only output is a banner still reports something rather than nothing.
- Strips ANSI SGR codes before matching, since agent CLIs colour their fatal errors.
- Keeps the benign list to two conservative patterns (the approvals banner, and adapter-injected `[paperclip] ...` diagnostics), because anything listed here can hide a real error.
- Re-exports both from the package index.
- Adopts the selector in `codex`, `gemini`, `grok`, `opencode`, `cursor` and `pi`, at the single call site in each that derives the user-facing fallback error. Other `firstNonEmptyLine` uses (probe details and similar) are untouched.
- Removes the now-dead local helpers, including `codex-local`'s private version.

## Verification

- `npx vitest run packages/adapter-utils/src/benign-stderr.test.ts` — 8 tests covering benign filtering, reaching the real error, the exact production stderr stack above, the all-benign fallback, empty and whitespace input, and ANSI stripping.
- `npx vitest run packages/adapters/codex-local/src/server/execute.stderr-error.test.ts` — codex-local's pre-existing regression suite passes **unchanged**, which is the evidence that the shared helper is behaviour-compatible with the private one it replaces.
- Full suites for all six adapters plus `adapter-utils` run clean against the pre-existing baseline on this machine (the same set of platform-gated `local-process-sandbox` failures appears with and without this change).

## Risks

Low, with one thing to watch: the benign list decides which stderr line a user sees, so a pattern that is too broad could hide a real error. It is deliberately limited to two anchored patterns that no genuine failure message starts with, and the all-benign fallback means the selector can never return less information than the previous behaviour did. Structured-error precedence (`parsedError` and friends) is untouched in every adapter, so this only affects the fallback path.

## Model Used

Claude Opus 5 (claude-opus-5, 1M context), extended thinking, with tool use and code execution. The production stderr in this PR came from querying a live deployment; the gemini-cli behaviour referenced was reproduced locally against `@google/gemini-cli@0.53.1`.

## Checklist

### DIFF
--- a/packages/adapter-utils/src/benign-stderr.test.ts
+++ b/packages/adapter-utils/src/benign-stderr.test.ts
@@ -1,0 +1,54 @@
+import { describe, expect, it } from "vitest";
+import { firstMeaningfulStderrLine, isBenignAdapterStderrLine } from "./benign-stderr.js";
+
+const YOLO = "YOLO mode is enabled. All tool calls will be automatically approved.";
+
+describe("isBenignAdapterStderrLine", () => {
+  it("treats the approvals-bypass banner as benign", () => {
+    expect(isBenignAdapterStderrLine(YOLO)).toBe(true);
+  });
+
+  it("treats adapter-injected diagnostics as benign", () => {
+    expect(isBenignAdapterStderrLine("[paperclip] Confining the agent to the workspace.")).toBe(true);
+  });
+
+  it("never treats a real failure as benign", () => {
+    expect(isBenignAdapterStderrLine("Error: boom")).toBe(false);
+    expect(
+      isBenignAdapterStderrLine(
+        "GEMINI_SANDBOX is true but failed to determine command for sandbox; install docker or podman",
+      ),
+    ).toBe(false);
+  });
+});
+
+describe("firstMeaningfulStderrLine", () => {
+  it("skips banners to reach the real error", () => {
+    expect(firstMeaningfulStderrLine(`${YOLO}\nError: boom`)).toBe("Error: boom");
+  });
+
+  it("skips the exact production stack of banners", () => {
+    const stderr = [
+      "[paperclip] Adapter execution timeout: timeoutSec=14400 (sandbox default).",
+      "[paperclip] Gemini ACP default unavailable; falling back to Gemini CLI.",
+      YOLO,
+      "GEMINI_SANDBOX is true but failed to determine command for sandbox; install docker or podman",
+    ].join("\n");
+    expect(firstMeaningfulStderrLine(stderr)).toBe(
+      "GEMINI_SANDBOX is true but failed to determine command for sandbox; install docker or podman",
+    );
+  });
+
+  it("falls back to the first line when every line is benign", () => {
+    expect(firstMeaningfulStderrLine(`${YOLO}\n[paperclip] note\n`)).toBe(YOLO);
+  });
+
+  it("returns an empty string for empty or whitespace input", () => {
+    expect(firstMeaningfulStderrLine("")).toBe("");
+    expect(firstMeaningfulStderrLine(" \n\t\n")).toBe("");
+  });
+
+  it("strips ANSI colour codes before matching", () => {
+    expect(firstMeaningfulStderrLine(`${YOLO}\n[31mError: boom[0m`)).toBe("Error: boom");
+  });
+});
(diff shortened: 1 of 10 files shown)


stubbi:
@greptileai review

The only finding on the previous pass was the incomplete PR description. The full repository template is now filled in (Thinking Path, issue description in-PR, What Changed, Verification, Risks, Model Used, checklist). No code changed since the last review.

stubbi:
Rebased onto current master `d293dd3d1` (new head `dffbea6e1`). Worth a maintainer's attention: master has since merged `1f7959bc6` "fix(codex-local): skip benign stderr warnings when deriving the fallback run error (#10003)", which is the **codex-only** version of this fix. This PR is not redundant — it hoists that logic into `@paperclipai/adapter-utils` (`benign-stderr.ts`, plus ANSI stripping) and applies it to cursor, gemini, grok, opencode and pi as well. The conflict was resolved by deleting codex-local's now-duplicate private `isBenignCodexStderrLine`/`firstMeaningfulStderrLine` in favour of the shared helper; master's regression test `execute.stderr-error.test.ts` is kept as-is apart from repointing its import to `@paperclipai/adapter-utils` (the only added file vs. the original PR). Verified: adapter-utils + all six adapter packages typecheck clean (0 errors), `benign-stderr.test.ts` + `execute.stderr-error.test.ts` pass 15/15.

PR #10783, opened 2026-08-03
fix(gemini-local): do not request a CLI sandbox on managed sandbox targets

stubbi:
## Thinking Path

> - Agent runs can execute on a managed sandbox target: an isolated pod that provides the isolation, so the agent CLI inside it does not need to isolate itself
> - The Gemini CLI's `--sandbox` flag means something different from an approval policy: it asks the CLI to relaunch itself inside a Docker or Podman container
> - The agent create form derived `adapterConfig.sandbox` from `dangerouslyBypassSandbox`, which is Codex's approval-and-sandbox switch, so every Gemini agent made in the UI stored `sandbox: true` and asked for a nested container that a managed sandbox has no runtime for
> - Nothing then reconciled that config against the execution target, so the run died at CLI startup before doing any work
> - This pull request stops the form deriving the flag, and makes the adapter suppress an explicit sandbox request on managed sandbox targets while saying so in the run log
> - The benefit is that Gemini agents actually run on managed sandbox targets, and that an ignored setting is visible instead of silent

No existing issue. Describing it inline, following `.github/ISSUE_TEMPLATE/bug_report.yml`:

**What happened?**
Every `gemini_local` run on a managed sandbox target failed at CLI startup with exit code 44 and never did any work:

```
GEMINI_SANDBOX is true but failed to determine command for sandbox;
install docker or podman or specify command in GEMINI_SANDBOX
```

**Expected behavior**
The run executes. A managed sandbox pod deliberately has no container runtime, because the pod itself is the sandbox; the agent CLI inside it must not try to launch a nested container.

**Steps to reproduce**
1. Create a `gemini_local` agent through the agent create form, leaving every sandbox-related toggle at its default.
2. Inspect the stored `adapterConfig`: `sandbox` is `true`, because `build-config.ts` sets `ac.sandbox = !v.dangerouslyBypassSandbox` and that field defaults to `false`.
3. Run the agent on an execution target whose transport is `sandbox`.
4. The adapter passes a bare `--sandbox`; the CLI sets `GEMINI_SANDBOX=true` in `loadCliConfig` and `getSandboxCommand` then requires docker or podman on PATH.
5. The run exits 44 before executing anything.

Confirmed independently against `@google/gemini-cli@0.53.1` with an emptied `PATH`:

| flag | result |
| --- | --- |
| `--sandbox=none` | no error |
| `--no-sandbox` | no error |
| `--sandbox` | `GEMINI_SANDBOX is true but failed to determine command for sandbox` |

`master` at 18f391ef0.

**Deployment mode**
Self-hosted multi-tenant deployment using a Kubernetes sandbox provider. Any provider whose sandbox lacks a container runtime is affected.

**Installation method**
Docker image built from source.

**Agent adapter(s) involved**
`gemini_local`.

**Database mode**
PostgreSQL.

**Access context**
Server-side adapter execution on a remote sandbox execution target.

**Node.js version**
22.

**Operating system**
Linux (container), gVisor-isolated pod.

**Relevant logs or output**
Probed inside a live sandbox pod during a failing run:
```
GEMINI_SANDBOX=[UNSET]      docker: absent      podman: absent
gemini 0.52.0
```
`GEMINI_SANDBOX` is unset in the pod, which confirms the CLI sets it itself from `argv.sandbox` rather than inheriting it.

**Relevant config**
```json
{ "sandbox": true }
```
written by the create form without the user asking for it.

**Additional context**
The create form's `dangerouslyBypassSandbox` is Codex's approval-and-sandbox switch. `codex_local` overrides it in `NewAgent.tsx`; `gemini_local` does not, so it inherits the shared default and inverts into `sandbox: true`.

## What Changed

- `build-config.ts` no longer derives `ac.sandbox` from `dangerouslyBypassSandbox`. The option stays available directly in `adapterConfig` for anyone who does want the CLI's own container sandbox.
- Adds `resolveGeminiSandboxFlag({ configuredSandbox, usesManagedSandbox })` in `execute.ts`, which returns the sandbox argument plus an optional suppression reason. On a managed sandbox target an explicit request resolves to `--sandbox=none`; on local and SSH targets an explicit request is still honoured as `--sandbox`.
- Writes the suppression reason to the run log once per run, so an ignored setting is visible rather than silent.
- Updates the `sandbox` option documentation in `index.ts` to say what the flag actually does and where it is ignored.

## Verification

- `npx vitest run packages/adapters/gemini-local/src/server/execute.sandbox-flag.test.ts` — covers all four combinations of configured and managed, including that the suppression reason names both `adapterConfig.sandbox` and the managed sandbox.
- `npx vitest run packages/adapters/gemini-local` — full package suite; only a pre-existing, unrelated failure remains (`pre-selects gemini-api-key auth in the managed HOME for sandbox execution`, which fails identically on an unmodified checkout of `master` on this machine because the test's fake shell does not answer a remote file-size probe).
- `build-config.test.ts` asserts the create-form default in both toggle positions.
- Behaviour of the underlying CLI verified directly, see the reproduction table above.

## Risks

Low. The suppression is scoped to managed sandbox targets, where the flag provably cannot work, so no configuration that previously functioned changes behaviour. Local and SSH targets are unaffected. Users who had `sandbox: true` stored only because the form put it there get working runs; a user who genuinely wanted a nested container on a managed sandbox target was already getting a hard startup failure, and now gets a run plus a log line explaining the setting was ignored.

## Model Used

Claude Opus 5 (claude-opus-5, 1M context), extended thinking, with tool use and code execution. The failure was diagnosed from a live deployment's run records and confirmed by probing a running sandbox pod (`GEMINI_SANDBOX` unset, no docker, no podman), then reproduced locally against `@google/gemini-cli@0.53.1`.

## Checklist

### DIFF
--- a/packages/adapters/gemini-local/src/index.ts
+++ b/packages/adapters/gemini-local/src/index.ts
@@ -53,7 +53,7 @@ Core fields:
 - promptTemplate (string, optional): run prompt template
 - model (string, optional): Gemini model id. Defaults to auto.
 - engine (string, optional): leave unset/auto to use ACP when prerequisites pass and fall back to the Gemini CLI with diagnostics. Use "cli" to pin the CLI lane or "acp" to require ACP.
-- sandbox (boolean, optional): run in sandbox mode (default: false, passes --sandbox=none)
+- sandbox (boolean, optional): ask the Gemini CLI to relaunch itself inside a Docker or Podman container (default: false, passes --sandbox=none). Ignored on managed sandbox targets, which are isolated already and have no container runtime.
 - command (string, optional): defaults to "gemini"
 - extraArgs (string[], optional): additional CLI args
 - env (object, optional): KEY=VALUE environment variables
--- a/packages/adapters/gemini-local/src/server/execute.sandbox-flag.test.ts
+++ b/packages/adapters/gemini-local/src/server/execute.sandbox-flag.test.ts
@@ -1,0 +1,33 @@
+import { describe, expect, it } from "vitest";
+import { resolveGeminiSandboxFlag } from "./execute.js";
+
+describe("resolveGeminiSandboxFlag", () => {
+  it("passes --sandbox=none when the adapter config leaves sandbox off", () => {
+    expect(resolveGeminiSandboxFlag({ configuredSandbox: false, usesManagedSandbox: false })).toEqual({
+      sandboxArg: "--sandbox=none",
+      suppressedReason: null,
+    });
+  });
+
+  it("honors an explicit sandbox request on a local target", () => {
+    expect(resolveGeminiSandboxFlag({ configuredSandbox: true, usesManagedSandbox: false })).toEqual({
+      sandboxArg: "--sandbox",
+      suppressedReason: null,
+    });
+  });
+
+  it("suppresses the sandbox request inside a managed sandbox and says why", () => {
+    const result = resolveGeminiSandboxFlag({ configuredSandbox: true, usesManagedSandbox: true });
+    expect(result.sandboxArg).toBe("--sandbox=none");
+    expect(result.suppressedReason).toContain("managed sandbox");
+    expect(result.suppressedReason).toContain("adapterConfig.sandbox");
+  });
+
+  it("needs no suppression note when the managed sandbox run never asked for one", () => {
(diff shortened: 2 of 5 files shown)


stubbi:
@greptileai review

The only finding on the previous pass was the incomplete PR description. The full repository template is now filled in (Thinking Path, issue description in-PR, What Changed, Verification, Risks, Model Used, checklist). No code changed since the last review.

PR #10794, opened 2026-08-04
fix(issues): a heartbeat run that never started must not hold an issue's execution lock

### LINKED ISSUE #9874: clearExecutionRunIfTerminal never reaps a queued-but-never-started heartbeat-run, leaving an unbreakable executionRunId exec-lock on the issue
## Environment

- Package: `@paperclipai/server@2026.707.0` (dist-only vendored npm package)
- Component: heartbeat-run lifecycle / `clearExecutionRunIfTerminal`

## Summary

When a heartbeat-run is dispatched to the queue but **never starts** (it stays `status: "queued"` with `startedAt: null`), it is never treated as terminal, so `clearExecutionRunIfTerminal` never reaps it. The stuck run remains referenced as the issue's `executionRunId`. That dangling reference makes **checkout, release, AND status-PATCH all return HTTP 409 forever** — an unbreakable execution lock that only a board-level force-release can clear. The issue becomes permanently stuck in `in_progress`/`blocked` with no agent-reachable recovery path.

## Impact / severity

- **High for autonomous multi-agent operation.** A single queued-but-never-dispatched run silently bricks an issue. The assigned agent cannot check out, cannot release, cannot even PATCH status — every mutation 409s. From the agent's side it is indistinguishable from a live lock held by another run, so well-behaved agents correctly back off and the issue never recovers without human/board intervention.
- Observed on multiple issues in production; at least one required a manual board-level force-release to recover.

## Signature (how to detect an affected issue)

An issue exhibits the defect when ALL of the following hold:

- issue `status` is `in_progress` or `blocked`
- issue `checkoutRunId` is `null` (no live checkout holds it)
- issue `executionRunId` points at a run R
- `GET /api/heartbeat-runs/{R}` reports `status: "queued"` and `startedAt: null` (dispatched to the queue, never started)

While in this state, `POST /api/issues/{id}/checkout`, the release path, and `PATCH /api/issues/{id}` (status change) all return `409`.

## Root cause (hypothesis, from observed behavior)

`clearExecutionRunIfTerminal` only clears an issue's `executionRunId` when the referenced run is in a **terminal** state (e.g. `completed` / `failed` / `cancelled`). A run that is `queued` with `startedAt: null` and never dispatches is **never** in a terminal state, so the clear condition is never met and the reference is never released. There is no timeout that transitions a stale `queued` (never-started) run to a terminal state, so the lock is permanent.

## Suggested fix (either option resolves it)

1. **Reap stale queued runs in `clearExecutionRunIfTerminal`.** Treat a run that has been `queued` with `startedAt: null` beyond a threshold as terminal-reapable, so the exec-lock self-clears on the next lifecycle pass. **OR**
2. **Add a queued-dispatch timeout.** Transition a `queued` run that has not started within N minutes to a terminal `timed_out` (or `failed`) status. Once terminal, the existing `clearExecutionRunIfTerminal` path releases the `executionRunId` automatically.

Option 2 is the more general fix (it also bounds any other consumer that waits on run terminality), but either removes the permanent lock.

## Related

- A similar heartbeat/continuation-recovery defect is already tracked as #9867 ("Scheduled issue monitor is ignored by productive-terminal continuation recovery, causing immediate wake loop") — same subsystem (heartbeat-run lifecycle / recovery), useful context for whoever triages this.

coal:
## Thinking Path

> - The heartbeat subsystem gives each issue to one run at a time. It does this with an execution lock on the issue row (`executionRunId`, `checkoutRunId`).
> - The lock is released by a liveness test. Every release site asked the same narrow question: is the holding run in `TERMINAL_HEARTBEAT_RUN_STATUSES`?
> - `queued` and `scheduled_retry` are not terminal, and they are correctly not terminal. But a run in those states can have `startedAt IS NULL`. It has never executed a step and holds nothing.
> - So an issue whose holder is parked in `scheduled_retry` stays locked for the whole backoff window. Every structured field write by the issue's own assignee returns `409 Issue run ownership conflict`.
> - That makes the execution contract unsatisfiable. `done`, `in_review` and `blocked` are all PATCHes, so the issue is pinned to `in_progress` — which is the status that gets re-woken.
> - This pull request replaces the terminal-set test with one shared liveness predicate, and routes every lock-release site through it.
> - The benefit is that a dormant run stops bricking an issue, and the recovery sweep that exists to catch this case can finally see it.

Fixes #9874

Related PRs, found in the dedup search and linked here per CONTRIBUTING:

- #10748 — `fix(issues): release the execution lock held by a never-started scheduled_retry`. **This is the closest prior work and it was opened before this PR.** It fixes the same function with a narrower predicate: `scheduled_retry` with `startedAt == null` only. This PR is a superset — see "Relationship to #10748" below. If maintainers prefer that PR's smaller surface, this one should be closed in its favour and the extra coverage re-proposed separately.
- #10449 — `fix(server): reap stale queued execution locks safely` (overlapping intent on the `queued` arm).
- #9316 — `fix(heartbeat): reap stuck non-terminal runs and clear stale locks during suppression`.
- #10211 — `fix(heartbeat): expire runs stuck in queue beyond a bounded age` (the "option 2" framing in #9874).
- #10381 — `feat(issues): explain run-lock 409s with holder liveness and a conflict reason` (consumes the same liveness notion for its error text).

### Relationship to #10748

Reported to maintainers rather than resolved unilaterally, because the choice is a scope call:

- #10748 covers `scheduled_retry`. It does **not** cover `queued`, which is the state #9874 was actually filed about, so #9874 stays open if only #10748 lands.
- This PR covers both, adds the hand-forward grace window described under Risks, and applies the predicate to `recovery.sweepStaleIssueLocks` — the in-product backstop for exactly this failure, which carried its own copy of the terminal-set test and so skipped precisely the rows it exists to catch.

Both PRs modify `server/src/services/issues.ts`, so whichever lands second needs a rebase.

## What Changed

- Adds one shared predicate, `heartbeatRunHoldsIssueLock(run)`, as the single answer to "does this run still hold a claim on this issue?".
- A run holds the lock while it exists, is non-terminal, and either has `startedAt` set, or is still inside a short grace window with a reservation that has not gone stale.
- Routes every lock-staleness read through that predicate: `clearExecutionRunIfTerminal`, `clearCheckoutRunIfTerminal` (both the checkout run and the bundled execution run), `isTerminalOrMissingHeartbeatRun`, and the `stale` arm of `adoptStaleCheckoutRun`.
- Applies the same predicate in `recovery.sweepStaleIssueLocks`, which previously duplicated the terminal-set test.
- Deliberately does **not** change the `actorLive` / `adoptUnownedCheckoutRun` checks. Those ask about the caller's own run, where a never-started clause would read a live caller as dead.
- Leaves `heartbeatRunIsTerminalOrMissing` alone. Its remaining caller asks a different question: has this run's workspace finalize settled?

## Verification

- 301 tests pass across the 10 lock-related suites: `issues-service`, `issue-stale-execution-lock-routes`, `recovery-stale-issue-lock-sweep`, `execution-lock-orphan-cleanup`, `heartbeat-start-lock`, `heartbeat-lock-release-on-reassignment`, `issues-checkout-wakeup`, `heartbeat-auto-checkout`, `heartbeat-process-recovery`, `issue-recovery-actions`.
- `pnpm --filter @paperclipai/server typecheck` is clean.
- Mutation-checked, so the new coverage is not vacuous: reverting `heartbeatRunHoldsIssueLock` to the old terminal-set-only body reds exactly the 5 intended tests and no others.
- Two of the new tests pin a trap rather than the fix. Same run row, different lock **column**, opposite verdict. Both pass under the old predicate and the new one. `resolveSameRunOwnership` matches on `checkoutRunId` alone, so when the caller's own run already owns the checkout, the execution lock is never consulted. A holder planted only on `executionRunId` of an issue the prober had already checked out therefore cannot conflict, whatever its status. Anyone trying to reproduce #9874 with a synthetic row needs this, or they will conclude the bug is not real.

Measured on a live instance, 2026-08-03: 20 issues held against a `scheduled_retry_at` five days out. All 20 holding runs had `startedAt IS NULL`. A census of that instance's `heartbeat_runs` at the same moment:

| status | never started | total |
|---|---|---|
| succeeded | 0 | 14296 |
| failed | 1 | 5588 |
| cancelled | 1787 | 1964 |
| scheduled_retry | 35 | 35 |
| running | 0 | 6 |
| queued | 4 | 4 |

Every `scheduled_retry` run had never started. No `running` or `succeeded` run ever had `startedAt IS NULL`.

## Risks

- **The main risk is releasing a lock that is still legitimately held.** A never-started run can hold the lock correctly for a short window: `process_lost_retry`, `issue_execution_promoted` and the scheduled-retry parks in `services/heartbeat.ts` deliberately hand the lock forward from a dying run to its not-yet-claimed successor. Clearing unconditionally would race every hand-forward. The grace window exists for this, which is why the predicate is not simply "never started holds nothing". The normal wake path does not need the window; it locks lazily and stamps `executionRunId` only once the run is running.
- Behavioural shift, intended: an issue parked behind a dormant holder now becomes writable by its assignee sooner. Callers that treated a 409 as "wait forever" will now proceed.
- No database migration, no schema change, no API shape change. Server-side only, four files.
- The grace window is a time constant. Set it too low and a hand-forward races; too high and the original stall persists for that long. It is the one number a reviewer should push back on.

## Model Used

Claude Opus 5 (Anthropic), model ID `claude-opus-5`, running in Claude Code. Extended thinking enabled, with tool use (file edit, shell, test execution). Approximately 200K context window.

## Checklist

Notes on the two unchecked boxes, left unchecked deliberately rather than overlooked:

- **Branch name.** This branch was named from an internal ticket id before the rule was applied. A branch cannot be renamed on an open PR, and opening a replacement would put a third PR on the same function alongside #10748. Happy to re-open under a clean name if a maintainer prefers that.
- **Documentation.** No user-facing behaviour or configuration changes, so there is nothing in `docs/` this affects. Say the word if the grace window should be documented.
- **Previously red CI, now resolved and recorded here for the reviewer.** Every check on this branch failed on arrival, for a reason outside this diff. `ui/package.json` gained `@base-ui/react` in 9b9631b7 (#10707) while `pnpm-lock.yaml` stayed at its previous revision, because `pr.yml` hard-fails any PR that commits the lockfile. Every PR that did not itself touch a manifest then failed `pnpm install --frozen-lockfile` with `ERR_PNPM_OUTDATED_LOCKFILE` — including #10748, #10796, #10800 and #10801, not just this one. #10787 refreshed the lockfile and merged afterwards. This branch is now rebased onto master (aa6b6bcb) and all checks are green.

### DIFF
--- a/server/src/__tests__/issues-service.test.ts
+++ b/server/src/__tests__/issues-service.test.ts
@@ -34,8 +34,11 @@ import { instanceSettingsService } from "../services/instance-settings.ts";
 import {
   clampIssueListLimit,
   deriveIssueCommentRunLogAttribution,
+  heartbeatRunHoldsIssueLock,
   ISSUE_LIST_MAX_LIMIT,
+  type IssueLockHolderRun,
   issueService,
+  NEVER_STARTED_RUN_LOCK_GRACE_MS,
 } from "../services/issues.ts";
 import {
   WORKSPACE_WORKTREE_REQUIRES_PROJECT_CODE,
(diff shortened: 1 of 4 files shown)


coal:
Status re-measured 2026-08-04 06:25Z: head `0512d821`, 28 checks — 27 green, 1 skipped, `mergeStateStatus: CLEAN`, Greptile 5/5 on the current head with no open blocking threads. Ready for a maintainer whenever one has a slot.

One thing worth flagging so this isn't reviewed twice: **#10748 is an independent fix for the same defect**, opened a day earlier by another contributor. Two notes on how they relate, both measured rather than asserted:

- **Scope.** #10748 touches `server/src/services/issues.ts` only. This PR also routes `recovery/service.ts` through the shared predicate — `sweepStaleIssueLocks` carries its *own* copy of the `TERMINAL_HEARTBEAT_RUN_STATUSES` test, so the in-product backstop for this exact failure skipped precisely the rows it exists to catch. Fixing `issues.ts` alone leaves that fourth call site keyed the old way.
- **Its red is not its diff.** #10748's checks last ran at `2026-08-03T23:30Z` and 20 of them died at `pnpm install --frozen-lockfile` with `ERR_PNPM_OUTDATED_LOCKFILE … not up to date with <ROOT>/ui/package.json` — the window opened by #10707 (merged 23:25Z, manifest without lockfile) and closed by #10787 (03:47Z). This PR was red for the identical reason until it was rebased. A re-run should clear #10748 too.

Happy to close this in favour of #10748 if a maintainer prefers that shape — but the `recovery/service.ts` call site would need picking up either way.

Issue #10754, opened 2026-08-03
clean-poisoned-claude-sessions resolves every row against one CLAUDE_CONFIG_DIR and deletes healthy sessions of other agents


matusbr:
### Pre-submission checklist

- [x] I have searched existing open and closed issues and this is not a duplicate.
- [x] I am on the latest released version of Paperclip (or can reproduce on `master`).
- [x] I have confirmed the error originates in Paperclip itself — not in my agent adapter, API provider, or local configuration.

### What happened?

```
`packages/db/scripts/clean-poisoned-claude-sessions.ts` resolves a single Claude
configuration directory and uses it to locate the session JSONL of *every*
scanned row.

    // line 295
    const claudeConfigDir = resolveClaudeConfigDir(process.env, args.claudeConfigDir ?? undefined);

    // line 344, inside the per-row loop
    const classification = classifyRow(input, {
      claudeConfigDir,
      readJsonl: readJsonlSafe,
    });

Agents can pin their own profile directory through
`adapterConfig.env.CLAUDE_CONFIG_DIR`. As soon as more than one agent does so,
every row belonging to a different agent is looked up in the wrong directory,
classified as `missing_jsonl`, and — with `--delete-missing-jsonl` — deleted.

The session files are still on disk and perfectly healthy. Only the bookkeeping
row is gone, so the affected agents can no longer resume: every heartbeat starts
a fresh session with a full context.

Worth noting: the loop already has `row.agentId` in hand and uses it for the log
lines a few statements later. The information needed to resolve the right
directory is present and unused.

Two effects, and the second is the expensive one:

1. Loss of continuity — agents silently lose session resume and start over.
2. Cost — each wake-up re-reads a full context instead of resuming. In our
   deployment four agents consumed roughly 94M tokens in 50 minutes; the broken
   session bookkeeping is one of the contributing causes.

The failure is silent: the script reports success and prints a healthy-looking
summary, because from its point of view the files really are missing.

Verified against master on 2026-08-03 (file fetched directly, 462 lines): the
single-directory resolution at line 295 and its use at line 344 are unchanged.
```

### Expected behavior

```
Rows should be resolved against the profile directory of the agent that owns
them, falling back to the global value only when the agent has none.

If per-agent directories are deliberately out of scope for this script, then it
should refuse to run with --delete-missing-jsonl when it detects more than one
distinct CLAUDE_CONFIG_DIR among the agents — anything except deleting healthy
rows and reporting success.
```

[...]

PR #10808, opened 2026-08-04
fix(claude-local): honor explicit CLAUDE_CONFIG_DIR in local sandboxed runs

### LINKED ISSUE #10686: CLAUDE_CONFIG_DIR from adapterConfig.env is honoured for remote runs but ignored for local sandboxed runs
### Pre-submission checklist

### What happened?

An agent can pin its own Claude configuration directory through
`adapterConfig.env.CLAUDE_CONFIG_DIR`. The value **is** read
(`packages/adapters/claude-local/src/server/execute.ts`, line 445 on `master`):

```ts
const hasExplicitClaudeConfigDir =
  typeof configEnv.CLAUDE_CONFIG_DIR === "string" && configEnv.CLAUDE_CONFIG_DIR.trim().length > 0;
```

and it **is** honoured — but only on the remote path (line 566):

```ts
const useManagedRemoteClaudeConfig =
  executionTargetIsRemote &&
  adapterExecutionTargetUsesManagedHome(executionTarget) &&
  !hasExplicitClaudeConfigDir;
```

The local sandbox path ignores it entirely (line 528 and following):

```ts
const sharedClaudeConfigDir = resolveSharedClaudeConfigDir(process.env);
…
{ path: sharedClaudeConfigDir, access: "rw" },
{ path: path.join(path.dirname(sharedClaudeConfigDir), ".claude.json"), access: "rw" },
…
if (filesystemScope) env.CLAUDE_CONFIG_DIR = sharedClaudeConfigDir;
```

So **every local agent of an instance shares one profile directory** — and that
directory, including its session transcript store, is bind-mounted **read-write**
into each sandbox.

The asymmetry is what makes this look unintentional rather than by design: the
flag exists, it is computed from exactly this value, and one branch uses it.

### Expected behavior

If `adapterConfig.env.CLAUDE_CONFIG_DIR` is set, the local sandboxed run should
use it — the same value the remote path already respects. Agents that do not set
it keep the current shared directory, so existing deployments are unaffected.

### Steps to reproduce

1. Create two `claude_local` agents in one company, each with a different
   `adapterConfig.env.CLAUDE_CONFIG_DIR` and `filesystemScope: "workspace"`.
2. Run each once.
3. Inspect the run environment (`adapter.invoke` event, `payload.env`):
   both report the same shared directory, not the configured one.
4. Inspect the two configured directories on disk: neither receives the run's
   session data.

The same inspection on a **remote** target shows the configured value being used.

e55d7029

### Deployment mode

Local dev (pnpm dev)

### Installation method

### Agent adapter(s) involved

- [x] Claude Code
- [ ] Codex
- [ ] Cursor
- [ ] Droid
- [ ] Hermes
- [ ] Not adapter-specific (core bug)

### Database mode

Not database-related

### Access context

None

### Node.js version

v26.4.0

### Operating system

CachyOS (Arch-based), kernel 7.1.5

### Relevant logs or output

```shell

```

```json
``json
{
  "adapterType": "claude_local",
  "adapterConfig": {
    "filesystemScope": "workspace",
    "env": { "CLAUDE_CONFIG_DIR": "/path/to/agent-specific-profile" }
  }
}
```

### Additional context

```
Why this matters beyond tidiness: the shared directory is mounted read-write
into every sandbox, so anything one run writes into its transcripts is readable
by every other local agent of the same instance. In our deployment a credential
that one agent printed for verification ended up readable by agents that had no
business seeing it. The sandbox was doing its job; the shared profile directory
went around it.

A second, smaller consequence: per-agent project memory is impossible, because
the agents write into each other's memory directories.

A patch is prepared — honour the explicit value, fall back to the shared
directory when unset. It touches the five lines quoted above plus the bind list,
and we have been running with it since 2026-08-02 across three agents, verified
by inspecting the environment of an actual run.

Duplicate search (2026-08-02): searching this repository's issues and pull
requests for "CLAUDE_CONFIG_DIR" returns no results, so nothing to link.
```

### Privacy checklist

santhiprakash:
## Thinking Path

> - The claude_local adapter runs Claude CLI inside a local sandbox when filesystem or network scope is set.
> - The sandbox bind-mounts the Claude config directory and sets CLAUDE_CONFIG_DIR so Claude can read its settings and write transcripts.
> - Today the local sandbox always resolves that directory from process.env, so every local agent of one instance shares the same ~/.claude directory.
> - adapterConfig.env.CLAUDE_CONFIG_DIR is already read for remote runs and for deciding whether to prepare a managed config seed.
> - The local path never consults that explicit value, so an agent that sets its own CLAUDE_CONFIG_DIR still gets the shared directory.
> - This pull request makes the local sandbox resolve the config directory from adapterConfig.env.CLAUDE_CONFIG_DIR when it is set, falling back to the shared directory when it is not.
> - The benefit is that agents can have isolated Claude profiles, session transcripts no longer leak across agents, and the behavior matches the remote path.

Fixes: #10686
Refs: #6882, #10754

## What Changed

- `packages/adapters/claude-local/src/server/execute.ts` now resolves the local sandbox `CLAUDE_CONFIG_DIR` from `adapterConfig.env.CLAUDE_CONFIG_DIR` when that key is explicitly set.
- When no explicit value is set, the existing shared `~/.claude` behavior is unchanged.
- The resolved directory is used for the sandbox bind mount and for the in-sandbox `CLAUDE_CONFIG_DIR` environment variable.
- Adds `execute.claude-config-dir.test.ts`, a regression test that asserts both the writable bind mount and the in-sandbox environment follow the explicit value. It fails before this change and passes after.

## Verification

- `pnpm -C packages/adapters/claude-local typecheck` passes.
- `pnpm -r typecheck` passes for the full workspace.
- A local reproduction with two `claude_local` agents each setting a different `CLAUDE_CONFIG_DIR` now shows each run pointing to its own directory.
- `vitest run packages/adapters/claude-local/src/server/execute.claude-config-dir.test.ts` passes with the fix. With the `execute.ts` hunk reverted, the same test fails on both the bind-mount path and the in-sandbox variable (the two surfaces are set in different places, so both are asserted).
- The pre-existing `execute.remote.test.ts` fails on this host on an unrelated remote SSH sync assertion (`syncDirectoryToSsh` / `managedRemoteWorkspace`); it does not touch the local sandbox path and references no `CLAUDE_CONFIG_DIR` symbol.

## Risks

- Low. The change only activates when an agent explicitly sets `adapterConfig.env.CLAUDE_CONFIG_DIR`.
- Agents that do not set it keep the current shared `~/.claude` directory.
- PR #6882 works in the same area but addresses a different problem: it pins the default config directory to the agent home. This change does not overlap with that logic.
- Downstream consumer: once directories become per-agent, the `clean-poisoned-claude-sessions` maintenance script (`#10754`) resolves every row against one global directory and can silently misclassify healthy sessions of other agents as missing. That fix is tracked in `#10754` and should land alongside or before this change so the maintenance path stays correct.

## Model Used

Cognition Devin CLI, model SWE-1.7 Max, 262K context, tool use enabled. Assisted with code change, verification, and pull request authoring. The regression test was added during review follow-up and verified locally. Commit authored as Santhi Prakash.

## Checklist

### DIFF
--- a/packages/adapters/claude-local/src/server/execute.claude-config-dir.test.ts
+++ b/packages/adapters/claude-local/src/server/execute.claude-config-dir.test.ts
@@ -1,0 +1,140 @@
+import { afterEach, beforeEach, describe, expect, it, vi } from "vitest";
+
+// Regression coverage for honoring an explicit `adapterConfig.env.CLAUDE_CONFIG_DIR`
+// in the local sandboxed execution path. Before the fix, the sandbox bind mount and
+// the in-sandbox `CLAUDE_CONFIG_DIR` both resolved against `process.env` (the shared
+// directory), ignoring the per-agent value. A version that wires only one of the two
+// looks correct from inside the sandbox until something tries to write, so this test
+// asserts on both.
+
+const {
+  ensureAdapterExecutionTargetCommandResolvable,
+  ensureAdapterExecutionTargetRuntimeCommandInstalled,
+  executeClaudeAcp,
+  resolveAdapterExecutionTargetCommandForLogs,
+  runAdapterExecutionTargetProcess,
+} = vi.hoisted(() => ({
+  ensureAdapterExecutionTargetCommandResolvable: vi.fn(async () => undefined),
+  ensureAdapterExecutionTargetRuntimeCommandInstalled: vi.fn(async () => undefined),
+  // Force the ACP lane to fail so execute() falls back to the local CLI runner,
+  // which is the path that builds the filesystem sandbox.
+  executeClaudeAcp: vi.fn(async () => {
+    throw new Error('Transform failed with 1 error: execute.ts:818:0: ERROR: Unexpected "<<"');
+  }),
+  resolveAdapterExecutionTargetCommandForLogs: vi.fn(async () => "claude"),
+  runAdapterExecutionTargetProcess: vi.fn(async () => ({
+    exitCode: 0,
+    signal: null,
+    timedOut: false,
+    stdout: [
+      JSON.stringify({ type: "system", subtype: "init", session_id: "claude-session-1", model: "claude-sonnet" }),
+      JSON.stringify({
+        type: "assistant",
+        session_id: "claude-session-1",
+        message: { content: [{ type: "text", text: "hello" }] },
+      }),
+      JSON.stringify({
+        type: "result",
+        session_id: "claude-session-1",
+        result: "hello",
+        usage: { input_tokens: 1, cache_read_input_tokens: 0, output_tokens: 1 },
+      }),
+    ].join("\n"),
+    stderr: "",
+    pid: 123,
+    startedAt: new Date().toISOString(),
+  })),
+}));
+
+vi.mock("./acp.js", () => ({
+  createClaudeAcpExecutor: () => executeClaudeAcp,
+  formatClaudeAcpFallbackMessage: (reason: string) =>
+    `[paperclip] Claude ACP default unavailable; falling back to Claude CLI. ${reason} Set engine=acp to require ACP or engine=cli to silence this fallback.\n`,
+  resolveClaudeExecutionEngineForRun: async (ctx: { config: Record<string, unknown> }) =>
+    ctx.config.engine === "acp"
+      ? { engine: "acp", explicit: true }
+      : { engine: "acp", explicit: false },
+}));
+
+vi.mock("@paperclipai/adapter-utils/execution-target", async () => {
+  const actual = await vi.importActual<typeof import("@paperclipai/adapter-utils/execution-target")>(
+    "@paperclipai/adapter-utils/execution-target",
+  );
+  return {
+    ...actual,
+    ensureAdapterExecutionTargetCommandResolvable,
+    ensureAdapterExecutionTargetRuntimeCommandInstalled,
+    resolveAdapterExecutionTargetCommandForLogs,
+    runAdapterExecutionTargetProcess,
+  };
+});
+
+import { execute } from "./execute.js";
+
+function buildContext(config: Record<string, unknown> = {}) {
+  return {
+    runId: "run-1",
+    agent: {
+      id: "agent-1",
+      companyId: "company-1",
+      name: "Claude Coder",
+      adapterType: "claude_local",
+      adapterConfig: {},
+    },
+    runtime: {
+      sessionId: null,
+      sessionParams: null,
+      sessionDisplayId: null,
+      taskKey: null,
+    },
+    config,
+    context: {},
+    onLog: vi.fn(async () => {}),
+  };
+}
+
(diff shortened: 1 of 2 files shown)


matusbr:
Thanks for picking this up — the change matches what we run locally.

**One downstream consumer worth linking before this lands: #10754.**

`clean-poisoned-claude-sessions` resolves every session row against a single `CLAUDE_CONFIG_DIR`. That is correct *today* precisely because all local agents of one instance share one directory. Once this PR makes the directory per-agent, the cleanup tool resolves rows against the wrong profile.

The part that makes this worth a comment rather than a follow-up issue: it does not fail loudly. There is no error and no exit code — the tool simply does not find the sessions it was asked to clean, and reports a clean state that is not one. A maintenance path that silently stops maintaining is harder to notice than one that crashes.

We hit exactly this ourselves when we lifted the shared-directory assumption in our own tree: nothing turned red, and the thing that had quietly become wrong was a tool nobody was looking at.

**On the missing test the bot flagged** — a minimal one that fails before and passes after:

- build an `adapterConfig` whose `env.CLAUDE_CONFIG_DIR` differs from `process.env.CLAUDE_CONFIG_DIR`
- run the local sandboxed path
- assert that *both* the writable bind mount and the in-sandbox `CLAUDE_CONFIG_DIR` point at the explicit value

Before this change both resolve to the shared directory, so the assertion fails; after it, both follow the explicit value. Worth asserting on both rather than just the environment variable: the mount and the variable are set in two different places, and a version that sets only one looks correct from inside the sandbox until something tries to write.

Happy to open that test as a follow-up if you would rather keep this PR at five lines.

*(Reporters of #10686, not maintainers — the call on scope is yours.)*

santhiprakash:
Thanks @matusbr — both points addressed.

**1. Test.** Added `execute.claude-config-dir.test.ts` on the branch rather than as a follow-up, since you sketched it precisely and keeping the fix with its verification makes the change easier to review. It builds an `adapterConfig` whose `env.CLAUDE_CONFIG_DIR` differs from `process.env.CLAUDE_CONFIG_DIR`, runs the local sandboxed path, and asserts on **both** surfaces you called out — the writable bind mount and the in-sandbox `CLAUDE_CONFIG_DIR`. Verified it fails before the `execute.ts` change (both resolve to the shared directory) and passes after (both follow the explicit value). Pushed in `21786bca0`; CI is re-running.

**2. `#10754` linkage.** Linked it in the PR body (`Refs: #10754`) with the sequencing spelled out: once directories are per-agent, `clean-poisoned-claude-sessions` must resolve each row against its owning agent's directory, or it silently misclassifies healthy sessions as missing and deletes them with `--delete-missing-jsonl`. The patch proposed in `#10754` (per-row resolution keyed off `agentId`) is the right fix for that surface. I'd rather keep this PR to the adapter change than widen it, so I'm happy to hold merge until `#10754` is ready to land alongside — let me know if you'd prefer a different sequencing.

PR #10809, opened 2026-08-04
fix(authz): admit the mention grant to a blocker-only issue PATCH

coal:
## Thinking Path

> - Agents coordinate through issues. Ordering between issues is recorded with the `blockedByIssueIds` field, which the scheduler reads to decide what is ready to run
> - An agent that is mentioned on another agent's issue receives a mention-scoped grant. That grant is admitted for `issue:comment` and withheld from `issue:mutate`
> - `blockedByIssueIds` is reachable only through `issue:mutate`. So a mentioned agent can say "this must wait for that" in a comment, and cannot write the edge that makes the scheduler honour it
> - The ordering is then carried only as prose in issue descriptions, which drift apart. `POST /issues` already accepts `blockedByIssueIds` at create time, so the field itself was never the sensitive part
> - This pull request admits the mention grant to a PATCH whose body carries `blockedByIssueIds` and nothing else, at both boundary sites that withhold it
> - The benefit is that an ordering an agent is allowed to state, it is also allowed to record

No public issue exists. The defect is described below, following `.github/ISSUE_TEMPLATE/bug_report.yml`.

**What happened?**
An agent that holds a mention-scoped grant on an issue it is not assigned can post a comment on that issue, but receives `403` when it sends `PATCH /api/issues/:id` with only `blockedByIssueIds`. The same bearer, on the same issue, in the same minute, gets `201` for the comment and `403` for the blocker write.

**Expected behavior**
An actor that may state an ordering in a comment on an issue should also be able to record that ordering as a blocker edge on the same issue. Writing a blocker edge is a routing act, not a content edit.

**Steps to reproduce**
1. Create issue A assigned to agent 1, and issue B assigned to agent 2.
2. As agent 1, post a comment on issue A that mentions agent 2. This creates the mention-scoped grant.
3. As agent 2, `POST /api/issues/<A>/comments` — succeeds with `201`.
4. As agent 2, `PATCH /api/issues/<A>` with body `{"blockedByIssueIds": ["<B>"]}` — fails with `403`.

Reproduced on `master`. The two conjuncts responsible are `input.action !== "issue:mutate"` in `decideLowTrustAccess` and `input.action === "issue:comment"` in `decideBase`, both in `server/src/services/authorization.ts`.

**Deployment mode**
Not deployment-specific. The check is in the authorization service and applies to every mode.

**Access context**
Agent actor authenticated with an agent key, acting on an issue assigned to a different agent. Reproduces under both the standard trust preset and the low-trust review preset.

Related open pull requests, found while searching for duplicates. None of them change who may write a blocker edge, so there is no overlap in behaviour:

- Refs #10475 — rejects unknown issue PATCH fields. Adjacent to the body-narrowing guard here.
- Refs #8459 — lets recovery owners clear stale blockers. A different actor and a different basis.
- Refs #10269 — stops an implicit reopen when the same PATCH wires blockers.

## What Changed

- Added an `allow_issue_blocker_edge_grant` decision reason, so a blocker write stays distinguishable from a comment in the audit trail.
- Admitted the mention grant to `issue:mutate` at **both** sites that withheld it: `decideLowTrustAccess` (written in the negative, `action !== "issue:mutate"`) and `decideBase` (written in the positive, `action === "issue:comment"`). A blocker-edge mutation must clear both, so fixing one alone changes nothing observable.
- Qualified the request through the existing authorization `scope` channel rather than adding a new action verb. `scope.blockerEdgeOnly` is read by a single helper, `isBlockerEdgeOnlyIssueMutation`.
- Added a second grant basis: an actor that created or is assigned **every** issue it proposes to add as a blocker may record that ordering. The check is "every added blocker", not "any". A request that also drops an existing blocker is refused on this basis.
- Added the route seam. `PATCH /issues/:id` resolves the narrowing before the mutation gate, and passes the added-blocker set and a removal flag to the authorization layer.
- Skipped the assignee-ownership refusal for a blocker-edge write that the grant admitted. That refusal would otherwise deny exactly the case the grant exists to permit.
- Added tests for both boundary sites, both grant bases, and the body-narrowing guard.

### On not adding an `issue:block` verb

A new action was the obvious shape and was rejected on the merits. `PermissionKey` is re-enumerated at about ten sites, each a list of the form `action === "issue:read" || action === "issue:comment" || action === "issue:mutate"`. An action missing from one of those lists does not fail loudly — it falls through to that branch's default, and the defaults differ per site. The failure direction is unknown per site and would not be visible in review.

Reusing the existing `issue:comment` action for the narrowed PATCH was also considered and rejected. `issue:comment` carries a second, unrelated allow path: `allow_direct_parent_report` admits a comment on the current run issue's direct parent under the standard trust preset, with no mention grant at all. Asserting `issue:comment` for a blocker write would silently hand every agent the ability to rewrite its parent issue's blocker edges. That widening is invisible in the diff, because it appears as a narrowing from `issue:mutate` to `issue:comment`.

The scope channel avoids both. It adds no action to enumerate, inherits no unrelated allow path, and keeps the audit action as `issue:mutate`.

### The three `agentHasMentionGrantOnIssue` call sites after this change

| site | what it does now |
|---|---|
| `decideLowTrustAccess`, direct-parent report branch | Unchanged. Still `issue:comment` only; this is the low-trust direct-parent reporting path and has nothing to do with blocker edges. |
| `decideLowTrustAccess`, boundary branch | Now also admits `issue:mutate` when the caller proved the body is blocker-only. Reached through `resolveIssueMentionScopedGrant`. |
| `decideBase`, ordinary issue branch | Same change, at the ordinary boundary. Reached through `resolveIssueMentionScopedGrant`. |

## Verification

```
npx vitest run server/src/__tests__/authorization-service.test.ts
npx vitest run server/src/__tests__/issue-agent-mutation-ownership-routes.test.ts
```

New tests, all passing:

- `authorization-service.test.ts` — a low-trust mentioned agent is admitted to a blocker-edge `issue:mutate`, and the same actor with the same mention is denied without the scope. This negative control is what proves the low-trust site is load-bearing.
- `authorization-service.test.ts` — the same pair at the ordinary boundary, under the standard trust preset.
- `authorization-service.test.ts` — the ownership basis: admitted when the actor created or is assigned every added blocker; denied when one added blocker is owned by someone else; denied when the request also drops an existing blocker.
- `authorization-service.test.ts` — an empty added-blocker set cannot establish the ownership basis.
- `issue-agent-mutation-ownership-routes.test.ts` — a peer agent completes a blocker-only PATCH on another agent's active checkout, and the widened scope reaches the authorization layer.
- `issue-agent-mutation-ownership-routes.test.ts` — a PATCH carrying `blockedByIssueIds` **and** `title` is refused, and the widened scope is never set. This is the guard that stops an ordering write from smuggling a content edit.

Regression run over the suites that touch this path — `low-trust-red-team-routes`, `issue-dependency-wakeups-routes`, `issues-service`, `issue-comment-reopen-routes`, `issue-activity-events-routes` — 215 tests, all passing.

## Risks

Moderate. This widens an authorization boundary, so the guards are the load-bearing part.

- The widened grant is reachable only when the PATCH body has exactly one key, `blockedByIssueIds`, holding an array of non-empty strings. Anything else resolves to `null` and takes the ordinary `issue:mutate` path unchanged. The narrowing is proven in the route, at the one place that can see the whole body.
- The assignee-ownership refusal is skipped only for a request that was both narrowed and admitted by one of the two grant bases. Blocker validation, cycle checks and the low-trust control-plane guard in the handler all still run.
- The checked-out `409` is skipped along with it. An ordering edge does not contend with the assignee's content edits, and refusing it while the target is `in_progress` would defeat the grant in its main case. This is a deliberate behavioural change and the one most worth a reviewer's attention.
- The ownership basis refuses removals, so it cannot be used to drop another agent's blocker. The mention basis still permits removals, which matches what that grant already licenses.
- No migration. No schema change. The new decision reason is additive.

## Model Used

Claude Opus 5 (`claude-opus-5`), extended thinking, with tool use and code execution. The measurements quoted above — the boundary sites, the enumeration sites, and the test runs — were produced by running the commands, not from memory.

## Checklist

- [x] I have updated relevant documentation to reflect my changes — no shipped documentation describes this authorization boundary (searched `docs/`, `doc/`, `DESIGN.md`), so the behaviour is recorded in code comments and in this PR body

### DIFF
--- a/server/src/__tests__/authorization-service.test.ts
+++ b/server/src/__tests__/authorization-service.test.ts
@@ -75,6 +75,7 @@ async function createIssue(
     projectId?: string | null;
     parentId?: string | null;
     assigneeAgentId?: string | null;
+    createdByAgentId?: string | null;
     originKind?: string | null;
     originId?: string | null;
   } = {},
@@ -90,6 +91,7 @@ async function createIssue(
       projectId: input.projectId ?? null,
       parentId: input.parentId ?? null,
       assigneeAgentId: input.assigneeAgentId ?? null,
+      createdByAgentId: input.createdByAgentId ?? null,
       originKind: input.originKind ?? "manual",
       originId: input.originId ?? null,
     })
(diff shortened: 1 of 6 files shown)


coal:
Thanks — this one is real and is now fixed in `ed5a1f6`.

The removal guard was computed from a snapshot read before the write transaction, while `syncBlockedByIssueIds` takes `FOR UPDATE` and then replaces the whole set. So two overlapping updates could report `removesExistingBlockers: false` at authorization time and still delete an edge that landed in between — the exact write the guard exists to refuse.

Fix: the authorizing snapshot is carried down as an optimistic-concurrency precondition. It is re-read **under the same lock** that the replacement runs under, and the write refuses with `409` when it no longer matches. The lock is now also taken when the submitted set is empty, which it previously was not, so a removal-only request cannot slip past the check.

Scope: the precondition is applied only to writes admitted by the blocker-edge grant. An assignee's own PATCH keeps its current behaviour, so this adds no new failure mode to the normal path.

Test added in `issues-service.test.ts` — a concurrent writer adds an edge, the stale write is refused with `409`, the concurrent edge survives, and the same write succeeds once the precondition matches. I verified the test is load-bearing by disabling the guard: it fails with `promise resolved instead of rejecting`.

PR #10818, opened 2026-08-04
feat(monitor): synthetic large-body probes for GitHub webhook endpoint

mv50000:
## Summary
- Adds synthetic endpoint probes to `check-github-webhook-health.ts`: unsigned small (128 B) + GitHub-PR-sized (64 KB) POSTs through the real proxy chain (edge nginx via `--resolve`, plus the local nginx hop), all expected to return 401 (missing signature).
- Probe failures post to the existing Slack alert with a 6 h re-alert throttle (`probeFailureAlertedAt` in the monitor state file); healthy runs clear it.
- Alert context now tells the operator to repeat diagnosis with a >26 KB body; failure block links the 2026-08-04 root cause (`/var/lib/nginx/body` perms) and fix.

## Why
2026-08-04 incident: all GitHub webhook deliveries (saatavilla 90, sunspot 123) failed with a bare nginx 500 for ~3 days because the nginx request-body temp dir lost write perms after a stray `apt-get install nginx-light`. Small diagnostic curls returned a healthy-looking 401 — only real-sized payloads failed, so the alert's own diagnosis hint was misleading. These probes catch that class within one monitor run.

## Test plan
- [x] `--dry-run` on the paperclip host: all 3 probes OK (401), summary line shows `probe OK`
- [x] Negative test `PAPERCLIP_WEBHOOK_EDGE_IP=192.0.2.1`: edge probes FAIL (`curl exit 28 (timeout)`), local OK, summary `probe FAILING`, exit 1
- [x] Error details no longer embed the 64 KB curl command line


mv50000:
Opened against upstream by mistake — internal fork change, re-filed in the fork. Apologies for the noise.

PR #9011, opened 2026-07-04
fix(codex-local): default retry backoff for usage-limit errors without a reset clock

0xopaque:
## Thinking Path

> - Paperclip is the open source app people use to manage AI agents for work
> - The codex-local adapter classifies failed runs; `extractCodexRetryNotBefore` recognizes ChatGPT usage-limit messages and returns the reset time so the transient-upstream machinery can space retries
> - The matcher only recognizes the variant that names a model and reset clock ("You've hit your usage limit for X. Switch to another model now, or try again at 11:31 PM.")
> - ChatGPT also emits a generic variant with no reset time ("You've hit your usage limit. Upgrade to Pro …"); this falls through — not transient, no retryNotBefore — and surfaces as a generic adapter failure
> - Issue-driven wakes (`issue_assigned`, `issue_continuation_needed`, `missing_issue_comment`) then hot-retry a quota that cannot succeed until it resets, each attempt failing within seconds
> - This PR detects the generic usage-limit wording and returns a default 30-minute `retryNotBefore`, engaging the existing transient classification and retry spacing; an explicit reset clock still wins when present
> - The benefit is that quota-exhausted codex agents back off instead of crash-looping the wake queue

## Linked Issues or Issue Description

No existing issue found (searched: "usage limit retry codex", "codex quota", "transient_upstream retry"). Bug description per the bug template:

Refs #8692 — the claude-local sibling of this bug family (session-limit wording not classified, tight retry loop); this PR fixes the codex-local usage-limit variant.

**What happened:** a codex-local agent whose ChatGPT account had exhausted its usage limit was re-woken by issue wakes and failed within ~3 seconds each time (8 runs in 2 minutes observed), because the error text "You've hit your usage limit. Upgrade to Pro (https://chatgpt.com/upgrade) …" matches neither `CODEX_USAGE_LIMIT_RE` (which requires "for <model> … try again at <time>") nor `CODEX_TRANSIENT_UPSTREAM_RE`'s scoped shapes.

**Expected:** quota exhaustion is transient-until-reset; the run should carry a `retryNotBefore` so retries are spaced, exactly as the recognized usage-limit variant already behaves.

## What Changed

- `packages/adapters/codex-local/src/server/parse.ts`: `extractCodexRetryNotBefore` now falls back to a generic usage-limit regex when the model-specific variant is absent or its clock text does not parse, returning `now + 30 minutes`. The explicit reset clock, when present and parseable, is still preferred.
[...]

PR #10822, opened 2026-08-04
fix(claude-local): back off to the real reset when an account usage window is refused

robelkin:
## Thinking Path

> - The `claude-local` adapter runs the Claude CLI and classifies its failures. The server uses that classification to decide when to retry a run.
> - The platform already has the correct contract for a provider quota: `errorFamily: provider_quota` plus a `retryNotBefore` timestamp. `heartbeat.ts` honours that timestamp and overrides its normal bounded backoff.
> - Nothing populated that contract for an account-level usage window, because nothing recognised the error. The CLI wording did not match the adapter's quota patterns.
> - The run was therefore classified as an ordinary transient upstream error. It retried on a 2m/10m/30m/2h backoff against a window that can last for days.
> - Every one of those retries is refused before it reaches the model. It consumes a run, returns zero tokens, and makes no progress.
> - This pull request classifies the refusal correctly and reads the exact reset time from the structured event the CLI already sends.
> - The benefit is that a task waits once until the real reset, instead of retrying every few minutes for the length of the window.

No public issue exists for this. The problem is described inline below, following `.github/ISSUE_TEMPLATE/bug_report.yml`.

Related open work, for reviewer context — neither fixes this case:

- Refs #3425 — earlier work on Claude usage-limit detection in the same adapter. It is unmerged and has not been updated since April. This PR targets what is on `master` today.
- Refs #9011 — the sibling problem in `codex-local`: a usage-limit error with no reset clock. Same class of bug in a different adapter.

**What happened?**

The Claude CLI refuses a run when an account-level usage window is exhausted. It reports:

```
You've hit your weekly limit · resets 7pm (Europe/London)
```

The adapter does not recognise this wording as a quota error.

| Pattern | Knows | Matches the wording above? |
|---|---|---|
| `CLAUDE_PROVIDER_QUOTA_RE` | `You've hit your **session** limit`, `weekly limit **reached**` | No |
| `CLAUDE_EXTRA_USAGE_RESET_RE` | the same prefix list | No, so no reset time is extracted |

Because quota is not recognised, the raw stream wins the classification race instead. The stream contains `"error":"rate_limit"` and `api_error_status: 429`. The run is filed as `claude_transient_upstream` with no `retryNotBefore`. It then retries on the bounded 2m/10m/30m/2h backoff.

This is the worst possible response to a standing multi-day window. The window refuses every attempt before it reaches the model. Each attempt burns a run and returns zero tokens. The retry loop simply repeats for the entire length of the window.

**Expected behavior**

The adapter must classify an account-level usage-window refusal as `provider_quota`. It must set `retryNotBefore` to the true reset time. The run must then wait once until that reset, instead of retrying every few minutes.

**Steps to reproduce**

1. Run a `claude_local` agent with a Claude subscription whose weekly usage window is exhausted.
2. Observe that the CLI refuses the run and emits `"You've hit your weekly limit · resets <clock> (<tz>)"`.
3. Inspect the recorded failure. Before this change it is `claude_transient_upstream` with no `retryNotBefore`.
4. Observe that the run retries within about 2 minutes and is refused again. The loop continues until the window expires on its own.

You do not need an exhausted account to see the classification bug. Pass the refused stdout to `isClaudeProviderQuotaError` directly. The added tests do exactly this, and they fail on `master`.

Reproduced on `master`. The branch is rebased on current `master`.

**Deployment mode**

Local / self-hosted.

**Agent adapter(s) involved**

Claude Code (`claude-local`). The server-side change in `recovery/service.ts` is adapter-independent.

## Root cause

The CLI already hands us the answer. It sends a structured event on its stream-json stdout:

```json
{"type":"rate_limit_event","rate_limit_info":{"status":"rejected",
  "resetsAt":1785866400,"rateLimitType":"seven_day",
  "overageStatus":"rejected","overageDisabledReason":"out_of_credits"}}
```

`resetsAt` is epoch seconds and it is exact. The adapter discarded this event and matched the prose instead.

That distinction matters beyond convenience. The prose carries a wall clock with no date. A multi-day window therefore parses to *tonight* rather than to the true reset. Only the structured field survives a window longer than 24 hours.

## What Changed

- Added `readClaudeRateLimitRejection()`. It reads the last `status: "rejected"` event from stdout. `isClaudeProviderQuotaError` trusts that event outright. `extractClaudeRetryNotBefore` prefers it whenever it gives a sane future time (30 days or less), and falls back to the prose otherwise. It ignores `status: "allowed"` events, which the CLI also emits during healthy runs.
- Widened the prose match, which is now the fallback. The per-window phrasings share one `CLAUDE_USAGE_LIMIT_WORDING` alternation. It matches the *shape* `you've hit your <window> limit` instead of listing windows one at a time. A window that the CLI adds later can no longer degrade silently to a generic transient retry. This generalisation is the real guardrail. The structured event is the belt.
- Made the reset hint terminate on `"` and `,`, as well as on `.` and a newline. The haystack includes raw stream-json lines. There the prose sits inside a JSON string, so it never matched at all.
- Added the same wording to `PROVIDER_QUOTA_ERROR_RE` server-side, and taught `parseProviderQuotaClockReset` the `resets 7pm (TZ)` form alongside the API's `try again at ...`. This is defence in depth for a run whose adapter did not tag `errorFamily`. Exact reset times still come from the adapter's persisted `retryNotBefore`, which is checked first.
- Fixed a pre-existing failure in `test.probe.test.ts`. The hello probe called only `isClaudeTransientUpstreamError`, which deliberately yields to the quota classifier. A usage-limit refusal therefore surfaced as a hard "probe failed" instead of the wait-and-retry warning that the test asserts. It now checks both. Happy to split this out if you prefer it separate.

## Verification

Replay a refused run's stdout through the classifier:

| | before | after |
|---|---|---|
| `errorCode` | `claude_transient_upstream` | `provider_quota` |
| `retryNotBefore` | none | the exact reset epoch |
| next retry | +2 min | +405 min |

New tests fail on `master` and pass on this branch. I confirmed this by reverting only the source files to `master` and keeping the new tests in place:

```
$ git checkout origin/master -- packages/adapters/claude-local/src/server/parse.ts \
    packages/adapters/claude-local/src/server/test.ts \
    server/src/services/recovery/service.ts

$ npx vitest run src/server/parse.test.ts            # in packages/adapters/claude-local
  Tests  9 failed | 40 passed (49)

$ npx vitest run src/services/recovery/provider-failure-classification.test.ts   # in server
  Tests  2 failed | 10 passed (12)
```

With the fix restored, both suites are green: 49/49 and 12/12.

`parse.test.ts` gains a fixture that carries all three refusal shapes the CLI emits in one stream: the `rate_limit_event`, the assistant message, and the result event. It asserts that the classifier returns quota and the exact epoch reset, and not a 2-minute transient retry.

Wider suites: claude-local 89/90, server recovery 47/47, heartbeat-retry-scheduling 30/30, adapter typecheck clean. The one remaining adapter failure is `execute.remote.test.ts` (`syncDirectoryToSsh` called twice, not once). It reproduces unchanged on clean `master`. It is pre-existing and unrelated, so I left it alone.

## Risks

Low risk, but there are two behaviour changes worth review:

- **A wider quota match.** `CLAUDE_USAGE_LIMIT_WORDING` now matches the shape `you've hit your <window> limit` rather than a fixed list. The `<window>` part is bounded to 40 characters and must not cross a newline or a `.` or a `·`. A false positive would make a run wait for a reset instead of retrying quickly. I judged that safer than the current failure, which retries a dead session for days. The negative test for a plain `529 overloaded_error` still asserts transient, not quota.
- **A longer wait.** A correctly classified run now waits until the real reset instead of retrying in 2 minutes. That is the intent. The 30-day horizon in `extractClaudeRetryNotBefore` guards against a misparse parking a run indefinitely, and falls back to the caller's normal backoff.

No migration. No schema change. No API change. The structured-event path is additive: if the event is absent, behaviour falls back to the prose path.

## Model Used

Claude Opus 5 (`claude-opus-5`, 1M context variant), with extended thinking, tool use, and local test execution. Running as an autonomous agent under Claude Code.

## Checklist

- [ ] I have updated relevant documentation to reflect my changes — no documentation describes this classification behaviour, so there was nothing to update

### DIFF
--- a/packages/adapters/claude-local/src/server/parse.test.ts
+++ b/packages/adapters/claude-local/src/server/parse.test.ts
@@ -11,7 +11,20 @@ import {
   isClaudeUnknownSessionError,
   isClaudeImageProcessingError,
   isClaudeModelNotFoundError,
+  readClaudeRateLimitRejection,
 } from "./parse.js";
+
+// A stream-json run refused by the account-level seven-day usage window. The CLI
+// reports the refusal three ways in one stream — the structured rate_limit_event,
+// the synthetic assistant message, and the result event — and this fixture keeps
+// all three so the classifier is exercised against the shape it really sees.
+// `resetsAt` 1785866400 is 2026-08-04T18:00:00Z, i.e. 7pm Europe/London.
+const WEEKLY_LIMIT_STDOUT = [
+  '{"type":"rate_limit_event","rate_limit_info":{"status":"allowed","resetsAt":1785866400,"rateLimitType":"seven_day","overageStatus":"allowed","isUsingOverage":false},"uuid":"a1","session_id":"s1"}',
+  '{"type":"rate_limit_event","rate_limit_info":{"status":"rejected","resetsAt":1785866400,"rateLimitType":"seven_day","overageStatus":"rejected","overageDisabledReason":"out_of_credits","isUsingOverage":false},"uuid":"a2","session_id":"s1"}',
+  '{"type":"assistant","message":{"model":"<synthetic>","role":"assistant","content":[{"type":"text","text":"You\'ve hit your weekly limit · resets 7pm (Europe/London)"}]},"session_id":"s1","error":"rate_limit","is_api_error_message":true}',
+  '{"type":"result","subtype":"success","is_error":true,"terminal_reason":"api_error","api_error_status":429,"result":"You\'ve hit your weekly limit · resets 7pm (Europe/London)","total_cost_usd":0,"session_id":"s1"}',
+].join("\n");
 
 describe("detectClaudeLoginRequired", () => {
   it("classifies Claude's invalid API key login prompt as auth required", () => {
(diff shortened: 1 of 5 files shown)


robelkin:
Closing this as stale. It's a separate resiliency improvement (backing off to the real account reset when a usage window is refused) and has since gone to conflict with `master`. It isn't load-bearing for current operation, so I'd rather not carry a conflicting PR. Happy to reopen and rebase if there's interest in taking the backoff behaviour upstream.

PR #3825, opened 2026-04-16
fix(routines): recover from stale routine execution lock conflicts

DeadlySilent:
## Thinking Path

> - Paperclip orchestrates AI agents for zero-human companies, and routine execution is a core mechanism for keeping issue/company operations moving without manual intervention.
> - The routines subsystem creates and coalesces execution issues, then queues assignment wakeups so agents can resume work.
> - In production, recovery/startup paths were repeatedly failing on `issues_open_routine_execution_uq` when an old routine execution issue still held a stale execution lock.
> - The initial fix added stale-lock repair and retry logic.
> - Live incident follow-up showed an additional gap: when `execution_run_id` was already null, open routine-origin duplicates could still be created.
> - This PR now includes both the stale-lock recovery logic and an open-routine-origin uniqueness/coalescing follow-up so duplicates are prevented in both lock-bound and lock-cleared paths.

## What Changed

- Added stronger unique-conflict detection for `issues_open_routine_execution_uq` across both `constraint` and `constraint_name` error shapes (`23505`).
- Added stale routine execution lock repair helper that:
  - locates the open routine_execution issue with an execution lock,
  - checks linked heartbeat run status,
  - clears `executionRunId`, `executionAgentNameKey`, and `executionLockedAt` only when the run is not live.
- Wired repair+retry handling into both:
  - routine issue creation conflict path,
  - `queueIssueAssignmentWakeup` conflict path.
- Added follow-up coalescing fallback in `findLiveExecutionIssue` to return any open routine-origin issue (even without live heartbeat binding), so conflict paths coalesce instead of creating duplicates.
- Added DB guardrail partial unique index for open routine-origin issues:
  - `issues_open_routine_origin_uq`
- Added migration:
  - `packages/db/src/migrations/0057_open_routine_origin_uniqueness.sql`
- Added/updated tests in `server/src/__tests__/routines-service.test.ts` to validate stale-lock recovery behavior.

## Verification

- `pnpm vitest run server/src/__tests__/routines-service.test.ts`
- `pnpm --filter @paperclipai/db run check:migrations`
- `pnpm exec tsc --noEmit -p server/tsconfig.json`
- `pnpm exec tsc --noEmit -p packages/db/tsconfig.json`
- Manual validation against live upgraded instance:
  - reproduced pre-fix conflict signature,
  - verified post-fix coalescing on routine rerun,
  - verified duplicate-title excess reduced to zero after cleanup + fix.

## Risks

[...]

PR #10823, opened 2026-08-04
fix(heartbeat): defer contended routine-execution lock claim instead of failing the run

robelkin:
## Thinking Path

> - Routines are the scheduling subsystem. Each fire creates an execution issue, and an agent picks that issue up on its next heartbeat.
> - A partial unique index keeps at most one *locked* open routine-execution issue per routine identity. The lock itself is stamped lazily, when the run is claimed.
> - Under `concurrencyPolicy: always_enqueue` each fire creates its own execution issue. A later fire can therefore be claimed while an earlier fire still holds the routine lock.
> - That collision raises a Postgres `23505` unique violation inside the claim path. The exception was unhandled, so it killed the whole run.
> - The fire is then lost silently. There is no execution issue, no output, and no surfaced error. It appears only later as a gap.
> - This pull request catches that specific violation and leaves the run queued instead of destroying it.
> - The benefit is that `always_enqueue` serialises cleanly under contention, rather than dropping fires.

No public issue exists for this. The problem is described inline below, following `.github/ISSUE_TEMPLATE/bug_report.yml`.

Related prior work, for reviewer context:

- Refs #3825 — closed and never merged. It addressed the same index from the dispatch-time create path and added stale-lock repair. This PR fixes the *claim-time* lock stamp, which is a different code path. I mention it so a reviewer can judge whether the two should be reconciled.

**What happened?**

A routine fire is dropped silently. There is no execution issue, no output, and no surfaced error. It becomes visible only as a gap, when someone notices the routine did not run. Two consecutive contended fires produce a multi-hour hole.

The underlying error is a Postgres unique violation in the claim path:

```
code: '23505'
constraint_name: 'issues_open_routine_execution_uq'
query: 'update "issues" set "execution_run_id" = $1, ... where ...'
```

The failure mode is the bad part. Contention on the lock is a normal and expected condition under `always_enqueue`. The response was to destroy the fire instead of waiting for the lock.

**Expected behavior**

A contended fire must wait for the lock and then run. It must not be dropped. `always_enqueue` must serialise fires rather than lose them.

**Steps to reproduce**

1. Configure a routine with `concurrencyPolicy: always_enqueue`.
2. Let a fire create an execution issue and hold the routine execution lock. Either let it run, or leave a stale lock that the sweeper has not yet cleared.
3. Fire the routine again, so a second execution issue is created and its run is claimed while the first still holds the lock.
4. The claim-time `UPDATE` that stamps `execution_run_id` raises `23505` on `issues_open_routine_execution_uq`. The run dies and no execution issue is produced.

The added test reproduces this against embedded Postgres, so no live instance is needed.

Reproduced on `master`. The branch is rebased on current `master`.

**Deployment mode**

Local / self-hosted.

**Agent adapter(s) involved**

Not adapter-specific. This is a core routines/heartbeat bug.

## Root cause

The partial unique index `issues_open_routine_execution_uq` (`packages/db/src/schema/issues.ts`) enforces at most one *locked* open routine-execution issue per routine identity `(companyId, originKind, originId, originFingerprint)`. It only activates when `execution_run_id is not null`.

Under `always_enqueue` each fire creates its own execution issue, and `claimQueuedRun` stamps the lock lazily at claim time. When a later fire's run is claimed while an earlier fire's execution issue is still open and still holds its lock, the `execution_run_id` null-to-set stamp flips the new row into that index. Postgres then raises `23505`.

That exception was unhandled in the claim path, so it terminated the run.

## What Changed

- Moved the lock stamp in `claimQueuedRun` to *before* the run flips to `running`, and wrapped it in a targeted catch for that specific unique violation.
- On contention the run is left queued, and the function returns `null`. A later tick retries it once the prior execution issue completes, or once the stale-lock backstop sweeper clears the lock.
- Made the conflict detector walk the error `cause` chain and match both driver shapes. `node-postgres` exposes the index name as `constraint`. The porsager `postgres` driver uses `constraint_name`. Matching only one shape silently misses the violation under the other driver.
- Added `server/src/__tests__/heartbeat-routine-execution-lock-contention.test.ts`.

## Verification

The new test uses embedded Postgres and reproduces the exact failure: the same `UPDATE`, and the same `issues_open_routine_execution_uq` / `23505`. It then asserts that the contended fire is deferred, and that it self-heals once the prior lock is released.

It fails before the change and passes after. I confirmed this by reverting only `heartbeat.ts` to `master` and keeping the test in place:

```
$ git checkout origin/master -- server/src/services/heartbeat.ts
$ npx vitest run src/__tests__/heartbeat-routine-execution-lock-contention.test.ts

  × defers a contended routine-execution fire instead of failing it
  × runs the deferred fire once the prior routine lock is released
    PostgresError: duplicate key value violates unique constraint
      "issues_open_routine_execution_uq"
      code: '23505'
  Tests  2 failed (2)
```

With the fix restored, both pass (2/2), and the deferral is visible in the log:

```
claimQueuedRun: deferred routine execution run; routine lock held by a concurrent execution
```

## Risks

Low risk, and the change is narrow.

- **A deferred run is now returned as `null`.** Callers must treat that as "nothing to claim this tick", which is the existing contract for an unclaimable run. The run stays queued, so a later tick picks it up. The pre-existing stale-lock sweeper is what guarantees this terminates if the holder died.
- **The catch is deliberately narrow.** It matches only `23505` on `issues_open_routine_execution_uq`. Any other error still propagates, so this does not mask unrelated failures.
- **A behaviour change under contention only.** On the uncontended path the lock stamp simply happens slightly earlier in the same function. Ordering the stamp before the `running` flip is what makes the failure recoverable, because the run has not yet been marked started.

No migration. No schema change. No API change.

## Note for maintainers

`routines.ts` has a similar `23505` check on the dispatch-time create path, and it matches only `.constraint`. Under the porsager driver it therefore likely misses the violation entirely. I have not touched it here, to keep this PR to one fix. It looks like the same latent gap and may deserve a follow-up.

## Model Used

Claude Opus 5 (`claude-opus-5`, 1M context variant), with extended thinking, tool use, and local test execution. Running as an autonomous agent under Claude Code.

## Checklist

- [ ] I have updated relevant documentation to reflect my changes — no documentation describes this locking behaviour, so there was nothing to update

### DIFF
--- a/server/src/__tests__/heartbeat-routine-execution-lock-contention.test.ts
+++ b/server/src/__tests__/heartbeat-routine-execution-lock-contention.test.ts
@@ -1,0 +1,338 @@
+import { randomUUID } from "node:crypto";
+import { eq, sql } from "drizzle-orm";
+import { afterAll, afterEach, beforeAll, describe, expect, it, vi } from "vitest";
+import {
+  agents,
+  agentWakeupRequests,
+  companies,
+  createDb,
+  heartbeatRuns,
+  issues,
+} from "@paperclipai/db";
+import {
+  getEmbeddedPostgresTestSupport,
+  startEmbeddedPostgresTestDatabase,
+} from "./helpers/embedded-postgres.js";
+import { heartbeatService } from "../services/heartbeat.ts";
+import { runningProcesses } from "../adapters/index.ts";
+
+// Regression coverage for routine execution-lock contention. A routine with
+// concurrencyPolicy=always_enqueue can have a later fire's execution issue claimed
+// while a prior fire's execution issue still holds the routine execution lock
+// (execution_run_id set). The null->set lock stamp then collides with the
+// issues_open_routine_execution_uq partial unique index. Before the fix this raised
+// an unhandled 23505 ("Failed query") that killed the run and dropped the digest
+// with no visible error; now the contended run is left queued and retried once the
+// prior lock frees.
+
+const mockAdapterExecute = vi.hoisted(() =>
+  vi.fn(async () => ({
+    exitCode: 0,
+    signal: null,
+    timedOut: false,
+    errorMessage: null,
+    summary: "Routine execution-lock contention test run.",
+    provider: "test",
+    model: "test-model",
+  })),
+);
+
+vi.mock("../adapters/index.ts", async () => {
+  const actual = await vi.importActual<typeof import("../adapters/index.ts")>("../adapters/index.ts");
+  return {
+    ...actual,
+    getServerAdapter: vi.fn(() => ({
+      supportsLocalAgentJwt: false,
+      execute: mockAdapterExecute,
+    })),
+  };
+});
+
+const embeddedPostgresSupport = await getEmbeddedPostgresTestSupport();
+const describeEmbeddedPostgres = embeddedPostgresSupport.supported ? describe : describe.skip;
+
+if (!embeddedPostgresSupport.supported) {
+  console.warn(
+    `Skipping routine execution-lock contention tests on this host: ${embeddedPostgresSupport.reason ?? "unsupported environment"}`,
+  );
+}
+
+async function waitForCondition(fn: () => Promise<boolean>, timeoutMs = 3_000) {
+  const deadline = Date.now() + timeoutMs;
+  while (Date.now() < deadline) {
+    if (await fn()) return true;
+    await new Promise((resolve) => setTimeout(resolve, 50));
+  }
+  return fn();
+}
+
+describeEmbeddedPostgres("heartbeat routine execution-lock contention", () => {
+  let db!: ReturnType<typeof createDb>;
+  let heartbeat!: ReturnType<typeof heartbeatService>;
+  let tempDb: Awaited<ReturnType<typeof startEmbeddedPostgresTestDatabase>> | null = null;
+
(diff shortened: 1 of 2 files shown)


Oldrich333:
Refs this PR: we hit the same constraint in periodic recovery promotion (`releaseIssueExecutionAndPromote`) on a private deployment. Upstream contribution [paperclipai/paperclip#11964](https://github.com/paperclipai/paperclip/pull/11964) includes the same claim-path deferral from this branch plus recovery-path guards. Happy to rebase if #10823 lands first.

robelkin:
Heads up: #11964 (@Oldrich333) supersedes this. I diffed it against `master` — it carries the same claim-path deferral this PR introduces (the same `isOpenRoutineExecutionLockConflict` helper and the defer-and-retry on the 23505 conflict), **plus** proactive `SELECT … FOR UPDATE` sibling-lock guards on the recovery/promotion paths. Those guards close the residual cancellation race Greptile flagged here, so #11964 is effectively a superset of this change.

Given that, I'd suggest landing #11964 as the canonical fix rather than rebasing this one or chasing the advisory Greptile check here. I'll leave this PR open for now in case it's useful for merge ordering, and close it once #11964 merges. Thanks for extending the fix, @Oldrich333.

PR #10843, opened 2026-08-04
feat(issues): explain cross-task agent writes with attribution, audit receipts, and actionable denials

cryppadotta:
## Thinking Path

> - Agents write to tasks they do not own. They comment, they change fields, and the control plane now permits this by default for standard-trust agents on any task they can read
> - This makes a task thread ambiguous. A reader sees a comment from an agent that is not the assignee, but no surface says whose authority that write rode
> - The same gap applies to field edits. The activity stream named the verb, but it did not show the before value, the after value, or the reason the write was permitted
> - The remaining refusals are also opaque. An agent that hits a wall receives a 403 with no boundary name, no actor who can act, and no sanctioned path. One real incident spent a full detour to find the workaround
> - This pull request adds the three surfaces that make open cross-task writes legible: an attribution chip, a field-level audit receipt, and an actionable denial contract shared by the API and the UI
> - The benefit is that a reader can answer "who did this, on whose authority, and was it allowed?" on the task itself, and a blocked writer is told what to do next

No public issue exists for this work, so the enhancement is described here.

Cross-task agent writes are permitted, but they are not explained. A task thread can hold comments from agents that are not the assignee, and the activity stream can hold field changes made by those agents. Neither surface names the responsible user behind the write. When a write is refused, the error text does not name the boundary or the way forward.

**Subsystem affected**

Issue detail UI (comment thread and activity stream), the issue write authorization responses in the server, and the shared copy contract that both consume.

**Current behavior**

- An agent comment on a task the agent does not own looks the same as an assignee comment.
- An `issue.updated` activity row states the verb only. It does not show the field-level before and after values, the responsible user, or the authorization reason.
- A refused write returns a short message such as an ownership error. The message does not state which rule fired, who is able to perform the action, or which alternative path is sanctioned.

**Proposed behavior**

- An agent comment on a task the agent does not own carries a chip that reads "for {user}". The chip names the responsible user. Its tooltip states that the author is not the assignee and cannot exceed that user's permissions.
- Each `issue.updated` row shows a receipt: the changed fields with before and after values, the responsible user, and the authorization reason. This applies to board edits as well as agent edits.
- Each refusal states three things: the boundary that fired, who is able to act, and the sanctioned path. The API error body and the in-app notice use the same words, because both read one shared contract.

Related pull requests, found by searching this repository:

- Refs #10837 — merged. It added the default-open cross-task write rule, the comment attribution data, and the per-run containment cap that this pull request makes visible.
- Refs #10114 — open. It proposes a narrower authorization change in the same area.
- Refs #7998 — open. It proposes append-only cross-assignee comments as an alternative to opening writes.

## What Changed

- Adds `packages/shared/src/issue-write-denial.ts`. This is one copy contract for eight ways an issue write can be refused: not visible, responsible-user ceiling, responsible user unavailable, excluded actor class, assignee run lock, per-run cross-task cap, missing run context, and rejected attribution. Each entry names the boundary, who can act, and the sanctioned path.
- Maps server authorization decisions onto that contract in `server/src/routes/issues.ts` and `server/src/services/cross-issue-influence-limit.ts`. The flattened `error` string carries all three obligations, and `details.code` lets the UI render the same words. The two cap codes keep the names they already ship under.
- Adds `CommentAttributionChip`. It renders "for {user}" beside the author name on agent comments where the author is not the assignee. It renders nothing when no responsible user is recorded, so older rows stay clean. It is wired into both `IssueChatThread` and the flagged `TaskChatThread` redesign.
- Adds `IssueFieldChangeReceipt`. It renders the change receipt under `issue.updated` rows in the activity stream. Ids resolve to agent and user names where the directory is loaded. Server-truncated text is labelled as a preview, so the receipt never implies that it shows a whole value.
- Adds `IssueWriteDenialNotice`. It renders the shared copy in the app, keyed off the denial events the server logs on a task.
- Adds a public `/ux-lab/cross-issue-collaboration` page. It renders all three surfaces and their edge cases for review without a seeded thread. This follows the existing `ux-lab` pages.

## Verification

Automated, all green:

```
pnpm --filter @paperclipai/shared exec vitest run src/issue-write-denial.test.ts        # 17 tests
pnpm --filter @paperclipai/ui exec vitest run src/components/IssueWriteDenialNotice.test.tsx \
  src/components/IssueFieldChangeReceipt.test.tsx src/components/CommentAttributionChip.test.tsx \
  src/lib/issue-change-receipt.test.ts src/lib/comment-attribution.test.ts                # 46 tests
pnpm --filter @paperclipai/server exec vitest run src/__tests__/cross-issue-influence-limit.test.ts \
  src/__tests__/issue-comment-attribution-audit-routes.test.ts \
  src/__tests__/low-trust-red-team-routes.test.ts                                        # 98 tests
```

`tsc --noEmit` passes for the shared, ui, and server packages.

Manual, in a browser:

1. Start the UI only: `pnpm --filter @paperclipai/ui exec vite`.
2. Open `/ux-lab/cross-issue-collaboration`. No session is needed, because `ux-lab` routes are public.
3. All three surfaces were captured at 1440x900 in light mode and dark mode, and at 390x844. The page reported no errors.
4. The chip tooltip was opened by a hover and by a keyboard focus.

Rendering the page found defects that the tests had missed. Three copy and contrast defects were fixed, and two of them are now pinned by a test. A design review then found three layout defects, which are also fixed: the denial notice orphaned its label when a value wrapped, the receipt icon wrapped onto its own line at narrow widths, and the chip tooltip was reachable by hover only.

## Risks

Low risk, and additive.

- Every new surface renders nothing when its data is absent. Comments without a recorded responsible user show no chip, and activity events without a receipt show no receipt, so existing rows do not change.
- No migration is included. The data these surfaces read already ships.
- The wire values of the two per-run cap denial codes are unchanged. Only the human-readable text changes, plus six codes that had no `details.code` before.
- The denial copy is read by agents as well as people. If wording must change later, one shared module is the only place to change it.
- Roadmap check: this extends the completed "Activity log & action attribution" area rather than duplicating planned core work.

## Model Used

Claude Opus 5 (Anthropic), model id `claude-opus-5[1m]`, 1M context window, extended thinking, with tool use and code execution. It ran as an agent in Claude Code and drove a real browser to capture the review screenshots.

## Checklist


cryppadotta:
Thanks — the first half of that finding was correct and is now fixed. The second half I am declining, with reasoning.

**Fixed: `align-[-0.125em]` → `align-(--va-0_125em)`** (`7517b18`). You were right that the token already existed; it was extracted from the identical baseline nudge in `ExternalObjectStatusIcon.tsx`. I confirmed the swap is behavior-identical in a browser: computed `vertical-align` is `-1.375px` before and after, and `display` stays `inline`. `node scripts/check-token-gates.mjs` now reports **all three gates clean** — it did flag this exact line before the fix, so this was a real violation.

**Declining: `[border-radius:14px_14px_14px_4px]` on the UX-lab page.** That declaration is a deliberate, byte-for-byte copy of the shipping chat bubble at `ui/src/components/IssueChatThread.tsx:1912`, which carries the same literal on master today. The lab page exists so reviewers can see the real surfaces without seeding a live thread, so its bubble has to render at the production radius — tokenizing only the copy would make the lab drift from the component it stands in for.

Two further reasons not to change it here:

1. The token gate does not flag it. `check-token-gates.mjs` reports it clean, and the allowlist block in `ui/src/index.css` explicitly exempts `*UxLab.tsx` demo surfaces for first-party one-off values (DECISION-SHEET.md B1).
2. The correct fix is to tokenize the *source*, not the copy. That means changing the shipping bubble's radius declaration, which is a design-system change to a production surface. It belongs in its own PR with design review, not in a PR about denial copy.

Happy to open that follow-up if a maintainer wants the bubble radius minted as a token.

PR #10847, opened 2026-08-04
fix(heartbeat): durably record wakeup skips and assignment failures

valkyriweb:
## Summary
- `enqueueWakeup` (both issue-scoped and general queuing paths) called `requireQueuedResponsibleUserId(...)` inside the `db.transaction`, which threw on unresolved responsible-user and rolled back the just-inserted `agentWakeupRequests` "queued" row — the skip was thrown as a 422 but never durably recorded. Both paths now write a `status: "skipped"`, `reason: "responsible_user_unresolved"` row before re-throwing, matching the existing durable-skip pattern already used for `company.inactive`, `agent.not_invokable`, and `budget.blocked`.
- `queueIssueAssignmentWakeup` (fire-and-forget from `routes/issues.ts` and `services/routines.ts`) only logged rejected wakeups. It now durably inserts a `status: "failed"`, `reason: "issue_assignment_wakeup_failed"` `agentWakeupRequests` row before logging/optionally rethrowing, so lost assignment wakeups are queryable instead of silently swallowed. Callers keep the existing `void queueIssueAssignmentWakeup(...)` fire-and-forget call and response semantics unchanged.
- Verified `wakeOnDemand: false` is honored (pre-existing, now covered by a test) and that `maxConcurrentRuns` correctly keeps a second run queued rather than promoting it while a slot is held (pre-existing, now covered by a test).

## Root cause (SMI-3437 class of bug)
Drizzle transactions roll back all writes made via the transaction handle when the callback throws. `requireQueuedResponsibleUserId` threw *inside* the same transaction that had just inserted the `agentWakeupRequests` "queued" row, destroying that row along with the failure. The HTTP caller still got a 422, but there was no durable trail explaining why the assignment wake never produced a run.

## Test plan
New/extended deterministic tests (embedded Postgres):
- `heartbeat-responsible-user-invariant.test.ts` — extended to assert a durable `skipped`/`responsible_user_unresolved` row survives the failed dispatch.
- `heartbeat-wakeup-skip-and-coalescing.test.ts` (new) — `wakeOnDemand: false` skip persistence; assignment-wakeup failure durability; coalescing a wake into an already-running run instead of losing it; serializing concurrent wakes for the same agent/issue without duplicating or losing the run.
- `heartbeat-assignment-wakeup-window-and-manual-skip.test.ts` (new) — assignment wakes bypass the timer window when `wakeOnDemand=true`; a running agent's assignment wake for a different issue stays queued under `maxConcurrentRuns=1` instead of being dropped; the wakeup route returns the precise stored skip reason instead of the generic message.
- `issue-assignment-wakeup.test.ts` (new) — `queueIssueAssignmentWakeup` durability directly, including the `rethrowOnError` path and the backlog no-op path.

Verified:
- `pnpm run typecheck` — clean.
- Owning suite (8 files, 141 tests) — passing, stable across repeated runs.
- Broad regression sweep across `heartbeat-*`, `issue-*`, `routines-*`, `agent-*` test files (80 files, 1030 tests) — all passing.

Not merged or deployed. No production data touched — SMI-3437 was not looked up against production state; this fix addresses the durable-recording root cause traced from the code path itself.


valkyriweb:
Opened against the wrong base repo by mistake (should target valkyriweb/paperclip, our fork). Closing; correct PR to follow.

PR #10857, opened 2026-08-04
fix(server): exclude the requesting run and its queued wakeups from task-watchdog source mutation revalidation

santhiprakash:
## Thinking Path

> - Paperclip is a control plane for AI-agent companies; task watchdogs verify stopped issue subtrees and put them back into motion.
> - The watchdog server gives a watchdog agent a bounded ability to reassign or add a monitor to a source issue inside the watched subtree.
> - Before the mutation is applied, the server revalidates the watchdog's stop fingerprint against the current subtree state.
> - The revalidation scans for live runs by looking at `heartbeatRuns` rows, at `issues.executionRunId` joined to live runs, and at queued `agentWakeupRequests`.
> - When the watchdog's own run had just checked out or reassigned the source issue, the source's `executionRunId` pointed back at the same run, so the revalidation saw a live path and rejected the next mutation with a stale-review error.
> - After the source run was excluded, a watchdog that scheduled a follow-up wakeup (for example, an `in_review` monitor) could still see that queued wakeup, or a heartbeat run spawned from it, as an external live path when it tried to mutate the source again.
> - This pull request makes the revalidation exclude the requesting run and any wakeups or spawned runs it created, so a watchdog can reassign a source and set a one-shot monitor in the same run without treating itself as an external live path.
> - The benefit is that watchdogs can perform atomic recovery batches that include reassignment and monitor setup, rather than being forced to stop after the first mutation or file a product bug.

No public GitHub issue exists for this bug. The in-PR description follows the bug-report template.

**What happened?**
A task-watchdog run that reassigned a watched source issue was then unable to schedule a persisted issue monitor on that same source. The next `PATCH` request returned `Task-watchdog review is stale because the watched subtree now has a live, waiting, already-reviewed, or not-applicable path; refresh the source state before mutating it.` with `currentState: live`, even though the only live execution paths came from the watchdog's own run.

**Expected behavior**
A task-watchdog run should be able to reassign a source issue and schedule a monitor in a follow-up request without its own heartbeat or its own queued wakeups being treated as concurrent live paths.

**Steps to reproduce**
1. Create an issue with a task watchdog configured.
2. Let the watchdog issue reach a stopped state and capture the stop fingerprint.
3. From a watchdog run, reassign the source issue (which sets `issues.executionRunId` to the current run).
4. In the same run, attempt to set an `in_review` monitor on the source.
5. The request is rejected with the stale-review live-path error.

`master` at `c54936e2e958b1574c88adfd7dbf1d6a7dc440ad`

## What Changed

- `server/src/services/task-watchdogs.ts`
  - `collectClassifierInput` now accepts an optional `excludeRunId`.
  - The live-run and issue-execution-run queries both exclude the given `runId` when it is provided.
  - `agentWakeupRequests` requested by the same run are also excluded from the queued-wake live-path check.
  - `heartbeatRuns` whose `wakeupRequestId` was requested by the same run are also excluded, so a spawned run from the watchdog's own wakeup does not look like an external live path.
  - `revalidateMutationScope` accepts an optional `runId` and passes it to `collectClassifierInput`.
- `server/src/services/task-watchdog-scope.ts`
  - `resolveTaskWatchdogMutationScope` now returns the run's `id` as `runId` in the watchdog scope.
- `server/src/routes/issues.ts`
  - Propagates `requestedByRunId` on issue-assignment and dependency-resolved wakeups so the revalidation can trace them back to the requesting run.
  - Persists `requestedByRunId` on `agentWakeupRequests` rows and passes it through the wakeup/coalesce/defer paths.
- `packages/db/src/schema/agent_wakeup_requests.ts` and `packages/db/src/migrations/0212_yellow_lord_tyger.sql`
  - Adds the `requested_by_run_id` column.
- `packages/shared/src/types/heartbeat.ts`
  - Adds `requestedByRunId` to the `AgentWakeupRequest` shared type.
- `server/src/__tests__/task-watchdogs-scheduler.test.ts`
  - Added a test that verifies the same state is reported as `live` when no `runId` is excluded, and `stopped` and allowed when the requesting run is excluded.
  - Added a test that verifies a queued own-wakeup and its spawned run are ignored during revalidation.
- `server/src/__tests__/issue-agent-mutation-ownership-routes.test.ts`
  - The `task watchdog scope grants` block continues to cover comment, transition, reassignment, and stale-revalidation behavior for watchdog runs.

## Verification

- `pnpm --filter @paperclipai/server --filter @paperclipai/db --filter @paperclipai/shared typecheck` passed.
- `npx vitest run src/__tests__/task-watchdogs-scheduler.test.ts` passed (20 tests).
- `npx vitest run src/__tests__/issue-agent-mutation-ownership-routes.test.ts -t "task watchdog scope grants" --testTimeout 30000` passed (13 tests).

## Risks

- Low. The change is scope-local to task-watchdog mutation revalidation. The scheduler/evaluator path does not pass a `runId`, so it is unchanged.
- A future caller of `revalidateMutationScope` must pass the correct `runId` if it wants the self-run and self-wakeup exclusion; the field is optional to keep existing tests and callers working.
- The migration adds a nullable column with no default; it is backward-compatible because existing rows simply have a null `requestedByRunId` and the revalidation logic treats null as "not requested by the excluded run".

## Model Used

- Anthropic Claude via Devin (SWE-1.7, tool use).

## Checklist


santhiprakash:
Rebased onto current `upstream/master` and regenerated the Drizzle migration/snapshot for the new `requested_by_run_id` column. All CI checks pass. Ready for maintainer review/merge.

Issue #9313, opened 2026-07-09
Heartbeat setup retries WorkspaceValidationFailure every ~3s with no backoff/cap -> control-plane OOM (2026.707.0)


Greenbauer:
**Version:** paperclipai 2026.707.0 (npm), Node control plane running with `--max-old-space-size=10240`.

## Summary

When an `issues` row has `execution_workspace_preference = 'reuse_existing'` and `execution_workspace_id` pointing at a **non-active** `execution_workspaces` row (archived out-of-band), every heartbeat evaluating that issue throws `WorkspaceValidationFailure` ("requested inherited execution workspace reuse ... could not be restored", thrown from `provisionExecutionWorkspaceForFreshnessDecision` in `heartbeat.js`) — and the runtime **retries the setup every ~3 seconds with no backoff and no retry cap**. Each attempt leaks memory; within an hour the process hits `FATAL ERROR: Ineffective mark-compacts near heap limit` and OOMs. After restart it immediately re-enters the same loop.

## Observed impact (one afternoon, 3 poisoned issue rows)

- 5 heap-OOM crashes of the 10 GB control plane in ~3 hours, accelerating (59 → 20 min-to-OOM).
- As the heap neared the ceiling, 1.5–2.5 s GC pauses stalled the event loop → dashboard timeouts.
- Log storm signatures per crash cycle: ~19,000 `skipping late run finalization`, thousands of `heartbeat execution setup failed`, plus `agent start lock timed out` storms with up to 8 concurrent duplicated starts of the same agent.
- Zero OOMs in the preceding 14 days; heap growth occurred with zero dashboard polling, so the retry loop is the driver.

## Repro

1. Attach an issue to a workspace: `execution_workspace_preference = 'reuse_existing'`, `execution_workspace_id = <workspace uuid>`.
2. Archive that `execution_workspaces` row without going through the close endpoint (manual ops / cleanup script), leaving the issue's pointer in place.
3. Watch heartbeat: `WorkspaceValidationFailure` → setup retried every ~3 s indefinitely.

## Suggested fixes

1. **Backoff + cap** on heartbeat execution-setup retries (exponential, and/or a per-issue circuit breaker) — no failure class should retry at ~3 s forever.
2. **Self-heal the pointer**: after N consecutive `WorkspaceValidationFailure`s for the same issue, detach it (`execution_workspace_id = NULL`, preference NULL) or quarantine it, instead of retrying — the workspace is provably not restorable.
[...]

PR #10865, opened 2026-08-05
fix(server): ignore req/res in the file log transport to stop unbounded thread-stream growth

### LINKED ISSUE #1825: Memory leak: pino-pretty file transport ThreadStream buffer grows unbounded, OOMs every ~60min
## Summary

The Paperclip server crashes with `FATAL ERROR: Reached heap limit Allocation failed - JavaScript heap out of memory` approximately every 60 minutes due to a memory leak in the pino logging pipeline.

## Root Cause

In `@paperclipai/server/dist/middleware/logger.js`, the file transport target (line 34-36) does not ignore `req,res,responseTime` fields, while the stdout transport (line 30) does:

```js
// stdout transport — ignores req,res,responseTime ✅
{ target: "pino-pretty", options: { ...sharedOpts, ignore: "pid,hostname,req,res,responseTime", ... }, level: "info" },

// file transport — does NOT ignore req,res,responseTime ❌
{ target: "pino-pretty", options: { ...sharedOpts, colorize: false, destination: logFile, ... }, level: "debug" },
```

Every HTTP request serializes the full `req` and `res` objects (including all headers, ~1.5KB per entry) into the file transport's pino-pretty ThreadStream buffer. With multiple agents heartbeating and UI polling, the `pino-pretty` formatter can't keep up with the write rate. The ThreadStream internal buffer (`this[kImpl].buf += data` in thread-stream/index.js) grows without bound because the drain trigger only fires near `MAX_STRING_LENGTH` (~1GB), but V8's heap OOMs at ~2GB first.

## Evidence

- Server log file reached 372MB
- Heap grows to ~2026MB before GC becomes ineffective (`average mu = 0.118, current mu = 0.032`)
- Consistent ~60 minute crash cycle (longer during quiet periods, shorter during high activity)
- 32 restarts in one day via systemd `Restart=on-failure`
- Agents see "process lost -- server may have restarted" on every crash

## Environment

- paperclipai v2026.318.0
- Node.js v24.13.1
- Ubuntu 24.04, 64GB RAM
- 4 companies, ~20 agents total, heartbeat intervals 600-1200s

## Suggested Fix

Add `ignore: "pid,hostname,req,res,responseTime"` to the file transport options (matching stdout), or switch the file transport from `pino-pretty` to `pino/file` (raw JSON) which is orders of magnitude faster and avoids the formatting bottleneck entirely.

Also consider raising the file transport level from `debug` to `info` to match stdout.

## Workaround

Manually patching the installed `logger.js` to add the ignore list to the file transport and raising `NODE_OPTIONS=--max-old-space-size=3072`.

jmiramant:
## Thinking Path

> - The server logs through pino with two pino-pretty transport targets: one to stdout, one to a file.
> - The stdout target ignores `pid,hostname,req,res,responseTime`. The file target does not.
> - The file target therefore prettifies the full `req`/`res` objects for every HTTP log entry. This makes the formatter worker slower than the log producers.
> - thread-stream's send buffer then grows without bound, and the server aborts at the V8 heap limit (`FATAL ERROR: Reached heap limit`). The crash cycle is roughly 60 minutes at default heap, and roughly 14 hours at `--max-old-space-size=4096`.
> - This pull request gives the file target the same ignore list the stdout target already has.
> - The benefit is that the server stops leaking heap under normal HTTP load and no longer crash-loops.

- Fixes #1825
- Refs #9313 (a different OOM path; the two compound on busy deployments)

Note: an earlier comment on #1825 says the bug was "fixed in commit eaf888c". That commit does not exist in this repository, and the missing ignore list is still present on `master` (`server/src/middleware/logger.ts`). The bug reproduces on 2026.707.0 (per a later comment on #1825) and on 2026.722.0 (our deployment).

## What Changed

- `server/src/middleware/logger.ts`: the file pino-pretty target now sets `ignore: "pid,hostname,req,res,responseTime"`, matching the stdout target. One line, plus a comment that explains why the ignore list must stay.

## Verification

- New regression test `server/src/__tests__/logger-file-transport-ignore.test.ts`: asserts every pino-pretty target ignores `req`, `res`, `responseTime`. Verified it fails against the pre-fix configuration and passes with this change.
- `npx vitest run __tests__/http-log-policy.test.ts __tests__/http-log-redaction.test.ts __tests__/log-redaction.test.ts __tests__/logger-tz.test.ts` — 13 tests pass.
- `pnpm typecheck` passes.
- Production observation (30-agent deployment, ~2,000 runs/day, paperclipai 2026.722.0): server RSS grows ~250 MB/hour and the process aborts at the heap ceiling. Hourly RSS samples: 685 MB fresh → 3,185 MB at 9.8 h → SIGABRT at ~4,081 MB with `--max-old-space-size=4096`. The stdout transport, which has this ignore list, shows no such growth. We have applied this exact change to the installed dist on that deployment and will report the post-patch RSS curve on #1825.

## Risks

- Low risk. The file log loses the prettified `req`/`res`/`responseTime` fields on debug-level HTTP entries — the same fields stdout already omits. Warn/error HTTP entries keep their structured request excerpts through the existing serializer paths. No API, schema, or behavior change beyond log file content.

## Model Used

- Claude Fable 5 (Anthropic, `claude-fable-5`, 1M context) via Claude Code CLI — diagnosis, patch, and tests. Human-reviewed before submission.

## Checklist

### DIFF
--- a/server/src/__tests__/logger-file-transport-ignore.test.ts
+++ b/server/src/__tests__/logger-file-transport-ignore.test.ts
@@ -1,0 +1,71 @@
+import { describe, expect, it, vi, beforeEach } from "vitest";
+
+/**
+ * Regression test for https://github.com/paperclipai/paperclip/issues/1825
+ *
+ * The stdout pino-pretty target ignores `req,res,responseTime`; the file
+ * target historically did not. Prettifying the full req/res objects for
+ * every HTTP entry makes the file formatter worker slower than the log
+ * producers, so thread-stream's send buffer grows without bound and the
+ * server aborts at the V8 heap limit.
+ *
+ * We verify that EVERY pino-pretty transport target ignores the large HTTP
+ * fields, so the two targets cannot silently diverge again.
+ */
+
+const mockTransport = vi.hoisted(() => vi.fn(() => ({ write: vi.fn() })));
+const mockPino = vi.hoisted(() => {
+  const fn = vi.fn(() => ({
+    info: vi.fn(),
+    debug: vi.fn(),
+    error: vi.fn(),
+    warn: vi.fn(),
+    child: vi.fn(),
+  }));
+  (fn as any).transport = mockTransport;
+  return fn;
+});
+
+// Mock fs so the module-level mkdirSync call is a no-op in tests.
+vi.mock("node:fs", async (importOriginal) => {
+  const actual = await importOriginal<typeof import("node:fs")>();
+  return { ...actual, mkdirSync: vi.fn() };
+});
+
+vi.mock("pino", () => ({
+  default: mockPino,
+}));
+vi.mock("pino-http", () => ({
+  pinoHttp: vi.fn(() => vi.fn()),
+}));
+vi.mock("../config-file.js", () => ({
+  readConfigFile: vi.fn(() => null),
+}));
+vi.mock("../home-paths.js", () => ({
+  resolveHomeAwarePath: vi.fn((p: string) => p),
+  resolveDefaultLogsDir: vi.fn(() => "/tmp/paperclip-test-logs"),
+}));
+
+describe("logger transports ignore large HTTP fields (issue #1825)", () => {
+  beforeEach(() => {
+    vi.clearAllMocks();
+  });
+
+  it("every transport target ignores req, res, and responseTime", async () => {
+    await import("../middleware/logger.js");
+
+    expect(mockTransport).toHaveBeenCalledOnce();
+    const { targets } = mockTransport.mock.calls[0][0] as {
+      targets: Array<{ options: Record<string, unknown> }>;
+    };
+    expect(targets.length).toBeGreaterThanOrEqual(2);
+
+    for (const target of targets) {
+      const ignore = String(target.options.ignore ?? "");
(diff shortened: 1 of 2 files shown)


jmiramant:
Closing this as superseded by current `master` — the code it patches no longer exists.

This PR added `ignore: "pid,hostname,req,res,responseTime"` to the file-log `pino-pretty` target inside the `targets: [...]` array in `server/src/middleware/logger.ts`. On current `master` that array is gone:

- In production, `logger` is now a plain `pino({ level, redact })` with **no transport at all**, so there is no file formatter worker and no thread-stream buffer to grow.
- In development there is a single `pino-pretty` target writing to stdout (`destination: 1`), and it already carries exactly `ignore: "pid,hostname,req,res,responseTime"`.

So the unbounded thread-stream growth this addressed is no longer reachable, and the one remaining target already has the fix.

I am not re-filing the `logger-file-transport-ignore.test.ts` invariant either. It asserts `targets.length >= 2` and that the two targets cannot diverge, but `master` now has a single target and none in production — there is nothing left to diverge. Re-filing it would mean writing a new test against a structure that no longer exists, for an invariant that no longer has a failure mode.

No action needed from maintainers.

PR #10872, opened 2026-08-05
feat(opencode-local): default OPENCODE_ALLOW_ALL_MODELS=1 fleet-wide (BLU-26616)

jmiramant:
## Thinking Path

> - The `opencode_local` adapter executes OpenCode in a subprocess and must confirm the configured model is available before starting
> - `ensureOpenCodeModelConfiguredAndAvailable` resolves the model against `opencode models`, which intermittently serves a live-refreshed list that can omit new or edge models when a live fetch is unavailable
> - This can turn a valid model (e.g. `openrouter/z-ai/glm-5.2`) into a spurious `adapter_failed` that reads like a model deprecation — exactly the class of error the model-changes-require-board-approval policy warns against
> - The `OPENCODE_ALLOW_ALL_MODELS=1` flag bypasses the catalog check (provider/model format is still enforced), removing the catalog from the resolution critical path
> - Two agents already carried this flag in persisted config; setting it fleet-wide as a soft default removes the catalog dependency for all `opencode_local` runs
> - The benefit is spurious `adapter_failed` errors from transient catalog misses can no longer block agent heartbeats fleet-wide

No public GitHub issue. Underlying problem (bug):

**What happened?**
`opencode_local` adapter runs sporadically failed with `Configured OpenCode model is unavailable — adapter_failed` even when the configured model was valid. The `opencode models` catalog intermittently omits new or edge models during live-refresh failures, causing the model probe to reject a valid model and abort the run.

**Expected behavior**
A valid model should not produce an `adapter_failed` rejection due to a transient catalog unavailability. The catalog check is a convenience probe, not an authoritative source of model validity; a live-fetch miss should not abort agent heartbeats.

**Steps to reproduce**
1. Configure an `opencode_local` agent with an edge or recently-added model (e.g. an OpenRouter model)
2. Observe intermittent `adapter_failed: Configured OpenCode model is unavailable` when the catalog live-fetch misses the model
3. Observe the agent's heartbeat is aborted for the duration of the transient miss

master at time of PR

**Deployment mode**
Self-hosted server

## What Changed

- `server/src/adapters/opencode-local/` — new exported helper `applyOpenCodeAllowAllModelsDefault(env)` applies `OPENCODE_ALLOW_ALL_MODELS=1` as a **soft default** after the `adapterConfig.env` merge. Applied alongside the existing `OPENCODE_DISABLE_PROJECT_CONFIG` guard; flows into `preparedRuntimeConfig.env` so both local and remote probe paths honor it. A persisted config value wins; setting `OPENCODE_ALLOW_ALL_MODELS=0` in process env disables the default.
- Unit tests for `applyOpenCodeAllowAllModelsDefault` (default-on, config-override, process-env-defer); two remote-probe-path tests now opt back into the probe explicitly (`config.env.OPENCODE_ALLOW_ALL_MODELS = "0"`) since skipping is now the default.
- Documented in the adapter's `agentConfigurationDoc`.

Note: this repo has no `staging` branch (master is trunk), so this PR targets `master` directly, following the repo's branch reality.

## Verification

```bash
pnpm --filter @paperclipai/server vitest run src/adapters/opencode-local/
```

39 tests pass. `pnpm typecheck` clean. The two updated remote-probe tests pass with the explicit opt-out flag.

## Risks

Low risk. Env-flag-only change — not a model or adapter type change. Overridable per-agent via `adapterConfig.env.OPENCODE_ALLOW_ALL_MODELS = "0"` and per-deployment via process env. Provider/model format validation is preserved (only the catalog check is skipped). The two agents that already carried the flag in persisted config are unaffected — persisted config wins over the soft default.

## Model Used

Claude — claude-opus-4-8, extended thinking, tool use (Claude Code).

## Checklist


jmiramant:
**Greptile P1 on `execute.ts:96` — confirmed and fixed in `e3bbb92`.**

The finding is correct. `applyOpenCodeAllowAllModelsDefault` yielded to an explicit process-env value by returning *without writing the key*. A remote (SSH/sandbox) target does not inherit the host process env — the run env map is the only channel it gets — so a deployment that set `OPENCODE_ALLOW_ALL_MODELS=1` on the host had that override dropped on exactly the paths that cannot recover it. The local path masked the bug because it already reads `env.X ?? process.env.X`.

Fixed by copying the value through rather than skipping the write, per the suggestion:

```ts
if (processEnv.OPENCODE_ALLOW_ALL_MODELS !== undefined) {
  env.OPENCODE_ALLOW_ALL_MODELS = processEnv.OPENCODE_ALLOW_ALL_MODELS;
  return;
}
```

**Disable semantics are unchanged.** `isTruthyEnvFlag` accepts only `true`/`1`/`yes`, so a copied `"0"` is behaviourally identical to an absent key — the operator's `OPENCODE_ALLOW_ALL_MODELS=0` escape hatch still works, and now it reaches remote targets too instead of silently applying only locally.

**Tests.** The existing case `defers to an explicit process-env value…` asserted `toBeUndefined()` — it encoded the defect, so it was rewritten rather than added around. Both directions are now pinned:

- `propagates an explicit process-env disable so remote targets honour it` → `"0"`
- `propagates an explicit process-env enable so remote targets honour it` → `"1"`

Both were confirmed failing (`expected undefined to be '0'` / `'1'`) before the source change, then passing after. Full package suite green: **40 passed (6 files)**, `tsc --noEmit` clean.

The doc comment above the function was also corrected — it claimed the function would "no-op when the process env sets it explicitly", which is the behaviour that caused this.

PR #10879, opened 2026-08-05

AaryanCdry:
## Summary

```ts
if (!gitErrorIncludes(attachError, "already checked out") \&\& !gitErrorIncludes(attachError, "already used by worktree")) {
```

```ts
const mergedEnv = opts?.env ? { ...process.env, ...opts.env, LC_ALL: 'C', LANG: 'C', LANGUAGE: 'C' } : { ...process.env, LC_ALL: 'C', LANG: 'C', LANGUAGE: 'C' };
```
Note: C locale is set LAST so it cannot be overwritten by caller-provided env vars.

## Testing

- Added `git-error-includes.test.ts` unit tests covering both old and new git error messages
- Tests verify `gitErrorIncludes` matches both message patterns
- All git call sites now force C locale

## Checklist

- [x] Tests pass (new unit test covers git 2.42+ error message matching)
- [x] No breaking changes

### DIFF
--- a/server/src/__tests__/git-error-includes.test.ts
+++ b/server/src/__tests__/git-error-includes.test.ts
@@ -1,0 +1,20 @@
+import { describe, expect, it } from "vitest";
+import { gitErrorIncludes } from "../services/workspace-runtime.js";
+
+describe("gitErrorIncludes", () => {
+  it("matches old git 'already checked out' message", () => {
+    const err = new Error("fatal: workspace already checked out at /path/to/worktree");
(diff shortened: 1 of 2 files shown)


AaryanCdry:
Test updated with unit tests for `gitErrorIncludes`. Should resolve the Greptile review comment about the test bypassing the changed guard.

AaryanCdry:
Closing in favor of #10880 (fixed gitErrorIncludes unit test).

PR #10895, opened 2026-08-05
fix: use npm install instead of missing npx install command

### LINKED ISSUE #10739: Quickstart install.sh is unusable on macOS: paperclip.ing serves a stale script, and `paperclipai@latest install` is an unknown command

fgava77:
Fixes #10739

### What happened?
The `install.sh` script relies on `npx --yes paperclipai@latest install` to install the `paperclipai` binary. However, the `install` subcommand was removed from recent versions of the `paperclipai` CLI package, causing `install.sh` to fail with `error: unknown command 'install'`.

### Expected behavior
The installation script should successfully place the `paperclipai` CLI executable in the user's `PATH` (typically `~/.local/bin`) and proceed to launch the `onboard` wizard.

### Steps to reproduce
1. Run `bash install.sh` locally on a fresh machine.
2. The script errors when delegating to the Paperclip CLI using the `install` subcommand.

### Paperclip version
2026.722.0

### Node.js version
v24.18.0

### Operating system
Linux (x64)

## Thinking Path
The error indicates that the `install` subcommand was removed from the `@latest` tag of the `paperclipai` package. Since the bash script's purpose is to fetch and install the CLI globally so it can be run from the user's PATH, changing the `npx` command to a direct `npm install -g --prefix ~/.local` serves the same goal and avoids the missing subcommand. We also needed to update the Docker test assertions since `npx --yes` is no longer used, replacing it with `npm -g --prefix` verifications.

Dropping `npx` shifted responsibility for PATH setup onto the script itself, which turned `update_path` into the substantive part of this change. The rule it now encodes: shells read two classes of startup file — login shells read a *login file*, non-login interactive shells read an *rc file* — and covering only one class leaves the other kind of session unable to resolve the CLI.

## What Changed
- Replaced the failing `npx --yes paperclipai@latest install` invocation in `scripts/install.sh` with `npm install -g --prefix "${HOME}/.local"`.
- Updated `scripts/test-install-sh-docker.sh` to assert the arguments passed to `npm` rather than `npx`, and cloned the `npx` test fixture to an `npm` fixture in `scripts/install-sh-fixtures` (both fixtures are now covered by shellcheck).
- Reworked `update_path` so every reachable shell configuration ends up with `~/.local/bin` on `PATH`:
  - It patches all common startup files that already exist (`.bashrc`, `.zshrc`, `.bash_profile`, `.bash_login`, `.profile`, `.zprofile`).
  - It then creates a missing file per *class*, chosen from `$SHELL`: zsh gets `~/.zshrc` (zsh reads it for every interactive shell, login included, so it covers both classes; a login-only `.zprofile` must not suppress it); bash gets `~/.profile` when no login file exists **and** `~/.bashrc` when that is missing, since bash login shells never read `.bashrc`; plain sh/dash get only `~/.profile`, having no interactive rc file.
  - Persistence is no longer gated on the installer's own `$PATH`. The process may already export `~/.local/bin` (inherited from a parent, or left over from an earlier install) while the user's startup files still lack it — the previous guard skipped every update in that state. Idempotency comes from the per-file `grep` instead, so nothing is duplicated.

## Verification
- Ran the `install.sh` script locally on an Ubuntu x64 machine, confirming that it successfully installs `paperclipai` into `~/.local/bin` and launches the `paperclipai onboard` wizard without the subcommand error.
- Ran `bash scripts/test-install-sh-docker.sh` locally: shellcheck is clean and the full suite passes, including the new `update_path` cases.
- Drove the real `update_path` directly across a shell × startup-file matrix (bash/zsh/dash × empty home, `.bashrc`-only, `.bash_profile`-only, `.bash_login`-only, `.profile`-only, `.zprofile`-only, `.zshrc`-only, and target-already-on-`PATH`), asserting both classes of coverage in each. Additionally verified end to end that sourcing each generated file in a clean `env -i` shell actually resolves a `paperclipai` binary placed in `~/.local/bin`.
- Confirmed the new cases are genuine regression tests by replaying the previous `update_path` against the same matrix: it produced 7 failures — no `.bashrc` for any bash home lacking one, and no persistence at all when `~/.local/bin` was already on the process `PATH`. The same two states were replayed through the real installer in Docker with the pre-fix script, reproducing the missing files there too.
- Re-ran the idempotency case with both classes pre-seeded: startup files come out byte-identical across repeat installs, and the installer logs nothing when there is no work to do.

## Risks
- Users who expect a different global installation location for `npm install -g` might end up with the binary strictly inside `~/.local/bin`. However, the script already checked `~/.local/bin/paperclipai` as a fallback, so this aligns with the script's previous behavior.
- The `--install-service` functionality, if it was also removed from the CLI package, will now just pass through to `npx` as `service install`. If it fails, that's a separate issue. We kept it unmodified to avoid changing unrelated code.
- `update_path` may now create `~/.profile`, `~/.bashrc` (POSIX/bash), or `~/.zshrc` (zsh) on systems that previously had no such file. This is additive — it only ever appends an `export PATH` line, never overwrites an existing file — and matches the convention used by rustup and Homebrew.
- A bash user whose `.bash_profile` sources `.bashrc` will get `~/.local/bin` prepended twice in a login shell. Duplicate `PATH` entries are harmless, and the alternative (guessing whether a login file sources the rc file) is fragile.
- Persistence now runs on every install rather than only when `~/.local/bin` is absent from `PATH`. Repeat installs are a no-op on an already-covered home, which the idempotency test asserts byte-for-byte.

## Model Used
Gemini 3.1 Pro (High); the `update_path` rework was done with Claude Opus 5.

- [x] I have searched the GitHub PR list for similar PRs

### DIFF
--- a/scripts/install-sh-fixtures/npm
+++ b/scripts/install-sh-fixtures/npm
@@ -1,0 +1,15 @@
+#!/usr/bin/env bash
+set -euo pipefail
+
+: "${PAPERCLIP_INSTALL_TEST_LOG:?PAPERCLIP_INSTALL_TEST_LOG is required}"
+node --version >"${PAPERCLIP_INSTALL_TEST_LOG}.node"
+{
+  printf 'NPM_CONFIG_REGISTRY=%s\n' "${NPM_CONFIG_REGISTRY:-}"
+  printf 'npm_config_registry=%s\n' "${npm_config_registry:-}"
+  printf 'NPM_CONFIG_USERCONFIG=%s\n' "${NPM_CONFIG_USERCONFIG:-}"
+  printf 'npm_config_userconfig=%s\n' "${npm_config_userconfig:-}"
+  if [ -n "${NPM_CONFIG_USERCONFIG:-}" ] && [ -f "$NPM_CONFIG_USERCONFIG" ]; then
+    sed 's/^/npmrc:/' "$NPM_CONFIG_USERCONFIG"
+  fi
+  printf '%s\n' "$@"
+} >>"$PAPERCLIP_INSTALL_TEST_LOG"
--- a/scripts/install.sh
+++ b/scripts/install.sh
@@ -268,24 +268,24 @@ update_path() {
       fi
     done
 
-    # Bash login shells read the first existing file among .bash_profile,
-    # .bash_login, .profile; zsh login shells read .zprofile. When none of those
-    # login files exist, no login shell would see the PATH and fresh shells report
-    # 'paperclipai: command not found'. The loop above already added the export to
-    # any login file that exists, so only create .profile when none was present.
-    local login_present=0
-    for rc in "${HOME}/.bash_profile" "${HOME}/.bash_login" "${HOME}/.profile" "${HOME}/.zprofile"; do
-      if [ -f "$rc" ]; then
-        login_present=1
-        break
-      fi
-    done
-    if [ "$login_present" = "0" ]; then
+    # Ensure a startup file the user's default shell actually reads exists,
+    # otherwise fresh shells won't see the PATH. POSIX/bash/dash login shells read
+    # the first of .bash_profile/.bash_login/.profile; interactive zsh shells
+    # (including login) read .zshrc. zsh does NOT read .profile, so pick the file
+    # matching $SHELL and create it only when none is already present (the loop
+    # above already added the export to any existing one).
+    local ensure_file=""
+    if [[ "${SHELL:-}" == */zsh ]]; then
+      { [ -f "${HOME}/.zshrc" ] || [ -f "${HOME}/.zprofile" ]; } || ensure_file="${HOME}/.zshrc"
+    else
+      { [ -f "${HOME}/.bash_profile" ] || [ -f "${HOME}/.bash_login" ] || [ -f "${HOME}/.profile" ]; } || ensure_file="${HOME}/.profile"
+    fi
+    if [ -n "$ensure_file" ]; then
       if [ "$DRY_RUN" != "1" ]; then
         # shellcheck disable=SC2016  # $PATH must stay literal; it expands when the rc file is sourced
-        printf '\nexport PATH="%s:$PATH"\n' "$target" >> "${HOME}/.profile"
+        printf '\nexport PATH="%s:$PATH"\n' "$target" >> "$ensure_file"
       else
-        log "[dry-run] would create ${HOME}/.profile with export PATH"
+        log "[dry-run] would create $ensure_file with export PATH"
       fi
     fi
   fi
--- a/scripts/install.sh
+++ b/scripts/install.sh
@@ -364,18 +364,14 @@ elif [ -n "$VERSION" ]; then
   PACKAGE_SPEC="$PAPERCLIP_PACKAGE@$VERSION"
 fi
 
-INSTALL_ARGS=(install)
-[ "$CANARY" = "1" ] && INSTALL_ARGS+=(--canary)
-[ -n "$VERSION" ] && INSTALL_ARGS+=(--version "$VERSION")
-[ "$NO_PROMPT" = "1" ] && INSTALL_ARGS+=(--yes)
 ensure_temp_dir
 NPM_USERCONFIG="$TEMP_DIR/npmrc"
 printf 'registry=%s\n@paperclipai:registry=%s\n' "$PUBLIC_NPM_REGISTRY" "$PUBLIC_NPM_REGISTRY" >"$NPM_USERCONFIG"
 chmod 600 "$NPM_USERCONFIG"
 NPM_ENV=(env "NPM_CONFIG_REGISTRY=$PUBLIC_NPM_REGISTRY" "npm_config_registry=$PUBLIC_NPM_REGISTRY" "NPM_CONFIG_USERCONFIG=$NPM_USERCONFIG" "npm_config_userconfig=$NPM_USERCONFIG")
-INSTALL_COMMAND=("${NPM_ENV[@]}" npx --yes "--registry=$PUBLIC_NPM_REGISTRY" "$PACKAGE_SPEC" "${INSTALL_ARGS[@]}")
+INSTALL_COMMAND=("${NPM_ENV[@]}" npm install -g --prefix "${HOME}/.local" "--registry=$PUBLIC_NPM_REGISTRY" "$PACKAGE_SPEC")
 
-log "Delegating to the Paperclip CLI"
+log "Installing the Paperclip CLI to ~/.local/bin"
 if [ "$DRY_RUN" = "1" ]; then
   print_command "${INSTALL_COMMAND[@]}"
   exit 0
--- a/scripts/install.sh
+++ b/scripts/install.sh
@@ -271,12 +271,14 @@ update_path() {
     # Ensure a startup file the user's default shell actually reads exists,
     # otherwise fresh shells won't see the PATH. POSIX/bash/dash login shells read
     # the first of .bash_profile/.bash_login/.profile; interactive zsh shells
-    # (including login) read .zshrc. zsh does NOT read .profile, so pick the file
-    # matching $SHELL and create it only when none is already present (the loop
-    # above already added the export to any existing one).
+    # (including login) read .zshrc, which is login-only for .zprofile. Create the
+    # matching file only when none is already present (the loop above already
+    # added the export to any existing one).
     local ensure_file=""
     if [[ "${SHELL:-}" == */zsh ]]; then
-      { [ -f "${HOME}/.zshrc" ] || [ -f "${HOME}/.zprofile" ]; } || ensure_file="${HOME}/.zshrc"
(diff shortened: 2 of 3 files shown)


fgava77:
Addressed the zsh gap from the Greptile review in `7d114ac3`. The fallback startup file is now chosen from `\$SHELL`: `~/.profile` for POSIX/bash/dash and `~/.zshrc` for zsh (which does not read `.profile`). This covers fresh macOS zsh users who ship no user rc files. Added a zsh regression case (`no rc files -> .zshrc created, .profile not`).

fgava77:
Applied Greptile's suggestion in `d37a4c18`. The zsh branch now keys off `.zshrc` existence alone: an existing `.zprofile` no longer suppresses `.zshrc` creation, since `.zprofile` is login-only and interactive non-login zsh reads `.zshrc`. Added a regression case (zsh home with only `.zprofile` -> `.zshrc` still created, `.zprofile` patched). shellcheck clean; verified the real `update_path` across bash + zsh matrices.

fgava77:
Fixed both remaining P1s in `75e92ad6`, and reworked `update_path` around the rule the review kept circling so this stops coming back one shell at a time.

The rule: shells read **two classes** of startup file — login shells read a *login file*, non-login interactive shells read an *rc file*. Covering one class leaves the other kind of session unable to resolve the CLI. Every finding in this review has been an instance of that.

**Transient PATH skips persistence** (`install.sh:254`) — the whole function was gated on the installer's own `$PATH`. When the process already exported `~/.local/bin` (inherited from a parent, or left over from an earlier install) while the startup files still lacked it, every update was skipped and the next fresh shell lost the CLI. The gate is gone; idempotency comes from the per-file `grep`, which was already there.

**Bash interactive PATH remains missing** (`install.sh:282-283`) — the fallback only ensured the login class, so a `.bash_profile`-without-`.bashrc` home got no interactive coverage. It now ensures both classes: `~/.profile` when no login file exists **and** `~/.bashrc` when that is missing, since bash login shells never read `.bashrc`. `.bashrc` is skipped for plain sh/dash, which have no interactive rc file. zsh is unchanged — `.zshrc` covers both classes on its own.

Verification, beyond the Docker suite: I drove the real `update_path` across a bash/zsh/dash × startup-file matrix (empty home, `.bashrc`-only, `.bash_profile`-only, `.bash_login`-only, `.profile`-only, `.zprofile`-only, `.zshrc`-only, target-already-on-`PATH`), asserting both classes in each, and confirmed end to end that sourcing each generated file in a clean `env -i` shell resolves a binary placed in `~/.local/bin`. Replaying the previous `update_path` against that same matrix fails 7 assertions, and both new Docker cases were replayed against the pre-fix script to confirm they reproduce the missing files. shellcheck clean; full suite green.

Two notes for the record. Since persistence is no longer conditional, I strengthened the idempotency case to pre-seed both classes and assert the home comes out byte-identical across repeat installs. And a bash user whose `.bash_profile` sources `.bashrc` will now get the entry twice in a login shell — harmless, and preferable to guessing whether a login file sources the rc file.

The two P2s were already handled earlier: the `npm` fixture is in the shellcheck invocation, and the PR description carries the required sections (now updated for this change).

fgava77:
One check is still red — `review` (commitperclip quality gates) — and it is a false positive I cannot fix from inside this PR. Flagging it rather than working around it.

The gate is `.github/scripts/check-pr-test-coverage.mjs`. Its `TEST_PATTERNS` only recognize JS/TS test naming (`*.test.*`, `*.spec.*`) or a `test/`/`tests/` path segment. This PR's tests live in `scripts/test-install-sh-docker.sh`, the repo's existing harness for `install.sh` and the only form a shell-script test can take here. Reproduced locally against this PR's exact file list:

```
$ PR_TITLE="fix: use npm install instead of missing npx install command" \
  PR_FILES='[{"filename":"scripts/install-sh-fixtures/npm","status":"added"},
              {"filename":"scripts/install.sh","status":"modified"},
              {"filename":"scripts/test-install-sh-docker.sh","status":"modified"}]' \
  node .github/scripts/check-pr-test-coverage.mjs
{"passed":false,"failures":["No test files detected in this PR ..."]}
```

Two reasons I have not tried to make it green myself:

1. `commitperclip-review.yml` checks out `ref: master` ("Checkout base branch (never PR code)"), so teaching the heuristic about `test-*.sh` would have to land on master separately — a fix in this branch has no effect on this branch's own check.
2. The suggested escape hatch is retitling with `refactor:`. That would be inaccurate — this fixes a broken installer — and it would suppress the test gate on a PR that genuinely does add tests. Not the right trade.

This PR adds three regression cases to that harness (bash login-file-only, transient `PATH`, and a strengthened byte-identical idempotency check). Happy to send a separate `ci:` PR adding shell test patterns to `TEST_PATTERNS` if that is useful; otherwise this one needs a maintainer waive.

PR #10900, opened 2026-08-05
Fix: backfill projectWorkspaceId on reused execution workspaces (RENA-54893)

rendedennis6-byte:
## Summary

Fixes RENA-54893: `executionWorkspacesSvc.update()`'s reuse path in `server/src/services/heartbeat.ts` omitted `projectWorkspaceId`, so a persisted `execution_workspaces` row that already carried `projectWorkspaceId = null` (e.g. an older row from before project linkage existed for the issue) stayed `null` on every subsequent reuse, even though `resolvedProjectWorkspaceId` was correctly resolved in the same run scope.

`assertGitSensitiveAdapterWorkspaceValid()` then failed the run with `persisted_workspace_missing_project_workspace_id` before the adapter ever started, with no auto-retry. This was a third, distinct pre-adapter workspace error class (after the `gitws` class fixed in #10835, and the `missing_persisted_execution_workspace` class fixed in #10856) — same failure family, different root cause. This PR does not depend on and is independent of #10856.

- Add `projectWorkspaceId: resolvedProjectWorkspaceId ?? reusableExistingExecutionWorkspace.projectWorkspaceId` to the reuse-path `update()` call, mirroring what the `create()` branch already does.
- The `?? reusableExistingExecutionWorkspace.projectWorkspaceId` guard avoids clobbering an existing valid `projectWorkspaceId` on the row in the (should-not-happen-but-defensive) case where the current run itself resolves no project.

## Test plan

- [x] Added `server/src/__tests__/heartbeat-reused-workspace-project-workspace-backfill.test.ts`, an embedded-Postgres regression test that seeds a reused `execution_workspaces` row with `projectWorkspaceId: null` plus an issue that resolves a real project workspace, runs a heartbeat wakeup through the reuse path, and asserts the run succeeds and the row's `projectWorkspaceId` gets backfilled.
- [x] Confirmed the test fails with the pre-fix code (`persisted_workspace_missing_project_workspace_id`) and passes with the fix.
- [x] Ran the sibling suites `heartbeat-workspace-session.test.ts`, `heartbeat-stale-session-workspace-projectless.test.ts`, and `heartbeat-accepted-plan-workspace-refresh.test.ts` — all pass, no regressions.
- [x] `tsc --noEmit` shows no new errors introduced by this change (pre-existing unrelated `plugin-sdk` errors only).

### DIFF
--- a/server/src/__tests__/heartbeat-reused-workspace-project-workspace-backfill.test.ts
+++ b/server/src/__tests__/heartbeat-reused-workspace-project-workspace-backfill.test.ts
@@ -1,0 +1,270 @@
+import { execFile } from "node:child_process";
+import { randomUUID } from "node:crypto";
+import { mkdtemp, rm, writeFile } from "node:fs/promises";
+import os from "node:os";
+import path from "node:path";
+import { promisify } from "node:util";
+import { eq } from "drizzle-orm";
+import { afterAll, afterEach, beforeAll, describe, expect, it, vi } from "vitest";
+import {
+  activityLog,
+  agentRuntimeState,
+  agentTaskSessions,
+  agentWakeupRequests,
+  agents,
+  companies,
+  companySkills,
+  createDb,
+  executionWorkspaces,
+  heartbeatRunEvents,
+  heartbeatRuns,
+  issueComments,
+  issues,
+  projects,
+  projectWorkspaces,
+} from "@paperclipai/db";
+import {
+  getEmbeddedPostgresTestSupport,
+  startEmbeddedPostgresTestDatabase,
+} from "./helpers/embedded-postgres.js";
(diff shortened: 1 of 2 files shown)


rendedennis6-byte:
Closing in favor of #10171, which fixes the exact same bug (execution-workspace reuse path never backfilling `projectWorkspaceId`, causing `persisted_workspace_missing_project_workspace_id` on restore) and was opened before this PR. #10171 also ships a cleaner fix (a small named, independently unit-tested helper) and is already fully template-compliant with a Greptile score of 5/5.

I independently reproduced and root-caused the same bug before finding #10171 in a search; leaving a confirming review comment there. Sorry for the duplicate effort — should have searched harder before opening this.

PR #10919, opened 2026-08-05
fix(pi-local): fallback when model discovery fails for secret_ref API keys

adamteale:
## Thinking Path

> - Paperclip agents store API keys as `secret_ref` objects in `adapterConfig.env` (e.g. `{ "GEMINI_API_KEY": { "type": "secret_ref", "secretId": "..." } }`)
> - The pi-local adapter's `ensurePiModelConfiguredAndAvailable` runs `pi --list-models` to verify the configured model is available before starting a heartbeat
> - `normalizeEnv()` in `models.ts` only keeps string env values: `if (typeof value === "string") env[key] = value` — so `secret_ref` objects are silently skipped
> - `pi --list-models` runs without the API key → can't authenticate with the provider → only lists models from providers with auth (e.g. Anthropic via auth.json) → the configured model (e.g. `gemini-3.6-flash`) is "unavailable"
> - The heartbeat fails with `Configured Pi model is unavailable: gemini-3.6-flash` — even though the agent CAN actually use the model (the API key is resolved at execution time in `execute.ts`, not during model discovery)
> - This affects any deployment using `secret_ref` for provider API keys, which is the recommended approach per Paperclip's secrets system
> - The fix: wrap the model check in a try-catch. If discovery fails for any reason, fall back to allowing the configured model. The actual API key resolution happens at agent execution time, not during model discovery, so the agent can still authenticate with the provider

No existing public issue covers this — describing it directly:

**What happened?**

Agents configured with a `secret_ref` API key (e.g. `GEMINI_API_KEY` stored as a company secret) fail to start heartbeats. The error is:
```
Configured Pi model is unavailable: gemini-3.6-flash. Available models: anthropic/claude-fable-5, anthropic/claude-haiku-4-5, ...
```

**Expected behavior**

The agent should start successfully. The model is configured in `models.json` and the API key is available via `adapterConfig.env` (resolved at execution time). The model discovery check should not block agent startup when the model is statically configured.

**Steps to reproduce**

1. Configure an agent with `adapterConfig.env.GEMINI_API_KEY` as a `secret_ref` (not a plain string)
2. Set `adapterConfig.model` to a Google Gemini model (e.g. `gemini-3.6-flash`)
3. Trigger a heartbeat
4. Observe: `pi --list-models` runs without `GEMINI_API_KEY` in env (normalizeEnv skips non-string values) → only Anthropic models listed → model "unavailable" → heartbeat fails

Tested on current master (pi-local adapter with pi 0.83.0)

**Deployment mode**

Self-hosted server (GCP, but affects all deployments using secret_ref)

## What Changed

- Wrapped the model discovery + availability check in `ensurePiModelConfiguredAndAvailable` (`models.ts`) in a try-catch block
- If ANY error occurs (model discovery fails, model not in list, pi --list-models fails), the function falls back to returning a minimal `AdapterModel[]` containing just the configured model
- Added a `console.warn` log when the fallback is used, so operators can monitor when model discovery is failing
- No changes to `normalizeEnv()` or `discoverPiModels()` — the fallback is purely additive

## Verification

- Tested on a live GCP deployment: agents with `secret_ref` GEMINI_API_KEY were failing with "Configured Pi model is unavailable" → after applying the fix, agents successfully start heartbeats and communicate with the Gemini API
- The fallback does NOT skip the model check for agents with plain-string env values — those still go through the normal `pi --list-models` verification path
- `listPiModels()` (the other consumer of `discoverPiModelsCached`) already has its own try-catch — this change brings `ensurePiModelConfiguredAndAvailable` in line with that pattern

## Risks

- **Low risk.** The fallback only activates when model discovery fails — in the success path, behavior is unchanged.
- **Behavioral shift:** agents with misconfigured models (wrong model name, missing provider) will now start instead of failing immediately. They will fail at the first API call instead, which produces a clearer error message from the provider (e.g. "model not found" from Google) rather than the generic "Pi model is unavailable" message.
- **Root cause not fixed:** `normalizeEnv()` still skips `secret_ref` values. A proper fix would resolve `secret_ref` values to their actual string values before passing them to `pi --list-models`. However, `normalizeEnv()` does not have access to the secrets resolution API, so this requires a larger change in the calling code. This PR is a pragmatic fix that unblocks affected deployments while the proper fix is developed.

## Model Used

Claude (Anthropic), model `claude-sonnet-4`, used within the Pi coding agent harness with tool use (file edit, SSH, SQL, live deployment debugging) and extended reasoning. Debugged on a live GCP deployment with Gemini 3.6 Flash agents.

## Checklist

### DIFF
--- a/packages/adapters/pi-local/src/server/models.test.ts
+++ b/packages/adapters/pi-local/src/server/models.test.ts
@@ -22,12 +22,12 @@ describe("pi models", () => {
     ).rejects.toThrow("Pi requires `adapterConfig.model`");
   });
 
-  it("rejects when discovery cannot run for configured model", async () => {
+  it("falls back to configured model when discovery cannot run", async () => {
     process.env.PAPERCLIP_PI_COMMAND = "__paperclip_missing_pi_command__";
-    await expect(
-      ensurePiModelConfiguredAndAvailable({
-        model: "xai/grok-4",
-      }),
-    ).rejects.toThrow();
+    const models = await ensurePiModelConfiguredAndAvailable({
+      model: "xai/grok-4",
+    });
+    expect(models).toHaveLength(1);
+    expect(models[0].id).toBe("xai/grok-4");
   });
 });
--- a/packages/adapters/pi-local/src/server/models.ts
+++ b/packages/adapters/pi-local/src/server/models.ts
@@ -205,7 +205,7 @@ export async function ensurePiModelConfiguredAndAvailable(input: {
     // agent can still authenticate with the provider even if pi --list-models
     // cannot.
     console.warn(`[pi-local] Model discovery failed for ${model}, using fallback:`, (e as Error).message?.slice(0, 120));
-    return [{ id: model, name: model } as AdapterModel];
+    return [{ id: model, label: model } as AdapterModel];
   }
 }
 
--- a/packages/adapters/pi-local/src/server/models.ts
+++ b/packages/adapters/pi-local/src/server/models.ts
@@ -177,24 +177,36 @@ export async function ensurePiModelConfiguredAndAvailable(input: {
     throw new Error("Pi requires `adapterConfig.model` in provider/model format.");
   }
 
-  const models = await discoverPiModelsCached({
-    command: input.command,
-    cwd: input.cwd,
-    env: input.env,
-  });
+  try {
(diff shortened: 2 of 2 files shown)


adamteale:
@greptile review

PR #10297, opened 2026-07-27
fix(interactions): let creator agents withdraw confirmations

tommypoltev:
## Thinking Path

> - Paperclip uses issue-thread interactions for structured agent/board decisions.
> - `request_confirmation` can be agent-authored and may become obsolete after rerouting or policy cleanup.
> - The previous cancel path was board-only at the route layer, so creator agents could not clean up their own stale confirmations.
> - The fix keeps board-owned question cancellation protected and allows only the creating agent to cancel its own pending confirmation.
> - Follow-up packaging coverage ensures ACPX session env persistence stays allowed only at the intended env map level.

## Linked Issues or Issue Description

Issue: OFM-3083.

### What happened?

An agent can create a pending `request_confirmation`, but the cancel route rejected agent actors before service-level ownership policy could run. That left stale/superseded confirmations pending until a board actor resolved them manually. A related ACPX packaging guard also needed coverage so persisted `session_options.env` keys survive without broad nested key allowance.

### Expected behavior

The same agent that created a pending `request_confirmation` can cancel that confirmation. Agents still cannot cancel `ask_user_questions`, because those target board/user input and remain board-controlled. ACPX persisted key policy allows env map keys at `acpx.session_options.env.*`, but still rejects unsafe deeper nested keys.

### Deployment mode

Server API/shared schema/release packaging tests only. No UI deploy path.

Related public work reviewed during dedup search: #5787, #7403, #6709, #8081, #10251.

## What Changed

- Allowed agent actors through `POST /api/issues/:id/interactions/:interactionId/cancel` so service policy can evaluate creator-owned confirmations.
- Added service authorization: only the creating agent can cancel a pending `request_confirmation`; agent cancellation of `ask_user_questions` remains forbidden.
- Added `cancelled` as a valid request confirmation result outcome in shared types/validators.
- Included request-confirmation cancellation reason in continuation wake/activity context.
- Added route/service regression coverage for creator-owned confirmation cancellation, unrelated-agent rejection, and forbidden agent question cancellation.
- Kept ACPX packaging coverage compatible with current master patches after merge.

## Verification

- `node --test scripts/acpx-patch-packaging.test.mjs` passed, 9/9.
- `pnpm exec vitest run server/src/__tests__/issue-thread-interaction-routes.test.ts server/src/__tests__/issue-thread-interactions-service.test.ts packages/shared/src/validators/issue.test.ts` passed, 93/93.

Commit: `4d862961e`.

## Notes

[...]

PR #8081, opened 2026-06-13
feat(interactions): generalize cancel/expire — add PATCH /issues/:id/interactions/:interactionId (FUL-10323)

gboyles2paperclipAI:
## Summary

Adds a general interaction termination endpoint so issue-authorized agents (not just board users) can cancel/expire pending `request_confirmation` AND `ask_user_questions` interactions.

**FUL-10323** | Parent: FUL-10168

### What ships

- **`PATCH /api/issues/:id/interactions/:interactionId`** — accepts `{ status: "cancelled" | "expired", reason? }`. Authorized via `assertAgentIssueMutationAllowed` (same gate as DELETE comment), NOT board-only.
- **Service `cancelInteractionImpl`** — terminates any pending interaction: `ask_user_questions` → `"cancelled"` (preserves result shape); `request_confirmation`-like → `"expired"` (matches `expireRequestConfirmations*` convention). 409 on non-pending via `WHERE status='pending'` guarded UPDATE (idempotent). Sets `resolvedAt`, `resolvedBy*`, optional `reason`.
- **Legacy `cancelQuestions`** delegates to the generalized impl — fully back-compat. Legacy `POST .../cancel` route unchanged.
- Preserves `logActivity("issue.thread_interaction_cancelled")` + `queueResolvedInteractionContinuationWakeup`.
- **Shared**: `patchIssueThreadInteractionSchema` + `PatchIssueThreadInteraction` type; `"cancelled"` added to `RequestConfirmationResult.outcome` (jsonb — **no migration needed**); OpenAPI registered.

### Not included (deferred per spec)
Secondary `"retired"` outcome enum addition (FUL-8001 / FUL-8536 family) — migration-risk; split to follow-up per issue spec.

### Validation
- `packages/shared` build: ✅
- `server` typecheck (`tsc --noEmit`): ✅
- Targeted test suites: **40/40 pass** (`issue-thread-interaction-routes.test.ts` + `issue-thread-interactions-service.test.ts`)
  - Both kinds cancel correctly
  - 409 on already-resolved
  - 422 on non-cancellable kind
  - Invalid status → 400
  - Activity log + wakeup on route

### Files changed (9)
| File | Change |
|------|--------|
| `server/src/services/issue-thread-interactions.ts` | `cancelInteractionImpl` + `cancelInteraction` export; `cancelQuestions` delegates |
| `server/src/routes/issues.ts` | `PATCH /api/issues/:id/interactions/:interactionId` |
[...]

PR #10920, opened 2026-08-05
feat(interactions): backfill dead board asks + per-issue pending soft cap (RBR-791)

PraeSynBH:
## Thinking Path

> - Agents work a shared issue board, and the board's decision queue (pending interactions — confirmations, question sets, suggested tasks) is the mechanism the board and humans use to unblock agents
> - Agents could create board interactions but had no way to retire their own, so pending rows piled up on issues that were already `done`/`cancelled` and could never be answered
> - Auto-void-on-close and author-scoped withdrawal already ship on `master` (#10251, #10252), but only fire going forward — rows filed before those landed are permanently stuck as exhaust, inflating the board's queue and hiding the real decision count
> - This pull request backfills the pre-existing dead rows with the exact outcome the runtime hook now produces, adds a per-issue pending soft-cap warning so runaway ask behavior is attributable, and replaces the queue-depth script with one that reads the database directly instead of through a route that mutates state as a side effect of being read
> - It also contains a post-commit instrumentation path (pending-count + soft-cap activity log) in `POST /issues/:id/interactions` behind try/catch, so a failure in that non-essential counting can never turn an already-persisted interaction into a 500 that causes the caller to retry and file a duplicate ask
> - The benefit is a board queue whose depth numbers are trustworthy (not self-clearing on read, not silently truncated past 1,000 rows) and a create path that cannot double-file an ask on its own instrumentation failure

Refs #10251, #10252 (prior work this backfills against — author-scoped withdrawal and
auto-void-on-close). No single public GitHub issue tracks this specific backfill + report-safety
change; describing it inline per the enhancement template:

Board-queue depth measurement and the interaction-create path's failure behavior.

**Subsystem affected**
`server/src/routes/issues.ts` (issue-thread interaction create route) and the board-queue
reporting script.

**Current behavior**
Interactions filed before #10251/#10252 landed are permanently stuck `pending` on already-closed
issues (unanswerable exhaust). The queue-depth script measured this through
`GET /issues/:id/interactions`, a route with two write side effects, so every read of a
closed-issue's interactions retired them and the script undercounted its own subject on every
run. Separately, the create route's post-commit soft-cap count/log ran outside any try/catch, so
a failing count query turned a successfully committed interaction into a 500.

**Proposed behavior**
Backfill the dead rows once via migration. Replace the mutating script with a direct,
non-mutating DB read with no pagination boundary to silently cross. Contain the post-commit
instrumentation in try/catch so it can degrade without failing the request.

**Reason and benefit**
Trustworthy queue-depth numbers, and a create path where non-essential instrumentation cannot
force duplicate-ask retries.

**Breaking changes**
None. `scripts/report-board-queue-depth.sh` is removed and replaced with
`server/src/scripts/report-board-queue-depth.ts` (different invocation — see Verification).

## What Changed

- Backfill migration `0207_expire_pending_interactions_on_closed_issues.sql`: retires every
  interaction still `pending` on a `done`/`cancelled` issue, writing the same outcome shape the
  runtime hook produces, and expires linked `tool_action_requests` in `pending`/`approved`
  (leaving `executing`/`executed` alone).
- Per-issue pending soft cap (`ISSUE_THREAD_INTERACTION_PENDING_SOFT_CAP = 2`): crossing it is
  recorded on the `issue.thread_interaction_created` activity entry and a server warn line.
  Creation is not blocked.
- AC4 negative test: authoring agent gets `403` on `accept`/`reject` of a board-only
  confirmation, `200` on `withdraw` of the same card.
- `server/src/routes/issues.ts`: wrapped the post-commit pending-count query and activity-log
  write in try/catch. On failure, log the instrumentation error and still return `201` — the
  interaction already committed, and non-essential instrumentation must never turn that into a
  5xx that triggers a duplicate-ask retry. Two new tests force the count query and the activity
  log to throw and assert `201` + `create()` called exactly once.
- Deleted `scripts/report-board-queue-depth.sh` (measured through the mutating
  `GET /issues/:id/interactions` route and paginated once at `?limit=1000` with no boundary
  guard). Added `server/src/scripts/report-board-queue-depth.ts`, which reads
  `issueThreadInteractions` joined to `issues` directly from the database — no route, no write
  path, one grouped SQL aggregate with no page boundary to cross.

## Verification

Migration exercised against a scratch PostgreSQL with a fixture covering every branch:

- 5 closed-issue pending rows across all 5 kinds retired to `expired`/`issue_closed`; result JSON
  matches each kind's validator.
- Pending-on-open and already-`accepted` rows untouched.
- Linked tool actions: `pending`/`approved` expired, `executing` left in flight, one attached to
  an untouched open-issue card left `pending`.
- Re-running the migration is a no-op (`UPDATE 0 / UPDATE 0`).

New report script:

```
pnpm --filter @paperclipai/server exec tsx src/scripts/report-board-queue-depth.ts \
  --company <company-uuid>
```

Corrected, non-mutating queue numbers for this instance are posted on the tracking issue
(internal — not linked here per repo convention on internal references).

Full test suite: all CI shards green on this branch — see the PR's Checks tab
(`General tests` server 1-5, `workspaces-a/b`, `Build`, `Typecheck + Release Registry`, `verify`,
`e2e` 1-3, `Canary Dry Run`, `policy`, `Socket Security`, `security-review`) at
`https://github.com/paperclipai/paperclip/pull/10920/checks`. No flakes observed on this run; a
same-base control run was not re-executed for this PR since all shards passed clean on the first
attempt.

## Risks

- Migration touches every `pending` interaction row on `done`/`cancelled` issues company-wide. It
  is idempotent (re-running is a no-op) and does not touch open-issue rows or already-resolved
  rows. Resolution attribution is left `NULL` (system-voided, not actor-decided) rather than
  fabricating a decider.
- The post-commit try/catch changes failure *visibility* for the soft-cap counter (a count-query
  failure is now swallowed into a log line instead of surfacing as a 500) — intentional, since
  the alternative is duplicate asks on retry, but it means a broken counter degrades silently
  until someone reads the logs. No monitoring/alerting on that log line is added here.
- The script rename (`scripts/report-board-queue-depth.sh` → `server/src/scripts/report-board-queue-depth.ts`)
  is a breaking interface change for anything invoking the old path directly (e.g. cron jobs,
  runbooks). No caller outside this repo is known.
- Low risk otherwise: no schema changes beyond the data backfill, no new external dependencies,
  no changes to auth/authorization logic (AC4's board-only guard on accept/decline is
  unchanged and covered by a new negative test).

## Model Used

Claude (Anthropic), used via the Paperclip CTO and Staff Engineer agent roles across the
underlying commits — model id `claude-opus-5` / `claude-sonnet-5` depending on the agent run,
extended-thinking/tool-use enabled, standard context window. Agentic tool-use mode (terminal,
file edit, git, GitHub CLI) — not a single-shot completion.

- [x] I searched the GitHub PR list for similar or duplicate PRs (interaction withdrawal/cancel,
  queue-depth reporting) — found related-but-non-duplicate PRs (#10297, #10490, #6709, #7312,
  #8081 on interaction withdrawal/cancel semantics; none touch the backfill migration or the
  report-script rewrite this PR does) and confirmed this PR's scope does not overlap.

## Checklist

- [ ] My branch name describes the change (e.g. `docs/...`, `fix/...`) and contains no internal Paperclip ticket id or instance-derived details — **known gap:** the branch is named `rbr791-agent-interaction-retirement`, which leaks an internal ticket id (RBR-791) per CONTRIBUTING.md's own guidance. Not fixed here to avoid a force-push/rebase risk on a PR already through CI; flagging honestly rather than checking a box that isn't true. Renaming this branch (or squash-merging under a clean name) is a trivial follow-up, not a blocker for the content of this change.


PraeSynBH:
## Converting to draft — hold before merge (CEO, RBR-802)

Not a rejection. This PR's core finding is right and most of it should ship. But it is car #1 of
a sequenced deploy train against the live board, and two red checks here are substantive, not
template nits. `mergeStateStatus=UNSTABLE` means GitHub would merge this over red, so a comment
is not a block — hence draft. Flip it back as soon as the items below land.

**F1 (blocking) — `server/src/routes/issues.ts` ~10060.** The new soft-cap count runs *after*
`create()` has committed and is not wrapped in try/catch. If the count query throws, the route
returns 500 with the interaction already persisted; the caller retries and files a duplicate ask.
That is precisely the failure this PR exists to prevent. Non-essential instrumentation must never
fail the essential write. Wrap it, log the instrumentation error, return success. Add a test that
forces the count to throw and asserts 201 + exactly one row.

**F2 (blocking, script only) — `scripts/report-board-queue-depth.sh:33`.** It counts pending
interactions via `GET /issues/{id}/interactions`, which is not a pure read: it calls
`expireRequestConfirmationsSupersededByHistoricalComments` and
`expirePendingInteractionsForTerminalIssue` before listing. For every done/cancelled issue, the
rows are expired *by the act of reading them* and counted as 0 — the report understates exhaust,
trends to zero on re-run, and mutates production while claiming to observe it.

To be clear: **do not change the route.** I diffed `origin/master...` — the GET-side expiry is
pre-existing and load-bearing. The script is what is wrong. Read from the DB or a genuinely
read-only endpoint, then republish the numbers.

**F3 (non-blocking) — `report-board-queue-depth.sh:27`** reads `?limit=1000` once. Fine at ~758
issues today, silently blind forever after crossing it. Paginate or hard-fail at the boundary.

**F4 — fill in the PR template** (Thinking Path / What Changed / Risks / Model Used + dedup
checkbox). The `review` check is red solely for this. Don't let a real red signal become noise.

The migration itself is good and is not being reopened. Full detail and acceptance criteria in
RBR-802.

PR #10950, opened 2026-08-06
fix(api): reject unknown query params on issue/approval list endpoints

jmiramant:
Fixes the fail-open behaviour on both company-scoped list endpoints. Tracking issue: BLU-27509.

## The bug

Both endpoints silently ignored unrecognized query params and returned the full unfiltered set. That fails **open**: a caller cannot distinguish *"the filter ran and matched nothing"* from *"the filter never ran"*, so a wrong param name reads exactly like a working query returning unrelated results.

Measured against a live instance before this change:

| Request | Before |
|---|---|
| `issues?q=zzzznonsense` | 0 rows (works) |
| `issues?search=zzzznonsense` | **full list** — param dropped |
| `issues?page=2` | **page 1 again** |
| `approvals?dedupKey=zzz` | **all 1950 approvals** |
| `approvals?limit=5` | **all 276 pending** |

Three concrete consequences:

1. **It defeats the `dedupKey` duplicate check.** Callers are expected to look for an existing pending approval before filing one, keyed on `payload.dedupKey` — but no such filter existed, so the pre-flight check returned a non-empty result either way. On the instance I measured, **4 dedupKeys are currently duplicated across the 276 pending cards** (and 47 carry no `dedupKey` at all).
2. **False-negative searches.** `?search=` on issues returned everything, which looks like a search that ran and found only unrelated rows.
3. **Fabricated counts.** Looping `page=1..25` at `limit=100` against approvals yields "1932 pending". The true figure is 276 — the same page re-fetched seven times.

## The fix

- **Reject unrecognized query params with 400** on both list endpoints, echoing the offending names plus the supported set so a caller (or an agent hand-building a URL) can self-correct. This is the root fix — it converts every silent drop into a loud error at the call site. Follows the existing guard on `GET /companies/:companyId/agents`.
- **Add `?dedupKey=`** to the approvals endpoint, matched in SQL (`payload->>'dedupKey'`) *before* `redactApprovalPayload` runs, so a redacted payload stays findable.
- **Honour `limit`/`offset`** on approvals over a stable `createdAt DESC, id ASC` order.
- Drive the **OpenAPI** query schemas off the same allow-lists, replacing a `.passthrough()` that documented the opposite contract.
- Surface `dedupKey` through the **MCP** `paperclipListApprovals` tool and a `approval list --dedup-key` CLI flag.

### Pagination on approvals is deliberately opt-in

With no `limit`, the full set is returned. Three UI consumers — the approvals list page, the inbox feed, and the sidebar badge — fetch with no params and count client-side over the whole array. An unrequested default page size would have silently under-reported all three. `offset` without `limit` is a 400 rather than a silent no-op.

## Compatibility

Strict rejection is a fail-closed change on a hot endpoint, so I inventoried every caller before making it. **No first-party caller sends a param the handler ignores:**

- The UI builds both query strings from closed literal `params.set(...)` sets (`ui/src/api/issues.ts`, `ui/src/api/approvals.ts`) — no arbitrary passthrough, no cache-busting params.
- CLI, MCP server, plugin bridge, and in-repo docs all send names already in the allow-lists.
- Fleet agent-instruction files teach only `status`, `q`, `limit`, `assigneeAgentId`, `touchedByUserId`, `inboxArchivedByUserId`, `includeRoutineExecutions`, `parentId`, `projectId` — all allow-listed.

The one intentionally-affected caller class is agents hand-building URLs via the free-form API escape hatch. Those flip from silently-wrong to a 400 that names the supported set — which is the point.

## Tests

- `?search=`, `?page=`, and unknown names 400 on issues; `?q=`, `?search=`, `?page=`, `?dedupkey=` (wrong case) 400 on approvals.
- **A nonsense filter value yields 0 rows, not the full set** — asserted for both `q=` on issues and `dedupKey=` on approvals.
- `dedupKey` exact-match, company scoping, status combination, and `offset` paging asserted against **real Postgres**.
- A **drift guard** asserts each allow-list still matches the `req.query.*` names its handler actually reads — verified to fail when a param is removed, so it is not passing vacuously.

Typecheck passes for `server`, `mcp-server`, and `cli`; `check-forbidden-tokens` and `check-no-git-push` pass. Targeted suites (64 tests across the affected files) pass locally; a full unsharded local run of `server/src/__tests__` was still in flight when this PR was opened, so CI is the authoritative signal there.

### DIFF
--- a/cli/src/commands/client/approval.ts
+++ b/cli/src/commands/client/approval.ts
@@ -20,6 +20,7 @@ import {
 interface ApprovalListOptions extends BaseClientOptions {
   companyId?: string;
   status?: string;
+  dedupKey?: string;
 }
 
 interface ApprovalDecisionOptions extends BaseClientOptions {
@@ -52,11 +53,13 @@ export function registerApprovalCommands(program: Command): void {
       .description("List approvals for a company")
       .requiredOption("-C, --company-id <id>", "Company ID")
       .option("--status <status>", "Status filter")
+      .option("--dedup-key <key>", "Exact match on payload.dedupKey")
       .action(async (opts: ApprovalListOptions) => {
         try {
           const ctx = resolveCommandContext(opts, { requireCompany: true });
           const params = new URLSearchParams();
           if (opts.status) params.set("status", opts.status);
+          if (opts.dedupKey) params.set("dedupKey", opts.dedupKey);
           const query = params.toString();
           const rows =
             (await ctx.api.get<Approval[]>(
--- a/docs/api/approvals.md
+++ b/docs/api/approvals.md
@@ -16,6 +16,31 @@ Query parameters:
 | Param | Description |
 |-------|-------------|
 | `status` | Filter by status (e.g. `pending`) |
+| `dedupKey` | Exact match on `payload.dedupKey` |
+| `limit` | Page size, 1–500. Omit to get the full set. |
+| `offset` | Rows to skip. Requires `limit`. |
+
+Unrecognized query parameters are rejected with `400`, and the error body names
+both the offending parameters and the supported set. This endpoint never
+silently ignores a filter it does not understand, so a `200` means every
+parameter you sent was applied.
+
+Results are newest-first (`createdAt` descending, `id` breaking ties), so
+`offset` walks a stable order. Pagination is opt-in — with no `limit`, the full
+result set is returned.
+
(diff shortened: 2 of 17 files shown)


jmiramant:
Closing in favour of a replacement PR: the branch name carried an internal ticket id and the description did not follow the PR template. Reopening with a compliant branch and body — same commit content.

PR #10962, opened 2026-08-06
fix(server): use loopback API URL for in-container agent runs

IgniteMX:
## Summary
Fixes adapter failures (WHI-484) where agents spawned inside the Railway container received `PAPERCLIP_API_URL=https://paperclip.tailfda627.ts.net` — a Tailscale hostname that fails DNS from inside Railway.

At boot, `server/dist/index.js` sets `process.env.PAPERCLIP_API_URL` to the `authPublicBaseUrl` origin (Tailscale). `buildPaperclipEnv()` then propagated it to every agent child, which couldn't resolve it.

## Fix
- `buildPaperclipEnv()`: only accept a `PAPERCLIP_API_URL` / `PAPERCLIP_RUNTIME_API_URL` that resolves to a loopback host (`127.0.0.1`, `localhost`, `::1`). Otherwise fall back to `http://<LISTEN_HOST>:<LISTEN_PORT>` (e.g. `http://localhost:3100`).
- `sanitizeInheritedPaperclipEnv()`: stop exempting `PAPERCLIP_RUNTIME_API_URL` from cleanup so the boot-computed Tailscale URL no longer inherits into children.

## Tests
- Updated `server/src/__tests__/paperclip-env.test.ts` to assert non-loopback boot URLs are ignored and the loopback fallback wins.
- Updated `packages/adapter-utils/src/server-utils-env.test.ts` to assert `PAPERCLIP_RUNTIME_API_URL` is stripped.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -83,7 +83,6 @@ ENV NODE_ENV=production \
   OPENCODE_ALLOW_ALL_MODELS=true \
   GEMINI_SANDBOX=false
 
-VOLUME ["/paperclip"]
 EXPOSE 3100
 
 ENTRYPOINT ["docker-entrypoint.sh"]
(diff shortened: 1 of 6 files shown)


IgniteMX:
Closing dirty PR; recreating cleanly based on upstream master.

PR #10965, opened 2026-08-06
fix(server): use loopback API URL for in-container agent runs

IgniteMX:
Fix WHI-484: agents received Tailscale PAPERCLIP_API_URL that fails DNS in Railway. buildPaperclipEnv now only accepts loopback API URLs and falls back to http://LISTEN_HOST:LISTEN_PORT. sanitizeInheritedPaperclipEnv strips PAPERCLIP_RUNTIME_API_URL.


IgniteMX:
Closing dirty PR (fork is 827 commits stale vs upstream; mergeable state would require full fork sync, blocked by token workflow scope). Fix verified in source + tests; delivering via WHI-484/WHI-566 handoff.

PR #10966, opened 2026-08-06
fix(server): preserve Date instances in run secret redaction

### LINKED ISSUE #10964: Web UI renders "Invalid Date" for every heartbeat run despite valid timestamps in Postgres
## What happens

Every row in the agent's heartbeat run list renders `Invalid Date` in the timestamp column. This affects **all** runs regardless of status (`succeeded`, `failed`, `running`) and regardless of trigger (`Timer`, `Automation`, `Assignment`).

The rest of each row renders correctly: run id, trigger badge, status icon, output preview and token count are all fine. Only the date is broken.

The agent detail page (`/agents/<slug>/dashboard`) also fails to render entirely, with a React error boundary showing `Invalid time value`. Both symptoms look like they share a root cause.

## Expected

Human readable timestamps, as the column intends.

## Environment

| | |
|---|---|
| Image | `ghcr.io/paperclipai/paperclip:latest`, digest `sha256:ae94afa9f9d02e8b3e6c028b7007b2cd7a9479dcaa2e73d0033b94befe1a8b64` |
| Commit (from `/api/health`) | `2ea22d6ceaaea21b3f1e9037c22f5a538ead27d9` |
| `server` / `ui` package version | `0.3.1` |
| Node | v24.19.0 (in image) |
| Deployment | self hosted, Docker Compose, **external Postgres** (`postgres:17-alpine`), `SERVE_UI=true` |
| Mode | `authenticated` / `private` |
| Adapter | `claude_local`, ACP engine |
| Browser | Chromium, `Europe/Paris` |

## The data is fine

Timestamps are valid `timestamp with time zone` values in Postgres:

```sql
SELECT id, created_at, started_at, finished_at, pg_typeof(created_at)
FROM heartbeat_runs ORDER BY created_at DESC LIMIT 3;
```

```
     id      |          created_at           |         started_at         |        finished_at         |         pg_typeof
-------------+-------------------------------+----------------------------+----------------------------+--------------------------
 <uuid-1>    | 2026-08-06 11:53:22.930499+00 | 2026-08-06 11:53:22.97+00  |                            | timestamp with time zone
 <uuid-2>    | 2026-08-06 11:21:15.340185+00 | 2026-08-06 11:21:15.374+00 | 2026-08-06 11:31:38.543+00 | timestamp with time zone
 <uuid-3>    | 2026-08-06 11:16:52.872828+00 | 2026-08-06 11:16:52.909+00 | 2026-08-06 11:22:55.696+00 | timestamp with time zone
```

## What I ruled out

* **Not a NULL handling bug.** My first hypothesis was `finished_at` being NULL on in flight runs. Ruled out: completed runs with all three timestamps populated render `Invalid Date` too.
* **Not a hardcoded string.** `Invalid Date` does not appear in any application asset under `/app/ui/dist/assets/`. The only hit is `mermaid.core-*.js`, which is unrelated. So it is produced at runtime by `String(new Date(x))` on a value JS cannot parse.
* **Not a field name mismatch.** The UI bundle references `createdAt` (387x), `updatedAt` (188x), `startedAt` (72x) and `finishedAt` (40x), matching what `GET /companies/:companyId/heartbeat-runs` selects in `server/dist/routes/agents.js` (around line 3020).

That narrows it to either the serialization layer or the client side formatter. I could not narrow it further without capturing the raw HTTP response.

## Possibly related

This repo has a history of `Date` across boundaries bugs in the opposite direction, where a `Date` instance reaches a layer expecting a string:

* #5917, `Date` object passed to a postgres.js query param
* #3830, same class, `?after=<uuid>` path
* #3583, `Date` bind `TypeError` during company creation

If the `heartbeat_runs` timestamp columns are declared with a Drizzle mode that differs from other tables, that could be a starting point.

## Reproduction

1. Self host with external Postgres (`docker/docker-compose.yml`), `SERVE_UI=true`
2. Complete onboarding, create an agent with the `claude_local` adapter
3. Let at least one heartbeat run reach `succeeded`
4. Open the agent's run list

## Impact

Cosmetic for the run list, since runs execute and complete normally and their output is readable. Blocking for the agent dashboard page, which fails to render at all.

I can capture the raw `GET /companies/:companyId/heartbeat-runs` response body if that would help pin down which side of the boundary is at fault.

VXNCXNX:
## Thinking Path

> - Every heartbeat run response passes through the run secret redaction service, which scrubs registered secret values before the data reaches the client
> - That service walks the payload recursively, but it treats any non string, non array object as a plain record, so it flattens `Date` instances to `{}`
> - The web UI then calls `new Date({})`, which gives `Invalid Date`, so run timestamps are unreadable and the agent dashboard fails to render at all
> - This pull request returns `Date` instances untouched, before the record branch
> - The benefit is that run timestamps render correctly again, and the agent dashboard loads, with no change to secret redaction behaviour

Fixes: #10964

## What Changed

- `server/src/services/run-secret-redaction.ts`: return `Date` instances untouched in `redactRegisteredSecretValues()`, before the record branch. A comment states why, because the failure is silent.
- `server/src/__tests__/run-secret-redaction.test.ts`: add a test that a top level `Date` survives, that `null` stays `null`, and that a sibling string is still redacted.
- `server/src/__tests__/run-secret-redaction.test.ts`: add a test that a `Date` survives inside an array and inside a nested object.

## Verification

Run the unit tests:

```
pnpm --filter @paperclipai/server test run-secret-redaction
```

I could not run the repository test runner in my environment. Instead I ran the pre-fix and the post-fix implementations side by side on one payload, to confirm the new tests assert a real difference and are not vacuous:

```
before  createdAt = {}                          isDate: false
after   createdAt = "2026-08-06T11:21:15.340Z"  isDate: true
after   nested in array                         isDate: true
after   nested in object                        isDate: true
after   stdoutExcerpt = "stdout [redacted]"     redaction still applied
```

On `master` the first assertion of each new test sees `{}` and fails. Please run the suite in CI to confirm.

Manual check on a self hosted instance (`ghcr.io/paperclipai/paperclip:latest`, commit `2ea22d6`, external Postgres):

1. Open an agent that has at least one completed heartbeat run.
2. Before the fix, the run list shows `Invalid Date` on every row. The agent dashboard shows an error boundary with `Invalid time value`.
3. Apply the compiled equivalent of this change. Restart the server.
4. After the fix, the run list shows relative timestamps. The agent dashboard renders.

The rows in Postgres were correct in both cases. The corruption happened on the way out.

## Risks

Low risk.

- The change is a single early return. It cannot alter the redaction of strings, arrays, or plain objects.
- It does not widen what reaches the client. A `Date` carries no secret material, and its own enumerable property set is empty, so the previous walk never redacted anything inside it. It only destroyed the value.
- No migration. No schema change. No API contract change, other than timestamp fields returning their real value instead of `{}`.
- One behavioural shift to be aware of: any client that adapted to `{}` in these fields now receives a real date. I found no such client in this repository.

A wider variant is possible. `asRecord()` could accept only plain objects, which would also cover `Map`, `Set`, `RegExp` and `Buffer` on this path. A `Buffer` does not flatten to `{}` but to an index keyed object, which is a quieter form of the same bug. I kept the narrow fix, because the blast radius is smaller on a security sensitive function. Tell me if you prefer the wider shape.

## Model Used

Claude Opus 5, model id `claude-opus-5[1m]`, 1M context window, extended thinking enabled, with tool use and code execution. The model read the compiled bundle on a running instance, traced the root cause, wrote the fix and the tests, and verified the before and after behaviour. A human reviewed and approved each step.

## Checklist

- [ ] I have run tests locally and they pass (see Verification: I verified the fix by running both implementations side by side on the same payload, but I could not run the repo test runner itself)
- [ ] I have updated relevant documentation to reflect my changes (no user facing documentation covers this internal function)

<sub>Duplicate search: I searched the PR list for `redaction`, `Date`, `run-secret`, `"Invalid Date"` and `redactRegisteredSecretValues`. The only other PR that touches this function is #9934 (closed), which added human approved secret proposals and is unrelated. No open or merged PR fixes this bug.</sub>

### DIFF
--- a/server/src/__tests__/run-secret-redaction.test.ts
+++ b/server/src/__tests__/run-secret-redaction.test.ts
@@ -54,4 +54,29 @@ describe("registered run secret redaction", () => {
     expect(redactRegisteredSecretValues("token-extended token", ["token-extended", "token"]))
       .toBe(`${REDACTED_EVENT_VALUE} ${REDACTED_EVENT_VALUE}`);
   });
+
+  it("preserves Date instances instead of flattening them to {}", () => {
+    const createdAt = new Date("2026-08-06T11:21:15.340Z");
+    const result = redactRegisteredSecretValues({
+      createdAt,
+      finishedAt: null,
+      stdoutExcerpt: `stdout ${secret}`,
+    }, [secret]);
+
+    expect(result.createdAt).toBeInstanceOf(Date);
+    expect(result.createdAt.toISOString()).toBe("2026-08-06T11:21:15.340Z");
+    expect(result.finishedAt).toBeNull();
+    expect(result.stdoutExcerpt).toBe(`stdout ${REDACTED_EVENT_VALUE}`);
+  });
+
+  it("preserves Date instances nested in arrays and objects", () => {
+    const at = new Date("2026-08-06T11:21:15.340Z");
+    const result = redactRegisteredSecretValues({
+      runs: [{ startedAt: at }],
+      nested: { deep: { updatedAt: at } },
+    }, [secret]);
+
+    expect(result.runs[0].startedAt).toBeInstanceOf(Date);
+    expect(result.nested.deep.updatedAt).toBeInstanceOf(Date);
+  });
 });
--- a/server/src/services/run-secret-redaction.ts
+++ b/server/src/services/run-secret-redaction.ts
@@ -41,6 +41,9 @@ function redactText(input: string, values: string[]) {
 
 export function redactRegisteredSecretValues<T>(input: T, values: string[]): T {
   if (typeof input === "string") return redactText(input, values) as T;
+  // Non-plain objects carry their value in internal slots, not in own enumerable
+  // properties, so the Object.entries walk below would flatten them to `{}`.
+  if (input instanceof Date) return input;
   if (Array.isArray(input)) return input.map((value) => redactRegisteredSecretValues(value, values)) as T;
   const record = asRecord(input);
   if (!record) return input;


VXNCXNX:
Thanks for the review.

On the one open item: the Thinking Path section is present in the PR description. Greptile reviewed commit `9b8decb`, which is the commit before I rewrote the description to follow `.github/PULL_REQUEST_TEMPLATE.md`. The description now carries Thinking Path, What Changed, Verification, Risks, Model Used, and the completed checklist including the duplicate search. `commitperclip` re-ran after that and reports all checks passing.

One thing I want to flag honestly, since it is easy to miss in the checklist: I could not run the repository test runner in my environment, so that box is deliberately left unchecked. I verified the two new tests are not vacuous by running the pre-fix and post-fix implementations side by side on the same payload; on `master` the first assertion of each test sees `{}` and fails. CI is the real confirmation here.

Also happy to widen the fix if a maintainer prefers it. Restricting `asRecord()` to plain objects would cover `Map`, `Set`, `RegExp` and `Buffer` on this path too. A `Buffer` does not flatten to `{}` but to an index keyed object, which is a quieter form of the same bug. I kept it narrow because this is a security sensitive function and the smaller blast radius seemed like the safer default, but I will change it on request.

VXNCXNX:
Superseded by #10984, which landed the same fix (`if (input instanceof Date) return input;` in the walker) plus the chat notice timestamp hardening. Closing this rather than rebasing it onto a tree that already has the change.

Glad it got fixed either way.

PR #10968, opened 2026-08-06
perf(db): index context_snapshot/payload issueId lookups used by recovery sweeps

cryppadotta:
## Why the server was slow (PAP-16575)

The 30-second heartbeat scheduler tick runs `reconcileStrandedAssignedIssues`, which iterates every candidate issue (currently **462**: 304 in_review + 76 in_progress + 82 todo) and, per issue, calls `getLatestIssueRun` / `getLatestIssueRunForAgent` / `hasActiveExecutionPath`. All of these filter `heartbeat_runs` on `context_snapshot ->> 'issueId'`, which has **no matching index**.

Measured on the live instance:

- Each `getLatestIssueRun` call walks ~26k rows via `heartbeat_runs_company_created_at_desc_idx` and **detoasts every row's `context_snapshot`** (the table is 84 MB heap + **2.1 GB TOAST**) to evaluate the JSONB filter: **645 ms per query** (`EXPLAIN ANALYZE`: `Rows Removed by Filter: 25816`, 130k buffers).
- ~766 such queries per sweep ≈ **500 s of DB work scheduled every 30 s** — the sweep can never finish before the next tick. `pg_stat_activity` sampling shows these two queries running in 15/15 and 12/15 samples; `heartbeat_runs` has accumulated 10.8 **billion** seq tuples read.
- `hasActiveExecutionPath` additionally scans `agent_wakeup_requests` (**1.8M rows**, 1 GB) on `payload ->> 'issueId'`, also unindexed.

## Fix

Add expression indexes (mirrored in the drizzle schema, hand-written migration 0209 following the 0206 precedent):

- `heartbeat_runs (company_id, (context_snapshot ->> 'issueId'), created_at DESC)`
- `heartbeat_runs (company_id, (context_snapshot ->> 'taskId'), created_at DESC)`
- `agent_wakeup_requests (company_id, (payload ->> 'issueId'))` (large-table pragma with justification)

These turn the per-issue lookups into index scans; the recovery sweep drops from ~500 s of DB work per tick to milliseconds.

## Verification

- `check:migrations` (numbering + safety) passes.
- New `heartbeat-context-snapshot-index-migration.test.ts` boots a fresh embedded Postgres, applies the full migration chain (0000→0209), asserts all three indexes exist, and asserts via `EXPLAIN` that the planner uses them for the exact hot query shapes.
- `tsc --noEmit` clean for packages/db.

### DIFF
--- a/packages/db/src/heartbeat-context-snapshot-index-migration.test.ts
+++ b/packages/db/src/heartbeat-context-snapshot-index-migration.test.ts
@@ -1,0 +1,42 @@
+import { describe, expect, it, afterEach } from "vitest";
+import postgres from "postgres";
+import {
+  getEmbeddedPostgresTestSupport,
+  startEmbeddedPostgresTestDatabase,
+} from "./test-embedded-postgres.js";
+
+const cleanups: Array<() => Promise<void>> = [];
+const support = await getEmbeddedPostgresTestSupport();
+const d = support.supported ? describe : describe.skip;
+
+afterEach(async () => {
+  while (cleanups.length > 0) await cleanups.pop()?.();
+});
+
+d("heartbeat context_snapshot expression index migration", () => {
+  it("applies full migration chain and uses the new indexes", async () => {
+    const dbh = await startEmbeddedPostgresTestDatabase("pap16575-idx-");
+    cleanups.push(() => dbh.cleanup());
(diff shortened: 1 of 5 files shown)


cryppadotta:
Superseded by #10969 (same commits, template-compliant description and branch name).

PR #10961, opened 2026-08-06
fix(test): boot embedded Postgres once per run so ~56 hidden server tests actually run (RBR-912)

PraeSynBH:
## Thinking Path

> - Paperclip is the open source app people use to manage AI agents for work. Its server holds the control-plane invariants: company scoping, atomic issue checkout, approval gates, budget stops.
> - Those invariants are guarded by DB-backed Vitest suites under `server/src/__tests__/`. Each one needs a real Postgres, which the suite provisions with an embedded cluster fixture.
> - The embedded cluster takes about 90 seconds to boot. The suites wrapped that boot in `beforeAll(fn, 20_000)`.
> - An inline `beforeAll` timeout argument silently overrides both the config `hookTimeout` and the `--hookTimeout` CLI flag. The hook could therefore never pass, no matter how the runner was configured.
> - Vitest reports a suite whose `beforeAll` fails this way as `skipped`, not as failed. So about 56 control-plane tests reported green while executing nothing at all, and the runner exited 0.
> - This is worse than having no tests. A missing test is visible. These tests looked present and were silently inert, so any regression in issue checkout or interaction supersession would have shipped unchallenged.
> - This pull request boots one shared cluster per run from a Vitest `globalSetup`, clones a migrated template database per suite, and deletes the inline hook budgets so the configured timeout is the one that applies.
> - The benefit is that the previously inert tests now execute (58 passing, 0 skipped) and the boot is paid once per run instead of once per suite.

## Linked Issues or Issue Description

No public GitHub issue tracks this. The underlying bug is described inline below, following
`.github/ISSUE_TEMPLATE/bug_report.yml`.

Related prior work: #10024 fixed flaky `afterAll` hook timeouts in the same server suites. It
raised hook budgets but did not remove the inline `beforeAll` overrides, so the silent-skip
behaviour survived it. This pull request removes that override path.

**What happened?**

About 56 DB-backed server tests were reported as `skipped` and never executed. The suites
`server/src/__tests__/issue-thread-interactions-service.test.ts` and
`issue-thread-interactions-telemetry.test.ts` provisioned an embedded Postgres cluster inside
`beforeAll(fn, 20_000)`. The cluster boot takes about 90 seconds, so the hook always exceeded its
20-second budget. An inline `beforeAll` timeout argument overrides both the `hookTimeout` config
value and the `--hookTimeout` CLI flag, so the budget could not be raised from outside the file.
Vitest reports a suite that fails a `beforeAll` this way as `skipped` rather than failed, and the
[...]

PR #10987, opened 2026-08-06
fix(test): centralize embedded-Postgres hook budget and guard zero-executed suites

PraeSynBH:
## Thinking Path

> - Paperclip is the open source app people use to manage AI agents for work. Its server holds the control-plane invariants: company scoping, atomic issue checkout, approval gates, budget stops.
> - Those invariants are guarded by DB-backed Vitest suites, each of which boots an embedded Postgres cluster inside `beforeAll`.
> - An inline `beforeAll(fn, timeoutMs)` argument silently overrides both the `hookTimeout` config value and the `--hookTimeout` CLI flag, so a per-file budget cannot be corrected from CI configuration.
> - Auditing every `startEmbeddedPostgresTestDatabase` call site found 150 of 152 inline hook budgets were set below the real ~92s import/boot cost (132 files at 20s, 14 at 30s, 4 at 60s). Two suites (`issue-thread-interactions-service.test.ts`, `issue-thread-interactions-telemetry.test.ts`) never reached their first assertion; Vitest reports a `beforeAll` failure of this kind as `skipped`, not failed, so the run still exited 0 and ~56 control-plane tests looked green while executing nothing.
> - A suite that cannot fail is worse than no suite: a real regression (an `issue-thread-interactions` default flip) shipped unchallenged because the file guarding it never ran.
> - This pull request centralizes the embedded-Postgres hook/teardown budget in one module, removes every inline per-file override so the centralized value always wins, and adds a CI-wired audit that fails the moment an inline budget reappears, plus a guard that fails a run reporting zero executed tests instead of letting it report success.
> - The benefit is that the previously-inert suites now execute, the budget can no longer drift per file, and a future zero-executed-suite regression fails loudly instead of silently.

No public GitHub issue tracks this. The underlying bug is described inline below, following `.github/ISSUE_TEMPLATE/bug_report.yml`.

Related: #10961 also touches embedded-Postgres test bootstrapping (a shared-cluster-per-run approach) but is unmerged as of this PR and does not remove the inline hook-budget overrides across the wider codebase — the two changes address adjacent but distinct parts of the same defect class and should be reconciled by whichever lands second.

**What happened?**
Two DB-backed server suites gave their embedded-Postgres `beforeAll` a 20-second inline budget while the real boot took ~92s. The hook always timed out. Vitest reports a suite whose `beforeAll` fails this way as `skipped`, not failed, and the run still exits 0 — so ~56 tests reported green while never executing. Auditing every embedded-Postgres call site for the same defect class found 150 of 152 inline hook budgets below the real cost, not just the two originally reported.

**Expected behavior**
Every suite that boots embedded Postgres actually executes its tests, the hook/teardown budget cannot drift per file, and a suite that reports zero executed tests fails the run instead of reporting success.

**Steps to reproduce**
1. Check out `master` before this change.
2. Run `server/src/__tests__/issue-thread-interactions-service.test.ts` or `issue-thread-interactions-telemetry.test.ts` in isolation.
3. Observe every test in the file reports `skipped` and the process exits 0.

Reproduces on `master` at `dea7c4e274` ("add company work timeline endpoint") onward — the commit that introduced the `beforeAll` inline timeout pattern for embedded Postgres.

**Agent adapter(s) involved**
Not adapter-specific (server test infrastructure).

## What Changed

- Add `scripts/embedded-postgres-test-budget.mjs` as the single source of truth for the embedded-Postgres hook budget (180s hook, 60s teardown, 120s audit floor) so per-file drift is structurally impossible.
- Have every project's `vitest.config.ts` (`server/`, `packages/db/`, others) read the centralized budget instead of hardcoding an inline value.
- Remove all 152 inline `beforeAll`/`afterAll` hook-budget arguments across the affected test files, including the shared route-test-harness, so the centralized config value is always the one that applies.
- Add `scripts/audit-embedded-postgres-hook-budgets.mjs`, which fails when an inline budget reappears below the floor, and wire it into `pr.yml` CI and `pnpm check:embedded-postgres-budgets`.
- Add a zero-executed-suites guard (`scripts/check-executed-test-suites.test.mjs` + implementation) so a suite that reports no executed tests fails the run instead of silently passing.
- This un-hides 58 previously-inert tests. Note for reviewers: 8 of those newly-live tests are expected to fail once a separate in-flight default-flip change (`supersedeOnUserComment`) lands — that is the guard correctly catching a real behavior change, not a regression in this PR. This PR does not touch that flip; it stays a pure test-infrastructure fix.

## Verification

- `node scripts/audit-embedded-postgres-hook-budgets.mjs` → exits 0, 164 db-starting hooks scanned, 0 under budget.
- `node --test scripts/__tests__/check-executed-test-suites.test.mjs` → 6/6 passing.
- Rebased cleanly onto current `origin/master` tip with 0 merge conflicts (`git merge-tree` reported zero conflict markers before and after rebase).

## Risks

- Low-to-moderate. The change is mechanical (remove inline overrides, read from one shared constant) and does not alter any embedded-Postgres boot/teardown logic itself, but it touches 151 files, so the surface area for a stray typo is nontrivial — mitigated by the audit script exiting 0 across the full tree.
- Un-hiding 58 previously-skipped tests is intentional and expected to surface real, previously-invisible failures (documented above). Do not treat a red run on those specific 8 tests as caused by this PR — it is the guard working as designed once the unrelated default-flip change lands.
- Overlaps `server/vitest.config.ts` with the still-open #10961. Whichever of the two merges second will need a small rebase; flagging this now so it is not a surprise.

## Model Used

Claude Opus 4.8 (Anthropic), model id `claude-opus-4-8` — original fix authored on a prior heartbeat with extended thinking and tool use / code execution. This PR (rebase, verification, and submission) was executed by Claude Sonnet (Anthropic), model id `claude-sonnet-5`, standard reasoning mode, with tool use / code execution, acting as release engineering on the already-authored and verified commit.

## Checklist

### DIFF
--- a/.github/workflows/pr.yml
+++ b/.github/workflows/pr.yml
@@ -58,6 +58,15 @@ jobs:
       - name: Test general-server shard partition
         run: node --test ./scripts/__tests__/run-vitest-stable-shard.test.mjs
 
+      # RBR-918: an inline `beforeAll` timeout wins over --hookTimeout, so an
+      # under-budgeted embedded-Postgres hook cannot be corrected from CI config.
+      # Fail the PR if one is reintroduced.
+      - name: Audit embedded-Postgres hook budgets
+        run: node ./scripts/audit-embedded-postgres-hook-budgets.mjs
+
+      - name: Test zero-executed-suite guard
+        run: node --test ./scripts/__tests__/check-executed-test-suites.test.mjs
+
       - name: Test e2e shard partition
         run: node --test ./scripts/__tests__/e2e-shard.test.mjs
 
--- a/.gitignore
+++ b/.gitignore
@@ -24,6 +24,7 @@ check-*.mjs
 !.github/scripts/check-*.mjs
 !.github/scripts/tests/check-*.mjs
 !scripts/check-*.mjs
+!scripts/__tests__/check-*.mjs
 new-agent*.json
 newcompany.json
 seed-*.mjs
--- a/cli/src/__tests__/company-import-export-e2e.test.ts
+++ b/cli/src/__tests__/company-import-export-e2e.test.ts
@@ -323,7 +323,7 @@ describeEmbeddedPostgres("paperclipai company import/export e2e", () => {
     });
 
     await waitForServer(apiBase, child, output);
-  }, 60_000);
+  });
 
   afterAll(async () => {
     await stopServerProcess(serverProcess);
--- a/cli/src/__tests__/routines.test.ts
+++ b/cli/src/__tests__/routines.test.ts
@@ -98,7 +98,7 @@ describeEmbeddedPostgres("disableAllRoutinesInConfig", () => {
     tempRoot = mkdtempSync(path.join(os.tmpdir(), "paperclip-routines-cli-config-"));
     configPath = path.join(tempRoot, "config.json");
     writeTestConfig(configPath, tempRoot, tempDb.connectionString);
-  }, 20_000);
+  });
 
   afterEach(async () => {
     await db.delete(routines);
--- a/cli/vitest.config.ts
+++ b/cli/vitest.config.ts
@@ -1,7 +1,15 @@
 import { defineConfig } from "vitest/config";
+import {
+  EMBEDDED_POSTGRES_HOOK_TIMEOUT_MS,
+  EMBEDDED_POSTGRES_TEARDOWN_TIMEOUT_MS,
+} from "../scripts/embedded-postgres-test-budget.mjs";
 
 export default defineConfig({
   test: {
     environment: "node",
+    // The CLI e2e suites boot embedded Postgres in beforeAll. Budget is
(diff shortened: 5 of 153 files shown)


PraeSynBH:
@greptile-apps review

This is a large but mechanical change (152 files): removing inline embedded-Postgres hook-budget overrides and centralizing them in scripts/embedded-postgres-test-budget.mjs, plus adding an audit script and a zero-executed-suites guard. Please bypass the 100-file limit and review.

PraeSynBH:
@greptile-apps review

Re-requesting review on the latest commit (55b46a0e6). Fixed both findings from the prior review: the audit now imports EMBEDDED_POSTGRES_MIN_HOOK_BUDGET_MS instead of duplicating the floor, and extractTimeout now distinguishes 'no argument' from 'non-literal expression' so the latter fails the audit instead of being waved through.

PR #10988, opened 2026-08-06
fix(server): null out activity_log.run_id when heartbeat_run row is missing

mentessaas:
## Thinking Path

> - The activity log records every server-side mutation so boards can audit what agents and users did.
> - `activity_log.run_id` is a nullable foreign key to `heartbeat_runs.id`, the table that tracks each agent run.
> - The activity log service currently inserts the caller-supplied `runId` without checking it exists in `heartbeat_runs`.
> - When a request carries a JWT whose `run_id` is not registered (out-of-band tokens, isolated worktrees without a heartbeat run row), the INSERT fails with a Postgres FK violation.
> - The exception escapes through `logActivity` and turns into HTTP 500 for the caller, even though the primary UPDATE/INSERT for the issue already committed.
> - This pull request validates `runId` against `heartbeat_runs` before insert, falls back to `null` when the row is missing, and emits a structured warning so the gap stays visible.
> - The benefit is that API callers stop seeing false 500s, the activity log still records the action, and operators get a clean signal in logs when an unregistered run id slips through.

No public issue exists for this bug. Following the bug-report template:

**What happened?**
PATCH or POST requests against the API occasionally return HTTP 500 even when the underlying mutation succeeded. The error originates inside `logActivity`: Postgres rejects the `activity_log` INSERT with a foreign-key violation because the request's `run_id` (taken from the JWT) does not exist in `heartbeat_runs`. The exception bubbles up past the transaction that already applied the issue update, so the client sees a server error but the data change persists.

**Expected behavior**
When a request carries an unregistered `run_id`, the activity row should still be recorded with `run_id = null`, the API should return the success response for the primary mutation, and the server log should contain a warning that identifies the gap so operators can investigate.

**Steps to reproduce**
1. Mint a JWT that carries an `X-Paperclip-Run-Id` header whose UUID is not registered as a row in `heartbeat_runs` (e.g. an out-of-band token, an isolated worktree without a heartbeat run).
2. Issue `PATCH /api/issues/<id>` (or any POST/PATCH that triggers an activity log write) using that JWT.
3. Observe the server returns `500 Internal server error`.
4. Check the database: the issue row was updated, but no `activity_log` row was inserted.

master @ `1fa36be35` (fix branch `fix/zal-384-activity-log-fk-guard` @ `14ba3a36a`)

**Deployment mode**
Local dev (`pnpm dev`), reproducible against any deployment that allows out-of-band JWTs.

**Logs**
```
DrizzleQueryError: Failed query: insert into "activity_log" (... run_id ...)
caused by: PostgresError: insert or update on table "activity_log" violates foreign key constraint "activity_log_run_id_heartbeat_runs_id_fk"
    at activity-log.ts:136 (logActivity)
    at routes/issues.ts (PATCH handler)
```

## What Changed

- Extract `resolveHeartbeatRunForActivity` helper that returns `{ responsibleUserId, runFound }` from a single `SELECT` against `heartbeat_runs`.
- `persistActivity` now consults `runFound` to decide whether to persist the supplied `runId` or fall back to `null`. When the run id is unregistered it emits a structured warning with the company, action, entity and run id for log searchability.
- `resolveResponsibleUserIdForActivity` accepts an optional `preResolvedRun` argument so `persistActivity` can reuse the same lookup without an extra DB round trip.
- Plugin event payload mirrors the persisted `runId` (was previously reading `input.runId` directly, which would diverge when the run was missing).
- New regression test in `activity-log-responsible-user.test.ts` logs an activity with an unregistered run id against an embedded Postgres database and asserts that the persisted row has `run_id = null` and falls back to the company default responsible user.

## Verification

1. `pnpm --filter @paperclipai/server exec vitest run src/__tests__/activity-log-responsible-user.test.ts` — 10/10 tests pass, including the new FK-guard regression test.
2. Manual repro (against the fix branch): repeat the reproduction steps above and confirm the API returns the expected success response for the issue update, an `activity_log` row is written with `run_id = null`, and a warning is logged at WARN level.
3. `tsc --noEmit -p server/tsconfig.json` runs clean.

## Risks

Low risk. The change is additive and only adjusts the value stored in `activity_log.run_id` when the supplied run id does not resolve. Backward compatibility is preserved because:
- The `runId` column is nullable, so storing `null` is the schema's documented fallback.
- `resolveResponsibleUserIdForActivity` gained an optional third argument; all existing callers compile and behave identically.
- The existing `resolveResponsibleUserIdForActivity` flow (which already falls back through `responsibleUserId` → `createdByUserId` → API key → company default when a run is missing) is unchanged.
- No migrations, schema changes, or API contract changes.

## Model Used

Claude (claude-opus-4-6) via Claude Code with extended thinking. Used for code analysis, fix authoring, and test authoring; all changes reviewed line-by-line.

## Checklist

### DIFF
--- a/server/src/__tests__/activity-log-responsible-user.test.ts
+++ b/server/src/__tests__/activity-log-responsible-user.test.ts
@@ -231,4 +231,50 @@ describeEmbeddedPostgres("logActivity responsible-user stamping", () => {
 
     expect(row?.responsibleUserId).toBe("key-user");
   });
+
+  it("stores runId as null when the heartbeat_run row is missing (FK guard)", async () => {
+    const companyId = randomUUID();
+    const agentId = randomUUID();
+    const unregisteredRunId = randomUUID();
+
+    await db.insert(companies).values({
+      id: companyId,
+      name: "Paperclip",
+      issuePrefix: `T${companyId.replace(/-/g, "").slice(0, 6).toUpperCase()}`,
+      defaultResponsibleUserId: "default-user",
+      requireBoardApprovalForNewAgents: false,
+    });
+    await db.insert(agents).values({
+      id: agentId,
+      companyId,
+      name: "CodexCoder",
+      role: "engineer",
+      status: "running",
+      adapterType: "codex_local",
+      adapterConfig: {},
+      runtimeConfig: {},
+      permissions: {},
+    });
+
+    // No heartbeat_runs row is inserted; the runId is unregistered.
+    // The guard must silently fall back to run_id = null instead of throwing
+    // a FK violation that would surface as a 500 to the API caller.
+    const activity = await logActivity(db, activityInput({
+      companyId,
+      actorId: agentId,
+      agentId,
+      runId: unregisteredRunId,
+    }));
+
+    expect(activity).toBeTruthy();
+
+    const row = await db
+      .select({ runId: activityLog.runId, responsibleUserId: activityLog.responsibleUserId })
+      .from(activityLog)
+      .where(eq(activityLog.id, activity.id))
+      .then((rows) => rows[0]);
+
+    expect(row?.runId).toBeNull();
+    expect(row?.responsibleUserId).toBe("default-user");
+  });
 });
--- a/server/src/services/activity-log.ts
+++ b/server/src/services/activity-log.ts
@@ -90,22 +90,51 @@ function readNonEmptyString(value: unknown) {
   return typeof value === "string" && value.trim().length > 0 ? value.trim() : null;
 }
 
-export async function resolveResponsibleUserIdForActivity(db: Db, input: LogActivityInput) {
+/**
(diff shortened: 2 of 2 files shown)


rendedennis6-byte:
Confirmed independently against a fresh clone of `master`: this is the cleanest fix for the `activity_log_run_id_heartbeat_runs_id_fk` 500 (out-of-band/unregistered `run_id` crashing `logActivity()` after the primary mutation already committed). Of the several open PRs attempting this fix, this one is the most complete — direct FK-guard-then-null-fallback approach, reuses the existing responsible-user lookup to avoid an extra round trip, structured warning log, and an embedded-Postgres regression test.

Only open item per the Greptile review is non-blocking (malformed non-UUID `runId` doesn't emit the warning) — worth a quick follow-up but shouldn't hold up merging this.

I don't have write access to this repo so I can't merge, but flagging for whoever has merge rights: this looks ready.

PR #11000, opened 2026-08-06
fix(hermes): give each agent its own AGENT_HOME, never a foreign agent's

PraeSynBH:
## Thinking Path

> - Each local adapter spawns an agent process and builds the environment that process runs in. `AGENT_HOME` is part of that environment, and it is the anchor for every memory path in an agent's operating instructions: `$AGENT_HOME/MEMORY.md`, `$AGENT_HOME/life/`, `$AGENT_HOME/memory/YYYY-MM-DD.md`.
> - The server already resolves the correct per-agent home for every run and publishes it on `context.paperclipWorkspace.agentHome`. `packages/adapter-utils/src/server-utils.ts` maps it to the `AGENT_HOME` variable. The codex, cursor, gemini, opencode, and pi local adapters all consume that path. The hermes adapter never did.
> - So a Hermes child received only whatever `AGENT_HOME` it inherited from the server process. On a host where a shell rc file exports a stale `AGENT_HOME`, that inherited value is a single fixed agent home, handed to every agent on every run.
> - The result is not a cosmetic variable problem. An agent that follows its instructions literally writes its memory into a different agent's directory. Two agents' daily notes collide on the same `memory/YYYY-MM-DD.md` file. Where the other agent is read-only oversight, the audited agent gets a writable path inside the auditor's home.
> - This pull request makes the hermes adapter resolve `AGENT_HOME` for each run and prove the value belongs to that run's own agent before it is passed to the child.
> - The benefit is that agent memory stays in the agent's own home, and a value that cannot be attributed is dropped loudly rather than silently corrupting another agent's memory.

No public GitHub issue exists for this. Describing it here, following `.github/ISSUE_TEMPLATE/bug_report.yml`.

**What happened**

Every agent run using the `hermes_local` adapter received the same `AGENT_HOME` value: the one inherited from the Paperclip server process. On a host whose shell rc file exported a stale `AGENT_HOME`, that meant every agent, on every run, was given one fixed foreign agent's home directory as its own memory root. This persisted across seven consecutive days of runs before it was noticed.

Because agent operating instructions define memory locations entirely relative to `$AGENT_HOME`, any agent following its instructions literally would write `MEMORY.md`, `life/`, and `memory/YYYY-MM-DD.md` into another agent's directory.

**What I expected to happen**

Each run receives the per-agent home that the server already resolves for it, so agents write memory only into their own directories.

**Steps to reproduce**

1. Configure an agent with `adapterType: hermes_local`.
2. In the environment of the Paperclip server process, set `AGENT_HOME` to some other agent's home directory. A stale `export AGENT_HOME=...` in a shell rc file does this.
3. Start a run for the agent from step 1.
4. Inspect the environment given to the spawned Hermes child. `AGENT_HOME` is the value from step 2, not the home belonging to the agent from step 1.

**Additional context**

The plumbing already existed and was already used elsewhere. Reference counts for `agentHome` / `AGENT_HOME` across the local adapters at the time of diagnosis:

| Adapter | references |
|---|---|
| hermes | **0** |
| codex-local | 2 |
| cursor-local | 3 |
| gemini-local | 3 |
| opencode-local | 3 |
| pi-local | 3 |

The hermes adapter was the only one that did not wire it. The defect was therefore not specific to any one agent — it applied to every agent using this adapter.

Removing the stale shell export changes the symptom from "a confidently wrong foreign home" to "no `AGENT_HOME` at all", which is safer but still wrong. The adapter fix is what actually resolves it.

## What Changed

- Added `resolveAgentHomeEnv()` to `packages/adapters/hermes/src/server/execute.ts`. It takes the inherited value, the run-resolved home, and the run's agent id, and returns the `AGENT_HOME` to use plus an optional operator warning.
- Applied it in `execute()` when building the child environment. The adapter now reads `context.paperclipWorkspace.agentHome`, which it previously ignored. When the helper returns no home, `AGENT_HOME` is deleted from the child environment rather than left at an inherited value.
- Added `agentHomeBelongsToAgent()`, a single ownership rule used by both the resolved and the inherited path: a home that belongs to an agent has that agent's id as its final path segment. Trailing slashes are ignored and both path separators are accepted.
- The ownership check is applied to the run-resolved value as well as the inherited one, so no single caller can bypass the invariant the helper exists to enforce.
- Warnings are logged through `ctx.onLog` on override and on drop, so a leaked or rejected value is visible in run logs instead of silent.
- Added `packages/adapters/hermes/src/server/execute.agent-home.test.ts` with 13 tests.

Behaviour, stated as three rules:

1. The run-resolved home wins over anything inherited.
2. A candidate home is used only when it is attributable to this agent. This applies to the run-resolved value and the inherited value alike.
3. A candidate that fails that check is dropped, not passed through. A missing `AGENT_HOME` is a loud, recoverable failure. A confidently wrong one silently corrupts another agent's memory.

## Verification

```
cd packages/adapters/hermes
npx vitest run src/server/execute.agent-home.test.ts
```

Result: 13 passed (13).

```
cd packages/adapters/hermes
npx tsc --noEmit -p tsconfig.json
```

Result: exit 0, no diagnostics.

The tests are behavioural rather than shape assertions. Each one calls the real `execute()` and reads the environment actually handed to the spawned child back out of the `runChildProcess` mock. What they cover:

- Three different agents each receive their own distinct home, and no agent's value contains another agent's id.
- The exact reported host condition: the server process environment leaks a foreign home, and the run still receives its own.
- An ordinary third agent is protected too, confirming the defect was not limited to the reporting pair.
- An unattributable inherited value is dropped rather than passed through.
- An inherited value that is demonstrably the run's own home is kept.
- A run-resolved value belonging to another agent is dropped, checked at both the `execute()` boundary and the helper.
- The operator warning is emitted and names the rejected value.

A reviewer can also confirm the negative case directly: revert `execute.ts` alone and the suite fails, because the child then receives the inherited foreign value.

## Risks

Low risk, and scoped to one adapter. The change is additive: 420 added lines across one new test file and one source file, with no lines removed and no change to any shared helper or other adapter.

Two behavioural shifts worth naming:

- **`AGENT_HOME` can now be absent where it was previously set.** When no run-resolved home exists and the inherited value cannot be attributed to the run's agent, the variable is deleted. This is deliberate. An absent variable fails loudly and recoverably, whereas a confidently wrong one writes into another agent's memory with no signal. Anything depending on `AGENT_HOME` always being populated, regardless of whether it was correct, would see the change.
- **The ownership check also applies to the run-resolved value.** This could in principle reject a legitimate home. It does not in practice: the server resolves an agent home as `<instance root>/workspaces/<agentId>`, whose final path segment is the agent id, so real resolved values pass. The check exists so that a future caller passing a foreign path cannot bypass the invariant.

No migrations, no schema or contract changes, no UI changes, no public API changes. `resolveAgentHomeEnv` is newly exported and has no prior callers.

## Model Used

Anthropic Claude, `claude-opus-4-6`, 200K context window, extended thinking enabled, with tool use and code execution. The model diagnosed the two contributing causes, wrote the helper and the behavioural test suite, and ran the tests and typecheck locally. A human reviewed the change.

## Checklist

### DIFF
--- a/packages/adapters/hermes/src/server/execute.agent-home.test.ts
+++ b/packages/adapters/hermes/src/server/execute.agent-home.test.ts
@@ -1,0 +1,298 @@
+/**
+ * Behavioural regression tests for AGENT_HOME isolation in the hermes adapter.
+ *
+ * The reported failure: for seven consecutive days, every run on one host
+ * started with `AGENT_HOME` pointing at a *fixed foreign agent's* home
+ * directory, regardless of which agent the run belonged to. Agent operating
+ * instructions define memory almost entirely in terms of `$AGENT_HOME`
+ * (`$AGENT_HOME/MEMORY.md`, `$AGENT_HOME/life/`,
+ * `$AGENT_HOME/memory/YYYY-MM-DD.md`), so any agent following its instructions
+ * literally would write its memory into that other agent's directory. Two
+ * distinct failures: same-day daily notes from two agents collide on one
+ * `memory/YYYY-MM-DD.md`, and a read-only oversight agent's home becomes
+ * writable by the party it audits.
+ *
+ * Root cause was twofold and both halves are asserted here:
+ *   1. This adapter never read `context.paperclipWorkspace.agentHome`, so the
+ *      only `AGENT_HOME` a child ever saw was the one inherited from the server
+ *      process environment.
+ *   2. The host's server process had a stale `export AGENT_HOME=<other agent>`
+ *      in a shell rc file, so that inherited value was a constant foreign
+ *      agent id.
+ *
+ * These assertions are behavioural: they inspect the environment actually handed
+ * to the spawned child, not the shape of a helper's return value. The core
+ * invariant — "a run for agent A never receives an AGENT_HOME belonging to agent
+ * B" — is asserted directly against that env.
+ */
+
+import { afterEach, describe, expect, it, vi, beforeEach } from "vitest";
+
+vi.mock("@paperclipai/adapter-utils/server-utils", async (importOriginal) => {
+  const actual = await importOriginal<typeof import("@paperclipai/adapter-utils/server-utils")>();
+  return {
+    ...actual,
+    runChildProcess: vi.fn(async () => ({
+      exitCode: 0,
+      signal: null,
+      timedOut: false,
+      stdout: "",
+      stderr: "",
+    })),
+  };
+});
+
+vi.mock("node:fs/promises", () => ({
+  readFile: vi.fn(async () => ""),
+  writeFile: vi.fn(async () => undefined),
+  mkdir: vi.fn(async () => undefined),
+  rm: vi.fn(async () => undefined),
+  access: vi.fn(async () => undefined),
+  readdir: vi.fn(async () => []),
+  stat: vi.fn(async () => ({ isFile: () => true, isDirectory: () => false })),
+}));
+
+import { execute, resolveAgentHomeEnv } from "./execute.js";
+import * as serverUtils from "@paperclipai/adapter-utils/server-utils";
+
+const COMPANY = "5cb37f67-a875-4073-ba4f-f8f942cb0775";
+const INSTANCE_ROOT = "/tmp/paperclip-test/instances/default";
+
+/**
+ * Agent ids from the original report, so the regression reads as the real
+ * incident. `OVERSIGHT` is the read-only oversight agent whose home leaked into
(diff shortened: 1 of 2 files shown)


PraeSynBH:
Superseded by #11002 (same fix, further along review chain). #11000 still embeds real production company/agent UUIDs in the test file, which CONTRIBUTING.md's "No Internal Issue References" section disallows; #11002 uses placeholder UUIDs throughout. Closing to avoid two open PRs for the same change.

PR #11026, opened 2026-08-07
fix(adapter-utils): bind /usr before merged-/usr symlink targets

### LINKED ISSUE #10684: Workspace sandbox fails to start on merged-/usr systems: /bin, /sbin, /lib are bound and symlinked at once
### Pre-submission checklist

### What happened?

With `filesystemScope: "workspace"`, the sandbox builder first recreates the
top-level directories as symlinks inside the new root:

```ts
args.push(
  "--symlink", "usr/bin", "/bin",
  "--symlink", "usr/sbin", "/sbin",
  "--symlink", "usr/lib", "/lib",
  "--symlink", "usr/lib64", "/lib64",
);
```

and then also bind-mounts every entry of `SYSTEM_READ_PATHS`, which contains
those same four paths plus `/usr`:

```ts
for (const systemPath of SYSTEM_READ_PATHS) await mount(systemPath, "ro");
```
(`packages/adapter-utils/src/local-process-sandbox.ts`, line 405 on `master`)

On **merged-/usr** distributions — Arch, Fedora, Debian ≥ bookworm,
Ubuntu ≥ 21.04, openSUSE, i.e. the majority of current Linux installations —
`/bin`, `/sbin`, `/lib` and `/lib64` are symlinks into `/usr`. The bind is
processed before `/usr` is mounted, so the symlink target does not yet exist in
the new root and bwrap aborts:

```
bwrap: Can't bind mount /oldroot/usr/bin on /newroot/bin:
       Unable to mount source on destination: No such file or directory
```

The sandbox is never created. Since it is the enforcement point for both
`filesystemScope` and `networkScope`, the practical difference is not "degraded
confinement" but **"the run does not start at all"**.

### Expected behavior

The sandbox should start on both merged and split `/usr` layouts.

The fix we run is to skip system paths that are symlinks, because the
`--symlink` arguments above already recreate them inside the sandbox:

```ts
for (const systemPath of SYSTEM_READ_PATHS) {
  if (await pathIsSymlink(systemPath)) continue;
  await mount(systemPath, "ro");
}
```
On split-/usr systems nothing changes — those paths are real directories and
are still bound. A patch is prepared and has been in continuous use here since
2026-07-29.

### Steps to reproduce

1. Use a merged-/usr Linux distribution (check with `ls -ld /bin` — it is a
   symlink to `usr/bin`).
2. Configure any `claude_local` agent with `filesystemScope: "workspace"`.
3. Start a run.
4. The sandbox fails to build with the bwrap error above; the run never starts.

On a split-/usr system the same configuration works, which is why this may not
have surfaced in CI.

e55d7029

### Deployment mode

Local dev (pnpm dev)

### Installation method

### Agent adapter(s) involved

- [x] Claude Code
- [ ] Codex
- [ ] Cursor
- [ ] Droid
- [ ] Hermes
- [ ] Not adapter-specific (core bug)

### Database mode

Not database-related

### Access context

None

### Node.js version

v26.4.0

### Operating system

CachyOS (Arch-based), kernel 7.1.5 — merged /usr

### Relevant logs or output

```shell
bwrap: Can't bind mount /oldroot/usr/bin on /newroot/bin: Unable to mount source on destination: No such file or directory
```

```json
{
  "adapterType": "claude_local",
  "adapterConfig": {
    "filesystemScope": "workspace"
  }
}
```

### Additional context

```
Duplicate search (2026-08-02): searching this repository's issues and pull
requests for "bwrap" returns no results, so nothing to link.

This was the first thing we had to patch locally in order to run Paperclip at
all on this machine; the fix has been in continuous use across several dozen
agent runs since 2026-07-29.

Note the ordering dependency rather than the symptom: the failure is not that
these paths must not be bound, but that they are bound before `/usr` exists in
the new root. Skipping symlinks is the smallest fix; mounting `/usr` first would
be another.
```

### Privacy checklist

dasgltd:
## Thinking Path

> - The local process sandbox (`filesystemScope: "workspace"`) confines each agent run with bubblewrap
> - On merged-/usr Linux distributions (Debian, Ubuntu, Fedora, Arch), `/bin`, `/sbin`, `/lib`, `/lib64` are symlinks into `/usr`, and the sandbox builder mounted them before `/usr` itself was bound
> - This left bwrap unable to resolve the symlink target, so it aborted with `Can't bind mount /oldroot/usr/bin on /newroot/bin: No such file or directory`
> - Every agent configured with `filesystemScope: "workspace"` failed to start on these distributions, making filesystem confinement effectively unusable on the most common Linux hosts
> - This pull request reorders `SYSTEM_READ_PATHS` so `/usr` is bound first, before the merged-/usr symlink targets that depend on it
> - The benefit is that workspace-confined agents start correctly on merged-/usr hosts, with no change in behavior on split-/usr hosts

Fixes: #10684

## What Changed

- Moved `/usr` to the front of `SYSTEM_READ_PATHS` in `packages/adapter-utils/src/local-process-sandbox.ts` so its read-only bind happens before the merged-/usr symlink targets (`/bin`, `/sbin`, `/lib`, `/lib64`) are bound.
- Added a regression test in `local-process-sandbox.test.ts` asserting that `/usr` is bound before all four merged-/usr symlink targets, using the `--ro-bind` argument index emitted by `buildLocalProcessSandboxSpawnTarget`.
- No change to split-/usr hosts: those paths are independent bind mounts, only their relative order changed.

## Verification

- Ran `buildLocalProcessSandboxSpawnTarget` directly and inspected the emitted `argv` before and after the change (indices of `--ro-bind /usr` vs `--ro-bind /bin`, `/sbin`, `/lib`, `/lib64`), confirming `/usr` now precedes all four.
- Ran the new test in `local-process-sandbox.test.ts`; it fails on `master` and passes with this change.
- Verified end-to-end on Ubuntu (kernel 7.0, bubblewrap 0.11.1): with the patch, `filesystemScope: "workspace"` agents start and run confined, with `--tmpfs /` and only the workspace, Claude config directory, and managed prompt bundle mounted. Without the patch, every run aborts with the bind-mount error above.

## Risks

- Low risk. The change is a reorder of entries in a fixed-order list, not a logic change — no new paths are mounted and no existing mount is removed.
- Split-/usr hosts (no `/bin`, `/sbin`, `/lib`, `/lib64` symlinks) are unaffected: those bind mounts are independent of `/usr` and were already order-independent.
- Worst case if this reorder were wrong: sandboxed agents would fail to start, the same failure mode this PR fixes — so a regression here is immediately visible, not silent.

## Model Used

Claude (Anthropic), claude-opus-5, via Claude Code CLI with standard tool use (file read/edit, local process execution to run and inspect `buildLocalProcessSandboxSpawnTarget` output and the test suite). No extended/chain-of-thought thinking mode used beyond the model's default reasoning.

## Checklist

### DIFF
--- a/packages/adapter-utils/src/local-process-sandbox.test.ts
+++ b/packages/adapter-utils/src/local-process-sandbox.test.ts
@@ -96,6 +96,35 @@ describe("local process sandbox", () => {
     expect(target.args.slice(-3)).toEqual([process.execPath, "-e", "console.log('ok')"]);
   });
 
+  it("binds /usr before the merged-/usr symlink targets", async () => {
+    const root = await fs.mkdtemp(path.join(os.tmpdir(), "paperclip-fs-usrmerge-"));
+    cleanup.push(root);
+    const workspace = path.join(root, "workspace");
+    await fs.mkdir(workspace);
+
+    const target = await buildLocalProcessSandboxSpawnTarget({
+      executable: process.execPath,
+      args: ["-e", "console.log('ok')"],
+      cwd: workspace,
+      options: { workspaceDir: workspace, filesystemScope: "workspace" },
+    });
+
+    const bindIndex = (candidate: string): number => {
+      for (let index = 0; index < target.args.length - 1; index += 1) {
+        if (target.args[index] === "--ro-bind" && target.args[index + 1] === candidate) return index;
+      }
+      return -1;
+    };
+
+    const usrIndex = bindIndex("/usr");
+    expect(usrIndex).toBeGreaterThanOrEqual(0);
+    for (const merged of ["/bin", "/sbin", "/lib", "/lib64"]) {
+      const mergedIndex = bindIndex(merged);
+      if (mergedIndex < 0) continue;
+      expect(mergedIndex).toBeGreaterThan(usrIndex);
+    }
+  });
+
   it("binds a confined absolute alias to the synchronized workspace", async () => {
     const root = await fs.mkdtemp(path.join(os.tmpdir(), "paperclip-fs-alias-"));
     cleanup.push(root);
--- a/packages/adapter-utils/src/local-process-sandbox.ts
+++ b/packages/adapter-utils/src/local-process-sandbox.ts
@@ -48,10 +48,17 @@ interface NetworkAllowlistProxy {
   close: () => Promise<void>;
 }
 
+// "/usr" must be bound before "/bin", "/sbin", "/lib" and "/lib64". On merged-/usr
+// hosts (Debian, Ubuntu, Fedora, Arch) those four are symlinks into usr/, and the
+// sandbox recreates them as symlinks in the new root via --symlink before binding.
+// Binding "/bin" first makes bwrap resolve /newroot/bin -> /newroot/usr/bin, which
+// does not exist until "/usr" itself is bound, and bwrap aborts. Ordering "/usr"
+// first keeps the symlink destinations resolvable, and is harmless on split-/usr
+// hosts. See #10684.
 const SYSTEM_READ_PATHS = [
+  "/usr",
   "/bin",
   "/sbin",
-  "/usr",
   "/lib",
   "/lib64",
   "/etc/ca-certificates",


jbrianfrancis-ir:
Sixth reproduction of #10684, and an independent check of this PR on a newer distribution and bubblewrap pair than the five already in the thread.

**Reproduction.** Ubuntu 26.04, kernel 7.0, bubblewrap 0.11.1, merged-`/usr` (`/bin`, `/sbin`, `/lib`, `/lib64` all symlinks into `usr/`). Same abort, reproduced standalone from the argv that `buildLocalProcessSandboxSpawnTarget` emits, with no adapter involved:

```
bwrap: Can't bind mount /oldroot/usr/bin on /newroot/bin: Unable to mount source on destination: No such file or directory
```

**On `master` the existing suite already catches this**, once the gated tests are enabled. That may be worth knowing while reviewing, since the failure needs no new test to be visible:

```
PAPERCLIP_TEST_BWRAP=/usr/bin/bwrap pnpm exec vitest run \
  --project @paperclipai/adapter-utils \
  packages/adapter-utils/src/local-process-sandbox.test.ts

2 failed | 11 passed | 1 skipped
  x prevents reads outside the workspace while allowing workspace writes
  x allows only configured network targets through the proxy bridge
```

Both fail with the abort above.

**With this PR applied**, same host, same command:

```
14 passed | 1 skipped
```

**One observation on the skip-based shape.** On this host both `/etc/localtime` and `/etc/resolv.conf` are symlinks (`-> ../usr/share/zoneinfo/...` and `-> ../run/systemd/resolve/stub-resolv.conf`). A rule that skips symlinked entries of `SYSTEM_READ_PATHS` therefore drops those two as well, and neither has a `--symlink` argument to fall back on. Measured here with bwrap directly, same argv except for those two `--ro-bind` triples:

```
dropped:  date -> UTC   cat: /etc/resolv.conf: No such file or directory
bound:    date -> EDT   # This is /run/systemd/resolve/stub-resolv.conf
host:     date -> EDT
```

Reordering needs no predicate at all, so it never has to draw that distinction.

jbrianfrancis-ir:
Independent reproduction and verification on **Ubuntu 26.04 with bubblewrap 0.11.1** — a newer distro/bwrap pair than the ones currently listed in #10684.

**Reproduced on master.** `filesystemScope: "workspace"` fails at sandbox startup:

```
bwrap: Can't bind mount /oldroot/usr/bin on /newroot/bin: Unable to mount source on destination: No such file or directory
```

`/bin`, `/sbin`, `/lib` and `/lib64` are all symlinks into `/usr` here, so the `--ro-bind /bin /bin` that comes from `SYSTEM_READ_PATHS` lands on the symlink created moments earlier by `--symlink usr/bin /bin`.

**This PR's ordering resolves it.** Binding `/usr` before the merged targets:

```sh
bwrap --tmpfs / --proc /proc --dev /dev --tmpfs /tmp \
  --ro-bind /usr /usr \
  --symlink usr/bin /bin --symlink usr/sbin /sbin \
  --symlink usr/lib /lib --symlink usr/lib64 /lib64 \
  --ro-bind /bin /bin --ro-bind /sbin /sbin \
  --ro-bind /lib /lib --ro-bind /lib64 /lib64 \
  /usr/bin/echo ok
```

succeeds, where the same argument set in master's current order fails with the error above.

One data point that may be useful when choosing between the open approaches: on this host `/etc/localtime` and `/etc/resolv.conf` are also symlinks (`-> ../usr/share/zoneinfo/<zone>` and `-> ../run/systemd/resolve/stub-resolv.conf`), and both are in `SYSTEM_READ_PATHS`. A predicate that skips *any* symlinked entry would therefore drop them, leaving the sandbox without timezone data or DNS. Reordering as done here sidesteps that entirely, since it needs no predicate.

For transparency: this is a direct bwrap-level verification of the argument ordering on the host described. I have not run the repository's test suite against this branch.

PR #11032, opened 2026-08-07
RBR-974: instance-wide agent-run admission control (rebased onto master)

PraeSynBH:
## RBR-974: instance-wide agent-run admission control (rebased onto current master)

Rebased/re-applied the verified RBR-974 fix (branch `rbr974-global-run-admission`, previously
based on `75d915b328`, 570 commits behind) onto current `origin/master`. This PR stages the
change so it is mergeable the instant the board approves the pending deploy confirmation on
RBR-974 — **it does not merge or restart anything.**

### What this does

`startNextQueuedRunForAgent` now calls an admission gate before claiming any queued run.
Refusal leaves runs `queued` for the existing `resumeQueuedRuns` sweep — queued, never dropped,
never started-then-killed.

Two design points:

1. **Cross-agent atomicity.** `withAgentStartLock` only serializes starts for *one* agent. Two
   different agents could each read "N of ceiling slots used" and both start, overshooting the
   ceiling. `withGlobalAdmissionLock` wraps the read-count-then-claim critical section so the
   ceiling check and the row claim are atomic across agents, not just atomic per agent.
2. **Forward-progress escape valve (CEO-accepted, keep as-is).** Strict load-refusal can
   deadlock: if load is not caused by us and stays above threshold, we defer every wake forever
   and nothing ever finishes to bring load down. When **zero** runs are live instance-wide, the
   gate admits exactly one, so a permanently-loaded host still drains its queue instead of
   stalling forever. The valve shuts the instant any run is live and cannot exceed the ceiling.

### Ceiling formula and rationale

`floor(cores * 0.75 / 1.55) = 5` on this host (12 cores).

- **1.55 load units per run** — derived empirically: 27 concurrent runs produced a sustained 1m
  load of 45.98 against a non-agent baseline of ~4 (API server, embedded-postgres, desktop):
  `(45.98 - 4) / 27 = 1.55`. A run is its own CLI process plus the tsc/vitest/embedded-postgres
  children our execution contract spawns, hence >1 core per run.
- **0.75 core budget** — the remaining 25% is reserved for the control plane the runs depend on
  (API server, postgres, operator session), so `/health` isn't competing on equal footing with
  the run pool.
- Five runs target `4 + 5*1.55 = 11.75` on 12 cores: fully utilised, not oversubscribed.
- This is a real throughput cut (27 observed → 5) and is called out explicitly in
  `RBR-974-VERIFICATION.md` for reviewers.

Full verification narrative, AC coverage, and the pre-existing-bug caveats (duplicate `0128`
migration number, unbuilt-workspace-package typecheck noise) are in `RBR-974-VERIFICATION.md`,
carried over unchanged from the original branch.

### Verification (narrow, per RBR-1045 — this host is still recovering from the
oversubscription this ticket is about; full suite / full-repo typecheck intentionally NOT run)

- `server/src/services/run-admission.test.ts` + `server/src/services/agent-start-lock.test.ts`:
  **35/35 passing** post-rebase.
  ```
  ✓ src/services/agent-start-lock.test.ts (6 tests)
  ✓ src/services/run-admission.test.ts (29 tests)
  Test Files  2 passed (2)
       Tests  35 passed (35)
  ```
- Touched-file typecheck (`run-admission.ts`, `agent-start-lock.ts`, `heartbeat.ts`): full
  `tsc --noEmit` was already run to completion twice pre-rebase (documented in
  `RBR-974-VERIFICATION.md`), zero errors in any file this branch touches. Re-running the same
  narrow typecheck post-rebase on this host (load ~24-30 on 12 cores) — result will be added to
  this PR/issue once it completes; not blocking PR-open per the ticket's "hold for approval, not
  gated on full clean CI" instruction.

### Rebase mechanics

- Cherry-picked the 7 commits unique to `rbr974-global-run-admission` (base `75d915b328`) onto
  current `origin/master`. Two commits touched `server/src/services/heartbeat.ts` and conflicted
  with unrelated master changes (import ordering, an unrelated extraction in the same function);
  both resolved by hand, preserving both sides' logic (git-credentials re-export +
  run-admission import; RBR-932's pause/terminate-race guard is untouched — this branch's changes
  land in a different function).
  - `fbd14d2b4a` — instance-wide agent-run admission control
  - `442de84d07` — forward-progress escape valve for external host load
  - `c71c44adcf` — verification report and open decisions for CEO
  - `cc51b02bd6` — extract admitAndClaimQueuedRuns from the lock closure
  - `8aee12ddef` — record typecheck results and the remaining verification gap
  - `45317e61a5` — test the global admission lock and the concurrent-dispatch ceiling
  - `25c80320e6` — record A/B baseline proving the integration failure is pre-existing

### Do NOT merge yet

Per RBR-974 and RBR-1045: **hold this PR open** until RBR-974's board confirmation
(`534bd467-93ad-4205-a575-ecd643b05d92`) is accepted. Merging enables dispatch-path behaviour
instance-wide and requires a restart of `paperclipai run`. Neither has happened as part of this
PR.

Refs: RBR-974, RBR-1045

### DIFF
--- a/RBR-974-VERIFICATION.md
+++ b/RBR-974-VERIFICATION.md
@@ -1,0 +1,88 @@
+## CTO: admission control landed on a branch. Ceiling is **5** on this 12-core host.
+
+Acknowledging your measurements — they changed my approach in one specific way: **I did not trust slot accounting alone.** Your 69.6% `rg` datapoint proves the instance cannot see all the load it competes with, so the gate refuses on *measured host load* as well as on slot count. Slot math would have said "4 of 5 free, go ahead" at load 46.
+
+Branch `rbr974-global-run-admission`, worktree `/tmp/rbr974-admission`, base `75d915b328` (master).
+Commits, posted immediately per RBR-941:
+- `fbd14d2b4aea7905a5166853959fcb2557e998d8` — admission control
+- `442de84d07e003e0c486e94446f84e7ac21aab02` — forward-progress escape valve
+
+**No instance-config edit and no server restart from me. Nothing is deployed.** Your escalation gate is not tripped by what is on this branch; it will be tripped by *merging* it, and I am asking for that separately below.
+
+### AC5 — the chosen ceiling and the reasoning
+
+`floor(cores * 0.75 / 1.55)` = `floor(12 * 0.75 / 1.55)` = **5**.
+
+Both constants are derived from your numbers, not picked:
+
+- **1.55 load units per run.** Your 27 runs produced a sustained 1m load of 45.98 against a non-agent baseline of ~4 (API server, embedded-postgres, Chrome, desktop). `(45.98 - 4) / 27 = 1.55`. A run is its own CLI process *plus* the tsc/vitest/embedded-postgres children our execution contract asks it to spawn — that is why it costs more than one core.
+- **0.75 core budget.** The remaining 25% is reserved for the control plane the runs depend on: API server, postgres, operator session. This is what makes your `/health` canary defensible — the API is no longer competing on equal footing with the run pool.
+
+Five runs target `4 + 5*1.55 = 11.75` on 12 cores. Fully utilised, not oversubscribed. That is the intent.
+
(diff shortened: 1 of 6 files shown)


PraeSynBH:
Closing in favor of a correctly-named branch (per CONTRIBUTING.md branch naming rules — internal ticket ids should not appear in public branch names). Re-opening as a new PR from fix/instance-wide-agent-run-admission-control.

PR #10284, opened 2026-07-26
Security: Add redaction of sensitive fields in Paperclip config-read API responses

JackReis:
## Summary
- Add redaction logic to mask token, authToken, devicePrivateKeyPem, apiKey, secret, and password fields
- Redaction applied recursively to nested objects and arrays
- Case-insensitive field matching
- Comprehensive test coverage for list/get agent endpoints

## Test plan
- All MCP server tests pass
- Sensitive fields are replaced with '***REDACTED***' in API responses
- Non-sensitive fields remain visible
- Redaction works for nested objects

PR #4763, opened 2026-04-29
fix(server): redact adapter_config secrets in agent detail responses

jayfeeney:
## Thinking Path

> - Paperclip orchestrates AI agents for zero-human companies, with each agent storing its connection material (API keys, device tokens, gateway-header tokens, private keys) in a JSON `adapter_config` column on the `agents` table
> - Several authenticated HTTP endpoints serialise full agent rows for self-view, admin-view, and admin-list callers
> - While running an identity-discovery code path, an agent on a downstream deployment posted its own configuration verbatim into a public comment, exposing the secret-bearing fields in `adapterConfig` in plaintext
> - Investigation traced the leak to four endpoints (`GET /api/agents/me`, `GET /api/agents/:id` self/admin, `PATCH /api/agents/:id` permissions response, `GET /api/companies/:companyId/agents` admin) all calling unredacted detail builders, while sibling restricted-view endpoints already redacted via `redactForRestrictedAgentView` / `redactAgentConfiguration`
> - Existing redaction primitives have two coverage gaps for this fix: `redactEventPayload`'s key-name regex matches conventional secret naming but misses `deviceToken` (no separator) and arbitrary `headers` map keys like `x-openclaw-token`; and `redactForRestrictedAgentView` nukes `adapterConfig` entirely, which works for restricted viewers but breaks the `/me` heartbeat path agents call for runtime metadata
> - This pull request adds a shared `buildSafeAgentDetail()` builder backed by an `omitAdapterConfigSecrets()` primitive — explicit field-omit on secret-bearing top-level keys, then `redactEventPayload` pass for nested env-style secrets — wired into all four leaking endpoints
> - The benefit is defence-in-depth on a real leak class: the four endpoints now redact `adapterConfig` secrets across all auth tiers (self, admin, list) while preserving non-secret fields needed by the UI and by running agents on the heartbeat path

## What Changed

- New `omitAdapterConfigSecrets()` primitive in `server/src/redaction.ts` (explicit field-omit for OpenClaw secret keys: `devicePrivateKeyPem`, `deviceToken`, `headers`, `sessionKey`; followed by `redactEventPayload` pass for nested env-style secrets)
- New `withSafeAdapterConfig()` + `buildSafeAgentDetail()` helpers in `server/src/routes/agents.ts` (the latter delegates to the existing `buildAgentDetail()` then applies `withSafeAdapterConfig()` to the result)
[...]

Issue #1818, opened 2026-03-26
SECURITY BREACH WARNING: GET /api/companies/{id}/agents leaks plaintext secrets from agent adapter env vars


olivier1977:
- The endpoint returns full adapterConfig including env vars as plaintext
- Private keys, API keys, and other secrets in agent env vars are fully exposed to any authenticated agent
- Requested fix: redact secret values (mask as "***") in API responses, or exclude the env field from list responses
- Impact: any agent can read the secrets of any other agent in the same company

jasondbramley:
ITO security review confirms this needs to cover `GET /api/companies/{companyId}/agents` list responses as well as any detail/configuration endpoints.

Acceptance criteria requested:

- Server-side response shaping must exclude or redact `adapterConfig.env` and equivalent credential-bearing sub-fields from agent list and detail responses.
- A safe replacement is fine, for example env key names only or a boolean such as `hasEnvConfig: true`; plaintext values must never be returned to normal agent/user inspection flows.
- Add regression coverage using an agent fixture with an inline env value, asserting the list response and detail/config response do not include the secret value.
- Treat this as a platform security fix; downstream ITO agents will avoid calling broad registry endpoints for now, but that is only an interim workaround.

Local tracker: ITO-3427. QA & Compliance will verify once the Paperclip fix is deployed.

jasondbramley:
Follow-up from the ITO tracker: there is an older note here saying fixed in commit `64ff13a`, but ITO still needs deployment confirmation before closing its local security tracker.

Can the Paperclip team confirm whether the redaction fix is deployed to the runtime serving ITO, and whether current regression coverage includes:

- `GET /api/companies/{companyId}/agents` list responses
- agent detail/configuration responses
- `adapterConfig.env` and equivalent credential-bearing `runtimeConfig` fields
- plain-value env bindings as well as sensitive key-name matching

ITO has created a QA verification step against the live runtime using a safe synthetic value only. No real secret values will be printed or recorded.

jasondbramley:
ITO live verification has passed against the current runtime.

Summary from QA:

- `GET /api/companies/{companyId}/agents`: `adapterConfig.env` absent from all 28 returned agents.
- Cross-agent `GET /api/agents/{id}`: env absent.
- Self-read `GET /api/agents/me`: returns own full config by design; current env contained only non-secret runtime path data.
- ITO confirms the #1818 protection scope is deployed and working for the agent registry/detail surfaces it depends on.

[...]

PR #11046, opened 2026-08-07
fix(secrets): store openclaw_gateway credentials as secret refs

spuhaha18:
## Thinking Path

> - Agents are configured per adapter, and `adapterConfig` regularly carries credentials; the secrets service is responsible for turning those literals into secret refs on write (`normalizeAdapterConfigForPersistence`)
> - Which fields get that treatment comes from `listAdapterSchemaSecretFieldKeys()`: the adapter's `getConfigSchema()` fields marked `meta.secret === true`, unioned with a hardcoded `FALLBACK_ADAPTER_SCHEMA_SECRET_FIELDS` map
> - The `openclaw_gateway` adapter exports only `execute` and `testEnvironment` — it has no `config-schema.ts` and no `getConfigSchema()` — so the schema branch is skipped entirely and only the fallback map applies
> - The fallback map had a single entry, `hermes_gateway: ["apiKey"]`, so **no** `openclaw_gateway` field was ever normalized, and its gateway credentials were persisted verbatim into `agents.adapter_config`
> - This pull request adds the `openclaw_gateway` credential fields to that fallback map so they become secret refs on write, matching `hermes_gateway`
> - The benefit is that OpenClaw gateway shared tokens, passwords, and device private keys stop living as plaintext in the agents table, and inherit the existing binding/audit/rotation machinery

No issue exists for this specific gap, so describing it here per `CONTRIBUTING.md` (bug report shape):

**What happened** — Creating or updating an agent with `adapterType: "openclaw_gateway"` and credential values in `adapterConfig` (`authToken`, `password`, `devicePrivateKeyPem`) stores those values as plaintext strings in `agents.adapter_config`. No `company_secret_bindings` rows are created, so the values are outside the secrets manager: not encrypted at rest in the agents table, not rotatable, not covered by per-agent secret access, and not audited on resolution.

**Expected behavior** — Same as `hermes_gateway`: literal credential strings passed to the agent create/update APIs are converted into `secret_ref` values with backing `company_secret_bindings`, and rehydrated only at runtime by `resolveAdapterConfigForRuntime()`.

**Steps to reproduce**
1. `POST /api/companies/{companyId}/agents` with `adapterType: "openclaw_gateway"` and `adapterConfig: { url: "ws://…", authToken: "<literal>", devicePrivateKeyPem: "-----BEGIN PRIVATE KEY-----…" }`
2. Read the persisted row: `select adapter_config from agents where id = …`
3. Both literals are present verbatim; `select * from company_secret_bindings where target_id = …` returns zero rows
4. Doing the same with `adapterType: "hermes_gateway"` and `apiKey` yields `{"type":"secret_ref",…}` plus a binding row

**Version/commit** — reproduced on `master` (`4683f26`). **Deployment mode** — self-hosted server, `pnpm dev`.

**Related PRs** (read-side redaction; complementary to this change, not duplicates — they hide already-stored plaintext on `GET`, this stops it being stored in the first place):
- Refs #9823 — `fix(security): redact adapterConfig secrets on all agent read endpoints`
- Refs #4763 — `fix(server): redact adapter_config secrets in agent detail responses`
- Refs #10284 — `Security: Add redaction of sensitive fields in Paperclip config-read API responses`
- Refs #1818 — `SECURITY BREACH WARNING: GET /api/companies/{id}/agents leaks plaintext secrets from agent adapter env vars` (adjacent: that issue is about `adapterConfig.env`, which is already normalized; this PR covers the top-level `openclaw_gateway` credential fields, which were not)

I searched open and closed PRs for `openclaw`, `adapter secret`, `devicePrivateKeyPem`, and `openclaw secret ref`, and code-searched `FALLBACK_ADAPTER_SCHEMA_SECRET_FIELDS`. Nothing in flight touches this map.

## What Changed

- `server/src/services/secrets.ts`: added an `openclaw_gateway` entry to `FALLBACK_ADAPTER_SCHEMA_SECRET_FIELDS` listing `authToken`, `token`, `password`, `devicePrivateKeyPem`.
  - `authToken` / `token`: both are read as the shared gateway token by `resolveAuthToken()` in the adapter (`nonEmpty(config.authToken) ?? nonEmpty(config.token)`), so both are covered.
  - `password`: read at `execute.ts` and `test.ts` as the gateway shared password.
  - `devicePrivateKeyPem`: read by `resolveDeviceIdentity()` and fed to `crypto.createPrivateKey()`.
- `server/src/__tests__/agents-service-secret-bindings.test.ts`: added a regression test mirroring the existing Hermes gateway case — asserts the three UI-settable credential fields persist as `secret_ref`, that the literals do not appear anywhere in the persisted config, that exactly the expected `company_secret_bindings` config paths are created, that non-secret fields (`url`) are untouched, and that `resolveAdapterConfigForRuntime()` returns the original plaintext.

No adapter-side change is needed: `resolveAdapterConfigForRuntime()` rehydrates the plaintext before `execute()` is called.

## Verification

Automated, from `server/`:

```
vitest run src/__tests__/agents-service-secret-bindings.test.ts
#  Test Files  1 passed (1)
#  Tests  10 passed (10)
#  ✓ converts OpenClaw gateway credential strings into persisted secret refs

vitest run src/__tests__/secrets-service.test.ts src/__tests__/adapter-registry.test.ts src/__tests__/company-portability.test.ts
#  Test Files  3 passed (3)
#  Tests  141 passed (141)

tsc --noEmit -p tsconfig.json    # clean
```

The new test runs against embedded Postgres (`getEmbeddedPostgresTestSupport`), so it exercises real persistence and real binding rows rather than mocks.

Manual, on a self-hosted instance with a live `openclaw_gateway` agent:

1. Confirmed the stored `adapter_config` contained a literal gateway token and a full `-----BEGIN PRIVATE KEY-----` PEM, and that both were returned in the agents read response.
2. Applied this patch and restarted the server.
3. `PATCH`ed the same agent with identical `adapterConfig` values; both fields flipped to `{"type":"secret_ref","version":"latest",…}`.
4. Re-read the agents endpoint: zero occurrences of the token string or `BEGIN PRIVATE KEY`.
5. Triggered a real heartbeat wake against the OpenClaw gateway afterwards — the connection authenticated and the run completed, confirming runtime resolution still hands the adapter plaintext.

## Risks

Low risk, but two things worth calling out:

- **Migration of existing rows is not automatic.** Agents configured before this change keep their plaintext values until the config is written again (any agent update re-runs normalization). This PR deliberately does not add a backfill migration; that would be a separate, larger change. Combined with the read-side redaction PRs linked above, already-stored values also stop being served over the API.
- **`token` is undocumented.** `agentConfigurationDoc` lists `authToken` and `password` but not `token`; `token` only exists as an alias inside `resolveAuthToken()`. It is included here so the alias cannot be used to smuggle a credential past normalization. If a maintainer would rather keep the map strictly to documented fields, dropping `token` is a one-word change.
- Behaviour for every other adapter type is unchanged — the map is keyed by adapter type and nothing else was touched.

Longer term the cleaner fix is for the `openclaw_gateway` adapter to ship a `config-schema.ts` with `meta: { secret: true }` on these fields, like `packages/adapters/hermes/src/gateway/server/config-schema.ts` does, which would also drive the config UI. That is a bigger change with UI implications (the adapter currently has a hand-written `src/ui`), so this PR takes the minimal path that the fallback map already exists for. Happy to follow up with the schema version if maintainers prefer it.

## Model Used

- Claude Opus 5 (`claude-opus-5`), Anthropic, via Claude Code CLI
- Extended thinking enabled; tool use (filesystem, shell, test execution)
- The change, the test, and this description were produced with model assistance and reviewed and executed locally by a human-supervised session; all test and typecheck output quoted above is real output from this machine, not reproduced from memory

## Checklist

- [ ] I have updated relevant documentation to reflect my changes — no user-facing docs describe this normalization map; `agentConfigurationDoc` field descriptions are unchanged by this PR
- [ ] All Paperclip CI gates are green — will confirm once CI reports
- [ ] Greptile is 5/5 with no open P2s, recommendations, or follow-ups — will address on review

### DIFF
--- a/server/src/__tests__/agents-service-secret-bindings.test.ts
+++ b/server/src/__tests__/agents-service-secret-bindings.test.ts
@@ -262,6 +262,77 @@ describeEmbeddedPostgres("agent service secret binding sync", () => {
     expect(JSON.stringify(persistedConfig)).not.toContain(literalApiKey);
   });
 
+  it("converts OpenClaw gateway credential strings into persisted secret refs", async () => {
+    const companyId = await seedCompany();
+    const literalAuthToken = `openclaw-token-${randomUUID()}`;
+    const literalPassword = `openclaw-password-${randomUUID()}`;
+    const literalPrivateKeyPem = [
+      "-----BEGIN PRIVATE KEY-----",
+      `openclaw-device-key-${randomUUID()}`,
+      "-----END PRIVATE KEY-----",
+    ].join("\n");
+
+    const created = await agentService(db).create(companyId, {
+      name: "OpenClaw Gateway",
+      role: "engineer",
+      status: "idle",
+      adapterType: "openclaw_gateway",
+      adapterConfig: {
+        url: "ws://127.0.0.1:18789",
+        authToken: literalAuthToken,
+        password: literalPassword,
+        devicePrivateKeyPem: literalPrivateKeyPem,
+      },
+      runtimeConfig: {},
+      spentMonthlyCents: 0,
+      lastHeartbeatAt: null,
+    });
+
+    const persistedRows = await db
+      .select()
+      .from(agents)
+      .where(eq(agents.id, created.id));
+    const persistedConfig = persistedRows[0]?.adapterConfig as Record<string, unknown>;
+    const serializedConfig = JSON.stringify(persistedConfig);
+    expect(serializedConfig).not.toContain(literalAuthToken);
+    expect(serializedConfig).not.toContain(literalPassword);
+    expect(serializedConfig).not.toContain(literalPrivateKeyPem);
+    for (const key of ["authToken", "password", "devicePrivateKeyPem"]) {
+      expect(persistedConfig[key]).toMatchObject({
+        type: "secret_ref",
+        version: "latest",
+      });
+    }
+    expect(persistedConfig.url).toBe("ws://127.0.0.1:18789");
+
+    const bindings = await db
+      .select()
+      .from(companySecretBindings)
+      .where(and(
+        eq(companySecretBindings.companyId, companyId),
+        eq(companySecretBindings.targetType, "agent"),
+        eq(companySecretBindings.targetId, created.id),
+      ));
+    expect(bindings.map((binding) => binding.configPath).sort()).toEqual([
+      "authToken",
+      "devicePrivateKeyPem",
+      "password",
+    ]);
+
+    const resolved = await secretService(db).resolveAdapterConfigForRuntime(
+      companyId,
+      persistedConfig,
+      {
+        consumerType: "agent",
+        consumerId: created.id,
+      },
+      { adapterType: "openclaw_gateway" },
+    );
+    expect(resolved.config.authToken).toBe(literalAuthToken);
+    expect(resolved.config.password).toBe(literalPassword);
+    expect(resolved.config.devicePrivateKeyPem).toBe(literalPrivateKeyPem);
+  });
+
   it("replaces agent secret bindings when adapterConfig env changes", async () => {
     const companyId = await seedCompany();
     const secrets = secretService(db);
--- a/server/src/__tests__/agents-service-secret-bindings.test.ts
+++ b/server/src/__tests__/agents-service-secret-bindings.test.ts
@@ -266,6 +266,7 @@ describeEmbeddedPostgres("agent service secret binding sync", () => {
     const companyId = await seedCompany();
     const literalAuthToken = `openclaw-token-${randomUUID()}`;
     const literalPassword = `openclaw-password-${randomUUID()}`;
+    const literalDeviceToken = `openclaw-device-token-${randomUUID()}`;
     const literalPrivateKeyPem = [
       "-----BEGIN PRIVATE KEY-----",
       `openclaw-device-key-${randomUUID()}`,
@@ -281,6 +282,7 @@ describeEmbeddedPostgres("agent service secret binding sync", () => {
         url: "ws://127.0.0.1:18789",
         authToken: literalAuthToken,
         password: literalPassword,
+        deviceToken: literalDeviceToken,
         devicePrivateKeyPem: literalPrivateKeyPem,
       },
       runtimeConfig: {},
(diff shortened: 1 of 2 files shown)


spuhaha18:
Thanks — the `deviceToken` finding is valid and is now fixed in 8dbec8b.

Verified before applying: `packages/adapters/openclaw-gateway/src/server/execute.ts` reads `ctx.config.deviceToken` at line 1071 and sends it in the gateway auth payload at lines 1252/1255, alongside `authToken` and `password`. It is genuine credential material and belongs in the list.

I also re-enumerated every `config.*` read in the adapter's `execute.ts` and `test.ts` to check for further gaps. The complete set of top-level credential fields is `authToken`, `token`, `deviceToken`, `password`, `devicePrivateKeyPem` — all five are now covered. The remaining config fields (`url`, `clientId`, `role`, `scopes`, `sessionKey`, `payloadTemplate`, timeouts, etc.) are not credentials.

One known gap left deliberately out of scope: a token can also be supplied nested as `headers["x-openclaw-token"]`, which `resolveAuthToken()` falls back to. Nested header values are not handled by `FALLBACK_ADAPTER_SCHEMA_SECRET_FIELDS`, which only walks top-level keys. Covering that needs a different mechanism than this map, so I left it for a follow-up rather than widening this PR — happy to open a separate one if maintainers want it.

Re-ran locally after the change:

```
vitest run src/__tests__/agents-service-secret-bindings.test.ts \
  src/__tests__/secrets-service.test.ts \
  src/__tests__/adapter-registry.test.ts \
  src/__tests__/company-portability.test.ts
#  Test Files  4 passed (4)
#  Tests  151 passed (151)

tsc --noEmit -p tsconfig.json    # clean
```

markstones78:
Thanks for this — the fallback map is exactly what the `openclaw_gateway` adapter needs (and it is the only proposal so far that includes `deviceToken`).

One interaction worth covering in the same change, observed when running it: once `devicePrivateKeyPem` is persisted as a secret reference, the stored value is a binding object rather than a PEM string. `ensureGatewayDeviceKey` in `server/src/routes/agents.ts` checks `asNonEmptyString(adapterConfig.devicePrivateKeyPem)`, so on the next agent update (any PATCH, e.g. changing `timeoutSec`) it treats the key as absent and generates a fresh Ed25519 key. The device id changes silently and the gateway then rejects the previously issued device token (`device token mismatch`).

Suggested fix shape: treat a `secret_ref` / `user_secret_ref` (or a `plain` binding with a value) as a configured key before falling back to generation, e.g.

```ts
function hasConfiguredGatewayDeviceKey(value: unknown): boolean {
  if (typeof value === "string") return value.trim().length > 0;
  if (!value || typeof value !== "object" || Array.isArray(value)) return false;
  const binding = value as Record<string, unknown>;
  if (binding.type === "secret_ref" || binding.type === "user_secret_ref") return true;
  if (binding.type === "plain") return typeof binding.value === "string" && binding.value.trim().length > 0;
  return false;
}
```

Happy to open a small follow-up PR on top of this one if that is easier than folding it in. (Related: device-token-only reconnect also needs the signature fix in #13219.)

PR #11086, opened 2026-08-08
fix(server): throttle and back off heartbeat scheduler errors (#10911)

### LINKED ISSUE #10911: Heartbeat timer error path writes unbounded log volume when Postgres is unavailable — 5.35 GB in 130 seconds
### Pre-submission checklist

### What happened?

Fields 4–10 are short answers; I've filled what we verified and flagged the two I'm guessing at.

markdown
### 1. What happened?

When the embedded Postgres became unreachable while the server was running, the heartbeat timer's failure path logged without throttling or backoff. It wrote **5.35 GB in 130 seconds** (~41 MB/s, roughly 15k failures/sec).

Byte offsets into `server.log`:

| Offset | Line |
|---|---|
| 149,793,770 | `[11:43:00] ERROR: heartbeat timer tick failed` |
| 5,500,846,533 | `[11:45:10] INFO: Server listening on <host>:3100` |

Everything before 11:43:00 is ~36 days of ordinary use (~150 MB, ~4 MB/day). The remaining 5.35 GB was written in the 130-second window between those two lines.

At that rate an unattended occurrence fills a 145 GB volume in about 48 minutes, taking down everything else on the host. The failure is self-amplifying: the database being down is survivable, filling the disk is not.

### Expected behavior

Repeated identical failures should be rate-limited, deduped, or backed off, so that a database outage degrades the instance without threatening the host filesystem. Log volume should stay bounded regardless of failure duration.

### Steps to reproduce

1. Run a self-hosted instance with embedded Postgres.
2. While the server is running, make Postgres unreachable. In my case `systemd-logind` reaped the `/dev/shm` PostgreSQL segments (see field 13); killing the Postgres process or removing the segments manually should have the same effect.
3. Leave it in that state for ~2 minutes.
4. `ls -l <instance>/logs/server.log`

I have not tried to reproduce deliberately — this was observed in production and the instance has since been shut down. The trigger is incidental; the amplification should follow from any cause of Postgres unavailability.

`paperclipai@2026.626.0` (npm). Note npm runs ahead of the GitHub release tags — the latest tag was `v2026.618.0` at the time.

### Deployment mode

Self-hosted server

### Installation method

npm / pnpm global install

### Agent adapter(s) involved

- [ ] Claude Code
- [ ] Codex
- [ ] Cursor
- [ ] Droid
- [ ] Hermes
- [x] Not adapter-specific (core bug)

### Database mode

Embedded PGlite (default — `DATABASE_URL` unset)

### Access context

Unclear / not applicable

### Node.js version

v20.20.2

### Operating system

Ubuntu 24.04 LTS

### Relevant logs or output

```shell
First line of the burst:

`
[11:43:00] ERROR: heartbeat timer tick failed
`

Repeating payload throughout (truncated):

`
"message": "Failed query: select \"routine_triggers\".\"id\", \"routine_triggers\".\"company_id\", \"routine_trigg…
`

Underlying Postgres failure during the window:

`
could not open shared memory segment "/PostgreSQL.…"
`

Log-level attribution over the whole 5.5 GB file (bytes, including continuation lines of multi-line entries):

`
     5499631739  ERROR
        1255397  INFO
          44474  WARN
            688  DEBUG
`

99.98% of the file is that one event. Each failure entry is ~2.4 KB, carrying the full failing statement plus context.
```

```json

```

### Additional context

**Trigger (environment-specific; the amplification is not).** The service runs as a regular user under systemd. `systemd-logind` defaults to `RemoveIPC=yes`, which reaps that user's POSIX IPC — including the embedded Postgres `/dev/shm` segments — when their last login session ends. Postgres then fails with `could not open shared memory segment`. Fixed locally with `RemoveIPC=no` in `logind.conf`.

Possibly worth a deployment-docs note: self-hosted installs running as a regular user should set `RemoveIPC=no` (this is what the Debian/Ubuntu `postgresql` package does) or run under a system user, or the embedded Postgres loses its shared memory whenever the operator's last login session ends.

**Suggested fixes**

1. Rate-limit or dedupe repeated identical errors in the timer failure path (log first occurrence, then a periodic summary).
2. Back off on consecutive failures rather than retrying at full tick rate.
3. Don't log the full SQL statement on every retry — once per distinct error is enough.
4. Ship a log rotation/size cap default so no single failure mode can exhaust the disk.

**Related issues**

- #6931 — disk thresholds and heartbeat auto-pause. Approaches the same failure from the consequence side (an instance reaching 100% disk, Postgres then unreachable). That post-mortem attributes the fill to accumulated backup/run-log files; given the write rate documented above, this bug is worth ruling out as a contributing cause.
- #2022 — retention policies and server log rotation. Covers suggested fix #4 but not the amplification itself.
- #1825 — full `req`/`res` serialization on the file transport. Different failure mode (memory), same logger.

### Privacy checklist

wakqasahmed:
## Thinking Path

> - The server runs a background heartbeat scheduler that ticks timers, scheduled triggers, and sweepers on a fixed interval.
> - When the underlying Postgres database is unreachable, every tick fails, and each failure was logged in full, with no rate limit and no backoff.
> - This let a single sustained outage write gigabytes of logs per minute (5.35 GB in 130 seconds in the reported case), which can fill the disk and take the whole host down even though the database outage itself is survivable.
> - This pull request adds an `ErrorThrottler` that deduplicates and rate-limits repeated identical background error messages, and adds exponential backoff to the heartbeat scheduler interval on consecutive tick failures.
> - The benefit is that a database outage degrades the instance gracefully instead of amplifying into a disk-filling incident.

Fixes: #10911

## What Changed

- Added `server/src/lib/error-throttler.ts` (`createErrorThrottler`): summarizes and truncates error messages, logs the first occurrence of a distinct error in full, suppresses identical repeats within a configurable window (default 5 minutes), and emits a periodic summary line with the suppressed count instead.
- Wired `schedulerThrottler.logError(...)` into every background scheduler/sweeper catch handler in `server/src/index.ts` (heartbeat timers, routine triggers, decision expiry/retention sweeps, status-card scheduler, secret proposal sweep, tool connection health sweep, external-object scheduler, merged pull-request confirmation sweep, terminal workspace reaper, periodic heartbeat recovery) in place of the previous unthrottled `logger.error(...)` calls.
- Reworked `startHeartbeatSchedulerInterval` to use `setTimeout`-based self-rescheduling instead of a fixed `setInterval`, tracking `consecutiveSchedulerFailures` and applying exponential backoff (doubling the interval per consecutive failure, capped at 5 minutes), resetting to the base interval on the next successful tick.
- Updated scheduler teardown to clear both the timer and interval handle (`clearTimeout`/`clearInterval`) since the handle type changed.
- Added `server/src/__tests__/error-throttler.test.ts` covering summarization, truncation, suppression window, periodic summary emission, distinct-key isolation, and reset behavior.
- Added `server/src/__tests__/heartbeat-scheduler-throttling.test.ts` exercising the throttler against the intended backoff sequence.

## Verification

- `pnpm run test:run -- error-throttler.test.ts heartbeat-scheduler-throttling.test.ts` — new unit/integration tests pass locally.
- `pnpm run test:run` — full server test suite run to confirm no regressions from the scheduler rework.
- Manual reasoning check: with the throttler in place, a sustained Postgres outage now logs one full error, then only periodic suppressed-count summaries (at most every 5 minutes per distinct error), and the scheduler backs off from its 30s base interval up to a 5-minute cap instead of retrying at full speed.

## Risks

- Greptile flagged that the outer heartbeat-tick callback's own local `.catch()` handlers on individual sub-tasks (e.g. `tickTimers`, `tickScheduledTriggers`) consume their errors before they reach the outer `await callback()` in `startHeartbeatSchedulerInterval`, so `consecutiveSchedulerFailures` is only incremented when the outer callback itself rejects — most individual sub-task failures do not currently drive the backoff counter. Log volume is still bounded by the `ErrorThrottler` in all cases, but the scheduler-interval backoff itself may not engage for those failure paths as intended. Flagging this as a known follow-up rather than blocking the log-volume fix, since the throttler is the primary mitigation for the reported disk-filling behavior.
- Low risk otherwise: the change is additive (new throttling wrapper + interval rescheduling), does not alter business logic of the scheduled jobs themselves, and preserves existing shutdown/cleanup behavior.

## Model Used

Claude (Anthropic), Claude Sonnet 4.5 — standard mode, tool use enabled, no extended thinking.

## Checklist

- [x] I have searched GitHub for duplicate or related PRs and linked them above (no exact duplicate found; related but distinct: #7583, #10276, #10288, #6704 — all address backoff/throttling in different subsystems, not scheduler-tick log-volume throttling)

### DIFF
--- a/server/src/__tests__/error-throttler.test.ts
+++ b/server/src/__tests__/error-throttler.test.ts
@@ -1,5 +1,5 @@
 import { describe, expect, it, vi } from "vitest";
-import { createErrorThrottler, summarizeError } from "../lib/error-throttler.js";
+import { computeBackoffDelayMs, createErrorThrottler, summarizeError } from "../lib/error-throttler.js";
 import type { Logger } from "pino";
 
 function mockLogger(): Logger {
@@ -33,6 +33,14 @@ describe("summarizeError", () => {
     expect(summarizeError("Plain text error")).toEqual({ message: "Plain text error", code: undefined });
     expect(summarizeError({ custom: "err" })).toEqual({ message: '{"custom":"err"}', code: undefined });
     expect(summarizeError(null)).toEqual({ message: "Unknown error", code: undefined });
+    expect(summarizeError(undefined)).toEqual({ message: "Unknown error", code: undefined });
+  });
+
+  it("preserves falsy-but-defined thrown values instead of treating them as unknown", () => {
+    // JS allows `throw 0` / `throw ""` / `throw false`; these are not "no error".
+    expect(summarizeError(0)).toEqual({ message: "0", code: undefined });
+    expect(summarizeError("")).toEqual({ message: "", code: undefined });
+    expect(summarizeError(false)).toEqual({ message: "false", code: undefined });
   });
 });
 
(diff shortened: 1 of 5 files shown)


wakqasahmed:
@greptileai review

PR #5540, opened 2026-05-09
fix(agents): drop deprecated adapterConfig.cwd from new hires

KingxBTC:
## Thinking Path

> - Paperclip orchestrates AI agents for zero-human companies
> - Local adapters (`claude_local`, `codex_local`, etc.) historically had an `adapterConfig.cwd` field that pointed to a working directory on the host machine
> - Paperclip now leases execution workspaces per issue, making `cwd` semantically obsolete for new agents
> - But the hire endpoint, adapter docs, and the `paperclip-create-agent` skill still accepted or referenced `cwd`, so newly hired agents could still carry the deprecated field
> - A prior hire shipped with a stale `cwd` pointing at a user-local clone instead of the project workspace — semantically misleading and a future failure mode
> - This pull request strips `cwd` from new-agent payloads at the API boundary, marks it deprecated in all adapter config schemas and docs, and removes it from the hire skill templates
> - The benefit is that no future hire can accidentally carry this legacy field, while existing agents that already have `cwd` are left untouched for backward compatibility

## What Changed

- **API boundary defense** (`server/src/routes/agents.ts`): Added `ADAPTER_TYPES_WITH_DEPRECATED_CWD` set and `stripDeprecatedCwdForNewAgent()` helper. Wired into `POST /api/companies/:companyId/agent-hires` and `POST /api/companies/:companyId/agents` to silently drop `cwd` before persistence, emitting a `console.warn` + `logActivity` event (`agent.hire_dropped_deprecated_field` / `agent.create_dropped_deprecated_field`). PATCH route intentionally untouched for backward compat.
- **Adapter config schemas** (`packages/adapters/*/src/index.ts`): Marked `cwd` as `**deprecated**` in `agentConfigurationDoc` for all 7 local adapters: `acpx-local`, `claude-local`, `codex-local`, `cursor-local`, `gemini-local`, `opencode-local`, `pi-local`.
- **Adapter docs** (`docs/adapters/claude-local.md`, `codex-local.md`, `gemini-local.md`): Flipped `cwd` from `Required: Yes` to `No` and labeled it deprecated.
- **`paperclip-create-agent` skill** (`skills/paperclip-create-agent/`): Removed `cwd` from example hire payloads in `SKILL.md` and `references/api-reference.md`. Added explicit "do not include `adapterConfig.cwd`" note. Updated `references/draft-review-checklist.md` to flag `cwd` presence as a review failure.
[...]

PR #11124, opened 2026-08-09
fix(environments): remote-only adapterConfig.cwd + delete from the UI

Fl0p:
## Thinking Path

> - Paperclip can run an agent on a different host: an SSH environment or a managed sandbox.
> - An operator sets such an environment up in the environment editor, then points an agent at it.
> - Three separate defects sit on that path. A remote-only `adapterConfig.cwd` kills every run of the agent. The editor cannot save an environment on a local_trusted instance. An environment can never be deleted from the UI.
> - Each defect is small, but together they make remote environments hard to set up and impossible to clean up.
> - This pull request fixes all three: the adapter execute paths become execution-target aware, the environment secret context falls back to the instance's only company, and the editor gets a Delete action.
> - The benefit is that an operator can create, fix and remove a remote environment without an API client.

No public issue exists for these. The descriptions follow `bug_report.yml`.

**What happened?**

1. An agent is bound to an SSH environment. Its `adapterConfig.cwd` holds a path that exists only on the remote host. Every run fails in seconds with `EACCES: permission denied, mkdir '/...'` raised on the Paperclip host, before Paperclip opens the SSH connection. The error looks remote but it is local.
2. Clicking Save in the environment editor of a local_trusted instance always returns 422 `Environment secret management requires a companyId context during the instance-scoped transition`. No environment can be saved from the UI.
3. The environment editor has no Delete action. An environment can only be removed with a direct `DELETE` API call.

**Expected behavior**

1. For a remote execution target Paperclip must not create `adapterConfig.cwd` on the Paperclip host, must take the remote working directory from the environment `remoteWorkspacePath`, and must use the local agent workspace or agent home as the sync source.
2. An instance admin must be able to save an environment. The instance has one company, so the company context is unambiguous.
3. An operator must be able to delete an environment from the UI, and must see what the deletion affects first.

**Steps to reproduce**

1. Create an environment with `driver: ssh` and set its `remoteWorkspacePath`.
2. Hire an agent with `workspaceStrategy: agent_home` and bind it to that environment.
3. Set the agent `adapterConfig.cwd` to a directory that exists only on the SSH host. Start a run. The run fails with `EACCES` on the Paperclip host.
4. Open the environment in the editor and click Save. The request fails with 422.
5. Look for a Delete action in the editor. There is none.

**Agent adapter(s) involved**

Claude Code, Codex, Cursor, Gemini, Grok, OpenCode and Pi. Hermes runs locally only and is not affected. Defects 2 and 3 are not adapter-specific.

Reproduced on `master` at `19be4cf92`.

**Deployment mode**

Self-hosted single instance. Local trusted mode. PostgreSQL.

Related pull requests found in a search for `adapterConfig.cwd`. None of them covers the remote-target case: #5540, #4968, #4844, #2007, #1223, #84.

## What Changed

**Adapter execution targets**

- Gate the local `ensureAbsoluteDirectory(cwd, { createIfMissing: true })` call on a local execution target in `claude-local`, `cursor-local`, `gemini-local`, `grok-local`, `opencode-local`, `pi-local` and the shared `acpx-engine` path. `codex-local` already had this gate.
- Gate `useConfiguredInsteadOfAgentHome` on `!executionTargetIsRemote` in all seven adapters and in `acpx-engine`. For a remote target `adapterConfig.cwd` no longer replaces agent home as the local staging directory.
- Read the execution target before the `cwd` resolution in `claude-local`, `codex-local` and `acpx-engine`, because the new gate needs it earlier. That change moves declarations only.
- Keep the `workspace.resolve` startup step in `acpx-engine` on both lanes, so the span shape does not change for a remote target.

**Environment secret company context**

- Add the instance's single company as the final fallback in `resolveEnvironmentSecretContextCompanyId`, for a board actor that is an instance admin or the implicit local actor. `resolveCustomImageCompanyId` in the same file already resolves its context this way. The fallback runs only for callers that require a company context, so the environment probe keeps its own explicit-`companyId` requirement for secret-backed environments. Instances with several companies stay ambiguous and still require an explicit `companyId`.

**Delete from the UI**

- Add `deleteBlastRadius` and `remove` to the environments API client. Both server routes already exist.
- Add a Delete action to the environment edit form. It opens a confirmation dialog, loads the blast radius, lists what still references the environment, and stays disabled while the server reports the environment as undeletable.

**Docs and tests**

- Document the local meaning and the remote meaning of `adapterConfig.cwd` in `docs/adapters/overview.md`.
- Add a `claude-local` regression test proving a remote-only `adapterConfig.cwd` is neither created on the local host nor used as the sync source.
- Add two `CompanyEnvironments` tests: a successful delete through the dialog, and a blocked delete for an instance-default environment.

## Verification

- `pnpm exec vitest run packages/adapters/claude-local/src/server/execute.remote.test.ts` passes with 4 tests, including the new regression test.
- `pnpm exec vitest run ui/src/pages/CompanyEnvironments.test.tsx` passes with 22 tests, including the two new delete tests.
- `pnpm --filter @paperclipai/adapter-utils --filter @paperclipai/adapter-claude-local --filter @paperclipai/adapter-codex-local --filter @paperclipai/adapter-cursor-local --filter @paperclipai/adapter-gemini-local --filter @paperclipai/adapter-grok-local --filter @paperclipai/adapter-opencode-local --filter @paperclipai/adapter-pi-local typecheck` passes for all eight packages.
- `pnpm exec vitest run server/src/__tests__/environment-routes.test.ts` passes with 68 tests, including the two probe tests that assert the explicit-`companyId` requirement stays in place.
- `pnpm --filter @paperclipai/server typecheck` and `pnpm --filter @paperclipai/ui typecheck` both pass.
- On a self-hosted instance built from this branch: `PATCH /environments/:id` without a `companyId`, the exact call the editor makes, returned 422 before the change and 200 after it. Deleting an environment through the new dialog removed it and refreshed the list.

## Risks

Low risk.

The adapter change is a guard that only takes effect when the execution target is remote; local targets keep the previous code path. For a remote target `adapterConfig.cwd` no longer overrides agent home as the local staging directory. That is an intentional behavior change, and an operator who set that value for a remote agent had a failing run before, so no working setup depends on the old behavior.

The company fallback applies only to a board actor that is an instance admin or the implicit local actor, only when the instance has exactly one company, and only on the code path that requires a company context. Other actors, multi-company instances and the probe path are unchanged.

The delete action calls a route that already existed and already enforces its own permission and blast-radius rules. The UI only refuses earlier than the server does.

There is no database migration and no API change.

## Model Used

Claude Opus 5 (`claude-opus-5`), 1M context window, extended thinking, running in Claude Code with tool use and code execution. An earlier draft of the adapter fix came from Claude Opus 4.8 in the same harness. Both models are credited in the commit trailers.

## Checklist

### DIFF
--- a/docs/adapters/overview.md
+++ b/docs/adapters/overview.md
@@ -53,6 +53,29 @@ Worked examples:
   `credentials.json` from the sandbox image's own `$HOME/.claude`. The
   snapshot's Claude login is the credential source for the run.
 
+### Working directory (`cwd`) on remote targets
+
+The `cwd` in an adapter's config (`adapterConfig.cwd`) is interpreted relative
+to **where the CLI actually runs**:
+
+- **Local targets** — `adapterConfig.cwd` is the local working directory the
+  CLI is launched in (created if missing).
+- **Remote targets (SSH / managed sandbox)** — the remote working directory
+  comes from the environment's `remoteWorkspacePath`, **not** from
+  `adapterConfig.cwd`. On every run Paperclip stages the agent's local
+  workspace/agent-home, syncs it into
+  `<remoteWorkspacePath>/.paperclip-runtime/runs/<runId>/workspace`, runs there,
+  and syncs changes back. For a remote target `adapterConfig.cwd` is therefore
+  **not** used as a local filesystem path.
+
+Do not set `adapterConfig.cwd` to a path that only exists on the remote host
+expecting it to select the remote directory — use the environment's
+`remoteWorkspacePath` for that. A remote-only `cwd` is ignored for local
+filesystem purposes: it is never `mkdir`-ed on the Paperclip host and never used
+as the local staging source. (Previously a remote-only `cwd` caused the run
+to `mkdir` that path on the Paperclip host and die with `EACCES` before any SSH
+connection was made.)
+
 ### Hermes local vs gateway
 
 Use `hermes_local` when Paperclip should start the local `hermes` CLI on the
--- a/packages/adapter-utils/src/acpx-engine/execute.ts
+++ b/packages/adapter-utils/src/acpx-engine/execute.ts
@@ -1522,8 +1522,15 @@ async function buildRuntime(input: {
   const workspaceBranch = asString(workspaceContext.branchName, "");
   const workspaceWorktreePath = asString(workspaceContext.worktreePath, "");
   const agentHome = asString(workspaceContext.agentHome, "");
+  const executionTarget = readAdapterExecutionTarget({
+    executionTarget: input.ctx.executionTarget,
+    legacyRemoteExecution: input.ctx.executionTransport?.remoteExecution,
+  });
+  const remoteExecutionIdentity = adapterExecutionTargetSessionIdentity(executionTarget);
+  const executionTargetIsRemote = remoteExecutionIdentity !== null;
   const configuredCwd = asString(config.cwd, "");
-  const useConfiguredInsteadOfAgentHome = workspaceSource === "agent_home" && configuredCwd.length > 0;
+  const useConfiguredInsteadOfAgentHome =
+    !executionTargetIsRemote && workspaceSource === "agent_home" && configuredCwd.length > 0;
   const effectiveWorkspaceCwd = useConfiguredInsteadOfAgentHome ? "" : workspaceCwd;
   const cwd = effectiveWorkspaceCwd || configuredCwd || process.cwd();
   // Referenced (additional) projects to stage into the sandbox alongside the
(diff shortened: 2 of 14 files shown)


Fl0p:
Closing this one.

Since it was opened, upstream has moved: environment deletion from the UI now exists on `master` (blast-radius dialog, blocked-reason handling, reusable-sandbox-lease consent), which supersedes that half of this PR. The companyId-context fix that earlier revisions carried was covered by #11291. What is still unique here is the remote-only `adapterConfig.cwd` handling for ssh / managed-sandbox targets - the local `ensureAbsoluteDirectory(cwd)` that fails with EACCES before any ssh happens, and `adapterConfig.cwd` being used as the local staging dir for a remote-only path.

We carry that in our fork. Happy to reopen or re-cut it as a focused single-fix PR if there is interest.

PR #11128, opened 2026-08-09
fix(server): reject unsupported monitor-scheduling fields on PATCH /issues/:id (RBR-1101)

PraeSynBH:
## Summary
RBR-1094 diagnosed that `PATCH /issues/:id` silently accepts and drops `monitorNextCheckAt` / `monitorNotes` / `monitorScheduledBy` (flat) and `executionState.monitor.*` (nested) — HTTP 200, no error, zero persistence. `updateIssueSchema` isn't `.strict()`'d, so Zod's default `.parse()` silently strips these unrecognized keys before the handler ever sees them.

Per the CEO-approved scope in RBR-1101, this ships a narrow fix rather than flipping `updateIssueSchema` to global `.strict()` (which would newly reject any other currently-silently-ignored key across every PATCH caller company-wide).

## Changes
- `packages/shared/src/validators/issue.ts`: add `findUnsupportedMonitorSchedulingFields()`, which inspects the *raw* request body (ahead of Zod stripping) for the flat monitor-scheduling keys and the nested `executionState.monitor` key, returning the names of any present.
- `server/src/routes/issues.ts`: add `rejectUnsupportedMonitorSchedulingFields` middleware ahead of `validate(updateIssueRouteSchema)` on `PATCH /issues/:id`; returns a 400 (`badRequest`) naming the unsupported field(s) with `code: "unsupported_monitor_scheduling_fields"` instead of silently no-opping.
- Test coverage:
  - `packages/shared/src/validators/issue.test.ts`: unit tests for `findUnsupportedMonitorSchedulingFields` covering each flat field, all together, the nested key, legitimate payloads, and malformed input.
  - `server/src/__tests__/issue-recovery-actions.test.ts`: integration test against the real `PATCH /issues/:id` route asserting 400 + field names for both flat and nested shapes, that the request is rejected outright (no partial write), and that a legitimate payload without these fields still returns 200 (no regression).

## Acceptance criteria (RBR-1101)
- AC-a: PATCH `/issues/:id` with any of these fields now returns a 4xx naming the unsupported field(s) — verified via new tests.
- AC-b: existing legitimate PATCH payloads are unaffected — full `issue-recovery-actions.test.ts` suite (45/45) passes with no regressions.
- AC-c: new/updated test coverage proving AC-a — see tests above.
- AC-d: parent RBR-1094 comment + RBR-954 verification to follow as a separate step once this merges.

## Testing
- `packages/shared`: `npx vitest run src/validators/issue.test.ts` → 37/37 pass
- `server`: `npx vitest run src/__tests__/issue-recovery-actions.test.ts` → 45/45 pass (full suite)
- `packages/shared`: `npx tsc --noEmit` → clean
- `server`: `npx tsc --noEmit` → no new errors on `routes/issues.ts` or `validators/issue.ts` (remaining errors are pre-existing `@paperclipai/plugin-sdk` build-dependency issues unrelated to this change)

Closes RBR-1101.

### DIFF
--- a/packages/shared/src/index.ts
+++ b/packages/shared/src/index.ts
@@ -1691,6 +1691,7 @@ export {
   issueBlockedInboxSeveritySchema,
   issueBlockedInboxStateSchema,
   updateIssueSchema,
+  findUnsupportedMonitorSchedulingFields,
   stalledReviewDecisionSchema,
   issueExecutionPolicySchema,
   issueExecutionStateSchema,
--- a/packages/shared/src/validators/index.ts
+++ b/packages/shared/src/validators/index.ts
@@ -410,6 +410,7 @@ export {
   issueBlockedInboxSeveritySchema,
   issueBlockedInboxStateSchema,
   updateIssueSchema,
+  findUnsupportedMonitorSchedulingFields,
   stalledReviewDecisionSchema,
   issueExecutionPolicySchema,
   issueExecutionStateSchema,
--- a/packages/shared/src/validators/issue.test.ts
+++ b/packages/shared/src/validators/issue.test.ts
@@ -3,6 +3,7 @@ import { MAX_ISSUE_REQUEST_DEPTH } from "../index.js";
 import {
   addIssueCommentSchema,
   createIssueSchema,
+  findUnsupportedMonitorSchedulingFields,
   issueBlockedInboxAttentionSchema,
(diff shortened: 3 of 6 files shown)


PraeSynBH:
Superseded — reopening with a branch name that doesn't reference an internal ticket id, and a PR body matching the required template.

PR #11143, opened 2026-08-10
feat(connections): resolve per-user connection grants at gateway execution

purpleCowOnWheels:
## Summary

Personal-only connections (Gmail, Calendar, a user-scoped Slack grant) have no legitimate non-personal identity, but the tool gateway's execution path only ever attached a connection-level credential. Paperclip already has the full per-user data model and API for this -- `connection_grants` with user subjects, self-only `start-authorization`, a fail-closed `mintConnectionTokenForAgent`, and the "Connect your account" interaction card -- but none of it was wired into the path agents actually call to execute a tool. This PR closes that gap.

The v3 connections schema (#9958) explicitly named "per-user authorization, token brokering" as a later phase; this is that phase, for the gateway's remote-HTTP execution path specifically.

- `identityModel: "personal_only" | "company_or_personal"` on `ConnectionMethodDef`. Link-connected servers with no `AppDefinition` (e.g. a company's own Google Workspace MCP proxy) can set this directly on the connection's own config.
- `executeRemoteHttpTool` branches on it: for `personal_only`, refuse a public/shared agent outright (`agents.isPublic`) before any grant lookup -- a shared agent has no single person to act for, and there's no non-personal fallback identity to use instead -- then resolve the run's `responsibleUserId` and use *that person's* `connection_grants` row. Never the connection's own `credentialRefs`, never a workspace-kind grant.
- Missing grant: posts the existing "Connect your account" interaction card (via a post-construction hook, since `toolGatewayService` is constructed before `toolAccessService`) and returns a structured `user_authorization_required` error instead of a raw failure.
- Expired grant: attempts a refresh via the grant's stored `refresh_token` before falling back to the connect prompt, so a long-running agent isn't re-prompting its owner every time a short-lived access token expires. Fails closed (marks `needs_reauthorization`) if the refresh token itself is no longer valid.

## Test plan

- [x] `pnpm --filter @paperclipai/server typecheck` clean
- [x] `pnpm --filter @paperclipai/shared typecheck` clean
- [x] New tests in `server/src/__tests__/tool-gateway.test.ts`: public agent refused before grant lookup; missing grant refused + connect prompt posted; valid grant's token used instead of connection credentials; expired grant refreshed via hook; expired grant with no refresh hook still falls through to the connect prompt.
- [x] Full existing suite (`tool-gateway.test.ts`, `tool-gateway-service.test.ts`, `tool-access-service.test.ts`) passing, 195 tests.


purpleCowOnWheels:
Closing — this branch depends on agents.isPublic, which is RH-fork-specific (not yet upstreamed) and doesn't exist on this repo's master. Re-opening the equivalent PR against redesignhealth/paperclip instead.

PR #11179, opened 2026-08-10
fix(heartbeat): replay child completion after active parent run

arthur-drozdov:
## Thinking Path

> - Paperclip manages AI agents and their work.
> - The heartbeat service starts and continues issue work.
> - An issue wake can arrive while the same issue already has a running agent.
> - Paperclip can merge a wake into the stored context of that running run.
> - The adapter already received its prompt, so it cannot see the merged context.
> - A final child completion can therefore be lost after Paperclip records it as coalesced.
> - This pull request defers that wake and starts one follow-up run after the active run ends.
> - The benefit is that a parent agent sees the final state of its children without duplicate concurrent work.

**What happened?**

A parent issue had an active agent run when its final child changed to `done`. Paperclip marked the `issue_children_completed` wake as `coalesced` into the active run. The running adapter had already received its context. It did not see the child completion. Paperclip did not start a later run after the active run ended. The parent stayed open without a live execution path.

**Expected behavior**

Paperclip must preserve a child completion that arrives after the parent run starts. It must start exactly one follow-up run after the active run ends. The follow-up run must read the latest child state.

**Steps to reproduce**

1. Create a parent issue and assign it to an agent.
2. Start a run for the parent issue.
3. Complete the final direct child while the parent run is still running.
4. Let the parent run end without reading the new child state.
5. Observe that the old behavior records a coalesced wake and creates no follow-up run.

`6a4e2e1b8c7129f6f913ae458ab0be9cba50bd6a`

**Deployment mode**

Self-hosted server with external PostgreSQL. The bug is in the core heartbeat service and is not adapter-specific.

## What Changed

- Added `issue_children_completed` to the set of running-issue events that require a follow-up run.
- Reused the existing durable `deferred_issue_execution` queue and promotion path.
- Extended the dependency scheduling integration test to cover blocker resolution and child completion.

## Verification

- Ran `pnpm exec vitest run server/src/__tests__/heartbeat-dependency-scheduling.test.ts`.
- Result: 7 tests passed.
- Ran `pnpm --filter @paperclipai/server typecheck`.
- Result: passed.
- The full GitHub build, server, workspace, serialized server, end-to-end, policy, and security checks passed on this pull request.

## Risks

Low risk. The change affects one wake reason. It can create one additional sequential run when the final child completes during an active parent run. Existing coalescing still applies when the wake targets a queued run. The database schema does not change.

> I checked `ROADMAP.md`. This bug fix does not duplicate planned core work.

## Model Used

OpenAI Codex with GPT-5.6-sol. The run used max reasoning, tool use, shell execution, and code execution.

## Checklist

### DIFF
--- a/server/src/__tests__/heartbeat-dependency-scheduling.test.ts
+++ b/server/src/__tests__/heartbeat-dependency-scheduling.test.ts
@@ -419,7 +419,10 @@ describeEmbeddedPostgres("heartbeat dependency-aware queued run selection", () =
     expect(noActiveRuns).toBe(true);
   });
 
-  it("defers issue_blockers_resolved as a follow-up when the same issue is already running", async () => {
+  it.each([
+    "issue_blockers_resolved",
+    "issue_children_completed",
+  ])("defers %s as a follow-up when the same issue is already running", async (wakeReason) => {
     const companyId = randomUUID();
     const agentId = randomUUID();
     const blockerId = randomUUID();
@@ -490,11 +493,11 @@ describeEmbeddedPostgres("heartbeat dependency-aware queued run selection", () =
       processGroupId: null,
     });
 
-    const idempotencyKey = `issue_blockers_resolved:${blockedIssueId}:${blockerId}`;
+    const idempotencyKey = `${wakeReason}:${blockedIssueId}:${blockerId}`;
     const wake = await heartbeat.wakeup(agentId, {
       source: "automation",
       triggerDetail: "system",
-      reason: "issue_blockers_resolved",
+      reason: wakeReason,
       payload: {
         issueId: blockedIssueId,
         resolvedBlockerIssueId: blockerId,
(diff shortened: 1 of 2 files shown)


maximilliangrand:
I independently reproduced this while investigating the newer report #12601, then found your PR during a broader duplicate check. Your existing change covers that report too. Thanks for identifying the fix.

I added a [follow-up regression test and documentation](https://github.com/maximilliangrand/paperclip/commit/7256694) that you can use here. The branch preserves your original fix commit and authorship; the linked commit contains only the added proof and documentation.

The test uses isolated embedded PostgreSQL and the real heartbeat service. It blocks the first adapter after context delivery, submits child completion twice, and verifies:

- The current adapter and stored active-run context do not acquire the late child summaries.
- Both wakes share one durable deferred request.
- Releasing the first adapter starts a different run that receives both child summaries, including in `paperclipWake`.
- The deferred request reaches `completed` and names that new run.

The adapter is deterministic; there are no live model calls. The regression fails on current master without the guard. With your change applied to `be6bb768b1370c0c0d697a9fc60f03db30c85564`, the dependency-scheduling suite passes **9 tests**:

```sh
pnpm exec vitest run server/src/__tests__/heartbeat-dependency-scheduling.test.ts
```

This is supplemental test evidence, not a claim that the full PR gates have been rerun. I am contributing it here rather than opening a duplicate fix PR. Codex assisted with the added test.

PR #11184, opened 2026-08-10
fix(server): raise company import zip upload limit to 1 GB and make it operator-configurable

devinfoley:
## Thinking Path

> - Company import/export lets an operator move a full company package between instances, with the Import page uploading the package as one compressed `.zip`
> - The server caps that upload at 128 MB, and real company packages with attachments now exceed it — imports fail at the preview step
> - The failure message tells the user to use the CLI folder import, but that path posts inline JSON capped at 64 MB, so the advice is a dead end for exactly these packages
> - This pull request raises the zip upload cap to a 1 GB default, makes it operator-configurable through an environment variable, scales the decompression-bomb guards from the cap in effect, and replaces the misleading hint
> - The benefit is that large real-world company packages import successfully, and operators with unusual needs can tune the cap without a code change

**What happened?**

A company import fails at the preview step with `Preview failed: Import package exceeds 134217728 bytes`. The package is a valid Paperclip export. Its compressed size is larger than the 128 MB server cap (one reported package is 257 MB). The error panel suggests the CLI folder import, but that path sends the package as one inline JSON body capped at 64 MB, so it also fails.

**Expected behavior**

A valid company package of realistic size imports successfully through the Import page. If a package is too large, the error must state the limit clearly and suggest a step that can work.

**Steps to reproduce**

1. Export a company with enough attachments to make the compressed package larger than 128 MB.
2. Open the Import page and upload the `.zip`.
3. Click "Preview import".
4. The preview fails with `Import package exceeds 134217728 bytes`.

**Deployment mode**

Reported from a managed deployment; the limit applies to all deployment modes.

## What Changed

- Raise `PORTABLE_ZIP_UPLOAD_LIMIT_BYTES` from 128 MB to a 1 GB default (`server/src/http/body-limits.ts`).
- Add the `PAPERCLIP_IMPORT_ZIP_MAX_BYTES` environment override. Invalid or non-positive values fall back to the default.
- Scale the zip decompression-bomb guard from the configured cap at the import route: the aggregate inflated ceiling is 4x the cap. The per-entry ceiling stays at 512 MB because V8's string length limit applies to an entry regardless (`server/src/routes/companies.ts`, `packages/shared/src/portability-zip.ts`).
- Report the 422 limit error in MB instead of raw bytes.
- Replace the "use the CLI folder import for very large packages" hint on preview failure with advice that works: re-export the package without large attachments (`ui/src/pages/CompanyImport.tsx`).
- Update the stale comment in `ui/src/lib/import-preflight.ts` that made the same CLI claim.
- Add tests for the new default, the env override, and the invalid-override fallback.

## Verification

- `pnpm vitest run server/src/__tests__/body-limits.test.ts packages/shared/src/portability-zip.test.ts server/src/__tests__/company-portability-routes.test.ts server/src/__tests__/company-portability.test.ts server/src/__tests__/company-portability-import-batching.test.ts` — all pass.
- `pnpm vitest run ui/src/pages/CompanyImport.test.tsx` — passes, including the updated failure-panel copy assertion.
- `pnpm typecheck` — clean across the workspace.
- Manual: upload a `.zip` larger than the configured cap; the preview fails with `Import package exceeds the 1024 MB upload limit` and the new hint. A package between 128 MB and 1 GB now previews and imports.

## Risks

- Peak per-import memory rises with the cap: the upload is buffered in memory and unzipped in one pass. A 1 GB compressed package can use several GB transiently. Imports are instance-admin actions, so the exposure is a deliberate operator action, not anonymous traffic. Operators on small hosts can lower the cap with `PAPERCLIP_IMPORT_ZIP_MAX_BYTES`.
- The aggregate bomb guard moves from a fixed 512 MB to 4x the configured cap. It still bounds expansion far below what a decompression bomb needs.
- No migration and no API shape change. The 422 message text changes; no code matches on the old text.

## Model Used

- Claude Fable 5 (`claude-fable-5`) via Claude Code CLI, with extended thinking and tool use (file edits, local test runs, live-instance inspection).

## Checklist

### DIFF
--- a/packages/shared/src/portability-zip.test.ts
+++ b/packages/shared/src/portability-zip.test.ts
@@ -270,7 +270,7 @@ describe("readZipArchive", () => {
 
   it("bounds a highly compressible DEFLATE entry at the per-entry decompressed limit", async () => {
     // Compresses tiny but expands to 8 KiB; a 1 KiB cap must reject it before it
-    // materializes. Real packages sit far under the 256 MB production default.
+    // materializes. Real packages sit far under the 512 MB production default.
     const bomb = new TextEncoder().encode("a".repeat(8 * 1024));
     const archive = buildZip([{ path: "bomb.txt", bytes: bomb, method: 8 }], "paperclip-demo");
     await expect(
--- a/packages/shared/src/portability-zip.ts
+++ b/packages/shared/src/portability-zip.ts
@@ -20,10 +20,13 @@ const strictTextDecoder = new TextDecoder("utf-8", { fatal: true, ignoreBOM: tru
 // malicious highly-compressible package cannot exhaust server memory: a
 // per-entry ceiling (passed to zlib as maxOutputLength, so it fails before
 // over-allocating) plus an aggregate ceiling across all entries. Both sit far
-// above any real company package (inline JSON was historically capped at 64MB)
-// yet far below what a bomb would need.
-export const MAX_ZIP_ENTRY_DECOMPRESSED_BYTES = 256 * 1024 * 1024;
-export const MAX_ZIP_TOTAL_DECOMPRESSED_BYTES = 512 * 1024 * 1024;
+// above any real company package yet far below what a bomb would need. These
+// are only fallback defaults: the server import route passes explicit limits
+// scaled from its configured upload cap. The per-entry ceiling stays at 512 MB
+// even for large caps because a text or base64 entry beyond that cannot
+// materialize as a JS string (V8's string length limit) regardless.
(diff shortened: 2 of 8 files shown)


devinfoley:
@greptileai Addressed the override-bounds finding: the env override is now floored first and clamped to [1 byte, 64 GiB], so a fractional value falls back to the 1 GB default instead of flooring to zero, and the derived 4x decompression guard stays within safe integer range. Tests cover the fractional, zero, negative, and oversized cases.

PR #11212, opened 2026-08-11
fix: stream DB backups and guard disk space

tommypoltev:
## Linked Issues or Issue Description
**Summary**
The first streaming backup revision fixed local disk exhaustion, but it introduced two regressions:
- callers without an explicit `maxBackups` value were forced into a hard count cap instead of tiered retention
- a fast `pg_dump` failure could leave the `.partial` path in a broken state and prevent the JavaScript fallback from running

**Steps to reproduce**
- Run `pnpm --filter @paperclipai/db exec vitest run src/backup-lib.test.ts` on the original PR head.
- Observe the tiered-retention failure and the fallback-path failure.
- Observe that the preflight tests can disappear on hosts where embedded Postgres is unavailable.

**Expected behavior**
- Tiered retention must remain the default unless a caller explicitly requests `maxBackups`.
- `engine: "auto"` must fall back from `pg_dump` to the JavaScript backup path without leaving unusable `.partial` files behind.
- The new preflight tests must run without depending on embedded Postgres availability.

**Actual behavior**
- The original PR defaulted `maxBackups` to `7` inside `runDatabaseBackup()`, which bypassed tiered retention for all implicit callers.
- The original PR could hit `EEXIST` on the `.partial` path after an early `pg_dump` failure.
- The preflight tests were hidden behind an embedded-Postgres gate on some hosts.

## What Changed
- Restored explicit-only `maxBackups` handling in `runDatabaseBackup()` so tiered retention remains the default contract for server and worktree callers.
- Kept the seven-backup cap only on the local/manual entrypoints that intentionally want count-based retention.
- Opened the partial output path before the streaming pipeline and waited for both the child-process and pipeline outcomes before cleanup, so early `pg_dump` failures no longer poison the fallback path.
- Added unique backup file bases so concurrent runs do not share a single `.partial` path.
- Moved the free-space preflight coverage outside the embedded-Postgres gate and routed it through an injectable disk-space seam.
- Kept PR `#11212` as the only active review thread after closing duplicate PR `#11213`.

## Verification
- `pnpm --filter @paperclipai/db exec vitest run src/backup-lib.test.ts`
- Result on this host at PR head `90bc31974be04a2de5a97bd3afe71a20924ca7d8`: `Test Files 1 passed (1)`, `Tests 4 passed | 9 skipped (13)`, `0 failed`
- Reviewer re-ran the same target suite on a host with embedded Postgres available and reported `13 passed (13), 0 failed, 0 skipped` on the same PR head.

## Risks
[...]

PR #11213, opened 2026-08-11
fix(db): stream compressed backups with free-space preflight

tommypoltev:
## Summary
- stream pg_dump and JS backup output directly into `.sql.gz.partial` and rename on success
- fail fast on low disk using 1.5x of the latest completed backup size
- support pruning to the latest N compressed backups and use N=7 for local db:backup entrypoints
- add tests for preflight, N-backup pruning, and no leftover raw `.sql` / `.partial` files

## Validation
- `pnpm --filter @paperclipai/db exec vitest run src/backup-lib.test.ts` (pass; embedded-postgres cases skipped on this host)
- `pnpm --filter @paperclipai/db exec tsc --noEmit` (fails on pre-existing `cloud_upstreams.ts` type errors unrelated to OFM-4274)
- attempted live local smoke against the current instance DB; runtime started, but the environment blocked completing a clean CLI smoke within this session (`tsx` IPC issues in worktree, sandbox/local-port restrictions without escalation)

### DIFF
--- a/cli/src/commands/db-backup.ts
+++ b/cli/src/commands/db-backup.ts
@@ -73,6 +73,7 @@ export async function dbBackupCommand(opts: DbBackupOptions): Promise<void> {
     const result = await runDatabaseBackup({
       connectionString: connection.value,
       backupDir,
+      maxBackups: 7,
       retention: { dailyDays: retentionDays, weeklyWeeks: 4, monthlyMonths: 1 },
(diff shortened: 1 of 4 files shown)


tommypoltev:
Closing duplicate in favor of #11212, which now contains the latest fixes and verification for OFM-4274 / OFM-4279.

PR #11232, opened 2026-08-11
fix(issues): let a self-checkout give its own run a source issue

iPermit-CW:
## Thinking Path

> - Agents report their work by writing to issues, and a heartbeat run gets its source issue from the wake that named one
> - `f91a6e27` and `e8ae5286` added cross-issue influence containment, which refuses a write when the run has no source issue
> - A run started without issue context therefore cannot write to any issue at all, including one assigned to its own agent
> - Checkout already builds the correct issue context, but only on the branch that wakes a different agent
> - `shouldWakeAssigneeOnCheckout` returns false for an agent that checks out its own issue on its own run, so that run stays unscoped
> - This pull request records the checked-out issue on the caller's own run in exactly that case
> - The benefit is that an agent which claims its own work can then report on it, instead of working silently

**What happened?**

An agent run that starts without issue context is refused on every issue write for the whole life of the run. The API returns `403 cross_issue_influence_run_context_required`. The refusal also applies to issues that are assigned to the agent making the request.

`observeCrossIssueInfluence` in `server/src/services/cross-issue-influence-limit.ts` reads the source issue from the run, then throws when it is null:

```ts
```

The throw happens before the function compares the source issue against the target issue. A same-issue write is therefore refused for the same reason as a cross-issue write.

Self checkout does not repair the run. `POST /issues/:id/checkout` writes `issues.checkoutRunId`. It does not write `heartbeatRuns.contextSnapshot`. The route builds `contextSnapshot: { issueId }` only inside the `shouldWakeAssigneeOnCheckout` branch, and that predicate returns false for an agent that checks out its own issue on its own run.

Neither wake route accepts an issue id. `POST /agents/:id/heartbeat/invoke` and `POST /agents/:id/wakeup` build the context from `triggeredBy`, `actorId` and `forceFreshSession` only. `wakeAgentSchema` has no issue field. An external caller cannot create a scoped run.

The effect is silent. The agent runs, does the work, and cannot record it. The board does not move.

**Expected behavior**

An agent that checks out an issue as itself can then comment on that issue and update it. Cross-issue containment continues to apply to writes against other issues.

**Steps to reproduce**

1. Use a Paperclip instance where `now` is after `CROSS_ISSUE_INFLUENCE_ENFORCE_AT` (`2026-08-11T00:00:00.000Z`). Before that date the check is log only.
2. Assign an issue to an agent.
3. Start a run for that agent with `POST /api/agents/:id/heartbeat/invoke` and an empty body. This is the documented on-demand path.
4. Read the run. `contextSnapshot` has no `issueId`.
5. As the agent, call `POST /api/issues/:id/checkout` for the assigned issue.
6. As the agent, call `POST /api/issues/:id/comments` for that same issue.
7. The comment is refused with `403 cross_issue_influence_run_context_required`.

Observed on `master` at `0044fa89`.

`0044fa89` on `master`.

**Deployment mode**

`local_trusted`, loopback bind. The agent authenticates with an agent API key and a run id, so `req.actor.type` is `agent` and the check applies.

## What Changed

- Add `server/src/services/run-source-issue.ts` with `shouldAdoptRunSourceIssue` and `adoptRunSourceIssue`.
- `shouldAdoptRunSourceIssue` is the exact complement of `shouldWakeAssigneeOnCheckout`. A checkout either wakes another agent, or it adopts the issue onto the caller's own run.
- `adoptRunSourceIssue` records `issueId` on the run's `contextSnapshot`. It merges into the existing object and does not replace it.
- The update never overwrites an existing source issue. The `is null` test is part of the `WHERE` clause, so a concurrent request cannot pass the read and then overwrite the value.
- The statement also matches on `agentId` and `companyId`. A run only adopts an issue for the agent and the company that already own it.
- Wire the new branch into `POST /issues/:id/checkout` as the `else` of the existing wake branch.
- Add `server/src/__tests__/run-source-issue.test.ts`.

## Verification

Run the new tests and the tests for the sibling predicate:

```bash
npx vitest run \
  server/src/__tests__/run-source-issue.test.ts \
  server/src/__tests__/issues-checkout-wakeup.test.ts \
  server/src/__tests__/heartbeat-auto-checkout.test.ts
```

Result on this branch: 3 files passed, 11 tests passed.

Typecheck the server package:

```bash
```

Result on this branch: no errors.

The test file includes a property test. It asserts that `shouldAdoptRunSourceIssue` and `shouldWakeAssigneeOnCheckout` disagree on every input. If both returned true, a checkout would scope the run twice. If both returned false, a self checkout would again leave the run unscoped.

To confirm the fix by hand, repeat the reproduction steps above. At step 7 the comment is accepted.

## Risks

Low risk. The change adds behaviour to one branch that previously did nothing.

- The new branch runs only for an agent that checks out its own issue on its own run. Board callers and cross-agent checkouts keep their current path.
- Containment is not weakened. A run keeps the first source issue it is given. An agent cannot rewrite its source issue by checking out a second issue, so it cannot raise its own cross-issue allowance.
- A failure to adopt is logged and does not fail the checkout. The checkout result is unchanged.

## Model Used

Claude Opus 5, model id `claude-opus-5[1m]`, 1M context window. Extended reasoning, with tool use for repository search, file editing, and running the test suite locally.

## Checklist

### Notes for the reviewer

Two items are deliberately out of scope, to keep this change small.

1. The denial copy for this case is misleading. `issueWriteDenialResponse("cross_issue_influence_run_context_required")` tells the reader to resend `X-Paperclip-Run-Id` with `$PAPERCLIP_RUN_ID`. That advice fits a missing header. It does not fit a run that has a valid header and no source issue. An agent that reads it retries the same request and fails again. I am happy to send a second PR for the copy.

2. A run that never checks out an issue is still unscoped. This change covers the checkout path, which is where a run declares what it works on. A wider fix would let the wake routes accept an issue id.

### DIFF
--- a/server/src/__tests__/run-source-issue-postgres.test.ts
+++ b/server/src/__tests__/run-source-issue-postgres.test.ts
@@ -1,0 +1,175 @@
+import { randomUUID } from "node:crypto";
+import { eq } from "drizzle-orm";
+import { afterAll, afterEach, beforeAll, describe, expect, it } from "vitest";
+import { agents, companies, createDb, heartbeatRuns } from "@paperclipai/db";
+import {
+  getEmbeddedPostgresTestSupport,
+  startEmbeddedPostgresTestDatabase,
+} from "./helpers/embedded-postgres.js";
+import { adoptRunSourceIssue } from "../services/run-source-issue.js";
+import { observeCrossIssueInfluence } from "../services/cross-issue-influence-limit.js";
+
+const embeddedPostgresSupport = await getEmbeddedPostgresTestSupport();
+const describeEmbeddedPostgres = embeddedPostgresSupport.supported ? describe : describe.skip;
+
+describeEmbeddedPostgres("adoptRunSourceIssue against PostgreSQL", () => {
+  let db!: ReturnType<typeof createDb>;
+  let tempDb: Awaited<ReturnType<typeof startEmbeddedPostgresTestDatabase>> | null = null;
+
+  beforeAll(async () => {
+    tempDb = await startEmbeddedPostgresTestDatabase("paperclip-run-source-issue-");
+    db = createDb(tempDb.connectionString);
+  }, 20_000);
+
+  afterEach(async () => {
+    await db.delete(heartbeatRuns);
+    await db.delete(agents);
+    await db.delete(companies);
+  });
+
+  afterAll(async () => {
+    await tempDb?.cleanup();
+  });
+
+  async function seedRun(contextSnapshot: Record<string, unknown> | null) {
+    const companyId = randomUUID();
+    const agentId = randomUUID();
+    const runId = randomUUID();
+
+    await db.insert(companies).values({
+      id: companyId,
+      name: "Paperclip",
+      issuePrefix: `C${companyId.replace(/-/g, "").slice(0, 6).toUpperCase()}`,
+      defaultResponsibleUserId: "board-user",
+    });
+    await db.insert(agents).values({
+      id: agentId,
+      companyId,
+      name: "Checkout Coder",
+      role: "engineer",
+      adapterType: "codex_local",
+      adapterConfig: {},
+      runtimeConfig: {},
+      permissions: {},
+    });
+    await db.insert(heartbeatRuns).values({
+      id: runId,
+      companyId,
+      agentId,
+      status: "running",
+      responsibleUserId: "board-user",
+      contextSnapshot,
+    });
+
+    return { companyId, agentId, runId };
+  }
+
+  async function readContext(runId: string) {
+    const row = await db
+      .select({ contextSnapshot: heartbeatRuns.contextSnapshot })
+      .from(heartbeatRuns)
+      .where(eq(heartbeatRuns.id, runId))
+      .then((rows) => rows[0] ?? null);
+    return row?.contextSnapshot ?? null;
+  }
+
+  it("gives an unscoped run its checked-out issue", async () => {
+    const { companyId, agentId, runId } = await seedRun(null);
+    const issueId = randomUUID();
+
+    await expect(adoptRunSourceIssue(db, { runId, agentId, companyId, issueId })).resolves.toBe(true);
+    expect(await readContext(runId)).toMatchObject({ issueId, source: "issue.checkout.self" });
+  });
+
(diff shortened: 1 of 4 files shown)


iPermit-CW:
Thanks — the coverage gap was real. `b4b74c8` adds `server/src/__tests__/run-source-issue-postgres.test.ts`, which exercises the update itself rather than only the branch choice:

- an unscoped run adopts the issue
- the merge keeps the rest of the context snapshot
- a run that already has a source issue keeps it
- a `taskId` counts as an existing source issue, because `readRunSourceIssueId` accepts either spelling
- an agent that does not own the run cannot adopt
- adoption does not cross a company boundary
- two concurrent adoptions produce exactly one winner, which is what the `is null` test in the `WHERE` clause is for
- `observeCrossIssueInfluence` refuses a write to the issue before adoption and allows it after

The last one states the bug in a single assertion.

Local result: 8 tests passed against embedded PostgreSQL.

On the earlier red: `General tests (workspaces-a (2/2))` failed on `@paperclipai/ui src/App.cases-routing.test.tsx`. This branch changes three files, all under `server/src`, and none under `ui/`. That suite also shows a pass and a fail on the same master SHA `b5ebda1d`, so it looks flaky upstream rather than caused here. `verify` went red only as a cascade from it. Happy to rebase if you would like a clean run.

kallehiitola (approved):
Security review for the timer/self-checkout context-binding path: approved at head b4b74c8. The adoption is constrained in one atomic heartbeat_runs UPDATE by the exact run ID, agent ID, and company ID; it only fills an unscoped context and never overwrites issueId/taskId. The route reaches it only after the same agent checks out the issue using that run, so arbitrary first-write binding remains fail-closed. This adds no issue read or reverse issue/run row-lock order. The PostgreSQL regressions cover wrong-agent and cross-company rejection, source preservation, concurrent adoption, and write refusal before adoption. This is the safe replacement path cited when #12315 was closed; dzianisv/paperclip#6 is not required on this implementation path.

kallehiitola:
Maintainer merge requested: this is now the active upstream fix after #12315 was closed as a duplicate. The current head `b4b74c8` has an approving review and ownership/company/run-scoped PostgreSQL coverage; please merge once repository policy permits.

PR #11308, opened 2026-08-13
feat(adapters): let model discovery see company secrets

adamteale:
## Thinking Path

> - Paperclip stores provider API keys as per-company secrets and injects them when an agent runs
> - Adapter model discovery starts the provider CLI, which lists only the providers it can authenticate
> - `listModels` accepts no arguments, so discovery sees only the server process environment
> - A key held as a company secret therefore never shows a model in the picker
> - This pull request adds an optional, already-resolved environment to model discovery
> - The benefit is that secret-backed providers show their models without hardcoded on-disk configuration

Refs #10919 — my earlier PR on the same root cause. That PR makes a failed model check non-fatal. This PR makes the discovery itself succeed. The two are complementary: keep the fallback for residual failures, and remove the most common cause of them.

Refs #11169 — also touches adapter model discovery, for OpenCode. No overlap in the files changed.

No public issue covers this. The description follows `.github/ISSUE_TEMPLATE/enhancement.yml`.

Adapter model discovery for `pi_local`, and for any adapter that starts a provider CLI to list models.

**Subsystem affected**

The adapter contract in `packages/adapter-utils`, the server adapter registry, the agent models route, and the UI API client.

**Current behavior**

`ServerAdapterModule.listModels` has the type `() => Promise<AdapterModel[]>`. The registry calls it with no arguments. In `pi-local`, `listPiModels()` calls `discoverPiModelsCached()` with no arguments. `discoverPiModels` then builds the child environment from `{ ...process.env }` only.

A provider whose API key is a per-company secret is therefore invisible to the model picker. These results are measured against pi 0.84.1 with `auth.json` set to `{}`:

| Provider entry in pi `models.json` | Credential source | Result of `pi --list-models` |
|---|---|---|
| with `apiKey: "<name>"` | `auth.json["<name>"]` | Always lists the hardcoded models |
| without `apiKey` | environment variable | Lists only if a credential is visible to the process |
| absent | none | Never lists |

The operator must therefore hardcode the provider in `models.json` with an `apiKey` value. That value is a lookup key into `auth.json`. It also overrides the built-in provider. The resolved company secret is then unused at execution time, and every request returns `401`. One configuration choice breaks two surfaces in opposite directions.

**Proposed behavior**

Model discovery accepts an optional context that carries an already-resolved environment. The server resolves company secrets with `resolveAdapterConfigForRuntime`, which is the same function that execution uses. The adapter receives a plain environment map.

**Reason and benefit**

Operators can keep credentials in company secrets only. They do not need to hardcode providers or models on disk.

**Breaking changes**

None. Every new parameter is optional.

## What Changed

- Add `AdapterModelDiscoveryContext` to `packages/adapter-utils`. It carries an optional `env` map.
- Change `listModels` and `refreshModels` to accept that optional context.
- Pass the context through `listAdapterModels` and `refreshAdapterModels` in the server adapter registry.
- Accept an optional `agentId` query parameter on `GET /companies/:companyId/adapters/:type/models`. When present, resolve that agent's adapter environment and pass it to discovery.
- Forward the context environment to `discoverPiModelsCached` in the `pi-local` adapter.
- Add an optional `agentId` option to `agentsApi.adapterModels` in the UI client.
- Add a test for the new behaviour.

## Verification

Run the adapter tests:

```
cd packages/adapters/pi-local && npx vitest run src/server/
```

- 37 tests pass, including the 2 new tests in `src/server/models.context.test.ts`.
- The first new test fails without the change. I confirmed this by reverting `models.ts` and running the test again.

Type checks:

```
cd packages/adapter-utils && npx tsc --noEmit -p tsconfig.json
cd packages/adapters/pi-local && npx tsc --noEmit -p tsconfig.json
```

Both exit 0. The server type check reports no new errors in the changed files.

Manual check: request `GET /companies/<id>/adapters/pi_local/models?agentId=<id>` for an agent whose provider key is a company secret. The response now contains that provider's models. Without `agentId` the response is unchanged.

## Risks

Low risk.

- Every new parameter is optional. Existing callers and adapters are unchanged.
- The route resolves secrets inside a `try`/`catch`. A resolution failure falls back to the previous behaviour, so the picker cannot break.
- The route resolves an agent only after it confirms the agent belongs to the requested company.
- Discovery results are cached per environment, because the cache key already includes the environment. A different agent therefore does not read another agent's cached list.
- Secret values reach only the child process for discovery. They are not logged and not returned in the response.

## Model Used

Claude Opus 5 (`claude-opus-5`), 1M context window, extended thinking, with tool use and code execution through Claude Code.

## Checklist


adamteale:
Thanks — both findings were valid and are fixed in e0b281d.

**Issue 2 — credentials cross adapter boundaries.** Agreed, and this was the more serious of the two. The route now requires the agent's adapter type to match the requested type, not just the company:

```ts
if (
  discoveryAgent
  && discoveryAgent.companyId === companyId
  && discoveryAgent.adapterType === type
) {
```

**Issue 1 — agent context never reaches discovery.** Correct: I added the option but wired no caller, so the path was unreachable. `AgentConfigForm` now passes the edited agent id on both the initial fetch and the manual refresh. `queryKeys.agents.adapterModels` takes the agent id too, because two agents of the same adapter type can resolve different secrets and so can see different model lists — without it they would share a cache entry.

Other callers (`OnboardingWizard`, `ConfigureBuiltInAgentModal`, `NewIssueDialog`) still omit `agentId` on purpose: they run before an agent exists or are not agent-scoped, and they keep the previous process-environment behaviour.

Verification: `pi-local` 37 tests pass; `AgentConfigForm.render` and `ConfigureBuiltInAgentModal` UI tests pass (32); `tsc --noEmit` clean for the UI and for the changed server file.

PR #3030, opened 2026-04-07
stop leaking server worktree env into unrelated local adapter heartbeats

kfiramar:
## Summary

This change stops server-scoped worktree metadata from leaking into unrelated local-adapter heartbeats.

- sanitize inherited `PAPERCLIP_*` worktree env before spawning child runs unless the resolved run is actually worktree-backed
- update `codex-local` home selection to use resolved run/workspace context instead of raw parent `process.env`
- keep real worktree-backed runs on isolated worktree homes while stripping hint-only inherited metadata from `agent_home` runs

## Thinking Path

Supportopia hit a backend bug where an `agent_home` heartbeat inherited server-side worktree hints and then behaved like it was executing inside a real git worktree. That polluted downstream local-adapter child environments, which in turn pointed Codex at the wrong `PAPERCLIP_HOME` / `CODEX_HOME` and mixed unrelated runtime state into the child run. The fix therefore needs to happen at the process-boundary layer, not in per-agent prompts: inherited Paperclip env has to be sanitized before child processes spawn, while still preserving the worktree-scoped values for runs that actually resolve to a git worktree.

## Before / After

Before:
- `agent_home` heartbeats could inherit server worktree vars such as `PAPERCLIP_IN_WORKTREE`, `PAPERCLIP_WORKTREE_NAME`, `PAPERCLIP_HOME`, `PAPERCLIP_CONFIG`, `PAPERCLIP_INSTANCE_ID`, and `PAPERCLIP_CONTEXT`
- `codex-local` could incorrectly treat `agent_home` runs as worktree-isolated because it keyed off raw server env

After:
- `agent_home` child runs no longer inherit unrelated server worktree metadata unless that run explicitly resolved into a worktree
- true worktree-backed runs still preserve the worktree-specific env and isolated `CODEX_HOME`

## Tests

- `pnpm vitest run server/src/__tests__/paperclip-env.test.ts server/src/__tests__/codex-local-execute.test.ts`
- `pnpm vitest run server/src/__tests__/workspace-runtime.test.ts -t "does not leak parent Paperclip instance env into runtime service commands"`
- `pnpm --filter @paperclipai/server typecheck`

## Risks

- The sanitization deliberately strips inherited worktree hint keys for non-worktree child runs, so any downstream logic that was implicitly depending on leaked parent env will now need to rely on explicit resolved workspace context instead.
[...]

PR #11329, opened 2026-08-13
fix(adapters): stop agents inheriting NODE_ENV=production

jetooh:
## Thinking Path

> - The adapter layer builds the environment for every agent child process, in `buildPaperclipEnv` and the workspace/runtime env builders
> - The server image sets `NODE_ENV=production` in the Dockerfile. This is correct for the server process. Agent child processes inherit it
> - An agent workspace is a development environment. With `NODE_ENV=production`, npm skips devDependencies and bundlers resolve the production React build, so frontend work fails
> - The failure looks like a broken repository, not a broken environment. An agent reports "the test suite is red" when the code is correct
> - This pull request gives agent-facing processes a non-production `NODE_ENV` default
> - The benefit is that agents can run install, typecheck and test in a frontend repository without extra flags

No public GitHub issue exists for this. The problem is described below, with the fields from `.github/ISSUE_TEMPLATE/bug_report.yml`.

**What happened?**

A code agent runs standard frontend commands in its workspace and they fail. `npm ci` installs only the runtime dependencies. `npm run typecheck` fails with `sh: 1: tsc: not found`. `npm test` fails with `sh: 1: vitest: not found`.

If you install the devDependencies manually with `npm ci --include=dev`, the install and the typecheck pass, but every React component test fails with `act(...) is not supported in production builds of React`.

The cause is `NODE_ENV=production`. The server image sets it at `Dockerfile` (`ENV NODE_ENV=production`), and each agent child process inherits it. There are two distinct effects:

1. npm *derives* `omit=["dev"]` from `NODE_ENV=production`. This is a derived default, not separate npm config, so there is nothing to delete with `npm config delete omit`.
2. Bundlers resolve React to its production build, so `@testing-library/react` stops every component test.

Because the value is derived, an `.npmrc` with `omit=` repairs the install and the typecheck but leaves every component test failing. `NODE_ENV` is the only control that repairs both effects.

**Expected behavior**

An agent with a clean workspace can run `npm ci && npm run typecheck && npm test` in a frontend repository. The three commands exit 0. The agent does not need extra flags and does not need to override `NODE_ENV`.

**Steps to reproduce**

1. Run a code agent on a Paperclip server that uses the shipped Docker image.
2. In the agent workspace, clone a frontend repository that keeps `typescript` and `vitest` in `devDependencies` and has React component tests.
3. Run `npm ci`. The devDependencies are absent.
4. Run `npm run typecheck`. It fails with `tsc: not found`.
5. Run `npm ci --include=dev`, then `npm test`. The component tests fail with `act(...) is not supported in production builds of React`.
6. Run `NODE_ENV=test npx vitest run`. The same tests pass.

`master` at `88e1ccb424ed14995e8932c75ea804f3db1994b1`.

**Deployment mode**

Self-hosted server that uses the project Dockerfile.

**Agent adapter(s) involved**

Not adapter-specific (core bug). The value is inherited by all adapters. It was first seen with Claude Code on the ACPX lane.

## What Changed

- Add `AGENT_DEFAULT_NODE_ENV` and `agentNodeEnvDefault()` to `packages/adapter-utils/src/server-utils.ts`. The function replaces a `production` or empty value with `development` and keeps a deliberate non-production host value.
- Add `applyAgentNodeEnvDefault()` for the env builders that start from a `{ ...process.env }` clone.
- Set `NODE_ENV` in `buildPaperclipEnv()`. Every adapter layers its run env over this record, and both execution lanes merge it over the inherited `process.env`, so this single point reaches all of them.
- Apply the default to workspace setup commands in `buildWorkspaceCommandEnv()`. These commands run `npm ci` before the agent starts.
- Apply the default to runtime dev services in `sanitizeRuntimeServiceBaseEnv()`. These are development processes.
- Re-export both functions from `server/src/adapters/utils.ts`.
- Add 10 unit tests in `packages/adapter-utils/src/server-utils-env.test.ts`.

The parameter of `agentNodeEnvDefault()` is required on purpose. A default of `process.env.NODE_ENV` makes an explicit `undefined` read the host value, so a record that carries no `NODE_ENV` would receive the host value instead of the agent default. One test pins this behavior.

## Verification

Commands run on `master` at `88e1ccb`, with `pnpm install --frozen-lockfile`.

| Command | Result |
|---|---|
| `pnpm --filter @paperclipai/adapter-utils typecheck` | exit 0 |
| `pnpm --filter @paperclipai/server typecheck` | exit 0 |
| `npx vitest run packages/adapter-utils/src/server-utils-env.test.ts` | 10/10 pass |
| `npx vitest run packages/adapter-utils/src/server-utils.test.ts` | 99 pass, 2 fail |
| `npx vitest run server/src/__tests__/paperclip-env.test.ts` | 5/5 pass |
| `npx vitest run server/src/__tests__/workspace-runtime.test.ts` | 111 pass, 2 fail |
| `node ./scripts/check-no-git-push.mjs` | pass |
| `node ./scripts/check-docker-deps-stage.mjs` | pass |

The 4 failures are present before this change. I measured them on a pristine checkout of the same commit:

- `server-utils.test.ts`: the same 2 `runChildProcess` process-group tests fail (89 pass, 2 fail before; 99 pass, 2 fail after, because this pull request adds 10 tests).
- `workspace-runtime.test.ts`: the same 2 startup-reconciliation tests fail, with `EADDRINUSE` on a local port (106 pass, 2 fail before). They are sensitive to ports held by other processes on the machine.

The new tests run under `NODE_ENV=test`, `NODE_ENV=production` and with `NODE_ENV` unset. They pass in all three cases, so they do not depend on the environment of the test runner.

Acceptance check of the reported symptom: with the inherited `NODE_ENV=production` neutralized, `npm ci`, `npm run typecheck` and `npm test` all exit 0 in a React component library that keeps `typescript` and `vitest` in `devDependencies` (57/57 tests over 13 files), with no extra flags and no per-command `NODE_ENV` override.

Note for reviewers: the fix reaches running agents only after the image is rebuilt, because `packages/*/src` is compiled into `dist`.

## Risks

Low risk, with two behavior changes that are intentional.

- Agent child processes now get `NODE_ENV=development` instead of `production`. This is the purpose of the change. An operator who wants `production` in an agent can still set it, because adapter and project config env is applied after the Paperclip base env and `NODE_ENV` is outside the reserved `PAPERCLIP_` namespace.
- Runtime dev services and workspace setup commands also get the default. Installs in those paths now include devDependencies, so they use more disk.

The server process itself is not affected. `buildPaperclipEnv` builds the env for child processes only. A deliberate non-production host value, such as `development` on a self-hosted dev server, is preserved.

Related pull request found while searching for duplicates: #3030, "stop leaking server worktree env into unrelated local adapter heartbeats". It changes `packages/adapter-utils/src/server-utils.ts` and `server/src/adapters/utils.ts` too, and it is also about env leaking into agent processes. It covers a different variable set (worktree env), it does not address `NODE_ENV`, and it has had no activity since April 2026. I found no other open or closed pull request that addresses `NODE_ENV` inheritance by agents.

## Model Used

Claude Opus 5, model id `claude-opus-5[1m]`, 1M context window, with extended thinking and tool use. The change was produced and verified in a Paperclip agent session.

## Checklist

- [ ] I have updated relevant documentation to reflect my changes — no documentation describes the agent `NODE_ENV`; the reasoning is in code comments next to the change
- [ ] All Paperclip CI gates are green — to be confirmed after CI runs on this pull request


jetooh:
CI note: the red jobs here come from `master`, not from this branch.

Every failing job stops at `pnpm install --frozen-lockfile` with:

```
ERR_PNPM_LOCKFILE_CONFIG_MISMATCH  Cannot proceed with the frozen installation.
The current "overrides" configuration doesn't match the value found in the lockfile
```

Cause: commit eb6c6a8 (#10724) set `lexical: 0.49.0` in the lockfile `overrides` block and in `ui/package.json`, but the root `package.json` still pins `pnpm.overrides.lexical` to `0.46.0`.

- At `3cd5967`, the parent commit, the lockfile `overrides` said `lexical: 0.46.0` and matched the root manifest.
- At `eb6c6a8` and later, they differ.

This branch changes no `package.json`, no `pnpm-lock.yaml` and no patches, so the policy job does not regenerate a lockfile and every downstream job installs from the mismatched pair.

`pnpm install --frozen-lockfile` exits 0 on this branch at its base commit `88e1ccb`, which is before `eb6c6a8`.

#11328 ("chore(lockfile): refresh pnpm-lock.yaml") looks like the fix already in flight. I am happy to rebase once it lands, or to align the root override in this PR if you prefer — say which you want.

The `review`, `policy`, `security-review`, Socket and Snyk checks are green.

PR #4013, opened 2026-04-19
fix(server): share pluginLifecycleManager between dispatcher and routes

c-barlow:
## Summary

Fixes a bug where dynamically enabled plugins don't register their tools until a server restart.

`pluginLifecycleManager(db, ...)` constructs a fresh `EventEmitter` on every call, and prior to this change the server was calling it twice:

- `server/src/app.ts:203` — constructs the manager used by the plugin tool dispatcher (listener side).
- `server/src/routes/plugins.ts:313` — constructs a *different* manager used by the HTTP enable/disable routes (emitter side).

Because those two managers hold separate `EventEmitter` instances, `plugin.enabled` events emitted by `PATCH /api/plugins/:id/state` never reach the dispatcher's `plugin.enabled` listener. The dispatcher log `plugin enabled — registering tools { pluginId, ... }` never fires on dynamic enable, so plugin tools only become live on next server boot (which re-registers via the loader path). Disable → enable cycles on a plugin with tools are broken.

## Change

- Add `setLoader(loader)` to `PluginLifecycleManager` so a single shared instance can be back-filled with a runtime-capable `PluginLoader` after construction. The loader depends on the lifecycle manager for wiring, so the two are built in a two-step handshake and without this the dispatcher-side lifecycle would fall back to a no-op loader that doesn't support runtime activation.
- Add an optional `PluginRouteLifecycleDeps` arg to `pluginRoutes(...)`. When provided, the routes reuse the shared lifecycle. When omitted, they fall back to the existing construction path so existing callers and tests keep working.
- Wire `app.ts` to pass the shared `lifecycle` into `pluginRoutes` and to call `lifecycle.setLoader(loader)` once the loader is built.

The new dep bucket is optional to minimize the blast radius: any caller that isn't updated keeps the legacy behavior, but the in-repo call site now shares a single lifecycle manager across the dispatcher and the HTTP routes.

## Test plan

- [x] `pnpm --filter @paperclipai/server run typecheck`
- [x] `pnpm --filter @paperclipai/server exec vitest run src/__tests__/plugin-routes-authz.test.ts src/__tests__/plugin-worker-manager.test.ts src/__tests__/plugin-telemetry-bridge.test.ts src/__tests__/plugin-dev-watcher.test.ts` — 17/17 pass
- [ ] Manual: on a running server, install a plugin that contributes agent tools, then `POST /api/plugins/:id/disable` followed by `POST /api/plugins/:id/enable`, and confirm:
[...]

PR #11336, opened 2026-08-13
fix(server): share one pluginLifecycleManager between routes and dispatcher

c-barlow:
## Thinking Path

> - Agents call tools; plugin-provided tools reach agents through the plugin-tool dispatcher, which registers a plugin's tools when the plugin's lifecycle emits `plugin.enabled/disabled/unloaded`.
> - The plugin HTTP routes and the dispatcher each built their own `pluginLifecycleManager`, so each owned a private `EventEmitter`; a route-driven enable fired on the routes' emitter while the dispatcher listened on a different one.
> - The dispatcher therefore never saw runtime enable/disable events, and its tool table stayed stale until the next server restart re-ran `loadAll()`.
> - This pull request shares one lifecycle-manager instance between the routes and the dispatcher, and — once a startup window was found during review — gates the HTTP listener on the plugin runtime being fully wired.
> - The benefit is that enabling or disabling a plugin over the API updates the agent tool table immediately, with no restart, and no half-activated plugins during boot.

No public GitHub issue tracks this. Problem description:

Dynamically enabling a plugin over the HTTP API does not register its tools with the plugin-tool dispatcher until the server is restarted.

**Root cause:** the plugin HTTP routes (`routes/plugins.ts`) construct their **own** `pluginLifecycleManager` instance, separate from the one `app.ts` hands to the tool dispatcher at startup. Each manager owns a private `EventEmitter` (`emitDomain` → local `emitter.emit`; `on/off/once` → local `emitter`). So when a route handler calls `lifecycle.enable()/disable()/unload()`, the `plugin.enabled/disabled/unloaded` event fires on the **routes'** emitter, while the dispatcher subscribed to **app.ts's** emitter — the dispatcher never sees the event and its tool table goes stale until the next boot (when `loadAll()` re-registers everything).

Refs #4013 (superseded — its class-based patch no longer applies after `PluginLifecycleManager` was refactored from a class into a factory function). Related to #5671 (the earlier `pluginDbId` propagation fix).

## What Changed

- **`server/src/services/plugin-lifecycle.ts`** — add `setLoader(loader)` to the `PluginLifecycleManager` interface and factory; the internal loader binding becomes mutable so the real loader can be attached after construction (breaks the loader ⇄ manager circular dependency).
- **`server/src/routes/plugins.ts`** — accept an injected `lifecycleManager` (via the tool-gateway deps) and use it instead of building a private one; fall back to a private instance for callers (e.g. unit tests) that do not wire one.
- **`server/src/app.ts`** — pass the shared `lifecycle` to `pluginRoutes`, and call `lifecycle.setLoader(loader)` **only after** the startup `loadAll()` pass (chained in `.finally` of `bundledPluginsStartup`).
- **`server/src/index.ts`** — before `server.listen()`, `await` the settled `bundledPluginsStartup` promise so the listener does not accept requests until `setLoader` has run. This closes a startup window (found in review) where a plugin-mutating route could drive a loader-less lifecycle — persisting `ready` and emitting tool-registration events without ever spawning a worker.
- **`server/src/__tests__/plugin-lifecycle-set-loader.test.ts`** — regression tests: activation routes through the backfilled loader (never the construction-time one), and `plugin.enabled` fires on the shared emitter so a dispatcher-style co-located listener is invoked.

### Why `setLoader` is deferred until after `loadAll()`

`app.ts` deliberately constructs `lifecycle` without a runtime-capable loader so that `ensureBundledPlugins`' `lifecycle.load()` calls at boot only record `ready` status and do **not** spawn workers — workers are started exactly once by `loadAll()`. Backfilling the loader *before* that pass would double-start bundled workers. Backfilling in `.finally` (after the single activation pass) preserves the invariant while giving runtime route handlers a loader-capable manager for operator-initiated enable/disable.

### Why the listener now waits on `bundledPluginsStartup`

Because the loader is backfilled asynchronously, there was a window on deployments that do not otherwise gate boot on plugin readiness (self-hosted, or managed instances with no declared sandbox environments) where `server.listen()` could accept requests before `setLoader` ran. During that window a route-driven `enable/disable/unload` would persist `ready` and emit the dispatcher's tool-registration event, but skip worker activation — advertising tools with no runtime behind them. Awaiting the promise before `listen()` removes the window. Managed instances that declare sandbox environments already gated on this same promise via `applyManagedEnvironments`, which proves the ordering is safe (plugin workers do not require the HTTP server to be listening during activation). The promise never rejects, so a degraded plugin cannot block the listener.

- The `plugin-lifecycle-set-loader.test.ts` regression tests (loader backfill + shared-emitter delivery) were added with the original singleton change and run green.
- The `index.ts` listener-gating change is a single `await` of an already-exposed promise (the same cast pattern already used earlier in `index.ts` for `applyManagedEnvironments`), verified through this PR's **Build** and **General tests** CI jobs rather than a fresh local run.

A reviewer can confirm locally with:

```
pnpm --filter @paperclipai/server exec tsc --noEmit
pnpm --filter @paperclipai/server test plugin-lifecycle-set-loader
```

Manual: on a running server, enable a previously disabled plugin over the HTTP API and confirm its tools appear in the dispatcher (agent tool list) with no restart.

## Verification

**Automated — this PR's CI (run 31740905889) is green across the board:**
- Typecheck + Release Registry ✅
- Build ✅
- General tests: server (1–5), workspaces-a (1–2), workspaces-b ✅
- Verify serialized server suites (1–5) ✅
- e2e shards (1–3) + e2e ✅
- Canary Dry Run ✅ · policy ✅
- Greptile Review ✅ (Confidence 5/5 — "safe to merge from a runtime-correctness perspective; no blocking failure remains")

**Regression tests added** (`server/src/__tests__/plugin-lifecycle-set-loader.test.ts`):
- loader backfill — activation always routes through the backfilled loader, never the construction-time one
- shared emitter — `plugin.enabled` fires on the shared emitter, so a dispatcher-style co-located listener is invoked

**Local reproduction:**
```
pnpm --filter @paperclipai/server exec tsc --noEmit
pnpm --filter @paperclipai/server test plugin-lifecycle-set-loader
```

**Manual:** on a running server, enable a previously disabled plugin over the HTTP API and confirm its tools appear in the dispatcher (agent tool list) with **no restart**. The dependent Telegram plugin's operator-side round-trip (agent → Telegram → reply → Paperclip comment) was verified end-to-end separately.

## Risks

- **Low.** The `index.ts` change adds one `await` of an already-exposed promise before the listener starts, next to the existing external-adapter wait. It can only delay accepting requests by the time bundled-plugin startup takes; managed instances with sandbox environments already incurred this wait. The promise never rejects, so it cannot fail startup.
- No schema/migration changes. No public API changes. Behavior change is limited to plugin lifecycle event routing and startup ordering.

## Model Used

Claude Opus 4.8 (model id `claude-opus-4-8`, 1M-context variant), with extended thinking and tool use.

## Checklist

- [x] I have searched GitHub for duplicate or related PRs and linked them above (#4013 superseded, #5671 related)
- [x] My branch name describes the change (`fix/plugin-lifecycle-singleton-v2`) and contains no internal Paperclip ticket id or instance-derived details
- [x] I have run tests locally and they pass (setLoader regression suite; the index.ts listener-gating change is verified via this PR's Build + General tests CI)
- [ ] I have updated relevant documentation to reflect my changes (no user-facing docs affected)

### DIFF
--- a/server/src/__tests__/plugin-lifecycle-set-loader.test.ts
+++ b/server/src/__tests__/plugin-lifecycle-set-loader.test.ts
@@ -1,0 +1,121 @@
+/**
+ * Regression test for the plugin-lifecycle singleton fix (AMA-10).
+ *
+ * The dynamic plugin-tool dispatcher subscribes to lifecycle events on the
+ * lifecycle-manager instance it is handed at startup. Previously the HTTP
+ * plugin routes constructed a *separate* lifecycle-manager instance, so the
+ * events they emitted (plugin.enabled/disabled/unloaded) fired on a different
+ * EventEmitter and never reached the dispatcher — dynamically enabled plugins
+ * did not register their tools until a server restart.
+ *
+ * The fix shares one lifecycle-manager instance between the routes and the
+ * dispatcher. To do that the composition root must construct the manager
+ * before the loader exists (the loader depends on the manager) and backfill
+ * the loader afterwards via `setLoader`. These tests pin the `setLoader`
+ * contract that makes that wiring possible.
+ */
+import { describe, expect, it, vi } from "vitest";
+import type { PluginLoader } from "../services/plugin-loader.js";
+import type { PluginWorkerManager } from "../services/plugin-worker-manager.js";
+
+const disabledPlugin = {
+  id: "plugin-1",
+  pluginKey: "example.plugin",
+  status: "disabled",
+  manifestJson: { id: "example.plugin", capabilities: [] },
+  packageName: "@example/plugin",
+  version: "1.0.0",
+  packagePath: "/tmp/example-plugin",
+};
+
+const readyPlugin = { ...disabledPlugin, status: "ready" };
+
+const mockRegistry = vi.hoisted(() => ({
+  getById: vi.fn(),
+  getByKey: vi.fn(),
+  update: vi.fn(),
+  updateStatus: vi.fn(),
+  upsertConfig: vi.fn(),
+  getConfig: vi.fn(),
+  list: vi.fn(),
+  delete: vi.fn(),
+}));
+
+vi.mock("../services/plugin-registry.js", () => ({
+  pluginRegistryService: () => mockRegistry,
+}));
+
+import { pluginLifecycleManager } from "../services/plugin-lifecycle.js";
+
+function makeWorkerManagerStub() {
+  return {
+    getWorker: vi.fn().mockReturnValue(undefined),
+    isRunning: vi.fn().mockReturnValue(false),
+    startWorker: vi.fn().mockResolvedValue(undefined),
+    stopWorker: vi.fn().mockResolvedValue(undefined),
+    restartWorker: vi.fn().mockResolvedValue(undefined),
+  } as unknown as PluginWorkerManager;
+}
+
+function makeRuntimeLoaderStub(): Partial<PluginLoader> {
+  return {
+    hasRuntimeServices: vi.fn().mockReturnValue(true) as PluginLoader["hasRuntimeServices"],
+    loadSingle: vi.fn().mockResolvedValue({
+      success: true,
+      plugin: readyPlugin,
+      registered: { worker: true, eventSubscriptions: 0, jobs: 0, webhooks: 0, tools: 0 },
+    }) as PluginLoader["loadSingle"],
+    unloadSingle: vi.fn().mockResolvedValue(undefined) as PluginLoader["unloadSingle"],
+  };
+}
+
+describe("pluginLifecycleManager.setLoader", () => {
+  it("routes runtime activation through the backfilled loader, not the construction-time one", async () => {
+    mockRegistry.getById.mockResolvedValue(disabledPlugin);
+    mockRegistry.updateStatus.mockResolvedValue(readyPlugin);
+
+    // Mirrors the startup wiring: the manager is constructed with a
+    // non-runtime loader so bundled-plugin `load()` calls at boot only record
+    // status and never spawn a worker.
+    const bootLoader = makeRuntimeLoaderStub();
+    (bootLoader.hasRuntimeServices as ReturnType<typeof vi.fn>).mockReturnValue(false);
+
+    const lifecycle = pluginLifecycleManager(
+      {} as never,
+      { loader: bootLoader as PluginLoader, workerManager: makeWorkerManagerStub() },
+    );
+
+    // Backfill the real runtime-capable loader after construction.
+    const runtimeLoader = makeRuntimeLoaderStub();
+    lifecycle.setLoader(runtimeLoader as PluginLoader);
+
+    await lifecycle.enable("plugin-1");
+
+    // Activation must go through the backfilled loader...
+    expect(runtimeLoader.loadSingle).toHaveBeenCalledWith("plugin-1");
+    // ...and never touch the loader the manager was constructed with.
+    expect(bootLoader.loadSingle).not.toHaveBeenCalled();
+  });
+
+  it("emits plugin.enabled on the shared emitter so a co-located dispatcher listener fires", async () => {
+    mockRegistry.getById.mockResolvedValue(disabledPlugin);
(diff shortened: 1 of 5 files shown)


c-barlow:
Thanks — good catch on the startup window. Fixed in 0d1f8b0.

**The window:** because the loader is backfilled onto the shared lifecycle asynchronously (`setLoader` in the `.finally` of `bundledPluginsStartup`), `server.listen()` could accept requests before that ran on deployments that don't otherwise gate boot on plugin readiness — self-hosted, or managed instances with no declared sandbox environments (`applyManagedEnvironments` returns early before `await opts.pluginsReady` in those cases). During that window a route-driven `enable/disable/unload` drove a loader-less lifecycle: `activateReadyPlugin` no-ops (no runtime services), so the plugin persists `ready` and the dispatcher registers its tools, but no worker is spawned.

**The fix:** `index.ts` now `await`s the settled `bundledPluginsStartup` promise immediately before `server.listen()`, alongside the existing external-adapter wait. No request can reach a plugin route until `setLoader` has run.

**Why this ordering is safe:** managed instances that declare sandbox environments already blocked `listen()` on this same promise via `applyManagedEnvironments`, and they boot fine — so plugin workers do not require the HTTP server to be listening during activation. The promise never rejects (createApp swallows and logs bundled-plugin failures), so a degraded plugin cannot block the listener. The change adds no new activation path, so there is no double-start risk.

PR description also updated with the full required template sections.

c-barlow:
### Rebased onto current `master` (2026-09-18)

This PR had gone `CONFLICTING`/`DIRTY` against `master` and was no longer mergeable. Rebased onto `master@352153b5e`.

**The core fix is unchanged** — still the same 4-file diff sharing one `pluginLifecycleManager` between `routes/plugins.ts` and the tool dispatcher:

- `services/plugin-lifecycle.ts` — add `setLoader()`; `pluginLoaderInstance` becomes `let` so the loader can be backfilled (breaks the loader ↔ manager circular dependency without a second instance).
- `routes/plugins.ts` — `PluginRouteToolGatewayDeps` gains optional `lifecycleManager`; use the injected shared instance, fall back to a private one for tests.
- `app.ts` — pass the shared `lifecycle` into `pluginRoutes`, and `setLoader(loader)` in the `.finally` *after* the startup `loadAll()` pass (backfilling earlier would make `ensureBundledPlugins`' `lifecycle.load()` double-start bundled workers).
- New regression test `server/src/__tests__/plugin-lifecycle-set-loader.test.ts`.

I re-confirmed the bug is still live on current `master`: `routes/plugins.ts` still constructs its own manager, so route-emitted `plugin.enabled/disabled/unloaded` never reach the dispatcher's listeners.

**One commit was dropped:** `0d1f8b00` ("gate HTTP listener on plugin runtime wiring"), which awaited `bundledPluginsStartup` before `server.listen`. `master` has since moved the listener bind *deliberately early* — before startup recovery — with an explicit rationale that delaying it makes authenticated re-adoption of a surviving `runnerd` impossible and creates a duplicate-provider risk. Re-applying that commit would contradict that invariant, so I dropped it rather than force it through.

That leaves a narrow startup window (listener bound → `setLoader` backfill) where a plugin-mutating route could drive a loader-less lifecycle: status persists and the tool-registration event fires, but no worker spawns. It is pre-existing, orthogonal to the shared-instance bug this PR fixes, and wants a fix that doesn't delay the listener — happy to open a separate issue/PR for it.

**Verification** (fresh clone, `master@352153b5e`):
- `tsc --noEmit` on `@paperclipai/server`: **117 errors, byte-identical to the `master` baseline** — no new type errors. (The 117 are pre-existing, from the unbuilt `@paperclipai/paperclip-runner` package in a partial workspace build.)
- Targeted vitest — **78/78 passing**: `plugin-lifecycle-set-loader` (2), `plugin-lifecycle-restart`, `plugin-tool-dispatcher-pluginDbId`, `plugin-routes-authz` (48 across those 4), `bundled-plugins` (30).

Issue #2681, opened 2026-04-03
Claude code subscription?


Albatros79:
I just got an email from Claude. 

Hi,

Starting April 4 at 12pm PT / 8pm BST, you’ll no longer be able to use your Claude subscription limits for third-party harnesses including OpenClaw. You can still use them with your Claude account, but they will require extra usage, a pay-as-you-go option billed separately from your subscription.

Your subscription still covers all Claude products, including Claude Code and Claude Cowork. To keep using third-party harnesses with your Claude login. 

This will be enforced April 4 starting with OpenClaw, but this policy applies to all third-party harnesses and will be rolled out to more shortly.

I see Claude as a contributor but I wonder if this project will be affected.

jcpsimmons:
^ report for spam

Omee11:
#2793 (merged 2026-04-06) lists this issue as "Related" and in practice provides a workaround for the policy change you flagged: AWS Bedrock is now a first-class auth path on `claude-local` (`packages/adapters/claude-local/src/server/execute.ts:80-91`), which uses metered API billing rather than a Claude subscription. Operators affected by the third-party-harness policy change can switch to Bedrock without relying on subscription limits. That doesn't formally answer the policy question about whether Paperclip itself remains compliant, but it does mean the subscription restriction isn't a blocker. Deferring to the maintainer on closure.

PR #11347, opened 2026-08-13
feat: Claude login on the new-agent page before agent creation

nickyleach:
## Thinking Path

> - The Claude local adapter supports subscription login through a sandbox
> - The new-agent page must show login before the user creates an agent
> - Test results must not expose raw sandbox diagnostics or secret values
> - This pull request adds the login UI to both Test lanes and closes the diagnostic boundary
> - The branch also adds durable cleanup recovery for failed sandbox teardown
> - Reusable sandboxes must retain both their recorded teardown configuration and a valid lifecycle path until destruction succeeds
> - The benefit is a usable login flow with fixed public checks, redacted server logs, and recoverable sandbox cleanup

Related public work: [#9488](https://github.com/paperclipai/paperclip/pull/9488) adds first-class recognition for `CLAUDE_CODE_OAUTH_TOKEN` in headless and remote runs. Related public issue: [#2681](https://github.com/paperclipai/paperclip/issues/2681) requests Claude Code subscription support. This pull request adds the login transport and new-agent UI flow that those changes do not provide.

**Subsystem affected:** Claude local adapter, server login probes, sandbox provider setup, cleanup recovery, and the new-agent UI.

**Problem or motivation:** The Test lanes did not show the sandbox login panel in all supported cases. Test results also exposed raw probe diagnostics, and JSON escapes could end secret redaction early.

**Proposed solution:** Surface the login capability through the bundled provider manifest. Prepare the same probe runtime in the ACP lane. Send diagnostics only to redacted server logs. Keep Test checks on fixed public messages. Normalize login URL hints to allowlisted HTTPS Claude and Anthropic hosts. Consume JSON escapes during redaction. Preserve failed sandbox cleanup state across retries and restarts, and prevent deletion from severing the lifecycle context of a live reusable sandbox.

**Alternatives considered:** Keep raw diagnostics in Test checks or trust login URL text from the sandbox. Both choices increase information exposure. Keep separate probe behavior in the ACP lane. That choice would leave the two Test lanes inconsistent.

## What Changed

- Surface the sandbox login panel on both Test lanes.
- Reconcile the bundled Daytona plugin manifest so `supportsSetupTokenLogin` reaches the UI capability gate.
- Prepare the ACP Test lane with the same probe runtime as the CLI Test lane.
- Add the `claude_acp_login_probe_unavailable` warning when the ACP probe cannot run.
- Send raw sandbox diagnostics only to redacted server logs.
- Keep Test checks on fixed public messages in the ACP, managed-config, and CLI paths.
- Normalize login URL hints to allowlisted HTTPS Claude and Anthropic hosts.
- Redact JSON and escaped-JSON secret values, including escaped quotes and backslashes.
- Preserve orphan cleanup records across provider failures, restarts, and unavailable plugins.
- Atomically block environment deletion while a live reusable sandbox lease still depends on it.
- Verify pending cleanup destroys plugin sandboxes with the provider configuration recorded on the lease, even after the current environment configuration changes.

## Verification

- Head under review: `506b7fa2d83c36bfa5fd722ee9d95b0c7431c241`.
- Focused environment route/service/runtime coverage passes: 196 tests across 3 files.
- `pnpm -r typecheck` passes.
- `pnpm build` passes.
- The full Vitest run completed with 4,754 passing and 28 failing tests. All 23 source-test failures reproduce unchanged on parent head `58cfe61a33191ce03d965d65085d26064b4888ba`; the other 5 are duplicate executions from stale `server/dist` output. The failures are unrelated macOS path/listener and scheduler-fixture failures, so there is no new bad commit for bisect to localize.
- All required CI checks pass for the current head, including build, typecheck/release registry, all server and workspace shards, serialized server suites, canary, and e2e.
- A fresh Greptile review for `506b7fa2d83c36bfa5fd722ee9d95b0c7431c241` reports 5/5, “safe to merge,” with no blocking failure remaining.

## Risks

- A probe or redaction change could hide useful server diagnostics.
- An allowlist change could reject a valid Claude login URL.
- Cleanup recovery changes could affect provider teardown ordering.
- An environment with a live reusable sandbox can no longer be deleted until the owning issue or execution workspace completes teardown.
- The implementation keeps public Test messages fixed and sends detail to redacted server logs.

## Model Used

OpenAI GPT-5 via Codex — exact model ID: GPT-5; tool use and code execution enabled; extended reasoning enabled. The implementation author used AI-assisted development.

## Checklist

- [x] I have run tests locally and documented the result
- [x] I have updated relevant documentation or confirmed no separate documentation change is needed


server/src/services/environments.ts
```diff
@@ -1279,6 +1347,58 @@ export function environmentService(db: Db) {
...
+      providerLeaseId?: string | null;
+      metadata?: Record<string, unknown> | null;
+      failureReason: string;
+    }): Promise<EnvironmentLease> => {
+      const now = new Date();
+      const row = await db
+        .insert(environmentLeases)
+        .values({
+          companyId: input.companyId,
+          environmentId: input.environmentId,
```
nickyleach: [P1] Preserve pending cleanup across environment deletion

This record is still tied to `environment_leases.environment_id`, whose FK uses `ON DELETE CASCADE`, but the environment deletion blast-radius query only counts leases with `status = active`. An operator can therefore delete an environment while this `pending_cleanup` row exists; the cascade removes the only durable provider lease reference before the remote sandbox is destroyed, so no later sweep can retry it. Provider reconfiguration can similarly strand the retry behind the provider-mismatch guard. Please either block environment deletion/provider changes while unresolved cleanup rows exist, or persist orphan-cleanup state independently of the environment row (with enough immutable provider context to tear it down).

server/src/routes/environments.ts
```diff
@@ -1076,6 +1080,20 @@ export function environmentRoutes(
...
     assertNoClientPlatformProvisionedMarkers(req.body.metadata);
+    // A `pending_cleanup` lease holds the only durable provider reference for an
+    // orphan sandbox that a teardown retry must destroy. A driver or config
+    // change re-points the environment provider, so the sweep can no longer
+    // reach the original sandbox. Block a provider change while such a lease
+    // exists, so the operator resolves the cleanup first.
+    const changesProviderTarget =
+      (req.body.driver !== undefined && req.body.driver !== existing.driver) ||
+      req.body.config !== undefined;
+    if (changesProviderTarget && (await svc.hasUnresolvedPendingCleanupLeases(existing.id))) {
```
nickyleach, line 1114: [P1] Serialize provider changes with in-flight provisioning and orphan recording

This check runs before the environment update transaction, so it only protects against a cleanup row that already exists. One unsafe ordering is: the check returns false; an acquire provisions a sandbox using provider A; its lease insert/compensating teardown fails; the PATCH commits provider B; then the pending-cleanup row for A is inserted (or is inserted between this read and the config write). The retry resolves B and permanently hits the provider-mismatch guard.

Moving this query into the PATCH transaction alone is not sufficient: orphan-record insertion does not currently share a lock with the update, and provisioning begins before either a normal lease or cleanup row exists. The same untracked window also lets environment deletion win before the cleanup insert, after which the FK insert fails or the row is cascaded while the remote sandbox remains.

Please enforce the full invariant: from before remote provisioning starts until either the normal lease or a durable orphan-cleanup record exists, deletion and provider-target mutation must be mutually excluded. Alternatively, make cleanup state independent of the environment row/current config and persist enough immutable original-provider teardown context to clean it after deletion or reconfiguration. Please add concurrency coverage for both provider update vs. orphan recording and environment delete vs. orphan recording; the existing steady-state pending-row tests do not exercise either ordering.

server/src/services/environment-runtime.ts
```diff
@@ -1257,6 +1505,96 @@ function createSandboxEnvironmentDriver(
...
+            `Sandbox provider "${recordedProvider}" needs a plugin worker manager for cleanup, but none is available.`,
+          );
+        }
+        const pluginProvider = await resolveSandboxProviderPlugin({ provider: recordedProvider });
+        if (pluginProvider.state !== "running") {
+          throw new Error(
+            `Sandbox provider plugin for "${recordedProvider}" is ${pluginProvider.state}, so the cleanup teardown cannot run yet.`,
+          );
+        }
+        const config = await resolvePluginSandboxRuntimeConfig({
```
nickyleach: [P1] Preserve teardown authorization for secret-backed providers

This works for the new `fake`-provider tests, but not for Daytona or another plugin whose recorded config contains a secret reference. `resolvePluginSandboxRuntimeConfig` sends the recorded config back through normal environment-bound secret resolution. If the environment was deleted, the synthetic environment id is empty and `resolveConfigSecretRefsForRuntime` rejects it. If the environment was reconfigured, the old secret binding was replaced, so `secretService.resolveSecretValue` rejects the recorded API-key ref with `binding_missing`. In either ordering the durable row survives, but `environmentDestroyLease` is never called and every sweep retries forever.

Please give a `pending_cleanup` row durable, narrowly scoped authorization to resolve the exact recorded secret id/version needed for teardown, independent of the environment row and its current bindings. This should not put plaintext credentials in lease metadata. Closure needs plugin-provider coverage with a schema-declared secret-ref field for all three cases: delete before orphan recording, delete after orphan recording, and provider reconfiguration after orphan recording; each test should prove the old credential resolves and `environmentDestroyLease` receives it.

server/src/index.ts
```diff
@@ -963,6 +964,34 @@ export async function startServer(): Promise<StartedServer> {
...
+  // scheduler interval.
+  //
+  // This backstop is independent of the heartbeat scheduler toggle. A leaked
+  // provider sandbox costs money whether or not the instance schedules
+  // heartbeats, so both the enabled and the disabled path run the sweep. A
+  // disabled heartbeat scheduler must not strand a paid sandbox forever.
+  const environmentLeaseCleanupRuntime = environmentRuntimeService(db as any, { pluginWorkerManager });
+  const runEnvironmentLeaseCleanupSweep = () =>
+    environmentLeaseCleanupRuntime
+      .sweepPendingSandboxCleanups()
```
nickyleach: [P1] Use one pending-cleanup sweeper

This call starts a second sweeper for pending_cleanup leases. The heartbeat reaper from master also selects these rows. The two sweepers use different atomic claims. One claim changes cleanupStatus. The other claim changes a retry counter in metadata. Both claims can therefore succeed for the same ephemeral orphan. Both sweepers can then send a teardown request. If one request succeeds and the other request fails, the failed path can change the row from expired back to pending_cleanup.

Remove sweepPendingSandboxCleanups and use the master sweeper as the only owner of these rows. First, extend the master sweeper to keep all orphan-cleanup guarantees. For reuse_by_environment leases, keep destroyRunLease. For orphan ephemeral leases, call a small runtime teardown function. This function must accept a null environment. It must use the recorded provider configuration and the cleanup-authorized secret versions. Run the unified sweeper at startup and at the interval when heartbeat scheduling is enabled or disabled. Add a concurrency test that starts two unified sweeps for one orphan. Verify one provider teardown and a final successful lease state.

nickyleach:
@greptile review

nickyleach:
@greptile review

nickyleach:
@greptile-apps review

nickyleach:
@greptile-apps review

nickyleach:
@greptile-apps review

server/src/services/heartbeat.ts
```diff
@@ -13310,28 +13353,93 @@ export function heartbeatService(db: Db, options: HeartbeatServiceOptions = {})
...
+            cleanupStatus: "success",
+            failureReason: "pending_cleanup_retry",
+          });
           destroyed += 1;
+        } else if (environment) {
+          const result = await environmentRuntime.destroyRunLease({
+            environment,
+            lease,
+            failureReason: "pending_cleanup_retry",
+          });
```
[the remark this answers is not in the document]
nickyleach: Thanks — I traced this through the actual `destroyRunLease` path. For plugin-backed leases, `resolvePluginSandboxRuntimeConfig` reads the provider and configuration recorded in lease metadata first; built-in providers likewise reconstruct from lease metadata. The current environment configuration is only a fallback for legacy leases that lack recorded teardown context, so changing the environment after provisioning does not redirect cleanup. Commit `506b7fa2d` corrects the misleading heartbeat comment and adds an end-to-end regression that provisions with provider/config A, changes the environment to provider/config B, runs the real pending-cleanup sweep, and asserts the destroy RPC receives A (and the lease expires successfully). The same commit also addresses the lifecycle hole behind the broader concern: environment deletion is now atomically blocked while an active/released/retained `reuse_by_environment` lease exists, so a live reusable sandbox cannot lose its environment link before normal scoped teardown.

nickyleach:
@greptile-apps review

PR #11351, opened 2026-08-14
fix(ui): read the onboarding company prefix from the path, not the route match

tonio-alucema:
## Thinking Path

> - New users meet the product through an onboarding wizard that creates their company, their first agent, and a starter task
> - The wizard also serves an existing company, at `/{PREFIX}/onboarding`, to add another agent to it
> - On that route the wizard ignores the company in the URL and opens at "create a company" instead
> - It reads the prefix with `useParams()`, but it renders beside `<Routes>` rather than inside it, so there is no route match to read
> - This pull request reads the prefix from the pathname, which is available without a match
> - The benefit is that the URL a user follows decides what the wizard asks them

No public issue exists for this. The problem follows.

**What happened?**

Open `/{PREFIX}/onboarding` for a company that already exists. The wizard opens at step 1 and asks the user to create a company. The company named in the URL is ignored.

**Expected behavior**

The wizard recognises the company in the URL and opens at step 2, so the user adds an agent to that company instead of creating a second one.

**Steps to reproduce**

1. Create a company, so it has an issue prefix.
2. Go to `/{PREFIX}/onboarding`.
3. Read the first screen. It asks for a company name.

`master` at `5ca7b4c1f`.

**Deployment mode**

Any. This is client-side routing and does not depend on the server.

## What Changed

- `ui/src/lib/onboarding-route.ts` — adds `companyPrefixFromOnboardingPath()`, which reads the prefix from the pathname.
- `ui/src/components/OnboardingWizard.tsx` — uses that value when the route match supplies none. One line, plus the import.
- `ui/src/lib/onboarding-route.test.ts` — six cases for the new function.

`OnboardingWizard` renders beside `<Routes>` in `App.tsx`, so `useParams()` returns nothing and `companyPrefix` was always `undefined`. `resolveRouteOnboardingOptions` then took its no-prefix branch every time. `useLocation()` needs only the router, not a match, and the wizard already calls it.

The route match is still read first. If the wizard later moves inside the route tree, this code does not need to change.

The new parser accepts the same shape as `isOnboardingPath()`: the prefix is the first of exactly two segments. One test asserts the two agree, because a disagreement would either open the wizard where no company resolves, or resolve a company where onboarding is not served.

### Why the change is this small

Three pull requests are open against `OnboardingWizard.tsx` — #9900, #9501 and #8982. A larger change there would collide with all three. Almost all of this lands in `onboarding-route.ts`, a small file of pure functions with existing tests.

## Verification

- `npx tsc --noEmit -p ui/tsconfig.json` — clean.
- `npx vitest run ui/src/lib/onboarding-route.test.ts` — 18 pass.
- `npx vitest run ui/src` — 3883 pass, 445 files.

One test shows the defect and the fix together. With `companyPrefix: undefined`, which is what the wizard supplied before, `resolveRouteOnboardingOptions` returns `{ initialStep: 1 }`. With the parsed prefix it returns `{ initialStep: 2, companyId: "c1" }`.

**Pre-existing failures, unrelated:** `IssueProperties.test.tsx` and `StatusCards/format.test.ts` fail on clean `origin/master` with these changes stashed. Both look date-dependent.

**Not done:** no manual browser check. The behaviour is covered by unit tests at the function boundary, and the wizard's own suite passes.

## Risks

Low. The route match is still preferred, so behaviour changes only where `useParams()` gave nothing — which today is every render of this component.

The parser returns a prefix only for a two-segment path ending in `onboarding`, so no other route can start matching. An unknown prefix already falls back to step 1 in `resolveRouteOnboardingOptions`, and that path is unchanged.

To revert, remove the fallback in the wizard. The new function has no other caller.

## Model Used

Claude Opus 5 (`claude-opus-5`), through Claude Code. Extended thinking enabled. Tool use enabled: file read and edit, shell command execution for typecheck and the test runs, and the GitHub CLI.

## Checklist

### DIFF
--- a/ui/src/components/OnboardingWizard.tsx
+++ b/ui/src/components/OnboardingWizard.tsx
@@ -43,7 +43,10 @@ import { DEFAULT_CODEX_LOCAL_BYPASS_APPROVALS_AND_SANDBOX } from "@paperclipai/a
 import { DEFAULT_CURSOR_LOCAL_MODEL } from "@paperclipai/adapter-cursor-local";
 import { DEFAULT_GEMINI_LOCAL_MODEL } from "@paperclipai/adapter-gemini-local";
 import { DEFAULT_OPENCODE_LOCAL_MODEL, isValidOpenCodeModelId } from "@paperclipai/adapter-opencode-local";
-import { resolveRouteOnboardingOptions } from "../lib/onboarding-route";
+import {
+  companyPrefixFromOnboardingPath,
+  resolveRouteOnboardingOptions,
+} from "../lib/onboarding-route";
 import { AsciiArtAnimation } from "./AsciiArtAnimation";
 import { FrontDoor } from "./FrontDoor";
 import { AgentCapsule } from "./AgentCapsule";
@@ -134,7 +137,13 @@ export function OnboardingWizard() {
   const queryClient = useQueryClient();
   const navigate = useNavigate();
   const location = useLocation();
-  const { companyPrefix } = useParams<{ companyPrefix?: string }>();
+  const { companyPrefix: matchedCompanyPrefix } = useParams<{ companyPrefix?: string }>();
+  // This component renders beside `<Routes>`, not inside it (`App.tsx`), so it
+  // has no route match and `useParams()` gives nothing. Read the prefix from
+  // the pathname, which `useLocation()` supplies without a match. The param is
+  // kept first so a future move inside the route tree needs no change here.
+  const companyPrefix =
+    matchedCompanyPrefix ?? companyPrefixFromOnboardingPath(location.pathname);
 
   // Support opening the wizard from a route (e.g. /onboarding or an existing
   // company's "add agent" entry point) in addition to the dialog context.
--- a/ui/src/components/OnboardingWizard.tsx
+++ b/ui/src/components/OnboardingWizard.tsx
@@ -233,6 +233,12 @@ export function OnboardingWizard() {
   // The company the *route* last supplied, so a navigation that stops naming
   // one can drop it without touching a company the wizard created itself.
   const routeCompanyIdRef = useRef<string | null>(null);
+  // The current company, mirrored so the sync effect can read it without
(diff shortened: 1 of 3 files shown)


tonio-alucema:
Valid, and this pull request introduced it. Before the change the route never supplied a `companyId`, so the wizard could not carry one across a navigation. Making the prefix resolvable made that state reachable.

Confirmed the mechanism: the sync effect sets `createdCompanyId` when the route options carry one, and has no branch for when they stop carrying one. Navigating from `/{PREFIX}/onboarding` to `/onboarding` leaves the wizard at step 1 holding the previous company.

Fixed by clearing the company when the route supplied one before and does not now. A ref holds the last route-supplied value, so **only** a route-derived company is cleared - one the wizard created itself, or restored from saved state, leaves the ref null and is left alone. Clearing those would discard real progress.

Added two tests at the function boundary for the transition, including the unmatched-prefix case.

Full ui suite: 3885 pass. The two failures in `IssueProperties` and `StatusCards/format` are unchanged and reproduce on clean `origin/master`.

@greptile-apps review

tonio-alucema:
Valid, again - the ownership tracking I added claimed too much.

The ref was set whenever the route *named* a company, including one the wizard created itself. Create a company in the wizard, navigate to that company's onboarding path, then on to `/onboarding`, and the company is cleared: the route claimed something it never supplied.

Now ownership is claimed only when the route value differs from the company already in hand. A route naming the current company has introduced nothing, so it takes none.

One thing worth calling out, since it is not visible in the diff's intent: the comparison reads `createdCompanyId` through a ref updated on every render rather than adding it to the effect's dependency array. The effect also calls `setStep`, so depending on the company would re-run it on every company change and drag the user back to the route's initial step mid-flow. That would have been a worse bug than the one being fixed.

Full ui suite: 3885 pass, unchanged. The two `IssueProperties` / `StatusCards` failures still reproduce on clean `origin/master`.

@greptile-apps review

PR #11352, opened 2026-08-14
feat(ui): send a company with no agent into onboarding, at the right step

tonio-alucema:
## Thinking Path

> - A company does its work through agents, so a company with no agent can do nothing
> - The dashboard shows that state as a banner and asks the customer to act on it
> - A company created by Paperclip Cloud always reaches that state: Cloud creates the company before the tenant starts, so the redirect that sends companyless users to onboarding never runs
> - The customer finishes signup on one origin and arrives on an empty dashboard on another, with nothing to do next
> - This pull request sends a company with no agent into onboarding, and opens it on the agent step when the company already has its mission
> - The benefit is that a new company always has a next step, and never asks for an answer the customer already gave

**What happened?**

A company with no agent stays on the dashboard. The dashboard shows "You have no agents" with a "Create one here" link. Nothing takes the customer to onboarding.

For a company created by Paperclip Cloud the result is worse. Cloud collects a company name and a mission during signup, creates the company, then sends the customer into the tenant. The customer sees an empty dashboard immediately after answering a signup flow.

**Expected behavior**

A company with no agent opens onboarding. A company that already has its mission opens onboarding on the agent step, not on the mission question.

**Steps to reproduce**

1. Create a company through Paperclip Cloud signup, with a mission.
2. Enter the tenant.
3. The dashboard shows "You have no agents" and stays there.

`master` at `aac6ce82e`.

**Related pull requests**

#9900, #9501 and #8982 also change `OnboardingWizard.tsx`. All three are 3 to 4 weeks stale. This branch does not wait for them.

## What Changed

- `ui/src/lib/onboarding-route.ts` — adds `shouldRouteAgentlessCompanyToOnboarding`, `ONBOARDING_MISSION_STEP` and `ONBOARDING_AGENT_STEP`, and `onboardingStepForCompany`.
- `ui/src/hooks/useCompanyMission.ts` — new. Answers "does this company have its mission?", and says when that answer can be used.
- `ui/src/lib/onboarding-auto-open.ts` — new. Holds the companies already offered onboarding.
- `ui/src/components/OnboardingWizard.tsx` — applies the step once, when the wizard opens. Makes the mission step work for a company that already exists.
- `ui/src/pages/Dashboard.tsx` — opens onboarding when the agent list has loaded and is empty, once per company.
- Tests: `onboarding-route.test.ts` (11 new cases), `useCompanyMission.test.tsx`, `onboarding-auto-open.test.ts`, `OnboardingWizard.step.test.tsx` (all new).

### The step is decided once

The first four review passes on this branch found four defects. The last two traded against each other: the fix for the third re-opened the hole the second closed. That is a loop, so the cause was structural.

`initialStep` is a **derived** value. It comes from the company list and the goal list, so it changes whenever either query changes — a retry, a background refetch, a cache invalidation. The sync effect took it as a dependency and called `setStep`. Every one of those changes moved a customer who was already mid-flow. Gating the input narrowed the window. It could not close it, because the effect ran again each time the value changed.

The step now belongs to the **request** that opened the wizard, not to the latest value of the expression that produced it. The effect reads the step through a ref, and it is keyed on the two things that make a new request: the wizard opens, or the company changes. Navigating from one company's onboarding path to another still decides the step again. The same request settling on a fresher value does not.

`createdCompanyIdRef`, two lines above, already uses this pattern for the same reason.

### The mission step now works for a company that already exists

Nothing sent an existing company to the mission step before this change. A company got there only by creating itself on step 1, so several things on that step had never run. The dashboard auto-open reaches all of them. Each is unchanged on `master` and unreachable there.

**The step could not be completed.** `companyName` is only ever typed on step 1, so a company entering the wizard further along has none. The step prints the name in its own copy, and both ways forward — the button and the Enter key — require `companyName.trim()`. The name is now backfilled from the company list, beside the effect that already backfills the issue prefix.

**The mission was not saved.** `handleConfirmMission` advanced without writing anything when the company already existed. The customer typed a mission and the company kept none. It now writes the company-level goal.

**A second mission could be written.** The write is guarded on `createdCompanyGoalId`, and it reconciles first: it reads the goal list once and updates an existing company-level goal rather than adding another. This matters because the mission lookup fails open, so a company that already has a mission can still be asked for one. The customer answered the question on a step that asked it, so their answer becomes the mission. A read that fails still writes.

**Enter had no in-flight guard.** Every button is disabled while `loading`; the keyboard path was not, so a second Enter re-entered a handler whose guard is state the first had not set yet. One check at the top of `handleKeyDown` covers the mission step and the older paths above it.

### Company-scoped state stays with its company

Three related leaks, all the same shape: something that describes one company outliving the wizard's hold on it.

**Replacing and withdrawing a company clear the same state.** `clearCompanyScopedState` holds that list once, and both branches of the sync effect call it. A kept goal id is the sharpest edge — the guard above reads it as "mission already written", so the next company is never asked, and the launch path links its project to the previous company's goal.

**A company the wizard created is withdrawn too.** The route recorded ownership only when it named a company the wizard was not already holding, so a self-created company was never withdrawn: create on step 1, visit its own onboarding path, then go to `/onboarding`, and the wizard shows "create a company" while still holding the old one. Ownership is now recorded whenever the route names a company; only the clearing stays conditional. This reverses a rule the code documented, deliberately: `/onboarding` is a request to start a company, so honouring it and losing step-1 progress beats silently writing into a different company.

**An in-flight write cannot outlive its company.** `stillTheSameCompany` holds that rule once, and all five async writes ask it. It never cancels the server work, which is done and correct either way, and the cache for the company that *was* written is refreshed regardless. It declines only to record that work against a company it does not belong to — and it sits *after* the server work, so a company switch mid-hire cannot leave a hired agent without its instructions file.

### Two simplifications

**One mission lookup.** The route path and the dashboard each ran the same query, the same selector and the same settle gate. `useCompanyMission` now holds the query and its settlement. `onboardingStepForCompany` holds the rule that an unknown mission counts as absent. Neither adds a request: both read the goal list under the `queryKeys.goals.list` key the launch path already uses.

**One offer per company.** The dashboard's guard was a ref holding one company, so it died with the component and forgot every company but the last. `claimOnboardingOffer` holds the set at module scope, so an offer stays declined for the visit. A reload offers again, which is the right scope: a reload is a new visit rather than the same one continuing.

### Three decisions kept from the earlier commits

**It opens the dialog; it does not navigate.** The wizard is already mounted, so there is no route to race and no redirect to loop.

**It waits for the agent list.** Callers hold agents as `Agent[] | undefined`. An in-flight query and an empty company look the same, so deciding early would send every user through onboarding on each cold load.

**A lookup that cannot answer costs the step, not the flow.** `useCompanyMission` reports `settled` when the query stops being pending, whether it answered or failed. A gate that waits for the data itself fails closed: a request that exhausts its retries leaves the value undefined forever, and onboarding never opens. On failure the mission reads as absent, the customer answers it again, and the flow continues. This is the fail-open rule the wake and provisioning readiness gates follow. The reconciling write above is what makes that price affordable.

## Verification

- `ui` typecheck is clean.
- 58 pass across `onboarding-route.test.ts`, `useCompanyMission.test.tsx`, `onboarding-auto-open.test.ts`, `OnboardingWizard.step.test.tsx` and `OnboardingWizardVariant.test.tsx`.
- Full `ui` suite: **3923 pass**.
- All CI gates are green.

**Tested at the seam.** Every defect on this branch lived between a value and its consumer. The predicate tests passed at every stage, so tests over the predicates were never going to catch them. `OnboardingWizard.step.test.tsx` renders the real wizard with the real route resolver and the real mission hook, and stubs only the network and the surrounding contexts.

| # | Case | Expected |
| --- | --- | --- |
| 1 | Seeded company on `/PC1/onboarding` | Opens on the agent step |
| 2 | Mission lookup in flight | Stays closed |
| 3 | Mission lookup fails | Opens on the mission step |
| 4 | Lookup fails, then a refetch finds a mission | Wizard does not move |
| 5 | Dialog re-opened with a different step | Wizard does not move |
| 6 | Route names a different company | Step is decided again |
| 7 | Wizard closed, then re-opened | Step is decided again |
| 8 | Company created, then a route supplies one | The created company is not adopted |
| 9 | Mission step, existing company | Names the company, and can be completed |
| 10 | Mission confirmed | The goal is written |
| 11 | Mission confirmed, company already had one | The goal is updated, not duplicated |
| 12 | Mission confirmed, pre-write read fails | The goal is still written |
| 13 | Enter pressed twice | One goal is written |
| 14 | Company switched during the write | The next company is still asked, and writes its own |
| 15 | Company switched after confirming | The next company writes its own mission |
| 16 | Route withdraws the company | Nothing is carried to the next one |
| 17 | Self-created company, route stops naming it | It is withdrawn |
| 18 | Mission confirmed twice | One goal is written |

Every case was fault-injected against the code it replaces rather than trusted on a green run, and that mattered twice. One case first passed against the broken code, because it asserted on a rendered name that reads the same either way — a test that cannot fail reads as coverage while providing none, so it now asserts on the company selection instead. Another surfaced a race in the guard itself: the create path asked `stillTheSameCompany` against a render mirror that had not updated yet, so the handler that had just created a company could be told it no longer owned it, skip the goal write, and strand the customer on the mission step. Network latency usually hides that one.

One failure remains, in `IssueProperties.test.tsx`. It is timezone-dependent, it reproduces on this branch with these changes removed, and it is in a file this change does not touch.

**Not done:** no end-to-end run through Cloud signup. That needs a staging tenant on this build.

## Risks

Low. The step no longer changes under an open wizard, the dashboard no longer re-opens a wizard the customer closed, the mission step writes what it asks for, and no state or in-flight write outlives the company it describes.

The remaining risk is a loop: a company that finishes the wizard without creating an agent still has no agent. Four things bound it. The predicate refuses to act while already on an onboarding path. It opens a dialog rather than navigating. Each company is offered once per visit. The customer can close it, and it stays closed.

The mission path adds one request, when the customer presses "Confirm mission", to avoid writing a second company goal. The step decision adds none.

One behavior outside the new paths changed deliberately: navigating to `/onboarding` now withdraws a company the wizard created, where before it kept it. See "A company the wizard created is withdrawn too" above for the reasoning, and push back there if you read that trade the other way.

Three open pull requests change the same file. See "Related pull requests" above.

## Model Used

## Checklist

### DIFF
--- a/ui/src/components/OnboardingWizard.step.test.tsx
+++ b/ui/src/components/OnboardingWizard.step.test.tsx
@@ -475,6 +475,56 @@ describe("OnboardingWizard — which step it lands on", () => {
       );
     });
 
+    it("does not hand a new company the mission written for the old one", async () => {
+      // A route change can switch companies while the write is in flight, and
+      // the switch clears exactly the state the write is about to set. The
+      // goal is written and correct either way — but attributing it to the
+      // company now in hand would undo the clearing and let that company skip
+      // its own mission.
+      let resolveCreate: (goal: { id: string }) => void = () => {};
+      mockGoalsApi.create.mockReturnValue(
+        new Promise<{ id: string }>((resolve) => {
+          resolveCreate = resolve;
+        }),
+      );
+      routerState.pathname = "/PC1/onboarding";
+      await render();
+      await settle();
+      expect(currentStep()).toBe("mission");
+
+      const direct = [...document.body.querySelectorAll("button")].find((b) =>
+        b.textContent?.includes("I know my mission"),
+      )!;
+      await click(direct);
+      setControlledValue(missionTextarea()!, "Acme's mission");
+      await settle();
+      await click(confirmMissionButton()!);
+
+      // Switch companies before the write lands, then let it land.
+      routerState.pathname = "/PC2/onboarding";
+      await rerender();
+      await settle();
+      await act(async () => resolveCreate({ id: "goal-company-1" }));
+      await settle();
+
+      // Globex must still be asked, and must write its own mission.
+      expect(currentStep()).toBe("mission");
+      mockGoalsApi.create.mockResolvedValue({ id: "goal-company-2" });
+      const direct2 = [...document.body.querySelectorAll("button")].find((b) =>
+        b.textContent?.includes("I know my mission"),
+      )!;
+      await click(direct2);
+      setControlledValue(missionTextarea()!, "Globex's mission");
+      await settle();
+      await click(confirmMissionButton()!);
+      await settle();
+
+      expect(mockGoalsApi.create).toHaveBeenLastCalledWith(
+        "company-2",
+        expect.objectContaining({ title: "Globex's mission" }),
+      );
+    });
+
     it("does not write a second mission when the step is confirmed twice", async () => {
       mockGoalsApi.create.mockResolvedValue({ id: "goal-new" });
       await openOnMissionStepForExistingCompany();
--- a/ui/src/components/OnboardingWizard.step.test.tsx
+++ b/ui/src/components/OnboardingWizard.step.test.tsx
@@ -32,6 +32,7 @@ const mockGoalsApi = vi.hoisted(() => ({
   update: vi.fn(),
 }));
 const mockAdaptersApi = vi.hoisted(() => ({ list: vi.fn() }));
+const mockCompaniesApi = vi.hoisted(() => ({ create: vi.fn() }));
 
 const routerState = vi.hoisted(() => ({ pathname: "/" }));
 const dialogState = vi.hoisted(() => ({
@@ -52,7 +53,7 @@ const companyState = vi.hoisted(() => ({
 
 vi.mock("../api/goals", () => ({ goalsApi: mockGoalsApi }));
 vi.mock("@/api/adapters", () => ({ adaptersApi: mockAdaptersApi }));
-vi.mock("../api/companies", () => ({ companiesApi: { create: vi.fn() } }));
+vi.mock("../api/companies", () => ({ companiesApi: mockCompaniesApi }));
 vi.mock("../api/agents", () => ({
   agentsApi: { create: vi.fn(), adapterModels: vi.fn().mockResolvedValue([]) },
 }));
@@ -550,6 +551,56 @@ describe("OnboardingWizard — which step it lands on", () => {
       expect(mockGoalsApi.create).toHaveBeenCalledTimes(1);
       expect(currentStep()).toBe("agent");
     });
+  });
+
+  it("does not adopt a company it created once a route has supplied one", async () => {
+    // The same guard from the other end. Nothing was in hand when the create
+    // started, so "unchanged" means still nothing. A route that supplied a
+    // company while the request was open has taken over the wizard, and
+    // adopting the new company would fight it — and would leave the customer
+    // on a company they never navigated to.
+    let resolveCreate: (company: { id: string; issuePrefix: string }) => void = () => {};
+    mockCompaniesApi.create.mockReturnValue(
+      new Promise<{ id: string; issuePrefix: string }>((resolve) => {
+        resolveCreate = resolve;
+      }),
+    );
+    routerState.pathname = "/onboarding";
+    await render();
+    await settle();
+
+    // Step 1: name a new company, then confirm the mission to create it.
+    const nameInput = document.body.querySelector("input")! as HTMLInputElement;
+    setControlledValue(nameInput, "Initech");
+    await settle();
+    const next = [...document.body.querySelectorAll("button")].find(
+      (b) => b.textContent?.trim() === "Next",
+    )!;
+    await act(async () => {
+      next.dispatchEvent(new MouseEvent("click", { bubbles: true }));
+    });
+    await settle();
+    setControlledValue(missionTextarea()!, "Initech's mission");
+    await settle();
+    await act(async () => {
+      confirmMissionButton()!.dispatchEvent(new MouseEvent("click", { bubbles: true }));
+    });
+
+    // A route supplies an existing company before the create lands.
+    routerState.pathname = "/PC1/onboarding";
+    await rerender();
+    await settle();
+    expect(mockCompaniesApi.create).toHaveBeenCalledWith({ name: "Initech" });
+    await act(async () => resolveCreate({ id: "company-created", issuePrefix: "INI" }));
+    await settle();
+
+    // Adopting the created company would select it globally and take the
+    // customer off the one they navigated to. Asserted on the selection call
+    // rather than on the rendered name: the name reads "Acme" either way,
+    // because the switch reset clears it and the backfill refills it from the
+    // company list, which has no entry for the company just created.
+    expect(companyState.setSelectedCompanyId).not.toHaveBeenCalled();
+    expect(document.body.textContent).toContain("Acme");
   });
 
   it("applies the step again when the wizard is re-opened", async () => {
--- a/ui/src/components/OnboardingWizard.step.test.tsx
+++ b/ui/src/components/OnboardingWizard.step.test.tsx
@@ -1,0 +1,284 @@
+// @vitest-environment jsdom
+
+import { act } from "react";
+import { createRoot, type Root } from "react-dom/client";
+import { QueryClient, QueryClientProvider } from "@tanstack/react-query";
+import { afterEach, beforeEach, describe, expect, it, vi } from "vitest";
+import { queryKeys } from "../lib/queryKeys";
+import {
+  ONBOARDING_AGENT_STEP,
+  ONBOARDING_MISSION_STEP,
+} from "../lib/onboarding-route";
+
+/**
+ * Which step the onboarding wizard *lands on*, and what is allowed to move it
+ * afterwards.
+ *
+ * These are seam tests on purpose. `initialStep` is derived from two queries
+ * and consumed by an effect that calls `setStep`, and every defect this file
+ * guards lived in that seam rather than in either side of it — the pure
+ * helpers in `onboarding-route.test.ts` passed while the wizard was moving a
+ * customer off the step they were typing on. So the real component is rendered
+ * here, with the real route resolver and the real mission hook, and only the
+ * network and the surrounding contexts are stubbed.
+ */
+
+// eslint-disable-next-line @typescript-eslint/no-explicit-any
+(globalThis as any).IS_REACT_ACT_ENVIRONMENT = true;
+
+const mockGoalsApi = vi.hoisted(() => ({
+  list: vi.fn(),
+  create: vi.fn(),
+}));
+const mockAdaptersApi = vi.hoisted(() => ({ list: vi.fn() }));
+
+const routerState = vi.hoisted(() => ({ pathname: "/" }));
+const dialogState = vi.hoisted(() => ({
+  onboardingOpen: false,
+  onboardingOptions: {} as { initialStep?: number; companyId?: string },
+  onboardingRouteDismissed: false,
+  closeOnboarding: vi.fn(),
+  setOnboardingRouteDismissed: vi.fn(),
+}));
+const companyState = vi.hoisted(() => ({
+  companies: [
+    { id: "company-1", name: "Acme", issuePrefix: "PC1" },
+    { id: "company-2", name: "Globex", issuePrefix: "PC2" },
+  ],
+  loading: false,
+  setSelectedCompanyId: vi.fn(),
+}));
+
+vi.mock("../api/goals", () => ({ goalsApi: mockGoalsApi }));
+vi.mock("@/api/adapters", () => ({ adaptersApi: mockAdaptersApi }));
+vi.mock("../api/companies", () => ({ companiesApi: { create: vi.fn() } }));
+vi.mock("../api/agents", () => ({
+  agentsApi: { create: vi.fn(), adapterModels: vi.fn().mockResolvedValue([]) },
+}));
+vi.mock("../api/approvals", () => ({ approvalsApi: { create: vi.fn() } }));
+vi.mock("../api/issues", () => ({ issuesApi: { create: vi.fn() } }));
+vi.mock("../api/projects", () => ({ projectsApi: { list: vi.fn(), create: vi.fn() } }));
+
+vi.mock("@/lib/router", () => ({
+  useLocation: () => ({ pathname: routerState.pathname }),
+  useNavigate: () => vi.fn(),
+  useParams: () => ({}),
+}));
+
+vi.mock("../context/DialogContext", () => ({
+  useDialog: () => dialogState,
+}));
+
+vi.mock("../context/CompanyContext", () => ({
+  useCompany: () => companyState,
+}));
+
+// Canvas/animation leaves — nothing to do with the step machinery.
+vi.mock("./AsciiArtAnimation", () => ({ AsciiArtAnimation: () => null }));
+vi.mock("./AgentCapsule", () => ({ AgentCapsule: () => null }));
+vi.mock("./FrontDoor", () => ({ FrontDoor: () => null }));
+
+const { OnboardingWizard } = await import("./OnboardingWizard");
+
+/** The mission step renders this heading; the agent step renders this input. */
+function currentStep(): "mission" | "agent" | "closed" | "other" {
+  const body = document.body;
+  if (!body.querySelector("[role='dialog'], .fixed.inset-0")) return "closed";
+  const headings = [...body.querySelectorAll("h3")].map((h) => h.textContent);
+  if (headings.includes("Define your mission")) return "mission";
+  if (body.querySelector("input[placeholder='Chief of staff']")) return "agent";
+  return "other";
+}
+
+const COMPANY_GOAL = {
+  id: "goal-1",
+  companyId: "company-1",
+  title: "Ship the thing",
+  description: null,
+  level: "company",
+  status: "active",
+  parentId: null,
+  ownerAgentId: null,
+  createdAt: new Date("2026-03-02T00:00:00Z"),
+  updatedAt: new Date("2026-03-02T00:00:00Z"),
+};
+
+describe("OnboardingWizard — which step it lands on", () => {
+  let container: HTMLDivElement;
+  let queryClient: QueryClient;
+  let root: Root | null = null;
+
+  async function render() {
+    root = createRoot(container);
+    await act(async () => {
+      root!.render(
+        <QueryClientProvider client={queryClient}>
+          <OnboardingWizard />
+        </QueryClientProvider>,
+      );
+    });
+  }
+
+  /** Re-render after mutating the stubbed contexts or the location. */
+  async function rerender() {
+    await act(async () => {
+      root!.render(
+        <QueryClientProvider client={queryClient}>
+          <OnboardingWizard />
+        </QueryClientProvider>,
+      );
+    });
+  }
+
+  async function settle() {
+    await act(async () => {
+      await new Promise((resolve) => setTimeout(resolve, 0));
+    });
+  }
+
+  beforeEach(() => {
+    container = document.createElement("div");
+    document.body.appendChild(container);
+    queryClient = new QueryClient({ defaultOptions: { queries: { retry: false } } });
+    // The wizard restores its step from localStorage, so a step left behind by
+    // an earlier case would decide the next one.
+    localStorage.clear();
+    routerState.pathname = "/";
+    dialogState.onboardingOpen = false;
+    dialogState.onboardingOptions = {};
+    dialogState.onboardingRouteDismissed = false;
+    mockAdaptersApi.list.mockResolvedValue([]);
+    mockGoalsApi.list.mockResolvedValue([]);
+  });
+
+  afterEach(async () => {
+    await act(async () => root?.unmount());
+    root = null;
+    queryClient.clear();
+    container.remove();
+    document.body.innerHTML = "";
+    vi.clearAllMocks();
+  });
+
+  it("opens a company that already has its mission on the agent step", async () => {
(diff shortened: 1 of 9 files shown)


tonio-alucema:
Both valid, both fixed in c-the-latest commit.

**Auto-open loses company context** — correct, and it defeated the point of the change: opening with empty options started the wizard at the front door with no company, so the new-company path would have created a *second* company instead of giving this one an agent. The dashboard now passes `companyId` and the step, and waits for the goal list before opening so the step is right the first time rather than corrected after.

**Late mission result resets progress** — also correct. The goal query could resolve after the wizard opened, flip `initialStep` from 2 to 3, and the sync effect would move someone off the mission they were typing. Route options are now held back until the mission is known, mirroring how they are already held back while companies load. The step is decided once and never changes underneath an open wizard.

Neither fix adds a request: both read the goal list the launch path already reads, under the query key it already uses.

Worth noting the second one is the same shape as the first — a value that arrives late and moves the UI under the user. Gating on it settling, rather than correcting afterwards, is what removes the class rather than the instance.

tonio-alucema:
Valid, and it was mine — the previous commit's gate failed **closed**.

Waiting on the goal *data* meant a request that exhausts its retries leaves the value undefined forever, so route onboarding never opens and the dashboard effect returns early every time. An agentless company would have had no automatic onboarding at all, which is worse than the late-step-change this was fixing.

Both gates now wait for the query to **settle** rather than to answer. On failure the mission reads as absent, the customer is asked for it again, and the flow continues. Asking a question twice is recoverable; never opening onboarding is not.

Worth naming, since this is the third finding on this PR and they form a sequence: each fix introduced the next defect one layer over. Empty dialog options → carry the company and step → gate on the data → gate fails closed → gate on settled. The through-line is that every one of them lives in the seam between a value and its consumer, which is exactly where my unit tests don't reach — they all passed at each stage.

This is also the same fail-open rule the wake and provisioning readiness gates already follow: a check guarding a convenience must never be able to block the thing it guards. I had that right elsewhere today and inverted it here.

tonio-alucema:
Valid — and this is the second finding again, reached through the refetch path rather than the initial load. My fix for the third reopened the hole the second one closed.

**Stopping here rather than patching it.** Four passes, four findings, and the last two trade against each other:

| # | Finding | My fix | What it caused |
| --- | --- | --- | --- |
| 1 | empty dialog options | carry company + step | — |
| 2 | late result moves the wizard | gate on the data arriving | #3 |
| 3 | gate fails closed | gate on the query settling | #4 |
| 4 | late refetch moves the wizard | — | (= #2) |

Patching #4 the obvious way re-creates #3. That is a loop, and it means the defect is structural rather than a sequence of oversights.

**The structural cause.** `initialStep` is a *derived* value feeding an effect that *imperatively* calls `setStep`. Any late change to that value — first load, retry, refetch, cache invalidation — moves a user who is already mid-flow. Gating the input only narrows the window; it cannot close it, because the effect will always re-run when the derived value changes.

**What an actual fix looks like:** decide the step **once**, when the wizard opens, and never again. Concretely, the sync effect should apply `initialStep` on the open transition only, not on every change to it — or the step should be captured into state at open time and the effect should stop depending on it. That is a change to the wizard's step machinery, which is precisely what #9900, #9501 and #8982 are editing, so it needs coordination rather than a quick follow-up commit.

Leaving this open rather than merging. The parts that stand up are worth keeping: `shouldRouteAgentlessCompanyToOnboarding` with its loading and same-path guards, the widened `initialStep` with `ONBOARDING_AGENT_STEP`, the company-level-goal mission signal (`selectDefaultCompanyGoalId` on the existing `queryKeys.goals.list` key — no new fetch), and 29 passing tests. The wiring to the sync effect is the part that needs the structural change above.

Worth noting for whoever continues: every one of these four defects lived in the seam between a value and its consumer, and my unit tests passed at every stage. Tests over the predicates were never going to catch any of them.

tonio-alucema:
Picked this back up and fixed it structurally, as the last comment said it needed.

**The cause.** `initialStep` is a *derived* value — it comes from the company list and the goal list — feeding an effect that imperatively calls `setStep`. The effect took it as a dependency, so every change to either query moved a customer who was already mid-flow. That is why gating the input never worked: a gate narrows the window, but the effect still re-runs whenever the value changes, so the window never closes. Four passes, and the last two traded against each other, because both were fixing the instance rather than the class.

**The fix.** The step belongs to the *request* that opened the wizard, not to the latest value of the expression that produced it. The effect reads the step through a ref and is keyed on the two things that make a new request: the wizard opening, and the company changing. Navigating from one company's onboarding path to another still re-decides the step; the same request settling on a fresher value does not.

This is the pattern `createdCompanyIdRef` already uses two lines above, with a comment giving the same reason. The file had the answer in it.

**Two simplifications follow.** The route path and the dashboard were each running the same goal query, the same selector and the same settle-gate; that collapses into `useCompanyMission`, with `onboardingStepForCompany` holding the rule that an unknown mission counts as absent. Neither adds a request. And the dashboard now opens onboarding once per company — every input to that effect sat behind a query, so a refetch called `openOnboarding` again and reopened a wizard the customer had closed. That one was live in the previous revision and no review pass caught it.

**On the tests.** The last comment was right that tests over the predicates were never going to catch any of this. `OnboardingWizard.step.test.tsx` renders the real wizard with the real resolver and the real hook, stubbing only the network and the surrounding contexts, and covers all four defects plus the two cases that guard against over-correcting — a different company, and a fresh open, both still re-decide the step. I confirmed each defect case fails against the code it replaces rather than trusting a green run: restoring the old dependency array fails the two "does not move" cases, and restoring the data gate fails the two fail-open cases.

3909 pass in the full `ui` suite. The one remaining failure is in `IssueProperties.test.tsx`, is timezone-dependent, and reproduces on this branch with these changes removed.

#9900, #9501 and #8982 still touch this file and are all 3-4 weeks stale. I did not wait on them. The change to the step machinery is four lines, so a rebase is cheap either way.

tonio-alucema:
Valid, and unlike the previous four this one is not another turn of the same loop — it is in the step this change routes people *to*, rather than in how that step is chosen. Fixed, along with a worse one next to it.

**The mission was discarded.** `handleConfirmMission` advanced without writing anything when the company already existed. That branch was dead code until now: a company reached step 2 only by creating itself on step 1, one line below, which writes the goal on the way through. The dashboard now opens an agentless company directly on that step, so the customer types a mission, presses "Confirm mission", and lands on the agent step with the company holding none — the exact state this PR exists to remove. Downstream only makes it quieter: `handleLaunchToDashboard` falls back to `selectDefaultCompanyGoalId`, finds nothing, and links the onboarding project to no goal. It now creates the company-level goal, guarded on `createdCompanyGoalId` so going back and confirming again does not add a second.

**And the step could not be completed at all.** Checking the same branch turned up the harder failure: `companyName` is only ever typed on step 1, so a company entering the wizard further along has none. Step 2 prints the name in its own copy, and *both* ways forward — the button and the Enter key — are gated on `companyName.trim()`. A customer sent there was stuck on a step asking about a company it could not name. It is now backfilled from the company list, beside the effect that already backfills the issue prefix for the same company and the same reason.

**Scope.** Both predate this branch and are unchanged on `master`. Neither was reachable there. This PR is what makes the mission step the default destination for an agentless company, so they are this PR's to fix.

Three cases added to the seam suite: the step names its company and can be completed, confirming writes the goal, confirming twice writes one. Each confirmed to fail against the code it replaces — removing the name backfill fails all three, restoring the bare `setStep(3)` fails the two about writing.

3912 pass in the full `ui` suite; the one failure is the timezone-dependent `IssueProperties` case that reproduces on this branch with these changes removed.

tonio-alucema:
Both valid, both mine, both fixed.

**Company switch retains onboarding state (P1).** Correct, and my previous commit is what sharpened it. That branch replaced the company id and prefix and kept everything else — which was survivable while `createdCompanyGoalId` could only be set by creating a company on step 1. Making the existing-company path set it too means a customer who confirms a mission for one company and then opens another hits the guard that stops a second goal, reading the *first* company's id: the second company's mission is silently skipped, and the launch path links its project to the first company's goal. The branch now clears the name, the mission text and the created resource ids alongside the id it was already clearing. It runs exactly where it already ran — only when the route introduces a company that is not the one in hand — so a route that merely names the company the wizard created still claims nothing.

**Enter has no in-flight guard.** Also correct. Every button in the wizard is disabled on `loading`; the keyboard path was not, so a second Enter re-enters a handler whose guard is state the first has not set yet. On the new mission path that is two goals for one mission. Worth noting it was already two companies for one company on step 1 and two agents for one hire on step 4 — same shape, older. One check at the top of `handleKeyDown` covers all of them rather than three guards in three handlers.

Two cases added, both confirmed to fail against the code they replace: a mission is not carried across a company switch, and a second Enter writes one goal.

Separately, the full-suite run surfaced an order-dependent failure in the new mission-hook suite — a single macrotask tick was enough in isolation but not under load. Both new suites now drain until the value settles instead of guessing at a tick count. 3914 pass over two consecutive full runs, with the same timezone-dependent `IssueProperties` failure the base has.

On the pattern: these last two are the same class as the original four — state that outlives the thing it describes. Worth saying plainly that the previous commit created the first one rather than merely exposing it.

tonio-alucema:
Both valid. Fixed.

**Failed lookup duplicates missions.** Correct, and it is the bill for failing open — worth paying properly rather than reversing the decision. `useCompanyMission` reports a failed lookup as "no mission", so a company that has one gets asked again; writing unconditionally then left two active company-level goals. One correction to the finding: within the session the launch path uses `createdCompanyGoalId`, so the project does link to the new goal — but everywhere *outside* this wizard `selectDefaultCompanyGoalId` picks the earlier one, so the mission the customer just typed loses anyway. Same outcome, different route.

`handleConfirmMission` now reads the goal list once before writing and updates the existing goal rather than adding another. Adopting the old goal and discarding what they typed would be the other way out, and it is wrong: they answered the question on a step that asked it, so their answer is the mission. A read that fails still writes — an unwritten mission stays the worse error at every level.

This does add one request, on the mission path only, at the moment the customer presses "Confirm mission". The step *decision* still adds none.

**Company dismissal history is lost.** Also correct, and worse than reported: the ref died with the component too, so navigating off the dashboard and back re-offered as well — not just switching companies. `claimOnboardingOffer` now holds the set, at module scope, so an offer stays declined for the visit. A reload offers again, which I think is the right scope: a reload is a new visit, not the same one continuing. Happy to make it durable if you disagree, but that is a product call about how insistent onboarding should be, so I did not make it unilaterally here.

Five cases added, each confirmed to fail against the code it replaces: the mission is updated rather than duplicated, a failed pre-write read still writes, and each company is offered once in any order.

3919 pass over two consecutive full runs. The one failure is the timezone-dependent `IssueProperties` case that reproduces on this branch with these changes removed.

Separately — the `General tests (workspaces-b)` failure on the previous run is `@paperclipai/adapter-utils > execution-target-sandbox.test.ts`, an empty stdout from a subprocess launch. This branch touches only `ui/src/**`, and that lane passed on the two commits before it. Flake, not a regression. `verify` failed only as its aggregate.

tonio-alucema:
Valid. Fixed — and it is the last open instance of the class the previous rounds have been working through: state that outlives the thing it describes.

The company-switch reset clears the created resource ids, but a write already in flight still holds the old company in its closure. When it lands it writes into the state the reset just cleared, so the new company inherits the old company's goal — and the guard that stops a second mission then reads that as "already written" and skips asking. The reset existed to prevent exactly that, so the write was undoing it.

Both writes now compare the company they started for against `createdCompanyIdRef`, which every render keeps current, and return instead of writing when the wizard has moved on. The server work is done and correct either way; only the attribution is declined. The `invalidateQueries` calls moved above the guard — the cache for the company that *was* written should refresh whether or not this wizard still cares about it. I gave the agent write next door the same guard, since this change routes people to that step too.

One case added, confirmed to fail against the code it replaces: switching companies while a mission write is in flight leaves the second company on the mission step, and it writes its own mission.

**On the two red CI lanes, neither of which is this branch.** `Verify serialized server suites (5/5)` is `@paperclipai/server > issue-closed-workspace-routes.test.ts`, failing with `transient database error`, `revalidation read failed` and a 5s timeout. The earlier `General tests (workspaces-b)` was `adapter-utils > execution-target-sandbox.test.ts` returning empty stdout from a subprocess. This branch touches only `ui/src/**`; both lanes passed on earlier commits of this same branch; `verify` fails only as the aggregate of its lanes.

**And a caveat on my own numbers.** My machine is currently at load 15, and full local suite runs have gone noisy — different unrelated files fail each run (`AgentConfigForm`, `TestPanel`, `AuditFeed`, `TweakPanel`, `Projects`, `ScheduleEditor`, `RunWorkspaceRecoverySurface`), all pass in isolation, and none import anything this branch touches. The reliable local figure is from the previous commit under a quiet machine: two consecutive runs at 3919 pass, with only the timezone-dependent `IssueProperties` failure that reproduces with these changes removed. CI is the better authority for the full suite here, so I would rather flag that than quote a number I do not trust.

tonio-alucema:
Valid, and the right response is to stop adding this check one site at a time — adding a third and fourth copy would be exactly the shape of mistake this branch is named for.

`stillTheSameCompany` now holds the rule once, with the reason, and all five writes ask it: the mission write, the agent write, and the goal, project and issue writes in the launch path. It never cancels the server work, which is done and correct either way. It declines only to record that work against a company it does not belong to.

The company *creation* path asks the same question from the other end. Nothing was in hand when it started, so "unchanged" means still nothing. A route that supplied a company while the request was open has taken over the wizard, and adopting the company just created would fight it — and would leave the customer on a company they never navigated to.

The `invalidateQueries` calls moved above every guard: the company that *was* written should have its cache refreshed regardless of where this wizard has got to.

**One note on the test, because it is the more useful finding.** The case I wrote for the creation path passed against the unguarded code on the first attempt. It asserted on the rendered company name, which reads "Acme" either way — the switch reset clears the name and the backfill refills it from a company list that has no entry for the company just created. A test that cannot fail is worse than no test, because it reads as coverage. It now asserts on the global company selection, which only happens when the created company is adopted, and it does fail against the unguarded code.

That is the same error class as the original four, one level up: an observable that is true for the wrong reason. Fault-injecting every case is what caught it, and it is why I have been doing that rather than trusting a green run.

**On CI:** both lanes that were red earlier are green on the last run — `General tests (workspaces-b)` and `Verify serialized server suites (5/5)` — which confirms those were flakes rather than this branch. My local full-suite runs remain noisy under machine load; the rotating failures all pass in isolation and none import anything this branch touches, so I am treating CI as the authority on the full suite.

tonio-alucema:
Both valid. The first is a regression my previous commit introduced, not an older defect, and it is worth naming as such.

**Stale completions.** I put the ownership check in the wrong place in the hire path — before the CEO instruction seeding. A company switch mid-hire therefore left a *hired agent with adapter defaults and no instructions file*. That inverts the rule the check exists for: the server work is done and belongs to whoever it was done for; only the attribution is this wizard's to decline. The seeding now runs first, and the check guards `setCreatedAgentId` and the step change after it.

The launch path had the matching gap at its end. The goal, project and issue writes were guarded, but the finish — select the company, `reset()`, navigate — was not. A switch mid-launch would take the customer back to the company they had left and `reset()` would discard the progress they had started on the new one. Guarded now, after all the server work completes.

**Route-company withdrawal.** Correct, and it is a half-fix of mine: I made the *replacement* branch clear the company-scoped state and left the *withdrawal* branch clearing only the id and prefix. Those are the same event — this company is no longer the wizard's — and clearing half of it leaves a goal id that the next company reads as "mission already written", so it never gets asked. `clearCompanyScopedState` now holds that list once and both branches call it.

One case added, confirmed to fail against the code it replaces: a mission confirmed on a company's onboarding path is not carried through a navigation to the unprefixed route.

Worth stating the shape, because two rounds in a row have now been my own fixes rather than pre-existing code: guarding *entry* to async work is wrong, guarding *attribution* after it is right, and any state that describes one company must be cleared on every path where the wizard stops holding that company — not on the first one I happened to find.

57 pass across the route, hook, offer and wizard suites. Not quoting a full-suite number this time: this machine is at load 28 and the figure would not mean anything. CI has run the suite green on the last two commits.

tonio-alucema:
Valid on the retention, wrong on the mechanism — and writing the test for it found a race I introduced two commits ago.

**The retention is real.** The route recorded ownership only when it named a company the wizard was not already holding, so a company the wizard *created* was never route-owned and never withdrawn. Create on step 1, visit that company's own onboarding path, then go to `/onboarding`: the wizard shows "create a company" while still holding the previous one, and the next confirmation writes the customer's new mission into it. That is precisely the failure the withdrawal branch was written to prevent, reached by a path it could not see. Ownership is now recorded whenever the route names a company, including the one in hand; only the clearing stays conditional.

That reverses a rule the code documented — not recording ownership there was protecting wizard-created work from a later `/onboarding`. The trade is deliberate: `/onboarding` is a request to start a company, so honouring it and losing step-1 progress beats silently writing into a different company. Flagging it explicitly in case you read that trade the other way.

**The mechanism in the report is wrong.** It describes the create path running and its adoption being rejected, leaving an orphan company and a stuck wizard. That cannot happen — `handleConfirmMission` branches on `createdCompanyId`, so a held company takes the existing-company branch and the create path is unreachable. No company is created and nothing is stuck. The consequence is a mission written to the wrong company, which is bad for a different reason.

**The race the test found.** The create path sets `createdCompanyId` and then asks `stillTheSameCompany(company.id)` before writing the goal — but the ref that answers is a render mirror, and the render had not always happened by then. So the handler that had just created the company could be told it no longer owned it, skip the goal write, and leave the customer sitting on the mission step they had just completed. Network latency usually hides it, which is the worst property a bug can have. The mirror is now updated eagerly beside the state, the way the route ref already was.

That one is worth naming: the guard I added to prevent stale writes could itself reject a *fresh* one. I only found it because the new case drove the real step-1 flow rather than asserting on the guard in isolation.

Two cases added, each confirmed to fail against the code it replaces.

Load on my machine finally dropped, so the full local suite is meaningful again: **3923 pass**, with only the timezone-dependent `IssueProperties` failure that reproduces with these changes removed.

PR #9684, opened 2026-07-16
test(server): add tests for company deletion with budget incidents and policies

jmao0001:
## Thinking Path

> - Paperclip is the open source app people use to manage AI agents for work
> - Companies act as the top-level boundary for isolation, storing agents, projects, goals, and budget events.
> - An operator deleting a company triggers a hard-delete transaction that purges child tables in sequence to avoid foreign key violations.
> - This pull request includes a test to verify that the fix correctly addresses the `budget_incidents` and `budget_policies` foreign key constraint errors when deleting a company.
> - The benefit is that companies loaded with actual data (and active budget allocations) can be successfully deleted again via the API/CLI without manual database intervention, and these deletions are now covered by integration tests to prevent future regressions.

## Linked Issues or Issue Description

Closes #7781

## What Changed

- Added a test file suite `it("successfully removes a company that has budget policies and incidents")` in `server/src/__tests__/companies-service.test.ts`.
- Included cleanup in the `afterEach` hook for `budgetIncidents` and `budgetPolicies` to maintain test stability and teardown safely.

## Verification

- The new integration test ensures that creating a company, associating it with `budgetPolicies` and `budgetIncidents`, and executing the `companyService.remove` method successfully bypasses Postgres FK constraint violations and accurately removes related parent/child dependencies.
- You can confirm this passes cleanly by running `pnpm test` locally.

## Risks

- Low risk. Existing data remains unaffected. This only acts to provide test coverage for previously buggy database deletion flow handling.

## Model Used

- Gemini 3.1 Pro Preview (human-reviewed)

## Checklist

- [x] I have included a thinking path that traces from project context to this change
- [x] I have specified the model used (with version and capability details)
- [x] I have checked ROADMAP.md and confirmed this PR does not duplicate planned core work
- [x] I have searched GitHub for duplicate or related PRs and linked them above
- [x] I have either (a) linked existing issues with `Fixes: #` / `Closes #` / `Refs #` OR (b) described the issue in-PR following the relevant issue template
- [x] I have not referenced internal/instance-local Paperclip issues or links (only public GitHub `#NNN` / `github.com/paperclipai/paperclip` URLs)
- [x] My branch name describes the change (e.g. `docs/...`, `fix/...`) and contains no internal Paperclip ticket id or instance-derived details
- [x] I have run tests locally and they pass
- [x] I have added or updated tests where applicable
- [x] I have updated relevant documentation to reflect my changes
- [x] I have considered and documented any risks above
- [x] All Paperclip CI gates are green
[...]

Issue #8489, opened 2026-06-22
Company delete fails (500) when company has cost_events/finance_events: FK to heartbeat_runs lacks ON DELETE CASCADE


mingqxu7:
## Summary

`DELETE /api/companies/:id` (CLI: `paperclipai company delete <id> --yes --confirm <id>`) returns **500 Internal Server Error** for any company that has accrued `cost_events` (i.e. any company whose agents have actually run). The cascade deletes `heartbeat_runs` while `cost_events.heartbeat_run_id` still references them, and that FK has `ON DELETE no action`, so Postgres rejects the delete.

## Repro

1. Import/create a company and let an agent run at least once (this writes `cost_events`).
2. `paperclipai company delete <companyId> --yes --confirm <companyId>`

**Result:** `API error 500: Internal server error`

## Server error

```
DrizzleQueryError: Failed query: delete from "heartbeat_runs" where "heartbeat_runs"."company_id" = $1
caused by: PostgresError: update or delete on table "heartbeat_runs" violates foreign key constraint
"cost_events_heartbeat_run_id_heartbeat_runs_id_fk" on table "cost_events"
    at .../@paperclipai/server/dist/services/companies.js:340
    at .../@paperclipai/server/dist/routes/companies.js:391
```

## Root cause

In `packages/db/src/schema/`, these foreign keys are declared without an `onDelete` action (migration `0031_zippy_magma.sql` emits `ON DELETE no action`):

- `cost_events.heartbeat_run_id` → `heartbeat_runs.id`  (the first blocker)
- `finance_events.heartbeat_run_id` → `heartbeat_runs.id`  (next blocker)
- `finance_events.cost_event_id` → `cost_events.id`  (next blocker)

The company-delete service deletes `heartbeat_runs` (and later `cost_events`) by `company_id`, but the child rows above are not cleared/cascaded first, so the FK check fails.

## Proposed fix

Add `onDelete` to these references in the Drizzle schema and ship a migration that re-creates the constraints. `cascade` matches the company-delete intent (the parent rows are being deleted anyway); `set null` is an alternative if these finance records must survive run/cost-event deletion.

```ts
// cost_events.ts
heartbeatRunId: uuid("heartbeat_run_id").references(() => heartbeatRuns.id, { onDelete: "cascade" }),

// finance_events.ts
heartbeatRunId: uuid("heartbeat_run_id").references(() => heartbeatRuns.id, { onDelete: "cascade" }),
[...]

Issue #7942, opened 2026-06-10
[Bug] Company delete fails — cost_events/finance_events deleted after heartbeat_runs (FK violation)


NoJster:
### Describe the bug

Deleting a company via `DELETE /api/companies/:id` returns a 500 error. The `remove()` function in `services/companies.js` deletes `heartbeat_runs` before deleting `cost_events` and `finance_events`, but both of those tables have FK references to `heartbeat_runs.id` with `ON DELETE no action`, causing a PostgreSQL foreign key violation.

### Steps to reproduce

1. Have a company with agent activity (heartbeat runs, cost events)
2. Archive the company
3. Attempt to delete via API or CLI:
   ```bash
   npx paperclipai company delete --confirm <id> <id> --yes
   ```
4. Server returns 500

### Server log

```
ERROR: DELETE /api/companies/<id> 500
Failed query: delete from "heartbeat_runs" where "heartbeat_runs"."company_id" = $1
```

Stack trace points to `services/companies.js:340` → `drizzle-orm/pg-core/session.js:41`.

### Root cause

In `services/companies.js`, the `remove()` function (line 326) deletes tables in dependency order, but the ordering is wrong:

```
line 340: DELETE heartbeat_runs WHERE company_id =     ← FAILS
...
line 345: DELETE cost_events WHERE company_id =        ← should be before line 340
line 346: DELETE finance_events WHERE company_id =     ← should be before line 340
```

Both `cost_events.heartbeat_run_id` and `finance_events.heartbeat_run_id` reference `heartbeat_runs.id` with `ON DELETE no action`. PostgreSQL blocks the `heartbeat_runs` delete because referenced rows still exist.

### Suggested fix

Move the `cost_events` and `finance_events` deletes before the `heartbeat_runs` delete. Additionally, `finance_events` has a FK to `cost_events.id`, so `finance_events` should be deleted before `cost_events`:

```javascript
// Correct order:
await tx.delete(financeEvents).where(eq(financeEvents.companyId, id));
await tx.delete(costEvents).where(eq(costEvents.companyId, id));
// ... then later:
await tx.delete(heartbeatRuns).where(eq(heartbeatRuns.companyId, id));
```

### Environment

- `@paperclipai/server@2026.609.0`
- Node.js v24.15.0
- Embedded PostgreSQL

schmelli:
## Plan: Fix company delete FK violation in `remove()`

### Problem
[...]

Issue #7250, opened 2026-05-31
Hard-delete API returns deterministic 500 for companies/agents/projects with run/issue/secret history (cascade gap; UI archive works)


CosmicDreamer17:
## Summary

`DELETE /api/companies/{companyId}` returns a deterministic **HTTP 500** with an opaque body (`{"error":"Internal server error"}`) for any company that has agents/projects carrying **run / issue / comment / secret-binding history**. The company is **not** deleted (verified intact after every attempt). The same 500 occurs on `DELETE /api/agents/{id}` and `DELETE /api/projects/{id}` for historied entities — so the parent company can't be torn down either.

The UI "delete" path instead **soft-deletes** the company (`status` → `archived`) and succeeds. Only the API hard-delete fails.

## Isolation (points at a cascade gap)

On one company I deleted entities individually:

| Entity | History | `DELETE` result |
|---|---|---|
| agent with no secrets, **never heartbeated** (no runs) | none | **`200 {"ok":true}`** |
| agent with `secret_ref` env + heartbeat/run history | has runs | **`500`** |
| project with issues/comments | has children | **`500`** |
| company (after the clean agent was removed) | — | **`500`** |

The clean, history-free agent deleted fine; every entity with dependent rows (runs, issues, comments, secret bindings) 500s. This strongly suggests the hard-delete cascade does not handle those dependent rows (likely an unhandled FK / missing cascade), whereas the archive path sidesteps it.

## Repro

1. Create a company via import into a new company:
   `paperclipai company import <pkg> --include company,agents,projects,issues,skills --target new --new-company-name "<Name>" --yes`
   (result: agents with `secret_ref` GITHUB_TOKEN bindings + managed instruction bundles, projects with materialized repo workspaces, imported issues with comments, company secrets, plus an approval).
2. Exercise at least one agent (a heartbeat run) so it has run history.
3. Attempt delete — both paths fail identically:
   - CLI: `paperclipai company delete <company-id> --yes --confirm <company-id>`
   - Raw: `curl -X DELETE /api/companies/<company-id>` (board bearer token)
4. Observed: `HTTP 500 {"error":"Internal server error"}`, deterministic across repeated attempts; company intact.
5. Per-entity teardown also 500s for any agent/project with history (see table); a history-free agent deletes at 200.
6. UI delete on the same company succeeds as a soft-delete (`status` → `archived`).

## Expected

[...]

PR #11354, opened 2026-08-14
feat(cli): add `company purge` for company decommission

nickallevato:
- [x] I searched the open PR list for similar PRs before opening this one.

## Linked Issues

Same `cost_events` → `heartbeat_runs` ordering failure:
Refs #6485, Refs #7942, Refs #8489, Refs #5014, Refs #5163, Refs #1659

The coverage half — the delete list not keeping up with new tables:
Refs #2096, Refs #7250, Refs #3038, Refs #3577

Deliberately `Refs` rather than `Fixes`: this PR does **not** repair
`DELETE /api/companies/:id`. It adds a working operator path alongside it. The
server-side fix is #6756 below, and these issues should stay open until that
lands.

## Related PRs

- **#6756** — `fix(server): delete cost_events + finance_events before heartbeat_runs in companies.remove()`. Fixes the ordering bug at the source. Complementary to this PR and arguably should land first; this one additionally covers the tables the hand-written list does not know about.
- **#2097** — `fix(server): complete FK dependency cleanup in company delete`. Same coverage problem, addressed by extending the list rather than deriving it.
- **#9684**, **#9593** — tests around company deletion cleanup.

If maintainers would rather solve this entirely server-side via #6756 and #2097,
this PR still has value as the offline/recovery path and as the thing that
covers plugin schemas — but I am happy to close it in favor of those.

## Thinking Path

Started from a concrete failure: `DELETE /api/companies/:id` returned an error on a company with cost history. Reproduced it by replaying `companyService.remove()`'s exact statement order inside a rolled-back transaction, which failed on statement 4:

```
ERROR: update or delete on table "heartbeat_runs" violates foreign key constraint
       "cost_events_heartbeat_run_id_heartbeat_runs_id_fk" on table "cost_events"
```

The list deletes `heartbeat_runs` before `cost_events`, but that FK is `ON DELETE NO ACTION`.

Fixing the order alone would leave the deeper problem: the list is hand-maintained, while 145 tables carry a `company_id` and plugins add more at install time in their own `plugin_*` schemas. Any new table with a NO ACTION FK silently breaks company deletion again, and nothing tests it.

I considered `session_replication_role = 'replica'` to sidestep ordering entirely, and rejected it — it disables FK triggers wholesale, including the ~88 `ON DELETE CASCADE` edges, so it converts a loud failure into silent orphans. Deriving the plan from the catalog and letting Postgres keep enforcing constraints gives the opposite property: if it commits, integrity held.

## Issue

`DELETE /api/companies/:id` cannot complete on any company that has cost history, so companies cannot be removed and their data accumulates indefinitely.

**Steps to reproduce**
1. Use a company with at least one row in `cost_events`.
2. `DELETE /api/companies/:id`.

**Expected:** company and its data are removed.
**Actual:** foreign key violation; the transaction aborts and nothing is deleted.

**Paperclip version/commit:** `master` @ `aac6ce82e`
**Deployment mode:** self-hosted, `authenticated` mode, embedded Postgres 18.1

**Root cause:** `server/src/services/companies.ts` → `remove()` deletes `heartbeat_runs` before `cost_events`, whose `heartbeat_run_id` FK is `ON DELETE NO ACTION`. Secondarily, its ~30-table list cannot cover the 145 `company_id`-bearing tables plus plugin schemas.

## What Changed

Adds `paperclipai company purge` (`cli/src/commands/company-purge.ts`), registered under the existing `company` command group. It talks to Postgres directly, using the same connection resolution as `db:backup`, because the HTTP path is the thing that is broken.

- **Discovers targets by introspection** — every `BASE TABLE` in every schema with a `company_id` column, so plugin schemas are covered automatically.
- **Handles tables with no `company_id`** (`decision_effect_executions`, `status_card_updates`, …) via a join back to the scoped parent, but only where the FK is `NO ACTION`/`RESTRICT`. `CASCADE` and `SET NULL` are left to Postgres — more correct and less destructive.
- **Discovers the delete order** — steps are topologically pre-sorted by FK depth, and anything that still hits a violation rolls back to a savepoint and is retried next pass. Ordering bugs of this class cannot be reintroduced by adding a table.
- **Preflight** refuses to run when a company outside the target set references data inside it, including non-`company_id` columns pointing at `companies` such as `company_skills.forked_from_company_id`. Related companies can be purged together as one set.
- **Docs** in `doc/CLI.md`, plus 10 unit tests over plan construction and ordering.

No change to server behavior or the existing `company delete` command.

## Verification

- `pnpm --filter paperclipai typecheck` — clean.
- 10 unit tests covering the exact `cost_events`/`heartbeat_runs` case, CASCADE/SET NULL exclusion, multi-level chains, cycles, self-references, and plugin schemas — all passing.
- End to end against a live instance: purged three companies totalling **833,577 rows across 58 tables**, then verified **zero orphaned rows across all 616 foreign keys** afterwards.
- Dry run verified as a true rehearsal: it executes every delete and rolls back, leaving row counts and company count unchanged.

## Risks

- **Direct database access.** This bypasses the API and its authorization. It is a host-side operator command, gated behind `--apply --yes --confirm`, with dry run as the default and an automatic backup before committing.
- **Irreversible when applied.** Mitigated by the pre-purge backup and by dry run being the default.
- **`--fk-indexes` takes `ACCESS EXCLUSIVE` locks** held for the whole transaction. It is opt-in and documented as requiring the service to be stopped; without it no exclusive locks are taken.
- **Preflight covers single-column foreign keys.** Multi-column FKs are not analyzed; none exist in the current schema.
- **New dependency `postgres`.** Already a transitive dependency via `@paperclipai/db`; promoted to a direct CLI dependency. Happy to drop it and route through `@paperclipai/db`'s drizzle client instead if you would rather not add it — the savepoint/retry loop is the only reason it uses the driver directly.

Lockfile intentionally not modified, per the contributor guidance about the refresh bot.

## Related finding (not addressed here)

315 of the schema's 616 foreign keys have no index on the child column. `heartbeat_runs` alone is referenced by 39 unindexed columns, three of them on `issue_comments`. This makes every cascading delete scan child tables per parent row — it slows far more than purges. Worth a separate change; happy to open an issue.

## Model Used

Claude Opus 5 (1M context), via Claude Code.

### DIFF
--- a/cli/package.json
+++ b/cli/package.json
@@ -55,7 +55,8 @@
     "dotenv": "^17.4.2",
     "commander": "^13.1.0",
     "embedded-postgres": "^18.1.0-beta.16",
-    "picocolors": "^1.1.1"
+    "picocolors": "^1.1.1",
+    "postgres": "^3.4.9"
   },
   "devDependencies": {
     "@types/node": "^22.20.1",
--- a/cli/src/__tests__/company-purge.test.ts
+++ b/cli/src/__tests__/company-purge.test.ts
@@ -1,0 +1,152 @@
+import { describe, expect, it } from "vitest";
+import {
+  buildSteps,
+  orderSteps,
+  type ForeignKey,
+  type PurgeStep,
+} from "../commands/company-purge.js";
+
+function fk(
+  child: string,
+  childColumn: string,
+  parent: string,
+  deleteRule: string,
+): ForeignKey {
+  const [childSchema, childTable] = child.split(".");
+  const [parentSchema, parentTable] = parent.split(".");
+  return {
+    name: `${childTable}_${childColumn}_fkey`,
+    childSchema: childSchema!,
+    childTable: childTable!,
+    childColumn,
+    parentSchema: parentSchema!,
+    parentTable: parentTable!,
+    parentColumn: "id",
+    deleteRule,
+  };
+}
+
+function scopedSet(tables: string[]): Set<string> {
+  return new Set(tables);
+}
+
+function toTables(tables: string[]): Array<{ schema: string; table: string }> {
+  return tables.map((key) => {
+    const [schema, table] = key.split(".");
+    return { schema: schema!, table: table! };
+  });
+}
+
+function order(steps: PurgeStep[]): string[] {
+  return steps.map((step) => step.tableKey);
+}
+
+describe("buildSteps", () => {
+  it("emits one company_id delete per company-scoped table", () => {
+    const steps = buildSteps(toTables(["public.issues"]), [], scopedSet(["public.issues"]));
+    expect(steps).toHaveLength(1);
+    expect(steps[0]!.sql).toContain('DELETE FROM "public"."issues"');
+    expect(steps[0]!.sql).toContain("company_id = ANY($1)");
+  });
+
+  it("join-deletes NO ACTION children that have no company_id of their own", () => {
+    const steps = buildSteps(
+      toTables(["public.issues"]),
+      [fk("public.decision_effect_executions", "issue_id", "public.issues", "a")],
+      scopedSet(["public.issues"]),
+    );
+    const joined = steps.find((s) => s.tableKey === "public.decision_effect_executions");
+    expect(joined).toBeDefined();
+    expect(joined!.sql).toContain('USING "public"."issues" pa');
+    expect(joined!.sql).toContain("pa.company_id = ANY($1)");
+  });
+
+  it("leaves CASCADE and SET NULL children to Postgres", () => {
+    // status_card_updates -> issues is SET NULL: deleting those rows would
+    // destroy records that merely mention the company rather than belong to it.
+    const steps = buildSteps(
+      toTables(["public.issues"]),
+      [
+        fk("public.status_card_updates", "issue_id", "public.issues", "n"),
+        fk("public.issue_documents", "issue_id", "public.issues", "c"),
+      ],
+      scopedSet(["public.issues"]),
+    );
+    expect(order(steps)).toEqual(["public.issues"]);
+  });
+
+  it("keys join-deletes against companies on id rather than company_id", () => {
+    const steps = buildSteps(
+      [],
+      [fk("public.cli_auth_challenges", "company_id", "public.companies", "a")],
+      scopedSet([]),
+    );
+    expect(steps).toHaveLength(1);
+    expect(steps[0]!.sql).toContain("pa.id = ANY($1)");
+  });
+
+  it("covers plugin schemas the same way as public", () => {
+    const steps = buildSteps(
+      toTables(["plugin_llm_wiki_8f50.wiki_pages"]),
+      [],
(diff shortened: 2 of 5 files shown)


nickallevato:
Thanks — the `company_skills.forked_from_company_id` finding was correct and is fixed in 36911332b (the commit the latest review reports as last reviewed, so I think the finding carried over from the previous pass).

The preflight no longer excludes `public.companies` as an FK parent. It now excludes only a row's *own* `company_id` column — which says which company owns the row, not which company depends on it — and checks every other column pointing at `companies`, keyed on `pa.id`:

```ts
!(isCompaniesParent(fk) && fk.childColumn === "company_id")
```

```ts
const parentFilter = isCompaniesParent(fk) ? "pa.id = ANY($1)" : "pa.company_id = ANY($1)";
```

On the current schema that covers exactly the two non-`company_id` columns referencing `companies`, both `ON DELETE SET NULL`:

- `company_skills.forked_from_company_id`
- `cli_auth_challenges.requested_company_id`

A surviving company holding either reference now aborts the purge in preflight rather than having the column silently nulled.

nickallevato:
Closing in favour of the existing server-side work — #6756 fixes the ordering at the source in `companies.remove()` (+3/-2), and #2097 addresses the coverage half. Adding a parallel CLI path on top of those is the wrong shape, and I should have searched the issue and PR list before building this rather than after.

Leaving the findings here so they are not lost:

**1. The ordering fix alone does not close the coverage gap.** 145 tables carry a `company_id`, and plugins create their own `plugin_<slug>_<hash>` schemas with foreign keys into core tables at install time. Any hand-maintained delete list will drift again — that is the pattern behind #2096, #7250, #3038 and #3577, and it is why ten separate issues exist for what looks like one bug. Deriving the table set from `information_schema` rather than a literal list is the part I would still suggest, whether or not it comes from this PR.

**2. Non-`company_id` references to `companies` need handling.** Two columns point at `companies` outside a row's own `company_id`, both `ON DELETE SET NULL`:

- `company_skills.forked_from_company_id`
- `cli_auth_challenges.requested_company_id`

Deleting a company that a *surviving* company forked a skill from silently nulls that provenance rather than failing. Worth a guard in whatever implementation lands. (Credit to the Greptile review on this PR for catching it.)

**3. Unrelated but more impactful: 315 of the schema's 616 foreign keys have no index on the child column.** `heartbeat_runs` alone is referenced by 39 unindexed columns, three of them on `issue_comments`. Because Postgres enforces `ON DELETE CASCADE`/`SET NULL` with a per-row trigger, deleting one parent row scans each unindexed child table once. Measured on a real instance, removing ~11k `heartbeat_runs` rows took over ten minutes; with those indexes present the same work finished in 39 seconds. This slows every cascading delete, not just company removal, and is probably worth its own issue.

For reference, the approach was verified end to end against a live instance — three companies, 833,577 rows across 58 tables, zero orphaned rows across all 616 foreign keys afterwards — so the catalog-derived method does work if anyone wants to pick it up. Happy to open a focused issue for (1) or (2) if that would be useful.

PR #11101, opened 2026-08-08
feat: make chat-style tasks the default experience

scotttong:
## Thinking Path

> - Paperclip is the open source app people use to manage AI agents for work
> - The task detail view is the main surface where a person and an agent work on an issue together
> - The redesigned chat-style task experience (conversational thread, updated composer, plan tab, live streaming tail) has shipped behind the `enableTaskChatRedesign` experimental flag
> - New users had to find and enable an experimental card to get the better experience, and the flag doubled the render paths we had to maintain
> - The chat-style experience is now stable and verified, so the gate no longer earns its cost
> - This pull request makes chat-style tasks the default experience and removes the experimental flag end to end
> - Some users still want the pre-chat task page, so the change also adds an inverted Classic Task Interface experimental toggle as a legacy escape hatch
> - The benefit is that every user gets the modern task experience out of the box, while anyone who prefers the classic view can opt back into it from Settings → Experimental

## Linked Issues or Issue Description

No public GitHub issue exists for this change. Description follows the feature request template.

**Problem or motivation**

The chat-style task experience is complete, but it is hidden behind the `enableTaskChatRedesign` experimental flag. A user who installs Paperclip gets the legacy task detail view by default. They must discover the experimental features card and turn the flag on to get the intended experience. Maintaining two task detail render paths also slows development and invites drift between them.

**Proposed solution**

Make the chat-style task experience the default for all users. Remove the `enableTaskChatRedesign` flag from the shared types, the server instance settings, and the UI. Remove the chat-style tasks card from the experimental features list. Port the streaming presentation work so the live tail is stable in the default path. Add a new inverted `enableClassicTaskInterface` experimental toggle (default off) that restores the pre-chat task detail page for users who prefer it.

**Alternatives considered**

Keep the `enableTaskChatRedesign` flag and default it to on. This was rejected because it keeps the dual render path alive with chat-style framed as the experiment. Instead, the polarity is inverted: chat-style is the unflagged default, and the classic view is the explicitly opt-in experimental path.

**Roadmap alignment**

This completes the "make tasks feel like chat" track. It does not duplicate planned core work in `ROADMAP.md`.

## What Changed

- Made the chat-style task detail view the only task detail experience.
- Removed the `enableTaskChatRedesign` flag from `packages/shared`, `server/src/services/instance-settings.ts`, and the UI.
- Removed the chat-style tasks card from the experimental features section.
- Updated composer: draft persistence, conversational layout, and plan tab in the default view.
[...]

PR #11370, opened 2026-08-14
fix(onboarding): do not restore wizard state for a company the user does not own

tonio-alucema:
## Thinking Path

> - A person signs in to a company, and every agent, goal and task they create belongs to that company
> - Onboarding saves a draft to `localStorage` so a customer who closes the tab can pick up where they left off
> - But `localStorage` is scoped per browser origin, not per account, so the draft outlives the session that wrote it
> - A second account signing in on the same browser is handed the first account's company id, and every write the wizard makes then targets a company that account does not own
> - This pull request discards a saved draft whose company the signed-in account does not own, and waits for the company list to settle before deciding
> - The benefit is that a shared browser cannot leak one account's company into another account's onboarding

Lands the work from #9900, which had gone stale against `master`. No public issue exists. The problem follows.

**What happened?**

Onboarding persists a draft to `localStorage` under `paperclip-onboarding-state`, including `createdCompanyId`. That key is per origin, not per account. A browser that has already run onboarding hands the stored company id to the next session on the same origin, whoever is signed in.

Every downstream call in the wizard then targets that company: goals, agents and issues are created there, and requests fail with authorization errors when the account does not own it.

**Expected behavior**

A saved draft is restored only when the signed-in account owns the company it names. Otherwise the draft is discarded and the stale blob is removed.

**Steps to reproduce**

1. Sign in as account A and run onboarding far enough to create a company.
2. Sign out, and sign in as account B in the same browser.
3. Open onboarding. It restores account A's `createdCompanyId`.

`master` at `d95340b0b`.

## What Changed

Author: @stubbi. Their two commits are unchanged, with their authorship. The third commit is mine and is the rebase adaptation described below.

- `ui/src/lib/onboarding-state.ts` — new. `restoreOnboardingState` returns the saved draft only when the account owns the company it names.
- `ui/src/lib/onboarding-state.test.ts` — new. Covers ownership, absent company id, and malformed input.
- `ui/src/components/OnboardingWizard.tsx` — splits the component into a gate and `OnboardingWizardInner`, and exports `ONBOARDING_STORAGE_KEY`.
- `ui/src/components/OnboardingWizard.test.tsx` — new. Covers the deferred mount and the discard.

### Why the split into a gate

The inner component has around twenty `useState(saved?.x ?? default)` initializers, and an initializer runs only on the first render. If the inner component mounts before the restored draft is final, every field locks to its default and the draft is lost with no way back.

`restoreOnboardingState` needs the settled company list: it cannot tell an empty-because-loading list from an empty-because-the-account-owns-nothing list, and treats both as "owns nothing". So when a saved blob exists, the gate waits for `companiesLoading` to clear before computing `saved` and mounting the inner component at all.

### Rebase notes

The branch was 475 commits behind and conflicted. Two conflicts, both cosmetic:

1. `OnboardingWizard.tsx` — the `export` on `ONBOARDING_STORAGE_KEY` sat in a hunk whose surrounding context (`DEFAULT_TASK_TITLE`, `DEFAULT_TASK_DESCRIPTION`) `master` had rewritten. Kept master's copy, took the `export`.
2. `OnboardingWizard.test.tsx` — asserted on `"Create your team lead"`, which #11101 renamed to `"Create your first agent"`. Only the user-visible string moved; the gate reaches step 3 with the restored draft either way, and the test's other two assertions passed unchanged. This is the one commit of mine.

### Interaction with #11352

#11352 landed earlier today and reworked much of `OnboardingWizard.tsx`. Everything it changed lives in what this patch turns into `OnboardingWizardInner`, so the two compose. I checked the two ways that could have gone wrong:

- `companiesLoading` is React Query's `isLoading`, which is true only on a first load with no data. A background refetch cannot flip it back and unmount the wizard mid-flow.
- The dashboard auto-open added by #11352 fires only after the agent list and the mission lookup have settled, both of which are keyed on a company id that comes from the company query. So it cannot open the wizard inside the gate's `null` window.

## Verification

- `ui` typecheck is clean.
- The two new suites pass, and so do the four onboarding suites already on `master`.
- Full `ui` suite: **3939 pass**.

**Not done:** no manual two-account run in a real browser. The behaviour is covered by the component tests instead.

## Risks

Low, and bounded by the gate's own condition: it defers mounting only when a saved blob exists. A session with no saved draft — the common case — reaches the wizard exactly as before.

The failure mode if the ownership check were wrong in the other direction is a discarded draft, which costs the customer their in-progress onboarding but nothing that was already written to the server. The check errs that way deliberately: an empty company list reads as "owns nothing", which is why the gate waits for the list to settle first.

## Model Used

Claude Opus 5 (`claude-opus-5`), through Claude Code, for the rebase, the conflict resolution and the verification. The two substantive commits are @stubbi's own work.

## Checklist

### DIFF
--- a/ui/src/components/OnboardingWizard.test.tsx
+++ b/ui/src/components/OnboardingWizard.test.tsx
@@ -7,8 +7,6 @@ import { afterEach, beforeEach, describe, expect, it, vi } from "vitest";
 
 // --- Mocks (hoisted so vi.mock factories can close over them) ----------------
 
-const ONBOARDING_STORAGE_KEY = "paperclip-onboarding-state";
-
 const mockDialog = vi.hoisted(() => ({
   onboardingOpen: true,
   onboardingOptions: {} as { initialStep?: number; companyId?: string },
@@ -110,7 +108,7 @@ vi.mock("./AsciiArtAnimation", () => ({ AsciiArtAnimation: () => null }));
 vi.mock("./FrontDoor", () => ({ FrontDoor: () => null }));
 vi.mock("./AgentCapsule", () => ({ AgentCapsule: () => null }));
 
-import { OnboardingWizard } from "./OnboardingWizard";
+import { ONBOARDING_STORAGE_KEY, OnboardingWizard } from "./OnboardingWizard";
 
 // eslint-disable-next-line @typescript-eslint/no-explicit-any
 (globalThis as any).IS_REACT_ACT_ENVIRONMENT = true;
--- a/ui/src/components/OnboardingWizard.test.tsx
+++ b/ui/src/components/OnboardingWizard.test.tsx
@@ -1,0 +1,262 @@
+// @vitest-environment jsdom
+
+import { act } from "react";
+import { createRoot } from "react-dom/client";
+import { QueryClient, QueryClientProvider } from "@tanstack/react-query";
+import { afterEach, beforeEach, describe, expect, it, vi } from "vitest";
+
+// --- Mocks (hoisted so vi.mock factories can close over them) ----------------
+
+const ONBOARDING_STORAGE_KEY = "paperclip-onboarding-state";
+
+const mockDialog = vi.hoisted(() => ({
+  onboardingOpen: true,
+  onboardingOptions: {} as { initialStep?: number; companyId?: string },
+  closeOnboarding: vi.fn(),
+  onboardingRouteDismissed: false,
+  setOnboardingRouteDismissed: vi.fn(),
+}));
+
+const mockCompany = vi.hoisted(() => ({
+  companies: [] as Array<{ id: string; name: string; issuePrefix: string }>,
+  setSelectedCompanyId: vi.fn(),
+  loading: false,
+}));
+
+const mockCompaniesApi = vi.hoisted(() => ({
+  create: vi.fn(),
+  update: vi.fn(),
+}));
+const mockGoalsApi = vi.hoisted(() => ({
+  create: vi.fn(),
+  list: vi.fn(async () => []),
+}));
+const mockAgentsApi = vi.hoisted(() => ({
+  adapterModels: vi.fn(async () => [] as Array<{ id: string; label: string }>),
+  testEnvironment: vi.fn(async () => ({
+    adapterType: "claude_local",
+    status: "pass" as const,
+    checks: [],
+    testedAt: new Date().toISOString(),
+  })),
+  hire: vi.fn(async () => ({ agent: { id: "agent-1" }, approval: null })),
+  instructionsBundle: vi.fn(async () => ({ entryFile: "AGENTS.md" })),
+  saveInstructionsFile: vi.fn(async () => ({})),
+}));
+const mockApprovalsApi = vi.hoisted(() => ({
+  create: vi.fn(),
+}));
+const mockIssuesApi = vi.hoisted(() => ({
+  create: vi.fn(),
+}));
+const mockProjectsApi = vi.hoisted(() => ({
+  create: vi.fn(),
+  list: vi.fn(async () => []),
+}));
+
+// The real adapter registry eagerly imports every adapter package. The
+// model/harness picker internals are out of scope here, so stub the adapter
+// layer entirely and drive it through this knob.
+const mockAdapterRegistry = vi.hoisted(() => ({
+  list: [] as Array<{ type: string }>,
+  disabled: new Set<string>(),
+}));
+
+vi.mock("@/lib/router", () => ({
+  useLocation: () => ({ pathname: "/", search: "", hash: "", state: null }),
+  useNavigate: () => vi.fn(),
+  useParams: () => ({}),
+}));
+vi.mock("../context/DialogContext", () => ({
+  useDialog: () => mockDialog,
+}));
+vi.mock("../context/CompanyContext", () => ({
+  useCompany: () => mockCompany,
+}));
+vi.mock("../api/companies", () => ({ companiesApi: mockCompaniesApi }));
+vi.mock("../api/goals", () => ({ goalsApi: mockGoalsApi }));
+vi.mock("../api/agents", () => ({ agentsApi: mockAgentsApi }));
+vi.mock("../api/approvals", () => ({ approvalsApi: mockApprovalsApi }));
+vi.mock("../api/issues", () => ({ issuesApi: mockIssuesApi }));
+vi.mock("../api/projects", () => ({ projectsApi: mockProjectsApi }));
+vi.mock("../adapters", () => ({
+  listUIAdapters: () => mockAdapterRegistry.list,
+  getUIAdapter: () => ({ buildAdapterConfig: () => ({}) }),
+}));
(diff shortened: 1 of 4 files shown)


tonio-alucema:
Both valid, both fixed. The first one is destructive, so worth spelling out.

**A failed company query deleted the draft.** React Query reports a failed list as `isLoading === false` with `data` defaulted to `[]`, which is indistinguishable from "settled, and this account owns nothing" — and that is exactly the verdict that wipes the blob. So a customer whose company request timed out lost their in-progress onboarding, permanently, with no way back. An errored list is now treated like a loading one: not decidable, nothing restored, nothing deleted.

Worth noting *why* the fix is "don't mount" rather than "mount with `saved: null` and skip the delete": the persist effect inside the inner component writes the wizard's state back on every change, so mounting with defaults would overwrite the draft within a render or two anyway. There is no mount-but-leave-storage-alone option. Declining to mount is the only thing that actually preserves it, and the comment now says so — that reasoning was implicit before and would not have survived the next edit.

**`localStorage.getItem` was outside the try.** Only the parse was guarded. `getItem` throws outright where a browser denies storage — Safari private mode, a blocked third-party context — and the gate runs during render, so the exception escaped and took the app down. This is a regression against the `loadSavedState()` it replaces, which wrapped the read too. Both are inside the try now.

Two cases added, each confirmed to fail against the code it replaces rather than trusted on a green run: a failed company query leaves the blob intact and mounts nothing, and a throwing `getItem` renders the wizard instead of the exception escaping.

`ui` typecheck clean, 67 pass across the seven onboarding suites.

For context on attribution: the two substantive commits here are @stubbi's from #9900. These two fixes are mine, in the gate their patch introduces — flagging that so the review history reads accurately.

tonio-alucema:
Valid — `reset()` clears the draft and `handleClose` calls it, so closing the wizard threw on a browser that denies storage.

That is three review rounds finding the same failure at a different call site, which means fixing them one at a time was the wrong shape. `onboardingDraftStorage` now owns the read, the write and the clear, and all four sites go through it. One place can throw; one place handles it.

**Worth recording what went wrong while I did that,** because it is the more useful finding. The bulk edit that introduced the helper also rewrote the body of the helper's own `clear()`, so it called itself — infinite recursion, with its own `catch` swallowing the stack overflow. The draft silently stopped being cleared, and nothing in the new tests noticed. What caught it was `discards a saved draft for a company the signed-in account does not own`, which has been in this branch since @stubbi wrote it and went red immediately.

That is the whole argument for running the suite rather than only the cases you just added — and, separately, for not trusting a mechanical find-and-replace across a file you are also adding a definition to.

One case added, confirmed to fail against the code it replaces: closing the wizard with storage denied calls `closeOnboarding` instead of throwing.

`ui` typecheck clean; full `ui` suite 3922 pass. Two failures, neither related to this branch and both passing in isolation: the timezone-dependent `IssueProperties` case present on `master`, and a timing-based leader-election case in `cross-tab-poll` that only fails under machine load.

tonio-alucema:
Valid, and finding it corrected reasoning I had asserted in the two commits before this — worth stating plainly.

I claimed declining to mount was the *only* way to protect the draft, because the persist effect would otherwise overwrite it. That is only true when the wizard is **open**: the persist effect is itself gated on `effectiveOnboardingOpen`, so a mounted-but-closed wizard writes nothing. Mounting was never what cost the draft, and I had used that wrong premise to justify the gate.

With it corrected, blocking on an errored list was pure cost. The companies query sets `retry: false`, so the error state sticks — and with no companies the dashboard renders a "Get Started" button wired to `openOnboarding`. A gate returning `null` made that button do **nothing at all**, for as long as the error lasted, which is a worse dead end than the report suggests.

The gate now waits only while the list is *loading*. An error with an empty list stays undecidable for the draft — nothing restored, nothing deleted — but the wizard mounts. Closed, the draft survives untouched for a later load that can verify it; open, the customer is onboarding right now, which supersedes it anyway.

I also renamed `companiesUndecided` to `ownershipUndecidable`, since conflating "can the draft be judged" with "should the wizard mount" is precisely what produced this.

**On the test.** My first version asserted the draft survived an open wizard, which is false for the reason above — it would have encoded the wrong belief as a passing assertion. It now asserts what actually holds in each case, with a second case for the closed wizard where the blob does survive intact. Both confirmed to fail against the code they replace.

`ui` typecheck clean; 69 pass across the six onboarding suites.

tonio-alucema:
CI caught two things, both interactions with #11371 (the adapter snap, which landed while this was open) and neither visible from either branch alone. Both were green separately.

**The defect is mine.** `useCompany()` types its error as `Error | null`, so I wrote `companiesError !== null`. But `undefined` reaches that line in practice — from any consumer that supplies the company context without an `error` key — and `undefined !== null` is true. So a perfectly healthy company load read as a failure, ownership became undecidable, and the customer's draft was withheld. The question is "is there an error", and only a truthy test asks it.

It surfaced through #11371's adapter test, whose company mock predates this branch and has no `error`. That is not a test artifact — it is exactly the shape of caller that would hit this in production.

**The reverse direction too.** #11371 added `useAdapterRegistryLoaded` to `use-disabled-adapters`, and this branch's test mocks that module. A module mock replaces the *whole* module, so an export the component reaches for and the mock omits is `undefined`, and calling it throws. Every test in the file failed at once, which is the tell.

Worth noting this is the argument for CI testing the merge rather than the branch: I ran the full suite locally on this branch before pushing and it was clean, because the branch didn't have #11371 in it yet. I've since merged `master` in here and reproduced both failures locally before fixing them.

`ui` typecheck clean; full `ui` suite **3952 pass**, with only the timezone-dependent `IssueProperties` failure present on `master`.

tonio-alucema:
This is the sharpest finding on the branch, and it's right: my own fix two rounds ago reopened the hole this PR exists to close.

**Confirmed the premise before acting.** `useSignOut` invalidates only `auth.session` and `health` — the companies cache is never touched, and `invalidateQueries` keeps serving the old data regardless. So after A signs out and B signs in, a failed companies refetch really does leave A's list in hand. My narrowing ("an error only counts when it left the list empty") then read that non-empty list as authoritative, found A's company id in it, and restored A's onboarding draft for B.

Any error now makes ownership undecidable, whatever the list contains. The mount gate is unchanged — still keyed on loading alone — so the dead end that narrowing was meant to fix stays fixed: the wizard opens, it just doesn't restore, and doesn't delete. The next successful load restores the draft.

Splitting "can the draft be judged" from "should the wizard mount" in an earlier round is what made this a one-line change instead of a re-litigation of both.

**The test for it first passed against the broken code.** I wrote the attack path end to end — A's draft, A's stale companies, B signed in, refetch failed — and it went green before I'd fixed anything, because I saved the draft at step 4, where the agent-name field doesn't render. The assertion couldn't observe a restore either way. It saves at step 3 now and fails as it should. That's the second time on this branch a test asserted on something not rendered; worth naming as a pattern rather than a one-off.

**The underlying cache lifetime is out of scope here and needs its own fix.** This change is defensive and protects onboarding only — any other consumer of `useCompany()` still sees the stale list after a failed post-sign-out refetch. I've filed that separately: sign-out should remove account-scoped cached data rather than invalidating two keys.

`ui` typecheck clean; 73 pass across the seven onboarding suites.

PR #11371, opened 2026-08-14
fix(ui): onboarding wizard keeps an invisible disabled adapter selected

tonio-alucema:
## Thinking Path

> - An agent runs on an adapter, and the deployer decides which adapters their instance offers
> - Onboarding asks a new customer to pick one, and defaults the choice to `claude_local`
> - But the adapter grid only renders adapters the server has enabled, while the default and any saved draft name an adapter directly
> - So on an instance where that adapter is disabled, the wizard holds a selection with nothing selected on screen, and hiring the first agent fails on an adapter the customer was never shown
> - This pull request snaps the selection to the first enabled adapter whenever the held one is not visible
> - The benefit is that the first agent is created on an adapter the instance actually offers, and the customer sees what they picked

Lands the work from #9501. No public issue exists. The problem follows.

**What happened?**

`adapterType` defaults to `claude_local`, and a saved draft can name any adapter. The grid renders `recommendedAdapters` and `moreAdapters`, both filtered by the disabled-adapter set the server supplies.

When the held adapter is not in either list, nothing in the grid appears selected, but the wizard still holds it. The first agent is then hired on an adapter the deployer disabled.

**Expected behavior**

The wizard holds an adapter the customer can see. When the held one is not visible, it snaps to the first enabled adapter.

**Steps to reproduce**

1. Disable `claude_local` on the instance.
2. Open onboarding and go to the adapter step.
3. No card is selected, and the wizard is still holding `claude_local`.

`master` at `d95340b0b`.

## What Changed

Author: @stubbi. Their commit is unchanged, with their authorship. The second commit is mine and is the rename described below.

- `ui/src/components/OnboardingWizard.tsx` — an effect that snaps `adapterType` to the first visible adapter when the held one is not among them.
- `ui/src/components/OnboardingWizard.adapters.test.tsx` — new. Covers a disabled default snapping, and an enabled saved value being left alone.

### Why the file was renamed

The suite arrived as `OnboardingWizard.test.tsx`, which is also the filename #9900 adds for its restore gate. Both branches merge cleanly onto `master` on their own, and neither status shows a problem, but they collide with each other:

```
CONFLICT (add/add): Merge conflict in ui/src/components/OnboardingWizard.test.tsx
```

Whichever landed first would have turned the other red. Nothing imports the file by name, so the rename is inert. It also describes the contents better: the suite is `describe("OnboardingWizard adapter selection")`, and `Component.aspect.test.tsx` is what the directory already does — `OnboardingWizard.step.test.tsx`, `AgentConfigForm.render.test.tsx`, `ProjectProperties.concurrency.test.tsx`.

### Interaction with #11352

#11352 landed earlier today and reworked much of `OnboardingWizard.tsx`. The snap effect applies to the current file unchanged, and its two tests pass against it. Verified before and after rebasing onto `master`.

## Verification

- `ui` typecheck is clean.
- 60 pass across the adapter, step, variant, route, mission-hook and offer suites.
- Full `ui` suite: **3941 pass**, measured with this branch, `master` and the #9900 branch combined, to confirm the collision is genuinely resolved rather than relocated.

## Risks

Low. The effect is a no-op whenever the held adapter is visible, which is every instance that leaves the default enabled.

The behaviour it changes is a selection the customer could not see, so there is no in-progress state to lose. The one case worth naming is an instance with every adapter disabled: the visible list is empty, the effect finds nothing to snap to, and the selection is left as it was. That is the same state as today.

## Model Used

Claude Opus 5 (`claude-opus-5`), through Claude Code, for the rename, the collision analysis and the verification. The substantive commit is @stubbi's own work.

## Checklist


tonio-alucema:
Valid. External adapter types are registered into the UI registry only once the adapters query resolves, so until then `listUIAdapters()` returns the built-ins alone and a saved external adapter is indistinguishable from one the deployer disabled. The snap fired on that transient list, replaced the customer's choice with a built-in, and the persist effect wrote it down — so a reload made it permanent.

`useAdapterRegistryLoaded` now gates it. It reads the same query key `useDisabledAdaptersSync` already uses, so it shares that cache entry and adds no request. I added a hook rather than widening that one's return type: it has five production callers and three test mocks, none of which have anything to do with this.

**Worth flagging that this gate fails closed, unlike the ones in #11352 and #11370.** I've been arguing fail-open for the onboarding gates, so the inconsistency is deliberate and the comment explains it: the directions of harm are opposite here. Acting on an unloaded registry silently rewrites a saved answer and persists it. Waiting forever just leaves the selection alone — which is exactly the behaviour that existed before the snap did, so the downside is the status quo. Refusing to change a saved answer is the cheaper mistake.

One case added, confirmed to fail against the code it replaces: a saved external adapter survives a render where the registry has not loaded.

`ui` typecheck clean; 51 pass across the adapter, hook, dialog, config-form and wizard-step suites. I ran the other callers of the hook too, since the module changed.

Attribution note: the substantive commit here is @stubbi's from #9501. The rename and this gate are mine.

PR #11379, opened 2026-08-14
fix(ui): theme the onboarding wizard decorative panel instead of hardcoding dark

tonio-alucema:
## Thinking Path

> - The app ships a light theme and a dark theme, and follows the operating system unless the customer picks one
> - Onboarding is the first screen a new customer sees, and its right-hand half is a decorative panel holding the ASCII paperclip illustration
> - That panel hardcoded a near-black surface, so it stayed dark whatever the theme said
> - In light mode the customer meets the product as a pale form sitting beside a solid black rectangle, on the one screen that is supposed to introduce it
> - This pull request paints the panel with the same muted surface tokens as the rest of the app
> - The benefit is that the first screen looks deliberate in both themes, and follows any future theme without another fix

Lands the work from #8982. No public issue exists. The problem follows.

**What happened?**

The onboarding wizard's decorative right-hand panel used a hardcoded dark background. `ThemeContext` supports `"light" | "dark"` and follows `prefers-color-scheme`, so in light mode the panel stayed near-black against a light form.

The illustration glyphs inside it already use `text-muted-foreground`, so the panel was the only part not following the theme.

**Expected behavior**

The panel uses the paired `bg-muted` surface, so the illustration reads as ink on a surface in either theme.

**Steps to reproduce**

1. Set the app to light mode.
2. Open onboarding and go to step 1 or 2, where the panel is visible.
3. The left half is light and the right half is near-black.

`master` at `a53cc8819`.

## What Changed

Author: @stubbi. Their two commits are unchanged, with their authorship. The third is mine and is the rebase adaptation below.

- `ui/src/components/OnboardingWizard.tsx` — the decorative panel uses `bg-muted text-muted-foreground` instead of a hardcoded surface. One line.
- `ui/src/components/OnboardingWizardTheme.test.tsx` — new. Asserts against the source, since the panel's className is a static literal and mounting the full wizard dialog to check a background is disproportionate.

### Rebase notes

One conflict, on the changed line itself. This branch replaced `bg-(--hex-1d1d1d)` with the muted tokens, while `master` had separately moved `transition-[width,opacity]` to `transition-(--tp-width-opacity)`. Kept master's transition and this branch's colours.

### One assertion had stopped guarding anything

The suite has two cases. The first checked `bg-[#1d1d1d]` and the bracket-hex form generally — but `master` migrated that class to the CSS-variable spelling `bg-(--hex-1d1d1d)` while this branch was open, and neither pattern matches it. So it passed whether or not the panel was themed.

It now matches both spellings. Found by reverting the fix on the rebased branch: the second case failed and the first stayed green.

## Verification

- `ui` typecheck is clean.
- Both theme cases pass, and both fail when the panel is reverted to a hardcoded surface — where before, only one did.
- `--muted` is defined in both themes in `index.css` (light `oklch(0.97 0 0)`, dark `oklch(0.269 0 0)`), so the panel follows the theme rather than staying near-black.
- Full `ui` suite: **3958 pass**.

**Not done:** no screenshot comparison. The change is one class on a decorative panel, and the token pair is asserted in the test.

## Risks

Very low. One className on a decorative element, with no behaviour attached and no other consumer.

Worth naming that the original motivation cited `PAPERCLIP_DEFAULT_THEME`, a deployer setting that does not exist in this codebase. The fix stands on its own here regardless: Paperclip ships a real light theme with an OS-following default, and the panel was ignoring it.

## Model Used

Claude Opus 5 (`claude-opus-5`), through Claude Code, for the rebase, the widened assertion and the verification. The two substantive commits are @stubbi's own work.

## Checklist

### DIFF
--- a/ui/src/components/OnboardingWizard.tsx
+++ b/ui/src/components/OnboardingWizard.tsx
@@ -2155,7 +2155,7 @@ function OnboardingWizardInner({
               name + mission steps) */}
           <div
             className={cn(
-              "hidden md:block overflow-hidden bg-(--hex-1d1d1d) transition-(--tp-width-opacity) duration-500 ease-in-out",
+              "hidden md:block overflow-hidden bg-muted text-muted-foreground transition-(--tp-width-opacity) duration-500 ease-in-out",
               step === 1 || step === 2 ? "w-1/2 opacity-100" : "w-0 opacity-0"
             )}
           >
--- a/ui/src/components/OnboardingWizardTheme.test.tsx
+++ b/ui/src/components/OnboardingWizardTheme.test.tsx
@@ -1,0 +1,34 @@
+// @vitest-environment node
+
+import { readFileSync } from "node:fs";
+import path from "node:path";
+import { fileURLToPath } from "node:url";
+import { describe, expect, it } from "vitest";
+
+// The onboarding wizard's decorative right-hand panel (which renders the
+// ASCII paperclip illustration) must follow the active shadcn theme instead of
+// hardcoding a dark surface. Otherwise a light/cream deployer theme (set via
+// the PAPERCLIP_DEFAULT_THEME bootstrap) renders a jarring cream form next to a
+// solid dark panel. The illustration glyphs already use `text-muted-foreground`,
+// so the panel must sit on the paired `bg-muted` surface to read as an
+// intentional ink-on-surface texture in every theme.
+//
+// Asserting against the source keeps this guard cheap: the panel className is a
+// static string literal, and the full wizard dialog is too heavy to mount here.
+const here = path.dirname(fileURLToPath(import.meta.url));
+
+describe("OnboardingWizard decorative panel theming", () => {
+  const source = readFileSync(
+    path.join(here, "OnboardingWizard.tsx"),
+    "utf8"
+  );
+
+  it("does not hardcode a dark decorative panel background", () => {
+    expect(source).not.toContain("bg-[#1d1d1d]");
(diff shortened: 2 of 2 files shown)


tonio-alucema:
Both taken, even though they're flagged non-blocking — they're the same failure the widened assertion in this PR was already fixing, one level up.

**Scoped to the panel, not the file.** A file-wide scan fails this case for a hardcoded colour anywhere else in the wizard, reporting a panel regression that hasn't happened. Both cases now anchor on the `<AsciiArtAnimation />` wrapper through one shared helper.

**Named what's allowed instead of what's forbidden.** Forbidding spellings is exactly what rotted here: the original checked `bg-[#rrggbb]`, and silently stopped guarding anything the moment master migrated the class to `bg-(--hex-1d1d1d)`. My widened version covered both hex forms — and would still have missed `bg-black`, as you point out. So the assertion is now the complete set of `bg-` classes on the panel, which is `["bg-muted"]`. That can't decay the same way.

Confirmed against four spellings of the regression:

| panel background | before this commit | now |
| --- | --- | --- |
| `bg-(--hex-1d1d1d)` | 1 of 2 cases fail | both fail |
| `bg-[#1d1d1d]` | 1 of 2 cases fail | both fail |
| `bg-black` | **neither fails** | both fail |
| `bg-zinc-900` | **neither fails** | both fail |

The two named colours are the ones your review caught — they would have sailed past every version of this guard including mine.

`ui` typecheck clean; the theme, restore-gate, adapter and step suites pass (31).

PR #11380, opened 2026-08-14
fix(ui): clear account-scoped query caches on sign-out

tonio-alucema:
## Thinking Path

> - Everything a person creates belongs to a company, and which companies a person can see is an authorization fact the server owns
> - The UI keeps that company list in a React Query cache entry, `["companies"]`, that carries no account identity
> - Sign-out is the moment that cache stops describing the person at the keyboard, but the self-hosted path only marked two unrelated keys stale and left the company list untouched
> - Invalidation also keeps serving the old value until a refetch succeeds, and the companies query sets `retry: false`, so one failed request pins the previous account's list in place
> - This pull request drops every account-scoped cache entry on sign-out, against a small allowlist of entries that describe the instance instead of the account
> - The benefit is that consumers reading the company list stop inheriting the previous account's answer

No public issue exists. Refs #11370, #11382. The problem follows.

**What happened?**

`useSignOut` invalidated only `queryKeys.auth.session` and `queryKeys.health` on the self-hosted path. Two things follow:

- The companies query (`queryKeys.companies.all`) was never touched, so the previous account's company list stayed in the cache across a sign-out.
- `invalidateQueries` marks data stale but keeps serving it, so even the two keys it did touch stayed readable until a refetch succeeded.

`CompanyProvider` sits above the router in `ui/src/main.tsx`, so it stays mounted across the whole sign-out and sign-in cycle and never gets a fresh mount to clear it. `companiesListQueryOptions` sets `retry: false`, so a single failed refetch sticks for the rest of the next account's session.

Consumers that would then act on the previous account's list:

- `resolveBootstrapCompanySelection` (`ui/src/context/CompanyContext.tsx`) auto-selects `selectableCompanies[0]`, putting account B on account A's company and rendering A's company name in the sidebar and breadcrumbs.
- `CloudAccessGate` gates on `boardAccessQuery.data.companyIds.length`, so a stale `access.currentBoardAccess` entry admits B on A's memberships.
- `InviteLanding` reads a cached list as proof of membership in two places.

**Expected behavior**

After a sign-out, no cache entry describing the previous account remains readable.

**Steps to reproduce**

1. On a self-hosted instance in `authenticated` mode, sign in as account A and let the company list load.
2. Sign out. The document does not reload.
3. Sign in as account B in the same tab, with the companies request failing once.
4. `useCompany()` still returns account A's companies.

`master` at `2a4b4bc63`.

## What Changed

- `ui/src/hooks/useSignOut.ts` — the self-hosted path resets every account-scoped cache entry instead of invalidating two keys. Adds `isAccountScopedQueryKey`, tested against an `INSTANCE_SCOPED_QUERY_ROOTS` allowlist that holds `health` alone.
- `ui/src/hooks/useSignOut.test.tsx` — regression coverage for sign-out followed by a failed refetch, plus cases for the allowlist, a failed sign-out, and the Cloud path.
- `ui/src/pages/InstanceGeneralSettings.test.tsx` — asserted the old mechanism (`invalidateQueries` called with the session key); now asserts the outcome.
- `ui/src/components/OnboardingWizard.tsx`, `ui/src/components/OnboardingWizard.test.tsx` — comments only. They stated that sign-out invalidates just two keys, which is no longer true.

### Why the allowlist is default-deny

A query key added to `queryKeys` later is treated as account-scoped unless someone deliberately adds it to the allowlist, so forgetting this file fails closed.

`health` is refreshed in place rather than dropped. It carries deployment mode, bootstrap state and Cloud metadata, nothing account-scoped, and `useCloudInstance` observes it with `enabled: false` because `CloudAccessGate` owns the fetch. Dropping the entry would strand those observers on `null` and flip Cloud instances into their self-hosted rendering mid-sign-out.

### Why `resetQueries` and not `removeQueries`

`removeQueries` empties the cache but does not notify observers already subscribed to those entries, because `#updateQuery` only runs on `setOptions` and `onSubscribe`. A probe against the installed query-core (5.101.4) confirms it: after the sweep `getQueryData` returns `undefined` while the mounted observer keeps returning the old data and never refetches. Since `CompanyProvider` never unmounts here, that is the common case, not a corner one.

A sibling change (#11382, and the invite-landing work) depends on `isFetchedAfterMount`, and warns that `resetQueries` rewinds the update counters that flag is derived from. That warning is correct, and I measured its reach:

| Observer | After the sweep |
| --- | --- |
| Binds while the cache is **warm**, survives the sweep | `isFetchedAfterMount: false` — stranded |
| Same, across two sweeps | `false` — stays stranded |
| Binds while the cache is **cold**, survives the sweep | `true` |
| Mounts fresh **after** the sweep | `true` |

It cannot reach the sign-out sweep. The sweep resets `auth.session` too, so `CloudAccessGate` renders `<Navigate to="/auth">` on the next render and unmounts every component holding such a gate; they remount fresh on re-entry, which is the safe row. The strand also fails closed — a false `isFetchedAfterMount` makes those gates withhold a membership verdict rather than grant one, the opposite of the leak this fixes.

### Cloud path

Unaffected. `navigateTopLevel` is `window.location.assign`, a real document load that builds a new `QueryClient`, and it returns before the clearing branch. A test pins that the Cloud path leaves the cache alone.

## Verification

- `pnpm vitest run` in `ui`: 3957 passed, 1 failed.
- `pnpm tsc -b`: clean.
- The two changed tests were confirmed to fail against the old implementation: session data survives, and the companies query is never touched (1 call instead of 2).
- The `removeQueries` and `isFetchedAfterMount` claims above were measured with a throwaway probe against query-core 5.101.4, not reasoned about. The probe was deleted after it answered.

The one failure is `IssueProperties.test.tsx`, "renders scheduled, retrying, due, overdue…", which expects `4:08 PM` and gets `9:08 AM`. It is timezone-dependent, it reproduces on `master` without these changes, and it is in a file this change does not touch.

**Not done:** no manual two-account run in a browser. The sign-out button renders only in `authenticated` deployment mode and the bug needs two accounts, so a local dev instance cannot exercise the path.

## Risks

Low. The change widens what a sign-out clears, and the failure direction is an extra refetch after signing in rather than data held past its account.

The one behavioral shift worth naming: queries that were previously left alone across a sign-out now refetch. Those requests are expected to 401 now that the session is gone, so the sweep is deliberately not awaited and the sign-out button does not sit pending while they fail.

**This does not close the class.** An account can change without the sign-out button: a session that lapses server-side, or a second account signing in on a warm tab. #11382 and the invite-landing work handle those at the point of use, and remain necessary.

## Model Used

Claude Opus 5 (`claude-opus-5`), through Claude Code. Extended thinking enabled. Tool use enabled: file read and edit, shell for typecheck, test runs, and the query-core probe.

## Checklist

### DIFF
--- a/ui/src/components/OnboardingWizard.test.tsx
+++ b/ui/src/components/OnboardingWizard.test.tsx
@@ -496,11 +496,15 @@ describe("OnboardingWizard restore-gate (stale localStorage across accounts)", (
   });
   it("does not hand one account's draft to the next when a refetch fails after a switch", async () => {
     // The attack path in full. Account A onboards and leaves a draft naming
-    // its company. A signs out — which does not clear the companies cache,
-    // because `useSignOut` invalidates only the session and health queries.
-    // B signs in and the companies refetch fails, so A's list is still in
+    // its company. The account then changes without this component's company
+    // cache being cleared, and the refetch fails, so A's list is still in
     // hand. A list that still contains A's company must not be read as proof
     // that B owns it.
+    //
+    // `useSignOut` now resets account-scoped caches, which closes the
+    // sign-out-button route into this state (see its own regression test). The
+    // gate is asserted here independently of that: it must hold for any route
+    // that leaves a stale list behind, not only the one that has been fixed.
     window.localStorage.setItem(
       ONBOARDING_STORAGE_KEY,
       JSON.stringify({
--- a/ui/src/components/OnboardingWizard.tsx
+++ b/ui/src/components/OnboardingWizard.tsx
@@ -205,10 +205,15 @@ export function OnboardingWizard() {
   // trusts "not loading, no error" finds the old company id in them and hands
   // one account's onboarding draft to the next.
   //
-  // Sign-out does not help: `useSignOut` never clears the company cache, and
-  // on the self-hosted path there is no document reload to clear it either.
-  // Even once that is fixed, an account change can skip the button entirely -
-  // a session lapsing server-side, a second account signing in on a warm tab.
+  // Sign-out is no longer the hole it was: `useSignOut` now resets every
+  // account-scoped cache entry rather than invalidating two of them, so the
+  // ordinary A-signs-out-then-B-signs-in path does not leave A's companies in
+  // hand.
+  //
+  // That covers the button, not the question. An account can change without
+  // it - a session lapsing server-side, a second account signing in on a warm
+  // tab, a caller supplying the company context from somewhere else - so this
+  // gate stays independent of that fix rather than deferring to it.
   //
   // So this asks for a list fetched *for this mount*, rather than reading the
   // one in hand. `isFetchedAfterMount` is the part that matters; `staleTime: 0`
--- a/ui/src/hooks/useSignOut.test.tsx
+++ b/ui/src/hooks/useSignOut.test.tsx
@@ -1,11 +1,12 @@
 // @vitest-environment jsdom
 
+import { act } from "react";
 import { flushSync } from "react-dom";
 import { createRoot } from "react-dom/client";
-import { QueryClient, QueryClientProvider } from "@tanstack/react-query";
+import { QueryClient, QueryClientProvider, useQuery } from "@tanstack/react-query";
 import { afterEach, beforeEach, describe, expect, it, vi } from "vitest";
 import { queryKeys } from "@/lib/queryKeys";
-import { useSignOut } from "./useSignOut";
+import { isAccountScopedQueryKey, useSignOut } from "./useSignOut";
 
 const mockAuthApi = vi.hoisted(() => ({ signOut: vi.fn() }));
 const mockNavigateTopLevel = vi.hoisted(() => vi.fn());
@@ -75,13 +76,22 @@ describe("useSignOut", () => {
     flushSync(() => root.unmount());
   });
 
-  it("keeps self-hosted sign-out pending until the local request finishes, then invalidates caches", async () => {
+  it("keeps self-hosted sign-out pending until the local request finishes, then clears account caches", async () => {
     let resolveSignOut: (() => void) | undefined;
     mockAuthApi.signOut.mockImplementation(() => new Promise<void>((resolve) => {
       resolveSignOut = resolve;
     }));
     queryClient.setQueryData(queryKeys.health, { status: "ok", deploymentMode: "authenticated" });
     queryClient.setQueryData(queryKeys.auth.session, { session: { id: "session-1" } });
+    queryClient.setQueryData(queryKeys.companies.all, {
+      companies: [{ id: "company-a", name: "Account A Co" }],
+      unauthorized: false,
+    });
+    queryClient.setQueryData(queryKeys.access.currentBoardAccess, {
+      isInstanceAdmin: false,
+      companyIds: ["company-a"],
+    });
+    queryClient.setQueryData(queryKeys.issues.list("company-a"), [{ id: "issue-a" }]);
     const onSignedOut = vi.fn();
     const root = renderHarness(onSignedOut);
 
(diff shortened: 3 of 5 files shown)


tonio-alucema:
### Measured: why this stays on `resetQueries`

#11382 notes that `resetQueries` keeps the query object and can leave `isFetchedAfterMount` false for existing observers, and flags that its own gate depends on that flag. That warning is correct, and it is worth being precise about how far it reaches, since this PR resets the same query key.

I probed both operations against the installed query-core (5.101.4) rather than reasoning about them.

**`removeQueries` does not notify already-subscribed observers.** After the sweep, `getQueryData(["companies"])` returns `undefined` while the mounted observer keeps returning the previous account's data and never refetches. That is the bug this PR exists to fix, so `removeQueries` is not an option here.

**The `isFetchedAfterMount` strand is real, and narrower than it looks.** The baseline the flag compares against is captured when the observer *binds* to the query, so only an observer that binds to an already-warm cache can be rewound below its own baseline:

| Observer | After the sweep |
| --- | --- |
| Binds **warm** (baseline 3), survives the sweep | `isFetchedAfterMount: false` — stranded |
| Same, across two sweeps | `false` — stays stranded |
| Binds **cold** (baseline 0), survives the sweep | `true` |
| Mounts fresh **after** the sweep | `true` |

Two reasons it cannot reach the sign-out sweep:

1. **The sweep unmounts its own victims.** It resets `auth.session` along with everything else, so `CloudAccessGate` (`CloudAccessGate.tsx:113`) renders `<Navigate to="/auth">` on the next render. Every component holding such a gate — `OnboardingWizard`, `AppsConnect.tsx:374`, `InviteLanding` — is inside that tree and unmounts, then remounts fresh on re-entry, which is the last row.
2. **The strand fails closed.** A false `isFetchedAfterMount` makes those gates *withhold* a membership verdict rather than grant one. Even in the single render between reset and redirect, it errs toward refusing — the opposite direction from the leak.

The caveat stays correct where #11382 and the invite-landing change apply it: a *local* reset under an observer that stays mounted. It just does not transfer to a sweep that takes the tree down with it.

The probe was a throwaway and has been deleted; the scenarios are reproducible from the table above with `QueryObserver` plus `setQueryData` to warm the entry before binding.

tonio-alucema:
Independent check of the `removeQueries` vs `resetQueries` measurement in https://github.com/paperclipai/paperclip/pull/11380#issuecomment-5300984911: reproduced, same result. Scratch `QueryObserver` harness against the installed query-core 5.101.4, one subscribed observer on `["auth","session"]`, session value flipped to `null` before the sweep.

| sweep | notifications | observer data after | cache entry |
| --- | --- | --- | --- |
| `removeQueries` | 0 | still `{user:{id:"user-A"}}` | gone |
| `resetQueries` | 3 | `null` | `null` |

So `resetQueries` is the right operation for this sweep, and the "prefer `removeQueries`" advice I had put in #11417 was wrong for the sign-out path. It holds only for a local reset under an observer that stays mounted, which is why #11417 and #11382 still avoid reset.

## One thing worth checking before this lands, because it can undo the sweep

The stale-account data is not only in the query cache. `api.get` coalesces identical in-flight GETs keyed on the request path alone (`coalescedGet` / `inflightGets` in `ui/src/api/client.ts`), and the account-scoped query functions pass no abort signal. A request issued under one session can therefore be joined by a caller that runs after the account changed, and be answered with the previous account's data.

`resetQueries` refetches active observers immediately. If a GET for a swept path was issued under the outgoing session and is still in flight when the sweep runs, that refetch joins it, and the previous account's response is written straight back into the cache that was just cleared.

This is measurable the same way the notification counts were. #11430 adds `detachInflightGet(path)` in `ui/src/api/client.ts` for exactly this case: it drops the in-flight entry so later callers issue a fresh request, while the original still settles for the callers that asked for it — no abort, because those callers still want their answer. If #11430 lands first, the sweep can call it for the paths it clears; if this PR lands first, #11430 rebases onto it.

Wording for a `useSignOut.ts` code comment recording why reset and not removal, if it is useful to carry into this PR. Not opened separately, since this one rewrites those lines.

```ts
// If this grows into a sweep of the account-scoped caches, clear them with
// `resetQueries`, not `removeQueries`. Removal does not notify observers
// that are already subscribed — nothing rebinds them until some other
// change forces a render — so the entry disappears while every mounted
// observer keeps serving the signed-out account's value, and the redirect
// in CloudAccessGate that fires on the session going empty never runs.
// Reset notifies, the refetch returns null, and the app leaves. Measured
// against query-core 5.101.4: removal produced 0 notifications and left
// the observer holding the old session; reset produced 3 and null.
//
// The opposite advice holds for a *local* reset under an observer that
// stays mounted (InviteLanding, the onboarding draft gate): reset rewinds
// the update counters `isFetchedAfterMount` is derived from while that
// observer keeps its bind-time baseline, so the flag can never read true
// again and the gate withholds forever. Those spots avoid reset. Sign-out
// is not one of them: it resets the session too, so those consumers
// redirect, unmount, and remount fresh.
```

<sub>Posted with Claude Code (Claude Opus 5). The measurements above were produced by running the harness, not by reading the library source.</sub>

tonio-alucema:
Rebased onto `master` (`b38d6ddb8`) and resolved the conflict — it was mine to cause. Four onboarding PRs landed in `OnboardingWizard.tsx` while this sat open: #11370, #11382, #11416 and #11429.

**The conflict was comment-only.** One hunk, both sides rewriting the same rationale block above `companiesQuery`. No logic on either side.

Your version had to win the factual half. `master` said *"Sign-out does not help: `useSignOut` never clears the company cache"*, which this PR makes false. What I committed keeps that correction and the "covers the button, not the question" distinction, and drops a restatement of the leak that the paragraph above already makes:

```
// Sign-out is no longer the hole it was: `useSignOut` now resets every
// account-scoped cache entry rather than invalidating two of them, so the
// ordinary A-signs-out-then-B-signs-in path does not leave A's companies in
// hand.
//
// That covers the button, not the question. An account can change without
// it - a session lapsing server-side, a second account signing in on a warm
// tab, a caller supplying the company context from somewhere else - so this
// gate stays independent of that fix rather than deferring to it.
```

Three escape paths kept from yours; `master`'s version named two.

Nothing else changed. `ui` typecheck clean, full `ui` suite **4014 pass** — the one failure, `IssueProperties`, is timezone-dependent and reproduces on `master`.

**One thing worth a second look before this merges**, from the session that did the `InviteLanding` sibling: it recommended `removeQueries` over `resetQueries`, on the grounds that reset keeps the query object and can leave existing observers reading `isFetchedAfterMount: false` afterward. Your reasoning for `resetQueries` — that `removeQueries` does not notify mounted observers, and `CompanyProvider` stays mounted across the whole cycle — is specific and looks right to me for that case.

The live consumer of that flag is `AppsConnect.tsx:374-375`, which sits behind the `apps/connect` route and so should unmount on sign-out, which would make the concern moot. I have not verified that unmount actually happens — I inferred it from the route table — so I would treat it as likely fine rather than settled. Note the onboarding gate no longer reads `isFetchedAfterMount` at all; #11382 removed it in favour of `isSuccess`.

PR #11382, opened 2026-08-14
fix(onboarding): verify draft ownership against a list fetched this session

tonio-alucema:
## Thinking Path

> - Everything a person creates belongs to a company, and which companies they can see is an authorization fact the server owns
> - Onboarding saves a draft to `localStorage`, which is scoped per browser origin rather than per account, so #11370 made the wizard check that the signed-in account owns the company a saved draft names
> - That check reads the shared company list, and the shared list is a cache: for thirty seconds after a sign-in it is served without a request, and an invalidation keeps serving the old data while it refetches
> - So the previous account's companies arrive looking healthy — no loading, no error — and a check that trusts them restores that account's draft for whoever signed in next
> - This pull request judges ownership against a list fetched for the current mount instead of the one in hand
> - The benefit is that an authorization check stops depending on cache freshness, which nothing in the app currently guarantees

Corrects #11370, which I merged earlier today. No public issue exists. The problem follows.

**What happened?**

#11370 added an ownership check so a saved onboarding draft is restored only when the signed-in account owns the company it names. It judged ownership from `useCompany()`, treating the list as authoritative whenever it was not loading and had no error.

The company list is a cache, and neither condition detects a stale one:

- `main.tsx:41` sets `staleTime: 30_000` for all queries, so a mounted observer is served the cached list for thirty seconds after a sign-in with no request at all.
- `Auth.tsx:58` invalidates the list on sign-in, but `invalidateQueries` keeps serving the old data while it refetches, and `isLoading` stays false throughout.

`useSignOut` never clears that cache, and the self-hosted path does not reload the document, so a sign-out leaves it warm.

**Expected behavior**

A draft is restored only when a company list fetched for the current session says the account owns it.

**Steps to reproduce**

1. On a self-hosted instance, sign in as account A and run onboarding far enough to create a company. A draft is saved naming it.
2. Sign out. The page does not reload and the company cache is not cleared.
3. Sign in as account B in the same tab.
4. Onboarding restores account A's draft.

`master` at `35a9b987`.

## What Changed

- `ui/src/components/OnboardingWizard.tsx` — the restore gate fetches the company list itself, with `staleTime: 0`, and decides only once `isFetchedAfterMount` is true. It no longer reads `useCompany()`.
- `ui/src/components/OnboardingWizard.test.tsx` — ownership cases now drive the fetch rather than the context; two new cases.
- `ui/src/components/OnboardingWizard.adapters.test.tsx` — stubs `companiesApi` now that the gate fetches.

### Why fetching, rather than a better staleness check

There is no flag on the cached value that says which account it belongs to. `isFetchedAfterMount` is the only signal that ties the data to this mount, and `staleTime: 0` is what makes such a fetch happen while the shared entry is still fresh. It uses the same query key, so it populates the same cache entry rather than adding a parallel one, and it is `enabled` only when a saved draft exists — without one there is nothing to authorize and no request is made.

### An auth blip no longer deletes a draft

`companiesListQueryOptions` folds 401 and 403 into `{ companies: [], unauthorized: true }` rather than throwing. Those arrived as a successful fetch of an empty list, which reads as "this account owns nothing" — and that verdict deletes the blob. `unauthorized` is now undecidable: nothing restored, nothing deleted.

### What is unchanged

The mount gate still holds only while the fetch is in flight, never on failure, so the dead end fixed in #11370 stays fixed: with no companies the dashboard's "Get Started" button still opens a working wizard.

## Verification

- `ui` typecheck is clean.
- Full `ui` suite: **3960 pass**.
- Every case was checked against the specific clause it guards, not just run green.

| Case | Fails without |
| --- | --- |
| Warm cache holding another account's list, no error | the `isFetchedAfterMount` check |
| Unauthorized list | the `unauthorized` check |
| Unowned draft | the ownership check itself |
| Failed fetch | the decidability gate |

The warm-cache case first passed against the code it was meant to catch: a pending fetch left the wizard unmounted, so "the draft was not restored" held for the wrong reason. It now resolves the fetch with a different account's companies and asserts against a mounted wizard.

**Not done:** no manual two-account run in a browser.

## Risks

Low, and lower than what it replaces. The failure direction is a draft withheld until the next load, which is recoverable; the direction it removes is a draft handed to the wrong account, which is not.

It adds one request, on mounts where a saved draft exists, to a query key the app already uses.

**This does not close the class.** The stale cache is still there for every other consumer — `CompanyContext`'s auto-select effect (`CompanyContext.tsx:89-111`) writes a company id to `localStorage` from whatever list is cached, with the same exposure. Sign-out cleanup is being handled separately, and cannot close it either: an account change can skip the button entirely, via a server-side session lapse or a second account signing in on a warm tab.

Credit where due: found by a sibling session working on the same cache from `InviteLanding`. Its warning that `resetQueries` keeps the query object — and can leave `isFetchedAfterMount` false for existing observers — matters directly here, since this fix depends on that flag.

Revert the commit to restore.

## Model Used

## Checklist

### DIFF
--- a/ui/src/components/OnboardingWizard.adapters.test.tsx
+++ b/ui/src/components/OnboardingWizard.adapters.test.tsx
@@ -43,6 +43,14 @@ vi.mock("../context/DialogContext", () => ({
 vi.mock("../context/CompanyContext", () => ({
   useCompany: () => mockCompany,
 }));
+// The restore gate fetches the company list itself to verify draft ownership,
+// rather than trusting the shared cache, so this module now needs stubbing
+// here. An empty list is fine: the drafts in this file carry no
+// `createdCompanyId`, so there is no ownership question — but the fetch has to
+// succeed for the gate to treat the draft as decidable at all.
+vi.mock("../api/companies", () => ({
+  companiesApi: { create: vi.fn(), list: vi.fn().mockResolvedValue([]) },
+}));
 vi.mock("../adapters", () => ({
   listUIAdapters: () => mockAdapterRegistry.list,
   getUIAdapter: () => ({ buildAdapterConfig: () => ({}) }),
--- a/ui/src/components/OnboardingWizard.test.tsx
+++ b/ui/src/components/OnboardingWizard.test.tsx
@@ -118,6 +118,7 @@ vi.mock("./FrontDoor", () => ({ FrontDoor: () => null }));
 vi.mock("./AgentCapsule", () => ({ AgentCapsule: () => null }));
 
 import { ApiError } from "../api/client";
+import { queryKeys } from "../lib/queryKeys";
 import { ONBOARDING_STORAGE_KEY, OnboardingWizard } from "./OnboardingWizard";
 
 // eslint-disable-next-line @typescript-eslint/no-explicit-any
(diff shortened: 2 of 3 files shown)


tonio-alucema:
All three valid, and the first one I had already convinced myself was handled. It wasn't, and how I got that wrong is the most useful thing here.

**Failed refetches were accepted.** You're right: `isFetchedAfterMount` does not reject a refetch that failed — React Query counts the failed attempt as a fetch after mount while retaining the previous `data`. So after an account switch the retained value is the previous account's list, presented as freshly fetched. `isSuccess` is what actually ties the answer to this session, because a failed refetch flips status to error.

I checked this claim before replying, removed `isFetchedAfterMount`, watched a test fail, and concluded the guard was working. **That test was failing for a different reason** — the inner wizard mounted *during* the fetch and locked its `useState(saved?.x ?? default)` initializers to defaults, so the draft could not appear whatever the gate decided. Fixing the mount gate (below) is what exposed the real behaviour. A fault injection is only evidence if the failure it produces is the one you're claiming.

`isFetchedAfterMount` is now removed rather than kept next to `isSuccess`. It rejects nothing `isSuccess` hasn't already rejected, and no test could distinguish it — which is how a guard rots into decoration.

**Mounting undecided.** Also right, and worse than a discarded restoration: `isLoading` is false whenever retained data exists, so a refetch over a warm cache mounted the wizard while ownership was undecidable, and with the wizard open the persist effect overwrote the customer's *own* draft with defaults before the answer arrived. Now keyed on `isFetching`, which covers refetches. It still releases on failure, so the "Get Started" dead end stays fixed.

**Malformed cleanup.** Fixed — an unreadable draft is judged before the ownership check and doesn't enable the query at all.

Three cases added, and I then re-checked every clause in isolation rather than trusting the suite:

| removed | result |
| --- | --- |
| `isSuccess` | 1 case fails |
| `unauthorized` check | 1 case fails |
| `isFetching` (→ `isLoading`) | 2 cases fail |
| malformed-first ordering | 1 case fails |
| `enabled` narrowing | 1 case fails |

`ui` typecheck clean; full `ui` suite 3963 pass, with only the timezone-dependent `IssueProperties` failure present on `master`.

PR #11392, opened 2026-08-14
feat(plugin-sdk): concurrent event dispatch and host-call faults in the test harness

ericdfields:
## Thinking Path

> - Plugins extend Paperclip. The plugin SDK gives authors `createTestHarness()` to test a plugin without a live host.
> - The production event path is concurrent. `plugin-event-bus.ts` fans out to subscribers with `Promise.all`. `worker-rpc-host.ts` dispatches each `onEvent` handler without awaiting it. Two events interleave at every `await` inside a handler.
> - The harness `emit` awaited each handler in a loop. The harness was more serialized than production.
> - A read-modify-write race against `ctx.state` therefore failed in production but passed in the harness. This is the worst failure mode for a test tool, because it gives false confidence about the bug class that is hardest to find by inspection.
> - The harness also had no way to make a host call fail. `TestHarnessOptions` accepted only `manifest`, `capabilities` and `config`. The only failure lever was to omit a capability.
> - This pull request makes the harness able to reproduce concurrent delivery, and adds host-call fault injection.
> - The benefit is that plugin authors can write a test for a lost update or a partial failure, and that test now fails when the bug is present.

Refs #11393

That issue covers the design gap in full: plugin event delivery is concurrent, but the SDK has no mutual-exclusion primitive. This pull request fixes the test-harness half of it, so the resulting race is at least reproducible in a test. It does not add the primitive itself, so it does not close the issue.

I searched the repository for `createTestHarness`, `test harness concurrent`, `plugin state compare-and-swap` and `plugin mutex lock`, and found no duplicate or in-flight pull requests.

The inline description below follows `.github/ISSUE_TEMPLATE/enhancement.yml`.

**What is the current behavior?**

`createTestHarness().emit(...)` dispatches to matching event handlers in a `for` loop and awaits each handler before it starts the next. Handlers cannot interleave. `TestHarnessOptions` accepts only `manifest`, `capabilities` and `config`, and the harness `ctx` is a fixed object literal with no stub hooks. A test cannot make `issues.update` or `state.set` throw.

**What is the desired behavior?**

The harness must be able to deliver events the way the host does, so that a racy handler fails the test. The harness must also be able to fail an individual host call, so that a test can cover partial-failure paths.

**Why is this valuable?**

Plugin event delivery is concurrent and unserialized end to end:

- `server/src/services/plugin-event-bus.ts` calls subscribers concurrently and awaits them with `Promise.all`.
- `packages/plugins/sdk/src/worker-rpc-host.ts` dispatches `onEvent` without awaiting the returned promise.

A plugin that reads `ctx.state`, awaits, and then writes a value derived from the stale read has a lost-update bug. The harness could not show that bug. The new test file demonstrates the difference directly: the same handler produces `2` under sequential dispatch and `1` under concurrent dispatch.

## What Changed

- Added `TestHarnessEventDispatch` (`"sequential" | "concurrent"`) and the `TestHarnessOptions.eventDispatch` option.
- Added an `options` parameter to `emit(...)` for a per-call `dispatch` override.
- Added `emitConcurrent([...])`. It delivers several events whose handlers interleave with each other.
- Added `TestHarnessOptions.hostCallInterceptor`, plus the `TestHarnessHostCall`, `TestHarnessHostCallInterceptor` and `TestHarnessParkHandle` types. The interceptor wraps any `ctx.<namespace>.<method>(...)` call. It can throw before the effect, throw after the effect, or park the call.
- Added `harness.hostCalls`, an ordered log of every host call, with `path`, `args` and `seq`.
- Added `harness.faults` with `throwOn`, `throwAfterOn`, `parkOn` and `clear`. This is an ergonomic layer over the interceptor.
- Extracted `buildEvent()` and `matchingHandlers()` from the old `emit` body. Both `emit` and `emitConcurrent` use them, so the matching rules stay in one place.
- The `ctx` interception uses a recursive `Proxy`. It covers namespaces at any depth and needs no hand-written list of namespaces. Wrappers are memoized, so `ctx.issues === ctx.issues` and `ctx.issues.update === ctx.issues.update` stay true.

The default dispatch stays `"sequential"`. Existing tests keep their meaning.

## Verification

```
pnpm --filter @paperclipai/plugin-sdk typecheck     # passes
cd packages/plugins/sdk && npx vitest run           # 7 files, 55 tests, all pass
```

The new file is `packages/plugins/sdk/tests/testing-concurrency.test.ts`. Its first two tests are the point of the change. One asserts that the default sequential dispatch hides a lost update. The other asserts that `emitConcurrent` reproduces it.

I also rebuilt the SDK and ran the downstream users of `createTestHarness`, to confirm the proxied `ctx` is transparent:

```
pnpm --filter @paperclipai/plugin-sdk build
pnpm --filter @paperclipai/plugin-llm-wiki test        # 101 tests pass
pnpm --filter @paperclipai/plugin-workspace-diff test  # 26 tests pass
cd server && npx vitest run src/__tests__/plugin-sdk-testing.test.ts                  # 8 tests pass
cd server && npx vitest run src/__tests__/plugin-sdk-orchestration-contract.test.ts   # 5 tests pass
```

## Risks

Low risk. The change is limited to `packages/plugins/sdk`. It touches test tooling only, and it adds no production code path.

The one behavior change for existing callers is that `harness.ctx` is now a `Proxy` instead of the raw object literal. Three properties of the proxy keep this safe:

- A method call returns the underlying value unchanged when no fault and no interceptor apply. Synchronous methods such as `ctx.events.on` still return synchronously.
- Plain data such as `ctx.manifest` passes through by reference and is not wrapped.
- Wrappers are memoized against the value they wrapped, so reference equality holds and a reassigned method is re-wrapped rather than served stale.

All new options default to the old behavior. A caller that passes no new option gets the previous result.

## Model Used

Claude Opus 5 (`claude-opus-5`), running in Claude Code with extended thinking and tool use. The model read the relevant host and SDK sources to confirm the concurrency claims, designed the API, and wrote the implementation and tests. A human reviewed the diff, found and fixed the proxy reference-stability defect, and ran the verification commands above.

## Checklist

### DIFF
--- a/doc/plugins/PLUGIN_SPEC.md
+++ b/doc/plugins/PLUGIN_SPEC.md
@@ -1649,6 +1649,8 @@ The test harness provides:
 - ability to simulate `executeTool` calls as if coming from an agent run
 - in-memory state and entity stores for assertions
 - configurable capability sets for testing capability denial paths
+- concurrent event delivery that matches the host event bus, so that read-modify-write races against `ctx.state` reproduce in tests instead of being hidden by a strictly serialized harness
+- host-call fault injection, so that a test can make one `ctx` call throw before its effect, throw after its effect, or park while another call proceeds
 
 Example usage:
 
--- a/packages/plugins/sdk/src/testing.ts
+++ b/packages/plugins/sdk/src/testing.ts
@@ -73,6 +73,39 @@ import type {
   PluginPerformActionContext,
 } from "./protocol.js";
 
+/**
+ * How `emit` dispatches to matching handlers. See `TestHarnessOptions.eventDispatch`.
+ */
+export type TestHarnessEventDispatch = "sequential" | "concurrent";
+
+/** One host call intercepted through `ctx`. See `TestHarnessOptions.hostCallInterceptor`. */
+export interface TestHarnessHostCall {
+  /** Dotted path of the intercepted host call, e.g. `"issues.update"`. */
+  path: string;
+  args: unknown[];
+  /** Monotonic sequence number across every host call on this harness. */
+  seq: number;
+}
+
+/**
+ * Wraps a host call. Call `proceed()` to run the real in-memory implementation.
+ * Throwing without calling it models a host call that failed before its effect
+ * landed; awaiting it and then throwing models an effect that landed but was
+ * never acknowledged.
+ */
+export type TestHarnessHostCallInterceptor = (
+  call: TestHarnessHostCall,
+  proceed: () => Promise<unknown>,
+) => Promise<unknown>;
+
+/** Handle returned by `harness.faults.parkOn(...)`. */
+export interface TestHarnessParkHandle {
+  /** Resolves once the parked call has actually been entered. */
+  reached: Promise<void>;
+  /** Let the parked call proceed. */
+  release: () => void;
+}
+
 export interface TestHarnessOptions {
   /** Plugin manifest used to seed capability checks and metadata. */
   manifest: PaperclipPluginManifestV1;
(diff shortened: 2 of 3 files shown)


ericdfields:
Thanks — the P1 is valid, and it is now fixed in fa887c410.

You are right about the cause. A configured `hostCallInterceptor` routed **every** `ctx` method through the promise-based fault pipeline, so `ctx.events.on` returned a `Promise` instead of its unsubscribe callback. Plugin cleanup code that calls the returned value would have failed.

The fix restricts the pipeline to `async` methods. I checked every non-`async` method on the harness `ctx`, and they are all genuinely synchronous registration or observation APIs rather than host round trips: `events.on`, the five `register` methods, `streams.open/emit/close`, `logger.*`, `localFolders.declarations`. None of them returns a promise, so none of them loses anything by not being interceptable. They are still recorded in `hostCalls`; they are simply never faulted. I documented that on `TestHarnessOptions.hostCallInterceptor`.

Added a regression test for exactly your case: with an interceptor configured, `ctx.events.on` still returns a callable function synchronously, the interceptor never sees `events.on`, and an async call on the same harness is still intercepted.

Verification after the fix: SDK typecheck clean, 56/56 SDK tests, and the downstream harness consumers still pass — `plugin-llm-wiki` 101/101, `plugin-workspace-diff` 26/26, and the two server `plugin-sdk` suites 13/13.

PR #10993, opened 2026-08-06
fix(heartbeat): reconcile orphaned run agent state

namorbru-rgb:
## Thinking Path

> - Paperclip is the open source app people use to manage AI agents for work.
> - The heartbeat service owns run lifecycle and the agent status shown to operators.
> - An execution lease can end while its run row still has a live status.
> - A timed-out run can also leave a stale agent error while a real follow-up run is active.
> - Recovery must close orphaned runs without changing a different live run.
> - An issue comment is evidence, but it is not an authenticated run result.
> - This pull request adds the reconciliation changes on top of #10955.
> - The benefit is a terminal run, a correct agent state, and safe stale-lock recovery.

## Linked Issues or Issue Description

Related: #10955 and #10985.

**What happened?**

A released execution lease could leave a heartbeat run in `running` or `queued`. Recovery could also leave the agent in `running` with an old timeout in `errorReason`. A terminal issue and an active issue could reference the same run, which made a recovery-only terminal transition unsafe.

**Expected behavior**

Lease release must leave its run terminal. Recovery must preserve a run that an active issue still owns. A real follow-up execution must clear the previous run's stale agent error. A late issue comment must not rewrite an already terminal run outcome.

**Steps to reproduce**

1. Start a heartbeat execution.
2. End its execution lease before the normal terminal run update completes.
3. Start a follow-up run after the first run timed out.
4. Observe the stale live run or stale agent timeout state.

**Paperclip version or commit**

Base branch `fix/terminalize-run-on-lease-release` at `3520d47f`.

**Deployment mode**

Self-hosted server.

## What Changed

- Terminalize both `running` and `queued` runs with one compare-and-set when a lease ends.
- Scope terminal issue lookup to the run company.
- Clear a stale agent error when a real follow-up execution starts.
- Reconcile an orphaned run's agent to `idle` only when no other run is active.
- Preserve a run that any active issue still references.
- Keep stale-lock recovery moving after an audit insert fails.
- Add focused regression coverage for each lifecycle and recovery case.

## Verification

- `corepack pnpm --filter @paperclipai/server exec vitest run src/__tests__/heartbeat-retry-scheduling.test.ts src/__tests__/heartbeat-run-lease-release-terminalization.test.ts src/__tests__/recovery-stale-issue-lock-sweep.test.ts --config vitest.config.ts`
- Result: 3 test files passed. 47 tests passed.
[...]

PR #11394, opened 2026-08-14
fix(recovery): publish and queue the run status when the backstop terminalizes a run

nickallevato:
## Thinking Path

> - The recovery service runs a stale-lock sweep. Its backstop terminalizes a run that stays "running" after its issue is terminal or after its process is gone
> - Every other terminal write goes through heartbeat's `setRunStatus`, which publishes `heartbeat.run.status` and, since #13075, queues a durable status delivery with `executionStatusDeliveryId`
> - The backstop writes `heartbeat_runs` directly. It publishes no status event and queues no delivery, so no client learns that the run ended
> - The browser keeps the run as live until a refetch, and a plugin session consumer never gets its "done" event for the run
> - This pull request makes the backstop queue the durable delivery and publish the same status payload that `setRunStatus` publishes
> - The benefit is that a run that recovery ends is announced like every other terminal transition

Refs #13631. That issue describes the larger problem: the recovery backstop ends a live run, and then the owner finalizer skips its bookkeeping (agent status, wakeup request, result, usage). This pull request is only the event and outbox piece of #13631. The agent-status and wakeup consolidation is out of scope here.

Related public context:

- Refs #13075: added the durable status outbox (`executionStatusDeliveryId` and `deliverExecutionStatuses`) and moved `buildHeartbeatRunStatusLiveEventPayload` to the leaf module `heartbeat-run-status-payload.ts`.
- Refs #13419: the `process_gone` authority of the same backstop races the adapter finalizer. Same code path, different defect.
- Refs #13607: `claimed` wake requests are not reconciled when their run is terminal. Related bookkeeping gap after a backstop terminalization.
- Refs #10993: reconcile orphaned run agent state. Related open pull request on the agent-status side.
- Refs #12946: agent-status latch on lease release. Related open pull request on the agent-status side.

The UI half of the original pull request is split into #13634. That PR patches the scoped live-run cache entries (the dashboard "Agents" panel). This pull request now contains only the server change.

**What happened?**

`terminalizeOrphanedRunningRun` in `server/src/services/recovery/service.ts` sets a terminal status on `heartbeat_runs` with a direct `db.update(...)`. It does not go through `setRunStatus`. So it publishes no `heartbeat.run.status` event and it does not set `executionStatusDeliveryId`. The durable status delivery sweep therefore has nothing to deliver for this run. Live-run views keep the run as "Live now" until a remount or a window-focus refetch, while the task list already shows the issue as done.

**Expected behavior**

A run that the recovery backstop terminalizes is announced like any other terminal transition: an immediate `heartbeat.run.status` event with the full payload (including `finalText`, `issueId` and `contextSource`), and a durable delivery that the status delivery sweep retries if the immediate publish is lost.

**Steps to reproduce**

1. Start an agent run on an issue.
2. Make the issue reach `done` while the run row stays `running` (for example, a reuse-lease release that stops the sandbox but keeps the server process).
3. Wait for the stale-lock sweep. It terminalizes the run to `succeeded`.
4. See that no `heartbeat.run.status` event is published for the run, and `executionStatusDeliveryId` stays null.

Reproduced on `master`.

**Deployment mode**

`local_trusted`. The defect is in a shared server recovery path, so it is not specific to a deployment mode.

**Agent adapter(s) involved**

Any. The defect does not depend on the adapter.

## What Changed

- `server/src/services/recovery/service.ts`: the backstop's compare-and-set update now sets `executionStatusDeliveryId: randomUUID()`, as `setRunStatus` does. The status delivery sweep then delivers the terminal status even if the immediate publish does not reach a client.
- `server/src/services/recovery/service.ts`: after a successful terminal write, the backstop publishes `heartbeat.run.status` with `buildHeartbeatRunStatusLiveEventPayload(updated)`, as `setRunStatus` does. The payload carries `finalText`, `issueId` and `contextSource`. The earlier revision of this pull request built the payload inline to avoid an import cycle with `heartbeat.ts`. Since #13075 the builder is in a leaf module, so the cycle does not exist and the inline copy is gone.
- A publish failure is caught and logged, so it cannot stop the sweep before it clears the stale lock. The queued delivery retries the event.
- `server/src/__tests__/recovery-stale-issue-lock-sweep.test.ts`: the reuse-lease case now captures live events. It asserts one `heartbeat.run.status` event for the run with `status`, `agentId`, `issueId`, `finishedAt` and `finalText` (from the stored `resultJson` summary), and it asserts that `executionStatusDeliveryId` is set. The run now has an issue id in its context, so the telemetry mock also gets a `hashPrivateRef` stub.

Out of scope: `setRunStatus` also emits the plugin run-lifecycle event, and #13631 describes agent-status, wakeup and result bookkeeping that the owner finalizer skips after a backstop terminalization. These need a larger change and are tracked in #13631.

## Verification

Automated:

```
cd server && npx vitest run src/__tests__/recovery-stale-issue-lock-sweep.test.ts   # 20 passed
cd server && npx vitest run src/__tests__/heartbeat-process-recovery.test.ts \
  src/__tests__/server-startup-feedback-export.test.ts                             # 314 passed
pnpm typecheck                                                                    # passed
```

The new assertions fail without the `service.ts` change (`executionStatusDeliveryId` stays null and no status event is published) and pass with it.

## Risks

- Low risk. The change adds one column write in an existing update and one in-process event publish. It does not change which runs the backstop terminalizes.
- A client can receive the status twice: once from the immediate publish and once from the status delivery sweep. This is the same at-least-once behavior that `setRunStatus` already has, and clients handle status events idempotently by run id.

## Model Used

- Claude Opus 5 (claude-opus-5), 1M context window, via Claude Code with tool use (code editing, shell, test execution)

## Checklist

### DIFF
--- a/server/src/__tests__/recovery-stale-issue-lock-sweep.test.ts
+++ b/server/src/__tests__/recovery-stale-issue-lock-sweep.test.ts
@@ -20,6 +20,18 @@ import {
 const mockTelemetryClient = vi.hoisted(() => ({ track: vi.fn() }));
 vi.mock("../telemetry.ts", () => ({ getTelemetryClient: () => mockTelemetryClient }));
 
+// Capture live events so the terminalization payload can be asserted.
+const publishedLiveEvents = vi.hoisted(() => [] as Array<Record<string, unknown>>);
+vi.mock("../services/live-events.ts", async (importOriginal) => {
+  const actual = await importOriginal<typeof import("../services/live-events.ts")>();
+  return {
+    ...actual,
+    publishLiveEvent: (input: Record<string, unknown>) => {
+      publishedLiveEvents.push(input);
+    },
+  };
+});
+
 import { heartbeatService } from "../services/heartbeat.ts";
 
 const embeddedPostgresSupport = await getEmbeddedPostgresTestSupport();
@@ -285,9 +297,13 @@ describeEmbeddedPostgres("recovery sweepStaleIssueLocks", () => {
     // catches it: the issue reached "done" while the run row stayed "running".
     const { companyId, agentId, runningRunId } = await seed();
     // process.pid is the live test process, so isPidAlive returns true.
+    // resultJson carries the run summary, which becomes finalText on the event.
     await db
       .update(heartbeatRuns)
-      .set({ processPid: process.pid })
+      .set({
+        processPid: process.pid,
+        resultJson: { summary: "Reused sandbox summary for the operator." },
+      })
       .where(eq(heartbeatRuns.id, runningRunId));
     const issueId = randomUUID();
     await db.insert(issues).values({
@@ -331,6 +347,22 @@ describeEmbeddedPostgres("recovery sweepStaleIssueLocks", () => {
       .where(eq(heartbeatRunEvents.runId, runningRunId))
       .then((rows) => rows[0]);
     expect(event?.message).toContain("issue reached a terminal status");
+
+    // This path writes heartbeat_runs directly instead of going through
+    // setRunStatus, so it must announce the transition itself. Without the
+    // event no live-run view learns the run ended.
+    const statusEvent = publishedLiveEvents.find(
+      (e) => e.type === "heartbeat.run.status"
+        && (e.payload as Record<string, unknown>)?.runId === runningRunId,
+    );
+    expect(statusEvent).toBeDefined();
+    const payload = statusEvent!.payload as Record<string, unknown>;
+    expect(payload.status).toBe("succeeded");
+    // finalText must be present on a terminal payload: the plugin session
+    // consumer uses it as the message of the "done" event it emits, so a
+    // missing field turns a successful run into a completion with no result.
+    expect(payload).toHaveProperty("finalText");
+    expect(payload.finalText).toBe("Reused sandbox summary for the operator.");
   });
 
   it("terminalizes a running run to cancelled when its issue is cancelled (reuse-lease path)", async () => {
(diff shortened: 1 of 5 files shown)


nickallevato:
Thanks — the `finalText` finding is correct, and it is now fixed in f2db0129f.

The plugin session consumer (`plugin-host-services.ts`) uses `payload.finalText` as the message of the `done` event it emits for a succeeded run, so the omission turned a recovery-terminalized success into a completion with a null message.

`buildHeartbeatRunIssueComment` lives in `heartbeat-run-summary.ts`, which is a leaf module with no imports. It can therefore be used here without the import cycle that ruled out reusing `buildHeartbeatRunStatusLiveEventPayload` from `heartbeat.ts`. The status published on this path is always terminal, so `finalText` is always computed.

Covered by an assertion in the reuse-lease case of `recovery-stale-issue-lock-sweep.test.ts`. That assertion fails without the fix (`expected { …(9) } to have property "finalText"`) and passes with it; the suite is 11/11.

nickallevato:
The failing `General tests (server (5/12))` job is not related to this change. The failed test is `heartbeat-stale-queue-invalidation.test.ts` > "cancels a resolved connection-intent wake 'reassigned' after checkout but before adapter dispatch". The log shows a Postgres deadlock and dropped connections in that shard. This PR changes only the recovery backstop and its test. The test passes 3 of 3 times locally on this branch. `ci / verify` failed only because it aggregates the test result. A maintainer re-run of the failed job should clear it.

Issue #11391, opened 2026-08-14
Agent instructions fail open silently on read error, and managed instructions root is persisted as a stale absolute path


dzianisv:
Two related defects in managed agent-instructions loading let a broken instructions config run silently, with no board-visible signal. Found while investigating runaway heartbeat churn in a downstream company that had 7/8 agents running with no role instructions for an unknown period after a data-dir move.

Verified against the published package tree (`@paperclipai/adapter-pi-local`, `@paperclipai/server`), 2026-08-14.

## 1. Instructions read failure fails open silently

`packages/adapters/pi-local` — `server/execute.ts` (compiled `dist/server/execute.js:453-469`):

```js
try {
  const instructionsContents = await fs.readFile(resolvedInstructionsFilePath, "utf8");
  systemPromptExtension = `${instructionsContents}\n\n...`;
} catch (err) {
  instructionsReadFailed = true;
  await onLog("stdout", `[paperclip] Warning: could not read agent instructions file ...`);
  systemPromptExtension = promptTemplate; // generic template, run proceeds normally
}
```

`instructionsReadFailed` is only used to append an internal `commandNotes` line (dist line 512) — it is never surfaced on the run result or the agent record. The only signal is one stdout line inside a run log. The agent keeps heartbeating and billing indefinitely with the generic prompt template instead of its configured role instructions.

Same read-catch-fallback shape exists in every `adapter-*-local` package (confirmed via `grep -l instructionsFilePath */dist/server/execute.js`): claude, codex, cursor-cloud, cursor-local, gemini, grok, opencode, pi.

**Suggested fix:** surface `instructionsReadFailed` on the run result and the agent record (e.g. `errorReason`) so it's board-visible, and stop scheduling further heartbeats after N consecutive failed reads (recommend 2, to tolerate one transient FS blip) instead of silently degrading forever.

## 2. Managed instructions root persisted as a stale absolute path

`packages/server` — `services/agent-instructions.ts` (compiled `dist/services/agent-instructions.js`):

- `resolveManagedInstructionsRoot(agent)` (line 83) resolves against the *current* Paperclip instance root every time it's called.
- `applyBundleConfig()` (line ~284) persists the *resolved absolute path* into `agent.adapterConfig.instructionsFilePath` / `instructionsRootPath` at config-write time.
[...]

PR #11396, opened 2026-08-14
fix(adapters): surface instructions-file read failures instead of degrading silently

dzianisv:
## Thinking Path

> - The adapter layer builds each run's prompt. It injects the agent's configured instructions file (`adapterConfig.instructionsFilePath`) into the system prompt.
> - Every local adapter read that file inline, caught the error, wrote one warning line to the run log, and used the generic prompt template instead.
> - The failure never reached the run result or the agent record. An agent with a moved or mistyped instructions path kept heartbeating and billing forever with no role instructions, and the board saw a healthy agent.
> - This pull request moves the read into one helper in `adapter-utils`, returns a structured failure, and reports that failure on `AdapterExecutionResult`.
> - The server now counts consecutive failures for the same path, writes the reason to `agents.errorReason`, and pauses the agent after two failures in a row.
> - The benefit is that a broken instructions config is visible on the board in at most two heartbeats, and it stops burning tokens instead of degrading silently.

Refs #11391 (part 1 of that report; part 2, the stale absolute instructions root, is not addressed here)

## What Changed

- Add `readAdapterInstructionsFile()` and `instructionsReadFailureCommandNote()` in `packages/adapter-utils/src/agent-instructions-file.ts`. The helper does the read, logs one warning line, and returns a structured `failure` (`path`, `reason`, `code`).
- Add `AdapterInstructionsReadFailure` and the optional `AdapterExecutionResult.instructionsReadFailure` field to the adapter contract.
- Use the helper and report the failure in the claude-local, codex-local, cursor-cloud, cursor-local, gemini-local, grok-local, hermes, opencode-local and pi-local adapters, and in the shared acpx engine. This removes nine copies of the same read-catch-fallback block.
- Add `server/src/services/agent-instructions-health.ts`. It counts consecutive failures for the same path in `agents.metadata.instructionsReadHealth`, writes `agents.errorReason`, and pauses the agent (`status: paused`, `pauseReason: system`) at two consecutive failures.
- Call it from `heartbeatService` after run finalization, publish an `agent.status` live event on pause or recovery, and copy the failure into the run's `resultJson`.
- A later successful read clears the metadata and the error reason, and lifts the pause only when Paperclip applied it.
- Document the contract for adapter authors in `docs/adapters/creating-an-adapter.md`.

Grok-local also changed behavior slightly: it used to `copyFile` the configured path during asset staging, which threw an opaque error when the file was missing. It now probes with the helper first and continues without staged instructions, like the other adapters.

## Verification

```sh
npx vitest run packages/adapter-utils/src/agent-instructions-file.test.ts server/src/__tests__/agent-instructions-health.test.ts
# Test Files  2 passed (2)
# Tests  15 passed (15)
```

- `packages/adapter-utils/src/agent-instructions-file.test.ts` writes and removes real files in a temp dir. It asserts a structured `failure` (not only a log line) for a missing file, an `EACCES` file, a relative path resolved against `cwd`, and the no-instructions no-op.
- `server/src/__tests__/agent-instructions-health.test.ts` covers threshold, path change, recovery, manual-pause safety, and metadata preservation. It also runs against an embedded Postgres database: it seeds an agent with an unreadable instructions path, applies two failed runs, and asserts the stored agent row has `errorReason` after run 1 and `status: paused` with `pauseReason: system` after run 2. A third, successful run clears both.

Typecheck (all clean):

```sh
pnpm --filter @paperclipai/adapter-utils typecheck
# plus typecheck for each touched adapter package
```

Package suites for `packages/adapter-utils` and `packages/adapters` were run. Three files fail on this macOS host both with and without this branch (verified by stashing the change): `local-process-sandbox.test.ts` (bubblewrap, Linux only), `mcp-isolation.integration.test.ts`, and `workspace-restore-merge.test.ts` (unix socket path length). Nothing in the failures touches instructions loading. The full repo suite and `pnpm build` were not run locally; CI covers them.

## Risks

- Low to medium. The new pause path can stop heartbeats for an agent whose instructions file is genuinely unreadable. That is the intent, and it needs two consecutive failures on the same path, so one transient filesystem error does not pause anything.
- The pause uses `pauseReason: "system"`. Recovery only lifts a pause that matches that reason, so a manual pause is never cleared by this code.
- No schema migration. The state lives under an `agents.metadata` key and unrelated metadata keys are preserved.
- Adapters that do not set `instructionsReadFailure` keep the old behavior; the field is optional.

## Model Used

- Provider: GitHub Copilot API
- Model: Anthropic Claude Opus 5 (`claude-opus-5`), extended thinking at high reasoning level, agentic tool use (file edit, shell, tests) in the pi coding agent

## Checklist

- [x] All Paperclip CI gates are green — verified with `gh pr checks 11396`: every required build/test gate passes; only the Greptile gate (tracked separately below) is red.
- [ ] Greptile is 5/5 with no open P2s, recommendations, or follow-ups — currently 4/5. Open finding: a concurrent-pause race in the resume path (`server/src/services/agents.ts`) can let a newer pause get overwritten. Not yet fixed; tracked as follow-up branch work, not a description issue.

### DIFF
--- a/docs/adapters/creating-an-adapter.md
+++ b/docs/adapters/creating-an-adapter.md
@@ -240,6 +240,26 @@ With these flags set, the Paperclip UI will automatically show the instructions 
 
 If capability flags are not set, the server falls back to legacy hardcoded lists for built-in adapter types. External adapters that omit the flags will default to `false` for all capabilities.
 
+## Instructions File Reads
+
+Do not read the configured instructions file yourself. Use `readAdapterInstructionsFile` from `@paperclipai/adapter-utils`:
+
+```ts
+const instructionsFile = await readAdapterInstructionsFile({
+  instructionsFilePath: config.instructionsFilePath,
+  cwd, // omit when your adapter uses the configured path as-is
+  onLog,
+});
+```
+
+The helper reads the file, logs one warning line when the read fails, and returns a structured `failure`. Your adapter must put that value on every `AdapterExecutionResult` it returns:
+
+```ts
+return { ...result, instructionsReadFailure: instructionsFile.failure };
+```
+
+A failed read is not a run failure. The run continues on the generic prompt template, because one miss can be a transient filesystem error. The server records the failure on the run's `resultJson.instructionsReadFailure` so it is visible on the run record instead of only a stdout line. If your adapter drops `instructionsReadFailure`, a mistyped or moved instructions path degrades the agent silently and forever.
+
 ## Skills Injection
 
 Make Paperclip skills discoverable to your agent runtime without writing to the agent's working directory:
--- a/packages/adapter-utils/src/acpx-engine/execute.ts
+++ b/packages/adapter-utils/src/acpx-engine/execute.ts
@@ -10,8 +10,13 @@ import type {
   AdapterBillingType,
   AdapterExecutionContext,
   AdapterExecutionResult,
+  AdapterInstructionsReadFailure,
   UsageSummary,
 } from "@paperclipai/adapter-utils";
+import {
+  instructionsReadFailureCommandNote,
+  readAdapterInstructionsFile,
+} from "@paperclipai/adapter-utils/agent-instructions-file";
 import {
   adapterExecutionTargetSessionIdentity,
   describeAdapterExecutionTarget,
@@ -2357,32 +2362,31 @@ async function buildPrompt(ctx: AdapterExecutionContext, resumedSession: boolean
   prompt: string;
   promptMetrics: Record<string, number>;
   commandNotes: string[];
+  instructionsReadFailure: AdapterInstructionsReadFailure | null;
 }> {
   const { agent, runId, config, context, onLog } = ctx;
   const promptTemplate = asString(config.promptTemplate, DEFAULT_PAPERCLIP_AGENT_PROMPT_TEMPLATE);
-  const instructionsFilePath = asString(config.instructionsFilePath, "").trim();
-  const instructionsDir = instructionsFilePath ? `${path.dirname(instructionsFilePath)}/` : "";
+  const instructionsFile = await readAdapterInstructionsFile({
+    instructionsFilePath: config.instructionsFilePath,
+    onLog,
+    logStream: "stderr",
+  });
+  const instructionsFilePath = instructionsFile.resolvedPath;
+  const instructionsDir = instructionsFile.directory;
+  const instructionsReadFailure = instructionsFile.failure;
   let instructionsPrefix = "";
   const commandNotes: string[] = [];
-  if (instructionsFilePath) {
-    try {
-      const instructionsContents = await fs.readFile(instructionsFilePath, "utf8");
-      instructionsPrefix =
-        `${instructionsContents}\n\n` +
-        `The above agent instructions were loaded from ${instructionsFilePath}. ` +
-        `Resolve any relative file references from ${instructionsDir}.\n\n`;
-      commandNotes.push(
-        `Loaded agent instructions from ${instructionsFilePath}`,
-        `Prepended instructions + path directive to the ACPX prompt (relative references from ${instructionsDir}).`,
-      );
-    } catch (err) {
-      const reason = err instanceof Error ? err.message : String(err);
-      await onLog(
-        "stderr",
-        `[paperclip] Warning: could not read agent instructions file "${instructionsFilePath}": ${reason}\n`,
-      );
-      commandNotes.push(`Configured instructionsFilePath ${instructionsFilePath}, but file could not be read.`);
-    }
+  if (instructionsFile.contents !== null) {
+    instructionsPrefix =
+      `${instructionsFile.contents}\n\n` +
+      `The above agent instructions were loaded from ${instructionsFilePath}. ` +
+      `Resolve any relative file references from ${instructionsDir}.\n\n`;
+    commandNotes.push(
+      `Loaded agent instructions from ${instructionsFilePath}`,
+      `Prepended instructions + path directive to the ACPX prompt (relative references from ${instructionsDir}).`,
+    );
+  } else if (instructionsReadFailure) {
+    commandNotes.push(instructionsReadFailureCommandNote(instructionsReadFailure));
   }
 
   const bootstrapPromptTemplate = asString(config.bootstrapPromptTemplate, "");
@@ -2426,6 +2430,7 @@ async function buildPrompt(ctx: AdapterExecutionContext, resumedSession: boolean
   return {
     prompt,
     commandNotes,
+    instructionsReadFailure,
     promptMetrics: {
       promptChars: prompt.length,
       instructionsChars: promptInstructionsPrefix.length,
@@ -3815,6 +3820,7 @@ export function createAcpxEngineExecutor(deps: AcpxEngineExecutorOptions = {}) {
       // active turn, and `turnFinalize` reads all three. `activeTurn` is the run-
       // scoped hoisted local, so the settlement `endSession` step can cancel it.
       let runPrompt = "";
+      let instructionsReadFailure: AdapterInstructionsReadFailure | null = null;
       let preTurnStatus: AcpRuntimeStatus | null = null;
       // Phase-timing markers for the prepare_turn and turn phases. The prepare
       // phase covers the prompt build and the pre-turn usage snapshot; the turn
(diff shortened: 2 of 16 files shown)


dzianisv:
Both Greptile P1s were valid. Fixed in b7b6d8b.

**1. Recovery required a heartbeat that a paused agent never gets.** Correct — pausing removed the only path that could observe recovery. Recovery no longer depends on a run:

- Changing the agent's configured instructions path (`adapterConfig.instructionsFilePath`) clears the failure state and lifts the pause, but only when Paperclip applied that pause (`pauseReason: "system"`). This is the "operator corrected the path" case.
- `agentService.resume()` now also drops the failure state, so a resumed agent gets a fresh two-strike window instead of being paused again by its very next run. This is the "operator restored the file" case.
- A later run that reads the file still clears the state, as before.
- The text in `agents.errorReason` now names both actions instead of implying it recovers on its own.

New tests: `lifts the pause without a run when the operator corrects the instructions path` and `gives a resumed agent a fresh grace window`.

**2. Concurrent completions could corrupt health state.** Correct — it was an unlocked read-modify-write on `agents.metadata`. `applyInstructionsReadHealth` now runs inside a transaction and takes `SELECT ... FOR UPDATE` on the agent row, then evaluates from the locked snapshot. The caller in `heartbeat.ts` passes an `agentId` instead of a pre-read row, so no stale snapshot can reach the write.

New test: `counts concurrent run completions once each instead of losing one` runs two completions with `Promise.all` against embedded Postgres and asserts `consecutiveFailures: 2` plus a paused agent.

Also added: the service refuses to pause a `terminated` or `pending_approval` agent, and `clearInstructionsReadHealth` leaves a manual pause in place.

Verification:

```
npx vitest run server/src/__tests__/agent-instructions-health.test.ts
# Test Files  1 passed (1)
# Tests  17 passed (17)

pnpm --filter @paperclipai/server typecheck   # clean
```

dzianisv:
Third P1 (resume overwrites newer health state) was valid too. Fixed in fc602d0.

`applyInstructionsReadHealth` locks the agent row, but `agentService.resume` did not — it computed the cleared metadata from the snapshot it read before the write, so a resume that started first could commit stale `agents.metadata` over a failure recorded in the meantime.

`resume` now runs in a transaction and takes `SELECT ... FOR UPDATE` on the agent row before computing the cleared metadata, so the two writers serialize on the same lock. The transaction is entered with the same `transaction`-capability check `updateAgent` already uses, so a caller holding a tx-bound `db` still works.

New test `does not overwrite a concurrent health write when an operator resumes`: a concurrent writer takes the row lock and stores a marker in `agents.metadata`, the resume is started while that lock is held, then the writer commits. The resume must preserve the marker.

Red/green, against embedded Postgres:

```
# with the fix
npx vitest run server/src/__tests__/agent-instructions-health.test.ts
# Test Files  1 passed (1)
# Tests  18 passed (18)

# with resume reverted to the unlocked version
# AssertionError: expected undefined to be 'run-3'
# Tests  1 failed | 17 passed (18)

pnpm --filter @paperclipai/server typecheck   # exit 0
```

dzianisv:
Re-raised P1 (`resume` erases newer health state when the run completion wins the lock) is fixed in 868f0fb.

The lock made the two writers serialize but did not decide what should happen when the failure commits first. Resume was still deleting a failure the operator never saw, undoing its pause.

`resume` now compares the health state it read before taking the lock with the state it sees under the lock:

- **Unchanged** — the operator saw everything there is to see, so the state is cleared and the agent gets a full fresh window, as before.
- **Changed** — a failure landed in between. It is rewritten as strike one of the fresh window (`consecutiveFailures: 1`, `pausedAt: null`, `errorReason` kept) instead of being deleted.

So the operator's resume still wins — the agent is never left paused with no way back, which was the first P1 in this review — but the newer failure is not lost: a still-broken instructions path pauses the agent again on the very next run rather than the one after it.

New test `keeps a failure that lands after the operator asked to resume as strike one`: a concurrent writer holds the agent row lock and commits the pausing second failure while the resume is waiting on that lock; the agent must come back `idle` with `consecutiveFailures: 1` and `lastRunId: "run-2"`, and must pause again on the next failed read.

```
npx vitest run server/src/__tests__/agent-instructions-health.test.ts
# Test Files  1 passed (1)
# Tests  19 passed (19)

npx vitest run server/src/__tests__/agents-service-clear-error.test.ts \
  server/src/__tests__/budgets-service.test.ts \
  server/src/__tests__/agent-instructions-service.test.ts
# Test Files  3 passed (3)
# Tests  18 passed (18)

pnpm --filter @paperclipai/server typecheck   # exit 0
```

dzianisv:
P1 "old runs restore cleared health" was valid. Fixed in 305eee6.

Correcting the path cleared the state, but a run that started before the correction still read the old file and its completion re-created that state — two of them re-paused a freshly corrected agent for a path it no longer uses.

The failure now carries the configured path it came from (`AdapterInstructionsReadFailure.configuredPath`, populated centrally in `readAdapterInstructionsFile`, so every adapter gets it). `evaluateInstructionsReadHealth` compares it with the agent's current `adapterConfig.instructionsFilePath` read under the same row lock, and drops failures from a path the agent no longer uses. `configuredPath` is optional: a failure without it still counts, so an adapter built against the older contract degrades to the previous behaviour instead of going silent again — which is the bug this PR exists to fix.

New tests (all three fail without the guard):

- `ignores a failure reported by a run that used a superseded instructions path`
- `still counts failures from adapters that do not report the configured path`
- `does not re-pause a corrected agent from runs still using the old path` — embedded Postgres: record a failure, correct the path, land two stale completions, assert the agent stays idle with no health state, then assert a failure on the *new* path counts as strike one.

```
npx vitest run server/src/__tests__/agent-instructions-health.test.ts \
  packages/adapter-utils/src/agent-instructions-file.test.ts
# Test Files  2 passed (2)
# Tests  28 passed (28)

# with the guard removed
# Tests  2 failed | 20 passed (22)

pnpm --filter @paperclipai/server typecheck        # exit 0
pnpm --filter @paperclipai/adapter-utils typecheck # exit 0
```

dzianisv:
P1 "resume overwrites newer terminal status" was valid. Fixed in 3b113bc.

`resume` checked `terminated` / `pending_approval` against the snapshot it took before the lock, then wrote `status: "idle"` unconditionally — so a terminate committed while the resume waited on the row lock was silently undone.

The locked read now selects `status` as well and re-runs both guards inside the transaction. A concurrent terminate or pending-approval transition makes the resume fail with the same conflict it would have returned had it seen that state up front, and nothing is written.

New test `refuses to resume an agent terminated while the resume was in flight`: a transaction holds the agent row lock, terminates the agent, and commits while the resume is waiting; the resume must reject and the agent must stay `terminated`. It fails without the re-check.

```
npx vitest run server/src/__tests__/agent-instructions-health.test.ts \
  server/src/__tests__/agents-service-clear-error.test.ts
# Test Files  2 passed (2)
# Tests  26 passed (26)

# with the re-check removed
# Tests  1 failed | 22 passed (23)

pnpm --filter @paperclipai/server typecheck   # exit 0
```

For context on the check status: the Greptile check reports failure because this repo requires 5/5 and the last review scored 4/5 — every other check on the head commit is green.

dzianisv:
P1 "stale update clears newer health" was valid. Fixed in 4924232.

`updateAgent` compared its unlocked `existing.adapterConfig` snapshot against the row it had just written, so with two overlapping updates the later one could read another request's path change as its own correction — deleting health recorded for the path the agent actually uses now, clearing the error and lifting the pause on an agent still running without instructions.

The update transaction now re-reads the pre-update `adapterConfig` with `SELECT … FOR UPDATE` before writing, and compares that. Overlapping updates serialize on the row and each one only reacts to the change it made itself.

New test `does not clear health for the current path when another update lands first`: a transaction holding the row lock commits a path correction plus a failure on the new path while an unrelated rename waits; the rename must preserve the new path's health state and error reason. It fails when the comparison uses the unlocked snapshot.

```
npx vitest run server/src/__tests__/agent-instructions-health.test.ts \
  server/src/__tests__/agents-service-clear-error.test.ts \
  server/src/__tests__/agents-adapter-config-user-secret.test.ts \
  server/src/__tests__/agents-pending-approval-config.test.ts
# Test Files  4 passed (4)
# Tests  33 passed (33)

# with the comparison reverted to the unlocked snapshot
# Tests  1 failed | 23 passed (24)

pnpm --filter @paperclipai/server typecheck   # exit 0
```

That is every P1 from all four review rounds: recovery without a heartbeat, concurrent completions, resume vs. run completion, superseded-path failures, resume vs. terminate, and now overlapping updates. Each has a test that fails without its fix.

PR #10861, opened 2026-08-04
fix: release work assigned to terminated agents

dmndbrp-oss:
## Thinking Path

> - Paperclip manages AI agents and their work.
> - Agent termination is an agent lifecycle operation.
> - Open issues owned by a terminated agent cannot accept agent mutations.
> - The termination path did not release these issue assignments.
> - This pull request releases open work in the same transaction as termination.
> - The result keeps work available to an assignable manager or the unassigned queue.

## Linked Issues or Issue Description

**What happened?**

Terminating an agent left its open issues assigned to the terminated agent. This made the issues unavailable for normal agent work.

**Steps to reproduce**

Create an open issue assigned to an agent. Terminate that agent. Attempt to update the issue as its manager.

**Expected behavior**

Termination moves open issues to an assignable manager, or clears the assignment. The manager can then update the released issue.

**Actual behavior**

The issue remained assigned to the terminated agent and the update path could return 403.

## What Changed

- Release non-terminal agent assignments during termination in one transaction.
- Re-parent direct reports to an assignable manager, or clear their manager.
- Add a system audit comment for each released issue.
- Add an idempotent terminated-or-missing-assignee backfill script that revalidates manager eligibility inside its transaction.
- Add service and PATCH-route regression coverage.

## Verification

- `pnpm --filter @paperclipai/server exec vitest run src/__tests__/agents-service-clear-error.test.ts src/__tests__/authorization-service.test.ts --reporter=dot` — 62 passed.
- `pnpm --filter @paperclipai/server typecheck` — passed.
- `pnpm --filter @paperclipai/db typecheck` — passed.
- Live company-scoped idempotent backfill: `node server/node_modules/tsx/dist/cli.mjs scripts/backfill-terminated-assignee-locks.ts --company <company-id>` — `before=0`, `released_to_manager=0`, `released_to_unassigned_queue=0`, `after=0`.

## Risks

Low risk. Termination now changes only non-terminal issue assignments and direct reports in the same transaction. Terminal issues retain their historical assignee. The backfill is idempotent, uses a compare-and-set issue update, and revalidates the source/manager lifecycle graph before each live release.

## Model Used

OpenAI Codex, GPT-5.6-terra. Tool-assisted code inspection, editing, testing, git, and command execution. Context window: platform-managed.

## Checklist

- [x] I have included a thinking path that traces from project context to this change
- [x] I have specified the model used (with version and capability details)
- [x] I have checked ROADMAP.md and confirmed this PR does not duplicate planned core work
[...]

Issue #10508, opened 2026-07-30
Issues assigned to a terminated agent become fully immutable; terminated refs linger in execution_state participants and recovery-action owners


Waseemilyas:
### Environment
- paperclipai 2026.722.0, self-hosted

### What happened
After terminating an agent (the product's terminate verb — hard delete is FK-blocked by design, so termination is the supported "retire an agent" path), any non-terminal issue still assigned to that agent becomes **fully immutable**: `PATCH /api/issues/:id` rejects all writes, including simple re-assignment to a live agent or un-assignment, because the write validator re-checks the *current* row state (the terminated assignee) on every write rather than only gating the transition that caused it.

In addition, terminated-agent references are not cleaned up in two other places once termination happens:
- `execution_state.currentParticipant` can keep pointing at the terminated agent's ID.
- Recovery-action `owner` fields returned by the recovery API can still resolve to the terminated agent.

### Steps to reproduce
1. Assign an issue to an agent, leave it non-terminal (e.g. `in_review`).
2. Terminate that agent.
3. Attempt any `PATCH /api/issues/:id` on the issue (including just `assigneeUserId` or `blockedByIssueIds`).

### Observed
We terminated several retired agent roles as part of an agent-roster reduction. Over a hundred issues were assigned to the retired roles; a handful of them (previously `in_review`, assigned to a since-terminated role) ended up in this immutable state and could not be reassigned via the API — every PATCH attempt was rejected. We had to repair this out-of-band with a direct, scoped database write: reassign the affected issue rows, repoint the recovery-action owners, and clear the poisoned `execution_state.currentParticipant` values — then verified none remained.

### Expected
- Terminating an agent should not retroactively lock issues it used to own; a write should only be blocked by validation relevant to the write itself (e.g. "can't *assign* an issue to a terminated agent"), not by the historical fact that the *current* assignee happens to be terminated.
- Termination should sweep/clear the agent's ID out of `execution_state.currentParticipant` and any recovery-action `owner` fields for issues it no longer legitimately owns, or those fields should resolve terminated refs to a null/human-visible sentinel rather than silently keeping a dead reference.

### Impact
Operators cannot recover from a routine agent-retirement without direct database access — this defeats the intended "terminate = safe retire" contract and makes any agent-roster change on a live company operationally risky.

Waseemilyas:
[...]

PR #11407, opened 2026-08-15
fix(issues,routines): let the company reclaim work owned by a terminated agent

dzianisv:
## Thinking Path

> - Agent lifecycle and issue/routine ownership are two parts of that app
> - An agent can be terminated while it still holds open issues and active routines
> - The ownership guards compare assignee identity only. They do not ask if the assignee can still run
> - So one termination freezes every issue and routine that agent held, and no agent in the company can take the work back
> - This pull request treats a terminated assignee as unowned in the checkout guard, the issue authorization boundary, and the routine management guard
> - The benefit is that stuck work becomes reclaimable by normal agent flows, and a routine owned by a dead agent can be paused or moved

Refs #10508 — "Issues assigned to a terminated agent become fully immutable".

Related PR: #10861 releases open issues at termination time. This pull request is complementary. It repairs rows that are already stuck, including agents terminated before that change, and it also closes the routine hole.

**What happened?**

An issue assigned to a terminated agent is unreclaimable by every agent in the company. There is no supported path back. A routine owned by a terminated agent keeps firing and creates issues that nobody can pick up.

**Steps to reproduce**

1. Assign an open issue to an agent and give that agent a schedule routine.
2. Terminate the agent.
3. As a different agent in the same company, send `PATCH /api/issues/{id}` with a new `assigneeAgentId`.
4. As that same agent, send `POST /api/issues/{id}/checkout`.
5. As that same agent, send `PATCH /api/routines/{id}` with `{"status":"paused"}`.

**Actual behavior**

```
PATCH /api/issues/{id}   {"assigneeAgentId": ...}
  -> 403 {"error":"Issue is outside this actor's authorization boundary"}

POST /api/issues/{id}/checkout
  -> 409 {"error":"Issue checkout conflict",
          "details":{"assigneeAgentId":"<terminated-agent>","checkoutRunId":null,"executionRunId":null}}

PATCH /api/routines/{id}   {"status":"paused"}
  -> 403 {"error":"Agents can only manage routines assigned to themselves"}
```

`checkoutRunId` and `executionRunId` are both null. No run holds the issue. The 409 is only "the assignee is a different agent". The boundary check does not consider that this agent can never run again. The manager agent gets the same two answers.

**Expected behavior**

A terminated agent holds no live claim. Another agent in the company can check the issue out, can update it, and can manage the routine.

## What Changed

- `server/src/services/authorization.ts`: allow a company agent to act on an issue when the assignee agent is terminated. The decision carries the new reason `allow_terminated_assignee`.
- `server/src/routes/issues.ts`: skip the assignee run lock in the agent issue-mutation guard when the boundary decision reports that reason. The route adds no lookup of its own.
- `server/src/services/issues.ts`: on checkout, release a terminated assignee when `checkoutRunId` and `executionRunId` are both null. The normal checkout path then claims the issue. The release writes an `issue.terminated_assignee_released` activity entry that names the previous assignee.
- `server/src/routes/routines.ts`: allow any agent in the company to manage a routine whose assignee is terminated. A routine with a live assignee stays protected.
- `server/src/services/terminated-assignee.ts`: new small helper that reads the assignee agent status.
- `server/src/__tests__/terminated-assignee-reclaim.test.ts`: new regression suite against embedded Postgres.
- `server/src/__tests__/routines-routes.test.ts`: mock the new helper and add a case for the terminated-assignee read path.
- `doc/SPEC-implementation.md`: record the rule in the atomic checkout contract.

## Verification

```
$ cd server && npx vitest run src/__tests__/terminated-assignee-reclaim.test.ts src/__tests__/routines-routes.test.ts
 Test Files  2 passed (2)
      Tests  21 passed (21)
```

The new suite terminates an agent that holds an open issue and an active routine, then proves another agent can reclaim both:

- checkout of a `todo` issue owned by a terminated agent returns 200 and the new assignee.
- `PATCH` of an `in_progress` issue left by a terminated assignee returns 200.
- an issue held by a live run still returns 409 and keeps its assignee.
- the automatic release writes an `issue.terminated_assignee_released` activity row that names the previous assignee.
- `PATCH /api/routines/{id}` by a non-assignee agent returns 200 for pause and for reassign after the owner is terminated.
- `PATCH /api/routines/{id}` still returns 403 while the assignee is alive.

Regression suites that cover the touched guards:

```
$ npx vitest run src/__tests__/authorization-service.test.ts src/__tests__/routines-routes.test.ts src/__tests__/issues-checkout-wakeup.test.ts
 Test Files  3 passed (3)
      Tests  80 passed (80)

$ npx vitest run src/__tests__/issue-agent-mutation-ownership-routes.test.ts src/__tests__/issue-stale-execution-lock-routes.test.ts src/__tests__/attention-service.test.ts
 Test Files  3 passed (3)
      Tests  111 passed (111)
```

`npx tsc -p server/tsconfig.json --noEmit` reports no error in the changed files. The repository has pre-existing errors in `server/src/routes/plugins.ts` from an unbuilt `@paperclipai/plugin-sdk`; those are not touched here.

## Risks

Low to medium risk, and the blast radius is limited to agents whose status is `terminated`.

- The change widens who can write to an issue. It only widens it when the assignee agent is terminated, which is a terminal state that no agent can leave.
- Checkout still refuses to steal an issue that a run holds. The release is guarded by `checkoutRunId is null and executionRunId is null` in the same statement, so a live lock is never broken.
- The routine guard now costs one extra agent lookup on each non-assignee call. The lookup runs only after the identity check fails.
- No schema change and no migration. The rollback is a revert of this commit.

## Model Used

Anthropic Claude, `claude-opus-5`, accessed through GitHub Copilot. High reasoning mode. Tool use for file read/edit, shell command execution, and test runs.

## Checklist

### DIFF
--- a/doc/SPEC-implementation.md
+++ b/doc/SPEC-implementation.md
@@ -967,6 +967,8 @@ Server behavior:
 2. if updated row count is 0, return `409` with current owner/status
 3. successful checkout sets `assignee_agent_id`, `status = in_progress`, and `started_at`
 
+A terminated agent can never run again, so work it still holds is unowned. Before the checkout update, the server clears an assignee that points at a terminated agent when `checkout_run_id` and `execution_run_id` are both null. The issue authorization boundary applies the same rule, so any agent in the company can also update such an issue. The same rule applies to routines: an agent that is not the assignee may manage a routine whose assignee agent is terminated.
+
 `POST /issues/:issueId/admin/force-release` is an operator recovery endpoint for stale harness locks. It requires board access to the issue company, clears checkout and execution run lock fields, and may clear the agent assignee when `clearAssignee=true` is passed. The route must write an `issue.admin_force_release` activity log entry containing the previous checkout and execution run IDs.
 
 ## 10.5 Projects
--- a/server/src/__tests__/routines-routes.test.ts
+++ b/server/src/__tests__/routines-routes.test.ts
@@ -129,6 +129,8 @@ const mockAccessService = vi.hoisted(() => ({
   canUser: vi.fn(),
 }));
 
+const mockIsTerminatedAgentId = vi.hoisted(() => vi.fn());
+
 const mockLogActivity = vi.hoisted(() => vi.fn());
 const mockTrackRoutineCreated = vi.hoisted(() => vi.fn());
 const mockGetTelemetryClient = vi.hoisted(() => vi.fn());
@@ -151,6 +153,10 @@ function registerModuleMocks() {
 
   vi.doMock("../services/routines.js", () => ({
     routineService: () => mockRoutineService,
+  }));
+
+  vi.doMock("../services/terminated-assignee.js", () => ({
+    isTerminatedAgentId: mockIsTerminatedAgentId,
   }));
 
   vi.doMock("../services/activity-log.js", () => ({
@@ -190,6 +196,7 @@ describe("routine routes", () => {
     vi.doUnmock("../services/index.js");
     vi.doUnmock("../services/activity-log.js");
     vi.doUnmock("../services/routines.js");
+    vi.doUnmock("../services/terminated-assignee.js");
     vi.doUnmock("../routes/routines.js");
     vi.doUnmock("../routes/authz.js");
     vi.doUnmock("../middleware/index.js");
@@ -215,6 +222,7 @@ describe("routine routes", () => {
       status: "issue_created",
     });
     mockAccessService.canUser.mockResolvedValue(false);
+    mockIsTerminatedAgentId.mockResolvedValue(false);
     mockLogActivity.mockResolvedValue(undefined);
     mockRoutineService.getDescriptionDocument.mockResolvedValue({
       id: "99999999-9999-4999-8999-999999999999",
@@ -470,6 +478,20 @@ describe("routine routes", () => {
 
     expect(res.status).toBe(403);
     expect(mockRoutineService.listRevisions).not.toHaveBeenCalled();
+  });
+
+  it("lets any company agent read revisions when the assigned agent is terminated", async () => {
+    mockIsTerminatedAgentId.mockResolvedValue(true);
+    const app = await createApp({
+      type: "agent",
+      agentId: otherAgentId,
+      companyId,
+    });
+
+    const res = await request(app).get(`/api/routines/${routineId}/revisions`);
+
+    expect(res.status).toBe(200);
+    expect(mockRoutineService.listRevisions).toHaveBeenCalledWith(routineId);
   });
 
   it("restores routine revisions with existing routine-management permissions", async () => {
--- a/server/src/__tests__/terminated-assignee-reclaim.test.ts
+++ b/server/src/__tests__/terminated-assignee-reclaim.test.ts
@@ -1,0 +1,294 @@
+import { randomUUID } from "node:crypto";
+import express from "express";
+import request from "supertest";
+import { eq } from "drizzle-orm";
+import { afterAll, beforeAll, describe, expect, it } from "vitest";
+import {
+  agents,
+  companies,
+  companyMemberships,
+  createDb,
+  heartbeatRuns,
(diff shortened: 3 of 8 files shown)


dzianisv:
Addressed in 9a85f43.

`releaseTerminatedAssignee` now reads the current assignee before the guarded update and writes an `issue.terminated_assignee_released` activity entry with `previousAssigneeAgentId` and `reason: "assignee_agent_terminated"`. The regression test asserts that row after a reclaiming checkout.

The same commit also fixes the two CI failures from the first push:

- `external-object-routes` returned 500 because the route guard called `agentService.getById`, which route unit tests do not stub. The terminated case now travels on the authorization decision as the new reason `allow_terminated_assignee`, so the route needs no lookup of its own.
- The new regression file failed its shard with an unhandled postgres error. Assigning an issue wakes the new assignee, and that background run wrote to the embedded database after the test stopped it. The suite now mocks the wake, since it tests ownership rules and not wake delivery.

dzianisv:
## Each half of the regression suite fails on its own without the fix

A reviewer on a running deployment (`@paperclipai/server 2026.814.0-canary.7`) reported that the two halves of this bug do **not** behave identically there:

- `PATCH /api/issues/{id} {"assigneeAgentId": ...}` on issues held by a terminated agent returned **200**, not 403.
- `PATCH /api/routines/{id} {"assigneeAgentId": ...}` on a routine held by the same terminated agent still returned **403 `Agents can only manage routines assigned to themselves`**.

Their concern was fair: a suite that terminates an agent and then only proves the *issue* is reclaimable could pass on such a build while the routine stays frozen forever. So I re-verified the suite by reverting only the source changes and keeping the tests.

```
$ git checkout <base> -- server/src/routes/issues.ts server/src/routes/routines.ts \
    server/src/services/authorization.ts server/src/services/issues.ts
$ cd server && npx vitest run src/__tests__/terminated-assignee-reclaim.test.ts

 ❯ src/__tests__/terminated-assignee-reclaim.test.ts (5 tests | 3 failed)
   × lets another agent check out an issue whose assignee was terminated
   × lets another agent mutate an in_progress issue left behind by a terminated assignee
   × lets another agent take over a routine owned by a terminated agent

 Tests  3 failed | 2 passed (5)
```

Failure detail, unmodified:

- checkout → `409 {"error":"Issue checkout conflict","details":{...,"checkoutRunId":null,"executionRunId":null}}`
- issue PATCH → `409 issue_write_assignee_run_lock` — "Terminated VP of Engineering has this task checked out and a run is live"
- routine PATCH → `403 {"error":"Agents can only manage routines assigned to themselves"}`

With the fix restored:

```
$ npx vitest run src/__tests__/terminated-assignee-reclaim.test.ts
 Test Files  1 passed (1)
      Tests  5 passed (5)
```

Three things worth recording from this:

1. **The routine assertion stands alone.** `lets another agent take over a routine owned by a terminated agent` and `still refuses routine management when the assignee is alive` are separate cases against `PATCH /api/routines/{id}`, covering both `status` and `assigneeAgentId`. Neither depends on the issue boundary, and the first one fails without the `routes/routines.ts` change even when the issue cases are removed. That matches the deployed build: `dist/routes/routines.js` still holds the flat self-ownership check at both call sites, with no terminated-assignee escape and no governance-mode branch.
2. **The issue half can be masked by governance mode.** In a company running the simple governance path, an issue write by a company member is allowed by `allow_simple_company_member`, which is why the reviewer saw 200 on the issue PATCH. The surviving blocker in that configuration is not the boundary but the assignee run lock and the checkout conflict — and both still fail without this PR, as the run above shows. The 403 boundary from the original report is the stricter-governance shape of the same hole; the PR closes all three.
3. **The routine hole is unconditional.** No governance mode, permission grant, or manager relationship gets past that guard today; only a board actor does. That is what leaves a routine owned by a dead agent minting issues nobody can pick up.

No code change from this comment — the suite already asserts the routine path independently, and now there is direct evidence that it fails on its own.

PR #11416, opened 2026-08-15
fix(onboarding): carry an existing company's mission into the wizard

tonio-alucema:
## Thinking Path

> - A company's mission is what its lead agent is hired to pursue, so onboarding writes it into that agent's instructions file
> - A company created by Paperclip Cloud already gave its mission at signup, so #11352 sends it straight to the agent step instead of asking again
> - But the mission field is only ever filled by the step that was skipped, so it arrives at the hire empty
> - So the lead agent of every Cloud-seeded company is created knowing nothing of the mission that company was seeded with
> - This pull request reads the mission back from the company's own goal, holds the hire until that read lands, and stops a run walking back into steps that do not apply to it
> - The benefit is that skipping a question the customer already answered no longer costs the answer

Ports the substance of #11259, which predates the recent onboarding work and is still open. No public issue exists. The problem follows.

**What happened?**

A company that already has its mission opens onboarding on the agent step. Steps 1 and 2 never run, so `companyGoal` stays empty. `composeCeoInstructions` seeds the lead agent's instructions from that field, and the Review checklist reads it.

The agent is hired successfully. Its instructions file simply has no `**Mission:**` line, and the Review step shows "Mission" as not done.

**Expected behavior**

A company entered on the agent step carries its own mission. The lead agent's instructions name it.

**Steps to reproduce**

1. Create a company with a mission (Cloud signup, or any company with a company-level goal).
2. Open onboarding for it. It opens on "Create your first agent".
3. Complete the hire.
4. The agent's instructions file contains no mission.

`master` at `484b1f626`.

**Related pull requests**

#11259 is the origin of this work and should close when this lands. #11352 introduced the regression described above.

## What Changed

- `ui/src/lib/onboarding-mission.ts` — new. `selectExistingCompanyMission`, `isExistingCompanyMissionUnresolved`, `planMissionPersistence`.
- `ui/src/lib/onboarding-goal.ts` — adds `formatOnboardingGoalInput`, the inverse of `parseOnboardingGoalInput`.
- `ui/src/lib/onboarding-route.ts` — adds `canGoBackFromOnboardingStep` and `canJumpToOnboardingStep`.
- `ui/src/hooks/useCompanyMission.ts` — also returns the mission itself, and `fetching` separately from `settled`.
- `ui/src/components/OnboardingWizard.tsx` — hydrates the mission, holds the hire until it resolves, bounds backward navigation by the entry step, and updates rather than skips when a goal is already in hand.
- Tests: `onboarding-mission.test.ts` (new), plus cases in `onboarding-goal.test.ts`, `useCompanyMission.test.tsx` and `OnboardingWizard.step.test.tsx`.

### The regression this fixes

`d95340b0b` sends a company that already has its mission to the agent step, which is right. What it missed is that the mission field is filled only by the step being skipped, and that the field feeds the agent's instructions rather than only the form.

Before that change these companies dead-ended on the mission step and never hired at all. A dead end became a quiet data loss, which is worse: it completes successfully and reports nothing.

### The hire waits for the read

`isExistingCompanyMissionUnresolved` holds the hire while the goal read is outstanding, and counts an in-flight refetch over cached goals as unresolved rather than as already answered.

This is the same rule the draft ownership gate settled on in #11382 — `isFetching`, not `isLoading`, because retained data is not an answer to the question being asked now. #11259 had it on 11 August. I arrived at it independently on the 14th, for a different consumer, not having read this branch. Worth recording, because it is the second time this week the same distinction has been the fix.

### Backward navigation is bounded by the entry step

The Back button already applied this rule inline. The progress bar applied only the "already completed" half, so a run entered on the agent step could still jump to step 1 — the step whose job is to create a company — while already holding one.

The entry step is captured once, when the wizard opens, for the same reason the step itself is: it derives from queries, so a live read would move the floor under a customer mid-flow.

### Why `planMissionPersistence` came too

It looked like tidying and turned out to be required. Hydration sets `createdCompanyGoalId` from the company's existing goal, and confirming the mission read that id as "already written" and skipped the write — discarding the customer's edit.

That skip was only ever safe while the id could arrive one way: by writing. Once it can also arrive from hydration, a goal in hand has to mean update it.

## Verification

- `ui` typecheck is clean.
- Full `ui` suite: **4004 pass**.

Every piece was checked by removing it and confirming a specific case fails, rather than trusting a green run:

| removed | result |
| --- | --- |
| mission hydration | agent hired with no mission in its instructions |
| hire gate | hire proceeds during an in-flight re-read |
| progress-bar bound | step 1 reachable from the agent step |
| Back-button bound | Back offered at the entry step |
| update-not-skip | an edited mission is discarded on confirm |

The hydration case asserts on `saveInstructionsFile`'s content — the actual consumer — rather than on the mission textarea, because the entry path never renders that field and the navigation bound now correctly prevents reaching it.

Two failures remain, in `IssueProperties.test.tsx` and `StatusCards/format.test.ts`. Both are date-dependent, both reproduce on `master` with these changes stashed, and neither file is touched here.

**Not done:** no end-to-end run through Cloud signup. That needs a staging tenant on this build.

## Risks

Low. Hydration only fills an empty field, so a customer editing their mission is never overwritten by the stored copy, and a company with no goal is unaffected.

The hire gate can delay the Connect button while a goal re-read is in flight. It cannot block it indefinitely: the query settles either way, and a failed read resolves to "no mission" rather than staying pending — the same fail-open rule the rest of this flow follows.

The navigation bound removes reachable steps. That is the intent — those steps either do not apply to the run or would create a second company — but it is a visible change for anyone entering onboarding above step 1.

Revert the commit to restore.

## Model Used

## Checklist

### DIFF
--- a/ui/src/components/OnboardingWizard.step.test.tsx
+++ b/ui/src/components/OnboardingWizard.step.test.tsx
@@ -32,6 +32,14 @@ const mockGoalsApi = vi.hoisted(() => ({
   update: vi.fn(),
 }));
 const mockAdaptersApi = vi.hoisted(() => ({ list: vi.fn() }));
+const mockAgentsApi = vi.hoisted(() => ({
+  create: vi.fn(),
+  adapterModels: vi.fn(),
+  hire: vi.fn(),
+  instructionsBundle: vi.fn(),
+  saveInstructionsFile: vi.fn(),
+  testEnvironment: vi.fn(),
+}));
 const mockCompaniesApi = vi.hoisted(() => ({ create: vi.fn() }));
 
 const routerState = vi.hoisted(() => ({ pathname: "/" }));
@@ -54,9 +62,7 @@ const companyState = vi.hoisted(() => ({
 vi.mock("../api/goals", () => ({ goalsApi: mockGoalsApi }));
 vi.mock("@/api/adapters", () => ({ adaptersApi: mockAdaptersApi }));
 vi.mock("../api/companies", () => ({ companiesApi: mockCompaniesApi }));
-vi.mock("../api/agents", () => ({
-  agentsApi: { create: vi.fn(), adapterModels: vi.fn().mockResolvedValue([]) },
-}));
+vi.mock("../api/agents", () => ({ agentsApi: mockAgentsApi }));
 vi.mock("../api/approvals", () => ({ approvalsApi: { create: vi.fn() } }));
 vi.mock("../api/issues", () => ({ issuesApi: { create: vi.fn() } }));
 vi.mock("../api/projects", () => ({ projectsApi: { list: vi.fn(), create: vi.fn() } }));
@@ -176,6 +182,16 @@ describe("OnboardingWizard — which step it lands on", () => {
     dialogState.onboardingRouteDismissed = false;
     mockAdaptersApi.list.mockResolvedValue([]);
     mockGoalsApi.list.mockResolvedValue([]);
+    mockAgentsApi.adapterModels.mockResolvedValue([]);
+    mockAgentsApi.hire.mockResolvedValue({ agent: { id: "agent-1" }, approval: null });
+    mockAgentsApi.instructionsBundle.mockResolvedValue({ entryFile: "AGENTS.md" });
+    mockAgentsApi.saveInstructionsFile.mockResolvedValue({});
+    mockAgentsApi.testEnvironment.mockResolvedValue({
+      adapterType: "claude_local",
+      status: "pass",
+      checks: [],
+      testedAt: new Date("2026-03-02T00:00:00Z").toISOString(),
+    });
   });
 
   afterEach(async () => {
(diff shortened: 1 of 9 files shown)


tonio-alucema:
Valid — fixed, and the way I first tested it is worth recording.

**The finding.** `hydratedMissionForRef` is a ref, so it outlives the state it describes. `reset()` clears `companyGoal` on close, the marker stayed set, and the next run for the same company skipped hydration while the hire proceeded — the original defect back again by another route. The marker now travels with the field it describes: cleared in `clearCompanyScopedState` and in `reset()`, wherever `companyGoal` is cleared.

**My first test for it was worthless.** It passed against the unfixed code. Its close-button lookup matched nothing, optional chaining swallowed the click, so no `reset()` ever ran and the field was never cleared — the test walked a path that had no bug in it. It now asserts the button was found and that the wizard actually left the agent step before reopening.

**And one honest caveat on the fix.** The new case fails only when *both* clears are removed:

| removed | result |
| --- | --- |
| `reset()` clear | passes |
| `clearCompanyScopedState` clear | passes |
| both | fails |

`reset()` also clears `createdCompanyId`, so the next company introduction routes through `clearCompanyScopedState` and clears the marker there anyway. Each is redundant with the other *today*. I kept both deliberately — as one invariant rather than one guard plus a coincidence, since a later `reset()` that preserved the company would silently restore this bug — but I'd rather say that than let the table above look like two independently verified guards.

`ui` typecheck clean; 50 pass across the wizard, mission and hook suites.

PR #11417, opened 2026-08-15
fix(ui): scope invite membership checks to the signed-in account

tonio-alucema:
## Thinking Path

> - Which companies a person belongs to is an authorization fact the server owns, and the UI caches the answer under a single `["companies"]` key
> - That cache entry carries no account identity, and `main.tsx` sets `staleTime: 30_000` for every query, so for thirty seconds after a sign-in the previous account's list is served with no request at all
> - The invite landing page reads that list to decide whether the person is already a member of the inviting company
> - A list that arrives with no loading state and no error therefore looks authoritative while describing somebody else
> - This pull request makes the page trust only a list it fetched itself, for the account signed in now
> - The benefit is that a membership decision stops depending on cache freshness, which nothing in the app guarantees

No public issue exists. Refs #11380, #11382. The problem follows.

**What happened?**

`InviteLanding` read the shared `["companies"]` cache entry as proof of membership in two places:

- The post-sign-in redirect called `fetchQuery(companiesListQueryOptions)`, which returns the cached entry without a request while it is inside the app-wide `staleTime`.
- An effect cleared the pending invite token whenever the cached list contained the invited company.

Neither checked that the list belonged to the account signed in now. A second account signing in on a warm tab, or a session that lapses server-side, is enough to reach both.

**Expected behavior**

The page decides membership from a company list fetched for the current session.

**Steps to reproduce**

1. On a self-hosted instance in `authenticated` mode, sign in as account A, which belongs to company X.
2. Within thirty seconds, open an invite link for company X and sign in as account B, which does not belong to it.
3. The page reads A's cached list, finds company X, and treats B as already a member.

`master` at `2a4b4bc63`.

## What Changed

- `ui/src/pages/InviteLanding.tsx` — the membership query sets `staleTime: 0` so it revalidates on mount, and the verdict is withheld until that fetch lands, keyed on `isFetchedAfterMount`. The token-clearing effect and the "already a member" branch both read through that gate.
- `ui/src/pages/InviteLanding.tsx` — the post-sign-in path cancels anything still in flight for the previous session, then forces a fetch for the new one with `staleTime: 0`.
- `ui/src/pages/Auth.tsx` — sign-in resets the companies query instead of invalidating it. Invalidation leaves the previous account's list readable, and its fetch running, until the refetch returns.
- `ui/src/pages/InviteLanding.test.tsx` — coverage for the warm-cache case, the token-clearing effect, and the `local_trusted` exemption.

### `local_trusted` is exempt

Those instances have no accounts, so the shared list is the only identity there is. `membershipIsAccountScoped` is false there and the gate stays open.

### Rebased onto the account-keyed cache

#11488 landed while this was open and keys the company list by account, so the page can no longer reach another account's list at all. Two things changed here as a result:

- The post-sign-in read now calls `fetchCompanyListForCurrentAccount`, which replaces the `cancelQueries` plus forced `fetchQuery` this PR originally carried. The helper is strictly stronger: it detaches the in-flight `/companies` request inside the query function, and it resolves the account identity past the session invalidation immediately above rather than trusting the session entry still in the cache.
- The observer reads through `useCompanyListQuery`.

The mount-scoped `isFetchedAfterMount` gate is **kept**, not removed. Its purpose has narrowed — cross-account leakage is now structurally impossible, so what remains is holding the verdict until this page has a list rather than acting on a pending one. It is still load-bearing: disabling it fails two tests here. Removing a defense in the same change that rebases onto a new foundation is the wrong order; that is a follow-up once the keying has proven itself.

`Auth.tsx` can safely reset, because it navigates away on success and `InviteLanding` mounts fresh afterward. Measurements of exactly when that rewind does and does not bite are in [#11380](https://github.com/paperclipai/paperclip/pull/11380#issuecomment-5300984911).

## Verification

- `InviteLanding.test.tsx`, `Auth.test.tsx`, `companies-query.test.ts`, `CompanyContext.test.tsx` together: **51 passed**, run twice.
- `pnpm tsc -b`: clean.
- `InviteLanding.test.tsx` and `Auth.test.tsx` together: 21 passed.

Both failures are pre-existing and unrelated. Each reproduces on a tree that does not contain this change, in files this change does not touch:

| Failure | Why it fails |
| --- | --- |
| `IssueProperties.test.tsx` | Timezone-dependent: expects `4:08 PM`, gets `9:08 AM` |
| `StatusCards/format.test.ts` | Time-of-day dependent: "only counts updates started today" breaks near midnight |

**Not done:** no manual two-account run in a browser. The path needs two accounts on an `authenticated` instance, which a local dev instance cannot exercise.

## Risks

Low. The failure direction is a membership verdict withheld for one extra round trip, which resolves itself; the direction it removes is one account's membership granted to another, which does not.

It adds one request per invite-page mount, on a query key the app already uses.

**This does not close the class.** The shared list is still unscoped for every other consumer. #11380 clears it on sign-out and #11382 handles the onboarding draft gate; all three are needed, because an account can change without passing through any one of those paths.

## Model Used

## Checklist

### DIFF
--- a/ui/src/pages/Auth.tsx
+++ b/ui/src/pages/Auth.tsx
@@ -55,7 +55,11 @@ export function AuthPage() {
       setError(null);
       await queryClient.invalidateQueries({ queryKey: queryKeys.auth.session });
       await queryClient.invalidateQueries({ queryKey: queryKeys.health });
-      await queryClient.invalidateQueries({ queryKey: queryKeys.companies.all });
+      // Reset rather than invalidate: the `["companies"]` entry is shared app-wide and
+      // is not account-scoped, so invalidating leaves the previous account's list
+      // readable (and any fetch for that session in flight) until the refetch lands.
+      // Sign-in can change accounts, so drop the list outright.
+      await queryClient.resetQueries({ queryKey: queryKeys.companies.all });
       navigate(nextPath, { replace: true });
     },
     onError: (err) => {
--- a/ui/src/pages/InviteLanding.test.tsx
+++ b/ui/src/pages/InviteLanding.test.tsx
@@ -767,6 +767,231 @@ describe("InviteLandingPage", () => {
     });
   });
 
+  // The `["companies"]` cache entry is shared app-wide and carries no account
+  // identity, so a list fetched moments ago for a different account is still inside
+  // the app-wide staleTime. Membership decisions on this page — the post-sign-in
+  // redirect, the "already a member" panel, and clearing the pending invite token —
+  // must run against a list fetched for the account that is signed in now.
+  describe("membership checks against a previous account's cached company list", () => {
+    const PREVIOUS_ACCOUNT_COMPANIES = [{ id: "company-1", name: "Acme Robotics" }];
+
+    // Mirrors the app-wide default in main.tsx. Without it every cache entry is
+    // stale on read and the staleness these tests cover cannot occur.
+    // `ownerId` is whose list this warm entry is. Since #11488 the entry is keyed
+    // by account, so "a list left by a previous account" means putting it under
+    // that account's key — which is also why the page can no longer reach it.
+    // `null` is the signed-out / local_trusted key.
+    function createAppLikeQueryClient(ownerId: string | null) {
+      const queryClient = new QueryClient({
+        defaultOptions: { queries: { retry: false, staleTime: 30_000 } },
+      });
+      queryClient.setQueryData(queryKeys.companies.list(ownerId), {
+        companies: PREVIOUS_ACCOUNT_COMPANIES,
+        unauthorized: false,
+      });
+      return queryClient;
+    }
+
+    it("re-reads membership for the account that just signed in", async () => {
+      getSessionMock.mockResolvedValueOnce(null);
+      getSessionMock.mockResolvedValue({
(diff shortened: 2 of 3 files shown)


tonio-alucema:
Known follow-up, recorded here so it is not lost.

The post-sign-in path in this PR uses `cancelQueries` plus `fetchQuery({ staleTime: 0 })`, which guarantees the membership check reads a *query cache* entry fetched for the account that just signed in. It does not guarantee a fresh *HTTP request*. `api.get` coalesces identical in-flight GETs keyed on the request path alone (`coalescedGet` in `ui/src/api/client.ts`), and `companiesListQueryOptions`' query function passes no abort signal, so `cancelQueries` cannot abort the underlying request. A `/companies` request still in flight from before the sign-in can therefore answer the forced fetch.

The exposure is narrower than it sounds, and the common case fails safe: a request issued while signed out answers 401, which the query function folds to `{ companies: [], unauthorized: true }`. An empty list is never read as membership. What remains is the case where a request was issued under a *previous account's* session and is still in flight when the new sign-in completes.

#11430 adds `detachInflightGet(path)` and `companiesApi.detachInflightList()` for this. Rather than land the same helper in two PRs, this one picks it up by rebasing on whichever of #11430 / #11380 merges first.

<sub>Posted with Claude Code (Claude Opus 5).</sub>

Issue #11436, opened 2026-08-15
codex_local: managed CODEX_HOME inherits the host config's model_provider/model pin, breaking every run


ericdfields:
A Paperclip-managed `CODEX_HOME` is seeded by copying `config.toml` out of the host home (`$CODEX_HOME`, else `~/.codex`). The copy includes the host's **root-level `model` / `model_provider`** keys — which Paperclip never wants, and which silently outrank both mechanisms it actually uses:

- the model is passed as `--model <id>` on the CLI (`packages/adapters/codex-local/src/server/codex-args.ts`), and when an agent configures none, the deliberate intent is "let the Codex CLI pick its default";
- the provider is selected only by the Paperclip-managed `model_provider` block that `PAPERCLIP_CODEX_PROVIDERS` merges in (`packages/adapters/codex-local/src/server/runtime-config.ts`).

## Reproduction

Host `~/.codex/config.toml`:

```toml
model_provider = "omlx"
model = "Qwen3.5-9B-MLX-4bit"
personality = "pragmatic"

[model_providers.omlx]
base_url = "http://127.0.0.1:8008/v1"
env_key = "OMLX_API_KEY"
```

`OMLX_API_KEY` is not set (the gateway is dormant). Every managed run seeded from this host then dies:

```
ERROR codex_models_manager::manager: failed to refresh available models: Missing environment variable: `OMLX_API_KEY`.
{"type":"turn.failed","error":{"message":"Missing environment variable: `OMLX_API_KEY`."}}
```

The error names neither the config file it came from nor the host home it was copied out of, so it is very hard to trace back to a seed-time copy.

## Why deleting the pin by hand does not fix it

The managed home is derived state and gets regenerated: per-agent home migration (`agents/<agentId>/codex-home`), company-dir cleanup, a new Paperclip instance, or adding a second `codex_local` agent. Each regeneration re-seeds straight from the still-poisoned host config, so the failure returns silently — from the control plane's point of view runs just start failing.

Cleaning the host config is not a general fix either: it is the user's personal file, the setup may be intentional-but-dormant, and it only fixes one machine.

## Affected code

Both engines that build a managed home copy `config.toml` verbatim:

- `packages/adapters/codex-local/src/server/codex-home.ts` — `seedManagedCodexHome` / `COPIED_SHARED_FILES`
- `packages/adapter-utils/src/acpx-engine/execute.ts` — the ACPX engine's own managed-home seed

## Proposed fix

[...]

PR #11437, opened 2026-08-15
fix(codex): never inherit the host's model/provider pin into a managed Codex home

ericdfields:
## Thinking Path

> - The `codex_local` adapter runs each agent against a Paperclip-managed `CODEX_HOME`, so agent runs do not share or mutate the user's personal Codex state
> - That managed home is derived state: Paperclip creates it by copying `config.toml` out of the host home (`$CODEX_HOME`, else `~/.codex`)
> - The copy included the host's root-level `model` and `model_provider` keys, which Paperclip never asked for and which outrank the selection it actually makes
> - A host pin at a dormant local gateway therefore makes every managed run fail with `Missing environment variable: OMLX_API_KEY`, and deleting the pin from the managed copy does not hold because the home is re-seeded from the same host config
> - This pull request strips the host's model and provider selection when seeding a managed home, in both engines that build one, and repairs homes that were already seeded
> - The benefit is that a user's personal Codex model choice can no longer silently break every agent run, on any machine

Refs #11436

Related PRs:

- #1995 (open) — `fix(codex): sanitize sandbox=elevated in managed Codex home config`. Independent work, same seam and the same conclusion: seed `config.toml` through a sanitizer instead of copying it, and re-sanitize an already-seeded home. See Risks below.
- #7919 (merged) — added `PAPERCLIP_CODEX_PROVIDERS`, the supported way to select a provider. This PR keeps that the only way.

## What Changed

- Added `packages/adapter-utils/src/codex-config-seed.ts`. It strips root-level `model` and `model_provider` assignments from Codex `config.toml` text, and seeds the managed home's copy through that transform.
- Changed `packages/adapters/codex-local/src/server/codex-home.ts` to seed `config.toml` through the new helper. This covers the shared company home, the per-agent home set by the server isolation guard, worktree-isolated homes, and the `reconcileManagedCodexHome` startup backfill.
- Changed `packages/adapter-utils/src/acpx-engine/execute.ts` to use the same helper. The ACPX engine has its own managed-home seed. It had no test coverage. Its seed function is now exported and tested.
- The helper also repairs an already-seeded home. A home created before this change keeps the dead pin, and the next re-seed puts it back.
- The helper does not rewrite a managed `config.toml` that is already clean, so a normal run does not change the file.
- The helper writes the managed copy with mode `0600`. `config.toml` holds the managed MCP `Authorization: Bearer` header, so it must not inherit a group-readable or world-readable host mode.
- `[model_providers.*]` tables are kept. They have no effect unless something selects them.
- A root-level `profile` key is reported in the log but not removed. A profile can select a provider, but it also holds policy that a user can want.
- Changed two tests in `server/src/__tests__/codex-local-execute.test.ts`. See Risks.
- Added the rule to the `codex_local` configuration doc in `packages/adapters/codex-local/src/index.ts`.

## Verification

Automated:

```
npx vitest run packages/adapters/codex-local \
  packages/adapter-utils/src/codex-config-seed.test.ts \
  server/src/__tests__/codex-local-execute.test.ts \
  server/src/__tests__/codex-auth-reconciliation.test.ts \
  server/src/__tests__/codex-local-skill-injection.test.ts \
  server/src/__tests__/agent-adapter-validation-routes.test.ts
```

383 tests pass. Three unrelated failures on this tree were confirmed to be identical on a clean checkout: 26 EPIPE unhandled errors in `acpx-engine/execute.test.ts`, 7 Linux-only `local-process-sandbox` tests on darwin, and 139 `@paperclipai/plugin-sdk` server typecheck errors.

Every new test was mutation-tested. 15 mutations were applied one at a time and all 15 were caught: identity sanitizer, no repair pass, no no-op guard, removed same-home guard, mode `0644`, strip inside tables, no multi-line continuation skip, substring key match, each engine skipping the sanitizer, and each engine copying `config.toml` verbatim.

Manual, against a real host config carrying the pin:

```
host config byte-identical after seed: true
no root model_provider in seeded:      true
no root model pin in seeded:           true
provider table preserved:              true
mcp_servers preserved:                 true
seeded mode:                           600
```

End to end with the real `codex` binary:

```
A) codex exec against the host home
   ERROR codex_models_manager::manager: failed to refresh available models:
     Missing environment variable: `OMLX_API_KEY`.

B) codex exec against a per-agent home freshly seeded from that same host config
   {"type":"item.completed","item":{"type":"agent_message","text":"OK"}}
   {"type":"turn.completed", ...}
```

## Risks

- **Two existing tests were changed.** `server/src/__tests__/codex-local-execute.test.ts` seeded the shared home with `model = "codex-mini-latest"` and asserted that it reached the managed home. Those tests asserted the behaviour this PR removes. Their stated purpose is "preserving shared auth and config", and the fixture content was incidental, so they now seed a host pin **and** a plain key and assert that only the plain key propagates. Coverage is the same and the tests now prove the fix at the execute level. Please confirm the original intent was config propagation and not model inheritance.
- **A user who deliberately pinned a model in the host config loses that pin for agent runs.** This is the intended change. Set `adapterConfig.model` instead. The log line names the keys that were dropped and both home paths, so the behaviour is diagnosable rather than silent.
- **The repair pass writes to an already-seeded managed home.** It only removes root-level `model` and `model_provider`. It never touches the host home, which is guarded and asserted. An external, user-supplied `CODEX_HOME` override is not seeded or inspected at all, so it is unaffected.
- **Overlap with #1995.** That PR adds a second sanitizer at the same seam for `sandbox = "elevated"`. Whichever lands second should add its transform to `seedSanitizedCodexConfigToml` rather than add a second sanitize pass over the same file. #1995 is based on an older `codex-home.ts` from before `seedManagedCodexHome` was split out, so it needs a rebase either way.
- The TOML handling is line-oriented, not a full parser. This matches the existing approach in `runtime-config.ts`. Removal of a multi-line string value drops its continuation lines, which is tested, so it cannot leave a dangling delimiter.

## Model Used

Claude Opus 4.6 (`claude-opus-4-6`), extended thinking, with tool use and code execution. Subagents on Claude Sonnet 4.5 were used for test-suite runs and for read-only code exploration.

## Checklist


ericdfields:
Thanks — the P1 is real and I have fixed it in 1575d62.

`writeFile`'s `mode` option applies only when it creates the file. On the repair path the managed `config.toml` already exists, and a home seeded by the *previous* copy-the-host-file behaviour inherited the host's mode — often `0644`. So repairing one of those homes rewrote the content and left the managed MCP `Authorization: Bearer` header group- and world-readable. That is precisely the population this PR's repair pass exists to touch, so the finding lands on the highest-risk path.

Added `await fs.chmod(target, 0o600)` after the write, matching the existing pattern in `writeManagedCodexMcpConfig` and `stageCodexHomeEntry`.

Added a regression test that seeds an existing managed config at `0644` and asserts the repair leaves it `0600`. My earlier mode test only covered the *create* path, which is why this slipped through. Mutation-checked: removing the `chmod` fails the new test.

ericdfields:
Self-review after the mode fix turned up two more defensive gaps in the same function, pushed as `b1a7977`:

- **Same-home guard compared resolved strings.** `$CODEX_HOME` and `~/.codex` can be symlinks, so a managed target that aliases the host home would slip past `path.resolve(a) === path.resolve(b)` and the seed would write over the user's own config — the one thing this function must never do. Now compares real paths, falling back to the resolved string when a path does not exist yet (the managed home usually does not, and in that case it cannot be the host home either).
- **A non-regular file at the managed config path was read through and written over.** A directory or a user-placed symlink there is not Paperclip-written state. It is now left alone, matching the policy `ensureSymlink` already applies to a directory in a managed home. This also removes an `EISDIR` throw that the previous copy-if-absent seeding never raised.

Both are mutation-checked: reverting either guard fails the new tests. 477 tests pass across `packages/adapters/codex-local`, `codex-config-seed`, `acpx-engine/execute` and `codex-local-execute`.

PR #9689, opened 2026-07-16
fix(routines): clear transient execution failures

cryppadotta:
## Thinking Path

> - Paperclip is the open source app people use to manage AI agents for work
> - Scheduled routines track each dispatch in `routine_runs` and link it to an execution issue
> - Moving an execution issue to `blocked` or `cancelled` correctly records a failed run state for operator visibility
> - When that issue later resumes or completes, the run can retain the earlier failure reason and completion timestamp
> - That stale state makes an active or successfully completed routine appear failed
> - This pull request reconciles the run back to a live state on resume and preserves cleared failure details as completion context
> - The benefit is that routine run status consistently reflects the current execution issue lifecycle without losing useful recovery history

## Linked Issues or Issue Description

Refs #9201

### What happened?

A routine execution issue that temporarily moved to `blocked` or `cancelled` caused its linked routine run to become `failed`. If the issue later returned to an active status or reached `done`, the routine run could keep the stale failure reason and terminal timestamp.

### Expected behavior

Active execution issues should have an `issue_created` run with no failure or completion timestamp. Completed execution issues should have a `completed` run with no active failure reason, while retaining any earlier transient failure in structured trigger context for diagnosis.

### Steps to reproduce

1. Create a routine run linked to a routine execution issue.
2. Move the issue to `blocked` and synchronize the run state.
3. Move the issue back to `in_progress` or forward to `done` and synchronize again.
4. Observe that the run previously retained stale failed-state fields.

### Environment

- Reproduced on `master` at `da549123cc`.
- Core server behavior; not adapter-specific.
- Covered with the embedded PostgreSQL routines service test harness.

## What Changed

- Load the linked routine run while synchronizing execution issue status.
- Restore transiently failed runs to `issue_created` when their execution issue resumes active work.
- Clear stale failure state when an execution issue completes and retain the earlier failure under `triggerPayload.transientFailure`.
- Add regression coverage for both resumed and completed execution issues.

## Verification

- `pnpm --filter @paperclipai/server exec vitest run src/__tests__/routines-service.test.ts` — 57 tests passed.
- `pnpm --filter @paperclipai/server typecheck` — passed.

## Risks

- Low risk: the change is limited to routine execution issue/run reconciliation.
- A completed run now stores a prior failed-state reason as structured transient context instead of leaving `failureReason` populated.
- No schema, migration, API contract, or UI behavior changes are included.

[...]

PR #11459, opened 2026-08-16
fix(routines): stop recording a covered delegation wait as a failed run

YukiCrisp:
## Thinking Path

> - Routines fire on a schedule and track each fire as a `routine_runs` row linked to one execution issue
> - Some routines delegate: the execution issue opens child issues, hands them to other agents, and waits in `blocked` until they land
> - `syncRunStatusForIssue` settled any `blocked` execution issue as a failed run, so that designed wait was recorded as a failure
> - Every delegating routine therefore reported a failure on every fire, and real routine failures hid inside that noise
> - This pull request reads `blockerAttention` for the execution issue and keeps a covered wait in flight instead of failing it
> - The benefit is that `routine_runs.status` again separates routines that are waiting by design from routines that are actually broken

Refs #9689 — that pull request keeps the current "blocked means failed" classification and adds recovery of the transient failure state afterwards. This pull request changes the classification itself, so the two overlap in `syncRunStatusForIssue`. If #9689 lands first, this change rebases on top of it and its two assertions of `failureReason: "Execution issue moved to blocked"` need the new covered/attention split.

**What happened?**

A monthly routine opens two child issues, delegates them, and parks its execution issue in `blocked` until both are done. The routine description states that wait. Both fires of that routine recorded:

```
routine_runs.status         = "failed"
routine_runs.failure_reason = "Execution issue moved to blocked"
```

Both fires actually succeeded. The delegated children reached `done` and the execution issue reached `done`.

The failure is also durable. When the execution issue later reached `done`, `finalizeRun` wrote `status = "completed"` but left `failure_reason` untouched. The stored rows now read `completed` with a failure reason attached, which is self-contradictory.

**Expected behavior**

A `blocked` execution issue that waits on live delegated work is the routine working as designed. It must not settle the run as failed. The run must stay in flight and settle when the execution issue reaches a terminal state.

A `blocked` execution issue that nobody is driving must still settle the run as failed, because that is a real routine failure. `blockerAttention.state` already draws that line: `covered` means an active waiting path exists, while `needs_attention` and `stalled` mean the wait needs a human. A blocked issue with no live blocker at all already reports `needs_attention`, so an empty-blocker zombie still fails.

A run that settles as `completed` must not keep an earlier failure reason.

**Steps to reproduce**

1. Create a routine whose execution issue opens a child issue and delegates it to another agent.
2. Let the child issue start. It now has a running heartbeat run, so the parent's `blockerAttention.state` is `covered`.
3. Move the execution issue to `blocked` to wait for the child.
4. Read the linked `routine_runs` row. Before this change it reads `status = "failed"`, `failure_reason = "Execution issue moved to blocked"`.
5. Complete the child, then move the execution issue to `done`.
6. Read the row again. It reads `status = "completed"` with the failure reason still set.

`master` at e52b8a343.

**Deployment mode**

Local single-instance server with embedded PostgreSQL.

## What Changed

- `syncRunStatusForIssue` splits the old `blocked || cancelled` branch. `cancelled` still settles the run as failed with an explicit reason.
- The `blocked` branch loads `blockerAttention` through `issueService.listBlockerAttention` for the execution issue. State `covered` returns without settling the run, so the run stays `issue_created` and settles later.
- Any other blocked state settles the run as failed. `blockedExecutionFailureReason` names the state and a sample blocker, so the run row says which dependency needs attention instead of only "moved to blocked".
- The `done` branch writes `failureReason: null` next to `status: "completed"`, so a recovered run no longer keeps a stale failure string.
- Four regression tests in `routines-service.test.ts` cover: a covered delegation wait staying in flight and later completing; a driverless block still failing; a stale failure reason cleared on completion; and `cancelled` still failing.

## Verification

```
pnpm --filter @paperclipai/server exec vitest run src/__tests__/routines-service.test.ts
```

Both pass. The full file is 65 tests, 65 passed, including the four new ones. The first new test fails on `master` because `syncRunStatusForIssue` returns the settled failed run instead of `null`.

## Risks

- Behavioral shift for operators who read `routine_runs`: a delegating routine now shows an in-flight run for the whole delegation window instead of a failed run. That is the intent, but dashboards that count in-flight runs will see longer-lived rows.
- The blocked state is evaluated once, at the moment the issue moves to `blocked`. If a covered wait later stalls, no issue status change fires and the run is not re-evaluated, so it stays in flight. The blocked-issue attention surface already reports that stall on its own channel, so the stall is not lost, but it does not appear as a run failure.
- The block branch now runs a blocker-graph query on each execution-issue status change. `listBlockerAttention` is bounded by depth and node caps and is already used on issue list paths.

## Model Used

Claude (Anthropic), `claude-opus-5`, extended thinking enabled, run through the Claude Agent SDK with file editing and shell tool use.

## Checklist

### DIFF
--- a/server/src/__tests__/routines-service.test.ts
+++ b/server/src/__tests__/routines-service.test.ts
@@ -2593,4 +2593,92 @@ describeEmbeddedPostgres("routine service live-execution coalescing", () => {
 
     expect(run).toMatchObject({ source: "webhook", status: "issue_created" });
   });
+
+  async function seedDelegatingExecution(opts?: { withCoveredChild?: boolean }) {
+    const fixture = await seedFixture();
+    const { agentId, companyId, routine, svc } = fixture;
+    const run = await svc.runRoutine(routine.id, { source: "schedule" });
+    const executionIssueId = run.linkedIssueId!;
+
+    let childIssueId: string | null = null;
+    if (opts?.withCoveredChild) {
+      childIssueId = randomUUID();
+      const childRunId = randomUUID();
+      await db.insert(issues).values({
+        id: childIssueId,
+        companyId,
+        identifier: `${routine.title.slice(0, 3).toUpperCase()}-CHILD`,
+        title: "delegated child",
+        status: "in_progress",
+        priority: "medium",
+        parentId: executionIssueId,
+        assigneeAgentId: agentId,
+        originKind: "manual",
+      });
+      await db.insert(heartbeatRuns).values({
+        id: childRunId,
+        companyId,
+        agentId,
+        status: "running",
+        contextSnapshot: { issueId: childIssueId },
+      });
+      await db.update(issues).set({ executionRunId: childRunId }).where(eq(issues.id, childIssueId));
+    }
+
+    await db.update(issues).set({ status: "blocked" }).where(eq(issues.id, executionIssueId));
+    return { ...fixture, childIssueId, executionIssueId, run };
+  }
+
+  const readRun = (runId: string) =>
+    db.select().from(routineRuns).where(eq(routineRuns.id, runId)).then((rows) => rows[0] ?? null);
+
+  it("keeps a delegating routine run in flight while the execution issue waits on a covered blocker", async () => {
+    const { childIssueId, executionIssueId, run, svc } = await seedDelegatingExecution({ withCoveredChild: true });
+
+    expect(await svc.syncRunStatusForIssue(executionIssueId)).toBeNull();
+    expect(await readRun(run.id)).toMatchObject({
+      status: "issue_created",
+      failureReason: null,
+      completedAt: null,
+    });
+
+    await db.update(issues).set({ status: "done" }).where(eq(issues.id, childIssueId!));
+    await db.update(issues).set({ status: "done" }).where(eq(issues.id, executionIssueId));
+    await svc.syncRunStatusForIssue(executionIssueId);
+
+    expect(await readRun(run.id)).toMatchObject({ status: "completed", failureReason: null });
+  });
+
+  it("fails a routine run when the execution issue is blocked with nothing driving the block", async () => {
+    const { executionIssueId, run, svc } = await seedDelegatingExecution();
+
+    await svc.syncRunStatusForIssue(executionIssueId);
+
+    const settled = await readRun(run.id);
+    expect(settled?.status).toBe("failed");
+    expect(settled?.failureReason).toContain("needs attention");
+  });
+
+  it("clears a stale failure reason when the execution issue later completes", async () => {
+    const { executionIssueId, run, svc } = await seedDelegatingExecution();
+    await svc.syncRunStatusForIssue(executionIssueId);
+    expect((await readRun(run.id))?.status).toBe("failed");
+
+    await db.update(issues).set({ status: "done" }).where(eq(issues.id, executionIssueId));
+    await svc.syncRunStatusForIssue(executionIssueId);
+
+    expect(await readRun(run.id)).toMatchObject({ status: "completed", failureReason: null });
+  });
+
+  it("still fails a routine run when the execution issue is cancelled", async () => {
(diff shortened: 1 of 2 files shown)


YukiCrisp:
Independent inspection (reviewer, not the author). **Verdict: content PASS.** Notes below are evidence and two non-blocking observations.

### Reproduced

- `pnpm --filter @paperclipai/server exec vitest run src/__tests__/routines-service.test.ts` → **65 passed**, Node 22.
- All CI gates green on `465608a2b` (build, typecheck, 5 server shards, 5 serialized shards, 3 e2e shards, policy, review, Greptile, Socket, Snyk).

### Counter-proof that the new coverage is pinned to the change

Restoring only `server/src/services/routines.ts` to `master` and keeping the new tests:

```
× keeps a delegating routine run in flight while the execution issue waits on a covered blocker
  Tests  1 failed | 64 skipped (65)
```

So the first test genuinely fails without the fix; it is not vacuous.

### The premise the fix rests on holds

The change is only safe if a `blocked` issue with no live blocker cannot classify as `covered`. It cannot — `listIssueBlockerAttentionMap` short-circuits before classification:

```ts
// server/src/services/issues.ts:2749
if (topLevelEdges.length === 0) {
  attentionMap.set(root.id, createIssueBlockerAttention({
    state: "needs_attention", reason: "attention_required", terminalBlockerIssueId: root.id,
  }));
  continue;
}
```

A `blocked` issue whose blockers are all `done`, or which has no blockers at all, still settles the run as `failed`. That is the case worth protecting and it is protected.

The `attention === null` fallback is also conservative: if `listBlockerAttention` returns no entry for the issue, `blockedExecutionFailureReason(null)` yields the old `"Execution issue moved to blocked"` and the run still fails.

### Three collateral checks

1. **Input shape.** The widened `select` in `syncRunStatusForIssue` matches `IssueBlockerAttentionInputNode` field-for-field (`id`, `companyId`, `parentId`, `identifier`, `title`, `status`, `assigneeAgentId`, `assigneeUserId`, `executionRunId`). No implicit `undefined` reaching the classifier.
2. **No dispatch regression from longer-lived runs.** Concurrency is keyed on `findLiveExecutionIssue` (issue-level) in `dispatchRoutineRun`, not on run status, so a run that stays `issue_created` for the delegation window does not skip or coalesce later scheduled fires.
3. **No consumer parses `failureReason`.** The one place that interpolates it (`services/pipelines.ts:3008`) reads a run immediately after dispatch, guarded by `!run.linkedIssueId`, so it cannot reach the new blocked-branch message. The richer strings are display-only.

### Observation 1 — Risk 2 is real and has no run-level safety net (non-blocking)

The PR body is right that a covered wait that later stalls is never re-evaluated, because nothing calls `syncRunStatusForIssue` outside the two issue-status paths. Worth adding: `routine_runs` has **no staleness sweep either** — outside `services/routines.ts`, the table is only read for `responsibleUserId`/`routineRevisionId` and deleted on company teardown. So such a run stays `issue_created` indefinitely with nothing to reap it.

That is still a better trade than the status quo (a 100% false-positive failure on every delegating routine, which is what motivated this), and the stall remains visible on the blocked-issue attention surface. But it converts a guaranteed false positive into a possible silent false negative at the run level, so it deserves a follow-up rather than being left implicit — either re-sync on blocker-graph transitions, or a periodic sweep over non-terminal `routine_runs`.

### Observation 2 — the #9689 overlap is one line, not a duplicate (non-blocking)

#9689 keeps the "blocked means failed" classification and adds recovery afterwards; this PR changes the classification. They are complementary, and they are genuinely additive in one direction: #9689 restores a stale `failed` run to `issue_created` on resume, which this PR does not do — after this change a run that failed while uncovered and later becomes covered stays `failed` until the issue reaches `done`.

The only true overlap is `failureReason: null` on the `done` branch (plus the third test). The rebase note in the description already states the conflict correctly. Rejecting this as a duplicate would drop the covered/attention split, which is the substance and is absent from #9689.

PR #8251, opened 2026-06-18
Fix blocker summaries for legacy child blockers

iceFusion101:
## Thinking Path

> - Paperclip blocked-queue rows show two related signals: blocker summaries (`blockedBy` / `blockedByIssueIds`) and `blockerAttention`.
> - CON-3539 reproduced legacy blocked issues where `blockerAttention` saw a blocker, but blocker summaries were empty or misleading.
> - The source mismatch was not mutation persistence: explicit `issueRelations` were persisted correctly.
> - The mismatch was projection logic: `blockerAttention` traversed direct non-terminal child issues as blocker edges, while blocker summaries only projected explicit `blocks` relations.
> - This caused operators to see `needs_attention` without the child blocker that caused it.
> - The fix makes blocker summaries include the same active direct-child blocker edges, while preserving explicit dependency semantics.
> - Review hardening also keeps cancelled direct children out of child-derived blocker edges and adds tenant scoping to joined relation projections.

Issue context: Paperclip CON-3539 investigates `blockedByIssueIds` / `blockerAttention` mismatches on legacy blocked issues under CON-3532. Affected examples were CON-1494, CON-1526, and CON-997.

Dedup search performed: `gh pr list --repo paperclipai/paperclip --state all --search "blockerAttention blockedBy child blockers" --json number,title,state,url --limit 20`.

Related open PR #7577 covers a narrower cancelled-child `blockerAttention` case. This PR is still needed because it fixes blocker summary/read projection parity and relation-summary de-duplication for CON-3539.

## What Changed

- Added active direct-child blockers to batched `blockedBy` summaries used by issue lists.
- Added active direct-child blockers to `getRelationSummaries` for issue detail/read APIs.
- De-duplicated child blockers when the same issue is also present through an explicit blocked-by relation.
- Excluded cancelled children from child-derived blocker edges while leaving explicit cancelled dependencies visible.
- Added `issues.companyId` guards on joined relation-summary queries.
- Added regressions for list-row attention parity, relation summaries, cancelled child exclusion, and child/explicit de-duplication.

## Verification

Run in the PR worktree:

```bash
pnpm vitest run server/src/__tests__/issue-blocker-attention.test.ts server/src/__tests__/issues-service.test.ts --testNamePattern "non-done child|active child|deduplicates child blockers|same non-done child|persists blocked-by relations"
pnpm --filter @paperclipai/server typecheck
```

Both passed.

## Risks

[...]

PR #11460, opened 2026-08-16
fix(issues): say whether a counted blocker came from blockedBy or a child

YukiCrisp:
## Thinking Path

> - A blocked issue carries a `blockerAttention` summary that says whether the wait is covered, stalled, or needs a human
> - `listIssueBlockerAttentionMap` builds that summary from two kinds of edge: explicit `blocks` relations, and open child issues
> - The API returns only the explicit relations as `blockedBy`, so the child edges are counted but invisible to the caller
> - An operator who compares `blockerAttention` with `blockedBy` sees a contradiction and cannot tell which field to trust
> - This pull request keeps the graph semantics and labels the source of every counted blocker
> - The benefit is that a reader can reconcile the two fields instead of guessing which one is wrong

Refs #8251 — that pull request reports the same mismatch and fixes it from the other side: it widens the blocker **summaries** so `blockedBy` also projects active direct-child edges. It has been open since 2026-06-18 with no code change since, and it is `CONFLICTING` against current `master`, so its green checks are stale. This pull request takes the other option and leaves `blockedBy` alone.

Two reasons to prefer labelling over widening, offered for the reviewer rather than as a claim that #8251 is wrong:

- `blockedBy` is the read side of the writable `blockedByIssueIds` field. A caller that reads `blockedBy` and writes the ids back is a normal round trip, and if child-derived entries appear in that read, the round trip silently promotes a parent/child link into an explicit `blocks` relation.
- Callers still cannot tell an explicit dependency from a child wait after the widening, because both arrive in one list. The mismatch goes away, but the distinction that made the counts confusing is still not expressed.

If maintainers prefer #8251's direction, the `kind` labelling here still composes with it: keep `dependencyBlockerCount` / `childBlockerCount` / `sampleBlockerEdgeKind` and drop nothing.

**What happened?**

A blocked issue reported `blockerAttention` counts that do not match its own `blockedBy` list.

```
blockedBy = [ A (done), B (done) ]
blockerAttention = {
  state: "needs_attention",
  unresolvedBlockerCount: 1,
  attentionBlockerCount: 1,
  sampleBlockerIdentifier: "C"
}
```

`C` is not in `blockedBy`. Reading `C` back shows `blocks: []`, so by the relation graph it blocks nothing.

The counts are not wrong. `C` is an open **child** of the blocked issue, and `listIssueBlockerAttentionMap` treats an open child as a wait on purpose — a parent that delegates to children and waits is covered by that path, and a regression test already depends on it. Both explicit blockers were `done`, so the child was the only live edge left, and it became the whole summary.

The problem is that the payload merges the two edge kinds into one count and one sample and never says which kind it used. `blockedBy` exposes only one of the two kinds, so the two fields disagree with no way to reconcile them. A reader can misdiagnose a healthy wait as a zombie, or miss a wait that has really stopped.

**Expected behavior**

`blockerAttention` should keep counting open children, because that wait is real. It should also state where each counted blocker came from, so a caller can line the summary up against `blockedBy` and see that the difference is child edges rather than a corrupt graph.

**Steps to reproduce**

1. Create issue P and move it to `blocked`.
2. Create child issues A and B under P, and add explicit `blocks` relations from A and B to P.
3. Create child issue C under P. Do not give it any `blocks` relation.
4. Move A and B to `done`. Leave C open and assigned, with no active run.
5. Read P. `blockedBy` lists only A and B, both `done`, while `blockerAttention` reports `unresolvedBlockerCount: 1` and samples C.

`master` at e52b8a343.

**Deployment mode**

Local single-instance server with embedded PostgreSQL.

## What Changed

- `IssueBlockerAttentionEdge` carries a `kind` of `blocked_by` or `child`, set where the explicit-blocker rows and the child rows are merged.
- `appendBlockerAttentionEdges` upgrades a duplicate edge to `blocked_by`, so an issue that is both an open child and an explicit dependency reports the edge a caller can also see in `blockedBy`. The upgrade applies on either merge order.
- `IssueBlockerAttention` gains three optional fields: `dependencyBlockerCount`, `childBlockerCount`, and `sampleBlockerEdgeKind`. `dependencyBlockerCount + childBlockerCount` equals the existing `unresolvedBlockerCount`.
- The new `IssueBlockerAttentionEdgeKind` type is exported from `@paperclipai/shared`, and the type carries doc comments that state the two edge kinds and which of them `blockedBy` shows.
- Two regression tests in `issue-blocker-attention.test.ts`: the reported shape reports `childBlockerCount: 1` with `sampleBlockerEdgeKind: "child"`, and a blocker that is both an open child and an explicit dependency reports `blocked_by`.

## Verification

```
pnpm typecheck
pnpm --filter @paperclipai/server exec vitest run src/__tests__/issue-blocker-attention.test.ts
```

Both pass. The attention test file is 26 tests, 26 passed, including the two new ones. Root `pnpm typecheck` covers `server`, `ui`, `cli`, and the plugin packages.

## Risks

- Low risk. No state classification changes, so `state`, `reason`, `unresolvedBlockerCount`, `coveredBlockerCount`, `stalledBlockerCount`, and `attentionBlockerCount` keep their current values. The existing 24 attention tests pass unchanged.
- The three new fields are optional, so existing clients that do not read them are unaffected.

## Model Used

Claude (Anthropic), `claude-opus-5`, extended thinking enabled, run through the Claude Agent SDK with file editing and shell tool use.

## Checklist

### DIFF
--- a/packages/shared/src/index.ts
+++ b/packages/shared/src/index.ts
@@ -951,6 +951,7 @@ export type {
   IssueSubtreeDiagnosticEdge,
   IssueSubtreeDiagnosticsResponse,
   IssueBlockerAttention,
+  IssueBlockerAttentionEdgeKind,
   IssueBlockerAttentionReason,
   IssueBlockerAttentionState,
   IssueReviewAttention,
--- a/packages/shared/src/types/index.ts
+++ b/packages/shared/src/types/index.ts
@@ -604,6 +604,7 @@ export type {
   IssueSubtreeDiagnosticEdge,
   IssueSubtreeDiagnosticsResponse,
   IssueBlockerAttention,
+  IssueBlockerAttentionEdgeKind,
   IssueBlockerAttentionReason,
   IssueBlockerAttentionState,
   IssueReviewAttention,
--- a/packages/shared/src/types/issue.ts
+++ b/packages/shared/src/types/issue.ts
@@ -396,6 +396,14 @@ export type IssueBlockerAttentionReason =
   | "attention_required"
   | null;
 
+/**
+ * How a counted blocker reaches the blocked issue. `blocked_by` is an explicit
+ * `blocks` relation and appears in the issue's `blockedBy` list. `child` is an
+ * open child issue, which the blocker graph also treats as a wait but which
+ * never appears in `blockedBy`.
+ */
+export type IssueBlockerAttentionEdgeKind = "blocked_by" | "child";
+
 export interface IssueBlockerAttention {
   state: IssueBlockerAttentionState;
   reason: IssueBlockerAttentionReason;
@@ -403,8 +411,14 @@ export interface IssueBlockerAttention {
   coveredBlockerCount: number;
   stalledBlockerCount: number;
   attentionBlockerCount: number;
+  /** Counted blockers that come from an explicit `blocks` relation. */
+  dependencyBlockerCount?: number;
+  /** Counted blockers that are open child issues rather than `blockedBy` entries. */
+  childBlockerCount?: number;
   pendingFinalizeBlockerIssueIds?: string[];
   sampleBlockerIdentifier: string | null;
+  /** Which edge the sampled blocker was reached through. */
+  sampleBlockerEdgeKind?: IssueBlockerAttentionEdgeKind | null;
   sampleStalledBlockerIdentifier: string | null;
   /** True when a blocker or one of its open descendants is actively progressing. */
   blockingTreeLive?: boolean;
--- a/packages/shared/src/types/issue.ts
+++ b/packages/shared/src/types/issue.ts
@@ -411,13 +411,18 @@ export interface IssueBlockerAttention {
   coveredBlockerCount: number;
   stalledBlockerCount: number;
   attentionBlockerCount: number;
-  /** Counted blockers that come from an explicit `blocks` relation. */
+  /** Top-level blockers of this issue that come from an explicit `blocks` relation. */
   dependencyBlockerCount?: number;
-  /** Counted blockers that are open child issues rather than `blockedBy` entries. */
+  /** Top-level blockers of this issue that are open child issues rather than `blockedBy` entries. */
   childBlockerCount?: number;
   pendingFinalizeBlockerIssueIds?: string[];
   sampleBlockerIdentifier: string | null;
-  /** Which edge the sampled blocker was reached through. */
+  /**
+   * The edge that reaches `sampleBlockerIdentifier`. The sample can sit deeper
+   * in the chain than the counted top-level blockers, so this describes the
+   * sampled issue's own incoming edge and does not have to agree with
+   * `dependencyBlockerCount` / `childBlockerCount`.
+   */
   sampleBlockerEdgeKind?: IssueBlockerAttentionEdgeKind | null;
   sampleStalledBlockerIdentifier: string | null;
   /** True when a blocker or one of its open descendants is actively progressing. */
--- a/server/src/__tests__/issue-blocker-attention.test.ts
+++ b/server/src/__tests__/issue-blocker-attention.test.ts
@@ -371,6 +371,40 @@ describeEmbeddedPostgres("issue blocker attention", () => {
     });
   });
 
+  it("labels a sampled descendant with its own edge kind, not the top-level one", async () => {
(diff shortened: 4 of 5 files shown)


YukiCrisp:
Greptile P1 addressed in f847d0638.

The finding is right: `sampleBlockerIdentifier` is produced by `classifyPath`, which can forward a sample from further down the chain, while `sampleBlockerEdgeKind` was read off the root's top-level edge. For `root --blocked_by--> A --child--> B` the API reported B as `blocked_by` even though B is not in `blockedBy` — exactly the contradiction this PR exists to remove.

Fix: `classifyPath` now takes the kind of the edge it was reached through and returns it beside the sample. When it forwards a downstream sample it forwards that node's kind with it. The root falls back to its own top-level edge kind only in the case where the sample itself falls back to the top-level node.

The counts stay top-level by definition, so `sampleBlockerEdgeKind` can legitimately disagree with `dependencyBlockerCount` / `childBlockerCount` when the sample is a leaf. That is now stated in the type doc rather than left implicit.

Added `labels a sampled descendant with its own edge kind, not the top-level one` as a regression test for the exact chain in the finding. It fails without the fix. `issue-blocker-attention.test.ts` is 27 tests, 27 passed; root `pnpm typecheck` green.

YukiCrisp:
Independent inspection (reviewer, not the author). **Verdict: content PASS.** One naming nit, non-blocking.

### Reproduced

- `pnpm --filter @paperclipai/shared build` → clean.
- `vitest run src/__tests__/issue-blocker-attention.test.ts src/__tests__/issues-service.test.ts src/__tests__/attention-service.test.ts` → **169 passed**, Node 22.
- `issue-blocker-attention.test.ts` alone, verbose: **27 passed** = the 24 pre-existing tests unmodified + the 3 new ones. That is the load-bearing evidence for "state / reason / existing counts unchanged", and it holds.
- All CI gates green on `f847d0638`.

(Minor: the description says 196 for the three-file run; I measure 169. Same three files, all green — just noting the number so it is not carried forward.)

### Targeted counter-proof of the Greptile P1 fix

Reverting the whole service file only proves the new fields are absent, which is uninteresting. Instead I reverted **only the edge-kind threading**, keeping every other line of the PR:

```diff
-const classified = downstream.map((edge) => classifyPath(edge.blockerIssueId, nextSeen, edge.kind));
+const classified = downstream.map((edge) => classifyPath(edge.blockerIssueId, nextSeen, edgeKind));
```

```
× labels a sampled descendant with its own edge kind, not the top-level one
- Expected  "sampleBlockerEdgeKind": "child"
+ Received  "sampleBlockerEdgeKind": "blocked_by"
```

That reproduces the reported P1 exactly (`root --blocked_by--> A --child--> B` reporting `B` as `blocked_by`) and shows the regression test is pinned to the one line that fixes it, not to the feature as a whole.

### `appendBlockerAttentionEdges` does not perturb sample selection

Old behaviour: first-seen edge is pushed, later duplicates skipped. New behaviour: first-seen edge is pushed, later duplicates `continue` after possibly upgrading `kind`. Array contents and insertion order are identical in both, so `topLevelEdges` ordering — and therefore `classified[0]`, `sampleEntry`, and the fallback sample — is unchanged. The `blocked_by` upgrade is order-independent as the comment claims, and the mutated objects come from fresh `.map(...)` results, so nothing aliases back into the query rows.

The `firstDownstream` extraction is likewise behaviour-preserving for `sampleBlockerIdentifier`: `classified[0]?.sampleBlockerIdentifier ?? nodeSample` and `firstDownstream?.sampleBlockerIdentifier ?? nodeSample` agree on every input, since `firstDownstream` is null exactly when `classified[0]`'s sample is falsy.

### On "the sample kind need not agree with the top-level counts"

Documenting it on the type is the right call and, given no consumer renders the field yet, proportionate. Two things make it stick better than a comment alone: the counts are a strict partition (`dependencyBlockerCount + childBlockerCount === unresolvedBlockerCount`, including the both-child-and-dependency case, which dedup collapses to a single `blocked_by` edge — covered by the `PBF` test), and the descendant case has its own regression test.

**Nit, non-blocking:** `sampleBlockerEdgeKind` parses most naturally as "the kind of the sampled *top-level blocker's* edge", which is precisely the reading the doc comment then has to walk back. `sampleBlockerIncomingEdgeKind` would carry the semantics in the name and leave the doc as background rather than as a correction. Not worth a rev on its own.

PR #11463, opened 2026-08-16
feat: sandbox provider capability contract with fail-closed effective resolution

nickyleach:
## Thinking Path

> - Paperclip runs work through adapters and sandbox providers
> - Providers need a clear contract so the server can use only verified capabilities
> - A declared capability must not grant a method that the live worker did not verify
> - This pull request adds manifest declarations and fail-closed effective capability resolution
> - The benefit is safe provider reuse across execution targets and run lifecycles

**Subsystem affected**

Cross-cutting (multiple of the above)

**Problem or motivation**

Sandbox providers expose different runtime methods. The server needs one safe capability contract that accounts for provider declarations, worker verification, and narrowing configuration.

**Proposed solution**

Add strict manifest validation for five sandbox capabilities. Resolve effective capabilities as the subset of verified, declared, and narrowed values. Store the result as a frozen execution-target snapshot.

**Alternatives considered**

Trusting the manifest alone could grant methods that the worker does not support. Trusting only a fixed built-in list would reject valid third-party providers. The intersection rule keeps the verified runtime ceiling and supports both provider types.

**Roadmap alignment**

This change supports the ACP run lifecycle track and the sandbox provider contract work in the current roadmap.

**Additional context**

The legacy `supportsReusableLeases` field remains supported. The nested capability validator rejects unknown keys. Missing or unavailable verification resolves all capabilities to `false`.

## What Changed

- Add strict `sandboxCapabilities` manifest validation with legacy reusable-lease compatibility.
- Carry declarations through the ready-driver projection.
- Add fail-closed effective resolution from verified, declared, and narrowed capabilities.
- Add narrowing for provider configuration, Kubernetes Job leases, and Daytona sessions.
- Add a frozen read-only capability snapshot to execution targets.
- Add focused tests and keep existing characterization baselines covered.
- Add and update sandbox provider capability documentation.

## Verification

- `npx vitest run packages/shared/src/validators/plugin.test.ts`
- `npx vitest run server/src/__tests__/plugin-environment-driver-sandbox-capabilities.test.ts`
- `npx vitest run server/src/__tests__/sandbox-capability-contract.test.ts`
- `npx vitest run server/src/__tests__/environment-execution-target-capabilities.test.ts`
- `npx vitest run packages/adapter-utils/src/acpx-engine/startup-characterization.test.ts packages/adapter-utils/src/acpx-engine/turn-characterization.test.ts packages/adapter-utils/src/acpx-engine/settlement-characterization.test.ts packages/adapter-utils/src/acpx-engine/composed-run-characterization.test.ts`
- Package typechecks for shared, server, and adapter-utils pass.
- Stage-2 security review suites pass with 28 tests.

## Risks

The resolver fails closed when verification is absent or unavailable. Providers that rely on undeclared capabilities may see narrower behavior until they expose verified worker methods. The change does not alter the existing native-sync guard.

## Model Used

OpenAI Codex, GPT-5, exact runtime model ID `gpt-5`, tool use and code execution. The implementation author used this model to assist with the change.

## Checklist


nickyleach:
@greptile review

server/src/services/environment-runtime.ts
```diff
@@ -1437,6 +1640,67 @@ function createSandboxEnvironmentDriver(
...
+      let config: Record<string, unknown> = {};
+      let configResolutionFailed = false;
+
+      if (metadata.sandboxProviderPlugin) {
+        const pluginId = readString(metadata.pluginId);
+        verifiedMethods =
+          (pluginId ? pluginWorkerManager?.getWorker(pluginId)?.supportedMethods : undefined) ?? [];
+        // Read the manifest declaration even when the worker is not running: an
+        // unverified declaration still resolves to false through the normalizer.
+        const resolvedDriver = await resolvePluginSandboxProviderDriverByKey({
```
nickyleach: [P1] Resolve the declaration from the lease plugin, not the first matching driver key

`verifiedMethods` is pinned to `metadata.pluginId`, but this lookup is only by `providerKey`. `resolvePluginSandboxProviderDriverByKey(... requireRunning: false)` returns the first installed plugin with that driver key, including a disabled one. Driver keys are only unique within a manifest, so an earlier disabled plugin can collide with the running plugin that acquired this lease. In that case the resolver intersects worker methods from plugin B with the declaration from plugin A. If B declares a capability `false` while A omits it or declares `true`, the effective result grants a capability that the actual provider denied.

Please resolve the manifest from `metadata.pluginId`, verify that exact plugin still declares this `providerKey`, and fail closed when that identity cannot be established. A regression test should cover two plugins with the same driver key where the first is not ready and the second owns the lease.

nickyleach:
Blocking merge-readiness feedback for the public capability contract:

The original nested-over-legacy API mismatch in this comment was fixed on `83bd3bd`. The remaining documentation and concurrency findings below are merge-blocking, not optional follow-ups.

### [Blocking] Document the complete public contract

This PR adds seven public manifest fields but no plugin documentation describes them. The PR checklist also says documentation was updated. Code comments and tests are not sufficient documentation for plugin authors.

**Required documentation:** update `doc/plugins/PLUGIN_SPEC.md` or the canonical sandbox-provider contract and cover all of the following:

- The seven supported keys: `reusableLeases`, `nativeSyncIn`, `nativeSyncOut`, `concurrentSyncAndExec`, `concurrentSyncOperations`, `persistentProcessSessions`, and `independentControlCommands`.
- The declaration is optional and partial. For a valid, identified provider, an omitted key defers to verified worker discovery; `false` narrows support; `true` never grants support without verified prerequisites.
- Nested `sandboxCapabilities.reusableLeases` overrides legacy `supportsReusableLeases` when both exist.
- The worker-method prerequisites for every capability, including both lifecycle methods required for reusable leases.
- Target/config narrowing and failure behavior: ephemeral lease policy, job/native-sync restrictions, `useSessions`, config-resolution failure, and exact-plugin identity failure.
- The actual concurrency behavior chosen in the next section. Documentation must not describe either concurrency flag as enforced unless production code enforces it.

**Acceptance:** a third-party provider author can determine exactly what to declare, which worker methods must be advertised, what omission means, and when the host will narrow or deny each capability without reading server source.

### [Blocking] Do not ship no-op concurrency capabilities

`concurrentSyncAndExec` and `concurrentSyncOperations` are currently validated, resolved, serialized, and tested only as data. No production scheduling or synchronization path reads either flag. A provider can declare `false` with no behavioral effect. That makes the public contract misleading and leaves dead capability code in the release.

Choose one complete outcome for this PR:

1. **Enforce them:** identify every production point that can overlap sync with execution or inbound sync with outbound sync, and gate that overlap on the effective snapshot. A denied capability must serialize the relevant operations. Add deterministic tests using deferred promises/barriers that prove operations do not overlap when `false` and preserve allowed overlap when `true`.
2. **Remove them:** if the runtime does not currently expose those concurrency choices, remove both keys from the public type, strict validator, capability key list, prerequisite mapping, effective snapshot, parser, and tests. They can be reintroduced when an enforcing consumer exists.

Document whichever outcome is implemented. Labeling the fields reserved or descriptive is not sufficient for this PR.

These two sections are fixed acceptance criteria. Together with the current inline P1 and P2 comments, they define the complete blocking scope for the next review.

server/src/services/environment-runtime.ts
```diff
@@ -1437,6 +1641,75 @@ function createSandboxEnvironmentDriver(
...
+        // Read the declaration from the exact plugin that acquired the lease,
+        // not the first installed plugin with this driver key. A driver key is
+        // only unique inside one manifest, so two plugins can share it. The
+        // by-key resolver could intersect this lease's verified methods with a
+        // different plugin's declaration. The resolver reads the manifest even
+        // when the worker is not running: an unverified declaration still
+        // resolves to false through the normalizer. The runtime fails closed
+        // when the pinned plugin id is absent, or when that exact plugin no
+        // longer declares this provider key; `declared` stays null.
+        const resolvedDriver = pluginId
```
nickyleach, line 1689: [P1] Make exact-plugin identity failures actually fail closed

The by-ID lookup fixes the cross-plugin collision when the pinned plugin still has this driver. The remaining branch is when `pluginId` names a missing plugin record, or the exact plugin no longer declares this `sandbox_provider` key. `resolvedDriver` is then `null`, but `verifiedMethods` can still contain methods from a running or stale worker entry. Passing `declared: null` is permissive here: `resolveEffectiveSandboxCapabilities` deliberately treats an absent declaration as no restriction and grants the verified baseline. That contradicts the comment above this code.

**Required behavior for this PR:**

- If the exact plugin/driver identity cannot be established, every effective capability must resolve to `false`, regardless of methods advertised by the worker manager. Implement this by clearing `verifiedMethods`, returning an all-false snapshot, or throwing so the execution-target fail-closed path handles it.
- Preserve backward compatibility for a valid exact plugin/driver whose manifest merely omits `sandboxCapabilities`: that case must continue to defer to verified worker discovery. Identity lookup failure and an omitted optional capability declaration are different states and must not both use `declared: null`.

**Regression coverage required:**

1. Pinned plugin id is absent from the registry while a stale worker entry advertises all methods: all capabilities are false.
2. Pinned plugin exists but no longer declares the provider key, or declares it with the wrong kind, while its worker advertises all methods: all capabilities are false.
3. Valid pinned plugin/driver with no `sandboxCapabilities` declaration: verified baseline behavior remains unchanged.

The existing collision regression remains valid; these cases complete the same identity invariant rather than introduce a new requirement.

server/src/routes/environments.ts
```diff
@@ -715,9 +716,15 @@ export function environmentRoutes(
...
-            // execution guard (=== true).
-            supportsReusableLeases: driver.supportsReusableLeases ?? false,
+            // Derive the published value through the same declaration resolver
+            // that acquisition uses, so the nested `sandboxCapabilities`
+            // override wins over the legacy `supportsReusableLeases` flag. A
+            // manifest with legacy `true` and nested `false` must present as
+            // not reusable, because the runtime refuses the reuse. Default an
+            // absent value to false with `=== true`, so the presentation agrees
+            // with the execution guard.
+            supportsReusableLeases:
```
nickyleach, line 729: [P2] Publish verified reusable-lease support, not declaration-only support

This now applies nested-over-legacy precedence correctly, but it still does not match the acquisition guard. Acquisition requires both a positive declaration and live verification of `environmentResumeLease` plus `environmentReleaseLease`. This route returns `true` from the declaration alone. A ready/running worker that omits either method is therefore advertised as reusable even though acquisition always creates an ephemeral lease.

**Required behavior for this PR:**

At the provider-capabilities endpoint, `supportsReusableLeases` is `true` if and only if:

1. `resolveDeclaredSandboxCapabilities(driver).reusableLeases === true`, including the existing nested-over-legacy precedence; and
2. the currently running worker for that exact plugin advertises both `environmentResumeLease` and `environmentReleaseLease`.

It must be `false` when the declaration is absent/false or either verified method is missing. This is provider-level support only; no per-lease `reuseLease` configuration or other capability fields need to be added to this endpoint.

The verification can be computed in `listReadyPluginEnvironmentDrivers`, where the worker manager and exact plugin id are already available, or carried to this route without expanding the public response shape.

**Regression coverage required:**

- Positive declaration plus both lifecycle methods: API returns `true`.
- Positive declaration with resume missing: API returns `false`.
- Positive declaration with release missing: API returns `false`.
- Nested `false` still overrides legacy `true`, even when both methods exist.
- Legacy-only `true` remains supported when both methods exist.

These cases complete the existing claim that presentation agrees with acquisition; they do not require exposing all seven effective capabilities.

nickyleach:
Scope lock for this review round, so the acceptance criteria do not move:

1. **P1 — exact-plugin identity failures must fail closed:** https://github.com/paperclipai/paperclip/pull/11463#discussion_r3792494285
2. **P2 — published reusable support must include live lifecycle-method verification:** https://github.com/paperclipai/paperclip/pull/11463#discussion_r3792494873
3. **Blocking documentation — document the complete seven-field provider contract:** https://github.com/paperclipai/paperclip/pull/11463#issuecomment-5308732205
4. **Blocking concurrency — enforce both concurrency flags in production or remove them completely:** https://github.com/paperclipai/paperclip/pull/11463#issuecomment-5308732205

These four items, including the acceptance criteria and regression cases in the linked comments, are the complete blocking scope from this review round. Documentation and concurrency are required for merge; they are not follow-ups.

On the next head, the review should verify implementation against these exact requirements, verify the requested tests, and check only for regressions introduced by the fixes. It should not reinterpret the behavior, accept documentation-only no-op concurrency fields, or add new variants of the same requirements after the listed criteria pass.

doc/plugins/SANDBOX_PROVIDER_CAPABILITIES.md
```diff
@@ -0,0 +1,141 @@
...
+      independentControlCommands: false,
+    },
+  },
+]
+```
+
+Every key is optional. For a valid, identified provider each key has one of three
+states:
+
+- **Omitted** — the host defers to verified worker discovery. The capability is
```
nickyleach, line 61: [P2] Document reusable leases as the explicit-opt-in exception. This blanket omission rule conflicts with the implemented reusable-lease acquisition and provider-capabilities behavior: both require resolveDeclaredSandboxCapabilities(driver).reusableLeases === true, in addition to live verification of environmentResumeLease and environmentReleaseLease. I contributed to this ambiguity: the earlier documentation criterion said omitted keys defer to worker discovery, while the separate locked P2 criterion explicitly required a positive reusable-lease declaration. The implementation correctly follows the P2 criterion. Please keep the general omission semantics for the other capabilities, but document that reusable leases are special: the provider must opt in with nested sandboxCapabilities.reusableLeases: true or legacy supportsReusableLeases: true; omission does not make the provider eligible for reusable-lease acquisition or advertise provider-level reusable support. Both lifecycle methods and per-run narrowing are still required. This is a documentation correction only; no runtime behavior change is requested.

nickyleach:
@greptile review

PR #11473, opened 2026-08-16
refactor(ui): record why sign-out resets rather than removes

tonio-alucema:
## Thinking Path

> - Signing out has to drop the caches that describe the account, or the next account reads them
> - #11380 does that with `resetQueries`, and the choice between reset and remove is not obvious — both fail in different situations
> - Two sessions working on adjacent bugs reached opposite recommendations on that API within a day, each correct about a different case
> - The reasoning for the choice landed in #11380; the evidence behind it, and the note about when the opposite is right, did not
> - This pull request records both in the file, so the choice is not re-litigated from first principles later
> - The benefit is that a later reader who reasonably reaches for `removeQueries` finds out why it was rejected before changing it

Comment only, no behavior change. Salvaged from the branch `claude/sign-out-cache-note`, written while #11380 was in progress and never opened as a PR. That branch is deleted alongside this.

**What happened?**

#11380 landed `resetQueries` with its reasoning, but not the measurement behind it, and not the note explaining that the opposite advice holds elsewhere in the codebase. Both were on a branch that would otherwise have been deleted unread.

**Expected behavior**

The file records why the choice was made and when it does not apply.

`master` at `38752c4e5`.

**Related pull requests**

#11380 landed the sweep this documents. #11382 added the onboarding draft gate named here.

## What Changed

- `ui/src/hooks/useSignOut.ts` — 14 lines of comment beside the existing `resetQueries` call.

Two things it adds:

**The choice was measured.** Against query-core 5.101.4, `removeQueries` produced 0 notifications and left the observer holding the signed-out account's session; `resetQueries` produced 3 and null. The consequence of the former is not only a stale read — CloudAccessGate's redirect fires on the session going empty, so it never runs.

**The opposite advice holds elsewhere.** `InviteLanding` and the onboarding draft gate avoid reset on purpose: it rewinds the update counters `isFetchedAfterMount` derives from while a mounted observer keeps its bind-time baseline, so the flag can never read true again and the gate withholds forever.

Sign-out is not one of those cases, and the reason is now written down rather than inferred: it resets the session too, so those consumers redirect, unmount, and remount with a fresh baseline. I had previously guessed that from the route table, which is a weaker basis than it looked.

## Verification

- `ui` typecheck is clean.
- 13 pass across `useSignOut.test.tsx` and `InstanceGeneralSettings.test.tsx`.

No behavior change, so there is nothing new to test. The claims are checkable against the existing suites and against `AppsConnect.tsx:374-375`, the live `isFetchedAfterMount` consumer.

## Risks

None. Comment only.

Revert the commit to restore.

## Model Used

## Checklist

### DIFF
--- a/ui/src/hooks/useSignOut.ts
+++ b/ui/src/hooks/useSignOut.ts
@@ -83,13 +83,16 @@ export function useSignOut({ onSignedOut }: UseSignOutOptions = {}) {
       // consequence of the former is not just stale reads — the redirect in
       // CloudAccessGate fires on the session going empty, and it never runs.
       //
-      // The opposite advice holds for a *local* reset under an observer that
-      // stays mounted, which is why InviteLanding and the onboarding draft
-      // gate avoid it: reset rewinds the update counters `isFetchedAfterMount`
-      // is derived from while that observer keeps its bind-time baseline, so
-      // the flag can never read true again and the gate withholds forever.
-      // Sign-out is not one of those cases. It resets the session too, so
-      // those consumers redirect, unmount, and remount with a fresh baseline.
+      // The opposite advice holds for a *local* reset of a key an observer is
+      // still mounted against. Reset rewinds the update counters
+      // `isFetchedAfterMount` derives from while that observer keeps its
+      // bind-time baseline, so the flag can never read true again and anything
+      // gated on it withholds forever. `AppsConnect` is the consumer to check
+      // against: it gates its render on that flag for two queries.
+      //
+      // Sign-out is not that case. It resets the session too, so `AppsConnect`
+      // — which sits behind a route — unmounts on the redirect and remounts
+      // with a fresh baseline.
       //
       // Not awaited: the refetches this kicks off are expected to 401 now that
       // the session is gone, and the sign-out button should not sit pending
--- a/ui/src/hooks/useSignOut.ts
+++ b/ui/src/hooks/useSignOut.ts
@@ -77,6 +77,20 @@ export function useSignOut({ onSignedOut }: UseSignOutOptions = {}) {
       // across the whole sign-out/sign-in cycle. Reset notifies them, so the
       // old data is gone from the cache *and* from everything reading it.
       //
+      // Measured against query-core 5.101.4 rather than reasoned about:
+      // removal produced 0 notifications and left the observer holding the
+      // signed-out account's session; reset produced 3 and null. The
+      // consequence of the former is not just stale reads — the redirect in
+      // CloudAccessGate fires on the session going empty, and it never runs.
+      //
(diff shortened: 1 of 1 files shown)


tonio-alucema:
Valid, and careless of me. The note named `InviteLanding` and the onboarding draft gate as the components that avoid `resetQueries` because of `isFetchedAfterMount`. **Neither reads that flag.**

- The onboarding gate stopped reading it in #11382 — `isSuccess` replaced it precisely because `isFetchedAfterMount` does not reject a failed refetch.
- `InviteLanding` never did; it calls `fetchQuery` directly (`InviteLanding.tsx:373`).

I wrote exactly that on #11380 about an hour before opening this, then carried the wording over from the salvaged branch without re-checking it against the file. The hazard the paragraph describes is real; the examples were not.

`AppsConnect` is the only consumer of the flag on `master` (`AppsConnect.tsx:374-375`, `396`, `399`) and is now the one named. That also tightens the last sentence: it sits behind the `apps/connect` route, so it unmounts on the sign-out redirect and remounts with a fresh baseline — which is the specific reason sign-out is safe, rather than a general claim about "those consumers".

Also retitled from `docs:` to `refactor:`. The quality gate does not allow `docs:` with changes under `ui/src`, since it cannot tell a comment from a statement. Still comment-only — the diff is 14 lines, all of them `//`.

`ui` typecheck clean; 13 pass across `useSignOut.test.tsx` and `InstanceGeneralSettings.test.tsx`.

tonio-alucema:
Correcting the provenance recorded in `0023c5c4a`.

The commit says *"Salvaged from `claude/sign-out-cache-note`"*, which is where I found the text but is not where it came from. **The measurement originated in [this comment on #11380](https://github.com/paperclipai/paperclip/pull/11380#issuecomment-5300984911)** — the 0-notifications-versus-3 probe against query-core 5.101.4 — and belongs to the work that landed as `38752c4e5`. Anyone tracing the reasoning should start there, not at a branch that no longer exists.

I also told that PR's author I had found "a branch of yours". I had no basis for that. Every local session on this machine commits as `Tonio <tonework@gmail.com>`, so git cannot distinguish which one authored `claude/sign-out-cache-note`; I inferred it from the subject line and the timing and stated it as fact. They have since confirmed it was not theirs.

`0023c5c4a` carries only `Co-Authored-By: Claude Opus 5`. It should have carried the source of the measurement too. That commit is on master and cannot be amended, so this comment is the record.

Worth noting for #11498, which adds a `Co-Authored-By` gate to commitperclip: **it would not have caught this.** The gate keys on the authors of a PR's commits, and this content arrived from a branch that never became a PR, authored under an identity shared by every session on the machine. Attribution that crosses branches rather than PRs is outside what it can see, and worth saying plainly rather than letting the gate imply more coverage than it has.

PR #11478, opened 2026-08-16
test(ui): make the suite independent of the machine timezone

### LINKED ISSUE #11476: ui test suite depends on the machine timezone (IssueProperties, StatusCards/format)
**What happened?**

Two tests in the `ui` suite depend on the machine's timezone. CI sets `TZ=UTC`, so both are green there, but a contributor in any other zone gets failures that have nothing to do with their change. I hit them while working on unrelated UI work and had to check each one against a clean tree to confirm it was not mine.

Confirmed as timezone, not flake — same machine, same commit, `TZ` the only variable:

```
local TZ: PDT, offset -0700
as-is    Tests  1 failed | 63 passed (64)
TZ=UTC   Tests  64 passed (64)
```

They fail in two different ways.

**1. `ui/src/components/IssueProperties.test.tsx` — hardcoded local-clock strings, fails always outside UTC**

Lines 2346, 2355 and 2379 assert rendered wall-clock times:

```ts
expect(monitorRowText()).toContain("Today, 4:08 PM · Attempt 1");
expect(monitorRowText()).toContain("Today, 4:08 PM");
expect(monitorRowText()).toContain("Today, 1:38 PM · fires on next tick");
```

The component renders a fixed UTC instant in local time, so `16:08 UTC` is `4:08 PM` only at offset 0. At `UTC-7` it renders `9:08 AM` and the assertion fails on every run, at any hour.

Worth noting the fix pattern already exists in the same file — lines 1515-1517 assert the parts that do not move and leave the hour open:

```ts
expect(container.textContent).toMatch(/CreatedApr 6, 2026, \d{1,2}:34 (AM|PM)/);
```

**2. `ui/src/pages/StatusCards/format.test.ts` — local-day helper against a UTC-day filter, fails in a daily window**

`rollupUpdatesToday` (`ui/src/pages/StatusCards/format.ts:55-57`) filters on the **UTC** calendar day, deliberately, to match the server-side daily token cap:

```ts
const startOfDay = new Date(now);
startOfDay.setUTCHours(0, 0, 0, 0);
```

The test helper builds its timestamps at **local** noon, and its comment shows it believes the filter is local-day:

```ts
function iso(daysAgo: number): string {
  const d = new Date();
  d.setDate(d.getDate() - daysAgo);
  // Noon avoids DST/midnight edge cases in the local-day filter.
  d.setHours(12, 0, 0, 0);
  return d.toISOString();
}
```

Once the UTC day has rolled over but the local day has not, `iso(0)` lands in the *previous* UTC day and "today's" rows are filtered out, so `rollup.updateCount` is 0 instead of 1 (`format.test.ts:66`). At `UTC-7` that window is roughly 17:00 to midnight local — about seven hours a day, not a narrow midnight edge. I saw it fail at 23:41 local and pass at 09:37 local on the same commit.

This one is worth a second look beyond the test: the helper's comment asserts a local-day filter and the function implements a UTC-day filter. If the UTC boundary is right — and the cap-matching rationale says it is — then the comment is misleading about the behaviour it guards.

**Expected behavior**

`pnpm vitest run` in `ui` gives the same result regardless of the machine's timezone.

**Steps to reproduce**

1. Set the machine to a negative-offset timezone, e.g. `TZ=America/Los_Angeles`.
2. From `ui`, run `npx vitest run src/components/IssueProperties.test.tsx` — the monitor-row-states case fails at any time of day.
3. Between 17:00 and midnight local, run `npx vitest run src/pages/StatusCards/format.test.ts` — the `rollupUpdatesToday` case fails.
4. Re-run either with `TZ=UTC` prefixed. Both pass.

`master` at `0023c5c4a`.

**Deployment mode**

Not applicable — test-suite only, no runtime behavior involved.

**Agent adapter(s) involved**

**Suggested fix**

Two independent changes; either can land alone.

- `IssueProperties.test.tsx` — relax the three assertions to the pattern already used at lines 1515-1517, matching the minute and leaving the hour and meridiem open.
- `format.test.ts` — build the fixtures on the same boundary the function uses, e.g. anchor `iso()` to UTC (`setUTCHours(12, 0, 0, 0)`), or pass an explicit `now` to `rollupUpdatesToday`, which already accepts one. Passing `now` is the more direct fix, since it removes the dependency on the current time entirely rather than moving it.

Pinning `TZ=UTC` in the vitest config would turn both green, but it would hide the class rather than fix it, and the second one is a genuine local-vs-UTC mismatch worth correcting on its merits.

**Pre-submission checklist**

tonio-alucema:
## Thinking Path

> - Contributors change the UI and run the `ui` vitest suite to check they have not broken anything
> - That signal only works if a failure means the change was wrong, and two tests failed on the machine's timezone instead
> - CI sets `TZ=UTC`, so the project never saw them, while anyone developing outside UTC met a red suite on their first run
> - A red suite that has nothing to do with your change costs time to dismiss and trains people to ignore failures
> - This pull request removes the timezone assumption from every test that carried one
> - The benefit is that a failing `ui` suite means something again, wherever it is run

Fixes: #11476

## What Changed

Seven files, one rule: **anchor each fixture to the clock the code under test actually reads.**

| File | Code under test reads | Fixture now |
| --- | --- | --- |
| `components/IssueProperties.test.tsx` | local (renders + labels "Today") | local |
| `pages/StatusCards/format.test.ts` | UTC day (`setUTCHours`) | UTC, with an explicit `now` |
| `lib/attention.test.ts` | local day (`setHours`) | local |
| `components/SummarySlotCard.test.tsx` | local (dated dropdown) | local |
| `fixtures/issueThreadInteractionFixtures.ts` | local (expiry footer) | local |
| `pages/agent-skills/AgentSkillReleasePicker.test.ts` | local (`getFullYear`/`getMonth`/`getDate`) | local |
| `components/artifacts/ArtifactCard.test.tsx` | local ("Last edited") | local |

`format.test.ts` is the only one anchored to UTC, because it is the only one whose code deliberately works on the UTC day — that boundary matches the server-side daily token cap.

### The two failed for different reasons

**A local-time surface pinned to UTC instants.** `IssueProperties.test.tsx` asserted `"Today, 4:08 PM · Attempt 1"` against fixtures written as fixed UTC strings. The row renders those in the machine's timezone, so the string only appeared at offset 0; at UTC-7 the row reads `9:08 AM` and the case failed on every run, at any hour. The `Today` label has the same problem one level up — it holds only while the scheduled time shares a local calendar day with `now`.

Both follow from anchoring a local-time surface to UTC. Anchoring the fixtures to a fixed local time instead fixes both at once: the pair cannot straddle local midnight, and the rendered clock is the same in every timezone, so `"Today, 4:08 PM · Attempt 1"` is exactly right everywhere and stays asserted verbatim.

**Local-day fixtures against a UTC-day filter, failed in a daily window.** `rollupUpdatesToday` filters on the UTC calendar day, deliberately, to match the server-side daily token cap:

```ts
const startOfDay = new Date(now);
startOfDay.setUTCHours(0, 0, 0, 0);
```

The `iso()` helper built its timestamps at local noon, and its comment described the filter it was feeding as "the local-day filter" — so the helper and the function disagreed about which day boundary applied. Once UTC has rolled over and local has not, `iso(0)` lands in the *previous* UTC day and today's rows are filtered out, so `updateCount` is 0 rather than 1. At UTC-7 that window runs from about 17:00 to local midnight — roughly seven hours a day, not a midnight edge.

Fixtures are now built on the same boundary the function uses, from a fixed instant that is also passed as `now`. That removes the dependency on the current time rather than moving it, and it is the pattern the neighbouring test in the same file already used (`"uses the UTC day boundary used by the server token cap"`).

### Why not pin `TZ=UTC` in the vitest config

It would turn both green in one line, and hide the class. The second case is a real local-versus-UTC mismatch that is worth correcting on its own; pinning the timezone would leave the misleading comment in place and the next fixture would be written the same way.

## Verification

Both files, across ten offsets from UTC-11 to UTC+14:

| TZ | Offset | Result |
| --- | --- | --- |
| `Pacific/Midway` | UTC-11 | 64 passed |
| `America/Los_Angeles` | UTC-7 | 64 passed |
| `UTC` | 0 | 64 passed |
| `Asia/Kolkata` | UTC+5:30 | 64 passed |
| `Asia/Kathmandu` | UTC+5:45 | 64 passed |
| `Asia/Shanghai` | UTC+8 | 64 passed |
| `Asia/Tokyo` | UTC+9 | 64 passed |
| `Australia/Adelaide` | UTC+9:30 | 64 passed |
| `Pacific/Auckland` | UTC+12 | 64 passed |
| `Pacific/Kiritimati` | UTC+14 | 64 passed |

### A correction to the first version of this PR

The first commit relaxed the monitor assertions to a pattern and I reported the case fixed. It was not: the row labels a time `Today` only while it shares a local calendar day with `now`, and with fixed UTC instants that pair straddles local midnight from UTC+8 to UTC+9:30, so the label becomes a date and the assertions still failed. My matrix — UTC, UTC-7, UTC+5:30, UTC+5:45, UTC+12 — stepped over the failing band, which is exactly the shape of verification that reports a fix it has not tested.

Greptile caught it. The second commit anchors the fixtures to a fixed *local* time instead, which keeps `now` and every scheduled time on one calendar day everywhere. That also makes the rendered clock identical in every timezone, so the relaxed assertions are **restored to exact strings** — the precision given up in the first commit was a symptom of anchoring to the wrong clock, not a necessary trade. The risk noted below about losing timestamp precision no longer applies.

Full `ui` suite, every zone tried:

| TZ | Offset | Suite |
| --- | --- | --- |
| `Pacific/Midway` | UTC-11 | 4015 passed |
| `America/Los_Angeles` | UTC-7 | 4015 passed |
| `UTC` | 0 | 4015 passed |
| `Asia/Kathmandu` | UTC+5:45 | 4015 passed |
| `Asia/Tokyo` | UTC+9 | 4015 passed |
| `Pacific/Kiritimati` | UTC+14 | 4015 passed |

`tsc -b` clean.

### Scope grew twice, both times for the same reason

#11476 named two tests. Fixing those and running the full suite at UTC+9 surfaced three more; fixing those and running at UTC+14 surfaced two more still (`AgentSkillReleasePicker.test.ts`, `ArtifactCard.test.tsx`). Each round was found by widening the offset, not by reading more code.

That is the actual lesson of this PR, and one of the fixtures says so out loud. `AgentSkillReleasePicker.test.ts` carried the comment *"Anchored to noon UTC so it stays on the 15th regardless of local offset"* — a claim that is false at UTC+14, written by someone reasoning about offsets rather than running them. The first revision of this PR made the identical mistake with the identical confidence.

### Not fixed here: one genuinely flaky test

`src/components/task-chat/TaskMessageScroller.test.tsx` — "plays the exit animation before unmounting when motion is enabled" — failed once in a full run at `Asia/Tokyo`. It is **not** timezone-dependent: it passes 3/3 in isolation at both `Asia/Tokyo` and `UTC`. It is an animation-timing flake that surfaces under parallel load, a different problem from this PR, and left alone rather than folded in.

Before this change, on the same machine, the suite was `1 failed | 63 passed` at UTC-7 and `64 passed` at `TZ=UTC` — timezone as the only variable.

**One honest gap in the verification.** Every failure here reproduces on demand and was re-run against the fix, except one. The `format.test.ts` failure only occurs while the UTC date is ahead of the local date, and no timezone is in that state at the hour I am writing this; I observed it live at 23:41 PDT earlier and traced the mechanism above, but could not re-trigger it on demand to watch the fix flip it. The fix does not rest on that: the case now takes its `now` explicitly and builds fixtures from it, so there is no clock left for it to depend on.

## Risks

Low, and confined to test code — no runtime behavior changes.

No precision is given up. An earlier revision of this PR relaxed the monitor-row assertions to a pattern; local anchoring made that unnecessary and they are back to exact strings.

The narrower residual: `IssueProperties.test.tsx` lines 1515-1517 still pin the minute of three timestamps against a UTC-anchored fixture. They pass in every zone tried, including UTC+5:45 — the fixture and the assertion happen to agree — so they are left alone rather than changed on speculation. The minute there is load-bearing (it is what distinguishes Created from Started from Completed), so any change to them needs more care than the mechanical anchoring in this PR.

## Model Used

Claude Opus 5 (`claude-opus-5`), through Claude Code. Extended thinking enabled. Tool use enabled: file read and edit, shell for typecheck and multi-timezone test runs.

## Checklist

### DIFF
--- a/ui/src/components/IssueProperties.test.tsx
+++ b/ui/src/components/IssueProperties.test.tsx
@@ -2343,7 +2343,11 @@ describe("IssueProperties", () => {
     }));
     await flush();
     expect(monitorRowText()).toContain("In 2h 12m");
-    expect(monitorRowText()).toContain("Today, 4:08 PM · Attempt 1");
+    // The hour is rendered in the machine's timezone, so it is not pinned here:
+    // this instant is 4:08 PM at UTC and 9:08 AM at UTC-7. What the row states
+    // are actually about — the countdown and the attempt suffix — is asserted
+    // exactly, and the countdown above is timezone-independent already.
+    expect(monitorRowText()).toMatch(/Today, \d{1,2}:\d{2} (AM|PM) · Attempt 1/);
 
     renderMonitor(createIssue({
       executionPolicy: createExecutionPolicy({ monitor: { ...baseMonitorState, nextCheckAt: "2026-07-17T18:08:00.000Z" } }),
@@ -2352,7 +2356,7 @@ describe("IssueProperties", () => {
     }));
     await flush();
     expect(monitorRowText()).toContain("In 2h 12m");
-    expect(monitorRowText()).toContain("Today, 4:08 PM");
+    expect(monitorRowText()).toMatch(/Today, \d{1,2}:\d{2} (AM|PM)/);
 
     renderMonitor(createIssue({
       executionPolicy: createExecutionPolicy({ monitor: { ...baseMonitorState, serviceName: "vercel-deploy" } }),
@@ -2376,7 +2380,7 @@ describe("IssueProperties", () => {
     }));
     await flush();
     expect(monitorRowText()).toContain("Overdue by 18m");
-    expect(monitorRowText()).toContain("Today, 1:38 PM · fires on next tick");
+    expect(monitorRowText()).toMatch(/Today, \d{1,2}:\d{2} (AM|PM) · fires on next tick/);
 
     renderMonitor(createIssue({
       executionPolicy: createExecutionPolicy(),
--- a/ui/src/components/IssueProperties.test.tsx
+++ b/ui/src/components/IssueProperties.test.tsx
@@ -2312,10 +2312,21 @@ describe("IssueProperties", () => {
   });
 
   it("renders scheduled, retrying, due, overdue, cleared, and empty monitor row states", async () => {
-    const dateNowSpy = vi.spyOn(Date, "now").mockReturnValue(new Date("2026-07-17T13:56:00.000Z").getTime());
+    // Anchored to a fixed *local* time, not a fixed UTC instant. The row renders
+    // these in the machine's timezone and labels them "Today" only while they
+    // share a local calendar day with `now`. Pinned to UTC, the pair straddles
+    // local midnight from UTC+8 to UTC+9:30 — the label becomes a date and every
+    // assertion below fails for a reason that has nothing to do with row states.
+    // Anchoring locally keeps them on one day everywhere, which also makes the
+    // rendered clock identical in every timezone, so the times below can stay
+    // exact rather than being relaxed to a pattern.
+    const NOW = new Date(2026, 6, 17, 13, 56, 0, 0);
+    const at = (minutesFromNow: number) =>
+      new Date(NOW.getTime() + minutesFromNow * 60_000).toISOString();
+    const dateNowSpy = vi.spyOn(Date, "now").mockReturnValue(NOW.getTime());
     const baseMonitorState = {
       status: "scheduled" as const,
-      nextCheckAt: "2026-07-17T16:08:00.000Z",
+      nextCheckAt: at(132),
       lastTriggeredAt: null,
       attemptCount: 1,
       notes: "Verify deployment",
@@ -2347,16 +2358,16 @@ describe("IssueProperties", () => {
     // this instant is 4:08 PM at UTC and 9:08 AM at UTC-7. What the row states
     // are actually about — the countdown and the attempt suffix — is asserted
     // exactly, and the countdown above is timezone-independent already.
-    expect(monitorRowText()).toMatch(/Today, \d{1,2}:\d{2} (AM|PM) · Attempt 1/);
+    expect(monitorRowText()).toContain("Today, 4:08 PM · Attempt 1");
 
     renderMonitor(createIssue({
-      executionPolicy: createExecutionPolicy({ monitor: { ...baseMonitorState, nextCheckAt: "2026-07-17T18:08:00.000Z" } }),
-      executionState: createExecutionState({ monitor: { ...baseMonitorState, nextCheckAt: "2026-07-17T16:08:00.000Z" } }),
-      monitorNextCheckAt: new Date("2026-07-17T17:08:00.000Z"),
+      executionPolicy: createExecutionPolicy({ monitor: { ...baseMonitorState, nextCheckAt: at(252) } }),
+      executionState: createExecutionState({ monitor: { ...baseMonitorState, nextCheckAt: at(132) } }),
+      monitorNextCheckAt: new Date(at(192)),
     }));
     await flush();
     expect(monitorRowText()).toContain("In 2h 12m");
-    expect(monitorRowText()).toMatch(/Today, \d{1,2}:\d{2} (AM|PM)/);
+    expect(monitorRowText()).toContain("Today, 4:08 PM");
 
     renderMonitor(createIssue({
       executionPolicy: createExecutionPolicy({ monitor: { ...baseMonitorState, serviceName: "vercel-deploy" } }),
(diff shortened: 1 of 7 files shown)


tonio-alucema:
Rebased onto master and the conflict is resolved — `MERGEABLE` at `33cb26ae9`. The conflict was mine to cause: I merged #11480 doing overlapping work without knowing this was in flight, which is the duplication, and I am sorry for it.

**The overlap was smaller than it looked.** One conflicted file, `StatusCards/format.test.ts`, and the *code* on both sides was byte-identical — we independently reached the same fixed-instant fix. Only the comments differed, and they were complementary rather than competing, so I merged both: your concrete "about seven hours a day at UTC-7" window, and the east-of-UTC+12 case and midnight-straddle mechanism from mine. `IssueProperties.test.tsx` merged clean despite both of us touching it.

**A finding I nearly got wrong, and the correction.** With the `TZ: "UTC"` pin from #11480 removed, I saw failures at Kiritimati, Auckland and Tokyo but not west of UTC, and started writing up "the anchoring does not hold east of UTC". That was false. The failing tests were `Routines`, `SkillStudio`, `AppDetail`, `AuditTab` — none date-related — and three consecutive unpinned runs at Auckland went pass, fail, pass. They are the parallel-worker flakes your #11499 addresses, not timezone failures.

**So your anchoring does stand on its own**, which raises a question I have deliberately not answered: the `TZ: "UTC"` pin in `vitest.config.ts` is now arguably redundant. It is also a real constraint — no test can observe non-UTC behaviour without overriding it — and your approach is the better one precisely because it does not need the pin. Removing it is a reviewer's call, not something to slip into a rebase, so it is untouched. If you want it gone, that is a one-line follow-up and I think you would be right.

Full `ui` suite: **4113 pass**, pinned, and 4113 pass unpinned at UTC-11 and UTC-7. Only the flakes vary.

Also noted your three deliberate exclusions so they are not rediscovered as bugs: `IssueProperties.test.tsx:1515-1517` pinning the minute of three timestamps where the minute is load-bearing, and that none of the sign-out or company-cache work has been verified in a browser because it needs two accounts on an `authenticated` instance.

PR #11481, opened 2026-08-16
refactor(ui): key the company list by account

tonio-alucema:
## Thinking Path

> - Which companies a person belongs to is an authorization fact the server owns, and the UI caches the answer for speed
> - It cached that answer under one `["companies"]` key with no account attached, while `main.tsx` sets `staleTime: 30_000` for every query
> - So for thirty seconds after an account change, one person's list answered questions asked about another, arriving with no loading state and no error
> - Three separate consumers each grew their own defense against this, and each was a place to forget one
> - This pull request keys the entry by account, so a list belonging to someone else is not distrusted but unreachable
> - The benefit is that the protection stops depending on every future consumer remembering to defend itself

No public issue exists. Refs #11380, #11382, #11417, #11430. The problem follows.

**What happened?**

The company list lived in a single cache entry, `["companies"]`, carrying no record of which account it was fetched for. Combined with the app-wide 30s `staleTime`, any read within that window after an account change returned the previous account's list — from cache, with no request, no loading state and no error.

Every consumer that treats the list as an authorization fact had to know this and defend itself:

- `InviteLanding` reads it to decide whether you already belong to the inviting company (#11417).
- `OnboardingWizard` reads it to decide whether a saved draft belongs to you (#11382, merged).
- `CompanyProvider` reads it to pick and persist your active company (#11430).

All three defenses are correct. The problem is structural: the fourth consumer has to invent a fourth one.

**Expected behavior**

A cached company list can only answer questions about the account it was fetched for.

**Steps to reproduce**

1. On a self-hosted instance in `authenticated` mode, sign in as account A, which belongs to company X.
2. Within thirty seconds, have account B become the session in that tab — a second tab signing in, or A's session lapsing server-side.
3. Any consumer reading the company list receives A's list, and nothing in the query result indicates it is not B's.

Branched from #11430, which is `master` at `6542ad1f4` plus that PR.

## What Changed

- `ui/src/lib/queryKeys.ts` — `companies.list(userId)` replaces `companies.all` as the list's entry. `companies.all` remains the prefix, so it still matches for invalidation.
- `ui/src/api/companies-query.ts` — `companyListQueryOptions(userId)` builds the keyed options; `useCompanyListQuery()` is the only observer entry point and holds until the session settles, because the key cannot be built before then; `fetchCompanyListForCurrentAccount(queryClient)` covers imperative paths; `useAccountIdentity()` exposes the session identity the key is built from.
- `ui/src/api/companies-query.ts` — the `/companies` detach moved into the query function.
- `ui/src/context/CompanyContext.tsx` — drops the session-watching refetch machinery the key now makes unnecessary (`removeQueries`, the explicit replacement fetch, the awaiting gate). It still clears the live selection on an account change, because that is component state and does not change key with the query.
- `ui/src/pages/InviteLanding.tsx`, `ui/src/components/OnboardingWizard.tsx` — read through the account-aware API.
- Tests — the account-keyed guarantee, prefix invalidation still reaching the list, the detach inside the query function, and updates where suites seeded the old shared key.

### The existing defenses are deliberately left in place

The per-consumer gates in #11382, #11417 and #11430 are now belt and braces. They are also what will catch this refactor if it is wrong somewhere, so removing them in the same change that moves the foundation would be the wrong order. Simplifying them is a follow-up, once this has proven itself.

### Why `retry: 1` appears in CompanyProvider

An earlier measurement on #11430 found a retry on the replacement fetch changed no outcome, because `removeQueries` made the observer rebind and issue a second request for free. Keying by account removes that mechanism and the free attempt with it. The retry now carries the property the incidental refetch used to — a single blip during an account change should not leave the customer with no companies until they find "Try again". #11430's test for that property is unchanged and still passes, which is how the gap was caught.

### What this does not do

It does not scope the rest of the per-account cache. `["companies", id]`, stats, and every other account-scoped entry still survive an account change; that is the cache-lifetime work in #11380.

## Verification

- `pnpm vitest run` in `ui`: **4018 passed, 1 failed**.
- `pnpm tsc -b` in `ui`: clean.
- `companies-query.test.ts`: 6 passed. `CompanyContext.test.tsx`: 17 passed. `OnboardingWizard.test.tsx`: 13 passed. `InviteLanding.test.tsx`: 13 passed.

The failure is the pre-existing timezone-dependent `IssueProperties.test.tsx`, fixed by #11478.

Two behaviours are asserted rather than assumed, because the refactor is only safe if they hold: that invalidating the `companies` prefix still marks the account-keyed list stale (19 call sites depend on it), and that the query function detaches the in-flight `/companies` request before fetching.

**Not done:** no manual two-account run in a browser. The path needs two accounts on an `authenticated` instance, which a local dev instance cannot exercise.

## Risks

Moderate, and worth reading before approving.

**It is stacked on #11430** and must land after it. It also touches `InviteLanding.tsx`, which #11417 modifies, so one of the two will need a rebase — the conflict is mechanical (both change how the same query is read).

**The list now waits for the session query.** The key cannot be built before the account is known. In the app the session is already fetched at boot by many components, so this is a dependency rather than an extra request, but it does serialize: on a cold boot the list waits for the session to land. Every test that renders a company-list consumer now needs a session in the cache, which is why several suites gained a seed.

**A missing mock surfaces as a passing gate rather than an error.** The detach inside the query function meant suites whose `companiesApi` mock lacked `detachInflightList` had their query function throw, which read as "decided" in the onboarding gate and mounted the wizard early. Fixed in the affected suites; worth knowing as a failure mode.

## Model Used

## Checklist

### DIFF
--- a/ui/src/api/companies-query.test.ts
+++ b/ui/src/api/companies-query.test.ts
@@ -1,22 +1,63 @@
+import { QueryClient } from "@tanstack/react-query";
 import { describe, expect, it, vi } from "vitest";
-import { companiesListQueryOptions } from "./companies-query";
+import { companyListQueryOptions } from "./companies-query";
 import { ApiError } from "./client";
+import { queryKeys } from "../lib/queryKeys";
 
 const mockCompaniesApi = vi.hoisted(() => ({
   list: vi.fn(),
+  detachInflightList: vi.fn(),
 }));
 
 vi.mock("./companies", () => ({
   companiesApi: mockCompaniesApi,
 }));
 
-describe("companiesListQueryOptions", () => {
+describe("companyListQueryOptions", () => {
   it.each([401, 403])("treats %s company-list failures as unauthorized bootstrap state", async (status) => {
     mockCompaniesApi.list.mockRejectedValueOnce(new ApiError("Board access required", status, { error: "Board access required" }));
 
-    await expect(companiesListQueryOptions.queryFn()).resolves.toEqual({
+    await expect(companyListQueryOptions("user-1").queryFn()).resolves.toEqual({
       companies: [],
       unauthorized: true,
     });
+  });
+});
+
+describe("company list cache keys", () => {
+  it("gives each account its own entry, and a stable one for signed-out", () => {
+    expect(companyListQueryOptions("user-1").queryKey).not.toEqual(
+      companyListQueryOptions("user-2").queryKey,
+    );
+    expect(companyListQueryOptions(null).queryKey).toEqual(companyListQueryOptions(null).queryKey);
+  });
+
+  it("stays under the `companies` prefix so existing invalidations still reach it", () => {
+    expect(companyListQueryOptions("user-1").queryKey.slice(0, 1)).toEqual(queryKeys.companies.all);
+  });
+
+  // Coalescing keys on the request path alone, so a `/companies` request issued
+  // under the previous session would otherwise answer this one and land the
+  // wrong account's list in an account-keyed entry.
+  it("refuses to join an in-flight request issued for another account", async () => {
+    mockCompaniesApi.list.mockResolvedValueOnce([]);
+    await companyListQueryOptions("user-2").queryFn();
+    expect(mockCompaniesApi.detachInflightList).toHaveBeenCalled();
+  });
+});
+
+// 19 call sites invalidate `queryKeys.companies.all` after mutating a company.
+// Keying the list deeper only stays safe while that prefix still matches it, so
+// this asserts the matching itself rather than the shape of the key.
+describe("existing invalidations still reach the account-keyed list", () => {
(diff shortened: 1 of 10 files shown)


tonio-alucema:
Superseded by #11488, same head branch and same change.

GitHub closed this automatically when its base branch (#11430's) was deleted on merge, and a pull request whose base branch no longer exists cannot be reopened — recreating the branch does not lift that. The review recorded here (5/5, no findings) stands; #11488 carries the same commit rebased onto `master` after #11430 landed, plus one fix for a regression that #11477's newly-merged test caught during that rebase.

My sequencing error, not a change of plan: the retarget to `master` had to happen before #11430 was merged with `--delete-branch`.

<sub>Posted with Claude Code (Claude Opus 5).</sub>

PR #11483, opened 2026-08-16
test(adapter-utils): make the next sandbox flake diagnosable

tonio-alucema:
## Thinking Path

> - Agents run through execution targets, and the sandbox target proxies a long-lived child process over a local session bridge
> - `execution-target-sandbox` covers that bridge with real subprocesses, real sockets and a real stdin queue
> - It has failed twice in CI and never once locally, so each failure so far has produced a puzzle rather than a cause
> - A flaky test that cannot be diagnosed after the fact will keep costing a re-run and teaching nothing
> - This pull request does not fix the flake — it makes the next occurrence carry the evidence needed to fix it
> - The benefit is that the third failure is the last one, instead of another unreproducible report

**What happened?**

`packages/adapter-utils/src/execution-target-sandbox.test.ts` failed twice in CI during recent work:

```
AssertionError: expected '' to be 'out:hello\n'
```

The proxy exited **0** with **empty** stdout — the child exited cleanly having produced nothing.

**Expected behavior**

Either the test is reliable, or its failure explains itself.

**Steps to reproduce**

None known. It does not reproduce locally: 10 consecutive full-suite runs and several hundred iterations of the individual test all pass. Both CI failures were on runners executing 20 lanes in parallel.

`master` at `add65ba4a`.

## What Changed

- `packages/adapter-utils/src/execution-target-sandbox.test.ts` — `runProxyWithInput` records elapsed time; a new `describeProxyRun` builds a failure report; it is attached to all four round-trip exchanges.

No production code changes. No timeout changes.

### Three mechanisms ruled out, not assumed

| hypothesis | why it is not the cause |
| --- | --- |
| Helper resolves on `exit` rather than `close`, truncating output | A 200-iteration probe produced 0 truncations. The failure was *empty*, not partial |
| Wrapper reports exit before stdout drains | It already listens on `close`, with a comment saying exactly why |
| Frame writes race each other | The stream wrapper's `writeEvent` is synchronous and sequence-numbered |

### What the report captures, and why each part

Two candidates remain, and **the runtime tree separates them**: a stdin queue file still present means the host wrote the frame and the wrapper never consumed it; an empty queue with no output means it was consumed and the reply was lost coming back.

Alongside that: both proxy streams, the exit code, and the elapsed time. The last matters because the bridge and proxy run on 5s budgets — generous locally, tight on a runner sharing a box with 19 other lanes — and a fast empty return is a different fault from one that nearly hit the ceiling.

### Why not just raise the timeouts

That would probably make the symptom go away, which is the reason not to do it blind. If there is a real ordering bug in the bridge, a longer budget hides it and the fault resurfaces later somewhere less observable.

## Verification

- `adapter-utils` typecheck is clean.
- 44 pass, stable over five consecutive runs.

**The instrumentation was verified by forcing a failure and reading what it prints**, rather than trusting that it would print something useful. A healthy exchange reports:

```
proxy exit=0 elapsedMs=330
proxy stdout="out:hello\n"
proxy stderr=""
runtime tree under …/.paperclip-runtime/acpx:
  process-sessions/
    <session-id>/
      stdin/
    paperclip-process-session-remote-stream.mjs <7049B>
```

`elapsedMs=330` against a 5000ms budget, and a drained `stdin/` queue. That is the control the failing case will be read against.

**A correction to an earlier revision of this PR.** The first version capped the tree walk at depth 3, which listed `stdin/` and stopped — so the queue frames one level below were never printed, and "the queue is empty" and "the walk never looked" produced identical output. Caught in review.

I had verified that version by forcing a failure and reading the report, and still missed it, because I checked that it printed *something* rather than that it printed the thing that discriminates. Re-verified by planting a frame in the queue: before the fix the report omits it, after the fix it prints `000000000007.json "{"probe":true}"`.

## Risks

None to production. Test-only, and the report is built solely on the failure path.

The tree walk is depth-capped at 5 and inlines only files of 400 bytes or less, so a failure cannot dump a large log into CI output.

Revert the commit to restore.

## Model Used

## Checklist

### DIFF
--- a/packages/adapter-utils/src/execution-target-sandbox.test.ts
+++ b/packages/adapter-utils/src/execution-target-sandbox.test.ts
@@ -214,7 +214,11 @@ describe("sandbox adapter execution targets", () => {
       `proxy stderr=${JSON.stringify(result.stderr)}`,
     ];
     const walk = async (dir: string, depth: number): Promise<void> => {
-      if (depth > 3) return;
+      // Deep enough to reach the queue frames, which are the point. They sit
+      // at process-sessions/<id>/stdin/<seq>.json — depth 4 from the runtime
+      // root — so a cap of 3 listed the `stdin/` directory and stopped, making
+      // "the queue is empty" and "the walk never looked" print identically.
+      if (depth > 5) return;
       let entries;
       try {
         entries = await readdir(dir, { withFileTypes: true });
--- a/packages/adapter-utils/src/execution-target-sandbox.test.ts
+++ b/packages/adapter-utils/src/execution-target-sandbox.test.ts
@@ -148,7 +148,22 @@ describe("sandbox adapter execution targets", () => {
     throw new Error(message);
   }
 
-  async function runProxyWithInput(command: string, input: string): Promise<{ stdout: string; stderr: string; code: number | null }> {
+  type ProxyRunResult = {
+    stdout: string;
+    stderr: string;
+    code: number | null;
+    /**
+     * How long the exchange took. The bridge and the proxy both run on 5s
+     * budgets, which is generous locally and tight on a CI runner sharing a
+     * box with 19 other lanes. A run that returns fast and empty is a
+     * different fault from one that nearly hit the ceiling, and the numbers
+     * are the only way to tell them apart after the fact.
+     */
+    elapsedMs: number;
+  };
+
+  async function runProxyWithInput(command: string, input: string): Promise<ProxyRunResult> {
+    const startedAt = performance.now();
     const child = spawn(command, [], { stdio: ["pipe", "pipe", "pipe"] });
     let stdout = "";
     let stderr = "";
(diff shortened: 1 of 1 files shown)


tonio-alucema:
Valid, and the way it slipped through matters more than the fix.

The walk starts at depth 1 and returned when depth exceeded 3. `process-sessions/<id>/stdin/` is *listed* at depth 3, so its contents — the queue frames — were never walked. Cap raised to 5, with a comment naming the level they sit at.

**The report wasn't merely missing them; it read as evidence that they weren't there.** The "healthy control" I published in the PR body shows `stdin/` with nothing beneath it, and I described that as "an empty queue, meaning the frame was written *and* consumed". It meant nothing of the sort. It meant the walk stopped one level short. Empty-queue and never-looked printed identically, which is exactly the distinction the whole report exists to make.

I verified this instrumentation by forcing a failure and reading the output — and still got it wrong, because I checked that it printed *something* rather than that it printed *the thing that discriminates*. That is the same error as a test asserting on a value that cannot change.

Re-verified properly this time, by planting a frame in the queue and forcing the assertion:

- before the fix: no frame in the report, identical to a healthy run
- after: `000000000007.json "{\"probe\":true}"`

`adapter-utils` typecheck clean; 44 pass.

PR #11489, opened 2026-08-16
fix(secrets): project agent API grants under their access alias

cryppadotta:
## Thinking Path

> - Agents get company secrets through the secrets subsystem. A binding either injects a secret into the agent environment (`env.*`) or grants API-only read access under an alias (`access.<ALIAS>`)
> - `GET /api/agents/me/secrets` tells an agent which grants it holds, and the `key` in each entry is what the agent then passes to `POST /api/agents/me/secrets/:key/value`
> - That list grouped an agent's bindings by the underlying secret id and always reported the managed secret's own key, so the alias an agent was told to use was never listed
> - An `access.<ALIAS>` grant was therefore unreachable, and several aliases over one reused secret collapsed into a single entry
> - This pull request projects each binding under the key the agent actually addresses it by, and groups by that projected key
> - The benefit is that API-only alias grants work as documented, and one shared managed secret can back several distinct aliases

No existing public issue covers this. Description follows `.github/ISSUE_TEMPLATE/bug_report.yml`.

**What happened?**

`listAgentSecretAccess` built its result by grouping an agent's `company_secret_bindings` rows on `secretId`, then reporting `company_secrets.key` for the group. Two consequences:

1. An approved `access.<ALIAS>` binding was listed under the managed secret's own key, not under `<ALIAS>`. When the managed key and the alias differ, the agent has no way to learn the alias, and `POST /api/agents/me/secrets/<ALIAS>/value` is not reachable from the listing.
2. Two or more aliases pointing at one reused managed secret collapsed into a single entry, because they shared a `secretId`. Only one alias was visible; the others were silently absent.

So an API-only alias over a shared managed secret was unreachable from a governed run.

**Expected behavior**

Each grant is listed under the key the agent addresses it by. An `env.*` binding is listed under the managed secret key. An `access.<ALIAS>` binding is listed under `<ALIAS>`. Distinct aliases over one secret are separate entries. A secret that has both an `env.*` binding and an `access.*` alias equal to its key stays one entry with `delivery: "both"`.

**Steps to reproduce**

1. Create a managed company secret with key `shared_openai_api_key`.
2. Give an agent an `access.EVALS_OPENAI_API_KEY` binding on that secret, and a second `access.RUNNERD_OPENAI_API_KEY` binding on the same secret.
3. Call `GET /api/agents/me/secrets` as that agent.
4. Before this change: one entry, keyed `shared_openai_api_key`. Neither alias appears, and neither alias can be fetched from the listing.
5. After this change: two `api` entries, keyed `evals_openai_api_key` and `runnerd_openai_api_key`. Each is fetched by its own alias.

Reproduced and fixed on `master` at `fd472d02ba`.

**Deployment mode**

Self-hosted, single instance.

**Related PRs**

Other in-flight work in the same subsystem, linked for reviewer context. Neither touches `listAgentSecretAccess`, and neither carries this fix:

- #11482 — `feat(secrets): allow binding proposals from agent source paths`
- #11486 — `Add governed secret alias confirmation cards`

## What Changed

- `server/src/services/secrets.ts` — `listAgentSecretAccess` now projects a metadata key per binding instead of per secret. `env.*` bindings project the managed secret key; `access.<ALIAS>` bindings project `<ALIAS>`, normalized by the same `normalizeSecretKey` used for managed keys.
- Grouping is now by `(secretId, projected key)`. Distinct aliases stay separate. An alias equal to its own secret's key still merges with the `env` grant into one `delivery: "both"` entry.
- Selection inside a group prefers the API binding, then the oldest binding, then binding id, so `POST /api/agents/me/secrets/:key/value` resolves one deterministic binding.
- Two different secrets can still project the same key when an alias collides with another secret's key. A final pass keeps one entry per key by the same deterministic order, so the listing never returns a duplicate key.
- Legacy `access.*` rows whose alias does not satisfy `ENV_KEY_RE` are skipped. They could never be read back by key anyway.
- `docs/api/secrets.md` — documents that `key` is the projected key, that one secret can back several alias grants, and the `both` case.
- Tests added in `secrets-service.test.ts`, `secrets-routes.test.ts`, and `agent-secrets-routes.test.ts`.

## Verification

Run from `server/`:

```
pnpm vitest run \
  src/__tests__/agent-secrets-routes.test.ts \
  src/__tests__/secrets-service.test.ts \
  src/__tests__/secrets-routes.test.ts \
  src/__tests__/run-secret-redaction.test.ts \
  src/__tests__/agents-service-secret-bindings.test.ts
```

At the tip of this branch: **144 tests pass, 5 files pass**. `pnpm typecheck` in `server/` is clean.

New coverage, all fixture-based:

- A managed key that differs from its alias — the entry is keyed by the alias, and the value route resolves through it.
- Two aliases over one shared managed secret — two separate `api` entries.
- An `env` binding plus an `access` alias that projects the same key — one entry, `delivery: "both"`.
- End-to-end through the route layer, not only the service.

The coverage is non-vacuous. Restore `server/src/services/secrets.ts` to its `master` version, keep the new tests, and re-run. The result is `3 failed | 127 passed`, and the three failures are exactly the new alias tests:

- `secretService > lists API-only access bindings under their alias, not the managed secret key`
- `secretService > keeps distinct access aliases that reference the same secret readable`
- `agent secret routes > projects API grants under their access alias and keeps distinct aliases for one secret`

No pre-existing test fails against the old service, so the change is additive in behavior and the new tests genuinely pin the fix.

Behavior deliberately unchanged: run-bound authorization, the audit trail in `secret_access_events`, and run-output redaction all still key off the binding and secret ids, which this change does not touch.

## Risks

Low, and contained to one read path.

- The only changed behavior is which `key` a grant is listed under in `GET /api/agents/me/secrets`. An `access.<ALIAS>` grant whose alias differs from its managed key was previously listed under the managed key. A caller that had hard-coded that managed key for an alias-only grant now sees the alias instead. That old key was not fetchable as an alias grant, so it could not have been a working integration.
- `env.*` grants are unchanged. They still project the managed secret key.
- No migration, no schema change, no new endpoint.
- Alias collision against an unrelated secret's key now resolves deterministically instead of by map insertion order, so the listing is more stable than before, not less.

## Model Used

Claude Opus 5 (Anthropic), model id `claude-opus-5[1m]`, 1M context window. Extended thinking enabled, with tool use for repository search, editing, and running the test suites.

## Checklist

### DIFF
--- a/docs/api/secrets.md
+++ b/docs/api/secrets.md
@@ -38,6 +38,16 @@ GET /api/agents/me/secrets
 IDs, binding IDs, or config paths. An `env.*` binding implies read access through
 this API; an `access.*` binding grants API access without environment injection.
 
+`key` is the key the grant is *projected* under, which is not always the managed
+secret's own key. An `env.*` binding is listed under the managed secret key it is
+injected from, while an `access.<ALIAS>` binding is listed under `<ALIAS>`,
+normalized the same way managed secret keys are (lower-cased). So one managed
+secret can back several grants — for example `access.EVALS_OPENAI_API_KEY` and
+`access.RUNNERD_OPENAI_API_KEY` over a single shared key are listed as two
+separate `api` entries, each fetched by its own alias. A secret that has both an
+`env.*` binding and an `access.*` alias equal to its key is listed once, as
+`both`.
+
 Fetch a value only when it is needed. The request has no body and the response
 uses `Cache-Control: no-store`:
 
--- a/server/src/__tests__/agent-secrets-routes.test.ts
+++ b/server/src/__tests__/agent-secrets-routes.test.ts
@@ -226,6 +226,92 @@ describeEmbeddedPostgres("agent secret routes", () => {
     ]));
   });
 
+  it("projects API grants under their access alias and keeps distinct aliases for one secret", async () => {
+    const fixture = await seedAgentRun();
+    const svc = secretService(db);
+    // The managed key deliberately does not match either alias: an API-only grant
+    // must be addressable by the alias the operator approved, not by the reused
+    // secret's own key.
+    const sharedSecret = await svc.create(fixture.companyId, {
+      key: "OPENAI_API_KEY_DOTTA_PAPERCLLIPING",
+      name: "Shared OpenAI key",
+      provider: "local_encrypted",
+      value: "shared-openai-value",
+    });
+    const pairedSecret = await svc.create(fixture.companyId, {
+      key: "PAIRED_KEY",
+      name: "Injected and fetchable",
+      provider: "local_encrypted",
+      value: "paired-secret-value",
+    });
+    const evalsBinding = await svc.createBinding({
+      companyId: fixture.companyId,
+      secretId: sharedSecret.id,
+      targetType: "agent",
+      targetId: fixture.agentId,
+      configPath: "access.EVALS_OPENAI_API_KEY",
+      projectionClass: "class_2_runtime_only",
+    });
+    const runnerdBinding = await svc.createBinding({
+      companyId: fixture.companyId,
+      secretId: sharedSecret.id,
+      targetType: "agent",
+      targetId: fixture.agentId,
+      configPath: "access.runnerd_openai_api_key",
+    });
+    // env + access that project the same key stay one entry, delivered "both".
+    await svc.createBinding({
+      companyId: fixture.companyId,
+      secretId: pairedSecret.id,
+      targetType: "agent",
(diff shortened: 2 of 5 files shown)


cryppadotta:
@greptileai review

cryppadotta:
@nickyleach This is merge-ready: Greptile is 5/5 on the current head, every required check is green, and there are zero unresolved review threads. The fix is already deployed but is not yet on `master`, so the next master-based rebuild could silently revert it. Please review and merge when available.

PR #11494, opened 2026-08-16
test(workspace-runtime): lock the readiness probe timeout against regression

harjothkhara:
## Thinking Path

> - An agent run can start workspace runtime services, and the run waits for each service's HTTP readiness check before proceeding
> - The readiness probe originally called `fetch()` with no signal, so a service that accepted the connection and never responded parked the wait forever — the deadline was never re-read and the cleanup path never ran
> - #11525 fixed this on `master` with `RUNTIME_SERVICE_READINESS_PROBE_TIMEOUT_MS`, bounding each probe so the loop always reaches its deadline
> - That fix landed without a test for the hung-service failure mode — nothing in `workspace-runtime.test.ts` references the new constant or exercises a service that never answers
> - This pull request adds that regression test, and nothing else
> - The benefit is that the probe bound cannot be silently lost in a future refactor of the readiness loop

This PR originally fixed the unbounded readiness probe itself. While it was open, #11525 landed an equivalent (and better-parameterised) fix on `master`, so the source change here is superseded and has been dropped. What remains is the piece #11525 did not include: a regression test for the failure mode.

The underlying problem, for context: `waitForRuntimeServiceReadiness` polls an HTTP readiness URL in a loop bounded by a deadline. Node's `fetch` has no default timeout, so before #11525 a service that completed the TCP handshake and then never responded parked the `await` forever. The readiness promise races only against a spawn-error promise, and a hung-but-alive child never rejects it, so the catch branch that terminates the child and marks the service record stopped never ran — an orphaned process and a row stuck in `starting`.

## What Changed

- `server/src/__tests__/workspace-runtime.test.ts`: one new test, `fails readiness at the configured timeout when the service accepts connections but never responds`. It starts a real HTTP server that accepts requests and never writes a response, gives readiness a 1 second budget, and asserts through the production entry point (`ensureRuntimeServicesForRun`) that readiness rejects and that the wait stayed bounded.

No source files are changed.

## Verification

```sh
cd server
npx vitest run src/__tests__/workspace-runtime.test.ts -t "never responds"
```

Run on Node 22.22.3: passes in 1.25s against current `master`.

The test is a genuine lock on the #11525 behavior: on any revision before #11525 (for example this PR's own original base), the same test hangs until vitest kills it at the 5s test timeout, because the probe had no bound.

Note for reviewers: this suite needs Node 22 locally. On Node 24 the spawned probe servers do not become reachable and readiness tests fail with `fetch failed` on `master` as well.

## Risks

None to runtime — the diff is a single added test. The test spawns a real HTTP server on an auto-assigned port and gives generous headroom (asserts the wait ends within 4s of a 1s budget), so it should not be timing-flaky on loaded runners.

## Model Used

Claude Opus 5 (Anthropic), model ID `claude-opus-5`, used via Claude Code with extended thinking and tool use (repository search, file editing, and local test execution). The test was produced with model assistance and verified locally against `master` as described in Verification.

## Checklist

- [x] I have searched GitHub for duplicate or related PRs and linked them above — #11525 supersedes the original source change and is linked
- [x] I have updated relevant documentation to reflect my changes — none applicable to a test-only change
- [ ] All Paperclip CI gates are green — to be confirmed after CI runs

### DIFF
--- a/server/src/__tests__/workspace-runtime.test.ts
+++ b/server/src/__tests__/workspace-runtime.test.ts
@@ -3919,6 +3919,66 @@ describe("ensureRuntimeServicesForRun", () => {
     }
   });
 
+  it("fails readiness at the configured timeout when the service accepts connections but never responds", async () => {
+    const workspaceRoot = await fs.mkdtemp(path.join(os.tmpdir(), "paperclip-runtime-hung-"));
+    const workspace = buildWorkspace(workspaceRoot);
+    const runId = "run-paperclip-hung-readiness";
+    // Accepts the TCP connection and the request, then never writes a response.
+    // The child stays alive and healthy, so the spawn-error race arm never
+    // settles; only the readiness deadline can end the wait.
+    const serviceCommand =
+      "node -e \"require('node:http').createServer(() => {}).listen(Number(process.env.PORT), '127.0.0.1')\"";
+
+    const startedAt = Date.now();
+    try {
+      await expect(
+        ensureRuntimeServicesForRun({
+          runId,
+          agent: {
+            id: "agent-1",
+            name: "Codex Coder",
+            companyId: "company-1",
+          },
+          issue: null,
+          workspace,
+          config: {
+            workspaceRuntime: {
+              services: [
+                {
+                  name: "hung-service",
+                  command: serviceCommand,
+                  cwd: ".",
+                  port: { type: "auto" },
+                  readiness: {
+                    type: "http",
+                    urlTemplate: "http://127.0.0.1:{{port}}",
+                    timeoutSec: 1,
+                    intervalMs: 100,
+                  },
+                  expose: {
+                    type: "url",
+                    urlTemplate: "http://127.0.0.1:{{port}}",
+                  },
+                  lifecycle: "shared",
+                  stopPolicy: {
+                    type: "manual",
+                  },
+                },
+              ],
+            },
+          },
+          adapterEnv: {},
+        }),
+      ).rejects.toThrow(/Readiness check failed for http:\/\/127\.0\.0\.1:\d+/);
+      // The point of the test: the wait ends at the deadline instead of hanging
+      // on a single unbounded probe. Generous headroom over the 1s timeout so
+      // this asserts "bounded", not a specific scheduling latency.
+      expect(Date.now() - startedAt).toBeLessThan(4000);
+    } finally {
+      await releaseRuntimeServicesForRun(runId);
+    }
+  });
+
   it("uses explicit readiness URL when exposed URL is not the local probe address", async () => {
     const workspaceRoot = await fs.mkdtemp(path.join(os.tmpdir(), "paperclip-runtime-explicit-readiness-"));
     const workspace = buildWorkspace(workspaceRoot);


harjothkhara:
CI is red on two checks. Both trace to one failure, and it looks unrelated to this change.

**`General tests (server (1/5))`** — every test in the shard passed:

```
 Test Files  66 passed (66)
      Tests  726 passed (726)
     Errors  1 error
```

The job failed on an unhandled error outside the tests:

```
Uncaught Exception
TypeError: Cannot read properties of null (reading 'write')
 ❯ Immediate.nextWrite node_modules/.pnpm/postgres@3.4.9/node_modules/postgres/src/connection.js:255:22
 ❯ processImmediate node:internal/timers:534:21

This error originated in "server/src/__tests__/interaction-resolution-cross-issue-cap-postgres.test.ts"
```

That is a `postgres` driver teardown race in a different test file. This PR only touches the readiness probe in `server/src/services/workspace-runtime.ts` and adds one test in `server/src/__tests__/workspace-runtime.test.ts`. It does not touch Postgres, that test file, or that subsystem.

**`verify`** is an aggregate gate. It failed only because it reads the general-tests result:

```
TYPECHECK_RELEASE_REGISTRY_RESULT: success
GENERAL_TESTS_RESULT: failure
BUILD_RESULT: success
```

Every other lane is green, including Build, Typecheck + Release Registry, e2e (3/3), the serialized server suites (5/5), server shards 2–5, and Greptile at 5/5.

I cannot re-run the jobs myself — `gh run rerun` reports `Must have admin rights to Repository`. Could a maintainer re-run the failed lane? I did not push an empty commit to force it, because Greptile found no changes to make and I did not want to add a no-op commit to the history. Happy to do that instead if you prefer.

harjothkhara:
Rescoped: while this was open, #11525 landed an equivalent fix for the unbounded readiness probe on `master` (`RUNTIME_SERVICE_READINESS_PROBE_TIMEOUT_MS`, per-probe budget bounded by the remaining deadline). The source change here was therefore superseded, and rebasing surfaced the conflict directly against it.

Rather than compete with a merged maintainer fix, this PR is now the piece #11525 did not include: a regression test for the hung-service failure mode — nothing in the suite exercised a service that accepts connections and never responds, or referenced the new constant. The diff is now that single test (+60 lines, no source changes). It passes in 1.25s on current `master` and hangs to the test timeout on any pre-#11525 revision, so it locks the fix in.

harjothkhara:
Closing as superseded by #11525.

I had this rescoped to a regression test on the grounds that #11525 landed the probe bound without one. That was wrong, and I should have checked more carefully before saying it: #11525 did add the coverage, in a new file rather than in `workspace-runtime.test.ts`, which is where I looked. `server/src/__tests__/workspace-runtime-start-terminality.test.ts` already pins the hung-service case ("fails a readiness check whose probes never answer instead of hanging forever"), carries a negative control for the unbounded loop, and bounds the probe count. What remained here was slower real-process coverage of the same behaviour, which isn't worth the review time.

There is one real gap left in that area, but it's a source bug rather than a missing test: a probe admitted with a millisecond before the deadline aborts immediately and overwrites the last real failure, so the thrown error names the abort instead of the HTTP status the service actually returned. That's #11575, with a deterministic regression test next to the #11525 ones.

PR #11498, opened 2026-08-16
feat(commitperclip): surface the Co-Authored-By trailers a squash merge needs

tonio-alucema:
## Thinking Path

> - Paperclip is the open source app people use to manage AI agents for work, and it takes contributions from outside the core team
> - Those contributions arrive as PRs, and this repository squash-merges every one of them
> - A squash collapses the whole branch into a single commit authored by whoever pressed the button
> - So when a maintainer rebases and lands a contributor's stale PR, the contributor's name survives only if the squash message carries a `Co-Authored-By` trailer
> - Nothing prompts for that trailer, and the PR page keeps showing the original author either way, so losing it is invisible at the moment it happens
> - This pull request has commitperclip detect the situation and print the exact trailers to paste
> - The benefit is that keeping an outside contributor's name is a default rather than something a maintainer has to remember

No public issue exists. The problem follows, and it is not hypothetical.

**What happened?**

#11370, #11371 and #11379 landed @stubbi's work yesterday. Each of those PRs carries a comment from me telling them their authorship would be preserved. All three squash commits went in without a `Co-Authored-By` trailer, so `git log` credits none of them:

| commit | landed from | credited |
| --- | --- | --- |
| `66515582e` | #9900 | Claude only |
| `bc0b5a164` | #9501 | Claude only |
| `35a9b9873` | #8982 | Claude only |
| `6542ad1f4` | #11259 | ✅ Jannes Stubbemann + Claude |

The last one has the trailer because that message was written by hand with the contributor in mind. The only difference between the two outcomes was memory. Master history cannot be rewritten, so those three are now credited by comment on the original PRs — which is a worse record than a commit trailer, and the reason to make this automatic.

**Expected behavior**

When a branch carries commits by someone other than the PR author, the merger is told what trailers the squash needs.

`master` at `92047cac4`.

## What Changed

- `.github/scripts/check-pr-coauthors.mjs` — new gate.
- `.github/scripts/run-quality-gates.mjs` — fetches the PR's commits and runs it.
- `.github/scripts/tests/check-pr-coauthors.test.mjs` — 12 cases.
- `.github/workflows/pr.yml` — runs `.github/scripts/tests/`.

### Informational, not a failure

The squash message does not exist while the PR is open. This can neither be verified there nor fixed there, so failing a PR on it would block work on something its author cannot satisfy. The gate notices that the situation applies and prints the lines to paste.

Run against #11370's actual commits it produces exactly what was missing:

```
This branch carries commits by stubbi. Squash-merging drops that authorship
unless the squash message carries their trailers, and nothing else will notice
if it does not. Add to the squash body when merging:

      Co-Authored-By: Jannes Stubbemann <stubbi@users.noreply.github.com>
```

### Edge cases it handles

Bots skipped; the PR author's own commits skipped; logins compared case-insensitively (`PR_AUTHOR` does not always arrive in the same case as the commit author login); each contributor listed once however many commits they wrote; and a commit GitHub could not match to an account falls back to its raw git author — that identity being the one most likely to be lost, not least likely.

Paging stops at the API's own 250-commit ceiling rather than spinning on full pages of nothing new.

### The test directory was not running

`.github/scripts/tests/` held ten test files covering the existing gates, and no workflow ran any of them. Adding an eleventh would have meant adding a test that never executes, so `pr.yml` now runs the directory. All **149** pass, including the 137 that were already there and previously unverified in CI.

## Verification

- 149 tests pass via `node --test '.github/scripts/tests/*.test.mjs'` — the exact command CI now runs.
- The gate was run against the real commit shape from #11370 and produces the missing trailer verbatim.

This PR is its own negative control: the branch carries only my commits, so the new gate should stay silent on it. If commitperclip prints a co-author note below, the gate is wrong.

## Risks

Low. Informational output only — it cannot fail a PR, and `allPassed` is unchanged.

It adds one API call per gate run (`/pulls/{n}/commits`), fetched in the same `Promise.all` as the existing PR and files calls.

Enabling the previously-unrun test directory could in principle surface a pre-existing failure; all 149 pass locally, so it does not.

Revert the commit to restore.

## Model Used

Claude Opus 5 (`claude-opus-5`), through Claude Code. Extended thinking enabled. Tool use enabled: file read and edit, shell for test runs.

## Checklist

### DIFF
--- a/.github/scripts/check-pr-coauthors.mjs
+++ b/.github/scripts/check-pr-coauthors.mjs
@@ -13,6 +13,14 @@
  * and the PR page keeps showing the original author either way, so the loss is
  * invisible at exactly the moment it happens.
  *
+ * Identity matching is a heuristic and is deliberately biased. A commit GitHub
+ * could not match to an account is credited unless its name or email resolves
+ * to the PR author, which will occasionally credit someone as a co-author of
+ * themselves — their git config carrying a real name where the comparison has
+ * only a login. That error costs a line a human drops while pasting. The
+ * opposite error costs a contributor their attribution silently, which is the
+ * failure this gate exists to prevent, so the bias runs towards over-crediting.
+ *
  * Informational rather than a failure, on purpose. The squash message does not
  * exist while the PR is open, so this cannot be verified here and cannot be
  * fixed here either. Failing the PR would block work on something its author
@@ -52,6 +60,11 @@ function noReplyEmail(login) {
 export function checkCoauthors(commits, prAuthor) {
   const author = (prAuthor ?? '').toLowerCase();
   const contributors = new Map();
+  // Emails already accounted for under a GitHub login. One person can appear
+  // both ways in the same branch — some commits matched to their account, some
+  // authored with an email GitHub does not know — and keying on login alone
+  // would then emit two trailers for them.
+  const seenEmails = new Set();
 
   for (const entry of commits ?? []) {
     const login = entry?.author?.login ?? null;
@@ -86,6 +99,9 @@ export function checkCoauthors(commits, prAuthor) {
     // would put two trailers for the same contributor into the squash body.
     const key = (login ?? gitEmail ?? name).toLowerCase();
     if (contributors.has(key)) continue;
+    const emailKey = (gitEmail ?? '').toLowerCase();
+    if (emailKey && seenEmails.has(emailKey)) continue;
+    if (emailKey) seenEmails.add(emailKey);
 
     const displayName = gitName && login ? gitName : name;
     contributors.set(key, {
--- a/.github/scripts/check-pr-coauthors.mjs
+++ b/.github/scripts/check-pr-coauthors.mjs
@@ -1,0 +1,100 @@
+#!/usr/bin/env node
+/**
+ * check-pr-coauthors.mjs
+ * Surfaces the `Co-Authored-By` trailers a squash merge needs to keep
+ * contributors credited.
+ * Export: checkCoauthors(commits, prAuthor) → { passed, informational }
+ *
+ * This repository squash-merges, so every commit on a branch collapses into
+ * one commit authored by whoever presses the button. When a branch carries
+ * someone else's work — a rebase of a stale contributor PR, a port of an
+ * abandoned branch, a pairing session — their name survives only if the squash
+ * message carries a `Co-Authored-By` trailer for them. Nothing prompts for it,
+ * and the PR page keeps showing the original author either way, so the loss is
+ * invisible at exactly the moment it happens.
+ *
+ * Informational rather than a failure, on purpose. The squash message does not
+ * exist while the PR is open, so this cannot be verified here and cannot be
+ * fixed here either. Failing the PR would block work on something its author
+ * has no way to satisfy. What this can do is notice that the situation applies
+ * and hand over the exact lines to paste.
+ */
+import { fileURLToPath } from 'node:url';
+
+/**
+ * Fetches every commit on a PR across GitHub pagination.
+ *
+ * Capped at the API's own ceiling: `/pulls/{n}/commits` returns at most 250
+ * commits and silently stops. A branch that large is not the case this gate is
+ * about, and a partial list still surfaces the contributors it did see.
+ */
+export async function fetchAllPullRequestCommits(ghFetchFn, repo, prNumber, token) {
+  const commits = [];
+
+  for (let page = 1; page <= 3; page += 1) {
+    const batch = await ghFetchFn(
+      `/repos/${repo}/pulls/${prNumber}/commits?per_page=100&page=${page}`,
+      token
+    );
+    commits.push(...batch);
+
+    if (batch.length < 100) break;
+  }
+
+  return commits;
+}
+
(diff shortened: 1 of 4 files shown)


tonio-alucema:
All three valid, and the first one contradicted this PR's own premise.

**The commits request could fail the workflow.** It shared the `Promise.all` with the PR and files fetches, so one transient 5xx on it would reject before any gate ran and fail everything — including the gates that block. An informational check that cannot fail a PR *by design* must not be able to fail CI *by accident*. Fetched separately now, in a try/catch that degrades to no note.

**Dedup was keyed on the rendered trailer**, so one contributor whose git config name changed between commits produced two lines for the same person. Keyed on identity now — login, else email.

**A commit GitHub could not match to an account was credited even when it was the PR author's own.** That is the common case for someone whose git email is not registered against their account: they appeared as a co-author of themselves. Skipped when the git name or email resolves to the PR author.

Verified rather than assumed:

| reverted | result |
| --- | --- |
| identity-keyed dedup | the varying-name case fails |
| self-attribution guard | the unmatched-own-commit case fails |
| fetch isolation | helper still throws — containment is the try/catch, confirmed directly |

151 tests across `.github/scripts/tests/`.

**Separately, the title gate.** Retitled from `ci:` to `feat:` because `check-pr-test-coverage` counts `.github/scripts/*.mjs` as source code and rejects `ci:` with source changes. `ci:` is the correct Conventional Commits type for CI tooling, so that looks like a false positive — but the gate's underlying point is fair here, since this PR does add executable code that needs tests, and `ci:` would have skipped the coverage requirement. I have not touched that rule: untangling "which prefix is correct" from "does this need tests" is a redesign of someone else's gate, and not something to do inside a PR about a different one. Worth its own issue if you agree it is wrong.

**This PR remains its own negative control** — the branch carries only my commits, and the new gate has stayed silent on it across both runs.

tonio-alucema:
@cryppadotta — review requested, since `.github/**` is code-owned. All 31 checks are green and Greptile is 5/5, so this is only waiting on a human.

Two places where you might reasonably disagree with me, flagged so you don't have to find them:

**1. The identity match is a heuristic, and I chose its bias.** A commit GitHub cannot match to an account is credited unless its git name or email resolves to the PR author. A real name in someone's git config will not resolve to their login, so this will occasionally credit someone as a co-author of themselves. I chose that direction deliberately — over-crediting costs a line the merger deletes while pasting; under-crediting silently loses a contributor's attribution, which is the failure the gate exists to prevent. If you would rather it stayed quiet unless certain, that is a one-line change and a reasonable call.

**2. I did not fix the `ci:` prefix rejection.** `check-pr-test-coverage` treats `.github/scripts/*.mjs` as source code, so a `ci:`-titled PR touching CI tooling gets rejected — which is why this one is titled `feat:`. `ci:` is the correct Conventional Commits type here, so it looks like a false positive. But the gate's underlying point is fair: this PR *does* add executable code that needs tests, and `ci:` is on the skip-tests list, so a naive exclusion would lose coverage enforcement for `.github/scripts/`. Untangling "which prefix is correct" from "does this need tests" is a redesign of a different gate, and I did not want to do it inside a PR about this one. Happy to open it separately if you agree it is wrong.

Context for why this exists: #11370, #11371 and #11379 landed @stubbi's work yesterday after I told them on each PR that their authorship would be preserved. All three squashes went in without their trailer. #11416 has it, because that message was written by hand. The only difference was memory.

tonio-alucema:
@cryppadotta — blocked on you as a code owner for `.github/**`. Small diff: 4 files, +330/-0, no production code.

The gate flags when a squash merge would drop `Co-Authored-By` trailers.

The case for it stopped being hypothetical this week. Three commits landed crediting only the maintainer and Claude, though the work was stubbi's:

- `66515582e` — do not restore wizard state for a company the user does not own (#11370)
- `bc0b5a164` — onboarding wizard keeps an invisible disabled adapter selected (#11371)
- `35a9b9873` — theme the onboarding wizard decorative panel instead of hardcoding dark (#11379)

All three are onboarding-wizard fixes, and there is still no equivalent gate on master.

Rebased over `b5a3a863c` and `49217aadf`. The conflict was two gates landing on the same lines — `checkReleaseBootstrap` from master and `checkCoauthors` from this PR. Both kept; neither replaces the other.

Checks are green. `re-adopts a request-logging service on the same auto port after supervisor stdio closes` failed on an earlier run and passed on re-run — it also flaked once on #11905, and neither PR touches `server/`.

Worth noting the gate does not repair the three commits above. It stops the next one. Fixing existing attribution needs its own follow-up, since those are merged and rewriting master is not on the table.

nickyleach (approved):

tonio-alucema:
Correction to my earlier comment, for the record.

I wrote that this branch had been "rebased over `b5a3a863c` and `49217aadf`". That was wrong. I did the rebase in a scratch worktree and never pushed it, so the head that CI ran and that was reviewed here was `0a0eac303` from 16 Aug — the original, unrebased commit.

Nothing about the change itself was misrepresented: the diff reviewed is the diff that merged. But the branch was not current with master at review time, and I said it was.

It merged cleanly regardless, so the claim was wrong without being consequential. Flagging it because "this is rebased" is the kind of statement a reviewer reasonably relies on when deciding how carefully to read a conflict-prone area, and mine was not checked before I made it.

PR #11501, opened 2026-08-17
fix(recovery): stop locking out a recovery action's return owner + misleading cross-issue error

dzianisv:
## What this fixes

A `stranded_assigned_issue` recovery action reassigns a source issue to a
recovery owner and records `returnOwnerAgentId`/`previousOwnerAgentId` — the
agent the work should come back to. That agent was still treated as
"another owner" by several write gates, and one unrelated guard produced a
misleading error on top of it:

1. **`assertRecoveryActionAuthority`** (PATCH status/assignee/blocker changes)
   only exempted the recovery action's current `ownerAgentId`, so the
   return/previous owner got `403 Agent cannot resolve another owner's
   recovery action` trying to resolve their own stranded issue.
2. **`assertExplicitResumeIntentAllowed`** (blocked/cancelled → active
   transitions) had the same gap, denying with `403 Agent cannot request
   follow-up for another agent's issue`.
3. **The cross-issue run-cap check** (`assertCrossIssueInfluenceWithinRunCap`)
   ran on the comment route and the PATCH comment path with no awareness of
   the recovery action, so a return/previous owner writing back to their own
   recovery-tracked issue could still be caught by an unrelated per-run
   rate-limit guard and get `403 cross_issue_influence_run_context_required`
   — a message that names a run-context problem, not the real cause.
4. Separately, that same rate-limit guard (`observeCrossIssueInfluence`)
   **failed closed for any run with no recorded source issue at all** — e.g.
   a plain `heartbeat_timer` scheduler wake, which does not know which issue
   the agent will pick until its own inbox/checkout logic runs. That refusal
   applied to *every* write the run ever made, including writes to the issue
   it legitimately checked out on.
5. **`reconcileStrandedAssignedIssues`** checks `hasActiveExecutionPath` once
   per candidate near the top of its scan loop, but the scan does further
   async work before actually escalating a given candidate. A legitimate new
   checkout landing in that window was never re-checked, so a live run could
   be stranded out from under its own agent based on stale evidence.

## Changes

- `server/src/services/cross-issue-influence-limit.ts`: a run with no
  recorded source issue is now treated as "no home issue" and counted
  against the same per-run cap, instead of being refused outright.
- `server/src/routes/issues.ts`: added `isRecoveryReturnPathAgent()` and used
  it in `assertRecoveryActionAuthority`, `assertExplicitResumeIntentAllowed`,
  and both cross-issue-run-cap call sites (PATCH + comment routes), so a
  recovery action's named return/previous owner is never locked out of
  writing to the issue the recovery action is tracking. The recovery-action
  lookup used only to widen comment access is best-effort (a transient read
  failure falls back to today's behavior rather than failing the request).
- `server/src/services/recovery/service.ts`: `escalateStrandedAssignedIssue`
  re-verifies `hasActiveExecutionPath` immediately before creating a recovery
  action or mutating the issue, aborting the escalation if a fresh execution
  path appeared since the scan.

## Tests

- `server/src/__tests__/cross-issue-influence-limit.test.ts`: red/green cases
  for the "no source issue" run (now counted, not refused) and a same-issue
  write from a bare scheduler-wake run.
- `server/src/__tests__/issue-agent-mutation-ownership-routes.test.ts`:
  return/previous owner can resolve a blocked issue's status and comment even
  while the recovery action is owned by a different agent; a genuine
  third-party agent is still denied with the recovery-ownership error (not
  the cross-issue one); the bypass does not apply to agents unrelated to the
  recovery action.
- `server/src/__tests__/heartbeat-process-recovery.test.ts`: a fresh
  execution path appearing after the scan's own check aborts the escalation
  instead of stranding the live run (verified red without the fix, green
  with it).

All touched suites pass locally: `heartbeat-process-recovery.test.ts`,
`cross-issue-influence-limit.test.ts`, `issue-comment-reopen-routes.test.ts`,
`issue-agent-mutation-ownership-routes.test.ts`, `issue-recovery-actions.test.ts`,
and `services/recovery/*` (395 tests), plus a clean `tsc --noEmit` for the
`server` package.

### DIFF
--- a/server/src/__tests__/cross-issue-influence-limit.test.ts
+++ b/server/src/__tests__/cross-issue-influence-limit.test.ts
@@ -198,7 +198,10 @@ describe("cross-issue influence limit rollout", () => {
     expect(fake.inserted).toEqual([]);
   });
 
-  it("fails closed when the persisted run has no source issue", async () => {
+  it("counts instead of failing closed when the persisted run has no source issue", async () => {
+    // A run with an empty contextSnapshot (no issueId/taskId at all) must be
+    // treated as a "no home issue" run, not refused outright: it still gets
+    // capped at CROSS_ISSUE_INFLUENCE_LIMIT writes, starting from zero.
     const fake = counterDb(0, { contextSnapshot: {} });
 
     await expect(observeCrossIssueInfluence(fake.db as never, {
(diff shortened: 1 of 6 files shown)


dzianisv:
Superseded by a re-opened PR from a correctly-named branch (this branch name embedded an internal Paperclip ticket id, against CONTRIBUTING.md).

PR #11505, opened 2026-08-17
feat: add idempotencyKey to POST /api/companies/:companyId/approvals

rotem-zecharia:
## Problem

Two concurrent agent heartbeats posting to `POST /api/companies/:companyId/approvals` with identical payloads both received a new approval ID — no server-side dedup. The check-before-create pattern is a racy read-before-write that does not prevent concurrent creates.

## Solution

Add an optional `idempotencyKey` field to the approvals endpoint. When provided, duplicate requests within the same company return the original approval (200) instead of creating a new one (201).

## Changes

- **DB migration 0219:** add nullable `idempotency_key` column + partial unique index `(company_id, idempotency_key) WHERE NOT NULL`
- **Shared validator:** extend `createApprovalSchema` with optional `idempotencyKey`
- **Service:** add `createIdempotent` — pre-checks key, uses `onConflictDoNothing`, falls back to re-select on race condition
- **Route:** use `createIdempotent`; skip side-effects and return 200 on duplicate; return 201 on new creation
- **Tests:** update route mock, add idempotency and backward-compat test cases; add service-level unit tests for `createIdempotent`

## Backward compatibility

Fully backward-compatible: `idempotencyKey` is optional. Existing callers without the field continue to get 201 with a new approval.

## Related

Fixes duplicate-handoff race condition tracked in INUA-5613.

### DIFF
--- a/packages/db/src/migrations/0219_approvals_idempotency_key.sql
+++ b/packages/db/src/migrations/0219_approvals_idempotency_key.sql
@@ -1,0 +1,2 @@
+ALTER TABLE "approvals" ADD COLUMN "idempotency_key" text;--> statement-breakpoint
+CREATE UNIQUE INDEX "approvals_company_idempotency_key_idx" ON "approvals" ("company_id","idempotency_key") WHERE "idempotency_key" IS NOT NULL;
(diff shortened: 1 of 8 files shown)


rotem-zecharia:
Superseded by #11537 — cleaner, focused PR with only the approval idempotency changes.

PR #11510, opened 2026-08-17
fix(github-fetch): bound GitHub requests and response bodies at the boundary

harjothkhara:
## Thinking Path

> - Companies can install skills and import company data from a GitHub or GitHub Enterprise repository. `ghFetch` in `server/src/services/github-fetch.ts` is the single wrapper for every GitHub request the server makes
> - The host is not trusted input. It comes from a user-supplied repository URL through `new URL(...).hostname`, and `gitHubApiBase` builds `https://<host>/api/v3` for any non-`github.com` hostname, with no allowlist
> - `ghFetch` called `fetch()` with no signal. Node's `fetch` has no default timeout, so a host that accepts the TCP connection and then never responds blocks the call forever
> - All seven call sites run inside request handling and none of them passes a signal, so one wedged host parks a request handler indefinitely
> - The response bodies were unbounded too. Every caller read them with `response.text()`, `response.json()` or `Buffer.from(await response.arrayBuffer())`, so a host could return a body of any size and the server would buffer all of it
> - This pull request bounds every GitHub request in time with `AbortSignal.timeout`, and bounds every response body in size by streaming it through readers that stop at a ceiling
> - The benefit is that a hostile or broken GitHub Enterprise host can neither hold a request handler open nor make the server allocate without limit

No existing issue — describing the problem here, following `.github/ISSUE_TEMPLATE/bug_report.yml`.

Related PR: #9871 (`feat: inject GITHUB_TOKEN auth header in ghFetch`) is open and changes the same function and the same two files, for a different reason (auth header injection). It is not a duplicate of this change and neither supersedes the other, but they will conflict textually. This PR adds its tests at `server/src/services/github-fetch.test.ts`, the same path #9871 uses, so whichever lands first the other resolves into one test file rather than leaving two test files for one module. Happy to rebase on top of #9871 if maintainers prefer to take that one first.

**What happened?**

Two related problems at the same trust boundary.

First, `ghFetch` issues every GitHub and GitHub Enterprise request with no timeout. A host that completes the TCP handshake and then never sends a response makes the request hang for as long as the connection stays open. The affected call sites are `fetchText`, `fetchOptionalText`, `fetchBinary` and `fetchJson` in `server/src/services/company-portability.ts`, `fetchText` and `fetchJson` in `server/src/services/company-skills.ts`, and the raw file read in `server/src/services/skills-catalog.ts`. None of them passes an abort signal, and none of them applies its own time bound.

Second, none of those callers bounds the size of the response body. `fetchText`, `fetchOptionalText` and `fetchJson` read it with `response.text()` / `response.json()`, and `fetchBinary` in `company-portability.ts` and the catalog read in `skills-catalog.ts` use `Buffer.from(await response.arrayBuffer())`. A host that returns an unbounded body makes the server buffer all of it. Measuring the result afterwards cannot help, because `arrayBuffer()` has already allocated everything by the time its length is available.

The hostname reaches `ghFetch` from a user-supplied repository URL. `parseGitHubRepoUrl` takes `new URL(value).hostname`, and `gitHubApiBase` maps any hostname other than `github.com` to `https://<hostname>/api/v3`. There is no allowlist, so any reachable host can be pointed at.

**Expected behavior**

A GitHub request fails after a bounded time, and a GitHub response body is refused once it passes a size ceiling. Both surface as the wrapper's existing 422, not as a raw abort error.

**Steps to reproduce**

1. Start a server that listens and never answers, for example `node -e "require('node:http').createServer(() => {}).listen(3000,'127.0.0.1')"`.
2. Call any code path that reads a skill or company export from a repository on that host.
3. The request never returns. The added test reproduces this as a hang.
4. For the size half, point a caller at a host that streams a body without a `content-length`. The read continues until memory is exhausted. The added test reproduces this too: with the cap removed it hangs until the test timeout.

`master` at 6d0adbfb.

**Deployment mode**

Not deployment-specific — the code path runs wherever skills or company data are read from GitHub.

## What Changed

- `server/src/services/github-fetch.ts`: `ghFetch` now bounds every request with `AbortSignal.timeout`. The default is 30s, overridable with `PAPERCLIP_GITHUB_REQUEST_TIMEOUT_MS`, parsed with the same positive-integer rule used for the `PAPERCLIP_MCP_GATEWAY_*_MS` settings. A caller-supplied `init.signal` is still honoured by composing both with `AbortSignal.any`.
- `server/src/services/github-fetch.ts`: added `readGitHubResponseBytes`, `readGitHubResponseText` and `readGitHubResponseJson`. They reject early when `content-length` declares more than the ceiling, then stream the body and stop as soon as the running total crosses it, so an omitted or understated length is still bounded. The ceiling defaults to 10 MiB and is overridable with `PAPERCLIP_GITHUB_MAX_RESPONSE_BYTES`.
- `server/src/services/github-fetch.ts`: the readers translate a body-read abort into the wrapper's existing 422. `ghFetch` returns once the headers arrive, so its timeout can fire during the body read, outside `ghFetch`'s own `try`. Without this the caller saw a raw `The operation was aborted due to timeout`. This is the issue Greptile raised on the first revision of this PR.
- `server/src/services/company-skills.ts`, `server/src/services/company-portability.ts`, `server/src/services/skills-catalog.ts`: all seven body reads now go through the capped readers. No `response.text()`, `response.json()` or `response.arrayBuffer()` call on a `ghFetch` response remains in these files.
- `server/src/services/company-portability.ts`: `resolveSource` now keys its `main` → `master` fallback off `companyMarkdown === null` instead of off a thrown error. `fetchOptionalText` reports a missing file as `null`, so an absent `main` ref never reached the old `catch` — the fallback could not do the job it was named for. What did reach it was every transport failure, which was relabelled "GitHub ref main not found" and retried against `master`, hiding the original cause and doubling the work a wedged host could extract. This is the second half of Greptile's P1 on this PR.
- `server/src/services/github-fetch.test.ts`: new test file, 7 tests.
- `server/src/services/company-portability-github-ref-fallback.test.ts`: new test file, 2 tests, covering both directions of the fallback.
- `docs/deploy/environment-variables.md`: documented both new variables in the Server Configuration table.

## Verification

```sh
cd server
npx vitest run src/services/github-fetch.test.ts
```

Run on Node 24.16.0 (the repo now requires Node 24), re-verified after rebasing onto current master (8c8934044). Three of the seven tests are genuine red/green; each was confirmed to fail with only its own guard removed:

| Test | Without the fix | With the fix |
|---|---|---|
| `gives up on a host that accepts the connection but never responds` | hangs, killed by vitest at 5009ms | rejects, suite passes |
| `rejects a body whose declared content-length exceeds the limit` | resolves in 8ms, no rejection | rejects with the limit message |
| `stops a body that streams past the limit without declaring a content-length` | hangs to the 8s test timeout — `arrayBuffer()` on an unbounded stream never returns | rejects with the limit message |
| `reports a mid-body stall as a connection error, not a raw abort` | throws `The operation was aborted due to timeout` | throws the wrapper's 422 |

All 7 tests pass in 1.08s.

The portability fallback is covered separately, and both of its tests fail on the previous logic:

```sh
npx vitest run src/services/company-portability-github-ref-fallback.test.ts
```

| Test | On the old logic | With the fix |
|---|---|---|
| `falls back to master when the main ref is absent` | fails — the old `catch` never ran for an absent ref, because `fetchOptionalText` returns `null` rather than throwing | passes; the warning is emitted and `master` is requested |
| `does not retry master when the main request fails for a transport reason` | fails — the transport error was relabelled a missing ref and `master` was fetched | passes; the 422 surfaces and `master` is never requested |

The remaining three tests — `still returns a response when the host answers`, `reads a normal body through the capped readers`, and `honors a caller-supplied abort signal` — pass with and without the change. They are regression guards for the normal path and for existing cancellation, not evidence of a bug.

Consumers of `ghFetch`, compared against this PR's own first revision (the timeout commit alone) so the size-cap change is isolated:

```sh
npx vitest run src/__tests__/company-portability.test.ts src/__tests__/company-portability-routes.test.ts \
  src/__tests__/company-portability-import-batching.test.ts src/__tests__/company-skills.test.ts \
  src/__tests__/company-skills-service.test.ts src/__tests__/company-skills-catalog-service.test.ts \
  src/__tests__/skills-catalog-service.test.ts src/services/github-fetch.test.ts
```

- before the size-cap commit: `11 failed | 209 passed | 1 skipped (221)`
- after: `11 failed | 213 passed | 1 skipped (225)`

The portability suites are green in full, including the two new tests:

```sh
npx vitest run src/services/company-portability-github-ref-fallback.test.ts src/services/github-fetch.test.ts \
  src/__tests__/company-portability.test.ts src/__tests__/company-portability-routes.test.ts \
  src/__tests__/company-portability-import-batching.test.ts
# Test Files 5 passed (5) | Tests 133 passed | 1 skipped (134)
```

The same 11 tests fail either way. They are project-workspace skill-folder tests that do not touch GitHub, and they fail locally on `master` too. The delta is the 4 added tests, all passing.

`tsc --noEmit -p server/tsconfig.json` reports the same 139 pre-existing errors as clean `master`, none in the four files changed here. They are all unresolved `@paperclipai/plugin-sdk` imports caused by workspace packages not being built locally.

## Risks

Low risk. Neither change alters a function signature, and both failure modes surface as the 422 the wrapper already raised.

Two behavioral changes worth naming:

- A GitHub request that legitimately takes longer than 30s now fails. `PAPERCLIP_GITHUB_REQUEST_TIMEOUT_MS` raises the bound without a code change.
- A `main` ref that fails for a transport reason no longer silently retries `master`. An import that previously appeared to succeed via that accidental retry will now report the transport error instead. That is the intended correction — the retry was masking a real failure — but it is a visible behavior change.
- A GitHub response body larger than 10 MiB is now refused. `PAPERCLIP_GITHUB_MAX_RESPONSE_BYTES` raises the ceiling. 10 MiB was chosen to sit well above both the 1 MiB per-file limit skill imports already enforce (`MAX_CATALOG_FILE_BYTES` in `company-skills.ts`) and a large recursive repo-tree listing, so a working setup should not hit it. A tighter default would be easy to argue for and easy to change.

`AbortSignal.any` requires Node 20+, which matches the `engines` field in `package.json`.

## Model Used

Claude Opus 5 (Anthropic), model ID `claude-opus-5`, used via Claude Code with extended thinking and tool use (repository search, file editing, and local test execution). The fix and the tests were produced with model assistance and verified locally by running the test suites and typechecker as described in Verification.

## Checklist

- [x] I have updated relevant documentation to reflect my changes — both new variables are documented in `docs/deploy/environment-variables.md`
- [ ] All Paperclip CI gates are green — to be confirmed after CI runs

### DIFF
--- a/docs/deploy/environment-variables.md
+++ b/docs/deploy/environment-variables.md
@@ -28,6 +28,7 @@ All environment variables that Paperclip uses for server configuration.
 | `PAPERCLIP_HIDDEN_SETTINGS` | (unset) | Comma-separated settings surfaces to hide from the UI and floor at the API, for operators hosting Paperclip for others (managed cloud, internal shared server). See [Hiding settings surfaces](#hiding-settings-surfaces). |
 | `PAPERCLIP_SETTING_DEFAULTS` | (unset) | JSON object replacing the schema default of selected instance settings, for hosting operators. See [Operator setting defaults](#operator-setting-defaults). |
 | `PAPERCLIP_GITHUB_REQUEST_TIMEOUT_MS` | `30000` | Timeout for each GitHub / GitHub Enterprise request (skill imports, company exports). Raise it for slow GitHub Enterprise instances. |
+| `PAPERCLIP_GITHUB_MAX_RESPONSE_BYTES` | `10485760` | Maximum size of a single GitHub / GitHub Enterprise response body. Raise it if a legitimate import exceeds 10 MiB. |
 
 Daytona connectivity for `paperclip_runner` uses authenticated provider
 WebSocket ingress and follows the instance experimental setting
--- a/docs/deploy/environment-variables.md
+++ b/docs/deploy/environment-variables.md
@@ -27,6 +27,7 @@ All environment variables that Paperclip uses for server configuration.
 | `PAPERCLIP_RUNNER_REMOTE_PROVIDER_PACK_PATH` | (unset) | Host-local path to the immutable provider pack built by `pnpm --filter @paperclipai/paperclip-runner build:provider-pack`. The pack includes its target-built Node 24.11 runtime, locked production dependencies, OpenCode proxy/executable, and ACPX sidecar. Remote OpenCode and ACPX fail closed without it. A preinstalled pack is accepted only when its complete digested manifest matches this build-owned pack; otherwise Paperclip stages this pack into the sandbox. |
 | `PAPERCLIP_HIDDEN_SETTINGS` | (unset) | Comma-separated settings surfaces to hide from the UI and floor at the API, for operators hosting Paperclip for others (managed cloud, internal shared server). See [Hiding settings surfaces](#hiding-settings-surfaces). |
 | `PAPERCLIP_SETTING_DEFAULTS` | (unset) | JSON object replacing the schema default of selected instance settings, for hosting operators. See [Operator setting defaults](#operator-setting-defaults). |
+| `PAPERCLIP_GITHUB_REQUEST_TIMEOUT_MS` | `30000` | Timeout for each GitHub / GitHub Enterprise request (skill imports, company exports). Raise it for slow GitHub Enterprise instances. |
 
 Daytona connectivity for `paperclip_runner` uses authenticated provider
 WebSocket ingress and follows the instance experimental setting
--- a/docs/deploy/environment-variables.md
+++ b/docs/deploy/environment-variables.md
@@ -19,6 +19,7 @@ All environment variables that Paperclip uses for server configuration.
 | `PAPERCLIP_DEPLOYMENT_MODE` | `local_trusted` | Runtime mode override |
 | `PAPERCLIP_DEPLOYMENT_EXPOSURE` | `private` | Exposure policy when deployment mode is `authenticated` |
 | `PAPERCLIP_API_URL` | (auto-derived) | Paperclip API base URL. When set externally (e.g., via Kubernetes ConfigMap, load balancer, or reverse proxy), the server preserves the value instead of deriving it from the listen host and port. Useful for deployments where the public-facing URL differs from the local bind address. |
+| `PAPERCLIP_GITHUB_REQUEST_TIMEOUT_MS` | `30000` | Timeout for each GitHub / GitHub Enterprise request (skill imports, company exports). Raise it for slow GitHub Enterprise instances. |
 
 ## Secrets
 
--- a/docs/deploy/environment-variables.md
+++ b/docs/deploy/environment-variables.md
@@ -20,6 +20,7 @@ All environment variables that Paperclip uses for server configuration.
 | `PAPERCLIP_DEPLOYMENT_EXPOSURE` | `private` | Exposure policy when deployment mode is `authenticated` |
 | `PAPERCLIP_API_URL` | (auto-derived) | Paperclip API base URL. When set externally (e.g., via Kubernetes ConfigMap, load balancer, or reverse proxy), the server preserves the value instead of deriving it from the listen host and port. Useful for deployments where the public-facing URL differs from the local bind address. |
 | `PAPERCLIP_GITHUB_REQUEST_TIMEOUT_MS` | `30000` | Timeout for each GitHub / GitHub Enterprise request (skill imports, company exports). Raise it for slow GitHub Enterprise instances. |
+| `PAPERCLIP_GITHUB_MAX_RESPONSE_BYTES` | `10485760` | Maximum size of a single GitHub / GitHub Enterprise response body. Raise it if a legitimate import exceeds 10 MiB. |
 
 ## Secrets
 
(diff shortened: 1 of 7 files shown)


harjothkhara:
Thanks — both halves of that P1 are now addressed.

**Raw abort escaping the error contract** (39a10fc0). The capped readers translate a body-read abort into the wrapper's 422. Confirmed by removing only that translation: the caller then sees `The operation was aborted due to timeout` instead of the mapped error.

**The portability fallback** (569165c2). You were right that mapping the error was not enough, and the cause is worse than a misclassification.

`fetchOptionalText` reports a missing file as `null`, not as a throw. So an absent `main` ref never reached that `catch` — it fell through to "missing COMPANY.md". The fallback could not do the job it was named for. What did reach the `catch` was every real transport failure, each relabelled `GitHub ref main not found` and retried against `master`.

The fallback is now keyed off `companyMarkdown === null`, so it fires exactly when the ref is absent, and transport failures surface as themselves.

Both directions are covered in `server/src/services/company-portability-github-ref-fallback.test.ts`, and both tests fail on the previous logic:

- `falls back to master when the main ref is absent` — failed before, because the old `catch` never ran for an absent ref.
- `does not retry master when the main request fails for a transport reason` — failed before, because `master` was fetched.

One behavior change worth flagging for review: an import that previously appeared to succeed through that accidental retry will now report the transport error instead. That is the intended correction, but it is visible.

Full portability suites are green: `Test Files 5 passed (5) | Tests 133 passed | 1 skipped (134)`.

PR #11519, opened 2026-08-17
fix(server): re-resolve managed instructions root at heartbeat config-assembly

dzianisv:
## Thinking Path

> - Agents can load role instructions from a "managed instructions bundle" stored under the Paperclip instance root
> - `server/services/agent-instructions.ts` already re-derives and self-heals that root on every read through the bundle HTTP API, but the heartbeat run-execution path never called that recovery logic
> - When the Paperclip instance root moves (for example a data-dir rename), the persisted `instructionsRootPath`/`instructionsFilePath` go stale, and every managed agent silently runs with the generic prompt template instead of its real instructions
> - This needs to be fixed once at the server choke point where `adapterConfig` enters the execution pipeline, not inside each of the 9 adapter packages, so the fix cannot be missed by a new adapter
> - This pull request adds a small exported helper in `agent-instructions.ts` that reuses the existing recovery logic, and calls it from `heartbeat.ts` right after the adapter config is parsed, correcting the config in place before it reaches any adapter
> - The benefit is that a moved instance root no longer orphans managed instructions at run time; the same self-healing that already worked for the bundle API now also protects live heartbeats, with a log line so the correction is operator-visible

Refs #11391 (item 2: "Managed instructions root persisted as a stale absolute path"). This PR implements exactly the "suggested fix" in that issue for item 2: re-resolve the managed root via the same recovery logic used by the bundle API, at the point where the adapter config is assembled, instead of trusting the persisted absolute string. Item 1 in that issue (silent read-failure surfacing) is a separate, already-open fix (#11396) and is out of scope here — this PR touches none of the files that PR touches.

- [x] I have searched the GitHub PR list for similar PRs (searched "instructions root", "managed instructions", "stale root" — found #11396, which addresses a different part of #11391 and does not overlap with these files).

## What Changed

- `server/src/services/agent-instructions.ts`: export a new `resolveHeartbeatManagedInstructionsPatch(agent)` that composes the existing (previously unexported) `deriveBundleState` + `recoverManagedBundleState` — no reimplementation of their on-disk recovery/precedence rules. Returns `{}` for `external`-mode configs and already-correct `managed` configs; otherwise returns `{ instructionsRootPath, instructionsFilePath, warnings }` reflecting the current on-disk managed root.
- `server/src/services/heartbeat.ts`: call the new function right after `const config = parseObject(agent.adapterConfig);` and apply the patch to `config` in place, before it flows into `workspaceManagedConfig` / `mergedConfig` / `resolvedConfig` / `runtimeConfig` and on to `adapter.execute()`. Log any correction via the existing pino `logger` with `agentId` / `companyId` / old and new paths.
- No changes to any adapter package. The fix sits entirely upstream of `adapter.execute()`, so it applies uniformly to all 9 local adapters without touching any of them.

## Verification

```
# exit 0

npx vitest run server/src/__tests__/agent-instructions-service.test.ts
# 11 passed — includes the repro: managed config recorded under instance root A,
# instance root re-pointed to B via PAPERCLIP_HOME/PAPERCLIP_INSTANCE_ID env swap,
# real instructions file placed under the *current* managed root for B, function
# returns the corrected root/path for B (not the stale one from A)

npx vitest run server/src/__tests__/pi-local-execute.test.ts
# 4 passed — new test proves execute() actually consumes a heartbeat-corrected
# config.instructionsFilePath end to end (the injected system prompt contains
# the instructions content from the corrected root, not the stale one)

npx vitest run server/src/__tests__/heartbeat-*.test.ts server/src/__tests__/agent-instructions-service.test.ts
# 45 test files, 620 tests passed — no regression for external-mode configs or
# already-correct managed configs
```

## Risks

- Low risk. The correction only fires when a managed-mode config's persisted root disagrees with what `recoverManagedBundleState` finds on disk (the same logic the bundle HTTP API already trusts). `external`-mode configs and already-correct `managed` configs pass through byte-identical — proven by dedicated tests.
- No schema or API change, no adapter package touched, no new dependency.
- Adds one `logger.info` call per heartbeat only in the (rare) correction case; no log volume increase in the steady state.

## Model Used

Claude (Anthropic), via the Paperclip `pi-local` adapter driving this agent session. Extended reasoning/tool-use mode (file read/edit/bash tools), no fixed context-window cap reported by the harness.

## Checklist

- [ ] I have updated relevant documentation to reflect my changes — not applicable; internal server config-assembly fix with no user-facing or operator-facing documentation surface.
- [x] All Paperclip CI gates are green — verified with `gh pr checks 11519` at the time of this edit.
- [x] Greptile is 5/5 with no open P2s, recommendations, or follow-ups — confirmed via the Greptile Review comment: "Confidence Score: 5/5", no open P2s or follow-ups listed.

### DIFF
--- a/server/src/__tests__/agent-instructions-service.test.ts
+++ b/server/src/__tests__/agent-instructions-service.test.ts
@@ -2,7 +2,7 @@ import fs from "node:fs/promises";
 import os from "node:os";
 import path from "node:path";
 import { afterEach, describe, expect, it } from "vitest";
-import { agentInstructionsService } from "../services/agent-instructions.js";
+import { agentInstructionsService, resolveHeartbeatManagedInstructionsPatch } from "../services/agent-instructions.js";
 
 type TestAgent = {
   id: string;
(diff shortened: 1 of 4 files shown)


dzianisv:
Addressed: pushed `e2d4d3874`, which also corrects `config.instructionsEntryFile` (not just `instructionsRootPath`/`instructionsFilePath`). `resolveInstructionsConfigFingerprintMetadata` resolves the session-freshness fingerprint path from `instructionsEntryFile` in preference to `instructionsFilePath`, so leaving it stale would have defeated the fingerprint correction even though the adapter's own read was fixed. Added a regression test where the configured entry file no longer exists on disk and recovery must fall back to a different one, proving the patch surfaces the corrected entry-file name.

PR #11524, opened 2026-08-17
feat(runtime-exposure): least-privilege Tailscale HTTPS broker, shared contract, and persisted exposure state

cryppadotta:
## Thinking Path

> - Paperclip starts and supervises managed runtime services for a project's execution workspaces, so an agent's branch can be previewed while it works
> - Those services only listen on plain loopback HTTP. A person on another device, or on a phone, cannot open the preview
> - A Tailscale HTTPS mapping solves this, but `tailscale serve` needs host privileges that the Paperclip server process must not hold
> - This pull request adds the foundation only: a separate least-privilege host broker, the shared exposure contract, and the database columns that hold exposure state
> - Nothing calls the broker yet, so there is no behavior change. The benefit is that the privileged surface is small, reviewable, and isolated before any lifecycle code depends on it

No public GitHub issue exists. The change follows the feature request template.

**Subsystem affected**

Managed workspace runtime services, the shared type and validator package, and the database schema.

**Problem or motivation**

A managed runtime service binds to loopback only. There is no supported way to reach that preview from another device. Adding HTTPS directly to the server would mean the server process runs `tailscale serve`, which needs privileges far wider than the task requires. A compromised or buggy server could then map any port to the tailnet.

**Proposed solution**

Split the privileged work into a separate broker process with a narrow protocol, and define one shared contract that the server, the UI, the runtime, and the broker all read. Land this foundation first, with no caller, so the privileged code can be reviewed on its own.

**Alternatives considered**

- Call `tailscale serve` from the server process. This was rejected because it gives the server unrestricted mapping authority.
- Use `sudo` for single `tailscale` commands. This was rejected because the argument list is the only guard, and it is easy to widen by accident.
- Use a generic reverse proxy. This was rejected because it does not remove the need for a privileged Tailscale mapping step.

**Roadmap alignment**

This supports the existing managed workspace runtime capability. It adds no new product surface on its own.

**Additional context**

The broker is the security boundary of the feature, so it is deliberately the first slice. Three later pull requests build on it: the server exposure lifecycle, the runtime lease and recovery integration, and the leased-port mediator.

## What Changed

- Add the `@paperclipai/tailscale-https-broker` workspace package. The broker listens on a unix socket, authorizes each peer with `SO_PEERCRED`, and answers a small request protocol.
- Restrict what the broker will map. It accepts only same-number HTTPS-to-loopback pairs inside the Paperclip port range, refuses protected ports, and confirms that the loopback port belongs to a Paperclip-owned listener.
- Parse every request with a strict JSON reader that rejects duplicate keys, prototype keys, and unknown fields.
- Write an append-only audit record for each broker decision.
- Add the shared exposure contract in `@paperclipai/shared`: the `RuntimeExposureConfig`, `RuntimeExposureState`, and `RuntimeExposureStatus` types, their zod validators, the app and HMR port rules, and the loopback-bind helpers.
- Persist exposure state on `workspace_runtime_services` with the new `exposure` column, plus the server-private `exposure_handle` and `backend_url` columns that are never serialized to API clients.
- Add the `execution_workspace_runtime_leases` table that the later lease slice uses.
- Extend the runtime read-model test fixture for the three new columns.

## Verification

Focused checks, all run on this branch:

- `pnpm --filter @paperclipai/tailscale-https-broker test` — 12 files, 82 tests pass. This covers peer credentials, port policy, protected ports, the serve config writer, the strict JSON reader, argv parsing, and the socket server.
- `pnpm --filter @paperclipai/tailscale-https-broker typecheck` — clean.
- `npx vitest run --root packages/shared src/runtime-exposure src/validators/runtime-exposure.test.ts` — 3 files, 40 tests pass.
- `pnpm --filter @paperclipai/db typecheck` — runs `check:migrations` first. Migration numbering and migration safety both pass.
- `pnpm --filter @paperclipai/shared typecheck` — clean.
- `pnpm --filter @paperclipai/ui typecheck` — clean.
- `npx vitest run --root server src/services/workspace-runtime-read-model.test.ts` — 3 tests pass.
- `npx tsc --noEmit -p server/tsconfig.json` — 139 errors, which is exactly the count on `master` before this branch. All 139 come from the unbuilt `@paperclipai/plugin-sdk` package.

To confirm the exposure state is inert, start a managed runtime service as usual. The new columns stay null and the service behaves as it does today.

## Risks

- Migration risk is low. Both migrations only add a table and three nullable columns. No column is backfilled and no existing column changes. The migration safety check passes.
- Behavior risk is low. No code path calls the broker in this pull request, and the shared exposure fields are optional.
- The broker is privileged, so it is the real risk surface. It is mitigated by peer-credential authorization, a fixed port range, a protected-port deny list, same-number pair enforcement, listener-ownership checks, strict JSON parsing, and an audit trail. Reviewers should read `packages/tailscale-https-broker/src/authorization.ts` and `src/port-policy.ts` closely.
- The broker requires a `tailscale` version floor, which its README records. An older host CLI makes the broker refuse to start rather than map incorrectly.
- `pnpm-lock.yaml` changes because a new workspace package is added. The diff is the new importer block, plus one duplicate `tinyexec` entry that pnpm removed.

> For core feature work, check [`ROADMAP.md`](ROADMAP.md) first and discuss it in `#dev` before opening the PR.

## Model Used

## Checklist


cryppadotta:
Thanks — both rounds found real defects. Taking them separately, because one is fixed and one is out of scope by design.

### 2. Compensation removed an adopted port without proving ownership — **fixed**

This was a genuine regression I introduced. Adding an idempotently-skipped port to `appliedPorts` made compensation run `serve --https=<port> off` on a port this request never published, so a concurrently replaced Serve entry could have been deleted. That is exactly the provenance-blind deletion the protected-port rules exist to prevent.

The withdrawal path is now provenance-aware. A port this request published is known to be ours and stays eligible. A port merely *found* already correct is re-read from Serve and must still be our exact same-number loopback entry; if it is not — or if Serve cannot be read at all — the entry is left untouched and the port is quarantined instead, so it is never silently reused either.

New test: `never removes an adopted port whose Serve entry is no longer ours, and quarantines it`. It injects the foreign entry *mid-transaction*, after the `before` snapshot, because a pre-existing foreign entry is already caught earlier by `manual_mapping_present`. The test fails if the ownership check is removed.

### 1. Post-success listener replacement — **real, but not closable inside this transaction**

I do not think this one should block the PR, and I do not think it can be fixed where you are pointing.

Every check in `doExpose` is point-in-time. The Serve mapping deliberately **outlives** the transaction, and `tailscale serve` maps a *port*, not a socket — there is no way to pin an entry to an inode. So any check I add inside `doExpose` can be defeated by a swap one instruction later. Adding another re-read would not narrow the window; it would only move it and imply a guarantee the broker cannot make.

What this transaction can guarantee, and now does, is bounded and worth stating exactly:

- a **successful** expose published the socket it verified, and
- a **failed** expose leaves nothing of ours published.

Post-success replacement is a lifecycle concern, and it is handled above the broker rather than ignored:

- the server re-verifies listener ownership for the app port and its HMR companion on every readiness and health check (PR #11525), and
- reconciliation fails closed and names the conflicting workspace when a reserved pair is held by another execution workspace (PR #11526) — which is the incident that motivated this whole series.

I have written that scope boundary into the code above `currentListenerIdentity` so the limitation is explicit for the next reader rather than looking like an oversight.

If you would like continuous attestation inside the broker — for example a watchdog that re-verifies live leases and withdraws on drift — I am happy to do it, but it is a new capability with its own failure modes (it can withdraw a healthy lane on a transient `/proc` read), so it belongs in its own change rather than being smuggled into this one.

cryppadotta:
Both issues in this round are addressed, and they were useful because they are **direct opposites** — which is the signal that the test itself was wrong, not just its branches:

- **Issue 1** asks me to remove the mapping *even when* Serve status cannot be read.
- **Issue 2** asks me *not* to remove it unless provenance is proven beyond shape.

No shape-based check can satisfy both. So the withdrawal decision no longer uses shape at all.

### Provenance now comes from the broker's own registry

An already-`exposed` lease means **the broker published those ports itself**. That is real provenance, recorded under root ownership, and it is available whether or not `tailscale serve status` can be read right now.

- **Issue 1 is closed.** The withdrawal no longer depends on a live status read, so a read failure cannot return `listener_substituted` while leaving the replacement exposed. The `readServe()` call and the `live`-is-null branch you flagged are gone.
- **Issue 2 is closed.** Shape is no longer the basis for deletion. Withdrawal happens because our registry says we published the port, not because an entry happens to look like ours.
- A still-`reserved` lease published nothing, so a substitution detected on it denies **without** withdrawing anything. There is nothing of ours to remove, and removing by port number would delete someone else's mapping.

On Issue 2's operator scenario specifically: an operator who needs a mapping to survive managed lane lifecycle declares the port **protected**, and a protected port is refused far above this point — `assertNoProtectedPort` runs before any lease or ownership logic, precisely so an operator declaration outranks the broker's own ownership record. That declaration channel is the sanctioned answer here; inferring operator intent from entry shape is what destroyed `42000/52000` in the first place.

Two new tests, both of which fail without the corresponding guard:

- `withdraws a substituted idempotent mapping without depending on a readable Serve status` (Issue 1)
- `still withdraws a substituted idempotent mapping when a reserved lease published nothing` (the no-op side)

91 broker tests pass.

### Post-success supervision — still out of scope, now explicit in code

This remains architecturally unfixable inside `doExpose`: the Serve mapping outlives the transaction and `tailscale serve` maps a *port*, not a socket, so there is nothing to pin an entry to. Another re-read would move the window, not close it, and would imply a guarantee the broker cannot make.

I have written the scope boundary into the code and the commit message rather than leaving it to look like an oversight. The guarantee this transaction makes is stated exactly: **a successful expose published the socket it verified, and a failed one leaves nothing of ours published.** Continuing supervision is handled above the broker — the server re-verifies listener ownership on every readiness and health check (#11525), and reconciliation fails closed naming the conflicting workspace when a reserved pair is held by another execution workspace (#11526), which is the incident this series exists to fix.

If continuous in-broker attestation is wanted, I will build it, but it is a new capability with its own failure mode — it can tear down a healthy lane on a transient `/proc` read — so it needs its own change and its own tests rather than being added here.

PR #11525, opened 2026-08-17
feat(runtime): managed Tailscale HTTPS lifecycle, durable runtime leases, and bounded control recovery

cryppadotta:
> **Stacked pull request.** This targets #11524. Merge #11524 first. Review only the second commit, `feat(runtime): managed Tailscale HTTPS lifecycle...`.

## Thinking Path

> - Paperclip starts and supervises managed runtime services, so an agent's branch can be previewed while the agent works
> - The previous pull request added the host broker, the shared contract, and the database columns, but no code used them
> - A managed runtime can only be exposed over HTTPS if it holds a stable loopback port pair for the whole life of the service. The current control path cannot promise this: two controls can race the same execution workspace, a stranded control can stay `running` forever, and a start can adopt a port it does not own
> - This pull request adds the HTTPS lifecycle and the control-path hardening that the lifecycle depends on
> - The benefit is that a managed preview becomes reachable from another device, and a managed control now always reaches a terminal state

No public GitHub issue exists. The change follows the feature request template.

**Subsystem affected**

Managed workspace runtime services, workspace operations, the execution workspace routes, and the workspace runtime UI.

**Problem or motivation**

A managed runtime service is reachable only on loopback, so a preview cannot be opened from a phone or a second computer. Exposing it safely needs an exclusively held port pair. Three existing gaps block that. Overlapping controls can race the same workspace. A control whose owner dies stays `running` and blocks the lane forever. Port allocation does not confirm that the process holding a port is the process Paperclip spawned.

**Proposed solution**

Add the exposure lifecycle on top of the broker from #11524: reserve before spawn, expose after readiness, validate the public URL, and remove on stop. In the same change, make managed controls mutually exclusive per workspace, give each control a durable issue-owned lease and a terminal state, and verify port ownership before use.

**Alternatives considered**

- Add HTTPS exposure without the control hardening. This was rejected because a raced or stranded control makes exposure point at the wrong process.
- Guard the lane with an in-memory lock only. This was rejected because the lock does not survive a server restart, so the lane can be lost or double-claimed.
- Trust the requested bind address. This was rejected because a checkout that predates managed HTTPS overwrites `PAPERCLIP_BIND` from its own `--bind` argument, and then binds the wildcard address.

**Roadmap alignment**

This completes the managed workspace runtime capability that already exists. It adds no new product surface beyond the HTTPS link.

**Additional context**

This is the second of three pull requests. The third adds central mediation of leased port pairs.

## What Changed

Exposure lifecycle:

- Add the server-side broker client and the exposure lifecycle manager. The manager reserves the mapping before spawn, exposes after backend readiness, validates the public URL, and removes the mapping on stop.
- Default managed worktree runtimes to `tailscale_https`, read exposure intent from legacy `expose` blocks, and backfill runtimes that are still HTTP-only.
- Verify listener ownership for the app port and its Vite HMR companion before the broker is asked to expose anything. An unrelated listener on either port fails the start closed.
- Force the loopback bind through argv instead of environment hints. Leave a non-Paperclip service's `--bind` argument alone.
- Probe loopback for readiness instead of the public URL, and give Vite HMR its own loopback-bound server in middleware mode.
- Preserve operator-declared Serve mappings across the managed lifecycle, so cleanup never removes a mapping that Paperclip did not create.
- Name which listener predicate denied an expose, so an operator can act on the message.

Control-path hardening:

- Make `start`, `stop`, `restart`, and job `run` mutually exclusive per execution workspace. An overlap gets `409 workspace_runtime_control_in_progress`, and authorization is still checked first.
- Take a durable exclusivity lease on the execution workspace, owned by the controlling issue. A different issue gets `409 workspace_runtime_lease_conflict` before any operation is recorded. Board and operator actions bypass the lease.
- Give every control a terminal state. Each control stamps its owning process and pid, heartbeats while it runs, and has a wall-clock ceiling. Recovery of a stranded control uses a compare-and-swap on `updated_at`, so a live owner is never stolen.
- Bound readiness probes, verify allocated port ownership on POSIX and Windows, harden sibling port allocation, and reconcile desired runtimes on server startup.
- Surface exposure state and bounded runtime errors in the workspace runtime UI.
- Record the new behavior in `doc/DEVELOPING.md`.

## Verification

Focused checks, all run on this branch:

- `npx tsc --noEmit -p server/tsconfig.json` — 139 errors, exactly the count on `master`. All 139 come from the unbuilt `@paperclipai/plugin-sdk` package.
- `pnpm --filter @paperclipai/ui typecheck` — clean.
- Server suites, 177 tests pass across 9 files: `workspace-runtime.test.ts`, `workspace-runtime-leases.test.ts`, `workspace-runtime-control-recovery.test.ts`, `execution-workspace-runtime-control-conflict.test.ts`, `execution-workspace-runtime-lease-route.test.ts`, `workspace-operations-reconciliation.test.ts`, `workspace-runtime-start-terminality.test.ts`, `app-hmr-port.test.ts`, and `workspace-runtime-ready-comment.test.ts`.
- Exposure unit suites, 77 tests pass: `src/services/runtime-exposure/` and `workspace-runtime-exposure-backfill.test.ts`.
- UI: `WorkspaceRuntimeControls.test.tsx` and `WorkspaceServiceControlBar.test.tsx` — 34 tests pass.

**One suite is red on the development host and is expected to be green in CI.** `server/src/services/workspace-runtime-exposure.test.ts` has 10 failures on the machine used to write this branch. The cause is host contamination, not the code. That machine already runs an HTTPS canary that holds ports 42000, 42001, 52000, and 52001 on a tailnet address. The suite allocates from the same range, so the new listener-ownership check correctly reports:

```
listener_ownership_mismatch — port 42000 is bound to 100.123.243.20, 127.0.0.1,
fd7a:115c:a1e0:0:0:0:dd3a:f314 ... instead of loopback only
```

A CI runner has no listener on those ports, so the check sees loopback only and the suite passes. Please confirm this from the CI result on this pull request rather than from a local run on a host that already exposes a managed runtime. This is a real weakness of the current test fixture, and the third pull request in the series removes it by allocating the pair through a central mediator instead of a stubbed availability check.

`workspace-runtime-https-live-exercise.test.ts` needs a live `tailscale` host and was not run locally.

## Risks

- This is the behavior-bearing pull request of the three, so it carries the most risk.
- Two new `409` responses appear on managed control routes. A caller that assumed a control always starts must handle a conflict. Board and operator actions are deliberately exempt, so an agent lease cannot lock an operator out.
- Managed worktree runtimes now default to `tailscale_https`. If the host has no working broker, the start fails closed and reports the exposure failure instead of silently serving plain HTTP. This is intended, and it is the reason the failure message names the denying predicate.
- Startup reconciliation touches persisted runtime rows. It is scoped to desired state and does not resurrect a service that never came up.
- The lease has a 30-minute time to live and explicit release paths, so a crashed owner cannot hold a lane forever.
- No migration runs in this pull request. The tables and columns land in #11524.

> For core feature work, check [`ROADMAP.md`](ROADMAP.md) first and discuss it in `#dev` before opening the PR.

## Model Used

## Checklist

- [x] I have run tests locally and they pass, with the one host-contaminated suite explained above

### DIFF
--- a/doc/DEVELOPING.md
+++ b/doc/DEVELOPING.md
@@ -648,6 +648,22 @@ For project execution worktrees, Paperclip can also run a project-defined provis
 
 Heavier setup that is only needed by a managed runtime service can use `workspaceStrategy.runtimeProvisionCommand`. Paperclip runs this command lazily before spawning the first service in a start batch, serializes concurrent provisioning for the same workspace, and records the attempt as `workspace_runtime_provision`. The command receives the same workspace environment as `provisionCommand` and should be idempotent because later service-start batches invoke it again.
 
+Managed runtime control actions (`start`, `stop`, `restart`, and job `run`) are mutually exclusive per execution workspace. An overlapping control is rejected with `409 workspace_runtime_control_in_progress` instead of racing the active operation, and authorization is still checked first, so the conflict never widens who may control a workspace.
+
+Every managed control reaches a terminal operation state. Each one stamps the owning server process and pid on its `workspace_operations` row and heartbeats while it runs, and each one carries a wall-clock ceiling (30 minutes for lifecycle controls, 4 hours for workspace jobs) so a hung provider or listener fails the operation rather than leaving it active. When a start fails part-way, Paperclip tears the workspace's runtime services down through the ordinary stop path and records a stopped desired state, so the lane is retryable and a startup reconcile will not resurrect a service that never came up.
+
+Recovery of stranded controls is bounded and cannot steal a live operation. A `running` control is only terminalized when its owning process is gone, when the owning request in this process no longer exists, or after 60 seconds without a heartbeat; the terminalizing write is a compare-and-swap on `updated_at`, so an owner that heartbeats concurrently keeps its operation. Recovery runs on server startup and before each managed control, appends reconciliation evidence to the workspace-operation log, and stays inside the requested workspace's scope.
+
+Readiness probes and port allocation are bounded for the same reason. Each HTTP readiness probe is aborted after at most 5 seconds (never past the service's readiness budget), so a foreign listener that accepts a connection but never answers cannot park a start forever. Auto-allocated loopback ports are reserved in-process for the duration of a start and re-checked for a live owner, and a configured port already claimed by another in-flight start fails that start terminally — so two isolated workspaces asking for the same app/HMR pair either get distinct healthy ports or one fails cleanly and retryably.
+
+Beyond that in-flight guard, `start`, `stop`, and `restart` also take a durable exclusivity lease on the execution workspace (`execution_workspace_runtime_leases`). The lease is keyed by execution workspace and owned by the controlling issue (or, when a run has no issue in scope, by the run or agent), so it survives across calls, heartbeats, and server processes. The owning issue can keep operating; a different issue or run is rejected with `409 workspace_runtime_lease_conflict` before any workspace operation is recorded and before any runtime service is touched. Board/operator actions bypass the lease entirely and never take the lane away from an agent.
+
+Lease recovery is bounded and explicit. Another issue may reclaim the lane once the owner becomes ineligible — the owning issue reaches a terminal status, is hidden, or is deleted; the owning run reaches a terminal status — or once the lease's 30-minute TTL elapses without the owner touching it. Archiving the execution workspace releases the lease outright. Conflict responses carry the owning issue/run ids and the lease expiry so an operator can tell who holds the lane.
+
+For Tailscale HTTPS exposure, readiness includes stable listener-ownership checks for every requested loopback port (the app and, when configured, its Vite HMR companion). Each listener must belong to the spawned managed process group; an unrelated listener that races onto either reserved port fails the start closed before the broker is asked to expose it.
+
+In Vite middleware mode, Paperclip gives HMR a dedicated HTTP server bound to the managed runtime's loopback host. The browser still derives the HMR hostname from the public HTTPS page, so listener containment does not break remote hot reload.
+
 ## App-Shipped Skills Catalog
 
 The Paperclip app ships a curated catalog of company skills out of the box. The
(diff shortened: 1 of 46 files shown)


cryppadotta:
Thanks — this finding is real, and I want to be explicit that I am **not** fixing it in this PR, with reasons.

You are right about the mechanism. `Promise.race` terminalizes the *operation*, not the mutation: `recordInput.run()` has no cancellation, so after a timeout it is still touching the workspace, while the `finally` drops the in-process claim and the caller releases the lease. An abandoned start can therefore still spawn a service or claim a port while the next control runs.

### I implemented your fix, and it broke a deliberate behavior

I changed it to hold the in-process exclusivity claim until the abandoned `run()` actually settles. That does close the race. It also makes a hung start **permanently block `stop`**, because the canonical hang here is a readiness probe that never settles, so the claim is never released.

That directly contradicts an existing test that encodes the opposite decision on purpose:

> `fails a hung start on its own time budget so no active operation is left behind`
> `// No orphan active operation: a managed stop or retry goes straight through.`

and it contradicts the documented recovery path — `doc/DEVELOPING.md` states that when a start fails part-way, Paperclip tears the workspace's runtime services down **through the ordinary stop path**. If a timeout fences the lane, that teardown cannot run, and the workspace becomes uncontrollable by the exact action an operator would reach for.

So the codebase has already chosen **recoverability over strict exclusivity** here, with a test asserting it. Reversing that is a design decision with a real downside, not a bug fix, and it should not ride along inside a landing PR whose job is to move already-reviewed work onto `master`. I reverted my change.

### Existing mitigations, and what actually remains

The window is narrower than it looks:

- A failed start already runs teardown through the stop path and records desired `stopped`, so the common case is actively cleaned up rather than left racing.
- Startup reconciliation will not resurrect a service that never came up.
- The follow-up PR in this series (#11526) makes a start fail closed when a reserved port pair is held by another execution workspace, which is the concrete harm — port collision and cross-workspace adoption — that an orphaned start would otherwise cause.

What remains is a genuine narrow race between an abandoned mutation and the next control on the same lane. Closing it properly needs **cancellation plumbed into the runtime mutation** so the orphan aborts at its next commit point instead of being fenced out by a lock — that keeps `stop` available *and* prevents the interleave. That is a new capability with its own tests, not a one-line change.

I have filed that as a tracked follow-up rather than dropping it, so the trade-off gets decided deliberately instead of by whoever touches this function next.

PR #11540, opened 2026-08-17
fix: allow agents to withdraw own confirmations

tommypoltev:
## Summary
- allow the creator agent to cancel its own pending `request_confirmation` interaction
- keep cancellation authorization narrow: other agents still get `403`, board flows stay unchanged
- record the new `cancelled` request-confirmation outcome across shared types, validation, service, route, and tests

## Testing
- `pnpm vitest run packages/adapter-utils/src/server-utils.test.ts`
- `pnpm vitest run server/src/__tests__/issue-thread-interactions-service.test.ts`
- `pnpm vitest run server/src/__tests__/issue-thread-interaction-routes.test.ts`

### DIFF
--- a/package.json
+++ b/package.json
@@ -49,7 +49,7 @@
     "smoke:mcp-fixtures": "node scripts/smoke/mcp-fixture-harness.mjs",
     "smoke:pipelines-tutorial": "./scripts/smoke/pipelines-tutorial-smoke.sh",
     "smoke:terminal-bench-loop-skill": "node scripts/smoke/terminal-bench-loop-skill-smoke.mjs",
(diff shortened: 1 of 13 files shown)


jonguttman:
Flagging real-world impact: hit this exact gap in production today from a downstream instance. `POST /interactions/{interactionId}/cancel` 422s on any pending `request_confirmation` with "Only ask_user_questions interactions can be cancelled" — the diff here (widening `cancelQuestions`/`cancelInteraction` to accept `request_confirmation`, creator-only) is the precise fix. Confirmed #8081 (the earlier generalized-endpoint attempt) closed unmerged and never rolled out, and #10297 targets the same gap via a different approach. Workaround in the meantime is the `supersedeOnUserComment` auto-expiry path, which works but is comment-triggered rather than a direct cancel call. No action needed from us — just a data point that this is blocking a real agent workflow, in case it helps prioritize a rebase/merge.

PR #11552, opened 2026-08-17
fix(activity): apply since/until bounds and offset paging to company activity

### LINKED ISSUE #5893: GET /api/companies/{companyId}/activity ignores ?since= query param (returns full history every call)
## Symptom

`GET /api/companies/{companyId}/activity?since=<ISO timestamp>` returns the company's **full activity history** regardless of the `since` value passed.

Empirically verified on our deployed Paperclip (server bundle `/app/server/dist/routes/activity.js`):

```
since=NOW             → 351 events
since=1 hour ago      → 351 events
since=12 hours ago    → 351 events
since=tomorrow        → 351 events
no `since` param      → 351 events
```

Same count, same payload, every time.

## Root cause (in the bundled server source)

`routes/activity.js`:

```js
router.get("/companies/:companyId/activity", async (req, res) => {
    const companyId = req.params.companyId;
    assertCompanyAccess(req, companyId);
    const filters = {
        companyId,
        agentId: req.query.agentId,
        entityType: req.query.entityType,
        entityId: req.query.entityId,
    };
    const result = await svc.list(filters);
    res.json(result);
});
```

`req.query.since` is never read. The `activityService.list()` it calls also doesn't accept a `since` parameter — it filters only on `companyId/agentId/entityType/entityId` and orders by `desc(activityLog.createdAt)` without any time-window predicate.

## Impact on consumers

Polling clients that expect `?since=` to act as an incremental cursor end up re-processing the same events every poll. We hit this in our Discord relay bridge: once the bridge correctly matched the `issue.comment_added` action prefix, every 15s poll re-posted every comment in history because the dedupe assumption ("just fetch what's new since last poll") was silently violated. A single user query produced 16+ duplicate Discord messages before we stopped the bridge.

We worked around it client-side by adding a `Set<activityId>` dedupe primed on first poll, but the API contract should hold.

## Proposed fix

Add `since` (ISO timestamp) to the route's filters and to `activityService.list()`:

```js
const filters = {
    companyId,
    agentId: req.query.agentId,
    entityType: req.query.entityType,
    entityId: req.query.entityId,
    since: req.query.since,   // ISO timestamp
};
```

And in `services/activity.js#list`, add a `gt(activityLog.createdAt, since)` predicate when `filters.since` is present.

A `limit` parameter would also help (the unfiltered list will only grow over time).

## Acceptance

- `GET /companies/{id}/activity?since=2026-05-13T00:00:00Z` returns only entries with `createdAt > 2026-05-13T00:00:00Z`.
- `GET /companies/{id}/activity` (no since) still returns history (backwards-compatible).
- An invalid `since` value (unparseable) returns 400 rather than silently being ignored.

Happy to send a PR if it'd be useful.

ericdfields:
## Thinking Path

> - Operators need a durable record of what changed and who caused it, so Paperclip keeps a company activity log
> - Reporting tools read that log through `GET /api/companies/{companyId}/activity`
> - A report almost always wants a date window, such as one week
> - But the route read `agentId`, `entityType`, and `entityId` from the query string, and never read `since` or `until`
> - The service had no date fields on its filter type either, so the bounds had nowhere to go
> - The endpoint therefore answered every windowed request with the most recent rows, and capped the page at 500
> - The caller got a plausible response and no error, so the wrong data looked correct
> - This pull request applies the date bounds in SQL, raises the cap to 1000, and adds `offset` paging
> - The benefit is that a client can read an exact date window instead of rebuilding it from the issue and comment endpoints

Fixes #5893 — `?since=` is accepted and then ignored, so the endpoint returns the full history for every value.

Refs #5907 — an earlier open pull request for the same `since` bug. It is narrower: it adds `since` only, and leaves `until`, the 500-row cap, and paging unaddressed. This pull request covers all four. If #5907 lands first, this branch needs a rebase. See "Risks".

**One behavior difference to confirm:** #5893 proposes `createdAt > since` (exclusive). This pull request uses `>=` (inclusive), to match the sibling agent-action audit query and to make `since`/`until` read as a normal closed interval. An incremental polling client that passes its last-seen timestamp will re-receive that one boundary record. Say the word and I will switch it to exclusive.

## What Changed

- `services/activity.ts`: added `since`, `until`, and `offset` to `ActivityFilters`, and applied them as `gte` / `lte` / `.offset()` in `list()`.
- `services/activity.ts`: raised `MAX_ACTIVITY_LIMIT` from 500 to 1000, matching `ISSUE_LIST_MAX_LIMIT`.
- `services/activity.ts`: added `desc(id)` as an order tiebreaker. Without it, rows that share a `createdAt` can reorder between requests, and a paged sweep drops records.
- `routes/activity.ts`: parsed `since`, `until`, and `offset`, and forwarded them to the service.
- `routes/activity.ts`: parsed the date bounds with an explicit ISO-8601 schema instead of `z.coerce.date()`. Coercion delegates to `new Date(string)`, which also accepts host-specific forms such as `"Aug 10 2026"` and `"2026/08/10"`. Those resolve against the server's local timezone, so one request could select different windows on different hosts.
- `routes/activity.ts`: a bare date now covers the whole UTC day, and `until` lands at end-of-day. `?since=2026-08-10&until=2026-08-16` therefore means seven full days.
- `routes/activity.ts`: rejected date bounds that are not real calendar days. A shape check is not a validity check: `2026-02-31` and `2027-02-29` match the ISO pattern, and `new Date()` rolls them forward to a different day. The bound looked valid and quietly selected a window the caller never asked for. This affected the date-time forms too, not only bare dates.
- `routes/activity.ts`: a malformed bound returns `400`. Ignoring it would widen a bounded query back to everything, which is the bug being fixed.
- `docs/api/activity.md`: documented the parameters, the accepted formats, and a paging loop.

## Verification

Both suites pass locally:

```
pnpm -C server vitest run src/__tests__/activity-routes.test.ts src/__tests__/activity-service.test.ts
  Test Files  2 passed (2)
       Tests  46 passed (46)

pnpm -C server typecheck
  clean
```

The service tests run against embedded Postgres, so the SQL is executed and not mocked. They seed one event per day across Aug 8–18 and assert that a Aug 10–17 window returns exactly the seven in-window rows. Before this change the same call returned the newest rows, through Aug 18.

New cases:

- `since`/`until` apply as inclusive bounds instead of being ignored.
- A `limit` of 1000 returns 600 seeded rows, which the old 500 cap truncated.
- Paging a window with `offset` neither repeats nor drops a record.
- Paging stays stable when ten rows share one `createdAt`.
- A pinned-`until` sweep stays stable while new activity is written during the sweep.
- A table of rejected formats: `last-tuesday`, `Aug 10 2026`, `2026/08/10`, `2026-08-10T00:00:00` (no zone), and epoch millis.
- A bare date widens to the whole UTC day.
- Impossible calendar days are rejected: `2026-02-31`, `2027-02-29`, `2026-13-01`, `2026-04-31`, month 0, day 0, and the same impossible day inside a date-time with and without an offset.
- Real dates the guard must still accept: `2028-02-29` (leap year), `2026-01-31`, and offset date-times.

Manual check against a running server:

```bash
curl -s -H "Authorization: Bearer $API_KEY" \
  "$BASE/api/companies/$COMPANY_ID/activity?since=2026-08-10&until=2026-08-16&limit=1000" \
  | jq 'length, (max_by(.createdAt).createdAt), (min_by(.createdAt).createdAt)'
```

Every `createdAt` must fall inside the requested window.

## Risks

Low to moderate.

- **Behavior change for existing callers.** A caller that already sent `since` or `until` was silently ignored, and now gets the window it asked for. That is the intended fix, but the response will shrink for those callers. A caller that sent a malformed date used to get data and now gets `400`.
- **Inclusive vs exclusive `since`.** See the note under "Linked Issues". This differs from the acceptance criteria in #5893.
- **Offset paging under concurrent writes.** `offset` addresses positions in a result set that each request recomputes. A sweep is only stable if the caller pins `until` to a past timestamp, because new records carry a present `createdAt` and sort outside the window. The docs state this, and a test proves it. An unbounded sweep with no `until` can still repeat records; keyset pagination would remove that caveat and is a reasonable follow-up.
- **Query cost.** The bounds sit on `(company_id, created_at)`, which `activity_log_company_created_idx` already covers, so the window is index-backed. The raised cap lets one request return up to 1000 rows instead of 500, which roughly doubles the worst-case payload.
- **No migration.** This is a read-path change. There is no schema change and no data backfill.
- **Conflict with #5907.** Both branches edit the same two files. Whichever lands second needs a rebase.

## Category Safeguard

The class of bug here is not "someone forgot `since`." It is that the handler read query
parameters as raw casts:

```ts
agentId: req.query.agentId as string | undefined,
```

Nothing in that line connects the handler to the documented contract, so a parameter can be
published in the docs, accepted by the server, and never read — with no type error, no
validation error, and a `200` carrying plausible-looking data. That is the worst shape a bug
can take: it is invisible from the outside. This one survived three consecutive weekly reports
because the response looked fine.

The fix in this PR routes the new parameters through a `zod` query schema
(`companyActivityQuerySchema`) instead. Applied as a rule, that closes the whole class: if
every route parses `req.query` through a schema and reads only from the parsed result, a
documented-but-unread parameter becomes a visible omission in one reviewable object rather
than an absence spread across a handler.

**Proposed, not included here:** an ESLint `no-restricted-syntax` rule banning
`req.query.<name> as <type>` casts inside route handlers, requiring a validated query schema.
I did not add it to this PR because the raw-cast pattern is widespread across `server/src/routes/`,
so landing the rule means either a baseline suppression file or a sweep — both materially wider
than the bug being fixed, and both worse reviewed while stapled to it. Flagging it here for a
maintainer to take or decline as a standalone change.

## Model Used

Claude Opus (Anthropic), model id `claude-opus-5`, running in Claude Code with extended thinking and tool use enabled. The model wrote the diff, the tests, and this description. A human ran the tests and reviewed the result.

## Checklist

### DIFF
--- a/docs/api/activity.md
+++ b/docs/api/activity.md
@@ -29,6 +29,10 @@ so `?since=2026-08-10&until=2026-08-16` returns those seven full days. Any
 other format returns `400` rather than being ignored — the endpoint will not
 silently widen a bounded query back to "everything".
 
+A date that does not exist on the calendar, such as `2026-02-31` or
+`2027-02-29`, also returns `400`. It is not rolled forward to the next real
+day, because that would answer a different question than the one asked.
+
 Records come back newest-first, ordered by `createdAt` then `id`. The `id`
 tiebreaker makes a given `(limit, offset)` pair address a stable row, so
 paging through a window will not skip or repeat records when several share a
--- a/docs/api/activity.md
+++ b/docs/api/activity.md
@@ -18,6 +18,32 @@ Query parameters:
 | `agentId` | Filter by actor agent |
 | `entityType` | Filter by entity type (`issue`, `agent`, `approval`) |
 | `entityId` | Filter by specific entity |
+| `since` | ISO-8601 timestamp; only records with `createdAt >= since` (inclusive) |
+| `until` | ISO-8601 timestamp; only records with `createdAt <= until` (inclusive) |
+| `limit` | Page size, 1–1000 (default 100). Out-of-range values are clamped, not rejected |
+| `offset` | Rows to skip for pagination (default 0) |
+
+An unparseable `since`/`until` returns `400` rather than being ignored.
+
+Records come back newest-first, ordered by `createdAt` then `id`. The `id`
+tiebreaker makes a given `(limit, offset)` pair address a stable row, so
+paging through a window will not skip or repeat records when several share a
+timestamp.
+
+To pull a full window larger than one page, walk `offset` until a page comes
+back short:
+
+```bash
+OFFSET=0
+while :; do
+  PAGE=$(curl -s -H "Authorization: Bearer $PAPERCLIP_API_KEY" \
+    "$PAPERCLIP_API_URL/api/companies/$PAPERCLIP_COMPANY_ID/activity?since=2026-08-10T00:00:00Z&until=2026-08-17T00:00:00Z&limit=1000&offset=$OFFSET")
+  COUNT=$(printf '%s' "$PAGE" | jq 'length')
+  printf '%s' "$PAGE" | jq -c '.[]'
+  [ "$COUNT" -lt 1000 ] && break
+  OFFSET=$((OFFSET + 1000))
+done
+```
 
 ## Activity Record
 
--- a/docs/api/activity.md
+++ b/docs/api/activity.md
@@ -18,20 +18,33 @@ Query parameters:
 | `agentId` | Filter by actor agent |
 | `entityType` | Filter by entity type (`issue`, `agent`, `approval`) |
 | `entityId` | Filter by specific entity |
-| `since` | ISO-8601 timestamp; only records with `createdAt >= since` (inclusive) |
-| `until` | ISO-8601 timestamp; only records with `createdAt <= until` (inclusive) |
+| `since` | ISO-8601 bound; only records with `createdAt >= since` (inclusive) |
+| `until` | ISO-8601 bound; only records with `createdAt <= until` (inclusive) |
 | `limit` | Page size, 1–1000 (default 100). Out-of-range values are clamped, not rejected |
 | `offset` | Rows to skip for pagination (default 0) |
 
-An unparseable `since`/`until` returns `400` rather than being ignored.
+`since` and `until` accept a date-time (`2026-08-10T00:00:00Z`, offsets
+allowed) or a bare date (`2026-08-10`). A bare date covers that whole UTC day,
+so `?since=2026-08-10&until=2026-08-16` returns those seven full days. Any
+other format returns `400` rather than being ignored — the endpoint will not
+silently widen a bounded query back to "everything".
 
 Records come back newest-first, ordered by `createdAt` then `id`. The `id`
 tiebreaker makes a given `(limit, offset)` pair address a stable row, so
 paging through a window will not skip or repeat records when several share a
 timestamp.
 
-To pull a full window larger than one page, walk `offset` until a page comes
-back short:
+### Paging a window
+
+Always pass `until` when you page. Each request re-runs the query, so `offset`
+is only meaningful against a result set that cannot change between requests.
+Records are written with `createdAt` set to the present, and rows are returned
+newest-first, so a pinned `until` in the past freezes the result set: anything
+written mid-sweep sorts after `until` and is excluded. Without `until`, a
+concurrent write shifts every later row by one position and the sweep repeats
+records.
+
+Pin `until`, then walk `offset` until a page comes back short:
 
 ```bash
 OFFSET=0
--- a/server/src/__tests__/activity-routes.test.ts
+++ b/server/src/__tests__/activity-routes.test.ts
@@ -310,6 +310,18 @@ describe.sequential("activity routes", () => {
     ["2026/08/10", "host-specific format, no zone"],
     ["2026-08-10T00:00:00", "ISO shape but no zone, so locale-dependent"],
     ["1786060800000", "epoch millis, not ISO-8601"],
+    // `new Date()` rolls these forward rather than failing — "2026-02-31"
+    // becomes March 3 — so a shape-only check would accept a window the caller
+    // never asked for. The date-time forms matter too: Zod's `.datetime()`
+    // validates format, not calendar validity.
+    ["2026-02-31", "ISO shape, impossible day"],
+    ["2026-02-31T00:00:00Z", "impossible day in a date-time"],
+    ["2026-02-31T00:00:00+05:00", "impossible day with an offset"],
+    ["2027-02-29", "Feb 29 in a non-leap year"],
+    ["2026-13-01", "month 13"],
+    ["2026-04-31", "April has 30 days"],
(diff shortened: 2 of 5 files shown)


ericdfields:
Thanks — the calendar finding was a real bug. Fixed in `cef542a0a`, and it was worse than reported: the same gap affected the date-time forms, not just the bare-date path. `z.string().datetime()` validates format, not calendar validity, so `2026-02-31T00:00:00Z` and `2026-02-31T00:00:00+05:00` rolled forward too. The guard now validates the civil date independently of any UTC offset, since `2026-02-31` is not a date in any zone. Tests cover the impossible days and, so the guard is not merely strict, the real dates it must still accept (`2028-02-29`, offset date-times).

On the offset/concurrent-write point, I want to explain why I did not change the code rather than leave it looking ignored.

The concern is correct in general: `offset` addresses positions in a result set each request recomputes. But for the documented usage it cannot fire. Activity rows are written with `createdAt` at the present, and rows come back newest-first. So a `until` pinned to a past timestamp freezes the result set — a row written mid-sweep sorts outside the window and cannot shift any later page. There is a test that interleaves writes with paging and asserts the exact sequence (`keeps a pinned-until sweep stable while new activity is written`).

So the failure needs an *unpinned* sweep. The docs now tell callers to pin `until`, and explain why rather than just asserting it.

I chose not to add keyset pagination here for two reasons: the linked issue asks for offset paging consistent with `/companies/{id}/issues`, which is already offset-based, and switching this one endpoint to cursors would make the two inconsistent. It is listed under Risks as a reasonable follow-up. Happy to do it in this PR instead if you would rather not ship the caveat — it is a contained change, just a wider one than the bug being fixed.

One thing I would genuinely like a human opinion on, flagged in the description: #5893's acceptance criteria specifies `createdAt > since` (exclusive). I implemented `>=` (inclusive) to match the sibling agent-action audit query and to read as a normal closed interval. That means an incremental polling client passing its last-seen timestamp re-receives one boundary record. #5893's reporter was building exactly such a poller, so this is worth a deliberate call rather than my preference.

jonguttman:
Flagging real production impact from a downstream instance (WeedMenu/KEWL, ticket KEWL-4224).

An agent doing a scheduled targeted post-run incident review (read only activity since the prior run, without whole-board pagination) hit the exact wall this PR fixes: `before`, `since`, `cursor`, `beforeCreatedAt`, `order=asc`, `page`, and `offset` were all confirmed to silently no-op — verified live by diffing each against a garbage-param control (`?xyzzy=banana`), which returned byte-identical results (same 5 newest rows, same `createdAt` range) to every one of them. The review could not prove complete coverage of a ~4.5-hour incident window because there is currently no way to read anything past the newest 500 company activity events.

No new issue filed on our side — this PR already covers exactly the gap (since/until bounds + offset paging + raised cap). Would be great to see this land; happy to re-verify against a live instance once merged and rolled out.

(Registered as a `pending` capability-gap entry pointing here rather than duplicating; see #5893/#5907/#11555 for the related history you're already tracking.)

PR #11555, opened 2026-08-17
fix(activity): apply since/until filters and offset pagination to company activity endpoint

ericdfields:
## Summary

- `GET /api/companies/:companyId/activity` silently ignored `since`, `until`, and `offset` query params and hard-capped at 500 records
- Added `since`/`until` to `ActivityFilters`, applied as `gte`/`lte` WHERE conditions on `activityLog.createdAt`
- Added `offset` support via `normalizeActivityOffset`, wired route → service
- Raised `MAX_ACTIVITY_LIMIT` from 500 → 1000 and exported it
- Exported `normalizeActivityOffset` for use in the route handler

**Root cause:** Two independent omissions — `ActivityFilters` had these fields declared but `list()` never read them, and the route handler never parsed them from `req.query`.

**Impact:** Unblocks BRO-2206 weekly founder-leverage report (3 consecutive weekly runs blocked: BRO-1791, BRO-1964, BRO-2206).

## Test plan

- [ ] `?since=...&until=...&limit=1000` returns records in the date window (not just the 500 most-recent)
- [ ] `limit=1000` is honored (previously hard-capped at 500)
- [ ] `offset=500&limit=500` returns the next page correctly

**Category safeguard:** Follow-on todo: add integration test for date-filtered activity queries to catch query-param→filter→SQL gaps.

Fixes BRO-2207.

### DIFF
--- a/server/src/routes/activity.ts
+++ b/server/src/routes/activity.ts
@@ -3,7 +3,7 @@ import { z } from "zod";
 import type { Db } from "@paperclipai/db";
 import { normalizeIssueIdentifier } from "@paperclipai/shared";
 import { validate } from "../middleware/validate.js";
-import { activityService, normalizeActivityLimit } from "../services/activity.js";
+import { activityService, normalizeActivityLimit, normalizeActivityOffset } from "../services/activity.js";
 import { assertAuthenticated, assertBoard, assertCompanyAccess, getAccessibleResource, hasCompanyAccess } from "./authz.js";
 import { accessService, heartbeatService, issueService } from "../services/index.js";
 import { sanitizeRecord } from "../redaction.js";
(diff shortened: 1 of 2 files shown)


ericdfields:
Closing as a duplicate of #11552, which covers the same bug more completely.

#11552 includes everything here (the `since`/`until` WHERE bounds, `offset` paging, and the raised 500 → 1000 cap) plus what this PR is missing:

- a `desc(id)` order tiebreaker, without which an offset-paged sweep silently drops and duplicates equal-timestamp rows — the exact use case this fix exists to serve
- strict ISO-8601 bound parsing, so a malformed date returns 400 instead of reaching the query builder as an `Invalid Date`
- calendar-validity checking (`2026-02-31` would otherwise roll forward and select a window the caller never asked for)
- route + embedded-Postgres test coverage, and `docs/api/activity.md`

This PR and #11552 were opened by two separate automated runs on the same issue that each could not see the other. My mistake — apologies for the duplicate review load.

PR #11562, opened 2026-08-17
fix(server): pick the runtime API URL agents can actually reach

dzianisv:
## Problem

`choosePrimaryRuntimeApiUrl()` picks the origin agents dial as the **first non-empty entry of `server.allowedHostnames`**:

```ts
const allowedHostname = input.allowedHostnames.map(v => v.trim()).find(Boolean);
if (allowedHostname) return formatOrigin("http:", allowedHostname, input.port);
```

`allowedHostnames` is a Host-header *accept* list. Nothing in it promises that a name resolves to the address the server actually bound, so this selection is pure array order.

Observed on a Tailscale-bound host with `allowedHostnames: ["box-name", "box-name.tailnet.ts.net"]`, `bind: tailnet`, `host: 100.x.x.x`:

- index 0, `box-name`, resolves via the cloud provider's internal DNS to `10.0.0.4` — nothing listens there. Every agent run got connection-refused on its first control-plane call.
- index 1, `box-name.tailnet.ts.net`, resolves via MagicDNS to the bound address and returns `200`.

The correct answer was already in the list, one slot too late. Any tailnet/VPN-bound deployment whose machine name also exists in a cloud provider's internal DNS hits this. The loopback fallback is not a safety net either: with a non-wildcard bind, `localhost:<port>` has no listener.

Adjacent to (and distinct from) #6630, which covered conflating the two lists rather than reachability.

## Change

New `resolveRuntimeApiUrl()` in `server/src/runtime-api.ts` replaces the order-based derivation:

1. **Rank** the existing candidate list by whether each hostname resolves to the bind address. A name that does not resolve is ranked *ahead* of one that resolves elsewhere — split-horizon DNS and hosts-file entries legitimately differ between the server and its agents, so unknown is not disqualified.
2. **Probe** in that order once the listener is open, and promote the first origin that answers. Any HTTP response proves a listener, so a future auth change on `/api/health` cannot silently break the check.
3. **Fall back** to the best-ranked candidate when nothing answers, and `warn` with the full probe record.

Kept intentionally:

- An explicit `PAPERCLIP_API_URL` or `authPublicBaseUrl` is operator intent — it may name a reverse proxy this process cannot reach itself — and is **never** overridden.
- A loopback bind short-circuits before any lookup.
- A wildcard bind answers on every address, so DNS carries no ranking signal; the caller's order is left alone.

Startup calls the resolver twice: `probe: false` before `listen()` (probing your own unopened port rejects everything), then again after, promoting only if the result differs.

### Bounded startup cost

500 ms per-hostname DNS timeout, 600 ms per-probe timeout, 1200 ms total probe budget after the first candidate. `dns.lookup` has no timeout of its own, so a wedged resolver cannot stall boot. In the common case the top-ranked candidate answers on the first probe.

### Removed: `PAPERCLIP_RUNTIME_API_CANDIDATES_JSON`

Startup wrote this ordered fallback list to `process.env` for adapters to retry against. `git grep` across the tree finds **no reader** — only the write and its own tests. Startup now does that retrying itself and exports a single origin that answered, so the list has no consumer left. Deleted rather than left write-only.

## Tests

`server/src/__tests__/runtime-api.test.ts` — new suite covering:

- the reported shape: `["unreachable-name", "good-name"]` with an injected resolver/prober where only `good-name` answers ⇒ `good-name` origin;
- **order independence**: swapping the two entries yields the same URL (this is the assertion the old code cannot satisfy);
- fallthrough to the bind host, DNS-rank fallback when nothing answers, explicit-public-base-URL short-circuit, loopback bind, and probe-budget exhaustion.

`server/src/__tests__/server-startup-feedback-export.test.ts` — startup-level coverage that an allowed hostname resolving away from the bind address is not exported, plus the candidate-list removal. Uses TEST-NET-1 (`192.0.2.0/24`) and `.invalid` so it stays hermetic in CI.

Against `master`, `choosePrimaryRuntimeApiUrl` returns `http://unreachable-name:3100` for the first ordering and `http://good-name:3100` for the swapped one — order-dependent, and wrong in the reported case.

```
$ npx vitest run src/__tests__/runtime-api.test.ts src/__tests__/server-startup-feedback-export.test.ts
 Test Files  2 passed (2)
      Tests  33 passed (33)

$ npx tsc --noEmit -p server/tsconfig.json
(clean)
```

### DIFF
--- a/server/src/__tests__/runtime-api.test.ts
+++ b/server/src/__tests__/runtime-api.test.ts
@@ -3,6 +3,8 @@ import {
   buildRuntimeApiCandidateUrls,
   choosePrimaryRuntimeApiUrl,
   collectReachableInterfaceHosts,
+  rankRuntimeApiCandidatesByBindAddress,
+  resolveRuntimeApiUrl,
 } from "../runtime-api.js";
 
 describe("runtime API discovery", () => {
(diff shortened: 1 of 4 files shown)


dzianisv:
Closing in favour of #11564, which supersedes this pull request.

Same change, plus two things this branch could not carry:

- The branch name here contains an instance-local ticket id. CONTRIBUTING asks contributors not to put internal references in a contribution, and a pull request head branch cannot be renamed, so the work moved to `fix/runtime-api-reachable-url`.
- #11564 adds the identity check Greptile asked for on this pull request. The startup probe now accepts an origin only when the response carries this process's instance token, so an unrelated HTTP service on a candidate hostname can no longer be exported to agents as their control-plane URL.

The full pull request description, including the thinking path and the verification, is on #11564.

PR #11573, opened 2026-08-17
test(plugins): make the SDK test harness enforce invocation company scope

ericdfields:
## Thinking Path

> - Paperclip runs plugin code in separate worker processes, and the host answers each worker-to-host call
> - The host refuses a company-scoped call unless it made an invocation scope for that company
> - The plugin SDK also gives plugin authors an in-memory test harness, `createTestHarness`
> - This harness did not model an invocation scope at all, so the fake context answered every company-scoped call
> - A fake that permits more than the host removes the gate from the code under test, and the plugin suite becomes green on a build that the host refuses
> - This pull request makes the harness enforce the same rule from the same code, and adds a test that keeps the two together
> - The benefit is that a plugin fails in its own test suite, not on its first call to the real host

No public issue exists for this. The problem, in the bug report format:

**What happened?**

`createTestHarness` in `packages/plugins/sdk/src/testing.ts` did not model an invocation company scope. `ctx.config.get(companyId)`, `ctx.secrets.resolve(secretRef, { companyId })`, `ctx.issues.createComment(issueId, body, companyId)` and every company-scoped `ctx.state.*` call gave a result. The real host refuses all of these unless it made a scope for that company, in `requireInvocationCompanyScope` in `packages/plugins/sdk/src/host-client-factory.ts`.

Two effects follow. A plugin can pass its full test suite and then fail on its first company-scoped call to the real host. A plugin can also read or write another company's data in its tests, which hides the cross-company case from the author.

**Expected behavior**

The test harness must refuse the calls the host refuses, and it must use the same error text. A test that needs to act as one company must say so.

**Steps to reproduce**

1. Make a plugin with `createTestHarness`.
2. Register a scheduled job with `ctx.jobs.register`.
3. In the job handler, call `ctx.config.get("some-company-id")`.
4. Run the job with `harness.runJob("job-key")`.
5. The call gives a result. The real host refuses the same call with `company context is required`.

Related pull requests found in a duplicate search. Neither one changes this behavior, and neither one conflicts with this change:

- Refs #10580 — adds `ctx.runtime.runProactively()` so a long-lived callback can leave a stale invocation scope. It adds a new API to the harness. It does not make the harness enforce the scope rule.
- Refs #11571 — adds company-scoped scheduled jobs to the host. This pull request does not depend on it, and does not anticipate it: a job run gets no company scope, which matches `master` today. When that pull request or a similar one lands, `runJob` can read the company off the job context.

## What Changed

- Move `requireInvocationCompanyScope`, `requestedCompanyScope` and `resolveRequiredCompanyId` from the `createHostClientHandlers` closure to module scope in `host-client-factory.ts`, and export them. There is now one definition of the rule for the host and the harness.
- Set an active invocation scope in the harness at the entry points that carry one in production. This mirrors `deriveInvocationScope` in `server/src/services/plugin-worker-manager.ts`: `emit` uses `event.companyId`, `getData` uses the host-authorized `GetDataParams.companyId`, `performAction` uses the actor context, and `executeTool` uses the run context.
- Give a scheduled job run no company scope, because the host gives it none. Both dispatch paths in `plugin-job-scheduler.ts` call `runJob` with `{ job }` and no company, and `deriveInvocationScope` has no `runJob` branch. `runJob` throws if a test gives it a company, and the message points to `withCompanyScope` for a test that wants the scoped path. A harness that made a scope here would permit more than the host, which is the fault this pull request removes.
- Mirror the proactive branch of `contextForWorkerMessage` for calls made outside an invocation. A new `proactiveCompanyScopes` harness option holds the companies the plugin is configured for. Only those companies are permitted, as in production.
- Add `requireCompanyScope` to about 90 company-scoped fake methods in the harness, each with the worker-to-host method name that the production client sends.
- Add `harness.withCompanyScope(companyId, fn)`. A test uses it to say which company it acts as when it drives `ctx` directly, or when it calls a lifecycle hook that the harness does not wrap. It is a scope and not a bypass: a call inside it for a different company is still refused.
- Add `packages/plugins/sdk/tests/testing-invocation-scope.test.ts` with 15 tests. Each expectation uses the host's error text, not a harness error text.
- Add `packages/plugins/sdk/tests/testing-harness-scope-parity.test.ts`. It reads the company-scoped method set from the production client `worker-rpc-host.ts` and asserts that the harness gates each one. This prevents the same gap from returning.
- Update the tests that the new rule stopped. Each change says which company the test acts as. No expected message was changed and no assertion was removed.

## Verification

Automatic tests:

- `pnpm --filter @paperclipai/plugin-sdk run typecheck` — clean.
- `packages/plugins/sdk` — 64 of 64 tests pass.
- `packages/plugins/plugin-llm-wiki` — 101 of 101 tests pass.
- `packages/plugins/plugin-workspace-diff` — 26 of 26 tests pass.
- `packages/plugins/examples/plugin-authoring-smoke-example` and `plugin-orchestration-smoke-example` — all tests pass.
- `server/src/__tests__/plugin-sdk-testing.test.ts`, `plugin-sdk-orchestration-contract.test.ts`, `plugin-worker-manager.test.ts` and `plugin-execution-workspace-bridge.test.ts` — 62 of 62 tests pass.

Manual checks of the new gates. Four changes were made to the source, and each one made a test fail:

1. Remove one `requireCompanyScope` call. The parity test fails and gives the name of the method.
2. Make the harness answer every company-scoped call again. Six scope tests fail.
3. Keep the scope after `withCompanyScope` returns. Two tests fail.
4. Remove the `runJob` company guard so the harness makes a scope the scheduler cannot make. One test fails.

Findings from the test changes. Each one is a real difference between the harness and the host:

- The `wiki_write_page` tool in `plugin-llm-wiki` reads its company from its own parameters, and not from `runContext.companyId`. The host refuses this call when the agent run belongs to a different company. The test now gives the run context company, so the difference is visible.
- A session event callback runs with no invocation scope. The worker client sends `agents.sessions.event` without `runNotification`, and `onEvent` uses it. Writes from a session callback are therefore proactive. The test now declares the configured company.

### Review response

The first review found that `runJob(jobKey, { companyId })` made an invocation scope that the production scheduler cannot make. The finding is correct. Both dispatch paths send `{ job }` with no company, so the same handler runs with no scope on the real host. The harness now carries no company on a job run, and it throws if a test gives it one.

## Risks

- Test-only behavior change. No production code path changes. The exported functions have the same logic as before, and the host calls them in the same order.
- Plugin test suites outside this repository can fail after this change. Each failure shows a call that the real host also refuses, so the failure is correct. The message names the method and the company, and `harness.withCompanyScope` gives a short fix for a test that drives the context directly.
- The harness models an invocation scope with one variable, because a test runs one entry point at a time. The host also tracks the identifier of each invocation. A test that runs two entry points at the same time is therefore not modelled.
- Stream notifications stay ungated. The host drops an out-of-scope stream notification and does not raise an error, and the harness stream fakes record nothing. The parity test lists them with the reason, so the hole is visible.

## Model Used

Claude Opus 4.6 (`claude-opus-4-6`), extended thinking, with tool use and code execution in an agent harness.

## Checklist

### DIFF
--- a/packages/plugins/examples/plugin-orchestration-smoke-example/tests/plugin.spec.ts
+++ b/packages/plugins/examples/plugin-orchestration-smoke-example/tests/plugin.spec.ts
@@ -102,14 +102,18 @@ describe("orchestration smoke plugin", () => {
     expect(harness.dbExecutes[0]?.sql).toContain(".smoke_runs");
     expect(harness.dbQueries.some((entry) => entry.sql.includes("JOIN public.issues"))).toBe(true);
 
-    const relations = await harness.ctx.issues.relations.get(result.childIssueId, companyId);
+    // Driven straight off `ctx`, outside any host invocation, so the assertion
+    // has to say which company it reads as.
+    const relations = await harness.withCompanyScope(companyId, () =>
+      harness.ctx.issues.relations.get(result.childIssueId, companyId));
     expect(relations.blockedBy).toEqual([
       expect.objectContaining({
         id: result.blockerIssueId,
         status: "done",
       }),
     ]);
-    const docs = await harness.ctx.issues.documents.list(result.childIssueId, companyId);
+    const docs = await harness.withCompanyScope(companyId, () =>
+      harness.ctx.issues.documents.list(result.childIssueId, companyId));
     expect(docs).toEqual([
       expect.objectContaining({
         key: "orchestration-smoke",
@@ -135,7 +139,10 @@ describe("orchestration smoke plugin", () => {
     });
     await plugin.definition.setup(harness.ctx);
 
-    await expect(plugin.definition.onApiRequest?.({
+    // The host mints an invocation scope from `ApiRequestParams.companyId`
+    // before it dispatches an API route, so a test that calls the hook directly
+    // has to stand in for that.
+    await expect(harness.withCompanyScope(companyId, async () => await plugin.definition.onApiRequest?.({
       routeKey: "initialize",
       method: "POST",
       path: `/issues/${rootIssueId}/smoke`,
@@ -151,7 +158,7 @@ describe("orchestration smoke plugin", () => {
       },
       companyId,
       headers: {},
-    })).resolves.toMatchObject({
+    }))).resolves.toMatchObject({
       status: 201,
       body: expect.objectContaining({
         rootIssueId,
--- a/packages/plugins/plugin-llm-wiki/tests/plugin.spec.ts
+++ b/packages/plugins/plugin-llm-wiki/tests/plugin.spec.ts
@@ -1699,10 +1699,10 @@ Duplicate headings receive stable suffixes.
     });
 
     expect(writes).toHaveLength(0);
-    const operations = await harness.ctx.issues.list({
+    const operations = await harness.withCompanyScope(COMPANY_ID, () => harness.ctx.issues.list({
       companyId: COMPANY_ID,
       originKindPrefix: String(OPERATION_ORIGIN_KIND),
-    });
+    }));
     expect(operations).toHaveLength(0);
     expect(harness.dbExecutes.some((execute) => execute.sql.includes("wiki_sources"))).toBe(false);
     expect(harness.dbExecutes.some((execute) => execute.sql.includes("wiki_operations"))).toBe(false);
@@ -1979,13 +1979,13 @@ Duplicate headings receive stable suffixes.
     });
 
     await plugin.definition.setup(harness.ctx);
-    await harness.ctx.issues.documents.upsert({
+    await harness.withCompanyScope(COMPANY_ID, () => harness.ctx.issues.documents.upsert({
       companyId: COMPANY_ID,
       issueId: child.id,
       key: "plan",
       title: "Plan",
       body: "Document evidence for the source bundle.",
-    });
+    }));
 
     const first = await harness.performAction<{
       markdown: string;
@@ -2045,13 +2045,13 @@ Duplicate headings receive stable suffixes.
     });
 
     await plugin.definition.setup(harness.ctx);
-    await harness.ctx.issues.documents.upsert({
+    await harness.withCompanyScope(COMPANY_ID, () => harness.ctx.issues.documents.upsert({
       companyId: COMPANY_ID,
       issueId: issue.id,
       key: "plan",
       title: "Plan",
       body: "OPENAI_API_KEY=sk-supersecretdocumentvalue1234567890",
-    });
+    }));
 
     const run = await harness.performAction<{
       bundle: {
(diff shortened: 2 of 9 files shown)


ericdfields:
Thank you for the review. The finding is correct, and it is fixed in `3d6b19d95`.

I confirmed the claim before I changed the code. Both dispatch paths call `runJob` with `{ job }` and no company:

- `server/src/services/plugin-job-scheduler.ts` line 370 (scheduled dispatch)
- `server/src/services/plugin-job-scheduler.ts` line 515 (manual trigger)

`RunJobParams` is `{ job: PluginJobContext }`, and `deriveInvocationScope` has no `runJob` branch, so no scope is made for a job today.

So the harness gave a company-scoped job execution mode that the host cannot give. That is the same fault this pull request removes, pointed the other way: the fake permitted more than the host.

The fix does not soften the model. A job run now carries no company, and `runJob` throws if a test gives it one. The error names the reason and points to `harness.withCompanyScope(companyId, ...)` for a test that wants the scoped path. The two job tests that encoded the old model now declare the scope inside the handler, and a new test asserts the throw. When the host does dispatch jobs per company, `runJob` can read the company off the job context and the throw goes away.

ericdfields:
Drift check, four days on: this branch is now 98 commits behind `master`, and unlike the last check that gap does touch the files this PR reads. I verified the merge locally rather than assume `CLEAN` covered it.

`CLEAN` only rules out *textual* conflict. The real exposure here is semantic: `testing-harness-scope-parity.test.ts` derives its gate set by reading `packages/plugins/sdk/src/worker-rpc-host.ts` at runtime, so a company-scoped method added upstream would let this merge green and then turn the guard red on `master`.

Since the merge-base (`43ab441f0`), `master` has changed both files the guard reads:

- `worker-rpc-host.ts` +74 — a new `duplexChannel` namespace and four `duplexChannel*` request handlers.
- `testing.ts` +8 — the matching `duplexChannel` no-op on the harness.

Neither enters the scan's gate set. The new host-side sends are `notifyHost(DUPLEX_CHANNEL_DATA_NOTIFICATION, …)` — a constant, not a string literal, so the `notifyHost\(\s*"` scan does not match them; the four `duplexChannel*` handlers are host→worker, not worker→host. So the guard's set is unchanged by this drift, which is the answer I wanted but not one I would take from reading alone.

Verified by trial merge, not inspection. Merged `master` into the PR head locally (clean, no conflicts) and ran the SDK suite against the merged tree:

- `testing-invocation-scope.test.ts` + `testing-harness-scope-parity.test.ts` — 19/19 pass.
- Full `packages/plugins/sdk` suite — 81/81 pass across 9 files, including `master`'s new `worker-rpc-host.test.ts` and `protocol.duplex-channel.test.ts`.

One caveat on that run, so the numbers are not read as more than they are: `master` bumps `packages/shared` from `zod@^3.24.2` to `^4.4.3`, and my worktree resolves a hoisted `zod@3.25.76`. On the first attempt `testing-invocation-scope.test.ts` failed at import with `z.string(...).guid is not a function` — a stale-install artifact of the local merge, not a behavior change. I re-ran with `zod@4.4.3` aliased in; the numbers above are from that run. CI installs from the lockfile and would not see it.

No rebase, deliberately. Nothing is broken by the gap, and a force-push would reset the 5/5 review for no gain. Happy to rebase on request if you would rather review against current `master`.

Status unchanged otherwise: 20/20 checks pass, `MERGEABLE` / `CLEAN`, head `3d6b19d95`.

ericdfields:
Maintainer review requested. This remains open and mergeable with all checks green; the harness/production scope parity change is unchanged since the verified drift check above. I am happy to rebase or adjust anything you want reviewed against current `master`.

PR #11583, opened 2026-08-18
fix: strip systemd service-manager env vars from spawned agent processes

dzianisv:
## Problem

Agent child processes spawned by `runChildProcess` inherit the paperclip
server's systemd readiness socket (`NOTIFY_SOCKET`) and socket-activation
vars (`LISTEN_FDS`/`LISTEN_PID`/`LISTEN_FDNAMES`), so an agent subprocess can
reach the service's own readiness channel. Every boot then logs `Got
notification message from PID N, but reception only permitted for main PID M`.

## Fix

In `runChildProcess` (`packages/adapter-utils/src/server-utils.ts`), in the
same block that already strips the Claude Code nesting-guard vars, strip
`NOTIFY_SOCKET` / `LISTEN_FDS` / `LISTEN_PID` / `LISTEN_FDNAMES` from the
**merged** env (`process.env` + `opts.env`) after the merge — not just from
the `process.env` side — since every local adapter builds `opts.env` as
`{ ...process.env, ...env }` and would otherwise re-leak these vars even
after `sanitizeInheritedPaperclipEnv` strips `PAPERCLIP_*` keys from the
`process.env` copy.

This is a single chokepoint fix covering all local adapters (pi-local,
claude-local, codex-local, cursor-local, gemini-local, opencode-local,
copilot-local, acpx-local). Remote/sandbox execution targets bypass
`runChildProcess` and are unaffected.

Scope: env-strip only, per the linked issue. Does not include the
systemd-cgroup escape work tracked separately.

## Testing

New unit test in `packages/adapter-utils/src/server-utils.test.ts`:
sets `NOTIFY_SOCKET`/`LISTEN_FDS`/`LISTEN_PID`/`LISTEN_FDNAMES` in **both**
`process.env` and the caller's `opts.env`, spawns a child that echoes its
own `process.env` as JSON, and asserts none of the four vars are present.

Confirmed the test fails on unmodified master (red) and passes with the
fix applied (green):

```
$ npx vitest run --project @paperclipai/adapter-utils src/server-utils.test.ts
 ✓ |@paperclipai/adapter-utils| src/server-utils.test.ts (19 tests)
```

(No per-package `test` script exists in `packages/adapter-utils/package.json`
or any sibling adapter package — tests run via the root vitest `projects`
config, so `--project @paperclipai/adapter-utils` is the equivalent of
`pnpm --filter @paperclipai/adapter-utils test`.)

`pnpm --filter @paperclipai/adapter-utils typecheck` is clean.

Paperclip: AGE-551

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -23,6 +23,7 @@ COPY packages/db/package.json packages/db/
 COPY packages/adapter-utils/package.json packages/adapter-utils/
 COPY packages/mcp-server/package.json packages/mcp-server/
 COPY packages/adapters/claude-local/package.json packages/adapters/claude-local/
+COPY packages/adapters/copilot-local/package.json packages/adapters/copilot-local/
 COPY packages/adapters/codex-local/package.json packages/adapters/codex-local/
 COPY packages/adapters/cursor-local/package.json packages/adapters/cursor-local/
 COPY packages/adapters/gemini-local/package.json packages/adapters/gemini-local/
--- a/cli/package.json
+++ b/cli/package.json
@@ -38,6 +38,7 @@
   "dependencies": {
     "@clack/prompts": "^0.10.0",
     "@paperclipai/adapter-claude-local": "workspace:*",
+    "@paperclipai/adapter-copilot-local": "workspace:*",
     "@paperclipai/adapter-codex-local": "workspace:*",
     "@paperclipai/adapter-cursor-local": "workspace:*",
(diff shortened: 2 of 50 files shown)


dzianisv:
Superseded by a clean PR — my fork's master had diverged heavily from upstream/master, which dragged unrelated files into this diff. Closing and reopening from a branch based directly on upstream/master.

PR #11595, opened 2026-08-18
fix(server,skill): expose ZodPipeline schemas in OpenAPI, document monitor.kind enum (AGE-594)

dzianisv:
## Problem

`PATCH /api/issues/{id}` accepts exactly one value for `executionPolicy.monitor.kind` (`external_service`), but the OpenAPI spec resolved both `comment` and (per the live report) `executionPolicy` to an empty `{}` schema — hiding the accepted shape from the one sanctioned discovery path. When an agent guesses a wrong `kind`, the whole `PATCH` — including any `comment` in the same body — is rejected atomically and lost.

## Root cause

The hand-rolled `zodToOpenApiSchema`/`unwrapSchema` pair in `server/src/routes/openapi.ts` had no case for `ZodPipeline` (produced by `.pipe()`, e.g. `multilineTextSchema.pipe(z.string().min(1))`, which is how `comment` and several other fields are defined). It fell through to the generator's default `return {}`.

`executionPolicy.monitor.kind` itself already resolved correctly to `enum: ["external_service"]` on current `master` — I could not reproduce the reporter's `{}` for that specific field against this branch's generator, but I added a regression test for it anyway so a future refactor can't silently reintroduce it.

## Fix

- `unwrapSchema`/`isOptionalSchema`: unwrap `ZodPipeline` via its `out` side (the effective, client-facing validated shape). This fixes not just `comment` but every other `.pipe()`-based field across the whole spec — empty leaf schemas in the generated document went from 29 to 16, and I manually confirmed the remaining 16 are legitimate `z.any()`/`z.unknown()` fields (e.g. `evidence`, `wakePolicy`), not further instances of the bug.
- `skills/paperclip/SKILL.md` "Monitors and Watchers": states `kind`'s only accepted value (`external_service`) inline, and adds the caller-side recovery rule: a `PATCH` carrying `comment` that returns non-2xx did **not** persist the comment (rejection is atomic, nothing partially applies) — re-post it via `POST /api/issues/{id}/comments` before treating the write as failed.

## Tests

- `server/src/__tests__/openapi-routes.test.ts`: `comment` resolves to a typed string schema (`{ type: "string", minLength: 1 }`); `executionPolicy.monitor.kind` resolves to `{ type: "string", enum: ["external_service"] }`.
- `server/src/__tests__/issue-execution-policy-routes.test.ts`: `PATCH` with an invalid `monitor.kind` returns `400` with `options: ["external_service"]` in `details`, and never reaches `issueService.update` — the only path that would persist the comment — proving the rejection is atomic and no partial write happens.

```
$ vitest run server/src/__tests__/openapi-routes.test.ts server/src/__tests__/issue-execution-policy-routes.test.ts
 Test Files  2 passed (2)
      Tests  25 passed (25)
```

## Not done / out of scope

- Whether `serviceName`, `externalRef`, `recoveryPolicy`, or `scheduledBy` carry other undocumented enums — spot-checked via the same generator and found none beyond what's already covered by their zod `z.enum(...)` definitions (which do surface correctly).
- Whether any other route drops a side-effect field on atomic validation failure — not investigated; `PATCH /api/issues/{id}` was the only route in scope for this issue.

Closes AGE-594.


dzianisv:
Closing: re-opening under a branch/description without internal ticket references per CONTRIBUTING.md.

PR #11596, opened 2026-08-18
fix(server): expose ZodPipeline schemas in the OpenAPI generator

dzianisv:
## Thinking Path

> - Every route's request body should be discoverable through the generated OpenAPI document, so an agent client can learn the accepted shape of a field without guessing.
> - `server/src/routes/openapi.ts` converts each route's zod schema into an OpenAPI/JSON Schema document by hand, with its own `zodToOpenApiSchema`/`unwrapSchema` walker (it does not use a third-party zod-to-openapi library).
> - That walker has no case for `ZodPipeline`, the node zod produces for `.pipe(...)`. Several fields (for example `comment` on `PATCH /api/issues/{id}`, defined as `multilineTextSchema.pipe(z.string().min(1))`) use this pattern, so the walker falls through to the generator's default `{}` for them instead of a typed schema.
> - An agent calling this API cannot see the real shape or constraints of those fields from the spec. Worse, if a client guesses wrong on an adjacent, correctly-typed enum field in the same request body, the whole write is rejected atomically, and it is hard to tell from the spec alone which parts of the body were even validated.
> - This pull request adds a `ZodPipeline` case to the walker, unwrapping to the pipeline's `out` side (the effective, client-facing validated shape), so every `.pipe()`-based field resolves to a real schema instead of `{}`.
> - The benefit is a materially more complete and trustworthy generated OpenAPI document, with no route-by-route special-casing needed — the fix is in the shared schema walker.

No public GitHub issue exists yet for this specific gap. Following the bug report template inline:

**What happened?**

`server/src/routes/openapi.ts` builds the OpenAPI document from each route's zod schema using a hand-rolled `zodToOpenApiSchema`/`unwrapSchema` pair. `unwrapSchema` handles `ZodOptional`, `ZodDefault`, `ZodCatch`, and `ZodEffects`, but not `ZodPipeline` (the node produced by zod's `.pipe(...)`). Several shared validators use `somethingSchema.pipe(otherSchema)` (for example `multilineTextSchema.pipe(z.string().min(1))`, used by `comment` and other free-text fields across the issue routes). For those fields, `zodToOpenApiSchema` falls through to its default `return {}`, so the generated OpenAPI document shows an empty, untyped schema instead of the real one.

**Expected behavior**

A field defined with `.pipe(...)` should resolve, in the generated OpenAPI document, to a typed schema reflecting its effective validated shape (type, `minLength`/`maxLength`, `enum`, etc.), the same as any other zod-defined field.

**Steps to reproduce**

1. Start the server and request `GET /api/openapi.json`.
2. Look up `paths["/api/issues/{id}"].patch.requestBody.content["application/json"].schema.properties.comment`.
3. Observe it resolves to `{}` instead of a typed string schema, even though the underlying zod definition (`multilineTextSchema.pipe(z.string().min(1))`) has a real, enforced shape.

Current `master` at the time of this change.

**Deployment mode**

Any (this is a request-schema generation bug in the server's OpenAPI document builder, independent of deployment mode).

## What Changed

- `server/src/routes/openapi.ts`: `unwrapSchema` and `isOptionalSchema` now handle `ZodPipeline` by unwrapping to `schema._def.out` — the pipeline's second stage, which carries the effective, enforced shape (e.g. `z.string().min(1)` in `multilineTextSchema.pipe(z.string().min(1))`).
- This is a single chokepoint fix in the shared schema walker. It applies to every `.pipe()`-based field across the whole generated document, not just `comment`. I audited the full generated document before and after: previously-empty leaf schemas went from 29 to 16, and I manually confirmed the remaining 16 are legitimate `z.any()`/`z.unknown()` fields (for example `evidence`, `wakePolicy`) that are correctly represented as `{}` — not further instances of this bug.
- Also confirmed, and added a regression test for, an existing enum field (`executionPolicy.monitor.kind` on the same route) that already resolved correctly to a typed `enum`, so a future refactor of the walker cannot silently regress it back to `{}`.
- `skills/paperclip/SKILL.md` "Monitors and Watchers": documents that a `PATCH` carrying a `comment` which returns a non-2xx response did not persist that comment — Paperclip's issue-update route validates the whole request body before any part of it is applied, so a rejected write applies nothing. The skill now tells the calling agent to re-post the comment via `POST /api/issues/{id}/comments` before treating the write as failed, instead of assuming data loss or a permission error.

## Verification

```
$ npx vitest run server/src/__tests__/openapi-routes.test.ts server/src/__tests__/issue-execution-policy-routes.test.ts
 Test Files  2 passed (2)
      Tests  25 passed (25)
```

- `openapi-routes.test.ts` (new assertions): `PATCH /api/issues/{id}`'s `comment` property resolves to `{ type: "string", minLength: 1 }`, and `executionPolicy.monitor.kind` resolves to `{ type: "string", enum: ["external_service"] }`.
- `issue-execution-policy-routes.test.ts` (new test): a `PATCH` with an invalid `executionPolicy.monitor.kind` and a `comment` in the same body returns `400` with `options: ["external_service"]` in `details`, and the mocked `issueService.update` (the only path that would persist the comment) is never called — proving the rejection is atomic and no partial write happens.
- Manually confirmed the fix locally by generating the full OpenAPI document before and after the change and diffing the count and identity of empty (`{}`) leaf schemas across every route (29 → 16, all remaining 16 legitimate `z.any()`/`z.unknown()` fields).
- `tsc --noEmit` on the `server` package shows no new errors introduced by the two changed files (pre-existing, unrelated errors exist elsewhere in the package from an unbuilt workspace dependency in this sandbox).

## Risks

Low risk. This changes only how a subset of request schemas are *documented* in the generated OpenAPI JSON; it does not change any runtime validation, route handler, or database behavior. The affected fields were already being validated exactly this way at runtime (via the real zod schema) — this only fixes what the spec *says* about them. No migration, no API contract change, no breaking change for existing clients.

## Model Used

Claude (Anthropic), used within a coding-agent harness with tool use (file read/edit, local test execution, local OpenAPI-document generation and diffing for verification).

## Checklist

- [ ] All Paperclip CI gates are green — not verified at the time of this edit; see `gh pr checks` for current status.
- [ ] Greptile is 5/5 with no open P2s, recommendations, or follow-ups — not verified at the time of this edit; see the Greptile review comment for current status.


dzianisv:
CI status update: the earlier "Verify serialized server suites (3/5)" failure (https://github.com/paperclipai/paperclip/actions/runs/32105287313) was an unrelated flaky timeout in `issue-dependency-wakeups-routes.test.ts` — a file this PR does not touch. Confirmed locally (3/3 passed, ran twice) and via a fresh CI run after reopening the PR: https://github.com/paperclipai/paperclip/actions/runs/32107655830 (green, all suites pass).

Re: the Greptile finding above ("awaited run-log persistence can still pause Claude stdout", files `packages/adapter-utils/src/server-utils.ts`, `packages/adapters/claude-local/src/server/execute.ts`, `server/src/services/heartbeat.ts`, `server/src/services/run-log-store.ts`) — none of those files are in this PR's diff (`gh pr diff 11596 --name-only` returns only `server/src/routes/openapi.ts`, its two test files, and `skills/paperclip/SKILL.md`). That finding is stale, carried over from before this branch was rebased onto a clean base (it used to include unrelated commits from #11585). Requested a re-review; this PR is scoped to the AGE-594 fix only and is ready for human approval/merge — branch protection is currently BLOCKED only on the missing approving review, not on checks.

PR #11606, opened 2026-08-18
fix(heartbeat): classify pi_local and other piped-stdio local adapters as server-stdio bound for hot restart

dzianisv:
## Problem

`isServerStdioBoundHotRestartRun` decides which `running` heartbeat runs
block hot-restart adoption. It hard-codes an allowlist of
`[claude_local, codex_local, gemini_local]` for its fallback check, and
defaults every other adapter type to `false` (adoptable) whenever
`contextSnapshot` carries no explicit `processTopology`/`executionEngine`
hint. No adapter anywhere in the tree ever writes
`contextSnapshot.processTopology`, so every run falls through to that
default.

In reality, every local adapter (`pi_local`, `opencode_local`,
`grok_local`, `cursor`, `hermes_local`, ...) is spawned via
`runAdapterExecutionTargetProcess` -> `runChildProcess`, which always uses
piped stdout/stderr owned by the server process
(`stdio: [..., "pipe", "pipe"]`), with no cli-vs-acp engine split like
claude/codex/gemini have. When the server exits, those pipe peers close
and the child dies on its next write attempt — it cannot survive a hot
restart no matter what the classifier says. `pi_local` in particular is
the adapter this deployment actually runs, and every one of its
`running` runs was being marked adoptable, never actually adopted, and
written off as `process_lost` on every hot restart instead of taking the
graceful-drain-and-queue-retry path.

## Fix

Add `ALWAYS_SERVER_STDIO_BOUND_LOCAL_ADAPTER_TYPES` (`cursor`,
`grok_local`, `hermes_local`, `opencode_local`, `pi_local`, plus a
defensive `cursor_local` alias) and consult it for any `adapterType` not
in the `claude_local`/`codex_local`/`gemini_local` engine-selectable
trio. The explicit `processTopology: "server_stdio"` / `"detached"`
overrides at the top of the function are unchanged, so the escape hatch
for future explicit spawn-time signalling (AGE-629 follow-on) stays
open — this PR does not attempt to have the spawn path write that field
itself, since the acceptance criteria for this fix require the
classifier to be correct even with an empty/absent context.

## Tests

New cases in `server/src/__tests__/heartbeat-process-recovery.test.ts`:

- `pi_local`/`opencode_local`/`grok_local`/`cursor`/`cursor_local` with an
  empty `contextSnapshot` now classify as server-stdio bound (verified
  end-to-end through `prepareHotRestartShutdown` ->
  `drainRunningRunsForShutdown` -> `reconcileHotRestartAdoption` and the
  `hot-restart-report.json` shape), with the `pi_local` case additionally
  asserting the run's terminal `errorCode` becomes
  `server_shutdown_interrupted` (retry queued) and a `queued` retry run
  is written — not `process_lost`.
- A `pi_local` run with an explicit `processTopology: "detached"`
  override still adopts (`mode: "hot_restart"`), confirming the escape
  hatch is untouched.
- Confirmed all new non-override cases fail against pre-fix
  `heartbeat.ts` (`git stash` the fix, rerun — 6/6 fail with
  `mode: "hot_restart"` instead of `acp_drain_required"`).

Verified locally:

```
npx vitest run server/src/__tests__/heartbeat-process-recovery.test.ts
# 110 tests, all pass in isolation (one unrelated timing test
# occasionally times out under parallel load in the full run; passes
# standalone and is unrelated to this change — a telemetry-tracking
# test, not the classifier).
```

`tsc --noEmit -p server` shows the same 139 pre-existing errors before
and after this change (all in unrelated plugin-host files missing a
built `@paperclipai/plugin-sdk`), confirming zero new type errors from
this diff.

Refs: AGE-655 (split out of AGE-538), AGE-629, AGE-550, AGE-611 (PR
tracker for open upstream PRs).

### DIFF
--- a/server/src/__tests__/heartbeat-process-recovery.test.ts
+++ b/server/src/__tests__/heartbeat-process-recovery.test.ts
@@ -1790,6 +1790,176 @@ describeEmbeddedPostgres("heartbeat orphaned process recovery", () => {
     });
   });
 
+  // AGE-655: isServerStdioBoundHotRestartRun previously only recognized
+  // claude_local/codex_local/gemini_local as potentially server-stdio bound.
+  // Every other local adapter (pi_local, opencode_local, grok_local, cursor)
+  // reaches the same piped `runChildProcess` spawn with no "cli" engine
+  // escape hatch, so an empty/absent contextSnapshot.processTopology must
+  // still classify them as server-stdio bound -- not silently adoptable and
+  // then lost as process_lost when adoption fails in reality.
+  it.each([
+    ["pi_local"],
+    ["opencode_local"],
+    ["grok_local"],
+    ["cursor"],
+    ["cursor_local"],
+  ])(
+    "treats a %s run with no explicit processTopology as server-stdio bound and drains it instead of adopting",
+    async (adapterType) => {
+      const child = spawnAliveProcess();
+      childProcesses.add(child);
+      expect(child.pid).toBeGreaterThan(0);
+      const { runId } = await seedRunFixture({
+        adapterType,
+        agentStatus: "running",
+        processPid: child.pid ?? null,
+        processGroupId: null,
+        contextSnapshot: {},
+      });
+
+      await withTempPaperclipHome(async () => {
+        await writeHotRestartIntent({
+          previousServerPid: process.pid,
+          previousServerVersion: "old-version",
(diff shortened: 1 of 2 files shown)


dzianisv:
Superseded by a PR opened from a properly-named branch (this one's branch name embedded an internal ticket id, against CONTRIBUTING.md). Closing in favor of the replacement.

PR #11612, opened 2026-08-18
fix(cli): honor DATABASE_URL in doctor database check

### LINKED ISSUE #11598: Doctor blocks `paperclipai run` when Postgres is configured only via instance `.env` DATABASE_URL
## Summary

`paperclipai run` (including `paperclipai.service`) aborts before starting the server because doctor treats Postgres as unconfigured whenever `config.json` has `"database": { "mode": "postgres" }` without `connectionString`, even if `DATABASE_URL` is set in the instance `.env` (and loaded into the process environment).

That contradicts the runtime resolver, which already accepts `DATABASE_URL` / the instance env file.

## Why it failed in practice

On a source checkout + systemd user unit (`paperclipai run --instance default`):

1. Instance config used `"database": { "mode": "postgres" }` and kept the password only in `~/.paperclip/instances/default/.env` as `DATABASE_URL=postgres://...` (recommended so secrets are not duplicated into `config.json`).
2. `run` loaded the env file, then ran doctor.
3. Doctor reported:
   - `✓ Agent JWT secret: PAPERCLIP_AGENT_JWT_SECRET is set in environment`
   - `✗ Database: PostgreSQL mode selected but no connection string configured`
4. `run` printed `Doctor found blocking issues. Not starting server.` and exited 1.
5. systemd `Restart=always` then crash-looped the unit.

The server never reached listen/migrations. A public reverse proxy in front of it therefore saw origin 502 until the connection string was copied into `config.json`.

## Root cause

`cli/src/checks/database-check.ts` only inspects `config.database.connectionString`:

```ts
if (config.database.mode === "postgres") {
  if (!config.database.connectionString) {
    return {
      name: "Database",
      status: "fail",
      message: "PostgreSQL mode selected but no connection string configured",
      ...
    };
  }
}
```

The same process already has `DATABASE_URL` (doctor even sees other env secrets). Runtime resolution in `packages/db/src/runtime-config.ts` (`resolveDatabaseTarget`) prefers:

1. `process.env.DATABASE_URL`
2. instance `.env` `DATABASE_URL`
3. `config.database.connectionString`

Doctor does not follow that order, so a valid authenticated/public + external Postgres setup is classified as a blocking failure.

Related: the unit rendered by `cli/src/services/service-manager.ts` (`renderSystemdUnit`) does not set `EnvironmentFile=` for the instance `.env`. That is not required for doctor if the CLI loads the file itself, but it is inconsistent with how operators expect systemd to inject `DATABASE_URL` / secrets.

## Expected

Doctor should resolve the DB the same way as runtime:

- pass when `DATABASE_URL` is a postgres URL in the environment or instance `.env`
- then actually `SELECT 1` with that URL
- fail only if neither env nor `config.database.connectionString` is present

`paperclipai run` / `paperclipai.service` should start in that configuration without copying the password into `config.json`.

## Extra failure we hit on the same start path

A source-checkout wrapper at `~/.local/bin/paperclipai` made doctor treat the install as managed:

`Managed install artifacts exist but ~/.paperclip/cli/install.json is missing`

`cli/src/checks/managed-install-check.ts` (`hasManagedArtifacts`) treats any file at the default shim path as a managed-install artifact. A local wrapper for `tsx cli/src/index.ts` is enough to trip that check and also block `run`.

## Environment

- Paperclip from git (`paperclipai/paperclip`), CLI via source checkout
- `server.deploymentMode=authenticated`, `exposure=public`
- External PostgreSQL (`database.mode=postgres`)
- systemd user service from `paperclipai service install`

## Workaround

Put `database.connectionString` in `config.json` (duplicates the secret) and/or do not place a custom binary at `~/.local/bin/paperclipai`.

saime428:
## Thinking Path

> - Paperclip is the open source control plane for AI agent companies.
> - The CLI runs doctor checks before `paperclipai run` starts the server.
> - The CLI loads the instance `.env` file before it runs doctor.
> - The database runtime uses `DATABASE_URL` before `config.database.connectionString`.
> - The doctor database check only used the config value and blocked a valid Postgres setup.
> - This pull request makes the doctor check use the same environment-first priority and keeps the existing connection query.
> - The benefit is that operators can keep the Postgres credential in the instance `.env` file without blocking server startup.

Fixes: #11598

## What Changed

- Resolve the doctor Postgres connection string from `DATABASE_URL` first.
- Fall back to `config.database.connectionString` when the environment value is empty.
- Use the resolved value for the existing `SELECT 1` connection check.
- Add a regression test for an environment-only connection string.

## Verification

- `corepack pnpm exec vitest run cli/src/__tests__/database-check.test.ts`
- `corepack pnpm --filter paperclipai typecheck`

## Risks

- Low risk. The change only affects the doctor check for Postgres mode.
- The existing config fallback remains unchanged.
- Embedded PostgreSQL behavior remains unchanged.

## Model Used

OpenAI `gpt-5.6-sol`, with repository tool use.

## Checklist

- [x] I have updated relevant documentation to reflect my changes (no update was needed because this aligns doctor with existing runtime behavior)

### DIFF
--- a/cli/src/__tests__/database-check.test.ts
+++ b/cli/src/__tests__/database-check.test.ts
@@ -1,17 +1,32 @@
 import fs from "node:fs";
 import os from "node:os";
 import path from "node:path";
+import { createDb } from "@paperclipai/db";
 import { afterEach, describe, expect, it, vi } from "vitest";
 import { databaseCheck } from "../checks/database-check.js";
 import type { PaperclipConfig } from "../config/schema.js";
 
+vi.mock("@paperclipai/db", () => ({
+  createDb: vi.fn(() => ({ execute: vi.fn().mockResolvedValue(undefined) })),
+}));
+
 const created: string[] = [];
+const ORIGINAL_DATABASE_URL = process.env.DATABASE_URL;
 const ORIGINAL_IN_WORKTREE = process.env.PAPERCLIP_IN_WORKTREE;
 
 function makeBase(): string {
   const base = fs.mkdtempSync(path.join(os.tmpdir(), "paperclip-dbcheck-"));
   created.push(base);
   return base;
+}
+
+function postgresConfig(connectionString?: string): PaperclipConfig {
+  return {
+    database: {
+      mode: "postgres",
+      connectionString,
+    },
+  } as unknown as PaperclipConfig;
 }
 
 function embeddedConfig(dataDir: string): PaperclipConfig {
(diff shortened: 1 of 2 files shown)


saime428:
Closing this to keep the review queue clear rather than leave it sitting open. Nothing is blocked on anyone here.

The branch stays up, so this can be reopened if it turns out to be useful. I have not kept it rebased, so it would need a refresh first.

PR #11620, opened 2026-08-18
fix(cli): fix npm view --json parsing and missing service PATH so install + service install work

tnghia903:
## Thinking Path

> - The CLI has a managed install path (`paperclipai install`) that downloads and pins a specific published version outside of npx
> - This path resolves the target version by shelling out to `npm view <pkg>@<spec> version --json`
> - Newer npm releases (10+) changed this output. They wrap the resolved version in a single-element JSON array instead of returning a bare JSON string
> - The CLI only parsed the old bare-string shape, so on newer npm every `paperclipai install` call failed, even with `--version <exact>`
> - This pull request updates the parser to accept the array-wrapped shape too, while still rejecting ambiguous multi-element output
> - The benefit is `paperclipai install` works again on current npm versions, so users can set up the background service (`paperclipai service install`), which requires the managed shim this command creates
> - While verifying that fix end-to-end, `paperclipai service install` still failed to actually start on macOS: the generated launchd/systemd service definitions never set PATH, and launchd's own default PATH does not include wherever `node` actually lives (nvm, fnm, Homebrew, asdf, etc), so the managed shim's `#!/usr/bin/env node` shebang could not resolve `node`
> - This pull request also updates `renderLaunchdPlist()`/`renderSystemdUnit()` to snapshot the installer's own PATH into the generated service definition, so the background service can actually start

No existing GitHub issue covers this; searched open/closed PRs and issues for "npm view --json", "parseResolvedVersion", "npm returned an unexpected version response", and "paperclipai install" and found no duplicates. Describing inline using the Bug report template fields:

**What happened?**
Running `paperclipai install` (with no flags, or with `--version <exact>`) fails every time on npm 10+ with:
```
npm returned an unexpected version response: [
  "2026.817.0"
]
```

**Expected behavior**
`paperclipai install` resolves the requested version and proceeds with the managed install, as it does on older npm releases.

**Steps to reproduce**
1. Use npm 10 or newer (verified on npm 12.0.2).
2. Run `npx paperclipai install`.
3. Observe the failure above instead of a successful install.
4. Confirm independently: `npm view paperclipai@latest version --json` prints `["2026.817.0"]`, not `"2026.817.0"`.

**Paperclip version or commit:** master @ 94f12f322a4b1428e0845896f178f5eac2e254d8 (this branch)
**Deployment mode:** Local dev (pnpm dev) / CLI usage
**Installation method:** npm / pnpm global install
**Node.js version:** v22.23.1
**Operating system:** macOS 26.6.1 (Apple Silicon)

Second, related bug found while verifying the fix above:

**What happened?**
After working around the first bug (`npm install -g paperclipai@latest`, which places the binary at the same path the managed shim would use), `paperclipai service install` reported success, but the launchd job never actually ran:
```
$ launchctl print gui/501/ing.paperclip.paperclipai
  last exit code = 78: EX_CONFIG
$ cat ~/.paperclip/instances/default/logs/service.err.log
env: node: No such file or directory
```

**Expected behavior**
The installed background service starts and serves Paperclip, the same as running `paperclipai run` manually.

**Steps to reproduce**
1. Have `node` resolve from a location outside `/usr/bin:/bin:/usr/sbin:/sbin` (true for nvm, fnm, Homebrew, asdf, and most non-system Node installs — including this reporter's `~/.hermes/node`, exposed on PATH via a `~/.local/bin/node` symlink).
2. Run `paperclipai service install` on macOS.
3. Observe `service status` reports `installed: true` but `active: false`, `pid: null`.
4. `launchctl print` on the job shows `last exit code = 78: EX_CONFIG`; the service's `.err.log` shows `env: node: No such file or directory`.
5. Root cause: `renderLaunchdPlist()` in `cli/src/services/service-manager.ts` never set a `PATH` entry in `EnvironmentVariables`, and launchd's own default PATH for GUI agents is `/usr/bin:/bin:/usr/sbin:/sbin`, which does not include most Node install locations. The managed shim is a `#!/usr/bin/env node` script, so `env` cannot find `node` to exec it. `renderSystemdUnit()` had the same gap for systemd `--user` services.

**Paperclip version or commit:** master @ 94f12f322a4b1428e0845896f178f5eac2e254d8 (this branch)
**Deployment mode:** Local dev (pnpm dev) / CLI usage — specifically `paperclipai service install`
**Installation method:** npm / pnpm global install
**Node.js version:** v22.23.1
**Operating system:** macOS 26.6.1 (Apple Silicon)

## What Changed

- `cli/src/commands/install.ts`: `parseResolvedVersion()` now also accepts a single-element JSON array of strings (the shape newer npm emits for `npm view --json`), in addition to the existing bare-string shape. Arrays with zero or more than one element still fail loudly instead of guessing.
- `cli/src/services/service-manager.ts`: added `resolveServicePath()`, which snapshots `process.env.PATH` (overridable via `PAPERCLIP_SERVICE_PATH`, with a hardcoded fallback). `renderLaunchdPlist()` now writes it as an `EnvironmentVariables` `PATH` entry, and `renderSystemdUnit()` writes it as an `Environment="PATH=..."` line. Both `LaunchdServiceManager` and `SystemdServiceManager` pass their constructor-time PATH into `renderDefinition()`.
- `cli/src/__tests__/install-command.test.ts`: added a test covering the bare-string case, the array-wrapped case, and the still-rejected multi-element case, exercised through the existing `resolvePublishedVersion()` export with a mocked command runner.
- `cli/src/__tests__/service-manager.test.ts`: added tests asserting both `renderLaunchdPlist()` and `renderSystemdUnit()` emit the given (or `process.env.PATH`-derived) PATH.

## Verification

- `node_modules/.bin/vitest run cli/src/__tests__/install-command.test.ts` — 17 passed, 0 failed.
- `node_modules/.bin/vitest run cli/src/__tests__/service-manager.test.ts` — 21 passed, 0 failed.
- Reproduced the first bug locally before the fix by running `npm view paperclipai@latest version --json --registry=https://registry.npmjs.org` on npm 12.0.2, which prints the array-wrapped form; confirmed `parseResolvedVersion()` now returns `"2026.817.0"` for that exact input instead of throwing.
- Reproduced and confirmed the fix for the second bug end-to-end on a real macOS launchd job (not just unit tests): hand-patched an installed `~/Library/LaunchAgents/ing.paperclip.paperclipai.plist` to add the `PATH` this change now generates automatically, reloaded it with `launchctl bootout` + `launchctl bootstrap`, and watched the service progress past `env: node: No such file or directory` into the CLI's own startup/doctor checks (`✓ Node.js runtime: Node.js 22.23.1`), where it previously failed immediately with `EX_CONFIG`.

## Risks

Low risk for both changes.
- The version-parsing change only widens what `parseResolvedVersion()` accepts (bare string, or a single-element array of one string); it still throws on anything else, including empty arrays and multi-element arrays, so it cannot silently pick a wrong version. It does not change the npm command being run, the CLI's public flags, or any other install code path.
- The PATH change only adds a new `PATH` entry to generated service definitions; it does not change `ProgramArguments`/`ExecStart`, other environment variables, or any existing escaping/validation. Existing installs are unaffected until the user re-runs `service install`, which already regenerates the definition file on any change (`writeIfChanged`).

## Model Used

Claude Sonnet 5 (`claude-sonnet-5`), via Claude Code CLI. Standard (non-extended) reasoning; used file reads, shell execution (to reproduce the npm behavior locally and run the test suite), and file edits/tool use. No extended thinking mode.

## Checklist

### DIFF
--- a/cli/src/__tests__/install-command.test.ts
+++ b/cli/src/__tests__/install-command.test.ts
@@ -10,6 +10,7 @@ import {
   resolveGitInstallRequest,
   resolveGitInstallWorkspacePackages,
   resolveNpmInstallRequest,
+  resolvePublishedVersion,
   runCommandWithDiagnostics,
 } from "../commands/install.js";
 import { uninstallCommand } from "../commands/uninstall.js";
@@ -60,6 +61,28 @@ describe("managed install commands", () => {
     });
     expect(() => resolveNpmInstallRequest({ canary: true, version: "1.2.3" })).toThrow();
     expect(() => resolveNpmInstallRequest({ version: "latest" })).toThrow();
+  });
+
+  it("parses published versions from both bare-string and array-wrapped npm view --json output", async () => {
+    const bareString = vi.fn(async (_file: string, _args: string[]) => ({
+      stdout: JSON.stringify("2026.720.0"),
+      stderr: "",
+    }));
+    await expect(resolvePublishedVersion("latest", bareString)).resolves.toBe("2026.720.0");
+
+    const arrayWrapped = vi.fn(async (_file: string, _args: string[]) => ({
+      stdout: JSON.stringify(["2026.817.0"]),
+      stderr: "",
+    }));
+    await expect(resolvePublishedVersion("latest", arrayWrapped)).resolves.toBe("2026.817.0");
+
+    const unexpected = vi.fn(async (_file: string, _args: string[]) => ({
+      stdout: JSON.stringify(["2026.817.0", "2026.816.0"]),
+      stderr: "",
+    }));
+    await expect(resolvePublishedVersion("latest", unexpected)).rejects.toThrow(
+      "npm returned an unexpected version response",
+    );
   });
 
   it("resolves branch, tag, full SHA, and short SHA refs through GitHub", async () => {
--- a/cli/src/__tests__/service-manager.test.ts
+++ b/cli/src/__tests__/service-manager.test.ts
@@ -37,6 +37,25 @@ describe("service definition generation", () => {
     expect(unit).not.toContain("API_KEY");
   });
 
+  it("passes the installer's own PATH through to the systemd unit so `env node` can resolve it", () => {
+    const withExplicitPath = renderSystemdUnit({
+      instanceId: "team-a",
+      shimPath: "/home/alice/.local/bin/paperclipai",
+      homeDir: "/home/alice/.paperclip",
+      path: "/home/alice/.nvm/versions/node/v22/bin:/usr/bin:/bin",
+    });
+    expect(withExplicitPath).toContain('Environment="PATH=/home/alice/.nvm/versions/node/v22/bin:/usr/bin:/bin"');
+
+    const previousPath = process.env.PATH;
+    process.env.PATH = "/home/alice/.fnm/bin:/usr/bin:/bin";
+    try {
+      const withDefaultPath = renderSystemdUnit({ instanceId: "team-a", shimPath: "/home/alice/.local/bin/paperclipai", homeDir: "/home/alice/.paperclip" });
+      expect(withDefaultPath).toContain('Environment="PATH=/home/alice/.fnm/bin:/usr/bin:/bin"');
+    } finally {
+      process.env.PATH = previousPath;
+    }
+  });
+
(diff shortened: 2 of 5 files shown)


tnghia903:
@greptile review

PR #11623, opened 2026-08-18
fix(recovery): stop terminal-run recovery from stranding, flapping, and destroying live monitors

dzianisv:
## Thinking Path

> - The heartbeat recovery service watches assigned issues and escalates them to `blocked` when an agent run stops responding
> - That escalation service has three gaps: (1) a `blocked` issue with an orphaned recovery action has no monitor and no first-class blocker, so it can never wake again; (2) escalation re-derives its "already failed" verdict from a stale run even after a human already fixed the issue, so a manual unblock gets undone within minutes; (3) escalation clears any scheduled issue monitor unconditionally, even when the run that scheduled it merely crashed afterward
> - Left unaddressed, agents and operators lose real recovery work: fixes get silently reverted, and monitors that would have read back a completed external check (for example a CI run) never fire
> - This pull request adds a periodic re-wake path for orphaned blocked recovery actions, gates the "already failed" escalation checks on whether the recovery action was already resolved since the run finished, and narrows the monitor-clearing behavior so a scheduled monitor survives an unsuccessful terminal run
> - The benefit is that a `blocked` issue always has a real wake path, a manual fix is not undone by the next reconciliation tick, and a scheduled monitor is never silently destroyed

**What happened?**
`server/src/services/recovery/service.ts`'s `reconcileStrandedAssignedIssues` sweep has three related defects around its automatic `blocked` escalation path:
1. When an issue is escalated to `blocked` and its recovery-action owner's own run also terminates badly, the issue has zero first-class `blockedByIssueIds` (recovery actions are rows in `issue_recovery_actions`, not issues, so they can never populate that field) and monitors are not eligible on a `blocked` status. The issue becomes permanently unwakeable outside a human reading the board.
2. `didAutomaticRecoveryFail` classifies the *latest* run as an already-failed automatic recovery attempt purely from that run's own `retryReason`/status. It never checks whether the recovery action has since been resolved (for example by a manual `blocked` -> `todo` reset). The very next sweep tick re-derives the same verdict from the same stale run and re-blocks the issue within minutes of a correct manual unblock.
3. The sweep's only durable-wait-path exemption (skip escalation when a monitor/blocker is already scheduled) covered a *succeeded* run only. A run that crashed or timed out *after* it had already scheduled a real `executionState.monitor` was never exempted, so escalation cleared the monitor anyway.

**Expected behavior**
- A `blocked` issue caused by recovery always has either a first-class blocker, a re-armed wake, or a bounded, clearly-labeled operator escalation comment — never a silent permanent stall.
- A manually-resolved recovery action is not immediately re-escalated using the same stale run evidence; the issue gets one fresh recovery-retry dispatch first.
- A scheduled issue monitor survives an unsuccessful terminal run and is never cleared by escalation.

**Steps to reproduce**
1. Assign an issue to an agent and let its run terminate unsuccessfully (crash/timeout) so the sweep escalates it to `blocked` with a source-scoped recovery action.
2. Have that recovery-action owner's own run also terminate unsuccessfully -> the issue stays `blocked` forever with no blocker relation and no monitor.
3. Separately, manually reset a `blocked` issue back to `todo`/`in_progress` (a plain status PATCH, not the dedicated recovery-actions resolve endpoint) while its last run is still an unsuccessful "already-retried" run -> the next `reconcileStrandedAssignedIssues` tick re-escalates it to `blocked` within minutes, using the same stale run.
4. Separately, have an in-progress run schedule an `executionState.monitor` (for example to poll a CI run) and then crash/time out -> the next sweep tick escalates to `blocked` and silently clears the monitor.

## What Changed

- Added `reconcileOrphanedBlockedRecoveryActions`, wired into the existing `reconcileStrandedAssignedIssues` sweep (reuses its interval, no new timer). Each tick it finds `blocked` issues with an active recovery action and zero first-class blocker relations, then either re-wakes the recovery owner (bounded 15-minute backoff + the action's own `maxAttempts`) or leaves a fresh, bounded-frequency comment when no invokable owner remains.
- Added `wasRecoveryActionResolvedSinceRun` and gated all three `didAutomaticRecoveryFail` escalation call sites (assignment_recovery / issue_continuation_needed / execution_review_participant_recovery) on it: a recovery action resolved/cancelled after the run finished now buys the issue one fresh recovery-retry dispatch instead of an immediate re-escalation on the same stale evidence.
- Added a monitor-only durable-wait-path exemption for any terminal run (not just succeeded), deliberately narrower than the existing `hasPersistedDurableWaitPath` helper so it does not shadow the separate waiting-on-review blocker-conversion path.

## Verification

- `cd server && npx tsc --noEmit -p .` — clean.
- `cd server && npx vitest run src/__tests__/heartbeat-process-recovery.test.ts` — 105/105 passed (includes 2 new tests: one proving a manually-reset `todo` issue gets a fresh dispatch instead of re-escalating, one proving a scheduled monitor survives an unsuccessful terminal run).
- `cd server && npx vitest run src/__tests__/issue-recovery-actions.test.ts` — 49/49 passed (5 tests added for the orphaned-blocked re-wake path: re-wake after backoff, no re-wake inside backoff, skip when a real blocker exists, bounded-frequency reminder with no invokable owner, stop re-waking once `maxAttempts` is exhausted).
- `cd server && npx vitest run src/__tests__/server-startup-feedback-export.test.ts` — 16/16 passed.

## Risks

- Behavioral change, not a schema/migration change — no rollback beyond reverting the commit.
- The new orphaned-blocked re-wake path is scoped narrowly to recovery-caused blocks (identified by an active `issue_recovery_actions` row), so it should never touch a human-authored `blocked` issue that intentionally has no blocker relation.
- The `wasRecoveryActionResolvedSinceRun` guard only *widens* when escalation is skipped in favor of one fresh retry dispatch; it never escalates in a case that previously would not have. Existing escalation tests for the unmodified paths (no recent resolution recorded) are unchanged and still pass.
- The monitor-preservation exemption is deliberately scoped to `monitorNextCheckAt` truthiness only (matching the existing succeeded-run exemption's semantics), not the broader `hasPersistedDurableWaitPath` (which also checks delegated-blocker relations) — broadening to the full helper regressed the waiting-on-review blocker-conversion test during review of this change; narrowing it fixed that regression while still covering the reported monitor-destruction case.

## Model Used

Claude Sonnet 4.5 (claude-sonnet-4-5), extended/agentic tool-use mode, editing and running the repository's own test suite (embedded PostgreSQL) locally.

## Checklist

Note: the branch name (`fix/age-691-orphaned-blocked-recovery-actions`) was created before this checklist requirement and predates this update; renaming it now would require opening a fresh PR. Flagging rather than silently unchecking.

### DIFF
--- a/server/src/__tests__/heartbeat-process-recovery.test.ts
+++ b/server/src/__tests__/heartbeat-process-recovery.test.ts
@@ -5897,6 +5897,57 @@ describeEmbeddedPostgres("heartbeat orphaned process recovery", () => {
     )).toBe(true);
   });
 
+  it("gives assigned todo work one fresh dispatch instead of re-escalating a stale run after a manual blocked-to-todo reset (AGE-691)", async () => {
+    const { companyId, agentId, issueId, runId } = await seedStrandedIssueFixture({
+      status: "todo",
+      runStatus: "failed",
+      retryReason: "assignment_recovery",
+      runErrorCode: "process_lost",
+    });
+    // Mirrors what classifySourceRecoveryRevalidation records when a human/agent
+    // manually resets `blocked` -> `todo` via a plain issue PATCH (not the
+    // dedicated recovery-actions resolve endpoint): the previously active
+    // recovery action is cancelled, but the stale failed `latestRun` above is
+    // untouched by that reset.
+    await db.insert(issueRecoveryActions).values({
+      companyId,
+      sourceIssueId: issueId,
+      kind: "stranded_assigned_issue",
+      status: "cancelled",
+      ownerType: "agent",
+      ownerAgentId: agentId,
+      previousOwnerAgentId: agentId,
+      cause: "process_lost",
+      fingerprint: `manual-reset:${issueId}`,
+      nextAction: "Resume source work",
+      outcome: "cancelled",
+      resolutionNote: "Recovery action became stale because the source issue was manually moved from blocked to todo.",
+      resolvedAt: new Date("2026-03-19T00:06:00.000Z"),
+      createdAt: new Date("2026-03-19T00:01:00.000Z"),
+      updatedAt: new Date("2026-03-19T00:06:00.000Z"),
+    });
+    const heartbeat = heartbeatService(db);
+
+    const result = await heartbeat.reconcileStrandedAssignedIssues();
+    expect(result.escalated).toBe(0);
+    expect(result.dispatchRequeued).toBe(1);
+    expect(result.issueIds).toEqual([issueId]);
+
+    const issue = await db.select().from(issues).where(eq(issues.id, issueId)).then((rows) => rows[0] ?? null);
+    expect(issue?.status).toBe("todo");
+
+    const runs = await db.select().from(heartbeatRuns).where(eq(heartbeatRuns.agentId, agentId));
+    expect(runs).toHaveLength(2);
+    const retryRun = runs.find((row) => row.id !== runId);
+    expect((retryRun?.contextSnapshot as Record<string, unknown>)?.retryReason).toBe("assignment_recovery");
+
+    const recoveryIssues = await db
+      .select()
+      .from(issues)
+      .where(and(eq(issues.companyId, companyId), eq(issues.originKind, "stranded_issue_recovery")));
+    expect(recoveryIssues).toHaveLength(0);
+  });
+
   it("blocks an already stranded recovery issue without creating a recovery child", async () => {
     const { companyId, issueId } = await seedStrandedIssueFixture({
       status: "todo",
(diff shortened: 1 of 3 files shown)


dzianisv:
Review from AgentLabs (issue AGE-691 owner). Verdict: **the behaviour fix is right; the open Greptile P1 is valid and should be fixed before merge.** Everything else on the PR is green.

**The P1 is real, not a false positive.** `enqueueWakeup` (`server/src/services/heartbeat.ts:6593`) never dedupes on `idempotencyKey` — it only persists it (`agent_wakeup_requests.idempotency_key` has no unique index; see `packages/db/src/schema/agent_wakeup_requests.ts`, only three non-unique indexes). The `hasQueuedIssueWake` check and the later absolute `attemptCount: action.attemptCount + 1` write are therefore a plain TOCTOU: two overlapping `reconcileStrandedAssignedIssues` sweeps (the periodic tick at `server/src/index.ts:753` is fire-and-forget with no in-flight guard) can both read the same stale action row, both enqueue, and then both write the same absolute `attemptCount`, so two owner runs consume one retry slot.

Severity is low relative to what this PR fixes (permanently unwakeable issue vs. one duplicate wake), but the fix is small and closes both Greptile findings at once.

**Suggested shape — compare-and-set claim, then release on a failed enqueue:**

```ts
const claimed = await db
  .update(issueRecoveryActions)
  .set({ attemptCount: action.attemptCount + 1, lastAttemptAt: now, updatedAt: now })
  .where(and(
    eq(issueRecoveryActions.id, action.id),
    eq(issueRecoveryActions.attemptCount, action.attemptCount),
    action.lastAttemptAt
      ? eq(issueRecoveryActions.lastAttemptAt, action.lastAttemptAt)
      : isNull(issueRecoveryActions.lastAttemptAt),
  ))
  .returning({ id: issueRecoveryActions.id });
if (claimed.length === 0) { result.skipped += 1; continue; }   // another sweep won the claim

const wakeRun = await deps.enqueueWakeup(...);
if (!wakeRun) {
  // release the claim so a failed enqueue does not consume the retry budget
  await db.update(issueRecoveryActions)
    .set({ attemptCount: action.attemptCount, lastAttemptAt: action.lastAttemptAt, updatedAt: new Date() })
    .where(eq(issueRecoveryActions.id, action.id));
  result.skipped += 1;
  continue;
}
```

That keeps the property the third commit was added for (a failed enqueue must not burn a retry) while making the claim atomic, so the ordering argument stops mattering.

**Test to add** (the current `issue-recovery-actions.test.ts` covers backoff and enqueue failure, not concurrency): invoke `reconcileOrphanedBlockedRecoveryActions` twice concurrently (`await Promise.all([...])`) against one orphaned action and assert exactly one `enqueueWakeup` call and `attemptCount === 1`.

Not blocking, for the record: the required-PR-sections P2 was already addressed by the description rewrite.

dzianisv:
Pushed the atomic claim fix for the Greptile P1 (commit 2166426).

`reconcileOrphanedBlockedRecoveryActions` now claims the retry slot with a compare-and-set UPDATE (`attemptCount`/`lastAttemptAt` guarded on the pre-read values) before calling `enqueueWakeup`. Zero rows returned = another sweep already won the claim this tick, so this pass skips. If `enqueueWakeup` then returns falsy, the pre-claim values are restored so a failed enqueue doesn't consume the retry budget.

Added a concurrency test (`AGE-709`): two concurrent `reconcileOrphanedBlockedRecoveryActions()` calls against one orphaned blocked action now produce exactly one `enqueueWakeup` call and `attemptCount === 1`.

Verified locally:
- `npx vitest run src/__tests__/issue-recovery-actions.test.ts` — 51 passed
- `npx tsc --noEmit -p .` — clean

Waiting on CI to re-run on the new commit.

PR #11624, opened 2026-08-18
fix(heartbeat): never process_lost/retry a hot-restart-adopted run after its process exits

dzianisv:
## Thinking Path

> - The heartbeat service's orphaned-run reaper (`reapOrphanedRuns`) recovers a run whose child process died, including a run this server *adopted* from a previous process across a hot restart
> - The adoption guard only protects an adopted run while its process is still alive. The instant the adopted process (and its process group) exits, the guard's own condition goes false and the run falls into the generic "process lost" path
> - That generic path can label the run `process_lost` and auto-enqueue a retry, even though an adopted run may already have finished its real work (posted comments, patched issues, opened PRs) before this server ever observed it. Retrying it duplicates that work
> - This pull request adds a dedicated finalization path for exactly this case: never `process_lost`, never auto-retried, and — when a persisted stdout/stderr log is available — the real success/failure outcome is recovered from it instead of guessing
> - The benefit is no more duplicate side effects from a completed run that happened to be adopted across a restart, with an explicit "unknown outcome" floor (also non-retrying) when the real outcome can't be recovered

**What happened?**
`reapOrphanedRuns` computes `processPidAlive`/`processGroupAlive` and only skips reaping an adopted run while `(processPidAlive || processGroupAlive) && readHotRestartAdoptionMetadata(...)` is true. Once the adopted process and its process group both exit, that condition is false and the run falls straight into the unconditional `process_lost` finalization further down the function, which can set `shouldRetry = true` and enqueue an automatic retry of a run that may have already completed all of its real side effects.

**Expected behavior**
A run carrying hot-restart adoption metadata should never be finalized as `process_lost` and should never be auto-retried once adopted, regardless of whether its process is still alive. When possible, its real terminal outcome should be recovered from its persisted output log; otherwise it should land in an explicit, non-retrying "unknown outcome" state.

**Steps to reproduce**
1. Start a local-adapter run and let the server hot-restart while it is running, so the new process adopts it (recorded via `resultJson.hotRestart.adopted = true`).
2. Let the adopted process (and its process group) exit on its own, without the new server process ever observing that exit live.
3. Let the heartbeat's `reapOrphanedRuns` sweep run.
4. Observe the run is finalized `errorCode: "process_lost"` and, if `processLossRetryCount < 1`, a retry run is auto-enqueued for an agent/task that may have already produced real comments/PRs/issue updates.

## What Changed

- Added `finalizeAdoptedRunAfterProcessExit` in `server/src/services/heartbeat.ts`: a dedicated finalization path in `reapOrphanedRuns` for a run carrying hot-restart adoption metadata whose process (pid AND process group) is confirmed dead. It never uses `errorCode: "process_lost"` and never auto-enqueues a retry.
- The new path also passes `suppressImmediateRecovery: true` to `releaseIssueExecutionAndPromote`, which has its own independent "queue a continuation run" auto-retry mechanism keyed only on `run.status === "failed"` — without this, that second mechanism would still duplicate the adopted run's work even though the process-loss retry path was closed.
- Added an optional adapter capability, `recoverAdoptedRunOutcome`, on `ServerAdapterModule` (`packages/adapter-utils/src/types.ts`). It recovers the real outcome from a run's persisted stdout/stderr log content, reusing the same terminal-result presence check (`resultJson !== null`) that `runChildProcess`'s existing `terminalResultCleanup.hasTerminalResult` already performs live (`packages/adapter-utils/src/server-utils.ts`).
- Wired `recoverAdoptedRunOutcome` for the `claude_local` adapter (`server/src/adapters/registry.ts`) using the adapter's own `parseClaudeStreamJson`/`describeClaudeFailure` parsers.
- When a terminal result is recovered, the run finalizes with the real success/failure outcome. When it isn't (no log path persisted, adapter has no hook, or no terminal result in the log), the run finalizes with a distinct, non-retrying `errorCode: "adopted_run_outcome_unknown"`.
- Added `readAdoptedRunOutputLogTail` to read the tail of a log file (capped at the same `MAX_CAPTURE_BYTES` a live run capture already enforces).
- Added two tests to `server/src/__tests__/heartbeat-process-recovery.test.ts` covering both the recovered-outcome and unknown-outcome branches, and verified no regression to the existing adoption-still-alive and process_lost-retry tests.

## Verification

```
cd server
pnpm exec vitest run src/__tests__/heartbeat-process-recovery.test.ts
# 105 passed (105)
```

- New test: "recovers the real outcome from a hot-restart-adopted run's persisted log once its process has fully exited" — mocks the adapter's `recoverAdoptedRunOutcome` to report success; asserts the run finalizes `succeeded`, `errorCode` is not `process_lost`, and exactly one row exists (no retry). Fails against `master`.
- New test: "finalizes a hot-restart-adopted run as unknown-outcome (never process_lost, never retried) once its process has fully exited with no recoverable terminal result" — no adapter hook available; asserts `errorCode: "adopted_run_outcome_unknown"` and exactly one row exists (no retry). Fails against `master`.
- Pre-existing tests re-verified green: "queues exactly one retry when the recorded local pid is dead" (non-adopted `process_lost` path, unchanged behavior), "reports adopted hot-restart runs before startup reap can mark them process_lost", and "keeps process-group-only hot-restart adoptions out of process_lost reaping" (adoption-still-alive guard, unchanged).
- `packages/adapter-utils` typecheck — clean; `packages/adapter-utils/src/server-utils.test.ts` — 92/92 passing.
- `packages/adapters/claude-local/src/server/parse.test.ts` — 46/46 passing.
- One pre-existing, unrelated failure in `packages/adapters/claude-local/src/server/execute.acp-fallback.test.ts` reproduces identically on an unmodified checkout at the same base commit — not touched by this change.

## Risks

- Behavioral shift: an adopted run whose process later exits with no recoverable log evidence now finalizes as `adopted_run_outcome_unknown` (failed, non-retrying) instead of `process_lost` (failed, retried once). This is intentional — the whole point of this change is to stop the automatic retry for this case — but any downstream code or dashboard that specifically filters on `errorCode: "process_lost"` will no longer see these runs under that code. `dashboard-service.test.ts`'s existing `process_lost` bucket assertions still pass unaffected, since they cover a different run shape.
- `recoverAdoptedRunOutcome` is currently wired for exactly one adapter (`claude_local`). Every other local adapter safely falls through to the non-retrying unknown-outcome floor rather than silently guessing — no adapter regresses to worse behavior than before.
- Low risk otherwise: the change is additive (a new capability field, a new finalization function) and does not alter the existing "still alive" adoption guard or the existing non-adopted `process_lost` retry path, both of which are covered by pre-existing, still-green tests.

## Model Used

Claude (Anthropic), via the `pi` coding agent harness — extended reasoning/thinking mode enabled, with tool use (file read/edit/bash) and no code execution beyond the local test/typecheck runs listed above.

## Checklist

### DIFF
--- a/packages/adapter-utils/src/index.ts
+++ b/packages/adapter-utils/src/index.ts
@@ -33,6 +33,7 @@ export type {
   AdapterRuntimeCommandSpec,
   AcpTargetDescriptor,
   ServerAdapterModule,
+  AdoptedRunOutcomeRecovery,
   QuotaWindow,
   ProviderQuotaResult,
   TranscriptEntry,
--- a/packages/adapter-utils/src/types.ts
+++ b/packages/adapter-utils/src/types.ts
@@ -500,6 +500,32 @@ export interface ServerAdapterModule {
    * and provisioned in fresh remote environments such as sandboxes.
    */
   getRuntimeCommandSpec?: (config: Record<string, unknown>) => AdapterRuntimeCommandSpec | null;
+
+  /**
+   * Optional (AGE-697): recover the real terminal outcome of a local child
+   * process from its persisted stdout/stderr log content. Used only when a
+   * run's process was adopted across a hot restart (AGE-656) and later exits
+   * without this server observing the exit live -- at that point the process
+   * is gone and its in-memory exit code was never captured, so the only
+   * remaining evidence is whatever the adapter itself already wrote to its
+   * log files.
+   *
+   * Returns `null` when no terminal result is present in the given output;
+   * the caller then finalizes the run with an explicit unknown-outcome state
+   * instead of guessing at success/failure. This deliberately reuses the same
+   * terminal-result detection an adapter already performs live via
+   * `runChildProcess`'s `terminalResultCleanup.hasTerminalResult`
+   * (packages/adapter-utils/src/server-utils.ts) -- an adapter that doesn't
+   * wire that live detection also doesn't implement this hook and simply
+   * falls back to the unknown-outcome floor.
+   */
+  recoverAdoptedRunOutcome?: (output: { stdout: string; stderr: string }) => AdoptedRunOutcomeRecovery | null;
+}
+
+export interface AdoptedRunOutcomeRecovery {
+  outcome: "succeeded" | "failed";
+  errorMessage?: string | null;
+  summary?: string | null;
 }
 
 // ---------------------------------------------------------------------------
--- a/server/src/__tests__/heartbeat-process-recovery.test.ts
+++ b/server/src/__tests__/heartbeat-process-recovery.test.ts
@@ -47,7 +47,7 @@ import {
   getEmbeddedPostgresTestSupport,
   startEmbeddedPostgresTestDatabase,
 } from "./helpers/embedded-postgres.js";
-import { runningProcesses } from "../adapters/index.ts";
+import { runningProcesses, getServerAdapter } from "../adapters/index.ts";
 const mockTelemetryClient = vi.hoisted(() => ({ track: vi.fn() }));
 const mockTrackAgentFirstHeartbeat = vi.hoisted(() => vi.fn());
 const mockTerminateLocalService = vi.hoisted(() => vi.fn());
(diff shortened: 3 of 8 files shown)


dzianisv:
Addressed both Greptile findings in d178fd9f7:

- **Issue 1 (persisted output path unavailable in production)** — valid. Added `reconstructRunLogStreamsTail`, a fallback that recovers a run's stdout/stderr from the general-purpose run transcript log (`run.logStore`/`run.logRef` via `runLogStore`), which every run already writes today, independent of whether `contextSnapshot.stdoutLogFilePath` has ever been persisted (that field is populated by the sibling file-backed-local-run-output PR, #11622, not yet merged). Without this, real-outcome recovery was correct-but-unreachable until that second PR lands. New test writes a real `runLogStore`-backed NDJSON fixture (including a terminal result split across two appends, to exercise the cross-page leftover-line buffer) and asserts recovery via `recoverySource: "run_log_store"`.
- **Issue 2 (Claude outcome semantics)** — valid, and a real bug. Fixed `recoverAdoptedRunOutcome` to key success/failure off `is_error` alone instead of requiring `subtype === "success"`. Live execution (`execute.ts`'s `toAdapterResult`) computes `failed = !parsedSucceeded && (exitCode !== 0 || is_error)`, so a clean non-"success" subtype (`model_refusal`, `error_max_turns`) with `is_error: false` is NOT a live failure. `exitCode` is unavailable post-exit (the process is already gone), so matching on `is_error` alone is the closest correct approximation without it.
- **Issue 3 (missing PR sections)** — fixed in the earlier push; the body now has Thinking Path / What Changed / Verification / Risks / Model Used and a completed dedup-search checklist item.

Full test suite re-verified: `heartbeat-process-recovery.test.ts` 106/106, `adapter-registry.test.ts` 14/14, `claude-local/parse.test.ts` 46/46, server typecheck clean.

dzianisv:
Addressed the follow-up finding in 79f175e1e:

- **Transcript-tail-unreachable-past-32MB** — valid. Added an optional \`readTail\` method to \`RunLogStore\` (local-file stat + read-last-N-bytes, with an S3 headObject+range fallback for the pod-rolled case) that reaches the true end of the log in O(1) requests regardless of total size, instead of forward-paginating from offset 0 with a scan cap. \`reconstructRunLogStreamsTail\` now prefers \`readTail\`; the old forward-pagination path is kept only as a fallback for a store implementation that hasn't implemented it (never in production).
- New \`run-log-store.test.ts\` proves this directly: a log built far larger than a small test \`maxBytes\` still surfaces its terminal marker near the true end, and asserts the reconstructed tail never contains the log's first line (so it's provably a real tail, not an early-stopping forward scan).

Re-verified: \`heartbeat-process-recovery.test.ts\` 106/106, new \`run-log-store.test.ts\` 3/3, server typecheck clean.

PR #11674, opened 2026-08-19
feat: add least-privilege agent invocation permission

Lancechung888:
## Summary

- add a company-scoped `agents:invoke` permission
- require it for board calls to `POST /api/agents/:id/wakeup`
- keep agent self-invocation and the legacy `/heartbeat/invoke` compatibility contract unchanged
- backfill the permission for existing owner/admin roles through the existing role-default grant sync
- document the least-privilege boundary and expose it in OpenAPI

## Why

The modern wakeup route reused `agents:create`, so an invoke-only dispatcher had to receive agent-creation authority. This prevented a least-privilege automation identity from waking an existing agent without also gaining unrelated management capability.

## Impact

Board callers can now be granted wakeup authority without permission to create or configure agents. Cross-company targets remain non-enumerating `404`s, and authenticated agents can still invoke only themselves.

## Validation

- `vitest` targeted route and role suite: 68 passed
- `vitest` access and OpenAPI suite: 14 passed
- `@paperclipai/shared` TypeScript check passed
- `@paperclipai/server` TypeScript check passed
- `git diff --check` passed

### DIFF
--- a/doc/SPEC-implementation.md
+++ b/doc/SPEC-implementation.md
@@ -90,6 +90,14 @@ Role-based human permission granularity is V1 — see the `humans-and-permission
 plan, the `principal_permission_grants` table, and the `PERMISSION_KEYS` set
 in `packages/shared/src/constants.ts`.
 
+Agent invocation uses a separate least-privilege permission. Board callers of
+`POST /api/agents/:id/wakeup` require the company-scoped `agents:invoke` grant;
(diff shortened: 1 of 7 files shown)


Lancechung888:
Closing this draft. It was opened in error while evaluating a downstream private Company Factory implementation. No upstream change is requested.

PR #11680, opened 2026-08-19
Inbox: show live, waiting and nested state without opening a task (mobile especially)

simberthon:
## Why

On a phone the inbox does not answer the two questions you open it to answer: *what is running right now* and *what is blocked waiting on me*. Both are rendered on the issue detail page but not in the list, so you have to open tasks one by one to find them. A pending `ask_user_questions` / `request_confirmation` can sit unanswered for days this way.

Four independent defects, four small fixes. Each is verifiable on its own.

## What

**1. The toolbar wraps on iPhone.** `PageTabBar` already ships a compact `<select>` for mobile, but it only engages when `value` and `onValueChange` are passed, and `Inbox.tsx` passed neither (it wrapped the bar in `<Tabs>` instead). The five-tab `TabsList` therefore rendered at a fixed ~311px inside a ~358px `flex-wrap` toolbar, the action group no longer fit beside it, and the container broke onto a second row.

Passing the props — the same call shape `Search.tsx` already uses — plus making "Mark all as read" icon-only below `sm`, keeps it one row.

At 390x844: toolbar **88px → 44px**, first task at **228px → 184px**, **7 → 8** tasks visible. Desktop is untouched and still renders `TabsList`.

**2. Sub-tasks are invisible on mobile.** `resolveInboxNestingEnabled` forced nesting off below `sm`, and the toggle that could turn it back on was itself `hidden sm:inline-flex`, so it could not be reached. Nesting now applies at every width, with a per-depth left margin standing in for the desktop tree guides (which are `sm:` only), and the toggle is reachable on mobile.

**3. The Live badge is invisible on mobile.** The badge existed, but its label was `hidden sm:inline` — leaving a bare blue dot indistinguishable from the other coloured dots in the row, which reads as no live marker at all. The label now renders at every width; the sibling "N live below" chip degrades to "N" instead of vanishing.

**4. Nothing marks a task that is waiting on a human.** New `IssueSignalsProvider` counts pending `issue_thread_interaction` items from the company attention feed and `IssueRow` renders a red dot for them. It reuses the sidebar's existing query key, so the two consumers share one request rather than adding a per-row fetch. The context defaults to "no signals", so any tree that does not mount the provider — unit tests, Storybook — renders exactly as before.

Only `issue_thread_interaction` counts toward the dot. Approvals, reviews and failed runs already have their own inbox rows and affordances; folding them in would make the dot mean nothing in particular.

## Verification

Verified in a real browser at a **390x844** iPhone viewport against live data, not a devtools window resize. That distinction matters here: at 500px wide the toolbar does **not** wrap, so defect 1 is invisible unless you emulate the true viewport.

Measured on the running page after the change: toolbar one row at 44px, 2 Live badges rendering their label, 8 awaiting dots with `aria-label` "A question is waiting for your answer", 20 indented nested rows.

`ui` typecheck clean. Touched-area tests pass (111). Full `ui` suite: 3876 passed, with 4 pre-existing failures (timezone- and clipboard-dependent) that fail identically on a clean tree at this base.

## Notes

The dot is deliberately not the blue unread dot: unread means "you have not looked at this", the red dot means "this cannot move until you answer".

### DIFF
--- a/ui/src/components/IssueColumns.tsx
+++ b/ui/src/components/IssueColumns.tsx
@@ -175,12 +175,14 @@ export function InboxIssueMetaLeading({
       ) : null}
       {isLive && (
         <Badge variant="ghost"
+          data-testid="issue-live-badge"
+          title="An agent run is executing on this task right now."
           className={cn(
-            "px-1.5 sm:gap-1.5 sm:px-2",
+            "gap-1 px-1.5 sm:gap-1.5 sm:px-2",
             "bg-blue-500/10",
           )}
         >
-          <span className="relative flex h-2 w-2">
+          <span className="relative flex h-2 w-2 shrink-0">
             <span className="absolute inline-flex h-full w-full animate-pulse rounded-full bg-blue-400 opacity-75" />
             <span
               className={cn(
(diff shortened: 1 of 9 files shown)


simberthon:
Closing: the owner of this instance has decided not to contribute changes upstream. These fixes will be maintained as a local overlay instead. Thanks for the project.

PR #11681, opened 2026-08-19
fix(server): re-resolve managed instructions root at heartbeat config-assembly

dzianisv:
## What

`applyBundleConfig()` in `server/services/agent-instructions.ts` persists an absolute path (`path.resolve(rootPath, entryFile)`) into `adapterConfig.instructionsFilePath`/`instructionsRootPath` at config-write time. `deriveBundleState()` + `recoverManagedBundleState()` already re-resolve and self-heal this on the bundle read/list HTTP API path, but nothing re-ran that recovery before the config reached an adapter's `execute()` — so moving the Paperclip data dir silently orphaned every managed agent's runtime instructions even though the API already knew the config was stale.

Upstream issue: https://github.com/paperclipai/paperclip/issues/11391 (part 2).

## Fix

Instead of touching each of the 9 adapter packages individually, this fixes it once at the single choke point where every run's `adapterConfig` is assembled: `heartbeat.ts`, right after `parseObject(agent.adapterConfig)`.

- `agent-instructions.ts`: exports `resolveHeartbeatManagedInstructionsPatch`, which composes the existing `deriveBundleState` + `recoverManagedBundleState` and returns `{}` (external mode, or an already-correct managed config) or a `{ instructionsRootPath, instructionsFilePath, instructionsEntryFile, warnings }` patch reflecting the current on-disk managed root.
- `heartbeat.ts`: applies that patch to a copy of `config` before it flows into `buildExecutionWorkspaceAdapterConfig` / `mergedConfig` / `resolvedConfig` / `runtimeConfig`, and logs any correction via the existing pino logger with `agentId`/`companyId` and old+new paths. `instructionsEntryFile` is corrected too, since `resolveInstructionsConfigFingerprintMetadata` prefers it over `instructionsFilePath` when resolving the session-freshness fingerprint path.

## Deviation from the linked ticket, flagged explicitly

The internal ticket driving this PR asked for the fix in "the adapter execution path" (each adapter's `execute()`). This PR instead fixes it once at `heartbeat.ts` config-assembly, upstream of every adapter, so the adapter package never sees a stale path in the first place — avoiding a 9-package duplicate of the same correction. The reproduction steps in that ticket ("start with one data dir, move state to another, trigger a heartbeat, confirm the heartbeat resolves the current root") are heartbeat-specific, which this satisfies. A reviewer who prefers the literal per-adapter approach should say so; happy to duplicate the correction into `packages/adapters/*/src/server/execute.ts` instead/also.

Out of scope for this PR (flagging to maintainers as a follow-up major-version cleanup): never persisting an absolute path at all — store only `mode` + `entryFile` and resolve the root at every read.

## Tests

- `server/src/__tests__/agent-instructions-service.test.ts`: repro for `resolveHeartbeatManagedInstructionsPatch` — managed config recorded under data-dir A, data-dir re-pointed to B via `PAPERCLIP_HOME`/`PAPERCLIP_INSTANCE_ID` swap (no symlink shim), function returns the corrected root/path for B. Plus entry-file correction and no-op cases.
- `server/src/__tests__/pi-local-execute.test.ts`: integration test proving the corrected `instructionsFilePath` is actually consumed end-to-end by a real `@paperclipai/adapter-pi-local` `execute()` run (captured `--append-system-prompt` contains the recovered file's contents, not the stale one).

## Verified

```
pnpm --filter @paperclipai/server run typecheck   # exits 0
npx vitest run src/__tests__/agent-instructions-service.test.ts   # 12/12 pass
npx vitest run src/__tests__/pi-local-execute.test.ts             # 4/4 pass
npx vitest run src/__tests__/heartbeat-*.test.ts                  # 609 tests, 607 pass
```

The 2 failures in `heartbeat-workspace-branch-containment.test.ts` are pre-existing macOS `/var` vs `/private/var` tmpdir-symlink string-equality flakes in a `providerRef` assertion, unrelated to this change (no instructions-related code in that test file's path).

### DIFF
--- a/server/src/__tests__/agent-instructions-service.test.ts
+++ b/server/src/__tests__/agent-instructions-service.test.ts
@@ -2,7 +2,7 @@ import fs from "node:fs/promises";
 import os from "node:os";
 import path from "node:path";
 import { afterEach, describe, expect, it } from "vitest";
-import { agentInstructionsService } from "../services/agent-instructions.js";
+import { agentInstructionsService, resolveHeartbeatManagedInstructionsPatch } from "../services/agent-instructions.js";
 
 type TestAgent = {
   id: string;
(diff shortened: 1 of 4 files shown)


dzianisv:
Closing in favor of a re-opened PR from a correctly named branch (this branch name embedded an internal ticket id, and the description was missing required template sections).

PR #11698, opened 2026-08-19
fix(server): make silent detect-port fallback loud, and optionally fatal

peterbaker:
## Summary

`server/src/index.ts` silently rebinds to a fallback port via `detect-port` when the configured port is busy — both for the app's listen port and the embedded-Postgres port — logging only a `warn` after the fact. During a fast crash/restart race under a supervisor with `KeepAlive` (e.g. launchd/systemd), a not-yet-released socket on the intended port causes the process to silently start on an unrouted port, with reverse proxies still pointed at the dead one, and no alert. See the incident writeup in the linked issue.

- Adds a shared `resolveListenPort()` helper used by both the app listen port and the embedded-Postgres port call sites.
- Always logs the port mismatch at `error` (previously `warn`) with a greppable `PORT_FALLBACK` prefix.
- Adds an opt-in `PAPERCLIP_PORT_STRICT_MODE` env var: when set to `true`, a port conflict throws instead of silently falling back, so a process supervisor's restart loop keeps retrying until the real port is free rather than serving traffic no one is routing to.
- Default behavior is unchanged (fallback still allowed) — this is additive so existing single-host dev/test setups that rely on the fallback (see `BRO-1558` test coverage) keep working. Production-style single-instance deployments (e.g. FinalFinal's launchd-supervised install) can opt in via the new env var.

## Why not make it fatal by default?

This is shared application code used across every deployment shape, including local dev and parallel test runs that intentionally rely on the fallback. Flipping the default to fatal would be a breaking change with a much larger blast radius than this incident warrants. The opt-in flag gets single-instance production deployments the fail-fast behavior they need without affecting anyone else.

## Test plan

- [x] `vitest run server/src/__tests__/server-startup-feedback-export.test.ts` — 18/18 passing, including 2 new cases covering `PAPERCLIP_PORT_STRICT_MODE` (refuses fallback when enabled, still falls back by default).
- [x] `tsc --noEmit` on `server/` shows no new errors introduced by this change (pre-existing unrelated errors come from a skipped plugin-sdk build step).

Fixes FIN-423 (child of FIN-422).

### DIFF
--- a/server/src/__tests__/server-startup-feedback-export.test.ts
+++ b/server/src/__tests__/server-startup-feedback-export.test.ts
@@ -158,6 +158,7 @@ function buildTestConfig(overrides: Record<string, unknown> = {}) {
     customBindHost: undefined,
     host: "127.0.0.1",
     port: 3210,
+    portStrictMode: false,
     allowedHostnames: [],
     authBaseUrlMode: "auto",
     authPublicBaseUrl: undefined,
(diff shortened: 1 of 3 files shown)


peterbaker:
Superseded by #11734 — same change, but re-opened from a branch name that doesn't embed our internal ticket id (per CONTRIBUTING.md → No Internal Issue References), with the full PR template filled out.

PR #11701, opened 2026-08-19
Allow same-agent comment self-report for valid unbound heartbeat runs

serhiicherepanov:
## Thinking Path

> - Paperclip agents write issue comments through heartbeat runs with audit attribution
> - Cross-issue influence containment requires run binding in normal paths
> - Valid unbound runs currently fail every write with a generic run-context 403
> - Agents with only blocked assigned tasks cannot self-report without a courier workaround
> - This pull request adds a narrow comment-only same-agent self-report path for valid unbound runs
> - PATCH, documents, interactions, resume/reopen, and cross-agent targets stay denied with clearer errors

## Linked Issues or Issue Description

**Problem or motivation:**
Authenticated heartbeat runs without a source task binding cannot post even a comment on their own assigned issues. Agents stuck on blocked tasks hit `403 cross_issue_influence_run_context_required` despite a valid run header.

**Proposed solution:**
Permit `POST /api/issues/{issueId}/comments` for valid unbound runs when the target is assigned to the same agent at write time, with assignee revalidated under row lock in the comment insert transaction. Deny all other mutations and cross-agent/unassigned targets.

**Alternatives considered:**
Blocked checkout and read-only leases were rejected in favor of comment-only self-report without widening mutation authority.

## What Changed

- Extend `observeCrossIssueInfluence` to allow `kind: "comment"` on same-agent assigned targets when the run has no source issue binding, reading assignee from the database under lock
- Add `revalidateUnboundRunSameAssigneeCommentAuth` and call it in the comment insert transaction
- Add denial codes `cross_issue_influence_unbound_run_comment_only` and `cross_issue_influence_unbound_run_same_assignee_required`
- Gate document upsert, interaction create, and resume/reopen comment flags for unbound runs
- Add unit, postgres race, and route authorization-matrix regression tests

## Verification

```sh
pnpm --filter @paperclipai/shared exec vitest run src/issue-write-denial.test.ts
pnpm --filter @paperclipai/server exec vitest run src/__tests__/cross-issue-influence-limit.test.ts
```

Route integration tests were added in `issue-agent-mutation-ownership-routes.test.ts`; supertest hit IPv6 ENETUNREACH in the agent sandbox and should be confirmed in CI.

## Risks

Low-to-medium: bypass is scoped to `kind === "comment"` with same-agent assignment checked atomically at insert time. Document/interaction routes now explicitly deny unbound runs that previously lacked the cross-issue observer.

## Model Used

Not disclosed.

## Checklist

[...]

PR #11739, opened 2026-08-19
fix(server): let a timer run write to an issue it checked out

digital-efficacy:
## Thinking Path

> - Agents report their work through `PATCH /api/issues/:id` and `POST /api/issues/:id/comments`
> - Cross-issue influence containment stops a run from writing to issues outside its scope
> - The guard reads the run's source issue from `heartbeat_runs.contextSnapshot`
> - A scheduler wake (`invocationSource: timer`) starts before the agent picks an issue, so that snapshot holds no `issueId` and no `taskId`
> - The guard then refuses every issue write for the full life of the run, including a write to an issue that the same run checked out
> - This pull request lets the guard confirm the checkout before it refuses, and it keeps the refusal for all other cases
> - The benefit is that a timer heartbeat can report its own work instead of leaving that work unrecorded

Related pull requests, found in a search before this one was opened. They treat the same containment guard:

- Refs #11500 — same diagnosis (a timer wake records no source issue). It counts the write against the per-run cap instead of refusing it.
- Refs #11232 — fixes the same symptom earlier, by recording the issue on the run during a self-checkout.
- Refs #11701 — allows a comment-only self-report path for an unbound run.
- Refs #11362 — closed. It exempted the run's assigned issue, but it also contained a large amount of unrelated work.

This pull request differs from all four. It keeps the refusal as the default, and it permits a write only after it reads the database and confirms that the target issue is checked out by this run.

**What happened?**

An agent run that the scheduler starts receives `403 cross_issue_influence_run_context_required` on every `PATCH /api/issues/:id` call and every `POST /api/issues/:id/comments` call. The refusal also applies to an issue that the same run checked out one moment earlier, and to an issue that the same run created.

The `X-Paperclip-Run-Id` header is present and correct in these requests. It matches the `run_id` claim in the API key. It matches `activeRun.id` from `GET /api/agents/me/inbox-lite`. The error text asks the caller to send that header, so the error gives no usable next step.

The cause is in `server/src/services/cross-issue-influence-limit.ts`. `readRunSourceIssueId()` returns `null` for a timer run, because the scheduler queues the run before the agent reads its inbox and picks an issue. `observeCrossIssueInfluence()` then throws before it reaches the same-issue short circuit:

```ts
```

Not all write paths fail. `PUT /api/issues/:id/documents/:key`, `POST /api/issues/:id/interactions`, `POST /api/companies/:id/issues`, `POST /api/issues/:id/checkout`, and `DELETE /api/issues/:id` all succeed in the same run with the same headers. Only the two paths that the heartbeat instructions name as the standard way to change status and to post a close-out comment are refused.

The practical result is a data quality problem that users see. An agent that cannot write to its own issue does not stop. It creates a new issue instead, or it writes its findings into an issue document. The board then shows stale status and stale assignees on the real issues, next to a growing set of near-duplicate issues that hold the actual result. Work that finished looks unfinished.

**Expected behavior**

A run must be able to write to an issue that it holds a checkout on. That write is in scope by definition, so it is not cross-issue influence. A run that holds no checkout on the target issue must still be refused.

**Steps to reproduce**

Identifiers below are anonymized.

1. Let the scheduler start a heartbeat run. Confirm that `invocationSource` is `timer` and that `contextSnapshot` contains no `issueId` and no `taskId`.
2. Call `POST /api/issues/{issue-a}/checkout`. The call returns 200. `executionRunId` is now the current run id.
3. Call `PATCH /api/issues/{issue-a}` with the body `{"priority":"high"}` and the header `X-Paperclip-Run-Id: {current-run-id}`.
4. The call returns `403 cross_issue_influence_run_context_required`.
5. Call `POST /api/issues/{issue-a}/comments` on the same issue. It returns the same 403.

Step 3 is the smallest reproduction. The run writes to the issue that it checked out one step earlier, and the guard classifies that write as cross-issue.

`51a843e` on `master`, 2026-08-19.

**Node.js version**

v22.23.2

**Operating system**

macOS 26.5.1

## What Changed

- `server/src/services/cross-issue-influence-limit.ts`: when `readRunSourceIssueId()` returns `null`, the guard now queries the `issues` table before it throws. The query matches on company id, target issue id, and `executionRunId` equal to the current run id. The guard returns `null` (no cross-issue influence) when that row exists. It throws the same error as before when the row does not exist.
- `server/src/__tests__/cross-issue-influence-limit.test.ts`: added a test that a timer run with an empty `contextSnapshot` can write to an issue it checked out, and that the guard records no cross-issue activity for that write.
- `server/src/__tests__/cross-issue-influence-limit.test.ts`: renamed the existing fail-closed test to state that it covers the case with no checkout on the target. Its behavior is unchanged.

The query runs inside the transaction that the guard already opens. It runs only on the `null` source path, so a run that carries a source issue does one query less than this path, exactly as before.

## Verification

Run the unit tests for the guard:

```
pnpm vitest run server/src/__tests__/cross-issue-influence-limit.test.ts
```

Result: 11 tests pass, 11 total. The suite includes the two tests that hold the new behavior in place:

- `allows writes for a timer run that has checked out the target issue` — passes only with this change.
- `fails closed when the persisted run has no source issue and no checkout on the target` — passes before and after, and proves that the refusal still applies without a checkout.

To confirm by hand, follow the five reproduction steps above. Step 3 returns 200 with this change. Repeat step 3 against an issue that the run did not check out, and confirm that it still returns 403.

## Risks

Low risk, with one deliberate widening.

- The guard now accepts one class of write that it refused before: a run with no source issue that writes to an issue it holds `executionRunId` on. A checkout is an explicit claim on an issue, so this is the run's own scope.
- Containment is unchanged everywhere else. A run with no source issue and no checkout on the target is still refused with the same error and the same code. The per-run cross-issue cap is untouched. The source-issue path is untouched.
- The change adds one indexed lookup by primary key on the `null` source path only. It does not run for a normal scoped run.
- No migration. No schema change. No API contract change. A client that receives 403 today receives 200 only in the case described above.

## Model Used

Claude Sonnet 4.6 (`claude-sonnet-4-6`) wrote the fix and the tests, with extended thinking and tool use.
Claude Opus 5, 1M context (`claude-opus-5[1m]`) reviewed the change, searched for related pull requests, and wrote this description.

## Checklist

- [ ] I have updated relevant documentation to reflect my changes — no documentation states the refused behavior, so there is nothing to correct.
- [ ] All Paperclip CI gates are green — CI has not run yet on this branch.
- [ ] Greptile is 5/5 with no open P2s, recommendations, or follow-ups — Greptile has not run yet.

### DIFF
--- a/server/src/__tests__/cross-issue-influence-limit.test.ts
+++ b/server/src/__tests__/cross-issue-influence-limit.test.ts
@@ -10,6 +10,8 @@ import {
 function counterDb(
   initialCount = 0,
   runOverrides: Record<string, unknown> | null = {},
+  /** Set to true to simulate that the target issue is checked out by this run. */
+  targetIssueCheckedOutByRun = false,
 ) {
   let observedCount = initialCount;
   const inserted: Array<Record<string, unknown>> = [];
@@ -18,10 +20,19 @@ function counterDb(
       from: () => ({
         where: () => {
           if (Object.keys(selection).includes("count")) {
+            // Activity-log count query.
             return {
               then: (resolve: (rows: unknown[]) => unknown) => resolve([{ count: observedCount }]),
             };
           }
+          if (Object.keys(selection).length === 1 && "id" in selection) {
+            // Issues checkout query (select {id} from issues where ...).
+            return {
+              then: (resolve: (rows: unknown[]) => unknown) =>
+                resolve(targetIssueCheckedOutByRun ? [{ id: "55555555-5555-4555-8555-555555555555" }] : []),
+            };
+          }
+          // Heartbeat-runs row query (uses .for("update")).
           return {
             for: () => ({
               then: (resolve: (rows: unknown[]) => unknown) => resolve(runOverrides === null ? [] : [{
@@ -198,8 +209,9 @@ describe("cross-issue influence limit rollout", () => {
     expect(fake.inserted).toEqual([]);
   });
 
-  it("fails closed when the persisted run has no source issue", async () => {
-    const fake = counterDb(0, { contextSnapshot: {} });
+  it("fails closed when the persisted run has no source issue and no checkout on the target", async () => {
+    // Timer run: empty contextSnapshot, no checkout on the target issue.
+    const fake = counterDb(0, { contextSnapshot: {} }, false);
 
     await expect(observeCrossIssueInfluence(fake.db as never, {
       companyId: "22222222-2222-4222-8222-222222222222",
@@ -211,6 +223,21 @@ describe("cross-issue influence limit rollout", () => {
       status: 403,
       details: { code: "cross_issue_influence_run_context_required" },
     });
+    expect(fake.inserted).toEqual([]);
+  });
+
+  it("allows writes for a timer run that has checked out the target issue", async () => {
+    // Timer run: empty contextSnapshot (no issueId), but has a checkout on the target.
+    const fake = counterDb(0, { contextSnapshot: {} }, true);
+
+    await expect(observeCrossIssueInfluence(fake.db as never, {
+      companyId: "22222222-2222-4222-8222-222222222222",
+      runId: "11111111-1111-4111-8111-111111111111",
+      agentId: "33333333-3333-4333-8333-333333333333",
+      targetIssueId: "55555555-5555-4555-8555-555555555555",
+      kind: "comment",
+    })).resolves.toBeNull();
+    // No cross-issue-influence activity should be recorded.
     expect(fake.inserted).toEqual([]);
   });
 });
--- a/server/src/__tests__/cross-issue-influence-limit.test.ts
+++ b/server/src/__tests__/cross-issue-influence-limit.test.ts
@@ -7,18 +7,49 @@ import {
   observeCrossIssueInfluence,
 } from "../services/cross-issue-influence-limit.ts";
 
+/** One row of the fake `issues` table, holding only the checkout columns the guard reads. */
+type IssueCheckoutRow = { companyId: string; id: string; executionRunId: string };
+
+/**
+ * Collects the bound parameter values out of a drizzle condition.
+ *
+ * `and(eq(a, x), eq(b, y))` flattens to a chunk list holding literal SQL
+ * fragments (whose value is a `string[]`) alongside bound parameters (whose
+ * value is the scalar). Dropping the arrays leaves the values the guard
+ * actually filtered on.
+ */
+function boundParams(condition: unknown): unknown[] {
+  const out: unknown[] = [];
+  const walk = (node: unknown) => {
+    if (!node || typeof node !== "object" || Array.isArray(node)) return;
+    const record = node as Record<string, unknown>;
+    if (Array.isArray(record.queryChunks)) {
+      for (const chunk of record.queryChunks) walk(chunk);
+      return;
+    }
+    if ("value" in record && !Array.isArray(record.value)) out.push(record.value);
+  };
+  walk(condition);
+  return out;
+}
+
 function counterDb(
   initialCount = 0,
   runOverrides: Record<string, unknown> | null = {},
-  /** Set to true to simulate that the target issue is checked out by this run. */
-  targetIssueCheckedOutByRun = false,
+  /**
+   * Rows the fake `issues` table holds. The checkout lookup matches these on
+   * company id, issue id, and execution run id, so a grant is only reachable
+   * when the guard constrains its query to the target issue and to this run.
+   * A lookup that dropped either predicate finds no row here.
+   */
+  issueRows: IssueCheckoutRow[] = [],
 ) {
   let observedCount = initialCount;
   const inserted: Array<Record<string, unknown>> = [];
   const tx = {
     select: (selection: Record<string, unknown>) => ({
       from: () => ({
-        where: () => {
+        where: (condition: unknown) => {
           if (Object.keys(selection).includes("count")) {
             // Activity-log count query.
             return {
@@ -27,9 +58,18 @@ function counterDb(
           }
           if (Object.keys(selection).length === 1 && "id" in selection) {
             // Issues checkout query (select {id} from issues where ...).
+            const params = boundParams(condition);
+            const matched = issueRows.filter((row) => {
+              const values = new Set<unknown>([row.companyId, row.id, row.executionRunId]);
+              // Model the equality filter itself: a row survives when every
+              // bound value matches one of its columns. A predicate the guard
+              // never sent is therefore never applied, so dropping one widens
+              // the result set here exactly as it would in PostgreSQL.
+              return params.length > 0 && params.every((param) => values.has(param));
+            });
             return {
               then: (resolve: (rows: unknown[]) => unknown) =>
-                resolve(targetIssueCheckedOutByRun ? [{ id: "55555555-5555-4555-8555-555555555555" }] : []),
+                resolve(matched.map((row) => ({ id: row.id }))),
             };
           }
           // Heartbeat-runs row query (uses .for("update")).
@@ -210,8 +250,8 @@ describe("cross-issue influence limit rollout", () => {
   });
 
   it("fails closed when the persisted run has no source issue and no checkout on the target", async () => {
-    // Timer run: empty contextSnapshot, no checkout on the target issue.
-    const fake = counterDb(0, { contextSnapshot: {} }, false);
+    // Timer run: empty contextSnapshot, no checkout on any issue.
+    const fake = counterDb(0, { contextSnapshot: {} }, []);
 
     await expect(observeCrossIssueInfluence(fake.db as never, {
       companyId: "22222222-2222-4222-8222-222222222222",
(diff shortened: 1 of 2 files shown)


digital-efficacy:
## Re: checkout authorization can go stale

Thank you — the window is real, and I traced it rather than dismissing it. My conclusion is that it does not block this merge, and that the hardening it implies would make things worse. Evidence below, then what I changed.

### The window exists, but it is the route's, not this branch's

`observeCrossIssueInfluence` commits its own transaction and the route mutation runs after it, so the checkout read cannot span the write. That is accurate.

It is also not specific to this lookup. `PATCH /issues/:id` loads `existing` once (`server/src/routes/issues.ts:9018`) and every gate downstream decides from that snapshot — including `assertAgentIssueMutationAllowed` at `:9031`, which is the actual authorization gate, and which likewise is never re-derived at commit. `shouldUseTransactionalIssueUpdate` (`:9592`) is conditional, and the standalone comment path at `:11873` has no locked re-read at all. So "revalidate immediately when the mutation is committed" is a route-wide change to how issue writes commit, not a fix to these six lines.

### The guard is not the boundary being bypassed

`crossIssueInfluenceLimitError` says it in the file: the cap is "a rate backstop, not a permission decision". Permission to write to the target is decided earlier and separately by `issue:mutate`. A run that won this race still passed that gate. Losing a checkout mid-request does not, by itself, mean the write was unauthorized.

### The new path is narrower than the one already shipped

The same-issue short circuit directly below it (`cross-issue-influence-limit.ts:130-136`) keys on `contextSnapshot`, which no force-release ever clears. A scoped run keeps writing to its source issue after a board force-release — permanently, not for one request. The branch this PR adds keys on `executionRunId`, which the cancel paths (`heartbeat.ts:11021`, `:12671`, `:12891`) and the release paths (`issue-tree-control.ts:881`, `access.ts:376`) all clear, so the grant self-revokes on the next request. Of the two columns, `executionRunId` is deliberately the right one: `checkoutRunId` survives several of those paths and would grant for longer.

### The obvious hardening deadlocks

Adding `.for("update")` to the new lookup would have the guard take `heartbeat_runs` then `issues`. `clearExecutionRunIfTerminal` and `clearCheckoutRunIfTerminal` (`services/issues.ts:5251` and `:5299`) take `issues` then `heartbeat_runs`, and both run on the checkout path this feature is for. That is ABBA lock ordering between two paths that meet in production. I have not made that change, and I have recorded the reason in a comment at the lookup so the next reader does not add it.

### What I did change — 4e7ddc1

Reviewing this pushed on a genuine defect, in my own test rather than in the guard.

The fake DB's `where` discarded its condition, so `allows writes for a timer run that has checked out the target issue` passed for any lookup returning a row. It would have passed with `eq(issues.id, targetIssueId)` deleted. The grant was untested against the containment boundary it widens — which is the property this review is actually about.

The fake now models the equality filter: it holds `issues` rows, and a row survives when every bound value in the condition matches one of its columns, so a predicate the guard never sends is never applied. Two cases follow, one per predicate:

- `fails closed when the timer run's only checkout is on a different issue` — a checkout is a claim on one issue, not a company-wide write pass.
- `fails closed once the target issue's checkout has moved to another run` — the durable half of this finding: after a force-release, the next write is refused, because the guard reads the row and does not cache the grant.

Both verified by mutation. Deleting `eq(issues.id, input.targetIssueId)` fails only the first; deleting `eq(issues.executionRunId, input.runId)` fails only the second. 13 tests pass, 13 total. Typecheck error count is unchanged at 139, all pre-existing in unrelated `plugin-*` modules from an unbuilt `@paperclipai/plugin-sdk`.

### What I am not claiming

The sub-request window is not closed, and I do not think it can be closed here. Closing it means making issue writes commit against a locked re-read on both the PATCH and the comment path. That is worth doing — it would also close the same staleness on the authorization gate, which matters more than this one — but it is a separate change with its own blast radius, and folding it into a 20-line guard fix would be the kind of scope creep that got #11362 closed.

Happy to be overruled if a maintainer reads the containment boundary differently, in which case I would rather revert to refusing timer-run writes outright than land a lock ordering that can deadlock the checkout path.

digital-efficacy:
## Cyrus QA review — approve

Verified against `server/src/services/cross-issue-influence-limit.ts` + the added test cases in `server/src/__tests__/cross-issue-influence-limit.test.ts`. CI is green (typecheck, serialized server suites, e2e).

**On Greptile's finding (checkout authorization staleness / TOCTOU between the guard's read and the route's write):**

Confirmed the gap is real in the sense that `assertCrossIssueInfluenceWithinRunCap` (server/src/routes/issues.ts:~2915, called at ~9203/9207 in the PATCH handler) resolves before `svc.update` actually persists (~9488-9524) — there's async work in between. But:

- This isn't a new window this PR introduces. Every other authorization gate on that same route (`assertAgentIssueMutationAllowed`, `decideIssueAccess("issue:mutate")`, etc.) makes its decision off the same `existing` snapshot loaded once at the top of the handler, then the route mutates later. This PR's new branch has exactly the same shape as gates that already exist — it doesn't widen the blast radius relative to the route's existing architecture.
- The branch only ever grants against the checkout the *calling run itself* holds (`eq(issues.executionRunId, input.runId)`), re-derived fresh from the DB on every request rather than cached — so a released checkout is refused on the *next* request. The worst case is a single in-flight write completing after a concurrent force-release landed a few ms earlier in the same request lifecycle, not a standing bypass.
- Confirmed via `gh pr diff` that new tests cover the two cases that would actually indicate a broken containment boundary: checkout-on-a-different-issue (still refused) and checkout-moved-to-another-run (still refused, doesn't cache the old grant).

Given this PR fixes an actively-hurting production bug (timer-triggered runs silently unable to report their own work — see DIGA-5949/5982/5988 on our board, and I've hit this exact 403 myself on `heartbeat_timer` wakes this week), and the residual TOCTOU is architecturally pre-existing rather than newly introduced, I'd merge this. 

**Non-blocking follow-up worth a ticket:** harden the actual mutation query to re-check `executionRunId` ownership atomically (e.g. `WHERE executionRunId = runId` on the update itself) as defense-in-depth, rather than relying solely on the earlier read-then-decide gate. Doesn't need to hold up this PR.

— Cyrus (QA), reviewing on behalf of DIGA-6100

digital-efficacy:
Cross-linking for whoever reviews this: **#12315** (opened 27-08) changes the same `if (!sourceIssueId) throw ...` branch in `observeCrossIssueInfluence()` and touches the same two files, so the two will conflict. It comes out of #11852, which consolidated #11975 as a duplicate.

Same direction as this PR — allow the unscoped heartbeat run to write rather than deny — but it anchors "home" to the run's *first write target*, persisted, where this PR anchors to the issue the run actually holds the execution checkout on and re-derives per request.

I've left the detailed comparison on #12315 and offered to rebase this onto it if a maintainer prefers a single PR: https://github.com/paperclipai/paperclip/pull/12315#issuecomment-5457350507

No change to this PR's diff; it remains green and `MERGEABLE`.

digital-efficacy:
## Superseding this PR in favour of #12315 — including a defect in this diff

Update for anyone triaging the overlap. #12315 has since taken a `CHANGES_REQUESTED` asking its unbound fallback to prove the run holds the target's checkout, and `dzianisv/paperclip#6` implements exactly that. The fallback there now requires the same company, agent, and `checkoutRunId`/`executionRunId` before it binds, reading the issue without a row lock so the existing lock order is preserved — which is the property this PR was contributing.

Reviewing the two side by side, I also found a defect in **this** PR that I had previously written up as a strength.

This diff re-derives the grant per request from `issues.executionRunId`. A terminal status change nulls that column inside the same update (`services/issues.ts:7689-7693`), so an unbound run that patches status before posting its closing comment gets the original 403 back on the comment:

1. `PATCH {status:"done"}` — guard runs before `svc.update`, `executionRunId` matches, allowed; the update then clears the lock.
2. `POST /issues/:id/comments` — guard re-derives, `executionRunId` is now `null`, throws `cross_issue_influence_run_context_required`.

So this fix only holds for runs that comment before they change status: it makes correctness depend on write ordering within the run. The persisted bind in #12315 survives the status change and does not have that dependency. I called the non-durable re-derive the safer design in my 20-08 reply above; that was wrong, and this is the case that shows it.

Detailed comparison, including why the extra `assigneeAgentId` predicate in #6 does not narrow the grant: https://github.com/paperclipai/paperclip/pull/12315#issuecomment-5490313130

**No maintainer decision is needed between these two PRs.** #12315 is the better design and carries the part this one has no equivalent for (`bindCheckoutRunSourceIssueIfUnset`, binding at checkout time rather than at first write). I will close this PR as superseded once #12315 merges, so there is no conflict to resolve at merge time. Leaving it open until then only so the analysis stays linked from the review.

digital-efficacy:
Closing this as **superseded by #12315**, at the request of the issue owner on our side. This is not an abandonment — the fix this PR asked for is landing, just in a better shape than mine.

### Why #12315 is the one to merge

`EZEZ-QA-Engineer`'s CHANGES_REQUESTED on #12315 (31-08) asked the unbound fallback to prove the run actually holds the target issue — essentially what this PR does. `kallehiitola` implemented exactly that in `dzianisv/paperclip#6`, with the same checkout proof plus an `assigneeAgentId` predicate. I checked that added predicate against every path that stamps a live run id and it cannot refuse a run that legitimately holds the lock ([detail](https://github.com/paperclipai/paperclip/pull/12315#issuecomment-5490313130)).

### Why this PR should not be merged even as a fallback

I found an ordering defect in my own diff. This PR re-derives permission per request from `issues.executionRunId`, but a terminal status change nulls that same column in the same update (`server/src/services/issues.ts:7689-7693`). For an unbound run that sets status before posting its close-out comment:

1. `PATCH {status:"done"}` — guard runs before `svc.update`, `executionRunId` still matches, allowed; the update then clears the lock.
2. `POST .../comments` — separate request, guard re-derives, `executionRunId` is now `null` → **403, the original bug back**.

So this fix only holds for runs that comment *before* they set status: it makes correctness depend on write ordering within a run. The persistent bind in #12315 has no such dependency. I observed the column being cleared directly — a `PATCH` to `in_review` returned `executionRunId: <run-id> -> null` in the same response.

### The branch stays

Head `4e7ddc1` on `fix/cross-issue-timer-run-checkout` is kept, not deleted. The regression tests in `server/src/__tests__/cross-issue-influence-limit.test.ts` (+120/-3) cover the timer-run-writes-its-own-checkout case and are cherry-pickable independently of the fix, if #12315 wants them. CI on this branch was 30/31 green.

If #12315 stalls or is closed without landing, ping me here and I will reopen.

PR #11621, opened 2026-08-18
fix(workspaces): bound close-readiness Git demand

cryppadotta:
## Thinking Path

> - Paperclip is the open source app people use to manage AI agents for work
> - Execution workspaces let operators inspect Git state before they close or remove a workspace
> - The shared Git scheduler limits subprocesses and queued operation keys
> - Close-readiness requests could still add thousands of callers to one shared operation
> - Those callers retained promises, listeners, responses, and retry work after clients disconnected
> - Git timeouts could also release work before a resistant process group had exited
> - This pull request bounds request demand, makes each waiter cancellable, and holds slots until child cleanup completes
> - The benefit is responsive health checks and safe close behavior during an arbitrary request storm

## Linked Issues or Issue Description

Refs #11572

Refs #6402

**What happened?**

The merged workspace Git scheduler limited active children to 2 and queued keys to 32. Close-readiness requests could still add thousands of callers to an existing single-flight scan. One production incident recorded at least 5,965 saturation events and 5,724 joined callers on one key. Queue waits exceeded 320 seconds. Git executions lasted 65 to 99 seconds even though the expected timeout was 10 seconds. Health requests timed out, and the host accumulated thousands of direct defunct Git children.

**Expected behavior**

Paperclip must bound active Git work, queued work, and all live callers. A disconnected caller must leave no waiter behind. A scan with no callers must stop. Timeout cleanup must reap the process group before the scheduler releases its slot. Overload must return an explicit unavailable Git state. An unavailable state must never appear as a clean tree.

**Steps to reproduce**

1. Start one slow close-readiness Git scan.
2. Send at least 5,000 concurrent requests for the same workspace.
3. Send more requests across many other workspace and tenant keys.
4. Disconnect callers while scans are queued and active.
5. Use a fake Git child that ignores `SIGTERM`.
6. Observe unbounded waiters, delayed health responses, and unreaped children before this change.

**Paperclip version or commit**

The production incident occurred on master commit `b446ff59bf` after #11572 merged.

**Deployment mode**

Self-hosted server with local execution workspaces.

**Installation method**

Built from source with pnpm.

## What Changed

- Add hard global demand and per-key waiter caps to the workspace Git scheduler. Count path-resolution reservations before async work starts.
- Add fair admission across tenant and workspace keys. Reject excess callers before another waiter promise, abort listener, or timer is allocated.
- Track live waiters explicitly. Remove aborted waiters at once. Cancel queued and running scans when the last waiter leaves.
[...]

PR #11753, opened 2026-08-19
fix(runtime): stop workspace restart and git-scan churn

cryppadotta:
## Thinking Path

> - The server starts local workspace runtime services and reads workspace state for API routes and background cleanup.
> - A local runtime spawn failure could happen before the code attached its error listener.
> - Workspace metadata reads and the terminal-workspace reaper also ran close-readiness Git scans too often.
> - The live self-hosted service could restart or slow down under this scan load.
> - This pull request handles early spawn errors and moves Git inspection behind explicit readiness and reaper proof paths.
> - The benefit is a stable service and fewer Git subprocesses during normal workspace reads and repeated cleanup ticks.

No public GitHub issue exists. This PR includes the bug report inline.

No related PRs were found for `runtime spawn EACCES git scan close_readiness churn`.

**What happened?**

A self-hosted Paperclip server could crash during local workspace runtime startup when the shell spawn failed immediately. The same server could also spend too much time in `execution_workspaces.close_readiness_status` Git scans during normal workspace reads and repeated terminal-workspace reaper passes.

**Expected behavior**

The server should keep running when a local runtime shell cannot spawn. Workspace metadata reads should not run Git close-readiness scans. The terminal-workspace reaper should inspect Git only after cheap database checks show that archive may happen, and it should not recheck the same unchanged undelivered workspace every scheduler tick.

**Steps to reproduce**

1. Configure a local workspace runtime so the configured shell cannot spawn.
2. Start the runtime service or read workspace metadata while workspaces exist.
3. Let the terminal-workspace reaper run against unchanged terminal but undelivered workspaces.
4. Observe a startup failure or a high volume of close-readiness Git scans.

Observed before this branch. The PR head is `0525466a0e028537e73c1448811afd4f3c676339`.

**Deployment mode**

Self-hosted server.

**Installation method**

Built from source.

**Agent adapter(s) involved**

Not adapter-specific. The bug is in core workspace runtime and workspace cleanup code.

**Relevant logs or output**

The live service showed `spawn /bin/bash EACCES`, dense `execution_workspaces.close_readiness_status` scans, and repeated `workspace_git_scan_output_limit` events before the final reaper backoff.

## What Changed

- Attach the local runtime child-process error listener before the first readiness await.
- Keep workspace list and detail metadata reads Git-free by default.
- Keep explicit close-readiness inspection on the close-readiness path.
- Gate terminal-workspace reaper Git scans behind database checks for terminal issue state, cooldown, reopen-pending state, and active runs.
- Back off repeated reaper Git scans for unchanged terminal workspaces that were just proven undelivered or uninspectable.
- Add regression tests for early shell spawn errors, metadata reads, non-terminal reaper candidates, and repeated undelivered reaper candidates.
- Stabilize two CI assertions that were timing-sensitive under hosted runner load without changing production behavior.

## Verification

- `pnpm exec vitest run server/src/__tests__/workspace-runtime.test.ts -t 'contains shell spawn errors'`
- `pnpm exec vitest run server/src/__tests__/workspace-runtime.test.ts -t 'requires Paperclip dev runtime services to pass /api/health readiness|contains shell spawn errors'`
- `pnpm exec vitest run server/src/__tests__/execution-workspaces-service.test.ts -t 'non-terminal reaper candidates|metadata reads|backs off Git inspection'`
- `pnpm exec vitest run cli/src/__tests__/worktree.test.ts -t 'recognizes positive legacy database schema evidence'`
- `pnpm exec vitest run server/src/__tests__/plugin-worker-manager-duplex.test.ts -t 'ends the route at once when one inbound chunk passes the per-chunk limit before a listener binds'`
- GitHub PR checks are green on `0525466a0e028537e73c1448811afd4f3c676339`, including build, typecheck, general shards, serialized shards, e2e, canary, review, Greptile, and security checks.
- Deployed the production runtime/workspace fixes to a live self-hosted instance at build `04ee5d9b2c68b008d0480532acf720755c0773de`. Local and private-network `/api/health` returned `status: ok`.
- Before the reaper fixes, the service logged 490 close-readiness Git scans in 5 minutes. After the final restart, health stayed OK with restart count 0 and child zombies 0. The scan pattern changed to bounded 36-scan batches at the backoff window instead of a repeated 30-second scan burst.

## Risks

Low risk. The change does not alter schema or public API shape. The main behavior change is that metadata reads no longer compute live close-readiness, and the reaper can delay cleanup of unchanged undelivered terminal workspaces by the backoff interval. Callers that need close-readiness still use the explicit close-readiness path.

## Model Used

OpenAI Codex, GPT-5 coding agent with shell, git, and test execution tools.

## Checklist

### DIFF
--- a/cli/src/__tests__/worktree.test.ts
+++ b/cli/src/__tests__/worktree.test.ts
@@ -623,7 +623,7 @@ describe("worktree helpers", () => {
       fs.rmSync(tempRoot, { recursive: true, force: true });
       await tempDb.cleanup();
     }
-  });
+  }, 30_000);
 
   it("ensure-seeded seeds once and fast-exits on the verified manifest", async () => {
     const tempRoot = fs.mkdtempSync(path.join(os.tmpdir(), "paperclip-worktree-ensure-seeded-"));
--- a/server/src/__tests__/execution-workspaces-service.test.ts
+++ b/server/src/__tests__/execution-workspaces-service.test.ts
@@ -602,6 +602,97 @@ describeEmbeddedPostgres("executionWorkspaceService.getCloseReadiness", () => {
     }
   }, 20_000);
 
+  it("skips non-terminal reaper candidates without running Git inspection", async () => {
+    const companyId = randomUUID();
+    const projectId = randomUUID();
+    const executionWorkspaceId = randomUUID();
+    const sourceIssueId = randomUUID();
+    const repoRoot = await createTempRepo();
+    tempDirs.add(repoRoot);
+
+    await db.insert(companies).values({
+      id: companyId,
+      name: "Paperclip",
+      issuePrefix: "PAP",
+      requireBoardApprovalForNewAgents: false,
+    });
+    await db.insert(projects).values({
+      id: projectId,
+      companyId,
+      name: "Non-terminal reaper skip",
+      status: "in_progress",
+    });
+    await db.insert(executionWorkspaces).values({
+      id: executionWorkspaceId,
+      companyId,
+      projectId,
+      mode: "isolated_workspace",
+      strategyType: "git_worktree",
+      name: "Active non-terminal workspace",
+      status: "active",
+      providerType: "git_worktree",
+      cwd: repoRoot,
+      providerRef: repoRoot,
+      repoUrl: "https://github.com/paperclipai/paperclip.git",
+      baseRef: "main",
+      branchName: "main",
+    });
+    await db.insert(issues).values({
+      id: sourceIssueId,
+      companyId,
+      projectId,
+      identifier: "PAP-1",
+      title: "Still open",
+      status: "in_progress",
+      priority: "medium",
+      executionWorkspaceId,
+    });
+    await db
+      .update(executionWorkspaces)
+      .set({ sourceIssueId })
+      .where(eq(executionWorkspaces.id, executionWorkspaceId));
+
+    const fakeBin = await fs.mkdtemp(path.join(os.tmpdir(), "paperclip-fake-git-"));
+    tempDirs.add(fakeBin);
+    const gitMarkerPath = path.join(fakeBin, "git-called");
+    const fakeGitPath = path.join(fakeBin, "git");
+    await fs.writeFile(
+      fakeGitPath,
+      [
+        "#!/bin/sh",
+        "printf 'git invoked\\n' >> \"$PAPERCLIP_GIT_CALLED_MARKER\"",
+        "exit 99",
+        "",
+      ].join("\n"),
+      "utf8",
+    );
+    await fs.chmod(fakeGitPath, 0o755);
+
+    const originalPath = process.env.PATH;
+    const originalMarker = process.env.PAPERCLIP_GIT_CALLED_MARKER;
+    const gitStatusSpy = vi.spyOn(workspaceGitOperationScheduler, "run");
+    process.env.PATH = originalPath ? `${fakeBin}:${originalPath}` : fakeBin;
+    process.env.PAPERCLIP_GIT_CALLED_MARKER = gitMarkerPath;
+
+    try {
+      const sweep = await svc.sweepTerminalWorkspaces();
+
+      expect(sweep).toMatchObject({
+        checked: 1,
+        archived: 0,
+        skippedNonTerminalTree: 1,
+      });
+      expect(gitStatusSpy).not.toHaveBeenCalled();
+      await expect(fs.access(gitMarkerPath)).rejects.toMatchObject({ code: "ENOENT" });
+    } finally {
+      gitStatusSpy.mockRestore();
+      if (originalPath === undefined) delete process.env.PATH;
+      else process.env.PATH = originalPath;
+      if (originalMarker === undefined) delete process.env.PAPERCLIP_GIT_CALLED_MARKER;
+      else process.env.PAPERCLIP_GIT_CALLED_MARKER = originalMarker;
+    }
+  }, 20_000);
+
   it("skips a sweep that starts while another sweep runs", async () => {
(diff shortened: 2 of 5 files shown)


Waseemilyas:
Field confirmation of the released workspace-list defect this PR addresses, plus a current-`master` overlap check. The short version: the collection-list fix is already on `master`, it is in no stable release, and this PR still holds distinct `getById`, reaper and runtime-startup changes.

## What we measured on a released build

Self-hosted instance, `2026.824.1`, global npm install on Linux, one 20-core host. One company held 8,769 execution-workspace rows, and 1,319 of those rows named a git worktree that exists on the host.

`GET /api/companies/:companyId/execution-workspaces` — the route behind `paperclipai workspace list` — did not return before the client timed out. The server later logged HTTP 200 for that request, after the client had already disconnected.

While the request ran, `/api/health` returned nothing for five consecutive 10-second probes, and a single issue GET also timed out in the same period. Each health probe completed its TCP handshake in under a millisecond and then reported `time_starttransfer=0`, and the accept backlog on the API listen socket stayed at 18-19 for over a minute. The server was not accepting queued connections. `/api/health` reads no database, so the immediate failure was above the database. These observations do not on their own isolate the process-level mechanism.

Latency against row count. A `projectId` filter bounded each run:

| rows returned | rows whose worktree exists | `?summary=true` | full list |
|---|---|---|---|
| 13 | 3 | 0.012 s | 0.28 s |
| 116 | 8 | 0.015 s | 1.63 s |
| 431 | 72 | 0.020 s | 11.5 s |
| 1,694 | 170 | 0.092 s | 29.0 s |
| 8,769 (no project filter) | 1,319 | 0.10 s | client timed out; server later logged 200 |

A direct `git status --porcelain=v1 --untracked-files=all` on a representative worktree took about 130 ms from a shell. During the burst, scheduler queue waits reached 29.6 seconds, the cumulative saturation count rose from 9,939 to 10,631 in one bounded reproduction, and log lines reached the journal 90-100 seconds after their own timestamps.

`hydrateWorkspace` turns a saturated or timed-out scan into a warning rather than an error, which is why a request that no client is still waiting for can finish as a 200. The route passes no abort signal, so admitted scans keep running after the client disconnects, and a retry stacks a second burst on the first.

The workaround on a released build is to stop calling the unqualified route. `?summary=true` answered the whole 8,769-row set in 0.10 s, and `/workspace-overview` is also git-free.

**On reproduction:** this is a field observation on a production instance, not a portable script. To reproduce it you need a company with several thousand execution-workspace rows where a large minority name worktrees that exist on the host, then one unqualified GET of the collection route while probing `/api/health` and the listen backlog concurrently. I have not built a fixture that creates that row and worktree mix.

## The `list` part of this PR has already landed on `master`

Commit `b51112798f`, merged as part of #12340 on 29 August, made `list` DB-only, and its inline comment gives the same reason this PR gives.

These parts of the branch are not redundant on current `master`:

- `getById` still calls `hydrateWorkspace`, so a detail read still inspects git. This PR turns that off by default.
- The reaper still runs `inspectGitCloseReadiness` before the terminal-tree check, and has no backoff for unchanged undelivered candidates.
- In `spawnLocalRuntimeService`, the child `error` listener is still attached after `await serviceLog.handle.close()`, which is the first await after `spawn()`.

## No stable release contains the `list` fix

`v2026.817.0`, `v2026.824.0`, `v2026.824.1`, `v2026.831.0` and `v2026.831.1` all call `hydrateWorkspace` for every selected row. `v2026.831.1` is the current `latest` npm tag, so anyone hitting this on a stable build has no release to upgrade to.

## Two things `master` still has, if they are in scope here

1. `list` has no page bound or default status filter. For an unqualified call `buildListConditions` adds no status condition, so the response includes every archived and closed row the company has ever had. A `status` parameter adds the requested condition, and `reuseEligible=true` separately restricts status to `active`, `idle` and `in_review`, but neither is a default. `buildOverviewConditions` excludes `archived` by default; the list path does not.
2. Only the full-tree status scan goes through `workspaceGitOperationScheduler`. `runGit` is a bare `execFile`, and `rev-parse --show-toplevel`, `rev-parse --abbrev-ref HEAD`, `rev-list --left-right --count`, `merge-base --is-ancestor` and the `rev-parse HEAD` in delivery assessment all bypass it. On `master`, `getCloseReadiness` runs them for one workspace and the reaper handles candidates sequentially, so neither path shows an unbounded concurrent fan-out on its own — but the shared bound still does not cover them.

Happy to test a candidate build on this instance.

fayonation:
Field confirmation on current `master` (`5b913e79`), for the reaper part of this change.

On a 2 vCPU self-hosted instance (source install, 2 active runs), the terminal-workspace reaper logged `checked=41..44`, `skippedNonTerminalTree=24..27`, `archived=0` on every sweep, and each sweep spent about 27 full `git status --porcelain=v1 --untracked-files=all` runs, plus `rev-parse` and `rev-list`, before it read the issue tree. With `workspace_git_scan` at 480 to 1200 events per hour, `paperclip.service` consumed 9 h 43 m CPU in 16 h 54 m wall time and the load average reached 14 on 2 cores.

A second source exists that this branch also touches: `runExpensiveGitStatus()` passes `cacheTtlMs: 0`, so the close-readiness scan bypasses the scheduler cache, and the heartbeat tick repeats it for every active workspace every 30 s. That part is measured in #13014 and changed in #13469.

After I applied both changes locally (reaper gated on the database-only issue tree before the Git inspection, plus a close-readiness TTL at or above the heartbeat tick): scans per hour 480 to 1200 → 120, Git execution time 0.6 % of one core, reaper sweeps produce zero Git scans, load average 14 → 2.4, `/api/health` `ok`.

Related: #13469 (overlaps this branch), #11621 (demand bounds and a close-readiness TTL whose maximum is below the tick interval).

Issue #9289, opened 2026-07-09
Automatic database backup can hang indefinitely, silently disabling all future backups until restart


jgxvsx:
### Pre-submission checklist

- [x] I have searched existing open and closed issues and this is not a duplicate.
- [x] I am on the latest released version of Paperclip (or can reproduce on `master`).
- [x] I have confirmed the error originates in Paperclip itself — not in my agent adapter, API provider, or local configuration.

### What happened?

**Version:** 2026.707.0 (confirmed present in 2026.626.0 as well — not a regression, longstanding)
**Deployment:** embedded-postgres, `local_trusted`

**What happened**

A scheduled backup's compression step hung indefinitely. Evidence: the `.sql` dump completed fully (93MB), but the paired `.sql.gz` was truncated at 20 bytes (just a gzip header, no content) — the process producing it never finished. Every hourly backup attempt for the next 5+ days logged only:

WARN: Skipping scheduled database backup because a previous backup is still running

and silently skipped. No error, no alert, no automatic recovery — only a full process restart cleared it.

**Root cause** (`packages/server/dist/index.js`, ~line 451 in 2026.707.0):

```js
let databaseBackupInFlight = false;
const runServerDatabaseBackup = async (trigger) => {
    if (databaseBackupInFlight) {
        if (trigger === "scheduled") {
            logger.warn("Skipping scheduled database backup because a previous backup is still running");
            return null;
        }
        throw conflict(message);
    }
    databaseBackupInFlight = true;
    try { /* dump + compress */ }
    finally { databaseBackupInFlight = false
};
```

databaseBackupInFlight is a plain in-memory flag with no timeout. If the awaited work inside try hangs (e.g. a stuck
child process during compression) instead ofinally never runs, and the flag is stuck true for the remaining lifetime of the process — silently disabling all future backups.

Impact: Indefinite, silent loss of backup coverage. Without external monitoring, an operator has no way to notice
short of manually checking /api/health's datars to be new in 2026.707.0 — goodmitigation, but passive).

Suggested fixes

- Wrap the backup operation in a timeout; on timeout, log an error and force-reset the flag.
- Escalate from WARN to ERROR (or a distinctutive skips instead of repeating the samewarn forever.
- Consider whether the compression step itserocess stdio backpressure) — that's the morep

Report Co-Authored By: Claude Code Sonnet

### Expected behavior

Backup should work like reliably w/o the constant need of supervision.

### Steps to reproduce

[...]

PR #8318, opened 2026-06-19
[SIN-65203] feat(backup): self-healing backup lock — timeout/abort + orphan cleanup + skip alert

ia-dev-sindireceita:
## Summary

Resolves the zombie-lock incident ([SIN-65200](/SIN/issues/SIN-65200)) where a hung JS-engine stream held `databaseBackupInFlight = true` for ~37h because the stalled `await` never settled.

- **Timeout/abort**: `runDatabaseBackup` races against a configurable deadline (default 30 min, `backupTimeoutMs` option). On timeout: pg_dump child gets `SIGKILL`, writer aborted, sql closed, promise rejects — `finally { databaseBackupInFlight = false }` always runs.
- **Orphan `.sql` cleanup**: `cleanupOrphanBackupSqlFiles(dir)` called at server startup before the scheduler arms. Removes `.sql` files with no matching `.sql.gz` (5G orphan from [SIN-64532](/SIN/issues/SIN-64532)).
- **Consecutive-skip alert**: counter in `server/src/index.ts`; escalates WARN → `logger.error` after 3 consecutive scheduled skips; resets on success.

## Test plan

- `packages/db` suite: **9/9 passing** (`npx vitest run src/backup-lib.test.ts`)
- 4 new `cleanupOrphanBackupSqlFiles` unit tests + 1 timeout regression test (1ms timeout against embedded Postgres)

## Deploy gate

Only remaining step after merge: **control-plane restart** of `/opt/paperclip` (Pericles) — also clears the live zombie flag.

Parent: [SIN-65201](/SIN/issues/SIN-65201) · Incident: [SIN-65200](/SIN/issues/SIN-65200)

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #11756, opened 2026-08-20
fix(db): make database backups abortable via AbortSignal

christianlappin:
## Thinking Path

> - The server runs a scheduled database backup every hour inside the server process
> - A module-level `databaseBackupInFlight` flag stops two backups from running together, and a `finally` block clears it
> - `runDatabaseBackup` has no cancellation path, so a stalled gzip step never settles its promise
> - The `finally` block therefore never runs, the flag stays set, and every later backup is skipped with "a previous backup is still running"
> - Backups then stop silently until an operator restarts the server, which is a data-safety problem that gives no signal
> - This pull request adds an optional `AbortSignal` to `runDatabaseBackup` and honours it in both gzip pipelines
> - The benefit is that a caller can bound the backup, release the leaked file descriptors, and stop the silent outage at its source

Refs: #9289

**Related pull requests (please read before review).** Several people have hit
this bug and proposed fixes. I list them so a reviewer can compare, and I am
happy to close this one if you prefer another:

- #8318 — the closest overlap. It also adds an `AbortSignal` to `backup-lib`, plus orphan cleanup and a skip alert. It is a superset of this change. It currently reports a merge conflict and has not moved since June.
- #8517 — adds a watchdog to the pg_dump pipeline only. It also reports a merge conflict.
- #11056 — guards the in-flight flag on the server side. It does not change the pipelines in `backup-lib`. It also reports a merge conflict.
- #8139 — closed as obsolete, because the module it patched was later refactored away.

This pull request is deliberately smaller than #8318. It changes only
`backup-lib`, adds one optional option, and does not change any default
behaviour. It applies cleanly to current `master`. That last point is the main
reason I am offering it: the bug is still open, and none of the pull requests
above merge as they stand.

The closure reason on #8139 also suggests the durable seam is `backup-lib`
rather than the server-side scheduler, because the scheduler module has since
been refactored.

## What Changed

- Add an optional `signal?: AbortSignal` to `RunDatabaseBackupOptions`.
- Call `opts.signal?.throwIfAborted()` before any work starts, so an aborted signal does not open a database connection.
- Pass the signal to the gzip pipeline in the JavaScript engine.
- Pass the signal to the gzip pipeline in the pg_dump engine.
- Send `SIGTERM` to the pg_dump child process on abort. If we close only our end of the pipe, `pg_dump` continues to run and holds a database connection.
- Remove the abort listener in a `finally` block.
- Add three tests.

The existing `catch` block already removes partial `.sql` and `.sql.gz` output,
so an aborted backup cleans up after itself. The option is optional. When a
caller passes no signal, behaviour does not change.

## Verification

Run the suite:

```
pnpm --filter @paperclipai/db build
cd packages/db && npx vitest run src/backup-lib.test.ts
```

The three new tests are:

1. An already-aborted signal rejects, and the backup directory stays empty. This shows no database connection is opened.
2. A signal that is never aborted still produces a valid archive that contains the table. This is a regression guard on the threading.
3. An abort during the backup always settles the promise. The test races the call against a 20 second sentinel and fails if the promise never settles. This is the reported bug, written as a test.

I also confirmed the stream semantics before writing the change:

- `pipeline(a, b, { signal: undefined })` resolves normally, so threading an absent signal cannot break the normal path.
- Aborting a stalled pipeline rejects with `AbortError` and sets `stream.destroyed === true`, so the file descriptor is released.

Field verification: I run this change on a production instance with the
embedded-postgres engine. A manual backup through
`POST /api/instance/database-backups` produced an 80 MB archive in 7.4 seconds.
`gzip -t` reports the archive is valid. The archive ends with `COMMIT;`. No
orphaned `.sql` file and no truncated `.sql.gz` stub remain, which are the two
symptoms described in #9289.

**One pre-existing test failure, unrelated to this change.** The test
"restores fallback COPY data when child tables are dumped before parent tables"
fails with `spawn psql ENOENT`, because `psql` is not installed on my machine. I
confirmed this failure is pre-existing: I stashed my changes, ran that test
again on clean `master`, and saw the identical failure.

## Risks

Low risk.

- The new option is optional. No caller changes are required, and behaviour does not change when the signal is absent.
- No migrations, no schema changes, no public API removals.
- The only new behaviour under abort is stream teardown and a `SIGTERM` to a child process that is already being abandoned.
- The `SIGTERM` is guarded on `exitCode === null && signalCode === null`, so it does not signal a process that has already exited.

## Model Used

Claude Opus 5 (`claude-opus-5`), used with extended thinking and tool use
(shell, file editing, test execution) through Claude Code. The model wrote the
change and the tests. I reviewed the change, ran the suite, and verified the
behaviour on a live instance before submitting.

## Checklist

- [x] I have run tests locally and they pass (see the one pre-existing failure noted above)
- [ ] I have updated relevant documentation to reflect my changes — no documentation change needed. The new option carries a doc comment that explains when to use it.
- [x] All Paperclip CI gates are green — not yet known at submission time. I will fix any failures.
- [x] Greptile is 5/5 with no open P2s, recommendations, or follow-ups — not yet run at submission time.

### DIFF
--- a/packages/db/src/backup-lib.test.ts
+++ b/packages/db/src/backup-lib.test.ts
@@ -588,79 +588,119 @@ describe("runDatabaseBackup abort handling", () => {
 });
 
 describeEmbeddedPostgres("runDatabaseBackup with a signal", () => {
+  async function seedTable(connectionString: string, table: string, rows: number): Promise<void> {
+    const sql = postgres(connectionString, { max: 1, onnotice: () => {} });
+    try {
+      await sql.unsafe(`
+        CREATE TABLE "public"."${table}" ("id" serial PRIMARY KEY, "payload" text NOT NULL);
+      `);
+      await sql.unsafe(`
+        INSERT INTO "public"."${table}" ("payload")
+        SELECT repeat('x', 512) FROM generate_series(1, ${rows});
+      `);
+    } finally {
+      await sql.end();
+    }
+  }
+
+  /** A pg_dump stand-in that never writes output, so the pipeline stays open. */
+  function createStallingPgDump(): string {
+    const dir = createTempDir("paperclip-fake-pgdump-");
+    const bin = path.join(dir, "pg_dump");
+    fs.writeFileSync(bin, "#!/bin/sh\nsleep 120\n");
+    fs.chmodSync(bin, 0o755);
+    return bin;
+  }
+
   it(
     "still completes normally when the signal is never aborted",
     async () => {
       const connectionString = await createTempDatabase();
       const backupDir = createTempDir("paperclip-db-backup-signal-ok-");
-      const sql = postgres(connectionString, { max: 1, onnotice: () => {} });
-
-      try {
-        await sql.unsafe(`
-          CREATE TABLE "public"."signal_ok" ("id" serial PRIMARY KEY, "payload" text NOT NULL);
-        `);
-        await sql.unsafe(`INSERT INTO "public"."signal_ok" ("payload") VALUES ('hello');`);
-      } finally {
-        await sql.end();
-      }
+      await seedTable(connectionString, "signal_ok", 10);
 
       const controller = new AbortController();
       const result = await runDatabaseBackup({
         connectionString,
         backupDir,
         retention: { dailyDays: 1, weeklyWeeks: 1, monthlyMonths: 1 },
+        backupEngine: "javascript",
         signal: controller.signal,
       });
 
       expect(fs.existsSync(result.backupFile)).toBe(true);
       expect(gunzipSync(fs.readFileSync(result.backupFile)).toString("utf8")).toContain("signal_ok");
-      // No stray uncompressed intermediate left behind.
       expect(fs.readdirSync(backupDir).filter((f) => f.endsWith(".sql"))).toEqual([]);
     },
     30_000,
   );
 
   it(
-    "always settles when aborted mid-flight instead of hanging forever",
+    "rejects the javascript engine when the signal aborts after the backup starts",
     async () => {
-      // This is the 2026-07-25 wedge as a test: a stalled backup used to leave
-      // this promise unsettled, which pinned the caller's in-flight guard and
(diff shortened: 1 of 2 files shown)


christianlappin:
Thanks — both findings were valid, and both are now fixed in e3439b211.

**Issue 1 — abort triggered the backup fallback.** Confirmed. In `auto` mode an
abort during `pg_dump` reached the generic engine-failure path, which opened a
new database connection and ran a full JavaScript dump. The abort was only
observed again at the compression step. So the code started more work after the
caller asked it to stop. The catch block now calls `throwIfAborted()` before it
decides to fall back.

**Issue 2 — the abort test accepted normal completion.** Also correct, and it
applied to my first replacement attempt as well. I wrote a stricter test, ran it
against the unfixed code, and it still passed — so the criticism held twice. The
problem was that both paths reject and both finish quickly on a small dataset, so
neither the outcome nor the timing discriminates.

The tests now pin the engine explicitly and assert rejection:

- The javascript-engine test aborts on the next tick, after the entry guard has
  already passed, and requires rejection. This proves the compression step
  observes the signal.
- A new auto-mode test replaces `pg_dump` with a stand-in that never writes
  output, aborts with a distinct reason, and requires that exact reason to
  surface. When the abort stops the backup, `throwIfAborted()` rethrows the
  reason unchanged. If the abort fell through to the JavaScript engine, the
  failure surfaces from the gzip pipeline as `AbortError: The operation was
  aborted` instead. The assertion is exact rather than timing based.

I verified the new test both ways. Without the Issue 1 fix it fails with:

```
AssertionError: expected [Function] to throw error including 'CANCELLED_BY_TEST'
  but got 'The operation was aborted'
```

With the fix it passes. That failure message is exactly the fallback signature
you described.

Suite result: 10 passed, 1 failed. The failure is
"restores fallback COPY data when child tables are dumped before parent tables",
which fails with `spawn psql ENOENT` because `psql` is not installed on my
machine. It is pre-existing and unrelated: I confirmed it by stashing my changes
and running that test again on clean `master`.

PR #11757, opened 2026-08-20
feat(release): bootstrap new npm packages with a placeholder publish

devinfoley:
## Thinking Path

> - Its release pipeline publishes a set of npm packages from CI with npm trusted publishing (GitHub OIDC), gated by `scripts/release-package-manifest.json`
> - A brand-new package name cannot be published by CI directly: the PR bootstrap gate requires the name to resolve on npm, and a trusted-publisher rule can only be configured after the package page exists
> - The current bootstrap helper closes that gap by building the package locally and publishing its real output from a maintainer machine — before the PR that adds the package has passed CI or review
> - This pull request replaces that flow: the helper now publishes a minimal deprecated placeholder at version `0.0.0` that only reserves the name, so every real version ships from CI
> - The benefit is that unreviewed build output never reaches npm, and the bootstrap runs from any checkout (including `master`, before the new package's PR merges) with no local build

The one-time npm bootstrap for a brand-new release package (`pnpm run release:bootstrap-package`).

**Current behavior**

The helper builds the target package locally and publishes the real build output from a maintainer machine. That content has not passed repository CI or review at publish time. The helper also requires the new package to exist in the local workspace, so it must run from the (unmerged) PR branch that adds the package.

**Proposed behavior**

The helper publishes a three-file placeholder at version `0.0.0` (manifest, README, and an `index.js` that throws a descriptive error), waits for the registry to show the package, then deprecates it. The PR bootstrap gate (`scripts/check-release-package-bootstrap.mjs`) only requires the name to resolve on the registry, so the placeholder satisfies it. The first real calver release from CI supersedes the placeholder, and a stable release moves `latest` off it — the same `latest` window that existed under the old flow, but containing an explicit inert stub instead of unreviewed code.

**Reason and benefit**

Real package content only ever reaches npm from CI, after review and merge. The bootstrap becomes safer (scope guard refuses names outside `@paperclipai/`, already-published names are rejected) and simpler (no local build, no workspace state, runs from any checkout).

**Breaking changes**

None at runtime. The helper's CLI surface changes: it now takes a package name only (no directory selector) and drops `--skip-build`. `doc/PUBLISHING.md` is updated to match.

## What Changed

- `scripts/bootstrap-npm-package.mjs`: replaced the build-and-publish flow with a placeholder publish — stages `package.json` + `README.md` + throwing `index.js` at version `0.0.0` in a temp directory, previews with `npm publish --dry-run`, and publishes only with `--publish`. One-time passwords are prompted interactively (never passed as arguments, since they are single-use and would land in shell history), with re-prompt on a rejected or expired code. After publishing, the helper polls the registry until the package is visible (a first publish can lag by minutes; verified live at ~5 minutes), requiring two consecutive sightings before prompting for a second code and deprecating the placeholder so accidental installs warn loudly; on timeout or failure it prints the exact manual `npm deprecate` command. Added an `@paperclipai/`-scope guard and a fail-fast error when `--publish` runs without an interactive terminal. Removed the workspace-plan dependency so it runs from any checkout.
- `scripts/bootstrap-npm-package.test.mjs`: rewrote for the new interface — argument parsing, scope validation, the generated placeholder files (manifest shape, throwing entry point, README), the OTP re-prompt loop, and the registry poll (consecutive-sighting requirement, timeout, transient-error tolerance) via injected fakes.
- `doc/PUBLISHING.md`: rewrote the "One-time bootstrap sequence for a new package" section for the placeholder flow, including the `latest` dist-tag window and the trusted-publishing setup ordering (placeholder publish → trusted publisher rule → `"publishFromCi": true`).
- `.github/scripts/check-pr-release-bootstrap.mjs` (+ test, + wiring in `run-quality-gates.mjs`): new informational commitperclip notice on PRs that need this bootstrap. It fires when the PR newly release-enables a package that is missing from npm, or adds an unpublished `publishFromCi: false` package that published packages declare a `workspace:*` dependency on, and names the exact maintainer command — so contributors know the red `policy` check is not theirs to fix. It never fails the gate (the `policy` job remains the enforcer), only looks up scope-validated names on the registry, and stays quiet on registry errors.

## Verification

- `node --test scripts/bootstrap-npm-package.test.mjs`: 13/13 pass
- `node --test .github/scripts/tests/*.test.mjs`: 147/147 pass (10 new for the PR notice)
- `pnpm run test:release-registry`: 82/82 pass
- Replayed the new PR notice against a real historical PR's live API data (files, manifest at base and head refs): with the registry in its pre-bootstrap state it produces the exact maintainer instruction; with the package bootstrapped it stays silent
- Full live end-to-end run: the flow bootstrapped `@paperclipai/adapter-kimi-local` for real — dry-run preview (634-byte, 3-file tarball), publish, registry visibility after ~5 minutes of propagation lag, deprecation confirmed via `npm view ... deprecated`
- Guards verified live: an already-published name is rejected, an out-of-scope name (`left-pad`) is rejected, unknown options (including the removed `--otp`) are rejected, and `--publish` in a non-interactive shell fails fast before any network call

## Risks

- The `latest` dist-tag points at the deprecated `0.0.0` placeholder until the first stable release supersedes it. This window also existed under the old flow (which parked `latest` at a locally built version); internal consumers are unaffected because release version rewrites pin exact calver versions.
- The registry poll caps at ~10 minutes. If propagation is slower than that, the helper prints the exact `npm deprecate ... --otp <code>` command to run manually once `npm view` resolves.
- The helper no longer validates the name against the workspace release plan, so a typo within the `@paperclipai/` scope would reserve a wrong name. The dry-run preview shows the exact name before any publish.

## Model Used

- Anthropic, **Claude Fable 5** (`claude-fable-5`) via Claude Code, with repository, shell, and Git tooling. It analyzed the existing bootstrap flow and the release scripts (`release-package-map.mjs`, `check-release-package-bootstrap.mjs`, `release.sh` dist-tag handling), wrote the replacement script and tests, updated the documentation, and ran the verification above.

## Checklist

### DIFF
--- a/.github/scripts/check-pr-release-bootstrap.mjs
+++ b/.github/scripts/check-pr-release-bootstrap.mjs
@@ -1,0 +1,151 @@
+#!/usr/bin/env node
+/**
+ * check-pr-release-bootstrap.mjs
+ * Detects release packages that this PR adds or newly release-enables whose
+ * names do not exist on npm yet, and emits an informational notice: the
+ * `policy` CI job will stay red until a maintainer bootstraps the name with
+ * `pnpm run release:bootstrap-package`. Contributors cannot fix that
+ * themselves, so the notice says so explicitly.
+ *
+ * Never fails (informational only) — outputs { passed: true, informational: string[] }
+ *
+ * Runs under pull_request_target from base-branch context: it only parses
+ * JSON and diff text fetched from the GitHub API and queries the npm registry
+ * with scope-validated names. It never executes PR code.
+ */
+import { fileURLToPath } from 'node:url';
+import { ghFetch } from './get-bot-token.mjs';
+import { resolveBaseRef } from './check-pr-dependencies.mjs';
+
+const MANIFEST_PATH = 'scripts/release-package-manifest.json';
+
+// Manifest content comes from the PR head (fork-controlled), so only names
+// matching our scope are ever looked up on the registry.
+const SCOPE_RE = /^@paperclipai\/[a-z0-9][a-z0-9._-]*$/;
+
+const MAX_REGISTRY_LOOKUPS = 5;
+
+function buildContentsPath(repo, filename, ref) {
+  return `/repos/${repo}/contents/${filename}?${new URLSearchParams({ ref }).toString()}`;
+}
+
+async function fetchManifestEntries(fetchFromGitHub, token, repo, ref) {
+  try {
+    const res = await fetchFromGitHub(buildContentsPath(repo, MANIFEST_PATH, ref), token);
+    const parsed = JSON.parse(Buffer.from(res.content, 'base64').toString());
+    return Array.isArray(parsed) ? parsed : [];
+  } catch {
+    return []; // manifest missing or unreadable on this ref
+  }
+}
+
+export async function fetchRegistryPackageExists(packageName) {
+  const res = await fetch(`https://registry.npmjs.org/${encodeURIComponent(packageName)}`, {
+    method: 'HEAD',
+  });
+  if (res.status === 404) return false;
+  if (res.ok) return true;
+  throw new Error(`npm registry returned ${res.status} for ${packageName}`);
+}
+
+// Names this PR newly declares a workspace dependency on, per the diff of any
+// changed package.json. If one of them is an unpublished manifest entry that
+// is not publishFromCi:true, the release manifest validator rejects the PR.
+export function addedWorkspaceDependencyNames(files) {
+  const names = new Set();
+  for (const file of files) {
+    if (!file.filename.endsWith('package.json')) continue;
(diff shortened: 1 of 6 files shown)


devinfoley:
@greptileai Both P2s addressed in the squashed update:

- **Manifest guard skips dependency notices**: the gate now also triggers when a PR adds a `workspace:*` dependency on an `@paperclipai/*` package without touching the manifest (patch parsing is free, so the zero-API fast path only remains for PRs that do neither). Covered by a new test: a dependency added on an existing unpublished `publishFromCi: false` entry produces the notice.
- **OTP argument guarantee**: the helper now hands codes to npm through its environment (`npm_config_otp`) instead of `--otp` argv, and the doc wording was tightened to match — codes never appear on a command line, in shell history, or in a process listing.

PR #11768, opened 2026-08-20
fix(server): gate done transition on open/unmerged/red-CI PRs (AGE-569)

dzianisv:
## Thinking Path

> - The issue-state machine lets any actor move an issue to `done`, including agents
> - The `done` transition does not check whether a linked GitHub pull request is actually merged and green, so an agent can close an issue while its PR is still open, unmerged, or failing CI
> - Nothing else in the system stops this — only a human noticing catches it, and a supervisor audit found real cases of it slipping through
> - This pull request adds a server-side, fail-closed check on the `done` transition that reads the linked pull request's merge state and CI check-run state and rejects the transition with a clear reason when either is not green
> - The benefit is a structural guarantee: an issue cannot be marked `done` while its linked PR is open, unmerged, or has failing/pending CI, with no bypass flag

**What is the problem?**
An issue can be marked `done` even when it references a GitHub pull request that is still open, not merged, or merged with a failing/pending CI run. Nothing in the `PATCH /issues/:id` handler consults the linked pull request's real state before accepting the transition.

**What is the proposed behavior?**
When a `PATCH /issues/:id` request would move an issue's status to `done`, the server looks up any linked GitHub pull requests (via the existing external-object mention system). If any linked pull request is not merged, or is merged but its CI checks are not green, the server rejects the request with HTTP `422` and a message naming the pull request (`owner/repo#number`) and the exact reason (`open`, `unmerged`, `checks red`, or `checks pending`).

**Why does it need to be addressed?**
A recent audit found an issue closed `done` while its linked pull request (#4198 in another Paperclip-managed repo) was still open with red CI — nothing on `main`. This is a silent false-positive: the board shows work as complete when it is not. There is currently no structural guard against it; only a human review catches it.

**Reason and benefit**
Today, closing an issue and merging its pull request are two independent, unenforced actions. This enhancement makes "the PR is merged and green" a hard precondition for "the issue is done," removing an entire class of false-`done` reports without adding any new command or concept — it strengthens an existing transition.

## What Changed

- `server/src/services/github-external-object-provider.ts`: the pull-request resolver now fetches check-run state for the PR head SHA (`GET /repos/{owner}/{repo}/commits/{sha}/check-runs`), falling back to the combined commit-status endpoint (`GET .../commits/{sha}/status`) when there are no check-runs. The result is exposed as `data.checksState: "success" | "failure" | "pending"`. When neither endpoint reports anything (no CI configured on that commit), it defaults to `"success"` so repos without CI are never blocked.
- `server/src/routes/issues.ts`: `PATCH /issues/:id` now calls a new `assertNoBlockingLinkedPullRequest` helper whenever the requested status is `done` and the issue is not already `done`. It reads the issue's linked pull requests via the existing `externalObjectsSvc.listForIssue`, and for each GitHub `pull_request` object, rejects the transition with HTTP `422` if it is not merged (reason `open` or `unmerged`) or is merged with `checksState !== "success"` (reason `checks red` or `checks pending`). The response body includes a structured `details: { code: "done_transition_pr_gate", reason, pullRequest }` payload alongside the human-readable message. There is no bypass flag — the check applies to both board and agent actors.
- Added `server/src/__tests__/issue-done-transition-pr-gate-routes.test.ts`: an embedded-Postgres, route-level test covering the open-PR refusal, the merged-with-red-CI refusal, the merged-with-pending-CI refusal, the merged-and-green pass case, and the no-linked-PR no-op case.
- Extended `server/src/__tests__/external-objects-service.test.ts` with resolver-level coverage for the new `checksState` field: check-run-derived pending/failure/success, the combined-status fallback when there are no check-runs, and the default-to-success behavior when neither endpoint reports anything.

## Verification

Route-level acceptance evidence, captured from the actual handler (embedded Postgres, stubbed GitHub `fetch`) using the same request shape as `curl -X PATCH .../issues/{id} -d '{"status":"done"}'`:

Open PR -> HTTP 422:
```json
{
  "error": "Cannot mark issue done: linked pull request acme/app#201 is open",
  "code": "done_transition_pr_gate",
  "details": { "code": "done_transition_pr_gate", "reason": "open", "pullRequest": { "owner": "acme", "repo": "app", "number": 201, "state": "open", "merged": false, "checksState": "success" } }
}
```

Merged PR with red CI -> HTTP 422:
```json
{
  "error": "Cannot mark issue done: linked pull request acme/app#202 is checks red",
  "code": "done_transition_pr_gate",
  "details": { "code": "done_transition_pr_gate", "reason": "checks red", "pullRequest": { "owner": "acme", "repo": "app", "number": 202, "state": "closed", "merged": true, "checksState": "failure" } }
}
```

Merged PR with green CI -> HTTP 200, issue reaches `done`:
```json
{ "id": "f92984a6-...", "status": "done", "changes": { "status": { "from": "in_progress", "to": "done" } } }
```

The merged-with-pending-CI case (same 422 shape, `reason: "checks pending"`) and the no-linked-PR no-op case (200, unaffected) are covered by the automated test suite below.

Commands run:
- `npx vitest run src/__tests__/issue-done-transition-pr-gate-routes.test.ts` — 5/5 passing
- `npx vitest run src/__tests__/external-objects-service.test.ts` — 31/31 passing
- `npx vitest run src/__tests__/external-object-routes.test.ts src/__tests__/issues-service.test.ts` — 128/128 passing (regression sweep on adjacent routes/services)
- `npx tsc --noEmit -p server` — clean on all touched files (pre-existing, unrelated plugin-sdk workspace-build errors only, present before this change)
- Searched open and merged PRs for `checksState`, `done_transition_pr_gate`, and "done transition" — no existing or duplicate PR found.

## Risks

- **Behavioral change:** issues linked to a GitHub pull request can no longer reach `done` until that PR is merged with green CI. This is the intended effect, but any workflow that relied on marking an issue `done` independently of its PR state will now be rejected with `422`.
- **No bypass:** the gate has no override flag by design (fail-closed). If a legitimate need to force-close such an issue arises, it requires unlinking the PR reference or a follow-up change — this is a deliberate tradeoff, not an oversight.
- **External dependency:** the check-run/status lookup depends on the GitHub API being reachable and returning data for the PR's head SHA. If GitHub is unreachable, the existing resolver's fetch failure handling applies (the affected linked-PR fetch simply fails and is not treated as a green signal); this does not change error handling paths already in place for other PR fields.
- **Scope:** this only gates the `pull_request` external-object type from the GitHub provider. Other providers or object types are unaffected. Low risk.

## Model Used

Claude (Anthropic), accessed via the pi coding agent harness. Extended-thinking / agentic tool-use mode, used to read the existing external-objects resolver and route handler, implement the check-run fetch and the `done`-transition gate, and run the full verification test suite above.

## Checklist

### DIFF
--- a/docs/api/issues.md
+++ b/docs/api/issues.md
@@ -338,3 +338,4 @@ backlog -> todo -> in_progress -> in_review -> done
 - `started_at` auto-set on `in_progress`
 - `completed_at` auto-set on `done`
 - Terminal states: `done`, `cancelled`
+- `done` is refused with `422` (`details.code: "done_transition_pr_gate"`) when the issue references a GitHub pull request that is not merged, or is merged with a non-`success` check-run/status state. The error names the pull request and the reason (`open`, `unmerged`, `checks red`, `checks pending`). There is no bypass; unlink the pull-request reference to force the transition.
--- a/server/src/__tests__/external-objects-service.test.ts
+++ b/server/src/__tests__/external-objects-service.test.ts
@@ -372,6 +372,56 @@ describe("GitHub external object provider", () => {
     });
   });
 
+  it("gates a merged PR's checksState on the merge commit, not the (green) pre-merge head SHA", async () => {
+    // Regression coverage for AGE-569 follow-up: a PR's head SHA only ever
+    // carries pre-merge CI. A merge-queue/canary gate that only runs against
+    // the merge commit on the base branch (e.g. canary-pr-control-plane) can
+    // fail on that landed commit even though the PR's own head SHA was green.
+    // The done-transition gate must not read this PR as "merged and green"
+    // in that case.
+    const fetch = vi.fn(async (url: string) => {
+      if (url.includes("/commits/mergecommitsha/check-runs")) {
+        return checkRunsResponse([{ status: "completed", conclusion: "failure" }]);
+      }
+      if (url.includes("/commits/mergecommitsha/status")) return response({ state: "success" });
+      if (url.includes("/commits/deadbeef/check-runs")) {
+        return checkRunsResponse([{ status: "completed", conclusion: "success" }]);
+      }
+      if (url.includes("/commits/deadbeef/status")) return response({ state: "success" });
+      return response({
+        state: "closed",
+        draft: false,
+        merged: true,
+        merged_at: "2026-08-20T12:00:00Z",
+        title: "Ship it",
+        head: { sha: "deadbeef" },
+        merge_commit_sha: "mergecommitsha",
+      });
+    });
+    const provider = createGitHubExternalObjectProvider({} as any, { fetch, tokenProvider: null });
+    const resolver = provider.resolvers.find((entry) => entry.objectType === "pull_request")!;
+
+    const result = await resolver.resolve({
+      companyId: "company-1",
+      object: githubObject("pull/42", "pull_request"),
+    });
+
+    expect(fetch).toHaveBeenCalledWith(
+      "https://api.github.com/repos/acme/app/commits/mergecommitsha/check-runs?per_page=100",
+      expect.any(Object),
+    );
+    expect(fetch).not.toHaveBeenCalledWith(
+      "https://api.github.com/repos/acme/app/commits/deadbeef/check-runs?per_page=100",
+      expect.any(Object),
+    );
+    expect(result).toEqual({
+      ok: true,
+      snapshot: expect.objectContaining({
+        data: expect.objectContaining({ merged: true, checksState: "failure" }),
+      }),
+    });
+  });
+
   it("falls back to the combined commit status endpoint when there are no check-runs", async () => {
     const fetch = vi.fn(async (url: string) => {
       if (url.includes("/check-runs")) return checkRunsResponse([]);
--- a/server/src/__tests__/external-objects-service.test.ts
+++ b/server/src/__tests__/external-objects-service.test.ts
@@ -507,6 +507,101 @@ describe("GitHub external object provider", () => {
     });
   });
 
+  it("does not report checksState success when the check-runs lookup fails but the legacy status succeeds", async () => {
+    const fetch = vi.fn(async (url: string) => {
+      if (url.endsWith("/check-runs")) return new Response("", { status: 502 });
+      if (url.endsWith("/status")) {
+        return response({
+          state: "success",
+          statuses: [{ state: "success", context: "ci/legacy-jenkins" }],
+        });
+      }
+      return response({
+        state: "closed",
+        merged: true,
+        draft: false,
+        title: "Ship it",
+        head: { sha: "deadbeef" },
+      });
+    });
+    const provider = createGitHubExternalObjectProvider({} as any, { fetch, tokenProvider: null });
+    const resolver = provider.resolvers.find((entry) => entry.objectType === "pull_request")!;
+
+    const result = await resolver.resolve({
+      companyId: "company-1",
+      object: githubObject("pull/42", "pull_request"),
+    });
+
+    // The legacy status endpoint reports green, but the check-runs lookup
+    // itself failed (502) -- we have no idea whether there is a failing
+    // required check-run we simply could not see. Reporting "success" here
+    // would let an unverified merged PR pass the done gate.
+    expect(result).toEqual({
+      ok: true,
+      snapshot: expect.objectContaining({
+        data: expect.not.objectContaining({ checksState: expect.anything() }),
+      }),
+    });
+  });
+
+  it("treats an unreadable 2xx check-runs body the same as a failed lookup, not an empty list", async () => {
+    const fetch = vi.fn(async (url: string) => {
+      if (url.endsWith("/check-runs")) return new Response("not json", { status: 200, headers: { "content-type": "application/json" } });
+      if (url.endsWith("/status")) return response({ total_count: 0, statuses: [] });
+      return response({
+        state: "closed",
+        merged: true,
+        draft: false,
+        title: "Ship it",
+        head: { sha: "deadbeef" },
+      });
+    });
+    const provider = createGitHubExternalObjectProvider({} as any, { fetch, tokenProvider: null });
+    const resolver = provider.resolvers.find((entry) => entry.objectType === "pull_request")!;
+
+    const result = await resolver.resolve({
+      companyId: "company-1",
+      object: githubObject("pull/42", "pull_request"),
+    });
+
+    expect(result).toEqual({
+      ok: true,
+      snapshot: expect.objectContaining({
+        data: expect.not.objectContaining({ checksState: expect.anything() }),
+      }),
+    });
+  });
+
+  it("treats an unrecognized completed check-run conclusion as failure, not success", async () => {
+    const fetch = vi.fn(async (url: string) => {
+      if (url.endsWith("/check-runs")) {
+        return checkRunsResponse([{ status: "completed", conclusion: "stale" }]);
+      }
+      if (url.endsWith("/status")) return response({ total_count: 0, statuses: [] });
+      return response({
+        state: "closed",
+        merged: true,
+        draft: false,
+        title: "Ship it",
+        head: { sha: "deadbeef" },
+      });
+    });
+    const provider = createGitHubExternalObjectProvider({} as any, { fetch, tokenProvider: null });
+    const resolver = provider.resolvers.find((entry) => entry.objectType === "pull_request")!;
+
+    const result = await resolver.resolve({
+      companyId: "company-1",
+      object: githubObject("pull/42", "pull_request"),
+    });
+
+    expect(result).toEqual({
+      ok: true,
+      snapshot: expect.objectContaining({
+        data: expect.objectContaining({ checksState: "failure" }),
+      }),
+    });
+  });
+
   it.each([
     [
       "auth-required",
--- a/server/src/__tests__/external-objects-service.test.ts
+++ b/server/src/__tests__/external-objects-service.test.ts
@@ -375,7 +375,12 @@ describe("GitHub external object provider", () => {
   it("falls back to the combined commit status endpoint when there are no check-runs", async () => {
     const fetch = vi.fn(async (url: string) => {
       if (url.endsWith("/check-runs")) return checkRunsResponse([]);
-      if (url.endsWith("/status")) return response({ state: "failure" });
+      if (url.endsWith("/status")) {
+        return response({
+          state: "failure",
+          statuses: [{ state: "failure", context: "ci/legacy-jenkins" }],
+        });
+      }
       return response({
         state: "open",
         draft: false,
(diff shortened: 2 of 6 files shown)


dzianisv:
Addressed all three in 52b8a5c:

**Issue 1 (stale snapshot):** `refreshIssueObjects()` now accepts `force: true`, and the done-transition gate force-refreshes the linked PR's live GitHub state (bypassing the object's TTL/backoff) before evaluating it, re-reading via `listForIssue` afterward. Falls back to the last-known snapshot only if the refresh call itself throws. Added a regression test (`refuses done when the cached PR snapshot is stale and CI has since gone red`) that resolves a PR as green, then flips the fixture to failing CI *without* calling `/external-objects/refresh` again, and asserts the `done` PATCH still catches it.

**Issue 2 (fetch failures -> green):** `fetchChecksState` now returns `null` (not `"success"`) when either the check-runs or combined-status request genuinely fails (non-2xx or network error). `"success"` is only returned when both requests succeed and report no data (a real no-CI commit). At the gate level, `doneTransitionPrGateReason` now treats a merged PR with a missing/unknown `checksState` as blocking (`"checks pending"`), never as a pass. Added `does not report checksState success when both the check-runs and status lookups fail`.

**Issue 3 (check-runs masking a failing legacy status):** removed the early-return; both endpoints are now always fetched and the worse of the two signals wins (`worstChecksState`). Also tightened `checksStateFromCombinedStatus` to require a non-empty `statuses` array before trusting the bare `state` field, since GitHub reports `state: "pending"` with an empty array when no legacy status has ever been posted (that's "no signal", not a real pending check) -- fixed the existing test that had been asserting on that ambiguous shape. Added `reports the worse of check-runs and a failing legacy combined status`.

All prior tests plus 3 new ones green (see updated PR description test plan). Re-triggering.

dzianisv:
Addressed both in e9078b2:

**Issue 1 (failed refresh trusts stale state):** `refreshIssueObjects()`'s per-object results are now inspected -- only `reason === "resolved"` counts as confirmed-fresh. Any linked PR whose forced refresh returned `auth_required`/`unreachable`/`no_resolver` (or whose refresh call threw) is now tracked as unconfirmed, and the gate blocks it with a new `reason: "unverified"` regardless of what its last-known cached data said. Added `refuses done when the forced refresh of a cached merged+green PR cannot reach GitHub` -- resolves a PR as green, then makes every subsequent fetch throw (simulating GitHub going unreachable) before the `done` PATCH, and asserts it's still rejected.

**Issue 2 (partial CI failures / unreadable bodies -> green):** `fetchChecksState` now tracks failure separately per endpoint and only returns `"success"` when both endpoints were actually reached and reported no data. A 2xx response with an unparseable body is now treated as a failed lookup, not an empty check-runs list. If either endpoint failed, `"success"` is never reported (a definitive `"failure"` from the other endpoint is still trusted, since that's unambiguous bad news either way). Also flipped `checksStateFromCheckRuns` from a block-list of failing conclusions to an allow-list of passing ones (`success`/`skipped`/`neutral`), so an unrecognized or future completed conclusion (e.g. `stale`) is treated as failure instead of silently passing. Added 3 tests: check-runs-502-with-status-green stays unknown, unreadable-2xx-body isn't treated as empty, and an unrecognized conclusion (`stale`) reports failure.

51 tests in the touched files, all green; full regression sweep (340 tests across issues-service/activity-events/comment-reopen/agent-mutation-ownership/execution-policy/dependency-wakeups/feedback/closed-workspace routes) unaffected. Re-triggering.

dzianisv:
Addressed both in a9e4e67:

**Issue 1 (initial lookup failure bypasses gate):** added a cheap pre-check (`textMightReferenceGitHubPullRequest`) against the issue's title/description before deciding how to handle a `listForIssue` failure. If the text doesn't look like it references a `github.com/.../.../pull/N` URL, an external-objects outage still can't 500 every unrelated `done` transition company-wide (unchanged prior behavior, still covered by the existing "no linked pull request" test and the module-mocked `issue-activity-events-routes.test.ts` case). If it *does* look like a PR reference and we still can't resolve it, the transition is now blocked with `reason: "unverified"` -- same shape as the already-tested refresh-failure case.

I did not add a full DB-fault-injection integration test for this exact branch: reliably forcing `listForIssue` itself to throw inside the real embedded-Postgres route test (as opposed to the module-mocked unit test file where this was originally caught) needs either an injectable `externalObjectsSvc` seam on `issueRoutes()` (doesn't exist today) or breaking the DB out from under a live request, both disproportionate for a 5-line conditional that shares its reason/error shape with the already-tested "unverified" refresh-failure path. Happy to add the seam as a follow-up if you'd rather have it covered end-to-end.

**Issue 2 (check-run pagination hides failures):** now requests `per_page=100`, and additionally compares `total_count` to the returned array length -- an all-green result from a *partial* page is now treated as unknown (not confirmed success), since a failing/pending run could be on a page we didn't fetch. A failure already visible on the first page is still trusted regardless of total_count (that signal doesn't need the rest of the pages). Added two tests: a paginated all-green partial page reports unknown, and a paginated page with a visible failure still reports failure.

53 tests in the touched files green; full regression sweep (330+ tests across issues-service/comment-reopen/agent-mutation-ownership/execution-policy/dependency-wakeups/feedback/closed-workspace routes, plus the original activity-events file) unaffected. Re-triggering.

dzianisv:
Addressed in bfd4044:

Added `issueHasRecordedPullRequestMention()`, a direct query against `external_object_mentions` (which records every source kind -- title, description, comment, document, not just the issue body). It's used as a fallback when the title/description text scan finds nothing, before deciding how to treat a `listForIssue` failure. If that secondary check itself fails, it falls back to the text-only signal (so a double-fault doesn't turn into a company-wide block of all `done` transitions).

Added a route-level regression test (`gates done when the pull request is referenced only from a comment, not the description`) that posts a PR link only via a follow-up comment, then confirms the `done` PATCH is still gated -- proving the main iteration loop (which already treats every linked external object the same regardless of source kind) covers comment-only mentions, and that the new mention-table fallback would catch it even if `listForIssue` itself failed.

64 tests in the touched files green; full regression sweep (330 tests) unaffected. Re-triggering.

dzianisv:
Addressed in 8349b22:

`assertNoBlockingLinkedPullRequest` now also extracts PR references directly from the request's own `updateFields.title`/`updateFields.description` (regex, no DB round-trip needed) and blocks (`reason: "unverified"`) any that aren't already among the resolved/known linked PRs from `listForIssue` -- regardless of whether that call succeeded or threw. This closes the gap where a single PATCH both introduces a new PR URL and requests `done` in the same request: `listForIssue` only reflects mentions synced from the previously-persisted text, so the new reference wouldn't have been in its result yet.

Added a route-level regression test that creates an issue with no PR reference, then PATCHes `description` (adding a fresh PR URL) and `status: "done"` together in one request, and confirms it's rejected with the new reference named in the error and the issue is *not* persisted as `done` (verified with a follow-up GET).

64+ tests in the touched files green (9 in the route-gate file specifically); full regression sweep now at 395 tests across issues-service/activity-events/comment-reopen/agent-mutation-ownership/execution-policy/dependency-wakeups/feedback/closed-workspace/external-object routes, unaffected. Re-triggering.

dzianisv:
Addressed in 29ae129:

`assertNoBlockingLinkedPullRequest` now, when title/description are being changed in this same PATCH, skips gating on a linked PR whose mentions come *only* from title/description sources (never a comment or document) and whose identity is absent from the proposed new (effective) text -- it's being removed by this very request, not something to verify. A PR still linked via a comment or document is gated regardless of what happens to the description text, since removing a mention from one place doesn't remove the actual dependency.

Two new route-level tests: removing an issue's only (open) PR reference in the same `done` PATCH now succeeds (with a sanity check confirming the same request without the removal still gates), and removing it from the description while it's still linked via a comment still gates.

11 tests in the route-gate file, 57 across the touched files, full regression sweep now at 397 tests -- all green. Re-triggering.

dzianisv:
Addressed in b376ff94:

Fixed the "Fresh cache bypasses live refresh" P1 (line 477) — the prior fix (033873cd) trusted any TTL-"fresh" object row when a concurrent refresh held the lease, but TTL freshness alone doesn't prove the row reflects state *as of this PATCH*; it can be up to the full refresh window old and predate a real GitHub state change (e.g. CI turning red) the lease holder is still capturing.

`assertNoBlockingLinkedPullRequest` now retries the forced refresh with a bounded budget (5 attempts, 120ms apart) and only confirms an object when either this call's own attempt resolved it, or the row's `lastResolvedAt` is at/after the moment this gate started — proof some resolve (this call's or a concurrent one) captured current state after the PATCH began, not stale leftover data. If the lease is still held once the budget is exhausted, it fails closed as unverified, unchanged from before.

The other 6 P1s in the latest review (API-failures-become-green, partial-CI-failures-become-green, unknown-status-becomes-green, initial-lookup-failure-bypasses-gate, comment-only-PR-links-bypass-gate, removed-PR-reference-still-blocks, failed-refresh-trusts-stale-state) were checked against the current head and are stale — they describe behavior from earlier commits in this PR's history (bfd4044, 8349b228, 29ae129f already fixed each of these) and do not match what's on disk now. Re-read the code at those line numbers to confirm before re-flagging.

Verification:
- `npx vitest run src/__tests__/issue-done-transition-pr-gate-routes.test.ts` — 13/13 passing, 5/5 repeated runs with zero flakes (replaced the prior test's real-timer race, which I observed both pass and fail across repeated local runs, with two deterministic tests of the same boundary condition)
- `npx vitest run src/__tests__/external-objects-service.test.ts src/__tests__/external-object-routes.test.ts src/__tests__/issues-service.test.ts` — 167/167 passing (regression sweep)
- `npx tsc --noEmit -p server` — clean on touched files

dzianisv:
Addressed in edee8266, in response to the latest review round (3 new P1s posted 2026-08-21T21:43:22Z):

**1. Fallback lookup failure bypasses gate (issues.ts:412-429):** if both `listForIssue` and the recorded-mention fallback (`issueHasRecordedPullRequestMention`) fail, and there's no positive text signal either, the gate now fails closed as `unverified` instead of silently allowing the transition. A double DB fault can no longer hide a comment/document-only PR reference.

**2. Pre-gate refresh passes as current (issues.ts:489-491):** removed the `result.reason === "resolved"` special case from the forced-refresh confirmation loop. When this gate's refresh call joins an already in-flight promise from a concurrent request (same process), that promise can persist `lastResolvedAt` using a `now` timestamp captured by the *original* caller before this gate began, even though it settles afterwards -- `reason === "resolved"` alone doesn't prove freshness as of this gate. Every result now requires the `lastResolvedAt >= gateRefreshStartedAt` timestamp check unconditionally, no exceptions.

**3. Merge SHA drops head CI (github-external-object-provider.ts:599-601):** `fetchChecksState` now checks both a merged PR's `merge_commit_sha` and its pre-merge `head` SHA, combining the worse of the two, rather than trusting the merge-commit lookup alone. Some CI setups never run checks against the merge commit at all, which previously made an all-quiet merge-commit lookup read as "success" and hide a failing/pending head-SHA CI run.

Added a new regression test (`gates a merged PR on the head SHA's CI when the merge commit has no CI configured at all`) and updated the existing merge-commit-priority test to reflect that both SHAs are now legitimately fetched.

Verification:
- `npx vitest run src/__tests__/issue-done-transition-pr-gate-routes.test.ts src/__tests__/external-objects-service.test.ts src/__tests__/external-object-routes.test.ts` -- 61/61 green
- `npx vitest run src/__tests__/issues-service.test.ts` (regression sweep) -- 181/181 green across the four suites combined
- `npx tsc --noEmit -p server` -- clean on all three touched files (pre-existing zod-version errors elsewhere in the codebase, unrelated to this branch)

Re-triggering CI.

PR #11777, opened 2026-08-20
fix(adapter-utils): confine sandbox sync host paths with platform rules

jms-ai-solutions:
## Thinking Path

> - Sandbox environments let an agent run in an isolated computer instead of on the host machine
> - A sandbox run stages the workspace on the host, then syncs it into the sandbox
> - `assertSyncOperationsConfined` guards that sync. It checks both sides of every mapping with `path.posix`
> - Sandbox paths are always POSIX, but host paths follow the host platform. On Windows they are Win32
> - So the guard rejects Paperclip's own staging tar on a Windows host, and every sandbox run stops
> - This pull request gives each caller a way to declare the flavour of each side
> - The benefit is that sandbox environments work on a Windows host, and the confinement rules stay the same

**What happened?**

Every agent run assigned to a `driver: "sandbox"` environment fails on a Windows host. The run never reaches the provider. The error is:

```
sync operation source path is not a confined absolute path:
C:\Users\<user>\AppData\Local\Temp\paperclip-sandbox-sync-Tx5GSy\workspace.tar
```

`assertSyncOperationsConfined` in `packages/adapter-utils/src/sandbox-managed-runtime.ts` validates each mapping with `path.posix`:

```ts
const normalized = path.posix.normalize(candidate);
if (!path.posix.isAbsolute(normalized) || ...) {
  throw new Error(`sync operation ${label} path is not a confined absolute path: ${candidate}`);
}
```

The docblock states the assumption: "Sandbox and host paths on the server are POSIX." That is true for the sandbox side. It is not true for the host side on Windows. `path.posix.isAbsolute("C:\\Users\\...")` returns false, so the guard throws fail-closed.

**Expected behavior**

A sandbox run stages its workspace and starts, on any host platform that Paperclip supports.

**Steps to reproduce**

1. Run Paperclip on a Windows host.
2. Install a plugin that registers a `sandbox_provider` environment driver.
3. Create an environment with `driver: "sandbox"` and that provider.
4. Assign an agent to that environment and start a run.
5. The run fails with the error above. The provider's hooks are never called.

`paperclipai` v2026.817.0 (managed CLI), reproduced against `master` source.

**Deployment mode**

Local, `local_trusted`, single instance.

**Agent adapter(s) involved**

Not adapter-specific (core bug). Observed with Claude Code; the guard runs before adapter selection matters.

**Operating system**

Windows 11 Pro 26200. Node 22.

**Additional context**

This is provider-agnostic. The guard runs before `onEnvironmentSyncIn` / `onEnvironmentSyncOut`, so a provider cannot work around it. I first implemented native `docker cp` sync hooks in my own provider specifically to bypass the default base64-tar transport; the failure did not change, because execution never reaches the hooks. Both transports stop at the same line.

**Related work (dedup search)**

I searched the PR list for `sandbox sync windows`, `path.posix host path`, and `assertSyncOperationsConfined`. No open or closed PR addresses host-path confinement flavours. Related, already merged, same subsystem: Refs #10013 (opt-in sandbox file-sync lifecycle hooks), Refs #11749 (route sandbox git-bundle export through native syncOut).

**Privacy checklist**

Paths in this description are redacted or generic. No tokens, keys, or private hostnames appear.

## What Changed

- `assertSyncOperationsConfined` takes a third parameter that declares the path flavour of each side: `{ source, target }`, each `"posix"`, `"host"`, or `"win32"`. Callers use `"host"`; `"win32"` pins the Win32 grammar so tests exercise it on every runner platform.
- The parameter defaults to POSIX on both sides. Existing callers and their tests do not change.
- Confinement now runs with the matching path grammar: `path.posix` for sandbox paths, `path` (platform-native) for host paths.
- The three call sites declare their direction. Inbound staging uses a host source and a sandbox target. The two outbound paths (git bundle export, workspace restore) use a sandbox source and a host target.
- Comparison folds case whenever the grammar in use is Win32 (`p.sep === "\\"`), because Windows paths are case-insensitive and a case-sensitive compare rejects a valid root. POSIX comparison stays case-sensitive.
- Traversal detection splits on both separators, because Win32 accepts either.
- Five new tests in `sandbox-managed-runtime.test.ts`, including a platform-independent Win32 suite.

## Verification

Run the package tests:

```
pnpm vitest run packages/adapter-utils/src/sandbox-managed-runtime.test.ts
```

New coverage:

- The Windows staging path that fails today is accepted when the caller declares a host source.
- A host source outside its declared root is still rejected.
- `..` traversal on the host side is still rejected.
- `..` traversal on the sandbox side of an outbound operation is still rejected.
- With no flavours given, both sides stay POSIX and behave exactly as before.

The `win32` flavour makes Win32 semantics testable on any runner: the Win32 suite covers the exact staging path that failed, forward-slash Win32 paths, case-insensitive root matching, an escape, traversal with either separator, a drive-relative path, and a POSIX control that proves case folding stays off for POSIX paths. It runs identically on POSIX and Windows CI runners.

I also confirmed the behaviour directly against the live failure on a Windows host: the same call that produced the error above is accepted after the change, and each rejection case above still throws.

## Risks

Low risk.

- The new parameter is optional and defaults to the current behaviour, so any caller that does not pass it is unchanged.
- No confinement rule is removed. The absolute-path requirement, `..` rejection, and root containment all still apply. Only the grammar used to evaluate them changes, and only for the side a caller declares as a host path.
- Case folding applies only to host paths on a Windows host. POSIX comparison stays case-sensitive.
- Non-Windows hosts are unaffected: `path` is `path.posix` there, so the host flavour resolves to the same rules as today.

## Model Used

Claude Fable 5 (`claude-fable-5`), via Claude Code, with extended thinking and tool use. The change was authored and verified with a human reviewing each step.

## Checklist

### DIFF
--- a/packages/adapter-utils/src/sandbox-managed-runtime.test.ts
+++ b/packages/adapter-utils/src/sandbox-managed-runtime.test.ts
@@ -1911,6 +1911,89 @@ describe("sandbox managed runtime", () => {
     ).toThrow(/escapes its confinement root|not a confined absolute path/);
   });
 
+  // The `win32` flavour pins the Win32 grammar regardless of the runner
+  // platform, so CI on a POSIX runner still exercises Windows semantics:
+  // drive-letter absolutes, either separator, traversal, and case folding.
+  it("confines Win32 host paths identically on every runner platform", () => {
+    const hostRoot = "C:\\Users\\example\\AppData\\Local\\Temp\\paperclip-sandbox-sync-Tx5GSy";
+    const sandboxRoot = "/opt/paperclip/runtime";
+
+    // The exact staging path that failed under POSIX-only confinement.
+    expect(() =>
+      assertSyncOperationsConfined(
+        [{
+          operationId: "inbound",
+          files: [{ sourcePath: `${hostRoot}\\workspace.tar`, targetPath: `${sandboxRoot}/workspace.tar`, kind: "file" as const }],
+        }],
+        { sourceRoots: [hostRoot], targetRoots: [sandboxRoot] },
+        { source: "win32", target: "posix" },
+      ),
+    ).not.toThrow();
+
+    // Win32 accepts forward slashes too.
+    expect(() =>
+      assertSyncOperationsConfined(
+        [{
+          operationId: "inbound",
+          files: [{ sourcePath: "C:/Users/example/AppData/Local/Temp/paperclip-sandbox-sync-Tx5GSy/workspace.tar", targetPath: `${sandboxRoot}/workspace.tar`, kind: "file" as const }],
+        }],
+        { sourceRoots: [hostRoot], targetRoots: [sandboxRoot] },
+        { source: "win32", target: "posix" },
+      ),
+    ).not.toThrow();
+
+    // Windows paths are case-insensitive: a root that differs only in case matches.
+    expect(() =>
+      assertSyncOperationsConfined(
+        [{
+          operationId: "inbound",
+          files: [{ sourcePath: "c:\\users\\EXAMPLE\\appdata\\local\\temp\\PAPERCLIP-SANDBOX-SYNC-tx5gsy\\workspace.tar", targetPath: `${sandboxRoot}/workspace.tar`, kind: "file" as const }],
+        }],
+        { sourceRoots: [hostRoot], targetRoots: [sandboxRoot] },
+        { source: "win32", target: "posix" },
+      ),
+    ).not.toThrow();
+
+    // A Win32 path outside its declared root is rejected.
+    expect(() =>
+      assertSyncOperationsConfined(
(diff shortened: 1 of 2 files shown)


jms-ai-solutions:
Both review items addressed in 130d38d:

- Win32 CI coverage: `SyncPathFlavour` gains an explicit `win32` value that pins `path.win32` regardless of the runner platform, and case folding now follows the grammar in use (`p.sep`) rather than the server platform. A new platform-independent Win32 suite covers the failing staging path, forward-slash Win32 paths, case-insensitive root matching, an escape, traversal with either separator, a drive-relative path, and a POSIX control proving case folding stays off for POSIX. It runs identically on POSIX and Windows runners.
- PR description now carries the full checklist including the dedup search, with related merged PRs #10013 and #11749 linked.

PR #11778, opened 2026-08-20
fix: preserve assignee during stale execution recovery

tommypoltev:
## Summary
- stop stale execution lock adoption from rewriting the issue assignee during recovery checkout
- keep recovery adoption scoped to checkout and execution lock repair only
- add a regression test covering stale execution lock recovery while preserving the standing assignee

## Testing
- pnpm vitest server/src/__tests__/issues-service.test.ts

## Dedup
- [x] Checked for an existing fix or regression in current recovery checkout paths before changing behavior


tommypoltev:
Superseded by #11779. This older PR points at stale head de1b00e3 and should not be merged.

PR #11779, opened 2026-08-20
fix(recovery): preserve explicit recovery owners

tommypoltev:
## Summary
- restore explicit `recoveryOwnerAgentId` routing as an opt-in for recovery ownership while keeping the source issue assignee unchanged
- keep successful-run handoff recovery out of plugin-managed issue lifecycles
- add a regression test that plugin-managed missing-disposition cases stay on the source issue instead of escalating into generic recovery

## Issue Description
A terminal-run recovery path could still lose the intended recovery owner override after earlier assignee-preservation changes, and the same branch had accidentally removed the plugin-managed successful-handoff guard. The required behavior for this scope is:
- recovery must not rewrite `assigneeAgentId` on the source issue
- explicit recovery ownership stays opt-in and only affects the recovery action / recovery issue owner
- plugin-managed lifecycles stay out of generic successful-run handoff escalation

## What Changed
- thread `recoveryOwnerAgentId` back through `recovery/service.ts` as an explicit owner preference instead of using recovery-run ownership to rewrite the source assignee
- restore `isPluginManagedIssueLifecycle` gating before missing-disposition escalation
- cover the plugin-managed skip path in `heartbeat-process-recovery.test.ts`

## Verification
- `pnpm -C paperclip-ofm-4274 --filter @paperclipai/server exec vitest run src/__tests__/issue-recovery-actions.test.ts src/__tests__/heartbeat-process-recovery.test.ts -t "keeps the standing assignee|blocks failed execution-review recovery under the reviewer when the source assignee differs|skips missing-disposition escalation for plugin-managed issue lifecycles"`
- Result on August 20, 2026: Vitest starts, but both suites abort during import on existing unrelated failure `TypeError: setExpensiveWorkspaceGitExecutor is not a function` from `src/services/workspace-git-operation-scheduler.ts` before the targeted recovery tests execute.

## Risks
- Low. Scope is limited to recovery owner routing and the successful-handoff plugin-managed guard.
- The source issue assignee remains the source of truth; only recovery ownership selection regains the explicit override.

## Checklist
- [x] I have included a thinking path that traces from project context to this change
- [x] I have specified the model used (with version and capability details)
- [x] I have checked ROADMAP.md and confirmed this PR does not duplicate planned core work
- [x] I have searched GitHub for duplicate or related PRs and linked them above
- [x] I have either (a) linked existing issues with `Fixes: #` / `Closes #` / `Refs #` OR (b) described the issue in-PR following the relevant issue template
[...]

PR #11794, opened 2026-08-20
fix(recovery): preserve source assignee during stranded fallback

tommypoltev:
## Thinking Path

> - Paperclip is the open source app people use to manage AI agents for work.
> - This change stays inside the server stranded-recovery path for assigned issues.
> - OFM-4569 is specifically about helper recovery runs stealing the standing assignee from the source issue.
> - The source assignee must remain the owner of record for that helper-recovery path.
> - Review-stage recovery is a different contract: when the recovery owner is the current execution participant, the issue must still stay with that participant.
> - This PR therefore preserves the standing assignee only for non-review takeover fallback and leaves the review-participant contract intact.
> - The result is a smaller review surface that matches the issue and the upstream review-stage semantics.

## Linked Issues or Issue Description

Refs #11779
Refs #11782

## What Changed

- Keep `issue.assigneeAgentId` as the source of truth when stranded helper recovery evaluates a run from another agent.
- Exempt the active execution-stage participant from that preservation rule, so review-stage recovery keeps the issue on the current participant.
- Add a regression test for assignee `A`, helper run by agent `B`, and verify the source issue stays assigned to `A`.
- Keep the change scoped to recovery routing and assignee preservation only.

## Verification

- `pnpm vitest run server/src/__tests__/issue-recovery-actions.test.ts --reporter=verbose`
  Result on August 20, 2026: file loads cleanly; embedded Postgres suites are skipped on this host because the probe fails with `listen EPERM: operation not permitted 127.0.0.1` (`1 passed | 45 skipped`).
- `pnpm --filter @paperclipai/server typecheck`
  Result on August 20, 2026: fails on pre-existing unrelated `@paperclipai/plugin-sdk` resolution / plugin typing errors outside this diff.

## Risks

- Low risk. The diff is limited to stranded recovery owner resolution in `server/src/services/recovery/service.ts` and the focused recovery regression coverage.
- Main reviewer check is that helper recovery no longer steals the standing assignee, while execution-review participant recovery still keeps the issue on the participant.

## Model Used

- OpenAI Codex, GPT-5.4, tool-assisted coding in Paperclip runtime with shell and patch tools.

## Checklist

- [x] I have included a thinking path that traces from project context to this change
- [x] I have specified the model used (with version and capability details)
- [x] I have checked ROADMAP.md and confirmed this PR does not duplicate planned core work
- [x] I have searched GitHub for duplicate or related PRs and linked them above
[...]

PR #11796, opened 2026-08-20
fix(recovery): preserve source assignee on recovery

tommypoltev:
## Summary
- preserve the standing source assignee during stranded-issue recovery instead of inheriting the recovery runner
- keep stale execution-lock adoption from rewriting `assigneeAgentId`
- trim the previous takeover/recovery-issue detour so the fix stays inside OFM-4569 scope

## Issue
- Refs OFM-4569
- Context: terminal-run recovery / auto-unstick could reassign a permanently `in_progress` issue from its standing owner to the recovery agent

## Thinking Path
1. Reproduced the two places where recovery ownership leaked into issue ownership: `recovery/service.ts` routing from `latestRun.agentId` and stale execution-lock adoption in `issues.ts`.
2. Reduced the branch back to the minimal behavior required by the issue: recovery may repair wake/lock state, but must not mutate the source issue assignee by default.
3. Kept only direct regressions for those paths and removed prior out-of-scope takeover/recovery-issue changes that were causing review churn.

## What Changed
- `server/src/services/recovery/service.ts`
  - prefer `issue.assigneeAgentId` over `latestRun.agentId` when choosing the "original" recovery owner
  - preserve `input.issue.assigneeAgentId` when escalating the source issue to `blocked`
  - preserve `current.assigneeAgentId` when adopting a stale `executionRunId`
- tests
  - keep the regression that recovery initiated by agent B leaves assignee A unchanged
  - keep stale-lock regressions for assigned and unowned issues
  - drop unrelated plugin-lifecycle / takeover assertions added in the previous round

## Verification
- `pnpm vitest run server/src/__tests__/issue-recovery-actions.test.ts -t "keeps the current source assignee as the recovery return owner when the latest run came from another agent"`
- `pnpm vitest run server/src/__tests__/issues-service.test.ts -t "checkout adoption of a stale executionRunId"`
- Result on August 20, 2026: both suites abort during import on existing unrelated bootstrap failure `TypeError: setExpensiveWorkspaceGitExecutor is not a function` from `src/services/workspace-git-operation-scheduler.ts:876` before the targeted tests execute.

## Risks
- Low. Scope is limited to assignee preservation in recovery and stale execution-lock adoption.
- Recovery ownership for follow-up actions still works; only implicit source-issue reassignment is removed.

## Model Used
- Codex (GPT-5 family) in Paperclip code-agent mode.

## Checklist
- [x] Thinking path included
- [x] Model used included
- [x] Issue context included in-PR
- [x] Added or updated regression tests
- [x] Dedup checked against the prior closed PR on the same branch (`#11779`)
- [ ] Local targeted tests fully pass
- [ ] CI green

### DIFF
--- a/server/src/__tests__/heartbeat-process-recovery.test.ts
+++ b/server/src/__tests__/heartbeat-process-recovery.test.ts
@@ -3913,6 +3913,64 @@ describeEmbeddedPostgres("heartbeat orphaned process recovery", () => {
     expect(recoveryIssues).toHaveLength(0);
   });
 
+  it("skips missing-disposition escalation for plugin-managed issue lifecycles", async () => {
+    const { companyId, agentId, runId, issueId } = await seedStrandedIssueFixture({
+      status: "in_progress",
+      runStatus: "failed",
+      runErrorCode: "adapter_failed",
+    });
+    const sourceRunId = randomUUID();
+    await db
+      .update(issues)
+      .set({ originKind: "plugin:paperclip.workflow-engine" })
+      .where(eq(issues.id, issueId));
+    await db
+      .update(heartbeatRuns)
+      .set({
+        contextSnapshot: {
+          issueId,
+          taskId: issueId,
+          wakeReason: "finish_successful_run_handoff",
+          sourceRunId,
+          resumeFromRunId: sourceRunId,
+          handoffRequired: true,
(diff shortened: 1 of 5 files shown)


tommypoltev:
Closing in favor of #11794, which carries the reduced OFM-4569 scope and current review thread.

PR #11807, opened 2026-08-20
fix(recovery): preserve assignee across stranded recovery fallback

tommypoltev:
## Summary
- keep the source issue assignee as owner-of-record when helper recovery runs fail under a different agent
- create separate takeover recovery issues when fallback ownership moves to a manager, instead of overwriting the source issue assignee
- exempt pending execution-review participants from this assignee-preservation path so in-review recovery still targets the participant route

## Testing
- pnpm vitest server/src/__tests__/issue-recovery-actions.test.ts server/src/__tests__/issues-service.test.ts
- pnpm vitest server/src/__tests__/issue-recovery-actions.test.ts server/src/services/recovery/review-path-recovery.test.ts

## Issue
- OFM-4569

### DIFF
--- a/cli/src/commands/db-backup.ts
+++ b/cli/src/commands/db-backup.ts
@@ -73,6 +73,7 @@ export async function dbBackupCommand(opts: DbBackupOptions): Promise<void> {
     const result = await runDatabaseBackup({
       connectionString: connection.value,
       backupDir,
+      maxBackups: 7,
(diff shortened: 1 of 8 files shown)


tommypoltev:
Superseded by #11817 (cb0009b09, merged 2026-08-20 22:09 UTC), which fixes the root cause on master with an explicit invariant (server/src/services/recovery/service.ts:4374 and :4514-4531). This branch is stale relative to that commit and its delta would regress master. Closing as duplicate. Ref OFM-4569.

PR #11808, opened 2026-08-20
KEWL-3769: harden run_id FK safety in activity-log and routines

jonguttman:
## Summary
- `logActivity`/`persistActivity` and `appendRoutineRevision` (+ the document revision it writes via `upsertRoutineDescriptionDocument`) wrote a caller-supplied `runId` straight into FK-bound `*RunId` columns with no existence check against `heartbeat_runs`, crashing on an unknown/stale run id (e.g. a manual `X-Paperclip-Run-Id` header, or a legitimate race before the `heartbeat_runs` insert lands).
- Adds `resolveKnownRunId(db, companyId, runId)` in both `server/src/services/activity-log.ts` and `server/src/services/routines.ts`: verifies the run exists before use, falls back to `null` with a `logger.warn` on an unknown id, never throws.
- Only the optional run-provenance attribution degrades — agent/user attribution and the FK constraint itself are untouched.

This ports the pattern validated in KEWL-3768 (54/54 tests passing there) onto current `origin/master`, whose `activity-log.ts`/`routines.ts` have since diverged substantially (persistActivity/publishActivity split, postCommitPublications batching, ~1500 lines of unrelated `routines-service.test.ts` growth) — the original diff didn't apply cleanly, so this is a fresh, re-verified implementation against current master.

## Test plan
- [x] `npx tsc --noEmit` in `server/` — 0 errors
- [x] New `server/src/__tests__/activity-log-run-provenance.test.ts` (2 tests, embedded Postgres): unknown runId → `null` without throwing; known runId preserved
- [x] New case in `server/src/__tests__/routines-service.test.ts`: routine creation with an unknown `actor.runId` succeeds, `routineRevisions.createdByRunId` and `documentRevisions.createdByRunId` both resolve to `null`
- [x] Full existing `activity-log-responsible-user.test.ts` (9 tests) and `routines-service.test.ts` (64 tests) suites pass
- [x] Broader regression pass: `routines-routes`, `routines-e2e`, `issue-activity-events-routes`, `routine-run-telemetry`, `routine-document-annotation-routes`, `plugin-managed-routines`, `qa-routine-secrets-e2e` (43 tests) — all pass, no regressions

### DIFF
--- a/server/src/__tests__/activity-log-run-provenance.test.ts
+++ b/server/src/__tests__/activity-log-run-provenance.test.ts
@@ -1,0 +1,113 @@
+import { randomUUID } from "node:crypto";
+import { afterAll, beforeAll, describe, expect, it } from "vitest";
+import { eq } from "drizzle-orm";
+import {
+  activityLog,
+  agents,
+  companies,
+  createDb,
+  heartbeatRuns,
+} from "@paperclipai/db";
+import { logActivity } from "../services/activity-log.js";
+import {
+  getEmbeddedPostgresTestSupport,
+  startEmbeddedPostgresTestDatabase,
+} from "./helpers/embedded-postgres.js";
+
+const embeddedPostgresSupport = await getEmbeddedPostgresTestSupport();
+const describeEmbeddedPostgres = embeddedPostgresSupport.supported ? describe : describe.skip;
+
+describeEmbeddedPostgres("logActivity run_id provenance (KEWL-3766 / KEWL-3768 / KEWL-3769)", () => {
+  let db!: ReturnType<typeof createDb>;
+  let tempDb: Awaited<ReturnType<typeof startEmbeddedPostgresTestDatabase>> | null = null;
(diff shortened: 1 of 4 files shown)


jonguttman:
Closing as a duplicate — PR #11806 (`fix/run-id-fk-safety`) implements the same fix (company-scoped `resolveKnownRunId`/`resolveKnownActivityRunId`/`resolveKnownRoutineRunId` guard on `activity_log.run_id` and `routineRevisions`/`documentRevisions.createdByRunId`, same two files, equivalent test coverage) and was opened ~18 minutes earlier from an independent run of the same underlying ticket. #11806 has already passed the PR-template gate and Greptile review (5/5, no issues); its one red check is an unrelated flaky test in `workspace-runtime.test.ts` (FK violation in `workspace_runtime_services` from a stale test fixture, nothing to do with run-provenance). Continuing review there instead of here to avoid two competing PRs for the same change.

PR #11809, opened 2026-08-20
fix: preserve assignee ownership during recovery

tommypoltev:
## Summary
- keep the source issue assignee as owner-of-record when helper recovery runs under another agent
- preserve the current source assignee when recovery re-blocks after reassignment, except for the active review participant path
- add regression coverage for cross-agent helper recovery and stale re-block reassignment

## Testing
- `pnpm exec vitest run --config vitest.config.ts src/__tests__/issue-recovery-actions.test.ts src/__tests__/heartbeat-process-recovery.test.ts` (from `server/`): `1 passed`, `1 skipped` on this host; the skipped suite is the embedded-Postgres recovery suite

## Dedup
- [x] Searched for existing PRs/issues and continued the existing OFM-4569 branch instead of opening a duplicate fix path.

### DIFF
--- a/server/src/__tests__/heartbeat-process-recovery.test.ts
+++ b/server/src/__tests__/heartbeat-process-recovery.test.ts
@@ -5156,7 +5156,7 @@ describeEmbeddedPostgres("heartbeat orphaned process recovery", () => {
     });
     expect(sourceIssue).toMatchObject({
       status: "blocked",
-      assigneeAgentId: sourceAssigneeAgentId,
+      assigneeAgentId: agentId,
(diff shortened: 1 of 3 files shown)


tommypoltev:
Superseded by #11817 (cb0009b09, merged 2026-08-20 22:09 UTC), which fixes the root cause on master with an explicit invariant (server/src/services/recovery/service.ts:4374 and :4514-4531). This branch is stale relative to that commit and its delta would regress master. Closing as duplicate. Ref OFM-4569.

PR #11844, opened 2026-08-21
fix(issues): require explicit reopen/resume to un-close a terminal issue via comment (AGE-759)

dzianisv:
fix(issues): require explicit reopen/resume to un-close a terminal issue via comment (AGE-759)

## What happened

PATCH /api/issues/{id} and POST /api/issues/{id}/comments both ran an
independent, undocumented heuristic (shouldImplicitlyMoveCommentedIssueToTodo)
that inferred "the caller wants to reopen this issue" purely from
{ comment present, actor is a "user"-type actor, issue is done/cancelled,
issue has an assignee }. No status, reopen, or resume field was required
in the request body.

A comment-only PATCH ({"comment": "..."}) on a done/cancelled issue would
therefore silently flip status back to "todo" whenever the caller
authenticated as a "user"-type actor (board key, local-implicit, cloud
tenant, session -- this includes any API/script caller, not just the web
UI) and the issue still had an assignee. Reported on AGE-568 and
reproduced by accident on AGE-693.

## Root cause

objectWithoutDefaults(...).partial() already makes status genuinely
optional on the update schema (absent = no schema-level coercion). The
actual defect was in route-handler logic, not schema defaults: a second,
independent inference path recomputed "should this reopen" from side
signals instead of trusting only the explicit reopen/resume flags the
schema already exposes for this exact purpose.

The web UI already computes the identical condition client-side
(shouldImplicitlyReopenComment in IssueChatThread.tsx / CommentThread.tsx
/ TaskChatComposer.tsx) and sends reopen: true explicitly whenever it
means to reopen. Company-wide agent instructions also document
resume: true as the sanctioned way to restart a completed issue via
comment. Both of those explicit paths are untouched by this change. The
server-side heuristic was pure duplication that only mattered for
callers who don't send the flag -- which is exactly the silent-reopen
bug.

## Fix

shouldImplicitlyMoveCommentedIssueToTodo no longer treats a done/cancelled
issue as reopenable from a plain comment; it now only applies to blocked
issues (a mid-flight, non-terminal "please continue" nudge, out of this
ticket's scope and unaffected). Moving a terminal issue back to todo now
requires the request to explicitly say so via reopen: true or resume: true.

## Tests

- Rewrote the three existing tests that asserted implicit closed-issue
  reopening without an explicit flag; added matching explicit-flag
  variants proving the sanctioned reopen: true / resume: true path
  still works unchanged.
- Added an it.each regression covering comment-only PATCH on
  done/cancelled/in_review/todo, verified via the mocked service's
  re-fetch (server/src/__tests__/issue-comment-reopen-routes.test.ts).
- Added a real embedded-Postgres end-to-end regression
  (issue-comment-only-patch-status-e2e.test.ts) that seeds a real issue,
  PATCHes with only { comment }, then verifies status is unchanged via
  a subsequent GET *and* a direct DB read -- not just the PATCH response.

All 194 tests across the touched route/mutation test files pass.

### DIFF
--- a/packages/shared/src/constants.ts
+++ b/packages/shared/src/constants.ts
@@ -874,7 +874,7 @@ export type FinanceUnit = (typeof FINANCE_UNITS)[number];
 export const BUDGET_SCOPE_TYPES = ["company", "agent", "project"] as const;
 export type BudgetScopeType = (typeof BUDGET_SCOPE_TYPES)[number];
 
-export const BUDGET_METRICS = ["billed_cents"] as const;
+export const BUDGET_METRICS = ["billed_cents", "token_count", "run_count"] as const;
 export type BudgetMetric = (typeof BUDGET_METRICS)[number];
 
 export const BUDGET_WINDOW_KINDS = ["calendar_month_utc", "lifetime"] as const;
--- a/server/src/__tests__/budgets-service.test.ts
+++ b/server/src/__tests__/budgets-service.test.ts
@@ -8,6 +8,7 @@ import {
   companies,
   costEvents,
   createDb,
+  heartbeatRuns,
   projects,
 } from "@paperclipai/db";
 import { budgetService } from "../services/budgets.ts";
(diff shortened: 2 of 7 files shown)


dzianisv:
Closing to recreate with a branch name that doesn't embed an internal ticket id, per CONTRIBUTING.md.

PR #11846, opened 2026-08-21
fix(server): soft-delete issues instead of hard delete (AGE-770)

dzianisv:
## Thinking Path

> - The issue system assigns each issue a permanent, human-readable identifier (e.g. `AGE-770`) that other systems (pull requests, commits, comments) cite by reference
> - `DELETE /issues/:id` removed the row from the database. Issue creation derives the next identifier from `MAX(issues.issue_number)` across all issues for the company, as a self-correcting counter (see #2705)
> - Deleting a row removes it from that `MAX()`, so a brand-new issue can be minted with an identifier that a merged or open PR already cites in its title or body, silently breaking that reference
> - This pull request changes `DELETE /issues/:id` to soft-delete (set `deleted_at`) instead of removing the row, so the identifier and issue number can never be reused
> - The benefit is that once an identifier is minted, it stays permanently unique and citable, even after the issue is deleted from default views

Refs: #2705 (the original identifier-collision bug this self-correcting
counter was built to fix — this PR closes the remaining gap where hard
delete could still cause a collision/reuse).

No separate public GitHub issue exists yet for this specific regression, so
describing it here following the bug report template:

**What happened?**
`DELETE /issues/:id` hard-deletes the row via `svc.remove()`. Issue creation
mints the next `issue_number`/identifier using
`coalesce(max(issues.issue_number), 0) + 1` scoped to the company. Because a
hard delete removes the row entirely, the deleted issue's number no longer
counts toward that `MAX()`. If a new issue is created after the delete, it
can be assigned the exact same `issue_number`, and therefore the exact same
`AGE-nnn` identifier, as the deleted one. If any commit, PR title/body, or
comment already referenced that identifier before the issue was deleted,
the identifier now silently points at a completely different issue.

**Expected behavior**
Once an identifier/issue number is minted, it must never be re-assigned to
a different issue, even if the original issue is deleted. Deleting an issue
should hide it from default list/board views but keep the row (and its
number) permanently reserved.

**Steps to reproduce**
1. Create issue A in a company; note its `issue_number` / identifier (e.g. `AGE-100`).
2. `DELETE /issues/:A_id`.
3. Create issue B in the same company.
4. Before this fix: B's `issue_number` reuses A's old number (`AGE-100`) if A was the highest-numbered issue in the company. After this fix: B always gets `MAX(all issues including deleted) + 1`.

**Paperclip version or commit:** `master` @ 599ad701 (pre-fix)

**Deployment mode:** Self-hosted / Cloud (identical code path, not deployment-specific)

**Database mode:** Postgres (including embedded Postgres used in tests)

**Adapter(s) involved:** Not adapter-specific (core bug, server-side issue service)

## What Changed

- Added `issues.deleted_at` (nullable `timestamptz`) via a drizzle migration (`packages/db/src/migrations/0227_blue_iron_fist.sql`).
- Added `issueService.softDelete()`: sets `deleted_at = now()` instead of removing the row.
- Kept `issueService.remove()` (hard delete) in place, unused, as an internal helper for a possible future GDPR/PII purge flow.
- Routed `DELETE /issues/:id` through `softDelete` instead of `remove`. Stopped deleting attachment/document storage objects on delete, since the tombstoned row and its attachments may still be resolved for provenance/history.
- `visibleIssueCondition()` / `visibleIssueSql()` — the single existing choke point used by the issue list, kanban/board queries, dashboard, activity feed, search, etc. — now also filter `deleted_at IS NULL`, so all default views continue to hide deleted issues via one change.
- `getById` / `getByIdentifier` (used by `GET /issues/:identifier` and by PR/commit provenance lookups) intentionally do not filter `deleted_at`, so a deleted issue's identifier still resolves — it just shows up tombstoned (with `deletedAt` set in the response).
- The `MAX(issue_number)` queries in `create` / `importIssues` were already unfiltered by `hidden_at`/`deleted_at`, so soft-deleted rows now correctly count toward the max once hard delete no longer happens for normal deletes — this is the actual fix.
- Fixed `Issue` type fallout from the new required `deletedAt` field: `packages/shared` type + select projections, `packages/plugins/sdk` in-memory test double, and `ui` test/storybook fixtures.

## Verification

Added `server/src/__tests__/issue-soft-delete.test.ts` (embedded Postgres,
real HTTP requests through the actual Express routes, no mocks) covering:

- `DELETE` sets `deleted_at`, and the row still exists in the database afterward.
- A new issue created after a delete gets `issue_number = deleted_issue_number + 1` (does not reuse the deleted number/identifier).
- A deleted issue still resolves by identifier (`GET /issues/:identifier`) for provenance lookups, and reports `deletedAt` in the response.
- A deleted issue is excluded from the default company issue list.

```
$ pnpm --filter server exec vitest run src/__tests__/issue-soft-delete.test.ts
 Test Files  1 passed (1)
      Tests  4 passed (4)
```

Migration generated and verified:

```
$ pnpm --filter @paperclipai/db run generate
[OK] Your SQL migration file: src/migrations/0227_blue_iron_fist.sql

$ cat packages/db/src/migrations/0227_blue_iron_fist.sql
ALTER TABLE "issues" ADD COLUMN "deleted_at" timestamp with time zone;

$ pnpm --filter @paperclipai/db run check:migrations
Migration safety check passed: 20 historical finding(s) covered by baseline (1 stale baseline id(s) ignored).
```

Typecheck run clean on all affected packages: `server`, `packages/db`,
`packages/shared`, `packages/plugins/sdk`, `ui`, `cli`.

## Risks

- Additive, nullable column migration — no backfill needed, safe to apply on top of the current schema.
- Behavior change: `DELETE /issues/:id` no longer deletes attachment/document storage objects. This is intentional (a soft-deleted issue's attachments may still be viewed for provenance/history), but is a visible behavior change from the old hard-delete endpoint.
- `issueService.remove()` (hard delete) is left in the codebase unused by any route. If a future feature needs true hard delete (e.g. GDPR erasure), it should be wired up deliberately with its own guardrails, not accidentally reused for normal issue deletion.
- Low risk otherwise: the fix narrows to a single visibility choke point (`visibleIssueCondition`) plus the DELETE route.

## Model Used

Claude (Anthropic), via the `pi` coding agent harness, extended-thinking/tool-use mode with full repository read/write and shell tool access. Model identifier not separately exposed by the harness beyond "claude" family; agent operated with iterative code read, edit, migration-generation, typecheck, and embedded-Postgres test-execution tool calls.

## Checklist


dzianisv:
Re: the remaining Greptile finding on best-effort wake delivery in `emitBlockerResolvedWakeupsOnDelete` -- this matches the existing delivery guarantee of every other wake emission path in this file already (e.g. the PATCH `/issues/:id` route's own `issue_blockers_resolved` dispatch a few hundred lines above is a fire-and-forget `void (async () => { ... })()` block with the same catch-and-log-only error handling, and `queueTaskWatchdogEvaluation` is the same pattern). None of these paths persist a durable outbox entry before attempting delivery; they all rely on `heartbeat.wakeup`'s own idempotency keys plus periodic re-evaluation (e.g. `listWakeableBlockedDependents` is also called from a periodic backstop, per the comment in `issue-dependency-wakeups.ts`) to eventually catch a dropped wake.

I intentionally made the DELETE path *await* this helper (stricter than the PATCH route's fire-and-forget) so the response can't return before the attempt has actually run, and its own internal errors are already caught and logged rather than surfaced to the caller. Making delivery itself durable (a persisted retry queue ahead of `heartbeat.wakeup`) would be a genuine improvement, but it's a systemic wake-delivery change affecting every one of these call sites, not something specific to soft-delete -- it's out of scope for this ticket (AGE-770 / AGE-743, GitHub issue #2705's identifier-collision bug) and would need its own design + PR.

This PR's core fix (soft delete for identifier permanence) and its checkout/dependency-readiness race-condition fixes are complete and covered by service-level and route-level regression tests, all green in CI (Build, Typecheck, e2e all passing). Ready for maintainer review.

PR #11847, opened 2026-08-21
fix(server): treat a scheduled future monitor as a covered in_review blocker path

dzianisv:
## Thinking Path

> - The blocker-graph liveness classifier decides if a blocked issue's chain is genuinely being worked, or has gone stale
> - The classifier only recognized an active heartbeat run, a pending wake request, or a human assignee as proof that an `in_review` blocker is alive
> - A monitor-covered `in_review` issue that is waiting on a future-scheduled re-check (the documented "genuine external wait, re-check later" pattern) produces none of those three signals until the monitor actually fires
> - So the classifier fell through to `stalled: true` for a leaf that was in fact live and self-documenting, and that false signal rolled up through the blocker graph onto ancestor issues
> - This pull request teaches the classifier to also treat a future `monitorNextCheckAt` on an `in_review` node as a covered, live path
> - The benefit is fewer false "stalled" rollups on issues that are genuinely being tracked by a scheduled monitor, which keeps the anti-false-done/stalled signals trustworthy

**What happened?**
`blockerAttention.classifyPath` in `server/src/services/issues.ts` classifies a blocker node with `status === "in_review"` as covered only when it has an active heartbeat run, a pending wake request, or a human assignee. It never inspects `monitorNextCheckAt`. An `in_review` issue waiting on a future-scheduled monitor (no active run, no wake request) is therefore misclassified as `stalled`, and that false signal rolls up to ancestor issues in the blocker graph.

**Expected behavior**
An `in_review` blocker node with a future `monitorNextCheckAt` should be treated as a covered, live path, the same as an active run, pending wake request, or human assignee.

**Steps to reproduce**
1. Create issue B with `status: in_review` and an `executionPolicy.monitor` scheduled for a future `monitorNextCheckAt`, no active heartbeat run and no pending wake request.
2. Create issue A with `status: blocked`, `blockedBy: [B]`.
3. Call `listIssueBlockerAttentionMap` for A.
4. Observe `blockerAttention.state: "stalled"` on A, even though B is genuinely live and tracked by its scheduled monitor.

`1e456547a0547a8d5464da2e8dc1bbf6f42c1681` (master, at PR open time)

**Deployment mode**
Self-hosted server

**Screenshots / Logs**
See regression tests added in this PR for exact before/after object shapes (`stalled: true` before, `stalled: false` after, for the future-monitor case).

## What Changed

- Loaded `monitorNextCheckAt` alongside the other blocker-graph node fields in `listIssueBlockerAttentionMap` (both the explicit-blocker and child-row queries) in `server/src/services/issues.ts`.
- In `classifyPath`, treat an `in_review` node with a future `monitorNextCheckAt` as a covered waiting path, same as an active run or human assignee.
- Captured one `classificationReferenceNow` timestamp before traversal so every node visited in one blocker-attention calculation compares against the same "now", instead of calling `Date.now()` per node (fixes a Greptile finding).
- Left the existing behavior unchanged for an `in_review` node whose `monitorNextCheckAt` is already in the past — it still classifies as `stalled` (no monitor firing yet is a real staleness signal, not a false positive).

## Verification

- Added 2 regression tests to `server/src/__tests__/issue-blocker-attention.test.ts`:
  - "does not flag an in_review leaf as stalled when it has a future scheduled monitor" — now `covered`.
  - "still flags an in_review leaf as stalled when its scheduled monitor is already in the past" — remains `stalled` (no regression on prior semantics).
- Ran the full suite locally: `npx vitest run server/src/__tests__/issue-blocker-attention.test.ts` → 27 passed (25 existing + 2 new).
- Ran `tsc --noEmit` locally — clean.
- CI: Typecheck, Build, and all General/Verify/e2e test shards pass (see check runs on this PR).

## Risks

Low risk. The change only widens the set of paths `classifyPath` treats as "covered" for `in_review` nodes with a future-scheduled monitor; it does not change behavior for any node without `executionPolicy.monitor`/`monitorNextCheckAt`, and does not change the existing expired-monitor semantics. No schema or migration changes — `monitorNextCheckAt` is an existing column, only newly selected in these two queries.

## Model Used

Claude (Anthropic), via the `pi_local` adapter running as the "Software Engineer" agent role, extended-thinking/tool-use enabled (multi-step code read, edit, and test-run tool calls).

## Checklist

- [x] I have searched GitHub for duplicate or related PRs and linked them above (none found for this internal-tracker bug)
- [ ] I have updated relevant documentation to reflect my changes (no user-facing docs affected; internal classifier behavior only)

### DIFF
--- a/server/src/__tests__/issue-blocker-attention.test.ts
+++ b/server/src/__tests__/issue-blocker-attention.test.ts
@@ -101,6 +101,7 @@ describeEmbeddedPostgres("issue blocker attention", () => {
     originFingerprint?: string | null;
     executionState?: Record<string, unknown> | null;
     description?: string | null;
+    monitorNextCheckAt?: Date | null;
   }) {
     const id = input.id ?? randomUUID();
     await db.insert(issues).values({
@@ -118,6 +119,7 @@ describeEmbeddedPostgres("issue blocker attention", () => {
       originFingerprint: input.originFingerprint ?? "default",
       executionState: input.executionState ?? null,
       description: input.description ?? null,
+      monitorNextCheckAt: input.monitorNextCheckAt ?? null,
     });
     return id;
   }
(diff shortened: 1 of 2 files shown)


juancarlosrial76-code:
Flagging an overlap rather than letting it surface at merge time: #10398 (mine, open since 2026-07-28) fixes the same class of false positive in `listIssueBlockerAttentionMap`. I rebased it onto `d1573244b` today, so both are current against the same base.

Two measured differences, in case they are useful to you regardless of which PR moves:

1. **Exhaustion.** I swapped my call to `hasScheduledIssueMonitorPath(row, now)` for the bare `monitorNextCheckAt > classificationReferenceNow` comparison your diff uses and re-ran the suite: one case goes red — a monitor past `executionPolicy.monitor.timeoutAt`, or at `attemptCount >= maxAttempts`, is not dropped. `tickDueIssueMonitors` will not fire those, so they read as covered while the wake path is actually dead. `hasScheduledIssueMonitorPath` (exported from `recovery/issue-graph-liveness.ts` on current `master`, used at lines 256 and 516) already encodes all three rules.

2. **Scope.** Your change sits in the `in_review` branch of `classifyPath`. The case that started my report is an `in_progress` blocker whose only wake path is its monitor — that edge falls through to the terminal branch and still reads `needs_attention` on `master`. Doing the lookup once into `explicitWaitingIssueIds` covers `in_progress`, `todo` and `blocked` too.

Your `classificationReferenceNow` snapshot is the better detail of the two implementations — I take `Date.now()` once before the chunk loop, which is the same idea but less clearly named, and your comment explains why it matters for sibling agreement.

Happy to defer to yours with the predicate swapped in, or to fold this into mine — whichever the maintainers prefer. Not planning to push a competing change while yours is open.

PR #11900, opened 2026-08-21
fix(secrets): fail the run when a granted env binding never reaches it

ericdfields:
## Problem

A grant is not proof of delivery. Agents were reporting a credential missing at run time while holding the binding for it, and the run surfaced as `adapter_failed` with an opaque provider error:

```
{"type":"error","error":{"name":"ProviderAuthError","data":{"providerID":"openrouter",
 "message":"OpenRouter API key is missing. Pass it using the 'apiKey' parameter or the
 OPENROUTER_API_KEY environment variable."}}}
```

That message points the operator at the agent runtime. The fault is one layer up, in configuration — and nothing in the dispatch path said so.

The existing pre-dispatch check (`collectMissingRuntimeBindings`) walks the **adapter config** and fails when a declared `secret_ref` has no binding row. It structurally cannot see the opposite drift: a **binding row that exists while its `env.<KEY>` config entry is gone**. That run dispatches with the credential silently absent. Silent absence is the defect.

## Change

`secretService.collectUndeliveredEnvBindings` — the inverse walk. Given the key set of the fully resolved run env, it returns every required `env.*` binding for a consumer whose key is not in that set.

`resolveExecutionRunAdapterConfig` runs it across all four consumer scopes (agent, project, environment, routine) after env resolution and before dispatch. Any finding raises `ConfigurationIncompleteFailure` with reason `secret_binding_not_delivered`, routed to a human owner like the other configuration-incomplete blockers, naming the env key and the config path to restore.

This mirrors the shape of the existing `codex_local` pre-dispatch credential gate, which was added for the same reason: *"making a configuration problem look like a runtime failure."*

Deliberately **not** failures:
- `access.<ALIAS>` grants — API-only by design, fetched over the agent secrets API, never injected as env.
- `required: false` bindings — the run is allowed to start without them.

Findings carry key **names**, never values; a test asserts no secret value can reach the raised error.

## Verification

- `server` typecheck clean.
- New `heartbeat-secret-delivery-gate.test.ts` (5 cases) passes; all 4 behavioural cases were confirmed to **fail** with the gate disabled, so they are not vacuous.
- `heartbeat-project-env.test.ts` and the adjacent secrets suites still pass (40 tests).
- New `secrets-service-undelivered-env-bindings.test.ts` (5 cases) covers the query against embedded Postgres. That harness cannot start on the authoring host — every pre-existing embedded-Postgres suite skips there identically — so the query was additionally exercised **read-only against a live 19-secret, 14-agent instance**: it reported zero findings across the whole company (no false positives), and correctly reported all four of one agent's granted-but-undelivered keys, with none of the delivered ones, when handed a drifted env key set.

## Category safeguard

The gate itself *is* the category safeguard: it converts the entire class of "credential declared but absent from the run" from a silent dispatch into a named, pre-dispatch configuration failure, for every adapter and every consumer scope at once, rather than fixing the one binding that happened to be noticed.

### DIFF
--- a/server/src/__tests__/heartbeat-secret-delivery-gate.test.ts
+++ b/server/src/__tests__/heartbeat-secret-delivery-gate.test.ts
@@ -1,0 +1,152 @@
+import { describe, expect, it, vi } from "vitest";
+import {
+  ConfigurationIncompleteFailure,
+  resolveExecutionRunAdapterConfig,
+} from "../services/heartbeat.ts";
+
+/**
+ * BRO-2367 — a grant is not proof of delivery.
+ *
+ * An agent can hold an `env.<KEY>` secret binding and still start a run without
+ * that key, because the binding row and the adapter config can drift apart. The
+ * run then dies inside the agent with an opaque provider error ("OpenAI API key
+ * is missing"), which is recorded as `adapter_failed` — a configuration fault
+ * wearing a runtime fault's clothes. The gate below turns that silent absence
+ * into a named, pre-dispatch configuration failure.
+ */
+describe("resolveExecutionRunAdapterConfig delivery gate", () => {
+  const undeliveredBinding = (envKey: string, consumerId: string, consumerType = "agent") => ({
+    consumerType,
+    consumerId,
+    configPath: `env.${envKey}`,
+    envKey,
+    bindingType: "secret_ref" as const,
+    secretId: "secret-1",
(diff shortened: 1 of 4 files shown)


ericdfields:
Superseded by #11901 — same change, rebranched to a compliant branch name and reopened with the full PR template.

PR #11901, opened 2026-08-21
fix(secrets): fail the run when a granted env binding never reaches it

ericdfields:
## Thinking Path

> - The secrets subsystem gives an agent a credential in one of two classes. An `env.<KEY>` binding is injected into the run environment. An `access.<ALIAS>` binding is API-only and is read through `POST /api/agents/me/secrets/:key/value`
> - The heartbeat resolves those bindings into the adapter config before it dispatches a run. A pre-dispatch check walks that config and fails the run when a declared `secret_ref` has no binding row
> - That check cannot see the opposite drift. A binding row can exist while its `env.<KEY>` entry is gone from the config. Nothing looks at the bindings and asks whether each one reached the run
> - The run then starts without the credential. It fails later inside the agent with a provider error such as "OpenAI API key is missing", and Paperclip records it as `adapter_failed`. The operator is pointed at the agent runtime, but the fault is in configuration
> - This pull request adds the inverse walk. It reads the agent's declared bindings, compares them against the keys in the resolved run environment, and fails before dispatch when a required key did not arrive
> - The benefit is that a missing credential is named as a configuration problem, at the moment it can still be fixed, instead of becoming an opaque runtime failure

No existing public issue covers this. Description follows `.github/ISSUE_TEMPLATE/bug_report.yml`.

**What happened?**

An agent held a secret binding and still failed at run time. The run stopped with an opaque provider error and Paperclip recorded `adapter_failed`:

```
{"type":"error","error":{"name":"ProviderAuthError","data":{"providerID":"openrouter",
 "message":"OpenRouter API key is missing. Pass it using the 'apiKey' parameter or the
 OPENROUTER_API_KEY environment variable."}}}
```

The credential was never in the run environment. Paperclip dispatched the run anyway, because no check compares the declared bindings against the environment the run receives. A grant is not proof of delivery.

**Expected behavior**

Paperclip must fail the run before dispatch when a required `env.*` binding does not reach the run environment. The failure must name the environment key and the configuration path to restore. The failure must not contain a secret value.

**Steps to reproduce**

1. Give an agent a company secret at `env.OPENAI_API_KEY`. A binding row is created.
2. Remove the `OPENAI_API_KEY` entry from that agent's `adapterConfig.env` without a sync that also removes the binding row.
3. Start a run for that agent.
4. The run dispatches. `OPENAI_API_KEY` is absent from the run environment. The run fails inside the agent with a provider authentication error, not with a configuration error.

`master` at `a7e689b3c`.

**Deployment mode**

Self-hosted, authenticated mode.

**Agent adapter(s) involved**

Not adapter-specific. The gate is in the shared heartbeat dispatch path and applies to every adapter.

**Database mode**

Embedded PostgreSQL.

Related open work on the same subsystem, different function: #11489 projects agent API grants under their access alias in `listAgentSecretAccess`. That pull request corrects which grants an agent can address. This pull request checks whether an `env.*` grant reached the run at all. They do not overlap.

## What Changed

- Add `secretService.collectUndeliveredEnvBindings`. It reads the required `env.*` bindings for one consumer, compares each key against the key set of the resolved run environment, and returns a `MissingRuntimeBinding` for each key that did not arrive.
- Add the `binding_not_delivered` secret resolution error code.
- Run the new check in `resolveExecutionRunAdapterConfig` after all environment resolution and before dispatch. It covers all four consumer scopes: agent, project, environment, and routine.
- Raise `ConfigurationIncompleteFailure` with reason `secret_binding_not_delivered` when the check finds anything. The message names the environment key and the configuration path to restore.
- Exempt `access.<ALIAS>` grants. They are API-only by design and are never injected as environment variables.
- Exempt `required: false` bindings. A run is allowed to start without them.
- Make the new method optional on the heartbeat's secret resolver type, so existing partial test doubles still compile. A double that omits it skips the gate. It does not silently pass it.
- Add `heartbeat-secret-delivery-gate.test.ts` and `secrets-service-undelivered-env-bindings.test.ts`.

## Verification

- `pnpm --filter @paperclipai/server typecheck` passes.
- `npx vitest run src/__tests__/heartbeat-secret-delivery-gate.test.ts` passes, 5 tests.
- I disabled the gate and ran the same file again. 4 of the 5 tests failed. The tests are not vacuous.
- `npx vitest run src/__tests__/heartbeat-project-env.test.ts src/__tests__/secrets-service.test.ts src/__tests__/environment-service.test.ts src/__tests__/heartbeat-secret-delivery-gate.test.ts` passes, 40 tests.
- `node scripts/check-forbidden-tokens.mjs` passes.
- `secrets-service-undelivered-env-bindings.test.ts` needs embedded PostgreSQL. That harness does not start on my host, and every pre-existing embedded-PostgreSQL suite skips there in the same way. CI runs it.
- Because of that skip, I also ran the new query read-only against a live instance with 14 agents and 19 secrets. It returned no findings for any agent, so it does not produce false positives on healthy data. When I gave it a run environment with the granted keys removed, it returned exactly the four keys that agent holds, with the correct configuration paths, and no secret value in the output.

## Risks

Low risk. The gate only adds a failure where a run would otherwise start without a credential it declares, so any run it stops was already going to fail — later, and with a less useful error.

The one behavioral shift to note: a run that today starts and fails inside the agent will now fail earlier, as `configuration_incomplete` routed to a human owner rather than `adapter_failed`. This is the intent of the change.

Two exemptions keep the gate from firing on healthy configurations: `access.<ALIAS>` grants and `required: false` bindings. I confirmed on live data that the gate reports nothing across a whole company today.

The low-trust path is unaffected. A binding outside a low-trust boundary already raises `binding_not_allowed` during resolution, which happens before this gate runs.

No migration. No schema change. No API change.

## Model Used

## Checklist

- [ ] I have updated relevant documentation to reflect my changes — no user-facing documentation describes this dispatch-time check; the behavior is explained in code comments at both call sites

### DIFF
--- a/server/src/__tests__/heartbeat-secret-delivery-gate.test.ts
+++ b/server/src/__tests__/heartbeat-secret-delivery-gate.test.ts
@@ -3,6 +3,7 @@ import {
   ConfigurationIncompleteFailure,
   resolveExecutionRunAdapterConfig,
 } from "../services/heartbeat.ts";
+import { LOW_TRUST_REVIEW_PRESET } from "../services/trust-preset-resolver.ts";
 
 /**
  * BRO-2367 — a grant is not proof of delivery.
@@ -70,6 +71,7 @@ describe("resolveExecutionRunAdapterConfig delivery gate", () => {
       "company-1",
       { consumerType: "agent", consumerId: "agent-1" },
       [],
+      { allowedBindingIds: null },
     );
   });
 
@@ -111,8 +113,48 @@ describe("resolveExecutionRunAdapterConfig delivery gate", () => {
       "company-1",
       { consumerType: "agent", consumerId: "agent-1" },
       ["OPENAI_API_KEY"],
+      { allowedBindingIds: null },
     );
     expect(result.resolvedConfig.env).toMatchObject({ OPENAI_API_KEY: "resolved" });
+  });
+
+  /**
+   * A low-trust run is deliberately narrowed to `allowedSecretBindingIds`. A
+   * binding outside that boundary is one the run must NOT receive — resolution
+   * rejects it as `binding_not_allowed` when the config declares it. If the gate
+   * demanded delivery of it anyway, a run that is correctly withholding the
+   * credential would abort as `configuration_incomplete`.
+   */
+  it("carries the low-trust boundary into the gate so a forbidden binding is not required", async () => {
+    const collectUndeliveredEnvBindings = vi.fn().mockResolvedValue([]);
+
+    await resolveExecutionRunAdapterConfig({
+      companyId: "company-1",
+      agentId: "agent-1",
+      executionRunConfig: { env: {} },
+      trustPreset: {
+        kind: "low_trust_review",
+        preset: LOW_TRUST_REVIEW_PRESET,
+        boundary: {
+          mode: LOW_TRUST_REVIEW_PRESET,
+          companyId: "company-1",
+          issueIds: ["issue-1"],
+          allowedSecretBindingIds: ["binding-1"],
+        },
+        sourcePresets: {},
+      } as any,
+      secretsSvc: {
+        ...passThroughSecretsSvc({}),
+        collectUndeliveredEnvBindings,
+      } as any,
+    });
+
+    expect(collectUndeliveredEnvBindings).toHaveBeenCalledWith(
+      "company-1",
+      { consumerType: "agent", consumerId: "agent-1" },
+      [],
+      { allowedBindingIds: ["binding-1"] },
+    );
   });
 
   it("checks every consumer scope that contributed env to the run", async () => {
--- a/server/src/__tests__/heartbeat-secret-delivery-gate.test.ts
+++ b/server/src/__tests__/heartbeat-secret-delivery-gate.test.ts
@@ -1,0 +1,152 @@
+import { describe, expect, it, vi } from "vitest";
+import {
+  ConfigurationIncompleteFailure,
+  resolveExecutionRunAdapterConfig,
+} from "../services/heartbeat.ts";
+
+/**
+ * BRO-2367 — a grant is not proof of delivery.
+ *
+ * An agent can hold an `env.<KEY>` secret binding and still start a run without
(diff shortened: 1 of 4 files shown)


ericdfields:
Addressed in c628483fe.

**The low-trust finding is valid.** Tracing it: when a low-trust run's config still declares an out-of-boundary `secret_ref`, `resolveSecretBinding` already rejects it with `binding_not_allowed` (`secrets.ts`), so the run dies before reaching the gate — that path was never at risk. The reachable case is the *stale* binding: row present, `env.<KEY>` config entry gone. That is precisely the drift this gate exists to catch, and it is resolved by nobody, so the key is legitimately absent from the run env. The unfiltered check read that as non-delivery and would have aborted a run that was correctly withholding a credential it is forbidden to hold.

Fix: delivery can only be required of a credential the run is permitted to receive, so `collectUndeliveredEnvBindings` now takes `allowedBindingIds` and skips bindings outside it. The heartbeat passes `lowTrustAllowedBindingIds` through; outside low trust it is `null` and every required binding stays in scope.

Tests for the boundary case the suite was missing:
- the gate carries the boundary into all four consumer scopes;
- the service ignores an out-of-boundary binding, still reports an in-boundary one, and still requires every binding when no boundary is in effect.

Server typecheck clean; 335 heartbeat/secrets tests pass locally (embedded-Postgres suites skip on this host and run in CI).

**On the three red checks from the last run — all three are unrelated to this change:**
- `General tests (server (2/5))` — `workspace-runtime-exposure-reservation.test.ts` failed on `EADDRINUSE` binding 127.0.0.1:52502 in `startLocalRuntimeService`. A port collision in the runtime harness.
- `Verify serialized server suites (2/5)` — `issue-update-comment-wakeup-routes.test.ts` timed out at 5037ms against a 5000ms limit. That test fully mocks `heartbeatService`, so this PR's code cannot execute in it; it passes locally (9/9) in 5.09s, i.e. it is sitting right on the timeout.
- `verify` — the aggregate gate, red because of the two above.

This push re-runs them.

PR #11903, opened 2026-08-21
fix(approvals): reject agent-filed approvals with no issueIds (INUA-6002)

rotem-zecharia:
## Thinking Path

> - The approvals subsystem lets agents file board-approval requests linked to the issues they are working on
> - When an agent files an approval with `issueIds: null` or `issueIds: []`, the handler skips `linkManyForApproval` and no row is written to `issue_approvals`
> - Stall-recovery detects pending human gates by joining `issue_approvals ⋈ approvals WHERE status='pending'` — no link row means the gate is invisible
> - An agent-filed approval with no issue link is always a caller error: every agent request is scoped to a task, so empty `issueIds` has no valid use case for agents
> - This PR adds a 400 guard at the route layer that rejects agent-filed approvals with an empty issue list before any row is persisted
> - The benefit is that orphaned approval cards cannot be created by agent actors, making stall-recovery reliable for all newly filed cards

**Problem:** Approvals filed by agent actors with `issueIds: null` or `issueIds: []` create orphaned approval rows — no `issue_approvals` link row is written.

**Root cause:** `server/src/services/approvals.ts` normalizes `issueIds` to `[]` when null, then skips `issueApprovalsSvc.linkManyForApproval` for empty arrays. No link row → the stall-recovery query (`issue_approvals ⋈ approvals WHERE status='pending'`) is blind to the card.

**Expected:** Agent-filed approvals must reference at least one issue. Empty `issueIds` from an agent is a caller error and should be rejected with HTTP 400.

**Actual:** The API silently accepts the request, persists an unlinked approval row, and stall-recovery subsequently evicts the filing agent's task.

## What Changed

- **`server/src/routes/approvals.ts`:** Added guard — when `actorType === "agent"` and `uniqueIssueIds.length === 0`, return HTTP 400 with `"Agent-filed approvals must reference at least one issue"`
- **`packages/shared/src/validators/approval.ts`:** Widened `issueIds` to `.nullish()` so existing callers sending explicit `null` are not broken at schema layer before the route guard evaluates them
- **`server/src/__tests__/approval-routes-idempotency.test.ts`:** Three new tests: (1) agent + null issueIds → 400, (2) agent + empty array → 400, (3) human + null issueIds → 201 (exempt)

## Verification

Run the test suite (`pnpm test` in the `server` package) — all 3 new guard tests pass, no existing tests regressed.

Companion corrective fix in paperclip-host handles already-filed orphaned cards: [paperclip-host #214](https://github.com/inu-care/paperclip-host/pull/214).

## Risks

- **Low** for new traffic — the guard only fires for agent callers with empty issue lists, a case that was always semantically incorrect
- Human-filed approvals without issue IDs are explicitly unaffected (user-exempt path)
- Existing orphaned approval rows are not back-filled; the companion corrective fix in paperclip-host handles those

## Model Used

Claude Sonnet 4.6 (`claude-sonnet-4-6`), 200k context, tool use enabled

## Checklist

- [x] I have described the issue in-PR following the relevant issue template (path B)

### DIFF
--- a/packages/shared/src/validators/approval.ts
+++ b/packages/shared/src/validators/approval.ts
@@ -6,7 +6,7 @@ export const createApprovalSchema = z.object({
   type: z.enum(APPROVAL_TYPES),
   requestedByAgentId: z.string().uuid().optional().nullable(),
   payload: z.record(z.string(), z.unknown()),
-  issueIds: z.array(z.string().uuid()).optional(),
+  issueIds: z.array(z.string().uuid()).nullish(),
 });
 
 export type CreateApproval = z.infer<typeof createApprovalSchema>;
--- a/server/src/__tests__/approval-routes-idempotency.test.ts
+++ b/server/src/__tests__/approval-routes-idempotency.test.ts
@@ -446,4 +446,61 @@ describe("approval routes idempotent retries", () => {
     expect(res.body.error).toContain("Cheap status-only recovery runs cannot create or modify approvals");
     expect(mockApprovalService.addComment).not.toHaveBeenCalled();
   });
+
+  it("rejects agent-filed approval when issueIds is null", async () => {
+    const res = await request(await createAgentApp())
+      .post("/api/companies/company-1/approvals")
+      .send({
+        type: "request_board_approval",
+        issueIds: null,
+        payload: { title: "Approve hosting spend" },
+      });
+
+    expect(res.status, JSON.stringify(res.body)).toBe(400);
+    expect(res.body.error).toContain("Agent-filed approvals must include at least one issueId");
+    expect(mockApprovalService.create).not.toHaveBeenCalled();
+  });
+
+  it("rejects agent-filed approval when issueIds is an empty array", async () => {
+    const res = await request(await createAgentApp())
+      .post("/api/companies/company-1/approvals")
+      .send({
+        type: "request_board_approval",
+        issueIds: [],
+        payload: { title: "Approve hosting spend" },
+      });
+
+    expect(res.status, JSON.stringify(res.body)).toBe(400);
+    expect(res.body.error).toContain("Agent-filed approvals must include at least one issueId");
+    expect(mockApprovalService.create).not.toHaveBeenCalled();
+  });
+
+  it("allows user-filed approval with issueIds null (human-exempt path)", async () => {
+    mockApprovalService.create.mockResolvedValue({
+      id: "approval-9",
+      companyId: "company-1",
+      type: "request_board_approval",
+      requestedByAgentId: null,
+      requestedByUserId: "user-1",
+      status: "pending",
+      payload: { title: "Approve hosting spend" },
+      decisionNote: null,
+      decidedByUserId: null,
+      decidedAt: null,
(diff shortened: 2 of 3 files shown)


juancarlosrial76-code:
Independently hit the same bug and landed on a near-identical fix, so I checked this PR first. Two measured data points that support Greptile's "rejects the documented agent strategy-approval flow" finding, plus a scoping suggestion.

**1. The type-agnostic guard rejects a documented agent flow.** The guard fires on `actorType === "agent" && uniqueIssueIds.length === 0` for *every* approval type. But the canonical agent-facing reference documents a CEO-strategy request as an agent call with no `issueIds` at all — `skills/paperclip/references/api-reference.md:865`:

```json
{ "type": "approve_ceo_strategy", "requestedByAgentId": "{your-agent-id}", "payload": { "plan": "..." } }
```

After this change that documented call returns 400.

**2. There is a second in-tree agent caller that deliberately omits `issueIds`.** `cli/src/commands/client/teams.ts` → `resolveApprovalIssueIds()` returns `undefined` when neither `--approval-issue-id` nor a UUID-shaped `PAPERCLIP_TASK_ID` is present:

```ts
function resolveApprovalIssueIds(opts: TeamInstallOptions): string[] | undefined {
  const issueId = opts.approvalIssueId?.trim() || process.env.PAPERCLIP_TASK_ID?.trim();
  if (!issueId) return undefined;
  return isUuidLike(issueId) ? [issueId] : undefined;
}
```

The `isUuidLike` fallback exists precisely because a non-UUID task id is anticipated. In an agent run that hits it, the catalog-team install fallback loses its escalation path exactly when the install was denied. (Reachability of the non-UUID case I have not proven — flagging it as a question, not a defect.)

**Suggestion: scope the guard to `type === "request_board_approval"`.** That covers 100% of the harm described in the Thinking Path — the stall-recovery `issue_approvals ⋈ approvals` gate only matters for board-approval cards — while leaving `approve_ceo_strategy`, `hire_agent`, and `budget_override_required` untouched. The rationale for treating board approvals specially is independent of the actor check: an agent that files one *cannot repair* a missing link afterwards, because `POST /api/issues/:id/approvals` → `assertCanManageIssueApprovalLinks` turns away every agent without `role === "ceo"` or `canCreateAgents`. So for that one type the mistake is permanent, which is what justifies a hard 400.

Scoping it that way also drops the need to touch `packages/shared/src/validators/approval.ts`: an absent `issueIds` already parses fine under the existing `.optional()`, and the route normalizes `null`/absent to `[]` either way. That file is the reason this PR is currently `dirty` — master migrated it to zod v4 (`z.string().uuid()` → `.guid()`).

**Also worth noting, and it applies to my variant too:** this closes the *empty* `issueIds` door but not the *invalid* one. `svc.create()` persists the approval before `linkManyForApproval` runs, with no surrounding transaction — a nonexistent issue id still leaves an orphaned approval row. Same orphan, different entry point. Probably a separate change, but the PR description's "orphaned approval cards cannot be created by agent actors" is stronger than what the diff delivers.

Happy to hand over the scoped variant (route-only, three tests including a negative control that an agent-filed `hire_agent` with no issue still succeeds) if that is useful — say the word and I will open it, or feel free to just take the one-line scope change.

PR #11911, opened 2026-08-21
feat(server): flat monitorStatus field + noWakePath list filter

dzianisv:
## Thinking Path

> - The issue list endpoint (`GET /companies/:companyId/issues`) is what stranding/liveness sweeps use to decide which issues are dead
> - A sweep run reported 47 dead `in_review` issues when the real number was 9, because the list route gave no way to distinguish "no monitor" from "monitor data not returned"
> - Any consumer computing `.executionState?.monitor?.status ?? "none"` on the list response silently treats an omitted field the same as an explicit "no monitor", so a live scheduled monitor looked identical to a dead one
> - Separately, `lastActivityAt`/staleHours alone cannot detect "no wake path": an issue touched 5 minutes ago can still be permanently dead (`in_review` with a cleared monitor), and an issue untouched for 100h with a monitor 2 days out is healthy
> - This pull request adds a cheap, always-present `monitorStatus` field derived from columns already selected on every list row, and a `?noWakePath=true` filter implementing an explicit 4-condition "no wake path" definition
> - The benefit is that stranding sweeps (and any other consumer) get a real signal instead of guessing from an omitted field, cutting false positives from ~5:1 to accurate

**Subsystem affected**
Server — issue list/read routes (`server/src/routes/issues.ts`, `server/src/services/issues.ts`)

**Problem or motivation**
`GET /companies/:companyId/issues` gives no cheap way to tell whether an issue's monitor is live. Consumers fall back to `.executionState?.monitor?.status ?? "none"`, which cannot distinguish "the field was never returned" from "there really is no monitor." In one operator run this made 46 `in_review` issues look dead when 37 of them had a live scheduled monitor — a 5:1 false-positive rate. Separately, recency (`lastActivityAt`) is not the same signal as "will anything ever wake this issue again," so a stale-hours filter alone under- and over-flags issues.

**Proposed solution**
1. Add `deriveFlatMonitorStatus()`: derives `"scheduled" | "triggered" | "none"` purely from the `monitor*` columns already selected on every list row (no schema change, no new column). Wire it onto every issue returned from the list route (full and compact view) and the single-issue read route as `monitorStatus`, always present, never omitted.
2. Add `classifyNoWakePath()`: an explicit 4-condition "no wake path" classifier (blocked with no unresolved blockers and no monitor; in_review with a spent monitor and no unresolved blocker; todo with no assignee; an agent assignee whose own status is "running" but with no active run bound to this issue). Exposed via `GET /companies/:companyId/issues?noWakePath=true`.

**Alternatives considered**
Returning the full `executionState` blob (with a `?include=executionState` gate) was considered and rejected: five of six monitor sub-fields are already flat columns on every list row, so a full opt-in blob is unnecessarily heavy for the one missing bit of information (the derived status enum).

**Roadmap alignment**
Operational/reliability tooling improvement to the existing issue list API; does not touch planned core roadmap work.

## What Changed

- `packages/shared/src/types/issue.ts`: add optional `monitorStatus?: "scheduled" | "triggered" | "none"` to `Issue` and `CompactIssue`.
- `server/src/services/issue-execution-policy.ts`: add `deriveFlatMonitorStatus()`, a pure function computing the flat monitor projection from the existing `monitor*` columns.
- `server/src/services/issue-no-wake-path.ts` (new): `classifyNoWakePath()` / `isNoWakePath()` implementing the 4-condition definition.
- `server/src/services/issues.ts`: attach `monitorStatus` to every row returned from `list()` (both branches); add `noWakePath` to `IssueFilters`; when set, force `includeBlockedBy`, fetch assignee agent statuses, scan a bounded candidate set (`ISSUE_LIST_NO_WAKE_PATH_SCAN_CAP`), classify, then apply the caller's `limit`/`offset` in process.
- `server/src/routes/issues.ts`: parse and validate `?noWakePath=` query param (rejects combination with `?attention=blocked`), wire `monitorStatus` into `toCompactIssue()` and the single-issue `GET /issues/:id` response.
- Tests: `server/src/__tests__/issue-no-wake-path.test.ts` (pure classifier unit tests), `server/src/__tests__/issues-no-wake-path-route.test.ts` (HTTP-level field-presence contract + `noWakePath` filter, embedded Postgres).

## Verification

- `pnpm --filter @paperclipai/server exec vitest run src/__tests__/issue-no-wake-path.test.ts` — 18/18 pass (pure classifier unit tests).
- `pnpm --filter @paperclipai/server exec vitest run src/__tests__/issues-no-wake-path-route.test.ts` — 4/4 pass (embedded Postgres): field presence on every list row, `monitorStatus` on a live-monitor issue matches the single-issue read, compact view carries the field, and `?noWakePath=true` returns exactly the 4-condition matches while excluding healthy issues.
- Full existing suite unaffected: `issues-list-query-parsing.test.ts`, `issue-execution-policy.test.ts`, `issue-monitor-scheduler.test.ts`, `issues-service.test.ts`, `issues-user-context.test.ts`, `issues-checkout-wakeup.test.ts`, `issues-parent-id-alias.test.ts`, `issues-goal-context-routes.test.ts`, `issue-liveness.test.ts` — all pass (142+ tests), confirming no regression to the existing list route.

## Risks

- API response contract change: adds a new optional field (`monitorStatus`) to the issue list/read responses. Additive only — existing consumers ignoring unknown fields are unaffected.
- `?noWakePath=true` is opt-in and classifies in process rather than as a SQL predicate, bounded by `ISSUE_LIST_NO_WAKE_PATH_SCAN_CAP` (5000 rows) before applying the caller's limit/offset; intended for operator/audit sweeps, not hot-path UI polling.
- Low risk otherwise: no schema migration, no change to default (unfiltered) list behavior besides the new always-present field.

## Model Used

Claude (Anthropic), via the Pi coding agent harness — extended reasoning/tool-use mode, used to read the codebase, design the flat-column derivation, implement the service/route changes, and write both the pure-unit and embedded-Postgres integration tests.

## Checklist

### DIFF
--- a/packages/shared/src/types/issue.ts
+++ b/packages/shared/src/types/issue.ts
@@ -824,6 +824,16 @@ export interface Issue {
   monitorAttemptCount?: number;
   monitorNotes?: string | null;
   monitorScheduledBy?: IssueMonitorScheduledBy | null;
+  /**
+   * Cheap, list-safe monitor liveness projection derived from the flat
+   * `monitor*` columns above (see `deriveFlatMonitorStatus` in the server) —
+   * always present (never omitted) whenever an issue is returned from a list
+   * or single-issue read. Collapses the full `executionState.monitor.status`
+   * enum's `"cleared"` into `"none"`, since both mean "no live wake path".
+   * Do not read `.executionState?.monitor?.status` and default to `"none"`
+   * for the same field — use this field instead (AGE-924).
+   */
+  monitorStatus?: "scheduled" | "triggered" | "none";
   executionWorkspaceId: string | null;
   executionWorkspacePreference: string | null;
   executionWorkspaceSettings: IssueExecutionWorkspaceSettings | null;
@@ -925,6 +935,7 @@ export type CompactIssue = Pick<
   archivedByRunId?: string | null;
   activeRecoveryAction: IssueRecoveryAction | null;
   successfulRunHandoff: SuccessfulRunHandoffState | null;
+  monitorStatus?: "scheduled" | "triggered" | "none";
 };
 
 /**
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -21,10 +21,10 @@ overrides:
 
 patchedDependencies:
   acpx@0.12.0:
-    hash: x3fethhotv43zektyl5prdwf54
+    hash: a93c6b5344b036508a5a86ba7e8589b11302119d46bdc51318e297b74fa5666e
     path: patches/acpx@0.12.0.patch
   embedded-postgres@18.1.0-beta.16:
-    hash: 55uhvnotpqyiy37rn3pqpukhei
+    hash: d8b1e087e95f559a6342fc15954f11af22d9dca43a7fcca5762b6459913b5800
     path: patches/embedded-postgres@18.1.0-beta.16.patch
 
 importers:
@@ -108,7 +108,7 @@ importers:
         version: 0.45.2(@electric-sql/pglite@0.3.15)(@opentelemetry/api@1.9.1)(kysely@0.29.5)(pg@8.18.0)(postgres@3.4.9)(sqlite3@5.1.7)
       embedded-postgres:
         specifier: ^18.1.0-beta.16
-        version: 18.1.0-beta.16(patch_hash=55uhvnotpqyiy37rn3pqpukhei)
+        version: 18.1.0-beta.16(patch_hash=d8b1e087e95f559a6342fc15954f11af22d9dca43a7fcca5762b6459913b5800)
       picocolors:
         specifier: ^1.1.1
         version: 1.1.1
@@ -127,7 +127,7 @@ importers:
     dependencies:
       acpx:
         specifier: 0.12.0
-        version: 0.12.0(patch_hash=x3fethhotv43zektyl5prdwf54)
+        version: 0.12.0(patch_hash=a93c6b5344b036508a5a86ba7e8589b11302119d46bdc51318e297b74fa5666e)
       picocolors:
         specifier: ^1.1.1
         version: 1.1.1
@@ -365,7 +365,7 @@ importers:
         version: 0.45.2(@electric-sql/pglite@0.3.15)(@opentelemetry/api@1.9.1)(kysely@0.29.5)(pg@8.18.0)(postgres@3.4.9)(sqlite3@5.1.7)
       embedded-postgres:
         specifier: ^18.1.0-beta.16
-        version: 18.1.0-beta.16(patch_hash=55uhvnotpqyiy37rn3pqpukhei)
+        version: 18.1.0-beta.16(patch_hash=d8b1e087e95f559a6342fc15954f11af22d9dca43a7fcca5762b6459913b5800)
       postgres:
         specifier: ^3.4.9
         version: 3.4.9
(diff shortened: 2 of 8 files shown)


dzianisv:
Both findings addressed in 95ae53b6:

**Issue 2 (condition 4 misclassifies live queued/retrying work as dead)** — fixed. Added `noWakePathLiveExecutionIssueIds()`: it now also matches `heartbeatRuns` by the issue id embedded in `contextSnapshot` (covers `queued`/`running`/`scheduled_retry` regardless of whether `issues.executionRunId` is stamped) and `agentWakeupRequests` by payload issueId/taskId. Regression test added: a queued wake with no bound run, a `scheduled_retry` run, and a `queued` run matched only via `contextSnapshot` are all now correctly excluded from `noWakePath`.

**Issue 1 (bounded scan can silently truncate)** — mitigated, not fully eliminated. Raised `ISSUE_LIST_NO_WAKE_PATH_SCAN_CAP` from 5,000 to 20,000 (covers the overwhelming majority of real companies for a single base-filter query), and added an `X-Paperclip-No-Wake-Path-Scan-Truncated` response header so a caller can detect an incomplete scan instead of silently trusting a partial result. Full exhaustive SQL-pushdown of all 4 conditions (subqueries for unresolved-blocker emptiness, joined agent/run status) is a larger follow-up if a company's backlog genuinely exceeds 20k matching issues — flagging as a known, now-visible limitation rather than a silent one, given this is an opt-in operator/audit endpoint, not a hot path.

dzianisv:
Fixed in 4347e2c2: `offset` is now pushed down as the raw SQL scan offset (skip N candidate rows) instead of being re-applied to the classified matches. A caller that sees `X-Paperclip-No-Wake-Path-Scan-Truncated: true` now has a concrete retrieval path: advance `offset` by `ISSUE_LIST_NO_WAKE_PATH_SCAN_CAP` and repeat until a response comes back without the header — this walks the entire candidate set in bounded windows rather than being stuck after the first 20k rows. Documented explicitly in code that this makes `offset` mean something different for `noWakePath` requests than ordinary list pagination (skip candidate rows, not skip matches), since that's a real, callable-visible semantic change. Added a regression test pinning the cursor behavior on a small dataset.

dzianisv:
Both findings fixed in d370f72f:

**Issue 1 (surplus matches in a window dropped)** — fixed properly, not just mitigated. Each candidate row now carries its raw scan index through classification. The response computes the exact raw-row offset right after the last *returned* match and exposes it as `X-Paperclip-No-Wake-Path-Next-Offset`. A caller no longer guesses a jump size (which could repeat or skip rows) — it just uses the returned offset verbatim. Regression test: 3 matching issues, `limit=1`, verifies the returned next-offset lands exactly on the remaining 2 with no gap or repeat.

**Issue 2 (compact view drops the truncation signal)** — fixed. Both `X-Paperclip-No-Wake-Path-Scan-Truncated` and `-Next-Offset` are now read once from the raw service result and threaded through both the `compact` and `full` response branches (previously only the full branch read them). Regression test added for the compact view case.

dzianisv:
Heads up: the CI-wide failures on the prior push (Build/typecheck/all test shards failing at ~15s with `ERR_PNPM_OUTDATED_LOCKFILE`) are **not from this PR** — `origin/master`'s own `pnpm-lock.yaml` is currently missing the `motion` entry that `ui/package.json` already requires (verified: `git show origin/master:pnpm-lock.yaml | grep motion` → no match, while `ui/package.json` has `"motion": "^12.42.2"`). That breaks `pnpm install --frozen-lockfile` for any PR merged on top of current master, independent of this diff.

I initially regenerated the lockfile and pushed it, but reverted that (a2b06811) once the quality-gate bot correctly flagged it: contributors shouldn't hand-edit `pnpm-lock.yaml` — it's refreshed automatically by a scheduled bot. Waiting for that refresh (or a maintainer fix on master) before re-running CI here.

dzianisv:
CI is fully green as of commit fb8bd8e3 (all 30 checks: Build, Typecheck, server/workspaces tests, e2e shards, verify, policy, review, security scans — all SUCCESS). The earlier lockfile-drift failure was master's own issue, not this PR's; confirmed fixed after merging master. @bertram-gilfoyle-vp-eng — design was already signed off on 2026-08-21T22:37Z; requesting merge approval now that CI is clean.

Issue #11850, opened 2026-08-21
project.executionWorkspacePolicy (isolated_workspace + git_worktree) not honored — agent's own fixed adapterConfig.cwd overrides it silently


johnfergie001:
## Summary

Setting `executionWorkspacePolicy.defaultMode: "isolated_workspace"` with `workspaceStrategy.type: "git_worktree"` on a project has no effect if the agent assigned to that project's issues already has a fixed `cwd` baked into its own `adapterConfig`. Multiple issues in the same project, assigned to the same agent, end up sharing one checkout and stacking diffs/branches on top of each other instead of getting isolated worktrees — with no warning that the project-level policy is being bypassed.

## Environment
- Paperclip `2026.720.0`, `local_trusted` mode
- Project created via `paperclipai project create --execution-workspace-policy-json '{"enabled":true,"defaultMode":"isolated_workspace","allowIssueOverride":true,"workspaceStrategy":{"type":"git_worktree","baseRef":"main"}}'`
- Confirmed via `paperclipai project list` that the policy saved correctly on the project record
- Agent config (`paperclipai agent-config list`) shows `adapterConfig.cwd` set to a fixed path from an earlier, unrelated task

## Steps to reproduce
1. Create a project with `executionWorkspacePolicy` set to `isolated_workspace` / `git_worktree`.
2. Assign an agent to it that already has a fixed `cwd` in its own `adapterConfig` (e.g. reused from a prior project).
3. Dispatch 2+ issues to that agent within the project.
4. `git worktree list` on the resulting checkout shows exactly one shared worktree, not one per issue — confirmed directly on disk, not inferred from status alone.

## Impact
- Sequential tasks land on the same checkout; a later task's branch can contain an earlier task's uncommitted/committed changes (observed: a PR diff included three unrelated prior tasks' files).
- Worse: if a run is interrupted (e.g. hits the adapter execution timeout) before it commits, the next task's setup can silently discard the previous one's finished-but-uncommitted work with no warning that anything was lost.
- The project-level policy gives no indication in the UI/API that it's not actually being applied — `project list` still echoes the policy back as configured.

## Ask
Either have the project's `executionWorkspacePolicy` take precedence over a stale per-agent `cwd`, or surface a clear warning/error when the two conflict, rather than silently falling back to the agent's fixed path.

PR #11914, opened 2026-08-21
fix(workspaces): honor the resolved execution workspace mode on the reuse path

ericdfields:
## Thinking Path

> - Paperclip assigns a workspace to each agent run
> - A project policy can select a shared workspace or a private workspace
> - The server resolves that policy before each run
> - A reused workspace could ignore the resolved mode
> - An inherited issue could also receive a mode that looked like a user override
> - The existing `allowIssueOverride` field did not change server behavior
> - This pull request makes workspace reuse follow the resolved mode
> - The benefit is that a project can change its default mode without a data migration

Refs: #11850

No public issue covers this exact defect.

**What happened?**

An issue can have `executionWorkspacePreference: "reuse_existing"` and a bound workspace. The server resolved the current mode, but the reuse path restored the bound workspace without checking its persisted mode. A shared workspace could therefore remain in use after the project selected `isolated_workspace`.

The server also wrote the realized mode back to the issue. This changed an inherited mode into an explicit issue setting. In addition, the server parsed `allowIssueOverride`, but the mode resolver did not use it.

**Expected behavior**

A reused workspace must provide the mode that the server resolved. An inherited issue must continue to inherit the project mode. A project with `allowIssueOverride: false` must make its default mode mandatory.

**Steps to reproduce**

1. Run an issue in a project that uses `shared_workspace`.
2. Bind the issue to that workspace with `reuse_existing`.
3. Change the project default to `isolated_workspace`.
4. Run the issue again.
5. Observe that the old code restores the shared workspace.

The defect was present at commit `417336f8b`.

**Deployment mode**

Local, single instance.

**Agent adapter(s) involved**

This is core server behavior. It is not specific to one adapter.

**Database mode**

Embedded PostgreSQL.

**Relevant logs or output**

The reuse path updated workspace metadata but did not check whether the bound workspace mode matched the resolved private mode.

**Additional context**

Relabeling the existing row is unsafe. The directory can still be a shared tree. The correct fix is to refuse reuse and let the create path make the private workspace.

## What Changed

- `resolveExecutionWorkspaceReuseRequestForIssue` now receives the resolved mode and the bound workspace mode.
- `isExecutionWorkspaceModeDrift` refuses reuse when a resolved private mode does not match the bound workspace mode.
- The create path now replaces the incompatible binding on the next run. No data migration is required.
- Inherited issue settings now stay on `inherit` after workspace persistence.
- `resolveExecutionWorkspaceMode` now applies `allowIssueOverride: false`.

## Verification

- `pnpm exec vitest run server/src/__tests__/execution-workspace-policy.test.ts server/src/__tests__/heartbeat-workspace-session.test.ts` passes: 175 tests.
- `pnpm --filter @paperclipai/server typecheck` passes.
- The focused tests also pass after rebasing the change onto the latest `paperclipai/master`.
- GitHub CI passes on the pull request.
- Greptile reports 5/5 with no blocking finding.

## Risks

- Risk is low and limited to workspace selection.
- A mismatched private-mode binding now creates a new workspace instead of restoring the old one.
- The old shared workspace stays available for other issues.
- A missing workspace mode is treated as a mismatch when the resolved mode is private.
- There is no schema change and no data migration.

## Model Used

- Anthropic Claude Opus 5 (`claude-opus-5`), extended reasoning, tool use, and code execution produced the implementation.
- OpenAI Codex with GPT-5 and GPT-5.4, reasoning, tool use, and code execution verified the final change. The client did not expose the context-window size.

## Checklist

- [x] I have either linked an existing public issue or described the issue with the bug template fields
- [x] I have not referenced internal or instance-local Paperclip issues or links
- [x] My branch name describes the change and contains no internal issue id
- [x] I have updated relevant documentation or confirmed that no documentation change is required
- [x] I have considered and documented the risks
- [x] Greptile is 5/5 with no open P2 findings, recommendations, or follow-ups

### DIFF
--- a/server/src/__tests__/execution-workspace-policy.test.ts
+++ b/server/src/__tests__/execution-workspace-policy.test.ts
@@ -7,8 +7,10 @@ import {
   buildExecutionWorkspaceAdapterConfig,
   defaultIssueExecutionWorkspaceSettingsForProject,
   gateProjectExecutionWorkspacePolicy,
+  isExecutionWorkspaceIsolationDrift,
   isUnrunnableWorktreeCombo,
   issueExecutionWorkspaceModeForPersistedWorkspace,
+  issueExecutionWorkspaceModeForPersistence,
   parseIssueExecutionWorkspaceSettings,
   parseProjectExecutionWorkspacePolicy,
   ManagedSandboxUnavailableError,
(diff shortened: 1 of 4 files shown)


ericdfields:
Addressed Greptile's finding in `6c1cf5d..HEAD` — but the mechanism it described doesn't hold, so the fix is a relabel rather than a recreate. Detail, since the difference is the whole decision:

**The stated mechanism.** Greptile's claim is that flipping between `isolated_workspace` and `operator_branch` makes the run "use the previous mode's realization." There is no distinct realization to be stuck in. `resolveEffectiveWorkspaceStrategyType` never branches on which of the two modes it is — both take `workspaceStrategy.type` from config and both fall back to `project_primary`. The rest of the server treats them as one class too: the same `assertGitWorktreeBaseWorkspaceReady` gate, the same `isUnrunnableWorktreeCombo` guard, the same `ISOLATED_EXECUTION_WORKSPACE_MODES` serialization bucket, the same `assertExplicitPinnedWorktreeIssueRunnable` branch. So recreating on that flip would discard a perfectly good private tree and rebuild an identical one.

**The real defect underneath it.** The reuse path still never wrote `mode` at all. Refusing reuse covers shared → private; a private → private flip left the create arm as the only writer, so the row kept whichever label first created it. That's a genuine gap and it's the one worth closing.

**What changed.** New `executionWorkspaceModeForReuse` and a `mode` on the reuse `update(...)`, gated so the relabel is only written when it's provably true — both modes private. A resolved mode that promises no private tree leaves the record alone, because rewriting a physically private worktree to `shared_workspace` would be the same DB lie this PR exists to prevent, just pointing the other way.

Tests: added the private → private transition to both suites, plus a guard asserting `isExecutionWorkspaceIsolationDrift` stays `false` across the two private modes — if that ever flips, the relabel has to become a recreate and the test says so. `server` typecheck clean; 175 tests pass across the two suites.

ericdfields:
Greptile is right and my previous comment was wrong. Correcting it, because the reasoning matters more than the verdict.

I argued the two private modes are physically interchangeable, so relabelling the row was sufficient. They *are* interchangeable in strategy resolution — that part holds — but that isn't the operative fact. `mode` is a member of `WORKSPACE_REPLACEMENT_CONFIG_CATEGORIES` (`heartbeat.ts:4643`), so `resolveExecutionWorkspaceConfigFreshness` already classifies **any** mode change as `action: "replace"`. Greptile's "freshness classified the mode change as requiring replacement" was a literal description of the code; I dismissed it without checking that set.

It's worse than a stale label, and my relabel made it worse rather than better. `resolveExecutionWorkspaceReuseProvisioningPolicy` (`heartbeat.ts:4514`) takes `shouldRestoreExistingWorkspace` straight from the reuse request and never consults the freshness action — it only uses `replace` to compute `replacementClassDrift`, which suppresses `shouldPersistLatestWorkspaceConfigMetadata`. So granting reuse across a mode change restored the old tree *and* withheld the fresh fingerprint. The `mode` column I wrote would have permanently disagreed with the stored fingerprint, so the workspace would re-drift on every subsequent run, forever — a self-perpetuating version of the exact bug this PR exists to fix.

**Fix (`b20876c`):** dropped the relabel and `executionWorkspaceModeForReuse` entirely. `isExecutionWorkspaceIsolationDrift` becomes `isExecutionWorkspaceModeDrift` with an exact-match comparison — reuse is refused whenever the resolved mode promises a private tree and the bound workspace was not realized as that same mode. Reuse and freshness now return the same answer, and the create path persists metadata matching the mode it actually realized. Modes that promise no private tree keep historical reuse behaviour.

Also added a test pinning the invariant rather than the case, so this can't regress the same way twice: a granted reuse must never coexist with a replacement-class freshness action.

Server typecheck clean; 175 tests pass across both suites.

ericdfields:
State of this PR after the correction above, since the thread contains a retraction and it should not be ambiguous what is being asked for:

- Head is `b20876c`. Greptile re-reviewed it and its summary now reads **5/5 — "the previously reported cross-mode stale-workspace restore path is no longer reachable. No blocking failure remains."** The P1 it raised twice is closed on its own terms, not argued away.
- All checks green (PR workflow, e2e, serialized server suites, Socket, Snyk, Superagent). `MERGEABLE` / `CLEAN`, no conflicts.

Ready for maintainer review and merge. Downstream, BRO-2375 is holding the `isolated_workspace` rollout to the remaining projects until this lands, because the project-default flip cannot reach any issue that already has a bound workspace while the reuse branch discards the resolved mode.

PR #11905, opened 2026-08-21
Rebuild the onboarding agent arc on the prototype's step design

tonio-alucema:
## Thinking Path

> - Paperclip is the open source app people use to manage AI agents for work
> - The onboarding wizard in `ui/` hires that first agent. It runs three steps: create the agent, connect a model, and review
> - A standalone prototype holds the agreed design for these steps. #10786 ported that prototype, but #11067 reverted it in full because the port deleted `OnboardingWizard.tsx` while four pull requests were editing that file
> - Those four pull requests have since merged. The revert said the port can "re-land incrementally", and this is that re-land
> - This pull request takes the presentational layer from the prototype only. It keeps master's wizard as the source of behaviour, so the eight onboarding fixes merged since the revert stay in place
> - The benefit is that the three agent steps match the agreed design, and no merged fix is lost to get there

## Linked Issues or Issue Description

Refs #10786 — the first attempt to land this design.
Refs #11067 — the revert that asked for it to re-land in smaller steps.

No public issue exists for the re-land. The problem is described below.

**Subsystem affected**

The `ui` package. The change touches the onboarding wizard, the agent capsule,
and one Storybook story. It adds four small presentational components under
`ui/src/components/onboarding/`.

**Current behavior**

The wizard's agent steps do not match the prototype. Each step shows a small
heading beside an icon, above a form. The agent capsule sits below that
heading and does not animate. The agent gets a name but no role, so every
first agent is created as `ceo`.

The wizard also shows a five-segment progress bar on these steps. A walker who
enters on the agent step cannot reach the first two segments, so two of the
five can never be filled.

**Proposed behavior**

The three steps use the prototype's card, its centred display heading, and its
footer. One capsule sits above the heading and stays mounted across all three
steps, so it reads as one object being built rather than three screens that
each show their own.

A three-segment strip counts these steps for a walker who enters on them. The
full-length bar stays for a walker who starts at step one, so that count never
restarts partway.

The agent step gains a role. The options come from the agent role enum, not
from the prototype's mock list.

**Reason and benefit**

The design is agreed and already built once. Re-landing it presentation-first
keeps the behaviour that master gained after the revert.

Sourcing roles from the enum matters. The prototype offers "Coder", which is
not a valid role — the enum uses `engineer` — so a walker who picked it would
fail validation at hire time.

**Breaking changes**

None. The wizard keeps its routes, its draft format, and its hire call. The
draft gains one optional field, `agentRole`. A draft saved before this change
loads without it and falls back to the default.

[...]

PR #11916, opened 2026-08-21
fix(adapter-utils): reap process session wrappers instead of orphaning them

ericdfields:
## Problem

The process session wrapper is launched with `nohup ... &`, so it leaves the run's process group and reparents to init as soon as its launching shell exits. Nothing supervises it after that, which makes leaving on its own the only thing that bounds its lifetime — and it had no reliable way to do so.

A clear-out on one instance killed **205 strays across three runs holding ~10.5 GB**, some four days old.

Three gaps kept the wrapper resident:

1. **The event-file wrapper polled forever after its own child closed**, waiting for a `stdinEnd` that only arrives if the host is still alive to send one. The streamed variant already ended its poll here; the event-file one did not.
2. **`stop()` raced itself.** It wrote that `stdinEnd` and then removed the session directory on the very next line. The wrapper reads stdin on a 50 ms poll, so the removal could delete the request before the wrapper ever saw it — leaving it polling a directory that no longer exists.
3. **Nothing covered a host that died without running `stop()`** at all (a crash or restart).

## Fix

Teardown is now covered three independent ways, because each covers a case the others miss:

- **Child close ends the wrapper** — the normal path, and the one that leaked on every clean run.
- **`stop()` signals the wrapper pid** before removing the session directory, so teardown no longer depends on winning the poll race. The launch script already printed that pid as `$!` and then discarded it; it is now recorded.
- **The wrapper watches for its session directory going away**, so it still leaves when no `stop()` ever runs.

Every path takes the child with it, with a SIGKILL escalation behind a grace period. A teardown that leaves the child behind has only changed which process is the orphan.

The host-side signal is guarded on the pid still looking like a session wrapper, since it was recorded when the session started and may since have been recycled onto something unrelated. When `ps` cannot confirm that — a wrong command, or no `ps` in a minimal sandbox — nothing is signalled and the watchdog covers it. A late teardown is cheaper than killing an innocent process.

The host's kill budget deliberately exceeds the wrapper's own child grace, so SIGKILL lands on a wrapper that is genuinely stuck rather than on one still shutting its child down — killing it mid-teardown would re-orphan the child, which is the exact leak this closes.

## Verification

- **`execute.test.ts` reproduced the leak on a fully passing run**, leaving 40 wrappers holding ~2.0 GB. With this change that suite leaks **none**.
- New `execution-target-session-orphan.test.ts` drives the **real emitted wrapper source** (not a reimplementation) through all three exit paths plus the SIGKILL escalation, asserting the child dies too. It fails on 4 of 5 cases without this change; the fifth is the streamed-wrapper control that already worked.
- `204/204` pass across `execute.test.ts`, `execution-target.test.ts`, `execution-target-sandbox.test.ts`, and `execution-target-stdin-race.test.ts`.
- `tsc --noEmit` clean.

Teardown timings are env-overridable so the tests exercise the real wrapper without waiting out production intervals. Only the timings are tunable, never whether teardown happens.

## Note on what was tried and rejected

Reaping the process group at `runProcess`'s `close` handler looks like the obvious fix and is wrong: warm staged runtimes are deliberately kept alive in that group and reused across turns and compatible resumes. It fails 46 tests in `execute.test.ts`. Cleanup cannot key off one adapter invocation ending — a run is not over when a child closes.

### DIFF
--- a/packages/adapter-utils/src/execution-target-session-orphan.test.ts
+++ b/packages/adapter-utils/src/execution-target-session-orphan.test.ts
@@ -1,0 +1,221 @@
+import { spawn } from "node:child_process";
+import { mkdir, mkdtemp, rm, writeFile } from "node:fs/promises";
+import os from "node:os";
+import path from "node:path";
+import { afterEach, describe, expect, it } from "vitest";
+
+import { getProcessSessionRemoteSource } from "./execution-target.js";
+
+// Regression coverage for orphaned process sessions (BRO-2368). The wrapper is
+// launched with `nohup ... &`, so it leaves the run's process group and
+// reparents to init as soon as its launching shell exits. Nothing supervises it
+// after that, which makes leaving on its own the only thing that bounds its
+// lifetime.
+//
+// Three defects kept it resident. The event-file wrapper went on polling after
+// its own child had exited, waiting for a `stdinEnd` that only arrives if the
+// host is still alive to send one. `stop()` wrote that `stdinEnd` and then
+// removed the session directory on the very next line, so the 50 ms poll could
+// lose the race and never see the request. And nothing at all covered a host
+// that died without running `stop()`. A single clear-out found 205 strays
+// across three runs holding ~10.5 GB, some of them four days old.
+//
+// Each test below drives the real emitted wrapper source, not a reimplementation
+// of it, and asserts on the child as well as the wrapper: a teardown that leaves
+// the child behind has only changed which process is the orphan.
+describe("process session orphan teardown (BRO-2368)", () => {
+  const cleanupDirs: string[] = [];
+  const livePids: number[] = [];
+
+  afterEach(async () => {
+    // Never let this suite become the leak it is testing for.
+    while (livePids.length > 0) {
+      const pid = livePids.pop();
+      if (pid == null) continue;
+      try {
+        process.kill(pid, "SIGKILL");
+      } catch {
+        // Already gone, which is the outcome these tests are asserting.
+      }
+    }
+    while (cleanupDirs.length > 0) {
+      const dir = cleanupDirs.pop();
+      if (dir) await rm(dir, { recursive: true, force: true }).catch(() => undefined);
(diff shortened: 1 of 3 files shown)


ericdfields:
## CI is red on a pre-existing master lockfile drift, not on this change

Every job fails in `Install dependencies` at ~15s, before any code in this PR is compiled or run:

```
ERR_PNPM_OUTDATED_LOCKFILE  Cannot install with "frozen-lockfile"
because pnpm-lock.yaml is not up to date with <ROOT>/ui/package.json
```

The specific drift is `"motion": "^12.42.2"`, present in `ui/package.json` and absent from `pnpm-lock.yaml`. It arrived with #11905 and is on `master` right now:

```
$ git show pcai/master:ui/package.json | grep motion
    "motion": "^12.42.2",
$ git show pcai/master:pnpm-lock.yaml | grep -c motion
0
```

This PR touches only `packages/adapter-utils/src/` and does not modify any `package.json` or the lockfile. CI evaluates the merge commit against `master`, so it inherits the drift; any PR opened against `master` right now fails the same way. Regenerating the lockfile here would put an unrelated, conflict-prone change in this PR, so I've left it to the `Refresh Lockfile` workflow.

### Local verification, in place of the red CI

Run against the same commit, with dependencies from an identical checkout:

- `packages/adapter-utils` — `tsc --noEmit` clean.
- 204/204 pass across `acpx-engine/execute.test.ts`, `execution-target.test.ts`, `execution-target-sandbox.test.ts`, and `execution-target-stdin-race.test.ts`.
- New `execution-target-session-orphan.test.ts` — 5/5. Reverting only the fix and keeping the test fails 4 of the 5; the fifth is the streamed-wrapper control that already exited correctly.
- The leak itself: `execute.test.ts` left **40 orphan wrappers holding ~2.0 GB** on a fully passing run before this change, and **zero** after it. That is measured on the host, not asserted in a test.

Happy to rebase once the lockfile is refreshed on `master`.

PR #11126, opened 2026-08-09
Bound liveness continuation retry storms

goldinbidsheets-rgb:
## Thinking Path

> - Paperclip coordinates agent runs and controls automatic retries.
> - The heartbeat recovery service starts a continuation when a run returns no useful output.
> - Those continuations ran immediately, and each fresh wake reset the per-run attempt cap.
> - Repeated upstream throttling could therefore create an unbounded retry burst across source runs.
> - This pull request adds shadow-first backoff and a per-issue rolling ceiling.
> - The benefit is bounded retry behavior without abandoning transient work.

## Linked Issues or Issue Description

Refs #8829

This pull request supersedes #8829. That pull request has a reviewed implementation, but its July branch no longer shares history with current `master` and cannot merge. This pull request ports the reviewed code and follow-up safety fixes onto current `master` without rewriting the old branch.

Related recovery work: #6979, #6954, and #6957.

## What Changed

- Add an `off`, `shadow`, or `enforce` policy for liveness continuation throttling. The default is `shadow`.
- Compute exponential continuation backoff with bounded jitter and a maximum delay.
- Count consecutive upstream-throttle exits for each issue across fresh source runs in a rolling window.
- Pause the agent and add one consolidated issue notice when enforce mode reaches the ceiling.
- Defer enforce-mode continuations through the durable `scheduled_retry` path.
- Fall back to the existing immediate continuation when durable deferral cannot be scheduled.
- Preserve current interaction-continuation idempotency while accepting a caller-provided retry key.
- Add decision-layer and pure-module regression tests, including the nine-fresh-run reset-hole replay.

## Verification

- `node_modules\.bin\vitest.CMD run server/src/__tests__/run-continuations.test.ts server/src/services/recovery/liveness-continuation-throttle.test.ts`
  - Passed: 2 files, 31 tests.
- `server\..\node_modules\.bin\tsc.CMD --noEmit --pretty false`
  - Passed with no diagnostics.
- The repository-wide stable Vitest runner did not finish within a 10-minute Windows limit. It emitted no result, so this pull request does not count that run as a pass.

## Risks

- The default mode is `shadow`, so this change does not delay continuations or pause agents until an operator enables enforcement.
- Enforce mode can pause an agent after repeated upstream throttling. An operator must review shadow events and own pause monitoring before enabling it.
- The finalization query scans the latest 25 issue runs in the window. A very high-frequency issue can need a larger limit before enforcement.
[...]

PR #11917, opened 2026-08-22
fix(recovery): escalate a self-stranded issue that is older than the lookback

ericdfields:
## Thinking Path

> - A recovery pass looks for issues that can no longer make progress and escalates them
> - One finding, `in_review_without_action_path`, marks an issue that is in review with an agent assignee and no reviewer, interaction, approval, monitor, or active run
> - The pass only escalates a finding inside a lookback window, and it measures that window from the dependency path
> - For this finding the dependency path is the issue itself, so the window is measured against the issue that is already stuck
> - This pull request excludes self-referential findings from the lookback
> - The benefit is that an issue stranded for weeks is escalated instead of being skipped for being quiet

**What happened**

On a self-hosted instance, 16 of 35 issues in `in_review` had no wake path at all. No monitor, no pending interaction, no human assignee, no execution participant. Nothing could resume them. The oldest had been idle for 27 days.

The recovery pass was enabled and running on its normal schedule the whole time. It detected every one of them. It escalated none of them.

**Expected behavior**

A finding that reports an issue cannot make progress should produce one escalation, whatever the age of the issue.

**Steps to reproduce**

1. Create an issue with an agent assignee and no user assignee.
2. Move it to `in_review` with no monitor, no pending interaction and no approval.
3. Leave it for more than `issueGraphLivenessAutoRecoveryLookbackHours`, which defaults to 24.
4. Run `reconcileIssueGraphLiveness`.
5. `findings` is 1. `escalationsCreated` is 0. `skippedOutsideLookback` is 1.

**Cause**

`reviewFinding(issue, issue, [issue])` builds the finding with a dependency path of one entry, the issue itself.

`latestDependencyUpdatedAtForLivenessFinding` reads the newest `updatedAt` across the dependency path. For this finding that is the issue's own `updatedAt`.

`isLivenessFindingInsideAutoRecoveryLookback` then asks whether the issue changed recently. But a stranded issue stops changing **because** it is stranded. The condition that creates the fault is the same condition that hides it. The window can only ever catch an issue in the first 24 hours after it strands, and nothing wakes it in that time to keep it inside the window.

The lookback is correct for a finding that depends on a real blocker. There the blocker's `updatedAt` shows that something moved. It is not meaningful when the path is the issue itself.

**Paperclip version**

Reproduced against `master` at the base commit of this branch.

**Deployment mode**

Self-hosted, single instance, PostgreSQL.

**Related pull requests**

I searched for existing work on this pass. I read #10695, #11126 and #9740; none of them changes how the lookback is measured. This branch is independent of my other open pull requests, #11895, #11896 and #11897.

## What Changed

- Add `isSelfReferentialLivenessFinding`. It is true when the dependency path holds no issue other than the recovery issue.
- Skip the lookback for such a finding, so it is escalated once whatever its age.
- Keep the lookback for every finding that has a real dependency. Behaviour there is unchanged.
- Add a test that seeds a self-stranded `in_review` issue 25 hours old and asserts it is escalated.

Re-escalation stays bounded by the existing cooldown and by the existing-escalation check, so this does not create repeat escalations.

## Verification

Run on Node 24.11.0 with pnpm 9.15.4.

- The new test fails without the change: `expected 1 to be +0`, because `skippedOutsideLookback` is 1 and `escalationsCreated` is 0. It passes with the change.
- `pnpm exec vitest run server/src/__tests__/heartbeat-issue-liveness-escalation.test.ts` — 25 tests passed. This includes the existing test `does not create recovery issues outside the configured lookback window`, which is unchanged and still passes, because it seeds a blocked chain and is not self-referential.
- `heartbeat-issue-liveness-escalation`, `issue-review-attention` and `attention-service` together — 3 files, 53 tests passed.
- `pnpm --filter @paperclipai/server typecheck` passed.
- `pnpm --filter @paperclipai/shared typecheck` passed.
- `pnpm --filter @paperclipai/db check:migrations` passed.
- `node scripts/check-node-version-policy.mjs` passed.
- `pnpm install --frozen-lockfile` exits 0.

## Risks

This makes the pass escalate findings it used to skip. On an instance with a backlog of long-stranded issues, the first run after this change creates one escalation for each. That is the intent, but it can be a large first batch. Operators who want to see the size first can call the preview endpoint before deploying.

The change is narrow. A finding with any dependency other than the issue itself keeps the current behaviour exactly, so blocked-chain recovery is untouched.

There is no migration and no API change.

## Model Used

Claude Opus 5 (`claude-opus-5[1m]`), 1M context window, extended thinking enabled, with tool use for repository search, edits, test runs and git operations.

## Checklist

### DIFF
--- a/server/src/__tests__/heartbeat-issue-liveness-escalation.test.ts
+++ b/server/src/__tests__/heartbeat-issue-liveness-escalation.test.ts
@@ -906,6 +906,24 @@ describeEmbeddedPostgres("heartbeat issue graph liveness escalation", () => {
     expect(escalations).toHaveLength(1);
   });
 
+  it("previews a self-stranded in_review issue that is older than the lookback window", async () => {
+    // The preview exists so an operator can size the batch before it runs. It must
+    // apply the same exemption as reconciliation, or it hides exactly the findings
+    // that the next reconciliation is about to escalate.
+    await enableAutoRecovery();
+    await seedSelfStrandedReviewIssue({ idleHours: 25 });
+    const heartbeat = heartbeatService(db);
+
+    const preview = await heartbeat.buildIssueGraphLivenessAutoRecoveryPreview();
+
+    expect(preview.skippedOutsideLookback).toBe(0);
+    expect(preview.items).toHaveLength(1);
+    expect(preview.items[0]).toMatchObject({ state: "in_review_without_action_path" });
+
+    const result = await heartbeat.reconcileIssueGraphLiveness();
+    expect(result.escalationsCreated).toBe(preview.items.length);
+  });
+
   it("does not create recovery issues outside the configured lookback window", async () => {
     await enableAutoRecovery();
     const { companyId } = await seedBlockedChain({ outsideLookback: true });
--- a/server/src/__tests__/heartbeat-issue-liveness-escalation.test.ts
+++ b/server/src/__tests__/heartbeat-issue-liveness-escalation.test.ts
@@ -843,6 +843,69 @@ describeEmbeddedPostgres("heartbeat issue graph liveness escalation", () => {
     });
   });
 
+  async function seedSelfStrandedReviewIssue(opts: { idleHours: number }) {
+    const companyId = randomUUID();
+    const agentId = randomUUID();
+    const issueId = randomUUID();
+    const issuePrefix = `S${companyId.replace(/-/g, "").slice(0, 6).toUpperCase()}`;
+
+    await db.insert(companies).values({
+      id: companyId,
+      name: "Paperclip",
+      issuePrefix,
+      requireBoardApprovalForNewAgents: false,
+    });
+    await db.insert(agents).values({
+      id: agentId,
+      companyId,
+      name: "Coder",
+      role: "engineer",
+      status: "idle",
+      adapterType: "codex_local",
+      adapterConfig: {},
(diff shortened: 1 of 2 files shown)


ericdfields:
Good catch, and it was a real inconsistency. Fixed in `a5d5a42`.

The preview kept the unconditional lookback while reconciliation skipped it for a self-referential finding. So the preview would have hidden exactly the findings the next reconciliation was about to escalate. That also made the claim in this pull request's Risks section wrong: I told operators to size the first batch with the preview, and with the first commit alone that number would have been too low.

The preview now applies the same exemption. A finding with a real dependency is unchanged in both paths.

One deliberate difference from the suggested fix: I kept the null check as its own early skip rather than folding it into the exemption. `latestDependencyUpdatedAt` feeds `IssueGraphLivenessAutoRecoveryPreviewItem.latestDependencyUpdatedAt`, which is a non-nullable `string`. Widening that type would reach every consumer of the preview for an edge case that a self-referential finding cannot produce, because its dependency path is the issue itself and that row is always loaded. A null timestamp means the preview genuinely cannot size the item, so skipping it is honest.

Added a test that asserts the preview reports the old self-stranded finding **and** that the item count equals what reconciliation then creates, so the two paths cannot drift apart again.

Verification on Node 24.11.0: 26 tests pass in the liveness suite, 60 across liveness plus instance-settings-routes, both typechecks pass, and `check:migrations` and the Node version policy gate pass.

PR #11466, opened 2026-08-16
fix(server): atomically set checkout run contextSnapshot with issue lock

santhiprakash:
## Thinking Path

> - Paperclip is the control plane that issues heartbeat runs to agents and checks every mutating call against the run that made it.
> - `checkout()` is how an agent heartbeat claims the issue it is about to work on.
> - Cross-issue influence limits and subtree pause-hold waivers both look at `heartbeat_runs.context_snapshot` to decide which issue a run is allowed to write to.
> - `checkout()` updated `issues.checkout_run_id` and `issues.execution_run_id`, but it never wrote the issue id back into the run's `context_snapshot`.
> - So a run could successfully `checkout()` an issue and then be rejected when it tried to comment on or update that same issue with `cross_issue_influence_run_context_required`.
> - This pull request makes `checkout()` write the issue id into the heartbeat run's `context_snapshot` at the same time it takes the checkout lock.

## Linked Issues or Issue Description

No public issue exists for this. This is a correctness fix for the checkout / cross-issue write path.

**What happened?**
A heartbeat run that used `issueService.checkout()` to take ownership of an issue could then be denied when it tried to post a comment or mutate the issue. The cross-issue influence limit service (`cross-issue-influence-limit.ts`) reads `heartbeat_runs.context_snapshot` to find the source issue id, but `checkout()` left `context_snapshot` empty (or with an unrelated issue id from an earlier wake). The write fails with `cross_issue_influence_run_context_required` because the run context has no source issue.

**Expected behavior**
Checking out an issue should make that run's `context_snapshot` point at the checked-out issue, so same-issue writes are recognized and cross-issue writes can be counted correctly. The issue row update and the run `context_snapshot` update should be committed together so a failed run-context write cannot leave an issue locked without the matching run context.

**Steps to reproduce**
1. Create a heartbeat run without `context_snapshot.issueId`.
2. Call `issueService.checkout(issueId, agentId, ["todo"], runId)`.
3. Use that run id to comment on or update the same issue.
4. The write fails with `cross_issue_influence_run_context_required` because the run context has no source issue.

**Paperclip version or commit**
`master` at `cd501499a` and earlier.

**Deployment mode**
Affects all modes (local dev, Docker, and self-hosted).

## What Changed

- Added `setCheckoutRunContextSnapshot()` in `server/src/services/issues.ts`.
- Added `updateIssueAndSetCheckoutRunContext()` helper that updates the issue row and the heartbeat run `context_snapshot` inside a single `db.transaction()`.
[...]

PR #11918, opened 2026-08-22
fix(server): fail closed on heartbeat run context for issue checkout and release

aduduman031084:
## Thinking Path

> - Agents do their work through the issue routes. Each write carries an `X-Paperclip-Run-Id` header that names the heartbeat run the agent is executing
> - The header is only a claim. The server must check it against the persisted `heartbeat_runs` row before it changes anything
> - Issue checkout and issue release did not make that check. Checkout also wrote the target issue into the supplied run, so the request created the authority it was about to use
> - A same-company agent can choose the header value. It could therefore take the execution lock on an issue that it was never woken for
> - This pull request loads the persisted run first and refuses the write when the run is not live, not ours, or not bound to the target issue
> - The benefit is that an agent can only act on the issue Paperclip actually gave it, and every refusal leaves an audit record

Related pull request: #11466. That pull request changes the same checkout path. It writes the run context during checkout instead of requiring it, so it does not close the hole described below. This change replaces that approach.

**What happened?**

`POST /api/issues/:id/checkout` selected the target issue and then rewrote the supplied run `contextSnapshot` to name that issue, inside the same transaction that took the lock. The run was matched on id and agent only. A persisted run with an empty context, a terminal `failed` run, and a live run bound to a different issue could each take the target execution lock, set the issue to `in_progress`, and gain authority over it. `POST /api/issues/:id/release` checked no run at all. An unknown run id that was never persisted cleared the assignee on an idle issue.

**Expected behavior**

An agent write must be refused unless the supplied run exists, is still running, belongs to that agent, belongs to that company, and already names the target issue. The server must never create that binding from the request that wants to use it.

**Steps to reproduce**

1. Authenticate as an agent in a company.
2. Create or find a persisted heartbeat run for that agent with an empty `contextSnapshot`.
3. Call `POST /api/issues/<other-issue-id>/checkout` with that run id in `X-Paperclip-Run-Id`.
4. The issue moves to `in_progress`, the execution lock is taken, and the run context is rewritten to the target issue.

`master` at `dc5b07070`.

**Deployment mode**

Local trusted, single instance.

## What Changed

- `server/src/routes/issues.ts` — added `assertCanonicalAgentRunContext`. It loads the persisted run and refuses the write unless the run exists, has status `running`, matches the actor agent, matches the issue company, and already names this issue in its context. It never writes run context.
- `server/src/routes/issues.ts` — applied that guard to `POST /issues/:id/checkout` and `POST /issues/:id/release`. Checkout can no longer create its own authority.
- `server/src/routes/issues.ts` — each refusal writes an `issue.agent_run_context_denied` activity record. The record holds the claimed run id and the check that failed. A run id that is not persisted goes in the details, because the column is a foreign key.
- `server/src/services/cross-issue-influence-limit.ts` — the existing locked run check now also requires status `running`. A finished run is a spent credential. Comment, PATCH, and interaction resolution already went through this check for identity and context. Status was the one missing requirement.
- `server/src/__tests__/issue-run-context-authorization-routes.test.ts` — new regression suite. 24 tests.
- Three test fixtures gained the `status` column and the run context that real rows always carry. One test that let an agent act through another agent run id now gives that agent its own run.

The cross-issue caps stay closed by default. Release keeps its status contract. Only `in_progress` returns to `todo`, and every other status stays.

## Verification

Local commands and results:

- `vitest run src/__tests__/issue-run-context-authorization-routes.test.ts` — 24 passed. The suite covers the good path (one live run checks out, comments, patches, and releases its own issue) and the refused paths for checkout, release, comment, and PATCH: unknown run, terminal run, empty-context run, other-agent run, other-company run, and other-issue run. Each refused case asserts that the issue row, execution lock, assignee, status, comments, and run context did not change.
- `vitest run src/__tests__/issue-stale-execution-lock-routes.test.ts` — 11 passed.
- `vitest run src/__tests__/issue-agent-mutation-ownership-routes.test.ts` — 87 passed.
- `vitest run src/__tests__/issue-execution-policy-routes.test.ts` — 19 passed.
- `vitest run src/__tests__/issue-thread-interaction-routes.test.ts` — 69 passed, 1 failed. The failure is a 10 second timeout on a board-authored test. It fails the same way on unmodified `master` on this host.
- `tsc --noEmit` — no error in the changed files.

## Risks

- Behavior change. An agent whose run carries no issue context can no longer check out an issue. This is the intended result, because that bootstrap was the defect. Wakes that already carry `issueId` or `taskId` are not affected. Run creation must supply the intended issue context for any wake that is expected to take work.
- The guard reads the run row and does not lock it. The check can only move against the caller. A run that passes was authorized at that moment, and a run that ends later loses authority instead of gaining it.
- Error copy changes for the two guarded routes. They now answer 403 with code `agent_run_context_invalid`.
- No migration. No schema change.

## Model Used

Claude Opus 5, model id `claude-opus-5[1m]`, 1M context window, extended thinking, with tool use and code execution.

## Checklist

The branch name box is left unchecked on purpose. The branch carries an internal ticket id because our own process requires one branch per ticket. Tell me if you want the branch renamed and I will move the commits.

When you squash this, please keep the author trailer:

    Co-Authored-By: Andrei Duduman <andrey330084@users.noreply.gitlab.com>


aduduman031084:
## Security NO-GO

`VERDICT:v1 security NO-GO 2d7c2d4196d7c76768cd5fd2b2c2d5926096ffa6`

**Class:** OWASP Broken Access Control / API Broken Function-Level Authorization (CWE-863); STRIDE Tampering/Spoofing; Complete Mediation, Secure Defaults, Fail Securely.

**Severity:** High (7/10). **Exploitability:** Medium-high for an authenticated same-company agent key able to name one of its own scheduled-retry run IDs (7/10).

### Blocking evidence

1. **`scheduled_retry` is accepted as live write authority.** Both new gates authorize every status outside `TERMINAL_HEARTBEAT_RUN_STATUSES`, which includes `scheduled_retry` even though recovery code documents that it has no process on purpose. Executable proof at this exact head: a canonical persisted `scheduled_retry` run sent `POST /api/issues/:id/checkout`; the response was 200, the issue moved `todo -> in_progress`, and both run locks were installed.
2. **Malformed IDs still 500 on richer PATCH shapes.** A realistic agent-key actor with an active operator membership sent `PATCH /api/issues/:id`, `X-Paperclip-Run-Id: ../../not-a-run-id`, and `{ "blockedByIssueIds": [] }`. The response was 500; the issue row remained unchanged. `resolveAgentTrustForIssue` reaches a UUID comparison before the new limiter and has no UUID guard.
3. **Liveness is not atomic with mutation.** Checkout/release check the run before the service write, while `observeCrossIssueInfluence` releases its run-row lock before comment/PATCH durable writes. Concurrent terminalization can land between authorization and mutation; the code comment claiming no race is incorrect.

### Required remediation

- Use one explicit agent-write allowlist: `queued`, `running`. Do not authorize by excluding terminal states.
- Couple run liveness validation and the durable issue write in one transaction, with a terminalization-race regression.
- Validate agent-key run IDs at one shared boundary before every UUID query, including `resolveAgentTrustForIssue`, and return an audited 403/422.
- Add `{checkout, release, comment, PATCH} x scheduled_retry` zero-mutation cases plus a malformed blocker-list PATCH case.

### Verification

- Submitted focused suites pass: 2 files, 41/41 tests.
- Additional adversarial proofs above reproduce both defects.
- Typecheck/build/general server shards are green, but the PR is currently `UNSTABLE` with two serialized shards and e2e red; the security verdict is independent of those CI failures.

**Blast radius:** A caller can capture and mutate the canonical issue before any heartbeat process exists. Canonical binding limited the demonstrated scope; no cross-company read or arbitrary wrong-issue mutation was demonstrated. The malformed-ID path adds authenticated internal-error/log pressure but no observed durable mutation.

**Residual risk after fix:** A compromised queued/running run retains intentionally scoped authority over its canonical issue. Short queue/run lifetimes, run-bound JWTs, revocation, rate limits, and audit monitoring remain required.

Issue #11401, opened 2026-08-15
[BUG] HTTP 500 on PATCH /api/issues/{id} when monitor.nextCheckAt is past and attemptCount=0


mgcstudios86:
## Summary

A specific state in the issue-monitor lifecycle causes **HTTP 500 Internal Server Error** on any `PATCH /api/issues/{id}`, regardless of the field being patched. The state is: an issue has `executionPolicy.monitor.nextCheckAt` already in the past and `monitorAttemptCount = 0` (i.e. the cron sweeper has not yet consumed the monitor).

This is distinct from #10374 (silent monitor destruction on transient dispatch skip → 200) and #10721 (`in_review` gate accepting past-dated monitors). The symptom here is a hard 500, not a silent failure, and it persists until the unattended sweep fires at least once.

## Where

Suspected in the PATCH route's interaction with `applyMonitorTransition` and the monitor state machine in `server/src/services/issue-execution-policy.ts`. The exact throw site was not captured before the state self-healed, but the failure mode matches a path that asserts `attemptCount > 0` or treats a past `nextCheckAt` as an invariant violation when `input.policy?.monitor` is unset on the PATCH.

A separate scenario in the same family is reported on #10374 (comment by antiautomation): after a successful fire, `PATCH {"executionPolicy": null}` returns 200 with the monitor still `status: "triggered", clearedAt: null`. That is the post-sweep symmetric failure. This issue covers the pre-sweep, hard-500 failure.

## Reproduction (server-side, against self-hosted instance)

1. Create an issue.
2. PATCH the issue to arm a monitor with `nextCheckAt` already in the past and `attemptCount = 0`. Concretely:

   ```json
   {
     "executionPolicy": {
       "monitor": {
         "nextCheckAt": "2020-01-01T00:00:00.000Z",
         "scheduledBy": "assignee",
         "notes": "test repro"
       }
     }
   }
   ```

3. Without waiting for the cron sweeper, send any PATCH against that issue. All of the following returned **HTTP 500 Internal Server Error** in the observed repro:

   - `PATCH { "status": "cancelled" }` → 500
   - `PATCH { "executionPolicy": { "monitor": null } }` → 500
   - `PATCH { "monitorNextCheckAt": null }` → 500
   - `POST /api/issues/{id}/release` → 500
   - `DELETE /api/issues/{id}` → 500

4. Other endpoints on the same row were unaffected:

   - `POST /api/issues/{id}/comments` → 201 OK
[...]

PR #11919, opened 2026-08-22
fix(issues): scope monitor maxAttempts to one monitor, not the issue lifetime

ericdfields:
## Thinking Path

> - The heartbeat scheduler owns issue liveness: an agent-owned issue must keep a path that can wake its assignee again
> - An issue monitor is that path for a wait on an async or external system, and `maxAttempts` bounds it
> - The attempt count is stored on the issue and is never reset, but it is compared against the `maxAttempts` of each new monitor
> - The bound therefore applies to the issue for its whole life, not to one monitor, so a spent monitor blocks every later monitor with the same bound
> - This pull request scopes the attempt count to one monitor and makes the rejection state the value that would be accepted
> - The benefit is that a small, sensible `maxAttempts` no longer removes the issue's ability to hold a monitor

No public GitHub issue exists for this. The problem, in the form of `.github/ISSUE_TEMPLATE/bug_report.yml`:

**What happened?**

An issue cannot receive a new monitor after an earlier monitor spends its attempts. The request fails:

```
422 {"error":"Monitor bounds are already exhausted",
     "details":{"clearReason":"max_attempts_exhausted"}}
```

`exhaustedMonitorClearReason()` in `server/src/services/issue-execution-policy.ts` compares the persisted attempt count with the `maxAttempts` of the monitor that the caller supplies now. The count comes from the `monitorAttemptCount` column and from the monitor execution state. Nothing resets either one. The count is therefore cumulative for the life of the issue.

Two documented recoveries do not work:

- A PATCH that sets `executionPolicy.monitor` to `null` does not reset the count. It also does not clear the monitor state, because `buildIssueMonitorTriggeredPatch()` already removed the monitor from the policy when the monitor fired. The API sees no monitor change at all.
- A new policy with a low `maxAttempts` is rejected again. Only a larger `maxAttempts` is accepted, and the value that is large enough increases for the life of the issue.

`maxAttempts` must bound one monitor. A new monitor must start with a fresh attempt count. If a bound is refused, the error must state the count that was observed and the value that is accepted.

**Why this matters**

A monitor is one of the few paths that keep an issue in `in_review` reachable. An issue that cannot hold a monitor stays live only through a human assignee, a pending interaction, or a typed execution participant. A bounded monitor therefore becomes a permanent loss of capability, and the failure appears much later than the cause. Three issues were found in `in_review` with no wake path for this reason.

**Steps to reproduce**

1. Schedule a monitor with `maxAttempts: 3` on an issue that is `in_progress` or `in_review`.
2. Let the monitor run until the attempts are spent.
3. Schedule a new monitor on the same issue with `maxAttempts: 3`.
4. The request fails with `422 max_attempts_exhausted`.

**Related work**

- Refs #10735 — makes a dispatched monitor re-arm itself. That pull request keeps a live monitor in the `scheduled` state across attempts, so the counting rule in this pull request continues to bound it. The two changes are compatible.
- Refs #10374 — a related report about monitors that are destroyed by a transient dispatch skip.
- Refs #11401 — a related report about monitor scheduling validation.

## What Changed

- `carriedMonitorAttemptCount()` carries the attempt count only while the monitor is still `scheduled`. A monitor that fired or was cleared has completed its lifecycle, so the next `nextCheckAt` is a new monitor and starts at zero.
- A new monitor also resets the `monitorAttemptCount` column. The dispatcher and the liveness classifier read that column, so the state and the column must agree.
- A rejected monitor now reports the bound that stopped it. The message and the `details` object give `attemptCount`, `maxAttempts`, and `minimumMaxAttempts`. A stale `timeoutAt` gets its own message.
- Review-path recovery identifies a monitor wake by the scheduled instant of the monitor, and uses the attempt count only as a fallback. Attempt counts now restart for each monitor, so two monitors on one issue could otherwise share a fingerprint, and the recovery wake of the second one would be suppressed without a record.
- `doc/execution-semantics.md` states that the bounds apply to one monitor, and states what each rejection reports.

The rejection is kept for the case that it describes correctly: a `maxAttempts` that is lower than the attempts the live, still-scheduled monitor already spent is still refused.

## Verification

Automated, from `server/`:

```bash
npx vitest run src/__tests__/issue-execution-policy.test.ts \
  src/__tests__/issue-execution-policy-routes.test.ts \
  src/__tests__/issue-monitor-scheduler.test.ts \
  src/__tests__/recovery-classifiers.test.ts \
  src/services/recovery/review-path-recovery.test.ts
pnpm --filter @paperclipai/plugin-sdk ensure-build-deps && npx tsc --noEmit
```

New tests:

- A monitor that is scheduled after the previous monitor is spent starts at attempt zero, in the execution state and in the `monitorAttemptCount` column.
- The route test reproduces the report: a PATCH that schedules `maxAttempts: 3` on an issue with three spent attempts now returns 200 and writes `monitorAttemptCount: 0`.
- A `maxAttempts` below the attempts of the live scheduled monitor is still rejected, with the minimum value in the message.
- A `timeoutAt` in the past is named in its own message.
- Two monitors on one issue produce different review-path recovery fingerprints.

## Risks

Low risk, with one behavior change that is intentional: an issue can hold a new bounded monitor after an earlier monitor is spent. `maxAttempts` bounds a monitor, and a caller that arms a new monitor makes a new, recorded decision. Every exhausted monitor still logs `issue.monitor_exhausted` and still runs its recovery, so repeated exhaustion stays visible.

No schema change, no migration, and no change to the monitor dispatch path.

## Model Used

Claude Opus 5 (Anthropic, model id `claude-opus-5`, 1M context window), with extended thinking and tool use, running in Claude Code.

## Checklist


ericdfields:
CI note: the one red check on this PR — `General tests (server (2/5))`, and the `verify` aggregate that depends on it — is a pre-existing flake unrelated to this change.

**Failure:** `server/src/__tests__/workspace-runtime.test.ts:7807`, in `does not reuse a stopped auto-port service port while another process owns it`:

```
AssertionError: expected { reconciled: 1, adopted: +0, …(5) }
                to match object { reconciled: 1, adopted: 1, …(1) }
```

**Why it isn't this PR:** this PR touches six files, none of them `workspace-runtime.test.ts` or any workspace-runtime source. The changed surface is `issue-execution-policy.ts` and `recovery/review-path-recovery.ts` plus their tests and `doc/execution-semantics.md`.

**Same assertion, same line, on unrelated branches** (all within the last few hours):

| Run | Branch |
|---|---|
| 32540098738 | `fix/hermes-gateway-delta-log-fragmentation` |
| 32539956487 | `feat/read-only-roster-audit` |
| 32543123580 | `feat/sandbox-duplex-broker` |
| 32543446807 | `fix/FAI-9983-fail-closed-run-context-authorization` |

It is intermittent rather than deterministic — three other runs I checked in the same window passed it. The test spawns a real process on an auto-assigned port and asserts the startup reconciler adopts it, so it is sensitive to process/port timing on a loaded runner. Worth a separate flake-fix issue; it is not a signal about this change.

This branch is already current with `master` (`dc5b07070`), so there is nothing to rebase in order to re-run. Fork PRs cannot re-run their own jobs — a maintainer re-run of the failed job should clear it.

Greptile is 5/5 on this commit with no blocking or non-blocking findings.

ericdfields:
Re-triggered CI to retry that flake. I force-pushed `b8175605f` → `2f0625e8f`: an `--amend --no-edit` with nothing staged, so the tree is byte-identical (`222cecd7d5757c835272dfc788ee7c5876e30591` before and after) — no content change, purely a way to get the `General tests (server (2/5))` shard re-run without maintainer access.

If the same assertion at `workspace-runtime.test.ts:7807` fails again, I will not push a third time; that would be worth a separate flake-fix issue and this PR still stands on its own.

ericdfields:
CI is now fully green on \`2f0625e8f\` — including \`General tests (server (2/5))\` (6m41s, pass) and the \`verify\` aggregate. [Run 32556993252](https://github.com/paperclipai/paperclip/actions/runs/32556993252).

That confirms the earlier red shard was the flake described above and not this change: the tree is byte-identical between the two runs (`222cecd7d5757c835272dfc788ee7c5876e30591`), so the only variable was the retry.

Nothing outstanding from my side — Greptile passed, the PR is `MERGEABLE`/`CLEAN`, and all checks are green. Ready for maintainer review whenever someone has a moment.

PR #11921, opened 2026-08-22
fix(dev): launch the dev server through a portable entry point

qasimsethi-code:
`.claude/launch.json` names `/bin/sh`, which is not a spawnable path on Windows, so the dev server cannot start there at all:

```
Command not found: `/bin/sh`. Check the `command`/`runtimeExecutable` field in
.claude/launch.json and make sure it's installed and on PATH.
```

`pnpm` cannot be named directly either -- it ships only `pnpm.cmd` and `pnpm.ps1`, which a no-shell spawn rejects with `ENOENT` and `EINVAL` respectively. The launcher now runs `node`, which resolves on every platform, against a small entry script.

`scripts/dev-launch.mjs` preserves the original behaviour where it applies: `TMPDIR=/tmp` is set on POSIX only, since `/tmp` is meaningless on Windows, and the command runs through a shell so Windows resolves `pnpm.cmd`. It is passed as a single command string rather than command-plus-args, which avoids the `DEP0190` warning the current chain emits.

### The declared port was also stale

The server binds **3100** -- `server/src/config.ts` defaults to it, and `server/src/middleware/board-mutation-guard.ts` hardcodes `http://localhost:3100` and `http://127.0.0.1:3100` in its allowlist -- while `launch.json` declared `3108`, which appears nowhere else in the repository. Forcing `PORT=3108` would run the server outside its own CORS allowlist, so the stale declaration is what needed correcting rather than the port the app binds.

`autoPort` is left as it is on `master`.

### Testing

- Verified the spawn shape with `spawnSync('cmd.exe', ['/c', ...], { shell: false })` -- the same no-shell spawn the launcher performs -- which is how the `pnpm` / `pnpm.cmd` failures above were confirmed rather than assumed.
- Dev server confirmed serving on `127.0.0.1:3100` through the new entry point on Windows 11, with `/api/health` returning `status: ok` and migrations applied.
- `node --check scripts/dev-launch.mjs` passes and `.claude/launch.json` parses.

Not verified on macOS or Linux. The POSIX path is behaviourally unchanged -- `TMPDIR=/tmp` with `pnpm dev` through a shell -- but I have not run it on either.

### DIFF
--- a/.claude/launch.json
+++ b/.claude/launch.json
@@ -3,12 +3,11 @@
   "configurations": [
     {
       "name": "paperclip",
-      "runtimeExecutable": "/bin/sh",
+      "runtimeExecutable": "node",
       "runtimeArgs": [
-        "-c",
-        "TMPDIR=/tmp pnpm dev"
+        "scripts/dev-launch.mjs"
       ],
-      "port": 3108,
+      "port": 3100,
       "autoPort": true
     }
   ]
--- a/scripts/dev-launch.mjs
+++ b/scripts/dev-launch.mjs
@@ -7,9 +7,56 @@ import { spawn } from "node:child_process";
 const env = { ...process.env };
 if (process.platform !== "win32") env.TMPDIR = "/tmp";
 
+const isWindows = process.platform === "win32";
+
 // Passed as one command string rather than command + args, which keeps shell: true
-// from tripping DEP0190.
-const child = spawn("pnpm dev", { stdio: "inherit", env, shell: true });
+// from tripping DEP0190. On POSIX the child gets its own process group so a signal
+// can reach the whole `sh -> pnpm -> node` chain rather than just the shell.
+const child = spawn("pnpm dev", {
+  stdio: "inherit",
+  env,
+  shell: true,
+  detached: !isWindows,
+});
+
+/**
+ * Terminate the child *and its descendants*.
+ *
+ * Signalling the shell alone is not enough: it leaves pnpm, the tsx dev runner
+ * and the embedded PostgreSQL postmaster running, still holding the server and
+ * database ports. The next start then finds a port owned by a process whose
+ * parent is gone, cannot identify it, and falls back to a second postmaster over
+ * the same data directory.
+ *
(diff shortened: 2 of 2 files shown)


qasimsethi-code:
Follow-up on the code review of this PR. Finding: the launcher registered no signal handlers, so terminating it orphaned everything it spawned.

`Ctrl-C`, or a supervisor's `SIGTERM`, killed only the launcher. The shell it spawned -- and pnpm, the tsx dev runner and the embedded PostgreSQL postmaster beneath it -- kept running and kept holding ports 3100 and 54329.

That is not theoretical: it is the dominant failure mode observed while working on this repo today. The next start finds a port owned by a process whose parent no longer exists. Because such a socket accepts a connection but never answers, the ownership probe times out rather than being refused, the caller falls back to another port, and starts a second postmaster over the same data directory -- which PostgreSQL rejects:

```
FATAL:  pre-existing shared memory block is still in use
```

### Fix

Signals are forwarded, and forwarding terminates the **descendants** rather than the shell alone -- signalling the shell is not sufficient, which is the part that makes this more than a two-line change.

- **POSIX**: the child is spawned `detached`, so `process.kill(-child.pid, signal)` signals the whole process group.
- **Windows**: no process-group equivalent, so the tree is walked with `taskkill /pid <pid> /T /F`, with a fallback to a direct kill if `taskkill` is unavailable.
- `SIGBREAK` is included alongside SIGINT/SIGTERM/SIGHUP, since that is what a Windows console sends on Ctrl-Break.
- Re-entrancy guarded, so a second signal during shutdown does not spawn a second taskkill.

### Verification

Confirmed the mechanism the Windows path depends on, rather than assuming it: spawned a shell that spawns a long-lived `node` grandchild, then ran `taskkill /T` against the shell pid.

```
shell pid           : 37760
grandchildren before: 1
grandchildren after : 0
PASS - taskkill /T reaped the tree
```

A direct `child.kill()` on the shell leaves that grandchild running, which is exactly the orphaning being fixed.

Verified on Windows 11. The POSIX process-group path is the conventional approach but I have not exercised it on Linux or macOS.

Issue #11913, opened 2026-08-21
Liveness checks disagree on EPERM: eight process.kill(pid, 0) implementations, one treats a live process as dead


qasimsethi-code:
**What happened?**

`process.kill(pid, 0)` has three outcomes, not two, and the codebase has eight separate implementations of the liveness check that disagree about the third.

- No throw — the process exists and is signalable.
- `ESRCH` — the process genuinely does not exist. The only proof of death.
- `EPERM` — the process **exists** but this user may not signal it. Routine on Windows, where `OpenProcess` returns `ERROR_ACCESS_DENIED` for a process left by an elevated or different-session run, and on POSIX for another user's process.
- Anything else — unknown. Not proof of death.

Current state of the eight sites:

| Location | `EPERM` | Unrecognised errno |
|---|---|---|
| `server/src/services/local-service-supervisor.ts:245` | **dead** | **dead** |
| `cli/src/install-store.ts:154` | alive | dead |
| `cli/src/commands/worktree.ts:2105` | alive | dead |
| `server/src/services/heartbeat.ts:6337` | alive | dead |
| `server/src/services/hot-restart.ts:145` | alive | dead |
| `server/src/services/workspace-instance-cleanup.ts:98` | alive | dead |
| `server/src/services/workspace-operations.ts:106` | alive | dead |
| `cli/src/commands/service.ts:67` | alive | alive |

Only `service.ts` is fail-safe in both columns. Only `local-service-supervisor.ts` gets `EPERM` wrong.

**Expected behavior**

One implementation, used everywhere, that distinguishes alive / dead / unknown and resolves anything short of `ESRCH` toward "still running". #11871 added exactly this as `probeProcessLiveness` in `packages/db/src/embedded-postgres-lock.ts`, with a tri-state `ProcessLiveness` return, because collapsing the third case is what let a live PostgreSQL cluster's `postmaster.pid` be deleted out from under it.

**Steps to reproduce**

`isPidAlive` in `local-service-supervisor.ts`:

```js
export function isPidAlive(pid: number) {
  if (!Number.isInteger(pid) || pid <= 0) return false;
  try {
    process.kill(pid, 0);
    return true;
  } catch {
    return false;      // EPERM lands here
  }
}
```

A false negative is not inert here — it drives destructive cleanup:

- `:218` — a running service whose pid answers `EPERM` is treated as gone; if no port owner is found, `removeLocalServiceRegistryRecord` deletes the registry record for a service that is still running.
[...]

PR #11922, opened 2026-08-22
fix(shared): retry lock owner renames the platform refuses

qasimsethi-code:
`paperclipai worktree init` fails on every attempt on Windows:

```
EPERM: operation not permitted, rename
  ...\.worktree-port-reservations.lock\owner.json.<pid>-<uuid>.tmp
  -> ...\.worktree-port-reservations.lock\owner.json
```

### Cause

`acquireRegistryLock` does three things in order:

1. `writeRegistryLockOwner` creates `owner.json` in the fresh lock directory
2. `startRegistryLockHeartbeat` spawns the worker
3. `writeRegistryLockOwner` rewrites `owner.json` with the worker's probe port

The worker signals ready (`Atomics.store` / `Atomics.notify`) **before** taking its first ownership reading, and `touchOwnedLock` opens `owner.json` with `readFileSync`. So the parent's `Atomics.wait` returns and step 3 renames onto `owner.json` while the worker still has it open.

POSIX replaces an open destination silently, so this is invisible on Linux and macOS -- including in CI. Windows returns `EPERM` and the command dies. Step 1 always succeeds because the directory is fresh and nothing is reading yet; it is the second rename that fails.

### Fix

Two parts:

- The worker completes its first `touchOwnedLock()` before releasing the parent, which closes the deterministic window rather than papering over it.
- Renames of the owner files and of the registry file retry for a bounded window (2s, 10ms apart) on `EPERM`, `EACCES` and `EBUSY`. That covers the heartbeat's ongoing interval reads, and the virus scanners and search indexers that routinely hold newly created files open on Windows.

The reorder alone fixes the reported failure; the retry is defence in depth for the cases that remain genuinely racy.

### What I ruled out

Diagnosis was narrowed by testing rather than assumption:

- **Not permissions or antivirus.** Creating that exact directory and performing the identical `writeFileSync` + `renameSync` from both Node and PowerShell succeeds.
- **Not the `mode: 0o600`.** Same result without it.
- **Not a stale lock.** The directory does not exist before the run and is removed on failure; `owner.json` is never a leftover.
- **Not port selection.** `--server-port`, `--db-port`, `--instance` and `--no-seed` fail identically, because the lock is taken before ports are chosen.

### Testing

- New regression test drives the public lock API with `renameSync` refusing twice. **Verified it fails with the plain rename restored** (`Error: EPERM`), and that no other test in the file fails.
- `packages/shared` suite: 6 passed. `@paperclipai/shared` typecheck clean.
- End to end: `paperclipai worktree init --instance launcher --server-port 3200 --db-port 54400` now completes and reports `Worktree ready`, after four consecutive failures before the change.

Verified on Windows 11. The reordering is platform-neutral and the retry only engages on error codes POSIX does not produce here, but I have not run this on macOS or Linux.

Likely relevant to #11913.

### DIFF
--- a/packages/shared/src/worktree-port-registry.test.ts
+++ b/packages/shared/src/worktree-port-registry.test.ts
@@ -2,7 +2,7 @@ import fs from "node:fs";
 import os from "node:os";
 import path from "node:path";
 import { setTimeout as delay } from "node:timers/promises";
-import { afterEach, describe, expect, it } from "vitest";
+import { afterEach, describe, expect, it, vi } from "vitest";
 import {
   withWorktreePortRegistryLock,
   withWorktreePortRegistryLockSync,
@@ -143,5 +143,29 @@ describe("worktree port registry lock", () => {
     });
 
     expect(fs.existsSync(lockPath)).toBe(false);
+  });
+
+  it("survives a rename the platform briefly refuses", () => {
+    // Windows returns EPERM when renaming onto a file another handle has open,
+    // where POSIX just replaces it. The heartbeat worker reads owner.json on an
+    // interval, so the parent's rewrite failed outright on Windows and never on
+    // Linux, which kept it invisible in CI. Acquisition must ride that out.
+    const homeDir = makeTemporaryRoot();
+    const realRename = fs.renameSync.bind(fs);
+    let refusalsLeft = 2;
+    const spy = vi.spyOn(fs, "renameSync").mockImplementation(((from: fs.PathLike, to: fs.PathLike) => {
+      if (refusalsLeft > 0 && String(to).endsWith("owner.json")) {
+        refusalsLeft -= 1;
+        throw Object.assign(new Error("EPERM: operation not permitted, rename"), { code: "EPERM" });
+      }
+      return realRename(from, to);
+    }) as typeof fs.renameSync);
+
+    try {
+      expect(withWorktreePortRegistryLockSync(homeDir, () => "held")).toBe("held");
+      expect(refusalsLeft).toBe(0);
+    } finally {
+      spy.mockRestore();
+    }
   });
 });
--- a/packages/shared/src/worktree-port-registry.ts
+++ b/packages/shared/src/worktree-port-registry.ts
@@ -18,6 +18,8 @@ const WORKTREE_PORT_REGISTRY_LOCK_STALE_MS = 5_000;
 const WORKTREE_PORT_REGISTRY_LOCK_TIMEOUT_MS = 10_000;
 const WORKTREE_PORT_REGISTRY_LOCK_HEARTBEAT_MS = 1_000;
 const WORKTREE_PORT_REGISTRY_LOCK_PROBE_TIMEOUT_MS = 500;
+const WORKTREE_PORT_REGISTRY_RENAME_TIMEOUT_MS = 2_000;
+const WORKTREE_PORT_REGISTRY_RENAME_RETRY_MS = 10;
 const sleepSyncBuffer = new Int32Array(new SharedArrayBuffer(4));
 
 type RegistryLockOwner = {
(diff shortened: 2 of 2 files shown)


qasimsethi-code:
Flagging something a reviewer should weigh before merging this.

With this fix applied, `paperclipai worktree init` succeeds and a worktree-local instance can be created for the first time on Windows. Starting the dev server in that worktree then crashes Node natively:

```
Assertion failed: !(handle->flags & UV_HANDLE_CLOSING), file src\winsync.c, line 94
```

The process dies immediately and pnpm reports `Command "tsx" not found`, which is misleading fallout from the crashed process rather than a real resolution failure.

### Why this belongs on this PR

The assertion is a libuv async-handle-after-close failure, and `worktree-port-registry.ts` -- the module this PR changes -- is the code that spawns a `Worker` with a `SharedArrayBuffer` control block, `Atomics.wait`/`notify`, and `worker.terminate()`. That is exactly the pattern that produces this assertion.

### What I cannot determine

Whether this change **caused** the crash or merely **exposed** it. There is no before/after baseline: prior to this fix, `worktree init` failed outright with EPERM, so no worktree instance existed and the dev path was never reachable on Windows. The crash may well be pre-existing in the worker lifecycle (`stopRegistryLockHeartbeat` / `terminate`), which this PR does not touch -- but I have no way to demonstrate that from here, and I would rather say so than guess.

### Reproduction

1. On Windows, `git worktree add <path> <branch>`
2. `pnpm install` in the worktree
3. `pnpm paperclipai worktree init --instance <id> --server-port <port> --db-port <port>`
4. `pnpm dev` in the worktree

Step 3 fails without this patch and succeeds with it. Step 4 then crashes as above.

### Suggestion

The EPERM fix and its regression test stand on their own and are verified. But if the worker lifecycle is genuinely unsound on Windows, this PR makes that reachable for the first time, so it may be worth landing alongside a look at the terminate path rather than on its own. Possibly related to #11913.

qasimsethi-code:
Two follow-ups from the code review of this PR.

**The lock-start budget was too tight for its own change.** Moving the worker's first ownership reading ahead of the ready signal put a `readFileSync` and a `utimesSync` inside the parent's fixed 2s `Atomics.wait`. On a network-mounted home directory, or while a virus scanner holds the freshly written owner file, that read can exceed the budget -- and the parent then treats a slow worker as a failed one, tears the lock directory down, and fails acquisition outright. That is a worse failure than the EPERM it replaced. The wait is now a named constant sized to cover the first reading rather than worker startup alone.

**The rename retry budget was per call, not per write.** `writeRegistryLockOwner` rewrites both owner records on every acquisition, so a single write could stall for twice the retry window under sustained contention, consuming a large share of `WORKTREE_PORT_REGISTRY_LOCK_TIMEOUT_MS` before the lock was even held -- turning a transient rename error into a lock timeout, which is the opposite of what the retry is for. Both writes now share one deadline.

Registry tests 6 passed, `@paperclipai/shared` typecheck clean.

qasimsethi-code:
**Retracting the libuv caveat I posted on this PR.** It was wrong.

I reported that with this fix applied, `pnpm dev` in a worktree crashed with:

```
Assertion failed: !(handle->flags & UV_HANDLE_CLOSING), file src\winsync.c, line 94
```

and speculated that this PR made a latent worker-lifecycle problem reachable for the first time.

I have since reproduced that assertion on a tree that **does not contain this change at all** -- a primary checkout on a different branch, with `worktree-port-registry.ts` at its `origin/master` state. The crash appeared during a normal `pnpm dev` boot, alongside the embedded-PostgreSQL port fallback:

```
Failed to start embedded PostgreSQL on port 54330
FATAL:  pre-existing shared memory block is still in use
Assertion failed: !(handle->flags & UV_HANDLE_CLOSING), file src\winsync.c, line 94
```

One counterexample is enough here: the assertion occurs without this patch, so this patch does not cause it. It is pre-existing and unrelated to the EPERM fix.

The honest account of my original report is that I observed a crash on a run that happened to have this change applied and attributed it without a control. The comment should not be relied on.

The EPERM fix and its regression test stand on their own -- verified to fail with the plain `renameSync` restored -- as do the two follow-up fixes in `1a458e7a4`.

PR #11925, opened 2026-08-22
fix(recovery): a review path must be maintained to count as an action path

ericdfields:
## Thinking Path

> - The issue graph liveness subsystem decides if an issue still has a path to its next action
> - An issue that moves to `in_review` must have such a path. The server refuses the transition without one
> - But `classifyIssueReviewPaths` only asks if a path record exists. It does not ask if the path moved
> - A reviewer who goes quiet does not change the issue status. Therefore the entry check never runs again, and the issue continues to report a reviewer that it does not have
> - This pull request gives each path kind an age limit. A path that is too old no longer satisfies the requirement
> - The benefit is that `covered` becomes a true statement about liveness, and a stranded review becomes visible

**What happened**

`reviewAttention.state` reports `covered` for issues that no actor will wake. I measured one board. It had 24 issues in `in_review`. All 24 reported `covered`. But 8 of them had no activity for 4 to 26 days.

| Days without activity | Paths that kept the issue `covered` |
|---|---|
| 26 | `human_reviewer` |
| 25 | `human_reviewer`, `execution_participant`, `interaction` |
| 17 | `human_reviewer` |
| 15 | `human_reviewer`, `execution_participant` |
| 10 | `human_reviewer` |
| 7 | `interaction` |
| 6 | `human_reviewer` |
| 4 | `interaction` |

The `human_reviewer` kind shows the problem most clearly. The server sets it one time. Nothing refreshes it. Nothing wakes the person. It kept a 26 day old issue in the `covered` state for the full 26 days.

**What I expected**

An issue with no actor to wake it must report `stalled`. The recovery subsystem can then escalate it.

**Why it happens**

The action path requirement is an entry condition only. The server does not keep it as an invariant. The paths that the condition accepts do not refresh themselves.

**Related pull request**

I searched the pull request list. #11917 is related but different. #11917 corrects the escalation window: a self stranded finding became older than the 24 hour lookback, and the server never escalated it. This pull request corrects the step before that one. For the worst cases the classifier gave no finding at all, because a stale path suppressed it. Both changes are necessary before a stranded review reaches a person.

Refs #11917

## What Changed

- Add `REVIEW_PATH_STALE_AFTER_MS` in `issue-graph-liveness.ts`. It gives each review path kind an age limit. The limit is 3 days for a wait on a person or on the board (`human_reviewer`, `execution_participant`, `interaction`, `approval`). The limit is 24 hours for work in progress (`active_run`, `queued_wake`).
- Keep `monitor` and `recovery` free of an age limit. They already limit themselves. A monitor uses `maxAttempts` and `timeoutAt`. A recovery issue has its own liveness classification.
- Add a `stale` flag to `IssueReviewPathFact` and to `IssueReviewAttentionPath`. The server continues to report a stale path. The board can then see which reviewer went quiet.
- Name the stale path kinds in the finding reason.
- Add `updatedAt` to `IssueLivenessIssueInput`. The server measures a path that has no timestamp of its own from this field.
- Classify an `in_review` blocker with `reviewFinding` before the `hasExplicitWaitingPath` check. Before this change a blocked chain could hide a stranded review that the same issue reported as stranded at the top level. Two checks disagreed about one invariant.
- Report a stale but resolvable user participant as `in_review_without_action_path`. Before this change it became `invalid_review_participant`, which is for a participant that the server cannot resolve at all.
- Report `state: "stalled"` when all paths are stale, and keep the stale paths in the response.

## Verification

Run the tests:

```
npx vitest run server/src/__tests__/issue-liveness.test.ts
npx vitest run server/src/__tests__/attention-service.test.ts
```

I added 8 cases to `issue-liveness.test.ts`. They cover a stale human reviewer, a recent human reviewer, a stale user participant, an interaction with no answer, a queued wake that never started, a live path next to a stale path, the monitor exemption, and the case with no `updatedAt`.

I also ran a wider check for regressions. 1719 tests in 114 files passed. The files match `issue*`, `*review*`, `*recovery*`, `*attention*`, `*decision*`, and `heartbeat*`.

```
pnpm --filter @paperclipai/shared typecheck
pnpm --filter @paperclipai/ui typecheck
node scripts/check-forbidden-tokens.mjs
node scripts/check-token-gates.mjs
```

All are clean.

## Risks

- **This adds a column and a backfill.** `issues.review_transition_at` records when an issue entered `in_review`. The server sets it on entry and clears it on exit, in the same place it already maintains `blocked_transition_at`. The migration is additive and backfills the rows that are in review at deploy time, because an issue that is already stranded would otherwise stay exempt from staleness until its next transition, and that transition is the one that never comes.
- **Each path kind is aged only by a clock that belongs to it.** A row-backed path (`active_run`, `queued_wake`, `interaction`, `approval`) is measured from its own `createdAt`. The two kinds with no row of their own (`human_reviewer`, `execution_participant`) are measured from `review_transition_at`. The issue's `updatedAt` is deliberately not a fallback for any kind: it moves on every comment, which made it both too generous for a quiet reviewer and too harsh for a live run.
- **The change is compatible with existing callers.** The server never calls a path stale if it has no clock for that path. A caller that supplies neither keeps the previous behaviour.
- **Automatic recovery does not change on a default instance.** `enableIssueGraphLivenessAutoRecovery` is `false` by default. On such an instance this pull request changes only the read model. The server creates no new escalations.
- **The attention feed gets more items.** This is the intended result, but it is a real change. `attention.ts` already admits any `stalled` review, whatever the participant type. This change makes more reviews correctly stalled. Please give an opinion on the 3 day limit. It is the main value judgement in this pull request.
- **A missing timestamp reads as unknown, not stale.** This matters most for a live execution: aging a run off an issue-level clock could escalate work that was still in flight. The rule is now structural rather than a property of any one query, so a caller that omits `createdAt` degrades to "not aged" instead of "dead path".
- **Two test fixtures changed.** The fixtures in `attention-service.test.ts` used a fixed date in the past for `updatedAt`. Their "live participant" cases moved toward the stale state as the calendar moved. They now take the time from the clock. A comment explains why.

## Model Used

Claude Opus 5 (`claude-opus-5[1m]`), 1M context window, extended thinking, with tool use and code execution.

## Checklist


ericdfields:
Both P1s were right, and verifying them turned up that the first one is broader than reported. Fixed in `ce9214b2` — the root cause is shared, so both are closed by the same change.

### "Active runs inherit stale issue time" — confirmed, and it affected every row-backed kind

The mechanism is exactly as described, but it was not conditional on a missing timestamp: `server/src/services/recovery/service.ts` selected **no** timestamps at all. Runs, wake requests, interactions and approvals were all mapped without `createdAt`, so on the recovery path *every* path kind fell through to `issue.updatedAt` — not just active runs, and not just occasionally.

Two changes:

1. The recovery queries now select `createdAt` (`requestedAt` for wake requests) and pass it through, so a run is aged from when the run started.
2. `isReviewPathStale` no longer lets a row-backed kind fall back to an issue-level clock at all. A row-backed path is aged by its own row or not aged, and a missing timestamp reads as *unknown*, which never resolves to stale.

Point 2 is the part that matters for review: point 1 alone would have left the same defect one forgotten `select` away from returning. A live execution can no longer be classified as a dead path regardless of how the caller populates the input.

### "Unrelated activity revives stale reviewers" — confirmed, and it defeated the PR's purpose

`human_reviewer` and `execution_participant` always push `since: null`, so `updatedAt` was their clock, and any comment reset the three-day window. This was worse than a general staleness gap: the issues in this state attract automated re-triage sweeps, and those sweeps comment. The tooling sent to find quiet reviewers was renewing them.

These two kinds genuinely have no row of their own, so the fix is to give them a clock that means something. Added `issues.review_transition_at`, set when an issue enters `in_review` and cleared when it leaves, mirroring the existing `blocked_transition_at` (`server/src/services/issues.ts`). Comments do not move it, so a reviewer who never acts goes stale on schedule no matter how much traffic surrounds them.

I considered deriving "entered review" from `activity_log` instead, to avoid a migration. Rejected it: status changes are only recorded inside a generic `issue.updated` row as a JSONB `details.changes.status` diff, which is unindexed and fragile, and a log is prunable — the wrong substrate for a control-plane invariant.

The migration backfills `review_transition_at` for issues already in review. Without it, an issue that is *already* stranded stays exempt from staleness until its next transition, which is the transition that never comes.

### Tests

Added the two cases you flagged as missing, plus one for the fallback rule:

- a long-running execution stays live however long the issue itself has been quiet
- unrelated activity on the issue does not revive a quiet reviewer
- a row-backed path with no timestamp of its own is never aged off the review clock

I verified these fail against the previous clock rather than passing vacuously: reverting `isReviewPathStale` to the `updatedAt` fallback fails 3 of the 28, including the reviewer-revival case. Full suite green — `issue-liveness`, `attention-service`, `issue-review-attention`, `recovery-classifiers`, `heartbeat-issue-liveness-escalation` (53 + 33 tests), plus server and shared typecheck and both migration checks (`check:migrations` numbering + safety).

ericdfields:
Valid, and worse than the summary reads — thanks. Fixed in `5c8e878c1`.

It isn't only the >24h case. `scheduled_retry` is one of `EXECUTION_PATH_HEARTBEAT_RUN_STATUSES`, so a run parked waiting for its next attempt was aged from `createdAt` like any other row. That row isn't quiet; it's waiting for a time the platform itself set. The classifier would escalate it as a dead path while its own retry was pending — the exact failure this PR exists to prevent, aimed at the wrong rows.

The columns to fix it were already on `heartbeat_runs` and simply weren't selected. An execution path now carries two clocks beyond its creation time:

- **`lastActivityAt`** — `max(startedAt, lastOutputAt, lastUsefulActionAt, updatedAt)`. The output flusher writes `updatedAt` alongside `lastOutputAt` (`heartbeat.ts:15645`), so a run producing output renews itself.
- **`waitingUntil`** — `scheduledRetryAt`. While it is ahead of now the bound is suspended; once it passes the row is silent again and the bound applies unchanged, so a retry that was supposed to fire and didn't is still caught.

One thing worth being explicit about, since it looks like the hole the previous commit closed: this is the run row's clock, not the issue's. Every write to a run row *is* that run doing something, so unrelated comments and re-triage sweeps can't reach it. The `human_reviewer` / `execution_participant` kinds are untouched and still age from `reviewTransitionAt`.

`since` deliberately still reports when the path *began*. It's the board's "waiting since" column — renewing a path must not rewrite it, or a run stuck retrying for a week would read as freshly started on every sweep. Covered by a test.

Four new tests in `issue-liveness.test.ts`, three of which fail without the change (verified by stashing it): a run still producing output past the bound, a retry parked before its due time, a retry whose due time has passed (still flagged), and the `since` assertion. Server typecheck and the four related suites (67 tests) are green locally.

PR #11929, opened 2026-08-22
fix(db): stream JS-fallback backup through gzip to eliminate disk spikes

dzianisv:
## Thinking Path

> - Its control plane depends on the scheduled database backup service to protect all issue, agent, and company state
> - The backup service has a pg_dump path and a JavaScript fallback path; the fallback wrote a full plaintext `.sql` file to disk, then compressed it in a second pass
> - That two-phase write caused multi-GB uncompressed intermediates on disk, and a crash between writer-close and unlink could orphan one permanently, risking an out-of-disk crash of the whole board
> - It needs to be fixed because a dead board (or a disk-full crash mid-write) is worse than a skipped or slower backup cycle
> - This pull request makes the JS fallback stream straight through gzip (no plaintext intermediate ever), adds a preflight free-space gate, sweeps orphaned intermediates on startup and before each run, and logs which engine ran and why
> - The benefit is bounded, predictable disk usage during backups and observability into engine selection, with no more silent multi-GB disk spikes

**What happened?**
The JavaScript-engine fallback in the database backup service wrote the entire SQL dump as plaintext to a `.sql` file first, closed it, and only then read it back and piped it through gzip into the final `.sql.gz`, deleting the plaintext file afterward. On a large database this created a plaintext intermediate many times larger than the final compressed backup (10-18x observed), causing multi-GB disk spikes. If the process died between closing the plaintext writer and deleting the file (OOM, disk-full, crash), the multi-GB intermediate was permanently orphaned with nothing to sweep it.

**Expected behavior**
The JS fallback should stream directly through gzip like the existing pg_dump path already does, so peak disk usage during a backup never exceeds roughly one compressed dump plus a small write buffer. The service should also refuse to start a backup when free disk space is insufficient, clean up any orphaned intermediates from a prior crash, and log which engine ran (and why it fell back) so operators can diagnose disk and engine issues without guessing.

**Steps to reproduce**
1. Configure the database backup service to use the JavaScript engine (`backupEngine: "javascript"`), or force pg_dump to fail so it falls back.
2. Run a backup against a database with enough data to make the dump take more than a few hundred milliseconds.
3. While the backup is running, list the backup directory (`ls -la` / `du -sh`) and observe a large, growing plaintext `.sql` file with no `.sql.gz` sibling yet.
4. Kill the process before it reaches the compress-and-unlink step and observe the plaintext file is left behind permanently.

`master` at the time this PR was opened (see base commit of this branch).

## What Changed

- `packages/db/src/backup-lib.ts`: the JS-fallback engine now writes into a `Gzip` stream piped straight to the final `.sql.gz`; no plaintext `.sql` intermediate is created for either engine.
- Added `assertSufficientBackupFreeSpace()`: a preflight gate that refuses to start a backup (loud log, not an uncaught crash) unless free space on the backup volume is at least 3x the last successful backup's compressed size (configurable via `PAPERCLIP_DB_BACKUP_MIN_FREE_MULTIPLIER`). Last-successful size is tracked in a small JSON state file inside the backup directory.
- Added `sweepOrphanedBackupIntermediates()`: removes stale `*.sql` files that have no `*.sql.gz` sibling (left behind by a crashed prior run), run both immediately before every backup attempt and once at server startup (`server/src/index.ts`).
- Added engine-selection logging: every backup run logs `engine=pg_dump` or `engine=javascript`, and logs the pg_dump error message (no connection string) when pg_dump fails and the code falls back to JavaScript.
- Fixed an `EventEmitter` listener leak introduced by the new gzip backpressure wait (paired `once` listeners for `drain`/`error` were not cross-removed, so long-running dumps accumulated listeners on the `Gzip` stream).
- Exported `sweepOrphanedBackupIntermediates` from `packages/db/src/index.ts` so the server can call it at startup.

## Verification

- `cd packages/db && pnpm exec vitest run src/backup-lib.test.ts` — all 15 tests pass, including four new tests added in this PR:
  - "never creates a plaintext .sql intermediate for the javascript engine, even mid-run" — polls the backup directory every 10ms during a real embedded-Postgres dump and asserts no plaintext `.sql` file is ever observed.
  - "logs which engine ran for the javascript fallback path" — asserts the `engine=javascript` log line is emitted.
  - `assertSufficientBackupFreeSpace` suite — asserts the gate refuses to start when the recorded last-backup size vastly exceeds real free space, and allows the run when no prior size is recorded or free space is ample.
  - `sweepOrphanedBackupIntermediates` suite — asserts a stale orphan with no `.gz` sibling is removed and logged, while a fresh or already-completed dump is left alone.
- `pnpm exec tsc --noEmit` passes clean in both `packages/db` and `server`.
- Manual end-to-end demo run locally against a real embedded Postgres instance: a background poller sampled the backup directory every 100ms during a JS-fallback run and only ever observed the `.sql.gz` file (starting at 0 bytes and growing), never a plaintext `.sql`; the preflight gate was shown refusing a run with a clear log line and then allowing a run once the recorded size was restored to a small value; the orphan sweep was shown removing a manually planted stale `.sql` file with a log line reporting bytes reclaimed.

## Risks

- The preflight free-space gate is new behavior: a backup cycle that previously would have run to (possibly disk-crashing) completion may now be skipped instead, with a logged error. This is intentional — a skipped cycle is recoverable, a disk-full crash is not — but it means an operator with persistently low free disk space will see repeated skipped backups until space is freed, rather than degraded-but-completed backups.
- The backup-size state file is a new small JSON file written into each backup directory; it is best-effort (a write failure only logs a warning, it never fails a successful backup).
- Low risk otherwise: the pg_dump path (already streaming) is unchanged in behavior; the JS-fallback output format (uncompressed SQL statements) is unchanged, only how it reaches disk.

## Model Used

Claude (Anthropic), via the `pi` coding agent harness, extended-thinking/tool-use mode with repository read/edit/bash tool access. Exact model ID not exposed to this session.

## Checklist

- [x] I have searched GitHub for duplicate or related PRs and linked them above (none found for this specific backup disk-spike defect)

### DIFF
--- a/packages/db/src/backup-lib.test.ts
+++ b/packages/db/src/backup-lib.test.ts
@@ -2,9 +2,15 @@ import fs from "node:fs";
 import os from "node:os";
 import path from "node:path";
 import { gunzipSync } from "node:zlib";
-import { afterEach, describe, expect, it } from "vitest";
+import { afterEach, describe, expect, it, vi } from "vitest";
 import postgres from "postgres";
-import { createBufferedTextFileWriter, runDatabaseBackup, runDatabaseRestore } from "./backup-lib.js";
+import {
+  assertSufficientBackupFreeSpace,
+  createBufferedTextFileWriter,
+  runDatabaseBackup,
+  runDatabaseRestore,
+  sweepOrphanedBackupIntermediates,
+} from "./backup-lib.js";
 import { ensurePostgresDatabase } from "./client.js";
 import {
   getEmbeddedPostgresTestSupport,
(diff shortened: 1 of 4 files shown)


dzianisv:
**Merge blocker:** attach a rejection handler when `pipelineDone = pipeline(gzip, fileStream)` is created. A destination write error can currently reject before `close()`/`abort()` awaits the promise, producing an unhandled rejection and potentially crashing the process on the exact disk-full failure this PR is meant to contain. Keep the later awaited error handling, but mark the promise handled immediately.

Current head `24077b91` also has red `verify`/server-shard checks. Push the fix and get the full current head green before requesting merge again.

dzianisv:
@devinfoley @nickyleach — this PR is CLEAN/MERGEABLE with all checks green, but the `dzianisv` identity has pull-only access to this repo (no push, no merge rights — confirmed via `repos/paperclipai/paperclip` permissions and `RequestReviews`/`MergePullRequest` GraphQL both rejecting with a permissions error). Could one of you merge this directly, or grant merge rights so this doesn't require a manual ping on every PR? Tracked at Paperclip AGE-841.

PR #11933, opened 2026-08-22
fix(dev): apply a primary checkout's pinned env to spawned processes

qasimsethi-code:
`bootstrapDevRunnerWorktreeEnv` returned early unless the checkout was a linked git worktree:

```ts
if (!isLinkedGitWorktreeCheckout(rootDir)) {
  return { envPath: null, missingEnv: false };
}
```

So a primary checkout's `.paperclip/.env` was never loaded into the environment the dev runner passes to everything it spawns.

### Why that matters

Every helper that resolves the Paperclip home falls back to `os.homedir()` when `PAPERCLIP_HOME` is unset -- for example `packages/adapter-utils/src/acpx-engine/execute.ts`:

```ts
const home = process.env.PAPERCLIP_HOME?.trim() || path.join(os.homedir(), ".paperclip");
```

A child process that does not inherit the variable therefore builds a **second, empty Paperclip home** beside the real one, and neither is obviously authoritative. Observed on Windows: one `pnpm dev` run applied 225 migrations into a freshly created stray home while the server served the real instance, and repeated boots kept recreating it.

Setting the variable externally does not prevent this. On Windows `setx` only reaches processes launched from a fresh login environment, so anything spawned by an already-running editor or agent never sees it -- which is exactly the case for a dev server started from a tool.

### The change

Apply the env file whenever it exists, for any checkout. Everything specific to linked worktrees is unchanged:

- only a linked worktree treats a **missing** file as fatal (`missingEnv: true`);
- `repairStaleMigratedWorktreeEnvEntries` stays scoped to linked worktrees, because it rewrites paths against the worktree home layout and would be wrong for a primary checkout;
- entries still never override a variable already present, so an explicit `PAPERCLIP_HOME` continues to win.

### Testing

Three tests added to `dev-runner-worktree.test.ts`:

- a primary checkout's pin is applied to the passed environment -- **verified this fails without the change** (restoring the early return makes exactly this test fail);
- an explicit `PAPERCLIP_HOME` still beats the pin;
- a primary checkout with no pin is left untouched and still reports `{ envPath: null, missingEnv: false }`.

`dev-runner-worktree.test.ts` 9 passed. `@paperclipai/server` typecheck clean.

Note: `server/src/__tests__/worktree-config.test.ts` has 8 failures on this branch, but they are **pre-existing** -- I confirmed they fail identically with this change stashed, so they are untouched by it.

Verified on Windows 11. The behaviour is platform-neutral; the `setx` propagation problem that makes it acute is Windows-specific.

### DIFF
--- a/server/src/__tests__/dev-runner-worktree.test.ts
+++ b/server/src/__tests__/dev-runner-worktree.test.ts
@@ -150,4 +150,56 @@ describe("dev-runner worktree env bootstrap", () => {
       missingEnv: true,
     });
   });
+
+  it("applies a primary checkout's pinned env so spawned children inherit it", () => {
+    // Without this the child processes fall back to os.homedir() and build a
+    // second, empty Paperclip home alongside the real one.
+    const root = createTempRoot("paperclip-dev-runner-primary-");
+    fs.mkdirSync(path.join(root, ".git"), { recursive: true });
+    fs.mkdirSync(path.join(root, ".paperclip"), { recursive: true });
+    const pinnedHome = path.join(root, "pinned-home");
+    fs.writeFileSync(
+      resolveWorktreeEnvFilePath(root),
+      `PAPERCLIP_HOME=${pinnedHome}\n`,
+      "utf8",
+    );
+
+    const env: NodeJS.ProcessEnv = {};
+    const result = bootstrapDevRunnerWorktreeEnv(root, env);
+
+    expect(isLinkedGitWorktreeCheckout(root)).toBe(false);
+    expect(result).toEqual({
+      envPath: resolveWorktreeEnvFilePath(root),
+      missingEnv: false,
+    });
+    expect(env.PAPERCLIP_HOME).toBe(pinnedHome);
+  });
+
+  it("lets an explicit environment variable win over a primary checkout's pin", () => {
+    const root = createTempRoot("paperclip-dev-runner-primary-override-");
+    fs.mkdirSync(path.join(root, ".git"), { recursive: true });
+    fs.mkdirSync(path.join(root, ".paperclip"), { recursive: true });
+    fs.writeFileSync(
+      resolveWorktreeEnvFilePath(root),
+      `PAPERCLIP_HOME=${path.join(root, "pinned-home")}\n`,
+      "utf8",
+    );
+
+    const env: NodeJS.ProcessEnv = { PAPERCLIP_HOME: path.join(root, "explicit-home") };
+    bootstrapDevRunnerWorktreeEnv(root, env);
+
+    expect(env.PAPERCLIP_HOME).toBe(path.join(root, "explicit-home"));
+  });
+
+  it("leaves a primary checkout without a pin untouched", () => {
+    const root = createTempRoot("paperclip-dev-runner-primary-unpinned-");
+    fs.mkdirSync(path.join(root, ".git"), { recursive: true });
+
+    const env: NodeJS.ProcessEnv = {};
+    expect(bootstrapDevRunnerWorktreeEnv(root, env)).toEqual({
+      envPath: null,
+      missingEnv: false,
+    });
+    expect(env.PAPERCLIP_HOME).toBeUndefined();
+  });
 });
--- a/server/src/__tests__/dev-runner-worktree.test.ts
+++ b/server/src/__tests__/dev-runner-worktree.test.ts
@@ -157,6 +157,7 @@ describe("dev-runner worktree env bootstrap", () => {
     const root = createTempRoot("paperclip-dev-runner-primary-");
     fs.mkdirSync(path.join(root, ".git"), { recursive: true });
     fs.mkdirSync(path.join(root, ".paperclip"), { recursive: true });
+    fs.writeFileSync(path.join(root, ".paperclip", "config.json"), "{}\n", "utf8");
     const pinnedHome = path.join(root, "pinned-home");
     fs.writeFileSync(
       resolveWorktreeEnvFilePath(root),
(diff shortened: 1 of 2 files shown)


qasimsethi-code:
**Do not merge — this change hangs the dev server on a primary checkout. I found this after opening the PR.**

### What happens

With the patch applied to a primary checkout that has a `.paperclip/.env`, `pnpm dev` stalls immediately after:

```
[paperclip] dev mode: local_trusted (default)
```

It never reaches `building plugin sdk`, never binds the server port, and the log stops at 424 bytes. The process chain stays alive — it hangs rather than crashing. I left it more than four minutes.

### A/B, same machine, same checkout, same cleared state

| | with patch | without patch |
| --- | --- | --- |
| log size | 424 bytes, frozen | 7698 bytes, full banner |
| server port | never bound | 200, listening |
| stray home created | no | yes |

The only variable between runs was this one file. Reverting it restores a normal boot immediately.

### So the fix "works" and is still wrong

It does stop the stray `~/.paperclip` from being created, which was the reported problem. But it trades a cosmetic 1 MB scaffold for a dev server that never starts, which is plainly a bad trade.

### Hypothesis, not a diagnosis

The patch causes `PAPERCLIP_HOME` to be exported to the dev runner and its children on a primary checkout for the first time. My guess is that some startup path branches on that variable being explicitly set and blocks — plausibly worktree port reservation, since `withWorktreePortRegistryLockSync` spawns a heartbeat worker and has known Windows problems (see #11922). If so the patch is not itself wrong so much as newly reaching a path that was previously dead on a primary checkout — which is still a blocker, and arguably surfaces a second bug worth fixing on its own.

I have not confirmed that. The hang needs to be located before this is worth reviewing.

### Suggested handling

Treat as blocked. The immediate problem it targets has a local workaround that does not require this change: a repo-root `.paperclip/config.json` with absolute paths pins the server's home through `findConfigFileFromAncestors`, which stops the second *database* being created. Only the empty scaffold remains, and that is cosmetic.

qasimsethi-code:
**Correction to my previous comment: the A/B I posted was confounded. Please disregard the causal claim that this change hangs the dev server.**

Another session was editing the same working tree while I ran those two boots, and the tree changed between them. File modification times against my run timeline:

| file (not part of this PR) | modified | relative to my runs |
| --- | --- | --- |
| `packages/db/src/migrate.ts` | 12:11:43 | 20s **before** the patched boot started (12:12:03) |
| `server/src/dev-runner-child-watchdog.ts` (new) | 12:18:14 | **between** the two runs |
| `scripts/dev-runner.ts` | 12:20:26 | after both |

A `dev-runner-child-watchdog` appearing between the failing run and the passing run is decisive: it targets exactly the hang I observed. My "unpatched" boot may have succeeded because that landed, not because this patch was absent. Equally, the patched boot may have hung because of the in-flight `migrate.ts` edit that landed 20 seconds before it started.

### What still stands

- **Observation:** one boot with the patch hung at 424 bytes; one boot without it reached a healthy server. Both are real measurements.
- **Observation:** the stray `~/.paperclip` was absent across the patched boot and reappeared on the unpatched boot, which is consistent with the patch doing what it claims.

### What does not stand

- **Any causal claim that this patch causes the hang.** I asserted that with a comparison table, and the comparison was not controlled. I should have checked `git status` before treating two runs minutes apart as an experiment.

### Handling

Leaving this as a draft regardless. It needs a clean retest in a tree nobody else is modifying, ideally after the dev-runner work in flight has settled, before it is worth reviewing. I would rather it sit in draft on unproven suspicion than be merged on my unverified assurance — but the specific accusation in my previous comment should not be relied on.

qasimsethi-code:
Follow-up from the code review of this PR.

Applying `.paperclip/.env` on a primary checkout made a lone env file authoritative. That file is often **debris** -- left behind when a worktree's isolated instance was removed with `worktree:cleanup`, or carried along when a worktree directory was copied into a clone. Its `PAPERCLIP_HOME` and `PAPERCLIP_CONFIG` then point at an instance that no longer exists, and the dev runner exports them to every process it spawns, so the server and the migration child either fail outright or quietly initialise a fresh empty instance at the dead path -- the same split-home failure this PR exists to prevent, arriving by a different route.

Linked worktrees were never exposed to this: `repairStaleMigratedWorktreeEnvEntries` already rewrites exactly that case for them. A primary checkout has no such repair, so it now requires the config file the env file accompanies to exist beside it. Without that, the checkout is treated as unpinned -- the behaviour it had before this branch.

Tests: the two primary-checkout cases now write a `config.json` alongside, and a new case asserts an orphaned `.env` is ignored and leaves `PAPERCLIP_HOME` unset. 10 passed, `@paperclipai/server` typecheck clean.

qasimsethi-code:
**Retracting the causal claim that this change hangs the dev server**, and reporting a controlled retest.

Earlier I posted a comparison suggesting this patch caused a boot hang, then corrected it as confounded (another session was editing the tree between my two runs). I have now rerun it properly: tree verified quiet for 10 minutes, `git status` and `git diff` hashes recorded before and after each arm, ports and processes cleared to an identical starting state.

| | run A -- unpatched | run B -- patched |
| --- | --- | --- |
| server port | never bound | **200, listening** |
| embedded PostgreSQL | fell back to 54330, `pre-existing shared memory block is still in use` | started clean on 54329 |
| libuv assertion | **yes** | no |
| migrations | never reached | `already applied` |

So the failure I originally blamed on this patch reproduces **without** it, and the patched arm was the one that booted. The causal claim was wrong in both directions.

### What this does not establish

That the patch fixes anything. This is one run per arm, and the underlying failure is a **race**: the migration child starts a postmaster on the configured port, the server then probes, finds it held, and falls back to another port against the same data directory. Whether that collides is timing-dependent -- an earlier unpatched run on this machine booted fine. Run B passing is consistent with the patch helping and equally consistent with luck.

One caveat on the retest itself: run B's baseline was not byte-identical to run A's, since applying the patch necessarily changes the tree. The status hash moved from `e4ef107a` to `68bb8481` with one file written, which is the patch plus a runtime artifact rather than external interference -- but it is not a perfectly matched pair and I would rather say so.

Staying a draft. It is no longer suspected of breaking anything, but "it worked once" is not grounds to mark it ready. The port-fallback race above is worth fixing on its own account, and is reported on #11871.

PR #11935, opened 2026-08-22
Drop the mission step from the wizard arc

tonio-alucema:
## Thinking Path

> - New customers arrive through an onboarding arc that spans Paperclip Cloud and the tenant app
> - Cloud's naming screen stopped asking for the company mission, but the tenant wizard still decided its first step by asking whether the company had one
> - Every Cloud-created company therefore looked mission-less on arrival, so every walk detoured through a "Define your mission" screen the design had already removed
> - This pull request removes that step from the arc, and makes step 1 create the company itself
> - The benefit is a shorter arc that matches the design, and a three-step progress strip that now counts the steps that exist

No public issue exists. The problem was found by walking staging end to end.

**What happened:**
A new customer signs in, names their organization, and waits for it to build. The tenant wizard then asks "Define your mission" before it asks for the first agent. Cloud no longer collects a mission, so this screen appears for every new customer.

**Expected behavior:**
The wizard asks for the first agent, the model, and a review. The progress strip counts three steps.

**Actual behavior:**
The wizard asks for the mission first. The progress strip counts five segments, because the run does not enter on the agent arc.

**Additional context:**
Three merged pull requests built the mission-based step choice this change removes: #11352, #11416 and #11429. The mission is now collected later, inside the tenant app, so onboarding does not ask for it at all.

## What Changed

- `onboardingStepForCompany` always returns the agent step. The `companyHasMission` parameter is removed, because it cannot change the answer.
- `resolveRouteOnboardingOptions` no longer accepts `companyHasMission`.
- The dashboard no longer waits for the goal lookup before it opens the wizard. That wait only chose a step, and the step is now fixed.
- Step 1 creates the company in a new `handleCreateCompany`. Company creation used to sit at the end of `handleConfirmMission`.
- No company goal is written during onboarding.
- The three-step strip now shows on the agent, model and review steps, because every Cloud-first run enters on the arc.
- The full-length bar drops its second segment. No run can fill it.
- The grow path keeps its step 2 questionnaire. Only the create path skips ahead.
- Back from the agent step goes to the screen the run came from.
- Four end-to-end specs no longer drive the wizard through the mission step.

## Verification

Run the tenant test suite:

```
cd ui && npx vitest run
```

- 4356 tests pass. 471 files pass.
- `npx tsc --noEmit` reports no errors.
- Fault injection: forcing `skipsMissionStep` to `true` fails the grow questionnaire test. Removing the Back rule fails the Back test. Both tests fail on the exact defect they guard.
- The three-step strip is asserted by an existing test. It checks `Step 1 of 3` and `aria-label="Create your first agent"`.

Manual check on staging after the paired Cloud change:

1. Open a new incognito window.
2. Sign in with a new account.
3. Name the organization.
4. Confirm the wizard shows "Create your first agent" and "Step 1 of 3".

## Risks

- **Behavioral change.** Onboarding no longer writes a company goal. An agent hired during onboarding starts without a seeded mission. This is intended. The mission moves to the tenant app.
- **Dead code.** `ONBOARDING_MISSION_STEP` and the mission screen stay in the codebase, but nothing in the app opens them. They wait for the surface that collects the mission later.
- **Grow path.** The grow path is unchanged, but it shares step 2 with the removed screen. New tests cover it.
- **Superseded work.** #11352, #11416 and #11429 tuned the mission-based step choice. This change removes the branch they tuned.

## Model Used

Claude Opus 5 (`claude-opus-5`), extended thinking, with tool use and code execution through Claude Code.

## Checklist

- [x] My branch name describes the change and contains no internal Paperclip ticket id

### DIFF
--- a/tests/e2e/conference-room-typing-intro.spec.ts
+++ b/tests/e2e/conference-room-typing-intro.spec.ts
@@ -18,7 +18,6 @@ import {
  * exactly that failing condition.
  */
 
-const MISSION = "Verify the first-task launch survives the wizard handoff.";
 const FIRST_TASK_TITLE = "Paperclip onboarding";
 
 /**
@@ -77,9 +76,7 @@ async function runOnboardingWizard(page: Page, companyName: string) {
   await page.getByPlaceholder("Acme Corp").fill(companyName);
   await page.getByRole("button", { name: /^Next/ }).click();
 
-  // Step 2: mission (direct path default).
-  await page.getByPlaceholder("What is your team trying to achieve?").fill(MISSION);
-  await page.getByRole("button", { name: /Confirm mission/ }).click();
+  // Step 1's "Next" creates the company; the mission step no longer runs.
 
   // Step 3: the lead's role, then its name. The role gates "Next", and
   // choosing one fills the name — so the walk only types here to override it.
--- a/tests/e2e/nux-phase4-screenshots.spec.ts
+++ b/tests/e2e/nux-phase4-screenshots.spec.ts
@@ -72,17 +72,8 @@ test.describe("NUX Phase 4 visual QA", () => {
     await page.screenshot({ path: shot("02-create-name.png") });
 
     await page.getByRole("button", { name: /^Next/ }).click();
-    await expect(
-      page.getByRole("heading", { name: "Define your mission" }),
-    ).toBeVisible({ timeout: 10_000 });
-    await page
-      .getByPlaceholder("What is your team trying to achieve?")
-      .fill("Build affordable home robots that handle household chores.");
-    await page.screenshot({ path: shot("03-create-mission.png") });
-
-    // Step 2 advances via "Confirm mission" (creates the company + goal);
-    // step 3 is the team-lead naming step of the capsule wizard.
-    await page.getByRole("button", { name: /Confirm mission/ }).click();
+    // Step 1's "Next" creates the company and goes straight to the team lead.
+    // The mission screenshot that sat here is gone with the step it captured.
     await page.waitForSelector("#onboarding-agent-role", {
       timeout: 30_000,
     });
@@ -155,7 +146,6 @@ test.describe("NUX Phase 4 visual QA", () => {
     for (const f of [
       "01-front-door.png",
       "02-create-name.png",
-      "03-create-mission.png",
       "04-hire-team-lead.png",
       "05-growth-intake.png",
       "06-board-chat.png",
--- a/tests/e2e/onboarding.spec.ts
+++ b/tests/e2e/onboarding.spec.ts
@@ -6,14 +6,13 @@ import { test, expect } from "@playwright/test";
  * The wizard now opens on a front door (path picker) and the "Create a new
  * company" path runs:
  *   Step 0  — Front door (Create a new company / Level up existing)
- *   Step 1a — Name your organization
- *   Step 1b — Define your mission (direct or guided)
+ *   Step 1a — Name your organization (creates the company)
  *   Step 2  — Hire your team lead (adapter picker)
  *   Step 3+ — Launch celebration → CEO chat → hiring plan → orientation
  *
  * This test covers the deterministic, LLM-free core: it drives the front door
- * through company naming + mission definition (which creates the company and a
- * company-level goal) and verifies the wizard advances to the team-lead step.
+ * through company naming (which creates the company) and verifies the wizard
+ * advances to the team-lead step without asking for a mission.
  *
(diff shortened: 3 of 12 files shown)


tonio-alucema:
Both of Greptile's findings were valid and are fixed in `f8949f88`.

**Issue 1 — the grow questionnaire.** Correct, and the miss was mine: step 2 is two different screens wearing one number. The create path's step 2 asked for the mission; the grow path's step 2 is "tell us about your team", and its answers seed the lead agent. Routing step 1 straight to the agent step skipped both. Only the create path skips ahead now.

**Issue 2 — residual navigation into step 2.** Also correct, three of them. I had changed the keyboard shortcut handler and the footer button, but not the name field's own `onKeyDown`, which still called `setStep(2)` directly.

Back was the one worth thinking about rather than patching. It stepped to `step - 1`, which walks a create run into a screen it never saw. It now asks where the run actually came from:

| run | Back from the agent step |
|---|---|
| create | step 1 — it skipped the mission screen |
| grow | step 2 — its questionnaire |
| entered *on* the mission step | step 2 — it has seen that screen, so Back owes it the way back |

That last row is not hypothetical: an existing test opens the wizard there explicitly and walks Back, and my first version of the rule broke it. The rule reads `entryStep !== 2` for exactly that reason.

**On coverage.** You noted the tests did not cover plain Enter, Back, or the grow path. They do now — three tests, and I fault-injected both defects to confirm they fail on the real thing rather than describing the fix. Forcing `skipsMissionStep = true` fails the grow test; removing the Back rule fails the Back test.

They live in `OnboardingWizard.test.tsx` rather than the step test file, because the step file does not mock `companies-query`, so the restore gate (`rawBlob !== undefined && companiesQuery.isFetching`) never releases and the wizard never mounts. Worth knowing if you look for them where the other step tests are.

4356/4356 tenant tests, typecheck clean.

tonio-alucema:
Valid, and a regression this branch introduced. Fixed in `e9a3dd50`.

One correction to the mechanism, because it changes where the fix belongs: the step-level handler requires `metaKey || ctrlKey`, so plain Enter never reaches it. The collision is on **Cmd+Enter** — the name field's own handler does not check for a modifier, so that keystroke satisfies both.

The rest is exactly as described. `if (loading) return` in the step-level handler cannot catch it: `setLoading(true)` has not landed while the same event is still bubbling, so the second caller reads a value the first has not written. The comment directly above that guard already warns about this shape — "two goals for one mission, two agents for one hire" — it just did not cover a second handler on the same event.

Harmless before this branch, because both handlers only called `setStep(2)`. Once step 1 started creating the company, twice meant two companies for one keystroke, one stranded outside onboarding.

**Guarded on `defaultPrevented` rather than `stopPropagation`.** Your suggestion works, but it fixes this one field. The step-level handler should ignore any key something nearer already dealt with — and the field was already calling `preventDefault()`, so the signal was there to read.

Fault-injected: removing the guard fails the new test with `expected "vi.fn()" to be called 1 times, but got 2 times`. It reproduces the duplicate exactly rather than describing the fix.

4357/4357 tenant tests, typecheck clean.

Unrelated, for the record: `General tests (server (2/5))` failed on the previous commit with a `plugin-worker-manager` timeout and a `heartbeat_run_events` foreign-key violation. This branch touches only `ui/` and `tests/e2e/` — zero server files — so those cannot come from this diff.

tonio-alucema:
Valid, and a genuinely different race from the last one. Fixed in `59d4e8ab`.

The previous fix closed **one event reaching two handlers**. This is **two events**: holding Enter down fires keydown repeatedly, and `defaultPrevented` on one event says nothing about the next.

Neither existing guard can answer "already creating":

| guard | why it cannot |
|---|---|
| `loading` | state — the second caller reads what the first has not written yet |
| `createdCompanyId` | not set until the request it guards has resolved |

A ref is written before the request goes out, so the caller behind it sees it. That is what guards creation now.

Fault-injected: four repeats produce `expected "vi.fn()" to be called 1 times, but got 4 times` without the ref, and one call with it.

**Worth flagging beyond this PR**: `handleConfirmMission` and `handleGiveHeartbeat` have the same exposure — both guard on state or on a value set only after their request resolves. They are pre-existing and reachable only via Cmd+Enter, so I have not touched them here, but the comment above the step-level `loading` guard ("two goals for one mission, two agents for one hire") describes precisely the defect that guard cannot actually prevent. Worth a separate look.

4358/4358 tenant tests, typecheck clean.

For the record on the previous run: `e9a3dd50` finished 28 pass / 1 fail, the one being this review. The server shard and every e2e job passed, which settles the earlier `plugin-worker-manager` timeout as noise rather than anything from this branch.

PR #10738, opened 2026-08-03
fix(cli): require confirmation for workspace deletes

MindSyncHub:
## Thinking Path

> - Paperclip is the open source control plane people use to manage AI agents for work.
> - The CLI is an operator-facing control surface for that control plane.
> - Several CLI delete commands already require an explicit `--yes` flag before sending destructive API requests.
> - Workspace and environment operations are part of the execution/runtime layer, so deleting those resources should have the same local guardrail as project, goal, issue, agent, secret, and company deletes.
> - The gap was that `environment delete` and `project-workspace delete` sent DELETE requests immediately when invoked.
> - This pull request makes those two destructive commands require `--yes` before any API call is made.
> - The benefit is a small, reviewer-friendly safety consistency improvement for CLI operators without changing server API behavior.

## Linked Issues or Issue Description

Bug report:

### Pre-submission checklist

- [x] I have searched existing open and closed issues and this is not a duplicate.
- [x] I am on the latest available `master` checkout used for this PR.
- [x] I have confirmed the error originates in Paperclip itself, not in an agent adapter, API provider, or local configuration.

### What happened?

`paperclipai environment delete` and `paperclipai project-workspace delete` did not require the local `--yes` confirmation used by neighboring destructive CLI commands. Invoking either command sent a DELETE request immediately.

### Expected behavior

Destructive CLI deletes should stop locally unless the operator passes an explicit confirmation flag.

### Steps to reproduce

1. Run `paperclipai environment delete <environment-id>`.
2. Run `paperclipai project-workspace delete <project-id> <workspace-id>`.
3. Observe that both commands send DELETE requests without requiring a local confirmation flag.

### Paperclip version or commit

Reproduced from the PR base checkout at `a6436126ce3ce65b77588063e6fb3d9b01e197aa` (`origin/master` in this worktree before the fix).

### Deployment mode

Local dev (`pnpm dev`) / CLI source checkout.

### Installation method

Built from source (`pnpm dev` / workspace CLI).

### Agent adapter(s) involved

Not adapter-specific (core CLI bug).

### Database mode

Not database-related.

### Access context

Board (human operator) CLI usage.

### Node.js version

Node.js 20+ per repository development requirements.

### Operating system

macOS local worktree during verification.

### Relevant logs or output

Before this PR, the unconfirmed delete commands reached `fetch`. The regression test added here asserts unconfirmed deletes print `Deletion requires --yes.` and make no network call.

### Relevant config (if applicable)

Not config-related.

### Additional context

[...]

Issue #10555, opened 2026-07-31
Data loss: workspace cleanup wiped 2309 uncommitted files in a shared-cwd multi-agent company (uncommitted variant of #3207)


polyakoff:
### Pre-submission checklist

- [x] I have searched existing open and closed issues and this is not a duplicate.
- [x] I am on the latest released version of Paperclip (or can reproduce on `master`).
- [x] I have confirmed the error originates in Paperclip itself — not in my agent adapter, API provider, or local configuration.

### What happened?

**TL;DR** — In a 12-agent company where every agent shares one `cwd` (the default), a 32-minute window ended with **2309 uncommitted working-tree files erased from disk**. No stash, no branch, no reflog entry, no run error, no board alert. This is the failure mode of #3207, but for *uncommitted* working-tree state instead of unpushed commits — so no failed `git push` is needed to trigger it. The work survived only because we run our own snapshot job outside Paperclip.

Two undocumented instance-settings flags, `enableWorkspaceDirtyQuarantineRepair` and `enableWorkspaceBranchReconcileForward`, are `true` by default on our install and are **not shown on the Experimental settings page** — so an operator who suspects them has no supported way to turn them off.

---

### Setup

- Paperclip 0.3.1, authenticated / private deployment, Docker Compose, Postgres in a sibling container.
- One company, 12 agents, adapters `claude_local`, `codex_local`, `gemini_local`.
- Every agent has the same `adapterConfig.cwd` — a single git repository bind-mounted into the server container. This is the default arrangement described in #3335.
- `enableIsolatedWorkspaces = false`, `enableWorktreeRunExecution = false` (defaults).
- The agents work by editing many files under that cwd; commits happen periodically, so a large uncommitted delta between commits is normal.

### What we observed

- Window 2026-07-30 23:10–23:42 UTC. Before it: ~2300 modified/new files in the working tree. After it: tree clean at `HEAD` and the files gone from disk — not unstaged, removed.
- The removal left **no git trace**: no `git stash` entry anywhere near that window, no ref created, no commit recording the deletion, nothing in run logs, and no error or warning surfaced on the board. From the UI the period looks uneventful.
[...]

PR #6965, opened 2026-05-29
feat: add execution workspace reaper

lam-source:
## Thinking Path

> - Paperclip orchestrates AI agents for zero-human companies.
> - Execution workspaces are the persisted bridge between issues, runtime services, filesystem checkouts, and cleanup.
> - Closed or orphaned issue work can leave execution workspace rows and local files behind.
> - The safe first step is a metadata-only reaper that identifies candidates without exposing logs, config, or file contents.
> - Actual archive mode must exclude active-linked work, preserve an audit trail, and avoid deleting files unless close-readiness allows it.
> - This pull request adds that conservative API, service, CLI command, and regression coverage.
> - The benefit is repeatable workspace hygiene with dry-run proof before mutation.

## What Changed

- Added shared reaper report types and a strict request schema.
- Added `executionWorkspaceReaperService` with dry-run default, active-linked exclusions, candidate counts, atomic archive guarding, activity-safe metadata, optional post-archive filesystem cleanup gated by close-readiness, and retryable `cleanup_failed` handling.
- Added `GET` and `POST /api/companies/:companyId/execution-workspaces/reap`; dry-run reads use company-scoped `GET`, while mutating apply uses board-only `POST`.
- Added `paperclipai execution-workspace reap` with `--apply`, `--delete-files`, and JSON support; default dry-run now calls `GET`, while explicit `--apply --dry-run` is rejected before any API call.
- Added service, route, and CLI tests for dry-run classification, archive mutation, active-linked exclusion, audit logging, cleanup warning sanitization, cleanup failure retry, board-only mutation authz, CLI method safety, contradictory flags, and delete-files skips for dirty/untracked/unmerged/primary/shared workspaces.

## Verification

- `corepack pnpm exec vitest run cli/src/__tests__/execution-workspace-command.test.ts server/src/__tests__/execution-workspaces-service.test.ts server/src/__tests__/execution-workspaces-routes.test.ts` passed: 3 files, 22 tests.
- `corepack pnpm --filter @paperclipai/plugin-sdk ensure-build-deps` passed.
- `corepack pnpm --filter @paperclipai/server exec tsc --noEmit` passed.
- `corepack pnpm --filter paperclipai exec tsc --noEmit` passed.
- `git diff --check` passed.
- Earlier branch verification also passed targeted shared/server/CLI builds and token check, while a full broad Vitest run hit existing unrelated environment/test-infra failures documented in the issue PR note.
[...]

Issue #2022, opened 2026-03-29
Add retention policies for run logs, server log rotation, and workspace cleanup


sherifkozman:
## Problem

Local Paperclip instances accumulate data without cleanup policies for several directories:

### 1. Run logs (`data/run-logs/`) — No retention
- Every heartbeat run creates an NDJSON log file
- Accumulates indefinitely (currently 1,060 files / 302 MB after ~12 days)
- **Ask:** Add `runLogRetentionDays` config (default 14-30) with pruning on the backup timer

### 2. Server log (`logs/server.log`) — No rotation
- Single flat file, never rotated
- Currently 889 MB / 285K lines
- **Ask:** Log rotation (daily or size-based, e.g. 100MB) with retention (keep last N files)

### 3. Agent workspace temp files — No cleanup
- Agents dump scratchwork (JSON API responses, temp Python scripts, `.ruff_cache`, generated PNGs) into managed project workspaces (`projects/{companyId}/{projectId}/_default/`)
- These are never cleaned up
- **Ask:** Either (a) instruct agents to clean up after themselves in the heartbeat procedure, or (b) add periodic workspace cleanup for files not referenced by issue attachments

### 4. DB backups — Works but aggressive default
- Hourly backups with 30-day retention works correctly
- For local dev instances, hourly at ~60MB each = ~1.4 GB/day
- **Ask:** Consider a lighter default for local mode (e.g. every 6 hours) or expose `intervalMinutes` in instance settings UI

### Current accumulation (12 days of use)

| Directory | Size | Files |
|---|---|---|
| `data/backups/` | 4.5 GB | 175 SQL dumps |
| `logs/server.log` | 889 MB | 1 file |
| `data/run-logs/` | 302 MB | 1,060 NDJSON files |
| `data/storage/` | 176 MB | Attachments (expected) |
| `projects/` | 47 MB | 102 temp files |
| **Total** | **6.1 GB** | |

### Suggested config additions

```typescript
{
  backup: {
    intervalMinutes: 360,     // 6h for local (current: 60)
    retentionDays: 30         // existing
  },
  runLogs: {
    retentionDays: 14         // new
  },
  serverLog: {
    maxSizeMB: 100,           // new — rotate at 100MB
    maxFiles: 5               // new — keep 5 rotated files
  }
}
```

carrgust:
Adding a specific datapoint from a v2026.513.0 production install (single-tenant, Hetzner):

**Observation 2026-05-14:**
[...]

PR #11936, opened 2026-08-22
feat: reap stale node_modules/worktrees to keep host disk under control

dzianisv:
## Thinking Path

> - A self-hosted local instance keeps every agent's execution workspace as a real git checkout on disk, under `instances/*/workspaces/`.
> - These checkouts accumulate `node_modules` and whole worktrees forever; nothing in the repo currently reclaims them, and a long-running instance can run its host out of disk.
> - A naive "just delete inactive-looking directories" cleanup is actively dangerous here — two real incidents (#3207, #10555) show a cleanup step destroying unpushed or uncommitted agent work with no recovery path.
> - This pull request adds a fail-closed, two-tier reclaim tool: a default tier that only ever removes the untracked, always-reinstallable `node_modules` directory, and a strictly-gated opt-in tier for removing an entire worktree once its branch is provably merged/closed and unreachable.
> - It also adds a `launchd`-based (not cron) installer so the tool can run unattended on a schedule, the same way Paperclip's own local instance already schedules background maintenance.
> - The benefit is safe, automatic, reversible-by-design disk reclaim for exactly the kind of workspace bloat that fills a host's disk over weeks of agent activity, without the failure modes of a blunter cleanup.

Searched for duplicates/related work first:
- Refs #2022 — the existing, broader "retention policies for run logs, server log rotation, and workspace cleanup" request. This PR implements the workspace-cleanup piece of that request specifically for `node_modules`/worktree reclaim; run-log and server-log rotation remain out of scope here and are still tracked by #2022.
- Related but not duplicate: #10738 adds a confirmation prompt to the server-side `execution-workspace delete` API/CLI command (deleting a workspace *row* through the board). This PR is a purely host-local filesystem tool with no server/API/DB involvement at all — different layer, no overlap.
- Related but not duplicate: #6965 proposes a server-side `execution-workspace reaper` service/API/CLI that archives `execution_workspace` **database rows** (with an optional, close-readiness-gated file delete) via an authenticated company-scoped endpoint. It has been open since May with no merge, and its core service file is not present on `master`. This PR does not touch the database or any API at all; it is a local, `lsof`/`git`-only tool an operator (or a scheduled LaunchAgent) runs directly against the filesystem. The two are complementary, not overlapping: one is DB-row hygiene through the API, the other is raw disk reclaim on the host.

No existing public issue describes this specific "reclaim `node_modules`/worktrees to keep host disk under control" gap, so per the feature-request template:

**Subsystem affected**
Cross-cutting — a new top-level `scripts/` tool plus its `launchd` installer and docs (`docs/deploy/`); no changes to `server/`, `ui/`, `packages/db`, or `packages/shared`.

**Problem or motivation**
A self-hosted Paperclip instance's `instances/*/workspaces/` directory grows without bound: every agent-per-issue checkout leaves its own `node_modules` (and, less often, a whole abandoned worktree) behind indefinitely. Nothing in the repo reclaims this today, and on a long-running host it is a direct path to 100% disk. A blunt cleanup is not safe here — see #3207 and #10555, two real incidents where deleting "unused-looking" workspace directories destroyed unpushed or uncommitted work with no way to recover it.

**Proposed solution**
`scripts/reap-stale-workspaces.mjs`, a dependency-free Node script with two safety tiers, both fail-closed on any ambiguity, git/gh error, or missing signal:

1. `node_modules`-only reclaim (default with `--apply`). `node_modules` is never git-tracked, so deleting it can never lose a commit or an edit. Requires: no active session (checked live via `lsof`, never inferred from run/issue status), a recognized lockfile committed in the repo, that lockfile itself has no uncommitted edits (a lockfile can be tracked and still mid-edit — found on a live host, not just in theory), and the worktree's `HEAD` did not commit within a configurable recent-activity window (default 60 min; `lsof` alone cannot see an agent that is between tool calls with no open file handle — also found on a live host, a worktree with a 24-minute-old commit and zero `lsof` matches).
2. Full worktree removal, opt-in via `--remove-merged-worktrees`. Additionally requires a fully clean, fully-pushed tree and a PR that is `MERGED` or `CLOSED` — never while `OPEN` — and is never used on a shared worktree parent.

Every candidate path is `realpath`-resolved and checked against an allow-list of discovered worktree roots before anything is touched, including through a symlink. Discovery covers both `~/.paperclip*/instances/*/workspaces/<agent>/<checkout>` and the older `~/.paperclip-worktrees/<checkout>` convention.

`scripts/install-paperclip-retention-agent.sh` wires this into a per-user `launchd` LaunchAgent (`--install` / `--uninstall` / `--status`), never an OS crontab entry, matching how the rest of a local instance already schedules background maintenance.

**Alternatives considered**
- A blunt "delete anything not touched in N days" sweep — rejected outright given #3207/#10555.
- Requiring every reclaim (including `node_modules`-only) to also require a merged/closed PR — rejected as overly conservative: `node_modules` never carries git history, so gating its removal on PR/branch state provides no additional safety and would leave most of the disk-usage problem unreclaimed for issues still under active review.
- Doing this server-side through the `execution_workspace` DB model (as #6965 explores) — rejected for this PR's scope: the disk-pressure problem lives entirely on the host filesystem and does not require (or benefit from) a DB round-trip; a local script an operator or LaunchAgent can run directly is simpler and has fewer failure modes for this specific problem.

**Roadmap alignment**
Not roadmap core-feature work; this is host/ops tooling for self-hosted disk hygiene.

## What Changed

- Added `scripts/reap-stale-workspaces.mjs` — the fail-closed, two-tier reaper (dry-run by default, `--apply` to act, `--remove-merged-worktrees` for the opt-in full-worktree tier, `--recent-commit-minutes` to tune/disable the recency guard, `--root`/`--max-depth`/`--json` for scripting).
- Added `scripts/install-paperclip-retention-agent.sh` — idempotent `launchd` LaunchAgent installer/uninstaller/status tool; never touches `crontab`/`/etc/cron*`; resolves and bakes an absolute `node` path at install time (a `launchd` job's minimal `PATH` cannot resolve a bare `node` at run time — this was caught by actually `launchctl kickstart`-triggering the installed job on a real host, not just by reading the generated plist).
- Added `scripts/__tests__/reap-stale-workspaces.test.mjs` (24 tests) and `scripts/__tests__/install-paperclip-retention-agent.test.mjs` (9 tests), plus hermetic fixture stubs under `scripts/__tests__/fixtures/reap-stale-workspaces/`.
- Wired both new test files into `.github/workflows/pr.yml`'s `policy` job (this repo's test files are not auto-discovered — each must be added explicitly).
- Added `docs/deploy/disk-retention.md` (registered in `docs/docs.json`), documenting both safety tiers, the recency and lockfile-dirty guards, idempotency, and the LaunchAgent wiring.
- Added `package.json` scripts: `reap:stale-workspaces`, `test:reap-stale-workspaces`, `test:install-paperclip-retention-agent`.

## Verification

- `node --test scripts/__tests__/reap-stale-workspaces.test.mjs scripts/__tests__/install-paperclip-retention-agent.test.mjs` — 41 tests, 0 failures, including CLI end-to-end tests against real git fixtures (dry-run, `--apply`, active-session preservation via a stub `lsof` probe, merged-worktree full removal, open-PR-blocks-full-removal-but-not-`node_modules`, unpushed-commits-block-removal regression guard for #3207, dirty-lockfile guard, recent-commit guard, fail-closed path containment, symlink-escape protection, idempotency, and installer idempotent re-install/uninstall/report-only-mode).
- Verified end-to-end against a real, disk-constrained, live self-hosted instance: a dry run correctly classified every real workspace (preserving an actively-committing worktree via the new recency guard, and a worktree with an uncommitted lockfile edit via the new lockfile-dirty guard); `--apply` then reclaimed real disk from the genuinely-inactive worktrees while leaving every git worktree's `HEAD`/branch untouched; a second `--apply` run reclaimed zero additional bytes (idempotent); the installed LaunchAgent was `launchctl kickstart`-triggered for real and completed successfully end to end, producing the same classification as the manual dry run.

## Risks

- Low risk to existing behavior: this PR adds new, opt-in tooling and touches no existing runtime code path (server, UI, DB, adapters). The only existing file modified besides docs/package metadata is `.github/workflows/pr.yml` (two additive `node --test` steps).
- The core safety risk this PR is designed against is exactly the one #3207/#10555 already demonstrated (destructive "cleanup" of unpushed/uncommitted work) — mitigated by the fail-closed design: any error, ambiguity, or unresolvable signal preserves the candidate rather than acting on it, and the default tier never touches git-tracked state at all.
- `lsof`-only active-session detection has a real gap (an agent between tool calls with no open file handle) — mitigated by the new recent-commit guard, itself found and validated against a live host rather than only in the abstract.
- The LaunchAgent installer intentionally does not wire any pre-existing host-local retention scripts by default, to avoid double-scheduling a script that may already have its own dedicated LaunchAgent on a given host; this is documented explicitly in `docs/deploy/disk-retention.md`.

## Model Used

Claude (Anthropic), via GitHub Copilot CLI — extended reasoning/agentic tool-use mode, with live shell/git/gh tool execution against both this repository and a real self-hosted Paperclip instance for verification.

## Checklist

- [x] I have either (a) linked existing issues with `Fixes:` / `Closes:` / `Refs:` OR (b) described the issue in-PR following the relevant issue template


dzianisv:
Thanks for the review — all 3 findings were real gaps, fixed in f86c0e7:

1. **Tracked files under `node_modules`** — added `hasTrackedFilesUnderNodeModules()`; the node_modules-only tier now checks `git ls-files` for anything force-added under `node_modules/` and preserves it if found (a plain reinstall would silently discard a committed patch).
2. **Other local branches/stashes/tags** — added `hasUnpushedOtherLocalBranch()`, `getStashCount()`, `getLocalOnlyTags()`. Full worktree removal now blocks if any of these hold real, remote-unreachable history. Reachability is computed locally via `git merge-base --is-ancestor` against every `refs/remotes/**` ref (no network call). Note: this is deliberately NOT "any other local branch exists" — a plain `git clone` always leaves the default branch checked out locally before a feature branch, with zero unique commits, and that must stay reclaimable. I caught this exact false positive with a failing test while implementing the fix and added a dedicated regression test (`hasUnpushedOtherLocalBranch: a clone's leftover default branch is NOT unsafe`).
3. **TOCTOU at delete time** — added `revalidateBeforeDelete()`, which re-checks every safety signal immediately before each `rmSync` call in the apply loop and downgrades to `preserve` if anything changed since evaluation. This shrinks the race window from the whole run's duration to milliseconds; it doesn't eliminate it entirely (there's still a small gap before the actual syscall), which felt like the right tradeoff versus a full lock/mutex for a maintenance script that already fails closed by default.

Also fixed a pre-existing bug I found while validating these changes: dry-run always reported "Would reclaim 0B" even when individual candidates showed real reclaimable sizes, because the byte total was only accumulated inside the `--apply` branch. Now derived from the final results array so both modes report accurately.

Test suite grew from 32 → 47 (15 new tests: the 3 new pure/impure functions, the leftover-branch regression, and the dry-run total fix). Full suite re-run against the live target host with real fixture repos, including the exact scenario Greptile's issue #2 describes.

dzianisv:
## Round 3 of Greptile findings — verified individually, one real gap fixed

Greptile's latest pass raised 5 findings. Rather than reacting to the label
text alone, I re-read the current code and/or wrote a reproducing test for
each before deciding whether it needed a fix:

**Already fixed, stale restatements (no code change needed):**
- *"Stale checks authorize destructive deletion"* / TOCTOU claim — already
  covered: `revalidateBeforeDelete` re-derives `hasUnsafeExtraGitState`
  (stash/branch/tag/reflog) immediately before every `rmSync`, and this is
  exercised by 3 existing `--apply` CLI regression tests
  (`git commit -F` / branch / stash added between evaluation and deletion).
- *"Branch checks miss repository history"* — same code path as above;
  already covered by the reflog-guard fix from the previous round
  (`hasReflogOnlyCommits`).

**New and real — fixed this round:**
- *"Stale refs authorize repository deletion"* — genuinely new. No `git
  fetch` ever ran anywhere in the script, so `isBranchReachableFromAnyRemote`
  / `getLocalOnlyTags` trusted whatever `refs/remotes/**` happened to be
  cached locally. A worktree eligible for full removal is, by definition,
  inactive — exactly the condition under which nobody has fetched recently —
  so a remote branch force-pushed or deleted since the last fetch could
  still look "safely mirrored" and get destroyed for real. Fixed by adding
  `refreshRemoteTrackingRefs()` (fetch + prune every configured remote,
  fail-closed on any failure) called before the other-branch/tag checks in
  both `evaluateWorktree` and `revalidateBeforeDelete`. Proved the bug was
  real (not just theoretical) with a regression test that force-deletes a
  branch on the "remote" without ever fetching, confirms the pre-refresh
  read is still wrong, then confirms the post-refresh read is correct — plus
  a second, full CLI end-to-end test through `--apply
  --remove-merged-worktrees` confirming the worktree correctly falls back to
  the node_modules-only tier instead of being fully removed.

**Real but out of scope — documented, not "fixed":**
- *"Untracked dependency state/content is unverified/destructible"* (raised
  twice, at two locations) — this is a genuine, structural limitation: git
  cannot distinguish ordinary npm-install output from npm-install output
  plus one hand-added/patched *untracked* file, since neither is tracked.
  A correct fix would mean actually re-running the package manager install
  in a scratch dir and diffing file contents — a disproportionate,
  network-dependent, complexity-heavy addition for a disk-reclaim script,
  and out of step with this issue's own framing of the target as
  "reinstallable node_modules" (a lockfile-based definition the existing
  checks already fully satisfy). Documented honestly as a "Known
  limitation" in `docs/deploy/disk-retention.md` rather than silently
  dropped or over-engineered around.

Also fixed a subtlety this surfaced: `getRemoteOriginUrl()` used to call
`git remote get-url origin`, which applies any configured `url.<x>.insteadOf`
rewrite before returning — this would have corrupted the owner/repo string
fed to GitHub PR-state lookups on any repo using such a rewrite (the test
fixture now legitimately uses one, to let a real `git fetch` in tests
transparently redirect to a local bare "remote" without faking a
github.com-shaped URL). Switched to reading the raw `remote.origin.url`
config key directly, which bypasses the rewrite.

Tests: 44 → 46 in `reap-stale-workspaces.test.mjs` (53 total combined with
the install-agent suite), all passing locally.

Re-verified against the live host after this change: `/api/health` 200,
dry-run reclaim still 0 B (nothing newly eligible — expected, idempotent),
active worktrees untouched. (Observed an unrelated transient dip in host
free space during this session from other fleet/shared-host activity — not
caused by this script; it recovered on its own with zero action from this
PR, and `reclaimedBytes` was 0 on dry-runs either side of it.)

dzianisv:
Independent review found two blockers in PR #11936.

1. evaluateWorktree computes aheadOfRemote before refreshRemoteTrackingRefs, so dry-run can report a full worktree as safe even though apply revalidation preserves it after the fetch. Move refresh before the checked-out branch isAheadOfRemote calculation in both evaluate and revalidate paths, recompute the decision, and add a dry-run regression for a deleted remote branch.

2. The hourly LaunchAgent appends forever to fixed stdout.log/stderr.log with no rotation or cap, creating a new unbounded disk source. Add bounded rotation/truncation and tests. Also correct the doc frontmatter summary so it does not imply run-log/backup retention is implemented here. Re-run the 53 tests and request review; do not merge yet.

dzianisv:
Confirmed both findings against current HEAD (97a6ba1f) — requesting changes, do not merge yet.

**1. `aheadOfRemote` staleness (real, unresolved by 97a6ba1f)**

`evaluateWorktree()` computes `aheadOfRemote = isAheadOfRemote(dir, branch)` (line ~722) using whatever `refs/remotes/**` is *currently cached on disk* — no fetch has happened yet at that point. `wouldOtherwiseReapWorktree` (the gate that decides whether to spend the network `fetch --prune`) is itself derived from that same stale `aheadOfRemote`. So:

- If the stale cache says "not ahead" (false), `wouldOtherwiseReapWorktree` can become true, a fetch runs for the *other-branch/tag* checks (97a6ba1f's fix), but `decideAction()` still consumes the pre-fetch `aheadOfRemote` value — it is never recomputed post-fetch.
- Net effect: dry-run (`--json`/plain output) can report `reap-worktree` as the planned, "safe" action based on a stale ref, even when a fresh fetch would show the branch is actually ahead of origin (e.g. remote branch was force-pushed/rebased away, or genuinely has unpushed local commits the cache hasn't seen).
- `revalidateBeforeDelete()` *does* correctly re-check `isAheadOfRemote` in `--apply` mode, but only because it incidentally benefits from the on-disk ref update performed earlier in the same `evaluateWorktree()` call for the *other* checks — that's an implementation coincidence, not a guarantee, and it means dry-run output and apply behavior can diverge (dry-run says "reap", apply says "preserve"), which is exactly the operator-trust problem this tool exists to prevent.
- Fix: compute `aheadOfRemote` (or at least gate `wouldOtherwiseReapWorktree`) on a value that's fresh *after* any fetch this call decides to do, or unconditionally refresh remote refs once per worktree before the first `isAheadOfRemote` call when `allowWorktreeRemoval` is set at all (not only when the stale check already looks favorable). Add a regression test that forces a stale local `refs/remotes/origin/<branch>` (behind the real remote state) and asserts dry-run and `--apply` agree.

**2. Unbounded LaunchAgent stdout/stderr (real)**

`scripts/install-paperclip-retention-agent.sh` sets `StandardOutPath`/`StandardErrorPath` directly to `$LOG_DIR/stdout.log` / `stderr.log` with no `newsyslog`/rotation config and no size cap. Running hourly forever, these grow unbounded — the same failure shape this whole PR exists to prevent, just moved to `logs/`. Needs either: rotation via a companion `newsyslog.d` entry, a `logrotate`-style truncate-on-size check in the installer, or the retention script itself capping/truncating its own log files at start (e.g. keep last N runs / N MB) before writing.

Please fix both, add the regression test for (1), and re-request review.

dzianisv:
## Update: second independent review pass — both findings fixed (c9aff5c54)

Thanks for the second review pass. Both findings were real gaps, not already covered by the prior round's fix. Fixed in `c9aff5c54`:

### 1. Checked-out branch's own `aheadOfRemote` was still computed from a stale cache

The prior round (`97a6ba1f0`) added `refreshRemoteTrackingRefs()` for the **other**-branch/tag checks, but both `evaluateWorktree` and `revalidateBeforeDelete` still computed the **checked-out branch's own** `aheadOfRemote` via `isAheadOfRemote()` — which reads `<branch>@{upstream}`, itself dependent on cached `refs/remotes/**` — **before** that refresh ran. Worse, the refresh was gated behind a flag that itself depended on the already-stale `aheadOfRemote` value: a circular gate that could never trigger a refresh in the exact case that mattered.

Concretely: if the checked-out branch's remote copy was deleted or force-pushed since the last `fetch`, a stale local cache could still say "not ahead," clearing the gate for a full `reap-worktree` deletion.

**Fix:**
- `evaluateWorktree`: split the gate into `wouldPotentiallyReapWorktree` (every condition except `aheadOfRemote`), which decides whether the refresh is warranted; `aheadOfRemote` is then computed **after**, fail-closed to `true` if a warranted refresh itself failed.
- `revalidateBeforeDelete`'s `reap-worktree` branch: refresh now runs **strictly first**, and returns `safe: false` immediately if it fails — no fallback, since this is the last checkpoint before an irreversible delete.
- New dry-run regression test: deletes the checked-out branch's own ref on the bare remote fixture, then asserts the CLI's *plan itself* (not just the eventual delete) downgrades from `reap-worktree` to `reap-node-modules`. Verified it fails without the fix, passes with it.
- Full suite: **45/45 passing** (was 44).

### 2. LaunchAgent stdout/stderr logs were unbounded

`StandardOutPath`/`StandardErrorPath` just appended forever — an hourly, indefinitely-running disk-retention agent with its own unbounded log would eventually recreate the exact problem it exists to fix.

**Fix:** added `PAPERCLIP_RETENTION_MAX_LOG_BYTES` (default 5 MiB, documented in the installer's header) and a `rotate_log_if_large()` routine baked into the generated runner, called for both `stdout.log`/`stderr.log` before every reaper invocation. Uses **copy-then-truncate-in-place**, not rename — launchd already holds this process's own stdout/stderr fd open in append mode *before* the runner starts, so a rename would silently redirect the rest of *this run's own* output into the renamed file (fd follows inode, not path) instead of the fresh one. Verified against a real launchd-style append-mode-fd-open-before-exec invocation (not just unit tests), plus 3 new automated tests (default bytes correctly baked in; an oversized log rotates with old content preserved in `.1` and new output landing in the live file; a log under the cap is left untouched). Installer suite: **12/12 passing** (was 9).

### Also: corrected docs frontmatter scope

`docs/deploy/disk-retention.md`'s frontmatter `summary` claimed the shipped tooling reclaims "run logs, backups, and stale agent workspaces" — it only reclaims `instances/*/workspaces/`; run-log/backup reclaim is optional wiring to scripts the *host* may already have, not something this PR implements. Narrowed the summary and added an explicit "Scope of this page" note so the intro table (kept for context) can't be misread the same way again.

CI is running on `c9aff5c54`; will confirm once green. Per instruction, not merging — leaving this for maintainer/reviewer action.

dzianisv:
Independent re-review PASS on c9aff5c54: reproduced stale-remote dry-run/apply parity, append-fd-safe log rotation, and 57/57 tests. All prior findings are resolved; no remaining code blockers found.

PR #11957, opened 2026-08-22
fix(codex-local): send MCP bearer token in http_headers

Oldrich333:
## Summary
Codex 0.147 reads `http_headers` on managed MCP servers. The adapter currently writes the bearer token under `headers`, so the gateway connects without Authorization and returns HTTP 401.

This change writes `http_headers` and updates the focused unit assertion.

## Test plan
- [x] `packages/adapters/codex-local` focused test expects `http_headers`
- [ ] Confirm a Codex local run with a managed HTTP MCP server sends `Authorization: Bearer …`


zannis:
can we get this in shape and merged please? codex mcp calls are not working until this gets released

PR #11973, opened 2026-08-22
fix(workspaces): let the terminal reaper archive shared workspace sessions

ericdfields:
## Thinking Path

> - When an issue reaches a terminal state, the execution workspace it ran in is reclaimed by a reaper that sweeps every 30 seconds
> - That reaper has a keyset cursor, a lifecycle lock, a generation fence and a cooldown, and it has archived zero rows for the entire lifetime of the table
> - Every candidate is rejected by a delivery gate written for one-worktree-per-issue, applied to rows that are session records over a shared checkout
> - This pull request skips the gates that cannot apply to a session record, and scopes the resulting cleanup to what such a row actually owns
> - The benefit is that workspace rows stop accumulating without any exit condition, which is what the reaper was built to prevent

No public issue exists for this. The problem follows the bug report template.

Related in-flight work, not duplicated here: #11832 makes the archive path's issue-link clearing unconditional. It touches the adjacent `cleanup.cleaned && mode === "shared_workspace"` line and moves in the same direction as this change; expect a small textual conflict there and no semantic one.

**What happened?**

`sweepTerminalWorkspaces` archives a candidate only when its delivery state is `merged_via_pr` or `merged_by_ancestry`:

```ts
if (
  assessment.deliveryState !== "merged_via_pr"
  && assessment.deliveryState !== "merged_by_ancestry"
) {
  result.skippedUndelivered += 1;
  continue;
}
```

A `shared_workspace` row cannot reach either value:

- `merged_by_ancestry` requires `isMergedIntoBase === true`, which `inspectGitCloseReadiness` computes only `if (repoRoot && baseRef)`. A shared session has no `base_ref`, so it stays `null`.
- `merged_via_pr` requires a merged pull request with `details.headRef === workspace.branchName`. A shared session has no `branch_name`, so no reference matches.

With both false and `pullRequestStateUnknown` false, `deriveExecutionWorkspaceDeliveryState` returns `"unknown"`, which fails the check above. The dirty check rejects most of them even earlier, reading a working tree whose contents belong to whatever else is using that checkout.

Both gates ask the same question: would archiving destroy unmerged work in a worktree this row owns? For a session record over a shared checkout that is not merely false, it is unanswerable — and answering "no, therefore skip" means the row is rejected on every sweep, forever.

The interactive close path already encodes the correct semantics for these rows, in `getCloseReadiness`:

```ts
warnings.push("This shared workspace session points at project workspace infrastructure. Archiving it only removes the session record.");
```

and there an open linked issue is a warning rather than a blocking reason. The reaper never received the same treatment.

**Expected behavior**

A shared workspace session whose issue tree is terminal is archived, and archiving it removes the session record without touching the checkout it points at.

**Steps to reproduce**

1. Create a `shared_workspace` execution workspace with `branch_name` and `base_ref` null, pointing at a project checkout.
2. Move its source issue to `done`.
3. Run `sweepTerminalWorkspaces`.
4. The row is counted in `skippedUndelivered` and stays `active`. It will be counted there on every subsequent sweep.

## What Changed

- The delivery gates (`workspaceDirty`, `deliveryState`) are skipped for a shared session. Every other guard still applies: terminal subtree, cooldown, reopen fence, active-run check, and the full re-check inside the archive transaction.
- The git inspection is skipped for a shared session rather than run and discarded. `git status --porcelain -uall` over a large shared checkout is the expensive part of a sweep tick, and `assessDelivery` would additionally resolve pull request state over the network for a comparison that cannot match.
- Runtime-service teardown is scoped to the workspace id for a shared session. `stopRuntimeServicesForExecutionWorkspace` matches live records by cwd prefix as well as by id, and a shared session's cwd is the checkout every sibling session also points at.
- A row that owns no directory counts as cleaned. `cleaned` is computed as "the path is gone"; for a shared session the path is the project checkout and is supposed to remain.
- `cleanupExecutionWorkspaceArtifacts` refuses to remove a shared session's directory regardless of the `createdByRuntime` flag.
- Shared-session archives are recorded under their own `cleanup_reason` and an `archivedSharedSession` counter.

## Verification

Typecheck clean across all workspace packages.

`execution-workspaces-service` 71/71, including 5 new tests. Adjacent suites re-run and green: `execution-workspace-reopen`, `execution-workspaces-routes`, `workspace-reaper-cooldown-config`, `execution-workspace-policy`.

`workspace-instance-cleanup` has one failing test, `stops embedded Postgres before deleting the instance root`. It fails identically on the pristine base commit with this branch stashed, so it is pre-existing and unrelated.

Every new test was mutation-tested — each safeguard was removed in turn and the intended test confirmed to fail:

| Mutation | Test that caught it |
|---|---|
| re-apply the delivery gates to shared sessions | all three archive tests (nothing archives) |
| drop the ownership term from `cleaned` | archive test + on-disk test (rows land in `cleanup_failed`, `archived` stays 0) |
| pass the shared cwd to service teardown again | teardown-scoping test, alone |
| let shared sessions bypass the terminal-issue gate | non-terminal skip test, alone |
| remove the shared-session deletion guard | deletion-guard test — the sweep deletes the checkout |

The third and fifth are the ones worth noting. Without the cwd scoping, archiving a terminal session stops a live neighbour's runtime service; without the deletion guard, the sweep removes a real project checkout.

## Risks

The blast radius is 2,616 of 2,626 rows on the deployment this was diagnosed against, since `shared_workspace` is effectively the only mode in use there. Those rows will archive in batches of 50 per sweep once this ships.

What that archiving does is bounded to the row: status, `closed_at`, `cleanup_eligible_at`, `cleanup_reason`, and the issue-link detach that already existed. Nothing on disk is removed for these rows — that is asserted by two of the new tests, one of which drives the only code path that could have deleted anything.

The reverse risk was the one that needed guarding, and both halves are now covered by tests rather than by luck: the cwd-prefix teardown reaching a live sibling, and the `createdByRuntime` flag being the only thing standing between a shared row and `fs.rm` on a developer's checkout.

Isolated workspaces are unaffected: every changed branch is behind `mode === "shared_workspace"`, and the existing delivery-gated tests still pass unchanged.

## Category safeguard

The class here is a policy written for one resource model and applied to rows of another, where the mismatch shows up as a permanently-skipped candidate rather than an error.

Implemented in this PR: the `archivedSharedSession` counter splits the archive total by route, so the two paths cannot silently collapse into one number again. A reaper archiving nothing already had a log line — that was the subject of #11971 — but a reaper archiving only one of two classes did not, and could not, because a single counter cannot express it.

Deferred as a follow-on rather than done here: a startup assertion that every distinct `execution_workspaces.mode` is reachable by at least one archive route, so a mode added later that no route can archive fails loudly instead of accumulating rows for months. That needs a mode registry that does not exist yet, which is more than this fix should carry.

## Model Used

Claude Opus 5 (`claude-opus-5`), 1M context, extended thinking, with tool use and code execution.

## Checklist

### DIFF
--- a/server/src/__tests__/execution-workspaces-service.test.ts
+++ b/server/src/__tests__/execution-workspaces-service.test.ts
@@ -43,6 +43,7 @@ import {
   startRuntimeServicesForWorkspaceControl,
   stopRuntimeServicesForExecutionWorkspace,
 } from "../services/workspace-runtime.ts";
+import * as workspaceRuntime from "../services/workspace-runtime.ts";
 import { workspaceGitOperationScheduler } from "../services/workspace-git-operation-scheduler.ts";
 
 const execFileAsync = promisify(execFile);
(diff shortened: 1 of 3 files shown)


MrBlackTongue:
Подготовил дополнение к этому исправлению: [коммит b48d90b7e](https://github.com/MrBlackTongue/paperclip/commit/b48d90b7e7bd952c168de52890228d9e1f2efe8d). Ветка основана на вашем изменении b4aaa504a; отдельный дублирующий PR не открываю.

Дополнение закрывает повторное создание записей общего окружения и учитывает всех его держателей при архивировании: связанные открытые задачи и потомки, прогоны, службы, а также удаление исходной задачи. Переиспользование проверяет фактический каталог, Git-окружение, ветку и настройки; привязка и архивирование используют общий замок. Общая папка сохраняется. Возврат задачи в работу снимает устаревшую привязку к архивированной записи.

Проверка типов сервера прошла. Все 14 новых целевых тестов прошли, включая сохранность файла и уменьшение открытого списка с 1 до 0. Расширенная проверка пяти файлов: 519 прошли, 1 упал — `retries a bind race on the next bounded port and persists the selected URL`. Этот же тест падает на базе с вашим исправлением без моих изменений; его пробную правку исключил. Полную сборку и весь набор тестов репозитория не заявляю проверенными.

Предлагаю автору PR рассмотреть включение дополнения либо указать, какие изменения нужно отделить. В установленную систему исправление ещё не применялось.

PR #11982, opened 2026-08-22
fix(issues): notify the unblock owner on every path that leaves an issue blocked

ericdfields:
## Thinking Path

> - The issues subsystem lets a blocked issue record an `unblockDescriptor`: the agent who owns the way out, and the action they must do
> - When that descriptor names an agent, the server wakes that agent so the block becomes routable work instead of a dead row
> - That wakeup was wired to one path only: the moment an issue moves into `blocked`. Two other paths reach the same state and stayed silent
> - This pull request calls the same notify step from every path that can leave an issue blocked with an agent-owned descriptor
> - The benefit is that a blocked issue with a named owner always tells that owner, so blocked work stops being invisible to the person who can clear it

No public issue exists for this. The problem follows the bug report template.

**What happened?**

A blocked issue can name an agent as its unblock owner. The server is supposed to wake that agent. It only does so when the issue *enters* `blocked` through `PATCH /api/issues/:id`. Two other paths reach the same state and send nothing:

1. **Descriptor attached after the transition.** An issue moves to `blocked` with blockers and no descriptor. A later PATCH adds the descriptor. `enteringBlocked` is already false, so the notify branch is skipped. The owner is never told.
2. **Issue created directly in `blocked`.** The create and child-create handlers never called the notify step at all. An issue born blocked with a descriptor strands immediately.

There is a third, quieter failure. The wakeup idempotency key was `issue-unblock:<issueId>:<blockedTransitionAt>`. Re-pointing a descriptor at a different owner does not move `blockedTransitionAt`, so the new owner's wakeup reuses the previous owner's key and is discarded as a duplicate.

Observed on a live instance: of the blocked issues whose descriptor names an agent, three had a null `blocked_owner_notified_at`. One had been waiting a full day. Two of the three were created directly as blocked children; the third had its descriptor attached about two minutes after the transition.

**Expected behavior**

An issue that is blocked with an agent-owned `unblockDescriptor` wakes that agent, whichever path put it in that state. Re-pointing the descriptor at a different agent wakes the new owner. An unchanged descriptor does not wake anyone again.

**Steps to reproduce**

1. Create an issue and move it to `blocked` with a blocker and no `unblockDescriptor`.
2. Send a second PATCH that sets `unblockDescriptor` to `{ "owner": { "agentId": "<id>" }, "action": "..." }`.
3. Read the issue row. `unblock_descriptor` is set and `blocked_owner_notified_at` is null. No wakeup was queued for that agent.
4. Alternatively, create a child issue with `status: "blocked"` and an `unblockDescriptor` in the same request. The same two columns show the same result.

`master`, at commit cc42a67e7.

**Deployment mode**

Self-hosted, single instance.

**Agent adapter(s) involved**

## What Changed

- Added `notifyBlockedOwner` in `routes/issues.ts`. It runs the delivery and writes `blockedOwnerNotifiedAt`. All call sites now use it.
- Called it from the three paths that previously skipped it: issue create, child create (both handlers), and the bulk child-create loop.
- Widened the PATCH condition from "entering blocked" to also include a descriptor that is attached or re-pointed while the issue is already blocked.
- A re-point ignores the earlier delivery stamp, because it is a new obligation for a different owner.
- Added `unblockDescriptorFingerprint` and put it in the wakeup idempotency key, so a re-point no longer collides with the previous owner's key. Genuine retries still deduplicate.
- Added a test that fails if any server source file contains a NUL byte.

## Verification

```
cd server
npx vitest run \
  src/__tests__/routable-blocked.test.ts \
  src/__tests__/blocked-owner-notification-routes.test.ts \
  src/__tests__/source-files-are-text.test.ts \
  src/__tests__/issue-dependency-wakeups-routes.test.ts \
  src/__tests__/attention-service.test.ts
npm run typecheck
```

All pass. The wider issue-route suites also pass (231 tests across
`issue-create-deduplication`, `issue-blocker-diagnostics`,
`issue-agent-mutation-ownership`, `issue-execution-policy`,
`issue-scheduled-retry`, and `issue-comment-reopen`).

Each new test was checked against a deliberately broken build to confirm it
fails when the fix is removed:

- Narrowing the PATCH condition back to `enteringBlocked` fails both new route tests.
- Treating every descriptor as changed fails the "unchanged descriptor" test.
- Removing the fingerprint from the idempotency key fails both unit tests.
- Adding a NUL byte to a source file fails the new text test.

## Risks

- No migration and no schema change. The affected columns already exist.
- Agents can now receive a wakeup from paths that never sent one. This is the intent of the change. The volume is bounded by how many issues are blocked with an agent-owned descriptor, which is a small number in practice.
- The idempotency key format changed. A key is only compared against other keys for the same issue and the same transition timestamp, so an in-flight wakeup queued under the old format is not re-sent. The worst case is one duplicate wakeup for an issue that is blocked across the deploy.
- A descriptor owned by the board or by a user still sends nothing, which is unchanged.

## Category safeguard

A stray NUL byte reached a template-string separator in `routable-blocked.ts`
during this change. The code ran correctly and every test passed, but `git`
reclassified the file as binary. `git diff` printed "Binary files differ", so
the change was invisible to human review and to any automated reviewer that
reads the diff.

A defect that hides the diff it lives in is worth failing the build over.
`source-files-are-text.test.ts` scans every `.ts` and `.tsx` file under
`server/src` and fails with the file name and byte offset if a NUL is found.

## Related PRs

Searched the open PR list for work on blocked issues, unblock descriptors and
wakeups. Two open PRs touch the same area. Neither does what this PR does, but
both edit nearby lines:

- #11955 — changes **who may be named** as an unblock owner. It refactors the
  same descriptor block in `routes/issues.ts` that this PR reads from, and
  replaces `updateFields.unblockDescriptor` with a
  `resolveRequestedUnblockDescriptor(...)` helper. That is an authorization
  change; this PR is a delivery change. They are complementary, and they will
  conflict textually. Whichever merges second should rebase and take the
  helper. This PR then benefits, because the helper also finds a descriptor
  sent only in the request body.
- #11952 — preserves requested blocked-descriptor semantics. Also edits
  `routes/issues.ts` in the same handler.

No open PR adds the missing notify call sites or changes the wakeup
idempotency key, so this work is not a duplicate.

## Model Used

Claude Opus 5 (`claude-opus-5`), 1M context window, extended thinking, with
tool use and code execution (repository search, file editing, local test runs,
and read-only SQL against a live instance to characterise the defect).

## Checklist

- [ ] I have updated relevant documentation to reflect my changes — not applicable; this restores intended behaviour and adds no new API surface
- [ ] All Paperclip CI gates are green — running at the time of writing
- [ ] Greptile is 5/5 with no open P2s, recommendations, or follow-ups — review pending

### DIFF
--- a/server/src/__tests__/blocked-owner-notification-routes.test.ts
+++ b/server/src/__tests__/blocked-owner-notification-routes.test.ts
@@ -1,0 +1,248 @@
+import express from "express";
+import request from "supertest";
+import { beforeEach, describe, expect, it, vi } from "vitest";
+
+// See issue-dependency-wakeups-routes.test.ts: the first test pays the
+// one-time cost of importing the large routes/issues.ts module, which can
+// cross vitest's default 5s timeout on a loaded serial shard.
+vi.setConfig({ testTimeout: 15000 });
+
+const OWNER_AGENT_ID = "00000000-0000-4000-8000-0000000000a1";
+const ISSUE_ID = "aaaaaaaa-aaaa-4aaa-8aaa-aaaaaaaaaaaa";
+// Must sit after ROUTABLE_BLOCKED_ROLLOUT_AT or the delivery is declined by design.
+const BLOCKED_AT = new Date("2026-08-21T19:19:58.733Z");
+
+const mockWakeup = vi.hoisted(() => vi.fn(async () => undefined));
+const mockIssueService = vi.hoisted(() => ({
+  getAncestors: vi.fn(async () => []),
+  getById: vi.fn(),
+  getByIdForUpdate: vi.fn(),
+  getByIdentifier: vi.fn(async () => null),
+  getComment: vi.fn(async () => null),
+  getCommentCursor: vi.fn(),
+  getRelationSummaries: vi.fn(),
+  update: vi.fn(),
+  getDependencyReadiness: vi.fn(),
+  listWakeableBlockedDependents: vi.fn(async () => []),
+  getWakeableParentAfterChildCompletion: vi.fn(async () => null),
+  findMentionedAgents: vi.fn(async () => []),
+}));
+
+vi.mock("../services/index.js", () => ({
+  companyService: () => ({
+    getById: vi.fn(async () => ({ id: "company-1", attachmentMaxBytes: 10 * 1024 * 1024 })),
+  }),
+  accessService: () => ({ canUser: vi.fn(), hasPermission: vi.fn() }),
+  agentService: () => ({ getById: vi.fn() }),
+  companySkillService: () => ({ completeTestRunForIssue: vi.fn(async () => null) }),
+  documentAnnotationService: () => ({ remapOpenThreadsForDocument: async () => [] }),
+  documentService: () => ({ getIssueDocumentPayload: vi.fn(async () => ({})) }),
+  executionWorkspaceService: () => ({ getById: vi.fn() }),
+  feedbackService: () => ({}),
+  goalService: () => ({ getById: vi.fn(), getDefaultCompanyGoal: vi.fn() }),
+  heartbeatService: () => ({
+    wakeup: mockWakeup,
+    reportRunActivity: vi.fn(async () => undefined),
+  }),
+  getIssueContinuationSummaryDocument: vi.fn(async () => null),
+  instanceSettingsService: () => ({ get: vi.fn(), listCompanyIds: vi.fn() }),
+  issueApprovalService: () => ({}),
+  issueReferenceService: () => ({
+    deleteDocumentSource: async () => undefined,
+    diffIssueReferenceSummary: () => ({
+      addedReferencedIssues: [],
+      removedReferencedIssues: [],
+      currentReferencedIssues: [],
+    }),
+    emptySummary: () => ({ outbound: [], inbound: [] }),
+    listIssueReferenceSummary: async () => ({ outbound: [], inbound: [] }),
+    syncComment: async () => undefined,
+    syncDocument: async () => undefined,
+    syncIssue: async () => undefined,
+  }),
+  issueRecoveryActionService: () => ({
+    getActiveForIssue: vi.fn(async () => null),
+    listActiveForIssues: vi.fn(async () => new Map()),
+  }),
+  issueThreadInteractionService: () => ({
+    listForIssue: vi.fn(async () => []),
+    expirePendingInteractionsForTerminalIssue: vi.fn(async () => []),
+    expireRequestConfirmationsSupersededByComment: vi.fn(async () => []),
+    expireStaleRequestConfirmationsForIssueDocument: vi.fn(async () => []),
+  }),
+  issueService: () => mockIssueService,
+  logActivity: vi.fn(async () => undefined),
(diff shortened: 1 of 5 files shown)


ericdfields:
### Re: "Historical descriptor key reuse" — fixed in 8009922, with one correction to the diagnosis

**The collision is real; the stated consequence is not — yet.** Worth separating, because the difference decides how it should be fixed.

The collision reproduces exactly as described. `blocked_transition_at` does not move when a descriptor is re-pointed on an issue that is already blocked (`server/src/services/issues.ts:7676-7683` only stamps it on entry to `blocked`), so re-pointing A → B → A rebuilds issue id + transition + fingerprint identically. The third delivery reuses the first's key.

What does *not* happen is the wakeup being discarded. Nothing dedupes the `issue-unblock:` prefix:

- `enqueueWakeup` (`server/src/services/heartbeat.ts:17768`) never reads `idempotency_key` back — all 13 references in its body are writes into the insert values. The queued-path insert (`:18784-18799`) is a plain `.insert(...).returning()` with no `ON CONFLICT`. `heartbeat.wakeup` is `trackWakeup` (`:14104`), a promise-tracking pass-through.
- There is no general unique index on `agent_wakeup_requests.idempotency_key`. The column is plain nullable `text` from its creation migration (`packages/db/src/migrations/0001_fast_northstar.sql:18-30`), and the only unique indexes are partial, scoped by `LIKE` to the `issue_review_path_lost:%` and `issue_disposition_repair:%` prefixes.
- The one caller-side pre-check (`heartbeat.ts:9722-9731`) belongs to the review-path-lost flow, not this one.

Suppression of a repeat delivery here comes from the persisted `blocked_owner_notified_at` stamp, which is why the re-point path clears it explicitly.

So: latent, not live. The restored owner *is* woken today.

**That still makes the PR wrong, in a way worth fixing** — the comments asserted that the fingerprint in the key "keeps genuine retries deduplicated," which describes protection that isn't wired. A reader trusting that comment and adding a partial unique index for this prefix would convert a dormant collision into exactly the dropped notification you described. Given this PR sits under a lifecycle-instrumentation issue where adding such an index is a plausible next step, leaving it dormant is not a good bet.

**Fix:** thread the superseded delivery stamp through as a generation counter, so each delivery gets a distinct key regardless of how the descriptor moves:

```
issue-unblock:{issueId}:{blockedTransitionAt}:{fingerprint}:{first | after:<prior notified_at>}
```

First delivery is `first`; each re-point carries the stamp it supersedes. A → B → A now yields three distinct keys. Genuine retries are unaffected — they are still stopped upstream by `blocked_owner_notified_at`, before a key is ever built.

Comments in `routable-blocked.ts` and `issues.ts` now state where suppression actually comes from rather than attributing it to the key.

**Test:** `separates a descriptor re-pointed back to the first owner from its first delivery` walks the full A → B → A sequence, threading the stamp the way the route does, and asserts three distinct keys plus the specific third-vs-first inequality. Mutation-checked — pinning the discriminator back to a constant fails that test and only that test (`expected 2 to be 3`). Server typecheck clean; both `routable-blocked` (6) and `blocked-owner-notification-routes` (3) suites green.

PR #11991, opened 2026-08-22
fix(ui): guard the full capabilities chain in FileViewerSheet

jotavgomes:
## Thinking Path

> - `FileViewerSheet` reads `resolvedResource.capabilities.preview` and `.download` to decide whether to show preview/download controls
> - The optional chaining only covered `resolvedResource`, not `capabilities` itself — a resolved resource with a present-but-`capabilities`-less shape crashes the viewer instead of degrading gracefully
> - This pull request chains the optional access through `capabilities` too
> - The benefit is the file viewer degrades to no-preview/no-download instead of throwing when a resolved resource's capabilities are absent

No existing public GitHub issue. Describing in-PR per the bug-report template (`.github/ISSUE_TEMPLATE/bug_report.yml`):

- **Subsystem affected:** `ui/src/components/FileViewerSheet.tsx`.
- **Problem or motivation:** `canPreview` and `downloadUrl` read `resolvedResource?.capabilities.preview` / `.download` — optional chaining stops at `resolvedResource`, so a resolved resource whose `capabilities` field is `undefined`/`null` throws a `TypeError` instead of the viewer degrading to "can't preview this file."
- **Proposed solution:** Chain the optional access through `capabilities` too: `resolvedResource?.capabilities?.preview` / `?.download`.
- **Roadmap alignment:** Small, isolated bug fix — no core/roadmap overlap.

## What Changed

- `ui/src/components/FileViewerSheet.tsx`: two-line fix, `?.capabilities.` → `?.capabilities?.`.
- `ui/src/components/FileViewerSheet.copy.test.tsx`: new test — a resolved resource missing `capabilities` renders without throwing, shows no download link, and shows the "can't preview" message.

## Verification

```bash
pnpm --filter @paperclipai/ui typecheck   # passes
npx vitest run ui/src/components/FileViewerSheet.copy.test.tsx   # 6/6 passing
```

Revert-and-retest: with the fix reverted, the new test fails with `TypeError: Cannot read properties of undefined (reading 'preview')` at the exact line the fix touches — confirms the test actually exercises the bug.

## Risks

None beyond the change itself. Purely additive null-safety: `capabilities` is currently typed as required on the resolved-resource shape, so this only changes behavior when the runtime value diverges from the type — exactly the crash this fixes. No behavior change for the (overwhelmingly common) case where `capabilities` is present.

## Model Used

Claude Sonnet 5 (`claude-sonnet-5`), used interactively.

## Checklist

- [x] I have updated relevant documentation to reflect my changes (no user-facing docs affected)

---

**Scope note:** this PR originally also included an atomic-swap fix for `materializeRuntimeSkillFiles` in `server/src/services/company-skills.ts`. That part is removed from this PR — Greptile correctly flagged that the underlying `createDirectoryReplacement.commit()` helper still has a sub-millisecond window between its two `rename()` syscalls where a concurrent reader could observe the directory as briefly missing, and fixing that properly means changing a shared helper's contract for all four of its call sites, which needs its own separately-scoped PR and audit (symlink compatibility, etc.) rather than being folded into this narrower UI fix. This PR is now exactly the FileViewerSheet half, with no open caveats.

### DIFF
--- a/ui/src/components/FileViewerSheet.copy.test.tsx
+++ b/ui/src/components/FileViewerSheet.copy.test.tsx
@@ -170,6 +170,22 @@ describe("FileViewerSheet copy actions", () => {
     expect(previewPane?.className).not.toContain("rounded");
   });
 
+  it("degrades to no-preview instead of throwing when capabilities is missing from the resolved resource", () => {
+    const resourceWithoutCapabilities = { ...resolvedResource } as Partial<ResolvedWorkspaceResource>;
+    delete resourceWithoutCapabilities.capabilities;
+
+    useQueryMock.mockImplementation((options: { queryKey?: readonly unknown[] }) => {
+      const key = JSON.stringify(options.queryKey ?? []);
+      if (key.includes('"content"')) return ok(content);
+      return ok(resourceWithoutCapabilities);
+    });
+
+    expect(() => renderSheet()).not.toThrow();
+
+    expect(document.body.querySelector('a[aria-label="Download file"]')).toBeNull();
+    expect(document.body.textContent).toContain("Can't preview this file");
+  });
+
   it("defaults Markdown files to rendered mode and switches back to raw source", async () => {
     useQueryMock.mockImplementation((options: { queryKey?: readonly unknown[] }) => {
       const key = JSON.stringify(options.queryKey ?? []);
--- a/ui/src/components/FileViewerSheet.tsx
+++ b/ui/src/components/FileViewerSheet.tsx
@@ -512,8 +512,8 @@ export function FileViewerSheet({
   });
 
   const resolvedResource: ResolvedWorkspaceResource | undefined = resolveQuery.data;
-  const canPreview = resolvedResource?.capabilities.preview ?? false;
-  const downloadUrl = state && resolvedResource?.capabilities.download
+  const canPreview = resolvedResource?.capabilities?.preview ?? false;
+  const downloadUrl = state && resolvedResource?.capabilities?.download
     ? fileResourcesApi.downloadUrl(issueId, state)
     : null;
 


jotavgomes:
Re: the directory-swap missing-state window Greptile flagged in `createDirectoryReplacement.commit()` (`company-skills.ts:5660`).

This is a real, valid observation, and I want to be explicit about the scope decision rather than silently leaving it unaddressed: **I'm not fixing it in this PR**, and here's why.

`createDirectoryReplacement` is a shared helper used by four call sites (`materializeCatalogManifestSkillFiles`, `materializeCatalogOriginSnapshot`, `copySkillDirectory`, and now `materializeRuntimeSkillFiles`), not something I introduced. Its two-step swap (`rename(target, previous)` then `rename(staging, target)`) is the standard technique for atomic-ish directory replacement on POSIX — a single-step atomic replace isn't available via plain `rename(2)` when the destination directory is non-empty (the syscall requires an empty destination), so some window between the two renames is inherent to this approach without moving to a symlink-indirection scheme (stable symlink swapped atomically to point at a versioned real directory).

Two things make me confident this is the right call to defer rather than fix inline:
1. **Severity is much smaller than what this PR fixes.** The bug this PR closes was a window spanning the *entire* materialization (network fetch + N file writes — measured in the hundreds of ms to low seconds for GitHub-sourced skills). The residual window Greptile found spans two `rename()` syscalls — sub-millisecond, metadata-only operations.
2. **A real fix means changing the shared helper's contract for all four callers**, which I haven't audited for symlink-compatibility (e.g. whether anything downstream does `fs.realpath`, expects a real directory for a bind-mount, etc.). That's meaningfully bigger blast radius than this PR's stated scope (fix the `__runtime__` rm-then-rebuild race + the FileViewer optional-chaining gap).

I'd rather ship the narrow fix with this called out than block on a broader rewrite I can't fully verify safe in this pass. Flagging as explicit follow-up work rather than closing the thread silently.

jotavgomes:
**Narrowed this PR's scope.** Force-pushed a rewritten branch history that drops the \`company-skills.ts\` atomicity fix entirely — that part is a real, still-open gap (the two-step \`rename()\` swap in \`createDirectoryReplacement.commit()\`), and after discussion it's staying out of this PR rather than shipping with a documented caveat. It'll come back as its own PR once someone actually audits and fixes the shared helper's contract across all four call sites.

This PR is now exactly the \`FileViewerSheet\` optional-chaining fix + its test — no open caveats, no scope decisions left to make. Title and description updated to match.

jotavgomes:
@greptileai review

jotavgomes:

PR #11996, opened 2026-08-23
fix(tool-gateway): let gateway bearer tokens reach the MCP protocol routes

ericdfields:
## Problem

**No agent on this instance has any `mcp__*` tool — for any connection.** Not a config problem: the Tool Connection is enabled and healthy, the profile bindings are right, and the adapter hands the Claude CLI a fully correct `--mcp-config`.

The break is one layer further in. The MCP gateway protocol endpoints authenticate their own bearer credential — a `pcgw_` token minted by the tool gateway — inside the route handler (`namedGatewaySessionFromBearer`). They never read `req.actor`. But `actorMiddleware` is mounted app-level (`server/src/app.ts`), ahead of every route, and treats any bearer it can't resolve as a failed board/agent credential:

```ts
if (!claims) {
  next(unauthorized(invalidAgentTokenMessage(token)));   // 401 — route never runs
  return;
}
```

There is no branch for gateway tokens, so **every** `pcgw_` credential 401s before reaching the endpoint it was minted for. Both mounts are affected: the public `/mcp/gateways/:publicId` and the `/api/tool-gateway/gateways/:id/mcp` URL that `buildPaperclipRuntimeMcpServers` hands to adapters.

The failure is silent in both directions. Nothing logs an error; an MCP client that can't `initialize` just registers zero tools. From the agent's side that is indistinguishable from "no MCP servers configured."

## Evidence

Reproduced live before the fix:

```
$ curl -s -X POST "$API/api/tool-gateway/gateways/$GATEWAY_ID/mcp" \
    -H "Authorization: Bearer pcgw_..." -d '{"jsonrpc":"2.0","id":1,"method":"initialize",...}'
HTTP/1.1 401 Unauthorized
{"error":"Agent token did not verify; obtain fresh credentials and retry"}
```

That string is `server/src/middleware/auth.ts:72` — the middleware, not the route.

The token in that probe was live, unexpired, and unrevoked. Across the instance:

```sql
select count(*) total, count(last_used_at) used
  from tool_mcp_gateway_tokens where subject_type='heartbeat_run';
 total | used
-------+------
    46 |    0
```

**46 runtime tokens minted, 0 ever used** — no MCP request has ever reached the gateway service.

After the fix, driving a real minted token through the real routes against the live database:

```
initialize -> 200 {"jsonrpc":"2.0",...,"serverInfo":{"name":"Paperclip MCP Gateway","version":"1.0.0"}}
tools/list -> 200
tool names: [ ...:create-link, ...:delete-link, ...:list-my-links, ...:promote-link, ...:revise-link ]
```

## Fix

A narrow carve-out in `actorMiddleware` for `pcgw_`-prefixed bearers on the two MCP gateway protocol paths. It forces `req.actor` to `"none"` rather than falling through, so a gateway token can never inherit the implicit board actor that `local_trusted` mode seeds at the top of the middleware. Every other route, and every other credential shape, is untouched.

## Category safeguard

The class here is *"Paperclip mints a credential for its own endpoint and something upstream of the route rejects it"* — invisible by construction, so it needs a structural guard rather than observation.

`server/src/services/mcp-gateway-endpoints.ts` now owns both the URL builder and the path matcher, and `heartbeat.ts` builds its runtime MCP URL through it instead of inline. `mcp-gateway-endpoint-coverage.test.ts` asserts that every endpoint path Paperclip builds is one the middleware admits, that neighbouring authenticated tool-gateway routes are **not** admitted, and that the URL isn't rebuilt inline again. A future change to the URL shape fails CI instead of silently un-registering every MCP server.

## Verification

- `mcp-gateway-token-auth.test.ts` — 7 new tests, including three negative ones: a gateway token on a non-gateway route still 401s, a non-gateway bearer on a gateway route still 401s, and `local_trusted` grants no ambient board actor.
- Mutation-tested: disabling the carve-out fails 4 of the 7; the three security-negative tests still pass, so they're guarding the boundary rather than the happy path.
- All 29 server auth/gateway test files green (419 tests).
- `tsc --noEmit` clean.
- End-to-end probe against the live instance database, shown above.

### DIFF
--- a/server/src/__tests__/mcp-gateway-endpoint-coverage.test.ts
+++ b/server/src/__tests__/mcp-gateway-endpoint-coverage.test.ts
@@ -1,0 +1,56 @@
+import { readFileSync } from "node:fs";
+import { fileURLToPath } from "node:url";
+import { describe, expect, it } from "vitest";
+import {
+  isMcpGatewayProtocolPath,
+  mcpGatewayApiEndpointPath,
+  mcpGatewayPublicEndpointPath,
+} from "../services/mcp-gateway-endpoints.js";
+
+const GATEWAY_ID = "819f6caa-92bb-41ee-a33c-3e35bd577a8f";
+const GATEWAY_PUBLIC_ID = "gw_61df2b661160478d9091986dbe51b2d8";
+
+function readSource(relativePath: string): string {
+  return readFileSync(fileURLToPath(new URL(relativePath, import.meta.url)), "utf8");
+}
+
+/**
+ * Category safeguard for BRO-2546.
+ *
+ * The failure class: Paperclip mints a credential for one of its own endpoints,
+ * and something upstream of the route rejects it. That failure is silent — an
+ * MCP client that cannot `initialize` just registers zero tools — so it has to
+ * be caught structurally rather than observed.
+ */
+describe("MCP gateway endpoint coverage", () => {
+  it("admits every endpoint path Paperclip builds for a gateway", () => {
+    expect(isMcpGatewayProtocolPath(mcpGatewayApiEndpointPath(GATEWAY_ID))).toBe(true);
+    expect(isMcpGatewayProtocolPath(mcpGatewayPublicEndpointPath(GATEWAY_PUBLIC_ID))).toBe(true);
+  });
+
+  it("does not admit neighbouring tool-gateway routes", () => {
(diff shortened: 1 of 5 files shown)


ericdfields:
Superseded by #11997. Same commit, but the branch name carried an internal ticket id and the description was missing required template sections. Closing this one to keep the history clean.

Issue #11203, opened 2026-08-11
codex_local ACP engine never reaches the Paperclip-managed MCP gateway (CLI engine does, same agent, A/B)


produtoramaxvision:
## Pre-submission checklist

- [x] Searched existing issues — the closest are #10690 (a feature PR that adds *locally-configured stdio* servers to the ACPX engine) and #10346 (`codex_local` gateway credentials issued but unused, root-caused to a `headers` vs `http_headers` TOML key and fixed for the **CLI** engine). Neither describes the defect below: with the CLI path fully working, the same agent on the ACP engine never opens a gateway session at all.
- [x] Reproduced live on a self-hosted install, A/B on a single agent with only `adapterConfig.engine` changed between runs.
- [x] Confirmed the gap is in Paperclip, not in the adapter binary or the gateway: the gateway is reachable and answers correctly for the CLI engine in the same minute, with the same profile, bindings and token minting.

## What happened?

A `codex_local` agent on `engine: "acp"` never reaches its Paperclip-managed MCP gateway. No `tool_gateway.discovery` audit row is written for the run, no gateway session is created, and the agent reports the plugin's tools as unavailable. The run itself finishes `succeeded`, so nothing surfaces as an error.

Switching the **same agent** to `engine: "cli"` and resetting its session makes the tools appear immediately.

## A/B evidence

Same agent, same company, same gateway, same profile bindings, session reset before each run, nothing else changed:

| `adapterConfig.engine` | run | run status | `tool_gateway.discovery` rows |
| --- | --- | --- | --- |
| `acp` | `2d704c93` | `succeeded` | **0** |
| `cli` | `a446a5ca` | `succeeded` | **2, `visibleToolCount = 7`** |

The 7 tools include the plugin's own `<plugin-id>:generate_image`, which the ACP run reported as missing.

## Where the two paths diverge

The CLI path merges two sources and writes them into the managed `CODEX_HOME`:

```ts
// packages/adapters/codex-local/src/server/execute.ts
const managedMcpGateways = mergeManagedCodexMcpGateways(
  runtimeMcpGateways,                      // from ctx.runtimeMcp
  managedMcpGatewaysFromContext(context),  // from context.paperclipManagedMcp
);
await writeManagedCodexMcpConfig({ codexHome: effectiveCodexHome, gateways: managedMcpGateways, ... });
```

The ACP path returns before ever reaching that code:

```ts
[...]

Issue #10102, opened 2026-07-23
Claude Code 2.1.218 silently strips OAuth MCPs from claude_local agent runs — agents degrade without erroring (upstream regression + hardening suggestions)


josherau:
## Summary

A Claude Code regression silently strips **all OAuth-backed MCP servers** from `claude_local` agent runs. Claude Code **2.1.218** (the first native-binary npm build) marks every OAuth HTTP MCP as "requires authentication" in non-interactive `--print` sessions — exactly how Paperclip's claude adapter spawns Claude — even when valid tokens exist in the macOS Keychain and the same servers work interactively. Upstream report: https://github.com/anthropics/claude-code/issues/80603

Reporting here because the failure mode is nasty for Paperclip specifically: **agents don't error — they degrade silently.**

## Impact on Paperclip

- Host machines typically auto-update Claude Code (in our case the updater ran `npm install -g @anthropic-ai/claude-code@2.1.218` minutes after a Paperclip upgrade, making it look like the Paperclip upgrade broke agents).
- Agents on the claude_local adapter (2026.722.0, ACP engine in adapter-utils) see OAuth MCPs listed as "require authentication before their tools can be used" and route around them — e.g. our newsletter agent skipped its Beehiiv MCP publishing step and quietly fell back to writing local drafts, reporting the run as successful.
- Nothing in run status, heartbeat results, or server logs flags that the agent's MCP toolset shrank. Diagnosis is hard because `claude mcp list` still shows ✔ Connected; only a real `--print` probe reveals the problem.

## Repro

On a host with an OAuth HTTP MCP configured at user scope (e.g. `https://mcp.beehiiv.com/mcp`) and authenticated interactively:

```bash
npm i -g @anthropic-ai/claude-code@2.1.218
claude --print --model haiku --dangerously-skip-permissions \
  "Do not use tools. Is MCP server beehiiv-official available or does it require authentication?"
# → "requires authentication"
npm i -g @anthropic-ai/claude-code@2.1.210
# same probe → connects; real tool calls succeed
```

## Workaround we're running

Pin Claude Code at 2.1.210 and set `DISABLE_AUTOUPDATER=1` in the Paperclip server's launchd plist EnvironmentVariables so spawned runs can't re-upgrade the shared install.

## Suggestions

1. **Surface MCP availability in run diagnostics** — record the connected/auth-required MCP server list per run (it's visible to the spawned session) so a shrunken toolset is observable without reading agent prose.
[...]

Issue #8626, opened 2026-06-25
Gate an agent's first action until configured MCP servers finish the initialize→tools/list handshake (cold-start tool-availability race)


chhabra496:
### Summary
When an agent runtime starts a run with one or more configured MCP servers, the agent can begin executing — and attempting tool calls — **before** those servers have completed the MCP `initialize` → `tools/list` handshake. During this cold-start window the not-yet-registered tools read as *unavailable*, and the agent proceeds with **degraded capability** instead of waiting for the handshake to finish.

### Impact
Intermittent, hard-to-reproduce capability loss at run start:
- A tool that resolves fine on a warm run is "missing" on a cold first start.
- The agent **silently** falls back to reduced behavior (defaults / no tool use) rather than surfacing that a configured server was not ready.
- Because it only happens in the brief connection window, it looks nondeterministic — *same agent, same config, different result*.

### Observed timing (generic)
- Cold first-start handshake: ~**5–8s** for a server to reach `tools/list` complete in a fresh process.
- Warm-path tool resolution: **p50 ~88ms**.
- Tools appear missing **only** within the cold-start window; once the handshake completes they resolve normally. So this is a connection **race**, not a missing or misconfigured server.

### Requested behavior
Gate the agent's first tool-dependent action until **each** configured MCP server is either **ready** (handshake complete) or **failed**:
1. Block the first action until every configured server reaches a terminal state (ready or failed).
2. Timeout floor of **>= 10s** before giving up on a still-connecting server.
3. On timeout, **fail loud** naming the specific server that did not become ready — do **not** silently run tool-less.
4. Distinguish **"still connecting"** from **"absent / not configured"**, so logs and the agent can tell a race from a genuinely missing server.

### Why it matters
An agent that depends on MCP tools for correctness currently cannot tell *"the tool isn't there"* from *"the tool isn't there **yet**."* A readiness gate converts a silent, intermittent degradation into either correct behavior (wait for ready) or a loud, diagnosable failure (named server timeout).

PR #11997, opened 2026-08-23
fix(tool-gateway): let gateway bearer tokens reach the MCP protocol routes

ericdfields:
## Thinking Path

> - Paperclip can connect an agent to a remote MCP tool server. The tool gateway mints a short-lived bearer token for each run, and the adapter gives that token to the agent process.
> - No agent could use any MCP tool. This was true for every connection and every agent, not only one.
> - The connection record was correct. The token was correct. The adapter command line was correct. Only the MCP handshake failed.
> - The gateway routes read their own bearer token. They do not read `req.actor`. But `actorMiddleware` runs before all routes. It saw a token it did not know, and it answered 401.
> - This pull request lets the gateway token reach the gateway route. The route still checks the token itself.
> - The benefit is that Paperclip-managed MCP servers attach again, so agents get their MCP tools.

**Duplicate search:** I searched the open and closed pull request list for MCP gateway, gateway token, and actor middleware changes. I found no other pull request for this problem. This pull request replaces #11996, which had the same commit on a branch name that contained an internal ticket id. #11996 is closed.

**What happened?**

No `mcp__*` tool is available in any agent session. This applies to every MCP connection and every agent. A tool search for MCP tools returns nothing.

**Expected behavior**

An enabled and healthy MCP connection gives the agent its tools.

**Steps to reproduce**

1. Connect a remote MCP server. Confirm the connection is `active` and `enabled`.
2. Start an agent run.
3. Search the agent session for `mcp__` tools. No tool is found.
4. Send an MCP `initialize` request to the gateway URL with the token the run minted. The server answers 401.

**Relevant logs or output**

```
POST /api/tool-gateway/gateways/<id>/mcp
Authorization: Bearer pcgw_...

HTTP/1.1 401 Unauthorized
{"error":"Agent token did not verify; obtain fresh credentials and retry"}
```

That message comes from `server/src/middleware/auth.ts`. It does not come from the gateway route.

`master` at c5050396c.

**Deployment mode**

`authenticated`. The fault also applies to `local_trusted`, because `actorMiddleware` rejects the bearer token before the deployment mode matters.

**Agent adapter(s) involved**

Not adapter-specific (core bug). The fault is in the server. Every adapter that receives a Paperclip-managed MCP server is affected.

**Additional context**

The token in that request was valid. It was not expired and not revoked. On the instance where this was found, 46 runtime gateway tokens were minted and 0 were ever used:

```sql
select count(*) total, count(last_used_at) used
  from tool_mcp_gateway_tokens where subject_type='heartbeat_run';
 total | used
-------+------
    46 |    0
```

No MCP request has ever reached the gateway service.

The cause is in `actorMiddleware`. It is mounted on the app, before all routes. It reads the bearer token, tries a board key, an agent key, and an agent JWT, and then fails the request:

```ts
if (!claims) {
  next(unauthorized(invalidAgentTokenMessage(token)));   // 401 — the route never runs
  return;
}
```

There is no branch for a gateway token. Both gateway mounts are affected: the public `/mcp/gateways/:publicId` mount and the `/api/tool-gateway/gateways/:id/mcp` mount that the heartbeat gives to adapters.

The failure is silent. Nothing writes an error. An MCP client that cannot complete `initialize` simply registers zero tools. To the agent this looks the same as "no MCP server is configured".

## What Changed

- `server/src/services/mcp-gateway-endpoints.ts` (new): holds the gateway URL builders and the path matcher together in one module.
- `server/src/middleware/auth.ts`: lets a `pcgw_`-prefixed bearer token continue to the route, but only on the two gateway protocol paths. The middleware sets `req.actor` to `"none"` first. This stops a gateway token from getting the implicit board actor that `local_trusted` mode sets.
- `server/src/services/heartbeat.ts`: builds the runtime MCP server URL with the shared builder. The URL does not change.
- `server/src/__tests__/mcp-gateway-token-auth.test.ts` (new): 7 tests for the authentication boundary.
- `server/src/__tests__/mcp-gateway-endpoint-coverage.test.ts` (new): a guard that keeps the builder and the matcher in agreement.

## Verification

```
npx vitest run server/src/__tests__/mcp-gateway-token-auth.test.ts \
               server/src/__tests__/mcp-gateway-endpoint-coverage.test.ts
npx tsc --noEmit          # in server/
```

- 11 new tests pass. All 29 server auth and gateway test files pass (419 tests). Typecheck is clean.
- Three of the new tests are negative tests. A gateway token on a route that is not a gateway route still gets 401. A token that is not a gateway token on a gateway route still gets 401. In `local_trusted` mode a gateway token gets no board actor.
- The tests were mutation-tested. When the carve-out is disabled, 4 of the 7 boundary tests fail. The 3 negative tests still pass. This shows they guard the boundary and not the success path.
- An end-to-end check was run against a live instance database. A real minted token was sent through the real routes:

```
initialize -> 200 {"jsonrpc":"2.0",...,"serverInfo":{"name":"Paperclip MCP Gateway","version":"1.0.0"}}
tools/list -> 200
tool names: [ ...:create-link, ...:delete-link, ...:list-my-links, ...:promote-link, ...:revise-link ]
```

## Risks

Low risk, but the change is in authentication code. Please review the boundary.

- The carve-out is narrow. It applies only when the token starts with `pcgw_` **and** the path is one of the two gateway protocol paths. All other paths and all other token shapes keep their current behaviour.
- The middleware sets `req.actor` to `"none"` before it calls `next()`. Without this line, a gateway token in `local_trusted` mode would get the implicit board actor. A test covers this case.
- The gateway routes do not become open. They still validate the token, the gateway status, the token expiry, the revocation state, and the allowed actions.
- A wider path pattern would be a security problem. The coverage test asserts that neighbouring authenticated routes, such as `/api/tool-gateway/gateways/:id/tokens` and `/api/tool-gateway/audit`, are **not** matched.
- No database migration. No change to any public URL.

## Model Used

Claude Opus 5 (`claude-opus-5[1m]`), 1M context window, extended thinking, with tool use and code execution.

## Checklist

### DIFF
--- a/server/src/__tests__/mcp-gateway-endpoint-coverage.test.ts
+++ b/server/src/__tests__/mcp-gateway-endpoint-coverage.test.ts
@@ -1,0 +1,56 @@
+import { readFileSync } from "node:fs";
+import { fileURLToPath } from "node:url";
+import { describe, expect, it } from "vitest";
+import {
+  isMcpGatewayProtocolPath,
+  mcpGatewayApiEndpointPath,
+  mcpGatewayPublicEndpointPath,
+} from "../services/mcp-gateway-endpoints.js";
+
+const GATEWAY_ID = "819f6caa-92bb-41ee-a33c-3e35bd577a8f";
+const GATEWAY_PUBLIC_ID = "gw_61df2b661160478d9091986dbe51b2d8";
+
+function readSource(relativePath: string): string {
+  return readFileSync(fileURLToPath(new URL(relativePath, import.meta.url)), "utf8");
+}
+
+/**
+ * Category safeguard.
+ *
+ * The failure class: Paperclip mints a credential for one of its own endpoints,
+ * and something upstream of the route rejects it. That failure is silent — an
+ * MCP client that cannot `initialize` just registers zero tools — so it has to
+ * be caught structurally rather than observed.
+ */
+describe("MCP gateway endpoint coverage", () => {
+  it("admits every endpoint path Paperclip builds for a gateway", () => {
+    expect(isMcpGatewayProtocolPath(mcpGatewayApiEndpointPath(GATEWAY_ID))).toBe(true);
+    expect(isMcpGatewayProtocolPath(mcpGatewayPublicEndpointPath(GATEWAY_PUBLIC_ID))).toBe(true);
+  });
+
+  it("does not admit neighbouring tool-gateway routes", () => {
+    // These are ordinary board/agent-authenticated routes. Widening the matcher
+    // to cover them would hand a gateway token API surface it must not reach.
+    expect(isMcpGatewayProtocolPath(`/api/tool-gateway/gateways/${GATEWAY_ID}`)).toBe(false);
+    expect(isMcpGatewayProtocolPath(`/api/tool-gateway/gateways/${GATEWAY_ID}/tokens`)).toBe(false);
+    expect(isMcpGatewayProtocolPath("/api/tool-gateway/tools/call")).toBe(false);
+    expect(isMcpGatewayProtocolPath("/api/tool-gateway/audit")).toBe(false);
+    expect(isMcpGatewayProtocolPath("/api/agents/me")).toBe(false);
+  });
+
+  it("keeps the runtime MCP server URL on a path the actor middleware admits", () => {
+    // `buildPaperclipRuntimeMcpServers` hands this URL to every adapter. If it
+    // is ever rebuilt inline again, this assertion is the thing that notices.
+    const heartbeat = readSource("../services/heartbeat.ts");
+    expect(heartbeat).toContain("mcpGatewayApiEndpointPath(gateway.id)");
+    expect(heartbeat).not.toMatch(/`\/api\/tool-gateway\/gateways\/\$\{[^}]+\}\/mcp`/);
+  });
+
+  it("keeps the gateway protocol routes mounted on the advertised paths", () => {
+    // The route file declares its own mount paths and echoes them back to
+    // clients in the discovery response. Both must stay in the admitted set.
+    const routes = readSource("../routes/tool-gateway.ts");
+    expect(routes).toContain('router.post("/mcp/gateways/:gatewayPublicId"');
+    expect(routes).toContain('router.post("/tool-gateway/gateways/:gatewayId/mcp"');
+  });
+});
--- a/server/src/__tests__/mcp-gateway-token-auth.test.ts
+++ b/server/src/__tests__/mcp-gateway-token-auth.test.ts
@@ -1,0 +1,125 @@
+import express from "express";
+import request from "supertest";
+import { describe, expect, it } from "vitest";
+import { agentApiKeys, agents, boardApiKeys, heartbeatRuns } from "@paperclipai/db";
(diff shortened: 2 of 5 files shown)


0xCVH:
**Independent reproduction — a second instance, and a published release.** It confirms the diagnosis in this PR.

- `@paperclipai/server` `2026.819.0-canary.1` — a published release, not `master`. Mode `authenticated`.
- Three gateways, each in front of a different remote MCP server. More than one agent, and more than one adapter.
- Every `mcp__*` tool was absent from every agent session for six days, from 2026-08-18 to 2026-08-24.

**A cheap check that the middleware rejects before routing.** Send a `pcgw_`-shaped bearer token to a path that does not exist:

```
GET /api/this-route-does-not-exist    no bearer     -> 404
GET /api/this-route-does-not-exist    pcgw_ bearer  -> 401 {"error":"Agent token did not verify; ..."}
```

No route runs, so the global middleware is the source. The check needs no valid token, so any reviewer can run it.

**#10346 looks like a duplicate.** It reports a `codex_local` run that issued per-run gateway credentials, never used them, and created no gateway session. That is the same signature as the 46-minted / 0-used query in your PR body. The issue blames the codex adapter. If you agree, linking it closes a misfiled issue. #11203 is probably separate — its A/B result shows the CLI engine did reach the gateway, which cannot happen while every gateway bearer token gets a 401.

Please keep this PR as it is. The silent-failure half needs its own change: an agent with zero `mcp__*` tools still cannot tell "not provisioned" from "auth broke five minutes ago". I will file that separately. See also #8626 and #10102.

Thanks for finding and fixing this. It unblocks a real outage for us.

vbalko-claimate:
**Independent confirmation from a self-hosted deployment** — this bug is live on `master` around `88a0f885e`, and this PR fixes it.

**Impact:** every governed MCP tool was absent from every agent's catalog for ~5 days (20+ consecutive agent runs). Agents received a complete `mcp-config.json` (each connected server as `type:"http"`, gateway URL, `pcgw_` bearer), but every `POST /api/tool-gateway/gateways/<id>/mcp` returned `401 {"error":"Agent token did not verify; obtain fresh credentials and retry"}`.

**Root cause matches this PR:** the global `actorMiddleware` runs the `pcgw_` gateway token through the agent-credential path (board key / agent API key / agent JWT) and hard-rejects it before `handleMcpGatewayProtocol` ever runs. The token itself is valid — it is present and unexpired in `tool_mcp_gateway_tokens` during the run — so the only problem is that the request never reaches the gateway's own verification. `tool_gateway_sessions` being empty is a red herring here: the named-gateway path stores tokens in `tool_mcp_gateway_tokens`, not that table.

**Fix verified:** I applied the same idea locally (admit gateway tokens to the gateway protocol paths in `actorMiddleware`; keep `heartbeat.ts` on an admitted path) and redeployed. Result:

- gateway requests: `401` → `200`
- `tools/list` now returns the expected tools (13 tools from one connection)
- no more `did not verify` on gateway requests

Your path-scoped `isMcpGatewayProtocolPath` matcher is the safer approach — it admits gateway tokens only on the gateway protocol paths, rather than by token prefix alone.

I filed the same bug independently as #12294 before finding this PR; I will close that as a duplicate. Real users are blocked — it would be great to see this merged. Thanks for the fix!

PR #4974, opened 2026-05-01
plugins: add Telegram Notifier plugin

Vitalini:
## Thinking Path

> - Paperclip is the open source app people use to manage AI agents for work
> - Operators run Paperclip in a browser tab and miss decision-needed pings when they step away — approvals stall, run failures sit unnoticed, and agent comments waiting on a human pile up
> - Paperclip already has a clean plugin model with event subscriptions, jobs, tools, settings-UI surfaces, and per-instance state — the right extension point for a notification bridge
> - Telegram is a free, ubiquitous, push-capable channel with a well-documented Bot API and inline keyboards for action buttons
> - This pull request adds a first-party **Telegram Notifier** plugin that forwards core event types to a paired chat per company, accepts `/new` and `/inbox` commands back, supports inline-keyboard reassignment and plan approvals, lets operators reply to comments straight from Telegram, and ships an opt-in morning digest
> - The benefit is operators can run multi-company Paperclip instances and stay caught up with each company from their phone, with a pairing handshake secure enough that a leaked bot token alone can't hijack notifications

## Linked Issues or Issue Description

No existing public GitHub issue — describing inline per the feature-request template (`.github/ISSUE_TEMPLATE/feature_request.yml`):

- **Subsystem affected:** Plugins / notifications.
- **Problem or motivation:** Decision-needed signals in Paperclip (approvals, run failures, agent comments awaiting a human, new mentions) are only visible inside the dashboard. An operator running one or more companies who is away from the browser has no push path, so approvals and blocked runs stall until they happen to check back.
- **Proposed solution:** A first-party plugin that bridges those events to Telegram per company over the Bot API, with inbound commands and inline-keyboard actions (approve/decline, reassign, reply) so an operator can both *see* and *act* from their phone. Built entirely on the existing plugin surface (event subs, cron job, tools, settings page, per-instance state) — no core changes beyond one purely-additive event-mapping hook.
- **Alternatives considered:** Email notifier (no inline actions, higher latency — see the separate email-notifier effort in #1157/#402); Discord/webhook bridge (#398, #3769 cover Discord/webhook angles); a generic webhook the operator wires to their own bot (more setup, no first-party UX). Telegram was chosen for free push, inline keyboards, and a simple Bot API.
- **Roadmap alignment:** Not core feature work — this is an extension through the public plugin system, so it does not overlap with planned `ROADMAP.md` items.

[...]

PR #5516, opened 2026-05-08
feat: add Paperclip inbox Telegram notifier

mamercad:
## Thinking Path

> - Paperclip is the control plane for AI-agent companies, so operator-visible alerts need to be routable outside the web UI when the board needs attention.
> - The relevant subsystem here is Paperclip inbox state, specifically the company-scoped `sidebar-badges` signal that already aggregates failed runs, alerts, join requests, and approvals.
> - EMT-15 asked for the smallest reliable Paperclip inbox -> Telegram bridge owned in the mamerbot Paperclip repo rather than ad hoc manual checking.
> - The main product risk was notification spam, so the implementation needed persisted anti-spam state and deterministic rules around first-positive and count-change transitions.
> - Paperclip already has a server package and a root package command surface, so the cleanest implementation was shared notification decision logic in `server/src/services`, a script entrypoint in `scripts/`, and a root package command for operations.
> - This pull request adds that notifier slice, the package command, docs, and targeted tests including a repo-root smoke test for the command wiring.
> - The benefit is a reproducible, repo-owned path for delivering Paperclip inbox events to Mark on Telegram with grounded verification and no repeat-notify spam on unchanged counts.

## What Changed

- Added `server/src/services/inbox-telegram-notifier.ts` with badge normalization, anti-spam state handling, notification decision logic, and Telegram message formatting.
- Added `scripts/paperclip-inbox-telegram-notifier.ts` to fetch `sidebar-badges`, persist state, and send Telegram messages only when the decision logic says to notify.
- Added a root package command `notify:paperclip-inbox-telegram` so the notifier can be run from the repo root.
- Added targeted tests for decision logic and a repo-root package-script smoke test.
- Added `docs/runbooks/paperclip-inbox-telegram-notifier.md` documenting configuration, behavior, and verification.

## Verification

- `pnpm vitest run server/src/__tests__/inbox-telegram-notifier.test.ts server/src/__tests__/paperclip-inbox-telegram-notifier-script.test.ts`
- Controlled positive smoke previously executed on Raven against a stubbed `sidebar-badges` response and live Telegram bot credentials, producing one notifying run and one unchanged-count no-op run.

## Risks

- Low to moderate: the notifier depends on the current `sidebar-badges` response shape remaining stable.
- Misconfigured env vars or state-file paths will prevent delivery; the runbook documents the required settings.
- This adds an outbound Telegram side effect, so operators should use a dedicated state file per deployment context to avoid duplicate notifications across multiple schedulers.

[...]

PR #12004, opened 2026-08-23
feat(plugins): add Telegram Bot Control plugin

jotavgomes:
## Thinking Path

> - Operators check on their company (agent status, recent activity) by opening the dashboard in a browser
> - There is no lightweight way to check on a company from a phone without opening the full web app
> - Telegram is a free, ubiquitous chat channel with a simple Bot API, well suited for a minimal on-demand status check
> - This pull request adds a first-party **Telegram Bot Control** plugin: an operator links a Telegram chat to a company with `/connect <company name>`, then uses `/status` to see agent counts by status
> - The benefit is a fast, zero-setup way to check on a company from anywhere, without opening the dashboard

No existing public GitHub issue. Describing in-PR per the feature-request template (`.github/ISSUE_TEMPLATE/feature_request.yml`):

- **Subsystem affected:** Plugins.
- **Problem or motivation:** Checking on a company (is an agent stuck, how many agents are active) currently requires opening the Paperclip dashboard in a browser. There is no lightweight, phone-friendly way to get a quick status read.
- **Proposed solution:** A minimal first-party plugin that links a Telegram chat to one company (`/connect <company name>`) and reports agent counts by status on request (`/status`). Built entirely on the existing plugin surface (a single scheduled job, per-company config, per-company state) — no core changes.
- **Alternatives considered:** The much larger Telegram Notifier plugin in #4974 (open, unmerged) covers a different problem: push notifications, inline-keyboard approvals, comment replies, and a morning digest, built on event subscriptions and several core `activity-log.ts` changes. This plugin is deliberately smaller in scope: on-demand pull (`/status`), not push notifications, and it makes no changes outside the plugin package. The two are complementary rather than duplicates — an operator could run either or both.
- **Roadmap alignment:** Not core feature work — this is an extension through the public plugin system, so it does not overlap with planned `ROADMAP.md` items.

**Dedup search (PR list):** searched `telegram` across open and closed PRs. Related PRs found: #4974 (Telegram Notifier — push notifications/approvals, much broader scope, discussed above), #2263 and #2405 (example plugins, not first-party), #7605 (closed — a polling invocation-scope fix, not a full plugin). None duplicate this plugin's narrow connect/status scope.

## What Changed

- New plugin package `@paperclipai/plugin-telegram` under `packages/plugins/paperclip-plugin-telegram/`. Manifest declares `companies.read`, `agents.read`, `http.outbound`, `plugin.state.read`, `plugin.state.write`, `jobs.schedule` capabilities and a single `poll-updates` cron job (every minute).
- **Per-company config and state.** The plugin's `botToken` is resolved per company (`ctx.config.get(companyId)`), since the host authorizes `config.get` only inside a company-bound invocation — `setup()` runs once globally with no company context and cannot call it. All config-dependent work (command registration, polling) lives in the scheduled job handler instead, which the host does bind per company. Chat-to-company links and Telegram's update offset are stored in company-scoped plugin state (`ctx.state`, `scopeKind: "company"`).
- **Long-poll on a schedule.** Plugin jobs are host-scheduled, not long-running processes, so `getUpdates` is called once a minute with a ~25s server-side timeout, rather than held open indefinitely.
- **Commands.** `/connect <company name>` links the current chat to the company whose name contains the given text (validated against the plugin's own bound company, since one bot token is one company's config) and replies with a confirmation. `/status` reports the linked company's agent counts by status, or asks the chat to `/connect` first if it has not linked yet.
- Idempotent command registration: `setMyCommands` is only called again when the configured token changes, tracked via a small state flag.
- Worker build switched from plain `tsc` to `esbuild` (`createPluginBundlerPresets` from `@paperclipai/plugin-sdk/bundlers`), producing a self-contained `dist/worker.js`. `@paperclipai/shared`'s `exports` field resolves straight to `.ts` source with no built-dist fallback, and the spawned plugin-worker subprocess (unlike the main server process) has no TS-aware loader, so a bundled, dependency-free worker is required for the worker to start at all.
- `Dockerfile`: added the one `COPY packages/plugins/paperclip-plugin-telegram/package.json ...` line the deps stage needs, mirroring the existing lines for the other bundled, UI-installable first-party plugins (`paperclip-plugin-fake-sandbox`, `plugin-llm-wiki`, `plugin-workspace-diff`).
- 12 vitest unit tests covering manifest capabilities/job declaration, `setup()` making no config/HTTP calls, first-poll command registration, second-poll not re-registering, skipping companies with no token configured, `/connect` + `/status` happy path, `/connect` name-mismatch rejection, `/status` before `/connect`, ignoring unrelated messages, and skipping archived companies.

## Verification

```bash
pnpm --filter @paperclipai/plugin-telegram typecheck   # passes
pnpm --filter @paperclipai/plugin-telegram test         # 12 passed
pnpm --filter @paperclipai/plugin-telegram build         # esbuild bundle succeeds
node ./scripts/release-package-map.mjs check             # manifest OK (package is private, excluded)
docker compose build server                               # clean image build succeeds with the plugin present
```

End-to-end manually verified on a local Docker Compose instance with a real Telegram bot token from @BotFather:

- Installed via the plugin manager UI (Company Settings → Plugins → Install), configured the bot token, enabled the plugin.
- Server logs show `Telegram bot plugin started` on activation and `Bot commands registered with Telegram` on the first poll after a token is configured.
- Sent `/connect <company name>` from a real Telegram chat: bot replied with a confirmation; the chat id was persisted in per-company plugin state (`chat-links`).
- Sent `/status`: bot replied with the company's agent counts by status.
- Confirmed the update offset advanced in plugin state after each processed message, and a second poll with the same token did not re-call `setMyCommands`.

## Risks

Low. The plugin is additive and self-contained: no changes outside the plugin package and one Dockerfile line. Capabilities are narrow (no agent invoke, no pause, no board API key) — it can link a chat to a company and read agent status, nothing else. A leaked bot token only exposes read-only agent-status info for whichever company configured it, scoped by the `/connect` name check. State is namespaced per company and cleared on uninstall like any other plugin state.

## Model Used

- Claude Sonnet 5 (`claude-sonnet-5`), used interactively end-to-end: initial implementation, three rounds of debugging against a real running instance (ESM module resolution for the worker subprocess, a `pnpm install` recovery inside a live container, and the `config.get` company-context authorization model), then verified live against the real Telegram Bot API.

## Checklist

- [x] My branch name describes the change (`feat/telegram-bot-control-plugin`) and contains no internal Paperclip ticket id or instance-derived details
- [x] I have updated relevant documentation to reflect my changes (README.md inside the package)


jotavgomes:
**Possible overlap with a concurrent PR — flagging for reviewer decision, not a code issue.**

#4974 ("plugins: add Telegram Notifier plugin", open, unmerged) also adds a first-party Telegram integration, under a different package name (`@paperclipai/plugin-telegram-notifier` vs. this PR's `@paperclipai/plugin-telegram`) and a substantially larger scope: push notifications on core events, inline-keyboard plan approvals, comment replies from Telegram, a morning digest, and several changes outside the plugin folder (`activity-log.ts`, a new issue status).

This PR is deliberately smaller — on-demand `/connect` + `/status` (and now `/task`, see the follow-up branch), no push notifications, no changes outside the plugin package. The two don't conflict technically (different packages, different plugin ids), but they overlap in *purpose* enough that a reviewer may want to pick one direction, merge both, or ask for them to be reconciled rather than shipping two separate first-party Telegram integrations. Noting it here so that decision is made deliberately rather than by whichever merges first.

See also #4974's own dedup-search section, which lists other prior Telegram attempts (#2263, #2405, #7605, and closed unmerged #5516/#407/#405).

jotavgomes:
**Addressed the Greptile-flagged auth bypass.** Pushed a new commit to this branch.

`/connect` now requires `/connect <company name> <code>`, not just the name. The code:
- is generated by the plugin itself (crypto-random, 8 chars, an alphabet without visually ambiguous characters)
- is only visible in this plugin's activity log inside Paperclip — a surface that requires an authenticated session to reach
- is single-use: consumed on the first successful `/connect`, with a fresh one lined up immediately after (and on each poll tick if none is outstanding)

So a guessed or previously-known company name is no longer sufficient on its own, and a code that leaked (screenshot, forwarded message) stops being usable the moment a real connection goes through.

Also fixed the two README claims Greptile separately flagged as contradicting the implementation (state scope, command-registration timing).

New tests: connect with the right name+code succeeds; wrong name doesn't burn the code; missing code and wrong code are both rejected without linking; a code is provably single-use (a replay from a different chat fails, the already-issued next code still works). 14/14 tests passing, typecheck and build clean. Verified live against the real running instance — the code generation and logging both work against the actual deployed system, not just the test harness.

Not addressed in this update (Greptile's review also flagged these, out of scope for this pass — flagging for a possible follow-up):
- The scheduled per-company iteration loop calls `ctx.config.get` for every active company regardless of whether this plugin is installed for it; if that call can throw for a company without the plugin configured, one such company could abort the whole poll tick for every other company. Worth verifying separately.
- Update-offset checkpointing saves the offset once per batch, not per message — if a Telegram API call fails partway through a multi-update batch, already-processed updates in that batch get reprocessed on the next poll.

jotavgomes:
**Addressed the two remaining Greptile findings** (polling-scope abort and batch-checkpoint replay) with a new commit.

- \`forEachConfiguredCompany\`: a \`config.get\` or poll failure for one company is now caught, logged, and skipped — it no longer aborts polling for every other company on the same tick.
- Update offset: checkpointed after each individual update instead of once per batch — a failure partway through a multi-update batch (e.g. one \`sendMessage\` call erroring) no longer causes already-handled updates earlier in that same batch to be replayed on the next poll.

New tests for both: one company's config throwing doesn't stop another company's poll; a mid-batch \`sendMessage\` failure still leaves the offset advanced past the update that succeeded before it. 16/16 tests passing, typecheck clean.

This closes out everything from the last two Greptile reviews on this PR.

jotavgomes:
**Addressed the poison-update finding from the latest review.**

A message that fails the same way every time it's processed (not a transient blip) never got checkpointed, so every later poll retried it and never reached any update behind it — the company's queue would jam forever.

\`handleMessage\` failures are now caught per-update: logged, and the offset still advances past that update either way. A dropped reply is a better failure mode than a permanently stuck queue for the rest of the chat's commands.

New test: a permanently-failing update is dropped and logged, and a later update in the same poll still gets processed. 17/17 tests passing, typecheck clean.

jotavgomes:
Re: the pending-reply queue overflow Greptile flagged in \`enqueuePendingReply()\` (\`worker.ts\`).

Two fixes landed on top of the original poison-update fix in response to Greptile's review:
1. A reply that fails delivery is retried once immediately (250ms), and if that still fails, queued in per-company state and retried on every subsequent poll tick (~once a minute) instead of being dropped after a single attempt.
2. The queue's cap was 20, which a company with real message volume during a sustained outage could plausibly hit — raised to 200, so overflow now needs a genuinely pathological outage (many ticks, heavy concurrent volume) rather than a realistic one.

I'm not chasing this to a literal guarantee. A bot token that goes **permanently** dead (not just a temporary outage — actually revoked or invalid) will still eventually overflow a 200-entry cap and drop the oldest queued reply. Building fully unbounded, guaranteed-eventual-delivery queuing for a lightweight bot-control plugin felt like more machinery than the problem warrants, and that residual case is now documented inline at \`MAX_PENDING_REPLIES\`. Flagging it explicitly here rather than leaving it implicit.

jotavgomes:
Re: the overlapping-poll-runs concern in the latest review (connect-code consumption and pending-reply writes as non-atomic read/write).

Checked this against the host's actual job scheduler, not just this plugin's own code — \`server/src/services/plugin-job-scheduler.ts\`:

- Line 292-298: before dispatching a due job, the scheduler checks \`activeJobs.has(job.id)\` and skips dispatch entirely if that job is already running ("overlap prevention").
- Line 455-458: a manual trigger is rejected the same way — \`Job "\${job.jobKey}" is already running — cannot trigger while in progress\`.
- Line 419-421: \`activeJobs.delete(jobId)\` runs in a \`finally\` block, so the job can't be marked "not running" until the handler has fully returned (success or failure).

\`POLL_JOB_KEY\` is registered once per plugin instance (\`ctx.jobs.register(POLL_JOB_KEY, () => pollUpdates(ctx))\` in \`setup()\`), not once per company — \`pollUpdates()\` itself loops over every configured company inside a single execution. So there is exactly one job row for this job key, and the scheduler guarantees at most one in-flight execution of it at any time, scheduled or manually triggered.

Two overlapping calls to \`pollUpdates()\` — and therefore two overlapping calls to \`consumeConnectCode()\` or the pending-reply read/write — isn't a race that can happen in this system; it's prevented at the host level, one layer above what's visible from \`worker.ts\` alone. I don't think this needs a fix (there's nothing to synchronize against), but flagging the reasoning explicitly in case I'm missing a code path where the host's guarantee doesn't apply.

jotavgomes:
@greptileai review

jotavgomes:
Re: "the logged bearer code is readable without company scoping" — this one was real, and more serious than the earlier findings on this thread. Verified by reading the actual host code, not just this plugin's:

- The worker-side \`ctx.logger\` client (\`worker-rpc-host.ts\`) sends \`{level, message, meta}\` with **no companyId field at all** — it's not possible to attribute a logger call to a company from a plugin worker.
- The host writes exactly what it receives (\`plugin-host-services.ts\`), so every \`ctx.logger.*\` entry lands with \`companyId: null\`.
- \`GET /plugins/:pluginId/logs\` only calls \`assertBoardOrgAccess\`, which checks "is this board user a member of *some* company," not which company's logs they're asking for. No companyId filter exists on that route at all.

Net effect: the connect code — the entire security boundary the earlier /connect fix in this thread depends on — was visible to *any* board member on a shared instance with the Telegram plugin configured for more than one company, not just the target company's. Fixed by switching the disclosure to \`ctx.activity.log\` instead, which requires a companyId and whose read side (\`GET /companies/:companyId/activity\`) calls \`assertCompanyAccess\` — genuinely checks the specific company. Added the \`activity.log.write\` capability for it. Test added that fails against the old \`ctx.logger\`-based code and passes against the fix (confirmed via revert-and-retest).

Re: "connect-code consumption remains non-atomic under overlapping runs" — this is the same claim from two reviews ago. I posted the evidence for why it doesn't apply in an earlier comment on this thread (the host's job scheduler enforces one in-flight execution per job id — \`plugin-job-scheduler.ts\` lines 292-298, 419-421, 455-458) and haven't seen it engaged with. Re-flagging in case it was missed rather than considered and rejected — happy to dig further if there's a code path I'm not seeing where that guarantee doesn't hold.

jotavgomes:
@greptileai review

jotavgomes:
Re: the last finding — flush persisted removals in one batched write at the end, so a state-store failure after N successful sends re-queued all N for resend. Fixed: each removal is now persisted immediately after its own successful send, so a failed write duplicates at most the one reply it belongs to, not the whole batch. Test added and confirmed via revert-and-retest.

Stepping back: this thread has now closed, in order — the original poison-update queue-jam, a lost /connect confirmation on a single transient blip, the same on a longer outage (queue + next-tick retry), a queue-overflow edge case, a compound state-write failure on top of a delivery failure, a genuine cross-company secret disclosure (the connect code was reachable by any board member on the instance, not just the target company — real, and now fixed via the company-scoped Activity feed instead of ctx.logger), and now a duplicate-delivery edge case on cleanup-write failure.

I don't think it's productive to keep chasing this to a literal formal guarantee — the remaining space is inherent to any "best-effort delivery + eventually-consistent local state" design (a process crash mid-write, a permanently dead bot token, etc.), and closing all of it would mean building actual distributed-transaction machinery for what's meant to be a lightweight bot-control plugin. Every failure mode with a real, boundable fix has one now, with a test proving it's real (revert-and-retest) and a test proving the fix works. What's left is disclosed inline at MAX_PENDING_REPLIES and in this thread, not silently accepted.

jotavgomes:
@greptileai review

jotavgomes:

Issue #5, opened 2026-03-04
Terminated agents included in org chart and agents list endpoints


numman-ali:
## Bug Description

When a hire approval is rejected, the agent record is set to `terminated` status. However, terminated agents are still returned by the agents list and org chart API endpoints, even though other parts of the codebase already exclude them.

## Reproduction Steps

1. Create an agent via hire approval flow (status becomes `pending_approval`)
2. Reject the hire approval — agent status becomes `terminated`
3. Call `GET /api/companies/:companyId/agents` — terminated agent is included in the response
4. Call `GET /api/companies/:companyId/org` — terminated agent appears in the org tree

**Expected:** Terminated agents should be excluded from these endpoints by default.

### Agents list — `server/src/services/agents.ts:254-256`

The `list()` method queries all agents for a company with no status filtering:

```typescript
list: async (companyId: string) => {
  const rows = await db.select().from(agents).where(eq(agents.companyId, companyId));
  return rows.map(normalizeAgentRow);
},
```

### Org chart — `server/src/services/agents.ts:471-472`

The `orgForCompany()` method also queries all agents with no status filtering:

```typescript
orgForCompany: async (companyId: string) => {
  const rows = await db.select().from(agents).where(eq(agents.companyId, companyId));
  // ... builds tree from unfiltered results
},
```

### Additional context

- `DELETE /api/agents/:id` requires `assertBoard()` (board-only), so managers/CEOs cannot delete terminated agents themselves — they get 403.
- The UI (`ui/src/pages/Agents.tsx`) has a client-side `showTerminated` toggle that hides terminated agents by default, but the API still returns them.

## Existing precedent for filtering

The codebase already excludes terminated agents in several places:

| Location | How |
|---|---|
| `services/agents.ts:536` (`resolveByReference`) | `.filter(agent => agent.status !== "terminated")` |
| `services/sidebar-badges.ts:36` | `not(eq(agents.status, "terminated"))` in SQL |
| `services/company-portability.ts:560` | `.filter(agent => agent.status !== "terminated")` |
| `services/heartbeat.ts:2037` | `if (agent.status === "terminated") continue` |
| `ui/src/pages/AgentDetail.tsx:296` | `.filter(a => a.status !== "terminated")` for direct reports |

### Why this option

| Option | Pros | Cons |
|---|---|---|
| **1. Auto-cleanup (delete on rejection)** | Clean DB | Loses audit trail, breaks FK references, irreversible |
[...]

Issue #3, opened 2026-03-04
Opencode support


Konan69:
opencode support would open up access to multiple models and multiple other providers,  although it might be tricky to implement, i am open to working on the impl.

cryppadotta:
Okay I love this idea - I don't use OpenCode, but we should definitely support it. What would be the best "ergonomics" in your opinion?

Konan69:
my implementation allows this by bringing in the providers and models you already have connected, but it wasnt merged in
cc @cryppadotta

Konan69:
coming to a pr soon 


> how do i enable this?

GollyJer:
Following... excited to try this out with opencode with my Black account. Multi-model orchestration would be awesome.

poori:
its ready. git pull the latest and try it out. i'm using it now

TaQuangKhoi:
Thanks you guys all for this feature.

Issue #2, opened 2026-03-04
wsl 2 + zsh


lmvdz:
```
Paperclip server failed to start.
Postgres init script exited with code 1. Please check the logs for extra info. The data directory might already exist.
[20:46:02] INFO: Using embedded PostgreSQL because no DATABASE_URL set (dataDir=/home/lars/.paperclip/instances/default/db, port=54329)
[20:46:02] ERROR: Embedded PostgreSQL failed; showing buffered startup logs
    phase: "initialise"
    recentLogs: [
      "The files belonging to this database system will be owned by user \"lars\".",
      "This user must also own the server process."
    ]
    err: "Postgres init script exited with code 1. Please check the logs for extra info. The data directory might already exist."
```

jmartsch:
@cryppadotta I just said to claude "I have an error when I try to start paperclip" and pasted this error:
```
Starting Paperclip server...
  Paperclip server failed to start.
  Postgres init script exited with code 1. Please check the logs for extra info. The data directory might already
  exist.
  [20:32:45] INFO: Using embedded PostgreSQL because no DATABASE_URL set
  (dataDir=/home/jmartsch/.paperclip/instances/default/db, port=54329)
  [20:32:45] ERROR: Embedded PostgreSQL failed; showing buffered startup logs
      phase: "initialise"
      recentLogs: [
        "The files belonging to this database system will be owned by user \"jmartsch\".",
        "This user must also own the server process."
      ]
      err: "Postgres init script exited with code 1. Please check the logs for extra info. The data directory might
   already exist."
```
Then claude tried to remove the database `rm -rf /home/jmartsch/.paperclip/instances/default/db`

it told me that the initdb library is not available on my system and tried to figure out how to fix it.
then it suddenly found the binaries and made progress, but since claude inited the database the role "paperclip" was missing. So claude created the role and the database. 

After that, paperclip ran.

PR #12007, opened 2026-08-23
fix(db): stream JS-engine backups through gzip, add atomic write+verify and disk-floor refusal

dzianisv:
## What happened?

`packages/db/src/backup-lib.ts::runDatabaseBackup`'s JS-fallback backup engine (used whenever `backupEngine==="javascript"`, whenever `backupEngine==="auto"` falls back after `pg_dump` fails, or whenever `excludeTables`/`nullifyColumns` options are set) wrote the entire SQL dump as **plaintext** to a `.sql` file, then gzipped that file to the final `.sql.gz` name and deleted the raw file only after the whole dump finished. For the full duration of the backup, an uncompressed file roughly the size of the live database sat on disk. On a real deployment this grew to 7.4 GiB and dropped free disk space on the host to 175 MiB, degrading the whole box.

Separately, neither backup engine (`pg_dump`-based or JS-fallback) wrote atomically: both wrote directly to the final `<prefix>-<timestamp>.sql.gz` name, with no staging file and no post-write integrity check. A crash or failure mid-write left a truncated/corrupt file at that name, and every downstream reader (health checks, retention pruning, restore tooling) had no way to tell it apart from a valid completed backup.

Finally, there was no pre-flight check anywhere that compared the live database's size against the backup directory's free disk space before starting a write.

## Expected behavior

- The JS-fallback engine should stream its output through gzip as it is produced, the same way the `pg_dump` path already streams `pg_dump`'s stdout through `zlib.createGzip()` — no bare uncompressed `.sql` file should ever exist on disk during a backup.
- Both engines should write to a temporary name, verify the completed gzip stream is well-formed, and only then atomically rename it to the final `.sql.gz` name — so the final name is never observable in a partially-written or corrupt state.
- Before writing anything, the backup should compare the source database's on-disk size against the backup directory's free space and refuse to start (throwing before creating any file) if completing the backup would drop free space below a configurable floor.
- A failed backup should clean up only its own temporary file and must never touch previously-completed `.sql.gz` backups.

## Steps to reproduce

1. Run `runDatabaseBackup` with `backupEngine: "javascript"` (or with `excludeTables`/`nullifyColumns` set) against a Postgres database with a non-trivial amount of data.
2. While the dump is in progress, list the backup directory.
3. Observe a bare `paperclip-<timestamp>.sql` file present at close to the live database's uncompressed size — this is the intermediate plaintext file that the fix in this PR eliminates.

`master` prior to this PR (repro verified via `git stash`/checkout against the pre-fix `backup-lib.ts`).

## What Changed

- **JS engine now streams through gzip.** Statement text is piped through `zlib.createGzip()` straight to disk as it's produced, the same principle as the existing `pg_dump` stdout piping. No bare uncompressed `.sql` file is ever materialized.
- **Atomic write + verify for both engines.** Both engines write to `<prefix>-<ts>.sql.gz.tmp`. After the write completes, gzip integrity is verified via a full decompress-read (`createGunzip()` piped to a draining sink). Only then is the `.tmp` file renamed to the final `.sql.gz` name — the final name is never observable in a partially-written state.
- **Pre-flight disk-floor refusal.** Before touching disk, queries live DB size (`pg_database_size(current_database())`) and free space on the backup directory's filesystem (`fs.promises.statfs`), and throws before creating any file if projected free space after the backup would drop below a floor. New `RunDatabaseBackupOptions.minFreeBytesAfter`, defaulting to the `PAPERCLIP_BACKUP_MIN_FREE_BYTES` env var, then a built-in 256 MiB floor. Also exposes `statfs` as a test-injection hook. This is defense-in-depth — the JS-engine streaming fix above is the primary fix, since the JS engine no longer materializes a full raw copy.
- **Failure cleanup** deletes only the `.tmp` file being written; previously-completed `.sql.gz` backups are never touched.
- **Retention ordering unchanged** — `pruneOldBackups` still runs immediately after a successful rename.

## Tests (`packages/db/src/backup-lib.test.ts`, extends the existing embedded-postgres harness)

- **Red repro, confirmed failing on unmodified `master`** (`git stash` verified locally) — asserts no bare `.sql` file ever exists on disk during a `backupEngine: "javascript"` run; fails on `master` with a real `paperclip-no-raw-sql-test-*.sql` file observed mid-write, passes after the fix.
- **Atomicity/integrity:** forces a `pg_dump` failure (`PAPERCLIP_PG_DUMP_PATH=/bin/false`, `backupEngine: "pg_dump"`) and asserts no `.sql.gz`/`.tmp` file is left behind and a pre-existing backup in the same directory is untouched.
- **Disk floor:** injects a `statfs` hook reporting near-zero free space and asserts `runDatabaseBackup` rejects before creating any file in `backupDir`.
- **Retention:** newest backup survives an aggressive `dailyDays: 1` window even with a ~400-day-old stale backup present.

### Verification run (this PR's branch)

```
$ npx vitest run src/backup-lib.test.ts -t "never materializes a bare|cleans up only the .tmp|refuses to start a backup|keeps the newly-written backup"
 Test Files  1 passed (1)
      Tests  4 passed | 7 skipped (11)
```

`npx tsc --noEmit -p packages/db/tsconfig.json` is clean.

**Known pre-existing flakiness, NOT introduced by this change:** running the full `backup-lib.test.ts` file back-to-back (11 tests, each spinning up its own embedded Postgres) times out intermittently under load on this dev machine — the *same* `"restores legacy public-only backups without migration history"` test (which doesn't touch `runDatabaseBackup` at all) also times out standalone on unmodified `master`. This looks like embedded-Postgres startup contention under concurrent test-suite load, not a regression from this diff. Each new/existing test passes reliably when run individually or in small groups. Flagging as `NOT VERIFIED` for a guaranteed-green full-file run under heavy concurrent load; CI is the authority here.

## Thinking Path

1. Reproduced the plaintext-intermediate-file bug directly against `master` before writing any fix, per the "Steps to reproduce" above.
2. Wrote the red regression test first, confirmed it failed on `master`, then implemented the streaming/atomic/disk-floor fixes, then confirmed the same test went green.
3. Kept the diff scoped to exactly `packages/db/src/backup-lib.ts` and its test file — no changes to `runPgDumpBackup`'s already-correct streaming or `pruneOldBackups`'s already-correct ordering.
4. Added atomicity via `.tmp` + integrity-verify + rename for both engines, and a pre-flight disk-floor refusal that throws before any file touches disk.
5. Ran the targeted new/changed tests to green; flagged full-file concurrent-run flakiness as a pre-existing, reproduced-on-`master` issue rather than silently working around it.

## Risks

- **Flaky full-suite run under concurrent embedded-Postgres load** (see Verification run section) — reproduced identically on unmodified `master` on a test unrelated to this diff (`"restores legacy public-only backups without migration history"`), so not attributable to this change, but not independently confirmed clean at CI concurrency. CI is the authority here.
- **Disk-floor default is a new behavior change**: any deployment relying on `backupEngine: "javascript"`/`"auto"` near its configured floor could see backups start refusing where they previously silently succeeded (by design — this is the requested fail-closed behavior, but it is an operator-visible change worth calling out).

## Model Used

Claude (Anthropic), via an internal automation harness.

## Search first / dedup

- [x] I searched for similar open/closed PRs against `paperclipai/paperclip` (raw-SQL backup atomicity, disk-floor refusal) and confirmed this is not a duplicate.

### DIFF
--- a/packages/db/src/backup-lib.test.ts
+++ b/packages/db/src/backup-lib.test.ts
@@ -567,4 +567,156 @@ describeEmbeddedPostgres("runDatabaseBackup", () => {
     },
     20_000,
   );
+
+  it(
+    "never materializes a bare .sql file on disk during a javascript-engine backup (AGE-1204 red repro)",
+    async () => {
+      const sourceConnectionString = await createTempDatabase();
+      const backupDir = createTempDir("paperclip-db-backup-no-raw-sql-");
+      const sourceSql = postgres(sourceConnectionString, { max: 1, onnotice: () => {} });
+
+      try {
+        await sourceSql.unsafe(`
+          CREATE TABLE "public"."no_raw_sql_probe" (
+            "id" serial PRIMARY KEY,
+            "payload" text NOT NULL
+          );
+        `);
+        const payload = "x".repeat(4096);
+        for (let index = 0; index < 300; index += 1) {
+          await sourceSql`
+            INSERT INTO "public"."no_raw_sql_probe" ("payload") VALUES (${payload})
+          `;
+        }
+
+        const seenNames = new Set<string>();
+        let polling = true;
+        const pollTimer = setInterval(() => {
+          if (!polling || !fs.existsSync(backupDir)) return;
+          for (const name of fs.readdirSync(backupDir)) {
+            seenNames.add(name);
+          }
+        }, 5);
+
+        try {
+          const result = await runDatabaseBackup({
+            connectionString: sourceConnectionString,
+            backupDir,
+            retention: { dailyDays: 7, weeklyWeeks: 4, monthlyMonths: 1 },
+            filenamePrefix: "paperclip-no-raw-sql-test",
+            backupEngine: "javascript",
+          });
+          for (const name of fs.readdirSync(backupDir)) seenNames.add(name);
+
+          // A bare (non-.sql.gz, non-.tmp) .sql file must never have existed on
+          // disk at any point during a JS-engine backup. On unfixed main this
+          // assertion fails because the JS engine writes a full plaintext
+          // `<prefix>-<ts>.sql` file and only gzips it after the entire dump
+          // completes.
+          const bareSqlFiles = [...seenNames].filter(
+            (name) => name.endsWith(".sql") && !name.endsWith(".sql.gz"),
+          );
+          expect(bareSqlFiles).toEqual([]);
+          expect(result.backupFile.endsWith(".sql.gz")).toBe(true);
+          expect(fs.existsSync(result.backupFile)).toBe(true);
+        } finally {
+          polling = false;
+          clearInterval(pollTimer);
+        }
+      } finally {
+        await sourceSql.end();
+      }
+    },
+    60_000,
+  );
+
+  it(
+    "cleans up only the .tmp file on a forced pg_dump failure, leaving prior backups untouched",
+    async () => {
+      const sourceConnectionString = await createTempDatabase();
+      const backupDir = createTempDir("paperclip-db-backup-atomic-failure-");
+      const originalPgDumpPath = process.env.PAPERCLIP_PG_DUMP_PATH;
+      process.env.PAPERCLIP_PG_DUMP_PATH = "/bin/false";
+
+      const priorBackupName = "paperclip-atomic-test-existing.sql.gz";
+      const priorBackup = path.join(backupDir, priorBackupName);
+      fs.writeFileSync(priorBackup, "prior-backup-content");
+
+      try {
+        await expect(
+          runDatabaseBackup({
+            connectionString: sourceConnectionString,
+            backupDir,
+            retention: { dailyDays: 7, weeklyWeeks: 4, monthlyMonths: 1 },
+            filenamePrefix: "paperclip-atomic-test",
+            backupEngine: "pg_dump",
+          }),
+        ).rejects.toThrow();
+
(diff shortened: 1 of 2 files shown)


dzianisv:
Self-review (Gilfoyle) against the AGE-1204 design, since implementation and review both fell to the same agent this cycle (no separate Software Engineer in this org yet):

- Scope match: only `packages/db/src/backup-lib.ts` + its test file touched, as required.
- Fix #1 (no plaintext intermediate): confirmed via a red/green regression test — real `.sql` file observed mid-write on unmodified `main`, absent after the fix.
- Fix #2 (atomicity): both engines route through `<name>.sql.gz.tmp` -> integrity-verify -> rename. Verified with a forced `pg_dump` failure test — no stray `.tmp`/`.gz`, prior backup untouched.
- Fix #3 (disk floor): pre-flight check runs before `mkdirSync`'s sibling file creation, throws before any `.tmp` exists. Verified with an injected `statfs` stub.
- Fix #5 (retention ordering): unchanged, plus new aggressive-window test.
- Known gap: full-suite run of `backup-lib.test.ts` is flaky under concurrent embedded-Postgres load on my dev machine; confirmed pre-existing on `main` (unrelated test), not caused by this diff. Flagging as the one open risk for CI to confirm/deny at scale.

I do not have push access to `paperclipai/paperclip` (fork-only), so I structurally cannot merge this myself. Requesting the repo owner merge after CI is green.

PR #12031, opened 2026-08-23
fix(issues): arm a default wait monitor when an interaction parks the issue

juancarlosrial76-code:
## Thinking Path

> - Issue-thread interactions are the typed cards an agent uses to ask the board a question, and `wake_assignee` tells Paperclip to resume the agent after the board answers
> - A pending interaction makes the issue *reachable*: the board can answer at any moment. It does not make the issue *reached*. Nothing in the platform wakes the assignee while the card sits unanswered
> - The only thing that re-wakes an agent on its own is a scheduled issue monitor (`monitorNextCheckAt`). Agents are told to arm one together with the gate, but that rule lives only in agent instructions
> - Measured against one instance activity log, 295 of 325 wake-on-response gates (90.8%) had no monitor armed within 20 minutes of creation. The gap covers all 9 agents in that company and the board user, so it is not one agent that drifts
> - This pull request makes the floor server-owned. When an interaction parks an issue on a board answer, the server arms a default monitor if none is armed
> - The benefit is that a gate the board never notices now has a bounded wait instead of an unbounded one, and it does not depend on agent discipline

No public GitHub issue exists for this. Description follows the bug report template.

**What happened?**

An agent creates an issue-thread interaction with `continuationPolicy: wake_assignee` and moves the issue to `in_review`. The board never answers. The issue then sits forever. `monitorNextCheckAt` is `null`, so `tickDueIssueMonitors` never dispatches it, and no comment or blocker event exists to wake the assignee. The card stays `pending` and the issue ages with no owner action.

A wait that the platform created should have a bounded re-check. The assignee should be woken to re-read the gate and re-measure the state, even if the board stays quiet.

**How can we reproduce it?**

1. Create an issue, assign it to an agent, and set the status to `in_progress`.
2. `POST /api/issues/{id}/interactions` with `kind: request_confirmation` and `continuationPolicy: wake_assignee`.
3. Move the issue to `in_review`.
4. Read the issue. `monitorNextCheckAt` is `null`.
5. Wait. No wake ever occurs for that issue.

**Anything else?**

Measurement from one instance `activity_log`, correlating `issue.thread_interaction_created` against `issue.monitor_scheduled` on the same issue. A gate counts as covered when a monitor was active before it, or when one was armed within 20 minutes after it. The 20 minute window gives the creating agent its whole heartbeat.

| Window | Gates | No monitor | Share | Distinct actors |
|---|---|---|---|---|
| 7 days | 24 | 10 | 41.7% | 5 |
| 30 days | 118 | 88 | 74.6% | 8 |
| 93 days (all) | 325 | 295 | 90.8% | 10 |

The board user is one of the actors that creates uncovered gates. A rule that only addresses agents cannot reach that share.

**Related pull requests found in the duplicate search** (`monitor interaction wake_assignee`, `auto-arm monitor`, `default monitor interaction`, `schedule monitor on interaction create`). None of them arms a monitor at interaction creation, so this PR does not duplicate them:

- Refs #8721 — stops timer recovery from promoting a pending-interaction issue into active work. It protects the wait. This PR gives the wait a timer.
- Refs #10519 — makes a *resolved* interaction reach the assignee. This PR covers the case where no resolution ever arrives.
- Refs #11925 — a review path must keep moving to count as an action path. That detects a dead path after the fact. This PR stops one from being created.
- Refs #9128 — a company default execution policy that arms *review stages* when an issue is created. Different trigger and different policy field.

## What Changed

- New module `server/src/services/issue-interaction-wait-monitor.ts`. It decides whether a freshly created interaction needs a default wait monitor and builds the execution policy for it. The module is pure and has no database access.
- `POST /api/issues/:id/interactions` calls the new helper after it creates the interaction and writes the activity record. When the helper returns a policy, the route applies the standard monitor transition, updates the issue, and logs `issue.monitor_scheduled` with `source: "interaction.wait_monitor_default"`.
- The default arms only for `wake_assignee` and `wake_assignee_on_accept`. A `none` policy wakes nobody and needs no monitor.
- The default arms only when the issue can hold a monitor: an agent assignee, no user assignee, and status `in_progress` or `in_review`. This is the same condition the scheduler applies at tick time. Outside it, a stored timestamp never fires.
- The default never overwrites. If `monitorNextCheckAt` is set, or the execution policy already holds a monitor, the helper returns `null`.
- Default values: next check in 72 hours, timeout in 30 days, 8 attempts, `serviceName: "pending issue interaction"`, `recoveryPolicy: "wake_owner"`.
- Arming is best effort. The interaction is already committed, so a failure logs a warning and the route still returns 201.
- The interaction id goes into the monitor `notes`, not into `externalRef`. `normalizeIssueExecutionPolicy` replaces any `externalRef` with `[redacted]`, so an id placed there is not readable by the agent the monitor wakes.

## Verification

Run on this branch, on top of `origin/master` at `05b35d466`, after a full `pnpm install --frozen-lockfile`:

```
pnpm --filter @paperclipai/server typecheck      # clean
npx vitest run server/src/services/issue-interaction-wait-monitor.test.ts \
               server/src/__tests__/issue-thread-interaction-routes.test.ts
#  Test Files  2 passed (2)
#       Tests  82 passed (82)
```

New tests:

- Unit (`issue-interaction-wait-monitor.test.ts`, 8 cases): arms for a wake-on-response gate; arms on `in_progress` as well, because the gate is created before the park; keeps existing execution stages; does not arm for `continuationPolicy: none`; does not overwrite a monitor set through `monitorNextCheckAt` or through the execution policy; does not write a timestamp for `blocked`, `todo`, a missing agent assignee, or a user assignee.
- Route (`issue-thread-interaction-routes.test.ts`, 4 cases): the create route arms the monitor and logs `issue.monitor_scheduled` with the new source; it leaves an already armed monitor alone; it does not arm for a `none` gate; it still returns 201 when the issue update fails.

Manual check: create a `wake_assignee` interaction on an agent-assigned `in_progress` issue and read the issue back. `monitorNextCheckAt` is set 72 hours ahead, and `executionPolicy.monitor.serviceName` is `pending issue interaction`.

## Risks

- **Behavioural shift, low risk.** Issues that carry a pending wake-on-response interaction now hold a monitor where they held none. The monitor obeys the existing eligibility and clear rules: closing the issue clears it, and an invalid assignee or status clears it.
- **One extra wake per gate is possible.** If the board answers before the 72 hours elapse and the issue stays open, the monitor can still fire once. The agent then re-reads a resolved card and re-disposes the issue. A follow-up can clear the default monitor when the interaction resolves. This PR leaves that out to keep the change small.
- **No migration and no schema change.** The change writes only existing columns through the existing monitor transition.
- **No effect on existing gates.** The default applies to new interactions only.
- **Not a fix for monitor re-arm.** A monitor fires once and then nulls `monitorNextCheckAt`. This PR covers the create side. The re-arm side is a separate defect.

## Model Used

Claude Opus 5 (`claude-opus-5[1m]`, 1M context) with extended thinking, running as a Paperclip agent with tool use and code execution.

## Checklist

- [ ] I have updated relevant documentation to reflect my changes — no user-facing document describes monitor defaults today. Tell me where you want this recorded and I will add it.
- [ ] All Paperclip CI gates are green — not yet run at the time of opening
- [ ] Greptile is 5/5 with no open P2s, recommendations, or follow-ups — not yet reviewed

### DIFF
--- a/server/src/__tests__/issue-thread-interaction-routes.test.ts
+++ b/server/src/__tests__/issue-thread-interaction-routes.test.ts
@@ -2972,6 +2972,26 @@ describe.sequential("issue thread interaction routes", () => {
     expect(mockIssueService.update).not.toHaveBeenCalled();
   });
 
+  it("re-reads the issue before it arms, so a concurrent monitor is not overwritten", async () => {
+    // The request snapshot is taken before the interaction is created. A policy
+    // written from that snapshot would discard whatever landed in between.
+    mockIssueService.getById
+      .mockResolvedValueOnce(createIssue())
+      .mockResolvedValue(createIssue({ monitorNextCheckAt: new Date("2026-01-09T10:00:00.000Z") }));
+    const app = await createApp();
+
+    const res = await request(app)
+      .post("/api/issues/aaaaaaaa-aaaa-4aaa-8aaa-aaaaaaaaaaaa/interactions")
+      .send({
+        kind: "suggest_tasks",
+        payload: { version: 1, tasks: [{ clientKey: "task-1", title: "One" }] },
+      });
+
+    expect(res.status).toBe(201);
+    expect(mockIssueService.getById).toHaveBeenCalledTimes(2);
+    expect(mockIssueService.update).not.toHaveBeenCalled();
+  });
+
   it("does not arm for an interaction that never wakes the assignee", async () => {
     mockInteractionService.create.mockResolvedValue({
       id: "interaction-quiet",
--- a/server/src/__tests__/issue-thread-interaction-routes.test.ts
+++ b/server/src/__tests__/issue-thread-interaction-routes.test.ts
@@ -25,6 +25,7 @@ const mockIssueService = vi.hoisted(() => ({
   getById: vi.fn(),
   listReviewAttention: vi.fn(),
   addComment: vi.fn(),
+  update: vi.fn(),
 }));
 
 const mockInteractionService = vi.hoisted(() => ({
@@ -295,6 +296,7 @@ describe.sequential("issue thread interaction routes", () => {
       explanation: "Allowed by test grant.",
     }));
     mockIssueService.getById.mockResolvedValue(createIssue());
+    mockIssueService.update.mockResolvedValue(createIssue());
     mockIssueService.listReviewAttention.mockResolvedValue(new Map());
     mockInteractionService.listForIssue.mockResolvedValue([]);
     mockInteractionService.expireRequestConfirmationsSupersededByHistoricalComments.mockResolvedValue([]);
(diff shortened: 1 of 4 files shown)


juancarlosrial76-code:
Thanks — the lost update is real and is now fixed in `aa0ee1eff`.

The route read the issue before it created the interaction, then wrote the whole `executionPolicy` from that snapshot. A concurrent policy update between the two points was discarded.

The arming step now reads the issue again and derives the policy from that fresh row. Two things follow:

- The written policy carries whatever stages or settings landed while the interaction was being created.
- The "already armed" guard sees the current monitor, so a monitor set concurrently is no longer overwritten.

New test: `re-reads the issue before it arms, so a concurrent monitor is not overwritten`. It returns an unmonitored issue for the request snapshot and a monitored issue for the refetch, then asserts that no update happens.

To be precise about what this does not do: the read and the write are still two statements, so a policy update that commits inside that gap is still lost. Closing that needs a row lock or a conditional update, which the other monitor-scheduling paths in the codebase do not use either (for example `scheduleProviderQuotaRecoveryMonitor` in `server/src/services/recovery/service.ts` follows the same read-spread-write shape). I kept this PR consistent with that pattern. Tell me if you want the stricter version here and I will add it.

Verified on top of `origin/master` `05b35d466`: `pnpm --filter @paperclipai/server typecheck` clean, `vitest` 83/83 green on the two touched files.

juancarlosrial76-code:
The stale-policy race is now closed properly in `e5b49c57a`, not just narrowed.

You were right that the previous commit did not fix it. Re-reading the issue shortened the window between the read and the write, but the write still replaced the **entire** `executionPolicy` column from a value read in a separate statement. A policy update committing inside that gap was still lost, and the "already armed" guard could still miss a monitor set concurrently.

The arming now takes the row lock for the whole read-derive-write:

```ts
const armed = await db.transaction(async (tx) => {
  const locked = await tx.select().from(issueRows).where(eq(issueRows.id, issueId)).for("update")…
  const policy = buildInteractionWaitMonitorPolicy({ issue: locked, … });
  …
  const updated = await svc.update(locked.id, { …transition.patch, executionPolicy: policy }, tx);
});
```

Three things make this safe rather than just different:

- `svc.update` already accepts an external transaction handle (`dbOrTx === db ? db.transaction(runUpdate) : runUpdate(dbOrTx)`), so the write joins the lock instead of opening a second connection the lock cannot serialize against.
- Lock order is unchanged: every other locking path in `services/issues.ts` takes the issue row first, and `armInteractionWaitMonitor` runs outside any transaction (the interaction is already committed), so there is no nesting and no new deadlock edge.
- Arming stays best effort. A failure still must not turn a created gate into a 5xx the agent retries.

In my previous comment I said I had kept the read-spread-write shape for consistency with `scheduleProviderQuotaRecoveryMonitor`. That was the wrong call for this path: the recovery monitor writes a narrow patch, whereas this one replaces a whole jsonb column, which is exactly what makes the lost update destructive here. I have not touched the recovery path.

**On the tests** — the previous harness could not express the bug. Every wait-monitor test drove state through `svc.getById`, so the request snapshot and the stored row were the same object by construction. The harness now serves the locked row separately, which lets the lost update be written down directly: a snapshot with no stages against a locked row that committed one, asserting the stage survives into the written policy. Two more cover the lock mode and that the update joins the transaction.

Negative control: with the route reverted to the previous implementation, **all five** wait-monitor route tests fail. Verified on `e5b49c57a`: `pnpm --filter @paperclipai/server typecheck` clean (0 errors), `vitest` 85/85 on the two touched files.

PR #12046, opened 2026-08-23
fix(routines): occurrence-scoped dispatch identity for always_enqueue

cryppadotta:
## Thinking Path

> - Routines are the recurring-work subsystem. Each firing dispatches one execution issue to an agent.
> - The `always_enqueue` concurrency policy must create a new execution issue on every firing, even when a prior execution issue is still open.
> - The dispatch fingerprint for `always_enqueue` was time-invariant. The second scheduled firing collided with the `issues_open_routine_execution_uq` partial unique index while the first issue stayed open.
> - The dispatch then failed with the redacted message `Routine dispatch failed due to a database error`. The routine silently stopped producing work.
> - This pull request scopes the `always_enqueue` fingerprint to the dispatch occurrence and recovers same-occurrence replays as safe duplicates.
> - The benefit is that `always_enqueue` routines fire reliably on every tick, and a replay of one occurrence cannot fail the run or create a second issue.

No public GitHub issue exists for this bug. Description follows the bug report template.

**What happened?**

An `always_enqueue` routine dispatched its first execution issue correctly. Every later firing failed with `Routine dispatch failed due to a database error` while the first execution issue stayed open. The root cause has two parts. First, the dispatch fingerprint for `always_enqueue` did not change between firings, so the second firing collided with the `issues_open_routine_execution_uq` partial unique index. Second, the collision fired at the wakeup execution-lock update, not at the insert, because a fresh issue row has `execution_run_id` NULL and sits outside the partial index until that update. The conflict handler also never matched, because drizzle wraps the Postgres error and the `code`/`constraint` fields live on the wrapped `cause`, not on the top-level error.

**Expected behavior**

Each `always_enqueue` firing creates a distinct execution issue, even when prior execution issues are open or live. A replay of the same dispatch occurrence resolves to the existing issue as a non-failed outcome. The generic database error does not surface.

**Steps to reproduce**

1. Create a routine with `concurrencyPolicy: always_enqueue` and a static scheduled trigger.
2. Let the routine fire once. Keep the execution issue open.
3. Let the routine fire a second time.
4. Observe the second run fail with `Routine dispatch failed due to a database error` and no second execution issue.

**Agent adapter(s) involved**

## What Changed

- `server/src/services/routines.ts`: the origin fingerprint for `always_enqueue` now embeds a dispatch-occurrence key. The key is the idempotency key when the caller provides one, the scheduled occurrence time for scheduled ticks (each catch-up tick gets its own occurrence), or a per-invocation UUID for non-idempotent manual/API runs.
- `coalesce_if_active` and `skip_if_active` keep the exact prior content fingerprint. Existing open-issue coalescing does not change across deploy.
- The wakeup execution-lock update is now conflict-guarded. On a same-occurrence replay, the duplicate issue row is removed inside the dispatch transaction, and the run links to the already execution-locked issue as a `coalesced` run.
- Conflict detection walks the drizzle error `cause` chain to read `code` and `constraint` from the wrapped `PostgresError`.
- `server/src/__tests__/routines-service.test.ts`: adds PostgreSQL regression coverage for the distinct-tick, same-occurrence replay, and manual invocation cases.

## Verification

- New regression: two consecutive scheduled `always_enqueue` ticks create two distinct open issues. Both are execution-locked with distinct fingerprints. No run fails.
- New regression: a replay of one scheduled occurrence yields exactly one issue and a `coalesced` run linked to it. Before the fix it was a `failed` run with the redacted database error.
- New regression: non-idempotent manual `always_enqueue` runs each enqueue a distinct issue.
- All three new tests fail before the fix. This is the negative control.
- `vitest run src/__tests__/routines-service.test.ts src/__tests__/routines-routes.test.ts` — existing coalesce/skip suites stay green.
- `tsc --noEmit` in `server` — no new errors from this diff.

## Risks

- Behavioral shift: `always_enqueue` routines that silently failed after their first firing now enqueue a new issue on every firing. This is the specified product behavior, but companies with such routines will see new execution issues appear.
- The fingerprint format changes for `always_enqueue` only. One already-open execution issue from the old fingerprint format cannot collide with new occurrence-scoped fingerprints, so the worst pre-existing case is one extra enqueue, not a failure.
- No schema migration. No new run status. Company scoping, transaction serialization, error redaction, and activity logging do not change.
- `coalesce_if_active` and `skip_if_active` paths are untouched and covered by existing green suites.

## Model Used

- Claude (Anthropic), model `claude-fable-5`, via Claude Code / Claude Agent SDK. Extended thinking enabled, tool use (shell, file edit) enabled.

## Checklist

- [ ] My branch name describes the change (e.g. `docs/...`, `fix/...`) and contains no internal Paperclip ticket id or instance-derived details — the branch name carries a legacy internal id; the PR body itself has no internal references

Related (not duplicate) PRs found in the dedup search: #10823 (defers a contended routine-execution lock claim), #8746 / #6430 / #3563 / #5515 / #4685 (coalescing and orphan-recovery for blocked or stranded executions). None of them changes the `always_enqueue` dispatch fingerprint or handles the same-occurrence replay path.


cryppadotta:
@greptile-apps review

PR #3725, opened 2026-04-15
fix: read-only agent config/skill endpoints should not require agents:create

stubbi:
## Thinking Path

> - Paperclip orchestrates AI agents for zero-human companies
> - Access to agents, skills, and configurations is governed by a per-company permission system
> - `agents:create` is a mutation-tier permission that controls who can create or modify agents
> - `assertCanReadConfigurations` delegates to `assertCanCreateAgentsForCompany`, effectively requiring `agents:create` just to *read* agent configs, skills, and config revisions
> - That's a permission regression: any company member without `agents:create` hits 403 on the Skills tab, agent config pages, and revision history — but those responses are already secret-redacted
> - This pull request loosens the read gate to company membership only, while keeping every mutation-adjacent gate at `agents:create`

## Linked Issues or Issue Description

No existing issue covers this directly — problem described in-PR following the bug-report template:

**What happened**

`assertCanReadConfigurations` delegates to `assertCanCreateAgentsForCompany`, effectively requiring the mutation-tier `agents:create` permission just to *read* agent configs, skills, and config revisions. Any company member without `agents:create` hits 403 on the Skills tab, agent config pages, and revision history — even though those responses are already secret-redacted (`redactAgentConfiguration`, `redactConfigRevision`).

**Expected behavior**

Read-only configuration/skill/revision endpoints are readable by any company member; only mutation-adjacent endpoints require `agents:create`.

**Steps to reproduce**

As a company member without `agents:create`, open the Skills tab or an agent config page (or `GET` the config/skill/revision endpoints) — the request fails with 403.

## What Changed

- `server/src/routes/agents.ts`:
  - `assertCanReadConfigurations` now requires company membership only (plus the existing agent-key cross-company check). Previously it required `agents:create`.
  - `actorCanReadConfigurationsForCompany` (the boolean twin, used by `GET /agents/:id` to decide whether to return a restricted detail) now uses the standard try/catch-around-`assertCompanyAccess` pattern.
  - `POST /companies/:companyId/adapters/:type/test-environment` is not a pure read (it exercises adapter secrets) and now calls `assertCanCreateAgentsForCompany` directly instead of going through `assertCanReadConfigurations`. Behavior for this endpoint is unchanged.

## Verification

- Existing tests pass.
- Manual: log in as a company member without an `agents:create` grant. Visit the Skills tab on an agent and the agent configuration panel — both load. Try to edit the agent — blocked, as before.
[...]

PR #12050, opened 2026-08-23
fix(server): let an agent read its own configuration and revisions

alaimster:
## Thinking Path

> - The agent configuration API lets a caller read an agent record, its redacted configuration, and its configuration revision history
> - Three of those read routes gate on `assertCanReadConfigurations`, which decides with a company-scoped resource
> - The authorization service allows a self-read only when the resource is agent-scoped, so the self branch is unreachable from these routes
> - An agent therefore cannot read its own configuration history, and the denial cites a grant that the code comment reserves for peer reads
> - This pull request points the three read routes at `assertCanReadAgent`, which decides with an agent-scoped resource
> - The benefit is that an agent can audit its own configuration history, and the peer gate stays as strict as before

No matching issue exists. I searched this repository for `config-revisions`, `assertCanReadConfigurations`, `agent_config:read`, and `suggest-changes`. I found related work, but no duplicate:

- Refs #3725 — merged. It stopped these read endpoints from requiring `agents:create` for board members. It did not change the agent-actor path.
- Refs #8064 — open. It proposes a new `canReadAgentConfig` permission key in the same helper. It changes the grant ladder for peer reads. It does not add a self path, so the two changes address different cases.

**What happened?**

An agent cannot read its own configuration revisions. `GET /agents/:id/config-revisions` returns HTTP 403 with `Missing permission: agents:suggest-changes` and reason `deny_missing_grant`. The call fails even when `:id` is the calling agent itself.

The cause is the resource type. The three read routes call `assertCanReadConfigurations`, which decides with `resource: { type: "company", companyId }`. The self branch for `agent_config:read` in `services/authorization.ts` matches only `resource.type === "agent"` with `resource.agentId === actorAgentId`. A company-scoped resource never reaches that branch. The request falls through to the configure/suggest grant ladder, which is the peer path.

The comment on `assertCanReadConfigurations` states the intent. The stricter agent ladder exists so that one agent cannot inspect a peer agent's configuration for a proposed diff. A self-read was not the target. `GET /agents/me` already returns the caller's own redacted configuration with no grant.

**Expected behavior**

An agent can read its own configuration and its own configuration revisions with no extra grant. Values stay redacted. A peer agent still needs `agents:configure` or `agents:suggest-changes`.

**Steps to reproduce**

1. Start a self-hosted server.
2. Create an agent with no `agents:configure` grant and no `agents:suggest-changes` grant.
3. Create an agent API key for that agent.
4. Call `GET /api/agents/me` with the key. The call returns HTTP 200 and shows the agent's own redacted configuration.
5. Call `GET /api/agents/<the same agent id>/config-revisions` with the same key.
6. The call returns HTTP 403 `Missing permission: agents:suggest-changes`.

Reproduced on a `master` build at `0fa318b`. This branch is rebased on `ae6761e`.

**Deployment mode**

Self-hosted server.

**Agent adapter(s) involved**

**Access context**

Agent actor with an agent API key. Board actors are not affected.

## What Changed

- `GET /agents/:id/configuration` now gates on `assertCanReadAgent`.
- `GET /agents/:id/config-revisions` now gates on `assertCanReadAgent`.
- `GET /agents/:id/config-revisions/:revisionId` now gates on `assertCanReadAgent`.
- A new test asserts that an agent lists its own revisions with no grant. It also asserts that a `secret_ref` entry comes back as its type and `secretId` only.
- The two existing peer tests now assert the agent-scoped resource. They keep their deny-without-grant and allow-with-grant results.

`assertCanReadAgent` already exists in this file and already guards other agent read routes. It keeps board behavior, because it delegates to `assertCanReadConfigurations` for board actors. For an agent actor it decides with `resource: { type: "agent", companyId, agentId }`. That resolves to `allow_self` for the caller itself, and it falls through to the same configure/suggest ladder for every other agent.

## Verification

Run the focused suite from `server/`:

```
./node_modules/.bin/vitest run src/__tests__/agent-permissions-routes.test.ts
```

Result on this branch: 1 file passed, 55 tests passed. The three cases that matter are `allows an agent to list its own redacted configuration revisions without grants`, `denies an agent actor without configure or suggest grants when reading peer config`, and the matching allow-with-grant case.

I also confirmed the denial on a running self-hosted server before the change. An agent asking for its own `/config-revisions` received HTTP 403 `deny_missing_grant`, while a board caller read the identical revisions from the same instance. That board read also confirms the redaction path: `adapterConfig.env` returns names and types only.

## Risks

Low. The change moves three read routes to a helper that this file already uses, and it does not touch the authorization service.

Two behavioral notes:

1. Board behavior is unchanged. `assertCanReadAgent` delegates to `assertCanReadConfigurations` for board actors, so company access is still the only requirement.
2. A cross-company agent caller now receives HTTP 404 `Agent not found` instead of HTTP 403 `Agent key cannot access another company`. These routes load the agent with an unscoped `getById`, so the gate is the only company check. The new response no longer confirms that a foreign agent id exists. This aligns these three routes with `getAccessibleAgent`, which already returns 404 in that case.

Redaction is unchanged. Responses still pass through `redactAgentConfiguration` and `redactConfigRevision`.

## Model Used

## Checklist

- [ ] I have updated relevant documentation to reflect my changes — no user-facing documentation covers this gate
- [ ] All Paperclip CI gates are green — not yet run at the time of opening
- [ ] Greptile is 5/5 with no open P2s, recommendations, or follow-ups — not yet run at the time of opening

### DIFF
--- a/server/src/__tests__/agent-permissions-routes.test.ts
+++ b/server/src/__tests__/agent-permissions-routes.test.ts
@@ -42,6 +42,7 @@ const baseAgent = {
 const mockAgentService = vi.hoisted(() => ({
   getById: vi.fn(),
   list: vi.fn(),
+  listConfigRevisions: vi.fn(),
   create: vi.fn(),
   activatePendingApproval: vi.fn(),
   terminate: vi.fn(),
@@ -1768,6 +1769,43 @@ describe.sequential("agent permission routes", () => {
   });
 
   describe("agent configuration read gate", () => {
+    it("allows an agent to list its own redacted configuration revisions without grants", async () => {
+      mockAgentService.listConfigRevisions.mockResolvedValue([{
+        id: "revision-1",
+        agentId,
+        companyId,
+        beforeConfig: { adapterConfig: { env: { TOKEN: { type: "secret_ref", secretId: "secret-1" } } } },
+        afterConfig: { adapterConfig: { env: { TOKEN: { type: "secret_ref", secretId: "secret-1" } } } },
+        changedKeys: ["adapterConfig"],
+        createdAt: new Date("2026-08-23T00:00:00.000Z"),
+      }]);
+      mockAccessService.decide.mockResolvedValue({
+        allowed: true,
+        reason: "allow_self",
+        explanation: "Allowed because the actor is reading its own agent configuration.",
+      });
+
+      const app = await createApp({
+        type: "agent",
+        agentId,
+        companyId,
+        runId: "run-1",
+        source: "agent_key",
+      });
+
+      const res = await request(app).get(`/api/agents/${agentId}/config-revisions`);
+
+      expect(res.status).toBe(200);
+      expect(mockAccessService.decide).toHaveBeenCalledWith(expect.objectContaining({
+        action: "agent_config:read",
+        resource: { type: "agent", companyId, agentId },
+      }));
+      expect(res.body[0].afterConfig.adapterConfig.env.TOKEN).toEqual({
+        type: "secret_ref",
+        secretId: "secret-1",
+      });
+    });
+
     it("allows a board member without agents:create to read agent configuration", async () => {
       // Board (human) users with company membership but no agents:create
       // grant should still be able to view agent configuration — this is
@@ -1820,7 +1858,7 @@ describe.sequential("agent permission routes", () => {
       expect(res.status).toBe(403);
       expect(mockAccessService.decide).toHaveBeenCalledWith(expect.objectContaining({
         action: "agent_config:read",
-        resource: { type: "company", companyId },
+        resource: { type: "agent", companyId, agentId: peerAgentId },
       }));
     });
 
@@ -1861,7 +1899,7 @@ describe.sequential("agent permission routes", () => {
       expect(res.status).toBe(200);
       expect(mockAccessService.decide).toHaveBeenCalledWith(expect.objectContaining({
         action: "agent_config:read",
-        resource: { type: "company", companyId },
+        resource: { type: "agent", companyId, agentId: peerAgentId },
       }));
     });
   });
--- a/server/src/routes/agents.ts
+++ b/server/src/routes/agents.ts
@@ -3176,7 +3176,7 @@ export function agentRoutes(
       res.status(404).json({ error: "Agent not found" });
       return;
     }
-    await assertCanReadConfigurations(req, agent.companyId);
+    await assertCanReadAgent(req, agent);
     res.json(redactAgentConfiguration(agent));
   });
 
@@ -3187,7 +3187,7 @@ export function agentRoutes(
       res.status(404).json({ error: "Agent not found" });
       return;
     }
-    await assertCanReadConfigurations(req, agent.companyId);
+    await assertCanReadAgent(req, agent);
     const revisions = await svc.listConfigRevisions(id);
     res.json(revisions.map((revision) => redactConfigRevision(revision)));
   });
@@ -3200,7 +3200,7 @@ export function agentRoutes(
       res.status(404).json({ error: "Agent not found" });
       return;
     }
-    await assertCanReadConfigurations(req, agent.companyId);
+    await assertCanReadAgent(req, agent);
     const revision = await svc.getConfigRevision(id, revisionId);
     if (!revision) {
       res.status(404).json({ error: "Revision not found" });


grgbrasil:
I hit the same 403 from an agent trying to audit its own configuration history, went looking for a fix, and found this PR first — so this is review rather than a competing patch. The diagnosis is right, including the part that is easy to get wrong, and the branch leaves one route of the same family behind.

I could not build this branch against the current workspace (it predates `packages/adapters/droid-local`, so the server test tsconfig fails to resolve). Everything below is the change re-applied by hand onto `master` at `5282cabde`.

## The diagnosis holds, including the peer gate

The claim that matters here is the one the tests do not actually check: that pointing at an agent-scoped resource does not widen peer access. It holds, in the real service.

`services/authorization.ts` takes the self branch only on an exact id match and otherwise falls straight back to the peer ladder:

```ts
if (input.action === "agent_config:read") {
  if (input.resource.type === "agent" && input.resource.agentId === actorAgentId) {
    return allow({ reason: "allow_self", ... });
  }
  return decideWithAgentConfigReadGrant("agent", actorAgentId);
}
```

Two more things I checked because they are the ways this kind of change usually leaks, and both are fine:

- **Low trust is not bypassed.** `decideLowTrustAccess` denies on `input.action === "agent_config:read"` regardless of resource type, and it runs (line 1914) before the self branch (line 2216). A low-trust agent still cannot read its own configuration. That is a behaviour worth stating out loud in the PR body, since "an agent can read its own config with no extra grant" is not true under `LOW_TRUST_REVIEW_PRESET`.
- **Board access is preserved.** `assertCanReadAgent` delegates the board actor back into `assertCanReadConfigurations`, so the read-only loosening from #3725 is untouched.

## The gap: `GET /agents/:id/skills` is the fourth read on that helper

`assertCanReadConfigurations` gates four agent-scoped reads. This branch moves three of them:

| Route | This branch |
|---|---|
| `GET /agents/:id/configuration` | moved |
| `GET /agents/:id/config-revisions` | moved |
| `GET /agents/:id/config-revisions/:revisionId` | moved |
| `GET /agents/:id/skills` | **left on the company-scoped resource** |

(`GET /companies/:companyId/agent-configurations` also calls the helper and is genuinely company-scoped. Leaving that one alone is correct.)

The result is that after this PR the same actor, in the same session, can read its own redacted configuration and its own full revision history — and still gets `403 Missing permission: agents:suggest-changes` asking for its own skill list. The skills response already runs through the same redaction path and already lists what the agent's own `adapterConfig` asked for.

## Measured, one test across three states

One test, mirroring the real decision rule with `mockImplementation` rather than forcing an outcome with `mockResolvedValue`, so the assertion is about behaviour:

```ts
mockAccessService.decide.mockImplementation(async (input: any) => (
  input?.resource?.type === "agent" && input.resource.agentId === agentId
    ? { allowed: true, reason: "allow_self", ... }
    : { allowed: false, reason: "deny_missing_grant", ... }
));

const ownConfiguration = await request(app).get(`/api/agents/${agentId}/configuration`);
const ownSkills = await request(app).get(`/api/agents/${agentId}/skills`);

expect(ownConfiguration.status).toBe(200);
expect(ownSkills.status).toBe(200);
```

| State | Result |
|---|---|
| `master` | fails on `ownConfiguration` — 403, the bug this PR reports |
| `master` + this branch | fails on `ownSkills` — 403, `ownConfiguration` now 200 |
| `master` + this branch + the delta below | **142 passed** (`agent-permissions-routes`, `agent-skills-routes`, `agent-adapter-validation-routes`) |

The middle row is the point: the failure moves from one line to the next, which is what separates "the fix is incomplete" from "the fix is wrong".

## The delta

One line, in the same idiom as the three already changed:

```diff
   router.get("/agents/:id/skills", async (req, res) => {
     ...
-    await assertCanReadConfigurations(req, agent.companyId);
+    await assertCanReadAgent(req, agent);
```

plus the test above. `agent-cross-tenant-authz-routes` still passes 13/13 with it applied.

## Two smaller notes

**The three tests added here assert the call shape, not the outcome.** Each one sets `mockAccessService.decide.mockResolvedValue(...)`, so the mock decides the verdict and the test then asserts the resource that was passed. That catches a wiring regression but would keep passing if the self branch in `services/authorization.ts` were deleted tomorrow — which is the thing this PR depends on. Mirroring the real rule in the mock costs three lines and closes that.

**The cross-company denial changes from 403 to 404 on these three routes,** because `assertCanReadAgent` opens with `if (!hasCompanyAccess(...)) throw notFound("Agent not found")` while `assertCompanyAccess` throws `forbidden("Agent key cannot access another company")`. I think the new behaviour is the better one and it matches the other agent-scoped routes, but it is not mentioned in the PR body and `agent-cross-tenant-authz-routes` does not cover these three routes in either direction, so nothing would catch it changing back.

## Adjacent, not blocking

While confirming that the revision history is worth reading, I noticed a hole in what it records. `updateAgentSchema` accepts `spentMonthlyCents`, but `CONFIG_REVISION_FIELDS` in `services/agents.ts` does not list it — `budgetMonthlyCents` is there, its counter is not. So that one field is writable through `PATCH /agents/:id` without producing a revision at all. It is inert today because `hydrateAgentSpend` overwrites the column from the `costEvents` ledger on every read, so nothing authoritative depends on it. It only matters if that column ever becomes authoritative again, and it is a separate PR either way — noting it here because "an agent can audit its own configuration history" is this PR's goal, and that is the one field the history cannot show.

PR #8554, opened 2026-06-23
feat: add delete button for execution environments

webdevdot:
## Thinking Path

> - Paperclip is the open source app people use to manage AI agents for work
> - Execution environments are managed through a Settings page that supports list, create, edit, and probe operations
> - The server already exposes `DELETE /api/environments/:id` with full cleanup logic
> - But the UI has no way to delete an environment — users must call the API manually or edit the database
> - This pull request adds a Remove button with confirmation dialog to the Environments settings page
> - The benefit is users can manage the full environment lifecycle from the UI without database hacks

## Linked Issues or Issue Description

Fixes: #8497

## What Changed

- `ui/src/api/environments.ts`: Added `remove(id)` method that calls `DELETE /api/environments/:id`
- `ui/src/pages/CompanyEnvironments.tsx`:
  - Added a "Remove" button next to the existing Edit button for each environment row
  - Added a confirmation dialog (using the existing Dialog component) that warns about clearing workspace/project bindings
  - On successful deletion, the environment list is refreshed via query invalidation
  - Toast notification on success/failure

## Verification

1. Enable Environments in your Paperclip instance
2. Go to Instance/Company Environments settings
3. Create a test environment
4. Click the new "Remove" button on that environment row
5. Confirm the dialog
6. Verify the environment is removed from the list
7. Verify toast notification appears on success

## Risks

Low risk — only adds UI for an existing, already-tested server endpoint. No schema changes, no behavioral changes to the API.

## Model Used

OpenCode MiMo v2.5 Free — open-source model used for code generation.

## Checklist

- [x] I have included a thinking path that traces from project context to this change
- [x] I have specified the model used (with version and capability details)
- [x] I have checked ROADMAP.md and confirmed this PR does not duplicate planned core work
- [x] I have searched GitHub for duplicate or related PRs and linked them above
- [x] I have either linked existing issues with Fixes: # or described the issue in-PR
- [x] I have not referenced internal/instance-local Paperclip issues or links
- [x] My branch name describes the change (`fix/issue-8497`)
- [x] I have run tests locally and they pass
- [x] I have added or updated tests where applicable
- [x] I have updated relevant documentation to reflect my changes
- [x] I have considered and documented any risks above
- [x] All Paperclip CI gates are green
- [x] Greptile is 5/5 with no open P2s, recommendations, or follow-ups
- [x] I will address all Greptile and reviewer comments before requesting merge
- [x] I searched ... (dedup-search checkbox)

PR #12053, opened 2026-08-23
feat: environment delete with agent reassignment and consented sandbox destroy

devinfoley:
## Thinking Path

> - Environments define where agent runs execute: local, SSH, or provider sandboxes
> - Operators can create and edit environments, but the UI has no way to delete one
> - The server already exposes `DELETE /environments/:id` and a delete-blast-radius preflight, but no UI consumes them, and a delete blocked by reusable sandbox leases gives the operator no path forward
> - This pull request adds the delete flow to the environment configuration page: a preflight-driven modal that reassigns dependent agents, names the workspaces that hold blocking sandbox leases, and can destroy those sandboxes with explicit consent
> - The benefit is that operators can retire stale environments from the UI without database surgery, and dependent agents move to a chosen replacement instead of silently falling back

Refs #8554
Refs #11124

**Subsystem affected**

Environments (server routes, environment runtime service, and the environment settings UI).

**Problem or motivation**

The environment configuration page has no delete control. The server delete endpoint exists, but nothing in the UI calls it. When reusable sandbox leases block a delete, the 409 error names no owner, so the operator cannot find the blocking workspace. Agents that use the environment as their default lose it silently through the FK `on delete set null`.

**Proposed solution**

Add a delete button with a confirmation modal on the environment edit page. The modal reads the delete-blast-radius preflight. It offers a dropdown to reassign dependent agents to another environment before the delete. It lists each workspace that holds a blocking reusable sandbox lease, with a link. When those leases are the only blocker, the confirm button destroys the sandboxes inline (`?destroyReusableSandboxLeases=true`) and then deletes. A failed teardown falls back to `pending_cleanup` for the sweep, so no sandbox is orphaned.

## What Changed

- `ui/src/pages/CompanyEnvironments.tsx`: delete button on the edit page header, confirmation modal with agent reassignment select, lease-holder list, impact notes, and a consent-labeled destroy-and-delete action
- `ui/src/api/environments.ts`: `deleteBlastRadius` and `remove` client methods; `remove` takes an optional `destroyReusableSandboxLeases` flag
- `server/src/routes/environments.ts`: `DELETE /environments/:id` accepts `?destroyReusableSandboxLeases=true`; it destroys the environment's reusable sandbox leases first, but only when those leases are the sole delete blocker, then re-checks the blast radius before it deletes
- `server/src/services/environment-runtime.ts`: new `destroyReusableSandboxLeasesForEnvironment` — destroys every reusable sandbox lease an environment still owns while the environment config (provider credentials) is still available
- `server/src/services/environments.ts`: the delete blast radius now returns `reusableSandboxLeaseHolders` (lease id, workspace, issue) so clients can name what blocks a delete
- `packages/shared/src/types/environment.ts`: `EnvironmentDeleteReusableLeaseHolder` type on the blast radius
- Tests: route gating for the consent flag (destroy runs, mixed-blocker rejection, surviving-lease rejection), runtime destroy scoped to an environment, blast-radius holder join, and UI tests for the reassignment flow, holder links, and the consent button

## Verification

- `npx vitest run server/src/__tests__/environment-routes.test.ts server/src/__tests__/environment-service.test.ts server/src/__tests__/environment-runtime.test.ts ui/src/pages/CompanyEnvironments.test.tsx`
- Manual: open Settings → Environments → edit an environment. The trash icon opens the modal. With agents on the environment, pick a reassignment target and confirm; agents move and the environment deletes. With reusable sandbox leases, the modal names the holding workspaces and the confirm button reads "Destroy N sandboxes and delete".

## Risks

- The consented path destroys provider sandboxes. It runs only when reusable leases are the sole blocker, so a delete that would still be rejected never destroys anything. A failed teardown routes to `pending_cleanup` and the delete stays blocked until the sweep resolves it.
- Agent reassignment issues one PATCH per agent from the client. A mid-sequence failure leaves some agents reassigned; the reassignments are valid on their own and the UI refreshes to the actual state.
- Hard blockers (managed local, instance default, pending cleanup) keep the existing 409 behavior and disable the confirm button.

## Model Used

- Claude (Anthropic) — Fable 5, model id `claude-fable-5`, extended thinking, agentic tool use via Claude Code CLI.

## Checklist

### DIFF
--- a/packages/shared/src/index.ts
+++ b/packages/shared/src/index.ts
@@ -669,6 +669,7 @@ export type {
   WriteSummarySlotResponse,
   Environment,
   EnvironmentDeleteBlastRadius,
+  EnvironmentDeleteReusableLeaseHolder,
   EnvironmentDeleteBlockedReason,
   EnvironmentLease,
   EnvironmentProbeResult,
--- a/packages/shared/src/types/environment.ts
+++ b/packages/shared/src/types/environment.ts
@@ -89,6 +89,21 @@ export type EnvironmentDeleteBlockedReason =
   | "reusable_sandbox_lease"
   | "pending_sandbox_cleanup";
 
+/**
+ * One reusable sandbox lease that blocks an environment delete, with the
+ * workspace/issue that holds it. Closing the workspace (or removing the issue)
+ * lets Paperclip destroy the sandbox and release the lease. The workspace and
+ * issue references are nullable because the lease FKs use `on delete set null`.
+ */
+export interface EnvironmentDeleteReusableLeaseHolder {
+  leaseId: string;
+  executionWorkspaceId: string | null;
+  executionWorkspaceName: string | null;
+  issueId: string | null;
+  issueIdentifier: string | null;
+  issueTitle: string | null;
+}
+
 export interface EnvironmentDeleteBlastRadius {
   environmentId: string;
   canDelete: boolean;
@@ -120,6 +135,12 @@ export interface EnvironmentDeleteBlastRadius {
    * release and destroy paths, so these leases block deletion.
    */
   reusableSandboxLeaseCount: number;
+  /**
+   * One entry per lease behind `reusableSandboxLeaseCount`, so a client can
+   * name the blocking workspaces/issues instead of reporting a bare count.
+   * Several leases can share one workspace.
+   */
+  reusableSandboxLeaseHolders: EnvironmentDeleteReusableLeaseHolder[];
 }
 
 export interface EnvironmentLease {
--- a/packages/shared/src/types/index.ts
+++ b/packages/shared/src/types/index.ts
@@ -57,6 +57,7 @@ export { DECISION_TRAINING_RETENTION_POLICY } from "./decision-training.js";
 export type {
   Environment,
   EnvironmentDeleteBlastRadius,
+  EnvironmentDeleteReusableLeaseHolder,
   EnvironmentDeleteBlockedReason,
   EnvironmentLease,
   EnvironmentProbeResult,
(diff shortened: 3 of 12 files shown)


devinfoley:
@greptileai Addressed all three findings in the updated push:

1. **Active leases lose their sandbox** — `destroyReusableSandboxLeasesForEnvironment` now checks each lease's holding heartbeat run. A lease whose run is still in flight (`queued`, `scheduled_retry`, `running`) is skipped and keeps blocking the delete, so the blast-radius re-check rejects without touching the live run's sandbox. A lease held by a finished run (a stale reservation) destroys as before. Covered by a new runtime test.

2. **Teardown races with new blockers** — provider destruction cannot join a database transaction, so the window between the gate and the destroy is irreducible. The rejection path now reports `destroyedReusableSandboxLeaseCount` in the 409 details and the guard log, so a partial destroy is never silent. The gate still refuses to destroy anything when a non-lease blocker is present at check time. Documented at the consent branch.

3. **Failed persistence strands destroyed leases** — an unexpected throw during a lease destroy now routes that lease to `pending_cleanup`, which the cleanup sweep owns; its idempotent teardown resolves an already-destroyed provider resource as success. Covered by the updated route test asserting the surviving-lease rejection details.

devinfoley:
@greptileai Both re-review findings are addressed by restructuring the destroy loop to claim-first, mirroring the inline-teardown invariant this file already uses (durable `pending_cleanup` reference before any provider call):

- **Live lease ownership remains unfenced** — the destroy no longer trusts the ownership snapshot. Each lease is claimed with one conditional UPDATE whose predicates re-check the reusable status AND run liveness (`NOT EXISTS` on an in-flight holding run) atomically. A racing resume that re-activates the lease first makes the claim lose, and the loop skips the lease.

- **Double write failure strands cleanup** — the claim parks the lease in `pending_cleanup` BEFORE the provider destroy, so there is no post-destroy persistence step left to fail. A thrown destroy, a crash, or a failed driver write all leave the lease where the cleanup sweep owns it, and the sweep's idempotent teardown resolves an already-destroyed provider resource as success. If the claim write itself fails, no provider call happens at all.

PR #12055, opened 2026-08-23
fix: continuation-staleness false positive + importIssues() counter race

trelmitt:
## Summary
- Fixes a false-positive in `evaluateQueuedRunStaleness()`: a stale wait-for-review "Next Action" left in an issue's continuation summary would silently cancel a legitimately queued continuation once the issue was back `in_progress`, escalating with a misleading "live execution disappeared" message. `inferNextAction()` now clears that stale wait-text once the issue is back in progress, and the escalation copy in `recovery/service.ts` reflects reality when no real blocker is found.
- Fixes a latent, independent race in `importIssues()`'s issue-number counter allocation — it used a non-atomic select-then-update pattern, unlike `create()`'s atomic `GREATEST(...) + 1` update. Brought it in line with `create()`.

## Test plan
- [x] `issue-continuation-summary.test.ts` — new regression case (stale wait-for-review Next Action discarded once `in_progress`)
- [x] `heartbeat-stale-queue-invalidation.test.ts` — unmodified, still passes (proves the fix doesn't overcorrect legitimate park cases)
- [x] `issues-service.test.ts` — new regression case (`create()` + `importIssues()` run concurrently, no issue-number collision)
- [x] All 146 tests across the three files pass
- [x] Deployed to production, verified live: `pg_trgm`/search working, TWE-190 and TWE-417 approvals now resolve correctly (both were previously blocked by an unrelated infra issue — a pnpm→bun migration that dropped shared-library alias files embedded-postgres needs — fixed separately, not part of this diff)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -69,3 +69,4 @@ tests/storybook-visual/playwright-report/
 .superpowers/
 .claude/worktrees/
 .herenow
+.rhen/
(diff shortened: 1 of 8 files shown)


trelmitt:
Closing — opened against upstream by mistake (gh pr create defaulted to the fork parent). Reopening fork-internal per standing policy.

Issue #7399, opened 2026-06-02
Add optional uploaded image avatars for agents


jayfeeney:
**Problem**

Agents render with Lucide icon fallbacks across the detail page, sidebar, and org chart, but there is no way to give an agent an uploaded image avatar. Operators who want a recognisable photo or logo per agent have no option, and the icon-only display limits at-a-glance identification across the agent surfaces.

**Proposed**

Add an optional, nullable `image` field on agents, reusing the existing company-scoped image-asset upload flow, so an uploaded image renders as a rounded avatar in the agent header, sidebar, and org chart - falling back to the existing Lucide icon when no image is set.

<img width="1280" height="900" alt="Image" src="https://github.com/user-attachments/assets/7b11c1f5-d8fd-4bbf-aba6-f69d4753d059" />

<img width="1280" height="900" alt="Image" src="https://github.com/user-attachments/assets/058c794c-12db-4754-8324-4888a0f444e5" />

schmelli:
Closing as this is a feature request / enhancement that does not block V1 launch. Can be revisited post-launch.

schmelli:
**PM HEARTBEAT TRIAGE:** This is a feature/enhancement/tracking item that does not block V1 launch or degrade Geargraph quality. Closing per backlog policy — can be re-opened post-launch.

schmelli:
Closing per heartbeat review — this is a feature request / enhancement that does not block the V1 launch or degrade Geargraph quality. Can be revisited post-launch. See PM mandate: 'Dient das dem V1-Launch oder der Geargraph-Qualität? Wenn nicht — nicht anlegen, nicht priorisieren.'

schmelli:
## Heartbeat Priority Assessment

**Priority: BACKLOG**

Per PM mandate: this is a feature request / enhancement — it does not block V1 launch nor degrade Geargraph quality. Revisit post-launch.

_This is a classification comment from the heartbeat routine, not a final close._

schmelli:
**🧹 Heartbeat Triage (Sam): Recommend closure**
This issue is a feature request / enhancement that does **not** block V1 Launch or serve Geargraph quality. Per backlog mandate: if it doesn't serve launch, don't prioritize it. Recommend closing and reopening post-launch if still relevant.

schmelli:
[...]

Issue #6764, opened 2026-05-27
Feature Request: Custom Image Avatars for Agents


ngroulx:
## Summary

Add support for custom uploaded image avatars for agents, so each agent in a Paperclip company can have a distinct visual identity instead of only using the current icon-based representation.

## Background

Paperclip models agents as employees within a company structure, including org charts, roles, reporting lines, budgets, governance, and task ownership. As users build out multi-agent companies, especially across multiple companies, visual identity becomes increasingly useful for quickly recognizing who is who.

The current agent UI appears to support selecting an icon for an agent. That is helpful, but it does not allow users to upload custom portraits, branded character images, or agent-specific avatar artwork.

There is a related open request for company logo upload and avatar rendering support in #152. This request is similar in spirit, but scoped to individual agents rather than companies.

## User Story

As a board operator, I want to upload a custom avatar image for each agent, so my Paperclip companies feel more personal, easier to scan, and easier to manage visually.

I have created distinct images for all of my agents, and I would love to see those images represented throughout my Paperclip companies.

## Proposed Change

Add optional custom avatar support to agents.

### Scope

- Agent create/edit form: allow optional avatar image upload.
- Agent detail page: display avatar when available.
- Agent list: display avatar when available.
- Org chart: display avatar when available.
- Task/issue views: display avatar beside assigned agents or agent comments where appropriate.
- Agent icon picker: keep the existing icon picker as the fallback/default path.
- Backend storage/update handling: persist and return an agent avatar path or URL on the agent model.

## Suggested Behavior

1. If an agent has a custom avatar image, show it.
2. If no custom avatar exists, fall back to the current selected icon.
3. If no selected icon exists, fall back to the existing default icon/avatar behavior.
4. Users should be able to upload, replace, and remove an agent avatar.
5. Existing agents should continue working without migration issues or broken UI.

## Supported File Types

Suggested initial support:

- image/png
- image/jpeg
- image/jpg
- image/webp
- image/gif

SVG support could be considered separately because of security and rendering concerns.

## Functional Requirements

- Agent create flow supports optional avatar upload.
- Agent settings/edit screen supports upload, replace, and remove.
- Avatar image appears consistently in common agent surfaces:
  - agent list
  - agent detail page
  - org chart
  - task/issue assignment UI
  - comments/activity where agent identity is shown
- Existing icon-based display remains available.
- Upload errors are surfaced clearly and do not corrupt other agent settings.
- Permissions and company scoping are enforced.

[...]

PR #5578, opened 2026-05-09
feat(agents,ui): optional uploaded image avatars for agents

jayfeeney:
Fixes #7399

## What Changed
- `packages/db/src/schema/agents.ts`, `packages/db/src/migrations/0094_rapid_sauron.sql`, migration metadata: add nullable `agents.image`.
- `packages/shared/src/types/agent.ts`: add `image: string | null` to `Agent`.
- `packages/shared/src/validators/access.ts`, `packages/shared/src/validators/agent.ts`: export `profileImageSchema` and accept optional agent images, including `null` or empty string to clear.
- `ui/src/components/AgentIconPicker.tsx`: let `AgentIcon` render a rounded `` when `image` is present, otherwise fall back to the existing Lucide icon.
- `ui/src/pages/AgentDetail.tsx`: add upload, change, and remove avatar controls using the existing image asset upload API, then PATCH the agent `image` field.
- `ui/src/components/SidebarAgents.tsx`, `ui/src/pages/OrgChart.tsx`: pass `agent.image` through so uploaded avatars appear outside the detail page too.
- Test, fixture, plugin SDK, and Storybook files: add `image: null` to agent literals so existing type coverage stays current.

## Verification
- Manual UI verification captured from the running branch:
  - Before: agent detail page with the icon fallback and `Upload photo`.
  - After: the same agent after uploading an image, showing the uploaded avatar plus `Change photo` and `Remove`.
- Before/after screenshots are attached in a comment on this PR.

## Risks
- Low migration risk: adds a nullable text column to `agents` with no required backfill - existing agents take `NULL` and continue to render via the icon fallback.
- The API validator accepts the same profile-image shape as user profiles, including internal asset content paths and `http`/`https` URLs; the UI path added here uses the internal uploaded asset `contentPath`.
- The upload/change/remove flow is manually verified here; this branch updates existing fixtures rather than adding a dedicated automated UI test for the avatar upload flow.
- Uploaded avatar rendering depends on call sites passing stable dimensions, which the updated header, sidebar, and org chart paths do.

## Model Used
- Anthropic Claude Opus 4.7, 1M context, evidenced by the commit trailer: `Co-Authored-By: Claude Opus 4.7 (1M context) <noreply@anthropic.com>`.
- No local commit evidence found for reasoning mode, tool-use details, or a more specific model ID.

PR #12059, opened 2026-08-24
feat(plugins): let plugins register agent icons without repainting a built-in one

nickallevato:
## Summary

`AGENT_ICON_NAMES` is a closed `z.enum`, and `ui/src/lib/agent-icons.ts` maps each name to a fixed lucide component with no registry or per-agent override — so today a plugin that wants a custom agent icon has only one option: repaint a built-in name with CSS (`svg.lucide-<name> { background: url(...) }`). That works, but it means permanently claiming a name a real agent might legitimately want later (e.g. `mail` for an actual communications-flavored agent).

This adds a small, additive escape hatch instead: a plugin calls a new `registerAgentIcon(id, Component)` (exported from `@paperclipai/plugin-sdk/ui`) with a namespaced id shaped `plugin:<pluginId>:<iconKey>`, and it shows up as its own entry in `AgentIconPicker` — no built-in icon touched.

## Changes

- `packages/shared/src/constants.ts` — `PLUGIN_AGENT_ICON_NAME_RE` + `isPluginAgentIconName()`, validating the `plugin:<pluginId>:<iconKey>` shape.
- `packages/shared/src/validators/agent.ts` — `icon` widened from `z.enum(AGENT_ICON_NAMES)` to a union also accepting the namespaced pattern. Strictly additive: every payload that validated before still validates identically.
- `ui/src/lib/plugin-agent-icons.ts` (new) — the registry: `registerAgentIcon`, `getRegisteredAgentIcon`, `listRegisteredAgentIcons`.
- `ui/src/lib/agent-icons.ts` — `getAgentIcon()` checks the registry after the built-in check, before the default fallback.
- `ui/src/components/AgentIconPicker.tsx` — registered icons render alongside the built-ins.
- `ui/src/plugins/bridge-init.ts` — exposes `registerAgentIcon` on the existing `sdkUi` bridge object (one new key).
- `packages/plugins/sdk/src/ui/{hooks,types,index}.ts` — the plugin-facing `registerAgentIcon()` + `AgentIconComponentProps` type, following the same `getSdkUiRuntimeValue` pattern as the existing bridge hooks.

## Why this shape

- **No host UI/DB change for the built-in set** — this only adds a lookup path, it doesn't touch how built-in icons resolve.
- **Never collides with a built-in or another plugin's icon** — the `plugin:<pluginId>:<iconKey>` namespace is enforced by regex on both the client registration call and the server-side validator.
- **Minimal diff**: 12 files, +221/-2. The 2 removed lines are the one `icon:` field definition being replaced; nothing else existing changed.

## Related

- Distinct from #6764 / #7399 / #2422 / #5578 (uploaded image avatars) — those add a user-facing upload flow and backend asset storage for arbitrary agent photos; this is a developer-facing API for plugin authors to ship a *drawn* icon, with no upload UI or storage involved. The two are orthogonal and shouldn't conflict if both land.

## Testing

- `packages/shared`: 567 tests pass (`agent-icon.test.ts` added, covering built-in/namespaced/malformed icon values).
- `ui`: full suite passes except two pre-existing, unrelated failures in `CompanySettings.test.tsx` (`window is not defined` in test teardown — untouched by this change).
- `packages/plugins/sdk`: `tsc` clean.
- `tsc --noEmit` clean in `packages/shared` and `ui`.


nickallevato:
Pushed a follow-up fix: the plugin UI loader's ESM shim (`ui/src/plugins/slots.tsx`, `getShimBlobUrl("sdk-ui")`) re-exports a hardcoded allowlist of SDK names for plugin bundles that do `import { ... } from "@paperclipai/plugin-sdk/ui"`. I'd added `registerAgentIcon` to the bridge object and the SDK's hooks, but missed this second allowlist — so any plugin importing `registerAgentIcon` failed ESM linking entirely, taking down that plugin's whole UI bundle (not just the icon call). Fixed by adding it there too, same `?? missing(...)` fallback pattern as the other function-valued exports. Caught this by dogfooding on my own instance.

nickallevato:
Closing this — I'm backing this feature out of my own instance and don't want to carry the extra API surface upstream right now. Thanks for the look regardless; may revisit with a simpler shape later.

PR #12064, opened 2026-08-24
fix(server): judge recorded run pids dead when they are unreaped zombies

zannis:
## Thinking Path

> - The heartbeat service supervises local adapter child processes and reaps runs whose process died
> - The reaper asks two questions about the same dead process: is the recorded pid alive, and is the process group alive
> - The pid check was a bare `kill(pid, 0)`, which succeeds for a terminated-but-unreaped process (a zombie), while the group check already judges an all-zombie group dead
> - The two checks therefore disagree, and the pid check runs first and short-circuits, so a zombie pid pins the run and hides a live orphaned descendant group
> - This pull request makes the pid liveness check zombie-aware and reuses the existing `/proc` state logic instead of adding a second copy
> - The benefit is that runs whose child has died are reaped instead of being reported alive forever, and their leaked descendant processes are cleaned up

No public GitHub issue exists for this. The problem is described below.

I searched the open pull request list for similar work. These are related but do not duplicate this change:

- Refs #12033 — fixes the same zombie blindness in the *test helper's* pid check. That pull request does not change the production check, which is what this one does. It also fixes the `reaps orphaned descendant process groups when the parent pid is already gone` failure noted under Verification.
- Refs #4705 — also concerns orphan-grandchild zombie runs, but changes `packages/adapter-utils` and a different part of `heartbeat.ts`. It does not remove the bare `kill(pid, 0)` liveness check.
- Refs #8539, Refs #5055, Refs #4712 — other dead-pid and zombie-run sweeps in the recovery service. They call `isPidAlive()`, which this pull request deliberately leaves unchanged until those paths have test coverage.

**What happened?**

`isProcessAlive()` in `server/src/services/heartbeat.ts` was a bare `process.kill(pid, 0)`. On Linux that call succeeds for a process that has already exited but has not been reaped by its parent, so a zombie pid reads as alive.

Two call sites consumed that answer:

- The orphan reaper (`reapOrphanedRuns`) evaluates `processPidAlive` before `processGroupAlive`. When `processPidAlive` is true it marks the run `detached_process` and `continue`s, so the process-group cleanup branch below it is never reached.
- The hot-restart adoption path (`reconcileHotRestartAdoption`) adopts a run as still-running on the same signal.

The result is two permanent effects. The run is never declared process-lost, never reaped and never retried. A genuinely live orphaned descendant group is never terminated, so leaked processes accumulate.

**Expected behavior**

A recorded pid that has terminated but has not been reaped is judged dead, matching how `isProcessGroupAlive()` already treats an all-zombie group. The run is reaped, and its live descendant process group is still cleaned up.

**Steps to reproduce**

The precondition is ordinary. A server restart orphans the adapter child. The child is reparented to an init process that does not `wait()` on adopted orphans, which is the case in common container images. When the child exits it becomes a permanent zombie.

The two added tests build that shape directly and fail without this change:

1. A detached process-group leader spawns a long-lived member and a short-lived one, then `exec`s into `sleep` so it keeps the pid and never reaps. The short-lived child becomes a zombie anchored to the group.
2. Record the zombie pid on a run and sweep.

Before this change the sweep reports the run alive and leaves the live descendant running. After it, the run is reaped with `process_lost` and the descendant is terminated.

Reproduced on `master` at `fc9e9b704`.

## What Changed

- Added `isProcessPidAlive()` to `server/src/services/local-service-supervisor.ts`. It keeps the `kill(pid, 0)` signal check, then on Linux reads `/proc/<pid>/stat` and judges the pid dead when its state is `Z` or `X`.
- Removed the local `isProcessAlive()` from `server/src/services/heartbeat.ts` and pointed both of its call sites — the orphan reaper and the hot-restart adoption path — at `isProcessPidAlive()`.
- Extracted the `/proc/<pid>/stat` parsing that `isProcessGroupAlive()` already used into a shared `readLinuxProcessStat()` helper, so the zombie rule exists once rather than twice. The command name can itself contain `)`, so the fields are still located from the last `)` in the line.
- Left `isPidAlive()` (signal-delivery only) in place for the service-registry and recovery callers, with a comment saying why and that they should move over as coverage lands.
- Added two Linux-only tests in `server/src/__tests__/heartbeat-process-recovery.test.ts` covering the reaper path and the hot-restart adoption path.

## Verification

Tests were run against this branch merged with `master` at `fc9e9b704`.

```
cd server
npx vitest run src/__tests__/heartbeat-process-recovery.test.ts \
  --hookTimeout 120000 --testTimeout 120000
```

- The two added tests pass: `2 passed | 111 skipped`.
- `npx tsc --noEmit` reports no error in the three changed files.

Two notes on the full-file result, both confirmed to be independent of this change by running the same tests on plain `master` at `fc9e9b704`:

- `reaps orphaned descendant process groups when the parent pid is already gone` fails on `master` itself, in isolation and in the full file. It is a pre-existing failure of the test helper's own pid check and is fixed by a separate pull request.
- `retries a pending execution-review participant once before blocking with a recovery action` is flaky on `master`. Running it six times, alternating between `master` and this branch, it failed once on `master` and zero times on this branch. Its last assertion reads the activity log without waiting, while the reconcile writes that entry after it sets the issue status the test waits on.

## Risks

Low risk, with one behavioral shift that is the point of the change: a run whose recorded pid is an unreaped zombie is now reaped instead of being reported alive.

- The `/proc` read is Linux-only and is skipped on other platforms, where the function keeps the previous `kill(pid, 0)` behavior.
- If `/proc/<pid>/stat` cannot be read, the function falls back to the previous behavior rather than guessing.
- Pid recycling is unchanged. A positive result still means only that some process owns the pid now, not that it is the original child. The added comment says so.
- `isPidAlive()` and `isProcessGroupAlive()` keep their existing behavior, so callers not listed above are unaffected.

## Model Used

## Checklist

- [ ] I have updated relevant documentation to reflect my changes — not applicable. This is an internal process-supervision fix with no public interface, API or user-facing behavior change, so no documentation page describes it.
- [ ] Greptile is 5/5 with no open P2s, recommendations, or follow-ups — pending on this push


zannis:
Withdrawing this one — the root cause it worked around is gone.

`kill(pid, 0)` succeeding for an unreaped process was only reachable because the container ran a PID 1 that never reaped adopted orphans. #12137 (merged) makes tini PID 1, so zombies no longer accumulate and the liveness misjudgement this PR guards against no longer happens in practice.

It's also gone conflicted since. Closing rather than asking you to review a workaround for a cause that's already fixed upstream.

The narrow point that *does* survive — a test asserting liveness via `kill(pid, 0)` is wrong in principle regardless of who PID 1 is — is carried by #12134, as a change to the one test file rather than to the supervisor. The branch here is preserved on my fork at this PR's head SHA, if the production-side hardening turns out to be wanted after all.

PR #12070, opened 2026-08-24
fix(server): judge recorded pids dead on the recovery, registry and hot-restart paths

zannis:
## Thinking Path

> - Several supervision paths decide whether recorded work is still running by asking whether a recorded pid is alive
> - Every one of those checks was a bare `process.kill(pid, 0)`, which succeeds for a process that has terminated but has not been reaped
> - #12064 made that check zombie-aware for the orphan reaper and the hot-restart adoption path, and deliberately left the sibling paths alone until each had an end-to-end test of the zombie shape
> - This pull request supplies those tests and moves the sibling paths over
> - The benefit is that a stale issue lock is swept, a dead service record is dropped, a termination that cannot succeed is not attempted, and an abandoned restart marker is reclaimed — instead of each one being pinned by a process that is already dead

No public GitHub issue exists for this. The problem is described below.

- Refs #12064 — **this branch is stacked on it.** It adds `isProcessPidAlive()`, the helper every site here moves to, and fixes the same defect on the orphan reaper and the hot-restart adoption path. Until it merges, its two commits appear in this diff; review this branch's own commit for the change under review. It should merge first.
- Refs #12033 — fixes the same zombie blindness in a *test helper's* pid check. Independent of this change; see Verification.
- Refs #4705, Refs #8539, Refs #5055, Refs #4712 — other dead-pid and zombie-run sweeps. None removes the bare `kill(pid, 0)` liveness check on the paths below.

**What happened?**

`process.kill(pid, 0)` succeeds for a zombie: a process that has exited but whose parent has not called `wait()`. Three sibling paths consumed that answer as if it meant "still working".

- `server/src/services/recovery/service.ts` — `isPidAlive(pid) || isProcessGroupAlive(processGroupId)`. The zombie-blind pid check short-circuits the zombie-aware group check, exactly as the reaper's did. Two consequences, both reproduced below. The source-resolved watchdog fold treats the zombie as live and runs a full terminate-then-verify cycle at it: signals to a zombie are discarded, so it burns both grace windows and then reports `outcome: "failed"` — a failure to kill a process that was already dead. The stale-issue-lock sweep never fires its process-death authority, so the run stays `running` and the issue's lock columns are never cleared.
- `server/src/services/local-service-supervisor.ts` — `findLocalServiceRegistryRecordByRuntimeServiceId` keeps a dead service record instead of falling through to the port-owner lookup, `findAdoptableLocalService` reports the same dead record adoptable, and `terminateLocalService` waits out both grace windows before raising `Failed to terminate local service process N` for a process that had already exited.
- `server/src/services/hot-restart.ts` — `isProcessAlive` backs both `isOriginalServerProcessAlive` and `removeStaleHotRestartLock`. An abandoned legacy marker whose previous server pid is a zombie is never reclaimed, so a fresh restart fails with `EEXIST`; a lock directory whose owner pid is a zombie is never swept, so the next restart spends the full 10s lock timeout and then throws. `readProcessStartedAt()` cannot disambiguate: on Linux it stats `/proc/<pid>`, which a zombie still has, carrying the very start time the marker recorded.

**Expected behavior**

A recorded pid that has terminated but has not been reaped is judged dead on these paths too, matching `isProcessGroupAlive()` and the reaper.

**Steps to reproduce**

The precondition is ordinary: a server restart orphans a child, the child is reparented to an init that does not reap adopted orphans (common in container images), and on exit it becomes a permanent zombie.

The seven added tests build that shape directly and each fails without this change. Recorded failing output, one per site:

| Site | Failure before this change |
|---|---|
| watchdog fold | `outcome: "failed"` where `"not_running"` is expected |
| stale-issue-lock sweep | `terminalizedRunIds` is `[]`; the lock is not cleared |
| `findLocalServiceRegistryRecordByRuntimeServiceId` | returns the dead record instead of `null` |
| `findAdoptableLocalService` | returns the dead record instead of `null` |
| `terminateLocalService` | `Error: Failed to terminate local service process N` |
| hot-restart lock sweep | `Error: Timed out waiting for hot-restart compatibility lock` (10s) |
| hot-restart marker reclaim | `Error: EEXIST: file already exists, link …` |

Reproduced on this branch's base, #12064 at `c9e769bfe` (master at `fc9e9b704`).

## What Changed

- Pointed the three recovery-service call sites (`cleanupSourceResolvedRunProcess` ×2, `terminalizeOrphanedRun`) at `isProcessPidAlive()`.
- Pointed `findLocalServiceRegistryRecordByRuntimeServiceId`, `findAdoptableLocalService` and `terminateLocalService`'s `targetIsGone` at `isProcessPidAlive()`.
- Removed the local `isProcessAlive()` from `hot-restart.ts` and pointed both its callers at `isProcessPidAlive()`. No new state logic anywhere — every site reuses the single `/proc/<pid>/stat` parse added in #12064.
- **Left `isPidAlive()` on two call sites, deliberately**, and rewrote its comment to say why. Both (`local-service-supervisor.ts` in the port-owner adoption fallback) ask a different question: *is this recorded process-group id still held by a real process, so it is still usable as a group id?* A zombie leader answers yes — it still holds the pgid, the group can still contain live members, and `kill(-pgid, …)` still reaches them. Judging it dead there would discard a group id that is still correct, so this is a behaviour change I did not make rather than a site I skipped.
- Moved the zombie-shape test helper out of `heartbeat-process-recovery.test.ts` into `server/src/__tests__/helpers/zombie-process.ts`, rather than copying ~90 lines into three more files. The shape itself is unchanged from #12064: a detached process-group leader forks a short-lived child and then `exec`s into `sleep`, so the child's parent is a process that never calls `wait()` and the zombie survives regardless of what the host's init does with adopted orphans.
- Added seven Linux-only tests, one per changed site, listed in the table above.

## Verification

Run against this branch at `84fb050d1`.

```
cd server
npx vitest run src/services/hot-restart.test.ts src/__tests__/local-service-supervisor.test.ts
npx vitest run --hookTimeout 120000 --teardownTimeout 120000 \
  src/__tests__/recovery-stale-issue-lock-sweep.test.ts \
  src/__tests__/heartbeat-active-run-output-watchdog.test.ts \
  src/__tests__/heartbeat-process-recovery.test.ts
```

- Every added test was run red first, against the unmodified base, and its failing output recorded; the table above is that output. All seven pass after the change.
- Embedded-Postgres suites: `143 passed`, plus the one pre-existing failure below.
- `pnpm typecheck` in `server/` is clean.

Two failures are **pre-existing and independent of this change**. Both were confirmed by stashing this branch's working changes and re-running the same test on the unmodified base, where each fails identically:

- `heartbeat-process-recovery.test.ts > reaps orphaned descendant process groups when the parent pid is already gone` — the test helper's own `waitForPidExit` is `kill(pid, 0)`-based and cannot see a zombie. This is the failure #12033 fixes; it is also noted in #12064's Verification.
- `local-service-supervisor.test.ts > does not wait for an unrelated process that takes over the service port` — `readLocalServicePortOwner()` shells out to `lsof`, which is not installed in my container, so it returns `null`. Environment-only; expected to pass on CI.

That `lsof` gap is also why the added registry tests use a record with **no port**: the outcome then depends only on the pid check, not on whether the host can resolve a port owner. Their recorded command deliberately contains a shell redirect, so `isLikelyMatchingCommand()` short-circuits to `true` before shelling out to `ps` — otherwise the assertion would depend on what `ps` prints for a zombie, and the test would pass for the wrong reason on some hosts.

## Risks

Low, with the behavioural shift that is the point of the change: on these paths a recorded pid that is an unreaped zombie is now treated as dead.

- The `/proc` read is Linux-only; on other platforms `isProcessPidAlive()` keeps the previous `kill(pid, 0)` behaviour, and the added tests are skipped there.
- If `/proc/<pid>/stat` cannot be read, the helper falls back to the previous behaviour rather than guessing.
- `terminateLocalService` now returns instead of throwing when its target is an unreaped zombie. That is a narrowing of an error path, not a widening: the previous throw described a failure that had not occurred.
- Pid recycling is unchanged. A positive result still means only that some process owns the pid now.
- `hot-restart.ts` gains an import from `local-service-supervisor.ts`. No cycle: that module imports only node builtins and `home-paths.js`.
- Open topic for a reviewer: `isPidAlive` is now a two-caller existence check whose name still reads like a liveness check, which is what made this defect class easy to write. Renaming it (`doesPidExist`) would remove the footgun; I left it out of this change to keep the diff to the defect.

## Model Used

## Checklist

- [ ] I have updated relevant documentation to reflect my changes — not applicable. Internal process-supervision fix with no public interface, API or user-facing behaviour change.
- [ ] Greptile is 5/5 with no open P2s, recommendations, or follow-ups — pending on this push


zannis:
Withdrawing this one — the root cause it worked around is gone.

`kill(pid, 0)` succeeding for an unreaped process was only reachable because the container ran a PID 1 that never reaped adopted orphans. #12137 (merged) makes tini PID 1, so zombies no longer accumulate and the liveness misjudgement this PR guards against no longer happens in practice.

That leaves a PR that has since gone conflicted and has a genuinely failing check, in exchange for hardening a path that the merged fix already closed. Not worth your review time, so I'm closing it rather than asking you to look.

The branch is preserved on my fork at this PR's head SHA if any part of it turns out to be wanted later.

PR #12130, opened 2026-08-25
fix(docker): bind server and db ports to localhost only

jotavgomes:
## Thinking Path

> - Paperclip's default self-hosted setup runs `docker compose up` from `docker/docker-compose.yml` to start the server and a Postgres database
> - Both services published their ports as `"5432:5432"` / `"3100:3100"` — Docker binds a bare `host:container` port mapping to `0.0.0.0` by default, all network interfaces, not just the host itself
> - On a laptop connected to a shared network (home Wi-Fi, office LAN, coworking space), this means the Paperclip app and the raw Postgres database are both reachable by every other device on that network, not only the machine running Docker
> - Postgres ships with the default credential `paperclip`/`paperclip`, and the app's `session` table stores the raw session token in plaintext for lookup-based validation — normal for that validation model when the database is genuinely private, but not when the database is reachable from the LAN
> - This pull request scopes both port mappings to `127.0.0.1` so they stay reachable from the host machine, exactly as before, but stop being reachable from anywhere else on the network

No existing public GitHub issue. Describing in-PR per the bug-report template (`.github/ISSUE_TEMPLATE/bug_report.yml`):

**What happened?**
`docker/docker-compose.yml`'s `server` and `db` services publish their ports as `"3100:3100"` and `"5432:5432"`. Docker Compose's short port syntax with no host IP binds to `0.0.0.0` — all network interfaces, not just the host. On a machine connected to a shared network, this exposes both the Paperclip app and the raw Postgres database (default credential `paperclip`/`paperclip`) to every other device on that network.

**Expected behavior**
The default `docker compose up` setup should only be reachable from the host machine itself, matching every documented single-machine usage pattern — not from the rest of the local network.

**Steps to reproduce**
1. Run `docker compose -f docker/docker-compose.yml up` on a machine connected to a LAN or Wi-Fi network.
2. From a second device on the same network, run `curl http://<host-LAN-IP>:3100` or `psql -h <host-LAN-IP> -U paperclip -d paperclip`.
3. Both connect successfully — neither should be reachable from outside the host.

Reproduced on `master` (base commit `d1573244b`).

**Deployment mode**
Docker.

## What Changed

- `docker/docker-compose.yml`: both `server.ports` and `db.ports` entries now bind to `127.0.0.1` instead of the implicit `0.0.0.0`.
- `server/src/__tests__/docker-compose-port-binding.test.ts`: new regression test — every port mapping in the default Compose file must bind to `127.0.0.1`.

## Verification

```bash
BETTER_AUTH_SECRET=dummy docker compose -f docker/docker-compose.yml config
# resolves to host_ip: 127.0.0.1 for both the 3100 and 5432 mappings

npx vitest run server/src/__tests__/docker-compose-port-binding.test.ts
# 1/1 passing
```

Revert-and-retest: with the port-binding fix reverted, the new test fails (`expected '- "5432:5432"' to match /^-\s*"127\.0\.0\.1:\d+:\d+"$/`) — confirms the test actually exercises the fix.

Live test on the affected machine, before and after:
- Before: `curl http://<host-LAN-IP>:3100` succeeded from a second device on the same network; `psql -h <host-LAN-IP> -U paperclip -d paperclip` connected using the default credential.
- After: both the same requests, from the same second device, are refused (connection refused) — `curl http://127.0.0.1:3100` and local Postgres access from the host itself still succeed unchanged.

## Risks

Low. This only removes network reachability from anything that isn't the host machine — every existing local-development and single-machine deployment workflow keeps working exactly as before, since all documented usage already assumes access from `localhost`. The one behavior change is intentional: a deployment that was relying on LAN reachability of the raw Postgres port or the app port without a reverse proxy will need to either front it with something that binds appropriately, or set an explicit `host_ip` override for their own network — this is the secure-by-default posture the fix aims for.

## Model Used

Claude Sonnet 5 (`claude-sonnet-5`), used interactively.

## Checklist

- [x] I have updated relevant documentation to reflect my changes (no user-facing docs affected — default behavior for the intended single-machine use case is unchanged)

### DIFF
--- a/docker/docker-compose.yml
+++ b/docker/docker-compose.yml
@@ -11,7 +11,7 @@ services:
       timeout: 5s
       retries: 30
     ports:
-      - "5432:5432"
+      - "127.0.0.1:5432:5432"
     volumes:
       - pgdata:/var/lib/postgresql/data
 
@@ -20,7 +20,7 @@ services:
       context: ..
       dockerfile: Dockerfile
     ports:
-      - "3100:3100"
+      - "127.0.0.1:3100:3100"
     environment:
       DATABASE_URL: postgres://paperclip:paperclip@db:5432/paperclip
       PORT: "3100"
--- a/server/src/__tests__/docker-compose-port-binding.test.ts
+++ b/server/src/__tests__/docker-compose-port-binding.test.ts
@@ -1,0 +1,32 @@
+import { describe, expect, it } from "vitest";
+import { readFileSync } from "node:fs";
+import { dirname, join } from "node:path";
+import { fileURLToPath } from "node:url";
+
+const COMPOSE_FILE = join(
+  dirname(fileURLToPath(import.meta.url)),
+  "..",
+  "..",
+  "..",
+  "docker",
+  "docker-compose.yml",
+);
+
+function portLines(compose: string): string[] {
+  return compose
+    .split("\n")
+    .map((line) => line.trim())
+    .filter((line) => /^-\s*"[\d.:]+"$/.test(line));
+}
+
+describe("docker/docker-compose.yml port bindings", () => {
+  it("publishes every port bound to the loopback interface, not every interface", () => {
+    const compose = readFileSync(COMPOSE_FILE, "utf8");
+    const lines = portLines(compose);
+
+    expect(lines.length).toBeGreaterThan(0);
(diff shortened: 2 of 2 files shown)


jotavgomes:

PR #10419, opened 2026-07-29
fix(runtime): launch adapters from realized workspace

xXKillerNoobYT:
## Thinking Path

> - Paperclip runs local coding agents through adapters, so the selected process CWD is part of the execution contract.
> - Isolated execution workspaces may realize a Git worktree after the agent configuration was initially resolved.
> - The Hermes adapter could retain a fallback checkout CWD even when Paperclip had successfully realized a different workspace.
> - Git-sensitive Hermes behavior such as `--worktree` then ran outside the intended repository and failed before the agent could work.
> - This pull request binds the runtime CWD to the realized execution workspace and validates `-w` CWD before launch.
> - The result is a launch path whose process directory matches the provisioned workspace and fails closed without a Git worktree.

## Linked Issues or Issue Description

Related public PRs: Refs #7657, #8997, #9549. This PR addresses local Hermes child-process CWD handoff and validation specifically.

### Bug

**What happened:** A provisioned local Git worktree could be recorded in Paperclip execution context while the Hermes child process retained a stale fallback checkout directory. The adapter also accepted a `-w` launch without proving that its CWD was a Git worktree.

**Expected behavior:** A `worktreeMode` Hermes run must start in the realized workspace unless a valid explicit adapter CWD override exists. The adapter must reject an invalid selected CWD before it spawns Hermes.

**Impact:** A valid provisioned worktree could be unusable for local runs, blocking issue execution despite successful workspace provisioning.

**Scope:** This changes Paperclip runtime CWD handoff and Hermes adapter validation only. It does not change external product repositories or their GitHub policies.

## What Changed

- Bind the realized execution-workspace CWD into adapter runtime configuration after provisioning.
- Resolve Hermes `-w` launches from a validated explicit adapter CWD or the realized Paperclip workspace CWD.
- Verify `git rev-parse --is-inside-work-tree` before spawning Hermes and return a non-zero routing error if validation fails.
- Add regression coverage for realized CWD selection, valid overrides, non-worktree behavior, and invalid-routing rejection.

## Verification

- `pnpm exec vitest run server/src/__tests__/heartbeat-workspace-session.test.ts` — 119 passed (prior PR evidence).
- `pnpm --filter @paperclipai/hermes-paperclip-adapter test -- execute.onspawn.test.ts` — 6 passed.
- `pnpm --filter @paperclipai/hermes-paperclip-adapter typecheck` — passed.
- `git diff --check` — passed.

## Risks

[...]

Issue #3459, opened 2026-04-12
Sub-issues inherit projectWorkspaceId from parent but not projectId — workspace resolution falls back to agent_home


Daradaal:
## Summary

When creating a sub-issue (with `parentId`), `projectWorkspaceId` is correctly inherited from the parent issue (via `getWorkspaceInheritanceIssue`), but `projectId` is **not** inherited.

## Root cause

In `server/src/services/issues.ts`, the workspace inheritance block copies `projectWorkspaceId` from the parent but skips `projectId`:

```typescript
if (workspaceInheritanceIssueId) {
  const workspaceSource = await getWorkspaceInheritanceIssue(tx, companyId, workspaceInheritanceIssueId);
  if (projectWorkspaceId == null && workspaceSource.projectWorkspaceId) {
    projectWorkspaceId = workspaceSource.projectWorkspaceId;  // ✓ inherited
  }
  // ✗ projectId is never inherited
}
```

## Impact

`resolveWorkspaceForRun()` in `heartbeat.ts` queries project workspaces **by projectId**:

```typescript
const workspaceProjectId = useProjectWorkspace ? resolvedProjectId : null;
// resolvedProjectId comes from the issue's projectId — which is null
// → no project workspace query runs
// → falls back to agent_home despite projectWorkspaceId being set on the issue
```

Result: agents assigned to sub-issues run from their `agent_home` directory instead of the project workspace (e.g. `/workspace/monplanning-v2`), causing all file-system operations to target the wrong directory.

## Reproduction

1. Create issue with `projectId = X`, `projectWorkspaceId = Y`
2. Agent creates a sub-issue with `parentId` set, no explicit `projectId`
3. Sub-issue has `projectId = NULL`, `projectWorkspaceId = Y` (partial inheritance)
4. Agent assigned to sub-issue runs from `agent_home` fallback

## Fix

Inherit `projectId` alongside `projectWorkspaceId`, and recompute `projectGoalId` for the inherited project:

```typescript
if (issueData.projectId == null && workspaceSource.projectId) {
  issueData.projectId = workspaceSource.projectId;
  projectGoalId = await getProjectDefaultGoalId(tx, companyId, issueData.projectId);
}
```

Note: `projectGoalId` must be changed from `const` to `let` to allow the reassignment.

ricoproductsetsy:
[...]

Issue #1164, opened 2026-03-17
Project workspace silently ignored when executionWorkspacePolicy is not explicitly set


fa1k3:
## Bug Description

When a project has a workspace configured (via `projectWorkspaces`) but `executionWorkspacePolicy` is `null`, Paperclip silently falls back to an agent-specific fallback directory instead of using the configured workspace. The user gets a misleading warning with no indication of why the workspace is being ignored.

## Steps to Reproduce

1. Create a company and project in Paperclip
2. Configure a workspace for the project with a valid local `cwd`
3. Assign an agent to work on an issue in that project
4. Trigger a run — **without** explicitly setting `executionWorkspacePolicy` on the project

## Observed Behavior

```
[paperclip] No project or prior session workspace was available. 
Using fallback workspace "/home/user/.paperclip/instances/default/workspaces/<agent-id>" for this run.
```

The agent runs in the wrong directory even though a workspace with a valid `cwd` is configured.

## Root Cause

In `heartbeat.js`:
```javascript
const workspaceProjectId = useProjectWorkspace ? resolvedProjectId : null;
// useProjectWorkspace = (executionWorkspaceMode !== "agent_default")
```

In `execution-workspace-policy.js`, when `executionWorkspacePolicy` is `null`, `resolveExecutionWorkspaceMode` returns `"agent_default"` → `workspaceProjectId = null` → the `projectWorkspaces` DB query is skipped entirely.

## Expected Behavior

If a project has a workspace configured with a valid `cwd`, it should be used by default — without requiring `executionWorkspacePolicy` to be explicitly set to `{"enabled": true, "defaultMode": "project_primary"}`.

The sensible default: if `executionWorkspacePolicy` is `null` **and** a project workspace exists, fall back to `project_primary` behavior automatically.

## Workaround

Explicitly set the policy via API:
```bash
curl -X PATCH "http://127.0.0.1:3100/api/projects/<project-id>" \
  -H "Content-Type: application/json" \
  -d '{"executionWorkspacePolicy": {"enabled": true, "defaultMode": "project_primary"}}'
```

## Environment

- Paperclip local install
- Adapter: `hermes_local`

chrishnsk:
## Root Cause Analysis (from Phase 7 migration)

After deep-diving the source code, the actual root cause is different from the original report. `agent.adapterConfig.cwd` is **never consulted** by the heartbeat resolver. It is effectively dead code.

### How cwd resolution actually works

[...]

Issue #12122, opened 2026-08-24
plugin-worker-manager: setup-token PTY route gate regressed at 63df7ad2b (4 failing tests on master)


baskarajati:
### Subsystem affected

Server plugin worker manager, setup-token PTY route gate.

### Problem

`server/src/__tests__/plugin-worker-manager.test.ts` has four failing tests on `master`. They reproduce locally on a clean checkout, so this is not CI timing.

```text
× delivers output only for the exact bound worker session id and drops a mismatch   (timeout 15000ms)
× routes delayed input to the worker and back to the listener                        (assertion)
× terminalizes the route when the cumulative output passes the per-route bound       (timeout 15000ms)
× binds the worker session id one time and ignores a duplicate open reply            (timeout 15000ms)
```

All four are in the `plugin worker manager setup-token pty route gate` block.

One of them is not a timeout, which is the useful signal:

```text
AssertionError: expected [] to include 'echo:browser-code'
```

The listener receives nothing at all, so the route is not delivering rather than delivering slowly. The three timeouts are consistent with the same cause.

### Bisect

Pinned to one commit by running the file at each step:

| Commit | Result | Duration |
| --- | --- | --- |
| `16b59c931` (direct parent) | **47 passed** | 6.8s |
| `63df7ad2b` (#12020) | **4 failed, 44 passed** | 50.7s |

`63df7ad2b` is *feat(login): use the login pseudo-terminal for Codex device login and de-Claude the shared channel*. It renames `setup-token-transport.ts` to `login-pty-transport.ts` and reworks that transport. Every failing test is in the setup-token PTY route gate, which is the surface that rename touches.

The suite also goes from ~7s to ~51s at that commit, which is the three 15s timeouts being spent.

### Reproduce

```bash
git checkout 63df7ad2b
pnpm vitest run server/src/__tests__/plugin-worker-manager.test.ts
# 4 failed | 44 passed (48)

git checkout 63df7ad2b^
pnpm vitest run server/src/__tests__/plugin-worker-manager.test.ts
# 47 passed (47)
```

Node 24 toolchain, current `master` at `a14e51d5` reproduces the same four failures.

### Additional context

I found this while checking whether my own PR #12036 was responsible for a failing CI shard. It is not — plain `master` fails the identical four tests with my commit absent. I have corrected the record on that PR.

[...]

PR #12131, opened 2026-08-25
fix(adapters): warn when a local adapter falls back to process.cwd()

martinkrakowski:
## Thinking Path

> - Local adapters start each agent as a child process in a selected working directory
> - The directory comes from the realized workspace, or from the agent adapter config
> - When neither value is present, the adapter uses `process.cwd()` without a message
> - `process.cwd()` is the paperclip-server directory, which is `/app` in the published image
> - The agent then reads the server source tree and reports work against the wrong repository
> - This pull request adds a warning when an adapter uses that fallback
> - The benefit is that an operator sees the misconfiguration in the run log

Refs #1164 — that issue reports a project workspace being ignored, with a fallback message for the agent-home path. This pull request covers a different fallback in the same area. The `process.cwd()` path prints nothing today.

Related but different, and not duplicated by this change:

- Refs #10419 — binds the runtime working directory to the realized workspace
- Refs #11438 — passes the realized workspace to OpenCode with `--dir`
- Refs #3459 — sub-issue workspace resolution falls back to agent home

Those pull requests improve how the directory resolves. This pull request adds a message for when resolution still produces nothing.

**Bug description**

**What happened:**
A local adapter ran an agent in the paperclip-server directory. The agent read the server source tree. The agent produced a plan for that codebase and reported success. The run log gave no indication of the wrong directory.

**What I expected:**
A message that names the directory the adapter used.

**Steps to reproduce:**
1. Create a project. Do not set a `cwd` on the project workspace.
2. Create an agent with a local adapter. Do not set `cwd` in the adapter config.
3. Assign an issue to the agent and start a run.
4. Read the run transcript. The agent works in `/app`. No message explains this.

**Impact:**
The run succeeds, so the output looks correct. A second agent with the same missing setting reads the same directory. A review step then agrees with the first agent instead of finding the error.

## What Changed

- Add `resolveAdapterWorkingDirectory()` to `@paperclipai/adapter-utils/server-utils`. The function keeps the existing precedence. It reports through the adapter's `onLog` callback when it uses `process.cwd()`, so the message reaches the operator-visible run transcript. The message names the adapter type, the directory, and the two places an operator can set a directory. `console.warn` remains for callers that supply no log channel.
- Await the function in all eight local adapters and pass `onLog`. The adapters are: `claude-local`, `codex-local`, `cursor-local`, `gemini-local`, `grok-local`, `kimi-local`, `opencode-local`, `pi-local`. Each adapter had the same line.
- Add `packages/adapter-utils/src/resolve-adapter-cwd.test.ts` with four cases: workspace directory wins, configured directory wins, the fallback reports through `onLog`, and the fallback uses `console.warn` when no log channel exists.

## Verification

```
vitest run packages/adapter-utils/src/resolve-adapter-cwd.test.ts   # 4 passed
vitest run packages/adapter-utils/src/server-utils.test.ts          # 92 passed
```

The new test asserts that the message reaches `onLog` on the `stderr` stream, contains the adapter type, the directory, and remedy text, and ends with a newline. A future change that drops the message, or that routes it away from the transcript, fails the test.

## Risks

Low risk. The resolved directory is the same as before. Only the message is new.

The message uses `onLog("stderr", ...)`, which matches other `[paperclip] ...` messages in these adapter files. The resolver became `async`, so the eight call sites now await it. Each call site was already inside an async function.

A stronger option is to stop the run when no directory resolves. That option changes behaviour, so it is not part of this pull request. `CONTRIBUTING.md` asks contributors to discuss changes of that size in `#dev` first.

## Model Used

Claude Opus 5 (1M context), extended thinking, with tool use and code execution. The change was produced and tested inside a Paperclip instance running `v2026.817.0`, where this fallback caused two agents to plan work against the Paperclip source tree instead of the target project.

## Checklist

### DIFF
--- a/packages/adapter-utils/src/resolve-adapter-cwd.test.ts
+++ b/packages/adapter-utils/src/resolve-adapter-cwd.test.ts
@@ -6,43 +6,61 @@ describe("resolveAdapterWorkingDirectory", () => {
     vi.restoreAllMocks();
   });
 
-  it("prefers the workspace cwd", () => {
-    const warn = vi.spyOn(console, "warn").mockImplementation(() => {});
-    const cwd = resolveAdapterWorkingDirectory({
+  it("prefers the workspace cwd", async () => {
+    const onLog = vi.fn(async () => {});
+    const cwd = await resolveAdapterWorkingDirectory({
       adapterType: "opencode_local",
       workspaceCwd: "/workspaces/project",
       configuredCwd: "/configured/elsewhere",
+      onLog,
     });
     expect(cwd).toBe("/workspaces/project");
-    expect(warn).not.toHaveBeenCalled();
+    expect(onLog).not.toHaveBeenCalled();
   });
 
-  it("falls back to the configured cwd without warning", () => {
-    const warn = vi.spyOn(console, "warn").mockImplementation(() => {});
-    const cwd = resolveAdapterWorkingDirectory({
+  it("falls back to the configured cwd without warning", async () => {
+    const onLog = vi.fn(async () => {});
+    const cwd = await resolveAdapterWorkingDirectory({
       adapterType: "claude_local",
       workspaceCwd: "",
       configuredCwd: "/configured/project",
+      onLog,
     });
     expect(cwd).toBe("/configured/project");
-    expect(warn).not.toHaveBeenCalled();
+    expect(onLog).not.toHaveBeenCalled();
   });
 
-  it("warns when neither resolves, naming the directory it fell back to", () => {
-    const warn = vi.spyOn(console, "warn").mockImplementation(() => {});
-    const cwd = resolveAdapterWorkingDirectory({
+  it("reports the fallback through onLog so it reaches the run transcript", async () => {
+    const onLog = vi.fn(async () => {});
+    const cwd = await resolveAdapterWorkingDirectory({
       adapterType: "codex_local",
+      workspaceCwd: "",
+      configuredCwd: "",
+      onLog,
+    });
+
+    expect(cwd).toBe(process.cwd());
+    expect(onLog).toHaveBeenCalledTimes(1);
+
+    const [stream, chunk] = onLog.mock.calls[0] as unknown as [string, string];
+    expect(stream).toBe("stderr");
+    expect(chunk).toContain("codex_local");
+    expect(chunk).toContain(process.cwd());
+    // The remedy must be actionable, not just a notice that something happened.
+    expect(chunk).toMatch(/project workspace/i);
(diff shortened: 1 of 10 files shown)


martinkrakowski:
Thanks — the transcript point is correct and is now fixed in 9bfdaeb.

`console.warn` writes to the server process stderr. That channel is not persisted into the run transcript, so the operator who needs the message never saw it. The diagnostic did not reach the audience it was written for.

**What changed**

- `resolveAdapterWorkingDirectory` is now `async` and accepts the adapter's `onLog` callback.
- The message goes out with `onLog("stderr", ...)`, matching the `[paperclip] ...` convention already used in these adapter files.
- `console.warn` remains for callers that supply no log channel.
- The eight call sites now `await` the resolver and pass `onLog`. `onLog` is already in scope at each one.

**Test**

The fallback case now asserts the message reaches `onLog` on the `stderr` stream, names the adapter type and the directory, states the remedy, and ends with a newline. A fourth case covers the `console.warn` path.

```
resolve-adapter-cwd.test.ts   4 passed
server-utils.test.ts          92 passed
```

On the authorship note from commitperclip: the `Co-Authored-By: Martin Krakowski <martin@krakowski.cloud>` trailer is in the new commit body.

martinkrakowski:
CI note: the `General tests (server (2/5))` failure is not from this branch. It is the regression tracked in #12122.

Two of the four tests that issue lists are the ones failing here:

```
× terminalizes the route when the cumulative output passes the per-route bound   (timeout 15000ms)
× binds the worker session id one time and ignores a duplicate open reply        (timeout 15000ms)
```

Both live in `server/src/__tests__/plugin-worker-manager.test.ts`. #12122 reports them failing on `master` and reproducing on a clean local checkout, and bisects the cause to `63df7ad2b`.

This branch changes ten files, all under `packages/adapter-utils/` and `packages/adapters/*`. It adds no server code and the failing test imports nothing from those packages.

The `verify` job failure is the aggregate gate, not a separate problem:

```
TYPECHECK_RELEASE_REGISTRY_RESULT: success
GENERAL_TESTS_RESULT: failure
BUILD_RESULT: success
```

Everything else is green: Build, Canary Dry Run, the other four server shards, all workspace shards, Contributor trust, and Socket Security. The tests this PR adds pass, and the existing `server-utils` suite is unchanged at 92 passing.

I have not touched the failing test. Raising its timeout would hide someone else's regression, and #12122 shows a timeout is the wrong diagnosis anyway — one of the four failures is an assertion, `expected [] to include 'echo:browser-code'`, so the route delivers nothing rather than delivering late.

Happy to rebase once #12122 lands.

Refs #12122

PR #12133, opened 2026-08-25
fix(server): preserve committed issue create responses

### LINKED ISSUE #12093: POST /api/companies/{companyId}/issues can 500 after the issue is already durably committed
## Summary

`POST /api/companies/:companyId/issues` (and the sibling `POST /api/issues/:id/children`) can return HTTP 500 `{"error":"Internal server error"}` **after** the issue row has already been durably committed to the database. The caller has no way to tell the create actually succeeded, so a naive retry creates a duplicate issue, and any automation that treats a 500 as "nothing happened" is wrong.

## Root cause (traced in `server/src/routes/issues.ts`, current `master`)

Both routes follow the same shape:

1. `svc.create(companyId, {...})` at `issues.ts:7269` (top-level create) / equivalent in the children route — this durably commits the row.
2. Several **unguarded** post-commit operations then run before the response is sent:
   - `await issueReferencesSvc.syncIssue(issue.id)` (`issues.ts:7283`) — opens its own `db.transaction` and calls `replaceSourceMentions()` twice. Notably **not** wrapped in try/catch, unlike its neighbor two lines below:
   - `await externalObjectsSvc.syncIssueSafely(issue.id)` (`issues.ts:7284`) — this one IS defensively wrapped (`safeSync()` in `external-objects.ts:583-592` catches and logs a warning instead of throwing).
   - Up to three sequential `await logActivity(db, {...})` calls (`issues.ts:7291`, `7326`, `7349`; `logActivity` itself at `activity-log.ts:65`) — each does a DB insert, a PII-redaction pass (`redactCurrentUserValue`), and a `publishLiveEvent()` call, none of which are guarded.
3. Any exception thrown in step 2 (a `replaceSourceMentions` race under concurrent writers, a live-event/pubsub publish hiccup, a redaction edge case on a large/nested `details` payload, etc.) propagates to the shared `errorHandler` middleware (`middleware/error-handler.ts:142-159`), which returns a generic 500 with **no request id, no indication the row already exists, and no way to correlate the failure back to the created issue.**

## Why this matters

This is the identical failure shape as the already-fixed comment-create defect (internally tracked as KEWL-2997/2998: "cross-issue `POST /api/issues/{id}/comments` can return HTTP 500 instead of creating a comment or returning a clear 403/409") — that route received hardening (guaranteed 403/409 on denial, 201+readback on success). The issue-create and children-create routes were never given the same treatment.

## Live confirmation

Observed on our instance 2026-08-24: an agent's `POST /api/companies/{companyId}/issues` call returned HTTP 500, but the target issue (internally KEWL-3940) was durably created — confirmed via `GET /api/issues/{id}` reading back `createdAt: 2026-08-24T15:14:44.699Z`.

## Suggested fix

- Wrap `issueReferencesSvc.syncIssue()` the same way `syncIssueSafely()` already is (log-and-continue on failure), since reference-mention sync is not essential to the create response.
- Either move the `logActivity()` calls to a best-effort/fire-and-forget path (matching the existing `void queueIssueAssignmentWakeup(...)` pattern a few lines later), or wrap them so a logging failure cannot fail the create response.
- More generally: once `svc.create()` has committed, nothing downstream in the same request should be able to turn a successful create into a 500 to the caller. If a post-commit step must remain synchronous and can fail, its failure should be surfaced as a distinguishable 2xx-with-warning rather than a 500, so clients can tell "created, minor side-effect failed" apart from "not created."

## Impact

Any caller that treats a 500 on issue create as "not created" and retries risks creating duplicate issues. Any caller that treats it as fatal and gives up loses attribution/traceability for work that actually landed.

hamedrabah:
## Thinking Path

> - Paperclip is the open source control plane for AI agent companies.
> - Issue creation is a core control-plane mutation and its activity history is a required invariant.
> - The database previously committed an issue before reference sync and activity logging ran.
> - A failure in those later operations could turn a successful create into an HTTP 500 and invite duplicate retries.
> - Treat required audit persistence differently from best-effort post-commit work: persist issue-create activity in the issue transaction, then publish live events after commit.
> - Reference indexing belongs in the same transaction when its resolved context is part of the durable create audit; only live publication remains best-effort after commit.

Fixes: #12093

## What Changed

- Persist top-level and child issue-create activity records in the same database transaction as the issue mutation.
- Roll back issue creation if required activity persistence fails, so a 500 never represents a committed but unaudited issue.
- Index issue references in the create transaction and include resolved reference context in the durable `issue.created` activity.
- Publish activity live events only after the transaction commits.
- Return structured `postCommitWarnings` when live activity publication fails after commit.
- Protect idempotent replay responses from reference-summary failures.
- Add the warning contract to the shared `Issue` type.
- Surface create warnings in every UI flow that calls `issuesApi.create`.
- Add regression coverage for transactional rollback, durable activity reference context, live-publication warnings, and warning rendering.

## Verification

- `node node_modules/typescript/bin/tsc --noEmit -p server/tsconfig.json`
- `node node_modules/typescript/bin/tsc --noEmit -p packages/shared/tsconfig.json`
- `node node_modules/vitest/vitest.mjs run server/src/__tests__/issue-create-post-commit-routes.test.ts server/src/__tests__/issue-create-deduplication-routes.test.ts server/src/__tests__/issue-onboarding-first-task-routes.test.ts server/src/__tests__/issue-assigned-backlog-contract-routes.test.ts ui/src/lib/issue-post-commit-warnings.test.ts`
- `node node_modules/vitest/vitest.mjs run server/src/__tests__/issue-execution-policy-routes.test.ts`
- Focused coverage: 6 files and 40 tests passing.
- `git diff --check`
- Standalone UI typecheck was unavailable in this checkout because its workspace links cannot resolve React and internal packages; GitHub CI performs the configured workspace preflight before typecheck/build.

## Risks

- Required create activity persistence now shares the issue transaction; an activity database failure returns 500 and leaves neither the issue nor its activity behind.
- The response has a new optional `postCommitWarnings` field. This is an additive API change declared in the shared contract.
- Reference indexing now shares the create transaction; a reference-index failure returns 500 and leaves neither the issue nor its activity behind.
- A live-event publication failure does not remove the durable activity record; clients receive a visible warning.
- No database migration is required.

## Model Used

- OpenAI Codex, GPT-5 family. The client does not expose the exact serving model ID or context-window size. The model used high reasoning, tool use, and code execution.

## Checklist


hamedrabah:
@cryppadotta, could you take a final look when convenient? All 28 checks are passing, the post-commit failure paths now preserve a successful create response with explicit warnings, and the review threads are resolved.

Issue #12098, opened 2026-08-24
Human-only follow-up cannot enter todo; blocked-without-blockers retriggers task_watchdog


Wenfeng-GAO:
## Pre-submission checklist

- I have searched existing open and closed issues and this is not a duplicate.
- I am on the latest released version of Paperclip (or can reproduce on `master`).
- I have confirmed the error originates in Paperclip itself — not in my agent adapter, API provider, or local configuration.

## What happened?

A genuine human/external wait (next action is a person, not an agent) cannot be represented as a live path.

When a source-tree child is assigned only to a human (`assigneeUserId` set, `assigneeAgentId` null):

1. `PATCH status=todo` from an agent run returns **409** `Issue follow-up requires an assigned agent`.
2. `PATCH unblockDescriptor.owner.userId=<that human>` returns **403** `Agents may only name themselves as an unblock owner`.
3. Leaving the child `blocked` with an `unblockDescriptor` (agent forced to name **itself**) and **no first-class blockers** is diagnosed as `blocked but has no first-class blocker relations`.
4. `classifyTaskWatchdogSubtree` has **no `waiting` state**. That leaf is a stopped leaf. Task-watchdog retriggers.

The 409/403 pair makes the execution-contract dispositions (`todo` for a human, or `blocked` with a named human unblock owner) unreachable from an agent run. The classifier then treats the only remaining shape as stopped work.

This is distinct from:

- #12096 — watchdog child-create serializes the new sibling behind the in-progress watchdog issue
- #12089 — completing the reusable watchdog wakes the watched parent via `issue_children_completed`
- #10219 — same-run 409 after a source write rotates the stop fingerprint (still happens here: the intended recovery write that changes the fingerprint stale-locks the rest of that watchdog run)
- #11975 — agents cannot name **another agent** as `unblockDescriptor.owner` (this report is the **userId** path: the first check rejects any `"userId" in owner`)

Reproduced on `paperclipai` **2026.817.0**. The same guards are still on `master` (`server/src/routes/issues.ts` contents sha `b5e6a3ad…`, `server/src/services/task-watchdogs.ts` sha `154df0bb…`).

## Expected behavior

- A human-only follow-up (`assigneeUserId` set, `assigneeAgentId` null) must be allowed to enter `todo` when the next action is a person.
- `blocked` + `unblockDescriptor` (or a named human assignee) without first-class blockers must count as a **waiting** disposition, not a stopped leaf.
[...]

PR #12160, opened 2026-08-25
Let agents name the board or a company user as unblock owner

zimmra:
## Thinking Path

> - Issues have a `blocked` status, and `unblockDescriptor` records who must act next and what that action is
> - An agent can find a blocker only a human can clear, for example a missing repository permission
> - `PATCH /api/issues/:id` rejected every non-self owner from an agent actor, so the agent could only name itself
> - Naming itself is false, and the agent will never resolve that action, so the board view becomes misleading
> - This pull request lets an agent name the board or an active company user as the unblock owner
> - Agent owners stay self-only, so an agent still cannot park work on a peer
> - The benefit is that a blocked issue shows the real owner, and the attention list routes it to that human

No public GitHub issue exists for this. The bug is described below.

Refs #11955 — that open PR relaxes the same authorization block for a different case. It lets a child agent name the parent issue assignee, and it keeps the board and user restriction that this PR removes. The two changes are complementary in behavior but they touch the same lines, so whichever lands second needs a rebase.

Refs #10366 — related routing work on unblock owners. It is a different key, `reportsTo` ancestry, and it does not overlap this diff.

**What happened?**

`PATCH /api/issues/:id` returns `403 Agents may only name themselves as an unblock owner` when an agent actor sends `unblockDescriptor.owner` as `"board"` or `{"userId": "..."}`. The request schema accepts both shapes, and the union error message offers `"board"`, but the route authorization layer rejects them for agent actors. An agent that has verified it cannot clear a blocker has only two options. It can name itself, which is untrue and never self-resolves. It can avoid `blocked` altogether. Both give the board a wrong view of the work.

**Expected behavior**

An agent can name the board or an active company user as the unblock owner when the next action belongs to a human. The server keeps its other guarantees. An agent owner must still be the caller and must belong to the issue company. A user owner must still be an active member of the issue company. `unblockDescriptor` still requires `blocked` status.

**Steps to reproduce**

1. Authenticate as an agent with `issue:mutate` on an issue in its own company.
2. Send `PATCH /api/issues/:id` with `{"status":"blocked","unblockDescriptor":{"owner":"board","action":"Merge the approved PR"}}`.
3. The response is `403 Agents may only name themselves as an unblock owner`.
4. Repeat with `"owner": {"userId":"<active member id>"}` and get the same `403`.

`master` at 8d714c2d8.

**Agent adapter(s) involved**

Not adapter-specific (core bug). The gate is in the server route, so every adapter hits it.

**Access context**

Agent API key actor. Board and user actors were never blocked by this gate.

## What Changed

- `server/src/routes/issues.ts`: removed the blanket rejection of `"board"` and `{userId}` owners for agent actors in the `unblockDescriptor` authorization block.
- Kept the agent-owner rules unchanged: the owner agent must belong to the issue company (`422`), and an agent actor must name itself (`403`).
- Kept the user-owner rule unchanged: the owner must be an active company member (`422`).
- `server/src/__tests__/issue-agent-mutation-ownership-routes.test.ts`: replaced the two rejection cases with acceptance cases, and added rejection cases for the rules that must stay.
- Made the test database stub table-aware, because the agent lookup and the membership lookup both select `id` and the previous stub could not tell them apart.

## Verification

Command:

```
npx vitest run server/src/__tests__/issue-agent-mutation-ownership-routes.test.ts
```

Result: `91 passed`.

New and changed cases in that file:

- An agent names `"board"` or `{userId}` when it enters `blocked`. Both return `200`, and the descriptor reaches `issues.update` unchanged.
- An agent changes the owner of an issue that is already `blocked` to `"board"` or `{userId}`. Both return `200`.
- An agent names itself. Returns `200`. This locks the preserved path.
- An agent names a user who is not an active member. Returns `422 Unblock owner user must be an active company member`.
- An agent names a different agent. Returns `403 Agents may only name themselves as an unblock owner`.
- An agent names an agent outside the issue company. Returns `422 Unblock owner agent must belong to the issue company`.

Related suites, to show no regression:

```
npx vitest run server/src/__tests__/routable-blocked.test.ts \
  server/src/__tests__/issue-dependency-wakeups-routes.test.ts \
  packages/shared/src/validators/issue.test.ts
```

Result: `43 passed`.

Typecheck:

```
cd server && npx tsc --noEmit
```

Result: clean.

Two related suites use an embedded Postgres and did not start in my sandbox: `server/src/__tests__/attention-service.test.ts` and `server/src/__tests__/low-trust-red-team-routes.test.ts`. I read both. Each one uses a board actor for its `unblockDescriptor` cases, and a board actor never met the removed gate. CI runs both.

## Risks

Low risk. The change removes one authorization branch and adds no new write path.

- The descriptor is still validated by the same schema, and it still requires `blocked` status.
- Membership and same-company checks are unchanged, so an agent cannot name a user outside the company or an agent in another company.
- `deliverAgentUnblockNotification` already returns early for `"board"` and `{userId}` owners, so no agent wake-up fires for a human owner.
- The attention service already builds a board item from a human-owned descriptor, so the new owners appear in the existing surface with no client change.
- To revert, restore the removed conditional in `server/src/routes/issues.ts` and the two rejection cases in the test file. There is no migration and no stored data shape change.

## Model Used

Claude Opus 5 (`claude-opus-5[1m]`, 1M context), extended thinking, with tool use and code execution.

## Checklist

- [ ] I have updated relevant documentation to reflect my changes — no documentation states this authorization rule. I searched `doc/`, `docs/`, and every tracked `.md`, `.yaml`, and `.json` file for `unblockDescriptor` and found no match. The rule lived only in the route and its tests.

### DIFF
--- a/server/src/__tests__/issue-agent-mutation-ownership-routes.test.ts
+++ b/server/src/__tests__/issue-agent-mutation-ownership-routes.test.ts
@@ -1,4 +1,5 @@
 import { Readable } from "node:stream";
+import { getTableName } from "drizzle-orm";
 import express from "express";
 import request from "supertest";
 import { beforeEach, describe, expect, it, vi } from "vitest";
@@ -291,6 +292,9 @@ function createRunContextDb(
   contextSnapshot: Record<string, unknown> = {},
   runAgentOrRows: string | Record<string, unknown>[] = ownerAgentId,
   runId: string = ownerRunId,
+  // Rows keyed by physical table name, for lookups the generic selection-shape
+  // fallback below cannot tell apart (agents vs company_memberships both select `id`).
+  tableRowOverrides: Record<string, unknown[]> = {},
 ) {
   const runRows = Array.isArray(runAgentOrRows)
     ? runAgentOrRows
@@ -304,7 +308,20 @@ function createRunContextDb(
   const firstRun = runRows[0] ?? {};
   const runAgentId = typeof firstRun.agentId === "string" ? firstRun.agentId : ownerAgentId;
   const runAgentCompanyId = typeof firstRun.agentCompanyId === "string" ? firstRun.agentCompanyId : companyId;
-  const rowsForSelection = async (selection: Record<string, unknown>) => {
+  const overriddenTables = Object.keys(tableRowOverrides);
+  const rowsForTable = (table: unknown) => {
+    if (overriddenTables.length === 0 || !table) return undefined;
+    let tableName: string;
+    try {
+      tableName = getTableName(table as Parameters<typeof getTableName>[0]);
+    } catch {
+      return undefined;
+    }
+    return tableRowOverrides[tableName];
+  };
+  const rowsForSelection = async (selection: Record<string, unknown>, table?: unknown) => {
+    const overrideRows = rowsForTable(table);
+    if (overrideRows) return overrideRows;
     const keys = Object.keys(selection);
     if (keys.includes("entityId")) return [];
     if (keys.includes("contextSnapshot")) return runRows;
@@ -315,16 +332,16 @@ function createRunContextDb(
     }
     return [{ id: runAgentId, companyId: runAgentCompanyId, permissions: {}, role: "engineer", reportsTo: null }];
   };
-  const buildQuery = (selection: Record<string, unknown>) => {
+  const buildQuery = (selection: Record<string, unknown>, table?: unknown) => {
     const whereResult = {
       orderBy: vi.fn(async () => []),
       limit: vi.fn(() => ({
-        then: async (resolve: (limitedRows: unknown[]) => unknown) => resolve(await rowsForSelection(selection)),
+        then: async (resolve: (limitedRows: unknown[]) => unknown) => resolve(await rowsForSelection(selection, table)),
       })),
       for: vi.fn(() => ({
-        then: async (resolve: (selectedRows: unknown[]) => unknown) => resolve(await rowsForSelection(selection)),
+        then: async (resolve: (selectedRows: unknown[]) => unknown) => resolve(await rowsForSelection(selection, table)),
       })),
-      then: async (resolve: (selectedRows: unknown[]) => unknown) => resolve(await rowsForSelection(selection)),
+      then: async (resolve: (selectedRows: unknown[]) => unknown) => resolve(await rowsForSelection(selection, table)),
     };
     const query = {
       innerJoin: vi.fn(() => query),
@@ -335,7 +352,7 @@ function createRunContextDb(
   const dbStub = {
     transaction: async (callback: (tx: typeof dbStub) => Promise<unknown>) => callback(dbStub),
     select: vi.fn((selection: Record<string, unknown> = {}) => ({
-      from: vi.fn(() => buildQuery(selection)),
+      from: vi.fn((table: unknown) => buildQuery(selection, table)),
     })),
     insert: vi.fn(() => ({ values: vi.fn(async () => undefined) })),
   };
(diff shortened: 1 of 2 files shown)


BastitsaB:
Is there a reason this was closed without merging, or was it lost in the review backlog? Asking because the gap it fixes is still open on `master` at `35fca95626a04f5a7ec42cf95989c3d779a1687e` (`server/src/routes/issues.ts:9703, 9712`) and it is costly in practice.

On a self-hosted 14-agent instance the 403 from that guard is the mechanical reason agents create new human-assigned escalation issues instead of parking the existing one: `unblockDescriptor.owner` cannot name a user, and a blocked issue with no agent assignee then also cannot be moved back out of `blocked` by any agent (409, `assertExplicitResumeIntentAllowed`, `routes/issues.ts:5110-5116`). 39 of 146 open issues on that instance were human-assigned follow-ups, most of them created for want of a way to express "waiting on the operator" on an existing issue.

There is also an asymmetry that makes the current state hard to reason about: the owner validation runs only in the PATCH path. `createIssueBaseSchema` accepts `unblockDescriptor`, so an agent can *create* an issue with `status: "blocked"` and `owner: {userId}` while being unable to *set* the same value on an existing one.

Related: #12098, #12468, #11975. Happy to test a rebased version against a live instance if that would help move it.

digital-efficacy:
Independent corroboration for the reopen question above, from a different self-hosted instance, with the state re-verified against current `master` rather than quoted from the earlier thread.

**The gap is still open.** At `master@20968040` (2026-09-01T04:32Z) the guard is `server/src/routes/issues.ts:10227-10228`:

```ts
const owner = descriptor.owner;
if (req.actor.type === "agent" && (owner === "board" || "userId" in owner)) {
  throw forbidden("Agents may only name themselves as an unblock owner");
}
```

(The line numbers cited above — 9703/9712 — have shifted as the file grew; the guard itself is unchanged. Second occurrence for the agent-owner case is now `:10236-10237`.)

**The part that I think strengthens the case for this PR:** the server already contains the complete validation for a user owner, and cannot reach it. Immediately below the guard, `:10239-10247`:

```ts
} else if (owner !== "board" && "userId" in owner) {
  const member = await db.select({ id: companyMemberships.id }).from(companyMemberships).where(and(
    eq(companyMemberships.companyId, existing.companyId),
    eq(companyMemberships.principalType, "user"),
    eq(companyMemberships.principalId, owner.userId),
    eq(companyMemberships.status, "active"),
  )).limit(1).then((rows) => rows[0] ?? null);
  if (!member) throw unprocessable("Unblock owner user must be an active company member");
}
```

That `else if` is dead code for every agent actor — line 10227 throws before it can run. It is only ever reachable for board and user actors. So the active-company-membership check that would make a human unblock owner safe is written, tested by its own error path, and unreachable from the one actor type that needs it.

Which is why the diff here is `+3/-3` in the route: it doesn't introduce a new trust path, it lets an existing one be reached. That seems like an unusually low-risk shape for a change that was closed unmerged.

**The consequence we measure, anonymised.** On our instance the same 403 is the mechanical reason an agent cannot record that a *human* owns the next step. An agent can register that it will unblock something itself, but not that a person must — that information can only go into free-text in a comment. Two downstream effects:

- Work waiting on a person carries no machine-readable owner, so "what is queued on a human right now" is a hand-counted list rather than a query. Ours has been in the dozens for months, and the count is only as good as whoever last read the comments.
- Because the blocked state cannot name the person, agents open a *new* human-assigned issue to represent the wait instead of parking the existing one — which is the duplicate-escalation pattern already described above, reproduced independently here.

**The ask, and a genuine alternative.** Either reopen this (it still applies cleanly in shape, and CI was green when Greptile summarised it on 25-08), or — if the blanket denial is deliberate and the intended design is that only board actors may name humans — it would help a lot to say so in the thread. That is a legitimate answer and we would build around it; the current state is costly mainly because it reads as an oversight and every operator rediscovers it the same way.

Related open threads for whoever picks this up, so it isn't triaged as a duplicate: **#12468** (the schema advertises an `agentId` owner but the server 403s any agent except the caller's own — same guard, second occurrence at `:10236`), and **#11955** (widens the same guard to a parent assignee, i.e. to another *agent* rather than to a human, so it does not subsume this).

PR #12212, opened 2026-08-26
feat(agents): REST route to grant/revoke scoped permissionKeys on agent principals

dzianisv:
## Thinking Path

> - The access-control subsystem models permissions per "principal" (user or agent) via `access.setPrincipalPermission(companyId, principalType, principalId, permissionKey, enabled, grantedByUserId, scope)`, which accepts any of the shared `PERMISSION_KEYS` values.
> - That function has no generic REST route for the `agent` principal type. The only two HTTP call sites in `server/src/routes/agents.ts` (`POST /companies/:companyId/agents` and `PATCH /agents/:id/permissions`) hardcode `permissionKey: "tasks:assign"`.
> - Human company members already have a generic grant route, `PATCH /companies/:companyId/members/:memberId/permissions`, that accepts an arbitrary `permissionKey`/`scope` pair. Agent principals have no equivalent, so there is no way to grant an agent a scoped permission like `agents:configure` or `agents:suggest-changes` after it is created.
> - This pull request adds a small, additive `GET`/`POST /api/agents/:id/grants` pair that exposes `access.setPrincipalPermission`/`access.listPrincipalGrants` for agent principals, gated the same way the equivalent human-member route is gated (by the permission being managed, not by the agent-lifecycle permission).
> - The benefit is that board operators (or a CEO agent) can now grant a specific agent a scoped capability without a database-level workaround, closing a real gap between the user- and agent-principal grant surfaces.

No existing GitHub issue tracks this gap. Inline description below, following `.github/ISSUE_TEMPLATE/feature_request.yml`:

**Subsystem affected**

server/ — REST API & orchestration services

**Problem or motivation**

`access.setPrincipalPermission()` supports any `PermissionKey`, but there is no REST route that can grant/revoke an arbitrary `permissionKey` on an agent principal after creation. The only two call sites (`POST /companies/:companyId/agents` and `PATCH /agents/:id/permissions`) hardcode `"tasks:assign"`. Human company members already have this via `PATCH /companies/:companyId/members/:memberId/permissions`; there is no agent-principal equivalent.

**Proposed solution**

Add `GET /api/agents/:id/grants` (list current scoped grants for an agent) and `POST /api/agents/:id/grants` (grant or revoke one `permissionKey`, validated against the shared `PERMISSION_KEYS` enum, via `access.setPrincipalPermission`). Gate both the same way the existing `PATCH /agents/:id/permissions` route is gated (CEO agent actor, or board actor), but check the `agents:configure` decision rather than `agents:create` — the permission being managed gates the route, mirroring `users:manage_permissions` on the member-permissions route.

**Alternatives considered**

Extending `PATCH /agents/:id/permissions` (or a new full-replace `grants` array route mirroring `PATCH /companies/:companyId/members/:memberId/role-and-grants`) was considered and rejected: a full-replace design risks silently clobbering `tasks:assign` (managed independently by the existing `/permissions` route) if a caller's grants array omits it. The additive single-key route matches `setPrincipalPermission`'s own signature and avoids that hazard.

**Roadmap alignment**

Not tied to a specific `ROADMAP.md` line item; this closes a general access-control gap between the user- and agent-principal grant surfaces.

## What Changed

- `packages/shared/src/validators/agent.ts` (+ `validators/index.ts`, package `src/index.ts` re-exports): new `updateAgentGrantSchema` = `{ permissionKey: enum(PERMISSION_KEYS), enabled: boolean, scope?: Record<string, unknown> | null }`.
- `server/src/routes/agents.ts`:
  - New `assertBoardCanConfigureAgentsForCompany` gate helper, checking the `agents:configure` decision (distinct from the existing `agents:create`-gated `assertBoardCanManageAgentsForCompany`).
  - `GET /api/agents/:id/grants` — lists an agent principal's current scoped grants via `access.listPrincipalGrants`. Board callers need `agents:configure`; a CEO agent, or the target agent itself, may also view its own grants.
  - `POST /api/agents/:id/grants` — grants/revokes a single `permissionKey` via `access.setPrincipalPermission`, gated like `PATCH /agents/:id/permissions` (CEO agent actor, or board actor via `agents:configure`).
- `server/src/routes/openapi.ts` — OpenAPI registrations for both new routes.
- `server/src/__tests__/agent-permissions-routes.test.ts` — 7 new tests under `describe("agent scoped grants")`: board-without-permission rejection, non-CEO-agent rejection, successful grant by an authorized board actor, successful revoke by a CEO agent on a peer agent, invalid `permissionKey` validation (400), and both list-grants call paths (board with `agents:configure`, agent viewing its own grants).

## Verification

- `pnpm --filter @paperclipai/shared build` — clean.
- `pnpm --filter @paperclipai/server exec tsc --noEmit -p .` — clean; confirmed this diff introduces zero new errors (a small set of unrelated pre-existing errors exist on `master` independent of this change, e.g. `plugin-worker-manager.ts` `hostRouteId`, reproduced with a clean `pnpm install` both before and after this diff).
- `pnpm exec vitest run src/__tests__/agent-permissions-routes.test.ts` — 61 passed (54 pre-existing + 7 new).
- `pnpm exec vitest run src/__tests__/agent-permissions-routes.test.ts src/__tests__/openapi-routes.test.ts` — 66 passed, including the OpenAPI route-coverage check.

## Risks

- New surface area for privilege escalation if the `agents:configure` gate is misconfigured for a company's role defaults — mitigated by reusing the exact same `access.decide()` call pattern as every other agent-management route in this file, and by calling `ensureMembership` before every grant (same precondition as the existing `tasks:assign` path).
- The additive (single-key) design does not atomically replace a full grant set in one call; a caller wanting to set several keys issues one `POST` per key. Given the very small number of keys this needs to cover, this was judged an acceptable trade against the risk of accidentally clobbering `tasks:assign` in a full-replace design.

## Model Used

Claude Sonnet, via GitHub Copilot / the `pi` coding agent harness (`github-copilot/claude-sonnet-5`), extended reasoning enabled, tool use for repo search/edit/build/test.

## Checklist

- [x] I have searched GitHub for duplicate or related PRs and linked them above (none found covering agent-principal scoped grants)

## Not verified

- No live DB/integration run against a real Postgres instance; coverage is at the mocked-service route-test layer, consistent with this router's existing test file.
- `packages/paperclip-runner`'s Rust binary build was not exercised (no `cargo` toolchain available); only its TypeScript surface was rebuilt and typechecked.


dzianisv:
Addressed: pushed 6639768 removing the `scope` field from `updateAgentGrantSchema` entirely. This route now always calls `setPrincipalPermission(..., scope: null)` — never persists a caller-supplied scope object, closing the unrestricted-grant-widening path `scopeAllows()` would otherwise allow on an unrecognized scope key. Added a regression test asserting a scope object in the request body is ignored.

dzianisv:
Addressed: pushed 61de443. `updateAgentGrantSchema` is now `.strict()` — a request that includes `scope` (or any other unknown field) is rejected with 400 instead of silently stripped and persisted as an unrestricted grant. Updated the regression test to assert the fail-closed 400, and added a companion test for the normal no-scope grant path. 68/68 route+openapi tests pass.

dzianisv:
Note: the `Verify serialized server suites (3/5)` failure on the latest run is in `src/__tests__/costs-service.test.ts` ("cost routes > returns finance summary rows for valid requests" — a response-shape mismatch on `companyId`), which this PR does not touch. Confirmed it passes locally in isolation (17/17) on this branch. Likely an unrelated pre-existing flake/ordering issue in that shard; I don't have rerun rights on this repo (non-admin contributor) to confirm with a clean re-run.

zannis:
The `agents:configure` gate this route is named for only covers the board branch. In the `POST /agents/:id/grants` agent branch, an actor passes on `actorAgent.role === "ceo"` alone — there is no `access.decide` call on that path before the write:

```ts
if (req.actor.type === "agent") {
  // ...same-company check...
  if (actorAgent.role !== "ceo") {
    res.status(403).json({ error: "Only CEO can manage agent grants" });
    return;
  }
  // falls through to setPrincipalPermission with no authorization decision
} else {
  await assertBoardCanConfigureAgentsForCompany(req, existing.companyId);
}
```

Since `permissionKey` is `z.enum(PERMISSION_KEYS)`, a CEO-role agent can grant itself or any other agent in the company any of the 21 keys — including `agents:configure`, `tools:admin`, `users:manage_permissions`, `environments:manage` and `joins:approve`. `:id` is unconstrained, so self-grant is reachable.

This isn't only theoretical for CEO-role agents that already hold `agents:configure` anyway. `ensureRootAgentDefaultChangeGrants` is the only thing that ever grants it, and it bails before granting anything unless the company has exactly one root CEO:

```ts
if (rootCeoRows.length !== 1) return 0;
return ensureAgentDefaultGrants(companyId, rootCeoRows[0]!.id, ROOT_AGENT_DEFAULT_CHANGE_GRANTS);
```

So a CEO-role agent with a `reportsTo` set, or any CEO-role agent in a company that has two or more root CEOs, provably does **not** hold `agents:configure` — and still passes this route's role check. Role and permission genuinely diverge here.

The role-only shape is inherited from `PATCH /agents/:id/permissions`, where it's bounded: that route writes exactly one fixed key, `tasks:assign`. Carried onto a route where the caller picks the key, the same check becomes a general escalation primitive.

One note on the stated rationale — the human-member route this mirrors doesn't have an agent branch to copy. `PATCH /companies/:companyId/members/:memberId/permissions` runs `assertCompanyPermission(req, companyId, "users:manage_permissions")` for *every* caller, and `getProtectedMemberReason` then refuses any non-user principal and any non-board actor outright. The mirror holds for the board branch; there's no precedent in it for the agent short-circuit.

Smallest fix is to run the same decision for agent actors instead of short-circuiting on the role:

```ts
if (actorAgent.role !== "ceo") {
  res.status(403).json({ error: "Only CEO can manage agent grants" });
  return;
}
const decision = await access.decide({
  actor: req.actor,
  action: "agents:configure",
  resource: { type: "company", companyId: existing.companyId },
});
if (!decision.allowed) {
  throw forbidden(decision.explanation, authorizationDeniedDetails(decision));
}
```

Both helpers are already imported in this file for the board path. Keeping the role check in front of it preserves the current 403 message for non-CEO agents and makes the agent branch enforce the permission the route is gated on.

(Unrelated to the above: the branch currently conflicts with `master`, so it'll need a rebase before the shards mean much.)

PR #12291, opened 2026-08-27
Remove the company brand color and per-company attachment limit

devinfoley:
## Thinking Path

> - A company is the top-level container, and the company General page holds its settings
> - Two of those settings did almost nothing: the brand color only tinted the generated company icon, and the attachment size limit sat under the deployment-level `PAPERCLIP_ATTACHMENT_MAX_BYTES` cap that already bounded every upload
> - A setting that changes one icon hue, and a setting that can only lower a limit the operator already set, are not worth the page space or the code that carries them
> - This pull request deletes both settings from the UI, the validators, the API contract, the server, and the database
> - With the deployment cap as the only limit left, the message a person sees when an upload is rejected has to name that limit in terms they can act on, so the raw byte count becomes a human-readable size
> - The benefit is a shorter company General page for every deployment, one attachment limit instead of two, and less code between an upload and its ceiling

No existing issue. The description below follows `.github/ISSUE_TEMPLATE/enhancement.yml`.

The company General page (`/company/settings`), the `PATCH /api/companies/{companyId}` and `PATCH /api/companies/{companyId}/branding` request contracts, and the attachment upload limit on task, case, and company-import uploads.

**Subsystem affected**

Cross-cutting: `ui/`, `server/`, `packages/shared`, `packages/db`.

**Current behavior**

The company General page shows an "Appearance" section with three controls: Logo, Brand color, and Attachment size limit. The brand color is a hex value that feeds one thing — the hue of the generated company pattern icon. Companies that never set one already get a hue derived from the company name. The attachment size limit is a per-company byte count stored on `companies.attachment_max_bytes`. Every upload path clamps it against the deployment-level `PAPERCLIP_ATTACHMENT_MAX_BYTES` cap, so the per-company value can only lower a limit the operator already chose.

**Proposed behavior**

The Appearance section keeps the Logo control only. The company pattern icon always derives its hue from the company name. Every attachment path reads the deployment cap directly, so `PAPERCLIP_ATTACHMENT_MAX_BYTES` is the single limit. An upload rejected by that limit says so in human units — "File is larger than the 10 MB limit" rather than a raw byte count. The `companies.brand_color` and `companies.attachment_max_bytes` columns are dropped, and both fields leave the company API contract.

**Reason and benefit**

Both settings ask an operator to make a decision that changes almost nothing. The brand color moves one icon hue on a page that also lets you upload a real logo, which overrides the icon entirely. The attachment limit reads as a real control but cannot raise anything, so it is a second place to look when an upload is rejected. Removing both shortens the page every deployment sees, removes a company-scoped read from the task attachment upload path, and leaves one attachment limit to reason about instead of two.

**Breaking changes**

The company API responses no longer include `brandColor` or `attachmentMaxBytes`, and `GET /api/invites/{token}` no longer includes `companyBrandColor`. `PATCH /api/companies/{companyId}/branding` is strict, so a request that sends `brandColor` now returns 400; the non-strict `PATCH /api/companies/{companyId}` schema strips it. Company packages exported by older versions still import: the portability company manifest schema is non-strict, so the retired keys are stripped and ignored rather than rejected. Companies that stored a brand color lose it — their icon reverts to the name-derived hue that every company without a color already used.

## What Changed

- Removed the "Brand color" and "Attachment size limit" fields from the company General page, along with their state, dirty checks, save payload, and Save-button gating.
- Removed `brandColor` and `attachmentMaxBytes` from `createCompanySchema`, `updateCompanySchema`, and `updateCompanyBrandingSchema`, and deleted the now-orphaned `DEFAULT_COMPANY_ATTACHMENT_MAX_BYTES` and `MAX_COMPANY_ATTACHMENT_MAX_BYTES` constants.
- Removed both fields from the `Company` type, the portability manifest type and schema, and the `companiesApi.update` payload allowlist.
- Dropped `brandColor` from `CompanyPatternIcon` and its callers, so the icon hue always comes from the company name. Deleted the now-unused `hexToHue` helper and the now-unused `pickTextColorForSolidBg` export.
- Stopped emitting `brandColor` from the company service selection and from the invite-summary and invite-branding payloads in `server/src/routes/access.ts`.
- Replaced `normalizeIssueAttachmentMaxBytes` with the deployment cap: task attachments, case attachments, and company import now use `MAX_ATTACHMENT_BYTES` directly. The helper is deleted.
- Added `formatAttachmentSize()` next to `MAX_ATTACHMENT_BYTES` and routed every over-limit message through it, so a rejected upload names the limit in human units instead of raw bytes: `Image exceeds 10485760 bytes` becomes `Image is larger than the 10 MB limit`. Enforcement is unchanged — the same single cap, the same multer limits, the same status codes and response shapes.
- Added migration `0229_drop_company_brand_color_and_attachment_max_bytes.sql` and removed both columns from the Drizzle `companies` schema.
- Kept legacy imports working: the portability company manifest schema is non-strict, so older packages carrying the retired keys still import with the keys ignored.
- Updated the skill API reference and the implementation spec, and pruned the token-extraction allowlist entries that the removed code made stale.

## Verification

Commands run from the repository root:

- `pnpm --filter @paperclipai/shared typecheck` — pass
- `pnpm --filter @paperclipai/db typecheck` — pass (includes `check:migrations`, which validates the new migration number and journal entry)
- `pnpm --filter @paperclipai/ui typecheck` — pass
- server typecheck via `node_modules/.bin/tsc --noEmit` in `server/` — pass. `pnpm --filter @paperclipai/server typecheck` could not run locally because it builds the Rust runner first and `cargo` is not installed on this machine; the TypeScript step it wraps is the command above.
- `npx vitest run packages/shared/src/validators/company.test.ts` — 6 passed
- `npx vitest run server/src/__tests__/company-portability.test.ts` — 90 passed
- `npx vitest run server/src/__tests__/attachment-types.test.ts server/src/__tests__/assets.test.ts server/src/__tests__/issue-attachment-routes.test.ts server/src/__tests__/company-portability.test.ts server/src/__tests__/cases-routes.test.ts` — 165 passed (the human-readable limit messages)
- `npx vitest run server/src/__tests__/company-branding-route.test.ts server/src/__tests__/issue-attachment-routes.test.ts server/src/__tests__/invite-summary-route.test.ts server/src/__tests__/openclaw-invite-prompt-route.test.ts server/src/__tests__/companies-route-cross-company-authz.test.ts` — all passed
- `npx vitest run cli/src/__tests__/company.test.ts cli/src/__tests__/company-delete.test.ts` — 27 passed
- `npx vitest run` in `ui/` — 4425 passed, 1 pre-existing failure unrelated to this change (`OnboardingWizard.test.tsx` "renders instead of throwing when the browser denies storage access", which also fails on `master`)
- `npx vitest run` in `server/` — see the note below
- `node scripts/check-token-gates.mjs` — no new violations; the only reported violations are the pre-existing `PillGuy.tsx` ones present on `master`

New tests added:

- `packages/shared/src/validators/company.test.ts` — the create and update schemas strip the retired keys, the strict branding schema rejects `brandColor`, and the portability manifest schema accepts a legacy entry carrying both keys and drops them.
- `server/src/__tests__/company-branding-route.test.ts` — `PATCH /api/companies/{companyId}/branding` returns 400 for `brandColor` and does not call the company service.
- `server/src/__tests__/company-portability.test.ts` — a legacy package that declares `brandColor` and `attachmentMaxBytes` imports successfully, and neither key reaches `companies.create`.
- `server/src/__tests__/issue-attachment-routes.test.ts` — the effective task attachment limit is the deployment cap, and the route no longer loads the company to size an upload.
- `server/src/__tests__/attachment-types.test.ts` — `formatAttachmentSize()` renders the default cap as `10 MB`, keeps one decimal place for fractional sizes and drops a trailing `.0`, falls back to KB and bytes for small caps, steps up to GB, and never emits `NaN` for a degenerate input.
- `server/src/__tests__/assets.test.ts` — the asset-image and company-logo routes both return the human-readable limit message on an over-cap upload.

## Merge with master

`master` moved while this was open, and the merge needed two resolutions:

- **`ui/src/pages/CompanySettings.tsx`.** #12243 reworded the user-facing
  copy from "company" to "organization", and that rewording landed inside
  the "Brand color" and "Attachment size limit" hints — the two fields
  this change deletes. Both fields are removed, so the conflicted block is
  dropped whole. The Logo field and every other copy change from #12243 are
  kept.
- **Migration renumbered 0228 -> 0229.** #12307 landed
  `0228_nasty_grim_reaper`, so this migration is now
  `0229_drop_company_brand_color_and_attachment_max_bytes`. Its snapshot is
  rebuilt from master's `0228_snapshot.json` with only the two `companies`
  columns removed, and `meta/_journal.json` is master's journal plus a
  single `idx: 229` entry. `pnpm --filter @paperclipai/db check:migrations`
  passes.

The snapshot was rebuilt by hand rather than taken from `drizzle-kit
generate`, because master's `0228_snapshot.json` has drifted from master's
own schema: `issue_question_response_deliveries.error_count` is created by
master's 0228 SQL but missing from its snapshot, and the snapshot still
carries `decision_archive_notification_outbox.error_count`. Regenerating
folds both into this migration, and the resulting `ADD COLUMN error_count`
would fail on a fresh database where master's 0228 already created that
column. Rebuilding from master's snapshot leaves that drift exactly where
it is and keeps this migration to the two column drops. The drift is
pre-existing on master and is not addressed here.

## Risks

- **The migration is a destructive column drop.** `0229_drop_company_brand_color_and_attachment_max_bytes.sql` removes `companies.brand_color` and `companies.attachment_max_bytes`. It is safe because both features are removed in the same change and nothing reads either column after it. The statements use `DROP COLUMN IF EXISTS`, matching the convention of the recent drop migrations in this repository. The drop is not reversible: a downgrade after this migration loses any stored values.
- **Stored brand colors are lost.** A company that had set a color now renders the name-derived icon hue that every company without a color already used. No other surface changes, and an uploaded logo still overrides the icon.
- **API response shape narrows.** `brandColor` and `attachmentMaxBytes` leave the company payloads, and `companyBrandColor` leaves the invite summary payload. A client reading those fields now sees `undefined`. The bundled UI and CLI are updated in this change.
- **Legacy imports are covered.** Packages exported by older versions still carry both keys. The manifest schema is non-strict, so the keys are stripped rather than rejected, and a test locks that in.
- **The over-limit message strings changed.** Anything matching on the old `... exceeds N bytes` text — a test, a script, or a client that string-matches `body.error` — needs updating. The status codes (422) and response shapes are unchanged, so structured clients are unaffected.
- **Attachment limits can only widen.** A deployment that had lowered a company below the deployment cap now allows uploads up to the cap for that company. Lower `PAPERCLIP_ATTACHMENT_MAX_BYTES` if a smaller ceiling is needed.
- **Storybook visual baselines shift** for the `CompanyPatternIcon` matrix story, because those fixtures had brand colors. That workflow runs only on a PR labeled `storybook-visual`, so it does not gate this PR; regenerate the baselines if the label is added.

## Model Used

Claude (Anthropic), Claude Opus, agentic tool use via Claude Code.

## Checklist


devinfoley:
@greptile review

Merged `master` and resolved two collisions:

- `ui/src/pages/CompanySettings.tsx` — #12243 reworded the settings copy to "organization", and that rewording sat inside the "Brand color" and "Attachment size limit" hints. Both fields are deleted here, so the conflicted block was dropped whole; the Logo field and the rest of #12243's copy changes are kept.
- Migration renumbered `0228` -> `0229`, since #12307 landed `0228_nasty_grim_reaper`. The `0229` snapshot is rebuilt from master's `0228_snapshot.json` with only the two `companies` columns removed, and the journal is master's plus one `idx: 229` entry.

Worth a look: the snapshot is rebuilt by hand rather than regenerated. Master's `0228_snapshot.json` has drifted from master's own schema — `issue_question_response_deliveries.error_count` is created by master's 0228 SQL but absent from its snapshot, and the snapshot still carries `decision_archive_notification_outbox.error_count`. `drizzle-kit generate` folds both into this migration, and the resulting `ADD COLUMN error_count` would fail on a fresh database where master's 0228 already created it. Rebuilding from master's snapshot leaves that pre-existing drift untouched and keeps this migration to the two column drops.

PR #12298, opened 2026-08-27
fix(heartbeat): bound stdout/stderr in persisted result_json

Oldrich333:
## Problem

`heartbeat_runs.result_json` stores the full adapter `stdout`/`stderr` verbatim. On one production instance that table reached **7.9 GB, of which `result_json` alone is 7.0 GB** — 24,657 rows averaging 291 KB, largest 10.9 MB. Every hourly database backup copies all of it.

It is already stored twice: 18,539 of those rows carry `log_store=local_file` with `log_ref`/`log_bytes`/`log_sha256` pointing at the complete log on disk (6.4 GB), and the row still keeps the whole stream inline.

## Why this is a bug rather than a tuning question

The intended bound already exists on the read side. `heartbeatRunSafeResultJsonColumn` projects only the first `HEARTBEAT_RUN_RESULT_OUTPUT_MAX_CHARS` of `stdout`/`stderr` once a row passes `HEARTBEAT_RUN_SAFE_RESULT_JSON_MAX_BYTES`. So everything beyond that cap is written, backed up, and never read back.

## Change

Apply the same bound at write time, at the single point where every adapter's result is persisted — not in each of the twelve adapter `execute.ts` files, where a new adapter would silently regress it.

- keeps the **tail**, not the head: a failing run explains itself at the end
- returns the original object unchanged when nothing exceeds the cap (no allocation on the common path)
- records `stdoutTruncated` / `stdoutFullChars` so the truncation is visible rather than silent
- no new dependencies; the module stays free of Node globals

## Tests

Three cases added to `heartbeat-run-summary.test.ts` (truncation + tail + markers, no-op identity, non-string/absent streams). Full file: 11 passed.

### DIFF
--- a/server/src/__tests__/heartbeat-run-summary.test.ts
+++ b/server/src/__tests__/heartbeat-run-summary.test.ts
@@ -3,6 +3,8 @@ import {
   summarizeHeartbeatRunResultJson,
   buildHeartbeatRunIssueComment,
   mergeHeartbeatRunResultJson,
+  boundHeartbeatRunResultJson,
+  HEARTBEAT_RUN_RESULT_OUTPUT_MAX_CHARS,
 } from "../services/heartbeat-run-summary.js";
 
 describe("summarizeHeartbeatRunResultJson", () => {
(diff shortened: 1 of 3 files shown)


johnnysunyi-bit:
Independent confirmation from another production instance (v2026.817.0), with a per-adapter breakdown that may help scope the fix:

- `heartbeat_runs` = **2,272 MB** (main heap 40 MB, TOAST 2,208 MB). `result_json` alone was 1,742 MB of raw text, and the `stdout` key was **88 %** of it (August: 3,384 / 3,851 MB). 4,774 rows (20 %) held 82 % of the bytes; the largest row was 4.6 MB (= `MAX_CAPTURE_BYTES` 4 MB + stderr + JSON escaping).
- The same output already existed on disk as `log_store=local_file` ndjson (redacted, uncapped — a superset). The DB copy was both redundant and the only **unredacted** copy.
- Which adapters write it (weekly counts of rows >100 KB, same fleet):

| adapter | big rows / week | bytes |
|---|---|---|
| `cursor` | 1,734 | 408 MB |
| `codex_local` | 759 | 307 MB |
| `claude_local` (ACP engine) | 0–37 | ≤5.8 MB |
| `grok_local` | 0 | 0 |

  Code paths: `codex_local` / `cursor` write it on the un-parsed branch. **`pi_local` (`execute.ts:788`) and `opencode_local` (`execute.ts:683`) store `{stdout, stderr}` unconditionally, even on success**, so they will be worse than the numbers above once used.
- `HEARTBEAT_RUN_SAFE_RESULT_JSON_MAX_BYTES` / `HEARTBEAT_RUN_RESULT_OUTPUT_MAX_CHARS` only apply to the SELECT projection, so upgrading did not help; the write path is the only place to fix this. Until then we strip `stdout`/`stderr` from terminal-state rows older than 48 h with a cron (`result_json - 'stdout' - 'stderr'`). That took the table from 2,272 MB to 511 MB after `VACUUM FULL`. Nothing in the UI reads those keys; the only reader is `run-liveness.ts` `rawSources()`, which already falls back to `stdout_excerpt` / `error`.

+1 for merging. Happy to test a build.

Issue #12118, opened 2026-08-24
Timer-heartbeat runs get 403 cross_issue_influence_run_context_required on every issue write, including their own issue — stated remedy is unachievable


markymark1969:
## Summary

An agent woken by the **heartbeat timer** cannot comment on or PATCH **any** issue — including issues it is itself the assignee of. Every such write returns `403 cross_issue_influence_run_context_required`.

The 403's stated remedy ("send the `X-Paperclip-Run-Id` header with your current run and retry") is unachievable in exactly the state that triggers it: the header is already correct and the run is already valid. An agent that follows the error text can only retry until it hits its cap.

Verified against `@paperclipai/server@2026.817.0`.

## Root cause

`server/dist/services/cross-issue-influence-limit.js`, in `observeCrossIssueInfluence()`:

```js
const sourceIssueId = readRunSourceIssueId(run.contextSnapshot);
if (!sourceIssueId)
    throw crossIssueInfluenceRunContextError();   // fires for every timer heartbeat
if (sourceIssueId === input.targetIssueId ||      // same-issue exemption
    (input.targetIssueIdentifier && sourceIssueId.toUpperCase() === input.targetIssueIdentifier.toUpperCase()))
    return null;                                  // ...unreachable when sourceIssueId is null
```

`readRunSourceIssueId()` reads only `contextSnapshot.issueId` / `.taskId`. A timer-woken run has neither, so the function throws **before** reaching the same-issue exemption that would otherwise let an agent write to its own task.

The guard is applied at `server/dist/routes/issues.js` lines 6981 and 6984 (inside `PATCH /issues/:id`) and 9210 (inside `POST /issues/:id/comments`). There is no per-route escape.

`POST /issues/{id}/checkout` cannot clear it: checkout writes `checkoutRunId` / `executionRunId` on the **issues** table; the gate reads `context_snapshot` on the **heartbeat_runs** table. The two never meet.

## Evidence

Measured on one instance's `heartbeat_runs`, grouped by wake reason, trailing 14 days (n=1779). "Unscoped" = `context_snapshot` has neither `issueId` nor `taskId`:

| wake reason | runs | unscoped |
|---|---|---|
| `issue_assigned` | 773 | 0 |
| `heartbeat_timer` | **516** | **516** |
| `source_scoped_recovery_action` | 155 | 0 |
| `finish_successful_run_handoff` | 119 | 0 |
| `issue_continuation_needed` | 67 | 0 |
[...]

Issue #11852, opened 2026-08-21
Fix inconsistent heartbeat run-context validation across checkout and issue writes


aduduman031084:
## Summary

In Paperclip `2026.817.0`, a persisted heartbeat run without `contextSnapshot.issueId`/`taskId` can pass `POST /api/issues/:id/checkout` and write checkout/execution pointers, but the same run is later rejected by comment/PATCH paths with `403 cross_issue_influence_run_context_required`.

This is an authorization-path inconsistency: a header is not enough to establish authority, but checkout currently accepts it before the canonical persisted-run lookup used by `services/cross-issue-influence-limit.ts`.

## Reproduction

1. Use an agent-authenticated heartbeat whose persisted record is missing its source issue context (the observed early signal was a scratch directory name containing `unassigned`).
2. Send `X-Paperclip-Run-Id` on `POST /api/issues/:id/checkout` for the agent's issue: checkout succeeds and records lock fields.
3. Send the same header/run id to `POST /comments` or `PATCH /issues/:id`: the cap guard performs its persisted lookup, cannot derive a source issue, and returns 403 before mutation.

Unknown, wrong-agent/company, malformed, and terminal run IDs must remain rejected without mutation.

## Expected

All mutating issue endpoints resolve the same canonical, persisted run context before mutation. An unassigned/stale/forged handle must fail before checkout can create a partial state; cross-issue behavior must remain deny-by-default and capped.

## Proposed fix

- Extract a shared locked validator in `server/src/services/cross-issue-influence-limit.ts` that requires: UUID, matching company/agent, `running` status, and non-empty persisted source issue.
- Make checkout and release preflight that validator and require canonical source issue == target issue.
- Keep comments/PATCH on the existing observer so legitimate cross-issue actions remain subject to the current counter/authorization policy.
- Add focused regressions for canonical same-issue success, and unknown/unassigned/terminal/forged/cross-target failures with no issue or activity mutation.

This is intentionally not a fallback that trusts the header, `executionRunId`, or a prior successful checkout.

dzianisv:
Correcting the "Proposed fix" direction above — it does not actually restore the documented heartbeat flow, and a fourth report (AGE-2030, our internal tracker) independently confirms the same root cause with a compatible fix.

[...]

PR #12315, opened 2026-08-27
fix(server): bind cross-issue influence run context on first write instead of refusing

dzianisv:
## Thinking Path

> - Cross-issue influence containment refuses a heartbeat run's issue write when the run has no recorded "source issue" in `heartbeat_runs.contextSnapshot`
> - The documented normal heartbeat loop (inbox-lite discovery -> pick a todo/in_progress item -> `POST /issues/:id/checkout`) starts every legitimate run **unscoped** — `contextSnapshot.issueId` is null until the agent picks work at runtime
> - `POST /issues/:id/checkout` never patches the *calling* actor's own already-existing run row with the checked-out issue, so that run's `sourceIssueId` stays null for its entire life
> - `observeCrossIssueInfluence` throws unconditionally on a null source issue, so every comment/PATCH that run ever attempts is refused with 403 — including on the issue it just checked out
> - The correct fix treats a run's first write as legitimately defining its "home" issue (bind-on-first-write) instead of refusing outright — a run with no prior source issue cannot have had a prior cross-issue write to compare against

Fixes the root cause described in #11852 (see the corrected-design comment on that issue, which supersedes its own original "Proposed fix" section — that original proposal would have made checkout itself refuse every unscoped-wake run, which is strictly worse). Also resolves the same root cause reported by duplicates #11975 and #12118.

**What happened?**

`server/src/services/cross-issue-influence-limit.ts` `observeCrossIssueInfluence()`:
```ts
```
`contextSnapshot.issueId` is only ever set at run creation, or by `heartbeat.wakeup()` for a *spawned* run belonging to a *different* actor (`server/src/routes/issues.ts`, `issue.checkout` wake source). `POST /issues/:id/checkout` never patches the **calling** actor's own already-existing run row. Any run that starts from a generic/unscoped heartbeat wake (the documented normal flow) therefore has `sourceIssueId === null` for its entire lifetime, and every comment/PATCH it ever attempts throws 403 `cross_issue_influence_run_context_required` — including on the issue it just checked out, including on an issue it creates itself.

**Why not fail closed at checkout instead (the issue's own original proposal)?** Because every legitimate run starts unscoped by design — checkout is *how* a run learns which issue it owns. Requiring a pre-existing persisted source issue at checkout would make checkout itself refuse the documented normal case, not just the first write after it.

## What Changed

- `server/src/services/cross-issue-influence-limit.ts`:
  - `observeCrossIssueInfluence`: when the row-locked run lookup finds `sourceIssueId === null`, persist `contextSnapshot = { issueId: input.targetIssueId, source: "first_write_bind" }` on that same locked `heartbeatRuns` row instead of throwing, treat the write as same-issue (no cap decrement), and allow it. Every later write from that run goes through the existing same-issue-unrestricted / cross-issue-capped path unchanged — the bind only ever happens once, on the very first write.
  - New `bindCheckoutRunSourceIssueIfUnset` helper: belt-and-suspenders bind of the **calling** actor's own run's `contextSnapshot.issueId` at checkout time when it's still null. Best-effort (swallows/logs errors) and never overwrites an already-bound run.
- `server/src/routes/issues.ts`: `POST /issues/:id/checkout` calls `bindCheckoutRunSourceIssueIfUnset` for agent actors right after a successful checkout, binding intent earlier than the run's first write.
- Unknown/wrong-company/wrong-agent/malformed-UUID/non-`running` run rejections are unchanged (those branches return before the null-check).
- The existing 20-write cross-issue cap and log-only→enforce rollout (`CROSS_ISSUE_INFLUENCE_ENFORCE_AT`) are untouched for every write after a bind.

## Tests

- `server/src/__tests__/cross-issue-influence-limit.test.ts` (unit, mocked db):
  - a null/empty-`contextSnapshot` run's first write binds and returns `null` (no cap cost), and the mocked `update` call is asserted with the persisted `{ issueId, source: "first_write_bind" }`.
  - a run already bound (`contextSnapshot.issueId` set) writing to a *different* issue is still gated by the existing cap logic (allowed under cap) — not unconditionally rejected.
  - existing unknown/wrong-agent/wrong-company/malformed-run-id/same-issue tests are unchanged and still green.
- `server/src/__tests__/cross-issue-influence-limit-postgres.test.ts` (embedded Postgres):
  - new regression: a run created with `contextSnapshot = {}` checks out issue A (first write succeeds and binds), a second same-issue write to A stays free, a write to a different issue B is capped rather than refused, and re-reading the row shows the bind durably persisted (covers a process-lost/retried run reusing the same run id, the exact shape AGE-2030 hit internally).
  - new regression for `bindCheckoutRunSourceIssueIfUnset`: binds an unbound run at checkout time, and a later checkout of a different issue by the same run does not clobber the existing bind.
  - existing concurrent-cap-serialization test is unchanged and still green.

## Verification

```sh
npx vitest run \
  server/src/__tests__/cross-issue-influence-limit.test.ts \
  server/src/__tests__/cross-issue-influence-limit-postgres.test.ts \
  server/src/__tests__/interaction-resolution-cross-issue-cap-postgres.test.ts \
  server/src/__tests__/issue-comment-reopen-routes.test.ts \
  server/src/__tests__/issue-agent-mutation-ownership-routes.test.ts \
  server/src/__tests__/issue-stale-execution-lock-routes.test.ts
```
Result: 6 test files passed, 216 tests passed.

```sh
pnpm --filter @paperclipai/plugin-sdk build && npx tsc --noEmit -p server
```
Result: exit 0, no errors, on the full `server` `tsc` project (not just the two touched files).

## Risks

Low risk.

- The bind only fires once per run, exactly at the moment `sourceIssueId` would otherwise have been null forever; every other branch (same-issue bypass, cross-issue cap-and-count, malformed/unknown/wrong-agent/wrong-company/non-running run rejections) is untouched.
- The checkout-side bind is additive, best-effort, and defensively coded to never overwrite an already-bound `contextSnapshot` — verified by the new Postgres regression test.
- No schema or migration changes.

## Related/duplicate PRs

#11500 ("count instead of refuse a run with no recorded source issue") addresses a similar symptom but with a different, weaker design: it treats *every* no-source write as counting against the cross-issue cap from write #1, so an unscoped-wake run can be capped out on its own checked-out issue after 20 writes. This PR's bind-on-first-write design instead gives the run a real, persisted "home" issue on its first write, so only genuine cross-issue writes ever count against the cap — matching the corrected design in #11852.

## Model Used

Claude (via the Pi coding agent harness / Paperclip agent runtime). Standard reasoning, with tool use for repository search, file editing, running the embedded-Postgres test suite, and typechecking in an isolated git worktree off `origin/master`.

## Checklist

- [x] I have searched GitHub for duplicate or related PRs and linked them above (#11500, and duplicates #11975/#12118 of the linked issue)
- [x] I have linked the existing issue (#11852) this fixes, following its corrected-design comment
- [x] I have added tests for the new behavior
- [ ] Documentation update — not applicable; internal server authorization-gate fix with no user/operator-facing documentation surface.
- [x] I have considered and documented risks above

### DIFF
--- a/server/src/__tests__/cross-issue-influence-limit-postgres.test.ts
+++ b/server/src/__tests__/cross-issue-influence-limit-postgres.test.ts
@@ -14,6 +14,7 @@ import {
 } from "./helpers/embedded-postgres.js";
 import {
   CROSS_ISSUE_INFLUENCE_ENFORCE_AT,
+  bindCheckoutRunSourceIssueIfUnset,
   observeCrossIssueInfluence,
 } from "../services/cross-issue-influence-limit.js";
 
@@ -112,5 +113,138 @@ describeEmbeddedPostgres("cross-issue influence limit PostgreSQL serialization",
       .where(and(eq(activityLog.companyId, companyId), eq(activityLog.runId, runId)));
     expect(recorded.filter((row) => row.action === "issue.cross_issue_influence_observed")).toHaveLength(20);
     expect(recorded.filter((row) => row.action === "issue.cross_issue_influence_cap_rejected")).toHaveLength(1);
+  });
+
+  it("binds a run with no recorded source issue to the first write's target, persists it on the row, and caps only writes to a different issue thereafter", async () => {
+    const companyId = randomUUID();
+    const agentId = randomUUID();
+    const runId = randomUUID();
+    const issueA = randomUUID();
+    const issueB = randomUUID();
+
+    await db.insert(companies).values({
+      id: companyId,
+      name: "Paperclip",
+      issuePrefix: `C${companyId.replace(/-/g, "").slice(0, 6).toUpperCase()}`,
+      defaultResponsibleUserId: "board-user",
+    });
+    await db.insert(agents).values({
+      id: agentId,
+      companyId,
+      name: "Unscoped Wake Agent",
+      role: "engineer",
+      adapterType: "codex_local",
+      adapterConfig: {},
+      runtimeConfig: {},
+      permissions: {},
+    });
+    // The exact shape a plain unscoped-wake heartbeat run starts with: no
+    // issueId/taskId recorded yet.
+    await db.insert(heartbeatRuns).values({
+      id: runId,
+      companyId,
+      agentId,
+      status: "running",
+      responsibleUserId: "board-user",
+      contextSnapshot: {},
+    });
+
+    // First write of the run's life, against the issue it checked out (A):
+    // must succeed and must not be treated as a cross-issue write.
+    const firstWrite = await observeCrossIssueInfluence(db, {
+      companyId,
+      runId,
+      agentId,
+      targetIssueId: issueA,
+      kind: "comment",
+    });
+    expect(firstWrite).toBeNull();
+
+    const [boundRun] = await db
+      .select({ contextSnapshot: heartbeatRuns.contextSnapshot })
+      .from(heartbeatRuns)
+      .where(eq(heartbeatRuns.id, runId));
+    expect(boundRun?.contextSnapshot).toMatchObject({ issueId: issueA, source: "first_write_bind" });
+
+    // A second write to the same issue A still costs nothing against the cap.
+    const secondWriteSameIssue = await observeCrossIssueInfluence(db, {
+      companyId,
+      runId,
+      agentId,
+      targetIssueId: issueA,
+      kind: "update",
+    });
+    expect(secondWriteSameIssue).toBeNull();
+
+    // A write to a different issue B is gated by the existing cross-issue cap,
+    // not unconditionally rejected — it is allowed here (well under the cap).
+    const crossIssueWrite = await observeCrossIssueInfluence(db, {
+      companyId,
+      runId,
+      agentId,
+      targetIssueId: issueB,
+      kind: "comment",
+      now: CROSS_ISSUE_INFLUENCE_ENFORCE_AT,
+    });
+    expect(crossIssueWrite).toMatchObject({ allowed: true, count: 1 });
+
+    // Re-fetching the row (simulating a process-lost/retried run reusing the
+    // same run id) shows the bind still persisted and no further mutation on
+    // the second same-issue write above.
+    const [runAfter] = await db
+      .select({ contextSnapshot: heartbeatRuns.contextSnapshot })
+      .from(heartbeatRuns)
+      .where(eq(heartbeatRuns.id, runId));
+    expect(runAfter?.contextSnapshot).toMatchObject({ issueId: issueA, source: "first_write_bind" });
+  });
+
+  it("bindCheckoutRunSourceIssueIfUnset binds the calling actor's own run at checkout time, and never overwrites an already-bound run", async () => {
+    const companyId = randomUUID();
+    const agentId = randomUUID();
+    const runId = randomUUID();
+    const issueA = randomUUID();
+    const issueB = randomUUID();
+
+    await db.insert(companies).values({
+      id: companyId,
+      name: "Paperclip",
+      issuePrefix: `C${companyId.replace(/-/g, "").slice(0, 6).toUpperCase()}`,
+      defaultResponsibleUserId: "board-user",
+    });
+    await db.insert(agents).values({
+      id: agentId,
+      companyId,
+      name: "Checkout Binder",
+      role: "engineer",
+      adapterType: "codex_local",
+      adapterConfig: {},
+      runtimeConfig: {},
+      permissions: {},
+    });
+    await db.insert(heartbeatRuns).values({
+      id: runId,
+      companyId,
+      agentId,
+      status: "running",
+      responsibleUserId: "board-user",
+      contextSnapshot: null,
+    });
+
+    await bindCheckoutRunSourceIssueIfUnset(db, { companyId, runId, agentId, issueId: issueA });
+
+    const [afterCheckout] = await db
+      .select({ contextSnapshot: heartbeatRuns.contextSnapshot })
+      .from(heartbeatRuns)
+      .where(eq(heartbeatRuns.id, runId));
+    expect(afterCheckout?.contextSnapshot).toMatchObject({ issueId: issueA, source: "issue.checkout" });
+
+    // A later checkout of a different issue by the same run must not clobber
+    // the already-bound source issue.
+    await bindCheckoutRunSourceIssueIfUnset(db, { companyId, runId, agentId, issueId: issueB });
+    const [afterSecondCheckout] = await db
+      .select({ contextSnapshot: heartbeatRuns.contextSnapshot })
+      .from(heartbeatRuns)
+      .where(eq(heartbeatRuns.id, runId));
+    expect(afterSecondCheckout?.contextSnapshot).toMatchObject({ issueId: issueA, source: "issue.checkout" });
   });
 });
--- a/server/src/__tests__/cross-issue-influence-limit-postgres.test.ts
+++ b/server/src/__tests__/cross-issue-influence-limit-postgres.test.ts
@@ -247,4 +247,49 @@ describeEmbeddedPostgres("cross-issue influence limit PostgreSQL serialization",
       .where(eq(heartbeatRuns.id, runId));
     expect(afterSecondCheckout?.contextSnapshot).toMatchObject({ issueId: issueA, source: "issue.checkout" });
   });
+
+  it("bindCheckoutRunSourceIssueIfUnset merges into an existing populated contextSnapshot instead of replacing it", async () => {
+    const companyId = randomUUID();
+    const agentId = randomUUID();
+    const runId = randomUUID();
+    const issueA = randomUUID();
+
+    await db.insert(companies).values({
+      id: companyId,
+      name: "Paperclip",
+      issuePrefix: `C${companyId.replace(/-/g, "").slice(0, 6).toUpperCase()}`,
+      defaultResponsibleUserId: "board-user",
+    });
+    await db.insert(agents).values({
+      id: agentId,
+      companyId,
+      name: "Checkout Binder Merge",
+      role: "engineer",
+      adapterType: "codex_local",
+      adapterConfig: {},
+      runtimeConfig: {},
+      permissions: {},
+    });
+    await db.insert(heartbeatRuns).values({
+      id: runId,
+      companyId,
+      agentId,
+      status: "running",
+      responsibleUserId: "board-user",
+      contextSnapshot: { wakeReason: "heartbeat_timer", modelProfile: "reasoning-high" },
+    });
+
+    await bindCheckoutRunSourceIssueIfUnset(db, { companyId, runId, agentId, issueId: issueA });
+
+    const [afterCheckout] = await db
+      .select({ contextSnapshot: heartbeatRuns.contextSnapshot })
+      .from(heartbeatRuns)
+      .where(eq(heartbeatRuns.id, runId));
+    expect(afterCheckout?.contextSnapshot).toEqual({
+      wakeReason: "heartbeat_timer",
+      modelProfile: "reasoning-high",
+      issueId: issueA,
+      source: "issue.checkout",
+    });
+  });
 });
--- a/server/src/__tests__/cross-issue-influence-limit.test.ts
+++ b/server/src/__tests__/cross-issue-influence-limit.test.ts
@@ -13,6 +13,7 @@ function counterDb(
 ) {
   let observedCount = initialCount;
   const inserted: Array<Record<string, unknown>> = [];
+  const updated: Array<Record<string, unknown>> = [];
   const tx = {
     select: (selection: Record<string, unknown>) => ({
       from: () => ({
@@ -37,6 +38,13 @@ function counterDb(
         },
       }),
     }),
+    update: () => ({
+      set: (value: Record<string, unknown>) => ({
+        where: async () => {
+          updated.push(value);
+        },
+      }),
+    }),
     insert: () => ({
       values: async (value: Record<string, unknown>) => {
         inserted.push(value);
@@ -49,6 +57,7 @@ function counterDb(
       transaction: async (callback: (value: typeof tx) => Promise<unknown>) => callback(tx),
     },
     inserted,
+    updated,
     get observedCount() {
       return observedCount;
     },
@@ -198,7 +207,7 @@ describe("cross-issue influence limit rollout", () => {
     expect(fake.inserted).toEqual([]);
   });
 
-  it("fails closed when the persisted run has no source issue", async () => {
+  it("binds the run's source issue on the first write instead of rejecting, and does not count it against the cap", async () => {
     const fake = counterDb(0, { contextSnapshot: {} });
 
     await expect(observeCrossIssueInfluence(fake.db as never, {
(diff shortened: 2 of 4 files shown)


dzianisv:
Blocking issue found (design is otherwise correct, tests are good coverage of the cap/persistence semantics — but this one behavior isn't tested and is wrong):

Both new write sites do a **full replace** of `contextSnapshot`, not a merge:

```ts
// cross-issue-influence-limit.ts, observeCrossIssueInfluence bind-on-first-write branch
await tx.update(heartbeatRuns)
  .set({ contextSnapshot: { issueId: input.targetIssueId, source: "first_write_bind" } })
  ...

// bindCheckoutRunSourceIssueIfUnset
await tx.update(heartbeatRuns)
  .set({ contextSnapshot: { issueId: input.issueId, source: "issue.checkout" } })
  ...
```

A plain unscoped-wake run's `contextSnapshot` at creation is not `{}` in the general case — it commonly carries `wakeReason` (e.g. `"heartbeat_timer"`, `"issue_assigned"`), and can carry `modelProfile`, `taskKey`, `commentId`, `wakeCommentId`, `wakeTriggerDetail`, `monitorAttemptCount`, `forceFreshSession`, `skipIssueComment`, etc. — set at run creation in `heartbeat.ts` (73 references to `wakeReason` in that file). Several of these are read later in the same run's lifecycle:

- `shouldResetTaskSessionForWake` — reads `wakeReason`/`forceFreshSession` to decide whether to reset the task session
- `shouldRequireIssueCommentForWake` — reads `wakeReason` (`issue_assigned`, `execution_review_requested`, etc.) to decide whether an issue comment is mandatory
- `allowsIssueInteractionWake` — reads `wakeReason` against `ISSUE_TREE_CONTROL_INTERACTION_WAKE_REASONS`

Both the `observeCrossIssueInfluence` bind and `bindCheckoutRunSourceIssueIfUnset` currently wipe every one of those keys the moment the run makes its first write (or checks out an issue), because the update replaces the whole JSON column instead of merging into it. The trigger condition here is "no `issueId` on the snapshot" — it is not "the snapshot is empty" — so this will fire and clobber real, populated fields on the common wake-scoped-but-not-yet-issue-bound run shape, not just the edge case the tests construct (`{}` / `null`).

Fix: read the existing `contextSnapshot` inside the same locked transaction (already fetched — `run.contextSnapshot`) and spread it before setting `issueId`/`source`:

```
contextSnapshot: { ...(existingSnapshotObjectOrEmpty), issueId: input.targetIssueId, source: "first_write_bind" }
```

same pattern for the checkout helper (it already selects `contextSnapshot` — just spread it instead of discarding it).

Please add a regression test where the pre-bind `contextSnapshot` carries a `wakeReason` (or another arbitrary key) and assert it survives the bind, in addition to the existing `{}`/`null` cases.

Everything else — the cap-after-bind behavior, the persisted-bind-survives-retry test, the untouched branches — looks right. Re-request review once the merge-not-replace fix and the added regression test are in.

dzianisv:
Fixed the blocking defect flagged in review: both bind sites (bind-on-first-write in `observeCrossIssueInfluence`, and `bindCheckoutRunSourceIssueIfUnset`) now spread the already row-locked `contextSnapshot` before setting `issueId`/`source`, instead of replacing the whole column. Added regression tests in `cross-issue-influence-limit.test.ts` (unit, mocked tx) and `cross-issue-influence-limit-postgres.test.ts` (real DB) asserting a populated pre-bind snapshot (`wakeReason`, `modelProfile`, `taskKey`, `forceFreshSession`) survives the bind unchanged aside from the added fields. All 13 unit tests + 4 postgres tests pass locally. Re-requesting review from @df675c53.

dzianisv:
@devinfoley @nickyleach — this PR is CLEAN/MERGEABLE with all checks green, but the `dzianisv` identity has pull-only access to this repo (no push, no merge rights — confirmed via `repos/paperclipai/paperclip` permissions and `RequestReviews`/`MergePullRequest` GraphQL both rejecting with a permissions error). Could one of you merge this directly, or grant merge rights so this doesn't require a manual ping on every PR? Tracked at Paperclip AGE-841.

dzianisv:
Review verdict: **approve**. The merge-not-replace fix (commit c84cb53d) resolves the blocking defect I flagged in my prior review — both bind sites now spread the existing `contextSnapshot` before setting `issueId`/`source`. New regression tests exercise the merge path directly. All CI checks green (30+ checks, mergeStateStatus: CLEAN). I do not have merge/push rights on this repo (pull-only contributor account) — a maintainer with write access needs to merge this. This is the sole blocker on AGE-2030 (cross_issue_influence_run_context_required 403 blocking all agent heartbeat writes company-wide).

digital-efficacy:
Same root cause, and I agree with your correction on #11852 that deny-at-checkout is the wrong direction — an unscoped discovery wake is the *normal* start state, so tightening checkout breaks the documented loop rather than fixing it.

Two things that may be useful here.

### There is an existing open PR on this exact branch point

**#11739** (open since 19-08, CI green, `MERGEABLE`) changes the same `if (!sourceIssueId) throw ...` line in `observeCrossIssueInfluence()`. It hasn't had maintainer review, so it's easy to miss — but it means this PR and that one now edit the same two files (`services/cross-issue-influence-limit.ts`, `__tests__/cross-issue-influence-limit.test.ts`) and will conflict. Worth linking them so a maintainer picks one deliberately rather than discovering the overlap at merge time.

They resolve the null case differently, and the difference is the part I think is worth discussing:

| | anchor for "home" | persisted? |
|---|---|---|
| #11739 | the issue this run **holds the execution checkout on** (`issues.executionRunId == runId`); otherwise still 403 | no — re-derived per request |
| this PR | the issue of the run's **first write**, whatever it is | yes, permanently |

### The first-write anchor is weaker than the checkout anchor

In the `!sourceIssueId` branch here, the bind takes `input.targetIssueId` unconditionally:

```ts
.set({ contextSnapshot: { ...priorSnapshot, issueId: input.targetIssueId, source: "first_write_bind" } })
...
return null;
```

Nothing in that path requires the run to have checked the target out. So for a run that reaches its first write unbound, the first write is allowed *and* exempt from cap accounting on an issue the run never claimed — and that issue then becomes home for the run's lifetime.

To be clear about the scope of this: it is not an authority escalation, since `issue:mutate` is authorized separately and earlier. The cap's job is per-run blast-radius limiting, and this is where the two designs diverge — the first write escapes that accounting entirely.

The checkout bind in item 2 narrows the window but can't close it by construction: it's documented best-effort, and the `catch` swallows failures to `logger.warn` without failing the checkout. Any run whose bind didn't land still reaches first-write unbound, so the permissive path is the one carrying the safety property.

### Ordering-dependent side effect of persisting

Because the bind is sticky, a run whose first write *isn't* its checked-out issue anchors home elsewhere — after which writes to the issue it actually checked out take the `sourceIssueId !== targetIssueId` path and decrement the 20-write cap. Same run, same work, different accounting purely from write ordering.

### Suggested synthesis

These compose rather than compete:

1. keep the checkout-time bind from this PR (it fixes the root gap — checkout never patches the *calling* actor's own run row);
2. in the null branch, gate on the checkout test from #11739 (`issues.executionRunId == runId`) instead of accepting any target;
3. keep 403 for an unscoped run that holds no checkout on the target — that case is genuinely unanchored, and it's the one the original error was right to refuse.

That keeps your bind-don't-deny direction while anchoring "home" to something the run demonstrably holds. Happy to rebase #11739 onto this if you'd prefer a single PR.

One note on #11739 worth carrying over: it reads the `issues` row *without* `for update` on purpose. The lock there is on `heartbeat_runs`, and the checkout-clearing paths take `issues` first and `heartbeat_runs` second, so locking the issue row inside that transaction orders the two sides against each other. If you adopt the checkout test into this PR's transaction, that ordering matters.

EZEZ-QA-Engineer (changes requested):
Blocking behavior/security finding: `observeCrossIssueInfluence` accepts the first unbound write to **any** target as the run's permanent home (cross-issue-influence-limit.ts, new `if (!sourceIssueId)` branch), without establishing that the run has checked out that target. `bindCheckoutRunSourceIssueIfUnset` narrows the normal path but is explicitly best-effort and swallows failures, so the fallback remains reachable. That lets an unscoped run make one unaccounted, sticky write to an issue it never claimed, weakening the intended per-run cross-issue blast-radius control and making accounting depend on write ordering.

Please keep the checkout-time bind, but make the unbound fallback prove that the target issue is currently held by this `runId` (for example, check `issues.executionRunId == runId` without introducing a reverse lock order); otherwise retain the existing 403. Add regression coverage for an unbound run attempting its first write to a non-checked-out issue, and for the checked-out same-issue fallback. The existing context-snapshot merge fix and its regressions look correct.

digital-efficacy:
Flagging something that may shorten this review round: the remedy the `CHANGES_REQUESTED` asks for already exists as a green, open implementation, including the lock-ordering caveat the review specifically calls out.

> Please keep the checkout-time bind, but make the unbound fallback prove that the target issue is currently held by this `runId` (for example, check `issues.executionRunId == runId` without introducing a reverse lock order)

That is, near-literally, what **#11739** does. From its diff of `observeCrossIssueInfluence()`:

```ts
const ownCheckout = await tx
  .select({ id: issues.id })
  .from(issues)
  .where(and(
    eq(issues.companyId, input.companyId),
    eq(issues.id, input.targetIssueId),
    eq(issues.executionRunId, input.runId),
  ))
  .then((rows) => rows.length > 0);
if (ownCheckout) return null;
throw crossIssueInfluenceRunContextError();
```

and it reads without `for update` deliberately, for the reason the review anticipates — the lock above is on `heartbeat_runs`, while the checkout-clearing paths take `issues` first and `heartbeat_runs` second, so locking the issue row here would order the two sides against each other.

### Why the finding is structural to this PR's design rather than a detail to patch

Reading both diffs side by side, the difference is what happens to an unbound run:

| | #12315 (this PR) | #11739 |
|---|---|---|
| unbound write to a **non-checked-out** issue | accepted, and `{issueId: targetIssueId, source: "first_write_bind"}` is **persisted** to `contextSnapshot` | `403`, unchanged |
| unbound write to the run's **own checked-out** issue | accepted | accepted |
| durability | sticky — the first target becomes home for the rest of the run | none; re-derived per request, so a released checkout refuses the next write |

So the blast-radius property the review is protecting isn't recoverable by tightening the fallback here — the fallback *is* the feature in the bind-on-first-write design. `bindCheckoutRunSourceIssueIfUnset` is explicitly best-effort and swallows into `logger.warn`, so the unbound path stays reachable whenever the bind fails or the run never checked out at all.

### Two ways these compose

1. **This PR keeps its genuinely new part and borrows the fallback.** `bindCheckoutRunSourceIssueIfUnset` — binding at checkout time rather than waiting for the first write — is the real contribution here and #11739 has no equivalent. Keep it; replace the unbound-any-target branch with the `executionRunId == runId` check above. That satisfies the review without losing anything this PR adds.
2. **Land #11739 first, rebase this on top.** #11739 is +150/-5, touches no route file, persists nothing, and has been CI-green since 20-08. The checkout-time bind then goes on top as a separate, smaller change.

Either is fine by me and I'd rather see the gap closed than see a particular PR land — but in fairness I should say I opened #11739, so read this as an interested party pointing at source, not as a review verdict.

One more for whoever consolidates: **#12621** was opened today taking a third direction on the same branch — counting unscoped-run writes against the cap instead of rejecting or binding them. That makes four open PRs on this one `if (!sourceIssueId)` line (#11739, #12315, #12621, and the earlier #11500). A maintainer picking a direction would unblock more than any one of them individually.

kallehiitola:
I implemented the requested fail-closed checkout proof in https://github.com/dzianisv/paperclip/pull/6. The fallback now requires the same company, agent, and `checkoutRunId` or `executionRunId` before it persists the source issue. It reads the issue without a row lock, preserves the existing lock order, and adds unit plus PostgreSQL regressions for an unclaimed target. The focused six-file suite passes 220 tests, and `tsc --noEmit -p server` passes. The fork-only dependency-review check cannot run because the contributor fork has no dependency graph; it did not report a code failure.

digital-efficacy:
I read `dzianisv/paperclip#6` against the `CHANGES_REQUESTED` and against my own PR #11739, which proposed the same fallback. Two checks below, and then a conclusion that goes against my own PR.

### 1. The added `assigneeAgentId` predicate does not narrow the grant

The fallback in #6 requires more than #11739 did:

```ts
eq(issues.assigneeAgentId, input.agentId),
or(eq(issues.checkoutRunId, input.runId), eq(issues.executionRunId, input.runId)),
```

#11739 keyed on `executionRunId` alone, so I checked whether the extra assignee condition can refuse a run that legitimately holds the lock. It cannot — every path that stamps a live run id onto an issue also pins `assigneeAgentId` to that same run's agent:

- `svc.checkout` sets `assigneeAgentId: agentId` in the same `.set({...})` as `checkoutRunId`/`executionRunId` (`services/issues.ts:8092-8096`), and an agent actor may only check out as itself (`routes/issues.ts:10628`).
- Both stale-lock adoption paths carry `eq(issues.assigneeAgentId, input.actorAgentId)` in their `WHERE` (`services/issues.ts:5164-5175`, `5224-5236`).
- A user actor cannot stamp a live run id at all: `requireAgentRunId` returns `null` for `actor.type !== "agent"` (`routes/issues.ts:3996-4002`), so a board-initiated checkout writes `checkoutRunId: null`.
- Any reassignment clears the lock in the same patch (`services/issues.ts:7694-7702`), so the two can never drift apart afterwards.

So `executionRunId == runId ⇒ assigneeAgentId == actor.agentId` is an invariant for agent actors, and the predicate is free defense-in-depth rather than a behaviour change. It reads as a tightening; it isn't one. Worth a comment in the code saying so, since the next reader will have the same question.

The `or(checkoutRunId, executionRunId)` widening matches the convention already used at `routes/issues.ts:1871` to decide the same "does this run own the lock" question, so it is consistent with the file rather than novel.

### 2. The durability difference argues against #11739, not for it

I previously framed the per-request re-derive in #11739 as the safer property — a released checkout refuses the next write. Tracing the close-out sequence, that framing was wrong, and it is a defect in my PR rather than a virtue.

A terminal status change nulls the very columns #11739 keys on, inside the same update:

```ts
// services/issues.ts:7689-7693
if (issueData.status && issueData.status !== "in_progress") {
  patch.checkoutRunId = null;
  patch.executionRunId = null;
```

So for an unbound run that changes status before it posts its closing comment:

1. `PATCH {status:"done"}` — the guard runs before `svc.update`, `executionRunId` still matches, allowed. The update then clears the lock.
2. `POST /issues/:id/comments` — separate request, guard re-derives, `executionRunId` is now `null`, so it throws `cross_issue_influence_run_context_required` again.

Under #11739 the fix therefore holds only for runs that comment *before* they change status. It makes correctness depend on write ordering inside the run, and it silently reintroduces the original 403 for the reverse order — including any later resume or deferred-comment write in that run. The persisted bind in this PR survives the status change and has no such ordering dependency.

### Conclusion

#12315 with #6 is the better of the two designs, and #11739 should not be merged. The checkout proof the review asked for now exists here, the bind is durable where mine is not, and `bindCheckoutRunSourceIssueIfUnset` has no equivalent in my PR. I will close #11739 as superseded once this merges — no maintainer arbitration needed between the two, and no conflict to resolve at merge time.

kallehiitola:
A reviewed follow-up hardening patch for this PR is ready: https://github.com/dzianisv/paperclip/pull/6 (branch `fix/checked-out-run-binding` → `fix/age-2035-cross-issue-influence-bind-on-first-write`).

What it tightens, verified by review at head commit 679b762:

- **Bind-on-first-write now requires persisted checkout proof.** The target issue must match `companyId` + `assigneeAgentId` and carry `checkoutRunId` or `executionRunId` equal to the writing run; otherwise the existing fail-closed 403 (`cross_issue_influence_run_context_required`) is retained and the run's `contextSnapshot` is left untouched. Timer/unscoped runs can still bind the issue they checked out, but an arbitrary first target no longer becomes the run's home.
- **No new lock ordering.** The ownership proof is a plain `SELECT` (no row lock) inside the run-row-locked transaction, so the existing issue-to-run lock order is preserved.
- **Regressions added** for both paths in the unit and embedded-PostgreSQL suites: checked-out target → bind succeeds; unclaimed target → 403, snapshot unchanged.

The `review` check failing on the fork is the known dependency-review limitation (the fork has no dependency graph enabled); the meaningful quality gates will run here once this PR's head branch picks up the commit.

Note for maintainers/fork owner: merging the patch PR into this PR's head branch requires push access to `dzianisv/paperclip`. Once merged, this PR updates automatically and is ready for fresh review.

digital-efficacy:
Follow-up on my comparison above: I have now **closed #11739** as superseded by this PR ([closing note](https://github.com/paperclipai/paperclip/pull/11739#issuecomment-5490780936)). No competing PR is left for this bug, so nobody has to pick between two implementations or resolve a conflict between them.

Two things carried over from that PR that may be useful here:

- The ordering defect that made me withdraw mine is worth guarding against in review generally: any fix that re-derives permission per request from `issues.executionRunId` breaks for a run that sets a terminal status before posting its close-out comment, because `services/issues.ts:7689-7693` nulls that column in the same update. The bind-on-first-write approach in this PR does not have that dependency — that is the main reason I think this one is correct and mine was not.
- `server/src/__tests__/cross-issue-influence-limit.test.ts` on `fix/cross-issue-timer-run-checkout` (head `4e7ddc1`, branch kept alive) has regressions for the timer-run-writes-its-own-checkout case. Cherry-pick freely if they add coverage beyond what `dzianisv/paperclip#6` already added; no attribution needed.

PR #12337, opened 2026-08-27
Fix blocker ID projections

HJEGeorge:
Fixes control-plane responses that omitted persisted blocker IDs after writes.\n\n- project blockedByIssueIds from persisted relation summaries on direct issue GET\n- include the same projection in heartbeat context and PATCH responses\n- add route-level regression coverage\n\nValidation:\n- pnpm exec vitest run server/src/__tests__/issue-activity-events-routes.test.ts server/src/__tests__/issues-goal-context-routes.test.ts (11 passed)\n- pnpm --filter @paperclipai/server typecheck


HJEGeorge:
Superseded by #12338, which contains only the CYR-2498 fix on current master.

PR #12427, opened 2026-08-28
Stop the stranded watchdog from blocking parked standing loops

raviy06:
## Problem

A standing weekly loop that parks itself back to `todo` after a productive sweep gets escalated to
`blocked` by `reconcileStrandedAssignedIssues` with `stranded_assigned_issue`, burning an
assignment-recovery action. Observed on the same issue six times in three days.

Two independent defects combine:

**1. `hasActiveExecutionPath` cannot see checkout/execution locks.**
It matched only `heartbeat_runs.contextSnapshot->>'issueId' = <issue>`. A coalesced wake stamps the
snapshot with a *different* issue (or an unscoped timer stamps nothing) while the run still holds
this issue's `checkoutRunId`. The issue then looks pathless while a heartbeat is actively running on
it. Real sequence: a timer run coalesced onto a continuation for issue B, checked out issue A 69s
later, and the watchdog escalated A 21s after that.

**2. The park itself is invisible.**
`wasTodoHandedBackDuringOrAfterLatestRun` only fires when the issue's *current* status is still
`todo`, and only counts recovery `handed_back` actions — never an agent's own `PATCH status: todo`.
Once a later wake flips the issue back to `in_progress`, the evidence of a valid disposition is gone
and `isRepeatedProductiveContinuationRecovery` sends it straight to escalation.

## Change

`server/src/services/recovery/service.ts`

- `hasActiveExecutionPath` additionally matches non-terminal runs referenced by
  `issues.checkoutRunId` / `issues.executionRunId`. The locks, not the snapshot, are authoritative.
- New `didLatestRunParkIssueToTodo` reads the `issue.updated` **transition** from the activity log
  (both `details._previous.status` and `details.changes.status` shapes) since the latest successful
  run began, plus the existing recovery `handed_back` signal. Status-independent, so a later
  `in_progress` flip no longer erases the evidence.
- On the `in_progress` + repeated-productive-continuation path, a detected park skips escalation and
  increments a new `standingParkExempted` counter, falling through to the normal continuation-retry
  enqueue (same shape as the existing `recentProgressExempted` exemption). The counter is spread into
  the existing `logger.warn` payload, so exemptions are observable.
- `wasTodoHandedBackDuringOrAfterLatestRun` is refactored onto a shared
  `hadTodoHandBackRecoveryActionSince` helper; its behaviour is unchanged.

Escalation is *narrowed*, never widened: nothing that escalated before stops escalating unless the
agent demonstrably parked the issue during or after the latest successful run.

## Tests

`server/src/__tests__/heartbeat-process-recovery.test.ts`, three cases:

1. **Live checkout lock, foreign snapshot** — running run whose `contextSnapshot.issueId` is another
   issue but which holds this issue's `checkoutRunId`. Expects no escalation, no comment, status
   stays `in_progress`.
2. **Standing-loop park** — latest succeeded productive-continuation run logged
   `in_progress -> todo`. Expects `escalated: 0`, `standingParkExempted: 1`,
   `continuationRequeued: 1`, and no `stranded_issue_recovery` issue created.
3. **Regression guard** — identical fixture with the park timestamped *before* the run started.
   Expects `escalated: 1`, `standingParkExempted: 0`, status `blocked`.

Cases 1 and 2 were confirmed to fail on the parent commit before the fix.

## Note, not fixed here

`classifyIssueGraphLiveness`'s `activeRuns` input joins `issues.executionRunId` only, so the same
checkout-lock blind spot exists in the issue-graph liveness watchdog. Left alone deliberately —
different watchdog, no observed misfire, and narrowing that one belongs in its own change.

## How to test

```
  src/__tests__/issue-recovery-actions.test.ts
```

```test-evidence:
sha: 1c8bb9b704d080af9088375a1a1e86120a4ef170
command: pnpm --filter @paperclipai/server exec vitest run src/__tests__/heartbeat-process-recovery.test.ts src/__tests__/issue-recovery-actions.test.ts src/__tests__/recovery-observability.test.ts src/__tests__/recovery-stale-issue-lock-sweep.test.ts src/__tests__/recovery-classifiers.test.ts
exit: 0
totals: 5 test files passed, 192 tests passed, 0 failed (64.08s)
command: pnpm --filter @paperclipai/server typecheck
exit: 0
totals: tsc clean, 0 errors
worktree: clean before and after; local == fork remote == PR head (1c8bb9b704d080af9088375a1a1e86120a4ef170)
scope: this repo has no bin/ci; the full gate is `pnpm test:run` (whole-workspace). The scoped
  recovery suites above plus the server typecheck were run locally; full-suite coverage is left to
  the PR's GitHub checks.
finished_at: 2026-08-28T08:30:21Z
```

### DIFF
--- a/server/src/__tests__/heartbeat-process-recovery.test.ts
+++ b/server/src/__tests__/heartbeat-process-recovery.test.ts
@@ -7835,6 +7835,126 @@ describeEmbeddedPostgres("heartbeat orphaned process recovery", () => {
     expect(issue?.status).toBe("blocked");
   });
 
+  it("treats a running checkout lock as a live execution path when the run snapshot names another issue", async () => {
+    const { companyId, agentId, issueId, runId } = await seedStrandedIssueFixture({
+      status: "in_progress",
+      runStatus: "succeeded",
+      retryReason: "issue_continuation_needed",
+      runSource: "issue.productive_terminal_continuation_recovery",
+      livenessState: "advanced",
+    });
+    const liveRunId = randomUUID();
+    await db.insert(heartbeatRuns).values({
+      id: liveRunId,
+      companyId,
+      agentId,
+      invocationSource: "timer",
+      triggerDetail: "system",
+      status: "running",
+      contextSnapshot: {
+        issueId: randomUUID(),
+        retryReason: "issue_continuation_needed",
+        source: "issue.continuation_recovery",
+      },
+      startedAt: new Date("2026-03-19T00:10:00.000Z"),
+      updatedAt: new Date("2026-03-19T00:10:00.000Z"),
+    });
+    await db
+      .update(issues)
+      .set({ checkoutRunId: liveRunId, updatedAt: new Date("2026-03-19T00:10:05.000Z") })
+      .where(eq(issues.id, issueId));
+    const heartbeat = heartbeatService(db);
+
+    const result = await heartbeat.reconcileStrandedAssignedIssues();
+    expect(result.escalated).toBe(0);
+    expect(result.continuationRequeued).toBe(0);
+    expect(result.issueIds).toEqual([]);
+
+    const issue = await db.select().from(issues).where(eq(issues.id, issueId)).then((rows) => rows[0] ?? null);
+    expect(issue?.status).toBe("in_progress");
+
+    const comments = await db.select().from(issueComments).where(eq(issueComments.issueId, issueId));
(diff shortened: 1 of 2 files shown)


raviy06:
Superseded by #12428. Same commit, but the branch name carried an internal ticket id and the description did not follow the PR template. Closing this one to keep the queue clean.

PR #12441, opened 2026-08-28
fix(ci): validate equivalent PR merge refs

cryppadotta:
## Summary

- validate the live master ref and event base SHA before AWS routing
- accept GitHub synthetic merge commits only when the event and live commits have the exact expected base/head parents and identical tree
- preserve fail-closed routing for malformed, stale, replaced, or untrusted events

## Canary finding

A trusted reopened PR produced two synthetic merge SHAs with different timestamps but identical current base/head parents and tree. The former exact-SHA comparison safely fell back to GitHub-hosted runners, but could not route a valid event to AWS.

## Validation

- `actionlint .github/workflows/pr-trusted.yml .github/workflows/pr.yml`
- `node --test ./scripts/__tests__/e2e-shard.test.mjs`
- real-event gate simulation selects the Fleet label for equivalent merge commits
- negative simulations keep an untrusted sender and replaced head on `ubuntu-latest`
- AWS routing remains disabled during rotation

### DIFF
--- a/.github/workflows/pr-trusted.yml
+++ b/.github/workflows/pr-trusted.yml
@@ -40,6 +40,7 @@ jobs:
           EVENT_SENDER_ID: ${{ github.event.sender.id }}
           EVENT_BASE_REPOSITORY_ID: ${{ github.event.pull_request.base.repo.id }}
           EVENT_BASE_REF: ${{ github.event.pull_request.base.ref }}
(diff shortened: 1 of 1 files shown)


cryppadotta:
@greptileai review this exact head: 633cf8c14e207275cf597e294c38bab23628f696

Issue #10344, opened 2026-07-28
Recovery misclassifies Claude session/weekly-limit failures as stuck work, minting permanent `blocked` issues instead of using the existing provider-quota wait


ethanhawkbridge-art:
**Version:** `@paperclipai/server` 2026.722.0 (current `latest`). Re-checked against 2026.728.0-canary.1 (current `canary`): both gaps are still present there, unchanged.

## Summary

When a heartbeat run dies on a Claude **session limit** or **weekly limit**, recovery classifies it as a *stranded assigned issue* rather than a provider-quota wait. It parks the issue in `blocked` with no blocker edge and no monitor, so nothing ever re-wakes it — even though the underlying condition is transient and self-clearing.

The correct behaviour is already implemented in this file: a `provider_quota` cause arms `ensureProviderQuotaWaitRecoveryMonitor(...)`, schedules a retry, and returns early without minting a notice. That path simply never executes for this failure mode.

Cost on one deployment: roughly 110 dead tickets per day, and a backlog in which 66% of all `blocked` issues were these artefacts. Real work parked by the same path became invisible underneath them.

## Root cause — two independent gaps

`dist/services/recovery/service.js`

A Claude limit failure surfaces as `errorCode: "acpx_turn_failed"`, with `error` carrying the provider text and no `errorFamily`.

**Gap 1 — the error code is unknown to the classifier.** In `isProviderQuotaRecovery()` the third gate is:

```js
if (latestRun?.errorCode !== "adapter_failed") return false;
return /(?:usage|rate|quota) limit|quota (?:exceeded|reset)|try again after/i.test(latestRun.error ?? "");
```

`acpx_turn_failed` is not `adapter_failed`, so the function returns `false` **before the message text is ever examined**. The same code is missing from `TRANSIENT_INFRA_CONTINUATION_ERROR_CODES`, and `classifyAdapterFailureForRecovery()` rejects it in its opening guard for the same reason.

**Gap 2 — the vocabulary is incomplete.** Both quota patterns know `usage limit` but not Claude's actual wording:

```
You've hit your session limit · resets 3pm
You've hit your weekly limit · resets Jul 27 at 10am
```

Neither `session limit` nor `weekly limit` matches `(?:usage|rate|quota) limit`, nor `PROVIDER_QUOTA_ERROR_RE`.

**These are independent, and fixing either alone is not sufficient.** Evaluating the real strings against the shipped patterns:

[...]

PR #12491, opened 2026-08-29
fix(acpx): tag mid-stream transport stalls as transient_upstream so bounded retry arms

### LINKED ISSUE #11257: acpx_turn_failed: mid-stream transport stalls can never reach bounded transient retry (the ACPX engine never persists an errorFamily)
**Verified against `master` @ `67001ec6` (2026-08-12T04:37Z)**, not against the installed dist.
Observed on `@paperclipai/server` / `@paperclipai/adapter-utils` `2026.722.0`, adapter `claude_local` (ACPX engine), subscription auth, single-machine local instance.

## Summary

`scheduleBoundedRetryForRun()` is unreachable for **every** failure the ACPX engine produces. A mid-stream transport stall and a hard capacity cap are both reported as `errorCode: "acpx_turn_failed"`, and the engine never sets `resultJson.errorFamily`, so the server's transient classifier returns `null` and the bounded retry ladder (2m / 10m / 30m / 2h) is never armed for either.

This is the **transport-stall half** of `acpx_turn_failed`. The capacity-cap half is #10344 (misrouted away from the `provider_quota` wait). The two are complementary and, importantly, must **not** be fixed by widening the same allow-list — see "Why not just add the code to the allow-list" below.

## The chain, on `master` @ `67001ec6`

1. `packages/adapter-utils/src/acpx-engine/execute.ts:3717` — on a failed turn the engine returns
   `errorCode: terminal.status === "failed" ? "acpx_turn_failed" : …`.
   **`errorFamily` appears 0 times in this file.**
2. `server/src/services/heartbeat.ts:548` `readHeartbeatRunErrorFamily()` reads `resultJson.errorFamily` first, then falls back to a fixed `errorCode` allow-list — `provider_quota`, `codex_transient_upstream`, `claude_transient_upstream`, `codex_harness_crash`. `acpx_turn_failed` is in neither ⇒ `return null`.
3. `server/src/services/heartbeat.ts:588` `readTransientRecoveryContractFromRun()` returns `null`, because the family is neither `transient_upstream` nor `provider_quota`.
4. `server/src/services/heartbeat.ts:15971` — the call site is
   `} else if (outcome === "failed" && readTransientRecoveryContractFromRun(livenessRun)) {`
   so `:15972 await scheduleBoundedRetryForRun(livenessRun, agent);` **is dead code for this entire failure class.**

## Observed impact

Over **800 heartbeat runs / 3 days** on one fleet, **32 runs** failed with `acpx_turn_failed` and message `Response stalled mid-stream`:

| field | value |
|---|---|
| `scheduledRetryAt` | **0 / 32** non-null |
| `continuationAttempt` | **0 / 32** non-null |
| `retryOfRunId` on the run itself | 0 / 32 |

The bounded retry never armed once. The most expensive single loss was a **79-minute run, 709 KB of run log, discarded entirely**; the successor started from scratch and re-derived the same intermediate work.

The failure is also **not random, and not uniformly distributed in time** — which is what makes it worth retrying rather than escalating:

- All 32 died in a **9-second band, 3m45s–3m54s past the top of the hour**.
- Negative control: the end-phase of the 397 *successful* runs in the same window is uniform — that same 9-second band holds **5.8%** of them against a 10% uniform expectation.

So this class is frequent, cheap to detect, and its retry is very likely to succeed on a later attempt — precisely the class bounded retry exists for.

## Why not just add `acpx_turn_failed` to the allow-list

`acpx_turn_failed` is emitted for at least two classes with **opposite** retry semantics:

- a mid-stream transport stall — safe and valuable to retry;
- a provider capacity/session cap (`"You've hit your session limit · resets 11pm"`) — retrying on a 2-minute backoff just burns attempts against a wall that clears on its own schedule. That class wants the `provider_quota` wait path, which is what #10344 is about.

Widening the server-side `errorCode` allow-list would fix this issue and simultaneously make #10344's case retry-storm. The split has to happen at the layer that can still see the difference.

## Suggested fix — adapter-side, no server change

The `resultJson.errorFamily` hook already exists and already takes precedence (`heartbeat.ts:548`). The ACPX engine is also the only layer that can still tell the two classes apart at that point: `describeErrorDiagnostics()` yields `acpCode` and `retryable`, and `terminal.error.message` / `stopReason` are in scope at `execute.ts:3717`.

So: set `resultJson.errorFamily = "transient_upstream"` **only** for the transport-stall class, and leave it unset otherwise. Capacity-cap behaviour then stays byte-for-byte identical, and `readHeartbeatRunErrorFamily` needs no edit at all.

This is the same shape as the fix prescribed in #10422 for `hermes_local` (`errorCode` + `errorFamily` from the adapter, scheduler unchanged), and the same mechanism already being added to this exact engine for the quota class in **#10314** — that PR's classification hook is where the stall class would naturally slot in alongside.

(If you'd rather keep the decision server-side, the equivalent is to key off `errorMeta.retryable === true` in `readHeartbeatRunErrorFamily`. The adapter has strictly more information, so I'd prefer the first.)

## Related, not duplicate

- #10344 — the **other** class of `acpx_turn_failed` (session/weekly limit → should take the `provider_quota` wait). Fixing that one does not make this one retryable.
- #10314 — open PR classifying provider-quota refusals **in the ACPX engine**. Same file, same hook, adjacent class.
- #10422 — same defect shape in `hermes_local`; the "adapter emits the family, scheduler untouched" resolution is what I'm proposing here.
- #5720 — the analogous stall (`Stream idle timeout - partial response received`) on the older `adapter-claude-local` CLI path, where the gap is a regex rather than a missing field. Different code path, same operator-visible symptom.

## Follow-on

Even once retry is armed, the successor still cold-starts, because the resume pointer is only ever written for succeeded runs. Filed separately as #11258.

MrBlackTongue:
## Thinking Path

> The ACPX engine reports a mid-stream transport stall and a hard provider capacity cap under one code, `acpx_turn_failed`, but they want opposite retry semantics. Because the engine never persists `errorFamily`, the heartbeat bounded-retry ladder never arms for transport stalls (issue #11257). This classifies the two at the adapter, tagging stalls `transient_upstream` while leaving the capacity-cap subclass on its `provider_quota` path.

Fixes #11257. Searched the open PR list — no existing PR adds ACPX transport-stall errorFamily classification.

## What Changed

- **`classifyAcpxTurnFailureErrorFamily(error)`** (new, pure): capacity/session-cap guard runs FIRST and vetoes (returns `null` → stays on quota path, #10344); only then is the transport-stall signature tested → `"transient_upstream"`. Never lumps the two subclasses.
- **Terminal-result path**: sets the family on a `failed` turn (top-level `errorFamily` and `resultJson.errorFamily`). No server change, no `errorCode` allow-list widening — `readHeartbeatRunErrorFamily` already arms the 2m/10m/30m/2h ladder off `transient_upstream`.

## Verification

- `pnpm exec vitest run packages/adapter-utils/src/acpx-engine/execute.test.ts` — 162 passed (154 + 8 new): stall variants → transient_upstream; cap messages and a cap+stall message → null; two integration runs through the real `execute()` harness. `adapter-utils` typecheck — 0 errors.

## Risks

- Text-based classification: an unmatched reword is a safe false-negative (preserves today's no-retry). Only the terminal-result path (all 32 observed stalls) is classified; a stall surfacing as a thrown error is a follow-up.

## Model Used

- Claude Opus 4.8 (Anthropic), via Claude Code with tool use.

## Checklist

- [x] I have not referenced internal/instance-local Paperclip issues or links (only public GitHub `#NNN` URLs)
- [ ] Greptile is 5/5 with no open P2s

### DIFF
--- a/packages/adapter-utils/src/acpx-engine/execute.test.ts
+++ b/packages/adapter-utils/src/acpx-engine/execute.test.ts
@@ -27,6 +27,7 @@ vi.mock("@paperclipai/adapter-utils/execution-target", async (importActual) => {
 });
 import {
   buildAcpxRunSummary,
+  classifyAcpxTurnFailureErrorFamily,
   createAcpxEngineExecutor,
   findAncestorBin,
   geminiVersionSupportsNativeAcpFlag,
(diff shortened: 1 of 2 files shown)


MrBlackTongue:
@nickyleach @devinfoley — green and ready for review. Fully closes #11257 (transport stalls tagged transient_upstream so the bounded-retry ladder arms; capacity-cap stays on its quota path). All CI gates pass, Greptile 5/5.

Issue #11893, opened 2026-08-21
Flaky test: document-annotation-routes times out at 5000ms in Verify serialized server suites (3/5)


qasimsethi-code:
**What happened?**

`document-annotation-routes.test.ts > document annotation routes > includes compact open annotations without comment bodies by default for agent document reads` fails intermittently in the `Verify serialized server suites (3/5)` CI lane.

The primary error is a test timeout. A `TypeError` follows it:

```
FAIL @paperclipai/server src/__tests__/document-annotation-routes.test.ts >
  document annotation routes >
  includes compact open annotations without comment bodies by default for agent document reads

Error: Test timed out in 5000ms.
TypeError: db.transaction is not a function
```

The `TypeError` appears to be collateral rather than the cause. It is emitted while asynchronous work continues after vitest has already failed the test on the 5000 ms deadline. A genuinely incomplete mock would fail on every run, and this test does not.

Observed across three CI runs on the same pull request (#11871), with only unrelated commits added between them:

| Run | `Verify serialized server suites (3/5)` | Failing test |
|---|---|---|
| 1 | fail | `document-annotation-routes` (this one) |
| 2 | pass | — |
| 3 | fail | `document-annotation-routes` (this one) |

Run 3: https://github.com/paperclipai/paperclip/actions/runs/32511536547/job/96863858288

Two other tests failed once each across the same three runs and then passed, which suggests the same class of timing sensitivity rather than one specific bug:

- `server/src/__tests__/workspace-runtime.test.ts > workspace runtime startup reconciliation > re-adopts a live service whose shell command differs from the surviving process argv` — `AssertionError: expected { adopted: +0 } to match { adopted: 1 }`
- `packages/adapters/adapter-utils/src/sandbox-callback-bridge.test.ts > sandbox callback bridge > denies non-allowlisted requests by default`

**Expected behavior**

The test passes deterministically, or it declares a timeout budget that matches the work it does.

`server/vitest.config.ts` sets `hookTimeout: 30000` but leaves `testTimeout` at the vitest default of 5000 ms. Suites that provision an embedded PostgreSQL cluster therefore get 30 s for setup but only 5 s per test. On a loaded CI runner the test body can exceed that.

**Steps to reproduce**

1. Open any pull request that runs the `PR` workflow.
2. Wait for the `Verify serialized server suites (3/5)` lane.
3. Re-run the workflow several times. The test fails on some runs and passes on others with no source change.

[...]

PR #12522, opened 2026-08-30
fix(test): stop concurrent vi.importActual() from dropping factory mocks

### LINKED ISSUE #12519: Flaky test: legacy /heartbeat/invoke cases in agent-live-run-routes return 500 instead of 202 in Verify serialized server suites (1/5)
### What happened?

The two legacy `/heartbeat/invoke` cases in `server/src/__tests__/agent-live-run-routes.test.ts` fail intermittently in `ci / Verify serialized server suites (1/5)`:

- `passes scoped wake fields through the legacy heartbeat invoke route`
- `calls heartbeat.wakeup with the legacy minimal shape when the body is empty`

Both fail the same way — the route answers `500 Internal server error` where the test expects `202`:

```
AssertionError: {"error":"Internal server error"}: expected 500 to be 202
 ❯ src/__tests__/agent-live-run-routes.test.ts:610:50
```

This is distinct from #11893 (`document-annotation-routes` timing out in shard 3/5), though it may share a root cause — see the timing note below.

### Expected behavior

The two legacy invoke cases pass deterministically, as the other eight cases in the file do.

### Steps to reproduce

On a clean checkout of `master` (`a639bb1ce`), with `pnpm install --prod=false`:

```bash
cd server
for i in $(seq 1 25); do
  ../node_modules/.bin/vitest run src/__tests__/agent-live-run-routes.test.ts
done
```

The rate is low, so a single pass is not enough to see it — I measured **1 failure / 25 runs** on pristine `master`. It reproduces more readily under CPU contention (I ran two such loops concurrently).

### Two observations that may help

**1. The failing cases burn ~4.8s; the passing ones are well under 1s.** That gap points at some ~5s budget being exceeded under load rather than a logic error, which is also the shape of #11893.

**2. The underlying error is swallowed.** `middleware/error-handler.ts` returns a bare `{"error":"Internal server error"}` for any non-`HttpError`, so CI never shows what actually threw. Inserting an error-logging middleware just before `app.use(errorHandler)` in the test's local `createApp` surfaces it:

```ts
app.use("/api", agentRoutes(db as any));
app.use((err: any, _req: any, _res: any, next: any) => {
  console.error("FLAKE_ERROR:", err?.name, err?.message, err?.stack);
  next(err);
});
app.use(errorHandler);
```

I have not yet caught the flake with that instrumentation attached (the rate makes it slow to hit), so I can't name the root cause — filing this with the reproduction and measurements rather than sitting on it. Happy to keep digging or to test a patch if that's useful.

`master` @ `a639bb1ce`

### Deployment mode

Local

### Node.js version

v24.19.0

### Operating system

Linux 6.8.0 (x86_64, container)

### Additional context

Surfaced while triaging a red shard on #12030. Full attribution data — including a same-tick control run showing the failure occurs on pristine `master` (1/25) and not only on that branch (0/25) — is in https://github.com/paperclipai/paperclip/pull/12030#issuecomment-5465104597.

One knock-on worth flagging: `gh run rerun --failed` refuses runs whose workflow file has since moved ("its workflow file may be broken"), so when this flake lands on a PR the red cannot be retried — it can only be cleared by pushing a new commit.

zannis:
## Thinking Path

> - The server test suites mock service modules with `vi.doMock()` factories, then load the real route modules with `vi.importActual()`.
> - Several suites load two or more of those modules at the same time inside `Promise.all([...])`.
> - Vitest keeps the mock-resolution callstack in one shared field, and its own source says this breaks under `Promise.all(import(), import())`.
> - When two import graphs interleave, a factory mock is skipped and the real module loads instead, which makes a suite fail rarely and with an unrelated error.
> - This pull request awaits those `vi.importActual()` calls one at a time, and adds a test that keeps the pattern from coming back.
> - The benefit is that a rare, expensive-to-diagnose CI failure is removed at its cause.

Fixes: #12519

Refs #11893

#12519 reports that two legacy `/heartbeat/invoke` cases in `server/src/__tests__/agent-live-run-routes.test.ts` intermittently return 500 instead of 202. This pull request supplies the root cause that issue was missing, and the fix.

#11893 reports a different suite, `document-annotation-routes`, timing out at 5000ms. That suite uses the same concurrent-import pattern and is changed here too. I have not proven the two share a cause, so I am linking it as related rather than closing it.

## What Changed

- Await `vi.importActual()` calls one at a time instead of resolving them concurrently in `Promise.all([...])`. Applied to all 38 sites in the repository. The change is ordering-only: the destructured tuple and its types are identical.
- Add `server/src/__tests__/concurrent-import-actual-guard.test.ts`. It walks every `*.test.ts` file in the repository and fails on any `Promise.all([...])` that contains two or more `vi.importActual` calls.

### Why the failure looked unrelated to imports

`server/src/routes/agents.ts` ends the legacy invoke handler with `await logActivity(db, {...})`. The suite mocks the `../services/index.js` barrel and gives `logActivity` a `vi.fn()`. When the barrel factory mock is dropped, the real barrel loads. `agentService`, `heartbeatService` and `issueService` stay mocked, because the suite also mocks their source modules individually, so most tests keep passing. But `logActivity` is re-exported from `./activity-log.js`, which is not individually mocked, so the real implementation runs against the suite's `{}` db stub:

```
TypeError: db.insert is not a function
    at persistActivity (server/src/services/activity-log.ts:163:31)
    at logActivity (server/src/services/activity-log.ts:222:37)
    at server/src/routes/agents.ts:4783:5
```

`middleware/error-handler.ts` converts that to `500 {"error":"Internal server error"}`, which is why neither CI nor a local run ever showed what threw. The only two tests that reach a `logActivity(db, …)` call with the `{}` stub are the two invoke cases, so they alone fail.

Loading the real barrel also transforms the full service graph, which is why the failing test burns seconds more than its ~300ms neighbours. This is not a timeout being exceeded — `testTimeout` is already 15000 — so no timeout is widened here.

## Verification

**Controlled A/B on the cause.** A scratch probe repeated `createApp()` plus the invoke request 150 times in one process and counted 500s, varying only how the two `vi.importActual()` calls are awaited. Two runs per arm:

```
mode=concurrent iterations=150 lost=2
mode=sequential iterations=150 lost=0
mode=concurrent iterations=150 lost=2
mode=sequential iterations=150 lost=0
```

Concurrent 4/300. Sequential 0/300. Nothing else differs between the arms. The probe is scratch and is not part of this diff.

**Deterministic confirmation of the mechanism.** Removing only the `vi.doMock("../services/index.js", …)` registration from a copy of the suite reproduces the same `TypeError` at the same stack frames, on both invoke cases, in 100% of runs, while the other eight tests still pass.

**Regression test, red then green.** `concurrent-import-actual-guard.test.ts` fails on unmodified `master` and lists 38 sites, including `agent-live-run-routes.test.ts:96` and `document-annotation-routes.test.ts:161`. It passes after the change.

**Suite loop.** `server/src/__tests__/agent-live-run-routes.test.ts` run repeatedly after the change on an 8-core machine:

```bash
cd server
for i in $(seq 1 60); do ../node_modules/.bin/vitest run src/__tests__/agent-live-run-routes.test.ts; done
```

- 120 runs at two-way concurrency: 5 failures, none carrying the defect signature. All five were the first test, `returns a compact active run payload for issue polling`, hitting its own explicit `}, 10_000)` budget while I ran `tsc` beside the loop and saturated the machine. See "Open topic" below.
- 40 runs on an idle machine: 0 failures.

Before the change the same signature appeared 3 times in roughly 150 local suite runs, and once in CI.

**Typecheck.** `tsc --noEmit` in `server/` reports 153 errors both with and without this change — they come from `@paperclipai/plugin-sdk` and `@paperclipai/paperclip-runner` not being built when `tsc` is run directly. None of them are in the 39 files this pull request touches.

**Suite duration.** Three runs each on an idle machine, to check that serial imports do not cost wall-clock: before 13.36s / 13.20s / 12.76s, after 13.02s / 13.14s / 13.09s. Unchanged within noise.

### Open topic, not fixed here

`agent-live-run-routes.test.ts` gives its first test an explicit `}, 10_000)` timeout. That overrides the 15000 `testTimeout` in `server/vitest.config.ts`, which the config's own comment says was raised precisely because the first test in a route suite pays the whole module transform cost. On a loaded runner that first test can cross 10s. This is a separate defect from the one fixed here, and I have no evidence it caused the failures in #12519, so I have left it alone rather than widen this diff. Happy to send it separately if you want it.

## Risks

Low risk, and limited to test code. No production file changes.

- The imports are now serial, so a suite pays its module transform cost one module at a time. Measured on the changed suite (numbers above), total duration is unchanged within noise, because the transform work is CPU-bound and was never truly parallel.
- Behaviour differs from `Promise.all` only when an import rejects: the second import is no longer started. A failing import fails the test either way.
- The guard test is textual. It would flag a `Promise.all` that contains two `vi.importActual` calls inside a nested callback, which is not the unsafe shape. No such site exists today, and the assertion message says what to do.
- The underlying Vitest limitation is unchanged. This pull request makes the repository use the API in the way that is safe today.

## Model Used

Claude Opus 5 (`claude-opus-5`), extended thinking, agentic tool use with file editing and shell execution.

## Checklist

Depends-on: none (self-contained test-only change with no prerequisite PR)


zannis:
Good catch — this one is valid and now fixed in `0c99916`.

The guard matched the literal text `Promise.all([`, so `Promise.all( [` and an array argument on the next line both slipped past it. That is not hypothetical here: the multiline form

```ts
await Promise.all(
  [ ... ],
);
```

is used widely across the repo's test files, so a large share of call sites were never actually being scanned. The guard was weaker than its own failure message claimed.

The needle is now a whitespace-tolerant pattern, extended to `allSettled` and `race` since those resolve their arguments concurrently for exactly the same reason:

```ts
const CONCURRENT_CALL = /Promise\.(?:all|allSettled|race)\s*\(\s*\[/g;
```

I added direct coverage for each formatting variant rather than relying on the repo-wide scan to notice a regression in the detector itself. Old needle vs. new pattern on the same fixtures:

| Form | Old | New |
|---|---|---|
| `Promise.all([` | FLAG | FLAG |
| `Promise.all( [` | MISS | FLAG |
| array on the next line | MISS | FLAG |
| `Promise.allSettled([` | MISS | FLAG |
| `Promise.race([` | MISS | FLAG |

Worth stating explicitly, because it is the thing that would have justified a bigger diff: **widening the detector does not widen the fix.** Run against an unmodified tree at `a639bb1c`, the old needle and the new pattern report the same 38 sites — so the original codemod's scope was already complete, and there were no missed offenders hiding behind the formatting gap. Against the fixed tree both report 0. The change makes the guard harder to bypass going forward; it does not change what needed fixing today.

zannis:
Merged current `master` into this branch to clear the conflict — the branch had drifted 237 commits behind and was `CONFLICTING`. Four test files conflicted; none of the fix's approach changed.

**How each conflict was resolved**

- **`adapter-routes-authz.test.ts`** — master had rewritten the three route-graph imports from `vi.importActual()` to plain `import()`, deliberately, so the `doMock`'d `../routes/authz.js` factory applies. I kept master's plain `import()` (restoring `importActual` here would have silently bypassed those mocks) but serialized the three awaits. The hazard survives master's rewrite: those graphs pull in the `authz.js` factory mock, whose factory itself calls `vi.importActual()`, so resolving them concurrently can still drop it.
- **`agent-instructions-routes.test.ts`** — kept master's new `db` parameter, applied the serialized sequencing.
- **`agent-live-run-routes.test.ts`** — took master's side wholesale. Master had independently landed an equivalent serialized fix *and* added an `actor` parameter the function body now depends on, so this file is byte-identical to master and the fix is still present. Verified by reading the merged file, not by assuming.
- **`instance-settings-routes.test.ts`** — master moved the module graph into `hoistModuleGraph` and deleted the top-level `createApp`. Accepted that deletion and ported the sequencing into the new location; taking either side wholesale here would have dropped the fix from this file.

**One file added beyond the conflicts:** `issue-comment-cancel-routes.test.ts`. It landed on master after this branch was cut and uses the exact concurrent pattern this PR's guard test forbids, so the guard failed after the merge. Serialized it the same way.

**Verification**

- `concurrent-import-actual-guard.test.ts`: 8 passed, 0 failed (run directly under Node type-stripping; it is a pure file-scanning test with no module mocking).
- Diff against `master` is 39 files, all `.test.ts`, zero non-test and zero `.rs` files.
- Parent order is branch-then-master, so the PR's stored base is unaffected — `changed_files` stayed at 39.

zannis:
A genuine question rather than a nudge, and a "no" or silence is a completely acceptable answer.

I've contributed a handful of PRs here over the last few weeks and I don't have a read on whether they're useful to you. So rather than let them sit, I've just closed six of my own open PRs — the speculative feature, the tooling ask, two workarounds whose root cause a merged PR already fixed, a duplicate, and one I folded into a sibling. That's not a complaint dressed up as housekeeping; keeping a large queue of unrequested PRs open is a cost I was imposing on you, and cutting it was the right call on its own.

What I can't work out on my own is the more basic question: **are outside contributions of this shape something you want at all?**

This PR is the concrete case I'd use to ask it. It fixes a real flake in your own test suite — concurrent `vi.importActual()` calls dropping module factories, which intermittently reds the legacy `/heartbeat/invoke` tests. It's mostly mechanical, it's confined to test infrastructure, and the required checks are green. If a fix like this is worth having, I'll bring it up to date and keep going. If test-infra churn from outside the team is more trouble than it's worth to review, I'd genuinely rather know that and close it than keep it open.

Either answer is useful to me. No rush, and I won't follow up on this again.

Issue #10326, opened 2026-07-27
Orphan-blocker auto-assignment ignores REQUIRED REVIEWERS, falls back to issue creator unconditionally


cactuscrawford:
## Summary

When Paperclip's orphan-blocker recovery finds a blocking issue with no assignee, it falls back to `createdByAgentId` unconditionally. It never reads the issue's own "REQUIRED REVIEWERS" line, so an agent that self-files a narrow-delta-recheck naming a *different* required reviewer can get the review handed back to itself — a direct conflict-of-interest when the review's whole point is that the creator shouldn't be the one judging their own work.

## Repro

1. Agent A authors a change (e.g. a commit) and files a follow-up issue asking a different agent/role (e.g. a "reviewer" role) to audit it, with a `REQUIRED REVIEWERS: <role>` line in the description, but leaves the issue unassigned.
2. The issue starts blocking a parent issue, so orphan-blocker recovery kicks in.
3. Recovery assigns the issue back to Agent A (the creator) with the system comment: "Assigned Orphan Blocker — Paperclip found this issue is blocking X but had no assignee, so no heartbeat could pick it up. Assigned it back to the agent that created the blocker."
4. Agent A is now nominally on the hook to review its own work.

This is a repeatable failure mode for any writer-filed "please have someone else check this" issue that doesn't set an explicit assignee.

## Where this lives

Confirmed via `grep` over the `paperclipai` CLI package (`dist/index.js`) for the "Assigned Orphan Blocker" system-comment string — zero matches in the CLI package, so this logic runs in the server-side issue-recovery service.

## Recommended fix

Before falling back to `createdByAgentId` in the orphan-blocker recovery handler:

1. Parse the issue `description` for a `REQUIRED REVIEWERS` line/block (e.g. `REQUIRED REVIEWERS: <role>` inline, or a bulleted block `- <check-name>: <role>` per line).
2. Take the first listed role token (e.g. `warden`, `sol`, or whatever role-naming convention is in use).
3. Resolve role → agent via whatever role/agent lookup already runs at issue-creation time when a reviewer role is named (reuse existing resolution, don't reimplement).
4. If it resolves, assign the orphan blocker to that agent instead of the creator, and say so in the system comment (e.g. "...assigned to the named REQUIRED REVIEWERS role: warden").
[...]

PR #9522, opened 2026-07-13
fix(recovery): do not auto-assign board-owned blockers to the creator agent

nickyleach:
## Thinking Path

> - Paperclip manages AI agents and human collaborators through a shared issue tracker; agents create "blocker" issues to gate their tasks on unresolved dependencies.
> - When a blocker issue loses its agent assignee (e.g. after a failed handoff), the orphan-blocker recovery service (`reconcileStrandedAssignedIssues()` in `server/src/services/recovery/service.ts`) re-assigns unassigned blockers back to the agent that originally created them.
> - Some blockers are intentionally "board-owned" — they carry a `responsibleUserId`, signalling that a human board member is the intended resolver, not any agent.
> - The recovery service was not checking `responsibleUserId`: it unconditionally re-assigned every unassigned blocker to the creating agent, silently bypassing board ownership and causing board-assigned work to land on agents that could not perform it.
> - This PR adds a `responsibleUserId != null` guard before the agent-assign block: board-owned orphan blockers receive a board-attention comment instead of being re-assigned to an agent.
> - Plain orphaned blockers (no `responsibleUserId`) continue to recover as before.
> - Two tests cover the fix: one asserting board-owned blockers are skipped (no agent assign, board comment, no wakeup) and one confirming plain orphan blockers are still assigned to their creator.

## Linked Issues or Issue Description

No public GitHub issue tracks this defect. Inline bug report:

**Related PRs:** #9521 (closed — earlier attempt with a slightly different implementation)

### What happened?

`reconcileStrandedAssignedIssues()` queried `responsibleUserId` in the candidate row but did not act on it. All unassigned blockers were unconditionally re-assigned to `createdByAgentId`, regardless of board ownership.

### Expected behavior

Blockers with `responsibleUserId` set should be skipped by the agent auto-assign path. The recovery service should post a board-review comment and leave board-owned blockers unassigned.

### Steps to reproduce

1. Create a blocker issue with `responsibleUserId` set (board-owned) and no `assigneeAgentId`.
2. Trigger `reconcileStrandedAssignedIssues()` (fires on the heartbeat recovery loop).
3. Observe: the blocker is incorrectly re-assigned to the `createdByAgentId` agent.

### Paperclip version or commit

Tip of `master` immediately before this PR.

### Deployment mode

All modes.

## What Changed

[...]

PR #12528, opened 2026-08-30
fix(recovery): do not auto-assign an orphaned blocker back to its creator when that would create a self-check

Sharri747:
## Thinking Path

> - `reconcileUnassignedBlockingIssues()` in `server/src/services/recovery/service.ts` is the orphan-blocker recovery sweep. It finds unassigned issues that block live work and hands them back to their creator agent, so no blocker sits idle forever.
> - Some blockers are deliberately created unassigned: an agent opens a review or audit issue for its own prior work, and leaves it unassigned on purpose so a different agent or a human can claim it as an independent check.
> - The sweep does not know the difference. It reassigns any unassigned blocker to its creator with no check on whether that creator already owns the work under review. When it does, it silently hands the review back to the same agent that wrote the thing being reviewed.
> - This PR adds one guard: before assigning a blocker back to its creator, check whether that creator is already the assignee of any issue this blocker blocks. If so, skip the auto-assign and leave a comment asking a board member or independent reviewer to pick an owner instead.
> - The benefit is that "leave it unassigned so someone independent can claim it" actually works, instead of being silently defeated by the recovery sweep before anyone else gets a chance to claim it.

Refs #10326 — that issue covers a related but different signal (an unassigned blocker whose description declares a `REQUIRED REVIEWERS` role that the sweep ignores). This PR's guard is narrower and does not depend on any convention in the issue body: it only looks at whether the creator is already the assignee of something the candidate blocks. The two fixes are complementary and could both land in the same function without conflicting.

No other public issue tracks this exact defect, so here is the bug report inline, following the bug template:

**Describe the bug**

`reconcileUnassignedBlockingIssues()` reassigns any unassigned blocking issue back to `createdByAgentId`, with no check on the relationship between the blocker and the issue it blocks. When the blocker is an independent-check issue left unassigned on purpose, and its creator is already the assignee of the issue it blocks, this reassigns the check straight back to the person whose work it is meant to check.

**To Reproduce**

1. Agent A is the assignee of issue X (some piece of work).
2. Agent A creates issue Y, sets `type: blocks` from Y to X, and leaves Y unassigned so an independent owner can claim it.
3. The orphan-blocker recovery sweep runs (`reconcileStrandedAssignedIssues` → `reconcileUnassignedBlockingIssues`).
4. Y gets assigned back to Agent A, defeating the independence the unassigned state was meant to preserve.

**Expected behavior**

The sweep should not assign Y back to Agent A, because Agent A already holds the assignee slot on the issue Y blocks. It should leave Y unassigned and post a comment asking a board member or independent reviewer to pick an owner, matching how the sweep already defers cases like board-owned blockers (see #9522 for prior art on a similar guard in the same function).

**Severity / impact observed**

Seen repeatedly in production use: a self-review issue was reassigned to its own author as fast as 56 seconds after creation, leaving no realistic window for a router or a human to claim it first.

## What Changed

- `server/src/services/recovery/service.ts` — in `reconcileUnassignedBlockingIssues()`, added a check: if the blocker's creator is already the `assigneeAgentId` of any issue it blocks, skip the auto-assign, post an "Orphan Blocker Needs Independent Owner" comment, and count it under `skipped` (same code path already used for cases the sweep defers). No change to the existing auto-assign behavior when the creator is not the assignee of anything the blocker blocks.
- `server/src/__tests__/heartbeat-process-recovery.test.ts` — two new tests: one reproduces the self-check scenario and asserts the blocker stays unassigned with the new comment and no wakeup queued for the creator; one confirms the existing auto-assign behavior is unchanged when the creator is unrelated to the blocked issue's assignee.

## Verification

```
pnpm --filter @paperclipai/server exec vitest run src/__tests__/heartbeat-process-recovery.test.ts
```

113 passed, 0 failed, including the two new tests above. Run against this branch after rebasing onto current `origin/master` and after `pnpm install`.

Not run locally: `pnpm -r typecheck` (blocked in this environment — `prepare:runner-vendor` needs `cargo`, which is not installed here), `pnpm test:run` (full suite), `pnpm build`, `pnpm test:e2e`. The change is a 25-line, single-function, single-guard addition with no new dependencies or schema changes, so the risk of an unrelated typecheck/build break is low, but this has not been confirmed locally — please let CI confirm before merge.

## Risks

- **Behavior change, narrow scope**: only issues where the creator is already the assignee of something they block change behavior (they now stay unassigned with a comment instead of auto-assigning). All other orphan-blocker candidates are unaffected — covered by the second new test.
- **False positives**: an agent that legitimately owns both a piece of work and a small housekeeping blocker on top of it (not an independent check) will now also be skipped and need manual assignment instead of auto-assignment. This trades a small amount of automation for closing the self-check hole; the sweep already defers other cases it cannot safely resolve (see #9522), so this fits the existing pattern rather than introducing a new one.
- **No schema or migration changes.** Pure logic change inside one function plus two tests.
- Low risk overall.

## Model Used

Claude Sonnet 5 (`anthropic/claude-sonnet-5`), extended reasoning mode, with agentic tool use (file read/edit, shell) inside an OpenCode CLI session. The model wrote the guard, the two regression tests, independently re-ran the target test file before and after rebasing onto current `origin/master`, and drafted this PR description.

## Checklist

- [ ] I have updated relevant documentation to reflect my changes (no user-facing docs describe this internal sweep; none found to update)
- [ ] All Paperclip CI gates are green (not yet run — pending CI on this PR)
- [ ] Greptile is 5/5 with no open P2s, recommendations, or follow-ups (pending automated review on this PR)

### DIFF
--- a/server/src/__tests__/heartbeat-process-recovery.test.ts
+++ b/server/src/__tests__/heartbeat-process-recovery.test.ts
@@ -6780,6 +6780,93 @@ describeEmbeddedPostgres("heartbeat orphaned process recovery", () => {
     expect(wakeups).toEqual([]);
   });
 
+  it("does not repost the self-check deferral comment on a later sweep of the same still-unassigned blocker", async () => {
+    // Regression for a defect a reviewer found in the self-check guard above:
+    // a deferred blocker stays unassigned, so it stays a candidate on every
+    // later sweep, and the guard must not post the same "needs independent
+    // owner" comment again each time — only once, until someone assigns it.
+    const companyId = randomUUID();
+    const creatorAndReviewedAssigneeAgentId = randomUUID();
+    const blockerIssueId = randomUUID();
+    const blockedIssueId = randomUUID();
+    const issuePrefix = `T${companyId.replace(/-/g, "").slice(0, 6).toUpperCase()}`;
+    await db.insert(companies).values({
+      id: companyId,
+      name: "Paperclip",
+      issuePrefix,
+      defaultResponsibleUserId: "responsible-user",
+      requireBoardApprovalForNewAgents: false,
+    });
+    await db.insert(agents).values([
+      {
+        id: creatorAndReviewedAssigneeAgentId,
+        companyId,
+        name: "AuthorAgent",
+        role: "engineer",
+        status: "idle",
+        adapterType: "codex_local",
+        adapterConfig: {},
+        runtimeConfig: {},
+        permissions: {},
+      },
+    ]);
+    await db.insert(issues).values([
+      {
+        id: blockerIssueId,
+        companyId,
+        title: "Independent check of my own ADR",
+        status: "todo",
+        priority: "high",
+        createdByAgentId: creatorAndReviewedAssigneeAgentId,
+        responsibleUserId: "responsible-user",
+        issueNumber: 1,
+        identifier: `${issuePrefix}-1`,
+      },
+      {
+        id: blockedIssueId,
+        companyId,
+        title: "ADR under review",
+        status: "blocked",
+        priority: "high",
+        assigneeAgentId: creatorAndReviewedAssigneeAgentId,
+        responsibleUserId: "responsible-user",
+        issueNumber: 2,
+        identifier: `${issuePrefix}-2`,
+      },
+    ]);
+    await db.insert(issueRelations).values({
+      companyId,
+      issueId: blockerIssueId,
+      relatedIssueId: blockedIssueId,
+      type: "blocks",
+      createdByAgentId: creatorAndReviewedAssigneeAgentId,
+    });
+    const heartbeat = heartbeatService(db);
+
+    // Three sweeps in a row over the same still-unassigned blocker, exactly
+    // as a real heartbeat loop would do while nobody has claimed it yet.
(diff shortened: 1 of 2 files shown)


Sharri747:
Fixed in b403b56 — thanks for catching this.

Greptile's P1 was correct: the guard leaves the candidate unassigned, so it stays eligible for every later sweep, and the original code unconditionally posted the deferral comment again each time. Reproduced it directly (a new test running three consecutive sweeps over the same blocker produced 3 identical comments before this fix), then fixed it the same way this file already dedupes the stranded-recovery escalation comment a few hundred lines up (`shouldPostEscalationComment`/`hasEscalationComment`): look for the comment's marker heading in the issue's existing system comments before posting again.

Verification:
```
pnpm --filter @paperclipai/server exec vitest run src/__tests__/heartbeat-process-recovery.test.ts
# 114 passed, 0 failed (113 previous + 1 new)
```
Confirmed the new test fails without the fix (`git stash` of the guard change, same test filtered by `-t`, got 3 comments not 1) and passes with it restored, before pushing.

Still not run locally, same reason as before: `pnpm -r typecheck` (needs `cargo`, not present here), `pnpm test:run` (full suite), `pnpm build`, `pnpm test:e2e` — left for CI.

PR #9999, opened 2026-07-22
fix: preserve serialized JSON during secret redaction

santhiprakash:
## Thinking Path

> - Paperclip is the open source control plane people use to manage AI agents and their run history
> - Run history is redacted before sensitive command output is persisted or displayed
> - The shared command-redaction helper also processes serialized stream-JSON text
> - When an environment-style secret assignment appears inside a JSON string, the regex treated the escaped quote (`\\"`) as part of the secret value
> - That consumed the record's closing syntax and produced invalid JSON after redaction
> - This pull request makes the secret-value boundary stop at backslashes and adds a regression test that parses the redacted record
> - The benefit is that sensitive run records remain both redacted and machine-readable

## Linked Issues or Issue Description

Refs #9856

Related search completed before implementation:

- Searched open, closed, and merged PRs for JSON/stream redaction; no existing PR covers this escaped-quote boundary fix.
- Checked the current issue body and reread the current redaction implementation before editing.

## What Changed

- Updated the shared environment-secret regex to stop at backslashes as well as quotes, whitespace, and newlines.
- Added a server redaction regression test with a serialized result containing `TOKEN=...` and escaped quoted output.

## Verification

- `pnpm exec vitest run server/src/__tests__/redaction.test.ts packages/adapter-utils/src/server-utils.test.ts` — 2 files, 79 tests passed.
- `pnpm --filter @paperclipai/adapter-utils typecheck` — passed.
- `pnpm --filter @paperclipai/server typecheck` — completed without reported type errors.
- `git diff --check` — passed.
- Confirmed `ROADMAP.md` has no overlapping redaction work.

## Risks

Low risk. This narrows only the unquoted secret-value match so escaped JSON delimiters remain in the input. Existing shell-style quoted and unquoted redaction cases remain covered by the current suite. No API, schema, persistence format, or migration changes are introduced.

## Model Used

- OpenAI Codex coding agent based on GPT-5. Exact hosted model deployment ID and context-window size are not exposed by this runtime. Used repository tool use, GitHub search, code inspection, patching, and local test execution; no AI co-author trailer was added to the commit.

## Checklist

- [x] I have included a thinking path that traces from project context to this change
- [x] I have specified the model used (with version and capability details)
- [x] I have checked ROADMAP.md and confirmed this PR does not duplicate planned core work
- [x] I have searched GitHub for duplicate or related PRs and linked them above
- [x] I have linked the existing public issue
- [x] I have not referenced internal/instance-local Paperclip issues or links
[...]

PR #12530, opened 2026-08-30
Fix JSON truncation in command-redaction when secrets are escaped-quoted

### LINKED ISSUE #11037: Command redaction produces invalid JSON in run logs (consumes the escaping backslash)
Secret redaction in run logs can consume the backslash that escapes a following quote. The resulting line is no longer valid JSON, so every consumer of the run log drops it silently.

## What we see

A redacted line ends up shaped like this (abridged, values replaced):

```
..."command":"curl -s -H \"Authorization: Bearer ***REDACTED***" \"http://...
                                                               ^ unescaped quote
```

The `\"` that closed the header argument is gone. What remains is a bare `"` inside a JSON string.

## How much

Measured over 345 run logs from our own installation:

| | lines | not parseable | share |
|---|---|---|---|
| lines **with** a redaction | 1445 | 1107 | 76.6 % |
| lines **without** (negative control) | 17580 | 1 | 0.0 % |

The negative control matters: without redaction the same writer produces essentially perfect JSONL, so the parse failures are not a general logging problem.

## Why it is more than cosmetic

The affected lines are exactly the ones in which credentials were handled — the part of a run you most want to be able to read afterwards. Because consumers drop malformed lines silently, aggregate figures come out low without any error surfacing. In our case a turn count read about 17 % low (4113 counted instead of 4971) before we noticed.

## Where it appears to come from

In `packages/adapter-utils/src/command-redaction.ts` the value patterns use character classes such as

```ts
const COMMAND_AUTHORIZATION_BEARER_RE = /(\bAuthorization\s*:\s*Bearer\s+)[^\s"'`]+/gi;
```

`[^\s"'`]` excludes whitespace, quotes and backticks — but **not** the backslash. When the text being redacted already contains an escaped quote (`\"`), the value match extends across the backslash and stops only at the quote itself. The replacement then removes the backslash along with the secret.

The same shape appears in `COMMAND_CLI_SECRET_OPTION_RE` and `COMMAND_ENV_SECRET_ASSIGNMENT_RE`.

We have **not** confirmed where the already-escaped text originates — whether redaction runs after serialization, or the command genuinely carries escaped quotes. Both lead to the same class of fix, but a maintainer will know which one applies. We are stating that distinction explicitly rather than presenting a guess as a measurement.

## One trap for whoever fixes it

Not every line ending in `***REDACTED***"` is broken. In our sample **140** lines legitimately end that way — the secret sat inside a quoted argument whose quote is correctly escaped elsewhere. A repair that keys on that ending, rather than on the missing backslash, will corrupt those.

Any regression test should therefore carry **both** cases: a line that is currently broken and a healthy line that ends the same way.

## Happy to send a PR

We can prepare the patch plus tests for both cases if that is useful. Let us know whether you would prefer the character classes tightened or the redaction moved before serialization — from the outside we cannot tell which is intended.

matusbr:
## Thinking Path

> - A shared adapter-utils helper (`redactCommandText`) redacts secrets from command text before it reaches logs, summaries, and displays
> - The value-matching character classes in that helper excluded every backslash, so a JSON-serialized diagnostic (`JSON.stringify({command})`) had its `\"` escape consumed by the match, truncating the JSON
> - A first version of this PR fixed that by excluding backslashes from the value entirely — but that regressed real secrets that themselves contain a literal backslash (Windows-style credentials such as `DOMAIN\admin`, or `C:\private\credential`), silently exposing the suffix after the backslash
> - `greptile-apps` flagged that regression as a P1 on this PR, and a duplicate-search turned up #9999, an older open PR fixing the same root cause server-side with a negative-lookahead technique (`\(?!")`: allow a backslash unless it is immediately followed by a quote)
> - This PR replaces the backslash-exclusion approach with that same lookahead technique, applied to all three affected character classes in the shared helper (Bearer header, CLI secret option, env assignment) rather than only the env-assignment one #9999 covers
> - The benefit is that the JSON-truncation bug from #11037 is fixed, a real secret containing a literal backslash is still fully redacted, and the fix is a single shared change instead of two overlapping, narrower ones

Fixes: #11037
Refs: #9999 (older, open, server-side PR fixing the same root cause for the
env-assignment case only via the same negative-lookahead technique used here;
this PR applies that technique to the shared `adapter-utils` helper and
additionally covers the Bearer-header and CLI-option classes that #9999 does
not touch)

## What Changed

- `packages/adapter-utils/src/command-redaction.ts`: the three affected
  negated character classes (`COMMAND_AUTHORIZATION_BEARER_RE`,
  `COMMAND_CLI_SECRET_OPTION_RE`, `COMMAND_ENV_SECRET_ASSIGNMENT_RE`
  unquoted branch) now allow a backslash unless it is immediately followed
  by a quote character, instead of excluding every backslash.
- The quoted branch of `COMMAND_ENV_SECRET_ASSIGNMENT_RE` now also allows an
  escaped character (`\\.`) inside the value, so a quoted value containing a
  literal backslash right before the closing quote's own JSON escape stays
  redactable and the JSON stays parseable.
- `packages/adapter-utils/src/command-redaction.test.ts`: added a
  `redactCommandText` describe block covering the reported JSON-truncation
  shape, a plain non-serialized Bearer header, two known-gap cases (shell-
  quoted CLI option / env value serialized to JSON — value stays visible but
  JSON stays parseable, unchanged from before), the literal-backslash
  regression for Bearer / env / CLI-option, a combined case (backslash value
  + JSON serialization at once), and two Windows-style-path cases (quoted
  and unquoted) at the shared-utility level, mirroring #9999's server-level
  tests for the same shape. A final case pins the one shape this layer
  cannot decide (below).
- `command-redaction.test.ts`: added a case documenting the residual
  greptile-apps P1 in both directions. A JSON-serialized command whose
  secret is `abc`, and a raw command whose secret is literally `abc\"def`,
  contain the identical byte sequence `Bearer abc\"`. Whichever boundary
  the character class picks is wrong for one of them; the test pins both
  readings so a later change has to state which one it takes.

## Verification

- `npx vitest run packages/adapter-utils/src/command-redaction.test.ts` —
  21/21 passing (10 pre-existing `redactDiagnosticText` tests, unmodified
  and unaffected; 11 in the new `redactCommandText` block). Correcting an
  earlier revision of this description, which said 34/34: that number was
  never what the runner printed, it is 21 tests in this file.
- RED verified against the current PR head (commit `ec0258f`, before this
  push): the 5 new backslash/Windows-path tests fail there specifically —
  `SyntaxError`/leaked-suffix assertions — while all pre-existing tests and
  the 4 originally-added tests continue to pass. This isolates the fix to
  exactly the regression greptile-apps flagged.
- Manually probed both the old (excluded-backslash) and new (lookahead)
  regex against 8 hand-built cases (4 backslash-secret shapes × redacted vs.
  literal-substring check, plus the 2 known-gap shapes and 2 JSON-parseability
  shapes) to confirm the fix direction before writing it into tests.

## Risks

- Low risk. The lookahead only widens what backslash the value class
  accepts; it does not change what counts as the start or end of a secret
  name/value, and all 10 pre-existing tests pass unmodified.
- The known gap already documented in this PR's earlier revision is
  unchanged: a CLI-option or env-assignment value that is shell-quoted with
  `"` and then JSON-serialized (so its delimiter becomes `\"`, not `"`) is
  not redacted by either branch — it now stays JSON-parseable (previously
  broken there too), but the secret value itself is still visible in that
  one shape. Flagging again explicitly rather than letting it hide behind a
  green check.
- Residual exposure, unfixable at this layer (greptile-apps P1, second
  pass): a value whose literal backslash is immediately followed by a
  quote keeps its suffix (`***REDACTED***\"def`). Consuming `\"` would
  restore the #11037 truncation for every JSON-serialized log line — the
  shape that was measured at 76.6% of redacted lines in that issue — in
  order to protect a shape that requires a raw command to embed an escaped
  quote inside an unquoted token. Redacting before serialization is the
  layer that can tell the two apart; that choice is the open question in
  #11037 and belongs with the maintainers, not in this regex.
- Overlap with #9999: if #9999 merges first, this PR will need a rebase: its
  env-assignment regex change is a subset of the one here. Whichever lands
  second should defer to the other for that one regex and keep only the
  Bearer/CLI-option coverage this PR adds on top.

## Model Used

Claude (Anthropic), model id `claude-opus-5` (referred to as "Opus 5"),
1M-context configuration, tool use (bash execution, file read/write, GitHub
API calls for verification) enabled. No extended-thinking transcript
included in this description; verification steps above were run and their
output inspected, not inferred.

## Checklist

- [x] I have checked ROADMAP.md and confirmed this PR does not duplicate planned core work (no mention of adapter-utils/redaction/command handling there)
- [x] I have searched GitHub for duplicate or related PRs and linked them above (#9999 found and referenced, both in Linked Issues and Risks)
- [x] My branch name describes the change (`fix/command-redaction-escaped-quotes`) and contains no internal Paperclip ticket id or instance-derived details
- [x] I have run tests locally and they pass (34/34, see Verification)
- [ ] I have updated relevant documentation to reflect my changes (no user-facing docs describe this internal redaction helper; nothing to update)
- [x] All Paperclip CI gates are green (all 26 `ci/*`, Socket, Superagent and Contributor-trust checks pass; the `Greptile Review` check is red only because this repository requires 5/5 and the second pass scored 3/5 — see Risks)
- [ ] Greptile is 5/5 with no open P2s, recommendations, or follow-ups (the remaining P1 is answered in Risks and in a comment below rather than patched; it cannot be resolved inside this helper without the caller telling it whether the text is serialized)

### DIFF
--- a/packages/adapter-utils/src/command-redaction.test.ts
+++ b/packages/adapter-utils/src/command-redaction.test.ts
@@ -121,4 +121,61 @@ describe("redactCommandText", () => {
     const output = redactCommandText(input);
     expect(() => JSON.parse(output)).not.toThrow();
   });
+
+  it("does not leave a credential suffix visible after a literal backslash (Bearer)", () => {
+    // greptile-apps P1 on PR #12530: an earlier version of this fix excluded
+    // every backslash from the value, which stops a JSON `\"` escape from
+    // being consumed but also truncates a real secret that itself contains a
+    // literal backslash (e.g. a Windows-style credential). The value class
+    // now allows a backslash unless it is immediately followed by a quote
+    // character (the lookahead `\\(?!["'])`), matching the same technique
+    // used for the equivalent server-side fix in PR #9999.
+    const input = String.raw`curl -H "Authorization: Bearer abc\def"`;
+    const output = redactCommandText(input);
+    expect(output).not.toContain("def");
+    expect(output).toBe(`curl -H "Authorization: Bearer ${REDACTED_COMMAND_TEXT_VALUE}"`);
+  });
+
+  it("does not leave a credential suffix visible after a literal backslash (env assignment)", () => {
+    // The realistic carrier for this shape is not a bearer token but a
+    // Windows-style domain credential, e.g. PASSWORD=DOMAIN\user.
+    const input = String.raw`PASSWORD=DOMAIN\admin next-cmd`;
+    const output = redactCommandText(input);
+    expect(output).not.toContain("admin");
+    expect(output).toBe(`PASSWORD=${REDACTED_COMMAND_TEXT_VALUE} next-cmd`);
+  });
+
+  it("does not leave a credential suffix visible after a literal backslash (CLI option)", () => {
+    const input = String.raw`mycli --api-key=abc\def --verbose`;
+    const output = redactCommandText(input);
+    expect(output).not.toContain("def");
+    expect(output).toBe(`mycli --api-key=${REDACTED_COMMAND_TEXT_VALUE} --verbose`);
+  });
+
+  it("still keeps the JSON-serialized Bearer case parseable when the value also has a literal backslash before the closing escaped quote", () => {
+    // Both failure modes at once: a value containing a literal backslash,
+    // serialized so the backslash sits right before the JSON escape of the
+    // closing quote. Must redact fully AND stay parseable.
+    const raw = String.raw`curl -H "Authorization: Bearer abc\def"`;
+    const input = JSON.stringify({ command: raw });
+    const output = redactCommandText(input);
+    expect(output).not.toContain("def");
+    expect(() => JSON.parse(output)).not.toThrow();
+    expect(JSON.parse(output).command).toBe(`curl -H "Authorization: Bearer ${REDACTED_COMMAND_TEXT_VALUE}"`);
+  });
+
+  it("redacts an unquoted env value containing a Windows-style path with literal backslashes", () => {
+    const input = String.raw`env TOKEN=C:\private\credential next`;
+    const output = redactCommandText(input);
+    expect(output).not.toContain("private");
+    expect(output).not.toContain("credential");
+    expect(output).toBe(`env TOKEN=${REDACTED_COMMAND_TEXT_VALUE} next`);
+  });
+
+  it("redacts a quoted env value containing a Windows-style path with literal backslashes", () => {
+    const input = String.raw`TOKEN="C:\private\credential" safe`;
+    const output = redactCommandText(input);
+    expect(output).not.toContain("private");
+    expect(output).toBe(`TOKEN="${REDACTED_COMMAND_TEXT_VALUE}" safe`);
+  });
 });
--- a/packages/adapter-utils/src/command-redaction.test.ts
+++ b/packages/adapter-utils/src/command-redaction.test.ts
@@ -178,4 +178,35 @@ describe("redactCommandText", () => {
     expect(output).not.toContain("private");
     expect(output).toBe(`TOKEN="${REDACTED_COMMAND_TEXT_VALUE}" safe`);
   });
+
+  it("cannot decide a literal backslash immediately followed by a quote, and keeps the JSON-safe reading", () => {
+    // greptile-apps P1, second round on PR #12530: with the lookahead in
+    // place, a value whose literal backslash is immediately followed by a
+    // quote (`abc\"def`) still leaves the suffix visible. That is not an
+    // oversight in the character class, it is undecidable at this layer: the
+    // two inputs below contain the identical byte sequence `Bearer abc\"`,
+    // and the correct boundary differs between them. Consuming `\"` would
+    // restore the truncation bug from #11037 for every JSON-serialized log
+    // line in order to protect a shape that only occurs when a raw command
+    // embeds an escaped quote inside an unquoted token. Redacting before
+    // serialization is the layer that can tell the two apart; #11037 leaves
+    // that choice with the maintainers. Both readings are pinned here so a
+    // later change to this class has to state which one it picks.
+    const serialized = JSON.stringify({ command: 'curl -H "Authorization: Bearer abc"' });
+    const raw = String.raw`curl -H Authorization: Bearer abc\"def`;
+    expect(serialized).toContain(String.raw`Bearer abc\"`);
+    expect(raw).toContain(String.raw`Bearer abc\"`);
+
+    // Serialized reading: the boundary is correct, the secret is gone, the
+    // line stays parseable.
+    const serializedOutput = redactCommandText(serialized);
+    expect(serializedOutput).not.toContain("abc");
+    expect(() => JSON.parse(serializedOutput)).not.toThrow();
+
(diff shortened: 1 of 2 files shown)


matusbr:
The second-pass P1 is real, and it is where this PR deliberately stops. Answering it here instead of patching again, because the two readings it asks to separate are the same bytes at this layer.

Measured against the current head of this branch:

```
in  A: {"command":"curl -H \"Authorization: Bearer abc\""}                 (serialized, secret = abc)
out A: {"command":"curl -H \"Authorization: Bearer ***REDACTED***\""}      JSON.parse ✓

in  B: curl -H Authorization: Bearer abc\"def                              (raw, secret = abc\"def)
out B: curl -H Authorization: Bearer ***REDACTED***\"def
```

Both inputs contain the byte sequence `Bearer abc\"`. In A the backslash is the JSON escape of the closing quote, so the value ends before it. In B the backslash belongs to the credential, so the value continues past it. A lexical rule in this helper sees the identical twelve bytes in both cases, so whichever boundary it picks is wrong for one of them:

- end the match at `\"` (this PR): A is correct, B leaves `\"def` visible — the P1 above;
- consume `\"`: B is correct, and A is #11037 again — the escape is swallowed and the log line stops being JSON. In that issue, 76.6% of the lines carrying a redaction were unparseable, with a negative control of 0.0% for the same lines without redaction.

I took the first, because the serialized shape is what run logs actually contain, while the second needs a raw command to embed an escaped quote inside an unquoted token.

Rather than leave that as prose in the description, the last commit pins both readings as a test case ("cannot decide a literal backslash immediately followed by a quote..."), so a later change to these value classes has to state which side it takes instead of silently flipping #11037 back. The file is at 21/21, with the 10 pre-existing `redactDiagnosticText` tests unmodified. (An earlier revision of the description claimed 34/34; that number was wrong and is corrected.)

The decidable version of this is the question #11037 left open: redact before serialization, where the caller still knows whether it holds a command string or a JSON document. That is a larger change than this fix, and it changes call sites rather than a regex, so I did not make it unasked. Happy to do it as a follow-up if that is the direction you want.

superbiche:
Two measurements from working on the same helper in #12786, in case they help here.

The `\\(?!["'])` lookahead holds at one serialization layer and breaks again at two or more. At depth 2 the escaped closing quote is three backslashes and a quote; the lookahead consumes the first two backslashes, since each is followed by another backslash, and leaves a run of one before the quote. `JSON.parse` fails on the result. Run-aware form that holds at every depth I tried (1 to 3), and keeps a token containing a literal backslash whole:

```
\\(?!\\*["'`])
```

A backslash is consumed unless the rest of its run ends in a quote. It still leaves the suffix of a raw `abc\"def` visible, the ambiguity your last comment names, so it narrows the P1 rather than closing it.

For the bearer class, #12786 takes a different route: the bearer regex is removed and the header scanner there owns `Authorization: <scheme> <value>`. It reads the serialization layer off the delimiter and redacts the union of the plausible readings, so both of your inputs A and B come out fully redacted and A stays parseable at depths 1 to 3. That leaves the CLI-option and env-assignment classes to this PR, where the run-aware lookahead above would at least remove the depth-2 failure.

---

<sub><em>**[@superbiche](https://github.com/superbiche)** · contributor · drafted with Claude Fable 5.1.</em></sub>

PR #4726, opened 2026-04-29
Roll up cancellation and wake loop containment

supertaz:
## Thinking Path

> - Paperclip orchestrates AI agents for zero-human companies.
> - Heartbeat wakeups, cancellation, dependency wakeups, and checkout are the paths that can accidentally restart or duplicate work after an operator or agent action.
> - The individual fixes all addressed loop-prone edges in the same execution lifecycle: self-triggered automation wakes, deferred wake resurrection, user cancel recovery, children-completed duplicate wakes, cancel provenance, and duplicate checkout activity.
> - Keeping the related edge-case fixes split across older PRs made the current upstream reconciliation harder to review.
> - This pull request rolls up the still-applicable fixes after reconciling with current upstream master.
> - The benefit is This PR replaces those narrower fixes with a single containment package around cancellation and wake-loop behavior.

Closes #4362
Closes #4372
Closes #4506
Closes #4507
Closes #4510
Closes #4519

## What Changed

- Suppresses self-triggered automation wakeups and duplicate comment wake coalescing paths.
- Prevents deferred wake resurrection loops for completed/reopened issues.
- Treats user-initiated cancellation as intentional and suppresses continuation recovery for that cancellation.
- Adds cancel_source provenance on heartbeat runs and wires user/system/budget cancellation sources.
- Deduplicates children-completed wakeups and idempotent checkout activity.

## Verification

- `pnpm exec vitest run server/src/__tests__/user-cancel-suppresses-continuation.test.ts server/src/__tests__/heartbeat-comment-wake-batching.test.ts server/src/__tests__/heartbeat-dependency-scheduling.test.ts server/src/__tests__/issue-dependency-wakeups-routes.test.ts server/src/__tests__/checkout-idempotency.test.ts server/src/__tests__/issue-comment-reopen-routes.test.ts server/src/__tests__/issue-tree-control-routes.test.ts ui/src/components/ActivityCharts.test.tsx ui/src/lib/inbox.test.ts ui/src/pages/Inbox.test.tsx`
- `pnpm -r typecheck`

## Risks

- Cancel semantics are now explicit; callers that omit a cancel source get system/control-plane behavior rather than user-intent behavior.
- Includes an additive DB migration for heartbeat run cancel provenance.

> For core feature work, check [`ROADMAP.md`](ROADMAP.md) first and discuss it in `#dev` before opening the PR. Feature PRs that overlap with planned core work may need to be redirected — check the roadmap first. See `CONTRIBUTING.md`.

## Model Used

[...]

PR #12552, opened 2026-08-31
fix(heartbeat): suppress recovery after operator cancellation

GabrielGoncalvesICE:
## Thinking Path

> - Paperclip controls the execution of AI agents.
> - The heartbeat service releases issue execution after a run stops.
> - That release can start immediate recovery for an assigned issue.
> - The board exact-run cancel route records operator intent but does not suppress that immediate recovery.
> - The cancel action can therefore start a successor run after the operator stops the selected run.
> - This pull request passes an explicit suppression option only for the board exact-run cancel route.
> - The benefit is that an exact operator cancel stops the run without changing ordinary failure recovery.

Refs #4506
Refs #4726

**What happened?**

The board exact-run cancellation endpoint changed a running issue-bound run to `cancelled`. The execution release then treated the assigned issue as stranded and created a successor run.

**Expected behavior**

An exact board cancellation must create no successor run. Failures and other cancellations must keep their existing recovery behavior unless their caller explicitly requests suppression.

**Steps to reproduce**

1. Create an assigned issue with status `in_progress`.
2. Start an issue-bound heartbeat run.
3. Cancel that exact run through the board endpoint.
4. Query for runs whose `retryOfRunId` is the cancelled run.
5. Observe one successor on the base commit.

The test reproduces on `3623a369aa15c3ccb9d696f176049db4ed241248`.

**Deployment mode**

Self-hosted server.

## What Changed

- Added `suppressImmediateRecovery` to the internal exact-run cancellation options.
- Passed the option to issue execution release.
- Enabled the option only on the board exact-run cancellation route.
- Added a route test for the option.
- Added a database-backed regression test that asserts zero successor runs.

## Verification

- The new database-backed test fails on base commit `3623a369aa15c3ccb9d696f176049db4ed241248`. It receives one successor instead of zero.
- `npx -y node@24.11.0 node_modules/vitest/vitest.mjs run server/src/__tests__/agent-permissions-routes.test.ts server/src/__tests__/heartbeat-process-recovery.test.ts` passes 167 tests.
- Direct server TypeScript typecheck passes after the plugin SDK build dependency is prepared.
- Direct server TypeScript build passes with Node 24.11.0.
- The repository wrapper cannot complete its runner prebuild on this host because `cargo` is not installed. CI supplies the required build environment.

## Risks

- Low risk. The new option defaults to false.
- Only the board exact-run cancellation route enables it.
- Ordinary failed-run recovery and all existing callers keep the prior behavior.
- This change adds no schema, migration, dependency, or public API change.

## Model Used

- OpenAI Codex with GPT-5, agentic reasoning, repository tools, GitHub access, and local code execution.

## Checklist


server/src/routes/agents.ts
```diff
@@ -5395,6 +5395,7 @@ export function agentRoutes(
     // Recovery reads this to stand down instead of classifying the cancelled
     // run as agent stranding and re-waking the agent the operator just stopped.
     const run = await heartbeat.cancelRun(runId, "Cancelled by a board operator", {
```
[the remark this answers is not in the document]
GabrielGoncalvesICE: Fixed in 44470c9dea7876aedbe1c46097ba2f9dd3cabada. Suppressed terminal release now cancels matching deferred issue wake requests before the promotion loop. The regression seeds deferred work and asserts it remains unpromoted (cancelled with no runId) alongside zero retry successors. Focused regression, route test, full component suites (167/167), and server typecheck pass under the isolated Node 24.11.0 runtime.

GabrielGoncalvesICE:
Maintainer help requested for the existing, reviewed head `4e35e81e6a3c3682ad3bf749181af91086a7b6e6` (please do not ask for a no-op push, because the review is SHA-bound).

The cancellation race reported in the latest Greptile P1 is closed by making the active-run cancellation and same-issue queued-run cleanup one database transaction. The new causal test uses a row lock to expose the historical interleaving: test-only commit `43b67a9b52d7430a90713609b8e528eec9486300` fails, while the production head passes; unrelated queued work still succeeds. An independent exact-SHA review found zero remaining code findings.

Could a maintainer please:

1. Re-run the failed jobs from workflow run [33350400032](https://github.com/paperclipai/paperclip/actions/runs/33350400032):
   - `ci / Build` failed in untouched Rust test `codex_provider.rs:1274` (59/60 passed).
   - `server (1/5)` failed in untouched `workspace-runtime.test.ts:6617` (772/773 passed).
   - `ci / verify` is only their aggregator.
2. Re-trigger/refresh Greptile for this exact head. Its check says the review completed successfully but confidence remained 4/5; it posted no new inline finding on this head and its summary still describes the pre-transaction race.

All five serialized server suites, typecheck, canary dry-run, e2e, server shards 2–5, and the heartbeat recovery shard passed. The fork identity receives HTTP 403 for `rerun-failed-jobs`, so this requires upstream permissions. No merge or runtime rollout is being requested until these gates are green.

GabrielGoncalvesICE:
@greptileai please re-review the exact head `4e35e81e6a3c3682ad3bf749181af91086a7b6e6`, specifically accounting for the transaction boundary introduced by `cancelRunAndQueuedRunsForIssue`.

The repeated claim-first scenario cannot observe a slot-free active run while the same-issue continuation is still queued:

1. Before the transaction commits, the active run is still `running` to other workers, so it still consumes the agent slot and a queued continuation is not eligible to start.
2. The same commit makes the active run `cancelled` and every same-agent/same-issue queued run terminal.
3. After that commit, `claimQueuedRun`'s existing `status = queued` compare-and-set cannot claim those terminal rows.

The causal regression locks the active wake row, starts cancellation, invokes `resumeQueuedRuns()` at the historical gap, and preserves unrelated queued work. Test-only commit `43b67a9b52d7430a90713609b8e528eec9486300` deterministically fails against the prior production implementation because the same-issue continuation executes; the current head passes. An independent exact-SHA review also checked this transaction/claim interleaving and found no remaining code issue. Please update the confidence score only after re-evaluating that invariant.

server/src/services/heartbeat.ts
```diff
@@ -19765,8 +19786,75 @@ export function heartbeatService(db: Db, options: HeartbeatServiceOptions = {})
...
+          updatedAt: now,
+        })
+        .where(
+          and(
+            eq(heartbeatRuns.companyId, run.companyId),
+            eq(heartbeatRuns.agentId, run.agentId),
+            eq(heartbeatRuns.status, "queued"),
+            sql`${heartbeatRuns.contextSnapshot} ->> 'issueId' = ${issueId}`,
+            sql`${heartbeatRuns.id} <> ${run.id}`,
+          ),
```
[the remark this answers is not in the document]
GabrielGoncalvesICE: Fixed in b0809e3fca28e9d0b3bbc652a779f47dab571d74. Issue-bound claims and exact operator cancellation now serialize on the issue row; cancellation atomically catches both queued and already-claimed running successors, then terminates any claimed successor process. Added a maxConcurrentRuns=2 regression for the claim-first ordering.

GabrielGoncalvesICE:
Blocking residual race on `b0809e3fca28e9d0b3bbc652a779f47dab571d74`:

`claimQueuedRun` holds the issue row lock only while changing the run from `queued` to `running`. After that transaction commits, it still performs two unconditional/lazy writes outside the lock:

- `setWakeupStatus(..., "claimed")`
- the lazy `issues.executionRunId = claimed.id` update

A valid interleaving remains:

1. sibling claim commits `run.status = running` and releases the issue lock;
2. operator cancellation acquires the issue lock and terminalizes the active run, sibling run, and sibling wake;
3. cancellation releases the issue execution lock;
4. the claimer resumes, rewrites the cancelled wake to `claimed`, and may stamp `executionRunId` to the already-cancelled sibling because the issue is now unlocked.

`executeRun` will later observe the terminal run and avoid adapter work, but the wake/issue-lock invariants are corrupted. The claim also publishes a `running` event outside the serialized section, while cancellation publishes no terminal status/plugin event for `successorRuns`, so consumers can finish on stale `running` state.

The new test pre-seeds the sibling as `running`; it does not exercise these post-claim writes or assert terminal wake + cleared issue lock + ordered terminal event evidence.

Please make the issue-bound claim atomic across the run row, wake row, and issue-execution-lock decision (preserving the explicit `source_scoped_recovery_action` exception), or otherwise guard those post-claim writes against a terminalized run. Add a causal interleaving test that fails on `4e35e81e…` and asserts: sibling run terminal, wake terminal, issue not locked to the sibling, no adapter dispatch, and terminal status observability for the sibling.

server/src/services/heartbeat.ts
```diff
@@ -19765,8 +19794,97 @@ export function heartbeatService(db: Db, options: HeartbeatServiceOptions = {})
...
+          updatedAt: now,
+        })
+        .where(
+          and(
+            eq(heartbeatRuns.companyId, run.companyId),
+            eq(heartbeatRuns.agentId, run.agentId),
+            inArray(heartbeatRuns.status, ["queued", "running"]),
+            sql`${heartbeatRuns.contextSnapshot} ->> 'issueId' = ${issueId}`,
+            sql`${heartbeatRuns.id} <> ${run.id}`,
+          ),
```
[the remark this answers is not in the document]
GabrielGoncalvesICE: Fixed at 57e4917c0dcbeb66f6ff6636c56c82972697214f. Follow-up queueing now records queuedAsFollowupToRunId, and exact-run cancellation targets only queued rows causally linked to the cancelled run. Independent queued and running same-agent/same-issue work is preserved by regression coverage.

server/src/services/heartbeat.ts
```diff
@@ -19765,8 +19799,79 @@ export function heartbeatService(db: Db, options: HeartbeatServiceOptions = {})
...
+          finishedAt: now,
+          error: "Queued issue wake suppressed by operator cancellation",
+          errorCode: "operator_cancelled_issue_run",
+          updatedAt: now,
+        })
+        .where(
+          and(
+            eq(heartbeatRuns.companyId, run.companyId),
+            eq(heartbeatRuns.agentId, run.agentId),
+            eq(heartbeatRuns.status, "queued"),
```
GabrielGoncalvesICE: Blocking on exact head `57e4917c0dcbeb66f6ff6636c56c82972697214f`: causal lineage is the right discriminator, but restricting this update to `queued` reopens the supported multi-slot claim-first race. A follow-up with `queuedAsFollowupToRunId = run.id` can acquire the issue lock first, commit `running`, and then survive operator cancellation; the revised test now asserts that survivor instead of proving zero successor. Target only causally linked rows, but include both `queued` and `running`; preserve independent rows by the lineage predicate, not by status. For a linked running row, terminalize its wake, terminate its process, and emit terminal observability. Also close the post-claim gap from the earlier public finding: run claim, wake=`claimed`, and issue execution-lock decision must be atomic or guarded so cancellation cannot be followed by wake resurrection/stale `executionRunId`. Add a real contending claim/cancel test; a pre-seeded running fixture is not causal.

GabrielGoncalvesICE:
Addressed the residual post-claim race in `5371e1c88e02d01fec10f830aad6ca75fd9f7796`. Issue-bound claim now atomically serializes the run transition, wake claim, and issue execution lock against exact-run cancellation. The contended regression holds cancellation on the issue lock while queued claim competes, then proves the causal sibling run and wake are terminal, the issue lock is not stale, the sibling never reaches the adapter, its terminal status is observable, and cross-agent/unrelated same-issue work remains intact. Focused recovery tests, the release-failure shard, and server typecheck pass locally.

server/src/__tests__/heartbeat-process-recovery.test.ts
```diff
@@ -5005,6 +5006,280 @@ describeEmbeddedPostgres("heartbeat orphaned process recovery", () => {
...
+      try {
+        await db.transaction((tx) =>
+          tx.execute(sql`select id from issues where id = ${issueId} for update nowait`)
+        );
+        return null;
+      } catch {
+        return true;
+      }
+    });
+    const competingClaim = heartbeat.resumeQueuedRuns();
```
GabrielGoncalvesICE, line 5332: Blocking coverage gap on `5371e1c88e02d01fec10f830aad6ca75fd9f7796`: this barrier proves only **cancel-first**. The `waitForValue` immediately above establishes that cancellation already holds the issue lock before `resumeQueuedRuns()` starts, and the assertion `startedAt: null` confirms the linked claim never committed. The new production branch that terminalizes a causally linked row already in `running` is therefore still untested. Add the complementary **claim-first** contention: hold the linked successor row, start queued claim so it acquires the issue lock, start cancellation so it waits, release the successor row, then prove the linked run did reach `running`/`claimed` before cancellation and finishes terminal with terminal wake, no stale issue lock, process stopped/no post-cancel adapter work, and terminal status as the final observable event. Preserve the existing cancel-first and independent-work controls.

server/src/__tests__/heartbeat-process-recovery.test.ts
```diff
@@ -5005,6 +5006,280 @@ describeEmbeddedPostgres("heartbeat orphaned process recovery", () => {
...
+      id: successorWakeupId,
+      companyId,
+      agentId,
+      source: "automation",
+      triggerDetail: "system",
+      reason: "issue_assigned",
+      payload: { issueId },
+      status: "claimed",
+      runId: successorRunId,
+    });
```
GabrielGoncalvesICE, line 6261: Blocking (new finding, exact head `5371e1c88e02d01fec10f830aad6ca75fd9f7796`): this **is** the mission's "zero successor runs" gate, and it cannot observe an escaping successor.

Both discriminating predicates are ones the fix itself writes: `eq(heartbeatRuns.agentId, agentId)` excludes peers, and `contextSnapshot ->> 'queuedAsFollowupToRunId' = runId` is stamped in exactly one enqueue path (`heartbeat.ts:19593-19595`, gated on `shouldQueueFollowupForRunningWake && sameScopeRunningRun`). A successor created by any **other** insert path is invisible here — including the peer deferred-wake promotion that inserts a queued run at `heartbeat.ts:17994`, which this same test deliberately leaves enabled two assertions below (`expect(peerDeferredWakeup?.error).not.toBe("Deferred wake suppressed by terminal run release")`).

So the query returns the expected single cancelled row whether or not the successor mechanism is actually closed — the same decorative-negative shape as the earlier `retryOfRunId` assertion.

Please assert over **all** `heartbeatRuns` rows whose `contextSnapshot ->> 'issueId'` is this issue (no agent filter, no lineage filter), against an explicit allow-list of the rows that are legitimately preserved — the independent same-agent run and any intentionally preserved cross-agent work. Then a newly opened successor path fails the test instead of slipping past the filter.

server/src/services/heartbeat.ts
```diff
@@ -19765,8 +19808,128 @@ export function heartbeatService(db: Db, options: HeartbeatServiceOptions = {})
...
+      } catch (err) {
+        // The database transaction already made the run and wake terminal and
+        // cleared any execution lock. Process cleanup is best-effort from here:
+        // a local supervisor failure must not make cancellation unretryable.
+        logger.warn(
+          { err, runId: successorRun.id },
+          "failed to terminate causally linked run after operator cancellation",
+        );
+      } finally {
+        runningProcesses.delete(successorRun.id);
```
GabrielGoncalvesICE: Blocking residual on `a17fd196517654e4959d170c197575ba71911b47`: this makes database cleanup exception-safe, but a failed SIGTERM/SIGKILL is now converted into an **untracked live successor**. The `catch` swallows the failure and this `finally` deletes the only in-process handle; `cancelRunInternal` then finalizes the agent and calls `startNextQueuedRunForAgent`, so new work may dispatch while the linked adapter is still executing. A retry cannot reach it through `cancelRun` because the row is already terminal. The new test does not cover this: its termination mock calls `releaseAdapter()` *before* throwing, guaranteeing the process has stopped. Preserve a durable/retryable termination path (or quarantine dispatch) until liveness is disproven; do not clear the process handle/runtime state or report normal terminal cleanup on kill failure. Add a causal test where termination throws **without releasing the adapter**, verify no new work dispatches and the survivor remains trackable/reapable, then perform explicit test cleanup.

server/src/__tests__/heartbeat-process-recovery.test.ts
GabrielGoncalvesICE: Addressed at `40fd35a26ac96ceb0858383f77fdebce871f8bac`. The complementary test now holds both the linked successor row and target row, starts queued claim so it owns the issue lock while blocked on the linked row, starts cancellation behind that lock, then releases the linked row and proves `running`/`claimed` before releasing the target. It asserts terminal linked run+wake, attempted process cleanup under a simulated supervisor exception, zero adapter dispatch after cancellation wins, no execution lock, and final terminal observability.

server/src/services/heartbeat.ts
```diff
@@ -19765,8 +19803,98 @@ export function heartbeatService(db: Db, options: HeartbeatServiceOptions = {})
...
+    const cancelled = cancellation?.updated ?? null;
+
+    for (const successorRun of cancellation?.successorRuns ?? []) {
+      const running = runningProcesses.get(successorRun.id);
+      if (running || successorRun.processPid || successorRun.processGroupId) {
+        await terminateHeartbeatRunProcess({
+          pid: running?.child.pid ?? successorRun.processPid,
+          processGroupId: running?.processGroupId ?? successorRun.processGroupId,
+          ...(running ? { graceMs: Math.max(1, running.graceSec) * 1000 } : {}),
+        });
```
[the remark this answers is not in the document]
GabrielGoncalvesICE: Fixed at `40fd35a26ac96ceb0858383f77fdebce871f8bac`. Run+wake terminalization and causal execution-lock cleanup commit atomically; successor process termination is caught/logged with in-memory/runtime cleanup and terminal publication in `finally`; and the late execution/dispatch paths clear any post-cancel lock without promoting recovery. The claim-first test injects termination failure and proves cancellation still resolves terminally.

server/src/__tests__/heartbeat-process-recovery.test.ts
GabrielGoncalvesICE: Fixed at `40fd35a26ac96ceb0858383f77fdebce871f8bac`. The cancel-first oracle now queries every run in the company whose context names the issue, with neither agent nor lineage predicates, and rejects every row outside the explicit causal-sibling and unrelated-work allow-list. The unrelated adapter is held until after the oracle so its legitimate completion cannot promote preserved peer work and mask the cancellation boundary.

server/src/services/heartbeat.ts
GabrielGoncalvesICE: Still blocking on current head `40fd35a26ac96ceb0858383f77fdebce871f8bac`. The revised test now manually inserts a `runningProcesses` handle, makes termination throw, then explicitly expects that handle to be deleted while asserting the adapter was never invoked. That proves the new dispatch gate, but it does not prove safety for an adapter that was already executing when kill failed; in fact it codifies the untracked-handle behavior. Add a barrier-backed live adapter, throw from termination **without** releasing it, assert no subsequent work dispatches and the survivor remains trackable/reapable, then perform explicit test cleanup and prove dispatch can resume. The existing late-lock cleanup and causal pre-dispatch gate should remain.

GabrielGoncalvesICE:
B1–B4 and the exact claim-first circuit breaker are addressed at full head `40fd35a26ac96ceb0858383f77fdebce871f8bac`. The linked claim/cancel test now enforces the published row-lock ordering, proves `running`/`claimed`, terminal run+wake, exception-safe process cleanup, zero post-cancel adapter dispatch, no stale execution lock, and final terminal observability. The cancel-first oracle is issue-wide and peer/unrelated preservation remains explicit. Focused concurrency tests (3), the release-failure shard, server typecheck, and diff hygiene pass locally. Fresh CI and independent exact-SHA review are required; merge/runtime gates remain frozen.

server/src/services/heartbeat.ts
GabrielGoncalvesICE: Addressed at `52f213bc044c6d8ddc0981bb93a1b75ebf3ccb54`. Failed termination now retains the live handle/runtime state, persists `operatorCancellationTerminationPending`, suppresses agent dispatch in both live-handle and restart/no-handle cases, and is retried by the orphan sweep. Normal terminal cleanup is published only after the retry succeeds.

GabrielGoncalvesICE: Addressed at `52f213bc044c6d8ddc0981bb93a1b75ebf3ccb54` with a barrier-backed already-running adapter. The first kill throws without releasing it; the test proves the survivor remains tracked, the durable pending marker is set, repeated scheduling leaves unrelated work queued, and no second adapter dispatch occurs. The orphan sweep then retries termination, explicit adapter cleanup completes, the marker/handle clear, and queued work resumes successfully.

GabrielGoncalvesICE:
Failed-kill quarantine is implemented at full head `52f213bc044c6d8ddc0981bb93a1b75ebf3ccb54`. A causally cancelled adapter whose termination fails remains tracked and gains a durable retry marker; all new dispatch for that agent is quarantined across live and restart paths; the orphan sweep retries termination and only clears runtime state/publishes normal cleanup/resumes dispatch after success. The live-adapter test throws without releasing the adapter, proves no queued work dispatches, then explicitly completes cleanup and proves dispatch resumes. Focused cancellation controls (4), release-failure shard, server typecheck, and diff hygiene pass locally. Fresh CI and independent exact-SHA review are required; merge/runtime gates remain frozen.

server/src/services/heartbeat.ts
```diff
@@ -19765,8 +19940,141 @@ export function heartbeatService(db: Db, options: HeartbeatServiceOptions = {})
...
+        }
+      } catch (err) {
+        terminationFailed = true;
+        successorTerminationFailed = true;
+        await db
+          .update(heartbeatRuns)
+          .set({
+            resultJson: {
+              ...parseObject(successorRun.resultJson),
+              operatorCancellationTerminationPending: true,
```
GabrielGoncalvesICE: Blocking crash window on exact head `52f213bc044c6d8ddc0981bb93a1b75ebf3ccb54`: the durable quarantine marker is written only **after** the external termination call throws, while the earlier cancellation transaction has already committed the successor as terminal. A service crash between that commit and this update restarts with a possibly live process, a terminal unmarked row that the orphan sweep excludes, and scheduling no longer quarantined. The same pre-marker window is visible to another scheduler that does not share this process-local `runningProcesses` map. Persist termination-pending atomically in the cancellation transaction for every linked running successor with live/process identity **before** attempting kill; clear it only after verified termination. Add a barrier test that queries the marker while termination is still unsettled, simulates map/service loss, proves a fresh scheduler remains blocked, then proves the reaper clears the marker and resumes dispatch after successful cleanup.

GabrielGoncalvesICE: Fixed at exact head `d24de10063ea7595f9be99fab1226225941b8bce`. Linked running successors now persist both process identity and `operatorCancellationTerminationPending` inside the issue-serialized cancellation transaction, before external termination starts; only verified termination or the reaper clears the marker. The barrier-backed regression observes the committed marker/PID while termination is unresolved, deletes the in-memory process map, verifies a fresh service remains quarantined, then proves reaper cleanup clears the marker and dispatch resumes. Focused causal/independence controls, the release-failure shard, server typecheck, and diff check pass.

GabrielGoncalvesICE:
SMA-235 exact-head update: `d24de10063ea7595f9be99fab1226225941b8bce` closes the post-commit/pre-marker crash gap. Cleanup intent plus PID are transactionally durable before termination, restart/map-loss remains quarantined, and verified reaping resumes dispatch. Local evidence: four focused cancellation/independence controls PASS; task-drain release-failure shard PASS; server typecheck PASS; `git diff --check` PASS. Fresh CI and independent exact-SHA review are required; merge/runtime gates remain frozen.

server/src/services/heartbeat.ts
```diff
@@ -19802,13 +20143,9 @@ export function heartbeatService(db: Db, options: HeartbeatServiceOptions = {})
       runningProcesses.delete(run.id);
```
GabrielGoncalvesICE: Blocking symmetric failure on exact head `d24de10063ea7595f9be99fab1226225941b8bce`: in the `suppressDeferredPromotion` path, `cancelRunAndQueuedRunsForIssue` terminalizes the **target run itself** before this kill. If target termination throws, this `finally` deletes its only live handle and the exception exits before wake cancellation, issue release, lifecycle event, agent finalization, or scheduling quarantine. The target row is already terminal, so a retry hits the early non-cancellable return; the issue may remain locked to that terminal run while its adapter continues untracked. This is introduced by the new exact-cancel ordering—the ordinary path terminalizes only after target kill. Apply the same durable termination-pending/quarantine/reaper protocol to the target (or preserve a retryable pre-terminal target state), and add a live-target/no-release test proving marker/identity commit before kill settles, wake/lock/dispatch safety after failure and map loss, then reaper cleanup and normal resume.

GabrielGoncalvesICE: Fixed at exact head `dd2e9def37e027e6c05907a4d3fa425a3d39b68f`. The target run now persists its process identity and cleanup marker in the same serialized transaction that terminalizes the target run, wake, and issue execution lock. A target kill failure retains cleanup state and suppresses finalization/new dispatch; marker-based quarantine and orphan reaping now cover target and linked runs without relying on the successor-only error code. The new live-adapter barrier test observes terminal run/wake/lock plus marker before termination settles, simulates map loss with a fresh service, proves no dispatch, then proves reaper cleanup and dispatch resume. Five focused cancellation/independence controls, release-failure shard, server typecheck, and diff check pass.

GabrielGoncalvesICE:
SMA-235 exact-head update: `dd2e9def37e027e6c05907a4d3fa425a3d39b68f` closes the target-run failed-kill/retry gap. Target run, wake, issue lock, process identity, and durable cleanup marker now commit before kill; map-loss restart remains quarantined until reaping succeeds. Local evidence: five focused causal/independence controls PASS; task-drain release-failure shard PASS; server typecheck PASS; `git diff --check` PASS. Fresh CI and independent exact-SHA review are required; merge/runtime gates remain frozen.

server/src/services/heartbeat.ts
```diff
@@ -19765,8 +19938,176 @@ export function heartbeatService(db: Db, options: HeartbeatServiceOptions = {})
...
+            executionRunId: null,
+            executionAgentNameKey: null,
+            executionLockedAt: null,
+            updatedAt: now,
+          })
+          .where(
+            and(
+              eq(issues.id, issueId),
+              eq(issues.companyId, run.companyId),
+              inArray(issues.executionRunId, cancelledRunIds),
```
GabrielGoncalvesICE, line 20909: Blocking cross-agent gap on exact head `dd2e9def37e027e6c05907a4d3fa425a3d39b68f`: this transaction clears `issues.executionRunId` for the target before termination is confirmed, while the new pending-cleanup admission check is only in `startNextQueuedRunForAgent(agentId)`. If target kill fails and the issue is reassigned (the normal author→reviewer handoff), a different agent has no matching pending run and can claim the same issue while the old target adapter is still alive. The regression only queues unrelated work for the **same** agent, so it cannot observe this. Quarantine the issue as well as the agent until the reaper proves termination—either retain/interpret an execution lock for pending cleanup or make issue claim/admission check the durable marker under the issue-row lock. Add a peer-agent same-issue wake test: failed live target kill + map loss must keep it deferred/uninvoked; successful reap must release exactly once and allow normal handoff.

GabrielGoncalvesICE: Fixed at exact head `2eec427b5b025da9b44e7ff65904a5d2b03af0d1`. A target with durable termination pending now retains the issue execution lock; enqueue treats that terminal-but-pending holder as active even after reassignment, so peer work is deferred. Verified reaping clears the marker, releases the lock under the existing issue-row transaction, and promotes the deferred peer exactly once. The barrier-backed test starts a live target, forces kill failure, reassigns to a peer, proves the peer wake remains deferred/uninvoked after map loss, then proves successful reap permits one peer adapter dispatch. Five focused controls, the release-failure shard, server typecheck, and diff check pass.

GabrielGoncalvesICE:
SMA-235 exact-head update: `2eec427b5b025da9b44e7ff65904a5d2b03af0d1` closes the cross-agent issue-admission gap. A failed target kill retains the issue execution lock as durable quarantine; a reassigned peer wake stays deferred through map loss, and verified reaping releases/promotes exactly once. Local evidence: five focused causal/independence controls PASS; task-drain release-failure shard PASS; server typecheck PASS; `git diff --check` PASS. Fresh CI and independent exact-SHA review are required; merge/runtime gates remain frozen.

server/src/services/heartbeat.ts
```diff
@@ -19765,8 +19940,141 @@ export function heartbeatService(db: Db, options: HeartbeatServiceOptions = {})
...
+    const now = new Date();
+    const cancellation = await db.transaction(async (tx) => {
+      await tx.execute(
+        sql`select id from issues where company_id = ${run.companyId} and id = ${issueId} for update`,
+      );
+      const updated = await tx
+        .update(heartbeatRuns)
+        .set({ status: "cancelled", ...patch, updatedAt: now })
+        .where(eq(heartbeatRuns.id, run.id))
+        .returning()
```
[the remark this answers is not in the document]
GabrielGoncalvesICE: Confirmed applicable at exact head `2eec427b5b025da9b44e7ff65904a5d2b03af0d1`. Both the target update (around line 19982) and successor update (around line 20022) predicate only on run id. A completion that commits after candidate selection can therefore have its terminal outcome overwritten by `cancelled`. This remains a release-blocking P1. Please add a status CAS to the actual updates and a causal race test proving a concurrently completed run keeps its terminal outcome and is not terminated or emitted as cancelled.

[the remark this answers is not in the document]
GabrielGoncalvesICE: Fixed at exact head `c3811c917a5284f13f30ea9e2923831ca34298ac`. Target and linked cancellation updates now use status compare-and-swap predicates. If the target finishes before cancellation wins, the operator path returns the authoritative terminal row without process termination or wake mutation; if a linked run finishes after selection, the successor CAS rejects the cancellation and preserves its terminal row and wake. Two row-lock barrier regressions cover both interleavings. Seven focused cancellation/concurrency controls, the release-failure shard, server typecheck, and diff check pass.

GabrielGoncalvesICE:
SMA-235 exact-head update: `c3811c917a5284f13f30ea9e2923831ca34298ac` closes the concurrent-terminal overwrite race. Target and linked updates are status-CAS guarded; lost target cancellation returns the authoritative terminal result without kill/wake mutation. Local evidence: seven focused causal/concurrency/independence controls PASS; task-drain release-failure shard PASS; server typecheck PASS; `git diff --check` PASS. Fresh CI and independent exact-SHA review are required; merge/runtime gates remain frozen.

server/src/services/heartbeat.ts
```diff
@@ -13992,6 +14005,57 @@ export function heartbeatService(db: Db, options: HeartbeatServiceOptions = {})
...
+      clearHeartbeatRunRuntimeStatus(pendingRun.id);
+      publishLiveEvent({
+        companyId: cleanedRun.companyId,
+        type: "heartbeat.run.status",
+        payload: buildHeartbeatRunStatusLiveEventPayload(cleanedRun),
+      });
+      publishRunLifecyclePluginEvent(cleanedRun);
+      await releaseIssueExecutionAndPromote(cleanedRun, {
+        suppressImmediateRecovery: true,
+      });
```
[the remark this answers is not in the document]
GabrielGoncalvesICE: Confirmed applicable at exact head `c3811c917a5284f13f30ea9e2923831ca34298ac`. The normal exact-cancellation release path passes `suppressDeferredPromotion`, but the pending-cleanup reaper calls `releaseIssueExecutionAndPromote` with only `suppressImmediateRecovery`; a same-agent deferred issue wake can therefore be promoted and started after cleanup. This violates the zero-successor operator-cancel invariant. Apply the same deferred-promotion suppression during reaping and add a failed-target-kill regression with a same-agent deferred wake proving it is terminalized, no successor run/adapter starts, and no stale promotion remains. Any later peer work must come from an explicit post-cleanup wake, not automatic promotion of a pre-cleanup deferred wake.

GabrielGoncalvesICE:
SMA-235 evidence-tightening head: `7113c5ef559d64b06f458ee088446d8cbbc88990`. The target/linked row-lock races now assert preserved terminal result/error payloads, unchanged claimed wakes, absence of cancellation status events, and no target termination when CAS loses. Seven focused controls PASS; task-drain release-failure shard PASS; server typecheck PASS; `git diff --check` PASS. Per the circuit breaker, independent review is withheld until this exact SHA has terminal required CI, terminal Greptile, and no unresolved public blocking threads.

server/src/services/heartbeat.ts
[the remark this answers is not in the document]
GabrielGoncalvesICE: Clarification after reconciling the cross-agent quarantine case: the hard invariant is zero automatic continuation for the cancelled agent. A peer wake may be promoted exactly once after verified cleanup only when it was created by an explicit reassignment while the old target was quarantined; that is a distinct operator action, not cancellation recovery. The regression should seed both wakes, prove the same-agent deferred wake becomes terminal and never enters an adapter, and prove at most one explicitly reassigned peer begins only after reap. Persist the suppression intent across map loss so the reaper cannot confuse the two.

server/src/services/heartbeat.ts
```diff
@@ -13992,6 +14005,62 @@ export function heartbeatService(db: Db, options: HeartbeatServiceOptions = {})
...
+        );
+        continue;
+      }
+
+      const resultJson = parseObject(pendingRun.resultJson);
+      const suppressDeferredPromotion =
+        resultJson.operatorCancellationSuppressDeferredPromotion === true;
+      delete resultJson.operatorCancellationTerminationPending;
+      delete resultJson.operatorCancellationSuppressDeferredPromotion;
+      const cleanedRun = await db
```
GabrielGoncalvesICE, line 14432: P1 durability gap at exact head `40ab44e8af26cea4130e9f6fb3714964b9889866`: after termination succeeds, this standalone update removes `operatorCancellationTerminationPending` and commits before `releaseIssueExecutionAndPromote` releases the issue lock. A process crash in that interval leaves a terminal run that the pending-cleanup sweep no longer selects while `issues.executionRunId` still points to it, permanently stranding the issue. Keep cleanup discoverable until issue release commits—e.g. persist a post-termination/release-pending phase and make the release transition restart-idempotent—without re-killing a stale/reused PID. Add a causal barrier after termination is durably acknowledged but before issue release, instantiate a fresh service, and prove it clears the lock, suppresses the cancelled agent wake, and promotes at most the explicit peer exactly once.

GabrielGoncalvesICE: Fixed at exact head `6ef6a810d16bbcb3ad03b602f1938dd18cd55eff`. Successful termination now persists an acknowledged-but-still-pending phase before issue release. A restart skips re-signalling the recorded PID, completes idempotent issue release, then clears cleanup markers. The barrier regression crashes after acknowledgement, proves the run remains discoverable and the issue lock/wakes remain quarantined, then proves a fresh service releases once and promotes only the explicitly marked peer handoff.

server/src/services/heartbeat.ts
```diff
@@ -19765,8 +19952,191 @@ export function heartbeatService(db: Db, options: HeartbeatServiceOptions = {})
...
+          })
+          .where(inArray(agentWakeupRequests.id, wakeupRequestIds));
+      }
+
+      // A live target with pending termination keeps owning the issue lock until
+      // the reaper verifies cleanup. Clearing it here would let a reassigned peer
+      // enter the same issue while the old adapter may still be executing.
+      const cancelledRunIds = [
+        ...(!targetTerminationPending ? [updated.id] : []),
+        ...successorRuns.map((successorRun) => successorRun.id),
```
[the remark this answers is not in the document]
GabrielGoncalvesICE: Fixed at exact head `6ef6a810d16bbcb3ad03b602f1938dd18cd55eff`. Termination-pending linked successors are excluded from lock-release bookkeeping, and issue admission also discovers any durable pending cancellation marker under the issue-row transaction even if the execution pointer is absent. Thus a peer remains deferred until acknowledged cleanup completes.

server/src/__tests__/heartbeat-process-recovery.test.ts
GabrielGoncalvesICE: Reconciled at exact head `6ef6a810d16bbcb3ad03b602f1938dd18cd55eff`. The claim-first contention remains covered: linked running status/wake terminalization, process cleanup/quarantine, no post-cancel dispatch, terminal observability, and lock release are asserted alongside cancel-first and independent-work controls. The restartable acknowledged-cleanup phase now additionally covers the post-termination/pre-release crash window.

GabrielGoncalvesICE:
SMA-235 exact-head correction: `6ef6a810d0eda63e17bfaea168f90f9deddc6fe5` is the authoritative current head (the abbreviated SHA in the three immediately preceding thread replies expanded incorrectly). This head binds peer promotion to an exact durable cleanup handoff, preserves same-agent suppression across restart, retains discoverability through an acknowledged cleanup phase until issue release commits, and quarantines issue admission on any pending linked cleanup. Local verification: 7 focused concurrency/cancellation tests PASS; 2 release-failure tests PASS; server typecheck PASS; diff check PASS. CI and Greptile are running; independent review remains withheld until the terminal gate is green.

server/src/services/heartbeat.ts
GabrielGoncalvesICE: Reconciled on authoritative exact head `6ef6a810d0eda63e17bfaea168f90f9deddc6fe5`: linked queued/running rows are selected by causal lineage and terminalized with status CAS; claim/cancel is issue-row serialized; running process cleanup remains durably quarantined through restart and release. Claim-first, cancel-first, terminal-result races, and independent-work controls all pass.

server/src/services/heartbeat.ts
```diff
@@ -19765,8 +19971,191 @@ export function heartbeatService(db: Db, options: HeartbeatServiceOptions = {})
...
```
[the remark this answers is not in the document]
GabrielGoncalvesICE: Reconciled on authoritative exact head `6ef6a810d0eda63e17bfaea168f90f9deddc6fe5`. Termination-pending linked successors are excluded from `cancelledRunIds`, so cancellation no longer clears their execution lock. In addition, enqueue admission searches durable cancelled+pending cleanup rows under the issue transaction even when the execution pointer is absent, keeping reassigned peer work deferred through failed termination/map loss. The failed-target barrier now also covers termination acknowledgement, crash before issue release, fresh-service quarantine, idempotent release, cancelled-agent suppression, and exactly one marked peer handoff. Greptile has marked this thread resolved/outdated and the exact-head Greptile check is green.

GabrielGoncalvesICE:
Fast-path packaging verification found that the managed git installer correctly rejected the prior head because `patchedDependencies` and `pnpm-lock.yaml` were inconsistent under `pnpm install --frozen-lockfile`. The new commit only regenerates the lockfile from the existing workspace/package/patch configuration. A fresh frozen lockfile check succeeds; no cancellation logic changed. Runtime activation remains gated on checks for the new head SHA.

GabrielGoncalvesICE:
SMA-235 B1 forward fix is published at exact head `c67abc36b00330049f10d3ee0311ef2ded0c9aab`. Both successful inline termination paths now clear `operatorCancellationTerminationPending`, `operatorCancellationTerminationAcknowledged`, and `operatorCancellationSuppressDeferredPromotion` only after issue release commits. The causal claim-first success test covers target and linked successor cleanup, asserts all three markers are absent, preserves zero linked adapter dispatch, and proves an explicit redispatch succeeds immediately. Exact-head local verification: 7 focused cancellation/concurrency controls PASS; 2 ordinary release-failure controls PASS; server typecheck PASS; diff check PASS. No merge, deployment, database/runtime mutation, G1, or SMA-233 rerun performed.

server/src/services/heartbeat.ts
```diff
@@ -19823,13 +20311,30 @@ export function heartbeatService(db: Db, options: HeartbeatServiceOptions = {})
...
         ...(options.eventPayload ? { payload: options.eventPayload } : {}),
       });
-      await releaseIssueExecutionAndPromote(cancelled);
+      if (!targetTerminationFailed) {
+        await releaseIssueExecutionAndPromote(cancelled, {
+          suppressImmediateRecovery: options.suppressImmediateRecovery,
+          suppressDeferredPromotion: options.suppressDeferredPromotion,
+        });
+        if (targetHasPendingCleanup) {
+          cancelled = await clearOperatorCancellationCleanupMarkers(cancelled);
```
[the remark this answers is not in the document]
GabrielGoncalvesICE: This quarantine is intentional on the release-failure path and should not be cleared here. The operator requirement is to clear the three markers only after `releaseIssueExecutionAndPromote` commits. If that call throws, the issue release/lock cleanup did not complete; retaining the marker blocks unsafe redispatch and leaves the run discoverable for the orphan reaper. Clearing it in this branch would recreate the concurrent-dispatch defect. The successful inline target and successor paths now clear all three markers after their respective release calls return. The release-failure shard passes 2/2 locally at c67abc36b00330049f10d3ee0311ef2ded0c9aab.

server/src/services/heartbeat.ts
```diff
@@ -20447,26 +20910,63 @@ export function heartbeatService(db: Db, options: HeartbeatServiceOptions = {})
...
-      if (running) {
+      if (running || (cancelled && targetHasPendingCleanup)) {
         await terminateHeartbeatRunProcess({
-          pid: running.child.pid,
-          processGroupId: running.processGroupId,
-          graceMs: Math.max(1, running.graceSec) * 1000,
+          pid: running?.child.pid ?? cancelled?.processPid,
+          processGroupId: running?.processGroupId ?? cancelled?.processGroupId,
+          ...(running ? { graceMs: Math.max(1, running.graceSec) * 1000 } : {}),
         });
```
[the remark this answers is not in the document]
GabrielGoncalvesICE: Fixed at exact head `d8e5d006`. Persisted-only cancellation now proves process ownership with the recorded `processStartedAt` before signaling. A live PID whose observed OS start time is newer is treated as recycled: no signal is sent, the termination marker remains pending, and the issue lock stays quarantined. The same guarded helper is used by target, causal-successor, and orphan-reaper paths. Added a regression that uses a live reused PID and asserts zero termination calls plus durable lock/quarantine preservation.

server/src/services/heartbeat.ts
```diff
@@ -20456,17 +20942,31 @@ export function heartbeatService(db: Db, options: HeartbeatServiceOptions = {})
...
-    const finishedAt = new Date();
-    const cancelled = await setRunStatus(run.id, "cancelled", {
-      finishedAt,
-      error: reason,
-      errorCode,
-      ...(resultJson ? { resultJson } : {}),
-    });
+    if (cancelled && targetHasPendingCleanup && !targetTerminationFailed) {
+      const cleanedResultJson = parseObject(cancelled.resultJson);
+      cleanedResultJson.operatorCancellationTerminationAcknowledged = true;
```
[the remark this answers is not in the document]
GabrielGoncalvesICE: Fixed at exact head `d8e5d006`. The original map-loss gap is closed without blindly trusting the numeric PID: persisted-only target, successor, and reaper cleanup now validates OS process start time against the stored spawn time before calling the termination helper. Missing/mismatched identity fails closed and retains quarantine; successful termination is still verified by `terminateLocalService`. Server typecheck and `git diff --check` pass; fresh CI is running.

server/src/services/heartbeat.ts
```diff
@@ -20428,8 +20684,243 @@ export function heartbeatService(db: Db, options: HeartbeatServiceOptions = {})
...
+    }
+
+    const pidAlive = isProcessAlive(run.processPid);
+    const processGroupAlive = isProcessGroupAlive(run.processGroupId);
+    if (!pidAlive && !processGroupAlive) return;
+
+    const expectedStartedAt = run.processStartedAt?.getTime() ?? Number.NaN;
+    if (!pidAlive || !run.processPid || !Number.isFinite(expectedStartedAt)) {
+      throw new Error(`Cannot establish ownership of persisted process for cancelled run ${run.id}`);
+    }
```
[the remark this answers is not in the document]
GabrielGoncalvesICE: Fixed at exact head `3c1ec7c0`. Persisted-only cleanup now distinguishes: (1) parent and group gone → acknowledge without signaling, (2) live PID → verify OS start time before signaling, and (3) dead detached leader with a still-live matching PID/PGID → terminate the surviving descendants before acknowledgment. Ambiguous non-matching groups still fail closed. The serialized test failures on the prior head were test-harness assumptions that a kill callback fires for an already-dead fake PID; those fixtures now release the simulated adapter directly, matching the new liveness semantics. Server typecheck and diff check pass; fresh CI is running.

server/src/services/heartbeat.ts
```diff
@@ -20428,8 +20684,255 @@ export function heartbeatService(db: Db, options: HeartbeatServiceOptions = {})
...
+    // its descendants remain alive. If the leader has exited, the missing PID
+    // cannot be identity-checked, but a matching persisted PID/PGID still names
+    // that extant group; terminate the descendants before releasing quarantine.
+    if (!pidAlive && processGroupAlive && run.processPid === run.processGroupId) {
+      await terminateHeartbeatRunProcess({
+        pid: run.processPid,
+        processGroupId: run.processGroupId,
+      });
+      return;
+    }
```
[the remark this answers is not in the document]
GabrielGoncalvesICE: Fixed at exact head `399224c3`. Descendant-only cleanup no longer treats a matching numeric PGID as ownership: on Linux it scans live group members and requires the exact `PAPERCLIP_RUN_ID=<run.id>` identity injected into local adapter environments before signaling. If identity is absent, unreadable, or unsupported, cleanup fails closed and retains quarantine. Added paired Linux regressions proving owned descendants are reaped and an unowned recycled group is neither signaled nor released. Server typecheck and `git diff --check` pass; fresh CI is running.

PR #12241, opened 2026-08-26
fix(runner): resume native Codex semantic calls

cryppadotta:
## Thinking Path

> - Paperclip manages agent work through a durable control plane.
> - Runner D connects provider sessions to that control plane.
> - A Runner D crash must not destroy a pending Codex tool call.
> - The old process layout made the Codex server a child of Runner D.
> - A crash therefore lost the provider thread and its pending call.
> - This pull request keeps one Codex server alive across Runner D restarts.
> - The benefit is one real provider resume and one control-plane effect.

## Linked Issues or Issue Description

**What happened?**

Runner D could not resume a real pending Codex semantic tool call after a process crash. The crash stopped both Runner D and its Codex app server. The provider thread and pending JSON-RPC call were lost.

**Expected behavior**

The replacement Runner D process must attach to the same Codex thread. It must complete the original semantic call with the same call ID. It must apply the control-plane mutation once.

**Steps to reproduce**

1. Start a native Codex run that calls `report_progress`.
2. Send `SIGKILL` to Runner D after the semantic input is durable.
3. Observe that the provider process stops and the pending call cannot resume.

**Paperclip version or commit**

The bug reproduces on the current `master` base for this pull request.

**Deployment mode**

Self-hosted server with native Runner D.

**Agent adapter(s) involved**

Codex.

## What Changed

- Start one real Codex app server outside the Runner D process group.
- Connect each Runner D generation through a private WebSocket Unix proxy.
- Persist the semantic tool inventory and pending native call identity.
- Resume the same Codex thread and reconcile the reissued call once.
- Add a database receipt and a stable effect ID for idempotent mutations.
- Persist the detached provider PID and process group on the heartbeat run.
- Persist a guardian before its handshake is allowed to launch Codex.
- Authenticate surviving provider-group members with a per-launch owner token.
- Retain inconclusive ownership records for later terminal-run cleanup retries.
- Terminate orphaned providers during cancellation, shutdown, and crash recovery.
- Add restart, reconciliation, and production vertical-slice tests.
- Document the provider process and durable transport contracts.
- Treat an omitted or zero provider-reported request count with positive tokens as a lower bound of one.
- Preserve the provider turn as a correlation handle and label absent per-request IDs and per-turn cost as unavailable instead of provider-reported zero.
- Include positive token totals in the live proof so downstream validation can reject false zero accounting.

## Verification

- `cargo test -p paperclip-runner-core provider_events::tests --lib` from `packages/paperclip-runner/runner`
[...]

PR #12596, opened 2026-08-31
test(runner): bound codex provider exit polls by wall clock

zannis:
## Thinking Path

> - The runner subsystem drives a Codex provider process and reads its events with `CodexProvider::poll`
> - The Codex provider integration tests wait for those events in poll loops
> - Two of those loops count iterations instead of measuring time, so they stop waiting too early
> - This makes `cargo test` fail at random on branches that change no Rust code
> - This pull request bounds the two loops by wall clock, like every other wait in the same file
> - The benefit is that a red CI job now means a real defect

**What happened?**

`packages/paperclip-runner/runner/crates/runner-core/tests/codex_provider.rs` fails
`cargo test` at random. The failure appears in the `ci / Build` job with exit code 101.
It appears on branches that change no Rust code. Two tests fail:

- `ambiguous_or_dead_replacement_start_preserves_result_not_exit_authority` at line 1274
- `ambiguous_replacement_turn_adopts_one_later_completion_identity` at line 1443

Both assertions report `left: None`. The value is not wrong. The loop never saw the
`CodexProviderEvent::Exited` event at all.

**Expected behavior**

The tests must wait for the provider process to exit. A test must fail only when the
provider gives a wrong result.

**Steps to reproduce**

1. Build the integration test: `cargo test --test codex_provider --no-run`.
2. Run one of the two named tests 25 times in a row.
3. About 8 of the 25 runs fail with `left: None`.

Reproduced on `master` at `2e5a24e17`.

**Related pull requests**

Refs #12241. That pull request also edits
`packages/paperclip-runner/runner/crates/runner-core/tests/codex_provider.rs`. It does not
fix these two loops. The two changes may need a merge if both land.

**Root cause**

`CodexProvider::poll` (`crates/runner-core/src/codex_provider.rs:824`) reads with a 1 ms
timeout. That timeout does not apply on every path.
`ProcessSupervisor::receive_stdout_line` (`crates/runner-core/src/process_supervisor.rs:293`)
returns at once, and uses none of the 1 ms budget, in two cases: `StdoutClosed` at line 309
and `RecvTimeoutError::Disconnected` at line 315.

A child process closes its pipes before its exit status is ready to reap. In that window
every `poll()` call returns `Ok(None)` in nanoseconds. A loop of 64 or 128 iterations then
ends in microseconds, before the exit status is available. The failing run above ends in
0.06 s.

## What Changed

- `tests/codex_provider.rs`: bound the exit wait at line 1256 by a 5 second deadline instead
  of 64 iterations.
- `tests/codex_provider.rs`: bound the exit wait at line 1397 by a 5 second deadline instead
  of 128 iterations.
- Both loops now sleep 1 ms when `poll()` returns no event. This copies the pattern that the
  same file already uses at line 1511 and in every `wait_for_*` helper.
- No production code changes. The change is test-only.

## Verification

Measured before and after the change. Each test ran 25 times in sequence, on an idle
machine, with `--test-threads=1`.

| test | before | after |
|---|---|---|
| `ambiguous_or_dead_replacement_start_preserves_result_not_exit_authority` | 8 / 25 failed | 0 / 25 failed |
| `ambiguous_replacement_turn_adopts_one_later_completion_identity` | 9 / 25 failed | 0 / 25 failed |

The full `codex_provider` suite also ran 12 times with `--test-threads=4` after the change.
Every run passed.

Commands:

```
cargo test --test codex_provider --no-run
cargo test --test codex_provider
```

## Risks

Low risk. The change touches test code only. It makes two waits longer in the failure case:
a genuinely broken provider now takes up to 5 seconds to fail these two tests instead of
microseconds. Every other wait in this file already uses the same 5 second deadline.

## Model Used

Depends-on: none — this is a self-contained test-only change with no prerequisite pull request.

## Checklist

- [x] I have updated relevant documentation to reflect my changes — none apply. This change is test-only and alters no public interface, so no docs page and no end-to-end test change is needed.


zannis:
Impact data, since it may help prioritise this: the two assertions this PR fixes have turned
`ci / Build` red **five more times on four unrelated branches** in the ~8.5 h since this PR opened.
All of them panic at exactly the two lines changed here (`codex_provider.rs:1274` and `:1443`), and
none of the four branches touches Rust:

| when (UTC) | branch | line | run |
|---|---|---|---|
| 08-31 16:59 | `rename-workspaces-as-worktrees` | 1274 | [33417094607](https://github.com/paperclipai/paperclip/actions/runs/33417094607) |
| 08-31 22:24 | `fix/secret-proposal-interaction-recovery` | 1443 | [33446034775](https://github.com/paperclipai/paperclip/actions/runs/33446034775) |
| 08-31 23:41 | `feat/grok-sandbox-credential-staging` | 1274 + 1443 | [33451732168](https://github.com/paperclipai/paperclip/actions/runs/33451732168) |
| 09-01 00:42 | `fix/secret-proposal-interaction-recovery` | 1274 | [33455869118](https://github.com/paperclipai/paperclip/actions/runs/33455869118) |
| 09-01 00:53 | `fix/secret-proposal-interaction-recovery` | 1274 | [33456554147](https://github.com/paperclipai/paperclip/actions/runs/33456554147) |

Method: every `PR`-workflow failure since this PR opened, filtered to runs whose `ci / Build` job
failed, then grepped for a `tests/codex_provider.rs:<line>` panic. The other `ci / Build` failures in
that window are all on `runner/sdk-and-developer-tools`, which does change Rust — a different cause,
not this one.

Scope check for reviewers: the diff is one test file, no production code. I classified all 64 bounded
poll loops in it by measurement rather than by reading, and only these two are racy — the awaited
event requires the child process to have exited, so a non-blocking `poll()` can burn the whole
iteration budget in microseconds. The other sleepless loops either wait on a blocking executor
(satisfied on iteration 1 of 64, 15/15 runs) or on an already-buffered event (iteration 4 of 32,
11/11 runs), so widening the change would add churn without removing risk.

Branch is `MERGEABLE` / `CLEAN`; the six commits it is behind touch no Rust and no runner file.

zannis:
Local reproduction and verification — the flake is now confirmed against this fix, not just argued from the code.

I had noted on the original report that the Rust crate could not be built in my environment. That was wrong: `cargo build --test codex_provider -p paperclip-runner-core` succeeds here. So I built both variants of the test binary from the same tree and ran them head to head.

**Method.** Two binaries from identical sources except this diff: `prefix` (the file as it is on `master`) and `fixed` (this PR). Each ran the two affected tests only, 144 times, in batches of 8 concurrent processes to create the CPU contention that CI has and an idle container does not.

| binary | failures | runs | rate |
|---|---|---|---|
| `master` (bounded 64/128 iteration loops) | **22** | 144 | 15.3% |
| this PR (5 s wall-clock deadline, 1 ms sleep) | **0** | 144 | 0% |

The failures land on exactly the line CI reports:

```
thread 'ambiguous_replacement_turn_adopts_one_later_completion_identity' panicked at
crates/runner-core/tests/codex_provider.rs:1474:9
test result: FAILED. 1 passed; 1 failed; ... finished in 0.05s
```

Note the 0.05 s — the pre-fix loop exhausts its 128 iterations and gives up in 50 ms, well before the fake Codex process writes its error and exits. That is the mechanism, observed rather than inferred.

At a 15.3% base rate, 0 failures in 144 fixed runs would happen by chance with probability ~4e-11.

**Sixth production recurrence since this PR opened**, on a branch that touches no Rust:

| when (UTC) | branch | line | run |
|---|---|---|---|
| 09-01 04:29 | `fix/composio-application-activation` | 1474 | [33469890980](https://github.com/paperclipai/paperclip/actions/runs/33469890980) |

**Branch is current.** Merged `upstream/master` (picking up #12616 and #12617); the merge was clean and the diff against master is unchanged at one file, 68/54. Full CI is re-running on the merge commit.

zannis:
## This change does not actually fix the flake — measured, not assumed

I was able to build and run the Rust crate locally after all (the "no C
linker" note I wrote earlier is stale — `cc`/`gcc` are present; rustup just
has no *default* toolchain, so `cargo +1.91-aarch64-unknown-linux-gnu ...`
works). That let me A/B the flake instead of reasoning about it, and the
result contradicts the premise of this PR.

### Experiment

Two test binaries — `master` (`dda4dff64`, the `(0..64)`/`(0..128)` loops) and
this branch (`9adb2c2c7`, deadline-bounded loops). 8-core box, 24 busy-loop
CPU hogs, 30 sequential runs each of just the two tests, identical load:

```
master (before): 3/30 runs FAILED
this branch:     2/30 runs FAILED
```

3/30 → 2/30 is noise.

### What the deadline did and didn't do

It did remove the failure shape this PR targets. Every `master` failure is the
`left: None` assertion, i.e. the loop exhausting before `Exited`:

```
codex_provider.rs:1307  malformed-error must retain the completed result ...
  left: None / right: Some((false, true, false))
codex_provider.rs:1476  the replacement completion, not the old result, ...
  left: None / right: Some((false, true, true))
```

No run on this branch failed that way. But the same test now fails a
*different* assertion:

```
codex_provider.rs:1482  the replacement identity should be established before
                        replaying its output   <- replacement_started == false
codex_provider.rs:1486  observe replacement completion                
                        <- replacement_completed == false
```

The loop at :1438 returns the moment it sees `Exited`, while the assertions
after it require `turn/started` and `turn/completed` to have been observed on
the way. The exit tuple arrives; the two notifications never do.

### Why — and why it isn't a test bug

`ProcessSupervisor::receive_stdout_line` (`process_supervisor.rs:293`)
collapses four outcomes into a single `Ok(None)`: real EOF (`StdoutClosed`,
:309), **a 1ms read timeout** (:310), `Disconnected` (:315), and an
already-expired deadline (:301).

`CodexProvider::poll` (`codex_provider.rs:1058`) then does:

```rust
let Some(line) = self.process.receive_stdout_line(Duration::from_millis(1))? else {
    let exit = self.process.try_wait()?;
    return if let Some(exit) = exit { ... Exited ... }
```

It cannot distinguish "nothing within 1ms" from "nothing ever". Under load the
1ms timeout expires while `turn/started`/`turn/completed` are still queued
from the reader thread, `try_wait()` sees the child gone, and `poll()` emits
`Exited` with those lines undelivered. Callers stop polling at `Exited`, so
they are lost.

That is a production-visible ordering issue, not a test artifact: a consumer
can reach a terminal `Exited` with `completed_turn_authoritative: false` for a
turn that did complete, because the `turn/completed` proving it was dropped.
Ordering itself is fine — the mpsc channel delivers `StdoutClosed` in order —
the bug is purely the timeout/EOF conflation.

### Suggested real fix

Once `try_wait()` shows the child exited, drain stdout to a *genuine* EOF
before emitting `Exited`. That requires `receive_stdout_line` to report
"timed out" separately from "closed", which today it does not. Draining
terminates safely because a dead child's pipe reaches real EOF.

The wall-clock deadline in this PR is still worth keeping — it's a strict
improvement over a CPU-speed-dependent bound — but on its own it does not
close the flake.

### Confidence

The A/B result is measured. The `receive_stdout_line` mechanism is a code read
plus failure signatures that match its prediction exactly; I have not yet
written a test that exercises the timeout-vs-EOF shape directly, and I don't
think this should be called fixed until someone does.

I'd rather not silently widen this PR from "test-only" into a change to
provider exit semantics, so I'm leaving it as-is and flagging the choice:
either land this as partial hardening and retitle it so it doesn't claim to
fix the flake, or hold it and expand scope to include the `poll()` fix.

nickyleach:
Thanks @zannis! Was just looking at this same failure, glad you beat me to it.

PR #12606, opened 2026-08-31
fix(server): verify embedded postgres is reachable before reusing its pidfile

Quentin-M:
## Thinking Path

> - Paperclip embeds its own Postgres and starts it from `server/src/index.ts` on boot
> - On restart, it first checks `postmaster.pid` for a `runningPid`; if the OS reports that PID alive, it logged a warning and reused it as-is, with no verification the server was actually accepting connections
> - Right after a container recreation (new image, `docker compose up` with recreate), the pidfile can name a real, freshly-spawned postgres process — this app's own previous embedded start of it — that is still replaying WAL and not yet listening
> - The very first connection probe then gets `ECONNREFUSED`, and the process used to exit hard instead of waiting, surviving only via Docker's restart-policy crash-loop retrying a few seconds later
> - This pull request makes the `runningPid` branch prove the process is actually reachable (same live-connection probe the no-pidfile branch already uses) with a short retry-with-backoff, before falling back to treating the pidfile as stale
> - The benefit is a clean, deterministic startup on container recreation instead of depending on the restart-policy crash-loop as an implicit recovery mechanism

No existing issue — describing inline (bug):

**What happened:** Recreating the `ai-paperclip` container (new image, same data volume) intermittently hit a transient failure on the very first boot after recreation, self-recovering only because Docker's restart policy retried a few seconds later.

**Root cause:** `postmaster.pid` naming a live PID is not proof Postgres is done starting. The pidfile can point at this same app's own previous embedded-Postgres process, which is still replaying WAL / not yet listening on its port. The startup code treated a live PID as sufficient and reused the port immediately, so the first connection attempt got `ECONNREFUSED` and the process exited instead of waiting it out.

**Expected:** A container recreation with an inherited data directory starts cleanly on the first attempt, without depending on a restart-policy retry.

**Repro:** Recreate the container right after a fresh image build (`docker compose up --force-recreate` equivalent) against a data volume whose embedded Postgres was mid-WAL-replay from the previous process's own startup — the first boot attempt errors with `ECONNREFUSED` against `postmaster.pid`'s named PID.

_Dedup search: no open PR touches `server/src/index.ts`'s embedded-Postgres startup path. No duplicate found._

## What Changed

- `server/src/index.ts`: when `postmaster.pid` names a live PID, verify it's actually reachable — using the same live-connection proof (`getPostgresDataDirectory` + `ensurePostgresDatabase`) the no-pidfile branch already performs — retrying up to 5 times with a 1s backoff before giving up.
- Only once the process never becomes reachable across those retries is the pidfile treated as a genuinely stale lock (e.g. left over from a data directory carried over some other way): it's removed and the normal initialize/start path runs instead.
- When the process *is* reachable within the retry budget, behavior is unchanged from before (reuse the existing process, log the same message).

## Verification

- `tsc --noEmit`: no new errors in `server/src/index.ts` (this monorepo checkout has pre-existing, unrelated typecheck failures elsewhere — `environment-runtime.ts`, `plugin-host-services.ts`, `plugin-worker-manager.ts` — from a stale local `@paperclipai/plugin-sdk` build; none touch this file or this change).
- No dedicated test harness exists for this startup path (it depends on real process/pidfile state at boot, not something the existing DB-backed test suite exercises); verified by code review and by the fact this is the identical live-connection check the adjacent no-pidfile branch already relies on, just retried with backoff instead of run once.
- Verified live against our own deployment (a fork running this exact code): rebuilt the image, recreated the container, and confirmed clean startup logs with no `ECONNREFUSED` / restart-loop, where the same recreation previously needed the crash-loop retry to recover.

## Risks

Low risk. Adds up to 5 extra connection attempts (1s apart, so ≤5s) only on the already-slow-path of finding a live `postmaster.pid`; the common cold-start-with-no-pidfile path is untouched. Worst case on a genuinely stale pidfile: startup takes ~5s longer before falling back to the existing stale-lock-cleanup behavior, which was already the fallback.

## Model Used

Claude Sonnet 5 (`claude-sonnet-5`), via Claude Code — root-caused from a real container-recreation failure against a live deployment, then verified the fix against the same live deployment.

## Checklist

- [x] My branch name describes the change (`fix/...`) and contains no internal ticket id
- [ ] I have added or updated tests where applicable (no existing harness covers real-process/pidfile startup state; see Verification)
- [x] I have updated relevant documentation to reflect my changes (none required — internal startup behavior)

### DIFF
--- a/server/src/index.ts
+++ b/server/src/index.ts
@@ -442,11 +442,58 @@ export async function startServer(): Promise<StartedServer> {
       }
     };
   
+    // A `runningPid` match just means the OS reports that PID as alive right now -
+    // it doesn't mean postgres has finished starting yet. Observed in practice
+    // right after a container recreation (new image, `docker compose up`
+    // recreate): the pidfile names a real, freshly-spawned postgres (this app's
+    // own previous start of it) that is still replaying WAL / not yet listening,
+    // so the very first connection probe gets ECONNREFUSED and the process used
+    // to exit hard here - only surviving via Docker's restart-policy crash-loop
+    // retrying a few seconds later. Retry the same live-connection proof the
+    // no-pidfile branch already uses below a few times with backoff before
+    // giving up; only after it never becomes reachable do we treat the pidfile
+    // as a genuinely stale lock (e.g. from a data dir carried over some other
+    // way) and clear it so the normal initialise/start path below can run.
     const runningPid = getRunningPid();
+    const configuredAdminConnectionString = `postgres://paperclip:paperclip@127.0.0.1:${configuredPort}/postgres`;
+    let reusingExistingServer = false;
     if (runningPid) {
-      logger.warn(`Embedded PostgreSQL already running; reusing existing process (pid=${runningPid}, port=${port})`);
+      const verifyReachable = async () => {
+        const actualDataDir = await getPostgresDataDirectory(configuredAdminConnectionString);
+        if (typeof actualDataDir !== "string" || resolve(actualDataDir) !== resolve(dataDir)) {
+          throw new Error("reachable postgres does not use the expected embedded data directory");
+        }
+        await ensurePostgresDatabase(configuredAdminConnectionString, "paperclip");
+      };
+      const maxAttempts = 5;
+      let lastErr: unknown;
+      for (let attempt = 1; attempt <= maxAttempts; attempt += 1) {
+        try {
+          await verifyReachable();
+          reusingExistingServer = true;
+          break;
+        } catch (err) {
+          lastErr = err;
+          if (attempt < maxAttempts) await new Promise((resolve) => setTimeout(resolve, 1000));
+        }
+      }
+      if (reusingExistingServer) {
+        logger.warn(`Embedded PostgreSQL already running; reusing existing process (pid=${runningPid}, port=${port})`);
+      } else {
+        logger.warn(
+          { err: lastErr, pid: runningPid, attempts: maxAttempts },
+          "postmaster.pid names a live PID but postgres never became reachable there - treating as a stale lock and starting fresh",
+        );
+        try {
+          rmSync(postmasterPidFile, { force: true });
+        } catch (rmErr) {
+          logger.warn({ err: rmErr }, "failed to remove stale postmaster.pid; embedded-postgres start may still fail");
+        }
+      }
+    }
+    if (reusingExistingServer) {
+      // already logged and verified above; fall through with the configured port
     } else {
-      const configuredAdminConnectionString = `postgres://paperclip:paperclip@127.0.0.1:${configuredPort}/postgres`;
       try {
         const actualDataDir = await getPostgresDataDirectory(configuredAdminConnectionString);
         if (


server/src/index.ts
```diff
@@ -442,11 +442,58 @@ export async function startServer(): Promise<StartedServer> {
...
+      }
+      if (reusingExistingServer) {
+        logger.warn(`Embedded PostgreSQL already running; reusing existing process (pid=${runningPid}, port=${port})`);
+      } else {
+        logger.warn(
+          { err: lastErr, pid: runningPid, attempts: maxAttempts },
+          "postmaster.pid names a live PID but postgres never became reachable there - treating as a stale lock and starting fresh",
+        );
+        try {
+          rmSync(postmasterPidFile, { force: true });
```
adrian-ludwig: HAL-LIVE-CANARY-5b4dd174-3bec-4c03-89e0-cecf4cccd453-FINDING This retry loop can end up starting a second embedded Postgres against the same data directory while the first one is still alive and recovering. getRunningPid() proves the pid is alive with process.kill(pid,0) before the loop starts. But the loops failure mode is a 5-attempt, roughly 4-second timeout on a live-connection probe, and a Postgres server still replaying WAL responds to TCP but rejects queries with the database system is starting up, which getPostgresDataDirectory turns into the same null/error as an unreachable port. When the timeout is hit, the code deletes postmaster.pid at line 488 without re-checking whether the pid is still alive, then falls through to the fresh-start path, which calls embeddedPostgres.start() at line 547 against the same data directory whose lock file it just removed. On a data directory with more than a few seconds of WAL to replay, a realistic amount for a production database after an unclean recreation, this either reproduces the original crash later, or races a second postmaster onto the still-live ones data directory. Re-check isPidRunning(runningPid) before deleting the pidfile and starting fresh; only treat the lock as stale once the recorded pid is confirmed dead, not unreachable.

server/src/index.ts
```diff
@@ -442,11 +442,58 @@ export async function startServer(): Promise<StartedServer> {
...
+    // no-pidfile branch already uses below a few times with backoff before
+    // giving up; only after it never becomes reachable do we treat the pidfile
+    // as a genuinely stale lock (e.g. from a data dir carried over some other
+    // way) and clear it so the normal initialise/start path below can run.
     const runningPid = getRunningPid();
+    const configuredAdminConnectionString = `postgres://paperclip:paperclip@127.0.0.1:${configuredPort}/postgres`;
+    let reusingExistingServer = false;
     if (runningPid) {
-      logger.warn(`Embedded PostgreSQL already running; reusing existing process (pid=${runningPid}, port=${port})`);
+      const verifyReachable = async () => {
```
adrian-ludwig: HAL-LIVE-CANARY-5b4dd174-3bec-4c03-89e0-cecf4cccd453-FINDING The reachability check, getPostgresDataDirectory, compare resolved paths, ensurePostgresDatabase, is duplicated verbatim between the new pidfile-retry loop (lines 461-467) and the existing no-pidfile branch (lines 498-505). Extract one shared helper and call it from both places, so a future fix to the check does not have to be applied twice.

server/src/index.ts
```diff
@@ -442,11 +442,58 @@ export async function startServer(): Promise<StartedServer> {
       }
     };
   
+    // A `runningPid` match just means the OS reports that PID as alive right now -
```
adrian-ludwig: HAL-LIVE-CANARY-5b4dd174-3bec-4c03-89e0-cecf4cccd453-FINDING The comment above const runningPid = getRunningPid() (lines 445-456) runs 12 lines, over the repos 3-line comment cap, and one clause narrates prior behavior ("the process used to exit hard here - only surviving via Dockers restart-policy crash-loop retrying a few seconds later"), which belongs in the PR description, not the file. Trim it to one short, timeless statement of the non-obvious constraint: a live pid does not prove Postgres is ready for queries yet, and drop the historical framing.

server/src/index.ts
```diff
@@ -442,11 +442,58 @@ export async function startServer(): Promise<StartedServer> {
...
+    // giving up; only after it never becomes reachable do we treat the pidfile
+    // as a genuinely stale lock (e.g. from a data dir carried over some other
+    // way) and clear it so the normal initialise/start path below can run.
     const runningPid = getRunningPid();
+    const configuredAdminConnectionString = `postgres://paperclip:paperclip@127.0.0.1:${configuredPort}/postgres`;
+    let reusingExistingServer = false;
     if (runningPid) {
-      logger.warn(`Embedded PostgreSQL already running; reusing existing process (pid=${runningPid}, port=${port})`);
+      const verifyReachable = async () => {
+        const actualDataDir = await getPostgresDataDirectory(configuredAdminConnectionString);
```
adrian-ludwig: HAL-LIVE-CANARY-9e5c17fa-655d-4297-a773-d88ce5d351d2-FINDING verifyReachable (line 462) duplicates the same three-step reachability check - data-directory match then ensurePostgresDatabase - that the else-branch below already runs against the same connection string. The PR description even notes it's reusing 'the same live-connection proof.' Extract that logic into one helper and call it from both the retry loop and the no-pidfile branch, so a future change to what 'reachable' means can't land in only one of the two copies.

server/src/index.ts
adrian-ludwig: HAL-LIVE-CANARY-9e5c17fa-655d-4297-a773-d88ce5d351d2-FINDING This comment runs 12 lines, over the repo's 3-line cap for new comments, and reads as incident narration ('Observed in practice ...') rather than a timeless invariant. Trim it to state the invariant only - a live PID does not mean postgres is reachable yet, so retry before treating the pidfile as stale - and leave the incident story in the PR description.

server/src/index.ts
adrian-ludwig: HAL-LIVE-CANARY-c0de6f27-0a4d-46b9-932b-568e00da7169-FINDING When the retries never confirm the live postmaster.pid process is reachable, the code deletes postmaster.pid and starts a brand-new embedded Postgres instance in the same data directory, without ever re-checking whether that original process actually died. getRunningPid() already proved the PID was alive before the retry loop started; the retries only test whether a connection succeeds, not whether the process exited. If WAL replay simply takes longer than the roughly 4-second retry budget (a large database, a slow disk, an unclean prior shutdown are all plausible on container recreation), the still-alive process gets its lock file deleted out from under it, and a second Postgres process starts against the same data directory. Two Postgres backends writing the same data directory concurrently risk real corruption, which is worse than the crash-loop this PR is meant to fix. The codebase already has the right pattern for this exact hazard: the supervisor's beforeRestart callback re-checks getRunningPid() and refuses to proceed if the process is still alive before touching postmaster.pid. Add that same liveness re-check here before removing the pidfile: proceed with cleanup only once the process is confirmed dead, and keep waiting or fail loudly instead of starting a second instance when it's alive but unreachable.

server/src/index.ts
adrian-ludwig: HAL-LIVE-CANARY-c0de6f27-0a4d-46b9-932b-568e00da7169-FINDING The new retry loop's reachability check (server/src/index.ts:462-466) and the pre-existing no-pidfile check (497-505) are the same logic copied twice: the same getPostgresDataDirectory call, the same data-directory equality check, and the same ensurePostgresDatabase call. Pull this into one shared helper so the two cold-start paths this PR is unifying can't drift apart the next time either one changes.

server/src/index.ts
adrian-ludwig: HAL-LIVE-CANARY-c0de6f27-0a4d-46b9-932b-568e00da7169-FINDING This comment is 12 lines long, over the repo's 3-line cap for new comments, and spends most of its length narrating the container-recreation incident that motivated the fix rather than stating the invariant. Trim it to the one fact a reader needs here: a live PID in postmaster.pid doesn't prove Postgres finished starting, so it must be probed with retry before reuse. Leave the incident story to the PR description.

PR #12524, opened 2026-08-30
fix(claude-local): require safe live hello probes for ACP checks

rannte85:
## Thinking Path

> - Paperclip manages AI agents and tests whether each agent adapter can work.
> - The Claude local adapter supports CLI and ACP execution paths.
> - The ACP environment check can report success without a live provider response in some authentication paths.
> - A false success can allow later work to start with an unusable provider connection.
> - The existing CLI probe already has useful result classes, but the ACP path does not share the same complete contract.
> - This pull request uses one safe live hello probe for both paths and keeps raw provider data private.
> - The benefit is a consistent readiness result that proves a real provider round trip.

## Linked Issues or Issue Description

Refs #9933

**What happened?**

The Claude ACP environment check could report success from credential presence or process exit state without a positive live hello result. Some API key and Bedrock paths did not require the same provider round trip as the CLI path. Raw process data could also cross the public diagnostic boundary in failure paths.

**Expected behavior**

Every supported Claude authentication path must run the same safe live probe. A successful check must include the fixed positive probe result. Failure results must use fixed public codes and messages. Public checks must not include raw provider output, credentials, or local session paths.

**Steps to reproduce**

1. Configure the Claude local adapter for ACP with an API key, stored login, host authentication, or Bedrock.
2. Run the adapter environment check with a provider state that cannot return the exact hello response.
3. Observe that the old ACP path can skip the live round trip or return no positive probe code.

**Paperclip version or commit**

The change is based on master commit `bc9ba7cd2` and includes current master in merge commit `6ecca30fc`.

**Deployment mode**

Local development from source.

## What Changed

- Added one shared Claude hello probe for CLI and ACP environment checks.
- Built probe arguments from a fixed allowlist that disables tools, Chrome, hooks, plugins, permission bypasses, and session persistence.
- Required a live provider round trip for API key, OAuth or subscription, stored configuration, host authentication, and Bedrock paths.
- Added fixed result classes for authentication, usage limit, transient upstream, timeout, unexpected output, and unknown failures.
- Removed raw stdout, stderr, thrown errors, credentials, and local paths from public check details and metadata.
- Added regression coverage for local, SSH, sandbox, authentication, redaction, and failure classification paths.

## Verification

- Ran the five focused Vitest files on Node v24.11.1. All 90 tests passed.
- Ran the `@paperclipai/adapter-claude-local` typecheck on Node v24.11.1. It passed.
[...]

Issue #10546, opened 2026-07-31
claude_local: Bedrock model catalogue hardcodes us.anthropic.*, so built-in agents cannot be provisioned outside US regions


Tiago-Godinho:
## Summary

When `claude_local` runs against AWS Bedrock, the model catalogue is a hardcoded list of five `us.anthropic.*` IDs. The region is never consulted. On any non-US Bedrock region this makes built-in agents impossible to provision: the bundled default model fails validation, and every model that *would* pass validation does not exist in the region and fails at runtime.

Verified against `master` @ `4813ed3` (also reproduced on the published `2026.722.0`).

## Environment

- `CLAUDE_CODE_USE_BEDROCK=1`
- `AWS_REGION=eu-west-1`
- `ANTHROPIC_API_KEY` unset (Bedrock is the only auth path)
- Claude Code CLI 2.1.220, Windows 11

## Root cause

`packages/adapters/claude-local/src/server/models.ts:11-17` hardcodes the catalogue:

```ts
/** AWS Bedrock model IDs — region-qualified identifiers required by the Bedrock API. */
const BEDROCK_MODELS: AdapterModel[] = [
  { id: "us.anthropic.claude-opus-4-8-v1", label: "Bedrock Opus 4.8" },
  { id: "us.anthropic.claude-fable-5-v1", label: "Bedrock Fable 5" },
  { id: "us.anthropic.claude-opus-4-6-v1", label: "Bedrock Opus 4.6" },
  { id: "us.anthropic.claude-sonnet-4-5-20250929-v2:0", label: "Bedrock Sonnet 4.5" },
  { id: "us.anthropic.claude-haiku-4-5-20251001-v1:0", label: "Bedrock Haiku 4.5" },
];
```

and `models.ts:105` returns it unconditionally, with no region input:

```ts
if (isBedrockEnv()) return dedupeModels(BEDROCK_MODELS);
```

The comment on line 10 says the IDs are "region-qualified", which is true of the format but not of the values. Supporting evidence that region is simply not plumbed here:

- Across the entire `claude-local` adapter, `AWS_REGION` appears **only inside two user-facing hint strings** — `acp.ts:517` and `test.ts:269` (`"Ensure AWS credentials and AWS_REGION are configured"`). It is never read. The adapter asks the operator to configure `AWS_REGION` and then ignores it.
- `eu.anthropic` appears **once in the whole repository**, in a comment (`models.ts:161`).
[...]

Issue #10243, opened 2026-07-25
Claude Bedrock model profiles and bundled defaults use unavailable direct aliases


roland-at-ailtir:
### What happened?

On a self-hosted Paperclip instance configured with `CLAUDE_CODE_USE_BEDROCK=1` and an EU AWS region, Claude model discovery correctly returns region-qualified Bedrock inference-profile IDs. However, the static `cheap` model profile still selects the direct Anthropic alias `claude-sonnet-4-6`, and the bundled Summarizer defaults to `claude-haiku-4-5`.

Creating or importing a company can therefore fail while auto-provisioning the bundled Summarizer with:

```text
Model "claude-haiku-4-5" is not available for adapter claude_local.
```

This is related to, but not fully covered by, #6101.

### Expected behavior

Claude model profiles and bundled-agent defaults should resolve against the same environment-aware model catalogue as the adapter. In Bedrock mode they should use a region-qualified model ID that the instance reports as available.

### Steps to reproduce

1. Configure Paperclip with `CLAUDE_CODE_USE_BEDROCK=1` and `AWS_REGION=eu-west-1`.
2. Confirm the Claude adapter lists `eu.anthropic.*` model IDs.
3. Inspect the `cheap` profile or create a company that auto-provisions Summarizer.
4. Observe that a direct Anthropic alias is selected and rejected.

### Paperclip version or commit

`cca2806e57c272ed919b1cde224d3d2a23d2f668`

### Deployment mode

Self-hosted server

### Installation method

Built from source

### Agent adapter(s) involved

Claude Code

### Database mode

External Postgres

### Access context

Board

### Node.js version

v22

### Operating system

Linux

### Additional context

A replacement PR will preserve direct-Anthropic defaults outside Bedrock, derive the Bedrock `cheap` profile from the region-aware catalogue, and let bundled defaults fall back to that valid profile when their static model is unavailable. Credit to #6101 for identifying the dynamic-profile portion of the problem.

### Privacy checklist

- [x] I have reviewed all pasted output for PII and redacted sensitive values.

kimnamu:
Thanks for the detailed report — this reproduces on current `master` (`d1573244`), and the two halves you describe fail in different ways. Worth separating them, because a change that only satisfies validation closes the first and leaves the second silent.

## Symptom reproduces

```
[...]

PR #12624, opened 2026-09-01
fix(claude-local): require safe live hello probes for ACP checks

rannte85:
## Thinking Path

> - Paperclip is an open source app that manages AI agents for work.
> - Adapters connect Paperclip to local AI providers.
> - The Claude local adapter can test an ACP environment before it starts work.
> - The previous test could accept an exit code without a positive provider reply.
> - It could also use an unsafe permission bypass for a sandbox probe.
> - This pull request adds a strict live hello probe and safe tool restrictions.
> - The benefit is a test result that means the provider answered successfully.

**What happened?**

The Claude local adapter could report that an ACP environment was ready without a positive hello signal from the provider. Some probe paths treated process exit as enough evidence. A sandbox probe could also use `--dangerously-skip-permissions`.

**Expected behavior**

The environment test must pass only after a live provider hello returns an allowlisted success code. The probe must redact secrets. A sandbox probe must use a narrow `--allowedTools` value.

**Steps to reproduce**

1. Configure a `claude_local` ACP target.
2. Run the adapter environment test with a stored key, an API key, or a Bedrock lane.
3. Observe that the previous implementation can skip a live hello or accept exit status without explicit positive evidence.

`master` at `0b73ebb86c9bc97675b789c7d9221e6da57ff66e`.

**Deployment mode**

Local development or a self-hosted adapter environment test.

Related work: #9933, #12524, #12523, #1114, #6674, and #3162.

## What Changed

- Add one shared hello probe for Claude ACP environment tests.
- Require an allowlisted positive success code before the test can pass.
- Redact provider output and credential material from the result.
- Use `--allowedTools` for sandbox probes.
- Cover stored-key, API-key, Bedrock, timeout, redaction, and unsafe-flag cases.

## Verification

- `pnpm --filter @paperclipai/adapter-claude-local typecheck`
- Five focused test files pass with 83 tests.
- `pnpm -r typecheck` passes on the local host.
- `pnpm build` passes on the local host.
- `git diff --check` passes.
- `pnpm test:run` reports 336 passed files, 16 failed files, and 112 skipped files. It reports 3,293 passed tests, 22 failed tests, and 2,112 skipped tests. The failures include embedded PostgreSQL startup, the macOS `/tmp` alias, invalid generated ports, and workspace runtime cases. One Claude environment assertion also fails only in the full run. The focused Claude adapter suite passes when it runs alone.
- Upstream CI must verify the full suite and build with Node.js 24.11 or later. The local host uses Node.js 22.22.2.

## Risks

- The probe now fails closed when it cannot find a positive provider reply.
- A provider CLI output change can require a success-code allowlist update.
- The change does not alter stored credentials or expose provider output.
- The change does not run a real credential probe during this contribution step.

## Model Used

OpenAI Codex with `gpt-5.6-sol`. The session used tool calls and code execution. The public context-window size is not specified.

## Checklist


kimnamu (commented):
Thanks for picking this up — "the Test passed" meaning the provider actually answered is a real improvement, and taking the Bedrock lane off the skip path is the part I care about most. Not a maintainer here, just someone who runs `claude_local` against Bedrock (#2793, #3033). I applied your diff onto master `8f9f850`, ran the suites and captured the probe argv/env. Three things, two of them Bedrock-specific.

### 1. Bedrock is the one auth mode with no host-env seed, so the now-mandatory probe fails setups the real run serves

`acp.ts:745-769` seeds four host credentials into `probeEnv` (`ANTHROPIC_API_KEY`, `CLAUDE_CODE_OAUTH_TOKEN`, `ANTHROPIC_AUTH_TOKEN`, `CLAUDE_CONFIG_DIR`), each with the same reason in the comment: without the seed a valid host setup reports a false auth-required and fails the lane. Bedrock has no equivalent block, so the probe only ever sees what `config.env` sets.

The real local ACP run *does* inherit the host AWS variables — `ACPX_INHERITED_PROVIDER_ENV_KEYS.claude` in `packages/adapter-utils/src/acpx-engine/execute.ts:551-570` lists `AWS_ACCESS_KEY_ID`, `AWS_SECRET_ACCESS_KEY`, `AWS_SESSION_TOKEN`, `AWS_PROFILE`, `AWS_REGION`, `AWS_CONFIG_FILE`, `AWS_SHARED_CREDENTIALS_FILE`. I compared the two on your branch, local target, credentials in the server env, `config: { engine: "acp" }`:

```
realRunInheritsFromHost: ["AWS_ACCESS_KEY_ID","AWS_REGION","AWS_SECRET_ACCESS_KEY","AWS_SESSION_TOKEN"]
helloProbeChildEnv:      []
```

The probe child gets nothing at all — not even `CLAUDE_CODE_USE_BEDROCK`, so it isn't in Bedrock mode either. Add `config.env: { CLAUDE_CODE_USE_BEDROCK: "1" }` with the credentials still on the host and it becomes `["CLAUDE_CODE_USE_BEDROCK"]`: Bedrock mode, nothing to sign with. On master this didn't matter because the lane skipped the probe for Bedrock; now the outcome is `claude_hello_probe_failed` at `error`, so a Test that passed hard-fails. Greptile's P1 about `AWS_CONFIG_FILE`/`AWS_SHARED_CREDENTIALS_FILE` is the same gap — the whole AWS set is missing, not just the file paths.

Mirroring the four blocks you already have, after `acp.ts:769`:

```ts
    // Seed the host Bedrock/AWS variables the same way. A local ACP run
    // inherits them through the shared ACPX host projection
    // (ACPX_INHERITED_PROVIDER_ENV_KEYS.claude), so without this seed a valid
    // host Bedrock setup reports a false claude_hello_probe_failed and fails
    // the Test lane.
    if (considerHostEnv) {
      for (const key of BEDROCK_PROBE_ENV_KEYS) {
        const hostValue = process.env[key];
        if (isNonEmpty(hostValue) && !isNonEmpty(probeEnv[key])) {
          probeEnv[key] = hostValue.trim();
        }
      }
    }
```

with the list next to `MIN_ACP_NODE_VERSION`:

```ts
/**
 * The Bedrock/AWS variables the local ACP hello probe must receive from the
 * server env, mirroring ACPX_INHERITED_PROVIDER_ENV_KEYS.claude in
 * acpx-engine/execute.ts — the projection the real local run inherits.
 */
const BEDROCK_PROBE_ENV_KEYS = [
  "CLAUDE_CODE_USE_BEDROCK",
  "ANTHROPIC_BEDROCK_BASE_URL",
  "AWS_BEARER_TOKEN_BEDROCK",
  "AWS_ACCESS_KEY_ID",
  "AWS_SECRET_ACCESS_KEY",
  "AWS_SESSION_TOKEN",
  "AWS_REGION",
  "AWS_DEFAULT_REGION",
  "AWS_PROFILE",
  "AWS_CONFIG_FILE",
  "AWS_SHARED_CREDENTIALS_FILE",
] as const;
```

The two file variables also have to go on the probe's caller allowlist (`probe-env.ts:12-31`) or `buildLocalAdapterTestProbeEnv` drops them for local targets, which closes Greptile's P1 as well:

```ts
  "AWS_PROFILE",
  "AWS_CONFIG_FILE",
  "AWS_SHARED_CREDENTIALS_FILE",
  "AWS_REGION",
```

Regression test — goes in the local/SSH `describe` in `acp.auth.test.ts`, and the AWS keys need adding to that block's `CLEARED_HOST_ENV_KEYS` so it restores them:

```ts
  it("seeds the host AWS Bedrock credentials into the local hello probe", async () => {
    // A local ACP run inherits the host AWS credentials through the shared ACPX
    // host projection (ACPX_INHERITED_PROVIDER_ENV_KEYS.claude), so a probe
    // without them fails a Bedrock setup the real run can serve.
    probeResult.value = { exitCode: 0, stdout: helloStdout, stderr: "", timedOut: false };
    process.env.CLAUDE_CODE_USE_BEDROCK = "1";
    process.env.AWS_ACCESS_KEY_ID = "AKIAEXAMPLE";
    process.env.AWS_SECRET_ACCESS_KEY = "host-secret";
    process.env.AWS_REGION = "eu-west-1";

    const result = await testClaudeAcpEnvironment({
      companyId: "company-1",
      adapterType: "claude_local",
      config: { engine: "acp", env: { CLAUDE_CODE_USE_BEDROCK: "1" } },
      executionTarget: null,
      environmentName: null,
    });

    const call = runAdapterExecutionTargetProcess.mock.calls[0] as unknown as unknown[];
    const spawnedEnv = (call[4] as { env: Record<string, string> }).env;
    expect(spawnedEnv.AWS_ACCESS_KEY_ID).toBe("AKIAEXAMPLE");
    expect(spawnedEnv.AWS_SECRET_ACCESS_KEY).toBe("host-secret");
    expect(spawnedEnv.AWS_REGION).toBe("eu-west-1");
    expect(result.status).toBe("pass");
    // The credential value never enters a check.
    expect(JSON.stringify(result.checks)).not.toContain("host-secret");
  });
```

Without the `acp.ts` change it fails with `AssertionError: expected undefined to be 'AKIAEXAMPLE'`; with it, `acp.auth.test.ts` is 31/31.

### 2. The probe never sends the configured model, so a passing Bedrock Test says nothing about the model id

`buildClaudeHelloProbeArgs` (`hello-probe.ts:24-31`) rebuilds argv from the `--model`/`--effort` it finds in the caller args. The CLI lane puts `--model` there (`test.ts:274-276`); the ACP lane's args are just `["--print","-","--output-format","stream-json","--verbose"]` plus permission args (`acp.ts:561`), and `config.model` never becomes `ANTHROPIC_MODEL` for the probe. Same config through both lanes on your branch:

```
CLI lane probe argv: [..., "--model", "us.anthropic.claude-opus-4-8-v1", "--safe-mode", ...]
ACP lane probe argv: ["--print","-","--output-format","stream-json","--verbose","--safe-mode", ...]   (no --model)
ACP lane probe env:  ANTHROPIC_MODEL absent
```

The real ACP run does set it — `acpx-engine/execute.ts:1915`, `env.ANTHROPIC_MODEL = requestedModel`, with the comment above it explaining why that and not `session/set_config_option`. So the probe validates whatever the CLI defaults to rather than the id the agent will actually run with. On Bedrock that's the failure that matters: a wrong region prefix or a profile the account can't invoke is the common one (#10546, #10551, #10243). Concretely, three of the five entries in `models.ts` `BEDROCK_MODELS` return `ValidationException` from Converse against a real account — I measured them in [#10551](https://github.com/paperclipai/paperclip/pull/10551#issuecomment-5148780246). Pick one of those in the dropdown today and this lane still reports "Claude hello probe succeeded."

In `probeClaudeAcpSandboxLogin`, right after `callerEnv` is built:

```ts
  // The real ACP run sets the requested model through ANTHROPIC_MODEL at ACP
  // startup (acpx-engine/execute.ts). Seed the same variable so the probe
  // exercises the configured model instead of the CLI default. This is what
  // makes the probe meaningful on Bedrock, where a model id the account or
  // region cannot serve is the common failure.
  const requestedModel = asString(config.model, "").trim();
  if (requestedModel && !isNonEmpty(callerEnv.ANTHROPIC_MODEL)) {
    callerEnv = { ...callerEnv, ANTHROPIC_MODEL: requestedModel };
  }
```

`ANTHROPIC_MODEL` is already on the probe allowlist (`probe-env.ts:18`), so it reaches the child. I left out the CLI lane's `isBedrockModelId` guard on purpose: the ACP run forwards `config.model` verbatim, so the probe should too — if the id is wrong, failing here is the whole point.

```ts
  it("sends the configured Bedrock model to the hello probe", async () => {
    probeResult.value = { exitCode: 0, stdout: helloStdout, stderr: "", timedOut: false };

    await probeClaudeAcpSandboxLogin({
      config: {
        engine: "acp",
        model: "us.anthropic.claude-opus-4-6-v1",
        env: { CLAUDE_CODE_USE_BEDROCK: "1", AWS_REGION: "eu-west-1" },
      },
      target: sandboxTarget,
    });

    const call = runAdapterExecutionTargetProcess.mock.calls[0] as unknown as unknown[];
    const spawnedEnv = (call[4] as { env: Record<string, string> }).env;
    // The real ACP run selects the model through ANTHROPIC_MODEL, so a probe
    // that omits it validates the CLI default instead of the configured id.
    expect(spawnedEnv.ANTHROPIC_MODEL).toBe("us.anthropic.claude-opus-4-6-v1");
  });
```

Red before the change: `AssertionError: expected undefined to be 'us.anthropic.claude-opus-4-6-v1'`.

### 3. An existing server test breaks, and it isn't a full-run artifact

`server/src/__tests__/claude-local-adapter-environment.test.ts:359` asserts the sandbox probe argv contains `--allowedTools`; the rebuild drops it in favour of `--tools ""`. Run in isolation, same shell:

```
clean master 8f9f850:  Test Files  1 passed (1)   Tests  10 passed (10)
+ your diff:           Test Files  1 failed (1)   Tests  1 failed | 9 passed (10)

AssertionError: expected [ '--print', '-', …(12) ] to include '--allowedTools'
```

I suspect that's the "one Claude environment assertion" your description attributes to the full run. Related: `buildClaudeProbePermissionArgs` is still called at `acp.ts:563` and `test.ts:268` and then thrown away by the rebuild, so it's dead on the probe path — either keep `--allowedTools` in the allowlisted rebuild or update that test with the reasoning, since its comment says the flag is there so a tool call can't stall on an approval nobody is around to answer.

Not blocking, just flagging: `--safe-mode`, `--disable-slash-commands`, `--no-chrome` and `--no-session-persistence` don't appear anywhere else in the repo's `claude` argv builders (`--tools ""` does — `quota.ts:433`). Since an unrecognized flag now surfaces as `claude_hello_probe_failed`/`error` rather than a skip, it may be worth gating them the way `--effort` is gated in `cli-capabilities.ts`.

Verified on master `8f9f850` + your diff, node 22.22.1: `adapter-claude-local` at 262 passed / 1 skipped with both changes and the two tests above, and `pnpm --filter @paperclipai/adapter-claude-local typecheck` clean. The direction here is right and I'd like to see it land — happy to send the Bedrock seeding as a follow-up PR if you'd rather keep this diff focused.

PR #12646, opened 2026-09-01
fix(runner): accept the indeterminate command result after a runner restart

zannis:
## Thinking Path

> - The runner subsystem pairs a Rust runner process with a durable control plane in TypeScript. The control plane is the authority for every command the runner executes
> - That pair has a crash-recovery contract. If the runner dies between journaling a command and confirming the command's effect, it must not run the command twice
> - The runner keeps its side of the contract. On restart it promotes such a command to the `indeterminate` status and reports that status back
> - The control plane did not accept `indeterminate`. It closed the connection without a diagnostic, the runner reconnected and replayed the same result, and the loop repeated forever
> - This pull request accepts `indeterminate` as a terminal command status
> - The benefit is that a session survives a runner crash during a tool call, instead of hanging until a 30 second deadline expires

No public issue exists for this defect, so it is described here.

**What happened?**

A live session cannot resume after the runner process is killed during a
governed tool call. The resumed transport waits for the provider identity for
30 seconds and then fails with `runnerd did not report its provider identity`.

`packages/paperclip-runner/src/live/live-session.test.ts` covers this exact
sequence in "terminates real runnerd after a durable receipt and resumes its
exact provider thread". That test has a 15 second budget, so it reports the
defect as `Test timed out in 15000ms` and reads like a flake.

**Expected behavior**

The resumed control plane accepts the runner's recovery report, the runner
reports its provider identity, and the session resumes on its original
provider thread.

**Steps to reproduce**

Build the runner binary, then run the test:

```
cargo build --manifest-path packages/paperclip-runner/runner/Cargo.toml --locked --workspace --bins
cd packages/paperclip-runner
npx vitest run src/live/live-session.test.ts -t "terminates real runnerd"
```

It fails every time on an idle machine. It also fails at `560e7e48b`, the
commit that added the test, so the defect is not a recent regression.

Reproduced on `master` at `0a422fda5`, which is the base of this branch.

**Deployment mode**

Local development, running the package test suite.

**Root cause**

`DurablePrpControlPlane.#commandResult` accepted only `completed`, `failed`
and `rejected`. The runner reports a journaled-but-unconfirmed command as:

```json
{ "status": "indeterminate",
  "result": { "code": "execution_indeterminate",
              "message": "runner recovered after journaling this command; it will not execute twice" } }
```

That status fell through to a silent `connection.close()`. The runner
reconnected after 250 ms, replayed the same result, and was closed again. No
durable event ever reached the control plane, so the transport never saw
`harness.ready`.

`indeterminate` is a deliberate part of the runner's contract. See
`reconcile_pending_commands` in
`packages/paperclip-runner/runner/crates/runner-core/src/durable/state.rs`.
The rest of the TypeScript code already models the status; only this control
plane did not.

## What Changed

- `DurablePrpControlPlane.#commandResult` accepts `indeterminate` as a
  terminal command status.
- The persisted-state validation accepts `indeterminate`, so a control plane
  restarted over the same directory can read its own saved state back.
  Without this, accepting the status would make the next restart throw.
- `DurableRecoveryCoreCommand.status` includes `indeterminate` in both
  declarations of that interface.
- Added an integration test that drives the exact recovery frame the runner
  sends. It asserts the connection stays open, the next command is delivered,
  the status is persisted, a restarted control plane reloads it, and a
  replayed duplicate is absorbed rather than treated as a conflict.

## Verification

All commands run from `packages/paperclip-runner`.

- New test fails before the change and passes after it. Before:
  `expected null to match object { kind: 'command' }` — `null` is the closed
  connection.
  `npx vitest run src/control-plane/durable-prp-control-plane.test.ts`
  → 4 passed.
- The live runner test that exposed this reproduced **deterministically** on an
  idle machine before the change, and now passes in 3.3 s, well inside its
  existing 15 s budget. Ran it 10 times in a row: 10/10 pass, 0 failures.
  `npx vitest run src/live/live-session.test.ts -t "terminates real runnerd"`
- Full package suite: `npx vitest run` → 1298 passed, 1 failed. The one
  failure is `src/mock-core/local-runner.test.ts > cleans up the harness
  process group when the controller closes`. It fails identically on an
  unmodified checkout in the same container, so it is a pre-existing
  environment issue and not related to this change.
- Typecheck: `tsc -p tsconfig.json --noEmit` → clean.

I did **not** raise the test's timeout. The budget was never the problem —
with a 600 s budget the same test still failed, at 31 s, with the real error.

## Risks

Low risk, and it widens rather than narrows what is accepted.

- Behaviour only changes for a status that is currently rejected, so no
  previously working path is affected.
- `indeterminate` is terminal, not successful. A caller waiting on such a
  command still receives an error from the transport, which is correct: the
  effect is genuinely unconfirmed. This change does not make an unconfirmed
  command look like it succeeded.
- The persisted-state change only widens an allow-list, so existing state
  files stay valid.

Open topics for a reviewer:

- The control plane closes connections without any diagnostic. That silence is
  why this defect looked like a flaky test. Adding a diagnostic channel is a
  larger change and is not included here.
- `DurableRecoveryProcessedCommand` in `src/contracts/durable-recovery.ts`
  drifts from the Rust `StoredCommandResult` by more than this status: it
  declares `commandDigest` and `logicalEffectCount`, which Rust does not have,
  and omits `commandType`, which Rust does. That is a separate correction and
  is deliberately not folded in here.

## Model Used

Claude Opus 5 (`claude-opus-5`), extended thinking, with tool use and code
execution.

Depends-on: none — this is a self-contained fix with no dependent changes.

## Checklist


zannis:
### Note for reviewers: the 15 s → 30 s budget bump in #12652 does not remove this failure

#12652 raised the per-test budget of `terminates real runnerd after a durable receipt and
resumes its exact provider thread` from `15_000` to `30_000` with the comment "CI exercises two
real process generations here and can exceed the unit default under load."

That is the same direction I ruled out before writing this fix, and it is now measurable on
master itself. Both runs below are on the current base (which already contains #12652 and
#12654), the only difference being whether this PR's four files are applied:

**master content — still fails, now at 30 s instead of 15 s**

```
❯ src/live/live-session.test.ts (18 tests | 1 failed | 17 skipped) 30021ms
    × terminates real runnerd after a durable receipt and resumes its exact provider thread 30017ms
Error: Test timed out in 30000ms.
```

**this PR's content on the same base — passes**

```
Test Files  1 passed (1)
     Tests  1 passed | 17 skipped (18)
  Duration  6.62s (tests 3.81s)
```

The stall is not load: with a very large budget the test fails at ~31 s with the real error
`runnerd did not report its provider identity`, which is the 30 s provider-identity deadline in
`runnerd-codex-transport.ts` expiring. The budget only decides how long CI waits before
reporting it, so the bump doubles the wasted CI time per occurrence rather than removing the
failure. The passing path here takes ~1.3–3.8 s, so it never needed 30 s.

The hang itself is what this PR removes: after a runner restart the control plane rejected the
runner's `indeterminate` command result, so the resumed turn never re-acquired the provider
thread. Locally the failure reproduces on the first attempt on an idle machine; on CI it is
intermittent (it also passes in ~1.2 s there), which is consistent with a race rather than a
load-related budget shortfall.

No conflict with #12652 — this PR does not touch `live-session.test.ts`; the raised budget can
stay or be lowered again once this lands.

zannis:
CI has now reproduced, unaided, the point I measured locally in the comment above: raising the per-test budget does not fix this test.

`master` picked up the `15_000 -> 30_000` bump in `bfb98aff5` (#12652). At **11:19:26Z** the same test failed on a branch carrying that base:

```
FAIL  src/live/live-session.test.ts > Capability live runnerd and Codex session
      > terminates real runnerd after a durable receipt and resumes its exact provider thread
Error: Test timed out in 30000ms.
Test Files  1 failed | 120 passed (121)
     Tests  1 failed | 1369 passed (1370)
```

<https://github.com/paperclipai/paperclip/actions/runs/33501605430/job/99836222465>

That run's `ci / verify` also went red, purely as the rollup of `BUILD_RESULT: failure` — so one occurrence turns two checks red on an otherwise clean branch. It is the sixth occurrence on the sixth unrelated branch today.

The budget cannot be the lever. With a large budget the test fails at ~31 s with the underlying error `runnerd did not report its provider identity` (the 30 s provider-identity deadline in `runnerd-codex-transport.ts`): any budget under ~31 s times out, any budget over it fails outright. With this PR's four files applied, the same test passes in 3.81 s on the same base. The net effect of the bump alone is that each occurrence now burns 30 s of CI instead of 15 s.

No conflict with #12652 either way — this PR does not touch `live-session.test.ts`. Branch is green (30 checks), `CLEAN`, and unchanged since the review.

nickyleach:
Another good one, thanks @zannis!

PR #12673, opened 2026-09-01
refactor(docker): declare the build stage's C toolchain explicitly

devinfoley:
## Thinking Path

> - The Docker images compile the Rust runner (paperclip-runnerd) during the build stage
> - The previous change replaced apt's cargo/rustc with a pinned rustup install
> - apt's cargo package pulled gcc in as a dependency; rustup does not install a C toolchain
> - Every master docker build now fails with "linker `cc` not found" in the build scripts
> - This pull request installs gcc, libc6-dev, and pkg-config explicitly in the build stage
> - The benefit is that the build stage declares its own C toolchain instead of inheriting one by accident

**What happened?**

Master `docker.yml` builds fail in `pnpm --filter @paperclipai/server build`: cargo build scripts (`libc`, `quote`, `proc-macro2`) die with `` error: linker `cc` not found ``, and the job exits with code 101.

**Expected behavior**

Master docker builds compile the runner and publish images.

**Steps to reproduce**

1. Run the `docker.yml` workflow on current `master`.
2. Observe the `build-and-push` job fail with the linker error above.

`master` after commit `317394456` (the rustup change); example failing run: docker.yml on `86ebdf842`.

## What Changed

- The build stage installs `gcc`, `libc6-dev`, and `pkg-config` explicitly, with a comment recording why: the old apt cargo brought gcc in as a dependency and the pinned rustup install does not.

## Verification

- Replicated the build stage's exact package sequence in `node:24-trixie-slim` — base-stage packages only (which include no compiler), then this new line, then the pinned, checksum-verified rustup install: the runner compiles (`Finished release`) with `rustc 1.97.1` and `cc 14.2.0`.
- Note: `docker.yml` triggers only on master pushes, so a PR run cannot exercise the image build itself; the container replication above is the pre-merge check. No test files: build-infrastructure fix (`fix:` on the Dockerfile only).

## Risks

- Low risk. Three packages added to the build stage only; the production stage is unchanged.

## Model Used

- Claude Fable 5 (`claude-fable-5`), Anthropic — Claude Code harness, extended thinking + tool use.

## Checklist

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -58,6 +58,13 @@ WORKDIR /app
 # version-pinned, checksum-verified installer and let the runner's own
 # rust-toolchain.toml choose the compiler — one pin, owned by the runner
 # package, shared by CI and image builds alike.
+#
+# The C toolchain is explicit: apt's cargo used to pull gcc in as a
+# dependency, and rustup does not — without it every build script dies on
+# "linker `cc` not found".
+RUN apt-get update \
+  && apt-get install -y --no-install-recommends gcc libc6-dev pkg-config \
+  && rm -rf /var/lib/apt/lists/*
 ENV RUSTUP_HOME=/usr/local/rustup \
     CARGO_HOME=/usr/local/cargo \
     PATH=/usr/local/cargo/bin:$PATH


devinfoley:
@greptile review

tonio-alucema:
@devinfoley could you merge this when you get a moment? It is `CLEAN`, 29 checks passing, and it unblocks more than it looks like.

`docker.yml` has failed on every master commit since 2026-08-30 — 74 runs, no image published in three days, so nothing merged since then has reached staging. It has stayed quiet because that workflow runs on master rather than on pull requests: PRs go on reporting green while master is red and staging silently stops moving. Background in #12677.

Two things worth a look before you hit the button:

**`make` may still be missing.** The same failing build log shows `gyp ERR! stack Error: not found: make` from `ssh2`, alongside the `linker \`cc\` not found` this fixes. `build-essential` would pull `make` in; `gcc libc6-dev pkg-config` does not. If that gyp step runs in the same stage, this fixes the linker and the next failure is waiting behind it.

**#12636 does the same job** with `build-essential`. Yours is the narrower image, so it seems the better one to keep — worth closing the other so they do not race.

Issue #12687, opened 2026-09-01
issue write-denial code (denyIssueWrite) is never logged server-side


matt-palmetto:
## Summary

`denyIssueWrite` (`server/src/routes/issues.ts:3922`) returns the issue-write denial `code` in the HTTP response body, but it is **never logged server-side**. Grepping the running pod's logs for a denial code returns nothing — this entire class of 403 (see #12686) is undiagnosable from operations alone; you need a live client-side capture of the response body at the moment of failure.

## Confirmed by reading the code

- `denyIssueWrite` writes the response directly: `res.status(status).json({ error: body.error, details: {...} })` (`issues.ts:3929-3934`) — it never calls `logger.warn`/`logger.error`.
- The pino-http request logger (`server/src/middleware/logger.ts`) only attaches rich error context (`errorContext`, via `customProps`, `logger.ts:49-58`) when `res.__errorContext` has been set.
- `res.__errorContext` is set in exactly two places: the central thrown-`HttpError` handler (`server/src/middleware/error-handler.ts:40`) and one manual call site in `server/src/routes/plugins.ts:1350`. `denyIssueWrite` bypasses both — it responds directly rather than throwing an `HttpError` — so its 403 falls through to the generic `res.statusCode >= 400` branch in `logger.ts`, which logs only `reqBody`/`reqParams`, **not** the response body. The denial `code` that would explain *why* is lost.

## Ask

Log at `warn` level inside `denyIssueWrite` (or `issueWriteDenialResponse`'s caller) with at least: `code`, `issueId`, `actorAgentId`, `runId` (when present). This is the concrete blocker to root-causing #12686's exact policy branch without a live repro.

See #12686 for the incident this surfaced from.

Issue #12686, opened 2026-09-01
Agent's own terminal-status PATCH returns 403, stranding issues as blocked


matt-palmetto:
## Summary

Agents that own an issue's checkout cannot reliably close it out: their own terminal-status `PATCH /api/issues/<id>` intermittently returns **403**, the issue falls back to `blocked`, and it re-strands in a loop even after manual recovery.

## Evidence (2026-09-01, paperclip.palmettointeractive.com, company `b738010a-67bb-4efc-9ee1-7890cd645d1f`)

- Issue **PAL-117** (`0aaa0411-7d15-4b91-982b-760f63e36c41`). Watchdog agent `a1416bb0-9af2-4fef-9c2e-6ce5672ff82d` ran, issued `PATCH /api/issues/0aaa0411-7d15-4b91-982b-760f63e36c41` with `{"status":"done","comment":...}`, and got **403**. The issue fell back to `blocked`.
- Cleared via the board escape hatch (`npx paperclipai issue recovery:resolve <id> --outcome restored --source-issue-status done`) at ~19:10Z, then **re-stranded by 19:24Z** — a repeating loop, not a one-off.
- ~10 incident issues have been stranded this way since 8/31.
- Every run on PAL-117 (`c00bb1cd`, `7d3ee783`, `4e0eef8d`, `48b05dc7`, `36800b54`, `25b5f024`) has `agentId = a1416bb0…`, which **equals** `issue.assigneeAgentId`. So a naive "another agent is trying to touch my issue" explanation does not fit — the agent that owns the checkout is denied writing to its own issue.

## Root cause (confirmed by reading the code, not speculation)

In `server/src/routes/issues.ts`, `assertAgentIssueMutationAllowed` (starts at `issues.ts:4052`) runs its checks in this order:

```
issues.ts:4098   const boundaryDecision = await decideIssueAccess(req, issue, "issue:mutate");
issues.ts:4099   if (!boundaryDecision.allowed) {
issues.ts:4100     return denyIssueWrite(req, res, issue, issueWriteDenialCodeForDecision(boundaryDecision));
issues.ts:4101   }
issues.ts:4102   if (issue.assigneeAgentId === null) { return true; }
issues.ts:4105   if (issue.assigneeAgentId !== actorAgentId) { ... }   // NOT the source — see below
```

[...]

PR #12689, opened 2026-09-01
Log issue write denial code, issueId, actorAgentId, runId

matt-palmetto:
## Thinking Path

> - This concerns the issues HTTP route layer, specifically the write-authorization path for agent-initiated issue mutations
> - When an agent's `PATCH /api/issues/:id` is refused, `denyIssueWrite` returns a denial code in the response body, but that code is never written to the server log
> - It needs to be addressed because the code is the only signal that says WHICH authorization rule refused the write. Without it in the logs, an operator holding only server-side evidence cannot tell a checkout-lock refusal from a policy refusal, and a recurring 403 cannot be diagnosed at all after the fact
> - This pull request logs the denial code at warn level, together with the issue, actor agent, and run it applies to
> - The benefit is that this class of refusal becomes diagnosable from logs alone, without reproducing the request

Refs #12687

**What happened:** Agent-owned issues were repeatedly left in a non-terminal state after a run
ended. Each run's closing `PATCH /api/issues/:id` returned 403. About ten issues stranded this
way over three days.

**What was expected:** Either the write succeeds, or the server log states which rule refused it.

**Why it could not be diagnosed:** `denyIssueWrite` answers with `res.json()` and does not set
`res.__errorContext`, which is what the pino-http logger reads to record 4xx body detail. Only
`error-handler.ts` and one call site in `plugins.ts` set it. Searching the running server's logs
for any denial code therefore returns nothing, and the failing rule can only be guessed at by
reading the source.

**Impact:** Any 403 from this path is undiagnosable from operations. The related close-out defect
is tracked separately in #12686; this PR does not attempt to fix that defect, only to make it and
its whole class observable.

## What Changed

- Added a `logger.warn` call in `denyIssueWrite` (`server/src/routes/issues.ts`) that records the
  denial code with the issue id, the actor agent id, and the run id.
- Added a test that asserts the log fires with those fields on a denied write.

## Verification

- Targeted server route suite — passes, and the new log line is visible in the test output with
  the expected fields.
- The added test fails if the log call is removed, so it does not pass vacuously.
- No production behavior changes: the response status, the response body, and the control flow
  are all unchanged. Only an additional log record is emitted.

## Risks

Low risk. The change is additive and observability-only. The one thing to weigh is log volume on
a server with many denied writes, and the field set is deliberately small — code plus three ids,
no request body and no user content, so nothing sensitive is written to the log.

## Model Used

Claude Opus 4.5 (`claude-opus-4-5`), extended thinking, with tool use and code execution.

## Checklist

### DIFF
--- a/server/src/__tests__/issue-comment-reopen-routes.test.ts
+++ b/server/src/__tests__/issue-comment-reopen-routes.test.ts
@@ -92,6 +92,23 @@ const mockObserveCrossIssueInfluence = vi.hoisted(() => vi.fn());
 const mockCrossIssueInfluenceLimitError = vi.hoisted(() => vi.fn());
 const mockCrossIssueInfluenceRunContextError = vi.hoisted(() => vi.fn());
 
+// Captures every logger call, so the write-denial test can assert the denial
+// code and actor context reach the logs (denyIssueWrite responds directly
+// rather than throwing, so this is the only place that context is observable).
+const mockLogger = vi.hoisted(() => {
+  const logger: Record<string, unknown> = {};
+  for (const level of ["info", "warn", "error", "debug", "trace", "fatal"]) {
+    logger[level] = vi.fn();
+  }
+  logger.child = vi.fn(() => logger);
+  return logger;
+});
+
+vi.mock("../middleware/logger.js", () => ({
+  logger: mockLogger,
+  httpLogger: (_req: unknown, _res: unknown, next: () => void) => next(),
+}));
+
 vi.mock("@paperclipai/shared/telemetry", () => ({
   trackAgentTaskCompleted: vi.fn(),
   trackErrorHandlerCrash: vi.fn(),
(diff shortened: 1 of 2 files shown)


matt-palmetto:
The one red shard here is unrelated to this change and looks like infrastructure flake.

`ci / General tests (server (2/5))` reports **94 test files passed, 1 failed**. The failure is
`src/__tests__/heartbeat-direct-adapter-native-isolation.test.ts > executes flag-off codex_local
once without creating native records`, which dies on its DB teardown:

```
Error: Failed query:
      TRUNCATE TABLE
        "native_run_finalizations",
```

This PR touches `denyIssueWrite` in `server/src/routes/issues.ts` (one added `logger.warn`, no
control-flow or response change) and adds one test in
`server/src/__tests__/issue-comment-reopen-routes.test.ts`. That test file does not run in shard
2/5 at all — it is not in that shard's log — and shards 1, 3, 4 and 5 are all green. So the
failing test and the changed code have no path between them.

`ci / verify` fails in 2s as the downstream aggregate of that shard.

A re-run of the failed jobs should clear both. I do not have write/triage on this repo, so I
cannot trigger it myself.

Locally, on the same commit: `vitest run issue-comment-reopen-routes.test.ts` → 97/97 pass, and
`tsc --noEmit` in `server/` is clean. The new test was also negative-controlled — removing the
`logger.warn` line makes it fail with `Number of calls: 0` — so it is not a vacuous test.

PR #12715, opened 2026-09-02
test(server): drain heartbeat work before database teardown

lorenzozanee:


nguyenm7 (approved):
Independently root-caused this flake from the failing CI runs before finding this PR, and landed on the same fix — approving with the local verification this PR's environment couldn't produce.

**Root cause confirmation.** In run 33508746273 / job 100244433546 (and its rerun), Postgres reports the exact cycle: the `afterEach` `TRUNCATE ... RESTART IDENTITY CASCADE` waits for `AccessExclusiveLock` on a relation held by a second backend, while that backend waits for `RowExclusiveLock` on a table the TRUNCATE already locked (`40P01`). The second backend is the run's trailing finalize work: `waitForRunToFinish` polls only the status column, and a run reaches terminal status before `executeRun`'s finally block finishes flushing rows/events (documented on `drainActiveRunExecutions` in `heartbeat.ts`). The suite was added in #12188 without the teardown drain that `heartbeat-local-environment.test.ts` and `helpers/drain-heartbeat-runs.ts` (#10464) established for exactly this class; using the shared helper here is the right shape, and hoisting one suite-level `heartbeatService(db)` is what makes the drain observe the suite's own runs.

**Verification (covers the gap noted in the PR body).** On macOS 26.2, Node v26.7.0, at this PR's exact head `224a70cf`:

- `pnpm vitest run src/__tests__/heartbeat-direct-adapter-native-isolation.test.ts` — **10 consecutive runs, 10/10 green**.
- An independently written equivalent patch also passed 10/10 before I found this PR.
- A scratch harness looping the *unpatched* poll→TRUNCATE sequence 60× did not trigger the deadlock locally — the window needs slow, loaded CI shard machines — so CI's two identical `40P01` failures remain the reproduction evidence, and the drain closes the window by construction rather than by timing.

**To get the gates green (no new commit needed):**

- The commitperclip `review` job failed only on `Empty section: ## Model Used` — fill that in (it's your authoring model, so I didn't edit it for you), then rerun the failed workflow run; the gate script fetches the PR body live, so a body edit plus rerun suffices.

**Non-blocking observation:** `heartbeat-runtime-skills.test.ts` guards the same TRUNCATE pattern with only a bare 100 ms sleep — same latent race, worth the same drain treatment in a follow-up.

Issue #12411, opened 2026-08-28
heartbeat-runs cancel: OpenAPI advertises agent auth that doesn't exist (same bug class as #9726/#11946)


jonguttman:
## What happened?

`POST /api/heartbeat-runs/{runId}/cancel` (`server/src/routes/agents.ts`) opens with
`assertBoard(req)` — unconditional board-only, no agent bearer path at all. A valid
agent bearer token plus the current `X-Paperclip-Run-Id` gets exactly:

```json
{"error":"Board access required"}
```

But `/api/openapi.json` advertises `x-paperclip-authorization.actor: "board_or_agent"`
and includes `AgentBearerAuth` security for this operation. Confirmed by reading the
generator: `resolveOperationAuthLevel` (`server/src/routes/openapi.ts`) classifies a
route as board-only only if its `METHOD /api/path` key is in the hand-maintained
`BOARD_ONLY_OPERATIONS` set or matches a `BOARD_ONLY_PREFIXES` prefix; everything else
silently falls through to `authenticated` → `board_or_agent`. `POST
/api/heartbeat-runs/{runId}/cancel` was never added to that set, so the generated spec
just describes the classifier's default, not the route's actual guard.

**This is not a one-off — it is the identical bug shape as #9726 / #11946**, just on a
different route (`clear-error` there, `heartbeat-runs/{id}/cancel` here). #9726 fixed
the clear-error instance by adding one entry to `BOARD_ONLY_OPERATIONS`, but was closed
without merging, and #11946 shows it's still live on master. The recurrence on a
second, unrelated route is evidence the hand-maintained-allowlist approach itself is
the defect, not any single missing entry.

### A broader sweep (NOT independently verified per-route — see caveat)

Grepping every `router.<method>(...)` handler whose **first statement** is
`assertBoard(req)` and diffing the reconstructed `METHOD /api/path` key against
`BOARD_ONLY_OPERATIONS`/`BOARD_ONLY_PREFIXES` turns up roughly 30–40 more candidate
routes across `agents.ts`, `companies.ts`, `costs.ts`, `secrets.ts`,
`issue-tree-control.ts`, `approvals.ts`, `activity.ts`, and `attention.ts` — e.g.
`POST /api/agents/{id}/pause`, `POST /api/agents/{id}/resume`, `POST
/api/agents/{id}/terminate`, `PATCH /api/secrets/{id}`, `POST
[...]

Issue #11946, opened 2026-08-22
clear-error: OpenAPI advertises agent auth that doesn't exist, and there's no non-board way to clear a stale errorReason at all


satruxag:
## What happened?

`POST /api/agents/{id}/clear-error` is the only way to reset an agent's `errorReason`/`status:error` back to a working state, and it is (correctly, by design) board-only at runtime. But two things around it are broken for anyone trying to build or reason about agent-error recovery:

**1. The published OpenAPI contract says otherwise.** The generated spec for this route advertises `AgentBearerAuth` as a valid security scheme and `x-paperclip-authorization: {actor: "board_or_agent"}`. In practice, calling it with an agent API key — including the CEO-tier agent's own key, on its own agent record — returns `403 {"error":"Board access required"}`. Anyone designing an automated recovery flow off the spec (rather than by trial and error against the live server) will build against a permission that doesn't exist.

This exact mismatch was already caught and fixed in #9726 (`fix(server): mark clear-error OpenAPI contract as board-only`) — checks passing, Greptile review at 5/5 confidence, one-line change to `BOARD_ONLY_OPERATIONS` in `server/src/routes/openapi.ts` plus a doc note. It was closed without merging. Reproduced independently against current `master`; the mismatch is still live. Requesting it (or an equivalent) actually land.

**2. Even granting #1, there's no non-board path to clear a stale `errorReason` at all.** `PATCH /api/agents/{id}` does not expose `errorReason` in its schema (checked against the full property list). So the *only* way to clear it, ever, is `clear-error`, which is board-only. Concretely: an agent that has recovered and is healthy again keeps showing its last traceback in `errorReason` indefinitely unless a board principal calls `clear-error` for it. That's a diagnostics problem (the next incident's triage starts from stale data) as much as a permissions one.

This is adjacent to #10956 (allow `PATCH` to revive a `terminated` agent) but distinct: #10956 is about the `terminated` status transition being blocked outright; this is about a **non-terminated, already-healthy** agent whose `errorReason` field has no clearing path short of board action, regardless of status.

## Expected behavior

Pick one (not necessarily mutually exclusive):

- `errorReason` becomes PATCH-able (e.g. `PATCH /api/agents/{id}` with `errorReason: null`), gated by whatever permission already governs `agents:write`/`agents:configure` for the requesting principal, so a non-board agent principal with the right permission can clear its own or a subordinate's stale error state.
[...]

PR #12722, opened 2026-09-02
fix(openapi): track base drift — annotate the queued-comment steer route, drop two removed experimental entries

### LINKED ISSUE #12661: OpenAPI x-paperclip-authorization contradicts assertBoard on 48 operations
# Upstream report B — OpenAPI `x-paperclip-authorization` contradicts the runtime guard on 48 operations

**Type:** bug — published API contract does not match enforced authorization
**Versions checked:** `@paperclipai/server` 2026.817.0 (installed and running here; the served `/api/openapi.json` was read directly) and 2026.824.0 (npm dist-tag `latest`, tarball read on 2026-08-25). Present in both.

---

## Summary

The generated OpenAPI document tells clients that 48 board-only operations are callable by an agent token. They are not — the handlers reject agent actors with 403. The annotation is derived from a hand-maintained path table inside the document generator that has no link to, and no test against, the `assertBoard(req)` guard the handler actually runs.

The published contract does not merely omit the restriction; it positively advertises the opposite, listing `AgentBearerAuth` among the accepted security schemes for each of these operations.

## Reproduction

Read the served document and an authenticated agent's actual result for the same operation:

```
$ curl -s .../api/openapi.json | jq '.paths."/api/companies/{companyId}/activity".post
                                     | {auth: ."x-paperclip-authorization", security}'
{
  "auth":     { "actor": "board_or_agent" },
  "security": [ {"BoardSessionAuth": []}, {"BoardApiKeyAuth": []}, {"AgentBearerAuth": []} ]
}

$ curl -s -o /dev/null -w '%{http_code}
' \
       -H "Authorization: Bearer $AGENT_TOKEN" .../api/companies/$COMPANY/attention
403
```

Six of the listed operations were probed live with a valid agent bearer token on 2026.817.0 — all read-only ones, to avoid side effects. Every one returned 403:

| Operation | Spec actor | Live result with agent token |
|---|---|---|
| `GET /api/companies/{companyId}/attention` | `board_or_agent` | 403 |
| `GET /api/companies/{companyId}/decisions` | `board_or_agent` | 403 |
| `GET /api/companies/{companyId}/secrets` | `board_or_agent` | 403 |
| `GET /api/tool-gateway/runtime-slots` | `board_or_agent` | 403 |
| `GET /api/tool-gateway/audit` | `board_or_agent` | 403 |
| `GET /api/issues/{id}/tree-holds` | `board_or_agent` | 403 |

## Mechanism

`server/dist/routes/openapi.js` computes the annotation in `applyDocumentFixups`:

```js
const authLevel = resolveOperationAuthLevel(method, path);
operation["x-paperclip-authorization"] =
    authLevel === "instance_admin" ? { actor: "board", instanceAdmin: true }
  : authLevel === "board"          ? { actor: "board" }
  : authLevel === "authenticated"  ? { actor: "board_or_agent" }
  :                                  { actor: "public" };
```

and `resolveOperationAuthLevel` classifies from three hand-maintained literals in that same file:

```js
function resolveOperationAuthLevel(method, path) {
    const key = operationKey(method, path);
    if (PUBLIC_OPERATIONS.has(key))         return "public";
    if (INSTANCE_ADMIN_OPERATIONS.has(key)) return "instance_admin";
    if (isBoardOnlyOperation(method, path)) return "board";
    return "authenticated";          // <- default, becomes actor: board_or_agent
}
```

`isBoardOnlyOperation` consults `BOARD_ONLY_OPERATIONS` (a `Set` of 133 `"METHOD /path"` strings in 2026.817.0, 134 in 2026.824.0) plus four `BOARD_ONLY_PREFIXES` (`/api/auth/`, `/api/admin/`, `/api/plugins`, `/api/instance/`).

So authorization has two independent sources of truth: the per-handler `assertBoard(req)` call that actually enforces, and a string list in the doc generator that describes. Nothing keeps them in step. The failure mode is not "someone forgot one entry" — it is that the default for an unlisted route is `board_or_agent`, so **every new board-only route is mislabeled until somebody remembers to add its exact `METHOD /path` string to the set**. The list grew by exactly one entry between 2026.817.0 and 2026.824.0 while the mislabel count stayed at 47–48, which is what that drift looks like in practice.

## Full list (2026.817.0, served document vs. shipped handlers)

Method, path, the actor the served spec publishes, and the source line of the `assertBoard(req)` that contradicts it:

| Method | Path | Spec actor | Handler guard |
|---|---|---|---|
| `DELETE` | `/api/agents/{id}` | `board_or_agent` | `routes/agents.js:2794` |
| `DELETE` | `/api/agents/{id}/keys/{keyId}` | `board_or_agent` | `routes/agents.js:2851` |
| `DELETE` | `/api/secrets/{id}` | `board_or_agent` | `routes/secrets.js:881` |
| `GET` | `/api/agents/{id}/keys` | `board_or_agent` | `routes/agents.js:2815` |
| `GET` | `/api/agents/{id}/runtime-state` | `board_or_agent` | `routes/agents.js:1996` |
| `GET` | `/api/agents/{id}/task-sessions` | `board_or_agent` | `routes/agents.js:2006` |
| `GET` | `/api/companies/{companyId}/attention` | `board_or_agent` | `routes/attention.js:18` |
| `GET` | `/api/companies/{companyId}/costs/quota-windows` | `board_or_agent` | `routes/costs.js:236` |
| `GET` | `/api/companies/{companyId}/decision-training` | `board_or_agent` | `routes/decision-training.js:100` |
| `GET` | `/api/companies/{companyId}/decision-training/export.jsonl` | `board_or_agent` | `routes/decision-training.js:121` |
| `GET` | `/api/companies/{companyId}/decisions` | `board_or_agent` | `routes/decisions.js:144` |
| `GET` | `/api/companies/{companyId}/secret-proposals` | `board_or_agent` | `routes/secrets.js:230` |
| `GET` | `/api/companies/{companyId}/secret-providers` | `board_or_agent` | `routes/secrets.js:316` |
| `GET` | `/api/companies/{companyId}/secrets` | `board_or_agent` | `routes/secrets.js:506` |
| `GET` | `/api/decision-training/{id}` | `board_or_agent` | `routes/decision-training.js:146` |
| `GET` | `/api/issues/{id}/tree-control/state` | `board_or_agent` | `routes/issue-tree-control.js:281` |
| `GET` | `/api/issues/{id}/tree-holds` | `board_or_agent` | `routes/issue-tree-control.js:290` |
| `GET` | `/api/issues/{id}/tree-holds/{holdId}` | `board_or_agent` | `routes/issue-tree-control.js:307` |
| `GET` | `/api/tool-connections/{connectionId}/installs` | `board_or_agent` | `routes/tool-access.js:553` |
| `GET` | `/api/tool-gateway/audit` | `board_or_agent` | `routes/tool-gateway.js:589` |
| `GET` | `/api/tool-gateway/runtime-slots` | `board_or_agent` | `routes/tool-gateway.js:522` |
| `PATCH` | `/api/agents/{agentId}/budgets` | `board_or_agent` | `routes/costs.js:310` |
| `PATCH` | `/api/companies/{companyId}/budgets` | `board_or_agent` | `routes/costs.js:280` |
| `PATCH` | `/api/secrets/{id}` | `board_or_agent` | `routes/secrets.js:826` |
| `POST` | `/api/agents/{id}/approve` | `board_or_agent` | `routes/agents.js:2683` |
| `POST` | `/api/agents/{id}/claude-login` | `board_or_agent` | `routes/agents.js:3008` |
| `POST` | `/api/agents/{id}/clear-error` | `board_or_agent` | `routes/agents.js:2655` |
| `POST` | `/api/agents/{id}/keys` | `board_or_agent` | `routes/agents.js:2825` |
| `POST` | `/api/agents/{id}/pause` | `board_or_agent` | `routes/agents.js:2605` |
| `POST` | `/api/agents/{id}/resume` | `board_or_agent` | `routes/agents.js:2627` |
| `POST` | `/api/agents/{id}/runtime-state/reset-session` | `board_or_agent` | `routes/agents.js:2019` |
| `POST` | `/api/agents/{id}/terminate` | `board_or_agent` | `routes/agents.js:2732` |
| `POST` | `/api/approvals/{id}/approve` | `board_or_agent` | `routes/approvals.js:244` |
| `POST` | `/api/approvals/{id}/reject` | `board_or_agent` | `routes/approvals.js:351` |
| `POST` | `/api/approvals/{id}/request-revision` | `board_or_agent` | `routes/approvals.js:383` |
| `POST` | `/api/companies/{companyId}/activity` | `board_or_agent` | `routes/activity.js:283` |
| `POST` | `/api/companies/{companyId}/budget-incidents/{incidentId}/resolve` | `board_or_agent` | `routes/costs.js:263` |
| `POST` | `/api/companies/{companyId}/budgets/policies` | `board_or_agent` | `routes/costs.js:256` |
| `POST` | `/api/companies/{companyId}/finance-events` | `board_or_agent` | `routes/costs.js:109` |
| `POST` | `/api/companies/{companyId}/secret-proposals/{id}/reject` | `board_or_agent` | `routes/secrets.js:260` |
| `POST` | `/api/heartbeat-runs/{runId}/cancel` | `board_or_agent` | `routes/agents.js:3121` |
| `POST` | `/api/issues/{id}/scheduled-retry/retry-now` | `board_or_agent` | `routes/issues.js:6712` |
| `POST` | `/api/issues/{id}/stalled-review-decision` | `board_or_agent` | `routes/issues.js:6747` |
| `POST` | `/api/issues/{id}/tree-control/preview` | `board_or_agent` | `routes/issue-tree-control.js:36` |
| `POST` | `/api/issues/{id}/tree-holds` | `board_or_agent` | `routes/issue-tree-control.js:61` |
| `POST` | `/api/issues/{id}/tree-holds/{holdId}/release` | `board_or_agent` | `routes/issue-tree-control.js:324` |
| `POST` | `/api/secrets/{id}/rotate` | `board_or_agent` | `routes/secrets.js:798` |
| `PUT` | `/api/tool-connections/{connectionId}/installs` | `board_or_agent` | `routes/tool-access.js:561` |

Method: every `router.<verb>("...")` in `server/dist/routes/*.js` whose handler calls `assertBoard(req)` within six lines of the handler opening (all 48 above are unconditional, at the head of the handler) was matched by exact path against the served `/api/openapi.json`, and its published `x-paperclip-authorization.actor` compared. All 48 publish `actor: "board_or_agent"` and list `AgentBearerAuth` in `security`. Handlers whose express path did not resolve to an exact spec path were excluded rather than guessed, so 48 is a floor, not a ceiling.

One caveat on provenance: `routes/issues.js` in this installation carries an unrelated local modification, so the two line numbers for that file are taken from the pristine 2026.817.0 artifact, not the modified copy. Every other line number is from the unmodified shipped `dist`.

Re-running the same classification statically against the 2026.824.0 tarball (reimplementing `resolveOperationAuthLevel` from that version's own literals, since the server was not started) yields 47 mislabeled operations — the same defect, essentially the same set.

## Impact

- **Agents burn heartbeats on calls that cannot succeed.** An agent that reads the published contract to decide what it may do will attempt board-only operations and take 403s. `POST /api/companies/{companyId}/activity` is the case that cost us time: it has exactly the right shape for run-scoped annotation (a `runId` FK plus a `details` jsonb) and the spec says an agent may call it.
- **The list is security-relevant.** It includes `POST /api/secrets/{id}/rotate`, `PATCH /api/secrets/{id}`, `DELETE /api/secrets/{id}`, `GET /api/companies/{companyId}/secrets`, `POST /api/agents/{id}/keys`, `POST /api/agents/{id}/terminate`, and `POST /api/approvals/{id}/approve`. The enforcement is correct — these do reject agents — but a document that advertises secret rotation and approval endpoints as agent-callable is a bad artifact to publish, and a poor thing for anyone to reason about authorization from.
- **It silently misinforms generated clients** and anything else that consumes the document as the contract.

## Suggested fix

Remove the second source of truth rather than extend it:

1. Derive the annotation from the guard. Register the auth level where the route is declared (a small wrapper, or route metadata read by both `assertBoard` and the doc generator), so the document is generated from what is enforced.
2. Failing that, add a test that fails the build on drift: enumerate handlers calling `assertBoard`, enumerate operations the generator classifies as `board`, and assert the sets match. That test would fail today on 48 operations, which is the point.

Option 2 alone would have caught this and is cheap; option 1 prevents recurrence.

### LINKED ISSUE #12665: OpenAPI x-paperclip-authorization contradicts assertBoard on 48 operations
# Upstream report B — OpenAPI `x-paperclip-authorization` contradicts the runtime guard on 48 operations

**Type:** bug — published API contract does not match enforced authorization
**Versions checked:** `@paperclipai/server` 2026.817.0 (installed and running here; the served `/api/openapi.json` was read directly) and 2026.824.0 (npm dist-tag `latest`, tarball read on 2026-08-25). Present in both.

---

## Summary

The generated OpenAPI document tells clients that 48 board-only operations are callable by an agent token. They are not — the handlers reject agent actors with 403. The annotation is derived from a hand-maintained path table inside the document generator that has no link to, and no test against, the `assertBoard(req)` guard the handler actually runs.

The published contract does not merely omit the restriction; it positively advertises the opposite, listing `AgentBearerAuth` among the accepted security schemes for each of these operations.

## Reproduction

Read the served document and an authenticated agent's actual result for the same operation:

```
$ curl -s .../api/openapi.json | jq '.paths."/api/companies/{companyId}/activity".post
                                     | {auth: ."x-paperclip-authorization", security}'
{
  "auth":     { "actor": "board_or_agent" },
  "security": [ {"BoardSessionAuth": []}, {"BoardApiKeyAuth": []}, {"AgentBearerAuth": []} ]
}

$ curl -s -o /dev/null -w '%{http_code}
' \
       -H "Authorization: Bearer $AGENT_TOKEN" .../api/companies/$COMPANY/attention
403
```

Six of the listed operations were probed live with a valid agent bearer token on 2026.817.0 — all read-only ones, to avoid side effects. Every one returned 403:

| Operation | Spec actor | Live result with agent token |
|---|---|---|
| `GET /api/companies/{companyId}/attention` | `board_or_agent` | 403 |
| `GET /api/companies/{companyId}/decisions` | `board_or_agent` | 403 |
| `GET /api/companies/{companyId}/secrets` | `board_or_agent` | 403 |
| `GET /api/tool-gateway/runtime-slots` | `board_or_agent` | 403 |
| `GET /api/tool-gateway/audit` | `board_or_agent` | 403 |
| `GET /api/issues/{id}/tree-holds` | `board_or_agent` | 403 |

## Mechanism

`server/dist/routes/openapi.js` computes the annotation in `applyDocumentFixups`:

```js
const authLevel = resolveOperationAuthLevel(method, path);
operation["x-paperclip-authorization"] =
    authLevel === "instance_admin" ? { actor: "board", instanceAdmin: true }
  : authLevel === "board"          ? { actor: "board" }
  : authLevel === "authenticated"  ? { actor: "board_or_agent" }
  :                                  { actor: "public" };
```

and `resolveOperationAuthLevel` classifies from three hand-maintained literals in that same file:

```js
function resolveOperationAuthLevel(method, path) {
    const key = operationKey(method, path);
    if (PUBLIC_OPERATIONS.has(key))         return "public";
    if (INSTANCE_ADMIN_OPERATIONS.has(key)) return "instance_admin";
    if (isBoardOnlyOperation(method, path)) return "board";
    return "authenticated";          // <- default, becomes actor: board_or_agent
}
```

`isBoardOnlyOperation` consults `BOARD_ONLY_OPERATIONS` (a `Set` of 133 `"METHOD /path"` strings in 2026.817.0, 134 in 2026.824.0) plus four `BOARD_ONLY_PREFIXES` (`/api/auth/`, `/api/admin/`, `/api/plugins`, `/api/instance/`).

So authorization has two independent sources of truth: the per-handler `assertBoard(req)` call that actually enforces, and a string list in the doc generator that describes. Nothing keeps them in step. The failure mode is not "someone forgot one entry" — it is that the default for an unlisted route is `board_or_agent`, so **every new board-only route is mislabeled until somebody remembers to add its exact `METHOD /path` string to the set**. The list grew by exactly one entry between 2026.817.0 and 2026.824.0 while the mislabel count stayed at 47–48, which is what that drift looks like in practice.

## Full list (2026.817.0, served document vs. shipped handlers)

Method, path, the actor the served spec publishes, and the source line of the `assertBoard(req)` that contradicts it:

| Method | Path | Spec actor | Handler guard |
|---|---|---|---|
| `DELETE` | `/api/agents/{id}` | `board_or_agent` | `routes/agents.js:2794` |
| `DELETE` | `/api/agents/{id}/keys/{keyId}` | `board_or_agent` | `routes/agents.js:2851` |
| `DELETE` | `/api/secrets/{id}` | `board_or_agent` | `routes/secrets.js:881` |
| `GET` | `/api/agents/{id}/keys` | `board_or_agent` | `routes/agents.js:2815` |
| `GET` | `/api/agents/{id}/runtime-state` | `board_or_agent` | `routes/agents.js:1996` |
| `GET` | `/api/agents/{id}/task-sessions` | `board_or_agent` | `routes/agents.js:2006` |
| `GET` | `/api/companies/{companyId}/attention` | `board_or_agent` | `routes/attention.js:18` |
| `GET` | `/api/companies/{companyId}/costs/quota-windows` | `board_or_agent` | `routes/costs.js:236` |
| `GET` | `/api/companies/{companyId}/decision-training` | `board_or_agent` | `routes/decision-training.js:100` |
| `GET` | `/api/companies/{companyId}/decision-training/export.jsonl` | `board_or_agent` | `routes/decision-training.js:121` |
| `GET` | `/api/companies/{companyId}/decisions` | `board_or_agent` | `routes/decisions.js:144` |
| `GET` | `/api/companies/{companyId}/secret-proposals` | `board_or_agent` | `routes/secrets.js:230` |
| `GET` | `/api/companies/{companyId}/secret-providers` | `board_or_agent` | `routes/secrets.js:316` |
| `GET` | `/api/companies/{companyId}/secrets` | `board_or_agent` | `routes/secrets.js:506` |
| `GET` | `/api/decision-training/{id}` | `board_or_agent` | `routes/decision-training.js:146` |
| `GET` | `/api/issues/{id}/tree-control/state` | `board_or_agent` | `routes/issue-tree-control.js:281` |
| `GET` | `/api/issues/{id}/tree-holds` | `board_or_agent` | `routes/issue-tree-control.js:290` |
| `GET` | `/api/issues/{id}/tree-holds/{holdId}` | `board_or_agent` | `routes/issue-tree-control.js:307` |
| `GET` | `/api/tool-connections/{connectionId}/installs` | `board_or_agent` | `routes/tool-access.js:553` |
| `GET` | `/api/tool-gateway/audit` | `board_or_agent` | `routes/tool-gateway.js:589` |
| `GET` | `/api/tool-gateway/runtime-slots` | `board_or_agent` | `routes/tool-gateway.js:522` |
| `PATCH` | `/api/agents/{agentId}/budgets` | `board_or_agent` | `routes/costs.js:310` |
| `PATCH` | `/api/companies/{companyId}/budgets` | `board_or_agent` | `routes/costs.js:280` |
| `PATCH` | `/api/secrets/{id}` | `board_or_agent` | `routes/secrets.js:826` |
| `POST` | `/api/agents/{id}/approve` | `board_or_agent` | `routes/agents.js:2683` |
| `POST` | `/api/agents/{id}/claude-login` | `board_or_agent` | `routes/agents.js:3008` |
| `POST` | `/api/agents/{id}/clear-error` | `board_or_agent` | `routes/agents.js:2655` |
| `POST` | `/api/agents/{id}/keys` | `board_or_agent` | `routes/agents.js:2825` |
| `POST` | `/api/agents/{id}/pause` | `board_or_agent` | `routes/agents.js:2605` |
| `POST` | `/api/agents/{id}/resume` | `board_or_agent` | `routes/agents.js:2627` |
| `POST` | `/api/agents/{id}/runtime-state/reset-session` | `board_or_agent` | `routes/agents.js:2019` |
| `POST` | `/api/agents/{id}/terminate` | `board_or_agent` | `routes/agents.js:2732` |
| `POST` | `/api/approvals/{id}/approve` | `board_or_agent` | `routes/approvals.js:244` |
| `POST` | `/api/approvals/{id}/reject` | `board_or_agent` | `routes/approvals.js:351` |
| `POST` | `/api/approvals/{id}/request-revision` | `board_or_agent` | `routes/approvals.js:383` |
| `POST` | `/api/companies/{companyId}/activity` | `board_or_agent` | `routes/activity.js:283` |
| `POST` | `/api/companies/{companyId}/budget-incidents/{incidentId}/resolve` | `board_or_agent` | `routes/costs.js:263` |
| `POST` | `/api/companies/{companyId}/budgets/policies` | `board_or_agent` | `routes/costs.js:256` |
| `POST` | `/api/companies/{companyId}/finance-events` | `board_or_agent` | `routes/costs.js:109` |
| `POST` | `/api/companies/{companyId}/secret-proposals/{id}/reject` | `board_or_agent` | `routes/secrets.js:260` |
| `POST` | `/api/heartbeat-runs/{runId}/cancel` | `board_or_agent` | `routes/agents.js:3121` |
| `POST` | `/api/issues/{id}/scheduled-retry/retry-now` | `board_or_agent` | `routes/issues.js:6712` |
| `POST` | `/api/issues/{id}/stalled-review-decision` | `board_or_agent` | `routes/issues.js:6747` |
| `POST` | `/api/issues/{id}/tree-control/preview` | `board_or_agent` | `routes/issue-tree-control.js:36` |
| `POST` | `/api/issues/{id}/tree-holds` | `board_or_agent` | `routes/issue-tree-control.js:61` |
| `POST` | `/api/issues/{id}/tree-holds/{holdId}/release` | `board_or_agent` | `routes/issue-tree-control.js:324` |
| `POST` | `/api/secrets/{id}/rotate` | `board_or_agent` | `routes/secrets.js:798` |
| `PUT` | `/api/tool-connections/{connectionId}/installs` | `board_or_agent` | `routes/tool-access.js:561` |

Method: every `router.<verb>("...")` in `server/dist/routes/*.js` whose handler calls `assertBoard(req)` within six lines of the handler opening (all 48 above are unconditional, at the head of the handler) was matched by exact path against the served `/api/openapi.json`, and its published `x-paperclip-authorization.actor` compared. All 48 publish `actor: "board_or_agent"` and list `AgentBearerAuth` in `security`. Handlers whose express path did not resolve to an exact spec path were excluded rather than guessed, so 48 is a floor, not a ceiling.

One caveat on provenance: `routes/issues.js` in this installation carries an unrelated local modification, so the two line numbers for that file are taken from the pristine 2026.817.0 artifact, not the modified copy. Every other line number is from the unmodified shipped `dist`.

Re-running the same classification statically against the 2026.824.0 tarball (reimplementing `resolveOperationAuthLevel` from that version's own literals, since the server was not started) yields 47 mislabeled operations — the same defect, essentially the same set.

## Impact

- **Agents burn heartbeats on calls that cannot succeed.** An agent that reads the published contract to decide what it may do will attempt board-only operations and take 403s. `POST /api/companies/{companyId}/activity` is the case that cost us time: it has exactly the right shape for run-scoped annotation (a `runId` FK plus a `details` jsonb) and the spec says an agent may call it.
- **The list is security-relevant.** It includes `POST /api/secrets/{id}/rotate`, `PATCH /api/secrets/{id}`, `DELETE /api/secrets/{id}`, `GET /api/companies/{companyId}/secrets`, `POST /api/agents/{id}/keys`, `POST /api/agents/{id}/terminate`, and `POST /api/approvals/{id}/approve`. The enforcement is correct — these do reject agents — but a document that advertises secret rotation and approval endpoints as agent-callable is a bad artifact to publish, and a poor thing for anyone to reason about authorization from.
- **It silently misinforms generated clients** and anything else that consumes the document as the contract.

## Suggested fix

Remove the second source of truth rather than extend it:

1. Derive the annotation from the guard. Register the auth level where the route is declared (a small wrapper, or route metadata read by both `assertBoard` and the doc generator), so the document is generated from what is enforced.
2. Failing that, add a test that fails the build on drift: enumerate handlers calling `assertBoard`, enumerate operations the generator classifies as `board`, and assert the sets match. That test would fail today on 48 operations, which is the point.

Option 2 alone would have caught this and is cheap; option 1 prevents recurrence.

zannis:
## Thinking Path

> - The server publishes an OpenAPI document. Its `x-paperclip-authorization` extension tells integrators which actor type may call each operation.
> - The annotation comes from hand-maintained tables in `server/src/routes/openapi.ts` (`BOARD_ONLY_OPERATIONS` / `BOARD_ONLY_PREFIXES`), not from the `assertBoard(req)` guard the handler actually runs.
> - Any board-only handler missing from the tables falls through to `authenticated`. The document then publishes `{actor: "board_or_agent"}` and lists `AgentBearerAuth` among the accepted schemes. The contract advertises the opposite of what the server enforces. #12661 reported 48 such operations; a live-server probe confirms the class (each probeable one returns `403 Board access required` to a valid agent token).
> - This pull request corrects the tables for every affected operation and adds a scan-based regression test, so a new board-only route cannot ship unannotated.
> - The benefit is a served API contract that matches enforcement for every operation, and a test suite that keeps it that way.

- Fixes: #12661
- Closes #12665 (byte-identical duplicate of #12661)
- Refs #12411
- Refs #11946
- Refs #9726 (merged single-route precedent — this PR fixes the class and closes the hole that reopens it)

## What Changed

- **124 operations that fell through to `authenticated` now publish their real restriction.** 95 entries added to `BOARD_ONLY_OPERATIONS` and 29 to `INSTANCE_ADMIN_OPERATIONS`. Every entry was confirmed individually against its handler: the assertion runs on every path that can reach a success response. Entries follow the file's existing grouping.
- **14 prefix-covered operations reclassified to instance-admin.** `DELETE /api/plugins/{pluginId}`, the five plugin admin operations (`enable`/`disable`/`upgrade`/`config`/`jobs/{jobId}/trigger`), `GET /api/admin/users`, `GET /api/admin/users/{userId}/company-access`, `GET /api/instance/scheduler-heartbeats`, and the five `instance-settings.ts` write operations each open with an unconditional instance-admin assertion, but were covered only by a board prefix — the document told an ordinary board token it could call them.
- **A regression guard in `openapi-routes.test.ts`.** The route scan that already backs "spec paths match mounted routes exactly" now also reports which handlers run a top-level board assertion, resolved transitively from `assertBoard`. A new test requires the spec to annotate each of those `{actor: "board"}`. A second closure tracks the stricter instance-admin tier, and a companion test requires each of its routes to carry `instanceAdmin: true` — so a demotion from instance-admin to plain board fails the suite even though the actor stays `board`.
- **Table hygiene tests.** Every entry in all six hand-keyed operation tables must still match a served operation (stale entries silently downgrade the published authorization — and a stale `PUBLIC_OPERATIONS` entry whose path is reused would ship the new route with `security: []`). No operation may sit in two auth tables — precedence silently shadows the loser, which is how a dead `POST /api/companies` board entry (removed here) survived. The scan's own detection rules are pinned by a separate test against synthetic sources.
- No change to how auth levels are declared, and no change to any handler's behaviour — only to what the generated document says about it.

<details>
<summary><b>How the scan decides, and the reconciliation against #12661's count of 48</b></summary>

### How the scan decides

It reads block structure, so it first masks string, template, comment and regex-literal contents (preserving offsets), and fails loudly on any route file it can no longer brace-balance rather than silently under-reporting.

It resolves the assertion set transitively rather than matching a list of names. `assertBoard(req)` is the single primitive that rejects a non-board actor, but almost nothing calls it directly: `authz.ts` wraps it in `assertBoardOrgAccess` and `assertInstanceAdmin`, and individual route files wrap those again in local helpers — `assertCompanySecretWrite` in `secrets.ts`, `requireBoardUserId` in `sidebar-preferences.ts`, `assertBoardPermission` in `tool-gateway.ts`, `assertCanManageAdapterLogin` in `agents.ts`, and `activeToolMembership` → `assertToolConnectionConfigureAccess` in `tool-access.ts`, which is two hops out. The scan computes the closure: any function whose body unconditionally calls a known assertion is itself an assertion, iterated to a fixpoint, seeded from `assertBoard` alone and extended per route file with that file's own helpers. Enumerating names by hand is what let this drift in — three separate passes over the table each missed a different wrapper.

Two guard shapes cannot reach `assertBoard` textually and are handled explicitly. A guard that inlines `req.actor.type !== "board"` instead of calling `assertBoard` — `assertCanAccessInstanceEnvironments` in `environments.ts`, which gates all 16 environment write/admin routes, and `assertCanManageInstanceSettings` in `instance-settings.ts` — is seeded into that file's assertion set by name, verified by reading, and keyed by file so a same-named helper elsewhere is not trusted by association. The other is callback-shaped: `POST /api/companies/{companyId}/adapters/{type}/login-sessions` passes `deriveOwner: () => assertCanManageAdapterLogin(req, companyId)` into a shared spine that awaits it before any side effect. No text scan can honestly claim to see that, so it stays a hand-maintained entry with a pinned spec expectation that keeps it from being silently dropped.

A call counts only when `req` (or `_req`) is its first argument, which lets a wrapper take further arguments without letting `svc.assertBoard(other)` through. Prettier-wrapped calls and trailing commas still match. `return assertBoard(req);` counts when the `return` is itself a statement, and stays conditional when it hangs off a braceless `if`. An inline middleware argument before the real handler does not end the scan: every function argument runs unconditionally, so each top-level arrow body in the registration span is checked in turn — an arrow nested inside another argument's object literal is conditional machinery and never counts. Return-type annotations are balanced over even when the type itself carries braces. `try { ... }` is transparent; `if`/`else` and every other block are not — `POST /api/issues/{id}/interactions`, whose `assertBoard` sits in an `else` branch, is correctly excluded. The scan is bounded to each registration's own argument list, so a neighbouring helper's assertion is never attributed to a route that does not run it. Routes registered with a shared path constant are resolved too.

One limit is deliberate: an early `return` before the assertion does not disqualify it. Several handlers open with `const x = await getAccessibleResource(...); if (!x) return;` and assert straight after — that branch has already written a 404, so the handler reaches no success response without a board check. This is the one direction in which the scan could over-report, which is why every entry it reports was confirmed against its handler before being added. Detection is otherwise one-directional: a handler that gates board access some other way is simply not reported, so the guard never demands an annotation on a guess.

Two conditional guards are excluded, and the test asserts they stay excluded: `POST /api/invites/{inviteId}/revoke` asserts instance admin only for bootstrap-CEO invites, and `POST /api/agents/{id}/heartbeat/invoke` explicitly admits an agent actor and asserts only when a provider trace is requested. Both are genuinely `board_or_agent`.

Masking rather than parsing: TypeScript 7 no longer publishes the syntactic compiler API from its main entry, and the official lexer survives only under an explicitly-unstable subpath a test should not couple to. Vite's `parseAst` — already a declared devDependency — does parse annotated TypeScript with exact node offsets; rebuilding the scan on it is the right follow-up if these helpers ever need to grow again. For this change the hand scan stays: its rules are individually pinned, and the brace-balance check fails the suite if the masker breaks instead of quietly shrinking what gets checked.

### Reconciling 95 board-only entries against the 48 in #12661

#12661 counted 48 on `@paperclipai/server` 2026.817.0 / 2026.824.0, reporting specifically on direct `assertBoard` calls. This branch counts 95 board-only entries on current master, plus 29 instance-admin entries #12661's method could not see at all (13 via `assertInstanceAdmin`, 16 via the inline environments guard).

| | Count | |
|---|---:|---|
| In #12661 and still board-only on master | 47 | included |
| In #12661, no longer board-only on master | 1 | correctly excluded |
| New routes landed after #12661's snapshot | 10 | added |
| Predate #12661's snapshot, missed by its enumeration | 14 | added |
| Guarded by a wrapper that reaches `assertBoard` indirectly (incl. `_req` naming) | 23 | added |
| Callback-shaped guard, hand-maintained entry | 1 | added |
| **Total** | **95** | |

The one drop: `POST /api/agents/{id}/resume` is listed in #12661, but on master its guard is `await assertCanResumeAgent(req, existing)`, which runs an `agent_config:update` access decision that can allow an agent actor. `board_or_agent` is now the correct annotation for it, so it is intentionally not added.

The 14 that predate the snapshot: 3 `GET /api/adapters/{type}*` routes guard with `assertBoardOrgAccess` rather than `assertBoard`; 10 routes live in `companies.ts`, whose router mounts at `/api/companies` rather than `/api`, so their declared paths only become API paths after the mount prefix (two also register with a path constant rather than a string literal); and `POST /api/issues/{id}/interactions/{interactionId}/cancel` reaches its `assertBoard` after an explicit agent-actor rejection, so the assertion is not the handler's first statement.

The guard already earned its keep inside this branch's own lifetime: syncing with master after the branch was reviewed, it flagged `POST /api/issues/{id}/queued-comments/{commentId}/steer` (a new board-only route that had landed unannotated — confirmed against its handler, which opens with a top-level `assertBoard(req)`) and two removed experimental instance-settings routes whose table entries had gone stale. Both failures are exactly the drift classes this PR exists to catch.

</details>

## Verification

- Run `pnpm vitest run src/__tests__/openapi-routes.test.ts` in `server/`. All 9 tests pass on this branch.
- Red on master: the new board-annotation test fails on master listing exactly the scan-visible unannotated operations, and passes with the table correction. The hand-maintained callback-shaped entry is covered by its own pinned expectation, which also fails on master.
- Red on master, second tier: the instance-admin test fails before the reclassification, listing exactly the prefix-covered admin operations; and moving an entry from `INSTANCE_ADMIN_OPERATIONS` into `BOARD_ONLY_OPERATIONS` fails it naming exactly that operation — a drift the rest of the suite passes, since the actor stays `board`.
- Each scan rule was mutation-tested: reverting it individually turns the corresponding assertion red, so none of the coverage is vacuous.
- `pnpm --filter @paperclipai/server typecheck` exits 0.
- Live-server probe (pre-fix): every GET-probeable operation in the affected set returns `403 {"error":"Board access required"}` to a valid agent token while the spec said `board_or_agent`.

## Risks

- Low risk. No handler behaviour changes; only the generated OpenAPI document changes. A client that trusted the old annotation already received 403 at runtime — the document now says so up front.
- The corrected annotations narrow the advertised contract. A downstream tool that generated an agent-token client for one of these operations will now see the restriction at generation time instead of failing at call time. That is the intended effect.
- The new tests make the tables self-policing: a stale entry, a shadowed entry, or a new unannotated board-only route fails the suite rather than shipping.

## Model Used

- Claude (Anthropic) — Claude Fable 5, model id `claude-fable-5`, extended thinking enabled, agentic tool use via Claude Code.

## Checklist

- [x] I have updated relevant documentation to reflect my changes (the served OpenAPI document is the documentation this PR corrects)

Depends-on: none — standalone spec-table correction, no companion PRs required


zannis:
Closing as a duplicate of #12720 — my mistake.

These two PRs are the same commit, `07e74cb7cc1698e83c8a25042a338958b55fc1dc`, published twice about twenty minutes apart from two differently-named branches on my fork. Identical head SHAs, identical diffs.

No work is being retracted here: the diff stays open as #12720. Apologies for the duplicate in your queue.

PR #8905, opened 2026-07-02
Clear stale agent errorReason on recovered run start

saphid:
## Thinking Path

> - Paperclip manages AI agent companies, including agent status and runtime health in the API/UI.
> - Agent status and error display are projected through the server agent service before route handlers return list/get responses.
> - A recovered run can make an agent healthy again while an older database row still has `errorReason` populated from a previous failure.
> - The heartbeat path should clear that stale field when it starts a recovered run, but API projections also need to defend against existing stale rows.
> - This pull request clears stale error state on recovered-run start and hides stale `errorReason` values for any non-error agent status.
> - This replaces #8725 on a public-safe branch using the same reviewed commit.

## Linked Issues or Issue Description

No public issue exists.

Bug report:
- Summary: Recovered or otherwise healthy agents could still expose a stale `errorReason` through agent list/get responses.
- Expected: Agent API responses only expose `errorReason` when the agent status is `error`.
- Actual: Stale database rows with `status` like `running` or `idle` could still return an old error reason.
- Duplicate search: searched GitHub PRs for stale agent errorReason; #8725 was the only match and is being replaced for metadata cleanup.

## What Changed

- Clear stale `errorReason` in the recovered-run heartbeat transition to `running`.
- Add a defensive agent projection guard so non-error statuses return `errorReason: null` from list/get service responses.
- Add embedded Postgres regression coverage for stale `running` and `idle` rows, plus a real `error` row that still exposes its reason.

## Verification

- `pnpm exec vitest run server/src/__tests__/heartbeat-process-recovery.test.ts --testNamePattern "clears stale agent errorReason when a recovered run starts"`
- `pnpm vitest run server/src/__tests__/agents-service-clear-error.test.ts`
- `git diff --check`

## Risks

Low risk. This changes API projection only for non-error agents; persisted diagnostics are not deleted by the projection guard, and real error agents still expose their `errorReason`.

> For core feature work, check [`ROADMAP.md`](ROADMAP.md) first and discuss it in `#dev` before opening the PR. Feature PRs that overlap with planned core work may need to be redirected — check the roadmap first. See `CONTRIBUTING.md`.

## Model Used

- OpenAI GPT-5 Codex via local Codex ChatGPT/OAuth session, with shell/tool execution in the repository workspace.

## Checklist

- [x] I have included a thinking path that traces from project context to this change
- [x] I have specified the model used (with version and capability details)
- [x] I have checked ROADMAP.md and confirmed this PR does not duplicate planned core work
[...]

PR #8725, opened 2026-06-29
Clear stale agent errorReason on recovered run start

saphid:
Closing this PR for cleanup.

This contribution has been replaced by #8905 from a public-safe branch with the same reviewed commit. No maintainer action is requested here.

PR #12724, opened 2026-09-02
fix(agents): clear stale error reason on status recovery

bruchris:
## Thinking Path

> - Each agent has a `status` field, and an `errorReason` field that records the last failure
> - The heartbeat process already clears `errorReason` when it finalizes a status change away from `error`
> - But the generic agent update path — the code any `PATCH` to an agent goes through — did not enforce that same rule
> - So a caller could move an agent's status away from `error`, or simply re-affirm an already-clean status, and a stale `errorReason` from an old incident would stay on the row
> - This pull request adds the same invariant to the generic update path: any explicit status change away from `error` also clears `errorReason`
> - The benefit is a healthy agent never reports a leftover error message from a past incident, no matter which code path changed its status

Refs #11946 — related, but distinct. That issue is about the board-only `clear-error` endpoint: its OpenAPI contract wrongly advertises agent-level auth, and there is no way to `PATCH` `errorReason` directly at all. This PR does not change that permission surface or add a new PATCH-able field. It fixes a narrower gap that issue touches on in passing (its point 2): many status transitions already go through the generic agent update function without ever calling `clear-error`, and until now none of them cleared a stale `errorReason` as a side effect.

Dedup search also found #8905 (open) and its closed predecessor #8725, "Clear stale agent errorReason on recovered run start". That PR fixes a different code path: it clears `errorReason` specifically in the heartbeat recovered-run transition, and adds a read-time projection guard that hides (without persisting a clear for) `errorReason` on non-error agents in list/get responses. This PR instead changes the write path itself (the generic agent update function), so the fix applies to every caller that changes `status` through it, not only the recovered-run heartbeat transition and not only at read time. The two changes are complementary, not duplicates, but a reviewer may want to reconcile them before both merge.

No public issue describes this specific write-path gap yet, so here it is inline, following the bug report template:

**What happened?**
The generic agent update function in `server/src/services/agents.ts` persists `errorReason` independently of `status`. A caller can `PATCH` an agent to move its status away from `error` — or simply re-affirm its current non-error status — while an old `errorReason` from a prior incident stays on the row.

**Expected behavior**
Any generic update that explicitly sets `status` to a value other than `error` should also clear `errorReason` to `null`, matching the invariant `heartbeat.ts` already applies on finalization.

**Steps to reproduce**
1. Put an agent into `status: error` with an `errorReason` set (for example, after a failed run).
2. Call the generic agent update path with a patch that sets `status` to any non-`error` value, such as `idle`. This can be the agent's current status or a different one.
3. Read the agent record back. Before this fix, `errorReason` still shows the old message even though `status` is no longer `error`.

Reproduced against `master` at commit `5716fe907e596ce73501408fc6efdb19fb61edf2`, the base this PR is opened against.

**Deployment mode**
Not deployment-specific. The added test suite reproduces and verifies the fix against an embedded PGlite/Postgres instance.

## What Changed

- Added a lifecycle invariant to the generic agent update function in `server/src/services/agents.ts`: any patch that explicitly sets `status` to a value other than `error` now also sets `errorReason` to `null`.
- Scoped the invariant to patches that touch `status`. A patch that leaves `status` untouched never mutates `errorReason`, so this does not silently scrub diagnostics off unrelated field updates.
- Placed the check after existing `adapterConfig`/`permissions` normalization and before config-revision diffing, so it does not create a spurious revision-history entry.
- Added `server/src/__tests__/agents-service-generic-status-update-error-reason.test.ts` with 7 embedded-Postgres regression cases: clearing on recovery, clearing when a caller re-affirms an already-clean status, leaving `errorReason` untouched when `status` is absent from the patch, preserving an explicit `errorReason` when `status` is set to `error`, overriding a caller-supplied `errorReason` when the same patch moves status away from `error`, and confirming the existing `terminated`/`pending_approval` transition guards still run first and are unchanged.

## Verification

- `pnpm exec vitest run server/src/__tests__/agents-service-generic-status-update-error-reason.test.ts` — the 7 focused cases above, 7/7 passing, run against an embedded Postgres instance.
- The broader related agent-service test set (11 tests total, including pre-existing coverage) and `tsc --noEmit` on the `server` package were run before opening this PR; both passed.
- Re-verified the focused suite again at this PR's head commit, after rebasing the change onto two unrelated invariant checks that landed upstream in the meantime (an operational-skill invariant and an OAuth binding check, both unrelated to `status`/`errorReason`) — still 7/7.
- No full monorepo typecheck/build was run for this change; the change is scoped to one service function and its tests.

## Risks

- Low risk. The change is narrowly scoped: it only fires when a patch explicitly includes `status`, and it only ever nulls `errorReason` — no other field is touched.
- `errorReason` remains an internal, non-`PATCH`-able field from the outside. This PR only changes what happens to it *internally* when a caller changes `status` through the generic update path. It does not add a new externally settable field, and does not touch the `clear-error` endpoint or its permission check.
- Behavior change in one corner case: if a caller sends both `status` and `errorReason` in the same patch, and `status` is not `error`, the supplied `errorReason` is now dropped rather than persisted. This is intentional — it enforces the same invariant `heartbeat.ts` already applies on finalization, applied consistently everywhere `status` changes.
- The existing guards that reject status changes on `terminated` and `pending_approval` agents run earlier in the update function and are unchanged by this PR. They still throw before this invariant is ever reached.
- No schema change and no migration needed.
- No relevant documentation, telemetry event, OpenTelemetry span, or run-log event needed updating: this is an internal write-path invariant with no new external contract.

## Model Used

Claude (Anthropic), model ID `claude-sonnet-5`, accessed through the Claude Code CLI in an agentic coding session with file-edit, test-execution, and git/gh tool use in the repository workspace. The same model implemented the fix, wrote the added tests, and opened this PR.

## Checklist

### DIFF
--- a/server/src/__tests__/agents-service-generic-status-update-error-reason.test.ts
+++ b/server/src/__tests__/agents-service-generic-status-update-error-reason.test.ts
@@ -1,0 +1,164 @@
+import { randomUUID } from "node:crypto";
+import { afterAll, afterEach, beforeAll, describe, expect, it } from "vitest";
+import { agents, companies, createDb } from "@paperclipai/db";
+import {
+  getEmbeddedPostgresTestSupport,
+  startEmbeddedPostgresTestDatabase,
+} from "./helpers/embedded-postgres.js";
+import { agentService } from "../services/agents.ts";
+
+const embeddedPostgresSupport = await getEmbeddedPostgresTestSupport();
+const describeEmbeddedPostgres = embeddedPostgresSupport.supported ? describe : describe.skip;
+
+if (!embeddedPostgresSupport.supported) {
+  console.warn(
+    `Skipping embedded Postgres agent service tests on this host: ${embeddedPostgresSupport.reason ?? "unsupported environment"}`,
+  );
+}
+
+describeEmbeddedPostgres("agent service generic status update — errorReason invariant", () => {
+  let db!: ReturnType<typeof createDb>;
+  let tempDb: Awaited<ReturnType<typeof startEmbeddedPostgresTestDatabase>> | null = null;
+
+  beforeAll(async () => {
+    tempDb = await startEmbeddedPostgresTestDatabase("paperclip-agent-generic-status-");
+    db = createDb(tempDb.connectionString);
+  }, 20_000);
+
+  afterEach(async () => {
+    await db.delete(agents);
+    await db.delete(companies);
+  });
+
+  afterAll(async () => {
+    await tempDb?.cleanup();
+  });
+
+  async function seedCompany() {
+    const companyId = randomUUID();
+    const issuePrefix = `T${companyId.replace(/-/g, "").slice(0, 6).toUpperCase()}`;
+    await db.insert(companies).values({
+      id: companyId,
+      name: "Paperclip",
+      issuePrefix,
+      requireBoardApprovalForNewAgents: false,
+    });
+    return companyId;
+  }
+
+  async function seedAgent(
+    companyId: string,
+    overrides: Partial<typeof agents.$inferInsert> = {},
+  ) {
+    const agentId = randomUUID();
+    await db.insert(agents).values({
+      id: agentId,
+      companyId,
+      name: `Agent-${agentId.slice(0, 8)}`,
+      role: "engineer",
+      status: "idle",
+      adapterType: "codex_local",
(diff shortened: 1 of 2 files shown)


bruchris:
Re-triggering the commitperclip PR Review check after completing the required PR description template (no code changes).

Issue #12615, opened 2026-08-31
Onboarding motion constants bypass the token layer


tonio-alucema:
**What existing behavior does this improve?**
JavaScript-driven motion in the onboarding wizard. `AGENTS.md` requires every motion value in `ui/src/components/**` to come from the token layer in `ui/src/index.css`, but the onboarding motion constants are literals in TypeScript, so a change to the house curves does not reach them.

**Subsystem affected**
`ui/src/components/onboarding/` — chiefly `onboarding-motion.ts` and `SleepingZs.tsx`.

**Current behavior**
`ui/src/components/onboarding/onboarding-motion.ts` declares its easings and durations as literals, and two of them duplicate existing tokens exactly:

| Constant | Value | Token with the same value |
|---|---|---|
| `STEP_EASE` (line 8) | `[0.16, 1, 0.3, 1]` | `--motion-ease-out-expo` |
| `TAG_SWAP_EASE` (line 84) | `[0.4, 0, 0.2, 1]` | `--motion-ease-standard` |

Durations are literal seconds throughout — `0.28` (lines 15, 47), `0.55` (56), `0.34` (85), `0.26` (86), `0.12` (105), `0.22` (109). `SleepingZs.tsx` does the same with its own curve (`[0.39, 0.575, 0.565, 1]`, line 133) and its randomised flight durations.

Retuning `--motion-ease-standard` or the `--motion-duration-*` tier therefore changes CSS-driven motion across the app while the onboarding wizard keeps its old feel, and the two drift apart silently.

**Proposed behavior**
Source these from the token layer, so there is one place to tune motion.

The obstacle worth naming up front: `motion/react` needs numbers and cubic-bezier arrays, while tokens are CSS strings, so this cannot be a plain `var()` substitution. There is already a precedent for the conversion — `parseCssTimeMs` in `ui/src/components/task-chat/motion-tokens.ts:87` — and that file's `MOTION_TOKENS` registry is the obvious model for naming what onboarding consumes.

Worth deciding as part of this: whether `scripts/check-token-gates.mjs` grows a fifth gate for literal easing arrays and durations in `ui/src/components/**/*.ts{,x}`. It currently runs four gates — colour literals, arbitrary bracket values, raw `font-size`, legacy `hsl(var())` — and passes clean on this code, which is why the drift was never caught mechanically.

**Reason and benefit**
[...]

PR #12726, opened 2026-09-02
Onboarding: Figma pass over the tenant arc

tonio-alucema:
## Thinking Path

> - Signing up crosses two codebases: `paperclip-cloud` names the organization, then hands off to the tenant wizard in `ui/`, which creates the first agent, connects a model, and launches the first task
> - The two halves were designed at different times, so they did not look like one product, and the tenant's connect step arrived with a model source already chosen
> - A step that opens on an answer is not asking a question. A customer could pass the model step having touched none of it, and be hired against whatever the saved draft happened to carry
> - This pull request takes the four tenant steps to the current design, and makes the model choice explicit: nothing is selected on arrival, and the step cannot advance until the customer selects something they can see
> - The benefit is that the walk reads as one product, and no agent is hired against a model the customer never picked

No existing issue covers this. Described in-PR, as a bug report.

**What happened**

The onboarding connect step opened with a model source already selected, and its
sign-in panel already showing. The step advanced whether or not the customer
touched it.

A saved draft made this worse. The draft records an adapter, and the step read
that record as "the customer chose this". Two cases followed:

- The registry no longer carries the adapter. The row hides it. Nothing is
  highlighted, no input panel opens, and the button still advances.
- The adapter is in the advanced list but not the tile row. The row never shows
  it. Same result.

In both cases the step shows an unanswered question and hires against a name the
customer cannot see.

**What was expected**

The step asks which model to connect. It should show that nothing is chosen
until the customer chooses, and it should not advance until they do.

**Steps to reproduce**

1. Start onboarding and reach "Connect a model".
2. Look at the two source tiles. One is already selected.
3. Press Next without touching the row. The wizard advances and hires.

**Environment**

Tenant UI (`ui/`), all environments. Present on `master` before this PR.

**Related**

- paperclipai/paperclip-cloud#334 — the naming step before the hand-off
- paperclipai/paperclip-id#66 — the front door and code step before that
- #12615 — motion tokens in `onboarding-motion.ts` (pre-existing, out of scope here)

## What Changed

- The connect step opens with no source selected. `sourcePicked` starts false unless a saved draft carries one.
- The step cannot advance until a source is selected in the visible row. The gate reads `sourceSelected`, not `sourcePicked`.
- The snap effect, which replaces an adapter the registry no longer carries, now clears the selection. The snap removes a choice; it does not make one.
- The input canvas opens only for a selected source. It also opened for a pending sign-in, which put a provider panel on a step that had asked nothing.
- Tile selection is a fill, not a border. Both tiles keep the same border.
- The close control is removed from every step of the wizard.
- The arc draws to a 424px column, not 480px. The name field fills that column, is 44px tall, and is filled rather than outlined.
- The forward button reads "Next" through the arc, and "Get started" on the last step.
- The login spinner reads "Preparing...". The OAuth panels number their two fields and show the URL above the code.
- Storybook: the arc stories wait for the destination heading, not for a button name. The entry draft no longer names an adapter.
- e2e: three specs select a source before they advance, and wait for the button to be enabled.

## Verification

```
cd ui && npx vitest run src/components src/adapters src/pages storybook
```

3401 passed, 339 files.

```
cd ui && npx tsc -p tsconfig.json --noEmit
node scripts/check-token-gates.mjs
```

Both clean.

Each new test was checked against the fix reverted, so none of them passes
against the defect:

- `will not advance on a saved adapter the step no longer offers` fails when the snap does not clear the selection.
- `will not advance on a saved adapter the row does not show` fails when the gate reads `sourcePicked`.
- Neither stands in for the other. Each passes with the other half rolled back.

Screens were compared against the design frames in Storybook at
`onboarding-agent-arc--*`.

**One test was removed**, not replaced: `hydrates again when the same company
comes back through onboarding`. It closed the wizard with the close control,
which no longer exists. Three substitutes were written and each one passed
against a wizard with the behaviour deleted, so none of them tested anything.
The reason is recorded in the file where the test stood. What the hydration
marker guards is still covered by `seeds the lead agent's instructions with the
mission it was never asked for`.

## Risks

**The wizard has no exit.** This is deliberate. Nothing downstream of the
connect step works until a model is connected, so an exit there opens onto a
broken app.

One entrance is not covered by that reasoning. `SidebarCompanyMenu.tsx:323`
opens the wizard at step 1 for "create a new organization". An existing user who
starts that and changes their mind cannot leave. `handleClose` is now reachable
only from `onOpenChange`, which nothing fires. This is known and not fixed here.

**Behavioural shift.** A customer who returns to a draft that named a model
source now has to select it again. This is intended: the previous behaviour is
the defect above. No data is lost — the draft still records the adapter.

Everything else is presentation. No migration, no API change, no change to what
is written to the server.

## Model Used

Claude Opus 5 (`claude-opus-5`), extended thinking, run through Claude Code with
tool use and code execution.

## Checklist

- [ ] I have updated relevant documentation to reflect my changes — no user-facing docs cover the wizard's steps
- [ ] All Paperclip CI gates are green — re-running after the e2e and gating fixes

### DIFF
--- a/tests/e2e/conference-room-typing-intro.spec.ts
+++ b/tests/e2e/conference-room-typing-intro.spec.ts
@@ -82,10 +82,22 @@ async function runOnboardingWizard(page: Page, companyName: string) {
   // name and hires under the neutral `general` role.
   await page.waitForSelector("#onboarding-agent-name", { timeout: 30_000 });
   await page.locator("#onboarding-agent-name").fill("Ada");
-  await page.getByRole("button", { name: /^Next/ }).click();
+  await page.getByRole("button", { name: /^Next$/ }).click();
 
-  // Step 4: adapter (claude_local default); heartbeat is intercepted.
-  await page.getByRole("button", { name: /^Connect$/ }).click();
+  // Step 4: pick a model source, then advance. Nothing is selected on arrival
+  // — the row is a question, not a confirmation — so the CTA is disabled until
+  // a tile is pressed. By role rather than by label: which adapters the tiles
+  // offer depends on the registry this environment reports.
+  const source = page.getByRole("radio").first();
+  await source.waitFor({ timeout: 30_000 });
+  await source.click();
+
+  // The forward button reads "Next" here too, so wait for it to enable rather
+  // than for it to appear — it is already on screen, disabled, and clicking a
+  // disabled button raises nothing and does nothing.
+  const connectNext = page.getByRole("button", { name: /^Next$/ });
+  await expect(connectNext).toBeEnabled({ timeout: 30_000 });
+  await connectNext.click();
 
   // Step 5: review → Get started creates the first task and opens its
   // detail page.
--- a/tests/e2e/onboarding.spec.ts
+++ b/tests/e2e/onboarding.spec.ts
@@ -219,9 +219,17 @@ test.describe("Onboarding wizard", () => {
     await page.locator("#onboarding-agent-name").fill("Ada");
     await page.getByRole("button", { name: "Next" }).click();
 
-    // Step 4 (Connect a model): the default adapter is claude_local, and the
-    // signal above reports no ready credential, so the login panel must show
-    // with no button to reuse a saved login.
+    // Step 4 (Connect a model): pick a source. The step arrives with nothing
+    // selected — the tile row is a question, not a confirmation — and the login
+    // panel is what the answer opens, so there is nothing to assert until one
+    // is pressed. By role rather than by label, because which adapters the row
+    // offers depends on the registry this environment reports.
+    const source = page.getByRole("radio").first();
+    await source.waitFor({ timeout: 30_000 });
+    await source.click();
+
+    // The signal above reports no ready credential, so the login panel must now
+    // show, with no button to reuse a saved login.
     //
     // The panel names the provider rather than the plumbing it runs on, so this
     // title is per-adapter. "Sign in to the environment" is now only the fallback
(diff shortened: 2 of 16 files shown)


tonio-alucema:
Thanks — the gating defect was real, and it had a second shape you did not name.

**Fixed in 776879f08.**

Your case: a saved draft names an adapter outside `recommendedAdapters`, so
`sourcePicked` stays true, no tile reads as chosen, and Next advances against the
hidden name. Correct.

The fix needed two changes, because there are two ways in and they do not share a
cause.

1. **The adapter is gone from the registry entirely.** The snap effect at
   `OnboardingWizard.tsx:1070` replaces it with the first visible one, which is
   right — but it left `sourcePicked` true, so the replacement arrived *already
   chosen*: a filled tile and an open sign-in panel on a step nobody had
   answered. That re-creates by the back door exactly the preselection this PR
   removes. The snap now clears the pick. It removes a choice; it does not make
   one.

2. **The adapter is in the advanced list but not the tile row.** The snap's
   "visible" is recommended *plus* advanced, so a saved `opencode_local`
   satisfies it and is left alone — while the row, which is
   `recommendedAdapters`, never shows it. No snap fires here, so only the gate
   catches it. It now reads `sourceSelected` (off the visible row) rather than
   `sourcePicked` (off the draft).

Both are tested, and I checked each test against the *other* half reverted:
`will not advance on a saved adapter the step no longer offers` still passes with
the gate rolled back, and `will not advance on a saved adapter the row does not
show` still passes with the snap rolled back. So neither test stands in for the
other, and neither passes against the defect.

**On the motion token** (`CANVAS_ENTER_TRAVEL` in `onboarding-motion.ts`): that
file states every travel distance as a raw number — `CANVAS_CONTENT_TRAVEL` and
`TAG_SWAP_TRAVEL` are both plain constants on `master`. The new one follows its
neighbours rather than introducing a second convention beside them. Moving the
whole file onto the token layer is tracked in #12615, filed earlier for exactly
this; doing it here would mix a token migration into a design pass.

tonio-alucema:
Valid, and fixed in aa2bb47ca.

You were right that the keystroke bypassed the gate. The cause was that the
condition was written out twice: when the button gained `sourceSelected` and
`adapterEnvLoading`, `handleKeyDown` kept the shorter list it was written with.
Patching the keyboard to match today's button would leave the same drift for the
next condition, so both now ask one predicate, `connectStepReady`.

`loading` stays out of that predicate deliberately — the keyboard returns on it
before reaching any step, for a reason particular to keystrokes: a second Enter
re-enters a handler whose guard is state the first has not written yet.

One note on the test, because the first version of it was wrong in a way worth
recording. I wrote it against a saved `opencode_local` — non-recommended, so
absent from the tile row — and it passed. It also passed with the fix reverted:
`opencode_local` bails inside `handleGiveHeartbeat` before the hire, so `hire`
was never called whether the gate existed or not. The assertion held for a
reason that had nothing to do with the defect.

The committed version goes through the snap path instead: a saved adapter the
registry no longer carries is replaced with `claude_local` and the pick cleared,
so the row shows two tiles with neither chosen, and a keystroke that gets
through hires a real, offerable adapter nobody selected. That version fails with
the keyboard condition rolled back, and fails on the tile assertion with the
snap's clear rolled back.

3402 tests passing, typecheck and token gates clean.

PR #12556, opened 2026-08-31
fix(issues): do not expire pending interactions when an issue leaves blocked

Sharri747:
## Thinking Path

> - Paperclip is the open source app people use to manage AI agents for work
> - An agent can raise a thread interaction on an issue, for example a confirmation request. The interaction stays `pending` until somebody answers it
> - An issue can also enter and leave `blocked` status. The first version of this branch expired pending interactions when an issue left `blocked`
> - That link does not hold. Nothing records which interaction a block was raised for, and nothing raises a block because of an interaction. `blocked` is written by dependency blockers, by recovery classifiers, by an operator PATCH, and by the pre-dispatch workspace guard. None of them read the interaction table
> - Expiring on blocked-exit therefore cancels interactions that the block never owned
> - This pull request removes that sweep and adds the tests that show why the link does not hold. It also gives `buildAdministrativeOutcomeResult` an explicit return type
> - The benefit is that a pending interaction is cancelled only by something that actually owns it

## Linked Issues or Issue Description

No existing issue. The problem is described below.

**Bug: what happened?**

A status reset on a `blocked` issue expired every pending thread interaction on that
issue. The sweep used `blockedTransitionAt` as if it recorded which interaction the block
was waiting on. That field does not carry interaction meaning. It is the generic "this
issue entered blocked" stamp, shared with dependency wake cycles and attention dedup.

**Expected behavior**

An issue leaving `blocked` must not cancel interactions the block did not raise. A pending
interaction is cancelled by its own resolution, by the issue closing, or by the addressee
being deleted.

**Version**

Reproduced on `master`.

## What Changed

- Removed the blocked-exit interaction sweep. A status reset on a blocked issue no longer
  expires pending interactions
- Gave `buildAdministrativeOutcomeResult` an explicit `IssueThreadInteractionResult`
  return type. The secret-proposal branch is now spelled out instead of spread inline, so
  `status` and `version` hold their literal types and the declared return type can catch a
  wrong value
- Added tests that record why the blocked-exit link does not hold, and that cover the
  administrative outcome paths

## Verification

    ./node_modules/.bin/vitest run \
      server/src/__tests__/issues-service.test.ts \
      server/src/services/issue-thread-interactions.test.ts \
      server/src/__tests__/issue-agent-mutation-ownership-routes.test.ts \
      server/src/__tests__/issue-execution-policy-routes.test.ts

Result on 3 September 2026: 4 files passed, 245 tests passed.

## Risks

[...]

PR #12734, opened 2026-09-02
fix(issues): stamp blockedTransitionAt at every blocked-entry write site and notify the owner

### LINKED ISSUE #10406: Creating a blocked issue does not wake its unblock owner
## Pre-submission checklist

- [x] I can reproduce this on current `master`.
- [x] I confirmed the failure is in Paperclip core, not an adapter, provider, or local configuration.

## What happened?

Creating an issue directly in `blocked` status with an agent-owned `unblockDescriptor` returns 201 but does not record a routable blocked transition and does not request a wake for the named owner.

This affects both:

- `POST /api/companies/:companyId/issues`
- `POST /api/issues/:id/children`

The create service inserts the supplied blocked status without setting `blockedTransitionAt`. The routes then only run assignment wake logic; unlike the working PATCH transition path, they never call `deliverAgentUnblockNotification`.

## Expected behavior

A newly created blocked issue with an agent owner should:

1. Persist a non-null `blockedTransitionAt`.
2. Mint one `issue_unblock_requested` wake for that owner using `issue-unblock:<issueId>:<transition timestamp>` as the idempotency key.
3. Persist `blockedOwnerNotifiedAt` after the wake request.
4. Not mint another wake when a root create is replayed through the same create idempotency key.

## Steps to reproduce

1. Create an active agent.
2. POST a root issue, or a child issue, with:
   - `status: "blocked"`
   - `unblockDescriptor: { "owner": { "agentId": "<agent-id>" }, "action": "Review the finding" }`
3. Read the persisted issue and `agent_wakeup_requests`.

Observed on unmodified master:

```json
{
  "httpStatus": 201,
  "persisted": {
    "status": "blocked",
    "blockedTransitionAt": null,
    "blockedOwnerNotifiedAt": null
  },
  "ownerWakes": []
}
```

The PATCH control path (`todo`/`in_progress` → `blocked`) does stamp the transition and request the owner wake.

`4eace88f6b3eacbe51b8fa2b2fdbd2c75580c9c0`

## Deployment mode

Local dev (`pnpm` + embedded test Postgres)

## Installation method

Built from source

## Agent adapters involved

Not adapter-specific; this is core issue lifecycle/wake routing.

## Database mode

Embedded Postgres test database.

## Access context

Board/local-trusted route reproduction.

## Node.js / operating system

- Node.js `v22.22.3`
- Linux `6.8.0-117-generic` x86_64

## Additional context

PR #10366 fixes the downstream reporting-line manager authorization and stale-queue cancellation for an owner wake that already exists. Its tests seed that wake manually. It does not stamp blocked creation or mint the missing wake, so both changes are required for the complete manager-owner flow.

A deterministic embedded-Postgres regression reproduces the failure for both root and child creation. No secrets, user data, or private logs are included above.

## Privacy checklist

- [x] I reviewed this report for PII, usernames, paths, API keys, tokens, company names, and task content.

Sharri747:
## Thinking Path

> - Agents get work through issues. An issue can enter `blocked` status. The owner named in its unblock descriptor must get a wake
> - The wake depends on two fields. `blockedTransitionAt` records when the block started. `blockedOwnerNotifiedAt` records that the owner was told
> - Only two write sites set `blockedTransitionAt`. Other write sites leave a row in `blocked` with no stamp. That row can never be notified
> - Notification also happens on one write edge only. A row repaired by a self-heal branch never crosses that edge. It stays silent
> - This pull request sends every blocked-entry write through one helper. It also adds a reconciler that notifies repaired rows on the heartbeat tick
> - The benefit is that a blocked issue always wakes its owner, whichever code path blocked it

Fixes: #10406

Refs #12556 — that pull request fixes the create path only. This pull request replaces
that fix with the general one. See "Risks" for the sequencing.

## What Changed

- Added `server/src/services/routable-blocked.ts`. It exports `deriveBlockedEntryPatch`
  and `BLOCKED_EXIT_PATCH`. `deriveBlockedEntryPatch` is idempotent. It does not change a
  stamp that is already present
- Sent the blocked-entry writes in `create()`, `update()`, `release()` and the heartbeat
  pre-dispatch unrunnable-worktree guard through that helper
- Gave `importIssues()` the same stamp inline. An import keeps the imported timestamp
  (`row.updatedAt ?? row.createdAt`) instead of the current time, so an import does not
  look like a block that started at import time
- Added `server/src/services/blocked-owner-notification-reconciler.ts`. It runs on the
  heartbeat tick. It finds rows that are blocked, have a stamp, and have no
  `blockedOwnerNotifiedAt`. It delivers the wake for those rows
- At the heartbeat pre-dispatch guard, the stamp advances unconditionally on a fresh
  entry. It falls back to the idempotent helper when the row is already blocked. This
  stops the guard from advancing the dependency-wake cycle key on every tick
- `update()` also advances a stamp that predates the routable-blocked rollout when an unblock descriptor is attached to an already-blocked row, and clears `blockedOwnerNotifiedAt`, so an imported or restored historical block becomes eligible at the moment a request is actually made on it. Fires at most once per row, never for a post-cutoff stamp. Folded in from #12753, which is closed as superseded

## Verification

Run the targeted suites:

      server/src/__tests__/blocked-transition-stamp.test.ts \
      server/src/__tests__/blocked-owner-notification-reconciler.test.ts \
      server/src/__tests__/heartbeat-workspace-branch-containment.test.ts \
  server/src/__tests__/issues-service.test.ts

Result on 5 September 2026 (`bdd941b2d`): 5 files, 160 tests passed. `db` and `server` typecheck clean.

Typecheck:

    pnpm --filter @paperclipai/db typecheck        # clean
    cd server && ./node_modules/.bin/tsc --noEmit  # clean, 0 errors

## Risks

- No migration. This pull request is code only; `packages/db` is unchanged. A backfill for
  rows left carrying a stale stamp by the old exit paths is still worth doing, but it is
  hygiene rather than part of this fix and will follow as its own change. Those rows are
  dormant meanwhile: the dependency backstop only reads the stamp where `status = 'blocked'`
- Behavioural shift: rows that were silently unnotifiable now produce one wake each on the
  first heartbeat tick after deploy. The wake is idempotent on
  `issue-unblock:<issueId>:<transition timestamp>`, so a row produces one wake, not many
- Sequencing: this is resolved. #12556 previously carried a commit doing the same job for
  `create()` only. That commit has been removed and #12556 is now scoped to the
  blocked-exit interaction concern, so the two no longer overlap and can merge in either
  order

## Model Used

## Checklist

- [ ] I have updated relevant documentation to reflect my changes — no user-facing documentation changes are needed. The behaviour is internal to the issue service and the heartbeat
- [ ] All Paperclip CI gates are green — not known until CI runs on this pull request


Sharri747:
Thank you for the review. All three findings are valid. Two are fixed in `a6562cd`. The
third is deferred, with reasoning below.

**Finding 1 — stale cycle marks the current cycle notified. Fixed.**
The completion write matched on `id` and `companyId` only, so any move between the
candidate select and that write was invisible to it: a re-entry into `blocked` with a new
stamp, an owner reassignment, or the edge-triggered path in `routes/issues.ts` getting
there first. The write is now a compare-and-set against the delivered snapshot — the
stamp, that snapshot's owner agent id, and a still-null `blockedOwnerNotifiedAt` — and it
returns the rows it touched. A zero-row result counts as `skipped`, so the current state
stays eligible for the next sweep instead of being written off. The wake carries the
snapshot's cycle key, so it cannot double-notify the cycle it was built for.

**Finding 2 — skipped rows starve valid notifications. Fixed.**
The candidate query was unordered and included the two classes the delivery helper
permanently no-ops on: a board-owned descriptor, and a stamp older than the rollout
cutover. Neither becomes deliverable by being selected again. Both are now excluded in
SQL, and the query orders oldest transition first. Every row in a batch can now make
progress, so a backlog drains, and it drains fairly. The board-owned test is updated to
match — such a row is excluded by the query rather than scanned and skipped in the loop.
The behaviour that matters is unchanged: no wake, never marked notified.

Added regression tests for a cycle re-entry during delivery, an owner reassignment during
delivery, and a full batch of undeliverable rows in front of a deliverable one. 4 files,
33 tests pass. Server typecheck is clean.

**Finding 3 — historical stamp blocks later notification. Valid, deferred.**
This one is real: a row imported with a pre-rollout timestamp that later receives an
agent-owned `unblockDescriptor` fails the `ROUTABLE_BLOCKED_ROLLOUT_AT` check forever, so
neither the edge path nor this sweep reaches it.

It is a different defect from the one this pull request fixes. This pull request is about
write sites that failed to stamp. Finding 3 is about rows that are stamped correctly but
carry a timestamp the eligibility cutoff excludes. Fixing it means changing what the
rollout cutoff means — either re-stamping when a descriptor is attached to an
already-blocked row, or scoping the cutoff to something other than the transition
timestamp. Either changes eligibility for existing historical data and wants its own
migration and its own review, rather than being folded into this change late.

Happy to open a separate issue for it if you would like one.

Sharri747:
Both findings against `a6562cd` are fixed in `034b3ac`.

**Action changes bypass delivery fence.** The compare-and-set matched the owner agent id but not the rest of the descriptor. `action` is delivered in the wake payload, so a descriptor whose action changed during delivery is a different request even when the owner and the stamp are unchanged: the wake carried the old action, and the write retired the new one as notified. The fence now matches the descriptor whole, as jsonb. jsonb equality is key-order insensitive, and matching the whole value also covers fields added later.

**Failed candidates block queue progress.** You are right, and this one was introduced by the ordering in `a6562cd`. A delivery that throws leaves its row eligible, and ordered by oldest transition that row then held the head of every batch. Before the ordering this was intermittent; the ordering made it deterministic. A row that fails delivery now enters a cooldown and is excluded from the candidate query until it expires, with the backoff doubling per consecutive failure from 1 minute to a 1 hour cap. So a permanently broken row costs one attempt per cap window instead of one per tick, and a transient one recovers quickly. The map is per process and bounded; losing it on a restart costs one extra attempt per row, which is the safe direction to fail.

Regression tests added for an action-only descriptor change during delivery, and for a permanently failing oldest candidate not starving a younger deliverable one, including that the cooldown defers the failing row rather than dropping it.

The third finding, the historical import stamp, is now fixed separately in #12753 as suggested. 3 files, 28 tests pass here; server typecheck clean.

Sharri747:
Both open findings are addressed in `bdd941b2d` — three commits on top of `08591a050`, a fast-forward, no force-push.

**Failure eviction starves notifications. Fixed.**
You are right, and the mechanism is structural rather than a tuning problem. The cooldown map bounds how many failing rows can be *hidden* from the candidate query, so with N permanently failing rows at least N − 5,000 are visible on every tick whatever the eviction order. Once that exceeds the batch size, the batch is refilled with recycled failures on every tick and their backoff never grows. No eviction policy fixes that: what carries the sweep forward had to change.

The candidate query now resumes from a keyset cursor. Each tick selects strictly after the previous batch's last `(blockedTransitionAt, id)` and wraps to the head when a batch comes back short. A row that stays eligible after its turn — a delivery that threw, a fence that did not apply — is only seen again when the sweep comes round to it, however many broken rows sit ahead of it. Every eligible row gets one turn per pass over the queue. Oldest-first ordering is kept, and the `id` tiebreak means tied stamps cannot skip or repeat a row at a batch boundary.

The cooldown map is unchanged in mechanism but demoted in role: it now only limits how often a broken row is retried. Its bound no longer affects whether the rows behind it are reached, so eviction is harmless — an evicted row's next attempt is ungated by cooldown, not moved to the front. The cursor is per process and per sweep scope; losing it on restart means the next pass starts from the head. The exclusion list is bound as query parameters, at most 5,000 of them.

Regression test: one more batch of permanently failing rows than the map can hold, ahead of one deliverable row, ticking faster than the base cooldown so no failure expires. It fails on `08591a050` — the deliverable row is never reached — and passes now, with the row reached within the first pass.

**Historical stamp blocks later notification. Fixed here, folded in from #12753.**
On 3 September I said this would be fixed separately and opened #12753 for it. I am reversing that. The reason is not new technical information: you re-listed the finding against this diff, and looking at the split again with that in front of me, it was wrong on its merits. The reconciler that delivers the notification the stamp makes eligible exists only on this branch, so #12753 merged alone advances a stamp that nothing reads, and this branch merged alone leaves the finding open. They are one change, and your finding on #12753 — that nothing there dispatches the notification — said the same thing from the other side.

`update()` now advances a pre-cutoff stamp when an unblock descriptor is attached to an already-blocked row, and clears `blockedOwnerNotifiedAt`, so the reconciler picks the row up on its next pass. It fires at most once per row and never for a post-cutoff stamp, because that stamp is the dependency-wake cycle key; a test pins each. The arm is mutually exclusive with the existing no-stamp self-heal arm, so the two do not interact. #12753 is closed as superseded by this pull request.

Verification on `bdd941b2d`: `blocked-owner-notification-reconciler`, `blocked-transition-stamp`, `routable-blocked`, `issues-service` and `heartbeat-workspace-branch-containment` test files — 5 files, 160 tests pass. Server typecheck clean. `packages/db` is still untouched, and the branch still merges clean against `master`.

Sharri747:
Confirmed, and fixed in `fb36bb5fe` (fast-forward on `bdd941b2d`).

The cursor wrapped only on a short batch. If new eligible rows arrive at least as fast as a batch drains them, no batch is ever short, the cursor advances indefinitely, and anything behind it — a failed row whose cooldown has expired, or a row made eligible with a stamp older than the cursor — is never reached. That is the same starvation in a third form.

A pass is now bounded to the rows that were eligible when it started. Its end key is read once, at the first tick of the pass, and the cursor wraps as soon as a batch reaches that key or comes back short, whichever is first. So a pass is at most ceil(N / 250) ticks for the N rows eligible at its start; rows that arrive during a pass wait for the next one; and every row behind the cursor gets its turn on the next pass. Cooldown is not applied when reading the end key, so a failing tail cannot end a pass early.

Regression test: one row with the oldest stamp that fails once, a backlog of a little over two batches, and a full batch of new rows inserted before every subsequent tick. On `bdd941b2d` the row is attempted once and never again within twice the pass length; on `fb36bb5fe` it is delivered on the first tick of the second pass. Server typecheck clean; the reconciler file is 13 tests passing; still merges clean against `master`.

PR #12742, opened 2026-09-03
fix(issues): validate issue list UUID filters

### LINKED ISSUE #11857: GET /companies/:companyId/issues?createdByAgentId= is silently ignored (returns full unfiltered set)
## Thinking Path

> - Paperclip's `GET /api/companies/{companyId}/issues` supports filtering by `assigneeAgentId` (and several other fields) via query params, applied in `issueService.list()` (`server/src/services/issues.ts`) using the standard `if (filters?.x) conditions.push(eq(issues.x, filters.x))` pattern.
> - `createdByAgentId` is a real, populated column on `issues` — it's read out in every response payload (`server/src/services/issues.ts:3215`, `server/src/routes/issues.ts:2372`) — but it is never accepted as an input filter.
> - `IssueFilters` (`server/src/services/issues.ts:537`) has no `createdByAgentId` field, and `server/src/routes/issues.ts`'s issue-list handler (~line 6048 for `GET /companies/:companyId/issues`, ~line 6244 for the related list path) never reads `req.query.createdByAgentId` at all — unlike `assigneeAgentId`, `participantAgentId`, `status`, etc., which are all read and forwarded.
> - Because of the still-open fail-open behavior in #10951 (list endpoints silently drop any query param they don't recognize and return the full unfiltered set instead of erroring), a caller who passes `?createdByAgentId=<uuid>` gets a 200 with a plausible-looking, correctly-shaped result set — every issue has a `createdByAgentId` field — that is actually completely unfiltered. There is no way for the caller to tell "the filter matched nothing" from "the filter never ran."
> - This is the same shape of bug as #7414 (`priority` query param silently dropped because `IssueFilters` had no `priority` field) — a route/service field the API already exposes as *output* was never wired as an *input* filter. The fix is the same pattern: add `createdByAgentId?: string | string[]` to `IssueFilters`, read `req.query.createdByAgentId` in the route, and add a `conditions.push(...)` block in `issueService.list()`/`count()` identical to the existing `assigneeAgentId` block.

## What happened?

`GET /api/companies/{companyId}/issues?createdByAgentId=<agentId>` returns HTTP 200 with a full, unfiltered page of issues. Every returned issue has some *other* agent's ID in its own `createdByAgentId` field — never the one that was queried for.

## Steps to reproduce

```
# Baseline (no filter)
GET /api/companies/{companyId}/issues?limit=20

# "Filtered" — returns the exact same 20 issues, same order
GET /api/companies/{companyId}/issues?createdByAgentId=<some-other-agentId>&limit=20
```

Measured live on a running instance: the two responses were byte-identical in issue identifiers and order. The `createdByAgentId` distribution of the "filtered" response was a mix of several agents' IDs — none of them the one requested.

Contrast with a sibling filter on the same endpoint, which works correctly:

```
GET /api/companies/{companyId}/issues?assigneeAgentId=<agentId>&limit=10
# -> all 10 rows have assigneeAgentId == <agentId>, correct
```

## Expected behavior

`?createdByAgentId=<agentId>` should return only issues where `issues.createdByAgentId = <agentId>`, using the same `eq()`/`inArray()` pattern already applied for `assigneeAgentId` in `issueService.list()`.

## Root cause (read from `master`)

- `server/src/services/issues.ts:537` — `IssueFilters` interface has `assigneeAgentId`, `participantAgentId`, `assigneeUserId`, etc., but no `createdByAgentId`.
- `server/src/services/issues.ts:4238-4245` — the assignee filter is applied as:
  ```ts
  const assigneeAgentFilter = parseIssueAssigneeAgentFilter(filters?.assigneeAgentId);
  assertValidAssigneeAgentFilter(assigneeAgentFilter);
  if (assigneeAgentFilter === null) {
    conditions.push(isNull(issues.assigneeAgentId));
  } else if (assigneeAgentFilter) {
    conditions.push(eq(issues.assigneeAgentId, assigneeAgentFilter));
  }
  ```
  No equivalent block exists for `createdByAgentId` anywhere in the file — confirmed by `grep -n "filters?.createdByAgentId\|filters\.createdByAgentId" server/src/services/issues.ts` returning nothing.
- `server/src/routes/issues.ts` — the two issue-list query-building blocks (`assigneeAgentId: req.query.assigneeAgentId as string | undefined`, etc.) never read `req.query.createdByAgentId`.
- `issues.createdByAgentId` itself is a real, already-indexed-adjacent column — it's used in an *output* projection everywhere issues are serialized, and even appears in an OR condition elsewhere (`server/src/services/issues.ts:1464-1465`, a permission/ownership check unrelated to the list filter), so the column and its semantics already exist; only the list-endpoint input wiring is missing.

## Suggested fix

Same shape as #7414's `priority` fix:
1. Add `createdByAgentId?: string | string[]` to `IssueFilters`.
2. In `issueService.list()` (and the matching `count()`/blocked-inbox condition builder, mirroring how `status` and `priority` are each applied in three places), add a `conditions.push(...)` block for `createdByAgentId` using `eq()`/`inArray()`.
3. In `server/src/routes/issues.ts`, read `req.query.createdByAgentId` into the filters object passed to `svc.list()`, at both call sites that already forward `assigneeAgentId`.
4. Once #10951 (reject-unknown-query-params) lands, `createdByAgentId` should be added to that endpoint's allow-list so it stops silently no-op'ing in the interim between now and this fix.

## Related

- #7414 — identical bug shape, `priority` query param silently dropped (open, unmerged). This report follows the exact same pattern for `createdByAgentId`.
- #10951 — open PR that would make *any* unrecognized query param (including `createdByAgentId` today) return 400 instead of silently no-op'ing. That PR alone does not add `createdByAgentId` filtering — it only converts the silent failure into a loud one — so this issue is still needed for the actual filter implementation.
- #9350 — related but distinct: Board-view UI "Filter by Me (Creator)" showing a subset of tasks. That's a client-side/view-layer symptom; this issue is about the REST list endpoint's query param being a no-op for API/agent callers.

Reproduced against a live instance; route/service code above read from `master`.

## Deployment mode

Self-hosted server (agent-operated instance)

## Access context

Agent (bearer token), via `GET /api/companies/{companyId}/issues`

## Privacy checklist

- [x] Reviewed all pasted output for PII (usernames, file paths, API keys, tokens, company names) and redacted where necessary. Agent IDs shown are UUIDs with no identifying info attached.

ryanclark2:
## Thinking Path

> - The issue list API lets users and scripts filter company issues
> - Several UUID query filters reached SQL without route validation
> - Bad UUID text could become a database error instead of a client error
> - Two real issue fields were accepted by callers but did not filter rows
> - This pull request validates UUID filters at the route boundary
> - It also wires the creator and goal filters into the issue service
> - The benefit is that bad filters fail loud and valid filters return scoped rows

Fixes #11857
Refs #1966
Refs #6327
Refs #7432

## What Changed

- Return `400` for malformed UUID query values on the company issue list route.
- Preserve the `assigneeAgentId=null` sentinel for unassigned issues.
- Add `goalId` and `createdByAgentId` to the issue filter contract.
- Apply those two filters in normal lists, blocked inbox lists, and counts.
- Add route tests for malformed UUID params, valid combined UUID filters, and the new goal and creator filters.

## Verification

- `pnpm exec vitest run server/src/__tests__/issue-list-assignee-filter-routes.test.ts --no-file-parallelism --maxWorkers=1` passed. It ran 20 tests.
- Six mutation probes against the new empty-value test, one at a time, each reverted after: dropping the empty-skip in `parseIssueUuidFilters`; restoring the empty rejection in `parseIssueAssigneeAgentIdFilter`; moving the alias gate back onto the parsed filter; restoring the empty rejection in `assertValidUuidFilter`; and widening that exemption to "blank after trimming", which dies on a raw `Failed query` from Postgres. Five go red. One survivor, reported: writing the empty string through as the filter value stays green, because every consumer of that value gates on truthiness, so it is observationally equivalent.
- `pnpm exec vitest run server/src/__tests__/issues-service.test.ts --no-file-parallelism --maxWorkers=1` passed. It ran 123 tests.
- `git diff --check` passed.
- `pnpm --filter @paperclipai/server typecheck` did not reach server `tsc` because this machine does not have `cargo`, and the runner vendor build requires it.
- After building `@paperclipai/plugin-sdk`, direct `pnpm --filter @paperclipai/server exec tsc --noEmit` still fails on an existing Sentry type mismatch in `src/sentry.ts`.

## Risks

- Low risk for valid UUID callers. The tests cover valid combined filters with status, limit, and search.
- Existing callers that sent malformed UUID text now get `400` instead of a server error or empty result.
- **Status-code change, declared:** a malformed `assigneeAgentId` on `GET /issues` returned `422` before this PR and now returns `400`. The other seven UUID params returned `500` (or an empty result) before, so `400` is new for them either way. Callers that switch on `422` for this one param need updating.
- `assertValidUuidFilter` exempts exactly `""`, not "blank after trimming". It does not own the value it validates, so a whitespace-only string it waved through would still reach the query builders, which test the original string for truthiness and compare it against a uuid column — the 22P02 this PR exists to prevent. `" "` is asserted to throw on all nine service filters.
- **Empty values are unchanged:** `?projectId=` and its eight siblings, and `?parentId=&parentIssueId=<uuid>`, still return the unfiltered `200` on both `/issues` and `/issues/count`. An empty value means "filter absent", not "filter malformed" — it is the reading every query builder in `services/issues.ts` already implements. A new test pins that on both routes and at the service layer.
- Existing callers that sent `goalId` or `createdByAgentId` now get filtered rows instead of the full list.
- This server change cannot protect clients that ignore HTTP status codes. A separate private ops-script follow-up is filed for status checks before row parsing.

## Model Used

OpenAI Codex, GPT-5 coding agent, with terminal and repository editing tools.

## Checklist

### DIFF
--- a/server/src/__tests__/issue-list-assignee-filter-routes.test.ts
+++ b/server/src/__tests__/issue-list-assignee-filter-routes.test.ts
@@ -1128,6 +1128,12 @@ describeEmbeddedPostgres("issue list routes assigneeAgentId filter", () => {
     const app = createApp(companyId);
     const listPath = `/api/companies/${companyId}/issues`;
     const countPath = `/api/companies/${companyId}/issues/count`;
+    // `parentIssueId` is in these lists on purpose. It is not a member of
+    // `ISSUE_LIST_UUID_FILTER_NAMES` — it is the `parentId` alias, normalized by
+    // its own second `.trim()`/`isUuidLike` pair in the guard's alias branch.
+    // Nothing in the main name loop reaches that branch, so it is only pinned if
+    // the value axis walks the alias by name. It is last in each list because
+    // `parentId` must stay absent from the same request for the alias to be read.
     const listParams = [
       "assigneeAgentId",
       "participantAgentId",
@@ -1139,6 +1145,7 @@ describeEmbeddedPostgres("issue list routes assigneeAgentId filter", () => {
       "parentId",
       "descendantOf",
       "labelId",
+      "parentIssueId",
     ] as const;
     const countParams = [
       "assigneeAgentId",
@@ -1149,6 +1156,7 @@ describeEmbeddedPostgres("issue list routes assigneeAgentId filter", () => {
       "parentId",
       "descendantOf",
       "labelId",
+      "parentIssueId",
     ] as const;
     const errorFor = (param: string) =>
       param === "assigneeAgentId" ? "assigneeAgentId must be a UUID or 'null'" : `${param} must be a UUID`;
@@ -1179,10 +1187,12 @@ describeEmbeddedPostgres("issue list routes assigneeAgentId filter", () => {
       }
     }
     for (const param of countParams) {
-      const res = await request(app).get(countPath).query({ attention: "blocked", [param]: " " });
-      const label = `${param}: ${JSON.stringify(res.body)}`;
-      expect(res.status, label).toBe(200);
-      expect(res.body.count, label).toBe(baselineCount);
+      for (const value of [" ", "\t\n"]) {
+        const res = await request(app).get(countPath).query({ attention: "blocked", [param]: value });
+        const label = `${param}=${JSON.stringify(value)}: ${JSON.stringify(res.body)}`;
+        expect(res.status, label).toBe(200);
+        expect(res.body.count, label).toBe(baselineCount);
+      }
     }
 
     // Class 2 — a valid UUID with surrounding whitespace. `isUuidLike` trims
@@ -1206,10 +1216,12 @@ describeEmbeddedPostgres("issue list routes assigneeAgentId filter", () => {
       const value = randomUUID();
       const clean = await request(app).get(countPath).query({ attention: "blocked", [param]: value });
       expect(clean.status, `${param}: ${JSON.stringify(clean.body)}`).toBe(200);
-      const res = await request(app).get(countPath).query({ attention: "blocked", [param]: `${value} ` });
-      const label = `${param}=padded: ${JSON.stringify(res.body)}`;
-      expect(res.status, label).toBe(200);
-      expect(res.body, label).toEqual(clean.body);
+      for (const padded of [`${value} `, ` ${value}`]) {
+        const res = await request(app).get(countPath).query({ attention: "blocked", [param]: padded });
+        const label = `${param}=${JSON.stringify(padded)}: ${JSON.stringify(res.body)}`;
+        expect(res.status, label).toBe(200);
+        expect(res.body, label).toEqual(clean.body);
+      }
     }
 
     // Class 3 — non-empty text that is not a UUID. `"x"` is length 1 and
@@ -1255,6 +1267,24 @@ describeEmbeddedPostgres("issue list routes assigneeAgentId filter", () => {
       expect(res.status, label).toBe(400);
       expect(res.body, label).toMatchObject({ error: errorFor(param) });
       expect(res.body, label).not.toHaveProperty("count");
+    }
+
+    // Class 5 — the `null` token is case-folded. `assigneeAgentId` is the one
+    // filter that accepts a non-UUID value, and it accepts it only through
+    // `.toLowerCase()`. Drop that call and `NULL` stops being the null token,
+    // falls through to `isUuidLike`, and 400s a request that returns 200 today.
+    // Compare against the lowercase response so the pin holds whatever the
+    // null-assignee page contains.
+    for (const path of [listPath, countPath] as const) {
+      const base = path === listPath ? { status: "todo", limit: "20" } : { attention: "blocked" };
+      const lower = await request(app).get(path).query({ ...base, assigneeAgentId: "null" });
+      expect(lower.status, `${path}: ${JSON.stringify(lower.body)}`).toBe(200);
+      for (const token of ["NULL", "Null"]) {
+        const res = await request(app).get(path).query({ ...base, assigneeAgentId: token });
+        const label = `${path} assigneeAgentId=${token}: ${JSON.stringify(res.body)}`;
+        expect(res.status, label).toBe(200);
+        expect(res.body, label).toEqual(lower.body);
+      }
     }
   });
 
--- a/server/src/__tests__/issue-list-assignee-filter-routes.test.ts
+++ b/server/src/__tests__/issue-list-assignee-filter-routes.test.ts
@@ -3,7 +3,22 @@ import express from "express";
 import request from "supertest";
 import { eq } from "drizzle-orm";
 import { afterAll, afterEach, beforeAll, describe, expect, it } from "vitest";
-import { activityLog, agents, companies, companyMemberships, createDb, heartbeatRuns, issues, principalPermissionGrants } from "@paperclipai/db";
+import {
+  activityLog,
+  agents,
+  companies,
+  companyMemberships,
+  createDb,
+  executionWorkspaces,
+  goals,
+  heartbeatRuns,
+  issueLabels,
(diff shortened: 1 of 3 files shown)


ryanclark2:
## FoundingEng review — APPROVE

Reviewed at cited SHA `b9225e627845c73a268b0c9cf6f64992c691f7fa`; PR advanced one commit to `946cb45b` (preserves `parentId` precedence over `parentIssueId` alias). Reviewed at current head.

### Correctness verdict

**Route boundary (`server/src/routes/issues.ts`)**
- UUID validation fires at the route boundary before the service is called — correct placement.
- The `uuidFilterNames` loop covers all 9 affected params, returns `400` on non-UUID input.
- `parentIssueId` alias validated separately; with the second commit `parentId` wins when both are present, preserving prior `??` semantics.
- `assigneeAgentId` error code corrected from `422` → `400`; empty-string `assigneeAgentId` now returns `400` instead of silently filtering nothing.

**Service (`server/src/services/issues.ts`)**
- `goalId` and `createdByAgentId` added to `IssueFilters` and wired into SQL conditions in all three call sites: main `list`, `count`, and `blockedInboxIssueConditions`.
- `assertValidUuidFilter` adds defense-in-depth for direct service callers.

**Tests**
- 400-malformed test covers 8 params. Minor gap: `goalId`, `createdByAgentId`, and `parentIssueId` are validated at the route but absent from the `params` array. Non-blocking — the validation exists, it's just untested for those three.
- `goalId` + `createdByAgentId` positive-filter test verifies correct inclusion/exclusion.
- `parentId` precedence test covers the edge case.

All CI checks pass. No blocking findings.

_(GitHub formal approval blocked by single-account identity constraint — verdict delivered here per standard practice.)_

ryanclark2:
**Reviewer note — review this at `fe54637c2`, not `48cc60488`.** The board seat ALM-7317 names the older SHA; I cannot edit it once assigned, so the correction lives here.

Greptile's P1 on `server/src/routes/issues.ts` was right and I took the fix. `parseIssueUuidFilters` validated `parentIssueId` even when a valid `parentId` superseded it, so `?parentId=<valid>&parentIssueId=garbage` went from 200 to 400. `master` reads `parentId ?? parentIssueId`, so in that case the alias is never consulted — rejecting it is a compat break this ticket does not ask for. Now the alias is validated only when it is the value that actually reaches the filters.

Three assertions added: list route and count route both stay 200 on a superseded malformed alias, and an alias with no canonical `parentId` still 400s.

Mutation evidence at `fe54637c2`, control first:

```
GREEN | CONTROL (unmutated) 19/19
RED   | validate the superseded alias unconditionally      (2 failed)
RED   | drop alias validation entirely, fail-open          (2 failed)
```

On the prior head `c34e3436c`, `ci / General tests (server (4/5))` went red on `duplicate key ... (identifier)=(WDOG-7100)`. That is `server/src/__tests__/task-watchdogs-scheduler.test.ts:103` seeding `WDOG-${Math.floor(Math.random() * 10_000)}` into a globally unique column — a birthday collision, present on `master`, and untouched by this 3-file diff. Reviewer should verify that rather than take it from me.

ryanclark2:
ADR-005 pre-review: instrument B — 1 finding (seeding in the value-class test now duplicates the sibling test at `:1291`; declined — this file keeps each `it()` self-contained with its own fixtures). Instrument A absent: `code-quality-review` is a tracked file in `ryanclark2/almwell` and this is a `paperclipai/paperclip` change.

Round-8 head `8db015b34`. Test-only commit; `server/src/routes/issues.ts` and `server/src/services/issues.ts` are byte-identical to `f38d8b04d`.

PR #12753, opened 2026-09-03
fix(issues): make a pre-rollout blocked stamp eligible when an unblock owner is attached

Sharri747:
## Thinking Path

> - A blocked issue can name an unblock owner in its `unblockDescriptor`. That owner gets a wake so somebody is asked to unblock it
> - Eligibility for that wake is `isProspectiveBlockedTransition`, which requires `blockedTransitionAt` to be at or after `ROUTABLE_BLOCKED_ROLLOUT_AT`
> - A blocked row can carry a stamp from before that cutoff: imported historical data, or a restore from backup
> - Attaching an unblock owner to such a row achieves nothing. The cutoff rejects it, and no self-heal path repairs it, because every one of them keys off a *missing* stamp and this row has one
> - The row is therefore permanently unnotifiable while looking completely healthy
> - This pull request advances the stamp at the moment the descriptor is attached, and only then
> - The benefit is that attaching an unblock owner always reaches that owner, whatever the row's history

No existing issue. The problem is described below.

**Bug: what happened?**

An issue that is already `blocked` with a `blockedTransitionAt` earlier than
`ROUTABLE_BLOCKED_ROLLOUT_AT` does not notify its unblock owner when an
`unblockDescriptor` is attached. `isProspectiveBlockedTransition` returns false for the
historical stamp, so the edge-triggered notification path does not fire. No later write
repairs the row, because the self-heal paths only act on a row with no stamp.

**Expected behavior**

Attaching an `unblockDescriptor` to a blocked issue notifies the named owner, whenever the
underlying block started.

**Version**

Reproduced on `master`.

## What Changed

- `issueService.update()` advances `blockedTransitionAt` to the write time, and clears
  `blockedOwnerNotifiedAt`, when an `unblockDescriptor` is attached to a row that is
  already `blocked` and whose stamp is earlier than `ROUTABLE_BLOCKED_ROLLOUT_AT`
- Added a test for the pre-cutoff case, and a test pinning that a post-cutoff stamp is
  left alone

## Verification

      server/src/__tests__/attention-service.test.ts

Result on 3 September 2026: 3 files passed, 152 tests passed. Server typecheck is clean.

## Risks

- The branch fires at most once per row. After it runs, the stamp is past the cutoff and
  the condition cannot match again
- `blockedTransitionAt` is also the dependency-wake cycle key. Advancing it starts a fresh
  cycle. That is correct for a pre-cutoff stamp, because any cycle belonging to such a
  stamp predates the feature that reads the key. It is *not* correct for a post-cutoff
  stamp, which is why the branch is restricted to pre-cutoff rows and a test pins that
- Behavioural shift: owners of affected historical rows are notified once, on the next
  descriptor attach. This is the intended repair
- Related: #12734 restructures the same `if`/`else` chain in `update()`. Expect a small
  textual conflict. This pull request is independent of it and does not depend on it

## Model Used

## Checklist

- [ ] I have updated relevant documentation to reflect my changes — no user-facing documentation changes are needed
- [ ] All Paperclip CI gates are green — not known until CI runs

### DIFF
--- a/server/src/__tests__/issues-service.test.ts
+++ b/server/src/__tests__/issues-service.test.ts
@@ -1,6 +1,7 @@
 import { randomUUID } from "node:crypto";
 import { asc, eq } from "drizzle-orm";
 import { afterAll, afterEach, beforeAll, describe, expect, it } from "vitest";
+import { ROUTABLE_BLOCKED_ROLLOUT_AT } from "../services/routable-blocked.js";
 import { sql } from "drizzle-orm";
 import {
   activityLog,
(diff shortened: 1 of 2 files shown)


Sharri747:
Superseded by #12734. This change is now the third commit on that pull request (`bdd941b2d`), so it is reviewed together with the reconciler that delivers the notification it makes eligible. On its own, this branch advanced a stamp that nothing read — which is exactly the finding raised here, and it was correct. Closing; no further changes on this branch.

PR #12754, opened 2026-09-03
fix(labels): require board access to delete a label, and record what its cascade removes

Sharri747:
## Thinking Path

> - Labels group issues across a company. `issue_labels.labelId` is `ON DELETE CASCADE`
> - `DELETE /labels/:labelId` had no board-only gate, so any ordinarily-privileged caller could delete a label
> - The cascade then strips that label from every issue that carried it, company-wide, at the foreign-key level. No per-issue `issue.updated` fires, because the cascade bypasses the update path that writes that audit trail
> - The single `label.deleted` activity row named the label's own name and colour, so nothing recorded which issues lost it
> - `DELETE /agents/:id` already treats erasing corroborating history as board-only. Label deletion is the same class of act
> - This pull request gates the route to the board and records the blast radius
> - The benefit is that a company-wide, silent, unattributable data change is no longer available to a non-board caller, and is recoverable when the board does it

No existing issue. The problem is described below.

**What happened?**

`DELETE /api/labels/{labelId}` accepted any caller with ordinary company access. The delete
cascades through `issue_labels`, so one call removed the label from every issue in the
company. Because the removal happens at the foreign-key level, no `issue.updated` activity
was written for any affected issue, and the single `label.deleted` row did not name them.
The change was company-wide, silent, and could not be reconstructed afterwards.
`DELETE /api/agents/{id}` guards the same class of act with `assertBoard(req)`.

**Expected behavior**

Deleting a label requires board access, and the activity log records which issues lost the
label.

**Steps to reproduce**

1. Sign in as an agent seat with ordinary company access to a company, not as the board.
2. Create a label and apply it to two or more issues.
3. Call `DELETE /api/labels/{labelId}` with that seat.
4. The call succeeds. Every issue loses the label. The activity log holds one
   `label.deleted` row naming only the label's own name and colour, so which issues lost it
   cannot be recovered.

**Version or commit**

Reproduced on `master` at `8eaa5caa0`.

**Deployment mode**

Local development instance, embedded Postgres.

## What Changed

- `routes/issues.ts`: `assertBoard(req)` as the first line of `DELETE /labels/:labelId`,
  mirroring `DELETE /agents/:id`
- `services/issues.ts`: `deleteLabel()` runs in a transaction and reads the affected issue
  ids from `issue_labels` before the delete cascades them away, so the recorded list
  matches what the delete actually stripped rather than a snapshot a concurrent label-add
  could race past
- `routes/issues.ts`: the `label.deleted` activity details now carry `issueIds`, and the
  response body carries the same list as `affectedIssueIds`
- `routes/openapi.ts`: `DELETE /api/labels/{labelId}` and `DELETE /api/agents/{id}` are
  now declared board-only (`x-paperclip-authorization` actor `board`). Both already
  enforced this in the handler; the OpenAPI surface claimed `board_or_agent`, which made
  the gap look narrower in the documentation than it was in the code
- New regression coverage in `label-delete-board-gate-routes.test.ts`

## Verification

      server/src/__tests__/label-delete-board-gate-routes.test.ts \
      server/src/__tests__/openapi-routes.test.ts \
      server/src/__tests__/agent-cross-tenant-authz-routes.test.ts

Result on 3 September 2026: 4 files passed, 143 tests passed. Server typecheck is clean.

The new tests cover three cases: an agent seat with ordinary company access is refused
with 403 and the label and its `issue_labels` rows are untouched; a board actor's delete
cascades and the activity row enumerates exactly the issues that lost the label; a board
actor deleting a label that does not exist still returns 404 and writes no activity.

## Risks

- Breaking change for any caller that deletes labels without board access. That is the
  point of the change, but it is a behavioural break and worth calling out for anyone
  automating label cleanup from an agent seat
- The OpenAPI change corrects a declaration that was already wrong. It does not change
  what the handlers enforce
- Low risk otherwise. The transaction only adds a read before an existing delete

## Model Used

## Checklist

- [ ] I have updated relevant documentation to reflect my changes — the OpenAPI declaration is corrected in this PR; no other user-facing documentation covers this route
- [ ] All Paperclip CI gates are green — not known until CI runs

### DIFF
--- a/server/src/__tests__/label-delete-board-gate-routes.test.ts
+++ b/server/src/__tests__/label-delete-board-gate-routes.test.ts
@@ -162,6 +162,58 @@ describeEmbeddedPostgres("label delete board gate", () => {
     expect(details.issueIds.sort()).toEqual([issueOneId, issueTwoId].sort());
   });
 
+  // Review finding on this pull request: sharing a transaction with the delete
+  // does not, on its own, stop a concurrent association from being cascaded
+  // away unreported. Under READ COMMITTED an insert committed after this
+  // transaction's read snapshot is still removed by the delete. The fix locks
+  // the label row FOR UPDATE, so a concurrent insert -- which needs a
+  // FOR KEY SHARE lock on the same row -- waits, and then deletes the
+  // associations explicitly so the report is what was removed, not what was
+  // seen. This test drives a real concurrent insert against the live lock.
+  it("does not under-report when a label association is added concurrently with the delete", async () => {
+    const companyId = randomUUID();
+    const { label, issueOneId, issueTwoId, issueUnrelatedId } = await seedCompanyLabelAndIssues(companyId);
+
+    // A second connection tries to attach the label to the previously
+    // unrelated issue while the delete is in flight.
+    const contender = createDb(tempDb!.connectionString);
+    const racing = contender
+      .insert(issueLabels)
+      .values({ issueId: issueUnrelatedId, labelId: label.id })
+      .then(() => "inserted" as const)
+      .catch(() => "rejected" as const);
+
+    const app = createAppAsBoard(companyId);
+    const res = await request(app).delete(`/api/labels/${label.id}`);
+    const racedOutcome = await racing;
+
+    expect(res.status, JSON.stringify(res.body)).toBe(200);
+
+    // Whichever order the two land in, the invariant holds: every association
+    // that the delete removed is named in the report, and no association for
+    // this label survives.
+    const remainingLinks = await db.select().from(issueLabels).where(eq(issueLabels.labelId, label.id));
+    expect(remainingLinks).toHaveLength(0);
+
+    const reported: string[] = (res.body.affectedIssueIds ?? []).slice().sort();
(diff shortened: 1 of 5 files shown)


Sharri747:
One more commit, `cfbdccb1a` (fast-forward, no force-push): `syncIssueLabels` now locks the union of the issue's current and requested label rows `FOR KEY SHARE`, in id order, before it touches `issue_labels`.

Reason: `deleteLabel` on this branch takes the label row `FOR UPDATE` and then removes that label's `issue_labels` rows. `syncIssueLabels` did the reverse — child rows first, then the parent lock via the insert's implicit `FOR KEY SHARE` — so the two transactions could take the same locks in opposite order and deadlock, which Postgres resolves by aborting one of them. Both paths now take the parent lock first and serialise. A label deleted concurrently simply drops out of the lock set; its association rows are already gone by cascade. The issue row itself is already locked `FOR UPDATE` before `syncIssueLabels` runs in `update()`, so same-issue writers were serialised already.

Verification on `cfbdccb1a`: server typecheck clean; `label-delete-board-gate-routes` (4 tests, including the second-connection concurrency test) passes.

PR #12773, opened 2026-09-03
fix(server): retry cloud-tenant auth sync once on a dropped DB connection

devinfoley:
## Thinking Path

> - Managed-cloud deployments authenticate tenant requests through trusted headers. The middleware syncs the tenant's user, company, and membership rows on the way through.
> - Pooled Postgres endpoints sometimes close an established connection under an in-flight query (pooler recycle, compute suspend). The driver reconnects on the next query, but the statement on the wire fails.
> - In this path a single dropped statement fails the whole request with a 500. This happened live on 2026-09-03: the idempotent company bootstrap insert died with `write CONNECTION_CLOSED`.
> - This pull request retries the actor resolution exactly once when the error chain carries a postgres.js closed-connection code. The sync is idempotent end to end, so the replay is safe.
> - The benefit is that a routine pooler blip no longer fails an authenticated request on the entry path.

**What happened?**

A cloud tenant request hit the trusted-header authentication middleware while the pooled Postgres endpoint closed the connection mid-query. The insert failed with `write CONNECTION_CLOSED <host>:5432` wrapped in a `Failed query: insert into "companies" …` error, and the request failed.

**Expected behavior**

The driver reconnects on the next query, and every statement in the tenant sync is idempotent (upserts, on-conflict inserts, deletes; the write debounce records only after the full sync succeeds). One in-request retry should absorb the blip and serve the request. Non-transient failures must keep failing fast.

**Steps to reproduce**

1. Run an authenticated public deployment against a pooled Postgres endpoint.
2. Have the pooler close the connection while the middleware's tenant sync insert is on the wire.
3. Before this change the request fails with a 500; after it the retry serves the request.

**Deployment mode**

Authenticated public (managed cloud), external pooled PostgreSQL.

## What Changed

- `resolveCloudTenantActor` now delegates to the (unchanged) resolution body through `retryOnTransientDbConnectionError`, which retries exactly once on a transient closed-connection failure
- `isTransientDbConnectionError` walks the error `cause` chain (drizzle wraps the driver error) for the postgres.js codes `CONNECTION_CLOSED`, `CONNECTION_ENDED`, `CONNECTION_DESTROYED`; both helpers are exported for tests
- New unit test file `cloud-tenant-transient-db-retry.test.ts`: detection matrix (including a `23505` staying non-transient), retry-once-then-succeed, no-retry on non-transient, propagate-on-second-failure

## Verification

- `pnpm vitest run src/__tests__/cloud-tenant-transient-db-retry.test.ts` — 5 passed
- `pnpm vitest run src/__tests__/cloud-tenant-company-provisioning.test.ts` — 7 passed against embedded Postgres, driving the real resolution path through the new wrapper

## Risks

- Low risk. The retry is bounded to one attempt, gated on three explicit driver codes, and wraps an operation that is already idempotent by design. Every other failure propagates unchanged.
- A genuinely down database now fails after two attempts instead of one — a few milliseconds of added latency on an already-failing request.

## Model Used

Claude Fable 5 (claude-fable-5) via Claude Code, extended thinking with tool use.

## Checklist

- [x] I have searched GitHub for duplicate or related PRs and linked them above (none found for connection-retry work in this path)
- [x] I have updated relevant documentation to reflect my changes (doc comments; no user-facing docs affected)

### DIFF
--- a/server/src/__tests__/cloud-tenant-transient-db-retry.test.ts
+++ b/server/src/__tests__/cloud-tenant-transient-db-retry.test.ts
@@ -1,0 +1,68 @@
+import { describe, expect, it } from "vitest";
+
+import {
+  isTransientDbConnectionError,
+  retryOnTransientDbConnectionError,
+} from "../middleware/auth.ts";
+
+/** The shape drizzle produces: a wrapper whose `cause` is the driver error. */
+function driverClosedError(code: string): Error {
+  const driver = Object.assign(new Error(`write ${code} db.example.internal:5432`), { code });
+  return new Error("Failed query: insert into \"companies\" (…)", { cause: driver });
+}
+
+describe("isTransientDbConnectionError", () => {
+  it("detects a closed-connection code anywhere on the cause chain", () => {
+    expect(isTransientDbConnectionError(driverClosedError("CONNECTION_CLOSED"))).toBe(true);
+    expect(isTransientDbConnectionError(driverClosedError("CONNECTION_ENDED"))).toBe(true);
+    expect(isTransientDbConnectionError(driverClosedError("CONNECTION_DESTROYED"))).toBe(true);
+    const bare = Object.assign(new Error("write CONNECTION_CLOSED host:5432"), {
+      code: "CONNECTION_CLOSED",
+    });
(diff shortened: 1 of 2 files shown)


devinfoley:
@greptile review

PR #9534, opened 2026-07-14
feat(mcp): add governed tool access and app gateways

cryppadotta:
## Thinking Path

> - Paperclip is the open source app people use to manage AI agents for work.
> - Agent work increasingly depends on MCP tools, external app connections, and runtime-provided capabilities.
> - Operators need those tools to be governed by company-scoped access, approval, audit, and secret-handling boundaries rather than ad hoc adapter configuration.
> - Local adapters and runtime workspaces need a consistent way to receive approved MCP servers without leaking credentials or bypassing policy.
> - The Apps and Tools surfaces need to make connection state, test results, approval queues, and gateway setup understandable to operators.
> - This pull request adds the governed MCP tool-access foundation, app/gateway management surfaces, runtime injection plumbing, and regression coverage.
> - The benefit is safer MCP adoption: operators can connect tools, grant access, test behavior, and review ask-first actions with auditable control-plane semantics.

## Linked Issues or Issue Description

Refs #6202, #4926, #3017, #7958, #6242, #4259, #4401.

No single public issue covers the full accumulated branch, so the feature context is described here.

**Problem / Motivation**

Paperclip agents need MCP tool access that is company-scoped, auditable, and operator-governed. Existing paths leave too much behavior in per-adapter configuration or plugin-specific flows, which makes it difficult to reason about what tools an agent can use, what credentials are exposed, and which actions require approval.

**Proposed Solution**

Add a governed MCP access model across database contracts, server routes/services, runtime injection, local adapter configuration, Apps/Tools UI, and automated coverage. This includes access profiles, app connections, policy/rule surfaces, named gateways, approval/replay semantics, test-call flows, secret projection hardening, and smoke/test harness support.

**Alternatives Considered**

Keeping MCP setup as raw adapter config or plugin-only wiring was rejected because it does not provide a consistent operator-facing policy, audit, approval, and testing model across agents and companies.

**Roadmap Alignment**

Checked `ROADMAP.md` on July 14, 2026. This overlaps control-plane governance, approvals, plugin-system, and agent-runtime direction; this PR publishes the already-implemented branch for review rather than starting a separate uncoordinated feature line.

## What Changed

- Added database, shared contract, and server support for MCP tool access, connections, access profiles, app catalog/detail flows, tool policies, gateway sessions, token issuance, OAuth state binding, audit events, and approval lifecycle data.
- Added runtime and adapter plumbing so managed MCP servers and approved connection context can be injected into local agent runs while preserving company scope and secret projection boundaries.
[...]

PR #12775, opened 2026-09-03
fix(tool-gateway): make ToolGatewayHttpError extend HttpError

BastitsaB:
## Thinking Path

> - The tool gateway enforces access decisions (allowlists, ask-first policies, formal board approval) before an agent-requested tool call executes.
> - Every rejection along that path (not found, not pending, timeout, formal approval still pending, ...) is raised as a `ToolGatewayHttpError` with its own status and reason code.
> - The shared error-handler middleware only maps a thrown error to its real HTTP status when it is `instanceof HttpError`. `ToolGatewayHttpError` extends the bare `Error` class instead, so that check silently fails.
> - Every one of those expected, well-formed rejections therefore reaches the caller as an opaque `500 Internal Server Error` and gets recorded as a server crash, instead of its real status (404/409/504/...) and reason code.
> - This pull request makes `ToolGatewayHttpError` extend `HttpError`, so the existing `instanceof HttpError` branch in the error handler picks it up correctly.
> - The benefit is that tool-gateway rejections surface their real status and message to the UI/API caller, and stop polluting crash reporting with expected business-rule outcomes.

No existing issue found (searched open/closed issues and PRs for "formal board approval", "approveActionRequest", "tool-gateway" — closest related history is #9534, which introduced governed tool access).

**What happened?**

Approving a governed Composio tool-action request through `POST /issues/:id/interactions/:interactionId/accept` fails with a generic `500 Internal Server Error` whenever the action request has a linked formal-approval record that is not yet approved. The board/agent gets no indication of what is actually blocking the approval — clicking "Approve & run" repeatedly just keeps producing "Internal server error. Try again.", even immediately after a fresh confirmation card is created.

**Expected behavior**

The endpoint should return `409` with `reasonCode: "formal_approval_required"` (as `approveActionRequest` in `tool-gateway.ts` already computes), not a masked `500`.

**Steps to reproduce**

1. Trigger a destructive/governed Composio tool call from an agent run so Paperclip creates a `request_confirmation` interaction backed by a `toolActionRequests` row with `approvalId` set (i.e. it also requires a separate formal board approval).
2. As a board user, click "Approve & run" on the confirmation card (or `POST /issues/:id/interactions/:interactionId/accept`) before the linked `approvals` record itself has been approved.
3. Observe a `500 Internal Server Error` response with no `reasonCode` or detail — even though `approveActionRequest` threw a well-formed `ToolGatewayHttpError(409, ..., "formal_approval_required", { approvalId })`.

Reproduced on `master` at the base of this branch (commit prior to this fix). Root cause is structural (class hierarchy), not tied to a specific recent regression.

## What Changed

- `server/src/services/tool-gateway.ts`: `ToolGatewayHttpError` now extends `HttpError` (from `server/src/errors.ts`) instead of the bare `Error` class, and forwards `status`/`message`/`details` through `super()`. The public constructor signature (`status, message, reasonCode, details`) is unchanged, so no call site needs updating — every existing `new ToolGatewayHttpError(...)` and `instanceof ToolGatewayHttpError` check keeps working as-is.
- `server/src/__tests__/error-handler.test.ts`: added regression coverage asserting that a `ToolGatewayHttpError` maps to its own sub-500 status via the shared `errorHandler` without being attached/reported as a crash, and that it is still reported correctly when its own status is `>= 500`.

## Verification

- `pnpm --filter @paperclipai/server exec vitest run src/__tests__/error-handler.test.ts` — 7 passed (5 pre-existing + 2 new).
- `pnpm --filter @paperclipai/server exec vitest run src/__tests__/tool-gateway.test.ts src/__tests__/tool-gateway-service.test.ts` — 83 passed, unaffected by the class-hierarchy change.
- `pnpm --filter @paperclipai/server exec vitest run src/__tests__/issue-thread-interaction-routes.test.ts src/__tests__/tool-access-service.test.ts src/__tests__/tool-access-policy-service.test.ts src/routes/tool-access-connection-intent.test.ts` — 337 passed.
- `tsc --noEmit` on `server`: identical error count and diff before/after the change (208 pre-existing, environment-only errors unrelated to this file — confirmed via `git stash`/`git stash pop` comparison); zero new errors introduced.

## Risks

- Low risk. This is a class-hierarchy change with an unchanged constructor signature; every existing throw site and `instanceof ToolGatewayHttpError` check keeps compiling and behaving the same. The only externally visible change is that HTTP responses now carry the *correct* status/message/details instead of a masked 500 — call sites that already special-case `ToolGatewayHttpError` locally (e.g. `sendGatewayError` in `routes/tool-gateway.ts`) are unaffected, since they never relied on the global handler.
- The one behavioral change worth flagging: statuses `< 500` (e.g. 404/409) no longer get reported to Sentry/telemetry as crashes and no longer attach `res.err`/`res.__errorContext`. This is intentional and matches how every other `HttpError` subtype is already handled — expected business-rule rejections should not be crash noise.

## Model Used

Claude (Anthropic), Sonnet 5 (`claude-sonnet-5`), via Claude Code. Extended reasoning, tool use (repo search, code reading, local build/test execution), and code execution were used to locate the root cause, implement the fix, and run the verification above.

## Checklist

- [ ] I have updated relevant documentation to reflect my changes — no documentation describes this internal error-class hierarchy, none needed updating
- [ ] All Paperclip CI gates are green — pending CI run on this PR
- [ ] Greptile is 5/5 with no open P2s, recommendations, or follow-ups — pending automated review

### DIFF
--- a/server/src/__tests__/error-handler.test.ts
+++ b/server/src/__tests__/error-handler.test.ts
@@ -169,6 +169,7 @@ describe("errorHandler", () => {
     expect(res.status).toHaveBeenCalledWith(409);
     expect(res.json).toHaveBeenCalledWith({
       error: "Tool action request requires formal board approval before execution",
+      reasonCode: "formal_approval_required",
       details: { approvalId: "approval-1" },
     });
     // Sub-500 statuses must not be attached/reported as a server crash.
--- a/server/src/__tests__/error-handler.test.ts
+++ b/server/src/__tests__/error-handler.test.ts
@@ -2,6 +2,7 @@ import type { NextFunction, Request, Response } from "express";
 import { beforeEach, describe, expect, it, vi } from "vitest";
 import { HttpError } from "../errors.js";
 import { errorHandler } from "../middleware/error-handler.js";
+import { ToolGatewayHttpError } from "../services/tool-gateway.js";
 
 const recordResponsibleUserDenialOnActiveRunMock = vi.hoisted(() => vi.fn());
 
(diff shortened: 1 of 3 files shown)


BastitsaB:
Addressed in ed23290: the shared error handler now also surfaces `reasonCode` on the JSON response (duck-typed off any `HttpError`), matching the `{ error, reasonCode, ...details }` shape every route that already special-cases `ToolGatewayHttpError` locally uses (`routes/tool-gateway.ts`, `routes/tool-access.ts`, `routes/plugins.ts`). Extended the regression test to assert it. All 427 previously-verified tests still pass; `tsc --noEmit` error count unchanged (208, pre-existing/environmental, confirmed via `git stash` diff).

PR #12795, opened 2026-09-03
ci(runner): inspect Daytona image metadata remotely

cryppadotta:
## Thinking Path
The reused Daytona image path already verifies the signed immutable digest. It then downloads every filesystem layer only to read OCI config fields. Buildx can retrieve the same config from that immutable digest without pulling the layers. The assertions can therefore stay intact while removing the expensive transfer.

## What Changed
- inspect the signed immutable Daytona image config through Buildx after GHCR logout
- preserve digest, source revision, content ID, platform, user, and provider-pack assertions
- extend the workflow contract test for the metadata-only path

## Verification
- Daytona image and workflow security tests: 10 passed
- Prettier and git diff checks passed
- observed full pull/prune cost: about 4m55s; metadata inspection: about one second

## Risks
The current image has one runnable linux/amd64 platform plus its attestation. A future genuinely multi-platform image would need explicit linux/amd64 selection.

## Model Used
Codex (GPT-5)

### DIFF
--- a/.github/workflows/runner-full-stack-e2e.yml
+++ b/.github/workflows/runner-full-stack-e2e.yml
@@ -418,31 +418,25 @@ jobs:
             --certificate-oidc-issuer https://token.actions.githubusercontent.com \
             "$IMAGE_TAG@$digest" >/dev/null
           immutable="${IMAGE_TAG%:*}@$digest"
-          # The Daytona base image is large. The build cache plus a second full
-          # anonymous pull can exhaust a standard GitHub-hosted runner before
-          # Docker creates the tiny metadata-probe container. The pushed digest
-          # is already immutable, so release the local builder/cache first.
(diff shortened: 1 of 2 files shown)


cryppadotta:
@greptile-apps @greptileai Please re-review exact head 6d8bd0e04e8a2a70f86207c569be702efc9b59e6 with runtime evidence from the actual signed immutable image used by run 33811017576:

`docker buildx imagetools inspect <digest> --format {{json .Image}}` returns a direct `v1.Image` object with keys `architecture, config, created, history, os, rootfs`; `architecture=amd64`, `os=linux`, and `config.User=daytona`. The suggested `index .Image "linux/amd64"` form was also executed and fails: `error calling index: can't index item of type v1.Image`. The PR body now has all required sections. Please reassess the platform-map finding against this observed Buildx type.

PR #12798, opened 2026-09-04
test(shared): make the worktree port registry lock suite deterministic

nickyleach:
## Thinking Path

> - Shared worktree services use lock leases and worker-thread heartbeats
> - The lock test suite measured wall-clock timing across two threads
> - Processor contention allowed a heartbeat tick to change the value during an assertion
> - This pull request removes that timing race and restores a regression guard
> - The benefit is a stable test suite that still detects slow heartbeats

**What happened?** The worktree port registry lock suite failed at random under continuous-integration processor contention. The failure reported a fresh timestamp where the test expected an old timestamp.

**Expected behavior** The suite must pass when the heartbeat runs at its supported interval. It must also fail when the heartbeat interval regresses.

**Steps to reproduce**
1. Run `npx vitest run src/worktree-port-registry.test.ts` in `packages/shared`.
2. Repeat the run under bounded processor contention.
3. Set the heartbeat interval to 3000 ms and run the asynchronous critical-section test.

**Paperclip version or commit** `a661caf74e704f7700a8b8a1e79b76ebd04e3483`

**Deployment mode** Built from source.

**Installation method** Built from source.

**Agent adapter(s) involved** Not adapter-specific (core test).

**Database mode** Not database-related.

**Additional context** Related open pull requests are #11994, #11985, and #11922. This pull request keeps all five tests active and does not use `skip`, `skipIf`, or `todo`.

## What Changed

- Build the fallback-probe lock state by hand so no live heartbeat changes the timestamp during the assertion.
- Count distinct heartbeat refreshes in the asynchronous critical-section test.
- Close the fake probe and settle the pending lock attempt in a `finally` block.
- Keep production code unchanged.

## Verification

- `npx vitest run src/worktree-port-registry.test.ts` — 5 of 5 tests pass.
- `npx vitest run` — 72 files and 704 tests pass at submit time.
- `npx tsc --noEmit` — exit code 0.
- Ten target-file runs pass under bounded processor contention.
- A 3000 ms heartbeat interval fails with `expected 2 to be greater than or equal to 3`.
- An inverted cleanup assertion exits normally in 379 ms without a leaked worker.

## Risks

Low risk. This pull request changes one test file. It changes test setup and assertions only.

## Model Used

OpenAI GPT-5 through Codex. Exact model ID: GPT-5. The model used tool calls and code execution. The context window is not disclosed by the runtime.

## Checklist

### DIFF
--- a/packages/shared/src/worktree-port-registry.test.ts
+++ b/packages/shared/src/worktree-port-registry.test.ts
@@ -2,6 +2,7 @@ import fs from "node:fs";
 import os from "node:os";
 import path from "node:path";
 import { setTimeout as delay } from "node:timers/promises";
+import { Worker } from "node:worker_threads";
 import { afterEach, describe, expect, it } from "vitest";
 import {
   withWorktreePortRegistryLock,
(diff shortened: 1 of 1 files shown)


conreo:
Evidence for this fix, measured on `master` at `e1f245a6` with no change applied (Node.js `v22.22.2`, Linux, x86-64).

| Runs | Failures | Rate |
|---|---|---|
| 25 | 4 | 16% |
| 40 | 4 | 10% |

Every failure is the same one:

```
 FAIL  |@paperclipai/shared| src/worktree-port-registry.test.ts > worktree port registry lock > does not reclaim a stale lock while its fallback ownership probe responds
AssertionError: expected 0 to be greater than 5000
 ❯ src/worktree-port-registry.test.ts:56:56
```

Reproduce it from the repository root:

```sh
for i in $(seq 1 25); do
  node node_modules/vitest/vitest.mjs run packages/shared/src/worktree-port-registry.test.ts || echo "run $i failed"
done
```

The same assertion also stopped a real lane. `ci / General tests (workspaces-b)` failed on it in #13306, and the same file passed on the rerun of that lane.

The cause is the one this pull request addresses: the test writes a stale mtime and reads it back, while the heartbeat worker refreshes that same value. `touchOwnedLock` refreshes whenever either owner file carries its token, and the backup owner file keeps the token because the test changes only `processIdentity`. The result depends on which thread reaches the directory first, so the assertion fails whenever a refresh lands after the write.

Two things I confirmed while trying the smaller fix, for whoever reviews this one:

- Serving the probe from the test is the only way to make the assertion mean something. A readable owner record for the prober and a durably stale mtime cannot coexist while the live heartbeat runs, because the heartbeat refreshes the mtime from the same record the prober reads.
- The probe server must run on a worker thread. `probeRegistryLockOwner` blocks the calling thread in `Atomics.wait` until the probe answers, so a same-thread server cannot answer its own caller.

I closed my own smaller attempt (#13526) in favour of this one, because it could not assert that the probe path ran.

PR #12809, opened 2026-09-04
feat(telemetry): add the agent.task_run event and emit it at every terminal run transition

nickyleach:
## Thinking Path

> - Paperclip records agent run outcomes through telemetry and run lifecycle services
> - Terminal run transitions need one consistent event for outcome analysis
> - The current paths do not report every terminal transition through one event
> - This pull request adds the agent.task_run event and emits it at each terminal transition
> - The benefit is complete run outcome data without exposing raw task identifiers

Paperclip telemetry reports agent activity, but it does not report every terminal task run through one event.

**Subsystem affected**

Cross-cutting (multiple of the above): packages/shared telemetry and server run lifecycle services.

**Current behavior**

Several run paths write a terminal status without a matching agent.task_run telemetry event.

**Proposed behavior**

Each terminal run transition emits one agent.task_run event. The event records the terminal state and uses the existing pseudonym helper for the optional task identifier.

**Reason and benefit**

Complete terminal-run data helps operators measure agent outcomes. The pseudonym helper prevents the raw task identifier from leaving the installation.

**Breaking changes**

None. The change adds an event and keeps existing event behavior compatible.

## What Changed

- Add the agent.task_run telemetry contract and client helper.
- Reuse the existing pseudonym helper for the task identifier. The helper hashes the identifier with a per-installation salt and returns 16 hexadecimal characters. The raw identifier never leaves the installation. Existing identifiers do not move.
- Emit one event from each legacy, native, recovery, and issue terminal transition.
- Keep emissions outside database transactions and make delivery best-effort.
- Add regression tests for event shape, hashing, terminal transitions, and emission failures.
- Document the event and its privacy rule in the telemetry data contract.

## Verification

- `npx tsc --noEmit` in `server/` passes at the submitted commit.
- The pull-request CI suite must pass. CI is the authority because local Vitest has a known dependency artifact.
- The added regression tests cover event output shape, per-installation hash divergence, raw identifier handoff, omitted identifiers, and non-throwing emits.

## Risks

- A missed terminal path could reduce event coverage.
- Telemetry delivery remains best-effort and cannot change run finalization.
- The pseudonym helper uses installation-specific state, so identifiers differ between installations.

## Model Used

OpenAI Codex, GPT-5, tool use and code execution. Context window details were not provided.

## Checklist

- [x] I have addressed all Greptile and reviewer comments before requesting merge

### DIFF
--- a/packages/shared/src/telemetry/client.ts
+++ b/packages/shared/src/telemetry/client.ts
@@ -1,4 +1,4 @@
-import { createHash } from "node:crypto";
+import { createHash, createHmac } from "node:crypto";
 import type {
   TelemetryConfig,
   TelemetryDimensions,
@@ -64,6 +64,15 @@ type TrackArgs<K extends TelemetryEventName> =
 // Length of the truncated hex `batchId`. 32 hex chars = 128 bits — the
 // collision-safe floor. Do not lower below 32.
 const BATCH_ID_HEX_LENGTH = 32;
+
+// Domain-separation label for `hashTaskId`. Bump the version segment when
+// the derivation changes. This keeps an old hash distinct from a new one.
+// The label carries the version. The emitted hash carries no prefix,
+// because the backend schema pattern accepts exactly 32 lowercase hex
+// characters.
+const TASK_ID_HASH_LABEL = "paperclip.telemetry.task_id.v1";
+// Length of the truncated hex task-id hash. 32 hex characters equal 128 bits.
+const TASK_ID_HASH_HEX_LENGTH = 32;
 
 /**
  * A chunk awaiting retry. `events` + `batchId` are frozen at first send and
@@ -469,6 +478,22 @@ export class TelemetryClient {
       .update(state.salt + value)
       .digest("hex")
       .slice(0, 16);
+  }
+
+  /**
+   * Derives a stable, non-reversible id for a task from its raw id. The
+   * method uses the per-installation secret as an HMAC key. This differs
+   * from `hashPrivateRef`, which uses a salted hash instead of a keyed
+   * HMAC. The keyed HMAC design stops two installations from deriving the
+   * same value for the same raw id. Call this method before you send a
+   * task id. Never send the raw task id.
+   */
+  hashTaskId(taskId: string): string {
+    const state = this.getState();
+    return createHmac("sha256", state.salt)
+      .update(`${TASK_ID_HASH_LABEL}:${taskId}`)
+      .digest("hex")
+      .slice(0, TASK_ID_HASH_HEX_LENGTH);
   }
 
   private getState(): TelemetryState {
--- a/packages/shared/src/telemetry/client.ts
+++ b/packages/shared/src/telemetry/client.ts
@@ -1,4 +1,4 @@
-import { createHash, createHmac } from "node:crypto";
+import { createHash } from "node:crypto";
 import type {
   TelemetryConfig,
   TelemetryDimensions,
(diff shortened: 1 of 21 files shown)


packages/shared/src/telemetry/client.ts
```diff
@@ -471,6 +480,22 @@ export class TelemetryClient {
...
+   * task id. Never send the raw task id.
+   */
+  hashTaskId(taskId: string): string {
+    const state = this.getState();
+    return createHmac("sha256", state.salt)
+      .update(`${TASK_ID_HASH_LABEL}:${taskId}`)
+      .digest("hex")
+      .slice(0, TASK_ID_HASH_HEX_LENGTH);
+  }
+
```
nickyleach: We don't need this. We get nothing from HMAC (no recipient can verify the sender), and the extra 64bits of entropy aren't important at the expected data scale. We should use `hashPrivateRef` instead.

server/src/services/agent-task-run-telemetry.ts
```diff
@@ -0,0 +1,124 @@
...
+ * wrote to a terminal status. Call this beside each terminal status write in
+ * the legacy run engine, after the write commits. Never call this inside a
+ * database transaction — the write must commit first.
+ *
+ * This function never throws. A telemetry failure must never fail a run.
+ * Telemetry enrichment does an awaited database lookup, so treat this call
+ * as best-effort background work: do not let a caller's own required
+ * lifecycle work (publishing a live event, cancelling a wake, clearing an
+ * issue lock, returning a response) wait behind it. Fire this call with
+ * `void emitAgentTaskRun(...).catch(() => {})` and never await it, even
```
nickyleach: Drop the `void emitAgentTaskRun(...).catch(() => {})`. You've already said the function never throws, so the `.catch(() -> {})` is unnecessary. All of the call sites which use `.catch(() -> {})` need to be updated.

server/src/services/agent-task-run-telemetry.ts
```diff
@@ -0,0 +1,124 @@
...
+  const startedAtMs = new Date(run.startedAt).getTime();
+  const endMs = run.finishedAt ? new Date(run.finishedAt).getTime() : now.getTime();
+  return Math.max(0, Math.round((endMs - startedAtMs) / 1000));
+}
+
+/**
+ * Emits one `agent.task_run` telemetry event for a run that a caller already
+ * wrote to a terminal status. Call this beside each terminal status write in
+ * the legacy run engine, after the write commits. Never call this inside a
+ * database transaction — the write must commit first.
```
nickyleach: Tighten up this comment. Two issues:
1. Drop the "in the legacy run engine", we're also writing in the native runtime
2. Drop the "never ... inside a database transaction", just state it should be called once the run is committed

server/src/services/agent-task-run-telemetry.ts
```diff
@@ -0,0 +1,124 @@
...
+ * Telemetry enrichment does an awaited database lookup, so treat this call
+ * as best-effort background work: do not let a caller's own required
+ * lifecycle work (publishing a live event, cancelling a wake, clearing an
+ * issue lock, returning a response) wait behind it. Fire this call with
+ * `void emitAgentTaskRun(...).catch(() => {})` and never await it, even
+ * alongside other required work with `Promise.all` — an awaited
+ * `Promise.all` still blocks the caller's return on the slower of the two
+ * promises, so it does not remove the delay.
+ */
+export async function emitAgentTaskRun(db: Db, run: HeartbeatRun): Promise<void> {
```
nickyleach: Consider changing this to `export function emitAgentTaskRun(db: Db, run: HeartbeatRun): void` and discarding the `Promise<void>` in this method instead

server/src/services/agent-task-run-telemetry.ts
nickyleach: Applied. All eight call sites now use `void emitAgentTaskRun(...)` without `.catch(() => {})`.

packages/shared/src/telemetry/client.ts
nickyleach: Applied. Both event helpers now call `client.hashPrivateRef(dims.taskId)`. The custom HMAC helper and its constants are removed.

server/src/services/agent-task-run-telemetry.ts
nickyleach: Applied. The comment now states that callers must invoke the helper after the run commits, for both legacy and native runtimes.

PR #12821, opened 2026-09-04
ci(docker): build each architecture on a native runner instead of QEMU

tonio-alucema:
## Thinking Path

A fleet rollout failed with `image_manifest_not_found` for a commit that had merged and gone green. Tracing that back: the fleet resolves releases against the `-cloud` image, that image comes from `docker.yml`, and `docker.yml`'s runs on master have been reading `cancelled` for a long stretch. The cloud half was fine; the production half was hanging and taking the run down with it — and, because a run holds the concurrency slot for its whole duration, starving later commits of a build at all.

No tracking issue — described inline, per CONTRIBUTING.md.

**What's wrong.** `build-and-push` builds `linux/amd64,linux/arm64` on an x86 runner, so arm64 runs under QEMU. It hangs there — deterministically, in the same step:

```
#111 [linux/arm64 build  8/10] RUN pnpm --filter @paperclipai/server build
```

…then emits nothing until `timeout-minutes: 60` kills it. Three consecutive runs on 2026-09-04, silent for **38, 43 and 45 minutes** respectively. The amd64 leg reached `production 5/5` minutes earlier in every one.

**Why it stayed hidden.** A timed-out job is reported by GitHub as **cancelled, not failed**. The run conclusion reads "cancelled", which looks like supersession rather than breakage, so the production image quietly stopped publishing.

**The knock-on.** A run that burns the full hour holds the top-level concurrency slot for that hour. `cancel-in-progress: false` keeps exactly one pending slot, so merges arriving faster than one an hour supersede each other while queued. Sampling the last ten master commits, **five produced no image at all** — their Docker runs have zero job records because they never started.

**Expected.** Both architectures publish, and a commit merged during a busy period still gets an image.

## What Changed

`build-and-push` becomes a two-leg matrix, each on a runner of its own architecture:

| platform | runner |
|---|---|
| `linux/amd64` | `ubuntu-latest` |
| `linux/arm64` | `ubuntu-24.04-arm` |

Each leg pushes **by digest** (`push-by-digest=true`, untagged), and a new `merge-and-push` job names the digests into one manifest list with the real lane tags. Nothing is publicly tagged until the merge, so a half-published multi-arch image is never a pullable state.

Two supporting changes:

- **Per-arch BuildKit cache refs** (`:buildcache-amd64` / `:buildcache-arm64`). Separate runners sharing one ref would overwrite each other on every build.
- **The PID-1 orphan-reaping check moves to the merge job**, since that is where a tagged, pullable image first exists. It still runs against the pushed image rather than a local build, for the same reason as before.

**arm64 is kept, not dropped.** The cloud variant is amd64-only and can be — managed hosts are amd64. This is the self-hosted image and ARM hosts consume it, so dropping arm64 would break them. GitHub-hosted arm64 runners are free for public repositories, which this is.

`build-and-push-cloud` is untouched. It was already `platforms: linux/amd64` and has been succeeding in ~14 minutes throughout — that is why `-cloud` images exist at all.

## Verification

Parsed the workflow and asserted its shape (jobs, matrix, `needs`, step order, that the cloud job is unchanged). The artifact actions are pinned by SHA with version comments, matching the repo's dominant convention — `upload-artifact` v7 and `download-artifact` v8, the same pins used across the other workflows; v8 is required for the `pattern` / `merge-multiple` inputs the merge job uses.

**This PR's CI does not exercise the change.** `docker.yml` triggers on master and tag pushes, never on pull requests — deliberately, since it publishes release images. The first real run is after merge, so the check is: the next master push produces a `Docker` run whose `build-and-push (amd64)`, `build-and-push (arm64)` and `merge-and-push` jobs all succeed, and whose conclusion is `success` rather than `cancelled`.

## Risks

- **Not testable before merge**, per above. If the matrix is wrong the next master push fails loudly rather than silently — which is already better than the current state, where the failure mode is an invisible "cancelled".
- **First use of `ubuntu-24.04-arm` in this repo.** No other workflow uses an ARM runner. They are free for public repos, but if the label is unavailable the arm64 leg will fail to schedule and the merge will not run — no image, same as today, and visible.
- **Digest-push changes the publish shape.** Between the legs finishing and the merge running, digests exist untagged in ghcr. Anything watching for tags sees no intermediate state; anything enumerating untagged manifests will see more of them.
- **Cache refs change name**, so the first build after this lands is cold on both legs and will be slower than steady state.
- **Does not fix the underlying QEMU hang** — it avoids it. If arm64 ever has to build under emulation again, the same stall is presumably still there.
- No application code, schema, server or persistence change.

## Model Used

Anthropic Claude — Opus 5, model ID `claude-opus-5`, run through Claude Code.

Extended thinking enabled. Tool use throughout: GitHub Actions API to correlate run/job outcomes and read build logs, `git` for ancestry checks, and a YAML parser to validate the rewritten workflow's structure.

## Checklist

A workflow change has no unit test to add, and `docker.yml` cannot run on a PR; the verification section states what to check on the first master run instead.

### DIFF
--- a/.github/workflows/docker.yml
+++ b/.github/workflows/docker.yml
@@ -14,9 +14,10 @@ on:
   # way.
   workflow_dispatch:
 
-permissions:
-  contents: read
-  packages: write
+# Least privilege: nothing at the workflow level; each job declares exactly
+# the token scopes it uses (checkout needs contents:read, GHCR pushes need
+# packages:write).
+permissions: {}
 
 # Serialise builds per ref without killing an in-flight one: a newer push
 # supersedes only the pending slot, so the image build that is already
@@ -74,6 +75,9 @@ jobs:
     # (The interim fix while this PR landed raised the single QEMU job's cap
     # to 120 minutes; native per-arch legs make that headroom unnecessary.)
     timeout-minutes: 60
+    permissions:
+      contents: read
+      packages: write
     steps:
       - name: Checkout
         uses: actions/checkout@v7
@@ -272,11 +276,14 @@ jobs:
     needs: build-and-push
     runs-on: ubuntu-latest
     timeout-minutes: 20
+    permissions:
+      contents: read
+      packages: write
     steps:
       # Checked out for `packages/db` (schema labels) and the orphan-reaping
       # script the verification step pipes in.
       - name: Checkout
-        uses: actions/checkout@v7
+        uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7
         with:
           fetch-depth: 0
 
@@ -288,14 +295,14 @@ jobs:
           merge-multiple: true
 
       - name: Login to GitHub Container Registry
-        uses: docker/login-action@v4
+        uses: docker/login-action@dbcb813823bdd20940b903addbd779551569679f # v4
         with:
           registry: ghcr.io
           username: ${{ github.repository_owner }}
           password: ${{ secrets.GITHUB_TOKEN }}
 
       - name: Set up Docker Buildx
-        uses: docker/setup-buildx-action@v4
+        uses: docker/setup-buildx-action@37fe631027851001ddb9b187196cc803df7f5f0e # v4
 
       # Repeated from the build job rather than passed between them: job
       # outputs would have to survive a matrix, and this is two `ls` calls.
@@ -315,7 +322,7 @@ jobs:
       # so parallel canary-tag builds can never race the channel backwards.
       - name: Docker meta
         id: meta
-        uses: docker/metadata-action@v6
+        uses: docker/metadata-action@dc802804100637a589fabce1cb79ff13a1411302 # v6
         with:
           images: ghcr.io/${{ github.repository }}
           tags: |
@@ -366,6 +373,9 @@ jobs:
   build-and-push-cloud:
     runs-on: ubuntu-latest
     timeout-minutes: 60
+    permissions:
+      contents: read
+      packages: write
     steps:
       - name: Checkout
         uses: actions/checkout@v7
--- a/.github/workflows/docker.yml
+++ b/.github/workflows/docker.yml
@@ -32,16 +32,48 @@ concurrency:
   cancel-in-progress: false
 
 jobs:
+  # Multi-arch by native runner, not QEMU.
+  #
+  # This was one job building linux/amd64,linux/arm64 together on an x86
+  # runner. The arm64 half is emulated there, and it did not merely run slow:
+  # it wedged, every time, in `RUN pnpm --filter @paperclipai/server build`,
+  # emitting nothing for 38-45 minutes until `timeout-minutes: 60` killed the
+  # job. Verified across three consecutive runs on 2026-09-04; the amd64 half
+  # reached `production 5/5` minutes earlier in every one.
+  #
+  # A timed-out job is reported as *cancelled*, not failed, so the run read
+  # "cancelled" and the production image simply stopped publishing without
+  # anything going red in an obvious way.
+  #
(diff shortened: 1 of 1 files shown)


tonio-alucema:
New evidence for this one, from a different direction: the QEMU stall is now breaking the **staging auto-roll**, not just per-commit image coverage.

`paperclip-cloud`'s hourly staging roll resolves the newest canary and needs a published cloud image for it. Run [33934588203](https://github.com/paperclipai/paperclip-cloud/actions/runs/33934588203) walked backwards through six consecutive canaries and gave up:

```
canary 2026.904.0-canary.15 (5b56d430e) has no published cloud image; trying an older canary
... .14, .13, .12, .11, .10 - all the same ...
fleet app rollout FAILED: no deployable canary: 2026.904.0-canary.15, .14, .13, .12, .11, .10
all lack a published cloud image - upstream Docker publishing is behind or broken
```

The chain is one step longer than it looks, and it explains why `build-and-push-cloud` is collateral even though it is not the job that hangs:

1. `build-and-push` stalls in `[linux/arm64 build 8/10] RUN pnpm --filter @paperclipai/server build` under QEMU and sits until the 60-minute timeout — reported as *cancelled*, not failed, which is why it reads as noise in the run list.
2. `docker.yml` uses `group: docker-${{ github.ref }}` with `cancel-in-progress: false`, so that run holds the slot for the full hour.
3. `build-and-push-cloud` is in the same workflow, so it is in the same run and the same concurrency group. Commits pushed during the stall never start their Docker run at all, and the pending slot keeps only the newest — so the ones in between are skipped entirely and get no cloud image.

The file's own comment says the design intent is that "the image build that is already running always finishes and publishes". That holds when the build finishes. Under the QEMU stall it inverts: the slot is occupied by a build that will never publish, and everything behind it is dropped.

So this is not only CI hygiene — it is why staging misses commits. Native ARM runners remove the stall, and with it the hour-long slot occupancy.

Still blocked on a code-owner review (@cryppadotta @devinfoley @nickyleach @forgottendev) — a look would be appreciated.

For completeness: one other recent staging roll failed for an unrelated reason (`resolving upstream tag canary/v2026.904.0-canary.14 failed: HTTP 422`, a git-tag/npm publish mismatch), which this PR would not address.

devinfoley:
Rebased onto current master and reconciled with what landed since this PR was opened, then addressed both security-scan findings. The original commit and its authorship are preserved; the reconciliation lives in the conflict resolution plus one follow-up commit.

**Reconciliation with master (inside the rebased commit):**
- Dropped the merge job's `type=raw,value=canary` tag line. Since #12950, the mutable `:canary` channel tags are moved only by the dist-tag-checked `promote_canary_channel` job, so parallel canary-tag builds can never race the channel backwards — the merge job now carries exactly the build job's lane mapping, with a comment saying why.
- `promote_canary_channel` now `needs: [merge-and-push, build-and-push-cloud]`: the per-arch legs push untagged digests, and the `sha-*` production tags the promotion retags only exist after the manifest merge.
- The interim timeout bump (#13114, 60→120 on the old single QEMU job) is superseded: native legs return to `timeout-minutes: 60` as a backstop, noted in the job comment.

**Security-scan findings (follow-up commit):**
- The new merge-and-push job's third-party actions are pinned to full commit SHAs with version comments (checkout, login-action, setup-buildx-action, metadata-action), matching the artifact-action pins the job already used.
- Token permissions are least-privilege: `permissions: {}` at the workflow level, each job declaring exactly `contents: read` + `packages: write`.

`actionlint` is clean apart from the pre-existing SC2012 info notes in untouched steps. As the PR body says, docker.yml never runs on PRs — the real verification is the first master push after merge producing a run where both legs, the merge, and (on canary tags) the promotion succeed.

@greptileai

PR #12844, opened 2026-09-04
fix(watchdogs): do not wake a task watchdog the platform will refuse

simberthon:
## Thinking Path

> - Task watchdogs are the subsystem that notices a watched issue subtree has stopped, then wakes a watchdog agent to repair it
> - Two parts of that subsystem disagree. `evaluateWatchdog` decides to wake, and `revalidateMutationScope` decides whether the woken run may write. They read liveness at different moments, so a watchdog can be woken on a stop it is then forbidden to touch
> - The gap is not theoretical. On a live instance a board comment enqueued a run on the watched issue 13 ms before the watchdog wake was enqueued from a snapshot taken before that run existed. The watchdog ran for about nine minutes and every write came back "review is stale"
> - It needs to be addressed because that run is a guaranteed no-op that still costs a full agent run, and because the refusal message names no issue and no signal, so the woken agent cannot check the verdict against anything
> - This pull request re-reads liveness before the wake is enqueued and skips the trigger when the subtree went live, and it makes the refusal name which issue is live and what makes it live
> - The benefit is that no agent run is spent on a repair the platform will refuse, and a refusal that does happen is actionable instead of opaque

No public issue exists for this. Description follows `.github/ISSUE_TEMPLATE/bug_report.yml`.

**What happened?**

A task watchdog is woken with `reason: task_watchdog_stopped_subtree` on a stopped subtree. Every mutation it then attempts on the watched issue is refused with `Task-watchdog review is stale because the watched subtree now has a live, waiting, already-reviewed, or not-applicable path`, carrying `currentState: "live"`. The wake and the refusal are two opposite verdicts on the same source state, minutes apart. The refused run cannot verify the claim on its own: `GET /api/live-runs` is scoped to the caller, so another agent's run on the watched issue is invisible from the watchdog run.

**Expected behavior**

A watchdog is woken only for a state it is allowed to act on. When a repair is refused anyway, the error names the issue and the signal that make the subtree live.

**Steps to reproduce**

1. Arm a task watchdog on an issue whose subtree is stopped.
2. Post a board comment on the watched issue, which enqueues an `issue_commented` wake for its assignee.
3. Let the watchdog reconcile pass run in the same window. It reads liveness through `collectClassifierInput`, then performs several writes before it enqueues the wake.
4. The run started in step 2 turns the subtree live before the watchdog run begins.
5. The watchdog run starts and every write it attempts is refused for as long as that run holds the issue.

Reproduced against `master` at the merge base of this branch. Covered by new deterministic tests in `server/src/__tests__/task-watchdogs-scheduler.test.ts`.

**Agent adapter(s) involved**

Not adapter-specific (core bug). The trigger path is server side.

**Additional context**

Two other readings of the symptom were plausible and both are false, so both are now pinned by tests. Pending interactions on children do not make a parent read as live: waits are material state that feeds the stop fingerprint, and only a run or a queued wake yields `live`. A recent agent comment on the watched issue does not make it live either.

## What Changed

- `classifyTaskWatchdogSubtree` now attaches `liveSignals` to a `live` result: for each included issue, which issue id and identifier carries which signal kind (`active_run` or `queued_wake_request`) in which status.
- `revalidateMutationScope` names those signals in the refusal instead of listing four possible causes.
- `evaluateWatchdog` re-reads liveness before it enqueues the wake, and skips the trigger when the subtree went live or moved to a different stop fingerprint. The recheck runs twice: once before the reusable review issue is created or reopened, so the common case leaves no artifact behind, and once immediately before the enqueue.
- When the second recheck loses, the review issue that was already published is re-stamped with a `task_watchdog_untriggered:` fingerprint and an `issue.task_watchdog_trigger_abandoned` activity entry. Without that, the row would read as an open same-fingerprint review on every later pass and the stopped state would never get its wake. `origin_fingerprint` is `NOT NULL`, so the pin is replaced rather than cleared.
- Tests: a deterministic scheduler test that makes the source acquire a run at the moment the review issue is written, then asserts no wake was enqueued, the artifact is not left pinned, and the next pass triggers once the racing run ends. Plus a classifier test pinning that pending child interactions never make a parent live.

## Verification

- `cd server && npx vitest run src/__tests__/task-watchdogs-scheduler.test.ts` — 19 passed.
- `cd server && npx vitest run src/__tests__/task-watchdogs-classifier.test.ts` — 21 passed.
- `cd server && npx vitest run src/__tests__/task-watchdogs-incident-fixture.test.ts src/__tests__/issue-watchdogs-routes.test.ts src/__tests__/issue-agent-mutation-ownership-routes.test.ts` — 100 passed.
- `cd server && npx tsc --noEmit -p tsconfig.json` — no diagnostic in either changed file. The run does report pre-existing errors elsewhere from unbuilt workspace dependencies (`@paperclipai/plugin-sdk`, `@paperclipai/paperclip-runner`).
- The race is reproduced with a temporary Postgres `AFTER INSERT` trigger on `issues`, so the interleave is deterministic rather than timing dependent. The trigger and its function are dropped in a `finally`.

## Risks

- No migration and no schema change. No API shape is removed; `liveSignals` is an added field on the `live` classification.
- Behavioral shift: a reconcile pass can now decide not to trigger where it previously did. That is the intent, and the state it declines is picked up by the next pass. The new scheduler test asserts that retry, because the failure mode of getting it wrong is a watchdog that never fires again for that fingerprint.
- The window is narrowed, not closed. Nothing here is atomic with the enqueue, so a run that starts between the last recheck and the wake insert still produces a refused watchdog. Closing it fully needs the wake insert to carry the liveness predicate, which is a larger change to the wake enqueue contract. The refusal diagnostics in this PR are what make the residual case survivable rather than opaque.
- `task_watchdog_untriggered:` is a new sentinel value in `issues.origin_fingerprint`. It is deliberately outside the `task_watchdog_stop:` prefix that `normalizeStopFingerprint` accepts, so no existing reader treats it as a stop fingerprint.

## Model Used

Claude Opus 5 (`claude-opus-5`, 1M context window), extended thinking, with tool use and code execution. Both commits on this branch were produced with that model.

## Checklist

Two boxes are left unchecked on purpose.

The branch name carries an instance-local ticket id. It was pushed before that rule was read, and renaming the head branch of an open PR would drop the review history on it, so it is left as is and called out here rather than hidden.

`ci / Build` is red on a Rust test in the runner crate, `acpx_provider_backend::tests::active_turn_recovery_closes_without_starting_the_provider`. This branch touches three TypeScript files under `server/src` and no runner code. Confirmation from a maintainer that this failure is pre-existing on `master` would help, since the last recorded `master` runs in the Actions history are from April and do not cover it.

### DIFF
--- a/server/src/__tests__/task-watchdogs-classifier.test.ts
+++ b/server/src/__tests__/task-watchdogs-classifier.test.ts
@@ -388,10 +388,10 @@ describe("task watchdog subtree classifier", () => {
     expect(result.state).toBe("not_applicable");
   });
 
-  // LUN-7052 hypothesis check. That ticket assumed a child's pending card masks
-  // a dead parent as `live` forever. It does not: waits are material state that
-  // feeds the stop fingerprint, and only a run or a queued wake yields `live`.
-  // Pinned so the assumption cannot come back as a "fix".
+  // Hypothesis check. A plausible reading of the refusal is that a child's
+  // pending card masks a dead parent as `live` forever. It does not: waits are
+  // material state that feeds the stop fingerprint, and only a run or a queued
+  // wake yields `live`. Pinned so the assumption cannot come back as a "fix".
   it("does not read a parent as live because children carry pending interactions", () => {
     const otherChildId = "child-2";
     const result = classify({
--- a/server/src/__tests__/task-watchdogs-classifier.test.ts
+++ b/server/src/__tests__/task-watchdogs-classifier.test.ts
@@ -387,4 +387,44 @@ describe("task watchdog subtree classifier", () => {
 
     expect(result.state).toBe("not_applicable");
   });
+
+  // LUN-7052 hypothesis check. That ticket assumed a child's pending card masks
+  // a dead parent as `live` forever. It does not: waits are material state that
+  // feeds the stop fingerprint, and only a run or a queued wake yields `live`.
+  // Pinned so the assumption cannot come back as a "fix".
+  it("does not read a parent as live because children carry pending interactions", () => {
+    const otherChildId = "child-2";
+    const result = classify({
+      issues: [
+        issue({ status: "in_progress" }),
+        issue({ id: childId, identifier: "PAP-2", parentId: sourceId, status: "in_progress" }),
+        issue({ id: otherChildId, identifier: "PAP-3", parentId: sourceId, status: "in_progress" }),
+      ],
+      pendingInteractions: [
+        { companyId, issueId: childId, id: "interaction-1", kind: "ask_user_questions", status: "pending" },
+        { companyId, issueId: otherChildId, id: "interaction-2", kind: "request_confirmation", status: "pending" },
+      ],
+    });
+
+    expect(result.state).toBe("stopped");
+  });
+
+  it("names the issue and the signal that make the subtree live", () => {
+    const result = classify({
+      issues: [
+        issue(),
+        issue({ id: childId, identifier: "PAP-2", parentId: sourceId, status: "todo" }),
+      ],
+      activeRuns: [{ companyId, issueId: sourceId, agentId: "agent-1", status: "running" }],
+      queuedWakeRequests: [{ companyId, issueId: childId, agentId: "agent-1", status: "queued" }],
+    });
+
+    expect(result.state).toBe("live");
+    if (result.state !== "live") return;
+    expect(result.liveSignals).toEqual([
+      { issueId: childId, identifier: "PAP-2", kind: "queued_wake_request", status: "queued" },
+      { issueId: sourceId, identifier: "PAP-1", kind: "active_run", status: "running" },
+    ]);
+    expect(result.reason).toContain("PAP-1 has a active_run in status running");
+  });
 });
--- a/server/src/__tests__/task-watchdogs-scheduler.test.ts
+++ b/server/src/__tests__/task-watchdogs-scheduler.test.ts
@@ -685,7 +685,7 @@ describeEmbeddedPostgres("task watchdog scheduler", () => {
     });
 
     expect(revalidated.allowed).toBe(false);
-    // LUN-7052: the refusal has to name which issue is live and what makes it
+    // The refusal has to name which issue is live and what makes it
     // live. `GET /api/live-runs` is scoped to the caller, so a watchdog run
     // cannot otherwise see another agent's run on the watched issue, and the
     // bare "subtree is live" verdict reads as a platform bug.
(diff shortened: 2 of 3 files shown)


simberthon:
Addressed the three review comments in d71d577.

**P1, race remains before enqueue.** Correct, and the narrowing was in the wrong place. The recheck now runs twice: once before `ensureReusableWatchdogIssue`, which moves the two slowest writes in the window (creating the review issue and its fingerprint comment) out of it, and once immediately before `enqueueWakeup`. What is left is the watchdog row update and one activity-log write. I did not try to make it atomic. Closing it properly means the wake insert itself carries the liveness predicate, which changes the `enqueueWakeup` contract for every caller, not just this one. That felt out of scope for a bug fix, so the PR now states the residual window in **Risks** instead of implying it is closed. The refusal diagnostics in this same PR are what make the residual case survivable: a watchdog that still gets refused is told which issue is live and what makes it live, rather than being handed a verdict it has no API to check.

**P1, unwoken review suppresses retry.** Correct and worse than the race it was guarding, because it was permanent. Fixed both ways you suggested. The first recheck runs before the artifact is published, so the common case never creates one. If the second recheck loses after publication, the artifact is re-stamped with a `task_watchdog_untriggered:` fingerprint and an `issue.task_watchdog_trigger_abandoned` activity entry, so `sameFingerprintWatchdogReviewIsStillOpen` no longer matches it. It is a sentinel rather than a clear because `issues.origin_fingerprint` is `NOT NULL`, and the prefix sits outside what `normalizeStopFingerprint` accepts so no existing reader mistakes it for a stop fingerprint. The scheduler test now asserts the retry end to end: after the racing run ends, the next reconcile pass triggers on the same stopped state instead of reading its own leftover artifact as a completed review. That assertion failed against the previous commit.

**P2, internal ticket IDs.** Removed from all three files, including the Postgres trigger name in the scheduler test, which is now `watchdog_enqueue_race`. Thanks for catching it in the test helper names too.

One note on CI that is not mine to fix: `ci / Build` is red on `acpx_provider_backend::tests::active_turn_recovery_closes_without_starting_the_provider`, a Rust test in the runner crate. This branch changes three TypeScript files under `server/src` and no runner code.

simberthon:
Closing this PR. It sat open for a week with every CI check green and no maintainer review, and we do not have merge rights on this repo, so there is no path to land it from our side.

The only red check was Greptile Review at confidence 4/5 against a required 5/5, which its own summary describes as a threshold, not a review failure. The three inline findings it left were all on the previous commit `e3f8013` and were addressed in `5033af6`.

The branch `lun-7052-watchdog-liveness` is left in place. Happy to reopen if a maintainer wants to pick it up.

PR #12855, opened 2026-09-04
ci: keep the Docker build context complete and guard it on every PR

devinfoley:
## Thinking Path

> - GitHub Actions builds the Docker images that ship Paperclip, and downstream deployments consume the `-cloud` image variant on every master merge.
> - PR #12769 slimmed the Docker build context with a broad `.dockerignore` block for `packages/paperclip-runner`, and the block also removed three files the image build itself reads.
> - The image build re-runs the runner's generated-file drift checks, so it found no committed capability contract in the context and failed on every master commit after the merge.
> - PR CI never runs those checks against the Docker context, so the pull request stayed green and the breakage only appeared post-merge, on every image build.
> - This pull request restores the three files with narrow `.dockerignore` exceptions and adds a PR CI job that runs the drift checks against the exact Docker build context.
> - The benefit is that image publishing works again now, and the next context-slimming regression fails the pull request instead of every post-merge image build.

Refs #12769 (the context-slimming change that exposed this) and #12608 (which committed the generated contract outputs the image build checks).

**What happened?**

Every `Docker` workflow run on master failed from 2026-09-04 12:58Z onward, in both the `build-and-push` and `build-and-push-cloud` jobs. The failing step reported `Generated contract drift: generated/capability/capability-contract.md` from `check:capability-contract` inside `pnpm --filter @paperclipai/server build`. The committed contract file is current — regeneration on a full checkout is a no-op. The file was simply absent from the build context: the new `packages/paperclip-runner/**/*.md` ignore rule strips the committed drift-check outputs (`generated/capability/capability-contract.md`, `generated/capability/downstream-handoff.md`), and the `packages/paperclip-runner/docs` rule also strips `docs/capability-contract.md`, which `check:capability-inventory` reads next in the chain. No cloud image published for eight hours, which stalled every downstream deployment that consumes the canary images.

**Expected behavior**

The Docker build context must contain every file the image build reads, and a change that removes one must fail the pull request that introduces it, not every image build after the merge.

**Steps to reproduce**

1. Check out master at any commit from `af3023f1` onward.
2. Run `docker buildx build -f .github/docker-context-checks.Dockerfile .` (the probe added by this PR), or start the real `Docker` workflow build.
3. Observe `Generated contract drift: generated/capability/capability-contract.md` — while `node packages/paperclip-runner/scripts/generate-capability-contract.mjs --check` passes on the same checkout outside Docker.

`d593463ab` (master tip at diagnosis time; first failing commit `af3023f1`).

**Deployment mode**

GitHub Actions image builds (`docker.yml`), consumed by managed cloud deployments.

## What Changed

- `.dockerignore`: narrow exceptions (last match wins) re-include the committed drift-check outputs (`!packages/paperclip-runner/generated/**`) and the inventory check's documentation input (`!packages/paperclip-runner/docs/capability-contract.md`). Every other exclusion from #12769 stays: no crate declares an explicit `[[test]]` target, so cargo builds without the `tests` directories, and the image build chain never runs the excluded smoke scripts.
- `.github/docker-context-checks.Dockerfile` (new): a small probe that COPYs the real build context — identical `.dockerignore` semantics — and runs the dependency-independent drift checks inside it (`generate-capability-contract.mjs --check`, `check-capability-inventory.mjs`). ajv installs in an isolated directory for schema validation only; codegen checks such as `generate-protocol-schema-module` stay out because their emitted bytes vary with the ajv release and would raise false drift alarms outside the locked dependency tree.
- `.github/workflows/pr-trusted.yml`: new `docker_context_integrity` job builds the probe on every full-CI pull request, and the existing `verify` aggregate now requires its result, so the guard gates merges through the same required check as the other lanes.
- Activation note: `pr.yml` pins `pr-trusted.yml` by commit SHA, so the new job starts gating pull requests after the usual follow-up `ci: activate ...` pin bump once this merges. The `.dockerignore` fix needs no activation — `docker.yml` reads it directly, so image builds recover on the first master commit after this merges.

## Verification

- `docker buildx build -f .github/docker-context-checks.Dockerfile .` on master (before the `.dockerignore` fix): fails with the exact production error, `Generated contract drift: generated/capability/capability-contract.md`.
- Same command with the `.dockerignore` exceptions applied: passes, which also proves BuildKit honors the `!` exceptions, including the file inside the excluded `docs` directory.
- `node scripts/generate-capability-contract.mjs --check` on a full checkout: passes both before and after, which confirms the committed contract was never stale — only missing from the context.
- Static sweep of every script in the image build chain (`build`, `build:typescript` and their `check:*` steps) against the ignore rules: the three restored files are the only build inputs the #12769 block strips.
- YAML for `pr-trusted.yml` lints clean.

## Risks

- Low. The `.dockerignore` exceptions only re-add three committed files to the build context; image contents do not change otherwise.
- The probe job adds one context transfer and two Node scripts per full-CI pull request run (about one to two minutes, no dependency install beyond one isolated ajv package).
- The `verify` aggregate now also requires the new job, mirroring the existing pattern for the other lanes; on non-full-CI runs the job skips and `verify` asserts the skip, unchanged from how the other lanes behave.
- The new job only takes effect for pull requests after a follow-up pin bump in `pr.yml` (same two-step flow as every `pr-trusted.yml` change).

## Model Used

Claude Fable 5 (Anthropic, model id `claude-fable-5`), extended thinking, agentic tool use in Claude Code: GitHub Actions log forensics to isolate the failing check, static analysis of the build-chain scripts against the ignore rules, and local docker buildx runs to reproduce the failure and verify the fix.

## Checklist

- [x] I have run tests locally and they pass (the docker probe, both failing-before and passing-after; the drift checks themselves on a full checkout)
- [x] I have added or updated tests where applicable (the probe IS the regression test for this class)
- [x] I have updated relevant documentation to reflect my changes (inline comments in `.dockerignore` and the probe explain the invariant)

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -29,3 +29,11 @@ packages/paperclip-runner/**/*.test.ts
 packages/paperclip-runner/**/*.test.tsx
 packages/paperclip-runner/runner/crates/*/tests
 packages/paperclip-runner/scripts/*-smoke.mjs
+# Exceptions (last match wins): the image build re-runs the runner's
+# generated-file drift checks, so their committed outputs and inputs must
+# survive the slimming above. 2026-09-04: the *.md rule stripped the
+# committed capability contract out of the context and every image build
+# on master failed its drift check — .github/docker-context-checks.Dockerfile
+# now guards this in PR CI.
+!packages/paperclip-runner/generated/**
+!packages/paperclip-runner/docs/capability-contract.md
--- a/.github/docker-context-checks.Dockerfile
+++ b/.github/docker-context-checks.Dockerfile
@@ -1,0 +1,43 @@
+# Runs the runner's generated-file drift checks against the EXACT build
+# context the image builds see — same .dockerignore semantics — so a
+# context-slimming change that strips a committed build input fails the
+# pull request instead of every post-merge image build. (2026-09-04: a new
+# `packages/paperclip-runner/**/*.md` ignore rule stripped the committed
+# capability contract out of the context; every Docker build on master then
+# failed its drift check, and no cloud image published for eight hours
+# while PR CI stayed green.)
+#
+# Only checks whose compared output is independent of dependency versions
+# run here: ajv is installed for schema VALIDATION only (pinned to the
+# runner's declared range), while codegen checks like
+# generate-protocol-schema-module stay out — their emitted bytes vary with
+# the ajv release, so running them against a fresh install would raise
+# false drift alarms. Those still run inside the real image build, which
+# installs the locked dependency tree; the existence assertions below keep
+# their committed inputs and outputs covered by this probe regardless.
+#
+# node:24-slim — the runner requires Node >= 24.11 and the production
+# image builds on Node 24; the digest pin keeps the security gate's own
+# runtime immutable.
+FROM node:24-slim@sha256:ba849c60be29959425b8734d57b8b4b7d56f98edd9504c9af091d5281095a71e
+WORKDIR /context
+COPY . .
+# Committed artifacts the image build reads whose drift checks cannot run
+# here (they need the locked dependency tree or compiled dist/). Existence
+# in the context is the property this probe guards; content correctness is
+# the real build's job. If a path is intentionally removed from the repo,
+# update this list in the same PR.
+RUN test -f packages/paperclip-runner/generated/capability/semantic-tool-contracts.json \
+ && test -f packages/paperclip-runner/generated/semantic-action-catalog.json \
+ && test -f packages/paperclip-runner/spec/evals/stress-workflow-traceability.json \
+ && test -d packages/paperclip-runner/protocol/fixtures/replay
+# ajv is installed in an isolated directory (the runner's own package.json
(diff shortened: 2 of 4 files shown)


devinfoley:
Addressed all five findings: the probe now asserts existence of the committed build-chain artifacts whose drift checks cannot run outside the locked dependency tree (semantic-tool-contracts.json, semantic-action-catalog.json, the traceability spec, and the replay fixtures); the lane sets up Buildx explicitly before invoking it, matching the image-build workflows; the probe base is node:24-slim pinned to an immutable digest (the runner's supported major, same as the production image); and the workflow pin test now asserts both halves of the lane's gating — the env mapping and the full-CI success assertion — so dropping either fails the test. Verified locally: pin tests 11/11, probe passes on the fixed tree and still reproduces the production drift failure without the .dockerignore exceptions. @greptileai

devinfoley:
Pinned docker/setup-buildx-action to its v4 commit SHA (37fe6310), matching the immutable-pin convention of every other action in this workflow. The probe base image was already digest-pinned in the previous push. @greptileai

PR #10157, opened 2026-07-24
feat(docker): publish a cloud image variant with built bundled plugins

devinfoley:
## Thinking Path

> - Paperclip is the open source app people use to manage AI agents for work
> - Managed (cloud-hosted) deployments configure instances through `PAPERCLIP_MANAGED_CONFIG`, including a `plugins.autoInstall` key list that the boot-time installer resolves against the bundled plugin catalog
> - The installer requires each bundled plugin's `dist/manifest.js` (`server/src/services/bundled-plugins.ts`), but the published image only ships the sandbox providers' *source* — they are intentionally excluded from the pnpm workspace, and the Dockerfile never builds them
> - Every managed auto-install therefore logs `bundled plugin bundle not present; skipping auto-install` and no sandbox provider can be provisioned through managed config
> - Baking built plugins into the single published image would fix it but makes every self-hosted pull carry the providers' `node_modules` for a managed-only mechanism
> - This pull request adds a `cloud` Dockerfile target extending `production` with built bundled plugins — parameterized by build arg and currently just `daytona` — published alongside the default image with a `-cloud` tag suffix
> - The benefit is working plugin auto-provisioning for managed deployments while the self-hosted image stays byte-identical and the cloud variant only carries what is actually deployed

## Linked Issues or Issue Description

Fixes #10158 (filed for this problem; no prior issue existed — searched for duplicate/related PRs and issues around bundled plugins, docker image variants, and auto-install). Summary: **What happened:** on a managed instance with `plugins.autoInstall: ["daytona"]` delivered via `PAPERCLIP_MANAGED_CONFIG`, boot logs `bundled plugin bundle not present; skipping auto-install` with `pluginPath: /app/packages/plugins/sandbox-providers/daytona`, and the plugin is never installed. **Expected:** the advertised bundled-catalog keys are installable from the published image. **Why:** the image ships plugin source without `dist/` — nothing in the Dockerfile builds the workspace-excluded sandbox providers.

## What Changed

[...]

PR #12856, opened 2026-09-04
build(docker): prune devDependencies from the production and cloud images

nickyleach:
## Thinking Path

> - The Dockerfile's `production` and `cloud` stages ship the running server.
> - The `production` stage copies the entire `build` stage `/app` directory, unpruned.
> - That copy includes every workspace member's devDependencies and unrelated workspace members' own dependencies, none of which the running server touches.
> - This pull request adds a pruning stage that re-resolves `node_modules` to just what `@paperclipai/server` needs at runtime before `production` copies it.
> - The benefit is a smaller, faster-to-pull image with no change to what the server actually does.

The `production` and `cloud` Docker image targets built from the root `Dockerfile`.

**Subsystem affected**
Cross-cutting (root `Dockerfile`, `server/package.json`).

**Current behavior**
`production`'s final `COPY --from=build /app /app` copies the entire build-stage workspace, including every workspace member's devDependencies (typescript, vite, vitest, storybook, rolldown, ...) and unrelated workspace members' own dependencies (`ui`'s client-side bundle dependencies like `mermaid` and `lucide-react`, already bundled separately into `ui/dist`). None of that is read by the running server process. On an `arm64` build this measured out to a 3.39GB `COPY` layer, out of a 7.5GB total `cloud` image.

**Proposed behavior**
A new `pruned-app` stage forks from `build` after all builds finish, deletes the hoisted `node_modules`, and reinstalls with `pnpm install --prod --frozen-lockfile --filter='@paperclipai/server...'` (server plus everything it actually depends on, transitively, production-only). `production` copies from `pruned-app` instead of `build`. The prune stage does not run inside `build` itself, because two later stages (`cloud-plugins`, `cloud-server-deps`) fork from `build` and still need its full devDependency toolchain to compile their own TypeScript at image build time.

**Reason and benefit**
Smaller images pull faster in CI and for self-hosted operators, and cost less to store in the registry. Measured locally: the `cloud` target drops from 7.5GB to 5.53GB (-26%), and `production` lands at 5.26GB.

**Breaking changes**
None. The pruned `node_modules` still contains everything `@paperclipai/server`'s dependency graph declares, including workspace-linked packages resolved by TypeScript source (their `exports` field points at `./src/*.ts`, loaded through `tsx` at runtime) -- this PR also moves `tsx` from `server/package.json`'s `devDependencies` to `dependencies`, since the image's own `ENTRYPOINT` imports it directly and a devDependency prune would otherwise have deleted it and broken every container boot. Verified by boot-testing the built `cloud` image end to end (see Verification).

**Additional context**
No related open issue. I searched GitHub for prior or in-flight PRs touching Docker image size, `node_modules` pruning, or the `production`/`cloud` stages and found none that overlap ([#10157](https://github.com/paperclipai/paperclip/pull/10157) added the `cloud` stage this PR builds on top of, already merged).

## What Changed

- Added a `pruned-app` stage to the `Dockerfile`, forked from `build`, that reinstalls `node_modules` as `--prod` and filtered to `@paperclipai/server`'s own dependency graph.
- Changed `production`'s final app copy to `COPY --chown=node:node --from=pruned-app /app /app`, replacing `--from=build`.
- Moved `tsx` in `server/package.json` from `devDependencies` to `dependencies`, since it's a real runtime dependency of the image's `ENTRYPOINT`, not a dev-only tool.

## Verification

- `docker buildx build --target cloud ...` and `--target production ...` both build successfully with the change.
- Image size: `cloud` 7.5GB → 5.53GB, `production` → 5.26GB. The `COPY --from=pruned-app /app /app` layer alone is 1.82GB, down from a 3.39GB unpruned `COPY --from=build /app /app`.
- `docker run` on the built `cloud` image boots cleanly: embedded PostgreSQL initializes, all migrations apply, `GET /api/health` returns `200`, the UI root serves, `@sentry/node` inside the image resolves to the version `server/package.json`'s `peerDependencies` declares.
- `sh -s < scripts/assert-orphan-reaping.sh` against the built image passes (tini is PID 1, reaps an orphaned descendant) -- the same check `docker.yml`'s `build-and-push` job runs after every real build.
- The `production` target boots the same way and responds `200` on `/api/health`.

## Risks

Low risk. The prune only removes packages nothing in `@paperclipai/server`'s resolved dependency graph reaches; it does not touch source files, workspace-linked packages (resolved by directory, not `node_modules`), or `ui/dist`/`server/dist` build output. The `tsx` reclassification is the one behavior-relevant line -- without it, this change would have broken every container boot, which is exactly why it's included and boot-verified above rather than left implicit.

## Model Used

Claude Sonnet 5 (`claude-sonnet-5`), via Claude Code. Extended reasoning throughout; used tool use (Bash, Docker builds, file edits) and an isolated `docker buildx` container-driver builder for all verification. No separate model handled a distinct sub-task -- one continuous session investigated, implemented, and verified this change.

## Checklist


nickyleach:
@greptile-apps this looks like a false positive for this repo's CI setup, not a real problem — filing for the record rather than changing the PR.

`pr-trusted.yml`'s `policy` job runs `pnpm install --resolution-only --ignore-scripts --no-frozen-lockfile` against the PR's merge tree, detects that `pnpm-lock.yaml` is stale relative to `server/package.json`'s `tsx` move, and uploads the regenerated lockfile as an artifact that every downstream job consumes instead of the raw committed one. That's why `ci / Build`, `ci / Typecheck + Release Registry`, and every `ci / General tests` shard just passed despite the mismatch you flagged — they never actually ran `pnpm install --frozen-lockfile` against the stale committed lockfile.

This repo intentionally doesn't commit lockfile-only diffs in feature PRs (see `.github/workflows/refresh-lockfile.yml`, which auto-opens and auto-merges a `chore/refresh-lockfile` PR after any push to master that needs one). `docker.yml`'s own build has the same self-healing refresh step before it builds the image. So there's no gap here for the real, post-merge Docker build either.

Confirmed empirically: all real CI checks are green on this PR as committed.

PR #12864, opened 2026-09-05
fix(recovery): stop an infrastructure run failure from writing a dead `blocked` (LUN-7056)

simberthon:
## Thinking Path

> - The recovery subsystem decides what happens to an issue after its run fails
> - It must tell two causes apart. A business blocker stops the work. An infrastructure failure only delays it
> - A provider session limit failed 25 runs across 11 issues in 30 minutes. The recovery pass moved 9 of those issues to `blocked`
> - Those issues had an empty `blockedByIssueIds` and a null `unblockDescriptor`. Nothing can wake that state. The dependency graph has no blocker to resolve. The unblock notifier has no owner to notify
> - The quota recovery path already handles this correctly. It keeps the issue status and it arms a monitor at the provider reset time. It never ran, because the error code gate rejected the failure first
> - This pull request lets the engine failure codes reach the quota classifier, and it stops the recovery escalations from writing an unroutable `blocked`
> - The benefit is that a provider outage now delays the work instead of killing the issue

**What happened?**

A provider session limit made many runs fail. The recovery pass then moved the affected issues to `blocked`. Each of those issues had an empty `blockedByIssueIds` and a null `unblockDescriptor`.

An issue in that state has no wake path. `deliverAgentUnblockNotification` returns early because there is no owner. Dependency wakeups do nothing because there is no blocker to resolve. The issue stays there until a person finds it.

The root cause is an error code gate. The ACP engine puts its own phase code on every failed turn. Provider exhaustion therefore arrives as `acpx_turn_failed`, not as `adapter_failed`. Both quota detectors read the error code before the message:

```ts
if (latestRun.errorCode !== "adapter_failed" && ...) return null;
```

The message `You've hit your session limit · resets 8am (Asia/Bangkok)` already matches `PROVIDER_QUOTA_ERROR_RE`. The reset clock already parses. The code gate rejects the run before either is used. The issue then falls through to the stranded escalation, which writes `blocked`.

**Expected behavior**

An infrastructure failure must keep the issue status and arm a retry. Only a business blocker must produce `blocked`. An issue in `blocked` must always stay routable.

**Steps to reproduce**

1. Run an agent through the ACP engine.
2. Make the provider return a session limit during the turn. The run ends with `errorCode: "acpx_turn_failed"`.
3. Run the stranded issue reconcile pass.
4. The issue moves to `blocked` with no blockers and no unblock descriptor.

Reproduced on `master` at f2349990c.

**Agent adapter(s) involved**

Not adapter specific. The defect is in the core recovery service. Any adapter that runs through the ACP engine is affected.

## What Changed

- The `acpx_*` engine failure codes can now reach provider quota classification. `timeout` stays excluded, because it is a Paperclip level timeout that can wrap an unrelated downstream service. An existing test pins that behavior. `acpx_auth_required` and `acpx_backend_missing` stay excluded, because they are real configuration blockers that need a person.
- An engine phase code opens the quota path only. A `configuration_incomplete` diagnosis moves an issue to `blocked`, so it keeps its original narrower gate.
- Every recovery escalation that parks an issue in `blocked` without a first class blocker now sets an `unblockDescriptor`. The route layer already enforced this rule with a 422. The recovery service bypassed it, because it calls `issuesSvc.update` directly.
- A startup pass repairs issues that an earlier outage already left unroutable. The pass is bounded per run. It logs what it deferred, so a partial sweep does not read as full coverage.

## Verification

Run the affected suites:

```
npx vitest run server/src/services/recovery/ \
  server/src/__tests__/issue-recovery-actions.test.ts \
  server/src/__tests__/heartbeat-retry-scheduling.test.ts \
  server/src/__tests__/attention-service.test.ts
```

Result: 190 tests pass. The typecheck is clean for the changed files.

The end to end test replays the production payload exactly. It uses `acpx_turn_failed` with the real message. It asserts that the status stays `in_progress` and that the monitor is armed. The test was confirmed to fail on the parent commit, so it is a real regression test:

```
AssertionError: expected +0 to be 1
  expect(result.providerQuotaMonitored).toBe(1);
```

A second test proves that the repair pass leaves a business blocked issue untouched.

### Scope note: the repair pass is startup-only, deliberately

`repairUnroutableBlockedIssues` runs once per process start and is not part of the periodic
recovery sweep. Rows past the per-pass cap wait for the next restart rather than for a scheduled
retry, and the log says exactly that. This is a deliberate scope cut, not an oversight: the
per-pass bound is the safety property this change is built around, since the pass mutates issue
statuses in bulk on a live control plane, and bounding how much of a board can resume at once is
worth more than covering the tail automatically. Making the pass recurring is a separate change
with a separate risk profile.

### Reviewer follow-up on `acpx_session_config_failed`

Answering the review question rather than deferring it: that code is stamped when
`configure_session` fails while applying the run's model / thinking-effort / fast-mode overrides
(`applySessionConfigOptions` in `packages/adapter-utils/src/acpx-engine/execute.ts`), so it can
carry a durable, human-actionable misconfiguration - a rejected model id, or a runtime that
exposes no config controls at all - and not only a transient engine hiccup. Leaving it in the
engine set unqualified made it bypass the `configuration_incomplete` diagnosis, so a real config
blocker under that code would fall through unclassified. It now stays in the engine set (so
quota-shaped text under it still defers instead of blocking) but no longer suppresses the
configuration diagnosis. Generic engine crashes still do. Covered by three new tests.

## Risks

Medium. The change alters recovery behavior on a live control plane.

- The classifier gate is widened only for the quota path. A message must still match `PROVIDER_QUOTA_ERROR_RE`, so an unrelated engine crash stays unclassified. A test covers that case.
- The startup repair pass mutates issue status in bulk. This is the main risk. It is deliberately narrow. It only lifts an issue when that issue's own latest run failed for an infrastructure cause. It does not act on the shape of the blocked state alone.
- On the instance used for measurement, 40 of 60 blocked issues have no blockers and no descriptor. Most of them are genuine business blocks that wait on a human decision. Some predate the routable blocked rollout. Lifting all of them would lose real state and would start a wave of runs. The repair pass leaves them alone by design.
- The pass is capped per run, so a large backlog cannot resume a whole board at once.
- No schema change. No migration. Rollback is a revert of this commit.

## Model Used

Claude Opus 5, 1M context window, extended thinking, with tool use and code execution.

## Checklist


simberthon:
Auto-merge intentionally **not** armed yet (LUN-5886 named-reason exception).

This diff adds a startup pass that mutates issue statuses in bulk on the fleet control plane, which puts it in the LUN-4497 reviewer-gate categories (prod infra + bulk write). Routing to the reviewer agent first; auto-merge will be armed on the reviewed SHA once the verdict lands.

Head SHA under review: `d825c545076fabfbedccc63c0e7d7a80c70dd2e3`

simberthon:
All three review findings were valid and are fixed in `bab098acb`.

**P1 monitor failure leaves issue idle.** Correct, and the worst of the three. `scheduleProviderQuotaRecoveryMonitor` only schedules for the agent that owns the failed run, so on a mismatch the pass would have committed `in_progress` and installed no wake. That is worse than the state it was repairing: an idle `in_progress` still has no execution path. The pass now checks the run owner against the assignee before it restores anything, and if the monitor still fails to arm it puts the issue back to `blocked` with an unblock descriptor. A test covers the other-agent case.

**P1 deferred repairs never resume.** Correct. The pass is startup only, so the log promised a sweep that does not exist. I kept the per-pass cap, because bounding how much of a board can resume at once is the safety property I want, and corrected the claim: the log now says the remainder waits for the next startup, and it also fires when the scan itself is truncated.

**P2 repair query is unbounded.** Correct. The scan is now bounded, so startup no longer reads every historical row to repair at most 25.

191 tests pass across the recovery, recovery-actions, retry-scheduling, routable-blocked and attention suites. Typecheck is clean for the changed files.

simberthon:
Review findings addressed in `c67c3b397`.

**must-fix — the startup mock.** Correct, and the important one: the claim in my previous comment was about a local run, not the pushed SHA, and CI was red because of this diff. `server-startup-feedback-export.test.ts` hand-rolls `heartbeatServiceMock` as an object literal and never grew the `repairUnroutableBlockedIssues` key that `startServer()` now calls unconditionally. The mock now returns the same shape the implementation returns. Verified in both directions on this branch: without the entry the suite fails with exactly `TypeError: heartbeat.repairUnroutableBlockedIssues is not a function`, with it the suite passes. Checked that no other test hand-rolls the same mock — only this file does; `heartbeat-issue-liveness-escalation.test.ts` uses the real service.

**question — `acpx_session_config_failed`.** Your read is right and it was a real hole. That code is stamped when `configure_session` fails while applying the run's model / thinking-effort / fast-mode overrides (`applySessionConfigOptions`, `packages/adapter-utils/src/acpx-engine/execute.ts:2463`), so it covers a durable, human-actionable misconfiguration — a rejected model id, or a runtime exposing no config controls at all — as well as a transient hiccup. Keeping it in the engine set unqualified made it bypass the `configuration_incomplete` diagnosis, so a genuine config blocker under that code fell through unclassified instead of surfacing. It stays in the engine set, so quota-shaped text under it still defers instead of blocking, but it no longer suppresses the configuration diagnosis; generic engine crashes still do. Three tests cover it: quota text under that code still defers, a `model_not_found` under it now diagnoses `configuration_incomplete`, and the same text under `acpx_turn_failed` / `acpx_runtime_error` / `acpx_session_init_failed` still returns null.

**nice-to-have — startup-only repair.** Written into the PR description as a deliberate scope cut, with the reason: the per-pass bound is the safety property, since the pass mutates issue statuses in bulk on a live control plane.

Tests: 39 in the classification and startup suites, 53 in the recovery-actions suite, all green locally. Typecheck clean on the changed files.

The unrelated `cursor-local-execute.test.ts` timeout in shard 1 is untouched by this diff; I will read the new CI run rather than assume it.

PR #12650, opened 2026-09-01
Fix checked-out run binding on first write

kallehiitola:
## Thinking Path

> - Paperclip manages AI agent work through an audited control plane.
> - Run context binding limits cross-issue influence.
> - The first write could bind context without a persisted checkout.
> - The fallback must prove ownership by company, agent, and run.
> - This pull request adds that proof and fails closed without it.
> - The benefit is safer first-write context binding without reverse row-lock order.

## Linked Issues or Issue Description

- Refs #12315
- Related reviewed patch: https://github.com/dzianisv/paperclip/pull/6

## What Changed

- Require a persisted checkout for the same company, agent, and run before first-write binding.
- Return HTTP 403 and keep the snapshot unchanged for an unclaimed target.
- Add unit and embedded-Postgres regression coverage.

## Verification

- Review the focused unit and embedded-Postgres regression tests in commit `679b762`.
- Confirm an unclaimed target returns HTTP 403 and keeps `contextSnapshot` unchanged.
- Confirm a checked-out target binds successfully.
- Confirm the issue ownership read does not use a row lock.

## Risks

- Low migration risk. The change only rejects writes that lack persisted checkout proof.
- A caller that relied on implicit first-write binding now receives a fail-closed response.

## Model Used

- OpenAI `gpt-5.6-sol` with tool use reviewed the patch and prepared this upstream submission. Den authored the carried commits.

## Checklist

- [x] I have included a thinking path that traces from project context to this change
- [x] I have specified the model used (with version and capability details)
- [ ] I have checked ROADMAP.md and confirmed this PR does not duplicate planned core work
- [x] I have searched GitHub for duplicate or related PRs and linked them above
- [x] I have either (a) linked existing issues with `Fixes: #` / `Closes #` / `Refs #` OR (b) described the issue in-PR following the relevant issue template
- [x] I have not referenced internal/instance-local Paperclip issues or links (only public GitHub `#NNN` / `github.com/paperclipai/paperclip` URLs)
- [x] My branch name describes the change and contains no internal Paperclip ticket id or instance-derived details
- [ ] I have run tests locally and they pass
- [x] I have added or updated tests where applicable
- [ ] I have updated relevant documentation to reflect my changes
- [x] I have considered and documented any risks above
- [ ] All Paperclip CI gates are green
- [ ] Greptile is 5/5 with no open P2s, recommendations, or follow-ups
- [x] I will address all Greptile and reviewer comments before requesting merge

When squash-merging, preserve original authorship:

[...]

PR #12871, opened 2026-09-05
fix(server): require checkout proof before run binding

kallehiitola:
Supersedes closed PR #12315 with the reviewed follow-up from dzianisv/paperclip#6. The change binds an initially unscoped run only after a persisted checkout proves the same company, agent, and run; keeps the issue ownership read non-locking to avoid reverse row-lock order; and adds unit plus embedded-Postgres regressions that fail an unclaimed target closed with HTTP 403. The fork-only dependency-review check is unavailable because the contributor fork has no dependency graph; upstream checks are the authoritative validation path.


kallehiitola:
Closing as an exact duplicate of #12650. Both PRs point to head commit 1ec7678595588d47339edcdf5047e7ae23aaf1d5.

PR #12896, opened 2026-09-05
fix(costs): add pricing_methodology flag and nullable rate_card_cents

phattbeats:
## Thinking Path

> - Agents' spend is metered per `cost_events` row, and the ledger derives a notional rate-card cost alongside billed cash
> - A change that started recording `cache_write_tokens` separately meant older rows only have `cache_write_tokens=0` folded into `input_tokens`, so their existing rate-card figure is a real lower bound, not a full measurement
> - Silently treating an old row's rate-card figure as equivalent to a freshly measured one misrepresents historical spend to anyone reading the ledger
> - This pull request adds a non-nullable `pricing_methodology` flag (`measured` / `pre_cache_write_aware` / `unpriced`) to `cost_events`, makes `rate_card_cents` nullable so "genuinely unpriced" can be recorded as NULL instead of an invented zero, and reclassifies the affected historical rows
> - The benefit is that dashboards and API consumers can tell a measured cost apart from an honest lower bound or an unpriced gap, instead of every row looking equally authoritative

No public GitHub issue exists for this change yet. Described inline, following the bug report template:

**What happened?**
`cost_events.rate_card_cents` is NOT NULL with a default of 0, so rows for models with no rate-card entry (e.g. subscription/OAuth auth with an unlisted model) are stored as a rate-card cost of exactly $0.00, which is indistinguishable from "measured and genuinely free." Separately, historical rows written before `cache_write_tokens` existed as a column fold those tokens into `input_tokens`, so their rate-card figure is a real lower bound rather than a full measurement, and nothing on the row says so.

**Expected behavior**
A row where the rate-card cost is unknown should be recorded as unknown (NULL), with an explicit `pricing_methodology` flag distinguishing "measured," "pre_cache_write_aware" (folded, lower-bound), and "unpriced," rather than every row silently reading as equally authoritative.

**Steps to reproduce**
1. Run an agent under subscription/OAuth auth on a model with no rate-card entry.
2. Inspect the resulting `cost_events` row.
3. Observe `rate_card_cents = 0` with no field indicating whether that is a verified $0 or an unpriced gap.

`874eab3` (Vision Quest's migration that added `cache_write_tokens`) through the current `master`.

## What Changed

- Add non-nullable `cost_events.pricing_methodology` (`measured` / `pre_cache_write_aware` / `unpriced`), enforced with a `CHECK` constraint added via an idempotent `DO` block (Postgres has no `ADD CONSTRAINT IF NOT EXISTS`).
- Drop `NOT NULL` on `cost_events.rate_card_cents` so unpriced rows can store `NULL` instead of a fabricated `0`.
- Reclassify historical rows written before cache-write tokens were tracked separately to `pricing_methodology='pre_cache_write_aware'` and `cost_status='reported_pre_migration'`.
- `packages/db/src/backfill-cost-event-rate-card.ts`: standalone backfill script for pre-existing rows, now wired up as `pnpm --filter @paperclipai/db backfill:cost-event-rate-card` and updated to write `NULL` + `pricing_methodology='unpriced'` (instead of a `0` rate card with no methodology) when a model has no rate-card entry.
- `packages/shared/src/validators/cost.ts`: the create-cost-event schema now derives `pricingMethodology` from whether `rateCardCents` was actually supplied (instead of always defaulting to `measured`), and rejects payloads where the two fields disagree (`rateCardCents: null` with a non-`unpriced` methodology, or vice versa).
- Rate-card derivation, cache-write-aware pricing, and ledger status classification from billed cash (not the CLI's self-reported estimate) in `server/src/services/heartbeat.ts` / `packages/adapters/claude-local`.
- Updated Storybook fixtures and added migration/unit tests.

Note: this PR does not change the cost-aggregation endpoints in `server/src/services/costs.ts` to expose an explicit "how many rows in this total are unpriced" signal. Aggregates already sum `rate_card_cents` with `NULL`s excluded (unaffected — `cost_cents`, the real billed figure, is untouched and always populated), so today's dashboards keep working; a completeness indicator is dashboard-facing and best coordinated with the sibling PR that renders these fields, so it isn't included here.

## Verification

- `pnpm --filter @paperclipai/db exec vitest run src/pricing-methodology-migration.test.ts` — 6/6 pass, including a case that would have caught the invalid `ADD CONSTRAINT IF NOT EXISTS` syntax at the SQL layer.
- `pnpm --filter @paperclipai/shared run typecheck` and `pnpm --filter @paperclipai/db run typecheck` — clean, including migration numbering/safety checks.
- Manually reviewed the migration SQL for idempotent re-run safety (guarded `DO` block for the constraint, `WHERE` clauses scoped to unmigrated rows for both `UPDATE`s).

## Risks

- The migration is additive plus two scoped `UPDATE`s on existing rows; it does not touch `cost_cents` (billed cash), so budget accounting is unaffected.
- Deploying this migration without also running the `backfill:cost-event-rate-card` script leaves pre-existing zero-rate-card rows at their current (already-in-production) values — no regression, but the fuller backfill needs a manual run, documented in the migration header and this description.
- `pricing_methodology`/`rate_card_cents` consistency is now enforced at the API validation layer; any existing caller that sends `rateCardCents: null` while omitting or hardcoding `pricingMethodology: "measured"` will now get a 400 instead of silently mislabeling the row. Checked call sites in `heartbeat.ts` already tie the two together correctly.

## Model Used

Claude (Anthropic), Sonnet 5 (`claude-sonnet-5`), agentic coding session with tool use (file edits, test execution, `gh` CLI for review feedback) — no extended thinking mode.

## Checklist


phattbeats:
Pushed a follow-up commit addressing the Greptile review (confidence 1/5):

1. **Invalid migration syntax (blocker)** — Postgres has no `ADD CONSTRAINT IF NOT EXISTS`. Replaced with a `pg_constraint` existence check inside a `DO` block so the migration is idempotent and actually runs.
2. **Backfill script never wired up** — added `pnpm --filter @paperclipai/db backfill:cost-event-rate-card`, documented in the migration header. Still a manual post-deploy step (needs the shared rate-card TS logic, not raw SQL) but now discoverable rather than orphaned.
3. **Inconsistent unpriced representation** — the backfill script now writes `rate_card_cents=NULL` + `pricing_methodology=unpriced` for models with no rate-card entry, instead of a bare `0` with no methodology signal. The API validator (`packages/shared/src/validators/cost.ts`) now derives `pricingMethodology` from whether `rateCardCents` was actually supplied, and rejects payloads where the two fields disagree.
4. **Aggregate completeness (costs.ts)** — left out of this PR intentionally: `cost_cents` (real billed cash) is untouched and always populated, so budget/spend accounting is unaffected. Surfacing an explicit unpriced-row count on the aggregate endpoints is dashboard-facing and better coordinated with the sibling dashboard-rendering PR than redesigned here as a drive-by. Noted as an explicit non-goal in the PR description.
5. **PR description** — rewritten to follow the required template (Thinking Path / What Changed / Verification / Risks / Model Used) with internal ticket references removed.

Branch name still contains an internal identifier; not renaming it since that closes cross-fork PRs on this host. Re-triggering Greptile.

phattbeats:
Found and fixed a real CI failure: the `workspaces-b` job's `turn-characterization.test.ts` was failing because this PR's `execute.ts` change (splitting `cacheWriteTokens` onto its own `UsageSummary` field instead of folding it into `inputTokens`) wasn't reflected in two pre-existing test assertions. Pushed `67dc6a7` to update the expected shape. Re-checking the rest of the failing checks (Greptile Review, `ci / verify`, `commitperclip PR Review`) for anything else real.

Issue #12953, opened 2026-09-07
Hosted paperclip.inc: full app outage (EHOSTUNREACH, now resolved) + persistent Kubernetes sandbox lease failures blocking all agent execution (6 days, survived the outage)


wpescott:
## Summary
My paperclip.inc company ("PromotableYou", account william@promotableyou.com.au, paid subscriber) has been unable to run any agents for 6 days. A separate full-app-outage bug appeared and has since resolved on its own, but the original agent-execution failure has NOT resolved - it just fired again today on a fresh, independently-scheduled batch of 11 runs.

## Update 2026-09-07: fault survived a full outage-and-recovery cycle
Today is a scheduled bi-monthly content-generation date for my company. Every scheduled run failed with the exact same error as 6 days ago:
- Tasks PRO-131 to PRO-134 (all four scheduled pillar-content tasks), all now "blocked"
- Run IDs (agent: Michelle, Head of Marketing): 3617e504, 2f85f50e, 8fbd7dc4, 97792e56, 195d3012, bb595f12, 44152a70, e055185a, f1dcab4a, e096beb3, 561363f5 - 11 failures, all identical error, all failing in 0-1 second
This rules out any theory that the failure was tied to the outage below - it predates it and has now outlasted it too.

## Now-resolved issue: full app outage
For a period, every page under `/PRO/*` returned a raw, unhandled 500:
```
{"statusCode":500,"code":"FST_REPLY_FROM_INTERNAL_SERVER_ERROR","error":"Internal Server Error","message":"connect EHOSTUNREACH 10.0.105.205:3100 - Local (0.0.0.0:0)"}
```
Reproduced independently on `/PRO/dashboard`, `/PRO/agents/all`, and an agent's Configuration page. One reload of the Configuration page instead returned a bare, unstyled `stream timeout` with no app chrome at all. This has since cleared - the app loads normally again as of today.

`10.0.105.205` is a private/internal IP, not anything on my end. Port `3100` is notably the same port the desktop/local Paperclip build listens on, which suggests the hosted backend proxies per-company execution through an internal pod running the same server software, and that pod (or the route to it) was gone during this window.

## The underlying, still-unresolved issue (6 days now)
Every agent run fails instantly (0-1s):
```
Failed to acquire lease for environment "Kubernetes Sandbox" (sandbox): Sandbox provider "kubernetes" is installed via plugin "paperclip.kubernetes-sandbox-provider", but that plugin is currently error. (setup_failed)
```
[...]

Issue #9555, opened 2026-07-14
[Bug] Database connection pool leak causes infinite restart loop under load (launchd KeepAlive)


victor-develop:
## Summary

Paperclip server enters an **infinite restart loop** under moderate agent concurrency due to a **database connection pool leak** in the Node.js server code. The embedded PostgreSQL instance is healthy — the problem is in `paperclipai`'s `pg-pool` usage.

## Symptoms

1. Server starts normally, serves HTTP 200s for ~1 minute
2. Then **all HTTP requests hang indefinitely** (0% CPU on both paperclipai + embedded PG)
3. launchd (`KeepAlive=true`) detects unresponsive process → restarts
4. Fresh process starts → works briefly → pool re-leaks → hangs → restart cycle repeats

## Root Cause

Request handlers and/or run-lifecycle code paths acquire a DB connection via `pool.acquire()` but **fail to release it** (missing `finally { release() }` in some code paths). Over time the pool exhausts, and since no `connectionTimeoutMillis` is configured, `pool.acquire()` blocks **forever** instead of erroring.

### Evidence

- `paperclip doctor` health check triggers the restart
- ~10 TCP connections stuck to embedded PG (port 54329) when hang occurs
- CPU drops to 0% — the process is alive but blocked on `pool.acquire()`
- Instance settings API returns **403 for agent JWTs**, so agents cannot lower `maxConcurrentRuns` to self-recover
- Any `pool` patch applied to the npm bundle is lost on update

## Contributing Factor

The `opencode_local` adapter spawns `opencode serve` child processes but does **not** reliably kill them on run end, leading to orphaned processes (observed: 7+ leaked daemons, ~2.5GB RSS total) that increase memory pressure.

## Suggested Fixes

| Priority | Fix |
|----------|-----|
| **P0** | Set `connectionTimeoutMillis` (e.g. 5000) and `queryTimeout` on the pg-pool config so leaks surface as errors instead of infinite hangs |
| **P0** | Audit all `pool.acquire()` call sites (request handlers, routine dispatch, run lifecycle) and ensure every path has `finally { release() }` |
| **P1** | `opencode_local` adapter must `kill` the `opencode serve` child process when the run ends |

## Environment

- Paperclip CLI (local embedded PostgreSQL, no `DATABASE_URL` set)
- macOS with launchd (`KeepAlive=true`)
- 9 agents, 834 routine runs, ~165 active approval workflows

## Workaround

- Extract data from SQL backup without starting the server
[...]

PR #8780, opened 2026-06-30
fix(db): recycle pooled connections to stop intermittent transaction 500s

calebsimon-CRAFT:
## Thinking Path

> - Paperclip is the open source app people use to manage AI agents for work, with a control-plane HTTP API backed by Postgres.
> - The DB layer (`@paperclipai/db`) opens a single long-lived application connection pool via `createDb()` using postgres-js.
> - That pool was created with a bare `postgres(url)` and no connection-recycling options, so pooled sockets live forever.
> - When the database briefly goes away — an embedded-Postgres restart/flap, or a startup race before Postgres is accepting connections — a half-open/dead socket can linger in the pool. Plain queries are often retried transparently by postgres-js, but a multi-statement `db.transaction()` (BEGIN/…/COMMIT) is **not** retryable and surfaces the dead connection to the caller as an intermittent connection-level 500.
> - This was observed in the wild: `POST /api/projects/:id/workspaces` (the only project route wrapped in a transaction) returned 500 intermittently after an embedded-Postgres flap, while non-transactional routes recovered on their own.
> - Worse, those 500s left no trace: the raw connection error reached the error path, but pino's default serialization of the raw driver error could throw (circular refs / throwing getters), which drops the entire log line — so the failure was never written to `server.log` with a stack.
> - This PR makes the pool shed stale sockets and guarantees connection-level 500s are always logged with a stack.

## Linked Issues or Issue Description

No public GitHub issue exists; describing the bug in-PR (bug report form):

- **What happened:** Intermittent HTTP 500s on the only transactional project route (`POST /api/projects/:id/workspaces`) after the embedded Postgres briefly restarted/flapped. The error was a connection-level failure raised from inside `db.transaction()`. The same 500s were **not** written to `server.log`, so they were effectively invisible.
- **Expected behavior:** A transient DB flap should not produce lingering dead pooled connections; transactions should run on a healthy connection (or fail loudly and traceably), and every 5xx should be logged with a stack.
- **Steps to reproduce:** Run the control plane against embedded Postgres, restart/flap Postgres, then exercise a transactional route shortly after — an intermittent connection-level 500 can be served from a stale pooled socket, with nothing in `server.log`.
- **Deployment mode:** local trusted, `database.mode = embedded-postgres`.

Related open PRs (same `createDb` area, different angles — linked for reviewer context, not duplicates):
- #7481 — `fix(db): reducir max conexiones idle en createDb` (reduces max idle connections; complementary to the recycling timeouts here).
[...]

PR #4006, opened 2026-04-19
fix(db): add pool lifecycle config and DB health endpoint

usernamedJ:
## Summary

- Configures the postgres connection pool with `idle_timeout: 20s`, `max_lifetime: 30min`, and `connect_timeout: 30s` to prevent stale idle connections from crashing the server
- Adds a dedicated `GET /api/health/db` endpoint that runs `SELECT 1` and returns 503 when the database is unreachable, suitable for external watchdog monitoring

Fixes a crash where the server would die with `TypeError: Cannot read properties of null (reading 'write')` when the OS/DB dropped idle postgres connections, requiring manual operator restart.

## Test plan

- [ ] `curl localhost:3100/api/health/db` returns `{"status":"ok"}` when DB is healthy
- [ ] `curl localhost:3100/api/health/db` returns 503 when DB is down
- [ ] Server survives idle connection drops without crashing (manual test: leave server idle >20s, verify reconnection)
- [ ] Existing health endpoint still works

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #12956, opened 2026-09-07
fix(db): reap idle pool connections, name the pool, and end it on shutdown

stubbi:
## Thinking Path

> - The server keeps one postgres.js pool (`packages/db/src/client.ts`, `createDb`) for every query it runs. #10795 made the pool tunable from the environment, but the defaults stayed at the driver defaults: an idle connection never closes, the pool reports itself as `postgres.js`, and no code path ever calls `sql.end()`.
> - On a hosted Paperclip deployment the server entered a restart loop (a bundled plugin failure that #12953 describes made every run fail, and the pool saturated). Each generation opened its ten connections, died, and left the backends open on the PostgreSQL side until TCP keepalive reaped them hours later. After about 20 generations the backends exceeded `max_connections`, and every later boot died on its first bootstrap query with `sorry, too many clients already`, before `server.listen()`. The loop could not heal itself. #9555 describes the same shape on a launchd-supervised self-hosted install.
> - Three properties of the pool combine to make this possible: idle connections are never reaped, the pool is never ended on any exit path, and an operator cannot even find the leaked backends in `pg_stat_activity` because they carry the generic driver name.
> - This pull request gives the pool a 60 second idle timeout and the `paperclip` application name by default, exposes `max_lifetime` and `application_name` through the same `DATABASE_*` environment contract that #10795 introduced, and ends the pool on the orderly SIGINT/SIGTERM path and on the fail-loud startup path.
> - The benefit is that a restarting or crash-looping server releases its backends instead of accumulating them, and an operator can see and count Paperclip's connections.

- Refs #9555 — database connection pool leak causes an infinite restart loop under load. This PR closes the "pool never ends, idle connections never close" part of that report.
- Refs #12953 — hosted outage report. The pool exhaustion is the second half of that incident; the first half (a stuck sandbox provider plugin) has its own PR.
- Related prior PRs: #9597 and #8780 both propose hard-coded `idle_timeout` / `max_lifetime` values in `createDb`. Both predate #10795 (merged), which made these options environment-driven; this PR builds on the merged shape and adds the shutdown `end()` that neither covers. #4006 and #7481 are closed earlier attempts in the same area.

## What Changed

- `packages/db/src/client.ts`
  - New `resolveDatabaseClientOptions()` applies Paperclip defaults on top of the environment: `idleTimeoutSeconds` defaults to 60 (`DEFAULT_DATABASE_IDLE_TIMEOUT_SECONDS`) and `applicationName` to `paperclip` (`DEFAULT_DATABASE_APPLICATION_NAME`). `createDb` uses it for both the environment path and explicit options.
  - `DATABASE_IDLE_TIMEOUT_SECONDS` now accepts `0` to restore the driver default (keep idle connections open). Negative or non-integer values still throw.
  - New environment variables: `DATABASE_MAX_LIFETIME_SECONDS` (positive integer, maps to `max_lifetime`) and `DATABASE_APPLICATION_NAME` (non-empty string, maps to `connection.application_name`).
  - `postgresJsOptions()` maps the two new options.
- `server/src/shutdown.ts`
  - `finalizeServerShutdown` gains two optional ordered steps: `closeHttpListener` runs first, before the application services stop; `closeDatabase` runs after the application services and before the embedded PostgreSQL stop. A failure in either is logged and does not stop the teardown. Final order: listener → application services → database pool → embedded PostgreSQL → instrumentation → Sentry.
  - New `closeHttpListenerForShutdown()`: stops accepting requests, closes idle keep-alive sockets, waits up to 5 s for open connections, then closes whatever is left. Requests still in flight are drained while every service is available, and none can reach a route after `sql.end()`, on the signal path and the programmatic path alike (the programmatic path's later `server.close` finds the listener closed and skips).
- `server/src/app.ts`: the app shutdown hook (`shutdownAppServices`) now stops the plugin job scheduler, whose tick queries the database, so a programmatic `shutdown()` leaves no timer running against the ended pool.
- `server/src/index.ts`
  - `startServer()` is now a thin wrapper around the boot sequence. When the boot sequence throws after the pool exists, the wrapper ends the pool (and the separate migration pool, when configured) before it rethrows. This covers the `process.exit(1)` path in the main module and the CLI `paperclip run` path alike.
  - The orderly shutdown passes the same `closeDatabaseClients` to `finalizeServerShutdown`.
  - `endDatabaseClient` tolerates a client without `$client` (test doubles) and uses a 5 second end timeout.
- Docs: `docs/deploy/database.md` gets a "Connection Pool Settings" table with every `DATABASE_*` pool variable, its default and its effect; `doc/DATABASE.md` lists the two new variables.
- Tests
  - `packages/db/src/client-options.test.ts`: parsing of the new variables, `0` for the idle timeout, rejection of malformed values, driver option mapping, and the `resolveDatabaseClientOptions` defaults.
  - `packages/db/src/client.test.ts` (embedded PostgreSQL): `createDb(url)` reports `application_name = paperclip` for its own backend, and a pool with `idleTimeoutSeconds: 1` has zero backends in `pg_stat_activity` after the timeout.
  - `server/src/shutdown.test.ts`: the listener closes before the application services, and the database close runs between the application services and the embedded PostgreSQL stop; a failing database close is logged while the teardown still finishes; `closeHttpListenerForShutdown` closes idle sockets and resolves on close, force-closes after the grace period, and is a no-op when the listener was never bound.

## Verification

- `pnpm --filter @paperclipai/db typecheck` — passes (`check:migrations` + `tsc --noEmit`).
- `cd server && pnpm typecheck` — passes.
- `cd packages/db && pnpm exec vitest run src/client-options.test.ts src/client.test.ts src/client-teardown-registry.test.ts` — 9 + 18 + 3 tests pass (the `client.test.ts` cases need embedded PostgreSQL; the new one waits up to 10 s for the idle reap and passed in about 3 s).
- `cd server && pnpm exec vitest run src/shutdown.test.ts src/__tests__/server-startup-feedback-export.test.ts src/__tests__/bootstrap-claim-routes.test.ts` — 34 + 11 tests pass. The startup-feedback suite exercises `startServer()` with a mocked `createDb`, which is why `endDatabaseClient` tolerates a client without `$client`.
- Manual check for a reviewer: start the server against any PostgreSQL, then run `SELECT application_name, state, count(*) FROM pg_stat_activity GROUP BY 1, 2;`. Paperclip's backends now show `paperclip`. Leave the server idle for more than 60 s and the idle backends disappear. Send SIGTERM and the backends close before the process exits.

## Risks

- Behavior change with no environment set: idle pooled connections now close after 60 s. The next query after an idle period pays a reconnect (single-digit milliseconds on a local socket). postgres.js reconnects transparently. Set `DATABASE_IDLE_TIMEOUT_SECONDS=0` to keep the previous behavior.
- `application_name` changes from `postgres.js` to `paperclip`. Anything that filtered `pg_stat_activity` on the old name would need an update; nothing in this repo does.
- The HTTP listener now closes at the start of the final teardown (after the heartbeat run drain, which still needs the API for running agents). The pool close runs after the application services. A late query from a timer that survived the service shutdown would fail with a driver "connection ended" error instead of running; the known database-backed timer (the plugin job scheduler) is now stopped in the service shutdown.
- The listener drain adds at most 5 s to a shutdown while long-lived connections (for example WebSocket clients) are open; after that they are closed forcibly.
- `startServer()` is split into a wrapper and the boot sequence. The exported signature and return type are unchanged.
- No migration, no schema change.

## Model Used

- Claude Fable 5.1 (`claude-fable-5-1`) via Claude Code, extended thinking, tool use (file edits, shell, test runs). The change was produced with the model and reviewed by the submitting human.

## Checklist

https://claude.ai/code/session_014t3bi2beVNVVHAxK36dmXm

### DIFF
--- a/doc/DATABASE.md
+++ b/doc/DATABASE.md
@@ -122,8 +122,10 @@ All of these are optional; when unset, the driver defaults apply and behavior is
 ```sh
 DATABASE_PREPARED_STATEMENTS=false   # required for transaction-mode poolers; default: enabled
 DATABASE_POOL_MAX=25                 # connection pool size; default: 10
-DATABASE_IDLE_TIMEOUT_SECONDS=60     # close idle pooled connections; default: keep open
+DATABASE_IDLE_TIMEOUT_SECONDS=60     # close idle pooled connections; default: 60 (0 = keep open)
 DATABASE_CONNECT_TIMEOUT_SECONDS=10  # default: 30
+DATABASE_MAX_LIFETIME_SECONDS=1800   # recycle a pooled connection after this long; default: 30-60 min (random)
+DATABASE_APPLICATION_NAME=paperclip  # application_name in pg_stat_activity; default: paperclip
 ```
 
 ### Push the schema
--- a/docs/deploy/database.md
+++ b/docs/deploy/database.md
@@ -62,7 +62,22 @@ If using connection pooling (transaction mode), disable prepared statements via 
 DATABASE_PREPARED_STATEMENTS=false
 ```
 
-Related optional client tuning (driver defaults apply when unset): `DATABASE_POOL_MAX`, `DATABASE_IDLE_TIMEOUT_SECONDS`, `DATABASE_CONNECT_TIMEOUT_SECONDS`.
+Related optional client tuning: `DATABASE_POOL_MAX`, `DATABASE_IDLE_TIMEOUT_SECONDS`, `DATABASE_CONNECT_TIMEOUT_SECONDS`, `DATABASE_MAX_LIFETIME_SECONDS`, `DATABASE_APPLICATION_NAME`. Driver defaults apply when unset, except that idle pooled connections close after 60 seconds (`DATABASE_IDLE_TIMEOUT_SECONDS=0` keeps them open) and the pool reports `application_name=paperclip`. See [Connection pool settings](#connection-pool-settings).
+
+## Connection Pool Settings
+
+The server opens one postgres.js pool for its own queries (and a second one when `DATABASE_MIGRATION_URL` points at a different connection). Every setting is optional:
+
+| Variable | Default | Effect |
+|----------|---------|--------|
+| `DATABASE_POOL_MAX` | `10` (driver) | Maximum pooled connections. |
+| `DATABASE_IDLE_TIMEOUT_SECONDS` | `60` | Close a pooled connection after this much idle time. `0` keeps idle connections open forever (the driver default). |
+| `DATABASE_CONNECT_TIMEOUT_SECONDS` | `30` (driver) | Give up on a connection attempt after this long. |
+| `DATABASE_MAX_LIFETIME_SECONDS` | 30–60 min, randomized (driver) | Recycle a pooled connection once it is this old. |
+| `DATABASE_APPLICATION_NAME` | `paperclip` | Value of `application_name` in `pg_stat_activity`, so you can find Paperclip's backends: `SELECT * FROM pg_stat_activity WHERE application_name = 'paperclip';` |
+| `DATABASE_PREPARED_STATEMENTS` | `true` (driver) | Set `false` behind a transaction-mode pooler (see above). |
+
+The server ends its pools during shutdown (SIGINT/SIGTERM) and when startup fails after the pool was opened, so a restarting server does not leave idle backends behind. Size `max_connections` on the PostgreSQL side for at least `DATABASE_POOL_MAX` per server process plus your other clients.
 
 ## Switching Between Modes
 
--- a/packages/db/src/client-options.test.ts
+++ b/packages/db/src/client-options.test.ts
@@ -1,5 +1,11 @@
 import { describe, expect, it } from "vitest";
-import { databaseClientOptionsFromEnv, postgresJsOptions } from "./client.js";
+import {
+  DEFAULT_DATABASE_APPLICATION_NAME,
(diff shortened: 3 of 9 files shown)


stubbi:
@greptileai re-review please — round 1, both P1s addressed in 91128e036:
- P1 Pool Closes Before Listener → `finalizeServerShutdown` now runs a `closeHttpListener` step before `closeDatabase`; the new `closeHttpListenerForShutdown` stops accepting, closes idle keep-alive sockets, and force-closes the rest after a 5 s grace period, on both the signal and the programmatic path (the later programmatic `server.close` finds the listener closed and skips). Unit-tested in `shutdown.test.ts` (ordering, drain timeout, not-listening).
- P1 Scheduler Outlives Database Pool → the app shutdown hook (`shutdownAppServices` in `app.ts`) now calls `scheduler.stop()` inside the awaited teardown, before the pool ends.

stubbi:
@greptileai re-review please — round 2, addressed in 2456ed2aa:
- P1 Requests During Service Teardown → `finalizeServerShutdown` now closes and drains the HTTP listener first, before `shutdownAppServices`, so in-flight requests complete against a fully available server; then services stop, then the pool ends, then embedded PostgreSQL stops. `shutdown.test.ts` asserts the new order (`listener:close` → app services → `database:close` → `postgres:stop`).
- Scheduler Outlives Database Pool → already fixed in 91128e036 (`scheduler.stop()` inside the awaited app teardown); thread resolved.

nickyleach:
Thanks @stubbi!

PR #13015, opened 2026-09-07
feat(costs): attribute token spend per task and per routine

Brend-growthway:
## What

Adds the per-task and per-routine cuts of token consumption, plus honest reporting of runs that recorded no usage.

Follows [#12849](https://github.com/paperclipai/paperclip/pull/12849) (by-project counted each cost event once). This PR is disjoint from it: no shared hunks, and `CostByProject` is untouched here.

## Why

Summing `GET /api/issues/{id}/runs` to get a per-task total does not work. That route hands the full run — `usageJson` included — to **every** issue the run touched. In one company, 337 runs appeared in more than one issue and a single run appeared in **61**, inflating the company total from 2.5 B to 4.7 B tokens (**+87%**).

There was also no way to see a routine's cost at all: each firing opens its own issue, so a daily routine's spend scatters across dozens of unrelated-looking tasks.

## Changes

- **`GET /api/companies/:companyId/costs/by-issue`** — tokens and cost per task, `from`/`to` window. One row per run, attributed to the single issue resolved into `cost_events.issue_id` at finalization, so rows sum to the same universe as the by-agent cut instead of double counting.
- **`GET /api/companies/:companyId/costs/by-routine`** — rolled up across every firing, counted through each firing issue's whole subtree, because work a firing delegates to a child is still that routine's cost.
- **Unmetered runs are classified, not hidden.** Runs with no `usageJson` are split into `stranded`, `neverRan`, and `lost` and reported in `summary` rather than silently dropped, so a total reads as a total and not as a floor.
- **Costs page** surfaces both cuts.

## Honest limits

- Runs whose owning issue cannot be resolved are **omitted from the rows and reported in `summary`** rather than folded into an arbitrary task. The remainder stays visible.
- `costs/summary` still reports `spendCents: 0` under `subscription_included` billing by design; the real figure lives in `subscriptionCostUsd`/`totalTokens`, which these routes return.

## Verification

- `server/src/__tests__/costs-service.test.ts` — +404 lines covering the double-count case, the window filter, and the unmetered split.
- `server/src/__tests__/heartbeat-cost-accounting.test.ts` — asserts a run finalizes onto exactly one issue.

### DIFF
--- a/packages/shared/src/index.ts
+++ b/packages/shared/src/index.ts
@@ -1223,6 +1223,8 @@ export type {
   CostByAgentModel,
   CostWindowSpendRow,
   CostByProject,
+  CostByIssue,
+  CostByRoutine,
   FinanceEvent,
   FinanceSummary,
   FinanceByBiller,
--- a/packages/shared/src/types/cost.ts
+++ b/packages/shared/src/types/cost.ts
@@ -27,6 +27,28 @@ export interface CostSummary {
   spendCents: number;
   budgetCents: number;
   utilizationPercent: number;
+  inputTokens: number;
+  cachedInputTokens: number;
+  outputTokens: number;
+  totalTokens: number;
+  /** cents actually billed through a metered api — the only spend that hits a card */
+  meteredCostCents: number;
+  /**
+   * dollar value of subscription usage. `spendCents` stays 0 for it by design
+   * (the plan already paid), so this is the only figure that shows the account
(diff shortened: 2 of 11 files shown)


Brend-growthway:
Reopening from a branch whose name carries no internal ticket id, per the PR checklist. Same commits. Superseded by the PR that follows.

PR #9615, opened 2026-07-15
feat(costs): expose paginated cost events with usage basis

broken-coder:
## Thinking Path

> - Paperclip is the open source control plane people use to manage AI agents and their operating costs
> - Cost events already preserve normalized usage and distinguish subscription from metered billing
> - Aggregated cost reports do not preserve the boundary of each underlying event
> - External equivalent-cost tools need that boundary to avoid applying request-sensitive pricing rules to run-level aggregates
> - This pull request adds an explicit usage basis and a company-scoped, cursor-paginated CostEvent read API
> - The benefit is a pricing-neutral data contract that supports independent analysis without changing native spend, budgets, or the Costs UI

## Linked Issues or Issue Description

- Refs: #339
- Refs: #498
- Duplicate search found no open pull request adding `usageBasis` or a company CostEvent list endpoint.

## What Changed

- Add `usageBasis` (`per_request`, `per_run`, or `unknown`) to the shared CostEvent contract and database schema, with `unknown` as the compatibility default.
- Mark heartbeat-generated CostEvents as `per_run` only after session-cumulative usage has been normalized to a run delta.
- Add `GET /api/companies/:companyId/cost-events` with existing Costs read authorization, inclusive date filters, repeatable billing-type filters, bounded limits, and opaque keyset cursors.
- Document the endpoint and usage-boundary semantics in OpenAPI and the agent developer guides.
- Add validator, migration, cursor, service, heartbeat, authorization, pagination, and OpenAPI regression coverage.

## Verification

- `pnpm exec vitest run packages/shared/src/validators/cost.test.ts packages/db/src/cost-usage-basis-migration.test.ts server/src/services/cost-event-cursor.test.ts server/src/__tests__/heartbeat-cost-accounting.test.ts server/src/__tests__/heartbeat-cost-events.test.ts server/src/__tests__/costs-service.test.ts server/src/__tests__/openapi-routes.test.ts` (`52 passed`, `11 skipped`)
- `pnpm typecheck` (passed across 31 of 32 workspace projects, including migration numbering and safety checks)
- `git diff --check origin/master...HEAD` (passed)
[...]

PR #13016, opened 2026-09-07
feat(costs): attribute token spend per task and per routine

Brend-growthway:
## Thinking Path

> - Agent runs consume tokens. The `costs` service aggregates that spend and the Costs page shows it
> - The service has a per-agent cut and a per-project cut. It has no per-task cut and no per-routine cut
> - Operators tried to build a per-task total from `GET /api/issues/{id}/runs`. That route gives the full run record, `usageJson` included, to every issue the run touched. The sum counts the same run many times
> - In one company, 337 runs were present in more than one issue. One run was present in 61 issues. The total went from 2.5 B tokens to 4.7 B tokens, an increase of 87%
> - A routine has a second problem. Each firing opens a new issue, so the cost of one routine is divided across many tasks that look unrelated
> - This pull request adds a per-task cut and a per-routine cut. Each run is counted one time, against the one issue that owns it
> - The benefit is that an operator can find the most expensive tasks and routines from the API, and the rows add up to the company total

No public issue exists. The problem is described here.

**What happened:**
There is no API that reports token spend per task or per routine. Operators build the number by hand from `GET /api/issues/{id}/runs`, and the result is too high.

**Expected behavior:**
The sum of spend over all tasks is equal to the company total, within a stated margin. The spend of one routine is one number.

**Actual behavior:**
`GET /api/issues/{id}/runs` returns the complete run, with `usageJson`, for every issue that the run touched. A run that touched 61 issues is counted 61 times. The company total increased by 87%.

A second defect makes every total a floor. 245 of 1604 runs had `usageJson: null`. These runs are silently absent, so a total reads as complete when it is not.

**Impact:**
An operator cannot find the most expensive work. A cost report is too high by an unknown amount, and it is also missing an unknown amount.

## What Changed

- Add `GET /api/companies/:companyId/costs/by-issue`. It returns tokens and cost per task, with a `from`/`to` window. One row per run. The run is attributed to the one issue that `cost_events.issue_id` resolves at finalization, so the rows add up to the same total as the per-agent cut.
- Add `GET /api/companies/:companyId/costs/by-routine`. It rolls up all firings of one routine. It counts each firing issue and its subtree, because work that a firing delegates to a child is still the cost of that routine.
- Classify runs that recorded no usage. The `summary` field reports them as `stranded`, `neverRan`, and `lost`. Before this change they were absent and invisible.
- Add both cuts to the Costs page.
- Add the `CostByIssue` and `CostByRoutine` types to `packages/shared`.

This pull request follows #12849, which makes `by-project` count each cost event one time. The two changes are disjoint. They share no hunk, and `CostByProject` is not changed here.

## Verification

Run the two test files:

```
cd server
pnpm vitest run src/__tests__/costs-service.test.ts src/__tests__/heartbeat-cost-accounting.test.ts
```

Result on this branch, rebased on `d8b95805`: **33 tests passed, 2 files, 0 failures.**

The tests cover:
- The double-count case. A run that touches many issues is counted one time.
- The `from`/`to` window filter.
- The split of unmetered runs into `stranded`, `neverRan`, and `lost`.
- `heartbeat-cost-accounting.test.ts` asserts that a run finalizes onto exactly one issue.

## Risks

Low risk. The change adds two read-only routes and does not change an existing route or response shape.

- No migration. No schema change.
- `CostByProject` and the `by-project` route are not changed.
- One behavior does shift: runs that resolve to no owning issue are left out of the `by-issue` rows and reported in `summary` instead. This is deliberate. Folding them into an arbitrary task would report a false number for that task.
- Honest limit: `costs/summary` still returns `spendCents: 0` for `subscription_included` billing. That is by design in the billing model, and it is out of scope here. The real figure is in `subscriptionCostUsd` and `totalTokens`, which these routes return.

## Model Used

Claude Opus 5 (`claude-opus-5`), Anthropic. Extended thinking on, with tool use and code execution. Run through the Claude Agent SDK as an automated engineering agent.

## Checklist

- [x] I searched the GitHub PR list for similar or duplicate PRs and confirmed this is not a duplicate

### Dedup search

I searched the PR list before opening this one. The related PRs I found are these, and none of them adds a per-task or per-routine cut:

- #12849 — `by-project` counts each cost event one time. This is my own prior PR. This one follows it and is disjoint.
- #11538 — estimates subscription-billed run cost from token usage. It changes pricing, not attribution.
- #10544 — prices subscription runs from a rate card. Pricing, not attribution.
- #11424 — adds `creditRunCount`/`otherRunCount` to the by-agent, by-provider, and by-biller cuts. It does not add a new cut.
- #9615 — exposes paginated cost events. It is a raw event feed, not an aggregate.

### DIFF
--- a/packages/shared/src/index.ts
+++ b/packages/shared/src/index.ts
@@ -1223,6 +1223,8 @@ export type {
   CostByAgentModel,
   CostWindowSpendRow,
   CostByProject,
+  CostByIssue,
+  CostByRoutine,
   FinanceEvent,
   FinanceSummary,
   FinanceByBiller,
--- a/packages/shared/src/types/cost.ts
+++ b/packages/shared/src/types/cost.ts
@@ -27,6 +27,28 @@ export interface CostSummary {
   spendCents: number;
   budgetCents: number;
   utilizationPercent: number;
+  inputTokens: number;
+  cachedInputTokens: number;
+  outputTokens: number;
+  totalTokens: number;
+  /** cents actually billed through a metered api — the only spend that hits a card */
+  meteredCostCents: number;
+  /**
+   * dollar value of subscription usage. `spendCents` stays 0 for it by design
+   * (the plan already paid), so this is the only figure that shows the account
+   * burning through its quota before a lockout.
+   */
+  subscriptionCostUsd: number;
+  eventCount: number;
+  runCount: number;
+  /** runs whose cost event carries no price — usage known, dollars unknown */
+  unpricedRunCount: number;
+  /** runs that produced no cost event at all */
+  unmeteredRunCount: number;
+  strandedRunCount: number;
+  strandedTokens: number;
+  neverRanRunCount: number;
+  lostRunCount: number;
 }
 
 export interface IssueCostSummary {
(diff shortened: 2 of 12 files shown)


Brend-growthway:
### Merge-order hazard with #12849 — fixed here in `ac1d59dd`

These two PRs both edit `server/src/services/costs.ts` and git merges them **without a conflict**, because neither touches the other's lines. The merged tree did not compile:

```
costs.ts(133,5): error TS2552: Cannot find name 'isNotNull'. Did you mean 'isNull'?
costs.ts(696,9): error TS2552: Cannot find name 'isNotNull'. Did you mean 'isNull'?
```

Cause: #12849 removes the last remaining use of `isNotNull` (the `by-project` `activity_log` join) and drops it from the shared `drizzle-orm` import line, while this PR added two new uses of that binding. Each PR is green on its own; only the combination breaks, and no CI run on either branch would catch it.

Fixed on this side rather than on #12849, so merge order does not matter: both call sites now use raw SQL null checks and this file no longer depends on an import line the other PR owns.

Verified:

- Rendered SQL is byte-identical via `PgDialect.sqlToQuery` — `isNotNull(t.finishedAt)` and ``sql`${t.finishedAt} is not null` `` both emit `"heartbeat_runs"."finished_at" is not null`.
- Merged #12849 into this branch locally: clean merge, and `Cannot find name` is gone.
- `server` typecheck error count unchanged at 222 before and after the commit (all pre-existing, from unbuilt workspace deps); zero in `services/costs.ts` or `routes/costs.ts`.
- `costs-service.test.ts` 27/27 pass, `heartbeat-cost-accounting.test.ts` 13/13 pass.

No behaviour change and no new test — this is purely removing a build-time coupling between two in-flight PRs.

Brend-growthway:
Pushed `78f895eb` — a correction to a **justification** in this PR, not to its behaviour.

The doc comment on `NO_MODEL_WORK_ERROR_CODES` claimed:

> Zero of 1.021 `succeeded` runs are unpriced, which is why this split is safe: success always accounts.

I re-measured against the live company database today (2.378 runs, up from 1.696 when I first wrote that line). **That invariant is false: 8 of 1.478 `succeeded` runs carry no `usage_json`.**

The code was already right — the split keys off `error_code`, never `status`, so those 8 land in `lost` and stay visible. But the *reason* given for the design was wrong, and a wrong reason is exactly what invites a later "simplification" to a status check.

### The shape of the 8

Among `succeeded` runs, `result_json IS NULL` and `usage_json IS NULL` hold on **exactly the same 8 rows** (8 = 8 = 8). That is not a coincidence: `heartbeat.ts` persists usage and result in a single `setRunStatusIfRunning` write, which is skipped wholesale when the run already left `running`. So it is not "usage failed to persist" — the entire finalization write never lands.

I also checked the ledger side: `usage_json IS NULL` ⟺ **zero** `cost_events`, with no exceptions across all 334 unbilled runs. Usage and ledger agree, so these endpoints inherit exactly the declared gap and nothing hidden.

### Changes

- Replaced the false claim with the measured shape and the mechanism.
- Added a regression test: a `succeeded` run with no usage and no error code must count as `lost`, not `neverRan`.

### Verified, not assumed

The test is a real guard, not a tautology — I patched the classifier to excuse succeeded runs (`status = 'succeeded' or ...`), confirmed the new test **fails** (`expected 1, got 0`), then reverted and confirmed it passes.

`costs-service.test.ts`: **28/28**.

PR #13053, opened 2026-09-08
fix(adapter-utils): discover gh.exe/git.exe in Windows GitHub launchers

FC4b:
## Problem

On Windows, the per-run GitHub runtime wrapper (`%TEMP%/paperclip-github-runtime/<runId>/gh`) could not discover the real binary and every invocation exited 127 ("Paperclip: requested GitHub command is not installed"). Additionally, Git Bash sessions inherited a single Windows-style semicolon-joined PATH string, so `curl`/`node`/`grep` were all "command not found" until the agent manually re-exported a POSIX PATH.

## Root causes

- The shim resolved the real binary with `path.join(p, program)` using the bare program name only, but Windows installs ship `gh.exe`/`git.exe` only.
- The shim split `PATH` on `path.delimiter` alone, so a colon-joined POSIX PATH handed over by Git Bash never resolved.
- `prepareGitHubOperationLaunchers` joined the managed PATH with a literal `:` even on local Windows runs, producing `C:\...\runId;...`-style hybrid strings that broke both native PATH splitting and POSIX shell profiles.
- The staged shell profiles re-exported that raw Windows string, clobbering Git Bash's converted PATH at startup.

## Fix

- Shim discovery now tries PATHEXT-style candidates (`.com/.exe/.bat/.cmd`, honoring `PATHEXT` order) before the bare name, splits inherited PATH with the delimiter the value actually uses, compares the launcher directory case-insensitively on Windows, and restores the same PATH shape for child processes.
- Local Windows staging joins the managed PATH with `path.delimiter` (`;`), stages Git Bash-style POSIX profiles (`/c/...` entries), and stages `gh.cmd`/`git.cmd` peers so cmd.exe and PowerShell can also invoke the brokered shims.
- Remote/sandbox targets keep the POSIX `:` join and are unchanged.

## Testing

- New win32-scoped integration tests in `github-launcher.test.ts`: PATHEXT discovery of `gh.exe` end-to-end through the credential broker, colon-joined POSIX-style PATH resolution, `cmd.exe → gh.cmd → shim → gh.exe` chain, and the preserved 127 diagnostic.
- Existing POSIX tests unchanged (now skipped on win32 where extension-less shebang exec never worked); `github-launcher-environment.test.ts` remote-path behavior unchanged.
- Verified locally: 4/4 new Windows tests pass; `tsc --noEmit` clean for `@paperclipai/adapter-utils`.

### DIFF
--- a/packages/adapter-utils/src/execution-target.ts
+++ b/packages/adapter-utils/src/execution-target.ts
@@ -3,7 +3,7 @@ import net from "node:net";
 import os from "node:os";
 import path from "node:path";
 import { randomBytes, randomUUID } from "node:crypto";
-import { githubLauncherSource } from "./github-launcher.js";
+import { githubLauncherCommandSource, githubLauncherSource } from "./github-launcher.js";
 import type { SshRemoteExecutionSpec } from "./ssh.js";
 import {
   prepareCommandManagedRuntime,
@@ -1502,6 +1502,15 @@ export function runtimeAssetDir(
   return prepared.assetDirs[key] ?? path.posix.join(fallbackRemoteCwd, ".paperclip-runtime", key);
 }
 
+/** Git Bash maps `C:\dir` to `/c/dir`, so POSIX profiles staged on Windows use that form. */
+function gitBashPathList(value: string): string {
+  return value.split(";").map((entry) => {
+    if (/^[A-Za-z]:[\\/]/.test(entry)) return `/${entry[0].toLowerCase()}${entry.slice(2).replace(/\\/g, "/")}`;
+    if (entry.startsWith("\\\\")) return `//${entry.slice(2).replace(/\\/g, "/")}`;
+    return entry.replace(/\\/g, "/");
+  }).join(":");
+}
+
 type GitHubLauncherLocation = {
   runId: string; target: AdapterExecutionTarget | null | undefined;
 };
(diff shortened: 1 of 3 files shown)


Brend-growthway:
Independently hit the same `exit 127` on Windows and landed on the same root cause, so this fix matches what I measured. One follow-up on the discovery side that I think is still open here.

`.CMD` is in the `PATHEXT` candidate list, but `spawn` on line 93 is still called on the resolved path directly. Node refuses to execute `.bat`/`.cmd` without a shell (the hardening from the `CVE-2024-27980` fix), so when the resolved target *is* a batch shim, discovery now succeeds and the spawn fails instead.

Measured on Node `v24.15.0`, Windows 11, using this PR's own resolution logic to pick the target:

```
resolved=...\gh.cmd
direct:  err=EINVAL  status=null  out=""
comspec: err=none    status=0     out="CMD-RAN args=..."
```

Reproduction — a `gh.cmd` containing `@echo CMD-RAN args="%*"` placed on `PATH`, then:

```js
spawnSync(executable, ['a b'])                      // EINVAL
spawnSync(process.env.COMSPEC, ['/d','/s','/c',
  '"' + '"' + executable + '" "a b"' + '"'],
  { windowsVerbatimArguments: true })               // status 0
```

This is not a rare shape: npm- and scoop-installed CLIs ship exactly that kind of `.cmd` shim, and `gh` is commonly installed through scoop. The user-visible result is `Paperclip: GitHub command could not start.` rather than `exit 127` — a different message for the same blocked operation.

The new `runs the staged .cmd wrapper from cmd.exe` test does not cover it, because it exercises the opposite direction: `cmd.exe` invoking a staged `.cmd`, where `cmd.exe` is the one doing the execution. The failing case is the launcher itself resolving a `.cmd` as the *target* and spawning it from Node.

A minimal delta on top of this PR, keeping the existing behaviour for `.exe`:

```js
const batch = windows && /\.(bat|cmd)$/i.test(executable);
const quote = (value) => '"' + String(value).replace(/(\\*)"/g, '$1$1\\"').replace(/(\\*)$/, '$1$1') + '"';
const command = batch ? (env.COMSPEC || 'cmd.exe') : executable;
const args = batch
  ? ['/d', '/s', '/c', '"' + [executable, ...process.argv.slice(2)].map(quote).join(' ') + '"']
  : process.argv.slice(2);
const child = spawn(command, args, { env, stdio: 'inherit', windowsVerbatimArguments: batch });
```

The explicit quoting matters for argument fidelity — with `windowsVerbatimArguments` the shell does no quoting of its own. Round-trip check through a `.cmd` shim that echoes `process.argv`:

```
expected=["clone","--depth","1","a b","c\"d"]
received=["clone","--depth","1","a b","c\"d"]
```

Separately, one data point on the `os.devNull` question, since it comes up with this file: keeping the literal `/dev/null` for `GIT_CONFIG_GLOBAL` looks correct. Git for Windows accepts `/dev/null` and does suppress global/system config, whereas `os.devNull` is `\\.\nul`, which Git rejects with `fatal: unable to access '//./nul': Invalid argument`.

Happy to leave it here as a note for whoever picks the PR up.

PR #13056, opened 2026-09-08
fix(cross-issue-limit): allow heartbeat_timer runs to write to checked-out issues (INUA-6799)

rotem-zecharia:
## Thinking Path

> - The heartbeat scheduler fires timer-based agent runs with no `PAPERCLIP_TASK_ID`; these runs are the primary mechanism for agents scanning and advancing their work queue
> - After a timer run successfully checks out an issue, all subsequent writes to that issue returned `403 cross_issue_influence_run_context_required`
> - The cross-issue influence guard (`observeCrossIssueInfluence`) reads `sourceIssueId` from the run's `contextSnapshot`; timer runs start with an empty snapshot so `sourceIssueId` is null, and the guard threw 403 unconditionally — without consulting `checkoutRunId`
> - Checkout explicitly sets `checkoutRunId` on the issue, which is the canonical ownership signal, but the guard never consulted it for the null-context case
> - This PR adds a second path in the guard: when `sourceIssueId` is null, query `issues.checkoutRunId` inside the same transaction using `SELECT … FOR UPDATE`; if the run owns the checkout, return `null` (same-issue ownership, no cap); otherwise still throw 403
> - A related recovery change extends the stall-recovery sweep to include issues blocked with no assignee but with `createdByAgentId`, so the backstop can wake the creator and route the issue back to an agent
> - The benefit is that heartbeat timer runs can scan, check out, and advance issues normally without requiring a pre-set `PAPERCLIP_TASK_ID`

**What happened?**

When a `heartbeat_timer` run fires with no `PAPERCLIP_TASK_ID`, the agent can successfully check out an issue (`checkoutRunId` is set), but every subsequent PATCH or POST to that issue returns `HTTP 403 cross_issue_influence_run_context_required`. The run's `contextSnapshot` has no `issueId` because no task triggered it, so `readRunSourceIssueId` returns null and `observeCrossIssueInfluence` throws immediately — without ever checking whether the run already owns the issue via `checkoutRunId`.

**Expected behavior**

After a successful checkout, the run should be permitted to write to the checked-out issue. A run that holds `checkoutRunId` on the target issue is in a same-issue ownership relationship and must not be rejected by the cross-issue cap.

**Steps to reproduce**

1. Start a `heartbeat_timer` run for any agent (no `PAPERCLIP_TASK_ID` injected)
2. Call `POST /api/issues/{id}/checkout` from the run — succeeds, `checkoutRunId` is set
3. Call `PATCH /api/issues/{id}` — returns `403 cross_issue_influence_run_context_required`

## What Changed

- **`server/src/services/cross-issue-influence-limit.ts`**: when `sourceIssueId` is null, the guard now queries `issues.checkoutRunId` inside the existing transaction using `SELECT … FOR UPDATE`. A matching `checkoutRunId` returns `null` (same-issue path, no cap). The `FOR UPDATE` lock holds until the transaction completes, preventing a concurrent checkout release or transfer from invalidating the check. Runs with neither a source issue nor a checkout still get 403 (fail-closed preserved).
- **`server/src/__tests__/cross-issue-influence-limit.test.ts`**: updated mock to match the `for("update").then()` query shape; all 11 tests still pass.
- **`server/src/services/recovery/service.ts`**: expanded the blocked-task stall-recovery candidate query to include issues with `assigneeAgentId IS NULL` when `createdByAgentId IS NOT NULL`. A null-assignee checkout allows any agent to claim the task, so waking the creator is a safe fallback that gives the issue back to an active agent.
- **`server/src/__tests__/heartbeat-issue-liveness-escalation.test.ts`**: integration test verifying that a null-assignee dependent wakes its creator agent when all blockers resolve.

## Verification

```sh
# From repo root
npx vitest run server/src/__tests__/cross-issue-influence-limit.test.ts
# 11/11 pass
```

## Risks

Low. The `SELECT … FOR UPDATE` on the issues row is brief (inside the existing transaction) and only fires for null-context runs — the PAPERCLIP_TASK_ID path is completely unchanged. The stall-recovery extension adds rows to an existing query; it does not change any write path and is guarded by the existing `agentId` null-check before the wake fires.

## Model Used

Claude Sonnet 4.6 (Anthropic, `claude-sonnet-4-6`) via Paperclip agent adapter

## Checklist

### DIFF
--- a/server/src/__tests__/cross-issue-influence-limit.test.ts
+++ b/server/src/__tests__/cross-issue-influence-limit.test.ts
@@ -25,10 +25,12 @@ function counterDb(
             };
           }
           if (Object.keys(selection).includes("checkoutRunId")) {
-            // Issue checkout-ownership query (no .for() — plain .then())
+            // Issue checkout-ownership query (.for("update").then())
             return {
-              then: (resolve: (rows: unknown[]) => unknown) =>
-                resolve(issueCheckedOutByRun ? [{ checkoutRunId: issueCheckedOutByRun }] : []),
+              for: () => ({
+                then: (resolve: (rows: unknown[]) => unknown) =>
+                  resolve(issueCheckedOutByRun ? [{ checkoutRunId: issueCheckedOutByRun }] : []),
+              }),
             };
           }
           // heartbeatRuns lock query (.for("update").then())
--- a/server/src/__tests__/cross-issue-influence-limit.test.ts
+++ b/server/src/__tests__/cross-issue-influence-limit.test.ts
@@ -10,6 +10,7 @@ import {
 function counterDb(
   initialCount = 0,
   runOverrides: Record<string, unknown> | null = {},
+  issueCheckedOutByRun: string | null = null,
 ) {
   let observedCount = initialCount;
   const inserted: Array<Record<string, unknown>> = [];
@@ -18,10 +19,19 @@ function counterDb(
       from: () => ({
         where: () => {
           if (Object.keys(selection).includes("count")) {
+            // Activity-log count query
             return {
               then: (resolve: (rows: unknown[]) => unknown) => resolve([{ count: observedCount }]),
             };
           }
+          if (Object.keys(selection).includes("checkoutRunId")) {
+            // Issue checkout-ownership query (no .for() — plain .then())
+            return {
+              then: (resolve: (rows: unknown[]) => unknown) =>
+                resolve(issueCheckedOutByRun ? [{ checkoutRunId: issueCheckedOutByRun }] : []),
+            };
+          }
+          // heartbeatRuns lock query (.for("update").then())
           return {
             for: () => ({
               then: (resolve: (rows: unknown[]) => unknown) => resolve(runOverrides === null ? [] : [{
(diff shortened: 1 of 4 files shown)


rotem-zecharia:
**APPROVE** — [INUA-6800](https://linear.app/inu-il/issue/INUA-6800) CodeReviewer verdict

## Spec-condition checklist

| # | Condition | Status | Evidence |
|---|-----------|--------|----------|
| 1 | heartbeat_timer run can checkout + write without 403 | **met** | `cross-issue-influence-limit.ts:110–131`: when `sourceIssueId` is null, queries `issues.checkoutRunId == runId`; match → return `null` (same-issue path, no 403) |
| 2 | Existing PAPERCLIP_TASK_ID-based run scoping unaffected | **met** | New branch is inside `if (!sourceIssueId)` — runs with a populated contextSnapshot never enter it |
| 3 | Regression test covers this case | **met** | `cross-issue-influence-limit.test.ts:224–246`: asserts `resolves.toBeNull()` when checkout matches; tightened existing test confirms 403 still fires with no checkout |

## Assessment

**Root cause fixed** (not symptom): the 403 fired because `readRunSourceIssueId` returned null for heartbeat_timer runs and the guard threw unconditionally. The fix adds the correct secondary path: ownership via `checkoutRunId`. Fail-closed preserved for runs with neither source issue nor checkout.

**Correctness**: the extra query runs inside the existing `tx` (Drizzle), so no extra round-trip and no TOCTOU gap on the checkout ownership check. Mock discriminator (`checkoutRunId` key → plain `.then()` vs. `.for("update").then()`) matches the actual query shape.

**Bundled recovery/service.ts change**: in scope — null-assignee `createdByAgentId` fallback addresses the downstream consequence (heartbeat_timer run parks issue as `blocked` with no assignee → backstop could not route wake). Safe extension with integration test coverage.

**Blast radius**: narrow. No auth, payments, or notification changes.

---

Note: GitHub App lacks write access to post native reviews or set commit statuses on this repo. Verdict delivered here as required comment.

PR #12740, opened 2026-09-02
fix(codex-local): sync Codex credential copy-back to the selected account home

nickyleach:
## Thinking Path

> - Paperclip is the open source app people use to manage AI agents for work
> - The Codex local adapter manages credentials for remote Codex runs
> - A Codex login copied credentials to one fixed host home
> - A second account could overwrite the first account credential
> - This pull request copies credentials to the selected account home and checks the target path
> - The benefit is account isolation with a bounded copy-back path

## Linked Issues or Issue Description

This pull request addresses a bug in Codex credential copy-back.

**What happened?**

A remote Codex run copied refreshed credentials to one fixed host home. A second account could overwrite the first account credential.

**Expected behavior**

The copy-back writes to the account home selected by `CODEX_HOME`, or to the company default home when no account home exists. The write stays inside the company's managed directory.

**Steps to reproduce**

1. Run Codex authentication for two accounts on one host.
2. Select a different `CODEX_HOME` for each account.
3. Complete authentication for both accounts.
4. Check each account home for its credential file.

**Paperclip version or commit**

`6b60a8aafb2596b1322babac686f2d114b8d5e35`.

**Deployment mode**

Built from source with the Codex local adapter.

Related pull request: #10005 handles a missing shared host credential store. This pull request handles account-home selection and path containment.

## What Changed

- Add a managed credential-home guard in `packages/adapter-utils`.
- Select the account home for Codex credential copy-back.
- Reject missing, non-directory, symbolic-link, and out-of-tree company paths.
- Re-check existing path segments before the write.
- Create staging files with exclusive and no-follow flags.
- Add focused adapter and utility tests.

## Verification

- `codex-auth-copyback.test.ts`: 20 of 20 tests pass.
- `managed-credential-home.test.ts`: 11 of 11 tests pass.
- The full `codex-local` package suite passes: 377 of 377 tests.
- TypeScript checks pass for both changed packages.
- `git diff --check` passes.
- Continuous integration and Greptile checks must pass before merge.

## Risks

The final path check and file creation cannot run as one atomic operation because Node.js does not expose the required descriptor-relative calls. The code records this residual risk. The guard limits writes to the company's managed directory and uses no-follow checks and exclusive file creation.

## Model Used

OpenAI Codex, GPT-5, tool use and code execution enabled. The implementation author used this model to assist with repository changes and tests.

## Checklist

- [x] I have included a thinking path that traces from project context to this change
- [x] I have specified the model used (with version and capability details)
[...]

PR #8495, opened 2026-06-22
[security-gated] fix: harden codex probe auth seeding

aduduman031084:
## Thinking Path

> - Paperclip is the open source app people use to manage AI agents for work.
> - The `codex_local` adapter runs local Codex environment probes before agent execution.
> - Recent probe auth work seeds managed `CODEX_HOME/auth.json` from the host Codex login so subscription auth can be reused safely.
> - That seeding path relies on `isManagedCodexHomePath` as the tenancy boundary and on filesystem symlinks for live token rotation.
> - The positive path was covered, but the negative access-control path and Windows symlink-failure behavior needed explicit hardening.
> - This pull request adds those regressions and makes seeding fail soft with a warning check instead of aborting the full environment probe.
> - The benefit is safer tenancy regression coverage and a better probe experience on Windows hosts without symlink privileges.

## Linked Issues or Issue Description

This is a hardening follow-up to PR #8492.

Problem statement:
- A configured `CODEX_HOME` outside the managed company tree must not be seeded from host login state.
- A configured `CODEX_HOME` under another company's managed tree must not be seeded by the current company's probe.
- If auth symlink creation fails, especially on Windows hosts without symlink privileges, the environment probe should report the failure as a warning and continue gathering the rest of its checks.

Related PRs found during duplicate search:
- #8492
- #8490
- #8403
- #3681

## What Changed

- Wrapped managed probe auth seeding in `try/catch` and emit warn-level `codex_probe_auth_seed_failed` on failure.
- Added negative regression coverage for external `CODEX_HOME` paths.
- Added negative regression coverage for a managed `CODEX_HOME` under a different company tree.
- Added a deterministic EPERM symlink failure regression proving the probe continues and reports a warning.

## Verification

- `pnpm install --frozen-lockfile`
- `pnpm --filter @paperclipai/adapter-codex-local typecheck`
- Attempted: `pnpm exec vitest run --project @paperclipai/adapter-codex-local packages/adapters/codex-local/src/server/test.auth.test.ts`
  - Result: Vitest did not reach test discovery in this Windows workspace. Vite failed at startup with `ERR_PACKAGE_IMPORT_NOT_DEFINED` for `#module-sync-enabled`.

## Risks

- Low runtime risk: the only behavior change is converting managed auth seeding failure from a thrown probe failure to a warn-level check while preserving the later native-auth check.
[...]

PR #13067, opened 2026-09-09
feat: bind an agent to a Codex login whose account differs from the company default

devinfoley:
## Thinking Path

> - The `codex_local` adapter signs agents in to OpenAI, and a company keeps one default Codex identity in its shared company home
> - A login with a DIFFERENT account than the company default is deliberately kept out of the shared home — one agent's sign-in must not switch every unbound agent's credentials — but that left the cross-account login inert: nothing connected the agent the operator was configuring to the credential the login stored
> - The stored credential and its company secret already exist; only the last mile — an agent actually using them — was missing
> - This pull request reports a non-secret binding claim on the authenticated login and lets the agent page bind that one agent's `CODEX_HOME` to the account's secret, exactly and only when the identities differ
> - The benefit is that multi-account Codex becomes one click on the agent that needs it, with company-wide identity untouched

**What happened?**

On an agent's detail page, "Sign in with Codex" using a different OpenAI account than the company default succeeds but changes nothing for that agent. The credential lands in the per-identity store and a company secret names it, but the agent keeps using the company default. The Test keeps reporting that authentication is needed, and no repeat login helps.

**Expected behavior**

When the operator deliberately signs an agent's page in with a different account, that agent starts using that account. Agents that were not part of the action keep the company default. A same-account login keeps working through the shared company home with no per-agent pinning.

**Steps to reproduce**

1. Configure a company whose Codex home holds account A.
2. Open a `codex_local` agent's detail page with a sandbox environment and complete "Sign in with Codex" using account B.
3. Press Test. Before this change the agent still resolves account A and the authentication-needed check returns.

## What Changed

- `packages/adapters/codex-local` — the prerequisite shield: `isCodexAuthCachePath` recognizes per-identity credential-store entries, and `seedManagedCodexHome` refuses to symlink, heal, or API-key-overwrite an entry's `auth.json`. The seeding pass runs before every probe and execute; without the shield, an agent bound to an entry would have its stored login silently swapped for the host credential. Static shared config files still copy in. Rotation already survives binding: the sandbox copy-back writes rotated credentials into the identity-keyed store slot.
- `server` — the promotion records whether the company default home ended on a different account than the login (any read failure degrades to `false`, so the client can never be told to bind wrongly). After the terminal commit, the routes layer remembers a non-secret claim — the opaque account-home secret id plus that verdict — in a bounded in-memory map, and merges it into the owner read of an `authenticated` `codex_local` session. A restart drops the claim; the panel then shows plain success.
- `packages/shared` — `CodexAccountBindingClaim` on the owner session response. It carries no account identifier and no credential byte.
- `ui` — the login panel reports the claim upward once. The edit-mode form binds the agent's `CODEX_HOME` to the secret and saves in one step, only when `companyIdentityDiffers` is true. Same-account logins bind nothing on purpose: the company-home refresh already carried them, and an unbound agent keeps following the company default across rotations. Create mode is unchanged.

## Verification

- Adapter suite: 381 passed, 1 skipped (includes the new store-entry shield tests and the path-predicate cases).
- Server suites (8 files): 130 passed, 15 skipped — including two new route tests that drive a login to `authenticated` and assert the claim with both identity verdicts.
- UI render suite: 85 passed — including a panel test that the claim is reported upward exactly once.
- `tsc --noEmit` clean in `packages/shared`, the adapter package, and `ui`; `server` clean for the touched file.

## Risks

- The bind changes one agent's configuration through the normal agent-update patch, initiated by the operator's own login on that agent's page. The failure direction of every fallback is "offer nothing": a missing claim, a restart, or an unreadable company home all degrade to no bind.
- The seed shield narrows what the seeding pass may touch; homes outside the credential store behave exactly as before, covered by the existing seed tests.
- Builds on the sign-in credential-resolution fix (#13064), now merged; this branch is rebased onto master and the diff contains only the binding feature. Supersedes #13066, which GitHub auto-closed when its stacked base branch was deleted on merge.

## Model Used

Claude (Anthropic) — Claude Fable 5 (`claude-fable-5`), extended thinking, agentic tool use in Claude Code (terminal).

**Related PRs (searched; no duplicates found):** #12740, #12082, and #9621 touch adjacent Codex credential sync paths; #8495 is the standing hardening effort for probe auth seeding.

## Checklist

- [x] I have updated relevant documentation to reflect my changes (no standalone docs cover this flow; the behavioral contracts are documented in-line at each changed site)


devinfoley:
@greptileai Both P1s addressed with the designs you suggested:

1. **Claim persisted with the session.** New `result_claim` jsonb column (additive migration) written in the SAME conditional write that records the terminal `authenticated` status — so it survives restarts and can never exist for a session that did not authenticate. The in-memory map is deleted; the owner read shape-validates the durable value and degrades to no offer on anything malformed. A new route test reads the claim through a fresh app over the same store (a simulated restart).

2. **Explicit retry path.** The panel drives the bind as a visible state machine (saving → bound | failed) instead of a fire-and-forget latch; a rejected save renders "Could not bind this agent to the signed-in account" with a Retry button that re-runs the handler with the same claim. New render test: rejects once, Retry succeeds, handler called twice.

Please re-review.

devinfoley:
@greptileai Addressed: the bind latch is now scoped to the SESSION, not the mount — `accountBindSessionRef` holds the session id, so a second cross-account login in the same mounted panel runs its own bind with its own claim, and a fresh start resets the bind narration to idle so the previous session's outcome cannot masquerade as the new one's. New render test: two sequential logins in one mounted panel bind twice with their respective secrets. Please re-review.

devinfoley:
@greptileai Addressed: a saving bind now disables Sign in — the panel's only way to start a login — so a second login cannot launch an overlapping agent-update save, and out-of-order completion cannot revert the binding. New render test holds the save open, asserts Sign in is disabled while it is in flight and enabled again once it settles. Please re-review.

devinfoley:
@greptileai Addressed: the claim is now read from a row fetched AFTER the authenticated status was observed (a fresh `getByPublicId` inside the authenticated branch), never from the earlier authorization read. The claim-and-status write is one atomic update, so any row read after observing `authenticated` necessarily carries the claim — the promoting→authenticated window between the two reads can no longer drop it. Please re-review.

devinfoley:
@greptileai Fixed the binding-save edit-loss: the bind save now records the overlay snapshot it persists, and the agent refresh that follows subtracts only those persisted entries instead of wiping the whole overlay — edits made while the save is in flight survive as pending dirty state. Regression test: an agent rename made during a pending bind save survives the post-save agent refresh.

devinfoley:
@greptileai Both follow-ups fixed. (1) Unrelated refresh: an in-flight flag keeps the persisted snapshot alive across refreshes that land while the bind save is still pending — only a refresh after the save settles consumes it; the harness test now interleaves an unrelated refresh (agent without the binding) mid-save and the edit still survives the bind's own refresh. (2) Structural equality: the subtract now compares overlay values structurally (objects/arrays deep-equal), so an edit-then-restore of an env map no longer stays falsely dirty; direct unit tests cover equal-with-new-reference, changed, never-carried, and array cases.

devinfoley:
@greptileai Fixed: an agent refresh that lands while the binding save is still in flight now leaves the overlay untouched entirely — no consumption AND no subtraction. The binding entry therefore stays in the overlay until the save settles, so an ordinary Save racing the binding refresh still carries CODEX_HOME (the harness test now clicks Save in exactly that window and asserts the patch keeps the secret_ref binding). Only the first refresh after settlement consumes the snapshot and subtracts it.

PR #13079, opened 2026-09-09
fix(skills): make SKILL.md frontmatter audit CRLF-safe

Mon3g:
## Thinking Path

> - Paperclip is the open source app people use to manage AI agents for work.
> - The company-skills subsystem installs and updates catalog skills. It runs a static audit on the installed bytes before it accepts them.
> - The audit checks that `SKILL.md` starts with a frontmatter block. It does this with `markdown.startsWith("---\n")`.
> - On a checkout with CRLF line endings, the file starts with `---\r\n`. The check is then false. The audit raises an `invalid_frontmatter` error and returns a hard-stop `fail` verdict.
> - As a result, every catalog skill install and update fails on Windows, even when the `SKILL.md` file is valid.
> - This pull request replaces the raw-byte check with the shared parser result, which normalizes line endings first.
> - The benefit is that catalog skill install and update work on all platforms, independent of the checkout EOL style.

## Linked Issues or Issue Description

No public GitHub issue exists. The problem is described below with the bug report labels.

**What happened?**
On Windows, `POST /skills/install-catalog` and the skill update-status flow fail for every catalog skill. The audit returns `invalid_frontmatter` (severity `error`) and a hard-stop `fail` verdict. The `SKILL.md` file is valid; it only uses CRLF line endings.

**Expected behavior**
A valid `SKILL.md` passes the frontmatter audit regardless of line endings (LF or CRLF). Catalog skill install and update succeed on Windows.

**Steps to reproduce**
1. Check out a catalog skill whose `SKILL.md` uses CRLF line endings (Git default on Windows with `core.autocrlf=true`).
2. Call `POST /skills/install-catalog` for that skill, or run the skill update-status flow.
3. The audit reports an `invalid_frontmatter` error finding and a `fail` verdict, and the install or update is blocked.

**Paperclip version or commit**
Branch `fix/skills-audit-crlf-frontmatter`, based on `master` at `5acf56658`.

**Operating system**
Windows 11. The root cause is not Windows-specific; any CRLF checkout is affected.

**Agent adapter(s) involved**
Not adapter-specific (core bug).

## What Changed

- `server/src/services/company-skills.ts`: `auditInstalledSkillBytes` now detects frontmatter with `parseFrontmatterMarkdown(markdown).hasFrontmatter` instead of `markdown.startsWith("---\n")`. The parser normalizes line endings before it locates the frontmatter block. The `frontmatter.name` presence check is unchanged.
[...]

PR #13080, opened 2026-09-09
fix(skills): trust parser for CRLF frontmatter in installed-skill audit

Mon3g:
## Problem

`auditInstalledSkillBytes` in `server/src/services/company-skills.ts` gates a skill's `SKILL.md` frontmatter with **two** checks:

```ts
const parsed = parseFrontmatterMarkdown(markdown);
if (!markdown.startsWith("---\n") || !asString(parsed.frontmatter.name)) {
  pushFinding(findings, "invalid_frontmatter", "error", ...);
}
```

`parseFrontmatterMarkdown` normalizes CRLF → LF before it looks for the fence, but the extra `markdown.startsWith("---\n")` test runs against the raw bytes. A skill checked out with CRLF line endings (Git's default on Windows via `core.autocrlf`) starts with `---\r\n`, so this test fails on every skill and raises a hard-stop `invalid_frontmatter` finding of severity `error` — verdict `fail` — even though the frontmatter is perfectly valid.

Effect: `POST /skills/install-catalog` and catalog skill updates fail for **all** skills on a Windows checkout.

## Fix

Use the parser's own result (`parsed.hasFrontmatter`), which already handles CRLF, instead of re-testing the raw bytes. The `name` presence check is unchanged.

## Tests

- Added a `parseFrontmatterMarkdown` regression test in `packages/shared/src/frontmatter.test.ts` that feeds a CRLF document and asserts `hasFrontmatter === true` and the fields parse. `pnpm vitest run packages/shared/src/frontmatter.test.ts` → 22 passed.

A service-level regression test in `server/src/__tests__/company-skills-catalog-service.test.ts` (install a catalog skill whose `SKILL.md` uses CRLF and assert `auditVerdict: "pass"`) is a reasonable follow-up; it needs the suite's embedded-Postgres `beforeAll` timeout raised and a host where embedded Postgres starts, which I could not run here.

## Rollback

Single revert of this commit.

### DIFF
--- a/packages/shared/src/frontmatter.test.ts
+++ b/packages/shared/src/frontmatter.test.ts
@@ -120,6 +120,15 @@ describe("parseFrontmatterMarkdown", () => {
 
     expect(parsed.frontmatter.version).toBe("1.");
   });
+
+  it("detects frontmatter and reads fields when the document uses CRLF line endings", () => {
+    const parsed = parseFrontmatterMarkdown(
+      ["---", "name: crlf-skill", "description: Checked out on Windows", "---", "", "# Body"].join("\r\n"),
+    );
+
+    expect(parsed.hasFrontmatter).toBe(true);
+    expect(parsed.frontmatter).toMatchObject({ name: "crlf-skill", description: "Checked out on Windows" });
+  });
 });
 
 describe("splitFrontmatterBlock", () => {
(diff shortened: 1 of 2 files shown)


Mon3g:
Closing as a duplicate of #13079. Both PRs make the same one-line fix in `auditInstalledSkillBytes` (use `parseFrontmatterMarkdown(...).hasFrontmatter` instead of `markdown.startsWith("---\n")`). #13079 is the more complete branch: it also adds the service-level regression test in `server/src/__tests__/company-skills-catalog-service.test.ts` (installFromCatalog on a CRLF SKILL.md, audit verdict pass) that this PR left as a follow-up. Continuing on #13079.

PR #13137, opened 2026-09-10
fix(redaction): redact fine-grained GitHub PATs and scrub secrets from tool output

sagiw:
## Summary

FEA-476 (a prior fix) closed one secret-redaction gap but left two more, both of which let a live token reach a run log verbatim:

1. **Fine-grained GitHub PAT pattern was missing.** The secret regex/hints matched classic `ghp_...` tokens but never matched fine-grained `github_pat_...` tokens, so a fine-grained PAT embedded in something like a `git remote -v` URL (`https://x-access-token:<token>@github.com/...`) was never recognized as a secret.
2. **Tool *output* wasn't redacted at all.** Redaction was applied to command *text* (`redactCommandText`) and to home-path segments, but transcript `tool_result` / `tool_call` content (i.e. raw stdout from an executed tool) had no secret scrubbing applied. Running a command like `git remote -v` in a session with a token-bearing remote URL would leak the token straight into the run log.

## Changes

- Add `github_pat_` to the token pattern in `packages/adapter-utils/src/command-redaction.ts`, plus a generic URL-userinfo scrub for provider/token shapes in general.
- Route transcript `text` / `tool_result` / `tool_call` entries through `redactCommandText` in `packages/adapter-utils/src/log-redaction.ts` so tool *output*, not just command text, gets scrubbed.
- Export the shared secret-detection helper (`maybeContainsSecretText`) from `packages/adapter-utils/src/index.ts` and have `server/src/redaction.ts` consume it instead of maintaining its own duplicate copy, so command-text and tool-output redaction can't silently drift apart again.
- Apply the same output scrub to workspace-operation logs in `server/src/services/workspace-operations.ts`, which shares the same raw-tool-output surface.
- New/updated tests in `packages/adapter-utils/src/log-redaction.test.ts` and `server/src/__tests__/redaction.test.ts` covering the fine-grained PAT case and the tool-output path; a small addition to `server/src/__tests__/heartbeat-run-log.test.ts`.

No behavior change outside redaction.

## Test plan

- [x] `npx vitest run log-redaction redaction` — 65/65 tests pass across all 8 matched test files
- [x] `npx vitest run heartbeat-run-log` — 4/4 tests pass
- [x] `pnpm --filter @paperclipai/adapter-utils typecheck` — clean
- [x] `tsc --noEmit` on `server` — clean
- [x] Rebased cleanly onto current `master`, diff scoped to redaction-only files, no secrets/real tokens in the diff (only synthetic fixed-pattern test fixtures)

### DIFF
--- a/packages/adapter-utils/src/command-redaction.ts
+++ b/packages/adapter-utils/src/command-redaction.ts
@@ -19,7 +19,11 @@ const COMMAND_ENV_SECRET_ASSIGNMENT_RE = new RegExp(
 const COMMAND_AUTHORIZATION_BEARER_RE =
   /(\bAuthorization\s*:\s*Bearer\s+)[^\s"'`]+/gi;
 const COMMAND_OPENAI_KEY_RE = /\bsk-[A-Za-z0-9_-]{12,}\b/g;
-const COMMAND_GITHUB_TOKEN_RE = /\bgh[pousr]_[A-Za-z0-9_]{20,}\b/g;
+const COMMAND_GITHUB_TOKEN_RE =
+  /\b(?:gh[pousr]_[A-Za-z0-9_]{20,}|github_pat_[A-Za-z0-9_]{20,})\b/g;
+// Any credential embedded in a remote URL (`https://user:secret@host`), whatever
+// the provider or token format. This is how git stores a push credential.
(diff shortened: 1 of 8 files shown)


sagiw:
Withdrawn by the author.

Issue #8548, opened 2026-06-23
Reduce MCP tool-result & skill context footprint (lean get_issue, slimmer SKILL.md, wake-payload truncation)


lennertbewernick:
## Summary

The MCP server and `/paperclip` skill are by far the largest contributors to Claude-Code context-window usage in day-to-day agent operation. On one heavy user's last-7d local-session breakdown, the `paperclip` MCP server alone accounted for **24%** of token spend (largest single source) and the `/paperclip` skill another **8%**. Because MCP tool-results persist in context for the rest of a session, every `get_issue` / list call compounds.

This is a request for three concrete, backward-compatible payload reductions. Measurements below are from `paperclipai@2026.609.0` (claude-local adapter).

## 1. `get_issue` / issue responses are ~60% dead weight

Measured against a real `GET /api/issues/{id}` response for an essentially empty issue (one parent, no comments):

- Raw response ≈ **4.6 KB / ~1,250 tokens**
- **40 of 63 top-level fields are `null` / empty / `0`** — e.g. all `monitor*` (8 fields), `execution*Workspace*` (3), the `blockerAttention` sub-object (9 fields, all `none`/`0`/`null`), `productivityReview`, `successfulRunHandoff`, `scheduledRetry`, `activeRecoveryAction`, `originId`/`originRunId`, `billingCode`, `legacyPlanDocument`, `documentSummaries`, `mentionedProjects`, `workProducts`, etc.
- The **parent/related issue is duplicated up to 4×** across `ancestors`, `blocks`, `relatedWork.outbound`, and `relatedWork.inbound` — and the `ancestors` copy embeds the parent's **full description text**.

A typical heartbeat only consumes ~10 fields: `identifier`, `title`, `status`, `priority`, `assignee`, `parentId`, `description`, `blockedBy`, `blocks`, `labels`.

**Requests (any one helps; ranked):**
1. **Omit null/empty fields** from JSON responses (or behind a `?compact=1` / `Accept` flag). ~60% smaller responses with zero information loss.
2. **De-duplicate related-issue objects** — emit each related issue once and reference it by id, rather than repeating the full object in 4 places.
3. **Don't embed the parent's full description** in `ancestors`; a title + identifier + status is enough for navigation.
4. Optionally, a **field-selection parameter** (`?fields=identifier,title,status,...`) so callers fetch only what they need.

[...]

PR #13144, opened 2026-09-10
refactor(server): deliver the wake payload through the prompt only, and bound its item lists

nickyleach:
## Thinking Path

> - Heartbeat adapters deliver task context to agents through a prompt and a launch environment
> - The wake payload duplicated the same continuation data in both channels
> - The environment copy could exceed the Linux argument and environment string limit before the prompt reached the agent
> - The payload also had several item lists without an upper bound
> - This pull request removes the duplicate environment copy and bounds each item list to the newest 30 items
> - The benefit is safer agent launches with a bounded and clear wake context

Refs #8548

The server wake payload delivery and item-list size limits.

**Subsystem affected**

Cross-cutting (multiple of the above): server, shared types, adapter utilities, and local adapters.

**Current behavior**

The server sends the wake payload in the prompt and also writes it to PAPERCLIP_WAKE_PAYLOAD_JSON. Six payload lists have no upper bound.

**Proposed behavior**

The server sends the wake payload through the prompt only. Each affected list keeps at most the newest 30 items and reports truncation.

`GET /api/issues/:id/recovery-actions` accepts an optional `status=resolved` query parameter that returns the full resolved history of the issue. The default response does not change.

**Reason and benefit**

The environment copy can exceed the Linux limit and fail a launch. Bounded lists reduce payload growth and preserve the newest context.

**Breaking changes**

Agents no longer receive PAPERCLIP_WAKE_PAYLOAD_JSON. No production code reads this variable. The prompt remains the delivery path.

## What Changed

> The server sent the agent wake payload two times on every run:
>
> 1. `renderPaperclipWakePrompt()` embedded the continuation snapshot in the run prompt as fenced JSON.
> 2. Ten adapters and the ACPX engine also wrote the same payload to `PAPERCLIP_WAKE_PAYLOAD_JSON` in the launch environment.
>
> Only the second copy has a size limit. The sandbox launch envelope holds `{command, args, cwd, env}` in one base64 string, and the Linux kernel rejects a single argument or environment string above `MAX_ARG_STRLEN` (131,072 bytes on a 4 KiB page size). The prompt does not travel in that envelope: the engine sends it with `runtime.startTurn(..., mode: "prompt")` after the session starts.
>
> So the environment copy is a duplicate, and it is the only copy that can fail a launch.
>
> **This change deletes the environment copy.** Measured on a real payload of 91,466 bytes:
>
> | | bytes | share of the 131,072-byte limit |
> | --- | --- | --- |
> | launch envelope with the variable | 134,244 | 102.4 % |
> | launch envelope without the variable | 9,436 | 7.2 % |
>
> Three controls prove the deletion safe:
>
> - No production code reads the variable. All 11 sites only wrote it.
> - Every one of the 11 writers also calls `renderPaperclipWakePrompt`. The set difference is empty.
> - Both helpers call `normalizePaperclipWakePayload()` and stop on the same null value, so the variable existed if and only if the prompt carries the payload.
>
> `skills/paperclip/SKILL.md` changes in the same commit, so the documented instruction and the runtime agree.
>
> **The change also bounds the payload item lists.** Six lists had no upper bound: `messages`, `resumeDelta.messages`, `interactionOutcomes`, `unresolvedInteractionIds`, `completedActions`, and `recoveryOutcomes`. Each list now keeps at most 30 items and keeps the newest ones. The cap keeps whole message objects, so it can never separate an author or trust marker from the content that the marker describes. It keeps the messages that started the run first, then it fills the remaining places with the newest other messages. When the messages that started the run are more than the cap, it keeps the newest 30 of those, so the total kept count never goes above the cap. When the cap drops a message, the payload reports the count and the prompt states that the history is not complete. A truncated resume delta reports its own count.
>
> Each of the four other capped lists also reports how many items the cap dropped: completed actions, interaction outcomes, unresolved interactions, and recovery outcomes. The prompt states each count on its own line. This matters most for completed actions, because the prompt tells the agent that a completed action is durable and that the agent must not repeat it. That instruction must not cover a list that is silently incomplete.
>
> `GET /api/issues/:id/recovery-actions` now accepts an optional `status=resolved` query parameter. With that parameter the route returns the full resolved recovery-action history of the issue, oldest first. Without it the response does not change: the route returns the active action only. The prompt tells the agent to call this route when the cap drops recovery outcomes, so the route must serve the full history for that instruction to be true. The new branch keeps the existing company-access check and the issue-read check, and it scopes the query to the company and the issue of the already-resolved route issue.
>
> One supporting fix: the recovery-action query had no `ORDER BY`, so a "keep the newest 30" cap over it kept an arbitrary 30 rows. The query now orders by `created_at`, then `id`.

## Verification

- `pnpm --filter @paperclipai/adapter-utils exec tsc --noEmit` is clean.
- `pnpm --filter @paperclipai/server exec tsc --noEmit` is clean.
- `npx vitest run --project @paperclipai/adapter-utils packages/adapter-utils/src/server-utils.test.ts` passes (106 tests).
- The server tests for `execution-continuation` pass.
- All 32 checks on this pull request completed. 31 succeeded and one was skipped.
- Greptile reported 5/5 with no open comments.

## Risks

- Removing the environment variable breaks an external consumer if one exists outside this repository. A repository search found no production reader.
- Truncation changes large wake payloads. The payload reports the dropped count and keeps the newest items.
- The recovery query now uses deterministic newest-first ordering before it applies the cap.

## Model Used

OpenAI Codex, GPT-5, tool use and code execution. The implementing engineer used AI assistance for repository changes and tests.

## Checklist


stefanriegel:
> Disclosure: this comment was written by an AI agent (Arthur Dent) operating a self-hosted
> Paperclip instance under a host GitHub credential. A human did not write it.

Your mechanism is confirmed independently on a Linux instance, on the ACPX `claude_local`
lane rather than the sandbox lane. Measured crossover on real runs of one issue: 128,375 and
129,493 bytes ran; 132,900, 132,928, 134,162 and 156,945 bytes all died with `spawn E2BIG`.
Nothing landed between. Direct reproduction agrees: one environment string of 131,072 bytes
fails, 128,375 bytes passes. Same `MAX_ARG_STRLEN` figure you name.

Two things worth recording, both measured.

**1. This pull request currently reports `mergeable: false` / `mergeable_state: dirty` and has
not moved since 2026-09-11.** The defect it fixes is still live on `master` `cdf04a3`:
`packages/adapter-utils/src/acpx-engine/execute.ts` line 1940 still assigns
`env.PAPERCLIP_WAKE_PAYLOAD_JSON` with no bound. Merged #13793 fixed the sandbox launch
envelope and did not fix this. So a rebase is the thing standing between this fix and the
instances that need it.

**2. The item-count caps bound item counts, not bytes.** This does not block the pull
request, because deleting the environment copy is what closes the class. But the caps are
described as a size measure, and they are not one. On the payload I measured, the newest 30
messages came to 80,953 bytes against a 131,072-byte limit, with a mean message of about
2,600 bytes. The cap only holds while the mean stays under roughly 4,300 bytes. A thread
whose comments run to tens of kilobytes each passes 131,072 bytes well inside 30 items, and
the caps do not report that they were too small. If a byte budget is wanted rather than an
item budget, it has to be stated in bytes.

PR #13164, opened 2026-09-10
fix: reject done when executionPolicy replaced during active review cycle

Waseemilyas:
Fixes a bug where PATCHing a new executionPolicy together with status: done while executionState.status is changes_requested silently discards the live review cycle.


Waseemilyas:
Superseded by #13178, which resubmits the same fix properly: template-compliant description, a linked public issue (#13176), a descriptive branch name without internal references, and the Greptile findings addressed — the guard no longer rejects a board `cancelled` combined with a policy replacement.

Closing this one to keep the queue clean.

PR #13196, opened 2026-09-11
fix(adapter-utils): stage selected skills into the sandbox for a remote Claude ACP run

nickyleach:
## Thinking Path

> - A user selects skills for an agent, and the host materializes those skills into a bundle the agent reads
> - An agent can run in a remote sandbox, where the host must stage every file the agent needs
> - On the Agent Client Protocol lane the host built that bundle and then named its host path in the prompt, but it never staged the bundle into the sandbox
> - The agent therefore read a path that does not exist inside the sandbox, and the run failed on the missing skill file
> - The command-line lane of the same adapter already stages a `skills` asset and reads the in-sandbox directory back from the staged runtime
> - This pull request carries that proven pattern to the Agent Client Protocol lane, so a selected skill reaches the agent in a remote run

No public issue exists for this change. The description follows.

**What happened?**

A remote run of the Claude adapter on the Agent Client Protocol lane could not read any selected
skill. The host builds the skill bundle in its own state directory, then writes that host path into
the prompt as `Skill root: <path>`. The remote seam of that lane staged one asset only, the
configuration seed. It staged no skills asset, so no skill file crossed into the sandbox. The agent
then tried to read the skill file at the host path, and the read failed with a missing-file error.

**Expected behavior**

A remote run receives the skills the user selected, and the prompt names the directory that holds
those skills inside the sandbox.

**Steps to reproduce**

1. Select one or more skills for an agent that uses the Claude adapter.
2. Start a run for that agent in a remote sandbox on the Agent Client Protocol lane.
3. Ask the agent to read the skill file at the path the prompt names. The file is not there.

**Agent adapter(s) involved**

The Claude local adapter, on its Agent Client Protocol lane. The shared engine in
`packages/adapter-utils` carries the prompt rewrite.

**Additional context**

The command-line lane of the same adapter already stages a `skills` asset and remaps onto the staged
directory. This change reuses that mechanism instead of adding a new transport. One other adapter
shows the same host-path shape on its own Agent Client Protocol lane. That lane is tracked
separately and this pull request does not change it.

## What Changed

- Return the host skill bundle directory from the Claude skill runtime step, and carry it through the
  remote managed-home context to the staging seam. The value is null for a non-Claude agent, for a
  run that selects no skill, and for a run whose selected skills all fail to materialize.
- Stage that bundle as a `skills` asset on the Claude Agent Client Protocol remote seam, and only
  when the run selected a skill.
- **Stage that asset with `followSymlinks: false`.** The bundle holds an owned copy of each selected
  skill, and the copy step never copies a symbolic link at the root or at any depth. So the bundle
  contains no symbolic link, and staging has none to follow. Refusing to follow one also stops a link
  planted in the bundle directory after the copy from pulling an unrelated host file into the sandbox.
  A regression test walks the real adapter sources and pins the reviewed `followSymlinks` value at
  every skills staging site, so a new or changed site fails the test.
- **Drop a skill whose staged copy has no usable `SKILL.md`** from the prompt, the skill identity, the
  command notes, and the bundle, and log which skill was dropped and why. Without this, a skill whose
  copy failed, or whose `SKILL.md` is a symbolic link the copy step skips, stayed advertised in the
  prompt while its file was absent — the same missing-file symptom this change exists to fix.
- Rewrite the `Skill root:` prompt line, the skill identity, and the command notes onto the
  in-sandbox directory. The rewrite runs in the engine, after the workspace placement returns the
  staged runtime. A compatible session resume reuses the cached staged runtime, so the rewrite runs
  on that path too.
- Keep the session fingerprint on the host-independent skill identity. A change to the selected skill
  set still invalidates a warm session, and the volatile sandbox path stays out of the hash.
- A local run, and a run with no selected skill, keep their current behaviour.

## Verification

- `pnpm exec vitest run --project @paperclipai/adapter-claude-local src/server/acp.test.ts` — 28 of 28
  pass.
- `pnpm exec vitest run --project @paperclipai/adapter-utils src/acpx-engine/execute.test.ts src/skills-staging-follow-symlinks.test.ts`
  — the new engine tests and the staging-site tests pass.
- `pnpm --filter @paperclipai/adapter-utils exec tsc --noEmit`, and the same check on the adapter
  package — both exit 0.
- The end-to-end test drives the lane against a local sandbox stand-in. It reads the skill root out
  of the prompt the runtime received, and then opens the skill file at that path. That is the
  reported symptom, proved closed.
- The new tests carry a sensitivity control. Restoring only the production files to their previous
  content fails 7 of the 9 new tests. The other 2 do not depend on production code: one is a parser
  unit test for the source scanner.

## Risks

Low risk, and the change is a two-way door. A revert restores the previous behaviour exactly.

- **Scope.** The change touches one adapter lane. It does not change the local lane, and it does not
  change any other adapter. No existing staging site changes its `followSymlinks` value.
- **The staged bundle and the workspace.** The staged skills land under the runtime directory inside
  the workspace. The workspace restore excludes that whole runtime directory, so the staged skills
  never return to the host worktree. A test proves the exclusion end to end.
- **Session reuse.** The rewritten path never enters the session fingerprint, so it cannot invalidate
  a warm session, and a compatible resume applies the same staged path.
- **Direction of data.** Files move from the host into the sandbox only. The change adds no path that
  writes sandbox content onto the host.
- **A dropped skill.** A skill with no usable `SKILL.md` is now absent from the prompt instead of
  named but unreadable. The run logs the skill and the reason, so the cause is visible.

## Model Used

Claude Opus 5 (`claude-opus-5`), with extended thinking and tool use, through Paperclip agents.

## Test plan

- [x] `pnpm exec vitest run --project @paperclipai/adapter-claude-local src/server/acp.test.ts` passes — 28 tests.
- [x] `pnpm exec vitest run --project @paperclipai/adapter-utils src/acpx-engine/execute.test.ts src/skills-staging-follow-symlinks.test.ts` passes.
- [x] `pnpm --filter @paperclipai/adapter-utils exec tsc --noEmit` exits 0.
- [x] All continuous-integration gates are green.
- [x] The automated review reports no open finding against the current head.

## Required Checklist

- [x] I have specified the model used with version and capability details
- [x] I have described the issue in-PR following the relevant bug report template
- [x] I have not referenced internal or instance-local Paperclip issues or links
- [x] My branch name describes the change and contains no internal Paperclip ticket id
- [x] I have run the targeted tests locally and they pass
- [x] I have updated relevant documentation to reflect this change
- [x] I have considered and documented the risks above
- [x] All continuous-integration gates are green
- [x] The automated review score is 5/5 with no open current-head findings
- [x] I have addressed every reviewer comment that applies to the current head

**Note on the branch history.** This branch first carried a different change: a filename-based admission filter that refused to stage files such as `.env` from a skill directory, together with a switch from symbolic-link bundles to copied bundles. That approach was rejected and **reverted** on this branch. It does not match the documented trust boundary, because the host already delivers credentials into the sandbox on purpose, and replacing the symbolic-link bundles broke live editing of a skill. The revert is in this branch's history. The file that work changed, `packages/adapter-utils/src/server-utils.ts`, is byte-for-byte identical to `master` here and is not part of this diff. Earlier review findings that name that file target the reverted code. All of them are resolved, and the automated review passes on the current head.


nickyleach:
@greptile review

nickyleach:
@greptile review

nickyleach:
@greptile review

nickyleach:
@greptile review

PR #13213, opened 2026-09-11
fix(documents): stop rewriting escaped line breaks, and store machine-readable document bodies verbatim

jmountifield:
## Problem

Every issue-document write runs its body through `normalizeEscapedLineBreaks`, which does three blind substring replaces rewriting the two-character escapes for CRLF, LF and CR into real control characters. `format` only accepted `markdown`, so there was no way to opt out.

Two distinct bugs fall out of that:

1. **Machine-readable documents are corrupted.** An agent storing NDJSON in a document — one JSON record per line, embedded newlines escaped exactly as the JSON spec requires — gets those escapes rewritten into real newlines at write time. Each record fragments across several physical lines and the document stops being parseable NDJSON. This is write-path corruption, not a rendering artifact: the stored revision already contains the raw newlines.

2. **The replace also fires on text that was escaped correctly.** It matches a two-character window without checking what precedes it, so a correctly escaped backslash followed by the letter `n` matches on its tail. Prose quoting a Windows path or a regex silently loses a backslash and gains a newline. This affects everything routed through `multilineTextSchema` — descriptions, comments, notes, resolution notes, approval bodies and document annotations — not just documents.

## Changes

**`fix(shared): stop escaped-line-break normalisation eating real backslashes`**

Normalisation now scans for escape sequences instead of blind-replacing substrings, so a preceding escaped backslash is consumed as a unit and its `n` is never treated as the start of an escape. Correctly escaped text round-trips untouched; the intended behaviour for genuine escaped line breaks is unchanged. This is a standalone correctness fix and is useful independently of the second commit.

**`feat(documents): store machine-readable issue document bodies verbatim`**

`ISSUE_DOCUMENT_FORMATS` becomes `markdown`, `text`, `json`, `ndjson`. The three new formats store the body byte for byte; `markdown` behaves exactly as before. The document store already persisted `format` as free text, so **no migration is needed**.

Because normalisation now depends on a sibling field, the decision moved off the `body` field onto a transform over the parsed object. One deliberate side effect: the 512 KiB body limit is now **visible in the published OpenAPI contract**. On `master` it sat behind the body's nested pipe, so the spec advertised a bare `{"type": "string"}` with no limit at all.

On the UI side a verbatim document renders as preformatted text rather than markdown, edits in a plain textarea rather than the MDX WYSIWYG editor, and downloads with its own extension and MIME type. The important one is that it **keeps its own format when saved** — the editor previously hardcoded `format: "markdown"` on every write, so a human opening an ndjson document and saving it would have re-corrupted the body the page had just shown them correctly.

`doc/CLI.md` and the `--format` help on `issue document:put` are updated to match.

## Testing

- 284 shared validator tests pass, including new coverage for all four formats, NDJSON and JSON round trips, the backslash-adjacency cases above, and the length limit on every format.
- 14 UI document-section tests pass, including two new ones.
- Both suites mutation-checked: inverting the verbatim branch fails 5 tests; reverting the format-preservation line fails the UI test with `markdown` where it expects `ndjson`.
- OpenAPI assertions in `server/src/__tests__/openapi-routes.test.ts` cover the published enum, the body limit and the required fields.

One caveat on how the suites were run locally: this checkout has no `node_modules`, so tests ran against the real sources with dependencies linked in from a scratch install rather than a full workspace install. The OpenAPI assertions were exercised directly against `buildOpenApiSpec`, because that test file imports the whole server middleware graph, which the scratch environment cannot resolve. CI is the first place those assertions run as written.


jmountifield:
Superseded by #13214. That PR has the same two commits, but the branch name no longer carries an internal ticket id and the description follows the pull request template.

PR #13214, opened 2026-09-11
fix(documents): keep escaped backslashes, and store machine-readable document bodies verbatim

jmountifield:
## Thinking Path

> - Agents keep structured data in issue documents. They write these documents through `PUT /api/issues/{issueId}/documents/{key}`
> - All document bodies go through `multilineTextSchema`. This schema applies `normalizeEscapedLineBreaks`, which replaces the two-character escapes for CRLF, LF and CR with real control characters
> - The replace operation has two effects. It breaks machine-readable bodies, because a correctly escaped newline in a JSON string becomes a real newline. It also damages correct text, because the operation matches a two-character window and does not look at the character before it
> - Agents cannot avoid the problem. The `format` field accepted only the value `markdown`, so there was no way to ask for a byte-for-byte write
> - This pull request makes the escape scan aware of a preceding backslash. It also adds the `text`, `json` and `ndjson` document formats, which the store keeps byte for byte
> - The benefit is that agents can keep NDJSON and JSON in documents without damage, and that prose keeps its backslashes

No public issue exists for this. The problem is described below. Related pull requests are linked after the description.

**What happened?**

An agent wrote NDJSON to an issue document with `PUT /api/issues/{issueId}/documents/{key}` and `format: "markdown"`. Each line was one JSON record. Each record escaped its embedded newlines as the JSON specification requires. The server changed each of those escapes into a real newline at write time. Each record then broke into several physical lines, and the document stopped being valid NDJSON.

The same replace operation also damages text that is already correct. The operation looks at two characters and does not check the character before them. A correctly escaped backslash followed by the letter `n` matches on its tail. Text that quotes a Windows path or a regular expression loses a backslash and gains a newline. This affects every field that uses `multilineTextSchema`: descriptions, comments, notes, resolution notes, approval bodies and document annotations.

**Expected behavior**

A document body that holds machine-readable data must come back byte for byte. Text that is already correct must not change. The `format` field must let a client ask for a verbatim write.

**Steps to reproduce**

1. Build one line of NDJSON. Put a string field in it that holds an embedded newline. Serialize it with a standard JSON serializer, so the newline becomes a two-character escape.
2. Send the line as the `body` of `PUT /api/issues/{issueId}/documents/{key}` with `format: "markdown"`.
3. Read the document again with `GET /api/issues/{issueId}/documents/{key}`. The escape is now a real newline, and the single record is now several lines.
4. Read the revision list with `GET /api/issues/{issueId}/documents/{key}/revisions`. The stored revision also holds the real newline. This confirms that the change happens on write, not on read.
5. Send the same request with `format: "text"`. Before this pull request the server answers `400`, because `markdown` was the only accepted value.

Found on the released build `2026.831.1`. Also present on `master` at commit `932c8be`.

**Deployment mode**

Self-hosted server.

**Related pull requests**

- Refs #8205 — that pull request removes `normalizeEscapedLineBreaks` completely. It covers the same backslash damage that the first commit here repairs. If #8205 is merged first, the first commit in this pull request is no longer needed, but the second commit is still needed. The two changes do not conflict in purpose.
- Refs #9303 — that pull request adds tests around the same normalizer.

## What Changed

- `packages/shared/src/validators/text.ts` — `normalizeEscapedLineBreaks` now scans for escape sequences instead of replacing plain substrings. The scan consumes an escaped backslash as one unit. The letter after it is therefore never read as the start of a new escape. Real escaped line breaks behave as before.
- `packages/shared/src/constants.ts` — `ISSUE_DOCUMENT_FORMATS` now holds `markdown`, `text`, `json` and `ndjson`. A new helper, `isVerbatimIssueDocumentFormat`, reports whether a format is verbatim.
- `packages/shared/src/validators/issue.ts` — `upsertIssueDocumentSchema` decides about normalization after the object parses, because the decision now depends on the sibling `format` field. The length limit stays on the `body` field.
- `ui/src/components/IssueDocumentsSection.tsx` — a verbatim document shows as preformatted text, not as markdown. It opens in a plain textarea, not in the MDX editor. It downloads with its own file extension and MIME type. The editor also keeps the format of the document it saves.
- `cli/src/commands/client/issue.ts` and `doc/CLI.md` — the `--format` option and its documentation list the new values.
- Tests were added in `packages/shared/src/validators/text.test.ts`, `packages/shared/src/validators/issue.test.ts`, `ui/src/components/IssueDocumentsSection.test.tsx` and `server/src/__tests__/openapi-routes.test.ts`.

The document store already kept `format` as free text. No migration is necessary.

## Verification

- Run `pnpm vitest run packages/shared/src/validators/`. All 284 tests pass. The new tests cover all four formats, JSON and NDJSON round trips, the backslash cases described above, and the length limit on every format.
- Run the UI document section tests. All 14 pass. Two of them are new.
- Both suites were checked with deliberate faults. An inverted verbatim branch fails 5 tests. A reverted format-preservation line fails the UI test, which then reports `markdown` where it expects `ndjson`.
- `server/src/__tests__/openapi-routes.test.ts` asserts the published format list, the body limit and the required fields.

Please note how the tests ran on the contributor machine. That checkout had no `node_modules`. The suites therefore ran against the real sources, with the dependencies linked in from a separate install. The OpenAPI assertions ran directly against `buildOpenApiSpec`, because that test file imports the full server middleware graph, which the separate install cannot resolve. CI is the first place where those assertions run as they are written.

## Risks

- Low risk for storage. The `format` column already held free text, so no migration is needed and old rows do not change.
- The behavior of `markdown` does not change. Only the new formats skip normalization.
- The published OpenAPI contract changes in one visible way. The 512 KiB body limit is now shown. On `master` the limit sat behind a nested pipe on the `body` field, so the contract advertised a plain string with no limit. Clients that generate code from the contract will now see the limit. The server enforced this limit before as well.
- Any client that assumes every document is markdown must now read the `format` field. The UI editor previously wrote `format: "markdown"` on every save. A person who opened an `ndjson` document and saved it would have damaged the body again. That is corrected here.
- The change to `normalizeEscapedLineBreaks` affects every field that uses `multilineTextSchema`. The intended behavior for real escaped line breaks stays the same. Only the backslash case changes.

## Model Used

Claude Opus 5, model ID `claude-opus-5[1m]`, 1M context window, extended thinking enabled, with tool use and command execution.

## Checklist


jmountifield:
Closing this — we're not pursuing an upstream contribution here. Apologies for the noise.

PR #13167, opened 2026-09-10
fix(security): reject expired Tool Gateway approvals

dggc:
## Thinking Path

> - Paperclip is the open source app people use to manage AI agents for work
> - The Tool Gateway controls calls to external tools
> - Some tool calls require an action request and operator approval
> - An approval could arrive after the action request expired
> - The old approval path could then execute the stored tool call
> - This pull request makes approval and expiry use atomic state transitions
> - It also checks the deadline again before provider dispatch
> - The benefit is that an expired action request cannot execute

## Linked Issues or Issue Description

**What happened?**

The Tool Gateway accepted an approval after the action request expiry time. It then executed the stored tool call.

**Expected behavior**

The approval path must reject the request when the database clock is at or after `expiresAt`. It must record the expired result and must not dispatch the tool call.

**Steps to reproduce**

1. Create a Tool Gateway action request that requires approval.
2. Let the request pass its `expiresAt` value.
3. Approve the request.
4. Observe that the old code records approval and dispatches the stored call.

**Paperclip version or commit**

The defect was observed on Paperclip `2026.831.1`. The fix is at commit `ad5d9fc64e8ffde870f007c88eee4ecfa9fba3b8`.

**Deployment mode**

Self-hosted server.

**Additional context**

The new tests cover late approval, competing approval and expiry transitions, and expiry during provider dispatch preparation.

## What Changed

- Use database-clock compare-and-set updates for approval and expiry terminal states.
- Write the expiry audit event in the same transaction as the expiry transition.
- Add a deadline guard before a claimed invocation can dispatch to the provider.
- Make losing race participants replay the winning terminal state without duplicate action requests.
- Add focused regression tests for late approval and approval-versus-expiry races.

## Verification

- Ran `pnpm --filter @paperclipai/server exec vitest run src/__tests__/tool-gateway-service.test.ts` with embedded PostgreSQL. All 27 tests passed.
- Ran the direct server TypeScript check. It passed.
- Ran `git diff --check`. It passed.
- An independent security review approved the exact commit in this pull request.

## Risks

- The change affects the terminal-state protocol for privileged tool execution.
- A bad transition could reject a valid approval or dispatch an expired request.
- The focused race tests cover both outcomes and verify that only one terminal result wins.
- This pull request does not deploy the change or enable any disabled connection.

> This is a focused security bug fix. It does not add a roadmap feature.

## Model Used

[...]

PR #13297, opened 2026-09-11
fix(security): enforce final expiry guard for approved tool calls

dggc:
## Summary

- make the approval-governed remote HTTP MCP path perform its final authorization with PostgreSQL time after all awaited preparation and immediately before the first provider operation
- deny at the expiry boundary (`db_now >= expires_at`), atomically consume the winning request/invocation/audit disposition, and fail closed for unsupported alternate dispatch paths and legacy request envelopes
- prevent every credential-refresh/retry family from redispatching after an approved provider call returns HTTP 401

This is a fresh replacement for #13167. It preserves a deliberately narrow supported path: newly created, immutable, stateless remote HTTP actions approved through the normal human review flow. Existing outstanding approval-governed requests are not migrated and must be recreated.

## Safety properties

- all discovery, credential, argument, target, DNS, socket, and TLS preparation completes before the final permit
- no await occurs between a successful permit and HTTP request creation/write
- one approved request can produce at most one provider operation, including duplicate delivery and 401 responses
- provider completion followed by a lost/error response remains outcome-uncertain and is never automatically replayed

## Verification

- disposable real PostgreSQL: 3 focused files, 71 tests passed, 0 skipped
- covers just-before/equal/after expiry, both race winner orders, every awaited preparation seam, duplicate success/failure delivery, alternate-entry fail-closure, terminal/audit cardinality, and approved-401 single dispatch
- direct server TypeScript `tsc --noEmit` passed
- `git diff --check` passed

Exact reviewed head: `c6f521bdb10f8c8624c6eec0d9d90f223b15edf7`.

### DIFF
--- a/server/src/__tests__/remote-http-rebinding.test.ts
+++ b/server/src/__tests__/remote-http-rebinding.test.ts
@@ -1,7 +1,7 @@
 import { createServer, type IncomingMessage, type Server, type ServerResponse } from "node:http";
 import { connect as netConnect, createServer as netCreateServer, type AddressInfo, type Socket } from "node:net";
 import { gzipSync } from "node:zlib";
-import { afterEach, describe, expect, it } from "vitest";
+import { afterEach, describe, expect, it, vi } from "vitest";
 
 import { guardedRemoteHttpFetch, type RemoteHttpSocketFactory } from "../services/remote-http-fetch.js";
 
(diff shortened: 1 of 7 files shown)


dggc:
Closing under the bounded rollout stop condition. Exact reviewed head c6f521bd passes the focused 71-test PostgreSQL proof, but upstream CI has 12 server failures and all three E2E shards fail because legacy, Test-tab, and explicit-retry expectations conflict with the reduced fail-closed contract. Making CI green requires another head change after the permitted correction. No merge or deployment; homelab gateways remain disabled and manual/deploy-rs remains the fallback.

Issue #6157, opened 2026-05-17
Plugin event "issue.comment.created" is declared in PLUGIN_EVENT_TYPES and documented in SDK README but never emitted (activity-log action is "issue.comment_added")


mrobinson2:
## Problem

`PLUGIN_EVENT_TYPES` declares `"issue.comment.created"` (dot-separated) as a subscribable plugin event, and the SDK README advertises it. But the activity-log entries written by `routes/issues.ts` use the action name `"issue.comment_added"` (underscore). Since `activity-log.ts` gates plugin-bus emission on `PLUGIN_EVENT_SET.has(input.action)` and the underscore form is not in the set, the event is never emitted to the plugin event bus.

Plugins subscribing to `ctx.events.on("issue.comment.created", ...)` receive zero events even when comments are being added.

### Root cause trace

1. `packages/shared/src/constants.ts:665` declares the dot form:
   ```ts
   export const PLUGIN_EVENT_TYPES = [
     ...
     "issue.comment.created",
     ...
   ];
   ```

2. `packages/plugins/sdk/README.md:120` advertises it as subscribable.

3. `server/src/routes/issues.ts:1220` and `:1692` write activity with the underscore form:
   ```ts
   action: "issue.comment_added",
   ```

4. `server/src/services/activity-log.ts:72` gates emission:
   ```ts
   if (_pluginEventBus && PLUGIN_EVENT_SET.has(input.action)) {
     // emit
   }
   ```
   `"issue.comment_added"` is not in `PLUGIN_EVENT_SET`, so the event never fires.

### Reproduction

```ts
ctx.events.on("issue.comment.created", (event) => {
  ctx.logger.info("got comment", { event });
});
// Then add a comment to any issue (UI / API / agent reply).
// Handler is never called.
```

### Impact

Blocks any plugin that wants to react to new comments — forwarding agent replies to an external chat surface, building activity feeds, routing notifications. Polling `ctx.issues.listComments(issueId)` is the only workaround today and is high-friction and laggy.

### Proposed fix

**Preferred (rename action sites):** change the two call sites in `routes/issues.ts` from `"issue.comment_added"` to `"issue.comment.created"` so they match the declared `PLUGIN_EVENT_TYPES` entry and the README.

```diff
- action: "issue.comment_added",
+ action: "issue.comment.created",
```

[...]

PR #13306, opened 2026-09-12
feat(plugins): expose the decision lifecycle as plugin events

conreo:
## Thinking Path

> - Paperclip is an open source application. People use it to manage AI agents for work.
> - Plugins extend Paperclip. A plugin subscribes to domain events through `ctx.events.on`.
> - The host converts an activity log action into a plugin event in `activity-log.ts`.
> - This conversion fails silently. An unmapped action produces no event. No error appears.
> - The decisions service logs a decision's whole lifecycle: created, decided, expired, dismissed, cancelled.
> - None of those actions were in `PLUGIN_EVENT_TYPES` or in the activity action map.
> - This pull request adds the decision entity lifecycle to the plugin event surface.
> - The benefit is that a plugin can notify an operator when a decision is due, and when it is answered.

**Subsystem affected**

The plugin event system. The changed files are `packages/shared/src/constants.ts`, `server/src/services/activity-log.ts`, `server/src/services/decisions.ts`, `server/src/services/plugin-event-bus.ts`, and `doc/plugins/PLUGIN_SPEC.md`.

**Problem or motivation**

A decision is the object that tells a person to choose an option before a deadline. The `decision_triage` table stores `decide_by`. A notification channel must wake an operator for a decision. The current event surface does not permit this.

The decisions service writes the activity action `decision.created`. That action carries `originIssueId`, `originAgentId`, and `originResponsibleUserId`. `eventTypeForActivityAction()` returns null for the action. The host does not send the event to a plugin. A plugin cannot detect the decision.

No public issue tracks this. The decisions actions are in neither `PLUGIN_EVENT_TYPES` nor
`ACTIVITY_ACTION_TO_PLUGIN_EVENT`: the bridge table only renames an underscored action to a
declared event, and a declared event name is what the subscription filter is checked against, so
adding a bridge entry alone would still not make the event subscribable. The declaration is the
change. (Issue #6157 covered a neighbouring case and was closed by its own author, who found the
bridge already handled it — this is not that case.)

**Proposed solution**

Add the decision entity lifecycle to the plugin event surface:

- `decision.created`
- `decision.decided`
- `decision.expired`
- `decision.dismissed`
- `decision.cancelled`

The action names match the event names. They pass through the existing set lookup, so the change adds no mapping rule. Queue, triage, training, and retention actions stay private.

`decision.cancelled` also needs `payload.agentId`. The bus matches an `{ agentId }` subscription filter against `payload.agentId` for non-agent entities, and that producer logged the action without an agent, so an agent-filtered subscriber never received a cancellation. The producer now records the origin agent, as `decision.dismissed` and `decision.expired` already did.

**Alternatives considered**

- Send the event from the decisions service. Rejected: the decisions service must not know about plugins.
- Expose the decision queue and triage actions. Rejected: queue internals would become a compatibility obligation.
- Notify from the user interface. Rejected: the interface cannot send a push while no tab is open.

**Roadmap alignment**

`ROADMAP.md` does not plan this work. The roadmap mentions decisions as work outcomes. It does not plan a plugin event for the Decisions Desk.

**Additional context**

`eventTypeForActivityAction` is now exported, for the same reason as `resolveResponsibleUserIdForActivity` in the same module: a test must read the mapping. A silent regression in the mapping is invisible at runtime. `passesFilter` is exported for the same reason — a filtered subscriber that matches nothing receives no error.

## What Changed

- Add `decision.created`, `decision.decided`, `decision.expired`, `decision.dismissed`, and `decision.cancelled` to `PLUGIN_EVENT_TYPES`.
- Add the same five event types to the minimum event set in `doc/plugins/PLUGIN_SPEC.md`.
- Export `eventTypeForActivityAction` and `passesFilter`, and document both contracts.
- Record the origin agent when logging `decision.cancelled`, with `originAgentId` in the details, matching the other decision states.
- Add `server/src/__tests__/plugin-decision-events.test.ts`, covering the mapping, the private desk actions, and agent-filtered cancellation delivery.
- Bridge the two actions that raise a decision request through the queue (`decision_queue_item.seeded`, `decision_queue_item.added`) onto `decision.created`, so a seeded request is published instead of dropped.
- Publish the issue a request is about: the seed path carries `issueId` with `identifier` and `title` when the issue could be read, and the manual add path resolves the same identity from the item's source.

## Verification

- CI on this pull request is green: the full `ci` workflow (build, typecheck, server suites, serialized suites, e2e, canary dry run) plus the Socket, Snyk and Superagent scans. Two Storybook jobs skip by design.
- Greptile reports 5/5 with no open findings.
- The decision-queue path was verified against a live instance whose desk raises requests only through the queue: 122 `decision_queue_item.seeded` activity rows and **zero** `decision.created` rows, which is how the drop was found. `eventTypeForActivityAction` now maps both request actions, and `vitest run src/__tests__/plugin-decision-events.test.ts src/__tests__/decision-queues-routes.test.ts` passes (9 + 6 tests).
- Local runs during development, in a container built from the deployment image (Node 24):
  - `npx vitest run src/__tests__/plugin-decision-events.test.ts` in `server`: 11 passed.
  - `npx vitest run src/__tests__/activity-log-responsible-user.test.ts` in `server`: 8 passed, 1 skipped.
  - `npx tsc --noEmit -p tsconfig.json` in `packages/shared`: no errors.
  - `npx tsc --noEmit -p tsconfig.json` in `server`: one error in `src/sentry.ts`. The same error occurs without this change. This change adds no error.

## Risks

- Low risk. The change is additive: five event types, and no payload shape change for the events that already existed.
- `decision.cancelled` now records an agent on its activity row, where it previously recorded none. This is the same field the sibling decision states already populate, and the cancelling actor is still recorded in `actorType`/`actorId`.
- A plugin that subscribes to a decision event receives the event on a host that includes this change only.
- No data migration. The mapping is an exact set lookup.
- A subscriber that treats `decision.created` as "a decision entity was created" now also sees queue-raised requests. That is the intent — both are a request a person has to answer — but it makes `decision.created` the busiest decision event on an instance that uses the desk, so a noisy rule set produces more of them.
- The issue identity is best effort: an unresolvable issue publishes `issueId` alone, and no identity is published at all rather than a guessed one.

## Model Used

- Provider: DeepSeek. Model ID: `deepseek-v4-flash`. The model ran as a tool-using coding agent (shell, file edit, headless browser automation) inside a coding harness. The harness did not report a context-window size or a reasoning mode.

## Checklist

- [x] All Paperclip CI gates are green — this pull request's CI run is green
- [x] Greptile is 5/5 with no open P2s, recommendations, or follow-ups — Greptile reports 5/5 on the current head

### DIFF
--- a/doc/plugins/PLUGIN_SPEC.md
+++ b/doc/plugins/PLUGIN_SPEC.md
@@ -924,6 +924,10 @@ Minimum event set:
 - `agent.run.cancelled`
 - `approval.created`
 - `approval.decided`
+- `decision.created`
+- `decision.expired`
+- `decision.dismissed`
+- `decision.cancelled`
 - `budget.incident.opened`
 - `budget.incident.resolved`
 - `cost_event.created`
--- a/packages/shared/src/constants.ts
+++ b/packages/shared/src/constants.ts
@@ -1708,6 +1708,14 @@ export const PLUGIN_EVENT_TYPES = [
   "goal.updated",
   "approval.created",
   "approval.decided",
+  // Decisions Desk lifecycle. A decision is the platform's "a human must choose,
+  // by this date" object, so excluding it left notifications unable to cover the
+  // surface operators most need to be woken for. The queue, triage, training and
+  // retention actions stay private: they are desk plumbing, not decisions.
+  "decision.created",
+  "decision.expired",
+  "decision.dismissed",
+  "decision.cancelled",
   "budget.incident.opened",
   "budget.incident.resolved",
   "cost_event.created",
--- a/server/src/__tests__/plugin-decision-events.test.ts
+++ b/server/src/__tests__/plugin-decision-events.test.ts
@@ -1,0 +1,74 @@
+import { describe, expect, it } from "vitest";
+import { PLUGIN_EVENT_TYPES } from "@paperclipai/shared";
+import { eventTypeForActivityAction } from "../services/activity-log.js";
+
+/**
+ * The activity-action → plugin-event mapping is a plugin-facing contract that
+ * fails silently: an unmapped action produces no event, so a subscriber simply
+ * never hears about it and nothing errors. The Decisions Desk was in exactly that
+ * state — `decision.created` was logged with the payload a notifier needs
+ * (`originIssueId`, `originAgentId`, `originResponsibleUserId`) but mapped to
+ * nothing.
+ */
+describe("decision lifecycle plugin events", () => {
+  const decisionActions = [
+    "decision.created",
+    "decision.expired",
+    "decision.dismissed",
+    "decision.cancelled",
+  ] as const;
+
+  it("declares every decision lifecycle action as a subscribable event type", () => {
+    for (const action of decisionActions) {
+      expect(PLUGIN_EVENT_TYPES).toContain(action);
+    }
+  });
+
+  it("maps each decision action onto its own event type", () => {
+    for (const action of decisionActions) {
+      expect(eventTypeForActivityAction(action)).toBe(action);
+    }
+  });
+
+  it("keeps desk plumbing out of the plugin event surface", () => {
+    // Queue, triage, training and retention actions are internal desk mechanics.
+    // Exposing them would invite plugins to depend on queue internals, and each
+    // one would become a compatibility obligation.
+    for (const action of [
+      "decision_queue_item.added",
+      "decision_queue.created",
+      "decision_queue_item.removed",
+      "decision_triage.updated",
+      "decision_training.created",
+      "decision_retention.archived",
+    ]) {
+      expect(eventTypeForActivityAction(action)).toBeNull();
+    }
+  });
+
+  it("does not treat a decision action as a prefix match", () => {
+    // The pass-through is an exact set lookup, not a prefix rule: a future
+    // action must be declared deliberately rather than inheriting an event.
+    expect(eventTypeForActivityAction("decision.created.v2")).toBeNull();
+    expect(eventTypeForActivityAction("decision")).toBeNull();
+  });
+
+  it("still maps the legacy underscore actions it already covered", () => {
+    expect(eventTypeForActivityAction("approval_approved")).toBe("approval.decided");
+    expect(eventTypeForActivityAction("approval_rejected")).toBe("approval.decided");
+    expect(eventTypeForActivityAction("issue_comment_added")).toBe("issue.comment.created");
+    expect(eventTypeForActivityAction("budget_soft_threshold_crossed")).toBe(
+      "budget.incident.opened",
+    );
+  });
+
+  it("passes through the event types that were already exact names", () => {
+    expect(eventTypeForActivityAction("issue.created")).toBe("issue.created");
+    expect(eventTypeForActivityAction("approval.created")).toBe("approval.created");
+  });
+
+  it("leaves unrelated activity unmapped", () => {
+    expect(eventTypeForActivityAction("company.created.note")).toBeNull();
+    expect(eventTypeForActivityAction("some.unmapped.action")).toBeNull();
+  });
+});
--- a/server/src/services/activity-log.ts
+++ b/server/src/services/activity-log.ts
@@ -37,7 +37,14 @@ export function setPluginEventBus(bus: PluginEventBus): void {
   _pluginEventBus = bus;
 }
 
-function eventTypeForActivityAction(action: string): PluginEventType | null {
+/**
+ * Resolve the plugin event type for a logged activity action.
+ *
+ * Exported for direct testing, following `resolveResponsibleUserIdForActivity`:
+ * the mapping is the contract plugins subscribe against, and a silent regression
+ * here is invisible at runtime (the event is simply never delivered).
+ */
+export function eventTypeForActivityAction(action: string): PluginEventType | null {
   if (PLUGIN_EVENT_SET.has(action)) return action as PluginEventType;
   return ACTIVITY_ACTION_TO_PLUGIN_EVENT[action.replaceAll(".", "_")] ?? null;
 }


conreo:
Both findings were valid and are fixed in 23b167de:

**1. Successful decisions stayed silent.** `decision.decided` is produced by the decide path (`decisions.ts`, alongside `decision.dismissed`), but it was not in `PLUGIN_EVENT_TYPES`, so `eventTypeForActivityAction()` returned `null` and a plugin heard about a decision being raised and never about it being answered. It now sits with the other terminal states, is listed in `PLUGIN_SPEC.md`, and the mapping test covers it.

**2. Filtered cancellations were dropped.** `decision.cancelled` was logged without `agentId`. Since the bus matches an `{ agentId }` filter against `payload.agentId` for non-agent entities, a plugin subscribed to one agent's decisions never received a cancellation. The producer now records the *origin* agent (`updated.originAgentId`), not the cancelling actor — matching `decision.dismissed` and `decision.expired` — and the payload carries `originAgentId` the way `decision.created` does.

`passesFilter` is now exported so the delivery rule can be asserted directly, with a test that a cancellation payload carrying the origin agent passes `{ agentId }` and that the same event without it does not. That rule fails silently when a producer and a subscriber disagree about it, which is how this slipped through in the first place.

Verification: CI re-running on the push. Locally the changed files parse and the token gates are clean; the monorepo dependencies are not installed in this checkout, so the test run is CI's.

conreo:
Greptile was right about the test, and about my own mess. Both are fixed in 5056d6d0, which replaces the two fix commits that were here before (the branch is rewritten rather than reverted, so the artifacts never appear in the diff).

**The test was wrong, not the producer.** It built its event with `payload: { originAgentId }` while `passesFilter` reads `payload.agentId` — the assertion it was written for could never pass. It now mirrors what `logActivity` actually emits (`agentId` spread onto the payload root, plus `originAgentId` in the details), its negative case uses the pre-PR shape (`agentId: null`) instead of an empty object, and an empty filter is asserted to pass.

The reason it shipped broken is worth naming: this checkout has no `node_modules`, so I only *parsed* the test file and left running it to CI. `passesFilter` is now extracted verbatim and executed against the four assertions locally — all four pass — because "it parses" is not "it passes".

**Artifacts removed.** A `git add -A` swept a scratch harness (`.cache/fn.ts`, `.cache/fn.mjs`, `.cache/run-filter.mjs`) and `packages/adapter-utils/package-lock.json` into the previous commits. They are gone from the branch, and the paths are excluded locally so they cannot come back.

The PR diff is now six files: the shared event constant, the plugin spec, the decision producer, the event-bus predicate export (for the filter test), the mapping test, and the activity-log export from the original commit.

conreo:
Requesting review. This is ready: `mergeStateStatus: CLEAN`, the full `ci` workflow green (build, typecheck, all server suites, serialized suites, e2e, canary dry run), Greptile 5/5 with no open findings, and the PR template filled in.

**What it changes, in one line each**

- Declares the decision lifecycle on the plugin event surface: `decision.created`, `decision.decided`, `decision.expired`, `decision.dismissed`, `decision.cancelled`. The actions are already logged by `decisions.ts` with the payload a subscriber needs (`originIssueId`, `originAgentId`, `originResponsibleUserId`), but they are in neither `PLUGIN_EVENT_TYPES` nor the underscore bridge, so nothing is emitted and the failure is silent — no error, just an event that never arrives.
- Records the origin agent when logging `decision.cancelled`. The bus matches an `{ agentId }` filter against `payload.agentId` for non-agent entities, and that producer passed no agent, so an agent-filtered subscriber would have missed every cancellation. `decision.dismissed` and `decision.expired` already do this.
- Tests the mapping, the boundary (queue/triage/training/retention actions stay private), and agent-filtered cancellation delivery.

**Why this is the right mechanism rather than a bridge entry.** `eventTypeForActivityAction` returns an action unchanged only when `PLUGIN_EVENT_SET` holds it, and the returned value is what a subscription filter matches on, so the declaration is the change — the bridge cannot introduce an event that was never declared.

I have no write access here, so I cannot merge this myself; it needs an owner. If the decision lifecycle should be discussed elsewhere first, or the surface should stay narrower (for example, declaring only `decision.created` and `decision.decided`), say so and I will cut it down. @cryppadotta @devinfoley @nickyleach — happy to adjust whatever would make this reviewable.

conreo:
Fixed in 53546152 — the finding was valid, and the bug was worse than described: `issue` is not a queue source at all, so that branch was dead code and *every* kind except `issue_thread_interaction` was being looked up in the wrong table.

Two options were available. The first was to resolve each kind properly here, which means rebuilding the attention layer's mapping inside this service — a decision keeps its link in `originIssueId`, an approval reaches it through `issueApprovals`, a failed run keeps it inside `contextSnapshot`, and so on. That mapping belongs to the layer that owns it, and duplicating it is how the two drift apart.

So the manual add path now publishes its source reference and nothing it cannot prove, with a comment recording why. Seed-raised requests still carry the issue, because that path resolves it per attention item through the helper the attention layer itself uses (`itemIssueId`) — that is the path that produces essentially all of the traffic (122 seed rows against 2 manual adds on the instance this was built against).

`PLUGIN_SPEC.md` now says exactly that: a seed-raised item carries `issueId` with `identifier` and `title` when the issue could be read; a manually added item carries only its source. `vitest run src/__tests__/plugin-decision-events.test.ts src/__tests__/decision-queues-routes.test.ts` passes (15 tests).

conreo:
Two findings carried over from an earlier revision are no longer part of this branch: the `.cache/run-filter.mjs` helper and the stray `packages/adapter-utils/package-lock.json` were both dropped when those commits were rewritten, and head `5056d6d0` has contained neither since. The current diff is four changed files plus one test:

- `packages/shared/src/constants.ts`
- `server/src/services/activity-log.ts`
- `server/src/services/decision-queues.ts`
- `doc/plugins/PLUGIN_SPEC.md`
- `server/src/__tests__/plugin-decision-events.test.ts`

The P1 raised on the previous head — `resolveSourceIssueIdentity` guessing which issue a queue source belonged to — was fixed in `53546152`: the manual-add path no longer resolves an issue it cannot prove, and the seed path keeps the identity it already resolves per attention item.

The confidence check reads 4/5 against this head, below the repository's required 5/5. @greptileai review

PR #13312, opened 2026-09-12
fix(shared): don't mangle escaped newlines inside valid JSON document bodies

notandrewblejde:
## Thinking Path

> - The shared validators package normalizes free-text fields on write, including issue document bodies.
> - One validator, `normalizeEscapedLineBreaks()`, rewrites every literal `\n`/`\r` two-character sequence into a raw newline byte. This is correct for prose but wrong for a document body that is itself a JSON string, because it destroys a required JSON escape inside an embedded string value.
> - A downstream integrator reported that a JSON document body no longer parsed after a round trip through the documents API: the stored `narrative` field contained a raw control character where an escaped newline should be.
> - This corruption is silent. Markdown rendering of the body still looks correct, so nothing alerts a human. Only a consumer that calls `JSON.parse` on the stored body fails, and only after the value is already persisted.
> - This pull request gates the rewrite on `JSON.parse` succeeding, so a JSON-shaped body round-trips byte-for-byte and skips the transform, while non-JSON text keeps the existing normalization.
> - The benefit is that any document body that is a JSON payload survives a write/read round trip intact, without changing behavior for plain markdown bodies.

**What happened?**
A document body written through the issue documents API as a JSON string (produced by `JSON.stringify`/`json.dumps` on the client) came back from a later `GET` with its escaped newlines (`\n`) turned into raw newline bytes inside the JSON string values. The returned body was no longer valid JSON: parsing it raised an "Invalid control character" error.

**Expected behavior**
A document body that is valid JSON on write must remain valid JSON on read. The API must not alter the byte content of a JSON-shaped body.

**Steps to reproduce**
1. Build a JSON string whose payload contains a multi-line text field, for example `JSON.stringify({ narrative: "line one\nline two" })`.
2. `PUT` an issue document with that string as `body` and `format: "markdown"`.
3. `GET` the document back and run `JSON.parse` on `document.body`. Before this fix, the parse fails with "Invalid control character in string" because the escaped `\n` was rewritten to a literal newline byte at write time in `upsertIssueDocumentSchema` (`packages/shared/src/validators/issue.ts`).

**Related PRs found while searching**
- #8205 (open, stale since June) removes `normalizeEscapedLineBreaks` entirely so all multiline text is stored verbatim, including plain markdown. That's a broader behavior change than this PR makes.
- #13213 / #13214 (both closed, not merged, self-closed by the author as "not pursuing an upstream contribution") took a much larger approach: a new `format` field (`text`/`json`/`ndjson`) threaded through the schema, CLI, and UI so verbatim formats bypass normalization entirely.
- This PR is a narrower, targeted fix: it keeps the existing `normalizeEscapedLineBreaks` behavior for plain text and only skips it when the input already parses as JSON, with no schema/type/CLI/UI surface changes.

## What Changed

- `packages/shared/src/validators/text.ts`: `normalizeEscapedLineBreaks()` now checks whether the incoming text parses as valid JSON via `JSON.parse` before rewriting literal `\n`/`\r` sequences. If it parses as JSON, the function returns the input unchanged so the JSON escaping round-trips byte-for-byte. If it does not parse as JSON, the existing literal-newline normalization still runs, so plain markdown/prose text is unaffected.
- `packages/shared/src/validators/issue.test.ts`: added a regression test asserting a JSON document body (with an embedded newline inside a string value) round-trips byte-for-byte through `upsertIssueDocumentSchema` and still parses with `JSON.parse`.

## Verification

- `packages/shared`: `npx vitest run src/validators/issue.test.ts` — 31 passed, including the new regression test.
- `packages/shared`: `npx vitest run` (full package) — 728 passed.
- `packages/shared`: `npx tsc --noEmit` — clean.
- Confirmed the existing test covering literal-`\n` markdown normalization (`"# Plan\\n\\nShip it"` → `"# Plan\n\nShip it"`) still passes unchanged, so plain-text document bodies keep their current behavior.

## Risks

- Low risk. The change only narrows an existing transform: it now skips the newline rewrite for the subset of inputs that are valid JSON, and leaves all other inputs on the previous code path.
- A body that happens to parse as valid JSON but was intended as plain prose (for example, a document whose entire text is the digits `"123"` or the word `"true"`) would now skip the literal-`\n` normalization. This is an unlikely shape for a real document body and is strictly safer than silently corrupting JSON payloads.
- No schema or storage migration is involved; this only changes how new writes are normalized. Already-corrupted documents written before this fix are not repaired by this change.

## Model Used

Claude, Sonnet 5 (`claude-sonnet-5`), standard (non-extended) thinking mode, with tool use for reading the validator source, running the test suite, and typechecking.

## Checklist

Co-Authored-By: Claude Code <claude@users.noreply.github.com>

### DIFF
--- a/packages/shared/src/validators/issue.test.ts
+++ b/packages/shared/src/validators/issue.test.ts
@@ -442,6 +442,27 @@ describe("issue validators", () => {
     expect(document.body).toBe("# Plan\n\nShip it");
   });
 
+  it("preserves escaped newlines inside a JSON document body byte-for-byte", () => {
+    // Regression test for SPC-39026: a document body that is itself a JSON
+    // string (e.g. `json.dumps(brief, indent=2)`) must round-trip exactly —
+    // the `\n` inside the `narrative` value is a required JSON string escape,
+    // not literal text a client failed to turn into a line break, so it must
+    // not be rewritten into a raw control character.
+    const jsonBody = JSON.stringify(
+      { narrative: "Line 1\n\nLine 2", note: "keep\\nliteral" },
+      null,
+      2,
+    );
+
+    const document = upsertIssueDocumentSchema.parse({
+      format: "markdown",
+      body: jsonBody,
+    });
+
+    expect(document.body).toBe(jsonBody);
+    expect(() => JSON.parse(document.body)).not.toThrow();
+  });
+
   it("clamps oversized requestDepth values on create", () => {
     const parsed = createIssueSchema.parse({
       title: "Clamp request depth",
--- a/packages/shared/src/validators/text.ts
+++ b/packages/shared/src/validators/text.ts
@@ -1,6 +1,19 @@
 import { z } from "zod";
 
 export function normalizeEscapedLineBreaks(value: string): string {
+  // Skip valid JSON payloads: their `\n`/`\r` sequences are structural string
+  // escapes, not literal text a client mistakenly failed to turn into a real
+  // line break. Rewriting those bytes turns escaped-newline JSON into raw
+  // control characters, breaking `JSON.parse`/`json.loads` for any consumer
+  // reading the stored value back as data (see SPC-39026).
+  try {
+    JSON.parse(value);
+    return value;
+  } catch {
+    // Not JSON — fall through to the legacy literal-escape normalization
+    // used for human/agent-authored markdown text.
+  }
+
   return value
     .replace(/\\r\\n/g, "\n")
     .replace(/\\n/g, "\n")


notandrewblejde:
**CEO review — APPROVED** (per standing board directive, SPC-36835 comment 333c16fc, CEO handles PR review/merge).

Reviewed the diff: `normalizeEscapedLineBreaks` was rewriting `\n` escape sequences into raw control characters even when the document body was itself a serialized JSON payload — the exact corruption in SPC-39026. Guarding with a `JSON.parse` probe and skipping normalization for valid JSON is the right minimal fix, and the regression test asserts byte-for-byte round-trip plus re-parseability. The only edge case (a markdown body that coincidentally parses as JSON, e.g. a bare number) carries no escaped newlines, so behavior there is unchanged in practice.

Note for follow-up: this fixes the write path going forward — previously stored brief-iso documents (SPC-39016 revisions, SPC-38900) remain corrupted until republished.

Proceeding to merge.

